PR #8, opened 2015-04-06
fix observe to return a well-behaved subscription

vsavkin:


blesh:
Heh... yeah I got a little lazy there. Thank you Victor.

PR #11, opened 2015-04-09
In ES6 semantics, Derived Class must call 'super()' before to use 'this'  in constructor

saneyuki:


blesh:
Thank you so much for this pull request. I wasn't able to merge this one, directly, and I had to make those changes anyhow after updating BabelJS.

saneyuki:
> Thank you so much for this pull request. I wasn't able to merge this one, directly, and I had to make those changes anyhow after updating BabelJS.

okay!

PR #15, opened 2015-04-22
[DO NOT MERGE] This PR created for discussion and comparison

blesh:


src/Observable.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+    
+    Observable.create = function(subscribe) {
+        return new Observable({ subscribe: subscribe });
+    };
+    
+    Observable.template = function(SubscriberType) {
+        function ObservableTemplate(source) {
+            var argsIdx = 0;
+            var argsLen = arguments.length;
```
blesh: non-monomorphic function use of arguments /@jhusain

src/Observable.js
```diff
@@ -0,0 +1,69 @@
...
+    function Observable(source) {
+        this.source = source;
+    }
+    
+    Observable.create = function(subscribe) {
+        return new Observable({ subscribe: subscribe });
+    };
+    
+    Observable.template = function(SubscriberType) {
+        function ObservableTemplate(source) {
```
blesh: Why are we not just using standard JS sub classing? ObservableTemplate function is created multiple times.

src/Subscriber.js
```diff
@@ -0,0 +1,137 @@
+var noop = require("src/support/noop");
+var Disposable = require("src/Disposable");
+
+module.exports = function() {
+    
+    function destNext   (x) { return this.destination["next"](x);  }
+    function destThrow  (e) { return this.destination["throw"](e); }
+    function destReturn ( ) { return this.destination["return"](); }
+    
+    function Subscriber(onNextOrDest, onError, onCompleted, dispose) {
```
blesh: Why isn't `return` the same as `dispose`?

src/Disposable.js
```diff
@@ -0,0 +1,98 @@
...
+        this.length = 0;
+    }
+    
+    Disposable.prototype.dispose = function() {
+        if(!this.disposed) {
+            var result = true, f, xs, i, n;
+            (f = this._dispose) && (result = f.call(this));
+            if(!!result || result === void 0) {
+                this.disposed = true;
+                delete this._dispose;
```
blesh: deleting causes deoptimization in V8

src/Disposable.js
```diff
@@ -0,0 +1,98 @@
...
+    };
+    
+    Disposable.prototype.add = function () {
+        
+        var argsIdx = -1;
+        var argsLen = arguments.length;
+        var disposables = new Array(argsLen);
+        while(++argsIdx < argsLen) { disposables[argsIdx] = arguments[argsIdx]; }
+        
+        var xs = this._disposables || (this._disposables = []),
```
blesh: This will also cause deoptimization in V8. Adding members in an unknown order causes creation of a new "hidden class".

We should have a Disposable and a CompositeDisposable which inherits from it.

src/Disposable.js
```diff
@@ -0,0 +1,98 @@
...
+                (xs[j++] = x));
+        }
+        this.length = j;
+        return this;
+    };
+    
+    Disposable.prototype.remove = function () {
+        
+        var argsIdx = -1;
+        var argsLen = arguments.length;
```
blesh: Just again: Using arguments makes the function non-monomorphic. Just accept an array.

src/Disposable.js
```diff
@@ -0,0 +1,98 @@
...
+        
+        var xs = this._disposables || (this._disposables = []),
+            ys = disposables, j = xs.length, i = -1, n = ys.length,
+            xd = this.disposed, x;
+        while(++i < n) {
+            // 
+            // expression-ize all the things.
+            // 
+            // get the disposable to add
+            (x = ys[i]) &&
```
blesh: Minor thing: Almost all minifiers will convert if/thens into expressions. Expressions can be harder to grok and may turn away contributors. It might be better to use standard if/then syntax for readability.

src/observable/from-array.js
```diff
@@ -0,0 +1,59 @@
...
+            result = state ? state[3] : { done: false, value: undefined };
+        
+        subscriber = state ? state[0] : subscriber;
+        
+        if(subscriber.disposed) {
+            result.done = true;
+            return result;
+        }
+        
+        if(scheduler) {
```
blesh: should also compare against global immediate scheduler

src/observable/range.js
```diff
@@ -0,0 +1,66 @@
...
+        } else if(scheduler = this.scheduler) {
+            return scheduler.schedule({
+                to: to,
+                from: from,
+                result: result,
+                subscriber: subscriber
+            }, _subscribe);
+        } else {
+            while(++from < to) {
+                result = subscriber.next(from);
```
blesh: "There is no try" - yoda

We need to be doing tries and sending caught errors to `throw()`. Try blocks are not optimized by Chrome, so these benchmark numbers are probably off by a bit. (I'm not actually sure I'm doing this in my prototype, either)

src/observable/from-array.js
trxcllnt: @blesh why?

src/Observable.js
trxcllnt: Yes, `ObservableTemplate` is a shortcut for creating each specific `Observable` subclass for each operator, e.g. `MapObservable`, `ZipObservable`, etc., since its job is really just to instantiate a specific `Generator` constructor with the appropriate args.

src/observable/from-array.js
blesh: I'm trying to remember... I think the idea was there there would also be a global scheduler. I'm not really sold on that idea, though. But if there were a global scheduling override, we'd need to check it here.

PR #16, opened 2015-04-26
use prototype instead of 'this' before super

BerkeleyTrue:
'this' in subclass constructor before call to super exits is undefined
according to spec. Use prototype object to pass in method instead to
comply with spec. 
See https://github.com/babel/babel/issues/1284 for more details

### DIFF
--- a/src/observable/observable.js
+++ b/src/observable/observable.js
@@ -75,7 +75,7 @@ Observable.return = function(value) {
 
 export class ScheduledObservable extends Observable {
   constructor(source, observationScheduler) {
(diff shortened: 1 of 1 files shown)


blesh:
@BerkeleyTrue  --- thank you for this PR... we're actually migrating the project over to TypeScript (so we can transpile down to ES6, ES5 CJS, ES5 AMD and global files *and* support TypeScript .d.ts file output. Current work is being done in the `typescriptify` branch until it's stabilized enough to merge into `master`.

PR #27, opened 2015-05-26
fix global build (fixes #22)

### LINKED ISSUE #22: Fix global build
Currently the rx.ts file is manually exporting Observable and Subscription constructors on the window object.

blesh:
Removed babelify, and just use straight babel. Probably should use a local install of babel though.


blesh:
Actually, this is going to break tests, @vsavkin and @jeffbcross ... because I changed where the global was installing to Rx3.Observable ... Ideally, it will be Rx.Observable at some point...

blesh:
Tests are all straightened out now. Much better.

PR #32, opened 2015-06-10
Typescriptify refactor tayloresque

blesh:
After much toil.

Tests still not working... but I want to put my foot in the door.


blesh:
Oops... wrong branch.

PR #33, opened 2015-06-10
Typescriptify refactor tayloresque

blesh:
Incomplete... just getting my foot in the door for review.

Tests still need fixed/updated.


blesh:
I'm going to go ahead and merge this so Paul and I get split up efforts. It's in an okay place now, even though tests are still borked.

PR #39, opened 2015-06-16
Adds fromArray and of creation operators. Adds combine and zip operators. Adds tests. Fixes typos.

trxcllnt:


blesh:
I didn't merge this because of the ongoing discussion about the higher-kindness issues, and the flux of the TC39 spec.  But I did try to use it as a reference point for the work. Thanks, so much, @trxcllnt, your input and expertise on this project has been insanely valuable. In particular, working through the code you originally wrote that I ported for the beginnings of this project.

PR #86, opened 2015-07-17
Lift: Join Observer and Subscription into Subscriber

blesh:
All tests working.

There is a *LOT* that can be improved in this, I think.

Things to look at: Subject and ConnectableObservable. I hacked my way through these, they work, but I think they require review. @trxcllnt please have a look.

... also, I feel like we're about to have a few PRs that are trampling on one another? I hope not.


blesh:
Wrong branch....

PR #87, opened 2015-07-17
Lift: Join Observer and Subscription into Subscriber

blesh:
Tests are working, perf looks good.

This requires review. I hacked my way through Subject and ConnectableObservable. attn @trxcllnt 

There is a lot that can be improved, but this is a good starting point.


blesh:
Actually, perf is so fast, I have my doubts it's doing anything... 16x faster than what's in master? Doesn't seem right.

blesh:
Oh... another note, I think I completely commented out `fromEvent` because it needed more refactoring and I didn't reimplement it. This is okay, IMO, because there were no tests for `fromEvent` anyhow. We should probably set up a seperate issue to add `fromEvent` with tests. (also `fromEventPattern` doesn't have tests)

blesh:
This is an awful lot of commits now because I had to rebase `lift` from `master` and my branch from `lift`

blesh:
Merging because it's just to the `lift` branch. Also, it works, where the `lift` branch does not, yet. I'd really like to get some review on what I've done in the `lift` branch prior to merging to master though.

Issue #93, opened 2015-07-17
Run perf tests against previous build/version of RxJS Next


blesh:
So rather than just test against RxJS 2, perhaps it would be better to also run against the most recent published and/or master builds of RxJS Next?

I think to do this, we'd have to change the build target directory to something like `tmp/` which we wouldn't commit. Then use what's in there for the "local" test, vs "latest", "master" and "Rx2". Thoughts?

Pertains to some comments here #90

blesh:
I think this should definitely happen before we move to Beta. By the time we hit Beta, everything should be mostly stable and a new performance baseline can be established.

ktrott:
I'd like to see this even sooner as without it, we don't know build over build if we're improving/regressing our performance.

ktrott:
@trxcllnt going to look at the work that we did for Falcor and reuse some of that here.

blesh:
@ktrott has any of that been OSS'ed yet?

ktrott:
Still a private repo for the moment, but @trxcllnt has access

blesh:
This should really be something for the initial beta... since we won't really have something to run against other than RxJS 2/3 until this library is published to npm as at least an `alpha` version.

blesh:
@jeffbcross do you have anything you're doing on the angular team to do perf tests against previous versions of angular in a semi-contained environment? Can you use Travis for that?

PR #94, opened 2015-07-17
chore(dest): remove dest folder from committed source

### LINKED ISSUE #90: Remove dist folder from source
If we _must_ have dist as part of the repo, it should at least be something that only gets updated at each tag, so that deltas between branches are less drastic (and rebases are less difficult). But I propose keeping no build folders in the project, and just make a script that generates the npm distribution at publish-time. I can implement this script, similar to how we handle it in https://github.com/angular/angular/blob/master/scripts/publish/npm_publish.sh

@blesh

jeffbcross:
Closes #90


blesh:
I'd like to discuss this more in #90 and #93. I agree that we should do something about this so it's not crazy to rebase/diff.

blesh:
Closing this one for now until we can discuss #93 and #90. Also, it can't be automatically merged.

I don't disagree with the spirit of this PR at all.

PR #103, opened 2015-07-18
implement partition operator

robwormald:
Implements .partition operator because I have impulse control problems. Does not include dist/ changes per #94

### DIFF
--- a/spec/operator/partition-spec.js
+++ b/spec/operator/partition-spec.js
@@ -2,7 +2,7 @@
 var RxNext = require('../../dist/cjs/RxNext');
 var Observable = RxNext.Observable;
 
-describe('Observable.partition.filter()', function(){
+describe('Observable.prototype.partition()', function(){
    
    it('should partition an observable into two using a predicate', function(done){
        
--- a/spec/operator/partition-spec.js
+++ b/spec/operator/partition-spec.js
@@ -1,0 +1,85 @@
+/* globals describe, it, expect */
+var RxNext = require('../../dist/cjs/RxNext');
+var Observable = RxNext.Observable;
+
+describe('Observable.partition.filter()', function(){
+   
+   it('should partition an observable into two using a predicate', function(done){
+       
+       var expectedPositive = [0, 1, 2];
+       var expectedNegative = [-3, -2, -1];
+       var completed = false;
+       
+       function completer(){
+           if(completed){
+               expect(positiveValues).toEqual(expectedPositive);
+               expect(negativeValues).toEqual(expectedNegative);
+               done();
+           }
+           else{
+               completed = true;
+           }
+       }
+       
+       var values = [-3, -2, -1, 0, 1, 2];
+       
+       var numberStream = RxNext.Observable.fromArray(values);
+       
+       var streams = numberStream.partition(function(value){
+           return value >= 0;
+       });
+       
+       var positiveStream = streams[0];
+       var negativeStream = streams[1];
+       
+       var positiveValues = [];
+       var negativeValues = [];
+       
+       positiveStream.subscribe(function(value){
+           positiveValues.push(value);
+       }, null, completer);
+       
+       negativeStream.subscribe(function(value){
+           negativeValues.push(value);
+       }, null, completer);
+
+   });
+   
+   it('should pass errors to both returned observables', function(done){
+       
+       var values = [-3, -2, -1, 0, 1, 2];
+       var numberStream = RxNext.Observable.fromArray(values);
+       var errored = false;
+       
+       function rejecter(){
+           expect(true).toBe(false);
+       }
+       
+       function completer(error){
+           if(!errored){
+               errored = error;
+           }
+           else{
+               expect(errored).toBe(error);
+               done()
+           }
+       }
+       
+       var streams = numberStream
+       .map(function(value){
+           throw 'error'
+       })
+       .partition(function(value){
+           return value >= 0;
+       });
+       
+       var positiveStream = streams[0];
+       var negativeStream = streams[1];
+       
+       positiveStream.subscribe(rejecter, completer, rejecter);
+       negativeStream.subscribe(rejecter, completer, rejecter);
+       
+    });
+   
+    
+});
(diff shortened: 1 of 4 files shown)


blesh:
It's well developed. I'm unsure of the utility. Seems like groupBy would be a better choice. I don't want to unilaterally make a decision though. @trxcllnt @mattpodwysocki

headinthebox:
The advantage of partition is that you know there are *precisely* two groups.

blesh:
Would it be better if it returned an Observable with the array in it? it just seems I can do this with a filter or groupBy. 

What are identified use cases? Perhaps that would help me understand the utility

robwormald:
I stumbled across partition when messing with angular2 and http for the first time doing a boring autocomplete type scenario:
```js
const queryStream = someFormControl.valueChanges.toRx();
// "c"
// "ca"
// "cat"
```
piped into an http service to make queries against youtube
```js
const searchResults = 
queryStream.map(search => `search?q=${inputString}`).flatMapLatest(searchYouTube);
// [video, video]
// [video, video, video]
// [video, video, video, video]
```
with ng2 i can bind an observable directly to the view (nice) to display those results
```js
this.searchResults = 
  queryStream
    .map(search => `search?q=${inputString}`)
    .flatMapLatest(searchYouTube);
```
```html
  <ul>
    <li *ng-for="video of searchResults | async">{{video.title}}</li>
  </ul>
```
i wanted to then "reset" the search results (with an empty array) when somebody cleared the input:
```js
const queryStream = someFormControl.valueChanges.toRx();
// "cat" -> [video, video, video]
// "ca" -> [video, video, video]
// "c" -> [video, video, video]
// "" -> ???
```
my (incredibly noobish) first attempt i'm not even going to post here, but @robertpenner plugged in partition for me to end up with the following - this wouldn't be nearly as useful without es6 destructuring.
```js
const queryStream = someFormControl.valueChanges.toRx();

//partition
const [searchesToMake, emptyInputs] = queryStream.partition(query => (query.length));

//"parallel" streams 
//turn empty strings into empty arrays
const emptySearchResults = emptyInputs.map(() => []);
//turn search strings into query results
const youTubeSearchResults = searchesToMake.map(makeQueryUrl).flatMapLatest(searchYouTube);

//join back together and bind to the view
this.searchResults = youTubeSearchResults.merge(emptySearchResults);
// "c" -> [video, video, video]
// "ca" -> [video, video, video]
// "c" -> [video, video, video]
// "" -> []
```
The use case here is effectively "forking/splitting" the stream based on a condition - it feels to me like a stream based if/else statement, vs a groupBy which feels more like a switch or if/...if else/else statement. 

Other use cases I can think of (and this might be of interest to @jeffbcross re: ngHttp) 

I'm unabashedly a promise fan because of the clear resolve / reject partitioning - things either succeed or fail - they may fail for any number of reasons but that first "fork" is quite clear to me - now I'm getting all Reactive partition seems like a nice extension of that concept: 
```js
//isValid with promises
function checkStatus(response) {
  if (response.status >= 200 && response.status < 300) {
    return response
  } else {
    var error = new Error(response.statusText)
    error.response = response
    throw error
  }
}
let fetchResults = fetch('/someurl').then(checkStatus).then(res => res.json());

fetchResults.then(doStuffWithSuccesses, doStuffWithErrors);
```
the Rx version's mental model is better - its not so much an "error" (throw) as much as "a result other than the expected one":
```js
const [validResponses, invalidResponses] = someHttpQueryStream.partition(res => (res.status >= 200 && response.status < 300));
//use responseSuccess and responseError "interceptors" cleanly 
const jsonStream = validResponses.map(parseSuccessResponse);
const errorStream = invalidResponses.map(parseErrorResponse);
const responseStream = jsonStream.merge(errorStream);
```
You might use groupBy downstream from that for handling different sorts of errors, as there are many possible error codes:
```js
let errorStream = invalidResponses.map(parseErrorResponse).groupBy(res => res.statusCode)
```

Example quoted in the Rx docs is modulo -
```js
const numbers = [1, 2, 3, 4, 5]
const [evens, odds] = numbers.partition(x => x % 2 === 0);
```

Wayyy above my head but I think what I'm trying to explain (poorly) is something along the lines of https://en.wikipedia.org/wiki/Equivalence_class - partition is a way to express "is" or "is not" - which is different than "belongs to". A series of shapes could be subdivided into "squares" and "not squares", which is different to "squares", "triangles" and "hexagons". There are two and only two sets in the first, and any number of possible sets in the latter. 90% of that article is greek to me though, so I dunno. 

@blesh I would be curious in either case to see the groupBy equivalent - I think part of your point is that in a groupBy scenario the 2nd ("else") observable would be dynamically created as needed? Unless I'm missing something though that would require you to do some sort of check downstream from the groupBy, or some sort of if/else statement? 

See also the discussion where it got added to Rx : https://github.com/Reactive-Extensions/RxJS/issues/141 - the OP puts its much more succinctly than my mad rambling :

"it allows to switch between 2 (or in the case of groupBy, n) observables depending on condition, I want to conditionally notify 2 different observer depending on a condition." 

Assuming the code is valid I think it's worth the 3 lines, though I'd be quite happy to implement it myself if there's a nice way to plug it in from userland. If the ES7 fnBind operator gets in this sort of thing would go nicely in a standalone aggregator type lib, or whatever the solution is for doing it nicely (lift?) 
```js
const partition = (predicate, thisArg) => [this.filter(predicate), this.filter(x => !predicate.call(thisArg, x))];
const [successes, failures] = queryStream::partition(res => (res.status >= 200 && response.status < 300));
```
my 3c. Plunker of the original use case, looks nasty because I'm having to plug it in in the middle of the logic. http://plnkr.co/edit/5FfPoA?p=preview

apologies on the wall of text. this stuff is fun :D

mattpodwysocki:
@robwormald I'm all in favor of it since we did indeed implement it in RxJS, so why it would be excluded would be beyond me.  Totally in agreement with @headinthebox here

blesh:
> since we did indeed implement it in RxJS

Fair enough, @mattpodwysocki. I must have missed that one. Sounds good to me.

PR #105, opened 2015-07-18
Fix devDependencies and mention of global packages

staltz:
It is not necessary to have those prerequisite packages installed globally. When running npm scripts, the bin (e.g. browserify) is found automatically in /node_modules/.bin/, before it tries to find it globally. This commit also includes browserify as a devDependency, since it was the only one missing in order for
`build_all` to run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,14 +36,6 @@ The build and test structure is fairly primitive at the moment. There are variou
 npm run build_all
 ```
 
-### Prerequisites
-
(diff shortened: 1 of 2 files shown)


blesh:
Fair enough. I think this whole build pipeline needs a refactor/rewrite. But it works for now.

blesh:
... actually that's sort of what the global install thing stemmed from... just pushing it out the door.

PR #106, opened 2015-07-19
Add combineLatest (static and proto operator)

staltz:
First draft implementation, just to get things started.
Basic functionality is in place.


blesh:
Fantastic. I think this will do for now, however, I think we'll need to revisit this and other operators like `zip`, for example and reimplement them with `lift`. I'll merge this, but put an issue up.

Issue #116, opened 2015-07-20
Rename SubscriberFactory to Operator


blesh:
As @trxcllnt pointed out, in RxJava, what we're calling SubscriberFactory is just called Operator. This makes total sense, because the only type you'd make a new SubscriberFactory is when you're creating or adding a new Operator (for the most part).

The name "SubscriberFactory" was born out of me simply converting things over from Observer to Subscriber. Nothing more.

Thoughts @benjchristensen @trxcllnt?

Issue #112, opened 2015-07-20
All Creation Operators must use a Scheduler


mattpodwysocki:
I'm noticing operators such as `empty` lack a scheduler, which is required because you can parameterize when/where empty happens.  All creation operators need a scheduler.

blesh:
Yes, @trxcllnt and I had them in there, then removed them as a performance spike, then discussed re-adding them, but having gotten around to it. The idea that @trxcllnt and I discussed was creating fast-paths for things like the range operator if the `immediate` scheduler is being used.

mattpodwysocki:
@blesh that's fine, but the default should be in there especially if you move to a `nextTick` scheduler.

blesh:
In fact, labeling this as a bug, because it should have been re-added.

Issue #109, opened 2015-07-20
Convert `combineLatest` to use `lift`


blesh:
The static `combineLatest` and it's instance counterpart need to be refactored to use `lift`

Issue #108, opened 2015-07-20
Convert `zip` to `lift` method


blesh:
The static `zip` method and it's instance counterpart need to be refactored to use `lift`.

PR #120, opened 2015-07-27
Adds generics, Observer, Subscription, Subscriber, Operator, and Observable type classes, and updates all operators to use the Observable/Operator pair.

### LINKED ISSUE #69: Proposal: Implement CompositeSubscription behavior in Subscription, remove other Subscription types
Realistically, we can model all subscription logic with CompositeSubscription. Promoting composite behavior to the Subscription base class will reduce file size and complexity.

We can reduce the number of Subscription allocations by sharing a single Subscription instance between all intermediate Observers per subscribe call. Some operators don't allocate resources that need to be disposed. The ones that do (the flattening, splitting, and coincidence strategies) can add their inner Subscriptions to the Subscription shared between all Observers. This also reduces complexity inside those operators, as they don't have to worry about removing their inner subscriptions from the shared Subscription, because they'll be cleaned up when the shared Subscription is disposed (on throw, return, or dispose).

This is probably only feasible if Subscriptions store their inner subscriptions in a HashSet for O(1) insertion and removal.

trxcllnt:
Fixes #69, #70, #75, #72, #82, #96, #108, #109, #112, and #116.


jeffbcross:
Holy giant commit!

staltz:
Next time it would be friendly to use Github's *Assignee* feature before starting to fix the issue so multiple contributors don't do the same task in vain.

blesh:
> Next time it would be friendly to use Github's Assignee feature before starting to fix the issue so multiple contributors don't do the same task in vain.

I agree. I knew of some of the things that @trxcllnt was working on, and I had assigned them to myself. But I didn't know the full scope until yesterday.

blesh:
Either way, I'm thankful for this commit. We just need to keep it to smaller bites from here on out.

Issue #141, opened 2015-07-29
Improving `catch`, possibly removing `retryWhen`


blesh:
The proposal is a non-breaking change to catch:

**EDIT: see [comment below](https://github.com/ReactiveX/RxJS/issues/141#issuecomment-126120512)**

blesh:
cc/ @mattpodwysocki @staltz @trxcllnt for thoughts.

This was something I came up with while reimplementing the `retryWhen` operator today and discussing @trxcllnt's alternate version of that operator he called `retryMany`. I realized that both of these could be accomplished with `catch` if catch simply had a second argument of the source observable. @trxcllnt pointed out that it would have to be the "caught" observable, or really the whatever the last `catch` call returned, intialized with the source observable. Otherwise on the second catch, it would skip the catch operator all together.

blesh:
@zenparsing and @jhusain may find this issue interesting for their es-observable efforts as well, since it's a basic operator.

blesh:
So in working with this and talking with George Campbell at Netflix, it seems like the optional arguments should/could be:

```js

myObservable.catch((err:any, source:Observable<T>, caught:Observable<any>) => Observable<any>) => Observable<any>

```

Where the arguments are as follows:

- error: the error value
- source: the original source observable
- caught: the observable that is the result of the initial source.catch() call

With these arguments one can do just about any type of retry or catch handling. However, it's worth pointing out that any of this could be accomplished with closure as well. It's just that inside of the `catch` operator, we already have access to all of those observables, so we can easily pass them in without any closure.

### I'm torn on this one:

#### Pros:

- It's a non breaking change to catch
- It adds very little complexity
- It enables retryWhen functionality without an additional operator, and is perhaps easier to explain than retryWhen

#### Cons:

- While it's probably easier to explain than retryWhen, it's still pretty hard to explain. Like "why return the 'caught' Observable?", etc
- ??? 


I'm sure there are more cons... I'm waiting for others to weigh in.

blesh:
To elaborate:

ordinary catch:

```js
myObservable.catch(err => {
  if(isBadError(err)) {
    throw err;
  }
  return Observable.value('something happened, but here's a nice response');
});
```

retryWhen equivalent:

```js
myObservable.catch((err, source, caught) => {
  if(totallyWannaRetry(err)) {
    return caught;
  } else {
    throw err;
  }
})
```

blesh:
[...]

PR #144, opened 2015-07-29
add catch operator, related to #141, closes #130

### LINKED ISSUE #130: Add catch

blesh:
Note this implementation adds a second argument of the source observable to catch. See unit tests for details.

### DIFF
--- a/spec/operators/catch-spec.js
+++ b/spec/operators/catch-spec.js
@@ -1,0 +1,109 @@
+/* globals describe, it, expect */
+var Rx = require('../../dist/cjs/Rx');
+var Observable = Rx.Observable;
+
+describe('Observable.prototype.catch()', function () {
+  it('should pass the error as the first argument', function (done) {
+    Observable.throw('bad')
+      .catch(function (err) {
+        expect(err).toBe('bad');
+        return Observable.empty();
+      })
+      .subscribe(function () { },
+        function (err) {
+          expect('this was called').not.toBeTruthy();
+        },
+        done);
+  });
+  
+  it('should catch the error and allow the return of a new observable to use', function (done) {
+    var expected = [1, 2, 'foo'];
+    Observable.of(1, 2, 3)
+      .map(function (n) {
+        if (n === 3) {
+          throw 'bad';
+        }
+        return n;
+      })
+      .catch(function (err) {
+        return Observable.of('foo');
+      })
+      .subscribe(function (x) {
+        expect(x).toBe(expected.shift());
+      }, function (err) {
+        expect('this was called').not.toBeTruthy();
+      }, function () {
+        done();
+      });
+  });
+  
+  it('should catch and allow the observable to be repeated with the third (caught) argument', function (done) {
+    var expected = [1, 2, 1, 2, 1, 2];
+    var retries = 0;
+    Observable.of(1, 2, 3)
+      .map(function (n) {
+        if (n === 3) {
+          throw 'bad';
+        }
+        return n;
+      })
+      .catch(function (err, caught) {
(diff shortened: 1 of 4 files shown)


blesh:
@trxcllnt ... please review?

trxcllnt:
@blesh 👍 overall, but two questions jump to mind:

1. the utility of the `source` Observable inside the catch selector is dubious imo
2. I'm wary of saving the Observable returned from the catch selector as the "last" selected Observable. I know they're not supposed to, but someone could return operations on the `caught` Observable, which would be remembered the next time the catch selector is invoked... but wouldn't have any insight into why the new `caught` Observable is dispatching different events than expected from the `source` Observable.

Do you have the same concerns? Do you have examples that can't be accomplished without scanning the `caught` Observable?

blesh:
> the utility of the source Observable inside the catch selector is dubious imo

I completely agree. We can probably remove that.

> I'm wary of saving the Observable returned from the catch selector as the "last" selected Observable. 

Yes, definitely, I actually changed that, it always pumps in the `caught` observable now. It no longer saves the "last" observable. However, there may be some interesting utility to pumping in the `last` observable. Mostly utility in the form of what you just mentioned. I'm not sure why someone would want to do that though... Again, it seems dubious...

How about this for the signature?

```js
catch(handler: (err: any, caught: Observable<any>) => Observable<any>)
```

if we wanted to do the full monty:

```js
catch(handler: (err: any, caught: Observable<any>, last: Observable<any>, source: Observable<T>) => Observable<any>)
```

My gut tells me the first option is the better one though.

trxcllnt:
@blesh Agreed, I like the first option better. seems like the second option would be better suited as a separate `catchExpand` operator or similar.

blesh:
@trxcllnt okay, I'm going to go with that first option then...

trxcllnt:
@blesh LGTM

src/operators/catch.ts
```diff
@@ -0,0 +1,49 @@
...
+  caught: Observable<any>;
+  
+  constructor(destination: Observer<T>, selector: (err:any, caught:Observable<any>) => Observable<any>, caught: Observable<any>) {
+    super(destination);
+    this.selector = selector;
+    this.caught = caught;
+  }
+
+  _error(err) {
+    const result = tryCatch(this.selector).call(this, err, this.caught);
```
trxcllnt: @blesh @mattpodwysocki pointed out we can do without the `.call(this, ` part of the selector invocation.

spec/operators/catch-spec.js
```diff
@@ -0,0 +1,104 @@
...
+          throw 'bad';
+        }
+        return n;
+      })
+      .catch(function (err) {
+        expect(err).toBe('bad');
+        return Observable.of('foo');
+      })
+      .subscribe(function (x) {
+        expect(x).toBe(expected[i++]);
```
jeffbcross: Or expected.shift() and get rid of the i.

spec/operators/catch-spec.js
```diff
@@ -0,0 +1,104 @@
...
+    var i = 0;
+    Observable.of(1, 2, 3)
+      .map(function (n) {
+        if (n === 3) {
+          throw 'bad';
+        }
+        return n;
+      })
+      .catch(function (err) {
+        expect(err).toBe('bad');
```
jeffbcross: This repeated expectation seems like it should be its own spec, and removed from other specs.

src/operators/catch.ts
blesh: I was just following prior art elsewhere in the lib really. I thought it was strange. We should just call this directly, no?

Issue #126, opened 2015-07-28
Micro performance tests of operators


ktrott:
For each operator, run through tests w/ a range of sizes of streams of data
* 1
* 10
* 100
* 1000
* 10000
* 100000
* 1000000

With and without a scheduler

Idea: Array of ranges, operators, schedulers and run it.

blesh:
closed by #146

PR #146, opened 2015-07-30
Adds micro-benchmark performance tests.

trxcllnt:
Addresses #126.


mattpodwysocki:
@trxcllnt please use 3.0 which is the most current straight from GitHub instead of 2.5 which will not give you accurate results.

trxcllnt:
@mattpodwysocki I'm testing against the latest version in npm because that's what the public uses. I'm happy to bump the version when 3.0 is published.

blesh:
@trxcllnt the actual `.spec.js` files need to be updated, they still are trying to go to `perf/flatmap` rather than `perf/macro/flatmap`

PR #147, opened 2015-07-30
Adding thisArg and bindCallback

mattpodwysocki:
This should optimize the `.call` attempts on `map` and `filter` so that if the `thisArg` is not supplied, there is no perf hit.


trxcllnt:
@mattpodwysocki 👍 this looks good to me, thanks for the PR. I've got few questions:

 - I've heard rumblings that `Function.prototype.bind` is about to be heavily optimized. Should we have to include `thisArg` in the API when users can bind `thisArg` themselves?
 - If we *do* include `thisArg` in `map` and `filter`, should we also include it in `flatMap`, etc. for consistency?

blesh:
Pulling it down to test locally now...

Issue #95, opened 2015-07-17
Re-implement fromEvent/fromEventPattern with tests


blesh:
We need to reimplement `fromEvent` and `fromEventPattern`, only this time with tests.

PR #154, opened 2015-08-04
feat(observable): add fromEventPattern creator function

blesh:
cc @trxcllnt

### DIFF
--- a/src/observables/FromEventPatternObservable.ts
+++ b/src/observables/FromEventPatternObservable.ts
@@ -1,0 +1,40 @@
+import Scheduler from '../Scheduler';
+import Observable from '../Observable';
+import Subscription from '../Subscription';
+import tryCatch from '../util/tryCatch';
+import {errorObject} from '../util/errorObject';
+
+export default class FromEventPatternObservable<T, R> extends Observable<T> {
+
+  static create<T>(addHandler: (handler: Function) => any, removeHandler: (handler: Function) => void, selector?: (...args:Array<any>) => T) {
+    return new FromEventPatternObservable(addHandler, removeHandler, selector);;
+  }
+  
+  constructor(private addHandler: (handler:Function) => any, private removeHandler: (handler:Function) => void, private selector?: (...args:Array<any>) => T) {
+    super();
+  }
+
+  _subscribe(subscriber) {
+    const addHandler = this.addHandler;
+    const removeHandler = this.removeHandler;
+    const selector = this.selector;
+    
+    const handler = selector ? function(e) {
+      let result = tryCatch(selector).apply(null, arguments);
+      if (result === errorObject) {
+        subscriber.error(result.e);
+      } else {
+        subscriber.next(result);
+      }
+    } : function(e) { subscriber.next(e); }
+    
+    let result = tryCatch(addHandler)(handler);
+    if (result === errorObject) {
+      subscriber.error(result.e);
+    }
+    subscriber.add(new Subscription(() => {
+      //TODO: determine whether or not to forward to error handler
+      removeHandler(handler)
+    }));
+  }
+}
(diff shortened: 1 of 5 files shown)


blesh:
This addresses #95 in part.

src/observables/FromEventPatternObservable.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  static create<T, R>(addHandler: (handler: Function) => void, removeHandler: (handler: Function) => void, selector?: (x: R) => T, scheduler?: Scheduler) {
+    return new FromEventPatternObservable(addHandler, removeHandler, selector, scheduler);;
+  }
+  
+  static dispatch(state) {
+    const { addHandler, removeHandler, selector, subscriber } = state;
+    let scheduler = (<any>this).scheduler;
+    
+    let handler = selector ? (e) => {
```
mattpodwysocki: You shouldn't have an e here, instead should have ...args to ensure that in case the event handler has more than just one argument

src/observables/FromEventPatternObservable.ts
```diff
@@ -0,0 +1,72 @@
+import Scheduler from '../Scheduler';
+import Observable from '../Observable';
+import Subscription from '../Subscription';
+import tryCatch from '../util/tryCatch';
+import {errorObject} from '../util/errorObject';
+
+export default class FromEventPatternObservable<T, R> extends Observable<T> {
+
+  static create<T, R>(addHandler: (handler: Function) => void, removeHandler: (handler: Function) => void, selector?: (x: R) => T, scheduler?: Scheduler) {
+    return new FromEventPatternObservable(addHandler, removeHandler, selector, scheduler);;
```
mattpodwysocki: the addHandler should not return void, as it currently works, it returns whatever is returned invoking the addHandler.  This is useful in places like Dojo which return a handle for unsubscription instead of a subscribe/unsubscribe pair

src/observables/FromEventPatternObservable.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+  
+  static dispatch(state) {
+    const { addHandler, removeHandler, selector, subscriber } = state;
+    let scheduler = (<any>this).scheduler;
+    
+    let handler = selector ? (e) => {
+      let value = tryCatch(selector)(e);
```
mattpodwysocki: instead of tryCatch on e, it should be applied to the arguments coming in as it is currently with RxJS

src/observables/FromEventPatternObservable.ts
mattpodwysocki: why is a scheduler required here unless to allow someone to change where the unsubscribe happens (which is where subscribeOn) was meant to address

src/observables/FromEventPatternObservable.ts
blesh: Fair... so, should I also check `args.length` here and return just the one argument if there is only one?

mattpodwysocki: You should assign e to arguments[0] and a selector if there are more args should work, like we have in our existing RxJS [`fromEventPattern`](https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/linq/observable/fromeventpattern.js#L11-L16)

blesh: So you're passing all arguments to the selector, but only one argument back through. I see. I guess I haven't used that feature, thank you for the clarification.

src/observables/FromEventPatternObservable.ts
blesh: Your right, a scheduler probably isn't necessary here, I guess my mind defaulted to that since it's an Observable creation method.

blesh:
As it seems I've met requests, and I need to move on, I've merged this PR.

PR #155, opened 2015-08-04
feat(observable) add fromEvent creation method

blesh:
Addresses #95 

cc/ @mattpodwysocki @trxcllnt


blesh:
As this was so closely related to `fromEventPattern`, and the advice @mattpodwysocki gave there applied here, I've applied that advice, and I'm merging so I can continue.

PR #160, opened 2015-08-06
feat(operator): add toPromise operator. closes #159

### LINKED ISSUE #159: add toPromise operator
We need to add at least the basic implementation of this. I haven't figured out the best strategy for configuration.

blesh:
I didn't add a configuration option yet, because I'm still unsure where we're going to put that. But for now, it's the basic implementation.

It *might* be worth adding a custom `toPromise` method on `ScalarObservable` for performance reasons, but I have my doubts that will be game-changing for anyone.


blesh:
Merging after a week with no objections.

PR #162, opened 2015-08-10
Implement startWith() operator

### LINKED ISSUE #84: Add startWith operator

saneyuki:
Fix #84

### DIFF
--- a/src/operators/startWith.ts
+++ b/src/operators/startWith.ts
@@ -1,0 +1,8 @@
+import Observable from '../Observable';
+import ScalarObservable from '../observables/ScalarObservable';
+import concat from './concat';
+
+export default function startWith<T>(x: T): Observable<T> {
+  const init = new ScalarObservable(x);
+  return concat.call(init, null, <Observable<T>>this);
+}
--- a/spec/operators/startWith-spec.js
+++ b/spec/operators/startWith-spec.js
@@ -1,0 +1,20 @@
+/* globals describe, it, expect */
+var Rx = require('../../dist/cjs/Rx');
+var Observable = Rx.Observable;
+
+describe('Observable.prototype.startWith()', function () {
+  it('should start an obsevable with given value', function (done) {
+    var source = 'source'
+    var init = 'init';
+    var expected = [init, source];
+
+    var i = 0;
+    Observable.just(source)
+    .startWith(init)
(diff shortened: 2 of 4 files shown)


blesh:
Thanks! I'll check this out after a while unless someone else gets to it. I'm on vacation this week.

src/operators/startWith.ts
```diff
@@ -0,0 +1,7 @@
+import Observable from '../Observable';
+import ScalarObservable from '../observables/ScalarObservable';
+
+export default function startWith<T>(x: T): Observable<T> {
+  const init = ScalarObservable.just(x);
+  return init.concat(null, <Observable<T>>this);
```
trxcllnt: @saneyuki thanks for the PR! In the interest of modularity, would you mind updating this PR to pull in the module for the operator you're relying on here, like [`concat`](https://github.com/ReactiveX/RxJS/blob/master/src/operators/concat.ts#L1) does? Since we're aiming for complete modularity here, we can't assume `concat` will exist on the `ScalarObservable` prototype. I'd be satisfied with this PR if you pulled in either of the concat or merge modules, it's your choice. Thanks!

trxcllnt: It'd also be great if you changed this to use the `ScalarObservable` constructor, instead of `just`, since `ScalarObservable` isn't guaranteed to have `just` statically defined either.

saneyuki:
@trxcllnt thank you for your review!. I pushed the fixup commit.

trxcllnt:
@saneyuki thanks! lgtm.

blesh:
There's no way you could know this, because we don't have contribution guidelines, but I'd really like it if you could flatten your commits and  change the commit message to: `feat(operator): implement startWith"

The easiest way to do this would be an interactive rebase. `git rebase -i HEAD~2` then you'd fixup the second entry, and reword the first. If you need any additional help, let me know and I'm happy to assist.

Otherwise, LGTM... let me know when you've got those changes pushed and I'll merge it in.

Thanks again!

saneyuki:
@blesh 

okay! I squashed my changesets & change the message to `feat(operator): implement startWith().`

blesh:
PR merged with commit 1f36d99862c0e226cdb41b15eee8b40fd8d93b53

PR #166, opened 2015-08-13
feat(operator): add groupBy

### LINKED ISSUE #165: groupBy / groupByUntil
After speaking with @mattpodwysocki and @trxcllnt it's probably best to merge the functionality of grouipBy and groupByUntil into one operator.

The signature would essentially match `groupByUntil`, but the second two arguments would be completely optional.

```js

groupBy(keySelector:(value:T) => K, elementSelector?: (value:T) => R, durationSelector?: (group:Observable<R>) => Observable<any>): Observable<GroupedObservable<K, R>>

```

blesh:
closes #165

### DIFF
--- a/src/operators/groupBy.ts
+++ b/src/operators/groupBy.ts
@@ -1,0 +1,139 @@
+import Operator from '../Operator';
+import Observer from '../Observer';
+import Subscriber from '../Subscriber';
+import Observable from '../Observable';
+import Subject from '../Subject';
+import Map from '../util/Map';
+import FastMap from '../util/FastMap';
+
+import tryCatch from '../util/tryCatch';
+import {errorObject} from '../util/errorObject';
+import bindCallback from '../util/bindCallback';
+
+export default function groupBy<T, R>(keySelector: (value: T) => string,
+    elementSelector?: (value: T) => R,
+    durationSelector?: (grouped: GroupSubject<R>) => Observable<any>): Observable<GroupSubject<R>>
+{
+  return this.lift(new GroupByOperator<T, R>(keySelector, durationSelector, elementSelector));
+}
+
+export class GroupByOperator<T, R> extends Operator<T, R> {
+  constructor(private keySelector: (value: T) => string, 
+    private durationSelector?: (grouped: GroupSubject<R>) => Observable<any>,
+    private elementSelector?: (value: T) => R) {
+    super();
+  }
+  
+  call(observer: Observer<R>): Observer<T> {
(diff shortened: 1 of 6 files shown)


src/operators/groupBy.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+export class GroupDurationSubscriber<T> extends Subscriber<T> {
+  constructor(private group: GroupSubject<T, any>, private parent:GroupBySubscriber<any, T, any>) {
+    super(null);
+  }
+  
+  _next(value: T) {
+    this.group.complete();
+    this.parent.groups[this.group.key] = null;
```
trxcllnt: Nitpicking here, but this logic seems ripe to put into a function on the GroupBySubscriber, the way we've done it in [merge](https://github.com/ReactiveX/RxJS/blob/master/src/operators/merge.ts#L106). This would also make it easier to subclass.

blesh: Sounds reasonable

blesh:
There is a glaring bug in this one... if you select an object as a key. I'll need to rework this to add some sort of hash code or use an Array and lookup groups each time in the event that keys are objects.

blesh:
Made the desired changed for @trxcllnt ...

I'm not going to change the "bug" I mentioned above though, keys should be returned as strings or they will coerce as strings because it's using a hashtable under the hood for group storage. I'm sticking with this method for performance reasons. Native Map implementations and Map shims are much, much slower on the look up. Given that performance is a primary goal, and that creating a composite key in the  `keySelector` should be trivial, I see no reason to slow things down to accommodate using "anything" as a key.

blesh:
### Compromise 

After speaking with both @trxcllnt and @mattpodwysocki on this... I've come up with what I think is a solid compromise...

- the underlying `groups` collection on the `GroupBySubscriber` starts off as `null`
- when the first `key` arrives, the type is examined (just once) to determine if it's a string or not.
  - if the key is a string: Use the `FastMap`, which is a Map-duck-quack-quack that simply wraps a Hash table.
  - if the key is *not* a string: Use native `Map` or a stub.

My tests have determined that `FastMap` is twice as fast for getting and setting values than `Map`.

### Caveats

- If the first key is a string, but other keys are Objects, you're going to have a bad time.

### Future

I think this approach will enable us to easily remove the "FastMap" whenever ES6 Map is better optimized by more platforms.

blesh:
merged via rebase with 1e13aeab34c65732eb243662377ccca4c3f904c1

PR #169, opened 2015-08-14
feat(operator): add ensure

trxcllnt:


blesh:
I'm unfamiliar with the `ensure` operator. Isn't this `finally`?

staltz:
@blesh yup

```js
observableProto.ensure = ensure;
observableProto.finally = ensure;
```

blesh:
Weird, that's undocumented in the current version. 

@trxcllnt can we rename the file and commit to `finally`? I don't really care if there's an alias or not, but I'd bet most people, including myself, would look for a `finally.ts` if they were trying to locate the code behind this operator. I'd wager that most people only use `finally`.

Otherwise this looks good to me.

mattpodwysocki:
@blesh I'm with you as `finally` makes more sense here when it comes to something happening as we already have `catch`, why not `finally`?  We're not like Ruby with `rescue` and `ensure`.

trxcllnt:
closed in 526e4c92f0a2b1670aebf7cde39063657f47b684.

mattpodwysocki:
@blesh @trxcllnt and since it was undocumented, it has been subsequently removed from RxJS v3.0.1 because I wanted to limit aliases in RxJS v3.0, so I would expect us to remove `ensure` as well here.  I had forgotten those aliases existed in 2.x so they are now gone

blesh:
Well, this PR has been merged with [this commit](https://github.com/ReactiveX/RxJS/commit/526e4c92f0a2b1670aebf7cde39063657f47b684), so I'll keep it closed... so we'll need another `chore(operator): remove ensure alias` PR

Issue #174, opened 2015-08-14
chore(operator) : remove ensure alias for finally


mattpodwysocki:
This will allow us to remove the `ensure` alias from `finally` as it has already been removed from RxJS v3.0.1

blesh:
closed by 3f743ab12476f0b689c79a68637dff2f1cd1d475

PR #176, opened 2015-08-14
Removing ensure alias for finally

mattpodwysocki:
Fixes Issue #174 which removes the ensure alias for finally


blesh:
Ah, crap... should have asked you to change the commit message to `chore(operator): removing ensure` ... I guess it doesn't matter in this case.

(I'd like to use the Changelog generation tooling that the @jeffbcross and friends have)

mattpodwysocki:
@blesh so noted in the future, thanks

PR #181, opened 2015-08-17
feat(operator): add skipUntil

blesh:


blesh:
oops... looks like I conflated two PR branches here...

PR #185, opened 2015-08-18
feat(operator): Add do operator.

trxcllnt:


mattpodwysocki:
Do we not want Observer support here as well in addition to the three main functions?

trxcllnt:
@mattpodwysocki we do, but I need this now for my work on Falcor right now. Should also probably be `doObserver` or something similar, as we've decided against polymorphic functions for performance reasons.

mattpodwysocki:
@trxcllnt then are we going to support `doOnNext` etc to handle `onNext` calls, and for all of the other verbiage like we have in the current as well?

trxcllnt:
@mattpodwysocki I don't see why not.

PR #189, opened 2015-08-20
chore(operator): fix API signatures for merge, concat, combineLatest …

### LINKED ISSUE #184: Proposal: Argument order in concat, merge, zip, concatLatest should be changed
Currently the arguments for both `concat` and `merge` having optional arguments *first* with splat params for Observables after:

### Current

```js
merge(scheduler?:Scheduler, concurrent?:number, ...observables:Array<Observable<T>>)
concat(scheduler?:Scheduler, ...observables:Array<Observable<T>>)
```

### Proposed

I'd like to change the order and remove the splat params:

```js
merge(observables: Array<Observable<T>>, concurrent?:number, scheduler?:Scheduler)
concat(observables: Array<Observable<T>>, scheduler?:Scheduler)
```

Reasons are as follows:

- any function that has an optional Scheduler argument should have that argument be the last argument, for consistency.
- Passing an array rather than splat params means we no longer have to analyze  the `arguments` object and do any `slice` trickery to get the array of Observables out of it, and should optimize better.
- Having all optional arguments come last means we no longer have to do any `typeof` assertions to figure out what argument is what.

(TLDR: less polymorphic functions)

blesh:
…and zip

closes #184

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "benchmark": "^1.0.0",
     "benchpress": "^2.0.0-alpha.25",
     "browserify": "^11.0.0",
+    "colors": "^1.1.2",
     "glob": "^5.0.10",
     "http-server": "^0.8.0",
     "jasmine": "^2.3.1",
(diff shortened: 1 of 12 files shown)


blesh:
oops, accidentally left some micro perf tests commented out...

PR #190, opened 2015-08-20
Change CJS and AMD build steps to use typescript compiler

jeffbcross:


jeffbcross:
I'm looking into the Travis failure. Tests were passing at some point, but I made a few changes.

jeffbcross:
Not sure why the takeUntil test is failing, though I've been experiencing some flakes with other tests like skipUntil when testing locally, even on master branch without these changes.

trxcllnt:
@jeffbcross it's a difference in timing between your local env and the CI server. this will be cleared up when I do the virtual time stuff. this PR lgtm

blesh:
> this will be cleared up when I do the virtual time stuff

Yup, that's going to be a big deal .... I definitely need it to refactor the tests for `throttle` and `debounce`

blesh:
FWIW... the test in question can be fixed (for now) by adding larger delays to the tests.

trxcllnt:
closed by abe9a2418af67e09df7d0085f5c7bb63edd6bff2.

PR #204, opened 2015-08-24
proposal: Change `Operator` from class to interface.

saneyuki:
The motivation of this pull request is here:

A deriving class is unclear to understand "what we should need to implement this derived classs?". Thus, I think we should define interface and implement it for each implementations in these case:

- There is no default behavior, or a few default behavior.
- No need to create a base class instance.

`Operator` base class define only `call()` method, I think it will not be high cost for us to implement `call` method for each `Operator` implementations.

How do you think about this?

### DIFF
--- a/src/Operator.ts
+++ b/src/Operator.ts
@@ -1,8 +1,12 @@
 import Observer from './Observer';
 import Subscriber from './Subscriber';
 
-export default class Operator<T, R> {
-  call<T, R>(observer: Observer<R>): Observer<T> {
-    return new Subscriber<T>(observer);
-  }
+interface Operator<T, R> {
+  call<T, R>(observer: Observer<R>): Observer<T>;
+}
+
+export default Operator;
+
+export function defaultCallFn<T, R>(observer: Observer<R>): Observer<T> {
+  return new Subscriber<T>(observer);
 }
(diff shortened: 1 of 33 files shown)


blesh:
@saneyuki I don't at all disagree with this PR. If we use an interface, TypeScript will then complain if we don't implement the `call` method. And I don't think we're using the plain `Operator` type anywhere. SO this makes total sense.

I'll leave this PR sit for a while for others to chime in, but for now it looks fine to me.

blesh:
attn @trxcllnt

trxcllnt:
@saneyuki @blesh the only reason this is a base class is because TypeScript creates an empty file for interfaces. Do we expect anybody to ever import the `defaultCallFn` from Operator?

blesh:
> the only reason this is a base class is because TypeScript creates an empty file for interfaces. 

Doesn't seem like a big deal. Don't think of it as an "empty file"... think of it as "importing all of JavaScript in 0 bytes".

> Do we expect anybody to ever import the defaultCallFn from Operator?

I doubt it.


Otherwise LGTM. @trxcllnt ?

trxcllnt:
@blesh lgtm

blesh:
Thank you @saneyuki!

Merged with d565115d5c64c740dd5ce93060c816d0ad1daee1

saneyuki:
@blesh @trxcllnt thank you for your reviewing!

PR #210, opened 2015-08-26
feat(operator): add retry

jeremymwells:


blesh:
Thanks @jeremymwells! Other than my (generally nitpicky) comments, it looks good. If you get those few small issues cleaned up, I'll get this merged right away.

Issue #100, opened 2015-07-17
Roadmap: Alpha Release


blesh:
Below is a list of what is required for an Alpha release.

### Necessary Basic Static Methods
I'm cherry-picking from current RxJS2

- [x] concat
- [x] empty
- [x] from
- [x] fromArray
- [x] fromEvent
- [x] fromEventPattern
- [x] fromPromise
- [x] interval
- [x] just/value
- [x] merge
- [x] never
- [x] of
- [x] range
- [x] return
- [x] throw
- [x] timer
- [x] zip

### Necessary Basic Operators 

I'm cherry-picking [operators from RxJS 2](https://github.com/Reactive-Extensions/RxJS/tree/master/doc/api/core/operators) for now. I'm trying to narrow it down to what's essential to implement.

- [x] buffer
- [x] bufferCount (bufferWithCount)
- [x] bufferTime (bufferWithTime)
- [x] bufferToggle (buffer(obs, fn))
- [x] bufferWhen (buffer(fn))
- [x] catch
- [x] combineLatest
- [x] concat
- [x] concatAll
- [x] defaultIfEmpty
- [x] delay
- [x] flatMap
- [x] filter
- [x] forkJoin
- [x] groupBy
- [x] groupByUntil
- [x] map
- [x] mapTo
- [x] merge
- [x] mergeAll
- [x] multicast
- [x] observeOn
- [x] publish
- [x] reduce
- [x] repeat
- [x] retry
- [x] retryWhen
- [x] scan
- [x] skip
- [x] skipUntil
- [x] startWith
- [x] subscribeOn
- [x] switchMap (flatMapLatest)
- [x] take
- [x] takeUntil
- [x] timeout
- [x] toArray
- [x] toPromise
- [x] window
- [x] windowCount (windowWithCount)
- [x] windowTime (windowWithTime)
- [x] windowToggle (window(obs, fn))
- [x] windowWhen (window(fn))
- [x] withLatestFrom
- [x] zip
- [x] zipAll

### Necessary Support Types

(Not including Observable and Subscriber, which are a given)

- [x] BehaviorSubject
- [x] ConnectableObservable (multicast)
  - [x] refCount
- [x] Subject
- [x] ReplaySubject

### Schedulers

- [x] immediate
- [x] nextTick

### Macro Performance Tests

[...]

PR #211, opened 2015-08-26
Add micro-benchmark performance test for skip operator.

kwonoj:
Adding additional micro test coverage for skip operator listed in issue https://github.com/ReactiveX/RxJS/issues/100 by following existing one.


blesh:
Awesome thank you!

PR #216, opened 2015-08-27
bugfix(zip): Remove usage of sparse array

jinroh:
ES5 `every` has a weird behavior on sparse arrays which broke `zip` operator.

```js
var foo = []; foo[1] = "a";
foo.every(hasValue); // === true;
```


blesh:
Oh wow, good catch @jinroh. Yeah, I think `every` skips any item in an array that is `undefined`... 

merged as of 8c20e82185ff36802f3a35b0a2020277bf550ecb

PR #218, opened 2015-08-27
bugfix(take): Complete on limit reached

jinroh:
`take` should complete when the limit is reached, not waiting for another event and then complete.


blesh:
I've fixed the broken test and merged this as of 801a711f2f1e4a1c4e327bc3b63603e26042926a

blesh:
Thank you, @jinroh!

Issue #219, opened 2015-08-27
feat(operator): implement publishBehavior


staltz:
For symmetry with `publish` which multicasts on a Subject, we should have `publishBehavior` and `publishReplay`. In Reactive-Extension/RxJS these exist but are `publishValue` and `replay` respectively. The new naming would give a nice symmetry:

- `publish` multicasts on `Subject`
- `publishBehavior` multicasts on `BehaviorSubject`
- `publishReplay` multicasts on `ReplaySubject`

I can implement this, if you think this plan looks good.

PR #223, opened 2015-08-27
feat(operator): add publishBehavior operator and spec

staltz:
Implements #219.
We're not sure yet whether #219 should be done or not, but in any case, here's the PR in the pipeline awaiting for review/approval.


trxcllnt:
@staltz lgtm

blesh:
merged with 249ab8d

Thank you, @staltz!

PR #225, opened 2015-08-27
refactor(Subscriber): make Observer an interface, inherit from Subscr…

### LINKED ISSUE #224: refactor(Subscriber): Observer should just be an interface
I think we could change `Subscriber` to be a class that `extends Subscription` and `implements Observer` rather than the other way around. Then `Observer` could just be an interface, and we can clean up some logic most likely.

The reason I suggest this is we never actually `new` up an Observer anywhere... whereas `Subscription` does get `new`-ed up and reused frequently. So it doesn't make as much sense to have `Subscription` be the one that is being used like an interface on `Subscriber`.

cc/ @trxcllnt

blesh:
…iption

Changed Observer to an interface throughout the app since we were not using it directly as a class anywhere
Fixed macro perf tests that were calling  rather than  for some reason

closes #224

### DIFF
--- a/perf/macro/flatMap-scalar/perf.js
+++ b/perf/macro/flatMap-scalar/perf.js
@@ -18,7 +18,7 @@ var RxNextTestObservable = new RxNext.Observable(function(observer) {
   while(++index < numIterations) {
     observer.next(index);
   }
-  observer.completed();
+  observer.complete();
 });
 
 var Rx2TestObservable = Rx.Observable.create(function(observer) {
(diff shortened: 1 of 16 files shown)


blesh:
attn @trxcllnt - please review this change, since it was a lot of refactoring.  I was as careful as I could be, but I may have missed something.

attn @jeffbcross - The macro performance tests are really fast after this change. Can you double check that it's actually doing something? I manually ran the perf test and saw that it *was* doing something, but I want to make sure benchpress wasn't recording false values.

trxcllnt:
@blesh L**GREAT**TM

blesh:
nvm @jeffbcross I guess it was already that fast. We may have to up the iterations of the macro perf tests to make them register something a little higher. lol

PR #227, opened 2015-08-28
bugfix(combinelatest): No upper bound limit

jinroh:
The `combineLatest` operator was basically operating like `ignoreElements`, and the tests did not test anything except for the completion of the returned observable. This PR should fix the issue.


jinroh:
Note that the `combineLatest` operator is also using sparse arrays -- like `zip` was before #216. This PR does not change that as it would require some re-factoring.

blesh:
Thanks @jinroh ... merged as of 81e5dfb

PR #228, opened 2015-08-28
feat(observable): add Observable.forkJoin

blesh:


mattpodwysocki:
@blesh not a fan of this naming here.  Also, the forkJoin should also take a selector function at the end as the current implementation does.

blesh:
@mattpodwysocki ... I don't have a problem with what you're saying. Changing it to `all` was a request from multiple people. Perhaps it's worth just aliasing? [FWIW, the selector function isn't documented in current Rx](https://github.com/Reactive-Extensions/RxJS/blob/de3d945cf9cea8e78d61486e9220f10846ce3463/doc/api/core/operators/forkjoin.md)

blesh:
@mattpodwysocki ... also I'm not seeing the selector function you're talking about in the current implementation: https://github.com/Reactive-Extensions/RxJS/blob/de3d945cf9cea8e78d61486e9220f10846ce3463/src/core/linq/observable/forkjoin.js

blesh:
@jayphelps suggested `joinLast` as another alternative. `forkJoin` just doesn't make sense to me, personally. But if I'm alone in that sentiment, then I'm happy to leave it `forkJoin`.

blesh:
Changed `all` back to `forkJoin` since there were no strong opinions about it.

PR #229, opened 2015-08-28
chore(test): add micro-benchmark performance test for defaultIfEmpty …

kwonoj:
…operator.

Continue effort to cover items in issue https://github.com/ReactiveX/RxJS/issues/100


blesh:
Thanks again, @kwonoj!

Issue #220, opened 2015-08-27
feat(operator): implement publishReplay


staltz:
For symmetry with `publish` which multicasts on a Subject, we should have `publishBehavior` and `publishReplay`. In Reactive-Extension/RxJS these exist but are `publishValue` and `replay` respectively. The new naming would give a nice symmetry:

- `publish` multicasts on `Subject`
- `publishBehavior` multicasts on `BehaviorSubject`
- `publishReplay` multicasts on `ReplaySubject`

I can implement this, if you think this plan looks good.

blesh:
I *personally* don't have a problem with this other than API bloat. I mean, is it not just sugar for `observable.multicast(() => new BehaviorSubject(init))`? 

@mattpodwysocki @trxcllnt @headinthebox... have you seen broad needs for such operators?

staltz:
It indeed is sugar for `observable.multicast(() => new BehaviorSubject(init))`.
But `publish` is sugar for `observable.multicast(() => new Subject())` yet publish is already in the library.

staltz:
As some historical context: https://github.com/Reactive-Extensions/RxJS/commit/4e120cba62cbf716fe2131464b8e163fb4e8f9d9

blesh:
FWIW: I've always felt that really broad APIs make users feel uneasy when trying to learn.

headinthebox:
I'm not the right person to ask; I only use `xs.publish(_xs => ...)` never any of the other forms.

blesh:
> I'm not the right person to ask; I only use xs.publish(_xs => ...) never any of the other forms.

That's interesting, doesn't that one automatically subscribe? I've always found that one in particular to be an odd implementation.

I actually like @trxcllnt's idea of an operator that allows one to choose the connection and disconnection strategy of the hot observable. I think it was something like:

```js
var publishedObs = original.multicast(() => new Subject(), (original, subj) => {
   // connection strategy here
   var subscription = original.subscribe(subj);

   return () => {
      // disconnection strategy here
   }
});
```

trxcllnt:
@blesh I believe @headinthebox is referring to RxJava's implementation of publishWithSelector (which is different from current RxJS).

PR #230, opened 2015-08-29
feat(operator): add publishReplay operator and spec

staltz:
Implements #220. Similar to #223


blesh:
merged as of a0c47d6

Thank you @staltz!

PR #231, opened 2015-08-29
chore(test): add micro-benchmark performance test for repeat operator

kwonoj:
Continue effort to cover items in issue #100


blesh:
merged as of f64b81b

Thank you again, @kwonoj!

PR #234, opened 2015-08-31
refactor(of): of(1) returns a ScalarObservable, just, value, return r…

### LINKED ISSUE #232: `just`, `value`, and `return` vs `of`
In the name of reducing the API surface area. I'd like to remove `just`, `value`, and `return` (I think we already removed `just`) in favor of using `of`. 

In order to make of more comparable to `value` or `return`, we could return a `ScalarObservable` if `of` is just one value.

blesh:
…emoved

- updates of to return a ScalarObservable or EmptyObservable where appropriate
- removes just, value and return to reduce API surface
- updates tests that utilized the above methods

closes #232


trxcllnt:
lgtm

PR #235, opened 2015-08-31
Add more tests for publishBehavior and publishReplay

staltz:
**DO NOT MERGE THIS PR** Because tests fail, and for now it is just to spark a discussion about the semantic of these operators.

Refer to Lee Campbell's book:

http://introtorx.com/Content/v1.0.10621.0/02_KeyTypes.html#ReplaySubject
> Another difference is that a replay-one-subject will still cache its value once it has been completed. So subscribing to a completed BehaviorSubject<T> we can be sure to not receive any values, but with a ReplaySubject<T> it is possible. 

Yet tests fail like this:
```
Failures:
1) Observable.prototype.publishBehavior() should not emit next events to observer after completed
  Message:
    Expected [ 0 ] to equal [  ].
  Stack:
    Error: Expected [ 0 ] to equal [  ].
      at Object.<anonymous> (./RxJSNext/spec/operators/publishBehavior-spec.js:65:21)
2) Observable.prototype.publishReplay() should emit replayed events to observer after completed
  Message:
    Expected [  ] to equal [ 3, 4 ].
  Stack:
    Error: Expected [  ] to equal [ 3, 4 ].
      at Object.<anonymous> (./RxJSNext/spec/operators/publishReplay-spec.js:98:21)
145 specs, 2 failures
Finished in 12.951 seconds
```

I tried to look for the root cause, I digged into multicast, then into ConnectableObservable, but I couldn't find any apparent cause for this behavior. Can anyone explain why these tests fail, or if their behavior is actually correct (and tests are wrong, meaning a mismatch with the semantics of their corresponding subjects after `onCompleted`)?


blesh:
cc/ @jhusain  ... I think we were talking about your implementations of `share` and `shareReplay`(??) and this is related.

blesh:
FYI: when #281 lands, we'll have all the of the infrastructure in place for building "marble" tests.

blesh:
@staltz just so you know, the marble tests have landed, and there are examples in the `combineLatest-spec`. You may want to try that.

staltz:
Thanks, I'll try if I find some time. Not related to this issue, however. It's still open to all of us (it seems) why does it behave like it does now.

blesh:
if I had to guess, I'd say it's because we fixed multicast to always take a *factory* for a Subject, rather than a subject. This was for a few reasons. Primarily so that: 1. People didn't use the same subject instance for twenty multicasts and some other use on the side, which would be a really weird antipattern. and 2. So that "shared" or "refCounted" observables behaved appropriately, in that they are cold until subscribed to once, then hot, and when refCount drops to zero, they go cold again. 

If I had to guess as to way publishBehavior works differently now than what you're expecting, that would be my first suspicion.

blesh:
@staltz ... I'm not sure what to do with this PR. I'm going to close it for now. Perhaps you can open an issue to discuss concerns with `publishBehavior` and `publishReplay`?

staltz:
Yeah, close the PR and lets discuss it in an issue.

blesh:
:+1:

PR #242, opened 2015-08-31
chore(test): add micro-benchmark performance test for Observable.just

kwonoj:
Not sure how does alias being treated per coverage i.e, 'just' and 'return', one test considered covers all alias or not - any suggestions?


blesh:
I actually *just* removed `just`, `value` and `return` in favor of `of`. The first three were identical aliases, the latter can do the same thing

kwonoj:
Make sense, then this PR can be closed without merging - closing by myself.

PR #243, opened 2015-09-01
chore(test): update micro-benchmark performance test for repeat operator

kwonoj:
minor alignment for micro-perf test cases followed by change https://github.com/ReactiveX/RxJS/pull/234


blesh:
Oops... good catch, thanks @kwonoj

Issue #247, opened 2015-09-01
Get `observeOn` to utilize Notification class 


blesh:
Currently, `observeOn` is using it's [own Notification class](https://github.com/ReactiveX/RxJS/blob/03853ac4a510a6cede2ef294f56eaa503e5479d6/src/operators/observeOn.ts#L61) 

It might be beneficial to have it use the shared [Notification class](https://github.com/ReactiveX/RxJS/blob/03853ac4a510a6cede2ef294f56eaa503e5479d6/src/Notification.ts)

PR #249, opened 2015-09-01
refactor(observeOn): let observeOn uses notification instead of custo…

kwonoj:
…m implementation

Trying to make changes as suggested in https://github.com/ReactiveX/RxJS/issues/247

* Since this is first time to try to make in source tree, not sure if changes are correctly following recommended pattern. Used delay operator's implementation as reference, would be appreciate to let me know if it's not feasible way to do.

* Not sure about testcases, does current codebase has testscheduler in place? Wasn't able to find it so far.

### DIFF
--- a/src/operators/observeOn.ts
+++ b/src/operators/observeOn.ts
@@ -3,6 +3,7 @@ import Observer from '../Observer';
 import Scheduler from '../Scheduler';
 import Observable from '../Observable';
 import Subscriber from '../Subscriber';
+import Notification from '../Notification';
 
 export default function observeOn<T>(scheduler: Scheduler, delay: number = 0): Observable<T> {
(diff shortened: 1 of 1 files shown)


blesh:
>  does current codebase has testscheduler in place?

Not yet, there is a virtual scheduler in the works, hopefully landing soon.

blesh:
Otherwise this looks good to me.

kwonoj:
Thanks. in that case would it be acceptable PR's merged then test case comes with separate PR once scheduler is in place? or PR should be updated with test?

blesh:
@kwonoj yes, we're going to proceed with development, but we'll need to circle back and add a LOT of tests with the VirtualScheduler

blesh:
Thanks @kwonoj

Issue #239, opened 2015-08-31
Remove `fromArray`, `fromPromise` in favor of just `from`


blesh:
Now that `from` does exactly the same thing as `fromArray` or `fromPromise` if an array or a promise is passed to it, perhaps we can remove those two methods to reduce the API surface area.

This is similar to the removal of `just`, `value` and `return` in favor of just using `of`.

trxcllnt:
@blesh I am in favor of this

blesh:
After @kwonoj generously made a PR around this, and some experimentation with performance, `from(array)` has proven to be much slower than `fromArray(array)`. So much so that I'm going to close this issue.

Thanks @kwonoj, if not for your PR, we wouldn't have known.

kwonoj:
It is glad I could find out this :) Will continue effort on other topics. Thanks!

PR #251, opened 2015-09-02
refactor(from): fromArray, fromPromise removed

kwonoj:
- removes fromArray and fromPromise to reduce API surface
- updates tests that utilized the above methods

covers issue https://github.com/ReactiveX/RxJS/issues/239 .

### DIFF
--- a/perf/micro/immediate-scheduler/observable/from-array.js
+++ b/perf/micro/immediate-scheduler/observable/from-array.js
@@ -8,7 +8,7 @@ module.exports = function fromArray(suite) {
     }
 
     var oldFromArrayWithImmediateScheduler = RxOld.Observable.fromArray(args, RxOld.Scheduler.immediate);
-    var newFromArrayWithImmediateScheduler = RxNew.Observable.fromArray(args);
+    var newFromArrayWithImmediateScheduler = RxNew.Observable.from(args);
 
     // add tests
     return suite
--- a/perf/micro/current-thread-scheduler/observable/from-array.js
+++ b/perf/micro/current-thread-scheduler/observable/from-array.js
@@ -8,7 +8,7 @@ module.exports = function fromArray(suite) {
     }
 
     var oldFromArrayWithCurrentThreadScheduler = RxOld.Observable.fromArray(args, RxOld.Scheduler.currentThread);
-    var newFromArrayWithCurrentThreadScheduler = RxNew.Observable.fromArray(args, RxNew.Scheduler.immediate);
+    var newFromArrayWithCurrentThreadScheduler = RxNew.Observable.from(args, RxNew.Scheduler.immediate);
 
     // add tests
     return suite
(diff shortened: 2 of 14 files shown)


perf/micro/immediate-scheduler/observable/from-array.js
```diff
@@ -8,7 +8,7 @@ module.exports = function fromArray(suite) {
     }
 
     var oldFromArrayWithImmediateScheduler = RxOld.Observable.fromArray(args, RxOld.Scheduler.immediate);
-    var newFromArrayWithImmediateScheduler = RxNew.Observable.fromArray(args);
+    var newFromArrayWithImmediateScheduler = RxNew.Observable.from(args);
```
kwonoj: Found this specific test case is now slower than old Rx's fromArray(). Same test for current thread scheduler still performs as expected. (New implementation is faster)

blesh: Interesting. I'll have a look, perhaps it's something we can optimize further.

kwonoj: More interestingly this morning gave another try on different machine, results was different saying new implementation is faster. Not sure if slower result came out by some of my system configuration. Anyway, will try couple of more run later.

blesh: I ran this a few times, and didn't at all see what you're seeing. RxJS Next was still faster, with no noticeable difference in speed from what's in master. Can you run again to confirm? I have serious doubts that making this function polymorphic and one assertion of `isArray` would make such a large difference.

I'm going to go ahead and merge for now. It's easy enough to add `fromArray` back in if we need to.

kwonoj: Will do, and create separate issue if I could find noticeable repro on my end.

blesh: Ah, I see, it's still faster than Rx2, but it's half the speed of just using fromArray. Amazing that one `isArray` check would make such a difference.

kwonoj: Thanks for clarification, it's obviously interesting. What would be best plan in this case?

blesh: Hmm... it looks like perhaps we should leave the other methods in the library, but guide people to use `from` unless they know for a fact their variable is an `Array`.

Again, I'm pretty shocked at the difference a single `Array.isArray` check made here. I've tried a variety of different methods, but I think we're going to have to walk away from this issue/PR

blesh:
@kwonoj ... Thank you SO much for your effort on this, but after experimentation, it appears that the performance drop between `from(array)` and `fromArray(array)` makes this PR a no-go.

I'll have to close this, but we wouldn't have known this was the case if it wasn't for your efforts. As such, this was a very worthwhile PR.

perf/micro/immediate-scheduler/observable/from-array.js
kwonoj: It seems makes sense. I'm also quite curious about why Array.isArray is taking those noticeable amount of time, still not sure though.

kwonoj:
About to close PR on my end ;) Thanks.

PR #252, opened 2015-09-02
chore(test): add micro-benchmark performance test for Observable.throw

kwonoj:
Continue to cover https://github.com/ReactiveX/RxJS/issues/100.

### DIFF
--- a/perf/micro/current-thread-scheduler/observable/throw.js
+++ b/perf/micro/current-thread-scheduler/observable/throw.js
@@ -1,0 +1,20 @@
+var RxOld = require("rx");
+var RxNew = require("../../../../index");
+
+module.exports = function _throw(suite) {
+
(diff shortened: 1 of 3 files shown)


perf/micro/current-thread-scheduler/observable/throw.js
```diff
@@ -0,0 +1,20 @@
+var RxOld = require("rx");
+var RxNew = require("../../../../index");
+
+module.exports = function _throw(suite) {
```
kwonoj: '_' was chosen to avoid collision to language reserved keywords, would be appreciate to let me know if this is acceptable styles. Otherwise, will change it accordingly.

blesh: Perfect, that's what we were doing elsewhere, I believe.

blesh:
merged with 492f594

Thanks @kwonoj!

PR #254, opened 2015-09-03
chore(test): add micro-benchmark performance test for startwith op…

kwonoj:
…erator

Continue to cover https://github.com/ReactiveX/RxJS/issues/100.


blesh:
merged with 2028a61

Thanks again @kwonoj!

kwonoj:
My pleasure! :)

PR #256, opened 2015-09-03
feat(Observable): add static create method

### LINKED ISSUE #255: feat(Observable): add static Observable.create function
This was an oversight and should be a simple add.

blesh:
- adds more tests around Observable
- adds some code to assure isUnsubscribed is always boolean in Subscriber

closes #255


blesh:
This was a small and necessary change, so I'm merging it immediately.

PR #267, opened 2015-09-04
refactor(Scheduler): change order of arguments to schedule method

### LINKED ISSUE #265: Change scheduler `schedule` signature
Funny enough, I feel like this is the first way I did scheduling, but @trxcllnt is resurrecting it...

Instead of:

```TypeScript
schedule(delay: number, state:any, work:(state?: any) => Subscription | void) : Subscription
```

it should be:

```TypeScript
schedule(work:(state?: any) => Subscription | void, delay: number = 0, state?: any) : Subscription
```

This actually brings it *sort of* inline with methods like `setTimeout` to a degree.

blesh:
schedule is now: schedule(work, delay?, state?)
also cleaned up some tests that still used scheduling from the early days when subscription was still async

closes #265

### DIFF
--- a/spec/scheduler-spec.js
+++ b/spec/scheduler-spec.js
@@ -8,9 +8,9 @@ describe('Scheduler.immediate', function() {
     var call1 = false;
     var call2 = false;
     Scheduler.immediate.active = false;
-    Scheduler.immediate.schedule(0, null, function(){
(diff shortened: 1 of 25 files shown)


blesh:
I'm going to merge this right away to make room for the impending work on VirtualTimeScheduler

PR #268, opened 2015-09-04
feat(VirtualTimeScheduler): add VirtualTimeScheduler

### LINKED ISSUE #151: chore: add VirtualTime and Test Schedulers
Filing issue to track this chore. Since the current Scheduler interface is so small, I'd like to keep the VirtualTime/Test Scheduler interfaces small as well. Would love to hear feedback from @mattpodwysocki and @headinthebox on the minimal API required to get started.

We'll also need the ability to switch on the recursive-vs-trampolining strategies.

```es6
interface Scheduler {
  now() : number;
  schedule<T>(work: (state?: any) => void, delay: number = 0, state?: any): Subscription<T>;
}

interface VirtualTimeScheduler extends Scheduler {
  tick(timeElapsed: number): void;
}

interface TestScheduler extends VirtualTimeScheduler {
  advance(steps: number): void;
}
```

blesh:
- Refactors Action to be an interface
- Adds VirtualTimeScheduler
- Refactors Scheduler interface to include actions list

closes #151

### DIFF
--- a/spec/helpers/test-helper.js
+++ b/spec/helpers/test-helper.js
@@ -3,8 +3,26 @@
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 1000;
 
 var _ = require("lodash");
+var Rx = require('../../dist/cjs/Rx');
+
+global.rxTestScheduler = null;
+global.cold;
+global.hot;
+global.expectObservable;
+
 
 beforeEach(function () {
+  global.rxTestScheduler = new Rx.TestScheduler(assertDeepEqual);
+  global.hot = function () {
+    setupFlush();
+    return global.rxTestScheduler.createHotObservable.apply(global.rxTestScheduler, arguments);
+  };
+  global.cold = function () {
+    setupFlush();
+    return global.rxTestScheduler.createColdObservable.apply(global.rxTestScheduler, arguments);
+  };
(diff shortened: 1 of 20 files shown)


trxcllnt:
@blesh one thing I forgot to mention: we should probably refactor the [`Action.prototype.schedule`](https://github.com/ReactiveX/RxJS/blob/master/src/schedulers/Action.ts#L13) to `(state?: any, delay?: number)`.

Reasoning is that we want to recycle the existing Action as much as possible for recursive scheduling, but the Action may start out "immediate," but want to reschedule itself to run again in the future (as is potentially the case in the [`TimerObservable`](https://github.com/ReactiveX/RxJS/blob/master/src/observables/TimerObservable.ts#L26)).

This probably means we can get away with a single Action class that respects whatever "mode" it's presently scheduled as.

blesh:
@trxcllnt ... that's a good idea, but I think it can live in a separate issue.

Can you add an issue to track that?

blesh:
Ugh... yeah, more refactoring to do here. I think, you're right about needing to specify the delay on reschedule. One more commit...just one more...

blesh:
Since @trxcllnt reviewed and I made the requested changes, I've merged this as of af1cb68 so we can move on to converting over @mattpodwysocki's tests

PR #271, opened 2015-09-05
fix(bufferCount): set default value for skip argument, do not emit em…

kwonoj:
…pty buffer at the end.

Found while try to write micro perf test for bufferCount, when bufferCount does not have explicit skip argument (startBufferEvery parameter) it only emits very first buffer element then completes observable. This PR aligns bufferCount behaves as same as current bufferWithCount (https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/bufferwithcount.md) by set default skip argument to size of buffer if it's not specified.

### DIFF
--- a/src/operators/bufferCount.ts
+++ b/src/operators/bufferCount.ts
@@ -33,7 +33,7 @@ class BufferCountSubscriber<T> extends Subscriber<T> {
     const count = (this.count += 1);
     const destination = this.destination;
     const bufferSize = this.bufferSize;
-    const startBufferEvery = this.startBufferEvery;
+    const startBufferEvery = (this.startBufferEvery == null) ? bufferSize : this.startBufferEvery;
     const buffers = this.buffers;
     const len = buffers.length;
     let remove = -1;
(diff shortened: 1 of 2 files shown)


src/operators/bufferCount.ts
```diff
@@ -33,7 +33,7 @@ class BufferCountSubscriber<T> extends Subscriber<T> {
     const count = (this.count += 1);
     const destination = this.destination;
     const bufferSize = this.bufferSize;
-    const startBufferEvery = this.startBufferEvery;
+    const startBufferEvery = (this.startBufferEvery == null) ? bufferSize : this.startBufferEvery;
```
kwonoj: Some of other operator such as windowCount it has default value as 0 instead of null. Not sure if there's consistent recommendation or guidelines, so did not modify default value itself but simply take buffersize if value's not specified.

src/operators/bufferCount.ts
```diff
@@ -64,7 +64,10 @@ class BufferCountSubscriber<T> extends Subscriber<T> {
     const destination = this.destination;
     const buffers = this.buffers;
     while (buffers.length > 0) {
-      destination.next(buffers.shift());
+      var buffer = buffers.shift();
+      if (buffer.length > 0) {
```
kwonoj: This check is to prevent bufferCount emits empty buffer element at the end of observable on certain condition, such as ```Observable.range(0,6).bufferCount(2) ``` where empty buffer is pushed then no element is actually counted in.

blesh:
Good catch. I suspect we'll be finding lots of bugs like this as we convert tests over to use the Virtual Time Scheduler and expand testing to cover more scenarios.

blesh:
Thanks @kwonoj!

PR #273, opened 2015-09-06
fix(windowCount): set default value for skip argument, do not emit em…

kwonoj:
…pty buffer at the end.

As same as `bufferCount` implementation fixed in https://github.com/ReactiveX/RxJS/pull/271, windowCount also does not sets default skip argument if it's not specified. This PR `windowCount` to set default skip argument to size of window.

### DIFF
--- a/src/operators/windowCount.ts
+++ b/src/operators/windowCount.ts
@@ -25,29 +25,34 @@ class WindowCountOperator<T, R> implements Operator<T, R> {
 }
 
 class WindowCountSubscriber<T> extends Subscriber<T> {
-  private windows: { count: number, window: Subject<T> } [] = [];
+  private windows: { count: number, notified : boolean, window: Subject<T> } [] = [{ count: 0, notified : false, window : new Subject<T>() }];
   private count: number = 0;
   
   constructor(destination: Observer<T>, private windowSize: number, private startWindowEvery: number) {
-    super(destination); 
+    super(destination);
   }
   
   _next(value: T) {
     const count = (this.count += 1);
-    const startWindowEvery = this.startWindowEvery;
+    const startWindowEvery = (this.startWindowEvery > 0) ? this.startWindowEvery : this.windowSize;
     const windowSize = this.windowSize;
     const windows = this.windows;
+    const len = windows.length;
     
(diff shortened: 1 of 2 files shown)


src/operators/windowCount.ts
```diff
@@ -25,29 +25,34 @@ class WindowCountOperator<T, R> implements Operator<T, R> {
...
   
   constructor(destination: Observer<T>, private windowSize: number, private startWindowEvery: number) {
-    super(destination); 
+    super(destination);
   }
   
   _next(value: T) {
     const count = (this.count += 1);
-    const startWindowEvery = this.startWindowEvery;
+    const startWindowEvery = (this.startWindowEvery > 0) ? this.startWindowEvery : this.windowSize;
```
kwonoj: As mentioned in https://github.com/ReactiveX/RxJS/pull/271#discussion_r38808559, some function initialize default value to 0 and some sets it to null for `number` types. As same as previous changes only updates value but not default value since not sure if there's recommended guidelines.

src/operators/windowCount.ts
```diff
@@ -25,29 +25,34 @@ class WindowCountOperator<T, R> implements Operator<T, R> {
...
-      this.destination.next(window);
+      windows.push({ count: 0, notified : false, window : window });
     }
     
-    const len = windows.length;
     for (let i = 0; i < len; i++) {
       let w = windows[i];
       const window = w.window;
+      
+      if (!w.notified) {
```
kwonoj: This flag is introduce to let window to be notified only if window will have any values to be contained, such as `Observable.range(0,6).windowCount(2)` emits empty window at the end. Basic intention is to align operator behavior as same as `bufferCount` does.

note : it seems current RxJS (https://github.com/Reactive-Extensions/RxJS/issues/148) allows empty window being emitted at the end.

blesh:
Thanks for catching these, @kwonoj

merged as of a513dbb

src/operators/windowCount.ts
staltz: I'm now working on #378 and I changed the implementation of `windowCount` to align with RxJS legacy, and it seems like I brought the "empty window at the end" back.

@blesh @kwonoj

PR #274, opened 2015-09-07
chore(test): add micro-benchmark performance test for windowCount ope…

kwonoj:
…rator

Continue to cover #100.


blesh:
Merged with c632d04

You're a machine, @kwonoj! Your help is greatly appreciated!

kwonoj:
Thanks, :smile:

PR #279, opened 2015-09-08
refactor(Scalar): move to _isScalar from instanceof ScalarObservable

### LINKED ISSUE #276: Idea: Add a symbol for @@scalarValue to Observable for testing
This idea is to provide the same benefits to PromiseObservable that exist for ScalarObservable.

Once `PromiseObservable` resolves, we can stick that resolved value in a property on the Observable itself of `@@scalarValue`. In other operators were we're testing for `instanceof ScalarObservable`, that could be changed to simply examine for that property's existence and use that value. Effectively, resolved promise observables would become scalars at that point.

Benefits: 

1. It should perform better to check for a property than using `instanceof.
2. PromiseObservables would benefit from being treated as scalars after resolution.

Possible Risks:

1. Maybe some perf degradation from treating the observable instance as a hashtable? 

I have my doubts about the risk I listed above, but I wanted to list it.

cc/@trxcllnt @mattpodwysocki

blesh:
- add _isScalar property to Observables that could be scalar
- check for _isScalar rather than instanceof ScalarObservable
- add _isScalar property to PromiseObservable that defaults to false
- set _isScalar property to true when PromiseObservable's promise resolves
- also set the value property of the PromiseObservable

closes #276

### DIFF
--- a/src/observables/PromiseObservable.ts
+++ b/src/observables/PromiseObservable.ts
@@ -6,6 +6,9 @@ import immediate from '../schedulers/immediate';
 
 export default class PromiseObservable<T> extends Observable<T> {
 
+  _isScalar: boolean = false;
+  value: T;
+  
   static create<T>(promise: Promise<T>, scheduler: Scheduler = immediate) {
     return new PromiseObservable(promise, scheduler);
   }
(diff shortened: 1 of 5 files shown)


blesh:
What I ended up doing for the PR was actually adding a `_isScalar` property to Observables that *could* be scalar. From there I changed checks of `val instanceof ScalarObservable` to just check for `val._isScalar`.

I've changed the `PromiseObservable` to set `_isScalar` to `true` once the promise resolves, as well as set a `value` property to the resolved reason.

** Controversial? **

I've also made subsequent subscriptions to the PromiseObservable synchronous

blesh:
On the controversial bit, I didn't want users of Observables to be punished in performance because they wrapped a Promise.

cc/ @trxcllnt @jeffbcross @mattpodwysocki

blesh:
Merging in the name of progress. @trxcllnt did review.

PR #282, opened 2015-09-09
chore(test): add micro-benchmark performance test for catch operator

kwonoj:
Continue to cover #100.

Able to see noticeable slowness on new operator.

>old catch with immediate scheduler x 87,736 ops/sec ±3.78% (83 runs sampled)
>new catch with immediate scheduler x 61,858 ops/sec ±2.73% (79 runs sampled)
>       -29.5% slower than Rx2

>old catch with current thread scheduler x 73,027 ops/sec ±2.87% (78 runs sampled)
>new catch with current thread scheduler x 58,123 ops/sec ±2.95% (82 runs sampled)
>        -20.41% slower than Rx2

(number varies by system running tests, but shows same trend)

### DIFF
--- a/perf/micro/current-thread-scheduler/operators/catch.js
+++ b/perf/micro/current-thread-scheduler/operators/catch.js
@@ -1,0 +1,20 @@
+var RxOld = require("rx");
+var RxNew = require("../../../../index");
+
+module.exports = function (suite) {
+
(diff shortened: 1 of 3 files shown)


blesh:
Merged with bb66d4b

Thanks @kwonoj! I'll look into why catch is slower. I have a few ideas.

kwonoj:
Thanks, I'd like to investigate on my end too but since I'm new to codebase, wasn't trivial for me. Please let me know if I can followup on anything.

PR #284, opened 2015-09-09
refactor(test): let macro perf test use common generator function

kwonoj:
In macro perf test, I could see same generator function to create test observable is used by several test cases. This PR aims to have common generator function using existing perf-helper module can be used current & further test cases. PR does not include any functional changes or test coverage expansion.

### DIFF
--- a/perf/macro/flatMap-scalar/perf.js
+++ b/perf/macro/flatMap-scalar/perf.js
@@ -13,21 +13,8 @@ var RxNextFlatMapRange = document.querySelector('#rx-3-flatmap-range-to-scalar')
 var Rx2ObservableReturn = Rx.Observable.return;
 var RxNextObservableReturn = RxNext.Observable.of;
 
-var RxNextTestObservable = new RxNext.Observable(function(observer) {
-  var index = -1;
-  while(++index < numIterations) {
-    observer.next(index);
-  }
-  observer.complete();
-});
-
-var Rx2TestObservable = Rx.Observable.create(function(observer) {
-  var index = -1;
-  while(++index < numIterations) {
-    observer.onNext(index);
-  }
-  observer.onCompleted();
-});
+var RxNextTestObservable = RxNext.Observable.range(0, numIterations);
+var Rx2TestObservable = Rx.Observable.range(0, numIterations);
 
 Rx2FlatMapRange.addEventListener('click', function() {
   Rx2TestObservable.flatMap(projectionRx2).subscribe();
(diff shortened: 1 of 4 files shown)


blesh:
Actually looking at this, and given RxJS Next has no onNextor onCompleted functions (rather it has next and complete), I doubt things have been working correctly that use this generator code.

I love the spirit of this PR, though.

It might be a better idea to just start using the range method of each library. When the first macro perf tests were written for RxJS Next, we didn't have a range method, so what's been built off of was pretty old code.

So range might do for now.

The only caveat to that would be that range is a bit slower in RxJS 2/3 than it is in Next. So if that is decided to be an issue, I'd say make two functions that like generator, one for each library, only I'd have them just return an observable directly, rather that the subscriber function that's passed to the observable.

What are your thoughts?

kwonoj:
> It might be a better idea to just start using the range method of each library. 

Sounds plan for me. Basically what I did was simple copy-paste of those function from existing code and haven't look through deeply, seems missed those point until now. :(

> I'd say make two functions that like generator, one for each library, only I'd have them just return an observable directly

Seems I got bit lost for this. Would you able to share bit more detail? I came to think of pseudo like this at first sight, `getObservableRx2(iteration) {
  return Rx.Observable.range(...);
}` but feeling I don't understand correctly.

blesh:
I think just go with using `range` for now.

kwonoj:
OK, will update PR with `range` for now.

blesh:
@kwonoj can you flatten these commits into one commit?

blesh:
@kwonoj  ... to clarify, since the first of the commits introduced code that the second commit took away, it's probably worth doing an interactive rebase and "fixing up" the commits to just be one commit, which will slim it down to only the net change introduced.

kwonoj:
@blesh Yes, got point. Trying to flatten commits, will update once it's done.

kwonoj:
OK, gave a try and seems it's flattened into single commit, please let me know if it's not correct and need to be updated.

blesh:
Merged with 112005d

Thank you, as always @kwonoj

kwonoj:
@blesh Sorry for kind of off-topic question, I could see commit above but looking actual TOT of master branch commit history I see changes are not reflected somehow. Am I understanding something wrong or does not aware correctly?

blesh:
Hmm... perhaps I failed to merge this properly? I'll double check

blesh:
I think I got distracted and pushed this to master on my fork. It's in there now: ca42a32

kwonoj:
Thanks for checking!

PR #285, opened 2015-09-09
chore(perf): improve micro perf ergonomics

blesh:
- automatically pick up new files without maintaining a huge array
- add ability to filter down to specific tests: `node perf/micro testfilter` where "testfilter" is any string
  to be found inside the path of the test

@kwonoj ... this is for you.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,10 +33,22 @@
   ],
   "author": "Ben Lesh <blesh@netflix.com>",
   "contributors": [
-    { "name": "Ben Lesh", "email": "blesh@netflix.com" },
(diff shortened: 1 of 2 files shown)


kwonoj:
Cool! :+1:

blesh:
I've changed it so you must use a name that is the name of the test files to run. This is because `map` matched too many files. Now it must be exact. `zip`, or `take` or `map` or `flatMap` for example.

blesh:
Merged immediately so I can continue work.

blesh:
Incidentally, just running `node perf/micro` will still run all of the tests.

PR #286, opened 2015-09-09
chore(perf): fix catch micro perf test

### LINKED ISSUE #283: fix(catch): get catch perf up to snuff
As shown with the micro performance tests here #282, catch isn't performing as desired. This needs to be addressed.

In discussions with @trxcllnt, it seems perhaps the problem is around creation of too many subscribers. That's speculation. We'll need to dig in.

blesh:
The new method isn't polymorphic, and can only be called with a function

```sh
old catch with immediate scheduler x 235,384 ops/sec ±0.64% (94 runs sampled)
new catch with immediate scheduler x 1,621,709 ops/sec ±0.56% (96 runs sampled)
	588.96% faster than Rx2

old catch with current thread scheduler x 217,551 ops/sec ±1.02% (94 runs sampled)
new catch with current thread scheduler x 637,311 ops/sec ±0.54% (93 runs sampled)
	192.95% faster than Rx2
```

closes #283

### DIFF
--- a/perf/micro/current-thread-scheduler/operators/catch.js
+++ b/perf/micro/current-thread-scheduler/operators/catch.js
@@ -3,8 +3,10 @@ var RxNew = require("../../../../index");
 
 module.exports = function (suite) {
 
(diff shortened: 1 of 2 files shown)


blesh:
cc/ @kwonoj

kwonoj:
Thanks for pointing out.

PR #288, opened 2015-09-10
chore(test): add micro-benchmark performance test for groupby operator

kwonoj:
Continue to cover #100.


blesh:
Merged with 71841c9

Thanks again @kwonoj!

Issue #258, opened 2015-09-03
chore(test): add macro perf tests for groupBy


blesh:
Since groupBy is another operators that creates a lot of object instances (Observables and Subscriptions) under the hood, we need some macro performance tests for it.

PR #280, opened 2015-09-08
chore(test): add macro performance tests for merge

kwonoj:
Try to close https://github.com/ReactiveX/RxJS/issues/257.

After adding this test case I'm observing failures on my machine at the end of perf test using **Windows 10, Chrome 45.0.2454.85 m (64-bit)**. To verify it's occurring on other configuration created clean machine runnning **Ubuntu 12.02 LTS 45.0.2454.85 (64-bit) and test was exited without error**. Since test case itself ran successfully I suspect this is somewhat related with protractor issue (maybe this? https://github.com/angular/protractor/issues/2001) but do not have concrete clue. While trying to troubleshoot issue, creating PR to ask if anyone experiences similar cases.

>github\RxJS\node_modules\protractor\node_modules\selenium-webdriver\lib\webdriver\webdriver.js:377
>      throw new Error('This driver instance does not have a valid session ID ' +
>            ^
>Error: This driver instance does not have a valid session ID (did you call WebDriver.quit()?) and may no longer be used.
>    at checkHasNotQuit (Z:\github\RxJS\node_modules\protractor\node_modules\selenium-webdriver\lib\webdriver\webdriver.js:377:13)
>    at Z:\github\RxJS\node_modules\protractor\node_modules\selenium-webdriver\lib\webdriver\webdriver.js:350:5
>    at [object Object].webdriver.promise.ControlFlow.runInNewFrame_ (Z:\github\RxJS\node_modules\protractor\node_modules\selenium-webdriver\lib\webdriver\promise.js:1654:20)
>    at [object Object].webdriver.promise.ControlFlow.runEventLoop_ (Z:\github\RxJS\node_modules\protractor\node_modules\selenium-webdriver\lib\webdriver\promise.js:1518:8)
>    at [object Object].wrapper [as _onTimeout] (timers.js:265:14)
>    at Timer.listOnTimeout (timers.js:110:15)
>==== async task ====
>WebDriver.quit()
>    at [object Object].webdriver.WebDriver.schedule (Z:\github\RxJS\node_modules\protractor\node_modules\selenium-webdriver\lib\webdriver\webdriver.js:345:15)
>    at [object Object].webdriver.WebDriver.quit (Z:\github\RxJS\node_modules\protractor\node_modules\selenium-webdriver\lib\webdriver\webdriver.js:418:21)
>    at Z:\github\RxJS\node_modules\protractor\lib\driverProviders\driverProvider.js:59:14
[...]

PR #292, opened 2015-09-10
chore(test): add macro performance tests for groupBy

kwonoj:
Try to close https://github.com/ReactiveX/RxJS/issues/258.

Aware macro test cases are currently somewhat broken such as discussed in https://github.com/ReactiveX/RxJS/pull/280, still try to create coverage first as much.


blesh:
Merged with acd745e

Thank you so much for all of this work.

PR #293, opened 2015-09-10
refactor(operators): ensure _subscribe is being called where possible

### LINKED ISSUE #290: Some inner subscriptoins are calling `subscribe` and not `_subscribe`
If we already have a `Subscriber` instance, there's no need to call `subscribe` on an `Observable`, instead, `_subscribe` can be called directly.

blesh:
Many operators with inner subscriptions were calling `subscribe` when they could
have been calling `_subscribe`, thus avoiding an `instanceof Subscriber` check.

- makes sure `_subscribe` is being called where possible
- cleans up inner subscription handling in a few places
- seperates `switchAll` tests from `switchLatest` tests

closes #290

### DIFF
--- a/spec/operators/switchAll-spec.js
+++ b/spec/operators/switchAll-spec.js
@@ -1,0 +1,29 @@
+/* expect, it, describe */
+var Rx = require('../../dist/cjs/Rx');
+
+var Observable = Rx.Observable;
+var immediateScheduler = Rx.Scheduler.immediate;
+
+describe('Observable.prototype.switchAll()', function(){
+  it("should switch to each immediately-scheduled inner Observable", function (done) {
+    var a = Observable.of(1, 2, 3, immediateScheduler);
+    var b = Observable.of(4, 5, 6, immediateScheduler);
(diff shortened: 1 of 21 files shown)


blesh:
cc @trxcllnt

trxcllnt:
@blesh the only reason I had to call public subscribe when I did `flatMap` is because we're subscribing to Observables that might return something other than `void` from `_subscribe`. Is this PR worth moving to the Observable `start` API (which guarantees we can call `_subscribe` safely)?

blesh:
> Is this PR worth moving to the Observable start API (which guarantees we can call _subscribe safely)?

I'm going to say "not right now" only because the consensus I've gotten from others in the Rx community has been mixed. There haven't been any really strong arguments *for* the move to the `start` API unless we plan on supporting backpressure, which was nixed a long time ago.

For now, this is a small change, that only benefitted in small ways. 

Hopefully in the future we can test out the start API in ernest and show perf gains and solid ergonomics as a compelling argument.

PR #294, opened 2015-09-10
add basic esdoc generation

robwormald:


blesh:
Awesome! Thanks for getting this done, @robwormald! It's a good start.

blesh:
merged with 341475a

Thanks again!

Issue #260, opened 2015-09-03
chore(test): add macro perf test for combineLatest


blesh:
combineLatest has inner subscriptions, therefore we should be tracking it's macro performance.

PR #297, opened 2015-09-11
chore(test): add macro performance tests for combineLatest

kwonoj:
Trying to close https://github.com/ReactiveX/RxJS/issues/260.


blesh:
Merged with f258fc1

thanks, @kwonoj

PR #299, opened 2015-09-11
fix(root): use self as the root object when available

vsavkin:


blesh:
:raised_hands:

blesh:
This fixed #300

PR #303, opened 2015-09-12

kwonoj:
…Promise

Single specs created for default scheduler due to `RxOld.fromPromise()` doesn't accept scheduler for creation.


blesh:
merged as of 13ada1f

Thank you @kwonoj!

Issue #304, opened 2015-09-13
Marble diagram tests just don't work for synchronous scenarios


blesh:
For example: `Observable.of(1,2,3)` ... how are you going to represent 3 values and a completion in 1 frame? `123|`? and then how does that "line up" to expected results? What about 3 values and an error? what about 100ms, then 14 synchronous values, then 200ms then one value and it completes?

It's a shame because I find the look of the test DSL very compelling.

@mattpodwysocki can say "I told you so". :p

cc/@staltz and @trxcllnt

staltz:
Why do you even need a marble diagram for purely synchronous cases? I think the DSL should be used only where convenient.

headinthebox:
And since it is just a string, you can invent new notation for the sync case, like [1,2,3] or 1#2#3.

blesh:
>Why do you even need a marble diagram for purely synchronous cases? I think the DSL should be used only where convenient.

Imagine `last`,  observable runs for some amount of time, then when it completes, it synchronously nexts, then completes in the same frame. That's not `------a|` by our notation.

I've been thinking of something like @headinthebox was saying, but it messes with the ability to line them up.

```
e3.withLatestFrom(e1,e2)


e1 -a---|
e2 ---b-----|
e3 ------c---d----|
r  ------[cab]----|

// What about `[dab]`? 
```

I spouse the upside is, since they're tests you can specify that this shouldn't happen, but it worries me that I'll run into something else I haven't considered

staltz:
Then just define e3 differently:

e3.withLatestFrom(e1,e2)

```
e1 -a---|
e2 ---b-----|
e3 ------c-----d----|
r  ------[cab]-[dab]--|
```

blesh:
Right. That's what I was saying. I'm just worried that there's something else I'm missing

staltz:
Completely irrelevant to this thread, but since we're at it, for purely aesthetic purposes I prefer parentheses:

`r ----(cab)--|`

blesh:
I like that, it doesn't visually match the pipe as much

PR #305, opened 2015-09-13
feat(operator): add elementAt

kwonoj:
Effort to add operator exists in current RxJS, picking up relatively simple operator as starting point.

### DIFF
--- a/perf/micro/immediate-scheduler/operators/elementat.js
+++ b/perf/micro/immediate-scheduler/operators/elementat.js
@@ -1,0 +1,20 @@
+var RxOld = require("rx");
+var RxNew = require("../../../../index");
+
+module.exports = function (suite) {
+
+  var oldElementAtWithImmediateScheduler = RxOld.Observable.range(0, 25, RxOld.Scheduler.immediate).elementAt(5);
+  var newElementAtWithImmediateScheduler = RxNew.Observable.range(0, 25).elementAt(5);
+
+  return suite
+    .add('old elementAt with immediate scheduler', function () {
+      oldElementAtWithImmediateScheduler.subscribe(_next, _error, _complete);
+    })
+    .add('new elementAt with immediate scheduler', function () {
+      newElementAtWithImmediateScheduler.subscribe(_next, _error, _complete);
+    });
+
+  function _next(x) { }
+  function _error(e){ }
+  function _complete(){ }
+};
--- a/spec/operators/elementat-spec.js
+++ b/spec/operators/elementat-spec.js
@@ -1,0 +1,47 @@
+/* globals describe, it, expect, hot, cold, expectObservable */
+var Rx = require('../../dist/cjs/Rx');
+var Observable = Rx.Observable;
+
+describe('Observable.prototype.elementAt', function() {
+  it("should return first element by zero-based index", function() {
+    var source = hot('--a--b--c--|');
+    var expected =   '--(a|)';
+		
+    expectObservable(source.elementAt(0)).toBe(expected);
+  });
+	
+  it("should return non-first element by zero-based index", function() {
+    var source = hot('--a--b--c--d--e--f--|');
+    var expected =   '-----------(d|)';
+		
+    expectObservable(source.elementAt(3)).toBe(expected);
+  });
+  
+  it("should return last element by zero-based index", function() {
+    var source = hot('--a--b--c--|');
+    var expected =   '--------(c|)';
+		
+    expectObservable(source.elementAt(2)).toBe(expected);
+  });
+  
+  it("should throw if index is smaller than zero", function() {
+    expect(function() { Observable.range(0,10).elementAt(-1); })
+      .toThrow(new Rx.ArgumentOutOfRangeError);
+  });
+	
+  it("should raise error if index is out of range but does not have default value", function() {
+    var source = hot('--a--|');
+    var expected =   '-----#';
+    
+    expectObservable(source.elementAt(3))
+      .toBe(expected, null, new Rx.ArgumentOutOfRangeError);
+  });
+	
+  it("should return default value if index is out of range", function() {
+    var source = hot('--a--|');
+    var expected =   '-----(x|)';
+    var defaultValue = '42';
+    
+    expectObservable(source.elementAt(3, defaultValue)).toBe(expected, { x: defaultValue });
+  });
+});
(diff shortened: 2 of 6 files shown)


perf/micro/immediate-scheduler/operators/elementat.js
```diff
@@ -0,0 +1,20 @@
+var RxOld = require("rx");
+var RxNew = require("../../../../index");
+
+module.exports = function (suite) {
+
+  var oldElementAtWithImmediateScheduler = RxOld.Observable.range(0, 25, RxOld.Scheduler.immediate).elementAt(5);
```
kwonoj: As same as test case https://github.com/ReactiveX/RxJS/pull/303, this operator does not accept any schedulers. I do not feel let those kind of test belong under current micro test category (immediate / current thread) is correct categorization, maybe need new category such as 'default', or others?

spec/operators/elementat-spec.js
```diff
@@ -0,0 +1,40 @@
...
+    expectObservable(source.elementAt(3))
+      .toBe(expected, null, 'Argument out of range');
+	});
+	
+	it("should return default value if index is out of range", function() {
+		var source = hot('--a--|', { a: 'a' });
+    var expected =   '-----x|';
+    var defaultValue = '42';
+    
+    //expectObservable(source.elementAt(3, defaultValue)).toBe(expected, { x: defaultValue });
```
kwonoj: Tried to follow test case description using marble diagrams but having some trouble with test specific test case. In case of this, actual result is `42|complete()` arrives in same frame. Seems this is related with issue https://github.com/ReactiveX/RxJS/issues/304 - am I understand correctly?

spec/operators/elementat-spec.js
```diff
@@ -0,0 +1,47 @@
...
+    expectObservable(source.elementAt(3))
+      .toBe(expected, null, 'Argument out of range');
+  });
+	
+  it("should return default value if index is out of range", function() {
+    var source = hot('--a--|', { a: 'a' });
+    var expected =   '-----x|';
+    var defaultValue = '42';
+    
+    //expectObservable(source.elementAt(3, defaultValue)).toBe(expected, { x: defaultValue });
```
kwonoj: Tried to follow test case description using marble diagrams but having some trouble with specific case. In case of this, actual result is `42|complete()` arrives in same frame. Seems this is related with issue https://github.com/ReactiveX/RxJS/issues/304 - am I understand correctly?

staltz: You mean actual result is `-----[42,|]`? where `[...]` happens in the same frame

kwonoj: @staltz Yes, correct. My expression wasn't clear enough.

staltz: When, it seems clear we need the `[a,b]` syntax in the DSL then. Even Meijer suggested it

blesh:
I'm ahead of you. It'll be in there tomorrow

kwonoj:
@blesh Shall I close PR then?

blesh:
Nah, leave it open and wait for the testing changes to land.

blesh:
Testing changes are part of #307

spec/operators/elementat-spec.js
```diff
@@ -0,0 +1,47 @@
...
+    var source = hot('--a--|', { a: 'a' });
+    var expected =   '-----#';
+    
+    expectObservable(source.elementAt(3))
+      .toBe(expected, null, 'Argument out of range');
+  });
+	
+  it("should return default value if index is out of range", function() {
+    var source = hot('--a--|', { a: 'a' });
+    var expected =   '-----x|';
```
blesh: Okay, once you rebase, this can now be:

```js
var expected =   '-----(x|)';   
```

kwonoj: Yes, just saw changes merged and about to start work on changes. Thanks!

kwonoj:
Updated changes with 
- now test case updated with grouping marble ability
- add argumentoutofrange error type

spec/operators/elementat-spec.js
```diff
@@ -0,0 +1,47 @@
...
+    var source = hot('--a--|');
+    var expected =   '-----#';
+    
+    expectObservable(source.elementAt(3))
+      .toBe(expected, null, new Rx.ArgumentOutOfRangeError);
+  });
+	
+  it("should return default value if index is out of range", function() {
+    var source = hot('--a--|');
+    var expected =   '-----(x|';
```
blesh: This can be closed like `-----(x|)`, it will *technically* work like you've done it because of the naive implementation, but it's probably better to close the group in case the implementation changes.

kwonoj: Agreed. let me update quickly to close.

src/operators/elementAt.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+  
+  call(subscriber: Subscriber<T>): Subscriber<T> {
+    return new ElementAtSubscriber(subscriber, this.idx, this.defaultValue);
+  }
+}
+
+class ElementAtSubscriber<T, R> extends Subscriber<T> {
+  
+  idx: number;
```
blesh: please name `index` and make this `private`

`private index:number = 0;`

It's a good idea to use `private` or `protected` where appropriate, because we plan on using a TypeScript clojure-compiler-esque minifier from our friends at Google (@jeffbcross) that will use that information to rename properties and reduce output size.

src/operators/elementAt.ts
```diff
@@ -0,0 +1,58 @@
...
+  defaultValue: any;
+  
+  constructor(destination: Subscriber<T>, idx: number, defaultValue?: any) {
```
blesh: in fact, this whole block could just be: 

```TypeScript
constructor(destination: Subscriber<T>, private index: number, private defaultValue?: any) {
  super(destination);
}
```

src/operators/elementAt.ts
kwonoj: will do. Just curiosity - does guideline recommends to init with default value when it's declared? i.e, 
`private idx:number = 0` vs `private idx:number`. Personally do not have any specific preferences.

src/operators/elementAt.ts
kwonoj: Yes, thanks for pointing.

kwonoj:
Updated changes with
- closed parentheses for test marbles
- init private field in ctor directly

kwonoj:
> please name `index` and ...

Additional correction I forgot to do is updated.

blesh:
Awesome... I'll look at this later this evening or in the morning.

blesh:
merged with cd562c4

thanks @kwonoj!

PR #307, opened 2015-09-13
feat(operator): add last operator and update marble tests to support grouping

### LINKED ISSUE #304: Marble diagram tests just don't work for synchronous scenarios

### LINKED ISSUE #306: Add last operator
Just what it says, we need the `last` operator.

blesh:
- adds EmptyError error type
- adds ability to group values in marble tests
- adds last operator

closes #304
closes #306

### DIFF
--- a/spec/operators/last-spec.js
+++ b/spec/operators/last-spec.js
@@ -1,0 +1,22 @@
+/* globals describe, it, expect, expectObservable, hot, cold */
+var Rx = require('../../dist/cjs/Rx');
+
+describe('Observable.prototype.last()', function(){
+  it('should take the last value of an observable', function(){
+    var e1 = hot('--a--^--b--c--|');
+    var expected =    '---------(c|)';
+    expectObservable(e1.last()).toBe(expected)
+  });
+  
+  it('should error on empty', function() {
+    var e1 = hot('--a--^----|');
+    var expected =    '-----#';
+    expectObservable(e1.last()).toBe(expected, null, new Rx.EmptyError());
+  });
+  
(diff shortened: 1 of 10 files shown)


blesh:
now the marble tests support grouping.. where the `(` designates the frame a group is emitted on, and it gathers all of the values until it finds a `)` ... nested groups are *not* supported.

```
merging this:

----a------|
----(bc)---|

makes this:

----(abc)--|
```

blesh:
attn @kwonoj @staltz

spec/operators/last-spec.js
```diff
@@ -0,0 +1,22 @@
+/* globals describe, it, expect, expectObservable, hot, cold */
+var Rx = require('../../dist/cjs/Rx');
+
+describe('Observable.prototype.last()', function(){
+  it('should take the last value of an observable', function(){
+    var e1 = hot('--a--^--b--c--|');
+    var expected =    '---------(c|)';
```
kwonoj: Recommend to unclose parenthesis when completed?

blesh: That's a nexted value and a completion in the same frame

PR #296, opened 2015-09-11
refactor(operator): ensure build time parameter type safety for _subs…

kwonoj:
…cribe, call.

Try to close https://github.com/ReactiveX/RxJS/issues/291

PR #308, opened 2015-09-14
refactor(operator): ensure parameter type safety for each subscriber …

kwonoj:
…ctor

This PR is aligning changes followed by https://github.com/ReactiveX/RxJS/pull/296 for each subscriber construction. Previous PR ensures parameter type of `call` to `Subscriber<>` and each subscriber constructor still able to accept `Observer<>`. This isn't big issue since `call` already guards type so subscriber constructor won't be able to accept other than `Subscriber<>` but better to have build type safety.

### DIFF
--- a/src/observables/ForkJoinObservable.ts
+++ b/src/observables/ForkJoinObservable.ts
@@ -11,7 +11,7 @@ export default class ForkJoinObservable<T> extends Observable<T> {
     return new ForkJoinObservable(observables);
   }
   
-  _subscribe(subscriber: Observer<any>) {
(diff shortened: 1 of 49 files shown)


blesh:
LGTM

blesh:
merged with 9e7e560

It was a tough merge because I changed a lot of files in my big PR the other day (#316) ... some of your changes *may* have been reverted.

Thanks again!

kwonoj:
Thanks, I'll check TOT and create following PR if any following change's required.

Issue #309, opened 2015-09-14
Find all tests with `setTimeout` in them and replace them


blesh:
Now that we have the TestScheduler, I'm sick of waiting for the tests with *actual* timing in them. I need to find them and kill them, replacing them with TestScheduler "marble" tests.

kwonoj:
Candidates are (most of) all test cases exist under /spec/**, is that correct?

blesh:
Yes, that's what I'm talking about.

blesh:
I think we've accomplished this for the most part.

PR #311, opened 2015-09-14
refactor(test): use test marbles for combinelatest observable spec

kwonoj:
Starting effort to cover https://github.com/ReactiveX/RxJS/issues/309


blesh:
merged with 93a5bd9

thanks again!

PR #312, opened 2015-09-15
fix(last): emit value matches with predicate instead of result of pre…

kwonoj:
…dicate

Minor correction to `last` operator to emit actual values instead of emitting result of predicate if predicate matches. Also added small test case to verify behavior.


blesh:
Haha. Don't code tired. Good catch!

kwonoj:
:)

blesh:
merged with 0f635ee

again, thanks for finding that for me, @kwonoj. Probably should have added more tests.

PR #313, opened 2015-09-15
refactor(test): format test failure output

kwonoj:
This PR appends carriage return per each object of test marble array to have better visibility of output if test fails. 

*Before*
before

*After*
after


blesh:
THIS IS THE GREATEST PR IN THE HISTORY OF THIS PROJECT

I'll be merging this immediately, thank you very much.

blesh:
merged with daa5f1d

Awesome, thanks @kwonoj!

kwonoj:
My pleasure!

PR #314, opened 2015-09-15
refactor(test): let macro test for combinelatest use spec initializer

kwonoj:
Migrate all remaining macro test to use common initializer.


blesh:
Merged with a7ab267

thank you, @kwonoj

PR #315, opened 2015-09-15
fix(buffer): do not emit empty buffer when completes

kwonoj:
- do not emit empty buffer when observable completes
- update test spec to use marble test scheduler

effort to cover https://github.com/ReactiveX/RxJS/issues/309


blesh:
Merged with 252fccb

Thanks, @kwonoj

Issue #302, opened 2015-09-12
SwitchAll / SwitchLatest are broken


jinroh:
`switch` operators do not seem to unsubscribe properly their inner subscriptions.

Here is a small repro code:

```js
Observable.interval(1000)
  .switchLatest(function(i) {
    return Observable.interval(100).do(function(j) { console.log(i + "_" + j); });
  })
  .subscribe(function() {
  });
```

blesh:
Thank you for pointing this out. Looking into this and expanding tests uncovered a variety of anomalies. I've been heads down refactoring code in this area.

blesh:
@jinroh ... can you please checkout #317 and see if it solves your issues? I suspect it will. I was able to confirm your issue and write tests around it.

jinroh:
I think it solves one of the problem.
Here the example code I use:

```js
var Rx = require("./cjs/Rx");

var Observable = Rx.Observable;

Observable.interval(1000).switchLatest(function(i) {
  return Observable.interval(100).map(function(j) {
    console.log("inner", i + "_" + j);
    return i + "_" + j;
  });
})
  .subscribe(function(v) {
    console.log("outer", v);
  });
```

Outer subscription behaves as expected now:

```
outer 0_0
outer 0_1
outer 0_2
outer 0_3
outer 0_4
outer 0_5
outer 0_6
outer 0_7
outer 0_8
outer 1_0
outer 1_1
outer 1_2
outer 1_3
...
```

*However*, the inner subscriptions (from the map operator in my example) do not seem to be cleaned up. The "inner interval" side effects are still present. It looks like the subscription graph is not entirely disposed as it should be, causing a leak:

```
inner 0_0
inner 0_1
inner 0_2
inner 0_3
inner 0_4
inner 0_5
inner 0_6
inner 0_7
inner 0_8
inner 0_9
inner 1_0
inner 0_10
inner 1_1
inner 0_11
inner 1_2
inner 0_12
inner 1_3
inner 0_13
...
```

Not sure if this problem is linked to the `Switch..` operator or more general. In my tests, this is the only operator I've used that internally unsubscribe inner-subs.

blesh:
Returning subscriptions from the project function in the operator doesn't do anything special. They're not being cleaned up because you're creating new subscriptions and never unsubscribing therm. I'm actually amazed that worked at all. I'll check that out.

jinroh:
Thanks !

[...]

Issue #222, opened 2015-08-27
refactor(subscribers): reduce inheritence usage in subscribers


blesh:
We have a few operators that have gotten a little deep with inheritance. In particular `switchLatest` (aka flatMapLatest), in which

`SwitchLatestSubscriber extends FlatMapSubscriber`
`FlatMapSubscriber extends MergeSubscriber`
and `MergeSubscriber extends Subscriber`

The problems with this are: 
- It gets a little hard to follow after a bit. 
- refactoring becomes more difficult
- less modularity

If shared functionality/methods are a goal, perhaps it would be better to share them via module imports?

attn @trxcllnt

blesh:
It's my *opinion* that we should probably keep inheritance to one level deep, with the occasional exception.

blesh:
I think this has been met to some degree by #316. Closing for now.

PR #316, opened 2015-09-15
Merge refactor

blesh:
Due to problems outlined in #302, I ended up adding more virtual time tests and found quite some behaviors weren't working as expected. Since #222 is still open, and it's something I wanted to tackle, and since the high level of inheritance in operators like `switchAll` made debugging a little painful, I opted for a refactor.

Unfortunately, refactoring merge touched a lot of other operators (`concat`, `concatAll`, `switchAll`, `switchLatest`, `switchLatestTo`, `flatMap`, `flatMapTo`, `merge`, `mergeAll`, `Observable.merge`, `Observable.concat`, and `expand`)

The basics of the changes are as follows:

- `expand` has it's own standalone subscriber
- `flatMap` has it's own standalone subscriber
- `switchAll` has it's own standalone subscriber
- `mergeAll` has it's own standalone subscriber

- `switchLatest` (which should be `switchMap`) inherits from `flatMap` subscriber
- `flatMapTo` inherits from `flatMap` subscriber
- `switchLatestTo` inherits from `flatMapTo` subscriber

other changes were just changes to calls to reflect better typing information.

Also, I had to fix a bug in the TestScheduler that was having `hot` observables subscribe and emit *before* the expectation observables did. This meant that if a subject (hot observable) emitted on frame 0, it wouldn't be provided to the expectation observable being tested because it hadn't subscribed yet.

### sincerest apologies for such a large PR

... at least I broke it into several commits.


blesh:
Oh, one additional thing... 

- Added `concurrency` limit to `expand` operator.

src/operators/switchLatest.ts
```diff
@@ -4,50 +4,30 @@ import Observable from '../Observable';
...
-  }
-
-  _subscribeInner(observable, value, index) {
-    this.innerSubscription = new Subscription();
-    this.innerSubscription.add(super._subscribeInner(observable, value, index));
-    return this.innerSubscription;
+  constructor(destination: Observer<T>,
+              project: (value: T, index: number) => Observable<R>,
+              resultSelector?: (innerValue: R, outerValue: T, innerIndex: number, outerIndex: number) => R2) {
+    super(destination, project, resultSelector, 1);
```
trxcllnt: Does this work? It looks like this makes the `SwitchLatestSubscriber` equivalent to `concat`... right?

PR #317, opened 2015-09-16
refactor(switchLatest): make switchLatest a standalone impl, add virtual time tests

blesh:
- completely refactors switchLatest to be a standalone implementation
- adds a lot of tests around switch latest

### DIFF
--- a/spec/operators/switchLatest-spec.js
+++ b/spec/operators/switchLatest-spec.js
@@ -3,7 +3,7 @@ var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 var immediateScheduler = Rx.Scheduler.immediate;
 
-fdescribe('Observable.prototype.switchLatest()', function () {
+describe('Observable.prototype.switchLatest()', function () {
   it("should switch with a selector function", function (done) {
     var a = Observable.of(1, 2, 3);
     var expected = ['a1', 'b1', 'c1', 'a2', 'b2', 'c2', 'a3', 'b3', 'c3'];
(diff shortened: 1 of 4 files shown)


spec/operators/switchLatest-spec.js
```diff
@@ -3,15 +3,166 @@ var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 var immediateScheduler = Rx.Scheduler.immediate;
 
-describe('Observable.prototype.switchLatest()', function () {
+fdescribe('Observable.prototype.switchLatest()', function () {
```
kwonoj: **f**describe?

blesh: Oops. Good catch.

That's to limit jasmine tests to just that grouping during development. I meant to remove that.

src/operators/switchAll.ts
```diff
@@ -58,13 +66,18 @@ class SwitchSubscriber<T> extends Subscriber<T> {
 }
 
 class InnerSwitchSubscriber<T> extends Subscriber<T> {
-  constructor(destination: Observer<any>, private parent: SwitchSubscriber<T>) {
-    super(destination);
```
trxcllnt: couldn't we still set `destination` on the `InnerSwitchSubscriber`, but invoke `super` with no argument? I believe the only logic we want to short-circuit here is the [shared `_subscription` assignment](https://github.com/ReactiveX/RxJS/blob/master/src/Subscriber.ts#L45).

blesh: Yes, that would probably work. If you want to make that change, feel free to PR.

trxcllnt: I'm going to devote some thought to the whole shared-subscription approach and see if I can poke any more holes in it. The solution may be as simple as explicitly passing the shared subscription to Subscriber's constructor, instead of reading it off the destination. Will let you know.

PR #319, opened 2015-09-16
feat(flatMap) Add support for returning Arrays from a flatMap selector.

trxcllnt:


mattpodwysocki:
@trxcllnt instead of Arrays, I would say support iterables which could be arrays, maps, sets, and even generators as it is currently supported in RxJS

xgrommx:
@trxcllnt @mattpodwysocki Why not just `flatMap(...).toArray()`?

mattpodwysocki:
@trxcllnt @xgrommx that doesn't make any sense because it won't give us the same as a selector return value

trxcllnt:
@mattpodwysocki I agree, but I needed this in a project last night, so y'know... incremental development

blesh:
This is pre-alpha... the skateboard could be merged to the smiley-face's small intestines.

blesh:
This LGTM. I think we need a broader approach that utilizes techniques like this for dealing with inner subscriptions.

blesh:
This has been merged with 4ad6797

PR #320, opened 2015-09-16
bugfix(expand): Fix expand to stay open until the source Observable completes.

trxcllnt:


blesh:
Can you rename the Commit to start with `fix()` rather than `bugfix`? That way it falls into one category.

trxcllnt:
@blesh done

blesh:
LGTM

blesh:
merged with 20ef785

Thanks!

PR #321, opened 2015-09-16
refactor(expand): Export expand operator/subscriber in its own support file.

trxcllnt:


blesh:
Strange that this didn't pass the tests for expand if all you did was move the classes to another file. Where there other changes? It's hard to tell because all of the code is changing files.

kwonoj:
I'm seeing one difference in 
`private hasCompleted: boolean = false;`

where it's originally 
`private hasCompleted: boolean = true;`

since test fails due to difference of completion frame, maybe this is related?

trxcllnt:
@blesh @kwonoj sorry, this PR was made off master when I forgot I had to fix the test as well. See this line: https://github.com/ReactiveX/RxJS/pull/320/files#diff-b427369af61529d4d02959b22714161aL14

blesh:
@trxcllnt are you talking about #320? Do you need that to be merged first?

trxcllnt:
@blesh that would be nice. I'll fix that PR.

trxcllnt:
@blesh fixed the unit test the same way #320 does so there won't be any merge conflicts if these get merged separately.

blesh:
@trxcllnt ... it looks like this branch has conflicts now that I've merged #320.

trxcllnt:
@blesh merge conflicts resolved.

blesh:
Thanks, I'll take a look

blesh:
merged with 245e8d4

:)

PR #324, opened 2015-09-16
docs(contribution): change spec directory link

adrianomelo:


blesh:
merged with 118fe98

Thanks, @adrianomelo! Good catch.

PR #326, opened 2015-09-16
test(combineLatest): add tests with empty or never sources

adrianomelo:
Observable.combileLatest should be never when:
 - sources are never and never
 - one of the sources is never
 - one of the sources is never and the other has values
 - one of the sources is never and the other complete
 - sources are empty and empty
 - one of the sources is empty
 - one of the sources is empty and the other complete

with two failing tests:
 - Observable.combineLatest should be never when sources are empty and empty
 - Observable.combineLatest should be never when one of the sources is empty and the other complete

### DIFF
--- a/spec/observables/combineLatest-spec.js
+++ b/spec/observables/combineLatest-spec.js
@@ -3,15 +3,69 @@ var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
 describe('Observable.combineLatest', function () {
+  function concat(x, y) { return '' + x + y; }
+
+  it('should be never when sources are never and never', function () {
+    var e1 = Observable.never();
+    var e2 = Observable.never();
+    var expected = '-';
+    var combined = Observable.combineLatest(e1, e2, concat);
+    expectObservable(combined).toBe(expected);
+  });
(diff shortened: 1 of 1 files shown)


adrianomelo:
Hi guys,

This is my first patch with code to this project..
There are two failing tests in this patch, I would like to know your opinion about which code is incorrect.. combineLatest or these 2 tests?

spec/observables/combineLatest-spec.js
```diff
@@ -3,15 +3,69 @@ var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
 describe('Observable.combineLatest', function () {
+  function concat(x, y) { return '' + x + y; }
+
+  it('should be never when sources are never and never', function () {
```
kwonoj: Isn't this case already exists in [spec/operators/combineLatest-spec](https://github.com/ReactiveX/RxJS/blob/master/spec/operators/combineLatest-spec.js#L77)?

adrianomelo: You are right! I didn't know this file.. I'll close this PR and take a look in this file.

PR #327, opened 2015-09-16
refactor(test): use test marbles for skip operator spec

kwonoj:
Continuing effort to cover #309


blesh:
merged with 348c26a

Thanks, as always, @kwonoj!

PR #328, opened 2015-09-17
refactor(test): clean up micro benchmark test for last operator

kwonoj:
Minor cleanup for micro benchmark test for `last` operator to have better visibility.


blesh:
Merged with bf2434a

Small change: Commit message labeled `chore()`rather than `refactor()` since it isn't refactoring library code and is only affecting tests.

Thanks @kwonoj!

kwonoj:
Thanks, I'll follow those for further commit messages.

PR #331, opened 2015-09-17
feat(config): add global configuration of Promise capability

### LINKED ISSUE #115: Configuration points for Promise
We need to add a configuration point for using alternative Promise libraries. We also need to allow developers to pass a Promise constructor to each method that might create a Promise internally. `forEach` once such example.

- [ ] add configuration point for Promises
- [ ] add optional Promise constructor parameter to `forEach`

blesh:
- Adds a minimal global configuration for Promises
- Adds a PromiseCtor optional parameter to `Observable.prototype.forEach`
- Adds some basic tests for `Observable.prototype.forEach`

closes #115

### DIFF
--- a/spec/helpers/test-helper.js
+++ b/spec/helpers/test-helper.js
@@ -78,3 +78,29 @@ afterEach(function () {
 function assertDeepEqual(actual, expected) {
   expect(actual).toDeepEqual(expected);
 }
+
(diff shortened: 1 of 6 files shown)


blesh:
Merging since this is blocking alpha release.

PR #336, opened 2015-09-17
chore(test): add macro performance tests for bufferCount

kwonoj:
Continue effort to cover #100.


blesh:
merged with 1353924

Thanks @kwonoj!

PR #337, opened 2015-09-18
fix(bufferTime): inner intervals will now clean up properly

blesh:
- adds marble tests around bufferTime
- adds `maxFrames` property to `VirtualTimeScheduler` that will limit the execution of tests


blesh:
attn @kwonoj @trxcllnt @steveorsomethin @mattpodwysocki... This PR has changes in it to VirtualTimeScheduler/TestScheduler that allow developers to test "unlimited" observables, in that it defaults to max out execution at 750 frames (75 character marbles). This can be set via `rxTestScheduler.maxFrames = 1000` (or whatever number you pick) at any point during the test.

blesh:
merged with 4ef41b0

PR #342, opened 2015-09-18
chore(test): add micro-benchmark performance test for Observable.never

kwonoj:
Continue to cover #100.


kwonoj:
Personally bit doubt if this test case is legit, created PR to be reviewed if there's any suspicious.

blesh:
I've merged it for now. It's a foot in the door 1c6db5a

However, I expected the perf here to be a little closer. It's worth investigating.

PR #345, opened 2015-09-18
refactor(ArrayObservable): array observables of one will now be treat…

### LINKED ISSUE #278: Have ArrayObservable.create return an Empty or ScalarObservable if possible.
This could then be used in `from` and `fromArray` so the resulting Observable performed better in various operations.

blesh:
…ed as scalars

closes #278

### DIFF
--- a/src/observables/ArrayObservable.ts
+++ b/src/observables/ArrayObservable.ts
@@ -46,9 +46,16 @@ export default class ArrayObservable<T> extends Observable<T> {
 
     (<any> this).schedule(state);
   }
+  
+  // value used if Array has one value and _isScalar
+  value: any;
 
   constructor(private array: T[], private scheduler?: Scheduler) {
     super();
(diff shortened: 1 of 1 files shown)


src/observables/ArrayObservable.ts
```diff
@@ -46,9 +46,16 @@ export default class ArrayObservable<T> extends Observable<T> {
...
   }
+  
+  // value used if Array has one value and _isScalar
+  value: any;
 
   constructor(private array: T[], private scheduler?: Scheduler) {
     super();
+    if(array.length === 1) {
+      this._isScalar = true;
+      this.value = array[0];
```
trxcllnt: @blesh what happens if a Scheduler is specified?

blesh:
Also added a minor refactor to the loop to "golf" it down a little bit and make it more readable (in this beholder's eyes).

PR #349, opened 2015-09-18
chore(test): format test failure output more aligned

kwonoj:
one line of (no, actually couple of) change to align test failure output format between actual vs. expected, 

*Before*
before

*After*
after


blesh:
Nice, thanks!

PR #353, opened 2015-09-18
feat(operator): add isEmpty operator

kwonoj:


kwonoj:
Updated code per suggestion.

blesh:
Merged with 80f72c5

Thank you @kwonoj!

PR #355, opened 2015-09-19
chore(test): use test marbles for delay operator spec

kwonoj:
- replace test spec to use test marble
- virtual time scheduler returns virtual timeframe of current for now()

This PR replaces test spec for delay operator to use test marble.

In addition to  those, let TestScheduler & VirtualTimeScheduler supports `now()` to return 'current virtual time' based on test marble frame to support some operators utilizes it, such as `delay`.


blesh:
> virtual time scheduler returns virtual timeframe of current for now()

Nice.

blesh:
Merged with 8ba2e5f

Thanks!

PR #356, opened 2015-09-19
chore(test): add micro-benchmark performance test for single operator

kwonoj:
Fill one missing micro test case for `single` without any parameters.


blesh:
Merged with b4d0a0f

Issue #352, opened 2015-09-18
Add micro perf tests for creation methods


blesh:
From #100 

- [x] fromEvent
- [x] fromEventPattern
- [x] fromPromise
- [x] interval
- [x] timer
- [x] zip

kwonoj:
- `never` is covered at https://github.com/ReactiveX/RxJS/pull/342 though it might need some investigation separately?

- isn't `just/value` is deprecated and `of` covers it instead?

blesh:
@kwonoj you're correct. That was a copy-pasta mistake. Edited.

kwonoj:
`fromPromise` also seems covered via https://github.com/ReactiveX/RxJS/pull/303

blesh:
Looks like we can close this... thanks for all the hard work @kwonoj!

PR #357, opened 2015-09-20
chore(test): add micro-benchmark performance test for Observable.zip

kwonoj:
- move Observable.zip test from operator test location
- add Observable.prototype.zip test to operator test location
- remove zip test from current thread tests

This PR includes **changes to existing micro perf test cases**, details are below.

1. existing test case under immediate-scheduler/operator/ is moved to .../observable/
: Existing case was already testing `Observable.zip()` for creating method

2. existing test case under current-thread-scheduler/operators/ is removed
: since `zip()` itself does not accept scheduler, test under current thread scheduler does test over different schedulers only for sources. As same as other creation method `fromPromise` or `never`, let test to be located under immediate scheduler.

3. add new test case under immediate-scheduler/operator/ for operator
: Now ./operator/zip test case tests `Observable.prototype.zip()`

Little bit unsure all of this changes are legit, created PR to ask opinions. Any opinion / suggestions would be appreciated.


blesh:
Looks good to me.

blesh:
merged with a3917b6

Thanks @kwonoj!

PR #358, opened 2015-09-20

kwonoj:
…Event

Used `EventEmitter` as source then limit to single element.


blesh:
Merged with e7e95b9... Thanks again, @kwonoj!

Issue #350, opened 2015-09-18
Add micro perf tests for existing operators


blesh:
From #100 

- [x] buffer (buffer(obs))
- [ ] bufferTime (bufferWithTime)
- [ ] bufferToggle (buffer(obs, fn))
- [ ] bufferWhen (buffer(fn))
- [ ] delay
- [x] forkJoin
- [x] mapTo (map(obj))
- [ ] multicast
- [ ] publish
- [x] retry
- [ ] retryWhen
- [x] skipUntil
- [ ] subscribeOn
- [ ] switchMap
- [ ] takeUntil
- [x] toPromise
- [ ] window (window(obs))
- [ ] windowTime (windowWithTime)
- [ ] windowToggle (window(obs, fn))
- [ ] windowWhen (window(fn))
- [ ] withLatestFrom
- [ ] zipAll

mattpodwysocki:
@blesh maybe I'm naive here, but why is `mapTo` required when `map` for example in RxJS supports wrapping objects that are passed in.

blesh:
> @blesh maybe I'm naive here, but why is mapTo required when map for example in RxJS supports wrapping objects that are passed in.

reducing type checking in a really, really commonly used operator as well as making it less polymorphic.

blesh:
I think for the most part we have the metrics we need here. Anything missing can be added ad hoc. Closing this issue.

PR #359, opened 2015-09-20
chore(test): add micro-benchmark performance test for toPromise operator

kwonoj:
Try to cover https://github.com/ReactiveX/RxJS/issues/350.

One of test system shows slower numbers such as

> old toPromise() with immediate scheduler x 351,507 ops/sec ±4.64% (81 runs sampled)
> new toPromise() with immediate scheduler x 322,127 ops/sec ±9.00% (65 runs sampled)
>        -8.36% slower than Rx2
>
> old toPromise(ctor) with immediate scheduler x 351,730 ops/sec ±5.72% (78 runs sampled)
> new toPromise(ctor) with immediate scheduler x 327,025 ops/sec ±9.27% (55 runs sampled)
>        -7.02% slower than Rx2

Not all of system shows slower numbers, bit unsure if test case design is correct. Would be appreciate to let me know if something doesn't look right - will update accordingly.

### DIFF
--- a/perf/micro/immediate-scheduler/operators/topromise-ctor.js
+++ b/perf/micro/immediate-scheduler/operators/topromise-ctor.js
@@ -1,0 +1,18 @@
+var RxOld = require("rx");
+var RxNew = require("../../../../index");
+
+module.exports = function (suite) {
+
+  var oldToPromiseWithImmediateScheduler = RxOld.Observable.of(25, RxOld.Scheduler.immediate).toPromise(Promise);
+  var newToPromiseWithImmediateScheduler = RxNew.Observable.of(25).toPromise(Promise);
+
+  return suite
+    .add('old toPromise(ctor) with immediate scheduler', function () {
+      oldToPromiseWithImmediateScheduler.then(_then);
+    })
+    .add('new toPromise(ctor) with immediate scheduler', function () {
+      newToPromiseWithImmediateScheduler.then(_then);
+    });
+  
+  function _then(x) { }
+};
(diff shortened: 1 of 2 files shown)


blesh:
@kwonoj  it makes sense this one would be close, and perhaps a little slower. It looks like I'm introducing a little more closure and another assertion. I think we can trim the fat here.

kwonoj:
Reason I was suspicious about my test case design was test results was vary and only some of system shows slower results than original. I was able to see some of additional checks as you mentioned, bit unsure about results is not consistent over system configuration. Still it's interesting topic to track down further to see where does difference comes.

blesh:
Another thing to pay attention to, is that % accuracy at the end of the test.

> old toPromise() with immediate scheduler x 351,507 ops/sec **±4.64%** (81 runs sampled)
> new toPromise() with immediate scheduler x 322,127 ops/sec **±9.00%** (65 runs sampled)
-8.36% slower than Rx2

The "old" toPromise test there could be off by up to around 16,000 ops per sec, and the "new" toPromise could be off by up to almost 29,000 ops per second.

When I ran it, RxJS Next was "2% faster" and I had slightly higher accuracy.  I suspect these two are just going to be on par, because the design of Promise pretty much means you need to introduce some closure.

blesh:
> old toPromise() with immediate scheduler x 298,514 ops/sec ±4.15% (82 runs sampled)
> new toPromise() with immediate scheduler x 306,881 ops/sec ±3.29% (83 runs sampled)
> 	2.8% faster than Rx2

> old toPromise() with immediate scheduler x 309,282 ops/sec ±4.89% (77 runs sampled)
> new toPromise() with immediate scheduler x 330,398 ops/sec ±3.46% (83 runs sampled)
> 	6.83% faster than Rx2

A few runs on my machine and I'm unable to get to to lean the other way.

blesh:
merged with 5e8be38

Thanks, @kwonoj!!

PR #360, opened 2015-09-20
chore(test): add micro-benchmark performance test for Observable.timer

kwonoj:

Add small number of duetime and skipped duration to avoid test execution time increase.


blesh:
Great! merged with b9ab56e... thanks @kwonoj

PR #361, opened 2015-09-20
chore(test): use test marbles for Observable.timer spec

kwonoj:
Continue effort to cover https://github.com/ReactiveX/RxJS/issues/309.


blesh:
merged with 2888dda.... Thanks again!

PR #363, opened 2015-09-21

kwonoj:
…EventPattern

Try to cover #352.
Used EventEmitter as source then limit to single element as same as test for `fromEvent`.


blesh:
merged with 9ed7844... thanks!

PR #364, opened 2015-09-21
chore(test): fix README on running perf benchmarks

staltz:
No more recommendation of having globally installed packages.
Add `webdriver-manager update` execution prior to running protractor.


package.json
```diff
@@ -13,7 +13,7 @@
     "build_test": "rm -rf dist/ && npm run build_es6 && npm run build_cjs && jasmine",
     "build_docs": "./docgen.sh",
     "test": "jasmine",
-    "perf": "protractor protractor.conf.js",
+    "perf": "webdriver-manager update && protractor protractor.conf.js",
```
kwonoj: What about put `webdriver-manager update` into `build_perf` instead? that'll allow to run perf test without calling it each time.

staltz: Yup, fixed.

blesh:
merged as of 6e5256d, thank you @staltz

PR #367, opened 2015-09-22
chore(test): add micro-benchmark performance test for Observable.inte…

kwonoj:
…rval

Try to cover #352.

Due to nature of `interval` limited elements number relatively small to reduce general test execution time.


blesh:
Merged with 3a7d6b0 ... thanks @kwonoj

Issue #366, opened 2015-09-21
Fresh install unrunnable


pselle:
I tried updating my upstream and was unable to run any build commands. When checking out a fresh repo, I got the same errors:

```
$ npm install
npm ERR! install Couldn't read dependencies
npm ERR! Darwin 14.5.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "install"
npm ERR! node v4.1.0
npm ERR! npm  v2.14.3

npm ERR! Invalid name: "@reactivex/rxjs"
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/pselle200/Projects/freshRx/npm-debug.log
```

Looks like `@reactivex/rxjs` is an invalid name. How are you all resolving this as you're working on it? I also updated my node, but that didn't solve the issue.

blesh:
It looks like you're running Node v 4? What version of NPM are you using?

`@reactivex/rxjs` should be a valid name, unless NPM changed something... 

cc/ @jeffbcross @igorminar ... have you seen anything like this?

blesh:
FWIW: I've only been using Node v 0.12.0 for development... I'll try 4 when I get a chance, but it might be a while.

staltz:
Come on, Ben, it's just a matter of doing `nvm install v4.1.0`.

I tried reproducing this by deleting node_modules and running `npm install` with node v4.1.0 and npm v2.14.3 and everything worked as it should (my git HEAD is on upstream master right now). I don't think it's a problem of npm not recognizing scoped packages.

pselle:
I had the issue _before_ I upgraded node as well. Removing the `@` and `/` from the name allow `npm install` to run.. until it puked without  babel.
```
sh: babel: command not found

npm ERR! Darwin 14.5.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "run" "build_cjs"
npm ERR! node v4.1.0
npm ERR! npm  v2.14.3
npm ERR! file sh
npm ERR! code ELIFECYCLE
npm ERR! errno ENOENT
npm ERR! syscall spawn
npm ERR! reactivexrxjs@0.0.0-prealpha.3 build_cjs: `rm -rf dist/cjs && babel dist/es6 --out-dir dist/cjs --modules common --sourceMaps --loose all`
[...]

PR #368, opened 2015-09-22
fix(package): specify supported npm version

jeffbcross:


blesh:
@jeffbcross ... is this to address @pselle's issue #366, I assume?

blesh:
merged with f72e622. Thanks @jeffbcross

Issue #346, opened 2015-09-18
Notification `create` methods could return consts in some cases


blesh:
In #206, @abersnaze pointed out that we could save on object allocations in `materialize` and `dematerialize` (as well as our TestScheduler) by returning the same references, or perhaps tokens of some sort (to further reduce object allocations), for common return values like `Notification.createComplete()` and `Notification.createNext(null)`, etc.

kwonoj:
Interested in creating PR for this - which way would be feasible? Initial thought I could think some ways like
- create `NotificationLite` class as similar to RxJava does have
- allow additional parameter into existing interface like `createComplete(getStaticReference?:boolean)` to return specific object reference
- let existing interface automatically returns by value, i.e `createNext(value) { if (!value) { return existingReference }....}`

Or possibly other, better way. Any suggestion would be appreciated.

blesh:
I think our current implementation is already pretty "lite"... but we can probably just use a static reference for all `complete` notifications, as well as `undefined` value notifications.

kwonoj:
Got it, will try to create to return static reference per specific cases as similar to 3rd way's pseudo does.

PR #371, opened 2015-09-23
refactor(Notification): return static reference notification for comm…

kwonoj:
…on value

Try to close https://github.com/ReactiveX/RxJS/issues/346.

Take simplest approach to return static reference for 2 specific cases of `createComplete()` and  `createNext(undefined)`.

### DIFF
--- a/src/Notification.ts
+++ b/src/Notification.ts
@@ -53,8 +53,14 @@ export default class Notification<T> {
     }
   }
   
+  private static completeNotification: Notification<any> = new Notification('C');
+  private static undefinedValueNotification: Notification<any> = new Notification('N', undefined);
+  
   static createNext<T>(value: T): Notification<T> {
-    return new Notification('N', value);
+    if (typeof value !== 'undefined') {
+      return new Notification('N', value);
+    }
+    return this.undefinedValueNotification;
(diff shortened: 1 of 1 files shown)


blesh:
This looks fine. I debated a little about whether or not the static properties should be `public` or `private` (for comparison reasons) or whether they should just be `const`s in the module file. ... but I think we can cross that bridge when it becomes necessary and this will work for now.

Where you able to create any performance tests that proved this was more efficient? I don't see how it couldn't be; After all, fewer allocations are fewer allocations, but I was curious if you tested this change at all.

I think it's good to merge as is. Just curious.

blesh:
merged with 3b58521, thanks @kwonoj!

kwonoj:
> whether or not the static properties should be public or private
: for this, I'd choose to encapsulate as much first until it's explicitly required to be accessed. 

> Where you able to create any performance tests that proved this was more efficient? I don't see how it couldn't be.. 
: actually this was one of my interests too, but wasn't able to come up with feasible test cases to be included in this PR. So basically, this changes are written without concrete number based decisions. I'm trying to explore some of possibilities if possible, will create PR if that's available.

PR #372, opened 2015-09-23
chore(test): use test marbles for reduce operator spec

kwonoj:
Continue effort to cover https://github.com/ReactiveX/RxJS/issues/309.


blesh:
Merged with fef727c... thanks again!

PR #393, opened 2015-09-24
feat(operator): add find, findIndex operator

kwonoj:
- add find, findIndex operator
- add SingleValueSubscriber for common behavior to notify and complete
immediately

###### `find`, `findIndex` operator is added.

In document of [`findIndex`](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/findindex.md), return value is `undefined` if there isn't matching element
>... defined by the specified predicate, if found; otherwise, undefined.

while example, and actual code returns `-1` instead. This PR follows existing code behaviors to return `-1`.

###### add `SingleValueSubscriber` as base of `FindSubscriber` and others

One of common pattern of subscriber is let subscriber notify and complete immediately, usually related with operators using predicate like `find`, `findIndex`, `isEmpty`, `single`, `last`, and others not yet implemented. `SingleValueSubscriber` exposes `notifyComplete(value)` interface for those behavior to reduce duplication of code.

1. Performance implication? : Micro perf test for operators having inherited subscriber does not shows noticeable performance hits.
2. behavior of `notifyComplete` is quite short and small, is it really required? : Thinking about numbers of operators having similar behavior, I think it still makes sense to have common base. It still brings concern of inheritance depth usually want to avoid in general.

### DIFF
--- a/perf/micro/immediate-scheduler/operators/find-predicate-this.js
+++ b/perf/micro/immediate-scheduler/operators/find-predicate-this.js
@@ -1,0 +1,26 @@
+var RxOld = require("rx");
+var RxNew = require("../../../../index");
+
+module.exports = function (suite) {
+
+    var predicate = function(value, i) {
+      return value === 20;
+    };
+    
+    var testThis = {};
+
+    var oldFindPredicateThisArg = RxOld.Observable.range(0, 50, RxOld.Scheduler.immediate).find(predicate, testThis);
+    var newFindPredicateThisArg = RxNew.Observable.range(0, 50).find(predicate, testThis);
+    
+    return suite
+        .add('old find(predicate, thisArg) with immediate scheduler', function () {
+            oldFindPredicateThisArg.subscribe(_next, _error, _complete);
+        })
+        .add('new find(predicate, thisArg) with immediate scheduler', function () {
+            newFindPredicateThisArg.subscribe(_next, _error, _complete);
+        });
+
+    function _next(x) { }
+    function _error(e){ }
+    function _complete(){ }
+};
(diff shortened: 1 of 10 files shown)


blesh:
> while example, and actual code returns -1 instead. This PR follows existing code behaviors to return -1.

This is the correct behavior.

blesh:
Actually, @kwonoj ... I'm going to halt this PR for some bikeshedding. Nevermind my previous comments.

blesh:
Damn you're too fast, @kwonoj ... I was trying to delete my previous comments. I'd stop work on this for right now... I'm suspicious that these operators might not have a lot of value to the library.

I want to create an issue to discuss it.

kwonoj:
I saw issue is created. I'll followup issue and let's close issue based on conclusion.

Thanks!

blesh:
@kwonoj ... can you add these operators in the `extended` directory, and to the KitchenSink output only? (see `elementAt`).

kwonoj:
Sure, will do. https://github.com/ReactiveX/RxJS/pull/404 this should be merged first, is my understanding correct?

blesh:
@kwonoj that's correct... sorry, I thought I'd merged it already. It's merged now.

blesh:
Also @kwonoj, you may want to rebase now that I'm done merging PRs for the morning. There's an ugly merge in there from the CoreOperators/KitchenSink refactor

kwonoj:
* `find`, `findIndex` now placed under `extended`
* Subscriber inheritance removed

kwonoj:
* updated module import paths for test case 
* removed redundant changes to PR

blesh:
Thanks for your hard work, @kwonoj. Much appreciated!

PR #395, opened 2015-09-24
feat(operator): add ignoreElements operator

jinroh:
Add `ignoreElements` operator present in RxJS3/4.


blesh:
merged with fe1a952

Thanks so much, @jinroh!

Issue #378, opened 2015-09-23
comprehensive marble tests for windowCount


blesh:
Should duplicate the current RxJS 3/4 tests

Should adhere to our contribution guidelines and operator authoring docs.

staltz:
Should not yet be closed. Even though it has received some PRs, there are a few edge cases missing to be tested.

staltz:
I can do this.

staltz:
Please close

PR #398, opened 2015-09-24
chore(test): use test marbles for windowCount operator spec

kwonoj:
Trying to cover https://github.com/ReactiveX/RxJS/issues/309, also trying to cover https://github.com/ReactiveX/RxJS/issues/378 as well by migration of existing test cases. 

Will try to create PR to contain both (replacing current, migrate from RxJS) starting from this changes.

### DIFF
--- a/spec/operators/windowCount-spec.js
+++ b/spec/operators/windowCount-spec.js
@@ -1,35 +1,31 @@
-/* globals describe, it, expect */
+/* globals describe, it, expect, expectObservable, hot */
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
 describe('Observable.prototype.windowCount', function () {
-  it('should emit windows at intervals', function (done) {
-    var expected = [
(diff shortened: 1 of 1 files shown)


kwonoj:
Actually one [test case](https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/windowwithcount.js#L73) from current RxJS is not covered in this PR, where source is being disposed. Any suggestions for representing those case?

blesh:
@kwonoj that's a good one. I was talking with @staltz recently about ways to improve the marble tests so they cover things like inner unsubscriptions and outer unsubscribes.

I think this PR is good for now, but we definitely need to come up with some solutions for the unsubscription tracking.

blesh:
... and there are a lot of test cases we haven't covered for this operator anyway. We'll get there.

blesh:
Merged with 84e3684

Thanks as always, @kwonoj!

PR #399, opened 2015-09-24
feat(zip): zip now supports never-ending iterables

### LINKED ISSUE #397: Add special handling for Iterables passed to `zip` operators
So the idea is this (I'll use `Observable.zip` as an example):

Assume:
```
Observable.zip(observable1, observable2, iterable);
```

The `iterable` will be treated differently, in that instead of being converted to an Observable and firehosing out nexts, it will get an `iterator` and `next()` once, waiting for the other Observables to emit.. as soon as the `zip` emits, the `iterator` will `next()` once more, and so on.

Each time the `iterator.next()` is called, the `IteratorResult` will be examined to see if it's `done`, if it's `done`, it will flag itself as no longer `active` and the Observable will end

blesh:
- adds support for never-ending iterables
- adds fast-track for arrays
- adds additional tests for zip
- fixes issues where completion does not happen at the appropriate moment

closes #397

### DIFF
--- a/spec/observables/zip-spec.js
+++ b/spec/observables/zip-spec.js
@@ -1,4 +1,4 @@
-/* globals describe, it, expect */
+/* globals describe, it, expect, expectObservable, hot, cold */
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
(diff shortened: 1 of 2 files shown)


blesh:
The controversial thing here is that it reduces performance slightly from what is in master...

from 98,000 ops/sec
to 78,000 ops/sec

Still substantially faster than the older versions of RxJS.

**However**... this does include fixes to how the returned observable completes. It's hard to compare apples to apples until those fixes are in what's in master. So that will be next.

blesh:
With (one possible set of) fixes to `zip` in #400, the perf drops anyway to 89,000...

So really the comparison for this feature is likely:  89,000 vs 78,000 ops/sec.

The upside is it enables zipping with an "infinite" resource without locking a thread or creating a huge buffer.

blesh:
attn @trxcllnt, @mattpodwysocki, @benjchristensen

blesh:
@steveorsomethin ... since you're more in-tune with TVUI/Netflix perf needs for these operators, do you view the perf differences mentioned above as okay vs the value of this PR?

One reason I feel like this is okay is it makes zip *slightly* less dangerous.

blesh:
Per a conversation with @steveorsomethin and @ktrott: The small change in performance isn't enough to outway the benefit of the new functionality and shouldn't effect anything in the lower-end hardware or non-JIT'ed runtimes Netflix needs to support.

Merged.

PR #400, opened 2015-09-24
fix(zip): ensure zip observable completes properly

blesh:
- adds additional tests
- outerIndex is now public on InnerSubscriber
- updates zip to complete when any inner observable has completed and its buffer is empty

relates to #397


blesh:
This is an alternative PR to #399 

Ops per second here are ~89,000 ... discussion is in #399.

blesh:
Closing this in favor of #399.

PR #404, opened 2015-09-25
refactor(Rx): split core from extended operators

### LINKED ISSUE #401: API surface and packaging strategy
Need a strategy that improves on management of the API / operator surface and packaging.

Today, it is often the case people try to use lite but it has more than they need but missing an operator they need. Seems like we want people to build their own Rx module of the operators they want.

We would still define the base Rx that has the core set of operators. Extended operators would be allowed in an extended area.

If you want to use more specialized / custom operators then this could be the pattern...

```js
const Observable = Rx.Observable.withOperators({
    elementAt: require('rx/extras/elementAt'),
    toPromise: require('rx/extras/toPromise'),
});
```

The above would return a sub-classes Observable that would mix those properties in.

We'd also have an `rx-kitchensink.ts` module

Consumers could also create their own Rx module with exactly what they want.

blesh:
- moves non-core operators to entended directory
- adds Rx.KitchenSink module
- removes non-core operators from Rx module
- removes TestScheduler from Rx module
- removes VirtualTimeScheduler from Rx module
- updates tests appropriately
- adds Rx.KitchenSink to es6 build process
- reorganizes operators as they appear in files to be in alphabetical order and separated by static and members

closes #401

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "build_all": "npm run build_es6 && npm run build_amd && npm run build_cjs && npm run build_global",
     "build_amd": "rm -rf dist/amd && tsc typings/es6-shim/es6-shim.d.ts src/Rx.ts -m amd --outDir dist/amd --sourcemap --target ES5",
     "build_cjs": "rm -rf dist/cjs && babel dist/es6 --out-dir dist/cjs --modules common --sourceMaps --loose all",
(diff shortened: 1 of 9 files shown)


blesh:
This also cleans up the `Observable.ts` file by introducing an interface `CoreOperators`

PR #405, opened 2015-09-25
chore(test): format test failure output more aligned

kwonoj:
Built in `JSON.toStringify()` does not serialize `error` object, so test output was not able to format exception correctly. This PR overrides default behavior while running tests. 

*before*
before

*after*
after


blesh:
Merged with d40d73b... thanks @kwonoj

PR #406, opened 2015-09-25
test(count): add tests for count

adrianomelo:


blesh:
Merged with 8c9d67c, thanks @adrianomelo! Much appreciated.

PR #407, opened 2015-09-25
test(zip): add marble tests for zip

adrianomelo:
- add missing tests
- refactor tests to use marble tests


blesh:
Merged with 5a00700... thanks again @adrianomelo

PR #408, opened 2015-09-26
chore(test): use test marbles for bufferCount operator spec

kwonoj:
Trying to cover #309, also trying to cover https://github.com/ReactiveX/RxJS/issues/378 as well by migration of existing test cases.


blesh:
Merged with d80d86a, thanks as always, @kwonoj

PR #409, opened 2015-09-27
chore(test): use test marbles for sample, sampleTime operator spec

kwonoj:
Trying to cover #309, migrated existing RxJS3 test cases.


blesh:
Merged as of 060287a ... I've added an additional commit to cover two additional situations:

1. Where a hot observable may have emitted prior to subscription
2. Where the source observable has NOT emitted by the time of the sample

blesh:
Thanks again @kwonoj

PR #410, opened 2015-09-27
chore(package): Add 'typings' field to package.json

saneyuki:
This supports TypeScript 1.6 later's [`moduleResolution=node` option](https://github.com/Microsoft/TypeScript/wiki/Compiler-Options).

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,5 +82,6 @@
   },
   "engines": {
     "npm": "~2.0.0"
-  }
+  },
+  "typings": "./dist/es6/Rx.d.ts"
 }


blesh:
I'm unfamiliar with this feature of TypeScript/TSC... Does this fix an issue we were having?

saneyuki:
Without this change, we need to write `///<reference path='....../node_modules/@reactivex/rxjs/dist/es6/Rx.d.ts'/>` to the file header if we use this in TypeScript. I feel it is verbose.

By this option, we don't need to write `///<reference path='.....'/>` when we use this package with TypeScript 1.6 & add `moduleResolution='node'` option (see [more details](http://blogs.msdn.com/b/typescript/archive/2015/09/16/announcing-typescript-1-6.aspx)). This is very useful in NodeJS environment with TypeScript or browserify/webpack env.

However, I feel this has some this advantage. It's that we need to maintain this field if we'll change the distribution path in the future.

blesh:
That disadvantage is probably fine.

LGTM.... merging in a few.

blesh:
Merged with 8318089

Thank you @saneyuki

saneyuki:
Thak you @blesh !

saneyuki:
> However, I feel this has some this advantage.

oops. s/this advantage/disadvantage/

PR #412, opened 2015-09-28
chore(test): add micro-benchmark performance test for delay operator

kwonoj:
Try to cover https://github.com/ReactiveX/RxJS/issues/350


blesh:
Merged with e4c287e ... thank you @kwonoj

PR #413, opened 2015-09-29
fix(concat): let observable concat instead of merge

kwonoj:
- change behavior of concat instead of merge
- migrate test case for concat, concatall

`concat` was working as same as `merge`, changed its behavior to actually concat observables also migrated all test cases from RxJS3. Operators like `concatall` might need additional test coverage in further. 

*reproducible code*
```
RxNew.Observable.interval(100).take(5).map(function(x) { return x + ':  source'})
  .concat(RxNew.Observable.interval(200).take(3).map(function(x) { return x + ':  other'}))
  .subscribe(function(x) { 
      console.log(x); 
    }, function(x) { 
      console.log(x);
    },function() { 
      console.log('concat completed'); 
    });
```

*before fix*
>0:  source
>0:  other
>1:  source
>2:  source
>1:  other
>3:  source
>4:  source
>2:  other
>concat completed

*after fix*
>0:  source
>1:  source
>2:  source
>3:  source
>4:  source
>0:  other
>1:  other
>2:  other
>concat completed

### DIFF
--- a/spec/observables/concat-spec.js
+++ b/spec/observables/concat-spec.js
@@ -1,0 +1,14 @@
+/* globals describe, it, expect, expectObservable, cold */
+var Rx = require('../../dist/cjs/Rx');
+var Observable = Rx.Observable;
+
+describe('Observable.concat', function () {
+  it('should emit elements from multiple sources', function() {
+    var e1 =  cold('-a-b-c-|');
+    var e2 =  cold('-0-1-|');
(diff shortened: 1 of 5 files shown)


blesh:
Merged with 8fe9e39... I've added a commit to account for the `mergeAll` operators being on the CoreOperators interface now.

blesh:
Thank you @kwonoj

PR #416, opened 2015-09-29
refactor(test): modify module import path for test cases

kwonoj:
* let index.js used for perf test to import `KitchenSink` by default
* modify import path for `elementAt` test case


blesh:
Merged with d5403af

Thanks, @kwonoj

PR #420, opened 2015-09-29
Add resultSelector argument to both `first` and `last` operators

blesh:
Just what it says. This was primarily for

1. to enable a user to use `first` to select an `index` if they so choose, effectively replacing `findIndex` as any sort of need in CoreOperators.
2. Consistency, so that `first` and `last` have identical signatures...

To be bikeshedded: 

- the usefulness of the `thisArg`
- argument order

### DIFF
--- a/spec/operators/first-spec.js
+++ b/spec/operators/first-spec.js
@@ -23,7 +23,7 @@ describe('Observable.prototype.first()', function() {
   it('should return the default value if source observable was empty', function() {
     var e1 = hot('-----^----|');
     var expected =    '-----(a|)';
-    expectObservable(e1.first(null, null, 'a')).toBe(expected);
+    expectObservable(e1.first(null, null, null, 'a')).toBe(expected);
   });
 
   it('should propagate error from the source observable', function() {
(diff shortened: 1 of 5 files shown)


staltz:
In argument order, I see defaultValue being more common than thisArg or resultSelector.

blesh:
I think `thisArg` could be removed entirely, myself. I've never been a fan of `thisArg` parameters.

staltz:
Summoning @mattpodwysocki for this

blesh:
So I've ended up just removing the `thisArg`. Since `.bind()` is available in every browser we're targeting, I don't see the need to have the `thisArg`, when it seems to junk up the API.

However, @staltz, I didn't put the `defaultValue` first because it made using the resultSelector without the default argument weird... `obs.first(predicate, undefined, resultSelector)`, you can't do `null`, because someone might *want* a `defaultValue` of `null`... so the order is `obs.first(predicate, resultSelector, defaultValue)`, and if you want a default value with no result selector it's `obs.first(predicate, null, 'hi')` I agree it's not *as* pretty, but I think the odds are so high that users are going to do a `map` right after they do a `first` or `last`, that the resultSelector will be a perf win for those cases.

blesh:
FWIW, I did experiment with having the order be what you suggested, @staltz.

blesh:
I think we can merge this for now and let users try it out and see how it feels. As long as we settle on a final API by Beta.

staltz:
Looks good to me. If `defaultValue` first didn't work, then your solution has to be the only way. I'd merge

Issue #411, opened 2015-09-28
timeout(), timeoutWith() behavior differ to RxJS3?


kwonoj:
Could observe different behavior while try to migrate test cases, would like to confirm if this is expected.

##### timeout()

*RxNew*
```
RxNew.Observable.interval(100).take(5).map(function(x) { return x + ': new timeout source'})
  .concat(RxNew.Observable.interval(200).take(3))
  .timeout(150, new Error('Timeout has occurred.'))
  .subscribe(function(x) { 
      console.log(x); 
    }, function(x) { 
      console.log(x);
    },function() { 
      console.log('new timeout completed'); 
    });
```
>0: new timeout source
>[Error: Timeout has occurred.]

`dueTime` is total time passed after source starts.
> source : ----------0----------1----------2----------3----------4--------------------0'--------------------1'.....
> output : ---------------#

*RxOld*
```
Rx.Observable.interval(100).take(5).map(function(x) { return x + ': old timeout source'})
  .concat(Rx.Observable.interval(200).take(3))
  .timeout(150, new Error('Timeout has occurred.'))
  .subscribe(function(x) { 
      console.log(x); 
    }, function(x) { 
      console.log(x);
    },function() { 
      console.log('old timeout completed'); 
    });
```

>0: old timeout source
>1: old timeout source
>2: old timeout source
>3: old timeout source
>4: old timeout source
>[Error: Timeout has occurred.]

`dueTime` is time window between each source element emits
> source : ----------0----------1----------2----------3----------4--------------------0'--------------------1'.....
> output :  ----------0----------1----------2----------3----------4---------------#

##### timeoutWith()

*RxNew*
```
RxNew.Observable.interval(100).take(5).map(function(x) { return x + ': new timeoutWith source'})
  .concat(RxNew.Observable.interval(200).take(3))
  .timeoutWith(150, 
    RxNew.Observable.interval(100).take(3).map(function(x){return x+': new timeoutWith other'}))
  .subscribe(function(x) { 
      console.log(x); 
    }, function(x) { 
      console.log(x);
    },function() { 
      console.log('new timeoutWith completed'); 
    });
```

>0: new timeoutWith source
>0
>1: new timeoutWith source
>0: new timeoutWith other
>2: new timeoutWith source
>1: new timeoutWith other
>1
>3: new timeoutWith source
>2: new timeoutWith other
>new timeoutWith completed

source still emits after timeout?
[...]

PR #421, opened 2015-09-30
fix(timeout): update behavior of timeout, timeoutWith

kwonoj:
* modify behavior of `timeout`, `timeoutWith`
* migrate test cases from RxJS3

Try to close https://github.com/ReactiveX/RxJS/issues/411

### DIFF
--- a/spec/operators/timeout-spec.js
+++ b/spec/operators/timeout-spec.js
@@ -1,44 +1,53 @@
-/* globals describe, it, expect */
+/* globals describe, it, expect, expectObservable, hot, rxTestScheduler */
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
-describe('Observable.prototype.timeout', function () {
-  it('should timeout after a specified delay', function (done) {
-    Observable.never().timeout(100)
-      .subscribe(function (x) {
-        throw 'should not next';
-      }, function (err) {
-        expect(err.message).toBe('timeout');
-        done();
-      }, function () {
-        throw 'should not complete';
-      });
-  }, 2000);
+describe('Observable.prototype.timeout()', function () {
+  var defaultTimeoutError = new Error('timeout');
   
-  it('should timeout after a delay and send the passed error', function (done) {
-    Observable.never().timeout(100, 'hello')
-      .subscribe(function () {
-        throw 'should not next';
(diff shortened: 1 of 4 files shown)


spec/operators/timeout-spec.js
```diff
@@ -1,44 +1,102 @@
...
+    var expected = '-----#';
+    var value = 'hello';
+    
+    expectObservable(e1.timeout(50, value, rxTestScheduler)).toBe(expected, null, value);
+  });
+  
+  it('should timeout at a specified Date', function (done) {
+    var e1 = Observable.never();
+    
+    e1.timeout(new Date(Date.now() + 100), null)
```
kwonoj: Absolute timeout period can cause intermittent test failure due to time frame difference in test scheduler (usually 1/10 frame difference) between expected timeout to actual timeout, by doing calculation twice (to setup fixture value create `Date` via `Date.now() + n` then actual operator does `+due - Date.now()`). 
to avoid intermittent failure, test cases covers absolute timeout are not using exact frame comparison.

blesh: Did you try `rxTestScheduler.now()`?

kwonoj: Actually did not. will try quickly.

kwonoj: Gave a try, but not works by just replacing to testSchderuler's now(). This happens due to fixture setup using`Date`, 
1. `new Date(now + timeframe)` is delivered as absolute timeframe param
2. `timeout` calculate absolute time via `(+due - now)` where `due` is param above.

if use actual datetime it causes intermittent failure, and using testscheduler.now create 1 as negative timeframe since testscheduler.now returns 0 as starting timeframe.

I'd like to investigate this further as separate PR since it's test setup specific issue, may need few trial & errors.

blesh:
Okay, I think that we can merge this for now. Just remove any tests that are sketchy and we'll create a separate issue for handling absolute Dates in the TestScheduler.

blesh:
Let me know when you've removed the questionable tests, and I'll merge.

kwonoj:
Updated, removing test cases related absolute date having issues.

blesh:
merged with 16bd691

Thanks, @kwonoj

Issue #422, opened 2015-09-30
Merge semantics for race conditions between synchronous notifications 


staltz:
This is probably a bug.

While making tests for mergeMap I wrote this marble test based on a RxJS 4 test:

```js
  it('should mergeMap many regular interval inners', function () {
   var a =   cold('----a---a---a---(a|)'                    );
   var b =   cold(    '----b---b---(b|)'                    );
   var c =   cold(                '----c---c---c---c---(c|)');
   var d =   cold(                        '----(d|)'        );
   var e1 =   hot('a---b-----------c-------d-------|'       );
   var expected = '----a---(ba)(ba)(ba)c---c---(dc)c---(c|)';
  
   var observableLookup = { a: a, b: b, c: c, d: d };
   var source = e1.mergeMap(function (value) {
     return observableLookup[value];
   });
  
   expectObservable(source).toBe(expected);
  });
```

Notice 

`var expected = '----a---(ba)(ba)(ba)c---c---(dc)c---(c|)'`

This test passes for RxJS 4, but does not pass for RxJS Next, which behaves like this

`var expected = '----a---(ba)(ab)(ab)c---c---(cd)c---(c|)'`

I quickly took a look and tried to solve this for mergeMap, but the problem seemed deeper, related to `subscribeToResult`, `InnerSubscriber`, and potentially from `TestScheduler` and `VirtualTimeScheduler` too. Might affect many operators, so was beyond the scope of my mergeMap PR #423.

blesh:
I find it puzzling that the order would be `(ba)(ba)(ba)` I would expect `(ab)(ab)(ab)`. Although I think `(ba)(ab)(ab)` is even weirder. That `(ba)` at the beginning is the troubling part.

blesh:
Also, with the TestScheduler, there shouldn't be any "race conditions".

staltz:
Yes. Both existing RxJS 4 and RxJS Next behaviors are puzzling, I'd also expect `(ab)(ab)(ab)` for both. See here https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/selectmany.js#L926-L939

trxcllnt:
@blesh @staltz I have a hunch this is related to the switch from the default trampoline-style scheduling in RxJS4 to the default recursive-style scheduling in RxJS Next. `(ba)(ab)(ab)` feels like the pattern I'd expect from such a switch.

[...]

Issue #374, opened 2015-09-23
Comprehensive marble tests for mergeMap


blesh:
Should duplicate the current RxJS 3/4 tests

Should adhere to our contribution guidelines and operator authoring docs.

staltz:
I assign myself to this.

blesh:
@staltz ... since I can't "assign" you, I'll put myself on it and follow up.

staltz:
All tests for mergeMap are done in PR #423, except for one case: #422

staltz:
I would close this issue because we already have a separate issue for the problematic test, #422.

PR #423, opened 2015-09-30
Marble tests for mergeMap and unsubscription marbles for TestScheduler

### LINKED ISSUE #402: Unsubscribe from marble tests
Can't do tests with the marble tests to show if someone wants to unsubscribe at some point: explicit disposable vs. completion. Need some way to tell the test scheduler to unsubscribe from the test observable at a particular frame.

staltz:
This PR resolves #402 and #374.
All tests from RxJS 4 are present in this PR, except for one test: see #422.

### DIFF
--- a/src/operators/mergeMap-support.ts
+++ b/src/operators/mergeMap-support.ts
@@ -11,12 +11,14 @@ import InnerSubscriber from '../InnerSubscriber';
 
 export class MergeMapOperator<T, R, R2> implements Operator<T, R> {
   constructor(private project: (value: T, index: number) => Observable<R>,
-    private resultSelector?: (innerValue: R, outerValue: T, innerIndex: number, outerIndex: number) => R2,
-    private concurrent: number = Number.POSITIVE_INFINITY) {
+              private resultSelector?: (innerValue: R, outerValue: T, innerIndex: number, outerIndex: number) => R2,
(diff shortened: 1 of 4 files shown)


src/operators/mergeMap-support.ts
```diff
@@ -49,36 +51,39 @@ export class MergeMapSubscriber<T, R, R2> extends OuterSubscriber<T, R> {
...
+      if (result === errorObject) {
         destination.error(errorObject.e);
       } else {
         destination.next(result);
       }
+    } else {
+      destination.next(innerValue);
     }
-    destination.next(innerValue);
   }
```
staltz: This above wasn't just a refactor, I recall solving one of the tests failing by including this `else`, because otherwise there would be two calls to `destination.next()`.

blesh:
Overall this looks good.

One thing that is interesting, is it doesn't cover the expectations of unsubscription does it? It only covers the unsubscribe.

I say that's interesting, because originally my concern was the inability to examine when an inner observable was unsubscribed _from_, in say, switch or the like. So I guess we still have an unresolved issue.

blesh:
I'm going to merge this one and create a new issue.

staltz:
Yeah that crossed my mind. Gotta think about it.

blesh:
merged with 4d973a3

PR #427, opened 2015-09-30
test(skipUntil): add micro-benchmark performance test for skipUntil operator

kwonoj:
Trying to cover https://github.com/ReactiveX/RxJS/issues/350.


blesh:
from now on, can we label test-related commits with `test(skipUntil): yadda`? I just noticed the conventional changelog generator was omitting ones that started with `chore()`

kwonoj:
Sure, shall I change this commit's label also?

kwonoj:
Actually, just did and update PR with suggested commit message.

blesh:
merged with 7ce1486

thanks, @kwonoj!

PR #429, opened 2015-10-01
test(zip): migrating tests from RxJS 4

adrianomelo:
Add tests:
- should work with empty and non-empty
- should work with non-empty and empty
- should work with never and non-empty
- should work with non-empty and never
- should combine two observables and selector
- should work with empty and error
- should work with error and empty
- should work with never and error
- should work with error and never
- should work with error and error
- should work with some error
- should work with error some
- should work with n-ary symmetric selector
- should work with n-ary array symmetric
- should work with n-ary symmetric array selector
- should work with some data asymmetric 1
- should work with some data asymmetric 2
- should work with some data symmetric
- should work with selector throws
- should work with right completes first
- should work with with iterable never empty
- should work with with iterable empty empty
- should work with with iterable empty non-empty
- should work with with iterable non-empty empty
- should work with with iterable never non-empty
- should work with with iterable non-empty non-empty
- should work with with iterable error empty
- should work with with iterable error some
- should work with with iterable some data both sides
- should work with with iterable selector throws

### DIFF
--- a/spec/operators/zip-spec.js
+++ b/spec/operators/zip-spec.js
@@ -28,12 +28,43 @@ describe("zip", function () {
     var expected = '|';
     expectObservable(a.zip(b)).toBe(expected);
   });
+  it('should work with empty and non-empty', function () {
+    var a = Observable.empty();
+    var b = hot('---1--|');
+    var expected = '|';
+    expectObservable(a.zip(b)).toBe(expected);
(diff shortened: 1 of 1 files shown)


blesh:
Awesome! Thanks, @adrianomelo! Merged with ef3cbec

PR #430, opened 2015-10-01
test(skip): add marble tests for skip operator

kwonoj:
Migrating existing RxJS4 test cases.

### DIFF
--- a/spec/operators/skip-spec.js
+++ b/spec/operators/skip-spec.js
@@ -10,10 +10,88 @@ describe('Observable.prototype.skip()', function () {
     expectObservable(source.skip(3)).toBe(expected);
   });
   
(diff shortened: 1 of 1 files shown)


blesh:
Do you think can can add tests for `empty`, `never` and `throw`? (Throw being different than error tests, because an error test can emit a value first)

blesh:
I'm actually shocked the RxJS 4 tests didn't have that.

kwonoj:
Yes that's true.. didn't noticed it. Will update PR as suggested.

kwonoj:
Test cases are added for explicit `throw`, `empty`, and some variant of `never` cases.

blesh:
Merged as of 954a873

Thanks again, @kwonoj!!!

PR #431, opened 2015-10-01
refactor(KitchenSink): add first, ignoreElements to interface

kwonoj:
Add two missing interfaces from `CoreOperator` to `KitchenSink`.


kwonoj:
It'd be better to remove duplication from `KitchenSink` and let `KitchenSink` re-export from `Rx` directly in further.

blesh:
@kwonoj I agree. I think the problem I hit with that is that I didn't want to mutate the `prototype` of the `Rx` Observable... so I was forced to duplicate code until a better solution is found.

blesh:
Merged with 8342430!

Thanks, @kwonoj

PR #432, opened 2015-10-01
feat(operator): add every operator

kwonoj:
adding `every` operator.

Micro perf test shows noticeable slowness on new operator.

>old every(predicate) with immediate scheduler x 104,692 ops/sec ±1.20% (95 runs sampled)
>new every(predicate) with immediate scheduler x 87,236 ops/sec ±1.24% (90 runs sampled)
>        -16.67% slower than Rx2

>old every(predicate, thisArg) with immediate scheduler x 94,638 ops/sec ±1.37% (96 runs sampled)
>new every(predicate, thisArg) with immediate scheduler x 84,714 ops/sec ±0.53% (96 runs sampled)
>        -10.49% slower than Rx2

investigating root cause but not yet able to find it, created PR to share implementation to address mistake / errors that I'm missing.

### DIFF
--- a/spec/operators/every-spec.js
+++ b/spec/operators/every-spec.js
@@ -1,0 +1,96 @@
+/* globals describe, it, expect, expectObservable, hot */
+var Rx = require('../../dist/cjs/Rx');
+
+describe('Observable.prototype.every()', function() {
+  function truePredicate(x) {
+    return true;
+  }
+  
+  function predicate(x) {
+      return x % 5 === 0;
+  }
+  
+  it('should emit true if source is empty', function() {
+    var source = hot('-----|');
+    var expected =   '-----(x|)';
+    
+    expectObservable(source.every(predicate)).toBe(expected, {x: true});
+  });
+  
+  it('should emit false if single source of element does not match with predicate', function() {
+    var source = hot('--a--|');
+    var expected =   '--(x|)';
+    
+    expectObservable(source.every(predicate)).toBe(expected, {x: false});
+  });
+  
+  it('should emit false if none of element does not match with predicate', function() {
+    var source = hot('--a--b--c--d--e--|');
+    var expected =   '--(x|)';
+    
+    expectObservable(source.every(predicate)).toBe(expected, {x: false});
+  });
+  
+  it('should return false if only some of element matches with predicate', function() {
+    var source = hot('--a--b--c--d--e--|', {a: 5, b: 10, c: 15});
+    var expected =   '-----------(x|)';
+    
+    expectObservable(source.every(predicate)).toBe(expected, {x: false});
+  });
+  
+  it('should emit true if single source element match with predicate', function() {
+    var source = hot('--a--|', {a: 5});
+    var expected =   '-----(x|)';
+    
+    expectObservable(source.every(predicate)).toBe(expected, {x: true});
+  });
+  
+  it('should emit true if all source element matches with predicate', function() {
+    var source = hot('--a--b--c--d--e--|', {a: 5, b: 10, c: 15, d: 20, e: 25});
+    var expected =   '-----------------(x|)';
+    
+    expectObservable(source.every(predicate)).toBe(expected, {x: true});
+  });
+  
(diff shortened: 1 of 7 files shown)


spec/operators/every-spec.js
```diff
@@ -0,0 +1,67 @@
...
+describe('Observable.prototype.every()', function() {
+  function truePredicate(x) {
+    return true;
+  }
+  
+  function predicate(x) {
+      return x % 5 === 0;
+  }
+  
+  it('should return true if source is empty', function() {
```
kwonoj: This test case comes from [RxJS3](https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/every.js#L12). 

Little bit unsure if this behavior is expected, if source observable is empty then `every` will emit `true` always regardless of predicate.

src/operators/every.ts
```diff
@@ -0,0 +1,61 @@
...
+  
+  _next(value: T) {
+    const predicate = this.predicate;
+    
+    if (predicate === undefined) {
+      this.destination.error(new TypeError('predicate must be a function'));
+    }
+    
+    let result = tryCatch(predicate)(value, this.index++, this.source);
+    if (result == errorObject) {
```
roganov: Probably typo, should be `result === errorObject`.

kwonoj: Yes it is, thanks!

kwonoj:
* updated couple of testcases to ensure source element before subscription doesn't affect behavior of `every`

blesh:
@kwonoj ... I've found no such slowness.

> old every(predicate) with immediate scheduler x 100,478 ops/sec ±0.91% (90 runs sampled)
> new every(predicate) with immediate scheduler x 278,192 ops/sec ±0.75% (93 runs sampled)
	176.87% faster than Rx2

> old every(predicate, thisArg) with immediate scheduler x 91,583 ops/sec ±0.57% (92 runs sampled)
> new every(predicate, thisArg) with immediate scheduler x 247,939 ops/sec ±0.61% (95 runs sampled)
	170.73% faster than Rx2

blesh:
No worries about any more changes, I've got conflicts resolved...

kwonoj:
Interesting, seems my system configuration should have some oddness related with micro perf slowness, recall history of observing slowness only on my end :(

blesh:
It's the `Rx.config.slow_for_kwonoj = true` setting I added a month ago.

kwonoj:
:)

blesh:
Made a change that detects scalars for an increase in perf:

was

> old every(predicate) with immediate scheduler x 321,042 ops/sec ±0.75% (95 runs sampled)
> new every(predicate) with immediate scheduler x 1,921,189 ops/sec ±0.70% (96 runs sampled)
>	498.42% faster than Rx2

now

> old every(predicate) with immediate scheduler x 327,967 ops/sec ±0.70% (93 runs sampled)
> new every(predicate) with immediate scheduler x 4,126,076 ops/sec ±0.67% (94 runs sampled)
> 	1158.08% faster than Rx2

and also I did the same for detecting observables of static arrays:

was

> old static array observable every(predicate) with immediate scheduler x 190,344 ops/sec ±0.48% (96 runs sampled)
> new static array observable every(predicate) with immediate scheduler x 1,895,897 ops/sec ±0.62% (95 runs sampled)
	896.04% faster than Rx2

now 

> old static array observable every(predicate) with immediate scheduler x 179,314 ops/sec ±0.66% (93 runs sampled)
> new static array observable every(predicate) with immediate scheduler x 4,134,918 ops/sec ±0.72% (92 runs sampled)
	2205.96% faster than Rx2

Merged with d5120d2

Thanks @kwonoj!

blesh:
... in retrospect, I should have added those two new micro perf tests in separate previous commits... since i did them first... oh well. Next time.

blesh:
FWIW, the change I made is something we could do in other places for sure. `isEmpty` or `first` for example. Anything that returns a ScalarObservable could be optimized most likely.

PR #434, opened 2015-10-01
style(src): add lint tasks for TypeScript and JS

staltz:
Establish a code style using tslint (for src/) and eslint (for perf/
and spec/). The existing code style was approximately inferred by
analyzing which style was the most common. The lint tasks should also
run in Travis. An established code style is important for a large
project such as RxJS which expects multiple contributions from the
community, in order to avoid pollution of feature-related commits with
style refactors.

PS: I know this is a huge commit which touches everything, but it's meant to happen only once. Will be much easier to handle code style issues in the future if we have linting tasks in place.


kwonoj:
What about adding lint task to be executed when `build_all` is invoked?

blesh:
@staltz can you rebase this one?

blesh:
Also, it wouldn't be a bad thing to run this with build_test at least, so it was part of travis checks.

staltz:
Will do.

staltz:
All done

Issue #425, opened 2015-09-30
count(predicate) plus tests


staltz:
`count()` is currently missing the predicate function as argument.

Should we do this or not?

blesh:
Yes, that should be a thing.

staltz:
This is closed by #436 @blesh

staltz:
Actually I have closing powers, so here we go.

PR #436, opened 2015-10-01
feat(count): add predicate support in count()

staltz:
Add a predicate function argument (and a thisArg argument) to count()
operator. Also add respective tests that mirror RxJS 4 tests.

Resolves issue #425.


blesh:
merged with 42d1add... thanks @staltz

Issue #428, opened 2015-09-30
Marble testing for when inner observables unsubscribe


blesh:
This is related to discussion in #423 ... 

Basically the problem is that in something like `switch`, we don't have a way of asserting when inner subscriptions were unsubscribed from.

blesh:
one thought I had was reusing the `!` in the expectation marbles, and then matching it up to an array of unsusbscription assertions some how:

```js
let a = hot('--1--^--------2--------3-----');
let b = hot('--4--^-----5----------------6------7---');
let observableLookup = { a: a, b: b };
let e1 = hot('----^------a-------b-------', observableLookup);
let expected = '---------^-2-----(!^)----6------7---';

expectObservable(e1.switch()).toBe(expected).withSubs([a, b]).withUnsubs([a]);
```

.... it's not a particularly good idea, but it's a start.

I don't really like that it junks up the expected marbles, though.

blesh:
Perhaps the inner subs and unsubs could be in different marble diagrams.

The worst part though will be doing this *without* adding test specific code to the library. We may have to create custom `TestObservable` and `TestSubject` types that deal with this by creating a `TestSubscriber` instead of a `Subscriber` when `subscribe()` is called.

staltz:
To mirror RxJS 4, the equivalent of this https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/selectmany.js#L240-L244 in our marble diagram tests would be 
```js
let a = hot('--1--^--------2--------3-----');
let asubs = '------------^-------!';
let b = hot('--4--^-----5----------------6------7---');
let bsubs = '--------------------^------------------';
let observableLookup = { a: a, b: b };
let e1 = hot('----^------a-------b-------', observableLookup);
let expected =   '---------2-------------6------7---';

expectObservable(e1.switch()).toBe(expected);
expect(a.subscriptions).toBe(asubs);
expect(b.subscriptions).toBe(bsubs);
```

[...]

PR #440, opened 2015-10-01
feat(share): add the share operator

### LINKED ISSUE #439: Implement `share` operator
This is an alias for `publish().refCount()`, but is honestly preferable for simplicity's sake. It's easier to teach people `share()` than `publish().refCount()`, because to a newbie, the latter won't make much sense.

Ideally, this will be implemented by pulling in the `publish` method via import, rather than depending on it being on the Observable. Since this is a modular library, there is no guarantee publish has been added.

blesh:
closes #439

### DIFF
--- a/perf/micro/immediate-scheduler/operators/share.js
+++ b/perf/micro/immediate-scheduler/operators/share.js
@@ -1,0 +1,20 @@
+var RxOld = require("rx");
+var RxNew = require("../../../../index");
+
+module.exports = function (suite) {
+  
+    var oldShareWithImmediateScheduler = RxOld.Observable.range(0, 25, RxOld.Scheduler.immediate).share();
+    var newShareWithImmediateScheduler = RxNew.Observable.range(0, 25).share();
+    
+    return suite
+        .add('old share with immediate scheduler', function () {
+            oldShareWithImmediateScheduler.subscribe(_next, _error, _complete);
+        })
+        .add('new share with immediate scheduler', function () {
+            newShareWithImmediateScheduler.subscribe(_next, _error, _complete);
+        });
+
+    function _next(x) { }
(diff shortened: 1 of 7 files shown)


staltz:
I would merge :+1:

spec/operators/share-spec.js
```diff
@@ -0,0 +1,50 @@
...
+    var source = obs.share();
+    
+    expect(subscriptionCount).toBe(0);
+    
+    source.subscribe();
+    source.subscribe();
+    
+    expect(subscriptionCount).toBe(1);
+  });
+
```
staltz: Should we also have a marble test where `e1` is a cold Observable to make sense applying `share` on it? Also the case of multiple subscribers to a was-cold shared Observable makes a lot of sense to have a subscribe-and-unsubscribe marble diagram test that we mentioned in #428.

perf/micro/immediate-scheduler/operators/share.js
```diff
@@ -0,0 +1,23 @@
...
+        .add('old share with immediate scheduler', function () {
+            oldShareWithImmediateScheduler.subscribe(_next, _error, _complete);
+        })
+        .add('new share with immediate scheduler', function () {
+            newShareWithImmediateScheduler.subscribe(_next, _error, _complete);
+        });
+
+    function add(acc, x) {
+        return x + x
+    }
```
staltz: `add` is not used.

spec/operators/share-spec.js
blesh: Probably

perf/micro/immediate-scheduler/operators/share.js
blesh: heh... cruft.

PR #443, opened 2015-10-02
improving perf for mapping over ArrayObservable and ScalarObservables

blesh:
For mapping over scalar observables:

was

> old map of scalar with immediate scheduler x 648,735 ops/sec ±0.54% (94 runs sampled)
> new map of scalar with immediate scheduler x 1,329,174 ops/sec ±1.32% (94 runs sampled)
>	104.89% faster than Rx2

now

> old map of scalar with immediate scheduler x 632,782 ops/sec ±0.69% (94 runs sampled)
> new map of scalar with immediate scheduler x 4,032,053 ops/sec ±0.52% (95 runs sampled)
>	537.19% faster than Rx2

For mapping over array observables:

was

> old map of array with immediate scheduler x 147,364 ops/sec ±0.79% (91 runs sampled)
> new map of array with immediate scheduler x 267,194 ops/sec ±0.65% (94 runs sampled)
> 	81.32% faster than Rx2

now

> old map of array with immediate scheduler x 143,522 ops/sec ±1.18% (94 runs sampled)
> new map of array with immediate scheduler x 1,063,957 ops/sec ±0.63% (96 runs sampled)
>	641.32% faster than Rx2

### DIFF
--- a/perf/micro/immediate-scheduler/operators/map-array.js
+++ b/perf/micro/immediate-scheduler/operators/map-array.js
@@ -1,0 +1,27 @@
+var RxOld = require("rx");
+var RxNew = require("../../../../index");
+
+module.exports = function (suite) {
+
+    var oldSelectWithImmediateScheduler = RxOld.Observable.fromArray([1,2,3,4,5,6,7,8,9,10], RxOld.Scheduler.immediate).map(square).map(double);
+    var newSelectWithImmediateScheduler = RxNew.Observable.of(1,2,3,4,5,6,7,8,9,10).map(square).map(double);
+
+    return suite
+        .add('old map of array with immediate scheduler', function () {
+            oldSelectWithImmediateScheduler.subscribe(_next, _error, _complete);
+        })
+        .add('new map of array with immediate scheduler', function () {
+            newSelectWithImmediateScheduler.subscribe(_next, _error, _complete);
+        });
+
+    function square(x) {
+        return x * x;
+    }
+
+    function double(x) {
+        return x + x;
+    }
+    function _next(x) { }
+    function _error(e){ }
+    function _complete(){ }
+};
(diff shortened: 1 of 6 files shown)


blesh:
Okay, so the interesting thing here is you'll notice I had to change some tests...

One potential issue is that with these optimizations, if the user is doing something like mutating an outer-scoped variable in `map` and they're relying on that with something like `retry`, the mapping function really is only called once... In other words, an end user could be punished for using mutable state if I proceed with this PR.

For example:

```js
let arrObs = Observable.of(1,2,3,4,5,6);
let calls = 0;
let source = arrObs.map(x => {
   if(calls++ > 30) { throw 'bad' };
}).repeat();

source.subscribe();
```

In the example above, the mapping function will only really be called 6 times (one for each item in the original ArrayObservable returned by `of`), because `map` is returning an ArrayObservable directly, rather than a plain Observable that has been mapped.

I'd love input here as to whether or not this is a good idea @jhusain @mattpodwysocki @benjchristensen @headinthebox @steveorsomethin @trxcllnt

blesh:
Also, the "Rx2" is a misnomer. It's running against Rx3 right now. I'll update that to 4 soon.

robwormald:
It seems like an incredible performance boost. The 'gotcha' needs to be considered, but I'm in favor of faster things instead of slower things to cover use cases that are probably a bad idea anyway. 

My vote : add this and and document the case(s) where it might bite end-users. 537 percents is a lot of percents.

blesh:
Okay, after discussions with @staltz and @jhusain, among others, that occurred offline, this looks good. Although @jhusain pointed out a really good idea, that I could probably just override map on the ScalarObservable rather than add checking here, so I'm going to do that, then merge.

blesh:
One of the only reasons I haven't merged this yet is because I've got the heeby-jeebies about essentially removing the stream processing of Arrays (it becomes just Array.prototype.map under the hood, creating new array instances) I suspect there's a longer tail to this, were if you have `Observable.of(1,2,3)` and `map` on it 10 times or more consecutively, it ends up being faster to leave it as is. I think the *scalar* side of this is a huge win, but maybe not the array side...

PR #446, opened 2015-10-02
refactor(test): print out failure occurs from micro perf test

kwonoj:
Found out while working on https://github.com/ReactiveX/RxJS/pull/431, micro perf test runner silently terminates if some of test case fail and throw. Prints out stack trace (if available) or error to allow easier failure trace.

*After*
micro-fail-output


blesh:
Nice! Thanks @kwonoj... merged with 164fa4b

PR #447, opened 2015-10-02
feat(shareReplay): add the shareReplay() operator

staltz:
Add shareReplay() as a core operator. Vaguely relates to issue #439 and issue #298.

PR #448, opened 2015-10-02
perf(ReplaySubject): remove unnecessary computation

staltz:
Remove unnecessary call to `_getEvents()` in ReplaySubject's `_next()`. No apparent reason for doing this. All tests still pass after this commit.

Please merge PR #447 before merging this PR. When running the micro perf test for shareReplay after this commit is applied, we get this result:
```
node perf/micro share-replay
Testing against RxJS v 3.1.2
old shareReplay with immediate scheduler x 44,835 ops/sec ±16.27% (87 runs sampled)
new shareReplay with immediate scheduler x 92,828 ops/sec ±1.57% (92 runs sampled)
        107.05% faster than Rx v 3.1.2
```

I don't know why that line was there in ReplaySubject.ts. Seems to contribute nothing to the behavior of ReplaySubject.

### DIFF
--- a/src/subjects/ReplaySubject.ts
+++ b/src/subjects/ReplaySubject.ts
@@ -24,7 +24,6 @@ export default class ReplaySubject<T> extends Subject<T> {
   _next(value?) {
     const now = this._getNow();
     this.events.push(new ReplayEvent(now, value));
-    this._getEvents(now);
     super._next(value);
   }
 


blesh:
Bloacked by #447

staltz:
Blocked but #447 is now mergeable, so soon this should be as well.

blesh:
Merged with 488ac2e... thanks, @staltz

PR #450, opened 2015-10-02
chore(micro perf): bump up old rx version to latest

kwonoj:
RxJS v4 is now released, time to bump up version for micro perf test maybe?


blesh:
Merged with 75427a9... thanks @kwonoj!

Issue #390, opened 2015-09-23
Comprehensive marble tests for filter


blesh:
Should duplicate the current RxJS 3/4 tests

Should adhere to our contribution guidelines and operator authoring docs.

staltz:
I'll start working on this. There are lots of tests here https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/where.js

staltz:
Most RxJS legacy tests for filter are now done. Except these:

- https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/where.js#L288
- https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/where.js#L608
- https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/where.js#L668

luisgabriel:
@staltz Rewriting the test this way it passes:

```javascript
it('should invoke predicate once for each checked value', function () {
    var source = hot('-1--2--^-3-4-5-6--7-8--9--|');
    var expected =          '--3---5----7-------|';

    var invoked = 0;
    var predicate = function (x) {
      invoked++;
      return isPrime(x);
    };

    var r = source
      .filter(predicate)
      .do(null, null, function () {
        expect(invoked).toEqual(7);
      });

    expectObservable(r).toBe(expected);
  });
```

Could be a bug in `hot`?

kwonoj:
Seems subscriptionframe `^` is not honored with plain subscription, all elements prior to subscription is pushed.

blesh:
That's because plain subscribe would need to be scheduled at `0` on the test scheduler.

kwonoj:
Yes, assumed so. Think do not need to modify behavior of `testScheduler` or either `hot` for above specific cases, @staltz what do you think?

staltz:
Hmm, yeah, I think using `.do()` to make the assert on `invoked` is fine. And now it's clear that it should have been subscribed on the TestScheduler.

blesh:
Yeah, as a general rule, the only way to know for sure when something will happen is to schedule it on the scheduler.

blesh:
Are these complete now, @staltz? I feel like this got wrapped up after this discussion.

staltz:
@blesh did the missing test get submitted by @luisgabriel ? I can't recall seeing it.

luisgabriel:
@staltz @blesh I'll submit a PR. 

[...]

PR #451, opened 2015-10-02
test(filter): add filter tests from RxJS legacy

staltz:
Add most RxJS legacy tests. Just 3 tests are missing still, which are
related to test helpers not yet available in RxJS Next.

Partially resolves issue #390.


blesh:
rebased and merged with 3d833f3

Thanks @staltz!

PR #452, opened 2015-10-02
test(scan): add scalar, array micro perf test for scan operator

kwonoj:
- remove perf test for scan under current thread scheduler

add micro perf test for scalar, array specific for `scan` operator, also removes existing test case under current-thread-scheduler. As same as other operators like `zip` (https://github.com/ReactiveX/RxJS/pull/357), test under current thread scheduler does test over different schedulers only for sources.


blesh:
You're going to need to rebase this one. I rebased it locally, but it's not passing lint. Moving on to merge the next PR if possible. (trying to get caught up)

kwonoj:
I'll do rebase and update PR once it's satisfied.

kwonoj:
Rebased with lint-satisfied.

blesh:
Thank you, @kwonoj! Merged with 0b6e35e

PR #454, opened 2015-10-03
fix(buffer): change behavior of buffer

adrianomelo:
The behavior of buffer operator was different from RxJS 4. This patch changes the behavior of buffer and adds tests from RxJS 4 and tests with never, empty and error.

### DIFF
--- a/spec/operators/buffer-spec.js
+++ b/spec/operators/buffer-spec.js
@@ -1,12 +1,112 @@
 /* globals describe, it, expect, expectObservable, hot */
 var Rx = require('../../dist/cjs/Rx');
+var Observable = Rx.Observable;
 
 describe('Observable.prototype.buffer', function () {
+  it('should work with empty and empty selector', function () {
+    var a = Observable.empty();
+    var b = Observable.empty();
+    var expected = '|';
+    expectObservable(a.buffer(b)).toBe(expected);
+  });
+  it('should work with empty and non-empty selector', function () {
+    var a = Observable.empty();
+    var b = hot('-----a-----');
+    var expected = '|';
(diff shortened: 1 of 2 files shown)


blesh:
Good. What was the problem with the original behavior?

adrianomelo:
@blesh, the original behavior:
 - doesn't accept buffers of length zero [0];
 - flushes the buffer when selector completes [0], and
 - flushes the buffer when a non-empty source observable completes, even with a 'never' selector [1].

Despite the changes made in this patch, this operator still not fully compatible with RxJS 4... There is an alternate method with buffer opening selector **and** buffer closing selector as parameters, we only support buffer closing selector [2]. 

[0] - https://github.com/adrianomelo/RxJS-1/blob/buffer/spec/operators/buffer-spec.js#L80
[1] - https://github.com/adrianomelo/RxJS-1/blob/buffer/spec/operators/buffer-spec.js#L18
[2] - https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/buffer.md

adrianomelo:
btw, I need to change some names before you push into master..

blesh:
The alternate argument overloads were broken into `bufferWhen` and `bufferToggle`, because JavaScript doesn't really support overloads well, and those particular overloads actually completely changed the behavior of the operator.

blesh:
I'll review soon. Might not be until Monday or Tuesday

adrianomelo:
ok, no problem!
The commit I'll be ready in a couple of minutes..

blesh:
Thank you for all of the hard work! This been merged with af603bd.  Thanks again, @adrianomelo!

PR #456, opened 2015-10-03
feat(operator): add timeInterval operator

kwonoj:
* adding timeInterval operator, as well as migrating RxJS4 testcases

### DIFF
--- a/perf/micro/current-thread-scheduler/operators/timeinterval.js
+++ b/perf/micro/current-thread-scheduler/operators/timeinterval.js
@@ -1,0 +1,20 @@
+var RxOld = require('rx');
+var RxNew = require('../../../../index');
+
+module.exports = function (suite) {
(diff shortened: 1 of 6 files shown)


staltz:
Pretty cool :sunglasses:

kwonoj:
Thanks! :smiley:

blesh:
Is this something that should be a core operator? It seems like this could be accomplished with `scan`. I love it, and I want to keep it, but perhaps this should go in the extended set of operators.

kwonoj:
Do you suggest to move to extended operator? I'm feeling somewhat similar too.

blesh:
I think so. It's one of those operators I think is cool, and that I want documented and available to people, but ultimately probably shouldn't ship with core.

kwonoj:
OK, I'll update PR as suggested.

kwonoj:
* Now `timeInterval` operator is placed under extended operators
* updated PR to satisfy lint

blesh:
Thanks for your hard work!

Issue #461, opened 2015-10-05
changing signature of startWith() to accept scheduler as last parameter?


kwonoj:
While other operators like `throttle`, `timeout`, `delay`, etcs are accepting `scheduler` as last parameter, [`startWith`] (https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/startwith.md) accept scheduler as first parameter. 

```
Rx.Observable.prototype.startWith([scheduler] ...args)
Rx.Observable.prototype.timeout(dueTime, [other], [scheduler])
Rx.Observable.prototype.throttle(windowDuration, [scheduler])
```

Maybe better to align signature of `startWith` to accept scheduler as last parameter also?

staltz:
I agree

blesh:
Yes... schedulers should always be the last argument. At least according the the [Rx Design Guidelines from Microsoft, Article 6.11](http://blogs.msdn.com/b/rxteam/archive/2010/10/28/rx-design-guidelines.aspx)

kwonoj:
OK, seems I can update currently opened PR directly instead of creating new one.

staltz:
I guess we can close this.

PR #462, opened 2015-10-05
fix(operator): startWith operator accepts scheduler, multiple values

kwonoj:
- update startWith operator to accept multiple values as well
- update signature of startWith to accept scheduler as last parameter
- fix current micro perf test to accept scheduler properly, expand test for multiple values
- expand test coverage

update behavior to accept scheduler, and multiple values as same as [RxJS4](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/startwith.md), created issue https://github.com/ReactiveX/RxJS/issues/461 for signature alignment in further.

### DIFF
--- a/spec/operators/startWith-spec.js
+++ b/spec/operators/startWith-spec.js
@@ -1,20 +1,92 @@
-/* globals describe, it, expect */
+/* globals describe, it, expect, expectObservable, hot, rxTestScheduler */
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
 describe('Observable.prototype.startWith()', function () {
-  it('should start an observable with given value', function (done) {
-    var source = 'source';
-    var init = 'init';
-    var expected = [init, source];
+  var defaultStartValue = 'x';
 
-    var i = 0;
-    Observable.of(source)
-      .startWith(init)
-      .subscribe(function (x) {
-        expect(x).toBe(expected[i++]);
-      }, null, function () {
-        done();
-      });
+  it('should start an observable with given value', function () {
+    var e1 =   hot('--a--|');
+    var expected = 'x-a--|';
+
+    expectObservable(e1.startWith(defaultStartValue)).toBe(expected);
+  });
+
+  it('should start with given value and does not completes if source does not completes', function () {
+    var e1 =   hot('----a-');
+    var expected = 'x---a-';
+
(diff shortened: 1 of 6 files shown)


spec/operators/startWith-spec.js
```diff
@@ -1,20 +1,93 @@
...
+  it('should start with given value and source both if source emits single value', function() {
+    var e1 = Observable.of('a');
+    var expected = '(xa|)';
+    
+    expectObservable(e1.startWith(defaultStartValue)).toBe(expected);
+  });
+  
+  it('should start with given values when given value is more than one', function() {
+    var e1 =   hot('-----a--|');
+    var expected = '(yz)-a--|';
```
kwonoj: This marble actually brings one curiosity, I thought `()` grouped marble treated as single frame, like

```
'N-a-|'
 N === (yz)
```

but current testScheduler takes frame from each element in group,

```
'NNNN-a-|'
 NNNN === (yz)
```

Not sure if my understanding was incorrect until now.

staltz:
:+1:

spec/operators/startWith-spec.js
staltz: Yes, `(yz)` means that all those values happen exactly at the frame `(` occupies, and the remaining `yz)` don't affect at all in terms of time frame. They just specify the content of that synchronous time window. I.e. they don't push the remaining diagram to the right.

staltz: And there is no other way, because otherwise if the parenthesis group would push the diagram to the right it would mean the parenthesis group would **not** be synchronous at that single frame.

kwonoj:
PR's updated to reflect https://github.com/ReactiveX/RxJS/issues/461, now `scheduler` is accepted as last parameter.

blesh:
@kwonoj that seems impossible. If you follow the code, they're the same thing. fromArray === ArrayObservable.create === new ArrayObservable

If so, it's an interesting bug. Can you confirm?

kwonoj:
@blesh Just checked code and you're right, it is obviously interesting. I'll re confirm to ensure if I did any mistakes or errors.

kwonoj:
PR is now updated to correctly initialize arrayObservable, also lint-satisfied.

blesh:
Merged with d1d339a... great work @kwonoj

Issue #470, opened 2015-10-06
Use Closure Compiler on global output file


blesh:
In an effort to reduce library size, use Google Closure compiler on output global `Rx.js` file to create `Rx.min.js` ... currently, uglify isn't doing that great of a job reducing the size of the minified output.

staltz:
Seems like this can be closed.

PR #471, opened 2015-10-06
refactor(build): use closure compiler to generate global output

kwonoj:
Tried to create changes for https://github.com/ReactiveX/RxJS/issues/470.

* Uses [google-closure-compiler](https://www.npmjs.com/package/google-closure-compiler) package, requires JRE to be installed to execute `build_global`.
* Simply replacing `uglifyjs` to `closure` to generate output with source map, does not apply any other compilation options.

Not sure if this is feasible change for issue, trying to leverage as starting point. 
Generated size from closure is about 160kb, uglify does about 280kb. (except source map)


blesh:
Thanks for the quick turnaround on that, @kwonoj! Merged with e4a69cc

PR #473, opened 2015-10-07
fix(TestScheduler): stop sorting actual results

### LINKED ISSUE #422: Merge semantics for race conditions between synchronous notifications

blesh:
- fixes an issue where actual results were being sorted when they should not have been
- ensures rescheduled Actions will be added to subscription and torn down when appropriate
- uses state to convey notification information to action work function rather than closure

closes #422

### DIFF
--- a/spec/operators/merge-map-spec.js
+++ b/spec/operators/merge-map-spec.js
@@ -1,8 +1,25 @@
+/* globals expectObservable, cold, hot, describe, it, expect */
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 var Promise = require('promise');
 
 describe('Observable.prototype.mergeMap()', function () {
+  it('should mergeMap many regular interval inners', function () {
+    var a =   cold('----a---a---a---(a|)'                    );
+    var b =   cold(    '----b---b---(b|)'                    );
+    var c =   cold(                '----c---c---c---c---(c|)');
+    var d =   cold(                        '----(d|)'        );
+    var e1 =   hot('a---b-----------c-------d-------|'       );
+    var expected = '----a---(ab)(ab)(ab)c---c---(cd)c---(c|)';
+
+    var observableLookup = { a: a, b: b, c: c, d: d };
+    var source = e1.mergeMap(function (value) {
+      return observableLookup[value];
+    });
+
+    expectObservable(source).toBe(expected);
+  });
+
   it('should map values to constant resolved promises and merge', function (done) {
     var source = Rx.Observable.from([4,3,2,1]);
     var project = function (value) {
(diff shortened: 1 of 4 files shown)


blesh:
I'm going to merge this one right away, since it fixes critical bugs with the test scheduler.

spec/operators/merge-map-spec.js
```diff
@@ -1,8 +1,25 @@
...
 var Promise = require('promise');
 
 describe('Observable.prototype.mergeMap()', function () {
+  it('should mergeMap many regular interval inners', function () {
+    var a =   cold('----a---a---a---(a|)'                    );
+    var b =   cold(    '----b---b---(b|)'                    );
+    var c =   cold(                '----c---c---c---c---(c|)');
+    var d =   cold(                        '----(d|)'        );
+    var e1 =   hot('a---b-----------c-------d-------|'       );
+    var expected = '----a---(ab)(ab)(ab)c---c---(cd)c---(c|)';
```
staltz: It finally looks like it should. (ab) :+1:

src/testing/ColdObservable.ts
```diff
@@ -29,12 +29,11 @@ export default class ColdObservable<T> extends Observable<T> implements Subscrip
   scheduleMessages(subscriber) {
     const messagesLength = this.messages.length;
     for (let i = 0; i < messagesLength; i++) {
-      const message = this.messages[i];
+      let message = this.messages[i];
```
staltz: Is there a reason behind this? Do we default to `let`?

blesh: No. It was probably residue from me fiddling with code while debugging

PR #474, opened 2015-10-07
test(defaultIfEmpty): add marble tests for defaultifEmpty operator

kwonoj:
*apply marble test cases for `defaultIfEmpty` operator.

Coverage wise, existing test cases are already same as RxJS4 test cases.

### DIFF
--- a/spec/operators/defaultIfEmpty-spec.js
+++ b/spec/operators/defaultIfEmpty-spec.js
@@ -1,50 +1,41 @@
-/* globals describe, it, expect */
+/* globals describe, it, expect, expectObservable, hot */
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
 describe('Observable.prototype.defaultIfEmpty()', function () {
-  it('should return the argument if Observable is empty', function (done) {
-    var emptyObservable = Observable.empty();
-    emptyObservable.defaultIfEmpty(2)
-      .subscribe(function (x) {
-        expect(x).toBe(2);
-      }, null, done);
+  it('should return the argument if Observable is empty', function () {
+    var e1 = Observable.empty();
+    var expected = '(x|)';
+
+    expectObservable(e1.defaultIfEmpty('x')).toBe(expected);
   });
 
-  it('should return null if the Observable is empty and no arguments', function (done) {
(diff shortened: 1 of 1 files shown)


spec/operators/defaultIfEmpty-spec.js
```diff
@@ -1,50 +1,42 @@
...
 var Observable = Rx.Observable;
 
 describe('Observable.prototype.defaultIfEmpty()', function () {
-  it('should return the argument if Observable is empty', function (done) {
-    var emptyObservable = Observable.empty();
-    emptyObservable.defaultIfEmpty(2)
-      .subscribe(function (x) {
-        expect(x).toBe(2);
-      }, null, done);
+  var defaultValue = 'x';
```
staltz: I'd rather keep this var inside each test case, even if it looks like repetition, because it makes tests look more obvious, less magical.

kwonoj: as you mentioned, intention was reducing repeated same values over test cases, served as base fixture setups. Your suggestion also makes sense though.

staltz: The reasoning behind keeping repetition is: we should be able to copy and paste a test case and it should contain all information needed without context surrounding it.

kwonoj: Agreed, will update PR as suggested.

blesh:
@kwonoj ... for the future. Try to put the Issue # this relates to at the bottom of your commit message. (see the [Commit Message Guidelines](https://github.com/ReactiveX/RxJS/blob/master/CONTRIBUTION.md#commit)). 

If you add commits like:
```
fix(Foo): stop foo from being bad

- did something to foo
- update bar to reflect foo

fixes #10000
closes #10001
```

or 

```
feat(qux): added a qux to our dux

- all ducks no qux more loudly

closes #10000
```
... it will close the related issue(s) when the commit is merged in.

If it doesn't close it, but relates to it, you can just add "relates to #10000" as a footer

This will save us from having to figure out what has been finished and what hasn't.

kwonoj:
Thanks for pointing out, definitely will do.

kwonoj:
PR's updated as suggested to remove common values outer scope of test cases, along with updating commit message.

Issue #389, opened 2015-09-23
Comprehensive marble tests for map


blesh:
Should duplicate the current RxJS 3/4 tests

Should adhere to our contribution guidelines and operator authoring docs.

luisgabriel:
I'll start migrating tests from: https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/select.js

blesh:
Okay, great.

luisgabriel:
@blesh Is there a special reason to keep `map()` and `mapTo()` as separate operators?

blesh:
@luisgabriel there are a few reasons:

1. JavaScript doesn't really have "overloading" of methods. You can leverage dynamic typing and reflection to mimic it, but it's not the same thing. because....
2. reflection/type assertions are slow.

For such a commonly used operator as map, I didn't want to incur a penalty at each call to check to see what type of thing was passed to map.

... so.. we have two different operators.

luisgabriel:
@blesh got it. Makes sense. I'll work on some tests for `mapTo()` then.

PR #478, opened 2015-10-07
map tests

luisgabriel:
This PR adds most of RxJS 4 tests for map. 

There are two still missing that possibly will require changes on the operator side. I'll take a closer look at them next.

Related to #389.


luisgabriel:
Just updated the PR and now it includes all tests from RxJS 4. The only one missing is actually related to `mapTo()` here in RxJS Next: https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/select.js#L242

blesh:
Merged with 414c1b7 ... thank you very much, @luisgabriel!

PR #482, opened 2015-10-08
test(mapTo): add micro-benchmark performance test for mapTo operator

kwonoj:
relates to #350 

Effort to cover micro performance test. for old RxJs, used `map(object)` to corresponds to `mapTo` operator.


blesh:
Merged with e5dfd8c .... thanks!

PR #485, opened 2015-10-08
Subject.create fix

blesh:
Calling `Subject.create` with an `Observable` that had never had an operator applied to it resulted in this error. This was because it called `Observable.prototype._subscribe` internally, and that method assumed there was an operator instance in `this.operator`, which in that case there would not have been.

This fix adds a check to see if there's an operator or not, and just uses the original subscriber if there isn't one.

This seems like a case for any subclassed observables that call `super._subscribe`, prior to having any operators applied.

### DIFF
--- a/src/Subject.ts
+++ b/src/Subject.ts
@@ -16,12 +16,11 @@ const _subscriberNext = Subscriber.prototype._next;
 const _subscriberError = Subscriber.prototype._error;
 const _subscriberComplete = Subscriber.prototype._complete;
 
-const _observableSubscribe = Observable.prototype._subscribe;
 
 export default class Subject<T> extends Observable<T> implements Observer<T>, Subscription<T> {
   _subscriptions: Subscription<T>[];
   _unsubscribe: () => void;
-  
+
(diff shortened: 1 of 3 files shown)


blesh:
@trxcllnt ... probably. I'll fiddle with it.

blesh:
@trxcllnt  ... yeah... I was able to just call _subscribe on the source directly in `BidirectionalSubject`'s `_subscribe` implementation. It's probably a little more efficient now.

blesh:
I'm going to go ahead an merge this since it was a trivial fix.

src/Subject.ts
```diff
@@ -177,7 +176,8 @@ class BidirectionalSubject<T> extends Subject<T> {
   }
 
   _subscribe(subscriber: Subscriber<T>) {
-    return _observableSubscribe.call(this, subscriber);
+    const operator = this.operator;
+    return this.source._subscribe.call(this.source, operator ? operator.call(subscriber) : subscriber);
```
trxcllnt: @blesh is `call` necessary here?

blesh:
Haha. No. I'm not sure why I did that.

PR #487, opened 2015-10-08
test(reduce): add marble tests for reduce operator

kwonoj:
- reduce now throws if source is empty without seed
- add test coverage for reduce operator

relates to #309 

Updated operator behavior to follow [RxJS4](https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/reduce.js#L127-L138), now operator throws if source is empty and there isn't any seed.

### DIFF
--- a/spec/operators/reduce-spec.js
+++ b/spec/operators/reduce-spec.js
@@ -4,13 +4,163 @@ var Observable = Rx.Observable;
 
 describe('Observable.prototype.reduce()', function () {
   it('should reduce', function () {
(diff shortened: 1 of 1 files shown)


blesh:
You'll need to roll back that particular change, and change the tests accordingly.

kwonoj:
PR's updated to revert operator behavior, align test case accordingly.

blesh:
Awesome. Merged with 407a55e... thanks as always, @kwonoj

PR #490, opened 2015-10-09
Faster tests

blesh:
Started removing more timer-based tests.

- updates tests for `bufferWhen` to more of a full-suite of marble tests.
- fixes `windowWhen` to use marble tests

ATTENTION @staltz ... the windowWhen tests will interest you. We can probably weaponize that approach into something you can use for `groupBy`.

### DIFF
--- a/spec/helpers/test-helper.js
+++ b/spec/helpers/test-helper.js
@@ -155,3 +155,15 @@ global.lowerCaseO = function lowerCaseO() {
 
   return o;
 };
+
+var ___start;
+beforeEach(function () {
+  ___start = Date.now();
+});
+
+afterEach(function () {
+  var elapsed = Date.now() - ___start;
+  if (elapsed > 500) {
+    throw 'slow test!';
+  }
+});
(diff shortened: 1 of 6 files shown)


blesh:
So far I've cut ~2 seconds off of the tests with that. (down to 4s or so)

staltz:
You mean the materialize approach for the inner observables? Hmm, smart. But I already got my metastream testing problem solved, and we should probably use it because (1) easier to read the tests, (2) makes that test independent of materialize, just in case materialize eventually has a bug.

blesh:
@staltz that's cool. I'll try it with the new test methods. I was thinking what I did above could be made into a test helper or method, but whatever works.

blesh:
```
678 specs, 0 failures
Finished in 0.939 seconds
```

Down from ~ 6 seconds. (I think a month ago we had about 300 tests and it took ~15 seconds to run!)

kwonoj:
> Down from ~ 6 seconds

I love it. :heart_eyes:

blesh:
@staltz ... I know you have a better method for testing the `window` operators (from your `groupBy` tests), I haven't dug into that, but I'm going to merge this PR for productivity reasons. I don't want people sitting waiting for 6 seconds every build or test run for everything to get done.

I think I could do more with the Scheduler tests to introduce some DI and test them more thoroughly and quickly as well... but that's not a high priority at the moment.

This wasn't meant to be comprehensive, just a productivity booster. Especially when building and linting already takes so long.

PR #492, opened 2015-10-09
test(partition): add marble tests for partition operator

kwonoj:
relates to #309 

Replace existing test to marble test, migrating RxJS4 test also.

### DIFF
--- a/spec/operators/partition-spec.js
+++ b/spec/operators/partition-spec.js
@@ -1,78 +1,159 @@
-/* globals describe, it, expect */
+/* globals describe, it, expect, expectObservable, hot */
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
 describe('Observable.prototype.partition()', function () {
-  it('should partition an observable into two using a predicate', function (done) {
-    var expectedPositive = [0, 1, 2];
-    var expectedNegative = [-3, -2, -1];
-    var completed = false;
-    var positiveValues;
-    var negativeValues;
-
-    function completer() {
-      if (completed) {
-        expect(positiveValues).toEqual(expectedPositive);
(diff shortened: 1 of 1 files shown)


spec/operators/partition-spec.js
```diff
@@ -1,78 +1,81 @@
...
+    var unsub =     '-------!';
+    var expected = ['--a-',
+                    '----b-'];
+    var result = e1.partition(predicate);
 
-    positiveStream.subscribe(rejecter, completer, rejecter);
-    negativeStream.subscribe(rejecter, completer, rejecter);
+    for (var idx = 0; idx < result.length; idx++ ) {
+      expectObservable(result[idx], unsub).toBe(expected[idx]);
+    }
```
staltz: Maybe this one too should use `expectObservableArray` ?

kwonoj: This test case is only case accepting unsubscription as parameter, choose simplify `expectObservableArray` and let loop explicitly with unsubscription.

staltz: alright

blesh:
We still have a few missing tests for this operator. It looks like this was meant to just rewrite the existing tests, would you like to tackle the rest @kwonoj?

The ones I can think of that are missing are:

- empty
- throw
- error thrown in predicate
- never
- predicate never matches anything
- predicate always matches everything
- no predicate passed (should error I think?)

kwonoj:
Absolutely. As mentioned this PR was intended to 'migrate' only, but it makes more sense to add additional coverage. I'll cover them up and update PR.

kwonoj:
Updated PR by adding additional test coverage.

blesh:
merged with 0fcf52e... Thanks @kwonoj!

PR #502, opened 2015-10-10
chore(micro perf): align test case names

### LINKED ISSUE #501: defaultIfEmpty micro perf test broken for Rx 4.0.0
Currently it's not returning any value at all.

kwonoj:
- test case name matches to operator
- output for test lookup failure, primitive wildcard matching for test
runner
- bump up rx to latest

closes #501 

This PR includes couple of changes related with issue.
- test case names are changed to match with operator
: some of previous test cases were having dash-separated names per operator, like `default-if-empty` for `defaultIfEmpty`. those schemes are dropped and dash's only being used to have separate test case for operators, aligned by renaming.

- bump up rx to latest 4.0.1
: this is main cause of issue, 4.0.0 has known issue of `defaultIfEmpty` does not work (https://github.com/Reactive-Extensions/RxJS/issues/938), and fixed in 4.0.1 (https://github.com/Reactive-Extensions/RxJS/pull/940/files)

- added `perf_micro` task
: now able to execute run micro perf test via `npm run perf_micro`. Arguments can be passed like `npm run perf_micro defaultIfEmpty -- --csv output.csv`. Note `-- --csv` is not mistypo, `--` allows npm to pass following double-dash argument `--csv` into task instead of consume by npm itself.

- adding small behavior enhancement in micro perf test runner
: Micro perf test runner now will say if none of test can be found compare to given argument. Also added very primitive wildcard matching to execute tests like `npm run perf_micro merge*`. Single test case matching still works as same as before.


blesh:
Awesome... merged with 8e6d560. I've been wanting to do that one for a while.

I've also added a check in there to make sure the csv arguments were not treated like file names.

PR #503, opened 2015-10-10
Fix type signature of static operators

luisgabriel:
Fix the type signature of a lot of static operators.

We could avoid this kind of mismatch if the operator assignments on `Rx.ts` were type-checked. But I'm not sure how we could enforce that.


staltz:
:+1:

blesh:
Merged with e5364de... thanks @luisgabriel!!

PR #504, opened 2015-10-10
test(mapTo): add new test cases

luisgabriel:
add tests from `map()` that makes sense to `mapTo()`.

relates to #389.


luisgabriel:
Just updated the PR including the request changed.

blesh:
Thanks, @luisgabriel!

PR #505, opened 2015-10-11
test(retry): add micro-benchmark performance test for retry operator

kwonoj:
relates to #350


blesh:
Great! merged with 7333a46! Thanks @kwonoj

PR #506, opened 2015-10-12
test(forkJoin): add micro-benchmark performance test for forkJoin ope…

kwonoj:
…rator

relates to #350


blesh:
Thanks again! merged with b707361

Issue #375, opened 2015-09-23
Comprehensive marble tests for groupBy


blesh:
Should duplicate the current RxJS 3/4 tests

Should adhere to our contribution guidelines and operator authoring docs.

staltz:
I'm picking this one.

staltz:
Just a quick report on this issue: writing the marble tests for this operator is not trivial: I am seeing how to improve the TestScheduler so that it can assert also Observables-of-Observables, which is what this operator returns. It's a better solution to, say, applying .mergeAll() after groupBy, or reverting to raw manual tests with the scheduler.

I also fixed a small bug in groupBy, so eventually I'll submit a PR with 3 commits, one for tests, another for groupBy fix, another to TestScheduler improvements.

staltz:
Example: we will be able to write this

```js
it('should group values with a key comparer', function () {
    var values = {
      a: '  foo',
      b: ' FoO ',
      c: 'baR  ',
      d: 'foO ',
      e: ' Baz   ',
      f: '  qux ',
      g: '   bar',
      h: ' BAR  ',
      i: 'FOO ',
      j: 'baz  ',
      k: ' bAZ ',
      l: '    fOo    '
    };
    var e1 = hot('-1--2--^-a-b-c-d-e-f-g-h-i-j-k-l-|', values);
    var expected =      '--w---x---y-z-------------|';
    var w = cold(         'a-b---d---------i-----l-|', values);
    var x = cold(             'c-------g-h---------|', values);
    var y = cold(                 'e---------j-k---|', values);
    var z = cold(                   'f-------------|', values);
    var expectedValues = { w: w, x: x, y: y, z: z };
    //var expectedValues = { w: 'foo', x: 'bar', y: 'baz', z: 'qux' };

    var source = e1
      .groupBy(function (x) { return x.toLowerCase().trim(); });
      //.map(function (g) { return g.key; });
    expectObservable(source).toBe(expected, expectedValues);
  });
``` 

Relevant also for window operators.

blesh:
Ah, I see the problem.

I think perhaps the API should remain the same, and we'll just need to analyze expected values to see if they are either `ColdObservable` or `HotObservable`, and pull the marbles out of them, convert them to messages, and put them in the outer message value . In contrast, we'll need to subscribe to the resulting observables and do the same thing. The `toDeepEqual` assertion should handle the rest.

Basically, you're going to create Arrays of messages, that sometimes have values that are Arrays of messages.

[...]

PR #508, opened 2015-10-12
GroupBy(Until) tests and fixes

staltz:
Finishes the last bits and pieces of #375.

In summary:
- Add tests to match groupByUntil from RxJS legacy
- Add basic test cases (empty, never, etc) that were missing for groupBy
- Fix some bugs related to groupBy


blesh:
Merged with 57e4207... good stuff, @staltz

PR #513, opened 2015-10-13
feat(operator): add debounce, debounceTime operator

### LINKED ISSUE #493: debounce operator with duration selector
While trying to migrate test coverage in RxJS4, could observe difference between signature of `debounce` operator

RxJS4
```
Rx.Observable.prototype.debounce(dueTime, [scheduler])
Rx.Observable.prototype.debounce(durationSelector)
```

while current signature only allows first one
`debounce<T>(dueTime: number, scheduler: Scheduler = nextTick)`

Not sure if this is intended design or not - should operator to be updated, or test coverage for latter can be dropped off?

kwonoj:
- add debounce operator for duration selector, debounceTime operator for numbered time behavior
- migrated RxJS4 test cases also expands test coverage
- added micro-perf benchmark test

Effort to close #493.

### DIFF
--- a/spec/operators/debounce-spec.js
+++ b/spec/operators/debounce-spec.js
@@ -1,18 +1,233 @@
-/* globals describe, it, expect */
+/* globals describe, it, expect, expectObservable, hot, cold, rxTestScheduler */
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
 describe('Observable.prototype.debounce()', function () {
-  it('should delay calls by the specified amount', function (done) {
-    var expected = [3, 4];
-    var source = Observable.concat(Observable.of(1),
+  function getTimerSelector(x) {
+    return function () {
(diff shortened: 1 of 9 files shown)


spec/operators/debounce-spec.js
```diff
@@ -1,18 +1,245 @@
...
+  });
+
+  it('should raise error when selector observable raises error', function () {
+    var e1 =   hot('--------a--------b--------c---------|');
+    var expected = '---------a---------b---------#';
+    var selector = [
+      cold('-x-y-'),
+      cold('--x-y-'),
+      cold('---#')
+    ];
```
staltz: Would be nice to align these with the e1 and the `expected`, for visual aid.

kwonoj: Yes, that'd be better. Will give it a try to update.

blesh:
Merged with a1e652f! ... thanks as always, @kwonoj!

PR #515, opened 2015-10-13
windowCount: tests and fixes

### LINKED ISSUE #378: comprehensive marble tests for windowCount

staltz:
Resolves #378 .

### DIFF
--- a/spec/operators/windowCount-spec.js
+++ b/spec/operators/windowCount-spec.js
@@ -3,31 +3,103 @@ var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
 describe('Observable.prototype.windowCount', function () {
-  function mergeMapfunction(x) {
-    return x.toArray();
-  }
+  it('should emit windows with count 2 and skip 1', function () {
+    var source = hot('^-a--b--c--d--|');
+    var subs =       '^             !';
+    var expected =   'u-v--x--y--z--|';
+    var u = cold(    '--a--(b|)      ');
+    var v = cold(      '---b--(c|)   ');
+    var x = cold(         '---c--(d|)');
+    var y = cold(            '---d--|');
+    var z = cold(               '---|');
+    var values = { u: u, v: v, x: x, y: y, z: z };
 
-  it('should emit windows at intervals', function () {
-    var e1 =   hot('--a--b--c--d--|');
-    var expected = '-----w--x--y--(z|)';
+    var result = source.windowCount(2, 1);
 
-    expectObservable(e1.windowCount(2,1).mergeMap(mergeMapfunction))
-      .toBe(expected, { w: ['a', 'b'], x: ['b', 'c'], y: ['c', 'd'], z: ['d'] });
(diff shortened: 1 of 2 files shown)


spec/operators/windowCount-spec.js
```diff
@@ -3,31 +3,103 @@ var Rx = require('../../dist/cjs/Rx');
...
-  }
+  it('should emit windows with count 2 and skip 1', function () {
+    var source = hot('^-a--b--c--d--|');
+    var subs =       '^             !';
+    var expected =   'u-v--x--y--z--|';
+    var u = cold(    '--a--(b|)      ');
+    var v = cold(      '---b--(c|)   ');
+    var x = cold(         '---c--(d|)');
+    var y = cold(            '---d--|');
+    var z = cold(               '---|');
```
staltz: @kwonoj @blesh Notice `z` is the empty window at the end. I noticed this is actually necessary. By the time `z` is opened, the observer simply does not know what will happen next, whether it'll be `next` or `complete`. And notice window `u` starts immediately at the beginning. This is to stay consistent with `windowTime` and `window` operators that do the same.

staltz:
Micro performance was not significantly affected:

Before this PR:
```
                                         |                     RxJS 4.0.1 |             RxJS 5.0.0-alpha.3 |          factor |      % improved
                 windowcount - immediate |               102,134 (±1.71%) |               140,875 (±3.91%) |           1.38x |           37.9%
                             windowcount |                45,378 (±2.02%) |               151,652 (±0.98%) |           3.34x |          234.2%
```


After this PR:
```
                                         |                     RxJS 4.0.1 |             RxJS 5.0.0-alpha.3 |          factor |      % improved
                 windowcount - immediate |                92,958 (±1.50%) |               159,954 (±4.25%) |           1.72x |           72.1%
                             windowcount |                44,771 (±1.20%) |               158,254 (±2.07%) |           3.53x |          253.5%
```

spec/operators/windowCount-spec.js
kwonoj: Internal behaviorwise I think that makes sense, but also I think it brings sort of confusions who uses those operators such as pre-RxJS4 issue asked about this behavior https://github.com/Reactive-Extensions/RxJS/issues/148, notable thing is some of operators does not behaves in same way due to it's internal implementation behavior. (in case of `buffer*` it makes sense since buffered collection will emit when selector's closing though). Maybe need additional documentation for this topics in further.

blesh:
It looks like a slight improvement in perf, actually.

spec/operators/windowCount-spec.js
staltz: The problem is also that I don't think we are able to have both "no empty window in the end" and "first window starts immediately". Try to fulfill one and the other is not met.

blesh:
LGTM.

blesh:
Merged with 908ae56... thanks again, @staltz

PR #523, opened 2015-10-14
fix(fromEvent): make selector argument optional in fromEvent static m…

jeffbcross:
…ethod

This makes it consistent with the signature of the implementation.


staltz:
LGTM

blesh:
Thanks, @jeffbcross

Issue #381, opened 2015-09-23
Comprehensive marble tests for buffer


blesh:
Should duplicate the current RxJS 3/4 tests

Should adhere to our contribution guidelines and operator authoring docs.

staltz:
Should not yet be closed.

Tests from RxJS 3/4 are mostly covered. Just a few still missing.

staltz:
I can this one.

blesh:
Looks like `bufferToggle` is the last one on this list requiring tests.

staltz:
Yep

PR #525, opened 2015-10-14
test(buffer): add a few remaining tests for buffer

staltz:
This should be enough to get #381 resolved.


blesh:
Merged with 8526206 ... thanks @staltz

PR #527, opened 2015-10-14
bufferWhen: tests and fixes

staltz:
Add comprehensive marble tests for bufferWhen.

bufferWhen, as far as I can see, mirrors RxJS 4 `buffer` using a closingSelector function as parameter.
If so, then this PR fixes bufferWhen to behave like RxJS 4 buffer in some corner cases.

Relates to #381.


blesh:
Merged with 5d28a38... thanks @staltz

PR #529, opened 2015-10-14
test(filter): test the number of calls to the predicate function

luisgabriel:
relates to #390.


blesh:
Merged with 6af8f2d... thanks @luisgabriel

PR #531, opened 2015-10-14
chore(micro perf): bump up rx version to latest

kwonoj:
RxJS 4.0.6 has released, PR trying to bump up version as well.

Instead of manually picking version all time, use `latest` to pick last published version always. Somewhat debatable though thinking about recent perf test issues with `defaultIfEmpty`, if new package has regression micro perf test will be impacted directly. Still believe this'll gives convenience and expect regression occurs raraly, considering framework is quite mature already.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "platform": "^1.3.0",
     "promise": "^7.0.3",
     "protractor": "2.2.0",
-    "rx": "^4.0.1",
+    "rx": "latest",
     "tslint": "^2.5.0",
     "typescript": "^1.7.0-dev.20150901",
     "watch": "^0.16.0"


kwonoj:
Updated PR to micro perf test runner does not complains when run without any argument.

kwonoj:
Timeout failures on one of async test case
>Failures:
1) Observable.prototype.combineLatest should combineLatest the source with the provided observables
  Message:
    Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
  Stack:
    Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
        at Timer.listOnTimeout (timers.js:119:15)

I'll try to convert those test into testscheduler based with separate PR.

blesh:
rebase... I think one of the jasmine async tests that was unnecessary was freaking out.

kwonoj:
It seems so. Rebased to include latest changes and now travis says happy.

kwonoj:
https://github.com/ReactiveX/RxJS/pull/543 filled changes of micro perf test runner already, updated PR to drop those changes.

blesh:
Merged with 82b313e. Thanks @kwonoj!

PR #535, opened 2015-10-15
test(bufferTime): add missing marble tests on bufferTime

staltz:
A few marble tests mirroring RxJS 4 tests were missing.


blesh:
Merged with 8f42b1d... thanks again, @staltz!

PR #537, opened 2015-10-15
fix(buffer): cleanup notifier subscription when unsubscribed

kwonoj:
Code snippet 

```
var source = Rx.Observable.timer(0, 25).buffer(Rx.Observable.interval(50)).take(3);
var subscription = source.subscribe(
  function (x) {
    console.log('Next: %s', x);
  },
  function (err) {
    console.log('Error: %s', err);
  },
  function () {
    console.log('Completed');
  });
```

emits buffered elements like 
>Next: 0,1
Next: 2,3
Next: 4,5
Completed
###### --> But process does not exit here

This PR let buffer unsubscribes closingNotifier to avoid above issue.

### DIFF
--- a/spec/operators/buffer-spec.js
+++ b/spec/operators/buffer-spec.js
@@ -1,8 +1,8 @@
-/* globals describe, it, expect, expectObservable, hot */
+/* globals describe, it, expect, expectObservable, expectSubscriptions, hot */
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
-describe('Observable.prototype.buffer', function () {
+describe('Observable.prototype.buffer()', function () {
   it('should work with empty and empty selector', function () {
     var a = Observable.empty();
     var b = Observable.empty();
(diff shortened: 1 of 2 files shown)


staltz:
Good catch. Should we also add a test for that?

kwonoj:
Would be better, bit unsure how to design test case for this. @staltz any suggestions?

blesh:
Looks like we need to add tests for the work? Coveralls is complaining on this one, @kwonoj

kwonoj:
Yes I agree, as asked above bit unsure how to design test case for this. May I borrow some tips from you for insight?

staltz:
@kwonoj here's a draft:

```js
  it('should', function () {
    var a = hot('--1--2--^--3--4--5---6----7--8--9---0---|');
    var unsub =         '              !                  ';
    var subs =          '^             !                  ';
    var b = hot('--------^--a-------b---cd|               ');
    var bsubs =         '^             !                  '; // <-----
    var expected =      '---a-------b---                  ';
    var expectedValues = {
      a: ['3'],
      b: ['4', '5']
    };
    expectObservable(a.buffer(b), unsub).toBe(expected, expectedValues);
    expectSubscriptions(a.subscriptions).toBe(subs);
    expectSubscriptions(b.subscriptions).toBe(bsubs); // <-----
  });
```

kwonoj:
@staltz Appreciate for help! I'll try to add those test cases.

kwonoj:
Updated PR by copy-paste of @staltz 's help :grin:

staltz:
Looks good. I wonder why Coveralls is reporting less coverage (by -0.01%, negligible in my opinion).

kwonoj:
>why Coveralls is reporting less coverage

@staltz This was reason I started looking into https://github.com/ReactiveX/RxJS/pull/557, for this specific it seems because current coverage try to check `*spec.js` also.

blesh:
merged with 1b30aa9 ... thanks @kwonoj!

PR #538, opened 2015-10-15
bufferToggle: fixes and tests

staltz:


blesh:
merged with 949fa31 ... thanks @staltz ..

Looking at that bit with `context`, I wonder if it would have been better to have a sub class of `Subscription` that carried the buffer with it and managed semantics around buffer size and completion... Dunno. I think your solution is fine.

staltz:
> I wonder if it would have been better to have a sub class of Subscription that carried the buffer with it

That didn't occur to me when I was fixing, you mean it would be better in performance or in elegance? I just tried to do minimal changes in order to get tests passing.

blesh:
I'd be curious about either, honestly. But it's not a big deal

PR #542, opened 2015-10-15
fix(repeat): add additional resubscription behavior

### LINKED ISSUE #516: repeat() operator behavior
In some cases it seems resubscription does not triggered correctly. for example, 

`Rx.Observable.interval(40).take(2).repeat(2).subscribe(console.log, null, function(){console.log('completed')});`

*RxJS4*
>0
1
0
1
completed

*RxJS5*
>0 
1 
(does not complete)

kwonoj:
- repeat operator now supports resubscription behavior properly along
with infinite repeat
- add, expand test coverage
- update micro perf test case

closes #516

Tried to update operator's behavior, as well as expanding test coverage.


blesh:
@kwonoj ... can you please have your tests added in a separate commit? This is so our change log shows that tests were added. Thank you!

kwonoj:
Absolutely, will do it now.

kwonoj:
now commits are separated per feature, tests.

blesh:
Merged with a12a5ae ... thanks again, @kwonoj

PR #544, opened 2015-10-15
fix(package): correct typings path

robwormald:
non critical, but entry point for typings should point to same location as CJS output fixed in https://github.com/ReactiveX/RxJS/commit/64cd865a798804ecf31113cd7a6889689ac7eb8c - my fault, probably should have caught this on review.


blesh:
Merged with a501b06. I appreciate you catching this!

Issue #83, opened 2015-07-16
Add distinctUntilChanged operator


steveorsomethin:
Might be nice to re-think the API on this one slightly. It seems very uncommon to use the keySelector first argument. Interested in hearing anecdotes to the contrary.

staltz:
>  It seems very uncommon to use the keySelector first argument.

Agreed.

The comparer can be used to select the key, if needed. So this

```js
var obs = observable.distinctUntilChanged(x => x.id, (x, y) => x === y);
```

becomes

```js
var obs = observable.distinctUntilChanged((x, y) => x.id === y.id);
```

blesh:
closed by f9ba4da3fe26db233c7553cf070bd769bb182515

PR #545, opened 2015-10-15
test(distinctUntilChanged): add marble tests for distinctUntilChanged…

kwonoj:
… operator

relates to #309 

Added additional test coverage as well as replacing existing test into marble test.
By following issue https://github.com/ReactiveX/RxJS/issues/83, some of test case for [`keySelector`](https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/distinctuntilchanged.js#L185-L197) is dropped and not migrated.


blesh:
merged with d02ef4b... nice work.

Issue #546, opened 2015-10-16
retry and retryWhen don't call unsubscription logic when the source observable errors


blesh:
Basically this scenario:

```js
Observable.create(observer => {
  observer.error('bad');
  return () => {
    console.log('tear down');
  };
}).retryWhen(errors => Observable.timer(3000))
```

that `"tear down"` is never logged to console.

staltz:
I can do this

staltz:
This can be now closed.

PR #547, opened 2015-10-16
retry: fixes and tests

staltz:
Resolves (half of) #546.

### DIFF
--- a/spec/operators/retry-spec.js
+++ b/spec/operators/retry-spec.js
@@ -35,10 +35,8 @@ describe('Observable.prototype.retry()', function () {
       observer.complete();
     })
       .map(function (x) {
-        if ((errors += 1) < retries) {
-          throw 'bad';
-        }
-        return x;
+        errors += 1;
+        throw 'bad';
       })
       .retry(retries - 1)
       .subscribe(
@@ -46,7 +44,7 @@ describe('Observable.prototype.retry()', function () {
           expect(x).toBe(42);
         },
         function (err) {
-          expect(errors).toBe(1);
+          expect(errors).toBe(2);
           done();
         }, function () {
           expect('this was called').toBe(false);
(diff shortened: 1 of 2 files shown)


kwonoj:
What happens if resubscription occurs sufficiently large, i.e) invoked with `retry()` without param to allow indefinite retry and source fails more than several thousands (>10000 maybe) ?

kwonoj:
Though I think that's sort of extreme cases usually do not expect to happen in real.

staltz:
> What happens if resubscription occurs sufficiently large, i.e) invoked with retry() without param to allow indefinite retry and source fails more than several thousands (>10000 maybe) ? 

In marble tests there is a time limit, `maxFrames = 750`. However, if the source is `cold('#')`, then the retries will happen synchronously and the time frame will not progress at all in the TestScheduler, so it's possible to put the test in an infinite loop until the test runner breaks.

kwonoj:
Actually question was about not creating test cases, trying with actual code such as `Observable.throw().retry(LAAARGE)` (pseudo, of course) not run by testscheduler. Just curiosity though.

staltz:
With RxJS 4, `Observable.throw().retry()` (indefinite) throws the error:
```
"Maximum call stack size exceeded"
```

kwonoj:
Interesting, thanks for confirmation.

staltz:
Actually, sorry, that was RxJS 2. With RxJS 4, the page freezes. Probably doing an infinite loop.

kwonoj:
I just did quick try with RxJS4 + node myself, and behavior was same, it seems does infinite retrying. I can give a try with this changes once it's merged. Again, thanks! :)

blesh:
@staltz ... this is precisely the solution I was thinking of when I first found this issue the other day. Nice work.

blesh:
Merged with 09b475a..

Also, I had to pin `eslint` to the original version we had installed, I had wiped out my `node_modules` directory and `npm install`ed, and it installed eslint `1.7.0`, which was giving false positives on our spec files. I just pinned it to a version I knew would work with our build.

staltz:
Good!

PR #552, opened 2015-10-16
test(VirtualTimeScheduler): add test using negative delay

luisgabriel:


blesh:
Merged with f414a0b ... thanks @luisgabriel!

PR #449, opened 2015-10-02
fix(VirtualTimeScheduler): ignore negative delays

staltz:
Fix the VirtualTimeScheduler to ignore scheduled actions that have
a negative `delay`. There seems to no reason to schedule these, and all
tests still pass.

In addition, it allows us to do this in tests:
```js
    var source = hot('-1--2--^-3-4-5-6--7-8--9--|');

    var invoked = 0;
    var predicate = function (x) {
      invoked++;
      return isPrime(x);
    };

    source.filter(predicate)
      .subscribe(null, null, function () {
        expect(invoked).toEqual(7); // <---- this here
      });
```

PR #553, opened 2015-10-16
test(timer): add more test cases from RxJS 4

luisgabriel:


luisgabriel:
The following test case is not passing because the test scheduler is emitting in timeframe `-1` while the test expects it to be emitted on zero. 

```javascript
it('should emit a single value immediately when a negative value is given', function () {
    var source = Observable.timer(-1, rxTestScheduler);
    var expected = '(x|)';

    expectObservable(source).toBe(expected, {x: 0});
  });
```

I'm not sure if this test case does not make sense or if we have a bug.

spec/observables/timer-spec.js
```diff
@@ -11,4 +11,27 @@ describe('Observable.timer', function () {
...
+
+  it('should start after delay and periodically emit values', function () {
+    var source = Observable.timer(40, 20, rxTestScheduler).take(5);
+    var expected = '----a-b-c-d-(e|)';
+    var values = { a: 0, b: 1, c: 2, d: 3, e: 4};
+
+    expectObservable(source).toBe(expected, values);
+  });
+
+  it('should start immediately and periodically emit values', function () {
```
kwonoj: related with this test case, what about adding one additional case when subscription to source is done?

luisgabriel: Good idea. I'll add it.

kwonoj:
I think negative scheduling was discussed bit with https://github.com/ReactiveX/RxJS/pull/449 before, so far I know basic idea is `VirtualTimescheduler` doesn't expect to schedule negative frames. For those specific cases, I'd consider to use actual scheduler instead since it's anyway to be expected to emit value immediately. Maybe @blesh or @staltz can correct me if I understood something wrong.

luisgabriel:
I was wondering if it makes sense to accept negative delays on `timer`. Currently, it's not doing any check and passing this negative value to the scheduler. Maybe it should check if the delay is negative and set it to zero internally as this is what will actually happen (the action will be executed immediately).

blesh:
Looks like this somehow reduced test coverage, so the build is failing.

kwonoj:
While [doc](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/timer.md) of RxJS4 does not specify behavior of negative,  [test](https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/timer.js#L36-L47) seems expect to behave is as immediate, as same as `0` is specified. I'd also consider same check-fallback, or let operator complains about negative (throw maybe?), but prior to be better to match operator behavior with RxJS4.

staltz:
The handling of negative depends on the scheduler, so either way (to support or to not support negatives), we should add negative tests to scheduler tests (e.g. `TestScheduler-spec.js`).

blesh:
@staltz is right... in any non-virtual scheduler, you would treat a negative value as a `0`, but in virtual scheduler (like the TestScheduler), you actually want to allow negative scheduling.

spec/observables/timer-spec.js
luisgabriel: @kwonoj What do you think about the following test case?

```javascript
it('should stop emiting values when subscription is done', function () {
  var invoked = 0;
  var source = Observable.timer(0, 30, rxTestScheduler)
    .do(function (x) {
      expect(invoked).toBe(x);
    });

  var subscription = source.subscribe(function () {
    invoked++;
    if (invoked === 3) {
      subscription.unsubscribe();
    }
  });
});
```

luisgabriel:
Ok. PR #552 adds a test case covering negative values on VirtualTimeScheduler.

spec/observables/timer-spec.js
kwonoj: Doesn't utilizing `expectUnsubscription` works like below?

```
var e1 = Observable.timer(x, y, testScheduler); 
var expected = '..................';
var unSub = '^                    !;

expectObservable(e1, unSub).toBe(expected);
expectSubscription(e1).toBe(unSub);
```

(pseudo, haven't tried it by myself yet)

luisgabriel: Sure! Much better this way. I was not aware of the second argument of `expectObservable()`.

```javascript
it('should stop emiting values when subscription is done', function () {
  var source = Observable.timer(0, 30, rxTestScheduler);
  var expected = 'a--b--c--d--e';
  var unsub   =  '             !'
  var values = { a: 0, b: 1, c: 2, d: 3, e: 4};

  expectObservable(source, unsub).toBe(expected, values);
});
```

But I 'm not sure if we can use `expectSubscriptions()` here since the observable returned by `timer()` does not have a subscription log.

kwonoj: Yeah.. since timer's generator and there isn't subscription. Think case would work still, may better to have explicit subscription in unsub as `='^               !';`

luisgabriel:
```javascript
it('should stop emiting values when subscription is done', function () {
  var source = Observable.timer(0, 30, rxTestScheduler);
  var expected = 'a--b--c--d--e';
  var unsub   =  '^            !'
  var values = { a: 0, b: 1, c: 2, d: 3, e: 4};

  expectObservable(source, unsub).toBe(expected, values);
});
```

Like this?

kwonoj: Yes, that was I meant.

luisgabriel: Ok. I'll update the PR.

luisgabriel: Done.

blesh:
merged with bc717a4 ... thanks @luisgabriel

Issue #558, opened 2015-10-17
Strategies for more accurate code coverage report


kwonoj:
Related with PR https://github.com/ReactiveX/RxJS/pull/557

Current code coverage report (i.e, [build 823] (https://coveralls.io/builds/3868169)) is indicating coverage of source points to transpiled results from original typescript based sources. Due to this reason it does not reports accurate coverage percentage also does not allows navigation to coverage detail, makes bit hassle to triage when coverage number goes down.

##### Coverage report shows coverage against transpiled
![image](https://cloud.githubusercontent.com/assets/1210596/10557176/298d7f54-7456-11e5-9006-70d1bc73e2af.png)

##### Not able to navigate actual source detail
![image](https://cloud.githubusercontent.com/assets/1210596/10557180/449bfeb0-7456-11e5-8309-f72b2acc20f2.png)

PR https://github.com/ReactiveX/RxJS/pull/557 partially resolves this issue by enabling coverage remapping via sourcemaps. Still it doesn't give right coverage to original source but es6 transpiled source because of current build mechanism,

1. tsc compiles into ES6
2. babel uses compiled ES6 and compiles into commonjs

since test cases are using commonjs output from 2, even after enabling sourcemap remapping coverage report shows result against ES6 output from 1 instead of original source doesn't give right accuracy of test coverage along with lacks of navigation. (example report can be found at [here](https://coveralls.io/builds/3868743))

##### source map enabled
![image](https://cloud.githubusercontent.com/assets/1210596/10557235/79b2e27e-7458-11e5-92b7-251ae467deb6.png)

##### detail's still not available
![image](https://cloud.githubusercontent.com/assets/1210596/10557197/1e83fbb4-7457-11e5-83a0-14e5616601d8.png)

Could think of few options to resolve this issues,

~~1. build commonjs module via typescript compiler~~
: Not considerable due to https://github.com/ReactiveX/RxJS/issues/248
~~2. leave current build mechanism as-is, build commonjs module once again via typescript compiler only for test coverage~~
: Though this can work, still bit hacky since it requires build couple, or triple for same build.
~~3. it's ok to live with current coverage status, let's leave as-is.~~
4. Maybe some other way I couldn't think of

I have tried option 2, could create test coverage shows quite different number than what we're currently having against transpiled results. (example report can be found at [here](https://coveralls.io/builds/3869129))

##### Ideal coverage report, shows accurate coverage to source
[...]

PR #557, opened 2015-10-17
chore(code coverage): enable coverage remapping to source via sourcemap

kwonoj:
- update script enables test coverage remapping to original (es6) source
- minor update to document, code

This PR introduces [`remap-istanbul` module](https://github.com/SitePen/remap-istanbul) allows to remap test coverage into original source code instead of transpiled results. (`istanbul` side discussion is ongoing at https://github.com/gotwarlost/istanbul/issues/212) Due to current build processes, this PR does not resolves remapping issue fully though.

PR #562, opened 2015-10-19
repeat: tests and fixes

### LINKED ISSUE #554: Add tests for repeat unsubscription
We need to make sure that the source observable's tear down logic is called before everything repeats again, so we have to make sure we're unsubscribing from the previous subscription before repeating.

I don't think we have tests around this.

staltz:
Resolves #554.

### DIFF
--- a/spec/operators/repeat-spec.js
+++ b/spec/operators/repeat-spec.js
@@ -4,39 +4,85 @@ var Observable = Rx.Observable;
 
 describe('Observable.prototype.repeat()', function () {
   it('should resubscribe count number of times', function () {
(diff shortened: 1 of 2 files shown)


staltz:
Coveralls is not happy :unamused:

kwonoj:
Think https://github.com/ReactiveX/RxJS/issues/558 and https://github.com/ReactiveX/RxJS/pull/557 need to be resolved soon for futher PR's coverage check failure. For now, it seems coverage's decreased though we're adding new test cases.

blesh:
merged with f67a596 ... thanks @staltz

PR #563, opened 2015-10-19
test(concat): add subscription tests for concat

staltz:
Add tests to assert subscription/unsubscription moments for concat
operator. This also changes most tests to use cold sources instead of
hot sources, since it makes more sense to concat cold sources, because
concatenating hot sources is essentially a merge.

*This is part of my effort to review all specs and make sure we have subscription assertions.*


luisgabriel:
:+1:

blesh:
Merged with ea000e7 ... thanks @staltz

Issue #549, opened 2015-10-16
supporting absoluteTime, selector for delay() operator?


kwonoj:
Current implementation of `delay` operator supports only relative time delay as stated in signature

`delay(delay: number, scheduler: Scheduler)`

compare to RxJS4, which supports `absoluteTime` also `delayDurationSelector`.

would like to ask if it's planned to expand current implementation to support rest of behaviors.

Personally, since it has somewhat polymorphic behavior would like to separate surface such as below (course function name could be different)

```
delay(delay: number | Date, scheduler: Scheduler)
delayWhen(delayDurationSelector: function, subscriptionDelay:Observable)
```

kwonoj:
Would like to take this one if this reaches agreement, since I'm working on expanding marble test for this operator.

blesh:
That seems reasonable. I'm on mobile, does delayWhen align with bufferWhen? If so, it sounds good to me

kwonoj:
somewhat similar, but not exactly. Major difference is `bufferWhen` accepts observable directly as closing selector, while above `delayWhen` simply mimic RxJS4 signature to accept `function` as selector. If it's to be aligned, signature will be looks like

`delayWhen(delayDurationSelector: Observable, subscriptionDelay?:Observable)`

not sure if selector observable can get rid of `subscriptionDelay` completely.

luisgabriel:
`Observable.timer()` in RxJS 4 also accepts absolute time: https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/timer.md

Should we add this behavior to RxJS Next?

blesh:
> Should we add this behavior to RxJS Next?

@luisgabriel yes we should... closing this issue though because it's done. I'll open another.

PR #566, opened 2015-10-19
fix(delay): accepts absolute time delay

kwonoj:
- delay operator accepts absolute time as well as relative time
- add additional test coverages

relates to #549

### DIFF
--- a/spec/operators/delay-spec.js
+++ b/spec/operators/delay-spec.js
@@ -1,11 +1,96 @@
-/* globals describe, it, expect, expectObservable, hot, rxTestScheduler */
+/* globals describe, it, expect, expectObservable, expectSubscriptions, hot, rxTestScheduler */
 var Rx = require('../../dist/cjs/Rx');
+var Observable = Rx.Observable;
 
 describe('Observable.prototype.delay()', function () {
   it('should delay by specified timeframe', function () {
-    var source = hot('--a--|');
-    var expected =   '-----a--|';
+    var e1 =   hot('--a--b--|');
+    var expected = '-----a--b--|';
+    var subs =     '^          !';
 
-    expectObservable(source.delay(30, rxTestScheduler)).toBe(expected);
+    expectObservable(e1.delay(30, rxTestScheduler)).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(subs);
+  });
+
+  it('should delay by absolute time period', function () {
+    var e1 =   hot('--a--b--|');
+    var expected = '-----a--b--|';
+    var subs =     '^          !';
+    var absoluteDelay = new Date(rxTestScheduler.now() + 30);
+
+    expectObservable(e1.delay(absoluteDelay, rxTestScheduler)).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(subs);
+  });
+
+  it('should delay by absolute time period after subscription', function () {
+    var e1 =   hot('---^--a--b--|');
+    var expected =    '------a--b--|';
+    var subs =        '^           !';
+    var absoluteDelay = new Date(rxTestScheduler.now() + 30);
+
+    expectObservable(e1.delay(absoluteDelay, rxTestScheduler)).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(subs);
+  });
+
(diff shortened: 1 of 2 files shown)


spec/operators/delay-spec.js
```diff
@@ -1,11 +1,79 @@
...
+
+  it('should raise error when source raises error by absolute time period', function () {
+    var e1 =   hot('--a--b--#');
+    var expected = '-----a--b--#';
+    var absoluteDelay = new Date(rxTestScheduler.now() + 30);
+
+    expectObservable(e1.delay(absoluteDelay, rxTestScheduler)).toBe(expected);
+  });
+
+  it('should raise error when source raises error by absolute time period after subscription', function () {
```
kwonoj: This behavior is bit different to RxJS4 [test case] (https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/delay.js#L156-L256), in RxJS4 test case is described as

```
source  = '---a---b---#'
delayed = '----a----b-#'; 
delayed = hot(source).delay(10);
```

that `error` is emitted as soon as source emits. Per [documantation](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/delay.md), behavior is defined as 

`Time shifts the observable sequence by dueTime. The relative time intervals between the values are preserved.`
think `error` should be considered as same as normal value element, but maybe that's something I understood behavior incorrectly(or either read RxJS4 test cases incorrectly?), asking confirmation. @blesh @staltz

spec/operators/delay-spec.js
```diff
@@ -1,11 +1,79 @@
...
+    var e1 =   hot('---^---a---b---#');
+    var expected =    '-------a---b---#';
+    var absoluteDelay = new Date(rxTestScheduler.now() + 30);
```
kwonoj: This behavior is bit different to RxJS4 [test case] (https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/delay.js#L156-L256), in RxJS4 test case is described as

```
source  = '---a---b---#'
delayed = '----a----b-#'; 
delayed = hot(source).delay(10);
```

that `error` is emitted as soon as source emits. Per [documantation](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/delay.md), behavior is defined as 

`Time shifts the observable sequence by dueTime. The relative time intervals between the values are preserved.`
think `error` should be considered as same as normal value element, but maybe that's something I understood behavior incorrectly(or either read RxJS4 test cases incorrectly?), asking confirmation. @blesh @staltz

staltz: IMO errors should be reported ASAP because they mean "unhandled exceptions".

kwonoj: OK, then I can consider this is my-incorrect-understanding, right?

spec/operators/delay-spec.js
```diff
@@ -1,11 +1,79 @@
...
+    var e1 =   hot('---^--a--b--|');
+    var expected =    '------a--b--|';
+    var absoluteDelay = new Date(rxTestScheduler.now() + 30);
+
+    expectObservable(e1.delay(absoluteDelay, rxTestScheduler)).toBe(expected);
+  });
+
+  it('should raise error when source raises error', function () {
+    var e1 =   hot('---a---b---#');
+    var expected = '------a---b#';
```
kwonoj: quickly mended `error` behavior per suggested. @staltz , whip me if I did something incorrect still.

staltz: Since you're at it, how about adding subscription assertions too?

kwonoj: subscription for sources, am I understanding correct? i.e for this testcase,

```
var e1 =   hot('---a---b---#');
var e1Sub =    '           !';
....
```

staltz: `var e1Sub = '^         !';` it was missing `^` to indicate when it's subscribed.

kwonoj: Yeah, too hesitate to write down stubs. Thanks for suggestion, trying to update it.

blesh:
Looks good.

blesh:
merged with b109100... thanks @kwonoj

PR #573, opened 2015-10-21
test(combineLatest): add subscription assertions

staltz:
Add tests for combineLatest (proto and static) that assert the
subscription and unsubscription times of the inner Observables.

### DIFF
--- a/spec/observables/combineLatest-spec.js
+++ b/spec/observables/combineLatest-spec.js
@@ -28,4 +28,397 @@ describe('Observable.combineLatest', function () {
       done();
     });
   });
+
+  it('should work with two nevers', function () {
+    var e1 = cold( '-');
+    var e1subs =   '^';
+    var e2 = cold( '-');
+    var e2subs =   '^';
+    var expected = '-';
+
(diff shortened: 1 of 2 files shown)


spec/observables/combineLatest-spec.js
```diff
@@ -28,4 +28,397 @@ describe('Observable.combineLatest', function () {
       done();
     });
   });
```
staltz: The following tests are just copy-pasted from tests on the operator version of combineLatest. What do people think about this? On one hand we need to test whether the static combineLatest works, but on the other hand it's really duplication of code to-be-maintained.

kwonoj: Personal preference is code to be strictly guarded even there is some of duplication may occurs. Course it'd be better if it's possible to have common fixture setup to reduce duplication..

blesh:
I think it's totally fine, and almost preferable to write tests that cover the same code twice when that code is exposed via two separate points on the API. Things change, and I'd hate for something to change with the static operator, cause breakage, and not be caught because we didn't have tests around it.

blesh:
merged with 7ca7857 thanks, @staltz

PR #575, opened 2015-10-21
test(buffer): add micro-benchmark performance test for buffer operator

kwonoj:
relates to #350 

PR https://github.com/ReactiveX/RxJS/pull/537 might need to be merged as well to make this test actually works.


blesh:
Merged with 89c750b .. thanks @kwonoj

PR #576, opened 2015-10-22
chore(tsconfig.json): update tsc build configuration

kwonoj:
- tsconfig points es6 build configuration by default

Though tsconfig.json is not being used via actual build script, it's no harm to align it with build script for anyone who want simply invoke `tsc` to get es6 build. One of minor changes while waiting PR related code coverage check is resolved, to avoid PR created with failed check.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.sublime-project
 *.sublime-workspace
 .settings
+.vscode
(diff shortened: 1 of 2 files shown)


blesh:
Merged with bccddb6 .. thanks @kwonoj

PR #585, opened 2015-10-23
fix(operator): update type definitions for union types

### LINKED ISSUE #581: Type definition complains for some operators with union type
For example, signature of `Observable.of` is defined as below

`Rx.Observable.of(value, [scheduler])`

and declaration is like below, accepts value or scheduler with union type definition.

`static of<T>(...array: (T | Scheduler)[]): Observable<T> `

which makes eventually type inference confusions like
image

since array now contains value:T and scheduler:Scheduler both.

While this doesn't hurt functionality at all, would be better to have way to suppress these type of complains.

kwonoj:
- allow multiple type parameters can be accepted for observable,
operator supports array of union types

closes #581

This PR replaces array of union type declaration more explicit way by language spec (https://github.com/Microsoft/TypeScript/issues/805), get rid of type inference error by typescript compiler when multiple type of parameters are actually provided.

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -133,8 +133,8 @@ export default class Observable<T> implements CoreOperators<T>  {
   }
 
   // static method stubs
-  static combineLatest: <T>(...observables: (Observable<any> | ((...values: Array<any>) => T) | Scheduler)[]) => Observable<T>;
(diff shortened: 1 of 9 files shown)


luisgabriel:
:+1:

blesh:
Merged with 9d90c75... thanks @kwonoj

PR #587, opened 2015-10-23
takeUntil: fixes and tests

staltz:


src/operators/takeUntil.ts
```diff
@@ -3,41 +3,56 @@ import Observer from '../Observer';
...
+  _complete() {
+    this.destination.complete();
+    this.notificationSubscriber.unsubscribe();
   }
 }
 
 class TakeUntilInnerSubscriber<T> extends Subscriber<T> {
-  constructor(destination: Subscriber<T>) {
-    super(destination);
+  constructor(public destination: Subscriber<T>) {
```
kwonoj: does `destination` need to be public here?

src/operators/takeUntil.ts
```diff
@@ -3,41 +3,56 @@ import Observer from '../Observer';
...
 class TakeUntilSubscriber<T> extends Subscriber<T> {
-  constructor(destination: Subscriber<T>,
-              observable: Observable<any>) {
-    super(destination);
-    this.add(observable._subscribe(new TakeUntilInnerSubscriber(destination)));
+  private notificationSubscriber: TakeUntilInnerSubscriber<any> = null;
+
+  constructor(public destination: Subscriber<T>,
+              private notifier: Observable<any>) {
+    super(null);
```
kwonoj: This is bit interesting for me - may I possibly ask some additional details for reason parent `Subscriber` doesn't handle destination for these cases? (xxxUntil)

src/operators/takeUntil.ts
```diff
@@ -3,41 +3,56 @@ import Observer from '../Observer';
...
 }
 
 class TakeUntilSubscriber<T> extends Subscriber<T> {
-  constructor(destination: Subscriber<T>,
-              observable: Observable<any>) {
-    super(destination);
-    this.add(observable._subscribe(new TakeUntilInnerSubscriber(destination)));
+  private notificationSubscriber: TakeUntilInnerSubscriber<any> = null;
+
+  constructor(public destination: Subscriber<T>,
```
kwonoj: same as 44 - does `destination` need to be public?

src/operators/takeUntil.ts
staltz: Yeah, if I'm not mistaken, `private destination` would raise a TypeScript compiler I think because the super class has destination defined as public.

src/operators/takeUntil.ts
staltz: `super(destination)` calls the constructor of `Subscriber`, which sets `this._subscription`. If `this._subscription` is set, when `Subscriber::unsubscribe()` is called it does this:
```ts
unsubscribe() {
    if(this._isUnsubscribed) {
      return;
    } else if(this._subscription) {
      this._isUnsubscribed = true;
    } else {
      super.unsubscribe();
    }
  }
```
And as you can see the actual unsubscribe wouldn't happen.

src/operators/takeUntil.ts
kwonoj: Yes, ts will complain `private` due to parent, but still `protected` will give more encapsulated scope.

staltz: Alright

src/operators/takeUntil.ts
kwonoj: Thanks for clarification, it's interesting. So for operator behavior strategies, is it possible to determine when it's available to lean on parent subscriber handles subscription vs. inherited operator specific subscriber need to handle? subscription test case is best way for those?

staltz: I don't understand your question

kwonoj: Sorry, seems I wasn't clear enough. 

Let's assume I write new operator `awesomexxx`. When I compose it, is there way I can determine if inherited subscriber need to handle subscription by itself instead calling subscriber's ctor (`super(destination)`) before actually verify it's behavior by writing subscription test cases? 

I was thinking about this can be some kind of 'composing operator guideline / references' in further.

staltz: Yeah I'm not sure what should be done, I'm not particularly happy with all the inheritance going on (`OperatorSubscriber ~> Subscriber ~> Subscription`), I was just replicating a pattern I've seen Ben and others apply. For instance, in `buffer`. I don't actually understand the rationale behind using `super(destination)`.

trxcllnt: @staltz the `Subscriber` constructor accepts the `destination` in order to reference the shared "composite" Subscriber.

blesh:
@staltz which test doesn't pass if you call `super(destination)`? I remember something like this coming up with another operator recently and discussing it with @trxcllnt, but I want to say in that case it was some inner subscription that was the issue, not the outer subscription.

staltz:
@blesh Actually in the case of `takeUntil` no tests fail if I use `super(destination)`. I used `super(null)` just to stay symmetric with skipUntil which needs `super(null)`. I'll update this PR to use `super(destination)` but let's talk about the problems in the skipUntil PR #586 .

staltz:
Actually @blesh you might want to unblock and merge this one since I updated it so the outer subscriber calls `super(destination)` and only the inner subscriber has `super(null)`. Or do you think also the inner one should call `super(destination)`?

blesh:
>  I updated it so the outer subscriber calls super(destination) and only the inner subscriber has super(null). Or do you think also the inner one should call super(destination)?

That sounds more correct. I'll review again in a bit.

blesh:
@staltz ... you're going to have to do something similar with #586

blesh:
merged with 9415196 ... nice work, @staltz

Issue #453, opened 2015-10-02
publishBehavior and publishReplay semantics when completed


staltz:
Continuing the discussion that started here: https://github.com/ReactiveX/RxJS/pull/235/files

RxJS Legacy (v4) passes these tests:
```js
var results1 = [];
var results2 = [];
var subscriptions = 0;

var source = Rx.Observable.create(function (observer) {
  subscriptions++;
  observer.onNext(1);
  observer.onNext(2);
  observer.onNext(3);
  observer.onNext(4);
  observer.onCompleted();
});

var hot = source.shareValue(0);

hot.subscribe(x => results1.push(x));

expect(results1).toBe([0,1,2,3,4]);
expect(results2).toBe([]);

hot.subscribe(x => results2.push(x));

expect(results1).toBe([0,1,2,3,4]);
expect(results2).toBe([]);
expect(subscriptions).toBe(2);
```
and
```js
var results1 = [];
var results2 = [];
var subscriptions = 0;

var source = Rx.Observable.create(function (observer) {
  subscriptions++;
  observer.onNext(1);
  observer.onNext(2);
  observer.onNext(3);
  observer.onNext(4);
  observer.onCompleted();
});

var hot = source.shareReplay(2);

hot.subscribe(x => results1.push(x));

expect(results1).toBe([1,2,3,4]);
expect(results2).toBe([]);

hot.subscribe(x => results2.push(x));

expect(results1).toBe([1,2,3,4]);
expect(results2).toBe([3,4]);
expect(subscriptions).toBe(2);
```

Yet RxJS Next behaves basically the opposite way:
```
Failures:
1) Observable.prototype.publishBehavior() should not emit next events to observer after completed
  Message:
    Expected [ 0 ] to equal [  ].
  Stack:
    Error: Expected [ 0 ] to equal [  ].
      at Object.<anonymous> (./RxJSNext/spec/operators/publishBehavior-spec.js:65:21)
2) Observable.prototype.publishReplay() should emit replayed events to observer after completed
  Message:
    Expected [  ] to equal [ 3, 4 ].
  Stack:
    Error: Expected [  ] to equal [ 3, 4 ].
      at Object.<anonymous> (./RxJSNext/spec/operators/publishReplay-spec.js:98:21)
145 specs, 2 failures
Finished in 12.951 seconds
```

Because in RxJS legacy there is
```js
    ConnectableObservable.prototype._subscribe = function (o) {
      return this._subject.subscribe(o);
    };
```

[...]

Issue #298, opened 2015-09-11
Too many sharing operators? Should we cull the herd?


blesh:
In a hallway discussion with @jhusain, we were talking about how most of the time people really just want `share` or `shareReplay` or `shareBehavior` (or the like), and rarely do they actually want to just `publish` something without that immediate `refCount` call.

Some of these operators are yet to implement, but the list is considerable when you realize they're all specializations of the same thing:

- `publish().refCount()`
- `publishReplay().refCount()`
- `publishBehavior().refCount()`
- `publishValue().refCount()`
- `share()`
- `shareReplay()`
- `shareBehavior()`
- `multicast(subjectFactory).refCount()`
- `publish()` then `connect()`
- `publishReplay()` then `connect()`
- `publishBehavior()` then `connect()`
- `multicast(subjectFactory)` then `connect()`
- `publish().connectionStrategyThing` - (tbd @trxcllnt proposed operator for defining connection and disconnection strategy)

### The problems

1. It's bloat, there are just too many operators that do pretty much the same thing.
2. Most of them aren't very straight forward to teach.
3. Many of them aren't used enough to justify a specialization.
4. The ones that are used frequently are always used in the same ways, but rarely used alone. (publish is  almost always paired with refCount, etc)

### My pitch

I think we should pare this back to being a refined set of operators that falls into two categories:

1. The most common combined operators (`share` rather than `publish().refCount()` for example)
2. The most basic operators that allow you to build any of the others `multicast` and `refCount` for example

### To cull or simply exclude?

An alternative to cutting them completely from the library is to keep them in the library, but not include them in the global output. People could simply use them from their modules.

attn @staltz who was a big proponent of `publishBehavior` (which is honestly one I'd consider cutting in favor of `shareBehavior`)

blesh:
I'd also appreciate the thoughts of @mattpodwysocki, who undoubtedly has faced more internal client needs (or dislikes) for the presence of these operators.

staltz:
I agree with the sentiment there are too many of them (I know I added yet more, but that was in an attempt to standardize them).

I have also been teaching these operators to developers, and they mostly prefer `share()` because it's so intuitive and has a clear intent. I sometimes use `publish().refCount()` just to emphasize the refCount and teach how it works under the hood.

Agreed that `share`, `shareBehavior`, `shareReplay` (**) should be the main ones to teach and advocate. (** although shareReplay is getting less useful, and ReplaySubject in general).

[...]

PR #588, opened 2015-10-23
feat(shareBehavior): add shareBehavior and its tests

staltz:


blesh:
This is bike-shedding, but wasn't this called "shareValue" before? Is that a better term? What would make the most sense for people new to the library? (FWIW: I found the name of "BehaviorSubject" to be essentially nonsense when I first started using Rx)

blesh:
Also, wasn't there a problem with `shareBehavior`, @staltz? Or was that only for `shareReplay`? (The problem being that it creates a new Subject in it's multicast)

staltz:
> This is bike-shedding, but wasn't this called "shareValue" before? Is that a better term?

It was, but we [discussed here and reached a consensus](https://github.com/ReactiveX/RxJS/issues/298#issuecomment-139609578) to align the sharing names to their respective subjects, for symmetry: 
- share uses Subject
- shareReplay uses ReplaySubject
- shareBehavior uses BehaviorSubject

To rename it to shareValue would mean also renaming BehaviorSubject to ValueSubject. I'm not contesting that actually. My brain got used to BehaviorSubject as a name, and perhaps to people migrating to v5 it would be easier to keep the same names, but I actually agree that ValueSubject would be more intuitive.

> Also, wasn't there a problem with shareBehavior, @staltz? Or was that only for shareReplay?

Both shareBehavior and shareReplay "have problems" ([this issue](https://github.com/ReactiveX/RxJS/issues/453)), and these tests are not covering those problems yet. So this PR here is agnostic to whatever happens to our decision regarding issue #453. When that's decided, we need to add more tests for these sharing operators.

blesh:
Okay, thanks for the clarification. Looking over this in more detail it looks fine. Sorry, I was going through 30+ PRs or something yesterday. LOL.

blesh:
merged with 97ff1ec, thanks @staltz

PR #589, opened 2015-10-23
test(mergeAll): add subscription/unsubscription assertions

staltz:


blesh:
Merged with 9f1722b ... thanks @staltz

PR #595, opened 2015-10-24
fix(NextTickAction): fix unsubscription behavior

### LINKED ISSUE #582: 'immediate' Observale.interval failure
Snippet below (duplication is intended, single execution works fine)

```
Rx.Observable.interval(0).take(6).subscribe(console.log);
Rx.Observable.interval(0).take(6).subscribe(console.log);
```

leads into unexpected failure in nexttick scheduler like below.

```
\NextTickAction.js:39
                _this.scheduler.scheduled = false;
                                          ^
TypeError: Cannot set property 'scheduled' of undefined
    at Immediate._onImmediate (Z:\github\RxJS\dist\cjs\schedulers\NextTickAction.js:39:43)
    at processImmediate [as _immediateCallback] (timers.js:367:17)
```

Creating issue while try to triage root cause for this.

kwonoj:
- fix clearImmediate behavior prevents nextTickScheduler throws if it's
being used multiple time

closes #582

### DIFF
--- a/src/schedulers/NextTickAction.ts
+++ b/src/schedulers/NextTickAction.ts
@@ -4,8 +4,7 @@ import ImmediateAction from './ImmediateAction';
 import Action from './Action';
 
 export default class NextTickAction<T> extends ImmediateAction<T> {
-
-  id: number;
+  private id: any;
 
   schedule(state?: any): Action {
     if (this.isUnsubscribed) {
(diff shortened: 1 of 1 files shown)


src/schedulers/NextTickAction.ts
```diff
@@ -4,8 +4,7 @@ import ImmediateAction from './ImmediateAction';
 import Action from './Action';
 
 export default class NextTickAction<T> extends ImmediateAction<T> {
-
-  id: number;
+  private id: any;
```
kwonoj: revised typeof `id` which wasn't number type.

blesh:
A strange oversight. Good catch. We need to add more tests around the schedulers to be sure.

blesh:
Merged with 3d8264c

PR #596, opened 2015-10-25
Add max and min operators

adrianomelo:
they are using reduce to select the min and max elements. maybe a similar approach can be used for sum and average.

### DIFF
--- a/src/operators/reduce-support.ts
+++ b/src/operators/reduce-support.ts
@@ -1,0 +1,56 @@
+import Operator from '../Operator';
+import Subscriber from '../Subscriber';
+
+import tryCatch from '../util/tryCatch';
+import {errorObject} from '../util/errorObject';
+
+export class ReduceOperator<T, R> implements Operator<T, R> {
+
+  acc: R;
+  project: (acc: R, x: T) => R;
+
+  constructor(project: (acc: R, x: T) => R, acc?: R) {
+    this.acc = acc;
+    this.project = project;
+  }
+
+  call(subscriber: Subscriber<T>): Subscriber<T> {
+    return new ReduceSubscriber(subscriber, this.project, this.acc);
+  }
+}
+
+export class ReduceSubscriber<T, R> extends Subscriber<T> {
+
+  acc: R;
+  hasSeed: boolean;
+  hasValue: boolean = false;
+  project: (acc: R, x: T) => R;
+
+  constructor(destination: Subscriber<T>, project: (acc: R, x: T) => R, acc?: R) {
+    super(destination);
+    this.acc = acc;
+    this.project = project;
+    this.hasSeed = typeof acc !== 'undefined';
+  }
+
+  _next(x) {
+    if (this.hasValue || (this.hasValue = this.hasSeed)) {
+      const result = tryCatch(this.project).call(this, this.acc, x);
+      if (result === errorObject) {
+        this.destination.error(errorObject.e);
+      } else {
+        this.acc = result;
+      }
+    } else {
+      this.acc = x;
+      this.hasValue = true;
+    }
+  }
+
+  _complete() {
+    if (this.hasValue || this.hasSeed) {
+      this.destination.next(this.acc);
+    }
+    this.destination.complete();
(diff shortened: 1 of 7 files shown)


src/Observable.ts
```diff
@@ -187,10 +187,12 @@ export default class Observable<T> implements CoreOperators<T>  {
   map: <T, R>(project: (x: T, ix?: number) => R, thisArg?: any) => Observable<R>;
   mapTo: <R>(value: R) => Observable<R>;
   materialize: () => Observable<any>;
+  max: () => Observable<any>;
   merge: (...observables: any[]) => Observable<any>;
   mergeAll: (concurrent?: any) => Observable<any>;
   mergeMap: <R>(project: ((x: T, ix: number) => Observable<any>), projectResult?: (x: T, y: any, ix: number, iy: number) => R, concurrent?: number) => Observable<R>;
   mergeMapTo: <R>(observable: Observable<any>, projectResult?: (x: T, y: any, ix: number, iy: number) => R, concurrent?: number) => Observable<R>;
+  min: () => Observable<any>;
```
kwonoj: is it expected signature does not match with actual implementation?

`min<T, R>(predicate?: (x: R, y: T) => R, thisArg?: R): Observable<R>`

src/operators/reduce-support.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import tryCatch from '../util/tryCatch';
+import {errorObject} from '../util/errorObject';
+
+export class ReduceOperator<T, R> implements Operator<T, R> {
+
+  acc: R;
+  project: (acc: R, x: T) => R;
+
+  constructor(project: (acc: R, x: T) => R, acc?: R) {
```
kwonoj: field can be assigned directly via ctor like

`ctor(private value1 : any)`

src/Observable.ts
adrianomelo: Yes, it is.. And I realized that this signature doesn't make sense. I'm using `thisArg` as a `seed` for reduce. Maybe a better signature is `min<T, R>(predicate?: (x: R, y: T) => R): Observable<R>`

kwonoj: As RxJS4 does, `comparer` seems bit more clear context for this operator.

adrianomelo: I agree :)

src/operators/reduce-support.ts
adrianomelo: @kwonoj I think that it's better leave this change to a new pull request. I didn't changed the code, just moved to a new file. There are some other minor changes such as a better signature, `seed` instead of `acc`. What do you think?

kwonoj: In my case I do apply small refactoring like this when I create PR, but if there's chunk of changes can be created separate PR that seems also way. Think @blesh can advise way feasible for maintaining repo history.

blesh: > I think that it's better leave this change to a new pull request. I didn't changed the code, just moved to a new file. 

This is preferable.

blesh:
Looks good, and I'm glad to have these operators.

I think, though, that since both of these operators can be accomplished trivially with the `reduce` operator, we can have them only exist in the `KitchenSink` output for now, and they don't need to be "core".

(Don't worry, I suspect the prevalence of transpiling and the ES7 Function Bind feature are going to make the whole "core" vs "kitchen sink" thing a solved problem)

adrianomelo:
@blesh No problem.. My goal with this patch was understand how the project is structured. So, I chose an easy to implement operator.

blesh:
@adrianomelo ... just let me know when this PR is updated. You'll also probably need to rebase, it looks like there are merge conflicts.

Thanks again for your hard work! I appreciate it.

adrianomelo:
@blesh this PR is now updated.
Thank you for helping me out reviewing code and for maintaining the project. You guys are doing an amazing work.

blesh:
@adrianomelo it is my distinct pleasure.

Your PR is merged with 79cb6cf ... thanks for your hard work!

PR #597, opened 2015-10-25
refactor(Subscriber): update type declaration for Subscriber

kwonoj:
- Subscriber applies generic type constraint

`Subscriber` itself does have type constraint via generic, it actually doesn't work for values or creating new observables, like below code snippet 

```
Rx.Observable.create((observer: Rx.Subscriber<number>) => {
    observer.next(42);
    observer.next('a');
    observer.complete();

    return () => {
        console.log('disposed');
    };
})
```

should be caught by compiler since type constraint of given subscriber is `number`, but it doesn't by `Subscriber::next()` does not apply type constraint for values.

**type constraint is not applied for values**
image

PR makes `Subscriber` now follows given types more explicitly let compiler allows type checking for build time.

**type constraint applied after changes**
image

This change also brings correct generic type checking for operators either operator passes same type or either different type like `<in, out>`. No functional changes are introduced, change affects only for type declaration.

### DIFF
--- a/src/OuterSubscriber.ts
+++ b/src/OuterSubscriber.ts
@@ -1,17 +1,16 @@
 import InnerSubscriber from './InnerSubscriber';
 import Subscriber from './Subscriber';
 
-
 export default class OuterSubscriber<T, R> extends Subscriber<T> {
-  notifyComplete(inner?: InnerSubscriber<T, R>) {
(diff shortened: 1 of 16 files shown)


luisgabriel:
:+1:

blesh:
Merged with 0329b50... thanks, as always @kwonoj!

PR #599, opened 2015-10-26
Add full tests for proto zip, static zip, and zipAll

staltz:


blesh:
Conflicts resolved and merged with 381d5bf ... thanks @staltz

PR #600, opened 2015-10-26
test(of): add a few more Observable.of tests

staltz:
Add tests to assert Observable.of's behavior with higher-order
Observables and the return of an ArrayObservable.


luisgabriel:
:+1:

blesh:
merged with 1d3f626. Thanks @staltz!

PR #602, opened 2015-10-27
refactor(src): remove unused references

kwonoj:
Clear most of unused references. 

[These variables](https://github.com/ReactiveX/RxJS/blob/master/src/util/root.ts#L24-L25) are only remaining unused references - wasn't sure if those can be safely removed or prepared for further usage. There are tslint rule `no-unused-variable` prevents further codebase to include unused, but to introduce rules remaining reference need to be clarified & resolved.

### DIFF
--- a/src/observables/IteratorObservable.ts
+++ b/src/observables/IteratorObservable.ts
@@ -173,10 +173,6 @@ function numberIsFinite(value) {
   return typeof value === 'number' && root.isFinite(value);
 }
 
-function isNan(n) {
-  return n !== n;
-}
-
 function sign(value) {
   let valueAsNumber = +value;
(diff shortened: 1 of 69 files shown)


src/observables/IteratorObservable.ts
```diff
@@ -173,10 +173,6 @@ function numberIsFinite(value) {
   return typeof value === 'number' && root.isFinite(value);
 }
 
-function isNan(n) {
-  return n !== n;
-}
```
kwonoj: Is there any particular usecase for this function? It's currently removed since there isn't any reference but functional behavior seems bit weird, curious if there is specific intended purpose.

PR #604, opened 2015-10-27
mergeMapTo: fixes and tests

staltz:
There was a rather embarrassing bug inside mergeMapTo, which should now be fixed. Also there were zero tests for this operator.


blesh:
Merged with 6f9859e... thanks @staltz

Issue #608, opened 2015-10-27
Update type declarations in codebase


kwonoj:
Such as PR https://github.com/ReactiveX/RxJS/pull/585, https://github.com/ReactiveX/RxJS/pull/597, some of type declaration signature in codebase need to be checked & revised if required.

kwonoj:
`repeat` also need to be updated, does not have default value for parameter.

![image](https://cloud.githubusercontent.com/assets/1210596/10953635/ea3bc228-82fd-11e5-8418-119170200f8b.png)

blesh:
Since this is such a generic issue, I'm going to close it for now. Ideally, we'll find more specific issues around typing and file them individually for specific files or something. "Update types in the whole library" isn't something easily resolved in a few commits.

PR #609, opened 2015-10-27
refactor(Subject): update type declaration for Subject

kwonoj:
relates to #608.


blesh:
merged 61e173b... thanks @kwonoj

PR #610, opened 2015-10-27
style(reduce): change signatures and parameter name

adrianomelo:
The second parameter of reduce isn't the accumulator, but the seed.
This commit change `acc` name to `seed`, use a shorthand for parameter properties
and fix some signature types.


blesh:
legit.

blesh:
merged as of e96d810... thanks @adrianomelo

Issue #607, opened 2015-10-27
Expand test coverage for base / utility implementations?


kwonoj:
Now test coverage reflects actual source code's test coverage shows most of operator implementation is being covered. Most of 'missing' coverage comes from underlying base, or utility function implementations such as `src\util\*` - Is it recommended to create test cases over it? It'll be pretty much basic test cases for functionality though.

staltz:
I'd say yes. That's just unit tests.

kwonoj:
Personally I thought so too, reason to create issue to get some agreement :)

blesh:
I agree 100%. I was thinking about just this. The Scheduling types are also woefully under-tested last I looked as well.

blesh:
As this is such a generic issue. I'm closing it. Ideally more specific issues will be filed around specific types that need more testing.

PR #616, opened 2015-10-28
test(bindCallback): add test cases for bindCallback

kwonoj:
relates to #607 

Effort to expand test coverage over utility implementations.


blesh:
merged with a5cb402... thanks @kwonoj!

PR #617, opened 2015-10-29
util/Immediate tests

blesh:
- refactor util/Immediate to make it testable.
- write all sorts of crazy tests to cover it thoroughly


blesh:
Still more to do here, I'm afraid.

PR #618, opened 2015-10-29
fix(take): deal with total <= 0 and add tests

adrianomelo:


adrianomelo:
@blesh, I like to be consistent :) #626

(PR is updated)

PS: I would like to comment on your previous review.. Is there a way to use the comments of a previous PR revision?

blesh:
> PS: I would like to comment on your previous review.. Is there a way to use the comments of a previous PR revision?

I'm not sure. I've not tried that. I think if you can look back in your commit history, unless it's been rewritten you can do that. Or if you have the original SHA, maybe?

blesh:
Merged with c5cc06f, thanks again, @adrianomelo!

PR #621, opened 2015-10-29
Static concat: fixes and tests

staltz:


src/operators/concat-static.ts
```diff
@@ -15,7 +15,6 @@ export default function concat<R>(...observables: Array<Observable<any> | Schedu
   let args = <any[]>observables;
   if (typeof (args[observables.length - 1]).schedule === 'function') {
     scheduler = args.pop();
-    args.push(1, scheduler);
```
staltz: Very weird line of code here, pushing the number `1` into `args` array and pushing `scheduler` back right after having popped it out from `args`.

kwonoj: Seems obviously bit weird. @blesh , was there some sort of reason? any kind of optimizations, or js runtime specific issues maybe?

blesh: It's because the lines below it were an oversight... I was building arguments to pass to `mergeStatic.apply`, then it looks like I fell asleep in my chair and forgot to do the rest.

Below, we should _not_ be calling `.mergeAll` on the observable itself, because due to modularity, it might not be there, rather, we should calling the static `merge` function with apply, which we should be importing

Something like:

```js
return mergeStatic.apply(this, args);
```

*THAT* is why I was building and maintaining args.

blesh: I can apply the fix and merge it.

kwonoj: OK, thanks.

blesh:
Merged with b9b37ae ... refactored the code to make the proper calls as well.. thanks for pointing that out, @staltz. I clearly dropped the ball there.

staltz:
Ok, cool

PR #622, opened 2015-10-29
test(every): add more tests on the every operator

staltz:
Add tests against the every operator in order to complete code coverage. Add tests regarding
predicates that throw, and sources which are SourceObservable or ArrayObservable.


blesh:
Merged with 6e82f27... thanks @staltz!

PR #624, opened 2015-10-29
test(first): expand test coverage for first operator

kwonoj:
relates to #607 

Journey to reach 99.99% code coverage.


blesh:
@kwonoj .. this is one of those that is experiencing the weird Jasmine problem where something about adding just one more test causes it to throw fail tests in ways that don't make any sense.

I suspect the test you added is making Jasmine choke for now for some reason.

blesh:
@kwonoj I take that back, the error is in the test that was added. Looking into it now.

blesh:
@kwonoj ... it was just a bad test, I have the fix, and I'm going to go ahead and merge it.

blesh:
Merged with 0a042c1... thanks @kwonoj!

PR #626, opened 2015-10-29
refactor(repeat): keep consistency on EmptyObservable instanciation

adrianomelo:
`take` uses `new EmptyObservable` instead of `EmptyObservable.create` to
instanciate a new empty observable. This minor change is to keep the consistency
between `repeat` and `take`.


blesh:
Merged with 958eeaa! Thanks again, @adrianomelo! I really appreciate this sort of thing. It's hard to find tiny inconsistencies like this in a large library.

PR #615, opened 2015-10-28
Engine setting: This library works with both npm2 and npm3

jesperronn:
Modified engines field so that it clearly states that RxJS also works with npm@3. 
This should get rid of the warning when installing a project dependent on RxJS
```
npm WARN engine @reactivex/rxjs@5.0.0-alpha.7: wanted: {"npm":"~2.0.0"} (current: {"node":"4.1.2","npm":"3.3.3"})
```

This commit fixes issue #572

Please note that I could not really test this locally: I read https://docs.npmjs.com/files/package.json#engines and modified the package.json

Does it look good?

Issue #550, opened 2015-10-16
Pin build dependencies to known-working versions


blesh:
There is a comment [here](https://github.com/ReactiveX/RxJS/pull/547#issuecomment-148829837) that describes an issue I had where eslint upgraded to 1.7.0 when I wiped my node_modules directory and reinstalled. 1.7.0 had some false positives. I don't want that to happen to others while trying to create PRs, so I think it's best that we pin our build dependencies to highest known working versions.

PR #629, opened 2015-10-30
fix(package.json): loosen the engines/npm semver range to prevent fal…

IgorMinar:
…se warnings

This should get rid the following warning when installing rxjs:

```
npm WARN engine @reactivex/rxjs@5.0.0-alpha.7: wanted: {"npm":"~2.0.0"} (current: {"node":"4.2.1","npm":"2.14.7"})
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "watch": "0.16.0"
   },
   "engines": {
-    "npm": "~2.0.0"
+    "npm": ">=2.0.0"
   },
   "typings": "./dist/cjs/Rx.d.ts"
 }


kwonoj:
Seems this is similar to existing PR https://github.com/ReactiveX/RxJS/pull/615, is there specific differences?

IgorMinar:
@kwonoj mine has a better commit message :-)

also, I wouldn't constrain the v to `<4` because you will have the same issue pop up in the future. Nobody expects npm to be backwards incompatible with regards to served packages in the future. If they do make a breaking change, they'll need to deal with it in npm because it's not reasonable to expect that all existing packages will add `<4` version constraint.

kwonoj:
For commit messages, you're right :)

For version management perspective, though it is expected to work without any regressions, but sometimes it happens. There was related discussions previously in https://github.com/ReactiveX/RxJS/issues/550 which was about each individual package instead of npm itself, I think same could be applied for npm itself too considering unmatched upversion of npm (for example npm 3 in current state) wouldn't break functionality but only gives warnings. I know it is bit hassle to manually manage version time to time though.

blesh:
@kwonoj... I think I'm less worried about npm versions than I am about different JS dependency package versions. I'm fine with merging this to squelch warnings. If it becomes a problem for us for some reason, we can always change it back.

Merged with df791c6 ... thanks @IgorMinar

PR #630, opened 2015-10-30
windowWhen: fixes and tests

### LINKED ISSUE #628: windowWhen tests and fixes
Just to inform so that no one else does the same task. I'm writing tests for windowWhen (we were missing this one). Also found some unsubscription-related bugs, which I'll fix too.

staltz:
Resolves #628.


kwonoj:
First effort to merging PR, merged with 6fe3786 . Thanks @staltz 

CC @blesh to point out if there's anything done incorrectly.

PR #638, opened 2015-11-01
refactor(CoreOperators): correct class type constraint passed to oper…

### LINKED ISSUE #635: Generic Typedef Issues with Operators
Will investigate this further tomorrow, but:

Imagine I have an Observable of Foos:
```typescript
interface foo {
  bar: string;
}
let o: Observable<foo> = new Observable(obs => {});

```
If I .subscribe, I get (as expected) the generic passed through appropriately:
```typescript
o.subscribe( v => {
  return v.bar //ok - v is appropriately cast to foo
});
```

however, if I use .map() 
```typescript
o.map( v => {
  return v.bar; //error! v is cast to any here.
});
```
I would expect the parameter passed into an operator (in this case, the project function) to maintain the cast type, no? Obviously the return from a map function is different, but looks like there's a bit of a snafu on typedefs here.

kwonoj:
…ator

relates to #608, closes #635 

Refactors type definition of operators to properly pass type constraint of `Observable` itself. 

As stated in issue #635, with current type definition operators explicitly redefines type of values for observable and does not passes type constraint of class.

Simple illustrations in below shows 

```
class A<T> {
  doSomething<T>(value: T): void {
   ....
  }
}

new A<string>().doSomething(...
```

function `doSomething` now expected type constraint explicitly instead of following type constraint of class itself, type inference is confused. 

image

Basically, above code snippet is behaves same as below.

```
class A<T> {
  doSomething<R>(value: R): void {
   ....
  }
}
```

After correcting function signature

```
class A<T> {
  doSomething(value: T) {
  ...
  }
}
```

compiler expects type of argument as below.

image

This PR aligns signature of functions for operators having explicit generic parameter overwriting class type constraint.


blesh:
Merged with a05426f after some conflict resolution and a lint fix.

Thanks, @kwonoj!

PR #639, opened 2015-11-02
feat(mergeScan): add new mergeScan operator.

trxcllnt:


blesh:
Cool! This looks interesting.

I think we discussed this before, and the basic idea is that it's exactly like a `mergeMap/flatMap` accept you pump the value back into the selector, right? 

How does this differ from `expand`? What are the use cases?

cc/ @mattpodwysocki - might be interested in this.

staltz:
Pretty cool addition, we talked about it in person already.

It's missing some thorough unsubscription tests. But we can merge this already and later I can add unsub tests.

blesh:
oh oh ... @trxcllnt this is that "exscand" operator we discussed a while back right? Yes, I like this, and there's really not another way to do it efficiently with existing operators.

trxcllnt:
@blesh not yet, I was going to put that in my next PR and call it "expandScan." This operator flattens multiple inner Observables and computes a aggregate value over all the inner values.

blesh:
@trxcllnt can you rebase and resolve the conflicts?

blesh:
@trxcllnt why did this close?

trxcllnt:
@blesh because it'd already been merged.

blesh:
Ah... I failed to mention it in a comment. :\ Thanks.

PR #640, opened 2015-11-02
fix(util): incorrect Symbol.iterator for es6-shim

mlaval:
In es6-shim, the string for Symbol.iterator has been changed from `_es6shim_iterator_` to `_es6-shim iterator_` in https://github.com/paulmillr/es6-shim/commit/f19261a5d84652f689834500f7a69aa93f6d174c

This PR retrieves it dynamically instead of hard-coding it, but it means that es6-shim must be loaded before rx.js

### DIFF
--- a/src/util/Symbol_iterator.ts
+++ b/src/util/Symbol_iterator.ts
@@ -10,8 +10,18 @@ if (!root.Symbol.iterator) {
   } else if (root.Set && typeof new root.Set()['@@iterator'] === 'function') {
     // Bug for mozilla version
     root.Symbol.iterator = '@@iterator';
+  } else if (root.Map) {
+    // es6-shim specific logic
+    let keys = Object.getOwnPropertyNames(root.Map.prototype);
+    for (let i = 0; i < keys.length; ++i) {
+      let key = keys[i];
+      if (key !== 'entries' && key !== 'size' && root.Map.prototype[key] === root.Map.prototype['entries']) {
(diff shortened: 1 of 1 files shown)


src/util/Symbol_iterator.ts
```diff
@@ -11,7 +11,14 @@ if (!root.Symbol.iterator) {
     // Bug for mozilla version
     root.Symbol.iterator = '@@iterator';
   } else {
-    root.Symbol.iterator = '_es6shim_iterator_';
+    // es6-shim specific logic
+    var keys = Object.getOwnPropertyNames(Map.prototype);
+    for (var i = 0; i < keys.length; ++i) {
+      var key = keys[i];
+      if (key !== 'entries' && key !== 'size' && Map.prototype[key] === Map.prototype['entries']) {
+        root.Symbol.iterator = key;
```
gkalpak: Couldn't you also break here ?

mlaval: You're right, thanks

mlaval:
This is exactly what I meant by "it means that es6-shim must be loaded before rx.js".
Retrieving this token is anyway specific to es6-shim.

If if is to much of a constraint, the correct value could just be hard-coded instead.

mlaval:
Comments integrated

blesh:
Looks legit to me.

blesh:
Merged with 15bf32c... thanks @mlaval!!!

PR #641, opened 2015-11-02
test(subscribeOn): expand test coverage for subscribeOn

kwonoj:
relates to #607 

Continue effort to expand test coverage.


blesh:
Merged with 6c0086e ... thanks @kwonoj!

Issue #642, opened 2015-11-03
Type definition for extended operator is not published


kwonoj:
Code snippet below works in JS, 

```
var Rx = require('....Rx.KitchenSink');
Rx.Observable.of(42).isEmpty().subscribe(console.log);
```

when it comes to TS 

```
import * as Rx from '....Rx.KitchenSink';
Rx.Observable.of(42).isEmpty().subscribe(console.log);
```

compiler will complain due to `Observable.of` does not returns extended type of operator instead of core.
`error TS2339: Property 'isEmpty' does not exist on type 'Observable<number>'.`

kwonoj:
This issue is related with modularity strategies. Tried some approaches ends up with vain, exploring few other approaches including mixins.

blesh:
cc/ @robwormald

blesh:
cc/ @jeffbcross

kwonoj:
Created PR https://github.com/ReactiveX/RxJS/pull/643 for initial proposal based on mixin, and possibly including ambient declaration merging if it's necessarily required.

robwormald:
Nice work OJ! I'll try this out later today but at first glance it looks
like it'll solve a bunch of issues for us.

kwonoj:
New changes in typescript compiler supporting polymorphic `this` (https://github.com/Microsoft/TypeScript/pull/4910) may be helpful to deal with this issues, but it's just rough guess and haven't explored to see if it's feasible with current codebases yet.

kwonoj:
Related with https://github.com/ReactiveX/RxJS/issues/750 .

blesh:
@kwonoj @robwormald is this issue resolved? Can we close this?

kwonoj:
No, it's still there and we don't have decent solution yet.

blesh:
@kwonoj @jeffbcross @robwormald ... is it reasonable to say this is a problem with TypeScript and not necessarily with how RxJS is structured?

kwonoj:
I'm considering half and half. TypeScript can possibly bring better feature resolves this issue, but I would expect it'll eventually requires change structures of RxJS codebase, just matter of how much. Short example would be interfaces are tied to return `Observable` only as well as implementation of `lift` does, etcs.

As an issue tracking perspective, I don't think this can be updated in relatively short time frame. I'm seeing couple of discussions are in TypeScript perspective would require some time to something's actually happening. I can suggest close this for now and bring it back once circumstances allowing further discussion.

[...]

PR #643, opened 2015-11-03
Module composition streategy proposal

kwonoj:
**This PR intended to show proof-of-concept of proposal, neither mature or intended to be actually merged**.

While seeing issues like https://github.com/ReactiveX/RxJS/issues/635, https://github.com/ReactiveX/RxJS/issues/642, came to feel if there's way to create module met requirements like

- type definition to be preserved over core, extended along with any custom observable module created in further
- reduce type signature duplication over difference places

This proposal is based on [Mixin](https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Mixins.md) of typescript, compose class implementations to extend modules if necessary.

**How composition works**

Attached PR illustrates rough mechanisms of composition. First, there is `Core` implementation of observable having bare minimum set in `CoreObservable.ts`. In this PR it implements interface of CoreOperator, this could be ambient merging introduced in typescript nightly. (There's one caveat of ambient merging, mixin class should declare all interfaces even if it's not implemented in class)
`Rx.Core.ts` does exports it directly without any extension.

`ExtendedObservable.ts` represents fully features set of Observable module, corresponds to current 'KitchenSink'. it does implements additional operator. `Rx.Extended.ts` is place where actual composition occurs, it declares mixin `Observable<T>` which implements Core & Extended both, then call `applyMixin` to compose mixin class at runtime. This runtime composition occurs very first time module `ExtendedObservable` is imported.

`Rx.My.ts` is representing cases where any consumer would like to create custom module for own flavor. In this PR, it utilizes core observable along with adding new operator implementation.

`testIndex.ts` shows usage of each module, simply import it and type definition starts to work.

**Declaring Mixin**

When declaraing mixin, it does do some specifics
1. should include all factory function signatures like `of`, `lift`, etcs
: Actual implementation of factory functions are injected while applying mixin (line 18 in `CoreObservable`) it is important to declare signature in each mixin explicitly. This prevents issues like https://github.com/ReactiveX/RxJS/issues/642, extended observale generator to return extended type to allow correct type inference.
2. Mixin operator's stub declaration is property, not function. And it does have 'different' return type than original.
: For example. `CoreObservable::filter():Operators` should be declared in mixin as `MyObservable::filter():MyObservable`. This is for type inference as same as 1.

**Possible pros**
- explicit prototype injection in module declaration can go away (though mixin internally does this)
- type definition works for all cases (core, extended, even custom)

**Cons**
- Mixin declaration still requires some of stubs though it's reduced compare to current approach.
- Requires specific formats to declare mixin to make it work, such as property stub declaration with corresponding types.

### DIFF
--- a/src/CoreObservable.ts
+++ b/src/CoreObservable.ts
@@ -1,0 +1,31 @@
+//Represents current 'Observable'
+
+export interface Operators<T> {
+  map?(): Operators<any>;
+  filter?(): Operators<any>;
+}
+
+export class CoreObservable<T> implements Operators<T> {
+  filter(): Operators<any> {
+    console.log('filter');
+    return this;
+  }
+
+  static of: () => CoreObservable<any>;
+}
+
+export function applyMixins(derivedCtor: any, baseCtors: any[]) {
+  derivedCtor.of = function () {
+    return new derivedCtor();
+  };
+
+  console.log('mixin called');
+
+  baseCtors.forEach(baseCtor => {
+      Object.getOwnPropertyNames(baseCtor.prototype).forEach(name => {
+          derivedCtor.prototype[name] = baseCtor.prototype[name];
+      })
+  });
+}
+
+export default CoreObservable;
--- a/src/ExtendedObservable.ts
+++ b/src/ExtendedObservable.ts
@@ -1,0 +1,8 @@
+// Represents current 'KitchenSink'
+
+export default class ExtendedObservable<T> {
+  isEmpty(): ExtendedObservable<any> {
+    console.log('isEmpty');
+    return this;
+  }
+}
--- a/src/Rx.Core.ts
+++ b/src/Rx.Core.ts
@@ -1,0 +1,6 @@
+import CoreObservable from './CoreObservable';
+
+//core rx does not have anything to add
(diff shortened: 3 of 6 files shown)


blesh:
I think I'll personally need to let this idea simmer for a while. Since TypeScript and Babel are now both going to support the `::` operator, and we have a two globals that handle the rest, I'm inclined to think that TypeScript adding `::` solved our problem for us for most people... Hmmm...

kwonoj:
@blesh agreed. This is one of proposals, and also could possibly write up another using new operator. BTW, `::` in typescript is not landed yet, right? When I checked issues for that some days ago it seems it was ongoing.

blesh:
> :: in typescript is not landed yet, right?

That's correct. It's sitting in a development branch last I looked.

kwonoj:
I came to think this PR can be closed for now, PR can be reopened anytime when time comes to evaluate this vs. any other possibilities.

kwonoj:
After discussion with @blesh , I'm closing this issue for now since alternative candidate #750 seems quite feasible solution. PR can be recreated anytime if this approach need to be revisited.

PR #652, opened 2015-11-05
Publish refactor

blesh:
This is to address #453 

- make `multicast` polymorphic to support new factory behavior and old subject argument behavior
- restore the following operators to RxJS 4 behavior: `publish`, `publishBehavior`, `publishReplay`
- ensure `share` returns an Observable that can be retried
- remove `shareBehavior` and `shareReplay` to prevent end-user-developer confusion.


staltz:
Looks all OK.

blesh:
Great. Putting this to bed.

PR #653, opened 2015-11-05
chore(travis): use node.js 4 on travis

### LINKED ISSUE #650: Figure out why the TypeScript build is so sloooooooowwww...
I'm not sure when or why it happened, but it seems like the ES6 build in particular got much, much slower over the last few weeks to the point where it's getting a little silly.

@jeffbcross @IgorMinar @robwormald ... have you run into this with Angular?

kwonoj:
closes #650

PR enables use of node.js 4 on travis, at least resolves build time issue on CI. In case of local dev machines, it'll still depends on each individual configurations.


package.json
```diff
@@ -13,9 +13,9 @@
   },
   "scripts": {
     "build_all": "npm run build_es6 && npm run build_amd && npm run build_cjs && npm run build_global",
-    "build_amd": "rm -rf dist/amd && tsc typings/es6-shim/es6-shim.d.ts src/Rx.ts -m amd --outDir dist/amd --sourcemap --target ES5",
+    "build_amd": "rm -rf dist/amd && tsc typings/es6-shim/es6-shim.d.ts src/Rx.ts -m amd --outDir dist/amd --sourcemap --target ES5 --diagnostics",
```
kwonoj: let compiler output diagnostics by default just in case.

.travis.yml
```diff
@@ -1,7 +1,15 @@
...
 node_js:
-  - '0.12'
+  - '4.2.2'
+
+addons:
+  apt:
+    sources:
+    - ubuntu-toolchain-r-test
+    packages:
+    - gcc-4.8
```
kwonoj: this compiler dependency is introduced for `selenium-webdriver` requires C++11 supported compilers with node.js 4.

kwonoj:
Seems selenium's still not satisfied with current configs, I'll update PR shortly.

package.json
```diff
@@ -100,7 +100,7 @@
     "multi-stage-sourcemap": "0.2.1",
     "platform": "1.3.0",
     "promise": "7.0.3",
-    "protractor": "2.2.0",
+    "protractor": "2.5.1",
```
kwonoj: `protractor` bumped to resolve `selenium-webdriver` dependency (https://github.com/angular/protractor/blob/master/CHANGELOG.md#240) to let it be built without issues on node v4.

kwonoj:
Now travis is happy, and tsc is happy too.

> rm -rf dist/es6 && tsc src/Rx.ts src/Rx.KitchenSink.ts --outDir dist/es6 --sourceMap --target ES6 -d --diagnostics
Files:             163
Lines:           27546
Nodes:          122767
Identifiers:     41558
Symbols:        216842
Types:          117400
Memory used:   244259K
I/O read:        0.02s
I/O write:       0.07s
Parse time:      0.56s
Bind time:       0.36s
Check time:      4.05s
Emit time:       0.73s
Total time:      5.70s

weswigham:
You may want to [consider adding 'stable' to your `node_js` travis config alongside the '4' series](https://github.com/Microsoft/TypeScript/pull/5536). Node 5 is the most recent stable version, and it makes sense to track the most recent stable build in CI, especially given `node`'s expected [release cadence](https://nodejs.org/en/blog/community/node-v5/).

kwonoj:
@weswigham Thanks, I was actually thought about that but choose LTS v4 instead. @blesh would it be acceptable to follow stable release on CI?

Widdershin:
Shouldn't we run both LTS and the currently supported releases on Travis?

blesh:
@kwonoj ... I've already merged this particular PR. I'm a little leary of adding `stable` at the moment. I'd sort of like to see how the Node community does with this faster cadence with regards to breakage and compatibility.

I'm not opposed to adding `stable` though, let's just wait for a couple of node releases and see how it goes.

blesh:
Merged with b5ecacc... and thank for the work @kwonoj, and also thanks for the input @weswigham and @Widdershin ... I have a feeling that we'll probably move to running the build on latest stable node, but I just want to proceed cautiously with the build side of things, since things are working alright for now.

kwonoj:
Agreed, that was my original intention to include specific version of node. We can observe release as you mentioned to decide include it or not.

PR #664, opened 2015-11-06
Refactor source to use named exports everywhere

### LINKED ISSUE #663: Move away from `export default` in all files?
WHY:

- So we can build everything TSC to fix issues like #647 
- Both TSC and Babel are no longer going to support setting `module.exports` directly for files with only an `export default`.
- CJS users on Node will most likely import Rx via `var Rx = require('@reactivex/rxjs');`, and not worry about modularity.
- Those worried about modularity probably won't care about importing like `var filter = require('@reactivex/rxjs/dist/cjs/operators/filter').filter`, because it's long enough.
- We can probably come up with a better distribution scheme, like publishing each operator on npm individually so users can do: `var filter = require('@reactivex/rx-filter')` or the like.

Thoughts? Concerns?

robwormald:
So it's kind of dark and cold in Sweden, and dirty little secret I've been maintaining this branch for a while anyway. 

Refactors all the source to use named exports everywhere. Also moves compilation to use typescript instead of babel and removes the copy task for moving d.ts files around.

Merge if you like. Or not. You can't stop me from maintaining my own copy MUHAHAHA.

Closes #663


robwormald:
Note - other than structural changes, the one thing to explicitly point out is a change to HotObservable:
https://github.com/robwormald/RxJS/blob/chore/remove-defaults/src/testing/HotObservable.ts#L35-L41

Note the anon IIFE wrapped here - this is due to the fact typescript currently complains with block scoped vars in loops. See https://eugenebichel.wordpress.com/2015/10/22/typescript-let-in-closures-and-how-let-works-after-compile-to-javascript/

robwormald:
TSLint passing now. That parameters not aligned check is a doozy - one thing we use over in angular2 is clang-format, which handles all that business as far as formatting, and produces consistent (albeit occasionally weird) formatting checks and automagic reformatting.

kwonoj:
>  this is due to the fact typescript currently complains with block scoped vars in loops

I remember I also hit this when I tried to build CJS via TS for code coverage but hasn't track down deeply. I think it's ok for now, may need revisit to refactor code to lint / compiler does not complain anymore.

> clang-format, which handles all that business as far as formatting

Obviously interesting. Is [this](https://www.npmjs.com/package/clang-format) right module?

package.json
```diff
@@ -14,7 +14,7 @@
   "scripts": {
     "build_all": "npm run build_es6 && npm run build_amd && npm run build_cjs && npm run build_global",
     "build_amd": "rm -rf dist/amd && tsc typings/es6-shim/es6-shim.d.ts src/Rx.ts -m amd --outDir dist/amd --sourcemap --target ES5 --diagnostics",
-    "build_cjs": "rm -rf dist/cjs && babel dist/es6 --out-dir dist/cjs --modules common --source-maps --loose all && node lib/copy_dts.js",
+    "build_cjs": "rm -rf dist/cjs && tsc typings/es6-shim/es6-shim.d.ts src/Rx.ts src/Rx.KitchenSink.ts -m commonjs --outDir dist/cjs --sourcemap --target ES5 -d --diagnostics",
```
kwonoj: I assume test coverage for this PR is not reported due to travis configuration try to remap sourcemap from babel's. Once #663's concluded, can include those changes too.

robwormald: Hmm good question. Wasn't really sure on the coverage stuff to be honest.

kwonoj: It'll be trivial change, and I can make those quickly. Let's not worry about it for now, see how discussions goes first if this `little secret` can be merged :)

robwormald:
See https://github.com/angular/angular/blob/master/DEVELOPER.md - there's a section about clang-format there, and see the .clang-format config file in the root.

kwonoj:
Appreciate for help, I'll look it up and prepare PR for RxJS as well.

trxcllnt:
@robwormald 👍🏽🙌🏽👏🏽🤘🏽🖖🏽👌🏽🙏🏽😘💕❤️💋 for removing intermediate Babel step 😍👽👻😻🤗

blesh:
This LGTM.

kwonoj:
:heart: for this PR. I have created PR to followup this for code coverage side at https://github.com/ReactiveX/RxJS/pull/672

blesh:
@robwormald ... can you rebase this one so we can get it merged?

robwormald:
Yup will do shortly

robwormald:
bases have been re'ed

blesh:
Thanks @robwormald! This one was a big deal.

kwonoj:
Yay!

Issue #661, opened 2015-11-05
Add test for type definition validation?


kwonoj:
There are some cases of type definition does not work as expected like reported issue https://github.com/ReactiveX/RxJS/issues/660, https://github.com/ReactiveX/RxJS/issues/642, https://github.com/ReactiveX/RxJS/issues/635, Came to feel it'd be helpful if there's compile-time type signature validation test is included CI. 

It'll not be kind of unit test, simply write down basic signatures of operator invocation in typescript instead of plain javascript, let CI try to compile it only (does not execute). Compilation result will tell if current type signature works in expected way or not. Course it can't cover whole aspect of signature, but at least can have minimum lines of basic usages.

blesh:
It sounds cool... @robwormald @jeffbcross what do you think about this? Do you guys do anything similar for Angular?

robwormald:
Yeah, I think this is a great idea. It's a little tricky to test, but the original issue that I posted cropped up by doing something similar:
```typescript
class Response {
  constructor(private _text:string){}
  toJson():Object { return JSON.stringify(this._text); }
}

let o = new Observable(obs => {
  obs.next(new Response('{"foo":"bar"}'));
  obs.complete();
});

obs.map(res => res.toJson()).subscribe() //complains - {} does not have a .toJson() method!

```
should be just a matter of setting up TS to compile some basic usage code with some typed classes. Would be happy to contribute here if you can get the mechanics set up @kwonoj

robwormald:
In angular2, our tests are written in TS and so such issues get caught when the tests are being compiled.

kwonoj:
As @robwormald mentioned, main difference between ng2 is RxJS test cases are written in JS. 

I'll try to spend some times to think about ergonomics setup for this and create PR for proposal.

kwonoj:
This could be achieved moving current test cases into typescript based one as same as Angular2 does, if circumstances allows.

kwonoj:
I'm assigning this myself for evaluating if current test cases can be moved into typescript based one.

kwonoj:
Currently there are 3 (including this) way proposed,

- convert unit test to typescript based, add additional test cases in existing tests (#826)
- separate test running for type validation only with custom runner using compiler api (#667)
- separate test, running with compiler out of box (#1189) by @david-driscoll

blesh:
@kwonoj, I'll defer to you on this one.

[...]

PR #667, opened 2015-11-06
Proposal for type validation test runner

kwonoj:
relates to #661 

This PR include small code snippet to check basic type definitions as discussed in issue.
There'll be chances of improvement to have additional type detection for verbose output, would like to consider this as starting point.

No actual tests & CI integration included yet.

**Output example**
image

### DIFF
--- a/types/index.js
+++ b/types/index.js
@@ -1,0 +1,37 @@
+var _ = require('lodash');
+var colors = require('colors');
+var glob = require('glob');
+var fs = require('fs');
+var ts = require('typescript');
+
+function diagnose(fileNames, options) {
+  var program = ts.createProgram(fileNames, options);
+
+  var diagnostics = _.groupBy(ts.getPreEmitDiagnostics(program),
+    function (x) {
+      return x.file.fileName;
+    });
+
+  _.forIn(diagnostics, function (value, key) {
+    fs.readFile(key, 'utf-8' ,function (err, file) {
+      var lines = file.split('\n');
+      _.forEach(value, function (diagnostic) {
+        var pos = diagnostic.file.getLineAndCharacterOfPosition(diagnostic.start);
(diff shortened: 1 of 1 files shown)


kwonoj:
cc @robwormald , would like to ask suggestions for output format or any other.

types/index.js
```diff
@@ -0,0 +1,38 @@
...
+
+      var lines = file.split('\n');
+      _.forEach(value, function (diagnostic) {
+        var pos = diagnostic.file.getLineAndCharacterOfPosition(diagnostic.start);
+        var message = ts.flattenDiagnosticMessageText(diagnostic.messageText, '\n');
+
+        console.log('Error : '.red + message);
+        console.log(lines[pos.line].gray);
+        console.log(_.repeat(' ', pos.character) + '^');
+        console.log();
```
staltz: How about `console.error()`?

kwonoj: Yes, that could be way. Will update.

kwonoj:
Small update suggestion from @staltz to use console.error with minor output changes.

image

blesh:
This is interesting. How long does it take to run, on average? (I'll pull it down sometime soon and give it a try, too)

kwonoj:
What I did was just small code snippet does not reflect real world cases (having lot of files, or codes) though, it took about a second. For comparison reference, on same system typescript build takes around 14 sec, unit test takes around 2.3 second.

blesh:
@trxcllnt ... what are your thoughts?

kwonoj:
Marked this as blocked while exploring feasibility of migrating unit test cases to typescript based.

blesh:
@kwonoj I'm going to close this for now since discussion here is dead. We can open a new issue for it to create more discussion, or we can discuss off-line and open an issue when we have some plans outlined.

kwonoj:
Agreed, thanks.

PR #668, opened 2015-11-07
test(FromObservable): expand test coverage for FromObservable

kwonoj:
Add couple of tests to cover `fromObservable`.


blesh:
Merged with 4d60edd. Thanks @kwonoj

PR #669, opened 2015-11-07
switch, mergeAll and concatAll signatures 

jinroh:
Some operators dealing with observable sequences (`switch`, `mergeAll` and `concatAll`) can have more precise and consistent signatures. (thanks @kwonoj)

### DIFF
--- a/src/CoreOperators.ts
+++ b/src/CoreOperators.ts
@@ -40,7 +40,7 @@ export interface CoreOperators<T> {
   mapTo?: <R>(value: R) => Observable<R>;
   materialize?: () => Observable<Notification<T>>;
   merge?: (...observables: any[]) => Observable<any>;
-  mergeAll?: (concurrent?: any) => Observable<any>;
+  mergeAll?: (concurrent?: number) => Observable<T>;
   mergeMap?: <R>(project: ((x: T, ix: number) => Observable<any>), projectResult?: (x: T, y: any, ix: number, iy: number) => R, concurrent?: number) => Observable<R>;
   mergeMapTo?: <R>(observable: Observable<any>, projectResult?: (x: T, y: any, ix: number, iy: number) => R, concurrent?: number) => Observable<R>;
   multicast?: (subjectFactory: () => Subject<T>) => ConnectableObservable<T>;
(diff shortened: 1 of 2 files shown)


src/CoreOperators.ts
```diff
@@ -15,7 +15,7 @@ export interface CoreOperators<T> {
   combineAll?: <R>(project?: (...values: Array<any>) => R) => Observable<R>;
   combineLatest?: <R>(...observables: Array<Observable<any> | ((...values: Array<any>) => R)>) => Observable<R>;
   concat?: <R>(...observables: (Observable<any> | Scheduler)[]) => Observable<R>;
-  concatAll?: () => Observable<any>;
+  concatAll?: () => Observable<T>;
```
kwonoj: I was actually thinking about `concat` as well, return type may need to be `Observable<any>` instead of specific types. I think there would be possible cases of `concatall` executed over multiple observable has different types - what do you think?

src/CoreOperators.ts
```diff
@@ -40,7 +40,7 @@ export interface CoreOperators<T> {
   mapTo?: <R>(value: R) => Observable<R>;
   materialize?: () => Observable<Notification<T>>;
   merge?: (...observables: any[]) => Observable<any>;
-  mergeAll?: (concurrent?: any) => Observable<any>;
+  mergeAll?: (concurrent?: number) => Observable<T>;
```
kwonoj: For return type, question maybe same as `concatall`.

src/CoreOperators.ts
kwonoj: Stub code might be like this,
```
Rx.Observable.range(0,3).map(function (x) {
  if (x === 1) {
    return Rx.Observable.range(0,3).mapTo('a');
  }
  return Rx.Observable.range(x, 2);
}).concatAll().subscribe(console.log);
```

output becomes `Observable<number | string>` in this case.
>0
1
a
a
a
2
3

jinroh: I'm not a heavy or experimented TS user so I may be wrong but don't you lose the interest of TS if you put `<any>` here ?

My reasoning with these typings was more: let's inform the "compiler" that `switch` / `mergeAll` / `concatAll` have the interesting type behaviour to keep the type of the original observable sequence they derive from. This way the compiler can give better type informations. Wouldn't you lose all that with `<any>` ?

kwonoj: Oops, yes, you're correct - apologize. I've read type constraint incorrectly and thought `Observable.range<number>.map().concatall(): Observable<number | string>`, which is not since `map()` does passes type constraint correctly. 

Thanks for pointing out.

src/CoreOperators.ts
kwonoj: Above question is also should be voided.

kwonoj:
Merged with e9b13d0, appreciate for taking time for this @jinroh .

kwonoj:
Actually c37cecf is commit sha, added one additional changes for lint friendly.

PR #670, opened 2015-11-07
chore(mergeall-perf): fix mergeall micro perf benchmarks

jinroh:
`mergeAll` operator micro perf benchmarks were incorrect, and in favor of RxJS5.
Using `mapTo` is error-prone in these kind of benchmarks. I just replaced them by simple `map`s.

### DIFF
--- a/perf/micro/current-thread-scheduler/operators/mergeall.js
+++ b/perf/micro/current-thread-scheduler/operators/mergeall.js
@@ -1,12 +1,18 @@
 var RxOld = require('rx');
 var RxNew = require('../../../../index');
 
+var source = Array.apply(null, { length: 25 });
+
 module.exports = function (suite) {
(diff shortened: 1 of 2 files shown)


trxcllnt:
@jinrih how did the previous test setup favor RxJS5? My concern here is now we can't compare to previous runs, because calling the map selector each time to create the inner Observable is significantly slower than `mapTo`, so now we're testing more than just mergeAll performance.

Perhaps a better test (if we don't want to use map/mapTo at all) would be:

```es6
Observable.of(
  Observable.range(0, 25), 
  Observable.range(0, 25), 
  Observable.range(0, 25),
  ...
).mergeAll();
```

jinroh:
@trxcllnt I think I see what you mean. I updated the benchmarks to measure only `mergeAll` instead of `map(To) + mergeAll` operators.

blesh:
Thanks @jinroh ... merged with f9e92df

PR #671, opened 2015-11-07
feat(forkJoin): accept promise, resultselector as parameter of forkJoin

### LINKED ISSUE #507: accepting promise for forkJoin()?
RxJS4's [`forkJoin`](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/forkjoin.md) accepts promise as well as observables

`args (Arguments | Array): An array or arguments of Observable sequences or Promises to collect the last elements for.`

while current implementation does not. Is this intended design to drop promise? If promise can be accepted, PR https://github.com/ReactiveX/RxJS/pull/506 might need to be updated to include it as well as updating implementation.

kwonoj:
closes #507 

`forkJoin` now accepts promises as well as result selector, and expanded test coverage.

### DIFF
--- a/src/observables/ForkJoinObservable.ts
+++ b/src/observables/ForkJoinObservable.ts
@@ -1,50 +1,96 @@
 import {Observable} from '../Observable';
 import {Subscriber} from '../Subscriber';
+import {PromiseObservable} from './PromiseObservable';
+import {EmptyObservable} from './EmptyObservable';
+import {isPromise} from '../util/isPromise';
 
 export class ForkJoinObservable<T> extends Observable<T> {
-  constructor(private observables: Observable<any>[]) {
-    super();
+  constructor(private sources: Array<Observable<any> |
+                                    Promise<any> |
+                                    ((...values: Array<any>) => any)>) {
+     super();
+   }
+
+  static create(...sources: Array<Observable<any> |
+                                  Promise<any> |
+                                  ((...values: Array<any>) => any)>)
+                                  : Observable<any> {
+    if (sources === null || sources.length === 0) {
+      return new EmptyObservable();
+    }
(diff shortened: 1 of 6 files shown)


kwonoj:
Updated PR. Appreciate take time to review with suggestions, @blesh

blesh:
Looks really good. I found the most minor thing in the world (a `let` instead of a `const`). You don't have to fix it if you don't want to, we can just merge the changes. Just thought I'd point it out.

kwonoj:
> You don't have to fix it if you don't want to

I've update it, though I usually try to be strict on code in this case I obviously missed it. Thanks for pointing it out.

blesh:
> I've update it, though I usually try to be strict on code in this case I obviously missed it. Thanks for pointing it out.

No problem, what I try to do is `const` everything and then let the compiler or IDE complain at me when I need a `let`

blesh:
Merged with 190f349 ... thanks! Good work!

src/observables/ForkJoinObservable.ts
```diff
@@ -1,50 +1,96 @@
...
+import {isPromise} from '../util/isPromise';
 
 export class ForkJoinObservable<T> extends Observable<T> {
-  constructor(private observables: Observable<any>[]) {
-    super();
+  constructor(private sources: Array<Observable<any> |
+                                    Promise<any> |
+                                    ((...values: Array<any>) => any)>) {
+     super();
+   }
```
staltz: I hope you don't mind nitpicking, but I noticed 3 spaces for indentation here. Should be 2.

kwonoj: I don't mind it at all, actually I also do similar thing when I review with my colleagues :) I'll check codes and apply changes if necessary.

PR #673, opened 2015-11-08
refactor(Immediate): fix minor syntax in test cases

kwonoj:
Minor correction to variable assignment in test case.


kwonoj:
Merged with d39e768.

PR #681, opened 2015-11-09
refactor(CoreOperators): update type definition for operator has defa…

### LINKED ISSUE #660: Property stub based type definition does not support parameter default value
Some of operators allows default value if it's not specified like below,

```
function repeat<T>(count: number = -1);
function defaultIfEmpty<T, R>(defaultValue: R = null);
```

and `Observable` has property stub to create prototype in module like 
`repeat: <T>(count: number) => Observable<T>`
default value for parameter is not allowed, end result causes typescript compiler complains without parameter.

```
Rx.Observable.of(42).defaultIfEmpty();
error TS2346: Supplied parameters do not match any signature of call target.
```

related with https://github.com/ReactiveX/RxJS/issues/608, and https://github.com/ReactiveX/RxJS/pull/643 for remove property based stubs.

kwonoj:
…ult parameter value

closes #660 

Appreciate @blesh for reminding optional parameter, seems I was dizzy when creating issue does not able to think about it.

### DIFF
--- a/src/CoreOperators.ts
+++ b/src/CoreOperators.ts
@@ -22,7 +22,7 @@ export interface CoreOperators<T> {
   dematerialize?: () => Observable<any>;
   debounce?: (durationSelector: (value: T) => Observable<any> | Promise<any>) => Observable<T>;
   debounceTime?: <R>(dueTime: number, scheduler?: Scheduler) => Observable<R>;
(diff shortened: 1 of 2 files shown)


kwonoj:
I'll check in this myself since changes are obvious and small.

kwonoj:
Merged with 3d5d050.

blesh:
:+1:

PR #683, opened 2015-11-10
chore(test): update wallaby configuration

kwonoj:
update wallaby configuration remove babel dependency


kwonoj:
Merged with 982f902 .

PR #684, opened 2015-11-10
test(Notification): enhance test case for error notification

kwonoj:
very minor enhancement to test `Notification.createError`.


kwonoj:
Merged with 5bfa87a .

PR #685, opened 2015-11-10
test(observeOn): add test cases for observeOn operator

kwonoj:
Expanding test coverage for remaining codes.


blesh:
Looks great, @kwonoj! Merged this as of b2e46aa

PR #687, opened 2015-11-10
test(ReplaySubject): add more tests for ReplaySubject

staltz:
Add tests that verify ReplaySubject's behavior with multiple subscribers, for both fixed bufferSize
and fixed windowTime parameters.


blesh:
Merged with 60859cc. Thanks @staltz

PR #688, opened 2015-11-10
test(BehaviorSubject): add more tests for BehaviorSubject

staltz:
Add tests that verify BehaviorSubject's behavior with multiple subscribers, also when it is
subscribed after it has completed.


blesh:
Merged with 6d81d32... thanks @staltz. Great work.

PR #689, opened 2015-11-10
refactor(ReplaySubject): rename an argument to ReplaySubject's ctor

staltz:
Rename constructor argument `_windowTime` to `windowTime`, for better developer experience when using IDEs that suggest usage by showing a dropdown autocomplete menu. This might also positively impact documentation, which will should show `windowTime` instead of `_windowTime`. The underscore was only needed in order to avoid collision with the operator method `windowTime()` in Observable class.

### DIFF
--- a/src/subjects/ReplaySubject.ts
+++ b/src/subjects/ReplaySubject.ts
@@ -11,11 +11,11 @@ export default class ReplaySubject<T> extends Subject<T> {
   private events: ReplayEvent<T>[] = [];
 
   constructor(bufferSize: number = Number.POSITIVE_INFINITY,
-              _windowTime: number = Number.POSITIVE_INFINITY,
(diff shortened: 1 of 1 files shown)


kwonoj:
:+1:

kwonoj:
what about renaming field `_windowTime` also? underbar for field is usually for backing field for property, but this doesn't have property.

staltz:
> The underscore was only needed in order to avoid collision with the operator method windowTime() in Observable class.

Because there is an operator called windowTime, it's a member of the Observable class and ReplaySubject extends Subject extends Observable.

kwonoj:
You're very correct. I didn't noticed by looking it via github. Thanks for clarification :)

kwonoj:
Merged with 9f0ecf7 . Thanks, @staltz

PR #691, opened 2015-11-10
fix(Immediate): nextTick scheduler should no longer throw in Chrome

### LINKED ISSUE #690: BUG: nextTick scheduler throwing when calling setImmediate in Chrome
Looks like an oversight on my part. Here is a [JSBin](http://jsbin.com/bezugabeji/1/edit?html,js,console) demoing the broken behavior.

blesh:
fixes #690

### DIFF
--- a/spec/util/Immediate-spec.js
+++ b/spec/util/Immediate-spec.js
@@ -1,4 +1,5 @@
 var ImmediateDefinition = require('../../dist/cjs/util/Immediate').ImmediateDefinition;
+var Rx = require('../../dist/cjs/Rx');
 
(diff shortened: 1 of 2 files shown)


kwonoj:
Would you mind share some details why chrome has issue with it? Seems other browser doesn't (tried with Edge / Chrome). Change itself looks ok in general.

blesh:
@kwonoj if you look at the issue, it's in `setImmediate`... the polyfill in Chrome is hitting code with a defect in it.

https://github.com/ReactiveX/RxJS/issues/690

kwonoj:
Thanks for details.

PR #698, opened 2015-11-11
chore(switch): add micro perf benchmarks for switch operator

jinroh:


blesh:
Merged with b9b58cc... thanks @jinroh!

PR #699, opened 2015-11-11
chore(concatall): fix concatall perf benchmarks

jinroh:
The `concatAll` benchmarks had the same issues as `mergeAll` (see #670).


kwonoj:
I'm feeling some of other tests are leaning on `map` need to follow this approach, is my understanding correct?

kwonoj:
Merged with d6c1e7a . @jinroh , thanks!

blesh:
> I'm feeling some of other tests are leaning on map need to follow this approach, is my understanding correct?

Yes, I'd say that's right.

kwonoj:
I'll open issue to not to forget and track down for it.

Issue #696, opened 2015-11-10
implement skipWhile


blesh:
we don't have `skipWhile` implemented yet, and it's probably a core operator.

jinroh:
I'd be happy to take this one !

PR #703, opened 2015-11-11
feat(operator): add skipWhile operator

jinroh:
(#696)

Waiting for your reviews !

### DIFF
--- a/spec/operators/skipWhile-spec.js
+++ b/spec/operators/skipWhile-spec.js
@@ -1,0 +1,181 @@
+/* globals describe, it, expect, expectObservable, expectSubscriptions, hot */
+var Rx = require('../../dist/cjs/Rx');
+var Observable = Rx.Observable;
+
+describe('Observable.prototype.skipWhile()', function () {
+  it('should skip all elements with a true predicate', function () {
+    var source = hot('-1-^2--3--4--5--6--|');
+    var sourceSubs =    '^               !';
+    var expected =      '----------------|';
+
+    expectObservable(source.skipWhile(function () { return true; })).toBe(expected);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+  });
+
+  it('should skip all elements with a truthy predicate', function () {
+    var source = hot('-1-^2--3--4--5--6--|');
+    var sourceSubs =    '^               !';
+    var expected =      '----------------|';
+
+    expectObservable(source.skipWhile(function () { return {}; })).toBe(expected);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+  });
+
+  it('should not skip any element with a false predicate', function () {
+    var source = hot('-1-^2--3--4--5--6--|');
+    var sourceSubs =    '^               !';
+    var expected =      '-2--3--4--5--6--|';
+
+    expectObservable(source.skipWhile(function () { return false; })).toBe(expected);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+  });
+
+  it('should not skip any elements with a falsy predicate', function () {
+    var source = hot('-1-^2--3--4--5--6--|');
+    var sourceSubs =    '^               !';
+    var expected =      '-2--3--4--5--6--|';
+
+    expectObservable(source.skipWhile(function () { return undefined; })).toBe(expected);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+  });
+
+  it('should skip all elements until predicate is false', function () {
+    var source = hot('-1-^2--3--4--5--6--|');
+    var sourceSubs =    '^               !';
+    var expected =      '-------4--5--6--|';
+
+    var predicate = function (v) {
+      return +v < 4;
+    };
+
+    expectObservable(source.skipWhile(predicate)).toBe(expected);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+  });
+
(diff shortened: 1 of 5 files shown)


spec/operators/skipWhile-spec.js
```diff
@@ -0,0 +1,142 @@
...
+
+    expectObservable(source.skipWhile(function () { return true; })).toBe(expected);
+  });
+
+  it('should skip all elements with a truthy predicate', function () {
+    var source = hot('-1-^2--3--4--5--6--|');
+    var expected =      '----------------|';
+
+    expectObservable(source.skipWhile(function () { return {}; })).toBe(expected);
+  });
```
staltz: Can you add a test case where the predicate throws an error?

spec/operators/skipWhile-spec.js
```diff
@@ -0,0 +1,142 @@
...
+  it('should be possible to skip using the element\'s index', function () {
+    var source = hot('--a--b-^-c--d--e--f--g--h--|');
+    var expected =          '--------e--f--g--h--|';
+
+    var predicate = function (v, index) {
+      return index < 2;
+    };
+
+    expectObservable(source.skipWhile(predicate)).toBe(expected);
+  });
```
staltz: Would be good to add subscription assertions for all these test cases.

spec/operators/skipWhile-spec.js
jinroh: Forgot this one !

spec/operators/skipWhile-spec.js
jinroh: Done.

jinroh:
In RxJS4, the predicate function [takes as third argument](https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/linq/observable/skipwhile.js#L29) (after the value and index) the observable object being subscribed...

I'm not sure what is the purpose of this and could not find any... Do you think we need to keep it ?

src/operators/skipWhile.ts
```diff
@@ -0,0 +1,47 @@
...
+  private skipping: boolean = true;
+  private index: number = 0;
+
+  constructor(destination: Subscriber<T>,
+              private predicate: (x: T, index: number) => boolean) {
+    super(destination);
+  }
+
+  _next(value: T): void {
+    const destination = this.destination;
```
staltz: Like for everything else, you could store `const skipping = this.skipping` since it's used multiple times.

jinroh: Not really because it is mutated on line 40.

staltz: Hmm, correct.

src/operators/skipWhile.ts
```diff
@@ -0,0 +1,47 @@
...
+import {tryCatch} from '../util/tryCatch';
+import {errorObject} from '../util/errorObject';
+import {bindCallback} from '../util/bindCallback';
+
+export function skipWhile<T>(predicate: (x: T, index: number) => boolean, thisArg?: any) {
+  return this.lift(new SkipWhileOperator(predicate, thisArg));
+}
+
+class SkipWhileOperator<T, R> implements Operator<T, R> {
+  predicate: (x: T, index: number) => boolean;
```
kwonoj: maybe this can be private? minor thing though.

kwonoj:
I think PR's OK. @jinroh , you're fast! :)

blesh:
> I'm not sure what is the purpose of this and could not find any... Do you think we need to keep it ?

I believe we kept it in `filter` and `map` as well.. so we should probably keep it here too... however, I totally agree that it's unnecessary.

src/operators/skipWhile.ts
```diff
@@ -0,0 +1,47 @@
+import {Operator} from '../Operator';
+import {Subscriber} from '../Subscriber';
+import {tryCatch} from '../util/tryCatch';
+import {errorObject} from '../util/errorObject';
+import {bindCallback} from '../util/bindCallback';
+
+export function skipWhile<T>(predicate: (x: T, index: number) => boolean, thisArg?: any) {
```
blesh: This should define a return value of `: Observable<T>`

jinroh:
@blesh I don't think [you](https://github.com/ReactiveX/RxJS/blob/master/src/operators/filter.ts#L44) [did](https://github.com/ReactiveX/RxJS/blob/master/src/operators/map.ts#L44).

I think we should drop this behaviour...

kwonoj:
Actually, there are mixture currently, refer [`count`](https://github.com/ReactiveX/RxJS/blob/master/src/operators/count.ts#L63). Seems it's good time to have conclusion for unified behavior.

src/operators/skipWhile.ts
jinroh: done.

blesh:
Oops didn't mean to close. Fat fingers.

Anyhow...

> I think we should drop this behaviour

That's fine. It really doesn't make much sense anyhow.

kwonoj:
> That's fine. It really doesn't make much sense anyhow.

: +1 too. Maybe can discuss later to find if anyone has usecase.

kwonoj:
I think most of issues in this PR's now settled down, will merge in around hr. (having around hr of buffer just in case if @blesh might have some additional suggestion)

blesh:
@kwonoj :+1: LGTM

kwonoj:
Merged with a2244e0. Appreciate for all effort & taking time, @jinroh

PR #707, opened 2015-11-11
test(FastMap): add test cases for FastMap

kwonoj:
Coverage says line is covered, but by other test cases only without test for fastmap itself. Added test coverage.

### DIFF
--- a/spec/util/FastMap-spec.js
+++ b/spec/util/FastMap-spec.js
@@ -1,0 +1,79 @@
+var FastMap = require('../../dist/cjs/util/FastMap').FastMap;
+
+describe('FastMap', function () {
+  it('should exist', function () {
+    expect(typeof FastMap).toBe('function');
+  });
+
+  it('should accept string as keys', function () {
+    var map = new FastMap();
+    var key1 = 'keyOne';
+    var key2 = 'keyTwo';
+
+    map.set(key1, 'yo');
+    map.set(key2, 'what up');
+
+    expect(map.get(key1)).toBe('yo');
+    expect(map.get(key2)).toBe('what up');
+  });
+
+  it('should allow setting keys twice', function () {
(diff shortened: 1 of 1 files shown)


src/util/FastMap.ts
```diff
@@ -1,23 +1,31 @@
...
   delete(key: string): boolean {
-    this._values[key] = null;
+    this.values[key] = null;
+    this._size--;
     return true;
   }
 
   set(key: string, value: any): FastMap {
-    this._values[key] = value;
+    if (!this.get(key)) {
```
kwonoj: I'm feeling this one's bit debatable since makes `set` slower.

- maybe let `size` simply returns `values.length` directly?
- or get rid of `size` interface?

It'll depends on usecase of `FastMap`, if we value efficiency of `get/set` operation, or expect `size` could be used frequently. So far, I haven't seen usage of `FastMap::size` in current codebase.

blesh: If we're not using `size` we can just get rid of it. There's no reason to assert and increment if we're not even using the value we're incrementing. Extra overhead for no reason.

kwonoj: Make sense. I'll get rid of `size` interface for now. It can be easily introduced once it's required.

kwonoj:
Now `fastMap` dropped `size` interface. Will be added once there's proper usecases.

kwonoj:
I think rest of changes are fine but will merge around late tonight to tomorrow morning to see if there's any other suggestions.

kwonoj:
Updated changes to not to include related with PR directly.

kwonoj:
Note : for discussions related with unused of `size` will be created as separate PR for refactoring only.

kwonoj:
Merged with af3d75a .

Issue #710, opened 2015-11-12
Source for ShareReplay accidentally re-introduced?


larrifax:
I just noticed that the source for shareReplay still exists in [src/operators/shareReplay.ts](https://github.com/ReactiveX/RxJS/blob/master/src/operators/shareReplay.ts). This was supposed to be removed in d0adf57fe2ac18736ee1b6e63e98d87108f70b8c, wasn't it? e1966574715bc526c4c0da9262699187d1a7fbcb seems to re-introduce it. Maybe a merge gone wrong there?

staltz:
Good catch. I'll remove it.

kwonoj:
I think this can be closed now. Please freely reopen if there's necessary task remains.

sbabeal:
Why were this members removed.  How to i get the same behaviour with auto connect.  I don't have to have to call connect every time I make a subscription because I don't know my callers

trxcllnt:
@sbabeal you can achieve the same behavior by calling `source.publishReplay().refCount()`. The share operators were removed to highlight the fact that the old share operator behaviors don't support retrying or repeating.

sbabeal:
Let me give it a try.  When I moved over to publishReplay(), the code running under the subscription never gets called.

const connectionObservable = rx.Observable.create((observer: rx.Observer<SignalRConnection>) => {
    writeToLog("creating new connection for " + url);
    const conn = new SignalRConnection(url);
    observer.next(conn);
    // need to add disconnection logic here.
}).publishReplay(1);

So when I call connectionObservable.subscribe( x => ….) the code in yellow never gets run.

If I remove publishReplay(1).  Then it works as expected.  My issue is that the next subscriber doesn’t get the SignalRConnection since its was already published.

I’ll write back in a second to let you know if it worked.


-Brandt

sbabeal:
Dude you’re awesome!!! Works perfectly.  Thanks

-Brandt

PR #712, opened 2015-11-12
fix(operators): Remove shareReplay and shareBehavior

staltz:
Remove two operators that were reintroduced by mistake. shareReplay and shareBehavior should not
exist anymore.

Resolves issue #710.


kwonoj:
Merged with 536a6a6, thanks @staltz

PR #713, opened 2015-11-12
debounce: more tests and some fixes

staltz:
Add some missing tests for debounce operator. While at it, I found some bugs with debounce, which are also fixed in this PR.

"Fix debounce operator to unsubscribe the ongoing duration Observable if the outer subscriber gets a
new 'next' value. Also fix the `clearDebounce()` method which was doing a null check, and is now doing
a truthyness check, to catch also undefined cases."


kwonoj:
:+1:

kwonoj:
Merged with dea7847, thanks @staltz

Issue #594, opened 2015-10-24
forkJoin operator doesn't accept an array of observables


sumigoma:
ForkJoin doesn't seem to maintain the behavior from Rx4 of supporting both .forkJoin(a,b,c) and .forkJoin([a,b,c]). If you try to do the latter (an array of observables) it gives the exception 
```
TypeError: observables[i].subscribe is not a function
    at ForkJoinObservable.execute.ForkJoinObservable._subscribe (ForkJoinObservable.ts:19)
```

endash:
I'm just gonna comment here because my issue is virtually identical. This is also an issue with `combineLatest`. In Rx4 you could pass an array of observables followed by a projection function. In Rx5 you can only use the arguments list. It is very common to have an array of observables rather than a set number of variables of type Observable.

trxcllnt:
@blesh I also think this is a concern. Unfortunately, TypeScript isn't exactly helpful in this regard. The only solution is to splat the Array when you call forkJoin/combineLatest, which you can only do in es6/ typescript.

I would be ok with a call to "flatten" arguments into a flat list of Observables, no matter how deeply-nested.

kwonoj:
I was thinking about this bit, as @trxcllnt mentioned it requires some additional code to make it work with typescript. Trying to create some changes to trigger perf comparison to see if it's not hurting.

blesh:
I'm sort of indifferent about this one.

We can try adding the checks and see how it affects the performance. If it's a hit, then I say we don't do it, and people can just use `.apply` like they would on any other function in JavaScript.

kwonoj:
>  how it affects the performance

: That's I'd like to verify :) if changes does not hurt performance, supporting this would be good. I should come up with numbers.

kwonoj:
Initial attempt blocked due to same reason as https://github.com/ReactiveX/RxJS/issues/660, property based stub does not allow overload.

PR #714, opened 2015-11-12
feat(forkJoin): accept array of observable as parameter

kwonoj:
relates to #594 

This PR enables `forkJoin` accepts array of observable.

Creating overloads of function having array parameter to rest parameter requires some explicit unfolds due to behaviors of typescript (bit more explanation can be found [here](https://www.stevefenton.co.uk/2013/11/allowing-array-or-rest-parameters-using-overloads-in-typescript/), way I used can be found in this PR as well.)

Performance wise, I haven't noticed big impact for general usecases of using rest parameters. 
(numbers may vary). Checking typeof array for single observable, or array might impact bit more though.

**before**
```
     RxJS 4.0.6 |            RxJS 5.0.0-alpha.10 |          factor |      % improved
------------------------------------------------------------------------------------
38,677 (±4.83%) |               411,112 (±4.14%) |          10.63x |          962.9%
```

**after**
```
     RxJS 4.0.6 |            RxJS 5.0.0-alpha.10 |          factor |      % improved
------------------------------------------------------------------------------------
40,614 (±4.33%) |               406,452 (±4.23%) |          10.01x |          900.8%
```


src/observables/ForkJoinObservable.ts
```diff
@@ -3,41 +3,46 @@ import {Subscriber} from '../Subscriber';
...
       return new EmptyObservable();
     }
-    return new ForkJoinObservable(sources);
-  }
 
-  private getResultSelector(): (...values: Array<any>) => any {
-    const sources = this.sources;
+    let resultSelector: (...values: Array<any>) => any = null;
+    if (typeof sources[sources.length - 1] === 'function') {
+      resultSelector = sources.pop();
```
kwonoj: Appreciate for suggestion of mutating sources directly @blesh , this seems way better than I did originally. I should've tried to look existing references before submitting PR.

src/observables/ForkJoinObservable.ts
```diff
@@ -3,41 +3,41 @@ import {Subscriber} from '../Subscriber';
...
+    }
 
-    let resultSelector = sources[sources.length - 1];
-    if (typeof resultSelector !== 'function') {
-      return null;
+    // if the first and only other argument besides the resultSelector is an array
+    // assume it's been called with `forkJoin([obs1, obs2, obs3], resultSelector)`
+    if (sources.length === 1 && isArray(sources[0])) {
+      sources = <Array<Observable<any>>>sources[0];
     }
```
kwonoj: Seems previous comment went away by updating. Appreciate @blesh for suggestion, it seems much better than I did first time. :)

src/observables/ForkJoinObservable.ts
```diff
@@ -3,41 +3,41 @@ import {Subscriber} from '../Subscriber';
...
+    super();
   }
 
   static create(...sources: Array<Observable<any> |
+                                  Array<Observable<any>> |
                                   Promise<any> |
-                                  ((...values: Array<any>) => any)>)
-                                  : Observable<any> {
-    if (sources === null || sources.length === 0) {
+                                  ((...values: Array<any>) => any)>) : Observable<any> {
```
kwonoj: After thinking bit more with updated code, came to think in this case function overload signature does not do much thing, simplified by having single signature does polymorphic.

kwonoj:
Updating code lint-friendly makes one comment goes away :( I should buy mac.
Comment was 

- removed overload signature, after code update overloaded signature actually doesn't do much thing, simplified into single signature and let it does polymorphics

Updated perf shows quite similar to original code

```
      RxJS 4.0.6 |            RxJS 5.0.0-alpha.10 |          factor |      % improved
-------------------------------------------------------------------------------------
40,950 (±19.35%) |               437,477 (±4.65%) |          10.68x |          968.3%
```

PR #718, opened 2015-11-13
chore(test): align test case location for util implementation

kwonoj:
Align placement for `util` implementation as same as other cases, this one only placed under `utils`.


kwonoj:
Merged with 0fae107.

Issue #465, opened 2015-10-06
lint_src (tslint) does not work on windows yet


kwonoj:
Creating issues to avoid possible confusions if someone new to codebase try to build on windows platform.

This issue is not because of RxJS codebase, occurs due to tslint (https://github.com/palantir/tslint) does not support explicit directory expanding other than glob match.

https://github.com/palantir/tslint/issues/688
https://github.com/palantir/tslint/issues/692

In short, while waiting tslint to be updated it will be bit difficult to windows users to run lint to source files. I've thought about creating wrapper script a bit but haven't reached conclusion if this is necessary instead of waiting tslint to be updated.

staltz:
Another option is running linting only in CI, and optionally locally. So, when creating PRs, we will still be able to see lint errors.

kwonoj:
I'd still like to suggest to lint as strict manner to be passed in CI server. At this moment, I'm not sure how many contributors are being impacted by this yet to decide lint as optional. (yes, I'm running windows though :wink:) It may possible to decide if lint can be optional or need to introduce interim script to aid windows platform users once we get to know numbers of people having problems.

kwonoj:
+Missing part, just build / run test is still works on windows platform by separated command. The only part windows users might have difficulties is once CI complains about lint, has to iterate between local edit-CI to fix lint issues.

kwonoj:
Recently added `npm cover` invokes `istanbul` has similar problem.

https://github.com/gotwarlost/istanbul/issues/13

blesh:
@kwonoj ... is this still an issue?

kwonoj:
Yes. tslint targeting this on 4.0 milestone, do not specified date yet. Maybe this kind can be updated in documentation and issue can be closed?

PR #719, opened 2015-11-13
chore(package.json): update coverage script to run on windows

kwonoj:
relates to #465 

This PR makes coverage script run on windows also. Still `tslint` is not available at this moment, looking for possible solutions.


kwonoj:
Merged with fed1ad9.

Issue #678, opened 2015-11-09
Bug: shared Observables not actually repeatable/retryable


staltz:
```js
var source = Rx.Observable.interval(100).take(3)
  .concat(Rx.Observable.throw(new Error('BOOM')))
  .share();

source.retry(3).subscribe(
  x => console.log(x), 
  e => console.error(e.message), 
  () => console.log('|'));

source.retry(3).subscribe(
  x => console.log('   '+x), 
  e => console.error('   '+e.message), 
  () => console.log('   |'));
```
Runs as
```
0
   0
1
   1
2
   2
```
while we expected
```
0
   0
1
   1
2
   2
0
   0
1
   1
2
   2
0
   0
1
   1
2
   2
BOOM
   BOOM
```

I'm investigating some solutions in ConnectableObservable to be able to support @blesh's use cases for retryable/repeatable Observables, but it seems like this is a bug.

blesh:
Weird, I thought I had tests around that from the beginning. It's probably my most recent changes to that that broke the behavior. I was using `share` and `retry` in my AngularConnect example (alpha.7 I think), and it was fine.

staltz:
We have this test:
```js
  it('should retry just fine', function () {
    var e1 =  cold('---a--b--c--d--e--#');
    var expected = '---a--b--c--d--e-----a--b--c--d--e--#';

    expectObservable(e1.share().retry(1)).toBe(expected);
  })
```
But notice how there is just one subscriber in this case.

staltz:
Also, this test doesn't have multiple subscribers: https://github.com/ReactiveX/RxJS/blob/master/spec/operators/multicast-spec.js#L88

blesh:
Hmm... it looks like `retryWhen` works fine, but the other two do not. My example was using `retryWhen`.

It doesn't seem like a problem with ConnectableObservable, my suspicion lies with the Subscribers for `repeat` and `retry`. I suspect they're being disabled by a flag when the error or completion hits them.

blesh:
> Also, this test doesn't have multiple subscribers: https://github.com/ReactiveX/RxJS/blob/master/spec/operators/multicast-spec.js#L88

It looks like it resubscribes whenever it completes: https://github.com/ReactiveX/RxJS/blob/e9b13d0178a7ebad2cccd217fc0249449804831d/spec/operators/multicast-spec.js#L102

[...]

PR #720, opened 2015-11-13
Add multicast tests and fix ConnectableObservable.

### LINKED ISSUE #657: comprehensive tests for multicast

staltz:
Resolves #657 and solves bug #678. Took some time because #678 was truly a nightmare.

### DIFF
--- a/src/observables/ConnectableObservable.ts
+++ b/src/observables/ConnectableObservable.ts
@@ -1,6 +1,7 @@
 import {Subject} from '../Subject';
 import {Observable} from '../Observable';
 import {Subscription} from '../Subscription';
+import {Subscriber} from '../Subscriber';
 
 export class ConnectableObservable<T> extends Observable<T> {
 
@@ -24,7 +25,16 @@ export class ConnectableObservable<T> extends Observable<T> {
     return (this.subject = this.subjectFactory());
   }
 
-  connect() {
+  connect(onSubscribe?: (subscription: Subscription<T>) => void): Subscription<T> {
+    if (onSubscribe) {
+      this._callbackConnect(onSubscribe);
+      return null;
+    } else {
+      return this._returningConnect();
+    }
+  }
+
+  _returningConnect(): Subscription<T> {
     const source = this.source;
     let subscription = this.subscription;
     if (subscription && !subscription.isUnsubscribed) {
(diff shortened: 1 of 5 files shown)


staltz:
By the way, @trxcllnt you might want to review this if I'm doing any performance mistake. At least there is one closure introduced there. I focused only on getting the bug fixed, we might want to do a second round to make it performant if there's opportunity for that.

staltz:
I just rebased, but now as you can see, tests fail. And the fix isn't immediate. I'm solving some concurrency issues now in ConnectableObservable. I don't know exactly which change to master happened that affected this.

src/observables/ConnectableObservable.ts
```diff
@@ -62,24 +92,60 @@ class RefCountObservable<T> extends Observable<T> {
 
   _subscribe(subscriber) {
     const connectable = this.connectable;
-    const subscription = connectable.subscribe(subscriber);
-    if (++this.refCount === 1) {
-      this.connection = connectable.connect();
+    const refCountSubscriber = new RefCountSubscriber(subscriber, this);
+    refCountSubscriber.myConnection = this.connection;
```
blesh: why is this not added via the constructor above it?

src/observables/ConnectableObservable.ts
```diff
@@ -62,24 +92,60 @@ class RefCountObservable<T> extends Observable<T> {
...
+
+  _complete() {
+    this._resetConnectable();
+    this.destination.complete();
+  }
+
+  _resetConnectable() {
+    const observable = this.refCountObservable;
+    const myConnection = this.myConnection;
+    if (myConnection && myConnection === observable.connection) {
```
blesh: could probably just get the `connection` out of `observable` in a const above, `const connection = observable.connection`

src/observables/ConnectableObservable.ts
staltz: It can. What's more important is line 101.

src/observables/ConnectableObservable.ts
staltz: Will do.

staltz:
@blesh I refactored it.

blesh:
haha... staltz, this is perfect, but I have _one_ more request. Can you flatten the fix commits? :)

staltz:
Yes sir I can. I made them separate in the first place because they hold a lot of story (just check the commit message) on how and why did I make those changes.

staltz:
@blesh Flat it is.

blesh:
:+1:  LGTM ... merging in 3... 2... 1...

blesh:
merged with 20b9a8c... thanks @staltz!

staltz:
That was one hell-ov-a PR.

PR #726, opened 2015-11-14
Add ES7 start event handler to Observer, Observable and Subject

blesh:
See commit messages for more detailed explanations, but the high level is this:

`start` is a handler on `Observer` that is called synchronously on subscription _before_ the Observable's initializer is called. This enables "synchronous unsubscribe". This is all part of the ES7 Observable spec (issue around it can be found here: https://github.com/zenparsing/es-observable/issues/65)

For Subjects, after some discussion with @abersnaze of RxJava noteriety, it was determined that the basic `Subject` should noop incoming start messages it _observes_. However, when `subscribe` is called on the Subject, the start handler on the passed observer will be called prior to adding that observer to the Subject's `observers` list.

Since this is a critical change, with interesting discoveries...
cc/ @trxcllnt @staltz @trxcllnt @mattpodwysocki @zenparsing @jhusain @kwonoj

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -84,7 +84,8 @@ export class Observable<T> implements CoreOperators<T>  {
    */
   subscribe(observerOrNext?: Observer<T> | ((value: T) => void),
             error?: (error: T) => void,
-            complete?: () => void): Subscription<T> {
+            complete?: () => void,
+            start?: (subscription: Subscription<T>) => void): Subscription<T> {
 
     let subscriber: Subscriber<T>;
 
@@ -96,9 +97,10 @@ export class Observable<T> implements CoreOperators<T>  {
       }
     } else {
       const next = <((x?) => void)> observerOrNext;
-      subscriber = Subscriber.create(next, error, complete);
+      subscriber = Subscriber.create(next, error, complete, start);
     }
 
+    subscriber.start(subscriber);
     subscriber.add(this._subscribe(subscriber));
 
     return subscriber;
--- a/spec/observable-spec.js
+++ b/spec/observable-spec.js
@@ -91,6 +91,13 @@ describe('Observable', function () {
       expect(mutatedByComplete).toBe(true);
     });
 
+    it('should accept a fourth argument for a start handler', function () {
+      Observable.create(function _subscriber(observer) {
+        expect(observer.isUnsubscribed).toBe(true);
+      })
+      .subscribe(null, null, null, function start(subscription) { subscription.unsubscribe(); });
+    });
+
     it('should return a Subscription that calls the unsubscribe function returned by the subscriber', function () {
       var unsubscribeCalled = false;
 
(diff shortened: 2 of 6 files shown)


src/Observable.ts
```diff
@@ -96,9 +97,10 @@ export class Observable<T> implements CoreOperators<T>  {
       }
     } else {
       const next = <((x?) => void)> observerOrNext;
-      subscriber = Subscriber.create(next, error, complete);
+      subscriber = Subscriber.create(next, error, complete, start);
     }
 
+    subscriber.start(subscriber);
     subscriber.add(this._subscribe(subscriber));
```
trxcllnt: @blesh should this line be protected with a conditional? `if (!subscriber.isUnsubscribed) { ... }`

blesh: Subscribers should always have a start method.  The conditional is in the Subscriber.

trxcllnt: @blesh yes, but if the `start` method synchronously calls `unsubscribe` (thus setting `isUnsubscribed` to `true`), then there's no reason to execute line 104 (`subscriber.add(this._subscribe(subscriber));`).

blesh: `add` checks `isUnsubscribed` internally. I could do that though.

staltz:
I reviewed all commits. Looks good and useful :+1:

src/Observable.ts
trxcllnt: @blesh yes, but that happens after `_subscribe` runs. Don't you think it'd be more efficient if we didn't even execute the subscription if all we do is immediately throw it all away?

blesh: Sure.

blesh: Well, I thought part of the purpose was to actually call the `_subscribe` and let user logic decide what to do.

kwonoj:
Change looks OK for me.

src/Observable.ts
trxcllnt: @blesh ah, I wasn't aware of that requirement. I'm thinking if subscribing is referentially transparent, and subscribing performs some action, then unsubscribing performs the inverse, then synchronous subscribe + unsubscribe is equivalent to not even subscribing in the first place. Does that sound right? What could the subscription do with a subscriber that's already disposed by the time it's called?

zenparsing: In es-observable, I'm not calling the subscribe function if the subscription is already closed.  I think it would be surprising for users to receive a closed subscription.

```js
    // I'll probably write this assuming that subscriber represents an active subscription
});
```

blesh: Oh? For some reason I thought that was part of this feature. "Look it's unsubscribed, I should do something different (like error or notify someone)"

I'm fine with this change though, it's minimal.

blesh:
I'm going to leave this one for a few more days while I wait to hear back from @jhusain and @zenparsing on what's happened with the spec at the TC39 this week.

blesh:
Closing this one for now, as it doesn't look like it's going to get passed the committee.

staltz:
Sad to have coded it in vain.

PR #728, opened 2015-11-15
refactor(repeat): remove logically unreachable code

kwonoj:
`repeat` returns `EmptyObservable` immediately if count is zero, doesn't need to check count inside of `FirstRepeatSubscriber` again. This PR removes logically unreachable code from subscribe implementation.

### DIFF
--- a/src/operators/repeat.ts
+++ b/src/operators/repeat.ts
@@ -29,28 +29,24 @@ class FirstRepeatSubscriber<T> extends Subscriber<T> {
               private count: number,
               private source: Observable<T>) {
     super(null);
-    if (count === 0) {
-      this.destination.complete();
-      super.unsubscribe();
-    }
     this.lastSubscription = this;
   }
 
-  _next(value: T) {
+  _next(value: T): void {
     this.destination.next(value);
   }
 
-  _error(err: any) {
+  _error(err: any): void {
     this.destination.error(err);
   }
 
-  complete() {
+  complete(): void {
     if (!this.isUnsubscribed) {
       this.resubscribe(this.count);
     }
   }
 
-  unsubscribe() {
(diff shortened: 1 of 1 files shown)


staltz:
Good catch

src/operators/repeat.ts
```diff
@@ -76,15 +72,15 @@ class MoreRepeatSubscriber<T> extends Subscriber<T> {
...
   }
 
-  _next(value: T) {
+  _next(value: T): void {
     this.parent.destination.next(value);
   }
 
-  _error(err: any) {
+  _error(err: any): void {
     this.parent.destination.error(err);
```
kwonoj: This one also seems unreachable but I wasn't completely sure. @staltz , would like to ask opinions, is there usecases of `FirstRepeatSubscriber` doesn't raises error than `MoreRepeatSubscriber` does?

staltz: Sure there are cases. An Observable execution is not always deterministic. Just imagine 
```js
source.map(() => {
  var random = Math.random();
  if (random === SOME_MAGIC_NUMBER) {
    throw new Error();
  } else {
    return random;
  }
})
```

The first execution may complete, but the second may not.

kwonoj: Very correct. I thought in a way of deterministic using fixed scalar values all time. Then this line requires test coverage expanded, I'll update that.

staltz: :+1: Good point to cover all cases

kwonoj:
PR updated to include expanded test coverage of `repeat` operator.

kwonoj:
I'll check in changes myself if there isn't additional suggestions for PR.

blesh:
This looks good. Nice work @kwonoj

PR #729, opened 2015-11-15
feat(callback): Add Observable.fromCallback

SomeKittens:


SomeKittens:
Ran into an odd situation.  With the additional tests included in this PR, I get this error:

```
Failures:
1) Observable.prototype.bufferWhen should NOT handle hot inner empty
  Message:
    RangeError: Maximum call stack size exceeded
  Stack:
    RangeError: Maximum call stack size exceeded
```

removing any one of the tests (or any test in `from-promise-spec.js`, but not `debouncetime-spec.js`, the other two I tried) results in `bufferWhen` passing.

I'm guessing it's a race condition somewhere, with the extra tests, the tests are running long enough to trigger the overflow.  Changing the test to the following results in it passing though I'm not sure if it invalidates the test:

```
  // bufferWhen is not supposed to handle a factory that returns always empty
  // closing Observables, because doing such would constantly recreate a new
  // buffer in a synchronous infinite loop until the stack overflows. This also
  // happens with buffer in RxJS 4.
  it('should NOT handle hot inner empty', function (done) {
    var source = Observable.of(1, 2, 3, 4, 5, 6, 7, 8, 9);
    var closing = Observable.empty();
    var TOO_MANY_INVOCATIONS = 30;

    var invoked = 0;
    source
      .bufferWhen(function () {
        if (invoked <= TOO_MANY_INVOCATIONS) {
          return closing;
        }
        return Observable.interval(100);
      })
      .subscribe(function (val) {
        invoked++;
        if (invoked > TOO_MANY_INVOCATIONS) {
          return done();
        }
        expect(Array.isArray(val)).toBe(true);
        expect(val.length).toBe(0);
      }, null, null);
  });
```

SomeKittens:
Ah, missed linting, fixing that now.

src/observables/CallbackObservable.ts
```diff
@@ -0,0 +1,101 @@
+import {Observable} from '../Observable';
+import {Subscriber} from '../Subscriber';
+import {Scheduler} from '../Scheduler';
+import {Subscription} from '../Subscription';
+import {immediate} from '../schedulers/immediate';
+import {tryCatch} from '../util/tryCatch';
+import {errorObject} from '../util/errorObject';
+
+export class CallbackObservable<T> extends Observable<T> {
```
kwonoj: What about align class name to observable, such as `Observable.fromEvent` has implementation named `FromEventObservable`?

SomeKittens:
Tried to match things to what's going on in `PromiseObservable.ts` with regards to scalar things and schedulers.  Not 100% confident, I'd appreciate it if you'd go over that section with a fine-tooth comb.

src/observables/CallbackObservable.ts
SomeKittens: I tried to mimic `PromiseObservable` with regards to that sort of thing.

kwonoj: Yes, just noticed that too.

src/observables/CallbackObservable.ts
```diff
@@ -0,0 +1,101 @@
...
+              subscriber.next(innerArgs[0]);
+            } else {
+              subscriber.next(innerArgs);
+            }
+          }
+          subscriber.complete();
+        };
+      }
+    } else {
+      let subscription = new Subscription();
```
kwonoj: this variable seems could be `const`

src/Observable.ts
```diff
@@ -146,6 +146,7 @@ export class Observable<T> implements CoreOperators<T>  {
                                removeHandler: (handler: Function) => void,
                                selector?: (...args: Array<any>) => T) => Observable<T>;
   static fromPromise: <T>(promise: Promise<T>, scheduler?: Scheduler) => Observable<T>;
+  static fromCallback: <T>(callbackFunc: Function, ctx: Object, selector: Function, scheduler: Scheduler) => Function;
```
kwonoj: parameter other than `callbackFunc` is allowed to be optional in signature of `create`, 

`create<T>(callbackFunc: Function, ctx: Object = undefined, selector: Function = undefined, scheduler: Scheduler = immediate)`

where this signature doesn't.

kwonoj:
Appreciate for effort, I didn't even noticed `fromCallback` is yet implemented :) I'll try to add some comment.

src/observables/CallbackObservable.ts
SomeKittens: Fixed

src/Observable.ts
SomeKittens: Fixed

SomeKittens:
Hello!  Any update on this?

kwonoj:
I am generally ok with current changes, but would like to ask @blesh 's opinion also may have better suggestions than me. Also, testing failure you've mentioned makes this as bit blocked, since those issue need to be addressed and resolved to correctly merge this PR.

SomeKittens:
Yep, willing to do whatever's needed to get those tests passing but I'm stumped.  I can put in the fix I mentioned if you or @blesh are ok with it (not sure if it breaks the spirit of the test).

blesh:
Sometimes async tests in `jasmine` get corrupt somehow. I'm not entirely sure what causes it, but I've seen it before. Anecdotally, I've pulled the PR down locally and it runs fine.

However, I don't think this PR matches with what `fromCallback` did previously. I'll add some comments to the commits.

blesh:
> However, I don't think this PR matches with what fromCallback did previously. 

... ignore that, I was just confused by one of the tests. Perhaps the test just needs refactored a little for readability.

SomeKittens:
@blesh Fixed, test is more specific now.

src/observables/CallbackObservable.ts
```diff
@@ -0,0 +1,103 @@
...
+  _isScalar: boolean = false;
+  value: T | T[];
+
+  static create<T>(callbackFunc: Function, ctx: Object = undefined, selector: Function = undefined, scheduler: Scheduler = immediate): Function {
+    return (...args): Observable<T> => {
+      return new CallbackObservable(callbackFunc, ctx, selector, args, scheduler);
+    };
+  }
+
+  constructor(private callbackFunc: Function, private ctx, private selector, private args: any[], public scheduler: Scheduler = immediate) {
```
staltz: Can you please give a newline for each argument? Like we do elsewhere, see e.g. https://github.com/ReactiveX/RxJS/blob/master/src/observables/IteratorObservable.ts#L61-L65

src/observables/CallbackObservable.ts
```diff
@@ -0,0 +1,103 @@
...
+    super();
+  }
+
+  _subscribe(subscriber: Subscriber<T | T[]>) {
```
staltz: This function is really big with lots of nested if statements. Can you refactor?

spec/observables/from-callback-spec.js
```diff
@@ -0,0 +1,93 @@
...
+      cb(datum);
+    }
+    var cbToObs = Observable.fromCallback(callback);
+
+    cbToObs(42)
+      .subscribe(function (x) {
+        expect(x).toBe(42);
+      }, null,
+      function (x) {
+        expect(x).toBe(undefined);
```
staltz: Since this is the `complete()` Observer function, it never expects any parameter, so you can remove `function(x)`. Otherwise the test case might be read in a awkward way, **as if** usually `complete()` takes a parameter. Actually you can altogether replace this function with just `done`.

Also, it is customary to use `done.fail` as the `error()` handler. So replacing `null` with `done.fail` is a good idea in order to catch unexpected error that a test case could possibly raise.

spec/observables/from-callback-spec.js
```diff
@@ -0,0 +1,93 @@
...
+
+    setTimeout(function () {
+      expect(nextSpy).not.toHaveBeenCalled();
+      expect(throwSpy).not.toHaveBeenCalled();
+      expect(completeSpy).not.toHaveBeenCalled();
+
+      clearTimeout(timeout);
+      done();
+    });
+  });
```
staltz: I guess for now this async test is ok. When we have Subscriber `start()` feature, we can rework this test.

staltz:
Thanks @SomeKittens for this PR.

spec/observables/from-callback-spec.js
SomeKittens: Added a few comments to make things clearer

src/observables/CallbackObservable.ts
SomeKittens: Fixed

SomeKittens:
@staltz as to the complexity of `_subscribe`, I based this off https://github.com/ReactiveX/RxJS/blob/master/src/observables/PromiseObservable.ts

Each path is slightly unique, not sure what to target for refactor.

SomeKittens:
(also rebased @kwonoj's test fixes into this branch)

spec/observables/from-callback-spec.js
```diff
@@ -0,0 +1,94 @@
...
+    cbToObs()
+      .subscribe(function () {
+        // Considered a failure if we don't go directly to err handler
+        done.fail('should not be called');
+      },
+      function (err) {
+        expect(err.message).toBe('Yikes!');
+        done();
+      },
+      null
```
staltz: This null here should also be done.fail

SomeKittens:
@staltz fixed missed null.

staltz:
Great

SomeKittens:
Anything else?

src/observables/CallbackObservable.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    let handler;
+
+    if (scheduler === immediate) {
+      if (this._isScalar) {
+        subscriber.next(this.value);
+        subscriber.complete();
+      } else {
+        handler = (...innerArgs) => {
+          let results;
```
blesh: you can move `let results` down to where it's being used on line `52`

src/observables/CallbackObservable.ts
```diff
@@ -0,0 +1,110 @@
...
+        };
+      }
+    } else {
+      const subscription = new Subscription();
+      if (this._isScalar) {
+        const value = this.value;
+        subscription.add(scheduler.schedule(dispatchNext, 0, { value, subscriber }));
+      } else {
+        handler = (...innerArgs) => {
+          let results;
```
blesh: same comment as before.

src/observables/CallbackObservable.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (scheduler === immediate) {
+      if (this._isScalar) {
+        subscriber.next(this.value);
+        subscriber.complete();
+      } else {
+        handler = (...innerArgs) => {
+          let results;
+
+          this._isScalar = true;
+          this.value = innerArgs;
```
blesh: The `_isScaler` and `value` properties are not necessary. I suspect this is just cruft from working off of the PromiseObservable. Since promises resolve once, and then cache the value, `_isScalar` and `value` were being set to enable PromiseObservables to fast-track in some operators. This can't be true for CallbackObservable, so you can remove all references to `_isScalar` and `value`.

src/observables/CallbackObservable.ts
```diff
@@ -0,0 +1,110 @@
...
+            } else {
+              subscriber.next(innerArgs);
+            }
+          }
+          subscriber.complete();
+        };
+      }
+    } else {
+      const subscription = new Subscription();
+      if (this._isScalar) {
```
blesh: again, this check/block isn't necessary, since CallbackObservable can't be scalar.

blesh:
Disregard my comments. I never use this creation method.

Frankly I think this is really a weird experience, and I dislike how it works. Every other `fromWhatever` method returns an Observable, not a function that returns a stateful observable.

blesh:
This LGTM. However, I'm not sure about including it in `Core`.

SomeKittens:
@blesh Agreed on it being weird, but that's how it works in Rx4 (and for that matter, Bluebird and the whole promise spectrum).  I'm open to updating the signature to return an observable, possibly passing in the callback args.

SomeKittens:
Sorry to continue pinging everyone, but is there anything else here that needs fixing?  (or is the holdup deciding if this should be in core or not?)

blesh:
> or is the holdup deciding if this should be in core or not?)

The hold up is mostly just me being too busy. :) Reviewing this now.

blesh:
Merged as of f6cebb6, with some minor lint fixes. Thanks @SomeKittens! This was a nice PR.

SomeKittens:
> The hold up is mostly just me being too busy. :) Reviewing this now.

It happens, especially around this time of year.  Thanks!

PR #734, opened 2015-11-16
chore(package.json): bump up typescript version

kwonoj:
- bump up typescript compiler version, compile with --pretty option
- markdown test is moved to travis

Small changes to build script,
- Typescript `pretty` option
: basically this option will prints out additional detail when compile fails like below, help someone does not have editor side compiler support.
image

- markdown test is moved to travis
: markdown test doesn't necessarily need to be executed per every local unit test run, moved into travis.


kwonoj:
I can check in this once got signed off.

blesh:
This needs to be rebased. Outside of that, I'm pretty indifferent about TypeScript versions, as long as we're up to date and things build properly.

kwonoj:
OK. I'll rebase and check this in by myself.

kwonoj:
Merged with 836b140, this requires `npm install` to update typescript compiler.

PR #735, opened 2015-11-16
test(ErrorObservable): add tests for ErrorObservable

kwonoj:
This PR added test coverage for `ErrorObservable`.

PR #737, opened 2015-11-16
test(switchMap): expand test coverage for switchMap

kwonoj:
This PR expands coverage of `switchMap` to cover most of cases.


kwonoj:
Will check in this with https://github.com/ReactiveX/RxJS/pull/735 around tomorrow to see if there's suggestions.

kwonoj:
Merged with 030f35c.

PR #740, opened 2015-11-17
chore(package.json): target closure compiler to ES5

### LINKED ISSUE #692: Closure compiler warnings
This might not be big issue..

after PR https://github.com/ReactiveX/RxJS/pull/664, closure compiler gives warning like below

```
./dist/global/Rx.js:7401: WARNING - Keywords and reserved words are not allowed as unquoted property names in older versions of JavaScript. If you are targeting newer versions of JavaScript, set the appropriate language_in option.
    if (typeof root_1.root.Symbol.for === 'function') {
```

Assume some of generated codes are not supported old version of runtime.

kwonoj:
- closure compiler will target ES5, suppress ES3 specific warnings

closes #692 

This PR changes closure compiler build targets, now all builds are targeting ES5+ as minimum supported runtime.


kwonoj:
Merged with 645a1b2.

Issue #578, opened 2015-10-22
ReplaySubject memory leak


roganov:
`ReplaySubject` would clear its cache of events only when a new observable subscribes, otherwise the array of events grows indefinitely.

Steps to reproduce:
```js
var rs = new ReplaySubject(1);
while (true) {rs.next(1)};
```
Apparently introduced by this https://github.com/ReactiveX/RxJS/commit/488ac2e4708665b71fbd3f3b109766c0e70f3689 commit.

trxcllnt:
@roganov good catch, I knew there was a reason we need to trim on each `next`. @blesh @staltz comments?

staltz:
Yeah you can blame me for taking it out. Certainly looked fishy, perhaps a refactor could help clarify it's not just "getting" the events.

blesh:
@staltz ... has this been fixed?

staltz:
Not yet. I'll do it.

blesh:
Okay, thanks @staltz ... I wasn't asking you to do it, I was just quickly checking to see if I missed that this needed to be closed.

kwonoj:
Closing this issue by PR https://github.com/ReactiveX/RxJS/pull/742 now merged.

PR #742, opened 2015-11-17
Fix ReplaySubject memory leak

staltz:
Resolves issue #578.

### DIFF
--- a/src/subjects/ReplaySubject.ts
+++ b/src/subjects/ReplaySubject.ts
@@ -22,11 +22,12 @@ export class ReplaySubject<T> extends Subject<T> {
   _next(value: T): void {
     const now = this._getNow();
     this.events.push(new ReplayEvent(now, value));
+    this._trimBufferThenGetEvents(now);
     super._next(value);
   }
 
   _subscribe(subscriber: Subscriber<any>): Subscription<T> {
-    const events = this._getEvents(this._getNow());
+    const events = this._trimBufferThenGetEvents(this._getNow());
     let index = -1;
     const len = events.length;
     while (!subscriber.isUnsubscribed && ++index < len) {
@@ -39,7 +40,7 @@ export class ReplaySubject<T> extends Subject<T> {
     return (this.scheduler || immediate).now();
   }
 
-  private _getEvents(now): ReplayEvent<T>[] {
+  private _trimBufferThenGetEvents(now): ReplayEvent<T>[] {
     const bufferSize = this.bufferSize;
     const _windowTime = this._windowTime;
     const events = this.events;


package.json
```diff
@@ -19,16 +19,17 @@
...
-    "build_test": "rm -rf dist/ && npm run lint && npm run build_cjs && jasmine",
+    "build_test": "rm -rf dist/ && npm run lint && npm run build_cjs && npm run jasmine",
     "build_cover": "rm -rf dist/ && npm run lint && npm run build_cjs && npm run cover",
     "build_docs": "./docgen.sh",
     "lint_perf": "eslint perf/ --fix",
     "lint_spec": "eslint spec/**/*.js --fix",
     "lint_src": "tslint -c .tslintrc src/*.ts src/**/*.ts src/**/**/*.ts",
     "lint": "npm run lint_src && npm run lint_spec && npm run lint_perf",
-    "cover": "istanbul cover -x \"*-spec.js index.js *-helper.js spec/helpers/*\" ./node_modules/jasmine/bin/jasmine.js && npm run cover_remapping",
+    "cover": "istanbul cover -x \"*-spec.js index.js *-helper.js spec/helpers/*\" npm run jasmine && npm run cover_remapping",
```
kwonoj: actually specifying path to `jasmine` directly was to support windows also, though it is not clean as using common script command. Would you able to make changes to original one to have windows compatibility?

staltz: Did you notice line 31 where the path do jasmine directly is still used?

kwonoj: Yes. I would say it's ugly and weiredness of windows.. Long story in short,

`istanbul ./node_modules/..../jasmine.js` works,
```
istanbul npm run jasmine
npm jasmine : node ./node_modules/..../jasmine.js
```
doesn't works.

kwonoj:
Personally I would consider those test as different category then unit tests, as same as we do for perf tests. One thing not sure is will we expand stress / memory tests for any other operators.

staltz:
It's probably a really good idea to add memory tests for other operators. I could update this PR to separate the memory/stress tests from `spec`. What does @blesh think?

kwonoj:
> It's probably a really good idea to add memory tests for other operators

: I'm also +1 for this at least for some heavyweight operators.

staltz:
Yeah but some operators have unbounded buffers, like zip, so just need to keep in mind they can eat memory indefinitely.

kwonoj:
> so just need to keep in mind they can eat memory indefinitely.

: I think that's 'known behavior`, isn't it? ;) Each cases can be discussed in further to organize tests correctly, in general I think it'd be good compare with RxJS4 to not behave worse at least.

blesh:
@staltz ... we might be able to leverage benchpress tests for memory leak testing. @jeffbcross what do you think?

package.json
staltz: By the way, how is it possible that this works in Windows when there is `&& npm run cover_remapping` at the end?

kwonoj: Not sure if I understood question correctly, chaining command via `&&` works with windows command prompt. Problem in here is `istanbul` does not able to consume supplied parameter correctly only on windows, causes trouble.

staltz: Ok, didn't know `&&` worked in windows. Then we should be fine.

staltz:
I updated the PR to make istanbul work on Windows, but we might scratch the whole memory-tests-in-jasmine approach if we are to use benchpress or something else for memory leak tests.

kwonoj:
Seems so. What about narrow down this PR to include fix only, and create separate issue for discuss where / how to place those kind of tests? Once decision's made, test can be added at the moment.

spec/subjects/replay-subject-spec.js
```diff
@@ -206,4 +206,21 @@ describe('ReplaySubject', function () {
...
+    global.gc();
+    var memoryBefore = process.memoryUsage().heapUsed;
+
+    for (var i = 0; i < 100; i++) {
+      replaySubject.next(event.slice());
+    }
+    global.gc();
+    var memoryAfter = process.memoryUsage().heapUsed;
+
+    expect(memoryAfter / memoryBefore).toBeLessThan(1.2);
```
blesh: I'm a little concerned about the potential for brittleness here in this test. Our tests could start failing because of a defect in a newly deployed runtime.

staltz: Yeah, it's a valid concern. Suggestions? Remove this test altogether perhaps?

kwonoj: Thought this test belongs kind of 'memory benchmark' test, maybe isolate to not to be executed with unittest?

staltz: So, should we
- Make separate jasmine specs directory just for memory tests?
- Wait for Jeff to help with setting up benchpress memory tests?
- Drop memory tests?

blesh:
I think drop the memory tests for now. It's not a bad *idea*, but I think they'll just be too brittle for automation.

staltz:
Will do.

staltz:
It's done.

kwonoj:
I'm merging this, let's consider memory benchmark as separate issue.

kwonoj:
Merged with 0a73b4d, thanks @staltz

PR #743, opened 2015-11-17
chore(micro-perf): ensure csv parameter available

kwonoj:
- let test runner does not prints error due to csv parameter is not
  supplied

This PR makes micro perf test runner does not print out error when run `npm run perf_micro` without supplying csv output parameter, by having additional checking.


kwonoj:
Merged with 1127f13.

PR #744, opened 2015-11-18
fix(last): accept thisArg parameter

kwonoj:
- last operator accepts thisArg for predicate function
- update test cases accordingly

Found differences between type definition in `CoreOperators` to actual implementation, `last` operator does not accept `thisArg` correctly. Updated implementation to accept parameter also expanded test coverage.

### DIFF
--- a/perf/micro/immediate-scheduler/operators/last-predicate-this.js
+++ b/perf/micro/immediate-scheduler/operators/last-predicate-this.js
@@ -9,7 +9,7 @@ module.exports = function (suite) {
   var testThis = {};
 
   var oldLastPredicateThisArg = RxOld.Observable.range(0, 50, RxOld.Scheduler.immediate).last(predicate, testThis);
-  var newLastPredicateThisArg = RxNew.Observable.range(0, 50).last(predicate, testThis);
+  var newLastPredicateThisArg = RxNew.Observable.range(0, 50).last(predicate, null, testThis);
 
   function _next(x) { }
   function _error(e) { }
--- a/spec/operators/last-spec.js
+++ b/spec/operators/last-spec.js
@@ -41,7 +41,21 @@ describe('Observable.prototype.last()', function () {
   it('should return a default value if no element found', function () {
     var e1 = Observable.empty();
     var expected = '(a|)';
-    expectObservable(e1.last(null, null, 'a')).toBe(expected);
+    expectObservable(e1.last(null, null, null, 'a')).toBe(expected);
+  });
+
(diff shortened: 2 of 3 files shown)


staltz:
PR looks fine as usual.

A bit of "meta" discussion though is: starts to look rather unergonomic to have defaultValue param as the last, so one might need to write `source.last(null, null, null, 'default')`. RxJS 4 has no resultSelector, and as an alternative you can give a settings object, allowing you to write `source.last({default: 'foo'})`. Does the settings parameter idea incur additional performance costs?

kwonoj:
That's interesting point. Thinking about we're already doing similar for `Observable.subscribe` it's doable options, question might be how many operators we want to support those behavior. (+ performance impact as you mentioned). Maybe better to create separate issue to have some opinions?

kwonoj:
Created issue for anonymous object acceptance for separate discussion.

blesh:
> Does the settings parameter idea incur additional performance costs?

Polymorphic function calls, extra object allocation, and property access on that option. So, there will be performance impacts. It's debatable on how large those will be, and it's one that will change with scale and complexity of the calls.

blesh:
Do we really need to be supporting `thisArg` these days? If this is targeting IE9+, then `bind` should be available everywhere.

kwonoj:
> Do we really need to be supporting thisArg these days? 

: This PR tried to align existing signature (as well as RxJS4) to match it's behavior. So opposite could also be answer drop support of `thisArg`, depends on decision to made.

blesh:
For right now, I'm going to pass on this PR and let it simmer for a while. The signature of `first` should match the signature of `last`. 5-6 years ago, when RxJS was born, having `thisArg` everywhere made total sense, because there was no `.bind` Now if people want it, all they really have to do is call `bind` and they're good. If they don't simply close-over the variable they care about.

kwonoj:
Sounds fair, can be revisited anytime when necessary.

kwonoj:
Mark this with clean, red label :)

staltz:
Yeah, I agree with Ben about `thisArg`, we could drop it everywhere possible. IMO 3-param function is Ok (hence no polymorphic call), 4 parameters starts to be not Ok.

blesh:
@kwonoj I think we can close this PR, as we're going to remove `thisArg` from every operator that isn't made to mirror a native function (so everything but `filter`, `map`, `reduce` and `forEach`)

kwonoj:
OK, I'll close this PR. Thanks!

Issue #701, opened 2015-11-11
Idea: TestScheduler to support marble diagrams for promises


staltz:
It's annoying to have to fallback to timers and timeouts when testing code that needs to use Promises. Would be good to be able to write

```js
var e1 = promise('---|', 'theResolveValue'); // or ---#
```
just like we can write
```js
var e1 = cold('---(a|)', {a: 'someValue'});
```

As a by-product, test execution might get a bit faster.

blesh:
I'm for it. But would this mean writing our own thennable pseudo-promise?  Or is there a promise  impl we can use that supports schedulers we can shoehorn?

staltz:
I was hoping a regular virtual time `action.execute()` to resolve the promise would do the trick. Do you think it's possible?

blesh:
That's not a bad idea. I like it.

staltz:
See #745 for the reasons why I'm closing this issue.

PR #745, opened 2015-11-18
feat(TestScheduler): add support for testing promises synchronously

staltz:
**DO NOT MERGE.**

FAILED attempt at adding support for testing promises synchronously with a VirtualTimeScheduler. This
commit is just to illustrate the attempt and highlight how exactly does it not work.

For issue #701.

- - -

After spending some time punching Promises in the face trying to make them behave synchronously like Observables, I gave up on this endeavor because I'm fairly convinced it's not possible. Here is one simple example:

```js
describe('debounce', function () {
  it('should accept a Promise as the duration selector ', function () {
    var e1 =    hot('--a--bc--d----|');
    var ptemplate =   '--|          ';
    var expected =  '----a---c--d--|';

    var result = e1.debounce(function durationSelector() {
      return promise(ptemplate, 42);
    });

  });
```

And when executed:
```
Failures:
1) Observable.prototype.debounce() should accept a Promise as the duration selector 
  Message:
    Expected 
    {"frame":140,"notification":{"kind":"N","value":"d","hasValue":true}}
    {"frame":140,"notification":{"kind":"C","hasValue":false}}
    
    to deep equal 
    {"frame":40,"notification":{"kind":"N","value":"a","hasValue":true}}
    {"frame":80,"notification":{"kind":"N","value":"c","hasValue":true}}
    {"frame":110,"notification":{"kind":"N","value":"d","hasValue":true}}
    {"frame":140,"notification":{"kind":"C","hasValue":false}}
 ```

In other words this means: **because Promises call `then()` at the next tick**, which happens **after** the rxTestScheduler has passed through all those frames, the promise as the duration selector for debounce will debounce the events for the next tick, **and not** for frames 40, 80, 110, as we specified. There is not much we can do since this is due to debounce's implementation, not related to the TestPromise helper I made.

I quit the attempt to support marble diagrams for Promises because it seems the ROI won't be positive. This PR is just to report my activity. I'll close this PR and issue #701 too.

CC @blesh @kwonoj

### DIFF
--- a/spec/helpers/marble-testing.js
+++ b/spec/helpers/marble-testing.js
@@ -12,6 +12,13 @@ function cold() {
   return global.rxTestScheduler.createColdObservable.apply(global.rxTestScheduler, arguments);
 }
 
+function promise() {
+  if (!global.rxTestScheduler) {
+    throw 'tried to use promise() in async test';
+  }
+  return global.rxTestScheduler.createPromise.apply(global.rxTestScheduler, arguments);
+}
+
 function expectObservable() {
   if (!global.rxTestScheduler) {
     throw 'tried to use expectObservable() in async test';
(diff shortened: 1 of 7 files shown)


SLaks:
Can this be implemented using `await` instead?

PR #746, opened 2015-11-18
refactor(TestScheduler): replace let with const and add TS types

staltz:
Use TypeScript return type for createColdObservable(), and replace some `let` occurrences with `const`
in TestScheduler.


kwonoj:
:+1:

kwonoj:
Merged with c2551ea, thanks @staltz !

PR #754, opened 2015-11-19

kwonoj:
Originated from https://github.com/ReactiveX/RxJS/pull/729#issuecomment-156780408 , with some of async test observables are not unsubscribed after test case is completed. 

In most cases it'll be harmless but would be better to unsubscribe it explicitly.

@staltz , would like to ask opinions tests for `bufferWhen`, if this revise would be legit or there should be better way to handle it.

### DIFF
--- a/spec/observables/fromEvent-spec.js
+++ b/spec/observables/fromEvent-spec.js
@@ -94,11 +94,12 @@ describe('Observable.fromEvent', function () {
       }
     };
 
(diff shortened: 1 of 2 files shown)


staltz:
Yes this PR makes perfect sense. Good discovery.

kwonoj:
Merged with 505f5b7.

PR #756, opened 2015-11-19
Make signature of .publishReplay() consistent with publishReplay()

flyingmutant:


kwonoj:
Yes, this seems good catch. Appreciate @flyingmutant

kwonoj:
Merged with 3f85538, thanks again @flyingmutant

kwonoj:
I just noticed my mistake of merging without aligning commit message...

kwonoj:
cc @blesh to have visibility, instead of trying to mend master branch by myself possibly can cause another disaster.

Issue #697, opened 2015-11-11
Generating marble diagram PNGs from tests to documentation


staltz:
As we approach the phase of writing docs, I had an idea. I'm proposing a neat addition that might help us in documentation.

`1.` In tests, we use `byDefinition` in `it.byDefinition('should something something...')` to tag a test case to be parsed for marble diagram generation.

E.g.

```js
// delay-spec.js

  it.byDefinition('should delay by specified timeframe', function () {
    var e1 =   hot('-a-b-c-d-e--|');
    var expected = '-----a-b-c-d-e--|';

    expectObservable(e1.delay(30, rxTestScheduler)).toBe(expected);
  });
```

`2.` We make a script to run as `npm run build_marbles` which gathers those `byDefinition` tests and uses https://github.com/aheckmann/gm to generate respective marble diagrams as png, storing them as e.g. `delay.png` (Just by dropping `spec` in `delay-spec.js`) in doc/img folder.

`3.` Something like this comes out

<img width="640" alt="delay" src="https://cloud.githubusercontent.com/assets/90512/11089368/d8c174c4-8871-11e5-8683-59bc19ea62d8.png">

What do you think?
I can make a proof of concept of this, with a small case, if you think it's worth it.

trxcllnt:
I like it.

blesh:
I think it sounds pretty cool. You could probably subclass or patch the TestScheduler to capture the metadata for the diagrams.

staltz:
> You could probably subclass or patch the TestScheduler to capture the metadata for the diagrams.

Precisely my thought. It sounds easy in theory but let see how it unfolds.

kwonoj:
I like this idea too.

staltz:
Turns out this is being very fun to develop.

So far I got a small script that given:
```js
var inputTestMessages = [
  { frame: 20, notification: { kind: 'N', value: 'a' } },
  { frame: 50, notification: { kind: 'N', value: 'b' } },
  { frame: 80, notification: { kind: 'C' } }
];

var outputTestMessages = [
  { frame: 60, notification: { kind: 'N', value: 'a' } },
  { frame: 90, notification: { kind: 'N', value: 'b' } },
  { frame: 120, notification: { kind: 'C' } }
];
```
(The above is what ASCII marble diagrams generate through the TestScheduler)

We get as output:

[...]

PR #757, opened 2015-11-20
chore(tests2png): add first version of PNG diagram generator

staltz:
Build Jasmine helper that detects tests tagged with asDiagram() and runs a painter script to create
a PNG marble diagram out of them.

First attempt at resolving #697.


kwonoj:
Seems travis is not happy with changes - is this PR work in progress?

staltz:
Now passing tests.

kwonoj:
Does images supposed to be checked in, or to be generated on each local machine?

package.json
```diff
@@ -88,6 +89,7 @@
     "fs-extra": "0.24.0",
     "ghooks": "0.3.2",
     "glob": "5.0.14",
+    "gm": "^1.21.1",
```
kwonoj: let's remove `^` to stay with specific version as same as other libraries.

staltz: Will do.

staltz:
> Does images supposed to be checked in, or to be generated on each local machine?

I don't know, that's up to us. What do you suggest? They could be generated at the same time docs are generated. Primary reason why I checked them in is to show how the diagrams look like, for review. :)

staltz:
This is now feature complete. It generates diagrams for the trickiest cases:

**Cold observables that start emitting after a while has passed**

concat

**Observable of Observables as input**

switch

**Observable of Observables as output**

groupby

When this is merged, all we need to do is add `it.asDiagram(label)` test cases to all operator spec files.

kwonoj:
> They could be generated at the same time docs are generated.

: Think this makes sense, doc could include this diagrams so doesn't require to generate it manually per each users.

spec/operators/buffer-spec.js
```diff
@@ -3,6 +3,18 @@ var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
 describe('Observable.prototype.buffer()', function () {
+  it.asDiagram('buffer')('should emit buffers that close and reopen', function () {
```
kwonoj: so does `it.asDiagram()` serves as test case as well as diagram generator?

spec/helpers/tests2png/painter.js
```diff
@@ -0,0 +1,271 @@
...
+  });
+  out = drawOperator(out, operatorLabel, heightSoFar);
+  heightSoFar += OPERATOR_HEIGHT;
+  out = drawObservable(out, maxFrame, heightSoFar, outputStream);
+
+  out.write('./img/' + filename + '.png', function (err) {
+    if (err) {
+      return console.error(arguments);
+    }
+    //console.log(this.outname + ' created  :: ' + arguments[3]);
```
kwonoj: unused code?

spec/operators/buffer-spec.js
staltz: Yes, both. Only `it.asDiagram` specs are used for `tests2png`, but both `it` and `it.asDiagram` are used for `npm test`.

spec/helpers/tests2png/painter.js
staltz: The console log? I can remove it. I used it for debugging, might be useful later.

kwonoj: Yes, console.log part. My 'personal' preference is do not leave unused code as commented in general. Since it's generally harmless output, what about simply enable it as verbose output?

blesh:
WOOO! That is some fancy work, @staltz

staltz:
:)

staltz:
@kwonoj images removed from git, and fixed that `//` commented line.

kwonoj:
I'll try to look in detail around tomorrow, stuck in bed with flu all day.

blesh:
> I'll try to look in detail around tomorrow, stuck in bed with flu all day.

Yuck. Get better.

blesh:
@staltz can you rebase?

staltz:
@blesh Rebased.

blesh:
@staltz I've pulled it down and tried it out. It definitely works. 

I ran into a couple of issues I think we should address in this PR though:

1. I didn't know what "imagemagick", "graphicsmagick" or "ghostscript" were or how to install them. I figured it out eventually (used brew), but we'll need to have that part documented.
2. I still don't really know how it decides which tests to make PNGs from. How is this configured? Can O do anything ad hoc with it? 
3. it wrote to `img/` and that's fine, but it should probably be in `.gitignore`.

Overall a VERY COOL PR!

staltz:
> I still don't really know how it decides which tests to make PNGs from. How is this configured? Can O do anything ad hoc with it? 

I can write down how to configure a test to be converted to a PNG. But the general case of ad hoc generating PNGs is not supported. If it would, it would make sense to make that a separate library.

staltz:
@blesh some few commits added to fix issues you mentioned.

staltz:
Tests failing just because markdown-doctest doesn't recognize `it()` from Jasmine and prefers to blow up.

staltz:
@Widdershin how I configure `.markdown-doctest-setup.js`? Do I need to require Jasmine or can I somehow mock `it()` calls?

Widdershin:
@staltz You could probably do either, but my instinct is just to mock it out.

If you add something like this to `.markdown-doctest-setup.js`:

```js
const it = {
  asDiagram: function (name) {
    return function (label, callback) {
      callback()
    }
  }
}
```

And then add `it` to the exported globals object:

```js
  globals: {
    ...
    it: it
  },
```

That should do the trick. Does that make sense?

staltz:
Hmm, I'll try that! Thanks!

staltz:
@blesh I fixed the markdown-doctest. Maybe now it's good to merge.

blesh:
Awesome stuff @staltz! Merged with 30a6214

🎉🎉🎉

staltz:
Woot!!!

I'll start going wild by making diagrams for all operators :)

PR #761, opened 2015-11-21
chore(build): generate package.json for npm distro

robwormald:
Simple script that runs post build to add package.json with specific overrides for ES6/CJS deployment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     }
   },
   "scripts": {
-    "build_all": "npm run build_es6 && npm run build_amd && npm run build_cjs && npm run build_global",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     }
   },
   "scripts": {
-    "build_all": "npm run build_es6 && npm run build_amd && npm run build_cjs && npm run build_global",
+    "build_all": "npm run build_es6 && npm run build_amd && npm run build_cjs && npm run build_global && npm run generate_packages",
```
kwonoj: what about place `generate_packages` into `prepublish` script?

blesh:
Merged with 185abdf ... thanks @robwormald 

So I think what's left here is to have some sort of script that publishes from both of these directories. We at least need something clean we can do.

PR #768, opened 2015-11-23
test(PromiseObservable): add more tests for fromPromise operator

staltz:
Add more tests to increase test coverage of fromPromise operator (PromiseObservable.create).

### DIFF
--- a/spec/observables/from-promise-spec.js
+++ b/spec/observables/from-promise-spec.js
@@ -1,19 +1,95 @@
 /* globals describe, it, expect */
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
-var Promise = require('promise');
 
 describe('Observable.fromPromise', function () {
-  it('should emit one value from that promise', function (done) {
+  it('should emit one value from a resolved promise', function (done) {
     var promise = Promise.resolve(42);
     Observable.fromPromise(promise)
-      .subscribe(function (x) {
-        expect(x).toBe(42);
-      }, null,
-      function (x) {
-        expect(x).toBe(undefined);
-        done();
-      });
+      .subscribe(
+        function (x) { expect(x).toBe(42); },
+        done.fail,
+        done);
+  });
+
+  it('should raise error from a rejected promise', function (done) {
+    var promise = Promise.reject('bad');
+    Observable.fromPromise(promise)
+      .subscribe(done.fail,
+        function (e) {
+          expect(e).toBe('bad');
+          done();
+        },
+        done.fail);
+  });
+
+  it('should share the underlying promise with multiple subscribers', function (done) {
+    var promise = Promise.resolve(42);
+    var observable = Observable.fromPromise(promise);
+
+    observable
+      .subscribe(
+        function (x) { expect(x).toBe(42); },
+        done.fail,
+        null);
+    setTimeout(function () {
+      observable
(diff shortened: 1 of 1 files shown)


spec/observables/from-promise-spec.js
```diff
@@ -1,21 +1,115 @@
...
-  it('should emit one value from that promise', function (done) {
+  it('should emit one value from a resolved promise', function (done) {
     var promise = Promise.resolve(42);
     Observable.fromPromise(promise)
       .subscribe(function (x) {
         expect(x).toBe(42);
-      }, null,
+      }, done.fail,
       function (x) {
         expect(x).toBe(undefined);
```
kwonoj: Is this required assertion?

spec/observables/from-promise-spec.js
```diff
@@ -1,21 +1,115 @@
...
+  it('should share the underlying promise with multiple subscribers', function (done) {
+    var promise = Promise.resolve(42);
+    var observable = Observable.fromPromise(promise);
+
+    observable
+      .subscribe(function (x) {
+        expect(x).toBe(42);
+      }, done.fail,
+        function (x) {
+          expect(x).toBe(undefined);
```
kwonoj: same as above - is this intention `completed` is being called?

spec/observables/from-promise-spec.js
staltz: Not in my opinion.

kwonoj: agreed. same as line 38 and others, I'd like to suggest to remove `undefined` assertion in `completed` along with correcting function signature for `completed`.

blesh: Those are likely leftovers from when `complete` actually did pass a value per the es-observable spec.

staltz: Did that exist? Ew ugly

staltz:
@kwonoj Updated the PR.

spec/observables/from-promise-spec.js
```diff
@@ -32,4 +108,25 @@ describe('Observable.fromPromise', function () {
...
+      expect(reason).toBe('fail');
+      done();
+    });
+
+    Observable.fromPromise(Promise.reject('bad'))
+      .subscribe(
+        done.fail,
+        function (e) {
+          expect(e).toBe('bad');
+          throw 'fail';
```
kwonoj: Seems this test is for `onError` behavior of `subscribe`, does PromiseObservable has special handling for this case?

staltz: @kwonoj Can you elaborate? I don't know what you're saying.

kwonoj: this test (seems) can be applied any kind of observable, i.e 

`Observable.throw('a').subscribe(..., function(e) { throw 'x' }, ...)` 

which emits error. It looks like testing behavior of `Observer::error()`, got curious if `PromiseObservable` has specific differences to handle those.

staltz: Ah, you're correct. I'll try correcting it.

blesh: `fromPromise` is different in that Promises will _eat_ errors thrown in `then` blocks.

blesh:
This LGTM... merged with 2e9246e .. thanks @staltz

Issue #763, opened 2015-11-22
Catch() and its shared subscriptions


staltz:
When writing tests for catch(), I noticed this behavior:

```js
  it('should catch and allow the cold observable to be repeated with the third ' +
  '(caught) argument', function () {
    var e1 =  cold('--a--b--c--------|       ');
    var subs =    ['^                       !',
                   '        ^               !',
                   '                ^       !'];
    var expected = '--a--b----a--b----a--b--#';

    var retries = 0;
    var result = e1
      .map(function (n) {
        if (n === 'c') {
          throw 'bad';
        }
        return n;
      })
      .catch(function (err, caught) {
        if (retries++ === 2) {
          throw 'done';
        }
        return caught;
      });

    expectObservable(result).toBe(expected, undefined, 'done');
    expectSubscriptions(e1.subscriptions).toBe(subs);
  });
```

Special attention given to this part:
```js
    var e1 =  cold('--a--b--c--------|       ');
    var subs =    ['^                       !',
                   '        ^               !',
                   '                ^       !'];
```

This happens because CatchSubscriber does
```ts
  _error(err) {
    const result = tryCatch(this.selector)(err, this.caught);
    if (result === errorObject) {
      this.destination.error(errorObject.e);
    } else {
      this.add(result.subscribe(this.destination)); // <====
    }
  }
```

As a consequence, if someone uses `source.catch(e => source)` as a flavor of `retry()`, the number of subscriptions will indefinitely grow and stay retained.

What do you think should be done about this behavior?

kwonoj:
At immediate sight, I'd expect each subscription unsubscribed when next subscription starts, like below.

```
var subs =    ['^        !',
                '        ^       !',
                '                ^       !'];
```

blesh:
@staltz ... I believe this was solved with #769, no?

kwonoj:
I believe so (though I'm not @staltz :) ) PR still need to be merged though.

blesh:
okay... reopening until the merge.

kwonoj:
Now really closing issue.

PR #764, opened 2015-11-22
test(catch): add comprehensive marbles tests for catch

staltz:

PR #769, opened 2015-11-24
fix(catch): fix catch to dispose old subscriptions 

staltz:
**DEPENDS ON PR #764 MERGED FIRST.**

Fix catch operator to not have anymore a shared underlying Subscription, and instead reset the
subscription for each new observable replacing the caught error. This fixes a potential memory leak
if catch is used as an infinite retry, because subscriptions would be retained since the beginning,
and would increasing each time a catch is performed.

Resolves issue #763.

### DIFF
--- a/src/operators/catch.ts
+++ b/src/operators/catch.ts
@@ -1,6 +1,7 @@
 import {Operator} from '../Operator';
 import {Subscriber} from '../Subscriber';
 import {Observable} from '../Observable';
+import {Subscription} from '../Subscription';
 import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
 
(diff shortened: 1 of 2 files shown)


src/operators/catch.ts
```diff
@@ -50,7 +50,17 @@ class CatchSubscriber<T> extends Subscriber<T> {
...
+      this.lastSubscription = result.subscribe(this.destination);
     }
   }
+
+  _complete() {
+    this.lastSubscription.unsubscribe();
+    this.destination.complete();
+  }
+
+  _unsubscribe() {
```
kwonoj: Is this function being called?

staltz: According to test coverage, yes it is. `_unsubscribe()` is overriding Subscription._unsubscribe(), which is called inside Subscription.unsubscribe().

kwonoj: Ah, didn't aware it's inherited from `Subscription`. thought it's inherited from `Subscriber` and tried to look `Subscriber::_unsubscribe()` instead.. my bad. :(

kwonoj:
@staltz , would you able to rebase this one? #764 is now merged.

staltz:
Rebased

kwonoj:
Merged with 280f7ed, thanks @staltz

PR #770, opened 2015-11-24
test(retryWhen): add a few missing tests for retryWhen

staltz:
To achieve 100% coverage for retryWhen.


kwonoj:
Merged with 23f21e0, thanks @staltz

PR #771, opened 2015-11-24
test(bufferTime): add subscription tests for bufferTime

staltz:


kwonoj:
Merged with 36d32b0, thanks @staltz

PR #773, opened 2015-11-24
test(bufferCount): add subscription assertions for bufferCount tests

staltz:


kwonoj:
Merged with 9f71dc9, thanks @staltz

PR #774, opened 2015-11-24
test(bufferWhen): add subscription assertions for bufferWhen tests

staltz:


kwonoj:
Merged with a76869e, thanks @staltz

PR #776, opened 2015-11-24
test(concatAll): add concatAll tests to mirror concat tests

staltz:


kwonoj:
Merged with 09c1c45, thanks @staltz

PR #781, opened 2015-11-25
test(dematerialize): add subscription assertions for dematerialize tests

staltz:


kwonoj:
Merged with 316615d, thanks @staltz

PR #785, opened 2015-11-25
test(elementAt): add subscription assertions on elementAt tests

staltz:


kwonoj:
Merged with 07099cd, thanks @staltz

PR #788, opened 2015-11-25
expand: fixes and tests

staltz:
Fix expand() in order to emit each next value only once. It was previously emitting those values
multiple times when a concurrency limit was set.
Add subscription assertions on expand tests. Add concurrency-related tests.

### DIFF
--- a/spec/operators/expand-spec.js
+++ b/spec/operators/expand-spec.js
@@ -1,4 +1,4 @@
-/* globals describe, it, expect, expectObservable, hot, cold */
+/* globals describe, it, expect, expectObservable, expectSubscriptions, hot, cold */
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 var Promise = require('promise');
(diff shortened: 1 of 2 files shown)


kwonoj:
> Fix expand() in order to emit each next value only once. It was previously emitting those values
multiple times

: :+1: I was hitting while I try to add test coverage for this operator, hadn't look in detail at the moment.

kwonoj:
btw, huge PR combos :)

gif_20150606_203702

staltz:
:sunglasses:

spec/operators/expand-spec.js
```diff
@@ -27,14 +30,214 @@ describe('Observable.prototype.expand()', function () {
...
+    var e2shape =  '---(z|)           ';
+    var expected = 'a-ub-vc-xd-ye-(z|)';
+    var concurrencyLimit = 100;
+
+    var result = e1.expand(function (x) {
+      if (x === 16 || x === 160) {
+        return Observable.empty();
+      }
+      return cold(e2shape, { z: x + x });
+    }, concurrencyLimit);
```
kwonoj: how does concurrency limit ignorance works when it's delivered as param of `expand` compare to other cases where it delivers and expect to honor those limit?

staltz: I meant that if the concurrency limit is high enough, then the behavior should be the same as not specifying it at all (which means Infinity). It's important to test because there is an `if` statement that only runs when the parameter is finite. We want to be sure "very big number" works the same as "infinite".

kwonoj: understood. thought `not passed` means explicitly skips concurrency limits.

kwonoj:
In general change looks good to me.

kwonoj:
Merged with 01f86e5, thanks @staltz

PR #791, opened 2015-11-25
test(findIndex): add subscription assertions for findIndex tests

staltz:


kwonoj:
Merged with d84cd61, thanks @staltz

PR #792, opened 2015-11-25
refactor(find): remove unnecessary if statement from FindValueSubscriber

staltz:
Remove code that checks for a valid type of predicate given to find() operators. The check is
unnecessary since the error resulting from an invalid predicate will be caught in tryCatch() util. This also gives us a tiny performance boost.

### DIFF
--- a/src/operators/extended/find-support.ts
+++ b/src/operators/extended/find-support.ts
@@ -44,10 +44,6 @@ export class FindValueSubscriber<T> extends Subscriber<T> {
   _next(value: T): void {
     const predicate = this.predicate;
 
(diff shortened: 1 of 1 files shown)


kwonoj:
Change looks good to me. 

One minor suggestion is if there'd be more clear exception message when `tryCatch` fails by try to `apply` directly, results `{"kind":"E","exception":{"message":"Cannot read property 'apply' of undefined"}`. but this seems can be considered later.

kwonoj:
Merged with 14e2e25, exception messages can be revised later once if necessary. Thanks @staltz

Issue #766, opened 2015-11-23
Expand operator behavior


kwonoj:
Took example snippet from RxJS 4, 

```
var source = Rx.Observable.return(42)
    .expand(function (x) { return Rx.Observable.return(42 + x); })
    .take(5);

var subscription = source.subscribe(console.log, console.log, ...);
```

*RxJS4*
```
Next: 42
Next: 84
Next: 126
Next: 168
Next: 210
Completed
```

*RxJS Next*
```
RangeError: Maximum call stack size exceeded
    at ExpandSubscriber._next 
```

blesh:
This seems to be more related to `take`, as when I remove `take` it works fine.

cc/ @trxcllnt

kwonoj:
Took a look this bit more, this is behavior of `expand`. Internally it recursively calling `_next()` to expand, having 2 problems

- for infinite expand with subscription, recursive call causes stack overflow
- in case of scalar value result via projection at 
```
 if (result._isScalar) {
          this._next(result.value);
 }
```

it infinitely continues while there isn't terminating condition.

blesh:
> for infinite expand with subscription, recursive call causes stack overflow

`expand` is apparently missing the `scheduler` argument. We should still default to "no scheduler", which means immediate execution (and potential stack overflow).

kwonoj:
> expand is apparently missing the scheduler argument

: compare to RxJS4, yes it does. I'll try changes for this.

>  (and potential stack overflow)

: I think this is similar case to https://github.com/ReactiveX/RxJS/issues/651, eventually might need some  mechanism to avoid stack overflows.

blesh:
> eventually might need some mechanism to avoid stack overflows.

That mechanism is allowing users to provide a scheduler.

The problem is that any mechanism to avoid stack overflows also: 1. is less performant, and 2. makes debugging with call stacks gnarly. ... those are two of our top goals.

kwonoj:
I'm closing this issue since `scheduler` parameter is being tracked in #841 .

Issue #651, opened 2015-11-04
Repeat operator cannot repeat indefinitely


kwonoj:
Code snippet 
` Rx.Observable.of(42).repeat(-1).subscribe(console.log);`

eventually will stop unexpectedly based on runtime due to maximum call stack exceeded. This is happening due to current resubscription occurs recursively. In previous implementation I used [scheduler](https://github.com/ReactiveX/RxJS/pull/562/files#diff-0864900d001ae5c1b7070f19e5e13a2eL28) to disconnect recursive chain at the moment lacks of better idea.

I'm assuming `retry` will behaves same since it's using same mechanics.

kwonoj:
As discussed in https://github.com/ReactiveX/RxJS/pull/547, I think this is bit extreme cases though.

blesh:
this is true for `retry` and `repeat`... when you're applying those to completely synchronous observables, you're going to get that because the default scheduler is recursive. The only other thing we could do would be to force the resubscription to happen on the trampoline scheduler in these operators. I think I discussed this with @trxcllnt and he convinced me we didn't want to do that.

Perhaps he can remind me what the reasoning is?

blesh:
As of right now, I'm inclined to say this isn't an issue. If someone really wants to repeat a completely synchronous observable indefinitely, they should really use a `subscribeOn` directly before the repeat call.

What do you think @kwonoj? Is that answer satisfactory?

kwonoj:
I agree. As already commented, I believe this is extreme cases that usually would not attempted to be used. Issue was created to confirm behavior to check difference between RxJS4.

kwonoj:
Maybe closing issue for now?

mattpodwysocki:
@kwonoj @blesh no, it's a bug, keep it open.  It's not an extreme case at all, there's a bug with their scheduler as RxJS v4 does not have this issue

blesh:
@mattpodwysocki okay.

kwonoj:
Do we consider this issue to be handled as Scheduler side behavior? (maybe changing subj.?)

blesh:
It's a result of default recursive scheduling on synchronous observable types. Where RxJS 4 used trampoline scheduling by default, we're using recursive scheduling for performance reasons. The side effect to this is that if you retry or repeat more than 255 times, you blow the stack.

Perhaps we could implement some sort of progressive scheduling in these scenarios. But I'm just spitballing... I'd like @trxcllnt's opinion on the matter.

[...]

PR #793, opened 2015-11-25
fix(expand): terminate recursive call when destination completes

kwonoj:
relates to #766 

Partially resolves expand operator behavior by adding recursive call terminating condition. Still inifnite expand causes stack overflow though.

Feeling this might not be best way to solve issue, any suggestions would be gladly welcome.

### DIFF
--- a/src/operators/expand-support.ts
+++ b/src/operators/expand-support.ts
@@ -33,12 +33,19 @@ export class ExpandSubscriber<T, R> extends OuterSubscriber<T, R> {
   }
 
   _next(value: any): void {
+    const destination = this.destination;
+
+    if (destination.isUnsubscribed) {
+      this._complete();
+      return;
+    }
+
     const index = this.index++;
     if (this.active < this.concurrent) {
-      this.destination.next(value);
+      destination.next(value);
       let result = tryCatch(this.project)(value, index);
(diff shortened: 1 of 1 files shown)


staltz:
@kwonoj this PR might make sense to merge together with #788. With both PRs, the infinite expand of ScalarObservables causes stack overflow eventually, but at a very large number, `579096`. 

If I add a small delay (hence converting the synchronous ScalarObservable to a normal async Observable), like ` .expand(function (x) { return Rx.Observable.of(42 + x).delay(1); })`, then the infinite expand can easily pass larger numbers like `1000000` and keep on going indefinitely.

So, I don't think very long living synchronous expands is something we need to support. This PR and PR #788 should be enough.

kwonoj:
I agree this PR need to be merged with #788 to correct its behaviors.

Supporting infinite `expand` is somewhat debatable, I'm thinking this is similar cases to https://github.com/ReactiveX/RxJS/issues/651 . In general, with current design operators have weakness to repeat its behavior with sufficiently large number (depends on system, runtime). It would be better to have mechanism instead of recursive calling.

staltz:
Mmm, yes, agreed.

kwonoj:
Merged with 3b8cf94.

PR #795, opened 2015-11-26
refactor(Notification): update type definition for Notification

kwonoj:
Minor type definition updates for `notification` to allow `error` parameter to be optional.


kwonoj:
Merged with fc6c3c4.

PR #798, opened 2015-11-26
test(isEmpty): add subscription assertions on isEmpty tests

staltz:


kwonoj:
Merged with 3dff5b6, thanks @staltz

PR #800, opened 2015-11-26
test(map): add subscription assertions on map tests

staltz:


spec/operators/map-spec.js
```diff
@@ -88,10 +116,12 @@ describe('Observable.prototype.map()', function () {
     });
 
     expectObservable(r).toBe(expected, values);
+    expectSubscriptions(a.subscriptions).toBe(asubs);
   });
 
   it('should map with index until completed', function () {
     var a = hot('-5-^-4--3---2----1--|--8--|--#');
```
kwonoj: (I know this is out of scope for PR) do we allow this kind of marble diagram? maybe this need to be refactored in further too?

staltz: AFAIK RxJS 4 tests do this kind of thing too. I think it's testing whether the Subject behaves well even though we are pushing it beyond its limits, asking it to do what it cannot do.

kwonoj: Yes, some of RxJS 4 test does and while migrated into RxJS 5 some of them were removed those (emit-after-completion). Got curious whether we'll apply same consistency here..

staltz: Yeah IMO it's enough if the Subject spec does this test for emission after completion and operator specs like this file don't.

kwonoj:
Merged with 417dc0a, thanks @staltz

PR #801, opened 2015-11-26
test(mapTo): add subscription assertions on mapTo tests

staltz:


kwonoj:
Merged with bf3c13c, thanks @staltz

PR #803, opened 2015-11-26
test(max): add subscription assertions on max() tests

staltz:


kwonoj:
Merged with 9b4d86c, thanks @staltz

PR #804, opened 2015-11-26
test(mergeMap): improve subscription assertions on tests

staltz:


kwonoj:
Just minor curiosity, was commit message supposed to be includes operator name?
>improve subscription assertions on **mergeMap** tests

staltz:
In this case I think I shortened the commit message by removing mergeMap. I'm unsure if the commit message after `type(scope):` should be a complete message or if it should assume the `scope` is known.

kwonoj:
>  should be a complete message or if it should assume the scope is known

: I think this is fine. Just wanted confirm if this is known and intended :)

kwonoj:
Merged with e11f5d8, thanks @staltz

PR #806, opened 2015-11-26
test(min): add subscription assertions on min() tests

staltz:


kwonoj:
Merged with a6720f2, thanks @staltz

PR #807, opened 2015-11-26
test(partition): add subscription assertions on partition tests

staltz:


kwonoj:
Merged with 9922809, thanks @staltz

PR #808, opened 2015-11-26
test(reduce): add subscription assertions on reduce tests

staltz:


kwonoj:
Merged with b7a7598, thanks @staltz .

PR #813, opened 2015-11-27
test(scan): add subscription assertions on scan tests

staltz:


kwonoj:
Merged with 3f51033, thanks @staltz

PR #814, opened 2015-11-27
test(sampleTime): add subscription assertions on sampleTime tests

staltz:


kwonoj:
Merged with a761685, thanks @staltz

PR #815, opened 2015-11-27
test(sample): add early unsubscription test for sample

staltz:


kwonoj:
Merged with 640a2d4, thanks @staltz

PR #817, opened 2015-11-27
test(skipUntil): add early unsubscription test for skipUntil

staltz:


kwonoj:
Merged with 11e61b8, thanks @staltz

PR #818, opened 2015-11-27
test(skipWhile): improve never/empty/throw tests on skipWhile

staltz:


kwonoj:
Merged with fee03d8, thanks @staltz

PR #819, opened 2015-11-27
test(startWith): add subscription assertions on startWith tests

staltz:


kwonoj:
Merged with bfdb934, thanks @staltz

PR #821, opened 2015-11-27
test(buffer): expand test coverage for buffer

kwonoj:
Add additional coverage for `buffer` operator.


kwonoj:
Merged with f6ef25a.

PR #822, opened 2015-11-27
test(switchFirst): improve switchFirst subscription assertions

staltz:


kwonoj:
Merged with cfb82b7, thanks @staltz

PR #823, opened 2015-11-28
refactor(mergeMapTo): remove explicit errorobject checking

kwonoj:
This PR applies minor refactoring to `mergeMapTo` operator checks explicitly given parameter is `errorObject`.

current implementation of `mergeMapTo` check if given observable parameter is errorObject everytime, which can be represented as test cases 

```
var x = errorObject;

var e1 =   hot('a---b-----------c-------d-------|');
var expected = '#';
var source = e1.mergeMapTo(x);

expectObservable(source).toBe(expected);
```

which doesn't have usage in general since `mergeMapTo` requires observable as parameter.

Maybe there's possible usecases I could not think of - it'd be appreciate for opinions. If passing errorObject is valid usecase, will convert this PR to add above test case instead.

### DIFF
--- a/src/operators/mergeMapTo-support.ts
+++ b/src/operators/mergeMapTo-support.ts
@@ -38,12 +38,9 @@ export class MergeMapToSubscriber<T, R, R2> extends OuterSubscriber<T, R> {
       const index = this.index++;
       const ish = this.ish;
       const destination = this.destination;
-      if (ish === errorObject) {
-        destination.error(ish.e);
-      } else {
-        this.active++;
-        this._innerSub(ish, destination, resultSelector, value, index);
-      }
+
+      this.active++;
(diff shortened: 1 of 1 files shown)


staltz:
Strange. My intuition is to remove it as this PR suggests.

kwonoj:
As same as @staltz I'm thinking this change's legit, but better to have confirmation if there's any historical reason. Marking this as blocked for now, /cc @blesh for opinion. Will merge once this is approved & good to go.

blesh:
LGTM

I'd bet that was dirty leftovers from a copy-pasta deal when `mergeMapTo` was created from `mergeMap`.

kwonoj:
Thanks for clarification, will merge this. :)

kwonoj:
Merged with 08af3e8.

PR #826, opened 2015-11-29
Proposal : migrate unit test to typescript 

kwonoj:
**Not to be merged since it's for proposal**

Related with https://github.com/ReactiveX/RxJS/issues/661, https://github.com/ReactiveX/RxJS/pull/667

One of other possibilities of enhancing type definition validation is using unit test instead of having additional test for definition only, as same as other (such as angular 2) does. This PR proposes basic setups of migrating existing test cases into typescript based, to gather opinions to proceed or not.

This PR does
- reuses most of existing test fixture setup (marble-testing helper function, test-helper) as is, with small modifications. Additional refactoring for typescript specific can be considered further once if test case is actually migrated.
- picks one existing case (`count-spec`) convert it into typescript based. Major difference is importing helper function (`hot`, `cold`, ...) explicitly instead of using global.
- for rxjs build, test case still uses `cjs` output instead of directly referencing typescript source to reduce build time for test. 

Travis currently fails by `markdown-doctest` is not updated with this changes.


blesh:
I'm on my phone, so I can't fully review this, but it's probably worth building all jasmine test helpers into their own module. At some point we'll want to publish that for others to use anyhow.

kwonoj:
> it's probably worth building all jasmine test helpers into their own module.

: Yes, good point.

I'll leave this specific PR without updating it to discuss go / no-go first, once it's decided to proceed will update PR accordingly.

staltz:
> Travis currently fails by markdown-doctest is not updated with this changes. 

By the way, I think we should make `markdown-doctest` errors be just warnings. Either that, or we need to find a way of explicitly ignoring some parts. @Widdershin ?

kwonoj:
> I think we should make markdown-doctest errors be just warnings. 

: Isn't purpose of `markdown-doctest` is let it fail if it 'fails'? ;)

Widdershin:
@staltz If a specific example shouldn't be tested, you can prefix it with `<!-- skip-example -->`.

Widdershin:
I'm also happy to add support for Typescript to `markdown-doctest`.

staltz:
> If a specific example shouldn't be tested, you can prefix it with `<!-- skip-example -->`.

Ok, that's good, I might use it. My case wasn't typescript code, it was Jasmine code, it didn't recognize `it('should be')` type of specs.

Widdershin:
@staltz yell out if you want any help integrating `markdown-doctest` into a project, I've been meaning to set it up for some of the cycle repos.

kwonoj:
> I'm also happy to add support for Typescript to markdown-doctest

: Appreciate for it, let's discuss in further depends on decisions made :)

staltz:
@Widdershin let's discuss in #757, so this PR discussion stays on topic.

blesh:
@kwonoj Discussion is dead here. While I'd like to move our tests to TypeScript for a variety of reasons, I don't see this particular PR going anywhere. Closing for now, feel free to reopen if you disagree, I'm just cleaning up.

kwonoj:
I think it's ok to close, issue https://github.com/ReactiveX/RxJS/issues/661 can be used to get conclusion where to start. I own responsibility to wrap up those discussion to which way to go.

Issue #825, opened 2015-11-29
Using Subjects as Observers


staltz:
I think this is a bug. Currently RxJS 5 Subjects cannot be used as Observers in `source.subscribe(subject)`, even though Subject implements the Observer interface.

```js
// Rx 5
observable.subscribe(
  subject.next.bind(subject),
  subject.error.bind(subject),
  subject.complete.bind(subject)
)

//Rx 4
observable.subscribe(subject.asObserver())
```

This should be simple to fix, but should be accompanied by tests too.

blesh:
This is totally a bug. They should work as observers. It's half the point.

blesh:
Also, you shouldn't have to use `asObserver`.

staltz:
Agreed.

staltz:
@blesh Apparently I'm not able to reproduce the bug, it seems everything is ok. See PR #828. I'm awaiting for a bug report from @TylorS who originally reported this bug to me.

blesh:
@staltz ... I did fix a bug semi-related to this a while ago: #724 

That might have been the culprit... I don't know.

ntilwalli:
@staltz Isn't this similar to issue https://github.com/ReactiveX/RxJS/issues/748?

blesh:
@staltz @ntilwalli I just tried every Subject type from the latest build and I couldn't reproduce this issue.

If we can't reproduce it, can we close this?

staltz:
Yeah let's close, Tylor couldn't reproduce it either.

ntilwalli:
This actually did not fix the issue from #748.  I've created a pull request with a fix for that issue.  https://github.com/ReactiveX/RxJS/pull/863

blesh:
@ntilwalli is right: There is a *minor* issue that Subjects are being unnecessarily wrapped in Subscribers. This is less of a "bug" and more of a "refactor".

PR #828, opened 2015-11-30
test(Subject): add tests for subjects used as Observer

staltz:
Add tests for Subject, BehaviorSubject and ReplaySubject to check whether they can be used as plain
observers when given to Observable.subscribe().

Related to issue #825.


kwonoj:
Change looks good to me.

kwonoj:
Merged with 3c72bca, thanks @staltz

PR #829, opened 2015-11-30
test(switchMap): add subscription assertions on switchMap tests

staltz:


kwonoj:
Merged with 12eb98e, thanks @staltz

PR #832, opened 2015-11-30
test(take): add subscription assertions on take() tests

staltz:


kwonoj:
Merged with affaf23, thanks @staltz

PR #834, opened 2015-11-30
test(takeWhile): improve subscription assertions

staltz:
Mainly related to empty/never/throw tests.


kwonoj:
Merged with e8d4546, thanks @staltz

PR #835, opened 2015-11-30
test(throttleTime): improve subscription assertions

staltz:
Add a subscription assertion for a test on throttleTime that was missing it.


kwonoj:
Merged with 9873401, thanks @staltz

PR #836, opened 2015-11-30
test(timeInterval): add subscription assertions

staltz:
Add subscription assertions on timeInterval tests.


kwonoj:
Merged with a643efa, thanks @staltz

PR #837, opened 2015-11-30
test(timeout): add subscription assertions

staltz:
Add subscription assertions on timeout() tests.


kwonoj:
Merged with f157258, thanks @staltz

PR #840, opened 2015-11-30
timeoutWith: fixes and tests

staltz:
Because of the underlying shared Subscriber, timeoutWith was not unsubscribing from the source observable once the timeout happened. [To align with RxJS 4](https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/timeout.js#L151-L187), this PR fixes that.

### DIFF
--- a/spec/operators/timeoutWith-spec.js
+++ b/spec/operators/timeoutWith-spec.js
@@ -1,119 +1,260 @@
-/* globals describe, it, expect, expectObservable, hot, cold, rxTestScheduler */
+/* globals describe, it, expect, expectObservable, expectSubscriptions, hot, cold, rxTestScheduler */
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
 describe('Observable.prototype.timeoutWith()', function () {
   it('should timeout after a specified period then subscribe to the passed observable', function () {
-    var e1 = Observable.never();
-    var e2 =  cold('--x--y--z--|');
+    var e1 =  cold('-');
+    var e1subs =   '^    !           ';
+    var e2 =       cold('--x--y--z--|');
+    var e2subs =   '     ^          !';
     var expected = '-------x--y--z--|';
 
-    expectObservable(e1.timeoutWith(50, e2, rxTestScheduler)).toBe(expected);
+    var result = e1.timeoutWith(50, e2, rxTestScheduler);
+
+    expectObservable(result).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    expectSubscriptions(e2.subscriptions).toBe(e2subs);
   });
 
   it('should timeout at a specified date then subscribe to the passed observable', function () {
-    var e1 = Observable.never();
-    var e2 = cold('--x--y--z--|');
+    var e1 =  cold('-');
+    var e1subs =   '^         !           ';
+    var e2 = cold(           '--x--y--z--|');
+    var e2subs =   '          ^          !';
     var expected = '------------x--y--z--|';
 
-    expectObservable(e1.timeoutWith(new Date(rxTestScheduler.now() + 100), e2, rxTestScheduler)).toBe(expected);
+    var result = e1.timeoutWith(new Date(rxTestScheduler.now() + 100), e2, rxTestScheduler);
+
+    expectObservable(result).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    expectSubscriptions(e2.subscriptions).toBe(e2subs);
   });
 
-  it('should timeout after a specified period between emit then subscribe to the passed observable when source emits', function () {
+  it('should timeout after a specified period between emit then subscribe ' +
+  'to the passed observable when source emits', function () {
     var e1 =     hot('---a---b------c---|');
-    var e2 =    cold('-x-y-|');
-    var expected =   '---a---b----x-y-|';
+    var e1subs =     '^          !       ';
+    var e2 = cold(              '-x-y-|  ');
+    var e2subs =     '           ^    !  ';
+    var expected =   '---a---b----x-y-|  ';
 
-    expectObservable(e1.timeoutWith(40, e2, rxTestScheduler)).toBe(expected);
+    var result = e1.timeoutWith(40, e2, rxTestScheduler);
+
+    expectObservable(result).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    expectSubscriptions(e2.subscriptions).toBe(e2subs);
   });
 
-  it('should timeout after a specified period then subscribe to the passed observable when source is empty', function () {
-    var e1 =   hot('-------------|');
-    var e2 =  cold('----x----|');
+  it('should allow unsubscribing explicitly and early', function () {
+    var e1 =     hot('---a---b-----c----|');
+    var e1subs =     '^          !       ';
+    var e2 = cold(              '-x---y| ');
+    var e2subs =     '           ^  !    ';
+    var expected =   '---a---b----x--    ';
+    var unsub =      '              !    ';
+
+    var result = e1.timeoutWith(40, e2, rxTestScheduler);
+
+    expectObservable(result, unsub).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    expectSubscriptions(e2.subscriptions).toBe(e2subs);
+  });
+
+  it('should not break unsubscription chain when unsubscribed explicitly', function () {
+    var e1 =     hot('---a---b-----c----|');
+    var e1subs =     '^          !       ';
+    var e2 = cold(              '-x---y| ');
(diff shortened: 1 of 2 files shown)


src/operators/timeoutWith.ts
```diff
@@ -43,12 +43,12 @@ class TimeoutWithSubscriber<T, R> extends OuterSubscriber<T, R> {
     return this._hasCompleted;
   }
 
-  constructor(destination: Subscriber<T>,
+  constructor(public destination: Subscriber<T>,
               private absoluteTimeout: boolean,
               private waitFor: number,
               private withObservable: Observable<any>,
               private scheduler: Scheduler) {
-    super(destination);
```
blesh: I'm unsure, but I don't think we want to do this here, because it may break the unsubscription chain for the "outer" observable. I think what we want to do is add an "InnerSubscriber" that _does_ do this, though.

staltz: Ok, I'll look into that.
However, we use this same approach in many other operators. Either we add tests for keeping the subscription chain that you mentioned, or we change all those other operators too.

blesh: I know I've seen this pattern, but I'm pretty sure it's relegated to "inner subscribers". If not, we need to have a look at those other operators.

staltz: Gut feeling says there's a good amount of those that don't relegate.

blesh:
@staltz ... where did we leave this? Are you changing this, or do you think it's good as is? If the latter, we're going to need to verify the unsubscription chain isn't broken by this operator.

staltz:
I'm going to try changing it first. But anyway can you clarify how exactly does "broken unsubscription chain" happen?

blesh:
@staltz ... my concern is that if you have something like `e1.mergeMapTo(e2).timeoutWith(e3).mergeMapTo(e4)`, and you subscribe to it, then call `unsubscribe` on the subscription, the teardowns for `e1` and `e2` won't be called.

I'm not *sure* this will happen if you don't pass the destination to `super`, but I feel like it could.

staltz:
> I'm not sure this will happen if you don't pass the destination to super, but I feel like it could.

You're right. It does happen. Is it ok if I add precisely that test case `mergeMapTo > timeoutWith > mergeMapTo` to timeoutWith-spec.js? I also think we need to add this same test `mergeMapTo > whatever > mergeMapTo` for every `whatever`.

staltz:
> I know I've seen this pattern, but I'm pretty sure it's relegated to "inner subscribers". 

So far, there is no InnerSubscriber which handles unsubscription this way. Actually all of the InnerSubscribers are simple vanilla InnerSubscribers, all of the variation happens with OuterSubscribers. I mean, zero subclasses of InnerSubscriber, but plenty of subclasses of OuterSubscriber.

I can solve this issue by allowing the shared subscription to persist over time. Let me explain.

```diff
  it('should timeout after a specified period then subscribe to the passed observable', function () {
    var e1 =  cold('-');
-   var e1subs =   '^    !           ';
+   var e1subs =   '^               !';
    var e2 =       cold('--x--y--z--|');
    var e2subs =   '     ^          !';
    var expected = '-------x--y--z--|';

    var result = e1.timeoutWith(50, e2, rxTestScheduler);

    expectObservable(result).toBe(expected);
    expectSubscriptions(e1.subscriptions).toBe(e1subs);
    expectSubscriptions(e2.subscriptions).toBe(e2subs);
  });
```

In this case, it's harmless. But what about other operators such as repeat?

```diff
  it('should resubscribe multiple times', function () {
    var e1 =   cold('--a--b--|                        ');
-   var subs =     ['^       !                        ',
-                   '        ^       !                ',
-                   '                ^       !        ',
-                   '                        ^       !'];
+   var subs =     ['^                               !',
+                   '        ^                       !',
+                   '                ^               !',
+                   '                        ^       !'];
    var expected =  '--a--b----a--b----a--b----a--b--|';

    expectObservable(e1.repeat(2).repeat(2)).toBe(expected);
    expectSubscriptions(e1.subscriptions).toBe(subs);
  });
```

That's not good, we'll start leaking memory the more the observable repeats.

I'm pointing this out as an overall issue, not related only to this PR and timeoutWith. The shared Subscription is retained in all operators, because of this:

```typescript
class Subscriber<T> ... {
  // ...
  unsubscribe(): void {
    if (this._isUnsubscribed) {
      return;
    } else if (this._subscription) {
      this._isUnsubscribed = true;
      // NO ACTUAL UNSUBSCRIPTION HERE  <-----------
    } else {
      super.unsubscribe();
    }
  }
  // ...
}
```

src/operators/timeoutWith.ts
```diff
@@ -69,31 +68,27 @@ class TimeoutWithSubscriber<T, R> extends OuterSubscriber<T, R> {
   }
 
   _next(value: T) {
-    if (!this.timedOut) {
```
staltz: Noticed I removed `timedOut`. It's not necessary because the check for `isUnsubscribed` that happens in the superclass `Subscriber` already guards against these cases.

spec/operators/timeoutWith-spec.js
```diff
@@ -1,119 +1,260 @@
...
+
+    var result = e1
+      .mergeMap(function (x) { return Observable.of(x); })
+      .timeoutWith(40, e2, rxTestScheduler)
+      .mergeMap(function (x) { return Observable.of(x); });
+
+    expectObservable(result, unsub).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    expectSubscriptions(e2.subscriptions).toBe(e2subs);
+  });
```
staltz: @blesh notice this test above.

staltz:
PR updated. But I'll take the whole "shared subscription" problem to another issue.

trxcllnt:
> @staltz That's not good, we'll start leaking memory the more the observable repeats.

Looking at [`repeat`](https://github.com/ReactiveX/RxJS/blob/master/src/operator/repeat.ts#L49), I don't think that's true.

Every `Subscriber` should pass the destination to the super constructor. There might've been an operator early on that we didn't do this with, but if it's still around, we should probably rethink it. If a Subscriber needs to create inner subscriptions, it should manage them like mergeAll does, e.g. [add](https://github.com/ReactiveX/RxJS/blob/master/src/operator/mergeAll-support.ts#L32) and [remove](https://github.com/ReactiveX/RxJS/blob/master/src/operator/mergeAll-support.ts#L48) them as needed.

staltz:
> Looking at repeat, I don't think that's true.

It's not true, I added that unsubscribe and used `super(null)` in order to make it not leak memory the more the observable repeats. I just meant that if we use `super(destination)` naively, we retain previous shared subscriptions and then we start leaking memory. But...

> If a Subscriber needs to create inner subscriptions, it should manage them like mergeAll does, e.g. add and remove them as needed.

This might be the solution. I'll try that approach, thanks.

staltz:
And again, I completely updated this PR to **not** use `super(destination)` but at the same time support all the use cases we need. @blesh @trxcllnt I'm interested in hearing if you think there is any other use case we should support, **or** a suggestion on how to solve the uses cases when using `super(destination)`.

blesh:
@staltz did you have a chance to make the changes necessary to this commit? I don't see where it's maintaining the subscription chain or the tests around maintaining that chain. But I'm reviewing a lot today so perhaps I'm just burnt out... lol

staltz:
Yep @blesh this PR is all good.

spec/operators/timeoutWith-spec.js
```diff
@@ -1,119 +1,260 @@
...
+    var unsub =      '              !    ';
+
+    var result = e1.timeoutWith(40, e2, rxTestScheduler);
+
+    expectObservable(result, unsub).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    expectSubscriptions(e2.subscriptions).toBe(e2subs);
+  });
+
+  it('should not break unsubscription chain when unsubscribed explicitly', function () {
```
staltz: @blesh here is the test about maintaining the chain

src/operator/timeoutWith.ts
```diff
@@ -43,12 +42,13 @@ class TimeoutWithSubscriber<T, R> extends OuterSubscriber<T, R> {
...
 
-  constructor(destination: Subscriber<T>,
+  constructor(public destination: Subscriber<T>,
               private absoluteTimeout: boolean,
               private waitFor: number,
               private withObservable: Observable<any>,
               private scheduler: Scheduler) {
-    super(destination);
+    super(null);
+    destination.add(this);
```
staltz: @blesh this line 51 makes sure we maintain the Subscriber chain.

blesh: I see, thanks @staltz

blesh: Yesterday I pulled it down and ran it and it looked fine. I just didn't see that bit.

blesh:
Merged with 6e63752... thanks @staltz

PR #842, opened 2015-11-30
test(toArray): add subscription assertions

staltz:
Add subscription assertions on toArray tests.


kwonoj:
Merged with a687695, thanks @staltz

PR #843, opened 2015-11-30
Add prototype-patching operator modules

### LINKED ISSUE #750: Add prototype-patching operator modules
(Recapping some offline discussion between @blesh, @IgorMinar, @robwormald and myself)

In order to allow exporting a minimal `Observable` and to increase composition of operators, while maintaining nice chaining syntax, we propose exporting each operator as an easily-importable module that will have a side effect of patching the `Observable` prototype. 

```typescript
import {Observable} from '@reactivex/rxjs';
// Observable.prototype.map = undefined
import {map} from '@reactivex/rxjs/map';
// Observable.prototype.map = function (...) {...}
```

With TypeScript 1.8, there will be support for this type of module to re-open and extend types, so we can not have to stub out all types on the `Observable` class like we do now. In anticipation of this change, we should go ahead and add interfaces for all operators, with [optional methods](https://github.com/microsoft/TypeScript/issues/5269#issuecomment-148530995). The `Observable` class will implement all operator interfaces instead of manually stubbing out methods on the class itself.

It's not ideal for an import to have side effects, and to mutate a shared object. In this case, I think the mutation is not likely to cause issues for people, since most folks will only import operators in one place inside an application. And if folks are only importing operators from rxjs, there's no risk of method name collision.

Eventually, when JS has sugar like [function bind](https://github.com/zenparsing/es-function-bind) to make extension-like functions more composable, there should be a breaking change to remove the side effect of the imports.

robwormald:
Step one (broken into two commits - operators and statics) in modularization strategy.

Moves prototype patching into individual modules, so Observable prototype can be patched incrementally. 

Before:
```typescript
import {Observable} from '@reactivex/rxjs'; //ALL THE THINGS
//or
import {Observable} from '@reactivex/rxjs/Observable'; //NONE OF THE THINGS
```

After:
```typescript
import {Observable} from '@reactivex/rxjs/Observable'; //NONE OF THE THINGS
//SOME OF THE THINGS!
import '@reactivex/rxjs/observables/PromiseObservable';
import '@reactivex/rxjs/operators/map';
import '@reactivex/rxjs/operators/filter';
import '@reactivex/rxjs/operators/reduce';
```

Part two will be using TS 1.8's module-reopening functionality to patch the interface, at which point the stub methods will be removed from Observable.ts. Not implemented yet - see https://github.com/Microsoft/TypeScript/issues/5269#issuecomment-148871082

Current side effect (minor) is that type-completion will show methods that have possibly not been imported yet.

Closes #750


robwormald:
cc @IgorMinar @jeffbcross - i couldn't work out optional syntax here, might be the pumpkin pie hangover.

blesh:
> //SOME OF THE THINGS!
> import '@reactivex/rxjs/observables/PromiseObservable';

This confuses me. Is this to say that `fromPromise` doesn't exist until `PromiseObservable` is imported? While I'm generally in favor of this approach for operators. I think for the static methods it's harder to figure out.

blesh:
Perhaps it should be:

```js
import `@reactivex/rxjs/observable/fromPromise`;
```

(NOTE: The `reactivex/rxjs` will shrink to just `rxjs` at some point)

robwormald:
Yes, correct, it doesn't exist until imported.

Definitely like this better, it reads much more nicely :+1: 
```typescript
import `@reactivex/rxjs/observable/fromPromise`;
```

blesh:
I'd go so far as to have it also export the fromPromise function (same for the others).

So:

```js
import {fromPromise} from 'rxjs/observable/fromPromise';

fromPromise(makePromise()).subscribe();
```

or

```js
import {Observable} from 'rxjs/observable';
import 'rxjs/observable/fromPromise';

Observable.fromPromise(makePromise()).subscribe();
```
would both work.

blesh:
... well I guess it could export `PromiseObservable`... I dunno. As long as the name is easier to read.

trxcllnt:
> @blesh:
> I'd go so far as to have it also export the fromPromise function (same for the others). [snip] ... well I guess it could export PromiseObservable... I dunno. As long as the name is easier to read.

Could we export both `fromPromise` and `PromiseObservable` from that module, or would that go against our "only export one thing per module" rule?

blesh:
> Could we export both fromPromise and PromiseObservable from that module, or would that go against our "only export one thing per module" rule?

No such rule, IMO. So I don't see why not.

robwormald:
observables => observable :heavy_check_mark:

blesh:
LGTM.

blesh:
Merged as of 95279cc... VERY nice work, @robwormald, and great idea overall @IgorMinar @jeffbcross et al.

kwonoj:
:+1: yay!

robwormald:
@trxcllnt i like the idea of exporting fromPromise et al, but they're mostly static methods on classes (PromiseObservable.create) - should i just
```js
const fromPromise = PromiseObservable.create;
export {fromPromise}
```
or is there something more clever i'm missing?

PR #849, opened 2015-12-01
refactor(Symbol): introduce unified Symbol module

### LINKED ISSUE #844: Move Symbol shims into single file.
Considering we're shimming `Symbol` and items on it frequently, I think it would be best to move it into a single file.

That would make this:

```js
import {$$iterator} from '../utils/Symbol_iterator';
import {$$observable} from '../utils/Symbol_observable';

const something = {
  [$$iterator]() { }
  [$$observable]() { }
};
```

... into this...

```js
import {Symbol} from '../utils/Symbol';

const something = {
  [Symbol.iterator]() { }
  [Symbol.observable]() { }
};
```

kwonoj:
closes #844 

This PR creates single module for symbols, integrates separated `symbol_xxx` modules. As mentioned in https://github.com/ReactiveX/RxJS/issues/844 typescript supports es6 global symbol which causes name collision, picked `SymbolShim` as initial name. Might be better name for module, creating PR to get some suggestions.


blesh:
@kwonoj would it be possible, while you're looking at this to implement and export a shim factory you can unit test (similar to what I did for `Immediate` and `Map`)?

basically it's just:

```js
import {root} from './root';

export function symbolPolyfill(root) {
   // do stuff to create Symbol and patch root
  return Symbol;
}

const Symbol = symbolPolyfill(root);
export Symbol;
```

from there, you can unit test `symbolPolyfill` and get coverage on it for various scenarios: (Symbol exists, but `iterator` doesn't, `Symbol` exists, but `observable` doesn't, `Symbol.observable` does exist, etc)

I realize this is all voluntary, so if you don't have the time, let me know and I'll take a swing at it.

kwonoj:
Gladly, I'll try to mimic those behavior and update PR.

blesh:
Conflicts resolved and merged with 2eb7eec... thanks @kwonoj!

Issue #850, opened 2015-12-01
Add AsyncSubject


mattpodwysocki:
Add AsyncSubject to RxJS v5. Other subjects can be added in the future, but this one is tried and true and used by many community members.

mattpodwysocki:
@blesh @staltz correct me if I'm wrong, but wouldn't it be as simple as this?

```typescript
import {Subject} from '../Subject';
import {Subscriber} from '../Subscriber';
import {Subscription} from '../Subscription';

export class AsyncSubject<T> extends Subject<T> {
  constructor () {
    super();
    this.value = void 0;
    this.hasNext = false;
  }

  _subscribe(subscriber: Subscriber<any>): Subscription<T> {
    const subscription = super._subscribe(subscriber);
    if (!subscription) {
      return;
    } else if (!subscription.isUnsubscribed && this.hasNext) {
      subscriber.next(this.value);
      subscriber.complete();
    }
    return subscription;
  }

  _next(value: T): void {
    this.value = value;
    this.hasNext = true;
  }
}
```

blesh:
LGTM.

I think once this completes, you could set `_isScalar` to `true` on it and get some benefits from operators that fast-track scalar observables.

architruc:
:+1:  for `AsyncSubject`, thanks!

Just one question: what would be the difference with `new ReplaySubject(1)` (if any)? I just learnt about this alternative from angular2 gitter.

staltz:
@architruc very different. AsyncSubject does not emit next values before it completes. It only emits on completion, and after completion for late subscribers.

architruc:
@staltz Okay, if I understand well `ReplaySubject` emits at each value (let's say each `next()` in the source), but `AsyncSubject` on `complete()` in the source, right?

Yes, it is different, seems I misunderstood, although `AsyncSubject` could still help in my usecase. Thanks!

blesh:
This can be closed now, I think.

PR #852, opened 2015-12-01
feat(AsyncSubject): initial implementation

mattpodwysocki:
Initial implementation of the `AsyncSubject` from RxJS v4.x which has the following behavior:

1. Can take as many `next` calls as it wishes, but will not emit a value until `complete` has been called.
2. If no `next` calls have been made, then no subscriptions should be called.

Closes Issue #850

### DIFF
--- a/src/Rx.KitchenSink.ts
+++ b/src/Rx.KitchenSink.ts
@@ -128,6 +128,7 @@ import './operators/zipAll';
 import {Subject} from './Subject';
 import {Subscription} from './Subscription';
 import {Subscriber} from './Subscriber';
+import {AsyncSubject} from './subjects/AsyncSubject';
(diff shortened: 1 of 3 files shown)


blesh:
@mattpodwysocki when you're done, can you reword the commit to be `feat(AsyncSubject): add AsyncSubject`? That way it will show up in the generated changlog as a new feature. :)

mattpodwysocki:
@blesh done

blesh:
Nice! would you mind rebasing and flattening away that merge commit?

mattpodwysocki:
@blesh done

blesh:
rebased, merged conflicts, added some simple tests.

Merged with master as of a365fe3... Thank you, open sorcerer.

PR #853, opened 2015-12-01
feat(BehaviorSubject): add getValue method to access value

### LINKED ISSUE #758: Behaviour getValue functionality missing
Is the `value` property of **BehaviourSubject** supposed to be public? Or is there supposed to be a public `getValue` function like there is in the previous version?
There is currently no way of reading the value without subscribing.

blesh:
- adds `getValue` method to get inner value
- further hides the value as `_value`
- adds `value` property getter
- purposefully excludes `value` property setter
- adds tests around `getValue` method and `value` property

fixes #758

### DIFF
--- a/src/Rx.ts
+++ b/src/Rx.ts
@@ -117,7 +117,7 @@ import {ConnectableObservable} from './observable/ConnectableObservable';
 import {Notification} from './Notification';
 import {EmptyError} from './util/EmptyError';
 import {ArgumentOutOfRangeError} from './util/ArgumentOutOfRangeError';
-import {nextTick} from './scheduler/nextTick';
+import {ObjectUnsubscribedError} from './util/ObjectUnsubscribedError';
 import {immediate} from './scheduler/immediate';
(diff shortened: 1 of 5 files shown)


src/Rx.ts
```diff
@@ -110,7 +110,7 @@ import {ConnectableObservable} from './observable/ConnectableObservable';
 import {Notification} from './Notification';
 import {EmptyError} from './util/EmptyError';
 import {ArgumentOutOfRangeError} from './util/ArgumentOutOfRangeError';
-import {nextTick} from './scheduler/nextTick';
+import {ObjectUnsubscribedError} from './util/ObjectUnsubscribedError';
```
kwonoj: Build failed by removing importing `nextTick` in here still being used in line 121.

kwonoj:
Change looks good to me except missing `nextTick` makes build failed.

blesh:
Merged with ef83066... thank you for the review @kwonoj

PR #856, opened 2015-12-01
refactor(timer): use isScheduler() to check optionals

luisgabriel:


blesh:
merged as of 79b38e4... thanks @luisgabriel

PR #858, opened 2015-12-01
fix(expand): accept scheduler parameter

### LINKED ISSUE #841: expand missing scheduler argument
I pointed out [here](https://github.com/ReactiveX/RxJS/issues/766#issuecomment-160740919), that we are missing a scheduler argument for `expand`. This is necessary for changing the behavior of when `expand` subscribes to observables returned by the selector.

It should default to no/null/undefined scheduler, which means immediate execution.

Related to #766 #788 #793

luisgabriel:
Also moves the handling of the default value for optional parameters to
the expand function instead of the operator's ctor.

Closes #841.

### DIFF
--- a/src/operator/expand-support.ts
+++ b/src/operator/expand-support.ts
@@ -1,5 +1,6 @@
 import {Operator} from '../Operator';
 import {Observable} from '../Observable';
+import {Scheduler} from '../Scheduler';
 import {Subscriber} from '../Subscriber';
 import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
@@ -8,12 +9,13 @@ import {InnerSubscriber} from '../InnerSubscriber';
 import {subscribeToResult} from '../util/subscribeToResult';
 
 export class ExpandOperator<T, R> implements Operator<T, R> {
-  constructor(private project: (value: T, index: number) => Observable<any>,
-              private concurrent: number = Number.POSITIVE_INFINITY) {
+  constructor(private project: (value: T, index: number) => Observable<R>,
+              private concurrent: number,
+              private scheduler: Scheduler) {
   }
 
   call(subscriber: Subscriber<R>): Subscriber<T> {
-    return new ExpandSubscriber(subscriber, this.project, this.concurrent);
+    return new ExpandSubscriber(subscriber, this.project, this.concurrent, this.scheduler);
   }
 }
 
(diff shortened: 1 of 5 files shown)


luisgabriel:
@kwonoj @blesh could you guys take a look if I'm going in the right direction here? ;)

src/operators/expand-support.ts
```diff
@@ -22,12 +26,25 @@ export class ExpandSubscriber<T, R> extends OuterSubscriber<T, R> {
   private active: number = 0;
   private hasCompleted: boolean = false;
   private buffer: any[];
+  private _concurrent: number = Number.POSITIVE_INFINITY;
```
kwonoj: is it necessarily required to have 2 fields, `_concurrent` and `concurrent`?

src/operators/expand-support.ts
```diff
@@ -41,24 +58,31 @@ export class ExpandSubscriber<T, R> extends OuterSubscriber<T, R> {
...
-        }
+        const state = { subscriber: this, result, value, index };
+        this.add(this.scheduler.schedule(disptch, 0, state));
       }
     } else {
       this.buffer.push(value);
     }
   }
 
+  _innerSub(result: any, value: any, index: number): void {
```
kwonoj: is this function inherited? if not, would like to suggest to name like `private innerSub(): void`

src/operators/expand-support.ts
luisgabriel: It is not. I'll fix it.

src/operators/expand-support.ts
luisgabriel: Ok, I can fix this. Does the `_` means something special?

kwonoj: I don't think so in general in RxJS codebases so far I know. Patterns I could see was some of inherited (such as `_next`, etcs) are starts with underbar, along with some private fields. Just thought in typical cases of instance method wouldn't need to start with underbar, especially `_innerSub` is public and being called externally via `dispatch`.

luisgabriel: Yes, I was suspecting that it means private. Ideally `innerSub` should be private. Currently, it isn't because `dispatch` cannot be an instance method (AFAIK).

kwonoj: One possible way is making `dispatch` as static method for class if you'd like to encapsulate `innerSub` as private. You may see those approach in other places like `DelaySubscriber::dispatch`.

luisgabriel: Hmm, nice. Will try this approach. 
I'll also rename `_innerSub()` to `subscribeToProjection()`.

luisgabriel:
Just updated this PR addressing the previous comments.

Still missing some test cases using a custom scheduler. Suggestion are very welcome. ;)

src/operator/expand-support.ts
```diff
@@ -46,19 +53,26 @@ export class ExpandSubscriber<T, R> extends OuterSubscriber<T, R> {
       let result = tryCatch(this.project)(value, index);
       if (result === errorObject) {
         destination.error(result.e);
+      } else if (!this.scheduler) {
+        this.subscribeToProjection(result, value, index);
```
blesh: Looks like we've lost the scalar optimization for this branch of code. `if (result._isScalar) {` ...etc.

luisgabriel: It was moved to `subscribeToProjection()`.

blesh: I see. Thanks.

blesh:
Merged with 79e9084. Thanks @luisgabriel!

PR #861, opened 2015-12-02
refactor(SymbolShim): export SymbolShim using factory

kwonoj:
Following changes of PR https://github.com/ReactiveX/RxJS/pull/849, supposed to update PR but stuck whole day with meetings. apologize for being late :(

This PR introduces changes
- mimic export behavior of `Immediate`, symbol is exported by simple factory does root patching interally
- create test coverages of symbols

Maybe this could be not the recommended way, any suggestions would be welcome. will update PR accordingly.

### DIFF
--- a/spec/util/SymbolShim-spec.js
+++ b/spec/util/SymbolShim-spec.js
@@ -1,0 +1,102 @@
+/* globals __root__ */
+var SymbolDefinition = require('../../dist/cjs/util/SymbolShim').SymbolDefinition;
+var Map = require('../../dist/cjs/util/Map').Map;
+var Rx = require('../../dist/cjs/Rx');
+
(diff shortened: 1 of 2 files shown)


blesh:
I feel like we're not using `Symbol.dispose` for anything and we can remove it. Thoughts?

kwonoj:
>  we can remove it.

: Yes, that seems better. We do not have any reference to `dispose` for now. I'll update PR to remove those.

kwonoj:
PR's updated to remove `dispose` symbol.

blesh:
LGTM.

kwonoj:
Thanks, I'll merge it myself.

kwonoj:
Merged with 2c5bf41 .

PR #862, opened 2015-12-02
test(mergeAll): add tests for limited concurrency cases

staltz:
Add tests for mergeAll to assert its behavior when given a finite concurrency parameter, such as
concurrent=1 and concurrent=2.


kwonoj:
Merged with 75abef1, thanks @staltz

PR #863, opened 2015-12-02
refactor(Observable): Subjects no longer wrapped in Subscriber

### LINKED ISSUE #748: Observable.subscribe checks 'instanceof' Subscription instead of shape
Currently `Subject` extends `Observable` and only implements `Subscription` so it doesn't see the `Subscription` function on it's inheritance chain.  This is an issue because currently the `subscribe` method on `Observable` checks whether the `subscriber` is an `instanceof Subscription`.  If the `subscriber` is a `Subject` it returns `false` and then wraps `subscriber` in a `Subscriber` and proceeds.  
```
      if (observerOrNext instanceof Subscriber) {
        subscriber = (<Subscriber<T>> observerOrNext);
      } else {
        subscriber = new Subscriber(<Observer<T>> observerOrNext);
      }
```
This seems faulty because this behavior will hide any attempt to override any of the `Subscriber` methods in the `Subject`. 
```
export class CustomSubject<T> extends Subject<T> {
  constructor(private _numObservables: number = 0) {
    super();
  }

  // Never gets called
  add(subscription?) {
    console.log("Add called")
  }
}
```
I was trying to create a custom `Subject` that reference counts the number of Observables it is subscribed to so it can defer calling `complete` until the last observable completes but this became an issue.  I don't know if this is a bug, but it seems like non-ideal behavior.  Would it be possible to add `Subscriber` to the prototypal chain of `Subject`, or,  in `Observable.subscribe` change the `instanceof` call to instead check the shape of the passed in `subscriber`?

### LINKED ISSUE #825: Using Subjects as Observers

ntilwalli:
Observable.subscribe updated to check for instanceof Subject, test added, import order changed in Rx and KitchenSink to avoid circular dependency issue. closes #825, closes #748

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -89,7 +89,7 @@ export class Observable<T> implements CoreOperators<T>  {
     let subscriber: Subscriber<T>;
 
     if (observerOrNext && typeof observerOrNext === 'object') {
-      if (observerOrNext instanceof Subscriber) {
+      if (observerOrNext instanceof Subscriber || observerOrNext instanceof Subject) {
         subscriber = (<Subscriber<T>> observerOrNext);
       } else {
         subscriber = new Subscriber(<Observer<T>> observerOrNext);
--- a/spec/observable-spec.js
+++ b/spec/observable-spec.js
@@ -109,6 +109,13 @@ describe('Observable', function () {
       expect(unsubscribeCalled).toBe(true);
     });
 
+    it('should return the given subject when called with a Subject', function () {
+      var source = Observable.of(42)
+      var subject = new Rx.Subject()
+      var subscriber = source.subscribe(subject)
+      expect(subscriber).toBe(subject)
+    })
+
     describe('when called with an anonymous observer', function () {
       it('should accept an anonymous observer with just a next function', function () {
         Observable.of(1).subscribe({
--- a/src/Rx.KitchenSink.ts
+++ b/src/Rx.KitchenSink.ts
@@ -1,3 +1,6 @@
+/* tslint:disable:no-unused-variable */
+import {Subject} from './Subject';
+/* tslint:enable:no-unused-variable */
 import {Observable} from './Observable';
 import {CoreOperators} from './CoreOperators';
 import {Scheduler as IScheduler} from './Scheduler';
@@ -125,7 +128,6 @@ import './operator/zip';
 import './operator/zipAll';
 
 /* tslint:disable:no-unused-variable */
-import {Subject} from './Subject';
 import {Subscription} from './Subscription';
 import {Subscriber} from './Subscriber';
 import {AsyncSubject} from './subject/AsyncSubject';
--- a/src/Rx.ts
+++ b/src/Rx.ts
@@ -1,3 +1,9 @@
+/* tslint:disable:no-unused-variable */
+// Subject imported before Observable to bypass circular dependency issue since
+// Subject extends Observable and Observable references Subject in it's
+// definition
+import {Subject} from './Subject';
+/* tslint:enable:no-unused-variable */
 import {Observable} from './Observable';
 
 // statics
@@ -101,7 +107,6 @@ import './operator/zip';
 import './operator/zipAll';
 
 /* tslint:disable:no-unused-variable */
-import {Subject} from './Subject';
 import {Subscription} from './Subscription';
 import {Subscriber} from './Subscriber';
 import {AsyncSubject} from './subject/AsyncSubject';


ntilwalli:
This PR addresses the issue where Subjects are unnecessarily wrapped in a Subscriber during Observable.subscribe.  Conceptually there is no reason they should be since they are designed to be both Observable and Subscriber.  This fix does introduce a circular dependency between Subject and Observable that must be managed (a bit inelegantly) during initial import in Rx and KitchenSink.  Is introducing module load ordering dependencies okay?  (Also after checkin I noticed I forgot to remove a commented out line of code in Observable.ts)

blesh:
LGTM.

blesh:
@ntilwalli can you rephrase this commit as `refactor(Observable): Subjects no longer wrapped in Subscriber` with a description of the work done, also add `closes #825, closes #748` to the commit message at the end.

The reason for the change from `fix` to `refactor`, is technically the functionality wasn't "broken" before, it was just sub-optimal.

Other than that, LGTM! Thanks for the PR!

src/Observable.ts
```diff
@@ -89,7 +89,8 @@ export class Observable<T> implements CoreOperators<T>  {
     let subscriber: Subscriber<T>;
 
     if (observerOrNext && typeof observerOrNext === 'object') {
-      if (observerOrNext instanceof Subscriber) {
+      //if (observerOrNext instanceof Subscriber) {
```
kwonoj: would you able to remove comment if this is leftover?

kwonoj:
Except one minor comment about commented code, change looks good to me too.

ntilwalli:
Great.  Done.

ntilwalli:
Great.  Done.

kwonoj:
@ntilwalli , would you able to flat 2 commit 42258b27b059f60eb5fb64d5532ef76be7e9f854 & 849d5b25b71e3e0e2743981965a504a5e2c5bbc1 into single one? both has identical commit message, and change seems quite similar too.

staltz:
From issue #825, @blesh said:

> Subjects are being unnecessarily wrapped in Subscribers

I originally said the same. But after looking at code, I noticed that Subjects are **not** Subscribers. A Subject is an Observable and an Observer. And a Subscriber is an Observer. Every Observer is wrapped by a Subscriber when we subscribe with an observer. That's why a Subject is wrapped by a Subscriber.

Look at this inheritance diagram:

photo_2015-12-03_10-33-00

Yet this PR introduces:

```js
if (observerOrNext instanceof Subscriber || observerOrNext instanceof Subject) {
  subscriber = (<Subscriber<T>> observerOrNext);
}
```

Does TypeScript allow us to (unsafe?) cast `observerOrNext` to Subscriber if it's a Subject?

kwonoj:
> Does TypeScript allow us to (unsafe?) cast...

: Since typescript only does type inference on build time, above code snippet could get passed.

`observerOrNext` is union type with primary type of `Observer<T>`, so regardless of runtime type checking (`observerOrNext instanceof ...`) typescript allows given parameter to be casted as `Subscriber`. Course, when code is being executed if assigned value is not `Subscriber`, it'll cause runtime issue.

if you're trying to cast `Subject` explicitly to `Subscriber` it'll complains, but also interestingly this snippet passes type inference.

```
let t:Subject<T> = ...;
subscribe(t);
```

while `t` is neither `Observer` or `((value: T) => void`.

Anyway, you're inheritance diagram seems legit, `Subject` seems does not inherit `Subscriber`.

kwonoj:
Removed code snippet that I was thinking incorrectly.

kwonoj:
> code is being executed if assigned value is not Subscriber, it'll cause runtime issue. 

: and seems this is not occurring since execution will be reach `add(...)` which is available implementation on `Subject` & `Subscriber` both.

staltz:
Is reimplementing Subject as extending both Observable and Subscriber too much work? Since Subject implements both Observer and Subscription anyway.

kwonoj:
I'm curious if subject need to extend subscriber. Nature of subject requires to be Observable and Observer as minimum - is there any specific usecases subscriber would benefit subject?

thinking about inheritance & implementation hierarchy, extending subscriber will makes subject

`Subject<T> extends Observable<T>, Subscriber<T> implements Observer<T>, Subscription<T>` (pseudo, let's forget about typescript doesn't support multiple inheritance. course in practical code it'll be more complex cause it need workaround for multiple inheritance..)

makes subject quite complex object.

staltz:
> makes subject quite complex object.

I know, so maybe it's ok to keep it wrapped in a Subscriber.

ntilwalli:
@staltz said:
>Is reimplementing Subject as extending both Observable and Subscriber too much work? Since Subject >implements both Observer and Subscription anyway.
What does *reimplementing* mean? Since Typescript is purely duck-typed, from a compile-time/type-checking/semantic  perspective, there's no difference between the statements:

1. `class Subject extends Observable<T> implements Observer<T>, Subscription<T>`
2. `class Subject extends Observable<T> implements Subscriber<T>`

Which means the cast is not unsafe. There *is* a difference between statements 1 and 2, and this:

3. `class Subject extends Observable<T> implements Observer<T>`

If the goal from from a semantic perspective is statements 1/2 then from a clarity perspective, 2 might be better and this PR is valid.  If the goal is 3 then this PR is not valid and there should be a PR to modify Subject accordingly.  When I first proposed this issue, I assumed Subject was coded correctly and Observable was inconsistent. It sounds like you're saying it's the other way around.

ntilwalli:
I edited the above comment fix a reasoning error and to be more precise as to my intended meaning.

blesh:
It will be fine. Are we tricking TypeScript? Yes. But we're doing it because TypeScript types aren't a real thing, they're helpful for build-time and development-time sanity checks. When you see `<casting>obj` anywhere, it's basically saying "I don't want your sanity checks, because I know what I'm doing". In this case, it works out. Subject and Subscriber have all of the same API points. This is by design.

This PR should be a small performance improvement when using Subjects to subscribe, and I don't see anything about it that could be bad. ... Other than people who _really_ like typing freaking out a little because we're being dirty, rotten liars about Subject being a Subscriber on that one line.

ntilwalli:
@kwonoj I've flattened the commits.

kwonoj:
@ntilwalli Thanks! I'll check in this after waiting couple of more hrs to see if there's any second opinion.

staltz:
> This PR should be a small performance improvement when using Subjects to subscribe, and I don't see anything about it that could be bad. ... Other than people who really like typing freaking out a little because we're being dirty, rotten liars about Subject being a Subscriber on that one line.

Fair enough...

kwonoj:
Merged with 5cb0f2b, thanks @ntilwalli !

ntilwalli:
@kwonoj I just realized, when flattening the commits, I forgot to update the commit message to say "refactor" as @blesh asked.  It currently says "fix".  Should I amend my message and push it again?

kwonoj:
Whew, sorry for that. I should've keep track detail before merge.. I think current branch is protected from force update, would like to hear @blesh which would be best.

PR #865, opened 2015-12-02
refactor(toPromise): update type definition for toPromise

### LINKED ISSUE #864: toPromise on Observable requires promise constructor argument
https://github.com/ReactiveX/RxJS/blob/master/src/Observable.ts#L247

The constructor should be optional, as it is implemented in the operator itself https://github.com/ReactiveX/RxJS/blob/master/src/operator/toPromise.ts

kwonoj:
- make Promistctor parameter as optional in type definition

closes #864


kwonoj:
Merged with dee84a3.

blesh:
Good catch, @kwonoj

PR #866, opened 2015-12-02
fix(AsyncSubject): emit value when it's subscribed after complete

luisgabriel:
Currently, if a new subscription happens after `complete`, AsyncSubject will emit just `complete` instead of `next` and then `complete`.


trxcllnt:
LGTM 👍🏽

blesh:
LGTM.

kwonoj:
Change looks good to me. I'll merge this around tonight.

kwonoj:
Merged with ed0eaf6, thanks @luisgabriel ! :+1:

PR #867, opened 2015-12-03
chore(eslint): disable auto fix

kwonoj:
- disable eslint autofix option

This PR disable autofix option of eslint. Originally I have introduced this to all js codes are being fixed automatically before creating PR, but could see possible side effect when author does not run lint locally, travis passes since it's correcting lint failure automatically which is not being reflected into PR. 

For now by disabling makes lint explicit while looking for better ergonomics could satisfy dev / travis check both.


kwonoj:
Merged with f1dc764.

PR #870, opened 2015-12-03
Adding plumbing to support more granular and maintainable typings

david-driscoll:
This is based on the original #715, I've broken it out into 2 branches so far, with other branches to follow after getting some feedback.

Today Rx has 3 or 4 places where the typings have to be updated, for the consumer to be able to actually get type information and consume that type information.

* Observable.ts
* Rx.KitchenSink.ts
* CoreOperators.ts
* The operator in question

The goal here is to be able to add well defined typings for an operator, and then extract that type information using a common convention, into another interface.  This extraction is an automated process, such that as changes are made to the operator the interface will automatically be updated.
This new interface for the operator can then be used where it needs to be, by simply using `typeof`.  

#715 was a proof of concept for me, and I only added all the typings I did because I was testing it with `omnisharp-atom` and `omnisharp-client`, both which use Rx4 extensively.  Once all the work was done, the compile times were reduced, and there were minimal changes that needed to be made to the source for things to "compile" correctly (I ran into other issues, but that isn't important to this case)

#### Pros
* All the type information now lives with the operator
* Only have to make a change to 1 file, to update the consumer.
* Allows us to have enhanced types for the operators, things like tuples for `combineLatest` instead of `Observable<any[]>

#### Cons
* Additional step in the build workflow.
* Adds some magic contributors need to know about
* The operator injection onto `Observable` might be obsolete with TS 1.8
  * I don't think we'll have the ability to type `this` yet, which is still one thing this tooling would be valuable for.
* Currently doing this one operator at a time breaks the build, for no other reason than a stackoverflow. :(
* _I'm sure there are cons I'm missing at the moment_

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -12,6 +12,25 @@ import {Subject} from './Subject';
 import {Notification} from './Notification';
 import {rxSubscriber} from'./symbol/rxSubscriber';
 
+import {combineLatest as combineLatestStatic} from './operator/combineLatest-static';
+import {concat as concatStatic} from './operator/concat-static';
+import {merge as mergeStatic} from './operator/merge-static';
+import {zip as zipStatic} from './operator/zip-static';
+import {BoundCallbackObservable} from './observable/bindCallback';
+import {DeferObservable} from './observable/defer';
+import {EmptyObservable} from './observable/empty';
+import {ForkJoinObservable} from './observable/forkJoin';
+import {FromObservable} from './observable/from';
+import {ArrayObservable} from './observable/fromArray';
+import {FromEventObservable} from './observable/fromEvent';
+import {FromEventPatternObservable} from './observable/fromEventPattern';
+import {PromiseObservable} from './observable/fromPromise';
+import {IntervalObservable} from './observable/interval';
+import {TimerObservable} from './observable/timer';
+import {RangeObservable} from './observable/range';
+import {InfiniteObservable} from './observable/never';
+import {ErrorObservable} from './observable/throw';
+
 /**
  * A representation of any set of values over any amount of time. This the most basic building block
  * of RxJS.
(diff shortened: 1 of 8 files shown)


src/Observable.ts
```diff
@@ -11,6 +11,28 @@ import {ConnectableObservable} from './observable/ConnectableObservable';
...
+import {CallbackObservable} from './observable/fromCallback';
+import {ArrayObservable} from './observable/fromArray';
+import {FromEventObservable} from './observable/fromEvent';
+import {FromEventPatternObservable} from './observable/fromEventPattern';
+import {PromiseObservable} from './observable/fromPromise';
+import {IntervalObservable} from './observable/interval';
+import {TimerObservable} from './observable/timer';
+import {RangeObservable} from './observable/range';
+import {InfiniteObservable} from './observable/never';
+import {ErrorObservable} from './observable/throw';
```
blesh: I think this is going to introduce circular dependencies. If you rebase, you can run `npm run check_circular_dependencies` and see.

blesh:
I'm still not 100% about this PR. It looks like something that will require maintenance over time, and the benefit seems dubious. I'm willing to admit I could be _totally_ wrong, though. I'm also really waiting for more feedback from heavy TypeScript consumers like the Angular team. @IgorMinar @jeffbcross @robwormald

src/Observable.ts
david-driscoll: They'll get thrown away, because they're just used strictly for the types.

david-driscoll:
I would love some more input as well because right now the TypeScript experience with RxJS5 is honestly quiet terrible.  It forces you down a path of typing everything, when the types should just flow through without issue, if there were proper typings.  In fact in some situations, if the class / interface isn't exported, you may not even be able to explicitly type anything without having to redefine the type yourself.

Additionally with the completed solution, I was seeing compile times reduced significantly, only because I think the compiler didn't have to work as hard keeping track of all anonymous types.  So once it's all finished, there may be some compile perf wins as well.

blesh:
> I would love some more input as well because right now the TypeScript experience with RxJS5 is honestly quiet terrible.

Same here, honestly. I'm mostly consuming this as ES6 with Babel or with Node and CJS.

> Additionally with the completed solution, I was seeing compile times reduced significantly, only because I think the compiler didn't have to work as hard keeping track of all anonymous types.

That sounds good.

My only concern is whether the additional maintenance is worth the benefits. So far I have very little evidence to work off of. I did a _little_ work with TypeScript and Angular using RxJS 5 for AngularConnect a few months ago, and I didn't notice any TS ergonomics issues.

Perhaps as we move the unit tests over the TypeScript (one proposal here #826), the issues this is attempting to solve will become more apparent. /cc @kwonoj

trxcllnt:
@blesh I disagree, this PR looks great. I had no idea TS let you export types like this. Great work @david-driscoll. Perhaps the `fileResult` data in typingsgen.js could be in a separate `.ts` (so we don't have to maintain a string as code in a script file)?

kwonoj:
In general I like essence of this PR itself. at least aspect of precise type definition of interface, results created by this PR (or extended PR https://github.com/ReactiveX/RxJS/pull/871 to illustrate example) would serve better than current type declarations. But also have few curious regarding ergonomics of development like when someone try to add new operator, or if it's required to modify signatures of few operator only, etcs and other I might not able to think of by lean on code based interface generation.

robwormald:
I definitely like the end result.

 The pain points @david-driscoll mentions are real, in terms of having types properly propagate through operators. I'm less excited about the manual code generation bit, if only because it adds complexity to the process for contribs. I'll pull this down this evening and have a proper play with it. Either way, nice work!

typingsgen.js
```diff
@@ -0,0 +1,74 @@
...
+import {Observable} from \'./Observable\';\n\
+import {ConnectableObservable} from \'./observable/ConnectableObservable\';\n\
+import {Scheduler} from \'./Scheduler\';\n\
+import {Notification} from \'./Notification\';\n\
+import {Subject} from \'./Subject\';\n\
+import {Observer} from \'./Observer\';\n\
+import {GroupedObservable} from \'./operator/groupBy-support\';\n\
+import {GroupByObservable} from \'./operator/groupBy\';\n\
+import {TimeInterval} from \'./operator/extended/timeInterval\';\n\
+import {ObservableOrPromise, ArrayOrIterator, _Selector, _IndexSelector, _SwitchMapResultSelector, _ObservableMergeMapProjector, _IteratorMergeMapProjector, _Predicate, _PredicateObservable, _Comparer, _Accumulator, _MergeAccumulator} from \'./types\';\n\n\
```
blesh: Is there any way we can not have to manually maintain this?

david-driscoll: It would still have to be maintained in some way most likely.  That said we have a few things we could do to change how it works.

Change it around such that we edit the existing file, instead of writing a new one.  That way if any new definitions bring in additional Observables / Operators that need to be referenced, we can just update the file itself, and no rely only on the file generation.

david-driscoll:
In reality we need the following issues in TypeScript to land, to avoid having to deal with the code generation.

https://github.com/Microsoft/TypeScript/pull/5290 and https://github.com/Microsoft/TypeScript/issues/5269 plus we also need one one of https://github.com/Microsoft/TypeScript/issues/3508 (function bind syntax `::`) or https://github.com/Microsoft/TypeScript/issues/3694 (ability to type `this` for a function)

david-driscoll:
It may also be worth investigating https://github.com/Microsoft/TypeScript/pull/5290 to see if that could help (perhaps have the interfaces live inside the operators themselves?)

kwonoj:
> It may also be worth investigating Microsoft/TypeScript#5290 to see

: would you able to elaborate usecases for this PR to those feature?

david-driscoll:
I don't know if it it'll work, may be blocked by some of the comments from https://github.com/Microsoft/TypeScript/issues/5292

I just tried, and it only works if the class and interface declarations happen to be in the same file currently, I even tried `export {Observable} from './operators/combineLatest';` to see if it would work, and while TypeScript wouldn't complain, it also wouldn't propagate type information.

david-driscoll:
Rebased onto origin/master

blesh:
I'm not sure why my other comment didn't make it.

I'm by _no means_ "against" this PR. I just don't want to make uneducated decisions and wanted to lean on the larger community of people that spend more type using RxJS 5 with TypeScript. I'm going to hand off handling of this to @kwonoj, and I'll review it before it's merged. My only concerns lay around maintenance, so if we can keep that as smooth as possible, I have no qualms.

src/Observable.ts
kwonoj: Would you able to share how this will be removed?

robwormald: when it's transpiled to JS, it'll strip anything that's only used for type definitions (but retained in the .d.ts file)

```typescript
import {Foo} from './foo'

function doStuff(foo: Foo){}
```
=>
```javascript
function doStuff(foo){}
```

kwonoj: I see. I'm feeling this specific changes (import types then assign static method with bring it via `typeof`) is orthogonal and can be separated - is my understanding correct? (at least just taking this changes locally into TOT seems working)

david-driscoll:
@blesh My tone may have sounded a little harsher than I was intending, sorry about that :smile:.

My goal is to help make something that we call an appreciate for years to come, by no means is it the only, or even the best solution, I do feel it fits in with the overall design that is currently here today.

src/operator-typings.ts
```diff
@@ -0,0 +1,14 @@
...
+import {Scheduler} from './Scheduler';
+import {Notification} from './Notification';
+import {Subject} from './Subject';
+import {Observer} from './Observer';
+import {GroupedObservable} from './operator/groupBy-support';
+import {GroupByObservable} from './operator/groupBy';
+import {TimeInterval} from './operator/extended/timeInterval';
+import {ObservableOrPromise, ArrayOrIterator, _Selector, _IndexSelector, _SwitchMapResultSelector, _ObservableMergeMapProjector, _IteratorMergeMapProjector, _Predicate, _PredicateObservable, _Comparer, _Accumulator, _MergeAccumulator} from './types';
+
+/* ||| MARKER ||| */
```
david-driscoll: As part of my rebase I also made a tweak here, any generate-rated interfaces will be but between the two markers, allowing maintenance to happen directly in the `operator-typings.ts` file, if we need to add in a new import for example.

src/Observable.ts
david-driscoll: Yeah, those changes by themselves don't really depend on everything else

kwonoj: What do you think about separate this change into another PR? this change's quite straightforward and small, can be checked easily. Also it'll reduce this PR's scope.

PR #871, opened 2015-12-03
Update typings for combineLatest

david-driscoll:
This is an example of the enhanced typings for #870

You can put your focus on `src/operator/combineLatest.ts` all the other files are related to #870.

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -19,11 +19,11 @@ import {combineLatest as combineLatestStatic} from './operator/combineLatest-sta
 import {concat as concatStatic} from './operator/concat-static';
 import {merge as mergeStatic} from './operator/merge-static';
 import {zip as zipStatic} from './operator/zip-static';
-import {BoundCallbackObservable} from './observable/bindCallback';
 import {DeferObservable} from './observable/defer';
 import {EmptyObservable} from './observable/empty';
 import {ForkJoinObservable} from './observable/forkJoin';
 import {FromObservable} from './observable/from';
+import {CallbackObservable} from './observable/fromCallback';
 import {ArrayObservable} from './observable/fromArray';
 import {FromEventObservable} from './observable/fromEvent';
 import {FromEventPatternObservable} from './observable/fromEventPattern';
@@ -178,7 +178,7 @@ export class Observable<T> implements CoreOperators<T>  {
   }
 
   // static method stubs
-  static bindCallback: typeof BoundCallbackObservable.create;
+  static bindCallback: typeof CallbackObservable.create;
   static combineLatest: typeof combineLatestStatic;
   static concat: typeof concatStatic;
   static defer: typeof DeferObservable.create;
@@ -206,9 +206,7 @@ export class Observable<T> implements CoreOperators<T>  {
   bufferWhen: (closingSelector: () => Observable<any>) => Observable<T[]>;
   catch: (selector: (err: any, source: Observable<T>, caught: Observable<any>) => Observable<any>) => Observable<T>;
   combineAll: <R>(project?: (...values: Array<any>) => R) => Observable<R>;
-  combineLatest: <R>(...observables: Array<Observable<any> |
-                                     Array<Observable<any>> |
-                                     ((...values: Array<any>) => R)>) => Observable<R>;
+  combineLatest: operator.operator_proto_combineLatest<T>;
   concat: <R>(...observables: (Observable<any> | Scheduler)[]) => Observable<R>;
   concatAll: () => Observable<any>;
   concatMap: <R>(project: ((x: T, ix: number) => Observable<any>), projectResult?: (x: T, y: any, ix: number, iy: number) => R) => Observable<R>;
@@ -269,9 +267,9 @@ export class Observable<T> implements CoreOperators<T>  {
   startWith: (x: T) => Observable<T>;
   subscribeOn: (scheduler: Scheduler, delay?: number) => Observable<T>;
   switch: <R>() => Observable<R>;
-  exhaust: <T>() => Observable<T>;
+  switchFirst: <T>() => Observable<T>;
   switchMap: <R>(project: ((x: T, ix: number) => Observable<any>), projectResult?: (x: T, y: any, ix: number, iy: number) => R) => Observable<R>;
-  exhaustMap: <T, R, R2>(project: (x: T, ix: number) => Observable<R>, rSelector?: (x: T, y: R, ix: number, iy: number) => R2) => Observable<R>;
+  switchFirstMap: <T, R, R2>(project: (x: T, ix: number) => Observable<R>, rSelector?: (x: T, y: R, ix: number, iy: number) => R2) => Observable<R>;
   switchMapTo: <R>(observable: Observable<any>, projectResult?: (x: T, y: any, ix: number, iy: number) => R) => Observable<R>;
   take: (count: number) => Observable<T>;
   takeUntil: (notifier: Observable<any>) => Observable<T>;
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -12,6 +12,25 @@ import {Subject} from './Subject';
 import {Notification} from './Notification';
 import {rxSubscriber} from'./symbol/rxSubscriber';
 
+import {combineLatest as combineLatestStatic} from './operator/combineLatest-static';
+import {concat as concatStatic} from './operator/concat-static';
+import {merge as mergeStatic} from './operator/merge-static';
+import {zip as zipStatic} from './operator/zip-static';
+import {BoundCallbackObservable} from './observable/bindCallback';
+import {DeferObservable} from './observable/defer';
+import {EmptyObservable} from './observable/empty';
+import {ForkJoinObservable} from './observable/forkJoin';
+import {FromObservable} from './observable/from';
+import {ArrayObservable} from './observable/fromArray';
+import {FromEventObservable} from './observable/fromEvent';
+import {FromEventPatternObservable} from './observable/fromEventPattern';
+import {PromiseObservable} from './observable/fromPromise';
+import {IntervalObservable} from './observable/interval';
+import {TimerObservable} from './observable/timer';
+import {RangeObservable} from './observable/range';
+import {InfiniteObservable} from './observable/never';
+import {ErrorObservable} from './observable/throw';
+
 /**
  * A representation of any set of values over any amount of time. This the most basic building block
  * of RxJS.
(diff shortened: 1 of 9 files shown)


src/Observable.ts
```diff
@@ -134,33 +156,25 @@ export class Observable<T> implements CoreOperators<T>  {
...
+  static fromEventPattern: typeof FromEventPatternObservable.create;
+  static fromPromise: typeof PromiseObservable.create;
+  static interval: typeof IntervalObservable.create;
+  static merge: typeof mergeStatic;
+  static never: typeof InfiniteObservable.create;
+  static of: typeof ArrayObservable.of;
+  static range: typeof RangeObservable.create;
+  static throw: typeof ErrorObservable.create;
+  static timer: typeof TimerObservable.create;
+  static zip: typeof zipStatic;
```
blesh: the `typeof` trick is pretty cool. I didn't know that one. I'm a little concerned about the implications of importing all of the other Observable types in this file, though.

src/operator-typings.ts
```diff
@@ -0,0 +1,44 @@
...
+  <T2, T3, TResult>(second: ArrayOrIterator<T2>, third: ArrayOrIterator<T3>, project: (v1: T, v2: T2, v3: T3) => TResult): Observable<TResult>;
+  <T2, T3, TResult>(second: ObservableOrPromise<T2>, third: ArrayOrIterator<T3>, project: (v1: T, v2: T2, v3: T3) => TResult): Observable<TResult>;
+  <T2, T3, TResult>(second: ObservableOrPromise<T2>, third: ArrayOrIterator<T3>, project: (v1: T, v2: T2, v3: T3) => TResult): Observable<TResult>;
+  <T2, T3, TResult>(second: ArrayOrIterator<T2>, third: ObservableOrPromise<T3>, project: (v1: T, v2: T2, v3: T3) => TResult): Observable<TResult>;
+  <T2, T3, TResult>(array: [ObservableOrPromise<T2>, ObservableOrPromise<T3>], project: (v1: T, v2: T2, v3: T3) => TResult): Observable<TResult>;
+  <TResult>(array: ObservableOrPromise<any>[], project?: Function): Observable<TResult[]>;
+  (...observables: Array<ObservableOrPromise<T>>): Observable<T[]>;
+  (...observables: Array<ArrayOrIterator<T>>): Observable<T[]>;
+  <R>(...observables: Array<ObservableOrPromise<T> | ((...values: Array<T>) => R)>): Observable<R>;
+  <R>(...observables: Array<ArrayOrIterator<T> | ((...values: Array<T>) => R)>): Observable<R>;
```
blesh: This is a completely crazy amount of definitions for one method. Is this really necessary? It seems like we shouldn't have to do this, and that if this many definitions are required it's really an issue with TypeScript that should be reported and dealt with.

tsconfig.json
```diff
@@ -7,8 +7,22 @@
...
+    "convertTabsToSpaces": true,
+    "insertSpaceAfterCommaDelimiter": true,
+    "insertSpaceAfterSemicolonInForStatements": true,
+    "insertSpaceBeforeAndAfterBinaryOperators": true,
+    "insertSpaceAfterKeywordsInControlFlowStatements": true,
+    "insertSpaceAfterFunctionKeywordForAnonymousFunctions": false,
+    "insertSpaceAfterOpeningAndBeforeClosingNonemptyParenthesis": false,
+    "placeOpenBraceOnNewLineForFunctions": false,
+    "placeOpenBraceOnNewLineForControlBlocks": false
+  },
```
blesh: What's this for?

typingsgen.js
```diff
@@ -0,0 +1,74 @@
...
+    } while(result);
+
+    if (!operators[filename]) {
+      operators[filename] = true;
+      fileResult += 'export interface operator_proto_' + filename + '<T> {\n  ' + methods.join(';\n  ') + ';\n}\n';
+    }
+  }
+}
+
+fs.writeFileSync('./src/operator-typings.ts', fileResult);
```
blesh: I see the above stated "crazy number of definitions" was generated by this script. Pretty cool, but again, is this something that's really necessary? How does it help? Do we have an example?

blesh:
I'd like people who *really* care about the TS usability of this lib to comment: @IgorMinar, @jeffbcross, @robwormald

src/operator-typings.ts
kwonoj: This seems similar to patterns of some C# interfaces, including some of Rx operators like `combinelatest`. basically this'll allows to have better typing inference experiences (such as intellisense?). Matter of if this is necessarily required for typescript land with RxJS, not sure. Bit lean on we might not need this.. especially thinking about effort to hand crafted operator signature would be increased in great amount, since lot of Rx operators has similar parameter patterns.

p.s : In case of Rx.Net, these type are generated via t4 (https://github.com/Reactive-Extensions/Rx.NET/blob/859e6159cb07be67fd36b18c2ae2b9a62979cb6d/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable/CombineLatest.Generated.tt)

blesh:
@david-driscoll I can see you've put a TON of work and thought into this. I'm really just trying to wrap my head around the needs and use cases this work is meant to meet. Whenever I see this much time and thought put into something, I _really_ want to merge it, but it's on me to make sure it makes sense to do so.

src/operator-typings.ts
david-driscoll: I'll give some more concrete examples in the normal comments to demonstrate how it really helps to flow the types, so you can basically avoid having to manually cast your method calls as you're coding.  Basically lets you take full advantage of TypeScript.

src/Observable.ts
david-driscoll: If the imports are just use for types, they dissolve when TypeScript is emitting the results.

For example the `cjs` output for `Observable.ts` starts with

```
var Subscriber_1 = require('./Subscriber');
var root_1 = require('./util/root');
var SymbolShim_1 = require('./util/SymbolShim');
/**
 * A representation of any set of values over any amount of time. This the most basic building block
 * of RxJS.
 *
 * @class Observable<T>
 */
var Observable = (function () {
```

david-driscoll:
@blesh I totally understand 100%.  I did the same amount of work for RxJS4 just to make my life working with the library that much easier.  In some ways the value of TypeScript is kind of lost if your typings don't describe all the features.  Sadly it's a case of JavaScript can do so much, so differently, describing it can be hard.

I'll take apart the `combineLatest` definitions to sort of explain what they do, and an example of how they can help.

Legend:
* o: `Observable`
* i: `Iterable`
* ar: `Array`
* p: `Promise`

Today `combineLatest` is defined as:

```
export function combineLatest<R>(...observables: Array<Observable<any> |
  Array<Promise<any>> |
  ((...values: Array<any>) => R)>): Observable<R> {
  let project: (...values: Array<any>) => R = null;
  if (typeof observables[observables.length - 1] === 'function') {
    project = <(...values: Array<any>) => R>observables.pop();
  }

  // if the first and only other argument besides the resultSelector is an array
  // assume it's been called with `combineLatest([obs1, obs2, obs3], project)`
  if (observables.length === 1 && isArray(observables[0])) {
    observables = <any>observables[0];
  }

  observables.unshift(this);

  return new ArrayObservable(observables).lift(new CombineLatestOperator(project));
}

Observable.prototype.combineLatest = combineLatest;
```

So basically it can be called many many different ways.  With an array, with a rest param array, without a projection function, with a projection function, etc.

`.combineLatest([o])`
`.combineLatest([p], (res_o) => res_o + 1)`
`.combineLatest(i)`
`.combineLatest(ar, (res_ar) => res_ar + 'abc')`

This example is fairly arbitrary as, `p` will only emit one value, but `subscribeToResult` allows for all `Obserable`'s, `Promise`'s, `Iterator`'s and `Array`'s.  All can have their own generic value from the type system, and there is no guarantee that they will be the same type.

The objective for me as a consumer, is that I want to have the type of `res_o` available as it were the value that is known to be emitted by the `Observable<T2>`, when I'm working in the `project` function, preferably without having to declare a type annotation.

For a more concrete demo, I'll use the playground.

[Today's typings](http://www.typescriptlang.org/Playground/#src=interface%20Promise%3CT%3E%20%7B%7D%0Ainterface%20Observable%3CT%3E%20%7B%0A%09combineLatest%3CR%3E(...observables%3A%20Array%3Cany%20%7C%20Observable%3Cany%3E%20%7C%0A%20%20Array%3CPromise%3Cany%3E%3E%20%7C%0A%20%20((...values%3A%20Array%3Cany%3E)%20%3D%3E%20R)%3E)%20%3A%20Observable%3CR%3E%3B%0A%20%20%20%20map%3CR%3E(project%3A%20(x%3A%20T%2C%20index%3F%3A%20number)%20%3D%3E%20R)%3A%20Observable%3CR%3E%3B%0A%7D%0A%0Avar%20a%20%3A%20Observable%3C%7Ba%3A%20string%7D%3E%3B%0Avar%20b%20%3A%20Observable%3C%7Bb%3A%20number%7D%3E%3B%0A%0A%2F%2F%20error%0Aa.combineLatest(b).map((%5Baa%2C%20bb%5D)%20%3D%3E%20(%7B%20valueB%3A%20bb.b%2C%20valueA%3A%20aa.a%20%7D))%3B%0A%0A%2F%2F%20annotated%0Aa.combineLatest%3C%5B%7Ba%3A%20string%7D%2C%20%7Bb%3A%20number%7D%5D%3E(b).map((%5Baa%2C%20bb%5D)%20%3D%3E%20(%7B%20valueB%3A%20bb.b%2C%20valueA%3A%20aa.a%20%7D))%3B%0A%0A)

[Enhanced typings](http://www.typescriptlang.org/Playground/#src=interface%20Promise%3CT%3E%20%7B%7D%0Ainterface%20Observable%3CT%3E%20%7B%0A%20%20%20%20combineLatest%3CT2%3E(second%3A%20Observable%3CT2%3E)%3A%20Observable%3C%5BT%2C%20T2%5D%3E%3B%0A%20%20%20%20map%3CR%3E(project%3A%20(x%3A%20T%2C%20index%3F%3A%20number)%20%3D%3E%20R)%3A%20Observable%3CR%3E%3B%0A%7D%0A%0Avar%20a%20%3A%20Observable%3C%7Ba%3A%20string%7D%3E%3B%0Avar%20b%20%3A%20Observable%3C%7Bb%3A%20number%7D%3E%3B%0A%0Aa.combineLatest(b).map((%5Baa%2C%20bb%5D)%20%3D%3E%20(%7B%20valueB%3A%20bb.b%2C%20valueA%3A%20aa.a%20%7D))%3B%0A%0A)


    <TResult>(project: (v1: T) => TResult): Observable<TResult>;
    <TResult>(project: (v1: T) => TResult): Observable<TResult>;
These are duplicated... whoops. :smile: 

    <T2>(second: ObservableOrPromise<T2>): Observable<[T, T2]>;
By default without a project method, we will output what is essentially a Tuple, so here we're typing it as a tuple of `T` coming from the observable (which is really `this`) and `T2` being the observable or promise that is being handed over.

Today with the code defined as above, we must intentionally say `.combineLatest<[T, T2]>(o)` to get the desired effect, otherwise all you get is `{}`.  This adds a lot of type annotation noise, and really defeats the purpose behind the value TypeScript is trying to give us.

    <T2>(array: [ObservableOrPromise<T2>]): Observable<[T, T2]>;
Arrays are a valid input type, this lets the consumer hand in a tuple, and get an appropriate output type.  Otherwise we would require a type annotation again.

    <T2, TResult>(second: ObservableOrPromise<T2>, project: (v1: T, v2: T2) => TResult): Observable<TResult>;
    <T2, TResult>(array: [ObservableOrPromise<T2>], project: (v1: T, v2: T2) => TResult): Observable<TResult>;
The further along definitions are really where things come together, we can strongly type `T` and `T2` and not have to create any type annotations in our methods, and TResult can simply be the return value of your method.

    <T2, T3>(second: ObservableOrPromise<T2>, third: ObservableOrPromise<T3>): Observable<[T, T2, T3]>;
Again we get to now type three different types, without having to declare types.

  <T2, T3>(second: ArrayOrIterator<T2>, third: ObservableOrPromise<T3>): Observable<[T, T2, T3]>;
  <T2, T3>(second: ArrayOrIterator<T2>, third: ArrayOrIterator<T3>): Observable<[T, T2, T3]>;
  <T2, T3>(second: ObservableOrPromise<T2>, third: ArrayOrIterator<T3>): Observable<[T, T2, T3]>;
   .... <snip> ....
The real complicated bit comes down to what I see as a compiler bug, and I just need to see if there is a bug out there to fix it.  Basically if you Union `ArrayOrIterator` with `ObservableOrPromise` then type inference breaks down, having the distinct overloads to handle each matrix (T, T2, T3, etc) solves the problem, but it makes the number of overloads grow exponentially.

This is perhaps something we don't need to model, because consumers could simply call `Obserable.of`.  That said I left it for now for demonstration purposes, I don't have a super strong opinion either way, my goal was to make the typings work with all the scenarios vanilla JS could potentially handle, it doesn't mean that we have to do that.  I would be happy to consider `Observable` and `Promise` first class citizens and delegate `Iterable` and `Array` to something that has to be converted to an `Observable` for TypeScript consumers.

david-driscoll:
Doing some reading and https://github.com/Microsoft/TypeScript/pull/5738 looks like it might solve the issue with array inference, I'll do some testing with `typescript@next` and see if that's the case.

PR #873, opened 2015-12-03
Remove duplicate links from docs

pselle:
Remove duplicates of combineLatest and concat


blesh:
@pselle I feel like this might have been automagically generated by esdoc, from something like the `/doc/index.md` file

pselle:
aha! I wondered where the docs came from (I figured they were generated given your last commit touched 300 files)

PR #880, opened 2015-12-03
AsyncSubject: Add more test cases

luisgabriel:


spec/subjects/async-subject-spec.js
```diff
@@ -1,19 +1,23 @@
 var Rx = require('../../dist/cjs/Rx');
 
 var AsyncSubject = Rx.AsyncSubject;
-var nextTick = Rx.Scheduler.nextTick;
 var Observable = Rx.Observable;
 
 describe('AsyncSubject', function () {
+  var results = [];
```
kwonoj: I remember this has been discussed between me and @staltz before, and generally agreed to allow duplicate to let test case to be self-contained. What do you think?

spec/subjects/async-subject-spec.js
```diff
@@ -52,34 +55,50 @@ describe('AsyncSubject', function () {
...
+    expect(results).toEqual([]);
+    subject.next(2);
+    expect(results).toEqual([]);
+    subject.complete();
+    expect(results).toEqual([2, 'done']);
+    subject.next(3);
+    expect(results).toEqual([2, 'done']);
+  });
+
+  it('should not emit values if unsubscribed before complete', function () {
```
kwonoj: Do we have test cases for 
- `subject` is unsubscribed early to see does not emit anything
- throws if try to emit after subject is completed

also?

spec/subjects/async-subject-spec.js
staltz: Definitely yes, avoid global state in tests, it's totally ok to repeat some common code in tests. The rule of thumb is: you should be able to copy a test and paste it in a discussion or in some execution environment, and everything you need for executing it is there available with no outer context assumed.

luisgabriel: I agree with you guys. I was also not very happy reusing var. I'll fix it and update this PR.

spec/subjects/async-subject-spec.js
luisgabriel: I think the first one is covered by the test case pointed out in this line.

The second point I'm not sure if I understood correctly. Should the subject throw an error or the observer?

kwonoj: for first case, what I'm thinking is 

```
subject.next(1);
subject.next(2);
subject.unsubscribe(); --> represents it's unsubscribed early, instead of completion 
expect(result)...
```

for second case, something like

```
subject.next(1);
subject.next(2);
subject.completed();
subject.next(3); --> per current implementation, expect to throw.
```

luisgabriel: I think both of these cases are covered in:
1. `should not emit values if unsubscribed before complete`
2. `should not emit values after complete`

But currently the second case does not throw an error. This seems to be also the behavior of RxJS 4: http://jsbin.com/mumujemesa/edit?js,console

What do you think?

kwonoj: first case seems right, I missed :)

actually for second, I think I wrote snippet incorrectly.

```
subject.subscribe(subscriber);
subscriber.unsubscribe(); -> subscriber has unsubscribed explicitly
subject.next('boom');
```

is this already covered also?

luisgabriel: Humm. Indeed there is no test case where the unsubscription happens right after subscription. I'll add this one and update the PR in a few minutes.

kwonoj: It isn't necessarily have to be unsubscribe right after , just shortened code.

it could be like
```
subscribe();
next();
next();
unsubscribe();
next(boom);
```

luisgabriel: I think this one is covered by `should not emit values if unsubscribed before complete`.

kwonoj: Cool, seems I'm fully-blinded eye today :sunglasses: sorry for those.

luisgabriel: No problem. ;)

blesh:
Merged with d2e6318... thanks again, @luisgabriel!

PR #884, opened 2015-12-04
test(publish): update test case for publish operator

kwonoj:
Small update to `publish` test cases.

cc @staltz - would you able to confirm if this is right changes?


staltz:
Yes, that's correct. Good catch.

kwonoj:
Cool, I'll check in this PR.

kwonoj:
Merged with 0574bd9.

PR #885, opened 2015-12-04
feat(publishLast): add publishLast operator

### LINKED ISSUE #883: Implement `publishLast`
Now that we have AsyncSubject #850, we need to add `publishLast`.

It's basically just:

```js
import {multicast} from './multicast';
import {AsyncSubject} from '../subject/AsyncSubject';

export function publishLast() {
  return mutlicast.call(this, new AsyncSubject());
}
```

kwonoj:
closes #883

This PR adds `publishLast` operator with initial test coverage, mostly mirrored from `publish` operator's test cases.

### DIFF
--- a/spec/operators/publishLast-spec.js
+++ b/spec/operators/publishLast-spec.js
@@ -1,0 +1,222 @@
+/* globals describe, it, expect, expectObservable, expectSubscriptions, hot, cold */
+var Rx = require('../../dist/cjs/Rx');
+var Observable = Rx.Observable;
+var Subject = Rx.Subject;
+
+describe('Observable.prototype.publishLast()', function () {
+  it('should return a ConnectableObservable', function () {
+    var source = Observable.of(1).publishLast();
+
+    expect(source instanceof Rx.ConnectableObservable).toBe(true);
+  });
+
+  it('should emit last notification of a simple source Observable', function () {
+    var source = cold('--1-2---3-4--5-|');
+    var sourceSubs =  '^              !';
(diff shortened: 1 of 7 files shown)


kwonoj:
current implementation of `multicast` and other `publish...` does not have `selector` parameter implementation in RxJS4 (https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/multicast.md#rxobservableprototypemulticastsubject--subjectselector-selector), this PR also doesn't have it yet. (not sure if `selector` is dropped or not)

staltz:
Very good question about `selector`. I know Erik Meijer uses it idiomatically. What does @blesh @trxcllnt think?

staltz:
Otherwise, the PR looks perfect.

blesh:
The selector in RxJS 4 is meant to make shared things semi reusable by hiding everything you're doing with it in a closure that you can repeat. This isn't _really_ necessary now that you can make reusable connectable observables.

When I had asked Erik about it a long time ago, he told me that he only uses `publish()` with connect for the most part. Overall, I'm indifferent other than I don't want to introduce it if it's going to result in heavily polymorphic or slow implementations of the operators.

kwonoj:
Since I personally does not have strong opinions of selector, I will not include those changes in this PR and would like to see if there's demand for those, such as someone might filing issue allows further discussion if necessary.

kwonoj:
PR is updated with rebase and add corresponding changes of `add` auto patching.

PR #886, opened 2015-12-04
catch: fixes and tests related to Subscriber chains

staltz:
Fix Catch() operator to include CatchSubscriber in the destination Subscriber, in order not to break
unsubscription chains. Add test that catch() should not break unsubscription chain.

@trxcllnt and @blesh continuing discussions in #840, see this. I did not use `super(destination)` but I made sure the Subscriber chain is preserved. In other words, it solves the bug, perhaps not the way you wanted. I don't know how to solve it with `super(destination)`.

Here's some reasoning of the problems I found with `super(destination)` **specific to catch**:

We need to support this:
```js
  it('should catch error and replace it with a cold Observable', function () {
    var e1 =   hot('--a--b--#          ');
    var e1subs =   '^       !          ';
    var e2 =  cold(        '1-2-3-4-5-|');
    var e2subs =   '        ^         !';
    var expected = '--a--b--1-2-3-4-5-|';

    var result = e1.catch(function (err) { return e2; });

  });
```

In other words, we NEED to unsubscribe from `e1` when it errors.

screen shot 2015-12-04 at 11 55 31

Because `e1`'s Subscription is added to the shared `destination` at the bottom, it would **only** be unsubscribed when the `destination` is unsubscribed. That's **not** what we want. We need to unsubscribe from `e1` when it errors, without unsubscribing from `destination`, which in turn should happen only when `destination` see's `e2`'s `complete`.

Makes sense?

### DIFF
--- a/spec/operators/catch-spec.js
+++ b/spec/operators/catch-spec.js
@@ -24,7 +24,7 @@ describe('Observable.prototype.catch()', function () {
   });
 
   it('should catch error and replace it with a cold Observable', function () {
-    var e1 =   hot('--a--b--#----|     ');
+    var e1 =   hot('--a--b--#          ');
     var e1subs =   '^       !          ';
     var e2 =  cold(        '1-2-3-4-5-|');
     var e2subs =   '        ^         !';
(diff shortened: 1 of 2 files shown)


blesh:
This looks good to me, @staltz. I don't see anything wrong with this approach.

I think this should work because there's no way the CatchSubscriber could have a `destination` that was't a Subscriber. This is because all non-subscriber or non-subject observers are wrapped in a subscriber when `subscribe` is called on the Observable.  If that weren't true, there might be a problem here, but that's never going to be true, so we're good.

blesh:
Merged with e1447ac... thanks for this one @staltz! Some of the subscription chain stuff is a mind-bender at times.

PR #890, opened 2015-12-04
chore(package.json): bump up jasmine version to latest

kwonoj:
This PR bump up jasmine to latest stable, 2.4.1.

I do not think this is direct resolution of intermittent test failure issues, but no harm to have some other fixes in place.


kwonoj:
Merged with 7db4964.

PR #896, opened 2015-12-05
docs(README): Update markdown-doctest config and fix import typo

Widdershin:
In order to avoid having to update the config for this style of operator in future, I think I will add the ability to specify a fallback require function for `markdown-doctest`, so that we can handle this automatically.

Happy to note that it did catch a legitimate error :smile:

### DIFF
--- a/.markdown-doctest-setup.js
+++ b/.markdown-doctest-setup.js
@@ -17,7 +17,11 @@ it.asDiagram = function asDiagram() {
 
 module.exports = {
   require: {
-    '@reactivex/rxjs': Rx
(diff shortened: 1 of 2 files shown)


kwonoj:
I noticed travis failure and just about to start look into this, thanks :)

Widdershin:
Thanks @kwonoj :smile:

kwonoj:
Merged with b3a227d, hope now build failure to be recovered. Appreciate for quick triage, @Widdershin :)

kwonoj:
And seems travis is back to green. :)

PR #897, opened 2015-12-05
refactor(operators): expand usage of const over let for variables

kwonoj:
- replaces variable declared by let to const where applicable
- use constructor parameter where applicable
- minor whitespace, lint, type update


luisgabriel:
:+1:

kwonoj:
Merged with 0bffca6.

PR #900, opened 2015-12-05
fix(repeat): preserve Subscriber chain in repeat()

staltz:
Fix repeat() operator to add its own Subscriber to the destination Subscriber, in order to avoid


blesh:
Merged with d9a7328... thanks @staltz

PR #901, opened 2015-12-05
fix(retry): preserve Subscriber chain in retry()

staltz:
Fix retry() operator to add its own Subscriber to the destination Subscriber, in order to avoid


blesh:
Merged with b429dac. Thanks @staltz !

PR #902, opened 2015-12-05
fix(retryWhen): preserve Subscriber chain in retryWhen()

staltz:
Fix retryWhen() operator to add its own Subscriber to the destination Subscriber, in order to avoid


blesh:
merged with c9cb958. Thanks @staltz

Issue #875, opened 2015-12-03
Tests asserting unsubscription chains don't break 


staltz:
We need to add tests that each operator propagates unsubscription from the destination to the source.

Here is an example:
```js
  it('should not break unsubscription chain when unsubscribed explicitly', function () {
    var e1 =     hot('--a--b--c--d--e----|');
    var unsub =      '         !          ';
    var e1subs =     '^        !          ';
    var skip =   hot('-------------x--|   ');
    var skipSubs =   '^        !          ';
    var expected =  ('----------          ');

    var result = e1
      .mergeMap(function (x) { return Observable.of(x); })
      .skipUntil(skip)
      .mergeMap(function (x) { return Observable.of(x); });

    expectObservable(result, unsub).toBe(expected);
    expectSubscriptions(e1.subscriptions).toBe(e1subs);
    expectSubscriptions(skip.subscriptions).toBe(skipSubs);
  });
```

Notice the use of the `mergeMap` sandwich there.

- - - -

- [x] buffer
- [x] bufferCount
- [x] bufferTime
- [x] bufferToggle
- [x] bufferWhen
- [x] catch
- [x] combineAll
- [x] combineLatest
- [x] concat-static
- [x] concat
- [x] concatAll
- [x] concatMap
- [x] concatMapTo
- [x] count
- [x] debounce
- [x] debounceTime
- [x] defaultIfEmpty
- [x] delay
- [x] dematerialize
- [x] distinctUntilChanged
- [x] distinctUntilKeyChanged
- [x] do
- [x] elementAt
- [x] every
- [x] exhaust
- [x] exhaustMap
- [x] expand
- [x] filter
- [x] find
- [x] findIndex
- [x] first
- [x] groupBy
- [x] ignoreElements
- [x] isEmpty
- [x] last
- [x] map
- [x] mapTo
- [x] materialize
- [x] max
- [x] merge-static
- [x] merge
- [x] mergeAll
- [x] mergeMap
- [x] mergeMapTo
- [x] mergeScan
- [x] min
- [x] multicast
- [x] observeOn
- [x] partition
- [x] publish
- [x] publishBehavior
- [x] publishLast
- [x] publishReplay
- [x] reduce
- [x] repeat
- [x] retry
- [x] retryWhen
- [x] sample
- [x] sampleTime
- [x] scan
- [x] share
- [x] single
- [x] skip
- [x] skipUntil
- [x] skipWhile
- [x] startWith
- [x] subscribeOn
[...]

PR #903, opened 2015-12-05
test(groupBy): add test against breaking unsubscription chain

staltz:
Add test for groupBy() operator, to verify that unsubscription chains are not broken when the
observable is unsubscribed explicitly.

For issue #875.


blesh:
Merged with 078f124. Thanks @staltz !

PR #904, opened 2015-12-05
test(share): add test against breaking unsubscription chain

staltz:
Add test for share() operator, to verify that unsubscription chains are not broken when the last
subscriber unsubscribes.

For issue #875.


kwonoj:
After PR https://github.com/ReactiveX/RxJS/pull/867, lint try to bug changes as much as they can :)

Otherwise, change looks good to me.

staltz:
Ok thanks, PR updated

blesh:
Merged with 5db1352. Thanks @staltz !

PR #905, opened 2015-12-05
fix(bufferToggle): fix unsubscriptions of closing Observable

staltz:
Fix bufferToggle to unsubscribe from the closing Observables once the bufferToggle Observable is
unsubscribed.


blesh:
Ooo.. good catch. :)

blesh:
Merged with 439b641. Thanks @staltz!

PR #906, opened 2015-12-05
docs(README): Fix README imports, simplify md-doctest config

Widdershin:
@blesh it looks like the way things are currently set up, you have to do `import Rx from 'rxjs/Rx';` instead of `import {Rx} from 'rxjs/Rx';`. I'm unsure which you were aiming for, but I've updated the README to reflect the current state of affairs.

I've also added support to `markdown-doctest` for the `require('rxjs/operator/map')` style of require, and hopefully any operator should be requirable without additional configuration.

### DIFF
--- a/.markdown-doctest-setup.js
+++ b/.markdown-doctest-setup.js
@@ -17,11 +17,7 @@ it.asDiagram = function asDiagram() {
 
 module.exports = {
   require: {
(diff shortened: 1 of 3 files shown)


blesh:
Thanks. I added that in a rush

blesh:
Merged with b28149e. Thanks @Widdershin !

PR #908, opened 2015-12-07
chore(package.json): bump esdoc version to latest (0.4.3)

Widdershin:
Currently documentation generation appears to be broken because of esdoc/esdoc#73, which was fixed in the 0.2.6 release.

I have verified that it works correctly using 0.4.3.


blesh:
Merged with e44b2d2. Thanks @Widdershin

Issue #899, opened 2015-12-05
Circular reference between Observable and others


kwonoj:
*Observable.ts*

```
...
import {ConnectableObservable} from './observable/ConnectableObservable';
import {Subject} from './Subject'; --> import subject

...
if (observerOrNext instanceof Subscriber || observerOrNext instanceof Subject) {
```

*Subject.ts*
```
...
import {Observer} from './Observer';
import {Observable} from './Observable'; --> import observable
...
export class Subject<T> extends Observable<T>
...
```

ends up with below snippet

```
import {Observable} from 'dist/cjs/Observable';

Observable.of(1).subscribe(console.log);
```

```
 d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
                                                                   ^
TypeError: Cannot read property 'prototype' of undefined
    at __extends (D:\github\RxJS\dist\cjs\Subject.js:4:68)
    at D:\github\RxJS\dist\cjs\Subject.js:20:5
    at Object.<anonymous> (D:\github\RxJS\dist\cjs\Subject.js:135:3)
```

Subject does not resolves reference to Observable.

kwonoj:
Not sure if https://github.com/ReactiveX/RxJS/issues/894 is related or not..

blesh:
It's probably #894. TypeScript is pretty good about hoisting dependencies that would otherwise be circular.

`import {Observable} from 'dist/cjs/Observable';` will just get you the bare-bones Observable without `from` and `of` right now.

kwonoj:
Did couple of try, seems #894 and this need to be updated both. even import `from` / `of` to observable, importing observable directly would not work due to this.

kwonoj:
Changed subject to describe this more general, such as

- Observable <--> GroupedObservable
- Observable <--> CoreOperators <--> GroupedObservable ..(direction is not precise, just an example)

and lot of others.

This reference could possibly cause runtime reference resolve issue as described above and also noticeably increases build times.

Tried to solve this via simple ambient declaration but wasn't work as expected, ambient merging seems possible way but still need to prove.

ntilwalli:
@kwonoj Just a quick comment, dunno if it's relevant, I introduced this circular import with https://github.com/ReactiveX/RxJS/pull/863 and to deal with it I updated the KitchenSink and Rx files to move the Subject import before Observable. I made a comment to that effect in the discussion and in the checkin notes.  Until I reordered these imports, I was seeing the same error.

[...]

Issue #894, opened 2015-12-05
Observable.of and Observable.from must be on Observable


blesh:
Right now the Observable exported by `src/Observable` doesn't have `of` and `from` on it by default. It really should.

`from` is required for interop at the very least.

blesh:
cc/ @robwormald @jeffbcross

jeffbcross:
whatev

kwonoj:
Closing this issue per discussion of #922, importing into Observable creates circular dependencies. For now will require manual import until better solution comes in.

PR #909, opened 2015-12-07
Proposal : remove circular reference using ambient merge

kwonoj:
### Proposal : not to be merged yet

This PR is proposal to resolve https://github.com/ReactiveX/RxJS/issues/899, via ambient declaration merging introduced in typescript (https://github.com/Microsoft/TypeScript/issues/3332).

In this PR, for any parents requires to have declaration of inherited childs, it imports ambient declarations instead of actual implementation of child to remove circular dependencies. Since reference to child from parents is mostly for type declaration only, this doesn't affect actual functionality in general.

Still for some cases functionality's breaking, such as this PR regresses https://github.com/ReactiveX/RxJS/pull/863 due to failed to check type of `Subject` since ambient declaration doesn't allow type checking. Yet this is proposal, PR does not include those refactoring but only include necessary changes.

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -6,10 +6,11 @@ import {Subscription} from './Subscription';
 import {root} from './util/root';
 import {CoreOperators} from './CoreOperators';
 import {SymbolShim} from './util/SymbolShim';
-import {GroupedObservable} from './operator/groupBy-support';
-import {ConnectableObservable} from './observable/ConnectableObservable';
-import {Subject} from './Subject';
 import {Notification} from './Notification';
+
+import {Subject} from './ambient/Subject';
+import {GroupedObservable} from './ambient/GroupedObservable';
+import {ConnectableObservable} from './ambient/ConnectableObservable';
 
 /**
  * A representation of any set of values over any amount of time. This the most basic building block
@@ -89,7 +90,7 @@ export class Observable<T> implements CoreOperators<T>  {
     let subscriber: Subscriber<T>;
 
     if (observerOrNext && typeof observerOrNext === 'object') {
-      if (observerOrNext instanceof Subscriber || observerOrNext instanceof Subject) {
+      if (observerOrNext instanceof Subscriber) { //|| observerOrNext instanceof Subject) {
         subscriber = (<Subscriber<T>> observerOrNext);
       } else {
         subscriber = new Subscriber(<Observer<T>> observerOrNext);
@@ -255,3 +256,9 @@ export class Observable<T> implements CoreOperators<T>  {
   zip: <R>(...observables: Array<Observable<any> | ((...values: Array<any>) => R)>) => Observable<R>;
   zipAll: <R>(project?: (...values: Array<any>) => R) => Observable<R>;
 }
+
+//let import occurs after Observable<T> is exported
+import {ArrayObservable} from './observable/fromArray';
+
+Observable.fromArray = ArrayObservable.create;
+Observable.of = ArrayObservable.of;
--- a/src/ambient/ConnectableObservable.ts
+++ b/src/ambient/ConnectableObservable.ts
@@ -1,0 +1,1 @@
+export interface ConnectableObservable<T> {}
--- a/src/CoreOperators.ts
+++ b/src/CoreOperators.ts
@@ -1,9 +1,10 @@
-import {Observable} from './Observable';
 import {Scheduler} from './Scheduler';
-import {ConnectableObservable} from './observable/ConnectableObservable';
-import {Subject} from './Subject';
-import {GroupedObservable} from './operator/groupBy-support';
 import {Notification} from './Notification';
+
+import {Observable} from './ambient/Observable';
+import {Subject} from './ambient/Subject';
+import {GroupedObservable} from './ambient/GroupedObservable';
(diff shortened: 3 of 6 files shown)


src/Observable.ts
```diff
@@ -89,7 +90,7 @@ export class Observable<T> implements CoreOperators<T>  {
     let subscriber: Subscriber<T>;
 
     if (observerOrNext && typeof observerOrNext === 'object') {
-      if (observerOrNext instanceof Subscriber || observerOrNext instanceof Subject) {
+      if (observerOrNext instanceof Subscriber) { //|| observerOrNext instanceof Subject) {
```
kwonoj: This is part where regresses https://github.com/ReactiveX/RxJS/pull/863, ambient declaration of `Subject` cannot check its type via `instanceof` since it's interface.

src/Observable.ts
```diff
@@ -255,3 +256,9 @@ export class Observable<T> implements CoreOperators<T>  {
   zip: <R>(...observables: Array<Observable<any> | ((...values: Array<any>) => R)>) => Observable<R>;
   zipAll: <R>(project?: (...values: Array<any>) => R) => Observable<R>;
 }
+
+//let import occurs after Observable<T> is exported
+import {ArrayObservable} from './observable/fromArray';
```
kwonoj: This resolves https://github.com/ReactiveX/RxJS/issues/894, but import occurs after `Observable` is actually declared to let `ArrayObservable` able to reference `Observable`.

kwonoj:
Due to regression, this PR does not passes travis and it's expected.

src/ambient/ConnectableObservable.ts
```diff
@@ -0,0 +1 @@
+export interface ConnectableObservable<T> {}
```
kwonoj: maybe this declarations could be single module such as

*ambient.ts*
```
export interface A{}
export interface B{}
export interface C{}
```
to allow single import to declarations like

`import {A,B,C} from 'ambient'`

kwonoj:
Note : this change has side effect of reducing build time speed. On travis, numbers are below (may vary per run though)

*without PR(cjs)*
```
Files:             280
Lines:           27854
Nodes:          128684
Identifiers:     42695
Symbols:       8453476
Types:          477658
Memory used:   766295K
I/O read:        0.02s
I/O write:       0.07s
Parse time:      0.55s
Bind time:       0.33s
Check time:     14.64s
Emit time:       1.00s
Total time:     16.52s
```

*with PR(cjs)*
```
Files:             284
Lines:           27853
Nodes:          128728
Identifiers:     42705
Symbols:       8038078
Types:          193698
Memory used:   282858K
I/O read:        0.03s
I/O write:       0.06s
Parse time:      0.81s
Bind time:       0.60s
Check time:      8.22s
Emit time:       0.92s
Total time:     10.56s
```

src/Observable.ts
jeffbcross: How about 

```typescript
if (observerOrNext instanceof Subscriber || 
    (typeof (<Subject>observerOrNext).next === 'function' && typeof (<Subject>observerOrNext).subscribe === 'function')) {

```

jeffbcross: Only reason for the explicit `<Subject>` is to make the code tool-friendly.

src/ambient/ConnectableObservable.ts
jeffbcross: +1

src/Observable.ts
kwonoj: Yes, need to cast `<any>` instead of `Subject` (since compile time imported ambient declaration does not have interfaces) but this works, thanks for pointing out. I was just spending time in opposite way of bringing interface check availability to runtime..

kwonoj:
I'll leave this PR as is for now to get opinions of direction itself, once it's agreed to go will update PR accordingly.

blesh:
@kwonoj, I like what you're doing with the `ambient` interface declarations here. We can probably use that. I have an alternative PR for fixing the Subject as a subscriber issue here #912. We can probably merge both of these PRs if you want to remove the changes you made to `subscribe` here.

kwonoj:
@blesh Just saw PR for `rxSubscriber` and seems that resolves one of failings to deal with Subject above gracefully. Once PR #912 is merged, I'll update this PR accordingly by utilizing those changes.

src/ambient/ConnectableObservable.ts
blesh: I agree, make these all into one file.

src/Observable.ts
blesh: I have an alternative PR for this particular bit here #912

kwonoj:
Seems #912 solves circular reference and this PR might not have much values to be added except few side effect like reduce build time. Once got confirmation from @blesh , I'll close this without merge.

blesh:
Yeah, I think we can probably close this one for now, since we no longer have the circular reference issue. I really appreciate the thought and hard work here though. It's an idea I'll have to remember if we do run into the circular reference issue again.

blesh:
Thanks again, @kwonoj

kwonoj:
It is cool to have changes landed to resolve issue. Thanks for taking time to reviewing this, @blesh :)

PR #910, opened 2015-12-07
test(map/mapTo): add test against breaking unsubscription chain 

luisgabriel:
Related to #875.


blesh:
Merged with 9e702fe. Thanks again @luisgabriel

PR #911, opened 2015-12-07
test(filter): add test against breaking unsubscription chain

luisgabriel:
Relates to #875.


blesh:
Merged with dedbfc8. Thanks again @luisgabriel

PR #912, opened 2015-12-07
Add rxSubscriber symbol

blesh:
Add an `rxSubscriber` symbol that can be used to designate any object as "safe" to use an an Rx Subscriber (sort of)...

The idea is basically this: If any object has `Rx.Symbol.rxSubscriber` implemented as a function, that function is expected to return an `Rx.Subscriber`. This allows Rx to shortcut the wrapping of observers (such as Subjects and the like) by calling this method and using what it returns instead.

It seems like a more flexible solution for this scenario than other solutions.

Other work done:

- Refactored SymbolShim (almost completely) to include a `for` method on the polyfill.
- Added `rxSubscriber` to both `Rx` and `Rx.KitchenSink` output
- Tests

### DIFF
--- a/spec/util/SymbolShim-spec.js
+++ b/spec/util/SymbolShim-spec.js
@@ -1,23 +1,28 @@
 /* globals __root__ */
-var SymbolDefinition = require('../../dist/cjs/util/SymbolShim').SymbolDefinition;
+var SymbolShim = require('../../dist/cjs/util/SymbolShim');
 var Map = require('../../dist/cjs/util/Map').Map;
 var Rx = require('../../dist/cjs/Rx');
+var polyfillSymbol = SymbolShim.polyfillSymbol;
+var ensureIterator = SymbolShim.ensureIterator;
 
-describe('SymbolDefinition', function () {
+describe('SymbolShim', function () {
   it('should setup symbol if root does not have it', function () {
     var root = {};
 
-    var result = new SymbolDefinition(root);
+    var result = polyfillSymbol(root);
     expect(root.Symbol).toBeDefined();
     expect(result.observable).toBeDefined();
     expect(result.iterator).toBeDefined();
+    expect(result.for).toBeDefined();
   });
 
-  it('should have observable, iterator symbols', function () {
-    var result = new SymbolDefinition(__root__);
+  it('should add a for method', function () {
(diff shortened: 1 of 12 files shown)


spec/util/SymbolShim-spec.js
```diff
@@ -43,60 +48,54 @@ describe('SymbolDefinition', function () {
...
-    it('should patch root if for symbol does not exist', function () {
-      var root = {};
-
-      var result = new SymbolDefinition(root);
-      expect(result.iterator).toBe('@@iterator');
-    });
+  it('should patch using Set for mozilla bug', function () {
+    function Set() {
+    }
+    Set.prototype['@@iterator'] = function () {};
```
blesh: Changed this around to use an actual class.

src/Observable.ts
```diff
@@ -89,8 +90,10 @@ export class Observable<T> implements CoreOperators<T>  {
...
 
     if (observerOrNext && typeof observerOrNext === 'object') {
-      if (observerOrNext instanceof Subscriber || observerOrNext instanceof Subject) {
+      if (observerOrNext instanceof Subscriber) {
         subscriber = (<Subscriber<T>> observerOrNext);
+      } else if (observerOrNext[rxSubscriber]) {
+        subscriber = observerOrNext[rxSubscriber]();
       } else {
         subscriber = new Subscriber(<Observer<T>> observerOrNext);
       }
```
blesh: The magic happens above this line.

jeffbcross: :+1:

jeffbcross:
LGTM (I mostly just looked at the `instanceof` and `Symbol` fix)

jeffbcross:
I ran made before and after this change to verify the circular dependency is gone.

jeffbcross:
See [this pr](https://github.com/ReactiveX/RxJS/pull/917) to see how I checked.

kwonoj:
Changes look good to me, :)

staltz:
Uuuh, interesting PR and interesting alternative to multiple inheritance.

kwonoj:
@staltz it is interesting, isn't it? I was spending time to find way via sort of inheritance chain and @blesh brings this beauty :)

Issue #838, opened 2015-11-30
Scheduler name bike-shedding


blesh:
Currently the trampoline scheduler is named `immediate`, and the microtask scheduler is named `nextTick`. There is als

I propose that we rename `immediate` to `sync`. I say this because `immediate` sort of implies that the scheduled action is going to take place right away, and that's not *really* true. It's going to be bumped from it's current call stack and executed next, but synchronously.

For `nextTick`, it's a little more debatable, it's pretty aptly named, but I think it would be more descriptive if it were named `microtask`, because ultimately that's the goal. To schedule it as a microtask after the current frame but before things like `setTimeout`.

... which brings me to another thing, we don't have a scheduler based off of `setTimeout` or `requestAnimationFrame`. The latter is DOM-specific, but given that we're modular now, I think there's grounds for including it in this library, it's an important scheduler.

Once we land on these names, I'll be writing some improved documentation on the matter.

staltz:
I see no objections to this...

blesh:
In talking to @chriscareycode, @abersnaze and @stealthcode at Netflix, there were a few other options presented for names around `immediate` in particular:

- `queueImmediate`: because it gives users the sense that scheduled items are being buffered in some way.
- `trampoline`: because it's a brand new word with no connotations in the JS community which may make it more memorable, and describes what it does in a way that can be googled.
- `jenga`: because it topples the call stack over with in the same event on the event loop. Again, descriptive and memorable.
- `immediate`: less favored because people have some idea in their heads of what "immediate" means.
- `sync`: Because you're scheduling, but it's synchronous. This was less well recieved.

... all parties thought that `currentThread` didn't make any sense to them because of JavaScript's lack of "real" multithreading.

staltz:
Trampoline sounds the best IMO.

kwonoj:
+1 for `trampoline` and 'currentThread didn't make sense ' both. (intrigued into 'jenga' for sec..)

blesh:
I'd be interested in thoughts from @zenparsing, @domenic, @wycats or @stefanpenner on this, as well. I'm sure this sort of scheduling has come up in conversations at the TC39, and they might have more specific vernacular we could leverage.

mattpodwysocki:
@blesh that's been the problems with transporting the names from runtime to runtime.  CurrentThread has a history of doing recursive scheduling right, placing things in the right order whereas the immediate was indeed immediate.  

Consider the difference:
```js
[...]

PR #914, opened 2015-12-07
Scheduler rename

blesh:
This is just a straight renaming, per discussions on #838 

The result is as follows:

|RxJS 4| RxJS 5|
|---|---|
|currentThread|queue|
|default|asap|

Additional schedulers are still to be implemented.

### DIFF
--- a/spec/observables/IteratorObservable-spec.js
+++ b/spec/observables/IteratorObservable-spec.js
@@ -88,7 +88,7 @@ describe('IteratorObservable', function () {
       [10, 20, 30, 40],
       function (x) { return x; },
       null,
-      Rx.Scheduler.immediate
+      Rx.Scheduler.queue
     );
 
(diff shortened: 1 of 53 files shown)


blesh:
I know some people weren't enthused about the `asap` name. But I settled on it, in lieu of variants of names with "job" in it (per @zenparsing and @jhusain telling me that the TC39 refers to micro task queues as "jobs")... this is because _googling "RxJS job scheduler" or "RxJS job" or the like would yield horrendous results.

On the other hand, googling "javascript asap" or the like, yields extremely relevant results even today.

staltz:
D:

blesh:
Naming these things sucks. haha. I don't expect everyone to be happy with the choices. Hell, _I'm_ not even happy with them, completely.

Just need to make a choice and keep moving.

kwonoj:
:+1:

PR #917, opened 2015-12-07
test(build): add test to ensure no circular dependencies in CJS output

jeffbcross:
The line in travis yaml should be uncommented when ready to run this check


blesh:
This has been merged with 0f0dbe8. I've uncommented the circular dependency checking for travis.

Thanks @jeffbcross!

PR #918, opened 2015-12-07
test(take): add test against breaking unsubscription chain

figueredo:
Relates to #875.


blesh:
merged with b1ae119. Thanks again, @figueredo

PR #919, opened 2015-12-07
test(takeWhile): add test against breaking unsubscription chain

figueredo:
Relates to #875.


blesh:
Merged with d7438b5. Thanks again, @figueredo !

PR #920, opened 2015-12-07
test(takeUntil): add test against breaking unsubscription chain

figueredo:
Relates to #875.


blesh:
Merged with c9da8cb. Thanks yet again, @figueredo!

PR #922, opened 2015-12-08
feat(Observable): export of, from by default

### LINKED ISSUE #894: Observable.of and Observable.from must be on Observable

kwonoj:
closes #894

Now circular reference issue is resolved in TOT, this changes will allow import Observable only makes to work with `of` and `from`.

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -258,3 +258,12 @@ export class Observable<T> implements CoreOperators<T>  {
   zip: <R>(...observables: Array<Observable<any> | ((...values: Array<any>) => R)>) => Observable<R>;
   zipAll: <R>(project?: (...values: Array<any>) => R) => Observable<R>;
 }
+
+//import after Observable<T> is declared and exported
+import {ArrayObservable} from './observable/fromArray';
+
+Observable.fromArray = ArrayObservable.create;
+Observable.of = ArrayObservable.of;
+
+import {FromObservable} from './observable/from';
+Observable.from = FromObservable.create;
--- a/src/Rx.KitchenSink.ts
+++ b/src/Rx.KitchenSink.ts
@@ -29,8 +29,6 @@ import './add/observable/bindCallback';
 import './add/observable/defer';
 import './add/observable/empty';
 import './add/observable/forkJoin';
-import './add/observable/from';
-import './add/observable/fromArray';
 import './add/observable/fromEvent';
 import './add/observable/fromEventPattern';
 import './add/observable/fromPromise';
--- a/src/Rx.ts
+++ b/src/Rx.ts
@@ -15,8 +15,6 @@ import './add/observable/bindCallback';
 import './add/observable/defer';
 import './add/observable/empty';
 import './add/observable/forkJoin';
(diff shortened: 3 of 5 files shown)


src/Rx.KitchenSink.ts
```diff
@@ -30,7 +30,6 @@ import './add/observable/defer';
 import './add/observable/empty';
 import './add/observable/forkJoin';
 import './add/observable/from';
-import './add/observable/fromArray';
```
kwonoj: doesn't need to re import again since `Observable` always patches this.

src/Observable.ts
```diff
@@ -258,3 +258,9 @@ export class Observable<T> implements CoreOperators<T>  {
   zip: <R>(...observables: Array<Observable<any> | ((...values: Array<any>) => R)>) => Observable<R>;
   zipAll: <R>(project?: (...values: Array<any>) => R) => Observable<R>;
 }
+
+//import after Observable<T> is declared and exported
+import {ArrayObservable} from './observable/fromArray';
+
+Observable.fromArray = ArrayObservable.create;
```
blesh: We actually only need `from` and `of`. But `fromArray` is fine, since it's part of the same module anyhow.

kwonoj: You're correct, why I did import `fromArray` only except `from`? :( Since it's same module I also think leaving `fromArray` would be fine, updating to include `from` actually..

kwonoj:
Updated PR to patch Observable to include

`from`
`of`
`fromArray` (which based on same module with `of`, importing it at once)

blesh:
@kwonoj unfortunately, this introduces circular dependencies to the file.

```
Observable -> observable/from
Observable -> observable/from -> observable/IteratorObservable
Observable -> observable/from -> observable/fromArray
Observable -> observable/from -> observable/fromArray -> observable/ScalarObservable
Observable -> observable/from -> observable/fromArray -> observable/ScalarObservable -> observable/empty
Observable -> observable/from -> observable/fromArray -> observable/ScalarObservable -> observable/throw
Observable -> observable/from -> observable/fromPromise
Notification -> Observable -> observable/from -> operator/observeOn-support
```

Honestly, at this point if we just rename `observable/fromArray` to `observable/of`, I think I'm okay just leaving these things out for now. We'll see how it works out.

I _am_ minorly worried about `from` though, since it's a requirement for interop.

kwonoj:
Oops, that's bad.. 

let me rephrase to see if I understood correctly - 
```
Observable.from = ArrayObservable.create;
Observable.of = ArrayObservable.of;
```

and does not import `FromObservable` into `Observable`. is that correct?

blesh:
@kwonoj I think this approach just won't work. 

Honestly, I'm okay with that for now. I'm cool with the modular approach of needing to import `of` and `from`.

kwonoj:
OK, understood. I'd rather avoid of dependency issue than import too. Closing this PR.

PR #927, opened 2015-12-08
style(test): update test helper lint-friendly

kwonoj:
This PR updates `painter.js` lint friendly.

/cc @staltz for couple of comments.

### DIFF
--- a/spec/helpers/tests2png/painter.js
+++ b/spec/helpers/tests2png/painter.js
@@ -1,3 +1,4 @@
+/*eslint-disable no-param-reassign, no-use-before-define*/
 var gm = require('gm');
 var _ = require('lodash');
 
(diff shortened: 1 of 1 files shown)


spec/helpers/tests2png/painter.js
```diff
@@ -1,3 +1,4 @@
+/*eslint-disable no-param-reassign, no-use-before-define*/
```
kwonoj: `no-use-before-define` is disabled due to recursive calling between `drawNestedObservable` and `drawObservableMessages`

`no-param-reassign` is disable due to reassign to `out` parameters.

Maybe both of them can be refactored instead?

staltz: Maybe yeah, but I tried not to spend too much time with this script to make the code elegant, because it was still a proof of concept. We can refactor and improve it over time.

staltz:
Otherwise, PR looks good.

kwonoj:
Yup, also consider those refactoring as 'good to have', can be visited later. For now, I'll suppress via lint configuration.

kwonoj:
Merged with 132836f.

Issue #878, opened 2015-12-03
Remove `thisArg` from non-nativesque operators


blesh:
Per discussion here #787, we will remove the `thisArg` arguments from all operators that aren't meant to mirror something in native JavaScript.

So really, only `map`, `filter`, `reduce` and `forEach` should have `thisArg`s.

Furthermore, the internal bind helper introduces a closure and should probably be removed in favor of using `apply` or `call` directly within those operators, or it should be otherwise refactored to not use closures.

benjamingr:
+1 I think this is important for consistency.

Just adding that the following also need a `thisArg`: [`some`](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/some.md) [`find`](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/find.md) [`findIndex`](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/findindex.md) and [`every`](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/every.md)

blesh:
Reopening until #993 is merged.

blesh:
Closed by #993

PR #928, opened 2015-12-08
feat(single): remove thisArg

luisgabriel:
Relates to #878.


luisgabriel:
Not sure if there is any native JS function similar to single(). You can close this PR if this is the case.

blesh:
Merged with 43af805... thanks @luisgabriel!

PR #929, opened 2015-12-08
test(zip): add test against breaking unsubscription chain

luisgabriel:
Relates to #875.


kwonoj:
Merged with 4f02090, thanks @luisgabriel .

PR #930, opened 2015-12-08
test(mergeMap): add tests for limited concurrency cases

staltz:
Add tests for mergeMap to assert its behavior when concurrency parameter is set to 1 and 2.


blesh:
merged with 1553524... thanks @staltz

PR #931, opened 2015-12-08
test(mergeMapTo): add tests for limited concurrency cases

staltz:
Add tests for mergeMapTo, to assert its behavior when the concurrency parameter is set to 1 and 2.


blesh:
Merged with 451a1e2... thanks @staltz

PR #932, opened 2015-12-08
test(zipAll): add test against breaking unsubscription chain

luisgabriel:
Relates to #875.


blesh:
Merged with 70266d0... thanks again @luisgabriel

PR #937, opened 2015-12-08
test(expand): add test against breaking unsubscription chain

staltz:
For issue #875.


kwonoj:
Merged with 425f44a, thanks @staltz

PR #964, opened 2015-12-08
test(reduce): add test against breaking unsubscription chain

staltz:
For issue #875.


kwonoj:
Merged with 460d510, thanks @staltz . (maybe I need merge-and-thanks to staltz script? :sunglasses: )

PR #970, opened 2015-12-08
test(skip): add test against breaking unsubscription chain

staltz:
For issue #875.


kwonoj:
Merged with d2bd860, thanks @staltz

PR #987, opened 2015-12-08
test(windowToggle): add test against breaking unsubscription chain

staltz:
For issue #875.

### DIFF
--- a/spec/operators/windowToggle-spec.js
+++ b/spec/operators/windowToggle-spec.js
@@ -117,6 +117,35 @@ describe('Observable.prototype.windowToggle', function () {
     expectSubscriptions(e1.subscriptions).toBe(subs);
   });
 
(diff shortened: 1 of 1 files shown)


kwonoj:
Couple of change seems need to be lint friendly :)

staltz:
Yes, I thought `npm test` was running lint (at some point I'm sure it was, it's not anymore), that's why these leaked.

kwonoj:
You may use `npm run build_test` instead of `test` to lint kick off. `test` itself just invokes jasmine only.

staltz:
Yeah but `build_test` is really slow since it also compiles TypeScript.

kwonoj:
I agree, still you have && operator at least for now. :) I always felt there might be better ergonomics instead of using npm script, but far, far away story.

PR #990, opened 2015-12-08
test(withLatestFrom): add unsubscription-related specs

staltz:
Add test case to verify that unsubscription chains are not broken. Add test that asserts
subscriptions when an unsubscription happens explicitly.

Related to issue #875.


blesh:
merged with 97e1fdf. Thanks @staltz

PR #991, opened 2015-12-08
docs(count): remove reference to thisArg

staltz:
Remove reference to a thisArg argument given to count() operator in its JSDoc.

For issue #878.


kwonoj:
Merged with 1298b57, thanks @staltz

PR #992, opened 2015-12-08
Updates for thisArgs issue

blesh:
- removes `thisArgs` from `skipWhile`
- adds `thisArgs` to `forEach`
- removes some `thisArgs` cruft from `CoreOperators` interface.
- eliminates some closure in the `forEach` method


blesh:
This should close #878

kwonoj:
Changes look good to me.

blesh:
Thanks for the review @kwonoj

Issue #889, opened 2015-12-04
The _helper_ `bindCallback` should probably be removed


blesh:
On operators with `thisArg`, we're using a util/helper called `bindCallback`. It has two problems:

1. it introduces a closure, which we're actively trying to avoid.
2. it's got the same name as the `bindCallback` method (formerly `fromCallback`).

Solution:

I think we should remove this everywhere and just use `apply` or `call` where appropriate.

kwonoj:
So for some of operator still has `thisArg` will directly apply context instead of using `bindCallback`?

blesh:
right... if you still need to have `thisArg`, it should just use `apply` or `call` directly.

PR #995, opened 2015-12-08
Remove bindCallback utility

blesh:
The bindCallback util file introduced closure and wasn't completely useful to us over all. Especially now that most `thisArg` arguments have been removed.

This PR removes `bindCallback` in favor of a simple `fn.call(thisArg, a, b, c)` call.

This also fixes some issues with type checking errors in the operators touched, because some of them were sort of leveraging `bindCallback` for type checking when there was an expectation of a function.


blesh:
closes #889

kwonoj:
Was trying to make changes for this, minutes behind :)

kwonoj:
Can we add one test case into `every-spec` for `should accept thisArg`? rest (map, filter..) does have it.

kwonoj:
Otherwise change looks good to me.

blesh:
> Can we add one test case into every-spec for should accept thisArg? rest (map, filter..) does have it.

Oops, good catch. I'll add that.

kwonoj:
:+1:

Issue #988, opened 2015-12-08
Symbol shim is created as an object but should be a function


wkwiatek:
Hey,

I had an issue with `zone.js` regarding checking whether `Symbol` exists in the browser or not and it turned out that `RxJS` is initializing `Symbol` as an object the real implementation resolves it as a function. This incompatibility may affect other libraries which relies on symbols (what happens with `zone.js`). 

In the library code it's probably following line responsible for that: https://github.com/ReactiveX/RxJS/blob/master/src/util/SymbolShim.ts#L12

Here is a plunker to see the problem:
http://plnkr.co/edit/hpvSjZDERVJHiLrgLd23

And the output from browser which doesn't support symbols is as follows (Win 7 / IE 11 - BrowserStack):
![bs_win7_ie_11 0 1](https://cloud.githubusercontent.com/assets/3719897/11663210/35e5c12c-9ddb-11e5-9230-02e160a0e14c.jpg)

blesh:
This is related to #999 for sure. And also correct, it needs to be a class. :\ I did that.

PR #1000, opened 2015-12-08
fix(SymbolShim): ensure for function even if Symbol already exists

### LINKED ISSUE #999: Errors with SymbolShim in some browsers
According the the Angular core Slack, there is an issue in some browsers when trying to call `Symbol.for`. Namely in Safari 8 and all versions of IE.

I haven't reproduced this yet, but I can see/understand where the issue is. Such is the danger with DIY polyfills.

You can see that [Symbol is supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) in more browsers than [Symbol.for is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for).

blesh:
In some engines, Symbol exists but does not impelment the for method.

fixes #999

### DIFF
--- a/spec/util/SymbolShim-spec.js
+++ b/spec/util/SymbolShim-spec.js
@@ -25,6 +25,18 @@ describe('SymbolShim', function () {
     expect(test).toBe('@@test');
   });
 
+  it('should add a for method even if Symbol already exists but does not have for', function () {
+    var root = {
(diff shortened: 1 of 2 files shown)


blesh:
also fixes #988

PR #993, opened 2015-12-08
refactor(KitchenSink): remove thisArg from distinctUntilChanged signature

staltz:
Remove an obsolete thisArg from the signature of distinctUntilChanged in KitchenSink file.

For issue #878.

PR #1004, opened 2015-12-09
refactor(distinctUntilKeyChanged): remote thisArg

kwonoj:
PR https://github.com/ReactiveX/RxJS/pull/993 updates interface of `distinctUntilKeyChanged`, this PR removes thisArg from implementation.

### DIFF
--- a/src/operator/extended/distinctUntilKeyChanged.ts
+++ b/src/operator/extended/distinctUntilKeyChanged.ts
@@ -1,9 +1,9 @@
 import {distinctUntilChanged} from '../distinctUntilChanged';
 
-export function distinctUntilKeyChanged<T>(key: string, compare?: (x: any, y: any) => boolean, thisArg?: any) {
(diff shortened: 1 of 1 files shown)


src/operator/extended/distinctUntilKeyChanged.ts
```diff
@@ -1,10 +1,10 @@
 import {distinctUntilChanged} from '../distinctUntilChanged';
 
-export function distinctUntilKeyChanged<T>(key: string, compare?: (x: any, y: any) => boolean, thisArg?: any) {
+export function distinctUntilKeyChanged<T>(key: string, compare?: (x: any, y: any) => boolean) {
   return distinctUntilChanged.call(this, function(x, y) {
```
kwonoj: actually this also seems bit .. to be modified. is this function need to call `distinctUntilChanged` with `this` context?

blesh:
Oops. Okay, thanks @kwonoj

blesh:
Merged with 1979d5f. Good catch, @kwonoj

PR #1006, opened 2015-12-09
refactor(operators): remove redundant observable patch

kwonoj:
- remove observable patch does not have exposed interface and not being
  referenced

This PR removes 3 specific patch imports, `IteratorObservable`, `ScalarObservable`, `SubscribeOnObservable`.

Theses are not referenced at all, also `Observable` does not exposes interfaces build will break if this are actually referenced.


kwonoj:
Still this might have possible usecases I couldn't think of, I'll not check in until have confirmation.

blesh:
LGTM.

kwonoj:
Merged with cd217a0. thanks for confirmation, @blesh

Issue #998, opened 2015-12-08
Run tests in all major browsers and engines


blesh:
This needs to be run against:

- IE 9+,
- Chrome
- FF
- Safari
- Android
- iOS
- Node > 0.12

in an automated way. Currently we're only running against Node, and that's not good.

luisgabriel:
The idea here is to use a CI like [saucelabs](http://saucelabs.com)?

kwonoj:
Yes it is. I was trying this for some time but wanted to land https://github.com/ReactiveX/RxJS/pull/826 this before, so hold for a while. Requires two steps I assume,  

- setup test environment using runner like `karma` can do browser test with existing spec
- setup travis pipeline to saucelab to run node test on travis as ci process, rest of browser test being pipelined to saucelabs.

kwonoj:
I'll try to spend couple of days to see if I can make land `karma` into current codebase quickly. That can be starting point.

kwonoj:
Seems https://github.com/ReactiveX/RxJS/issues/733 is related.

kwonoj:
As referenced in PR #1023 , it's required to create official RxJS saucelabs account (https://saucelabs.com/) and configure travis environment values. By saucelabs account requires email account when creating account and it should not be my personal one, asking @blesh to create one.

kwonoj:
For some reason I thought email address was permanent login credential which is not actually and can be changed easily. I'm creating account by my own, change those information accordingly later to close out this issue.

kwonoj:
#1187 is decoration change, so will close this issue by #1199 which now expands browser matrix. Each test failures will be tracked separately instead of using this broad issue.

PR #1007, opened 2015-12-09
chore(test): setup karma test runner against chrome browser

kwonoj:
- initial bootstrap config for karma runner against chrome
- ci server is not configured yet

relates to #998 

This PR is very initial step to resolve issue, setting up karma configuration against chrome browser. Once this PR is merged, will expand browser support as well as CI server setup.


spec/helpers/test-helper.js
```diff
@@ -1,6 +1,6 @@
 //Fail timeouts faster
 //Individual suites/specs should specify longer timeouts if needed.
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 1000;
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 5000;
```
kwonoj: timeout interval is increased to safely executed under browser.

karma.conf.js
```diff
@@ -0,0 +1,62 @@
...
+    // list of files / patterns to load in the browser
+    files: [
+      'spec/helpers/marble-testing.js',
+      'spec/helpers/test-helper.js',
+      'spec/**/*-spec.js'
+    ],
+
+    // list of files to exclude
+    exclude: [
+      'spec/observables/from-promise-spec.js' //'should globally throw unhandled errors' fails
```
kwonoj: as commented, one of test under `fromPromise 'should globally throw unhandled errors'` is failing under browser environment. For now this test case is excluded to establish basic running environment without test failure.

karma.conf.js
```diff
@@ -0,0 +1,62 @@
...
+
+module.exports = function (config) {
+  config.set({
+
+    // base path that will be used to resolve all patterns (eg. files, exclude)
+    basePath: '',
+
+    // frameworks to use
+    // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
+    frameworks: ['browserify', 'jasmine'],
```
kwonoj: to resolve commonjs lookup, using `karma--browserify` to bundling modules. there is `karma-commonjs` does job without bundling, but has some issues of resolving node_modules (https://github.com/karma-runner/karma-commonjs/issues/29) blocks adaptation.

karma.conf.js
staltz: I predicted that test would bring some headaches (I wrote that test in order to increase test coverage). One option is we can move just that test to `spec/observables/fromPromise-node-spec.js`.

karma.conf.js
```diff
@@ -0,0 +1,62 @@
...
+    // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
+    browsers: ['Chrome'],
+
+    // Continuous Integration mode
+    // if true, Karma captures browsers, runs the tests and exits
+    singleRun: true,
+
+    // Concurrency level
+    // how many browser should be started simultanous
+    concurrency: Infinity
```
staltz: I would default to `concurrency: 1`, because I have some experience seeing browsers failing their tests just because they are not the currently focused window, and we can afford waiting a bit longer on CI for all tests to complete.

karma.conf.js
kwonoj: I agree there'll be possible different ergonomics. For now I'd like to establish whole testing pipeline with non-invasive to current codebases as much, then think about those as following changes.

karma.conf.js
kwonoj: Think good suggestion. Simply followed default values and oversighted it.

karma.conf.js
blesh: I think it's acceptable to make conditional tests:

```js
if (process) {
  it('should do process tests', function () { });
} else if (window) {
  it('should do window.onerror test', function () { });
}
```

kwonoj: OK, will try and update PR accordingly.

spec/observables/from-promise-spec.js
```diff
@@ -109,24 +112,26 @@ describe('Observable.fromPromise', function () {
...
-    var invoked = false;
-    process.on('uncaughtException', function (reason, p) {
-      if (invoked) {
-        return;
-      }
-      invoked = true;
-      expect(reason).toBe('fail');
-      done();
-    });
+  if (isNode) {
```
kwonoj: Changed to use simple if conditional for now. Thought about adding some function like `it.node` or `it.browser` to be executed under certain runtime, but it could be added in further depends on how many test we'll have per different runtime.

spec/observables/from-promise-spec.js
```diff
@@ -1,6 +1,9 @@
 /* globals describe, it, expect */
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
+var isNode = typeof process === 'object' &&
+             process + '' === '[object process]' &&
+             typeof window === 'undefined';
```
blesh: All you really need to do below is check to see if it has `process` and run that test:

```js
if (typeof process === 'object' && Object.prototype.toString.call(process) === '[object process]') {
   /* test `process.on('uncaughtException', doThings)` in here */
} else if (typeof window === 'object' && Object.prototype.toString.call(window) === '[object global]') {
  /* test `window.addEventListener('error', doThings);` in here */
}
```

blesh: (in other words, right there you're not *technically* testing if it's "Node", you're testing to see if it has `process`, which browsers shouldn't have, but I suppose they *could*. We deal with some pretty weird runtimes here at Netflix.

kwonoj: > but I suppose they could

: It is very correct, what I did above was simple check if process exists. Updating PR as suggested.

kwonoj:
PR updated to have test cases per `process` and `browser`.

spec/observables/from-promise-spec.js
```diff
@@ -109,24 +109,49 @@ describe('Observable.fromPromise', function () {
...
-    Observable.fromPromise(Promise.reject('bad'))
-      .subscribe(
-        done.fail,
-        function (e) {
-          expect(e).toBe('bad');
-          throw 'fail';
-        },
-        done.fail);
-  });
+      window.onerror = onException;
```
kwonoj: used `onerror` instead of `addeventlistener(error)`, on chrome eventlistener triggers listener but still prints out error like http://jsbin.com/dipizajumu/3/edit?html,js,console
```
Uncaught error
..listener behavior
```

I'm feeling I missed some obvious thing and did some dumb thing with code snippet.

blesh: > n chrome eventlistener triggers listener but still prints out error like http://jsbin.com/dipizajumu/3/edit?html,js,console (not first error actually thrown immediately)

That's right! I was thinking to myself when I was writing that I had learned `onerror` was what you wanted to use, but I could find any supporting evidence, so I posted `addEventListener`. :)

Good find.

blesh:
LGTM

kwonoj:
Merged with 9996d77. Moving forward to next step to wire with CI server..

PR #1011, opened 2015-12-09
windowTime: fixes and tests

staltz:
Fix windowTime() operator to dispose window Subjects when the destination Subscriber is
unsubscribed.

Also allow a Subscription to add a child Subscription parameterized by `<any>`, not by `<T>`. There isn't a clear reason why all child Subscriptions should be type `<T>`, and there are cases where we need to
have a `<any>` child.

### DIFF
--- a/spec/operators/windowTime-spec.js
+++ b/spec/operators/windowTime-spec.js
@@ -40,50 +40,58 @@ describe('Observable.prototype.windowTime', function () {
 
   it('should return a single empty window if source is empty', function () {
     var source =   cold('|');
+    var subs =          '(^!)';
     var expected =      '(w|)';
-    var w =         cold('|');
+    var w =        cold('|');
     var expectedValues = { w: w };
 
     var result = source.windowTime(50, 100, rxTestScheduler);
 
     expectObservable(result).toBe(expected, expectedValues);
+    expectSubscriptions(source.subscriptions).toBe(subs);
   });
 
   it('should split a Just source into a single window identical to source', function () {
     var source =   cold('(a|)');
+    var subs =          '(^!)';
     var expected =      '(w|)';
-    var w =         cold('(a|)');
+    var w =        cold('(a|)');
     var expectedValues = { w: w };
 
     var result = source.windowTime(50, 100, rxTestScheduler);
 
     expectObservable(result).toBe(expected, expectedValues);
+    expectSubscriptions(source.subscriptions).toBe(subs);
   });
 
   it('should be able to split a never Observable into timely empty windows', function () {
-    var source =    hot('^-----------');
-    var unsub =         '           !';
-    var expected =      'a--b--c--d--';
-    var a =        cold('---|        ');
-    var b =        cold(   '---|     ');
-    var c =        cold(      '---|  ');
-    var d =        cold(         '---');
+    var source =    hot('^----------');
+    var subs =          '^         !';
+    var expected =      'a--b--c--d-';
+    var a =        cold('---|       ');
+    var b =        cold(   '---|    ');
+    var c =        cold(      '---| ');
+    var d =        cold(         '--');
+    var unsub =         '          !';
     var expectedValues = { a: a, b: b, c: c, d: d };
 
     var result = source.windowTime(30, 30, rxTestScheduler);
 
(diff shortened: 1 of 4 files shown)


blesh:
Yup. This looks good. Thanks for catching this with the `window` operators.

staltz:
Cool. Next PR coming up also for `windowWhen`.

spec/operators/windowTime-spec.js
```diff
@@ -123,6 +131,37 @@ describe('Observable.prototype.windowTime', function () {
...
+    virtualTimeScheduler.schedule(function () { source.next('e'); }, 50);
+    virtualTimeScheduler.schedule(function () { source.next('f'); }, 60);
+    virtualTimeScheduler.schedule(function () { source.next('g'); }, 70);
+    virtualTimeScheduler.schedule(function () {
+      try {
+        window.subscribe();
+      }
+      catch (err) {
+        expect(err.message).toBe('Cannot subscribe to a disposed Subject.');
+        done();
```
kwonoj: should this test be async test while using virtualtimescheduler? locally tried sync assertion as other PR (https://github.com/ReactiveX/RxJS/pull/1012/files#diff-f0b7f1f59929d2619cbf53504f594616R219) seems works.

blesh: VirtualTimeScheduler is the basis for TestScheduler. You should be able to just `flush()` the whole thing synchronously. ... I'm more curious about why this is in here.

spec/operators/windowTime-spec.js
```diff
@@ -123,6 +131,37 @@ describe('Observable.prototype.windowTime', function () {
...
+      }
+    }, 80);
+
+    virtualTimeScheduler.flush();
```
blesh: Yeah, if you make assertions after this point, it doesn't need to be an async test.

spec/operators/windowTime-spec.js
```diff
@@ -123,6 +131,37 @@ describe('Observable.prototype.windowTime', function () {
     expectObservable(result, unsub).toBe(expected, values);
   });
 
+  it('should dispose window Subjects if the outer is unsubscribed early', function (done) {
+    var virtualTimeScheduler = new Rx.VirtualTimeScheduler();
```
blesh: in all synchronous tests, there's a global `rxTestScheduler` you can use to do this work. It's based off of `VirtualTimeScheduler`, and you don't have to call `flush` on it per-say. (but you could).

blesh: @staltz is there any reason this can't be done with standard marble tests?

staltz: So, my standard approach was initially marble tests of course. Then I got into problems where I wanted to subscribe to a window (an inner Observable of the outer result) **AFTER** the whole outer is unsubscribed manually. I didn't figure out how to do that with marble diagrams without creating additional weirdness around it. I can give it another try. (the same applies to all other windowVariation operator tests).

staltz:
@blesh I updated this PR to use `rxTestScheduler` with marble tests, not `virtualTimeScheduler`. I can use the same approach with other variants of window.

blesh:
@staltz ... perfect. That's exactly the right thing to do there. You can schedule anything you like on `rxTestScheduler` and it will `flush` before making the assertions.

spec/operators/windowTime-spec.js
```diff
@@ -108,19 +116,45 @@ describe('Observable.prototype.windowTime', function () {
...
+    expectObservable(result, unsub).toBe(expected, values);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+    rxTestScheduler.schedule(function () {
+      try {
+        window.subscribe();
+      }
+      catch (err) {
+        expect(err.message).toBe('Cannot subscribe to a disposed Subject.');
+      }
+    }, 150);
```
kwonoj: Will this test also to use `time` in https://github.com/ReactiveX/RxJS/pull/1032 in later?

staltz: Yes.

kwonoj: Thanks for clarification. Then let'a make #1032 land first, update this PR.

blesh:
Merged with b73e260. Thanks @staltz

PR #1012, opened 2015-12-09
window: tests and fixes

staltz:
Fix window() operator to dispose window Subjects when the destination Subscriber is unsubscribed.


kwonoj:
Change looks good to me, I'll check in this PR.

kwonoj:
Merged with 5168f73, thanks @staltz

PR #1018, opened 2015-12-10
fix(operators): force patch modules to not emit empty declarations

### LINKED ISSUE #1010: Type Definitions are empty for "add" modules
The generated `d.ts` files that correspond to the "patch" operator and observable files are empty, so they are picked up by typescript as invalid modules and the compiler errors:
```
error TS2656: 
Exported external package typings file 'node_modules/rxjs/add/observable/fromEvent.d.ts' is not a module. Please contact the package author to update the package definition.
```

This means TS users are unable to do the desired `'import rxjs/add/operator/map' ` use case. 

Workaround for angular2 users discovering this issue is to simply `'import rxjs/Rx'` which will add all the operators and observables in the meantime without error. 

This appears to go all the way back to alpha.11, which is when the /add syntax was introduced.

robwormald:
closes #1010

### DIFF
--- a/src/add/observable/bindCallback.ts
+++ b/src/add/observable/bindCallback.ts
@@ -1,3 +1,5 @@
 import {Observable} from '../../Observable';
 import {BoundCallbackObservable} from '../../observable/bindCallback';
 Observable.bindCallback = BoundCallbackObservable.create;
+
+export var _void: void;
--- a/src/add/observable/defer.ts
+++ b/src/add/observable/defer.ts
@@ -1,3 +1,5 @@
 import {Observable} from '../../Observable';
 import {DeferObservable} from '../../observable/defer';
 Observable.defer = DeferObservable.create;
+
+export var _void: void;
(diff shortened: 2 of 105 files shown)


kwonoj:
Just curiosity, what happens if delete all `.../add/**/*.d.ts' ?

robwormald:
That works, but only until we bump up to TS 1.8 and patch the operator definition in the 'add' module.

kwonoj:
Get it, thanks. (Just clarify, it wasn't suggestion to delete it, purely based on curiosity)

robwormald:
Yeah, I hadn't even considered that option, haha!

kwonoj:
I think changes good for immediate solution. (sorry, may require rebase due to my latest check in).

robwormald:
whoops, screwed up that merge. fixing...
oops

blesh:
`file mode changed`

robwormald:
okay, good to go.

kwonoj:
Cool, I'll check in in today once travis says good to go.

kwonoj:
Merged with 676f82d, appreciate @robwormald !

PR #1021, opened 2015-12-10
More tests for debounce and debounceTime

staltz:


blesh:
merged with e3512a9. Thanks @staltz

PR #1027, opened 2015-12-10
test(every): add unsubscription-related tests

staltz:
Add tests for every() operator, to verify it behaves correctly when the result Observable is
unsubscribed explicitly.


kwonoj:
Change looks good to me.

blesh:
merged with 47f2d46. thanks @staltz

PR #1032, opened 2015-12-10
Refactor window tests

staltz:
Add createTime() function to TestScheduler, which takes a simple marble diagram as string and
outputs a number for the time frame which the diagram represents.

Refactor window (and its variants) tests that check for window Subjects being disposed, to use
marble tests and not the VirtualTimeScheduler.

Makes @blesh happy.


kwonoj:
Looks fine for me.

blesh:
Merged with 72ab2e1. Thanks @staltz

PR #1035, opened 2015-12-10
Observable: fix and tests related to thisArg 

luisgabriel:


spec/operators/findindex-spec.js
```diff
@@ -41,6 +41,35 @@ describe('Observable.prototype.findIndex()', function () {
...
   });
 
+  it('should return index of matching element from source emits multiple elements', function () {
+    var sourceValue = 7;
+    var source = hot('--a--b---c-|', { b: sourceValue });
+    var subs =       '^    !';
+    var expected =   '-----(x|)';
+
+    var predicate = function (value) {
+      return value === sourceValue;
```
kwonoj: just minor ergonomics question, what about remove `sourcevalue` and make predicate to compare to value directly as same as https://github.com/ReactiveX/RxJS/pull/1035/files#diff-8e3c11ee811cb3af7dbf4d6817481356R54 does?

kwonoj:
Change looks good to me otherwise above comment. Cool to catch signature differences between implementation :)

spec/operators/findindex-spec.js
luisgabriel: Sure. No problem. Will update the PR in a few minutes.

blesh:
Merged with 43304a7. Thanks again, @luisgabriel!

PR #1036, opened 2015-12-10
style(typings): Use typeof to capture typings for all static Observab…

david-driscoll:
…le methods

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -12,6 +12,25 @@ import {Subject} from './Subject';
 import {Notification} from './Notification';
 import {rxSubscriber} from'./symbol/rxSubscriber';
 
+import {combineLatest as combineLatestStatic} from './operator/combineLatest-static';
+import {concat as concatStatic} from './operator/concat-static';
+import {merge as mergeStatic} from './operator/merge-static';
+import {zip as zipStatic} from './operator/zip-static';
+import {BoundCallbackObservable} from './observable/bindCallback';
+import {DeferObservable} from './observable/defer';
+import {EmptyObservable} from './observable/empty';
+import {ForkJoinObservable} from './observable/forkJoin';
+import {FromObservable} from './observable/from';
+import {ArrayObservable} from './observable/fromArray';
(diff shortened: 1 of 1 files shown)


david-driscoll:
This pulls out the static methods as per the discussion in #870

kwonoj:
As already discussed in original PR, I like this changes and this looks good to me. Thanks for taking time with effort @david-driscoll :)

I'll check in this once @blesh agrees.

david-driscoll:
I've also gone and rebased the other branches to be based on this one, so the changes should disappear from those PRs once this is merged.

luisgabriel:
Super cool! :+1:

kwonoj:
It's off-topic question, is this applicable to non-static instance also?

david-driscoll:
Unfortunately no, because there is no way to consider the type of `this` currently with TypeScript.

Consider:

    export function combineLatest<T2>(second: Observable<T2>): Observable<[T, T2]>;

With this function we have no idea what `T` is, so to define it properly we have to replace it with `any` to compile.

    export function combineLatest<T2>(second: Observable<T2>): Observable<[any, T2]>;

And then we loose our type information.  #870 works around this, copying the definitions from the operator, defined as a function, and strips out any references to `T` from the function declaration.

The `typinggen.js` basically converts:

    export function combineLatest<T, T2>(second: Observable<T2>): Observable<[T, T2]>;

to:

    interface Observable<T> {
        combineLatest<T2>(second: Observable<T2>): Observable<[T, T2]>;
    }

So then we can then use `typeof operator.xyz`, as seen in #871.

kwonoj:
Make quite sense, thanks. Think this topic need to be continued under original PR.

kwonoj:
@david-driscoll Would you able to align commit message start with lowercase as same as other commit history?

kwonoj:
Otherwise, change looks cool to me. I'll let this opened for a while to have any other suggestions / opinions around after commit message's aligned. Expecting to take some time due to holiday.

blesh:
Merged with d4dd7f1. Thanks @david-driscoll!

PR #1037, opened 2015-12-10
chore(tslint): Use tslint.json instead of .tslintrc

david-driscoll:
A lot of tooling is standardizing around `tslint.json` as it's the default filename referenced by tslint.  I know that it can be changed from the command line, but editors that are unaware of this name, default to settings that do not match what the settings are for this repo.


kwonoj:
I don't have strong preferences to object change name, just curiosity question - which editor do you use doesn't support configure tslint file name?

david-driscoll:
Atom with `atom-typescript` and [`linter-tslint`](https://github.com/AtomLinter/linter-tslint)

I'm sure I send a PR in to fix it, and I work with the linter team as they're making changes to linter.  If we make it a configuration then you might have to flip back and forth between as you switch projects which would get fairly tedious after a time.

I realize that `.tslintrc` more closely matches the convention of every other linter (`.jshintrc`, `.eslintrc`, etc) but it gets in the way and sometimes I catch my self seeing erroneous errors or not catching linting errors.

Even looking at the vscode source, while it lets you configure the `configFile`, it ironically only adds a watcher for tslint.json. (https://github.com/Microsoft/vscode-tslint/issues/16)

kwonoj:
Thanks for detail. For current development ergonomics, I think this PR serves better. Looks good to me. :)

kwonoj:
I'll check in this around later today to see if there's any other opinion.

staltz:
I originally put the `.tslintrc` there. I'm fine if we change it to `tslint.json`.

kwonoj:
Merged with fc3dcf7, thanks @david-driscoll

PR #1038, opened 2015-12-10
chore(tsconfig): Added codeFormatOptions

david-driscoll:
One more :smile: 

--------------------------------------------

This follows the spec defined by the typescript language services.
https://github.com/Microsoft/TypeScript/blob/master/tests/cases/fourslash/fourslash.ts#L70-L87

This is used by atom-typescript to pass off formatting options to the compiler.  I'm not aware of any other tools that use this, but extensions to tsconfig.json are not illegal unless they clash with the default properties.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,6 +7,10 @@
     "target": "es6",
     "outDir": "dist/es6"
   },
+  "formatCodeOptions": {
+    "indentSize": 2,
+    "tabSize": 2
+  },
    "files": [
      "src/Rx.ts",
      "src/Rx.KitchenSink.ts"


kwonoj:
I was curious about this changes in original PR, in RxJS currently does not uses `tsconfig.json` at all but manually specifying options in build script due to populating different build targets, instead of creating separate tsconfig.json project configuration. How does it affect to build outputs?

david-driscoll:
It doesn't affect the build at all, the compiler ignores any settings it doesn't know about.  It's just for editor experience only.  Extensions are allowed as per https://github.com/Microsoft/TypeScript/issues/1955

It's just a quality of life thing for anyone using `atom-typescript` ( probably just myself, lol :smile: )

kwonoj:
I see. I think it's ok since changes are non-invasive, but let me leave this for a while to see if there's any other suggestions around before check in changes.

tsconfig.json
```diff
@@ -7,8 +7,12 @@
     "target": "es6",
     "outDir": "dist/es6"
   },
+  "formatCodeOptions": {
+    "indentSize": 2,
+    "tabSize": 2
+  },
```
staltz: Oh, this existed. I didn't know. Thanks!

kwonoj:
Merged with bcb01d1, thanks @david-driscoll :)

kwonoj:
Forgot to close..

Issue #1041, opened 2015-12-11
Repeat and retry bugging on synchronous sources


staltz:
This happens:
```js
Rx.Observable.of(10,20,30)
  .concat(Rx.Observable.throw('poop'))
  .multicast(() => new Rx.ReplaySubject(2))
  .refCount()
  .retry(5)
  .subscribe(x => console.log(x), e => console.error(e), () => console.info('|'));

// 10
// 20
// 30
// 10
// 20
// 30
```

And this:
```js
Rx.Observable.of(10,20,30)
  .multicast(() => new Rx.ReplaySubject(2))
  .refCount()
  .repeat(5)
  .subscribe(x => console.log(x), e => console.error(e), () => console.info('|'));

// 10
// 20
// 30
// 10
// 20
// 30
```

The bug manifests itself only for `repeat(x)` where `x >= 3`, and `retry(x)` where `x >= 2`.

staltz:
I'm working on solving this.

staltz:
It's a problem with multicast and refCount, since this works:

```js
Rx.Observable.of(10,20,30)
  .concat(Rx.Observable.throw('poop'))
  .retry(4)
  .subscribe(x => console.log(x), e => console.error(e), () => console.info('|'));

// 10
// 20
// 30
// 10
// 20
// 30
// 10
// 20
// 30
// 10
// 20
// 30
// 10
// 20
// 30
// poop
```

staltz:
Seems highly related to #678. I suspect this happens because we reverted/refactored from callback-based connect() to synchronous connect.

staltz:
This can be closed, we have tests for it and everything seems green.

PR #1042, opened 2015-12-11
Tests for bug #1041 with ConnectableObservable and retry/repeat

staltz:
Add the `connect(callback)` flavor to `ConnectableObservable`, which is used internally in `refCount()` to
allow `connect(callback)` to give its subscription to its caller through the callback, before actually
performing the connection.

Resolves issue #1041.

Add test cases for this.

### DIFF
--- a/spec/operators/multicast-spec.js
+++ b/spec/operators/multicast-spec.js
@@ -219,21 +219,53 @@ describe('Observable.prototype.multicast()', function () {
       function subjectFactory() { return new Subject(); }
       var source = cold('(123#)');
       var multicasted = source.multicast(subjectFactory).refCount();
-      var subscribe1 =  's         ';
-      var expected1 =   '(123123#) ';
-      var subscribe2 =  ' s        ';
-      var expected2 =   ' (123123#)';
+      var subscribe1 =  's               ';
+      var expected1 =   '(123123123123#) ';
+      var subscribe2 =  ' s              ';
+      var expected2 =   ' (123123123123#)';
       var sourceSubs = ['(^!)',
                         '(^!)',
+                        '(^!)',
+                        '(^!)',
+                        ' (^!)',
+                        ' (^!)',
                         ' (^!)',
                         ' (^!)'];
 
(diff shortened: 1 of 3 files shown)


src/observable/ConnectableObservable.ts
```diff
@@ -25,7 +25,16 @@ export class ConnectableObservable<T> extends Observable<T> {
     return (this.subject = this.subjectFactory());
   }
 
-  connect(): Subscription<T> {
+  connect(onSubscribe?: (subscription: Subscription<T>) => void): Subscription<T> {
+    if (onSubscribe) {
+      this._callbackConnect(onSubscribe);
+      return null;
```
trxcllnt: @staltz Shouldn't `connect` still return a `Subscription`?

staltz: In the callback version? I don't know. What should that be? Where would it be used?

trxcllnt: @staltz I'm not sure, just thinking about whether somebody can call `connect` like this publicly, and whether `onSubscribe` will always synchronously invoked. If there's any (potential) asynchrony at all, we should have a `Subscription` to track it.

staltz: Well, it doesn't hurt to return the subscription. I can do that.

staltz:
@trxcllnt PR updated according to your feedback

blesh:
I'm going to label this one blocked while we discuss changing the `connect` method.

staltz:
By the way, the changes to `connect` are completely backwards compatible.

blesh:
> By the way, the changes to connect are completely backwards compatible.

Yes, they are. I'm more concerned about external users starting to use it, and then us wanting to remove it later.

staltz:
@blesh This PR is now all good to be merged, I rebased it with Paul's latests pushes to master, so now it contains only tests, no changes to ConnectableObservable.

blesh:
merged with 69ebd9a. Thanks @staltz

PR #1043, opened 2015-12-11
test(concat): add test against breaking unsubscription chain

staltz:
Add test case for static concat operator, to verify it does not break unsubscription chains to other
operators when the result is unsubscribed explicitly.


kwonoj:
Changes look good to me.

blesh:
I had to read this test twice. haha.  LGTM, though. I'd like these tests a little more if they were easier to understand.

blesh:
merged with ba70c13. thanks @staltz

PR #1045, opened 2015-12-11

kwonoj:
Minor refactoring to few observable complete when test completes.


kwonoj:
Merged with 65e65e2.

PR #1046, opened 2015-12-11
fix(distinctUntilChanged): implement optional keySelector

trxcllnt:
Currently, `distinctUntilChanged` holds on to the [previous distinct `value`](https://github.com/ReactiveX/RxJS/blob/master/src/operator/distinctUntilChanged.ts#L48), which is incorrect. `distinctUntilChanged` should use an optional `keySelector` to [pick a key from each event](https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/linq/observable/distinctuntilchanged.js#L20), and only retain the `key` itself.

### DIFF
--- a/spec/operators/distinctUntilChanged-spec.js
+++ b/spec/operators/distinctUntilChanged-spec.js
@@ -109,7 +109,7 @@ describe('Observable.prototype.distinctUntilChanged()', function () {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
-  it('should emit once if comparer returns true always regardless of source emits', function () {
+  it('should emit once if comparator returns true always regardless of source emits', function () {
     var e1 =   hot('--a--b--c--d--e--f--|');
     var e1subs =   '^                   !';
     var expected = '--a-----------------|';
@@ -118,7 +118,7 @@ describe('Observable.prototype.distinctUntilChanged()', function () {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
-  it('should emit all if comparer returns false always regardless of source emits', function () {
+  it('should emit all if comparator returns false always regardless of source emits', function () {
     var e1 =   hot('--a--a--a--a--a--a--|');
     var e1subs =   '^                   !';
     var expected = '--a--a--a--a--a--a--|';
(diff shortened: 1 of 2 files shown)


kwonoj:
Doesn't issue https://github.com/ReactiveX/RxJS/issues/83 decided to drop `keySelector` in `distinctUntilChanged` interface?

/cc @blesh

trxcllnt:
@kwonoj @blesh This PR doesn't break the existing API, just adds `keySelector` back in. This is still in line with #83, because that was just a request to swap the order of the arguments.

I have an Observable that emits a single mutable instance multiple times (because it's more efficient than being immutable), but I need to compare a specific property on the instance between emissions.

blesh:
Why is the key selector necessary? Also, having them out of order does seem strange.

blesh:
(Not opposed, just investigating)

trxcllnt:
@blesh I have an Observable of Falcor Models. Even though the Models are different, they're bound to the same branch in the cache. When the first Model is emitted, I need the `keySelector` to call `model.getVersion()`, which returns an integer. When the second Model is emitted, `keySelector` calls `model.getVersion()`, and compares against the first Model's version. I can't rely on a custom comparator, because if the comparator calls `getVersion()` on the first and second models, they'll both return the same integer.

blesh:
I see. Sounds reasonable. I'll try to review the PR in earnest soon.

spec/operators/distinctUntilChanged-spec.js
```diff
@@ -118,7 +118,7 @@ describe('Observable.prototype.distinctUntilChanged()', function () {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
-  it('should emit all if comparer returns false always regardless of source emits', function () {
+  it('should emit all if comparator returns false always regardless of source emits', function () {
```
blesh: HA! pedantatorer!

src/operator/distinctUntilChanged.ts
```diff
@@ -3,49 +3,62 @@ import {Subscriber} from '../Subscriber';
 import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
 
-export function distinctUntilChanged<T>(compare?: (x: T, y: T) => boolean) {
-  return this.lift(new DistinctUntilChangedOperator(compare));
+export function distinctUntilChanged<T>(compare?: (x: any, y: any) => boolean, keySelector?: (x: T) => any) {
```
blesh: Good catch. The typings there made no sense.

blesh:
Merged with f6a897c. Thanks, @trxcllnt

PR #1047, opened 2015-12-12
feat(Observable): add pairwise operator

justinwoo:
bring in pairwise operator from [RxJS4](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/pairwise.md).

Uses the same "has previous" and "previous" value with new array each time approach.

edit: @Frikki said that this is pretty much `bufferCount` but with a static buffer size, so maybe this isn't needed. At least pairwise will be faster!!!! Maybe. I use pairwise in RxJS4 :frowning:

### DIFF
--- a/spec/operators/pairwise-spec.js
+++ b/spec/operators/pairwise-spec.js
@@ -1,0 +1,88 @@
+/* globals describe, it, expect, expectObservable, expectSubscriptions, cold, hot */
+var Rx = require('../../dist/cjs/Rx');
+var Observable = Rx.Observable;
+
+describe('Observable.prototype.pairwise()', function () {
+  it('should pairwise things', function () {
+    var e1 = hot('--a--^--b--c--d--e--f--g--|');
+    var e1subs =      '^                    !';
+    var expected =    '------v--w--x--y--z--|';
+
+    var values = {
+      v: ['b', 'c'],
+      w: ['c', 'd'],
+      x: ['d', 'e'],
+      y: ['e', 'f'],
+      z: ['f', 'g']
+    };
+
+    var source = e1.pairwise();
+
+    expectObservable(source).toBe(expected, values);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should not emit on single-element streams', function () {
+    var e1 = hot('-----^--b----|');
+    var e1subs =      '^       !';
+    var expected =    '--------|';
+
+    var values = {
+    };
+
+    var source = e1.pairwise();
+
+    expectObservable(source).toBe(expected, values);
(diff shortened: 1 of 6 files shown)


trxcllnt:
@justinwoo thanks for the PR! I agree, a specific implementation for `pairwise` will be faster than `bufferCount`, and probably easier to understand when you're using it. @blesh +1 for merging.

kwonoj:
Change looks fine to me too in general, maybe we can consider to place this under `kitchenSink`? just my 2¢ opinion.

blesh:
@justinwoo ... great PR. This looks good so far, however I'd like a few things before I can merge it:

1. Tests covering: `empty`, `never`, error (an error mid-stream), and `throw` (Observable.throw).
2. A Test covering an observable that only emits one value.
3. Probably remove this from CoreOperators and Rx.ts, and just leave it in KitchenSink. 
4. A micro perf test vs RxJS 4's impl (to make sure we're doing the right thing)
4. Some inline docs.

We'll be publishing beta in short order. In light of that, I'd like any new operators to be very well covered with tests.

justinwoo:
Turns out DFW has good wifi :open_mouth: 

Updated to use the KitchenSink type (using similar definition to `elementAt`) and added tests for `empty`, `never`, and `throw` using the marble diagram tests.

src/Rx.ts
```diff
@@ -65,6 +65,7 @@ import './add/operator/mergeMap';
 import './add/operator/mergeMapTo';
 import './add/operator/multicast';
 import './add/operator/observeOn';
+import './add/operator/pairwise';
```
kwonoj: This would not be required if this operator's placed under `kitchenSink`, since importing `add` will patch prototype immediately.

justinwoo: Oh, oops, accidentally unstaged this.

spec/operators/pairwise-spec.js
```diff
@@ -0,0 +1,88 @@
...
+    var values = {
+    };
+
+    var source = e1.pairwise();
+
+    expectObservable(source).toBe(expected, values);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('handle mid-stream throw', function () {
```
kwonoj: I know it's quite minor thing, would you able to align test case names as same as others? (starts with `should..`)

kwonoj:
Change looks good to me.

spec/operators/pairwise-spec.js
justinwoo: Sure, pushed the change now.

blesh:
Merged with 1432e59... great work, @justinwoo!

justinwoo:
:+1: :beers:

PR #1048, opened 2015-12-12

kwonoj:
- Updates remaining cases to complete observables when test finishes.


kwonoj:
Merged with 7f67b09.

PR #1050, opened 2015-12-13
chore(typings): Enabled noImplictAny and updated typings

david-driscoll:
With this change compliation time and performance is improved

--------------------------------------

Testing the waters a little bit here, but this change doesn't include any of the massive changes to how the external typings are generated.

### DIFF
--- a/src/CoreOperators.ts
+++ b/src/CoreOperators.ts
@@ -35,9 +35,9 @@ export interface CoreOperators<T> {
                 projectResult?: (x: T, y: any, ix: number, iy: number) => R,
                 concurrent?: number) => Observable<R>;
   flatMapTo?: <R>(observable: Observable<any>, projectResult?: (x: T, y: any, ix: number, iy: number) => R, concurrent?: number) => Observable<R>;
-  groupBy?: <R>(keySelector: (value: T) => string,
+  groupBy?: <TKey, R>(keySelector: (value: T) => string,
                 elementSelector?: (value: T) => R,
-                durationSelector?: (group: GroupedObservable<R>) => Observable<any>) => Observable<GroupedObservable<R>>;
+                durationSelector?: (group: GroupedObservable<TKey, R>) => Observable<any>) => Observable<GroupedObservable<TKey, R>>;
   ignoreElements?: () => Observable<T>;
   last?: <R>(predicate?: (value: T, index: number) => boolean,
              resultSelector?: (value: T, index: number) => R,
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -21,7 +21,7 @@ import {rxSubscriber} from'./symbol/rxSubscriber';
 export class Observable<T> implements CoreOperators<T>  {
   source: Observable<any>;
   operator: Operator<any, T>;
-  _isScalar: boolean = false;
+  _isScalar = false;
 
   /**
    * @constructor
@@ -30,7 +30,7 @@ export class Observable<T> implements CoreOperators<T>  {
    * can be `next`ed, or an `error` method can be called to raise an error, or `complete` can be called to notify
    * of a successful completion.
    */
-  constructor(subscribe?: <R>(subscriber: Subscriber<R>) => Subscription<T>|Function|void) {
+  constructor(subscribe?: <R>(subscriber: Subscriber<R> | Function) => Subscription<T>) {
     if (subscribe) {
       this._subscribe = subscribe;
     }
@@ -45,8 +45,8 @@ export class Observable<T> implements CoreOperators<T>  {
    * @returns {Observable} a new cold observable
    * @description creates a new cold Observable by calling the Observable constructor
    */
-  static create: Function = <T>(subscribe?: <R>(subscriber: Subscriber<R>) => Subscription<T>|Function|void) => {
-    return new Observable<T>(subscribe);
+  static create: Function = <T>(subscribe?: <R>(subscriber: Subscriber<R> | Function) => Subscription<T>) => {
+    return new Observable(subscribe);
   };
 
   /**
@@ -56,8 +56,8 @@ export class Observable<T> implements CoreOperators<T>  {
    * @description creates a new Observable, with this Observable as the source, and the passed
    * operator defined as the new observable's operator.
    */
-  lift<T, R>(operator: Operator<T, R>): Observable<T> {
-    const observable = new Observable();
+  lift<T, R>(operator: Operator<T, R>): Observable<R> {
+    const observable = new Observable<R>();
     observable.source = this;
     observable.operator = operator;
     return observable;
@@ -98,7 +98,7 @@ export class Observable<T> implements CoreOperators<T>  {
         subscriber = new Subscriber(<Observer<T>> observerOrNext);
       }
     } else {
-      const next = <((x?) => void)> observerOrNext;
+      const next = <((x?: any) => void)> observerOrNext;
       subscriber = Subscriber.create(next, error, complete);
     }
 
@@ -115,7 +115,7 @@ export class Observable<T> implements CoreOperators<T>  {
    * @returns {Promise} a promise that either resolves on observable completion or
    *  rejects with the handled error
    */
-  forEach(next: (value: T) => void, thisArg: any, PromiseCtor?: PromiseConstructor): Promise<void> {
+  forEach(next: (value: T) => void, thisArg: any, PromiseCtor?: PromiseConstructor): Promise<any> {
     if (!PromiseCtor) {
       if (root.Rx && root.Rx.config && root.Rx.config.Promise) {
         PromiseCtor = root.Rx.config.Promise;
@@ -128,7 +128,7 @@ export class Observable<T> implements CoreOperators<T>  {
       throw new Error('no Promise impl found');
     }
 
-    let nextHandler;
+    let nextHandler: any;
 
     if (thisArg) {
       nextHandler = function nextHandlerFn(value: any): void {
@@ -141,17 +141,17 @@ export class Observable<T> implements CoreOperators<T>  {
       nextHandler = next;
     }
 
-    const promiseCallback = function promiseCallbackFn(resolve, reject) {
+    const promiseCallback = function promiseCallbackFn(resolve: Function, reject: Function) {
       const { source, nextHandler } = <any>promiseCallbackFn;
       source.subscribe(nextHandler, reject, resolve);
     };
     (<any>promiseCallback).source = this;
     (<any>promiseCallback).nextHandler = nextHandler;
 
-    return new PromiseCtor<void>(promiseCallback);
+    return new PromiseCtor(promiseCallback);
   }
 
-  _subscribe(subscriber: Subscriber<any>): Subscription<T> | Function | void {
+  _subscribe(subscriber: Subscriber<any>): Subscription<T> | Function | any {
     return this.source._subscribe(this.operator.call(subscriber));
   }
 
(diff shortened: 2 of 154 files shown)


david-driscoll:
## Current times
### ES6
Files:             280
Lines:           28651
Nodes:          130697
Identifiers:     43291
Symbols:       8548835
Types:          483111
Memory used:   824318K
I/O read:        0.03s
I/O write:       0.13s
Parse time:      1.20s
Bind time:       0.70s
Check time:     25.11s
Emit time:       1.79s
Total time:     28.80s

### AMD
Files:             249
Lines:           26793
Nodes:          123074
Identifiers:     41029
Symbols:       7474980
Types:          446931
Memory used:   720169K
I/O read:        0.03s
I/O write:       0.05s
Parse time:      0.72s
Bind time:       0.48s
Check time:     15.07s
Emit time:       0.65s
Total time:     16.93s

### CommonJS
Files:             281
Lines:           27998
Nodes:          129465
Identifiers:     42856
Symbols:       8507685
Types:          481090
Memory used:   773381K
I/O read:        0.01s
I/O write:       0.08s
Parse time:      0.74s
Bind time:       0.49s
Check time:     16.75s
Emit time:       1.17s
Total time:     19.15s

## New Times
### ES6
Files:             281
Lines:           28737
Nodes:          131467
Identifiers:     43494
Symbols:       7811638
Types:           57411
Memory used:   161333K
I/O read:        0.02s
I/O write:       0.08s
Parse time:      0.63s
Bind time:       0.40s
Check time:      2.23s
Emit time:       0.55s
Total time:      3.81s

### AMD
Files:             250
Lines:           26858
Nodes:          123798
Identifiers:     41237
Symbols:       6807061
Types:           55249
Memory used:   150043K
I/O read:        0.02s
I/O write:       0.05s
Parse time:      0.77s
Bind time:       0.65s
Check time:      2.62s
Emit time:       0.47s
Total time:      4.50s

### CommonJS
Files:             282
Lines:           28084
Nodes:          130235
Identifiers:     43059
Symbols:       7775095
Types:           57210
Memory used:   165002K
I/O read:        0.02s
I/O write:       0.08s
Parse time:      0.72s
Bind time:       0.51s
Check time:      3.01s
Emit time:       0.76s
Total time:      5.00s

kwonoj:
I personally like this changes and was considered quite while ago, wasn't sure if this is feasible for current codebases yet. Again, generally I prefer this idea to be strict manner.

Few suggestion for changes though,

- would like to suggest to enable `suppressImplicitAnyIndexErrors`. Lot of cases in RxJS, it requires force casting to `any` to indexer based fields but does not gives quite benefit.
- in case of build, you may need to update `npm build script` as well as `tsconfig.json`, since typescript project configuration is not actually being used in RxJS builds.

May have some comment in each code as well..

and cc @blesh , @staltz here also since this'll impact code styles from now on expect opinions.

src/util/isPromise.ts
```diff
@@ -1,3 +1,3 @@
-export function isPromise(value: any): boolean {
-  return value && typeof value.subscribe !== 'function' && typeof value.then === 'function';
-}
\ No newline at end of file
+export function isPromise<T>(value: any | Promise<T>): value is Promise<T> {
```
kwonoj: Would you able to share some details of
- what does generic <T> does do?
- reason to change return type from boolean?

src/util/MapPolyfill.ts
```diff
@@ -29,9 +29,9 @@ export class MapPolyfill {
     return true;
   }
 
-  forEach(cb, thisArg) {
+  forEach(cb: any, thisArg: any) {
```
kwonoj: this case seems `cb` can be typeof `Function`

david-driscoll:
@kwonoj I totally expect opinions, hence testing the waters first. :smile: 

As far as `suppressImplicitAnyIndexErrors` I never ran into a case, inside the current `src`, so I didn't even look at the flag.  Now that might change once unit tests start to get converted over.

I tried to make the changes as non-invasive as possible, let me know if you notice something left behind.  No actual code changes, just the types.

The one thing that impressed me was the drastic drop in memory usage, and the drastic reduction in types that were being tracked by the compiler.

src/types.ts
```diff
@@ -0,0 +1,18 @@
+import {Observable} from './Observable';
```
kwonoj: I think this file (`types.ts`) is not landed in current TOT of codebase, isn't it?

src/operator/zip-support.ts
```diff
@@ -221,7 +221,7 @@ class ZipBufferIterator<T, R> extends OuterSubscriber<T, R> implements LookAhead
     }
   }
 
-  notifyNext(outerValue, innerValue, outerIndex, innerIndex) {
+  notifyNext(outerValue: any, innerValue: any, outerIndex: any, innerIndex: any) {
```
kwonoj: `index` can be number maybe?

src/operator/zip-support.ts
```diff
@@ -56,8 +56,8 @@ export class ZipSubscriber<T, R> extends Subscriber<T> {
     this.active = len;
     for (let i = 0; i < len; i++) {
       let iterator = iterators[i];
-      if (iterator.stillUnsubscribed) {
-        iterator.subscribe(iterator, i);
+      if ((iterator as any).stillUnsubscribed) {
```
kwonoj: what about cast type of `iterator` in line 58 once? and is there any feasible type other than `any`?

kwonoj:
@david-driscoll Thanks for effort, this requires lot of changes. :) 

One another suggestion in addition to above, I could see some changes like `types.ts` (https://github.com/david-driscoll/RxJS-1/blob/all-types-without-extras/src/types.ts) are included.

Given circumstances of PR itself is already huge due to nature of changes, I'd like to suggest to isolate PR as much as possible to not include directly related changes. Those refactoring can be separated into another PR, (maybe I presume too early?)

kwonoj:
I'll hold on to add comment into each individual files for now to gather opinions regarding essence of this PR first. Once it's agreed to land this, detail can be updated from that point.

david-driscoll:
I left in types on purpose because of the usefulness of the common types for handing complicated declaration.

In some cases we see method signatures like `(outerValue: T1, innerValue: T2, outerIndex: number, innerIndex: number) => TResult` show up 3 times in a file with the `function`, `Operator` and `Subscriber`.  It feels excessive and eats up a ton of space and is prone to failure.  Also it opens up the ability to change what a `_Selector` is in a central location and begin able to see that ripple through.  I suppose they could be replaced out, but I felt they offered some value, so left them in at least as a discussion point.

kwonoj:
> I left in types on purpose because of the usefulness of the common types for handing complicated declaration.

: Yes, I do not deny those and I agree in general, suggestion is simply better to have separate PR for those. Once you're able to create another PR for those single changes only, that can be easily discussed in checked in, while this PR is on discussion. :) 

In this PR currently, it contains 2 refactoring of

- applying `noimplicitany`
- adopting common type definition (types.ts)

and first change is quite huge, mixed discussion in this single PR could possibly delay check in of 2nd as well, and reviewer also requires additional effort (to discuss types.ts, have to open changes tab has >100 files). Please consider it as suggestion for purely logistics perspective.

luisgabriel:
Could you update the subject of the commit message to use lower case and present tense? Just a minor detail to keep the same [pattern](https://github.com/ReactiveX/RxJS/blob/master/CONTRIBUTING.md#commit) as the others commits in the history. ;)

src/util/isPromise.ts
david-driscoll: The return type still behaves as a boolean, but it also acts as a type guard as well.

Given the following:

    var value: Observable<T> | Promise<T>
    if (isPromise(value)) {
        // to the type system value is now just Promise<T>
    } else {
        // to the type system value is now just Observable<T>
    }

staltz:
I am in general favorable of merging this PR, but I haven't reviewed all the changes since there are many.

blesh:
> I am in general favorable of merging this PR, but I haven't reviewed all the changes since there are many.

Same.

kwonoj:
OK, think we're in consensus. I'll drive initial effort for in-detail review for changes, and will bring in @blesh , @staltz later before finally merge it.

Before proceed further comments, @david-driscoll would you kindly separate PR between 2 I commented above (https://github.com/ReactiveX/RxJS/pull/1050#issuecomment-164224748)?
I still believe separate PR serves better to deliver this vehicle into codebase.

src/CoreOperators.ts
```diff
@@ -35,9 +35,9 @@ export interface CoreOperators<T> {
                 projectResult?: (x: T, y: any, ix: number, iy: number) => R,
                 concurrent?: number) => Observable<R>;
   flatMapTo?: <R>(observable: Observable<any>, projectResult?: (x: T, y: any, ix: number, iy: number) => R, concurrent?: number) => Observable<R>;
-  groupBy?: <R>(keySelector: (value: T) => string,
+  groupBy?: <TKey, R>(keySelector: (value: T) => string,
```
luisgabriel: What about `K` instead of `TKey`? 

It would keep the single letter pattern. It would also keep it similar to the [signature used by RxJava](http://reactivex.io/RxJava/javadoc/rx/Observable.html#groupBy(rx.functions.Func1,%20rx.functions.Func1)).

src/Observable.ts
```diff
@@ -216,9 +216,9 @@ export class Observable<T> implements CoreOperators<T>  {
                projectResult?: (x: T, y: any, ix: number, iy: number) => R,
                concurrent?: number) => Observable<R>;
   flatMapTo: <R>(observable: Observable<any>, projectResult?: (x: T, y: any, ix: number, iy: number) => R, concurrent?: number) => Observable<R>;
-  groupBy: <R>(keySelector: (value: T) => string,
+  groupBy: <TKey, R>(keySelector: (value: T) => string,
```
luisgabriel: Same here.

luisgabriel:
There are a lot of good stuff in this PR. But I agree with @kwonoj here. This PR is too big to be reviewed properly.

I also have mixed feeling about some changes. For example:
```diff
-  if (result === errorObject) {
-    return new ErrorObservable(errorObject.e);
+  if (result as any === errorObject) {
+    return new ErrorObservable<any>(errorObject.e);
```

```diff
-const observableProto = (<KitchenSinkOperators<any>>Observable.prototype);
+const observableProto = (<KitchenSinkOperators<any>><any>Observable.prototype);
```

```diff
-Observable.prototype.exhaust = exhaust;
+(Observable.prototype as any).exhaust = exhaust;
```

I think this kind of change adds verbosity without many gains.

kwonoj:
I'll put on hold reviewing this PR and wait for separation of changes for now.

david-driscoll:
@kwonoj I'll see if I have time tonight to gets things apart, might take a little bit of work.

@luisgabriel 

```diff
-  if (result === errorObject) {
-    return new ErrorObservable(errorObject.e);
+  if (result as any === errorObject) {
+    return new ErrorObservable<any>(errorObject.e);
```

In the context of the method it makes sense, we could instead say it's of type T to match the return value.  The big thing is errors in Rx aren't the same type as the observable might have contextually, in fact it adds value here, we're saying the error can actually be anything.

```diff
-const observableProto = (<KitchenSinkOperators<any>>Observable.prototype);
+const observableProto = (<KitchenSinkOperators<any>><any>Observable.prototype);
```

This could just be converted to `const observableProto = (<any>Observable.prototype)` or to maintain typing `const observableProto: KitchenSinkOperators<any> = (<any>Observable.prototype);`, without changes it does become an error with noImplictAny.

```diff
-Observable.prototype.exhaust = exhaust;
+(Observable.prototype as any).exhaust = exhaust;
```
This is again just a typing that can be corrected, I just hadn't corrected it.  exhaust was conceptually moved to the extended space, but still exists on the base Observable.

luisgabriel:
> In the context of the method it makes sense, we could instead say it's of type T to match the return value. The big thing is errors in Rx aren't the same type as the observable might have contextually, in fact it adds value here, we're saying the error can actually be anything.

I see the value of making the type parameter of `ErrorObservable` explicit. But what about the `result as any` casting?


> `const observableProto: KitchenSinkOperators<any> = (<any>Observable.prototype);`

Much better. ;)

david-driscoll:
best we can do is `(<any>result) === errorObject` because the type of the result is not the same type as errorObject.  TypeScript is helping us here saying they should never be equal, now under the covers that is clearly not the case because we pass around `errorObject` into result if an error occurs.  

I just thought the as any syntax was easier to write/type than `(<any>result)`, I don't care either way I don't really care how we cast it, lol.

luisgabriel:
Yeah, I was not questioning the casting syntax, but the casting itself. I prefer not to do casting in situations such as this one. But I suppose it is needed to enable `noImplictAny`.

david-driscoll:
we could make the `errorObject:`, but then we'd lose it's definition, now I don't know if that matters to much.

kwonoj:
Marked as blocked for now to prevent accidental merging, https://github.com/ReactiveX/RxJS/pull/1077 should be resolved first.

kwonoj:
@david-driscoll , I think https://github.com/ReactiveX/RxJS/pull/1086 is end conclusions for this PR's orignally aimed for and this can be closed by checking ing #1086 - is my understanding correct?

david-driscoll:
Yeah, we should be good for this issue.

PR #1051, opened 2015-12-14
chore(package.json): add Andre as contributor

staltz:
Because, why not...
And @kwonoj could do the same.


kwonoj:
Haha, curious if there's rule to be included as contributor in package definition. I'll leave this PR since I can't decide it. :)

blesh:
REJECTED! 😛


I've added you both myself. :tada:

kwonoj:
Woohoo! /dance

PR #1055, opened 2015-12-14
Fix current-thread micro perf tests

luisgabriel:


blesh:
Thanks @luisgabriel! Good catch. Clearly I haven't run the tests since that change last week.

PR #1058, opened 2015-12-14
Add UMD and System.register bundles

jeffbcross:


jeffbcross:
I am doing manual testing.

blesh:
@jeffbcross is there no known way to automate a test of this?

jeffbcross:
@blesh I was manually testing the system output with actual angular app code that uses system format. When these get published and consumed inside Angular, our e2e tests test them. I also tested UMD by putting the bundle in an empty HTML page and playing with Rx in the console.

I'm sure there's a good way I could automate tests inside of rxjs, just not sure the best way to approach in a time crunch :). I'll try to think of something.

jeffbcross:
I should add more comments to the script after it's been reviewed.

There is a hack inside the script that creates an `rxjs` symlink to `cjs` so that the builder will register all modules as `rxjs/foo`. The script cleans up the symlink when done.

robwormald:
> There is a hack inside the script that creates an rxjs symlink to cjs so that the builder will register all modules as rxjs/foo. The script cleans up the symlink when done.

You should be able to achieve this with baseURL: 'dist/cjs'

robwormald:
Also note for a simple smoke test, you could probably do something like (in node)
```js
var System = require('systemjs');
//populate registry
System.import('Rx.js')
  .then(() => {
     //load from populated registry
     return System.import('rxjs/Rx');
  })
```

jeffbcross:
> You should be able to achieve this with baseURL: 'dist/cjs'

@robwormald can you give a more complete example? i.e. the first argument to `builder.build`

robwormald:
assuming you are building from dist/cjs/**:
```js
var Builder = require('systemjs-builder');
//create new builder instance
var builder = new Builder();

//configure it
builder.config({
  //force all path resolution to start here
  baseUrl: 'dist/cjs'
});
//now we're resolving starting relative to dist/cjs
builder.build('rxjs/Rx', 'dist/outputfile').then(...);
```

jeffbcross:
@robwormald awesome, thanks, I'll try that.

jeffbcross:
@robwormald wait, you realize there's not actually an `rxjs` folder inside `cjs`, right?

robwormald:
```js

var Builder = require('systemjs-builder');

var b = new Builder();

b.config({
	baseURL: 'dist',
	paths: {
		'rxjs/*': 'cjs/*.js'
	}
});

b.build('rxjs/Rx', 'rxbundle.js').catch(function(err){
	console.log(err);
})
```

jeffbcross:
@robwormald YTB that does the trick.

blesh:
Merged with c2daaf3. thanks @jeffbcross

jeffbcross:
Thanks, @blesh! I'll work on getting the spec added in a separate PR once I figure out `System` config

PR #1059, opened 2015-12-14
Cleaning up file structure

blesh:
This PR does two things:

1. Aligns spec names with their implementation counterParts (`async-subject-spec` becomes `AsyncSubject-spec` for example)
2. Moves all "extended" operators out of the "extended" folder and down with their siblings.

Change number 2 above is because we're moving to a more modular architecture and I don't think people should have to know arbitrary distinctions about what's "extended" and what's not when importing modules. `import {map} from 'rxjs/operators/map'` vs `import {min} from 'rxjs/operators/extended/min'`. The "extended" feels unimportant.


alxhub:
LGTM

blesh:
Thanks for the review @alxhub

PR #1061, opened 2015-12-15
refactor(Observable): remove exhaust from core rx

kwonoj:
Relates to PR https://github.com/ReactiveX/RxJS/pull/1002,

found out reading https://github.com/justinwoo/RxJS/commit/767b1f54e2d73551f3f1c156f6e1714502cd79e4#commitcomment-14966118, `exhaust` and `exhausemap` lives on Observable core and KitchenSink both. Removed from `Observable` to match with original PR intended as 'removes both operators from Rx core output'.


kwonoj:
I'll check in this change once got confirmation from @blesh if this is correctly intended in original PR.

kwonoj:
Merged with 0f38730.

PR #1062, opened 2015-12-15
fix(Subscription): instead of defining new const, just check private …

willheslam:
…variable

With the current implementation, after building to ES6 (or using latest published version with: npm install rxjs-es), trying to import any aspect of Rx results in this error in my webpack build:

```
ERROR in ./~/rxjs-es/Subscription.js
Module build failed: SyntaxError: /Users/wheslam/streamtest/node_modules/rxjs-es/Subscription.js: "unsubscribe" is read-only (This is an error on an internal node. Probably an internal error. Location has been estimated.)
```

I'm not really sure why this error is occurring as looking at the ES6 version of rxjs looks fine. Perhaps it is something to do with the bundling process?
Renaming the variable from 'const unsubscribe' to 'const unsubscribe2' fixes the problem - changing it to 'let unsubscribe' gives:

```
ERROR in ./~/rxjs-es/Subscription.js
Module build failed: TypeError: /Users/wheslam/streamtest/node_modules/rxjs-es/Subscription.js: Duplicate declaration "unsubscribe"
```

Making this change fixes it too - however I'm not particularly happy with this 'fix' as I'm not sure why it fixes it!
What are your thoughts on this? Any idea why this is happening?
(maybe I'm missing something obvious with the new ES6 module syntax)
Has anyone had any luck using rxjs-es in a webpack build yet? What was the original intention behind declaring a const in that way in the original implementation? Lots of questions. :)

See here for a simple test case:
https://github.com/willheslam/streamtest
git clone, npm install then npm run dev

### DIFF
--- a/src/Subscription.ts
+++ b/src/Subscription.ts
@@ -28,13 +28,12 @@ export class Subscription<T> {
 
     this.isUnsubscribed = true;
 
-    const unsubscribe = this._unsubscribe;
     const subscriptions = this._subscriptions;
 
     this._subscriptions = void 0;
 
-    if (unsubscribe) {
-      unsubscribe.call(this);
+    if (this._unsubscribe) {
+      this._unsubscribe.call(this);
     }
 
     if (subscriptions != null) {


kwonoj:
(Just try to have some more detail) So is this happening when executing webpack-bundled?

willheslam:
Yeah - I get this error when I run 'npm run dev' (see package.json in the test case) which kicks off my webpack bundle process - this is before even running the code.

If I try running the resulting bundle, I get 'Error: Cannot find module "./Subscription"' as a result of the syntax error during bundling.

kwonoj:
This is bit interesting to me for couple of reasons, first const variable `unsubscribe` is actually readonly and have not assigned to new value in function scope, and secondly same pattern is being used in next line `const subscriptions = this._subscriptions;` also but does not causes any issues (also this pattern's widely being used over all codebases).

I'd like to spend some time to figure out possible root cause first before check in this changes to understand what's happening. If this is really something related with es6 code generated by typescript, it'll require update all codebases too.

willheslam:
Agreed, this is very weird. The code resulting from typescript looks fine to me, so this could well be a webpack and/or babel bug. Thanks for sanity checking this - one way to test if it's a babel problem would be to find a copy of babel 5 (and the requisite babel-loader, etc.) and see if it changes - something for tomorrow night. :)

Have you had any experience using rxjs-es with babel yet?

kwonoj:
We actually had once to transpiliing using babel such as ts->es6(using ts)->cjs(using babel) but eventually migrated to typescript completely, before babel releases latest 6 version.

As an initial step, I'd like to try some some different bundler, as well as transpiler such as babel(with latest 6) etcs to check out what the end result is.

willheslam:
The plot thickens!
Using rollup as a bundler without babel works (though the resulting code uses ES7, so running it via node chokes)
```
streamtest wheslam$ npm run rollup

> streamtest@1.0.0 rollup /Users/wheslam/streamtest
> rollup -c rollup.config.js

Treating 'rollup-plugin-npm' as external dependency
Treating 'rollup-plugin-babel' as external dependency
ML0544:streamtest wheslam$ node dist/rollup.bundle.js
/Users/wheslam/streamtest/dist/rollup.bundle.js:394
                    const { thisArg, next } = nextHandlerFn;
                          ^

SyntaxError: Unexpected token {
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:404:25)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:313:12)
    at Function.Module.runMain (module.js:457:10)
    at startup (node.js:138:18)
    at node.js:974:3
```

But if I use rollup with its babel plugin, I get exactly the same issue:
streamtest wheslam$ npm run rollup

```
> streamtest@1.0.0 rollup /Users/wheslam/streamtest
> rollup -c rollup.config.js

Treating 'rollup-plugin-npm' as external dependency
Treating 'rollup-plugin-babel' as external dependency
/Users/wheslam/streamtest/node_modules/rxjs-es/Subscription.js: "unsubscribe" is read-only (This is an error on an internal node. Probably an internal error. Location has been estimated.)
SyntaxError: /Users/wheslam/streamtest/node_modules/rxjs-es/Subscription.js: "unsubscribe" is read-only (This is an error on an internal node. Probably an internal error. Location has been estimated.)
```

This is looking more and more like a problem with Babel!

Here's my test case with rollup integration now:
https://github.com/willheslam/streamtest/tree/rollup

Next up, I'm going to try webpack with an alternative transpiler like Traceur. Any suggestions about bundlers/transpilers I should try? I actually couldn't get browserify to work with babel 6 at all.

kwonoj:
Immediate candidate on my mind was browserify, but you already mentioned it :)

willheslam:
Can confirm it works beautifully with Traceur and Webpack! Bundles and runs in node just fine.
https://github.com/willheslam/streamtest/tree/webpack-traceur

Looks like this is a babel bug.

If you run:

```
babel Subscription.js --presets='es2015'
```

directly inside rxjs-es, you'll get exactly the same error.

Reported to Babel here:
https://phabricator.babeljs.io/T6830

kwonoj:
Thanks for effort to find out, would you be ok to close this issue on RxJS end since it's reported to babel?

willheslam:
Sure thing - thanks for all the help! :) Can't wait for Rxjs-es to be ready for prime time!

kwonoj:
@willheslam It is very appreciate to take effort to find out issues, also triage root cause. Thanks again, and feel freely create issue / PR anytime! :)

PR #1065, opened 2015-12-15
refactor(Subscription): remove the type parameter from `Subscription`

saneyuki:
This type parameter is not used actually.

### DIFF
--- a/src/Subscription.ts
+++ b/src/Subscription.ts
@@ -1,14 +1,14 @@
 import {noop} from './util/noop';
 
-export class Subscription<T> {
-  public static EMPTY: Subscription<void> = (function(empty){
+export class Subscription {
+  public static EMPTY: Subscription = (function(empty){
     empty.isUnsubscribed = true;
     return empty;
-  }(new Subscription<void>()));
+  }(new Subscription()));
 
   isUnsubscribed: boolean = false;
 
-  _subscriptions: Subscription<any>[];
+  _subscriptions: Subscription[];
 
   _unsubscribe(): void {
     noop();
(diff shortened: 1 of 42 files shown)


src/Subscription.ts
```diff
@@ -1,14 +1,14 @@
 import {noop} from './util/noop';
 
-export class Subscription<T> {
```
staltz: Hmm, makes sense to me. :+1:

kwonoj:
Yes, this seems quite legit to me too.

src/operator/groupBy-support.ts
```diff
@@ -3,8 +3,8 @@ import {Subject} from '../Subject';
 import {Subscriber} from '../Subscriber';
 import {Observable} from '../Observable';
 
-export class RefCountSubscription<T> extends Subscription<T> {
-  primary: Subscription<T>;
+export class RefCountSubscription<T> extends Subscription {
```
luisgabriel: I think you can also remove the type parameter from `RefCountSubscription<T>`. What do you think?

src/operator/groupBy-support.ts
```diff
@@ -44,7 +44,7 @@ export class GroupedObservable<T> extends Observable<T> {
   }
 }
 
-export class InnerRefCountSubscription<T> extends Subscription<T> {
+export class InnerRefCountSubscription<T> extends Subscription {
```
luisgabriel: Same here.

saneyuki:
> I think you can also remove the type parameter from RefCountSubscription<T>. What do you think?
> Same here.

okay. I removed them!

luisgabriel:
LGTM.

kwonoj:
Change looks good to me too.

saneyuki:
rebased! I resolved the conflict

blesh:
LGTM. cc/ @kwonoj

kwonoj:
I'll check in this today.

kwonoj:
Actually, doing it now...

kwonoj:
Merged with a8941d7, thanks @saneyuki . Appreciate for effort along with nice catch.

PR #989, opened 2015-12-08
test(operators): add tests against breaking unsubscription chain

staltz:
For issue #875.

PR #1073, opened 2015-12-15
docs(scan): add inline docs to scan

justinwoo:
I'm going to end up having a whole lot of these doc PRs. Should I open a PR for each one so tracking changes/editing is easier, or smash them into a big commit/series of commits in this PR?

Also, is there a convention we should use for these docs? oldRxJS docs use a variety of terms like "source/observable/sequence/stream" and "input/output/element/event" and "trigger/emit(?)". Maybe I just missed where these conventions are defined.

### DIFF
--- a/src/operator/scan.ts
+++ b/src/operator/scan.ts
@@ -4,6 +4,13 @@ import {Subscriber} from '../Subscriber';
 import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
 
+/**
+ * Applies an accumulator function to each input. Optional seed value is used as the initial value if supplied; otherwise, the first input is used.
+ * For aggregation with no intermediate results, see Observable.aggregate
+ * @param {Function} accumulator An accumulator function called on each input.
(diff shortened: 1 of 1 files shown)


kwonoj:
Thinking about PR like https://github.com/ReactiveX/RxJS/pull/989, I think it makes sense to create aggregated PR for this as initial check in, then adding separate following PR if there's missing.

justinwoo:
Closing this in favor of #1079, unless someone prefers this PR instead.

src/operator/scan.ts
```diff
@@ -4,6 +4,13 @@ import {Subscriber} from '../Subscriber';
 import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
 
+/**
+ * Applies an accumulator function to each input. Optional seed value is used as the initial value if supplied; otherwise, the first input is used.
+ * For aggregation with no intermediate results, see Observable.aggregate
```
staltz: What is Observable.aggregate? Shouldn't it be `reduce`?

Issue #1030, opened 2015-12-10
change script to add comments to all auto-patching operators to indicate generated file


jeffbcross:
Since there is a script to generate all of the auto-patching operator modules inside the `add` folder whenever operators change, comments should be added to those files indicating that they should not be manually edited without updating the script.
i.e.
```typescript
/**
 * Everything in this file is generated by the whatever-script.js
 * Any manual edits to this file will be lost next time the script is run.
 **/
```

The script is located at `tools/generate-operator-patches.js` and the affected files are in `src/add*`.

jeffbcross:
Unassigning for now since I'll be offline for a few weeks.

luisgabriel:
I can work on this one.

jeffbcross:
@luisgabriel awesome! And to be clear, it's the script that should be updated to include comments in those files. I'll update the issue title to make it less ambiguous.

luisgabriel:
@jeffbcross ok. ;)

kevinvandervlist:
Three small things I happened to come across while doing [something unrelated](https://github.com/ReactiveX/RxJS/pull/1105):

a) On a fresh clone the generator fails because the following folder does not exist: ./src/operator/extended

b) The following files seem to be generated incorrectly (related to the kitchensink stuff?)
* src/add/operator/distinctUntilKeyChanged.ts
* src/add/operator/elementAt.ts 
* src/add/operator/exhaust.ts 
* src/add/operator/exhaustMap.ts 
* src/add/operator/find.ts 
* src/add/operator/findIndex.ts 
* src/add/operator/isEmpty.ts 
* src/add/operator/max.ts 
* src/add/operator/mergeScan.ts 
* src/add/operator/min.ts 
* src/add/operator/timeInterval.ts

c) Since the files will be autogenerated anyway, shouldn't they be removed from the source tree, and generated on build?

kwonoj:
a / b can be discussed in PR #1097 directly - @luisgabriel , does current PR reflects this?

c - Think it depends on how we'll treat auto generated files. These files are not change frequently (mostly won't, unless adding / removing new operator) doesn't necessarily to be generated per build everytime.

luisgabriel:
I think `a` and `b` are being covered by #1097.

kwonoj:
Closing this issue by #1097 is checked in now.

PR #517, opened 2015-10-13
fix(Subscriber): avoid implicit any

robwormald:
picky, but ts compiler complains about it if you're consuming the TS source.

PR #1077, opened 2015-12-15
chore(typings): Enabled noImplictAny

david-driscoll:
With this change compliation time and performance is improved

------

@kwonoj  This was the quick way, there might a few places where the formatting, or in some cases the variable names used in method signatures that have changed.  If I need to reverse those let me know and I'll make a pass at it shortly.

### DIFF
--- a/src/CoreOperators.ts
+++ b/src/CoreOperators.ts
@@ -35,9 +35,9 @@ export interface CoreOperators<T> {
                 projectResult?: (x: T, y: any, ix: number, iy: number) => R,
                 concurrent?: number) => Observable<R>;
   flatMapTo?: <R>(observable: Observable<any>, projectResult?: (x: T, y: any, ix: number, iy: number) => R, concurrent?: number) => Observable<R>;
-  groupBy?: <R>(keySelector: (value: T) => string,
+  groupBy?: <K, R>(keySelector: (value: T) => string,
                 elementSelector?: (value: T) => R,
-                durationSelector?: (group: GroupedObservable<R>) => Observable<any>) => Observable<GroupedObservable<R>>;
+                durationSelector?: (group: GroupedObservable<K, R>) => Observable<any>) => Observable<GroupedObservable<K, R>>;
   ignoreElements?: () => Observable<T>;
   last?: <R>(predicate?: (value: T, index: number) => boolean,
              resultSelector?: (value: T, index: number) => R,
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -51,7 +51,7 @@ export class Observable<T> implements CoreOperators<T>  {
    * can be `next`ed, or an `error` method can be called to raise an error, or `complete` can be called to notify
    * of a successful completion.
    */
-  constructor(subscribe?: <R>(subscriber: Subscriber<R>) => Subscription|Function|void) {
+  constructor(subscribe?: <R>(subscriber: Subscriber<R>) => Subscription | Function | void) {
     if (subscribe) {
       this._subscribe = subscribe;
     }
@@ -66,7 +66,7 @@ export class Observable<T> implements CoreOperators<T>  {
    * @returns {Observable} a new cold observable
    * @description creates a new cold Observable by calling the Observable constructor
    */
-  static create: Function = <T>(subscribe?: <R>(subscriber: Subscriber<R>) => Subscription|Function|void) => {
+  static create: Function = <T>(subscribe?: <R>(subscriber: Subscriber<R>) => Subscription | Function | void) => {
     return new Observable<T>(subscribe);
   };
 
@@ -77,8 +77,8 @@ export class Observable<T> implements CoreOperators<T>  {
    * @description creates a new Observable, with this Observable as the source, and the passed
    * operator defined as the new observable's operator.
    */
-  lift<T, R>(operator: Operator<T, R>): Observable<T> {
-    const observable = new Observable();
+  lift<T, R>(operator: Operator<T, R>): Observable<R> {
+    const observable = new Observable<R>();
     observable.source = this;
     observable.operator = operator;
     return observable;
@@ -132,7 +132,7 @@ export class Observable<T> implements CoreOperators<T>  {
       throw new Error('no Promise impl found');
     }
 
-    let nextHandler;
+    let nextHandler: any;
 
     if (thisArg) {
       nextHandler = function nextHandlerFn(value: any): void {
@@ -145,7 +145,7 @@ export class Observable<T> implements CoreOperators<T>  {
       nextHandler = next;
     }
 
-    const promiseCallback = function promiseCallbackFn(resolve, reject) {
+    const promiseCallback = function promiseCallbackFn(resolve: Function, reject: Function) {
       const { source, nextHandler } = <any>promiseCallbackFn;
       source.subscribe(nextHandler, reject, resolve);
     };
@@ -212,9 +212,9 @@ export class Observable<T> implements CoreOperators<T>  {
                projectResult?: (x: T, y: any, ix: number, iy: number) => R,
                concurrent?: number) => Observable<R>;
   flatMapTo: <R>(observable: Observable<any>, projectResult?: (x: T, y: any, ix: number, iy: number) => R, concurrent?: number) => Observable<R>;
-  groupBy: <R>(keySelector: (value: T) => string,
+  groupBy: <K, R>(keySelector: (value: T) => string,
                elementSelector?: (value: T) => R,
-               durationSelector?: (group: GroupedObservable<R>) => Observable<any>) => Observable<GroupedObservable<R>>;
+               durationSelector?: (group: GroupedObservable<K, R>) => Observable<any>) => Observable<GroupedObservable<K, R>>;
   ignoreElements: () => Observable<T>;
   last: <R>(predicate?: (value: T, index: number) => boolean,
             resultSelector?: (value: T, index: number) => R,
--- a/src/Subject.ts
+++ b/src/Subject.ts
@@ -9,7 +9,7 @@ import {rxSubscriber} from './symbol/rxSubscriber';
 export class Subject<T> extends Observable<T> implements Observer<T>, Subscription {
 
   static create<T>(source: Observable<T>, destination: Observer<T>): Subject<T> {
-    return new Subject(source, destination);
+    return new Subject<T>(source, destination);
   }
 
   constructor(source?: Observable<T>, destination?: Observer<T>) {
@@ -32,7 +32,7 @@ export class Subject<T> extends Observable<T> implements Observer<T>, Subscripti
   lift<T, R>(operator: Operator<T, R>): Observable<T> {
     const subject = new Subject(this, this.destination || this);
     subject.operator = operator;
-    return subject;
+    return <any>subject;
   }
 
   add(subscription: Subscription|Function|void): void {
--- a/src/observable/ScalarObservable.ts
+++ b/src/observable/ScalarObservable.ts
@@ -13,7 +13,7 @@ export class ScalarObservable<T> extends Observable<T> {
     return new ScalarObservable(value, scheduler);
   }
 
-  static dispatch(state): void {
+  static dispatch(state: any): void {
     const { done, value, subscriber } = state;
 
     if (done) {
@@ -72,7 +72,7 @@ proto.filter = function <T>(select: (x: T, ix?: number) => boolean, thisArg?: an
   } else if (result) {
     return this;
   } else {
-    return new EmptyObservable();
+    return new EmptyObservable<T>();
   }
 };
 
@@ -107,7 +107,7 @@ proto.count = function <T>(predicate?: (value: T, index: number, source: Observa
 
 proto.skip = function <T>(count: number): Observable<T> {
   if (count > 0) {
-    return new EmptyObservable();
+    return new EmptyObservable<T>();
   }
   return this;
 };
@@ -116,5 +116,5 @@ proto.take = function <T>(count: number): Observable<T> {
   if (count > 0) {
     return this;
   }
-  return new EmptyObservable();
+  return new EmptyObservable<T>();
 };
--- a/src/observable/forkJoin.ts
+++ b/src/observable/forkJoin.ts
@@ -7,16 +7,15 @@ import {isArray} from '../util/isArray';
 
 export class ForkJoinObservable<T> extends Observable<T> {
   constructor(private sources: Array<Observable<any> | Promise<any>>,
-              private resultSelector?: (...values: Array<any>) => any) {
+              private resultSelector?: (...values: Array<any>) => T) {
     super();
   }
 
-  static create(...sources: Array<Observable<any> |
+  static create<T>(...sources: Array<Observable<any> | Promise<any> |
                                   Array<Observable<any>> |
-                                  Promise<any> |
-                                  ((...values: Array<any>) => any)>): Observable<any> {
+                                  ((...values: Array<any>) => any)>): Observable<T> {
     if (sources === null || arguments.length === 0) {
-      return new EmptyObservable();
+      return new EmptyObservable<T>();
     }
 
     let resultSelector: (...values: Array<any>) => any = null;
@@ -49,7 +48,7 @@ export class ForkJoinObservable<T> extends Observable<T> {
 }
 
 class AllSubscriber<T> extends Subscriber<T> {
-  private _value: any = null;
+  private _value: T = null;
 
   constructor(destination: Subscriber<any>,
               private index: number,
@@ -60,7 +59,7 @@ class AllSubscriber<T> extends Subscriber<T> {
     super(destination);
   }
 
-  _next(value: any): void {
+  _next(value: T): void {
     this._value = value;
   }
 
@@ -95,7 +94,7 @@ function hasValue(x: any): boolean {
 }
 
 function emptyArray(len: number): any[] {
-  let arr = [];
+  let arr: any[] = [];
   for (let i = 0; i < len; i++) {
     arr.push(null);
   }
--- a/src/observable/fromEvent.ts
+++ b/src/observable/fromEvent.ts
@@ -15,7 +15,7 @@ export class FromEventObservable<T, R> extends Observable<T> {
   }
 
   private static setupSubscription<T>(sourceObj: any, eventName: string, handler: Function, subscriber: Subscriber<T>) {
-    let unsubscribe;
+    let unsubscribe: () => void;
     let tag = sourceObj.toString();
     if (tag === '[object NodeList]' || tag === '[object HTMLCollection]') {
       for (let i = 0, len = sourceObj.length; i < len; i++) {
@@ -35,18 +35,18 @@ export class FromEventObservable<T, R> extends Observable<T> {
     subscriber.add(new Subscription(unsubscribe));
   }
 
-  _subscribe(subscriber) {
+  _subscribe(subscriber: Subscriber<T>) {
     const sourceObj = this.sourceObj;
     const eventName = this.eventName;
     const selector = this.selector;
-    let handler = selector ? (...args) => {
+    let handler = selector ? (...args: any[]) => {
       let result = tryCatch(selector)(...args);
       if (result === errorObject) {
-        subscriber.error(result.e);
+        subscriber.error(errorObject.e);
       } else {
         subscriber.next(result);
       }
-    } : (e) => subscriber.next(e);
+    } : (e: any) => subscriber.next(e);
 
     FromEventObservable.setupSubscription(sourceObj, eventName, handler, subscriber);
   }
--- a/src/observable/fromEventPattern.ts
+++ b/src/observable/fromEventPattern.ts
@@ -2,6 +2,7 @@ import {Observable} from '../Observable';
 import {Subscription} from '../Subscription';
 import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
+import {Subscriber} from '../Subscriber';
 
 export class FromEventPatternObservable<T, R> extends Observable<T> {
 
@@ -17,19 +18,19 @@ export class FromEventPatternObservable<T, R> extends Observable<T> {
     super();
   }
 
-  _subscribe(subscriber) {
+  _subscribe(subscriber: Subscriber<T>) {
     const addHandler = this.addHandler;
     const removeHandler = this.removeHandler;
     const selector = this.selector;
 
-    const handler = selector ? function(e) {
+    const handler = selector ? function(e: any) {
       let result = tryCatch(selector).apply(null, arguments);
       if (result === errorObject) {
         subscriber.error(result.e);
       } else {
         subscriber.next(result);
       }
-    } : function(e) { subscriber.next(e); };
+    } : function(e: any) { subscriber.next(e); };
 
     let result = tryCatch(addHandler)(handler);
     if (result === errorObject) {
--- a/src/operator/combineLatest.ts
+++ b/src/operator/combineLatest.ts
@@ -13,10 +13,10 @@ import {isArray} from '../util/isArray';
  * @returns {Observable} an observable of other projected values from the most recent values from each observable, or an array of each of
  * the most recent values from each observable.
  */
-export function combineLatest<R>(...observables: Array<Observable<any> |
+export function combineLatest<T, R>(...observables: Array<Observable<any> |
                                                        Array<Observable<any>> |
                                                        ((...values: Array<any>) => R)>): Observable<R> {
-  let project: (...values: Array<any>) => R =  null;
+  let project: (...values: Array<any>) => R = null;
   if (typeof observables[observables.length - 1] === 'function') {
     project = <(...values: Array<any>) => R>observables.pop();
   }
@@ -24,10 +24,10 @@ export function combineLatest<R>(...observables: Array<Observable<any> |
   // if the first and only other argument besides the resultSelector is an array
   // assume it's been called with `combineLatest([obs1, obs2, obs3], project)`
   if (observables.length === 1 && isArray(observables[0])) {
-    observables = <Array<Observable<any>>>observables[0];
+    observables = <any>observables[0];
   }
 
   observables.unshift(this);
 
-  return new ArrayObservable(observables).lift(new CombineLatestOperator(project));
+  return new ArrayObservable(observables).lift<T, R>(new CombineLatestOperator<T, R>(project));
 }
--- a/src/operator/concat-static.ts
+++ b/src/operator/concat-static.ts
@@ -11,12 +11,12 @@ import {isScheduler} from '../util/isScheduler';
  * @params {Scheduler} [scheduler] an optional scheduler to schedule each observable subscription on.
  * @returns {Observable} All values of each passed observable merged into a single observable, in order, in serial fashion.
  */
-export function concat<R>(...observables: Array<Observable<any> | Scheduler>): Observable<R> {
+export function concat<T, R>(...observables: Array<Observable<any> | Scheduler>): Observable<R> {
   let scheduler: Scheduler = null;
   let args = <any[]>observables;
   if (isScheduler(args[observables.length - 1])) {
     scheduler = args.pop();
   }
 
-  return new ArrayObservable(observables, scheduler).lift(new MergeAllOperator(1));
+  return new ArrayObservable(observables, scheduler).lift<Observable<T>, T | R>(new MergeAllOperator<T | R>(1));
 }
--- a/src/operator/concat.ts
+++ b/src/operator/concat.ts
@@ -12,7 +12,7 @@ import {MergeAllOperator} from './mergeAll-support';
  * @params {Scheduler} [scheduler] an optional scheduler to schedule each observable subscription on.
  * @returns {Observable} All values of each passed observable merged into a single observable, in order, in serial fashion.
  */
-export function concat<R>(...observables: (Observable<any> | Scheduler)[]): Observable<R> {
+export function concat<T, R>(...observables: Array<Observable<any> | Scheduler>): Observable<R> {
   let args = <any[]>observables;
   args.unshift(this);
 
@@ -21,5 +21,5 @@ export function concat<R>(...observables: (Observable<any> | Scheduler)[]): Obse
     scheduler = args.pop();
   }
 
-   return new ArrayObservable(args, scheduler).lift(new MergeAllOperator(1));
+   return new ArrayObservable(args, scheduler).lift<Observable<T>, R>(new MergeAllOperator<R>(1));
 }
--- a/src/operator/concatAll.ts
+++ b/src/operator/concatAll.ts
@@ -1,4 +1,3 @@
-import {Observable} from '../Observable';
 import {MergeAllOperator} from './mergeAll-support';
 
 /**
@@ -12,6 +11,6 @@ import {MergeAllOperator} from './mergeAll-support';
  *
  * @returns {Observable} an observable of values merged from the incoming observables.
  */
-export function concatAll<T>(): Observable<T> {
+export function concatAll<T>(): T {
   return this.lift(new MergeAllOperator(1));
 }
--- a/src/operator/concatMap.ts
+++ b/src/operator/concatMap.ts
@@ -1,5 +1,5 @@
-import {Observable} from '../Observable';
 import {MergeMapOperator} from './mergeMap-support';
+import {Observable} from '../Observable';
 
 /**
  * Maps values from the source observable into new Observables, then merges them in a serialized fashion,
@@ -11,7 +11,7 @@ import {MergeMapOperator} from './mergeMap-support';
  *
  * @param {function} project a function to map incoming values into Observables to be concatenated. accepts
  * the `value` and the `index` as arguments.
- * @param {function} [projectResult] an optional result selector that is applied to values before they're
+ * @param {function} [resultSelector] an optional result selector that is applied to values before they're
  * merged into the returned observable. The arguments passed to this function are:
  * - `outerValue`: the value that came from the source
  * - `innerValue`: the value that came from the projected Observable
@@ -20,7 +20,7 @@ import {MergeMapOperator} from './mergeMap-support';
  * @returns {Observable} an observable of values merged from the projected Observables as they were subscribed to,
  * one at a time. Optionally, these values may have been projected from a passed `projectResult` argument.
  */
-export function concatMap<T, R>(project: (value: T, index: number) => Observable<any>,
-                                projectResult?: (outerValue: T, innerValue: any, outerIndex: number, innerIndex: number) => R) {
-  return this.lift(new MergeMapOperator(project, projectResult, 1));
+export function concatMap<T, R, R2>(project: (value: T, index: number) => Observable<R>,
+                                    resultSelector?: (outerValue: T, innerValue: R, outerIndex: number, innerIndex: number) => R2) {
+  return this.lift(new MergeMapOperator(project, resultSelector, 1));
 }
--- a/src/operator/concatMapTo.ts
+++ b/src/operator/concatMapTo.ts
@@ -5,7 +5,7 @@ import {MergeMapToOperator} from './mergeMapTo-support';
  * Maps values from the source to a specific observable, and merges them together in a serialized fashion.
  *
  * @param {Observable} observable the observable to map each source value to
- * @param {function} [projectResult] an optional result selector that is applied to values before they're
+ * @param {function} [resultSelector] an optional result selector that is applied to values before they're
  * merged into the returned observable. The arguments passed to this function are:
  * - `outerValue`: the value that came from the source
  * - `innerValue`: the value that came from the projected Observable
@@ -15,9 +15,10 @@ import {MergeMapToOperator} from './mergeMapTo-support';
  * with itself, one after the other, for each value emitted from the source.
  */
 export function concatMapTo<T, R, R2>(observable: Observable<R>,
-                                      projectResult?: (outerValue: T,
-                                                       innerValue: R,
-                                                       outerIndex: number,
-                                                       innerIndex: number) => R2): Observable<R2> {
-  return this.lift(new MergeMapToOperator(observable, projectResult, 1));
+                                      resultSelector?: (
+                                                outerValue: T,
+                                                innerValue: R,
+                                                outerIndex: number,
+                                                innerIndex: number) => R2): Observable<R2> {
+  return this.lift(new MergeMapToOperator(observable, resultSelector, 1));
 }
--- a/src/operator/distinctUntilChanged.ts
+++ b/src/operator/distinctUntilChanged.ts
@@ -2,6 +2,7 @@ import {Operator} from '../Operator';
 import {Subscriber} from '../Subscriber';
 import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
+import {Observable} from '../Observable';
 
 /**
  * Returns an Observable that emits all items emitted by the source Observable that are distinct by comparison from the previous item.
@@ -10,13 +11,15 @@ import {errorObject} from '../util/errorObject';
  * @param {function} [compare] optional comparison function called to test if an item is distinct from the previous item in the source.
  * @returns {Observable} an Observable that emits items from the source Observable with distinct values.
  */
-export function distinctUntilChanged<T>(compare?: (x: any, y: any) => boolean, keySelector?: (x: T) => any) {
-  return this.lift(new DistinctUntilChangedOperator(compare, keySelector));
+export function distinctUntilChanged<T>(compare?: (x: T, y: T) => boolean): Observable<T>;
+export function distinctUntilChanged<T, K>(compare: (x: K, y: K) => boolean, keySelector?: (x: T) => K): Observable<T>;
+export function distinctUntilChanged<T, K>(compare: (x: K, y: K) => boolean, keySelector?: (x: T) => K): Observable<T> {
+  return this.lift(new DistinctUntilChangedOperator<T, K>(compare, keySelector));
 }
 
-class DistinctUntilChangedOperator<T, R> implements Operator<T, R> {
-  constructor(private compare: (x: any, y: any) => boolean,
-              private keySelector: (x: T) => any) {
+class DistinctUntilChangedOperator<T, K> implements Operator<T, T> {
+  constructor(private compare: (x: K, y: K) => boolean,
+              private keySelector: (x: T) => K) {
   }
 
   call(subscriber: Subscriber<T>): Subscriber<T> {
@@ -24,13 +27,13 @@ class DistinctUntilChangedOperator<T, R> implements Operator<T, R> {
   }
 }
 
-class DistinctUntilChangedSubscriber<T> extends Subscriber<T> {
-  private key: any;
+class DistinctUntilChangedSubscriber<T, K> extends Subscriber<T> {
+  private key: K;
   private hasKey: boolean = false;
 
   constructor(destination: Subscriber<T>,
-              compare: (x: any, y: any) => boolean,
-              private keySelector: (x: T) => any) {
+              compare: (x: K, y: K) => boolean,
+              private keySelector: (x: T) => K) {
     super(destination);
     if (typeof compare === 'function') {
       this.compare = compare;
--- a/src/operator/every.ts
+++ b/src/operator/every.ts
@@ -17,10 +17,8 @@ import {errorObject} from '../util/errorObject';
 export function every<T>(predicate: (value: T, index: number, source: Observable<T>) => boolean,
                          thisArg?: any): Observable<boolean> {
   const source = this;
-  let result;
-
   if (source._isScalar) {
-    result = tryCatch(predicate).call(thisArg || this, source.value, 0, source);
+    const result: boolean = tryCatch(predicate).call(thisArg || this, source.value, 0, source);
     if (result === errorObject) {
       return new ErrorObservable(errorObject.e, source.scheduler);
     } else {
@@ -30,7 +28,8 @@ export function every<T>(predicate: (value: T, index: number, source: Observable
 
   if (source instanceof ArrayObservable) {
     const array = (<ArrayObservable<T>>source).array;
-    let result = tryCatch((array, predicate, thisArg) => array.every(<any>predicate, thisArg))(array, predicate, thisArg);
+    const result = tryCatch((array: T[], predicate: (value: T, index: number, source: Observable<T>) => boolean, thisArg: any) =>
+                                    array.every(<any>predicate, thisArg))(array, predicate, thisArg);
     if (result === errorObject) {
       return new ErrorObservable(errorObject.e, source.scheduler);
     } else {
--- a/src/operator/exhaustMap.ts
+++ b/src/operator/exhaustMap.ts
@@ -14,22 +14,20 @@ import {subscribeToResult} from '../util/subscribeToResult';
  * @returns {Observable} an Observable containing all the projected Observables of each item of the source concatenated together.
  */
 export function exhaustMap<T, R, R2>(project: (value: T, index: number) => Observable<R>,
-                                     resultSelector?: (outerValue: T,
-                                                       innerValue: R,
-                                                       outerIndex: number,
-                                                       innerIndex: number) => R2): Observable<R> {
+                                     resultSelector?: (
+                                            outerValue: T,
+                                            innerValue: R,
+                                            outerIndex: number,
+                                            innerIndex: number) => R2): Observable<R2> {
   return this.lift(new SwitchFirstMapOperator(project, resultSelector));
 }
 
-class SwitchFirstMapOperator<T, R, R2> implements Operator<T, R> {
+class SwitchFirstMapOperator<T, R, R2> implements Operator<T, R2> {
   constructor(private project: (value: T, index: number) => Observable<R>,
-              private resultSelector?: (outerValue: T,
-                                        innerValue: R,
-                                        outerIndex: number,
-                                        innerIndex: number) => R2) {
+              private resultSelector?: (outerValue: T, innerValue: R, outerIndex: number, innerIndex: number) => R2) {
   }
 
-  call(subscriber: Subscriber<R>): Subscriber<T> {
+  call(subscriber: Subscriber<R2>): Subscriber<T> {
     return new SwitchFirstMapSubscriber(subscriber, this.project, this.resultSelector);
   }
 }
@@ -39,7 +37,7 @@ class SwitchFirstMapSubscriber<T, R, R2> extends OuterSubscriber<T, R> {
   private hasCompleted: boolean = false;
   private index: number = 0;
 
-  constructor(destination: Subscriber<R>,
+  constructor(destination: Subscriber<R2>,
               private project: (value: T, index: number) => Observable<R>,
               private resultSelector?: (outerValue: T, innerValue: R, outerIndex: number, innerIndex: number) => R2) {
     super(destination);
@@ -51,7 +49,7 @@ class SwitchFirstMapSubscriber<T, R, R2> extends OuterSubscriber<T, R> {
       const destination = this.destination;
       let result = tryCatch(this.project)(value, index);
       if (result === errorObject) {
-        destination.error(result.e);
+        destination.error(errorObject.e);
       } else {
         this.hasSubscription = true;
         this.add(subscribeToResult(this, result, value, index));
--- a/src/operator/find-support.ts
+++ b/src/operator/find-support.ts
@@ -5,7 +5,7 @@ import {Subscriber} from '../Subscriber';
 import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
 
-export class FindValueOperator<T, R> implements Operator<T, R> {
+export class FindValueOperator<T> implements Operator<T, T> {
   constructor(private predicate: (value: T, index: number, source: Observable<T>) => boolean,
               private source: Observable<T>,
               private yieldIndex: boolean,
--- a/src/operator/first.ts
+++ b/src/operator/first.ts
@@ -13,7 +13,7 @@ import {EmptyError} from '../util/EmptyError';
  */
 export function first<T, R>(predicate?: (value: T, index: number, source: Observable<T>) => boolean,
                             resultSelector?: (value: T, index: number) => R,
-                            defaultValue?: any): Observable<T> | Observable<R> {
+                            defaultValue?: R): Observable<T> | Observable<R> {
   return this.lift(new FirstOperator(predicate, resultSelector, defaultValue, this));
 }
 
--- a/src/operator/last.ts
+++ b/src/operator/last.ts
@@ -7,7 +7,7 @@ import {EmptyError} from '../util/EmptyError';
 
 export function last<T, R>(predicate?: (value: T, index: number, source: Observable<T>) => boolean,
                            resultSelector?: (value: T, index: number) => R,
-                           defaultValue?: any): Observable<T> | Observable<R> {
+                           defaultValue?: R): Observable<T> | Observable<R> {
   return this.lift(new LastOperator(predicate, resultSelector, defaultValue, this));
 }
 
@@ -24,7 +24,7 @@ class LastOperator<T, R> implements Operator<T, R> {
 }
 
 class LastSubscriber<T, R> extends Subscriber<T> {
-  private lastValue: T;
+  private lastValue: T | R;
   private hasValue: boolean = false;
   private index: number = 0;
 
--- a/src/operator/mergeAll.ts
+++ b/src/operator/mergeAll.ts
@@ -1,6 +1,5 @@
-import {Observable} from '../Observable';
 import {MergeAllOperator} from './mergeAll-support';
 
-export function mergeAll<R>(concurrent: number = Number.POSITIVE_INFINITY): Observable<R> {
+export function mergeAll<T>(concurrent: number = Number.POSITIVE_INFINITY): T {
   return this.lift(new MergeAllOperator(concurrent));
 }
--- a/src/operator/mergeMap-support.ts
+++ b/src/operator/mergeMap-support.ts
@@ -39,7 +39,7 @@ export class MergeMapSubscriber<T, R, R2> extends OuterSubscriber<T, R> {
       const ish = tryCatch(this.project)(value, index);
       const destination = this.destination;
       if (ish === errorObject) {
-        destination.error(ish.e);
+        destination.error(errorObject.e);
       } else {
         this.active++;
         this._innerSub(ish, value, index);
--- a/src/operator/mergeMap.ts
+++ b/src/operator/mergeMap.ts
@@ -2,10 +2,7 @@ import {Observable} from '../Observable';
 import {MergeMapOperator} from './mergeMap-support';
 
 export function mergeMap<T, R, R2>(project: (value: T, index: number) => Observable<R>,
-                                   resultSelector?: (outerValue: T,
-                                                     innerValue: R,
-                                                     outerIndex: number,
-                                                     innerIndex: number) => R,
-                                   concurrent: number = Number.POSITIVE_INFINITY) {
-  return this.lift(new MergeMapOperator(project, resultSelector, concurrent));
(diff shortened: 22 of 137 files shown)


kwonoj:
Appreciate for effort :) few copy-paste from original PR

- would you able to align commit message as same as others, starts with no-capital
- npm script `package.json` should be updated as well as `tsconfig.json`

I'll try to come up with possible cases of `suppressImplicitAnyIndexErrors` in our codebase meanwhile..

david-driscoll:
done! :smile: 

* ~~would you able to align commit message as same as others, starts with no-capital~~
* ~~npm script package.json should be updated as well as tsconfig.json~~

Build is erroring for `suppressImplicitAnyIndexErrors` in a few places (that I wasn't seeing, which is strange but not super important here).

Do we want to suppress those messages, or define the types so they don't happen?

kwonoj:
> Build is erroring for suppressImplicitAnyIndexErrors in a few places 

: Yes, those are errors I could see while I work on this, reason of suggestion for enabling `suppressImplicitAnyIndexErrors`. I don't see lot of values to enforce type for those indexers in our codebases, i.e `if (observerOrNext[rxSubscriber]) {`, `ish[SymbolShim.observable]`, etcs.. (but would like to hear your opinion also if you think it'd benefit to have type for those)

david-driscoll:
I don't see much value in typing them, there aren't a lot of cases where Rx is using objects as a dictionary or similar (not when there is map, etc).

I will update to suppress those messages then.

src/CoreOperators.ts
```diff
@@ -35,9 +35,9 @@ export interface CoreOperators<T> {
                 projectResult?: (x: T, y: any, ix: number, iy: number) => R,
                 concurrent?: number) => Observable<R>;
   flatMapTo?: <R>(observable: Observable<any>, projectResult?: (x: T, y: any, ix: number, iy: number) => R, concurrent?: number) => Observable<R>;
-  groupBy?: <R>(keySelector: (value: T) => string,
+  groupBy?: <TKey, R>(keySelector: (value: T) => string,
```
luisgabriel: As I suggested in #1050, I think it would be better to rename `TKey` to `K` here.

src/Observable.ts
```diff
@@ -216,9 +216,9 @@ export class Observable<T> implements CoreOperators<T>  {
                projectResult?: (x: T, y: any, ix: number, iy: number) => R,
                concurrent?: number) => Observable<R>;
   flatMapTo: <R>(observable: Observable<any>, projectResult?: (x: T, y: any, ix: number, iy: number) => R, concurrent?: number) => Observable<R>;
-  groupBy: <R>(keySelector: (value: T) => string,
+  groupBy: <TKey, R>(keySelector: (value: T) => string,
```
luisgabriel: Same renaming here.

src/CoreOperators.ts
david-driscoll: It's the C# dev in me coming out. :smiling_imp: 

I don't care about naming semantics, updating the commit.

src/Subject.ts
```diff
@@ -63,11 +63,11 @@ export class Subject<T> extends Observable<T> implements Observer<T>, Subscripti
     return new SubjectSubscription(this, subscriber);
   }
 
-  add(subscription?) {
+  add(subscription?: Subscription<T>) {
```
luisgabriel: You may need to update this if #1065 gets merged first.

src/Observable.ts
```diff
@@ -21,7 +21,7 @@ import {rxSubscriber} from'./symbol/rxSubscriber';
 export class Observable<T> implements CoreOperators<T>  {
   source: Observable<any>;
   operator: Operator<any, T>;
-  _isScalar: boolean = false;
+  _isScalar = false;
```
kwonoj: I would like this to be discussed as separate. Lot of analysis tools or editor recommends this pattern (to not explicitly specify type if it's obvious by assigned value) but it's currently common in our codebases. Once we decide which way to go, single PR change it once would works.

src/observable/ScalarObservable.ts
```diff
@@ -80,8 +80,8 @@ proto.reduce = function <T, R>(project: (acc: R, x: T) => R, seed?: R): Observab
     return <any>this;
   }
   let result = tryCatch(project)(seed, this.value);
-  if (result === errorObject) {
-    return new ErrorObservable(errorObject.e);
+  if (result as any === errorObject) {
```
luisgabriel: I see that some places (as the change above in this one in this diff) you are not casting `result` to `any`. Can we drop this one?

src/add/operator/exhaust.ts
```diff
@@ -2,4 +2,4 @@ import {Observable} from '../../Observable';
 import {exhaust} from '../../operator/exhaust';
 Observable.prototype.exhaust = exhaust;
 
-export var _void: void;
\ No newline at end of file
+export var _void: void;
```
kwonoj: is this file has changed?

src/observable/ScalarObservable.ts
kwonoj: specific reason to use `as` casting in here only?

david-driscoll: in some cases result is already any, in the cases where it isn't, we have to.

image

We could change `errorObject` to just be type any, and then this goes away.

src/add/operator/exhaust.ts
luisgabriel: I think it's line ending related. There are other files with the same change.

david-driscoll: Hrm... I'll see if I can figure out to avoid that line ending issue.

src/observable/forkJoin.ts
```diff
@@ -7,16 +7,15 @@ import {isArray} from '../util/isArray';
 
 export class ForkJoinObservable<T> extends Observable<T> {
   constructor(private sources: Array<Observable<any> | Promise<any>>,
-              private resultSelector?: (...values: Array<any>) => any) {
+              private resultSelector?: (...values: Array<any>) => T) {
```
kwonoj: does resultSelector always return `T`? what about cases like below snippet?

```
Rx.Observable.forkJoin(
  Rx.Observable.just(42),
  Rx.Observable.just(56),
  function (x, y) { return x.toString() + y.toString(); }
);
```

src/operator/combineLatest.ts
```diff
@@ -13,10 +13,10 @@ import {isArray} from '../util/isArray';
  * @returns {Observable} an observable of other projected values from the most recent values from each observable, or an array of each of
  * the most recent values from each observable.
  */
-export function combineLatest<R>(...observables: Array<Observable<any> |
+export function combineLatest<T, R>(...observables: Array<any | Observable<any> |
```
kwonoj: does combinelatest allowed to accept other than `Observable` and `Array<Observable>`?

src/add/operator/exhaust.ts
david-driscoll: Is there a standard with no newlines at the end of files for the project, or was just that how things happened to end up?  (Just want to get it right! :) )

src/operator/concat-static.ts
```diff
@@ -12,12 +12,12 @@ import {isScheduler} from '../util/isScheduler';
  * @params {Scheduler} [scheduler] an optional scheduler to schedule each observable subscription on.
  * @returns {Observable} All values of each passed observable merged into a single observable, in order, in serial fashion.
  */
-export function concat<R>(...observables: Array<Observable<any> | Scheduler>): Observable<R> {
+export function concat<T, R>(...observables: Array<Observable<any> | Scheduler>): Observable<T> {
```
kwonoj: What'll happen if 2 source has different types like

```
var source1 = Rx.Observable.of(42);
var source2 = Rx.Observable.of('b');

var source = Rx.Observable.concat(source1, source2);
```

and try to subscribe into `source`?

src/observable/ScalarObservable.ts
luisgabriel: You mean to type [this](https://github.com/ReactiveX/RxJS/blob/master/src/util/errorObject.ts) declaration of `errorObject` to `any`? If this is the case I don't see any problem. What do you think @kwonoj?

src/operator/concat.ts
```diff
@@ -10,9 +10,9 @@ import {MergeAllOperator} from './mergeAll-support';
  * on to the next.
  * @params {...Observable} the observables to concatenate
  * @params {Scheduler} [scheduler] an optional scheduler to schedule each observable subscription on.
- * @returns {Observable} All values of each passed observable merged into a single observable, in order, in serial fashion.
+* @returns {Observable} All values of each passed observable merged into a single observable, in order, in serial fashion.
  */
-export function concat<R>(...observables: (Observable<any> | Scheduler)[]): Observable<R> {
+export function concat<T, R>(...observables: Array<Observable<any> | Scheduler>): Observable<R> {
```
kwonoj: This line https://github.com/ReactiveX/RxJS/pull/1077/files#diff-261ac0fcd1adcd646e3747b9b8bf73f0R15 does 
`concat<T, R>(...observables: Array<Observable<any> | Scheduler>): Observable<T>`

returns `Observable<T>` while this returns `Observable<R>`, is there any differences?

src/operator/concatAll.ts
```diff
@@ -12,6 +11,6 @@ import {MergeAllOperator} from './mergeAll-support';
  *
  * @returns {Observable} an observable of values merged from the incoming observables.
  */
-export function concatAll<T>(): Observable<T> {
+export function concatAll<T>(): T {
```
kwonoj: Doesn't this returns `Observable`?

src/add/operator/exhaust.ts
luisgabriel: I'm not sure, but I think it's a bit messy at the moment.

src/operator/expand.ts
```diff
@@ -2,7 +2,7 @@ import {Observable} from '../Observable';
 import {Scheduler} from '../Scheduler';
 import {ExpandOperator} from './expand-support';
 
-export function expand<T, R>(project: (value: T, index: number) => Observable<R>,
+export function expand<T, R>(project: (value: T, index: number) => Observable<R> | Promise<R>,
```
kwonoj: do we accept projection returns `promise`? It seems implementation does not checks if returned is promise and does not trying to convert it internally but directly subscribe.

src/operator/first.ts
```diff
@@ -5,14 +5,14 @@ import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
 import {EmptyError} from '../util/EmptyError';
 
-export function first<T, R>(predicate?: (value: T, index: number, source: Observable<T>) => boolean,
+export function first<T, R>(predicate?: (value: T, index: number, observable: Observable<T>) => boolean,
                             resultSelector?: (value: T, index: number) => R,
-                            defaultValue?: any): Observable<T> | Observable<R> {
+                            defaultValue?: R): Observable<R> {
```
kwonoj: what happens if 

`source.first()` 

doesn't have resultSelector also doesn't have defaultValue but expect first element from source, which is `T`?

src/operator/last.ts
```diff
@@ -5,14 +5,14 @@ import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
 import {EmptyError} from '../util/EmptyError';
 
-export function last<T, R>(predicate?: (value: T, index: number, source: Observable<T>) => boolean,
+export function last<T, R>(predicate?: (value: T, index: number, observable: Observable<T>) => boolean,
                            resultSelector?: (value: T, index: number) => R,
-                           defaultValue?: any): Observable<T> | Observable<R> {
+                           defaultValue?: R): Observable<R> {
```
kwonoj: same as `first` at https://github.com/ReactiveX/RxJS/pull/1077/files#diff-abf64ae32e96d927f534f6199ca88bc8R10

src/operator/mergeAll.ts
```diff
@@ -1,6 +1,5 @@
-import {Observable} from '../Observable';
 import {MergeAllOperator} from './mergeAll-support';
 
-export function mergeAll<R>(concurrent: number = Number.POSITIVE_INFINITY): Observable<R> {
+export function mergeAll<T>(concurrent: number = Number.POSITIVE_INFINITY): T {
```
kwonoj: this one does not return `Observable`?

src/operator/mergeMap-support.ts
```diff
@@ -7,9 +7,9 @@ import {errorObject} from '../util/errorObject';
 import {subscribeToResult} from '../util/subscribeToResult';
 import {OuterSubscriber} from '../OuterSubscriber';
 
-export class MergeMapOperator<T, R, R2> implements Operator<T, R> {
-  constructor(private project: (value: T, index: number) => Observable<R>,
-              private resultSelector?: (outerValue: T, innerValue: R, outerIndex: number, innerIndex: number) => R2,
+export class MergeMapOperator<T, R, TResult> implements Operator<T, R> {
+  constructor(private project: (value: T, index: number) => Observable<R> | Promise<R>,
```
kwonoj: same as `expand`, https://github.com/ReactiveX/RxJS/pull/1077/files#diff-5e32168db6f2f1aea024c1cfcae156f0R5

src/add/operator/exhaust.ts
kwonoj: I don't think we currently have strict rules for those. Personally I clean up endings if I could find it in my changes, but in this case if file hasn't changed it'd better to unstage simply.

src/observable/ScalarObservable.ts
kwonoj: first, my question was simply if there reason to choose between below :)

`result as any` vs `<any>result`

for typeof errorObject, not sure at this moment..since it's literally kind of  `any` object, so it could be typeof `any` explicitly though.

kwonoj:
I'm seeing some of function signatures are changed not related to implicit any, specifically `projection` function definition accepts to return `Promise`. Most of current operator implementation as far as I recall it does not checks projected result is promise or observable. I think this is kind of out-of scope changes requires discussion (will we support those behavior or not?) would like to suggest to not to change those signatures in this PR and create issue if it's necessary.

src/operator/sample.ts
```diff
@@ -19,7 +19,7 @@ class SampleSubscriber<T> extends Subscriber<T> {
   private lastValue: T;
   private hasValue: boolean = false;
 
-  constructor(destination: Subscriber<T>, private notifier: Observable<any>) {
+  constructor(destination: Subscriber<T>, private notifier: any) {
```
kwonoj: notifier can be other type than `observable`?

src/operator/switch.ts
```diff
@@ -1,11 +1,10 @@
 import {Operator} from '../Operator';
-import {Observable} from '../Observable';
 import {Subscriber} from '../Subscriber';
 import {Subscription} from '../Subscription';
 import {OuterSubscriber} from '../OuterSubscriber';
 import {subscribeToResult} from '../util/subscribeToResult';
 
-export function _switch<T>(): Observable<T> {
+export function _switch<T>(): T {
```
kwonoj: This operator does not returns `observable`?

src/add/operator/exhaust.ts
david-driscoll: I see the editor config, I'll fix that up.

src/observable/forkJoin.ts
david-driscoll: `T` isn't used anywhere in the sources of the selector, to `T` is really just the result.  In this case it would return a string based on the code snippet.

src/util/MapPolyfill.ts
```diff
@@ -20,7 +20,7 @@ export class MapPolyfill {
     return this;
   }
 
-  delete(key): boolean {
+  delete(key: any) {
```
kwonoj: is there reason to remove return type?

src/util/MapPolyfill.ts
```diff
@@ -29,9 +29,9 @@ export class MapPolyfill {
     return true;
   }
 
-  forEach(cb, thisArg): void {
+  forEach(cb: any, thisArg: any) {
```
kwonoj: assume `cb` can be `Function`, what do you think?

src/util/isDate.ts
```diff
@@ -1,3 +1,3 @@
-export function isDate(value) {
+export function isDate(value: any): value is Date {
```
kwonoj: return type `value is Date` is actually neat, I didn't aware of this. Thanks for bringing this :)

src/operator/concat-static.ts
david-driscoll: Yep, that's wrong, it should be `R`.

src/operator/concatAll.ts
david-driscoll: `concatAll`/`mergeAll` all assume that you're already dealing with `Observable<Observable<T>>`, so calling `concatAll` actually returns `T` since `T` is `Observable<S>`

src/operator/first.ts
david-driscoll: You introduce an additional overload for the method that only includes `T`, but I left those changes out of the first run as it was complicating matters.  Once we get `noImplictAny` and the other type changes in I want to go around and make sure the interfaces get a makeover.

david-driscoll:
They are technically supported today, because all of the methods call through to `subscribeToResult` which handles everything under the sun.

src/operator/switch.ts
david-driscoll: same as concat/merge, the type already assumes that T is `Observable<S>`

src/util/MapPolyfill.ts
david-driscoll: It's implicit based on the code

src/util/MapPolyfill.ts
david-driscoll: Yeah, missed that.

src/util/isDate.ts
david-driscoll: It is very useful!

src/operator/concatAll.ts
kwonoj: Makes sense, but just bit concerned if someone just read signature might came to think in similar way as I thought originally. Does specifying return type to explicit `Observable<T>` could cause side effect?

src/operator/first.ts
kwonoj: Sorry, seems I'm little bit confused.. My thought in here is return type could remain as-is `: Observable<T> | Observable<R>` since it's already satisfies noimplicitAny conditions, and if this signature change's required than could brought up with separate PR/issue - do I have some incorrect understandings?

kwonoj:
Correct in sense of availability, but still I'd like to suggest it to take out as separate discussion if we want to support those as interface. As same as original PR, thinking about size of this changes limiting scope would be better, or discussion would be expanded.

src/util/MapPolyfill.ts
kwonoj: Yes it does, but most of cases we're adding explicit return types but in here was removing. Was curious specific reason to remove which already exists.

src/util/isDate.ts
kwonoj: 100% agree :+1:

src/operator/first.ts
david-driscoll: fair enough, I can change it back.

src/add/operator/exhaust.ts
kwonoj: For this issue, I'll bring up small PR to introduce consistency.

src/operator/switch.ts
kwonoj: Yes, now can sense that. Let's continue discussion on https://github.com/ReactiveX/RxJS/pull/1077#discussion_r47714337.

src/observable/ScalarObservable.ts
luisgabriel: Oh, I prefer `result as any` but I think `<any>result` is more used throughout the codebase.

Regarding the use of casting in this case, changing the signature of [tryCatch()](https://github.com/ReactiveX/RxJS/blob/master/src/util/tryCatch.ts) to return something like `(...args: any[]) => any` instead of just `Function` solves the problem?

kwonoj: or like `=> Function | errorObject` maybe? (just throwing out thoughts, not sure if it works)

for casting syntax, since I was living in C# prefer `as` in usual cases.. but in here would like to go with consistency in single style. Does not feel to introduce different styles while original style serves.

david-driscoll: we could create `isErrorObject(result: any): result is typeof errorObject {...}` but that helper method may have performance implications, and I know that is a concern.

kwonoj: Let's not touch `errorObject` for now to avoid expansion to this changes.

src/operator/concatAll.ts
david-driscoll: what we're really looking for is something like

```typescript
concatAll<T extends Observable<InnerT>, InnerT>(): Observable<InnerT>
```

which sadly isn't supported with typescript, and we might have to get away with just docs here.  If we make it `Observable<T>` then true return type would show up as `Observable<Observable<T>>`

src/observable/ScalarObservable.ts
luisgabriel: I think my suggestion about changing the `tryCatch()` signature is less invasive. What do you think, @david-driscoll?

kwonoj: to be clear, my suggestion above to 'not touch' meant do not modify object could bring side effect. as long as it's type changes only, I think it's fine.

src/operator/concatAll.ts
kwonoj: Understood, thanks for clarification.

src/util/MapPolyfill.ts
david-driscoll: I don't remember explicitly removing it, I think it was added recently.  I just didn't combine it in the merge when I saw it.

src/observable/ScalarObservable.ts
david-driscoll: @luisgabriel could do the problem actually came into play because I improved the definition for `tryCatch` :smile: 

`export function tryCatch<T extends Function>(fn: T): T`, so it knows the type of the function it will flow through to the invocation, which gives you type safety there instead of just pretending it's a function.

I'll add in the errorObject type change, it is probably small enough that it shouldn't be a big deal.

src/util/MapPolyfill.ts
kwonoj: Ah yes, seems you're correct. It's been added recently after this PR.. Please just rebase to reflect latest status then. :) Thanks for pointing.

src/observable/ScalarObservable.ts
kwonoj: > I'll add in the errorObject type change, it is probably small enough that it shouldn't be a big deal.

: do you mean bringing change https://github.com/ReactiveX/RxJS/pull/1077#discussion_r47718354 mentioned in here?

src/add/operator/exhaust.ts
david-driscoll: I've gone and corrected the line endings that looked pretty obvious, it looks like some files have them and some don't so it's hard to know if it should or shouldn't lol

kwonoj: > some don't so it's hard to know if it should or shouldn't lol

: Yes, agreed. As commented above let me amend those. :)

src/util/errorObject.ts
```diff
@@ -1 +1,2 @@
-export var errorObject = { e: {} };
\ No newline at end of file
+// Any so that it we don't have to cast when comparing it's values.
+export var errorObject: any = { e: {} };
```
david-driscoll: @kwonoj @luisgabriel this is the change I was talking about.

The object is the same, but the type is now just `any`.  I've stripped out any of the remaining `as any` casts as well.

kwonoj: I think that's ok for this PR.

luisgabriel: :+1:

src/util/tryCatch.ts
```diff
@@ -11,7 +11,7 @@ function tryCatcher(): any {
   }
 }
 
-export function tryCatch(fn: Function): Function {
+export function tryCatch<T extends Function>(fn: T): T {
   tryCatchTarget = fn;
-  return tryCatcher;
-};
\ No newline at end of file
+  return <any>tryCatcher;
```
luisgabriel: `tryCatcher` is a function and `tryCatch` returns a function. Why are doing this casting here?

src/util/isArray.ts
```diff
@@ -1 +1 @@
-export const isArray = Array.isArray || (x => x && typeof x.length === 'number');
\ No newline at end of file
+export const isArray = Array.isArray || (<T>(x: any): x is T[] => x && typeof x.length === 'number');
```
luisgabriel: `<T>(x: any): x is T[] => x` is a bit confusing. Do we need all this complexity?

Isn't just `(x: any) => ...` enough?

david-driscoll: `x is T[]` lets the compiler know that if the method returns true, than the `typeof x` is `T[]`

src/operator/zip-support.ts
```diff
@@ -221,7 +221,7 @@ class ZipBufferIterator<T, R> extends OuterSubscriber<T, R> implements LookAhead
     }
   }
 
-  notifyNext(outerValue, innerValue, outerIndex, innerIndex) {
+  notifyNext(outerValue: any, innerValue: any, outerIndex: any, innerIndex: any) {
```
luisgabriel: Can we say that `outerIndex` and `innerIndex` are `number`?

src/operator/zip-support.ts
```diff
@@ -56,8 +56,8 @@ export class ZipSubscriber<T, R> extends Subscriber<T> {
     this.active = len;
     for (let i = 0; i < len; i++) {
       let iterator = iterators[i];
-      if (iterator.stillUnsubscribed) {
-        iterator.subscribe(iterator, i);
+      if ((iterator as any).stillUnsubscribed) {
+        (iterator as any).subscribe(iterator, i);
```
luisgabriel:
```typescript
let iterator: Iterator<T> = iterators[i];
if (iterator.stillUnsubscribed) {
  iteratorsubscribe(iterator, i);
} else {
...
```

Could we do this?

src/operator/zip-support.ts
```diff
@@ -27,7 +27,7 @@ export class ZipSubscriber<T, R> extends Subscriber<T> {
   private index = 0;
   private values: any;
   private project: (...values: Array<any>) => R;
-  private iterators = [];
+  private iterators: LookAheadIterator<any>[] = [];
```
luisgabriel: What about `Iterator<T>[]` instead of `LookAheadIterator<any>[]`?

src/util/tryCatch.ts
david-driscoll: Compiler returns `Type '() => any' is not assignable to type 'T'.` without it.

With `<T extends Function>` though we can see much nicer behavior when using the method.

Given

```
var predicate: (value: T, index: number, observable: Observable<T>) => boolean;
var tryCatchMethod: tryCatch(predicate);
```
The `typeof tryCatchMethod` will then be the same as `typeof predicate`.

src/operator/zip-support.ts
david-driscoll: `stillUnsubscribed` is a member of `ZipBufferIterator<,>` but not `StaticArrayIterator` or `StaticIterator`.  I could add it on as a member (just typed) then it would probably work without the cast.

src/operator/zip-support.ts
david-driscoll: `LookAheadIterator` defines `hasValue` and `hasCompleted`, used elsewhere with iterator values.

david-driscoll: Oh and for typing it to `T` or `R` causes casting issues later down the line.

src/util/tryCatch.ts
luisgabriel: Hmm. I see. But strictly speaking, `typeof tryCatchMethod` should not be the same `as typeof predicate` since the return type of `tryCatchMethod` is `boolean | any`, am I right?

david-driscoll: no the cast to `any` just means we're telling the compiler assume this value is whatever, then the return type will be whatever the function is defined as, this case it's `T`

src/util/isArray.ts
luisgabriel: Got it. My main concern is readability. But it seems a good tradeoff to keep this typing here.

src/util/tryCatch.ts
luisgabriel: I know. What I'm saying is that `typeof tryCatchMethod` should not be the same as `typeof predicate`. The return type of these functions is different.

src/operator/zip-support.ts
luisgabriel: Ok.

src/util/tryCatch.ts
luisgabriel: `predicate` returns `boolean`.
`tryCatchMethod` returns `boolean | any`.

david-driscoll: but the way we're using it, we want `tryCatch` to behave exactly the same as whatever method we give it, return values and everything.  Yes it can return the `errorObject` but there is no way to amend the return type of a method, so it's just a trade off.

It's useful because if for example we decide to add or remove a parameter to something like 

```
takeWhile<T>(predicate: (value: T, index: number) => boolean
```

Then the following method would have compiler errors
```
const result = tryCatch(this.predicate)(value, this.index++);
```

kwonoj:
I think changes are now good to be checked in. Is there additional suggestion @luisgabriel ?

david-driscoll:
:balloon: :tada: :balloon:

kwonoj:
Also cc @blesh to see if there's any suggestions around. @blesh , I think changes are good now and waiting @luisgabriel 's response if there's additional suggestion from him. If there isn't, I'll check this in around tomorrow. Meanwhile if you have any suggestions, would be appreciate to have it. :)

david-driscoll:
I'll also rebase if #1065 goes in first, as it will probably have a small
conflict.

src/util/tryCatch.ts
luisgabriel: Hmm. Got it. Maybe this trade-off is worthwhile.

luisgabriel:
@kwonoj I'll give another look. But I think it's good to go. ;)

Very good work, @david-driscoll! My suggestion to future contributions is to do the work incrementally with more self-contained commits/PRs. It can speed up a lot the review process. ;)

kwonoj:
Cool, thanks @luisgabriel 

> work incrementally with more self-contained commits/PRs

: I agree to this in general, but I also think this was bit exceptional case since enabling compiler flag will require change all file at once. :)

kwonoj:
@david-driscoll , would you able to rebase this PR? #1065 landed first..

david-driscoll:
I can later this evening
On Wed, Dec 16, 2015 at 6:24 PM OJ Kwon <notifications@github.com> wrote:

> @david-driscoll <https://github.com/david-driscoll> , would you able to
> rebase this PR? #1065 <https://github.com/ReactiveX/RxJS/pull/1065>
> landed first..
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ReactiveX/RxJS/pull/1077#issuecomment-165282607>.
>
-- 
Thanks,
David

chicoxyzzy:
such a huge work!

david-driscoll:
@kwonoj rebased.

src/observable/fromEvent.ts
```diff
@@ -35,18 +35,18 @@ export class FromEventObservable<T, R> extends Observable<T> {
...
+    let handler = selector ? (e: any) => {
       let result = tryCatch(selector)(e);
       if (result === errorObject) {
-        subscriber.error(result.e);
+        subscriber.error(errorObject.e);
       } else {
         subscriber.next(result);
       }
-    } : (e) => subscriber.next(e);
+  } : (e: any) => subscriber.next(e);
```
luisgabriel: I think the indentation level is wrong here.

src/observable/fromEventPattern.ts
```diff
@@ -17,19 +18,19 @@ export class FromEventPatternObservable<T, R> extends Observable<T> {
...
-    const handler = selector ? function(e) {
+    const handler = selector ? function(e: any) {
       let result = tryCatch(selector).apply(null, arguments);
       if (result === errorObject) {
         subscriber.error(result.e);
       } else {
         subscriber.next(result);
       }
-    } : function(e) { subscriber.next(e); };
+  } : function(e: any) { subscriber.next(e); };
```
luisgabriel: Same here.

src/operator/concat.ts
```diff
@@ -10,9 +10,9 @@ import {MergeAllOperator} from './mergeAll-support';
  * on to the next.
  * @params {...Observable} the observables to concatenate
  * @params {Scheduler} [scheduler] an optional scheduler to schedule each observable subscription on.
- * @returns {Observable} All values of each passed observable merged into a single observable, in order, in serial fashion.
+* @returns {Observable} All values of each passed observable merged into a single observable, in order, in serial fashion.
```
luisgabriel: Same here.

src/operator/every.ts
```diff
@@ -8,15 +8,13 @@ import {Subscriber} from '../Subscriber';
...
 
-export function every<T>(predicate: (value: T, index: number, source: Observable<T>) => boolean,
+export function every<T>(predicate: (value: T, index: number, observable: Observable<T>) => boolean,
                          thisArg?: any): Observable<boolean> {
   const source = this;
-  let result;
-
   if (source._isScalar) {
-    result = tryCatch(predicate).call(thisArg || this, source.value, 0, source);
+    let result: boolean = tryCatch(predicate).call(thisArg || this, source.value, 0, source);
```
luisgabriel: Since you moved this declaration, I think you can use `const` instead of `let` here.

src/operator/every.ts
```diff
@@ -24,9 +22,10 @@ export function every<T>(predicate: (value: T, index: number, source: Observable
 
   if (source instanceof ArrayObservable) {
     const array = (<ArrayObservable<T>>source).array;
-    let result = tryCatch((array, predicate, thisArg) => array.every(<any>predicate, thisArg))(array, predicate, thisArg);
+    let result = tryCatch((array: T[], predicate: (value: T, index: number, observable: Observable<T>) => boolean, thisArg: any) =>
```
luisgabriel: Same here.

src/operator/exhaustMap.ts
```diff
@@ -6,35 +6,33 @@ import {errorObject} from '../util/errorObject';
 import {OuterSubscriber} from '../OuterSubscriber';
 import {subscribeToResult} from '../util/subscribeToResult';
 
-export function exhaustMap<T, R, R2>(project: (value: T, index: number) => Observable<R>,
-                                     resultSelector?: (outerValue: T,
-                                                       innerValue: R,
-                                                       outerIndex: number,
-                                                       innerIndex: number) => R2): Observable<R> {
+export function exhaustMap<T, R, TResult>(project: (value: T, index: number) => Observable<R>,
```
luisgabriel: Keep this type `R2` instead of renaming it to `TResult`.

src/operator/exhaustMap.ts
```diff
@@ -6,35 +6,33 @@ import {errorObject} from '../util/errorObject';
...
+                                          resultSelector?: (
+                                            outerValue: T,
+                                            innerValue: R,
+                                            outerIndex: number,
+                                            innerIndex: number) => TResult): Observable<TResult> {
   return this.lift(new SwitchFirstMapOperator(project, resultSelector));
 }
 
-class SwitchFirstMapOperator<T, R, R2> implements Operator<T, R> {
+class SwitchFirstMapOperator<T, R, TResult> implements Operator<T, TResult> {
```
luisgabriel: Same here.

src/operator/exhaustMap.ts
```diff
@@ -6,35 +6,33 @@ import {errorObject} from '../util/errorObject';
...
   }
 
-  call(subscriber: Subscriber<R>): Subscriber<T> {
+  call(subscriber: Subscriber<TResult>): Subscriber<T> {
     return new SwitchFirstMapSubscriber(subscriber, this.project, this.resultSelector);
   }
 }
 
-class SwitchFirstMapSubscriber<T, R, R2> extends OuterSubscriber<T, R> {
+class SwitchFirstMapSubscriber<T, R, TResult> extends OuterSubscriber<T, R> {
```
luisgabriel: Same here.

src/operator/find-support.ts
```diff
@@ -5,8 +5,8 @@ import {Subscriber} from '../Subscriber';
 import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
 
-export class FindValueOperator<T, R> implements Operator<T, R> {
-  constructor(private predicate: (value: T, index: number, source: Observable<T>) => boolean,
+export class FindValueOperator<T> implements Operator<T, T> {
+  constructor(private predicate: (value: T, index: number, observable: Observable<T>) => boolean,
```
luisgabriel: Please keep the name `source` here.

src/operator/find-support.ts
```diff
@@ -21,7 +21,7 @@ export class FindValueSubscriber<T> extends Subscriber<T> {
   private index: number = 0;
 
   constructor(destination: Subscriber<T>,
-              private predicate: (value: T, index: number, source: Observable<T>) => boolean,
+              private predicate: (value: T, index: number, observable: Observable<T>) => boolean,
```
luisgabriel: Same here.

src/operator/find.ts
```diff
@@ -1,7 +1,7 @@
-import {FindValueOperator} from './find-support';
 import {Observable} from '../Observable';
+import {FindValueOperator} from './find-support';
 
-export function find<T>(predicate: (value: T, index: number, source: Observable<T>) => boolean, thisArg?: any): Observable<T> {
+export function find<T>(predicate: (value: T, index: number, observable: Observable<T>) => boolean, thisArg?: any): Observable<T> {
```
luisgabriel: Same here.

src/operator/findIndex.ts
```diff
@@ -1,6 +1,6 @@
 import {Observable} from '../Observable';
 import {FindValueOperator} from './find-support';
 
-export function findIndex<T>(predicate: (value: T, index: number, source: Observable<T>) => boolean, thisArg?: any): Observable<number> {
+export function findIndex<T>(predicate: (value: T, index: number, observable: Observable<T>) => boolean, thisArg?: any): Observable<number> {
```
luisgabriel: Same here

src/operator/first.ts
```diff
@@ -5,14 +5,14 @@ import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
 import {EmptyError} from '../util/EmptyError';
 
-export function first<T, R>(predicate?: (value: T, index: number, source: Observable<T>) => boolean,
+export function first<T, R>(predicate?: (value: T, index: number, observable: Observable<T>) => boolean,
```
luisgabriel: Same here.

src/operator/first.ts
```diff
@@ -5,14 +5,14 @@ import {tryCatch} from '../util/tryCatch';
...
                             resultSelector?: (value: T, index: number) => R,
-                            defaultValue?: any): Observable<T> | Observable<R> {
+                            defaultValue?: R): Observable<T> | Observable<R> {
   return this.lift(new FirstOperator(predicate, resultSelector, defaultValue, this));
 }
 
 class FirstOperator<T, R> implements Operator<T, R> {
-  constructor(private predicate?: (value: T, index: number, source: Observable<T>) => boolean,
+  constructor(private predicate?: (value: T, index: number, observable: Observable<T>) => boolean,
```
luisgabriel: Same here.

src/operator/first.ts
```diff
@@ -28,7 +28,7 @@ class FirstSubscriber<T, R> extends Subscriber<T> {
   private hasCompleted: boolean = false;
 
   constructor(destination: Subscriber<R>,
-              private predicate?: (value: T, index: number, source: Observable<T>) => boolean,
+              private predicate?: (value: T, index: number, observable: Observable<T>) => boolean,
```
luisgabriel: Same here.

src/operator/last.ts
```diff
@@ -5,14 +5,14 @@ import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
 import {EmptyError} from '../util/EmptyError';
 
-export function last<T, R>(predicate?: (value: T, index: number, source: Observable<T>) => boolean,
+export function last<T, R>(predicate?: (value: T, index: number, observable: Observable<T>) => boolean,
```
luisgabriel: Same here.

src/operator/last.ts
```diff
@@ -5,14 +5,14 @@ import {tryCatch} from '../util/tryCatch';
...
                            resultSelector?: (value: T, index: number) => R,
-                           defaultValue?: any): Observable<T> | Observable<R> {
+                           defaultValue?: R): Observable<T> | Observable<R> {
   return this.lift(new LastOperator(predicate, resultSelector, defaultValue, this));
 }
 
 class LastOperator<T, R> implements Operator<T, R> {
-  constructor(private predicate?: (value: T, index: number, source: Observable<T>) => boolean,
+  constructor(private predicate?: (value: T, index: number, observable: Observable<T>) => boolean,
```
luisgabriel: Same here.

src/operator/last.ts
```diff
@@ -24,12 +24,12 @@ class LastOperator<T, R> implements Operator<T, R> {
...
 
 class LastSubscriber<T, R> extends Subscriber<T> {
-  private lastValue: T;
+  private lastValue: T | R;
   private hasValue: boolean = false;
   private index: number = 0;
 
   constructor(destination: Subscriber<R>,
-              private predicate?: (value: T, index: number, source: Observable<T>) => boolean,
+              private predicate?: (value: T, index: number, observable: Observable<T>) => boolean,
```
luisgabriel: Same here.

src/operator/mergeMap-support.ts
```diff
@@ -7,9 +7,9 @@ import {errorObject} from '../util/errorObject';
 import {subscribeToResult} from '../util/subscribeToResult';
 import {OuterSubscriber} from '../OuterSubscriber';
 
-export class MergeMapOperator<T, R, R2> implements Operator<T, R> {
+export class MergeMapOperator<T, R, TResult> implements Operator<T, R> {
```
luisgabriel: Keep `R2` here.

src/operator/mergeMap-support.ts
```diff
@@ -20,7 +20,7 @@ export class MergeMapOperator<T, R, R2> implements Operator<T, R> {
   }
 }
 
-export class MergeMapSubscriber<T, R, R2> extends OuterSubscriber<T, R> {
+export class MergeMapSubscriber<T, R, TResult> extends OuterSubscriber<T, R> {
```
luisgabriel: Same here.

src/operator/mergeMap.ts
```diff
@@ -1,11 +1,8 @@
...
 import {MergeMapOperator} from './mergeMap-support';
 
-export function mergeMap<T, R, R2>(project: (value: T, index: number) => Observable<R>,
-                                   resultSelector?: (outerValue: T,
-                                                     innerValue: R,
-                                                     outerIndex: number,
-                                                     innerIndex: number) => R,
-                                   concurrent: number = Number.POSITIVE_INFINITY) {
-  return this.lift(new MergeMapOperator(project, resultSelector, concurrent));
+export function mergeMap<T, R, TResult>(project: (value: T, index: number) => Observable<R>,
```
luisgabriel: Same here.

src/operator/mergeMapTo-support.ts
```diff
@@ -8,13 +8,13 @@ import {OuterSubscriber} from '../OuterSubscriber';
 import {subscribeToResult} from '../util/subscribeToResult';
 import {InnerSubscriber} from '../InnerSubscriber';
 
-export class MergeMapToOperator<T, R, R2> implements Operator<T, R> {
-  constructor(private ish: any,
-              private resultSelector?: (outerValue: T, innerValue: R, outerIndex: number, innerIndex: number) => R2,
+export class MergeMapToOperator<T, R, TResult> implements Operator<Observable<T>, TResult> {
```
luisgabriel: Same here.

src/operator/mergeMapTo.ts
```diff
@@ -1,11 +1,8 @@
...
 import {MergeMapToOperator} from './mergeMapTo-support';
 
-export function mergeMapTo<T, R, R2>(observable: Observable<R>,
-                                     resultSelector?: (outerValue: T,
-                                                       innerValue: R,
-                                                       outerIndex: number,
-                                                       innerIndex: number) => R2,
-                                     concurrent: number = Number.POSITIVE_INFINITY): Observable<R2> {
-  return this.lift(new MergeMapToOperator(observable, resultSelector, concurrent));
+export function mergeMapTo<T, R, TResult>(observable: Observable<R>,
```
luisgabriel: Same here.

src/operator/single.ts
```diff
@@ -7,14 +7,12 @@ import {tryCatch} from '../util/tryCatch';
...
-                                       index: number,
-                                       source: Observable<T>) => boolean): Observable<T> {
+export function single<T>(predicate?: (value: T, index: number, observable: Observable<T>) => boolean): Observable<T> {
   return this.lift(new SingleOperator(predicate, this));
 }
 
-class SingleOperator<T, R> implements Operator<T, R> {
-  constructor(private predicate?: (value: T, index: number, source: Observable<T>) => boolean,
+class SingleOperator<T> implements Operator<T, T> {
+  constructor(private predicate?: (value: T, index: number, observable: Observable<T>) => boolean,
```
luisgabriel: Keep the name `source` here.

src/operator/single.ts
```diff
@@ -29,12 +27,12 @@ class SingleSubscriber<T> extends Subscriber<T> {
   private index: number = 0;
 
   constructor(destination: Observer<T>,
-              private predicate?: (value: T, index: number, source: Observable<T>) => boolean,
+              private predicate?: (value: T, index: number, observable: Observable<T>) => boolean,
```
luisgabriel: Same here.

src/operator/switchMap.ts
```diff
@@ -7,20 +7,18 @@ import {errorObject} from '../util/errorObject';
 import {OuterSubscriber} from '../OuterSubscriber';
 import {subscribeToResult} from '../util/subscribeToResult';
 
-export function switchMap<T, R, R2>(project: (value: T, index: number) => Observable<R>,
-                                    resultSelector?: (outerValue: T,
-                                                      innerValue: R,
-                                                      outerIndex: number,
-                                                      innerIndex: number) => R2): Observable<R> {
+export function switchMap<T, R, TResult>(project: (value: T, index: number) => Observable<R>,
```
luisgabriel: Keep `R2` here.

src/operator/switchMap.ts
```diff
@@ -7,20 +7,18 @@ import {errorObject} from '../util/errorObject';
...
+                                         resultSelector?: (
+                                             outerValue: T,
+                                             innerValue: R,
+                                             outerIndex: number,
+                                             innerIndex: number) => TResult): Observable<TResult> {
   return this.lift(new SwitchMapOperator(project, resultSelector));
 }
 
-class SwitchMapOperator<T, R, R2> implements Operator<T, R> {
+class SwitchMapOperator<T, R, TResult> implements Operator<T, R> {
```
luisgabriel: Same here.

src/operator/switchMap.ts
```diff
@@ -28,14 +26,14 @@ class SwitchMapOperator<T, R, R2> implements Operator<T, R> {
   }
 }
 
-class SwitchMapSubscriber<T, R, R2> extends OuterSubscriber<T, R> {
+class SwitchMapSubscriber<T, R, TResult> extends OuterSubscriber<T, R> {
```
luisgabriel: Same here.

src/operator/switchMapTo.ts
```diff
@@ -7,20 +7,18 @@ import {errorObject} from '../util/errorObject';
 import {OuterSubscriber} from '../OuterSubscriber';
 import {subscribeToResult} from '../util/subscribeToResult';
 
-export function switchMapTo<T, R, R2>(observable: Observable<R>,
-                                      projectResult?: (outerValue: T,
-                                                       innerValue: R,
-                                                       outerIndex: number,
-                                                       innerIndex: number) => R2): Observable<R2> {
-  return this.lift(new SwitchMapToOperator(observable, projectResult));
+export function switchMapTo<T, R, TResult>(observable: Observable<R>,
```
luisgabriel: Keep `R2` here.

src/operator/switchMapTo.ts
```diff
@@ -7,20 +7,18 @@ import {errorObject} from '../util/errorObject';
...
+                                           resultSelector?: (
+                                                 outerValue: T,
+                                                 innerValue: R,
+                                                 outerIndex: number,
+                                                 innerIndex: number) => TResult): Observable<TResult> {
+  return this.lift(new SwitchMapToOperator(observable, resultSelector));
 }
 
-class SwitchMapToOperator<T, R, R2> implements Operator<T, R> {
+class SwitchMapToOperator<T, R, TResult> implements Operator<T, R> {
```
luisgabriel: Keep `R2` here.

src/operator/switchMapTo.ts
```diff
@@ -28,14 +26,14 @@ class SwitchMapToOperator<T, R, R2> implements Operator<T, R> {
   }
 }
 
-class SwitchMapToSubscriber<T, R, R2> extends OuterSubscriber<T, R> {
+class SwitchMapToSubscriber<T, R, TResult> extends OuterSubscriber<T, R> {
```
luisgabriel: Same here.

src/operator/concatMap.ts
```diff
@@ -20,7 +20,7 @@ import {MergeMapOperator} from './mergeMap-support';
  * @returns {Observable} an observable of values merged from the projected Observables as they were subscribed to,
  * one at a time. Optionally, these values may have been projected from a passed `projectResult` argument.
  */
-export function concatMap<T, R>(project: (value: T, index: number) => Observable<any>,
-                                projectResult?: (outerValue: T, innerValue: any, outerIndex: number, innerIndex: number) => R) {
-  return this.lift(new MergeMapOperator(project, projectResult, 1));
+export function concatMap<T, R, TResult>(project: (value: T, index: number) => Observable<R>,
```
luisgabriel: Use `R2` here instead of `TResult`.

src/operator/concatMapTo.ts
```diff
@@ -14,10 +14,11 @@ import {MergeMapToOperator} from './mergeMapTo-support';
  * @returns {Observable} an observable of values merged together by joining the passed observable
  * with itself, one after the other, for each value emitted from the source.
  */
-export function concatMapTo<T, R, R2>(observable: Observable<R>,
-                                      projectResult?: (outerValue: T,
-                                                       innerValue: R,
-                                                       outerIndex: number,
-                                                       innerIndex: number) => R2): Observable<R2> {
-  return this.lift(new MergeMapToOperator(observable, projectResult, 1));
+export function concatMapTo<T, R, TResult>(observable: Observable<R>,
```
luisgabriel: Keep it `R2`.

luisgabriel:
@kwonoj @david-driscoll Just finished to review it. Comments are inline. Most of them are style-related.

src/observable/fromEvent.ts
david-driscoll: done

src/observable/fromEventPattern.ts
david-driscoll: done

src/operator/concat.ts
david-driscoll: done

src/operator/every.ts
david-driscoll: yep, done.

src/operator/exhaustMap.ts
david-driscoll: I can do that, but do you mind if I ask why specificly (outside keeping it like it was).  I just personally found `R` and `R2` fairly confusing looking at the code.

luisgabriel: It's not clear looking at a variable declaration, for example, if `TResult` is a type parameter or a class/interface. Also, it's not consistent with the pattern we're using to name type parameters in this codebase as well as in RxJava.

luisgabriel:
LGTM.

Thanks, @david-driscoll!

kwonoj:
Cool, I'll check this in around late today to this weekend, allow some time buffer to see any other suggestions around considering given size of PR.

src/operator/exhaustMap.ts
david-driscoll: Fair enough, in C# land generally generic types are just prefixed with `T`
to distinguish them, and you'll almost never see a class with TMyClass.
Worth asking the question. :)

Thanks.

src/add/operator/zip-static.ts
```diff
@@ -1,5 +1,5 @@
 import {Observable} from '../../Observable';
 import {zip} from '../../operator/zip-static';
-Observable.zip = zip;
+Observable.zip = <any>zip;
```
luisgabriel: While working on #1030, I remembered that this patch touches in some auto-generated files. I think you should remove this kind of change from this patch. You'll need to drop changes in any file inside `src/add/*` (I think you modified only zip and zip-static in this patch).

What do you think, @kwonoj?

kwonoj: I think that makes sense if some of files are being generated automatically. Question's in that case - doesn't auto generated file impact build? shouldn't auto generator to be modified to ensure build passes?

luisgabriel: Yes, if dropping this change breaks the build we should modify the script to generate code that deals with this case. But I'm not sure if this is the case since this is the only static operator that required casting. The other change (`src/add/operator/zip.ts`) was only a renaming, so it can be dropped without problems.

kwonoj: Seems auto generator has function `generateNewObservableFileContents` , `generateNewOperatorFileContents ` which handles and it could be updated to handle this, but also still these file need to be included PR cause our codebase generates does file and check in into codebases, isn't it? only difference would be instead of manually modifying each files, generator should be updated and file need to be regenerated via scripts.

luisgabriel: If this change is really needed (the build breaks without it), I think we should integrate this PR with this change and I can modify the script afterwards to handle this case. But I'd drop it if this is not the case.

kwonoj: In this PR specific, I believe this was added to ensure to not to break build. Asking @david-driscoll to confirm once again.

luisgabriel: I fetched this commit locally and did some tests. It seems that removing this casting does not break the build. Can you confirm, @david-driscoll?

luisgabriel:
@kwonoj I think it'd be better if we could merge this PR before the change @trxcllnt and @blesh are discussing on #1099 to avoid more rebasing conflicts.

What do you guys think?

david-driscoll:
I haven't had a chance to check it, but it's possible it's not needed
anymore.
On Wed, Dec 23, 2015 at 2:07 PM Luís Gabriel Lima <notifications@github.com>
wrote:

> @kwonoj <https://github.com/kwonoj> I think it'd be better if we could
> merge this PR before the change @trxcllnt <https://github.com/trxcllnt>
> and @blesh <https://github.com/blesh> are discussing on #1099
> <https://github.com/ReactiveX/RxJS/pull/1099> to avoid more rebasing
> conflicts.
>
> What do you guys think?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ReactiveX/RxJS/pull/1077#issuecomment-166972431>.
>
-- 
Thanks,
David

blesh:
Just a note: This PR should probably wait for @trxcllnt's changes (code reviewed in branch and PR incoming shortly) to the Subscription scheme to land. They're big changes, so it may mean a painful refactor of this PR... however, they're important changes to internals, so they take precedence.

kwonoj:
Agreed with @blesh here. It wasn't intended to make this PR's check in delayed though, now it's overlaps with major refactoring to subscriptions - in this case, I'd like to let those changes land first prior to this. (Also https://github.com/ReactiveX/RxJS/pull/1097 as well @luisgabriel , current assumption's those PR's nearly orthogonal to subscription changes, but want to be sure).

luisgabriel:
Ok, no problem. I was assuming that it would be less painful to adapt @trxcllnt's changes since merging this one after could also require adding explicit typing on the new code besides solving the conflicts.

trxcllnt:
@david-driscoll the effort that went into this PR is clearly enormous, so thanks for that. Apologies if rebasing from #1106 is a major headache. I'll do the legwork to reconcile with this PR if necessary.

kwonoj:
#1106 's now landed, @david-driscoll would you able to rebase this? I'll take this as priority to check in once it's rebased, given size of changes delaying it more would create more change conflicts between PRs. (Just note : @luisgabriel , I'll take #1097 as second order after this one. Let me know if you have any suggestions).

luisgabriel:
@kwonoj Ok, no problem from my side. I'll try to rebase this patch tomorrow to see if I can help @david-driscoll to update this PR.

luisgabriel:
@david-driscoll I've rebased your patch on the current master. You can find it in this branch: https://github.com/luisgabriel/RxJS/tree/pr-1077

It was less painful than I thought. ;)

blesh:
The sooner we get the rebased version in the PR (or PR'ed elsewhere) the sooner we can review the changes again after the rebase and merge this.

blesh:
Marking "Blocked" as it's awaiting rebase.

david-driscoll:
Will get working on this is in the next day or two, I've been enjoying some
much needed time off :)

On Mon, Jan 4, 2016 at 3:14 PM Ben Lesh <notifications@github.com> wrote:

> Marking "Blocked" as it's awaiting rebase.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ReactiveX/RxJS/pull/1077#issuecomment-168791363>.
>
-- 
Thanks,
David

luisgabriel:
@david-driscoll I recommend you use https://github.com/luisgabriel/RxJS/tree/pr-1077 as a base for updating this PR. A lot of the conflicts are already solved in this branch.

robwormald:
:+1: also @blesh, who's the masochist now? https://github.com/ReactiveX/RxJS/pull/517#issuecomment-147892560

blesh:
@robwormald haha. Yeah, that was before I knew it would reduce the amount of time I'd have to wait for builds. :package:

david-driscoll:
@luisgabriel sorry for the delay, life happens.  Now fully rebased and building correctly.

luisgabriel:
@david-driscoll No problem at all. Thank you for the effort! ;)

src/operator/distinctUntilChanged.ts
```diff
@@ -10,13 +11,15 @@ import {errorObject} from '../util/errorObject';
  * @param {function} [compare] optional comparison function called to test if an item is distinct from the previous item in the source.
  * @returns {Observable} an Observable that emits items from the source Observable with distinct values.
  */
-export function distinctUntilChanged<T>(compare?: (x: any, y: any) => boolean, keySelector?: (x: T) => any) {
-  return this.lift(new DistinctUntilChangedOperator(compare, keySelector));
+export function distinctUntilChanged<T>(compare?: (x: T, y: T) => boolean): Observable<T>;
+export function distinctUntilChanged<T, K>(compare: (x: K, y: K) => boolean, keySelector?: (x: T) => K): Observable<T>;
+export function distinctUntilChanged<T, K>(compare: (x: K, y: K) => boolean, keySelector?: (x: T) => K): Observable<T> {
```
luisgabriel: Is this correct?

david-driscoll: If you supply a `keySelector`, then the values that get handed to compare, are the type that is returned by the `keySelector` otherwise the values handed to compare are `T`.

luisgabriel: Got it. Thanks.

luisgabriel:
LGTM :+1:

kwonoj:
Cool, I'm checking in this.. (finally!)

kwonoj:
Merged with db7f0b1. Appreciate @david-driscoll  for this huge effort, and also @luisgabriel takes effort to review this changes too.

PR #1079, opened 2015-12-15
Inline docs for operators (D - F)

justinwoo:
Not much, just simple inline docs written for operators D-F that were missing inline docs.

Some things I did that you may find not so great for ease of reading:

- Each element of the source observable is called an `input`
- The source observable is called the `source`
- Observables are often referred to as simply `sequence` (as in, "input sequence" or "output sequence")
- Each element of the returned/resulting observable is called an `output`
- The returned/resulting observable is often referred to as the `[verb]ed sequence`
- Avoided "invoke" in favor of "called"
- Avoided "project" or "predicate" in favor of just saying "[a] function [that is] called"

With the assumption that anyone looking for the most accurate information possible will probably be looking at the type signatures first, specs second, implementation third, so having a simple-enough explanation is better than being as literate as possible.

Figured I'd stop here for now to see feedback and fix some of my terrible mistakes while there's only a little over a dozen for now.

Also: anyone with strong opinions on JSDoc type signatures?

### DIFF
--- a/src/operator/debounce.ts
+++ b/src/operator/debounce.ts
@@ -9,6 +9,16 @@ import {errorObject} from '../util/errorObject';
 import {OuterSubscriber} from '../OuterSubscriber';
 import {subscribeToResult} from '../util/subscribeToResult';
 
+/**
+ * Returns the source Observable delayed by the computed debounce duration,
+ * with the duration lengthened if a new source item arrives before the delay
+ * duration ends.
+ * In practice, for each item emitted on the source, this operator holds the
+ * latest item, waits for a silence as long as the `durationSelector` specifies,
+ * and only then emits the latest source item on the result Observable.
+ * @param {function} durationSelector function for computing the timeout duration for each item.
+ * @returns {Observable} an Observable the same as source Observable, but drops items.
+ */
 export function debounce<T>(durationSelector: (value: T) => Observable<any> | Promise<any>): Observable<T> {
   return this.lift(new DebounceOperator(durationSelector));
 }
--- a/src/operator/debounceTime.ts
+++ b/src/operator/debounceTime.ts
@@ -5,6 +5,18 @@ import {Scheduler} from '../Scheduler';
 import {Subscription} from '../Subscription';
 import {asap} from '../scheduler/asap';
 
+/**
+ * Returns the source Observable delayed by the computed debounce duration,
+ * with the duration lengthened if a new source item arrives before the delay
+ * duration ends.
+ * In practice, for each item emitted on the source, this operator holds the
+ * latest item, waits for a silence for the `dueTime` length, and only then
+ * emits the latest source item on the result Observable.
+ * Optionally takes a scheduler for manging timers.
+ * @param {number} dueTime the timeout value for the window of time required to not drop the item.
+ * @param {Scheduler} [scheduler] the Scheduler to use for managing the timers that handle the timeout for each item.
+ * @returns {Observable} an Observable the same as source Observable, but drops items.
+ */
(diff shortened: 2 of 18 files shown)


staltz:
> Also: anyone with strong opinions on JSDoc type signatures?

https://github.com/ericelliott/rtype

staltz:
@justinwoo Thanks for the initiative to do these. Because RxJS 4 docs were known to be very confusing, perhaps too terse, RxJS 5 needs good and friendly docs. Here is some general feedback from my perspective:

- Don't be afraid to use more text. It's more important to explain the operator in human terms so humans can understand.
- `sequence` is a terminology used only in RxJS 4 as far as I know. I think it's better to say Observable or *stream*. Erik Meijer has used stream sometimes, and "Observable" is Stream in Dart lang.
- I highly recommend following/imitating RxJava docs. http://reactivex.io/RxJava/javadoc/rx/Observable.html

Compare for instance your debounce description with RxJava's debounce description:
> Ignores values from an observable following an initial value within a timed debounce duration.

http://reactivex.io/RxJava/javadoc/rx/Observable.html#debounce%28long,%20java.util.concurrent.TimeUnit%29
> Returns an Observable that mirrors the source Observable, except that it drops items emitted by the source Observable that are followed by another item within a computed debounce duration.

If the sentence starts with "Ignores", it sounds like this operator is used for ignoring values. In RxJava, they use "mirrors the source, ... except for these details".

justinwoo:
Cool, made the wording changes `input -> item`, `sequence -> Observable` and rewrote a lot of these to be more like the JavaDoc descriptions of the operators.

src/operator/debounceTime.ts
```diff
@@ -5,6 +5,14 @@ import {Scheduler} from '../Scheduler';
 import {Subscription} from '../Subscription';
 import {asap} from '../scheduler/asap';
 
+/**
+ * Returns an Observable the same as source Observable, but drops items
+ * emitted by the source followed by additional items before the timeout
+ * value expires on the scedule. The timeout resets on each emission.
```
staltz: `scedule` => `schedule`

src/operator/debounce.ts
```diff
@@ -8,6 +8,13 @@ import {tryCatch} from '../util/tryCatch';
 import {isPromise} from '../util/isPromise';
 import {errorObject} from '../util/errorObject';
 
+/**
+ * Returns an Observable the same as source Observable, but drops items
+ * emitted by the source followed by additional items before the computed
+ * timeout based on each item in the source.
```
staltz: For debounce(Time), to keep it as straight-to-the-point as possible, I'd describe it as
"*Returns the source Observable delayed by the computed debounce duration, except it extends the duration if a new source item arrives before the previous source item was emitted on the result Observable. In practice, for each item emitted on the source, this operator holds the latest item, waits for 'silence' as long as the `durationSelector` specifies, and only then emits the latest source item on the result Observable.*"

justinwoo:
Updated debounce/debounceTime docs as suggested and rebased.

staltz:
:+1:

blesh:
Looks good, but it needs rebased since @trxcllnt's changes to Subscription scheme landed.

blesh:
Looks good, but it needs to be rebased.

justinwoo:
Rebased. Should be good to go.

blesh:
Merged with 84e99af. Thanks @justinwoo!

... When the next version is published, I'll regen and update the docs.

justinwoo:
:+1: I'll try to work on getting some more inline docs done. Only G-Z to go! :scream_cat:

PR #1080, opened 2015-12-15
prepend licenses to npm-published bundles

jeffbcross:
I created two scripts to generate the umd and System bundles as well, so that SystemJS bundles could be included in global build.


blesh:
Merged with 1ff3b89. Thanks @jeffbcross

PR #1082, opened 2015-12-16
docs(MIGRATION): Add the section about renamed classes from RxJS 4.

saneyuki:


saneyuki:
I think we need this guide for the migration document for RxJS 4 users. They may use `new Rx.CompositeDisposable()` or `Rx.Observer.create()`.

In the migration document, we have already written:

> In RxJS 4, there was the idea of a `CompositeSubscription`.

but there is no mention about `CompositeSubscription` which is same as `CompositeDisposable`. This would confuses a reader.

saneyuki:
@blesh 

How do you think about this pull request?

MIGRATION.md
```diff
@@ -10,6 +10,16 @@ three basic goals:
 Meeting the above goals meant breaking changes to the RxJS API, and a complete rewrite means that we had opportunity
 to change/fix/correct things we've wanted to correct about RxJS in general.
 
+## Key Component Classes Are Recomposed
+
+They are similar to other language implemetations of ReactiveX (e.g. RxJava).
```
staltz: `implemetations` => `implementations`

saneyuki: oops!

saneyuki:
Addressed.

blesh:
Looks good. Merged with 3dea4e5. Thanks @saneyuki

PR #1083, opened 2015-12-16
fix(debounceTime): align value emit behavior as same as RxJS4

### LINKED ISSUE #1081: debounceTime doesn't work with Subject
We had a discussion here https://github.com/angular/angular/issues/5925. It seems that `throttleTime` works with `Subject` and `debounceTime` doesn't work with `Subject`.
`throttleTime`: http://jsfiddle.net/bo6ynqf7/1/
`debounceTime`: http://jsfiddle.net/8g968xh7/1/

kwonoj:
due to current null check, debounceTime does not allow value
to be `undefined`. This change updates behavior to align behavior
as same as RxJS4 does.

closes #1081

### DIFF
--- a/src/operator/debounceTime.ts
+++ b/src/operator/debounceTime.ts
@@ -20,7 +20,8 @@ class DebounceTimeOperator<T, R> implements Operator<T, R> {
 
 class DebounceTimeSubscriber<T> extends Subscriber<T> {
   private debouncedSubscription: Subscription<any> = null;
-  private lastValue: any = null;
+  private lastValue: T = null;
+  private hasValue: boolean = false;
 
   constructor(destination: Subscriber<T>,
               private dueTime: number,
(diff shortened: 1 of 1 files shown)


kwonoj:
I think this change's legit, but will wait to see if any other opinion / suggestions around before checkin.

mattpodwysocki:
@kwonoj why are you checking for null/undefined, when you should have a separate flag instead for hasValue just as we did in RxJS v4?  After all, null is also a valid value to send along.  Take a look at the current v4 implementation: https://github.com/Reactive-Extensions/RxJS/blob/master/src/modular/observable/debounce.js

kwonoj:
Thanks for suggestion, it's quite valid point. Current implementation used `null` instead of having `hasValue` to check it has value or not, so as pointed out it won't be able to accept it as value. I'll update PR to reflect behavior correctly.

staltz:
Ah, this PR exists... nice.

kwonoj:
PR updated to introduce `hasValue` instead of current implementation uses `null` to check if value has been set or not, to avoid some of values are not being accepted correctly.

kwonoj:
Merged with 5ee11e0.

PR #1084, opened 2015-12-16
docs(MIGRATION): fix MIGRATION.md typos and missing operators

staltz:
Fix typos in MIGRATION.md, and add more operators to the list of renamed operators. Also extend
that list to show removed operators.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -32,27 +32,81 @@ var subscription = myObservable.subscribe(doSomething);
 subscription.unsubscribe();
 ```
 
-
 ## Subscription: All Subscriptions Are "Composite"
 
 In RxJS 4, there was the idea of a `CompositeSubscription`. Now all Subscriptions are "composite".
 Subscription objects have an `add` and `remove` method on them useful for adding and removing subscriptions
 enabling "composite" subscription behavior.
 
-
-
-## Operators Renamed
+## Operators Renamed or Removed
 
 |RxJS 4|RxJS 5|
 |---|---|
-|`flatMap`|`mergeMap` or `flatMap`(alias)|
-|`flatMapLatest`|`switchMap`|
-|`bufferWithTime`|`bufferTime`|
 |`bufferWithCount`|`bufferCount`|
+|`bufferWithTime`|`bufferTime`|
+|`flatMap` or `selectMany`|`mergeMap` or `flatMap`(alias)|
+|`flatMapFirst`|`exhaustMap`|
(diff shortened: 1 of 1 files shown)


MIGRATION.md
```diff
@@ -32,27 +32,81 @@ var subscription = myObservable.subscribe(doSomething);
...
+|`flatMapWithMaxConcurrent`|`mergeMap` or `flatMap`(alias)|
+|`publishValue`|`publishBehavior`|
+|`replay`|`publishReplay`|
+|`select`|`map`|
+|`selectConcat`|`concatMap`|
+|`switchFirst`|`exhaust`|
+|`tap`|`do`|
 |`windowWithTime`|`windowTime`|
 |`windowWithCount`|`windowCount`|
-|`switchFirst`|`exhaustFirst`|
```
staltz: `exhaustFirst` was wrong.

MIGRATION.md
```diff
@@ -32,27 +32,81 @@ var subscription = myObservable.subscribe(doSomething);
...
-|`switchFirstMap`|`exhaustFirstMap`|
```
staltz: `switchFirstMap` doesn't exist in RxJS 4. It's `flatMapFirst`.

MIGRATION.md
```diff
@@ -32,27 +32,81 @@ var subscription = myObservable.subscribe(doSomething);
...
+|`where`|`filter`|
+|`amb`|`-`|
+|`and`|`-`|
```
kwonoj: what about use term like (WIP) or similar for now except operators completely decided to be dropped? might possible some people think lot of operators are dropped by just reading this document for first time.

staltz: But I don't even know which of these do we plan to support. The only one I know, `pairwise`, is not on this "removed" list. Anyway, we are removing lots of operators.

kwonoj: Yes, that's reason for that. My suggestion is simply mark everything as 'wip' for now, then change it to 'dropped' once it's decided instead of looking it as lot of operators dropped.

blesh:
merged with 657132d

blesh:
(and thanks, as always) :p @staltz

MIGRATION.md
blesh: FYI: I decided to merge because it's just docs and `-` is better than _nothing_. 

Personally I like `WIP` better. But that's me.

kwonoj: Yup, makes sense. Those can be updated further.

PR #1086, opened 2015-12-16
manual typings for bindCallback.create

david-driscoll:
Comparing against generation of types to spur on discussion of the best approach.  I picked a single file here just as an example case.  See #1087

As a reference for TypeScript itself, these typings can't be simplified much more without losing the enhanced experience of operating against them.

https://github.com/Microsoft/TypeScript/issues/5453 may be the answer, not sure if it answers all the possible use cases.  It is currently on the road map for 2.0.

I realize there will be tons of opinions my _only_ goal is to make the TypeScript experience with RxJS5 as good as we have gotten the experience in RxJS4.

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -37,6 +37,8 @@ import {ErrorObservable} from './observable/ErrorObservable';
 import {AjaxCreationMethod} from './observable/dom/AjaxObservable';
 import {WebSocketSubject} from './observable/dom/WebSocketSubject';
 
+import {CombineLatestSignature} from './operator/combineLatest';
+
 export type ObservableOrPromise<T> = Observable<T> | Promise<T>;
 export type ArrayOrIterator<T> = Iterator<T> | ArrayLike<T>;
 export type ObservableInput<T> = ObservableOrPromise<T> | ArrayOrIterator<T>;
@@ -191,9 +193,7 @@ export class Observable<T> implements CoreOperators<T>  {
   bufferWhen: (closingSelector: () => Observable<any>) => Observable<T[]>;
   catch: (selector: (err: any, source: Observable<T>, caught: Observable<any>) => Observable<any>) => Observable<T>;
   combineAll: <R>(project?: (...values: Array<any>) => R) => Observable<R>;
-  combineLatest: <R>(...observables: Array<Observable<any> |
-                                     Array<Observable<any>> |
-                                     ((...values: Array<any>) => R)>) => Observable<R>;
+  combineLatest: CombineLatestSignature<T>;
   concat: <R>(...observables: (Observable<any> | Scheduler)[]) => Observable<R>;
   concatAll: () => Observable<any>;
   concatMap: <R>(project: ((x: T, ix: number) => Observable<any>), projectResult?: (x: T, y: any, ix: number, iy: number) => R) => Observable<R>;
--- a/src/operator/combineLatest.ts
+++ b/src/operator/combineLatest.ts
@@ -40,18 +40,18 @@ export function combineLatest<T, R>(...observables: Array<ObservableInput<any> |
 
 /* tslint:disable:max-line-length */
 export interface CombineLatestSignature<T> {
-  <T2>(v2: ObservableInput<T2>): Observable<[T, T2]>;
-  <T2, T3>(v2: ObservableInput<T2>, v3: ObservableInput<T3>): Observable<[T, T2, T3]>;
-  <T2, T3, T4>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>): Observable<[T, T2, T3, T4]>;
-  <T2, T3, T4, T5>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>): Observable<[T, T2, T3, T4, T5]>;
-  <T2, T3, T4, T5, T6>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>): Observable<[T, T2, T3, T4, T5, T6]>;
-
   <R>(project: (v1: T) => R): Observable<R>;
   <T2, R>(v2: ObservableInput<T2>, project: (v1: T, v2: T2) => R): Observable<R>;
   <T2, T3, R>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, project: (v1: T, v2: T2, v3: T3) => R): Observable<R>;
   <T2, T3, T4, R>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, project: (v1: T, v2: T2, v3: T3, v4: T4) => R): Observable<R>;
   <T2, T3, T4, T5, R>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, project: (v1: T, v2: T2, v3: T3, v4: T4, v5: T5) => R): Observable<R>;
   <T2, T3, T4, T5, T6, R>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>, project: (v1: T, v2: T2, v3: T3, v4: T4, v5: T5, v6: T6) => R): Observable<R>;
+
+  <T2>(v2: ObservableInput<T2>): Observable<[T, T2]>;
+  <T2, T3>(v2: ObservableInput<T2>, v3: ObservableInput<T3>): Observable<[T, T2, T3]>;
+  <T2, T3, T4>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>): Observable<[T, T2, T3, T4]>;
+  <T2, T3, T4, T5>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>): Observable<[T, T2, T3, T4, T5]>;
+  <T2, T3, T4, T5, T6>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>): Observable<[T, T2, T3, T4, T5, T6]>;
 
   <R>(...observables: Array<ObservableInput<any> | ((...values: Array<any>) => R)>): Observable<R>;
   <R>(array: ObservableInput<any>[]): Observable<R>;
--- a/src/operator/combineLatest.ts
+++ b/src/operator/combineLatest.ts
@@ -1,4 +1,4 @@
-import {Observable} from '../Observable';
+import {Observable, ObservableInput} from '../Observable';
 import {ArrayObservable} from '../observable/ArrayObservable';
 import {isArray} from '../util/isArray';
 import {Scheduler} from '../Scheduler';
@@ -19,8 +19,8 @@ import {subscribeToResult} from '../util/subscribeToResult';
  * @returns {Observable} an observable of other projected values from the most recent values from each observable, or an array of each of
  * the most recent values from each observable.
  */
-export function combineLatest<T, R>(...observables: Array<Observable<any> |
-                                                       Array<Observable<any>> |
+export function combineLatest<T, R>(...observables: Array<ObservableInput<any> |
+                                                       Array<ObservableInput<any>> |
                                                        ((...values: Array<any>) => R)>): Observable<R> {
   let project: (...values: Array<any>) => R = null;
   if (typeof observables[observables.length - 1] === 'function') {
@@ -38,6 +38,27 @@ export function combineLatest<T, R>(...observables: Array<Observable<any> |
   return new ArrayObservable(observables).lift<T, R>(new CombineLatestOperator<T, R>(project));
 }
 
+/* tslint:disable:max-line-length */
+export interface CombineLatestSignature<T> {
+  <T2>(v2: ObservableInput<T2>): Observable<[T, T2]>;
+  <T2, T3>(v2: ObservableInput<T2>, v3: ObservableInput<T3>): Observable<[T, T2, T3]>;
+  <T2, T3, T4>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>): Observable<[T, T2, T3, T4]>;
+  <T2, T3, T4, T5>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>): Observable<[T, T2, T3, T4, T5]>;
+  <T2, T3, T4, T5, T6>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>): Observable<[T, T2, T3, T4, T5, T6]>;
+
+  <R>(project: (v1: T) => R): Observable<R>;
+  <T2, R>(v2: ObservableInput<T2>, project: (v1: T, v2: T2) => R): Observable<R>;
+  <T2, T3, R>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, project: (v1: T, v2: T2, v3: T3) => R): Observable<R>;
+  <T2, T3, T4, R>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, project: (v1: T, v2: T2, v3: T3, v4: T4) => R): Observable<R>;
+  <T2, T3, T4, T5, R>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, project: (v1: T, v2: T2, v3: T3, v4: T4, v5: T5) => R): Observable<R>;
+  <T2, T3, T4, T5, T6, R>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>, project: (v1: T, v2: T2, v3: T3, v4: T4, v5: T5, v6: T6) => R): Observable<R>;
+
+  <R>(...observables: Array<ObservableInput<any> | ((...values: Array<any>) => R)>): Observable<R>;
+  <R>(array: ObservableInput<any>[]): Observable<R>;
+  <R>(array: ObservableInput<any>[], project: (...values: Array<any>) => R): Observable<R>;
+}
+/* tslint:enable:max-line-length */
+
 /**
  * Combines the values from observables passed as arguments. This is done by subscribing
  * to each observable, in order, and collecting an array of each of the most recent values any time any of the observables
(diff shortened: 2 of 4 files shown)


blesh:
@david-driscoll I think you proposed this at an inopportune time (before the holidays).

Do my heavy TypeScript users have any opinions? @robwormald @jeffbcross @kwonoj

kwonoj:
Discussion at the moment makes this discussion pause until https://github.com/ReactiveX/RxJS/pull/1077 is landed in codebase, to separate scope of changes. 

Shortly, I still need spend some time to review these changes.

kwonoj:
Just to ensure, PR intended to be checked in is https://github.com/ReactiveX/RxJS/pull/1087, this PR shows handcrafted changes to compare with original PR.

robwormald:
the better the types in the distribution, the better, as far as i'm concerned. i'll take an extra few kb's of npm download for better intellisense. Up to y'all in terms of maintainability.

david-driscoll:
@kwonoj yes, I would prefer to check in is #1087, but I offered both to allow for comparing the difference between the two.

I have also rebased against master.

david-driscoll:
@kwonoj / @luisgabriel 

After doing some work today, I'm thinking this branch might be the preferred method, I added some "unit tests" as well (basically to ensure that things compile as expected).

kwonoj:
> this branch might be the preferred method, 

meaning between #1087 and this?

> I added some "unit tests" as well

Would you able to separate those? You may able to see there are pending PR in discussion https://github.com/ReactiveX/RxJS/pull/826 https://github.com/ReactiveX/RxJS/pull/667 for enabling test against type definitions, haven't concluded which way'll go. If your changes separated, that can be considerable option too.

david-driscoll:
Correct, this over #1087, it's a little more manual work, but the quality will probably be better.

If you look at `bindCallback-tpec.ts` for example, it doesn't actually do anything meaningful, in terms of testing functionality, it just validates that interface defined for `bindCallback` will compile as expected with the correct type inferences, etc.

kwonoj:
> If you look at bindCallback-tpec.ts for example,

yes, aware of those. Current discussion regarding test is about organization of test since it's testing definition only - so it'll be matter of trying to use existing PR's mechanism, or following your PR's mechanism to invoke those. Once it's decided, test can be added fairly quickly.

david-driscoll:
@kwonoj done!

kwonoj:
Thanks! I still need to look changes to add some inputs.. (my work-for-pay-bill blocks me, sorry for late response)

src/add/operator/combineLatest.ts
```diff
@@ -5,6 +5,6 @@
 import {Observable} from '../../Observable';
 import {combineLatest} from '../../operator/combineLatest';
 
-Observable.prototype.combineLatest = combineLatest;
+Observable.prototype.combineLatest = <any>combineLatest;
```
kwonoj: this patch operator's now auto generated, so instead of changing this script need to be updated - is that correct?

src/observable/bindCallback.ts
```diff
@@ -9,6 +9,24 @@ import {AsyncSubject} from '../subject/AsyncSubject';
 export class BoundCallbackObservable<T> extends Observable<T> {
   subject: AsyncSubject<T>;
 
+  /* tslint:disable:max-line-length */
+  static create<TResult>(callbackFunc: (callback: (result: TResult) => any) => any, selector?: void, scheduler?: Scheduler): () => Observable<TResult>;
```
kwonoj: My assumption is any similar operator would need same amount of declarations (i.e, `BoundNodeCallbackObservable`) - is that correct?

src/operator/combineLatest.ts
```diff
@@ -31,3 +31,28 @@ export function combineLatest<T, R>(...observables: Array<Observable<any> |
 
   return new ArrayObservable(observables).lift<T, R>(new CombineLatestOperator<T, R>(project));
 }
+
+/* These types would be come standard types used through-out the library */
```
kwonoj: Means this might need to be exported into different module later in common usage?

src/operator/combineLatest.ts
```diff
@@ -31,3 +31,28 @@ export function combineLatest<T, R>(...observables: Array<Observable<any> |
...
+export type ObservableOrPromise<T> = Observable<T> | Promise<T>;
+export type ArrayOrIterator<T> = Iterator<T> | ArrayLike<T> | Array<T>;
+export type ObservableInput<T> = Observable<T> | Promise<T> | Iterator<T> | ArrayLike<T>;
+
+/* tslint:disable:max-line-length */
+export interface CombineLatestDeclaration<T> {
```
kwonoj: (It's just naming) I'm currently mixing usage of term between `Declaration` with `Definition`, which one looks more intuitive name? or maybe `CombineLatestSignature`, `CombineLatestType`, etcs?

kwonoj:
Let me ask few questions as 'newcomer' point of view, let's say I'm completely new to codebase and compose new operator require similar function signatures

- How do I determine if operator requires this 1 to n type definition or not?
- Where can I refer to create those signatures? such as base templates, examples, etcs
- How do I correctly verify once it's composed to send out PR? (Yes we do have https://github.com/ReactiveX/RxJS/issues/661, questions about 'how to compose those test')

There might be some other questions I could possible missing.

kwonoj:
+ would you agree to close #1087 in case if this PR's expected to be checked in? (confirming if I understood correctly)

src/add/operator/combineLatest.ts
luisgabriel: Is this change needed?

david-driscoll: yes, because `combineLatest` (the method) no longer has the same definition as `combineLatest` the prototype method.

I plan to update the script to add this in, and that's when I found out that the generation script didn't work the other day.

src/observable/bindCallback.ts
david-driscoll: I don't know how important that is to be honest.  My feeling is we don't need to, because generally no one is going to use the Observable class, they're almost always going to go through the method.

src/operator/combineLatest.ts
david-driscoll: Yeah, and we may need discussion of them as well.

src/operator/combineLatest.ts
david-driscoll: Naming is hard, I just picked something at the time that didn't clash.

I kind of like `CombineLatestSignature` it's descriptive as to what this interface really is about.

david-driscoll:
It only apply to methods that have a method that has some sort of applied / combined output or input.  Which is actually only a few, but they are the important ones to capture, because it makes type inference that much better.  Exaples are `combineLatest`, `zip`, `withLatestFrom`, etc.

The only other most common case is methods with an optional selector.  You'd want one overload without the selector, because generally those return the incoming type `T` in some form, and the one with the selector would return `R`.

src/add/operator/combineLatest.ts
luisgabriel: Hmn. Makes sense.

src/operator/combineLatest.ts
luisgabriel: +1 for `CombineLatestSignature`.

luisgabriel:
I don't now what do you guys think but I'd prefer to split this into two separate commits: one for `biundCallback` and another for `combineLatest`.

src/observable/bindCallback.ts
```diff
@@ -9,6 +9,24 @@ import {AsyncSubject} from '../subject/AsyncSubject';
...
+  static create<T, TResult>(callbackFunc: (v1: T, callback: (result: TResult) => any) => any, selector?: void, scheduler?: Scheduler): (v1: T) => Observable<TResult>;
+  static create<T, T2, TResult>(callbackFunc: (v1: T, v2: T2, callback: (result: TResult) => any) => any, selector?: void, scheduler?: Scheduler): (v1: T, v2: T2) => Observable<TResult>;
```
luisgabriel: I'd prefer to rename `TResult` to `R`. It's less verbose and keeps the same pattern we're using for the other operators.

src/observable/bindCallback.ts
kwonoj: Would you elaborate bit more? does that mean this signature update is not even required by `Observable` should have correct signature?

kwonoj:
> I'd prefer to split this into two separate commits: 

: +1 for those. Thanks for catching it.

kwonoj:
@david-driscoll Reason I asked above questions are feeling those will be raised at some point when anyone try to make changes to contribute, tried to play advocate for those perspective. As same as some guide we do have for contribution or migration, I think these also need to be noted for further reference / maintenance purpose. (including above questions + list of operators have those signatures, etcs)

src/observable/bindCallback.ts
kwonoj: +1 for now. I'm also quite familiar of `TResult` by living in the C# world, but let's make consistency first. If we'd like to have rename to this, can be discussed in separate PR.

src/operator/combineLatest.ts
kwonoj: `CombineLatestSignature` seems feasible, let's go with that.

cvrjd2oukaahvmy

src/add/operator/combineLatest.ts
david-driscoll: I've updated this in the last third commit as something separate from the other work. (Hence the larger number of files changed)

david-driscoll:
Updated.

src/Observable.ts
```diff
@@ -187,9 +194,7 @@ export class Observable<T> implements CoreOperators<T>  {
   bufferWhen: (closingSelector: () => Observable<any>) => Observable<T[]>;
   catch: (selector: (err: any, source: Observable<T>, caught: Observable<any>) => Observable<any>) => Observable<T>;
   combineAll: <R>(project?: (...values: Array<any>) => R) => Observable<R>;
-  combineLatest: <R>(...observables: Array<Observable<any> |
-                                     Array<Observable<any>> |
-                                     ((...values: Array<any>) => R)>) => Observable<R>;
+  combineLatest: CombineLatestSignature<this>;
```
kwonoj: How do we assign updated signature of `bindCallback` here for static creation method? seems this PR doesn't changes it in `Observable`.

david-driscoll: It's already bound, based on the static methods.

kwonoj: Just checked again and you're correct, sorry for misunderstanding.

src/Observable.ts
```diff
@@ -37,6 +37,13 @@ import {ErrorObservable} from './observable/throw';
 import {AjaxCreationMethod} from './observable/dom/ajax';
 import {WebSocketSubject} from './observable/dom/webSocket';
 
+import {CombineLatestSignature} from './operator/combineLatest';
+
+/* These types would be come standard types used through-out the library */
+export type ObservableOrPromise<T> = Observable<T> | Promise<T>;
```
kwonoj: Let's separate this. Thinking about PR like https://github.com/ReactiveX/RxJS/pull/1175, I think there might be enough common usages for these types. Those implementation only might need import this definition, not whole `Observable`.

david-driscoll: Separate commit or separate PR (to go first?)

kwonoj:
Change looks good to me in general, (might) need to separate common type definitions.

Still, separate from this PR itself which won't block check in this, I'd like to discuss how to organize some kind of document (or tutorial) for this kind of operators anyone who want to contribute either creating new one or modifying existing one.

src/Observable.ts
kwonoj: As long as we keep in mind further refactoring to use this common type in existing codebase either way's fine. but separate commit would be easier for you I assume? :)

david-driscoll: Not too worried about the work / effort, more just the ordering. :smile:

kwonoj:
cc @blesh since this PR's nearly complete review, giving visibility before actual check in happens. In my perspective this change's are good and only waiting one small update but there could be other suggestions that I was missing about.

david-driscoll:
@kwonoj added in the extra commit and have a PR for it, if we want a separate discussion.  See #1216

masaeedu:
@david-driscoll Why the `<any>` casts in the patch files? Not having it serves as a nice check to ensure the implementation satisfies the interface.

src/add/operator/combineLatest.ts
masaeedu: I'm not sure this should be needed. The implementation function should be assignable to the interface. If it doesn't do that, it means there's some overloads in the interface that the implementation isn't equipped to handle. 

In this case, the interface is advertising to users that they can pass in iterators, promises or arrays wherever observables are expected, however the function isn't written to expect these and will explode. This is causing the assignability error, which is a good thing.

masaeedu: I would recommend narrowing `ObservableInput<T>` to just `Observable<T>` until we can figure out a way to generically deal with iterators, arrays and promises in the actual implementations (which would be absolutely magical!)

david-driscoll:
With the separated signature the implementation no longer satisfies the interface :cry:

masaeedu:
Yeah, but if you change `ObservableInput<T>` from whatever it is now to just `Observable<T>`, it does. This is the reality of the implementations at the moment, but it would be amazing if we could update the implementations to deal with arraylikes, iterators and promises transparently. Possibly some kind of decorator for the arguments could be written.

david-driscoll:
I hadn't noticed that point... I'll take a look tonight.

kwonoj:
@david-driscoll , would you able to rebase this one? also do we have conclusions on `any` casting on patch operators?

david-driscoll:
@kwonoj rebased!  I was able to eliminate the <any> calls, so that's already been removed from the changes.

masaeedu:
@david-driscoll Right now the implementation for `combineLatest` accepts `Array<ObservableInput<any>>`. This won't work once the `lift` method is fixed, because you can't lift an `Observable<Observable<any> | Promise<any> | Iterator<any> | ArrayLike<any> ...>` through an `Operator<T, R>` and produce an `Observable<R>`. 

I'd suggest narrowing `ObservableInput` to `Observable` only for now, so that the API exposed to the user reflects what the operators can actually handle.

david-driscoll:
@masaeedu I don't see any cosntraints in the code that would cause any compilation problems.  To validate I merged with your changes (which merge cleanly btw! :heart: git).

Do you have a specific example where you have seen issues pop in?

masaeedu:
That's odd, it doesn't compile for me. The changes I'm making are to remove `T` from `Observable.lift` and removing `T` from `Operator.call`. Once you do that, you get an error on the last line of `combineLatest` explaining how `new ArrayObservable(observables).lift` expects an `Operator<Observable<any> | Promise<any> ...>`, which an `Operator<T, R>` does not satisfy. Could you try making those two changes and attempting a compile? I can provide a patch file if you want.

david-driscoll:
Want to push your local branch and I'll check it out and see if I can see what you're seeing?

masaeedu:
@david-driscoll Yup, here you go: https://github.com/masaeedu/RxJS/tree/typings-option-a. Look at the error for `combineLatest.ts`.

david-driscoll:
Found the problem.

Your changes removed `<T, R>`

david-driscoll:
Found the problem.

Your changes removed `<T, R>` from `new ArrayObservable(observables).lift(new CombineLatestOperator(project));` hence why I didn't the problem when merged. :smile: 

Now that said the real solution isn't actually anything that is wrong, the typings are just incorrect.

Currently ArrayObservable is typed like `ArrayObservable<Observable<any>>` or with my changes `ArrayObservable<ObservableInput<any>>`.  This isn't actually correct, instead ArrayObservable should container whatever type is contained by the inputs. (Observables, Promises, Iterables, Arrays, etc).

The best way I see to type around that is something like:

    new ArrayObservable<T>(<any[]>observables).lift<R>(new CombineLatestOperator<T, R>(project));

This produces the effect that we're looking for.   In call cases of `ArrayObservable`, should probably be typed similarly.

masaeedu:
@david-driscoll You have to remove at least `T` from `.lift`, since there's only one free generic parameter in `lift` after the fix. Specifying `R` is unnecessary since it can be inferred from the type of the `Operator` being passed in, but even if you specify it this doesn't fix the problem (at least on my setup).

masaeedu:
The reason I think the error is understandable and correct is that an `Operator<T, R>` can only turn an `Observable<T>` into an `Observable<R>`. The code does not exist to allow an operator to deal with `Promise`s or `Iterable`s or whatever else. With your typings, a user would assume they can pass promises or arrays into `combineLatest` where before `Observable`s were expected, but there is no additional code to deal with this and it will cause runtime errors.

david-driscoll:
Two different problems.

1) The problem you noted was dealing with an `Observable` of `Observable<T>`, which should only be `Observable<T>`.  Also for the code itself `T` isn't really used, so just confuses matters to be honest.

2) Once you dig into `combineLatest-support.ts` you can see that it just shifts each value from the `ArrayObservable` into an array.  From there on complete it calls through to `subscribeToResult`, which handles among other things Observables, Promises, Arrays and Iterables.

masaeedu:
Ah, I wasn't aware `subscribeToResult` transparently deals with other types of items. Sorry about that; I was assuming that because everything is typed to `Observable`, there was no code to deal with the other stuff. 

I actually have a version of `CombineLatestOperator` with updated typings [here](https://github.com/masaeedu/RxJS/blob/8a448ce26f8be52e9c09563162606b468cc1a80b/src/operator/combineLatest-support.ts), where the signature is:

```
class CombineLatestOperator<T extends Observable<U>, U, R> implements Operator<T, U[] | R>
```

I would just need to update this to:

```
CombineLatestOperator<T extends ObservableInput<U>, U, R> implements Operator<T, U[] | R>
```

I'll update stuff in my branch.

kwonoj:
I've already marked this as lgtm & latest update removes `<any>` casting resolves some of questions we had. I'll check this in around this afternoon if there's any other suggestions around.

blesh:
It looks like there are some merge conflicts... @david-driscoll can you rebase this?

@kwonoj, feel free to merge this whenever you think is appropriate if you feel this is an improvement to our typings.

kwonoj:
Will do. I've already completed reviewing this and so far I believe good to check in. I'll do it myself once it's rebased.

masaeedu:
Is this going to be a pattern that is followed uniformly for all the operators or just for the ones with complicated type signatures?

kwonoj:
@masaeedu Not all of operators. There are some operators allows these kind of n-numbers of parameter s, those need to be updated like zip, withlatestfrom (https://github.com/ReactiveX/RxJS/issues/1208 , https://github.com/ReactiveX/RxJS/issues/1209)

kwonoj:
@david-driscoll , I'll check this in once it's rebased. (Gentle reminder, not poking :wink: )

src/operator/combineLatest.ts
```diff
@@ -48,8 +69,25 @@ export function combineLatest<T, R>(...observables: Array<Observable<any> |
  * @returns {Observable} an observable of other projected values from the most recent values from each observable, or an array of each of
  * the most recent values from each observable.
  */
-export function combineLatestStatic<T, R>(...observables: Array<any | Observable<any> |
-                                                    Array<Observable<any>> |
+/* tslint:disable:max-line-length */
+export function combineLatestStatic<T>(v1: ObservableInput<T>): Observable<[T]>;
```
david-driscoll: Since these have been merged, I added definitions for `combineLatestStatic` that mirror the signatures.

david-driscoll:
@kwonoj rebased.

kwonoj:
Thanks! I'll check this in.

kwonoj:
Merged with 7c9547a, appreciate for huge effort @david-driscoll :smile:

PR #1087, opened 2015-12-16
generated typings for bindCallback.create

david-driscoll:
Comparing against manual maintenance of types to spur on discussion of the best approach.  I picked a single file here just as an example case.  See #1086

As a reference for TypeScript itself, these typings can't be simplified much more without losing the enhanced experience of operating against them.

https://github.com/Microsoft/TypeScript/issues/5453 may be the answer, not sure if it answers all the possible use cases.  It is currently on the road map for 2.0.

I realize there will be tons of opinions my _only_ goal is to make the TypeScript experience with RxJS5 as good as we have gotten the experience in RxJS4.


blesh:
cc @robwormald @jeffbcross @kwonoj 

... also @david-driscoll you'll probably want to rebase both of these.

david-driscoll:
@blesh rebased.

kwonoj:
Apologizes for making this PR long-hanging. I'll start look into this and will provide some inputs if I have :)

Issue #1053, opened 2015-12-14
Rename `sample` operators to `inspect`, reimplement `sample` operator


blesh:
Per some offline discussion with @headinthebox, @mattpodwysocki and @staltz... the current implementation of `sample` as discussed in #150 is probably incorrect.

Basically, in the original `sample` operator ala RxJS 4, the emission of a value meant that a new value had _arrived_ on the source. If the notifier notified, and a new value *hadn't* arrived, nothing would be emitted. This means the very emission of the value has some meaning/connotations WRT the source observable.

... _that said_ ... I think @staltz raised some interesting points about how people might expect a sample operator to work, or needs those people might have.

... so rather than throw away the current `sample` and `sampleTime` implementations, I'd like to rename them to `inspect` and `inspectTime` and then reimplement the `sample` operators in RxJS 4's likeness.

staltz:
I'd question the usefulness of both `inspect` and `sample` flavors, since we're just adding more operators.

staltz:
@blesh There's also the issue with `sample`/`inspect` that the completion of the sampler Observable should "sample" the sampled Observable, like Erik mentioned. That might be important for beta.

blesh:
> There's also the issue with sample/inspect that the completion of the sampler Observable should "sample" the sampled Observable, like Erik mentioned. That might be important for beta.

:+1: good catch. I don't think I included this in my PR. I'll add that.

blesh:
> I'd question the usefulness of both inspect and sample flavors, since we're just adding more operators.

As long as `inspect` is in KitchenSink, and we're going with this modular design. I don't think it matters.

If it came down to removing one, though, I'd remove `inspect`. Only because it could be composed from `sample`, where the reverse isn't really possible.

staltz:
I'm curious, how would `inspect` be written using `sample` and others?

blesh:
Thinking quickly: Something like... `notifier.withLatestFrom(source.sample(notifier), (_, a) => a)`? That might not be _perfect_, but it's an approximation, I'd guess. The emitting on completion behavior of `sample` might make it weird, but you could compose around that too.

staltz:
Why not just `notifier.withLatestFrom(source, (_, a) => a)`?

blesh:
> Why not just notifier.withLatestFrom(source, (_, a) => a)?

I don't know? lol.

staltz:
[...]

PR #1088, opened 2015-12-17
fix(inspect): remove inspect and inspectTime operators

staltz:
Remove inspect and inspectTime operators from this library, since they are unnecessary and don't
exist in previous versions of RxJS.

This change was suggested in issue #1053.

BREAKING CHANGE:
`inspect` and `inspectTime` were removed. Use `withLatestFrom` instead.


kwonoj:
I'll check in this changes around late today to tomorrow, general agreement's already made in issue.

kwonoj:
Merged with 17341a4, thanks @staltz and sorry for delay. Long holiday vacation started.. :sunglasses:

PR #1090, opened 2015-12-17
test(delay): use time() test helper in delay() tests

staltz:


kwonoj:
Merged with 3eebf50. Thanks @staltz (and have happy holiday)

Issue #1075, opened 2015-12-15
Implement Observable.amb (as Observable.race)


saneyuki:
We still don't have [`Observable.amb()`](http://reactivex.io/documentation/operators/amb.html) operator.

This operator is useful in these case to take a 1st returned value like `Promise.race()` with dispatching some operations optimistically.

blesh:
I discussed this one with @trxcllnt a while back. I think we even postulated that it should be named `race` instead of `amb`.  @mattpodwysocki what are your thoughts?

mattpodwysocki:
@blesh [`Observable.race`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) is ok with me given the `Promise.race`

justinwoo:
Now in master at https://github.com/ReactiveX/RxJS/commit/ee3b593996096c90a07670b728835424717e9936

kwonoj:
Closing issue by changes are checked in.

PR #1091, opened 2015-12-17
feat(race): add race operator

justinwoo:
Implements race operator per #1075 

Hopefully it works! (I added test cases I could find and think of)


spec/observables/race-spec.js
```diff
@@ -0,0 +1,160 @@
+/* globals expect, it, describe, hot, cold, expectObservable */
+
+var Rx = require('../../dist/cjs/Rx');
+var Observable = Rx.Observable;
```
staltz: Great looking tests. Can you also add the cases "should allow unsubscribing early" and "should not break unsubscription chain when unsubscribing"

src/operator/race-support.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (!this.hasFirst) {
+      this.hasFirst = true;
+      this.first = outerIndex;
+    }
+
+    if (outerIndex === this.first) {
+      this.destination.next(value);
+    }
+  }
+}
```
staltz: Good looking implementation. :+1:

spec/observables/race-spec.js
justinwoo: Sure, added now.

spec/observables/race-spec.js
```diff
@@ -0,0 +1,176 @@
...
+
+  it('should not break unsubscription chains when unsubscribed explicitly', function () {
+    var e1 =  cold('---a-----b-----c----|');
+    var e1subs =   '^           !';
+    var e2 =   hot('------x-----y-----z----|');
+    var e2subs =   '^           !';
+    var expected = '---a-----b---';
+    var unsub =    '            !';
+
+    var result = Observable.race(e1, e2);
```
staltz: This test should not be just a copy of the previous. You need to wrap e1 and e2 in a "mergeMap" sandwich, and also put a mergeMap after race. The idea is that unsubscribing from a mergeMap operator should send a message to race operator to unsubscribe which should send a message to a mergeMap operator to unsubscribe which in turn should send a message to e1 and e2 to unsubscribe.

justinwoo: Oops, worked off of the wrong part. Updated now.

staltz:
Ok, looks good to merge already for me.

src/operator/race-static.ts
```diff
@@ -0,0 +1,16 @@
...
+import {RaceOperator} from './race-support';
+import {ArrayObservable} from '../observable/fromArray';
+
+/**
+ * Returns an Observable that mirrors the first source Observable to emit an item.
+ * @param {...Observables} ...observables sources used to race for which Observable emits first.
+ * @returns {Observable} an Observable that mirrors the output of the first Observable to emit an item.
+ */
+export function race<T>(...observables: Array<Observable<any>>): Observable<T> {
+  if (observables.length === 1) {
```
trxcllnt: @justinwoo could you add support for a single `Array<Observable>` argument also? Similar to [combineLatest-static](https://github.com/ReactiveX/RxJS/blob/master/src/operator/combineLatest-static.ts#L35), this would bring it in line with the [old signature](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/amb.md) of Observable.amb().

justinwoo: Sure, updated and pushed now.

src/operator/race-support.ts
```diff
@@ -0,0 +1,61 @@
...
+      if ((this.active -= 1) === 0) {
+        this.destination.complete();
+      }
+    }
+  }
+
+  notifyNext(observable: any, value: R, outerIndex: number): void {
+    if (!this.hasFirst) {
+      this.hasFirst = true;
+      this.first = outerIndex;
```
trxcllnt: @justinwoo once we've found the Observable that wins, don't we want to unsubscribe from all the rest? Or is that happening somewhere else and I've just missed it?

src/operator/race-static.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @param {...Observables} ...observables sources used to race for which Observable emits first.
+ * @returns {Observable} an Observable that mirrors the output of the first Observable to emit an item.
+ */
+export function race<T>(...observables: Array<Observable<any> | Array<Observable<any>>>): Observable<T> {
+  // if the only argument is an array, it was most likely called with
+  // `pair([obs1, obs2, ...])`
+  if (observables.length === 1 && isArray(observables[0])) {
+    observables = <Array<Observable<any>>>observables[0];
+  }
+  if (observables.length === 1) {
```
blesh: We should probably assert length === 1 only once here.

src/operator/race-support.ts
blesh: Agreed. And that would mean you no longer need to memoize the index of the "winner" in `first`. Further, the "winner" could be stored, and the other observables could be garbage collected along with the array they're in.

This makes me think of another thing, too... Should this be derived from a `raceAll`?  Thoughts anyone?

justinwoo: I could try making it this way. I originally tried but didn't really know how to make eager unsubscription worked, since I was trying to call the method but it was failing in the unit tests. Sounds reasonable to unsub the rest instead of just keeping them like the other combination ops.

justinwoo:
Not sure what that error is about, I don't see any problems with any other occurrences of `Subscription<any>`.

Added a commit to make the operator unsubscribe all of the other subscriptions which then removes the need to have the first observable's index around.

src/operator/race-support.ts
```diff
@@ -0,0 +1,62 @@
...
+export class RaceOperator<T, R> implements Operator<T, R> {
+  call(subscriber: Subscriber<T>): Subscriber<T> {
+    return new RaceSubscriber(subscriber);
+  }
+}
+
+export class RaceSubscriber<T, R> extends OuterSubscriber<T, R> {
+  private hasFirst: boolean = false;
+  private observables: Observable<any>[] = [];
+  private subscriptions: Subscription<any>[] = [];
```
luisgabriel: `Subscription` does not have type parameter anymore (see #1065).

justinwoo: Ohh, didn't know this went in already. Thanks!

src/operator/race-support.ts
blesh: For eager unsub, all you need is to unsubscribe the other inner subscribers (so you'd keep them in an array or something), then remove those instances from the outer subscriber.

blesh: If it helps you: For parity with Rx4, it should unsubscribe from the other inner subscriptions _before_ it emits the value from the "winning" observable.

justinwoo: I have this on this branch now: https://github.com/justinwoo/RxJS/blob/d4cdfc63e52b6e9feb15d0ea39be121b69444afd/src/operator/race-support.ts#L52

Holds on to subscriptions and then when a winner is selected, unsubscribes all of the other subscriptions and removes them before emitting.

blesh:
@justinwoo ... looks good, just needs rebased.

justinwoo:
Rebased. Updated to use typeof like the other static methods.

Should be clean when this completes on travis.

blesh:
Oops... last thing, @justinwoo, can you flatten your commits into just one commit to add the race operator?

justinwoo:
Oh, right. Forgot about that. Done.

luisgabriel:
I think this PR is not blocked anymore.

kwonoj:
Yes, seems so. Will check this in around late today to weekend.

luisgabriel:
I think this PR is not blocked anymore.

kwonoj:
@luisgabriel clicked comment btn twice? ;)

luisgabriel:
@kwonoj oops. conection problems. sorry ;)

kwonoj:
Merged with ee3b593, thanks @justinwoo !

PR #1094, opened 2015-12-18
docs(buffer): add PNG diagrams to JSDoc for buffer operators

staltz:
Add PNG marble diagram images on the inline documentation for each buffer-like operator, and fix
some typos and mistakes with JSDoc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "cover_remapping": "remap-istanbul -i coverage/coverage.json -o coverage/coverage-remapped.json && remap-istanbul -i coverage/coverage.json -o coverage/coverage-remapped.lcov -t lcovonly && remap-istanbul -i coverage/coverage.json -o coverage/coverage-remapped -t html",
     "test": "jasmine",
     "test_karma": "karma start karma.conf.js",
-    "tests2png": "mkdirp img && JASMINE_CONFIG_PATH=spec/support/tests2png.json jasmine",
+    "tests2png": "mkdirp tmp/docs/img && JASMINE_CONFIG_PATH=spec/support/tests2png.json jasmine",
     "watch": "watch \"echo triggering build && npm run build_test && echo build completed\" src -d -u -w=15",
     "perf": "protractor protractor.conf.js",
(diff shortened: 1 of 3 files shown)


staltz:
@kwonoj This PR works as an example of how should we write the docs for all operators so that they have their marble diagram images embedded.

luisgabriel:
:+1:

kwonoj:
This one's cool. One thing - tried locally and could see marble diagram generation is not yet included in docgen process, is my understanding correct?

staltz:
Ah yeah, it's not yet included in docgen process.

kwonoj:
What about add those changes also in this PR? It may possible some people think img links are broken if this changes are checked in and docs published before image's actually generated.

staltz:
Yes I'll add them.

staltz:
@kwonoj I added PNG diagram generation to `docgen.sh`.

kwonoj:
Tried PR locally and seeing img links still not displayed - docs are placed under `docs` and looks for `docs/img` and seems images are not copied under `docs`.

staltz:
That's because `docgen.sh` moves index.html from `docs` to `./` (root).

kwonoj:
Shouldn't image also need to be copied accordingly in this case? (Not sure, checking it - I don't want to check in those into `gh-pages` myself to only check its results)

staltz:
Good point.

staltz:
Alright, fixed @kwonoj

kwonoj:
Cool, tried locally and works as expected. 

Change looks good to me. I'll check this around late today if @blesh does not have any other suggestions.

kwonoj:
Merged with abbf4e7, thanks @staltz

staltz:
@kwonoj can you run `docgen.sh` to update http://reactivex.io/RxJS ?

kwonoj:
@staltz Yes I can (technically). What about run it once after https://github.com/ReactiveX/RxJS/pull/1157 is checked in?

staltz:
Yeah that's ok

PR #1097, opened 2015-12-18
auto-patching operators: Fixes and updates

luisgabriel:
Closes to #1030.


luisgabriel:
Let me do some more tests before merging it. ;)

kwonoj:
Suuuuure. Is this PR closes issue completely which includes injecting comments also?

luisgabriel:
@kwonoj Not yet. I'll add some commits injecting the comments.

luisgabriel:
@kwonoj Now it's complete. ;)

tools/generate-operator-patches.ts
```diff
@@ -146,15 +163,16 @@ function checkForCreate (op: OperatorWrapper): boolean {
 }
 
 if (process.argv.find((v) => v === '--exec')) {
-  mkdirp('./src/add/operator/extended', () => {
+  mkdirp('./src/add/operator', () => {
+    var kitchenSinkContents = fs.readFileSync('./src/Rx.KitchenSink.ts');
```
kwonoj: can this be `const` also?

tools/generate-operator-patches.ts
```diff
@@ -57,19 +58,26 @@ const AliasMethodOverrides = {
 };
 
 function generateNewOperatorFileContents (op:OperatorWrapper): OperatorWrapper {
+  var header = `/**
```
kwonoj: can these variable can be `const`? I know it's script and not for actual codebases though.

tools/generate-operator-patches.ts
```diff
@@ -57,19 +58,26 @@ const AliasMethodOverrides = {
...
-${op.isExtended ? 'import {KitchenSinkOperators} from \'../../../Rx.KitchenSink\';' : ''}`;
-  var extendedProto = `${op.isExtended ? 'const observableProto = (<KitchenSinkOperators<any>>Observable.prototype);' : ''}`
+  var imports = `import {Observable} from '../../Observable';
+import {${op.exportedFnName}} from '../../operator/${op.path.replace('.ts','')}';
+${op.isExtended ? 'import {KitchenSinkOperators} from \'../../Rx.KitchenSink\';' + '\n': ''}`;
+  var extendedProto = 'const observableProto = (<KitchenSinkOperators<any>>Observable.prototype);'
   var patch = op.aliases.map((alias) => {
     return `${baseObject}.${optPrototype}${alias} = ${op.exportedFnName};`;
   }).join('\n');
+  var typescriptHack = 'export var _void: void;'
```
kwonoj: just an idea, what about add comment, or injected comment to explain why 'hack' is introduced for further reference?

tools/generate-operator-patches.ts
```diff
@@ -77,16 +85,24 @@ ${op.isExtended ? 'import {KitchenSinkOperators} from \'../../../Rx.KitchenSink\
 }
 
 function generateNewObservableFileContents (op:OperatorWrapper): OperatorWrapper {
+  var header = `/**
```
kwonoj: this seems repeating in line 61, can this be reduced?

tools/generate-operator-patches.ts
luisgabriel: Sure. I'll change this one. I did not touch the others because they're not related to the patch.

tools/generate-operator-patches.ts
luisgabriel: Ok.

tools/generate-operator-patches.ts
luisgabriel: Good idea. Since this hack is spread over almost all files (not only the auto-generated ones), I'll add a comment in this script instead of injecting the comment in the final patching files.

luisgabriel:
@kwonoj Just updated the PR addressing you comments.

kwonoj:
I think change's good, thinking which changes should be checked in first between https://github.com/ReactiveX/RxJS/pull/1077 and this. Still need some comment update from those PR regarding patch operators.

luisgabriel:
I think it doesn't make a difference which one gets merged first. If it is necessary to keep the change regarding the patch operators on #1077, it will require a separate PR to change the generator script.
What do you think?

kwonoj:
> I think it doesn't make a difference which one gets merged first.

: Yes I agree, just work order in my preferences :)

luisgabriel:
Given the size of #1077, maybe it would be better to merge it first as is. Then I can rebase and update this PR.

luisgabriel:
I'll update this PR as soon as #1091 gets merged.

kwonoj:
@luisgabriel , PR #1091 's checked in to allow this PR can be updated. Thanks for take care of those. Let me mark as blocked for now to avoid possible confusions.

luisgabriel:
Thanks, @kwonoj! I'll rebase and update this PR tomorrow.

luisgabriel:
Just updated this PR. Now it includes also `pairwise`, `race` and `let` that were recently checked in.

kwonoj:
Merged with 5e43f14, appreciate for effort @luisgabriel :)

PR #1101, opened 2015-12-20
fix(fromEvent): added spread operator for events that emit more than one argument

earlsioson:
I added the spread operator to the first expression of the ternary operator, i.e. the condition when a selector is passed. I'm not sure if it makes sense to add it to the second expression. I also modified the spec to test for multiple arguments.

### DIFF
--- a/spec/observables/fromEvent-spec.js
+++ b/spec/observables/fromEvent-spec.js
@@ -127,4 +127,75 @@ describe('Observable.fromEvent', function () {
 
     send('test');
   });
+
+  it('should not fail if no event arguments are passed and the selector does not return', function (done) {
+    var send;
+    var obj = {
+      on: function (name, handler) {
+        send = handler;
+      },
+      off: function () {
+      }
+    };
+
+    function selector() {
+    }
+
+    Observable.fromEvent(obj, 'click', selector).take(1)
+      .subscribe(function (e) {
+        expect(e).toBeUndefined();
+      }, function (err) {
+        done.fail('should not be called');
+      }, done);
+
+    send();
+  });
+
+  it('should return a value from the selector if no event arguments are passed', function (done) {
+    var send;
+    var obj = {
+      on: function (name, handler) {
+        send = handler;
+      },
+      off: function () {
+      }
+    };
+
+    function selector() {
+      return 'no arguments';
+    }
+
+    Observable.fromEvent(obj, 'click', selector).take(1)
+      .subscribe(function (e) {
+        expect(e).toBe('no arguments');
+      }, function (err) {
+        done.fail('should not be called');
+      }, done);
+
+    send();
+  });
+
+  it('should pass multiple arguments to selector from event emitter', function (done) {
+    var send;
+    var obj = {
+      on: function (name, handler) {
+        send = handler;
+      },
+      off: function () {
+      }
+    };
+
+    function selector(x, y, z) {
+      return [].slice.call(arguments);
+    }
+
+    Observable.fromEvent(obj, 'click', selector).take(1)
+      .subscribe(function (e) {
+        expect(e).toEqual([1,2,3]);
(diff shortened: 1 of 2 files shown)


spec/observables/fromEvent-spec.js
```diff
@@ -114,17 +114,17 @@ describe('Observable.fromEvent', function () {
       }
     };
 
-    function selector(x) {
-      return x + '!';
+    function selector(x, y, z) {
```
kwonoj: Would you able to expand test cases instead of modifying it, such as testing
- emitted none via fromEvent (zero maybe?)
- emitted 1 via fromEvent
- emitted >1 via fromEvent

kwonoj:
> I added the spread operator to the first expression of the ternary operator, i.e.

: What does behavior difference when selector is not being passed between RxJS4 / RxJS5?

spec/observables/fromEvent-spec.js
earlsioson: Yes, I will expand the test cases per your suggestion.

earlsioson:
@kwonoj,

>What does behavior difference when selector is not being passed between RxJS4 / RxJS5?

for the code below both RxJS4 and RxJS5 display x:baz

```javascript
var EventEmitter = require('events').EventEmitter;
//const Rx = require('@reactivex/rxjs'); //RxJS5
const Rx = require('rx'); //RxJS4

var eventEmitter = new EventEmitter();

var source = Rx.Observable.fromEvent(
  eventEmitter,
  'data');
  
var subscription = source.subscribe(
  function (x) {
    console.log('x:' + x);
  },
  function (err) {
    console.log('Error: ' + err);
  },
  function () {
    console.log('Completed');
  });

eventEmitter.emit('data', 'baz', 'hi');
```

spec/observables/fromEvent-spec.js
earlsioson: @kwonoj, I updated with 3 additional test cases.

kwonoj:
Change looks good to me.

spec/observables/fromEvent-spec.js
```diff
@@ -127,4 +127,75 @@ describe('Observable.fromEvent', function () {
...
+      }
+    };
+
+    function selector(x, y, z) {
+      return x + y + z;
+    }
+
+    Observable.fromEvent(obj, 'click', selector).take(1)
+      .subscribe(function (e) {
+        expect(e).toBe(6);
```
blesh: Can we assert the arguments and the order thereof, rather than adding them together? That's the important part of the test.

earlsioson: @blesh,

Yes, I will do that.

earlsioson: @blesh,

Updated the test per your suggestion.

spec/observables/fromEvent-spec.js
```diff
@@ -127,4 +127,81 @@ describe('Observable.fromEvent', function () {
...
+      off: function () {
+      }
+    };
+
+    function selector(x, y, z) {
+      return {
+        arg1: x,
+        arg2: y,
+        arg3: z
+      };
```
blesh: I was thinking more like:  `return [].slice.call(arguments)` here, and a check of `expect(e).toEqual([1,2,3])` below because what's currently here won't catch a bug where additional arguments are unintentionally passed. 

Nitpicky, I know. :)

I really appreciate your effort on this PR.

earlsioson: @blesh,

I can appreciate the nitpickyness :) Am happy to contribute. Updated per your suggestion.

kwonoj:
I'll check this in around today.

kwonoj:
Merged with 3f8eabb, thanks @earlsioson

PR #1105, opened 2015-12-22
feat(Observable): add let to allow fluent style query building

kevinvandervlist:
Add let (and letBind for legacy browsers) operators, exactly like currently available in RxJS 4
(https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/let.md)

I don't know if this operator is left out intentionally, but I like having it in RxJS 4 and it is missing here. Any specific reasons why it is not included, or has it just not been picked up yet?

Here's a pull request implementing the feature in the same way is it works in RxJS 4. Simple, but really useful.

### DIFF
--- a/src/add/operator/let.ts
+++ b/src/add/operator/let.ts
@@ -1,0 +1,4 @@
+import {Observable} from '../../Observable';
+import {letProto} from '../../operator/let';
+Observable.prototype.let = letProto;
+Observable.prototype.letBind = letProto;
--- a/tools/generate-operator-patches.ts
+++ b/tools/generate-operator-patches.ts
@@ -42,12 +42,14 @@ const SpecialCasePrototypes = {
   'switch.ts': 'switch',
   'do.ts': 'do',
   'finally.ts': 'finally',
-  'catch.ts': 'catch'
+  'catch.ts': 'catch',
+  'let.ts': 'let'
 };
 
 const AdditionalAliases = {
   'mergeMap.ts': ['flatMap'],
-  'fromArray.ts': ['of']
+  'fromArray.ts': ['of'],
+  'let.ts': ['letBind']
 };
 
 const AliasMethodOverrides = {
--- a/spec/operators/let-spec.js
+++ b/spec/operators/let-spec.js
@@ -1,0 +1,24 @@
+/* globals describe, it, expect */
+var Rx = require('../../dist/cjs/Rx');
+var Observable = Rx.Observable;
+
+describe('let', function () {
+  it('should be able to compose with let', function (done) {
+    var expected = ['aa', 'bb'];
+    var i = 0;
+
+    var foo = function (observable) {
+      return observable
+        .map(function (x) {
+          return x + x;
+        });
+    };
+
+    Observable
+      .fromArray(['a','b'])
+      .let(foo)
+      .subscribe(function (x) {
+        expect(x).toBe(expected[i++]);
+      }, null, done);
+  });
+});
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -223,6 +223,8 @@ export class Observable<T> implements CoreOperators<T>  {
   last: <R>(predicate?: (value: T, index: number) => boolean,
             resultSelector?: (value: T, index: number) => R,
             defaultValue?: any) => Observable<T> | Observable<R>;
+  let: <T, R>(func: (selector: Observable<T>) => Observable<R>) => Observable<R>;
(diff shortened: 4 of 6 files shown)


kwonoj:
As you may already aware, lot of operator's are currently not migrated from RxJS4 for reasons of it is just not landed yet, or some of operators can be simply composed so may not be required. Details can be found at https://github.com/ReactiveX/RxJS/blob/master/MIGRATION.md#unimplemented-operatorsfeatures.

kevinvandervlist:
Yeah, I saw the section in the migration guide. I just didn't see whether this was a conscious decision or just a matter of not getting around to it. That's why I opened this PR :)

blesh:
@kevinvandervlist it hasn't come up as a requirement, so it hasn't been discussed.

robwormald:
neat, i hadn't played with this one before. @kevinvandervlist out of sheer curiosity, what sort of situations do you use this in? you mention 'fluent query building' - what does that look like?

kevinvandervlist:
I normally use this to take reuseable parts out of a composition so I end up with less duplication, or to end up with more readable higher-level code.

A simple example in JS:

```js
// This can obviously be reused, put in a library, whatever
function asFahrenheit(obs) {
  return obs
    .map(function(c) {
      return (c - 32) * 5/9;
    });
}

var readings = ...; // An observable with temperature readings in celcius
// This is meant by the fluent style query building
readings.let(asFahrenheit).subscribe(...);
```

In RxJava they are called [Transformers](http://reactivex.io/RxJava/javadoc/rx/Observable.Transformer.html), see [the wiki](https://github.com/ReactiveX/RxJava/wiki/Implementing-Your-Own-Operators#transformational-operators) for details.

robwormald:
very cool, thanks for the example. :+1:

trxcllnt:
@kevinvandervlist thanks for the PR, I've been meaning to get this in! it's so simple, I've been poly-filling it in my projects instead of committing it >.<

blesh:
> I've been poly-filling it in my projects instead of committing it

ಠ_ಠ

LOL

src/add/operator/let.ts
```diff
@@ -0,0 +1,6 @@
+import {Observable} from '../../Observable';
+import {letProto} from '../../operator/let';
+Observable.prototype.let = letProto;
+Observable.prototype.letBind = letProto;
```
luisgabriel: This file is auto-generated by a script. In order to generate it properly next time the script is executed you have to:

1. Put the real name of the operator as `let` instead of `letProto` ([generate-operator-patches.ts:40](https://github.com/ReactiveX/RxJS/blob/master/tools/generate-operator-patches.ts#L40))
2. Add `letBind` as an alias ([generate-operator-patches.ts:48](https://github.com/ReactiveX/RxJS/blob/master/tools/generate-operator-patches.ts#L48))

kevinvandervlist:
Hi @luisgabriel,

Thanks for the remarks. I did not rename the operator to 'let', since its reserved and I got an error about it (*error TS1214: Identifier expected. 'let' is a reserved word in strict mode. Modules are automatically in strict mode.*), but I did include it in the generator. 

I also discovered two small issues with the generator, I'll add those as comments to the currently open issue about the autogenerated headers (https://github.com/ReactiveX/RxJS/issues/1030 -- I missed those as well as you probably figured ;)).

tools/generate-operator-patches.ts
```diff
@@ -42,12 +42,14 @@ const SpecialCasePrototypes = {
...
-  'catch.ts': 'catch'
+  'catch.ts': 'catch',
+  'let.ts': 'let'
 };
 
 const AdditionalAliases = {
   'mergeMap.ts': ['flatMap'],
-  'fromArray.ts': ['of']
+  'fromArray.ts': ['of'],
+  'letBind': ['let']
```
luisgabriel: I think this should be: `'let.ts': ['letBind']`

kevinvandervlist:
Thanks @luisgabriel, that was a mistake. Fixed.

kwonoj:
Changes looks good to me and I think it's good to be checked in - I'll wait bit more to see any other opinions around and check in around this weekend. cc @blesh @luisgabriel for notification since this PR was underwater for some days.

luisgabriel:
LGTM

kwonoj:
Merged with 5a2014c, thanks @kevinvandervlist ! Also side note - your PR achieved exact 2000th test case in repo :)

image

PR #1106, opened 2015-12-25
The Subscriber refactor, operator updates, and scheduling fixes.

trxcllnt:
attention: @blesh @staltz @kwonoj 

1. Factors out the shared `Subscription` list from `Subscribers` and updates the `Observable`, `Subject`, and `Operator` implementations accordingly.
2. Refactors some operators that manage inner `Subscriptions` but weren't extending `OuterSubscriber` or using our `subscribeToResult` util. The more operators that use this pattern, the more operators that will interop with Array/Iterable/Promise/Observable-esque instances.
3. Includes fixes to our Scheduler Action implementations to ensure:
  - Actions scheduled to occur on the next tick still execute if the first scheduled action is canceled
  - Asap/Queued Actions that reschedule with a delay fallback to using `setTimeout` instead of rescheduling like they normally would.
4. fixes ReplaySubject to route messages through its scheduler (if specified), which mimics current RxJS v4 behavior.
5. fixes `Observables` and operators that default to the `queue` scheduler. Nothing should *ever* default to the `queue` scheduler, because
  - our Observables fall back to iterative loops if unspecified
  - timing operators that don't ensure their relative times are greater than zero will schedule and execute their logic synchronously. While not technically a bug, this is a horrible thing to do to an end user. Default to `asap` scheduler for timing operators (like `delay`).
6. Ensures all tests are run **asynchronously**, because Jasmine has difficulty with the boundaries between synchronous and asynchronous tests, and will non-deterministically fail for no reason.
7. Style updates:
  - `null` instead of `void 0` expressions
  - use our utility functions in more places
  - use more inherited properties instead of redefining them (in Subjects, operators)
  - call global functions from the root (`root.setTimeout` instead of `setTimeout`)

### DIFF
--- a/spec/Observable-spec.js
+++ b/spec/Observable-spec.js
@@ -121,13 +121,6 @@ describe('Observable', function () {
       expect(unsubscribeCalled).toBe(true);
     });
 
-    it('should return the given subject when called with a Subject', function () {
-      var source = Observable.of(42);
-      var subject = new Rx.Subject();
-      var subscriber = source.subscribe(subject);
-      expect(subscriber).toBe(subject);
-    });
-
     describe('when called with an anonymous observer', function () {
       it('should accept an anonymous observer with just a next function', function () {
         Observable.of(1).subscribe({
--- a/spec/Subject-spec.js
+++ b/spec/Subject-spec.js
@@ -21,7 +21,7 @@ describe('Subject', function () {
 
   it('should have the rxSubscriber Symbol', function () {
     var subject = new Subject();
-    expect(subject[Rx.Symbol.rxSubscriber]()).toBe(subject);
+    expect(typeof subject[Rx.Symbol.rxSubscriber]).toBe('function');
   });
 
   it('should pump values to multiple subscribers', function (done) {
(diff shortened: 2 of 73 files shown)


blesh:
FWIW: I've already reviewed most of this over the last few days. I'll give it a final review over the next few days.  It's Christmas, so it's probably going to be more later than sooner.

blesh:
> Also this commit is a fix, not a refactor.

We'll need to update the CHANGELOG manually with what's been fixed, because the changes required to fix those issues were broad, required multiple commits, and weren't necessarily linked directly to those issues

blesh:
Awesome work, @trxcllnt... as always. This is merged as of de31f32... 

Note: @kwonoj don't let me forget to update the CHANGELOG manually (or perhaps add a commit to do so) so the fixes for other issues are shown.

kwonoj:
OK, will keep eye `๏_๏` on @blesh to see CHANGELOG is coming, send some poke around if it doesn't arrive. (feeling having separate commit would be feasible for this case?)

PR #1111, opened 2015-12-29
feat(Observable): add doOn{Next,Error,Completed} operators

kevinvandervlist:
Added said operators for easier 'do' usage.

I find myself to use doOnError or doOnCompleted() quite often, and providing null parameters to the generic 'do' is less readable in my opinion. Every other implementation has them as well, lets complete the do* family.

If this is against the small api surface/less core operators vision, the PR can be rejected.


luisgabriel:
The doOn* operators can be replaced by passing an object to `do()`:

```javascript
obs.do({ error: (err) => /* stuff */ })
```

staltz:
Thanks @kevinvandervlist for the efforts, but I have to agree with Luis on this one.

kevinvandervlist:
No worries, didn't know that was possible :) Thanks!

blesh:
Thanks again, @kevinvandervlist, for the PR! And even more, thanks for being so cool about closing it when we realized it wasn't necessary.  10/10 :+1:

PR #1115, opened 2015-12-30
feat(ajax): adds ajax methods from rx-dom.

trxcllnt:


mattpodwysocki:
@trxcllnt why is this necessary to be in core versus just as a separate module?

blesh:
@mattpodwysocki .... I assume by "separate module" you mean "different npm package"? or "different repo"?

Now that everything is modular, there's really no reason to have them all separated. There are some consistency benefits to having them all in one repository.

mattpodwysocki:
@blesh @trxcllnt I'm less concerned about the bringing in what you need, I'm more interested in the versioning and the separation of concerns. By default, RxJS is a platform neutral library with no real hooks into the DOM (outside of schedulers).  

Instead, I'd like to see this as a separate module that can age/version on its own and be combined with other DOM features.  That and bringing Rx via NPM or other means won't download things you won't ever use, such as Ajax, especially if you're using it on the server-side.  Just adds extra bloat to the node_modules folder without any real gain.

blesh:
I'd like to keep everything in this repository. Whether or not they should be in separate npm packages is subject for debate. [Here's a decent article on the advantages of the monolithic repository](http://danluu.com/monorepo/). While, I don't want to go Full-Facebook on it. I think it has merit on a smaller scale.

The HARD part of this is figuring out the best way to manage supporting 3 different delivery styles: CJS, ES6 and globals if we're making these separate packages.

blesh:
@trxcllnt for now I've merged this into a new branch [rxjs-dom](https://github.com/ReactiveX/RxJS/tree/rxjs-dom). I'm going to continue working against that to fix a few minor issues and refactor a bit to match the rest of the library. We'll merge it over when we decide how we're going to organize it.

PR #1116, opened 2015-12-30
refactor(subscriber): Make Subscriber and derived classes' internal methods protected

saneyuki:
This change make these members `protected`:

- `Subscriber._next()`
- `Subscriber._error()`
- `Subscriber._complete()`

At current codebase, these members is only used in drived class of
`Subscriber` or itself. Thus we should hide them from public.
(If we need to make them public, then we should make them so).

## Exception
- `WindowSubscriber`'s members are used from
  `WindowClosingNotifierSubscriber`. So `WindowSubscriber`'s members are
  not `protected`.
    - These classes are a private class of `window.ts` modules.
    - I think this exception would not be a serious problem.

### DIFF
--- a/src/InnerSubscriber.ts
+++ b/src/InnerSubscriber.ts
@@ -8,16 +8,16 @@ export class InnerSubscriber<T, R> extends Subscriber<R> {
     super();
   }
 
-  _next(value: R) {
+  protected _next(value: R) {
     this.parent.notifyNext(this.outerValue, value, this.outerIndex, this.index++);
   }
 
-  _error(error: any) {
+  protected _error(error: any) {
     this.parent.notifyError(error, this);
     this.unsubscribe();
   }
 
-  _complete() {
+  protected _complete() {
     this.parent.notifyComplete(this);
     this.unsubscribe();
   }
--- a/src/Subscriber.ts
+++ b/src/Subscriber.ts
@@ -63,16 +63,16 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
     super.unsubscribe();
   }
 
-  _next(value: T): void {
+  protected _next(value: T): void {
     this.destination.next(value);
   }
 
-  _error(err: any): void {
+  protected _error(err: any): void {
     this.destination.error(err);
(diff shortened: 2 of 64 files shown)


kwonoj:
in case of `windowSubscriber` it could be work like this if we want to strict manner of interface,

```
class WindowSubscriber {
  publicerrorfunction(value: T) {
    doThing();
  }
  protected _error(unused) { 
    //noop
  }
}

class WindowClosingNotifierSubscriber {
  protected _error (err: any) {
   this.parent.publicerrorfunction(err);
  }
}
```

But not sure if this is really needed though. (this can be issue if there's strict compiler / lint rules to have explicit interfaces, but I'm not sure if those are even exists for now).

kwonoj:
@saneyuki , would you able to rebase this one?

Also, I think this change's ok in general, would like to hear opinion regarding above way of ensuring strict interfaces. (Clarifying : it's not asking to change in those way. I'm also on the fence between those, would like to hear opinion around)

saneyuki:
@kwonoj I'm sorry for the delay in my response. I'll check again tonight (about 8 hours later).

kwonoj:
No hurries at all, just checking :)

saneyuki:
> in case of windowSubscriber it could be work like this if we want to strict manner of interface,

make sense!

I think we need to keep the operation in `WindowSubscriber._error`. `Subscriber` or other derived classes' `error()` might call it.

saneyuki:
I pushed the fixup https://github.com/ReactiveX/RxJS/commit/40d5ff9c232f3a7c2add861f604b8a70e1eb32a0 :)

kwonoj:
Thanks for update, change looks good. Would you mind squash two commits 1119220 and 40d5ff9 into single commit, to avoid 40d5ff9 remain as 'fixup` commits in history.

saneyuki:
@kwonoj squashed!

kwonoj:
Cool, thanks! Let me check in once travis confirms builds.

kwonoj:
Merged with 41d39e2, thanks @saneyuki

PR #1117, opened 2015-12-30
refactor(operators): Make merge related operators' `_innerSub` private

saneyuki:
This methods is only used in them. We can hide their methods as private.


kwonoj:
looks good to me.

blesh:
LGTM

kwonoj:
Merged with 9bd7649, thanks @saneyuki .

PR #1118, opened 2015-12-30
refactor(Subject): Make Subject and derived classes' internal methods protected

saneyuki:
This change make these members `protected`:

- `Subject._next()`
- `Subject._error()`
- `Subject._complete()`

At current codebase, these members is only used in drived class of
`Subject` or itself. Thus we should hide them from public.


kwonoj:
I think change makes sense and looks good to me. I'll check this in around late today if there isn't any other suggestions or opinions.

kwonoj:
Merged with ce3de7a, thanks @saneyuki

PR #1123, opened 2016-01-03
fix(throw): Fix signature of error object mistakenly being type T.

trotyl:


src/observable/throw.ts
```diff
@@ -4,7 +4,7 @@ import {Subscription} from '../Subscription';
 
 export class ErrorObservable<T> extends Observable<T> {
 
-  static create<T>(error: T, scheduler?: Scheduler) {
+  static create<T>(error: any, scheduler?: Scheduler) {
```
kwonoj: think function won't need generic type also such as

`create(error: any, scheduler?: Scheduler)`

, what do you think?

trotyl: For functionality it's not needed, but for type inference by TypeScript it may sometimes be needed.
For example, you need to assign it to an object of type `Observable<Item>` and there is not chance for TypeScript to infer that type by the error of type `any`, and `ErrorObservable<any>` is not assignable to `Observable<T>`, then you need to provider it by `<Item>` mannualy.

kwonoj: Would you able to elaborate some snippets represents those cases?
In my understanding, current `<T>` in function signature won't be used all time for example

`Observable.throw('xxx');` 

since error object is explicitly specified to `any`, `T` won't be specified. It's not type of `Observable` class itself in `ErrorObservable<T>`, since function itself redefines generic for function only.

Simplifying current PR will looks as

``` 
class A<T> {
 doSomething<R>(value: any);
}
```

where `<R>` is not being used anywhere. I could possibly miss something, it'd be appreciate if you could bring up snippets those type's being consumed.

trotyl: I am currently trying on some code, it seem static method can not use Type information in static methods.
Namely, 
```typescript
class ErrorObservable<T> extends Observable<T> {
  static create(value: any): ErrorObservable<T> {
    return new ErrorObservable<T>(); // Compile time error for name T.
  }
}
``` 
would be illegal, and must be
```typescript
class ErrorObservable<T> extends Observable<T> {
  static create<T>(value: any): ErrorObservable<T> {
    return new ErrorObservable<T>();
  }
}
```
And now I have
```typescript
class Usage {
  getItem(id: string): Observable<Item> {
    // if found
    return Observable.of(item);
    // if not found
    return Observable.throw(new Error('...')); // It's illegal, can not cast ErrorObservable to Observable<Item>.
    // then
    return Observable.throw<Item>(new Error('...')); // it would be fine.
  }
}
```

kwonoj: Got point. It can be bit confused by type signature looks as

`Observable<T>.throw<{}>()`

class type information is not being used as you've mentioned, but usecase wise it needs to remain as-is.

trotyl: Actually, I'm have tried the way to put generic type in class, it seems
```typescript
Observable<Item>.method();
```
is illegal anyway, as I'm from C# it somehow confused me.

kwonoj: Yes, that'd be more ideal but seems not work in that way with current Typescript syntax.

kwonoj:
Change looks good to me. I'll check in this around tomorrow if any other suggestion / opinions around.

kwonoj:
@trotyl , would you able to rebase this PR to check it?

trotyl:
I'm not quite sure how to rebase a PR...

kwonoj:
@trotyl this is way I usually do,

```
git pull --rebase upstream master -> first, rebase master to latest
git checkout feature_branch
git rebase master -> rebase branch to master
```

This is me while I'm doing it :sweat:

blesh:
> I'm not quite sure how to rebase a PR...

@trotyl 

be sure you have an `upstream`:

```
git remote add upstream git@github.com:ReactiveX/RxJS.git
```

Then rebase

```
git checkout your_branch
git fetch upstream
git rebase upstream/master
```

Then force it up onto your fork:

```
git push origin your_branch -f
```

trotyl:
It's now rebased..if correct..

kwonoj:
Merged with 8f1fc93, thanks for effort @trotyl :)

Issue #1063, opened 2015-12-15
Finally and share do not work as expected


pavadeli:
The `finally` operator has some problems when used in combination with `share`. Some combinations work as expected, some don't.

###### Interval with `finally` works as expected:
```javascript
Rx.Observable.interval(100)
    .finally(() => console.log('finally 1'))
    .subscribe()
    .unsubscribe();
```
Console:
```
finally 1
```

###### Interval with `finally` and `share` does not work (nothing in log):
```javascript
Rx.Observable.interval(100)
    .finally(() => console.log('finally 1'))
    .share()
    .subscribe()
    .unsubscribe();
```
Console:
```
```

###### Subject with `finally` and `share` works as expected:
```javascript
new Rx.Subject()
    .finally(() => console.log('finally 1'))
    .share()
    .subscribe()
    .unsubscribe();
```
Console:
```
finally 1
```

###### Subject with two times `finally` and `share` does not work (nothing in log):
```javascript
new Rx.Subject()
    .finally(() => console.log('finally 1'))
    .finally(() => console.log('finally 2'))
    .share()
    .subscribe()
    .unsubscribe();
```
Console:
```
```

I tested this on *beta.0*. It does seem to work on a previous alpha (JSBin provides *alpha.8*, it works there). Unfortunately I don't understand enough of the inner workings to explain why this happens.

pavadeli:
The problem seems to be with `share` and not so much with `finally`. `Rx.Observable.merge` has the same kind of behaviour when combined with `share`.

###### `Rx.Observable.merge` with one child and `share` works as expected
```javascript
Rx.Observable
    .merge(
        Rx.Observable.interval(100)
    )
    .share()
    .subscribe(e => console.log(e));
```
Console:
```
1
2
3
4
...
```

###### `Rx.Observable.merge` with two children and `share` unsubscribe immediately
```javascript
Rx.Observable
    .merge(
        Rx.Observable.interval(100),
        Rx.Observable.interval(100)
    )
    .share()
    .subscribe(e => console.log(e));
```
Console:
```
```

staltz:
@blesh since I've been digging into share and ConnectableObservable, I volunteer to debug and fix this one, if it seems like a good priority in your opinion.

blesh:
[...]

PR #1126, opened 2016-01-04
Add tests against bug #1063.

staltz:
Bug #1063 is fixed by @trxcllnt's rework of Subscriptions/Subscribers. This PR just adds a few tests to make sure #1063 is not happening, so we can safely close that issue.


blesh:
Merged with 0a81bfc. Thanks, @staltz

kwonoj:
Closing this by changes are checked in.

Issue #1076, opened 2015-12-15
Issue with ReplaySubject as a subscriber when delay() is used (Beta.0)


TylorS:
When using a ReplaySubject as a subscriber to another Observable, if `.delay()` is used in the Observable sequence the ReplaySubject never recieves any values.

Working verison without `delay()` - http://jsbin.com/gezaqe/85/edit
Non-working version with `delay()` - http://jsbin.com/gezaqe/edit?html,js,console

staltz:
@TylorS is this a JSBin problem? I wasn't able to reproduce it. The "non-working" worked for me after I refreshed, cleared, and ran it.

I also don't see how is this related to ReplaySubject.
```js
var number$ = Rx.Observable.fromArray([1, 2, 3])
  .concatMap(x => Rx.Observable.of(x).delay(500))
var replicatedNumber$ = new Rx.ReplaySubject(1)


replicatedNumber$.subscribe(x => {
  console.log(replicatedNumber$.isUnsubscribed)
  if (x === 2) {
    replicatedNumber$.unsubscribe()
  }
})

number$.subscribe(replicatedNumber$)

// This subscription is what matters for console.log output
// Other subscriptions, such as the one above, should be
// impossible to affect this subscription. number$ is a cold
// Observable and should work when subscribed.
number$.subscribe(x => console.log('number$: ' + x))
```

TylorS:
I'm having a similar problem in a codebase as well. I could be wrong, but I felt my issue is with the ReplaySubject I'm trying to use to replicate the number$. I apologize, I wasn't very clear in my description of the issue.

I'm currently having no issues receiving values from number$ in either version, but replicatedNumber$ never logs it's values when `delay()` is present in the concatMap attached to `number$`. However, it will log `--false--false--|` when `delay()` is removed from `number$`

staltz:
Ok, now I can reproduce it. The minimal code to reproduce it is:

```js
const number$ = Rx.Observable.of(1, 2, 3).delay(500);
const subject = new Rx.Subject();
subject.subscribe(x => console.log(x));
number$.subscribe(subject);
```

TylorS:
I can confirm that your example causes the same issue. http://jsbin.com/gezaqe/91/edit?html,js,console

staltz:
I think I figured the root cause.
@blesh seems to be shared subscriber issue again.

[...]

PR #1127, opened 2016-01-04
test(Subject): add a test for Subject as an Observer

staltz:
Bug #1076 is fixed by @trxcllnt's rework of Subscriptions/Subscribers. This PR just adds a test to make sure #1076 is not happening, so we can safely close that issue.


kwonoj:
Change looks good to me, I'll check this in around tomorrow.

blesh:
Looks like this was already merged. Closing.

Issue #1085, opened 2015-12-16
groupBy not using lift 


blesh:
The current implementation of groupBy is no longer using `lift`. This actually muck with some other plans in the works for logging and debugging hooks.

GroupBy should be refactored to use `lift`.

[If you look at it](https://github.com/ReactiveX/RxJS/blob/6303558668016409315bfcd81b5b5f188a4cb174/src/operator/groupBy.ts#L25-L32) the `GroupByObservable` is really just a factory for a `Subscriber`. That's what we use `Operator` class for. I clearly didn't review this one very well when I merged it a long time ago. It seems that `RefCountSubscription` instance could be folded into the operator's subscriber as well.

trxcllnt:
@blesh I noticed this too. I'm working through a fix for `ConnectableObservable`, but don't have time to update `groupBy` just yet.

staltz:
I did that. At the time groupBy wasn't passing tests at all, so I just followed how RxJS 4 implemented it.

PR #1128, opened 2016-01-04
Fix groupBy() to use lift()

staltz:
Addresses bug #1085.

- Adds tests on Observable to verify that basic lift() features work.
- Adds tests on groupBy to assert it does not break lift() features.
- Fixes groupBy to use lift().

BREAKING CHANGES:
groupBy() now unsubscribes all inner Observables when the outer
Observable is unsubscribed. This diverges from the behavior in RxJS 4,
where inner Observables continue even if the outer is unsubscribed.

RxJS 4 explicitly supports the "inners continue when outer is unsubscribed", see [this test](https://github.com/Reactive-Extensions/RxJS/blob/01114cd050c90ad3b3a434c66a8f6aac50996b19/tests/observable/groupby.js#L829). @blesh @trxcllnt @mattpodwysocki what do you think about the breaking change I'm proposing? It seems odd to me why shouldn't we dispose of inners. I don't know if and why there is anyone depending on this behavior.


mattpodwysocki:
@staltz I'm not a fan of this breaking change as the inners should be independent of the outers which makes them more reliable as you pass the inners around for streaming purposes and to be weirdly affected by something upstream.  It's best to be kept as it was in RxJS v4 cc @headinthebox

blesh:
@mattpodwysocki I can see what you're talking about, but I'm having a hard time coming up with example code that would do this without doing something obviously strange.

blesh:
Also, why wouldn't we terminate those groups if they couldn't possibly produce another value because the source was terminated? It sounds more like a bug than a feature, TBH.

blesh:
Perhaps a middle-ground would be to have the inner streams error when the upstream is unsubscribed with a custom error like `UnsubscribedSourceError`. Then if the desired behavior is for the stream to continue on, you could handle that specific error with `catch` and return an `Observable.never()`.

blesh:
@staltz ... this needs to be rebased before I can merge it.

staltz:
Rebased! :)

mattpodwysocki:
@staltz @blesh I still have concerns about the independence of the inner versus outer subscriptions. I wouldn't merge until we come up to some agreement on this because it diverges from all other Rx implementations.

staltz:
@mattpodwysocki window also behaves like this (inners are unsubscribed when the outer is unsubscribed). Why is groupBy different? Who actually needs the use case you are portraying?

trxcllnt:
@staltz `GroupedObservable` is similar to `ConnectableObservable`, since they both front for a shared Subject. Multiple people can connect to the same `ConnectableObservable` without interfering with each other. The same should be true for `GroupedObservable`.

I know that's how `GroupedObservable` works in this PR, but this leads me to my next point:

once `groupBy` has emitted a `GroupedObservable`, that `GroupedObservable` has left the safety and comfort of the `groupBy` operator. Disposing the `GroupBySubscriber` shouldn't dispose inner `GroupedObservables`, because they might have subscription side-effects that `groupBy` doesn't know about, and if `groupBy` unsubscribes its inner `GroupedObservables`, unrelated side-effects might occur.

For example, if you multicast a `GroupedObservable` with a `ReplaySubject`, unsubscribing from the `GroupBySubscriber` would call `unsubscribe` on the `ReplaySubject`, which shouldn't be disposed yet.

mattpodwysocki:
@trxcllnt well said, and exactly my concerns that I had, which is why inner independence was necessary.

staltz:
Does that apply also to `window`?

RxJS 4 https://jsbin.com/pukeni/edit?js,console
RxJS 5 https://jsbin.com/sajuwu/edit?js,console

mattpodwysocki:
@staltz no, this not independent of the outer subscription, upon disposal, it still ensures that the window is eventually emptied and then it stops.

staltz:
@blesh @mattpodwysocki this PR is now safe to merge, I re-implemented the solution to keep it backwards-compatible as Matt wishes. We can/could postpone the discussion on subscription independence. The priority was to get `groupBy` working with `lift()`.

mattpodwysocki:
To be fair, it was both @trxcllnt and myself who wanted to keep independence.

staltz:
> Also, I don't see the tests associated with this to check for the independence of the inner from the outer such as these from

The tests were not added in this PR because they are _already_ there, I had ported them from RxJS 4 to this repo long ago.

blesh:
LGTM.

mattpodwysocki:
@staltz ah, sorry about that.  LGTM then!

src/Observable.ts
```diff
@@ -211,7 +211,7 @@ export class Observable<T> implements CoreOperators<T>  {
                projectResult?: (x: T, y: any, ix: number, iy: number) => R,
                concurrent?: number) => Observable<R>;
   flatMapTo: <R>(observable: Observable<any>, projectResult?: (x: T, y: any, ix: number, iy: number) => R, concurrent?: number) => Observable<R>;
-  groupBy: <K, R>(keySelector: (value: T) => string,
+  groupBy: <T, K, R>(keySelector: (value: T) => K,
```
kwonoj: do we need to add `T` as function generic parameter? expect `T` comes from generic type of `Observable<T>`.

src/operator/groupBy.ts
```diff
@@ -1,20 +1,27 @@
...
 export function groupBy<T, K, R>(keySelector: (value: T) => K,
                                  elementSelector?: (value: T) => R,
-                                 durationSelector?: (grouped: GroupedObservable<K, R>) => Observable<any>): GroupByObservable<T, K, R> {
-  return new GroupByObservable(this, keySelector, elementSelector, durationSelector);
+                                 durationSelector?: (grouped: GroupedObservable<K, R>) => Observable<any>): Observable<GroupedObservable<K, R>> {
+  return this.lift(new GroupByOperator(this, keySelector, elementSelector, durationSelector));
 }
 
-export class GroupByObservable<T, K, R> extends Observable<GroupedObservable<K, R>> {
+export interface RefCountSubscription {
```
kwonoj: is it needed to export this interface?

src/Observable.ts
staltz: I remember TS was complaining and this solved it, but I can re-check.

staltz:
@kwonoj I updated the TS typing issue that you pointed out.

kwonoj:
Thanks, change looks good to me too.

blesh:
:) Sorry @staltz, can you rebase this one?

staltz:
Alright, done

blesh:
Merged as of 39ebae9... I basically took your whole PR as-is during rebase and reapplied my perf fixes.

Thanks @staltz! This was an important one.

PR #1129, opened 2016-01-04
test(throttleTime): update testcase to use test scheduler

kwonoj:
- resolves intermittent async test failure on some environments


kwonoj:
Merged with cd41e6c.

PR #1132, opened 2016-01-05
Adds animationFrame Scheduler.

trxcllnt:


blesh:
@trxcllnt ... can you rebase this on the rxjs-dom branch? We'll merge it into that. I'm just parking this stuff there until we decide how to organize and publish it.

blesh:
Nevermind, I see that you already did that I think.

blesh:
Merged into rxjs-dom branch as of 15bfc2c

Thanks! This was an important add.

PR #1141, opened 2016-01-06
refactor(ajax): move helper methods

blesh:
- moves get, post, and getJSON to live on Observable.ajax (e.g. Observable.ajax.get)
- changes signatures of helper methods to be more refined for common use cases
- adds post and delete helpers

BREAKING CHANGE: get, post and getJSON helpers no longer live directly on Observable, they have been moved to Observable.ajx (e.g. Observable.ajax.get, Observable.ajax.post, etc)
BREAKING CHANGE: helpers for get and post no longer take Request (options) objects as parameters. If that feature is required, use the ajax method directly.


blesh:
Merging into `rxjs-dom` since @trxcllnt reviewed this offline.

PR #1143, opened 2016-01-06
Fixing issues in ajax code

blesh:
- Fixes issue where headers were not being set.
- Fixes issue were `timeout` was being set before `open`
- Removes `create2` method which was cruft from the last refactor.
- Adds tests around header setting.


blesh:
Whoa... this got jacked up somehow. Closing and fixing the PR.

PR #1152, opened 2016-01-08
chore(test): update dependency references in tests

kwonoj:
relates to #998

This PR updates references to `promise` in test case where it uses, ensure browser doesn't support promise can reference it.


kwonoj:
send coverage to coverall fails for some reason makes travis failure. checking if this is related with changes in PR.

kwonoj:
Seems coverall's confused sha of commit for some reason in this particular PR...

PR #1153, opened 2016-01-08
chore(test): update dependency references in tests

kwonoj:
relates to #998 

This PR updates references to promise in test case where it uses, ensure browser doesn't support promise can reference it.


kwonoj:
Merged with 05ac8dd. One step close to expanding browser test coverage, though still some problems left. Continuing effort to triage those.

PR #1154, opened 2016-01-08
fix(ajax): fix xhr response/responseText/responseXML handling.

trxcllnt:


blesh:
Closed in favor of #1169

PR #1157, opened 2016-01-11
chore(tests2png): add exhaust PNG diagram

staltz:
Add test to draw PNG diagram for exhaust. Change painter.js to support exhaust (use case where an
inner Observable is emitted but never subscribed to).


kwonoj:
Change looks good to me. I'll check this in around later today if there isn't any suggestions.

blesh:
Merged with 2cc2035... thanks @staltz

PR #1158, opened 2016-01-11
refactor(do): Make `DoSubscriber`'s menbers private.

saneyuki:
These members are not referred from others.

### DIFF
--- a/src/operator/do.ts
+++ b/src/operator/do.ts
@@ -47,9 +47,9 @@ class DoOperator<T> implements Operator<T, T> {
 
 class DoSubscriber<T> extends Subscriber<T> {
 
-  __next: (x: T) => void;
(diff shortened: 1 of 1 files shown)


kwonoj:
Just checking - is this changes are same changes in https://github.com/ReactiveX/RxJS/pull/1116 did not added at the moment?

saneyuki:
@kwonoj 

>  is this changes are same changes in #1116 did not added at the moment?

No, This is the separated change. This makes them private which stores callback functions passed from `DoOperator`.

kwonoj:
Change looks good to me. I'll check this in around later today if there isn't any suggestions.

blesh:
merged with a5927ba. Thanks @saneyuki

PR #1160, opened 2016-01-11
docs(MIGRATION): update operators list

luisgabriel:
`race` and `let` were added recently.


kwonoj:
Merged with 2dd771e, thanks @luisgabriel

Issue #1150, opened 2016-01-08
Observable.concat() does not complete with a buffered scalar observable


emonkak:
This Observable does not complete.

```javascript
const { Observable } = Rx

Observable
  .create(observer => {
    setTimeout(function() {
      observer.next(1)
      observer.complete()
    }, 1000)
	})
  .concat(Observable.of(2))
  .subscribe(
    x => console.log('Next:', x),
    x => console.log('Error:', x),
    () => console.log('Completed')
  )
```

http://jsfiddle.net/bo6ynqf7/7/

Expected:

```
Next: 1
Next: 2
Complete
```

Actual:

```
Next: 1
Next: 2
```

Nupf:
I just ran into a similiar problem and figured out the solution as far as I can tell. The problem only arises for scalar observables.

Concat is based on [mergeAll-support](https://github.com/ReactiveX/RxJS/blob/master/src/operator/mergeAll-support.ts), which does not subscribe to scalar-observables but [instantly nexts their value](https://github.com/ReactiveX/RxJS/blob/master/src/operator/mergeAll-support.ts#L26-L29). 
However, for [notifyComplete](https://github.com/ReactiveX/RxJS/blob/master/src/operator/mergeAll-support.ts#L46-L55) to be called, the observable has to be subscribed to by an [innerSubscriber](https://github.com/ReactiveX/RxJS/blob/master/src/InnerSubscriber.ts#L20-L23).

I believe someone more experienced should decide how this is best resolved (i.e. I would just change it to subscribe to scalar & non-scalar observables alike)

kwonoj:
Thinking this might be another side effect of ScalarObservable, related to https://github.com/ReactiveX/RxJS/issues/1142.

trxcllnt:
@Nupf yes, `subscribeToResult` already handles ScalarObservable, so any operators that check for `_isScalar` and aren't using `subscribeToResult` should be refactored to always call `subscribeToResult`.

blesh:
@kwonoj @trxcllnt ... do you know if this is still an issue?

kwonoj:
Just tried again with TOT and seems it's still occurring.

Issue #1140, opened 2016-01-06
Strange behaviour of Observable.of([x]).map


chrisprice:
I have a use case for creating an observable of arrays and came across some odd behaviour. The map function is invoked twice when using ```Observable.of(x)``` but not when using ```Observable.from([x])```. Is this expected?

```js
  it('should emit one value', function (done) {
    var calls = 0;
    Observable.of([42]).map(function (x) {
      expect(++calls).toBe(1);
      expect(x).toEqual([42]);
      return x;
    }).subscribe(function (x) {
      expect(++calls).toBe(2);
      expect(x).toEqual([42]);
    }, function (x) {
      done.fail('should not be called');
    }, done);
  });
```

```
Failures:
1) Observable.of should emit one value
  Message:
    Expected 2 to be 1.
  Stack:
    Error: Expected 2 to be 1.
        at ScalarObservable.<anonymous> (/Users/chris/RxJS/spec/observables/of-spec.js:51:23)
        at ScalarObservable.proto.map (/Users/chris/RxJS/dist/cjs/observable/ScalarObservable.js:61:45)
        at Object.<anonymous> (/Users/chris/RxJS/spec/observables/of-spec.js:50:25)
  Message:
    Expected 3 to be 2.
  Stack:
    Error: Expected 3 to be 2.
        at SafeSubscriber.<anonymous> (/Users/chris/RxJS/spec/observables/of-spec.js:55:23)
        at SafeSubscriber.tryCatcher [as _next] (/Users/chris/RxJS/dist/cjs/util/tryOrOnError.js:4:31)
        at SafeSubscriber.next (/Users/chris/RxJS/dist/cjs/Subscriber.js:89:18)
        at ScalarObservable._subscribe (/Users/chris/RxJS/dist/cjs/observable/ScalarObservable.js:44:24)

1999 specs, 1 failure
```

However, swapping ```Observable.of([42])``` for ```Observable.from([[42]])``` as below, doesn't trigger the failure.

```js
  it('should emit one value', function (done) {
    var calls = 0;
    Observable.from([[42]]).map(function (x) {
      expect(++calls).toBe(1);
      expect(x).toEqual([42]);
      return x;
    }).subscribe(function (x) {
      expect(++calls).toBe(2);
      expect(x).toEqual([42]);
    }, function (x) {
      done.fail('should not be called');
    }, done);
  });
```

chrisprice:
Should have mentioned this was tested with `master (d2e852)`.

[...]

PR #1161, opened 2016-01-11
fix(ScalarObservable): remove ScalarObservable optimization

### LINKED ISSUE #1142: Remove ScalarObservable optimizations for map, filter, reduce, scan, etc?
There is an issue #1140 that just popped up (I was waiting for it). 

I'm neither hot nor cold on these optimizations. So let's discuss them:

#### The Good:

1. The optimizations only effect observables created like `Observable.of(a)`, but not `Observable.of(1,2,3)`. They only effect ScalarObservables.
2. The optimizations doubled or tripled the speed of the operators for these Observables.
3. No one should be doing `Observable.of(a).map(toSomethingElse)` very often. It's code that only really shows up in demos and playground code normally, and/or it's an anti-pattern usually.
4. Generally it's a bad idea to mutate outer scoped variables inside of a map or filter handler.

#### The Bad:

1. Users trying to leverage map to create side effects might see funky behavior for ScalarObservables
2. The mapping or filtering functions are effectively no longer "lazy". If they're expensive or they create side effects, unexpected behavior can result.

kwonoj:
- remove usage of ScalarObservable for optimization
- implementation of ScalarObservable still remain for possible further usage

closes #1142, #1150, #1140 

Created PR, though it's bit early to get conclusion. Can be closed if it's decided to use ScalarObservable.

This PR removes usage of `ScalarObservable` from operators, but remain implementation itself for possible further discussions around optimizations when it's required.

### DIFF
--- a/doc/operator-creation.md
+++ b/doc/operator-creation.md
@@ -1,7 +1,7 @@
 # Operator Creation
 
-There are many ways to create an operator for RxJS. In this version of RxJS, performance was the primary consideration, as such, operator creation 
-in a way that adheres to the existing structures in this library may not be straight forward. This is an attempt to document how to 
+There are many ways to create an operator for RxJS. In this version of RxJS, performance was the primary consideration, as such, operator creation
+in a way that adheres to the existing structures in this library may not be straight forward. This is an attempt to document how to
 create an operator either for yourself, or for this library.
 
 For how to develop a custom operator for *this* library, [see below](#advanced).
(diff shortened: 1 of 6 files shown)


staltz:
I think the idea was to remove the _optimization_ from ScalarObservable, not to remove ScalarObservable itself (I might reading Ben wrong though). I can imagine there are performance benefits in keeping ScalarObservable. At least I think #1142 is only about removing the eager optimizations from ScalarObservable, not removing itself from FromObservable e.g.

kwonoj:
Ah, seems I was in hurry and did not get point correctly. Thanks for pointing out, @staltz . Let me close this, if this is really considerable can be reopened later.

PR #1164, opened 2016-01-11
fix(Immediate): update setImmediate compatibility on IE

### LINKED ISSUE #1163: setImmediate compatibility in IE10
among major browser engines, IE10 (and above, IE11) have implementation of `window.setImmediate` (https://developer.mozilla.org/ko/docs/Web/API/Window/setImmediate) and current implementation of RxJS `immediate` tries to utilize it. Currently it has 2 issues

- current codebase cannot call `window.setImmediate`
: our codebase assigns it as 
```
if (root.setImmediate) {
      this.setImmediate = root.setImmediate;
```
which brings error like below
```
IE 11.0.0 (Windows 10 0.0.0) AsapScheduler should schedule an action to happen later FAILED
        TypeError: Invalid calling object
```

reason of IE's behavior (https://msdn.microsoft.com/library/gg622930(v=vs.85).aspx).

- `window.setImmediate` has known issue
: has unreliable behavior reported at
http://codeforhire.com/2013/09/21/setimmediate-and-messagechannel-broken-on-internet-explorer-10/ 
and 
https://connect.microsoft.com/IE/feedback/details/801823/setimmediate-and-messagechannel-are-broken-in-ie10

confirmed as 'won't fix'. 

#### Possible options
- solve RxJS side issue only (calling setImmediate), let user decide to use `window.setImmediate` or not explicitly
- ignore `window.setImmediate` and use polyfill (onreadystatechange?)

kwonoj:
closes #1163

### DIFF
--- a/spec/util/Immediate-spec.js
+++ b/spec/util/Immediate-spec.js
@@ -10,13 +10,25 @@ describe('ImmediateDefinition', function () {
 
   describe('when setImmediate exists on root', function () {
     it('should use the setImmediate and clearImmediate methods from root', function () {
+      var setImmediateCalled = false;
+      var clearImmediateCalled = false;
+
       var root = {
-        setImmediate: function () {},
-        clearImmediate: function () {}
+        setImmediate: function () {
+          setImmediateCalled = true;
+        },
+        clearImmediate: function () {
+          clearImmediateCalled = true;
+        }
       };
+
       var result = new ImmediateDefinition(root);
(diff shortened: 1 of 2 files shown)


mattpodwysocki:
@kwonoj LGTM

kwonoj:
Appreciate for review! :)

kwonoj:
Merged with 39e6c0e.

PR #1165, opened 2016-01-12
fix(fromObservable): expand compatibility for iterating string source

### LINKED ISSUE #1147: Observable.from('string') browser compatibility
`Observable.from` can take string and iterate it as 

```
Rx.Observable.from('foo').subscribe(console.log);
// => Next: f
// => Next: o
// => Next: o
// => Completed
```

but fails to do it on some browsers does not support string iterator (IE, for example) 

in RxJS4, seems it has own iterator for string object (https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/linq/observable/from.js#L3-L23) to support this behavior - does RxJS5 need to port same behavior?

I've used  https://jsfiddle.net/v0w220v6/ to test out between different browsers.

kwonoj:
closes #1147

This PR enables utilization of `StringIterator` implementation in current codebase. While `IteratorObservable` already have those mechanism in place, it was not able to be executed in `fromObservable` by not allowing create `IteratorObservable` via string type source.

With this PR, string based source will create IteratorObservable either string has built in iterator or not.

### DIFF
--- a/src/observable/from.ts
+++ b/src/observable/from.ts
@@ -26,7 +26,7 @@ export class FromObservable<T> extends Observable<T> {
         return new ArrayObservable(ish, scheduler);
       } else if (isPromise(ish)) {
         return new PromiseObservable(ish, scheduler);
-      } else if (typeof ish[SymbolShim.iterator] === 'function') {
+      } else if (typeof ish[SymbolShim.iterator] === 'function' || typeof ish === 'string') {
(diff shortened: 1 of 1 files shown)


blesh:
Nice and easy fix! :)

merged with 8f7924f, thanks @kwonoj

kwonoj:
:)

PR #1170, opened 2016-01-13
fix(Subscriber): errors in nextHandler no longer propagate to errorHa…

### LINKED ISSUE #1135: Remove tryOrOnError from subscription
Somehow we missed this, but in order to comply with the es-observable spec, and in order not to confuse RxJS 4 users more than we already have, errors thrown in the next handler shouldn't jump to the error handler:

```js
Observable.of('hi').subscribe(() => {
  throw 'haha!';
}, (err) => {
  // THIS SHOULD NEVER BE HIT.
});
```

It should just throw.

blesh:
…ndler

fixes #1135


staltz:
Would be good to have a test for this. I can PR it.

PR #1172, opened 2016-01-13
fix(ajax): fixes error in Chrome accessing responseText when responseType isn't text.

trxcllnt:


blesh:
Oops. :( good catch

blesh:
Merged as of f3e2f73. Thanks @trxcllnt

PR #1174, opened 2016-01-13
fix(forkJoin): fix forkJoin to complete if sources Array is empty.

trxcllnt:


kwonoj:
Change looks good to me, will check in shortly.

kwonoj:
Merged with 412b13b, thanks @trxcllnt

Andrey-Pavlov:
Should it works after this fix `Observable.forkJoin([]).subscribe(res => { console.log(res); });` ?

kwonoj:
If it intends to not have any error yes but won't emit anything in console since there isn't anything to forkjoin.

mattpodwysocki:
@kwonoj yes, the intent is not to have any errors as we have the same in .NET with no values. See [`ForkJoin_None`](https://github.com/Reactive-Extensions/Rx.NET/blob/371c83c621562a2259580a03f0cb5bf8680ea720/Rx.NET/Source/Tests.System.Reactive/Tests/Linq/ObservableExTest.cs#L863-L872)

kwonoj:
@mattpodwysocki yes, that's my understanding.

and @Andrey-Pavlov updated comment 

> Update: it doesn't trigger next, only complete. -> Observable.forkJoin([]).subscribe(,,()=> { console.log('completed'); });

: This is expected, designed behavior of `forkJoin`. Regardless of this fix, forkJoin won't emit to `next` if there isn't thing to join.

PR #1175, opened 2016-01-13
refactor(race): Make Type paramerter more compatible with RxJS v4

saneyuki:
- Clearfy `race()`'s type definition.
  - See also [RxJS v4's one](https://github.com/Reactive-Extensions/RxJS/blob/master/ts/core/linq/observable/amb.ts)
- ...And some micro refactoring

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -242,7 +242,7 @@ export class Observable<T> implements CoreOperators<T>  {
   publishBehavior: (value: any) => ConnectableObservable<T>;
   publishReplay: (bufferSize?: number, windowTime?: number, scheduler?: Scheduler) => ConnectableObservable<T>;
   publishLast: () => ConnectableObservable<T>;
-  race: (...observables: Array<Observable<any>>) => Observable<any>;
+  race: (...observables: Array<Observable<T>>) => Observable<T>;
   reduce: <R>(project: (acc: R, x: T) => R, seed?: R) => Observable<R>;
   repeat: (count?: number) => Observable<T>;
   retry: (count?: number) => Observable<T>;
--- a/src/operator/race-static.ts
+++ b/src/operator/race-static.ts
@@ -8,7 +8,7 @@ import {isArray} from '../util/isArray';
  * @param {...Observables} ...observables sources used to race for which Observable emits first.
  * @returns {Observable} an Observable that mirrors the output of the first Observable to emit an item.
  */
-export function race<T>(...observables: Array<Observable<any> | Array<Observable<any>>>): Observable<T> {
+export function race<T>(...observables: Array<Observable<T> | Array<Observable<T>>>): Observable<T> {
   // if the only argument is an array, it was most likely called with
   // `pair([obs1, obs2, ...])`
   if (observables.length === 1) {
--- a/src/operator/race-support.ts
+++ b/src/operator/race-support.ts
@@ -20,11 +20,11 @@ export class RaceSubscriber<T, R> extends OuterSubscriber<T, R> {
     super(destination);
   }
 
-  _next(observable: any): void {
+  protected _next(observable: any): void {
     this.observables.push(observable);
   }
 
-  _complete() {
+  protected _complete() {
     const observables = this.observables;
     const len = observables.length;
     if (len === 0) {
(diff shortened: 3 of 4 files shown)


src/Observable.ts
```diff
@@ -247,7 +247,7 @@ export class Observable<T> implements CoreOperators<T>  {
   publishBehavior: (value: any) => ConnectableObservable<T>;
   publishReplay: (bufferSize?: number, windowTime?: number, scheduler?: Scheduler) => ConnectableObservable<T>;
   publishLast: () => ConnectableObservable<T>;
-  race: (...observables: Array<Observable<any>>) => Observable<any>;
+  race: (...observables: Array<Observable<T>>) => Observable<T>;
```
kwonoj: Just came to thought (I know it's not related with this PR) - if signature to be perfect parity compatible to RxJS4's `amb`, it would be something like
`race: (...observables: Array<Observable<T> | Promise<T>>) => Observable<T>;`
but `race` doesn't accept promise in its implementation yet.

Is my understanding correct?

src/operator/race-static.ts
```diff
@@ -8,14 +8,14 @@ import {isArray} from '../util/isArray';
...
  */
-export function race<T>(...observables: Array<Observable<any> | Array<Observable<any>>>): Observable<T> {
+export function race<T>(...observables: Array<Observable<T> | Array<Observable<T>>>): Observable<T> {
   // if the only argument is an array, it was most likely called with
   // `pair([obs1, obs2, ...])`
   if (observables.length === 1) {
-    if (isArray(observables[0])) {
-      observables = <Array<Observable<any>>>observables[0];
-    } else {
+    if (!isArray(observables[0])) {
```
kwonoj: I'm generally in favor of pattern 
```
if () {
 return; //return early if condition met(or not met)
}
```

but in this flow is there strong differences compare previous one? Either case this operator returns only when `observable[0]` is array and only order seems been changed.

kwonoj:
Change looks good to me in general (comments left are not critical to proceed)

src/operator/race-static.ts
saneyuki: Ah, I thought this order (the following) would be more clearly because the previous one was 2->1->3 by the following items.

1. the case we don't need to create an `ArrayObservable` (return a result)
2. the case we need to override the `observables` argument.
3. create and return `ArrayObservable`.

If you don't favor in it, I'll revert this change.

src/Observable.ts
saneyuki: > Is my understanding correct? 

Yes. your understanding is correct. We still have not implemented to accept `Promise<T>` to `race()` yet.
Perhaps, my pull request (and commit message) misleads you easily. Sorry.

david-driscoll: Fundamentally it is implemented, it's just not obvious because [`subscribeToResult`](https://github.com/ReactiveX/RxJS/blob/master/src/util/subscribeToResult.ts) consumes Arrays, Iterables, Observabes and Promises.

src/operator/race-static.ts
kwonoj: Personally I don't have strong preferences for this, but would like to remain previous here. I do understand your intention and it makes sense, also it brings reversed conditional check flow (check `if x===true` to `if !x`). 

My take in this change is, let's PR to have specific scope and take this as separate topics if necessary.

src/Observable.ts
kwonoj: @david-driscoll 's correct. Further change might be required to full support of implementation perspective, shortly I think for now this signature's more correct to avoid any possible confusion from consumer point.

src/operator/race-static.ts
kwonoj: (Just checking) may I able to ask your take on this?

saneyuki: ah, sorry. I know you ask to me. I'll reply you in tonight.

kwonoj: No hurries, take time :)

saneyuki: > My take in this change is, let's PR to have specific scope and take this as separate topics if necessary.

okay. I revert this change once, and will create a new pull request

saneyuki:
@kwonoj updated!

kwonoj:
Thanks! let me check this in around late today to tomorrow to see if any other suggestions around.

kwonoj:
@saneyuki , would you able to amend commit message as previous PR (https://github.com/ReactiveX/RxJS/pull/1212#issuecomment-173322395) ? I'll check this in once updated.

saneyuki:
@kwonoj updated!

kwonoj:
yay! :wink:

kwonoj:
Merged with d4813a9, thanks @saneyuki .

PR #1176, opened 2016-01-13
refactor(fromEvent): clearfy the type definition of Observable.fromEvent()

saneyuki:
This makes the type definition more compatible with [RxJS v4](https://github.com/Reactive-Extensions/RxJS/blob/8f12f812d497acf639588e90f74d504a9fc801ec/ts/core/linq/observable/fromevent.ts)

### DIFF
--- a/src/observable/FromEventObservable.ts
+++ b/src/observable/FromEventObservable.ts
@@ -4,30 +4,59 @@ import {errorObject} from '../util/errorObject';
 import {Subscription} from '../Subscription';
 import {Subscriber} from '../Subscriber';
 
+export type NodeStyleEventEmmitter = {
+  addListener: (eventName: string, handler: Function) => void;
+  removeListener: (eventName: string, handler: Function) => void;
+};
+function isNodeStyleEventEmmitter(sourceObj: any): sourceObj is NodeStyleEventEmmitter {
+  return !!sourceObj && typeof sourceObj.addListener === 'function' && typeof sourceObj.removeListener === 'function';
+}
+
+export type JQueryStyleEventEmitter = {
+  on: (eventName: string, handler: Function) => void;
+  off: (eventName: string, handler: Function) => void;
+};
+function isJQueryStyleEventEmitter(sourceObj: any): sourceObj is JQueryStyleEventEmitter {
+  return !!sourceObj && typeof sourceObj.on === 'function' && typeof sourceObj.off === 'function';
+}
+
+function isNodeList(sourceObj: any): sourceObj is NodeList {
+  return !!sourceObj && sourceObj.toString() === '[object NodeList]';
+}
+
+function isHTMLCollection(sourceObj: any): sourceObj is HTMLCollection {
+  return !!sourceObj && sourceObj.toString() === '[object HTMLCollection]';
+}
+
+function isEventTarget(sourceObj: any): sourceObj is EventTarget {
+  return !!sourceObj && typeof sourceObj.addEventListener === 'function' && typeof sourceObj.removeEventListener === 'function';
+}
+
+export type EventTargetLike = EventTarget | NodeStyleEventEmmitter | JQueryStyleEventEmitter | NodeList | HTMLCollection;
+
 export class FromEventObservable<T, R> extends Observable<T> {
 
-  static create<T>(sourceObj: any, eventName: string, selector?: (...args: Array<any>) => T) {
+  static create<T>(sourceObj: EventTargetLike, eventName: string, selector?: (...args: Array<any>) => T): Observable<T> {
     return new FromEventObservable(sourceObj, eventName, selector);
   }
 
-  constructor(private sourceObj: any, private eventName: string, private selector?: (...args: Array<any>) => T) {
+  constructor(private sourceObj: EventTargetLike, private eventName: string, private selector?: (...args: Array<any>) => T) {
     super();
   }
 
-  private static setupSubscription<T>(sourceObj: any, eventName: string, handler: Function, subscriber: Subscriber<T>) {
+  private static setupSubscription<T>(sourceObj: EventTargetLike, eventName: string, handler: Function, subscriber: Subscriber<T>) {
     let unsubscribe: () => void;
-    let tag = sourceObj.toString();
-    if (tag === '[object NodeList]' || tag === '[object HTMLCollection]') {
+    if (isNodeList(sourceObj) || isHTMLCollection(sourceObj)) {
       for (let i = 0, len = sourceObj.length; i < len; i++) {
         FromEventObservable.setupSubscription(sourceObj[i], eventName, handler, subscriber);
       }
-    } else if (typeof sourceObj.addEventListener === 'function' && typeof sourceObj.removeEventListener === 'function') {
-      sourceObj.addEventListener(eventName, handler);
-      unsubscribe = () => sourceObj.removeEventListener(eventName, handler);
-    } else if (typeof sourceObj.on === 'function' && typeof sourceObj.off === 'function') {
+    } else if (isEventTarget(sourceObj)) {
+      sourceObj.addEventListener(eventName, <EventListener>handler);
+      unsubscribe = () => sourceObj.removeEventListener(eventName, <EventListener>handler);
+    } else if (isJQueryStyleEventEmitter(sourceObj)) {
       sourceObj.on(eventName, handler);
       unsubscribe = () => sourceObj.off(eventName, handler);
-    } else if (typeof sourceObj.addListener === 'function' && typeof sourceObj.removeListener === 'function') {
+    } else if (isNodeStyleEventEmmitter(sourceObj)) {
       sourceObj.addListener(eventName, handler);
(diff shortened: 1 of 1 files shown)


src/observable/fromEvent.ts
```diff
@@ -4,37 +4,63 @@ import {errorObject} from '../util/errorObject';
...
+}
+
+function isNodeListOrHTMLCollection(sourceObj: any): sourceObj is NodeList | HTMLCollection {
+  let tag = sourceObj.toString();
+  return tag === '[object NodeList]' || tag === '[object HTMLCollection]';
+}
+
+export type EventTargetLike = EventTarget | EventEmmitterLike1 | EventEmmitterLike2 | NodeList | HTMLCollection;
+
+function setupSubscription<T>(sourceObj: EventTargetLike, eventName: string, handler: Function, subscriber: Subscriber<T>) {
```
kwonoj: Is there specific reason this funtion's out of class?

saneyuki: I thought this should not be the part of it because this function does not depend on the instance of `FromEventObservable`.

Are there some reasons to define the functions which don't depend on instance `this` as a static method by performance or something else?

kwonoj: I also don't have historical detail why it's originally designed in those way, so curious about way you chose. Personally though function does not depends on instance of class, it still makes sense to have it organized under class. @blesh , would you mind explain detail?

saneyuki: > I also don't have historical detail why it's originally designed in those way, so curious about way you chose. 
> Personally though function does not depends on instance of class, it still makes sense to have it organized under class. 

As my personal feeling, I don't think that a class should not have private static method which don't refer other class' static members if the language has a namespace or modules system. In other words, I think a class should be slim and should only have a instance/static members which is related to itself as much as possible to clearfy code.

A traditional (~ES5) JavaScript does not have any their system, so a developer need to create "namespace" object or "namespace" closure, or use some module system like CommonJS. (It was doom....) But this project is TypeScript and uses ES6 module style. We can define a private function. So I separate its methods from the class.

src/util/isEventEmitter.ts
```diff
@@ -0,0 +1,15 @@
+export type EventEmmitterLike1 = {
```
kwonoj: does other codebases using this exported types? can these be encapsulated under `fromEvent`?

saneyuki: No. This is only used in `fromEvent`. We can encapsulate this.

...But I intended to add it as a generic utility. Is this a premature optimization?

kwonoj: I wouldn't say it's 'premature', but in general I'd like to avoid expose if there isn't any common usage. If it's certain this will be used in other places, then it should be exported though. Do you have expectations this exported type (`EventEmmitterLikeN`, etcs) being used elsewhere than `fromEvent`?

david-driscoll: They'll need to be exported otherwise you'll raise a compiler exception.  What we don't have to do is export the from `Observable` and/or `KitchenSink`, so they'll be hard to get at.

Also since they're exported types unless someone tries to explicitly import this type, into an environment where this is imported as an ambient type (anything with `node.d.ts` really), there shouldn't not be any problems.  And long as the type shape matches, the ambient type would fit inside the exported type.

We probably don't need a separate `isEventEmitter` file though do we? This could just live as a private method in `fromEvent.ts`

kwonoj: If it's necessary to avoid compiler exception, I'm ok. In general principal would like encapsulate as much as we can, such as suggestions like avoid a separate isEventEmitter file.

BTW, (I know it's picky..) - any good suggestions of naming can replace xxx1, xxx2? @david-driscoll ?

saneyuki: @kwonoj @david-driscoll 

okay. I'll encapsulate this.

src/observable/fromEvent.ts
```diff
@@ -1,40 +1,60 @@
 import {Observable} from '../Observable';
+import {isEventTarget} from '../util/isEventTarget';
+import {isHTMLCollection} from '../util/isHTMLCollection';
+import {isNodeList} from '../util/isNodeList';
 import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
 import {Subscription} from '../Subscription';
 import {Subscriber} from '../Subscriber';
 
+export type EventEmmitterLike1 = {
```
david-driscoll: to @kwonoj's point of naming, I don't have much better names tbh.

`addListener`/`removeListener` is generally associated with NodeJS and it's `EventEmitter`.
`on`/`off` are generally associated with JQuery.

So only thing I can think of is something like `EventEmitterLike` and `JQueryEventEmitterLike`.

kwonoj: I think that gives more clarity than ~1 naming. Then what about `NodeEventEmitter` / `JQueryEventEmitter`?

kwonoj: `on/off` is not JQuery specific, might need somewhat generic name though.

david-driscoll: I went with JQuery because that's where I personally first saw the pattern appear.  I know it is certainly not the only place, but I imagine it's in Rx today because of JQuery's rise to dominance years ago.

kwonoj: Yes, just thought name would be bit more generic to avoid possible confusions. Again, naming's one of most painful part ;)

luisgabriel: +1 for `Node` instead of `JQuery`.

kwonoj: Isn't `add/remove` for `Node` and `on/off` for `JQuery`(and others)?

saneyuki: > Isn't add/remove for Node and on/off for JQuery(and others)?

what is your meaning `Node`?  DOM's `Node`? or NodeJS's node?

kwonoj: > addListener/removeListener is generally associated with NodeJS 

: NodeJS from @david-driscoll 's comment. Just trying to find name gives bit more clarity - so any suggestions would be appreciated.

saneyuki: @kwonoj 

As the ínterim summary of this discussion, I think these naming might be a bit clearly: `NodeStyleEventEmitter` & `jQueryStyleEventEmitter`. How about this?

As my personal thinking, I favor the convention `BarStyle` or `FooLike` to open a library which follows a style of jQuery.Event or Node's EventEmmitter.

kwonoj: I like those names, would like another suggestion since I'm not great for naming in English in general ;) @blesh @staltz , would it be ok with above suggested names?

src/observable/fromEvent.ts
kwonoj: Got point. For now as same as https://github.com/ReactiveX/RxJS/pull/1175#discussion_r50033458 I would like to suggest limit this PR to specific to type changes. This is purely about change scope, I think it'd better to have separate discussion regarding general organization strategy and discuss type definition in here.

saneyuki:
I updated:

- Remove the change splitting `setupSubscription()
- Rename type names of EventEmitter kinds

--------

@kwonoj 

I derived the naming convention `EventTargetLike` from [TypeScript's `PromiseLike<T>`](https://github.com/Microsoft/TypeScript/blob/9b151b3090f7eb75d1f2121da5d5aed9af253e15/lib/lib.core.es6.d.ts#L1190) but I feel `EventTargetKind` may be a bit more better.
How about do you feel?

kwonoj:
Both of them seems fine to me. Wish there's kind of dictionary for these kind of naming ;) Though I'm bit leaning towards to latter (`EventTargetKind`), if there's existing reference especially from TypeScript, that may be better choice. 

**BUT** since I'm not good at English enough, consider it as personal opinion and gladly welcome any suggestions. :) Let's wait for some time if any opinion / suggestions around.

luisgabriel:
`Kind` in a type system context means [something else](https://en.wikipedia.org/wiki/Kind_(type_theory)). I think `Like` makes more sense here.

src/util/isEventTarget.ts
```diff
@@ -0,0 +1,3 @@
+export function isEventTarget(sourceObj: any): sourceObj is EventTarget {
```
kwonoj: are these functions(`isEventTarget`, `isHTMLCollection`, `isNodeList`) need to be exported with separate module? do we have usage other than `fromEvent`?

saneyuki: I seem we don't have usage other than `fromEvent`.

When I wrote this patch, I thought we should add them as a utility function which can be used from other modules for organizing a code. But now I feel it may be "premerture" refactoring. I'll merge their functions into `fromEvent`.

saneyuki:
updated

kwonoj:
@saneyuki, I'm seeing one of commit is marked as fixup (7f3b468), is it expected to be squashed?

saneyuki:
@kwonoj squashed!

kwonoj:
Merged with 6f53dbf, thanks @saneyuki .

PR #1177, opened 2016-01-13
refactor(Observable): Make the internal '_subscribe()' method protected.

saneyuki:
- There is no reason that this member method should be public.
- If a third party user need some customize the behavior of it, then the person should make a custom operator.


kwonoj:
Makes sense to me (+change looks good to me).

kwonoj:
Let me check this in around today afternoon.

kwonoj:
Merged with 7ca7ba0, thanks @saneyuki

PR #1178, opened 2016-01-13
chore(generate-operator): Fixed operator generator to not fail on the…

david-driscoll:
… new dom folder


kwonoj:
Ah cool, thanks for catching this :+1: Came to think might need to have kind of hook to validate this on travis later..

kwonoj:
BTW, can we align commit message with history? (I know it's not fully consistent, better to be though)

david-driscoll:
All I did was run the tool after fixing it, I can add a note about running it.  The extra files were just ones that hadn't been picked up since the last time it ran afaik.

kwonoj:
> I can add a note about running it. 

: I think it's ok for now. I was thinking further way to ensure this.

luisgabriel:
I think `IteratorObservable.ts`, `ScalarObservable.ts` and `SubscribeOnObservable.ts` should not be added. Can you remove them from this PR?

We have two other tasks related to the generator:
1. Update the script to ignore the observable subclasses that are not exported via static method;
2. Update the script to generate the new operators inside the `dom` directory.

david-driscoll:
@luisgabriel done! :)

luisgabriel:
Nice. ;)

LGTM.

kwonoj:
Thanks for finding out @luisgabriel , assume #1182 and #1183 's to address issue you've listed, is my understanding correct?

luisgabriel:
@kwonoj Yep. Forgot to mention here that I've created these issues. ;)

kwonoj:
Cool. Then I'll check in this PR, let's track remaining changes via issues.

kwonoj:
Merged with 52eea95, thanks @david-driscoll , also @luisgabriel effort to look into this :)

Issue #1156, opened 2016-01-11
.distinct() operator


rudijs:
Hi,

I'm new to RxJS and in one of the early chapters of my book the .distinct() operator is used in the examples.

I see it's not part of this library, I'm assuming that's for performance reasons.

What would be the recommended approach for emiting unique values only with RxJS?

Thank you.

staltz:
@rudijs you can always keep a buffer with previous values, in a `scan` operator. I suppose `distinct` was removed because it is misleading, usually people want distinctUntilChanged, and distinct has memory from the whole history. If people really want it:

```js
source.scan((acc, curr) => {
  if (acc.indexOf(curr) === -1) { 
    acc.push(curr);
    return curr;
  } else {
    return 'ignore this';
  }
}, []).filter(x => x !== 'ignore this');
```

bouzuya:
does it work?

saneyuki:
@staltz I feel your example would work only once because the operator callback of `scan()` does not return `acc`

staltz:
Right, I was mistaken, should have been:

```js
source.scan((acc, curr) => {
  if (acc.indexOf(curr) === -1) {
    acc.push(curr);
    acc.last = curr;
    return acc;
  } else {
    acc.last = 'ignore this';
    return acc;
  }
}, [])
.map(acc => acc.last)
.filter(last => last !== 'ignore this');
```
```

trxcllnt:
@staltz @rudijs we haven't implemented `distinct` yet, but I believe it should be added to the list of operators we still need to implement. Even though it can be implemented by composing other operators, so can pretty much everything else, but it's a common enough operator that it should exist on its own cc: @blesh

mattpodwysocki:
@trxcllnt @staltz I agree, that `distinct` should be an operator unto itself.  Also, due to its constraint size of the buffer, we might want to add a flushing feature else the inner hash is going to grow infinitely.

justinwoo:
I'd be up for writing this operator up. Would it make sense for there to be an optional "flushing observable" that gets sent in for triggering hash flushing for every event?

blesh:
Yes, this should be implemented, but probably included only in KitchenSink, since it can be accomplished easily with other common operators.

kwonoj:
This operator's landed via https://github.com/ReactiveX/RxJS/pull/1180, closing issue.

PR #1180, opened 2016-01-13
Distinct operator

justinwoo:
Per #1156, I implemented a distinct operator that behaves similar to RxJS4, but also includes a way for the internal hashset to be flushed.

If this flushing mechanism doesn't really make sense, I can just remove that part of the code.

Feedback please! cc @mattpodwysocki


src/operator/distinct.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (flushes) {
+      this.flushSubscription = flushes.subscribe(() => this.hashSet.flush());
+    }
+  }
+
+  private compare(x: T, y: T): boolean {
+    return x === y;
+  }
+
+  _next(value: T): void {
```
saneyuki: should be `protected`

src/operator/distinct.ts
```diff
@@ -0,0 +1,103 @@
...
+      this.destination.error(errorObject.e);
+      return;
+    }
+
+    if (result) {
+      this.destination.next(value);
+    }
+  }
+
+  _complete(): void {
```
saneyuki: should be `protected`

src/operator/distinct.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.hashSet = new HashSet(this.compare);
+
+    if (flushes) {
+      this.flushSubscription = flushes.subscribe(() => this.hashSet.flush());
+    }
+  }
+
+  private compare(x: T, y: T): boolean {
+    return x === y;
+  }
```
saneyuki: I think we need not define this `compare()` as the member of `DistinctSubscriber`.

justinwoo: It's done this way in `distinctUntilChanged` and some other operators with function arguments. I assume it's for making lookup faster with javascript engines.

staltz:
I just realized something: we might need `distinctByKey`. RxJS 4 has both variants in the same operator: https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/distinct.md

But RxJS 5 split them:
- `distinctUntilChanged`
- `distinctUntilChangedByKey`

So now we're missing `distinctByKey`.

mattpodwysocki:
@staltz correct, and I assume we'll have comparers for keys as well for both?

justinwoo:
Okay, implemented `distinctKey` and flattened my commits.

src/Rx.KitchenSink.ts
```diff
@@ -8,6 +8,8 @@ import {Scheduler as IScheduler} from './Scheduler';
 export interface KitchenSinkOperators<T> extends CoreOperators<T> {
   isEmpty?: () => Observable<boolean>;
   elementAt?: (index: number, defaultValue?: any) => Observable<T>;
+  distinct?: (compare?: (x: T, y: T) => boolean, flushes?: Observable<any>) => Observable<T>;
```
kwonoj: `distinct` seems added in `Rx` module but definition leaves in `KitchenSinkOperator` instead of `CoreOperator`. Which module do we want to have those operator?

src/operator/distinct.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+class DistinctSubscriber<T> extends Subscriber<T> {
+  private hashSet: HashSet<T>;
+  private flushSubscription: Subscription;
+
+  constructor(destination: Subscriber<T>, compare: (x: T, y: T) => boolean, flushes: Observable<any>) {
+    super(destination);
+    if (typeof compare === 'function') {
```
kwonoj: is this type checking needed? There isn't any specific error thrown after this check, so either passing `null/undefined` or non-function object will eventually end up throwing in `HashSet`.

kwonoj: parameter signature of `distinct` is `(compare?: (x: T, y: T) => boolean, flushes?: Observable<any>)`, what would happen if 

```js
distinct(null, flushObs);
```

is this allowed execution case?

justinwoo: I don't see why not, that'd be like just buying into the default compare operator I think.

src/Rx.KitchenSink.ts
justinwoo: This one should be in KitchenSink since it's not essential. Do I need to fix this?

justinwoo: Ah, right. Let me go get rid of that line.

src/operator/distinct.ts
kwonoj: thanks, wanted to check if it'll be falls back into default comparer. Just came to notice - what about add those `distinct(null, obs);` into test case?

justinwoo: Yeah, I finished it up yesterday but wasn't sure if anyone wanted it. Pushed it up now.

blesh:
@justinwoo, I've merged this as of d026c41. All of your commits are intact, however I had to do some perf work on it. It was a great basis to start on. Check the perf commit d026c41 to see what I changed.

Thanks so much for your hard work!

blesh:
FWIW: We should probably take the time to eliminate closures in the `distinctKey` implementation at some point. And also add some perf tests.

PR #1181, opened 2016-01-13
test(Subscriber): add test for no error propagation

staltz:
Add test to assert that error happening in the next() handler should not propagate to the error()
handler.

For issue #1135.

### DIFF
--- a/spec/Subscriber-spec.js
+++ b/spec/Subscriber-spec.js
@@ -7,4 +7,29 @@ describe('Subscriber', function () {
     var sub = new Subscriber();
     expect(sub[Rx.Symbol.rxSubscriber]()).toBe(sub);
   });
+
+  describe('when created through create()', function () {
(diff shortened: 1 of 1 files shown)


luisgabriel:
LGTM.

kwonoj:
Can we convert test case to synchronous? I've tried way like below snippet.

```js
var subscriber = Subscriber.create(
  function next(value) {
    if (value === 2) {
      throw 'error!';
    }
  },
  errorSpy,
  completeSpy
);

subscriber.next(1);

expect(function () {
  subscriber.next(2);
}).toThrow('error!');

expect(errorSpy).not.toHaveBeenCalled();
expect(completeSpy).not.toHaveBeenCalled();
```

staltz:
@kwonoj Done

kwonoj:
:+1: looks good to me.

kwonoj:
Merged with d6f9668, thanks @staltz

PR #1185, opened 2016-01-13
fix forEach: nextHandler errors will reject returned Promise.

blesh:
There are a few merge errors fixed in here as well. 

This introduces some closure in the forEach implementation, but I'm not super concerned about it for now.


luisgabriel:
LGTM.

blesh:
Thanks for the review @luisgabriel ... I messed up the commit message... but no biggy. Merging because those eslint errors are annoying.

PR #1189, opened 2016-01-13
test(typings): Added tests for typings.

david-driscoll:


david-driscoll:
This may break at the moment, I just split this off of #1086, I'll be making some adjustments to it later this evening.

kwonoj:
Thanks for splitting. Let's not push changes more for now, instead let's leverage this PR to discuss mechanics of type validation test.

Currently there are 3 (including this) way proposed,

- convert unit test to typescript based, add additional test cases in existing tests (https://github.com/ReactiveX/RxJS/pull/826)
- separate test running for type validation only with custom runner using compiler api (https://github.com/ReactiveX/RxJS/pull/667)
- separate test, running with compiler out of box (this)

kwonoj:
Mainly separation between suggestion is will we going to use existing test cases vs. creating separate set of tests for type validation only.

kwonoj:
As same as other proposal, I've marked it as blocked for now to get final conclusion.

kwonoj:
@david-driscoll , I have created PR https://github.com/ReactiveX/RxJS/pull/1364 to enable logistics of testing type definition using existing test environment. Once it's agreed, these test may live under existing specs.

kwonoj:
@david-driscoll , I've created initial proposal #1414 to migrate these test into current unit test environment.

blesh:
It seems like with the integration tests all in TypeScript now, this might be a redundant effort. What do you think @david-driscoll? @kwonoj?

kwonoj:
>  this might be a redundant effort.

: Sort of, but test case itself is still valid. Once #1414 is accepted and checked in this PR need to be updated to use those approach (or either create new PR based on those). I'd like to let @david-driscoll decide between updating this PR or close it.

kwonoj:
#1414 is now checked in, this PR can utilize it.

kwonoj:
@david-driscoll , per discussion at https://github.com/ReactiveX/rxjs/pull/1479 let me close this PR for now. Primarily we'd like to cover type definition sanity via consumption from unit test itself, unless it's explicitly required. If there's some cases need to be explicitly, please reopen anytime and let's continue discussion. 

Really appreciate for effort. Thanks again!

PR #1191, opened 2016-01-14
fix(subscriptions): unsubscribe correctly when a Subscriber throws during synchronous dispatch.

### LINKED ISSUE #1186: Errors thrown in Subscriber's next handler should call unsubscribe logic
It was _supposed_ to be this way in Rx 4.0.6, according to @mattpodwysocki ... but [when I tried it](http://jsbin.com/xuhopiravu/1/edit?html,js,output) it didn't work out that way.

trxcllnt:
Resolves #1186. Also adds more tests ensuring calls to `next`, `error`, and `complete` are called in the context of whatever Observer they were specified from (the `SafeSubscriber` in the case of callbacks, or the Object instance if given a duck-typed Observer).

### DIFF
--- a/spec/Observable-spec.js
+++ b/spec/Observable-spec.js
@@ -146,7 +146,7 @@ describe('Observable', function () {
       expect(unsubscribeCalled).toBe(true);
     });
 
-    it('should not run unsubscription logic when an error is thrown sending messages synchronously', function () {
+    it('should run unsubscription logic when an error is thrown sending messages synchronously', function () {
       var messageError = false;
       var messageErrorValue = false;
       var unsubscribeCalled = false;
@@ -167,7 +167,7 @@ describe('Observable', function () {
       }
 
       expect(sub).toBe(undefined);
-      expect(unsubscribeCalled).toBe(false);
+      expect(unsubscribeCalled).toBe(true);
       expect(messageError).toBe(true);
       expect(messageErrorValue).toBe('boo!');
     });
@@ -195,7 +195,7 @@ describe('Observable', function () {
 
       expect(sub).toBe(undefined);
       expect(subscriber.isUnsubscribed).toBe(true);
-      expect(unsubscribeCalled).toBe(false);
+      expect(unsubscribeCalled).toBe(true);
       expect(messageError).toBe(true);
       expect(messageErrorValue).toBe('boo!');
     });
@@ -268,7 +268,7 @@ describe('Observable', function () {
         }
 
         expect(sub).toBe(undefined);
-        expect(unsubscribeCalled).toBe(false);
+        expect(unsubscribeCalled).toBe(true);
         expect(messageError).toBe(true);
         expect(messageErrorValue).toBe('boo!');
       });
--- a/src/Subscriber.ts
+++ b/src/Subscriber.ts
@@ -9,8 +9,14 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
   static create<T>(next?: (x?: T) => void,
                    error?: (e?: any) => void,
                    complete?: () => void): Subscriber<T> {
-    return new Subscriber(next, error, complete);
+    const subscriber = new Subscriber(next, error, complete);
+    subscriber.syncErrorThrowable = false;
+    return subscriber;
   }
+
+  public syncErrorValue: any = null;
+  public syncErrorThrown: boolean = false;
+  public syncErrorThrowable: boolean = false;
 
   protected isStopped: boolean = false;
   protected destination: Observer<any>;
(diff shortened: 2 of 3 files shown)


src/Subscriber.ts
```diff
@@ -12,29 +12,29 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
   static create<T>(next?: (x?: T) => void,
                    error?: (e?: any) => void,
                    complete?: () => void): Subscriber<T> {
-    return new SafeSubscriber<T>(next, error, complete);
+    return new SafeSubscriber<T>(null, next, error, complete);
   }
 
+  public syncErrorValue: any = null;
+  public syncErrorThrown: boolean = false;
+  public syncErrorThrowable: boolean = true;
```
blesh: Do these need to be `public` or can they be `protected`?

src/Subscriber.ts
```diff
@@ -83,18 +83,33 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
...
+    this._next = isFunction(next) && tryOrHandleMessageError(next, context) || null;
+    this._error = tryOrHandleMessageError(isFunction(error) && error || throwError, context);
+    this._complete = isFunction(complete) && tryOrHandleMessageError(complete, context) || null;
   }
 
-  next(value?: T): void {
+  next(value: T): void {
     if (!this.isStopped && this._next) {
       this._next(value);
+      this._reportMessageError(false);
```
blesh: since `_reportMessageError` doesn't do anything unless `syncErrorThrown` is `true`, shouldn't we wrap this in a conditional?

```js
if (this.syncErrorThrown) {
  this._reportMessageError(false);
}
```

src/Subscriber.ts
```diff
@@ -111,7 +128,22 @@ class SafeSubscriber<T> extends Subscriber<T> {
     if (!this.isStopped) {
       if (this._complete) {
         this._complete();
+        this._reportMessageError(true);
```
blesh: same comment as line 112.

src/Subscriber.ts
```diff
@@ -102,8 +117,10 @@ class SafeSubscriber<T> extends Subscriber<T> {
     if (!this.isStopped) {
       if (this._error) {
         this._error(err);
+        this._reportMessageError(true);
```
blesh: same comment as line 112

src/Subscriber.ts
```diff
@@ -111,7 +128,22 @@ class SafeSubscriber<T> extends Subscriber<T> {
...
+      } else {
+        this.unsubscribe();
+      }
+    }
+  }
+
+  private _reportMessageError(isStopped: boolean): void {
+    const { parent, syncErrorThrown } = this;
+    if (parent) {
```
blesh: I think we could wrap this whole block in an `if (syncErrorThrown)` since it's innards don't do anything unless that is `true`.

src/util/tryOrHandleMessageError.ts
```diff
@@ -0,0 +1,18 @@
+export function tryOrHandleMessageError(target: Function, context?: any): (x?: any) => any {
+  function tryCatcher() {
+    const { target, context } = (<any> tryCatcher);
+    try {
+      target.apply(context || this, arguments);
```
blesh: We could probably move this `context || this` to line 16. One less conditional called during each nexted return.

src/Subscriber.ts
trxcllnt: @blesh they do, since they need to be accessible to Observable's [`subscribe`](https://github.com/trxcllnt/RxJS/blob/unsubscribe-on-sync-throws/src/Observable.ts#L116).

src/util/tryOrHandleMessageError.ts
trxcllnt: @blesh would if I could, but `this` at line 5 won't be the same thing as `this` at line 16.

spec/Observable-spec.js
```diff
@@ -166,6 +207,38 @@ describe('Observable', function () {
...
+        } catch (e) {
+          messageError = true;
+          messageErrorValue = e;
+        }
+
+        expect(sub).toBe(undefined);
+        expect(unsubscribeCalled).toBe(true);
+        expect(messageError).toBe(true);
+        expect(messageErrorValue).toBe('boo!');
+      });
```
blesh: Can we add an additional test to ensure that no more emissions occur once the next handler throws?

This shouldn't emit `3`, for example:

```js
let tornDown = false;

new Observable(observer => {
  observer.next(1);
  observer.next(2);
  observer.next(3);

  function () => {
    tornDown = true;
  }
})
.subscribe(x => {
  if (x === 2) {
    throw new Error('I hate twos');
  }
  console.log(x);
});
```

trxcllnt: @blesh sure thing

trxcllnt:
@blesh updated this PR

blesh:
Merged as of b1698fe... After some perf testing, there was no perceivable difference in performance.

Thanks for the solid work, @trxcllnt

Issue #1166, opened 2016-01-12
Observable.from should handle "array-like" objects like arguments


mattpodwysocki:
In order to be compliant with RxJS v4, Rx.Observable.from is missing the following functionality which brings it on par with [`Array.from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from).  The current `isArray` check is too literal and should only check if it has a `.length` property

```js
// Generate a sequence of numbers
Array.from({length: 5}, (v, k) => k);    
// [0, 1, 2, 3, 4]
```

And so does RxJS v4:
```js
// Generate a sequence of numbers
Rx.Observable.from({length: 5}, (v, k) => k).subscribe(x => console.log(x));    
// [0, 1, 2, 3, 4]
```

Filing this bug as I did not see it in the test suite as it is in [RxJS v4](https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/from.js#L46-L63)

blesh:
That's a horrible way to make a series of numbers. haha.  Wouldn't users be better off using `range` for this? Do we really want them using `from` like that?

The better reason is so we can handle "array-like" types like `arguments`... basically anything with this shape:

```
var arrayLike = {
  "length": "3",
  "0": "foo",
  "1": "bar",
  "2": "baz"
};
```

So you can do things like `Observable.from(arguments)`

A better catch is that I don't think we have that result selector implemented.

blesh:
### We need

1. To handle array-likes, in particular arguments
2. To provide an optional resultSelector

mattpodwysocki:
@blesh yes, that's not a great example, and no, range won't do it.  Instead, focus on array like objects such as any `NodeList` object to turn into an Observable

blesh:
I think this is closable as of now. The only hanging issue is defined by the discussion here #1255

PR #1195, opened 2016-01-15
feat(from): allow Observable.from to handle array-like objects

justinwoo:
per #1166, this brings the signature and functionality closer to `Observable.from`
that is in RxJS4.

Cover your eyes, folks, this ain't pretty. Would appreciate feedback on how this should be changed. (e.g. should the default signature just be the 4-arg one with the mapFn? should fromArray just do the mapFn stuff by default and merge definitions with this? etc.)

### DIFF
--- a/spec/observables/from-spec.js
+++ b/spec/observables/from-spec.js
@@ -13,6 +13,67 @@ describe('Observable.from', function () {
     }, null, done);
   }, 300);
 
+  it('should handle an ArrayLike', function (done) {
+    var arrayLike = {
+      length: 3,
+      0: 1,
+      1: 2,
+      2: 3
+    };
+    var expected = [1, 2, 3];
+    var i = 0;
+    Observable.from(arrayLike).subscribe(function (x) {
+      expect(x).toBe(expected[i++]);
+    }, null, done);
+  }, 300);
+
+  it('should handle an ArrayLike from arguments', function (done) {
+    function makeArrayLike() {
+      var expected = [1, 2, 3];
+      var i = 0;
+
+      Observable.from(arguments).subscribe(function (x) {
+        expect(x).toBe(expected[i++]);
+      }, null, done);
+    }
+
+    makeArrayLike(1, 2, 3);
+  }, 300);
+
+  it('should handle an ArrayLike with a mapFn', function (done) {
+    var arrayLike = {
+      length: 3,
+      0: 1,
+      1: 2,
+      2: 3
+    };
+    var expected = [1, 1, 1];
+    var i = 0;
+    var mapFn = function (v, k) {
+      return v - k;
+    };
+    Observable.from(arrayLike, mapFn).subscribe(function (x) {
+      expect(x).toBe(expected[i++]);
+    }, null, done);
+  }, 300);
+
+  it('should handle an ArrayLike with a thisArg', function (done) {
+    var arrayLike = {
+      length: 3,
+      0: 1,
+      1: 2,
+      2: 3
+    };
+    var expected = [123, 123, 123];
+    var i = 0;
+    var mapFn = function (x, y) {
+      return this.thing;
+    };
+    Observable.from(arrayLike, mapFn, {thing: 123}).subscribe(function (x) {
+      expect(x).toBe(expected[i++]);
+    }, null, done);
+  });
+
   it('should handle a promise', function (done) {
     var promise = Promise.resolve('pinky swear');
 
--- a/src/observable/ArrayLikeObservable.ts
+++ b/src/observable/ArrayLikeObservable.ts
@@ -1,0 +1,74 @@
+import {Scheduler} from '../Scheduler';
+import {Observable} from '../Observable';
+import {ScalarObservable} from './ScalarObservable';
+import {EmptyObservable} from './EmptyObservable';
+import {Subscriber} from '../Subscriber';
+import {Subscription} from '../Subscription';
+
+export class ArrayLikeObservable<T> extends Observable<T> {
+
+  private mapFn: (x: any, y: number) => T;
+
+  static create<T>(arrayLike: ArrayLike<T>, mapFn: (x: any, y: number) => T, thisArg: any, scheduler?: Scheduler): Observable<T> {
+    const length = arrayLike.length;
+    if (length === 0) {
+      return new EmptyObservable<T>();
+    } else if (length === 1 && !mapFn) {
+      return new ScalarObservable<T>(<any>arrayLike[0], scheduler);
+    } else {
+      return new ArrayLikeObservable(arrayLike, mapFn, thisArg, scheduler);
+    }
+  }
+
+  static dispatch(state: any) {
+    const { arrayLike, index, length, mapFn, subscriber } = state;
+
+    if (subscriber.isUnsubscribed) {
+      return;
+    }
+
+    if (index >= length) {
+      subscriber.complete();
+      return;
+    }
+
+    const result = mapFn ? mapFn(arrayLike[index], index) : arrayLike[index];
+    subscriber.next(result);
+
+    state.index = index + 1;
+
+    (<any> this).schedule(state);
+  }
+
+  // value used if Array has one value and _isScalar
+  private value: any;
+
+  constructor(private arrayLike: ArrayLike<T>, mapFn: (x: any, y: number) => T, thisArg: any, private scheduler?: Scheduler) {
+    super();
+    if (!mapFn && !scheduler && arrayLike.length === 1) {
+      this._isScalar = true;
+      this.value = arrayLike[0];
+    }
+    if (mapFn) {
+      this.mapFn = mapFn.bind(thisArg);
+    }
+  }
+
+  protected _subscribe(subscriber: Subscriber<T>): Subscription | Function | void {
+    let index = 0;
+    const { arrayLike, mapFn, scheduler } = this;
+    const length = arrayLike.length;
+
+    if (scheduler) {
+      return scheduler.schedule(ArrayLikeObservable.dispatch, 0, {
+        arrayLike, index, length, mapFn, subscriber
+      });
+    } else {
+      for (let i = 0; i < length && !subscriber.isUnsubscribed; i++) {
+        const result = mapFn ? mapFn(arrayLike[i], i) : arrayLike[i];
+        subscriber.next(result);
+      }
+      subscriber.complete();
+    }
+  }
+}
--- a/src/observable/FromObservable.ts
+++ b/src/observable/FromObservable.ts
@@ -1,8 +1,11 @@
 import {isArray} from '../util/isArray';
+import {isFunction} from '../util/isFunction';
 import {isPromise} from '../util/isPromise';
+import {isScheduler} from '../util/isScheduler';
 import {PromiseObservable} from './PromiseObservable';
 import {IteratorObservable} from'./IteratorObservable';
 import {ArrayObservable} from './ArrayObservable';
+import {ArrayLikeObservable} from './ArrayLikeObservable';
 
 import {Scheduler} from '../Scheduler';
 import {SymbolShim} from '../util/SymbolShim';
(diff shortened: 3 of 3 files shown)


src/util/isArrayLike.ts
```diff
@@ -0,0 +1 @@
+export const isArrayLike = ((x: any): boolean => x && typeof x.length === 'number');
```
david-driscoll: You can make this a type guard, so TS can better infer types inside the if block;

    export const isArrayLike = (<T>(x: any): x is ArrayLike<T> => x && typeof x.length === 'number');

src/observable/fromArrayLike.ts
```diff
@@ -0,0 +1,69 @@
+import {Scheduler} from '../Scheduler';
+import {Observable} from '../Observable';
+import {Subscriber} from '../Subscriber';
+import {Subscription} from '../Subscription';
+
+export class ArrayLikeObservable<T> extends Observable<T> {
+
+  private _mapFn: (x: number, y: any) => T;
+
+  static create<T>(arrayLike: ArrayLike<T>, mapFn: (x: number, y: any) => T, thisArg: any, scheduler?: Scheduler) {
```
kwonoj: do we need support `thisArg`? in RxJS5, https://github.com/ReactiveX/RxJS/issues/878 unless there are corresponding native (i.e, `Array.map` vs `Observable.map`) function signature, does not accept `thisArg` as parameter. Curious this is applicable case.

src/util/isArrayLike.ts
blesh: The extra parens around both of these examples confused me.

Also, protip: in Github if you start with ```js  it will format the code block as JavaScript, TypeScript works too, for the most part.

src/observable/fromArrayLike.ts
```diff
@@ -0,0 +1,69 @@
...
+      this._mapFn = mapFn.bind(thisArg);
+    }
+  }
+
+  _subscribe(subscriber: Subscriber<T>): Subscription | Function | void {
+    let index = 0;
+    const arrayLike = this.arrayLike;
+    const count = arrayLike.length;
+    const mapFn = this._mapFn;
+    const scheduler = this.scheduler;
```
kwonoj: I liked way to shorten this kind of assignment in our codebase like
`const { arrayLike, count, mapFn... } = this`
can those be applicable here too?

src/observable/fromArrayLike.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  static dispatch(state: any) {
+
+    const { arrayLike, index, count, mapFn, subscriber } = state;
+
+    if (index >= count) {
+      subscriber.complete();
+      return;
+
```
kwonoj: empty line? (sorry for picky ;) )

src/observable/fromArrayLike.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    let result = mapFn ? mapFn(index, arrayLike[index]) : arrayLike[index];
```
kwonoj: can this be `const` instead?

blesh:
Couldn't this have been as easy as adding:

```TypeScript
if (isArrayLike(ish)) {
  return new ArrayObservable(new Array(ish.length), mapFn);
}
```

justinwoo:
Right, could be just as simple as `return new ArrayObservable(Array.from(ish.length, mapFn, thisArg), scheduler);`, but that does front-load transforming the values of the array-like into the transformed Observable, right?

If there's nothing about the scheduler semantics that needs to be worked out, then that would be fine with me, but I was delaying the transformation of the values until they are called for either on subscription or scheduler dispatch.

src/observable/fromArrayLike.ts
justinwoo: the original does bind the thisArg though, but not sure if this is something people *really* care about. https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/linq/observable/from.js#L101

justinwoo:
Updated based on feedback, but would love to hear if I can just get rid of most of this code or just merge this into the normal ArrayObservable and change the `from` signature entirely.

src/observable/from.ts
```diff
@@ -15,19 +17,30 @@ export class FromObservable<T> extends Observable<T> {
     super(null);
   }
 
-  static create<T>(ish: any, scheduler: Scheduler = null): Observable<T> {
+  static create<T>(ish: any, mapFnOrScheduler: Scheduler | ((x: number, y: any) => T), thisArg?: any, lastScheduler?: Scheduler): Observable<T> {
+    let scheduler: Scheduler;
+    let mapFn: (x: number, y: any) => T;
+    if (typeof mapFnOrScheduler === 'function') {
```
luisgabriel: You can use [isFunction()](https://github.com/ReactiveX/RxJS/blob/master/src/util/isFunction.ts) here.

luisgabriel:
What about this:

```typescript
//import {map} from '../operator/map'
if (isArrayLike(ish)) {
  const source = new ArrayObservable(Array.from(ish.length), scheduler);
  return map.call(source, mapFn, thisArg);
}

**Comment by blesh** [comment]
`from` should have a result selector we can use. If it doesn't, we should add it.

**Comment by justinwoo** [comment]
Result selector for every kind of input so that we use the same method signature as rxjs4 observable.from?

**Comment by blesh** [comment]
Actually, I've thought it over. This is the proper implementation. Disregard my silly idea above, you don't want to allocate another array with `Array.from`.

File: spec/observables/from-spec.js [hunk 7]
```diff
@@ -13,6 +13,54 @@ describe('Observable.from', function () {
...
+    var expected = [123, 123, 123];
+    var i = 0;
+    var mapFn = function (x, y) {
+      return this.thing;
+    };
+    Observable.from(arrayLike, mapFn, {thing: 123}).subscribe(function (x) {
+      expect(x).toBe(expected[i++]);
+    }, null, done);
+  });
+
```
blesh: Can we add a test for `arguments`? I suspect this will be a primary use case for this, since it's the most common ArrayLike.

src/observable/fromArrayLike.ts
```diff
@@ -0,0 +1,66 @@
...
+import {Observable} from '../Observable';
+import {Subscriber} from '../Subscriber';
+import {Subscription} from '../Subscription';
+
+export class ArrayLikeObservable<T> extends Observable<T> {
+
+  private mapFn: (x: number, y: any) => T;
+
+  static create<T>(arrayLike: ArrayLike<T>, mapFn: (x: number, y: any) => T, thisArg: any, scheduler?: Scheduler) {
+    return new ArrayLikeObservable(arrayLike, mapFn, thisArg, scheduler);
```
blesh: If there's no `mapFn`, and `length` of the `arrayLike` is 1, you could return `ScalarObservable`, and if it's zero, `EmptyObservable`

src/observable/fromArrayLike.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    state.index = index + 1;
+
+    (<any> this).schedule(state);
+  }
+
+  // value used if Array has one value and _isScalar
+  value: any;
+
+  constructor(public arrayLike: ArrayLike<T>, mapFn: (x: number, y: any) => T, thisArg: any, public scheduler?: Scheduler) {
```
blesh: `mapFn`'s arguments are backwards, I think. It should be: `(item: any, index: number) => T`.

src/observable/fromArrayLike.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+    if (mapFn) {
+      this.mapFn = mapFn.bind(thisArg);
+    }
+  }
+
+  _subscribe(subscriber: Subscriber<T>): Subscription | Function | void {
+    let index = 0;
+    const { arrayLike, mapFn, scheduler } = this;
+    const count = arrayLike.length;
```
blesh: can we name this variable `length`, since that's what it came from? (Applies elsewhere).

src/observable/fromArrayLike.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (!scheduler && arrayLike.length === 1) {
+      this._isScalar = true;
+      this.value = arrayLike[0];
+    }
+    if (mapFn) {
+      this.mapFn = mapFn.bind(thisArg);
+    }
+  }
+
+  _subscribe(subscriber: Subscriber<T>): Subscription | Function | void {
```
saneyuki: this should be `protected`

src/observable/fromArrayLike.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    if (index >= count) {
+      subscriber.complete();
+      return;
+    }
+
+    const result = mapFn ? mapFn(index, arrayLike[index]) : arrayLike[index];
+    subscriber.next(result);
+
+    if (subscriber.isUnsubscribed) {
```
blesh: I think we want to check `isUnsubscribed` first thing in here. Because anything could have happened to unsubscribe prior to `dispatch` being called. It won't hurt much, as-is, because we don't allow `next` to do anything if it's already unsubscribed, but it would be more efficient to check at the top of this function.

spec/observables/from-spec.js
justinwoo: Added on L30 above.

justinwoo:
Updated and all based on feedback. Are we okay with the fromArray/fromArrayLike split?

src/util/isArrayLike.ts
```diff
@@ -0,0 +1 @@
+export const isArrayLike = (<T>(x: any): x is ArrayLike<T> => x && typeof x.length === 'number');
```
kwonoj: do we have common usage's of this function over codebase except `from`? if not, what about defer to export it once there's usage?

src/observable/fromArrayLike.ts
```diff
@@ -0,0 +1,74 @@
...
+    const result = mapFn ? mapFn(arrayLike[index], index) : arrayLike[index];
+    subscriber.next(result);
+
+    state.index = index + 1;
+
+    (<any> this).schedule(state);
+  }
+
+  // value used if Array has one value and _isScalar
+  value: any;
```
kwonoj: can this be private?

src/observable/fromArrayLike.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  constructor(public arrayLike: ArrayLike<T>, mapFn: (x: any, y: number) => T, thisArg: any, public scheduler?: Scheduler) {
```
kwonoj: same as 44, `arrayLike` and `scheduler` seems can be private.

src/observable/fromArrayLike.ts
```diff
@@ -0,0 +1,74 @@
...
+    const { arrayLike, mapFn, scheduler } = this;
+    const length = arrayLike.length;
+
+    if (scheduler) {
+      return scheduler.schedule(ArrayLikeObservable.dispatch, 0, {
+        arrayLike, index, length, mapFn, subscriber
+      });
+    } else {
+      for (let i = 0; i < length && !subscriber.isUnsubscribed; i++) {
+        let result = mapFn ? mapFn(arrayLike[i], i) : arrayLike[i];
```
kwonoj: result can be `const`, preferred if it's applicable.

src/observable/fromArrayLike.ts
justinwoo: They were public in fromArray, but yeah, I don't think they need to be public.

src/util/isArrayLike.ts
justinwoo: I'd rather not crowd `from` with barely related defs and end up having split history.

src/observable/fromArrayLike.ts
justinwoo: Public in fromArray, but probably can be private just fine.

src/util/isArrayLike.ts
kwonoj: if it contains multiple lines I'd definitely agree, in this case would like to follow suggestion in https://github.com/ReactiveX/RxJS/pull/1211#issuecomment-172620017 , similar case to this.

justinwoo: If you insist. Pushed it up now.

kwonoj: It wasn't trying to insist, apologizes for my tone makes you felt.

kwonoj:
Change looks good to me, let me leave PR opened bit more to see if any other suggestions around.

src/util/isArrayLike.ts
justinwoo: It's fine, I'm just biased towards separating it out.

kwonoj:
Merged with 7245005, thanks @justinwoo .

Issue #1113, opened 2015-12-29
Expose Observable.of as a standalone function


cburgdorf:
Currently one has to import `import 'rxjs/add/observable/fromArray';` in order to get `Observable.of` which is pretty confusing. Can we make it so that `import 'rxjs/add/observable/of';` works?

kwonoj:
My assumption's https://github.com/ReactiveX/RxJS/pull/922#issuecomment-162967611 was intention for this which was not landed yet - 

>if we just rename observable/fromArray to observable/of, I think I'm okay just leaving these things out for now

Will update once @blesh confirms my understanding's correct.

cburgdorf:
I see. Thanks for the quick response.

blesh:
I think we can reasonably add an `add/observable/of` file, actually.

cburgdorf:
I read that those are generated from [here](https://github.com/ReactiveX/RxJS/blob/master/tools/generate-operator-patches.ts#L204) but I don't get exactly what to do to make it generate an `add/observable/of` file. Or should I just create it manually?

blesh:
@cburgdorf you could create it manually, I believe, since there are no name conflicts. Just add a comment to the top of the file to indicate it was manually created (since the others have the same for auto created)

cburgdorf:
Ok, guess this should do it https://github.com/ReactiveX/RxJS/pull/1197
I have some problems with the build since I don't have a proper Java environment. It's currently untested.

luisgabriel:
`fromArray` is deprecated in RxJS 4. Couldn't we just rename `fromArray` to `of`?

PR #1197, opened 2016-01-15
chore(Observable.of): expose as standalone function

cburgdorf:
Adresses #1113

This is actually untested. I'm not sure if this is supposed to have a test but are happy to get any pointers :)

### DIFF
--- a/src/add/observable/of.ts
+++ b/src/add/observable/of.ts
@@ -1,0 +1,9 @@
+/**
+ * This file was created manually, hence not managed by the 'tools/generate-operator-patches.ts' script.
+ **/
+import {Observable} from '../../Observable';
+import {ArrayObservable} from '../../observable/fromArray';
+
+Observable.of = ArrayObservable.of;
+
+export var _void: void;


blesh:
@cburgdorf I'm not sure what tests you could add around this at this point in time. I guess maybe some day we can have some integration tests that try to use all of the module files and the like. But for now this PR is fine.

kwonoj:
Yes, there isn't any test cases regarding this type of changes but only manual verification would work. I can pull down and check this works (Assume there isn't no reason to not work)

kwonoj:
I thought of just re-importing existing one to serve as alias like

`import './fromArray';`

but that's just personal - change looks ok to me. I'll check in changes once after I check it's working.

cburgdorf:
Oh, I didn't think of that. Feel free to change the commit before merging.

kwonoj:
@cburgdorf , just pull down pr and actually this doesn't work, not because of change in this PR but it's not compiled by compiler. TypeScript compiler can compile any file referenced but this change's alias to existing operator and not to be included anywhere, so doesn't compiled too.

couple of way to make it to work
- reference this in `Rx`, `Rx.KitchenSink` - double import `of` occurs
- update package.json, add `of.ts` in compiler

I'd go second way. Would you able to update PR?

kwonoj:
Or I can add following commit for compiler script changes, feel freely let me know.

cburgdorf:
Sorry, I'm not exactly sure where to add it in the `package.json`. And I'm not familiar enough with the build to test it manually myself right now. Feel free to drop the PR and introduce the change yourself. I'm not trying to sneak myself into the contributors list this way ;)

kwonoj:
@cburgdorf I believe this change itself is quite valid, and your effort is appreciated. Then let me leverage this PR to add commit for compiler changes :)

cburgdorf:
Thanks, sure. Whatever works well for you ;)

kwonoj:
Merged with 1edb636, by adding one commit to allow compiler actually takes this changes.

Appreciate for taking time to create PR @cburgdorf :)

cburgdorf:
And now that I see your changes I even learned some more bits about the project :) Collaboration as it should be :+1:

blesh:
Awesome! Thanks again @cburgdorf!

kwonoj:
Interestingly I'm seeing travis not able to find newly added file ('of.ts') and fail to build. I'll look this into quickly and will create fix if something's not correct with my script.

PR #1198, opened 2016-01-15
chore(package.json): fix build failures on non-windows platform

kwonoj:
Stupidity of confusing to use uppercase in path, not caught while building in local machine...


kwonoj:
Merged immediately with bec366a to avoid further build error from travis as well as anyone uses non-windows.

@blesh , this was my fault.

PR #1199, opened 2016-01-16
chore(test): expand browser test coverage

### LINKED ISSUE #998: Run tests in all major browsers and engines

kwonoj:
- expand browser list to be tested
- update karma configuration for newly added rx.dom test cases

closes #998 

This PR updates karma configurations to able to execute tests correctly, also expand browser matrix to cover all major browsers. There are browsers failing some of test cases can be triaged separately per each.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -7,7 +7,7 @@ module.exports = function (config) {
     sl_chrome: {
       base: 'SauceLabs',
       browserName: 'chrome',
-      version: '47'
+      version: '46'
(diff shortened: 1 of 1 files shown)


kwonoj:
Merged with ebfa7d2.

PR #1200, opened 2016-01-16
Update year range to 2016

pra85:


blesh:
Merged as of 55a2f98... Thanks for catching this @pra85! Apologies it took so long to merge such a simple PR. :)

Issue #1009, opened 2015-12-09
Do we need project and thisArg for IteratorObservable?


staltz:
Currently, `Observable.from`, when given an iterator, will create `IteratorObservable`, passing null as values to `project` and `thisArg`.

**There is no other case where IteratorObservable is created**, which means `project` and `thisArg` are never used.

Can we take them away?

kwonoj:
What about cases of someone import `IteratorObservable` directly and want to apply projection? `thisArg` could be removed though.

staltz:
I was somewhat assuming that if it's not in `Rx.ts` or `Rx.KitchenSink.ts`, then people wouldn't want to import it directly.

blesh:
So we can do a few things:

1. Allow optional `project` and `thisArg` parameters to `Observable.from`. (I'm cool with the `project`, and I think the `thisArg` bit is gross)
2. Add a `fromIterator` static method that takes these arguments.
3. Remove `IteratorObservable`.

Thoughts?

staltz:
(2) sounds better.

blesh:
I'm going to resubmit this issue with a slightly clearer top comment so we can get some community concensus.  https://github.com/ReactiveX/RxJS/issues/1255

PR #1204, opened 2016-01-16
refactor(Observable): remove create() from internal sublasses

### LINKED ISSUE #1183: Fix auto-patching operators generator to handle observable subclasses
This issue was pointed in #1178.

The `tools/generate-operator-patches.ts` is generating patches to some observable subclasses such as `ScalarObservable` and `SubscribeOnObservable` that are not directly export via static operators.

luisgabriel:
Currently, there are some Observable subclasses that are not directly
exported by a static operator of Observable but still have a `create()`
factory method. This method is not being used anywhere inside the
codebase besides explicit spec tests. The other operators that create
these special Observables are instantiating it via the constructor.

By removing these methods, we address issue #1183 as the patch generator
script looks for observables with `create()` to generate a patch such
as: `Observable.from = FromObservable.create;`.

Closes #1183

### DIFF
--- a/spec/observables/IteratorObservable-spec.js
+++ b/spec/observables/IteratorObservable-spec.js
@@ -9,37 +9,31 @@ describe('IteratorObservable', function () {
     expect(source instanceof IteratorObservable).toBe(true);
   });
 
-  it('should create IteratorObservable via static create function', function () {
-    var s = new IteratorObservable([]);
-    var r = IteratorObservable.create([]);
-    expect(s).toEqual(r);
-  });
-
   it('should not accept null (or truthy-equivalent to null) iterator', function () {
     expect(function () {
-      IteratorObservable.create(null);
+      var source = new IteratorObservable(null);
     }).toThrowError('iterator cannot be null.');
     expect(function () {
-      IteratorObservable.create(void 0);
+      var source = new IteratorObservable(void 0);
     }).toThrowError('iterator cannot be null.');
   });
 
   it('should not accept boolean as iterator', function () {
     expect(function () {
-      IteratorObservable.create(false);
+      var source = new IteratorObservable(false);
     }).toThrowError('Object is not iterable');
   });
 
   it('should not accept non-function project', function () {
     expect(function () {
-      IteratorObservable.create([], 42);
+      var source = new IteratorObservable([], 42);
     }).toThrowError('When provided, `project` must be a function.');
   });
 
   it('should emit members of an array iterator', function (done) {
     var expected = [10, 20, 30, 40];
-    IteratorObservable.create([10, 20, 30, 40])
-      .subscribe(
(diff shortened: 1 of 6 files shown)


kwonoj:
Makes sense to me. (+Change looks good to me)

luisgabriel:
This change may conflict with #1009 if the decision there is to create a `fromIterator` operator.

kwonoj:
You're expecting `fromIterator = IteratorObservable.create`?

luisgabriel:
Yes, this is [one of the options](https://github.com/ReactiveX/RxJS/issues/1009#issuecomment-172089750) that was proposed.

kwonoj:
Yup, that was my understanding from issue, thanks for clarification. Would you be ok to block this PR until those decision's being made? Other suggestions?

luisgabriel:
I'm ok with marking this as blocked for now. We can wait until the end of this week to see what happens on #1009.

kwonoj:
Thanks for clarification. I'm marking it as blocked for now, let's see where #1009 goes.

blesh:
This change seems unnecessary, honestly. I realize they're internal classes for the _global_ exports, but they're all still usable via _modules_.

```TypeScript
import { ScalarObservable } from 'rxjs-es/observable/ScalarObservable';

let someArgs = ['hi', Scheduler.asap];
ScalarObservable.create.apply(this, someArgs);
````

While I appreciate the effort that went in here, I don't agree with the PR.

david-driscoll:
They're still available as standalone methods, but this lets us strongly type `Observable.create`, `Subject.create`, etc.   I'm totally open to alternatives, for instance I tried using a new method `factory` instead of create and that worked similarly.  Just you have to understand there are multiple verbs (which was less than ideal).

    import { create as createScalarObservable } from 'rxjs-es/observable/ScalarObservable';
    
    let someArgs = ['hi', Scheduler.asap];
    createScalarObservable.apply(this, someArgs);

The main goal is to have the ability, from TypeScript, to have a strongly typed `create` method. Without that creating `Observable`s/`Subject`s/etc goes from `Observable.create((subscriber) => { ... })` and devolves into `<Observable<myType>>Observable.create((subscriber: Subscriber<myType>) => { ... })`.

luisgabriel:
@blesh My main motivation for proposing this change is the fact that the generator script assumes that observable subclasses with `create()` are directly connected to some static operator. So I thought that was the case. But I can drop this PR and create a new one changing the generator script to remove this assumption and fix #1183 .

But I'm a bit confused after this snippet in your comment. It's not clear to me what is the actual public API of this library. If these "internal" classes can be used via modules, should we be concerned with API compatibility of these classes between releases, for example?

blesh:
I have a feeling maybe the generator script for the static methods may have been a bit of over-engineering.

PR #1206, opened 2016-01-17
chore(test): update karma, wallaby configuration

kwonoj:
This PR updates karma configuration to let browser test have more sufficient to not to be disconnected while test is running. Test disconnection usually occurs on VM based test (IOS, etcs).


kwonoj:
Merged with ff13ec2

PR #1210, opened 2016-01-18
refactor(Notification): use 'void' for Notification.observe/do/accept().

saneyuki:
These method only returns a `void` value. There is no reason to specify `any` type.

### DIFF
--- a/src/Notification.ts
+++ b/src/Notification.ts
@@ -11,11 +11,11 @@ export class Notification<T> {
   observe(observer: Observer<T>): void {
     switch (this.kind) {
       case 'N':
-        return observer.next(this.value);
+        observer.next(this.value);
       case 'E':
-        return observer.error(this.exception);
+        observer.error(this.exception);
       case 'C':
-        return observer.complete();
+        observer.complete();
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/Notification.ts
```diff
@@ -8,7 +8,7 @@ export class Notification<T> {
     this.hasValue = kind === 'N';
   }
 
-  observe(observer: Observer<T>): any {
+  observe(observer: Observer<T>): void {
     switch (this.kind) {
       case 'N':
         return observer.next(this.value);
```
kwonoj: actually I was thinking this also, is it necessarily have to return here while `Observer::next` (or error, complete) doesn't return anything? what about also refactor this at once?

saneyuki: okay. I removed these `return`.

mattpodwysocki:
@saneyuki incorrect, you can return values from `Notification` as noted in RxJS v4: https://github.com/Reactive-Extensions/RxJS/blob/master/src/modular/notification.js#L61-L67 so please revert

saneyuki:
@mattpodwysocki 

> you can return values from Notification as noted in RxJS v4: https://github.com/Reactive-Extensions/RxJS/blob/master/src/modular/notification.js#L61-L67 and its associated tests: https://github.com/Reactive-Extensions/RxJS/blob/master/src/modular/test/notification.js#L73-L91.

- Does your point related to `Notification.accept()` only, doesn't it?
- If so, I feel this method's type signature and [RxJS v5's `Observer`](https://github.com/ReactiveX/RxJS/blob/master/src/Observer.ts) are invalid... The type of `nextOrObserver` arguments of `Notifications.accept()` should be compatible with v4. So Should we need to change it?

PR #1212, opened 2016-01-18
refactor(ToArraySubscriber): Make ToArraySubscriber.array private

saneyuki:
This member should be private. This is just micro refactoring :)


kwonoj:
looks good to me. re triggering travis once before try to check in..

kwonoj:
Merged with d469667, thanks @saneyuki .

I'm amended commit message to use lowercase when check this in. For further PR, would be able to not capitalize first letter in commit message? (per https://github.com/ReactiveX/RxJS/blob/master/CONTRIBUTING.md#subject) RxJS repo does some of automatic jobs using commit messages, so it'd be appreciate if commit messages are aligned.

PR #1213, opened 2016-01-18
test(zipAll): update test case to utilize zipAll

kwonoj:
Noticed couple of test cases in `zipAll` was executing `mergeAll`, assume it needs to be updated.

@staltz - would you able to confirm if this PR's correct, especially if use of `mergeAll` was intended?


kwonoj:
Merged with 175efe3.

PR #1214, opened 2016-01-18
Typing fixes

masaeedu:
The `Observable.prototype.lift` and `Operator.prototype.call` have unbound generic type parameters at the function level. Current signatures are:

```
// src/Operator.ts
export class Operator<T, R> {
  call<T, R>(subscriber: Subscriber<R>): Subscriber<T> {
    return new Subscriber<T>(subscriber);
  }
}

// src/Observable.ts
...
lift<T, R>(operator: Operator<T, R>): Observable<R> {
  const observable = new Observable<R>();
  observable.source = this;
  observable.operator = operator;
  return observable;
}
...
```

The `T` type parameter in each of those functions shouldn't be there, since `T` in each of those definitions needs to be bound to the `T` argument of the containing type.

The other changes are removing the redundant type argument in a few operators. The change also revealed that `MergeMapToOperator` doesn't implement the `Operator` interface correctly:

```
export class MergeMapToOperator<T, R, R2> implements Operator<Observable<T>, R2> {
  ...
  call(observer: Subscriber<R2>): Subscriber<T> {
    return new MergeMapToSubscriber(observer, this.ish, this.resultSelector, this.concurrent);
  }
}
```

An `Operator<Observable<T>, R2>` needs to implement a `call` method that returns `Subscriber<Observable<T>>`, not `Subscriber<T>`. Since I'm not familiar with what the `mergeMapTo` operator is supposed to do (the doc comment is missing), I couldn't change `MergeMapToSubscriber` to implement `Subscriber<Observable<T>>`. I've fixed it by replacing `T` with `any` for now, but this may bear further investigation.

### DIFF
--- a/src/Operator.ts
+++ b/src/Operator.ts
@@ -1,7 +1,7 @@
 import {Subscriber} from './Subscriber';
 
 export class Operator<T, R> {
-  call<T, R>(subscriber: Subscriber<R>): Subscriber<T> {
+  call<R>(subscriber: Subscriber<R>): Subscriber<T> {
     return new Subscriber<T>(subscriber);
   }
 }
--- a/src/Operator.ts
+++ b/src/Operator.ts
@@ -1,7 +1,7 @@
 import {Subscriber} from './Subscriber';
 
 export class Operator<T, R> {
-  call<R>(subscriber: Subscriber<R>): Subscriber<T> {
+  call(subscriber: Subscriber<R>): Subscriber<T> {
     return new Subscriber<T>(subscriber);
   }
 }
--- a/src/operator/every.ts
+++ b/src/operator/every.ts
@@ -39,7 +39,7 @@ export function every<T>(predicate: (value: T, index: number, source: Observable
   return source.lift(new EveryOperator(predicate, thisArg, source));
 }
 
-class EveryOperator<T, R> implements Operator<T, R> {
+class EveryOperator<T, R> implements Operator<T, boolean> {
   constructor(private predicate: (value: T, index: number, source: Observable<T>) => boolean,
               private thisArg?: any,
               private source?: Observable<T>) {
--- a/src/operator/isEmpty.ts
+++ b/src/operator/isEmpty.ts
@@ -13,15 +13,14 @@ export function isEmpty(): Observable<boolean> {
   return this.lift(new IsEmptyOperator());
 }
 
-class IsEmptyOperator<T> implements Operator<boolean, boolean> {
-  call (observer: Subscriber<T>): Subscriber<boolean> {
+class IsEmptyOperator implements Operator<any, boolean> {
+  call (observer: Subscriber<boolean>): Subscriber<any> {
     return new IsEmptySubscriber(observer);
   }
 }
 
-class IsEmptySubscriber extends Subscriber<boolean> {
-
-  constructor(destination: Subscriber<any>) {
+class IsEmptySubscriber extends Subscriber<any> {
+  constructor(destination: Subscriber<boolean>) {
     super(destination);
   }
 
--- a/src/operator/pairwise.ts
+++ b/src/operator/pairwise.ts
@@ -11,12 +11,12 @@ import {Subscriber} from '../Subscriber';
  *
  * @returns {Observable<R>} an observable of pairs of values.
  */
-export function pairwise<T>(): Observable<T> {
+export function pairwise<T>(): Observable<[T, T]> {
   return this.lift(new PairwiseOperator());
 }
 
-class PairwiseOperator<T, R> implements Operator<T, R> {
-  call(subscriber: Subscriber<T>): Subscriber<T> {
+class PairwiseOperator<T> implements Operator<T, [T, T]> {
+  call(subscriber: Subscriber<[T, T]>): Subscriber<T> {
     return new PairwiseSubscriber(subscriber);
   }
 }
@@ -25,7 +25,7 @@ class PairwiseSubscriber<T> extends Subscriber<T> {
   private prev: T;
   private hasPrev: boolean = false;
 
-  constructor(destination: Subscriber<T>) {
+  constructor(destination: Subscriber<[T, T]>) {
     super(destination);
   }
 
--- a/src/operator/scan.ts
+++ b/src/operator/scan.ts
@@ -21,7 +21,7 @@ class ScanOperator<T, R> implements Operator<T, R> {
   constructor(private accumulator: (acc: R, x: T) => R, private seed?: T | R) {
   }
 
-  call(subscriber: Subscriber<T>): Subscriber<T> {
+  call(subscriber: Subscriber<R>): Subscriber<T> {
     return new ScanSubscriber(subscriber, this.accumulator, this.seed);
   }
 }
@@ -40,7 +40,7 @@ class ScanSubscriber<T, R> extends Subscriber<T> {
 
   private accumulatorSet: boolean = false;
 
-  constructor(destination: Subscriber<T>, private accumulator: (acc: R, x: T) => R, seed?: T|R) {
+  constructor(destination: Subscriber<R>, private accumulator: (acc: R, x: T) => R, seed?: T|R) {
     super(destination);
     this.seed = seed;
     this.accumulator = accumulator;
--- a/src/operator/skipWhile.ts
+++ b/src/operator/skipWhile.ts
@@ -16,7 +16,7 @@ export function skipWhile<T>(predicate: (value: T, index: number) => boolean): O
   return this.lift(new SkipWhileOperator(predicate));
 }
 
-class SkipWhileOperator<T, R> implements Operator<T, R> {
+class SkipWhileOperator<T> implements Operator<T, T> {
   constructor(private predicate: (value: T, index: number) => boolean) {
   }
 
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -91,7 +91,7 @@ export class Observable<T> implements CoreOperators<T>  {
    * @description creates a new Observable, with this Observable as the source, and the passed
    * operator defined as the new observable's operator.
    */
-  lift<T, R>(operator: Operator<T, R>): Observable<R> {
+  lift<R>(operator: Operator<T, R>): Observable<R> {
     const observable = new Observable<R>();
     observable.source = this;
     observable.operator = operator;
--- a/src/observable/dom/WebSocketSubject.ts
+++ b/src/observable/dom/WebSocketSubject.ts
@@ -61,7 +61,7 @@ export class WebSocketSubject<T> extends Subject<T> {
 
   lift<R>(operator: Operator<T, R>) {
     const sock: WebSocketSubject<T> = new WebSocketSubject(this, this.destination);
-    sock.operator = operator;
+    sock.operator = <any>operator;
     return sock;
   }
 
--- a/src/operator/combineLatest.ts
+++ b/src/operator/combineLatest.ts
@@ -35,7 +35,7 @@ export function combineLatest<T, R>(...observables: Array<ObservableInput<any> |
 
   observables.unshift(this);
 
-  return new ArrayObservable(observables).lift<T, R>(new CombineLatestOperator<T, R>(project));
+  return new ArrayObservable(observables).lift(new CombineLatestOperator(project));
 }
 
 /* tslint:disable:max-line-length */
@@ -107,7 +107,7 @@ export function combineLatestStatic<T, R>(...observables: Array<any | Observable
     observables = <Array<Observable<any>>>observables[0];
   }
 
-  return new ArrayObservable(observables, scheduler).lift<T, R>(new CombineLatestOperator<T, R>(project));
+  return new ArrayObservable(observables, scheduler).lift(new CombineLatestOperator<T, R>(project));
 }
 
 export class CombineLatestOperator<T, R> implements Operator<T, R> {
--- a/src/operator/concat.ts
+++ b/src/operator/concat.ts
@@ -30,5 +30,5 @@ export function concatStatic<T, R>(...observables: Array<Observable<any> | Sched
     scheduler = args.pop();
   }
 
-  return new ArrayObservable(observables, scheduler).lift<Observable<T>, T | R>(new MergeAllOperator<T | R>(1));
+  return new ArrayObservable(observables, scheduler).lift(new MergeAllOperator<T | R>(1));
 }
--- a/src/operator/defaultIfEmpty.ts
+++ b/src/operator/defaultIfEmpty.ts
@@ -11,12 +11,12 @@ export function defaultIfEmpty<T, R>(defaultValue: R = null): Observable<T | R> 
   return this.lift(new DefaultIfEmptyOperator(defaultValue));
 }
 
-class DefaultIfEmptyOperator<T, R> implements Operator<T, R> {
+class DefaultIfEmptyOperator<T, R> implements Operator<T, T | R> {
 
   constructor(private defaultValue: R) {
   }
 
-  call(subscriber: Subscriber<T>): Subscriber<T> {
+  call(subscriber: Subscriber<T | R>): Subscriber<T> {
     return new DefaultIfEmptySubscriber(subscriber, this.defaultValue);
   }
 }
@@ -24,7 +24,7 @@ class DefaultIfEmptyOperator<T, R> implements Operator<T, R> {
 class DefaultIfEmptySubscriber<T, R> extends Subscriber<T> {
   private isEmpty: boolean = true;
 
-  constructor(destination: Subscriber<T>, private defaultValue: R) {
+  constructor(destination: Subscriber<T | R>, private defaultValue: R) {
     super(destination);
   }
 
--- a/src/operator/distinct.ts
+++ b/src/operator/distinct.ts
@@ -13,11 +13,11 @@ import {Subscription} from '../Subscription';
  * @param {Observable} [flushes] optional Observable for flushing the internal HashSet of the operator.
  * @returns {Observable} an Observable that emits items from the source Observable with distinct values.
  */
-export function distinct<T>(compare?: (x: T, y: T) => boolean, flushes?: Observable<any>) {
+export function distinct<T>(compare?: (x: T, y: T) => boolean, flushes?: Observable<any>): Observable<T> {
   return this.lift(new DistinctOperator(compare, flushes));
 }
 
-class DistinctOperator<T, R> implements Operator<T, R> {
+class DistinctOperator<T> implements Operator<T, T> {
   constructor(private compare: (x: T, y: T) => boolean, private flushes: Observable<any>) {
   }
 
--- a/src/operator/groupBy.ts
+++ b/src/operator/groupBy.ts
@@ -31,7 +31,7 @@ export interface RefCountSubscription {
   attemptedToUnsubscribe: boolean;
 }
 
-class GroupByOperator<T, K, R> extends Operator<T, R> {
+class GroupByOperator<T, K, R> extends Operator<T, GroupedObservable<K, R>> {
   constructor(public source: Observable<T>,
               private keySelector: (value: T) => K,
               private elementSelector?: (value: T) => R,
--- a/src/operator/inspect.ts
+++ b/src/operator/inspect.ts
@@ -12,7 +12,7 @@ export function inspect<T>(durationSelector: (value: T) => Observable<any> | Pro
   return this.lift(new InspectOperator(durationSelector));
 }
 
-class InspectOperator<T, R> implements Operator<T, R> {
+class InspectOperator<T> implements Operator<T, T> {
   constructor(private durationSelector: (value: T) => Observable<any> | Promise<any>) {
   }
 
--- a/src/operator/inspectTime.ts
+++ b/src/operator/inspectTime.ts
@@ -9,7 +9,7 @@ export function inspectTime<T>(delay: number, scheduler: Scheduler = asap): Obse
   return this.lift(new InspectTimeOperator(delay, scheduler));
 }
 
-class InspectTimeOperator<T, R> implements Operator<T, R> {
+class InspectTimeOperator<T> implements Operator<T, T> {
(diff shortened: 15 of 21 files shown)


src/Operator.ts
```diff
@@ -1,7 +1,7 @@
 import {Subscriber} from './Subscriber';
 
 export class Operator<T, R> {
-  call<T, R>(subscriber: Subscriber<R>): Subscriber<T> {
+  call<R>(subscriber: Subscriber<R>): Subscriber<T> {
```
luisgabriel: In this case, the type `R` of `call` does not have any relation with `R` of `Operator`?

masaeedu: Good catch; that's incorrect. `R` should be bound to the type's `R` as well. I think fixing this will greatly expand the scope of operators that need to be touched however. I'll try it out.

luisgabriel:
Just a minor details regarding the commit messages: you should not capitalize the first letter, and you should use present tense. More info on [CONTRIBUTING.md#subject](https://github.com/ReactiveX/RxJS/blob/master/CONTRIBUTING.md#subject)

masaeedu:
@luisgabriel Sorry, missed that bit in the contribution guidelines. Looks like fixing that second parameter has indeed broken a bunch of other operators that can only be fixed by introducing `any` as a quick hack or figuring out how the types flow through the implementation. 

I've actually already worked out the latter approach in a different branch (I'm planning on retracing the changes in small commits), but it involves lots of code being touched throughout the codebase. Would it be better for now to `any` any operators that incorrectly implement `Operator`?

luisgabriel:
I'm not sure what would be the best solution here. Could you point out this branch you said? I think I can have a better idea of the problem size.

It's also worth to wait for some inputs from @kwonoj  and @david-driscoll in this matter.

masaeedu:
Sure; take a look at the `typingFixesAllAtOnce` [branch](https://github.com/masaeedu/RxJS/tree/typingFixesAllAtOnce). Unfortunately there's a number of changes in there that build on this bugfix that make things a bit hairy. They include deduplication of the function signatures by adding an interface to each operator file and various changes in the implementation to make types flow correctly. Those changes aren't necessarily in a finished state. 

You could probably get a better idea of which operators would need to be fixed solely as part of this  bugfix by:

 1. Cloning the `typingFixes` [branch](https://github.com/masaeedu/RxJS/tree/typingFixes)
 1. Removing the `R` parameter from `Operator`
 1. Trying to build

kwonoj:
I was trying same (get rid of `<R>`) at the same moment, and agree it requires amount of changes to make correct type inferences.

My vote in here is make changes correctly at once, avoiding one additional steps of introducing `any`. Maybe PR can be separated per effort required, but for me removing redundant generic `<R>` and updating signature is atomic, should be one changes.

masaeedu:
I agree that would be the ideal approach @kwonoj, but the problem is that making the types flow correctly inside the implementations requires a lot of domain knowledge and isn't a simple chore type task. For this reason it would be easier if it could be fixed in separate PRs on a case by case basis.

In the `typingFixesAllAtOnce` branch I linked, I fixed implementations for whichever operators included doc comments, but there were still a few where I couldn't find or understand the expected behavior and had to use `any`.

kwonoj:
As I already mentioned, I agree that requires sufficient time to make corresponding changes. I'm not object to make separate PR for those effort, while this PR leaving `<R>` for now if those requires noticeable input for making changes. Still, I'm bit opposing to introduce interim `any` for incremental changes. PR https://github.com/ReactiveX/RxJS/pull/1077 would be reference for this kind of cases - even though it's amount of changes if it's logical atomic chunk, it'd be better to be changes at once.

masaeedu:
Ok, I see your point. :) So in terms of direction forward, here's a couple of problems that pop up which I don't know how to address:

 - `webSocket.ts`: Mismatch in type of `operator` property
 - `lift<T, R>` problem exists in `Subject` as well. Removing `T` results in similar type mismatch of `operator` property

There are also compile errors in the following files, however I already have tentative fixes for each of these in `typingFixesAllAtOnce` which need to be checked for correctness:

 - defaultIfEmpty.ts
 - every.ts
 - isEmpty.ts
 - pairwise.ts
 - race-support.ts
 - reduce-support.ts
 - scan.ts
 - skipwhile.ts
 - withLatestFrom.ts

I will clean these up and add them as individual commits to this branch. Once all of these are addressed I think we will be rid of the extra `<T, R>`.

masaeedu:
@kwonoj Ok, so I've cleaned up all the ones I know how to fix.The only remaining failures at this point are the `webSocket` and `Subject` failures. If you could take a look and indicate out how those can be fixed that would be much appreciated.

kwonoj:
Wow, this is huge ;) appreciate for effort. Let's start look into, I'll add some inputs if I have some questions around. /cc @david-driscoll also if he could give some helps.

src/operator/pairwise.ts
```diff
@@ -8,12 +8,13 @@ import {Subscriber} from '../Subscriber';
  * The (N-1)th is stored in the internal state until Nth input occurs.
  * @returns {Observable<R>} an observable of pairs of values.
  */
-export function pairwise<T>(): Observable<T> {
-  return this.lift(new PairwiseOperator());
+export function pairwise<T>(): Observable<[T, T]> {
+  let _this: Observable<T> = this;
```
luisgabriel: Why `_this`?

src/Subject.ts
```diff
@@ -29,7 +29,7 @@ export class Subject<T> extends Observable<T> implements Observer<T>, Subscripti
   protected dispatching: boolean = false;
   protected hasCompleted: boolean = false;
 
-  lift<T, R>(operator: Operator<T, R>): Observable<T> {
+  lift<R>(operator: Operator<T, R>): Observable<T> {
```
david-driscoll: `lift` should return `Observable<R>` should it not?

masaeedu: @david-driscoll That's right, forgot to checkout.

src/operator/scan.ts
```diff
@@ -13,14 +13,15 @@ import {errorObject} from '../util/errorObject';
  * @returns {Obervable} An observable of the accumulated values.
  */
 export function scan<T, R>(accumulator: (acc: R, x: T) => R, seed?: T | R): Observable<R> {
-  return this.lift(new ScanOperator(accumulator, seed));
+  let _this: Observable<T> = this;
+  return _this.lift(new ScanOperator(accumulator, seed));
```
david-driscoll: casting to `Observable<T>` may be less intrusive here.

src/operator/skipWhile.ts
```diff
@@ -5,10 +5,11 @@ import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
 
 export function skipWhile<T>(predicate: (value: T, index: number) => boolean): Observable<T> {
-  return this.lift(new SkipWhileOperator(predicate));
+  let _this: Observable<T> = this;
+  return _this.lift(new SkipWhileOperator(predicate));
```
david-driscoll: casting again

src/operator/pairwise.ts
kwonoj: +whereever possible, let's use `const` instead.

src/Subject.ts
masaeedu: @david-driscoll Ok, I've committed my partial fix for this. From my understanding the signature here should look like:

```
  lift<R>(operator: Operator<T, R>): Observable<R> {
    const subject = new Subject<R>(this, this.destination || this);
    subject.operator = operator;
    return subject;
  }
```

The problem now is that the first line of the function tries to produce a `Subject<R>` using `this`, which doesn't make sense. I'm not sure what the fix is.

david-driscoll:
The point about `Subject.lift` makes me thing that you should either just cast `this` to `<any>` and not worry about the underlying details, or `Subject` needs an additional generic type parameter.

I'm inclined to say cast it to `any`, because most consumers will be making subjects, and having two type arguments would be confusing.

src/Operator.ts
```diff
@@ -1,7 +1,11 @@
 import {Subscriber} from './Subscriber';
 
-export class Operator<T, R> {
-  call<T, R>(subscriber: Subscriber<R>): Subscriber<T> {
+export interface Operator<T, R> {
```
kwonoj: do we expect export interface and implementation both under same name? would you able to elaborate use case for interface is required with same name of implementation?

src/operator/scan.ts
masaeedu: That's true. The other side of that coin is that enforcing the `let _this: Observable<T>` pattern consistently will prevent people from making mistakes due to `this` being `any` down the line. I have no strong opinion either way but just wanted to point it out. /cc @kwonoj

src/operator/pairwise.ts
masaeedu: @luisgabriel The idea is that using `this` gives you `any`, and then all type checking from that point on is disabled. Using only `_this` ensures your type signatures are compatible (i.e. lifting through the operator actually produces the output type of the function). We can also do this using casts, but I arrived at this pattern over the course of a number of changes across other operators.

masaeedu:
@kwonoj Would you agree just casting to `<any>` is fine here?

kwonoj:
> Would you agree just casting to <any> is fine here? 

: It'd be appreciate if you can point 'here', since comment is not pointing specific code and due to size of change, it's bit hard to find right context. ;)

src/operator/pairwise.ts
luisgabriel: Got it. Thanks for the explanation.

I agree with you. This pattern leads to a better readability over casting. I'd prefer to name it `self` instead of `_this`, but this is a minor detail.

david-driscoll: I would go with `self` probably because the compiler will user `_this` internally for fat arrow methods.

masaeedu:
Sorry, I was referring to David's comment there. If you look at [my last commit](https://github.com/masaeedu/RxJS/commit/3eb136851cd1a40db18adbb0d20154a380466592), you'll see the problem. The `lift` method on `Subject` is supposed to produce (presumably), an `Observable<R>`. However, the first line in `lift` tries to make a `Subject<R>` using nothing but the instance of `Subject<T>` itself. Fixing in a type safe way would involve adding another type parameter to `Subject`, which would be somewhat inconvenient and would probably involve a lot of changes wherever it is referenced.

src/operator/pairwise.ts
kwonoj: for naming specific I also vote for self, in addition to @david-driscoll 's suggestion
- it's name already using in codebases
- generally in RxJS codebase avoid underbar

for casting itself, let me follow context bit more..

masaeedu: Got it, will change to `self`. However, perhaps I should rever to `this` here and add a separate PR to introduce `self` across operators (and fix any typing bugs this reveals)?

kwonoj: I'm bit on fence between casting vs using explicit typed reference. So it'll be basically
```js
const self: Observable<T> = this;
self.doxxx();
```

vs. 

```js
((Observable<T>)this).doXXX()...
```

correct? While I like prior snippet's clarity of type to `self`, also makes feel does it necessarily have to hold another reference which is repeated over every operator. (course I also does agree @masaeedu about casting, just speaking out loud to hear some opinions)

masaeedu: @kwonoj That's correct. David raised the performance aspect of this as well. Just so we can make this explicit, the concern is that the instance of `Observable` would be disposed by the GC if not for the `_this`/`self` reference, so adding this alias will result in memory pressure/a full out leak. Is this a fair assessment?

I will try to set up a test in Chrome memory mode to see how the alias affects frequency of GC and memory usage.

masaeedu: If I was a betting man, my bet would be that this has no significant impact, since the locally scoped `self` reference is orphaned as soon as the operator function returns (provided we don't start passing it into closures). So the lifetime of the object is dictated only by how long external references to the observable persist.

However, as mentioned earlier, I'll do a test in Chrome to backup my handwaving :)

src/Operator.ts
masaeedu: @kwonoj I think this was as a result of some difference in parameter variance when implementing an interface instead of "implementing" a class. I can't recall exactly why I had this, so will remove.

masaeedu:
@david-driscoll, @kwonoj Would it make sense to just remove the `lift` operator from `Subject` and inherit the behavior from `Observable`? I don't really see that it makes sense to try and feed the transformed values produced by the lifted observable into the same destination. 

E.g. you have a `Subject<number>`, which consists of an `Observable` of numbers and an `Observer` of numbers. You lift this through `.map(n => n.toString())` to produce another `?<string>`. Would this be a `Subject<string>` or an `Observable<string>`? 

Right now the code is trying to produce a new `Subject`, with a few quirks:

 - It sets the `destination` of the new `Subject` to its own destination, but this will result in our `Observer<number>` getting fed `string`s if anyone actually tries to push values to the new `Subject`
 - The `source` of the new `Subject<string>` is set to the original `Subject<number>`, which is fine, but you already get this behavior from the `Observable.prototype.lift` operator this method is overriding
 - The newly created subject is returned as an `Observable`, which indicates to me that the return value isn't expected to be used as a `Subject` anymore

Removing the override of  `lift` in `Subject` doesn't affect tests at all. It would be great if someone who thoroughly understands this machinery could provide input.

luisgabriel:
@masaeedu I think @blesh and @trxcllnt can provide some input in this matter.

src/operator/pairwise.ts
masaeedu: Testing this properly is blocked by #1228, so I'll just change these back to `this` for now.

kwonoj:
As an immediate thought I think it's fine to creating chain of subject via lift, but interesting point. Interface already limits to return observable only, limiting created object to behave as subject.. @masaeedu What about creating those as separate issue? I don't think this PR's scope can contain those changes anyway. (also this PR's already has bit of size for its own discussion)

masaeedu:
@kwonoj The reason I'm bringing this up here is because we need to make `lift` compile again for `Subject` after removing the redundant generics. If you want I can revert all changes to `Subject` (including removal of the redundant generics), and address both problems in a separate issue.

kwonoj:
I think we can choose between
- leave this PR as is until get conclusion on subject behavior
- revert current changes on subject to make this PR progress while discuss (your proposal)

I'll leave up to you which way to proceed.

masaeedu:
Ok, opened #1234 and will nuke the `Subject` changes.

src/operator/every.ts
```diff
@@ -8,38 +8,41 @@ import {Subscriber} from '../Subscriber';
...
-export function every<T>(predicate: (value: T, index: number, source: Observable<T>) => boolean,
-                         thisArg?: any): Observable<boolean> {
-  const source = this;
-  if (source._isScalar) {
-    const result: boolean = tryCatch(predicate).call(thisArg || this, source.value, 0, source);
+export function every<T>(predicate: Predicate<T>, thisArg?: any): Observable<boolean> {
+    
+  let _this: Observable<T> = this;
+  if (_this._isScalar) {
+    let scalar: ScalarObservable<T> = <any>_this;
```
blesh: When this transpiles, won't it have two additional lines where we're setting `_this = this` and then `scalar = _this`? I'm all for good typing for the _exposed_ API. Our internals, though, I just care that they work.

src/operator/every.ts
```diff
@@ -8,38 +8,41 @@ import {Subscriber} from '../Subscriber';
...
  * @returns {Observable} an Observable of booleans that determines if all items of the source Observable meet the condition specified.
  */
-export function every<T>(predicate: (value: T, index: number, source: Observable<T>) => boolean,
-                         thisArg?: any): Observable<boolean> {
-  const source = this;
-  if (source._isScalar) {
-    const result: boolean = tryCatch(predicate).call(thisArg || this, source.value, 0, source);
+export function every<T>(predicate: Predicate<T>, thisArg?: any): Observable<boolean> {
+    
+  let _this: Observable<T> = this;
```
blesh: Was this necessary? 

`source` is idiomatic for `this` throughout the library.

src/operator/isEmpty.ts
```diff
@@ -6,13 +6,13 @@ export function isEmpty(): Observable<boolean> {
...
-class IsEmptyOperator<T> implements Operator<boolean, boolean> {
-  call (observer: Subscriber<T>): Subscriber<boolean> {
+class IsEmptyOperator implements Operator<any, boolean> {
+  call (observer: Subscriber<boolean>): Subscriber<any> {
     return new IsEmptySubscriber(observer);
   }
 }
 
-class IsEmptySubscriber extends Subscriber<boolean> {
+class IsEmptySubscriber extends Subscriber<any> {
```
blesh: Seems less specific. What's the reason for the change here?

src/operator/pairwise.ts
blesh: on the `self` vs `_this` vs `source`...

So far `source` has been an idiom in this library. While I realize that these all end up as members of a class, the reality is that it's only so we can compose the operators left to right.  These operators rely on almost no state at all from the instance, and under the right conditions (say the passage of the "pipe operator" `|>`), they could just be functions.

Let's stick with `source`. There isn't a compelling reason to change it.

src/operator/scan.ts
```diff
@@ -13,14 +13,15 @@ import {errorObject} from '../util/errorObject';
  * @returns {Obervable} An observable of the accumulated values.
  */
 export function scan<T, R>(accumulator: (acc: R, x: T) => R, seed?: T | R): Observable<R> {
-  return this.lift(new ScanOperator(accumulator, seed));
+  let _this: Observable<T> = this;
```
blesh: Again, for this sort of thing, I fail to see the benefit to our end users. It's just going to add to the transpiled output, I believe.

src/operator/isEmpty.ts
masaeedu: @blesh `IsEmptySubscriber` isn't actually a subscriber of booleans. It's a `Subscriber` of anything that accepts a `Subscriber` of booleans in its constructor (to which it will pass true or false when it receives its first anything).

src/operator/every.ts
masaeedu: I hadn't discovered `source` until I got to this operator (and I'm not sure there's more than two others where it is used), so I didn't realize this was the idiom for this codebase. I have no strong opinions on the name used however, and `source` is actually much nicer.

src/operator/every.ts
masaeedu: It's true that there's no direct benefit to end users for this kind of change, but since I was doing typing fixes in this PR, I also tried to prevent mistakes in typings in the future. The `isEmpty` operator you commented on earlier is a good example of this. You're trying to take an `Observable<any>` and produce an `Observable<boolean>`, but you're using an `Operator<boolean, boolean>` to do it, which the type system would disqualify. If you cast `this` to `(<Observable<T>>this)`, the compiler complains about how the operator has the wrong type, which has the knock on effect of forcing you to fix your types all the way down to `IsEmptySubscriber`.

In and of itself this has no effect on the end user, but in many cases the user facing type signature is wrong simply because of a break in type checking somewhere down the line. It's also true that things "just work" because the people that have been contributing so far understand the codebase well, and you have good unit tests to cover your work. In some cases people might not have as good an understanding of the overall codebase (yours truly is a good example), and its nice to have accurate static typing to hold your hand.

masaeedu: I'll also point out that the extra emits have an effect proportional to the number of calls to `every` in the program, not the number of items that pass through the `every` operator. An large sized program will have a thousand to two thousand invocations (generously) of the operator functions, so an extra local variable per shouldn't have any performance impact.

All this aside, I'll be removing the `_this` aliases for this PR and will open a separate issue to discuss the pros/cons of it.

masaeedu:
@kwonoj, @blesh Ok, I've nuked the `_this` changes. The discussion regarding what to do about `Subject` has stalled a bit, so I think the prudent thing to do is to skip fixing the `lift` signature on that and the derived websocket subject for now.

src/operator/isEmpty.ts
blesh: Ah, right, I had that turned around in my head when I read it.

blesh:
@masaeedu can you rebase this PR please?

Also: just a note on PRs like this in general:

If it's internal code typings your changing (read: Nothing that a developer-user would interface with)... 

My feeling is that `<any>` is fine, unless it's going to speed up the build to do more specific casting. (which I'm not sure about)... _or_ unless it's going to add a lot to our development process to have things properly typed. (cc @david-driscoll @saneyuki @kwonoj on this note)

david-driscoll:
Would love to see this one in fairly soon, working on additional operator signatures at the moment and ran into an issue were the changes ran into these same issues.  @masaeedu :+1:

masaeedu:
Sorry about the delay folks. I'll be cleaning this up and rebasing on Saturday.

masaeedu:
@blesh Ok, rebased and removed most of the changes since they were controversial. The changes now are limited to removing redundant generics from `Operator.call` and `Observable.lift` and changing the minimum number of operators necessary to compile the project again.

This still leaves a number of operators where the internal typings (and in some cases user-facing typings) are wrong, but they still compile since `this` is `any`. Possibly these can be done in a different PR/skipped since low priority.

cc @kwonoj

luisgabriel:
@masaeedu maybe it would be worth to create an issue listing the operators with wrong typing (especially the external ones) so we could try to fix it incrementally.

kwonoj:
Sorry for delays, I'll start look into updated changes.

kwonoj:
Change looks good to me, nice PR @masaeedu :smile:

kwonoj:
I'll check in this PR after leaving bit more to see if additional suggestions around, expecting later today to tomorrow morning. cc @david-driscoll @luisgabriel @blesh for visibility.

luisgabriel:
LGTM.

Thanks @masaeedu for this nice PR. It raised important discussions regarding the typings.

david-driscoll:
LGTM :+1:

kwonoj:
Merged with f27902d, thanks for great PR @masaeedu !

PR #1216, opened 2016-01-19
chore(typings): Added generic types for use as input parameters

david-driscoll:
see #1086
cc @kwonoj

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -37,6 +37,10 @@ import {ErrorObservable} from './observable/throw';
 import {AjaxCreationMethod} from './observable/dom/ajax';
 import {WebSocketSubject} from './observable/dom/webSocket';
 
+export type ObservableOrPromise<T> = Observable<T> | Promise<T>;
+export type ArrayOrIterator<T> = Iterator<T> | ArrayLike<T>;
+export type ObservableInput<T> = ObservableOrPromise<T> | ArrayOrIterator<T>;
+
 /**
  * A representation of any set of values over any amount of time. This the most basic building block
  * of RxJS.


src/Observable.ts
```diff
@@ -37,6 +37,10 @@ import {ErrorObservable} from './observable/throw';
 import {AjaxCreationMethod} from './observable/dom/ajax';
 import {WebSocketSubject} from './observable/dom/webSocket';
 
+export type ObservableOrPromise<T> = Observable<T> | Promise<T>;
```
kwonoj: Will there be cases of this type import would be required, but importing `Observable` is not required? I was thought of having this as separate module other than `Observable` if there's those kind of cases. In case of operator it won't by all operator would need to import `Observable`.

david-driscoll: I can't think of an operator that doesn't have to import Observable, because all operators I can think of, return observables.  That's kind of why I liked the types in here, it's less work on the operator side to bring them in and use them.

kwonoj: Yup, make sense to me too. Only reason I thought of separating module is case of import without `Observable`.

kwonoj:
BTW, sorry for minor and picky - would you able to amend commit message to not to use capital letter? (https://github.com/ReactiveX/RxJS/blob/master/CONTRIBUTING.md#subject) It'll help what repo does with commit messages for automatic jobs such as changelog generating. I can do it shortly though, it'd better to author handles it.

david-driscoll:
done.

kwonoj:
Thanks, change looks good to me. I'll check this in if there isn't other suggestions around, /cc @blesh for visibility.

kwonoj:
Let me check in this around late today - tomorrow.

kwonoj:
Merged with 9194bad, thanks @david-driscoll

Issue #1112, opened 2015-12-29
Add throttleLast/inspect operator


mgamsjager:
> ### EDIT: to clarify the purpose of this issue (@blesh)
> 
> There is discussion below that brings to light a missing `inspect` or `throttleLast` operator.
> 
> It [starts here](https://github.com/ReactiveX/RxJS/issues/1112#issuecomment-167877365)
> 
> ... now for the original text:

---

Trying out Angular2 with Rxjs beta0.

Copied from this http://jsfiddle.net/bo6ynqf7/3/

From my service: 

```javascript
import {Subject, Observable} from "rxjs/Rx";
var sub = new Subject();
sub.throttleTime(1000).subscribe(res => console.log(res) );
sub.next(value);
```

When I substitute throttle with debounce then it works. Because beta0 isn't on the a CDN I can't provide a failing jsfiddle (sorry :( )

kwonoj:
Would you able to share some more details of doesn't work, doesn't it emit any values? or if it emits, doesn't it work as you've expected?

mgamsjager:
Well it does emit values because the console get's filled with values. What does not work is the throttle itself. The value variables are keys that are entered in an input field. I would expect the throttle to slow down the speed at which the keys are logged to the console

trxcllnt:
@kwonoj I'd venture a guess what @mgamsjager wants is (what I've been calling) `inspectTime`. Where `throttleTime` emits the value that triggers the `throttle` operation (the first event), I believe it's also valuable to have an operator that emits the *last value* that came in before the `throttle` delay completes. Here's a reference implementation:

```es6
import { asap } from 'rxjs/scheduler/asap';
import { Scheduler } from 'rxjs/Scheduler';
import { Subscriber } from 'rxjs/Subscriber';
import { Observable } from 'rxjs/Observable';
import { Subscription } from 'rxjs/Subscription';

function inspectTime(delay, scheduler = asap) {
    return this.lift(new InspectTimeOperator(delay, scheduler));
}

class InspectTimeOperator {
    constructor(delay, scheduler) {
        this.delay = delay;
        this.scheduler = scheduler;
    }

    call(subscriber) {
        return new InspectTimeSubscriber(subscriber, this.delay, this.scheduler);
    }
}

class InspectTimeSubscriber extends Subscriber {

    constructor(destination, delay, scheduler) {
        super(destination);
        this.delay = delay;
        this.value = null;
        this.hasValue = false;
[...]

PR #1219, opened 2016-01-19
feat(operator): adds inspect and inspectTime operators

trxcllnt:
Addresses #1112.


src/Observable.ts
```diff
@@ -215,6 +215,8 @@ export class Observable<T> implements CoreOperators<T>  {
                elementSelector?: (value: T) => R,
                durationSelector?: (group: GroupedObservable<K, R>) => Observable<any>) => Observable<GroupedObservable<K, R>>;
   ignoreElements: () => Observable<T>;
+  inspect: (durationSelector: (value: T) => Observable<any> | Promise<any>) => Observable<T>;
```
kwonoj: doesn't `CoreOperators<T>` also need to update interface signature?

trxcllnt: ah, yes. Rx.DOM also.

spec/operators/inspectTime-spec.js
```diff
@@ -0,0 +1,136 @@
...
+var Observable = Rx.Observable;
+var Scheduler = Rx.Scheduler;
+
+describe('Observable.prototype.inspectTime()', function () {
+  it.asDiagram('inspectTime(50)')('should emit the last value in each time window', function () {
+    var e1 =   hot('-a-x-y----b---x-cx---|');
+    var subs =     '^                    !';
+    var expected = '------y--------x-----|';
+
+    var result = e1.inspectTime(50, rxTestScheduler);
```
staltz: Just an information (no need to update this PR), you could write `var t = time('------|')` to get `t=50` instead of hard coding 50.

trxcllnt: @staltz cool. I copied the tests from [`throttleTime`](https://github.com/ReactiveX/RxJS/blob/master/spec/operators/throttleTime-spec.js) so as to be consistent. perhaps we can clean up the tests in a followup PR.

blesh:
Merged with 54f957b... Thanks @trxcllnt 

... although I think we might want to just have this one in KitchenSink, since I'm unsure how common it's use-cases are.

PR #1224, opened 2016-01-20
refactor(observable): made create strongly typed by creating a factor…

david-driscoll:
…y for derived types

------
The last sticking point to enhance the typings, is that methods like `Observable.create` and `Subject.create` are defined as functions, due to how TypeScript inheritance works.  This also conflicts with other sub-typed observables, such as `ArrayObservable` and so on.

This changes sub-typed observables to no longer use the `create` method, but instead have a stand-alone `create` function that is exported alongside the Observable, instead of as a static member.

Subject is a special case, the default factory method is `createSubject` but we overwrite `Subject.create` to make sure this change is backwards compatible.

I have an alternative branch, that changes [sub-typed names to `factory`](https://github.com/david-driscoll/RxJS-1/tree/divide-subject) instead, but after working with it, this change seems better because it doesn't introduce a new name.

### DIFF
--- a/spec/observables/ErrorObservable-spec.js
+++ b/spec/observables/ErrorObservable-spec.js
@@ -1,6 +1,7 @@
 /* globals describe, it, expect, rxTestScheduler, expectObservable*/
 var Rx = require('../../dist/cjs/Rx');
 var ErrorObservable = require('../../dist/cjs/observable/throw').ErrorObservable;
+var create = require('../../dist/cjs/observable/throw').create;
 var Observable = Rx.Observable;
 
 describe('ErrorObservable', function () {
@@ -11,13 +12,13 @@ describe('ErrorObservable', function () {
 
   it('should create ErrorObservable via static create function', function () {
     var e = new ErrorObservable('error');
-    var r = ErrorObservable.create('error');
+    var r = create('error');
 
     expect(e).toEqual(r);
   });
 
   it('should accept scheduler', function () {
-    var e = ErrorObservable.create('error', rxTestScheduler);
+    var e = create('error', rxTestScheduler);
 
     expectObservable(e).toBe('#');
   });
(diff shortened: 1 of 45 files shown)


kwonoj:
Would you mind to take #1086 and #1216 as priority? I believe both PR nearly gets completed review, it'd better to let those checked in first.

david-driscoll:
Nope, just offering it up for discussion early.

kwonoj:
Thanks for clarification. Let me mark this as discussion label for now, to literally discuss about change itself.

kwonoj:
#1086 is now landed, so we can get back to this PR into discussion.

So for my understanding, this PR does

- now each observable or operator does not have static `create` 
- instead, those create are lives inside of same module and exported
- type reference is being resolved by importing create function

is my understanding correct? Would you mind to elaborate bit more this PR aims to solve about 'conflict' you mentioned in description?

blesh:
> now each observable or operator does not have static create

This is a red flag for me. Those create methods are important pieces of the API... They're consistent on all of our types, and provide a less "classy" way of instantiating those types. A way that supports `call` and `apply` and doesn't require `new`. Arguably this is still true with the exported creation methods, but having them as statics on the type itself has been a part of the RxJS API for quite some time.

david-driscoll:
Would an alternative name be something that's possible?

I suppose we can wait for Microsoft/TypeScript#4628 if it lands, it's just less than idea today.

kwonoj:
@david-driscoll , since static interface `create` is core design intention I came to think this PR loses major portions of its intentions. It is great appreciate for taking effort to drive this, but for now I'd like to see where https://github.com/Microsoft/TypeScript/issues/4628 is heading as you mentioned meanwhile leaving current design as-is. What do you think? Or do you have other suggestion to be updated in this PR?

david-driscoll:
We can probably close in favor of the issue in TypeScript, just have to live with the problem for now. :)

PR #1232, opened 2016-01-21
refactor(FutureAction): Make FutureAction & derived classes' members protected

saneyuki:
There are no reasons to make these members public!

- `FutureAction._schedule()`
- `FutureAction._unsubscribe()`


kwonoj:
Change looks good to me. @saneyuki , would you update commit message as same as https://github.com/ReactiveX/RxJS/pull/1212?

saneyuki:
@kwonoj okay. updated!

kwonoj:
Merged with 2ae8f93, thanks @saneyuki .

PR #1236, opened 2016-01-22
fix(MergeMapSubscriber): clearfy type definitions for MergeMapSubscriber's members

saneyuki:
- `MergeMapSubscriber.buffer` should be `T[]` (`Array<T>`) because it is given a source value `T`.
- The type of 1st argument of `MergeMapSubscriber._innerSub` should be `Observable<R>` because the result of `MergeMapSubscriber.project` will be passed to it.


kwonoj:
I think change makes sense. @saneyuki , what about squashing commits into single one? for these changes I could think of single commits for `MergeMapSubscriber` type definition updates.

saneyuki:
@kwonoj squashed!

kwonoj:
Merged with 4ee5f02, thanks @saneyuki

PR #1238, opened 2016-01-22
Reorganize files, remove patch generation

blesh:
**Removes "support" files**: The `-support` files were born when we were trying to use `export default` for everything. Now that we've done away with that, they don't make any sense.

**Removes "static" files**: Again, because we're not default exporting anything any more, we can have the static implementations live along side their member counterparts.

**Moves "-static" patch files**: Takes all the of the patch files like `add/operator/zip-static` and moves them to a more appropriate `add/observable/zip`.

**Removes patch generator**: We don't need to generate patch files anymore. It's something we can maintain manually, as it isn't a lot of work, and we do have special cases from time to time.

**Renames observable class files**. Since we no longer have a patch generator that is using file names to create other files, we can name the Observable files something sane again. `ArrayObservable` is once again under `ArrayObservable.ts` and not `fromArray.ts`.

**Renames InfiniteObservable to NeverObservable**: This is a non-breaking change as we don't export this type, per say. It was done because the name is more descriptive and, like other Observable subclasses, it now aligns with the method associated with it.


luisgabriel:
LGTM.

src/add/observable/fromEvent.ts
```diff
@@ -3,7 +3,7 @@
  * Any manual edits to this file will be lost next time the script is run.
```
kwonoj: Some patch operator removes this comments - expected to leave this?

blesh: Oops, nope, I meant to remove that.

tools/generate-operator-patches.ts
```diff
@@ -60,10 +60,7 @@ const AliasMethodOverrides = {
   }
 };
 
-const header = `/**
- * Everything in this file is generated by the 'tools/generate-operator-patches.ts' script.
- * Any manual edits to this file will be lost next time the script is run.
- **/`;
+const header = '';
```
luisgabriel: I think you should remove this script also.

PR #1239, opened 2016-01-23
chore(docs): make docgen.sh executable

rgbkrk:
The README suggests running `npm run build_docs`. In order
for that to work, docgen.sh needs to be executable.


kwonoj:
Merged with 446ae95, thanks @rgbkrk !

PR #1241, opened 2016-01-23
docs(Subject): Document changes to Subject.create

rgbkrk:
Guide those migrating from RxJS 4 who are using `Subject.create` to
let them know about the new argument ordering.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -31,6 +31,12 @@ Observers and Subjects alike) has changed.
 
 So what was once `subject.onNext("hi")` is now `subject.next("hi")`.
 
+## Changes to Subject.create
+
+Along with observers now being subscribers, the order of the arguments to `Subject.create` has changed.
+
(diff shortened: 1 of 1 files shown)


blesh:
Actually, it looks like we have the arguments reversed. It should be `Subject.create(subscriber, observable)`, because `Subscriber` is an `Observer`. :\

@trxcllnt, I think you implemented this one, is there a reason for the reversal of argument order here?

trxcllnt:
@blesh no, just typed it that way a while back and it stuck.

kwonoj:
Seems it's clarified and we can introduce changes to interface actually, so I have filed https://github.com/ReactiveX/RxJS/issues/1285 to track those and closing this PR. @rgbkrk , huge appreciate to catch this and take effort to bring PR for those - let's track root reason via filed issue. Thanks again!

rgbkrk:
Thanks for tackling this.

PR #1243, opened 2016-01-24
refactor(WebSocketSubject): make 'WebSocketSubject._unsubscribe' protected

saneyuki:


saneyuki:
fixed the conflict

kwonoj:
I'm seeing couple of commits (i.e, https://github.com/saneyuki/RxJS/commit/73cf3cda8dfe1565e11e573db9362097788a706c) are under same category of this refactoring. What about creating single commit-PR to amends this changes at once? it'll give easier history lookup (in my opinion).

saneyuki:
@kwonoj 

May I mix their change into single pull request? I felt these have difference context (https://github.com/saneyuki/RxJS/commit/73cf3cda8dfe1565e11e573db9362097788a706c is a part of cleanup for ConnectableObservable and this is a part of WebSocketSubject)

kwonoj:
I think so, if PR contains single scope of changes. Feel freely close PR as needed.

saneyuki:
okay, I cherry-pick this patch on #1244

PR #1244, opened 2016-01-24
refactor(ConnectableObservable): hide ConnectableObservable's members from public

saneyuki:
- `ConnectableObservable<T>`'s members may be overridden by user. So this change make them `protected` for extensibility.
- `RefCountSubscriber<T>_resetConnectable()` belongs to ``RefCountSubscriber<T>` which is private class in ConnectableObservable.ts. So this change make it `private`.
- To protect unexpected modifying `ConnectableObservable<T>.subject/subscription` from others, this change make them `protected` and add `_closeSubscription()` for  `ConnectableSubscription`

### DIFF
--- a/src/observable/ConnectableObservable.ts
+++ b/src/observable/ConnectableObservable.ts
@@ -5,19 +5,19 @@ import {Subscription} from '../Subscription';
 
 export class ConnectableObservable<T> extends Observable<T> {
 
-  subject: Subject<T>;
-  subscription: Subscription;
+  protected subject: Subject<T>;
+  protected subscription: Subscription;
 
-  constructor(public source: Observable<T>,
+  constructor(protected source: Observable<T>,
               protected subjectFactory: () => Subject<T>) {
     super();
   }
 
   protected _subscribe(subscriber: Subscriber<T>) {
-    return this._getSubject().subscribe(subscriber);
+    return this.getSubject().subscribe(subscriber);
   }
 
-  _getSubject() {
+  protected getSubject() {
     const subject = this.subject;
     if (subject && !subject.isUnsubscribed) {
       return subject;
@@ -31,13 +31,22 @@ export class ConnectableObservable<T> extends Observable<T> {
     if (subscription && !subscription.isUnsubscribed) {
       return subscription;
     }
-    subscription = source.subscribe(this._getSubject());
+    subscription = source.subscribe(this.getSubject());
     subscription.add(new ConnectableSubscription(this));
     return (this.subscription = subscription);
   }
 
   refCount(): Observable<T> {
     return new RefCountObservable(this);
+  }
+
+  /**
+   * This method is opened for `ConnectableSubscription`.
+   * Not to call from others.
+   */
+  _closeSubscription(): void {
+    this.subject = null;
+    this.subscription = null;
   }
 }
 
@@ -46,10 +55,9 @@ class ConnectableSubscription extends Subscription {
     super();
   }
 
-  _unsubscribe() {
+  protected _unsubscribe() {
     const connectable = this.connectable;
-    connectable.subject = null;
-    connectable.subscription = null;
+    connectable._closeSubscription();
     this.connectable = null;
   }
 }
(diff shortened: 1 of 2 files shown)


saneyuki:
@kwonoj updated. could you review this?

kwonoj:
Would you mind to squash some of commits? Seems it doesn't need to breakdown into those numbers of commits.

src/observable/ConnectableObservable.ts
```diff
@@ -5,10 +5,10 @@ import {Subscription} from '../Subscription';
 
 export class ConnectableObservable<T> extends Observable<T> {
 
-  subject: Subject<T>;
-  subscription: Subscription;
+  protected _subject: Subject<T>;
```
kwonoj: Is there reason to make private field starts with `_`? Current codebase usually do not use those convention only except if field has separate getter / setter.

src/observable/ConnectableObservable.ts
```diff
@@ -17,28 +17,35 @@ export class ConnectableObservable<T> extends Observable<T> {
...
+    return (this._subscription = subscription);
   }
 
   refCount(): Observable<T> {
     return new RefCountObservable(this);
   }
+
+  // This method is opened for `ConnectableSubscription`.
+  // Not to call from others.
+  _closeSubscription(): void {
```
kwonoj: Though it's specific usage only from certain caller it's public interface, would like to avoid `_`. Hope to have something similar to `internal` modifier in TypeScript in further.

src/observable/ConnectableObservable.ts
saneyuki: @kwonoj 

Ah, I felt there are some implicit customs which is start with `_` to mark a property as a private.
okay. I'll fix it

src/observable/ConnectableObservable.ts
saneyuki: @kwonoj 

I think we should keep `_` for this method. `internal` modifier is not yet. I agree to use `private` or `protected` instead of `_`.

But all public field of all exported object is published as type definitions which is used from some autocomplete system. We need to some marker style until `internal` modifier or slimier one to mark it is not a public callable member.

saneyuki:
updated

src/observable/ConnectableObservable.ts
kwonoj: Once it's published as public, regardless of name starts with _ or not it is marked as consumer-usable. In my opinion, for this case I'd revert back to original code to expose `subject` / `subscription` directly for simple changes. Or either could think of more strict dependencies but initial assumption it'll require amount of change and might be cumbersome impact performance.

saneyuki: okay. I'll remove `_`

saneyuki:
@kwonoj updated

src/observable/ConnectableObservable.ts
```diff
@@ -39,6 +39,13 @@ export class ConnectableObservable<T> extends Observable<T> {
   refCount(): Observable<T> {
     return new RefCountObservable(this);
   }
+
+  // This method is opened for `ConnectableSubscription`.
```
kwonoj: it's just curiosity and assume it's not - does this comment appears editor's intellisense for typings? does JSDoc style comment appears?

saneyuki: At least on vscode, JSDoc style will appear.

saneyuki:
fixed

src/observable/ConnectableObservable.ts
```diff
@@ -39,6 +39,15 @@ export class ConnectableObservable<T> extends Observable<T> {
   refCount(): Observable<T> {
     return new RefCountObservable(this);
   }
+
+  /**
+   * This method is opened for `ConnectableSubscription`.
+   * Not to call from others.
+   */
```
blesh: Rather than this comment, I'd just name it `_closeSubscription` ... generally, `_name` is idiomatic in the JavaScript community for "this is not a public API".

blesh:
This LGTM, other than the one comment I left. Once you get that fixed it can be merged.

src/observable/ConnectableObservable.ts
kwonoj: Actually, I was suggested `closeSubscription` in opposite to @saneyuki did in first way. My bad, and sorry @saneyuki for making redo work.

saneyuki:
@kwonoj @blesh 

I fixed non public api to `_closeSubscription`.

kwonoj:
Merged with 3c43c84, thanks @saneyuki

PR #1245, opened 2016-01-25
feat(delayWhen): add delayWhen operator

### LINKED ISSUE #1202: Missing polymorphic behavior (delayWhen) in RxJS 5
I was trying to create a randomly delayed Observable like this:

```javascript
"use strict";

var Rx = require('rx');
// var Rx = require('rxjs/Rx');

var source = Rx.Observable.range(0, 100)
    .take(20)
    .delay(x => Rx.Observable.timer(Math.floor(Math.random() * 10000)));

var delayedSubscriber = source.subscribe(
  function (x) { console.log('onNext: %s', x); },
  function (e) { console.log('onError: %s', e); },
  function () { console.log('onCompleted'); }
);
```

I am not sure if this is the correct way to do what I intend, but I noticed a 
different behavior when I tested in RxJS 5 and nothing happened.
Is it an issue, or did something change in the semantics that I didn't read ?

Thanks.

kwonoj:
closes #1202 

This PR implements `delayWhen` operator supports selector with subscription delay current `delay` operator doesn't supports from RxJS4.

### DIFF
--- a/src/operator/delayWhen.ts
+++ b/src/operator/delayWhen.ts
@@ -1,0 +1,144 @@
+import {Operator} from '../Operator';
+import {Subscriber} from '../Subscriber';
+import {Observable} from '../Observable';
+import {Subscription} from '../Subscription';
+
+/**
+ * Returns an Observable that delays the emission of items from the source Observable
+ * by a subscription delay and a delay selector function for each element.
+ * @param {Function} selector function to retrieve a sequence indicating the delay for each given element.
+ * @param {Observable} sequence indicating the delay for the subscription to the source.
+ * @returns {Observable} an Observable that delays the emissions of the source Observable by the specified timeout or Date.
+ */
+
+export function delayWhen<T>(delayDurationSelector: (value: T) => Observable<any>,
+                             subscriptionDelay?: Observable<any>): Observable<T> {
+  if (subscriptionDelay) {
+    return new SubscriptionDelayObservable(this, subscriptionDelay)
+            .lift(new DelayWhenOperator(delayDurationSelector));
+  }
+  return this.lift(new DelayWhenOperator(delayDurationSelector));
+}
+
+class DelayWhenOperator<T> implements Operator<T, T> {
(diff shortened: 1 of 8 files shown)


blesh:
Looks okay overall. I think we should try to get this method to use common logic such as `subscribeToResult` to enable things like Promise and Observablesque support.

kwonoj:
Thanks. I'll give a try to refactor now to see if I can make it.

blesh:
In fact, I bet there are quite a few more operators that should probably use that code path to ensure they support Promises and the like.

kwonoj:
OK, refactored by using `subscribeToResult` with `OuterSubscriber`. I came to remember reason why I dropped it first time, it was due to manage subscription for some cases (like notifier does not emits but completes) was bit tricky. For this change I've ended up using `Map` with outer value index as key, any suggestion would be appreciated if you think there's better way.

src/operator/delayWhen.ts
```diff
@@ -0,0 +1,155 @@
...
+
+interface DelayNotiferSubContext<T> {
+  index: number;
+  subscription: Subscription;
+  value: T;
+}
+
+class DelayWhenSubscriber<T, R> extends OuterSubscriber<T, R> {
+  private completed: boolean = false;
+  private delayNotifiers: Array<DelayNotiferSubContext<T>> = [];
```
kwonoj: now tried to use simple array instead of map, subscription deletion via lookup & splice.

kwonoj:
PR should be updated after #1252 checked in, marking as blocked for now.

kwonoj:
updated PR to utilize innersub passed by notifyNext, also removes trycatcher as recent changes.

kwonoj:
I'll check this in later today to see any other suggestions around.

kwonoj:
actually noticed code was using old trycatcher, updated PR to remove it.

kwonoj:
Merged with 17122f9.

PR #1249, opened 2016-01-25
refactor(performance): improves shape of SafeSubscriber so JIT can optimize call patterns.

trxcllnt:
@blesh


blesh:
Merged with b727355.

Per discussion with @trxcllnt ... while the initial commit seemed to _slightly_ reduce perf,  it was because it was also adding handling for errors thrown by user-land  handlers. Given that, it did indeed improve performance over an implementation I had that didn't change the shape of the type much.

... while playing with merging this PR, I made some changes that did indeed double the performance again.

So the net of this PR is:

1. Perf loss from necessary changes to error handling (@trxcllnt)
2. Perf gain from shape change (@trxcllnt)
3. Perf gain from changes to error handling (me).

```
                                         |                     RxJS 4.0.7 |              RxJS 5.0.0-beta.1 |          factor |      % improved
                    mergemap - immediate |                 3,000 (±1.01%) |                29,092 (±0.66%) |           9.70x |          869.6%
                                mergemap |                 1,263 (±0.82%) |                 6,964 (±0.65%) |           5.51x |          451.2%
```

blesh:
For contrast, here's what it was in master before it all went down:

```
                                         |                     RxJS 4.0.7 |              RxJS 5.0.0-beta.1 |          factor |      % improved
                    mergemap - immediate |                 2,988 (±0.86%) |                17,586 (±0.77%) |           5.88x |          488.5%
                                mergemap |                 1,235 (±0.54%) |                 5,890 (±0.56%) |           4.77x |          377.1%
```

PR #1251, opened 2016-01-26
Perf or no perf, there is no try...

blesh:
I discovered today that adding custom tryCatcher members to operators with try/catch optimization concerns actually improved performance quite a bit.

So far: `map` doubled in overall speed (from factor of 1.7x to 3.2x on my machine),
`mergeMap` improved by a "factor" of a little more than 1x
`zip` improved by a "factor" of a little more than 1x-2x.

("factor" coming from micro perf output).

This will be part of an effort to go through and examine improvements that can be made where `tryCatch`/`errorObject` is being used.

See commits for details.

All of these perf tests are of course in V8/Node, however I don't see any reason they wouldn't carry over to other JIT'ed runtimes.

cc/ @trxcllnt

### DIFF
--- a/src/operator/mergeMap.ts
+++ b/src/operator/mergeMap.ts
@@ -2,8 +2,6 @@ import {Observable} from '../Observable';
 import {Operator} from '../Operator';
 import {Subscriber} from '../Subscriber';
 import {Subscription} from '../Subscription';
-import {tryCatch} from '../util/tryCatch';
-import {errorObject} from '../util/errorObject';
 import {subscribeToResult} from '../util/subscribeToResult';
 import {OuterSubscriber} from '../OuterSubscriber';
 
@@ -39,7 +37,7 @@ export class MergeMapOperator<T, R, R2> implements Operator<T, R> {
 
 export class MergeMapSubscriber<T, R, R2> extends OuterSubscriber<T, R> {
   private hasCompleted: boolean = false;
-  private buffer: T[] = [];
+  private buffer: Observable<any>[] = [];
   private active: number = 0;
   protected index: number = 0;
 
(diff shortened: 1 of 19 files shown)


src/operator/mergeMap.ts
```diff
@@ -78,17 +81,22 @@ export class MergeMapSubscriber<T, R, R2> extends OuterSubscriber<T, R> {
...
-      }
+    if (this.resultSelector) {
+      this._notifyResultSelector(outerValue, innerValue, outerIndex, innerIndex);
     } else {
-      destination.next(innerValue);
+      this.destination.next(innerValue);
+    }
+  }
+
+  _notifyResultSelector(outerValue: T, innerValue: R, outerIndex: number, innerIndex: number) {
```
luisgabriel: private?

luisgabriel:
:+1:

kwonoj:
Cool. How do we determine other operators can apply same pattern? try & compare perf manually?

blesh:
Yes, I'm trying and comparing perf manually

blesh:
Merging this work for now, it's about half done, but it gives a lot of examples of how to improve perf by using tryCatch handling that's a little more focused than `tryCatch`/`errorObject`

blesh:
Eventually, `tryCatch` and `errorObject` can probably be removed from this project entirely.

kwonoj:
I'll try to apply same approach on other operators as well.

> Eventually, tryCatch and errorObject can probably be removed from this project entirely.

: makes sense - in that case if refactoring does not brings noticeable perf improvement, will it still considered as good to checked in by removing usage of custom tryCatcher?

saneyuki:
@blesh

> All of these perf tests are of course in V8/Node, however I don't see any reason they wouldn't carry over to other JIT'ed runtimes.

How about ChakraNode (https://github.com/Microsoft/node)? We may be able to use it as a sample.

PR #1254, opened 2016-01-27
perf(debounce): remove tryCatch/errorObject for custom tryCatching

kwonoj:
This PR applies approach to remove custom trycatcher on debounce operator.

Numbers are

```
before
debounce - immediate |  5,530 (±7.51%) |      23,628 (±3.99%) |   4.27x |   327.3%

after
debounce - immediate |  5,593 (±9.91%) |      27,302 (±2.89%) |   4.88x |   388.1%
```

not sure if this can be considered noticeable improvement.

### DIFF
--- a/src/operator/debounce.ts
+++ b/src/operator/debounce.ts
@@ -3,9 +3,6 @@ import {Observable} from '../Observable';
 import {Subscriber} from '../Subscriber';
 import {Subscription} from '../Subscription';
 
-import {tryCatch} from '../util/tryCatch';
-import {errorObject} from '../util/errorObject';
-
 import {OuterSubscriber} from '../OuterSubscriber';
 import {InnerSubscriber} from '../InnerSubscriber';
 import {subscribeToResult} from '../util/subscribeToResult';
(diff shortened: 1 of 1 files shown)


kwonoj:
I'll check this in around later today if there isn't any suggestion.

src/operator/debounce.ts
```diff
@@ -46,21 +43,25 @@ class DebounceSubscriber<T, R> extends OuterSubscriber<T, R> {
...
-      subscription = subscribeToResult(this, duration);
-      if (!subscription.isUnsubscribed) {
-        this.add(this.durationSubscription = subscription);
-      }
+    try {
+      duration = this.durationSelector.call(this, value);
+    } catch (err) {
+      this.destination.error(err);
+      return;
+    }
```
luisgabriel: I think @blesh's suggestion on [#1266 (comment)](https://github.com/ReactiveX/RxJS/pull/1266#discussion_r51300242) also applies here. (To move the try/catch block to a separate function.)

blesh: That's correct. A function that had the bare minimum try catch implementation in it. All other logic in other functions.

kwonoj: Got it, will update PR accordingly.

kwonoj:
Updated PR to separate trycatch as function.

kwonoj:
Merged with 90bf3f1.

Issue #1256, opened 2016-01-27
Potential resource leak in unsubscribe loop


briancavalier:
Subscription [builds up an array of subscriptions in `add()`](https://github.com/ReactiveX/RxJS/blob/master/src/Subscription.ts#L71), which is later [torn down in a loop](https://github.com/ReactiveX/RxJS/blob/master/src/Subscription.ts#L40-L45) when unsubscribe() is called.  It seems that if a particular `subscription.unsubscribe` call throws, the subscriptions later in the array will never be unsubscribed.  If those subscriptions represent resources, perhaps from Observable.create calls as mentioned in [Guidelines bullet 4](https://github.com/ReactiveX/RxJS/blob/master/doc/operator-creation.md#guidelines), it appears that the resources would be leaked.

Is there a potential leak there?

trxcllnt:
@briancavalier yes there is. unsubscribe should be called safely.

blesh:
Thanks for pointing that out @briancavalier! Nice of the author of Most.js to lend a hand. 💖

trxcllnt:
@blesh @briancavalier upon further review, this may not be so clear cut, and I have a feeling this is related to #1186.

When an error is thrown from the private `_unsubscribe` method, what should we do with it? We could:

1. let it interrupt execution, potentially leaking memory.
2. catch it, continue disposing, and re-throw the error at the end

If we take the second approach, we'll also have to wrap the `unsubscribe` calls on our child subscriptions, since __their__ unsubscribe methods may also throw. Subscription's `unsubscribe` method would look something like this:

```es6
  unsubscribe(): void {

    if (this.isUnsubscribed) {
      return;
    }

    this.isUnsubscribed = true;

    let errors: Array<any>;
    let errorHappened = false;

    const { _unsubscribe, _subscriptions } = (<any> this);

    (<any> this)._subscriptions = null;

    if (isFunction(_unsubscribe)) {
      try {
        _unsubscribe.call(this);
      } catch (e) {
        errorHappened = true;
        (errors || (errors = [])).push(e)
      }
    }

    if (isArray(_subscriptions)) {

      let index = -1;
      const len = _subscriptions.length;

      while (++index < len) {
        const subscription = _subscriptions[index];
        if (isObject(subscription)) {
          try {
            subscription.unsubscribe();
          } catch (e) {
            errorHappened = true;
            (errors || (errors = [])).push(e)
          }
        }
      }
    }

    if (errorHappened) {
[...]

PR #1258, opened 2016-01-27
Fix leaks when unsubscribe functions error

blesh:
Relates to discussion in #1256 

Basically this adds a context that is passed around internally that collects all errors and throws them as a group in an `UnsubscriptionError` when the subscription tree is unsubscribed completely.

### DIFF
--- a/src/Subscription.ts
+++ b/src/Subscription.ts
@@ -88,7 +88,7 @@ export class Subscription {
         if (sub.isUnsubscribed || typeof sub.unsubscribe !== 'function') {
           break;
         } else if (this.isUnsubscribed) {
-            sub.unsubscribe();
+          sub.unsubscribe();
         } else {
           ((<any> this)._subscriptions || ((<any> this)._subscriptions = [])).push(sub);
         }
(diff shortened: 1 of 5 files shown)


trxcllnt:
@blesh I'm all for this PR, but I think if we're going to have special error handling logic to avoid potential memory leaks here, we should be consistent and do it for subscription as well (#1191).

src/Subscription.ts
```diff
@@ -38,12 +46,26 @@ export class Subscription {
       const len = _subscriptions.length;
 
       while (++index < len) {
-        const subscription = _subscriptions[index];
-        if (isObject(subscription)) {
-          subscription.unsubscribe();
+        const sub = _subscriptions[index];
+        if (isObject(sub)) {
+          let trial = tryCatch(sub.unsubscribe).call(sub);
```
trxcllnt: @blesh is there a reason this is `let trial` instead of `const trial`? same for above

blesh: Nope. Sorry I missed this comment in my zeal to merge.

blesh:
@trxcllnt no doubt, I'm just waiting for a rehash of that PR, since I think a lot has changed around that code since you made the PR.

PR #1260, opened 2016-01-28
fix(Notification): ensure `do` and `observe` handle partial observation

blesh:
- `do` will now handle notifying a handler that isn't present. (e.g. `notification.do(null, errorFn, null)`)
- `observe` will now handle partial observers.

Relates to #1211...

Rather than enforce the shape of the Observer passed to `observe`, we can allow any shape of Observer. If someone wants to use a generator as an observer passed to `Notification.prototype.observe` I see no reason to prevent them just because they don't have `error` or `complete` functions.

### DIFF
--- a/src/Notification.ts
+++ b/src/Notification.ts
@@ -11,11 +11,11 @@ export class Notification<T> {
   observe(observer: Observer<T>): any {
     switch (this.kind) {
       case 'N':
-        return observer.next(this.value);
+        return observer.next && observer.next(this.value);
       case 'E':
-        return observer.error(this.exception);
+        return observer.error && observer.error(this.exception);
       case 'C':
-        return observer.complete();
+        return observer.complete && observer.complete();
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/Notification.ts
```diff
@@ -11,11 +11,11 @@ export class Notification<T> {
   observe(observer: Observer<T>): any {
```
kwonoj: type definition still requires `observer` interface, so typescript won't be able to pass partial implementation of observers. Should it be updated in this PR, or leave it for further type enhancement?

blesh: Hmm... I thought I also had a PR to change the Observer interface to allow partials as well...

kwonoj: Not yet, https://github.com/ReactiveX/RxJS/blob/master/src/Observer.ts#L3-L5 seems we may need those changes also. (better to include in this PR as one shot changes?)

blesh: Just filed an issue related to this here: https://github.com/Microsoft/TypeScript/issues/6806

kwonoj: This https://github.com/Microsoft/TypeScript/issues/4889 might be relevant to what we want?

blesh:
Closing in favor of #1282

PR #1261, opened 2016-01-28
feat(Subject): implement asObservable

### LINKED ISSUE #1108: Subject.asObservable() missing?
Hi

either it is missing, or it is missing in the migration guide?

Frank

kwonoj:
closes #1108 

This PR tries to implement `asObservable` for subjects.

### DIFF
--- a/spec/Subject-spec.js
+++ b/spec/Subject-spec.js
@@ -1,4 +1,4 @@
-/* globals describe, it, expect */
+/* globals describe, it, expect, hot, expectObservable */
 var Rx = require('../dist/cjs/Rx');
 
 var Subject = Rx.Subject;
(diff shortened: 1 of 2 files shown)


spec/Subject-spec.js
```diff
@@ -508,4 +508,77 @@ describe('Subject', function () {
 
     source.subscribe(subject);
   });
+
+  describe('asObservable', function () {
+    it('should hide subject', function () {
+      var subject = new Rx.Subject();
+      expect(subject).not.toEqual(subject.asObservable());
+    });
```
blesh: Seems like there should be a test like `expect(result instanceof Observable).toBe(true)` and `expect(result instanceof Subject).toBe(false)`

kwonoj: I'll quickly add it.

src/Subject.ts
```diff
@@ -123,6 +123,12 @@ export class Subject<T> extends Observable<T> implements Observer<T>, Subscripti
     this._complete();
   }
 
+  asObservable(): Observable<T> {
+    return new Observable((observer: Observer<any>) => {
```
blesh: I think all you needed to do here was:

```TypeScript
const observable = new Observable();
observable.source = this;
return observable;
```

kwonoj: Ah, thanks for suggestion. I'll try right away to update.

kwonoj: Actually, just noticed `source` is protected member so can't assign it directly. exposing it as public member seems not legit - what do you think?

blesh: Then make a derived class `SubjectObservable` or something, and it can do:

```TypeScript
export class SubjectObservable<T> extends Observable<T> {
  constructor(source: Subject<T>) {
    super();
    this.source = source;
  }
}
```

easy peasy.

david-driscoll: worst case just cast around it.  If the TypeScript gets in the way of getting things done, work around it. :smile:

kwonoj: Thought about casting too a bit,but wanted tobstay in type guarded world ;) think derived class would work in this case. Thanks!

kwonoj:
Refactored PR to using derived class to remove closure as suggested. Derived class isn't exported by subject is only consumer currently.

kwonoj:
Merged with aca3dd0.

PR #1264, opened 2016-01-29
chore(tools): remove script to generate operator patches

luisgabriel:
This is not being used anymore.


kwonoj:
Merged with bb308f4, thanks @luisgabriel .

PR #1268, opened 2016-01-30
feat(scanCase): adds scanCase operator and unit tests

jwoodhull:
This PR implements as scanCase operator.  ScanCase allows items to have different accumulator functions depending on the result of a select function.

### DIFF
--- a/spec/operators/scanCase-spec.js
+++ b/spec/operators/scanCase-spec.js
@@ -1,0 +1,247 @@
+/* globals describe, it, fit, expect, expectObservable, expectSubscriptions, cold, hot */
+var Rx = require('../../dist/cjs/Rx');
+var Observable = Rx.Observable;
+
+describe('Observable.prototype.scanCase()', function () {
+  var concatSelect = function () {
+    return 'concat';
+  };
+
+  var addSelect = function () {
+    return 'add';
+  };
+
+  var strategy = {
+    concat: function (acc, x) {
+      return [].concat(acc, x);
+    },
+    add: function (acc, x) {
+      return acc + x;
+    },
+    addVal: function (acc, x) {
+      return acc + x.value;
+    },
+    subtract: function (acc, x) {
+      return acc - x.value | x;
+    },
+    throw: function () {
+      throw 'bad!';
+    }
+  };
+
+  it.asDiagram(
+    'scanCase((curr) => curr.op,' +
+    '{ add: (acc, curr) => acc + curr.value, substract: (acc, curr) => acc - curr.value },' +
+    '0)')('should scanCase', function () {
+      var values = {
+        a: { op: 'addVal', value: 1 }, b: { op: 'addVal', value: 3 }, c: { op: 'subtract', value: 2 },
+        x: 1, y: 4, z: 2
+      };
+      var e1 =        hot('--a--b--c|', values);
+      var e1subs =        '^        !';
+      var expected = '--x--y--z|';
+
+      var select = function (v) {
+        return v.op;
+      };
+
+      expectObservable(e1.scanCase(select, strategy, 0)).toBe(expected, values);
+      expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    });
+
+  it('shoud scanCase things', function () {
+    var e1 = hot('--a--^--b--c--d--e--f--g--|');
+    var e1subs =      '^                    !';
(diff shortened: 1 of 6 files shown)


kwonoj:
Hi, appreciate for taking effort for this PR :) This operator's quite interesting - would you mind if I ask usecases you'd leverage via this operator? I'd like to have bit more background if I can.

blesh:
> would you mind if I ask usecases you'd leverage via this operator?

@kwonoj ... if I were a guessing man, I'd say that this operator aims to support Redux-like behaviors.

However it seems like reducer and sub-reducer patterns recommended by @danabramov with the existing `scan` and a `switch` statement can accomplish the same thing.

@jwoodhull, can you show how this is an improvement over a `scan` with a `switch` statement? The one thing I can see is it allows the developer to "build" the switch statement external to the implementation programmatically, which is sort of cool, but still doable with `scan`.

The only concern I have with it, besides operator bloat, would be the additional overhead with try/catch protection around another user-provided "select" function used to select the key from the object in the second argument.

If the community determines this is a valuable operator and is worth supporting, I don't have a problem with adding it to KitchenSink and forgoing adding it to core operators.

cc/ @robwormald @staltz

jwoodhull:
@blesh  is correct.  This operator is meant to be used as a reducer within a Flux store.

```
const strategy = {};
// store loader
strategy['load'] = (accum, action) => action.payload;

// store updater
strategy['update'] = (accum, action) => {
  accum[action.payload.index] = action.payload.newValue;
  return accum;
};

const store = Rx.Observable.merge(
  Dispatcher.getActions('load'),
  Dispatcher.getActions('update'))
  .scanCase(action => action.name, strategy, {})
```
Yes this same pattern could be implemented within a scan accumulator with a switch statement.  However, I think the scan and switch is a pattern that would be implemented repeatedly in a flux application.  Repeated patterns should be abstracted out.  Also, by using a strategy pattern instead of a switch statement, the code can directly access the needed accumulator, instead of testing each possibility in a switch statement.

robwormald:
Interesting idea. I like it, I think. 

Thoughts:
- the key-based scan object seems like the only reasonable way to describe this case, but I don't know how well it fits into the rest of Rx operators. @blesh?
- I wonder if this is the kind of thing we should really be using let/composition for, or if it's a user land thing. Feels rather specific (though I'd totally use it, personally)
- once operators can be patched after the fact with TS this kind of thing is easy to do in userland. So I dunno. 
- perf? Probably not a whole lot of difference really, but switches are fast. 
- I'm not super stoked on the name (this is RxJS after all, wouldn't be an issue without some bike shedding)

trxcllnt:
Can't this be implemented with `scan` already? Am I missing something?

```es6
const strategies = {
  add(memo, y) { return memo + y; }
  subtract(memo, y) { return memo - y; }
};

Observable
  .of({ type: 'add', value: 10}, { type: 'subtract', value: 5})
  .scan((memo, e) => {
    return strategies[e.type](memo, e.value);
  }, 0);
```

blesh:
@trxcllnt ... this is kind of my thought too. At first I thought, well, it eliminates a closure, but you can still do that by ferrying the strategies around on the accumulator if you choose, and honestly once the code is in user-land, closure is probably not the biggest perf hit.

@robwormald brings up a good point about perf. Also, I think Redux, @gaearon, et al. Have come up with solid reducer patterns that can be used in an `Rx` scan that would perform really well.

I guess it's a "pass" from me, @jwoodhull. It's a very interesting idea, and I hope that this discussion doesn't discourage you from contributing in the future. I know what it's like to put an idea out there that you're excited about and have people pass on it. It can be a bummer.

I'll of course leave this open wait to read what the rest of the community has to say. Maybe I'm underestimating the desire here.

robwormald:
Once Typescript 1.8 lands with operator patching, it would be cool to see a whole collection of these types of operators on NPM as modules one could easily plug in. @jwoodhull do feel free to ping me if you want some help publishing this kind of thing to npm.

blesh:
@jwoodhull ... I really hope this hasn't put you off from contributing. It's this sort of inventive thinking that really helps push libraries like RxJS to the next level. I'm going to close this for now, as it's become stale, but as @robwormald suggested, it would be cool to see this as it's own npm module that people could add if they wanted your operator.

Again, I sincerely hope this hasn't left you with a foul taste for our community, and I truly hope to see you contribute more. Thanks again!

Issue #1257, opened 2016-01-27
groupby test failures on some browsers


kwonoj:
Browser testing shows some of browser does have unexpected behavior on [test case] 
(https://github.com/ReactiveX/RxJS/blob/master/spec/operators/groupBy-spec.js#L7) , like below:

```
Mobile Safari 7.0.0 (iOS 7.1.0) Observable.prototype.groupBy() should group numbers by odd/even FAILED
	Expected 
	{"frame":20,"notification":{"kind":"N","value":"[
		{"frame":0,"notification":{"kind":"N","value":"1","hasValue":true}},
		{"frame":80,"notification":{"kind":"N","value":"3","hasValue":true}},
		{"frame":160,"notification":{"kind":"N","value":"5","hasValue":true}},
		{"frame":200,"notification":{"kind":"C","hasValue":false}}
	]","hasValue":true}}
	{"frame":60,"notification":{"kind":"N","value":"[
		{"frame":0,"notification":{"kind":"N","value":"2","hasValue":true}},
		{"frame":80,"notification":{"kind":"N","value":"4","hasValue":true}}
	]","hasValue":true}}
	{"frame":220,"notification":{"kind":"C","hasValue":false}}
	
	to deep equal 
	{"frame":20,"notification":{"kind":"N","value":"[
		{"frame":0,"notification":{"kind":"N","value":"1","hasValue":true}},
		{"frame":80,"notification":{"kind":"N","value":"3","hasValue":true}},
		{"frame":160,"notification":{"kind":"N","value":"5","hasValue":true}},
		{"frame":200,"notification":{"kind":"C","hasValue":false}}
	]","hasValue":true}}
	{"frame":60,"notification":{"kind":"N","value":"[
		{"frame":0,"notification":{"kind":"N","value":"2","hasValue":true}},
		{"frame":80,"notification":{"kind":"N","value":"4","hasValue":true}},
		{"frame":160,"notification":{"kind":"C","hasValue":false}}
	]","hasValue":true}}
	{"frame":220,"notification":{"kind":"C","hasValue":false}}
```

occurs on < android 5, < IE 11, ios 7.

staltz:
Is this the only test that fails or are there many?

kwonoj:
[...]

PR #1271, opened 2016-01-31
docs(CONTRIBUTING): Updated the test section to include Travis and saucelabs

jadbox:
In reference to @staltz comment on #1257 to adding saucelabs information to the Travis configuration.
>> Ah neat. We need those instructions in CONTRIBUTING too

Btw, this is my first pull request to Rx: hoping I followed the right procedure here.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -121,6 +121,12 @@ then it must cover the following cases:
 - Success with the context, if any allowed in the operator signature
 - If an error is thrown
 
+#### <a id="unit-tests"></a>CI Tests
+- Using [Travis](https://travis-ci.org/) on your forked version of RxJS will allow running CI tests on that fork before submitting a PR to master
+ - Simply create a `Travis` account and add your fork as a new project
+- [Sauce Labs](https://saucelabs.com/) setup will allow performing automated browser tests on the fork. Since `saucelabs` doesn't perform browser tests on a PR, this will help verify test results before PR's are checked into master.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -121,6 +121,9 @@ then it must cover the following cases:
 - Success with the context, if any allowed in the operator signature
 - If an error is thrown
 
+#### <a id="unit-tests"></a>CI Tests
+- Set your [Travis](https://travis-ci.org/) account to point against your fork
```
kwonoj: what about adding some more background explanation for these tests? below's few topics I can think of : 
- `travis` setup on forked repo will allow ci tests on forked repo without submitting PR to master
- `saucalabs` setup will allow run browser test on forked repo: due to saucelab doesn't allow browser test on PR, this will help verify test results before PR's checked in master
-  since master runs both of test per each check in, it'd be welcome to setup those test for creating PR

blesh: Good thought, @kwonoj. What do you think @jadbox?

jadbox: Sounds great to me: updating PR now

kwonoj:
Thanks for update! would you mind one small favor to flatten commit into single one? I think these change can be single commit, like

`docs(CONTRIBUTING): update .....`

I can handle those so feel freely ping me if you have difficulties with it, but would like to let you have full control over your contribution :)

jadbox:
Thanks @kwonoj, how does this look? (squash merged)

kwonoj:
Thanks for update, yes, change looks good to me.

kwonoj:
Merged with 142d99a, appreciate for effort @jadbox and hope to see further contributions! :)

PR #1273, opened 2016-01-31
perf(single): remove tryCatch/errorObject (~2.5x improvement)

luisgabriel:
Before:
```
                             |      RxJS 4.0.7 |  RxJS 5.0.0-beta.1 | factor | % improved
------------------------------------------------------------------------------------------
single-predicate - immediate | 26,054 (±0.55%) |    86,358 (±0.20%) |  3.31x |     231.5%
          single - immediate | 40,594 (±0.22%) | 1,068,278 (±0.22%) | 26.32x |   2,531.6%
```

After:
```
                             |      RxJS 4.0.7 |  RxJS 5.0.0-beta.1 | factor | % improved
------------------------------------------------------------------------------------------
single-predicate - immediate | 26,850 (±0.47%) |   222,364 (±0.39%) |  8.28x |     728.2%
          single - immediate | 40,725 (±0.75%) | 1,105,053 (±1.32%) | 27.13x |   2,613.5%
```


src/operator/single.ts
```diff
@@ -55,20 +52,25 @@ class SingleSubscriber<T> extends Subscriber<T> {
...
-      } else if (result) {
-        this.applySingleValue(value);
-      }
+      this.__tryNext(value);
     } else {
       this.applySingleValue(value);
     }
   }
 
+  private __tryNext(value: T): void {
```
kwonoj: do we need double underbar (_ _) here while there isn't worry to name conflict? Some of existing codebase does have it but it seems to avoid name conflict (i.e, _error and __error).

luisgabriel: No actually. I confess that I'm always a bit confused about the use of underbar across the codebase. Not sure why I used it here. I'll rename it to `tryNext()` and update this PR. ;)

kwonoj: I think single underbar would work for consistency of this changes, other `tryxxx` does that. I agree using underbar in codebase sometimes not clear - I generally try to avoid to use it. In cases like conflict such as `public next(): void` and `protected _next(): void`, each encapsulated function would start with underbar, but never for public interfaces.

luisgabriel:
@kwonoj the PR is now updated.

kwonoj:
Change looks good to me.

kwonoj:
Merged with 2515cfb, thanks @luisgabriel

PR #1276, opened 2016-01-31
fix(typings): don't expose PromiseConstructor dependency

### LINKED ISSUE #1270: TS project cannot depend only on es6-promise
Given an angular2 project, relying on rxjs types available from node_modules/rxjs and only user-installed es6 typings (working on https://github.com/angular/angular/issues/5807 )
If we ask users to eg.
`$ typings install es6-collections es6-promise --ambient --save`
Then we are missing `PromiseConstructor`:
```
node_modules/rxjs/CoreOperators.d.ts(77,31): error TS2304: Cannot find name 'PromiseConstructor'.
node_modules/rxjs/Observable.d.ts(65,67): error TS2304: Cannot find name 'PromiseConstructor'.
node_modules/rxjs/Observable.d.ts(159,31): error TS2304: Cannot find name 'PromiseConstructor'.
```

This type is defined in typescript's distributed `lib.es6.d.ts` and also in `es6-shim.d.ts` but not in `es6-promise.d.ts`

Workaround for now is we require users install typings for all of `es6-shim` but this has the disadvantage that they may type-check usage of all es6 features, even those not polyfilled at runtime.

alexeagle:
Fixes #1270 

(Note: in angular we have a "test.typings" task, which takes the dist folder and ensures the user can type-check against it given pre-requisite dependencies. That would prevent this problem from returning)

### DIFF
--- a/src/CoreOperators.ts
+++ b/src/CoreOperators.ts
@@ -84,7 +84,7 @@ export interface CoreOperators<T> {
   timeout?: (due: number | Date, errorToSend?: any, scheduler?: Scheduler) => Observable<T>;
   timeoutWith?: <R>(due: number | Date, withObservable: Observable<R>, scheduler?: Scheduler) => Observable<T> | Observable<R>;
   toArray?: () => Observable<T[]>;
-  toPromise?: (PromiseCtor: PromiseConstructor) => Promise<T>;
+  toPromise?: (PromiseCtor: typeof Promise) => Promise<T>;
   window?: (closingNotifier: Observable<any>) => Observable<Observable<T>>;
   windowCount?: (windowSize: number, startWindowEvery: number) => Observable<Observable<T>>;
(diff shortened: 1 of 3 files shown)


kwonoj:
Interesting PR, thanks for this. So does `PromiseCtor` is equivalent to `typeof Promise`?

alexeagle:
Yes, simply, because in `lib.es6.d.ts`
```
declare var Promise: PromiseConstructor;
```

It's a tricky distinction.
`PromiseConstructor` forces a particular constructor signature, static members, etc.
`typeof Promise` just means a factory of Promise instances.

kwonoj:
Thanks @alexeagle , merged with f59225b. I've amended commit message from `Fixed #` to `closes #` to met with current convention we're using for automate changelogs, wish you don't mind :) Thanks again!

alexeagle:
@blesh we need to hold the next angular2 release until this is released in rxjs (so that our quickstart guide doesn't need a temporary "install es6-shim.d.ts" step)
Any idea when the next release will land? thx!

PR #1282, opened 2016-02-02
fix(Observer): observer interface now allows partial Observers

blesh:
- caveat is that there is no way to require \_at least one\_ method on Observer in TypeScript, at least that I can find

related #1260

### DIFF
--- a/src/Notification.ts
+++ b/src/Notification.ts
@@ -1,4 +1,4 @@
-import {Observer} from './Observer';
+import {PartialObserver} from './Observer';
 import {Observable} from './Observable';
 
 export class Notification<T> {
@@ -8,14 +8,14 @@ export class Notification<T> {
     this.hasValue = kind === 'N';
   }
 
-  observe(observer: Observer<T>): any {
+  observe(observer: PartialObserver<T>): any {
     switch (this.kind) {
       case 'N':
-        return observer.next(this.value);
+        return observer.next && observer.next(this.value);
       case 'E':
-        return observer.error(this.exception);
+        return observer.error && observer.error(this.exception);
       case 'C':
-        return observer.complete();
+        return observer.complete && observer.complete();
     }
   }
 
@@ -23,17 +23,17 @@ export class Notification<T> {
     const kind = this.kind;
     switch (kind) {
       case 'N':
-        return next(this.value);
+        return next && next(this.value);
       case 'E':
-        return error(this.exception);
+        return error && error(this.exception);
       case 'C':
-        return complete();
+        return complete && complete();
     }
   }
 
-  accept(nextOrObserver: Observer<T> | ((value: T) => void), error?: (err: any) => void, complete?: () => void) {
-    if (nextOrObserver && typeof (<Observer<T>>nextOrObserver).next === 'function') {
-      return this.observe(<Observer<T>>nextOrObserver);
+  accept(nextOrObserver: PartialObserver<T> | ((value: T) => void), error?: (err: any) => void, complete?: () => void) {
+    if (nextOrObserver && typeof (<PartialObserver<T>>nextOrObserver).next === 'function') {
+      return this.observe(<PartialObserver<T>>nextOrObserver);
     } else {
       return this.do(<(value: T) => void>nextOrObserver, error, complete);
     }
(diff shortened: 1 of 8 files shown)


kwonoj:
Looks good to me for now. I'll track how typescript side changes / updates are going to followup further.

kwonoj:
Seems with current Typescript it can be achieved via below, not looking elegant bit though..

`type Observer<T> = NextObserver<T> | ErrorObserver | CompleteObserver | (NextObserver<T> & ErrorObserver & CompleteObserver) | (ErrorObserver & CompleteObserver) | (NextObserver<T> & ErrorObserver) | (NextObserver<T> & CompleteObserver);`

(copy-pasted from issue).

blesh:
I experimented for a while with different approaches and this was the best I could come up with for developer ergonomics.

I tried this also: 

```TypeScript

export interface NextObserver<T> {
  isUnsubscribed?: boolean;
  next: (value: T) => void;
}

export interface ErrorObserver {
  isUnsubscribed?: boolean;
  error: (err: any) => void;
}

export interface CompletionObserver {
  isUnsubscribed?: boolean;
  complete: () => void;
}

export declare type Observer<T> = NextObserver<T> | ErrorObserver | CompletionObserver;
```

But that resulted in ugly experiences on the consumer end, where trying to call `next` on the Observer gave you TypeScript errors because "next" doesn't exist on " NextObserver<T> | ErrorObserver | CompletionObserver"

blesh:
@kwonoj ... haha you typed that while I was adding additional comments. Yes, I did try that, but it ended up being ugly on the side of whatever _consumes_ the observer. For example and error like: `o.next does not exist on NextObserver<T> | ErrorObserver | CompleteObserver | (NextObserver<T> & ErrorObserver & CompleteObserver) | (ErrorObserver & CompleteObserver) | (NextObserver<T> & ErrorObserver) | (NextObserver<T> & CompleteObserver)`;

kwonoj:
Yes, it'll be quite unfriendly from consumer perspective. For now I'd rather allow observer to be fully optional interfaces given that most cases caller wouldn't pass empty object ({}) in usual cases. Course this is based on assumption there'll be better ergonomics support from typescript will be available. I hope https://github.com/Microsoft/TypeScript/issues/4889 can be answer.

blesh:
@kwonoj ... I got a solid answer from the TypeScript folks and reimplemented my solution. Now there is a `PartialObserver<T>` type that matches any partial observer. I've also updated all appropriate locations in code that could have partial observers.

kwonoj:
Saw update from issue and this looks much better :) Change looks good to me. 

p.s: just curious, how does error output now when partial observer does not have some interfaces? (as compare to above)

blesh:
@kwonoj if you try to pass an empty object, for example to something that accepts `PartialObserver<T>`, it complains that `{} does not match NextObserver<T> | ErrorObserver | CompletionObserver`... likewise if there's too many additional methods: `{ next() {}, blah () {} }`.

kwonoj:
Thanks for clarification :)

src/Notification.ts
```diff
@@ -23,17 +23,17 @@ export class Notification<T> {
...
-        return complete();
+        return complete && complete();
     }
   }
 
-  accept(nextOrObserver: Observer<T> | ((value: T) => void), error?: (err: any) => void, complete?: () => void) {
-    if (nextOrObserver && typeof (<Observer<T>>nextOrObserver).next === 'function') {
-      return this.observe(<Observer<T>>nextOrObserver);
+  accept(nextOrObserver: PartialObserver<T> | ((value: T) => void), error?: (err: any) => void, complete?: () => void) {
+    if (nextOrObserver && typeof (<PartialObserver<T>>nextOrObserver).next === 'function') {
```
saneyuki: By the way, I think this guard will accepts `IterableIterator<T>` (from the code which have only a runtime dynamically type). In today's RxJS context,  can a generator be a `Observer<T>`?

blesh: @saneyuki ... not really... generators have `next`, `throw` and `return` methods. Very, very early versions of this library where implemented in that manner, but it was abandoned when the es-observable spec moved away from that.

Issue #1193, opened 2016-01-14
Cannot opt-in import operator's Type Definitions from 'rxjs/add/operator/***'


saneyuki:
At `5.0.0-beta.1`, We can success to compile the following code.
However, this code should not success the compilation because the code does not import `rxjs/add/operator/map`.

```typescript
import {Observable} from 'rxjs';

// The compilation should not success
// because this code does not import 'rxjs/add/operator/map'.
Observable.timer(1).map(() => {}).subscribe(() => {});
```
I'm not sure this problem is a higher priority issue, but I think this may be a problem.

@kwonoj @david-driscoll  Do you recognize this problem?

saneyuki:
> Do you recognize this problem?

Ah, My intention was "Have you already know this problem?"

kwonoj:
That's expected currently, by the way it works. So answer is yes, I do aware of it.

Solving this is not trivial though. What we want (and do) is allowing import operator module dynamically, and for this we also want to bring type definition to be dynamically provided per import, while type definition is typically expected to be published statically. It's also somewhat related with operator import itself, if you look into `add/*`, you can see it's using prototype patching at runtime. Means at compile time, TypeScript does not able to do much. I was once thought about using mixin / ambient class merging for modular operator supports those behavior, but that also had couple of drawbacks at the moment.

kwonoj:
You may refer issue at https://github.com/ReactiveX/RxJS/issues/642, not directly related but somewhat related. It's also hit by statically published type information, so extended operator definition's not supported at the moment.

kwonoj:
Just adding few additional details from TypeScript side support so far I know
- Extension method : on discussion
- traits : probably won't
- ES7 bind operator (`::`) : on discussion

Those are mechanics I was considered as possible solutions related to some of type / modularity issues but all of them are not in our hand at this moment. It's still possible there's elegant way that I haven't thought about.

kwonoj:
Noticed this https://github.com/Microsoft/TypeScript/pull/6213 too, might be feasible solutions..

blesh:
So to clarify this issue, the issue is that TypeScript thinks it's okay to use `map` even if map isn't there because we didn't import it?

kwonoj:
Yes, that's correct by `Observable` already have all type defintions of `CoreOperator` regardless of importing actual operator or not.

kwonoj:
I'm assigning myself for evaluating once typescript side changes arrives.

[...]

PR #1284, opened 2016-02-02
WAITING FOR TSC 1.8: refactor(patches): move operator stubs from Observable.ts to patch files

masaeedu:
Use new external module augmentation feature to move Observable stubs into patch files. See #1193


kwonoj:
Appreciate for taking effort for this!

I tried this PR and found somewhat unexpected behavior, while consumption of augmented module works editor supports for typing isn't as below,

image

image

none of type definitions are now being supported by editors. 

This change solves build time detection of what issue #1193 points but this seems quite regression that prevents daily usage of type definitions. I assume current editor's TypeScript engines are yet able to support those.

kwonoj:
For checking in this changes, I came to bit hesitate by this change loses existing availability of typings - What do you think? Is there'll be immediate resolution for this? /cc @saneyuki, @blesh also for visibility.

masaeedu:
@kwonoj Are you importing the appropriate operators?

kwonoj:
@masaeedu That's occurring even importing full set, `import * as Rx from..`. Since full set imports every operator, I'd expect it work out of box as same as before.

kwonoj:
Just clarify, I'm currently using VS code. Curious if other editor / IDE might have same behavior.

masaeedu:
@kwonoj Weird. Trying to setup a test now. Just to be sure, have you set your `"typescript.tsdk"` setting at the user or workspace level to the latest typescript?

kwonoj:
> have you set your "typescript.tsdk" setting at the user or workspace level to the latest typescript?

: No, I'm using out of box configuration. Maybe that's reason?

masaeedu:
@kwonoj Ok, so that's probably one thing to fix. I got some great help with how to do this in Microsoft/vscode#1771. However, I already have this set up and can reproduce the issue you're seeing when I `import * as Rx from 'rxjs/Rx'`.

The issue is that because we're not re-exporting the contents of the patch files in the `Rx.ts` file, all the `import './add/operator/foo'` lines have no effect on the generated `.d.ts` file. What we want is to do `export * from './add/operator/foo'`. I've tried changing a few of the imports to re-exports and it seems to work.

kwonoj:
> What we want is to do export * from './add/operator/foo'

: What happens if import operators separately?

masaeedu:
@kwonoj You mean if I don't use the `rxjs/Rx` module and import directly from `add/operator/foo`? That scenario works fine for me right now.

kwonoj:
>  if I don't use the rxjs/Rx module and import directly from add/operator/foo?

: Yes, that was the case.

kwonoj:
When I tried both (importing Rx / importing operator individually) were not able to provide type definitions.

masaeedu:
So this is the scenario I'm trying right now:

```
// @filename:main.ts
import {Observable} from 'rxjs/Observable';
import 'rxjs/add/operator/buffer';

var obs = Observable.of(1, 2, 3);
obs.bu // I press Ctrl+Space to get autocomplete options here
```

The autocomplete options I get look like this:

Image of Yaktocat

(please ignore tslint squigglies :smile:). Once I change the corresponding `import`s in `Rx.ts` to `export * from`s, they start working the same way.

masaeedu:
@kwonoj Hopefully this is what the expected behavior is?

kwonoj:
Cool, it's lovely. We still need to test out couple of more editor / IDE to ensure this works though. (anyone uses other editor than VS code?)

kwonoj:
> We still need to test out couple of more editor / IDE to ensure this works though. 

: I think it'd be ok as long as other editor's not regressed. losing all type intellisense is case I'd like to avoid.

masaeedu:
@kwonoj I can test out Atom. There's no rush to get this merged, I don't think rebasing this should be too difficult because there's not too many people fooling around with the patch files.

masaeedu:
@kwonoj Also, now that I've fixed the `Rx.ts` file, the scenario you were talking about works :smile: :

image demonstrating how Rx.ts import works

kwonoj:
Thanks, let's test out couple of more and rebase. Might need see how does reexport looks like.

kwonoj:
> the scenario you were talking about works

: Cool, one step forward for better typing experiences :)

masaeedu:
I did however have to nuke the `export var _void: void;` in the patch files to make the re-exports work. Do you know if this is still needed?

kwonoj:
> this is still needed

Nope. That was introduced due to https://github.com/ReactiveX/RxJS/issues/1010, patch operator type definition becomes empty since there's nothing exported. If there's real export, that doesn't necessarily required anymore.

kwonoj:
What about `Rx.KitchenSink`? doesn't it need to be updated as well?

src/add/observable/combineLatest.ts
```diff
@@ -1,6 +1,4 @@
 import {Observable} from '../../Observable';
 import {combineLatestStatic} from '../../operator/combineLatest';
 
-Observable.combineLatest = combineLatestStatic;
-
-export var _void: void;
```
kwonoj: isn't this still needed? does typings contains exports after removing this?

masaeedu:
@kwonoj I guess so. I wasn't actually aware of the existence of `Rx.KitchenSink`. Is the goal of that file to export everything in `Rx` plus some extras? If so it would probably be a good idea to only do `export * from './Rx'` and then export the extras, instead of duplicating all of them.

src/add/observable/combineLatest.ts
masaeedu: I'm not sure what you mean by that. This file is still considered an external module after removal of `export var _void` if that's what you mean, because it has top level `import`s.

src/add/operator/buffer.ts
```diff
@@ -4,4 +4,8 @@ import {buffer} from '../../operator/buffer';
 
 Observable.prototype.buffer = buffer;
 
-export var _void: void;
\ No newline at end of file
+declare module '../../Observable' {
+  interface Observable<T> {
+    buffer: (closingNotifier: Observable<any>) => Observable<T[]>;
+  }
+}
```
luisgabriel: This may be silly but what do you think about augmenting the module before the prototype assignment?

masaeedu: @luisgabriel I don't mind either approach. Is there a reason you prefer that style?

luisgabriel: It makes more sense to me to declare an interface first and then implement it. But this is not a big deal. I'm just commenting to see if this makes sense to someone else other than me. ;)

src/add/operator/buffer.ts
```diff
@@ -4,4 +4,8 @@ import {buffer} from '../../operator/buffer';
 
 Observable.prototype.buffer = buffer;
 
-export var _void: void;
\ No newline at end of file
+declare module '../../Observable' {
+  interface Observable<T> {
+    buffer: (closingNotifier: Observable<any>) => Observable<T[]>;
```
david-driscoll: Could this be `buffer: typeof buffer`?

Then this will change as `buffer` changes.

src/add/observable/combineLatest.ts
kwonoj: Refer https://github.com/ReactiveX/RxJS/issues/1010 I mentioned. If there isn't any export makes d.ts to be empty, it makes consumer in trouble. So question's is there anything exported after removing this and type definition's published. I assume operator includes augmentation does.

kwonoj:
> Is the goal of that file to export everything in Rx plus some extras? 

: Yes, it have few additional operators in additions to Rx.

> If so it would probably be a good idea to only do export * from './Rx' and then export the extras, instead of duplicating all of them.

: if that works I think that'd better. I tried those while ago but wasn't able to make it.

src/add/operator/buffer.ts
kwonoj: That's be neat if possible.

src/add/operator/buffer.ts
kwonoj: I'd do as @luisgabriel said when I gave attempt. Honestly I think it's minor and does not impact to anything though.

src/add/observable/combineLatest.ts
masaeedu: @kwonoj Yes, I ran into that issue. Discussing with the TypeScript folks in Microsoft/TypeScript#6022

kwonoj: For our side of interim, we could leave this as same as before, right?

src/add/operator/buffer.ts
masaeedu: I was planning to add interface exports to the operator files, so that the signature would be:

```
declare module '../../Observable' {
  interface Observable<T> {
    buffer: buffer<T>;
  }
}
```

The problem with doing just `typeof buffer` I think will be that the output type for consumers will be `Observable<{}>` instead of `Observable<T>`.

kwonoj: That'll work too, both will achieve goals of repeating interfaces on different files.

david-driscoll: fair enough, then when we finish off things related to #1086  (which I will rebase this afternoon)

src/add/observable/combineLatest.ts
masaeedu: Yes, that's right. I'll revert the `_void: void` removals and re-exports for the `observable` patches.

kwonoj: Let's go that way. I don't think this need to block this whole PR. (export void if empty, remove if it's unnecessary)

src/add/operator/buffer.ts
masaeedu: Yeah, exactly. For now, is it okay to merge this as is, then start referencing David's interfaces as he adds them?

kwonoj: David's changes orthogonal to this so far I remember. You can apply those pattern without checking in his.

masaeedu: @kwonoj From what I understand, using `buffer: typeof buffer` is not equivalent to using `buffer: (closingNotifier: Observable<any>) => Observable<T[]>;` or `buffer: BufferSignature<T>;`. If we go with `buffer: typeof buffer`, information about `T`, or the type of items that the source produces, is lost. I put together an image to illustrate:

left is with the current signature (or `interface<T>`), right is with typeof

Left is with the current signature (or `interface<T>`), right is with typeof

kwonoj: It'd be great if it works, but in that case let's go with as-is and try to find better way in further.

david-driscoll:
Atom updates fairly regularly against typescript itself, so the definitions do propagate properly it looks like.

image

image

src/add/operator/buffer.ts
david-driscoll: @masaeedu you're right, because of `buffer` expects there to be a typed this context, that is `T`.  So we'd have to add `BufferSiganture<T>` once `BufferSiganture<T>` is added.  Definitely not blocking.

david-driscoll:
Does anyone know of a way that this can be applied to static methods as well?  Users will still sett `Observable.combineLatest` even if they haven't imported `rxjs/add/observable/combineLatest`.

kwonoj:
> that this can be applied to static methods as well

: So far I understand, augmentation's instance for class / interface and not applicable for static.

david-driscoll:
I know, just a random item for thought.  Not meant as anything to block, just maybe someone knew something. :)

kwonoj:
> , just maybe someone knew something

: Yup, sorry for repeating already-known-to-unknown :) Once this changes are checked in, let's track those as separate issue if needed.

blesh:
Just marking this as blocked and discussion until it's worked out.

Also it needs to be rebased.

masaeedu:
@david-driscoll I think namespace merging might work for the static methods.

masaeedu:
@david-driscoll Yup, works great:

```
import {Observable} from '../../../Observable';
import {WebSocketSubject} from '../../../observable/dom/WebSocketSubject';

Observable.webSocket = WebSocketSubject.create;

declare module '../../../Observable' {
  namespace Observable {
    export let webSocket: typeof WebSocketSubject.create;
  }
}
```

kwonoj:
Let's make static method as separate discussion. This PR itself has quite amount of changes.

david-driscoll:
That's fine, @masaeedu you want to make a separate PR or should I?  (I don't care)

src/add/observable/combineLatest.ts
masaeedu: I'll be reverting these to help the PR go through, but the good news is that empty .d.ts files won't be a problem for the consumer starting in tomorrow's builds (in other words we don't need the `_void: void` anymore).

kwonoj: That's cool news, but it's starting from tomorrow's nightly version, right? thinking about discussion for min version support at https://github.com/ReactiveX/RxJS/pull/1288#issuecomment-178859936, that might need to be hold off for moment :(

masaeedu: @kwonoj Yeah, unfortunately it looks like we'll have to let this and #1288 stew for a while. :disappointed:

robwormald:
same as #1288, i'm definitely for this change but it needs to wait until the functionality is in the released TS compiler.

masaeedu:
Now tracked by #1388

PR #1286, opened 2016-02-02
refactor(throttle): remove tryCatch/errorObject

saneyuki:
- This change is same kind of #1251.
- I have not benchmarked this yet. So this is purely refactoring to unify code style.
  - The current codebase does not have any benchmark about `throttle()`.
  - `throttle()` suppress the value traffic. How do we benchmark this without a throughput benchmark?

### DIFF
--- a/perf/micro/current-thread-scheduler/operators/throttle.js
+++ b/perf/micro/current-thread-scheduler/operators/throttle.js
@@ -1,0 +1,22 @@
+var RxOld = require('rx');
+var RxNew = require('../../../../index');
+
+module.exports = function (suite) {
+  // v4's `throttle()` does not accept a duration selector.
+  // So this test is just a sample for each revisions of RxNew.
+  var oldThrottleWithImmediateScheduler = RxOld.Observable.range(0, 25, RxOld.Scheduler.currentThread)
+    .throttle(1);
+  var newThrottleWithImmediateScheduler = RxNew.Observable.range(0, 25, RxNew.Scheduler.queue)
+    .throttle(function () { return RxNew.Observable.of(0); });
+
+  function _next(x) { }
+  function _error(e) { }
+  function _complete() { }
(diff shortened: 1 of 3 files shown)


blesh:
Can you flatten the first two commits? There's no reason for them to be separate, really. They're just TypeScript fixes.

saneyuki:
I added the micro perf benchmars for `throttle()` ( https://github.com/saneyuki/RxJS/commit/8963abb98f305c39ff9b5d84a1224436f7c01a0d ), but RxJS v4 does not accept `Observable<T>` as a durationSelector (see also #1321). So we should only watch the v5's result.
(Should we really add this benchmark to the tree?)

These are results of benchmarks (this table drops v4's results):

## Before this pull request
```
                                         |             RxJS 5.0.0-beta.1 |
----------------------------------------------------------------------------
                    throttle - immediate |                38,030 (±2.18%) |
                                throttle |                32,392 (±5.58%) |
```

## My first patch
```
                                         |             RxJS 5.0.0-beta.1 |  
----------------------------------------------------------------------------
                    throttle - immediate |                45,254 (±1.21%) |
                                throttle |                33,630 (±3.33%) |
```

## @blesh's proposal
```
                                         |             RxJS 5.0.0-beta.1 |  
----------------------------------------------------------------------------
                    throttle - immediate |                58,980 (±5.69%) |
                                throttle |                41,971 (±5.14%) |
```

By these results, @blesh's suggestion is right!

saneyuki:
@kwonoj 

updated :)
And, I add some comments which explain why we cannot compare with v4's `throttle(durationSelector)`  to https://github.com/saneyuki/RxJS/commit/6060977541430ec3ccd3b30a84ad18af4169e479.

kwonoj:
Merged with 7bae860, thanks @saneyuki !

PR #1288, opened 2016-02-02
WAITING FOR TSC 1.8: Move static observable stubs from Observable.ts to…

david-driscoll:
… patch files

Related to #1284, and I'm fine with rebasing once #1284 gets merged.

cc @masaeedu @kwonoj

### DIFF
--- a/src/add/observable/bindCallback.ts
+++ b/src/add/observable/bindCallback.ts
@@ -3,4 +3,8 @@ import {BoundCallbackObservable} from '../../observable/BoundCallbackObservable'
 
 Observable.bindCallback = BoundCallbackObservable.create;
 
-export var _void: void;
+declare module '../../Observable' {
+  namespace Observable {
+    export let bindCallback: typeof BoundCallbackObservable.create;
+  }
+}
--- a/src/add/observable/bindNodeCallback.ts
+++ b/src/add/observable/bindNodeCallback.ts
@@ -3,4 +3,8 @@ import {BoundNodeCallbackObservable} from '../../observable/BoundNodeCallbackObs
 
 Observable.bindNodeCallback = BoundNodeCallbackObservable.create;
 
-export var _void: void;
+declare module '../../Observable' {
+  namespace Observable {
+    export let bindNodeCallback: typeof BoundNodeCallbackObservable.create;
+  }
+}
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "systemjs": "^0.19.6",
     "systemjs-builder": "^0.10.6",
     "tslint": "3.3.0-dev.2",
-    "typescript": "1.9.0-dev.20160128",
+    "typescript": "1.9.0-dev.20160202",
     "validate-commit-msg": "1.1.1",
     "watch": "0.16.0",
     "watchify": "3.7.0",
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -14,28 +14,7 @@ import {toSubscriber} from './util/toSubscriber';
 import {tryCatch} from './util/tryCatch';
 import {errorObject} from './util/errorObject';
 
-import {combineLatestStatic} from './operator/combineLatest';
-import {concatStatic} from './operator/concat';
-import {mergeStatic} from './operator/merge';
-import {zipStatic} from './operator/zip';
-import {BoundCallbackObservable} from './observable/BoundCallbackObservable';
-import {BoundNodeCallbackObservable} from './observable/BoundNodeCallbackObservable';
-import {DeferObservable} from './observable/DeferObservable';
-import {EmptyObservable} from './observable/EmptyObservable';
-import {ForkJoinObservable} from './observable/ForkJoinObservable';
-import {FromObservable} from './observable/FromObservable';
-import {ArrayObservable} from './observable/ArrayObservable';
-import {FromEventObservable} from './observable/FromEventObservable';
-import {FromEventPatternObservable} from './observable/FromEventPatternObservable';
-import {PromiseObservable} from './observable/PromiseObservable';
-import {IntervalObservable} from './observable/IntervalObservable';
-import {TimerObservable} from './observable/TimerObservable';
-import {raceStatic} from './operator/race';
-import {RangeObservable} from './observable/RangeObservable';
-import {NeverObservable} from './observable/NeverObservable';
 import {ErrorObservable} from './observable/ErrorObservable';
-import {AjaxCreationMethod} from './observable/dom/AjaxObservable';
-import {WebSocketSubject} from './observable/dom/WebSocketSubject';
 
 import {CombineLatestSignature} from './operator/combineLatest';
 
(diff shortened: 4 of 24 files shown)


kwonoj:
:+1: Thanks for PR, let's take #1284 first, hope you do not mind. (I know you already mentioned it's ok ;) )

src/add/observable/bindNodeCallback.ts
```diff
@@ -3,4 +3,8 @@ import {BoundNodeCallbackObservable} from '../../observable/BoundNodeCallbackObs
 
 Observable.bindNodeCallback = BoundNodeCallbackObservable.create;
 
-export var _void: void;
\ No newline at end of file
+declare module '../../Observable' {
+  module Observable {
+    export let bindNodeCallback: typeof BoundNodeCallbackObservable.create;
```
kwonoj: Just out of curiosity - can `let` be `const`?

david-driscoll: It appears the compiler will then refuse to let you assign the value above.

kwonoj: hmm.. interesting, thanks. Not a major issue, so think it'd be ok.

kwonoj:
cc @blesh also for visibility, @blesh , this PR changes organizations of static method as below 

1. now `Observable.ts` does not include static method stubs by default
2. as same as patch operator, importing patch will dynamically import static method as well as its types
3. No changes for Rx / Rx.KitchenSink

consumer perspective 2 is major change, now user who imports `Observable` only for partial operator import would need to import static method as well with this PR. Question would be if we want this or not. In terms of 'everything is modular' it's somewhat expected way though.

david-driscoll:
One thing to note, today they still have to import the static method, but a TypeScript consumer will see that the method "exists" then when they try to use it they'll get a method not found error.

This change makes it obvious that method doesn't exist and that they need to import it.

kwonoj:
Yes, my terms are not precise. @david-driscoll 's correct.

blesh:
Will this all work correctly if the consumer is using an older version of TypeScript? (say 1.7 or 1.8?)

david-driscoll:
I believe some of the changes are being back ported to 1.8 but I would have to confirm that.  This would also affect #1284.  @masaeedu do you know if this is being back ported to 1.8?  I thought I saw something about that.

kwonoj:
I believe it should work at least on 1.8, for earlier version need to verify it. Primary goal is not to regress and consumer should not care about typescript version dependency. (this, #1284 both)

david-driscoll:
https://github.com/Microsoft/TypeScript/issues/6731 is tagged as version 1.8.2

kwonoj:
Just tried and verified this change won't work with previous version of typescript, i.e 1.7. So if typescript side fix are in 1.8.2, supported min version will be 1.8.2. This feels bit breaking - will there be user want to stay with 1.6, or 1.7?

masaeedu:
This change won't work with TypeScript 1.7.5, no. You need to have a version of the compiler with the external augmentations feature enabled in order for #1284 and this PR to work, so you need to have either the nightlies or 1.8 (which is indeed being back ported to).

masaeedu:
My personal opinion is that there aren't any very serious reasons for a developer to want to stick to TS 1.7, since the [breaking changes between 1.7-1.8](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes) are so few, and the new features so many. That said I appreciate that breaking changes should be taken seriously, and will leave it to the maintainers to decide how to proceed.

kwonoj:
I personally living edge of versions so wouldn't mind to upgrade and would vote for this changes, but due to that reason I'm not able to represent actual consumer perspective. @blesh , what do you think? If version support is concern, I think we can hold off these changes for now since this is not blocking any functional scope of current codebase.

blesh:
TypeScript is currently published at version 1.7 and 1.8 is beta. So we need to support 1.7, unfortunately.

kwonoj:
> published at version 1.7 and 1.8 is beta. So we need to support 1.7

: That seems quite legit, valid point. @masaeedu / @david-driscoll , in that case I'd like to suggest to hold this changes until TypeScript release bumps up to 1.8 to avoid versioning issues - any suggestions around?

blesh:
Let's have the Angular folks chime in on this too, since this is their big ask:  @IgorMinar @jeffbcross @vsavkin @robwormald

robwormald:
definitely :+1:  on this PR in general, but we do need to hold until this functionality is released properly. we can't ask users to install a specific TS edge version (not to mention then having to override their editor to use it as well...)

kwonoj:
I agree it's hard to recommend consumers to tie with bleeding edge version of infrastructure.

Once I got agreement from @david-driscoll and @masaeedu , I'll put on hold both of PR and re raise once TypeScript bumps up its stable version includes this support. Hope it comes soon, I also quite like this changes and it possibly can bring additional benefit to codebases.

david-driscoll:
Oh it's fine to hold it. We can't all ride on the bleeding edge. :+1:

masaeedu:
@kwonoj No objection here, waiting for 1.8 makes sense.

masaeedu:
Note for when we get back to this: Microsoft/TypeScript#6858 fixes module imports so that re-exports aren't needed. TL;DR, all those `import 'foo'` -> `import 'foo'; export * from 'foo';` changes are no longer required.

kwonoj:
Great news :)

src/add/observable/bindCallback.ts
```diff
@@ -3,4 +3,8 @@ import {BoundCallbackObservable} from '../../observable/BoundCallbackObservable'
 
 Observable.bindCallback = BoundCallbackObservable.create;
 
-export var _void: void;
\ No newline at end of file
+declare module '../../Observable' {
+  module Observable {
```
mhegazy: why not `namespace`  instead of `module`?

Reasoning: the key word module in these two lines mean two things, and it can be confusing. namespace would be clearer.

david-driscoll: mainly I'm just used to using module for so long.  I have no issues changing it to `namespace`.

masaeedu: Yes please, if that's not too much trouble. The less overloaded terms I need to disambiguate in my head the better :smile:

masaeedu:
@david-driscoll Since this only works in bleeding edge TypeScript anyways, I think you can also take advantage of Microsoft/Typescript#6858 and get rid of the `export * from`s. Imports are fixed so those are no longer needed.

david-driscoll:
@masaeedu did you pull them out in your branch?  I can do that no problem.

kwonoj:
@masaeedu Is that feature has planned tagged version? I.e 1.8.2 has module aug and if those comes later (1.9 or something else), it might need to be separated.

masaeedu:
@kwonoj Yup, it'll be in 1.8.

david-driscoll:
There are a bunch of changes in https://github.com/Microsoft/TypeScript/commits/release-1.8 (the release tracking branch that they are migrating changes into).

* https://github.com/Microsoft/TypeScript/pull/6846
* https://github.com/Microsoft/TypeScript/issues/6762
* https://github.com/Microsoft/TypeScript/pull/6906

masaeedu:
@david-driscoll Haven't had time, will get to it on the weekend. No rush, right? :cry:

david-driscoll:
@masaeedu just want to make sure we're lined up! :smile: 

Now just watch them release 1.8 tomorrow. :scream:

IgorMinar:
This looks good to me but we need to coordinate merging this with a TS release and an Angular release. We can't ask people to be on unstable release of typescript.

kwonoj:
TS 1.8 has released (https://github.com/Microsoft/TypeScript/releases/tag/v1.8.2). Might need align with ng2 still though.

robwormald:
as of ng2beta7 we're 1.8 friendly I believe.

kwonoj:
> as of ng2beta7 we're 1.8 friendly I believe.

: Lovely :heart:

kwonoj:
@david-driscoll , I'm closing this PR per #1388. Let me know if this is incorrect.

PR #1289, opened 2016-02-02
Ensure KitchenSink.min.js is published, and add gzip testing tool

blesh:
Here's the funniest thing I've read today:

```
Rx.min.js: 34.02kB gzipped
Rx.KitchenSink.min.js: 33.88kB gzipped
```

... that's right, we've stuck so many operators in Rx.ts that gzipped it's actually LARGER than Rx.Kitchensink.ts (presumably because the latter is missing a few operators, or perhaps because it just "gzips better")


kwonoj:
Very interesting. What's size difference between two of them 'before' gzipped?

blesh:
```
440K	dist/global/Rx.KitchenSink.js
236K	dist/global/Rx.KitchenSink.min.js
440K	dist/global/Rx.KitchenSink.umd.js
192K	dist/global/Rx.KitchenSink.umd.min.js
400K	dist/global/Rx.js
216K	dist/global/Rx.min.js
396K	dist/global/Rx.umd.js
180K	dist/global/Rx.umd.min.js
```

blesh:
The umd.min versions are a _LITTLE_ closer, but KitchenSink is still smaller:

```
Rx.umd.min.js: 33.64kB gzipped
Rx.KitchenSink.umd.min.js: 33.40kB gzipped
```

kwonoj:
So there seems some kind of gzip-magic is happening..

blesh:
Also, I noticed that something is blanking out the `KitchenSink.min.js` file in the global output... not sure what's happening there.

blesh:
Just for experimental purposes: Aggressively removing operators, the smallest I got it was 26K gzipped. (JQuery is 36K or something, for reference)

blesh:
oops, clicked the wrong button.

kwonoj:
Kind of off-topic though, noticed 1.8 typescript has concatenate output for AMD / SystemJS (https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#concatenate-amd-and-system-modules-with---outfile) maybe current browserify based can be replaced into this?

PR #1259, opened 2016-01-28
refactor(windowWhen): use subscribeToResult with outerSubscriber

kwonoj:
This PR refactors `windowWhen` operator utilizes `subscribeToResult`, inspired by recent changes of notifyNext (https://github.com/ReactiveX/RxJS/pull/1252) and removes custom inner subscriber.

PR #1291, opened 2016-02-03
refactor(window): use subscribeToResult with outerSubscriber

kwonoj:
This PR's following effort to https://github.com/ReactiveX/RxJS/pull/1259, utilizes `subscribeToResult` where applicable.


kwonoj:
Merged with 2878aed.

PR #1300, opened 2016-02-03
perf(skipWhile): remove tryCatch/errorObject (~1.6x improvement)

luisgabriel:
Before:
```
                      |      RxJS 4.0.7 | RxJS 5.0.0-beta.1 | factor | % improved
----------------------------------------------------------------------------------
skipwhile - immediate | 27,101 (±1.37%) |  122,827 (±0.29%) |  4.53x |     353.2%
            skipwhile |  8,034 (±1.32%) |   82,160 (±0.28%) | 10.23x |     922.7%
```

After:
```
                      |      RxJS 4.0.7 | RxJS 5.0.0-beta.1 | factor | % improved
----------------------------------------------------------------------------------
skipwhile - immediate | 27,040 (±0.30%) |  197,417 (±0.28%) |  7.30x |     630.1%
            skipwhile |  8,842 (±2.08%) |  112,757 (±0.55%) | 12.75x |   1,175.2%
```

### DIFF
--- a/src/operator/skipWhile.ts
+++ b/src/operator/skipWhile.ts
@@ -1,8 +1,6 @@
 import {Observable} from '../Observable';
 import {Operator} from '../Operator';
 import {Subscriber} from '../Subscriber';
-import {tryCatch} from '../util/tryCatch';
-import {errorObject} from '../util/errorObject';
 
 /**
  * Returns an Observable that skips all items emitted by the source Observable as long as a specified condition holds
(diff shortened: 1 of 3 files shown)


src/operator/skipWhile.ts
```diff
@@ -38,17 +36,21 @@ class SkipWhileSubscriber<T> extends Subscriber<T> {
...
-        destination.error(errorObject.e);
-      } else {
-        this.skipping = Boolean(result);
-      }
+    if (this.skipping) {
+      this.tryCallPredicate(value);
     }
-    if (this.skipping === false) {
+
+    if (!this.skipping) {
```
blesh: I don't like that we're keeping `skipping` in state on the instance, here. It would probably be better to return `skipping` or have this logic in another function.

blesh: Actually, you can't really return 'skipping', so it's probably better to either have it in the same function, or move it to it's own function. What do you think?

luisgabriel: Is the problem having the state, or updating it inside `tryCallPredicate` and check for its value on `_next()` right after? If the problem is the second, I agree with you that it's a bit confusing.

What about this?
```typescript
    if (this.skipping) {
      this.skipping = this.tryCallPredicate(value);
    }

    if (!this.skipping) {
      destination.next(value);
    }
  }

  private tryCallPredicate(value: T): boolean {
    let skipping = true;
    try {
      const result = this.predicate(value, this.index++);
      skipping = Boolean(result);
    } catch (err) {
      this.destination.error(err);
    }
    return skipping;
  }
```

blesh: Hmm... let me read the whole impl. I'm probably missing something.

blesh: Yeah... disregard my first comment... I wasn't thinking. The implementation you have is fine. Sorry. haha.

blesh:
Deleting my original critique. It was way off.

This LGTM.

src/operator/skipWhile.ts
luisgabriel: No problem. ;)

kwonoj:
Merged with cf002db, thanks @luisgabriel

PR #1309, opened 2016-02-05
refactor(delayWhen): use subscribeToResult with outerSubscriber

kwonoj:
This PR correctly applies outerSubscriber into delayWhen operator.


kwonoj:
Merged with 621ec68.

Issue #1246, opened 2016-01-25
Need to identify operators that don't support Promises and Should


blesh:
As noted in #1245, there are probably many operators that take functions that return Observables, and should support returning Promises, Arrays, or "observablesque" objects.

I'd greatly appreciate anyone with free time to comb the operators and identify the ones that are missing this feature. Then we can list them out here and create new, separate issues for each one of them.

staltz:
I also noticed this while updating the documentation for operators at reactivex.io. I went through the operators and identified these:

- [ ] buffer (maybe supports Promises, but maybe needs types updated)
- [ ] bufferToggle (!!)
- [ ] bufferWhen (maybe supports Promises, but maybe needs types updated)
- [ ] catch (!!)
- [ ] concat (maybe supports Promises, but maybe needs types updated)
- [x] concatAll
- [ ] concatMap (maybe supports Promises, but needs types updated)
- [ ] concatMapTo (maybe supports Promises, but needs types updated)
- [x] debounce
- [ ] distinct (the flushes parameter)
- [ ] distinctKey (the flushes parameter)
- [x] exhaust
- [ ] exhaustMap (maybe supports Promises, but needs types updated)
- [ ] expand (maybe supports Promises, but needs types updated)
- [ ] merge (maybe supports Promises, but maybe needs types updated)
- [x] mergeAll
- [ ] mergeMap (maybe supports Promises, but needs types updated)
- [ ] mergeMapTo (maybe supports Promises, but needs types updated)
- [ ] mergeScan (maybe supports Promises, but needs types updated)
- [ ] race (maybe supports Promises, but needs types updated)
- [ ] retryWhen (maybe supports Promises, but needs types updated)
- [ ] sample (maybe supports Promises, but needs types updated)
- [ ] skipUntil (maybe supports Promises, but needs types updated)
- [x] switch
- [ ] switchMap (maybe supports Promises, but needs types updated)
- [ ] switchMapTo (maybe supports Promises, but needs types updated)
- [ ] takeUntil (maybe supports Promises, but needs types updated)
- [x] throttle
- [ ] timeoutWith (maybe supports Promises, but needs types updated)
- [ ] window (!!)
- [ ] windowToggle (maybe supports Promises, but needs types updated)
- [ ] windowWhen (maybe supports Promises, but needs types updated)
- [ ] withLatestFrom (maybe supports Promises, but needs types updated)
- [ ] zip (maybe supports Promises, but needs types updated)
[...]

PR #1311, opened 2016-02-05
chore(typings): add typings for buffer and window based methods

david-driscoll:
_will probably need a rebase after #1292_


src/operator/buffer.ts
```diff
@@ -23,6 +23,10 @@ export function buffer<T>(closingNotifier: Observable<any>): Observable<T[]> {
   return this.lift(new BufferOperator<T>(closingNotifier));
 }
 
+export interface BufferSignature<T> {
+  (closingNotifier: Observable<any>): Observable<T[]>;
```
david-driscoll: `closingNotifier` calls through to `subscribeToResult`, so technically this could be `ObservableInput<T>`.  That doesn't really make sense for the more finite values from `Promise<T>`, `Iterable<T>` or `Array<T>`.  So I think the best option is to explicitly leave these as just `Observable<T>`.

kwonoj: I think this is related to https://github.com/ReactiveX/RxJS/issues/1246. Probably we want to support return values other than observable - but I'd like to take care it as separate changes set include type update / test cases to ensure its behavior if necessary.

src/operator/bufferCount.ts
```diff
@@ -19,7 +19,11 @@ import {Observable} from '../Observable';
  * @returns {Observable<T[]>} an Observable of arrays of buffered values.
  */
 export function bufferCount<T>(bufferSize: number, startBufferEvery: number = null): Observable<T[]> {
-  return this.lift(new BufferCountOperator(bufferSize, startBufferEvery));
+  return this.lift(new BufferCountOperator<T>(bufferSize, startBufferEvery));
+}
+
+export interface BufferCountSignature<T> {
```
kwonoj: Compare to previous type definitions having n-number of param (combinelatest, etcs) these kind of operators have single signature only. Is it necessarily required to export signature then import in coreoperators?

david-driscoll: Sadly yes because we can't use typeof bufferCount<T>, because bufferCount is a function that form of casting isn't supported.

kwonoj: OK, got point. Thanks for clarification.

kwonoj:
Change looks good to me, (need to be rebased though)

david-driscoll:
done.

kwonoj:
Merged with 51add30, thanks @david-driscoll .

src/operator/bufferCount.ts
masaeedu: One way to work around extra import is to use interface merging: `export interface bufferCount`. Now no extra `import` is required for `BufferSignature`, `bufferCount` acts as both function and signature. The benefit is marginal however.

kwonoj: Yes, I don't think this is serious issue need to be addressed. Mostly curious question.

PR #1312, opened 2016-02-06
refactor(distinct): use subscribeToResult with outerSubscriber

kwonoj:
Continue effort to utilize subscribeToResult where applicable.


kwonoj:
Merged with 8ee51b0 .

PR #1313, opened 2016-02-07
Adjust description as ES7 features already came out.

trotyl:
The ES7(ES2016) **DO** only has two features: Array.prototype.includes & ** operator.
As [Function Bind Syntax](https://github.com/tc39/ecma262/blob/master/stage0.md) is still a stage 0 proposal, it not proper to regard it as a ES7 feature. (Seems at least ES9).


blesh:
Merged as of c97d5f4... thanks for catching this @trotyl!

PR #1316, opened 2016-02-08
fix(Subject): throw ObjectUnsubscribedError when unsubecribed …

### LINKED ISSUE #859: Unsubscribed Subjects should throw ObjectDisposedError if next, error or complete is called.
In the current version of RxJS v4, [we check for disposal](https://github.com/Reactive-Extensions/RxJS/blob/master/src/modular/subject.js#L27) with [`Disposable.checkDisposed`](https://github.com/Reactive-Extensions/RxJS/blob/master/src/modular/disposable.js#L46) so any subsequent call to `onNext`, `onError` and `onCompleted` and even `subscribe` are guarded by throwing a specific error of `ObjectDisposedError`.

In RxJS v5, there is a change to simply return and do nothing in the [`Subject`](https://github.com/ReactiveX/RxJS/blob/52eea95cd37481718a79267437ae280b1d9e195f/src/Subject.ts#L85), is this intentional and if so why?  Yes, I realize this follows RxJava instead of RxJS, but at least what thought went into this would be good to know. 

cc @benjchristensen @headinthebox 

(EDIT (@blesh): updated link to "modern" equivalent of what _was_ linked, and pinned it to a SHA)

kwonoj:
This PR tries to close #859, now `Subject` throws `ObjectUnsubscribedError` if it's unsubscribed. Updated operator behavior and test cases also to not emit after subject completion(error, complete, unsub)

### DIFF
--- a/spec/Subject-spec.js
+++ b/spec/Subject-spec.js
@@ -129,9 +129,6 @@ describe('Subject', function () {
     subscription1.unsubscribe();
 
     subject.complete();
-    subject.next(9);
-    subject.complete();
-    subject.error(new Error('err'));
 
     subscription2.unsubscribe();
 
@@ -180,9 +177,6 @@ describe('Subject', function () {
     subscription1.unsubscribe();
 
     subject.error(new Error('err'));
-    subject.next(9);
-    subject.complete();
-    subject.error(new Error('err'));
 
     subscription2.unsubscribe();
 
@@ -221,9 +215,6 @@ describe('Subject', function () {
     subscription1.unsubscribe();
 
     subject.complete();
-    subject.next(9);
-    subject.complete();
-    subject.error(new Error('err'));
 
     subscription2.unsubscribe();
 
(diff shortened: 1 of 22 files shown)


spec/Subject-spec.js
```diff
@@ -221,9 +215,6 @@ describe('Subject', function () {
     subscription1.unsubscribe();
 
     subject.complete();
-    subject.next(9);
-    subject.complete();
-    subject.error(new Error('err'));
```
blesh: Are we added tests for expectations of throwing in these deleted scenarios?

kwonoj: Oh, thanks for catching it. Subject's spec's are not updated to include those. Updating it now.

kwonoj:
PR's updated to add test coverage for changed behavior.

blesh:
LGTM... merged as of 64bf6b1. Thanks @kwonoj

PR #1315, opened 2016-02-08
fix(merge/concat): merging or concatenating scalar observables will now com…

blesh:
…plete properly

- Removed a superfluous check for `_isScalar` that was doing the wrong thing and not completing the scalar observable

fixes #1150

PR #1317, opened 2016-02-09
refactor(expand): remove superfluous _isScalar check

blesh:
This was removed because it was redundant and it could cause errors, but I was unable to
produce any, I think because of the nature of the operator.

As mentioned by @trxcllnt in #1315


kwonoj:
Change looks good to me.

PR #1318, opened 2016-02-09
refactor groupBy behaviors for browser compatibility

kwonoj:
This PR fixes compatibility of `groupBy` on some browsers.

Root cause was removing groups for completion (error, complete..) inside of `map::foreach` to clear all groups, doesn't supported by polyfill implementation of map. PR introduces `map::clear()` interface for explicit clearing and utilize it.

cc @staltz for visibility.

### DIFF
--- a/src/operator/groupBy.ts
+++ b/src/operator/groupBy.ts
@@ -126,8 +126,9 @@ class GroupBySubscriber<T, K, R> extends Subscriber<T> implements RefCountSubscr
     if (groups) {
       groups.forEach((group, key) => {
         group.error(err);
-        this.removeGroup(key);
       });
+
+      groups.clear();
     }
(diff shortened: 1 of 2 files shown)


kwonoj:
For some reason travis does not detect this PR, force close-reopened PR to trigger it.

src/operator/groupBy.ts
```diff
@@ -137,8 +138,9 @@ class GroupBySubscriber<T, K, R> extends Subscriber<T> implements RefCountSubscr
     if (groups) {
       groups.forEach((group, key) => {
         group.complete();
-        this.removeGroup(key);
```
staltz: clever fix

kwonoj:
Merged with ace3932.

Issue #839, opened 2015-11-30
cache operator


blesh:
Currently, if you want to cache some results from an Observable, you're forced to use `publishReplay` or a `ReplaySubject` or something to that effect. There are a few problems with this approach:

1. It's not really apparent to users that this is what they need to do.
2. The default behavior of `publishReplay` isn't what most people would want for common scenarios. For example: If there is a error from the source, you'll likely want to `retry` it, but if there's not, you probably don't want to `repeat` it. (This community infamously had a passionate debate on that matter, LOL).
3. It probably doesn't have all of the control necessary to define the various caching needs for each application.

There is a need, overall for a solid caching operator, and perhaps a set of refined versions of it. That makes it simple and easy to do all sorts of caching. The problem is that there is _all sorts of caching_.

@jhusain brought up that he had some ideas around this, and I've talked to @abersnaze and @stealthcode about work they're doing around this same issue in RxJava. @mattpodwysocki and @benjchristensen may also have some thoughts.

### Cache Evictions are complex

@abersnaze brought up the [Google CacheLoader](https://code.google.com/p/guava-libraries/wiki/CachesExplained), which I think is a good resource to understand the types of [evictions](https://code.google.com/p/guava-libraries/wiki/CachesExplained#Eviction) that can occur.

Short list of common cache evictions:

- size-based evictions (only cache the last 10 values)
- evict each value Xms from the time it arrived
- evict all/some/one value(s) at X absolute time
- evict values based off of some analysis of the value as it is being cached
- evict values when no one is subscribing any more.
- evict values on error
- evict values on completion.
- on and on and on...

(Caching is hard)

### Events that could affect desired cache behavior

Possible cache modifying events in RxJS:

- Subscription to the source
- New value received from source
- Error from source
- Completion from source
- Source unsubscribed
- Consumer subscribed
- Consumer unsubscribed

### Goals

I think we should define a primary `cache` operator (or a set of operators existing or non-existant, that can be composed to get all desired caching behaviors). From there, I think we should identify the most desired caching behaviors and create operators specific for those needs (with the primary cache operator driving them).

blesh:
oops... still working on that... please hold...

trxcllnt:
@blesh I experimented a long time ago with a `memoize` operator (since Observables are async functions, memoization should be possible) but I think it's mostly already covered by `publishReplay`, or so implementation-specific that it's impossible to solve generically without becoming Falcor or GraphQL.

[...]

PR #1320, opened 2016-02-09
feat(cache): add cache operator

blesh:
The `cache` operator is effectively `publishReplay` and `refCount`.

The main reasons it's not named `shareReplay`:

1. "`share`" no longer means "publishTypeX + refCount" in RxJS 5. It's not publish().refCount() it's a replayable refCounted observable. So `shareReplay` doesn't make sense, as this has little to do with `share`.
2. `cache` is more terse and descriptive to the general user. What does it do? It caches values. Does it share a replay? sure. Will new users ever search for that... probably not? Debatable.

Bike shedding is bike shedding. I think it's cleaner.

(ATTN: @staltz ... I think there's probably a better way to write the tests than what I did)

### DIFF
--- a/spec/operators/cache-spec.js
+++ b/spec/operators/cache-spec.js
@@ -1,4 +1,4 @@
-/* globals describe, it, expect, expectObservable, hot, rxTestScheduler */
+/* globals describe, it, expect, expectObservable, hot, rxTestScheduler, time */
 var Rx = require('../../dist/cjs/Rx');
 
 var asap = Rx.Scheduler.asap;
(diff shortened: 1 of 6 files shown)


blesh:
Related #839

staltz:
@blesh tests look alright, except there could be more of them (but that's debatable since we already have tests for publishReplay).

blesh:
@staltz that's what I was thinking, I tried to add tests I thought were specific to the behavior, but I can add a few more and repeat those other tests I suppose.

staltz:
PS: what happened to the "advanced cache invalidation attributes" ideas for this operator?

blesh:
> PS: what happened to the "advanced cache invalidation attributes" ideas for this operator?

Lack of interest, mostly.

blesh:
I've gotten some feedback from a few people that perhaps this operator requires some sisters like:

`cacheUntil(observable)`, `cacheWhile(fn)`, `cacheTime(ms)`, and `cacheCount(n)`

blesh:
Instead of the above operators what about a config object?

`cache({ time: 1000, count: 5 })`... etc. Makes the operator more readable I think, and opens the door for other types of cache eviction in the future without breaking the API or needing to add new operators.

blesh:
Alternatively, how do people feel about reversing the order of arguments? `cache(ms, count, scheduler)`?

kwonoj:
maybe name it to `cacheTime(timespan, ?maxCount, ?scheduler)` and introduce pair operator such as `cacheUntil`, etcs? Thought came from operators like `buffer`, `window` splits.

justinwoo:
I said this on twitter, but I think most people want caching by count, usually just 1 or 2, even. Maybe best if that's the first param if this will take multiple params (but I think splitting this into multiple operators would be nicer...)

But if you change the options hash to allow more/less/different values, then that's a change in the type of object you can send through, making it messy.

staltz:
>  `cache({ time: 1000, count: 5 })` 

It's a neat API, we should consider it for the future, but right now it's not consistent with the rest. `cache` API as it was merged is consistent, also `cacheCount`, `cacheTime` would be consistent.

PR #1321, opened 2016-02-09
docs(MIGRATION): remove the non-exist usecase for throttle()

saneyuki:
RxJS v4's `throttle()` does not accept `Observable<T>` as a durationSelector. There are no need the migration path.

- https://github.com/Reactive-Extensions/RxJS/blob/8f12f812d497acf639588e90f74d504a9fc801ec/src/core/linq/observable/throttle.js#L9
- https://github.com/Reactive-Extensions/RxJS/blob/8f12f812d497acf639588e90f74d504a9fc801ec/ts/core/linq/observable/throttle.ts

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -204,11 +204,7 @@ To reduce polymorphism and get better performance out of operators, some operato
       <td><code>debounceTime(delay: number, scheduler?: Scheduler)</code></td>
     </tr>
     <tr>
-      <td rowspan="2"><code>throttle</code></td>
-      <td><code>throttle(durationSelector: Observable)</code></td>
-      <td><code>throttle(durationSelector: Observable)</code></td>
-    </tr>
-    <tr>
+      <td><code>throttle</code></td>
       <td><code>throttle(delay: number, scheduler?: Scheduler)</code></td>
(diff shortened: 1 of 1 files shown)


kwonoj:
Doesn't RxJS5 column need to maintain `throttle(durationSelector)` variant?

saneyuki:
@kwonoj 

I thought this as a migration doc to move from RxJS v4. So I removed `throttle(durationSelector)` which is just a new feature.  If the comparison doc introduces a new feature, I feel we should keep it.

kwonoj:
Those table's for operator splits, introduces new operator variant from RxJS4's polymorphic behavior as well such as `window`, `windowWhen`, `windowToggle` comes from RxJS4. in case of `throttle(selector)` is somewhat different though since original throttle didn't have those behavior though, thought it belongs to operator split allows easier lookup. Maybe I'm incorrect still.

kwonoj:
Merged with a3cc4e7, think it's ok as-is by not displaying RxJS 5 variant is trivial compare to showing non exist one. Thanks @saneyuki .

saneyuki:
@kwonoj Thank you for accepting my proposal!

PR #1324, opened 2016-02-10
fix(benchpress): fix issues with benchmarks

jeffbcross:
Some benchmarks had errors that were causing perf tests to log 0:00 time
values for scriptTime metrics.

Fixed Issues:
 - Was using Rx.js instead of Rx.umd.js in all benchmarks
 - Was calling mergeMap on Rx2 observable instead of flatMap
 - Had mis-named variables in mergeAll spec


blesh:
Thanks @jeffbcross.

blesh:
Merged with 16894bb. It looks like some of the tests still have some issues. So I'll start a separate issue to track that.

PR #1326, opened 2016-02-10
refactor(test): reorder some test case to avoid unexpected jasmine be…

kwonoj:
…havior

Latest travis log (https://travis-ci.org/ReactiveX/RxJS/builds/108206549) shows couple of tests are failing on some browser unexpectedly. This is similar to previous cases of test starts to fail by adding one or another - just simply reordering test makes passes it. Assume it's somewhat related with jasmine's async test handling, meanwhile simply reorder couple of tests to make test executes successfully.

### DIFF
--- a/spec/operators/publish-spec.js
+++ b/spec/operators/publish-spec.js
@@ -16,6 +16,27 @@ describe('Observable.prototype.publish()', function () {
     published.connect();
   });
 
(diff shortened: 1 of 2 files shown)


kwonoj:
Merged with 835b9e9.

PR #1330, opened 2016-02-10
fix(do): fix do operator to invoke observer message handlers in the right context.

trxcllnt:
Presently, if you pass a Subject as the Observer to the `do` operator, this error happens when the Observable emits:
```
Object #<DoSubscriber> has no method 'throwIfUnsubscribed
```

:(

This PR uses the `SafeSubscriber` internals to call the `do` Observer in the right context.


blesh:
Lint is broken:

```
src/operator/do.ts[2, 9]: unused variable: 'Observer'
src/operator/do.ts[5, 9]: unused variable: 'noop'
```

blesh:
Merged. Good catch, @trxcllnt. Surprised we haven't heard about that one before. Or that I haven't run into it. :\

PR #1331, opened 2016-02-10
style(every): fix typos

gluck:
(and a repeated GH issue # in changelog)


typings/node/node.d.ts
```diff
@@ -1292,7 +1292,7 @@ declare module "crypto" {
         update(data: any): void;
         sign(private_key: string, output_format: string): string;
     }
-    export function createVerify(algorith: string): Verify;
+    export function createVerify(algorithm: string): Verify;
```
kwonoj: this typing definition is not originated in this repo but comes from definitelytyped (https://github.com/DefinitelyTyped/tsd), may need to update type definition instead of update it.

Actually I noticed this definition is not being referenced in repo, maybe we can get rid of this.

gluck: Ok, I'll just remove this one from the change then, and let you decide whether the def should be removed.

Thx !

gluck:
Done, rebased & amended

kwonoj:
Merged with 519000b, amended commit message from `..(every)` possibly can be read as operator `every`. Appreciate @gluck for catching this and take effort to create PR!

PR #1332, opened 2016-02-11
chore(typings): update type definition references

kwonoj:
This PR introduces small organization changes to referencing external type definitions.

- Delete type definitions included in codebase not referenced
- Referenced type definition(es6-shim) is managed by type definition manager(https://github.com/typings/typings). `npm install` will resolves type definition also.
- Still refers to ambient es6-shim as is for now, instead of referring external type definitions.


kwonoj:
in further, `es6-shim` definition can be replaced into only required external type definition set, such as `es6-promise` and `es6-collections` since those are only necessary reference in current codebase.

kwonoj:
Marking this as blocked for now, https://github.com/ReactiveX/RxJS/pull/1364 better to be checked in prior to this PR (if it's accepted).

kwonoj:
I'll check this in around today.

kwonoj:
Merged with 03edd9a .

@blesh @staltz @trxcllnt @luisgabriel @saneyuki @david-driscoll please retrigger `npm install` once again with master, this change requires it.

PR #1334, opened 2016-02-11
fix(typings): set map function parameter for Observable.from as optional

mchmielarski:
without this change typescript returns TS2346 error when we use Observable.from with one parameter.

I think that it will be nice if tests will be written in typescript because it will eliminate situations like this.

### DIFF
--- a/src/observable/FromObservable.ts
+++ b/src/observable/FromObservable.ts
@@ -20,7 +20,7 @@ export class FromObservable<T> extends Observable<T> {
     super(null);
   }
 
-  static create<T>(ish: any, mapFnOrScheduler: Scheduler | ((x: any, y: number) => T), thisArg?: any, lastScheduler?: Scheduler): Observable<T> {
+  static create<T>(ish: any, mapFnOrScheduler?: Scheduler | ((x: any, y: number) => T), thisArg?: any, lastScheduler?: Scheduler): Observable<T> {
     let scheduler: Scheduler = null;
     let mapFn: (x: number, y: any) => T = null;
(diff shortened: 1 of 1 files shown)


src/observable/FromObservable.ts
```diff
@@ -20,7 +20,7 @@ export class FromObservable<T> extends Observable<T> {
     super(null);
   }
 
-  static create<T>(ish: any, mapFnOrScheduler: Scheduler | ((x: any, y: number) => T), thisArg?: any, lastScheduler?: Scheduler): Observable<T> {
+  static create<T>(ish: any, mapFnOrScheduler?: Scheduler | ((x: any, y: number) => T), thisArg?: any, lastScheduler?: Scheduler): Observable<T> {
```
kwonoj: this makes sense for signature of generic `from` behavior. however in this case, does array-like observable need to check existence of `mapFn` in line 45? so far I remember arraylikeobservable's mapfn is required parameter. @justinwoo , is my understanding correct? or can we skip `mapFn`?

justinwoo: oh, oops, mapFn should be optional. it's supposed to just use identity when not specified.

kwonoj: Cool, thanks for clarification. So it seems PR's good :)

kwonoj:
Merged with efa4dc3, thanks for catching this @mchmielarski (also @justinwoo for confirm this). 

For testing, we're currently trying to put some effort if it's feasible to migrate current test cases into typescript to have type validation tests as you mentioned. It's not there yet though, it's something we'd like to have.

PR #1335, opened 2016-02-11
feat(compatibility): interop with <= RxJS v4-style Observers and Subjects.

trxcllnt:
I know we have a [big writeup](https://github.com/ReactiveX/RxJS/blob/master/MIGRATION.md) about how we changed all the names, but I want to throw this out there anyway. Having these aliases baked in really helps when upgrading large projects from 4-to-5.

edit: This PR also enables people to mix 4 and 5 in the same project (e.g. someone multicasts an Rx 4 Observable with a 5 Subject), which may be helpful integrating with libraries that haven't upgraded to 5 yet.

cc: @blesh @mattpodwysocki @staltz @kwonoj @david-driscoll @masaeedu

### DIFF
--- a/spec/backwards-compatability-spec.js
+++ b/spec/backwards-compatability-spec.js
@@ -1,0 +1,187 @@
+/* globals describe, it, expect, hot, expectObservable */
+var Rx = require('../dist/cjs/Rx');
+var Subject = Rx.Subject;
+var Observable = Rx.Observable;
+
+describe('backwards-compatible', function () {
+  describe('Subscriber', function () {
+    it('should alias next as onNext', function (done) {
+      Observable.create(function (subscriber) {
+        subscriber.onNext('compatible');
+      })
+      .subscribe(function (message) {
+        if (message !== 'compatible') {
+          done(new Error('Something went wrong.'));
+        } else {
+          done();
+        }
+      });
+    });
+
+    it('should alias error as onError', function (done) {
+      Observable.create(function (subscriber) {
+        subscriber.onError('compatible');
+      })
+      .subscribe(null, function (message) {
+        if (message !== 'compatible') {
+          done(new Error('Something went wrong.'));
+        } else {
+          done();
+        }
+      });
+    });
+
+    it('should alias complete as onCompleted', function (done) {
+      Observable.create(function (subscriber) {
+        subscriber.onCompleted();
+      })
+      .subscribe(null, null, function () {
+        done();
+      });
+    });
+
+    it('should be backwards-compatible with Rx4-style Object Observers', function (done) {
+      var onNextCalled = false;
+      var onErrorCalled = false;
+      var onCompletedCalled = false;
+
+      var success = Observable.create(function (subscriber) {
+        subscriber.next('compatible');
+        subscriber.complete();
+      });
+
+      var failure = Observable.create(function (subscriber) {
+        subscriber.error('compatible');
+      });
+
+      success.subscribe({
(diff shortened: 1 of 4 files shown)


blesh:
I think this isn't a good idea. 

I'm all for an external patch library or something that adds compatibility, but I'm 100% against stubbing in `onNext`, `onError` and `onCompleted` on every observer type, etc. That will encourage people to use those, when they really shouldn't be at all.

We're a full major version different for a reason. We broke compatibility.

If we wanted to deprecate `onNext`, etc, we should do that in a new release of RxJS 4, by adding `next`, `error`, `complete` and adding deprecation warnings to the other methods.

Otherwise, we build a Rx4-Rx5 bridge library, probably somewhere else.

trxcllnt:
@blesh one could make that argument against `flatMap` as well, but we haven't. A library (while annoying) could accomplish everything in this PR _except_ for [this bit in `SafeSubscriber`](https://github.com/trxcllnt/RxJS/blob/backwards-compatibility/src/Subscriber.ts#L128), which ensures [these](https://github.com/trxcllnt/RxJS/blob/backwards-compatibility/spec/backwards-compatability-spec.js#L43) [tests](https://github.com/trxcllnt/RxJS/blob/backwards-compatibility/spec/backwards-compatability-spec.js#L129) pass.

blesh:
> one could make that argument against flatMap as well, but we haven't. 

I'd rather remove the alias for flatMap than merge this. LOL

`flatMap` is there because it's a functional programming idiom. That's it.

kwonoj:
I agree this can be useful migration paths where RxJS4 or below version's heavily used, but think it'd be better to be separate package instead of supported as main library feature.

trxcllnt:
> A library (while annoying) could accomplish everything in this PR _except_ for [this bit in `SafeSubscriber`](https://github.com/trxcllnt/RxJS/blob/backwards-compatibility/src/Subscriber.ts#L128), which ensures [these](https://github.com/trxcllnt/RxJS/blob/backwards-compatibility/spec/backwards-compatability-spec.js#L43) [tests](https://github.com/trxcllnt/RxJS/blob/backwards-compatibility/spec/backwards-compatability-spec.js#L129) pass.

staltz:
I generally agree with Ben in this case. Also I can't see how migration would be so hard on small naming issues like `onNext` => `next`. Also can't see how mixing Rx4 and Rx5 in the same codebase would be a good idea (a lot more kB wasted for no benefit).

trxcllnt:
@staltz I wouldn't mix either, but it could happen in node. Having just updated our projects at work, I can tell you it's a non-trivial task to find all the places someone's using the old Observer API, especially when Subjects have been liberally employed.

@blesh look I'm not thrilled about adding this either, but at the end of the day it's a tiny detail that will make a lot of people's lives easier. There's about to be an influx of `Math.pow(10, 7)` new RxJS users when Angular 2 is released, and most of the Rx documentation online says to use these names.

Of course it's fine we adjust them to match ES committee decisions/JS conventions, but is it really the end of the world to do the same thing React does and print deprecation warnings before breaking things?

**Reminder**: this is impossible to do in an external library unless we expose `SafeSubscriber` for instrumentation. This fit requires change to the `SafeSubscriber` implementation to be complete.

blesh:
After talking with @trxcllnt about this on Slack, I think we can just expose internals like `SafeSubscriber` via `Rx.internal.SafeSubscriber` or the like and enable an external library to take care of this bridge.

Closing this for now.

PR #1347, opened 2016-02-13
feat(let): adds the let operator to Rx.KitchenSink

trxcllnt:


kwonoj:
Merged with dca6504, :+1: @trxcllnt . In further may revise this to import and reexport from rx.ts directly.

Issue #1343, opened 2016-02-13
Implement static if creation operator


trxcllnt:
https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/if.md

kwonoj:
Closed via https://github.com/ReactiveX/RxJS/pull/1348.

PR #1348, opened 2016-02-13
feat(if): add static Observable.if creation operator.

trxcllnt:
Addresses #1343

### DIFF
--- a/src/Rx.KitchenSink.ts
+++ b/src/Rx.KitchenSink.ts
@@ -1,6 +1,7 @@
 export * from './Rx';
 
 // statics
+import './add/observable/if';
 import './add/observable/using';
 
 // Operators
--- a/src/observable/IfObservable.ts
+++ b/src/observable/IfObservable.ts
@@ -1,0 +1,42 @@
+import {Observable} from '../Observable';
+import {Subscriber} from '../Subscriber';
+import {Subscription} from '../Subscription';
+
+export class IfObservable<T, R> extends Observable<T> {
+
+  static create<T, R>(condition: () => boolean,
+                      thenSource?: Observable<T>,
+                      elseSource?: Observable<R>): Observable<T|R> {
+    return new IfObservable(condition, thenSource, elseSource);
+  }
+
+  constructor(private condition: () => boolean,
+              private thenSource?: Observable<T>,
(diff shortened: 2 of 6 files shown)


src/Rx.KitchenSink.ts
```diff
@@ -39,8 +39,10 @@ import './add/observable/fromArray';
 import './add/observable/fromEvent';
 import './add/observable/fromEventPattern';
 import './add/observable/fromPromise';
+import './add/observable/if';
```
kwonoj: is this for `kitchensink` only? not in `rx`?

trxcllnt: @kwonoj in Rx4 `Observable.if` is in the "experimental" package, so I only put it into KitchenSink, though I'm still left wondering why we distinguish between "core" and "kitchen sink" operators in the first place.

src/observable/IfObservable.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    if (errorHappened) {
+      subscriber.error(error);
+    } else if (result && thenSource) {
+      return thenSource.subscribe(subscriber);
+    } else if (elseSource) {
+      return elseSource.subscribe(subscriber);
+    } else {
+      subscriber.complete();
+    }
```
luisgabriel: What about moving this if to a separate function? I know this is not a very performance sensitive operator, but we can have improvements anyway.

trxcllnt: Sure, that's alright with me.

kwonoj: @trxcllnt , would you be update PR as suggested, or proceed to check in as-is?

blesh:
@trxcllnt I think this needs rebased/refactored to move the tests to TypeScript.

trxcllnt:
@blesh done

blesh:
@trxcllnt this still appears to be in need of a rebase. I'm sure that something just moved around you, but can you hit it again? After that, myself or @kwonoj will merge it in ASAP.

kwonoj:
I'll check this in today. :)

kwonoj:
Merged with f7ff7ec, thanks @trxcllnt including several rebase :)

PR #1351, opened 2016-02-14
chore(typings): added type signatures for concat/exhaust/merge/switch

david-driscoll:
also adds signatures for the "All", "Map", and "MapTo" variants.

PR #1352, opened 2016-02-14
chore(typings): added typings for minor operators cache, catch, debou…

david-driscoll:
…nce, etc

----
Will need to be rebased after #1351


src/util/input-types.ts
```diff
@@ -0,0 +1,10 @@
+import {Observable, ObservableInput} from '../Observable';
+
+export type _mapProject<T, R> = (value: T, index: number) => R;
```
kwonoj: are these need to start with `_`?

luisgabriel: I think these input types could be a separate PR as it's not a dependency for creating the *Signature interfaces.

kwonoj: It can be, in that case those new PR should be checked in prior to this though.

- is `util' feasible place for these? (I'm also not sure)
- as above comment, does `_` needed?

kwonoj: I tried this by myself and found it's indifferent to have `_` or not, by this are actually not being exposed as pubic type definition. If intention is to use idiom of non-exposed export, I'm ok with this. Up to you. :)

src/observable/IteratorObservable.ts
```diff
@@ -198,4 +198,4 @@ function sign(value: any) {
     return valueAsNumber;
   }
   return valueAsNumber < 0 ? -1 : 1;
-}
+}
```
kwonoj: is this file has changes other than line ending?

src/util/input-types.ts
kwonoj: one thing though, as other signature does I'd like to suggest name these such as `predicateSignature`, etcs to have clarity since it's not actual predicate but signature of predicate. Any suggestions?

src/operator/inspect.ts
```diff
@@ -12,6 +12,10 @@ export function inspect<T>(durationSelector: (value: T) => Observable<any> | Pro
   return this.lift(new InspectOperator(durationSelector));
 }
 
+export interface InspectSignature<T> {
+  (durationSelector: (value: T) => ObservableOrPromise<any>): Observable<T>;
```
kwonoj: can line 11 also use `ObservableOrPromise`?

src/operator/last.ts
```diff
@@ -16,14 +17,20 @@ import {EmptyError} from '../util/EmptyError';
...
-                           defaultValue?: R): Observable<T> | Observable<R> {
+export function last<T, R>(predicate?: _sourcePredicate<T>,
+                           resultSelector?: (value: T, index: number) => R  | void,
+                           defaultValue?: R): Observable<T | R> {
   return this.lift(new LastOperator(predicate, resultSelector, defaultValue, this));
 }
 
+export interface LastSignature<T> {
+  (predicate?: _sourcePredicate<T>): Observable<T>;
+  (predicate: _sourcePredicate<T>, resultSelector: void, defaultValue?: T): Observable<T>;
```
kwonoj: is `resultSelector` is just `void` in this case instead of function does not returns such as `(..) => void`? line 21 shows now there are 2 types of reselectselector and one is function type does not returns.

src/operator/last.ts
```diff
@@ -16,14 +17,20 @@ import {EmptyError} from '../util/EmptyError';
  * from the source, or an NoSuchElementException if no such items are emitted.
  * @throws - Throws if no items that match the predicate are emitted by the source Observable.
  */
-export function last<T, R>(predicate?: (value: T, index: number, source: Observable<T>) => boolean,
-                           resultSelector?: (value: T, index: number) => R,
-                           defaultValue?: R): Observable<T> | Observable<R> {
+export function last<T, R>(predicate?: _sourcePredicate<T>,
+                           resultSelector?: (value: T, index: number) => R  | void,
```
kwonoj: sorry for picky ( :wink: ), spacing between R / void?

src/operator/last.ts
david-driscoll: This is for those cases where you want a default value, but aren't using the result selector.  So we don't have to define a return type (`<R>`)

src/operator/inspect.ts
david-driscoll: fixed.

src/util/input-types.ts
david-driscoll: I'll just rip them out for now, just easier than splitting hairs. :+1:

src/add/operator/first.ts
```diff
@@ -2,6 +2,6 @@
 import {Observable} from '../../Observable';
 import {first} from '../../operator/first';
 
-Observable.prototype.first = first;
+Observable.prototype.first = <any>first;
```
masaeedu: Copying over my comment I guess; is this not assignable as is?

kwonoj:
Change looks good to me, might need response for @masaeedu 's comment only. Will leave this until question's fully resolved before check this in.

src/add/operator/first.ts
david-driscoll: without it there was a compiler error

masaeedu: Tracing it back, it looks like it's because you have an overload in `FirstSignature` with a `void` parameter. I don't think its possible to pass a void value in TypeScript, so that overload can't be called.

david-driscoll: It lets you supply null / undefined and skip the `R` argument. [see example](http://www.typescriptlang.org/Playground#src=interface%20Observable%3CT%3E%20%7B%7D%0Aexport%20interface%20FirstSignature%3CT%3E%20%7B%0A%20%20(predicate%3F%3A%20(value%3A%20T%2C%20index%3A%20number%2C%20source%3A%20Observable%3CT%3E)%20%3D%3E%20boolean)%3A%20Observable%3CT%3E%3B%0A%20%20(predicate%3A%20(value%3A%20T%2C%20index%3A%20number%2C%20source%3A%20Observable%3CT%3E)%20%3D%3E%20boolean%2C%20resultSelector%3A%20void%2C%20defaultValue%3F%3A%20T)%3A%20Observable%3CT%3E%3B%0A%20%20%3CR%3E(predicate%3F%3A%20(value%3A%20T%2C%20index%3A%20number%2C%20source%3A%20Observable%3CT%3E)%20%3D%3E%20boolean%2C%20resultSelector%3F%3A%20(value%3A%20T%2C%20index%3A%20number)%20%3D%3E%20R%2C%0A%20%20%20%20%20%20defaultValue%3F%3A%20R)%3A%20Observable%3CT%20%7C%20R%3E%3B%0A%7D%0A%0Alet%20f%3A%20FirstSignature%3Cnumber%3E%3B%0A%0Alet%20s%20%3D%20f(x%20%3D%3E%20x%20%3D%3D%3D%201%2C%20null%2C%200)%3B%0Alet%20b%20%3D%20f(x%20%3D%3E%20x%20%3D%3D%3D%201%2C%20x%20%3D%3E%20x.toString()%2C%20'7')%3B%0A%0A%0A%0A).
 Now it does look like that last overload shouldn't be `<T | R>` but just `<R>`, I'll try and fix that tomorrow morning.

david-driscoll:
Fixed up first/last so that the default value is `R` when using the selector and `T` when not using the selector.

kwonoj:
Will check in this around tomorrow. /cc @masaeedu for visibility to see if there's other suggestions before check this ins.

src/operator/last.ts
```diff
@@ -17,11 +17,18 @@ import {EmptyError} from '../util/EmptyError';
  * @throws - Throws if no items that match the predicate are emitted by the source Observable.
  */
 export function last<T, R>(predicate?: (value: T, index: number, source: Observable<T>) => boolean,
-                           resultSelector?: (value: T, index: number) => R,
-                           defaultValue?: R): Observable<T> | Observable<R> {
+                           resultSelector?: (value: T, index: number) => R | void,
```
masaeedu: Can this be changed to `_mapProject`? I think the compiler is interpreting the return type as `R | void`, rather than the interpreting the whole argument as `predicate | void`.

kwonoj: I thought projection signature (like _mapProject) was taken out in this PR to be dealt separately, isn't it?

masaeedu: @kwonoj Whoops, that's right. I was looking at my squashed changes. The fix here would instead be to add parens.

masaeedu:
LGTM. Some minor quibbles with the function signatures that aren't assignable, but users don't really see those.

src/operator/last.ts
kwonoj: Gotcha, since those to be dealt with projection signature in later, I assume it'd good to kick off this PR to be checked in first - what do you think? @masaeedu

masaeedu: :+1:

kwonoj:
Merged with 9b9b2de, appreciate @david-driscoll !

PR #1358, opened 2016-02-18
chore(KitchenSink): export Operator

kwonoj:
- export Operator as same as Rx core does

Found kitchenSink does not exports operator, amending quickly.


kwonoj:
Merged with 8f7168d .

PR #1364, opened 2016-02-19
Migrate test cases into typescript

### LINKED ISSUE #661: Add test for type definition validation?

kwonoj:
closes #661.

This PR aims to use existing unit test environment for type definition validation as well by migrating into typescript based. As in first step of incremental changes, this specific PR targets migrating test cases only without adding any other changes such as adding type validation tests, fix type definition in code bases, etcs.

To avoid test cases are breaking while migration this PR includes huge chunk of file changes, but most of them are simply amending existing test cases into typescript friendly. High level changes are as below.

- all test spec, helpers are now typescript along with migrated other test chains such as coverage, doctest
- introduced npm script `test_buildonly` : this will compile test only without executing it for cases like type validation test running.
- introduced npm script `test_nobuild`: in opposite to above command, this'll execute existing compiled test without rebuild, reduces test turnaround time when test cases are infrequently changes
- existing script `test` will execute `test_buildonly && test_nobuild`

#### Further goals ####
- add actual test cases for type validation, fix type definition as well
- refactor test helper, support test framework independent / modular / better injection (no global patches, etcs)
- migrate few missing tool chains (test2png)

Opinion, suggestions are gladly welcome.

/cc @david-driscoll , @robwormald for visibility.


staltz:
wow! :clap:

robwormald:
`Sorry, we could not display the entire diff because it was too big.` is when you know somebody is putting in serious work :+1:

david-driscoll:
@kwonoj holy wow!  :clap: :tophat: 

I like that code coverage increased too. :+1:

blesh:
This is nuts. In a good way. I'm super pleased with this PR. Looks good to me so far.

Question: When you were doing this, did you `git mv` the files so it realizes you just renamed them?

kwonoj:
> When you were doing this, did you git mv the files so it realizes you just renamed them?

: actually I didn't, some of files were auto recognized by git as renamed and some weren't. :( Shall I apply it and update PR accordingly?

blesh:
@kwonoj: Looks good to me other than some of the other npm scripts now seem "dead".  `build_test` and `build_cover` don't seem to find any specs. Those are the two I was entering most often. Perhaps remove them for now or just have them echo a message to try the new scripts?

kwonoj:
>   build_test and build_cover don't seem to find any specs

: I've completely missed those, my bad. Instead of removing it, amended those script to include test case build also. I'll think about fine grained script control in further. (such as combination of  build src / test, + run test + coverage...)

kwonoj:
Merged with 4bc32b7. Now one step closer to more type-friendly codebase :) 

I'll work on followup changes to update type definitions as well.

PR #1365, opened 2016-02-19
chore(package.json): rename docgen to publish_docs, add build_docs

staltz:
Renamed npm script build_docs => publish_docs, because that's what it actually does, and renamed
docgen.sh to publish_docs.sh. Then, made npm script build_docs that simply builds the docs webpage,
without publishing.

Just a minor PR to clean up before I start major work related to docs.


package.json
```diff
@@ -22,7 +22,8 @@
     "build_perf": "npm run build_cjs && npm run build_global && webdriver-manager update && npm run perf",
     "build_test": "rm -rf dist/ && npm run lint && npm run build_cjs && jasmine",
     "build_cover": "rm -rf dist/ && npm run lint && npm run build_cjs && npm run cover",
-    "build_docs": "./docgen.sh",
+    "build_docs": "npm run build_es6 && npm run tests2png && esdoc -c esdoc.json",
+    "publish_docs": "./publish_docs.sh",
```
kwonoj: just minor (and picky) suggestion - what about replace inner invocation in shell script to run `build_docs`? then if change's required, only one place can be updated.

staltz: Done :)

kwonoj:
Merged with 6e42174 , thanks @staltz

PR #1368, opened 2016-02-19
fix(forkJoin): accepts observables emitting null or undefined

### LINKED ISSUE #1362: ForkJoin not emit next if some of result is null
```js
var rx = require('rx')
var rxjs = require('rxjs')

var a = rx.Observable.forkJoin(
  rx.Observable.just(10),
  rx.Observable.just(20),
  rx.Observable.just(null),
  rx.Observable.just(40)
)

a.subscribe(
  r => console.log('Rx Next: ' + r),
  err => console.log('Rx Error: ' + err),
  () => console.log('Rx Complete')
)

var b = rxjs.Observable.forkJoin(
  rxjs.Observable.of(50),
  rxjs.Observable.of(60),
  rxjs.Observable.of(null),
  rxjs.Observable.of(80)
)

b.subscribe(
  r => console.log('Rxjs Next: ' + r),
  err => console.log('Rxjs Error: ' + err),
  () => console.log('Rxjs Complete')
)
```

Result
```
Rx Next: 10,20,,40
Rx Complete
Rxjs Complete
```

kwonoj:
closes #1362.

This PR updates behavior of forkJoin to check provided observable has emitted value by introducing explicit flag as same as RxJS 4 implementation to avoid emitting `null` or `undefined` considered as observable does not contain any value.

### DIFF
--- a/spec/observables/forkJoin-spec.js
+++ b/spec/observables/forkJoin-spec.js
@@ -13,6 +13,18 @@ describe('Observable.forkJoin', function () {
     var expected = '--------------(x|)';
 
     expectObservable(e1).toBe(expected, {x: ['d', 'b', '3']});
+
+    //Hack - just adding one new test cases in here or either jasmin-is-weird-spec.js,
+    //one of test breaks under publish-spec.js
+    var e2 = Observable.forkJoin(
+               hot('--a--b--c--d--|', { d: null }),
+               hot('(b|)'),
+               hot('--1--2--3--|'),
+               hot('-----r--t--u--|', { u: undefined })
+            );
+    var expected2 = '--------------(x|)';
+
+    expectObservable(e2).toBe(expected2, {x: [null, 'b', '3', undefined]});
   });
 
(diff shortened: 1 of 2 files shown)


spec/observables/forkJoin-spec.js
```diff
@@ -13,6 +13,18 @@ describe('Observable.forkJoin', function () {
     var expected = '--------------(x|)';
 
     expectObservable(e1).toBe(expected, {x: ['d', 'b', '3']});
+
+    //Hack - just adding one new test cases in here or either jasmin-is-weird-spec.js,
```
kwonoj: another weirdness of jasmine behavior, just separate this as test case makes another test fails. Placing under 'jasmine-is-weird' also doesn't work in this case.

kwonoj:
I'll check this in around today, leaving bit more to see any other suggestions around.

kwonoj:
Merged with 6279d6b.

PR #1371, opened 2016-02-20
feat(join): add join operator

### LINKED ISSUE #1341: Implement join operator
https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/join.md

luisgabriel:
Closes #1341.

### DIFF
--- a/src/operator/join.ts
+++ b/src/operator/join.ts
@@ -1,0 +1,186 @@
+import {Observable} from '../Observable';
+import {ArrayObservable} from '../observable/ArrayObservable';
+import {Operator} from '../Operator';
+import {Subscriber} from '../Subscriber';
+import {OuterSubscriber} from '../OuterSubscriber';
+import {InnerSubscriber} from '../InnerSubscriber';
+import {subscribeToResult} from '../util/subscribeToResult';
+import {take} from './take';
+
+/**
+ * Returns an Observable that combines the items emitted by two Observables, and selects which items to combine
+ * based on duration-windows that you define on a per-item basis via the durtionSelector functions. You implement
+ * these windows as Observables whose lifespans begin with each item emitted by either Observable. When such a
+ * window-defining Observable either emits an item or completes, the window for the item it is associated with
+ * closes. So long as an item's window is open, it will combine with any item emitted by the other Observable.
+ * `resultSelector` is the function you define to combine the items.
+ *
+ * @param {Observable} right  the right observable sequence to join elements for
+ * @param {Function} leftDurationSelector  a function to select the duration (expressed as an observable sequence)
+ * of each element of the left observable sequence, used to determine overlap
+ * @param {Function} rightDurationSelector  a function to select the duration (expressed as an observable sequence)
+ * of each element of the left observable sequence, used to determine overlap
+ * @param {Function} resultSelector  a function invoked to compute a result element for any two overlapping elements
+ * of the left and right observable sequences.
+ * @returns {Observable} an observable sequence that contains result elements computed from source elements that
+ * have an overlapping duration
+ */
+export function join<T1, T2, R>(right: Observable<T2>,
+                                leftDurationSelector: (value: T1) => Observable<any>,
+                                rightDurationSelector: (value: T2) => Observable<any>,
+                                resultSelector: (left: T1, right: T2) => R): Observable<R> {
+  const o = new ArrayObservable([this, right]);
+  return o.lift(new JoinOperator(leftDurationSelector, rightDurationSelector, resultSelector));
+}
+
+export interface JoinSignature<T> {
+  <T2, R>(right: Observable<T2>, leftDurationSelector: (value: T) => Observable<any>,
+          rightDurationSelector: (value: T2) => Observable<any>,
+          resultSelector: (left: T, right: T2) => R): Observable<R>;
+}
+
+class JoinOperator<T1, T2, R> implements Operator<any, R> {
+  constructor(private leftDurationSelector: (value: T1) => Observable<any>,
+              private rightDurationSelector: (value: T2) => Observable<any>,
+              private resultSelector: (left: T1, right: T2) => R) {
+  }
+
+  call(subscriber: Subscriber<R>): Subscriber<any> {
+    return new JoinSubscriber(subscriber, this.leftDurationSelector, this.rightDurationSelector, this.resultSelector);
+  }
+}
+
+class Context<T> {
+  public map: Map<number, T> = new Map<number, T>();
+  public done: boolean = false;
+
+  constructor(public idCounter: number, public source: Observable<T>) {
+  }
+
+  get empty(): boolean {
+    return this.map.size === 0;
+  }
+
+  add(value: T): number {
+    const id = this.idCounter;
+    this.idCounter += 2;
+    this.map.set(id, value);
+    return id;
+  }
+
+  remove(id: number): void {
+    this.map.delete(id);
+  }
+}
+
+class JoinSubscriber<T1, T2, R> extends OuterSubscriber<any, R> {
+  private left: Context<T1>;
+  private right: Context<T2>;
+
+  constructor(destination: Subscriber<R>,
+              private leftDurationSelector: (value: T1) => Observable<any>,
+              private rightDurationSelector: (value: T2) => Observable<any>,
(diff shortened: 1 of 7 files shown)


kwonoj:
I've thought about bit if it's possible to make this without exposing `outerIndex` as public using innersubscription management looks like 

```js
interface Context<T1, T2> {
  observable: Observable<T1 | T2>;
  subscription: Subscription;
}

class join {
  private left: Context<T1, T2>;
  private right: Context<T1, T2>; 

  protected _next(value: Observable<T1 | T2>): void {
    (this.left) ? this.right = { observable: value, subscription: null } : this.left = { observable: value, subscription: null};
  }
}
```

rough proof of concept can be seen at https://github.com/kwonoj/RxJS/commit/1e06664eb56856ba4c9ee9edbacc5b0381e002fc, what do you think? since it's rough test code, still few topics I'd like to improve such as

- not using map for managing values?
- better duration completion handling?

and some others I couldn't think of.

kwonoj:
+ clarification : I'm not suggesting to update PR as comment. Just kicking off discussion. Not even sure if innersubscription management is efficiently perform compare to original PR.

src/operator/join.ts
```diff
@@ -0,0 +1,172 @@
...
+ * @param {Function} rightDurationSelector  a function to select the duration (expressed as an observable sequence)
+ * of each element of the left observable sequence, used to determine overlap
+ * @param {Function} resultSelector  a function invoked to compute a result element for any two overlapping elements
+ * of the left and right observable sequences.
+ * @returns {Observable} an observable sequence that contains result elements computed from source elements that
+ * have an overlapping duration
+ */
+export function join<T1, T2, D1, D2, R>(right: Observable<T2>,
+                                        leftDurationSelector: (value: T1) => Observable<D1>,
+                                        rightDurationSelector: (value: T2) => Observable<D2>,
```
blesh: The type of the duration selector observables should be `any`, since the typing doesn't at all matter there. So we don't need D1 and D2, I don't think.

src/operator/join.ts
```diff
@@ -0,0 +1,172 @@
...
+    try {
+      duration = durationSelector(value);
+    } catch (err) {
+      this.destination.error(err);
+      return false;
+    }
+    const durationObservable = take.call(duration, 1);
+    this.add(subscribeToResult(this, durationObservable, null, id));
+    return true;
+  }
```
blesh: Does it improve performance at all to move lines 116-118 to their own function?

src/operator/join.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  private tryResultSelector(leftValue: T1, rightValue: T2): void {
+    let result: R;
+    try {
+      result = this.resultSelector(leftValue, rightValue);
+    } catch (err) {
+      this.destination.error(err);
+      return;
+    }
+    this.destination.next(result);
```
blesh: could probably put `destination` in a `const` here.

src/operator/join.ts
luisgabriel: Indeed. I'll remove D1 and D2.

src/operator/join.ts
luisgabriel: I'll add perf tests and check this.

src/operator/join.ts
luisgabriel: Ok. Is it some optimization or just style?

luisgabriel:
@kwonoj I think I can improve the solution with some of your suggestions. ;) I'll update this PR as soon as I can.

Answering some specific parts:

> rough proof of concept can be seen at kwonoj@1e06664, what do you think? since it's rough test code

I was not very happy with exposing `outerIndex` as a public member (maybe a getter would be better?), but I think this is a useful information to have on `notifyComplete()` the same way it is available in `notifyNext()`. Is there a reason not having it on `notifyComplete()`?

> not using map for managing values?

The items can be removed in any order depending on duration. The first item added can be the last to be removed, i.e. there is no relation between when the item is added to the list and when it is removed. We could use an array instead, but it'll require searching for the item. Map seems a more approriate solution.

> better duration completion handling?

I tried to keep it as similar as possible to how RxJS 4 is handling completion to make sure it behaves similarly.

kwonoj:
>  Is there a reason not having it on notifyComplete()?

I don't have clear visibility on those too. I think you may able to file an issue with compelling usecases. As similar, `notifyNext` didn't have `innerSub` as parameter and it's added recently, my poc branch utilized it.  Personally, yet not sure about having idx in error / complete.

 > Map seems a more approriate solution. 

In general I agree, that question was just came out of curiosity if it's possible to do so.

src/operator/join.ts
blesh: Sorry, just getting back to this... it's a bit of both. 

1. Every call to `this.something` incurs a property/hashtable type lookup. If you stuff it in a variable, you don't have to make that lookup. It's not a _huge_ optimization, but it's a simple one.

2. When minified, it'll be smaller:

`this.foo.something(); this.foo.whatever();` minified is the same thing -> `this.foo.something();this.foo.whatever()`.

But with the const before hand:
`const foo = this.foo; foo.something(); foo.whatever();` -> `const a = this.foo; a.something(); a.whatever()`

blesh:
Looks like this will need to be rebased.

blesh:
Also, I think I'd like to block this until #1364 is merged. After that you'll need to convert your tests to TypeScript. Cool?

blesh:
Okay, this just needs rebased now, I think.

luisgabriel:
I'll update it later today.

luisgabriel:
Just updated the PR. 

Changes:
- Remove `D1` and `D2` type parameter
- Expose `outerIndex` of **InnerSubscriber** as a getter instead of make it public
- Add a **Context** class to encapsulate the state related to each observable that is being joined
- Move `durationSelector` subscription to a separate function (try/catch optimization)
- Completion handling is simpler now due to the use of **Context**
- Change `join` declaration on **Observable** and **CoreOperators** to use **JoinSignature**
- Move unit tests to TypeScript

luisgabriel:
@kwonoj I'm still using the index/id approach but I think it's simpler now with Context.

kwonoj:
I'm still in quite favor of not exposing outerindex completely, but that's purely my own :) change looks ok to me. Once @blesh approves, I'll check in this.

luisgabriel:
@kwonoj I kept using the index/id approach for two reasons:

1. I don't like very much having an Map where the key is a reference type
2. Without something that can identify the observable that is completing we have to "guess" in which Map the Subscription is trying to delete it from the left first and if it's not there, deleting from the right.

The first is not a strong argument as it's a matter of taste/style. But the second have performance implications. If the subscription is in the right Map, it'll walk through the whole left Map looking for a subscription that is not there.

That said, I have no problem in changing the implementation if you guys prefer not to expose the outerIndex from InnerSubscription.

kwonoj:
@luisgabriel Yes I understand, reason I mentioned it was 'my favor', and not insist to change it :)

blesh:
This is a pretty in-depth implementation. It's going to take more than a while to review. Can you please add perf tests? It's important that we make sure this hits the perf numbers well.

luisgabriel:
@blesh sure. No problem.

blesh:
Hey, @luisgabriel ... sorry I haven't gotten back to this one yet.  Hey, @trxcllnt, can you give this a review if you have the time?

luisgabriel:
@blesh No problem. I didn't have time to add the perf tests yet. I'll add them as soon as a can.

blesh:
Sorry, @luisgabriel, this is going to need rebased again. While this PR looks fine to me, the major hold up is I'm looking for someone that actually uses `join` a lot to review the PR. @mattpodwysocki, can you lend a hand? Also pinging @trxcllnt again.

luisgabriel:
@blesh I'll take some time this weekend to work on this. I have to rebase it and change the interface to accept `Subscribable` instead of `Observable`.

blesh:
Ping @luisgabriel ... just checking in.

luisgabriel:
@blesh I'm having trouble with the tests after rebasing. I'll try to fix it.

luisgabriel:
@blesh Sorry for the long delay. I'm no sure when I can work on this. So I'm going to close this for now. As soon as I can, I'll revisit this patch and submit a new PR.

PR #1377, opened 2016-02-23
chore(package.json): update package dependencies

kwonoj:
Updates TypeScript to release 1.8 to prepare module augmentation, as well as other dependencies.


kwonoj:
Merged with 150a2a1 .

PR #1378, opened 2016-02-23
docs(spec): show tests in ESDocs

staltz:
Add ESDoc @test tag on each *-spec.js file in order to make it show in the ESDoc page, and also
tweak esdoc.json to show tests.

To see this page live, open http://rxjs5-esdoc-tests.surge.sh/test.html

screen shot 2016-02-23 at 11 19 12


spec/operators/bufferCount-spec.js
```diff
@@ -2,6 +2,7 @@
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
+/** @test {bufferCount} */
```
staltz: Usually these changes just add the tag, but in other cases...

spec/operators/buffer-spec.js
```diff
@@ -2,7 +2,8 @@
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
-describe('Observable.prototype.buffer()', function () {
+/** @test {buffer} */
+describe('Observable.prototype.buffer', function () {
```
staltz: Usually these changes just add the tag, but in other cases we need to keep it consistent across all files, so I removed the parenthesis

spec/operators/count-spec.js
```diff
@@ -2,7 +2,8 @@
 var Rx = require('../../dist/cjs/Rx');
 var Observable = Rx.Observable;
 
-describe('count', function () {
+/** @test {count} */
+describe('Observable.prototype.count', function () {
```
staltz: And I had to fix cases like these, for consistency

kwonoj:
(If) https://github.com/ReactiveX/RxJS/pull/1364 is landed, does it means this PR need to be updated as well?

staltz:
I don't know... it depends whether git is smart to identify that just the filename extension changed, and can merge the new changes.

kwonoj:
Seems we may know after changes are landed :) let's see how it'll goes, since #1364 is not certain to be accepted at this moment.

blesh:
This looks great.

kwonoj:
@staltz , #1364 's checked in and seems it requires rebase. I'll check it this once it's done.

staltz:
I'll rebase this tomorrow

staltz:
Ok, just FYI, this is taking me some stress to rebase. ESDoc explicitly doesn't support TypeScript, so I need to compile all `.ts` tests to `.js` files, and the TS compilation isn't passing. It'll take some time to manually work this out.

staltz:
@kwonoj we need to coordinate together the rebase I'm doing. I had to do some big changes:

1. change `spec/helpers/test-helper.ts` to patch `global.it = it`
2. not import `it` like `import {it, asDiagram} from '../helpers/test-helper'` in individual `-spec.ts` files because that would make TS compile to `test_helper_1.it('should ...` in `-spec.js` files, and that's bad for ESDoc which expects `-spec.js` files look like normal Mocha-style tests (with `it` calls inside `describe` calls)
3. Change the spec compilation output from `tmp` to `spec-js` because `tmp` also contains `tmp/docs` which in turn also has javascript files (ESDoc generates some javascript) and we don't want ESDoc looking at everything inside `tmp`. Also, we don't want the ESDoc page to show the official path for test files being `tmp/something`.

Can you provide some insight why did you change `1.` and `2.` from how it was before? Is it solving some problem?

staltz:
Another thing I may change is 

```diff
-import {hot, cold, expectObservable, expectSubscriptions} from '../helpers/marble-testing';
 import {DoneSignature, asDiagram, lowerCaseO} from '../helpers/test-helper';
+declare const hot: any;
+declare const cold: any;
+declare const expectObservable: any;
+declare const expectSubscriptions: any;
``` 
in `-spec.ts` files and then

```diff
+import {assertDeepEqual, hot, cold, expectObservable, expectSubscriptions} from './marble-testing';

+global.cold = cold;
+global.hot = hot;
+global.expectObservable = expectObservable;
+global.expectSubscriptions = expectSubscriptions;
```

in `spec/helpers/test-helper.ts`

The result is that it makes the test file in ESDoc more readable:

screen shot 2016-03-02 at 17 13 19

Otherwise it looks like this:

screen shot 2016-03-02 at 17 15 43

Which is bad since we want tests (specially marble diagrams) to be additional documentation, it should be readable.

With TS2JS compilation, we lose those important whitespaces to align the ASCII marble diagrams though. I might need to try again doing a PR to TypeScript so we could add `/* */` comments there to horizontally align those ASCII marble diagrams, like this:

screen shot 2016-03-02 at 17 17 23

But that has to be in a different PR, of course.

kwonoj:
>  why did you change 1. and 2. from how it was before? Is it solving some problem?

: It was ergonomics changes to test cases to import helpers explicitly, intention to move forward test help to be modular without patching global variables any more. (though current still does have some) also provides type inferences to helper method too. Global patch would work, loses some conveniences though. Thinking about further plans, I'd like to have explicit export instead of global patching, but do not have great idea to solve this issue with preserving current way.

staltz:
> intention to move forward test help to be modular without patching global variables any more.

modular so that RxJS users can use the helpers too? And which helpers? I assume mainly `hot`, `cold`, etc, right? They are currently modular, packaged as `helpers/marble-testing.ts`. It's `helpers/test-helper.ts` which is doing the global patch.

I think with the current setup that we have (ESDoc only supporting JS/ES, no TS support), we have to do global patch. If we don't global patch e.g. `it()`, then we won't see test cases in the ESDoc test page.

When it comes to type testing our `spec` folder, I think the priority should to type test operators and Observables and actual RxJS stuff.

kwonoj:
> I assume mainly hot, cold, etc, 

: and also was thought about test-helper to, which setup custom assertions to use hot / cold.

I think with the current setup that we have (ESDoc only supporting JS/ES, no TS support), we have to do global patch.

: Yes, I don't object that. Above comment was if I could achieve both, I'd like to have explicit export without global patching but for now global patching seems way to go for now.

staltz:
Yeah, ESDoc not supporting TS is actually quite annoying. I've been able to get around it most of the time. Perhaps we could have looked into some other library that does straight conversion from TS -> docs, but there's probably even less tools that do that type of stuff.

Overall, I'm ok with hacking around ESDoc as we can. It's mostly useful.

kwonoj:
> Perhaps we could have looked into some other library that does straight conversion from TS -> docs

: Off-topic, but makes me curious - have we considered TypeDoc? (http://typedoc.io/) i.e unsupported feature we'd like to have, etcs?

staltz:
Maybe we could use it, but it's unclear is it beneficial to move at this point. It doesn't seem to support showing the tests, and unclear does it support additional markdown pages organized in an outline.

kwonoj:
> it's unclear is it beneficial to move at this point.

: agreed, question's only for curiousity and not suggesting to migrate into. It could be cobsiderable only if pain of using esdoc is unbearable along with all features we'd like to have is being supported in typedoc.

staltz:
Rebased. Also notice that I removed `import {asDiagram} from 'test-helpers'` because we have two different jasmine configurations, `asDiagram` is different for each, and we need to patch globally.

kwonoj:
Sorry, may need another rebase :(  I'll check in this once it's rebased.

staltz:
Rebased :white_check_mark: :)

package.json
```diff
@@ -22,20 +22,20 @@
...
+    "cover": "istanbul cover -x \"*-spec.js index.js *-helper.js spec-js/helpers/*\" ./node_modules/jasmine/bin/jasmine.js && npm run cover_remapping",
     "cover_remapping": "remap-istanbul -b ./ -i coverage/coverage.json -o coverage/coverage-remapped.json && remap-istanbul -b ./ -i coverage/coverage.json -o coverage/coverage-remapped.lcov -t lcovonly && remap-istanbul -b ./ -i coverage/coverage.json -o coverage/coverage-remapped -t html",
     "test_nobuild": "jasmine",
-    "test_buildonly": "rm -rf tmp && tsc --project ./spec --pretty",
+    "test_buildonly": "rm -rf spec-js && mkdirp spec-js && tsc --project ./spec --pretty",
     "test": "npm run test_buildonly && npm run test_nobuild",
     "test_karma": "karma start karma.conf.js",
-    "tests2png": "mkdirp tmp/docs/img && cp -r spec/helpers/tests2png tmp/helpers/tests2png && jasmine JASMINE_CONFIG_PATH=spec/support/tests2png.json",
+    "tests2png": "npm run test_buildonly && mkdirp tmp/docs/img && mkdirp spec-js/support && cp spec/support/*.json spec-js/support/ && JASMINE_CONFIG_PATH=spec/support/tests2png.json jasmine",
```
kwonoj: just note : specifying `JASMINE_CONFIG_PATH` prior to command `jasmine` makes this script windows-incompatible (was reason I changed its order to let jasmine accept it as command argument)

no blocking issue though, I'll amend this later.

staltz: interesting to know, sorry for breaking it

kwonoj: this script is rarely being used, totally ok and I can amend it later quickly. :)

kwonoj:
Merged with bd45a90, thanks @staltz

PR #1385, opened 2016-02-24
docs(esdoc): add playground environment in ESDoc-generated page

staltz:
Add Rx.umd.js script to the ESDoc-generated page, so RxJS can be experimented in the DevTools. Also add a welcome message in the console, and a hint text in index.md.

To see this page live, open http://rxjs5-esdoc-rxjs-playground.surge.sh/

screen shot 2016-02-24 at 15 01 17

### DIFF
--- a/doc/asset/devtools-welcome.js
+++ b/doc/asset/devtools-welcome.js
@@ -1,0 +1,15 @@
+var welcomeText = (
+  ' ____           _ ____      \n'+
+  '|  _ \\ __  __  | / ___|    \n'+
+  '| |_) |\\ \\/ /  | \\___ \\  \n'+
+  '|  _ <  >  < |_| |___) |    \n'+
+  '|_| \\_\\/_/\\_\\___/|____/ \n'+
+  '\nTry this code to get started experimenting with RxJS:\n'+
+  '\n    var subscription = Rx.Observable.interval(500)'+
+  '.take(4).subscribe(function (x) { console.log(x) });\n'
+);
+if (console.info) {
+  console.info(welcomeText);
+} else {
+  console.log(welcomeText);
+}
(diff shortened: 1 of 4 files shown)


kwonoj:
I was thinking about adopt in-frame jsbin like http://xgrommx.github.io/rx-book/content/observable/observable_instance_methods/amb.html in further - would it be considerable? doesn't require to open up console, also supports different env (TypeScript, mostly)

staltz:
We can have both. Many libraries anyway have the library available in devtools console, so we should do that too for consistency.

About the JSBin, if we do that, we shouldn't do it for each operator, because the page will definitely get heavy with so many iframes.

kwonoj:
> We can have both.

: Yes, totally agree.

> , we shouldn't do it for each operator, because the page will definitely get heavy with so many iframes.
 
: Yes, totally agree (too). I liked idea of embedding jsbin but also not quite much in favor of having it per each operator.

kwonoj:
Looks good to me.

blesh:
This looks good. Especially as a first pass. The only thing I can think of is some people might create a Subscription (similar to what's been suggested in console) and not keep a handle to their subscription so they can tear it down. But it's a fine solution until we can get something more robust.

Also I <3 ASCII Art.

blesh:
Just needs a rebase it looks like.

staltz:
I'll rebase it tomorrow, and I can also change that suggested example to have `var subscription = ` prepended. :)

staltz:
Rebested.

kwonoj:
Merged with 98e183d, thanks @staltz .

> Re*best*ed.

: Yup I agree, this PR's best thing to docs. :sunglasses:

Issue #1234, opened 2016-01-22
Publishing to a lifted Subject is type-unsafe


masaeedu:
The `lift` method in `Subject.ts` looks [like this](https://github.com/ReactiveX/RxJS/blob/175efe37c5a9fe0457c983d9df2e16ad4d459792/src/Subject.ts#L32):

```
  lift<T, R>(operator: Operator<T, R>): Observable<T> {
    const subject = new Subject(this, this.destination || this);
    subject.operator = operator;
    return <any>subject;
  }
```

The `T` parameter there is redundant, and causes a loss of type safety. The `lift` method on a `Subject<Foo>` should accept an `Operator<Foo, ?>`, not an `Operator<?, ?>`. A similar change for `Observable.lift` and `Operator.call` has been proposed in #1214.

A few other issues to consider:

 - the signature specifies that it returns an `Observable<T>`. The new `Subject` it is returning has as its operator an `Operator<T, R>`, and will therefore behave as an `Observable<R>`
 - the `Observer` functionality of `subject` is hidden and cannot be accessed, since the return type is `Observable<T>`
 - the destination of the new `Subject` is set to `this.destination`. Assuming you took a `Subject<number>`, lifted it through `.map(n => n.toString())`, and tried `next`ing strings to the resulting `Subject<string>`, the destination of the original subject, which expects numbers, would be handed strings

Overall, I don't think this override is necessary. The behavior for `lift` provided in the `Observable` class (which is to return a new observable, set its source to `this`, and attach an operator), is the same as what is implemented here. The additional behavior of wrapping in a `Subject` that posts to `this.destination` is type-unsafe and unused (at least within the codebase).

trxcllnt:
@masaeedu when Subject's `lift` returns a Subject instance, we can maintain the two-way pipeline through operator chaining for things like web sockets.

masaeedu:
@trxcllnt Could you direct me to the code where this is done?

trxcllnt:
@masaeedu everywhere in the Subject class that references `destination`.

masaeedu:
Yes, I realize that, but my point was that you never (as far as I know), actually call `next` on the value produced by `lift`ing `Subject`. Are there any places in the code where you `lift` a `Subject<Something>`, cast the returned `Obserable<Something>` back to a `Subject<Something>`, and `next` values to it?

masaeedu:
[...]

PR #1388, opened 2016-02-24
Module Augmentation

david-driscoll:
This combines #1284 and #1288 with the latest additions to the signatures, and introduces "module augmentation" for use with Rx.

Thanks to @masaeedu for getting it going, this branch is just an extension of his work. :clap: 

In addition I was able to clean up `Rx.DOM.ts` and `Rx.KitchenSink.ts`, so that they simply export `Rx.ts`, plus add the additional augmented items.

cc @kwonoj @blesh @staltz @luisgabriel


kwonoj:
#1284 and #1288 can be closed on behalf of this? (actually, noticed one of them closed already)

src/Subject.ts
```diff
@@ -15,8 +15,9 @@ export class Subject<T> extends Observable<T> implements Observer<T>, Subscripti
     return new Subject<T>(destination, source);
   };
 
-  constructor(protected destination?: Observer<T>, protected source?: Observable<T>) {
+  constructor(protected destination?: Observer<T>, source?: Observable<T>) {
```
kwonoj: exposing source is expected changes in this PR?

masaeedu:
@kwonoj Yeah, this PR amalgamates the changes in both of those, alongside some changes that were only mentioned in comments (cleanup of Kitchensink to re-export from Rx, for example). Correct me if I'm wrong here @david-driscoll.

kwonoj:
Main question in here will be 'when' to check in this changes. I think it's ok to consider now since 1.8 is released, ng2 b7 is 1.8 friendly, so next release of RxJS will work with those mostly. cc @robwormald and @IgorMinar for visibility, to not proceed if that's incorrect expectation.

Also, maybe we need commit message for 'breaking changes'? (not sure, this is non-api changes but prerequisite dependency changes)

src/Observable.ts
```diff
@@ -127,7 +20,7 @@ export type ObservableInput<T> = ObservableOrPromise<T> | ArrayOrIterator<T>;
  *
  * @class Observable<T>
  */
-export class Observable<T> implements CoreOperators<T>  {
+export class Observable<T> {
```
kwonoj: so interface `coreoperators` and `kitchensinkoperators` are gone completely?

luisgabriel:
Amazing work guys! :clap: :clap:

kwonoj:
> Amazing work guys! 

: Yes, I forgot to add this emoji... :clap: :+1:

src/Subject.ts
masaeedu: This doesn't actually change the visibility of `source`, since it is already a protected property. You're right though, in that it doesn't really seem like it belongs in this PR. It's probably something that crept in while I was fooling around with #1234. @david-driscoll will probably have to pull this out :pray:

kwonoj: >  since it is already a protected property.  

: Yes, you're correct. I've missed inheritance.

Let's take this out. :)

david-driscoll: No that was unintentional (I was seeing a bug in my editor that was already fixed)

src/Observable.ts
david-driscoll: They are, I don't see a real need for them, though if we needed too we could bring them back I suppose.  After all the file was, was a list of signatures and to an interface, which is what we've replaced with this PR.

kwonoj: >  I don't see a real need for them 

: In general I have similar feeling since I didn't had lot of cases of use interface to observable instead of observable itself but might be some usecases I couldn't think of. cc @blesh for opinions.

kwonoj:
Seems Rx.TestScheduler is not exported in kitchensink?

david-driscoll:
@kwonoj yeah... I was missing `TimeInterval`, `TestScheduler` and `VirtualTimeScheduler` from the original file.  Those have been added back in.

src/Rx.DOM.ts
```diff
@@ -1,167 +1,22 @@
...
-import './add/observable/fromEvent';
-import './add/observable/fromEventPattern';
-import './add/observable/fromPromise';
-import './add/observable/interval';
-import './add/observable/never';
-import './add/observable/range';
-import './add/observable/throw';
-import './add/observable/timer';
-import './add/observable/zip';
+export * from './Rx';
```
masaeedu: Not 100% sure, but I think this might need to be `import './Rx'` now, since the augmentations are treated as a side effect. They recently changed what gets emitted from imports/exports.

david-driscoll: I'll look into it.

david-driscoll: @masaeedu you saying we need to import and export it?  If I just change it to import things break.  If I only export, I get correct typings my editor, and everything seems to be working fine.

masaeedu: @david-driscoll Actually, I just realized this is a non-issue, since `Rx` exports lots of stuff. The issue was that if you try to `export * from 'foo'`, and `foo` has important module augmentations but nothing exported, the `export * from` will be elided from the definition file, and you lose the augmentations.

david-driscoll: gotcha, so that means :+1: to me not being crazy! :smile:

robwormald:
great stuff! :clap: 

cc @IgorMinar

blesh:
Very nice @david-driscoll!

blesh:
@kwonoj, can I lean on you to verify this change?

kwonoj:
@blesh Sure. I already reviewed most of changes and mostly fine, was waiting good to go sign from you (and from ng2 also) - I'll respin verification and ping once it's ready to check in.

robwormald:
will verify on our end tomorrow

kwonoj:
@robwormald cool, I'll finish verification on my end before then, can be aligned.

kwonoj:
Change looks good to me. 

@blesh , one thing to be noted - as mentioned in here (https://github.com/ReactiveX/RxJS/pull/1388#discussion_r54024603) , now `CoreOperators` and `KitchenSinkOperators` interfaces are removed by this PR. I don't see strong usecases of those interface since most of behaviors are based on `Observable<T>` itself so think it's ok but letting you know since there may be some reason it should exist.

kwonoj:
Remaining things

- [ ] rebase : @david-driscoll 
- [ ] good to go sign from ng 2 for align release schedule : @robwormald

blesh:
> good to go sign from ng 2 for align release schedule

What's that?

Also: @kwonoj I've added a label for "needs rebase".

kwonoj:
@blesh nothing special - basically this change will make minimum requirement of TypeScript to 1.8.2. Check once again Angular 2 also bumps up minimum requirement before check in to avoid breaking build for further releases. I believe this won't happen but just in case to ensure.

david-driscoll:
@kwonoj @robwormald rebased

kwonoj:
Thanks, I'll check this in around later today.

kwonoj:
Merged with a08fabf, thanks @david-driscoll ! :tada:

PR #1389, opened 2016-02-25
refactor(Subscription): remove to use tryCatch

saneyuki:
- This is just refactoring to remove to use tryCatch. 
- To make types more strictly, we would need to overhaul `Subscription` and separate an interface from `Subscription` class...

### DIFF
--- a/src/Subscription.ts
+++ b/src/Subscription.ts
@@ -13,7 +13,9 @@ export class Subscription {
   public isUnsubscribed: boolean = false;
 
   constructor(_unsubscribe?: () => void) {
-    if (_unsubscribe) {
+    // Check `_unsubscribe`to avoid override `this._unsubscribe` accidentally.
+    // Dn't call `super(_unsubscribe)` from a derived class.
+    if (!!_unsubscribe) {
       (<any> this)._unsubscribe = _unsubscribe;
     }
   }
--- a/src/Subscription.ts
+++ b/src/Subscription.ts
@@ -19,14 +19,13 @@ export class Subscription {
   }
 
   unsubscribe(): void {
-    let hasErrors = false;
-    let errors: any[];
-
     if (this.isUnsubscribed) {
       return;
     }
-
     this.isUnsubscribed = true;
+
+    let hasErrors = false;
+    let errors: any[] = [];
 
     const { _unsubscribe, _subscriptions } = (<any> this);
 
@@ -36,7 +35,7 @@ export class Subscription {
       let trial = tryCatch(_unsubscribe).call(this);
       if (trial === errorObject) {
         hasErrors = true;
-        (errors = errors || []).push(errorObject.e);
+        errors.push(errorObject.e);
       }
     }
 
@@ -51,7 +50,6 @@ export class Subscription {
           let trial = tryCatch(sub.unsubscribe).call(sub);
           if (trial === errorObject) {
             hasErrors = true;
-            errors = errors || [];
             let err = errorObject.e;
             if (err instanceof UnsubscriptionError) {
               errors = errors.concat(err.errors);
--- a/src/Subscription.ts
+++ b/src/Subscription.ts
@@ -1,8 +1,6 @@
 import {isArray} from './util/isArray';
 import {isObject} from './util/isObject';
 import {isFunction} from './util/isFunction';
-import {tryCatch} from './util/tryCatch';
-import {errorObject} from './util/errorObject';
 
 export class Subscription {
   public static EMPTY: Subscription = (function(empty: any){
@@ -26,19 +24,23 @@ export class Subscription {
     }
     this.isUnsubscribed = true;
 
-    let hasErrors = false;
-    let errors: any[] = [];
+    const result: {
+      hasErrors: boolean;
+      errors: any[];
+    } = {
+      hasErrors: false,
+      errors: [],
+    };
 
-    const { _unsubscribe, _subscriptions } = (<any> this);
+    const { _unsubscribe, _subscriptions }: {
+      _unsubscribe: () => void;
+      _subscriptions: Subscription[];
+    } = (<any> this);
 
     (<any> this)._subscriptions = null;
 
     if (isFunction(_unsubscribe)) {
-      let trial = tryCatch(_unsubscribe).call(this);
-      if (trial === errorObject) {
-        hasErrors = true;
-        errors.push(errorObject.e);
-      }
+      callPrivateUnsubscribe(this, result);
     }
 
     if (isArray(_subscriptions)) {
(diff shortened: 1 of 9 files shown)


trxcllnt:
@saneyuki have you compared the performance between master and this branch? I'd be interested to see what sort of difference this makes here.

saneyuki:
@trxcllnt

I added [benchmarks](https://github.com/saneyuki/RxJS/commit/e43e115ad24212e67b54cf206d5e0622d154e3ea) and the following are results:


### Before

```
                                         |                     RxJS 4.0.7 |              RxJS 5.0.0-beta.2 |          factor |      % improved
                            add-function |             1,754,485 (±0.60%) |             8,995,730 (±0.50%) |           5.13x |          412.7%
                        add-subscription |             1,735,979 (±0.72%) |             8,677,772 (±0.63%) |           5.00x |          399.9%
                     remove-subscription |               917,388 (±0.70%) |             1,738,510 (±0.63%) |           1.90x |           89.5%
    unsubscribe-simple-with-unsubscriber |            44,508,246 (±0.69%) |            10,933,015 (±0.70%) |           0.25x |          -75.4%
 unsubscribe-simple-without-unsubscriber |            47,338,542 (±1.01%) |            43,357,329 (±0.57%) |           0.92x |           -8.4%
               unsubscribe-with-children |             1,447,677 (±0.72%) |             1,427,434 (±0.44%) |           0.99x |           -1.4%
             unsubscribe-with-descendant |             2,012,364 (±0.87%) |             2,417,973 (±0.68%) |           1.20x |           20.2%
```


### After to remove `tryCatch`

```
                                         |                     RxJS 4.0.7 |              RxJS 5.0.0-beta.2 |          factor |      % improved
                            add-function |             1,740,157 (±0.59%) |             8,609,207 (±0.50%) |           4.95x |          394.7%
                        add-subscription |             1,713,813 (±0.59%) |             8,671,869 (±0.57%) |           5.06x |          406.0%
                     remove-subscription |               920,657 (±0.58%) |             1,742,495 (±0.47%) |           1.89x |           89.3%
    unsubscribe-simple-with-unsubscriber |            42,486,389 (±0.78%) |            17,084,885 (±0.77%) |           0.40x |          -59.8%
 unsubscribe-simple-without-unsubscriber |            45,266,060 (±0.78%) |            39,835,860 (±0.47%) |           0.88x |          -12.0%
               unsubscribe-with-children |             1,424,311 (±0.70%) |             2,627,288 (±0.89%) |           1.84x |           84.5%
             unsubscribe-with-descendant |             2,002,804 (±0.67%) |             3,592,306 (±0.58%) |           1.79x |           79.4%
```

src/Subscription.ts
```diff
@@ -126,4 +126,24 @@ export class UnsubscriptionError extends Error {
     super('unsubscriptoin error(s)');
     this.name = 'UnsubscriptionError';
   }
+}
+
+function callPrivateUnsubscribe(sub: Subscription): [boolean, any] {
```
blesh: These external functions are going to introduce closure when they're used. If they were on the Subscription prototype, they would not.

src/Subscription.ts
```diff
@@ -46,17 +48,15 @@ export class Subscription {
       const len = _subscriptions.length;
 
       while (++index < len) {
-        const sub = _subscriptions[index];
+        const sub: Subscription = _subscriptions[index];
         if (isObject(sub)) {
-          let trial = tryCatch(sub.unsubscribe).call(sub);
-          if (trial === errorObject) {
+          const [ok, e] = callPublicUnsubscribe(sub);
```
blesh: I'm unsure about this approach, as it's allocating an array.

blesh: A faster approach is to call a function that *only* has the try/catch block in it, that in turn calls another method that doesn't have any try/catch blocks in it (and optimizes better).


```js
class Foo() {
  start(x) {
    //do stuff here
    this.trySomeThing(x);
  }

  trySomeThing(x) {
    let result;
    try {
      result = this.someThing(x);
    } catch (err) {
      this.handleError(err);
    }
    this.handleSuccess(result);
  }

  handleSuccess(result) {
    // do more stuff here.
  }
}
```

blesh: It's ugly, but we're not making these changes for prettiness.

src/Subscription.ts
saneyuki: > These external functions are going to introduce closure when they're used

This `callPrivateUnsubscribe` don't capture a variables in outer environments. This would not be a closure which capture an environment. What would be a problem?

saneyuki:
@blesh 

> 

----------

> A faster approach is to call a function that only has the try/catch block in it, that in turn calls another method that doesn't have any try/catch blocks in it (and optimizes better).
> https://github.com/ReactiveX/RxJS/pull/1389#discussion_r55091956

I tried add  `try***Unsubscribe()` methods to `Subscription` class, however, if we will do it, we cannot make them private methods because `Subject<T>` __implements__  `Subscription` and `Subject<T>.unsubscribe()` calls `Subscription.prototype.remove.call(this, subscription)`.

To add `try***Unsubscribe()` as private methods might be able to do it if we reorder their interfaces and class, but I think we should not do it in this pull request because it may be a large and complex.

So I propose that we concentrate to "remove `tryCatch` from `Subscription`" on this pull request.

How about do you think?

src/Subscription.ts
```diff
@@ -13,31 +11,36 @@ export class Subscription {
   public isUnsubscribed: boolean = false;
 
   constructor(_unsubscribe?: () => void) {
-    if (_unsubscribe) {
+    // Check `_unsubscribe`to avoid override `this._unsubscribe` accidentally.
+    // Dn't call `super(_unsubscribe)` from a derived class.
+    if (!!_unsubscribe) {
```
blesh: If you're asserting "truthiness" here, you can just do it, no need for the `!!`. If you truly need a boolean, then `Boolean(x)` is probably more efficient than two operators back to back (but I'm speculating)

src/Subscription.ts
```diff
@@ -126,4 +119,28 @@ export class UnsubscriptionError extends Error {
     super('unsubscriptoin error(s)');
     this.name = 'UnsubscriptionError';
   }
+}
+
+function callPrivateUnsubscribe(sub: Subscription, result: { hasErrors: boolean; errors: any[]; }): void {
```
blesh: This might perform better as a prototypal member. Since here it's external to the class, to use it in a class method you're forced to close over it, creating a closure.

blesh:
Left a few more comments. This needs rebased again. Sorry for the slowness with the merge, but Subscription is one area we have to get right.

blesh:
This seems to have died out. I'm going to close it for now, because I think that the code has diverged significantly since this PR was made.

PR #1394, opened 2016-02-26
fix(exhaustMap): remove innersubscription when it completes

### LINKED ISSUE #1391: Observable memory leak with long polling
I am trying to do a long polling example using angular2 and observables.  I had basically this snippet in order to perform the long polling. 

```
(<KitchenSinkOperators<any>>Observable.interval(50))
			.exhaustMap(() => this.getResourceFromBackend('state'))
			.map((resource) => resource.value)
			.distinctUntilChanged()
			.startWith(0);
```

This code created a MASSIVE memory leak when run in my various browsers, it kept climbing and climbing.

Yet, when I change that code out to do the same thing with Observable.create, everything works as expected.  This is the converted version of the above code:

```
return Observable.create((subscriber: Subscriber<number>) => {
			let register = 0;
			let subscription: Subscription = null;
			// Set an initial value of 0
			subscriber.next(register);

			let interval = setInterval(() => {
				if(subscription === null) {
					subscription = this.getResourceFromBackend('state')
						.subscribe(({value}) => {
							if (<number>value !== register)
								subscriber.next(register = <number>value);
						}, (err: any) => subscriber.error(err), () => { subscription.unsubscribe(); subscription = null; });
				}
			}, 50);

			return () => {
				clearInterval(interval);
				if(subscription !== null)
					subscription.unsubscribe();
			}
		});
```

Am I misunderstanding something about how Observables and their operators work?  Or is there truly a bug here that needs to be fixed?  For now it seems I have to stick with my custom implementation, although I would rather not if we can fix the initial code.

Thank you very much,

kwonoj:
closes #1391 

/cc @trxcllnt @Guardiannw for reviewing.

### DIFF
--- a/src/operator/exhaustMap.ts
+++ b/src/operator/exhaustMap.ts
@@ -1,6 +1,7 @@
 import {Operator} from '../Operator';
 import {Observable, ObservableInput} from '../Observable';
 import {Subscriber} from '../Subscriber';
+import {Subscription} from '../Subscription';
 import {OuterSubscriber} from '../OuterSubscriber';
 import {InnerSubscriber} from '../InnerSubscriber';
 import {subscribeToResult} from '../util/subscribeToResult';
@@ -95,7 +96,9 @@ class SwitchFirstMapSubscriber<T, I, R> extends OuterSubscriber<T, I> {
     this.destination.error(err);
   }
 
-  notifyComplete(): void {
+  notifyComplete(innerSub: Subscription): void {
+    this.remove(innerSub);
+
     this.hasSubscription = false;
     if (this.hasCompleted) {
       this.destination.complete();


Guardiannw:
Looks good!

kwonoj:
I'll leave this PR opened bit in case if it's available to find additional source of leaks shortly, to squash into this PR instead of create separate.

kwonoj:
Merged with 7ca0859 . Do not close issue yet due to additional change's required to close it out.

trxcllnt:
@kwonoj @Guardiannw [has verified](https://github.com/ReactiveX/RxJS/pull/1395#issuecomment-189572878) that the other leak was the interplay between Angular's zone.js stack traces and our present interval scheduling strategy (repeated calls to `setTimeout` instead of one call to `setInterval`). I've fixed the issue in #1395, and am awaiting a reply from @blesh, who I believe initially misunderstood the new behavior.

PR #1398, opened 2016-02-27
refactor(operator): make some _unsubscribe protected.

saneyuki:


kwonoj:
Merged with 9767158, thanks @saneyuki .

PR #1402, opened 2016-03-01
chore(forEach): add the benchmark

saneyuki:


saneyuki:
By #1434, using `tryCatch` was removed. This pull request changes to add the benchmark

blesh:
This isn't ever going to be an "apples-to-apples" test between RxJS 4 and RxJS 5, since they do two different things. I think the RxJS 4 test should just show "N/A" or something.

blesh:
@saneyuki... for the "Old" Rx4 test, do you think you can figure out something to have it just show as "N/A". This comparison isn't really applicable, but I'd like to capture the metrics. Perhaps we need to add a new type of perf tests to our micro perf suite that is for Rx5-only functionality?

blesh:
This seems to have gone stale, @saneyuki so I'm going to close it for now.

saneyuki:
> @blesh: This seems to have gone stale, @saneyuki so I'm going to close it for now.

Ah, okay. I'm sorry to my reply was too late.

PR #1412, opened 2016-03-01
fix(takeLast): fix takeLast behavior to emit correct order

### LINKED ISSUE #1407: takeLast bug
```js
var foo = Rx.Observable.interval(1000);

var bar = foo.take(5).takeLast(3);

bar.subscribe(
  function (x) { console.log('next ' + x); },
  function (err) { console.log('error ' + err); },
  function () { console.log('done'); },
);
```

This outputs:
```
"next 3"
"next 4"
"next 2"
"done"
```

When it should output
```
"next 2"
"next 3"
"next 4"
"done"
```

Happened with beta.2

- - -

And 
```js
var bar = foo.take(5).takeLast(2);
```

Caused output
```
"next 4"
"next undefined"
"done"
```

kwonoj:
closes #1407 

This PR tries to fix behavior of `takeLast` by changing ring buffer indexing routines. Maybe I miss something, but so far it looks work correctly.

/cc @staltz for visibility.

### DIFF
--- a/spec/operators/takeLast-spec.ts
+++ b/spec/operators/takeLast-spec.ts
@@ -14,6 +14,40 @@ describe('Observable.prototype.takeLast()', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
+  it('should take last three values', () => {
+    const e1 =  cold('--a-----b----c---d--|    ');
+    const e1subs =   '^                   !    ';
+    const expected = '--------------------(bcd|)';
+
+    expectObservable(e1.takeLast(3)).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should take all element when try to take larger then source', () => {
+    const e1 =  cold('--a-----b----c---d--|    ');
+    const e1subs =   '^                   !    ';
+    const expected = '--------------------(abcd|)';
+
+    expectObservable(e1.takeLast(5)).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should take all element when try to take exact', () => {
+    const e1 =  cold('--a-----b----c---d--|    ');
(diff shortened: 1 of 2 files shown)


kwonoj:
Found one failure cases, marking as blocked for now.

staltz:
Nice! Thanks for the fix

spec/operators/takeLast-spec.ts
```diff
@@ -14,6 +14,46 @@ describe('Observable.prototype.takeLast()', () => {
...
+  it('should take all element when try to take exact', (done: DoneSignature) => {
+    const expected = [0, 1, 2];
+
+    Observable.of(0, 1, 2).takeLast(3).subscribe((x) => {
+      expect(x).toBe(expected.shift());
+    }, done.fail, () => {
+      expect(expected.length).toBe(0);
+      done();
+    });
+  });
```
blesh: Can we change the above tests to use marble diagrams and virtual scheduling?

kwonoj: I think so, let me do that.

blesh: I don't mind keeping one synchronous test, of course, but the marble tests are more analogous to the issue that was reported (which was using an interval).

kwonoj:
Updated PR to use marble diagram for test cases.

spec/operators/takeLast-spec.ts
```diff
@@ -14,6 +14,40 @@ describe('Observable.prototype.takeLast()', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
+  it('should take last three values', () => {
+    const e1 =  cold('--a-----b----c---d--|    ');
+    const e1subs =   '^                   !    ';
+    const expected = '--------------------(bcd|)';
```
staltz: Much nicer

kwonoj:
Merged with 73eb658.

PR #1414, opened 2016-03-02
Proposal : type definition test cases

kwonoj:
Now test cases are migrated in typescript, it is possible to integrate type definition test into existing test case execution environment. This PR is initial proposal of logistics. 

Changes are straightforward, test helper exports interface `type` for type definition assertion actually doesn't do anything but only serves as container of all type definition test. These will be verified via compile time of test cases. 

In test cases, very bottom of each test contains `type` assertion and list out possible type definition test cases, majorly
- operator input signature accepts parameter as expected
- operator chain, or subscription correctly delivers context of value type `<T>`
- etcs

intentionwise this is same as #1189 does, only difference is logistics.

/cc @david-driscoll for visibility also.

### DIFF
--- a/spec/helpers/test-helper.ts
+++ b/spec/helpers/test-helper.ts
@@ -17,6 +17,7 @@ global.hot = marbleHelpers.hot;
 global.time = marbleHelpers.time;
 global.expectObservable = marbleHelpers.expectObservable;
 global.expectSubscriptions = marbleHelpers.expectSubscriptions;
+global.type = type;
 
 //amending type definition of jasmine which seems doesn't have this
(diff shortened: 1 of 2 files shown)


kwonoj:
Rebased.

kwonoj:
@david-driscoll , would you able to share opinion this would works for you for type definition test?

david-driscoll:
:+1:

kwonoj:
Cool, I'll check this in around today and it'll unblock your PR #1189.

kwonoj:
Merged with fffc586.

PR #1419, opened 2016-03-02
chore(from): add benchmarks for the path of IteratorObservable.

saneyuki:


kwonoj:
Merged with 2ec5b24, thanks @saneyuki

PR #1421, opened 2016-03-03
refactor(scheduler): specify an explicit type

saneyuki:
This is the set we need not require breaking change.


kwonoj:
Merged with 4377c99, thanks @saneyuki

PR #1422, opened 2016-03-03
chore(test): update type definitions

kwonoj:
This PR updates types in test cases. Some of type definitions were incomplete at the moment (before typing update from @david-driscoll was checked in), test cases were force casted into any to avoid migration. Most of them are now corrected, PR amends test case as well.


kwonoj:
Merged with 2c1a794.

PR #1426, opened 2016-03-03
refactor(Symbols): stop shimming Symbol, and just get the symbols needed

blesh:


src/symbol/observable.ts
```diff
@@ -0,0 +1,18 @@
...
+const Symbol: any = root.Symbol;
+
+export let $$observable: any;
+
+if (typeof Symbol === 'function') {
+  if (!Symbol.observable) {
+    if (typeof Symbol.for === 'function') {
+      $$observable = Symbol.for('observable');
+    } else {
+      $$observable = Symbol('observable');
```
zenparsing: Will this result in cross-frame issues?  Will an Rx.Observable defined in one frame by identified as an observable in another frame?

spec/observables/using-spec.ts
```diff
@@ -1,24 +0,0 @@
...
-
-    source.subscribe();
-
-    if (disposed) {
-      done();
-    } else {
-      done.fail('disposed should be true but was false');
-    }
-  });
-});
```
staltz: Why remove this file? It seems unrelated to the Symbol shimming refactor.

blesh: I'll check again, but when I was working on this, there was no `using` implementation for it to test in the library. It seemed to be a file that was accidentally merged.

trxcllnt: @blesh the `Observable.using` PR was merged, have you rebased from master yet?

src/symbol/observable.ts
blesh: Yeah, that's probably not going to work very well if two different modules default the symbol this way.

blesh: Well, wait, since this is patching `Symbol.observable` on the following line, any other properly implemented `Symbol.observable` module like this should return that value, as they should always check to see if they exist first before they try to get something else.

blesh: Third set of eyes? @trxcllnt? @staltz? @mattpodwysocki?

trxcllnt: @blesh does `Symbol('observable')` patch that on to the global `Symbol.observable`? I'm not too familiar with the feature.

zenparsing: @trxcllnt Line 14 below patches it onto the global Symbol.

@blesh Let's say you have frame A and frame B.  RxJS gets loaded into both frames separately.  Since the `Symbol` function itself is not shared between frames (each frame has its own copy) both RxJS instances will polyfill `Symbol.observable` with a different symbol value.  If you try to give an Observable from frame B to frame A, the RxJS in frame A won't recognize the symbol that was polyfilled into frame B, and therefore it won't recognize it as an "observable-able".

`Symbol.for` doesn't have that problem because it uses a cross-realm global registry of string names to symbol values.  Because of the cross-frame issues, I'd only polyfill if `Symbol.for` was available.  But in that case, there's not much point in polyfilling.  All libraries can just use `Symbol.for("observable")` if `Symbol.observable` doesn't exist.  And of course fall back to `"@@observable"` if `Symbol.for` is not available.

staltz: I generally agree with @zenparsing that we should just use `Symbol.for()` and not `Symbol()`

Issue #1423, opened 2016-03-03
Source maps "sourceRoot" needs to be relative to the source map file's directory


trxcllnt:
See here: https://gist.github.com/trxcllnt/6ec25fe7bc0d0e4fd546

Currently, the `sourceRoot` value in the map files (`filter.js.map` in the gist) is always `"src/"`, but it needs to be relative to the directory that contains the map file (e.g. `"sourceRoot":"../src/"`). This is true of all map files (for example, `observable/dom/AjaxObservable.map.js` needs to be `"sourceRoot":"../../src/"`).

I've tried all the combinations of `--sourceMapRoot` `--mapRoot` `--rootDir` and `--outDir` to no avail, so we probably need a post-build script to recurse and rewrite the `sourceRoot` value of the built files.

kwonoj:
Or what if we amend copy source script to place all source files reside next to sourcemap and transpiled results (and get rid of sourceroot configs) then map can pick up src directly without path lookup.

(it's just thought came out and haven't tested yet, I'll try out on my end)

kwonoj:
> amend copy source script to place all source files reside next to...

: wasn't simple as that, not working :(

alexeagle:
drive-by comment: we found that chrome devtools support for sourcemaps is spotty, and users can fall in a lot of holes where the sourcemaps don't load correctly (eg. server didn't serve the sourcemap, served it at a wrong path, sourcemap reference is not quite right)
For our internal use of TypeScript, we inline the sourcemaps and the sources (in dev mode) and they work every time.

blesh:
> For our internal use of TypeScript, we inline the sourcemaps and the sources (in dev mode) and they work every time.

Seems like a gross workaround, honestly. And not too helpful for debugging production code.

kwonoj:
I supposed this issue is closed via https://github.com/ReactiveX/RxJS/pull/1428, closing issue. If there's unexpected behavior, let's reopen issue or create separate issue to track.

PR #1427, opened 2016-03-03
chore(build): update source copy script with sourcemap to relative path

kwonoj:
@trxcllnt for visibility - would you able to give a shot if this fixes source path lookup issue #1423 ? For me it seems work, but not 100% sure and would like to have second opinion for verification.


kwonoj:
Blocked until it's verified.

kwonoj:
Actually seems I noticed few caveat for this.. closing PR but leave branch to allow anyone tryout. I'll reopen once I get more confidence.

trxcllnt:
@kwonoj doesn't seem to be working, for example in `util/toSubscriber.js.map`, the `"sources":["util/toSubscriber.ts"]` needs to be `"sources":["toSubscriber.ts"]`

kwonoj:
Yes, I noticed that so closed PR immediately :(

PR #1428, opened 2016-03-04
refactor(npm-scripts): fix build to compile source maps and test scripts to clean up after running

### LINKED ISSUE #1423: Source maps "sourceRoot" needs to be relative to the source map file's directory

trxcllnt:
Fixes #1423 :tada: 
```
                                                         (>,                  o (>,   
     oo                  (>,  oo                     ,<)oo |\                8 oo |\  
    8 "} > @ <            /  8 "} > @ <             / 88 "}| \                8 "}| \ 
   .8 .-._/|              |`.8 .-._/|              | `., ., ,'                , ., ,' 
 .'_'`')`_.'              `-.'`')`_.'               `-.'`')'                 ('`')'   
 \| ) /                      ) /                       ) /                   )\____,<)
(>'/  |_,_                  /  |__,                   / (__,_               / (__,_   
   |    (,|                 |    ( /                 |   (-,/               |   (-,/  
 .'    ,'\|               .'    , /                .'    ) /              .'    ) /   
  `._/    )                `._/  '`-                `._,\ '`-              `._,\ '`-  
   \|     '`-               \|                          `\                     `\     
   ''--                    -`' -                        -`'                    -`'    
```


package.json
```diff
@@ -13,36 +13,59 @@
...
     "build_docs": "npm run build_es6 && npm run build_global && npm run tests2png && esdoc -c esdoc.json",
-    "publish_docs": "./publish_docs.sh",
+
+    "check_circular_dependencies": "madge ./dist/cjs --circular",
+
+    "clean_dist_amd": "rm -rf ./dist/amd",
+    "clean_dist_cjs": "rm -rf ./dist/cjs",
+    "clean_dist_es6": "rm -rf ./dist/es6",
+
+    "copy_src_amd": "mkdir -p ./dist/amd && cp -r ./src/* ./dist/amd/",
```
kwonoj: quick note - this won't work on windows (make script compatible to windows is always hard, sigh) . I don't think this blocks this PR, I'll try on my end to find compatible solution for followup.

kwonoj:
Just tried, let's say

folder test->index.ts
```js
import * as Rx from '......\Rx' //import compiled module with this PR

Rx.Observable.of(1).subscribe();
```

tsc index.ts

and shows unexpected compilation failure like below (message may vary)

>../RxJS/dist/cjs/observable/BoundNodeCallbackObservable.ts(16,3): error TS2394: Overload signature is not compatible with function implementation.
../RxJS/dist/cjs/observable/FromEventObservable.ts(62,39): error TS2339: Property 'length' does not exist on type 'NodeList | EventTarget | { addListener: (eventName: string, handler: Function) => void; removeLis...'.
../RxJS/dist/cjs/testing/ColdObservable.ts(35,35): error TS2459: Type 'T' has no property 'message' and no string index signature.

and also typescript compiler tries to compile Rx while trying to compile `index.ts`. As a module consumer, at least build should not be blocked (course it'd better to not build whole Rx again also).

package.json
trxcllnt: @kwonoj it wasn't already incompatible with windows?

kwonoj: No, I always updated script windows compatible (yes, I'm one of few windows user ;) )

trxcllnt:
@kwonoj so is the issue that the TypeScript compiler can't tell the difference between the built JS module and the source TS module now that the source TS files are published with the compiled code? That sounds like we should have a separate dist folder for TypeScript.

kwonoj:
> TypeScript compiler can't tell the difference between the built JS module and the source TS module now that the source TS files are published with the compiled code? 

: It seems like if original TS file's reside next to transpiled JS in module, TS compiler seems pick up source instead of transpiled one when try to import it. So I agree, it may need separate distribution of source code, which may affect this PR as well about locating source root.

blesh:
I'm actually confusing myself a little here trying to figure out what the desired source mapping would be for our users. If they're using the `rxjs` npm package, you'd think they'd want to see source for the unminified CJS build in the debugger, and not TypeScript.

Is the goal of this PR to get the TypeScript source mapped for the rxjs and rxjs-es builds?

kwonoj:
I expect source map to locate typescript source, when import transpiled build like CJS, or else.

In case of Javascript / Typescript consumers both, when they locate source directly they get transpiled CJS output, while they'll available to look into original TS when enable source map lookup. If they're pipelining other bundling processes (like minify, browserify, etcs..) then it's up to users to not to chain source map to simply display CJS build, or chain it to TypeScript source map to original source code.

trxcllnt:
@blesh when we use the `--souceMap` compiler option, TypeScript creates sourceMaps from the generated ES6/CJS/AMD code back to the original TypeScript source. The problem is there's no combination of `tsc` compiler options that generate proper paths for our standalone distribution model. 

@kwonoj do you know if TypeScript supports some sort of package.json field (similar to how Browserify prefers the file specified as `browser` instead of `main`) that allows us to direct the TypeScript compiler to the `src` folder instead of `dist/cjs`?

kwonoj:
@trxcllnt this is full schema of `tsconfig` (http://json.schemastore.org/tsconfig) and so far I know there isn't similar options. Most options are identical to compiler commandline options (https://github.com/Microsoft/TypeScript/wiki/Compiler%20Options)

I assume you already tried `--sourceRoot` options? (it's indicating debugger to where to look source files, so seems not applicable though)

trxcllnt:
@kwonoj yes, `--sourceRoot` doesn't do what we need. the only other alternative I can think of is to segregate the built JS from the TS source, but still publish them under `dist/<build-type>`, which changes the distribution file structure like so: 

```
-dist/cjs
  - package.json
  - src
    /* source TS files */
  - lib
    /* built JS files */
```

kwonoj:
That's one way I thought, but what happens to npm packages for each module (i.e `rxjs` for CommonJS)? those package does not have hierarchy allows to put source as illustrated above.

trxcllnt:
@blesh I've amended this PR to write the TS source files to `dist/<build-type>/src/` and write the built JS files back to `dist/<build-type>/`, which should solve the issue @kwonoj was seeing.

There's also a pre-existing issue with source maps for the global UMD and SystemJS files in `dist/global` and `dist/cjs/bundles`, but that fix should probably go in a follow-up PR.

tools/make-umd-bundle.js
```diff
@@ -12,4 +12,4 @@ var b = browserify('dist/cjs/Rx.KitchenSink.js', {
   baseDir: 'dist/cjs',
   standalone: 'Rx.KitchenSink'
 });
-b.bundle().pipe(fs.createWriteStream('dist/global/Rx.KitchenSink.umd.js'));
\ No newline at end of file
+b.bundle().pipe(fs.createWriteStream('dist/global/Rx.KitchenSink.umd.js'));
```
kwonoj: is this file changed?

trxcllnt: @kwonoj I think my editor added a new line at the end

blesh:
LGTM

kwonoj:
Merged with 1963afe, thanks @trxcllnt

kwonoj:
and for expanded script list, created PR https://github.com/ReactiveX/RxJS/pull/1446 to display script description via cli.

PR #1431, opened 2016-03-04
Method `ofObjectChanges` Does not exist in RXJS 5

MartinMuzatko:
See https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/ofobjectchanges.md

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -90,6 +90,7 @@ enabling "composite" subscription behavior.
 |`maxBy`|`-`|
 |`minBy`|`-`|
 |`onErrorResumeNext`|`-`|
+|`ofObjectChanges`|`-`|
 |`pausable`|`-`|
 |`pausableBuffered`|`-`|
 |`shareReplay`|`-`|


blesh:
@MartinMuzatko given that `Object.observe` is being dropped from the running for standardization, it's probably best that this creation function is kept in a separate module/package. What do you think @mattpodwysocki?

blesh:
oh... haha.. this is to add that to the list in the docs. I thought it was a feature request.

LGTM

mattpodwysocki:
:-1: `ofObjectChanges` relies upon `Object.observe` which is now deprecated and no longer to be a standard.

mattpodwysocki:
Nevermind me, yes, and RxJS v4 should have deprecated it too

kwonoj:
Merged with f195395 .

Thanks @MartinMuzatko for effort to create PR, I have checked in changes with amending commit message to follow commit message guidelines (https://github.com/ReactiveX/RxJS/blob/master/CONTRIBUTING.md#commit-message-format) which current repo uses it for automatic changelog creation. 

Appreciate for PR, and hope to see another PR from you in further!

MartinMuzatko:
Thank you! Happy to help wherever possible ;)

fgalassi:
I don't think ofObjectChanges is viable anymore, but ES6 has Proxies. Would you mind an implementation of `ofProxyChanges` along these lines of https://stackoverflow.com/questions/44369387/how-to-watch-object-changes-with-rxjs-5/44396986#44396986 ?

The only thing that bothers me is returning a pair [Proxy, Observable] but I don't see a better way given the Proxy api.

PR #1433, opened 2016-03-07
test(behaviorSubject): refactor test case to compatible with IE9

### LINKED ISSUE #1408: [Compatibility] BehaviorSubject.value test fails on IE9
One test case for BehaviorSubject is failing on IE9 as below.

```
IE 9.0.0 (Windows 7 0.0.0) BehaviorSubject should not allow you to set `value` directly FAILED
	Expected function to throw an exception.
```

this is only test failure for now.

kwonoj:
- update test case to explicitly check value instead of set value throws, IE9 behaves to not throw
by not supporting strict mode in Object.defineProperty

closes #1408


kwonoj:
Merged with feb73df .

PR #1434, opened 2016-03-08
fix(forEach): ensure that teardown logic is called when nextHandler t…

### LINKED ISSUE #1411: forEach should unsubscribe if callback function throws
Test case:

```js
new Rx.Observable(sink => {
    sink.next(1); 
    sink.next(2); 
    sink.complete();
}).forEach(x => {
    console.log(x);
    throw new Error("boom");
}).catch(err => console.log(err));
```

Expected result: Only "1" is logged.
Actual result: "1" and "2" are logged.

I recently fixed this issue over on es-observable:

- [Spec](https://zenparsing.github.io/es-observable/#observable-prototype-foreach-next-functions) (3.2.1 step 3.d)
- [Reference Implementation](https://github.com/zenparsing/es-observable/blob/master/src/Observable.js#L298)

blesh:
…hrows

Also removes use of `tryCatch` util, since it's unlikely to add any optimization in this case.

fixes #1411

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -80,6 +80,62 @@ describe('Observable', () => {
       })
       .then(done);
     });
+
+    it('should handle a synchronous throw from the next handler and tear down', (done: DoneSignature) => {
+      let unsubscribeCalled = false;
+      const syncObservable = new Observable<number>((observer: Rx.Observer<number>) => {
+        observer.next(1);
+        observer.next(2);
+        observer.next(3);
+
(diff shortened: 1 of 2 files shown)


blesh:
attn @zenparsing.

kwonoj:
Change looks good to me.

zenparsing:
I like this approach.  Looks good.

blesh:
Thanks for the review @zenparsing and @kwonoj. Merged.

PR #1461, opened 2016-03-12
chore(package.json): correct RxJS 4 dependency

kwonoj:
This PR's minor correction to dependency version bump, to not to use specific version of RxJS 4 but to use latest.


kwonoj:
Merged with aa91475.

Issue #1392, opened 2016-02-26
SwitchMap on forkJoined observable


sumigoma:
I have an array of observables (ng2 http requests) that I would like to cancel simultaneously using switchMap, but I'm finding that wrapping a forkJoin in switchMap doesn't cause the inner observables to get unsubscribed and hence cancelled. (Obviously, this works if there is no forkJoin).

Is there a recommended way of doing something like this? and in general, would unsubscribing from a forkJoined observable cause the inner observables to be disposed as well?

kwonoj:
Maybe I'm saying dumb without knowing full picture, but is there specific reason to wrap observables with forkJoin? forkJoin's supposed to wait all provided observable completes and take last emitted value.

sumigoma:
The number of observables (ajax calls) I have is variable in my particular use case, and I thought forkJoin might be a way to cancel all of them at once (without having to have a handle to each one with a switchMap on each). If a switchMap is on a forkJoin, is it true that it has no way of completing/cancelling the wrapped observables? 

This may not be the right approach to this problem, but the alternative of keeping a dynamic array of observables seems sloppy (e.g. if I have ten ajax calls being made and the next "refresh" cycle has nine, I'd like to cancel all ten before moving on). Is switchMap not the right operator to use here?

kwonoj:
I'd still consider switchMap to get expected behavior, without involving forkJoin. It maybe helpful if you could share small snippet to replicate situation.

trxcllnt:
@sumigoma you're right, from what I can tell [here](https://github.com/ReactiveX/RxJS/blob/master/src/observable/ForkJoinObservable.ts#L39) forkJoin isn't keeping track of its inner subscriptions. @kwonoj @blesh this is certainly a bug and should be addressed asap.

sumigoma:
Thank you @kwonoj, I could share a snippet but I'm not sure how helpful it would be. If switchMap was able to cancel the inner subscriptions of a forkJoin, I could cancel numerous calls without having to keep a handle on each observable. e.g. If I have N ajax requests and I want to cancel all of them, but N is variable, the forkJoin would help me immensely.

@trxcllnt Thanks, I wasn't sure if this was by design or not.

jonnydevlive:
[...]

PR #1462, opened 2016-03-13
fix(forkJoin): dispose the inner subscriptions when the outer subscription is disposed

trxcllnt:
**Description:**
Track forkJoin's inner subscriptions so they're unsubscribed when the outer subscription is unsubscribed.

#1392

### DIFF
--- a/spec/observables/forkJoin-spec.ts
+++ b/spec/observables/forkJoin-spec.ts
@@ -263,4 +263,20 @@ describe('Observable.forkJoin', () => {
 
     expectObservable(e1).toBe(expected);
   });
+
+  it('should allow unsubscribing early and explicitly', () => {
+    const e1 =   hot('--a--^--b--c---d-| ');
+    const e1subs =        '^        !    ';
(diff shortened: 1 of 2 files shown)


blesh:
LGTM

kwonoj:
While I locally test this PR, seeing below which seems another weirdness of jasmine. I'll check if this can be easily amended and prepare fixup. (and interestingly travis says it's passing while test is failing..)

>Failures:
1) Observable.prototype.publishBehavior should follow the RxJS 4 behavior and NOT allow you to reconnect by subscribing again
  Message:
    ObjectUnsubscribedError thrown
  Stack:

kwonoj:
This https://travis-ci.org/ReactiveX/RxJS/builds/115633896 is travis results shows test failure above.

kwonoj:
OK, merged with 82cd6d9 includes one chore commit simply reorder test case to avoid jasmine weirdness (again). thanks @trxcllnt

Issue #1464, opened 2016-03-13
interval seems unusable after stream fails


ccrowhurstram:
I have a really basic subscription to an `interval`. Once the stream emitted by the `interval` fails, it seems that any further observable's created from an `interval` no longer produces notifications.

**RxJS version:**

5.0.0-beta.2

**Code to reproduce:**

See [this jsbin](https://jsbin.com/razomi/3/edit?html,js,console,output)

**Expected behavior:**

See [this jsbin](https://jsbin.com/regata/3/edit?html,js,console,output) that has exactly the same code except that it uses rxjs 4.0.7

kwonoj:
I've step through bit to see what's happening, and seems scheduler state is not reset if action's thrown, so further attempt to flush does not being executed.

ccrowhurstram:
Thanks @kwonoj for looking into this

blesh:
It's apparently that the `active` flag is never returned to a `false` state if an action throws while executing during `flush`. That renders the scheduler completely inert.

Thanks @ccrowhurstram, this is a valuable find.

staltz:
Maybe this can be closed?

kwonoj:
I think so too. Closing issue - @ccrowhurstram , you may able to try with master (may need some time to publish code as package). Please feel freely reopen if this still problem.

ccrowhurstram:
OK thanks.

I might wait on beta 3 before grabbing the up date (I have the luxury of just playing with rxjs 5 at the moment).

PR #1468, opened 2016-03-14
fix(Schedulers): ensure schedulers can be reused after error in execu…

blesh:
**Description:**

Relates to #1464 

This is *one of two* possible fix PRs for this. This particular one adds an `errorHandler` argument to the `Action.prototype.execute` method to enable error handling and reseting the `active` flag without deoptimizing Scheduler `flush`.

/cc @trxcllnt


blesh:
Closed in favor of #1469

PR #1469, opened 2016-03-14
fix(Schedulers): ensure schedulers can be reused after error in execu…

blesh:
**Description:**

Adds a property to all Actions of `error` that is set if executing the action results in an error. That property is then checked after execution in `flush` methods of schedulers. This is an alternative to #1468 

#1464 

cc/ @trxcllnt

### DIFF
--- a/spec/Scheduler-spec.ts
+++ b/spec/Scheduler-spec.ts
@@ -34,4 +34,25 @@ describe('Scheduler.queue', () => {
       done();
     }, 70);
   });
+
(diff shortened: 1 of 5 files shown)


blesh:
FWIW: neither of the proposed PRs seems to have a big effect on performance, at least at first blush. I'd suspect that _this_ PR has a smaller footprint in memory though, because of fewer function allocations.

cc/ @kwonoj

trxcllnt:
@blesh I like this one better

blesh:
So :+1: then?

Issue #1242, opened 2016-01-23
Docs: Strategy for Guides


rgbkrk:
While I have the enthusiasm for it, I'd love to kick out more docs for RxJS 5. Especially those that are user focused guides with examples. Let me know how and where and I'll start putting out some.

anthonybrown:
Awesome, I'd love to learn more about this awesome library

blesh:
@staltz  is working on updating reactivex.io. Perhaps he can comment here on that effort. Otherwise, we need work done to update code file documentation for the ESDocs we generate that are published to reactivex.io/RxJS.

cc/ @staltz

staltz:
I'm working on updating the basics of reactivex.io (just updating every mention of RxJS to be aligned with RxJS 5's API), which is a cross-platform documentation site. It sorts of limits how we can document the API, because everything is "framed" as cross-platform. For instance http://reactivex.io/documentation/operators/scan.html the structure is: (1) the operator explained in high-level for all languages, (2) RxJS specific variants and a simple example for each variant.

(2) is not really an appropriate space to elaborate on each operator.

I need more suggestions on what would be the **ideal** documentation site for RxJS 5. And then how we can fit that either in reactivex.io or the ESDocs.

Widdershin:
Definitely keen to contribute to some user focused documentation, especially with interactive examples.

staltz:
@Widdershin one thing which I think can be independently built and embedded anywhere is the decision tree https://github.com/Reactive-Extensions/RxJS/blob/master/doc/gettingstarted/which-instance.md in an interactive form. I imagine a series of questions displayed and answering one takes to another question, until you hit the last answer and the conclusion is the operator you should use.

HighOnDrive:
@staltz How about letting the visitor coming to the reactivex.io home page make a choice for what language (JS, Java, etc) they want to view the rest of the site in. This would work similar to how a locale (English, French, German, etc) can be selected on internationalized sites. 

If the visitor chooses RxJS then a given operator page would (1) have the header still explain the operator in high-level for all languages, (2) have everything else on the operator page be displayed in the language the visitor specified. 

This would let the whole operator page be used for the selected language and completely remove the need to dig through more obtrusive menus. The language choice can be stored in a cookie so it is remembered on return visits.

staltz:
[...]

Issue #1060, opened 2015-12-15
Roadmap to release (leaving Beta)


blesh:
This is just a start for a road map:

- [x] automated tests in all platforms
- [ ] update reactivex.io documentation
- [ ] updated migration documentation
- [ ] all operators and public APIs have inline docs
- [ ] PNG marble diagrams generated for every operator
- [ ] improved documentation on Schedulers, what they are and what they do.
- [ ] publish all Jasmine test helpers as separate global and module
- [x] ensure type definition / signature would be correct for typescript-based consumers

kwonoj:
Maybe can add below?

- ensure type definition / signature for typescript interface would be correct for typescript-based consumers (also having logistics to test those)

staltz:
> all operators and public APIs have inline docs

Maybe add "and their respective PNG diagrams" to this?

blesh:
done and done.

blesh:
Given that the checkboxes in this issue are too generic to "check". I'm going to close this issue for now. Each of those items should really be tracked in their own issues.

PR #1471, opened 2016-03-15
docs(manual): write content for scheduler.md

staltz:
**Description:** Write documentation content on Schedulers, for the docs site.

#1060, #1242

### DIFF
--- a/doc/scheduler.md
+++ b/doc/scheduler.md
@@ -1,5 +1,128 @@
 # Scheduler
 
-What is a Scheduler?
+**What is a Scheduler?** A scheduler controls when a subscription starts and when notifications are delivered. It consists of three components.
 
-TODO
+- **A Scheduler is a data structure.** It knows how to store and queue tasks based on priority or other criteria.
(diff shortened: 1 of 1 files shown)


mattpodwysocki:
LGTM.  I'd also include the Virtual Time Scheduler as well as TestScheduler stuff in there as well as those are scheduler instances you can use.  I'd also add a little blurb as to why the schedulers are there such as enabling virtual time.  In addition, if we're going to have a section of testing, we should link to that along with any notion of a virtual time or test scheduler

staltz:
I absolutely agree. The whole Virtual Time Scheduler story is not addressed in this PR but it'll get its own PR. One reason why VT Scheduler was not listed alongside `asap`, `async` and others is that currently even in the API they are not available under `Rx.Scheduler.*`.

mattpodwysocki:
@staltz I guess this is a question then for @trxcllnt then about having things like `Scheduler.requestAnimationFrame` and others here or are they in DOM specific documentation?

staltz:
Hard to tell in which category things fit.

kwonoj:
Merged with 9a9baac, thanks @staltz

Issue #1450, opened 2016-03-10
switch operator test case behavior with promises


kwonoj:
This seems test case design than operator behavior, 

```js
Observable.of<any>(Promise.resolve(1), Promise.reject(2), Promise.resolve(3))
      .switch()
      .subscribe((x: number) => {
        expect(x).toBe(3);
      }, (err: any) => {
        expect(err).toBe(2); //this is not actually asserted
      }, () => {
        done();
      });
```

seems promise elements are being emitted before either it's being resolved or rejected. (and if `error()` is actually being called, `complete` should not be called)

@staltz , is this expected test design? (asking you since it was added with your commit looking at [history](https://github.com/ReactiveX/RxJS/commit/57ad9861baf722fb7c5d07858faf70ed4354eab8#diff-7d20220faf9ec300584166298c9da813R155))

staltz:
That test looks wrong, I wonder why I wrote it like that.

staltz:
Could have been to make coverage larger.

kwonoj:
Closed via https://github.com/ReactiveX/rxjs/pull/1472 .

PR #1472, opened 2016-03-15
test(switch): fix odd-looking test with Promises

staltz:

For issue #1450.


kwonoj:
Seems legit to me, I'll check this in around tomorrow if there's any other suggestion.

kwonoj:
Merged with 34a0d3c, thanks @staltz

Issue #1463, opened 2016-03-13
subscribeToResult isObservable check


milutinovici:
2nd if checks if the result is an observable, using instanceof. This check is too strict. It should be enough that result has subsrcribe method. Something like:

``` typescript
 if (result instanceof Observable || result.subscribe !== undefined)
```

kwonoj:
Do you mean [this](https://github.com/ReactiveX/RxJS/blob/aa91475e09caa0fecd929400f74b29273851e859/src/util/subscribeToResult.ts#L21) checker?

trxcllnt:
I know it's annoying, but you should use [`Symbol.observable`](https://github.com/ReactiveX/RxJS/blob/aa91475e09caa0fecd929400f74b29273851e859/src/util/subscribeToResult.ts#L63) to duck-type custom Observables cc: @blesh @zenparsing

milutinovici:
@kwonoj yes, that's the one.

@trxcllnt I haven't used Symbol before, but looking at the code, I should do something like
```typescript
const custom = createCustomObservable();
custom[SymbolShim.observable] = () => custom;
```
But it's not working.

blesh:
We went with the instanceof check for perf reasons. In some of the resource constrained devices Netflix has to deal with, property checking is much slower.

I think it's probably a reasonable compromise to have an `instanceof` check *first* (since it will pass in most cases), then have a `typeof result.subscribe === 'function'` check somewhere after that. The only thing is, I think it will be much more common to return an Array or Promise in there before a custom "subscribable", or whatever. Perhaps that could be checked at the same time as custom observable (ala `Symbol.observable`).

Really though, `Symbol.observable` is what you want to do. It's got a guaranteed signature that should accept an Observer. Something that's "subscribable" in Rx terms, would have to accept not only three functions, but optionally an Observer.

milutinovici:
> I think it's probably a reasonable compromise to have an instanceof check first (since it will pass in most cases), then have a typeof result.subscribe === 'function' check somewhere after that.

I'm fine with that solution.

> Really though, Symbol.observable is what you want to do.

I'm having a bit of trouble getting that to work. I'll dig a bit more to see why.

milutinovici:
Well with beta 3, Symbol.observable just started to work. Although, I still think that object having subscribe function should be enough.

blesh:
closing as inactive.

PR #1475, opened 2016-03-15
fix(Observable): introduce Subscribable interface to use instead of Observable in input arguments

vladima:
public surface of `Subscribable` is much smaller than in `Observable` which significantly reduces space to explore in when doing `type-related-to`  checks in TypeScript compiler.

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -1,4 +1,4 @@
-import {PartialObserver} from './Observer';
+import {PartialObserver, Observer} from './Observer';
 import {Operator} from './Operator';
 import {Subscriber} from './Subscriber';
 import {Subscription} from './Subscription';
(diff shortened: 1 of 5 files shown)


blesh:
This sounds like a better change. How do the builds perform with this change?

blesh:
Also, for my own reference, this is semi-related to #1463 .. but only in the typings aspect.

vladima:
Perf is roughly on par with tag properties

vladima:
@blesh can you please tell if there is anything else that needs to be done on this PR?

blesh:
@vladima sorry, I've been sick and busy catching up from being sick...

blesh:
This LGTM.

vladima:
Great! Does it mean that it will be available in the next beta release of RxJS?

kwonoj:
> Does it mean that it will be available in the next beta release of RxJS?

: I believe so. I'll double check and check this in around today or tomorrow, can be in time before next release freeze hits.

kwonoj:
Merged with 2256e7b, great appreciate for effort @vladima ! :tada:

PR #1477, opened 2016-03-16
docs(reference): hide irrelevant Observable subclasses

staltz:
**Description:**
Transient Observables like ErrorObservable are made hidden by this commit. This cleans up the
documentation in Observable listing all its subclasses. Transient Observable subclasses are
irrelevant in the documentation since they cannot be directly used by the API.

## Before:

screen shot 2016-03-16 at 17 16 57

## After:

screen shot 2016-03-16 at 17 14 08

#1060, #1242

### DIFF
--- a/src/Subject.ts
+++ b/src/Subject.ts
@@ -9,6 +9,9 @@ import {rxSubscriber} from './symbol/rxSubscriber';
 import {throwError} from './util/throwError';
 import {ObjectUnsubscribedError} from './util/ObjectUnsubscribedError';
 
+/**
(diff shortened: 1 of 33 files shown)


kwonoj:
Change looks good to me. (Wish github reaction includes 'looks good to me' icon too!)

staltz:
We could agree that thumbs up is LGTM.

blesh:
LGTM

kwonoj:
Merged with d9855e2, thanks @staltz :)

PR #1479, opened 2016-03-17
test(deferObservable): add type definition test

kwonoj:
**Description:**

This PR adds type definition test for `deferObservable`, as well as amending parameter type to allow function can throw without return value.

**Additional**

Issue at Typescript (https://github.com/Microsoft/TypeScript/issues/1042) will allow non-void function can throw once it's arrived.

### DIFF
--- a/spec/observables/defer-spec.ts
+++ b/spec/observables/defer-spec.ts
@@ -1,5 +1,5 @@
 import * as Rx from '../../dist/cjs/Rx';
-declare const {hot, expectObservable, expectSubscriptions};
+declare const {hot, expectObservable, expectSubscriptions, type};
 
 const Observable = Rx.Observable;
 
@@ -40,9 +40,9 @@ describe('Observable.defer', () => {
 
   it('should create an observable when factory throws', () => {
     //type definition need to be updated
-    const e1 = Observable.defer(<any>(() => {
+    const e1 = Observable.defer(() => {
       throw 'error';
-    }));
+    });
     const expected = '#';
 
     expectObservable(e1).toBe(expected);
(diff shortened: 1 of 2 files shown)


spec/observables/defer-spec.ts
```diff
@@ -72,4 +72,16 @@ describe('Observable.defer', () => {
     expectObservable(e1, unsub).toBe(expected);
     expectSubscriptions(source.subscriptions).toBe(sourceSubs);
   });
+
+  type(() => {
```
blesh: This might be the wrong time to point this out... but I have NO IDEA what these tests do. I find them almost unreadable.

kwonoj: I don't think there's wrong time, or even it'd be great time since we don't have much test for this yet. :)

Basically this test is ensuring type definition is 'works' as it's actual functionality does, for instance in case of this one `defer`, type should support cases

- when factory function returns `observable<T>`, subscription should carry over type T so you can access its properties without type errors
- factory function can throw without return anything, so defer should accept function throws but does not return

both of declaration in line 82 / 83 illustrates those usecases, and since it's compile time it is being verified when test case is being compiled. If those are not working as intended, compiler will throw out messages of expected type vs. provided one.

I also agree it could be possibly confusing to consume this kind of test (small one would be ok, but case like this? https://github.com/ReactiveX/RxJS/pull/1189/files) and it'd be great if there's suggestion how to improve this or requirement should be added for better visibility. Since there is barely no test yet, changes can be easily introduced.

kwonoj:
Let me not check this in to have better proposed PR for typing tests. /cc @david-driscoll also assume he might be interested as well.

spec/observables/defer-spec.ts
blesh: it seems like it would be better just to work that into the TypeScript of our actual tests. Since those will cause the compilation of the spec file themselves to fail and thus cause the tests to fail. Do we really need an additional test that doesn't seem to do anything once it's transpiled?

kwonoj: feasible opinion. In most cases test cases already consumes type definitions, so probably won't need to be covered once again via explicit. I remember few cases that test case does not include all of possible type definition cases though - start with opt out and only add necessary cases once it's being noticed maybe better way to go?

kwonoj: overloaded signature like combinelatest (#1189) might be few edge cases need these kind of test cases.

kwonoj:
I have closed this PR without merging in favor of https://github.com/ReactiveX/RxJS/pull/1484.

PR #1480, opened 2016-03-17
docs(manual): update image references

kwonoj:
**Description:**

It is not clearly known, but github seems prevent to render svg content in repo as markdown content, serves as plain text directly. (http://stackoverflow.com/a/21521184 / https://github.com/isaacs/github/issues/316)
This PR updates image reference to use rawgit (https://rawgit.com/) serves correct content type.

Maybe just replace image to non-svg would be better solution? /cc @staltz

### DIFF
--- a/doc/operators.md
+++ b/doc/operators.md
@@ -96,7 +96,7 @@ To explain how operators work, textual descriptions are often not enough. Many o
 
 Below you can see the anatomy of a marble diagram.
 
-<img src="./asset/marble-diagram-anatomy.svg">
+![Anatomy of marble diagrams](https://cdn.rawgit.com/ReactiveX/RxJS/master/doc/asset/marble-diagram-anatomy.svg)
 
 Throughout this documentation site, we extensively use marble diagrams to explain how operators work. They may be really useful in other contexts too, like on a whiteboard or even in our unit tests (as ASCII diagrams).
 


staltz:
Oh, wait a minute. That SVG is primarily used for the ESDoc website, and works normally there. While it **is** possible to open the `docs/` directory in GitHub and just read it from here, the recommended docs should be reactivex.io/rxjs. This is normal also in Cycle.js docs https://github.com/cyclejs/cyclejs.github.io/blob/master/_posts/2015-01-30-dialogue.md and also in React docs https://github.com/facebook/react/blob/master/docs/docs/thinking-in-react.md

staltz:
And it's important that the image is SVG, for good and scalable resolution, but specially because the SVG file is text/code which can be edited. I.e. no one needs some "original" Illustrator file if they want to edit the image. It's open source.

kwonoj:
This PR's not supposed to check in immediately, wanted to confirm its behavior. So it's ok to leave as-is? I thought it's expected to display some sort of images, (may other ppl could think same if they visit github only)

staltz:
Let's make sure the see a link to http://reactivex.io/rxjs

kwonoj:
OK, let me close this PR without merging, thanks for clarification and sorry for inconvenience. Totally my misunderstanding.

Issue #1483, opened 2016-03-17
Update static operators to accept promises across the board


trxcllnt:
A lot of this is my fault, but given that we've decided to interop with promises in most places, we should probably ensure we do it consistently. That means updating the family of static `defer` creation operators (defer, if, using, etc.) to accept promises, lowercase-o observables, etc. to use our internal `subscribeToResult` function.

kwonoj:
Is this related to https://github.com/ReactiveX/RxJS/issues/1246?

trxcllnt:
@kwonoj yes it is

blesh:
Really, there shouldn't be anything public facing that "accepts" `Observable<T>`, it should always accept `SubscribableOrPromise<T>`. Or at the very least `Subscribable<T>`. Having a function that accepts a strict class type is a 101 mistake, we should have been using interfaces this whole time.

blesh:
We should probably make separate issues for the operators in question.

staltz:
Can I remind people that if future PRs change the signature of an operator, please remember to update also the docs that I wrote for them. Otherwise docs will be a moving target

kwonoj:
I think this one's covered mostly, can this be closed maybe? (Assume `ConnectableObservable` is not fall into this?)

kwonoj:
I think most static operators are now covered as commented, /cc @blesh for visibility before closing this out and switching to open issues for each individuals, in case I miss something should be covered before closing.

kwonoj:
Closing this, if there's remaining operator will file separate.

PR #1484, opened 2016-03-18
fix(deferObservable): accepts factory returns promise

kwonoj:
**Description:**

I came to notice PR https://github.com/ReactiveX/RxJS/pull/1479 becomes redundant by issue https://github.com/ReactiveX/RxJS/issues/1483 filed after, allowing promise returned by factory function. I have closed #1479 and create this one instead to reflect those.

- factory function now can return either `Observable` or `Promise`
- deferObservable internally utilizes `subscribeToResult`
- as suggested in #1479, no longer include duplicated type definition test but test case reflects definition usecases

relates to #1483

** Notes **

Not sure if this change's intended direction in #1483. /cc @trxcllnt to ask review if this change makes sense.


kwonoj:
Merged with 0cb44e1 .

PR #1473, opened 2016-03-15
docs(operators): write comprehensive docs on delay() and switch()

staltz:
**Description:**
I'm starting to document each operator in details and this first PR is just to demonstrate my general approach to document an operator. There are two operators documented here: `delay` and `switch`. The structure is:

1. **Short headline** describing the operator
2. **Short "informal" description** of the operator. Think of this as "the operator explained for dummies". Rx4 docs suffered from overly-technical and accurate descriptions which made it hard for normal humans to parse in their brains. I'm trying to provide both formal and informal descriptions.
3. **Marble diagram**
4. **Longer formal/technical description**
5. **Signature**: params and return
6. **Example code** that actually makes sense in the real world
7. **Link to the tests**

Here is how `delay` and `switch` look like. @blesh @trxcllnt @jhusain your feedback here is important, whether this structure is good. If it's good, I'll just go ahead and do this for all operators.

screen shot 2016-03-15 at 17 24 34
  
screen shot 2016-03-15 at 17 25 55

**Related issue (if exists):**
#1060, #1242

PR #1485, opened 2016-03-18
docs(operators): write comprehensive JSDoc on the buffer operators

staltz:
Merge after PR #1473 is merged

### DIFF
--- a/src/operator/switch.ts
+++ b/src/operator/switch.ts
@@ -6,20 +6,43 @@ import {InnerSubscriber} from '../InnerSubscriber';
 import {subscribeToResult} from '../util/subscribeToResult';
 
 /**
- * Converts an Observable that emits Observables into an Observable that emits the items emitted by the most recently
- * emitted of those Observables.
+ * Converts a higher-order Observable into a first-order Observable by only the
+ * most recently emitted of those nested Observables.
+ *
+ * <span class="informal">Flattens an Observable-of-Observables by dropping the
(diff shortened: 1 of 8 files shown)


src/operator/switch.ts
```diff
@@ -6,20 +6,43 @@ import {InnerSubscriber} from '../InnerSubscriber';
...
- * is emitted, switchOnNext stops emitting items from the earlier-emitted Observable and begins emitting items from the
- * new one.
+ * `switch` subscribes to an Observable that emits Observables,
+ * also known as a higher-order Observable. Each time it observes one of these
+ * emitted nested Observables, the output Observable begins emitting the items
+ * emitted by that nested Observable. So far, it behaves like {@link mergeAll}.
+ * However, when a new nested Observable is emitted, `switch` stops emitting
+ * items from the earlier-emitted nested Observable and begins emitting items
+ * from the new one. It continues to behave like this for subsequent nested
+ * Observables.
```
kwonoj: Will it makes sense to mention `switch` also handles subscription when switching occurs? I remember someone have asked if it's needed to explicitly manage inner subscription after switching.

staltz: Good point. But let's talk about this in PR https://github.com/ReactiveX/RxJS/pull/1473 which is the one that has the commit for switch.

kwonoj: Cool :)

kwonoj:
Merged with 56c9191, thanks @staltz

PR #1490, opened 2016-03-18
fix(windowToggle): handle closingSelector completes immediately

### LINKED ISSUE #1487: windowToggle bug in WindowToggleSubscriber#closeWindow
**RxJS version:** 5.0.0-beta.2

**Code to reproduce:**
```js
var clicks = Rx.Observable.fromEvent(document, 'click');
var openings = Rx.Observable.interval(1000);
var result = clicks.windowToggle(openings, i =>
  i % 2 ? Rx.Observable.interval(500) : Rx.Observable.empty()
).mergeAll();
result.subscribe(x => console.log(x));
```

**Expected behavior:**
No error thrown, some messages in the console when the user clicks.

**Actual behavior:**
`Uncaught TypeError: Cannot read property 'window' of undefined` after 1 second.

**Additional information:**
Seems to happen in `WindowToggleSubscriber#closeWindow`.

JSBin: http://jsbin.com/zenelokita/1/edit?html,js,output

kwonoj:
**Description:**

This PR handles cases where `closingSelector` returns immediately completing observables (i.e `emptyObservable`) which invokes `notifyComplete` synchronously, not able to close constructed window by not having its context. 

closes #1487

This may not be ideal fixes for issue, any suggestion would be appreciated. /cc @staltz also for visibility & review.

### DIFF
--- a/src/operator/windowToggle.ts
+++ b/src/operator/windowToggle.ts
@@ -131,10 +131,16 @@ class WindowToggleSubscriber<T, O> extends OuterSubscriber<T, any> {
         const context = { window, subscription };
         this.contexts.push(context);
         const innerSubscription = subscribeToResult(this, closingNotifier, context);
-        (<any> innerSubscription).context = context;
-        subscription.add(innerSubscription);
+
+        if (innerSubscription.isUnsubscribed) {
+          this.closeWindow(this.contexts.length - 1);
+        } else {
+          (<any> innerSubscription).context = context;
+          subscription.add(innerSubscription);
+        }
 
         this.destination.next(window);
+
       }
     } else {
       this.closeWindow(this.contexts.indexOf(outerValue));
@@ -151,7 +157,11 @@ class WindowToggleSubscriber<T, O> extends OuterSubscriber<T, any> {
     }
   }
 
-  closeWindow(index: number) {
+  private closeWindow(index: number): void {
+    if (index === -1) {
+      return;
+    }
+
     const { contexts } = this;
     const context = contexts[index];
     const { window, subscription } = context;
--- a/spec/operators/windowToggle-spec.ts
+++ b/spec/operators/windowToggle-spec.ts
@@ -407,4 +407,22 @@ describe('Observable.prototype.windowToggle', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
     expectSubscriptions(e2.subscriptions).toBe(e2subs);
   });
+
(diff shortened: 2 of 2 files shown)


src/operator/windowToggle.ts
```diff
@@ -131,10 +131,16 @@ class WindowToggleSubscriber<T, O> extends OuterSubscriber<T, any> {
         const context = { window, subscription };
         this.contexts.push(context);
         const innerSubscription = subscribeToResult(this, closingNotifier, context);
```
staltz: What about this inner subscription to the empty Observable, when is it unsubscribed?

kwonoj: it is immediately unsubscribed when it's empty observable, so synchronously invoke `notifyComplete` and continue execution to line 135. This change's using those, check innersubscription status at that moment - if it's immediately unsubscribed, close window without setting context.

staltz: LGTM

kwonoj:
Appreciate for review @staltz :) I'll leave this bit around if there's any suggestions around.

src/operator/windowToggle.ts
```diff
@@ -131,10 +131,16 @@ class WindowToggleSubscriber<T, O> extends OuterSubscriber<T, any> {
...
-        (<any> innerSubscription).context = context;
-        subscription.add(innerSubscription);
+
+        if (!innerSubscription.isUnsubscribed) {
```
blesh: Can we invert this `if` block so it checks `if(innerSubscription.isUnsubscribed) {`?

kwonoj: Sure.

src/operator/windowToggle.ts
```diff
@@ -151,7 +157,11 @@ class WindowToggleSubscriber<T, O> extends OuterSubscriber<T, any> {
     }
   }
 
-  closeWindow(index: number) {
+  private closeWindow(index: number): void {
+    if (index < 0) {
```
blesh: I think it's more idiomatic to check `index === -1` here. Will this ever be less than `-1`?

blesh:
Couple of small, nitpicky, things. Other than those two things, it looks fine.

src/operator/windowToggle.ts
kwonoj: It's my old habit :) will update it.

blesh:
LGTM, thanks @kwonoj

kwonoj:
Merged with c755587 .

PR #1491, opened 2016-03-18
chore(package.json): update contact address

kwonoj:
**Description:**

Small chore to update contact address (no, I'm still at my work), may not use corp address to ensure this is not corp-related contribution. I'll check this in myself soon, /cc @blesh for visibility.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,7 +128,7 @@
     },
     {
       "name": "OJ Kwon",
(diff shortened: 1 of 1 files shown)


blesh:
I _guess_ this looks like an "okay" change...

kwonoj:
Merged with b3c8ba2, appreciate for reviewing this @blesh :+1:

PR #1494, opened 2016-03-19
Update bufferToggle closingSelector to accept promise / empty

kwonoj:
**Description:**

This PR updates behavior of bufferToggle's `closingSelector`

- accepts closingSelector returns promise
- handle cases where closingSelector returns immediately

relates to https://github.com/ReactiveX/RxJS/issues/1487, https://github.com/ReactiveX/RxJS/issues/1246

### DIFF
--- a/spec/operators/bufferToggle-spec.ts
+++ b/spec/operators/bufferToggle-spec.ts
@@ -1,5 +1,6 @@
 import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
+import {DoneSignature} from '../helpers/test-helper';
 declare const {hot, cold, asDiagram, expectObservable, expectSubscriptions};
 
 const Observable = Rx.Observable;
@@ -341,5 +342,43 @@ describe('Observable.prototype.bufferToggle', () => {
     expectObservable(result).toBe(expected);
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
     expectSubscriptions(e2.subscriptions).toBe(e2subs);
+  });
+
+  it('should accept closing selector that returns a resolved promise', (done: DoneSignature) => {
+    const e1 = Observable.concat(Observable.of(1),
+      Observable.timer(10).mapTo(2),
+      Observable.timer(10).mapTo(3),
+      Observable.timer(100).mapTo(4)
+      );
+    const expected = [[1]];
+
+    e1.bufferToggle(Observable.of(10), () => new Promise((resolve: any) => { resolve(42); }))
+      .subscribe((x) => {
+        expect(x).toEqual(expected.shift()); },
+        done.fail,
+        () => {
+          expect(expected.length).toBe(0);
+          done();
+      });
+  });
+
+  it('should accept closing selector that returns a rejected promise', (done: DoneSignature) => {
+    const e1 = Observable.concat(Observable.of(1),
+      Observable.timer(10).mapTo(2),
+      Observable.timer(10).mapTo(3),
+      Observable.timer(100).mapTo(4)
+      );
+
+    const expected = 42;
+
+    e1.bufferToggle(Observable.of(10), () => new Promise((resolve: any, reject: any) => { reject(expected); }))
+      .subscribe((x) => {
+        done.fail();
+      }, (x) => {
+        expect(x).toBe(expected);
+        done();
+      }, () => {
+        done.fail();
+      });
   });
 });
--- a/spec/operators/bufferToggle-spec.ts
+++ b/spec/operators/bufferToggle-spec.ts
@@ -343,6 +343,48 @@ describe('Observable.prototype.bufferToggle', () => {
     expectSubscriptions(e2.subscriptions).toBe(e2subs);
   });
 
+  it('should accept openings resolved promise', (done: MochaDone) => {
+    const e1 = Observable.concat(
+      Observable.timer(10).mapTo(1),
+      Observable.timer(100).mapTo(2),
+      Observable.timer(150).mapTo(3),
+      Observable.timer(200).mapTo(4));
+
+    const expected = [[1]];
+
+    e1.bufferToggle(new Promise((resolve: any) => { resolve(42); }), () => {
+      return Observable.timer(50);
+    }).subscribe((x) => {
+      expect(x).to.deep.equal(expected.shift());
+    }, (x) => {
+      done(new Error('should not be called'));
+    }, () => {
+      expect(expected.length).to.be.equal(0);
+      done();
+    });
+  });
+
+  it('should accept openings rejected promise', (done: MochaDone) => {
+    const e1 = Observable.concat(Observable.of(1),
+      Observable.timer(10).mapTo(2),
+      Observable.timer(10).mapTo(3),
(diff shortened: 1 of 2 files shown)


spec/operators/publishReplay-spec.ts
```diff
@@ -23,6 +23,34 @@ describe('Observable.prototype.publishReplay', () => {
     expect(source instanceof Rx.ConnectableObservable).toBe(true);
   });
 
+  it('should follow the RxJS 4 behavior and NOT allow you to reconnect by subscribing again', (done: DoneSignature) => {
```
kwonoj: simple test case reordering again, due to jasmine behavior. It seems occuring repeatedly now.

spec/operators/bufferToggle-spec.ts
```diff
@@ -341,4 +342,54 @@ describe('Observable.prototype.bufferToggle', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
     expectSubscriptions(e2.subscriptions).toBe(e2subs);
   });
+
+  it('should accept closing selector returns promise resolves', (done: DoneSignature) => {
```
staltz: I would add the word `that` here. "should accept closing selector **that** returns **a resolved** promise"

spec/operators/bufferToggle-spec.ts
```diff
@@ -341,4 +342,54 @@ describe('Observable.prototype.bufferToggle', () => {
...
+      .subscribe((x) => {
+        expect(x).toEqual(expected.shift()); },
+        done.fail,
+        () => {
+          expect(expected.length).toBe(0);
+          done();
+      });
+  });
+
+  it('should accept closing selector returns promise rejects', (done: DoneSignature) => {
```
staltz: I would add the word `that` here. "should accept closing selector **that** returns **a resolved** promise"

staltz:
Can we merge this after #1485, and then update this PR to document the use of Promises?

kwonoj:
Since #1485 is ready to check in (plan to do it in morning) and this is still need to be reviewed, order should be as you suggested. I'll update PR once check in #1485.

spec/operators/bufferToggle-spec.ts
kwonoj: Sure will do.

src/operator/bufferToggle.ts
```diff
@@ -44,18 +46,18 @@ import {errorObject} from '../util/errorObject';
  * @owner Observable
  */
 export function bufferToggle<T, O>(openings: Observable<O>,
```
staltz: I'm not sure, but can't `openings` also be a Promise?

src/operator/bufferToggle.ts
```diff
@@ -44,18 +46,18 @@ import {errorObject} from '../util/errorObject';
...
-                                   closingSelector: (value: O) => Observable<any>): Observable<T[]> {
+                                   closingSelector: (value: O) => SubscribableOrPromise<any> | void): Observable<T[]> {
```
staltz: Why `| void`?

src/operator/bufferToggle.ts
```diff
@@ -35,8 +37,8 @@ import {errorObject} from '../util/errorObject';
  *
  * @param {Observable<O>} openings An observable of notifications to start new
  * buffers.
- * @param {function(value: O): Observable} closingSelector A function that takes
- * the value emitted by the `openings` observable and returns an Observable,
+ * @param {function(value: O): SubscribableOrPromise} closingSelector A function that takes
+ * the value emitted by the `openings` observable and returns an Subscribable or Promise,
```
staltz: Can you keep the column limited under 80 characters? Also `an Subscribable` => `a Subscribable`.

src/operator/bufferToggle.ts
kwonoj: I'm not sure either, so limited changes for sure thing first.

src/operator/bufferToggle.ts
kwonoj: if selector throws in one execution branch like
```
if (...) {
throw  
}
```
 
then compiler will complain there isn't any return value in function. There's open issue to support this(https://github.com/Microsoft/TypeScript/issues/1042), until then without this typescript user can't throw or have to cast it into `any` explicitly.

kwonoj:
I think change's are ok, but will leave bit more to see any other suggestions around before check this in.

src/operator/bufferToggle.ts
```diff
@@ -44,18 +46,18 @@ import {errorObject} from '../util/errorObject';
  * @owner Observable
  */
```
staltz: We need to update docs in line 22 to say "an Observable or Promise returned by...".
And update docs in line 40 too.

src/operator/bufferToggle.ts
blesh: Yes, anywhere we're accepting an `Observable` we should accept any of the acceptable types. Although synchronous things like iterators and Arrays don't make sense here. I think we should probably accept `Promise` though.

Basically, we need to accept interfaces of `Promise`, `Subscribable` and `IObservable` here. (We probably also need better names for those last two, haha)

src/operator/bufferToggle.ts
blesh: That seems like an issue with TypeScript, not us, right?

kwonoj: > That seems like an issue with TypeScript, not us, right?

: Yes it is. Once typescript introduce those changes, this can go away.

src/operator/bufferToggle.ts
kwonoj: OK, will try to update. Would it be ok to accept `SubscribableOrPromise` as same as closing?

src/operator/bufferToggle.ts
blesh: I wouldn't even add this in our library. If it's an issue with typescript, it's an issue with typescript. We might not remember to remove it, and returning `void` here is invalid.

kwonoj: So let typescript user won't be able to throw in selector functions by out of box and amend it for their own? (via `<any>` casting).

kwonoj: /cc @david-driscoll - I remember we already have couple of signature followed this pattern. May need to update them as well to clear.

blesh: It's going to allow/encourage TypeScript users to use this method incorrectly. `void` is not an okay return here. They have to return an `Observable` if they don't throw. If we must, we must, but that's _really_ gross.

kwonoj: Got point, agreed. I'll remove void signature and will look for other places as well if it does have.

blesh:
Looks like this could use a rebase.

kwonoj:
Absolutely, and this PR still need to be updated per suggestion - I'll update it soon.

kwonoj:
PR's updated per suggestion also rebased. I'll leave this PR around day before check in to see any other suggestions around.

kwonoj:
Merged with 3d22c7a.

PR #1500, opened 2016-03-21
docs(interfaces): document important interfaces and classes

staltz:
Write JSDoc documentation for Observer, Subscriber, Subscription, Notification, Scheduler, Action,
etc.

### DIFF
--- a/spec/schedulers/AsapScheduler-spec.ts
+++ b/spec/schedulers/AsapScheduler-spec.ts
@@ -3,8 +3,8 @@ import * as Rx from '../../dist/cjs/Rx.KitchenSink';
 
 const asap = Rx.Scheduler.asap;
 
-/** @test {AsapScheduler} */
-describe('AsapScheduler', () => {
+/** @test {Scheduler} */
+describe('Scheduler.asap', () => {
   it('should exist', () => {
     expect(asap).exist;
   });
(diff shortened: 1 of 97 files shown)


src/Observer.ts
```diff
@@ -21,6 +21,68 @@ export interface CompletionObserver<T> {
...
+ * call the Observer's `next(value)` method to provide notifications. A
+ * well-behaved Observable will call an Observer's `complete()` method exactly
+ * once or the Observer's `error(err)` method exactly once, as the last
+ * notification delivered.
+ *
+ * @interface
+ * @name Observer
+ * @noimport true
+ */
+export class ObserverDoc<T> {
```
blesh: interesting, does this build as part of the output? That wouldn't be ideal.

Too bad there isn't a way to document interfaces. Perhaps we add some files to the build that are purely for documenting interfaces and build them separately?

staltz: Yeah it's ugly, I know.

> Perhaps we add some files to the build that are purely for documenting interfaces and build them separately?

We can do that. The real reason is that TypeScript strips away JSDoc comments for interfaces. And I won't try another PR to TS in vain if they are so... 'courteous'.

kwonoj: Agree it's pain to deal with comment strip in typescript with docs. :/

staltz:
I rebased and also moved interface documentations to files `MiscJSDoc.ts` which are not compiled to any release bundle, but appear in the ESDoc site.

Can haz merge now?

cuteness

blesh:
LGTM.

staltz:
Rebased. :white_check_mark:

kwonoj:
Merged with 8da7a8f, thanks @staltz :)

PR #1449, opened 2016-03-10
chore(code-coverage): replace code coverage checker to codecov

kwonoj:
**Description:**

This PR migrates current code coverage checker (coveralls) into codecov.io (https://codecov.io/)
There were couple of known failures (https://github.com/ReactiveX/RxJS/issues/1304), (https://github.com/ReactiveX/RxJS/issues/1440), codecov seems more reliably works in general.

To complete migrate, repo owner need to signup at codecov, add this repo to be monitored.

While I'm bit on fence to actually check this in or not, would like to use this PR as discussion point.

**Related issue (if exists):**

closes #1440

PR #1501, opened 2016-03-21
feat(ajax): add FormData support in AjaxObservable and add percent encoding for parameter key in body

vvakame:
**Description:**

add FormData support in AjaxObservable.
I want to upload files by AjaxObservable.

https://developer.mozilla.org/en-US/docs/Web/API/FormData/Using_FormData_Objects
http://caniuse.com/#search=FormData

I can't find test for AjaxSubscriber#serializeBody.
https://github.com/ReactiveX/RxJS/blob/master/spec/observables/dom/ajax-spec.ts

None.

### DIFF
--- a/src/observable/dom/AjaxObservable.ts
+++ b/src/observable/dom/AjaxObservable.ts
@@ -245,6 +245,8 @@ export class AjaxSubscriber<T> extends Subscriber<Event> {
   private serializeBody(body: any, contentType: string) {
     if (!body || typeof body === 'string') {
       return body;
+    } else if (root.FormData && body instanceof root.FormData) {
+      return body;
     }
 
     const splitIndex = contentType.indexOf(';');
@@ -254,7 +256,7 @@ export class AjaxSubscriber<T> extends Subscriber<Event> {
 
     switch (contentType) {
       case 'application/x-www-form-urlencoded':
-        return Object.keys(body).map(key => `${key}=${encodeURI(body[key])}`).join('&');
+        return Object.keys(body).map(key => `${encodeURI(key)}=${encodeURI(body[key])}`).join('&');
       case 'application/json':
         return JSON.stringify(body);
     }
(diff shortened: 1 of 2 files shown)


src/observable/dom/AjaxObservable.ts
```diff
@@ -245,6 +245,8 @@ export class AjaxSubscriber<T> extends Subscriber<Event> {
   private serializeBody(body: any, contentType: string) {
     if (!body || typeof body === 'string') {
       return body;
+    } else if (typeof FormData !== 'undefined' && body instanceof FormData) {
```
blesh: this should check if `root.FormData` rather than just `FormData`. Also, the `typeof` isn't really necessary, a simple `FormData && body instanceof FormData` will work.

blesh:
@vvakame, this looks like a solid change on the surface. Can you please add tests to cover this scenario?

cc @trxcllnt for additional review.

src/observable/dom/AjaxObservable.ts
tbranyen: I think the check is to ensure FormData is not `undefined`. If it's `undefined`, the instanceof check will fail.

vvakame:
@blesh @tbranyen thank you for reviewing!
I was push additional commit. please check again.
I found another issue. I created patch to it too.

need squash & push -f ?

kwonoj:
Yes, you can fixup & squash commits with update commit comments accordingly. (brief description in commit message, add details in body)

vvakame:
okay, I'll working for it.

BTY, travis-ci failed in master/HEAD :crying_cat_face: https://travis-ci.org/ReactiveX/rxjs/branches

kwonoj:
> BTY, travis-ci failed in master/HEAD

Yes, sorry for that - it's not your fault. I opened up PR https://github.com/ReactiveX/rxjs/pull/1449 for resolve CI failure issue due to code coverage checker.

kwonoj:
Change looks good to me.

blesh:
Merged with d03b226, after I fixed an unrelated issue with a bad test.

Thank you so much for the contribution @vvakame! It's very much appreciated!

PR #1510, opened 2016-03-22
If, Using operator accepts promise

kwonoj:
**Description:**
Continue effort as same as https://github.com/ReactiveX/rxjs/pull/1484, this PR updates `using` and `if` to accept promise for factory functions.

relates to #1483


kwonoj:
I think change's are ok, but will leave bit more to see any other suggestions around before check this in.

blesh:
LGTM... merging now.

blesh:
Merged with f8d7d1b, thanks @kwonoj

PR #1509, opened 2016-03-21
chore(package.json): change typing update script execution lifecycle

kwonoj:
**Description:**

There isn't provided way to override `install` script itself (and strongly not recommended), have to rely on lifecycle scripts. `prepublish` is more appropriate lifecycle hook for this issue (https://docs.npmjs.com/misc/scripts) which expected to be triggered after local dev system's `install`. though name's somewhat not clear, `prepublish` is expected to chained after `npm install` (parameterless install) (https://github.com/npm/npm/issues/3059#issuecomment-12282746)

**Related issue (if exists):**

closes #1504

PR #1511, opened 2016-03-22
chore(package.json): remove redundant script

kwonoj:
**Description:**
This PR removes redundant `postInstall` script after https://github.com/ReactiveX/rxjs/pull/1509 is introduced.


kwonoj:
Merged with 1104af1.

PR #1512, opened 2016-03-22
refactor(audit): rename from 'inspect' to 'audit'

### LINKED ISSUE #1505: Rename `inspect` to `audit`
Relates to discussed issue here #1387. Basically, `inspect` is a name-collision in Node because it's an API that should have been a `Symbol`, but couldn't have been because it predates Symbols. /shrug

Anyhow, since the `inspect` operator is brand new with this major version, we can rename it. I think `audit` is a fine name, props to @kennethlawrence for the idea.

Labeling priority `high` because we don't want any breaking changes to existing APIs after production release.

saneyuki:
Fix #1505


kwonoj:
change looks good to me.

saneyuki:
rebased.

kwonoj:
I'll check this in around today if there isn't specific suggestion around. Since this is name change only, I think it'll be ok.

kwonoj:
Merged with 9630daa, thanks @saneyuki

Issue #1513, opened 2016-03-22
Add a subscription argument to next?


zenparsing:
I think we should consider adding an additional "subscription" argument to `observer.next`.

(See https://github.com/zenparsing/es-observable/pull/84 for the spec changes.)

Consider the following:

``` js
// Synchronously emits: 1, 2, 3, done
let observable = new Rx.Observable(sink => {
    sink.next(1);
    sink.next(2);
    sink.next(3);
    sink.complete();
});

observable.first(x => {
    console.log(`First called with: ${ x }`);
    return true;
}).subscribe({
    next(x) { 
        console.log(`Observer called with: ${ x }`);
    }
});
```

Output:

``` js
'First called with: 1'
'Observer called with: 1'
'First called with: 2'
'First called with: 3'
```

Expected output:

``` js
'First called with: 1'
'Observer called with: 1'
```

Looking past the implementation details of Rx's `first`, the general issue here is that it's tricky to implement correct termination semantics in the face of synchronous emission.  The operator wants to unsubscribe in response to a `next` call, but can't because the `subscription` hasn't been made available yet.  This also comes up in the implementation of `forEach`, when we want to unsubscribe if the user callback throws an error.

To correctly handle the synchronous case, you currently have to do something like track whether `subscribe` has returned yet and handle that case differently within the `next` handler.

The proposal is to provide the subscription as a second argument to the observer's next method.  The safe observer automatically adds the parameter when calling through to the consuming observer.  The implementation of `first becomes something like:

``` js
function first(pred) {
    return new Observable(sink => {
        return this.subscribe({
            next(x, subscription) { 
                let take = false;

                try { take = Boolean(pred(x)) } 
                catch (e) { return sink.error(e) }

                if (take) {
                    subscription.unsubscribe();
                    return sink.next(x);
                }
            },
            error(e) { return sink.error(e) },
            complete() { return sink.complete() },
        });
    });
}
```

Thoughts?

trxcllnt:
@zenparsing the first subscriber should be completing + unsubscribing upon the first event that passes its predicate test, thus setting the subscriber state to ignore future next values, regardless of whether the source next's again or not. If this is not the case we have a bug :/

zenparsing:
@trxcllnt Right, it's a bug.  In general though, it's not trivial to write bug-free combinators like this which might need to unsubscribe before the `subscribe` call has completed.  That's why I think it might be helpful to provide the subscription as a second argument to "next".

[...]

PR #1516, opened 2016-03-22
fix(observable): ensure the subscriber chain is complete before calling this._subscribe

trxcllnt:
**Description:**
If an operator subscriber completed (or unsubscribed) early after receiving values from a synchronous source, the subscribers upstream weren't disposed until after returning from the source's subscribe function. When any downstream subscriber unsubscribes, the upstream subscribers should be notified immediately and unsubscribe all the way to the source subscription.

@zenparsing pointed this out in #1513

cc: @blesh

### DIFF
--- a/spec/Subject-spec.ts
+++ b/spec/Subject-spec.ts
@@ -356,11 +356,12 @@ describe('Subject', () => {
 
     subject.subscribe(function (x) {
       expect(x).toBe(expected.shift());
-    }, null, done);
+    });
 
     subject.next('foo');
     subject.complete();
-    subject.next('bar');
+    expect(() => subject.next('bar')).toThrow(new Rx.ObjectUnsubscribedError());
+    done();
   });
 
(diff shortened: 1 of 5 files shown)


spec/Subject-spec.ts
```diff
@@ -353,14 +353,21 @@ describe('Subject', () => {
...
 
     subject.subscribe(function (x) {
       expect(x).toBe(expected.shift());
-    }, null, done);
+    });
 
     subject.next('foo');
     subject.complete();
-    subject.next('bar');
+    try {
```
blesh: This should probably be:

```js
expect(() => subject.next('bar')).toThrow(new Rx.ObjectUnsubscribedError())
```

Or something like that.

blesh:
once that one small change to the tests is in, LGTM.

blesh:
merged with e66b2d8, thanks @trxcllnt

kwonoj:
SHA seems 1631224 .

PR #1517, opened 2016-03-22
chore(travis): move coverage checker out of script step

kwonoj:
**Description:**
Alternative to https://github.com/ReactiveX/rxjs/pull/1449, make code coverage checker optional.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ script:
   - npm run build_spec && npm run test_jasmine && node ./node_modules/markdown-doctest/bin/cmd.js
   - npm run check_circular_dependencies
   - npm run cover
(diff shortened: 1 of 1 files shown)


kwonoj:
OK, this PR makes code coverage checker as optional step in CI. @blesh , which way you prefer? (https://github.com/ReactiveX/rxjs/pull/1449, replacing to different checker vs. this one). I'll check in one and close other based on suggestion.

One thing to be noted - if coverage checker fails, there won't be latest coverage information available which is happening now - https://coveralls.io/github/ReactiveX/RxJS stopped to gather code coverage since it starts failing.

kwonoj:
Merged with 7782a8f.

PR #1519, opened 2016-03-23
feat(timestamp): add timestamp operator

### LINKED ISSUE #1515: timestamp operator
Hey, I just wanted to show some love to the timestamp operator and make sure it's in the todo list.
Thanks!

kwonoj:
**Description:**
This PR implements `timestamp` operator in RxJS 4. As same as timeinterval, it's placed under `kitchenSink`.

closes #1515

### DIFF
--- a/spec/operators/timestamp-spec.ts
+++ b/spec/operators/timestamp-spec.ts
@@ -1,0 +1,147 @@
+import * as Rx from '../../dist/cjs/Rx.KitchenSink';
+declare const {hot, cold, asDiagram, expectObservable, expectSubscriptions};
+
+declare const rxTestScheduler: Rx.TestScheduler;
+const Observable = Rx.Observable;
+
+/** @test {timestamp} */
+describe('Observable.prototype.timestamp', () => {
+  asDiagram('timestamp')('should record the time stamp per each source elements', () => {
+    const e1 =   hot('-b-c-----d--e--|');
+    const e1subs =   '^              !';
+    const expected = '-w-x-----y--z--|';
+    const expectedValue = { w: 10, x: 30, y: 90, z: 120 };
+
+    const result = e1.timestamp(rxTestScheduler)
+      .map(x => x.timestamp);
+
+    expectObservable(result).toBe(expected, expectedValue);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should record stamp if source emit elements', () => {
+    const e1 = hot('--a--^b--c----d---e--|');
+    const e1subs =      '^               !';
+    const expected =    '-w--x----y---z--|';
+
+    const expectedValue = {
+      w: new Rx.Timestamp('b', 10),
+      x: new Rx.Timestamp('c', 40),
+      y: new Rx.Timestamp('d', 90),
+      z: new Rx.Timestamp('e', 130)
+    };
+
+    expectObservable(e1.timestamp(rxTestScheduler)).toBe(expected, expectedValue);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should completes without record stamp if source does not emits', () => {
+    const e1 =   hot('---------|');
+    const e1subs =   '^        !';
+    const expected = '---------|';
+
+    expectObservable(e1.timestamp(rxTestScheduler)).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should complete immediately if source is empty', () => {
+    const e1 =  cold('|');
+    const e1subs =   '(^!)';
+    const expected = '|';
+
+    expectObservable(e1.timestamp(rxTestScheduler)).toBe(expected);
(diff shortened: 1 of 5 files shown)


src/add/operator/timestamp.ts
```diff
@@ -0,0 +1,10 @@
+import {Observable} from '../../Observable';
+import {timestamp, TimeStampSignature} from '../../operator/timestamp';
+
+Observable.prototype.timestamp = timestamp;
```
kwonoj: This is following RxJS 4 naming, https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/timestamp.md

`timestamp` doesn't follow camelcase, `s` is not capital. Was there specific reason for this?

staltz: Because *Timestamp* is one word https://en.wikipedia.org/wiki/Timestamp

src/add/operator/timestamp.ts
```diff
@@ -0,0 +1,10 @@
+import {Observable} from '../../Observable';
+import {timestamp, TimeStampSignature} from '../../operator/timestamp';
+
+Observable.prototype.timestamp = timestamp;
+
+declare module '../../Observable' {
+  interface Observable<T> {
+    timestamp: TimeStampSignature<T>;
```
staltz: So if you want to make it consistent, this would be named `TimestampSignature`

spec/operators/timestamp-spec.ts
```diff
@@ -0,0 +1,147 @@
+import * as Rx from '../../dist/cjs/Rx.KitchenSink';
+declare const {hot, cold, asDiagram, expectObservable, expectSubscriptions};
+
+declare const rxTestScheduler: Rx.TestScheduler;
+const Observable = Rx.Observable;
+
+/** @test {timestamp} */
+describe('Observable.prototype.timestamp', () => {
+  asDiagram('timestamp')('should record the time stamp per each source elements', () => {
+    const e1 = hot('--a--^b-c-----d--e--|');
```
staltz: I think the PNG marble diagram generation won't like the `--a--^b`, so could you just start this hot observable from the beginning? I'm also not sure if PNG diagrams need `e1subs`, so maybe that can be removed as well.

src/Rx.KitchenSink.ts
```diff
@@ -20,7 +20,9 @@ import './add/operator/mergeScan';
 import './add/operator/min';
 import './add/operator/pairwise';
 import './add/operator/timeInterval';
+import './add/operator/timestamp';
 
 export {TimeInterval} from './operator/timeInterval';
+export {TimeStamp} from './operator/timestamp';
```
staltz: `TimeStamp` => `Timestamp`

src/operator/timestamp.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+}
+
+class TimeStampSubscriber<T> extends Subscriber<T> {
+  constructor(destination: Subscriber<TimeStamp<T>>, private scheduler: Scheduler) {
+    super(destination);
+  }
+
+  protected _next(value: T): void {
+    const now = this.scheduler.now();
```
staltz: I'm wondering what happens if the given scheduler was `null`? RxJS4 uses a default scheduler if the given one is not good https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/linq/observable/timestamp.js#L50

spec/operators/timestamp-spec.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  it('should raise error if source immediately throws', () => {
+    const e1 =  cold('#');
+    const e1subs =   '(^!)';
+    const expected = '#';
+
+    expectObservable(e1.timestamp(rxTestScheduler)).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+});
```
staltz: Good job

src/add/operator/timestamp.ts
kwonoj: Yeah, thanks. I should learn English more. :)

src/operator/timestamp.ts
kwonoj: You mean when someone explicitly set `null` for scheduler?

kwonoj: just note, current scheduler assign behavior's same as `timeInterval` does. if null check to assign default scheduler's preferred, I'll update both.

staltz: Yeah I don't know if these checks will affect perf. In general we don't have validation checks.

kwonoj: For now, I'd suggest to go as is. First, it has default scheduler set behavior when not provided. Secondly, it's case someone puts null intentionally - maybe good to have later but it doesn't feel to should be corrected internally. What do you think?

staltz: Let's do as you suggested :+1:

kwonoj:
Merged with 80b1646.

Issue #1466, opened 2016-03-14
WebSocketSubject cannot be retried.


blesh:
**RxJS version:** 5.0.0-beta.2

**Code to reproduce:**

```js
let socket = new WebSocketSubject('ws://badurlhere');

socket.retryWhen((errs) => errs.switchMap(() => Observable.timer(1000))
  .subscribe(::console.log);
```

**Expected behavior:**

connection fails and then retries once a second.

**Actual behavior:**

connection fails and then `UnsubscriptionError` is thrown, subsequent subscriptions do not work

**Additional information:**

This apparently has to do with some changes to Subject and unsubscription error handling. I'm not sure though. We should definitely add some tests around this scenario, because it's important. WebSocketSubject was supposed to be reusable.

I found this will trying out the `multiplex` method in production for the first time. :\ Bummer.

skaapgif:
I'm going to have a stab at this, I think I've identified the cause, but I'll write a failing test case to isolate it.

skaapgif:
I created a failing test case and a fix:
https://github.com/skaapgif/rxjs/commit/9fd31cc0a0fdb91fc5152c376995918a57689057

When I apply the fix to my own test environment it solves the problem, but the test still fails for some reason. Any ideas?

The problem is each time a multiplex subject unsubscribes, it sends the `unsub` message to announce this to the server. If, however, the underlying socket closes, the multiplex subject still tries to send it's `unsub` message over an already unsubscribed subject which throws an `UnsubscriptionError`.

The current fix check's the underlying socket's ready state with `self.socket.readyState < 3` before sending an `unsub` message, but it might also be possible to check `self.isUnsubscribed`.

fs123:
It looks, that the bug does no longer exists since beta 8. See added test in PR 1737 above.

fs123:
It looks, that the bug does no longer exists since beta 8. See added test in PR 1738 above.

blesh:
> I created a failing test case and a fix: skaapgif@9fd31cc

@skaapgif did you already PR your fix and test for the unsub message?

blesh:
Closed by #1738

skaapgif:
I was still awaiting feedback, but @fs123's test is very similar and more thorough. I was able to reproduce the error without calling `retry(1)` though.

PR #1521, opened 2016-03-23
refactor(Subscription): split 'Subscription' class into class and interface

saneyuki:
- __This will cause BREAKING CHANGE if we accept this proposal.__
- @blesh @trxcllnt how about do you think?

-------

## Motivation

At now, [`Subscription`](https://github.com/ReactiveX/rxjs/blob/master/src/Subscription.ts) used as a implementation class or as an interface in [`Action`](https://github.com/ReactiveX/rxjs/blob/master/src/scheduler/Action.ts), [`Subject`](https://github.com/ReactiveX/rxjs/blob/master/src/Subject.ts), and others.

However, we cannot use a class which has a private member as an interface. This prevents a some refactoring internally (e.g. #1389) and It confuses us (and users) to use a class as an interface.
 

## Detailed Change

- Split `Subscription` class into `CompositeSubscription` class and `Subscription` interface.

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -189,7 +189,7 @@ describe('Observable', () => {
       const sub = source.subscribe(() => {
         //noop
        });
-      expect(sub instanceof Rx.Subscription).toBe(true);
+      expect(sub instanceof Rx.CompositeSubscription).toBe(true);
       expect(unsubscribeCalled).toBe(false);
       expect(typeof sub.unsubscribe).toBe('function');
 
--- a/spec/Subscription-spec.ts
+++ b/spec/Subscription-spec.ts
@@ -2,7 +2,7 @@ import * as Rx from '../dist/cjs/Rx';
 import {DoneSignature} from './helpers/test-helper';
 
 const Observable = Rx.Observable;
-const Subscription = Rx.Subscription;
+const Subscription = Rx.CompositeSubscription;
 
 /** @test {Subscription} */
 describe('Subscription', () => {
--- a/spec/observables/using-spec.ts
+++ b/spec/observables/using-spec.ts
@@ -2,7 +2,7 @@ import * as Rx from '../../dist/cjs/Rx.KitchenSink';
 import {it} from '../helpers/test-helper';
 
 const Observable = Rx.Observable;
-const Subscription = Rx.Subscription;
+const Subscription = Rx.CompositeSubscription;
 
 describe('Observable.using', () => {
   it('should dispose of the resource when the subscription is disposed', (done) => {
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -1,7 +1,7 @@
 import {PartialObserver, Observer} from './Observer';
 import {Operator} from './Operator';
 import {Subscriber} from './Subscriber';
-import {Subscription} from './Subscription';
+import {Subscription, SubscriptionList} from './Subscription';
 import {root} from './util/root';
 import {$$observable} from './symbol/observable';
 import {toSubscriber} from './util/toSubscriber';
@@ -84,7 +84,7 @@ export class Observable<T> implements Subscribable<T> {
    */
   subscribe(observerOrNext?: PartialObserver<T> | ((value: T) => void),
             error?: (error: any) => void,
-            complete?: () => void): Subscription {
+            complete?: () => void): SubscriptionList {
 
     const { operator } = this;
     const subscriber = toSubscriber(observerOrNext, error, complete);
--- a/src/Rx.ts
+++ b/src/Rx.ts
@@ -119,7 +119,7 @@ import './add/operator/zipAll';
 /* tslint:disable:no-unused-variable */
 export {Operator} from './Operator';
 export {Observer} from './Observer';
-export {Subscription, UnsubscriptionError} from './Subscription';
+export {CompositeSubscription, Subscription, UnsubscriptionError} from './Subscription';
 export {Subscriber} from './Subscriber';
 export {AsyncSubject} from './subject/AsyncSubject';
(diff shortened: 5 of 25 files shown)


staltz:
I don't see what do we gain with this...

trxcllnt:
This seems like a good reason to pressure TS to ignore private fields when using a class as an interface...

kwonoj:
@trxcllnt does compiler supports it already? (I assume not yet)

blesh:
So to be clear, this just renamed `Subscription` to `CompositeSubscription` and then created a `Subscription` interface?

blesh:
The truth is that we failed in OOP 101 throughout our code base here. For the more part, all functions should accept interfaces, not classes.

I could see use for an `ISubscription` and an `ISubscriptionList` or something that both match the `Subscription` implementation

In fact, our subscriptions should all really be returning `ISubscription`, and methods that want the `add` and `remove` functionality should really accept `ISubscriptionList`. This gives us flexibility to change our architecture later.

```js
interaface ISubscription {
  unsubscribe(): void;
}

interface ISubscriptionList extends ISubscription {
  add(sub: ISubscription);
  remove(sub: ISubscription);
}
```

saneyuki:
@blesh 

> So to be clear, this just renamed Subscription to CompositeSubscription and then created a Subscription interface?
> https://github.com/ReactiveX/rxjs/pull/1521#issuecomment-200554236

Yes. I intended so.

> I could see use for an ISubscription and an ISubscriptionList or something that both match the Subscription implementation

Okay. I'll make the new patch for this. BTW, Do we add `I`-prefix to them? TypeScript's naming convention does not recommend `I`-prefix for an interface.
https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#names

kwonoj:
>  BTW, Do we add I-prefix to them?

: and also most of our codebases doesn't uses I-prefix for its interfaces, I'd +1 to avoid if possible to consistent. Unfortunately I can't suggest good names.. :(

saneyuki:
How about `Subscriptional`? (I'm not sure my english naming sense...)

staltz:
Unsubscribable or Disposable (interface)
Subscription (class implementing the interface above)

Just a suggestion...

saneyuki:
> Unsubscribable or Disposable (interface)

I feel: 
- `Disposable` need `dispose()` & `isDisposed` as tier 1 (primary) member.
- `Unsubscribable` might be good, but I don't have any concrete confidence that we can choose it as a pair of `Subscribable<T>`...

I'd like to wait @trxcllnt and @blesh 's opinions.
Thank you for your advice, @staltz.

blesh:
@saneyuki I dug into this a bit yesterday.

as I see it, there are really three things:

```TypeScript
interface AnonymousSubscription {
  unsubscribe(): void;
}

interface ISubscription extends AnonymousSubscription {
  unsubscribe(): void;
  add(sub: AnonymousSubscription | Function | void);
  remove(sub: RxSubscription);
  isUnsubscribed: boolean;
}

class Subscription implements ISubscription {
  // stuff
}
```

`AnonymousSubscription` would match the interface defined by the [es-observable spec](https://github.com/zenparsing/es-observable#observable). This is a type we would *accept* from methods that want a Subscription... *however* we should be returning `ISubscription` everywhere.

In fact, while digging into this, I found a lot of security issues that concerned me, and you may even want to hold off on these changes, as I think there might be a larger refactor around subscription and subscribers coming down the pipe.

blesh:
Also: I don't really want to support legacy `.dispose()` interfaces. I think it will encourage some bad practices.

staltz:
Security issues? :hushed:

blesh:
@staltz https://github.com/ReactiveX/rxjs/issues/1532

Since the Observer and Subscription have merged into one "Subscriber", and we're returning the Subscriber directly from `subscribe`, it's the subscriber. You can push values into it and/or otherwise use it like an Observer or something.

It's worth discussing and discussing soon.

saneyuki:
@blesh Should I close this pull request and hand over to you as #1533? Or Should I modify changesets in this just as what you suggested in #1533?

blesh:
@saneyuki the changes in #1533 are approximately what I'd like to land in the repo. I can tell you've put a lot of thought and work into this, and I want to give credit where credit is due. But I do need to get this change in relatively quickly so I can continue with what I'm working on. This is preparation for what I figure might be a larger refactor next week to resolve #1532 and #1466.

You can copy over my commits and keep them in this PR if you like. (I know contributors' commit counts are important to people). Or we can close this and I can just merge the other PR. I'll leave it up to you. You're a solid contributor, and I'd like you to be happy with the outcome. Just let me know fast ;)

saneyuki:
@blesh 

Okay. Thank you for your explain.

Could I take this pull request again?  I'd like to try to shape the patch. If I will have not finish until March 29. AM9:00 (UTC+8),  then please close this pull request.

blesh:
@saneyuki .... I'm sorry, but I'm in a bit of a hurry to get those changes in today so I can start working on refactoring Subjects and @trxcllnt can get to work fixing ConnectableObservable. If you literally want to copy the commits over to your PR so you can have your name on them, I'm totally cool with that, but it needs to be done in the next hour so I can get to work on the changes I need to complete.

blesh:
@saneyuki sorry, but given the time constraints, I had to merge my other PR in favor of this one so I could move on. I greatly appreciate the effort you put in, and I really hope you're not too disappointed. I'll try to find a way to make it up to you. ;)

saneyuki:
@blesh 

No problem! and Thank you for waiting me .

BTW, I found some questions about #1533, thus I may open a new issue/pull request.

stasberkov:
Please remove `I` prefix from `ISubscription`. This makes naming style inconsistent.

PR #1522, opened 2016-03-23
refactor(Action): add 'T' type param to Action.

saneyuki:


src/Scheduler.ts
```diff
@@ -6,6 +6,6 @@ export interface Scheduler {
   schedule<T>(work: (state?: T) => Subscription | void, delay?: number, state?: T): Subscription;
   flush(): void;
   active: boolean;
-  actions: Action[];
+  actions: Action<any>[]; // XXX: use `any` to remove type param `T` from `Scheduler`.
```
saneyuki: I chose this approach to keep compatibility without a breaking change.

Should we add `T` to `Scheduler` for more strongly typing?

blesh:
I'm unsure about this change. What is the goal of the PR? What issue is it fixing?

david-driscoll:
Looks like to add type checking for the scheduler internals, so that when `state` is destructured as part of the scheduler it gets those string types too.

blesh:
Merged with e66b2d8.. Thanks @trxcllnt!

blesh:
lol oops wrong window.

saneyuki:
@blesh 

> I'm unsure about this change. What is the goal of the PR? What issue is it fixing?

The goal of this PR is just what @david-driscoll says. I intend to make internal types of `Action` strongly. In the scope of this PR, I don't think to introduce a breaking change, to keep just as a polishing up.

src/Scheduler.ts
saneyuki: > Should we add T to Scheduler for more strongly typing?

This is a separated issue (Sorry!). [In this PR, I don't discuss about it](https://github.com/ReactiveX/rxjs/pull/1522#issuecomment-202701079)

blesh:
I think this change is fine. The usefulness is dubious I suppose, but otherwise LGTM. :shipit:

kwonoj:
Merged with 5b8cb3e , thanks @saneyuki

PR #1524, opened 2016-03-23
docs(operators): fix typo in delay JSDoc

staltz:
I have no idea how that "order" sneaked in but it really shouldn't be there.

Baby PR. :baby: :baby_bottle:


kwonoj:
Merged with 0f17d5f, thanks @staltz .

PR #1527, opened 2016-03-23
chore(typings): fixed signature for mergeAll to support concurrent as…

### LINKED ISSUE #1526: 5.0.0-beta.3 `mergeAll` does not accept concurrency parameter (typescript)
After upgrading from `5.0.0-beta.2` to `5.0.0-beta.3` following code won't compile:

```typescript
import * as Rx from 'rxjs'

Rx.Observable.range(1, 10)
    .map(x => Rx.Observable.defer(() => Rx.Observable.of(x).delay(1000)))
    .mergeAll(1)  // if I don't pass any value it will compile
    .subscribe(e => console.log(e))
```

Not sure if it is a `typescript` issue or incorrect typings.

the tsc error message is:
`src/test.ts(3,1): error TS2346: Supplied parameters do not match any signature of call target.`

david-driscoll:
**Description:**
The concurrent parameter is currently missing from the signature.

fixes #1526

### DIFF
--- a/spec/operators/mergeAll-spec.ts
+++ b/spec/operators/mergeAll-spec.ts
@@ -61,7 +61,7 @@ describe('Observable.prototype.mergeAll', () => {
     const e1subs =    '^                         !';
     const expected =  '--a---b---c---d---e---f---|';
 
-    expectObservable((<any>e1).mergeAll(1)).toBe(expected);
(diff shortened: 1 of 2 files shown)


kwonoj:
Change looks good to me.

kwonoj:
Merged with 3319a80, thanks @david-driscoll

PR #1530, opened 2016-03-24
chore(package.json): build_test cleans up existing test cases

kwonoj:
**Description:**
This PR makes `build_test` explicitly clean up existing test case build, to avoid possible confusions such as switching branches, etcs. Individual command chain (npm run build && ...) will allows to work with preserved test build.


kwonoj:
Merged with 9700a51.

PR #1535, opened 2016-03-25
refactor(forkJoinObservable): use subscribeToResult for internal beha…

kwonoj:
**Description:**

Compare to other implementation (https://github.com/ReactiveX/rxjs/pull/1510) `forkJoin` already supported promise and others but using its own way. This PR refactors to use `subscribeToResult` instead.

relates to #1483


blesh:
LGTM, merged with ac6923f.

I've added a commit that improves perf a little bit back up past where it was originally (this PR lowered it a little).

Issue #1543, opened 2016-03-29
Operator `take` does not unsubscribe from its source when it reaches the limit with sync source


Igorbek:
**RxJS version:**
master branch.

**Code to reproduce:**

```ts
Observable.create(observer => {
      expect(observer.isUnsubscribed).toBe(false);
      observer.next(42);
      expect(observer.isUnsubscribed).toBe(true); // fails here
    }).take(1).subscribe();
```

**Expected behavior:**

Test should have passed.

**Actual behavior:**

Test has failed.

kwonoj:
Closed via https://github.com/ReactiveX/rxjs/pull/1544.

PR #1544, opened 2016-03-29
Make 'take' unsubscribe when it reaches the total

Igorbek:
**Description:**

- add a test that highlights the issue being fixed
- add a fix

#1543

### DIFF
--- a/spec/operators/take-spec.ts
+++ b/spec/operators/take-spec.ts
@@ -124,4 +124,14 @@ describe('Observable.prototype.take', () => {
     expectObservable(result, unsub).toBe(expected);
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
+
+  it('should unsubscribe from the source when it reaches the limit', () => {
+    const source = Observable.create(observer => {
+      expect(observer.isUnsubscribed).toBe(false);
+      observer.next(42);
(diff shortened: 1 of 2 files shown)


staltz:
Funny, I thought this (existing) test was supposed to cover this case:

```js
  it('should take one values of an observable with many values', () => {
     const e1 = hot('--a--^--b----c---d--|');
     const e1subs =      '^  !            ';
     const expected =    '---(b|)         ';
 
     expectObservable(e1.take(1)).toBe(expected);
   });
```

kwonoj:
Maybe cause taksSubscriber completes destination but didn't do it for self? (which this PR added)

Igorbek:
I suspect that case was handled by scheduler logic. Since it's a hot observable, subscriptions linked with scheduled actions.
BTW, do you have any way to test cold synchronous subscriptions/unsubscriptions with marble diagrams? '(abc|)' doesn't really work if I need to test the order of occurrences.

blesh:
> '(abc|)' doesn't really work if I need to test the order of occurrences.

It should... `(abc|)` does NOT match `(bac|)` with our tests.

blesh:
This LGTM.

kwonoj:
Merged with 6cd2f25, added a commit amends assertion in test cases. Thanks @Igorbek ! :tada:

Issue #1460, opened 2016-03-12
Migrate test runner from jasmine to another


kwonoj:
**Additional information:**

It is known jasmine behaves unexpectedly time to time, test cases are fail without exact reason. As similar to recent failures (https://github.com/ReactiveX/RxJS/pull/1459) just removing test cases into separate file, or modify test execution order makes it passes again. Problem in here is it becomes hard to trust test failures sometimes, cause it's unclear if given assertion failure is due to actual code behavior or simply jasmine's behavior. Maybe it's time to consider to move into different test runner other than jasmine, depends on migration effort.

saneyuki:
How about [power-assert](https://github.com/power-assert-js/power-assert) or [ava](https://github.com/sindresorhus/ava)? Their failure dump are informative.

kwonoj:
I'd like to discuss go / no-go first in this discussion. I also have some preferences, but need to be evaluated based on certain requirements such as current toolchain support (coverage, esdocs, etcs) and it can be evaluated once got agreements to go. Though I think it's time to consider, there might be other difficulties I could not think of to block to move on.

david-driscoll:
I've had good success with Mocha over the year, it has built in support for promises, so for tests I've done in the past, so async tests can sometimes be as easy as...

```
it('should return 0', () => {
  return Observable.of(1).toPromise().then(result => expect(result).to.be.eql(1));
});
```

blesh:
Given what I saw while working on #1536 ... I find jasmine's fragility to be _very_ disturbing.

blesh:
I'm changing this to "high priority" and "PRs welcome" ... but this is NOT at all critical to release, so it's probably not something I'd want anyone tackling in the short term if they're looking to add the most value.

kwonoj:
I'll try this on my end if this can be done relatively quickly.

kwonoj:
Closing this via https://github.com/ReactiveX/rxjs/pull/1545. One remaining issue is being tracked separately via https://github.com/ReactiveX/rxjs/issues/1549.

PR #1545, opened 2016-03-29
Migrate test cases into mocha test runner

kwonoj:
**Description:** :coffee: 

This PR's effort to track #1460, migrate test runner from `jasmine` to `mocha` (https://mochajs.org/) with `chai` for assertion suite, `sinon` for mocking. 

PR contains amount of file changes, major changes are below.

- use `mocha` as test runner 
- migrate global assertion patching in `test-helper` into mocha's custom interface by mocha's designed behavior of having separate global context per test suite
- update test coverage, marble diagram test accordingly 
- now test coverage execution on CI is moved into optional phase completely, will make bit faster build turnaround time
- fixes couple of tests similar to https://github.com/ReactiveX/rxjs/pull/1536

#1460

### NOTE
: This PR doesn't migrate browser test. Current test environment using mocha with custom interfaces makes bit hard to easier migration of karma test configuration, haven't found simple solution yet.

/cc @staltz for visibility & comment.


spec/Notification-spec.ts
```diff
@@ -1,45 +1,46 @@
...
+declare const expectObservable;
 const Notification = Rx.Notification;
 
 /** @test {Notification} */
 describe('Notification', () => {
   it('should exist', () => {
-    expect(Notification).toBeDefined();
-    expect(typeof Notification).toBe('function');
+    expect(Notification).exist;
+    expect(Notification).to.be.a('function');
```
staltz: Wow, I'm sorry you had to go through each of these assertions. Could have used something like https://github.com/crysalead-js/chai-jasmine. Did you try that?

kwonoj: was aware but didn't tempted quite especially given repo is somewhat inactive state. Updating assertion was actually wasn't that hard, just simple replace.

spec/helpers/testScheduler-ui.ts
```diff
@@ -0,0 +1,152 @@
...
+    };
+
+    /**
+     * Number of attempts to retry.
+     */
+    context.it.retries = function(n) {
+      context.retries(n);
+    };
+  });
+};
```
staltz: Good job here

package.json
```diff
@@ -159,19 +160,16 @@
     "gzip-size": "^3.0.0",
     "http-server": "^0.9.0",
     "istanbul": "0.4.2",
-    "jasmine": "2.4.1",
-    "jasmine-ajax": "3.2.0",
-    "jasmine-core": "2.4.1",
     "karma": "^0.13.22",
     "karma-browserify": "5.0.2",
     "karma-chrome-launcher": "0.2.2",
-    "karma-jasmine": "0.3.7",
```
saneyuki: We don't have to replace this with [`karma-mocha`](https://github.com/karma-runner/karma-mocha)?

blesh:
Holy crap! Nice work.

package.json
kwonoj: This PR doesn't include browser test configuration update as noted. May need track it separately.

spec/helpers/testScheduler-ui.ts
```diff
@@ -0,0 +1,152 @@
...
+
+//setup sinon-chai
+chai.use(sinonChai);
+
+declare const module: any;
+declare const global: any;
+
+const diagramFunction = global.asDiagram;
+
+//mocha creates own global context per each test suite, simple patching to global won't deliver its context into test cases.
```
blesh: If mocha is creating it's own global context, should we be adding `rxTestScheduler`, `hot`, `cold`, etc to this context? Or are you doing this already and I'm not seeing it?

spec/helpers/testScheduler-ui.ts
```diff
@@ -0,0 +1,152 @@
...
+      });
+    }
+
+    const it = context.it = context.specify = function(title, fn) {
+      global.rxTestScheduler = null;
+      let modified = fn;
+
+      if (fn && fn.length === 0) {
+        modified = function (done: MochaDone) {
+          global.rxTestScheduler = new Rx.TestScheduler(chai.assert.deepEqual);
```
blesh: Ah... see here: Should we be adding `rxTestScheduler` to `context` here and not `global`? I've never liked the fact we were putting it in shared global state. It's caused me to check it a few times to make sure it wasn't being trampled by bad Jasmine behavior or something. I'd hate to have the same uncertainties with mocha.

spec/helpers/testScheduler-ui.ts
kwonoj: rxTestscheduler is set up in https://github.com/ReactiveX/rxjs/pull/1545/files#diff-044a0fafe3ba4e4178acb8de9fb82252R90. Actually that's interesting question, what I found of mocha's global context is it doesn't allow patching existing interfaces (i.e, `it`, etcs..) cause its interface constructs it everytime per test suite, as you can see in this implementation. Other global instance seems not being affected. Maybe my word on comment is somewhat misleading that all global contexts are completely ignored..

blesh:
Looks good overall, just left a note to spur _discussion_ about where `rxTestScheduler` should live (don't take my comments and run with them, I just want to discuss it).

As far as I'm concerned, :shipit: ship it. I couldn't be happier to get rid of Jasmine.

spec/helpers/testScheduler-ui.ts
kwonoj: Yup, I can try that quickly. I assume it'll work - I also wanted to avoid global patching as much.

Maybe it's good to move all other interfaces (`hot`, `cold`, `expectObservable`...) as well into interfaces?

kwonoj:
> As far as I'm concerned, :shipit: ship it. I couldn't be happier to get rid of Jasmine.

: Cool. Let me do this in this way then - check in this change first, then refactor custom interface per suggestion. It's bit hard to play with this PR since it contains too many files. :) Thanks for reviewing!

kwonoj:
Merged with a1687c3, say goodbye to jasmine and have :coffee: .

@blesh , @staltz , @david-driscoll , @saneyuki , @trxcllnt ... this requires `npm install` once again to trigger. test. Just fyi, test execution time hasn't been changed and nearly identical.

PR #1547, opened 2016-03-29
Write JSDoc for map, mapTo, pluck, filter, scan, do

staltz:


src/operator/do.ts
```diff
@@ -4,14 +4,50 @@ import {Observable} from '../Observable';
...
+ * <span class="informal">Intercepts each emission on the source and runs a
+ * function, but returns an output which is identical to the source.</span>
+ *
+ * <img src="./img/do.png" width="100%">
+ *
+ * Returns a mirrored Observable of the source Observable, but modified so that
+ * the provided Observer is called for every item emitted by the source.
+ *
+ * This operator is useful for debugging your Observables for the correct values
+ * or performing other side effects.
```
blesh: I'd swap this around to say it's useful for side effects *first*, and _then_ mention it's useful for debugging.

blesh: Nevermind, it looks like that's covered above. That's what I get for skimming.

src/operator/do.ts
```diff
@@ -4,14 +4,50 @@ import {Observable} from '../Observable';
...
- * This operator is useful for debugging your observables for the correct values or performing other side effects.
- * @param {Observer|function} [nextOrObserver] a normal observer callback or callback for onNext.
- * @param {function} [error] callback for errors in the source.
- * @param {function} [complete] callback for the completion of the source.
- * @reurns {Observable} a mirrored Observable with the specified Observer or callback attached for each item.
+ * Perform a side effect for every emission on the source Observable, but return
+ * an Observable that is identical to the source.
+ *
+ * <span class="informal">Intercepts each emission on the source and runs a
+ * function, but returns an output which is identical to the source.</span>
```
blesh: How about:

> Allows for "mid-stream" observation of an observable chain by registering an observer or handlers that to observe values, errors and completions, generally for the purpose of creating side effects. Passes all events from the source to the returned observable. Any errors that are thrown in aforementioned observer or handlers are safely sent down the error path of the returned observable.

Thoughts?

src/operator/do.ts
```diff
@@ -4,14 +4,50 @@ import {Observable} from '../Observable';
...
+ * execution, it does not trigger an execution to happen like `subscribe` does.
+ *
+ * @example <caption>Map every every click to the clientX position of that click, while also logging the click event</caption>
+ * var clicks = Rx.Observable.fromEvent(document, 'click');
+ * var positions = clicks
+ *   .do(ev => console.log(ev))
+ *   .map(ev => ev.clientX);
+ * positions.subscribe(x => console.log(x));
+ *
+ * @example <caption>Equivalent to the example above, where `do` is replaced by `map`</caption>
```
blesh: What is the intent of showing `do` replaced by `map` here?

src/operator/do.ts
staltz: Sure. I can add that somewhere, but I try to keep the `informal` description short (my rule of thumb so far is max 2 lines), so your suggestion is longer. I can put your sentence outside of this informal block, like under it. But in general the `informal` description is what I **verbally** explain to a work colleague, written down unaltered.

src/operator/do.ts
staltz: ok :)

src/operator/do.ts
staltz: To illustrate what `do` does, and that it's not anyhow magical. It just takes the input and returns it. But in between, may cause a side effect. If you want, I can remove it, no problem.

blesh: I think remove it. I wouldn't want to indirectly encourage people to use `map` for side effects.

src/operator/do.ts
blesh: I think it's important to stress it's basically about creating a side effect and sending any thrown errors down the error path.

blesh: The "errors down the error path" part is the part most people miss about `do`.

staltz:
@blesh Added "*Any errors that are thrown in the aforementioned Observer or handlers are safely sent down the error path of the output Observable.*" and removed the map example.

blesh:
LGTM. Needs a rebase though.

staltz:
Rebased. :white_check_mark:

kwonoj:
Merged with f209294, thanks @staltz :)

PR #1550, opened 2016-03-29
refactor(test-helper): migrate global patching into custom interface

kwonoj:
**Description:**

@blesh suggestion in #1545 was great, I came to notice global patching won't be needed anymore for most of test cases. This PR updates custom interfaces,

- now `testScheduler` instance works on each suite context scope
- custom interface exposes all custom assertion / marble functions such as `hot`, `cold`, `expectObservable`... 

Note: markdown test is not able to executed by removing all global patching.. /cc @Widdershin


Widdershin:
@kwonoj can we set up a context for markdown doctest execution?

Would love your help getting this running with markdown-doctest as I'm quite busy at the moment.

kwonoj:
@Widdershin problem I faced in here is now everything is under mocha interfaces custom context, not sure how this coordinates with markdown test setups. I've tried quick but no luck :/ I'm planning to check this in and track markdown test as separate.

blesh:
> It's possible to publish this custom interface as separate npm module, anyone can use hot, cold, expectObservable... in their mocha test if they want. Not sure it's feasible consideration though.

We need to publish some sort of testing suite at some point. I think a *lot* of our users will be using Jasmine, though, because that's the Angular team's go-to.

kwonoj:
> I think a lot of our users will be using Jasmine, though, because that's the Angular team's go-to. 

: Yes it is.. just rough idea came to mind since now this custom interface becomes quite independent piece.

Widdershin:
If it's hard to reuse with markdown-doctest it's probably going to be quite hard to reuse with something like tape. Should that be a consideration at this point?

kwonoj:
Maybe I could revive `marble-testing` to export necessary functions and custom interface imports it, instead of contain it all together. Let me try that.

kwonoj:
OK, seems it's working. I wanted to get rid of all access to `rxTestScheduler` in global context, but that'll cause some trouble as experienced in markdown test. I'll leave `marble-testing` as is for now, since anyway all test cases are having separate context for scheduler.

Widdershin:
This looks great. Thanks for going to the trouble of ensuring compatibility with markdown doctest. The new design looks like it will be much easier to package up as well.

:)

kwonoj:
Merged with 3dabea1, thanks @Widdershin for feedback.

PR #1552, opened 2016-03-29
test(helper): enable global leak check for test cases

kwonoj:
**Description:**

This PR enables global leak check in mocha test runner, to ensure if any global patching in test case to be cleared. Seems this is somewhat inaccurate such as some teardown in test hook is not detected, those are marked as ignored.

### DIFF
--- a/spec/operators/toPromise-spec.ts
+++ b/spec/operators/toPromise-spec.ts
@@ -34,6 +34,8 @@ describe('Observable.prototype.toPromise', () => {
     Observable.of(42).toPromise().then((x: number) => {
       expect(wasCalled).to.be.true;
       expect(x).to.equal(42);
+
+      delete __root__.Rx;
       done();
     });
   });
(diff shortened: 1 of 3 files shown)


spec/operators/toPromise-spec.ts
```diff
@@ -34,6 +34,8 @@ describe('Observable.prototype.toPromise', () => {
     Observable.of(42).toPromise().then((x: number) => {
       expect(wasCalled).to.be.true;
       expect(x).to.equal(42);
+
+      delete __root__.Rx;
```
kwonoj: It's satisfying mocha's global leak checking, if test cade modifies global object and does not clear mocha catches it. In this test case it does not affect functionality but added for strict teardown.

kwonoj:
It's small ergonomics changes so assume it'll be ok, but leaving it bit while to hear any other suggestion around before check this in.

blesh:
:shipit: LGTM

kwonoj:
Merged with 26ad3db.

Issue #1295, opened 2016-02-03
RxJS 5 bufferWithTimeOrCount? (add maxCount to bufferTime)


avielfedida:
So RxJS 5 currently not containing `bufferWithTimeOrCount`, is it planned to be added in the future?, anyway I was looking for comparable way to "do" `bufferWithTimeOrCount` in RxJS 5, should I create my own operator or there is a nicer combo with the already ported operators?

blesh:
Hmmm... seems like an oversight. I think we can probably add an additional argument to `bufferTime` to cover this without impacting performance greatly.

Until we add that, you could use something like this: 

(I haven't tested it, and it looks a little gross but with some tweaks it may work):

```js
function bufferWithTimeOrCount(source, duration, bufferSize) {
  let reset = false;

  const buffers = source.scan((buffer, x) => {
    if (reset) {
      buffer = [];
      reset = false;
    }
    buffer.push(x);
    return buffer;
  }, []).share();

  const bufferCounted = buffers.filter(buffer => buffer.length < bufferSize);
  const timer = Observable.of(null).merge(bufferCounted).switchMap(() => Observable.timer(duration));
  
  return buffers.sample(timer.merge(bufferCounted)).do(() => reset = true);
}
```

blesh:
Thank you for your patience and for bring this issue up for us, @avielfedida. We'll get it fixed.

blesh:
### Contributors or those looking to implement this:

The idea I have here is that we can add an optional `maxBufferSize` parameter to `bufferTime` (likewise with `windowTime`) that would default to `Number.POSITIVE_INFINITY`. Then rework the the operator to support that functionality.

justinwoo:
Sorry if I have the details wrong, but `bufferTimeOrCount` only allows a single buffer at a time, right? And as soon as a buffer is flushed, the new buffer starts immediately? [(e.g. this JSBin for single/double clicks)](http://jsbin.com/fupexuw/1/edit?js,console,output)

It looks like in the current implementation of `bufferWithTime`, you'd have to purposefully limit the scheduling of opening multiple buffers and all, which sounds hard/like a lot of work.

I have a small implementation for `bufferTimeOrCount` with [this base spec](https://github.com/justinwoo/RxJS/blob/be8a6b89ae9c98542aeca0d0cb417523f289377b/spec/operators/bufferTimeOrCount-spec.js#L7-L22), but I don't know if this necessarily behaves as expected.

PR #1556, opened 2016-03-30
feat(bufferTime): add `maxBufferSize` optional argument

figueredo:
Add `maxBufferSize` to `bufferTime` as a way of implementing `bufferTimeOrCount` as discussed in #1295.

### DIFF
--- a/spec/operators/bufferTime-spec.ts
+++ b/spec/operators/bufferTime-spec.ts
@@ -18,7 +18,7 @@ describe('Observable.prototype.bufferTime', () => {
       z: []
     };
 
-    const result = e1.bufferTime(t, null, rxTestScheduler);
+    const result = e1.bufferTime(t, null, Number.POSITIVE_INFINITY, rxTestScheduler);
 
     expectObservable(result).toBe(expected, values);
     expectSubscriptions(e1.subscriptions).toBe(subs);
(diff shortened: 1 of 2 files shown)


blesh:
I think it would be better for our end users to allow this operator signature to be more polymorphic.

The following should work:

```
bufferTime(bufferTimeSpan);  // arguments.length === 1
bufferTime(bufferTimeSpan, scheduler);  // arguments.length === 2 && isScheduler(arguments[1])
bufferTime(bufferTimeSpan, bufferCreationInterval); // arguments.length === 2 && !isScheduler(arguments[1])
bufferTime(bufferTimeSpan, bufferCreationInterval, scheduler); // arguments.length === 3 && isScheduler(arguments[2])
bufferTime(bufferTimeSpan, bufferCreationInterval, maxBufferSize); // arguments.length === 3 && !isScheduler(arguments[2])
bufferTime(bufferTimeSpan, bufferCreationInterval, maxBufferSize, scheduler); // arguments.length === 4
```

basically, you're going to have to check the length of the arguments, and whether or not the last argument `isScheduler` in order to determine what's been passed.

blesh:
Other than that, it looks like a solid change. Thanks @figueredo

figueredo:
Yeah, makes much sense, I didn't think of implementing it in that way. I'll change that. Thanks for reviewing, @blesh.

figueredo:
Updated the PR making the signature more polymorphic.

blesh:
I was thinking on this change over this last week. I wonder if maybe RxJS needs to add the concept of buffers as a type of thing (ala js-csp). I think I'd like to put this change on hold until that's discussed.

blesh:
@figueredo I'm unblocking this and I think we can go ahead with a merge. Sorry for the delay. Do you think you can rebase for me?

figueredo:
@blesh I'll do it later today. What about the buffers idea (#1609)?

blesh:
Merged with cf45540... thanks @figueredo for your contribution! Good stuff!

Issue #1554, opened 2016-03-30
Add simple imports for observable creation methods


blesh:
### Problem

Currently when you want to use a creation method in an ad-hoc way, you need to import the type of Observable, then use it's static `create` method or new it up directly.

This poses two problems:

1. You have to know the name of the observable: `rxjs/observable/FromObservable` for example
2. It makes using it ugly:

```js
import { FromObservable } from 'rxjs/observable/FromObservable';
const from = FromObservable.create;

from([1,2,3]).subscribe(console.log.bind(console));
```

### Solution

Add *additional* files next to the observable files that export the `create` methods under a more useful name:

eg:

src/observable/from.ts
```TypeScript
import { FromObservable } from './FromObservable';

export const from = FromObservable.create;
```

src/observable/never.ts
```TypeScript
import { NeverObservable } from './NeverObservable';

export const never = NeverObservable.create;
```

etc.

That means that using these in an ad hoc way would be as simple as:

```js
import { from } from 'rxjs/observable/from'; //LOL: from from from

from([1, 2, 3]).subscribe(console.log.bind(console));
```

staltz:
How do you suggest to import `throw`?

blesh:
@staltz right now it's `import { _throw } from XXX` same for `_do`, `_finally` and `_catch`.

PR #1564, opened 2016-03-30
Creation method moves

blesh:
**Description:**
Adds modules to export creation methods in a more consistent/conventional fashion.

Basically anything that was available with `Rx.Observable.nameHere` is now available from `rxjs/observable/nameHere`. At least that's the idea.

#1554

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "watch": "Watch codebase, trigger compile when source code changes"
   },
   "scripts": {
+    "install": "typings install",
     "info": "npm-scripts-info",
     "build_all": "npm-run-all build_es6 build_amd build_cjs build_global generate_packages",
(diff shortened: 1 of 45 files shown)


blesh:
in other words, you can write in Babel:

```
import { timer } from 'rxjs/observable/timer';

timer(1000)::map(x => x + x).subscribe(::console.log);
```

kwonoj:
Cool, this looks good to me. :)

kwonoj:
Just came to thought - maybe need to update README to includes guidance of how-to, I'd expect this'll be de-facto recommendation of import?

kwonoj:
Let me check this in late today to tomorrow, if there isn't additional suggestions around.

blesh:
I've got it, thanks for the review @kwonoj

kwonoj:
:+1:

PR #1557, opened 2016-03-30
refactor(subjects): move Subjects to src/

figueredo:
Closes #1555.

PR #1565, opened 2016-03-31
chore(CHANGELOG): add breaking changes on beta 4 release

luisgabriel:
It was probably skipped by the script because the commit was a `refactor` instead of `feat` or `fix`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@
 * **forkJoin:** improve forkJoin perf slightly by removing unnecessary context tracking ([280b985](https://github.com/ReactiveX/RxJS/commit/280b985))
 
 
+### BREAKING CHANGES
+
+* Observable: `Observable.fromArray` was removed since it's deprecated on RxJS 4. You should use `Observable.from` instead.
+
+
 
 <a name="5.0.0-beta.3"></a>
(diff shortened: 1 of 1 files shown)


kwonoj:
Would it be possible to add one more https://github.com/ReactiveX/rxjs/pull/1557#issuecomment-203556339 ? It's my fault :cry: /cc @blesh also.

luisgabriel:
@kwonoj Sure. I'll update the PR shortly.

luisgabriel:
@kwonoj Actually, #1557 will be in the next release. So we'll have to add it when beta 5 is out.

kwonoj:
Thanks for clarification. I'll leave this PR to ensure aligned with release schedule and do not mess again.

luisgabriel:
I think this PR should be merged as quickly as possible since beta 4 is already released. Users may not be aware of the fact the `fromArray` was removed.

kwonoj:
I think it'll be quick (and @blesh 's probably already watching this.. :eyes: ) I'll ensure this does not pass this weekend.

blesh:
This PR looks fine. We'll just have to remember #1557 for Monday.

blesh:
Merged with 89c4a9b. Thanks @luisgabriel

PR #1566, opened 2016-03-31
chore(test): allow to throw non error object

kwonoj:
**Description:**

different to jasmine, mocha's async callback requires `Error` object explicitly. This PR updates custom interface bit to allow to throw any object (i.e `throw 'bad';`) by wrap into error object.

p.s : in case of explicit call to async callback, it's still requires to deliver error object.


kwonoj:
Merged with 9c29409.

PR #1567, opened 2016-03-31
Write JSDoc for many static creation operators

staltz:
empty, never, throw, of, create, defer, range, interval, timer.


spec/observables/empty-spec.ts
```diff
@@ -0,0 +1,14 @@
+import {expect} from 'chai';
```
blesh: Looks like this is unused and breaking the build.

spec/observables/never-spec.ts
```diff
@@ -0,0 +1,14 @@
+import {expect} from 'chai';
```
blesh: again, unused and breaking the build.

staltz:
Yep, we saw the same thing. I patched it.

blesh:
LGTM.

blesh:
Merged with 4b30b15. Thanks @staltz

Issue #1568, opened 2016-03-31
Type definition claims missing property?


kwonoj:
**RxJS version:**
master

**Code to reproduce:**
```js
///<reference path="./typings/main/ambient/es6-shim/index.d.ts" />
///<reference path="./spec/es5.d.ts" />

import * as Rx from '.dist/cjs/Rx';

Rx.Observable.of(1).subscribe(_ => {
  console.log(_);
});
```

**Actual behavior:**
`../RxJS/dist/cjs/Observable.d.ts(11,13): error TS2339: Property 'observable' does not exist on type 'SymbolConstructor'.`

**Additional information:**
tsconfig.json
```
{
  "compilerOptions": {
    "removeComments": false,
    "preserveConstEnums": true,
    "sourceMap": true,
    "declaration": true,
    "target": "es5",
    "module": "commonjs"
  },

  "files":["index.ts"]
}
```

Maybe I'm missing something essential after having es5 declarations, @david-driscoll - would you mind to point out what I did missing?

david-driscoll:
That is really weird, I'm seeing it too.... it's not picking up https://github.com/ReactiveX/rxjs/blob/master/src/symbol/symbol.d.ts for some reason.... must be the compiled output... I'll dig in tonight.

david-driscoll:
interesting in `dist/cjs/symbole` there is `/// <reference path="../src/symbol/symbol.d.ts" />
which is wrong...  I guess probably the only way to fix it up is part of the post build steps (where package.json is dropped in, etc).

Does that sound like a decent solution?

kwonoj:
isn't it because of ambient referencing in original code (observable.ts)?

david-driscoll:
It gets stranger... it's gotta be a bug in the TypeScript code, but I have a one line fix that appears to the solve the problem.

kwonoj:
Closed via https://github.com/ReactiveX/rxjs/pull/1569.

kwonoj:
@david-driscoll , reopened this issue by update with latest beta https://github.com/ReactiveX/rxjs/pull/1569#issuecomment-205857392 . Seems ambient type import is not way to go.

kwonoj:
re-closing this issue by https://github.com/ReactiveX/rxjs/pull/1582. To try out this with npm package, may need next version publish though.

PR #1569, opened 2016-03-31
style(typings): import ambient symbol inside Observable to make it av…

david-driscoll:
**Description:**
Adds an ambient reference to `symbol.d.ts` in Observable.  Strangely the ambient reference is not picked up from the `observable.ts`.

See #1568

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -1,3 +1,4 @@
+/// <reference path="./symbol/symbol.d.ts" />
 import {PartialObserver, Observer} from './Observer';
 import {Operator} from './Operator';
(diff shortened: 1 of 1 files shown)


kwonoj:
I think change's ok - can we file issue to typescript team to get some clarification if it's our side or compiler behavior to avoid possible regression in further?

kwonoj:
Merged with 7efcdc9, thanks @david-driscoll

rosendi:
Now I see the following error: `...Observable.d.ts(1,1): error TS2654: Exported external package typings file cannot contain tripleslash references. Please contact the package author to update the package definition.`

Issue #1360, opened 2016-02-18
Possible bug:  can no longer merge and takeUntil on the same stream


bgerm:
In RxJS 4 the code below would print 'up' and return to 0, 0 upon releasing the mouse from a drag.  In RxJS 5 this is no longer happens.

``` javascript
var box = document.getElementById('box');

var mouseDown = Rx.Observable.fromEvent(document, 'mousedown');
var mouseUp = Rx.Observable.fromEvent(document, 'mouseup');
var mouseMove = Rx.Observable.fromEvent(document, 'mousemove');

var mouseDrag = mouseDown.flatMap(function(md){
  var startX = md.offsetX;
  var startY = md.offsetY;

  return mouseMove
    .merge(mouseUp.map(x => 'up')) // <----
    .map(function(mm){
      if (mm === 'up') {
        console.log('up'); // to show it's no longer hit
        return {
          left: 0,
          top: 0
        };
      }
      return {
        left: mm.clientX - startX,
        top: mm.clientY - startY
      };
    }).takeUntil(mouseUp); // <----
});


mouseDrag.subscribe(function(e){
  box.style.top = e.top + 'px';
  box.style.left = e.left + 'px';
});
```

RxJS 4 jsbin:  https://jsbin.com/cigirovohe/1/edit?html,js,console,output
RxJS 5 jsbin:  https://jsbin.com/metosonayi/1/edit?html,js,console,output

trxcllnt:
@bgerm this is technically working as designed, though I can understand your confusion. In RxJS 4 `takeUntil` [subscribes to the source Observable](https://github.com/Reactive-Extensions/RxJS/blob/master/src/modular/observable/takeuntil.js#L32) before subscribing to the notifier Observable, but `takeUntil` in RxJS 5 [subscribes to the notifier](https://github.com/ReactiveX/RxJS/blob/master/src/operator/takeUntil.ts#L27) before subscribing to the source.

In your example, `takeUntil` hears about the `mouseup` event before `merge`, so it disposes the inner Observable before `merge` can next a value through.

Whether this behavior is desired can be debated. @blesh @mattpodwysocki @staltz do you guys feel strongly either way?

mattpodwysocki:
@trxcllnt then I would certainly say that's a bug here and not expected behavior for what was perfectly working code before.

staltz:
[...]

PR #1570, opened 2016-03-31
refactor(operator): operators should call source.subscribe

trxcllnt:
**Description:**
Passing the source into the operator lets the operator determine when the source subscription occurs (if at all). This will allow us to fix subscription ordering bugs like #1360.

#1360 is the only one that's filed as far as I know, but there are most likely others (`finally` comes to mind).

cc: @blesh


src/Operator.ts
```diff
@@ -1,7 +1,7 @@
 import {Subscriber} from './Subscriber';
 
 export class Operator<T, R> {
-  call(subscriber: Subscriber<R>): Subscriber<T> {
-    return new Subscriber<T>(subscriber);
+  call(subscriber: Subscriber<R>, source: any): any {
```
kwonoj: Is it possible to limit type of `source` instead of `any`?

blesh:
Looks like what we discussed. LGTM.

blesh:
Merged with fd0823b. Thanks, as always @trxcllnt

staltz:
Weird, how is this PR still open?

kwonoj:
I think it's just not closed. Let me close it.

PR #1573, opened 2016-04-01
docs(operators): fix JSDoc for the static combineLatest()

staltz:
**Description:**
For some reason I missed documenting the static combineLatest, which is roughly the same as the instance combineLatest. I wrote the static merge and instance merge in this style too, but just forgot to do it for combineLatest.


staltz:
Merged via c13408dda99aadd06b2942e2f90771a5eb7893dd

PR #1582, opened 2016-04-06
chore(typings): removed ambient symbol, and use of Iterable<T>

david-driscoll:
**Description:**
:arrow_forward: :horse: :arrow_backward: 

To early for prime time it seems. :cry:

These are currently not compatible with ES5 targets


kwonoj:
Thanks for quick approach. Seems travis's unhappy?

kwonoj:
And assume this is related to https://github.com/ReactiveX/rxjs/issues/1568, is my understanding correct?

kwonoj:
Merged with d185ff8, thanks @david-driscoll .

PR #1585, opened 2016-04-06
chore(test): enable source map for test stack trace

kwonoj:
**Description:**
This PR adds small ergonomics into test runner to enable source map of stack trace

before:
before

after:
after

It doesn't give clarity for test cases using testscheduler.flush() much though, gives small convenience over general failures.


kwonoj:
Merged with b43e90f. Please do `npm install`, @blesh , @staltz , @david-driscoll , @saneyuki , @luisgabriel

PR #1589, opened 2016-04-07
fix(AjaxObservable): support json responseType on IE

### LINKED ISSUE #1381: Observable.ajax response is string for IE11
I'm seeing an issue on 5.0.0-beta.2 where the ajax observable returns plain text when `responseType` is either set to `json` or left blank.

Example:
```js
Observable.ajax({
  method: 'GET',
  url: `api/blah`,
  crossDomain: true,
  resultSelector(res) {
    return res.response
  }
});
```

For a payload that is an Array, `res.response` is of type Array in Chrome, however, on IE11 it is a string.  Digging into it, it appears the `responseType` property on the xhr at https://github.com/ReactiveX/RxJS/blob/master/src/observable/dom/AjaxObservable.ts#L353 appears to always be empty, regardless of what `request.responseType` was at https://github.com/ReactiveX/RxJS/blob/master/src/observable/dom/AjaxObservable.ts#L215.

This is easy enough to work around in the resultSelector:
```js
function jsonResultSelector(res) {
  if(typeof res.response === 'string') {
    return JSON.parse(res.response);
  }
  return res.response;
}
```

kwonoj:
**Description:**
This PR updates internal behavior of AjaxObservable for IE specific by IE doesn't support json as response type, parse response as json object internally.

closes #1381

### DIFF
--- a/src/observable/dom/AjaxObservable.ts
+++ b/src/observable/dom/AjaxObservable.ts
@@ -375,11 +375,13 @@ export class AjaxResponse {
 
   constructor(public originalEvent: Event, public xhr: XMLHttpRequest, public request: AjaxRequest) {
     this.status = xhr.status;
-    this.responseType = xhr.responseType;
+    this.responseType = xhr.responseType || request.responseType;
+
     switch (this.responseType) {
       case 'json':
         if ('response' in xhr) {
-          this.response = xhr.response;
+          //IE does not support json as responseType, parse it internally
+          this.response = xhr.responseType ? xhr.response : JSON.parse(xhr.response || xhr.responseText || '');
         } else {
           this.response = JSON.parse(xhr.responseText || '');
         }


kwonoj:
I'm not clear at the moment if this approach is legit & acceptable, send out as PR for discussion to close out issue. If it's expected to not to support those and let it be handled externally, I'll simply close this PR as well as issue.

/cc @trxcllnt also for visibility.

src/observable/dom/AjaxObservable.ts
```diff
@@ -375,11 +375,13 @@ export class AjaxResponse {
...
     this.status = xhr.status;
-    this.responseType = xhr.responseType;
+    this.responseType = xhr.responseType || request.responseType;
+
     switch (this.responseType) {
       case 'json':
         if ('response' in xhr) {
-          this.response = xhr.response;
+          //IE does not support json as responseType, parse it internally
+          this.response = xhr.responseType ? xhr.response : JSON.parse(xhr.response || xhr.responseText || '');
```
blesh: Have we run this across all browsers? It _seems_ legit on the surface.

kwonoj: I've ran over Chrome / IE, can check FF (but not able to verify on safari).

kwonoj: Yes, this logic itself works on FF as well. (noticed response is null actually, but this seems somewhat different issue - api itself, or somewhere else that need to triage separately)

blesh:
Merged with bba13d8. Thanks, @kwonoj

danmarshall:
Cool fix @kwonoj - however I am seeing a specific issue when server replies with HTTP 204 - No Content. 

When this code executes:
```
JSON.parse(xhr.response || xhr.responseText || '')
```
If ```xhr.response``` or ```xhr.responseText``` are null then the statement will parse the empty string.
JSON.parse('') throws a syntax error (in any browser). Fix would be to parse ```'null'``` to create a ```null``` which is the response object for an HTTP 204 in non-IE browsers.

kwonoj:
@danmarshall Thanks for suggestion. I'll try to come up with PR shortly.

src/observable/dom/AjaxObservable.ts
danmarshall: the fallback '' should be 'null' instead. I will send a PR

kwonoj: @danmarshall , OK, then will wait PR instead of writing it up by my own.

maximusdominus:
Not sure if this is the appropriate place to put this, but I Just encountered this bug in a different place: when the server returns JSON in a 4xx / 5xx response. 

`typeof(error.xhr.response)` on IE is `string`, but on all other browsers it's an `object`

Here's what my code looks like

<pre>
(action$, store) => action$
    .ofType(MY_TYPE)
    .switchMap( action =>
      Observable.ajax({
        url: 'myUrl',
        method: 'POST',
        body: action.options,
        responseType: 'json',
      })
      .map(...)
      .catch(error => Observable.of({type: FETCH_FAILURE, errorResponse: error.xhr.response})),
    )
</pre>

adepatie:
@maximusdominus Were you able to resolve this issue? I'm facing the same issue.

PR #1599, opened 2016-04-09
typo on documentation.

vic:
**Description:**


kwonoj:
Merged with d47e1fb, thanks @vic . To follow commit message guideline (https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md#commit) which is being used for automatic changelogs, I've amended commit message bit. Hope it's ok for you, and thanks again for effort to create PR!

PR #1600, opened 2016-04-09
fix(typings): make an exported function's destructuring signature explicitly

saneyuki:
* This decrease cases which uses `any` type implicitly.

### DIFF
--- a/src/observable/BoundCallbackObservable.ts
+++ b/src/observable/BoundCallbackObservable.ts
@@ -138,11 +138,21 @@ function dispatch<T>(state: { source: BoundCallbackObservable<T>, subscriber: Su
   self.add(subject.subscribe(subscriber));
 }
 
-function dispatchNext({ value, subject }) {
+interface DispatchNextArg<T> {
+  subject: AsyncSubject<T>;
+  value: T;
+}
+function dispatchNext<T>(arg: DispatchNextArg<T>) {
+  const { value, subject } = arg;
   subject.next(value);
   subject.complete();
 }
 
-function dispatchError({ err, subject }) {
+interface DispatchErrorArg<T> {
+  subject: AsyncSubject<T>;
+  err: any;
+}
+function dispatchError<T>(arg: DispatchErrorArg<T>) {
+  const { err, subject } = arg;
   subject.error(err);
 }
--- a/src/observable/BoundNodeCallbackObservable.ts
+++ b/src/observable/BoundNodeCallbackObservable.ts
@@ -135,11 +135,21 @@ function dispatch<T>(state: { source: BoundNodeCallbackObservable<T>, subscriber
   self.add(subject.subscribe(subscriber));
 }
 
-function dispatchNext({ value, subject }) {
+interface DispatchNextArg<T> {
+  subject: AsyncSubject<T>;
+  value: T;
+}
+function dispatchNext<T>(arg: DispatchNextArg<T>) {
+  const { value, subject } = arg;
   subject.next(value);
   subject.complete();
 }
 
-function dispatchError({ err, subject }) {
+interface DispatchErrorArg<T> {
+  subject: AsyncSubject<T>;
+  err: any;
+}
+function dispatchError<T>(arg: DispatchErrorArg<T>) {
+  const { err, subject } = arg;
   subject.error(err);
 }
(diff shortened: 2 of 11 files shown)


saneyuki:
> - This also fixes the compile error if you use typescript@1.9.0-dev.20160408 or later in your project which uses rxjs.
>  -   Their versions checks implicitly any type in destructuring more strictly assignment with using noImplicitAny compiler option.

Ugh, Sorry,  TypeScript 1.8 don't emit destructuring assignment signature, but I forgot it before opening this pull request...

saneyuki:
So this just decreases cases which uses any type implicitly.

src/observable/BoundCallbackObservable.ts
```diff
@@ -138,11 +138,19 @@ function dispatch<T>(state: { source: BoundCallbackObservable<T>, subscriber: Su
   self.add(subject.subscribe(subscriber));
 }
 
-function dispatchNext({ value, subject }) {
+type DispatchNextArg<T> = {
```
kwonoj: Usually for signatures used `interface` unexported (https://github.com/ReactiveX/rxjs/blob/dd012796f09e2f76be0d7deb0e7d80310048bcd5/src/operator/windowToggle.ts#L77) - is there difference for this case? I assume interface will works.

saneyuki: Ah, I thought them just as an alias, not intended to introduce a new type which are implemented from other types.

But other parts uses `interface`, I'll sort with them

kwonoj: Yes, I don't think it's big deal, just for consistency. :)

saneyuki:
@kwonoj updated.

src/observable/BoundCallbackObservable.ts
```diff
@@ -138,11 +138,19 @@ function dispatch<T>(state: { source: BoundCallbackObservable<T>, subscriber: Su
   self.add(subject.subscribe(subscriber));
 }
 
-function dispatchNext({ value, subject }) {
+interface DispatchNextArg<T> {
+  subject: AsyncSubject<T>;
+  value: T;
+}
+function dispatchNext<T>({ value, subject }: DispatchNextArg<T>) {
```
kwonoj: it's also consistency suggestion - current codebase destructures inside of code - is there specific reason to do it in parameter?

```
function do(xxx : xxxType) {
const {value, subject} = xxx;
}
```

saneyuki: Umm, it would be just a matter of taste. But I don't have any strong options in my changed area. I'll change to the style to destructure inside of codes.

saneyuki:
@kwonoj updated again.

kwonoj:
Thanks, change looks good to me.

kwonoj:
Seems travis is not happy due to lint failure.

saneyuki:
oops :dizzy_face:

saneyuki:
updated!

src/observable/BoundCallbackObservable.ts
```diff
@@ -138,11 +138,21 @@ function dispatch<T>(state: { source: BoundCallbackObservable<T>, subscriber: Su
   self.add(subject.subscribe(subscriber));
 }
 
-function dispatchNext({ value, subject }) {
+interface DispatchNextArg<T> {
+  subject: AsyncSubject<T>;
+  value: T;
+}
+function dispatchNext<T>(arg: DispatchNextArg<T>) {
+  const { value, subject } = arg;
```
blesh: This might read cleaner (maybe?) as:

```TypeScript
function dispatchNext<T> ( { value, subject }: DispatchNextArg<T> ) {
 ....
}
```

One of the advantages of destructured function args is readability (but that's pretty subjective).

What do you think?

blesh:
@kwonoj @david-driscoll ... do these new interfaces need to be exported?

src/observable/BoundCallbackObservable.ts
kwonoj: Actually that was previous impl from @saneyuki , and I suggested current way. Do you prefer restructuring in param directly?

kwonoj:
I don't think so, most dispatch occurs internally and context argument is different per operator behaviors.

blesh:
This LGTM. I'm going to merge it so it's in this release.

blesh:
Merged with ef305af... thanks @saneyuki

saneyuki:
@blesh @kwonoj thank you for accepting this.

Issue #1604, opened 2016-04-10
rxjs-es d.ts file errors when compiling with tsc --allowJs


subhaze:
**RxJS version:**
rxjs-es@5.0.0-beta.5

**Code to reproduce:**
main.js file: `import * as Rx from './node_modules/rxjs-es/Rx';`

tsc cli: `tsc --allowJs --outFile public/main-dist.js --module system public/main.js`

**Expected behavior:**
for tsc compiler to emit Rx code and bundle up into main-dist.js

**Actual behavior:**
tsc compiler errors with the following
```
public/node_modules/rxjs-es/Observable.d.ts(12,6): error TS2304: Cannot find name 'Symbol'.
public/node_modules/rxjs-es/Observable.d.ts(14,86): error TS2304: Cannot find name 'Promise'.
public/node_modules/rxjs-es/Observable.d.ts(15,42): error TS2304: Cannot find name 'Iterable'.
public/node_modules/rxjs-es/Observable.d.ts(71,60): error TS2304: Cannot find name 'Promise'.
public/node_modules/rxjs-es/Observable.d.ts(71,70): error TS2304: Cannot find name 'Promise'.
public/node_modules/rxjs-es/observable/PromiseObservable.d.ts(22,31): error TS2304: Cannot find name 'Promise'.
public/node_modules/rxjs-es/observable/PromiseObservable.d.ts(23,26): error TS2304: Cannot find name 'Promise'.
public/node_modules/rxjs-es/operator/toPromise.d.ts(7,59): error TS2304: Cannot find name 'Promise'.
public/node_modules/rxjs-es/operator/toPromise.d.ts(7,69): error TS2304: Cannot find name 'Promise'.
public/node_modules/rxjs-es/operator/toPromise.d.ts(9,9): error TS2304: Cannot find name 'Promise'.
public/node_modules/rxjs-es/operator/toPromise.d.ts(10,26): error TS2304: Cannot find name 'Promise'.
public/node_modules/rxjs-es/operator/toPromise.d.ts(10,36): error TS2304: Cannot find name 'Promise'.
make: *** [build] Error 2
```
and no Rx code is emitted/bundled into main-dist.js

**Additional information:**
If I remove all of the d.ts files that are shipped with rxjs-es everything works correctly. Rx code is emitted and bundled up correctly into main-dist.js

kwonoj:
[...]

Issue #1066, opened 2015-12-15
Cannot find name 'Promise' in Observable.d.ts


d3lm:
I am trying RxJS 5.0.0-beta.0 and have installed it via npm via `npm install rxjs@5.0.0-beta.0`. Now I would like to import Subject and Observable from rxjs and did the following `import {Observable, Subject} from 'rxjs';`. This throws the error that 'Promise' cannot be found in Observable.d.ts. Is this an issue with RxJS or did I forget something? This is the full error:

    node_modules\rxjs\Observable.d.ts (154,66): error TS2304: Cannot find name 'Promise'
    node_modules\rxjs\CoreOperators.d.ts (77,31): error TS2304: Cannot find name
      'PromiseConstructor'
    node_modules\rxjs\CoreOperators.d.ts (22,67): error TS2304: Cannot find name 'Promise'

Maybe the declaration for Promise is missing here.

d3lm:
Was using RxJS with Typescript and had to `tsd install es6-shim`. Now its working.

blesh:
Thanks for the issue and the resolution, @d3lm :)

kwonoj:
Actually this has been asked before also, https://github.com/ReactiveX/RxJS/issues/592. Maybe better to update README to include typescript example, to notify type declaration for es6 shim is required? @blesh

app-sparkler:
`tsd install es6-shim --save` worked for me.

Issue #592, opened 2015-10-23
Type definition complains 


ziaukhan:
I am trying to transpile the following helloworld.ts file:

import Rx = require("@reactivex/rxjs");
Rx.Observable.of('Hello World!').subscribe(function(value) {
	console.log(value);
});

I have TypeScript 1.6.2 installed and have installed:
npm install @reactivex/rxjs

I am receiving the following errors, am I doing something wrong or this is a bug:

tsc -module commonjs helloworld
node_modules/@reactivex/rxjs/dist/cjs/CoreOperators.d.ts(21,70): error TS2304: Cannot find name 'Promise'.
node_modules/@reactivex/rxjs/dist/cjs/CoreOperators.d.ts(73,31): error TS2304: Cannot find name 'PromiseConstructor'.
node_modules/@reactivex/rxjs/dist/cjs/CoreOperators.d.ts(73,54): error TS2304: Cannot find name 'Promise'.
node_modules/@reactivex/rxjs/dist/cjs/Observable.d.ts(63,53): error TS2304: Cannot find name 'PromiseConstructor'.
node_modules/@reactivex/rxjs/dist/cjs/Observable.d.ts(63,74): error TS2304: Cannot find name 'Promise'.
node_modules/@reactivex/rxjs/dist/cjs/Observable.d.ts(74,38): error TS2304: Cannot find name 'Promise'.
node_modules/@reactivex/rxjs/dist/cjs/Observable.d.ts(97,69): error TS2304: Cannot find name 'Promise'.
node_modules/@reactivex/rxjs/dist/cjs/Observable.d.ts(149,30): error TS2304: Cannot find name 'PromiseConstructor'.
node_modules/@reactivex/rxjs/dist/cjs/Observable.d.ts(149,53): error TS2304: Cannot find name 'Promise'.

kwonoj:
You're missing type definition declaration for promise.

///<reference path="...../fileforpromise definition.d.ts" />

either you can manually pick those definition, or could lookup [definitelytyped](http://definitelytyped.org/).

kwonoj:
In my case, I use [`es6-promise`](https://github.com/borisyankov/DefinitelyTyped/blob/master/es6-promise/es6-promise.d.ts) for testing codes.

ziaukhan:
After adding the es6-promise it is still giving the following errors:

[...]

PR #1605, opened 2016-04-11
docs(installation): add instruction for es6-shim type definition

kwonoj:
**Description:**
This PR add small instructions to include `es6-shim` when try to compile via TypeScript.

#1604, #1066, #592

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -66,18 +66,16 @@ map.call(Observable.of(1,2,3), function (x) { return x + '!!!'; });
 You can also use the above method to build your own Observable and export it from your own module.
 
 ### CommonJS with TypeScript
-If you recieve an error like `error TS2304: Cannot find name 'Symbol'` or `error TS2304: Cannot find name 'Iterable'` when using RxJS you may need to install a supplemental set of typings.  This small set of interfaces allows RxJS to target ES6, while still allowing the TypeScript compiler to target ES5.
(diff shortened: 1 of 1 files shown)


staltz:
Can you instead put this change in `doc/installation.md`? I also think we should make the README have basically only the badges, a big link to reactivex.io/rxjs and information for contributors. All information for library consumers should be on reactivex.io/rxjs. Could you also remove this `## Installation and usage` section from README since it's the same as `doc/installation.md`?

kwonoj:
I'm OK with that, but isn't `doc` supposed to be posted as document at reactivex.io/rxjs instead of github repo instruction? I'm still feeling this might need to be posted in repo somewhere due to lot of first TypeScript users are lacks this is requirement.

staltz:
You mean we need to update reactivex.io/rxjs? We can do that asap I think.

kwonoj:
Ah, actually opposite. I assume your suggestion is to place under `doc`, which intended to published into reactivex.io.

staltz:
Yes...

kwonoj:
Updated, along with removing instruction added by https://github.com/ReactiveX/rxjs/pull/1582 which is not reflecting current code.

staltz:
LGTM

kwonoj:
Merged with 815f775, thanks for review @staltz

Issue #1602, opened 2016-04-09
Obserable Create methods docs and overview missing


SteveALee:
**RxJS version: Master branch

**Code to reproduce: n/a

**Expected behavior: n/a

**Actual behavior: n/a

**Additional information:**
Looking for the creation methods in the docs I found only a reference in [observable.md](https://github.com/ReactiveX/rxjs/blob/master/doc/observable.md) to **Creation Methods** fragment in overview.md. That appears to be missing.

staltz:
Mmm, I don't think it's missing: file:///Users/amed/Development/RxJSNext/tmp/docs/manual/overview.html#creation-operators

SteveALee:
I was looking here in github  https://github.com/ReactiveX/rxjs/tree/master/doc. No overview.md there.  I guess I'm miss understanding the docs generation process.

staltz:
https://github.com/ReactiveX/rxjs/pull/1607

SteveALee:
OK :)

PR #1607, opened 2016-04-11
docs(instructions): write instructions for not reading doc/ directory

staltz:
**Description:**
People have been mislead to look for docs by browsing through the doc/ directory through GitHub,
like they did with RxJS 4. This directory is not meant for content to be consumed directly through
GitHub, so this README will instruct people to look at reactivex.io/rxjs instead.

#1602

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,0 +1,3 @@
+# RxJS official docs at http://reactivex.io/rxjs/
+
+These files are not meant for reading directly in GitHub, they are just source code for generating the official page. You should find the docs at http://reactivex.io/rxjs/, containing all the documentation.


SteveALee:
People use to dealing will source code will naturally look in the docs folder as you observed. My guess is the README may not be looked at either. How about renaming the docs folder to something else?

staltz:
That's one option, but the README in that folder will make it easily visible to anyone browsing the repo.

staltz:
@kwonoj Is this mergeable?

kwonoj:
@staltz yes, looks good to me.

kwonoj:
Merged with c3c951e, thanks @staltz

PR #1612, opened 2016-04-12
chore(test): update test script for documentation

kwonoj:
**Description:**
This PR updates configurations to work with scripts for documentation purpose.

@blesh , sorry for causing regression. I believe now issue is resolve by checking locally, leave this bit if you'd like to check once again before check this in.


kwonoj:
Merged with 031013a.

PR #1617, opened 2016-04-14
Matching es-observable proposal `of` async emission

blesh:
**Description:**

1. Adds an `Rx.Scheduler.none`. This is a recursive scheduler, but in practice in our code it should never be called. Rather it should be referentially checked, then a recursive "hot path" should be taken.
2. Updates `Notification` to stop using static observable creation methods (it was using `Observable.of` internally)
3. Updates `Observable.of` to default to `asap` scheduling. This is done to match the [es-observable proposal's definition of `of`](http://github.com/zenparsing/es-observable)
4. Adds a perf test comparing the plain calls of `of` for both Rx 4 and 5

**Related issue (if exists):** ??? I'm sure it's come up, I'll have to search for one.

### Please direct hate mail regarding this design to @zenparsing, @jhusain and others at the [es-observable spec](http://github.com/zenparsing/es-observable). If you care that much, you should be more involved in the discussions on that spec.

😛

FWIW: The perf even with micro task scheduling in 5 vs trampoline scheduling in 4 is still solidly improved:

```

                     RxJS 4.1.0 |              RxJS 5.0.0-beta.6 |          factor |      % improved
                55,125 (±1.39%) |               351,630 (±0.80%) |           6.38x |          537.9%
```

### DIFF
--- a/perf/micro/immediate-scheduler/observable/of-plain-calls.js
+++ b/perf/micro/immediate-scheduler/observable/of-plain-calls.js
@@ -1,0 +1,34 @@
+var RxOld = require('rx');
+var RxNew = require('../../../../index');
+
+module.exports = function of(suite) {
+  var args = [];
+  for (var i = 0; i < 25; i++) {
+    args.push(i);
+  }
+
+  var oldOfWithImmediateScheduler = RxOld.Observable.of.apply(null, args);
+  var newOfWithImmediateScheduler = RxNew.Observable.of.apply(null, args);
+
+  function _next(x) { }
+  function _error(e) { }
+
+  // add tests
+  return suite
+    .add('old of called plain', {
+      defer: true,
+      fn: function (deferred) {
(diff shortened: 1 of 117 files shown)


blesh:
The above said, I will say that this change did cut the performance of `Observable.of()`'s bare call down to 1/3 of what it was.

trxcllnt:
> The above said, I will say that this change did cut the performance of Observable.of()'s bare call down to 1/3 of what it was.

I was there in the room when we all agreed (yes, even @jhusain) with @benjchristensen that we should deviate from the spec when it negatively impacts performance. It would seem this is one such scenario.

zenparsing:
😲 

You might be pleased to know that I think I may have finally convinced myself that `of` should just be sync in the spec.  I'll start (or restart) a thread over on es-observable tomorrow to discuss.

trxcllnt:
@zenparsing 😅🙊👌🏽💃❣

PR #1619, opened 2016-04-14
fix(race): handle observables completes immediately

### LINKED ISSUE #1615: Error with `race` and sync Observables
**RxJS version:**

5.0.0-beta.6

**Code to reproduce:**

At the Node (v5.0.0) console:

```
rx.Observable.race(rx.Observable.of(true), rx.Observable.of(false).delay(1500)).subscribe(console.log.bind(console))
```

**Expected behavior:**

```
true
```

**Actual behavior:**

```
true
TypeError: Cannot read property 'push' of null
    at RaceSubscriber._complete (/.../node_modules/@reactivex/rxjs/dist/cjs/operator/race.js:86:35)
    at RaceSubscriber.Subscriber.complete (/.../node_modules/@reactivex/rxjs/dist/cjs/Subscriber.js:113:18)
    at ArrayObservable._subscribe (/.../node_modules/@reactivex/rxjs/dist/cjs/observable/ArrayObservable.js:116:24)
    at ArrayObservable.Observable.subscribe (/.../node_modules/@reactivex/rxjs/dist/cjs/Observable.js:52:62)
    at Observable._subscribe (/.../node_modules/@reactivex/rxjs/dist/cjs/Observable.js:109:28)
    at RaceOperator.call (/.../node_modules/@reactivex/rxjs/dist/cjs/operator/race.js:55:23)
    at Observable.subscribe (/.../node_modules/@reactivex/rxjs/dist/cjs/Observable.js:52:38)
    at repl:1:81
    at REPLServer.defaultEval (repl.js:248:27)
    at bound (domain.js:280:14)
```

**Additional information:**

If I `delay` both, I don't get the TypeError. If I use `toPromise` and then `.then` it, rather than calling `subscribe` I don't get the TypeError.

It seems clear that it's a result of one of the Observables being synchronous. However if this is expected behavior I have two questions: (1) What should I do instead, if I receive Observables that may or may not be sync? (2) Why does the error appear to be an internal error, rather than a user-facing error?

cc @matthewwithanm

kwonoj:
**Description:**

This PR handles cases where observables are completes immedately, cleans up inner subscriptions
synchronously. 

closes #1615

### DIFF
--- a/spec/operators/race-spec.ts
+++ b/spec/operators/race-spec.ts
@@ -1,3 +1,4 @@
+import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
 declare const {hot, cold, expectObservable, expectSubscriptions};
 
@@ -146,5 +147,14 @@ describe('Observable.prototype.race', () => {
     expectObservable(result).toBe(expected);
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
     expectSubscriptions(e2.subscriptions).toBe(e2subs);
+  });
+
+  it('should allow observable emits immediately', (done: MochaDone) => {
+    const e1 = Observable.of(true);
+    const e2 = Observable.timer(200).map(_ => false);
+
+    Observable.race(e1, e2).subscribe(x => {
+      expect(x).to.be.true;
+    }, done, done);
   });
 });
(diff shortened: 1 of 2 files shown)


blesh:
Merged with abac3d1... Thanks @kwonoj

PR #1626, opened 2016-04-15
chore(typings): increase combineLatest's signature compatibility with v4

saneyuki:
- Add the case for ~`T9` type parameters which are defined in v4
  - https://github.com/Reactive-Extensions/RxJS/blob/9d6ea94244c2f40a29373f9ed6b12d951d3f2238/ts/core/linq/observable/combinelatestproto.ts
  - https://github.com/Reactive-Extensions/RxJS/blob/9d6ea94244c2f40a29373f9ed6b12d951d3f2238/ts/core/linq/observable/combinelatest.ts
- Restrict cases which accepts arrays as the 1st argument of `combineLatest()`
  - From `any` to `T`.
  - As my thought, It would be rare case to pass multiple types with stuffing into an array.

PR #1627, opened 2016-04-15
chore(typings): increase zip's signature compatibility with v4

saneyuki:
- Add the case for ~T9 type parameters which are defined in v4
  - https://github.com/Reactive-Extensions/RxJS/blob/9d6ea94244c2f40a29373f9ed6b12d951d3f2238/ts/core/linq/observable/zip.ts
- Restrict cases which accepts arrays as the 1st argument of `zip()`
  - From `any` to `T`.
  - As my thought, It would be rare case to pass multiple types with stuffing into an array.

### DIFF
--- a/src/operator/zip.ts
+++ b/src/operator/zip.ts
@@ -35,9 +35,9 @@ export interface ZipSignature<T> {
   <T2, T3, T4, T5>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>): Observable<[T, T2, T3, T4, T5]>;
   <T2, T3, T4, T5, T6>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>): Observable<[T, T2, T3, T4, T5, T6]>;
 
-  <R>(...observables: Array<ObservableInput<any> | ((...values: Array<any>) => R)>): Observable<R>;
-  <R>(array: ObservableInput<any>[]): Observable<R>;
-  <R>(array: ObservableInput<any>[], project: (...values: Array<any>) => R): Observable<R>;
+  <R>(...observables: Array<ObservableInput<T> | ((...values: Array<T>) => R)>): Observable<R>;
+  <R>(array: Array<ObservableInput<T>>): Observable<R>;
+  <TOther, R>(array: Array<ObservableInput<TOther>>, project: (v1: T, ...values: Array<TOther>) => R): Observable<R>;
 }
 /* tslint:enable:max-line-length */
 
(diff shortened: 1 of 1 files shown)


saneyuki:
related #1626

kwonoj:
> As my thought, It would be rare case to pass multiple types with stuffing into an array.

So this means `zip(Observable<string>, Observable<number>, Observable<other>)`..(just example) won't be supported out of box, is that correct?

saneyuki:
> So this means zip([Observable<string>, Observable<number>, Observable<other>])..(just example) won't be supported out of box, is that correct? 

Yes, it is about the current patch (`<R>(array: Array<ObservableInput<T>>, project: (v1: T, ...values: Array<T>) => R): Observable<R>;`).

After @kwonoj's  pointed out, I re-think typing of them, so I reproposal followings:

- We should accepts `observable1<T1>.zip(array: Array<T2>)` style on instance methods because we use the 1st value (observable1<T1>) as controlling signal of zipped groups.
   - So we treat `observable1`'s value as the "intention" event. This usecase is very similar to the usecase of `withLatestFrom()`.
- In static methods, we should only accepts `Rx.Observable.combineLatest(array: Array<T>)` because we pass a spawned observable list to it in almost case.
  - So there are no "intention" event in using static version.

src/operator/zip.ts
```diff
@@ -48,15 +54,21 @@ export function zipStatic<T, T2, T3>(v1: ObservableInput<T>, v2: ObservableInput
...
 export function zipStatic<T, T2, T3, T4, R>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, project: (v1: T, v2: T2, v3: T3, v4: T4) => R): Observable<R>;
 export function zipStatic<T, T2, T3, T4, T5, R>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, project: (v1: T, v2: T2, v3: T3, v4: T4, v5: T5) => R): Observable<R>;
 export function zipStatic<T, T2, T3, T4, T5, T6, R>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>, project: (v1: T, v2: T2, v3: T3, v4: T4, v5: T5, v6: T6) => R): Observable<R>;
-export function zipStatic<R>(...observables: Array<ObservableInput<any> | ((...values: Array<any>) => R)>): Observable<R>;
-export function zipStatic<R>(array: ObservableInput<any>[]): Observable<R>;
-export function zipStatic<R>(array: ObservableInput<any>[], project: (...values: Array<any>) => R): Observable<R>;
```
david-driscoll: I wouldn't remove these, instead just make sure the new ones are above them.

The reason being this allows the explicit typing of the return type, without having to type the input type.

saneyuki: My basic stance is that  it would causes a false negative case to detect compilation error if we leave `any` type on this.

> The reason being this allows the explicit typing of the return type, without having to type the input type.

Should we open a door to the case to customize the input/return type from this moment? I'm worry this might be a pitfail..

david-driscoll: While it's a possible pitfall, it would also be almost impossible to get around the case without some extremely painful casting or just simply `(<Function>Observable.zip)(....)`.

The typings should be a definition of _all of the capabilities_, not a subset that is arbitrarily "correct".  What is correct to you or me, most definitely won't be correct for everyone.

saneyuki: @david-driscoll 

> While it's a possible pitfall, it would also be almost impossible to get around the case without some extremely painful casting or just simply `(<Function>Observable.zip)(....)`.

I feel that's is the case which we should define as explicitly type signature in the upstream! Is it a difficult corner case to introduce it to the upstream.


> The typings should be a definition of all of the capabilities, not a subset that is arbitrarily "correct". What is correct to you or me, most definitely won't be correct for everyone.

I agree to accept the capabilities. But I also think that we cannot we should be careful to accept a pitfall trade-off, and we cannot accept all capabilities to design an APIs which equals to create an arbitrarily subset. I think we should introduce a new capability with the actual scenario.

saneyuki: @kwonoj @blesh 

How do you think about these discussions? (related to #1640, #1627)
I'd like to hear your thoughts.

kwonoj: I'm in general favor of explicit type by avoiding `any`, but in this case bit hard to comment. Do we have some example references of those edge cases we're expecting (and how frequently if we can estimate)? it is not possible to cover all cases like these kind of operator so might need compromise, but depends on how useful it'll be.

david-driscoll: Without them you cannot from TypeScript zip an arbitrary array, or zip an array as a rest Parma. While I don't know how common it is, I also don't know how we could quantify it until someone comes along and askes the question how come my code broke? Or how come I can't do `Observable.zip(...args)`?

kwonoj: >  how we could quantify it until someone comes along and askes 

: yes, that's right. I'm leaning on to allow this as long as it's legit case though there's some pitfall, for TypeScript user type definition becomes blocking issue to use api.

saneyuki: @kwonoj @david-driscoll 

The definitions of Rx v4 does not have the case of `any`. Is it really problem to lack the case of `any` for at this moment (at the first release timeline)?

kwonoj: >  Rx v4 does not have 

: maybe v5 having more TS userbases? (not 100% sure though).

> Is it really problem to lack the case of any for at this moment (at the first release timeline)

: my take on this is it can be acceptable for beta, but would avoid type definition / signature failure with official release once its introduced. As mentioned, for TS user type definition is critial issue even if it can be bypassed via cast to any since it makes to lose all type inferences. At this moment, I'm seeing this is tradeoff between allowing pitfall vs. avoid possible type definition failure on certain cases. Just hard to pick which would be more frequent. 

Might be available to disable those in next beta and gather some opinions?

david-driscoll: I'd be fine with that, I just worry we'll start breaking people by accident, or cause problems in the future.

david-driscoll:
I have a reservations as discussed but it may or may not an issue in the long term.  So LGTM with @kwonoj's comments to give a shot with beta and see what the feedback is.

saneyuki:
squashed

saneyuki:
rebased

blesh:
Looks like it's in need of another rebase, @david-driscoll ...

saneyuki:
@blesh @kwonoj (cc: @david-driscoll )

rebased.

saneyuki:
Sort with https://github.com/ReactiveX/rxjs/pull/1631#issuecomment-223558604

kwonoj:
Looks good to me. I'll check this in around late today.

PR #1631, opened 2016-04-17
chore(typings): increase combineLatest's signature compatibility with v4

saneyuki:
- retry #1626
- Add the case for ~`T9` type parameters which are defined in v4
  - https://github.com/Reactive-Extensions/RxJS/blob/9d6ea94244c2f40a29373f9ed6b12d951d3f2238/ts/core/linq/observable/combinelatestproto.ts
  - https://github.com/Reactive-Extensions/RxJS/blob/9d6ea94244c2f40a29373f9ed6b12d951d3f2238/ts/core/linq/observable/combinelatest.ts
- Restrict cases which accepts arrays as the 1st argument of `combineLatest()`
  - From `any` to `T`.
  - As my thought, It would be rare case to pass multiple types with stuffing into an array.

### DIFF
--- a/src/operator/combineLatest.ts
+++ b/src/operator/combineLatest.ts
@@ -80,9 +80,9 @@ export interface CombineLatestSignature<T> {
   <T2, T3, T4, T5>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>): Observable<[T, T2, T3, T4, T5]>;
   <T2, T3, T4, T5, T6>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>): Observable<[T, T2, T3, T4, T5, T6]>;
 
-  <R>(...observables: Array<ObservableInput<any> | ((...values: Array<any>) => R)>): Observable<R>;
-  <R>(array: ObservableInput<any>[]): Observable<R>;
-  <R>(array: ObservableInput<any>[], project: (...values: Array<any>) => R): Observable<R>;
+  <R>(...observables: Array<ObservableInput<T> | ((...values: Array<T>) => R)>): Observable<R>;
+  <R>(array: ObservableInput<T>[]): Observable<Array<T>>;
+  <TOther, R>(array: ObservableInput<TOther>[], project: (v1: T, ...values: Array<TOther>) => R): Observable<R>;
 }
 /* tslint:enable:max-line-length */
 
--- a/src/observable/combineLatest.ts
+++ b/src/observable/combineLatest.ts
@@ -11,17 +11,20 @@ export function combineLatest<T, T2, T3>(v1: ObservableInput<T>, v2: ObservableI
 export function combineLatest<T, T2, T3, T4>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, scheduler?: Scheduler): Observable<[T, T2, T3, T4]>;
 export function combineLatest<T, T2, T3, T4, T5>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, scheduler?: Scheduler): Observable<[T, T2, T3, T4, T5]>;
 export function combineLatest<T, T2, T3, T4, T5, T6>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>, scheduler?: Scheduler): Observable<[T, T2, T3, T4, T5, T6]>;
+
+export function combineLatest<T, R>(v1: ObservableInput<T>, project: (v1: T) => R, scheduler?: Scheduler): Observable<R>;
 export function combineLatest<T, T2, R>(v1: ObservableInput<T>, v2: ObservableInput<T2>, project: (v1: T, v2: T2) => R, scheduler?: Scheduler): Observable<R>;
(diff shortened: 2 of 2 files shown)


saneyuki:
@kwonoj How about this?

kwonoj:
Same question as https://github.com/ReactiveX/rxjs/pull/1627 . 

I'm seeing `combineLatest` has some usecases of having different types (such as using 1 for singal intent and second for value taking, etcs) - in that case array input will require explicit manual casting?

src/operator/combineLatest.ts
```diff
@@ -94,15 +100,21 @@ export function combineLatestStatic<T, T2, T3>(v1: ObservableInput<T>, v2: Obser
...
 export function combineLatestStatic<T, T2, T3, T4, T5, R>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, project: (v1: T, v2: T2, v3: T3, v4: T4, v5: T5) => R, scheduler?: Scheduler): Observable<R>;
 export function combineLatestStatic<T, T2, T3, T4, T5, T6, R>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>, project: (v1: T, v2: T2, v3: T3, v4: T4, v5: T5, v6: T6) => R, scheduler?: Scheduler): Observable<R>;
-export function combineLatestStatic<R>(...observables: Array<ObservableInput<any> | ((...values: Array<any>) => R) | Scheduler>): Observable<R>;
-export function combineLatestStatic<R>(array: ObservableInput<any>[], scheduler?: Scheduler): Observable<R>;
-export function combineLatestStatic<R>(array: ObservableInput<any>[], project: (...values: Array<any>) => R, scheduler?: Scheduler): Observable<R>;
```
david-driscoll: Again I wouldn't remove these, instead just make sure the new ones are above them.

The reason being this allows the explicit typing of the return type, without having to type the input type.

blesh: Is six deep really necessary? Honest question.

david-driscoll: I see 6 as being the highest amount you'd actually want to have, generally speaking I think I've only ever hit 4 before things go sideways.  9 honestly seems excessive to me.

kwonoj: I also think ~T9 seems bit excessive, what about to have ~T6 for now (as consistent with other current operators) to see if there's any actual usecases of more than 6? @saneyuki , do you actually have usecases?

saneyuki: @kwonoj 

> if there's any actual usecases of more than 6? @saneyuki , do you actually have usecases?

Yes. There are some actual usecases in the code base of my place of work (I cannot indicate them because they are not open sourced). 

Their code uses `combineLatest()` to develop a input form with Rx like following:

```javascript
const input1 = ...;
const input2 = ...;
const input3 = ...;
const input4 = ...;
const input5 = ...;
const input6 = ...;
const input7 = ...;

function combine(v1: T, v2: T2, v3: T3, v4: T4, v5: T5, v6: T6, v7: T7): FormViewModel {
  return new FormViewModel(v1, v2, v3, v4, v5, v6, v7);
}

const form = Rx.Observable.combineLatest(input1, input2, input3, input4, input5, input6, input7, combine);
```

- If we need more discussion about to introduce ~`T9` type parameters, I remove their changes from this pull request and I'll open a new pull request to add only them.
- As a this library's design, we cannot introduce ~`T9` type parameters at this moment, I (and my co-workers and places of work) may be able to accept its decision.

kwonoj: just curious, what about array?

```js
const input1 = ...;
const input2 = ...;
const input3 = ...;
const input4 = ...;
const input5 = ...;
const input6 = ...;
const input7 = ...;

combinLatest([nput1, input2, input3, input4, input5, ...]);
```

saneyuki: @kwonoj 

> just curious, what about array?

They have a different type, so we thought it is convenient way that to pass them separately in using a Rx v4.

With to define a tuple type and using destructuring assignment, we can use an array, but it need to define a temporary `Tuple` for convenience
```
type Tuple = [T1, T2, T3, T4, T5, T6, T7];
combinLatest([input1, input2, input3, input4, input5, ...]: Tuple, ([v1, v2, v3, v4, v5, ...]: Tuple):  => {
  ...
});
```

kwonoj: I assume different type with array can be solved via https://github.com/ReactiveX/rxjs/pull/1640? (https://github.com/ReactiveX/rxjs/pull/1640/files#diff-60de9008744d220f5a254bb9454716edR483)

saneyuki: @kwonoj 

- we already have `combineLatestStatic<T, R>(v1: ObservableInput<T>, project: (v1: T) => R, scheduler?: Scheduler): Observable<R>`. This would resolve to use an array with a tuple type.   - #1640 is not related to this usecase.
- #1640 still have an `any` type for this. This would causes a false negative case.

david-driscoll:
I have a reservations related to the discussion on #1627 but it may or may not an issue in the long term.  So LGTM with @kwonoj's comments to give a shot with beta and see what the feedback is.

kwonoj:
As shared above comment, I'm feeling more than ~6 signature seems bit excessive and need to conclusion on this - anyone pushing > 6 ? (I know @saneyuki will +1 :) )

saneyuki:
may related (long term): #5453

saneyuki:
squashed

saneyuki:
rebased

saneyuki:
@blesh @kwonoj (cc: @david-driscoll )

rebased.

kwonoj:
What's conclusions for having ~T9? I was suggesting limit ~T6 at this point (https://github.com/ReactiveX/rxjs/pull/1631#issuecomment-216583709) as well as @david-driscoll does.

saneyuki:
@kwonoj 

> I was suggesting limit ~T6 at this point (#1631 (comment))

Sorry. I had misunderstanding. I fixed it.

PR #1637, opened 2016-04-19
refactor(symbol): replace symbol polyfill to separate package

### LINKED ISSUE #1636: Update to use symbol-observable module
We should update this repository to use [symbol-observable](https://www.npmjs.com/package/symbol-observable) rather than having its own.

kwonoj:
**Description:**

This PR replaces existing symbol declaration into `symbol-observable` package.

closes #1636

**Notes**
Blocked by https://github.com/blesh/symbol-observable/pull/4 .


blesh:
@kwonoj FYI: the issue is resolved and a new version of `symbol-observable` has been published.

src/Observable.ts
```diff
@@ -3,12 +3,13 @@ import {Operator} from './Operator';
 import {Subscriber} from './Subscriber';
 import {Subscription, AnonymousSubscription, TeardownLogic} from './Subscription';
 import {root} from './util/root';
-import {$$observable} from './symbol/observable';
 import {toSubscriber} from './util/toSubscriber';
 
 import {IfObservable} from './observable/IfObservable';
 import {ErrorObservable} from './observable/ErrorObservable';
 
+import * as observableSymbol from 'symbol-observable';
```
blesh: Nitpick: I prefer `$$name` for symbols. Can we use `$$observable`? It's consistent with the rest of the library.

blesh:
So just one nitpick, other than that, it LGTM.

src/Observable.ts
kwonoj: Absolutely, updated per suggestion.

kwonoj:
I'll wait travis to be happy before check this in.

blesh:
Well, right now it looks like CJS will compile, but ES6 will not, for whatever reason. When compiling the ES6 it can't find the module.

blesh:
Specifically: `npm run build_cjs` works, but `npm run build_es6` does not

blesh:
Ugh, I'd bet anything that TypeScript is expected all imported modules from `node_modules` to be ES2015. :( Welcome to hell.

kwonoj:
Yes, noticed that, thanks for pointing out. Looking into it now..

kwonoj:
> Ugh, I'd bet anything that TypeScript is expected all imported modules from node_modules to be ES2015. :( Welcome to hell.

: Uh-oh, :fire: I'm burning..

kwonoj:
Just fyi, `moduleResolution` solves es15 build but not `systemjs-builder` complains about external dependencies. :/ looking into it also..

package.json
```diff
@@ -196,5 +196,8 @@
   "engines": {
     "npm": ">=2.0.0"
   },
-  "typings": "./dist/cjs/Rx.d.ts"
+  "typings": "./dist/cjs/Rx.d.ts",
+  "dependencies": {
+    "symbol-observable": "^0.2.2"
```
kwonoj: I've changed this to `dependencies` instead of `devDependencies` since it's required dependency not only dev perspective.

tools/make-system-bundle.js
```diff
@@ -4,8 +4,11 @@ var addLicenseToFile = require('./add-license-to-file');
 
 var config = {
   baseURL: 'dist',
+  meta: {
+    'symbol-observable': { build: false }
```
kwonoj: Since this is external dependency, I've marked to exclude in systemjs builder to not to include it and resolve it externally. I'm not daily user of systemjs, so not sure if this is correct - sending out changes to see if there's correct recommendation if I did wrong.

kwonoj:
@blesh - I've updated 2 parts more,

- `symbol-observable` is now `dependency` instead of `devDependency`
- systemjs bundler will mark `symbol-observable` externally resolved, let each consumer to provide it

and I believe both need to be approved. Will leave this to get suggestions & corrections if there's wrong I did.

tools/make-system-bundle.js
blesh: cc/ @robwormald

robwormald:
the use case for the system bundles is to provide tool-free usage, so it seems to be it would be simpler just to include it, as its unlikely to be provided elsewhere, no?

kwonoj:
> so it seems to be it would be simpler just to include it, as its unlikely to be provided elsewhere, no?

: Yes, I wasn't sure if current change's legit so would like to ask to confirm it. I'll try to include it as well, my initial attempt wasn't successful while resolving external modules.

kwonoj:
Updated PR as suggested.

kwonoj:
Merged with b626d97 .

PR #1640, opened 2016-04-21
Improved typings for combineLatestStatic, forkJoin and zipStatic

david-driscoll:
**Description:**
I noticed a few ergonomic issues with using `combineLatest` and `zip`. The single tuple type doesn't really make much sense, plus it conflicts with sending in an array value.

Also added typings for `forkJoin` as it's fairly similar to `combineLatest` in shape (obviously the behaviors are different).

### DIFF
--- a/spec/observables/combineLatest-spec.ts
+++ b/spec/observables/combineLatest-spec.ts
@@ -1,6 +1,6 @@
 import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
-declare const {hot, cold, expectObservable, expectSubscriptions};
+declare const {hot, cold, expectObservable, expectSubscriptions, type};
 
 const Observable = Rx.Observable;
 const queueScheduler = Rx.Scheduler.queue;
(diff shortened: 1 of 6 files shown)


kwonoj:
What's relation between other 2 PR https://github.com/ReactiveX/rxjs/pull/1631 / https://github.com/ReactiveX/rxjs/pull/1627 ?

david-driscoll:
It's not really related to those, although it looks like there will be a conflict that needs to be resolved.

I just found that trying to use an array with combineLatest caused the typings to freak out quiet a bit, this is just to help with that case.

kwonoj:
I see, thanks for clarification.

src/operator/combineLatest.ts
```diff
@@ -88,21 +88,23 @@ export interface CombineLatestSignature<T> {
...
-export function combineLatestStatic<R>(...observables: Array<ObservableInput<any> | ((...values: Array<any>) => R) | Scheduler>): Observable<R>;
+export function combineLatestStatic<T>(array: ObservableInput<T>[], scheduler?: Scheduler): Observable<T[]>;
 export function combineLatestStatic<R>(array: ObservableInput<any>[], scheduler?: Scheduler): Observable<R>;
+export function combineLatestStatic<T, R>(array: ObservableInput<T>[], project: (...values: Array<T>) => R, scheduler?: Scheduler): Observable<R>;
 export function combineLatestStatic<R>(array: ObservableInput<any>[], project: (...values: Array<any>) => R, scheduler?: Scheduler): Observable<R>;
+export function combineLatestStatic<T>(...observables: Array<ObservableInput<T> | Scheduler>): Observable<T[]>;
+export function combineLatestStatic<T, R>(...observables: Array<ObservableInput<T> | ((...values: Array<any>) => R) | Scheduler>): Observable<R>;
+export function combineLatestStatic<R>(...observables: Array<ObservableInput<any> | ((...values: Array<any>) => R) | Scheduler>): Observable<R>;
```
saneyuki: I'm worry about this `any` case which would cause a false negative for a compiling error.

david-driscoll: It's a fall through case to let you specify what the emitted type will be.  without that, you're stuck casting the inputs  `T` to massage them to be your return type.

Basically just trying to make things flexible enough that you don't get blocked into any corners.

david-driscoll:
@blesh rebased, may need an additional rebase after #1627 and #1631 land.

kwonoj:
change looks good to me.

kwonoj:
I'll check this in later today to tomorrow if there isn't other suggestions around.

kwonoj:
@david-driscoll , would you mind to rebase this one?

david-driscoll:
@kwonoj rebased :+1:

PR #1645, opened 2016-04-24
docs(MIGRATION): update distinctUntilChanged interface change

### LINKED ISSUE #1563: Mention distinctUntilChanged API change in migration guide
The function signature changed but I can't find it mentioned anywhere. See https://github.com/Reactive-Extensions/RxJS/issues/1188

kwonoj:
**Description:**
This PR adds one more table into migration document for operators does not split but only interface has been changed, cases like `distinctUntilChanged`.

closes #1563

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -220,6 +220,27 @@ To reduce polymorphism and get better performance out of operators, some operato
 </table>
 
 
+## Operator Interface Changes
+
+
+<table>
+  <thead>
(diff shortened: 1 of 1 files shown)


kwonoj:
Let me check this in around tomorrow morning if there isn't other suggestions around.

blesh:
Merged with 7a551cc. Thanks @kwonoj

PR #1652, opened 2016-04-26
Remove unused source variable

benjamingr:
**Description:**

Introduced in an earlier commit where it makes sense,
Doesn't really make sense now. Removed and used `this`
directly.


blesh:
LGTM

blesh:
Merged with 7cd8a43. Thanks @benjamingr!

Note, I changed the commit message to fit our conventions [which are listed here](https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md#commit)

benjamingr:
Hehe, OK, that's even stricter than the node guidelines :D

blesh:
It's mostly so they can be grepped, and even more than that so we can use conventional-changelog to generate the CHANGELOG.md.

Issue #1629, opened 2016-04-15
Support .Publish(...)


Dorus:
For reference, the [Rx Design Guidelines](http://blogs.msdn.com/b/rxteam/archive/2010/10/28/rx-design-guidelines.aspx) [5.10](http://xgrommx.github.io/rx-book/content/guidelines/using/index.html#use-the-publish-operator-to-share-sideeffects) suggest to use this operator.

The operator `.publish()` returns an `ConnectableObservable`, this means you first need to subscribe all your sources, `.connect()` it, keep a reference to your connect to prevent leaks, and then you also end up with a hot observable, so downstream subscribers better be ready as well.

Most of the time you only want to duplicate a stream locally. RxJS 4 and many other Rx implementations allow you to write 

    xs.publish(ob => ob.zip(ob.skip(1), (x, y) => {...}))

Here we use publish to peek at the next item in the source. Another favorite of mine is

    xs.publish(ob => ob.window(ob.debounce(100)))

Because often you want to use a source for its own window-boundary.

This operator is very similar to `.let`, but avoid subscribing to the source twice. Useful on cold observables, but also to avoid the overhead of subscribing multiple times to a hot source. (imagine you use this trick 4x in a row with `let`, you end up with 2^4th subscriptions.)

kwonoj:
Would you elaborate bit more? so far I know current `publish` already returns `ConnectableObservable` (https://github.com/ReactiveX/rxjs/blob/34cdef5a3d652c75a546e5bd73733fd21dc64fc0/src/operator/publish.ts#L15) would you mind to share some snippet of pusedo for suggested behavior?

Dorus:
The operator `publish([selector])` should not return an `ConnectableObservable`, instead it returns the result of creating the connectable observable, passing the connectable observable to the `selector`, 

aka
```JS
function publish(source, selector) {
  return Rx.Observable.create(ob => {
    var xs = source.publish();
    return selector(xs).subscribe(ob).add(xs.connect());
  });
}
```
Similar to publish in [RxJS 4](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/publish.md) and [Rx.Net](https://msdn.microsoft.com/en-us/library/hh229147%28v=vs.103%29.aspx).

> selector
    Type: System.Func<IObservable<TSource>, IObservable<TResult>>
    The selector function which can use the multicasted source sequence as many times as needed, without causing multiple subscriptions to the source sequence.

[...]

PR #1659, opened 2016-04-26
Support optional selector for publish / multicast

kwonoj:
**Description:**
This PR updates behavior of `multicast` and `publish` as well, support optional `selector` functions can be supplied.

#1629

***Note***
As I'm not frequent user of `publish`/`multicast`, have gut feeling this PR might be conceptually incorrect at all. Please feel freely point out / or close PR if this isn't correct approach.

### DIFF
--- a/src/add/operator/multicast.ts
+++ b/src/add/operator/multicast.ts
@@ -2,7 +2,7 @@
 import {Observable} from '../../Observable';
 import {multicast, MulticastSignature} from '../../operator/multicast';
 
-Observable.prototype.multicast = multicast;
+Observable.prototype.multicast = <any>multicast;
 
 declare module '../../Observable' {
   interface Observable<T> {
--- a/src/observable/MulticastObservable.ts
+++ b/src/observable/MulticastObservable.ts
@@ -1,0 +1,20 @@
+import {Observable} from '../Observable';
+import {Subscriber} from '../Subscriber';
+import {Subscription} from '../Subscription';
+import {ConnectableObservable} from '../observable/ConnectableObservable';
+
+export class MulticastObservable<T> extends Observable<T> {
+  constructor(protected source: Observable<T>,
+              private connectable: ConnectableObservable<T>,
+              private selector: (source: Observable<T>) => Observable<T>) {
+    super();
+  }
+
+  protected _subscribe(subscriber: Subscriber<T>): Subscription {
+    const {selector, connectable} = this;
+
+    const subscription = selector(connectable).subscribe(subscriber);
+    subscription.add(connectable.connect());
+    return subscription;
+  }
+}
--- a/src/operator/multicast.ts
+++ b/src/operator/multicast.ts
@@ -1,4 +1,6 @@
 import {Subject} from '../Subject';
+import {Observable} from '../Observable';
+import {MulticastObservable} from '../observable/MulticastObservable';
 import {ConnectableObservable} from '../observable/ConnectableObservable';
 
 /**
@@ -7,7 +9,10 @@ import {ConnectableObservable} from '../observable/ConnectableObservable';
  *
  * <img src="./img/multicast.png" width="100%">
  *
- * @param {Function} selector - a function that can use the multicasted source stream
+ * @param {Function|Subject} Factory function to create an intermediate subject through
+ * which the source sequence's elements will be multicast to the selector function
+ * or Subject to push source elements into.
+ * @param {Function} Optional selector function that can use the multicasted source stream
  * as many times as needed, without causing multiple subscriptions to the source stream.
  * Subscribers to the given source will receive all notifications of the source from the
  * time of the subscription forward.
@@ -17,7 +22,8 @@ import {ConnectableObservable} from '../observable/ConnectableObservable';
  * @method multicast
  * @owner Observable
  */
-export function multicast<T>(subjectOrSubjectFactory: Subject<T> | (() => Subject<T>)): ConnectableObservable<T> {
+export function multicast<T>(subjectOrSubjectFactory: Subject<T> | (() => Subject<T>),
+                             selector?: (source: Observable<T>) => Observable<T>): Observable<T> | ConnectableObservable<T> {
   let subjectFactory: () => Subject<T>;
   if (typeof subjectOrSubjectFactory === 'function') {
     subjectFactory = <() => Subject<T>>subjectOrSubjectFactory;
@@ -26,11 +32,15 @@ export function multicast<T>(subjectOrSubjectFactory: Subject<T> | (() => Subjec
       return <Subject<T>>subjectOrSubjectFactory;
     };
   }
-  return new ConnectableObservable(this, subjectFactory);
+
+  const connectable = new ConnectableObservable(this, subjectFactory);
+  return selector ? new MulticastObservable(this, connectable, selector) : connectable;
 }
 
 export type factoryOrValue<T> = T | (() => T);
+export type selector<T> = (source: Observable<T>) => Observable<T>;
 
 export interface MulticastSignature<T> {
   (subjectOrSubjectFactory: factoryOrValue<Subject<T>>): ConnectableObservable<T>;
+  (SubjectFactory: () => Subject<T>, selector?: selector<T>): Observable<T>;
 }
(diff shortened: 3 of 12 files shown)


blesh:
@kwonoj what were the perf changes for this addition?

kwonoj:
I haven't observed noticeable regression between this changes to master, but current master itself shows some interesting numbers on my machine

**master, without this PR**
```
                      |         RxJS 4.1.0 | RxJS 5.0.0-beta.6 | factor | % improved
-------------------------------------------------------------------------------------
multicast - immediate | 9,526,302 (±3.72%) |2,132,759 (±4.14%) |  0.22x |     -77.6%
```

may possible

- weird on my machine only (as always)
- perf test design is incorrect

I've added couple of commits for micro perf test as well.

blesh:
@kwonoj it's probably the new implementation. In Rx4, `multicast` is _highly polymorphic_. What I mean by that is [it has two wildly different code paths](https://github.com/Reactive-Extensions/RxJS/blob/76b20322cea524ef0bf4b408830981ce46b4c6c8/src/core/linq/observable/multicast.js#L36-L38) depending on the parameters provided.

I haven't dug into your implementation yet, but this was one of the primary reasons this functionality wasn't reintroduced immediately.

src/add/operator/multicast.ts
```diff
@@ -2,7 +2,7 @@
 import {Observable} from '../../Observable';
 import {multicast, MulticastSignature} from '../../operator/multicast';
 
-Observable.prototype.multicast = multicast;
+Observable.prototype.multicast = <any>multicast;
```
blesh: This seems like a smell. Why do we need to do this?

src/operator/multicast.ts
```diff
@@ -26,11 +33,35 @@ export function multicast<T>(subjectOrSubjectFactory: Subject<T> | (() => Subjec
       return <Subject<T>>subjectOrSubjectFactory;
     };
   }
-  return new ConnectableObservable(this, subjectFactory);
+
+  if (selector) {
```
blesh: Here you can create a _single_ `ConnectableObservable` and give it the subjectFactory, then provide that created observable to the `MulticastObservable`:

```
const connectable = new ConnectableObservable(this, subjectFactory);
if (selector) {
  return new MulticastObservable(this, connectable, selector);
} else {
  return connectable;
}
```

src/operator/multicast.ts
```diff
@@ -26,11 +33,35 @@ export function multicast<T>(subjectOrSubjectFactory: Subject<T> | (() => Subjec
...
+    return new MulticastObservable(this, subjectFactory, selector);
+  } else {
+    return new ConnectableObservable(this, subjectFactory);
+  }
+}
+
+class MulticastObservable<T> extends Observable<T> {
```
blesh: This can probably be in it's own module.

src/add/operator/multicast.ts
kwonoj: To support different signature `multicast(subject)` / `multicast(factory, selector?)`.

As similar to other operators have different signatures [`mergeMapTo`](https://github.com/ReactiveX/rxjs/blob/7cd8a4365c369a2d6e49ffa1dde5659057f6aad8/src/operator/mergeMapTo.ts#L63-L67) / [`add/mergeMapTo`](https://github.com/ReactiveX/rxjs/blob/7cd8a4365c369a2d6e49ffa1dde5659057f6aad8/src/add/operator/mergeMapTo.ts#L6]) if there's completely different two signatures, can't be assigned into prototype directly since function definition is single, polymorphic.

src/operator/multicast.ts
```diff
@@ -17,7 +23,8 @@ import {ConnectableObservable} from '../observable/ConnectableObservable';
  * @method multicast
  * @owner Observable
  */
-export function multicast<T>(subjectOrSubjectFactory: Subject<T> | (() => Subject<T>)): ConnectableObservable<T> {
+export function multicast<T>(subjectOrSubjectFactory: Subject<T> | (() => Subject<T>),
+                             selector?: (source: Observable<T>) => Observable<T>): Observable<T> | ConnectableObservable<T> {
```
blesh: I'm not in love with the name `selector` for this. But I'm not really sure what else to call it.

kwonoj: me too. followed RxJS 4 convention, since wasn't able to come up with better name.

blesh:
I've added a few comments, and pointed out one area I can see there might be a perf bottleneck.

src/operator/multicast.ts
blesh: Also, why not give it a static `create` function? It would then be possible to use it like:

```js
Observable.multicast(source, () => new Subject(), 
  (shared) => shared.map(x => x + '!')
                    .merge(shared.map(x => x + '?')));
```

src/operator/multicast.ts
staltz: `sandbox`?
`container`?
`shareAndJoin`?
`shareJoin`?
`divergeAndConverge`?
`converge`?
`work`?
`action`?

kwonoj:
Updated PR per suggestions. 

One thing about static `create` for `multicast` is not included, after give a couple of try I found it's quite challenging to create static signature due to overloaded signatures of `multicase` is completely different and incompatible which doesn't allow simple method signature overloading in TypeScript. May need some more exploration to see if I can bring it, will followup with separate PR once I can make it work.

blesh:
>  due to overloaded signatures of multicase is completely different and incompatible which doesn't allow simple method signature overloading in TypeScript.

Yeah, we ran into this with `Subject.create` as well. See that for a workaround. Also there's an issue I filed about it: https://github.com/Microsoft/TypeScript/issues/4628

kwonoj:
I've tried to apply similar practices as `Subject.create`, seems I did something not correctly. 

Basically what need to be done is (in pseudo)

```js
class Observable {
 static multicast(subjectOrSubjectFactory: Subject<T> | (() => Subject<T>))
        : ConnectableObservable<T>;
 static multicast(SubjectFactory: () => Subject<T>,
                 selector?: (source: Observable<T>) => Observable<T>): Observable<T> {
 }
}
```

function signature overloading in `Observable` without inheritance being applied. workaround for `Observable.create` seems doesn't allow to create method overload signatures - is there way for this?

/cc @david-driscoll also for visibility / tips.

blesh:
@kwonoj how is perf after the change?

blesh:
Where is @jayphelps with his 5th element "multicast" meme?

jayphelps:
image

Sorry I'm late!

kwonoj:
Numbers with current changes are below.

```
                                |         RxJS 4.1.0 |   RxJS 5.0.0-beta.6 |    factor |   % improved
------------------------------------------------------------------------------------------------------
 multicast-selector - immediate |    32,802 (±3.46%) |     59,618 (±3.62%) |     1.82x |        81.7%


                                |         RxJS 4.1.0 |   RxJS 5.0.0-beta.6 |    factor |   % improved
------------------------------------------------------------------------------------------------------
          multicast - immediate | 8,695,218 (±3.18%) |  2,351,401 (±2.43%) |     0.27x |       -73.0%
```

`muitlcast` itself doesn't seem regresses.

blesh:
@kwonoj I think we can merge this as-is and circle back to perf. I have a few ideas.

blesh:
Merged with 0f8bf28. Thanks @kwonoj

staltz:
By the way, I know it's kind of late, but I just remembered that this should also be done for `publishReplay` and `publishBehavior`, since they exist in both RxJS v4 and RxJava.

src/observable/MulticastObservable.ts
```diff
@@ -0,0 +1,20 @@
...
+import {ConnectableObservable} from '../observable/ConnectableObservable';
+
+export class MulticastObservable<T> extends Observable<T> {
+  constructor(protected source: Observable<T>,
+              private connectable: ConnectableObservable<T>,
+              private selector: (source: Observable<T>) => Observable<T>) {
+    super();
+  }
+
+  protected _subscribe(subscriber: Subscriber<T>): Subscription {
```
trxcllnt: @kwonoj shouldn't this really be done with an operator? Why introduce another Observable type?

kwonoj: I didn't have strong reasoning behind this, actually. So suggest to remove this observable type?

trxcllnt: Well, it'll make the ConnectableObservable changes I'm working on easier to integrate.

PR #1663, opened 2016-04-28
fix(Rx): remove kitchenSink and DOM, let Rx exports all

### LINKED ISSUE #1650: Proposal: No more KitchenSink
We should probably just have one global export file for the masses.

1. Is it huge? Yes.
2. Should people use RxJS 5 that way? Hell no. Use a bundler and only import what you're using.
3. In past tests, I noticed the gzipped size of Rx.KitchenSink.min.js was about the same as Rx.min.js, because of how the compression affected it.

If people are using the global export version of RxJS, it can be assumed that they're probably hacking around, and/or not really serious about reducing the size of their downloaded application.

At this point it's a little confusing to say "this is in Rx, but this is in Kitchensink" to people. Also, given how modular this library is, it would be easy enough for anyone to make their own custom build of RxJS 5 to keep around.

### Basically the idea is just to put everything in Rx.ts

Thoughts?

kwonoj:
**Description:**
This PR removes `Rx.KitchenSink`, let `Rx` exports everything instead as well as does not populate build anymore.

closes #1650


esdoc.json
```diff
@@ -10,11 +10,10 @@
   "title": "RxJS",
   "styles": ["./doc/styles/main.css"],
   "scripts": [
-    "./dist/global/Rx.KitchenSink.umd.js",
+    "./dist/global/Rx.umd.js",
     "./doc/asset/devtools-welcome.js",
     "./doc/scripts/custom-manual-styles.js",
-    "./doc/decision-tree-widget/dist/decision-tree-widget.min.js",
-    "./doc/scripts/setup-rx-script.js"
+    "./doc/decision-tree-widget/dist/decision-tree-widget.min.js"
```
kwonoj: /cc @staltz for document generate script related changes.

blesh: @kwonoj is this blocking? Because the rest of this PR looks great to me.

kwonoj: No, for visibility just in case if I miss anything. esdoc script is working on my local machine with this changes.

blesh:
@kwonoj can you get rid of `Rx.DOM.ts` as well? We can just merge it into `Rx.ts` too. It's silly and confusing to have two files where one is a superset of the other, IMO.

kwonoj:
Will do that as well. just got back and caught conversation's ongoing.

kwonoj:
quick confirmation @blesh - `Rx` suppose to not to export `Rx.dom` specific such as `requestanimationframeScheduler`, etcs, is that correct? or just export everything?

kwonoj:
Updated PR to remove `DOM` as well. @blesh - for now I made let `Rx` exports everything, let me know if this isn't expected to drop off some of exports.

blesh:
LGTM :shipit:

kwonoj:
Thanks for review / confirmation. I'll check this in late tonight to tomorrow.

blesh:
@kwonoj, be sure there is a breaking change message in the commit message. Since we're not putting those files out anymore. Will probably have to prefix it `fix` out `feat` and not `chore`

kwonoj:
ah, thanks for pointing out. Let me update commit message now.

kwonoj:
updated commit message as well.

blesh:
"Let Rx _export_ all"

:)

kwonoj:
Merged with f5090b4.

PR #1667, opened 2016-04-29
Improve finally tests

blesh:
- Adds additional, virtually scheduled tests to the finally battery of tests. Covers basic usage.
- Renames `finallySelector` as just `callback`, because it doesn't really "select" anything.

### DIFF
--- a/spec/operators/finally-spec.ts
+++ b/spec/operators/finally-spec.ts
@@ -1,7 +1,10 @@
 import {expect} from 'chai';
-import * as Rx from '../../dist/cjs/Rx';
+import * as Rx from '../../dist/cjs/Rx.KitchenSink';
 
 const Observable = Rx.Observable;
+
+declare const {asDiagram, expectObservable, Symbol, type};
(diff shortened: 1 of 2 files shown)


spec/operators/finally-spec.ts
```diff
@@ -71,4 +74,90 @@ describe('Observable.prototype.finally', () => {
...
+    let executed = false;
+    let s1 = hot(  '--a--b--c--#');
+    let subs =     '^          !';
+    let expected = '--a--b--c--#';
+    let result = s1.finally(() => executed = true);
+    expectObservable(result).toBe(expected);
+    expectSubscriptions(s1.subscriptions).toBe(subs);
+    // manually flush so `finally()` has chance to execute before the test is over.
+    rxTestScheduler.flush();
+    expect(executed).to.equal(true);
```
kwonoj: just note : now true / false assertion can be written as `expect(something).to.be.true`

kwonoj:
Change looks good to me.

PR #1674, opened 2016-04-29
fix(ScalarObservable): set _isScalar to false when initialized with a scheduler.

trxcllnt:
**Description:**
ScalarObservable's `_isScalar` flag was permanently `true`, even if initialized with a `scheduler`. This caused our internal `subscribeToResult` helper to short-circuit subscribing to the `ScalarObservable`, so the emission was never scheduled. This PR changes that.

None yet.

### DIFF
--- a/src/observable/ScalarObservable.ts
+++ b/src/observable/ScalarObservable.ts
@@ -34,6 +34,9 @@ export class ScalarObservable<T> extends Observable<T> {
 
   constructor(public value: T, private scheduler?: Scheduler) {
     super();
+    if (scheduler) {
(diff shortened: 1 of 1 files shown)


kwonoj:
Can we think of some test cases for this? (not sure, just throwing out idea)

blesh:
Agree. Should add a test or two

blesh:
I'm going to add a couple of tests and merge this today. No worries, @trxcllnt

blesh:
Merged as of bea15ca... thanks @trxcllnt!

Issue #1672, opened 2016-04-29
finally is not passing observable through properly


johnpapa:
**RxJS version:**
rxjs 5 beta 6

**Code to reproduce:**
```
  getSessions() {
    this.spinnerService.show();
    return <Observable<Session[]>>this.http
      .get(sessionsUrl)
      .map(res => this.extractData<Session[]>(res))
      .catch(this.exceptionService.catchBadResponse)
      .finally(() => this.spinnerService.hide());
  }
```

See code example here ...
https://github.com/johnpapa/event-view/commit/da21a922faf74aa3f2b6fbf1af5d1d666419b24e#diff-a115014ed5c151b197295d50f1ce2a54R24

**Expected behavior:**
TypeScript's `tsc` should think `getSessions` should return `Observable<Session[]>` 

**Actual behavior:**
TypeScript's `tsc` thinks `getSessions` is  returning `Observable<{}>` 

**Additional information:**
- When I comment out the `finally` TypeScript is happy
- When I cast the return type  as `<Observable<Session[]>>` tsc is happy

**Error message:**
```
> tsc -p ./src/client

src/client/app/+sessions/session-list/session-list.component.ts(34,9): error TS2322: Type '{}' is not assignable to type 'Session[]'.
```

cc // @david-driscoll @blesh @robwormald @wardbell

kwonoj:
Think I found cause.

kwonoj:
@johnpapa , I believe this is fixed by latest changes in master branch. If you're ok please check with master, or it'll be published in next release. Thanks for catching this.

blesh:
@johnpapa next release will be Monday unless I get a deluge of requests prior to ng-conf.

johnpapa:
trying it now.

just noticed i have this line ... which is not what you changed. i am on beta 6 (as it matches angular 2 beta 17)
```
   <T>(finallySelector: () => void): Observable<T>;
```

kwonoj:
Yes, that's prior to master including changes. Next package will be published in next week as @blesh mentioned, or for immediate trying you may need to pull down master branch.

johnpapa:
![image](https://cloud.githubusercontent.com/assets/1202528/14932229/4bab13c8-0e41-11e6-9321-98beade05681.png)

johnpapa:
Same error :(

david-driscoll:
worst case you could manually set the type `.finally<Session[]>(() => this.spinnerService.hide());`

[...]

PR #1675, opened 2016-04-30
chore(typings): updated catch signature to allow T to flow through

david-driscoll:
**Description:**
This allows catch to flow it's types by default, but also allows for additional casting if needed.

#1672


kwonoj:
Cool, this is reason I wasn't able to repro since I didn't try with catch? :/

david-driscoll:
Catch was causing `Observable<any>` to start flowing through, which isn't quiet correct :(

kwonoj:
Merged with 56a12ba, appreciate for quick triaging @david-driscoll

PR #1680, opened 2016-05-03
perf(combineLatest): avoid splice and indexOf

staltz:
Small speed up of `combineLatest` according to micro-perf benchmarks. CombineLatestSubscriber notifyNext() has `toRespond` as a number, not an array which is then spliced or searched with indexOf. We prefill the `values` array with a so-called `none` object, which is useful for the calculation of the `toRespond` number.

Before 
```

                                         |                     RxJS 4.1.0 |              RxJS 5.0.0-beta.7 |          factor |      % improved
               combinelatest - immediate |                19,142 (±4.60%) |               239,359 (±1.12%) |          12.50x |        1,150.4%
                           combinelatest |                18,787 (±1.83%) |               134,440 (±1.75%) |           7.16x |          615.6%
```
                           
After
```
                                        |                     RxJS 4.1.0 |              RxJS 5.0.0-beta.7 |          factor |      % improved
              combinelatest - immediate |                17,281 (±0.99%) |               263,042 (±1.21%) |          15.22x |        1,422.1%
                          combinelatest |                17,203 (±1.64%) |               144,099 (±1.57%) |           8.38x |          737.6%
```

### DIFF
--- a/src/operator/combineLatest.ts
+++ b/src/operator/combineLatest.ts
@@ -8,6 +8,7 @@ import {Subscriber} from '../Subscriber';
 import {OuterSubscriber} from '../OuterSubscriber';
 import {InnerSubscriber} from '../InnerSubscriber';
 import {subscribeToResult} from '../util/subscribeToResult';
+const none = {};
 
 /**
  * Combines multiple Observables to create an Observable whose values are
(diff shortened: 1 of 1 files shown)


blesh:
Merged with 33599cd... thanks @staltz

PR #1682, opened 2016-05-04
Fix npmcdn link

mjackson:
The default on npmcdn is to redirect to the latest version. The previous link 404s.


blesh:
LGTM! Thanks @mjackson!

Issue #1677, opened 2016-05-02
zipAll not working as expected ?


virtualvoid:
**RxJS version:** 5.0.0-beta.7

**Code to reproduce:** http://jsbin.com/milute/edit?js,console

**Expected behavior:** at the subscribe callback i expected to see the length of array that was made by zipAll

**Actual behavior:** something with ZipBufferIterator which i don't understand

**Additional information:** what's wrong with my or yours code ?

Thanks

virtualvoid:
for better explanation what I need to do :

```
    Observable
      .from(this.resourceStaffs)
      .map((resourceStaff: ResourceStaff, index: number) => this._resourceService.saveResourceStaff(resourceStaff, index))
      .zipAll()
      .subscribe((objectIndices: Array<ObjectIndex<ResourceStaff>>) => {
        debugger;
     ...
```
I have array of objects and I need to save them (using ng2 http) one-by-one, wait for all results and then zip it into one result (array of objectIndices).

Can anyone help me out please ? Thanks in advance.

virtualvoid:
```
EXCEPTION: TypeError: unknown type returned
   EXCEPTION: TypeError: unknown type returned
   angular2.dev.js (25644,9)
   STACKTRACE:
   angular2.dev.js (25644,9)
   TypeError: unknown type returned
   at SafeSubscriber.prototype.error (https://localhost:44300/app/libs/rx.js:10829:11)
   at Subscriber.prototype._error (https://localhost:44300/app/libs/rx.js:10769:7)
   at Subscriber.prototype.error (https://localhost:44300/app/libs/rx.js:10749:9)
   at OuterSubscriber.prototype.notifyError (https://localhost:44300/app/libs/rx.js:805:7)
   at InnerSubscriber.prototype._error (https://localhost:44300/app/libs/rx.js:891:7)
   at Subscriber.prototype.error (https://localhost:44300/app/libs/rx.js:10749:9)
   at subscribeToResult (https://localhost:44300/app/libs/rx.js:8257:7)
   at ZipBufferIterator.prototype.subscribe (https://localhost:44300/app/libs/rx.js:2505:7)
   at ZipSubscriber.prototype._complete (https://localhost:44300/app/libs/rx.js:2358:11)
   at Subscriber.prototype.complete (https://localhost:44300/app/libs/rx.js:10755:9)
[...]

PR #1683, opened 2016-05-04
chore(typings): removed implicit Array<T> signature for combineAll and zipAll

david-driscoll:
**Description:**
The implicit definition for combineAll / zipAll cannot properly infer the resulting type.  

See #1677 for more information.


kwonoj:
Merged with 4d773bc, thanks @david-driscoll

Issue #1685, opened 2016-05-04
Update reduce marble


Dorus:
http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-reduce => http://reactivex.io/rxjs/img/reduce.png

Reduce now has the seed as the first parameter, the marble shows it as the second.

staltz:
The documentation is wrong, the source code actually has the seed as the second. https://github.com/ReactiveX/rxjs/blob/master/src/operator/reduce.ts#L23

kwonoj:
I think this is closed via https://github.com/ReactiveX/rxjs/pull/1689.

PR #1689, opened 2016-05-06
docs(operators): write comprehensive JSDoc for reduce

staltz:
**Related issue:**

Fixes issue #1685.


staltz:
Merged

Dorus:
[reduce](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-reduce) still says the initial value is the first parameter, while it should be the second. Was this not fixed, or is http://reactivex.io/rxjs/ not yet updated?

kwonoj:
I think site is just not yet updated.

PR #1692, opened 2016-05-07
chore(commitizen): fix adapter path

### LINKED ISSUE #1691: `npm run commit` is not committing
Is it only me?

[Looks like latest version of `commitizen` expects the adapter path w/o `node_modules/` prefix.](https://github.com/commitizen/cz-cli#commitizen-for-project-maintainers)

```json
...
  "config": {
    "commitizen": {
      "path": "cz-conventional-changelog"
    }
  }
```

sarbbottam:
fixed #1691

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "config": {
     "commitizen": {
-      "path": "node_modules/cz-conventional-changelog"
+      "path": "cz-conventional-changelog"
     },
     "ghooks": {
       "commit-msg": "node ./node_modules/validate-commit-msg/index.js"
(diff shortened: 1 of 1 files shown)


kwonoj:
maybe can bump up version to `commitizen` also?

and you may able to use commit message `chore(package.json): ...` for general npm script related.

sarbbottam:
> maybe can bump up version to commitizen also?

Wondering if it should be done automatically via https://greenkeeper.io/?

kwonoj:
> Wondering if it should be done automatically via https://greenkeeper.io/?

: I considered it previously but some of package compatibility makes it should be managed by manually. commitizen is also one example, there's no way to automatically verify regression.

sarbbottam:
updated the commit message and bumped `commitizen` version.

kwonoj:
Merged with 97bcc15, thanks @sarbbottam

Issue #1697, opened 2016-05-10
type-check with typescript 1.9


alexeagle:
TypeScript 1.9 (@next) has been out for a while and will be the 2.0 beta in a couple weeks.
Newer TypeScript has many features which are useful in build tooling. Would be nice to be able to use these features while compiling rxjs.

$ npm install typescript@next
$ npm run build_es6

```
> @reactivex/rxjs@5.0.0-beta.7 compile_dist_es6 /Users/alexeagle/Projects/RxJS
> tsc                                          ./dist/es6/src/Rx.ts ./dist/es6/src/add/observable/of.ts                             -m es2015   --sourceMap --outDir ./dist/es6 --target ES6 -d --diagnostics --pretty --noImplicitAny --suppressImplicitAnyIndexErrors --moduleResolution node


130   const { callbackFunc, args, scheduler } = source;
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

dist/es6/src/observable/BoundCallbackObservable.ts(130,9): error TS2341: Property 'args' is private and only accessible within class 'BoundCallbackObservable<T>'.


130   const { callbackFunc, args, scheduler } = source;
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

dist/es6/src/observable/BoundCallbackObservable.ts(130,9): error TS2341: Property 'callbackFunc' is private and only accessible within class 'BoundCallbackObservable<T>'.


130   const { callbackFunc, args, scheduler } = source;
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

dist/es6/src/observable/BoundNodeCallbackObservable.ts(130,9): error TS2341: Property 'args' is private and only accessible within class 'BoundNodeCallbackObservable<T>'.


130   const { callbackFunc, args, scheduler } = source;
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

dist/es6/src/observable/BoundNodeCallbackObservable.ts(130,9): error TS2341: Property 'callbackFunc' is private and only accessible within class 'BoundNodeCallbackObservable<T>'.


102       unsubscribe = () => sourceObj.removeEventListener(eventName, <EventListener>handler);
                                        ~~~~~~~~~~~~~~~~~~~

[...]

PR #1700, opened 2016-05-11
chore(typings): fixed compiler errors when using @next version of typescript

david-driscoll:
**Description:**
This fixes two of the compiler errors that are issues with the `vNext` version of TypeScript (currently named 1.9, soon to be 2.0 I believe).

The third error is a compiler issue, so I have made no changes to `FromEventObservable`.

#1697

### DIFF
--- a/src/observable/BoundCallbackObservable.ts
+++ b/src/observable/BoundCallbackObservable.ts
@@ -80,7 +80,7 @@ export class BoundCallbackObservable<T> extends Observable<T> {
   constructor(private callbackFunc: Function,
               private selector: Function,
               private args: any[],
-              public scheduler: Scheduler) {
+              private scheduler: Scheduler) {
     super();
   }
 
@@ -100,7 +100,7 @@ export class BoundCallbackObservable<T> extends Observable<T> {
             const result = tryCatch(selector).apply(this, innerArgs);
             if (result === errorObject) {
               subject.error(errorObject.e);
-            } else {
+          } else {
               subject.next(result);
               subject.complete();
             }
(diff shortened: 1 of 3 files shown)


david-driscoll:
As per https://github.com/Microsoft/TypeScript/issues/7662 the error is a breaking change to TypeScript 2.0.  The fix is to block scope the variable.

src/observable/BoundCallbackObservable.ts
```diff
@@ -77,9 +77,9 @@ export class BoundCallbackObservable<T> extends Observable<T> {
     };
   }
 
-  constructor(private callbackFunc: Function,
-              private selector: Function,
-              private args: any[],
+  constructor(public callbackFunc: Function,
```
kwonoj: or maybe we can make `dispatch` static method in `boundcallbackObservable`? I would prefer to not expose these values as public.

rkirov:
Can we merge this change in? We would like to use TS 1.9 in angular2 as soon as possible.

Also, I see another error with @next, https://github.com/Microsoft/TypeScript/issues/8795
which will result in broken .d.ts files (but not broken compilation, which can be very confusing).

I recommend tacking on this change:

rewrite for `Subject<T>`:
```ts
  constructor(protected destination?: Observer<T>, protected source?: Observable<T>) {
    super();
  }
```
to
```ts
  constructor(destination?: Observer<T>, source?: Observable<T>) {
    super();
    this.destination = destination;
    this.source = source;
  }
```
so that it is clear that the arguments are optional but the field declarations are not. Or just make the field declarations in Observable optional too (but that would require all consumers to use 1.9).

david-driscoll:
Applied the feedback from @kwonoj and @rkirov that subject thing has bothered me in the past, I guess I never noted that the properties were optional... that's a strange bug for sure.

kwonoj:
Is my understanding correct above behavior (mark field optional) will be fixed by ts compiler eventually?

david-driscoll:
I'm sure it will, but doesn't seem like much harm to make them fields to get it working today.

kwonoj:
Those practices are widespread in our codebase not only in thise, so concerned even we apply immediate fix for this change there are still places have broken behavior. Maybe better to track compiler updates to resolve issue completely.

david-driscoll:
doesn't matter to me, I can flip it back. :smile:

blesh:
LGTM. :shipit:

PR #1701, opened 2016-05-11
BIG: Subject rewrite

blesh:
This is a rewrite of Subjects for RxJS 5 with the following effects:

- Better align with RxJS 4 behaviors around disposal and when ObjectUnsubscribedError is thrown
- Cleans up Subject implementation and separates out AnonymousSubject as a different class.
- This is prep work for a few fixes in ConnectableObservable.

This is a big set of changes. Breaking changes are listed in the commit.

Attn: @trxcllnt @mattpodwysocki @staltz @kwonoj @jayphelps

### DIFF
--- a/spec/operators/groupBy-spec.ts
+++ b/spec/operators/groupBy-spec.ts
@@ -70,7 +70,7 @@ describe('Observable.prototype.groupBy', () => {
       }, null, done);
   });
 
-  it('should group values with a duration selector', (done: MochaDone) => {
+  it('should group values with a duration selector', () => {
     const expectedGroups = [
       { key: 1, values: [1, 3] },
       { key: 0, values: [2, 4] },
@@ -78,19 +78,24 @@ describe('Observable.prototype.groupBy', () => {
       { key: 0, values: [6] }
     ];
 
+    const resultingGroups = [];
+
     Observable.of(1, 2, 3, 4, 5, 6)
       .groupBy(
         (x: number) => x % 2,
         (x: number) => x,
         (g: any) => g.skip(1))
       .subscribe((g: any) => {
-        const expectedGroup = expectedGroups.shift();
-        expect(g.key).to.equal(expectedGroup.key);
+        let group = { key: g.key, values: [] };
 
         g.subscribe((x: any) => {
-          expect(x).to.deep.equal(expectedGroup.values.shift());
+          group.values.push(x);
         });
-      }, null, done);
+
+        resultingGroups.push(group);
+      });
+
+      expect(resultingGroups).to.deep.equal(expectedGroups);
   });
 
   it('should handle an empty Observable', () => {
@@ -945,7 +950,7 @@ describe('Observable.prototype.groupBy', () => {
     };
     const e1 = hot('-1--2--^-a-b-c-d-e-f-g-h-i-j-k-l-|', values);
     const e1subs =        '^           !              ';
-    const expected =      '--v---w---x-#              ';
+    const expected =      '--v---w---x-(y#)              ';
     const v = cold(         'a-b---(d|)               ', values);
     const w = cold(             'c-----#              ', values);
     const x = cold(                 'e-#              ', values);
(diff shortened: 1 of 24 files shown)


blesh:
NOTE: Also fixes the behavior of `groupBy` to more closely match that of RxJS 4. (@staltz, I think you originally wrote those tests, can you verify?)

src/ReplaySubject.ts
```diff
@@ -2,55 +2,50 @@ import {Subject} from './Subject';
...
   private _windowTime: number;
 
   constructor(bufferSize: number = Number.POSITIVE_INFINITY,
               windowTime: number = Number.POSITIVE_INFINITY,
-              scheduler?: Scheduler) {
+              private scheduler: Scheduler = queue) {
     super();
-    this.scheduler = scheduler;
     this.bufferSize = bufferSize < 1 ? 1 : bufferSize;
     this._windowTime = windowTime < 1 ? 1 : windowTime;
```
paulpdaniels: I feel like these should have a consistent naming scheme. Either use the prefixed '_' or don't.

blesh: I agree... but that's been sitting there, and I was already knee-deep in yak hair.

src/Subject.ts
```diff
@@ -2,236 +2,136 @@ import {Operator} from './Operator';
...
+export class Subject<T> extends Observable<T> implements ISubscription {
+  observers: Observer<T>[] = [];
 
-  static create: Function = <T>(destination: Observer<T>, source: Observable<T>): Subject<T> => {
-    return new Subject<T>(destination, source);
-  };
+  isUnsubscribed = false;
 
-  constructor(protected destination?: Observer<T>, protected source?: Observable<T>) {
+  isStopped = false;
```
david-driscoll: Just an observation the lack of `private`/`protected`/`public` qualifiers makes all these fields essentially public to TypeScript consumers, some are useful (`hasError` for example) others may not be so useful.

blesh: Good note

blesh: For now, I think we should leave it and discuss later. They're all public in "plain JS" anyhow, it's more of a matter of how TS users will use it.

david-driscoll: I'm not super concerned with it, just was an observation.  The only thing it really affects is duck typing, if someone wanted to create a "clone" of subject, they'd have to implement these fields as well. (again not really super important)

src/ReplaySubject.ts
```diff
@@ -2,55 +2,50 @@ import {Subject} from './Subject';
...
 export class ReplaySubject<T> extends Subject<T> {
   private events: ReplayEvent<T>[] = [];
-  private scheduler: Scheduler;
   private bufferSize: number;
   private _windowTime: number;
 
   constructor(bufferSize: number = Number.POSITIVE_INFINITY,
               windowTime: number = Number.POSITIVE_INFINITY,
-              scheduler?: Scheduler) {
+              private scheduler: Scheduler = queue) {
```
trxcllnt: @blesh why are we defaulting to the queue scheduler here?

src/Subject.ts
```diff
@@ -2,236 +2,145 @@ import {Operator} from './Operator';
...
-    this.dispatching = false;
-
-    if (this.hasErrored) {
-      this._error(this.errorValue);
-    } else if (this.hasCompleted) {
-      this._complete();
+    this.hasError = true;
+    this.thrownError = err;
+    this.isStopped = true;
+    const { observers } = this;
```
trxcllnt: It seems we can get away without copying `observers` both here and in `complete`. After `hasError` or `isStopped` is true, all new applicants to `subscribe` con't make it into the observers list. Additionally, setting `this.observers` to null before notifying will save us the cost of each `SubjectSubscription` individually slicing their observer out of the observers list.

src/Subject.ts
```diff
@@ -2,236 +2,145 @@ import {Operator} from './Operator';
...
+  _subscribe(subscriber: Subscriber<T>): Subscription {
+    const { source } = this;
+    if (source) {
+      return this.source.subscribe(subscriber);
+    } else {
+      return Subscription.EMPTY;
     }
   }
-
-  [$$rxSubscriber]() {
```
trxcllnt: @blesh don't we still need this definition?

blesh: No, the only place it was used was changed because we have to wrap Subjects in a SafeSubscriber to prevent them from being `unsubscribed` and disabled (via throwing `ObjectUnsubscribedErrors`) in cases where they complete or error. This was done to match Rx4

src/util/toSubscriber.ts
```diff
@@ -1,19 +1,25 @@
...
 import {Subscriber} from '../Subscriber';
-import {$$rxSubscriber} from '../symbol/rxSubscriber';
+import {Subject} from '../Subject';
 
 export function toSubscriber<T>(
   nextOrObserver?: PartialObserver<T> | ((value: T) => void),
   error?: (error: any) => void,
   complete?: () => void): Subscriber<T> {
 
-  if (nextOrObserver && typeof nextOrObserver === 'object') {
```
trxcllnt: @blesh not a fan of the @@subscriber symbol anymore?

blesh: Actually, this change resulted in a circular reference issue I didn't notice until now... so I'm going to have to think about it.

src/Subject.ts
```diff
@@ -5,11 +5,26 @@ import {Subscriber} from './Subscriber';
...
+  }
+}
 
 /**
  * @class Subject<T>
  */
 export class Subject<T> extends Observable<T> implements ISubscription {
+
+  [$$rxSubscriber]() {
+    return new SubjectSubscriber(this);
```
trxcllnt: @blesh any reason this isn't just a regular `Subscriber`?

src/Subject.ts
```diff
@@ -2,236 +2,160 @@ import {Operator} from './Operator';
...
+  unsubscribe() {
+    this.isStopped = true;
+    this.isUnsubscribed = true;
+    this.observers = null;
   }
 
-  protected _next(value: T): void {
-    if (this.destination) {
-      this.destination.next(value);
+  _subscribe(subscriber: Subscriber<T>): Subscription {
```
saneyuki: why did you remove this (and derived-classes') `protected` qualifier?

blesh: Oops good catch

src/Subject.ts
blesh: I've readded it, FYI.

spec/operators/groupBy-spec.ts
```diff
@@ -1118,83 +1123,85 @@ describe('Observable.prototype.groupBy', () => {
...
-          hasSubscribed[group.key] = true;
-        }, subscriptionFrame - rxTestScheduler.frame);
-
-        return arr;
-      });
-
-    expectObservable(source).toBe(expected, expectedGroups);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
-  });
+  // HACK: I found this test hard to grok, and it was broken after the Subject refactor.
```
staltz: How exactly did it break?

spec/operators/multicast-spec.ts
```diff
@@ -548,29 +542,6 @@ describe('Observable.prototype.multicast', () => {
...
-          }, null, () => {
-            done();
-          });
-
-          expect(() => source.connect()).to.throw(Rx.ObjectUnsubscribedError);
-        });
-
-      source.connect();
-    });
-
```
staltz: Are we reverting the decision we did in https://github.com/ReactiveX/rxjs/issues/453 ?

spec/operators/publishReplay-spec.ts
```diff
@@ -193,12 +193,12 @@ describe('Observable.prototype.publishReplay', () => {
       expectObservable(subscriber1).toBe(expected1);
       expectObservable(subscriber2).toBe(expected2);
       expectObservable(subscriber3).toBe(expected3);
-      expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+      // expectSubscriptions(source.subscriptions).toBe(sourceSubs);
```
staltz: Why?

spec/operators/publishReplay-spec.ts
```diff
@@ -358,34 +358,4 @@ describe('Observable.prototype.publishReplay', () => {
...
-          done();
-        });
-
-        expect(() => source.connect()).to.throw(Rx.ObjectUnsubscribedError);
-      });
-
-    source.connect();
-
-    expect(results).to.deep.equal([4]);
-  });
```
staltz: :( ?

src/testing/HotObservable.ts
```diff
@@ -24,7 +24,7 @@ export class HotObservable<T> extends Subject<T> implements SubscriptionLoggable
     this.scheduler = scheduler;
   }
 
-  protected _subscribe(subscriber: Subscriber<any>): TeardownLogic {
+  _subscribe(subscriber: Subscriber<any>): Subscription {
```
saneyuki: nit: make `protected` :)

src/observable/dom/WebSocketSubject.ts
```diff
@@ -25,14 +25,15 @@ export interface WebSocketSubjectConfig {
...
-export class WebSocketSubject<T> extends Subject<T> {
+export class WebSocketSubject<T> extends AnonymousSubject<T> {
   url: string;
   protocol: string|Array<string>;
   socket: WebSocket;
   openObserver: NextObserver<Event>;
   closeObserver: NextObserver<CloseEvent>;
   closingObserver: NextObserver<void>;
   WebSocketCtor: { new(url: string, protocol?: string|Array<string>): WebSocket };
+  _output: Subject<T> = new Subject<T>();
```
saneyuki: nit: make `private` or `protected`

jadbox:
Any performance impact of the rewrite?

spec/operators/publishReplay-spec.ts
blesh: This was testing an inaccurate behavior. Rx4 didn't throw an error, it just didn't emit anything afterwards.

spec/operators/publishReplay-spec.ts
blesh: This is because now the subscriptions where playing over and over (since the Subject change). I wasn't sure how to test it.

```
  1) Observable.prototype.publishReplay with refCount() should NOT be retryable:
     AssertionError: 
Expected 
	{"subscribedFrame":0,"unsubscribedFrame":120}
	{"subscribedFrame":120,"unsubscribedFrame":120}
	{"subscribedFrame":120,"unsubscribedFrame":120}
	{"subscribedFrame":120,"unsubscribedFrame":120}
	{"subscribedFrame":120,"unsubscribedFrame":120}
	{"subscribedFrame":120,"unsubscribedFrame":120}
	{"subscribedFrame":120,"unsubscribedFrame":120}
	{"subscribedFrame":120,"unsubscribedFrame":120}
	{"subscribedFrame":120,"unsubscribedFrame":120}
	{"subscribedFrame":120,"unsubscribedFrame":120}
	
to deep equal 
	{"subscribedFrame":0,"unsubscribedFrame":120}
```

blesh: I mean, the *behavior* is okay. It seems like a problem with the test tooling.

blesh: What you see above is 3 observable trying to resubscribe 3 times and failing. 

Our test harness can't handle this scenario ATM. Not sure what to do there.

src/ReplaySubject.ts
blesh: Oops, I think that was because below it was using `queue` to get `now()` if there wasn't a scheduler.

blesh: Fixed.

blesh:
@jadbox overall the tests seem to be positive, in that this is a little faster. However, the tests I wrote were having some issues where V8 was optimizing `benchmark.js` itself WRT the Subject microperf tests. So I didn't commit those tests specifically. We'll need to move to more macro perf tests.

spec/operators/groupBy-spec.ts
blesh: FYI: I dropped the ball on this comment. @staltz, we'll need to circle back here, because I don't remember if there was any offline discussion about this test. I think we should capture what the purpose of this test was, and try to rewrite it to be a little more clear. This was broken by changes to how grouped subjects emitted, I believe, which was a cascading change from the Subject refactor to try to match Rx4 semantics.

Issue #1669, opened 2016-04-29
Broken reference counting when chaining refCount().first()


fterrazzoni:
**RxJS version:**
5.0.0-beta.6

**Code to reproduce:**
```js
subject = new Rx.Subject();
shared = subject.publishReplay(1).refCount();
shared.subscribe((x)=> console.log('Message:' + x));
subject.next('Hello');
shared.first().subscribe().unsubscribe();
subject.next('Goodbye');
```

**Expected behavior:**
```
Message:Hello
Message:Goodbye
```

**Actual behavior:**
```
Message:Hello
```

**Additional information:**

This code snippet behaves as expected in `RxJS 5 beta 3` and is broken since `RxJS 5 beta 4`. 

After some investigation, it looks like calling `shared.first().subscribe().unsubscribe()` on a `RefCountObservable` has the effect of decreasing `shared.refCount` by `1`, which is ultimately disconnecting the subject from the chain.

blesh:
@trxcllnt this is directly related to the work you're currently doing on `ConnectableObservable`.

trxcllnt:
resolved by #1705

Issue #1628, opened 2016-04-15
incorrect behavior of cache() 


unao:
**RxJS version:** 5.0.0-beta.6

**Code to reproduce:** 

```javascript
Rx.Observable
  .fromPromise(new Promise((res) => res(true)))
  .map(() => 'abc')
  .do(() => console.log('LOG_1'))
  .cache()
  .do(() => console.log('LOG_2'))
  .flatMap(() => Rx.Observable.timer(500)
    .flatMap(() => Rx.Observable.throw('ERR')))
  .retry(3)
  .subscribe(() => {})
```

**Expected behavior:** `LOG_1` should happen once and `LOG_2` four times:
```
LOG_1
LOG_2
LOG_2
LOG_2
LOG_2
```
Similarly to: 
```javascript
Rx.Observable.timer(1000)
  .map(() => 'abc')
  .do(() => console.log('LOG_1'))
  .cache()
  .do(() => console.log('LOG_2'))
  .flatMap(() => Rx.Observable.timer(500)
    .flatMap(() => Rx.Observable.throw('ERR')))
  .retry(3)
  .subscribe(() => {})
```

**Actual behavior:** `LOG_1` appears four times, moreover after `retry` the order of logs is reversed.
```
LOG_1
LOG_2
LOG_2
LOG_1
Uncaught ObjectUnsubscribedError
LOG_2
LOG_1
Uncaught ObjectUnsubscribedError
LOG_2
LOG_1
Uncaught ObjectUnsubscribedError
```

**Additional information:** Both code snippets generate: `Uncaught ObjectUnsubscribedError` after each retry. [demo](http://plnkr.co/edit/73s3KoDo8datezbLFc7W?p=preview)

unao:
By any chance did anyone take a look at it? Is it a bug or am I doing something wrong?

blesh:
I'll look into this very soon.

unao:
Just wonder if it may be _somehow_ connected to: #1466.

PR #1705, opened 2016-05-12
fix(ConnectableObservable): fix race conditions in ConnectableObservable and refCount.

trxcllnt:
**Description:**
Moves the refCount state to `ConnectableObservable`. Introduces a `Subscriber` between the source and the `Subject`, which can intercept `error` and `complete` calls before they go to the `Subject` and reset the state of the `ConnectableObservable`.

#1669 and possibly #1628


blesh:
LGTM 👍

PR #1706, opened 2016-05-12
fix(subscriptions): fixes bug that tracked subscriber subscriptions twice.

trxcllnt:
**Description:**
This fixes an issue where Subscribers were being added to their destination's subscription lists twice. I don't believe there was any functional regression, but performance theoretically suffered.

none that I'm aware of.


blesh:
👍 LGTM

blesh:
Merged with 29ff794. Thanks @trxcllnt

Issue #1698, opened 2016-05-10
windowCount + share + take(1) does not complete correctly.


Dorus:
```Js
Rx.Observable.interval(500)
.windowCount(1)
.do(v => console.log('before share'))
.share()
.do(v => console.log('after share'))
.take(1)
.mapTo(0)
.subscribe(console.log.bind(console), null, () => console.log('completed'));
```

prints:

```text
"before share"
"after share"
0
"completed"
"before share"
"before share"
"before share"
"before share"
"before share"
...
```
(endless repetition).

I'm not sure what's going on here, but after reading `completed` i would expect the source to stop also. Replace `.take(1)` with `.take(0)` or `.take(2)` and it does stop as expected.

trxcllnt:
resolved by #1705 and #1708

kwonoj:
Closing by PR's checked in.

PR #1708, opened 2016-05-12
fix(window): don't track internal window subjects as subscriptions.

trxcllnt:
**Description:**
This PR removes behavior in the window operators that tracked the window Subjects as Subscriptions. This could lead to race conditions where the Subscriber unsubscribes while the window operator is still active and attempting to send messages to the windows.

#1698

### DIFF
--- a/spec/operators/window-spec.ts
+++ b/spec/operators/window-spec.ts
@@ -1,4 +1,3 @@
-import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
 declare const {hot, cold, asDiagram, time, expectObservable, expectSubscriptions};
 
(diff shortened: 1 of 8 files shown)


spec/operators/window-spec.ts
```diff
@@ -201,28 +214,6 @@ describe('Observable.prototype.window', () => {
...
-      .do((w: any) => { window = w; });
-
-    expectObservable(result, unsub).toBe(expected, values);
-    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
-    rxTestScheduler.schedule(() => {
-      expect(() => {
-        window.subscribe();
-      }).to.throw(Rx.ObjectUnsubscribedError);
-    }, late);
-  });
```
blesh: I don't understand why this test went away. Can you elaborate? Thanks.

blesh:
It seems like #1705 resolves #1698. @trxcllnt, can you explain the relationship between #1698 and this PR?

spec/operators/window-spec.ts
trxcllnt: @blesh now that window's internal Subjects aren't being tracked as if they were Subscriptions, their `unsubscribe` method is never called when the subscriber disposes. This test attempts to subscribe to one of the window Subjects after the outer window subscription has been disposed, assuming the `WindowSubscriber` has called `unsubscribe` on the Subject. This PR invalidates that assumption.

trxcllnt:
@blesh #1705 resolves #1669 and partially fixed #1698, after which I discovered an unrelated race condition. The window operators would [continue to](https://github.com/ReactiveX/rxjs/blob/master/src/operator/windowCount.ts#L100) `next` or `complete` the internal window Subjects, even if a downstream subscriber unsubscribed in response to a `next` or `complete` event. Since the internal Subjects would be disposed at the time the downstream subscriber unsubscribed, attempting to message them again would throw `ObjectUnsubscribedErrors`. Minimal repro example:

```es6
Observable.range(0, 2)
  .windowCount(1).mergeAll()
  .take(1).subscribe();
```

spec/operators/window-spec.ts
blesh: Makes sense, thanks

blesh:
Looks good to me, @trxcllnt. :shipit:

PR #1679, opened 2016-05-03
feat(onErrorResumeNext): add onErrorResumeNext operator

kwonoj:
**Description:**
This PR implements `onErrorResumeNext` operators, including test coverages.

**Related issue (if exists):**

closes #1665

PR #1710, opened 2016-05-13
docs(MIGRATION): onErrorResumeNext was added to v5

jayphelps:
Was added in #1679 but MIGRATION guide still listed it as excluded from v5.

Cc/ @kwonoj


kwonoj:
Ah yes, I missed those :eyes: . Merged with cd14899, thanks @jayphelps :)

PR #1714, opened 2016-05-18
missing bracket

dreambo8563:
**Description:**


kwonoj:
Merged with 9b4c5d8, thanks @dreambo8563 . I've amended commit message bit while check in changes to conform commit message guidelines (https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md#commit) which is being used to automate changelog / history.

PR #1723, opened 2016-05-22
fix(combineLatest): rxjs/observable/combineLatest is now properly exp…

### LINKED ISSUE #1722: rxjs/observable/combineLatest missing.
Just what it sounds like. `import { combineLatest } from 'rxjs/observable/combineLatest';` doesn't do anything. :\

blesh:
…orted

fixes #1722


kwonoj:
:+1:

PR #1726, opened 2016-05-23
docs(manual): simplify language, improve flow

alex-wilmer:
**Description:**

Motivation:

First of all I would like to say bravo on the docs. It's a lot more user friendly than what I recall from v4. The language is more accessible and api generally easier to navigate. This pull request attempts to smooth some rough edges in an attempt to keep the barrier of entry low, so that people like me--very much on the beginner's end of Rx--don't feel overwhelmed or trip on some confusing sentences. 

Notes: 

 - Moved placement of adjectives / adverbs to be more typically "English"

  ex: `Multiply by 10 this value` -> `Multiply this value by 10`

 - Instances of `just`, `simply`, `obviously` don't usually add quality (they can, but not usually), and often give a condescending tone. It's safer to remove them I believe.

 - "Four core Observable concerns": `subscribe` was out of place here.. either they become all nouns or all verbs (subscription vs subscribe)

 - "every one second": awkward, same as `every second`

 - `next`, `error` and `complete` are notifications, not values. Improved clarity here

 - My change in the `multicast` section was the most drastic. I think there's a chance I over simplified or missed something important in my rewrite, however I think being more direct / explicit, less verbose and referring to fewer CS language (unless totally necessary) is a win here. For instance `subclass` is mentioned only once in the entire manual.. better to just leave that out.

Hope this is a welcome change. I'm enjoying my journey with RxJS so far and will happily put more effort into contributing here.

Cheers,
A

### DIFF
--- a/doc/observable.md
+++ b/doc/observable.md
@@ -69,9 +69,9 @@ ES2015 introduced [generator functions and iterators](https://developer.mozilla.
 | | Producer | Consumer |
 | --- | --- | --- |
 | **Pull** | **Passive:** produces data when requested. | **Active:** decides when data is requested. |
-| **Push** | **Active:** produces data at its own pace. | **Passive:** just reacts to data sent to it. |
+| **Push** | **Active:** produces data at its own pace. | **Passive:** reacts to received data. |
 
-**What is Push?** In Push systems, the Producer determines when to send data to the Consumer The Consumer itself is unaware of when the data will be received.
+**What is Push?** In Push systems, the Producer determines when to send data to the Consumer, who is unaware of when that data will be received.
 
 Promises are the most common type of Push system in JavaScript today. A Promise (the Producer) delivers a resolved value to registered callbacks (the Consumers), but unlike functions, it is the Promise which is in charge of determining precisely when that value is "pushed" to the callbacks.
 
@@ -149,7 +149,7 @@ console.log(foo.call());
 console.log('after');
 ```
 
-You will obviously see the output:
+You will see the output:
 
 ```none
 "before"
@@ -259,19 +259,19 @@ Conclusion:
 
 ## Anatomy of an Observable
 
-Observables are **created** using `Rx.Observable.create` or a creation operator, are **subscribed** to with an Observer, **execute** to deliver `next` / `error` / `complete` notifications to the Observer, and their execution may be **disposed**. These four aspects are all encoded in an Observable instance, but some of these aspects are related to other types, like Observer an Subscription.
+Observables are **created** using `Rx.Observable.create` or a creation operator, are **subscribed** to with an Observer, **execute** to deliver `next` / `error` / `complete` notifications to the Observer, and their execution may be **disposed**. These four aspects are all encoded in an Observable instance, but some of these aspects are related to other types, like Observer and Subscription.
 
 Four core Observable concerns:
-- **Creation**
+- **Create**
 - **Subscribe**
-- **Execution**
-- **Disposal**
+- **Execute**
+- **Dispose**
 
 ### Creating Observables
 
 `Rx.Observable.create` is an alias for the `Observable` constructor, and it takes one argument: the `subscribe` function.
 
-The following example creates an Observable to emit the string `'hi'` every one second to an Observer.
+The following example creates an Observable to emit the string `'hi'` every second to an Observer.
 
 ```js
 var observable = Rx.Observable.create(function subscribe(observer) {
@@ -308,11 +308,11 @@ A `subscribe` call is simply a way to start an "Observable execution" and delive
 
 The code inside `Observable.create(function subscribe(observer) {...})` represents an "Observable execution", a lazy computation that only happens for each Observer that subscribes. The execution produces multiple values over time, either synchronously or asynchronously.
 
-There are three types of values an Observable Execution can deliver:
+There are three types of events, or notifications, that an Observable can execute:
 
-- "Next" notification: an actual value such as a Number, a String, an Object, etc.
-- "Error" notification: a JavaScript Error or exception.
-- "Complete" notification: a valueless event.
+- "Next" notification: sends a value such as a Number, a String, an Object, etc.
+- "Error" notification: sends a JavaScript Error or exception.
+- "Complete" notification: does not send a value.
 
 Next notifications are the most important and most common type: they represent actual data being delivered to an Observer. Error and Complete notifications may happen only once during the Observable Execution, and there can only be either one of them.
 
@@ -373,7 +373,7 @@ When `observable.subscribe` is called, the Observer gets attached to the newly c
 var subscription = observable.subscribe(x => console.log(x));
 ```
 
-The Subscription represents the ongoing execution, and has a minimal API, which primarily simply allows you to cancel that execution. Read more about the [`Subscription` type here](./overview.html#subscription). With `subscription.unsubscribe()` you can cancel the ongoing execution:
+The Subscription represents the ongoing execution, and has a minimal API which primarily allows you to cancel that execution. Read more about the [`Subscription` type here](./overview.html#subscription). With `subscription.unsubscribe()` you can cancel the ongoing execution:
 
 ```js
 var observable = Rx.Observable.from([10, 20, 30]);
@@ -382,7 +382,7 @@ var subscription = observable.subscribe(x => console.log(x));
 subscription.unsubscribe();
 ```
 
-<span class="informal">When you subscribe, you get back a Subscription, which represents the ongoing execution. Just call `unsubscribe()` to cancel the execution.</span>
+<span class="informal">When you subscribe, you get back a Subscription, which represents the ongoing execution. Calling `unsubscribe()` will cancel the execution.</span>
 
 Each Observable must define how to dispose resources of that execution when we create the Observable using `create()`. You can do that by returning a custom `unsubscribe` function from within `function subscribe()`.
 
@@ -394,7 +394,7 @@ var observable = Rx.Observable.create(function subscribe(observer) {
   var intervalID = setInterval(() => {
     observer.next('hi');
   }, 1000);
-  
+
   // Provide a way of canceling and disposing the interval resource
   return function unsubscribe() {
     clearInterval(intervalID);
@@ -402,14 +402,14 @@ var observable = Rx.Observable.create(function subscribe(observer) {
 });
 ```
 
-Just like `observable.subscribe` resembles `Observable.create(function subscribe() {...})`, the `unsubscribe` we return from `subscribe` is conceptually equal to `subscription.unsubscribe`. In fact, if we remove the ReactiveX types surrounding these concepts, we get some JavaScript that is clearly simple:
+Just like `observable.subscribe` resembles `Observable.create(function subscribe() {...})`, the `unsubscribe` we return from `subscribe` is conceptually equal to `subscription.unsubscribe`. In fact, if we remove the ReactiveX types surrounding these concepts, we're left with rather straightforward JavaScript.
 
 ```js
 function subscribe(observer) {
   var intervalID = setInterval(() => {
     observer.next('hi');
   }, 1000);
-  
+
   return function unsubscribe() {
     clearInterval(intervalID);
   };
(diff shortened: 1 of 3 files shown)


doc/observable.md
```diff
@@ -69,9 +69,9 @@ ES2015 introduced [generator functions and iterators](https://developer.mozilla.
 | | Producer | Consumer |
 | --- | --- | --- |
 | **Pull** | **Passive:** produces data when requested. | **Active:** decides when data is requested. |
-| **Push** | **Active:** produces data at its own pace. | **Passive:** just reacts to data sent to it. |
+| **Push** | **Active:** produces data at its own pace. | **Passive:** reacts to received data. |
 
-**What is Push?** In Push systems, the Producer determines when to send data to the Consumer The Consumer itself is unaware of when the data will be received.
```
alex-wilmer: Missing punctuation.. could add a `.` if it makes more sense.

doc/observable.md
```diff
@@ -373,7 +373,7 @@ When `observable.subscribe` is called, the Observer gets attached to the newly c
 var subscription = observable.subscribe(x => console.log(x));
 ```
 
-The Subscription represents the ongoing execution, and has a minimal API, which primarily simply allows you to cancel that execution. Read more about the [`Subscription` type here](./overview.html#subscription). With `subscription.unsubscribe()` you can cancel the ongoing execution:
+The Subscription represents the ongoing execution, and has a minimal API which primarily allows you to cancel that execution. Read more about the [`Subscription` type here](./overview.html#subscription). With `subscription.unsubscribe()` you can cancel the ongoing execution:
```
alex-wilmer: I think "primarily" can be removed here as well. Primarily makes me think, "oh what else does it do?" only to find out it's all about cancelling really

doc/observable.md
```diff
@@ -69,9 +69,9 @@ ES2015 introduced [generator functions and iterators](https://developer.mozilla.
 | | Producer | Consumer |
 | --- | --- | --- |
 | **Pull** | **Passive:** produces data when requested. | **Active:** decides when data is requested. |
-| **Push** | **Active:** produces data at its own pace. | **Passive:** just reacts to data sent to it. |
+| **Push** | **Active:** produces data at its own pace. | **Passive:** reacts to received data. |
 
-**What is Push?** In Push systems, the Producer determines when to send data to the Consumer The Consumer itself is unaware of when the data will be received.
+**What is Push?** In Push systems, the Producer determines when to send data to the Consumer, who is unaware of when that data will be received.
```
blesh: I think I like:

> In Push systems, the Producer determines when to send data to the Consumer. The Consumer is unaware of when it will receive that data.

doc/observable.md
```diff
@@ -259,19 +259,19 @@ Conclusion:
...
+Observables are **created** using `Rx.Observable.create` or a creation operator, are **subscribed** to with an Observer, **execute** to deliver `next` / `error` / `complete` notifications to the Observer, and their execution may be **disposed**. These four aspects are all encoded in an Observable instance, but some of these aspects are related to other types, like Observer and Subscription.
 
 Four core Observable concerns:
-- **Creation**
+- **Create**
 - **Subscribe**
-- **Execution**
-- **Disposal**
+- **Execute**
+- **Dispose**
```
blesh: I disagree with this change. "Create" isn't a concern. I don't think people would say "I'm concerned with Create". Whereas, "I'm concerned with Creation" makes more sense.

doc/observable.md
blesh: You can technically add other subscriptions to a subscription.

blesh: But I agree, "primarily" feels awkward here, we can probably remove it.

blesh:
@alex-wilmer looks great! I've left a couple of comments for correction. Please review and I'm happy to merge this one.

doc/observable.md
alex-wilmer: Agreed, so do you think it should be left as is, or should `subcribe` be changed to `subscription`?

staltz: It's not that simple. The `Subscription` represents the "Execution" of the Observable, but the `subscribe` is an important event where everything "starts" and gets set up. So it's not a matter of changing verbs to nouns.

doc/observable.md
```diff
@@ -308,11 +308,11 @@ A `subscribe` call is simply a way to start an "Observable execution" and delive
...
 
-There are three types of values an Observable Execution can deliver:
+There are three types of events, or notifications, that an Observable can execute:
 
-- "Next" notification: an actual value such as a Number, a String, an Object, etc.
-- "Error" notification: a JavaScript Error or exception.
-- "Complete" notification: a valueless event.
+- "Next" notification: sends a value such as a Number, a String, an Object, etc.
+- "Error" notification: sends a JavaScript Error or exception.
+- "Complete" notification: does not send a value.
```
staltz: The notification doesn't send anything itself. It just "contains" values. So I'd say "Next notification: has a value such as a Number, etc".

doc/observable.md
```diff
@@ -382,7 +382,7 @@ var subscription = observable.subscribe(x => console.log(x));
 subscription.unsubscribe();
 ```
 
-<span class="informal">When you subscribe, you get back a Subscription, which represents the ongoing execution. Just call `unsubscribe()` to cancel the execution.</span>
+<span class="informal">When you subscribe, you get back a Subscription, which represents the ongoing execution. Calling `unsubscribe()` will cancel the execution.</span>
```
staltz: I think the "just" wasn't condescending here and it makes the sentence sound more conversational, which is the purpose of these spans. I'd rather keep the "just".

doc/subject.md
```diff
@@ -69,7 +69,7 @@ There are also a few specializations of the `Subject` type: `BehaviorSubject`, `
 
 ## Multicasted Observables
 
-Whenever we refer to "a multicasted Observable", that means an Observable execution that passes through a Subject. Otherwise, a "plain (unicast) Observable" behaves like an *invokable collection of future values*, with no sharing assumed.
+A "multicasted Observable" passes notifications through a Subject which may have many subscribers, whereas a plain "unicast Observable" only sends notifications to a single Observer.
```
staltz: Not a helpful change, I find it hard to understand this sentence.

doc/subject.md
```diff
@@ -146,7 +146,7 @@ setTimeout(() => {
 }, 2000);
 ```
 
-If we wish to avoid explicit calls to `connect()`, we use the method `refCount()` on the ConnectableObservable. It does reference counting: keeps track of how many subscribers are registered on the ConnectableObservable. `refCount()` calls `connect()` when that number goes from `0` to `1`, and keeps the subscription for the shared execution. When the number of subscribers goes from `1` to `0`, it finally unsubscribes the subscription to the shared execution. 
+If we wish to avoid explicit calls to `connect()`, we can use ConnectableObservable's `refCount()` method, which returns an Observable that keeps track of how many subscribers it has. When the number of subscribers increases from `0` to `1`, it will call `connect()` for us, which starts the shared execution. Only when the number of subscribers decreases from `1` to `0` will it be fully unsubscribed, stopping further execution.
```
staltz: I liked the "When the number of subscribers increases" change and onwards, but I'd rather keep the previous part because it clarifies that `refCount` means "reference counting", which may not be obvious to all.

staltz:
Thanks, this fixes a lot of weird idioms I use without noticing myself. I put some comments on the more critical bits, but I'm also welcome to this PR which is small enough to be decently reviewed and merged.

doc/observable.md
alex-wilmer: I agree this wasn't one of those places where I found it condescending. I found it to be an answer to the question "Hey how do I unsubscribe?" even though I wasn't asking. I feel way less strongly about this than my other changes. Will put it back how it was.

doc/subject.md
alex-wilmer: Fair enough, however `behaves like an *invokable collection of future values*, with no sharing assumed` is definitely confusing and out of place here. "Plain" observables have already had a lot literature up to this point, and this sentence doesn't help to make the distinction between multicast which is also a collection of future values.

doc/observable.md
alex-wilmer: I completely get that--but only after getting this far in with RxJS. Someone fresh reading the manual isn't so aware of which terms are Types and which are aliases, like Subscriber, which is not a true type.

For the same reason @blesh doesn't like `I am concerned with create`, `I am concerned with subscribe` is not a proper sentence. Maybe this section regarding "concerns" should be contain fuller sentences:

The four primary concerns are:

**Creating** Observables
What happens when an observer **subscribes** to an Observable.
**Executing** notifications.
What to do when an observer **disposes** a (or **unsubscribes** from) a subscription.

alex-wilmer:
Thanks guys, appreciate the feedback very much. Apologies for the delay on updating this PR. I'm hoping to have a bit of free time early next work week.

alex-wilmer:
Sorry for the delay! PR updated as best as I could without losing the intention or ignoring your comments.  Will rebase / squash when approved, if that's preferred. Thanks

doc/observable.md
```diff
@@ -308,11 +308,11 @@ A `subscribe` call is simply a way to start an "Observable execution" and delive
 
 The code inside `Observable.create(function subscribe(observer) {...})` represents an "Observable execution", a lazy computation that only happens for each Observer that subscribes. The execution produces multiple values over time, either synchronously or asynchronously.
 
-There are three types of values an Observable Execution can deliver:
+There are three types of events, or notifications, that an Observable can execute:
```
staltz: We don't use this language "executing notifications" anywhere, and it brings different connotations, it may be hinting something that doesn't actually happen. Observables "deliver" values/notifications/events, they don't "execute" them. I'd recommend reverting.

doc/observable.md
```diff
@@ -259,19 +259,19 @@ Conclusion:
...
 
-Four core Observable concerns:
-- **Creation**
-- **Subscribe**
-- **Execution**
-- **Disposal**
+Core Observable concerns:
+- **Creating** Observables
+- **Subscribing** to Observables
+- **Executing** Notifications
```
staltz: "**Executing** the Observable"

alex-wilmer:
Thanks @staltz, cleaned up. I have one question though.. Isn't the act of an Observer subscribing to an Observable trigger the Observable's execution? Aren't those two moments technically the same point in time, unless we're talking about subscribing to a Subject where an Observable execution may have already begun? 

I ask because as someone who maybe reading about Observables for the first time may think that "Executing the Observable" is something entirely different from subscribing to one. Am I wrong aboutthis?

alex-wilmer:
I'm fairly certain I personally understand Observable execution, kicking off that execution and stopping it. Being the first / only subscriber which kicks it off... receiving "deliveries" from Observables during it's execution. This stuff becomes a lot easier to grok as you use the library. I think it's important to let go of these particulars in the manual though--unless they are expertly communicated, possibly with diagrams. 

At the moment I think the manual suffers from jargon overloading and the concept of 'execution' is not well defined. 

Thoughts?

blesh:
It's probably a good idea to start a glossary of terms and use them consistently throughout the docs.

@staltz, given that this an iterative process, do you feel this PR is GTG? It LGTM.

doc/observable.md
LongLiveCHIEF: Idle thought here... what If we used "yields" instead of "sends"?

staltz:
Yes, it's quite ready to merge. Thank you for all the effort @alex-wilmer.

About the raised questions:

> Isn't the act of an Observer subscribing to an Observable trigger the Observable's execution? Aren't those two moments technically the same point in time, unless we're talking about subscribing to a Subject where an Observable execution may have already begun? 

It looks like you answered that yourself already. :)
If you `subscribe` to a Subject, you are not triggering any new execution at all. Conflating "subscribing" and "executing" would lead to misunderstandings.
The `subscribe` mostly coincides with "the beginning of the Observable execution", but not necessarily, and it's important that this distinction exists. Also when I say `subscribe` I don't mean "the Subscription". The Subscription mostly represents the execution of the Observable, but not necessarily. It's tricky.

In layman terms,
`subscribe` = "give me values"
Subscription = "a thing that lets me tell you to stop giving me values"
Execution = "the Observable/Subject is running, and delivering values"

I tried to write these associations in the docs, e.g.

> Subscribing to an Observable is analogous to calling a Function.

Issue #1718, opened 2016-05-20
cache behaving unexpected


Dorus:
```
const source = Rx.Observable.merge(
      Observable.interval(100),
      Observable.of(42).do(() => console.log("  effect A"))
  );

const source2 = Observable.of(52).do(() => console.log("  effect B"));
const o1 = source.cache(1);
const o2 = source2.cache(1);
const ob = (name) => ({next: x => console.log(name + x), complete: () => console.log(name + "c")})
o1.take(3).subscribe(ob("a1 "));
o2.take(1).subscribe(ob("a2 "));
setTimeout(() => {
  o1.take(3).subscribe(ob("b1 "));
  o2.take(1).subscribe(ob("b2 "));
}, 500);

// setTimeout(() => {
//   o1.take(3).subscribe(ob("c1 "));
//   o2.take(1).subscribe(ob("c2 "));
// }, 1000);

// setTimeout(() => {
//   o1.take(3).subscribe(ob("d1 "));
//   o2.take(1).subscribe(ob("d2 "));
// }, 1500);
```

result in:

```text
"  effect A"
"a1 42"
"  effect B"
"a2 52"
"a2 c"
"a1 0"
"a1 1"
"a1 c"
"b1 1"
"  effect A"
"b1 42"
"b2 52"
"b2 c"
"b1 0"
"b1 c"
```

What causes `effect B` to only play once, but `effect A` to play twice? Playing around with the code randomly seems to cause side-effects to appear and disappear on the source.

Because `cache`  works like `publishReplay().refCount()`, i would expect `effect B` to also play again after `a2 c` (a2 complete), but b2 does not seem to query the source again.

On a possibly unrelated note, `cache` combines poorly with cold observables (or startWith) because it replays later values and then subscribes again. But this behavior is rather random because of the bug above.

Dorus:
Also, an even worse example is

```JS
o1.take(2).subscribe(ob("a1 "));

setTimeout(() => {
  o1.take(2).subscribe(ob("b1 "));
}, 500);


setTimeout(() => {
  o1.take(2).subscribe(ob("c1 "));
}, 1000);

setTimeout(() => {
  o1.take(2).subscribe(ob("d1 "));
}, 1500);
```
where `effect A` plays 3 times, for `a1`, `b1` and `d1`, but not for `c1`.

trxcllnt:
[...]

PR #1727, opened 2016-05-23
fix(ConnectableObservable): fix ConnectableObservable connection handling issue

trxcllnt:
**Description:**
Fixes an issue that caused ConnectableObservable to assign itself a connection that should be disposed when working with synchronous Observables. This issue caused a race condition in synchronous refCounted Observables if the subscribers dispose before the published source completes, causing the ConnectableObservable not to re-subscribe to the original source when it should have.

#1718

### DIFF
--- a/src/observable/ConnectableObservable.ts
+++ b/src/observable/ConnectableObservable.ts
@@ -33,7 +33,9 @@ export class ConnectableObservable<T> extends Observable<T> {
   connect(): Subscription {
     let connection = this._connection;
     if (!connection) {
-      connection = this.source.subscribe(new ConnectableSubscriber(this.getSubject(), this));
+      connection = this._connection = new Subscription();
+      connection.add(this.source
+        .subscribe(new ConnectableSubscriber(this.getSubject(), this)));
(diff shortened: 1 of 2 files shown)


src/observable/ConnectableObservable.ts
```diff
@@ -123,9 +129,11 @@ class RefCountSubscriber<T> extends Subscriber<T> {
     }
 
     const { connection } = this;
-    if (connection) {
-      this.connection = null;
-      connection.unsubscribe();
+    const sharedConnection = (<any> connectable)._connection;
+    this.connection = null;
+
+    if (sharedConnection && (!connection || sharedConnection === connection)) {
```
blesh: Can you add a comment explaining this line? I don't want to grok it again later. haha.

blesh:
Other than the one comment I made, LGTM.

src/observable/ConnectableObservable.ts
trxcllnt: Sure can.

blesh:
:shipit:

PR #1738, opened 2016-05-31
Test for issues 1466

fs123:
Hello

This PR contains a test for the issue #1466 - WebSocketSubject cannot be retried. It looks, that the issue does no longer exists since beta 8. Because I want to fix it and I started my work with a test, I'm able to submit the test with this PR ;-)

On branch 5.0.0-beta.6 and 5.0.0-beta.7 the test fails with an "ObjectUnsubscribedError".

If something needs to be changed, let me know.

Finally I have to say thanks for the great work - its damn impressive!


blesh:
Cool! Thanks @fs123!

fs123:
It was my pleasure.

PR #1740, opened 2016-06-02
chore(typings): enable 'noImplicitReturns' compiler option

saneyuki:
I open this pull request to make our codebase more solidly.


saneyuki:
At now, we don't have to do any change to our codebase to enable this compiler option 🎉

kwonoj:
Cool thing. I'll check this in this changes around late today after try out couple of things.

Issue #1248, opened 2016-01-25
Typing for "Observable-y" returns


blesh:
### A Problem Of Return Typing

I'd like to examine functions such as `mergeMap` and `switchMap`... they take an argument that is like `(value: T, index: number) => Observable<R>` and we need something different, because `Observable<R>` could be a `Promise<R>` or `R[]` or `Iterable<R>` or `Iterator<R>` or `observable<R>` (which is a different type altogether, honestly).
### IObservable-able

For one thing, there are two weird interfaces in there...

``` TypeScript
interface IObservableable {
  [Symbol.observable]() : IObservable<any>
}

interface IObservable<T> extends IObservableable {
  subscribe(observer?: Observer<T>) => Subscription
}
```

That is a mirror, basically of `Iterable` and `Iterator` in ES6.

Thoughts?

/cc @david-driscoll for expertise and @zenparsing for vernacular.

david-driscoll:
So merged fairly recently are these types, though they are not used extensively yet.
https://github.com/ReactiveX/RxJS/blob/master/src/Observable.ts#L40-L42, names can be changed, no point fighting over, because naming is hard™. They're derived from similar types in RxJS4 ( https://github.com/Reactive-Extensions/RxJS/blob/master/ts/core/es5.ts#L3-L8 ).  I plan to progressively send PR's updating types as they move along.

Basic premise is that, `ObservableInput<T>` represents a value that can enter the system, with most operators, they get directed and enter through [`subscribeToResult`](https://github.com/ReactiveX/RxJS/blob/master/src/util/subscribeToResult.ts).  This can either be through a selector method (`switchMap`), or directly as an argument (`combineLatest`).  The results of all these methods would all be `Observable<T>`, we just need a type that "describes" the inputs, hence why I named it `ObservableInput<T>`.

Today `subscribeToResult` clearly supports:
* `Observable<T>`
* `Array<T>`
* `Promise<T>`
* `Iterator<T>` (symbol based)
  * Which means the type in `ObservableInput<T>` should include be `IterableIterator<T>` not 
`Iterator<T>` (sorry! :imp:)
* `Iterable<T>` (symbol based)
  * Which would mean `ObservableInpuit<T>` needs to be updated with `Iterable<T>` as well.
* `obsevable<T>`(~ish) (symbol based)

[...]

PR #1741, opened 2016-06-02
feat(Observable): initial IObservable interface

vankeer:
This PR is not meant to be merged, but to gauge for your opinion on this.
If approved, I can make the changes for the other operators and methods as well.

**Description:**

An interface for Observables.

This fixes a problem we were having when npm linking a library exposing methods that return Observables.

For example, when you `npm link ng2-example-library` ([ng2-example-library](https://github.com/vankeer/ng2-example-library)) in [ng2-example-consumer](https://github.com/vankeer/ng2-example-consumer), you get this:

> src\app\repository\repository.service.ts(14,10): error TS2322: Type 'Observable<any[]>' is not assignable to type 'Observable<any[]>'.
>   Property 'source' is protected but type 'Observable<T>' is not a class derived from 'Observable<T>'.

**Impact:**

Quite a lot of operator signatures would have to be adapted to accommodate this change, as well as changes by the user would be required (Observable > IObservable). We could avoid the latter by renaming the interface to Observable and the class to something else.

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -17,13 +17,19 @@ export interface Subscribable<T> {
 export type SubscribableOrPromise<T> = Subscribable<T> | Promise<T>;
 export type ObservableInput<T> = SubscribableOrPromise<T> | ArrayLike<T>;
 
+export interface IObservable<T> extends Subscribable<T> {
+  subscribe(observerOrNext?: PartialObserver<T> | ((value: T) => void),
+            error?: (error: any) => void,
+            complete?: () => void): Subscription;
+}
+
 /**
  * A representation of any set of values over any amount of time. This the most basic building block
  * of RxJS.
  *
  * @class Observable<T>
  */
-export class Observable<T> implements Subscribable<T> {
+export class Observable<T> implements IObservable<T> {
 
   public _isScalar: boolean = false;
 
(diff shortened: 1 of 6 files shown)


kwonoj:
I think this is related with issue https://github.com/ReactiveX/rxjs/issues/1248, currently bit on hold to clarify / ensure some of usecases. /cc @david-driscoll as well.

david-driscoll:
What version of the TypeScript compiler are you on?  I believe a lot of these issues were fixed already.

https://github.com/Microsoft/TypeScript/issues/6731 and https://github.com/Microsoft/TypeScript/pull/6786

If something still remains, it's likely something that can be fixed without a huge sweeping change here (hopefully anyhow).

vankeer:
I am using version 1.8.10 so that should be fine. I am still looking for solutions other than changing the return type to an interface.

kwonoj:
I'd like to suggest create issue first to illustrate issues with latest master to be discussed. While introducing interface to observable is on discussion, it has somewhat different goals this PR tries to achieve and possibly can take longer time. If current problem / issue can be fixed in other way, it could be introduced separately in current shape of codebases.

vankeer:
Thanks @kwonoj and @david-driscoll for your responses.
Given that this is indeed quite a huge refactoring, I have opened issue #1744 and will close this PR.

Issue #1729, opened 2016-05-24
Ajax: wrong Content-Type when uploading FormData


jachenry:
**RxJS version:**
5.0.0-beta.8

**Code to reproduce:**
```
Observable.ajax({
  url: ...,
  body: formData,
})
```

**Expected behavior:**
browser should be responsible for appending `Content-Type` of `multipart/form-data; charset=utf-8; boundary=...`

**Actual behavior:**
`Content-Type` is explicitly set in AjaxObservable

```
    if (!('Content-Type' in headers)) {
      headers['Content-Type'] = 'application/x-www-form-urlencoded; charset=UTF-8';
    }
```

kwonoj:
Closing this issue by #1746 is checked in.

PR #1746, opened 2016-06-07
fix(AjaxObservable): ignore content-type for formdata

jachenry:
**Description:**

allow undefined content-type header if body is formdata

references GitHub issue https://github.com/ReactiveX/rxjs/issues/1729


kwonoj:
change looks ok to me.

PR #1747, opened 2016-06-07
refactor(SubjectSubscription): remove 'any' type from SubjectSubscription

saneyuki:
This clarify arguments' types of `SubjectSubscription`.

### DIFF
--- a/src/SubjectSubscription.ts
+++ b/src/SubjectSubscription.ts
@@ -7,10 +7,10 @@ import {Subscription} from './Subscription';
  * @ignore
  * @extends {Ignored}
  */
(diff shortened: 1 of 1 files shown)


david-driscoll:
Do we expose SubjectSubscription as part of the public api? (I know it's exported, but the public api only returns subscription https://github.com/ReactiveX/rxjs/blob/39fb278b3a0163652b782b381b0d1a2b5708d035/src/Subject.ts#L102-L115)

I know they may parts internally where we consume it.  This change isn't bad, I just don't want to set a precedent that `Subscription` should be typed.

PR #1748, opened 2016-06-07
refactor(MergeMap): clarify any types which are used in MergeMapSubscriber

saneyuki:
This just intends to clarify & make strictly types.


david-driscoll:
:+1:

Issue #1745, opened 2016-06-06
WebSocketSubject not applying operators


fingermark:
**RxJS version:**
5.0.0-beta.8

**Code to reproduce:**

http://jsbin.com/kidawe/edit?js,console

```
const url = 'ws://ws.websocketstest.com:80/service';

Rx.Observable
  .webSocket({
    url: url,
    resultSelector: (e) => e.data
  })
  .map((x) => 'prefix-' + x)
  .map(x => x)
  .subscribe((x) => console.log(x));
```

**Expected behavior:**

`prefix-connected,`

**Actual behavior:**

`connected,`

**Additional information:**

I ran into an issue where I was trying to map on Observable.webSocket and noticed the `map`s weren't getting invoked.  Others have confirmed the bug.  See more here:

https://github.com/redux-observable/redux-observable/issues/42

jayphelps:
Just to clarify, if you only apply a single operator, it works. `Observable.webSocket().map(x => x)` but if you apply more than one, none of them get applied. `Observable.webSocket().map(x => 'prefix-' + x).map(x => x)`

blesh:
Ouch... nasty. Def a result of #1701 ... :\ I did that.

PR #1752, opened 2016-06-07
fix(WebSocketSubject): WebSocketSubject will now chain operators prop…

blesh:
Removed the custom `lift` method from WebSocketSubject. This is the quick and dirty fix for #1745. I think this class needs to be refactored again so it no longer overwrites it's own `destination`.

### DIFF
--- a/spec/observables/dom/webSocket-spec.ts
+++ b/spec/observables/dom/webSocket-spec.ts
@@ -52,6 +52,29 @@ describe('Observable.webSocket', () => {
     subject.unsubscribe();
   });
 
+  it('should allow the user to chain operators', () => {
+    let messageReceived = false;
+    const subject = Observable.webSocket('ws://mysocket');
+
+    subject
+      .map(x => x + '?')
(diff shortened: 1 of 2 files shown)


rgbkrk:
Awesome, thanks.

trxcllnt:
@blesh isn't the real culprit here the incomplete `WebSocketSubject` [constructor](https://github.com/blesh/RxJS/blob/7797fa0058b3a8b6dac0b7d74fb15dece4ce2088/src/observable/dom/WebSocketSubject.ts#L52)? It doesn't set `this.source` if the first parameter is an Observable, and the optional second parameter  `destination` is ignored entirely.

blesh:
@trxcllnt so you're saying here, we could add a check to see if `destination` is passed and if so, use that, otherwise use the ReplaySubject?

trxcllnt:
@blesh yes I think that should work. and probably shouldn't do the `assign()` if the first argument is an Observable.

PR #1764, opened 2016-06-12
fix(bufferTime): handle closing context when synchronously unsubscribed

### LINKED ISSUE #1763: Error on complete when using .bufferTime().take()
**RxJS version:**
rxjs@5.0.0-beta.8

**Code to reproduce:**

[jsbin example](https://jsbin.com/vojelidule/1/edit?html,js,console)
```
var source = Rx.Observable.interval(100)

source
  .bufferTime(1000)
  .take(2)
```

**Expected behavior:**

Should complete without any errors.

**Actual behavior:**

Throws an error on complete:

`TypeError: Cannot read property 'splice' of null`

**Additional information:**

Does work in RxJS 4 - [jsbin example](https://jsbin.com/palahucuso/1/edit?html,js,console)

kwonoj:
**Description:**
This PR updates behaviors of `openContext` and `closeContext`, do not try to update if contexts are cleared by synchronous unsubscription right after emitting value (i.e, `take`).

closes #1763

### DIFF
--- a/spec/operators/bufferTime-spec.ts
+++ b/spec/operators/bufferTime-spec.ts
@@ -284,4 +284,20 @@ describe('Observable.prototype.bufferTime', () => {
     expectObservable(result).toBe(expected, values);
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
+
+  it('should not throw when subscription synchronously unsubscribed after emit', () => {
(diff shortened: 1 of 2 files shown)


kwonoj:
I think change's ok, will check in around on this Sunday to next Monday if there isn't any suggestions.

PR #1765, opened 2016-06-13
fix(cache): get correct caching behavior

### LINKED ISSUE #1628: incorrect behavior of cache()

blesh:
This is an initial pass at fixing the cache operator. There is still a lot to do, but this is merge-worthy IMO. Ideally, the cache operator would use the lift mechanism. Also this is not well optimized as it is introducing a lot of closures. But it works, and that's the point for now.

fixes #1628

### DIFF
--- a/spec/operators/cache-spec.ts
+++ b/spec/operators/cache-spec.ts
@@ -35,11 +35,24 @@ describe('Observable.prototype.cache', () => {
     rxTestScheduler.schedule(() => expectObservable(s1).toBe(expected2), time(sub2));
   });
 
-  it('should replay values and error', () => {
-    const s1 = hot('---^---a---b---c---#     ').cache(undefined, undefined, rxTestScheduler);
-    const expected1 = '----a---b---c---#     ';
-    const expected2 = '                  (abc#)';
+  it('should not replay values after error with a hot observable', () => {
+    const s1 = hot('---^---a---b---c---#  ').cache(undefined, undefined, rxTestScheduler);
+    const expected1 = '----a---b---c---#  ';
+    const expected2 = '                  -';
     const t = time(   '------------------|');
+
+    expectObservable(s1).toBe(expected1);
+
+    rxTestScheduler.schedule(() => {
+      expectObservable(s1).toBe(expected2);
+    }, t);
+  });
+
+  it('should be resubscribable after error with a cold observable', () => {
+    const s1 = cold(  '----a---b---c---#                  ').cache(undefined, undefined, rxTestScheduler);
+    const expected1 = '----a---b---c---#                  ';
+    const expected2 = '                  ----a---b---c---#';
+    const t = time(   '------------------|                ');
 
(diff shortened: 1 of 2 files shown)


blesh:
(finally) lol

blesh:
Merged after quick review from @kwonoj

staltz:
Kind of sad that it's all "rx4", but you know that already.

blesh:
@staltz, yeah, I'll iterate on it.

trxcllnt:
@blesh Is the difference between this and `source.multicast(() => new ReplaySubject()).refCount()` that this only recycles the ReplaySubject when the source emits an error (as opposed to when the subscriber count drops to 0)?

To phrase it another way, it seems like this operator now permanently leaks memory until the source emits an error. If true, this troubles me greatly.

blesh:
Where is the leak? Can it be demonstrated?

trxcllnt:
@blesh 

```
// push 1000000 numbers into cache's ReplaySubject
const source = Observable.range(0, 1000000).cache();
source.take(100000).subscribe({
  complete() {
    console.log('cache's ReplaySubject now has 100000 values')
  }
});

// cache now has a reference to a ReplaySubject with 100000 values and the only way to release the reference for GC is to set source to null (but I can't, since source is a const).
```

matthewwithanm:
Is there documentation anywhere about how `cache()` is supposed to work now?

PR #1766, opened 2016-06-13
refactor(symbol-observable): upgrade symbol-observable to 1.0.0

blesh:
The breaking change in `symbol-observable` is that it now exports in a standard es module format. So it can be consumed as `require('symbol-observable').default` or with just `import whatever from 'symbol-observable';`

### DIFF
--- a/package.json
+++ b/package.json
@@ -199,6 +199,6 @@
   },
   "typings": "./dist/cjs/Rx.d.ts",
   "dependencies": {
-    "symbol-observable": "^0.2.4"
+    "symbol-observable": "^1.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -199,6 +199,6 @@
   },
   "typings": "./dist/cjs/Rx.d.ts",
   "dependencies": {
-    "symbol-observable": "^1.0.0"
+    "symbol-observable": "^1.0.1"
   }
 }
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -6,7 +6,7 @@ import {root} from './util/root';
 import {toSubscriber} from './util/toSubscriber';
 import {IfObservable} from './observable/IfObservable';
 import {ErrorObservable} from './observable/ErrorObservable';
-import * as $$observable from 'symbol-observable';
+import $$observable from 'symbol-observable';
 
 export interface Subscribable<T> {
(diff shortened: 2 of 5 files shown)


kwonoj:
Seems I can't build passes on local / travis both,

```
 shx rm -rf ./dist/global && mkdirp ./dist/global && node tools/make-umd-bundle.js && node tools/make-system-bundle.js && npm-run-all build_closure_core

dev died { Error: Error loading "lib/lib" at file:Z:/github/rxjs/dist/lib/lib.js
Error loading "lib/lib" from "rxjs/Rx" at file:Z:/github/rxjs/dist/cjs/Rx.js
ENOENT: no such file or directory, open 'Z:\github\rxjs\dist\lib\lib.js'
    at Error (native)
  errno: -4058,
  code: 'ENOENT',
  syscall: 'open',
  path: 'Z:\\github\\rxjs\\dist\\lib\\lib.js' }
```

blesh:
Looks like an issue with the `es:main` not importing deeply with the `symbol-observable` package. This is something @IgorMinar was talking about in another issue. I'm not really sure how to fix this, might need to be fixed in symbol-observable.

IgorMinar:
there are no deep imports in rxjs code base, so it's something else.

also the error message looks quite bizarre, looking at https://github.com/blesh/symbol-observable it does look like it's resolving symbol-observable to index.js rather than es/index.js. you might need to tweak your system builder configuration to recognize `jsnext:main` as the primary entry point if you want it to pull in the es version.

blesh:
Yeah, at this point, I'm thoroughly confused by the error message. I guess I'll have to dig in.

blesh:
It's apparently something about `symbol-observable` where saying `module.exports = require('./lib')` no longer finds `./lib/index`, but instead wants to find `./lib/lib`. :\

blesh:
I've updated to 1.0.1, but it seems like the problem still isn't fixed.

kwonoj:
it seems tries to resolve paths in rx itself directly, `/ReactiveX/rxjs/dist/lib/index.js` instead of module's paths. :/ weirdo.

kwonoj:
I've amended builder script bit in #1772 with bump up dependencies, to leverage it as discussion point. It seems updated script resolves `symbol-observable` successfully.

blesh:
Closed in favor of #1772

PR #1770, opened 2016-06-15
chore(package.json): update code coverage script

kwonoj:
**Description:**
This PR updates script for code coverage adopts `nyc` instead of manually remapping source maps aims to reduce scripts.


blesh:
LGTM

PR #1772, opened 2016-06-15
refactor(symbol-observable): bump up symbol-observable@1

kwonoj:
**Description:**
with my :heart: to @blesh ...

- bump up `symbol-observable@1`
- bump up `systemjs-builder@latest` with updating breaking change to its api

#1766 

I've created this PR only for demonstrating based on #1766 , so feel freely close this after updating original one. 

One thing I'm not sure is how to verify sanity of systemjs build with updated scripts.. /cc @robwormald  also.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,7 @@ tmp/
 coverage/
 img/
 spec-js/
-.nyc-output/
+.nyc_output/
 
 # Misc
 npm-debug.log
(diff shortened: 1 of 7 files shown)


kwonoj:
Marking blocked, discussion to prevent accidental check in.

blesh:
The update to SystemJS builder is what solves the global build issues?

kwonoj:
I just did start this by updating dependency version, I don't think version bump makes strong differences but configuration itself (https://github.com/ReactiveX/rxjs/pull/1772/files#diff-2702a38df95d385d0d224c110f92758eR8) was way to solve issue. Maybe applying same config to existing version of builder would works, didn't test it though.

blesh:
System JS hurts my brain.

blesh:
This LGTM.

PR #1774, opened 2016-06-16
fix(multicast): Fixes multicast with selector to create a new source connection per subscriber.

trxcllnt:
**Description:**
Multicast with a selector function should create a new ConnectableObservable for each subscriber to
the MulticastObservable. This ensures each subscriber creates a new connection to the source
Observable, and don't share subscription side-effects.

Multicast with selector now matches Rx4's behavior.

### DIFF
--- a/spec/operators/multicast-spec.ts
+++ b/spec/operators/multicast-spec.ts
@@ -49,9 +49,11 @@ describe('Observable.prototype.multicast', () => {
     connectable.connect();
   });
 
-  it('should accept selectors to factory functions', () => {
+  it('should accept a multicast selector and connect to a hot source for each subscriber', () => {
     const source =      hot('-1-2-3----4-|');
-    const sourceSubs =     ['^           !'];
+    const sourceSubs =     ['^           !',
+                            '    ^       !',
+                            '        ^   !'];
     const multicasted = source.multicast(() => new Subject(),
       x => x.zip(x, (a, b) => (parseInt(a) + parseInt(b)).toString()));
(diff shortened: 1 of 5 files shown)


spec/operators/multicast-spec.ts
```diff
@@ -51,7 +51,9 @@ describe('Observable.prototype.multicast', () => {
 
   it('should accept selectors to factory functions', () => {
     const source =      hot('-1-2-3----4-|');
-    const sourceSubs =     ['^           !'];
+    const sourceSubs =     ['^           !',
+                            '    ^       !',
+                            '        ^   !'];
     const multicasted = source.multicast(() => new Subject(),
       x => x.zip(x, (a, b) => (parseInt(a) + parseInt(b)).toString()));
```
blesh: In here, can we assert that `x` is something expected? Also, can we return something more clear? Maybe a simple `map` like: `x.map(v => v + '!')`. Then just use the values argument to `expectObservable` to assert.

blesh:
Good catch @trxcllnt! Just one little thing in the test

spec/operators/multicast-spec.ts
trxcllnt: @blesh the point of multicast with a selector is to allow multiple subscriptions to the multicast Observable inside the selector function, and one source connection per outer subscriber. By zipping `x` with itself, this test is guaranteeing two subscriptions to `x` for each connection to the multicast source.

I think this is a good test, but I'll add a test with a cold source as well to cover all cases.

PR #1773, opened 2016-06-16
refactor(errors): Updates custom errors to include more descriptive messages.

trxcllnt:
This PR updates our errors to include more information to make debugging easier. For example, this:

```es6
ObjectUnsubscribedError
```
will now look like this:
```es6
ObjectUnsubscribedError: object unsubscribed
    at ObjectUnsubscribedError.Error (native)
    at new ObjectUnsubscribedError (/Users/ptaylor/dev/RxJS/dist/cjs/util/ObjectUnsubscribedError.js:19:26)
    at Subject._subscribe (/Users/ptaylor/dev/RxJS/dist/cjs/Subject.js:94:19)
    at Subject.Observable.subscribe (/Users/ptaylor/dev/RxJS/dist/cjs/Observable.js:56:27)
    at repl:1:6
    at REPLServer.defaultEval (repl.js:272:27)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.<anonymous> (repl.js:441:10)
    at emitOne (events.js:101:20)
```

PR #1778, opened 2016-06-17
Better error messages

blesh:
This is a continuation of #1773 ... just added a commit to fix the other browser issue (I hope)... @kwonoj can you verify?

cc/ @trxcllnt

### DIFF
--- a/spec/helpers/testScheduler-ui.ts
+++ b/spec/helpers/testScheduler-ui.ts
@@ -104,12 +104,25 @@ module.exports = function(suite) {
       .replace(/\\n/g, '\n');
     }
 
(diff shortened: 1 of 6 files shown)


kwonoj:
Yes, I think it's good to go. Failures related with error object mutation's are gone. (just fyi, there was one failure but I suspect that's not related with this change, comes from fragility of current test runner)

trxcllnt:
@blesh @kwonoj thanks again for being so active on this issue (amongst others). It feels great knowing that even though this can't be my day-to-day anymore, you guys are still on top of things when I can make the time to contribute back in my own small way ❤️

PR #1788, opened 2016-06-27
fix(Observable.from): use scheduler if supplied

chrisprice:
**Description:**

Arguments supplied to `Observable.from` are not being propagated correctly. I've chased this a couple of levels down the rabbit hole and whilst this patch fixes some issues, I'm starting to think that Iterators and Schedulers just don't work together. I supply this patch only because the code is so obviously wrong.

N.B. I tried adding tests but I don't understand enough about your testing primitives and their behavior to get the results I would expect. However, I believe the code is currently so obviously wrong that the tests are unnecessary to prove correctness, only to prevent regression.

### DIFF
--- a/spec/observables/from-spec.ts
+++ b/spec/observables/from-spec.ts
@@ -1,6 +1,5 @@
 import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
-import {$$iterator} from '../../dist/cjs/symbol/iterator';
 
 declare const {asDiagram, expectObservable, Symbol, type};
 declare const rxTestScheduler: Rx.TestScheduler;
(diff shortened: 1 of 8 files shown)


blesh:
Yeesh. Good catch, @chrisprice . Do you think you can add a test that verifies the change? (i.e. one that was broken before the change, but works now?)

chrisprice:
This is the rabbit hole I was talking about 😬 The following test will pass -

```js
  it('should handle any iterable thing with a scheduler', (done: MochaDone) => {
    const iterable = <any>{};
    const iteratorResults = [
      { value: 'one', done: false },
      { value: 'two', done: false },
      { done: true }
    ];
    const expected = ['one', 'two'];

    expect($$iterator).to.equal(Symbol.iterator);

    iterable[Symbol.iterator] = () => {
      return {
        next: () => {
          return iteratorResults.shift();
        }
      };
    };

    Observable.from(iterable, x => x, null, Rx.Scheduler.async).subscribe((x: string) =>  {
      expect(x).to.equal(expected.shift());
    }, (x) => {
    }, () => {
      expect(expected.length).to.equal(0)
      done();
    });

    expect(expected.length).to.equal(2);
  });
```

But will fail if the following line is substituted -

```js
    Observable.from(iterable, Rx.Scheduler.async).subscribe((x: string) =>  {
```

The conditionals around the arguments seems very inconsistent across both the `create` function and the constructors invoked by it. E.g. what combinations of arguments are supported? How should arguments be tested? What happens to the `mapFnOrScheduler` and `thisArg` arguments for non-`IteratorObservable`s or `ArrayLikeObservable`s? 

I don't mind chasing a few of these down but are you able to give some some guidance around the API you're aiming for?

blesh:
If something takes a scheduler, it should be an optional last argument. If the arguments before the scheduler are optional, then I'd expect any order would work:

- object
- object, scheduler
- object, fn
- object, fn, scheduler
- object, fn, object
- object, fn, object, scheduler

Should all work.

chrisprice:
I've modified the `from-spec` suite of tests to run a standardised set of expectations for all of the possible Observable sources. Here are the results (after 04637ff910565f561b4b550dd21cd9de9226e580) -

name | object | object, scheduler | object, fn | object, fn, scheduler | object, fn, object | object, fn, object,  scheduler
---|---|---|---|---|---|---
observable|:white_check_mark:|:white_check_mark:|:x:|:x:|:x:|:x:
array|:white_check_mark:|:white_check_mark:|:x:|:x:|:x:|:x:
promise|:white_check_mark:|:white_check_mark:|:x:|:x:|:x:|:x:
iterator|:white_check_mark:|:x:|:x:|:x:|:x:|:x:
array-like|:white_check_mark:|:white_check_mark:|:white_check_mark:|:x:|:white_check_mark:|:white_check_mark:
string**|:white_check_mark:|:x:|:x:|:x:|:x:|:x:
arguments**|:white_check_mark:|:x:|:x:|:x:|:x:|:x:

** Implements iterator protocol so results duplicate iterator.

Before I spend any time fixing all the :x:s, is it worth reconsidering whether the projection and context arguments are useful? I would expect arrow-functions and `map` negate the need for them?

blesh:
@chrisprice I think you raise a valid question. I know that the idea of a projection function as an argument to `Observable.from` was shot down in @zenparsing's spec. Perhaps we should put this PR on hold until the community has time to discuss whether or not we really want to keep it in there. FWIW, I've never once used the projection argument to `from`. Generally, I think that `Observable.from(whatever).map(project)` is more idiomatic and easier to read. I also think we could work on collapsing things like that `map` so there isn't a performance concern there.

spec/observables/from-spec.ts
```diff
@@ -210,18 +24,6 @@ describe('Observable.from', () => {
     expect(r).to.throw();
   });
 
-  it('should handle object has observable symbol', (done: MochaDone) => {
```
blesh: Where did this test go? (Also, I think this test was invalid)

spec/observables/from-spec.ts
```diff
@@ -232,11 +34,71 @@ describe('Observable.from', () => {
...
+  it('should return T for arrays', () => {
     type(() => {
       /* tslint:disable:no-unused-variable */
-      let o1: Rx.Observable<number> = Observable.from(<number[]>[], x => x.toString(), null, Rx.Scheduler.asap);
+      let o1: Rx.Observable<number> = Observable.from(<number[]>[], Rx.Scheduler.asap);
       /* tslint:enable:no-unused-variable */
     });
   });
+
+  const fakerator = (...values) => ({
```
blesh: 👍  to the variable name!

spec/observables/from-spec.ts
```diff
@@ -232,11 +34,71 @@ describe('Observable.from', () => {
...
+  });
+
+  const sources: { name: string, value: any }[] = [
+    { name: 'observable', value: Observable.of('x') },
+    { name: 'array', value: ['x'] },
+    { name: 'promise', value: Promise.resolve('x') },
+    { name: 'iterator', value: fakerator('x') },
+    { name: 'array-like', value: { [0]: 'x', length: 1 }},
+    { name: 'string', value: 'x'},
+    { name: 'arguments', value: function(x) { return arguments; }('x') },
```
blesh: This is only missing one for `[Symbol.observable]()`... which should return an object with a `subscribe(observer: Observer<T>): Unsubscribable`

spec/observables/from-spec.ts
chrisprice: Folded into the matrix test below. I can restore it but I think it would just be a dupe.

src/observable/ArrayObservable.ts
```diff
@@ -101,7 +101,7 @@ export class ArrayObservable<T> extends Observable<T> {
   // value used if Array has one value and _isScalar
   value: any;
 
-  constructor(public array: T[], public scheduler?: Scheduler) {
+  constructor(private array: T[], private scheduler: Scheduler = null) {
```
blesh: Why `= null` and not optional?

src/observable/IteratorObservable.ts
```diff
@@ -61,56 +42,31 @@ export class IteratorObservable<T> extends Observable<T> {
...
 
-  private thisArg: any;
-  private project: (x?: any, i?: number) => T;
-  private scheduler: Scheduler;
-
-  constructor(iterator: any,
-              project?: ((x?: any, i?: number) => T) | any,
-              thisArg?: any | Scheduler,
-              scheduler?: Scheduler) {
+  constructor(iterator: any, private scheduler: Scheduler = null) {
```
blesh: Same thing here, why not just make `scheduler` optional?

src/observable/ArrayObservable.ts
chrisprice: I tried to unify across all the create and ctors, I just went with the most common and rolled it out.

chrisprice:
I think I've addressed all the comments with the latest update -
* Removed all default values from constructors
* Added an explicit test for observable-like to the coverage matrix

jayphelps:
@blesh how's this look besides needing rebase?

blesh:
LGTM

jayphelps:
@chrisprice just needs a rebase then it's merge time 🎉

chrisprice:
Rebased on to 3c35b94 :+1:

jayphelps:
Thanks !!

Issue #1784, opened 2016-06-25
WebSocketSubject seemingly ignores destination argument


deontologician:
**RxJS version:**
master branch

**Code to reproduce:**
https://github.com/ReactiveX/rxjs/blob/master/src/observable/dom/WebSocketSubject.ts#L67
**Expected behavior:**
The second argument to the constructor would be used somewhere
**Actual behavior:**
It is not used
**Additional information:**

I'm thinking the ReplaySubject should be subscribed to the `destination` argument being passed in so it gets its values.

blesh:
@deontologician this is a bit of a tricky one. The "destination" of the WebSocketSubject is the "observer side" of the subject. @trxcllnt and I are doing some fancy things here using ReplaySubject to queue messages that need to be sent over the WebSocket when it opens. There really isn't a reason a user would need to set the `destination` of this particular subject, so we're just leveraging it.

It's definitely confusing though. haha

:)

Does that satisfy your issue? Is there a reason you'd like to use `destination` in WebSocketSubject?

deontologician:
Ah, I thought it might be used to queue up a handshake message before any other requests were sent. So should the destination argument be removed from the constructor?

blesh:
@deontologician I trust this issue has been satisfied then? I'll close it for now, feel free to reopen if you still have questions or you feel there is still an issue?

trxcllnt:
@blesh what's the status on merging #1790?

deontologician:
This issue was just why the destination argument was not being used at all. It looks like #1790 fixed it, so I'm good

PR #1790, opened 2016-06-28
fix(WebSocketSubject): use the supplied WebSockeCtor, and support source/destination arguments in the constructor.

trxcllnt:
**Description:**

#1745, #1784


blesh:
LGTM.

PR #1792, opened 2016-06-30
fix(toPromise): do not resolve when does not emit values

### LINKED ISSUE #1736: toPromise behavior when observable ends without having emited any value
When executing the code below, the result of x is 'undefined'.

``` typescript
Rx.Observable.of().toPromise().then(x => console.log("x is undefined!!! " + x));
```

I think that a promise derived from an observable which is terminated and has not emitted any event should emit an error and not emit an undefined.

Otherwise, this would mean that all future usage of toPromise will result into a 'T | undefined'.

kwonoj:
**Description:**
This PR updates behavior of `toPromise`, let it explicitly resolve only if observable emits any value. If there isn't value emitted, promise will be rejected instead.

/cc @staltz also for review.

closes #1736

### DIFF
--- a/spec/operators/toPromise-spec.ts
+++ b/spec/operators/toPromise-spec.ts
@@ -22,6 +22,15 @@ describe('Observable.prototype.toPromise', () => {
     });
   });
 
+  it('should not resolve if does not emit any value', (done: MochaDone) => {
+    Observable.empty().toPromise(Promise).then(() => {
+      done(new Error('should not be called'));
+    }, (err: any) => {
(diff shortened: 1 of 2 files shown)


blesh:
Sorry, @kwonoj I'm going to have to reject this PR on the grounds that it deviates from RxJS 4's behavior.

See: http://jsbin.com/yuficabasu/1/edit?js,console,output

cc/ @mattpodwysocki

kwonoj:
got it, thanks.

PR #1797, opened 2016-07-01
chore(package.json): trigger lint against staged files only

kwonoj:
**Description:**
This PR introduces module `lint-staged` via wiring existing git hooks to pre-commit. Since checked in codebase is already guarded by full lint, it doesn't necessarily required to trigger full lint each time per check in. This PR makes automatically trigger lint only against staged files, reduced daily-flow time consumption on lint.

- Per each `commit`, lint will be triggered automatically only for staged files. Failure will block commit.
- All other `lint` tasks remain, allows manual full lint if possible.
- `lint_staged` task will allow trigger lint manually against staged files, do not need try commit each time for trigger lint.
- `travis` will run full lint per each build to serve as guard.

NOTE : I think #1731 is still valid when codebase grown, need frequent full lint including CI servers.

relates to #1731.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,13 @@
       "path": "cz-conventional-changelog"
     },
     "ghooks": {
-      "commit-msg": "node ./node_modules/validate-commit-msg/index.js"
+      "commit-msg": "node ./node_modules/validate-commit-msg/index.js",
+      "pre-commit": "npm run lint_staged"
     }
+  },
+  "lint-staged": {
+    "eslint": "*.@(js)",
+    "tslint" : "*.@(ts)"
   },
   "scripts-info": {
     "info": "List available script",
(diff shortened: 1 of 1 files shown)


kwonoj:
LGTM

kwonoj:
above comment was trying to see if I can pass LGTM checker. Seems not.

staltz:
LGTM

staltz:
I think LGTM.co needs at least two confirmations.

kwonoj:
> I think LGTM.co needs at least two confirmations. 

Make sense, let me try later.

kwonoj:
LGTM

kwonoj:
as I've verified this on windows machine only yet, I'll leave this opened and ask opinion to @staltz , @blesh and any other lives on non-windows system.

jayphelps:
LGTM

PR #1805, opened 2016-07-05
feat(PairsObservable): add PairsObservable creation method

### LINKED ISSUE #1804: Observable.pairs(dict) missing in RxJS 5
**RxJS version:**
5.0.0-beta.9

**Code to reproduce:**
`Observable.pairs({a: 1, b: 2})`

**Expected behavior:**
An Observable of `[['a', 1], ['b', 2]]` 

**Actual behavior:**
`Observable.pairs is not a function`

**Additional information:**
https://github.com/Reactive-Extensions/RxJS/issues/704 says the correct way to convert a dictionary to an Observable is with `Observable.pairs(dict)`, but that doesn't seem to be defined in RxJS 5.  Is there a creation operator for dictionaries/maps/Objects?

One of the nice things about collection libraries like Immutable and underscore is that they treat both lists and dictionaries as collections, letting you {map, filter, reduce} as appropriate.  To avoid bloat, it would be nice to use RxJS's collection methods for this, but I don't see any native way to operate on an Object.

kwonoj:
**Description:**
This PR adds `PairObservable` to support generate observable from object has keys.

Marking as 'discussion' for now, since may debatable if this need to be included or not, creating PR as discussion point as implementation is somewhat straightforward.

closes #1804

### DIFF
--- a/spec/observables/pairs-spec.ts
+++ b/spec/observables/pairs-spec.ts
@@ -1,0 +1,43 @@
+import {expect} from 'chai';
+import * as Rx from '../../dist/cjs/Rx';
+
+declare const {hot, asDiagram, expectObservable, expectSubscriptions};
+declare const rxTestScheduler: Rx.TestScheduler;
+const Observable = Rx.Observable;
+
+describe('Observable.pairs', () => {
+  asDiagram('pairs({a: 1, b:2})')('should create an observable emits key-value pair', () => {
+    const e1 = Observable.pairs({a: 1, b: 2}, rxTestScheduler);
+    const expected = '(ab|)';
+    const values = {
+      a: ['a', 1],
+      b: ['b', 2]
+    };
+
+    expectObservable(e1).toBe(expected, values);
+  });
+
+  it('should create an observable without scheduler', (done: MochaDone) => {
+    let expected = [
+      ['a', 1],
+      ['b', 2],
+      ['c', 3]
+    ];
+
+    Observable.pairs({a: 1, b: 2, c: 3}).subscribe(x => {
+      expect(x).to.deep.equal(expected.shift());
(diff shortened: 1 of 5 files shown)


blesh:
LGTM. Please add docs and use cases example

kwonoj:
Rebased and updated examples in jsdoc comment as well, (https://github.com/ReactiveX/rxjs/pull/1805/files#diff-e195dbaac260e2c3a7872befed9bc3ddR42) please let me know if there's document need to be added I missed.

jayphelps:
@kwonoj sorry bud...missed this before the last PR merge so it needs a rebase once again... 😢 I think we need to convince @blesh to disable forcing up-to-date even if no conflicts. In every other major repo I've been apart of, it's so rare that it would cause issues that wouldn't be caught vs. the pain of this "rebase and someone has to immediately merge it or else"

kwonoj:
> so it needs a rebase once again

: no worries, it's ok to me. :)

AFAIK this rebase checker was triggered when introducing LGTM checker. I also agree it'd be better to disable it to not require rebase if there isn't conflict, but it needs to play well with LGTM checker as well I believe.

jayphelps:
@kwonoj it's actually related to the fact that LGTM requires enabling "protected branches" and when it becomes enabled it defaults to some very strict settings.

image

The option "**Require branches to be up to date before merging**" just needs to be unchecked to remove this "out-of-date but no conflicts" block. We did that with [redux-observable](https://github.com/redux-observable/redux-observable).

kwonoj:
> The option "Require branches to be up to date before merging" just needs to be unchecked to remove this "out-of-date but no conflicts" block.

: Gotcha, that makes sense.

PR #1807, opened 2016-07-05
fix(babel): fix an issue where babel could not compile `Scheduler.async`

### LINKED ISSUE #1806: rxjs-es build cannot be complied by Babel
@blesh As a follow up from https://twitter.com/BenLesh/status/750378713905192960, here is an issue for this workaround that I've seen in a couple of code bases.

**RxJS version:**

5.0.0-beta.9

**Code to reproduce:**

Compile rxjs source with Babel.

See existing workarounds as example:
* https://github.com/mike-north/ember-rx-shim/blob/master/index.js#L10
* https://github.com/orbitjs/ember-orbit/blob/master/index.js#L28

**Expected behavior:**

Babel compiles rxjs-es source without errors.

**Actual behavior:**

Babel throws Unexpected Token error similar to this: https://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&code=let%20Scheduler%20%3D%20%7B%0A%20%20asap%2C%0A%20%20async%2C%0A%20%20queue%2C%0A%20%20animationFrame%0A%7D%3B

blesh:
This moves the `async` property to be the last one defined, because babel will then not choke on the `,` after the property.

fixes #1806

related to https://phabricator.babeljs.io/T7479

### DIFF
--- a/src/Rx.ts
+++ b/src/Rx.ts
@@ -187,9 +187,9 @@ import observable from 'symbol-observable';
  */
 let Scheduler = {
   asap,
-  async,
   queue,
-  animationFrame
+  animationFrame,
+  async
 };
 
 /**


kwonoj:
Change looks good to me. (Would like to see LGTM worksmmm ;) )

kwonoj:
LGTM

kwonoj:
hmm, seems I can't make LGTM passes. :/

blesh:
LGTM

kwonoj:
@blesh I think https://github.com/ReactiveX/rxjs/pull/1808 is reason for LGTM not working.

blesh:
LGTM LGTM LGTM

lol!

blesh:
Woot! it worked!

kwonoj:
:yay:

kwonoj:
@blesh I took this PR and just merged it :)

dacz:
I suppose to have this problem. RxJS installed from npm and it doesn't compile with babel. How may I apply the fix? Thx.

kwonoj:
@dacz unless you're directly using master via npm link or some other way, to get latest changes you may need wait next package published. I believe it'll be done quite soon, but hard to provide exactly when.

dacz:
ok, thanks, I'll wait.

PR #1813, opened 2016-07-08
fix(AjaxObservable): drop resultSelector support in ajax method

### LINKED ISSUE #1783: ajax() returns Observable<{}> instead of Observable<AjaxResponse>
**RxJS version:** 5.0.0-beta.9

**Code to reproduce:**

```typescript
const source = Observable.ajax({ method:"GET", url: "myurl" });
```

**Expected behavior:**

Mouse over _source_ to see that it's type _Observable&lt;AjaxResponse>_

**Actual behavior:**

It's type _Observable&lt;{}>_.

**Additional information:**

What might be going on is that _AjaxObservable&lt;T>.create_ has

```typescript
      return new AjaxObservable(urlOrRequest);
```
and I think it should be:
```typescript
      return new AjaxObservable<AjaxResponse>(urlOrRequest);
```

As a workaround I was able to do:

```typescript
const source = Observable.ajax<AjaxResponse>({ method:"GET", url: "myurl" });
```

... but _ajax_ shouldn't really take a type at all and I'm not sure why it even can.

kwonoj:
**Description:**

This PR drops support of `resultSelector` in `ajax*()` mehod.

- by removing `resultSelector`, ajax method delivers strict `AjaxResponse` type resolve type inference issues on subscription.
- when it's needed to select specific values, encourage chain `map` as necessary. 

closes #1783

BREAKING CHANGE: ajax.*() method no longer support resultSelector, encourage to use `map` instead

### DIFF
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -141,68 +141,6 @@ describe('Observable.ajax', () => {
     });
   });
 
-  it('should have an optional resultSelector', () => {
-    const expected = 'avast ye swabs!';
-    let result;
-    let complete = false;
-
-    const obj = {
-      url: '/flibbertyJibbet',
-      responseType: 'text',
-      resultSelector: (res: any) => res.response
-    };
-
-    (<any>Rx.Observable.ajax)(obj)
-      .subscribe((x: any) => {
-        result = x;
-      }, null, () => {
-        complete = true;
-      });
-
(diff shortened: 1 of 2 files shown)


jayphelps:
LGTM

jayphelps:
LGTM

jayphelps:
Needs rebase, then I'll merge.

kwonoj:
Thanks, rebased.

PR #1815, opened 2016-07-09
feat(switch): allow early completion triggered by the inner observable

tklever:
Currently, switch (and its derivatives) will unsubscribe from an inner observable when notifyComplete is triggered, however, it will remain open indefinitely until _complete is called. The patch optionally allows a user to choose to use switch as it currently works, or fall back to the root OuterSubscriber style behavior (call complete on the outer Observable)

With this option enabled, switch's Complete and Error will work very similarly, in that any Error or Complete in the inner Observable will short circuit the outer observable.


trxcllnt:
@tklever thanks for the PR. Is there any difference between the feature in this PR, and an inner observable emitting an error and converting to completion after the switch?

tklever:
In technical execution, no, there is not any difference (although one could make a case for ease-of-use). The issue is purely semantics between an "Error" state, and "Completed" state.

I understand that usually, semantics isn't a wildly overpowering reason to add a feature, as there are actually many existing ways to achieve what I'm attempting in this PR. However, I'd propose that this feature would allow a user to achieve more concise, fluent, self-describing code. 

The biggest issue I see with using the error functionality (which, to your point, absolutely works), is that I am left with error handlers that must make decisions about "fake" errors and "real" errors. I'm using the mechanics for "error" to express "working as designed". I'm left to catch `throw ErrorButNotReallyWinkWinkException`.

The reason I feel a change like this makes additional sense on the switch functionality (verses say, merge) is that switch is specifically dealing with a single inner observable. It disposes of the previous one when it gets a new one, essentially making it a wrapper for easy, iterative subscribe/unsubscribe. 

This feature optionally (I think optionally is important here, as I can see cases for both sides) allows the singular inner observable to ask the outer to complete.

trxcllnt:
@tklever I agree. I've run into this exact scenario recently, and used errors, but it would've been more convenient to have the early-termination behavior as an operator.

I'm primarily concerned that using a boolean in the result selector position to mean the operator should terminate early is somewhat cryptic, and that there's no precedent in other operators. What do you think? Maybe do a combination of flatten + takeUntil operators as `mergeUntil`, `concatUntil`, `switchUntil` etc.?

tklever:
`switchUntil`, in my opinion, would be a confusing departure from the the other `until` methods, as the existing `until`s receive signals from an outside 3rd party observable stream, not the stream itself.

As for precedence, I attempted to model the API after those that use `Scheduler` (`.timer()` specifically), floating free as the final argument, but if you find that confusing (or have an alternative), I'll certainly update this PR as you see fit. I'm not tied to the API and I agree that it's sort of an oddball now.

If we want to make it a separate set of operators, we could do `switchUntilComplete`, `switchMapUntilComplete` & `switchMapToUntilComplete`, but I worry those are a little verbose. Additionally, I'd probably want to export many of the `switch` internals, as we would be looking at a fair amount of code duplication. That very code duplication is why I proposed this is the first place, as it is SO CLOSE to `switch`, but just slightly different, it seemed like an update not a whole new thing.

I'll certainly update this to fall in line with the style and vision of the project maintainers, I just couldn't think of a better place for this.

blesh:
Can you explain the use cases you've hit here? It will help me understand the spirit of the PR.

tklever:
The use case I specifically hit was implementing something very similar to timeout, but using a custom observable type. This custom observable had a few variances such as counting down instead of up, emitting a "time remaining", but none of those specifics really matter. The key point is this observable emitted complete when it was finished counting down to it's goal instead of error (like timeout does). The error handler was used purely to handle full stop errors. Completely valid in isolation, it didn't play nice as soon as I started using operators on it.

Now, I certainly could have extended that original observable, and forced it (somewhat incorrectly) to translate completes to errors. But, that would have solved only that very specific use case, in a somewhat ugly way. This also forces my hand as a developer in a somewhat weird way. I can create an observable any way I like, but I must always remember that if I want allow my users to short circuit a switch, I must use the error handler (be it an error or not). At that point, I'm not building operators around how my observables work, I'm building observables around what the operators will allow me to do. That strikes me as somewhat backwards and a passive coupling.

Alternatively, this fix could probably be implemented as a "notifyCompleteHandler" parameter instead of a boolean, giving almost limitless freedom once notifyComplete is called. That just seemed like a bigger change than I should submit without running it by someone on the team. If that strikes your fancy, I could explore what that would look like.

@trxcllnt also mentioned he ran into this, so if he has an additional use case, I'd love to hear what he struggled with.

trxcllnt:
@tklever after giving it more thought, I think you can do this via `materialize` and `dematerialize`:
```es6
source
  .switchMap((x) => createInnerObs(x).materialize())
  .dematerialize()
  .subscribe();
```

tklever:
@trxcllnt, that does indeed satisfy my use case. Just ran the tests with this method and it works. I'm going to bundle it up in an operator (for ease of use) in my project, but I'm going to close and withdraw this PR. 

Thank you for your guidance.

PR #1816, opened 2016-07-09
test(switch): update test case to use correct operator

tklever:
This test case appears to be the victim of a copy paste error, this particular test for switch was not calling switch at all. Updated to call switch and it still passes.


kwonoj:
change looks good to me.

jayphelps:
LGTM

PR #1819, opened 2016-07-10
refactor(throwError): remove utility throwError() function

kwonoj:
**Description:**

This PR removes utility function `throwError`, which simply wraps `throw` and only being used in single place. Error can be thrown directly.

### DIFF
--- a/src/BehaviorSubject.ts
+++ b/src/BehaviorSubject.ts
@@ -1,7 +1,6 @@
 import {Subject} from './Subject';
 import {Subscriber} from './Subscriber';
 import {Subscription, ISubscription} from './Subscription';
-import {throwError} from './util/throwError';
 import {ObjectUnsubscribedError} from './util/ObjectUnsubscribedError';
 
 /**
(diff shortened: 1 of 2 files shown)


kwonoj:
Not sure why coverage's decreased :/

jayphelps:
Was this being done for performance reasons? I'm not aware of any, but wouldn't be surprised if wrapping the throw inside a function some how prevented a deopt or something. @blesh ?

kwonoj:
> Was this being done for performance reasons? I'm not aware of any

I was also bit curious, sending out PR as discussion point, especially this has been used in only single place of codebase.

blesh:
TMK, `throw` doesn't cause a deopt. I'm not sure why this function existed. 

Need to make sure the code coverage stays up, other than that, it LGTM

jayphelps:
coveralls doesn't make sense to me. It says in the list of changes that there was no change in coverage, but then claims -0.001% overall change. Seems like a rounding error or something...

jayphelps:
this LGTM

kwonoj:
> coveralls doesn't make sense to me. It says in the list of changes that there was no change in coverage, but then claims -0.001% overall change. Seems like a rounding error or something

I assume similar, some kind of threshold between line changes. Tried dummy changes but no luck. :/
Let me check this in once current travis result comes back. Thanks for review @blesh , @jayphelps .

PR #1825, opened 2016-07-12
refactor(Operator): replace operator<T, R> to interface

kwonoj:
**Description:**

This PR replaces implementation `Operator<T, R>`  into `interface` by base implementation `Operator::call` is not necessarily needed, since creating new operator requires to override it any time. Instead, this PR makes operator as contract interface only.

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -571,9 +571,8 @@ describe('Observable.lift', () => {
     }
 
     // The custom Operator
-    class LogOperator<T, R> extends Rx.Operator<T, R> {
+    class LogOperator<T, R> implements Rx.Operator<T, R> {
       constructor(private childOperator: Rx.Operator<T, R>) {
-        super();
       }
 
       call(subscriber: Rx.Subscriber<R>, source: any): TeardownLogic {
--- a/src/Operator.ts
+++ b/src/Operator.ts
@@ -1,8 +1,6 @@
 import {Subscriber} from './Subscriber';
 import {TeardownLogic} from './Subscription';
 
-export class Operator<T, R> {
-  call(subscriber: Subscriber<R>, source: any): TeardownLogic {
-    return source._subscribe(new Subscriber<T>(subscriber));
-  }
+export interface Operator<T, R> {
+  call(subscriber: Subscriber<R>, source: any): TeardownLogic;
 }


saneyuki:
I agree this change

trxcllnt:
@kwonoj initially I wrote the `Operator` as a base class because I intended to have a singleton instance on the Observable class's prototype, which would allow us to remove the [branching](https://github.com/ReactiveX/rxjs/blob/master/src/Observable.ts#L92) in `subscribe`. I'd have to circle back around to see if this would hook up the subscription chain in the right order (might require a change in the [`Subscriber` constructor](https://github.com/ReactiveX/rxjs/blob/master/src/Subscriber.ts#L72)), but if so, it'd look something like this:

```ts
export class Operator<T, R> {
  call(subscriber: Subscriber<R>, source: any): TeardownLogic {
    return source._subscribe(subscriber);
  }
}

export class Observable<T> implements Subscribable<T> {
  /* ... */
  subscribe(observerOrNext?: PartialObserver<T> | ((value: T) => void),
            error?: (error: any) => void,
            complete?: () => void): Subscription {

    const sink = toSubscriber(observerOrNext, error, complete);

    sink.add(this.operator.call(sink, this));

    if (sink.syncErrorThrowable) {
      sink.syncErrorThrowable = false;
      if (sink.syncErrorThrown) {
        throw sink.syncErrorValue;
      }
    }

    return sink;
  }
}
Observable.prototype.operator = new Operator<any, any>();
```

kwonoj:
Maybe I'm not catching all correctly, is it need all implementation of operator have base classes instead of `Observable` have specific impl for those behaviors like below?

```js
export interface Operator {
  call(...): TearDownLogic
}

class DefaultOperator<any, any> implements Operator {
  call(..): { return source._subscribe(...) };
}

export class Observable {
  private operator = new DefaultOperator();
}
```

trxcllnt:
@kwonoj yes, or even a private `operator` function on Observable:

```ts
export class Observable {
  private operator(source): TeardownLogic {
    // `this` refers to the Subscriber, because
    // Observable does `this.operator.call(sink, this)`
    return source._subscribe(this); 
  }
}
```

kwonoj:
Rebased. I think this change can be checked in and @trxcllnt 's change can follow later. Suggestions? /cc @trxcllnt , @jayphelps

jayphelps:
LGTM

kwonoj:
cool, thanks @jayphelps ! :)

PR #1828, opened 2016-07-12
chore(observable): remove hard dependency on symbol-observable

robwormald:
per discussion with @blesh, removing the hard dependency on symbol-observable to allow existing implementations (see: corejs/es7/observable) to polyfill instead of `symbol-observable`, and provide a fallback.

### DIFF
--- a/spec/symbol/iterator-spec.ts
+++ b/spec/symbol/iterator-spec.ts
@@ -3,7 +3,7 @@ import {expect} from 'chai';
 import {root} from '../../dist/cjs/util/root';
 import {$$iterator} from '../../dist/cjs/symbol/iterator';
 
-describe('rxSubscriber symbol', () => {
+describe('iterator symbol', () => {
   it('should exist in the proper form', () => {
     const Symbol = root.Symbol;
     if (typeof Symbol === 'function') {
--- a/spec/symbol/observable-polyfilled-spec.ts
+++ b/spec/symbol/observable-polyfilled-spec.ts
@@ -1,0 +1,10 @@
+import {expect} from 'chai';
+
+import {getSymbolObservable} from '../../dist/cjs/symbol/observable';
+
+describe('observable symbol', () => {
+  it('should exist in the proper form when Symbol does not exist', () => {
+    let $$observable = getSymbolObservable({Symbol: undefined});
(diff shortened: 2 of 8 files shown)


spec/symbol/iterator-spec.ts
```diff
@@ -3,7 +3,7 @@ import {expect} from 'chai';
 import {root} from '../../dist/cjs/util/root';
 import {$$iterator} from '../../dist/cjs/symbol/iterator';
 
-describe('rxSubscriber symbol', () => {
+describe('iterator symbol', () => {
```
blesh: lol... funny that was missed. No one EVER copies and pastes here... EVAR.

spec/symbol/observable-polyfilled-spec.ts
```diff
@@ -0,0 +1,15 @@
+import {expect} from 'chai';
+
+import {root} from '../../dist/cjs/util/root';
+import {getSymbolObservable} from '../../dist/cjs/symbol/observable';
+
+describe('rxSubscriber symbol', () => {
```
blesh: and then... LOL... you fell victim to the same thing.

src/symbol/observable.ts
```diff
@@ -0,0 +1,23 @@
...
+    } else {
+        $$observable = Symbol('observable');
+        Symbol.observable = $$observable;
+    }
+  } else {
+    $$observable = '@@observable';
+  }
+
+  return $$observable;
+}
```
blesh: can we have this just return the `$$observable` from the module? That way it's cached and we don't need to call `getSymbolObservable()` more than once.

I'd prefer that we had two named exports:  `getSymbolObservable(root)` (so we can test it with different root values), and `$$observable` which is really just `getSymbolObservable(root)` having been called and exported.

kwonoj:
(not sure) is it needed to move `symbol-observable` from `dependency` to `peerDependency` or `optionalDependencies`?

trxcllnt:
LGTM

jayphelps:
@trxcllnt I'd like to see what @robwormald says about Ben's comment  https://github.com/ReactiveX/rxjs/pull/1828#discussion_r70529070

robwormald:
@jayphelps I do believe I have made the change Ben suggested

PR #1830, opened 2016-07-14
fix(ObservableAjax): only set default Content-Type header on non-GET

elliotschi:
**Description:**

### DIFF
--- a/src/observable/dom/AjaxObservable.ts
+++ b/src/observable/dom/AjaxObservable.ts
@@ -198,7 +198,7 @@ export class AjaxSubscriber<T> extends Subscriber<Event> {
     }
 
     // ensure content type is set
-    if (!('Content-Type' in headers) && !(root.FormData && request.body instanceof root.FormData)) {
+    if (!('Content-Type' in headers) && !(root.FormData && request.body instanceof root.FormData) && request.method !== 'GET') {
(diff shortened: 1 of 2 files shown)


jayphelps:
nitpicking: although very odd to do so, some APIs do in fact require people to include a body with GET requests--I've seen it a number of times. So I'm thinking perhaps we should instead only omit it if they omitted a body as well?

```js
request.method !== 'GET' && typeof request.body === 'undefined'
```

Or even...if they don't provide any body just omit it regardless of what the method is. Not sure if there's any value in providing a content type with no content?

src/observable/dom/AjaxObservable.ts
```diff
@@ -191,7 +191,7 @@ export class AjaxSubscriber<T> extends Subscriber<Event> {
     }
 
     // ensure content type is set
-    if (!('Content-Type' in headers) && !(root.FormData && request.body instanceof root.FormData)) {
+    if (!('Content-Type' in headers) && !(root.FormData && request.body instanceof root.FormData) && !(typeof request.body === 'undefined')) {
```
jayphelps: `typeof request.body !== 'undefined'` please. The reason the others have it outside is because it's the only way to negate `in` or `instanceof`.

elliotschi: Fixed

jayphelps:
LGTM

Issue #1827, opened 2016-07-12
`WebSocketSubject` closes underlying socket when any observer unsubscribes


marshall007:
**RxJS version:** `5.0.0-beta.10`

**Code to reproduce:**

JSBin: http://jsbin.com/xisajopedo/3/edit?js,console

``` js
let socket = Rx.Observable.webSocket({
  url: 'ws://echo.websocket.org',
  resultSelector: e => e.data,
  openObserver: {
    next: () => console.log('open'),
    error: e => console.error(e)
  },
  closeObserver: {
    next: () => console.log('close')
  }
});

socket.subscribe(
  e => console.log(`res: ${e}`),
  e => console.error(e),
  () => console.log('complete')
)

socket.multiplex(() => 'handshake_start', () => 'handshake_end', () => true)
  //.takeUntil(n => n !== 'handshake_start')
  .subscribe(null, null, () => console.log('handshake_complete'))

socket.multiplex(() => 'query_start', () => 'query_end', () => true)
  .takeUntil(Rx.Observable.interval(1000).take(2))
  .subscribe(null, null, () => console.log('query_complete'))
```

**Expected behavior:**

I would expect that the socket would remain open after the `query` subscription completes since the `handshake` subscription is still open.

**Actual behavior:**

The underlying socket is closed once the first observer unsubscribes, thus unsubscribing all other observers as well.

deontologician:
Looking at this, it seems like this guard: https://github.com/ReactiveX/rxjs/blob/master/src/observable/dom/WebSocketSubject.ts#L206 might need to also check that `this._output.observers.length === 0`

PR #1831, opened 2016-07-14
fix(WebSocketSubject): prevent early close

deontologician:
**Description:**
This prevents the WebSocketSubject from prematurely closing the underlying websocket before all subscriptions have completed.

ReactiveX/rxjs#1827


kwonoj:
I think additional test case can be added for this behavior changes. (check stub socket.close being called only if condition mets)

deontologician:
Alright, test added. I've confirmed it fails on master and succeeds on this branch (on master the second `B next` is not received)

deontologician:
Has anyone had a chance to check this out?

kwonoj:
For me this change looks ok, need other reviewer's approval as well.

mattpodwysocki:
LGTM

deontologician:
Thanks!

PR #1838, opened 2016-07-19
proposal: chore(package.json): upgrade typescript to 2.0

saneyuki:
- **BREAKING CHANGE**
  - This does not cause the breaking change directly, but after merging this change, our codes will get a potentially breaking changes for all change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   },
   "lint-staged": {
     "eslint": "*.@(js)",
-    "tslint" : "*.@(ts)"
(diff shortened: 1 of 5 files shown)


kwonoj:
As similar to what we did in module augmentation in 1.8, I'd like to defer to get public release. There are number of dependencies (like ng2), better to align typescript compilers as well.

saneyuki:
@kwonoj okay. I'll close this at this moment.

Issue #1656, opened 2016-04-26
Subscription.add should return the subscription.


Dorus:
It would be really nice if we could chain subscriptions with `.add()`.

Previously i wrote the following code:

```JS
function publish(source, selector) {
  return Rx.Observable.create(ob => {
    var xs = source.publish();
    return selector(xs).subscribe(ob).add(xs.connect());
  });
}
```

But this is actually wrong. It would be nice if this would be possible. Currently we need to write:

```JS
function publish(source, selector) {
  return Rx.Observable.create(ob => {
    var xs = source.publish();
    var sub = selector(xs).subscribe(ob)
    sub.add(xs.connect())
    return sub;
  });
}
```

david-driscoll:
It looks like it's falling into this trap: https://github.com/ReactiveX/rxjs/blob/master/src/Subscription.ts#L133

We should probably `return this` in that scenario.

blesh:
> We should probably return this in that scenario.

Actually it should return `Subscription.EMPTY`. The subscription returned is the subscription made to wrap the teardown (if necessary). 

[See the docs here](https://github.com/ReactiveX/rxjs/blob/76dcafc7b45d035b1c3d348ae7d1a590ad8602c8/src/Subscription.ts#L124-L127)

(Also @david-driscoll, remember to hit <kbd>Y</kbd> when copying a link to lines of code in github to get the exact commit SHA in the url)

david-driscoll:
Thanks for the tip, didn't know that! :smile: 

Wouldn't there be two cases?  In the above scenario, if they pass in a specific subscription, shouldn't it return that subscription?  It's entirely possible I'm missing something though.

```
if (!teardown || teardown === Subscription.EMPTY)
    return Subscription.EMPTY;

if (teardown === this)
    return this;
```

BrainCrumbz:
Hello all, been pointed here by @Dorus , thanks for that. Don't know if the following request fits into current issue, let's see:

We're working with RxJs 5 in TypeScript, coming from a C# Reactive Extensions background. I'm looking on how to *extend* (as in C# Extension Methods) the `Subscription` class.
I'd like to provide a new (instance) method `Subscription.addTo(collector: Subscription)` which adds the instance to passed collector `Subscription`.
Aim is to be able to write this type of code:

```ts
class SomeClass {
  constructor(/*...*/) {
    someObservable$.subscribe(value => /*...*/).addTo(this.collector);
[...]

PR #1839, opened 2016-07-19
chore(typings): fix compiler errors when using tsc v2.0

saneyuki:
- This just for enhancement to fix the errors when using tsc v2.0.
  - ...and intend to decrease our effort when we upgrade our typescript compiler.

### DIFF
--- a/src/Subject.ts
+++ b/src/Subject.ts
@@ -125,8 +125,9 @@ export class Subject<T> extends Observable<T> implements ISubscription {
  * @class AnonymousSubject<T>
  */
 export class AnonymousSubject<T> extends Subject<T> {
-  constructor(protected destination?: Observer<T>, protected source?: Observable<T>) {
+  constructor(protected destination?: Observer<T>, source?: Observable<T>) {
     super();
+    this.source = source;
   }
 
   next(value: T) {
--- a/src/observable/GenerateObservable.ts
+++ b/src/observable/GenerateObservable.ts
@@ -245,7 +245,7 @@ export class GenerateObservable<T, S> extends Observable<T> {
     } while (true);
   }
 
-  private static dispatch<T, S>(state: SchedulerState<T, S>) {
+  private static dispatch<T, S>(state: SchedulerState<T, S>): Subscription | void {
     const { subscriber, condition } = state;
     if (subscriber.isUnsubscribed) {
       return;
--- a/src/observable/BoundNodeCallbackObservable.ts
+++ b/src/observable/BoundNodeCallbackObservable.ts
@@ -127,7 +127,8 @@ export class BoundNodeCallbackObservable<T> extends Observable<T> {
 function dispatch<T>(state: { source: BoundNodeCallbackObservable<T>, subscriber: Subscriber<T> }) {
   const self = (<Subscription> this);
   const { source, subscriber } = state;
-  const { callbackFunc, args, scheduler } = source;
+  // XXX: cast to `any` to access to the private field in `source`.
+  const { callbackFunc, args, scheduler } = source as any;
   let subject = source.subject;
 
   if (!subject) {


src/Subject.ts
```diff
@@ -125,8 +125,9 @@ export class Subject<T> extends Observable<T> implements ISubscription {
  * @class AnonymousSubject<T>
  */
 export class AnonymousSubject<T> extends Subject<T> {
-  constructor(protected destination?: Observer<T>, protected source?: Observable<T>) {
+  constructor(protected destination?: Observer<T>, source?: Observable<T>) {
```
kwonoj: reason for `source` is not protected anymore?

src/Subject.ts
```diff
@@ -125,8 +125,9 @@ export class Subject<T> extends Observable<T> implements ISubscription {
...
     super();
+    this.source = source;
```
kwonoj: (removed above comment, noticed this line later) reason for separating assignment?

src/observable/GenerateObservable.ts
```diff
@@ -245,7 +245,7 @@ export class GenerateObservable<T, S> extends Observable<T> {
     } while (true);
   }
 
-  private static dispatch<T, S>(state: SchedulerState<T, S>) {
+  private static dispatch<T, S>(state: SchedulerState<T, S>): Subscription | void {
```
kwonoj: can we consider this is returning `tearDownLogic`? it's bit broader than `subscription | void` though.

src/Subscription.ts
```diff
@@ -124,7 +124,7 @@ export class Subscription implements ISubscription {
    * `remove()` to remove the passed teardown logic from the inner subscriptions
    * list.
    */
-  add(teardown: TeardownLogic): Subscription {
+  add(teardown: TeardownLogic): Subscription | void {
```
kwonoj: This is related with https://github.com/ReactiveX/rxjs/issues/1656, better to make `add` explicitly return subscriptions for all cases instead of allowing void, which closes issue.

src/Subject.ts
saneyuki: @kwonoj 

The previous code will causes this error
```
dist/cjs/src/Subject.ts(127,14): error TS2415: Class 'AnonymousSubject<T>' incorrectly extends base class 'Subject<T>'.
  Property 'source' is optional in type 'AnonymousSubject<T>' but required in type 'Subject<T>'.
```

This is still protected because this `source` is defined in `Observable<T>` and it is inherited. The previous code just only overrides its definitions.

If you don't like this mixture of assignment property, I'll stop to use parameter property for `AnonymousSubject`

kwonoj: got it, thanks for clarification.

src/observable/GenerateObservable.ts
saneyuki: @kwonoj 

I thought it's bit broader for this. This `dispatch` will not return `Function` ever.

kwonoj: Yes, I know. just thought of reusing existing types if possible. It's too complex to introduce another type for this, maybe ok as is.

src/Subscription.ts
saneyuki: > This is related with #1656

I'll see it.

saneyuki:
@kwonoj 

I add the patch to fix  #1656

kwonoj:
I'd like to suggest to https://github.com/ReactiveX/rxjs/pull/1839/commits/feca0a681cc067e9a70ef6c2b4682bac97f1e01f as separate PR, which has different scope / issue than original PT does have.

saneyuki:
@kwonoj 

> I'd like to suggest to feca0a6 as separate PR, which has different scope / issue than original PR does have.


okay. I open a new pull request https://github.com/ReactiveX/rxjs/pull/1840

kwonoj:
LGTM

PR #1840, opened 2016-07-19
refactor(Subscription): return always subscription when calling Subscription.add()

### LINKED ISSUE #1656: Subscription.add should return the subscription.

saneyuki:
BREAKING CHANGE:

This fixes ReactiveX#1656.
This makes Subscription.add() returns always subscription.
if you checks the returned value of it. you might need to change it.

### DIFF
--- a/spec/Subscription-spec.ts
+++ b/spec/Subscription-spec.ts
@@ -78,4 +78,83 @@ describe('Subscription', () => {
       done();
     });
   });
+
+  describe('Subscription.add()', () => {
+    it('Should returns the self if the self is passed', () => {
+      const sub = new Subscription();
+      const ret = sub.add(sub);
+
+      expect(ret).to.equal(sub);
+    });
+
+    it('Should returns Subscription.EMPTY if it is passed', () => {
+      const sub = new Subscription();
+      const ret = sub.add(Subscription.EMPTY);
+
+      expect(ret).to.equal(Subscription.EMPTY);
+    });
+
+    it('Should returns Subscription.EMPTY if it is called with `void` value', () => {
+      const sub = new Subscription();
+      const ret = sub.add(undefined);
+      expect(ret).to.equal(Subscription.EMPTY);
+    });
+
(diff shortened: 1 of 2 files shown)


kwonoj:
sorry for nitpicking - doesn't `breaking` need to be in commit message footer? (https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md#footer)

saneyuki:
@kwonoj I fixed the message.

saneyuki:
@kwonoj 

rebased!

spec/Subscription-spec.ts
```diff
@@ -78,4 +78,83 @@ describe('Subscription', () => {
...
+      };
+      const ret = sub.add(arg);
+
+      expect(ret).to.equal(arg);
+    });
+
+    it('Should returns the passed one if passed a AnonymousSubscription having not function `unsubscribe` member', () => {
+      const sub = new Subscription();
+      const arg = {
+        isUnsubscribed: true,
```
saneyuki: oops. this should be `false` for testing correctly

saneyuki:
@kwonoj I rebased this on the latest master.

jayphelps:
LGTM, one more rebase and I'll merge!

jayphelps:
Thank you! :dancer:

saneyuki:
you're welcome!

Issue #1686, opened 2016-05-05
distinctUntilChanged seems to be comparing same value when used with combineLatest


eccheung4:
**RxJS version:5.0.0-beta.7**

**Code to reproduce:**
```
const test$ = Rx.Observable.from([1,2,3]);

Rx.Observable.combineLatest(test$)
    .do(x=>console.log('combineLatest:', x))
    .distinctUntilChanged((x,y)=>{console.log('Compare: ', x, y, x === y); return x === y})
    .subscribe(x=>console.log('Output: ', x));
```

**Expected behavior:**
```
combineLatest: [1]
Output:  [1]
combineLatest: [2]
Compare:  [1] [2] false
Output:  [2]
combineLatest: [3]
Compare:  [2] [3] false
Output:  [3]
```

**Actual behavior:**
```
combineLatest: [1]
Output:  [1]
combineLatest: [2]
Compare:  [2] [2] true
combineLatest: [3]
Compare:  [3] [3] true
```

trxcllnt:
I believe this is because `combineLatest` doesn't copy its values Array when it emits. This can be worked around by providing a selector function:

```es6
Observable.combineLatest(...sources, (...values) => values)
```

abijr:
I'm also experiencing this behavior when using `distinctUntilChanged` after `debounceTime`.

jadbox:
@abijr Did you try @trxcllnt suggested workaround? 

It seems that combineLatest reuses the same array reference for produced values. For performance, this is ideal to avoid creating new arrays, but it has the side affect of causing distinctUntilChanged to break since it's held reference will always be the same as the latest.

Perhaps a possible long term fix is to have distinctUntilChanged do a deep copy for its cached object to check against future values.

abijr:
I'm not using `combineLatest`, I'm using `debounceTime`, and it has a different type signature. I can't use @trxcllnt's workaround (well, I'm almost sure I can't). But, using `filter` instead of `distinctUntilChanged` solved my problem.

trxcllnt:
@abijr are you sure you can't use my workaround? another way of coding it (without ES6) is:
```
sourceA.combineLatest(sourceB, function() {
  return Array.prototype.slice.call(arguments);
})
.distinctUntilChanged(...);
```

david-driscoll:
Under the covers I can understand why you would want to avoid allocating a new array each call.

[...]

PR #1841, opened 2016-07-20
fix(combineLatest): emit unique arrays from the default projection

mark-buer:
**Description:**
Fixes `combineLatest` to emit unique array instances when using the default projection. Without this fix, the default projection emits a shared array (which causes all sorts of problems).

The included test fails against the un-patched `combineLatest` thusly:
```
combineLatest should emit unique array instances with the default projection
(expected c === 3, actual c === 1)
```

**Related issue:**
#1686

### DIFF
--- a/spec/operators/combineLatest-spec.ts
+++ b/spec/operators/combineLatest-spec.ts
@@ -461,4 +461,14 @@ describe('Observable.prototype.combineLatest', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
     expectSubscriptions(e2.subscriptions).toBe(e2subs);
   });
+
+  it('should emit unique array instances with the default projection', () => {
+    const e1 =   hot('-a--b--|');
+    const e2 =   hot('--1--2-|');
+    const expected = '-------(c|)';
+
+    const result = e1.combineLatest(e2).distinct().count();
+
+    expectObservable(result).toBe(expected, { c: 3 });
+  });
 });
--- a/src/operator/combineLatest.ts
+++ b/src/operator/combineLatest.ts
@@ -150,7 +150,7 @@ export class CombineLatestSubscriber<T, R> extends OuterSubscriber<T, R> {
       if (this.project) {
         this._tryProject(values);
       } else {
-        this.destination.next(values);
+        this.destination.next(values.slice());
       }
     }
   }


kwonoj:
Chang looks ok to me. Will leave opened for other reviewer's approval.

staltz:
Hmm, I'm not so welcoming to this change, we should see what impact it has on performance. I suppose it will have a meaningful negative impact. Also, I wouldn't recommend using plain `distinct` when the values are arrays or objects, I would use the keySelector parameter to calculate a hash of the array, in order to compare by contents, not by reference/pointer.

kwonoj:
> we should see what impact it has on performance. I suppose it will have a meaningful negative impact

Yes, there should be perf impact but I'm leaning on behavior first. but that seems very valid concern.

@mark-buer , would you mind to post micro-bench before vs. after to see how much new change loses?

staltz:
RxJS v4 doesn't clone the array. https://github.com/Reactive-Extensions/RxJS/blob/master/src/modular/observable/combinelatest.js#L49

Neither does xstream nor most.js.

Should we wait for more people to give their opinions?

kwonoj:
Discussions at https://github.com/ReactiveX/rxjs/issues/1686, we could use those issue to continue discussion. 

> RxJS v4 doesn't clone the array

it's interesting, I thought issue was created by different behavior between v4 to v5. I just didn't try it myself.

staltz:
`distinctUntilChanged` comparing references of arrays is almost always a bad idea:

- `a === b` may return true even though the *content of those two are different* if they point to the same array instance.
- `a === b` may return false even though the *content of those two are the same* if they point to different array instances.

david-driscoll:
Looking at the source Rxjs4 does clone the array, hence the comments in #1686.

see: https://github.com/Reactive-Extensions/RxJS/blob/29f85627e4ab71b050480f26dcc5d737f1b840f4/src/modular/observable/combinelatest.js#L23-L27

mark-buer:
@staltz This bug fix has no connection with distinct or distinctUntilChanged.

I'm of the opinion that performance should always take second place to correct operation.
If people prefer to have broken but fast behaviour, in this case, they can "opt-in" by supplying their own broken projection function.

I'll repeat my comment from #1686 here:

No additional operators are required to illustrate the incorrect and thus surprising behaviour of `combineLatest`.
This code:
```javascript
var as = new Rx.Subject();
var bs = new Rx.Subject();

// Default projection
Rx.Observable.combineLatest(as, bs)
    .observeOn(Rx.Scheduler.asap)
    .subscribe(function(c) { console.log('default projection [' + c[0] + ', ' + c[1] + ']') });

// Custom projection
Rx.Observable.combineLatest(as, bs, function(a, b) { return [a, b] })
    .observeOn(Rx.Scheduler.asap)
    .subscribe(function(c) { console.log(' custom projection [' + c[0] + ', ' + c[1] + ']') });

as.next(0);
bs.next('x');
as.next(1);
bs.next('y');
as.next(2);
```
Produces the output:
```
default projection [2, y]
 custom projection [0, x]

default projection [2, y]
 custom projection [1, x]

default projection [2, y]
 custom projection [1, y]

default projection [2, y]
 custom projection [2, y]
```
(note the different results between the default projection function and the custom projection function, crazy!)

The lack of array copy within the operator implementation causes incorrect asynchronous results for the default projection.

To fix this:

- for the case where a custom projection function is provided, _no array copy is required_,
- for the case of the default projection function, array copy is required to produce correct results.

Thus, only the default projection code path will be affected.

**If one values (GC) performance above and beyond "correct" behaviour, one could supply a suitably broken projection function. i.e. a projection function that returns a cached array/object.**

trxcllnt:
@mark-buer unfortunately not, since the array of combined values is applied to the projection function. I'm not against this change, but one could argue combineLatest shouldn't copy the Array (for performance) and if required, a unique Array could be created with a custom projection function.

mark-buer:
@trxcllnt Sorry, I wasn't meaning to imply that one could supply a projection function that returns the exact (private) array instance internal to the operator. What I meant was a projection function something like this:
```
let scaryReusedArray = [0, 0];
Rx.Observable.combineLatest(as, bs, function(a, b) {
    scaryReusedArray[0] = a;
    scaryReusedArray[1] = b;
    return scaryReusedArray;
})
```
i.e. the user takes on the responsibility for emitting this cached array that is acting as shared mutable state.

Yes, there is a obviously a performance hit with the apply and the extra redirection. And yes, that array is more reused than the internal one. But I hope that conveys the _gist_ of the idea.

With this bug, we are talking about _default operators_ with perfectly _normal input data_ producing _wrong results_.
Is anyone arguing that adding `.observeOn(Rx.Scheduler.asap)` (or other operators as in #1686) to `combineLatest` _should_ result in nonsensical output data?

I stumbled onto this "WTF" by sheer accident while tinkering with the API (assessing the suitability of rxjs for our new MVVM-based thick client). I lost an hour or so of my life getting to the bottom of this problem. Judging by #1686, I'm not the only one, and _I doubt we will be the last_.

If the broken behaviour is working as intended, it needs to be documented by someone who can justify and explain why it is so. i.e. don't make developers find this surprising behaviour by themselves. The full impact and ramifications of emitting this cached mutable array are widespread, it is a gift that just keeps on surprising as one chains more and more operators and adds asynchronicity into the mix.

**Suggestions (for those who value performance over correctness):**
Add a new operator which replicates the existing cached-array behaviour. This new operator will not allow the use of a custom projection function. Ideas for names are:
`combineLatestCached`, `combineLatestShared` or `combineLatestFast` (or tongue in cheek, `combineLatestSurprise` or `combineLatestPleaseDontUseForAsync`).

Or, add an overload for the existing operator. One could pass in boolean false as the projection function parameter to get `slice` behaviour (the new default), boolean true as the projection function parameter to get cached array behaviour (the current surprising default), or pass in a custom projection function as usual.

I'd like this PR be approved **to fix what is currently broken**, and that a follow up PR addresses potential performance concerns. Or it could be values of this framework don't align with what our team is looking for... no probs, sorry to waste everyones time. We'll seek out an alternative that values correctness and _then_ performance, in that order.

trxcllnt:
LGTM

trxcllnt:
I'm with @mark-buer on this, correct behavior first, then optimize performance as necessary. There are a number of options if this is a performance bottleneck for some use-cases, but honestly if that's the case, there are many other places we should look to optimize before this.

staltz:
For context, this is how RxJS v4 behaves
https://jsbin.com/juyurikofo/1/edit?js,console
because 
https://github.com/Reactive-Extensions/RxJS/blob/master/src/modular/observable/combinelatest.js#L104

david-driscoll:
And adding in rxjs5 behavior things break down.

https://jsbin.com/nunejirubi/1/edit?js,console

jayphelps:
hey @mark-buer we absolutely value correctness over performance as a general guideline. It just sounds like some of us don't believe this is truly "correct".

I'm in the camp of believing it could go either way as far as "correct", but would rather error on the side of copying because it appears to me it's more likely to reduce potential bugs than create them.

@blesh make the call please.

blesh:
This PR LGTM. I think we should add a perf test around shared default array use cases. But regardless, I like this PR.

:shipit:

jayphelps:
@mark-buer rebase plz :shipit:

jayphelps:
Thanks @mark-buer and everyone else for chiming in!

PR #1847, opened 2016-07-26
style(scan): updated signatures for scan and reduce

david-driscoll:
**Description:**
This change makes working with scan/reduce a little easier in TypeScript based applications.

None that I'm aware of.

This change makes working with scan and reduce a little easier without having to cast the types.

### DIFF
--- a/spec/operators/reduce-spec.ts
+++ b/spec/operators/reduce-spec.ts
@@ -1,5 +1,5 @@
 import * as Rx from '../../dist/cjs/Rx';
-declare const {hot, cold, asDiagram, expectObservable, expectSubscriptions};
+declare const {hot, cold, asDiagram, expectObservable, expectSubscriptions, type};
 
 const Observable = Rx.Observable;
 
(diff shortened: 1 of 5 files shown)


paulpdaniels:
Is there any danger in combining the two signature interfaces since they are identical?

kwonoj:
> Is there any danger in combining the two signature interfaces since they are identical?

: Yes, interesting point. Signature seems could be reused, most pain point maybe having good name for common type?

david-driscoll:
The common name I would go with is reduce, simply because they are both reduce operations.  `reduce` just happens on `complete`, and `scan` happens on `next`.

kwonoj:
for me those name looks fine.

blesh:
Nice work @david-driscoll

david-driscoll:
Branch updated.

kwonoj:
LGTM

kwonoj:
@david-driscoll , would you able to check PR? it still says diverged from master, not able to merge.

david-driscoll:
Whoops, one sec.

david-driscoll:
Rebased.

Issue #1834, opened 2016-07-16
Proposal: Throw in FromEventObservable if invalid source


matthewwithanm:
I just spent a while tracking down a bug. It turns out the emitter I was passing to `fromEvent` wasn't one that Rx knows how to handle. It would be cool if that threw an error instead of failing silently.

staltz:
I believe we could accept a PR for this.

matthewwithanm:
👍 otw

kwonoj:
I assume this issue is now resolved? /cc @jayphelps

jayphelps:
Indeed! thanks @kwonoj

PR #1852, opened 2016-07-29
feat(fromEvent): Throw if event target is invalid

matthewwithanm:
**Description:**

Throw instead of failing silently if the event target doesn't implement one of the supported protocols.

#1834

### DIFF
--- a/src/observable/FromEventObservable.ts
+++ b/src/observable/FromEventObservable.ts
@@ -133,6 +133,8 @@ export class FromEventObservable<T, R> extends Observable<T> {
       const source = sourceObj;
       sourceObj.addListener(eventName, handler);
       unsubscribe = () => source.removeListener(eventName, handler);
+    } else {
+      throw new TypeError('Invalid event target');
     }
 
     subscriber.add(new Subscription(unsubscribe));
--- a/spec/observables/fromEvent-spec.ts
+++ b/spec/observables/fromEvent-spec.ts
@@ -115,6 +115,17 @@ describe('Observable.fromEvent', () => {
     expect(offHandler).to.equal(onHandler);
   });
 
+  it('should error on invalid event targets', () => {
+    const obj = {
+      addListener: () => {
+        //noop
+      }
+    };
+
+    const subscribe = () => Observable.fromEvent(<any>obj, 'click').subscribe();
+    expect(subscribe).to.throw(TypeError, 'Invalid event target');
+  });
+
   it('should pass through options to addEventListener', () => {
     let actualOptions;
     const expectedOptions = { capture: true, passive: true };


jayphelps:
@matthewwithanm bunch of tests fail now due to this. Can you check it out?

kwonoj:
> bunch of tests fail now due to this

: most of test failure seems occurred in `markdown-doctest`, which currently triggered in travis only.

src/observable/FromEventObservable.ts
```diff
@@ -133,6 +133,8 @@ export class FromEventObservable<T, R> extends Observable<T> {
       const source = sourceObj;
       sourceObj.addListener(eventName, handler);
       unsubscribe = () => source.removeListener(eventName, handler);
+    } else {
+      throw new TypeError('Invalid event target');
```
blesh: I feel like we're not capitalizing our error messages anywhere else. Can you confirm? cc/ @kwonoj

kwonoj: While there are some of mixtures around, in majority error messages are not capitalized. May need to align existing codebase as well.

matthewwithanm: Sure, np. @jayphelps mentioned it should maybe be more descriptive. Any ideas?

jayphelps: @matthewwithanm nevermind. Let's just shipit :shipit:

blesh:
Looks like some tests need to be updated for this one.

blesh:
Just need to get the markdown doc tests to behave properly. @matthewwithanm basically that means there are code examples in the documentation that are throwing this new error now. You'll need to look at travis and see which ones, then go through and update them.

jayphelps:
@matthewwithanm in case it wasn't known, the code examples in the docs are actually run/evaluated to confirm they work. So it's not the doc generation process itself, it's code examples in the docs that now do not.

matthewwithanm:
Thanks! I'll take a look as soon as I get a sec…pushing out Nuclide releases atm 😉

matthewwithanm:
How do you want this fixed? It looks like the examples are just querying the dom for non-existent elements, so the errors are valid; the code isn't actually runnable without the elements that they take for granted (for example purposes).

kwonoj:
@matthewwithanm if I remember correctly, `markdown-doctest` allows to skip certain code to be tested (https://github.com/Widdershin/markdown-doctest#i-have-a-code-example-i-dont-want-tested). For those, may possibly simply skip them.

kwonoj:
it'd be better if those examples can be amended to conform with new changes, but unless replace queryselector into some other source (or querying actual element exists) I'm not sure if it can be achieved.

jayphelps:
Some of them could be changed to just add the event handler to `window`..but others I agree are more complex and re

kwonoj:
My suggestion is to make best effort as long as it doesn't require lot of boilerplate / bootstraps. Main purpose of example is to serve who read documentations. It is still be great if those code snippet can be verified via actual execution, but it should not distract by introducing additional codes that are not relevant to essence of example presented. (In short, agree with @jayphelps )

jayphelps:
@matthewwithanm we wanna get this out ASAP. We think adding `removeEventListener` noop [mock to here](https://github.com/ReactiveX/rxjs/blob/master/.markdown-doctest-setup.js#L26) might get the markdown tests to pass. If you don't have cycles right now to spare, can you enable "Allow edits from maintainers"? I'm happy to push this over the finish line for you.

image

jayphelps:
merged in #2107

matthewwithanm:
Oh my gosh, so sorry! This completely slipped through the cracks! 😔

jayphelps:
No biggie. It happens! 🤘

Issue #1744, opened 2016-06-06
npm linking consumer with library causes TS2322: Observable not assignable to Observable


vankeer:
**RxJS version:** 5.0.0-beta.6 and also tried npm linking with master today - no change

**Code to reproduce:**

You need:
- A library that exposes a method returning an Observable ([example](https://github.com/vankeer/ng2-example-library/blob/master/src/services/wiki.service.ts#L11))
- A consumer that consumes this method and returns an Observable as well ([example](https://github.com/vankeer/ng2-example-consumer/blob/master/src/client/app/%2Bwiki/wiki.component.ts#L32-L34))
- `npm link` the two before building

Example:

```
git clone https://github.com/vankeer/ng2-example-library.git
cd ng2-example-library
npm i
npm link
cd ..

git clone https://github.com/vankeer/ng2-example-consumer
cd ng2-example-consumer
npm i
npm link ng2-example-library
gulp build.dev
```

**Expected behavior:**

The library and the consumer build normally.

**Actual behavior:**

The consumer throws this error:

> src\client\app\+wiki\wiki.component.ts(33,12): error TS2322: Type 'Observable<any[]>' is not assignable to type 'Observable<any[]>'.
> Property 'source' is protected but type 'Observable<T>' is not a class derived from 'Observable<T>'.

**Additional information:**

This only happens when `npm link`ing the repos.

One way to avoid this problem is by having rxjs methods return an interface for Observable instead of Observable. Of course, this would require a rather large refactoring. PR for this suggested change: #1741 

Another way to avoid the problem is by explicitly installing the typings in the consumer (have not managed to do this yet with this specific example though).

blesh:
It's probably that both libraries should have `rxjs` as a peer dependency, rather than a direct dependency. You'd run into this issue with any library that passed through a type like that. I'm not sure there's anything rxjs can do about it.

vankeer:
Unfortunately having `rxjs` as peer dependency does not fix this problem. I updated my example repos.

I am unsure how to fix this (besides explicitly installing typings in the consumer) and I don't quite understand why this is not an issue for more people? Aren't more people npm linking different libraries like this and running into this same problem? Not sure what I'm missing.

blesh:
@vankeer oh, to clarify, this is a TypeScript build failure? cc/ @david-driscoll

david-driscoll:
I believe it's an issue with the compiler still. I get the issue as well.

[...]

PR #1853, opened 2016-07-29
[WIP] Interface all the things...

david-driscoll:
**Description:**
Let me prefix this with I really wish we didn't have to do this, but looking at the referenced issues I don't see the root problem being solved by the TypeScript compiler anytime soon.

There are problems when using `rxjs` with tools like `npm link` that cause the compiler to fail to resolve types properly and create lots of compiler errors, as it sees two copies of the same class, and they are not compatible.

The goal of this change is to do the following:
* Make the public `Observable` an interface.
* Make other public classes interfaces as well (`Subscription`, `Subscriber`, etc).
* Make sure `Observable` is still the same `Observable` to external consumers.

https://github.com/Microsoft/TypeScript/issues/6496
https://github.com/Microsoft/TypeScript/issues/7755
https://github.com/ReactiveX/rxjs/issues/1744

**Issues**
* Exporting new types will require jumping through similar interface hoops that this PR will solve.

**Work Inprogress Todos**
* [ ] Rename `Obsevable` -> `ObservableImpl`
* [ ] Rename `IObservable` -> `Observable`
* [ ] Create `ObservableStatic` interface
* [ ] Fixup TSLint errors
* [ ] Update tests to use new interfaces (so they compile!)
* [ ] Repeat renames for other interfaces such as `Subscription` (anything that the consumer may want to expose publicly)

cc @blesh @robwormald @kwonoj @staltz 

cc @mhegazy @ahejlsberg - maybe this can be fixed easily in the compiler?  :pray:   

As it stands `rxjs` (and any other library for that matter) cannot export any `class`es without causing potential compiler issues down the line for others.  Essentially we're being penalized because we're using TypeScript to develop the library.

Are there other simpler fixes?  Can the compiler emit an identifier in the `.d.ts` files that so that it knows "hey this is the same interface as the other one I have over here"?


package.json
```diff
@@ -65,7 +65,7 @@
     "build_test": "shx rm -rf ./dist/ && npm-run-all build_cjs clean_spec build_spec test_mocha",
     "build_cover": "shx rm -rf ./dist/ && npm-run-all build_cjs build_spec cover",
     "build_docs": "npm-run-all build_es6_for_docs build_global build_spec tests2png decision_tree_widget && esdoc -c esdoc.json",
-    "build_spec": "tsc --project ./spec --pretty",
+    "build_spec": "tsc --project ./spec",
```
saneyuki: @david-driscoll why did you remove `--pretty` flag?

david-driscoll: It was to get errors to be detected by vscode, this will be restored once the branch gets ready for land.

saneyuki: aha. thank you

david-driscoll:
Closing this for now, as there may be an alternative... opening an issue for that.

PR #1859, opened 2016-08-01
fix(catch): accept selector returns ObservableInput

### LINKED ISSUE #1857: catch() operator does not support selector returning ObservableInput (promise, array, etc)
It needs to be updated to use [subscribeToResult()](https://github.com/ReactiveX/rxjs/blob/master/src/util/subscribeToResult.ts). 

#1246 mentioned this (and other) operator missing it but seems it was closed without confirming all operators had support added.

kwonoj:
**Description:**
This PR utilizes `subscribeToResult` to accept selectors of `catch` can return any kind of `ObservableInput`.

closes #1857

### DIFF
--- a/spec/helpers/test-helper.ts
+++ b/spec/helpers/test-helper.ts
@@ -3,7 +3,10 @@ declare const global: any;
 declare const Symbol: any;
 
 import * as Rx from '../../dist/cjs/Rx';
+import {ObservableInput} from '../../dist/cjs/Observable';
 import {root} from '../../dist/cjs/util/root';
+import {$$iterator} from '../../dist/cjs/symbol/iterator';
+import $$symbolObservable from 'symbol-observable';
 
 export function lowerCaseO<T>(...args): Rx.Observable<T> {
   const values = [].slice.apply(arguments);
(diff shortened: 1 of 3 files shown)


jayphelps:
I believe coverage went down because `subscribeToResult` has numerous paths and the new test is just testing the Promise one of them.

Maybe if we are going to continue these integration style tests we can write a helper to test all the various inputs? something like:

```js
const createObservableInputs = value => Observable.of(
  Observable.of(value),
  Observable.of(value, Scheduler.async),
  [value],
  Promise.resolve(value),
  { [$$iterator]: function* () { yield value; } },
  { [$$observable]: () => Observable.of(value) }
);

it('should accept selector returns any ObservableInput', (done: MochaDone) => {
  const input$ = new createObservableInputs(42);

  input$.mergeMap(input =>
    Observable.throw('bad').catch(err => input)
  ).subscribe(x => {
    expect(x).to.be.equal(42);
  }, (err: any) => {
  }, () => {
    done();
  });
});
```

There's prolly a better way to write this using the TestScheduler helpers...

kwonoj:
> Maybe if we are going to continue these integration style tests we can write a helper to test all the various inputs?

: it's doable, I'll update PR.

> we could test subscribeToResult on its own, 

: While above suggestion (creating test helper) is doable, I think this is more legit way to maintain test coverage to `subscribeToResult`, since it needs to be covered by its own test coverage. I'll take those as well, see if I can create quick PR to enhance test coverage for `subscribeToResult` itself.

kwonoj:
@jayphelps updated test case as suggested, will take `subscribeToResult` as separate PR.

jayphelps:
👏 very nice!

LGTM

blesh:
Small note: Code coverage decreased slightly with this PR. Very, very slightly... but it's worth trying to remember to keep our test coverage high.

jayphelps:
@blesh it did? The last one says it _increased_

> Coverage increased (+0.04%) to 96.783% when pulling e55c62d on kwonoj:catch-observableinput into ff0c613 on ReactiveX:master.

We discussed this above and he updated. Where are you seeing it still decreased?

blesh:
oh I read the one higher up.

NEVERMIND!!!

PR #1862, opened 2016-08-02
test(QueueScheduler): use fake timers for async schedule

kwonoj:
**Description:**

This PR refactors test case of `QueueScheduler` to use fake timers instead of real datetime. There is intermittent test asserition failure by test case designed to check real execution time to verify asynchronous behavior, which can be vary per environment (i.e, virtual machines) leads into test failure. Instead, use fake timers and verify values synchronously.

In further, distinct test coverage for `Action<T>` might needed (`QueueAction`, `AsyncAction`...)


blesh:
LGTM

PR #1864, opened 2016-08-03
Updated introduction examples to use throttleTime in place of throttle

KevinMcIntyre:
**Description:**
While going through the introduction I found that the examples that used throttle threw an error due to it showing the previous usage of taking an integer to describe the interval length in milliseconds. It currently takes a function that returns an Observable, so I updated the document.

### DIFF
--- a/doc/introduction.md
+++ b/doc/introduction.md
@@ -73,7 +73,7 @@ With RxJS:
 ```js
 var button = document.querySelector('button');
 Rx.Observable.fromEvent(button, 'click')
-  .throttle(1000)
+  .throttleTime(1000)
   .scan(count => count + 1, 0)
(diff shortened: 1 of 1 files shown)


blesh:
Good catch. But I think it should have been `throttleTime` instead

KevinMcIntyre:
Ah, ok. I've updated the pull request accordingly.

blesh:
LGTM

blesh:
Thank you, @KevinMcIntyre

KevinMcIntyre:
I'm new to Rx and these docs are pretty awesome; so it's my pleasure to help!

I'll be attending React Rally in a couple weeks, looking forward to your talk!

PR #1871, opened 2016-08-09
refactor(build): No longer building AMD, umd bundle is Rx.js

### LINKED ISSUE #1678: Rx.umd.js vs Rx.js vs AMD
When we first got the "global build" working a while ago, @jeffbcross and @robwormald took the job. The outcome is that we have an `Rx.js` and `Rx.umd.js` file (plus minified counterparts).

1. I think that `Rx.umd.js` should just be `Rx.js` and the other version probably shouldn't exist. I'd like clarification on this separation from @jeffbcross or @robwormald
2. If we're exporting a UMD version of the library, is there any merit to continuing to publish every module in AMD format? Can't most AMD users just use the UMD one? Are there bundling benefits available in AMD (cc @jayphelps)

### Proposal

- Remove current `Rx.js` global build output.
- Rename `Rx.umd.js` global build output to `Rx.js`
- Possibly stop publishing AMD build in favor of guidance to just use `UMD` global output.

Thoughts?

blesh:
This is still using browserify for bundling. This is just the minimal change to close #1678.

If TSC is now providing a built-in way of bundling and UMD output, we should be using that, but for now, browserify is fine.

BREAKING CHANGE: AMD modules are no longer built and supported
BREAKING CHANGE: `Rx.umd.js` and the like are now just `Rx.js`

Closes #1678


blesh:
attn: @kwonoj @jayphelps 

Please review thoroughly. Since this change isn't one where it's easy to detect breakage via automation.

README.md
```diff
@@ -110,6 +110,13 @@ npm install @reactivex/rxjs@5.0.0-beta.1
...
 For CDN, you can use [npmcdn](https://npmcdn.com):
 
+5.0.0-beta.1 - 5.0.0-beta.11:
+
+https://npmcdn.com/@reactivex/rxjs/dist/global/Rx.umd.js
+
+
+5.0.0-beta.11 or higher:
+
 https://npmcdn.com/@reactivex/rxjs/dist/global/Rx.umd.js
```
kwonoj: b.11 or higher may need to point `Rx.js` instead of `Rx.umd.js`?

jayphelps:
Can we change the `BREAKING CHANGE: AMD modules are no longer built and supported` message to be more specific? It currently sounds like AMD is no longer supported at all, but in fact most commonly people are likely consuming the UMD module which has AMD support. Basically, we're no longer packaging AMD modules for each file, instead only supporting AMD as part of the single UMD build.

README.md
jayphelps: Actually, I believe this should be changed to `5.0.0-beta.12 or higher` need to point to `Rx.js` cause the previous section includes beta.11 (the current release) and correctly points to Rx.umd.js

kwonoj: yes, that seems more correct. Maybe just align with `installation.md`, saying `After beta.11`?

jayphelps:
- [ ] Make breaking change message not sound like there's zero support for AMD
- [ ] Correct/wordsmith the umd link for 5.0.0-beta.12 or higher in README.md

Otherwise looks good.

blesh:
@jayphelps updates made.

kwonoj:
change looks good to me.

jayphelps:
:shipit:

PR #1875, opened 2016-08-10
fix(ajaxObservable): remove implicit dependency to map operator patch

### LINKED ISSUE #1874: Probable regression in ajax.getJSON after updating to 5.0.0-beta.11
**RxJS version:**
5.0.0-beta.11

**Code to reproduce:**
```javascript
import { ajax } from 'rxjs/observable/dom/ajax';

console.log(ajax.getJSON('/api/v2/users'));
```

**Expected behavior:**
`AjaxObservable { _isScalar: false, request: Object }`

**Actual behavior:**
`Uncaught (in promise) TypeError: (intermediate value).map is not a function(… ) in ajaxGetJSON`

**Additional information:**
This is probably a regression caused by this commit: https://github.com/ReactiveX/rxjs/commit/7a77437205a18905b0c2974fedf0ee4e8c624bd6 
Introduced in 5.0.0-beta.11, since it works fine if I downgrade to 5.0.0-beta.10.

I'm using rxjs with redux-observable, if that makes any difference.

kwonoj:
**Description:**
This PR updates behavior of `ajax.getJSON`, internally relies on observable to have `map` operator patched to select response, which causes failure if consumer imports observable without operator patched. Instead, explicitly imports map operator.

closes #1874

### DIFF
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -440,39 +440,66 @@ describe('Observable.ajax', () => {
       expect(complete).to.be.true;
     });
 
-    describe('ajax.post', () => {
-      it('should succeed on 200', () => {
-        const expected = { foo: 'bar', hi: 'there you' };
-        let result: Rx.AjaxResponse;
-        let complete = false;
+    it('should able to select json response via getJSON', () => {
+      const expected = { foo: 'bar' };
+      let result;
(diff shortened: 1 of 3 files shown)


blesh:
LGTM

kwonoj:
@blesh Thanks for review! Merged changes by myself.

PR #1877, opened 2016-08-12
Update MIGRATION.md

gilesbradshaw:
just documentation - I was looking for throw

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -111,6 +111,7 @@ enabling "composite" subscription behavior.
 |`tapOnNext`|`-`|
 |`tapOnError`|`-`|
 |`tapOnCompleted`|`-`|
+|`throw`|`-`|
 |`timestamp`|`-`|
 |`toMap`|`-`|
 |`toSet`|`-`|


blesh:
`Observable.throw` still exists. Is there something else you were looking for?

jayphelps:
@blesh he was looking for it in the migration guide and it wasn't listed. Other operators that were not renamed are listed there so it seems this is a correct addition. Will let you decide, in case I'm thinking incorrectly.

blesh:
@jayphelps @gilesbradshaw ... but I guess I'm saying there's nothing to migrate. `Observable.throw` exists in both Rx 4 and Rx 5.

jayphelps:
@blesh OH I misunderstood the full context of the table. I thought it was an exhaustive list of all operators and that `-` meant it was implemented in v5 without being renamed. Instead, that means the operator _does not exist in v5_.

Because `Observable.throw` was indeed included in v5, it does not belong in the table.

@gilesbradshaw if you're importing operators individually, it's available in two ways:

```js
import 'rxjs/add/observable/throw';

Observable.throw(new Error('message'));
```

Or also as:

```js
import { _throw } from 'rxjs/observable/throw';

// underscore _throw!
_throw(new Error('message'));
```
Notice that in the second example the import is underscored because `throw` without underscore is a JavaScript reserved word for throwing exceptions `throw new Error('example');>`;

kwonoj:
just adding bit more context, I wanted to change `-` into `WIP` or `not implemented` or similar for bit more clarity. Just haven't looked those docs recently.

jayphelps:
@kwonoj that seems like a positive change indeed.

@gilesbradshaw we appreciate you taking the time to PR! Hope this clears up the confusion.

gilesbradshaw:
No worries - my misunderstanding :)

PR #1883, opened 2016-08-16
feat(sequenceEqual): adds sequenceEqual operator

### LINKED ISSUE #1882: sequenceEqual operator
What is the suggested means for checking equality between two observable streams with the removal of the _sequenceEqual_ operator in RxJS 5?   I am trying to make a comparison between a buffered or windowed list of integer values and a known array of values, in the context of a mergeMap.

blesh:
- [x] add necessary tests
- [x] documentation
- [x] perf tests

resolves #1882

### DIFF
--- a/src/operator/sequenceEqual.ts
+++ b/src/operator/sequenceEqual.ts
@@ -5,8 +5,6 @@ import {Subscriber} from '../Subscriber';
 import {tryCatch} from '../util/tryCatch';
 import {errorObject} from '../util/errorObject';
 
-export const defaultComparor = <T>(a: T, b: T) => a === b;
-
 export interface SequenceEqualSignature<T> {
   (compareTo: Observable<T>, comparor?: (a: T, b: T) => boolean): Observable<boolean>;
 }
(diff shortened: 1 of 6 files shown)


src/operator/sequenceEqual.ts
```diff
@@ -0,0 +1,136 @@
...
+      let a = _a.shift();
+      let b = _b.shift();
+      let areEqual = false;
+      if (comparor) {
+        areEqual = tryCatch(comparor)(a, b);
+        if (areEqual === errorObject) {
+          this.destination.error(errorObject.e);
+        }
+      } else {
+        areEqual = a === b;
```
kwonoj: what about making subscriber's ctor provides default value of (`defaultComparor`) instead of line 14, then remove this branch to check comparor exists? it'll makes comparison in `checkValues` and `_complete` lot similar, may can be extracted into common function.

blesh: conditionals are cheaper than function calls. I originally had what you suggested, but optimized.

kwonoj: Got it. please allow me one following question, per [line 66](https://github.com/ReactiveX/rxjs/pull/1883/files#diff-c404a29a686ef6f8878b3085e783ddafR66) it always supplies `comparor` to default value, so usual consumer directly uses operator won't hit this cases unless explicitly clears it like `sequenceEqual(other, null)` - is that correct?

blesh: Actually, in my most recent commit, I've removed the value checking from _complete entirely because it wasn't necessary, and also removed the defaultComparor.

kwonoj: ah, I can see it now. Thanks for clarification :+1:

kwonoj:
Change looks good to me, :shipit:

Issue #1876, opened 2016-08-11
Extending Observable - static `of` method


blesh:
While talking with @jayphelps about [ActionsObservable](https://github.com/redux-observable/redux-observable/blob/f3d5975db1fbf61edd4a47c14a2dac404474ca1b/src/ActionsObservable.js) from redux-observable/redux-observable, it's come up once or twice that someone would need to create an `ActionsObservable` from an array or set of values... 

Currently, this means you have to manually override `static of` in the ActionsObservable if you want to get an `ActionsObservable` as a result.

I think we could possibly change the static of method to do something like this:

```ts
static of(...values) {
  const observable = new this.constructor();
  const scheduler = undefined;
  if (values[values.length - 1] instanceof Scheduler) {
    scheduler = values.pop();
  }
  observable.source = new ArrayObservable(values, scheduler);
  return observable;
}
```

This would mean that extended classes like `ActionsObservable` would be able to benefit from `of` without needing to override it.

### Risks

1. I'm not sure what this would mean for things like Subjects yet. I'd have to dig in, but I can't imagine why someone would create `Subject` with `Subject.of`.
2. I suspect TypeScript will _hate_ this... because there's not really a great way to determine the returned type. We do know it should always at least be an Observable, though.
3. I have only ran into one use case for this so far, and that's ActionsObservable.

### Workaround

The current workaround if you needed this sort of thing is to simply implement `static of` yourself, (which is basically the same as above):

```ts
static of(...values) {
  const observable = new MyObservableType();
  const scheduler = undefined;
  if (values[values.length - 1] instanceof Scheduler) {
    scheduler = values.pop();
  }
  observable.source = new ArrayObservable(values, scheduler);
  return observable;
}
```

It's perhaps just a little gross to have to do this?

jayphelps:
@blesh I think you actually mean `new this()`

```js
static of(...values) {
  const observable = new this();
  const scheduler = undefined;
  if (values[values.length - 1] instanceof Scheduler) {
    scheduler = values.pop();
  }
  observable.source = new ArrayObservable(values, scheduler);
  return observable;
}
```

[...]

PR #1894, opened 2016-08-22
fix(Subject): do not expose static create method to inherited

### LINKED ISSUE #1890: BehaviorSubject created with "create" does not have getValue() method
**RxJS version:**   5.0.0-beta.9

**Code to reproduce:**

`
var a = new Rx.BehaviorSubject('start');

var b = Rx.BehaviorSubject.create(a, new Rx.Observable(o => {
     a.subscribe(o);
}));

console.log(b.getValue());

`

**Expected behavior:**
// output:
start

**Actual behavior:**

Uncaught TypeError: getValue is not a function

**Additional information:**

kwonoj:
**Description:**
This PR aligns interface of classes inherited from `Subject<T>`, do not expose static creation method by introducing abstract class `SubjectBase<T>`.

closes #1890

### DIFF
--- a/spec/subjects/AsyncSubject-spec.ts
+++ b/spec/subjects/AsyncSubject-spec.ts
@@ -1,5 +1,6 @@
 import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
+import {SubjectBase} from '../../dist/cjs/Subject';
 
 const AsyncSubject = Rx.AsyncSubject;
 
@@ -21,6 +22,11 @@ class TestObserver implements Rx.Observer<number> {
 
 /** @test {AsyncSubject} */
 describe('AsyncSubject', () => {
+  it('should extend SubjectBase', () => {
+    const subject = new AsyncSubject();
+    expect(subject).to.instanceOf(SubjectBase);
+  });
+
   it('should emit the last value when complete', () => {
     const subject = new AsyncSubject();
     const observer = new TestObserver();
--- a/spec/subjects/BehaviorSubject-spec.ts
+++ b/spec/subjects/BehaviorSubject-spec.ts
@@ -1,5 +1,6 @@
 import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
+import {SubjectBase} from '../../dist/cjs/Subject';
 declare const {hot, expectObservable};
 
 const BehaviorSubject = Rx.BehaviorSubject;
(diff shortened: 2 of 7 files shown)


kwonoj:
I'm feeling this might not be correct understanding of removing static interfaces or either incorrect implementation of having abstract classes - please feel freely close if PR is not feasible.

kwonoj:
..Or instead static creation method to comply inherited classes (similar to https://github.com/ReactiveX/rxjs/issues/1876)

jayphelps:
As discussed in #1890, there may be merit to adding support for `create` to other subject types, but in the interest of initial simplicity and correctness I think this is the right approach for now. This code looks good, but I'm worried that this is in fact a breaking change.

`(new AsyncSubject) instanceof Subject` will no longer pass, and that scares me and seems counter-intuitive. I'm probably being alarmist though..but I'm going to let others chime in here please.

kwonoj:
> (new AsyncSubject) instanceof Subject will no longer pass, and that scares me and seems counter-intuitive.

: I think that's legit point that I couldn't think of at the moment writing PR. do we have general usecases for checking types of inherited subjects?

jayphelps:
Not really for `instanceof`..but I wouldn't be surprised to see it. More commonly I imagine this would bite people with type systems:

```ts
class ActionsObservable extends Observable<Action> {
  constructor(source$: Subject<Action>) {
    this.source = source;
  }
}

const source$ = new ReplaySubject<Action>();
const action$ = new ActionsObservable(source$);
// type error: ReplaySubject does not match type Subject
```

kwonoj:
gotcha, you're correct. This PR makes breaking change to consumers should accept `SubjectBase` as base class instead of explicit `Subject`. Seems change's more broad than I originally expected..

kwonoj:
I tried small snippet [here](https://www.typescriptlang.org/play/#src=abstract%20class%20SubjectBase%20%7B%0D%0A%09foo()%3A%20void%20%7B%0D%0A%09%09%0D%0A%09%7D%0D%0A%7D%0D%0A%0D%0Aclass%20AsyncSubject%20extends%20SubjectBase%20%7B%0D%0A%09bar()%3A%20void%20%7B%0D%0A%09%09%0D%0A%09%7D%0D%0A%7D%0D%0A%0D%0Aclass%20Subject%20extends%20SubjectBase%20%7B%0D%0A%09static%20a()%20%7B%0D%0A%09%09console.log('meh')%3B%0D%0A%09%7D%0D%0A%7D%0D%0A%0D%0Afunction%20AAA(value%3A%20Subject)%20%7B%0D%0A%09console.log('bbb')%3B%0D%0A%7D%0D%0A%0D%0AAAA(new%20AsyncSubject())%3B) and at least as long as `subject` does not have it's own instance method type inference will work between `subjectbase` inherited vs `subject` itself. But as you've mentioned, I agree this isn't only cases hitting consumers and there might be more cases as well.

My current feeling is rather document `Subject.create` is only work for `Subject` itself for now until fully resolve static creating method inheritance design and close this PR without merging. what do you think?

kwonoj:
closing PR as intended design is to support `create` method for all subject implementation.

Issue #1863, opened 2016-08-03
WebSocketSubject does not reconnect with beta10 (change behaviour from beta7)


sebald:
**RxJS version:** 5.0.0-beta10

**Code to reproduce:**

```ts
    const RETRY_DELAY = 200;
    const openObserver:NextObserver<Event> = { next: () => console.log('connected') };
    const closeObserver:NextObserver<CloseEvent> = { next: () => console.log('disconnected') };

       const ws$ = new WebSocketSubject<APIMessage>({
            url,
            openObserver,
            closeObserver
        });

        const actions$ = ws$
            .retryWhen(errors => errors.mergeMap(error => {
                if (window.navigator.onLine) {
                    console.warn(`Retrying in ${RETRY_DELAY}ms.`);
                    return Observable.timer(RETRY_DELAY);
                } else {
                    return Observable.fromEvent(window, 'online').take(1);
                }
            }))
            .map(msg => msg /** usually maps APIMessage to ReduxMessages */ );
```

**Expected behavior:**
When the connected WS server is shut down, constantly push `errors` to the `retryWhen` until no errors are thrown anymore (=> server is back online again).

**Actual behavior:**
`retryWhen` is called and the timer observable is constantly repeated, but the `WebSocketSubject` seems not to recognise that the server is back online.

**Additional information:**
I wanted to update the dependencies of our app we wrote a few months back. We used beta7 before. Updating *rxjs* to beta8-10 will "break" the reconnecting feature we implemented with the above code .

It seems that with the rewrite of `Subject` in beta8 the behaviour completely changed. Bevor the console did show the following:

![beta8](https://cloud.githubusercontent.com/assets/985701/17356042/4c73b18e-5955-11e6-90fb-167a565b1384.png)

With beta10 the same code will behave like this:

![beta10](https://cloud.githubusercontent.com/assets/985701/17356050/5a4cffc2-5955-11e6-8693-747ca94d99ea.png)

The beta10 is (from my point a view) a lot nicer because it doesn't trigger the `closeObserver` multiple times and there is no error anymore :) But it also does not recognise when the WebSocket connection is availble again.

I tried to debug it myself but couldn't find a the issue.

EDIT: I added the fix applied in 848a527ed2c94ec017afa6395b286494c4454147 but didn't solve my problem.

blesh:
Thanks for reporting this. We'll have a look

deontologician:
[...]

PR #1899, opened 2016-08-24
feature(WebSocketSubject) unsubMsg undefined

deontologician:
**Description:**

When reconnecting a WebSocket, it's sometimes the case that a bunch of unsubscription messages are queued up before the socket was closed. Right now, they'll be unconditionally sent when the socket is reconnected. It's useful to be able to prevent sending them if they're undefined (which will just be serialized as the json string "undefined" anyway, unlikely to be what was expected)

If this makes sense, I'll add some tests

### DIFF
--- a/src/observable/dom/WebSocketSubject.ts
+++ b/src/observable/dom/WebSocketSubject.ts
@@ -104,7 +104,7 @@ export class WebSocketSubject<T> extends AnonymousSubject<T> {
         const result = tryCatch(unsubMsg)();
         if (result === errorObject) {
           observer.error(errorObject.e);
-        } else {
+        } else if (result !== undefined) {
           self.next(result);
         }
         subscription.unsubscribe();


blesh:
For clarity, were talking about multiplexing, right? If the unsubscribe message factory returns null or undefined then we don't send the message at all?

Although I'm not sure why anyone would have a multiplexed socket server that didn't require sending an unsubscribe message. Requires thought.

Do you have use cases?

deontologician:
My use case currently is after a disconnect from the server. The server drops all resources for multiplexed subscriptions and when we reconnect, the client sends a bunch of unsubscribe messages. The problem for me is that we have a handshake, so those queued unsubscriptions come flooding in and are recognized as invalid handshakes, which causes the connection to close again.

Ideally, I'd like to be able to just turn the unsubscriptions into no-ops if there is a disconnect.

deontologician:
Closing since I think this is better solved by fixing #1863

PR #1903, opened 2016-08-24
docs(MIGRATION): update operator table

kwonoj:
**Description:**
from https://github.com/ReactiveX/rxjs/pull/1877#issuecomment-239577722, this PR updates operator migration table to have better explanation of not yet implemented operators. For now everything is marked as `not yet implemented`, while some operator may need clarification like `will not implement`, etcs. 

also, additionally updated few operator status
- `sequenceEqual` removed as it's implemented
- `tapOn*` now refer to `do`


jayphelps:
LGTM, needs rebase.

PR #1904, opened 2016-08-24
Return null value from JSON.Parse

danmarshall:
JSON.parse('') will throw a syntax error (in any browser). Fix is to return a null value. I saw this when my server returned HTTP 204 No Content.

### DIFF
--- a/spec/helpers/ajax-helper.ts
+++ b/spec/helpers/ajax-helper.ts
@@ -117,7 +117,7 @@ export class MockXMLHttpRequest {
 
   private previousRequest: MockXMLHttpRequest;
 
-  private responseType: string = '';
+  protected responseType: string = '';
   private eventHandlers: Array<any> = [];
   private readyState: number = 0;
 
@@ -125,9 +125,9 @@ export class MockXMLHttpRequest {
   private password: any;
 
   private responseHeaders: any;
-  private status: any;
-  private responseText: string;
-  private response: any;
+  protected status: any;
+  protected responseText: string;
+  protected response: any;
 
   url: any;
   method: any;
@@ -176,6 +176,18 @@ export class MockXMLHttpRequest {
     this.triggerEvent('error');
   }
 
+  protected jsonResponseValue(response: any) {
+    try {
+      this.response = JSON.parse(response.responseText);
+    } catch (err) {
+      throw new Error('unable to JSON.parse: \n' + response.responseText);
+    }
+  }
+
+  protected defaultResponseValue() {
+    throw new Error('unhandled type "' + this.responseType + '"');
+  }
+
   respondWith(response: any): void {
     this.readyState = 4;
     this.responseHeaders = {
@@ -186,17 +198,13 @@ export class MockXMLHttpRequest {
     if (!('response' in response)) {
       switch (this.responseType) {
       case 'json':
-        try {
-          this.response = JSON.parse(response.responseText);
-        } catch (err) {
-          throw new Error('unable to JSON.parse: \n' + response.responseText);
-        }
+        this.jsonResponseValue(response);
         break;
       case 'text':
         this.response = response.responseText;
         break;
       default:
-        throw new Error('unhandled type "' + this.responseType + '"');
+        this.defaultResponseValue();
       }
     }
     // TODO: pass better event to onload.
@@ -218,4 +226,30 @@ export class MockXMLHttpRequest {
       }
     });
   }
+}
+
+export class MockXMLHttpRequestInternetExplorer extends MockXMLHttpRequest {
+
+  private mockHttp204() {
+    this.responseType = '';
+    this.responseText = '';
+    this.response = '';
+  }
+
+  protected jsonResponseValue(response: any) {
+    if (this.status == 204) {
+      this.mockHttp204();
+      return;
+    }
+    return super.jsonResponseValue(response);
+  }
+
+  protected defaultResponseValue() {
+    if (this.status == 204) {
+      this.mockHttp204();
+      return;
+    }
+    return super.defaultResponseValue();
+  }
+
 }
--- a/src/observable/dom/AjaxObservable.ts
+++ b/src/observable/dom/AjaxObservable.ts
@@ -396,9 +396,9 @@ export class AjaxResponse {
       case 'json':
         if ('response' in xhr) {
           //IE does not support json as responseType, parse it internally
-          this.response = xhr.responseType ? xhr.response : JSON.parse(xhr.response || xhr.responseText || '');
+          this.response = xhr.responseType ? xhr.response : JSON.parse(xhr.response || xhr.responseText || 'null');
         } else {
-          this.response = JSON.parse(xhr.responseText || '');
+          this.response = JSON.parse(xhr.responseText || 'null');
         }
         break;
       case 'xml':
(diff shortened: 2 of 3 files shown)


kwonoj:
Would you add test cases for this change as well?

danmarshall:
Where do you add your tests?

kwonoj:
I believe [`ajax-spec`](https://github.com/ReactiveX/rxjs/blob/80b817f41084440312954efa3bb0eef4947c8d8b/spec/observables/dom/ajax-spec.ts) is place to go.

danmarshall:
Hi @kwonoj - I cloned locally and tried doing a build, to get myself familiar with the test runner. At some point it failed when it tried to run java (which I don't have) so I think the level of setup is greater than the code contribution... what do you think?

kwonoj:
Jre is only needed to generate global build so you can safely ignore if you do not like to install jre. Just build / test would be sufficient for general workflow such as `npm run build_test`. You may able to check `npm run info` for other tasks available.

danmarshall:
@kwonoj do you have a test for your original fix that I can use as an example?

kwonoj:
No, I just noticed I skipped test cases and just verified manually. Seems that's reason I wasn't able to caught issue, test didn't cover it.

danmarshall:
I verified that this unit test fails on the master branch without the fix, passes on this branch with the fix.

danmarshall:
Hi @kwonoj would you mind taking a look at this? I spent a good deal of time yesterday writing the unit test, wanted to make sure that my additions are compatible with your framework style standards.

kwonoj:
Sure, please allow me some time as I'll be offline for some time. I've brifely looked changes, interested if it's possible to remove mock for IE only - but that'll need some trial on my end.

danmarshall:
Mocking {x} is the way to test {x}, so mocking IE would be the way to test IE :)

src/observable/dom/AjaxObservable.ts
```diff
@@ -396,7 +396,7 @@ export class AjaxResponse {
       case 'json':
         if ('response' in xhr) {
           //IE does not support json as responseType, parse it internally
-          this.response = xhr.responseType ? xhr.response : JSON.parse(xhr.response || xhr.responseText || '');
+          this.response = xhr.responseType ? xhr.response : JSON.parse(xhr.response || xhr.responseText || 'null');
         } else {
           this.response = JSON.parse(xhr.responseText || '');
```
kwonoj: isn't this `JSON.parse(xhr.responseText || '');` also need to be updated to `'null`' for fallback string? this isn't directly related to IE, but still seems possible unexpected failure cases.

spec/helpers/ajax-helper.ts
```diff
@@ -218,4 +226,33 @@ export class MockXMLHttpRequest {
...
+
+  protected jsonResponseValue(response: any) {
+    if (this.status == 204) {
+      this.mockHttp204();
+      return;
+    }
+    return super.jsonResponseValue(response);
+  }
+
+  protected defaultResponseValue() {
```
kwonoj: doesn't `jsonResponseValue` handles status 204? is there possibility to fall back into this overrided method?

src/observable/dom/AjaxObservable.ts
danmarshall: Do you mean on line 401 ? I did not exercise that code path, but I suppose that yes it should.

kwonoj: yes, 401. I know it's not directly related (for execution path to IE), but seems need to be updated as well.

spec/helpers/ajax-helper.ts
danmarshall: On other platforms, yes. This is the override to behave as IE

kwonoj: I just pulled down and while trying it, seems this specific (`MockXMLHttpRequestInternetExplorer::defaultResponseValue()`) hasn't executed. Maybe I'm missing something?

danmarshall: There is a scenario where one can write a unit test like this:
```
request.respondWith({
  'status': 204,
  'contentType': 'application/json',
  'responseType': 'json',
  'responseText': expected
});
```

The issue is that the IE does not actually provide ```  'responseType': 'json', ``` so this would be a presumptuous test. That is why when we mock IE that we actually unset ```responseType``` in ```private mockHttp204()```

That is the reason the MockXMLHttpRequestInternetExplorer overrides both, just in case.

kwonoj: > That is why when we mock IE that we actually unset responseType in private mockHttp204()

: yes, that I'm aware of.

My question in here is, if overrided `defaultResponseValue` doesn't do any roles in inherited `MockXMLHttpRequestIE`, is it necessarily need to do it? test for behavior of IE is fulfilled by response override at line 242, so default case may not be separated / overrideded but just leave as-is to throw directly in line 199 instead. Of course it's for 'just in case' and to be possibly served in further though, I'm seeing inherited class can be simplified.

spec/helpers/ajax-helper.ts
```diff
@@ -218,4 +226,33 @@ export class MockXMLHttpRequest {
       }
     });
   }
-}
\ No newline at end of file
+}
+
+export class MockXMLHttpRequestInternetExplorer extends MockXMLHttpRequest {
+  constructor() {
```
kwonoj: you can skip ctor if it doesn't do specific thing in inherited.

spec/helpers/ajax-helper.ts
danmarshall: Well, if it is used incorrectly, (test respondWith ```'responseType': 'json',```) this will give a false positive. Basically, my intention was to keep all IE mock behavior inside of the class. Otherwise every test writer has to know how IE behaves.

kwonoj: Got it, thanks for clarification. I don't see it as hard blocking issue.

kwonoj:
LGTM, I'll leave this PR around couple of days to see if other suggestions around before check in.

jayphelps:
LGTM, needs rebase plz and while you're add it please squash as a single commit using our [commit message format](https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md#commit-message-format) (see existing commits for reference)

For others reference, the behavior of having it be `xhr.response = null` is per spec when `responseType === 'json'`

https://xhr.spec.whatwg.org/#json-response

danmarshall:
Hmm I'm not comfortable with my local result when I try squashing. Looks like I will be squashing a few other folks commits in with mine, and I don't think that's what we want. I used the button on this page to do the merges from master, but it doesn't look like those were rebased. Do you know if I can safely just omit other people's commits during the interactive rebase?

danmarshall:
Sorry, just had to set upstream and fetch it. Nevermind :)

jayphelps:
@danmarshall I went ahead and squashed it. Check out http://push.cwcon.org/learn/squashing-commits.html for further info about how to squash your commits into a one. 💃 

Thanks for your fix!

danmarshall:
Awesome! When is your next publish to NPM ?

jayphelps:
@danmarshall the next beta release isn't set. There are usually a release at least once a month, so we're due for one shortly I imagine. It's up to @blesh

PR #1906, opened 2016-08-25
added ajax tests for 204 No Content

danmarshall:
**Description:**

### DIFF
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -440,6 +440,33 @@ describe('Observable.ajax', () => {
       expect(complete).to.be.true;
     });
 
+    it('should succeed on 204 No Content', () => {
+      const expected = null;
+      let result;
+      let complete = false;
+
+      Rx.Observable
+        .ajax.get('/flibbertyJibbet')
+        .subscribe(x => {
+          result = x.response;
+        }, null, () => {
+          complete = true;
+        });
+
+      const request = MockXMLHttpRequest.mostRecent;
+
+      expect(request.url).to.equal('/flibbertyJibbet');
+
+      request.respondWith({
+        'status': 204,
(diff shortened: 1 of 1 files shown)


danmarshall:
Hi @kwonoj I didn't realize this was going to make a 2nd PR :(

kwonoj:
Yes, please update original PR to include this changes.

One question - isn't this test supposed to 'fail' with current master's codebase?

danmarshall:
I suppose it would only fail if it simulates IE ?

kwonoj:
Yes, I thought test case should mock response to check beahvior of `json.parse`.

danmarshall:
I see. Well I am not sure how to handle this, as MockXMLHttpRequest behaves as Chrome / standards. Seems like there would need to be a MockXMLHttpRequest_IE ?

kwonoj:
It's rough idea that I haven't tried yet, but I might give a chance of mocking via `sinon`.

danmarshall:
So your original fix was handling IE which does not provide ```xhr.response``` or ```xhr.responseText```. In this unit test, we tell the ```MockXMLHttpRequest``` to respond:

```      request.respondWith({
        'status': 204,
        'contentType': 'application/json',
        'responseText': expected
      });
```
To simulate the IE behavior, we would just not respondWith responseText, like this:

```      request.respondWith({
        'status': 204,
        'contentType': 'application/json'
      });
```
I don't know if I can call ```request.respondWith()``` twice within the same unit test, do you? Otherwise there will be two tests.

kwonoj:
it is not prevented to use it multiple times, but as `respondWith` will update most recent mock response I'd go with separating test cases. You can see some of test cases are testing specific conditions like `'should try to create AXObject for XHR in old version of IE'`, assume could be applied for this case as well.

PR #1911, opened 2016-08-30
feat(operator): Add repeatWhen operator

DomiR:
**Description:**
It still has four test from retryWhen, that are disabled as they don't make sense for repeatWhen. Maybe there are some alternative tests that should be included instead. 

### DIFF
--- a/doc/decision-tree-widget/tree.yml
+++ b/doc/decision-tree-widget/tree.yml
@@ -286,7 +286,12 @@ children:
       children:
       - label: I want to re-subscribe
         children:
-        - label: repeat
+          - label: immediately
+            children:
+            - label: repeat
+          - label: when another Observable emits
+            children:
+            - label: repeatWhen
       - label: I want to start a new Observable
         children:
         - label: concat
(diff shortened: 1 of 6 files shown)


doc/decision-tree-widget/tree.yml
```diff
@@ -287,6 +287,7 @@ children:
       - label: I want to re-subscribe
         children:
         - label: repeat
+        - label: repeatWhen
```
staltz: To clarify how this file works, you have to add a description label before adding `children:` > `- label: repeatWhen` as the child of the description label. Just check how all other operators are formatted in this file. Not `repeatWhen` right under `repeat`.

DomiR: Thanks, fixed

src/operator/repeatWhen.ts
```diff
@@ -0,0 +1,126 @@
...
+    return source._subscribe(new RetryWhenSubscriber(subscriber, this.notifier, this.source));
+  }
+}
+
+/**
+ * We need this JSDoc comment for affecting ESDoc.
+ * @ignore
+ * @extends {Ignored}
+ */
+class RetryWhenSubscriber<T, R> extends OuterSubscriber<T, R> {
```
trxcllnt: `RetryWhenSubscriber` -> `RepeatWhenSubscriber`

DomiR: Thanks, done.

jayphelps:
LGTM, needs rebase.

PR #1916, opened 2016-09-01
chore(typings): update type definition for concat

### LINKED ISSUE #1915: Bad typings for concat
Version 5.0.0-beta.11

I'm having trouble with some of the provided typings. For example:

``` js
const x = Observable.of(1) // Observable<number>
const y = x.concat(x) // Observable<any>
```

I expected `y` to be `Observable<number>`. I tried to debug the typings but couldn't find where the generic was being lost. It seems to be related to `ObservableInput`, because if I change `concat`’s parameter typing from that to `Observable`, it works as expected.

Same problem with `Observable.merge(Observable.of(1))`

kwonoj:
**Description:**
This PR updates definition of `concat` to accept source observable type, conform `ConcatSignature`'s overload.

closes #1915

### DIFF
--- a/src/operator/concat.ts
+++ b/src/operator/concat.ts
@@ -44,7 +44,7 @@ import { MergeAllOperator } from './mergeAll';
  * @method concat
  * @owner Observable
  */
-export function concat<T, R>(...observables: Array<ObservableInput<any> | Scheduler>): Observable<R> {
+export function concat<T, R>(...observables: Array<ObservableInput<T | any> | Scheduler>): Observable<R> {
   return concatStatic<T, R>(this, ...observables);
 }
 
@@ -112,7 +112,7 @@ export function concatStatic<T, R>(...observables: (ObservableInput<any> | Sched
  * @name concat
  * @owner Observable
  */
-export function concatStatic<T, R>(...observables: Array<ObservableInput<any> | Scheduler>): Observable<R> {
+export function concatStatic<T, R>(...observables: Array<ObservableInput<T | any> | Scheduler>): Observable<R> {
   let scheduler: Scheduler = null;
   let args = <any[]>observables;
   if (isScheduler(args[observables.length - 1])) {


jayphelps:
@kwonoj I'm a bit confused.

> Same problem with Observable.merge(Observable.of(1))

I see the same thing with nearly all operators.

https://github.com/ReactiveX/rxjs/blob/master/src/operator/merge.ts#L53
https://github.com/ReactiveX/rxjs/blob/7b23e88db21302a10c2a163952b8da197d18cbfe/src/operator/combineLatest.ts#L49

etc.

Is this a problem across the board? Why did our tests (and the community) not catch this prior?

david-driscoll:
@jayphelps because we don't typically expose these methods to TypeScript users, since TypeScript still lacks `::`.  So we get the signature definition.

Also this would indeed, apply to merge / combineLatest and potentially others.

I'd be more inclined to focus on removing the Signature interfaces once we can safely transition to TS2.0 and merge the signatures into the methods with new ability to define the type for `this`.

jayphelps:
@david-driscoll I'll come back with a fresh head, cause I'm still confused since the issue described usage:

```ts
const x = Observable.of(1) // Observable<number>
const y = x.concat(x) // Observable<any>
```

Which isn't using the `::` bind operator

david-driscoll:
Commented on the other ticket, the issue is related to the extra catch overload (#1880).

As unlikely as it sounds it's related directly to a single unrelated overload and breaks type inference for every single merge operator.  I presume this is because it's actually conflicting with `Promise.catch` because if you `mergeMap(x => x.toPromise())` type inference works correctly.

kwonoj:
@jayphelps I wrote this as for kind of immediate amending, which occurs only when installed as module at repo (not inside as codebase). Try setup some test repo with failing example. For me also this is bit confusing, wanted to bring up as discussion point. Feel freely close this.

@david-driscoll this can occur with TS 1.8 as well, is this also related with issue you mentioned?

david-driscoll:
Yeah, I've seen the issue in 1.8 and 2.0.  Removing the extra overload solves the issue and type inference works again.

When I originally sat down with the problem I narrowed the problem down to the specific commit using git bisect, and tested various difference changes before I discovered the problem was that single overload.

kwonoj:
Gotcha. Makes me feel better to close this without merge and pursuit legit changes. Thughts?

david-driscoll:
I would agree.

kwonoj:
OK, closing without check it. @jayphelps , sorry for causing confusion to you :)

Issue #1892, opened 2016-08-22
Min(comparer) emits comparer result value instead of original stream value


gluck:
**RxJS version:**
5.0.0-beta.11
**Code to reproduce:**
`Observable.range(1,2).min((a,b) => a > b ? -1 : a < b ? 1 : 0)`
**Expected behavior:**
`next(2)` (reverse comparer)
**Actual behavior:**
`next(1 | -1)` (the comparer result is emitted instead)
**Additional information:**
existing rxjs 4 unit test showing expected behavior: [min with comparer non-empty](https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/min.js#L126)
existing rxjs 5 unit test showing the wrong behavior: [should handle a constant predicate on observable with many values](https://github.com/ReactiveX/rxjs/blob/master/spec/operators/min-spec.ts#L195) - 42 is emitted instead of some value from original observable.

I could work on a PR, but checking first that it's not an horrible mistake on my side, or an on-purpose change.

Thx.

gluck:
I'll work on one after vacation (2 weeks from now).

PR #1923, opened 2016-09-06
fix(min/max): do not return comparer values (#1892)

gluck:
**Description:**

min (&max) operators have now the expected behavior when used with a comparer:
- previously they were behaving like _reduce_ and emitting (last) comparer return value
- they now emit original observable min/max value by comparing the comparer return value to 0

method signatures were changed to have comparer return a number instead of T
unit tests have been updated, the ones relying on a constant comparer have been removed when their outcome should be undefined (more than 1 value in stream and no error/unsub) and replaced by ones with either a basic or reverse comparer.

I can split the PR in 2 commits (one for min and one for max) if you think that'd help on the release notes.

**Related issue :**

#1892

### DIFF
--- a/spec/operators/max-spec.ts
+++ b/spec/operators/max-spec.ts
@@ -184,26 +184,13 @@ describe('Observable.prototype.max', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
-  it('should handle a constant predicate on observable with many values', () => {
-    const e1 = hot('-x-^-a-b-c-d-e-f-g-|');
-    const e1subs =    '^               !';
-    const expected =  '----------------(w|)';
-
-    const predicate = () => {
-      return 42;
-    };
-
-    expectObservable((<any>e1).max(predicate)).toBe(expected, { w: 42 });
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
-  });
-
-  it('should handle a predicate on observable with many values', () => {
+  it('should handle a reverse predicate on observable with many values', () => {
     const e1 = hot('-a-^-b--c--d-|', { a: 42, b: -1, c: 0, d: 666 });
     const e1subs =    '^         !';
     const expected =  '----------(w|)';
 
     const predicate = function (x, y) {
-      return Math.min(x, y);
+      return x > y ? -1 : 1;
     };
 
     expectObservable((<any>e1).max(predicate)).toBe(expected, { w: -1 });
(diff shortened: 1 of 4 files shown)


kwonoj:
I prefer to split into each commit (personally). In either case, I'd like to suggest to remove issue number from commit message but include in description (https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md#commit-message-format)

spec/operators/max-spec.ts
```diff
@@ -184,42 +184,42 @@ describe('Observable.prototype.max', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
-  it('should handle a constant predicate on observable with many values', () => {
-    const e1 = hot('-x-^-a-b-c-d-e-f-g-|');
-    const e1subs =    '^               !';
-    const expected =  '----------------(w|)';
+  it('should handle a predicate on observable with many values', () => {
```
kwonoj: is this test case needed still? originally it was testing predicate returns constant value (which was incorrect assessment), testing execution of predicate is being covered by line 200 (https://github.com/ReactiveX/rxjs/pull/1923/files#diff-1e06648035d9126f1295caf9df494e1cR200) .

spec/operators/min-spec.ts
```diff
@@ -188,46 +188,46 @@ describe('Observable.prototype.min', () => {
...
-    expectObservable(result, unsub).toBe(expected, { w: 42 });
+    expectObservable(result, unsub).toBe(expected);
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
-  it('should handle a constant predicate on observable with many values', () => {
-    const e1 = hot('-x-^-a-b-c-d-e-f-g-|');
-    const e1subs =    '^               !';
-    const expected =  '----------------(w|)';
+  it('should handle a predicate on observable with many values', () => {
```
kwonoj: same as https://github.com/ReactiveX/rxjs/pull/1923/files#r77709205 .

spec/operators/max-spec.ts
gluck: fair enough, if the reverse predicate works, it's probably enough

gluck:
Done!

src/operator/min.ts
```diff
@@ -12,13 +12,13 @@ import { ReduceOperator } from './reduce';
...
+export function min<T>(comparer?: (x: T, y: T) => number): Observable<T> {
+  const min: (x: T, y: T) => T = (typeof comparer === 'function')
+    ? (x, y) => comparer(x, y) < 0 ? x : y
     : (x, y) => x < y ? x : y;
   return this.lift(new ReduceOperator(min));
 }
 
 export interface MinSignature<T> {
-  (comparer?: (x: T, y: T) => T): Observable<T>;
+  (comparer?: (x: T, y: T) => number): Observable<T>;
```
kwonoj: doesn't `MaxSignature` (https://github.com/ReactiveX/rxjs/pull/1923/files#diff-31007d130dd79eb825771f50f6eef837R24) need to be updated as well?

gluck: Definitely, and done.

kwonoj:
LGTM

PR #1928, opened 2016-09-08
feat(onErrorReturn): Add new Observable.prototype.onErrorReturn operator

jayphelps:
Note: RxJS v4 _does **not**_ have this operator either.

Some Rx implementations contain an `onErrorReturn` operator. http://reactivex.io/documentation/operators/catch.html (scroll down and select it from RxJava list)

> instructs an Observable to emit a particular item when it encounters an error, and then terminate normally

image

> The `onErrorReturn` method returns an Observable that mirrors the behavior of the source Observable, unless that Observable invokes onError in which case, rather than propagating that error to the observer, onErrorReturn will instead emit a specified item and invoke the observer’s onCompleted method. [javadoc](http://reactivex.io/RxJava/javadoc/rx/Observable.html#onErrorReturn(rx.functions.Func1))

This operator is usually similar to `.catch()` but that it expects your provided selector to return a single _value_ instead of an Observable.

In practice this is exactly what people do 99% of the time in redux-observable and _maybe_ in common RxJS usage as well?

#### Using .catch()

```js
// return an Observable
.catch(error => Observable.of({
  type: FETCH_USER_REJECTED,
  payload: error.xhr.response,
  error: true
}))
```
```js
const fetchUserEpic = action$ =>
  action$.ofType(FETCH_USER)
    .mergeMap(action =>
      ajax.getJSON(`/api/users/${action.payload}`)
        .map(fetchUserFulfilled)
        .catch(error => Observable.of({
            type: FETCH_USER_REJECTED,
            payload: error.xhr.response,
            error: true
        }))
    );
```

#### Using .onErrorReturn()
```js
// Return a value
.onErrorReturn(error => ({
  type: FETCH_USER_REJECTED,
  payload: error.xhr.response,
  error: true
})
```
```js
const fetchUserEpic = action$ =>
  action$.ofType(FETCH_USER)
    .mergeMap(action =>
      ajax.getJSON(`/api/users/${action.payload}`)
        .map(fetchUserFulfilled)
        .onErrorReturn(error => ({
            type: FETCH_USER_REJECTED,
            payload: error.xhr.response,
            error: true
        })
    );
```

Or more likely with action creators:

```js
.onErrorReturn(error => fetchUserRejected(error.xhr.response))
```

We originally started to discuss this here https://github.com/redux-observable/redux-observable/issues/102

Cc/ @blesh

### DIFF
--- a/src/operator/onErrorReturn.ts
+++ b/src/operator/onErrorReturn.ts
@@ -1,0 +1,66 @@
+import { Operator } from '../Operator';
+import { Subscriber } from '../Subscriber';
+import { Observable } from '../Observable';
+import { OuterSubscriber } from '../OuterSubscriber';
+
+/**
+ * Instructs an Observable to emit a particular item when it encounters an error, and then terminate normally.
+ * 
+ * You can use this to prevent errors from propagating or to supply fallback data should errors be encountered.
+ * @param {function} selector a function that takes as arguments `err`, which is the error. Whatever value
+ *  is returned by the `selector` will be emitted as the next value, just before terminating.
+ * @return {Observable} an observable that originates from the source
+ * @method onErrorReturn
+ * @owner Observable
+ */
+export function onErrorReturn<T, R>(selector: (err: any) => R): Observable<R> {
+  const operator = new OnErrorReturnOperator<T, R>(selector);
+  return this.lift(operator);
+}
+
+export interface OnErrorReturnSignature<T> {
+  <R>(selector: (err: any) => R): Observable<R>;
+}
+
+class OnErrorReturnOperator<T, R> implements Operator<T, R> {
+  constructor(private selector: (err: any) => R) {
+  }
+
+  call(subscriber: Subscriber<R>, source: any): any {
+    return source._subscribe(new OnErrorReturnSubscriber(subscriber, this.selector));
+  }
+}
+
+/**
+ * We need this JSDoc comment for affecting ESDoc.
+ * @ignore
+ * @extends {Ignored}
+ */
+class OnErrorReturnSubscriber<T, R> extends OuterSubscriber<T, R> {
+  constructor(destination: Subscriber<any>,
+              private selector: (err: any) => R) {
+    super(destination);
+  }
+
+  // NOTE: overriding `error` instead of `_error` because we don't want
+  // to have this flag this subscriber as `isStopped`.
+  error(err: any) {
+    if (!this.isStopped) {
+      const { destination } = this;
+      let result: R;
+
+      try {
+        result = this.selector(err);
+      } catch (err) {
+        this.destination.error(err);
+        return;
+      }
+
+      this.unsubscribe();
+      (<any>destination).remove(this);
+
+      destination.next(result);
+      destination.complete();
+    }
+  }
+}
--- a/spec/operators/onErrorReturn-spec.ts
+++ b/spec/operators/onErrorReturn-spec.ts
@@ -1,0 +1,111 @@
+import { expect } from 'chai';
+import * as Rx from '../../dist/cjs/Rx';
+declare const { hot, cold, asDiagram, expectObservable, expectSubscriptions };
+
+declare const rxTestSchdeuler: Rx.TestScheduler;
+const Observable = Rx.Observable;
+
+/** @test {onErrorReturn} */
+describe('Observable.prototype.onErrorReturn', () => {
+  asDiagram('onErrorReturn')('should catch error and emit single selected value', () => {
+    const e1 =   hot('--a--b--#        ');
+    const expected = '--a--b--(1|)';
+
+    const result = e1.onErrorReturn((err: any) => '1');
+
+    expectObservable(result).toBe(expected);
+  });
+
+  it('should catch error and replace it with Observable.of() that you can .switch()', () => {
+    const e1 =   hot('--a--b--c--------|');
+    const subs =     '^       !';
+    const expected = '--a--b--(XYZ|)';
+
+    const result = e1
+      .map((n: string) => {
+        if (n === 'c') {
+          throw 'bad';
+        }
+        return n;
+      })
+      .onErrorReturn((err: any) => {
+        return Observable.of('X', 'Y', 'Z');
+      })
+      .switch();
+
+    expectObservable(result).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(subs);
+  });
+
+  it('should allow unsubscribing explicitly and early', () => {
+    const e1 =   hot('--1-2-3-4-5-6---#');
+    const e1subs =   '^      !         ';
+    const expected = '--1-2-3-         ';
+    const unsub =    '       !         ';
+
+    const result = e1.onErrorReturn(() => 'x');
+
+    expectObservable(result, unsub).toBe(expected);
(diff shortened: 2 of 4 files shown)


src/operator/onErrorReturn.ts
```diff
@@ -0,0 +1,64 @@
...
+      } catch (err) {
+        this.destination.error(err);
+        return;
+      }
+
+      this.unsubscribe();
+      (<any>this.destination).remove(this);
+
+      this.destination.next(result);
+      this.destination.complete();
```
jayphelps: Is this the correct way to next a value and then complete?

src/operator/onErrorReturn.ts
```diff
@@ -0,0 +1,64 @@
...
+  // NOTE: overriding `error` instead of `_error` because we don't want
+  // to have this flag this subscriber as `isStopped`.
+  error(err: any) {
+    if (!this.isStopped) {
+      let result: R;
+
+      try {
+        result = this.selector(err);
+      } catch (err) {
+        this.destination.error(err);
```
kwonoj: do we have test coverage for this case? coverall says it's not covered. (selector throws error)

jayphelps: Nope, added! thanks! https://github.com/ReactiveX/rxjs/pull/1928/files#diff-a33389611bce8540bfdbcdcdafea23d3R92

kwonoj: Cool 💯 now coverall doesn't complains as coverage decreased.

src/operator/onErrorReturn.ts
kwonoj: Yes, there's similar use cases like [this](https://github.com/ReactiveX/rxjs/blob/7b23e88db21302a10c2a163952b8da197d18cbfe/src/operator/first.ts#L138) at first operator, and other operators as well.

kwonoj: maybe as same as other code does, could do 

```js

destination.xxx....
```

since `destination` is being accessed multiple times.

kwonoj:
Implementation wise this looks ok to me. I think adding this operator's ok too - kind of 'good to have'. Leave opened for other's suggestion around.

BerkeleyTrue:
Has there been discussion around changing the behavior of catch to accept an observable or a value in the selector?

I bring this up because my initial confusion around the catch operator came from the fact that promise.catch accepts a new value that continues the promise chain: http://jsbin.com/siqepiduyo/edit?js,console

This might reduce confusion of new users.

kwonoj:
> Has there been discussion around changing the behavior of catch to accept an observable or a value in the selector?

: Personally I'd vote for separated operators, for reasons of
- avoiding polymorphic behavior as possible
- align with other implementation's signatures

jayphelps:
@BerkeleyTrue if you mean adding this ability:

```js
.catch(() => 'value')
```

This would indeed be inconsistent with all other operators that expect an `ObservableInput` (Observable, Promise, Array, etc)

And if we supported it, it would make things even more confusing IMO when you wanted to emit an Observable _as the value itself._,

```js
.onErrorReturn(() => Observable.of(1, 2, 3))
// later
.switch() // or whatever
```

staltz:
I see why it's useful in redux-observable. But it's rather sad, we're just expanding RxJS's API, and I hear it quite often from beginners that they find the API to be humongous and therefore intimidating. Why make it harder for beginners through decision paralysis? Are a few characters worth it? (even if you're repeating .catch(e => Observable.of()) idiom)

roganov:
Aliasing `Observable` to `O` (which is pretty convenient regardless) as in
`import { Observable as O } from 'rx';`
`catch` + `O.of` becomes even shorter than `onErrorReturn`, so there might be no point in adding a new operator.

jayphelps:
For me (atleast) this isn't about reducing the number of characters. It's making common code easier to read.

In the redux-observable case, many will choose to use action creators:

```js
.catch(error => Observable.of(
  fetchUserRejected(error.xhr.response)
));
```
```js
.onErrorReturn(error => fetchUserRejected(error.xhr.response))
// or if your action creator expects the full error
.onErrorReturn(fetchUserRejected)
```

This is the bread butter for most, so it just _feels_ poor that we always have to do the wrapping.

That said, I think your arguments are valid as long as we decide on general criteria for accepting any new operators and we audit the existing and remove those we feel are similarly not critical (if any).

I understand where people are coming from--but don't share it personally because I look at it like a programming language (cause it sort of is); there are going to be tons of ways to do things you don't know about until you learn them. If you simply look at a giant list of all of them you're going to feel overwhelmed indeed.

staltz:
If you think in terms of the scopes of these two libraries, RxJS and redux-observable, it's easy to justify the existence of onErrorReturn in the context of redux-observable, but hard to justify its existence only considering the scope of RxJS. We need to be able to draw the line so that consumer libraries and frameworks (redux-observable, Angular 2, Cycle.js etc) *use* RxJS instead of writing shortcuts.

One of the reasons why I built xstream is that I noticed I couldn't influence RxJS to make it easier for Cycle.js usage if it would make it harder for other styles of using RxJS. Examples: the debate over reusable or non-reusable Subjects after they terminate, the `mapTo` operator that Ben considered removing once (while at that time I argued for keeping it because it'd make Cycle.js usage better, today I'd argue to remove it), etc.

It's that typical tradeoff of making something general and powerful versus making something specific and pragmatic. If we choose which of these sides RxJS assumes, we'll make it easier for beginners to grok it, to find it reusable and composable, and perhaps they can even independently deduce why some operators are built-in and others don't.

jayphelps:
Agreed in general, but it sounded as if I was somewhat implying that RxJS should accommodate redux-observable use cases at the expense of general RxJS usage--so I want to clarify that that's not my wish. Only using at a perfect example. In my RxJS code unrelated to redux-observable I very often use `catch` the same was as `onErrorReturn`. But if I'm honest, I don't nearly as often do _any_ catching in most of my non-redux-observable code so I can't say if this is typical.

I was hoping others who us RxJS for non-redux-observable uses would share some of there uses of `catch` and see if they too are often just returning an Observable of one. In that case, I don't think it'd be fair to say it's hard to justify--but perhaps still not something we want to do regardless.

blesh:
Technically this should work: `source.catch(err => ['value'])`

It's probably worth debating having operators that accept promises, iterables, observables, etc. accept any value and wrap it in an observable automatically. but I suspect there's problems with that being even more confusing.

trxcllnt:
> @blesh It's probably worth debating having operators that accept promises, iterables, observables, etc. accept any value and wrap it in an observable automatically. but I suspect there's problems with that being even more confusing.

This is what I wished we'd do with `catch`, but not `flatMap` etc. Also, I wish we had a nicer way of reporting which user-provided selector function causes us to throw the `unrecognized type` error in `subscribeToResult`.

jayphelps:
I'd be OK if `catch` allowed single value return (but not other operators). This absolutely would be a deviation of consistency, but one I think is mostly acceptable considering that `catch` is otherwise not really related to the `flatMap`, etc in purpose/behavior.

staltz:
> they would return an array value thinking that value would be emitted as-is but instead it's consumed as a stream.

jayphelps:
No consensus reached, so going to bow out.

PR #1931, opened 2016-09-09
chore(root): remove unused module/exports checking code

jayphelps:
Appears this is left over from a refactor--even has a lint disable to prove it 😆

### DIFF
--- a/src/util/root.ts
+++ b/src/util/root.ts
@@ -20,11 +20,7 @@ declare module NodeJS {
 
 export let root: any = (objectTypes[typeof self] && self) || (objectTypes[typeof window] && window);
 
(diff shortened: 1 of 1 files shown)


kwonoj:
:+1: was curios if it was intended or just leftover.

LGTM

jayphelps:
My best guess about coverage decrease is that unreachable code isn't counted against you as "untested" but _is_ included in the overall code quantity calculations; which means removing it will cause the calculation to shift towards a decrease since code that isn't counted as "uncovered" is inherently counted as covered.

Could be wrong though..just the best guess I can come up with considering this code is in fact completely unused.

kwonoj:
That makes sense to me.

PR #1933, opened 2016-09-09
fix(root): find global context (window/self/global) in a more safe way

### LINKED ISSUE #1930: Idenitifier self is not declared, breaks when `self` its leaked.
The `self` identifier is NOT declared locally in the module and hence it comes from global - in https://github.com/ReactiveX/rxjs/blob/3793e4768dd2655300b67a82be9b30c4c065bac1/src/util/root.ts#L21

So, depending on what `window.self` value might be (i.e from leaked variables, a common mistake of devs), RxJS might break. 

**RxJS version:**
`5.0.0-beta.11`

**Code to reproduce:**
Add a `window.self` variable with different values and run tests.

jayphelps:
Fixes #1930

Unfortunately this isn't something we can really unit test AFAIK. It's pretty meta.

The `objectTypes` stuffs appeared to be a left over from earlier refactors cause none of these should ever be a function.

### DIFF
--- a/spec/util/root-spec.ts
+++ b/spec/util/root-spec.ts
@@ -1,0 +1,9 @@
+import { expect } from 'chai';
+import { root } from '../../dist/cjs/util/root';
+
+/** @test {root} */
+describe('root', () => {
+  it('should exist', () => {
+    expect(typeof root).to.equal('object');
+  });
+});
--- a/src/util/root.ts
+++ b/src/util/root.ts
@@ -1,12 +1,3 @@
-let objectTypes = {
-  'boolean': false,
-  'function': true,
-  'object': true,
-  'number': false,
-  'string': false,
-  'undefined': false
-};
-
 declare let global: NodeJS.Global;
 declare let module: any;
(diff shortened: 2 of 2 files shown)


jayphelps:
ugh...remove code and even add one test and coverage goes down...wtf..

kwonoj:
Seems coveralls complaining about missing coverage of throwing out if `root` is nonexists. :/

https://coveralls.io/builds/7822965/source?filename=src%2Futil%2Froot.ts#L24

jayphelps:
I can remove it....but it seems stupid to do so just to get coveralls from complaining. Thoughts or workarounds?

kwonoj:
For this case I'd rather allow coverage decreasing. Thought way to test those cases bit, but wasn't able to find easy way to create legit coverage.

jayphelps:
@kwonoj planning to merge or waiting for others to chime in?

kwonoj:
@jayphelps sorry, just too busy with new hire orientation all day long. Merged. :)

Issue #1921, opened 2016-09-05
errors do not terminate source unless an observer is provided


colinfang:
```
Observable.interval(1000)
  .flatMap(() => this.http.get('abc'))
  .subscribe(
      // x => console.log('Next: ' + x) // Don't terminite without this.
  );
```

This http request gives an error due to 404. But the stream keeps polling.

However, as soon as I uncomment the commented line in subscribe, it terminates on the first error.

See example here

http://plnkr.co/edit/bQyBvboFJCPyZUEFjraX?p=preview

Why is that?

jayphelps:
Appears this isn't related to `flatMap`, it's more general than that. If you don't provide an observer to `subscribe(observer)` then the source does not terminate correctly.

jayphelps:
Still digging, but here's my minimal example:

http://jsbin.com/yayave/edit?js,output

So far my best guess is it's related to the internal `syncErrorThrowable`, whether it's true or false. Not providing an observer it gets the default `empty` observer which is `syncErrorThrowable: false` but when you provide your own its `syncErrorThrowable: true`. This logic is seemingly for optimizing try/catch.

@blesh this prolly needs your attention since I believe you're the one who knows this part best.

trxcllnt:
@jayphelps I'm responsible for `syncErrorThrowable

jayphelps:
@trxcllnt this is all YOUR FAULT!

trxcllnt:
@jayphelps I may have authored the code, but `syncErrorThrowable` is the product of a debate on error handling semantics that [I lost](https://github.com/tc39/proposal-observable/issues/47), so I'm going to blame @blesh and @zenparsing :-)

trxcllnt:
@jayphelps @colinfang fixed in #1935

PR #1935, opened 2016-09-11
Fixes 1921

trxcllnt:
**Description:**
Supply the Subscriber constructor the empty observer destination if `Observable.subscribe` is called with no arguments. This ensures the Subscriber wraps the empty Observer in a `SafeSubscriber` which will catch `error` messages and dispose of the subscription.

#1921


jayphelps:
👏 excellent!

LGTM

PR #1936, opened 2016-09-12
feat(filter): Observable<T>.filter() can take a type guard as the predicate function

saneyuki:
### Before

By the previous code base, we need to use `filter()` + `map()`

```typescript
interface SuperClass {}
interface DerivedClass extends SuperClass {}

const bar: Observable<SuperClass> = someobservable;
const derived: Observable<DerivedClass> = bar.filter((v): boolean => {
    // some type guard logic
  }).map<DerivedClass>((v) => v as DerivedClass);
```

### After

By this change, we can write this code:

```typescript
const bar: Observable<SuperClass> = someobservable;
const derived: Observable<DerivedClass> = bar.filter<DerivedClass>((v): v is DerivedClass => {
  // some type guard logic
});
```

## Related issue (if exists)

https://github.com/Microsoft/TypeScript/pull/10027

### DIFF
--- a/src/operator/first.ts
+++ b/src/operator/first.ts
@@ -58,9 +58,15 @@ export function first<T, R>(predicate?: (value: T, index: number, source: Observ
   return this.lift(new FirstOperator(predicate, resultSelector, defaultValue, this));
 }
 
+// We can cast `T -> R` in it if we pass the result selector.
+// Therefore we don't provide the signature which takes both a type guard function
+// as the predicate and the result selector.
+// (see #1936)
 export interface FirstSignature<T> {
   (predicate?: (value: T, index: number, source: Observable<T>) => boolean): Observable<T>;
+  <S extends T>(predicate?: (value: T, index: number, source: Observable<T>) => value is S): Observable<S>;
   (predicate: (value: T, index: number, source: Observable<T>) => boolean, resultSelector: void, defaultValue?: T): Observable<T>;
+  <S extends T>(predicate: (value: T, index: number, source: Observable<T>) => value is S, resultSelector: void, defaultValue?: S): Observable<S>;
   <R>(predicate?: (value: T, index: number, source: Observable<T>) => boolean, resultSelector?: (value: T, index: number) => R,
       defaultValue?: R): Observable<R>;
 }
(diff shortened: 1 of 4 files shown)


david-driscoll:
Let me preface this with, awesome job, this is something I wouldn't have thought of directly, but it does make it much easier to flow types, if you have to deal with unions 👏.

Only issue I have is that I think this applies to more operators than just filter.

I feel like this would benefit....
* `filter`
* `find`
* `first`
* `last`

cc @blesh @staltz @kwonoj 

It may even also apply to `partition` (although we can only narrow the positive result type).

kwonoj:
Agreeing with @david-driscoll , I think this change's good and would like to expand to other possibly supported operators as well.

saneyuki:
okay. I'll update this pull request.

saneyuki:
> okay. I'll update this pull request.

updated :)

david-driscoll:
LGTM

:+1:

src/operator/first.ts
```diff
@@ -60,9 +60,14 @@ export function first<T, R>(predicate?: (value: T, index: number, source: Observ
 
 export interface FirstSignature<T> {
   (predicate?: (value: T, index: number, source: Observable<T>) => boolean): Observable<T>;
+  <S extends T>(predicate?: (value: T, index: number, source: Observable<T>) => value is S): Observable<S>;
   (predicate: (value: T, index: number, source: Observable<T>) => boolean, resultSelector: void, defaultValue?: T): Observable<T>;
+  <S extends T>(predicate: (value: T, index: number, source: Observable<T>) => value is S, resultSelector: void, defaultValue?: S): Observable<S>;
   <R>(predicate?: (value: T, index: number, source: Observable<T>) => boolean, resultSelector?: (value: T, index: number) => R,
       defaultValue?: R): Observable<R>;
+  // We can cast `T -> R` in it if we pass the result selector.
```
kwonoj: it's just nitpicking, but I'm feeling this'd be better to be interface level comment (@ line 61) - what do you think?

kwonoj:
LGTM, only one nitpicking comment about comment location.

src/operator/first.ts
saneyuki: Ok. I'll change to it.

saneyuki:
@kwonoj I just updated & fixed what your pointed.

rob3c:
I just asked about how to flow types through Observable chains in this typescript issue: https://github.com/Microsoft/TypeScript/issues/7657#issuecomment-258544213

It'd be really nice to avoid what seems like a bunch of unnecessary casting, but I'm not sure if it's a typescript limitation or if it's addressable via better type definitions.

rob3c:
Just a quick update: I found a way to rewrite the type definitions to not require all of the casting above. I'm creating a pull request now...

PR #1937, opened 2016-09-12
style(typings): updated ObservableInput<T> to use PromiseLike<T> instead of Promise<T>

david-driscoll:
**Description:**
It appears that there are still some typing issues related to `catch` on both `Promise<T>` and `Observable<T>`.

This change simply reduces what we consider a `Promise<T>` just the thenable interface.  I double checked our code and our promise handling code only requires `then` and does not require `catch`.  This change will also make it easier for TypeScript consumers to return promise like items, instead of true promises.

Much like with the catch issue before I don't really have a valid test case, because all the unit tests do not use the `...Signature<T>` interfaces, and instead just the methods.

Once we move to TS2.0 that will change, as we remove the signature interfaces and move them to the methods themselves.

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -14,7 +14,7 @@ export interface Subscribable<T> {
             complete?: () => void): AnonymousSubscription;
 }
 
-export type SubscribableOrPromise<T> = Subscribable<T> | Promise<T>;
+export type SubscribableOrPromise<T> = Subscribable<T> | PromiseLike<T>;
 export type ObservableInput<T> = SubscribableOrPromise<T> | ArrayLike<T>;
 
 /**


jayphelps:
I'm initially 👍 to this because we have actually been hit by this with Bluebird, which doesn't actually conform 100% to the Promise interface.

> Neither type 'Bluebird<any>' nor type 'Promise<any>' is assignable to the other.
  Property '[Symbol.toStringTag]' is missing in type 'Bluebird<any>'.

Bluebirds prescribed "solution" was to ask you to wrap Bluebird promises in real ones with `Promise.resolve(bluebirdPromise)`

SomeKittens:
Agreed, I've also hit problems with Bluebird not strictly implementing `Promise` and would appreciate the extra flexibility that `PromiseLike<T>` gives.

david-driscoll:
It's the choice between giving users a foot gun, or causing them more pain to conform strictly to our interface.

In this case a miss behaving promise would have problems either way, this just makes it so that the types don't break down.  In TypeScript they could just force cast to `any`, and still cause themselves problems.

In my scenario today, I switched from `beta10` to `beta12` in one project, and all the types were failing.  `switchMap`, `mergeMap`, `Observable.combineLatest`.  It wasn't pretty.  This one change reversed all the problems though.

kwonoj:
For me this change looks ok. In general I'd follow strict interface as much, but only if it doesn't hurt real world usecases doesn't feel uncomfortable.

jayphelps:
LGTM

In some future version of RxJS, I think we'll make this more strict--certainly after Promise polyfills are no longer used. Until then, I think this is a very satisfying, practical compromise.

jayphelps:
@david-driscoll can I get a rebase?

david-driscoll:
rebased

jayphelps:
Thanks @david-driscoll !

PR #1939, opened 2016-09-13
fix: add missing expectSubscriptions constant

tklever:
Minor fix, declaring a constant that was missing. Just keeping my editor from barking at me.


jayphelps:
LGTM, thanks!

Issue #1829, opened 2016-07-14
Extending Observable


trxcllnt:
I think we're mature enough to start looking at this closer. As requested before (#1124, #1207, #1353, #1757), extending Observable should be seamless. Most of the time extending Observable and overriding `lift` does the job, but there are a few cases this breaks down.

If this sounds reasonable, I'll start work on a PR in the next few days.
#### Operators that call static methods

These are combineLatest, concat, merge, race, and zip. Implementing these with `lift` might be tedious, but doable.
#### Multicast/ConnectableObservable, and operators that emit inner Observables
1. We should implement `multicast` with selector as an operator instead of a custom Observable subclass.
2. `ConnectableObservable`, window, groupBy, and eventually join/groupJoin, should all use the prototype from the current `this` binding, and we can mixin their respective prototypes with `Object.create`. Fortunately, the `ConnectableObservable` implementation is already written with Operators and Subscribers, so aside from the public `multicast` method, this shouldn't be a lot of rework:

``` es6
class Observable {
  multicast(subjectFactory) {
    const connectable = Object.create(this, connectableObservableDescriptor);
    connectable.subjectFactory = subjectFactory;
    return connectable;
  }
}
const connectableObservableDescriptor = {
  connect: { value: function () { /* ... */} },
  refCount: { value: function () { /* ... */ } },
  _subscribe: { value: function _subscribe(subscriber) {
      return this.getSubject().subscribe(subscriber);
    }
  },
  getSubject: { value: function getSubject() {
      const subject = this._subject;
      if (!subject || subject.isStopped) {
        this._subject = this.subjectFactory();
      }
      return this._subject;
    }
  }
}
```

jayphelps:
regarding the static methods..are you referring to [this kind of stuff](https://github.com/ReactiveX/rxjs/blob/da91d6a3b4cff87fd16d7b438ebb4cb074dd9e49/src/observable/combineLatest.ts#L95):

```ts
return new ArrayObservable(observables, scheduler).lift(new CombineLatestOperator<T, R>(project));
```

If so, couldn't this be solved by deferring to the `lift` from the current prototype instead?

```ts
return this.lift.call(new ArrayObservable(observables, scheduler), new CombineLatestOperator<T, R>(project));
```

isiahmeadows:
@jayphelps That looks useful if you define `Observable.prototype.lift` to default to this:

```js
lift(o, f) {
  return o.lift(f)
}
```

[...]

PR #1941, opened 2016-09-13
feat(operators): Use lift in the operators that don't currently use lift.

trxcllnt:
**Description:**
Use lift in the operators that don't currently use lift.

#1829


src/operator/merge.ts
```diff
@@ -149,7 +148,7 @@ export function mergeStatic<T, R>(...observables: Array<ObservableInput<any> | S
     concurrent = <number>observables.pop();
   }
 
-  if (observables.length === 1) {
+  if (scheduler === null && observables.length === 1) {
```
jayphelps: did this fix an unreported bug?

src/Rx.ts
```diff
@@ -150,7 +150,6 @@ export {Subscriber} from './Subscriber';
 export {AsyncSubject} from './AsyncSubject';
 export {ReplaySubject} from './ReplaySubject';
 export {BehaviorSubject} from './BehaviorSubject';
-export {MulticastObservable} from './observable/MulticastObservable';
```
jayphelps: This will require a BREAKING CHANGE note, if accepted.

src/operator/race.ts
```diff
@@ -23,8 +23,7 @@ export function race<T>(...observables: Array<Observable<T> | Array<Observable<T
     observables = <Array<Observable<T>>>observables[0];
   }
 
-  observables.unshift(this);
-  return raceStatic.apply(this, observables);
+  return this.lift.call(raceStatic<T>(this, ...observables));
```
jayphelps: Without `observables.unshift(this)` doesn't this now have different behavior ? Was this a bug?

src/operator/merge.ts
trxcllnt: oh weird, that must have slipped in when I branched from master. yes, it's a bug fix that I was tracking down earlier today but didn't get to finish writing a test for. I can try to remove it from this change set, or alternatively add the test I was going to write and push again :[

src/Rx.ts
trxcllnt: bummertown. I'll amend the commit message.

jayphelps:
Maybe I'm missing something but doesn't this break static operator usage, which leads me to believe we're _not currently testing them like that_ 😱😱😱.

e.g. [static `merge`](https://github.com/trxcllnt/RxJS/blob/241787ed5b415f4ef6613b2c9de4dade89d7ae8c/src/observable/merge.ts) just references the instance operator function.

```js
import { merge } from 'rxjs/observable/merge';

Observable.merge(stream1, stream2);
// or
merge(stream1, stream2);
```

As well as
```js
import 'rxjs/add/observable/merge';

Observable.merge(stream1, stream2);
```

Did you have thoughts on the somewhat related issue of static methods not using lift?

https://github.com/ReactiveX/rxjs/issues/1829#issuecomment-239278815

jayphelps:
I totally missed that `Observable.merge` aka static merge does not use the regular merge, instead using `mergeStatic` which doesn't rely on `this`. So perhaps I'm drunk and this isn't an issue. Though I still would like static methods to lift too, but we could in fact punk on that regarding this PR.

src/operator/race.ts
trxcllnt: Shouldn't be. The instance method unwraps the argument if it's an Array of Observables, but I just looked and we don't have any test for this, which makes me nervous.

jayphelps: image

trxcllnt:
@jayphelps I'm also keen to update the static methods to use a creation/conversion method, but I wanted to get the instance methods out of the way first, as I imagine it'll be a bit before the static method is finalized.

src/operator/concat.ts
```diff
@@ -119,5 +119,9 @@ export function concatStatic<T, R>(...observables: Array<ObservableInput<any> |
     scheduler = args.pop();
   }
 
+  if (scheduler === null && observables.length === 1) {
+    return <Observable<R>>observables[0];
+  }
+
```
jayphelps: is this another one that slipped in or?

trxcllnt: heh yep

trxcllnt: added a test for this

src/operator/merge.ts
trxcllnt: added a test for this

src/operator/race.ts
trxcllnt: added a test for this

jayphelps:
LGTM, but needs rebase.

jayphelps:
@trxcllnt I forgot to remind you: this needs a BREAKING CHANGE added to the commit for the removal of `MulticastObservable`.

trxcllnt:
@jayphelps done

jayphelps:
LGTM

PR #1942, opened 2016-09-13
refactor(Operator): `call` renamed to `connect`

blesh:
because `call` is a method on `Function.prototype`, renaming this method to `connect` to prevent confusion. `connect` is more descriptive for what the method is doing, given than it connects two subscribers and subscribes to an observable

BREAKING CHANGE: `Operator.prototype.call` is now `Operator.prototype.connect`

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -629,8 +629,8 @@ describe('Observable.lift', () => {
       constructor(private childOperator: Rx.Operator<T, R>) {
       }
 
-      call(subscriber: Rx.Subscriber<R>, source: any): TeardownLogic {
-        return this.childOperator.call(new LogSubscriber<R>(subscriber), source);
+      connect(subscriber: Rx.Subscriber<R>, source: any): TeardownLogic {
+        return this.childOperator.connect(new LogSubscriber<R>(subscriber), source);
       }
     }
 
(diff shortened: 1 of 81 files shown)


trxcllnt:
@blesh 👎 from me. The method name and signature was designed to match `Function.prototype.call` (because in a non-JS environment, an operator _would_ just be a plain function). In the real world, I have operators which are implemented as pure functions, which expect `this` is the destination subscriber and the first argument is the source Observable.

jayphelps:
Because `call` natively already has meaning in JS world that is fairly widely known, I think this is a wise choice regardless of whether `call` would otherwise be the correctly descriptive name. I'm 👍 even though this will indeed be a breaking change for anyone with custom operators. Rip off the bandaid before 1.0.

@blesh did this come up because of anything in particular or just felt it was time?

Because of possible controversy, going to wait to see what others have to say.

trxcllnt:
The code `someOperator.call(subscriber, source)` should invoke `someOperator` on the `subscriber` with the supplied `source`, just like `someFunction.call(context, argument)` would invoke `someFunction` on the `context` with the supplied `argument`. This isn't a bug, it's a feature.

staltz:
Absolutely agree with Paul Taylor here, and I've always seen `operator.call` the way that he described, even though there was never such explanation written down. Also, why just do a breaking change just like that...

blesh:
Well, I'm going to close this for now. Not because I think it's good to ever use `call` as a name for a method on an Object in JavaScript, but because it's a breaking change that has little value to the upcoming release. cc/ @jayphelps

PR #1949, opened 2016-09-15
fix(bufferTime): do not create buffer when unsubscribed

### LINKED ISSUE #1944: Error on complete when using .bufferTime().take() and supplying a maxBufferSize
**RxJS version:**
rxjs@5.0.0-beta.12

**Code to reproduce:**
[jsbin example](https://jsbin.com/megavewube/1/edit?js,console)

```
Rx.Observable.interval(100)
  .bufferTime(1000, null, 5)
  .take(5)
  .subscribe(console.log.bind(console.log));
```

**Expected behavior:**
Stream should complete without any errors.

**Actual behavior:**
Throws a TypeError on complete:

```
Uncaught TypeError: Cannot read property 'push' of null
    BufferTimeSubscriber.openContext @ Rx.js:6288
    BufferTimeSubscriber.onBufferFull @ Rx.js:6280
    BufferTimeSubscriber._next @ Rx.js:6256
    Subscriber.next @ Rx.js:1301
```

**Additional information:**
- This bug only occurs when supplying a maxBufferSize and using a bufferCreationInterval of null.
- Looking at the code the problem is quite obvious, need to guard `this.contexts` in `BufferTimeSubscriber.prototype.openContext`
- Very similar to issue #1763 / PR #1764

kwonoj:
**Description:**
This PR fixes checking if given `bufferCreationInterval` is `null` or not, to execute legit paths when only timespan's supplied.

- closes #1944

### DIFF
--- a/src/operator/bufferTime.ts
+++ b/src/operator/bufferTime.ts
@@ -120,7 +120,7 @@ class BufferTimeSubscriber<T> extends Subscriber<T> {
               private scheduler: Scheduler) {
     super(destination);
     const context = this.openContext();
-    this.timespanOnly = bufferCreationInterval == null || bufferCreationInterval < 0;
+    this.timespanOnly = bufferCreationInterval === null || bufferCreationInterval < 0;
     if (this.timespanOnly) {
       const timeSpanOnlyState = { subscriber: this, context, bufferTimeSpan };
       this.add(context.closeAction = scheduler.schedule(dispatchBufferTimeSpanOnly, bufferTimeSpan, timeSpanOnlyState));
@@ -168,13 +168,13 @@ class BufferTimeSubscriber<T> extends Subscriber<T> {
     this.contexts = null;
   }
 
-  protected onBufferFull(context: Context<T>) {
+  protected onBufferFull(context: Context<T>): void {
     this.closeContext(context);
     const closeAction = context.closeAction;
     closeAction.unsubscribe();
     this.remove(closeAction);
 
-    if (this.timespanOnly) {
+    if (this.timespanOnly && this.contexts !== null) {
       context = this.openContext();
       const bufferTimeSpan = this.bufferTimeSpan;
       const timeSpanOnlyState = { subscriber: this, context, bufferTimeSpan };


src/operator/bufferTime.ts
```diff
@@ -120,7 +120,7 @@ class BufferTimeSubscriber<T> extends Subscriber<T> {
               private scheduler: Scheduler) {
     super(destination);
     const context = this.openContext();
-    this.timespanOnly = bufferCreationInterval == null || bufferCreationInterval < 0;
+    this.timespanOnly = bufferCreationInterval === null || bufferCreationInterval < 0;
```
jayphelps: Help me understand plz, why is there a notable difference between `undefined` and `null` here? i.e. why the fact that `== null` would match `undefined` too is a problem?

kwonoj: wow, it's my bad. apologize. This isn't legit complete fix and code should be updated more.

jayphelps: Phew! I was racking my brain!! 😅😅

src/operator/bufferTime.ts
```diff
@@ -168,13 +168,13 @@ class BufferTimeSubscriber<T> extends Subscriber<T> {
...
 
-  protected onBufferFull(context: Context<T>) {
+  protected onBufferFull(context: Context<T>): void {
     this.closeContext(context);
     const closeAction = context.closeAction;
     closeAction.unsubscribe();
     this.remove(closeAction);
 
-    if (this.timespanOnly) {
+    if (this.timespanOnly && this.contexts !== null) {
```
kwonoj: this is actual changes, do not attempt to open buffer if it's already unsubscribed.

jayphelps: Can we get a regression test around this?

kwonoj: Not sure. This occurs with async context, not able to think trivial way to create test cases at the moment writing PR. Let me think a bit more.

kwonoj: I thought this bit, but wasn't able to come up with clean way to create test coverage via test scheduler (or synchronous way) to mimic unsubscription occurs before scheduled action triggered.

kwonoj:
I think PR https://github.com/ReactiveX/rxjs/pull/1998 covers same issue and also having test coverage backing it, closing this PR without merging.

PR #1950, opened 2016-09-16
feat(operator): Add toSet operator

ahstro:
**Description:**
Bring back the `toSet` operator from RxJS 4.

**Disclaimer:**
I have no idea what I'm doing, I mostly just copied `toArray` stuff and modified it to make sense. It works, and the tests seem happy, but if you're even considering maybe thinking about possibly merging this, read through it or ask me simple questions that I'll fail to answer so you're forced to read through it.

Specific things I'm sure I must've messed up:
- The JSDoc in `src/operator/toSet.ts`
- The spec, probably. I mean, the tests pass, but I'm sure that's just luck.

### DIFF
--- a/spec/operators/toSet-spec.ts
+++ b/spec/operators/toSet-spec.ts
@@ -1,0 +1,112 @@
+import * as Rx from '../../dist/cjs/Rx';
+declare const {hot, cold, asDiagram, expectObservable, expectSubscriptions, type};
+
+const Observable = Rx.Observable;
+
+/** @test {toSet} */
+describe('Observable.prototype.toSet', () => {
+  asDiagram('toSet')('should reduce the values of an observable into an set', () => {
+    const e1 =   hot('-a-a--b-b|');
+    const e1subs =   '^        !';
+    const expected = '---------(w|)';
+
+    expectObservable(e1.toSet()).toBe(expected, { w: new Set(['a', 'b']) });
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should be never when source is never', () => {
+    const e1 =  cold('-');
+    const e1subs =   '^';
+    const expected = '-';
+
+    expectObservable(e1.toSet()).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should be never when source is empty', () => {
+    const e1 =  cold('|');
+    const e1subs =   '(^!)';
+    const expected = '(w|)';
+
+    expectObservable(e1.toSet()).toBe(expected, { w: new Set() });
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should be never when source doesn\'t complete', () => {
+    const e1 = hot('--x--^--y--');
+    const e1subs =      '^     ';
+    const expected =    '------';
+
+    expectObservable(e1.toSet()).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should reduce observable without values into an set of size zero', () => {
+    const e1 = hot('-x-^---|');
+    const e1subs =    '^   !';
+    const expected =  '----(w|)';
+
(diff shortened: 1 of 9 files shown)


ahstro:
Okay, two force pushes later.. This is why I need adult supervision.

spec/operators/toSet-spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  it('should work with throw', () => {
+    const e1 =  cold('#');
+    const e1subs =   '(^!)';
+    const expected = '#';
+
+    expectObservable(e1.toSet()).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  type(() => {
```
kwonoj: this isn't necessary in general unless tricky type definition verification is required.

src/operator/toSet.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * We need this JSDoc comment for affecting ESDoc.
+ * @ignore
+ * @extends {Ignored}
+ */
+class ToSetSubscriber<T> extends Subscriber<T> {
+
+  private set: Set<{}> = new Set();
+
+  constructor(destination: Subscriber< Set<T>>) {
```
kwonoj: nitpicking :), typo `< Set` -> `<Set`

jayphelps:
@ahstro I really do appreciate you taking the time for this.

I'm personally 👎 though because there has been discussion around keeping v5 possible API surface lean. Most recently in #1928. I struggle to see many use cases for this operator--particularly because it's so easy to populate a Set:

```js
item$.reduce((set, item) => set.add(item), new Set)
```

There is almost certainly an opportunity for a sort of "kitchen sink" library that people can add every sort of operator under the sun, but nothing exists yet.

I'll let others chime in with their thoughts.

kwonoj:
I'm bit curious of usecases of this operator too.

ahstro:
@jayphelps Yeah, that's what I figured, I just wanted to create the PR as an entrypoint for the discussion (didn't know about #1928) instead of just an issue, but was well aware that it might be deemed superfluous considering how easy it is to substitute with other operators, like you mentioned.

jayphelps:
Keeping this open for now, we're going to discuss the future of operators like this in our next team meeting so we have definitive alignment.

jayphelps:
@ahstro We met today and discussed this PR. ([notes here](https://github.com/ReactiveX/rxjs-core-notes/blob/master/2016-10/october-24.md#outstanding-prs)).

We want to continue reduce the size of the core set of operators, which might even mean removing some that already exist. We''ll continue to discuss the prospect of optional addons, which they may be a better fit for. There are many possibilities of what they would look like, but for now we want to keep our focus on shipping v5.0.0-final as soon as possible and will come back to this shortly after.

At that time, we'll undoubtedly need some changes to this PR to adapt. Do you want to keep this PR open until then or close for now? If you have extra cycles, we'd be open to someone spearheading the addon effort and would love your help if you're interested.

ahstro:
@jayphelps I read the notes and what you discussed sounds sane; I'm all for a smaller default import and moving this feature to an optional addon - whether just not imported or completely separate - feels, to me, like the best option.

I would love to help out with the effort of creating that solution. I'm not sure if _spearheading_ is appropriate since this is my first proverbial toe dip in this codebase, but I wouldn't mind getting involved and help out to the extent of my ability.

I _Watched_ `rxjs-core-notes`, so if future discussions will be held there, I'll be able to follow along, so if this PR should be open depends firstly on whether this operator should be included in the addon or not and secondly if the addon will be hosted in this repository or another? I don't care one way or the other, the code will be available when and if it's needed :)

kwonoj:
I'm going to close this PR as inactive discussion, meanwhile we've set our default recommendation to create operator as separate module first (https://github.com/ReactiveX/rxjs/blob/master/doc/operator-creation.md#publish-your-operator-as-a-separate-library) and evaluate to propagate into core library as needed. Appreciate for effort @ahstro and apologize makes PR hanging without having concrete decision soon enough.

Issue #1952, opened 2016-09-18
timer on animationFrame scheduler is broken in 5.0.0-beta.12


chirayuk:
**RxJS version:**  v5.0.0-beta.12

**Code to reproduce:**

```
Rx.Observable.timer(0, 1000, Rx.Scheduler.animationFrame).subscribe((x) => console.log(x));
```

**Expected behavior:**
I expect no exceptions and a console log every second.

**Actual behavior:**
A stacktrace every second (as long as there is no other animation frame scheduler action already queued by by something else.)

```
Uncaught TypeError: Cannot read property 'execute' of undefinedAnimationFrameScheduler.flush @ main_bundle.js?FXszpMTp_x7svlvayVh8cQ==:54276
setInterval (async)AsyncAction.requestAsyncId @ main_bundle.js?FXszpMTp_x7svlvayVh8cQ==:54449
AnimationFrameAction.requestAsyncId @ main_bundle.js?FXszpMTp_x7svlvayVh8cQ==:54225
AsyncAction.schedule @ main_bundle.js?FXszpMTp_x7svlvayVh8cQ==:54444
TimerObservable.dispatch @ main_bundle.js?FXszpMTp_x7svlvayVh8cQ==:43938
AsyncAction._execute @ main_bundle.js?FXszpMTp_x7svlvayVh8cQ==:54495
AsyncAction.execute @ main_bundle.js?FXszpMTp_x7svlvayVh8cQ==:54470
AnimationFrameScheduler.flush @ main_bundle.js?FXszpMTp_x7svlvayVh8cQ==:54276
requestAnimationFrame (async)AnimationFrameAction.requestAsyncId @ main_bundle.js?FXszpMTp_x7svlvayVh8cQ==:54232
AsyncAction.schedule @ main_bundle.js?FXszpMTp_x7svlvayVh8cQ==:54444
Scheduler.schedule @ main_bundle.js?FXszpMTp_x7svlvayVh8cQ==:40084
TimerObservable._subscribe @ main_bundle.js?FXszpMTp_x7svlvayVh8cQ==:43943
Observable.subscribe @ main_bundle.js?FXszpMTp_x7svlvayVh8cQ==:39607
(anonymous function) @ VM3011:1
```

**Additional information:**
I think this used to work in `v5.0.0-beta.9`

chirayuk:
[...]

PR #1953, opened 2016-09-18
fix(schedulers): Fix Asap and AnimationFrame Schedulers to act like the Async Scheduler if delay > 0

trxcllnt:
**Description:**
Asap and AnimationFrame Schedulers should act like the Async Scheduler if delay > 0.

#1952

### DIFF
--- a/spec/schedulers/AnimationFrameScheduler-spec.ts
+++ b/spec/schedulers/AnimationFrameScheduler-spec.ts
@@ -1,4 +1,5 @@
 import {expect} from 'chai';
+import * as sinon from 'sinon';
 import * as Rx from '../../dist/cjs/Rx';
 
 const animationFrame = Rx.Scheduler.animationFrame;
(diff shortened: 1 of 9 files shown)


spec/schedulers/AnimationFrameScheduler-spec.ts
```diff
@@ -20,6 +20,19 @@ describe('Scheduler.animationFrame', () => {
     }
   });
 
+  it('should act like the async scheduler if delay > 0', (done: MochaDone) => {
```
kwonoj: I'd like to suggest to use `fakeTimer` same as other test case ([here](https://github.com/ReactiveX/rxjs/blob/e91d113e49acc9b0e12e2524040fa84cf3fa2c78/spec/schedulers/QueueScheduler-spec.ts#L11-L35)) to avoid intermittent test failure by timing condition of test execution environment (usually occurs on mobile emulator browser test)

jayphelps:
@trxcllnt can we get a regression test for #1952, one where these changes are being flexed in a real observable?

trxcllnt:
@jayphelps sure thing

trxcllnt:
@jayphelps Scheduler with Observable tests added in `spec/observables/interval-spec.ts`

jayphelps:
LGTM

PR #1955, opened 2016-09-19
refactor(build): introduce closure-compiler-js for minified build

kwonoj:
**Description:**
This PR replaces current build step of minified build using `closure` compiler, by introducing js implementation (https://github.com/google/closure-compiler-js). One of drawback of `google-closure-compiler` package is it's just thin wrapper to java implementation of compiler, so requires JRE as dependency. With js implementation of compiler, JRE is no longer required.


jayphelps:
LGTM, needs rebase.

PR #1957, opened 2016-09-19
test(expand): add test coverage to expand

kwonoj:
**Description:**
This PR adds test coverage to `expand` operator, cases where scheduler is specified.

### DIFF
--- a/spec/operators/expand-spec.ts
+++ b/spec/operators/expand-spec.ts
@@ -3,6 +3,7 @@ import * as Rx from '../../dist/cjs/Rx';
 declare const {hot, cold, expectObservable, expectSubscriptions};
 
 declare const Symbol: any;
+declare const rxTestScheduler: Rx.TestScheduler;
 const Observable = Rx.Observable;
 
 /** @test {expand} */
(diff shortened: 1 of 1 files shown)


spec/operators/expand-spec.ts
```diff
@@ -21,6 +22,19 @@ describe('Observable.prototype.expand', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
+  it('should work with scheduler', () => {
+    const e1 =    hot('--x----|  ', {x: 1});
+    const e1subs =    '^        !';
+    const e2 =   cold('--c|      ', {c: 2});
```
staltz: We could make this test a bit easier to read. `e2` isn't used at the first time frame. So you could write instead `cold(  '--c|'`... (two spaces) because that's when the cold observable would be subscribed to.

spec/operators/expand-spec.ts
```diff
@@ -21,6 +22,19 @@ describe('Observable.prototype.expand', () => {
...
+    const expected =  '--a-b-c-d|';
+    const values = {a: 1, b: 2, c: 4, d: 8};
+
+    const result = e1.expand(x => x === 8 ? Observable.empty() : e2.mapTo(2 * x), Number.POSITIVE_INFINITY, rxTestScheduler);
```
staltz: Then, we could get the number `2` from `e2` itself. `: e2.map(c => c * x)`

kwonoj:
PR's updated per suggestion.

staltz:
LGTM

jayphelps:
LGTM too, needs rebase.

PR #1960, opened 2016-09-20
unsubscribe() instead of dispose()

natedean:
dispose() is no longer the way to dispose of a subscription handle


jayphelps:
LGTM

Issue #1958, opened 2016-09-19
Rename do back to tap (or atleast re add alias)


BLamy:
Choosing a reserved word just seems like it's asking for problems. 

I can't for the life of me figure out how to get do to work with function bind syntax.

![image](https://cloud.githubusercontent.com/assets/324928/18639359/989bb9b8-7e61-11e6-887b-31997e4758c0.png)

`import { _do } from 'rxjs/operator/do';`
Seems to work but the underscore makes me a little cautious.

staltz:
Look serious/legit.

jayphelps:
This isn't specific to `do`, we have numerous operators which are reserved words in this context.

We're going to try to be more aggressive in releasing v5 final asap, so the ship has most likely sailed for v5 on this--but we can discuss solutions for v6.

In the future I'd personally like to see us move completely away from reserved words, instead of aliasing. This will be tough to bikeshed since most (or all?) of these are inherited from other Rx implementations so changing them causes us to diverge even further from the common "reactive extensions" operator names. But we have unique usage and requirements around JS so I think it's still appropriate to do what's right for JS.

staltz:
There's so many names that coincide with reserved words. Observable.`of` to begin with. :(

jayphelps:
@staltz `of` isn't a reserved word 😄

jayphelps:
@BLamy remember that you can easily alias imports


```js
```

staltz:
Ah, I assumed it was, as you can do nowadays `for (var key of obj)`

BLamy:
The only two operators which conflict with reserved words I could find that were `let` and `do`. 

As jay said aliasing was a simple workaround but is using function bind still the preferred approach? I was only using it because it's labeled as the recommended approach but it's my first time ever using it.

jayphelps:
I believe this is an exhaustive list:

* `catch`
* `switch`
* `do`
* `finally`
* `let`
* `throw`

I think we need to remove the recommendation around using bind operator because I think it is too prescriptive considering it's still stage-0 and somewhat controversial among TC39. We can certainly still show it as an option, but we should clarify and not suggest it as the best solution. People take what we say as gospel without knowing the costs.

blesh:
[...]

PR #1962, opened 2016-09-20
docs(usage): Don't recommend function-bind operator as the "best" sol…

jayphelps:
In #1958 it was brought up that we have been recommending using the stage-0 function bind operator as the "best" way to use RxJS but I think this is misleading advice to people who don't realize that stage-0 proposals can be withdrawn without notice so they could be building their app using syntax that could vanish from Babel and have no future in JS.

I'm confident the advice was well intentioned, because it is one of the most flexible/ergonomic solutions. But the tradeoffs are risky and we need to be extremely careful because people sometimes take what we say as gospel.

This removes that suggestion and instead warns users of the nature of proposed syntax features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,9 @@ import 'rxjs/add/operator/map';
 Observable.of(1,2,3).map(x => x + '!!!'); // etc
 ```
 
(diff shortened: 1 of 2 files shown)


kwonoj:
LGTM

PR #1967, opened 2016-09-22
fix(operator): Fix take to complete when the source is re-entrant.

trxcllnt:
**Description:**
The TakeSubscriber compares the instance variable `count` to determine if it should complete or not.

If the source Observable is re-entrant, this check fails (because `count` is updated in each `next`), and take never completes.

I encountered this issue with expand and take on a multicast Observable, but here's a much simpler repro example:

```es6
import { Subject } from 'rxjs';

subject
  .take(10)
  .subscribe(
    (i) => {
      console.log(i);
      subject.next(i + 1);
    },
    (err) => {},
    () => console.log('done') // never gets printed
  );

subject.next(0);
```


kwonoj:
LGTM

PR #1976, opened 2016-09-25
fix(WebSocketSubject): ensure all internal state properly reset

### LINKED ISSUE #1863: WebSocketSubject does not reconnect with beta10 (change behaviour from beta7)

blesh:
When the internal socket is `null`-ed out, we weren't ensuring that all other Subjects that are being used internally were also being reset. This caused issues where sometimes internal subjects found themselves in a bad state when you tried to reconnect. This was even more nefarious in multiplexed web sockets.

This PR adds a single private method called `_resetState` that will reset the socket and all other Subjects to their original state.

fixes #1863

### DIFF
--- a/spec/observables/dom/webSocket-spec.ts
+++ b/spec/observables/dom/webSocket-spec.ts
@@ -394,6 +394,81 @@ describe('Observable.webSocket', () => {
   });
 
   describe('multiplex', () => {
+    it('should be retryable', () => {
+      const results = [];
+      const subject = Observable.webSocket('ws://websocket');
+      const source = subject.multiplex(() => {
+        return { sub: 'foo'};
+      }, () => {
+        return { unsub: 'foo' };
+      }, function (value: any) {
+        return value.name === 'foo';
+      });
+
+      source
+        .retry(1)
+        .map((x: any) => x.value)
+        .take(2)
+        .subscribe((x: any) => {
+          results.push(x);
+        });
+
+      const socket = MockWebSocket.lastSocket;
+      socket.open();
+
+      expect(socket.lastMessageSent).to.deep.equal({ sub: 'foo' });
+      socket.triggerClose({ wasClean: false }); // Bad connection
+
+      const socket2 = MockWebSocket.lastSocket;
+      expect(socket2).not.to.equal(socket);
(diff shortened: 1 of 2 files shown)


blesh:
I'm still thinking about how to test this in an automated way. I've got it working locally.

jayphelps:
LGTM

PR #1977, opened 2016-09-26
docs(Observable): bindNodeCallback

dancasey:
Change doc to match source: bindNodeCallback's selector parameter is optional.


dancasey:
Just noticed... I think this same change would also apply to `bindCallback`. I can update the PR if you want--just let me know.

kwonoj:
Yes, I think it'd better to update both.

dancasey:
Done.

blesh:
LGTM

blesh:
closed in favor of #2018, which retains @dancasey's commits and rebases.

PR #1978, opened 2016-09-26
style(typings): remove *Signature interfaces

david-driscoll:
**Description:**
This change removes the *Signature interfaces and moves them to the the methods themselves.  Using the new `this` syntax provided by TS2.0.

Requires TypeScript 2.0

#1968

**Details:**
This change was done entirely using a node script with regex's to deal with each error as they came.  So this script can be rerun at any point to run the conversion again.  This script is below.

The only manual change was for the Behavior subject specs, which is already solved in #1968 anyway.

```typescript
import * as _ from 'lodash';
import * as fs from 'fs';
import * as path from 'path';
const pathTo = '../RxJS/src';

const fileFlow = (path: string, ...flows: ((value: string) => string)[]): string => {
    const funcs = [
        (path: string) => fs.readFileSync(path).toString(),
        ...flows,
        (content: string) => fs.writeFileSync(path, content)
    ];
    return _.flow(...funcs)(path);
}

(() => {
    const p = path.join(pathTo, 'add', 'operator');
    const operators = fs.readdirSync(p);
    _(operators)
        .map(operator => ({ path: path.join(p, operator), name: operator.split('.')[0] }))
        .sortBy(x => x.name)
        .each(operator => {
            fileFlow(operator.path,
                // Replace all signature interfaces
                (file) => file.replace(/import \{\s?(.*?)(,\s)?(\w+Signature)\s?\} from/, 'import { $1 } from'),
                // Remove empoty imports (scan)
                (file) => file.replace(/import {\s+} from '.*?';/, ''),
                // Replace the signature implemenations
                (file) => {
                    const regex = new RegExp(`import {(.*?)} from '../../operator/${operator.name}'`);
                    const match = _.trim(file.match(regex)[1]);
                    return file.replace(/(\w+):\s\w+Signature<T>;/g, (str, firstItem: string) => {
                        if (firstItem.startsWith('_')) {
                            return `${firstItem}: typeof ${match};`;
                        }
                        return `${firstItem}: typeof ${match};`;
                    });
                });
        });
})();

(() => {
    const p = path.join(pathTo, 'operator');
    const operators = fs.readdirSync(p);
    let reduceSignatures;
    _(operators)
        .map(operator => path.join(p, operator))
        .sortBy(x => x)
        .each(operator => {
            fileFlow(operator,
                file => file.replace(/\/* tslint:disable:max-line-length *\//g, ''),
                file => {
                    if (!file.match(/import {\s?Observable.*?} from '(..\/)?..\/Observable';/)) {
                        file = "import { Observable } from '../Observable';\n" + file;
                    }
                    return file;
                },
                // Handle merge map operators
                file => {
                    if (file.indexOf('MergeMapOperator') > -1 || file.indexOf('SwitchMapOperator') > -1 || file.indexOf('SwitchMapToOperator') > -1) {
                        file = file
                            .replace(/: Observable<R> {/g, ': Observable<I | R> {')
                            // .replace(/: T/g, ': Observable<T>');
                    }
                    return file;
                },
                // Update the existing methods with this context
                file => {
                    const regex = /export function (.*?)(?:<(.*?)>)?\((.*)/;
                    const context = file.match(regex);
                    let generics = context[2];
                    if (!generics) {
                        generics = 'T';
                    }
                    else if (generics.indexOf('T') === -1) {
                        generics = `T, ${generics}`
                    }
                    if (_.trim(context[3])) {
                        let args: string = context[3];
                        const index = context[3].indexOf(')');
                        if (index > -1) {
                            args = context[3].substr(0, index);
                        }
                        if (_.trim(args)) {
                            return file.replace(regex, `export function $1<${generics}>(this: Observable<T>, $3`)
                        }
                    }
                    return file.replace(regex, `export function $1<${generics}>(this: Observable<T>$3`)
                },
                // Move signatures into function delcaration
                file => {
                    const funcName = (() => {
                        const result = file.match(/export function (\w+)/);
                        return result && result[1];
                    })();
                    let signatures = (() => {
                        const result = file.match(/export interface \w*Signature<T> {([\s|\S]*?)}/);
                        file = file.replace(/export interface \w*Signature<T> {([\s|\S]*?)}/, '');
                        if (result && result.length) {
                            const group = result[1];
                            const signatureContent = _.trim(result[1]);

                            const lines = _(signatureContent.split('\n'))
                                .map(_.trim)
                                .map(x => _.trimEnd(x, ';'))
                                .value();

                            const consolidatedLines = [];

                            for (let i = 0; i < lines.length; i++) {
                                if (_.startsWith(lines[i], '<') || _.startsWith(lines[i], '(')) {
                                    consolidatedLines.push(lines[i]);
                                } else {
                                    consolidatedLines[consolidatedLines.length - 1] += ' ' + lines[i];
                                }
                            }

                            return _(consolidatedLines)
                                .map(line => {
                                    // Ensure `T` is available
                                    if (line.startsWith('<T,')) {
                                        line = '<T,' + line.substr(3);
                                    } else if (line.startsWith('<')) {
                                        line = `<T, ` + line.substr(1);
                                    } else {
                                        line = '<T>' + line;
                                    }

                                    // Set `this`
                                    if (line.indexOf('()') > -1) {
                                        line = line.replace('()', '(this: Observable<T>)');
                                    } else {
                                        line = line.replace('(', '(this: Observable<T>, ');
                                    }

                                    return `export function ${funcName}${line};`;
                                })
                                .value();
                        }
                        return [];
                    })();

                    if (funcName === 'reduce') {
                        reduceSignatures = _.map(signatures, x => x.replace(' reduce', ' scan'));
                    }

                    if (funcName === 'scan') {
                        signatures = reduceSignatures;
                    }

                    const sigs = [
                        '/* tslint:disable:max-line-length */',
                        ...signatures,
                        '/* tslint:disable:max-line-length */'
                    ].join('\n');

                    if (signatures.length > 1) {
                        return file.replace(`export function`, sigs + '\nexport function');
                    }
                    return file;
                },
                // fix first / last
                file => file
                    .replace(/resultSelector\?: \(value: T, index: number\) => R \| void,$/mg, 'resultSelector?: (value: T, index: number) => R,')
                    .replace(/resultSelector\?: \(value: T, index: number\) => R,$/mg, 'resultSelector?: ((value: T, index: number) => R) | void,')
                    .replace(/ \| void \| void/, ' | void')
                    .replace(/result = this\.resultSelector\(value, index\)/, 'result = (<any>this).resultSelector(value, index)'),
                // fix groubBy
                file => file
                    .replace(/elementSelector\?:\ \(value:\ T\)\ =>\ R,$/mg, 'elementSelector?: ((value: T) => R) | void,'),
                // Handle merge operators
                file => {
                    if (file.indexOf('MergeAllOperator') > -1 || file.indexOf('SwitchOperator') > -1 || file.indexOf('FindValueOperator') > -1) {
                        file = file
                            .replace(/this\.lift\(new MergeAllOperator/g, '<any>this.lift<any>(new MergeAllOperator')
                            .replace(/this\.lift\(new FindValueOperator/g, '<any>this.lift<any>(new FindValueOperator')
                            .replace(/this\.lift\(new SwitchOperator/g, '<any>this.lift<any>(new SwitchOperator');
                            // .replace(/Observable<T>/g, 'Observable<Observable<T>>')
                            // .replace(/: T/g, ': Observable<T>');
                    }
                    return file;
                },
                // handle sequenceEqual
                file => {
                    if (file.indexOf('SequenceEqualOperator') > -1){
                        file = file
                            .replace('Operator<T, T>', 'Operator<T, boolean>')
                            .replace('subscriber: Subscriber<T>', 'subscriber: Subscriber<boolean>');
                    }
                    return file;
                },
                // reduce Consecutive blank lines
                file => {
                    while (file.match(/\n(?:\s*?)\n/g)) {
                        file = file.replace(/\n(?:\s*?)\n/g, '\n');
                    }
                    file = file.replace(/\n(?:\s*?)$/mg, '\n');
                    return file;
                }
            );
        });
})();
```

### DIFF
--- a/src/operator/skipUntil.ts
+++ b/src/operator/skipUntil.ts
@@ -18,12 +18,8 @@ import { subscribeToResult } from '../util/subscribeToResult';
  * @method skipUntil
  * @owner Observable
  */
-export function skipUntil<T>(notifier: Observable<any>): Observable<T> {
+export function skipUntil<T>(this: Observable<T>, notifier: Observable<any>): Observable<T> {
   return this.lift(new SkipUntilOperator(notifier));
-}
-
-export interface SkipUntilSignature<T> {
-  (notifier: Observable<any>): Observable<T>;
 }
 
 class SkipUntilOperator<T> implements Operator<T, T> {
--- a/src/add/operator/audit.ts
+++ b/src/add/operator/audit.ts
@@ -1,10 +1,10 @@
 import { Observable } from '../../Observable';
-import { audit, AuditSignature } from '../../operator/audit';
+import { audit } from '../../operator/audit';
 
 Observable.prototype.audit = audit;
 
 declare module '../../Observable' {
   interface Observable<T> {
-    audit: AuditSignature<T>;
+    audit: typeof audit;
   }
 }
--- a/src/add/operator/auditTime.ts
+++ b/src/add/operator/auditTime.ts
@@ -1,10 +1,10 @@
 import { Observable } from '../../Observable';
-import { auditTime, AuditTimeSignature } from '../../operator/auditTime';
+import { auditTime } from '../../operator/auditTime';
 
 Observable.prototype.auditTime = auditTime;
 
 declare module '../../Observable' {
   interface Observable<T> {
-    auditTime: AuditTimeSignature<T>;
+    auditTime: typeof auditTime;
   }
 }
--- a/src/add/operator/buffer.ts
+++ b/src/add/operator/buffer.ts
@@ -1,11 +1,11 @@
 
 import { Observable } from '../../Observable';
-import { buffer, BufferSignature } from '../../operator/buffer';
+import { buffer } from '../../operator/buffer';
 
 Observable.prototype.buffer = buffer;
 
 declare module '../../Observable' {
   interface Observable<T> {
-    buffer: BufferSignature<T>;
+    buffer: typeof buffer;
   }
 }
--- a/src/add/operator/bufferCount.ts
+++ b/src/add/operator/bufferCount.ts
@@ -1,11 +1,11 @@
 
 import { Observable } from '../../Observable';
-import { bufferCount, BufferCountSignature } from '../../operator/bufferCount';
+import { bufferCount } from '../../operator/bufferCount';
 
 Observable.prototype.bufferCount = bufferCount;
 
 declare module '../../Observable' {
   interface Observable<T> {
-    bufferCount: BufferCountSignature<T>;
+    bufferCount: typeof bufferCount;
   }
 }
--- a/src/add/operator/bufferTime.ts
+++ b/src/add/operator/bufferTime.ts
@@ -1,11 +1,11 @@
 
 import { Observable } from '../../Observable';
-import { bufferTime, BufferTimeSignature } from '../../operator/bufferTime';
+import { bufferTime } from '../../operator/bufferTime';
 
 Observable.prototype.bufferTime = bufferTime;
 
 declare module '../../Observable' {
   interface Observable<T> {
-    bufferTime: BufferTimeSignature<T>;
+    bufferTime: typeof bufferTime;
   }
 }
--- a/src/add/operator/bufferToggle.ts
+++ b/src/add/operator/bufferToggle.ts
@@ -1,11 +1,11 @@
 
 import { Observable } from '../../Observable';
-import { bufferToggle, BufferToggleSignature } from '../../operator/bufferToggle';
+import { bufferToggle } from '../../operator/bufferToggle';
 
 Observable.prototype.bufferToggle = bufferToggle;
 
 declare module '../../Observable' {
   interface Observable<T> {
-    bufferToggle: BufferToggleSignature<T>;
+    bufferToggle: typeof bufferToggle;
   }
 }
--- a/src/add/operator/bufferWhen.ts
+++ b/src/add/operator/bufferWhen.ts
@@ -1,11 +1,11 @@
 
 import { Observable } from '../../Observable';
-import { bufferWhen, BufferWhenSignature } from '../../operator/bufferWhen';
+import { bufferWhen } from '../../operator/bufferWhen';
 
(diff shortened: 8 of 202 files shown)


paulpdaniels:
I almost can't tell if I'm looking at JavaScript or C# anymore :trollface:

src/operator/skipUntil.ts
```diff
@@ -18,59 +17,46 @@ import { subscribeToResult } from '../util/subscribeToResult';
...
     super(destination);
     this.add(subscribeToResult(this, notifier));
   }
-
   protected _next(value: T) {
     if (this.hasValue) {
       super._next(value);
     }
   }
-
```
jayphelps: I would prefer if we didn't remove all these newlines--if we did, I'd like it as a separate PR.

jayphelps: i.e. we can certainly discuss if we want to switch to this style, but this PR isn't the good place. 👯

jayphelps:
(besides removing the newlines) I love this. We need to decide for sure about requiring TS 2.0 first, but I suspect we will since a large majority of our TS users are Angular2.

jayphelps:
Also--have you heard of [jscodeshifts/codemods](https://github.com/facebook/jscodeshift) :trollface: hahaha

david-driscoll:
haha, I was going for something quick an dirty knowing that this is a one time thing.

src/operator/skipUntil.ts
david-driscoll: Most definitely an artifact of the script, we can either tweak the script or tweak the files after the fact.   We can just wait to tweak those things to closer to when we want to merge these changes in.

blesh:
Can rebase this now since #2017 and #1968 are now merged.

david-driscoll:
Updated.

Keep in mind this *_will_* be a breaking change, that can't be fixed easily by removing `readonly` as this utilizes the new `this` functionality for function definitions.

blesh:
LGTM. Nice work @david-driscoll

kwonoj:
👍

PR #1985, opened 2016-09-27
Update example docs for `.throttle()`

connor4312:
**Description:** the example for `.throttle()` was incorrect and resulted in errors being thrown, if it was run.

**Related issue (if exists):** none


staltz:
Are you *sure*? Look at the signature of throttle: https://github.com/ReactiveX/rxjs/blob/master/src/operator/throttle.ts#L48

throttleTime is the operator that accepts a number for the period.

connor4312:
Ah, my bad. I dumped this in a "RxJS Playground" jsfiddle that is using rxjs 4.x instead of 5.x!

PR #1995, opened 2016-10-03
fix(Observer): fix Observable#subscribe() signature to suggest correct usage

awerlang:
**Description:**

This was annoying me a long time already. So this patch describes better how to call the `.subscribe()` method, ruling out invalid combinations (like a partial observer object and an error handler).

- Overloads subscribe signature for usage with an empty subscriber, PartialObserver<T>, or next, error and complete handlers


blesh:
LGTM

cc/ @david-driscoll @kwonoj @jayphelps can one of you confirm?

jayphelps:
LGTM

PR #1998, opened 2016-10-03
fix(bufferTime): no errors with take after bufferTime with maxBufferSize

### LINKED ISSUE #1944: Error on complete when using .bufferTime().take() and supplying a maxBufferSize

blesh:
Ensures that a buffer context is not created if the subscriber is closed

fixes #1944

### DIFF
--- a/spec/operators/bufferTime-spec.ts
+++ b/spec/operators/bufferTime-spec.ts
@@ -300,4 +300,23 @@ describe('Observable.prototype.bufferTime', () => {
     expectObservable(result).toBe(expected, values);
     expectSubscriptions(e1.subscriptions).toBe(subs);
   });
+
+  it('should not have errors when take follows and maxBufferSize is provided', () => {
+    const tick = 10;
+    const bufferTime = 50;
+    const expected = '-----a----b----c----d----(e|)';
+    const values = {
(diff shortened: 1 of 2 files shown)


kwonoj:
LGTM

PR #2001, opened 2016-10-04
Fix sample in /doc/subject.md

sdf611097:
Sample code in /doc/subject.md cannot work.

### DIFF
--- a/doc/subject.md
+++ b/doc/subject.md
@@ -15,15 +15,14 @@ In the example below, we have two Observers attached to a Subject, and we feed s
 ```js
 var subject = new Rx.Subject();
 
-subject.subscribe({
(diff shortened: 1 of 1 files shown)


staltz:
I think you were using RxJS v4. This site is for RxJS v5, where those APIs should work.

kwonoj:
I just tried previous snippet (before changes) and it works correctly.

And by assuming by changes https://github.com/ReactiveX/rxjs/pull/2001/files#diff-23d072ae68d9006ee6aed7b5b394181eR125, I think this change is targeting v4 api surface while this doc is intended for v5 api surface.

@sdf611097 , let me close this PR for now. If I misunderstood something and this PR needs to be checked in, please feel freely reopen and correct me anytime.

Issue #2010, opened 2016-10-06
'cache()' should allow subscribe after unsubscribe


newsash:
**RxJS version:** Latest
**Code to reproduce:**
``` typescript
var cached = Observable.timer(200).cache();
var subscription = cached.subscribe(v => console.log(v));
setTimeout(() => {
    subscription.unsubscribe();
    cached.subscribe(v => console.log(v));
}, 100);
```
**Expected behavior:**
"0" in console.
**Actual behavior:**
Nothing in console.
**Additional information:**
Subject should be nulled on outerSub unsubscribtion
https://github.com/ReactiveX/rxjs/blob/master/src/operator/cache.ts#L55

blesh:
... and here's another issue that leads me to believe `cache` should be removed from v5 until we more specifically bikeshed it's behavior.

newsash:
Oh, it's probably a dupe of #1959 
Use case from my project (`serviceCall` is angular 2 `http.post` with retry logic):
``` typescript
cache.observable = serviceCall
    .expand(() => Observable.timer(period).concatMap(() => serviceCall))
    .cache(1, period);
```

jayphelps:
@newsash we ended up removing the `.cache()` operator [for RC1](https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md#500-rc1-2016-10-11) because in it's current state it was too confusing and [poorly] trying to solve too many problems.

Issue #2006, opened 2016-10-05
cache: Weird behaviour on throw


Kamshak:
**RxJS version:** 5.0.0-beta.12

**Code to reproduce:** 
```
const throwsInstantly = Rx.Observable.throw("E");

throwsInstantly.cache(1).subscribe(
  x => console.log(x),
  error => console.error("Cache", error),
  () => console.log("Done")
);
```
Example: https://codepen.io/Kamshak-1472390889/pen/gwowkZ?editors=0010#0

**Expected behavior:** Console prints "Cache E": The error is passed on.

**Actual behavior:** No output.

**Additional information:** Is this expected? The behaviour is different if the observable throws after the cache is subscribed (error is passed on).

blesh:
I think this is a clincher for removing cache until we can further refine the desired behaviors.

Dorus:
This fails for `Rx.Observable.throw("E", Rx.Scheduler.queue);` but works correct under `Scheduler.asap`, `Scheduler.async`, `Scheduler.animationFrame` and `new Rx.TestScheduler()`.

jayphelps:
@Dorus we ended up removing the `.cache()` operator [for RC1](https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md#500-rc1-2016-10-11) because in it's current state it was too confusing and [poorly] trying to solve too many problems.

We do plan to revisit in the future and possibly provide one or more operators that have differing forms of caching.

Issue #1959, opened 2016-09-19
`cache()` behaves unexpectedly with non-terminating sources


chirayuk:
**RxJS version:** v5.0.0-beta.12

**Code to reproduce:**

```
var t$ = Rx.Observable.timer(0, 10).cache(1);
t$.take(5).subscribe();
// Wait until the previous subscription has ended.
// NOTE: t$ has NOT completed.
// So the following log should print one item from the cache and continue reading the rest from the timer.
// It should definitely NOT complete. But it does—prints a single item from the cache and completes.
setTimeout(() => t$.subscribe((x) => console.log(x)), 100);
```

**Expected behavior:**
Never ending console logs.

**Actual behavior:**
Single console log.

**Additional information:**
It's surprising behavior and undocumented (unless you read the code.)  It definitely bit me and I'd consider it a bug.

blesh:
Looks like the issue is that when `take()` unsubscribes, it reduces the [ref count to zero, and kills the source subscription](https://github.com/ReactiveX/rxjs/blob/646b38ddad9c8123bc5cc8db2008f5dddfc0a61d/src/operator/cache.ts#L54-L55)

Doesn't seem like a hard fix... just requires a little thought.

blesh:
In fact, I think it requires thought to figure out if this is a bug, or just counter-intuitive behavior.

blesh:
Thinking about it, it seems like this isn't a bug, so much as it is counter-intuitive behavior.

consider:

```js
const source = Observable.interval(1000).cache(1)
```

That part alone is a really strange use of `cache`, most likely... using it on an un-ending observable.

In the above case we're saying `source` is a ref-counted, hot observable. So it's not going to do anything until the first subscription to it. Since it's refCounted, we're going to stop the source when the subscriber count reaches zero.

If you use `take(5)` on it, and it completes before there are other subscribers to `source`, it's going to unsubscribe from the interval we're caching from.


### Questions

Do we want a "refCounting" behavior with `cache`?  Or shall we just say it goes until it `complete`s? What are the implications of that on other operations that could be performed? How does that affect our primary use case (which is mostly caching AJAX, I think)

blesh:
cc/ @jayphelps @trxcllnt @mattpodwysocki @jeffbcross @robwormald @kwonoj @staltz

jayphelps:
[...]

PR #2012, opened 2016-10-06
feat(cache): remove `cache` operator

blesh:
Because of all of the confusion around the `cache` operator, this removes the operator
from RxJS ahead of the 5.0.0 release. This is done for a few reasons\:

1. The name `cache` implies it supports all sorts of caching strategies, when this is pretty limited
2. Behavior around unending and hot observables is hard to reason about
3. Once we have a properly designed `cache` operator we can add it at a later time and it will not be a breaking change


blesh:
related #839 #1718 #2006 #1417 #1405 and many more I'm sure.

blesh:
cc/ @kwonoj @staltz @jayphelps @trxcllnt

blesh:
more related: #2010 #1959

jayphelps:
I'm very very 👍 , but since there's a chance this is controversial, going to wait for others to chime in.

kwonoj:
I'm 👍  too. If it's debatable about behavior need to be settled down, better to make it once instead of changing behavior over different release which can cause confusion.

trxcllnt:
LGTM

staltz:
LGTM

"Cache invalidation" :trollface:

tinchou:
Hey guys. I can't seem to find a simple alternative to `.cache()`. I asked this on [SO](http://stackoverflow.com/questions/41176489/rxjs-5-and-alternatives-to-the-cache-operator), but maybe you can help me out too.

Thanks!

zixia:
It takes me some time to finger out that the `cache` method has gone in v5.0.

I think the better way to _remove_ `cache` is to show a DEPRECATED message when `cache` method is called, and keep it for compatible from earlier code bases.

jayphelps:
@zixia sorry about the confusion. We made the change back in October when v5 was in beta, where we do not make any guarantees about API stability and deprecation cycles. Some of us have talked about possibly doing deprecation warnings if we do similar sweeping removals for future major versions e.g. v6, v7, etc. but we haven't commited to anything cause we don't yet have any such changes. The only breaking change we have queued up is #2174 which slightly changes the behavior of the buffer operator--it's a bugfix, but since it's a _breaking_ bugfix we're holding off on releasing it until v6.

PR #2015, opened 2016-10-09
style(docs): improve readability of publish_docs

igncp:
Hi,

Just saw this bash script (the only one in the repo) and thought to improve its readability, similar as it was initially but keeping the ampersands. Hope it helps.

Cheers


kwonoj:
LGTM

PR #2017, opened 2016-10-10
feat(package.json): update TypeScript to v2.0

blesh:
BREAKING CHANGE: TypeScript definitions are now for TS 2.0 and higher

Even if we use getter for class, they are marked with `readonly` properties
in d.ts.

this is a "fixed" version of #1968, retaining @saneyuki's work, just rebasing and updating the commit message.


kwonoj:
LGTM

Issue #2016, opened 2016-10-10
Include es6 type definition via TypeScript compiler or 3rd party module


kwonoj:
**RxJS version:**
.Beta.12

**Additional information:**
related to https://github.com/ReactiveX/rxjs/issues/2011#issuecomment-252418870

Lot of confusion from TypeScript user is RxJS rely on some of 3rd party type definition to support ES2015 signatures and they have to install it by themselves. From TS@2.0, there's newly introduced mechanism to supply 3rd party type definition by installing published npm packages, like `es6-shim`(https://www.npmjs.com/package/@types/es6-shim) . 

RxJS can specify them as dependency to remove those issues completely, or set loosely as peerDependency to consumer knows it's requires, feel freely choose between @types package or their own (i.e, typings).

I believe `redux-observable` is aim for similar, /cc @jayphelps as well for opinions.

jayphelps:
I was playing with the new @types and man this is such a confusing situation DefinitelyTyped vs typings vs types vs lib provides it already.

Can you share what changes you're recommending specifically? e.g. the package.json changes?

kwonoj:
I also tried this bit yesterday and found it's bit confusing to me as well, since it isn't super clear origin of @types package comes in, and possible conflict based on user configuration, etcs.

I don't have concrete configurations at this moment - but current suggestion is set as peerDependency to not cause trouble to existing configs but let user know they need to supply 3rd party type definition. Consumer can freely choose to supply it via installing peerDependency, or use their own existing config like `typings`. (Note: this only applies npm@3 and above which doesn't install peerDependency explicitly)

kwonoj:
in short, proposal will looks like

```js
"peerDependencies": {
    "@types/es6-shim": "version we want"
}
```

david-driscoll:
Do we need third part shims to be a part of the library itself?  With Ts2.0 we can simply set the `lib` appropriately, and then anyone consuming downstream just has to have the same or similar lib items.

kwonoj:
@david-driscoll you're referring https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#including-built-in-type-declarations-with---lib , is that correct? Yes, this seems neat and more feasible (if it works as expected)

kwonoj:
Changed subj.

david-driscoll:
[...]

Issue #1992, opened 2016-09-30
Consuming rxjs-es using webpack and babel causes invalid code while resolving imports


jh3141:
**RxJS version:** 5.0.0 beta 12

**Code to reproduce:**
In a project using webpack and babel to translate ES6 to traditional javascript:
* Add an 'include' directive to the babel loader to allow babel to translate the rxjs-es module
* Add `import Rx from 'rxjs-es';` to the main source file of the project
 
**Expected behavior:**
rxjs-es is transpiled down to standard javascript and can be used in the project

**Actual behavior:**
The transpiled code contains the following chunk of invalid javascript that causes a runtime error during initialisation:

```
/*!***********************************!*\
  !*** ../~/rxjs-es/util/assign.js ***!
  \***********************************/
/***/ function(module, exports, __webpack_require__) {

	'use strict';
	
	Object.defineProperty(exports, "__esModule", {
	    value: true
	});
	exports.assign = undefined;
	
	var _root = __webpack_require__(/*! ./root */ 417);
	
	var Object = _root.root.Object;

```

This does not work because the existence of the variable `Object` shadows the standard variable `Object` which the import code expects to be able to access, but because the definition is later in the file its value is not available at the time the import code needs it.  As `Object` is not exported from this module, renaming it to something else would be a suitable workaround for this issue.

**Additional information:**

See also this bug filed against webpack: https://github.com/webpack/webpack/issues/2618

While it does seem the problem is essentially webpacks, it doesn't seem they're making much progresss on solving a potentially difficult issue, whereas it would be trivial to work around it here.

kwonoj:
Related to https://github.com/ReactiveX/rxjs/issues/1671 . Actually there's discussion to not publish sole ES2015 package, which will remove these problems eventually (I hope).

jayphelps:
We've decided to stop publishing rxjs-es for now per #2019 because it's caused more problems than it solved. We'll continue considering similar solutions for the future!

Issue #1954, opened 2016-09-18
rxjs-es cannot be bundled by rollup or webpack


mistyharsh:
I am trying to use ES6 version of the rxjs by following instructions written in docs http://reactivex.io/rxjs/manual/installation.html:

``` bash
npm install rxjs-es --save
```

and using in my file as:

``` javascript

import Rx from 'rxjs/Rx';
Rx.Observable.of(1,2,3)
```

This throws the error: `Error: Could not resolve 'rxjs/Rx.js'`. So i modified my import to make it work:

``` javascript
import * as Rx from 'rxjs-es/Rx.js';
```

Now this generates a bundle but I see the error on browser console: `Uncaught ReferenceError: OuterSubscriber is not defined`
After debugging a bit I realized that bundle is not being generated the way it should be. However If I modify my imports to pick only what I need then it works:

``` javascript
import {Observable} from 'rxjs-es/Observable';
import 'rxjs-es/add/observable/of';
import 'rxjs-es/add/operator/map';

let observable = Observable.of(1,2,3);
```

I looked through old commits, issues and stackoverflow but could not find anything substantial. My rollup.js script looks like this:

``` javascript
gulp.task('jsbundle', () => {

    return rollup({
        entry: './src/commons/js/main.js',
        treeshake: true,
        plugins: [
            nodeResolve({
                module: true,
                jsnext: true,
                main: true
            }),
        ]
    }).then(bundle => {
        return bundle.write({
            format: 'iife',
            dest: './dist/commons/js/main.js'
        });
    });
});
```

Then I switched to commonjs package using `npm install rxjs` and tried to consume this using rollup via `rollup-plugin-commonjs` plugin. Then it works perfectly.

To further investigate, I did the same with **Webpack**. And I found the similar result. If I try to import entire core set of `Rx` functionality, then **rxjs-es** package doesn't work.

I am still not able to entirely figure out exact cause of the problem. I am willing to raise a PR if I can find out exact problem.

kwonoj:
as RxJS no longer publishes `rxjs-es` and recommend to use `rxjs` or `@reactivex/rxjs`, I think this issue can be closed by recommend to use CJS package published.

PR #2019, opened 2016-10-11
feat(ES2015): stop publishing `rxjs-es`

### LINKED ISSUE #1671: Stop publishing rxjs-es. Move ES6 build to output .mjs files alongside the CJS .js files.
In speaking with @robwormald the Angular folks are forced to swap between `rxjs` and `rxjs-es` during their build to enable proper tree-shaking. This is probably less than ideal. If all of the es6 files were under the same directories as `.mjs` files, then it would be as easy as configuring the bundler to look at that file extension to use ES6 modules.

This would mean that the `rxjs-es` build would likely not have a use anymore.

cc/ @IgorMinar @jeffbcross @bradlygreen

blesh:
BREAKING CHANGE: `rxjs-es` is no longer being published
BREAKING CHANGE: `@reactivex/rxjs` no longer has `/dist/es6` output

related #2016
related #1992
closes #1671
related #1954

### DIFF
--- a/.make-packages.js
+++ b/.make-packages.js
@@ -17,11 +17,6 @@ var cjsPkg = Object.assign({}, pkg, {
   main: 'Rx.js',
   typings: 'Rx.d.ts'
 });
-var es6Pkg = Object.assign({}, cjsPkg, {
-  name: 'rxjs-es',
-  main: 'Rx.js',
-  typings: 'Rx.d.ts'
-});
 
 fs.writeFileSync('dist/cjs/package.json', JSON.stringify(cjsPkg, null, 2));
 fs.writeFileSync('dist/cjs/LICENSE.txt', fs.readFileSync('./LICENSE.txt').toString());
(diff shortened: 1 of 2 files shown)


jayphelps:
LGTM

kwonoj:
Manual compilation via npm script will still be supported? (*_es6 scripts other than compile for docs)

jayphelps:
@kwonoj the plumbing around building the ES6 builds are still there, right now just used for docs. The es6 builds will no longer be included in `@reactivex/rxjs`  and `rxjs-es` will no longer get new publishes and we'll mark it as deprecated.

jayphelps:
@kwonoj I may be misunderstanding your question tho. lmk.

kwonoj:
@jayphelps yes, I'm aware of doc-related scripts and question is es6 build scripts are not being used for those. It's not a huge issue since it can be updated anytime.

jayphelps:
@kwonoj ahhh got it now. It has a minorly different npm script.

superherointj:
As 'rxjs-es' module is deprecated. If trying to:
import { Observable } from 'rxjs/Observable'
Gets error: node_modules\rxjs\Rx.js does not export Observable"
How should I be importing Observables using ES6 modules now?

PR #2021, opened 2016-10-11
fix(multicast): fix a bug that caused multicast to omit messages after termination

trxcllnt:
ConnectableObservable as the multicast selector function source argument would omit events after source termination when used with historic Subjects.

Simple repro example:

```js
Observable.of(1, 2).multicast(
  ( ) => new ReplaySubject(1),
  (x) => x.concat(x.takeLast(1)))
  .toArray()
  .subscribe((arr) => {
    if (arr.toString() !== `1,2,2`) {
      throw 'multicast missed the `takeLast(1)` event';
    }
  });
```


staltz:
LGTM

PR #2022, opened 2016-10-11
fix(operator/do): fix typings

Brooooooklyn:
**Description:**
```ts
Observable.of('abc')
  .do(v => {
    // type of v is { } here
    console.log(v.substr(2))
  })
```


kwonoj:
This looks ok to me, @david-driscoll is this caused by script running? or current state is expected?

blesh:
This looks fine to me, pending @david-driscoll's approval (because it's typings related)

src/operator/do.ts
```diff
@@ -48,7 +48,7 @@ import { TeardownLogic } from '../Subscription';
  * @owner Observable
  */
 /* tslint:disable:max-line-length */
-export function _do<T>(next: (x: T) => void, error?: (e: any) => void, complete?: (this: Observable<T>) => void): Observable<T>;
+export function _do<T>(this: Observable<T>, next: (x: T) => void, error?: (e: any) => void, complete?: (this: Observable<T>) => void): Observable<T>;
```
david-driscoll: definitely something I missed with the script.  This needs to be changed slightly, note the additional `this: Observable<T>` .

`complete?: (this: Observable<T>) => void` should be changed back to `complete?: () => void`

Brooooooklyn: fixed

david-driscoll:
LGTM

kwonoj:
LGTM

PR #2023, opened 2016-10-11
feat(groupBy): Adds subjectSelector argument to groupBy

trxcllnt:
Adds a `subjectSelector` argument to `groupBy`, allowing consumers to customize the messaging semantics of the inner GroupedObservables.

Sometimes it isn't possible to immediately subscribe to groupBy's inner GroupedObservables as they're emitted, and you need to replay some number of events between when the GroupedObservable was created and when you subscribe.

For example, one encounters this problem writing multitouch gestures. Each Gesture has a set of constraints to consider the gesture started, but then needs access to the historic events once those constraints have been satisfied.


blesh:
This appears to be a non-breaking change.

I *think* I can see the utility of this, but would you might adding some use cases in a comment on this PR so everyone can better understand this addition?

Also, we'll need to update the documentation. Ideally with an additional example showing using this argument for something useful, I guess.

cc/ @staltz @mattpodwysocki what are your thoughts on this addition?

src/operator/groupBy.ts
```diff
@@ -46,15 +48,15 @@ export interface RefCountSubscription {
 }
 
 class GroupByOperator<T, K, R> implements Operator<T, GroupedObservable<K, R>> {
-  constructor(public source: Observable<T>,
-              private keySelector: (value: T) => K,
+  constructor(private keySelector: (value: T) => K,
               private elementSelector?: ((value: T) => R) | void,
-              private durationSelector?: (grouped: GroupedObservable<K, R>) => Observable<any>) {
+              private durationSelector?: (grouped: GroupedObservable<K, R>) => Observable<any>,
+              private subjectSelector?: () => Subject<R>) {
```
mattpodwysocki: I'd rename this to `subjectFactory` since it's not taking any input into this function

mattpodwysocki:
@blesh it's an interesting choice here, although `subjectSelector` should be `subjectFactory` as it's not taking any input and always returning some sort of `Subject`.  The only thing about this that intrigues me is changing it probably from a normal `Subject` to `ReplaySubject` in that it persists data instead of a regular `Subject` which is lossy.  

@trxcllnt can you give us a concrete use case with working code as to why this is a good change?

paulpdaniels:
Would it make sense to also add this for the `window` operators as well? There are certain cases where this sort of stuff happens if you apply various flattening techniques after windowing.

mattpodwysocki:
@paulpdaniels yes, it's a slippery slope here once we allow you to specify the guts as it were of a given Observable of Observables.

blesh:
Slippery Slope

staltz:
The only thing I don't like about this is that it makes groupBy a method with 5 arguments. But otherwise I'm favorable and can see its utility, albeit rare.

src/operator/groupBy.ts
trxcllnt: Agreed

trxcllnt:
@mattpodwysocki I admit, the use-cases for this are relatively rare, which is why I like the subjectFactory in the last arg position.

The problem that spurred me to add this option was around multitouch gesture recognition:

```es6
Observable
  .fromEvent(domElement, "touchstart")
  .groupBy(getIdentifier)
  .take(2)
  .map((starts) => starts
    .mergeMap(start => Observable
      .fromEvent(window, "touchmove")
      .filter(isIdentifier(start.identifier))
      .startWith(start)
    )
  )
  .combineAll()
```

By the time the second touchstart event fires, the first inner group has already fired, so the first Observable in the combineAll never emits.

blesh:
Thinking more on this... I feel like adding this leaks out the underlying abstraction. People don't need to know we use a Subject under the hood. I say this because it's possible that at some later point we decide to implement a groupBy that doesn't use a Subject. Perhaps for performance reasons. If we add this argument we can't really do that because we'll be stuck with using Subjects... or at least we can't do that without some crazy conditionals.

trxcllnt:
@blesh I see what you're saying, but since GroupedObservables are hot and multicast to multiple subscribers, how else could they be implemented? It seems any performance improvements that could be realized from a specialized "GroupedObservableSubject" probably can and should be applied to Subjects generally.

One description of `groupBy` is that it transforms an Observable of values into an Observable of inner hot Observables. From that angle, it makes sense to allow some sort of control over the observability of inner groups.

The only other solution I've found to this problem is adding `autoConnect` to ConnectableObservable, which would immediately connect the ConnectableObservable to the GroupedObservable. I'm less inclined to want to go that route because, a. I'm certain that once autoConnect goes in, it's going to get a lot of beginners into the sort of trouble they aren't equipped to handle, and b. the double-multicasting offends my aesthetic sensibilities.

jayphelps:
We decided to ship this :shipit: at the meeting today.

Issue #1837, opened 2016-07-19
Use new features introduced from TypeScript Compiler v2.0


saneyuki:
After upgrade tsc to 2.0, we can use more compiler options to make our code more safety, cleanly, and massive.

## Examples:

- `strictNullChecks`compiler option.
- `noUnusedLocals` compiler option.
- `noImplicitThis` compiler option.
- `noUnusedParameters` compiler option.
- Use npm's @types packages to manage d.ts files.
- Use `readonly` modifiers.
- Use `never` types.
- Glob support in tsconfig.json
- And mores...

## Official release note:

- [What's new in TypeScript · Microsoft/TypeScript Wiki](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#typescript-20)

saneyuki:
@kwonoj could you change labels of this issues? I don't have the permission to edit it.

saneyuki:
@kwonoj thank you.

kwonoj:
related issue : https://github.com/ReactiveX/rxjs/issues/1915

maybe non-nullable type's also needed.

saneyuki:
We gathered related issues to this once, but I think this is too generic "meta" bug. We have been landed upgrading tsc 2.0. So we should considers for each issues.

Issue #1972, opened 2016-09-23
chore(package.json): Use npm's @types packages to manage d.ts files


saneyuki:
- We can remove the dependency for `typings` package.
- And it will simplify a dependency management.

saneyuki:
Duplicated #1837

saneyuki:
I'll reopen as a separated tracking issue.

kwonoj:
I believe this issue is already resolved in latest master.

saneyuki:
@kwonoj Oops, I missed it. Thank you!

Issue #1971, opened 2016-09-23
feat(TypeScript): enable 'noImplicitThis' options


saneyuki:
- This intent to make `this` type explicitly.
  - This would make our codes more type safe with a compiler enforcement.

saneyuki:
Duplicated #1837

saneyuki:
Relate? https://github.com/ReactiveX/rxjs/pull/1978

kwonoj:
I'm bit down for this. I've enabled this option for few codebases, and found it's sometimes not easy to infer correct type of `this` and eventually had to go down with explicitly `any` casting still. Problem of `this` is it's reserved keyword, you can't just cast simply like `this as any`. For example, this is actual erroring code in Rx codebase after enabling this option - which just need additional boilerplate to casting.

```js
//'this' implicitly has type 'any' because it does not have a type annotation.
state.context.closeAction = (<any>this).schedule(state, state.bufferTimeSpan); 
````

**Comment by saneyuki** [comment]
@kwonoj 

We can specify `this` as 1st argument of a signature like `(this: A) => void` or the following:

```ts
function dispatchBufferTimeSpanOnly<T>(this: Action<T>, state: any) {
    ...
    // this `this` is `Action<T>`
    state.context.closeAction = this.schedule(state, state.bufferTimeSpan);
}
```

If we land https://github.com/ReactiveX/rxjs/pull/2078, we can checks validness of it statically.

kwonoj:
> We can specify this as 1st argument of a signature like (this: A) => void or the ...

Yes I am aware of those. Question in here is is it worth to do it? as our codebase shows, when we use `this`, we are aware its usage and explicitly cast to `any` anyway. So enabling this option only requires additional boilerplate to cast into any since most of usage those not asks explicit type inference. Above snippet can be addressed via #2078  as you mentioned though, it is about general of accessing `this` in certain context.

saneyuki:
@kwonoj 



- I think this would  clarify `this` type for callbacks (e.g. `Observable<T>`'s constructor, `Scheduler.schedule()`). The signature express very strongly what this `this`. 
   - This increase our code's readability. It is nice for us and users.
- And, there is no (or less) cons as my feeling by that I'm trying to enable this option in my private branch.
   - In almost case, we need not add additional boilerplate. We need specify `this` type for only a function which uses implicit this type. If a function does not use `this`, then we don't have to specify `this` type.
[...]

Issue #1970, opened 2016-09-23
enable 'noUnusedParameters' options


saneyuki:
This would be useful to clean up our codes.

saneyuki:
Duplicated #1837

saneyuki:
We tried to enable once at #2024, but then we could not make a consensus to land it because we'd like to focus to ship v5 stable and there still be inconsistent the naming conventions. (see also [the meeting note](https://github.com/ReactiveX/rxjs-core-notes/blob/master/2016-10/october-24.md#outstanding-prs)).

So I'll reopen this as a tracking bug for the future.

kwonoj:
Note TSlint now deprecated `no-unused-variable` and consider to remove it in a further version.

https://github.com/palantir/tslint/issues/1617

benlesh:
Closing as stale.

PR #2024, opened 2016-10-11
style(linting): enable tsc's 'noUnusedParameters' compiler option

saneyuki:
**Description:**

* Make signatures for `public`one, `protected` member, or exported function only suppress compile error.
   * They are opened interface. We cannot cause breaking change.
* Change (remove) needless signatures of `private` member or private functions.
   * They are not opened interface. This change will not breaking change.
* Relate https://github.com/ReactiveX/rxjs/issues/1837


saneyuki:
umm, why is the coverage down?

kwonoj:
I saw coverage confused time to time. By given numbers of coverage, I think this is same case.

kwonoj:
Change itself looks ok to me.

blesh:
I'm not sure about this change.

1. What issue is it resolving? It's touching a LOT of files and I'm not sure what the gain is for the PR as a whole.

2. Sorting the compiler options, while helpful, will obfuscate changes visible in `git blame` on that config file. I understand the spirit, but I'm not sure about the result.

blesh:
Also, this should probably not be listed as a `feat(TypeScript)`, because it's not really a feature change for the library is it? It seems like it's just a `style(linting)` sort of change?

kwonoj:
> It seems like it's just a style(linting) sort of change

Yes, I'd agree this is more about style changes to explicitly mark unused variables in function signatures, etcs.

jayphelps:
My initial feeling is that a lot of these changes like

```diff
-    const result = source.retryWhen((errors: any) => notifier);
+    const result = source.retryWhen((_: any) => notifier);
```

Make the params even more frivolous, so wouldn't we just want to remove them instead of using the underscore unused convention? Thoughts?

kwonoj:
> so wouldn't we just want to remove them instead of using the underscore unused convention? Thoughts?

In my case I prefer to have param like `(_value, key) => { do(key); }` which shows signature supplies param, but it's not used inside of functions. My vote is let function have all param signatures (even it makes to disable `noUnusedParameters`) to have clarity regarding type definitions.

blesh:
> In my case I prefer to have param like (_value, key) => { do(key); } which shows signature supplies param, but it's not used inside of functions

Linting shouldn't complain about that anyhow.

jayphelps:
@kwonoj yeah I was referring to all the numerous places where we don't use _any_ of the the params at all. In your `(_value, key) => { do(key); }` example, any reasonable linter shouldn't complain, no matter what you name your first arg, because you use the second one. Dunno if TS is a reasonable linter haha but I imagine it is.

kwonoj:
@jayphelps yes, tslint and typescript compiler both should work with above snippet example, I have code working with ts@2 (https://github.com/kwonoj/rxjs-testscheduler-compat/blob/master/src/VirtualPromise.ts#L80)

saneyuki:
@jayphelps 

> Make the params even more frivolous, so wouldn't we just want to remove them instead of using the underscore unused convention? Thoughts?

As my thought of when I created this patch, it is just same with @kwonoj 's comment:

> My vote is let function have all param signatures (even it makes to disable noUnusedParameters) to have clarity regarding type definitions.

I prefer to keep their signatures explicitly, and this patch was created with simply grep+replace.

saneyuki:
@blesh 

> What issue is it resolving? It's touching a LOT of files and I'm not sure what the gain is for the PR as a whole.

This marks unused parameters explicitly. This will notify us that there are some cleanup chance which may be related to large refactoring. And we can suppress the error with adding `_` prefix  like this patch. I don't think this causes a trouble for our coding.

------------------------------------

> Sorting the compiler options, while helpful, will obfuscate changes visible in git blame on that config file. I understand the spirit, but I'm not sure about the result.

Okay. I'll remove its patch.

------------------------------------

> Also, this should probably not be listed as a feat(TypeScript), because it's not really a feature change for the library is it? It seems like it's just a style(linting) sort of change?

Okay. I'll change commit logs.

kwonoj:
Let me try to summarize to get some clarification, as well as to get agreement for this changes.

*1. current*
`(value: any, key: any) => { do(key); } `

*2. mark unused, leave param detail as-is*
`(_value: any, key: any) => { do(key); } `

*3. proposal in this pr, removes param detail*
`(_: any, key: any) => { do(key); } `

*4. remove param from signature*
`(key: any) => { do(key); } `

my vote is 2, if not available falls back to 1 (even if it does not allow to enable compiler option).

saneyuki:
@kwonoj 

> 2.mark unused, leave param detail as-is
> ```
> (_value: any, key: any) => { do(key); }
> ```

Ah, nice. I'll change to sort the style with 2. 

> 3.proposal in this pr, removes param detail
> ```
> (_: any, key: any) => { do(key); }
> ```

I may use this style in some case in other language, but in TypeScript in this project, I don't have strong opinion to use style 3.

saneyuki:
```diff
-  notifyComplete(unused: Subscriber<R>): void {
+  notifyComplete(_: Subscriber<R>): void {
```

...But this case, I'm favor of to use `_` to express "unused". It's a self-evident.

saneyuki:
> ...But this case, I'm favor of to use _ to express "unused". It's a self-evident.

Sorry. This is my wrong. They should use the original interface's argument name.

jayphelps:
I think we should leave everything as-is and leave it up to the discretion of the person writing a given block of code. I have different opinions for apps, but for libs like this we can't easily apply the rule across the entire codebase. Sometimes it makes sense to have the unused params because they will show up in IDE suggestion signatures--and in that case they _should not_ have _underscore _name because that implies they are private (by JS convention). Other times, in completely internal code, it make make sense to leave the params named for future documentation cases. They may be unused right now, but when I come back it's nice to see them and gives me immediate context around what this function is--especially when they're anonymous functions. And  certainly there are times where it's very obvious what they would be and having them at all is potentially confusing.

So this all leads me to believe _"it depends"_. So I think it's OK to just leave it to discretion.

saneyuki:
@jayphelps

> and in that case they should not have _underscore _name because that implies they are private (by JS convention)
> Other times, in completely internal code, it make make sense to leave the params named for future documentation cases.
> https://github.com/ReactiveX/rxjs/pull/2024#issuecomment-253119150

Yes. I agree to it completely. _We should not use _ prefix for public interface_.
So I don't change a public signature (e.g. `Observable.map()` or other public instance method of key fundamental components) in pull request, and I would not like to do it.

I also think that `MergeMapSubscriber` or other internal classes are exported to public and user can import to their codes, but they are private classes _factually_. Their classes are undocumented, I don't think many many user would use them as tier 1 public APIs. So I think we can change their classes signatures with `_` prefix

saneyuki:
> Ah, nice. I'll change to sort the style with 2. 

done.

blesh:
I'm sorry, @saneyuki. I think a LOT of work and thought went into this PR, but I'm just not on-board with the changes. 

1. It touches a lot of lines in a lot of files and doesn't fix any issues. Thus adding some indirection when using tools like `git blame` to figure out what's going on.
2. The outcome might be less ergonomic for other library authors.

I love the spirit and the hard work. I'd like to reward that sort of thing by merging it, however, I don't think I want to merge this particular PR, because I don't agree with the outcome.

In the future, it might be better to propose changes in an *Issue* first, and let the community discuss it before you put in all this hard work. I don't want any hard feelings, and I don't want to 
waste anyone's time.

saneyuki:
I fixed needless changing line feed. Sorry.

saneyuki:
@blesh 

> It touches a lot of lines in a lot of files and doesn't fix any issues. Thus adding some indirection when using tools like git blame to figure out what's going on.

If you said about my patch was changing some files' linefeed, I'm sorry my faults. I have been fixed them.

`git blame` is important, but we have a complete history in our repository and we're using GitHub that provides pretty nice UI to track a file not only `git blame`. I don't think that it would not be a serious problem about `git blame`.

I think we should build a more solid codes rather than clean `git blame`, and I hope it is valuable.

--------------------------

> The outcome might be less ergonomic for other library authors.

I just think this `noUnusedParameters` compiler option is just a builtin rules which is like tslint's `no-unused-variable` error or others. We can escape from it easily with adding `_` prefix. There is no problem. If we says about less ergonomic with lint errors, we should stop to use tslint to build a solid code.

And we enables tslint's `no-unused-variable` rules. This pull request do a same things on compiler's built feature.

--------------------------

> In the future, it might be better to propose changes in an Issue first, and let the community discuss it before you put in all this hard work. I don't want any hard feelings, and I don't want to
waste anyone's time.

I'm apologize that I missed to note about the issue https://github.com/ReactiveX/rxjs/issues/1837. at the time to open this. But there is no any objection, I created this pull request.

jayphelps:
@saneyuki Going to close for now and we may revisit later as the project progresses. Like @blesh said, we do truly appreciate the effort that went into this. We want you to continue to feel welcome to contribute! Regarding #1837, we did see it, but it included many suggestions and at the time we hadn't decided when we were going to upgrade to tsc 2.0 so we didn't give any input because those things are just lower on our radar--we're pretty heads down focusing on trying to get 1.0-final shipped ASAP. So tl;dr no objection isn't necessarily a consensus of approval.

We're going to be meeting much more frequently to review outstanding PRs and issues, with the intent to get better at preventing these sorts of things. Bare with us as we learn and grow together! 👍

saneyuki:
@jayphelps 

> Regarding #1837, we did see it, but it included many suggestions and at the time we hadn't decided when we were going to upgrade to tsc 2.0 so we didn't give any input because those things are just lower on our radar

As the history, I had filed each of their new features as separated issues (#1969, #1972, #1971, #1970), but I gathered them into #1837 because @kwonoj says they should be gathered into the single issue.

If we need more discussion at an issue, not at here, I’ll reopen their separate issues.

---------------------------------------------------------------------------------------------------

> we're pretty heads down focusing on trying to get 1.0-final shipped ASAP.

I agree that this issue would not be a release blocker to ship final 1.0-stable release. I agree we don't merge this and we don't merge or review this until shipping 1.0-final stable release too.

But it is not a reason to reject this pull request and to comment "file an issue at first" which is not listed in https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md. AT NOW, there is no rules about "file an issue before opening pull request". Should we note about “file an issue at first before opening a new pull request” to a contribution guide?

And so I don't close this pull request by my hand without to explain why we would not like to keep open this.

---------------------------------------------------------------------------------------------------

> So tl;dr no objection isn't necessarily a consensus of approval.

Do you mean "we must hear from all humans living on the Earth"? If you say so, we should build a new proposal process. AT NOW, we don't have concrete proposal process. I thought as a community contributor regards that there is no concrete process and we use implicit devolution if someone don't say any opinions. If my thought was wrong, please write your thought about proposal consensus process to a opened site.

---------------------------------------------------------------------------------

> We're going to be meeting much more frequently to review outstanding PRs and issues, with the intent to get better at preventing these sorts of things. Bare with us as we learn and grow together!

Yes. __I'd also like to continue to effort to grow this project together__. To achieve it, I’m going to discuss deeply.

Will you publish a public meeting note which are not published now? Openness very very important for contributors who would like to contribute a code. Private chatting & no publishing meeting notes are strong barrier for contributors. Could you publish them? I’m looking forward to see them!

jayphelps:
@saneyuki we do indeed plan to start publishing our meeting notes. We're working out where to put them and who'll record them. 👍

jayphelps:
@saneyuki Hey, we discussed this PR at our meeting today. ([meeting notes](https://github.com/ReactiveX/rxjs-core-notes/blob/master/2016-10/october-24.md#outstanding-prs))

We couldn't reach consensus on whether we want to impose this linting rule with some in favor and others against. Because we're trying to focus on shipping v5, we decided to keep the existing behavior for now.

We really do appreciate you taking the time! Having PRs closed can suck sometimes, but it does happen to all of us. We're also learning from this and going to discuss whether or not we should formalize a RFC process or at least document our suggestions better. 👍

saneyuki:
@jayphelps 

> We couldn't reach consensus on whether we want to impose this linting rule with some in favor and others against. Because we're trying to focus on shipping v5, we decided to keep the existing behavior for now.

Okay. I understand why we close this now.
I'll re-open https://github.com/ReactiveX/rxjs/issues/1970 as the tracking to `noUnusedParameters`.

PR #2025, opened 2016-10-11
fix(Observable/Ajax): mount properties to origin readystatechange fn

Brooooooklyn:
**Description:**
xhr.onreadystatechange property may be monkey patched by other library such as `zone.js`.
It would cause AjaxObservable could not find `subscriber`, `progressSubscriber` and `request` property on xhr.onreadystatechange.

it's a simple example here: http://plnkr.co/edit/oX1v8tTjTog1T57Z0JnI?p=preview


blesh:
@Brooooooklyn ... I'm struggling to understand the issue. I *think* I get it... but can you create a test case as part of this PR that definitively shows the issue this is trying to correct? You don't need to use zones, maybe just monkey patch the handler or something?

It seems if there's a problem like this, the alternative solution would be to use the `addEventHandler` means of registering handlers.

jayphelps:
@blesh Indeed the issue is that zone.js does not respect the identity of the function you assign to the event callbacks. [It wraps the function here](https://github.com/angular/zone.js/blob/737f8d834a95dc6f24d57d2055ca99d51694cb46/dist/zone.js#L640-L661) so a new function created internally is the one you actually receive when we `xhr.onreadystatechange.something = 123` etc.

IMO is solution is acceptable and effectively the same we'd need to do if we used `xhr.addEventListener` -- that is, we need to assign the function to a local scope var first, then use that local scope var to assign the props, independently of setting up the listener logic.

Brooooooklyn:
@blesh I've added some test in a separate commit, could you please review it.

blesh:
Oh. I understand. Not sure why it wasn't sinking in the first time I looked at it.

jayphelps:
LGTM

PR #2027, opened 2016-10-11
feat(typescript): remove dependency to 3rd party es2015 definition

### LINKED ISSUE #2016: Include es6 type definition via TypeScript compiler or 3rd party module

kwonoj:
**Description:**
This PR enables `lib` option in TypeScript compiler 2.0 to include es2015 native type definitions for cjs build, instead of supplying 3rd party type definition.

closes #2016


kwonoj:
This seems working by trying out local package generated via `npm pack`, not sure about additional verification approaches.

/cc @david-driscoll , @jayphelps for visibility.

david-driscoll:
Travis stuck?  Should I restart the build?

kwonoj:
yes, let's retrigger. Not sure what happened on travis. For documentation, I think release note woild be suffecient since supplying 3rd party typing was never officially documented. After this change, theoritically any consumer would not need 3rd party definition due to RxJS.

blesh:
LGTM ... merge if you're ready, @david-driscoll

Issue #2029, opened 2016-10-13
[rc.1] do operator breaks types. / Subjects need asObservable to keep types.


ovrmrw:
**RxJS version:** 5.0.0-rc.1

**Code to reproduce:** [https://github.com/ovrmrw/rxjs-marble-tests/blob/master/spec/forIssues/rc1-unexpected-behavior.ts](https://github.com/ovrmrw/rxjs-marble-tests/blob/master/spec/forIssues/rc1-unexpected-behavior.ts)

**Expected behavior:** `asObservable` is not needed for Subjects just to keep variable types. `do` operator should not break variable types.

**Actual behavior:** Subjects without `asObservable` all turn variable types into "any". `do` operator turns variable types into "{}".

**Additional information:** TypeScript 2.0.3 , VSCode 1.6.0

david-driscoll:
This should be fixed with #2022

ovrmrw:
`do` operator is ok by [#2222](https://github.com/ReactiveX/rxjs/pull/2022).
But an issue that Subject needs `asObservable` to keep types is still a problem.

ovrmrw:
I confirmed to be solved the Subject issue. Great work! Should I close this issue?

kwonoj:
Closing via #2035 .

PR #2035, opened 2016-10-14
fix(typings): fixed Subject<T>.lift to have the same shape as Observable<T>.lift

david-driscoll:
**Description:**
This addresses the problem described in #2029.

`Observable<T>` implements `lift<R>(...)`
`Subject<T>` implements `lift<T, R>(...)`

The extra generic type argument `T`, causes type inference to break in horrible ways.

#2029

### DIFF
--- a/src/Subject.ts
+++ b/src/Subject.ts
@@ -43,9 +43,9 @@ export class Subject<T> extends Observable<T> implements ISubscription {
     return new AnonymousSubject<T>(destination, source);
   };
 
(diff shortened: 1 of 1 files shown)


david-driscoll:
I did investigate a way to make a compile time test that would fail if this happens again... but it's really hard to make the compiler fail when it gives you an `any` type.

kwonoj:
LGTM

PR #2036, opened 2016-10-14
feat(operators): add `operators` operator

blesh:
This adds an operator call `operators` that enables users to create a new Observable with
operators added to it as methods based on a passed hash of operator methods.

``` ts
source.operators({ filter, map })
  .filter(x => x > 1)
  .map(x => x * 10)
```

related #2034


blesh:
I'm proposing this because:

1. I think it solves some of the type inference concerns.
2. It reads a little cleaner and is more ergonomic.
3. It's an alternative

blesh:
NOTE: We can move the method to `Observable` proper like `operate` from #2034 as well.

blesh:
... also the name can be bike-shedded, this is more of a proposal.

staltz:
Name suggestion: `include` or `including`. But still not super happy with the direction this is going.

PR #2038, opened 2016-10-14
fix(IteratorObservable): Observables `from` generators will now final…

### LINKED ISSUE #1938: Rx.Observable.from..take does not close iterator
**RxJS version:** 5.0.0-beta.12

**NodeJS version:** 6.5.0

**Code to reproduce:**
```
function *G() {
	try {
		while ( true )
			yield 42
	} finally {
		console.log( `ok` )
	}
}
Rx.Observable.from( G() ).take( 1 ).subscribe( x => console.log( x ) )
```

**Expected behavior:**
```
42
ok
```

**Actual behavior:**
```
42
```

**Additional information:**
Following native code closes iterator properly:
```
function *G() {
	try {
		while ( true )
			yield 42
	} finally {
		console.log( `ok` )
	}
}
for ( let x of G() ) {
	console.log( x )
	break
}
```
Output:
```
42
ok
```

blesh:
...ize when subscription ends

In order to model the behavior of `for..of` when consuming a Generator, if a `break` is hit in the `for..of`, `return()` is called on the generator and the generator will jump to a `finally` block if it has one. Observables created from generators will now have this same behavior.

```js
    Observable.from((function* () {
      try {
        yield 1;
        yield 2;
        yield 3;
      } finally {
        console.log('finalized');
      }
    })())
    .take(2)
    
    // should log
    // 1
    // 2
    // finalized
```
    
fixes #1938

### DIFF
--- a/spec/observables/IteratorObservable-spec.ts
+++ b/spec/observables/IteratorObservable-spec.ts
@@ -1,5 +1,6 @@
 import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
+import {queue} from '../../dist/cjs/scheduler/queue';
 import {IteratorObservable} from '../../dist/cjs/observable/IteratorObservable';
 
 declare const expectObservable;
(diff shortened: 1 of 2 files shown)


jayphelps:
Anyone else reading this later, keep in mind that generators like this are basically singletons, so if one subscribers forces the generator to `.return()`, then **every** subscriber is forced into that.

```js
const stream$ = Observable.from((function* () {
  try {
    yield 1;
    yield 2;
    yield 3;
  } finally {
    console.log('finalized');
  }
})());

stream$.take(1).subscribe();
stream$.take(3).subscribe();
// Second subscriber will only get one, not three because
// first subscriber is an scumbag
```

jayphelps:
looks good, will wait a few for others to chime in.

blesh:
Related is the issue I've created in the TC39 spec. We might even want to wait for a ruling there... https://github.com/tc39/proposal-observable/issues/115

blesh:
Given that this updates RxJS 5 to match the direction of the observable spec, I think it's the proper path. https://github.com/tc39/proposal-observable/issues/115

jayphelps:
LGTM

PR #2039, opened 2016-10-15
fix(WebSocketSubject.prototype.multiplex): no longer nulls out socket…

### LINKED ISSUE #2037: Unsubscribing from WebSocketSubject reference to Websocket
rxjs@5.0.0-rc.1

When unsubscribing from a multiplex WebSocketSubject I am seeing the socket getting set to null. There are still other subscribers and the socket still exists, but we are loosing the reference to the socket

I believe this was introduced in [62d242e](https://github.com/ReactiveX/rxjs/commit/62d242ecf5c7b513c86aceb6e6836efe6ddc909f).

Relevant Code
```ts
protected _subscribe(subscriber: Subscriber<T>): Subscription {
    const { source } = this;
    if (source) {
      return source.subscribe(subscriber);
    }
    if (!this.socket) {
      this._connectSocket();
    }
    let subscription = new Subscription();
    subscription.add(this._output.subscribe(subscriber));
    subscription.add(() => {
      const { socket } = this;
      if (this._output.observers.length === 0 && socket && socket.readyState === 1) {
        socket.close();
      }
      this._resetState(); //This is where to socket reference is getting lost
    });
    return subscription;
  }
```

blesh:
This fix ensure the observers count goes to zero before the state is reset on the `WebSocketSubject` instance

fixes #2037


kwonoj:
Change looks fine (and intentionally not enable l_gtm checker for additional reviewer)

jayphelps:
LGTM

PR #2042, opened 2016-10-16
chore(package.json): bump up package dependencies

kwonoj:
**Description:**

This PR cleans up `package.json` include below changes.

- removes `es6` specific build script (`es6_build_docs` remains) to avoid possible confusions
- bump up package dependencies
- removes `typings` definition, use `@types` for devdependency type definitions
- introduce `yarn` for travis build


jayphelps:
This is now failing because we merged the PR to [use rollup](https://github.com/ReactiveX/rxjs/pull/2064/files#diff-6d2911fe563068b8126098588db98a84R7) which depends on the ES6 build output and this PR removes the scripts to build it.

I've added those scripts back, though fairly soon we probably want to think that whole thing through and make sure we're not accidentally leaking es6 into any of our npm packages.

@kwonoj can you confirm that's OK for now? Wanna get this merged since it's primary around using yarn and bumping deps.

kwonoj:
@jayphelps , yes, updated looks 👍

jayphelps:
LGTM

Issue #2013, opened 2016-10-06
Operator to support AJAX persistence and multicasting


benlesh:
Currently, there's a PR to remove `cache` #2012. The original intention of `cache` was to support use-cases, mostly coming from the Angular 2 community, revolving around wanting to persist successful AJAX requests, but still enable `retry` and `retryWhen` on failed AJAX requests.
### This operator should
1. Multicast the source observable
2. Keep all values and "replay" them after a successful completion
3. Enable "retrying" on error
### Stuff to consider, and unanswered questions:
1. What are the behaviors when applied to a never-ending observable (`Observable.interval(100)`, for example)
2. How will this operator behave when applied to another `hot` observable?
### Possible Solutions
- Re-add the `cache` operator but rename it to something more descriptive of it's purpose.
- Add a simplified operator that takes no arguments called `shareResults` [proposed here](https://github.com/ReactiveX/rxjs/issues/2013#issuecomment-253053953)

For sake of discussion, we'll call this operator `persist`.

StephenFluin:
I love the `cache` operator. The name doesn't matter to me, but I hope it exists somewhere. I'm aware that I can manually do this with `publishReplay` and `connect`, but `cache` was simpler and cleaner.

I often use it to share data across components/routes when making HTTP requests that don't need to be refreshed often:

https://github.com/sjtrimble/fluin.io/blob/master/src/app/shared/post.service.ts

Or I used it to eliminate re-processing of incoming data from data streams.

https://github.com/StephenFluin/ames/blob/e5502d1d0aadb44d44ec1fde659ed94ebf74057f/src/app/shared/auth.service.ts#L33

blesh:
@StephenFluin it should be noted that `publishReplay()`/`connect()` can't be "retried", really. On failure, it's just done.

blesh:
What if we simplified it to always persist all values and multicast and we called it `shareResults`? I think that implies that even after all value have arrived, all of them are kept as "results" and shared amongst all subscribers. There would be no arguments.

```js
Observable.ajax.getJSON('http://something.com/whatever').shareResults();
```

Thoughts?

blesh:
I also think `shareResults` keeps some similarity to the `share` operator, which has a similar behavior, retry-wise, but adds that little extra to imply result persistence.

newsash:
Use-case to consider (shared auto-updating value, which should not be updated in periods without subscribers):
``` typescript
if (cache) { return cache; }

[...]

PR #2043, opened 2016-10-17
feat(shareResults): add shareResults operator

benlesh:
Adds a very basic implementation of the `shareResults` operator proposed in #2013 

TODO:
- [x] basic impl
- [x] documentation with examples
- [x] flush out test suite
- [ ] add perf tests (against nothing in Rx4, because it does not exist there)
- [ ] optimize closures


benlesh:
I think this PR should probably be closed in favor of #2443

PR #2046, opened 2016-10-17
fix(ReplaySubject): observer now subscribed prior to running subscrip…

### LINKED ISSUE #2044: ReplaySubject subscription handler is called before it is subscribed
**RxJS version:**
rxjs@5.0.0-rc.1

**Code to reproduce:**
```js
let subject = new Rx.ReplaySubject();
subject.next(1);
subject.subscribe(function (data) {
  console.log(data);
  if (data < 3) { data++; subject.next(data); }
});
```

**Expected behavior:**
Output: 1, 2, 3

**Actual behavior:**
Output: 1

**Additional information:**
This works properly in RxJS v4

It seems that the issue comes from this code: https://github.com/ReactiveX/rxjs/blob/master/src/ReplaySubject.ts#L40
Where we first call the subsrcibe handler with the previous values, before we actually subscribe him for `next` values.

blesh:
…tion function

Removed reliance on inherited `_subscribe` method from `Subject`. ReplaySubject has different behaviors during subscription for notifying new subscribers after a completion or error. After moving the `super._subscribe` call before the notification from saved `_events`, it was apparent that the inherited behavior from `Subject` was stopping the subscriber before it could be notified properly.

fixes #2044


jayphelps:
LGTM

PR #2049, opened 2016-10-18
`distinct` perf updates

blesh:
### First commit

- BREAKING CHANGE: changes `distinct` to use a keySelector and NOT a comparer. Comparer check was to mimic RxJS 2 or 3, which was the current version when this operator was originally created.
- BREAKING CHANGE: removes `distinctKey` as it's redundant.
- Adds the minimum necessary `Set` implementation if `Set` does not exist globally.
- Improves overall perf of `distinct()` by 5x in first commit.
- Improves overall perf of `distinct(keySelector)` (versus older `distinctKey`) by 8x in first commit

### Additional commit (do not flatten)

- Refactors error handling for V8 optimization. Moves deopt only to a path that is hit if a keySelector is actually provided. This further improves perf of `distinct()` by another 2x (total of 9-10x), and `distinct(keySelector)` slightly (total of 9x)

### Note

`comparer` function argument exists in Rx4, yes, but after discussion with @mattpodwysocki, we are removing it because it is of little use, and most use cases could be solved by the keySelector. We can add it at a later time if we choose with minimal breakage.


jayphelps:
commit message:

> BREAKING CHANGE: `distinct` no signature changed, first argument is
`keySelector`, not `comparer` function.

I think the "no" was accidental. I think something like this is more clear:

> BREAKING CHANGE: `distinct` operator has changed, first argument is
an optional `keySelector` and a `compare` function is no longer supported

kwonoj:
Just curiosity - does micro perf test number shows expected gain? (I know micro perf hasn't run quite long time though..)

blesh:
> Just curiosity - does micro perf test number shows expected gain? (I know micro perf hasn't run quite long time though..)

Yes... ~9-10x faster in micro perf test that previous.

blesh:
> I think the "no" was accidental. I think something like this is more clear

 yup, good catch

src/operator/distinct.ts
```diff
@@ -5,29 +5,33 @@ import { TeardownLogic } from '../Subscription';
...
+import { ISet, Set } from '../util/Set';
 
 /**
  * Returns an Observable that emits all items emitted by the source Observable that are distinct by comparison from previous items.
  * If a comparator function is provided, then it will be called for each item to test for whether or not that value should be emitted.
  * If a comparator function is not provided, an equality check is used by default.
  * As the internal HashSet of this operator grows larger and larger, care should be taken in the domain of inputs this operator may see.
  * An optional parameter is also provided such that an Observable can be provided to queue the internal HashSet to flush the values it holds.
+ * @param {function} [keySelector] optional function to select which value you want to check as distinct
  * @param {function} [compare] optional comparison function called to test if an item is distinct from previous items in the source.
```
jayphelps: Need to remove this I believe?

src/operator/distinct.ts
```diff
@@ -5,29 +5,33 @@ import { TeardownLogic } from '../Subscription';
 import { OuterSubscriber } from '../OuterSubscriber';
 import { InnerSubscriber } from '../InnerSubscriber';
 import { subscribeToResult } from '../util/subscribeToResult';
+import { ISet, Set } from '../util/Set';
 
 /**
  * Returns an Observable that emits all items emitted by the source Observable that are distinct by comparison from previous items.
  * If a comparator function is provided, then it will be called for each item to test for whether or not that value should be emitted.
```
jayphelps: docs need to be updated to not describe the compare function

jayphelps:
@blesh is going to update the jsdocs, then we'll merge.

jayphelps:
LGTM

PR #2050, opened 2016-10-18
fix(reduce/scan): both scan/reduce operators now accepts `undefined` itself as a valid seed

### LINKED ISSUE #2047: scan doesn't accept undefined as seed value
problem: `scan(reduceFn, undefined)` behaves identically to `scan(reduceFn)`
proposal: `undefined` is a valid seed value

(1) it is not consistent with `Array.prototype.reduce`
```
['a', 'b', 'c'].reduce(function(acc, s){ return acc + ' ' + s; })
"a b c"
['a', 'b', 'c'].reduce(function(acc, s){ return acc + ' ' + s; }, undefined)
"undefined a b c"
```

(2) with certain kinds of reducing functions, this behavior can lead to unexpected behavior
```javascript
var reducer = new Rx.Subject();
reducer.scan(function(acc, fn) { return fn(acc); }, initialState);
reducer.next(function nextState(state){ /* return new state */ })
```
if `initialState` is `undefined`, then subject will emit `nextState` function instead of `nextState(undefined)` leading to undesired behavior. Using a different value for `initialState` could also result in undesired behavior for example if combined with `distinctUntilChanged`

the [scan api doc](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-scan) has no mention that `undefined` is not a valid seed value
> Returns an Observable that applies a specified accumulator function to each item emitted by the source Observable. If a seed value is specified, then that value will be used as the initial value for the accumulator. If no seed value is specified, the first item of the source is used as the seed.

jayphelps:
Array#reduce supports `undefined` as a valid seed value, so we should too. This was mostly an oversight.

```js
of(1, 2, 3).reduce((acc, x) => acc + ' ' + x, undefined);
// "undefined 1 2 3"
```

```js
of(1, 2, 3).scan((acc, x) => acc + ' ' + x, undefined);
// "undefined 1"
// "undefined 1 2"
// "undefined 1 2 3"
```

fixes #2047

Cc/ @blesh


kwonoj:
Implementation side it looks fine, just need to clarify if this is expected behavior. (for me it seems legit to align with native)

trxcllnt:
@jayphelps @kwonoj I've always wondered, why do we use a getter/setter for the seed value? Is it faster?

jayphelps:
I'm betting there is no reason other than that's what the person who wrote it used.

kwonoj:
> I'm betting there is no reason other than that's what the person who wrote it used.

: Pretty much this case. Using getter / setter in this codebase is mostly depends on preference / chocies of writer. At least it does not gives any perf or so.

mattpodwysocki:
LGTM, we need to be careful as `undefined` is as valid a value as `null` in terms of initial values, such as `startWith` etc.  And `Function.prototype.bind` doesn't change the number of parameters in the end going to the function, so we're fine with checking the param count.

src/operator/reduce.ts
```diff
@@ -53,16 +53,24 @@ export function reduce<T>(this: Observable<T>, accumulator: (acc: T[], value: T,
...
 
 export class ReduceOperator<T, R> implements Operator<T, R> {
-
-  constructor(private accumulator: (acc: R, value: T) => R, private seed?: R) {
-  }
+  constructor(private accumulator: (acc: R, value: T) => R, private seed?: R, private hasSeed: boolean = false) {}
 
   call(subscriber: Subscriber<R>, source: any): any {
-    return source._subscribe(new ReduceSubscriber(subscriber, this.accumulator, this.seed));
+    return source._subscribe(new ReduceSubscriber(subscriber, this.accumulator, this.seed, this.hasSeed));
```
blesh: I can totally see *why* you're passing `hasSeed` to the Subscriber via the constructor. It makes it less polymorphic. However, I do feel like the subscriber itself should be able to determine if it has a seed. How does that effect performance?

Do you even think it matters? I'm on the fence. It probably doesn't. It just feels weird as an API, albeit an internal one to be like `seed`, `hasSeed` on the tail end of a call.

Thoughts?

jayphelps: we would have to have additional branching, would we not?

```js
if (this.hasSeed) {
  return source._subscribe(new ReduceSubscriber(subscriber, this.accumulator, this.seed));
} else {
  return source._subscribe(new ReduceSubscriber(subscriber, this.accumulator));
}
```

And then each class downstream would have to duplicate the `arguments.length >= 2` check wouldn't they?

I did it this way because it seemed to me the best performant and the other solutions had a lot of duplication. It's very non-obvious at first, but best I can tell you _have_ to check at the original call site otherwise you'll lose that info because of the `Operator#call` architecture.

mattpodwysocki: 👍

jayphelps: @trxcllnt sorry, I'm not following what your suggestion is? Is it to move the `arguments.length >= 2` check into the `ScanOperator` constructor and use `new ScanOperator(...arguments)`?

A little pedantic of me, but technically that will perform not as well because `new ScanOperator(...arguments)` transpiles to something like `new (Function.prototype.bind.apply(Wow, [null].concat(args)))();`. But this really isn't the typical hot code path, so prolly doesn't matter.

trxcllnt: @jayphelps no you're totally right and I should have clarified I was playing devil's advocate: I'm 100% against doing that.

jayphelps:
Approved during our core meeting. I don't like to self-merge, but I was tasked to merge all the approved PRs so #YOLO

PR #2060, opened 2016-10-23
docs(readme): add further instructions for building docs on Mac

boneskull:
**Description:**

Add instructions to install missing Ghostscript fonts on Mac.

Homebrew's `ghostscript` package does not install any fonts;
`gm convert` will fail without them.  Add instructions to install these
fonts manually.


staltz:
LGTM

PR #2068, opened 2016-10-25
refactor(util): remove unused variables in src/util/

saneyuki:
**Description:**
- This removes needless type parameters & declarations.
- Just a minor refactoring.

**Related issue (if exists):**
- I found this refactoring chance in playing with codes for #1969

PR #2069, opened 2016-10-25
refactor(AjaxObservable): remove needles type param T from AjaxObservable.getJSON()

saneyuki:
**Description:**

BREAKING CHANGE: 
- Remove the unused type parameter `T`
  - By this change, the user does not have to specify needless type parameter.

### DIFF
--- a/src/observable/dom/AjaxObservable.ts
+++ b/src/observable/dom/AjaxObservable.ts
@@ -67,7 +67,7 @@ export interface AjaxCreationMethod {
   post(url: string, body?: any, headers?: Object): Observable<AjaxResponse>;
   put(url: string, body?: any, headers?: Object): Observable<AjaxResponse>;
   delete(url: string, headers?: Object): Observable<AjaxResponse>;
-  getJSON<T, R>(url: string, headers?: Object): Observable<R>;
+  getJSON<T>(url: string, headers?: Object): Observable<T>;
 }
 
 export function ajaxGet(url: string, headers: Object = null) {


jayphelps:
Minor change: "BREAKING CHANGE" messages in commits need to be fully descriptive on their own, independent of the first part of the commit message because when the CHANGELOG is generated, they live in their own section.

e.g. https://github.com/ReactiveX/RxJS/commit/aa30af2 generates to https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md#breaking-changes-1

So in this case, something _like_

> BREAKING CHANGE: `Observable.ajax.getJSON()` now only supports a single type parameter, `getJSON<R>(url: string, headers?: Object): Observable<R>`. The extra type parameter it accepted previously was superfluous.

Double check all your other PRs have good BREAKING CHANGE messages too please. Also--related refactors like these should prolly have been in a single PR to ease in the discussion and tracking. #2072 #2071 #2070 #2068

saneyuki:
> Minor change: "BREAKING CHANGE" messages in commits need to be fully descriptive on their own, independent of the first part of the commit message because when the CHANGELOG is generated, they live in their own section.

Okay, I'll update.

> related refactors like these should prolly have been in a single PR to ease in the discussion and tracking. #2072 #2071 #2070 #2068

They are originated from the same issues, but I thought they still are different issue. cleaning up codes or changing signature which we may need some decisions.

saneyuki:
updated the commit message.

src/observable/dom/AjaxObservable.ts
```diff
@@ -67,7 +67,7 @@ export interface AjaxCreationMethod {
   post(url: string, body?: any, headers?: Object): Observable<AjaxResponse>;
   put(url: string, body?: any, headers?: Object): Observable<AjaxResponse>;
   delete(url: string, headers?: Object): Observable<AjaxResponse>;
-  getJSON<T, R>(url: string, headers?: Object): Observable<R>;
+  getJSON<R>(url: string, headers?: Object): Observable<R>;
```
blesh: should probably just be `T`... that's what we do everywhere else.

saneyuki: @blesh okay. I fixed in https://github.com/ReactiveX/rxjs/pull/2069/commits/501bac49bab8949a64f6b76a47ac47ec6b1bfec1

PR #2070, opened 2016-10-25
fix(IteratorObservable): clarify the return type of IteratorObservable.create()

saneyuki:
**Description:**
- Before this change, if call `IteratorObservable.create<number>()` with specifying type parameter, the return type of it was `IteratorObservable<{}>`. This pull request fix this.
  - This might be potential breaking change.


saneyuki:
updated the commit message

PR #2071, opened 2016-10-25
refactor(ErrorObservable): clarify an error type `E` for ErrorObservable.

saneyuki:
**Description:**
- BREAKING CHANGE - `ErrorObservable.create()` (`Obserbable.throw()`)
takes a new type parameter `E` that express a error type instead on old
one's `T`.
- Their functions return `ErrorObservable<E>`.
  - It is still the drived type of `Observable<any>`, it's not changed from
old one.
  - But it contains the error typed as `E` now.

### DIFF
--- a/src/observable/ErrorObservable.ts
+++ b/src/observable/ErrorObservable.ts
@@ -12,7 +12,7 @@ export interface DispatchArg {
  * @extends {Ignored}
  * @hide true
  */
-export class ErrorObservable extends Observable<any> {
+export class ErrorObservable<T> extends Observable<any> {
 
   /**
    * Creates an Observable that emits no items to the Observer and immediately
@@ -54,7 +54,7 @@ export class ErrorObservable extends Observable<any> {
    * @name throw
    * @owner Observable
    */
-  static create<T>(error: any, scheduler?: Scheduler) {
+  static create<T>(error: T, scheduler?: Scheduler): ErrorObservable<T> {
     return new ErrorObservable(error, scheduler);
   }
 
@@ -63,7 +63,7 @@ export class ErrorObservable extends Observable<any> {
     subscriber.error(error);
   }
 
-  constructor(public error: any, private scheduler?: Scheduler) {
+  constructor(public error: T, private scheduler?: Scheduler) {
     super();
   }
 


saneyuki:
updated the commit message.

mattpodwysocki:
@saneyuki wouldn't this break type inference such as if in our selector, we're depending on `Observable<T>` and we had conditional logic in our selector?

```js
function selector<Observable<number>>(x: number) {
  if (x % 2 === 0) {
    return Observable.throw<number>('nope nope nope');
  }

  return Observable.of<number>(1);
}
```

blesh:
I think @mattpodwysocki is right here. @david-driscoll or @kwonoj can you confirm?

saneyuki:
@mattpodwysocki @blesh 

- `ErrorObservable` is defined as `class ErrorObservable extends Observable<any> {}` , so [by your example](https://github.com/ReactiveX/rxjs/pull/2071#issuecomment-256489777), the return type of `selector()` is `Observable<number>` (and we need not specify type param `T`  to `Observable.throw()`).
- At all, is it a right way that `ErrorObservable` is derived from `Observable<any>` so that make it more generic? Should we change from `ErrorObservable` to `ErrorObservable<T>`>

kwonoj:
So this seems bit tricky.

First, I think generic of error observable itself is redundant, cause it does not being used to infer type of error. For example, in current interface `Observable.throw(1)` is inferred to 
```Observable<T>.throw: <{}>(error: any, scheduler?: Scheduler) => ErrorObservable```

Since ErrorObservable itself does not have generic type and force set base observable type to any.

But also it is questionable if it's right to give type to ErrorObservable itself, since all of error being emit in Observable is currently `any`, do not have specific types so allowed to error any value as needed. It may considerable option to adapt generic type to ErrorObservable to allow type inferred (cases like selector need to return single type T). 

Waiting @david-driscoll 's input as well.

blesh:
I think this should be:  `Observable.throw<T>(error: T)` returning `ErrorObservable<T>`. Since we know the type that we're passing to it, it can be inferred.

@saneyuki, if you want to make that change in this PR that would be cool. Otherwise, I don't think we'll merge this PR, but rather make that change in another PR.  :PR:

saneyuki:
@blesh 

> I think this should be: Observable.throw<T>(error: T) returning ErrorObservable<T>. Since we know the type that we're passing to it, it can be inferred.

Okay. I'll change to it.

BTW, I think it would be better to use type parameter `E` (`Observable.throw<E>`, `ErrorObservable<E>`) instead of `T`, to express the generics type which we intend to accept an error type to. How do you think about it?

saneyuki:
@blesh 

I have updated this pull request. [Your comment](https://github.com/ReactiveX/rxjs/pull/2071#issuecomment-258496627) is like this, isn't it?

blesh:
:shipit:

PR #2072, opened 2016-10-25
refactor(FromEventObservable): remove needless type parameter from FromEventObservable & FromEventPatternObservable

saneyuki:
**Description:**
- By this change, the consumer does not simplify the type to specify to use them
- BREAKING CHANGE: `FromEventObservable<T>` & `FromEventPatternObservable<T>` are no longer take type parameter `R`.
  You don't have to pass 2nd type parameter.


saneyuki:
update commit messages.

blesh:
:shipit:

PR #2077, opened 2016-10-25
fix(sample): source is now subscribed to before the notifier

### LINKED ISSUE #2075: providing .sample() the same observable it samples produces stale output
Thanks for the skeleton @jayphelps 

**RxJS version:**
5.0.0.beta-12

**Code to reproduce:**
http://jsbin.com/sapuju/4/edit?js,console,output

**Expected behavior:**
Open the console. 3

**Actual behavior:**
2

**Additional information:**

blesh:
- made sure that notification observable is subscribed to *after* the source observable.

fixes #2075

### Up For Debate

I'm not sure this fix is necessary.

- Almost all operators that accept an observable as an argument (usually as a notifier) subscribe to that observable *before* they subscribe to the source. 
- This change might misalign this operator with others. (`buffer` or `bufferWhen` come to mind)
- #2075 represents a edge-case or possibly a corner-case for `sample`. Why would someone `sample` an observable with itself?


jayphelps:
@mattpodwysocki can you provide any insight into the differing behavior between v4 and v5 in #2075?

mattpodwysocki:
@jayphelps this was intentional design.

trxcllnt:
@jayphelps @blesh also both `takeUntil` and `skipUntil` should subscribe to their notifiers before subscribing to the source Observable. This is to ensure that if the notifier synchronously `next` or `complete`, the source Observable isn't subscribed to.

jayphelps:
@trxcllnt O snap that's a really good point.

jayphelps:
@blesh I think we have pretty solid justification to proceed. Can I merge?

blesh:
:shipit:

PR #2078, opened 2016-10-25
refactor(Scheduler): clearify the signature which `Scheduler.schedule()` takes as `work`.

saneyuki:
**Description:**
- For more readability, more consistently.
- I think this would not be a breaking change by these reasons:
  - There are no problem if we pass the function which does not has `this` specifying.
  - There are no problem if we pass the function which has `this: any` or `this: Action<T>`.
    - This case is valid with the actual behavior.
  - If we pass the function which has `this: SomeNonValidTypeWithAction<T>`,
    then, basically, its code would be wrong.


blesh:
@saneyuki the build is failing after this change.

saneyuki:
@blesh

> the build is failing after this change.

 fixed.

blesh:
:shipit:

PR #2080, opened 2016-10-25
fix(Object.assign): stop polyfilling Object assign

blesh:
**Description:**

RxJS should not be responsible for polyfilling `Object.assign`.
This should also increase code coverage around the alternate method used for assign

BREAKING CHANGE: RxJS will no longer polyfill `Object.assign`. It does
not require `Object.assign` to function, however, your code may be
inadvertently relying on this polyfill.


jayphelps:
LGTM

PR #2082, opened 2016-10-25
fix(Symbol.iterator): will not polyfill Symbol iterator unless Symbol…

blesh:
… exists
- adds more robust tests
- removes strange logic added to support es6-shim, as no documentation or reasoning could be found to support its existence
- ponyfills Symbol rather than polyfills... meaning the module just returns something that can be used as the symbol, but
  if `Symbol` exists as a function, like it should, and it does not have `Symbol.iterator`, only then will it polyfill it

BREAKING CHANGE: RxJS will no longer polyfill `Symbol.iterator` if `Symbol` does not exist. This may break code that inadvertently relies on this behavior

### DIFF
--- a/spec/symbol/iterator-spec.ts
+++ b/spec/symbol/iterator-spec.ts
@@ -1,32 +1,100 @@
-import {expect} from 'chai';
-
-import {root} from '../../dist/cjs/util/root';
-import {$$iterator} from '../../dist/cjs/symbol/iterator';
+import { expect } from 'chai';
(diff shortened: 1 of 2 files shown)


blesh:
It looks like this may still be [an issue with es6-shim](https://github.com/paulmillr/es6-shim/blob/3e46993a95b1867665166e0290e5cbaaf2ccb2f6/es6-shim.js#L266), because they're using a substitute string that is unique to es6-shim, rather than `@@iterator` like other libraries.

jayphelps:
LGTM

PR #2084, opened 2016-10-26
chore(package.json): bump up dependencies to TS@2.0.6

kwonoj:
**Description:**

This PR updates TS to latest, which includes several fixes also improves build speed.

image

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -97,12 +97,6 @@ amdefine@>=0.0.4:
   version "1.0.0"
   resolved "https://registry.yarnpkg.com/amdefine/-/amdefine-1.0.0.tgz#fd17474700cb5cc9c2b709f0be9d23ce3c198c33"
 
-amdetective@~0.2.1:
-  version "0.2.1"
-  resolved "https://registry.yarnpkg.com/amdetective/-/amdetective-0.2.1.tgz#bd96fcb67e5dfd51dafd4afecff46269efa7dca9"
-  dependencies:
-    esprima "~2.7.0"
-
 anchor-markdown-header@^0.5.5:
   version "0.5.5"
   resolved "https://registry.yarnpkg.com/anchor-markdown-header/-/anchor-markdown-header-0.5.5.tgz#c8c7d23695689e69fdac2b4ce81680b4f38e0d6e"
@@ -119,6 +113,10 @@ ansi-styles@^2.2.1:
   version "2.2.1"
   resolved "https://registry.yarnpkg.com/ansi-styles/-/ansi-styles-2.2.1.tgz#b432dd3358b634cf75e1e4664368240533c1ddbe"
 
+any-promise@^1.0.0:
+  version "1.3.0"
+  resolved "https://registry.yarnpkg.com/any-promise/-/any-promise-1.3.0.tgz#abc6afeedcea52e809cdc0376aed3ce39635d17f"
+
 any-promise@~0.1.0:
   version "0.1.0"
   resolved "https://registry.yarnpkg.com/any-promise/-/any-promise-0.1.0.tgz#830b680aa7e56f33451d4b049f3bd8044498ee27"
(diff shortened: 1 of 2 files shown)


jayphelps:
What's the lowdown on how yarn works with these? I'm still not entirely clear how its lockfile handles using semver ranges in our `package.json` like we have `^2.0.3`...which means any `2.x.x` release...but if a lock file is used..why does it not negate using ranges at all--except for dependencies of dependencies, which we don't control their range, which I get that use case.

i.e. why do we need this PR to bump us and are ranges still even meaningful? A possibly related followup is whether or not bumping versions should also update the lockfile?

Sorry--I thought I understood yarn, but reading through their docs...the answers to these aren't clear for me.

kwonoj:
@jayphelps , oh my... good point. blame me that I used `npm install` instead of `yarn`. 

So intention of this PR was to pin down dependency version to @2.0.6 as minimum to clarify package installation, so it should include lock file for yarn and I've missed it. :/ As said, in case of npm install it should resolve to latest since it's within minor semver range. I assyne yarn works similar but wanted to update lockfile to ensure it.

jayphelps:
Future searching through twitter, hackernews (yuck), etc seems to suggest indeed semver ranges in yarn are now only used when using the `upgrade` command. If not provided with a higher version than the existing range, it find the highest matching in the existing range and update the lockfile.

https://yarnpkg.com/en/docs/cli/upgrade

Because we won't get upgrades within range for "free" anymore, we should prolly discuss at the next meeting how/when we'll manually PR upgrades. Without precedence, it's hard to know how often our dependencies bump minor/patch..but I can easily imagine someone PR'ing a bump every week and it might get pretty old dealing with them.

kwonoj:
makes sense. wish there's some automated way to manage this.

jayphelps:
There's https://greenkeeper.io which has been awesome for me in other OSS, but unclear if can be made to solve this particular thing with yarn and even if it did, it would be basically just repeated PRs to keep up updated.

As it's probably clear, I'm not entirely convinced yet that lockfiles are the amazing solution to the problem of "works on my machine". BUT I'm certainly cognizant that I'm not the authority on that discussion and lack full context to the choice by yarn. There may also be reasons other than that, that are attractive.

kwonoj:
> There's https://greenkeeper.io which has been awesome for me in other OSS, but unclear if can be made to solve this particular thing with yarn and even if it did, it would be basically just repeated PRs to keep up updated.

: Yes, I'm aware of those but since greenkeeper does not support yarn yet (creating PR including lockfile updates), still requires manual responsibilities of managing lockfile once notified about package version bumps up. Just roughly hoped if there's way to chain `npm install` automatically hooks lockfile update, and force include into commit.

> I'm not entirely convinced yet that lockfiles are the amazing solution to the problem of "works on my machine". BUT I'm certainly cognizant that I'm not the authority on that discussion and lack full context to the choice by yarn. There may also be reasons other than that, that are attractive.

: +1 here. that's reason currently this codebase only uses yarn and lockfile for travis CI, instead of every local contributor to utilize it. Even travis failed by yarn, it can be recovered (or disabled) quickly enough.

jayphelps:
> that's reason currently this codebase only uses yarn and lockfile for travis CI, instead of every local contributor to utilize it. Even travis failed by yarn, it can be recovered (or disabled) quickly enough.

I'm not following. If contributors don't use yarn, isn't yarn not solving anything for us really? My local node_modules can and will often be different then what runs in travis w/ yarn if I use npm?

kwonoj:
> If contributors don't use yarn, isn't yarn not solving anything for us really?

: Yes, correct. For now, single and only reason to introduce yarn (for CI only) is expecting boost build speed by reduce package installation time (what yarn claim, though I haven't done 1:1 comparison on this repo yet). I am expecting to expand it's usage over local devs as well once it's verified sufficiently, as well as resolve some of possible questions while using yarn on CI only while, such as this kind of bump-up cases.

> My local node_modules can and will often be different then what runs in travis w/ yarn if I use npm?

: While it is also very true, this repo itself's dependency is not complex and in most cases it doesn't causes problems between per-machine installations. I do not think this is immediate issue to be resolved by introducing yarn in this package.

yarn.lock
```diff
@@ -1229,57 +1235,6 @@ browserify@^13.0.0:
     vm-browserify "~0.0.1"
     xtend "^4.0.0"
 
-browserify@13.0.1:
```
kwonoj: Yes, totally agreed.

jayphelps:
> For now, single and only reason to introduce yarn (for CI only) is expecting boost build speed by reduce package installation time

FWIW build times haven't changed in any notable measure https://travis-ci.org/ReactiveX/rxjs/builds. Perhaps we're not using it in the ideal way, or it's because travis doesn't support it by default, or because maybe there is no caching going on and maybe that's where the real wins are. I dunno 🙁

kwonoj:
> FWIW build times haven't changed

:  😢 

I think at least this PR provides good chance to discuss further. Maybe we could simply start with `just drop it` to somewhere else.. Suggest to consider yarn and its lockfile as experimental stage. For now it won't block anyone at this moment. (I hope)

PR #2091, opened 2016-10-28
fix(build): UMD Wrap global bundle

zertosh:
In #2064, I bundled the build as a CommonJS package, completely neglecting web use. This PR UMD wraps the bundle so it's AMD/CommonJS/Web friendly.

Ideally, I would've liked to use `babel-plugin-transform-es2015-modules-umd` instead of using my own wrapper, but the perf cost of not including the parens is measurable:

**Without parens:**

```
$ for i in {1..5}; do node -p "s=Date.now();require('./dist/global/Rx.js');Date.now()-s"; done
49
48
48
43
40
```

**With parens:**

```
$ for i in {1..5}; do node -p "s=Date.now();require('./dist/global/Rx.js');Date.now()-s"; done
36
28
27
27
32
```

The wrapper I'm using came from `babel-plugin-transform-es2015-modules-umd`. I'll add the option to the plugin so it adds parens. Once that's in, I'll come back here and remove the custom wrapper in favor of the plugin.

## Test plan:

* In Node, made sure that `require('./dist/global/Rx.js').Observable` existed.
* Created an `index.html` with `<script src="dist/global/Rx.js"></script>` and made sure that `Rx.Observable` existed from the console.

### DIFF
--- a/tools/make-umd-bundle.js
+++ b/tools/make-umd-bundle.js
@@ -3,6 +3,26 @@ var fs = require('fs');
 
 var babel = require('babel-core');
 
+var umdPrelude = [
+  ';(function (global, factory) {',
+  '  if (typeof define === "function" && define.amd) {',
+  '      define(["exports"], factory);',
+  '  } else if (typeof exports !== "undefined") {',
+  '      factory(exports);',
+  '  } else {',
+  '      var _exports = {};',
+  '      factory(_exports);',
+  '      global.Rx = _exports;',
+  '  }',
+  '})(this, (function (exports) {', // The extra wrapping parens are important.
+  ''                                // https://github.com/nolanlawson/optimize-js
+].join('\n');
+
+var umdPostlude = [
+  '',
+  '}));'
+].join('\n');
+
 rollup.rollup({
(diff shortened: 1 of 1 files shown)


kwonoj:
Referring https://github.com/ReactiveX/rxjs/pull/2064#issuecomment-256869506. Seems it's possible to use TSC only with rollup, maybe it's worth to explore that before check in this.

blesh:
@kwonoj It sounds like this PR will at least solve problems people might have with using what's currently in master from the web. I suggest we get that problem fixed first, then explore options that are more efficient and build time.

Brooooooklyn:
@blesh @kwonoj https://github.com/ReactiveX/rxjs/pull/2093 this pr could also fix issue here

kwonoj:
For matters of make build usable, I'd like to suggest simply revert latest changes. Problem of this PR is we don't know what possible other failures could have and does not gaurantee 100% work. Next PR addresses this chane simply can include previous changeset as well.

zertosh:
Closing since #2093 is better

blesh:
❤️ @zertosh thank you

PR #2093, opened 2016-10-28
chore: Use TypeScript to bundle UMD directly

Brooooooklyn:
|  | Babel | TypeScript |
| --- | :-: | --: |
| Rx.js | 514K | 483k |
| Rx.min.js | 168k | 139k |

I've tested the Rx.js and Rx.min.js in my browser, and its work fine 😬


zertosh:
What's the load perf like?

Brooooooklyn:
TypeScript:
image

Babel:
image

zertosh:
Got the same results. Looks like `tslib` is leaking globals `__extends`, `__assign`, etc. when the build is run without a CJS wrapper (i.e. browser).

Brooooooklyn:
The leaking global problem would resolve when https://github.com/Microsoft/TypeScript/pull/9097 landed. And I think that does not affect anything.

blesh:
This looks great.

:shipit:

package.json
```diff
@@ -73,8 +73,8 @@
...
-    "compile_dist_es6_for_docs": "tsc                                 ./dist/es6/src/Rx.ts ./dist/es6/src/add/observable/of.ts ./dist/es6/src/MiscJSDoc.ts -m es2015   --sourceMap --outDir ./dist/es6 --target ES6 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --suppressImplicitAnyIndexErrors --moduleResolution node",
+    "compile_dist_es6": "tsc                                          ./dist/es6/src/Rx.ts ./dist/es6/src/add/observable/of.ts                             -m es2015   --sourceMap --outDir ./dist/es6 --target ES5 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --suppressImplicitAnyIndexErrors --moduleResolution node --noEmitHelpers --lib es5,es2015.iterable,es2015.collection,es2015.promise,dom ",
+    "compile_dist_es6_for_docs": "tsc                                 ./dist/es6/src/Rx.ts ./dist/es6/src/add/observable/of.ts ./dist/es6/src/MiscJSDoc.ts -m es2015   --sourceMap --outDir ./dist/es6 --target ES6 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --suppressImplicitAnyIndexErrors --moduleResolution node --noEmitHelpers",
```
kwonoj: is this needed if this build is only going to be used to generate document?

package.json
```diff
@@ -73,8 +73,8 @@
...
-    "compile_dist_es6": "tsc                                          ./dist/es6/src/Rx.ts ./dist/es6/src/add/observable/of.ts                             -m es2015   --sourceMap --outDir ./dist/es6 --target ES6 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --suppressImplicitAnyIndexErrors --moduleResolution node",
-    "compile_dist_es6_for_docs": "tsc                                 ./dist/es6/src/Rx.ts ./dist/es6/src/add/observable/of.ts ./dist/es6/src/MiscJSDoc.ts -m es2015   --sourceMap --outDir ./dist/es6 --target ES6 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --suppressImplicitAnyIndexErrors --moduleResolution node",
+    "compile_dist_es6": "tsc                                          ./dist/es6/src/Rx.ts ./dist/es6/src/add/observable/of.ts                             -m es2015   --sourceMap --outDir ./dist/es6 --target ES5 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --suppressImplicitAnyIndexErrors --moduleResolution node --noEmitHelpers --lib es5,es2015.iterable,es2015.collection,es2015.promise,dom ",
```
kwonoj: I know this is nitpicking and should not block check in, might need to rename script since it doesn't target es6 anymore.

kwonoj:
LGTM too. 👍

package.json
Brooooooklyn: updated 👌

PR #2100, opened 2016-11-03
fix(first): will now only emit one value in recursive cases

### LINKED ISSUE #2098: first() produce a lot more than 1 data in some recursive situation.
It is pretty hard to describe in text. The code below might be more understandable.

I have a `Subject` that is connected with `first()` and subscribed with a observer that will fire a data synchronously to the same `Subject` with `next()`. It turns out that the observer receives a lot of data even though it is chained with `first()`.

I found this bug in more personal complicated code, but below is a simplified repro one. Also, I'm not sure if this is expected behavior or not.

**RxJS version:**
5.0.0-beta.12

**Code to reproduce:**
```js
const valueSubject = new Subject();
const result = [];

valueSubject.first().subscribe(val => {
    result.push(val);
    valueSubject.next(false);
});

valueSubject.next(true);

// Expect result.length to be 1.
```

**Expected behavior:**
Data should be received once.

**Actual behavior:**
Large amount of data fired to observer. Sometimes, the Error `Maximum call stack size exceeded` shows up.

**Additional information:**
When data is pushed asynchronously, the test will pass. Here is the asynchronous test in Jasmine:
```js
it('should behave correctly', (done) => {
    const valueSubject = new Subject<boolean>();
    const result: boolean[] = [];

    valueSubject.first().subscribe(val => {
        result.push(val);
        setTimeout(() => valueSubject.next(false), 0);
    });

    valueSubject.next(true);

    setTimeout(() => {
        expect(result.length).toBe(1);
        done();
    }, 100);
});
```

In this case, the test passes.

blesh:
adds guard after first emitted value to prevent reentrant behavior

fixes #2098

### DIFF
--- a/spec/operators/first-spec.ts
+++ b/spec/operators/first-spec.ts
@@ -42,6 +42,20 @@ describe('Observable.prototype.first', () => {
     expectSubscriptions(e1.subscriptions).toBe(sub);
   });
 
+  it('should only emit one value in recursive cases', () => {
+    const subject = new Rx.Subject<number>();
+    const results = [];
+
+    subject.first().subscribe(x => {
+      results.push(x);
+      subject.next(x + 1);
+    });
+
+    subject.next(0);
+
+    expect(results).to.deep.equal([0]);
+  });
+
   it('should propagate error from the source observable', () => {
     const e1 = hot('---^---#');
     const expected =  '----#';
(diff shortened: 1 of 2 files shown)


kwonoj:
LGTM

kwonoj:
Travis failure seems coming from yarn packging installation, not sure why it fails suddenly though. Maybe revert to npm and need discuss introducing yarn seperately.

PR #2103, opened 2016-11-04
style(linting): enable 'noImplicitThis' compiler option

saneyuki:
**Description:**

- This intent to make this type explicitly.
  - This would make our codes more type safe with a compiler enforcement.
  - This makes our API documentation better.
- I think this would not be a breaking change by these reasons:
    - There are no problem if we pass the function which does not has this specifying.
    - There are no problem if we pass the function which has this: any or `this`.
      - This case is valid with the actual behavior.
    - If we pass the function which has invalid `this` type
      then, basically, its code would be wrong.

  

- Fix https://github.com/ReactiveX/rxjs/issues/1971.
- Related: https://github.com/ReactiveX/rxjs/pull/2078


jayphelps:
I don't believe this and #2104 are features (i.e. publicly accessible API/behavior additions). They are more like a refactor. Anyone else have thoughts?

kwonoj:
@jayphelps I agree, as these change should not impact public api surface. 

for this one, as shared in original issue I'm not in great favor of this changes while I would not suggest to block check in completely. My experience with this flag usually caused unnecessary pain point for some tricky cases

david-driscoll:
@kwonoj do you have specific cases? With the new this syntax I've not had any issues.  I've had lots of problems with tslint in the past though.

kwonoj:
@david-driscoll sorry, it was bit while and I don't have examples for now.

saneyuki:
@jayphelps 

> I don't believe this and #2104 are features (i.e. publicly accessible API/behavior additions). They are more like a refactor. Anyone else have thoughts?

Okay. I'll change to `style(linting)`.

jayphelps:
LGTM

Issue #1969, opened 2016-09-23
feat(TypeScript): enable 'noUnusedLocals' options


saneyuki:
- This would be useful to clean up our codes.
- We detect unused variables at compile time.
- We can replace tslint's `no-unused-variable`.

saneyuki:
And we could be able to disable tslint's `no-unused-variable` rule. It might be decrease the time to run tslint

kwonoj:
@saneyuki , please aggregate into single issue to track enabling TS 2.0 specific features. I don't think it's need to be tracked as separated issue for each options.

kwonoj:
Long-hanging task tracking issue https://github.com/ReactiveX/rxjs/issues/1837 can be used for those purpose.

saneyuki:
@kwonoj Ah, okay. I'll change title and description of #1837

saneyuki:
Duplicated #1837

saneyuki:
I'll reopen as a separated tracking issue.

saneyuki:
If we replace tslint's `no-unused-variable`, the exec time of tslint will be decreased.

## The result of `time npm run lint_src` 
env: MacBook Pro (Retina, 15-inch, Mid 2014), 2.8 GHz Intel Core i7


| | real | user | sys|
|---|----|------|-------
| before |  6.85 | 6.97 | 0.28|
| after |  5.26 | 5.16 | 0.29 |

kwonoj:
as this issue is discussed in https://github.com/ReactiveX/rxjs/pull/2104 and concluded to not to introduce compiler level option, I believe issue can be closed.

PR #2104, opened 2016-11-04
style(linting): enable ' --noUnusedLocals' option

saneyuki:
**Description:**

- This option detects more unused variables than tslint.
- We detect unused variables at compile time without doing lint.
- We can replace tslint's `no-unused-variable` (But not now).
   - We need to separate tslint configurations for src/ from single file.
- This pull request does not change a public interface.
  - Almost changes are related to private interface.

- Fix https://github.com/ReactiveX/rxjs/issues/1969
- Depends on https://github.com/ReactiveX/rxjs/pull/2071


src/observable/ArrayLikeObservable.ts
```diff
@@ -55,7 +55,8 @@ export class ArrayLikeObservable<T> extends Observable<T> {
 
   protected _subscribe(subscriber: Subscriber<T>): TeardownLogic {
     let index = 0;
-    const { arrayLike, scheduler } = this;
+    const arrayLike = this.arrayLike;
+    const scheduler = this.scheduler;
```
jayphelps: Any reason? I think we want to keep our existing, destructure style. Is there a subtle difference I'm missing?

src/operator/last.ts
```diff
@@ -58,7 +58,7 @@ class LastSubscriber<T, R> extends Subscriber<T> {
   constructor(destination: Subscriber<R>,
               private predicate?: (value: T, index: number, source: Observable<T>) => boolean,
               private resultSelector?: ((value: T, index: number) => R) | void,
-              private defaultValue?: any,
+              defaultValue?: any,
```
jayphelps: Why did this need to be changed to not be a private property?

saneyuki: `defaultValue` is not used in `LastSubscriber`

saneyuki: And this option make it the error.

src/observable/ArrayLikeObservable.ts
saneyuki: This is for to avoid a [TypeScript compiler's bug](https://github.com/Microsoft/TypeScript/issues/11324) (It will be [fixed in TS2.1](https://github.com/Microsoft/TypeScript/pull/11998)).

src/operator/delay.ts
```diff
@@ -134,7 +134,7 @@ class DelaySubscriber<T> extends Subscriber<T> {
 }
 
 class DelayMessage<T> {
-  constructor(private time: number,
-              private notification: any) {
+  constructor(public time: number,
+              public notification: Notification<T>) {
```
saneyuki: The old code was wrong. These members should be public to be used in [`DelaySubscriber`](https://github.com/ReactiveX/rxjs/blob/master/src/operator/delay.ts#L82)

jayphelps:
Discussed this one at our last meeting and we decided we want to keep the existing behavior because we haven't noticed any practical issues with our current way and this makes debugging harder--leaving unused variables while we're fixing something, but being reminded if we forget by the separate linting phase before commit.

Thanks @saneyuki for your contribution!

More info: https://github.com/ReactiveX/rxjs-core-notes/blob/master/2016-11/november-8.md#internal-lint-changes-2104-2103

saneyuki:
@jayphelps 

Okay. I accept this decision.

I might open a new pull request which remove some needless code detected by `noUnusedLocals` option but not enable it.

PR #2107, opened 2016-11-05
fix(fromEvent): Throw if event target is invalid

jayphelps:
Added `removeEventListener` mock to #1852 and rebased (didn't have permission to push -f to his  branch.


blesh:
:shipit: :ship:

PR #2109, opened 2016-11-05
chore(fix-test-helpers): fix Symbol.observable refernce for Safari 8

blesh:
This fix is related to saucelabs tests failing in Safari 8.
The suspicion is that it is a problem with the fact a line of code was
using `Symbol.observable` directly instead of `56526symbolObservable`
since we're no longer polyfilling `Symbol.observable` unless `Symbol`
exists properly. Might also be related to `symbol-observable` dependency.


jayphelps:
LGTM

PR #2114, opened 2016-11-06
test(browser): polyfill promise for unsuppoerted browser

kwonoj:
**Description:**
It seems IE browser testing fails by not able to look up promise implementation, which some of spec creates as test fixture. This PR patches global if promise doesn't exists to let test works correctly.


jayphelps:
Looking good, will require some typings love after #2120 so waiting on that

kwonoj:
Yes, I'll update once pr checked in.

kwonoj:
Updated PR to align with latest changes.

jayphelps:
LGTM

PR #2118, opened 2016-11-07
Fix the docs site, bug #2111

### LINKED ISSUE #2111: Documentation Generation Broken
Currently the documentation, when generated, is missing all of the operator documentation. This is a bit of a fire we need to put out because they've been generated and pushed to http://reactivex.io/rxjs ... so we need to fix this ASAP, (before any other changes are merged) and republish the docs.

staltz:
Fixes #2111.

There were two things to fix. I made one commit for each fix. Check the commit messages for more info.


jayphelps:
I broke your build. Digging into it..one of those "works on my machine...."

jayphelps:
LGTM

PR #2119, opened 2016-11-07
feat(filter, find, first, last): support type guard predicates that don't requiring casting

rob3c:
**Description:**
These updates allow the use of type guard predicates to flow narrowed types through Observable operator chains using the `filter`, `find`, `first`, and `last` operators without requiring casting. Unfortunately, direct use of `typeof x === 'string'` style boolean predicates are not able to be supported yet in this context, even though typescript understands how to use them in other contexts (e.g. `if/else` statements and such). However, type guard functions can still reduce a lot of boilerplate casting noise with these new definitions.  Examples for each operator appear below.

```typescript
// x is `Observable<string | number>`
let x = Observable.from([1, 'aaa', 3, 'bb']);

// This type guard will narrow a `string | number` to a string in the examples below
let isString = (x: string | number): x is string => typeof x === 'string';

// Here, `s` is a string in the second filter predicate after the type guard (yay - intellisense!)
let guardedFilter = x.filter(isString).filter(s => s.length === 2); // Observable<string>
// In contrast, this type of regular boolean predicate still maintains the original type
let boolFilter = x.filter(s => typeof s === 'number'); // Observable<string | number>

// After the type guard `find` predicate, the type is narrowed to string
let guardedFind = x.find(isString).filter(s => s.length > 1).map(s => s.substr(1)); // Observable<string>  
// In contrast, a boolean predicate maintains the original type
let boolFind = x.find(x => typeof x === 'string'); // Observable<string | number>

// After the type guard `first` predicates, the type is narrowed to string
let guardedFirst1 = x.first(isString).filter(s => s.length > 1).map(s => s.substr(1)); // Observable<string>
let guardedFirst2 = x.first(isString, s => s.substr(0)).filter(s => s.length > 1); // Observable<string>

// Without a resultSelector, `first` maintains the original type (TS can't do this yet)
let boolFirst1 = x.first(x => typeof x === 'string', null, ''); // Observable<string | number>
// `first` still uses the `resultSelector` return type, if it exists.
let boolFirst2 = x.first(x => typeof x === 'string', s => ({str: `${s}`}), {str: ''}); // Observable<{str: string}>

// Same for `last`:
let guardedLast1 = x.last(isString).filter(s => s.length > 1).map(s => s.substr(1)); // Observable<string>
let guardedLast2 = x.last(isString, s => s.substr(0)).filter(s => s.length > 1); // Observable<string>
let boolLast1 = x.last(x => typeof x === 'string', null, ''); // Observable<string | number>
let boolLast2 = x.last(x => typeof x === 'string', s => ({str: `${s}`}), {str: ''}); // Observable<{str: string}>
```

This enhances the initial attempt to integrate type guards for narrowing types in operator chains in https://github.com/ReactiveX/rxjs/pull/1936. I had initially asked about a solution to this in https://github.com/Microsoft/TypeScript/issues/7657#issuecomment-258544213, but arrived at this solution before there was any response.

### DIFF
--- a/src/operator/filter.ts
+++ b/src/operator/filter.ts
@@ -44,9 +44,10 @@ import { TeardownLogic } from '../Subscription';
  * @owner Observable
  */
 /* tslint:disable:max-line-length */
-export function filter<T>(this: Observable<T>, predicate: (value: T, index: number) => boolean, thisArg?: any): Observable<T>;
-export function filter<T, S extends T>(this: Observable<T>, predicate: (value: T, index: number) => value is S, thisArg?: any): Observable<S>;
-/* tslint:disable:max-line-length */
+export function filter<T, S extends T>(this: Observable<T>,
+                                       predicate: ((value: T, index: number) => boolean) |
+                                                  ((value: T, index: number) => value is S),
+                                       thisArg?: any): Observable<S>;
 export function filter<T>(this: Observable<T>, predicate: (value: T, index: number) => boolean,
                           thisArg?: any): Observable<T> {
   return this.lift(new FilterOperator(predicate, thisArg));
--- a/src/operator/find.ts
+++ b/src/operator/find.ts
@@ -36,8 +36,10 @@ import { Subscriber } from '../Subscriber';
  * @owner Observable
  */
 /* tslint:disable:max-line-length */
-export function find<T>(this: Observable<T>, predicate: (value: T, index: number, source: Observable<T>) => boolean, thisArg?: any): Observable<T>;
-export function find<T, S extends T>(this: Observable<T>, predicate: (value: T, index: number, source: Observable<T>) => value is S, thisArg?: any): Observable<S>;
+export function find<T, S extends T>(this: Observable<T>,
+                                     predicate: ((value: T, index: number, source: Observable<T>) => boolean) |
+                                                ((value: T, index: number, source: Observable<T>) => value is S),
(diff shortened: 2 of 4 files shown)


kwonoj:
Change looks fine to me, /cc @david-driscoll

david-driscoll:
LGTM

kwonoj:
LGTM

intellix:
Having issues with this in TS 2.4 within my Angular codebase.

@angular/router exports the following type:
```ts
export type Event = NavigationStart | NavigationEnd | NavigationCancel | NavigationError |
    RoutesRecognized | RouteConfigLoadStart | RouteConfigLoadEnd;
```

Pre 2.4 I was doing:
```ts
navigations: Observable<NavigationEnd> = this.router.events.filter(event => event instanceof NavigationEnd);
```

But now get: 
```
Type 'Observable<Event>' is not assignable to type 'Observable<NavigationEnd>'.
  Type 'Event' is not assignable to type 'NavigationEnd'.
    Type 'NavigationStart' is not assignable to type 'NavigationEnd'.
      Property 'urlAfterRedirects' is missing in type 'NavigationStart'.
```

Which looks to me like it's not understanding my instanceof filter, but it only started giving me trouble in 2.4 which apparently strengthens Generics somewhat: https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#improved-inference-for-generics

Issue #2116, opened 2016-11-07
rxjs/Rx has no default export


Modellium:
Hey,

Not sure if it applies as an "issue", but it sure would be great to update the README documentation.

Where it say: 

"To import the entire core set of functionality:

import Rx from 'rxjs/Rx'; "

There is no default export on rxjs/Rx and I don't think there have been for some time.

It can be confusing

Adam Kettani

jayphelps:
Thanks! I believe this is the same as #2085. Whether or not you see a default export depends on the loader you're using...we're going to try to rework our README instructions to be more explicit about these sorts of things.

AdamKettani:
Alright thank you, what loader is currently working?

awerlang:
Add this to your `tsconfig.json`:

```json
  "compilerOptions": {
   ...
    "allowSyntheticDefaultImports": true,
  }
```

sikharsaikia:
Changing Changing import Rx from 'rxjs/Rx'; to import * as Rx from 'rxjs/Rx'; fixes it. Thanks to Denis https://github.com/ReactiveX/rxjs/issues/2085

benlesh:
No longer relevant in v6

PR #2124, opened 2016-11-09
chore(example): add basic TS import bootstrap example

kwonoj:
**Description:**
This PR is  discussion point for initial bootstrap(module import, mostly) example of RxJS on various compiler / bundling configurations. Each example should include

- how to import all operators of Rx via entry point
- how to selectively import specific operator as needed
- npm script to invoke each cases
- (and link from readme)

discussion point
- [ ] code organization: where is best place to locate those? under main repo (like this PR) vs. separate repo?
- [ ] what's most compelling code snippet to show after module is imported?
- [ ] which compiler / bundler should be included? (babel, webpack, etcs..)

separately
- [ ] maybe instead of installing published package, always install from master (via `npm pack` generated) to use these example as very basic package sanity test as well?

#2116 and others

### DIFF
--- a/examples/typescript/import-all.ts
+++ b/examples/typescript/import-all.ts
@@ -1,0 +1,6 @@
+import * as Rx from 'rxjs';
+
+Rx.Observable
+  .range(1, 3)
+  .map(x => x * 2)
+  .subscribe(console.log.bind(console));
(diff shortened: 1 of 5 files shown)


kwonoj:
I'm closing this since this hasn't got attention and think it's not being needed at this moment.

PR #2125, opened 2016-11-10
style(linting): remove needless members detected by noUnusedLocals

saneyuki:
**Description:**

- This retake the par of https://github.com/ReactiveX/rxjs/pull/2104 to remove needless members or type parameters.
  -  But this does not enable `noUnusedLocals` option according to the decision of #2104.
- This would not be a breaking change.
  - Because these change only internal signatures.

- https://github.com/ReactiveX/rxjs/pull/2104
- https://github.com/ReactiveX/rxjs/issues/1969

### DIFF
--- a/src/Operator.ts
+++ b/src/Operator.ts
@@ -1,6 +1,7 @@
+import { Observable } from './Observable';
 import { Subscriber } from './Subscriber';
 import { TeardownLogic } from './Subscription';
 
 export interface Operator<T, R> {
-  call(subscriber: Subscriber<R>, source: any): TeardownLogic;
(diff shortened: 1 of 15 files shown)


saneyuki:
rebased!

saneyuki:
> Coverage decreased (-0.1%) to 97.59% when pulling 

Ugh, mysterious down is again 😢

blesh:
@saneyuki ... I'm so sorry. Given that this PR has gone stale, and there hasn't been much interest from the rest of the team. I'm going to close it for now. cc @kwonoj just in case.

saneyuki:
@blesh 

okay. I will open a new pull request which is more break downed.**

PR #2127, opened 2016-11-10
chore(build): Using local installed tsc

imcotton:
To ensure using `./node_modules/.bin/tsc` in npm run-script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,8 @@
     "test_browser": "npm-run-all build_spec_browser && opn spec/support/mocha-browser-runner.html",
     "test": "npm-run-all clean_spec build_spec test_mocha clean_spec",
     "tests2png": "npm run build_spec && mkdirp tmp/docs/img && mkdirp spec-js/support && shx cp spec/support/*.opts spec-js/support/ && mocha --opts spec/support/tests2png.opts spec-js",
-    "watch": "watch \"echo triggering build && npm run build_test && echo build completed\" src -d -u -w=15"
+    "watch": "watch \"echo triggering build && npm run build_test && echo build completed\" src -d -u -w=15",
+    "tsc": "tsc"
   },
   "repository": {
     "type": "git",


kwonoj:
I'm bit unsure if this is necessary change. If someone uses globally installed typescript compiler and intend to picks it up, it's probably each own dev may need take care of those. 

Also, `postinstall` lifecycle script is not being used in this repo. We've been used and take it out as it does not works as expected.

imcotton:
Thanks for noticing `postinstall`, I've removed it.

I think having local installed tsc to run would help to isolate the potential side effect, so that in case dosen't require to uninstall global version (e.g. v2.1) which incompatible with one specified in `package.json`.

jayphelps:
Can you clarify a few examples that this solves? I may totally be unfamiliar with voodoo at work here..it's my understanding that this is only useful if someone uses `npm run tsc` (or similar). If we use `tsc` directly in our other run-scripts, e.g.

```
"build_spec": "tsc --project ./spec --pretty"
```

This new code won't make any difference, will it? We'll already be picking up the locally installed `tsc` bin, as long as the `typescript` npm module is installed.

kwonoj:
Afaik this is related to globally installed pkg. If typescript is installed globally and its path resolving is happening prior to npm's local module resolution in script, it is possible npm can pickup global one instead of locally installed module. I don't think this is usual case and even in that case it's consumer's own to correctly manage installed global packages, so bit down (not strongly though) on this changes.

jayphelps:
@kwonoj It's my understanding (and testing appears to confirm) that any locally installed npm package with a bin will shadow any possibly installed bin in your path. Is that not happening?

Even if it didn't, I'm still not understanding how this PR would change any behavior in our scripts as-is? Or is this just for us to be able to run `npm run tsc` for other one-off purposes?

kwonoj:
>  I believe it does so by adding node_modules/.bin to your PATH, but perhaps the order in which it inherits existing PATHS depends on other configuration. 

: Long time ago I've seen somewhat similar cases of (not exactly to this though) that path resolving works weirdly due to global npm path resolve with npm resolves local module when run script. So comment above I said is 'it is kind of possible' case scenarios.

> Even if it didn't, I'm still not understanding how this PR would change any behavior in our scripts as-is? Or is this just for us to be able to run npm run tsc for other one-off purposes?

: this one's actually you're correct, for some reason I was naively thought this change will resolve those module path resolution issues. Not sure why I thought in those way. :/

imcotton:
@jayphelps @kwonoj 

You're right, this change should only useful for `npm run tsc`, I think I'd illusionary by the recent `yarn install` bumping local typescript version from 2.0.x to 2.1, at the meantime I've v2.1 installed globally so that lead me to the wrong thinking :dizzy_face:

Sorry for the trouble guys...

jayphelps:
@imcotton no worries! none of us know what we're doing.

image

PR #2129, opened 2016-11-12
Update migration docs

blesh:
The migration docs need a little love. The idea here was just to make it easier to find operators that you're trying to migration, and provide a little more guidance and context for people trying to deal with migrating operators we aren't going to add or haven't added yet.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -54,7 +54,7 @@ enabling "composite" subscription behavior.
 |---|---|
 |`amb`|`race`|
 |`and`|Not yet implemented|
-|`asObservable`|Not yet implemented|
+|`asObservable`|Exists on `Subject` only|
 |`average`|Not yet implemented|
 |`bufferWithCount`|`bufferCount`|
 |`bufferWithTime`|`bufferTime`|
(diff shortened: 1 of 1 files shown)


jayphelps:
Love love love this. How bout we change `Not yet implemented` to just `No longer implemented`. Because the former heavily implies to people that we _will_ implement it soon; we've had several people point to it as justification. We certainly may add some of them, either to core or to some future addon libraries, but I don't want to imply anything. If you're good with that, I'm happy to change it for you, just reply 👍

kwonoj:
I think this PR is ready to check in but only LGTM prevents it? (hope my LGTM unlocks this)

jayphelps:
@kwonoj I still think we should change `Not yet implemented` to `No longer implemented` or something else non-committal sounding. Maybe `No longer provided`. Any thoughts Mr. @blesh?

kwonoj:
@jayphelps totally agree to replace string other than current one. Not sure about what would be best to describe it though. (maybe just big red large **'X'**?)

blesh:
@jayphelps updated... what do you think?

jayphelps:
LGTM

PR #2131, opened 2016-11-15
chore(CHANGELOG): add message about chai typings fix

jayphelps:
The commit message for #2120 was not marked as `fix()` so it wasn't picked up. Understand why it wasn't, cause...it's not really a "bugfix" in the sense of a runtime bug...just a compile/typings bug.

Cc/ @kwonoj @blesh


kwonoj:
LGTM

PR #2135, opened 2016-11-16
Cancel scheduled timeout work, if no longer needed

### LINKED ISSUE #2134: .timeout doesn't cancel setTimeout
**RxJS version:**
rxjs beta.12
**Code to reproduce:**

**Expected behavior:**
cancel the internal setTimeout when it's not needed anymore (after first response)

**Actual behavior:**
setTimeout still resolves which delays universal since zone.js tracks it
**Additional information:**

Angular Universal uses zone.js to keep track of async calls before rendering. There seems to be a problem with `.timeout` on an http call
here's a repro
https://github.com/gdi2290/universal-starter-rxjs-timeout-issue/blob/master/src/app/shared/api.service.ts#L23
npm run watch:dev
comment out that line 
there are some console logs on the server that show the stable time taking 2000ms with the timeout rather than less than 1ms

jayphelps:
Fixes #2134

This ties the action subscription returned from a call to `schedule()` to the lifespan of the subscriber, so it handles error/complete. On reschedules it more explicitly cancels, if there's an existing.

***

I _think_ this fixes it, but to be honest, I can't be sure. It's not immediately clear _how_ to test it. At first glance, it doesn't appear we're testing this possible bug in the other operators which could have similar problems. e.g. debounceTime. So other operators could be hiding the fact that they don't clean up properly, because of guards in their work callback, or the subscriptions that protect against values continuing after error/complete.

I tried peeking into the internals of TestScheduler to see if maybe I could grab a reference to the `VirtualAction` and check `action.pending`, but _because_ it's entirely virtual, the action comes and goes only inside a `testScheduler.flush()`.

Perhaps @trxcllnt aka Mr. Scheduler can halp?


mattpodwysocki:
@jayphelps perhaps there's a test we can write for this?  I don't see anything here to that regard but should be.

jayphelps:
@mattpodwysocki see my original post 😆

mattpodwysocki:
@jayphelps * sigh * reading is fundamental...

src/operator/timeout.ts
```diff
@@ -45,6 +46,8 @@ class TimeoutOperator<T> implements Operator<T, T> {
 class TimeoutSubscriber<T> extends Subscriber<T> {
   private index: number = 0;
   private _previousIndex: number = 0;
+  private action: Subscription = null;
+
   get previousIndex(): number {
```
trxcllnt: If we're tracking the scheduled action subscription, do we still need `previousIndex` etc.? Same question for `hasCompleted`.

src/operator/timeout.ts
```diff
@@ -65,18 +68,34 @@ class TimeoutSubscriber<T> extends Subscriber<T> {
...
     }
   }
 
   private scheduleTimeout(): void {
-    let currentIndex = this.index;
-    this.scheduler.schedule(TimeoutSubscriber.dispatchTimeout, this.waitFor, { subscriber: this, index: currentIndex });
+    const currentIndex = this.index;
+    const timeoutState = { subscriber: this, index: currentIndex };
+
+    this.cancelTimeout();
```
trxcllnt: So here it seems like we can do this:

```es6
const { action } = this;
if (!action) {
  this.add(this.action = this.scheduler.schedule(...));
} else {
  action.schedule(this.waitFor, timeoutState);
}
```

Instead of scheduling an action, throwing it away, and scheduling again, we can recycle the Action and reschedule it for a new due time. This should be both more efficient, and allow us to remove more code from this file.

trxcllnt:
@jayphelps let's use `lift`! If the changes in my review pan out, something like this should do the trick:

```typescript
it('should unsubscribe from the scheduled timeout action when timeout is unsubscribed early', () => {
  const e1 =   hot('--a--b--c---d--e--|');
  const e1subs =   '^         !        ';
  const expected = '--a--b--c--        ';
  const unsub =    '          !        ';

  const result = e1
    .lift(function(source) {
      const timeoutSubscriber = this;
      const { action } = timeoutSubscriber; // get a ref to the action here,
      timeoutSubscriber.add(() => {         // because it'll be null by the
        if (!action.canceled) {             // time we get into this function.
          throw new Error('TimeoutSubscriber scheduled action wasn\'t canceled');
        }
      });
      return source._subscribe(timeoutSubscriber);
    })
    .timeout(50, null, rxTestScheduler);

  expectObservable(result, unsub).toBe(expected);
});
```

src/operator/timeout.ts
jayphelps: not sure, but `hasCompleted` indeed seems extraneous. I couldn't figure out why there is an index (aka count) check anyway? Why would it ever not be equal?

src/operator/timeout.ts
jayphelps: 👍 I'm on board with reusing the action.

> and allow us to remove more code from this file.

What other code could be removed?

trxcllnt:
@jayphelps I'm preparing a PR for your PR, if that's ok :-)

trxcllnt:
@jayphelps I fixed the lint, rebased, tried and failed to push straight to this PR ref (GH doesn't like that), so I re-PR'd to your timeout branch.

package.json
```diff
@@ -89,7 +89,7 @@
     "prepublish": "shx rm -rf ./typings && typings install && npm run build_all",
     "publish_docs": "./publish_docs.sh",
     "test_mocha": "mocha --opts spec/support/default.opts spec-js",
-    "debug_mocha": "node-debug _mocha --opts spec/support/debug.opts spec-js",
+    "debug_mocha": "node --inspect --debug-brk ./node_modules/.bin/_mocha --opts spec/support/debug.opts spec-js",
```
blesh: This change seems unrelated although I agree with it.

spec/operators/timeout-spec.ts
```diff
@@ -139,4 +139,27 @@ describe('Observable.prototype.timeout', () => {
...
   });
+
+  it('should unsubscribe from the scheduled timeout action when timeout is unsubscribed early', () => {
+    const e1 =   hot('--a--b--c---d--e--|');
+    const e1subs =   '^         !        ';
+    const expected = '--a--b--c--        ';
+    const unsub =    '          !        ';
+
+    const result = e1
+      .lift(function(source) {
```
blesh: I'm not sure this test is very straightforward. Perhaps it could be refactored into using the `let` operator, and returning a more basic observable

spec/operators/timeoutWith-spec.ts
```diff
@@ -260,4 +260,30 @@ describe('Observable.prototype.timeoutWith', () => {
...
+  it('should unsubscribe from the scheduled timeout action when timeout is unsubscribed early', () => {
+    const e1 =     hot('---a---b-----c----|');
+    const e1subs =     '^          !       ';
+    const e2 = cold(              '-x---y| ');
+    const e2subs =     '           ^  !    ';
+    const expected =   '---a---b----x--    ';
+    const unsub =      '              !    ';
+
+    const result = e1
+      .lift(function(source) {
```
blesh: Same as above

package.json
jayphelps: @trxcllnt I believe this is you, cool for me to remove?

spec/operators/timeout-spec.ts
jayphelps: can we pair on this? I couldn't figure out how to test this at all, so @trxcllnt came up with this solution and it's still not clear to me how we could do it better. 💃

package.json
trxcllnt: as long as it's in a follow up PR that's fine with me. I didn't want to clutter up the PR list with internal config changes that only affect the devs working on the project, but I leave it up to your discretion

spec/operators/timeout-spec.ts
blesh: @trxcllnt ugh.. this is a case where we're abusing the fact that `Operator` has a `call` method, and we're therefor passing a function to `lift` here? This is really convoluted, it took me a while to figure that out. If anything this test reminded me why `Operator` probably shouldn't have a `call` method.

I think we need to find a different way to test this.

blesh: @jayphelps I think the way to test this, and other things like it, would be to add functionality to the TestScheduler so that it tracks the Actions it's created and those Actions can be examined to see at what tick they were cancelled. For now, we can leave these tests in here, but they'll need comments explaining what they're doing and why it works.

jayphelps: > Basically we need comments explaining why lift is magically accepting a function here and not an Operator instance.

We could simply pass an object with a call method.

```js
.lift({
  call(source) { ... }
})
```

trxcllnt: @blesh this is _exactly_ why Operator has a call method. In an ideal world an operator _would_ just be a function, but because JS sucks, we've had to make them classes.

trxcllnt:
@blesh alternatively, we should be able to define `Operator` as a [hybrid type](https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Interfaces.md#hybrid-types) (so it's both an Object with a `call` method, and a Function with `thisArg: Subscriber<T>`), but my TypeScript-foo was not strong enough to defeat the compiler. maybe @david-driscoll @kwonoj could help?

trxcllnt:
@blesh what are the blockers to merging this? @jayphelps if possible, could you allow adding commits to this PR from other collaborators?

jayphelps:
@trxcllnt the edit bit is already enabled:

image


We discussed this PR today, minor changes needed: https://github.com/ReactiveX/rxjs-core-notes/blob/master/2016-12/december-20.md#review-prsissues

> Needs comments on why we're using lift this way, also remove npm script changes.

I won't be able to get to it for at least a couple days, but others with edit bit are welcome to modify.

spec/operators/timeout-spec.ts
jayphelps: Ben is OK with this for now, but wants some inline comments about why we're doing

trxcllnt:
ping @jayphelps https://github.com/jayphelps/rxjs/pull/3

trxcllnt:
lol wut

kwonoj:
@trxcllnt it is expected, latest change now guards to have type definition import for test cases explicitly if test case has been modified (remaining test will be updated gradually, it's my task) 

basically what you need is import type definitions for `hot`, `cold` like https://github.com/ReactiveX/rxjs/pull/2202/files#diff-4b0bd2db7bc00156389d58a6ee4d8e1cR6 . just copy-paste it and update PR would be sufficient. (if there's type error, it need to be fixed as well)

trxcllnt:
@kwonoj thanks for the heads up. @jayphelps I've made the fixes in https://github.com/jayphelps/rxjs/pull/4

kwonoj:
I'm going to lower commit message checker to warning so ignore failure for now please.

blesh:
LOL... why are there so many commits in this PR now?

trxcllnt:
@blesh lint/spec typings fixes and merge commits/conflicts *le sigh*

kwonoj:
@trxcllnt sorry for inconvenience but rebase to master once again will resolve commit message checker failure (though you'll see warning still).

blesh:
Ugh. I'm not sure what to do with this PR. It's older that dirt. It's a solid fix, but there's something crazy going on with the commits. Can we resubmit this one and/or clean it up?

jayphelps:
I'm not following all of @trxcllnt's additions. If he's not available to fix, we can abandon this PR and I can try to fix the underlying problem from scratch again.

trxcllnt:
@jayphelps @blesh they're just merge commits from master to resolve the merge conflicts

PR #2138, opened 2016-11-17
fix(partition): handles `thisArg` as expected

jooyunghan:
**Description:**

This fix ensures that the `partition` will call `predicate` with`thisArg` as expected if it is specified.

This fix ensures that the `partition` will call `predicate` with`thisArg` as expected if it is specified.


jayphelps:
bravo! 👏 LGTM

PR #2141, opened 2016-11-18
fix(timeout): wrap customError with TimeoutError

kwonoj:
**Description:**

This PR updates behavior of `timeout` when specify `errorToSend` for custom timeout message. Since custom timeout message can be anything (literally `any`), if caller does not specify it as timeouterror, when it's thrown it does not deliver context of error thrown. This PR checks type of error, wraps into timeout error to have additional context to debug. 

with given example snippet
`Rx.Observable.never().timeout(10, 'customerr').subscribe(console.log.bind(console));`

*Before*
```js
D:\github\rxjs\dist\cjs\scheduler\AsyncScheduler.js:45
            throw error;
            ^
customerr
```

*After*
```js
D:\github\rxjs\dist\cjs\scheduler\AsyncScheduler.js:45
            throw error;
            ^
TimeoutError: customerr
    at new TimeoutError (D:\github\rxjs\dist\cjs\util\TimeoutError.js:17:26)
    at NeverObservable.timeout (D:\github\rxjs\dist\cjs\operator\timeout.js:30:17)
    at Object.<anonymous> (D:\github\rxjs\dummy.js:3:23)
    at Module._compile (module.js:573:32)
    at Object.Module._extensions..js (module.js:582:10)
    at Module.load (module.js:490:32)
    at tryModuleLoad (module.js:449:12)
    at Function.Module._load (module.js:441:3)
    at Module.runMain (module.js:607:10)
    at run (bootstrap_node.js:382:7)
```

### DIFF
--- a/src/operator/timeout.ts
+++ b/src/operator/timeout.ts
@@ -20,7 +20,8 @@ export function timeout<T>(this: Observable<T>, due: number | Date,
                            scheduler: Scheduler = async): Observable<T> {
   const absoluteTimeout = isDate(due);
   const waitFor = absoluteTimeout ? (+due - scheduler.now()) : Math.abs(<number>due);
-  const error = errorToSend || new TimeoutError();
+  const error = errorToSend instanceof Error ? errorToSend : new TimeoutError(errorToSend);
+
   return this.lift(new TimeoutOperator(waitFor, absoluteTimeout, error, scheduler));
 }
 
--- a/spec/operators/timeout-spec.ts
+++ b/spec/operators/timeout-spec.ts
@@ -1,4 +1,6 @@
 import * as Rx from '../../dist/cjs/Rx';
+import {TimeoutError} from '../../dist/cjs/util/TimeoutError';
+
 declare const {hot, cold, asDiagram, expectObservable, expectSubscriptions};
 
 declare const rxTestScheduler: Rx.TestScheduler;
@@ -19,11 +21,35 @@ describe('Observable.prototype.timeout', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
-  it('should timeout after specified timeout period and send the passed error', () => {
+  it('should timeout after specified timeout period and send the passed error as TimeoutError', () => {
     const e1 =  cold('-');
     const e1subs =   '^    !';
     const expected = '-----#';
     const value = 'hello';
+
+    const result = e1.timeout(50, value, rxTestScheduler);
+
+    expectObservable(result).toBe(expected, null, new TimeoutError(value));
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should timeout after specified timeout period and send the passed error', () => {
+    const e1 =  cold('-');
+    const e1subs =   '^    !';
+    const expected = '-----#';
+    const value = new Error('hello');
+
+    const result = e1.timeout(50, value, rxTestScheduler);
+
+    expectObservable(result).toBe(expected, null, value);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should timeout after specified timeout period and send the passed TimeoutError', () => {
+    const e1 =  cold('-');
+    const e1subs =   '^    !';
+    const expected = '-----#';
+    const value = new TimeoutError('hello');
 
     const result = e1.timeout(50, value, rxTestScheduler);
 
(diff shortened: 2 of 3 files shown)


src/operator/timeout.ts
```diff
@@ -20,7 +20,8 @@ export function timeout<T>(this: Observable<T>, due: number | Date,
                            scheduler: Scheduler = async): Observable<T> {
   const absoluteTimeout = isDate(due);
   const waitFor = absoluteTimeout ? (+due - scheduler.now()) : Math.abs(<number>due);
-  const error = errorToSend || new TimeoutError();
+  const error = errorToSend instanceof Error ? errorToSend : new TimeoutError(errorToSend);
```
jayphelps: What about instead checking if it's a string? if yes, wrap, if no, use as-is. Otherwise if someone provides their own object that doesn't extend `Error`, it would log `TimeoutError: [object Object]`

kwonoj: That's good point. One concern about allowing custom object is in that case it won't have stack trace to trace down still which this PR try to avoid for custom errors. Maybe instead of allowing error to be any kind limit its type to `Error` would be more simpler approach? (but this causes breaking changes to interface).

jayphelps: I don't fully understand the use case for providing an error **instance**. Why would you want to do `.timeout(10, new SomeError())`? Instead, wouldn't you really want to pass the class itself, to be instantiated _when_ there is an error? `.timeout(10, SomeError)`? IMO that (along with the overload of providing just a string) is the more ideal API.

kwonoj: I agree. It's kind of compromise I could think of without deep consideration. In here, I think requirement to achieve are
- allow pass custom err, either string or other object
- wrap it as `error`, to have stack trace

So in case of caller deliver its own error object directly, it won't generate stack trace as expected. Any suggestion to solve both of those?

kwonoj:
@KrishMunot would you mind elaborate bit more `then pass the param as a trace?`

blesh:
Closing in favor of #2172

src/operator/timeout.ts
jayphelps: > Indeed, Observable#timeout() isn't much more than a specialization of Observable.throw()

They are mostly unrelated. `Observable.throw` is for creating a lazy observable which only does one thing: call `observer.error()` with your provided error value.

`.timeout(x)` is an operator which takes the source observable and mirrors it, until the timeout ms have been exceeded and then errors with a timeout exception.

So it *has* to throw _some error_ for the timeout meaning three options: Require devs provide an exception class to throw, use our own internal exception class, or a combination of the two. We used to support a combination of the two, but we're not happy with the way it worked, and we don't want to let our bikeshedding hold back the v5-final release, so we made the call to remove the ability to customize the exception, and instead just us our own consistently.

There was the second point of discussion which is creating the error object at the operator callsite so we have better stack traces in the event of a timeout, that is seemingly unrelated to whether or not we accept a custom error class or not cause the argument still stands for either option. We discussed the pros vs cons of creating that instance at the callsite vs. lazy, and generally came to the consensus that the developer experience outweighs the (in practice) _very_ small hit of allocating the object we may or not not use.

I can totally listen to reason, but I'd need more guidance than saying we did it for no good reason.

mattflix: Let me correct myself on two points:  (It was late.  :)

1) Sorry, by "specialization" , what I meant was that the _usage_ of the .throw() and .timeout() operators is similar (or should be, IMO) with regard to the supplied error value.  In both cases, the caller creates and supplies the error value in the context of the operator's instantiation rather than the operation's later execution.  This is sometimes a problem (esp. when creating an ```Error``` instance is desiring an execution-relative call stack), but not necessarily so.

2) The old Rx API, at least Rx4, despite what I said, does throw an ```Error``` instance from .timeout() (with .name set to "TimeoutError").  So, Rx4, too, has an opinion about what the error value should be in this case.  Yuck.

I think the way .timeout() works is a mistake, in both the previous Rx versions and also now the "improved" version in Rx5.  Rx should have no opinion about any error value.  Rx should not require or assume an error value is an instance of ```Error``` (Javascript does not).  If I want to time-out my streams using a string, number, or other object as the error value, I should be able to (without Rx getting in my way with "helpful" wrappers or assumptions).

So...  perhaps not in this release, if we want to "fix" or improve the situation with regards to the inability to create the error value in the context of the .timeout() (or even .throw()) operator's execution, we might consider having those operators accept an "error value creation" callback function as I described earlier.

Having .throw() forward the error value supplied to it as-is and .timeout() do something entirely different with the error value supplied it, seems rather inconsistent.

PR #2143, opened 2016-11-20
fix(event): uses `Object.prototype.toString.call` on objects

estobbart:
**Description:**

Objects created with a null prototype do not have a toString method.


jayphelps:
For completeness, can you give a real world example of why someone would listen to events on an object without a prototype?

estobbart:
If you create a WebKit browser, for example GTK, and then use JavaScriptCore to add a custom JSObjectRef that fires events, JSObjectMake accepts a JSClassRef structure, to be used as the prototype.

jayphelps:
Seems legit. Thanks.

LGTM

PR #2148, opened 2016-11-21
docs(README): add missing import in example

dliv:
**Description:**

Add a missing import statement for `Observable.of`. Without this fix, the
example code would generate an error message like
'Uncaught TypeError: _Observable.Observable.of is not a function(…)'
at runtime.


jayphelps:
There were several other instances of missing imports, I amended your PR with them. Thanks!

PR #2152, opened 2016-11-23
fix(errors): Better error message when you return non-observable things,

jayphelps:
```js
item$.mergeMap(() => {})
// You provided 'undefined' where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.

item$.mergeMap(() => ({}))
// You provided an invalid object where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.
```

One thing I thought about doing is having operators (who are the ones who call `subscribeToResult`) pass in the operator name, so that the error message could also say which operator it thinks it occured in. **But**, what about operator aliases? i.e. mergeMap/flatMap, etc. And also, this could potentially be _more_ confusing if it's a custom operator that calls other operators inside of it--but perhaps that's rare enough in the wild to discount it.

So for example: `You provided 'undefined' to 'mergeMap|flatMap' where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.`

Cc/ @kentcdodds

### DIFF
--- a/src/util/subscribeToResult.ts
+++ b/src/util/subscribeToResult.ts
@@ -1,6 +1,7 @@
 import { root } from './root';
 import { isArray } from './isArray';
 import { isPromise } from './isPromise';
+import { isObject } from './isObject';
 import { Subscriber } from '../Subscriber';
 import { Observable, ObservableInput } from '../Observable';
 import { $$iterator } from '../symbol/iterator';
@@ -31,9 +32,7 @@ export function subscribeToResult<T>(outerSubscriber: OuterSubscriber<any, any>,
     } else {
       return result.subscribe(destination);
     }
-  }
-
-  if (isArray(result)) {
+  } else if (isArray(result)) {
     for (let i = 0, len = result.length; i < len && !destination.closed; i++) {
       destination.next(result[i]);
     }
@@ -55,7 +54,7 @@ export function subscribeToResult<T>(outerSubscriber: OuterSubscriber<any, any>,
       root.setTimeout(() => { throw err; });
     });
     return destination;
-  } else if (typeof result[$$iterator] === 'function') {
+  } else if (result && typeof result[$$iterator] === 'function') {
     const iterator = <any>result[$$iterator]();
     do {
       let item = iterator.next();
@@ -68,15 +67,18 @@ export function subscribeToResult<T>(outerSubscriber: OuterSubscriber<any, any>,
         break;
       }
     } while (true);
-  } else if (typeof result[$$observable] === 'function') {
+  } else if (result && typeof result[$$observable] === 'function') {
     const obs = result[$$observable]();
     if (typeof obs.subscribe !== 'function') {
-      destination.error(new Error('invalid observable'));
+      destination.error(new TypeError('Provided object does not correctly implement Symbol.observable'));
     } else {
       return obs.subscribe(new InnerSubscriber(outerSubscriber, outerValue, outerIndex));
     }
   } else {
-    destination.error(new TypeError('unknown type returned'));
+    const value = isObject(result) ? 'an invalid object' : `'${result}'`;
+    const msg = `You provided ${value} where a stream was expected.`
+      + ' You can provide an Observable, Promise, Array, or Iterable.';
+    destination.error(new TypeError(msg));
   }
   return null;
 }
--- a/spec/util/subscribeToResult-spec.ts
+++ b/spec/util/subscribeToResult-spec.ts
@@ -125,7 +125,8 @@ describe('subscribeToResult', () => {
     const subscriber = new OuterSubscriber(x => {
       done(new Error('should not be called'));
     }, (x) => {
-      expect(x).to.be.an('error', 'invalid observable');
+      expect(x).to.be.an('error');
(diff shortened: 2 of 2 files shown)


jayphelps:
If we can be sure that `subscribeToResult` is only ever used when you _return_ something, I can use that word instead of "provided". It seems [`Observable.from` does not use it](https://github.com/ReactiveX/rxjs/blob/master/src/observable/FromObservable.ts#L98)

kentcdodds:
@jayphelps, you're one of my favorites <3

jayphelps:
@kentcdodds hahaha well, I'm sure the others were gonna jump on this too. I literally already had a branch to change this that's been sitting for several months and your issue reminded me to finish. :shipit:

This really is an annoyance I've had to deal with helping others on gitter, etc.

kentcdodds:
Yeah, one thing that really helped me with beginner questions on angular-formly was validating every API with [`api-check`](https://github.com/kentcdodds/api-check). Would not recommend using this package anymore, but the concept of validating APIs for development (I was inspired by React) is awesome. I imagine TypeScript/Flow could really help with this :)

jayphelps:
@kentcdodds Yeah, TS shouldn't let you return non-observable things. We could certainly consider adding more robust error checking wrapped in React-style `process.env.node_env` checks. Would you volunteer to create the ticket to track and (just maybe) give us some pointers?

kentcdodds:
Can definitely create a ticket!

robwormald:
> We could certainly consider adding more robust error checking wrapped in React-style process.env.node_env checks 

Could we avoid introducing any code like this that's going to require a specific build process? We have a ton of consumers of Rx not using Webpack, so I'd like to avoid penalizing them.

kentcdodds:
Yes, that would be best. Optimally we should probably do what React's done by having the minified version exclude this stuff. So people not using a build system should use the minified version, and people using a build system can remove that themselves as part of the build. I wonder if there's a better way...

If I recall correctly, @jmdobry had a `lib-name.debug.js` file for his js-data library which you were supposed to use in development which had helpful things like this and then in production you'd use (and minify yourself) `lib-name.js`. Seems legit to me.

kentcdodds:
All that said, one of the drawbacks of doing an "opt-in" system for this kind of thing is the people who generally need this kind of help wont know to opt into the help and you lose one of the main benefits of doing something like this in the first place. I think that making an opt-out version which removes this stuff is better.

kentcdodds:
In any case, this is something that'll take a bit of effort, but it's well worth it IMO :) `</done>`

jayphelps:
@robwormald @kentcdodds agreed. We can continue that discussion in #2153 💃

src/util/subscribeToResult.ts
```diff
@@ -68,15 +67,18 @@ export function subscribeToResult<T>(outerSubscriber: OuterSubscriber<any, any>,
...
-      destination.error(new Error('invalid observable'));
+      destination.error(new TypeError('Provided object does not correctly implement Symbol.observable'));
     } else {
       return obs.subscribe(new InnerSubscriber(outerSubscriber, outerValue, outerIndex));
     }
   } else {
-    destination.error(new TypeError('unknown type returned'));
+    const value = isObject(result) ? 'an invalid object' : `'${result}'`;
+    const msg = `You provided ${value} where a stream was expected.`
+      + ' You can provide an Observable, Promise, Array, or Iterable.';
```
blesh: It might be nitpicking, but one could return an Rx.Observable or a "lowercase-o observable" which is any object that implements `Symbol.observable`. How do we differentiate those two things there? Maybe `Observable, Promise, Array, Iterable or object implementing Symbol.observable`?

blesh: Also, can we just use a single template string `\`like this\`` here instead of concatenating two strings together? I feel like the newline in the message is fine.

jayphelps: Yeah, I just punted on `Symbol.observable` because it felt like one of those "if you know it exists, you know it's supported" things. But I'm cool with adding `of an object implementing Symbol.observable`

jayphelps: if you use template strings, the indention will matter and be wack. I would need to do this:

```js
    const msg = `You provided ${value} where a stream was expected.
You can provide an Observable, Promise, Array, or Iterable.`;
// Notice how this can't be indented, otherwise the indention will also transfer through
// e.g. this:
    const msg = `You provided ${value} where a stream was expected.
      You can provide an Observable, Promise, Array, or Iterable.`;
// would output:
You provided null where a stream was expected.
      You can provide an Observable, Promise, Array, or Iterable.
````

If you're OK with that, I can change it. lmk.

kentcdodds: Please no. Please, leave it like it is :)

blesh: I know it can't be indented. It just seems inefficient to allocate two strings, then concatenate them inot another string.

blesh: then again, check the TypeScript compiled output, maybe if accounts for this and fixes it automagically.

blesh:
:shipit:

blesh:
LGTM

PR #2158, opened 2016-11-29
fix(race): unsubscribe raced observables with immediate scheduler

ktkiiski:
**Description:**

This fixes a bug with the `race` operator: whenever a raced Observable emitted a value **immediately** when subscribed (with the default scheduler), any following Observables were subscribed but never unsubscribed.

Now, if any raced Observable emits a value immediately, any latter Observables are ignored, because for them "the race is already lost".

Also, it is now ensured that the immediately emitting Observable is still added to the `RaceSubscriber` and will be unsubscribed once the outer Observable is unsubscribed.

### DIFF
--- a/spec/operators/race-spec.ts
+++ b/spec/operators/race-spec.ts
@@ -1,4 +1,5 @@
 import {expect} from 'chai';
+import * as sinon from 'sinon';
 import * as Rx from '../../dist/cjs/Rx';
 declare const {hot, cold, expectObservable, expectSubscriptions};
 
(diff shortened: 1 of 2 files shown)


kwonoj:
Looks good to me.

trxcllnt:
LGTM 2

kwonoj:
haven't tried it yet but seems some of the test may able to use marbles. @ktkiiski, what do you think?

ktkiiski:
I actually first wrote the specs with marbles, but they did not bust the original bug. This is probably because the marble tests are based on `VirtualTimeScheduler` (?), and it seems that the bug only occurred with the default (immediate) scheduler, when the value was immediately emitted on subscription.

kwonoj:
Ahhhh forgot that. `VirtualTimeScheduler` is kind of `AsyncScheduler`, so if test aims for default scheduler it may not work. @trxcllnt, I think PR's good as-is. any suggestions?

trxcllnt:
After reviewing the source, it seems the test cold Observables are missing the ability to synchronously emit values on subscription. I was expecting something like `cold('(x)--|')` to emit `x` upon subscription. Don't want to block this PR until that gets implemented, so LGTM.

blesh:
LGTM as well.

PR #2159, opened 2016-11-29
fix(AjaxObservable): catch XHR send failures to observer

kwonoj:
**Description:**

Currently when `XMLHttpRequest::send()` throws, AjaxObservable does not deliver those failures to observer different to other XHR behaviors. This PR updates behavior of `AjaxObservable` to change those behavior to deliver error to observer as well.


jayphelps:
LGTM

PR #2164, opened 2016-11-30
fix(typings): Use overloading instead of union type for typings filter, find, first, last.

### LINKED ISSUE #2163: Type lost when using interfaces with filter, find, first, last
**RxJS version:** RC4

**Code to reproduce:**
```
interface Bar {
  bar?: string;
}
class Foo implements Bar {
  constructor(public bar: string = 'name') {}
}

let foo: Foo = new Foo();
Observable.of(foo)
    .filter(foo => foo.bar === 'name')
    .subscribe(foo => console.log(foo.bar));  <--- works fine

let foo2: Bar = new Foo(); <--- type is interface, not the class
Observable.of(foo2)
    .filter(foo => foo.bar === 'name')
    .subscribe(foo => console.log(foo.bar)); <-- "Property 'bar' does not exist on type '{}'"
```

**Expected behavior:**

Typing to be maintained through the filter() function.

**Actual behavior:**

Type is lost downstream from .filter,() instead being typed as '{}'

**Additional information:**

Worked in RC2. Occurs for [all functions changed in RC3](https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md#500-rc3-2016-11-15)

saneyuki:
## Description

- This fixes #2163 by reverts commit 922d04e.
- So this reverts #2119 sadly.

### Drawback

- We would need specify type parameters to their methods explicitly if we'd like to narrow the type via `predicate` as type guarde function in some cases.
  - However, I don't think this is a serious problem because we can avoid this drawback with specifying actual types only.
  - And this changes makes our typings more similar to [TypeScript's Array methods](https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es5.d.ts#L1086-L1097).

## Related issue (if exists)

- #2163 
- #2119

### DIFF
--- a/spec/operators/filter-spec.ts
+++ b/spec/operators/filter-spec.ts
@@ -272,4 +272,22 @@ describe('Observable.prototype.filter', () => {
     expectObservable(r, unsub).toBe(expected);
     expectSubscriptions(source.subscriptions).toBe(subs);
   });
+
+  it('should not be compile error', () => {
+    {
+      // x is `Observable<string | number>`
+      const x: Rx.Observable<string | number> = Observable.from([1, 'aaa', 3, 'bb']);
+      // This type guard will narrow a `string | number` to a string in the examples below
+      const isString = (x: string | number): x is string => typeof x === 'string';
+
+      // Here, `s` is a string in the second filter predicate after the type guard (yay - intellisense!)
+      const guardedFilter = x.filter(isString).filter(s => s.length === 2); // Observable<string>
+      // In contrast, this type of regular boolean predicate still maintains the original type
+      const boolFilter = x.filter(s => typeof s === 'number'); // Observable<string | number>
+
+      // To avoid the lint error about unused variables 
+      expect(guardedFilter).to.not.equal(true);
+      expect(boolFilter).to.not.equal(true);
+    }
+  });
 });
--- a/spec/operators/filter-spec.ts
+++ b/spec/operators/filter-spec.ts
@@ -281,13 +281,27 @@ describe('Observable.prototype.filter', () => {
       const isString = (x: string | number): x is string => typeof x === 'string';
 
       // Here, `s` is a string in the second filter predicate after the type guard (yay - intellisense!)
-      const guardedFilter = x.filter(isString).filter(s => s.length === 2); // Observable<string>
+      const guardedFilter = x.filter<string | number, string>(isString).filter(s => s.length === 2); // Observable<string>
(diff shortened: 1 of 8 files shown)


saneyuki:
> We would need specify type parameters to their methods explicitly if we'd like to narrow the type via predicate as type guarde function in some cases. 

You can see [the diff](https://github.com/ReactiveX/rxjs/pull/2164/commits/739593f90124c051badfc313df5942abfa1fdbb2)

saneyuki:
@kwonoj @david-driscoll 

How do you think about this?

david-driscoll:
That was my thought for the solution as well.

I have one idea... I need to validate first, I'll check that right now.

david-driscoll:
Nope, that didn't work sadly.

We can solve the guard case, or we can solve the much more common case.  I would argue that it makes more sense to fix the common case.

LGTM 👍

spec/operators/filter-spec.ts
```diff
@@ -272,4 +272,36 @@ describe('Observable.prototype.filter', () => {
     expectObservable(r, unsub).toBe(expected);
     expectSubscriptions(source.subscriptions).toBe(subs);
   });
-});
+
+  it('should not be compile error', () => {
+    // tslint:disable no-unused-variable
```
blesh: It's possible that a later version of TypeScript compiler might completely remove code related to unused variables as an optimization. I'm completely speculating, but it might be worth using `expect()` to set some obvious assertions around the variables you're not using, and remove these `tslint:disable` comments.

rob3c:
Oops - sorry, guys! I only found out about this today, but I put together what I hope is a fix for this that'll let us keep the new support without breaking the previous behavior.

Please take a peek at #2170 before fully reverting it with this PR. I seem to have covered the concerns in #2163, but maybe there's something else that I missed.

saneyuki:
@blesh @jayphelps 

- I fixed what you reviewed.
- BTW, which do you accept this or #2170?
   - This pull request was opened before #2170
   - #2170 is opened by the original developer who write the patch which causes https://github.com/ReactiveX/rxjs/issues/2163

kwonoj:
@saneyuki , appreciate for effort for creating PR. as original author provided fix in #2170 and quick checking seems it correctly addresses problems, those PR has been checked in instead of taking it out. Let me close this PR in favor of #2170 . Thanks again!

saneyuki:
@kwonoj 

OK. I agree to close this ;-)

PR #2170, opened 2016-12-03
Fix type guard additions

rob3c:
**Description:**
This is a fix for the recently-added type guard support in the `filter`, `find`, `first` and `last` operator typings. Please see the test cases for details of what's working now, and be sure to let me know if I missed anything this time! ;-)

Apparently, my previous PR in #2119 is being reverted in #2164 due to the important missed case mentioned in #2163. Unfortunately, I didn't receive any notifications about it, but I'm glad I saw it today while lurking through the repo! :-) Hopefully, this PR should address the problem (which I'm embarrased that I missed), and let us keep the new support without breaking previous behavior.

### DIFF
--- a/spec/operators/filter-spec.ts
+++ b/spec/operators/filter-spec.ts
@@ -272,4 +272,55 @@ describe('Observable.prototype.filter', () => {
     expectObservable(r, unsub).toBe(expected);
     expectSubscriptions(source.subscriptions).toBe(subs);
   });
+
+  it('should support type guards without breaking previous behavior', () => {
+    // tslint:disable no-unused-variable
+
+    // type guards with interfaces and classes
+    {
+      interface Bar { bar?: string; }
+      interface Baz { baz?: number; }
+      class Foo implements Bar, Baz { constructor(public bar: string = 'name', public baz: number = 42) {} }
+
+      const isBar = (x: any): x is Bar => x && (<Bar>x).bar !== undefined;
+      const isBaz = (x: any): x is Baz => x && (<Baz>x).baz !== undefined;
+
+      const foo: Foo = new Foo();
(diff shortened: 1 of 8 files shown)


jayphelps:
Cc/ @david-driscoll can you review?

david-driscoll:
I swear I had tried overload ordering the other day to validate if this was possible.

I tried again just now and these changes seem to be working just fine.  👍 

cc @saneyuki for an extra set of 👀

rob3c:
@david-driscoll Yep, the ordering is definitely necessary! It wasn't enough, though, so there are some strategic non-optional predicates in there, too, to help typescript resolve certain cases.

saneyuki:
@rob3c @david-driscoll 

I confirmed that in https://github.com/ReactiveX/rxjs/pull/2164/ we need to write the version which takes `predicate` returning `boolean` before to define the one which takes the type guard `predicate` so that the type inference works correctly for the case of that `predicate` returning `boolean` simply. (So we need to define the order of `bar<T>()` => `bar<T, S extends T>()`)

Is it resolved in this?  The current patches are not care about them.

rob3c:
@saneyuki That's what I originally thought, too.

Surprisingly, it turns out that `<T, S extends T>` (and `<T, S extends T, R>` for `first` and `last`) needs to be declared before the simpler `<T>` (and `<T, R>`) definitions, but it needs to be combined with carefully-defined required/optional parameters, so typescript can distinguish all desired cases. Otherwise, we get the dreaded `{}` type inferred instead of the actual types for some of them.

kwonoj:
LGTM

blesh:
Thanks so much, @rob3c!!

PR #2132, opened 2016-11-15
fix(timeout): throw traceable TimeoutError

kwonoj:
**Description:**
This PR updates behaivor of `timeout` operator for better readibility of stack trace when `TimeoutError` thrown. 

For example, with given code snippet

```js
Rx.Observable.never().timeout(10).subscribe(console.log.bind(console));
```

stacktrace will be like below
```js
D:\github\rxjs\dist\cjs\scheduler\AsyncScheduler.js:45
            throw error;
            ^
TimeoutError: Timeout has occurred
    at new TimeoutError (D:\github\rxjs\dist\cjs\util\TimeoutError.js:17:26)
    at TimeoutSubscriber.notifyTimeout (D:\github\rxjs\dist\cjs\operator\timeout.js:99:40)
    at AsyncAction.TimeoutSubscriber.dispatchTimeout [as work] (D:\github\rxjs\dist\cjs\operator\timeout.js:75:20)
    at AsyncAction._execute (D:\github\rxjs\dist\cjs\scheduler\AsyncAction.js:111:18)
    at AsyncAction.execute (D:\github\rxjs\dist\cjs\scheduler\AsyncAction.js:86:26)
    at AsyncScheduler.flush (D:\github\rxjs\dist\cjs\scheduler\AsyncScheduler.js:36:32)
    at ontimeout (timers.js:365:14)
    at tryOnTimeout (timers.js:237:5)
    at Timer.listOnTimeout (timers.js:207:5)
```

which loses original context of code but shows stack trace of async execution context (vary by scheduler specified). This makes hard to track down unhandled exception since it's disconnected to original codebases. Instead, this PR creates error to contain stack traces nearby originall caller, emits like below

```js
D:\github\rxjs\dist\cjs\scheduler\AsyncScheduler.js:45
            throw error;
            ^
TimeoutError: Timeout has occurred
    at new TimeoutError (D:\github\rxjs\dist\cjs\util\TimeoutError.js:17:26)
    at NeverObservable.timeout (D:\github\rxjs\dist\cjs\operator\timeout.js:24:32)
    at Object.<anonymous> (D:\github\rxjs\dummy.js:3:23)
    at Module._compile (module.js:573:32)
    at Object.Module._extensions..js (module.js:582:10)
    at Module.load (module.js:490:32)
    at tryModuleLoad (module.js:449:12)
    at Function.Module._load (module.js:441:3)
    at Module.runMain (module.js:607:10)
    at run (bootstrap_node.js:382:7)
```
[...]

PR #2172, opened 2016-12-05
feat(timeout): remove `errorToSend` argument, always throw TimeoutError

blesh:
The `errorToSend` argument did not make much sense, so we are removing it ahead of the 5.0 release. All timeout errors will be thrown as `TimeoutError`, which can be tested with an assertion against `instanceof Rx.TimeoutError`.

BREAKING CHANGE: `timeout` no longer accepts the `errorToSend` argument

related #2141

### DIFF
--- a/src/operator/timeout.ts
+++ b/src/operator/timeout.ts
@@ -8,32 +8,30 @@ import { TeardownLogic } from '../Subscription';
 import { TimeoutError } from '../util/TimeoutError';
 
 /**
- * @param due
- * @param errorToSend
- * @param scheduler
+ * @param {number} due
+ * @param {Scheduler} [scheduler]
  * @return {Observable<R>|WebSocketSubject<T>|Observable<T>}
  * @method timeout
  * @owner Observable
  */
-export function timeout<T>(this: Observable<T>, due: number | Date,
-                           errorToSend: any = null,
+export function timeout<T>(this: Observable<T>,
+                           due: number | Date,
                            scheduler: Scheduler = async): Observable<T> {
   const absoluteTimeout = isDate(due);
   const waitFor = absoluteTimeout ? (+due - scheduler.now()) : Math.abs(<number>due);
-  const error = errorToSend || new TimeoutError();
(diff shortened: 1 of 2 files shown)


blesh:
cc @kwonoj @jayphelps

src/operator/timeout.ts
```diff
@@ -97,6 +93,6 @@ class TimeoutSubscriber<T> extends Subscriber<T> {
   }
 
   notifyTimeout(): void {
-    this.error(this.errorToSend);
+    this.error(new TimeoutError());
```
kwonoj: I'd propose to create error object in ctor, allows bit easier stack to trace (was done in https://github.com/ReactiveX/rxjs/pull/2132)

blesh: Why allocate an object you don't need?

kwonoj: Not sure which object is unused, pr #2132 is simply changing where error object is created and pass into here. Would you elaborate bit what I'm missing?

jayphelps: @blesh because if you do end up needing it, the stack trace is awful if we don't do it at the constructor call site. See #2132

blesh: I see. Makes sense.

blesh:
The coverage results seem to be just the unfortunate math of removing a few tests we didn't need. I'm not worried about it.

kwonoj:
LGTM

samal84:
Note to anyone else who was using the errorToSend argument to create a contextual error message.
You can use catch immediately after the timeout and check the error for instanceof TimeoutError like this to achive the same:
```
source.timeout(1000)
.catch((error) => {
  if (error instanceof TimeoutError) {
    throw new Error('Timeout error. Value of someLocalVariable was: ' + someLocalVariable);
  } else {
    throw error;
  }
});
```
Quite a lot more lines I know.

samal84:
Also @benlesh why did the errorToSend argument not make any sense?

Issue #1754, opened 2016-06-08
Bug in buffer or bufferCount operation when used with subject


iFadey:
**RxJS version:** 5.0.0-beta.8

**Code to reproduce:**
https://jsfiddle.net/iFadey/qahvL1dm/

``` javascript
let section = document.querySelector('section');
let subject = new Rx.Subject();

Rx.Observable.fromEvent(section, 'click')
.subscribe(e => subject.next(e))
;

subject.buffer(subject.bufferCount(2))
.subscribe(x => console.log(x))
;
```

**Expected behavior:**
On first two clicks on section element must print pair of two event objects.

**Actual behavior:**
It prints single event object on first two clicks. Afterwards it starts printing correct pair.

**Additional information:**
I verified it in RxJS 4 and it works correctly with that version.

blesh:
To narrow this down, what you're seeing is happening in `buffer`. The same behavior occurs with `subject.buffer(subject.take(2))`. This seems a little edge-casey, and might be a quirk with changes to default scheduling, but, for now I'll label it as a bug until we can confirm otherwise, since it is odd.

Podlas29:
Currently on 
```
const source = Observable.from([1,2,3,4,5,6]);
source.buffer(source.take(2));
```
RxJS 4 returns:
```
[1]
[2]
[]
// complete
```
This last array is last opened buffer, which did not have time to be populated with values, since right after it's opening `take(2)` completes stream.

The same code in RxJS 5 results in:
```
[]
[]
// complete
```

There are actually two differences here: first is obvious subscription order bug, that was fixes here: https://github.com/ReactiveX/rxjs/pull/2195

On that branch results are: 
```
[1]
[2]
// complete
```
So second problem is that last one array. Currently code
```
Observable.never().buffer(Observable.empty());
```
in RxJS 4 results in
```
[]
// complete
```
and in RxJS 5
```
// complete
```
So real question here is: **when selector stream completes do we want to return last initiated buffer?**

PR #2174, opened 2016-12-07
fix(buffer): emit last buffer if source completes

JaminFarr:
**Description:**
This fixes an issue with the bufffer operator.
If the source emits values and completes before the notifier emits, those values are lost. 

```
source  = --a--b--c--d--e--|
notifer = ----x-----x-------
source.buffer(notifer)

results
[a]
[b, c]

d & e are never emitted
```

**Related issue:** https://github.com/ReactiveX/rxjs/issues/2154

Fix buffer operator not emitting the last buffer when the source
completes. This is closer to rxjs v4* behaviour and matches the
behaviour of the other buffer operators. The _complete method is very
similar to thoses in bufferCount, bufferTime and bufferWhen.

*rxjs v4 will always emit the buffer if the source completes even if the
buffer is empty. This fix only emits if the buffer is non empty.

BREAKING CHANGE:

extra value may be emitted when source completes

### DIFF
--- a/spec/operators/buffer-spec.ts
+++ b/spec/operators/buffer-spec.ts
@@ -219,4 +219,16 @@ describe('Observable.prototype.buffer', () => {
     expectObservable(a.buffer(b).take(1)).toBe(expected, expectedValues);
     expectSubscriptions(b.subscriptions).toBe(bsubs);
   });
+
+  it('should emit last buffer if source completes', () => {
+    const a =    hot('-a-b-c-d-e-f-g-h-i-|');
+    const b =    hot('-----B-----B--------');
+    const expected = '-----x-----y-------(z|)';
+    const expectedValues = {
+      x: ['a', 'b', 'c'],
+      y: ['d', 'e', 'f'],
+      z: ['g', 'h', 'i']
+    };
+    expectObservable(a.buffer(b)).toBe(expected, expectedValues);
+  });
 });
--- a/src/operator/buffer.ts
+++ b/src/operator/buffer.ts
@@ -69,6 +69,14 @@ class BufferSubscriber<T> extends OuterSubscriber<T, any> {
     this.buffer.push(value);
   }
 
+  protected _complete() {
+    const buffer = this.buffer;
+    if (buffer.length > 0) {
+      this.destination.next(buffer);
+    }
+    super._complete();
+  }
+
   notifyNext(outerValue: T, innerValue: any,
              outerIndex: number, innerIndex: number,
              innerSub: InnerSubscriber<T, any>): void {


Podlas29:
@JaminFarr does in this branch last buffer is emited also when `selector` observable completes? (for context see: https://github.com/ReactiveX/rxjs/issues/1754)

blesh:
We'll need to verify that this is how Rx4 behaved. As it stands right now this would be more of a breaking change than a "fix", I think. But if it behaved this way in Rx4, there's an argument to be made that this is a fix.

blesh:
@JaminFarr confirmed that Rx4 behaves this way and Rx5 does not... therefor this is a bug fix. :+1:

blesh:
This LGTM... Will merge after conferring with others. Pretty sure this should land in 5.0.2 as a bug fix though.

JaminFarr:
@Podlas29 No, This PR only changes the source complete behavior.

Examples from https://github.com/ReactiveX/rxjs/issues/1754
```javascript
Observable.never().buffer(Observable.empty());

// Result: Same, emits nothing. Source never completes so it is not affected by this fix. 
```
```javascript
const source = Observable.from([1,2,3,4,5,6]);
source.buffer(source.take(2));

// Result:
[] // Same
[] // Same
[1, 2, 3, 4, 5, 6] // new from this change
// The notifier emits twice and completes before source is subscribed to, producing
// the first two empty array results. Then the source completes and the array of the
// remaining values is emitted.
```

@blesh Great.

jayphelps:
Merged into `next` branch, as we decided in our [core team meeting](https://github.com/jayphelps/rxjs-core-notes/blob/d03095b98c2b4c26e4e340f85e50294823b859ff/2017-01/january-30.md#fixbuffer-emit-last-buffer-if-source-completes-2174) today this was indeed a breaking change that has to go into a major version bump.

jayphelps:
Thanks @JaminFarr !!

Jason3S:
@JaminFarr What is the behavior of this fix? Does it emit immediately or wait until the `closingNotifier`.  Please see #2361

JaminFarr:
@Jason3S This fix emits immediately when the source completes with the remaining values. This matches RxJS4 but looks like the opposite behavior to what you're after. I don't think there is a version of BufferWhen in RxJS4 to compare against.

As a workaround to delay the last buffer value to emit in time with the trigger, you can zip the result with the trigger.

```javascript
const Observable = Rx.Observable;

const source$ = Observable.interval(200).take(25)
const trigger$ = Observable.interval(2000).share()




// I'm using window and mergeMap as a workaround until my PR is merged
// This is the same as buffer(trigger$)
source$
  .window(trigger$)
  .mergeMap(x => x.toArray())

// Use zip to delay the last value and a selector to remove the trigger value
  .zip(trigger$, (sourceValue /*, triggerValue*/) => sourceValue)

2s: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
4s: [10, 11, 12, 13, 14, 15, 16, 17, 18, 19]
6s: [20, 21, 22, 23] -- without the zip this will emit at ~4.5s
```

Jason3S:
Nice workaround.

Issue #2178, opened 2016-12-08
Custom error object is no longer typeof Error from TypeScript 2.1


kwonoj:
**RxJS version:**

**Code to reproduce:**

**Expected behavior:**

**Actual behavior:**

**Additional information:**
Was shared in meeting, it is actually intended design of TS 2.1 (https://github.com/Microsoft/TypeScript-wiki/blob/master/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work) that no longer allows to extend built in object by default. 

Means, all of custom errors in RxJS v5 (i.e `TimeoutError`) is not typeof `Error`, will break existing code if anyone does type checking for those, as well as existing test cases too.

kwonoj:
Assigning myself for now.

kwonoj:
Seems 2.2 have enhancement support for this : https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#support-for-newtarget

saneyuki:
@kwonoj 

> Seems 2.2 have enhancement support for this

IE9 & IE10 does not have `__proto__`. I feel it would not work well for them.

kwonoj:
I'm closing this issue since we don't have exact plans to bump up compiler, can revisit later once things are getting clear.

Issue #2066, opened 2016-10-25
feat(TypeScript): enable 'strictNullChecks' option


saneyuki:
## Pros
- This would make our code base more solidly, more clearly to decrease a bug. 
- This option enables full control-flow basis analysis of tsc.
  - This option is introduced from tsc 2.0.
## Cons
- We may need change a public signatures as breaking change.

cc: @david-driscoll @kwonoj

jayphelps:
I'm absolutely for this in theory--huge advocate for non-nullable types. I'm mostly worried about consumers of RxJS who _don't_ have `strictNullChecks` enabled. Presumably they will still be able to pass `null` against our TS definitions right? i.e. this is mostly for our internal sake?

Off hand I can't think of any signatures we would _want_ to accept `null` as valid. We sometimes call having no scheduler the "null scheduler" but this is really a misnomer IMO and I'd like to stop doing that. Can anyone think of any?

kwonoj:
> Presumably they will still be able to pass null against our TS definitions right? i.e. this is mostly for our internal sake?

: I expect same behavior. this option's affecting compile time behavior, I don't think it can control consumer's compile time options.

> Off hand I can't think of any signatures we would want to accept null as valid. 

: Maybe someone might use `null` where `undefined` as valid value as well?

saneyuki:
@jayphelps 

> I'm absolutely for this in theory--huge advocate for non-nullable types. I'm mostly worried about consumers of RxJS who don't have strictNullChecks enabled. Presumably they will still be able to pass null against our TS definitions right? i.e. this is mostly for our internal sake?

- Yes. I think this would be a internal sake basically if we don't change public interface signatures.
- After we enables `strictNullChecks`, it would be a problem if the consumer of RxJS compiles their codes & rxjs's raw TypeScript code without `strictNullChecks`.
  - But I think this is very very rare case. Almost (maybe ~99.9%) consumers uses only rxjs via npm with bundled d.ts.

saneyuki:
In the following case, 

- It would not be a compile error if the consumer _disable_ `strictNullChecks`.
- It would be a compile error if the consumer _enable_ `strictNullChecks`.

Thus I think that there would not be a problem if we don't change public interface.

```ts
// bar.d.ts
export function bar(a: number): void;
```

```ts
// consumer's code
import {bar} from 'bar';

bar(1);
bar(null); // compile error with `strictNullChecks`
bar(undefined); // compile error with `strictNullChecks`
[...]

PR #2181, opened 2016-12-08
Patch custom Error object to inherit base Error

kwonoj:
**Description:**
Per breaking changes of TS 2.1 (https://github.com/Microsoft/TypeScript-wiki/blob/master/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work) Inheriting built in object no longer gives prototype chain. This PR updates each custom error object to have prototype of Error.

#2178 

*BREAKING CHANGE*: This change drops support of IE 10 and prior

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,9 @@ install:
   - npm install && npm run lint
 
 script:
-  - npm run build_spec && npm run test_mocha && node ./node_modules/markdown-doctest/bin/cmd.js
+  - npm install typescript@2.0.7 && npm run build_spec && npm run test_mocha
+  - npm install typescript@latest && npm run build_spec && npm run test_mocha
+  - node ./node_modules/markdown-doctest/bin/cmd.js
   - npm run check_circular_dependencies
 
 after_success:
--- a/package.json
+++ b/package.json
@@ -186,7 +186,7 @@
     "source-map-support": "^0.4.0",
     "tslib": "^1.0.0",
     "tslint": "^3.15.1",
-    "typescript": "^2.0.6",
+    "typescript": "^2.1.4",
     "typings": "^2.0.0",
     "validate-commit-msg": "^2.3.1",
     "watch": "^1.0.1",
--- a/spec/support/mocha.sauce.runner.js
+++ b/spec/support/mocha.sauce.runner.js
@@ -63,18 +63,6 @@ var customLaunchers = {
     platform: 'OS X 10.11',
     version: '9.1'
   },
-  sl_ie9: {
-    base: 'SauceLabs',
-    browserName: 'internet explorer',
-    platform: 'Windows 2008',
-    version: '9'
-  },
-  sl_ie10: {
-    base: 'SauceLabs',
-    browserName: 'internet explorer',
-    platform: 'Windows 2012',
-    version: '10'
-  },
   sl_ie11: {
     base: 'SauceLabs',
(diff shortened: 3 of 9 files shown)


.travis.yml
```diff
@@ -32,7 +32,9 @@ install:
   - npm install && npm run lint
 
 script:
-  - npm run build_spec && npm run test_mocha && node ./node_modules/markdown-doctest/bin/cmd.js
+  - npm install typescript@2.0.7 && npm run build_spec && npm run test_mocha
+  - npm install typescript@latest && npm run build_spec && npm run test_mocha
```
kwonoj: this is kind of hacky, monkey patching to test on TS@2.0.x and latest both. Maybe better to setup different CI configuration to test each compilers instead in further. 

Note : obviously this will slow down CI build.

saneyuki: @kwonoj 

Should we keep ts 2.0.x? We cannot continue to support IE 10 or prior even if we tried keep ts 2.0.x compatibility.

kwonoj: This is not for browser compatibility, not to introduce current change breaks on TS@2.0.x.

saneyuki: Do we have any reason to keep the compatibility with ts2.0? We can change the dependency to it at now before we'll ship to stable release.

And, we can release changes including a breaking change by little by little with semver. That's is just the benefit of semver to avoid any giant releases.

kwonoj: If it's certain new feature like module argumentation, I'd agree that. But for this case, I see no strong reason to not to keep compatibility between 2 versions, especially given custom object is expected to inherit Error Object itself. This test is guarding 'current' change to be safe with previous version of TS, while further version comes out and if decided to introduce version specific breaking changes can be taken out.

saneyuki: > This test is guarding 'current' change to be safe with previous version of TS, while further version comes out and if decided to introduce version specific breaking changes can be taken out.

I'm worrying about that we will need to maintain on 2 different version of compilers if we try to https://github.com/ReactiveX/rxjs/issues/2066, these type based analysis are improved by tsc incrementally, so the difference of compiler causes the situation we cannot compile it on both versions.
So do you means that is is an separate issue which we will face to and we should resolve if we tackle to https://github.com/ReactiveX/rxjs/issues/2066 or others? (If you said so, I agree it.)

kwonoj: as #2066 does not have near-term possibility to check in (and even for this), I am not worried too much since per each case this can be easily taken out as necessary.

david-driscoll:
I fell this change is a little dubious... dropping support for IE9 and IE10 where support is useful for Angular2 seems like it might be a conflict.

Wouldn't it make more sense to fall back and copy the properties in the event __proto__ doesn't exist?

kwonoj:
> Wouldn't it make more sense to fall back and copy the properties in the event proto doesn't exist?

Haven't try it - does it makes to inherit Error correctly?

blesh:
At Netflix, we have to assume the worst browsers in some cases... So I can't condone dropping support for < IE 10.

jayphelps:
I personally feel it's too early to drop IE9/10 as well. [ng2 supports them](https://angular.io/docs/ts/latest/guide/browser-support.html), so if we drop them, it puts us both in a very difficult position that will likely have backlash.

kwonoj:
As IE support is critical part in library, I believe this PR is not acceptable as-is. Closing it.

PR #2182, opened 2016-12-08
chore(package.json): upgrade typescript to v2.1

saneyuki:
BREAKING CHANGE: TypeScript definitions are now for TS 2.1 and higher.

Abstract
----------

- This change does not cause any breaking change directly, but we would causes some breaking change by using new features intoduced by v2.1 for the future.
- If we upgrade typescript to v2.1, there are some pros/cons.

### pros

- we can switch to npm:@types type dependencies by https://github.com/Microsoft/TypeScript/issues/11849 is fixed in v2.1.
- we can use more powerfull features introduced by v2.1.
  - We can use [tslib](https://github.com/Microsoft/tslib) to reduce our compiled code size.
- we don't have to upgrade major version v5 stable release if we merge this now before v5 stable release.

### cons

- This would introduce some breaking change for the future.
  - But we might cause some breaking change intentionally because we don't lock the range of our typescript version to patch version. This might causes to accept new features which uses the lang features introduced by a latest version of TypeScript

## Relations links

- [TypeScript 2.1 RC: Better Inference, Async Functions, and More](https://blogs.msdn.microsoft.com/typescript/2016/11/08/typescript-2-1-rc-better-inference-async-functions-and-more/)
- [Announcing TypeScript 2.1](https://blogs.msdn.microsoft.com/typescript/2016/12/07/announcing-typescript-2-a1/)

## Related issues

- https://github.com/ReactiveX/rxjs/issues/2178
  - We should wait it if we merge this into master.

### DIFF
--- a/package.json
+++ b/package.json
@@ -186,7 +186,7 @@
     "source-map-support": "^0.4.0",
     "tslib": "^1.0.0",
     "tslint": "^3.15.1",
-    "typescript": "^2.0.6",
+    "typescript": "^2.1.4",
     "typings": "^2.0.0",
     "validate-commit-msg": "^2.3.1",
     "watch": "^1.0.1",


kwonoj:
I'm closing this PR by https://github.com/ReactiveX/rxjs/pull/2181 as it's superset includes fixes for breaking changes.

saneyuki:
@kwonoj ah, okay.

PR #2183, opened 2016-12-08
Make subscribeOn use lift, refactor operators to call public subscribe method

trxcllnt:
Today, operators are passed the Observable they're attached to as the second "source" argument. They have to call a private `_subscribe` method, which is a pass-through inside Observable to `this.source.subscribe`. This is less ergonomic (someone writing a custom operator has to know to do this), and an extra function call per Observable in the chain. The solution to this is to pass `this.source` to the operators, who can then call the source's public `subscribe` method.

This PR also makes subscribeOn use lift, and fixes a bug where the subscribeOn subscriber wasn't tracking the source subscription.

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -93,7 +93,7 @@ export class Observable<T> implements Subscribable<T> {
     const sink = toSubscriber(observerOrNext, error, complete);
 
     if (operator) {
-      operator.call(sink, this);
+      operator.call(sink, this.source);
     } else {
       sink.add(this._subscribe(sink));
     }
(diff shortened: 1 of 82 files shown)


trxcllnt:
Ah also, I guess this is a breaking change, but it only affects weirdos like me who override lift and write custom operators.

staltz:
LGTM

blesh:
LGTM as well. The coverage fell as a result of basic math. I'll merge it

trxcllnt:
WOOOOOOOOO

Issue #2171, opened 2016-12-04
Examples do not have resulting output


stepanzarubin:
Documentation examples does not have resulting output, which is not obvious and makes study process difficult.

```
var weight = Rx.Observable.of(70, 72, 76, 79, 75);
var height = Rx.Observable.of(1.76, 1.77, 1.78);
var bmi = Rx.Observable.combineLatest(weight, height, (w, h) => w / (h * h));
bmi.subscribe(x => console.log('BMI is ' + x));
```

How can I know the output without running it myself? Am I missing something?

jayphelps:
I think that's a definitely fair critique. Thanks for letting us know!

As a side note, the marble diagrams do show you the results of using an operator, but it's not the results from the example section 🙁  http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-combineLatest

![image](https://cloud.githubusercontent.com/assets/762949/20895779/a2675470-bacf-11e6-8fbf-c74b1180893f.png)

We're also will missing docs for many of the operators. We know this, just lack the resources/contributors at the moment.

jrroman:
@jayphelps are these answers something that you would like added to the README.md? If so I would be more than happy to help out.

jayphelps:
@jrroman we don't wanna add operator specific docs to the README itself, but certainly if you have cycles to amend the inline ESDoc documentation and showing what the example would output, please do!

jrroman:
@jayphelps that sounds good to me!

stepanzarubin:
@jayphelps I can contribute on my free time by adding examples output whenever possible, just give me a single example so that we have a standard formatting.

stepanzarubin:
@blesh this does not affect public docs,  is this what community wants to avoid here?[http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-combineLatest](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-combineLatest)

jrroman:
@stepanzarubin What do you mean by, this does not affect public docs?

blesh:
@stepanzarubin I think it's that we didn't update the docs yet.

benlesh:
Closing as stale

PR #2184, opened 2016-12-08
docs(DeferObservable.ts, EmptyObservable.ts, FromEventObservable.ts, …

jrroman:
**Description:**

…FromObservable.ts, combineLates

Add output results for the examples in the documentation pertaining to the operators. Change

seqenceEqual.ts 'var code = Observable.from' to 'var code = Rx.Observable.from'. If the current

example is pasted into an editor and executed it will throw an error of 'Observable not defined'.

The addition of Rx. fixes this.

Issue: Examples do not have resulting output #2171

### DIFF
--- a/src/observable/DeferObservable.ts
+++ b/src/observable/DeferObservable.ts
@@ -38,6 +38,12 @@ export class DeferObservable<T> extends Observable<T> {
    *   }
    * });
    * clicksOrInterval.subscribe(x => console.log(x));
+   * 
+   * // Results in the following behavior:
+   * // If the result of Math.random() is greater than 0.5 it will listen
+   * // for clicks anywhere on the "document"; when document is clicked it
+   * // will log a MouseEvent object to the console. If the result is less
+   * // than 0.5 it will emit ascending numbers, one every second(1000ms).
    *
    * @see {@link create}
    *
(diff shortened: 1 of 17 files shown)


staltz:
Hi @jrroman. Have you tried generating the docs site (`npm run build_docs` if I'm not mistaken) after your changes? I'm afraid those `//` comments won't show up in the code block, because the documentation generator looks for adjacent lines. So to be safe the `//` comments should be after code, with no empty line in between.

jrroman:
Hi @staltz, I did run `npm run build_docs` and it looked like they appear in the code block with the associated example. I have attached a screenshot of my local, maybe I misunderstood something. If you would like I can move them directly adjacent so that there is no chance of error.

screen shot 2016-12-08 at 16 51 49

jrroman:
@staltz I have gone through and removed all spaces between the code and the comments. It now looks like the below screenshot. If this is what you mean, I can push these changes.

screen shot 2016-12-08 at 17 14 03

blesh:
@jrroman I personally like it better with a space in between. But that's me...

staltz:
Ok, cool. I was suggesting no lines in between just to make sure that they compiled as a code block. If it works even with an empty line in between, then sure, go ahead.

jrroman:
@blesh @staltz I cleaned the examples output up a bit just slight corrections with spacing and such. I kept the space between the example and the output per blesh's request.

jrroman:
screenshot of the output from the last commit for reference.

screen shot 2016-12-09 at 16 48 24

staltz:
LGTM

blesh:
LGTM

PR #2191, opened 2016-12-13
fix(TypeScript): pin to TypeScript 2.0.x, fix errors with Error subcl…

blesh:
…assing

Because the the last build was pinned to 2.x.x, it built with TypeScript 2.1.x, which had breaking changes but was not a major release.

This should fix that.

:(


staltz:
Same problem happened to me in some of my packages. If/when RxJS moves to TS 2.1, it will be a breaking change. So please don't do the same mistake as TS team, follow semver and bump it to v6.0.0.

PR #2195, opened 2016-12-14
fix(buffer): subscribe to source and closingNotifier in proper order

### LINKED ISSUE #1610: Different behavior with Rx 4 and buffer
**RxJS version:** 5.0.0-beta5

**Code to reproduce:**

``` js
const a = Observable.of(1, 2, 3, 4).share();
a
  .buffer(a.filter(x => x % 2 === 0))
```

**Expected behavior:**

```
[1, 2]
[3, 4]
[]
```

**Actual behavior:**

```
[]
[]
```

**Additional information:**

The "expected behavior" is what Rx 4 does, but maybe there's a more correct way to do this?

mpodlasin:
**Description:**

In buffer operator subscribe to source observable first, so that when
closingNotifier emits value, all source values emited before land in buffer

Closes #1610


blesh:
Is this the behavior of RxJS 4? Worth confirming.

blesh:
It looks like Rx5 and Rx4 currently differ in behavior, even given the same scheduling. So this is something we need to look at and fix.

Podlas29:
In RxJS 4
```javascript
const Observable = Rx.Observable;

const xs = Observable.from([0,1,2,3,4,5,6]);

xs.buffer(xs.filter(x => x % 2 === 0))
```
results in 
`[0]`
`[1, 2]`
`[3, 4]`

After my change rxjs 5 behaves same (as seen in my unit test).
Do you have any cases where behaviour is different?

Podlas29:
@blesh bare in mind that this pr will also partially help solve: https://github.com/ReactiveX/rxjs/issues/1754

jayphelps:
Needs a BREAKING CHANGE disclaimer in the commit message body so that it will show up in our automatic CHANGELOG.md generation

Podlas29:
@jayphelps done. please let me know if message is ok. ;)

jayphelps:
I changed to base branch aka merge target to `next` since this is a breaking change.

Still need one more person to review

PR #2196, opened 2016-12-15
fix(Observable): align type definition of subscriber with Observable

### LINKED ISSUE #2166: Why different generics for Subscriber and Observable
Not sure if this is a bug or not.

**RxJS version:**

`5.0.0-rc.4`

**Code to reproduce:**

https://github.com/ReactiveX/rxjs/blob/e91d113e49acc9b0e12e2524040fa84cf3fa2c78/src/Observable.ts#L56

**Expected behavior:**

```ts
  static create: Function = <T>(subscribe?: <T>(subscriber: Subscriber<T>) => TeardownLogic) => {
    return new Observable<T>(subscribe);
  };
```

**Actual behavior:**

```ts
  static create: Function = <T>(subscribe?: <R>(subscriber: Subscriber<R>) => TeardownLogic) => {
    return new Observable<T>(subscribe);
  };
```

**Additional information:**

Maybe I misunderstood this part of RxJS, but I thought the Observable and Subscriber would share the same generic type here? Aren't the values I pass to `subscriber.next` from the same type I get when I use `value$.map`?

kwonoj:
**Description:**
This PR align type of subscriber param to `<T>`, same as observable.

- closes #2166

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -37,7 +37,7 @@ export class Observable<T> implements Subscribable<T> {
    * can be `next`ed, or an `error` method can be called to raise an error, or
    * `complete` can be called to notify of a successful completion.
    */
-  constructor(subscribe?: <R>(this: Observable<T>, subscriber: Subscriber<R>) => TeardownLogic) {
+  constructor(subscribe?: (this: Observable<T>, subscriber: Subscriber<T>) => TeardownLogic) {
     if (subscribe) {
       this._subscribe = subscribe;
     }
@@ -53,7 +53,7 @@ export class Observable<T> implements Subscribable<T> {
    * @param {Function} subscribe? the subscriber function to be passed to the Observable constructor
    * @return {Observable} a new cold observable
    */
-  static create: Function = <T>(subscribe?: <R>(subscriber: Subscriber<R>) => TeardownLogic) => {
+  static create: Function = <T>(subscribe?: (subscriber: Subscriber<T>) => TeardownLogic) => {
     return new Observable<T>(subscribe);
   };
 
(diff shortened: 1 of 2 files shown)


kwonoj:
What would be concise example to test out if this causes not regression? sometimes there are regression even build passes correctly (like recent changes to `filter`, etcs)

kwonoj:
wow actually build caught this. :cry:

kwonoj:
I'm marking it as blocked for review approval / and sound test plan.

blesh:
The change looks good... we'll just need to ensure it's non-breaking so we can figure out where it lands.

kwonoj:
Agreed. Need to figure out some way to test.. /cc @david-driscoll for borrowing some idea.

kwonoj:
Enabling #2202 into all current test cases will unblock this PR for most common usecase.

src/Observable.ts
```diff
@@ -37,7 +37,7 @@ export class Observable<T> implements Subscribable<T> {
    * can be `next`ed, or an `error` method can be called to raise an error, or
    * `complete` can be called to notify of a successful completion.
    */
-  constructor(subscribe?: <R>(this: Observable<T>, subscriber: Subscriber<R>) => TeardownLogic) {
+  constructor(subscribe?: (this: Observable<T>, subscriber: Subscriber<T>) => TeardownLogic) {
```
david-driscoll: I've suspected this could be troublesome at some point, but I'm sure I forgot about just after I saw it.

blesh:
Okay... so this looks good, but it reminds me that we're actually leaking an implementation detail via TypeScript. We don't want to encourage our external users to use anything other than the Observer<T> interface. Subscriber is more of an implementation thing for our library.

PR #2199, opened 2016-12-19
chore(readme): Simplifes CDN usage directions

ichpuchtli:
**Description:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,10 +122,8 @@ npm install @reactivex/rxjs@5.0.0
 ### CDN
 
 For CDN, you can use [unpkg](https://unpkg.com/):
-
-##### 140k Minified, 30k Minified && Gzip'd  
   
-https://unpkg.com/@reactivex/rxjs/dist/global/Rx.min.js
+https://unpkg.com/rxjs/bundles/Rx.min.js
 
 #### Node.js Usage:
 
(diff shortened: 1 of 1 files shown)


kwonoj:
Looks ok, but maybe (still) need to expose non-minified package as well? (https://unpkg.com/@reactivex/rxjs@5.0.1/dist/global/Rx.js)

ichpuchtli:
Sure. I actually had that with a label saying Good for Debugging, so I will add that back.

blesh:
Honestly, I've been using `https://unpkg.com/rxjs/bundles/Rx.min.js`

ichpuchtli:
@blesh is it possible to selectively pull in operators using the cdn? i.e. 
```html
<script src="https://unpkg.com/rxjs/observable/observable.min.js" />
<script src="https://unpkg.com/rxjs/observable/of.min.js" />
<script src="https://unpkg.com/rxjs/operators/map.min.js" />
```
would be equivalent to
```ts
import { of } from 'rxjs/observable/of';
```

jayphelps:
@ichpuchtli it's not. The maintenance overhead to our build pipeline would be pretty onerous and it's already a nightmare. We make the assumption that people who care about file size should bundle--after all, patching with individual links would be pretty inefficient cause of number of HTTP requests

jayphelps:
Updated PR to use `https://unpkg.com/rxjs/bundles/Rx.min.js` per @blesh.

We may go back and forth on this so forgive us if we do. It's still not clear when someone should use `@reactivex/rxjs` vs `rxjs` and why we need two different packages.

PR #2200, opened 2016-12-19
Fix ajax upload progress event is not set correctly

Brooooooklyn:
**Description:**
progress event should be set before `xhr.open`,
see: https://developer.mozilla.org/en/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest

image

### DIFF
--- a/spec/helpers/ajax-helper.ts
+++ b/spec/helpers/ajax-helper.ts
@@ -200,7 +200,12 @@ export class MockXMLHttpRequest {
     throw new Error('unhandled type "' + this.responseType + '"');
   }
 
-  respondWith(response: any): void {
+  respondWith(response: any, progressTimes?: number): void {
+    if (progressTimes) {
+      for (let i = 1; i <= progressTimes; ++ i) {
+        this.triggerUploadEvent('progress', { type: 'ProgressEvent', total: progressTimes, loaded: i });
+      }
+    }
     this.readyState = 4;
     this.responseHeaders = {
       'Content-Type': response.contentType || 'text/plain'
@@ -238,6 +243,15 @@ export class MockXMLHttpRequest {
       }
     });
   }
+
+  triggerUploadEvent(name: any, eventObj?: any): void {
+    // TODO: create a better default event
+    const e: any = eventObj || {};
+
+    if (this.upload['on' + name]) {
+      this.upload['on' + name](e);
+    }
+  }
 }
 
 export class MockXMLHttpRequestInternetExplorer extends MockXMLHttpRequest {
@@ -264,4 +278,11 @@ export class MockXMLHttpRequestInternetExplorer extends MockXMLHttpRequest {
     return super.defaultResponseValue();
   }
 
+  triggerUploadEvent(name: any, eventObj?: any): void {
+    // TODO: create a better default event
+    const e: any = eventObj || {};
+    if (this['on' + name]) {
+      this['on' + name](e);
+    }
+  }
 }
--- a/spec/helpers/ajax-helper.ts
+++ b/spec/helpers/ajax-helper.ts
@@ -139,7 +139,7 @@ export class MockXMLHttpRequest {
   onerror: (e: ErrorEvent) => any;
   onprogress: (e: ProgressEvent) => any;
   ontimeout: (e: ProgressEvent) => any;
-  upload: XMLHttpRequestUpload;
+  upload: XMLHttpRequestUpload = <any>{ };
 
   constructor() {
     this.previousRequest = MockXMLHttpRequest.recentRequest;
@@ -158,6 +158,12 @@ export class MockXMLHttpRequest {
     this.password = password;
     this.readyState = 1;
     this.triggerEvent('readyStateChange');
+    const originalProgressHandler = this.upload.onprogress;
+    Object.defineProperty(this.upload, 'progress', {
+      get() {
+        return originalProgressHandler;
+      }
+    });
   }
 
   setRequestHeader(key: any, value: any): void {
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -652,6 +652,67 @@ describe('Observable.ajax', () => {
       expect(complete).to.be.true;
     });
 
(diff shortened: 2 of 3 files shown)


src/observable/dom/AjaxObservable.ts
```diff
@@ -304,14 +303,14 @@ export class AjaxSubscriber<T> extends Subscriber<Event> {
     (<any>xhrTimeout).request = request;
     (<any>xhrTimeout).subscriber = this;
     (<any>xhrTimeout).progressSubscriber = progressSubscriber;
-    if (xhr.upload && 'withCredentials' in xhr && root.XDomainRequest) {
+    if (xhr.upload && 'withCredentials' in xhr) {
```
kwonoj: is this passes browser test on IE9? Seems IE9 relies on `XDomainRequest` and `xdr.onprogress` as eventhandler.

spec/helpers/ajax-helper.ts
```diff
@@ -158,6 +158,14 @@ export class MockXMLHttpRequest {
     this.password = password;
     this.readyState = 1;
     this.triggerEvent('readyStateChange');
+    // https://developer.mozilla.org/en/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest
```
kwonoj: just fyi, since `ajax-helper` is mock this comment may need to be placed under `AjaxObservable` implementation.

spec/helpers/ajax-helper.ts
```diff
@@ -139,7 +139,7 @@ export class MockXMLHttpRequest {
   onerror: (e: ErrorEvent) => any;
   onprogress: (e: ProgressEvent) => any;
   ontimeout: (e: ProgressEvent) => any;
-  upload: XMLHttpRequestUpload;
+  upload: XMLHttpRequestUpload = Object.create(null);
```
kwonoj: what about simply do 

```js
upload = {
    progress: {
      get() {
        return this.upload.onprogress;
      }
    }
  }
```

? any strong reason to declare empty object then define property later?
(other than specifying type `XMLHttpRequestUpload`. This'll be easily solvable with TS2.1 by declare as `Partial<XMLHttpRequestUpload>`)

Brooooooklyn: It's not about Type declare... I just want mock the behavior that `listeners after open called would not be fired`. So I just want remove `setter` on `this.upload.progress` here

src/observable/dom/AjaxObservable.ts
Brooooooklyn: I don't know why `root.XDomainRequest`  here, it caused `progressSubscriber` only worked in IE9. @blesh

kwonoj: I think `&&` constraint is probably needed to be updated to deal each cases (`if xhr` vs `else if XDomainRequest`), but checking itself is still needed.

spec/helpers/ajax-helper.ts
kwonoj: I see. but `Object.create(null)` is required then?

Brooooooklyn: because https://github.com/Brooooooklyn/rxjs/blob/issue/ajax-upload-progress/src/observable/dom/AjaxObservable.ts#L319 would set `xhr.upload.onprogress` while progressSubscriber not empty.

kwonoj: OOOOOh I see. sorry for that. Sorry for those.

kwonoj: Seems it'd good to leave small comments to describe reason for this for further note.

src/observable/dom/AjaxObservable.ts
Brooooooklyn: I've checked the specs of XDomainRequest, and I'm sure the only difference between `xdr` and `xhr` in handling with events is here https://github.com/Brooooooklyn/rxjs/blob/9bc37772a71ec34b2cb5aef9f3f910b1177ee58f/src/observable/dom/AjaxObservable.ts#L316

spec/helpers/ajax-helper.ts
Brooooooklyn: updated commit

spec/observables/dom/ajax-spec.ts
```diff
@@ -652,6 +652,67 @@ describe('Observable.ajax', () => {
...
+
+      const request = MockXMLHttpRequest.mostRecent;
+
+      request.respondWith({
+        'status': 200,
+        'contentType': 'application/json',
+        'responseText': JSON.stringify({})
+      }, 3);
+
+      expect(spy).to.be.calledThrice;
```
blesh: LOL... is this really a thing? "Thrice"? Hahaha... How about just `callCount(3)` so I don't snicker every time I read it? 😄

Brooooooklyn: 😂

Brooooooklyn:
will this commit be landed in next version? I'm working on a fileUploader using RxJS, and I will be happy if this commit available before the end of November.

kwonoj:
>  if this commit available before the end of November

uh..since it's already nearby x-mas it missed train already? :)

If it's Dec. You mean, it depends on @blesh but may npt be available due to holiday season.

spec/observables/dom/ajax-spec.ts
```diff
@@ -763,12 +814,11 @@ describe('Observable.ajax', () => {
     const request = MockXMLHttpRequest.mostRecent;
 
     expect(() => {
-      request.onprogress((<any>'onprogress'));
+      request.upload.onprogress((<any>'onprogress'));
```
blesh: it seems like we should have to test both paths for this. It's tedious, but I want to make sure this works in every browser we can.

blesh: @Brooooooklyn @kwonoj thoughts?

blesh: On second thought. I think this can be remedied at a later time. This is fine for now.

blesh:
LGTM

PR #2201, opened 2016-12-19
fix(groupBy): Fix groupBy to dispose of outer subscription.

trxcllnt:
Not sure how groupBy is passing current unit tests, as we seem to be covering this case.

This PR fixes a bug where groupBy wasn't unsubscribing from its source when the outer subscription has been disposed and all inner subscriptions are inactive.

If you see [here](https://github.com/trxcllnt/rxjs/blob/04b7757dd6e661e7a7be48231d0261790327b9f5/src/operator/groupBy.ts#L250), the inner refCount subscription calls `parent.unsubscribe()` if `parent.count === 0` and `parent.attemptedToUnsubscribe` is true. But the parent only unsubscribes if `parent.attemptedToUnsubscribe` is false, which means the subscription to the source is never disposed.

This is easily replicated with groupBy on a Subject, where you'll see the observers list grow indefinitely.

### DIFF
--- a/spec/operators/groupBy-spec.ts
+++ b/spec/operators/groupBy-spec.ts
@@ -340,6 +340,34 @@ describe('Observable.prototype.groupBy', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
+  it('should unsubscribe from the source when the outer and inner subscriptions are disposed', () => {
+    const values = {
+      a: '  foo',
+      b: ' FoO ',
+      c: 'baR  ',
+      d: 'foO ',
+      e: ' Baz   ',
+      f: '  qux ',
+      g: '   bar',
+      h: ' BAR  ',
+      i: 'FOO ',
+      j: 'baz  ',
+      k: ' bAZ ',
+      l: '    fOo    '
+    };
+    const e1 = hot('-1--2--^-a-b-c-d-e-f-g-h-i-j-k-l-|', values);
+    const e1subs =        '^ !';
+    const expected =      '--(a|)';
+
+    const source = e1
+      .groupBy((val: string) => val.toLowerCase().trim())
+      .take(1)
+      .mergeMap((group: any) => group.take(1));
+
+    expectObservable(source).toBe(expected, values);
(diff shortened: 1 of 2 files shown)


kwonoj:
So we can't create create test coverage for this kind of cases?

trxcllnt:
@kwonoj I'm not sure. [This test](https://github.com/ReactiveX/rxjs/blob/master/spec/operators/groupBy-spec.ts#L314) looks to be covering this case. I tried making a test that broke without this patch applied, but couldn't get it to break outside of the production code I had. It's possible my production code synchronously unsubscribes, which would explain why I can't replicate with marble diagrams (yet).

trxcllnt:
@kwonoj if you run this example, you can see the `interval` subscription isn't being disposed:

```es6
import { Observable } from 'rxjs';

console.clear();

Observable
  .interval(100)
  .do(console.log.bind(console, 'interval'))
  .finally(console.log.bind(console, 'unsubscribed'))
  .take(10)
  .mapTo(1)
  .groupBy((x) => x)
  .take(1)
  .mergeMap((group) => group.take(1))
  .subscribe(
    console.log.bind(console, 'next'),
    console.log.bind(console, 'error'),
    console.log.bind(console, 'complete')
  );
/*
> Console was cleared
> interval 0
> next 1
> complete undefined
> interval 1
> interval 2
> interval 3
> interval 4
> interval 5
> interval 6
> interval 7
> interval 8
> interval 9
> unsubscribed
*/
```

blesh:
@trxcllnt can we definitively add a failing test case for this? I'd like to see that before we can merge this.

trxcllnt:
@blesh ok, I found a way to add a marble test that fails without this patch applied. the PR has been updated with this test.

mattpodwysocki:
LGTM

trxcllnt:
@blesh did this get merged??

blesh:
Hmmm... @trxcllnt I think this got merged into 5.0.x, which was deleted 😱. It's all good, I think we can merge this into master.

Issue #1633, opened 2016-04-18
Losing type inferences in marble tests


kwonoj:
**RxJS version:**
master

**Additional information:**

Currently to enhance `jsdoc` test code readability, test scheduler interface is not being directly imported but accessed globally with implicit ambient declaration, so `hot` or `cold` observable generated does not support any type inference to `Observable<T>`

``` js
declare const {hot, cold, asDiagram, expectObservable, expectSubscriptions};

var e1 = hot(...);

e1.audit().subscribe() //e1 does not gives any type context
```

This is somewhat problem since all unit test cases are intended to be used as type definition test cases too, which current test cases are losing its benefit in lot of cases.

kwonoj:
/cc @staltz for visibility, also @david-driscoll .

I've quickly tried if it's possible to use ambient declaration to test helper interfaces, but that wasn't successful since dependent types (`Observable`, `Subject`, `Scheduler`, etcs) are also need to be handcrafted since ambient declaration does not allow to import external type definition. 

@staltz for asking idea if we can solve this in `jsdoc` perspective, @david-driscoll for asking type definition perspective I couldn't think of.

staltz:
> Currently to enhance jsdoc test code readability, test scheduler interface is not being directly imported but accessed globally with implicit ambient declaration

I don't understand what does jsdoc have to do with the marble tests. Can't we make `hot()` and `cold()` generic functions `hot<T>()` and `cold<T>()`?

kwonoj:
> Can't we make hot() and cold() generic functions hot<T>() and cold<T>()?

: no, this isn't about it.

Do you remember original refactoring was importing test helper directly like

```js
import {hot} from ...

hot();
```

? it's transpiled by compiler something like

```js
var hot_1 = require(...);

hot_1();
```

so `jsdoc` test case will display some kind of ugly name. Instead, https://github.com/ReactiveX/rxjs/pull/1378 introduced to remove explicit import to come over those limitation of `jsdoc`. Now test cases are not importing explicitly, none of type inferences for generated Observable from `hot` or `cold`, type verification / inference including intellisense from editor does not work.

staltz:
Oh I see. Can't we declare the types of those functions in each file?

Like
```typescript
declare hot: string => Observable<any>;
```

(Just an example, not sure if that compiles)
It would incur some repeated code, on the other hand.

kwonoj:
[...]

PR #2202, opened 2016-12-20
test(filter): enable type inference to marble diagram

kwonoj:
⚠️ Targeting master ⚠️

- [x] Non-breaking change

**Description:**

This PR resolves long-pending issue #1633, provides type inference to marble-diagram based test cases while does not hurt documentation for specs. It is using TypeScript compiler's feature to import type signature without real module import and assign it to existing ambient declaration - ambient declaration will provide same type inference to exported modules.

**Why Targeting master?**

Originally test case were designed to be used as validating type definition as well for most common use cases to prevent issue like https://github.com/ReactiveX/rxjs/issues/2163 . Due to current setup ergonomics, all of tests are running without any compiler validation so there wasn't any regression testing for type definitions which would be considered as breaking change if occurs. 

This PR will enable all test cases to have correct compiler type inference, allows all test cases become first-guard to type definition validation. Once this PR is accepted, I'll work on other test cases to have same changes.

#1633

### DIFF
--- a/spec/operators/filter-spec.ts
+++ b/spec/operators/filter-spec.ts
@@ -1,6 +1,12 @@
 import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
-declare const {hot, cold, asDiagram, expectObservable, expectSubscriptions};
+import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
+
+declare const { asDiagram };
+declare const hot: typeof marbleTestingSignature.hot;
+declare const cold: typeof marbleTestingSignature.cold;
+declare const expectObservable: typeof marbleTestingSignature.expectObservable;
+declare const expectSubscriptions: typeof marbleTestingSignature.expectSubscriptions;
 
 const Observable = Rx.Observable;
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ cache:
 env:
   matrix:
   - NODE_VER=4 FULL_VALIDATE=false
-  - NODE_VER=6 FULL_VALIDATE=true alias grunt=./node_modules/grunt-cli/bin/grunt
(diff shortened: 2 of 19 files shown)


kwonoj:
I have assigned @david-driscoll , @staltz as reviewer for code / document side changes just in case.

kwonoj:
This is generated doc with PR, which preserves transpiled spec

image

with type inference supported.

image

spec/operators/filter-spec.ts
```diff
@@ -1,6 +1,12 @@
 import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
-declare const {hot, cold, asDiagram, expectObservable, expectSubscriptions};
+import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
```
staltz: How about `import * as marbleTestingSignature from '../helpers/marble-testing';`?

kwonoj: I haven't tried it but tempted to leave as-is to give clarity this import is unique kind and not being used in general.

staltz: That intent may be better expressed as a comment, not as the usage of `require`.

david-driscoll: could you not just destructure it with the import?

`import {hot, cold, asDiagram, expectObservable, expectSubscriptions} from '../helpers/marble-testing';`

kwonoj: No, it's #1633. If use imported type, TS will transpile it and doc will be affected, reason only use it as type to be removed while transpiling.

jayphelps: I think a comment here is super important. I can easily see myself wondering why this was done and changing it later.

kwonoj:
@jayphelps @blesh updated PR with changes to watch PR to validate certain behaviors like large PR, missing type definition import. It'll looks like https://github.com/ReactiveX/rxjs/pull/2327#issuecomment-277484394 

image

make status checker fail if those validation does not passes.

ladyleet:
🔥

PR #2211, opened 2016-12-22
docs(operator): Zip documentation

feloy:
**Description:**

Zip operator documentation

### DIFF
--- a/src/operator/zip.ts
+++ b/src/operator/zip.ts
@@ -67,23 +67,21 @@ export function zipStatic<R>(...observables: Array<ObservableInput<any> | ((...v
  * If the latest parameter is a function, this function is used to compute the created value from the input values. 
  * Otherwise, an array of the input values is returned.
  *
- * @example <caption>Combine age, name and gender from different sources</caption>
+ * @example <caption>Combine age and name from different sources</caption>
  *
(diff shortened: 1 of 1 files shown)


src/operator/zip.ts
```diff
@@ -61,6 +61,30 @@ export function zipStatic<R>(...observables: Array<ObservableInput<any> | ((...v
...
+ * of its input Observables.
+ *
+ * If the latest parameter is a function, this function is used to compute the created value from the input values. 
+ * Otherwise, an array of the input values is returned.
+ *
+ * @example <caption>Combine age, name and gender from different sources</caption>
+ *
+ * let Ages = Observable.of<number>(7, 5, 9);
+ * let Names = Observable.of<string>('Foo', 'Bar', 'Beer');
+ * let Gender = Observable.of<string>('F', 'M', 'F');
```
blesh: Can you lowercase these variable names? Also I think a lot of our other examples use Finnish Notation (i.e. `age$` `name$` `gender$`.

Finally, I think I'd like to exclude gender from this example and use something else. There are people who don't fit into the "M" or "F" pattern of gender, and in support of those folks, I think it would be best to stick to neutral material.

staltz:
LGTM

blesh:
Thanks @feloy!

PR #2213, opened 2016-12-22
docs(operators): examples for 'every' operator

feloy:
**Description:**

Examples for 'every' operator

### DIFF
--- a/src/operator/every.ts
+++ b/src/operator/every.ts
@@ -5,6 +5,25 @@ import { Subscriber } from '../Subscriber';
 
 /**
  * Returns an Observable that emits whether or not every item of the source satisfies the condition specified.
+ *
+ * @example <caption>A simple example emitting true if all elements are less than 5, false otherwise</caption>
+ *  Observable.of(1, 2, 3, 4, 5, 6)
+ *     .every(x => x < 5)
+ *     .subscribe(x => console.log(x)); // -> false
+ * 
+ * @example <caption>An example passing this (if you cannot use ES6 arrow function)</caption>
+ * class MyClass {
+ *     private isSmallEnough(n: number): boolean {
+ *         return n < 7;
+ *     }
+ *     constructor() {
+ *         Observable.of(1, 2, 3, 4, 5, 6)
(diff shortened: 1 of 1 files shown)


src/operator/every.ts
```diff
@@ -5,6 +5,25 @@ import { Subscriber } from '../Subscriber';
...
 /**
  * Returns an Observable that emits whether or not every item of the source satisfies the condition specified.
+ *
+ * @example <caption>A simple example emitting true if all elements are less than 5, false otherwise</caption>
+ *  Observable.of(1, 2, 3, 4, 5, 6)
+ *     .every(x => x < 5)
+ *     .subscribe(x => console.log(x)); // -> false
+ * 
+ * @example <caption>An example passing this (if you cannot use ES6 arrow function)</caption>
+ * class MyClass {
```
blesh: I think we can ditch the `class` here and just use a function and the observable.

src/operator/every.ts
```diff
@@ -5,6 +5,25 @@ import { Subscriber } from '../Subscriber';
...
+ *     private isSmallEnough(n: number): boolean {
+ *         return n < 7;
+ *     }
+ *     constructor() {
+ *         Observable.of(1, 2, 3, 4, 5, 6)
+ *             .every(function(x) { return this.isSmallEnough(x); }, this)
+ *             .subscribe(function(x) { console.log(x); });
+ *     } 
+ * }
+ * let c: MyClass = new MyClass(); // -> true
```
blesh: Related to above. I think the class just adds too much indirection to the example.

feloy: I don't understand. What would be 'this' in the case we remove the class and we keep only a function and the Observable?

blesh: The context argument to `every` is a legacy of `Array.prototype.every`. It's not really the important part of the example. Since `isSmallEnough` is a pure function, you can just call it directly. I don't think we need to use the `context` argument in the example.

feloy:
Ok, right. In fact I wanted to demonstrate the use of this context argument in this example. But I understand that it is not really the subject. 
This second example would not be necessary anymore, I propose to remove it completely

PR #2215, opened 2016-12-23
fix(TypeScript): interfaces that accepted `Scheduler` now accept `ISc…

blesh:
…heduler` interface

this should be a non-breaking change. It just makes sense to accept a minimal interface here rather than a full-on class

The full on class was giving me difficulties in another TypeScript project where I wanted to mock a Scheduler.

### DIFF
--- a/src/Scheduler.ts
+++ b/src/Scheduler.ts
@@ -1,6 +1,10 @@
 import { Action } from './scheduler/Action';
 import { Subscription } from './Subscription';
 
+export interface IScheduler {
+  now(): number;
+  schedule<T>(work: (this: Action<T>, state?: T) => void, delay?: number, state?: T): Subscription;
+}
 /**
  * An execution context and a data structure to order tasks and schedule their
  * execution. Provides a notion of (potentially virtual) time, through the
@@ -17,7 +21,7 @@ import { Subscription } from './Subscription';
  *
  * @class Scheduler
  */
-export class Scheduler {
+export class Scheduler implements IScheduler {
 
   public static now: () => number = Date.now ? Date.now : () => +new Date();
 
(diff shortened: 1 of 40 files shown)


src/Scheduler.ts
```diff
@@ -17,7 +21,7 @@ import { Subscription } from './Subscription';
  *
  * @class Scheduler
  */
-export class Scheduler {
+export class Scheduler implements IScheduler {
 
   public static now: () => number = Date.now ? Date.now : () => +new Date();
```
kwonoj: if `now` is defined in interface, shouldn't this be instance method as well to allow override?

blesh: Interesting. I don't think this should be static at all. Not sure how that crept in there.

trxcllnt: @blesh from what I recall, it's static to mirror the [Rx4 Scheduler API](https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/concurrency/scheduler.js#L45)

blesh: What was the purpose there? Not that it pertains to this PR

blesh: What was the purpose there? Not that it pertains to this PR

trxcllnt: I believe schedulers were at one point static classes in Rx.NET, but could be misremembering.

kwonoj: one question for this is, if this becomes instance method will this be breaking change for next major semver?

trxcllnt: It is already both static and instance method (assigned via the constructor)

kwonoj: 😮 oh, thanks for clarification.

blesh: If anything, we might deprecate and remove the static version at some point.

PR #2217, opened 2016-12-25
refactor(Error): Rewrite custom derived errors with Object.create()

saneyuki:
**Description:**

- This is the alternative proposal to fix https://github.com/ReactiveX/rxjs/issues/2178 with rewriting custom derived errors by `Object.create()`'s classical inheritance by hand.
  - This approach is based on TypeScript's definitions for build-in object.
  - By this change, we can upgrade TypeScript 2.1 or later without dropping support for IE9/IE10 for the future.
- This is not a breaking change because this patch creates an inheritance by hand instead of TypeScript's transformer.

https://github.com/ReactiveX/rxjs/issues/2178

### DIFF
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -262,9 +262,12 @@ describe('Observable.ajax', () => {
       'responseText': 'Wee! I am text!'
     });
 
+    expect(error instanceof Error).to.be.true;
     expect(error instanceof Rx.AjaxError).to.be.true;
+    expect(error.name).to.equal('AjaxError');
     expect(error.message).to.equal('ajax error 404');
     expect(error.status).to.equal(404);
+    expect(error.xhr instanceof XMLHttpRequest).to.be.true;
   });
 
   it('should fail on 404', () => {
@@ -294,9 +297,12 @@ describe('Observable.ajax', () => {
       'responseText': 'Wee! I am text!'
     });
 
+    expect(error instanceof Error).to.be.true;
     expect(error instanceof Rx.AjaxError).to.be.true;
+    expect(error.name).to.equal('AjaxError');
     expect(error.message).to.equal('ajax error 300');
     expect(error.status).to.equal(300);
+    expect(error.xhr instanceof XMLHttpRequest).to.be.true;
   });
 
   it('should succeed no settings', () => {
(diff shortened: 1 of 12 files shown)


saneyuki:
@kwonoj how about this?

kwonoj:
Personally, I'd like to land https://github.com/ReactiveX/rxjs/pull/2202 (and following PR) prior to any-targeting-master changes to enable type inference-validation on each changes. Though this looks solid (and local verification works), always better to have test supports.

saneyuki:
I added some test cases for this change.

blesh:
Interesting. I played around with this same approach for a few minutes and TS 2.1 didn't like it when I tried. Heh. Looks good to me, though.

DanielRosenwasser:
For the record, you can create an intermediate class that "does the right thing" by not explicitly returning from the constructor itself. You could write it as an ES5-style class (much the same way the old code would be compiled), and then derive your classes from *that* class. That way, much less of the code actually has to be rewritten.

```ts
var __extends = (this && this.__extends) || function (d, b) {
    for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
    function __() { this.constructor = d; }
    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
};

export interface ShimErrorStatic {
    new (message?: string): ShimError;
}
export interface ShimError extends Error {}

export const ShimError: ShimErrorStatic = (function (_super) {
    __extends(C, _super);
    function C(message) {
        _super.apply(this, arguments);
        // ensure that 'message' gets set correctly - wasn't case with old implementation.
        this.message = message;
    }
    return C;
}(Error)) as any;
```

The only downside is that if you're compiling to ES2015, this winds up being an extra layer that you may find undesirable.

kwonoj:
> compiling to ES2015, this winds up being an extra layer

@DanielRosenwasser would you able to share how much will it be? I expect ES15 package will be published eventually, so if those are unacceptable amount probably would better to go with existing changes in this PR.

DanielRosenwasser:
Well you'd have to add the `ShimError` function that I posted above, and then add an import for each class that extends `Error`.

DanielRosenwasser:
Actually, I wonder if you could get away with this, which doesn't need the helper function:

```ts
export interface RxErrorStatic {
    new (message?: string): RxError;
}
export interface RxError extends Error {}

export const RxError: RxErrorStatic = function (_super) {
    function RxError(this: Error, message: string) {
        const err = Error.call(this, message);
        this.message = message;
        this.stack = err.stack;
        return err;
    }
    RxError.prototype = Object.create(Error.prototype, {
        constructor: {
            value: RxError,
            enumerable: false,
            writable: true,
            configurable: true,
        },
    });
    return RxError;
}(Error) as any;
```

saneyuki:
@kwonoj 

What should I do about this? I feel the current patch is not different from @DanielRosenwasser proposed one.

benlesh:
We solved this issue a while back, but I think you were involved with that. Either way, thanks for the contribution, @saneyuki. Closing this one for now.

Issue #2088, opened 2016-10-26
FromEventPattern does not conform to v4


mattpodwysocki:
The current implementation of [`fromEventPattern`](https://github.com/ReactiveX/rxjs/blob/master/src/observable/FromEventPatternObservable.ts) deviates from the current v4 implementation in several ways:
- `removeHandler` is no longer optional
- `removeHandler` does not get return value from `addHandler`

Is the intent here to follow this behavior, or deviate?  If we are to deviate, then we must ask why the `addHandler` is `addHandler: (handler: Function) => any` instead of `addHandler: (handler: Function) => void` if we aren't returning anything, nor are we using the value produced from it.

**RxJS version:** v5 - current

**Code to reproduce:**

This is an example of using RxJS v4 with the `addHandler` returning a value, which would then be forwarded onto the remove handler.

``` js
Observable.fromEventPattern(
  (h) => { return elem.connect('data', h); },
  (_, r) => { r.disconnect(); }
);
```

**Expected behavior:**

The expected behavior would forward the return value from `elem.connect('data', h)` so that you could remove the handler from the value by calling `disconnect`.

**Actual behavior:**

Currently, RxJS v5 does not support doing anything with the `addHandler` despite its signature of allowing anything to be returned, hence why its signature of `addHandler: (handler: Function) => any` is confusing if nothing is ever done with the return value.

**Additional information:**

If this is indeed breaking change behavior, we should put that in the documentation and the reasons for this.  The reason it was added originally was that some libraries instead pass a subscription like object to the user which you can later disconnect instead of the standard Subject/Observer pattern with add/remove handlers.

blesh:
cc/ @trxcllnt who I believe is the original implementer. (this was added when RxJS was v 2 or 3 I think)

trxcllnt:
@blesh I don't know how I got associated with implementing this, I've [literally never touched](https://github.com/ReactiveX/rxjs/blame/master/src/observable/FromEventPatternObservable.ts) this file. I'm halfway convinced you're messing with me :P

@mattpodwysocki I agree 100%, this should exactly match v4 behavior, something I would've done if I was the [author](https://github.com/ReactiveX/rxjs/blob/1095d4cc98ad9ca7f959ebc631b37a43255de15d/src/observables/FromEventPatternObservable.ts) of this version \*cough\*@blesh\*cough\*

jayphelps:
[...]

Issue #1900, opened 2016-08-24
fromEventPattern signal missing


dwieeb:
**RxJS version:** 5.0.0-beta.11 

In v4, there was a way to take the _signal_ (the thing returned from `addHandler`) and use it in `removeHandler` for proper disposal. [See v4 docs for `fromEventPattern`](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/fromeventpattern.md#example), under the Dojo toolkit example. I think this functionality is still useful and it [appears to be missing](https://github.com/ReactiveX/rxjs/blob/master/src/observable/FromEventPatternObservable.ts#L60).

I can attempt to submit a PR, but I was wondering if there was a reason for its removal.

blesh:
cc @trxcllnt

trxcllnt:
@blesh was I responsible for `fromEventPattern`? If so, I apologize, and this should definitely be fixed.

blesh:
@trxcllnt no worries, it was an add you made very very early on in development. I can look at it too, I was just traveling when this crossed my screen.

kwonoj:
Closing via #2224.

PR #2224, opened 2016-12-28
Align fromEventPattern behavior to RxJS v4

kwonoj:
**Description:**

This PR aligns `fromEventPattern`'s behavior to RxJS v4, including 2 changes
- `removeHandler` is now optional
- `removeHandler` signature now accepts optional `signal` parameter, `(handler: Function, signal?: any) => void` forwarded value of `addHandler` for eventlistener disposal.

This PR is currently ⚠️ Targeting master ⚠️, since both of changes are optional probably won't breaking any existing usecases. 

#1900, #2088

### DIFF
--- a/src/observable/FromEventPatternObservable.ts
+++ b/src/observable/FromEventPatternObservable.ts
@@ -1,3 +1,4 @@
+import { isFunction } from '../util/isFunction';
 import { Observable } from '../Observable';
 import { Subscription } from '../Subscription';
 import { Subscriber } from '../Subscriber';
@@ -45,7 +46,7 @@ export class FromEventPatternObservable<T> extends Observable<T> {
    * @param {function(handler: Function): any} addHandler A function that takes
    * a `handler` function as argument and attaches it somehow to the actual
    * source of events.
-   * @param {function(handler: Function, signal?: any): void} removeHandler A function that
+   * @param {function(handler: Function, signal?: any): void} [removeHandler] An optional function that
    * takes a `handler` function as argument and removes it in case it was
    * previously attached using `addHandler`. if addHandler returns signal to teardown when remove,
    * removeHandler function will forward it.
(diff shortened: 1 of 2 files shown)


src/observable/FromEventPatternObservable.ts
```diff
@@ -45,7 +46,7 @@ export class FromEventPatternObservable<T> extends Observable<T> {
    * @param {function(handler: Function): any} addHandler A function that takes
    * a `handler` function as argument and attaches it somehow to the actual
    * source of events.
-   * @param {function(handler: Function): void} removeHandler A function that
+   * @param {function(handler: Function, signal?: any): void} [removeHandler] An optional function that
```
jayphelps: signal should be documented as well in this description

jayphelps:
@kwonoj the code looks good, but it's not clear to me what the use case is? Can you provide one for insight?

kwonoj:
I don't have strong usecase of `signals` by myself, it's aligning interface to v4 and support some of edge case eventhandler pattern like `dojo`(https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/fromeventpattern.md#example)

for `removeHandler` part, I do think it doesn't necessarily be required param, rather be choices per consumption.

mattpodwysocki:
@jayphelps here's the full reason on our [Bridging to Events](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/gettingstarted/events.md#bridging-to-custom-events-with-fromeventpattern) documentation.  Sometimes, when adding an event returns a complex object instead of the simple Subject/Observer pattern, so we need to thread it through to the remove handler.

jayphelps:
Seems legit.

src/observable/FromEventPatternObservable.ts
```diff
@@ -75,10 +77,15 @@ export class FromEventPatternObservable<T> extends Observable<T> {
...
+    const retValue = this._callAddHandler(handler, subscriber);
+
+    if (!isFunction(removeHandler)) {
+      return;
+    }
+
     subscriber.add(new Subscription(() => {
       //TODO: determine whether or not to forward to error handler
-      removeHandler(handler);
+      removeHandler(handler, retValue) ;
```
mattpodwysocki: Any teardown should never throw, so I'd let it be

kwonoj:
Uh...
```
1 failing
  1) Observable.prototype.observeOn should clean up subscriptions created by async scheduling (prevent memory leaks #2244):
      Uncaught AssertionError: expected 3 to equal 2
      + expected - actual
      -3
      +2
```

https://travis-ci.org/ReactiveX/rxjs/jobs/189285828

Seems we have intermittent test failure?

jayphelps:
@blesh ^^ can you take a look? https://github.com/ReactiveX/rxjs/commit/9664a38ec2bb269ef87ae51c1eabab1ab8f2a48b

blesh:
Ug... it looks like the observeOn test is breaking this one... The last run might predate the fixes to it.

kwonoj:
Rebased and CI becomes green. I'll check this is later.

Issue #2045, opened 2016-10-17
Export ajax-helper in the released module


rgbkrk:
**RxJS version:**

5.0.0-rc1

**Additional information:**

I'd love to be able to fully test a module I'm working on, where I make a call to `webSocket` (yet don't use it). I noticed that [the websocket tests](https://github.com/ReactiveX/rxjs/blob/418d59769e8e61d5a7dc37235fe3862494ea35a3/spec/observables/dom/webSocket-spec.ts) use a [`MockWebSocket` from ajax-helper](https://github.com/ReactiveX/rxjs/blob/418d59769e8e61d5a7dc37235fe3862494ea35a3/spec/helpers/ajax-helper.ts). Could that be exported so Rx users can rely on it for their own testing?

kwonoj:
There is ongoing effort to expose public interface for creating test (https://github.com/ReactiveX/rxjs/issues/1775), it may discussable to include those helper to public or not. 

For me it's bit on fence if it's needed to include those mock object or not, since those are generally easily mockable via existing mocking framework (based on test framework, like `sinon`, or `jest`, etcs).

blesh:
cc @jayphelps

blesh:
@rgbkrk this might even be worth just busting into it's own module, honestly. It's not really specific to RxJS.

kwonoj:
I was actually recently thought of get rid of those mocks and using `sinon` instead.

jayphelps:
I'm not a testing pro--but I've recently started testing the websocketsubject with redux-observable by just mocking the API I provide.

```js
class FakeWebSocket extends Observable {
  constructor(response$) {
    this.response$ = response$;
  }

  multiplex(sub, unsub, messageFilter) {
    this.subMsg = sub();
    this.unsubMsg = unsub();
    return this.response$.filter(messageFilter);
  }
}
```

So say you have an API
```js
// api.js
export const ticketSocket = Observable.webSocket('ws://whatever');
```

and an epic
```js
export const stockTickerEpic = (action$, store, { ticketSocket }) =>
  action$.ofType('START_TICKER_STREAM')
    .mergeMap(action => 
      ticketSocket.multiplex(
        () => JSON.stringify({ sub: action.ticker }),
        () => JSON.stringify({ unsub: action.ticker }),
        msg => msg.ticker === action.ticker
      )
      .map(tick => ({ type: 'TICKER_TICK', tick }))
    );
```

[...]

PR #2225, opened 2016-12-28
remove separate ajaxHelper module

kwonoj:
**Description:**

⚠️ Targeting master ⚠️

This PR is simple refactoring for test cases around `ajaxObservable`, by removing separate module `ajax-helper` and let each test case embeds it where it needed. 

Primary reason for this is removing `require` dependency in `mocha` tests - I'm currently trying to simplify test / build script pipeline, which needs to remove external helpers as much to make it possible. Also this will clears #2045 as well, now internal mock for test is no longer exported at all.

This PR does not includes any breaking changes.

#2045


blesh:
@kwonoj can you resolve these conflicts?

PR #2230, opened 2016-12-31
docs(tutorial): Log target values and not the events

jtefera:
We want to log the input values and not the keypress events.


blesh:
LGTM, thanks @jtefera! :heart:

PR #2231, opened 2016-12-31
docs(tutorial): Changed values to keys in the examples

jtefera:
**Description:**
It seems that there is a confusion between `event.key` and `event.target.value` in the doc.
As "hello world" is typed in the input box, a stream of keypress events will give as the output:

    "h"
    "e"
    "l"
    "l"
    ...

for `event.key` and will output:

    "h"
    "he"
    "hel"
    "hell"
    "hello"
    ...
for `event.target.value`.
The examples seem to imply to want the first behavior but they are using the second method.
Changed the examples code to use the first method. 

Another possibility would be to correct the examples having in mind the second case.

### DIFF
--- a/doc/tutorial/basics.md
+++ b/doc/tutorial/basics.md
@@ -81,23 +81,23 @@ input.takeUntil(stopStream)
 // typing "hello world"
 var input = Rx.Observable.fromEvent(document.querySelector('input'), 'keypress');
 
-// Pass on a new value
-input.map(event => event.target.value)
+// Pass on a new key press
+input.map(event => event.key)
   .subscribe(value => console.log(value)); // "h"
 
-// Pass on a new value by plucking it
-input.pluck('target', 'value')
+// Pass on a new key press by plucking it
+input.pluck('key')
   .subscribe(value => console.log(value)); // "h"
 
-// Pass the two previous values
-input.pluck('target', 'value').pairwise()
-  .subscribe(value => console.log(value)); // ["h", "e"]
+// Pass paired the two previous key presses
+input.pluck('key').pairwise()
+  .subscribe(value => console.log(value)); // ["h", "e"], ["l", "l"], ...
 
-// Only pass unique values through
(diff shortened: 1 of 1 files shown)


blesh:
@jtefera I think you're onto something here, I'm not sure this example makes sense. However, I'm not sure the changes make sense either. I think the whole example needs looked at maybe?

cc/ @staltz @kwonoj

doc/tutorial/basics.md
```diff
@@ -81,23 +81,23 @@ input.takeUntil(stopStream)
...
+// Pass on a new key press by plucking it
+input.pluck('key')
   .subscribe(value => console.log(value)); // "h"
 
-// Pass the two previous values
-input.pluck('target', 'value').pairwise()
-  .subscribe(value => console.log(value)); // ["h", "e"]
+// Pass paired the two previous key presses
+input.pluck('key').pairwise()
+  .subscribe(value => console.log(value)); // ["h", "e"], ["l", "l"], ...
```
staltz: Actually pairwise would produce `["h", "e"]`, `["e", "l"]`, `["l", "l"]`...

staltz:
I'm favorable for these changes. However, @christianalfoni originally wrote these examples, maybe we can pick feedback from him.

christianalfoni:
Hi there!

I would state that this change would make sense if you just listened to keypresses on the `window` or `document`, meaning that you actually track keypresses. This does not make sense on an input though, because you might for example have the word `hello` as an input, then you make a selection around `ll` and then hit a new character, `b`, which results in the value `hebo`. 

But I think `keypress` is the wrong event here actually, cause it does not trigger on backspace. 

So yeah, I would state it should either listen to `document` or `window` with a keypress event, as changed here. Or it stays as an `input`, but the `keypress` event needs to be changed to `input` :)

Hope this made some sense!

staltz:
Closing this PR for being too old without an update. We can easily reopen it if when the PR author wishes to update it.

Issue #2232, opened 2016-12-31
subject.hasObserver not deprecated in version 4 but not included in version 5


jamesrhaley:
**RxJS version:**
5.0.2

**Code to reproduce:**
var subject = new Rx.Subject();

console.log(subject.hasObservers());

**Expected behavior:**
false

**Actual behavior:**
TypeError: subject.hasObservers is not a function

**Additional information:**
hasObserver has not been listed as deprecated in version 4 or what it has been migrated to in version 5 but as far as I can see, it is no longer an instance method of subject anymore.

benlesh:
Closing this due to lack of interest.

PR #2233, opened 2017-01-01
fix(Subject): Limit Subject interface public accessibility

kwonoj:
**Description:**
This PR updates interface of `Subject`. I came to notice current Subject interface expose its properties too wide to public scope, any instance of Subject exposes state variables like `observers`, `isStopped` can be mutated. This PR limits those variables to inherited scope only, and expose it as readonly property.

It is certainly breaking change since anyone have used those property to change Subject's state now it won't work - but not sure how popular usecase it'll be. (and is it recommended pattern?) For now, I've targeted this into master until gets reviewed and approved.

BREAKING CHANGE: if there's code change state of Subject via publicly exposed properties, it won't work anymore

Related note, is it worth #2232 to add `Subject:hasObservers()` interface or should recommend to use `Subject::observers.length` and update migration guide?

### DIFF
--- a/src/ReplaySubject.ts
+++ b/src/ReplaySubject.ts
@@ -41,7 +41,7 @@ export class ReplaySubject<T> extends Subject<T> {
     } else if (this.isStopped) {
       subscription = Subscription.EMPTY;
     } else {
-      this.observers.push(subscriber);
+      this._observers.push(subscriber);
       subscription = new SubjectSubscription(this, subscriber);
     }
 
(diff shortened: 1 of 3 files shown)


blesh:
This is akin to my comments on #2234... I'm not sure what this is buying us.

kwonoj:
I'm closing this for now as same reason of https://github.com/ReactiveX/rxjs/pull/2234.

PR #2234, opened 2017-01-01
fix(Subscription): Limit Subscription::closed public accessibility to readonly

kwonoj:
**Description:**

This PR changes the signature of `ISubscription::closed` to `readonly`. 
Currently, it is allowed to change state of `closed` as well as `unsubscribe()` but those two are causing different outcome by `unsubscribe` actually does necessary teardown for unsubscription, while `closed` flag only changes state flag. Instead, this PR explicitly makes given flag as getter only to read state, otherwise let use `unsubscribe()` if needed.

This is indeed breaking change if anyone has been using `closed` to control subscription state, it won't work anymore. For now, I have targeted this to master until gets reviewed and approved.

BREAKING CHANGE: Cannot change closed state without unsubscribe()

### DIFF
--- a/spec/observables/ScalarObservable-spec.ts
+++ b/spec/observables/ScalarObservable-spec.ts
@@ -1,4 +1,5 @@
 import {expect} from 'chai';
+import * as sinon from 'sinon';
 import * as Rx from '../../dist/cjs/Rx';
 import {ScalarObservable} from '../../dist/cjs/observable/ScalarObservable';
 
@@ -18,11 +19,14 @@ describe('ScalarObservable', () => {
   });
 
   it('should not schedule further if subscriber unsubscribed', () => {
-    const s = new ScalarObservable(1, rxTestScheduler);
+    const schedulerMock = sinon.stub(rxTestScheduler);
+    const s = new ScalarObservable(1, schedulerMock as any);
     const subscriber = new Rx.Subscriber();
     s.subscribe(subscriber);
-    subscriber.closed = true;
+    subscriber.unsubscribe();
     rxTestScheduler.flush();
+
+    expect((schedulerMock as any).schedule).calledOnce;
   });
 
   it('should set `_isScalar` to true when NOT called with a Scheduler', () => {
--- a/src/SubjectSubscription.ts
+++ b/src/SubjectSubscription.ts
@@ -8,9 +8,7 @@ import { Subscription } from './Subscription';
  * @extends {Ignored}
  */
 export class SubjectSubscription<T> extends Subscription {
-  closed: boolean = false;
-
-  constructor(public subject: Subject<T>, public subscriber: Observer<T>) {
+  constructor(private subject: Subject<T>, private subscriber: Observer<T>) {
     super();
   }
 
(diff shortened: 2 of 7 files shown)


spec/observables/ScalarObservable-spec.ts
```diff
@@ -18,11 +19,14 @@ describe('ScalarObservable', () => {
...
-    const s = new ScalarObservable(1, rxTestScheduler);
+    const schedulerMock = sinon.stub(rxTestScheduler);
+    const s = new ScalarObservable(1, schedulerMock as any);
     const subscriber = new Rx.Subscriber();
     s.subscribe(subscriber);
-    subscriber.closed = true;
+    subscriber.unsubscribe();
     rxTestScheduler.flush();
+
+    expect((schedulerMock as any).schedule).calledOnce;
```
kwonoj: while looking codes I noticed this test case actually did not assert anything, modified to have assertion but not sure if this is intended one - will update accordingly as suggested.

blesh:
I'm not sure why this change is necessary. How does it effect perf?

kwonoj:
Way I'm seeing this change is enhancing ergonomics with small complexity addition in codebases. 

It is indeed this codebase is javascript and anyone can actually do anything if they try to avoid given interfaces, on the other hand if library's interface prevents any accidental incorrect usage, it would be better to prevent user's confusion.

Perfwise, previously it is just property access and with this change it'll be getter function, `() => boolean`. I do agree this adds complexity course, but it's small, though gain is also not quite hugh - small complexity, small gain.

If this change doesn't seems give sufficient benefit to library users, I'd like at least propose to update signature of `ISubscription` only which is free gain for typescript user without any codebase complexity- created https://github.com/ReactiveX/rxjs/pull/2249 for it.

blesh:
Honestly, @kwonoj, I think I'm :-1: on this change and other changes like it. I'm happy with TypeScript enforcement (and possible _name "enforcement" in "private" or "protected" variables cases) for now.

I'm willing to be convinced otherwise, of course. But right now I'm not feeling it.

kwonoj:
@blesh fine for me. Just one thing, can we add this as agenda for how much of runtime guarantee / validation we'd like to provide? Currently codebase barely does not provide anything on it, and this PR is kind of those to adding some more soundness on runtime (by introducing runtime complexity). As you can see, there are some of opened issues to demand certain level of runtime validation which non-TS user could benefit. User of TS this doesn't be too much issues for myself, but worth to think of where to draw those lines (such as it's already good as-is vs. we need something more)

Closing PR without checking it.

blesh:
@kwonoj that's fine. I'm happy to discuss it.

I'm on the side of: "If you want soundness, use a type system, otherwise this is JavaScript and I'll duck-punch all the things I want"

Although, I really would like to go through and rename all private and protected variables to have underscores before them. That's been bothering me.

kwonoj:
>  I really would like to go through and rename all private and protected variables to have underscores before them. That's been bothering me.

: That maybe sufficient soundness of runtime level we'd like to achieve. I think having settled down line would be great whatever level we'd like to have.

Issue #2237, opened 2017-01-01
Previously made subscriptions with same observer are unsubscribed when subsequent subscription completes


oscarlorentzon:
**RxJS version:**

`5.0.2`

**Code to reproduce:**

```ts
let source = Observable
    .timer(1000, 1000)
    .take(2);

let observer = {
    next: () => { }
};

source
    .do(v => console.log("a", v))
    .subscribe(observer);

source
    .do(v => console.log("b", v))
    .subscribe({
        next: () => { }
    });

Observable.empty()
    .subscribe(observer);
```

**Expected behavior:**

Expected console output:
```js
a 0
b 0
a 1
b 1
```

**Actual behavior:**

Actual console output:
```js
b 0
b 1
```

**Additional information:**

The root cause for this behavior is the following logic:

https://github.com/ReactiveX/rxjs/blob/6922b16f4a297a4acbe8ad2b086c947b7c0d8315/src/Subscriber.ts#L174-L177

The observer instance in the example above will get the unsubscribe function of the `Subscriber` attached to it in the first source `subscribe` call. In the final `subscribe` call to the empty observable the observer instance already has the attached unsubscribe function which will be added to the `Subscription`. When the empty observable subscription completes, which happens immediately, a chain reaction will occur. The empty observable will unsubscribe on complete and it will subsequently call the previously added unsubscribe function which is related to the first source subscription. That subsription will also be unsubscribed and closed.

Another example of the problem is when `subscribe` is called with no arguments:

```ts
let source = Observable
    .timer(1000, 1000)
    .take(2);

source
    .do(v => console.log("a", v))
    .subscribe();

source
    .do(v => console.log("b", v))
    .subscribe(() => { });

Observable.empty()
    .subscribe();
```

Expected console output:
```js
a 0
b 0
a 1
b 1
```

Actual console output:
```js
b 0
b 1
```

This was introduced in `5.0.0-rc.1` and is caused by the fact that the `Observer.empty` instance is used for all calls to `toSubcriber` with no arguments:

https://github.com/ReactiveX/rxjs/blob/6922b16f4a297a4acbe8ad2b086c947b7c0d8315/src/util/toSubscriber.ts#L21

This logic also means that a larger chain reaction can occur when multiple calls to subscribe with no arguments on the same or different source observables are made. If the last subscription completes, all previous subscriptions that were made with no arguments will be unsubscribed and closed.

[...]

PR #2238, opened 2017-01-01
test(Subscriber, toSubscriber, Observable): Add unit tests and solution for erroneous subscriber complete behavior

oscarlorentzon:
**Description:**

~~This PR adds failing unit tests related to the issue below. This PR does not resolve the underlying issue.~~ Now contains a fix of the problem.

**Related issue:**
#2237


trxcllnt:
@oscarlorentzon I've issued a pull to this PR at https://github.com/oscarlorentzon/rxjs/pull/1. If you hit merge there, we can merge it in cleaner here.

oscarlorentzon:
@trxcllnt PR merged.

jayphelps:
Is this ready for review? It looks like it now includes a solution, not just tests?

oscarlorentzon:
@jayphelps Correct, after merging @trxcllnt's PR this PR now contains a solution too. Updated the PR title accordingly. 

It is ready for review.

spec/Observable-spec.ts
```diff
@@ -193,6 +193,80 @@ describe('Observable', () => {
...
+      });
+
+      source.subscribe();
+
+      Observable.empty().subscribe();
+
+      setTimeout(() => {
+        let err;
+        let errHappened = false;
+        try {
```
blesh: This try-catching seems unnecessary. Maybe I'm misunderstanding the purpose. Can you clairify?

blesh:
@oscarlorentzon this looks good... I just have a question about the try-catching in the tests, can you have a look and respond? Otherwise, this LGTM.

spec/Observable-spec.ts
```diff
@@ -193,6 +193,80 @@ describe('Observable', () => {
       source.subscribe();
     });
 
+    it('should work when subscribe is called with no arguments and other subscription with no arguments completes', (done: MochaDone) => {
+      let nextCalled = false;
+      const source = new Observable((subscriber: Rx.Subscriber<string>) => {
+        const id = setTimeout(() => {
```
blesh: This should be using the `rxTestScheduler` instance rather than `setTimeout` so the test can be deterministic.

spec/Observable-spec.ts
```diff
@@ -193,6 +193,80 @@ describe('Observable', () => {
...
+
+      let observer = {
+        next: function () { /*noop*/ }
+      };
+
+      source.subscribe(observer);
+
+      Observable.empty().subscribe(observer);
+
+      setTimeout(() => {
```
blesh: Same thing here, this should use the test scheduler.

oscarlorentzon:
@blesh I have simplified the two tests in `spec/Observable-spec.ts` related to your comments by inverting the logic. After those changes there is no need to use the `rxTestScheduler`. The changes also makes the tests more readable. They are now synchronous. They still fail on the codebase before @trxcllnt's fix. They succeed on the codebase after that fix.

I am not sure why the coverage has decreased in the latest build when this PR basically only adds new tests. The last commit did not remove any test. Do I need to change something regarding that for this PR to be merged?

A note, the try catch was there because of how errors needs to be handled in async mocha tests: https://mochajs.org/#asynchronous-code

kwonoj:
Coveralls details page indicates few lines are missing with this PR in subscription / subject, but not sure why. It may possible coveralls detection is not accurate.

trxcllnt:
@blesh I think it's time we merged, as we're getting more and more reports on this

trxcllnt:
I'm gonna go ahead and merge this in since we're getting so many issues about it.

christophechevalier:
Hi, anyway to get this released very soon (in 5.2.1 or something like that)? Thanks :)
(FYI @victornoel)

victornoel:
@trxcllnt sorry to insist, but this bug is touching some subscriptions that are actually made by angular itself, so it is not possible for us to use the workaround where `() => {}` is passed to the subscribe function.
And the issue arises in a complete non-deterministic way (we add a some subscription to our ngrx store and it breaks another subscription made via angular html async operator…).

victornoel:
ping @benlesh too, sorry for the bother…

MadMub:
hello all, not sure on the rxjs release process, did this make it into 5.3?

trxcllnt:
@MadMub yes it did

jayphelps:
@MadMub btw if you're ever wondering if some merged PR made it into a release, click on the merge commit (or commit that was added on your behalf, for rebases) and you can see a list of branches, the PR number, and a list of Release tags this code is included in. In this case, you'll see the tag `5.3.0`. If you don't see a release tag, it probably hasn't been released--though sometimes people do weird rewriting of history that prevents this feature from working.

image

PR #2240, opened 2017-01-02
fix(windowTime): clean up closed window with timeSpanOnly

kwonoj:
**Description:**
This PR adds cleanup to `windowTime` operator to clear up closed instance of window when invoked via timespan only.

### DIFF
--- a/src/operator/windowTime.ts
+++ b/src/operator/windowTime.ts
@@ -63,7 +63,6 @@ export function windowTime<T>(this: Observable<T>, windowTimeSpan: number,
 }
 
 class WindowTimeOperator<T> implements Operator<T, Observable<T>> {
-
   constructor(private windowTimeSpan: number,
               private windowCreationInterval: number,
               private scheduler: Scheduler) {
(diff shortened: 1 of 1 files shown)


kwonoj:
Not sure how to test this cases? :/

src/operator/windowTime.ts
```diff
@@ -147,52 +163,35 @@ class WindowTimeSubscriber<T> extends Subscriber<T> {
...
-  windowTimeSpan: number;
-  subscriber: WindowTimeSubscriber<T>;
-}
-
-function dispatchWindowTimeSpanOnly<T>(this: Action<TimeSpanOnlyState<T>>, state: TimeSpanOnlyState<T>) {
+function dispatchWindowTimeSpanOnly<T>(this: Action<TimeSpanOnlyState<T>>, state: TimeSpanOnlyState<T>): void {
   const { subscriber, windowTimeSpan, window } = state;
   if (window) {
-    window.complete();
+    subscriber.closeWindow(window);
```
blesh: @kwonoj should've probably saved all of the typing refactors and formatting stuff for a different commit. It's hard for me to see what the "meat" of this PR is.. I'm assuming it's mostly this line here?

kwonoj: yes, correct. previously timspan'ed window closed, but not teared down as same as default window behavior. Now both uses same teardown path. Sorry for creating commit include mixed changes, should I split up and update PR?

blesh: Yeah, we might want to. This is a commit to merge in a fix to production code... so we might want to make sure the changes are isolated this time.

kwonoj: will do. I'll rather create as separate PR to have better scoped clarity.

PR #2242, opened 2017-01-02
docs(tutorial): wrong keyboard event

Gregoirevda:
The `keypress` event is send before the value has been set on the input element : it always returns the last value of the element.
When `keyup` is send, the value has already been set on the input; returning the actual value.

**Description:**

### DIFF
--- a/doc/tutorial/basics.md
+++ b/doc/tutorial/basics.md
@@ -48,7 +48,7 @@ Which one you choose depends on the scenario. The normal **Observable** is great
 ## Controlling the flow
 ```js
 // typing "hello world"
-var input = Rx.Observable.fromEvent(document.querySelector('input'), 'keyup');
+var input = Rx.Observable.fromEvent(document.querySelector('input'), 'input');
 
 // Filter out target values less than 3 characters long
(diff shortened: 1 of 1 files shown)


blesh:
@Gregoirevda actually, I think this should probably be `input` if we want to be "modern" about this. It'll have better behavior than `keyup` because it won't register arrow keys, etc. Works in IE9+.

doc/tutorial/basics.md
```diff
@@ -48,7 +48,7 @@ Which one you choose depends on the scenario. The normal **Observable** is great
 ## Controlling the flow
 ```js
 // typing "hello world"
-var input = Rx.Observable.fromEvent(document.querySelector('input'), 'keypress');
+var input = Rx.Observable.fromEvent(document.querySelector('input'), 'keyup');
```
blesh: use `var input = Rx.Observable.fromEvent(document.querySelector('input'), 'input');` instead. (Same for the other line)

Gregoirevda: Ok, i changed it to input

blesh:
👍  thanks, @Gregoirevda

Issue #2229, opened 2016-12-30
Run browser testing via global builds


kwonoj:
**Additional information:**

Currently all of testing even including browser test uses CJS build package. It'd be more ideal if browser testing uses global build package instead, to detect possible regressions on global build pipeline.

kwonoj:
Related PR to enable test pipeline: https://github.com/ReactiveX/rxjs/pull/2225 / https://github.com/ReactiveX/rxjs/pull/2227

benlesh:
This is old. We tried browser testing, and it was too flaky so we stopped.

PR #2227, opened 2016-12-28
Simplify mocha runner configuration

kwonoj:
**Description:**

⚠️ Targeting master ⚠️

This PR simplifies some of mocha runner configurations for further test pipeline refactoring.
Does not have breaking changes, targeting master.

**Related issue (if exists):**

#2225

PR #2243, opened 2017-01-02
build(global): enable aggressive closure optimization

kwonoj:
**Description:**
This PR enables 'ADVANCED' optimization in closure compiler to generate more compressed minified global build. 

Size comparsion:
image

gain around ~30kb of more compression.

I'm marking it as **BLOCKED** - cause there isn't reliable way to validate regressions of this optimization. I have filed related issue https://github.com/ReactiveX/rxjs/issues/2229, which'll allow to test global builds correctly on browsers.

https://github.com/ReactiveX/rxjs/issues/2229
https://github.com/ReactiveX/rxjs/pull/2227
https://github.com/ReactiveX/rxjs/pull/2225

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,7 +160,7 @@
     "ghooks": "^1.3.2",
     "glob": "^7.0.3",
     "gm": "^1.22.0",
-    "google-closure-compiler-js": "^20160916.0.0",
+    "google-closure-compiler-js": "^20161201.0.0",
     "gzip-size": "^3.0.0",
     "http-server": "^0.9.0",
     "lint-staged": "^3.1.0",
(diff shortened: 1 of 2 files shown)


trxcllnt:
nice looking forward to this!

blesh:
@kwonoj do you think we could validate regressions by somehow building and running the specs against the this optimized build? This seems like something we should prioritize.

kwonoj:
As noted in PR, I already filed https://github.com/ReactiveX/rxjs/issues/2229 to change browser testing's target to use UMD build instead of CJS. I gave a quick shot and found it wasn't trivial though.

kwonoj:
I'm going to close this for now and reopen once we have pipeline for testing against browser.

Issue #2244, opened 2017-01-02
Memory Leak with Queue Scheduler and Subject


MikeRyan52:
**RxJS version:** 5.0.2

**Code to reproduce:**
Observing a subject using the `queue` scheduler creates many `QueueAction`s that are never freed.

Plnkr demonstrating issue: https://plnkr.co/edit/BBkHlpIf6MiP2eOPdIlz

Minimal code to reproduce:
```ts
import 'rxjs/add/operator/observeOn';
import { interval } from 'rxjs/observable/interval';
import { BehaviorSubject } from 'rxjs/BehaviorSubject';
import { queue } from 'rxjs/scheduler/queue';

const source$ = new BehaviorSubject(0);

source$.observeOn(queue).subscribe();

interval(50).subscribe(source$);
```

Create two heap allocation snapshots about 20 seconds apart and use the comparison tools to see the allocation delta. Example:

![Heap Allocation Snapshot Example](http://i.imgur.com/Mh3q5Q1.png)

**Expected behavior:**
No memory leak of previous values.

**Actual behavior:**
Previous values are retained in `QueueAction`s

kwonoj:
So the issue seems occurring on `observeOn` operator by holding observed notification's subscription. Please check https://github.com/ReactiveX/rxjs/pull/2245 for reference, local verification shows leaks are gone with given changes.

MikeRyan52:
@kwonoj Pulled down the latest from your branch and I am still seeing a leak of `QueueAction`s. Also seeing a leak of instances of `ObserveOnContext`. All are being held by `ObserveOnSubscriber._subscriptions`

kwonoj:
@MikeRyan52 can you share setup you've used to repro issue (maybe github repo)?

MikeRyan52:
@kwonoj I created a demo repository here that reproduces the problem with both the latest published release of rxjs and your branch: https://github.com/MikeRyan52/rxjs-memory-leak

blesh:
I see the problem right now and I'm working on a solution.

blesh:
@MikeRyan52 try the PR #2248 ... it should solve your problem.

Thank you so much for reporting this one.

blesh:
@MikeRyan52 the patch is published. 5.0.3

JamesHenry:
Thanks @blesh and @MikeRyan52!

@blesh, just FYI the latest on npm appears to still be `5.0.2`

blesh:
[...]

PR #2245, opened 2017-01-02
fix(observeOn): remove observed subscription to clean up notification…

kwonoj:
… instance

**Description:**
This PR updates internals of `observerOn` to remove subscription once scheduled notification is observed. Wittout those, instance of `Notification` and related `Action` are hold by operator subscription and won't be freed. This is targeting master, as I believe there is no breaking changes.

- relates to #2244


kwonoj:
closing PR for now as this seems not legit fix and proper fix will follow.

blesh:
FYI: Fix PR is here #2248

PR #2249, opened 2017-01-04
fix(ISubscription): update type definition of ISubscription::closed

kwonoj:
**Description:**

Originated from https://github.com/ReactiveX/rxjs/pull/2234, doesn't add any code changes but only update type definition for interface clarity. 

Currently it's not so benefitable even for TS user since all of codebase just return `Subscription` everywhere instead of using interfaces, that need to be dealt with separate refactoring.

#2234

### DIFF
--- a/src/Subscription.ts
+++ b/src/Subscription.ts
@@ -13,7 +13,7 @@ export type TeardownLogic = AnonymousSubscription | Function | void;
 
 export interface ISubscription extends AnonymousSubscription {
   unsubscribe(): void;
-  closed: boolean;
+  readonly closed: boolean;
 }
 
 /**


src/Subscription.ts
```diff
@@ -13,7 +13,7 @@ export type TeardownLogic = AnonymousSubscription | Function | void;
 
 export interface ISubscription extends AnonymousSubscription {
   unsubscribe(): void;
-  closed: boolean;
+  readonly closed: boolean;
```
david-driscoll: 👍 this will encourage TypeScript users to not try to modify this value (as it shouldn't be mutable)

kwonoj: @david-driscoll check #2234 as well, this PR is subset of it only updated type definition, while it does updates actual code behaviors.

blesh:
> since all of codebase just return Subscription everywhere instead of using interfaces, that need to be dealt with separate refactoring.

Technically, a LOT of the codebase returns subclasses of `Subscriber`. So people *could* call `next` on them if they chose too.. however, that's not supported nor is it documented.

I guess what I'm saying is: Hold off on that refactor, there's likely one coming anyhow... especially with the planned scheduling changes.

jayphelps:
LGTM

kwonoj:
LGTM

Issue #2198, opened 2016-12-16
Record perf results vs RxJS 4, then drop RxJS 4 from perf testing


benlesh:
It's time to move on. Future versions of this library need to be compared against previous versions of itself, not Rx4.

We need to

1. Record and keep the perf results vs RxJS 4, if for no other reason than documentation purposes.
2. Fix the perf tests to target a previous version of RxJS 5.

blesh:
cc @kwonoj @jayphelps

manekinekko:
Hi @blesh are you still planning to use Benchpress for future perf tests? I'm actually building a tool called Perfy, on top of Benchpress which allows to keep track of previous tests campaign, which is exactly what you need. The tool lives here :https://github.com/manekinekko/perfy.

If this is something you wanna try out, I can prepare a POC for RxJS.

blesh:
@manekinekko that sounds awesome. I'd love to see it in action. I have a feeling our benchpress tests have probably fallen into disrepair. 

Do note, though, that we have both micro perf and macro perf tests.

manekinekko:
alright. I'll begin working on that just right after christmas 🎅

manekinekko:
@blesh here is a quick draft... Please tell if this is what you had in mind?
![image](https://cloud.githubusercontent.com/assets/1699357/21467873/3324c2b4-c9fc-11e6-9468-542bb09d3b9a.png)

kwonoj:
@manekinekko I'm curious if we could create github webhooks to display some critical info per each PR (let's not think about ergonomics like where to run server, etcs for now)

manekinekko:
@kwonoj yep I think it's doable. 

Here are some facts about Perfy:
- It runs after protractor/benchpress has ended executing specs tests.
- It overrides benchpress config in order to make benchpress outputs JSON files.
- It reads those JSON files and merges them into a single JSON file which serves as a database.
- It has a CLI interface.

Here is the complete workflow:

`npm run tests --> protractor --> benchpress --> perfy --> charts`

One thing I'm missing here is how we'll be able to provide a subset of spec tests to benchpress. Typically, we will need to:
1) run only the tests that are affected by the specific PR that is being created. 
2) Or do we need to run the whole tests? 

I think 2) would be a good approach 🤔

claudiordgz:
[...]

PR #2266, opened 2017-01-08
feat(Zone): add support for Zone.js

mhevery:
Add support for Zone.js in a way which should have minimal impact on performance.

Ensures that all callbacks are invoked in the same `Zone` as the `Zone` which was current at
the time of the callback registration with the `Rx`.

-  Add `getZone()` method which returs always `null` if no `Zone` is present or returrns
   the current zone.
 - All places where the `Zone` needs to be intered are quarded to only go through zone if the
   Zone is different from the current `Zone` for performance reasons.

### DIFF
--- a/src/Subscriber.ts
+++ b/src/Subscriber.ts
@@ -1,4 +1,5 @@
 import { isFunction } from './util/isFunction';
+import {getZone} from './util/getZone';
 import { Observer, PartialObserver } from './Observer';
 import { Subscription } from './Subscription';
 import { empty as emptyObserver } from './Observer';
@@ -92,7 +93,13 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
    */
   next(value?: T): void {
     if (!this.isStopped) {
-      this._next(value);
+      if (this._zone && this._zone != getZone()) {
+        // Current Zone is different from the intended zone.
+        // Restore the zone before `next`ing.
+        this._zone.run(this._next, this, [value]);
+      } else {
+        this._next(value);
+      }
     }
   }
 
@@ -106,7 +113,13 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
   error(err?: any): void {
     if (!this.isStopped) {
       this.isStopped = true;
-      this._error(err);
+      if (this._zone && this._zone != getZone()) {
+        // Current Zone is different from the intended zone.
+        // Restore the zone before `error`ing.
+        this._zone.run(this._error, this, [err]);
+      } else {
+        this._error(err);
+      }
     }
   }
 
@@ -119,7 +132,13 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
   complete(): void {
     if (!this.isStopped) {
       this.isStopped = true;
-      this._complete();
+      if (this._zone && this._zone != getZone()) {
+        // Current Zone is different from the intended zone.
+        // Restore the zone before `complete`ing.
+        this._zone.run(this._complete, this);
+      } else {
+        this._complete();
+      }
     }
   }
 
--- a/package.json
+++ b/package.json
@@ -72,10 +72,10 @@
     "copy_src_cjs": "mkdirp ./dist/cjs/src && shx cp -r ./src/* ./dist/cjs/src",
     "copy_src_es6": "mkdirp ./dist/es6/src && shx cp -r ./src/* ./dist/es6/src",
     "commit": "git-cz",
-    "compile_dist_cjs": "tsc ./dist/cjs/src/Rx.ts ./dist/cjs/src/add/observable/of.ts                         -m commonjs --lib es5,es2015.iterable,es2015.collection,es2015.promise,dom --sourceMap --outDir ./dist/cjs --target ES5 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --noImplicitThis --suppressImplicitAnyIndexErrors --moduleResolution node",
-    "compile_module_es6": "tsc                                          ./dist/es6/src/Rx.ts ./dist/es6/src/add/observable/of.ts                             -m es2015   --sourceMap --outDir ./dist/es6 --target ES5 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --noImplicitThis --suppressImplicitAnyIndexErrors --moduleResolution node --noEmitHelpers --lib es5,es2015.iterable,es2015.collection,es2015.promise,dom ",
-    "compile_dist_es6_for_docs": "tsc                                 ./dist/es6/src/Rx.ts ./dist/es6/src/add/observable/of.ts ./dist/es6/src/MiscJSDoc.ts -m es2015   --sourceMap --outDir ./dist/es6 --target ES6 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --noImplicitThis --suppressImplicitAnyIndexErrors --moduleResolution node",
-    "cover": "shx rm -rf dist/cjs && tsc src/Rx.ts src/add/observable/of.ts -m commonjs --lib es5,es2015.iterable,es2015.collection,es2015.promise,dom --outDir dist/cjs --sourceMap --target ES5 -d && nyc --reporter=lcov --reporter=html --exclude=spec/support/**/* --exclude=spec-js/**/* --exclude=node_modules mocha --opts spec/support/default.opts spec-js",
+    "compile_dist_cjs": "tsc ./dist/cjs/src/Rx.ts ./dist/cjs/src/add/observable/of.ts ./node_modules/zone.js/dist/zone.js.d.ts                         -m commonjs --lib es5,es2015.iterable,es2015.collection,es2015.promise,dom --sourceMap --outDir ./dist/cjs --target ES5 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --noImplicitThis --suppressImplicitAnyIndexErrors --moduleResolution node",
+    "compile_module_es6": "tsc                                          ./dist/es6/src/Rx.ts ./dist/es6/src/add/observable/of.ts ./node_modules/zone.js/dist/zone.js.d.ts                             -m es2015   --sourceMap --outDir ./dist/es6 --target ES5 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --noImplicitThis --suppressImplicitAnyIndexErrors --moduleResolution node --noEmitHelpers --lib es5,es2015.iterable,es2015.collection,es2015.promise,dom ",
+    "compile_dist_es6_for_docs": "tsc                                 ./dist/es6/src/Rx.ts ./dist/es6/src/add/observable/of.ts ./node_modules/zone.js/dist/zone.js.d.ts ./dist/es6/src/MiscJSDoc.ts -m es2015   --sourceMap --outDir ./dist/es6 --target ES6 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --noImplicitThis --suppressImplicitAnyIndexErrors --moduleResolution node",
+    "cover": "shx rm -rf dist/cjs && tsc src/Rx.ts src/add/observable/of.ts ./node_modules/zone.js/dist/zone.js.d.ts -m commonjs --lib es5,es2015.iterable,es2015.collection,es2015.promise,dom --outDir dist/cjs --sourceMap --target ES5 -d && nyc --reporter=lcov --reporter=html --exclude=spec/support/**/* --exclude=spec-js/**/* --exclude=node_modules mocha --opts spec/support/default.opts spec-js",
     "decision_tree_widget": "cd doc/decision-tree-widget && npm run build && cd ../..",
     "doctoc": "doctoc CONTRIBUTING.md",
     "generate_packages": "node .make-packages.js",
@@ -192,7 +192,8 @@
     "watch": "^1.0.1",
     "watchify": "3.7.0",
     "webpack": "^1.13.1",
-    "xmlhttprequest": "1.8.0"
+    "xmlhttprequest": "1.8.0",
+    "zone.js": "^0.7.4"
   },
   "engines": {
     "npm": ">=2.0.0"
--- a/spec/Zone-spec.ts
+++ b/spec/Zone-spec.ts
@@ -1,0 +1,139 @@
+import {expect} from 'chai';
+import 'zone.js';
+import {Subscriber, Observable} from '../dist/cjs/Rx';
+
+/**
+ * The point of these tests, is to ensure that all callbacks execute in the Zone which was active
+ * when the callback was passed into the Rx.
+ *
+ * The implications are:
+ * - Observable callback passed into `Observable` executes in the same Zone as when the
+ *   `new Observable` was invoked.
+ * - The subscription callbacks passed into `subscribe` execute in the same Zone as when the
+ *   `subscribe` method was invoked.
+ * - The operator callbacks passe into `map`, etc..., execute in the same Zone as when the
+ *   `operator` (`lift`) method was invoked.
+ */
+describe('Zone interaction', () => {
+  it('should run methods in the zone of declaration', () => {
+    const log: string[] = [];
+    const constructorZone: Zone = Zone.current.fork({ name: 'Constructor Zone'});
+    const subscriptionZone: Zone = Zone.current.fork({ name: 'Subscription Zone'});
+    let subscriber: Subscriber<string> = null;
+    const observable = constructorZone.run(() => new Observable<string>((_subscriber) => {
+      subscriber = _subscriber;
+      log.push('setup');
+      expect(Zone.current.name).to.eq(constructorZone.name);
+      return () => {
+        expect(Zone.current.name).to.eq(constructorZone.name);
+        log.push('cleanup');
+      };
+    })) as Observable<string>;
+    subscriptionZone.run(() => observable.subscribe(
+        () => {
+          expect(Zone.current.name).to.eq(subscriptionZone.name);
+          log.push('next');
+        },
+        () => null,
+        () => {
+          expect(Zone.current.name).to.eq(subscriptionZone.name);
+          log.push('complete');
+        }
+    ));
+    subscriber.next('MyValue');
+    subscriber.complete();
+
+    expect(log).to.deep.equal(['setup', 'next', 'complete', 'cleanup']);
+    log.length = 0;
+
+    subscriptionZone.run(() => observable.subscribe(
+        () => null,
+        () => {
+          expect(Zone.current.name).to.eq(subscriptionZone.name);
+          log.push('error');
+        },
+        () => null
+    ));
+    subscriber.next('MyValue');
+    subscriber.error('MyError');
+
+    expect(log).to.deep.equal(['setup', 'error', 'cleanup']);
+  });
+
+  it('should run methods in the zone of declaration when nexting synchronously', () => {
+    const log: string[] = [];
+    const rootZone: Zone = Zone.current;
+    const constructorZone: Zone = Zone.current.fork({ name: 'Constructor Zone'});
+    const subscriptionZone: Zone = Zone.current.fork({ name: 'Subscription Zone'});
+    const observable = constructorZone.run(() => new Observable<string>((subscriber) => {
+      // Execute the `next`/`complete` in different zone, and assert that correct zone
+      // is restored.
+      rootZone.run(() => {
+        subscriber.next('MyValue');
+        subscriber.complete();
+      });
+      return () => {
+        expect(Zone.current.name).to.eq(constructorZone.name);
+        log.push('cleanup');
+      };
+    })) as Observable<string>;
+
+    subscriptionZone.run(() => observable.subscribe(
+        () => {
+          expect(Zone.current.name).to.eq(subscriptionZone.name);
+          log.push('next');
+        },
+        () => null,
+        () => {
+          expect(Zone.current.name).to.eq(subscriptionZone.name);
+          log.push('complete');
+        }
+    ));
+
+    expect(log).to.deep.equal(['next', 'complete', 'cleanup']);
+  });
+
+  it('should run operators in the zone of declaration', () => {
+    const log: string[] = [];
+    const rootZone: Zone = Zone.current;
+    const constructorZone: Zone = Zone.current.fork({ name: 'Constructor Zone'});
+    const operatorZone: Zone = Zone.current.fork({ name: 'Operator Zone'});
+    const subscriptionZone: Zone = Zone.current.fork({ name: 'Subscription Zone'});
+    let observable = constructorZone.run(() => new Observable<string>((subscriber) => {
+      // Execute the `next`/`complete` in different zone, and assert that correct zone
+      // is restored.
+      rootZone.run(() => {
+        subscriber.next('MyValue');
+        subscriber.complete();
+      });
+      return () => {
+        expect(Zone.current.name).to.eq(constructorZone.name);
+        log.push('cleanup');
+      };
+    })) as Observable<string>;
+
+    observable = operatorZone.run(() => observable.map((value) => {
+      expect(Zone.current.name).to.eq(operatorZone.name);
+      log.push('map: ' + value);
+      return value;
+    })) as Observable<string>;
+
+    subscriptionZone.run(() => observable.subscribe(
+        () => {
+          expect(Zone.current.name).to.eq(subscriptionZone.name);
+          log.push('next');
+        },
+        (e) => {
+          expect(Zone.current.name).to.eq(subscriptionZone.name);
+          log.push('error: ' + e);
+        },
+        () => {
+          expect(Zone.current.name).to.eq(subscriptionZone.name);
+          log.push('complete');
+        }
+    ));
+
+    expect(log).to.deep.equal(['map: MyValue', 'next', 'complete', 'cleanup']);
+  });
+
+});
--- a/spec/support/default.opts
+++ b/spec/support/default.opts
@@ -1,3 +1,4 @@
+--require zone.js
 --require source-map-support/register
 --require spec-js/helpers/testScheduler-ui.js
 --ui spec-js/helpers/testScheduler-ui.js
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -4,6 +4,7 @@ import { Subscriber } from './Subscriber';
 import { Subscription, AnonymousSubscription, TeardownLogic } from './Subscription';
 import { root } from './util/root';
 import { toSubscriber } from './util/toSubscriber';
+import { getZone } from './util/getZone';
 import { IfObservable } from './observable/IfObservable';
 import { ErrorObservable } from './observable/ErrorObservable';
 import { $$observable } from './symbol/observable';
@@ -29,8 +30,17 @@ export class Observable<T> implements Subscribable<T> {
 
   protected source: Observable<any>;
   protected operator: Operator<any, T>;
+  /**
+   * The `Zone` which was captured at the time `Observable` got created.
+   * This is the `Zone` which will be used when invoking the `Observable` callbacks.
+   */
+  private _zone: Zone;
 
   /**
+   * Note about `zone.js`: When `zone.js` is loaded the `Observable` will capture the `Zone` on
+   * construction and then ensure that the `subscribe` function as well as the `TeardowLogic`
+   * execute in the `Zone` which was current an the time of the constructor call.
+   *
    * @constructor
    * @param {Function} subscribe the function that is  called when the Observable is
    * initially subscribed to. This function is given a Subscriber, to which new values
@@ -38,6 +48,7 @@ export class Observable<T> implements Subscribable<T> {
    * `complete` can be called to notify of a successful completion.
    */
   constructor(subscribe?: <R>(this: Observable<T>, subscriber: Subscriber<R>) => TeardownLogic) {
+    this._zone = getZone();
     if (subscribe) {
       this._subscribe = subscribe;
     }
@@ -60,6 +71,10 @@ export class Observable<T> implements Subscribable<T> {
   /**
    * Creates a new Observable, with this Observable as the source, and the passed
    * operator defined as the new observable's operator.
+   *
+   * Note about `zone.js`: When `zone.js` is loaded the all operator callback function will execute
+   * in the `Zone` which was current when the operator was registered using `lift` method.
+   *
    * @method lift
    * @param {Operator} operator the operator defining the operation to take on the observable
    * @return {Observable} a new observable with the Operator applied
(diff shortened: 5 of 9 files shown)


kwonoj:
Just curiosity - without this PR, how does RxJS works with `zone.js`? is there functional differences, or only impact perf?

jayphelps:
@mhevery this is a fairly fundamental change. I imagine not all rxjs members are familiar with Zones. Could you perhaps summarize Zones as well as the pros/cons to adding native support vs. the existing interactions with Zone's polyfills?

mhevery:
@kwonoj 
> Just curiosity - without this PR, how does RxJS works with zone.js? is there functional differences, or only impact perf?

Without this PR the Zone's are incorrectly propagated through RX. So this is a correctness PR with focus on performance. See [ZoneSpec](https://github.com/ReactiveX/rxjs/pull/2266/files#diff-e7fb16d246f015bd63a0031750582e34) to see how the PR changes the Zone propagation behavior.

RX is one of the few libraries which falls into the category of "User Queueing" Libraries. What this means is that RX retains multiple callbacks and then executes them at a later time in a single task. Without this PR all of the callbacks would execute in the same Zone, which is incorrect.

mhevery:
@jayphelps 

> @mhevery this is a fairly fundamental change. I imagine not all rxjs members are familiar with Zones. Could you perhaps summarize Zones as well as the pros/cons to adding native support vs. the existing interactions with Zone's polyfills?

You can read up on Zones in [The Zone Primer](https://docs.google.com/document/d/1F5Ug0jcrm031vhSMJEOgp1l-Is-Vf0UCNDY-LsQtAIY). As of now when Rx runs in the Zone environment it will propagate Zones according to VM Tasks. In the case of Rx this is not correct behavior because Rx does its own scheduling and the result is that all callbacks end up running in the same zone as Task zone. This behavior is mostly correct most of the time, but there are cases which will have wrong behavior. This PR addresses those. This PR is needed to enable propre implementation of https://github.com/angular/angular/issues/13248

This PR was authored with @blesh help, and so I think he may be one of the people who should have sufficient context to review it. I am happy to work with anyone else and explain all of the nuances and reasoning.

mhevery:
Push new SHA which adds more comments, and simplifies the code a bit.

blesh:
It's probably a good idea to add a good amount of documentation around what  RxJSs behavior is with Zones and why. 

Also perhaps some examples of using RxJS with Zones. I think that this will be interesting when it comes to @jhusain's proposed changes to observable wrt error handling. 

Overall looks good, but needs docs for sure. Otherwise it will appear to be magic to some Zones users I'm sure.

trxcllnt:
I'm not 100% up to speed on zones, but this does seem like something that should go into a zone scheduler, rather than baked into every Subscriber (see also [Rx Guidelines 5.8 on synchronize](https://go.microsoft.com/fwlink/?LinkID=205219) [pdf]).

Also, what's the impact on users (like me) who will never use zones?

mhevery:
@blesh I have updated the documentation, please let me know if it is sufficient.

mhevery:
@trxcllnt 
> I'm not 100% up to speed on zones, but this does seem like something that should go into a zone scheduler, rather than baked into every Subscriber (see also Rx Guidelines 5.8 on synchronize [pdf]).

The goal of this PR is to ensure that any application callbacks which are passed into the RX run in the same `Zone` as was active when the callback was passed in. This is too fine grained for the scheduler, since fully synchronous Rx may never use a scheduler, but still must abide by the rule above.

As an example the Observable must ensure that the `subscriber` callback and teardown must run in the same zone which was active at the time of constructor.

```
const constructorZone = Zone.current;
const observable = new Observable<string>((subscriber) => {
  expect(Zone.current).to.eq(constructorZone);
  return () => {
    expect(Zone.current).to.eq(constructorZone);
  };
}));
```

Same rules apply to operators, as well as to `subscribe` callbacks. These callbacks are too fine grained to be handled by the scheduler.


> Also, what's the impact on users (like me) who will never use zones?

It should not have any impact whatsoever. The Zone dependency is optional, and if there is no zone present the code should execute at the same speed as before. The semantics are always the same, the only difference is if the `Zone`s are preserved.

mhevery:
@blesh 
> I think that this will be interesting when it comes to @jhusain's proposed changes to observable wrt error handling.

This PR does not change the semantics of the Error handling only ensures that a correct `Zone` is propagated. Therefore any such changes should have no effect on this.

trxcllnt:
@mhevery looking at the tests, it seems like something that Angular could compose by returning all their Observables with `.subscribeOn(CurrentZoneScheduler).observeOn(CurrentZoneScheduler)`, where the scheduler checks the Subscriber's zone against `Zone.current`, and switches as appropriate. The try-catch runner seems more inaccessible, but could be resolved by the work to remove that function entirely in favor of running in the context of a global try/catch context, which @blesh has more details about.

I'm not against supporting zones, but would like to find a way that doesn't add extra debugger step commands for people who aren't using zones. Thoughts?

ladyleet:
👍

mhevery:
@trxcllnt 

1. I don't mind Angular doing more work, but I don't think what you are proposing would work.
2. But more importantly our developers can not be required to use Rx in that way. It is one thing for a framework to do extra work, but something harder for the developers. (Also our developers could chose to use a third part Rx library which is not Angular or `Zone` aware and the code should still work.)

The reason it would not work is imagine this situation.

```
new Observable(cb1).map(cb2).subscribe(cb3)
```
The above has three callbacks. It is possible that each callback will run in a separate `Zone` so there are three `Zone`s involved. The above code has only one scheduler. Hence one scheduler can not be associated with three callbacks.

As I said before the scheduler is too coarse grained for what is needed. Or to think about it differently the scheduler and `Zone`s are orthogonal. The kind of scheduler you choose does not have any influence on the way `Zone`s are propagated. It is possible that a developer would implement and install a custom scheduler, and we still need to have the `Zone`s propagated.

blesh:
> would like to find a way that doesn't add extra debugger step commands for people who aren't using zones.

I'll need to look at the call stacks, because that sort of visibility is an important goal for Netflix.

@trxcllnt Zones support is tricky because there might be discrepancies between the zone the observable is in when it's created, versus when it's subscribed, versus the zone the actual thing producing data is running in. Allowing nature to take it's course with the way observables are currently might lead to seemingly non-deterministic behavior with zones.

@mhevery I must admit though, I'm not entirely sold on the way it's broken down here though... for example, given this:

```js
let observable;

zone1.run(() => observable = new Observable(observer => {
  console.log('subscribing', Zone.current.name);
  observer.next('hi');
  observer.complete();
});

zone2.run(() => observable = observable.map(x => {
  console.log('map', Zone.current.name);
  return x + '!';
});

zone3.run(() => observable.subscribe(x => {
  console.log('next', Zone.current.name);
  console.log('next', x);
});
```

I would expect this:

```js
subscribing zone3
map zone3
next zone3
next hi!
```

Because really those things are just functions:

```js
let fn;

zone1.run(() => fn = observer => {
  console.log('subscribing', Zone.current.name);
  observer.next('hi');
  observer.complete();
});

zone2.run(() => fn = observer => fn({
  next(x) { 
    console.log('map', Zone.current.name);
    observer.next(x + '!');
  }
});

zone3.run(() => fn({
  next(x) {
    console.log('next', Zone.current.name);
    console.log('next', x);
  }
});
```

It seems like Zones might have this issue with any function that is created under a zone that might be called at a later time. Is that accurate?

blesh:
@mhevery ... I guess what I'm saying is: Are we sure that the "correct zone" for an Observable's ctor callback is the zone it was created in? Is that the standard behavior for all functions in JavaScript WRT to Zones?

trxcllnt:
> @blesh It seems like Zones might have this issue with any function that is created under a zone that might be called at a later time.

Yes, this was my first instinct too. It seems like this would be an issue for nearly every library. Is jQuery going to have to ship with zone support as well?

> @mhevery Also our developers could chose to use a third part Rx library which is not Angular or Zone aware and the code should still work.

Does this mean someone could drop in most.js in place of Rx, and zones should still work? If so, and considering Rx has to bake in zone support, how would zones still work if most.js doesn't do the same?

Exploring alternatives a bit more, is there a reason Angular can't extend (or worst case, monkey-patch) Observable with a custom `lift` implementation that creates zone-aware subscribers? Something like this:

```es6
class ZoneObservable extends Observable {
  lift(operator) {
    const obs = new ZoneObservable();
    obs.source = this;
    obs.operator = new ZoneOperator(operator);
    return obs;
  }
}
class ZoneOperator {
  constructor(op) { this.op = op; }
  call(subscriber, source) {
    return this.op.call(new ZoneSubscriber(subscriber), source);
  }
}
```

blesh:
@trxcllnt I think there is a larger issue here than just Angular, as Zones are on the TC39's docket. 

Looking at this now: https://docs.google.com/document/d/1F5Ug0jcrm031vhSMJEOgp1l-Is-Vf0UCNDY-LsQtAIY/edit#heading=h.eijd9xb9zfqx

It seems like the idea is that since using a zone is _explicit_, then it can be assumed that if someone adds a mapping function to an observable within a zone, their intention is that that function should run in that zone. Is that correct, @mhevery? In other words, if I've gone out of my way to create an Observable in some `zone.run()` then I must _really_ want the body of it to run in that zone, etc. etc.

mhevery:
> @blesh  ... I guess what I'm saying is: Are we sure that the "correct zone" for an Observable's ctor callback is the zone it was created in? Is that the standard behavior for all functions in JavaScript WRT to Zones?

Yes we are sure. I know it is not immediately obvious, so let me try to explain through an example.

Imagine you have an observable on `keydown` listener and you would like to debounce it before you pass it onto the subscriber.  In our case the subscriber would be an Angular application which should re-render. 

It is clear that the subscriber should run in Angular zone so that angular does proper rendering. But if we run the observable registration of `keydown` in Angular zone, then we will be rendering on each key stroke which is too much since we have debounce. The `keydown` events happen way more often then the subscriber callback, and we want to apply different rules to them.

```
var keydown$
ngZone.runOutside(() => {
  keydown$ = Observable.fromEvent(element, 'keydown').debounce(200);
});
ngZone.runInside(() => {
  keydown$.subscribe(...)
});
```

In the above example what we set up is Rx recipe. Part of the recipe is outside of Angular and part is in Angular. This way we can control which operations have side effects and which do not. 



> @trxcllnt  Yes, this was my first instinct too. It seems like this would be an issue for nearly every library. Is jQuery going to have to ship with zone support as well?

Good question! No most libraries do not need anything special to work. Some such as Rx, and Promise do. So the obvious question is Why? What is special about Rx and Promise. These libraries do something called user-queueing. This issue arrises only when the library takes callbacks from many different clients, stores these callbacks and then batch executes them in a single VM turn. jQuery does not do that. jQuery passes its callbacks to underling platform `addEventListener` and it does not batch process them, so there is no issue. Rx is different, it composes many callbacks into chains and then executes them together, taking the output of one and passing to next.

What we are interested in is to be able to know which callback comes from app, framework, or outside the app (such as another app or an ad) So Rx needs to do more work, where as most libraries do not.

This is why the current state sort of works. We correctly retain the Zone for the VM turn, but that implies that all callbacks in that VM turn will have the same Zone. This prevents the debounce example shown above.





> @trxcllnt Does this mean someone could drop in most.js in place of Rx, and zones should still work? If so, and considering Rx has to bake in zone support, how would zones still work if most.js doesn't do the same?

Sorry, that is not what I meant. If Angular would support most.js, then we would have to teach most.js how to propagate Zones as well (see above example of user-queueing). What I was getting at is imagine you have a third-party library such as RxFirebase, which was written on top of Rx but without the knowledge of Angular or Zones. An Angular developer should be able just pull that library in and use it in an Angular app.  Therefore, requiring that RxFirebase use a special calling convention which was suggested in original comment would break the Zone support and would make Angular and exclusive club. We want you to be able to use any library wether or not it was written with Angular (or Zones) in mind.




> @trxcllnt Exploring alternatives a bit more, is there a reason Angular can't extend (or worst case, monkey-patch) Observable with a custom lift implementation that creates zone-aware subscribers? Something like this:

That would solve the operator case, but not the subscribe callback nor creation callback case.



> @blesh It seems like the idea is that since using a zone is explicit, then it can be assumed that if someone adds a mapping function to an observable within a zone, their intention is that that function should run in that zone. Is that correct, @mhevery? In other words, if I've gone out of my way to create an Observable in some zone.run() then I must really want the body of it to run in that zone, etc. etc.

Yes, that is correct.

trxcllnt:
> @mhevery Some such as Rx, and Promise do. So the obvious question is Why? What is special about Rx and Promise. These libraries do something called user-queueing. 

> jQuery passes its callbacks to underling platform addEventListener and it does not batch process them, so there is no issue. Rx is different, it composes many callbacks into chains and then executes them together, taking the output of one and passing to next.

I see. Wouldn't jQuery Deferred's would still need to include Zone support, since they execute user-code asynchronously, like Promises and Observables. Overall, it sounds like the Zones approach is to monkey-patch all the platform globals that can execute user-code asynchronously to support Zones. Is that accurate?

> @mhevery If Angular would support most.js, then we would have to teach most.js how to propagate Zones as well (see above example of user-queueing). What I was getting at is imagine you have a third-party library such as RxFirebase, which was written on top of Rx but without the knowledge of Angular or Zones.

Thanks for clarifying. Based on this knowledge, I have some further questions (I appreciate you bearing with me here):

1. How do Zones interact with libraries that allow function currying/partial application, like lodash or ramda? Conceivably, one could declare a partially applied function in one Zone, curry some arguments (effectively redefining the function) in a different zone, and then execute the curried function in yet another zone. Which Zone does the original function run in? Do lodash and ramda have to build in zone support (or be monkey-patched themselves) to exhibit the correct behavior?
2. Are there any benefits to Zones in Rx core if I'm not using Angular 2? If I'm using Rx and React on the client, or running a high performance router and GPU render loop with Rx and websockets, what's this PR buy me?

Without knowing the potential benefits, I only see one big negative: at least one extra conditional per onNext and try/catch (something we're trying not to do). Regardless of whether the VM inlines the result of `getZone()`, the condition still results in 2x the number of debugger step commands I'll have to issue when I'm stepping through an operator chain.

> @mhevery That would solve the operator case, but not the subscribe callback nor creation callback case.

Ok, so it sounds like monkey-patching Observable is an option for Angular. If so, we can extend the monkey-patching example to have logic to run subscribe functions in the proper zone context:
```es6
class ZoneObservable extends Observable {
  constructor(subscribe) {
    super(Zone.current.runInContext(subscribe)); // <-- replace `runInContext` with whatever Zone method is appropriate here
  }
  /* ... the rest of the example ...*/
}
```

blesh:
@mhevery in talking with @jhusain about this, we determined that the zone should probably be attached to the subscription. In other words, whatever zone the subscribe call was in is the zone the other functions passed to the observable/map/filter/whatever should be run in. 

I don't think that's how this PR work though. Am I mistaken?

mhevery:
> @trxcllnt Wouldn't jQuery Deferred's would still need to include Zone support, since they execute user-code asynchronously, like Promises and Observables

Yes any implementation of Promise or Rx would need this. But the rest of the jQuery would not need it.


> @trxcllnt How do Zones interact with libraries that allow function currying/partial application, like lodash or ramda?

It should have not effect since lodash does not store the callback for later execution. It executes them synchronously. 


> @trxcllnt  Are there any benefits to Zones in Rx core if I'm not using Angular 2?

Yes, you would be able to get long stack traces, and debug information about which operator executed when etc (I will update the PR to add the string information for better debugger support)

You could also get timing information to identify which streams are slow.


> Ok, so it sounds like monkey-patching Observable is an option for Angular.

I don't really see it as an option. It would be very cumbersome. For example we can't patch native Promises and instead we have our own implementations in the Zone. Plus monkey patching has to happen at the right time.

Again having `ZoneObservable` would not work as it means third party libraries such us RxFirebase would not be usable in Angular since they use `new Observable` rather then `ZoneObservable`

I am happy to explore this option more, but given my experience with writing Zone and Promise implementation is think it is very unlikely.

mhevery:
> @blesh: @mhevery in talking with @jhusain about this, we determined that the zone should probably be attached to the subscription. In other words, whatever zone the subscribe call was in is the zone the other functions passed to the observable/map/filter/whatever should be run in.

Having `subscribe` is not the correct behavior. I was alluding to it here: https://github.com/ReactiveX/rxjs/pull/2266#issuecomment-271444203 but let me try again.

Imagine you have a library which creates an observable `obs` which is shared. We then hand `obs` to two consumers `A` and `B`. `A` subscribes to `obs` in `AngularZone` but `B` does not. Now the observable constructor will run either in `AngularZone` or not depending on wether `A` run before `B` or vice versa.

Here is another example. Let's say you have a `timer` observable which counts up periodically. You then add an operator, which only passes on `even` counts, finally one subscribes to the result.

```
var o = rootZone.run(() => new Observable((sub) => {
  var count = 0;
  setInterval(() => sub.next(count++));
} )
.even()
);
angulaZone.run(() => {
 o.subscribe((v) => {
    // do some work expect that angular renders
  })
}
``` 

If the subscription happens in `AngularZone` then you want to render the UI only when data arrives on `subscribe`. You don't want to cause an update on odd values is produced at `setInterval` since the odd value will not propagate to the final subscription.

A common use case is for example autocomplete inbox. Many events happen, but only some of the events should result in angular render. Running the `Observable` constructor callback in the same zone as the `subscription` produces the wrong result.

As a matter of fact the behavior you are asking for is how Rx behaves without support for Zones. Without this PR Rx will execute the `Observable` constructor callback in the same zone as that of the `subscribe` which leads to mostly correct behavior until you want to do things which require sub-selection of data, which then results in running AngularZone too often.

trxcllnt:
> @mhevery It should have not effect since lodash does not store the callback for later execution. It executes them synchronously.

This is true of the Array and Set operators, but I'm talking about methods like [`_.partial`](https://lodash.com/docs#partial), or any method in [Ramda](http://ramdajs.com), which automatically curries. The point I'm getting at is this seems to be a problem inherent with _any_ library that lets you declare a method in one Zone, and execute it in another.

> @mhevery Yes, you would be able to get long stack traces, and debug information about which operator executed when etc (I will update the PR to add the string information for better debugger support)
> 
> You could also get timing information to identify which streams are slow.

Assuming this only works if I compile Zones into my app? What about if I can't, say, because I'm in a high-performance GPU environment with gigabytes of data in memory, so the overhead of a global context that tracks and changes how the platform methods execute is a non-starter?

> @mhevery If the subscription happens in AngularZone then you want to render the UI only when data arrives on subscribe. You don't want to cause an update on odd values is produced at setInterval since the odd value will not propagate to the final subscription.
> 
> A common use case is for example autocomplete inbox. Many events happen, but only some of the events should result in angular render. Running the Observable constructor callback in the same zone as the subscription produces the wrong result.

This has been a solved problem in Rx for years, via Schedulers. Rx.NET synchronizes the events from a throttled stream UI thread scheduler for rendering. In RxJS, we achieve something similar with `noisyStream.auditTime(0, Scheduler.animationFrame).subscribe(render)`.

> @mhevery I don't really see it as an option. It would be very cumbersome. For example we can't patch native Promises and instead we have our own implementations in the Zone. Plus monkey patching has to happen at the right time.

Sure, but I'm sure that was considered when Zones were designed? If zones want to globally track and control execution contexts, surely someone acknowledged the impracticality of updating all 400,000 packages in npm with Zones support? Monkey-patching platform globals gets you most of the way there, and since Zones made the decision they want to own the world, it's their responsibility to monkey-patch the rest (or provide alternative solutions, like a Zone-aware transpiler, etc.).

> @mhevery Again having ZoneObservable would not work as it means third party libraries such us RxFirebase would not be usable in Angular since they use new Observable rather then ZoneObservable

From what I understand about commonjs and ES6 Modules, Angular should be able to replace the Observable export in `rxjs/Rx` and `rxjs/Observable` with one that's aware of Zones, which would solve the RxFirebase issue.

mhevery:
> @trxcllnt This is true of the Array and Set operators, but I'm talking about methods like _.partial, or any method in Ramda, which automatically curries. The point I'm getting at is this seems to be a problem inherent with any library that lets you declare a method in one Zone, and execute it in another.

I had a closer look and I don't see anything there which would require Zone interception. This does not fall under the user-queue umbrella.



> Assuming this only works if I compile Zones into my app? What about if I can't, say, because I'm in a high-performance GPU environment with gigabytes of data in memory, so the overhead of a global context that tracks and changes how the platform methods execute is a non-starter?

Not sure what you mean by compile. If the Zone is loaded than one can create a Zone which will track the invocation. If the Zone is loaded but the code is run in the default Zone, or Zone which does not create objects, than there is zero overhead. All of these features are opt-in. By default it is indistinguishable from no Zone at all. So I don't see an issue here.

> This has been a solved problem in Rx for years, via Schedulers. Rx.NET synchronizes the events from a throttled stream UI thread scheduler for rendering. In RxJS, we achieve something similar with noisyStream.auditTime(0, Scheduler.animationFrame).subscribe(render).

Scheduler and Zones are orthogonal. Its not that you have a ZoneScheduler, it is that you track invocations using zone for any scheduler.

> But surely that was considered when Zones were designed? If zones want to globally track and control execution contexts, surely someone acknowledged the impracticality of updating all 400,000 packages in npm with Zones support? Monkey-patching platform globals gets you most of the way there, and since Zones made the decision they want to own the world, it's their responsibility to monkey-patch the rest (or provide alternative solutions, like a Zone-aware transpiler, etc.).

No one is talking about monkey patching 400,00 libraries. We monkey patched the platform which gets us almost all the way there, with few exception of user queuing libraries. 

If it was possible to monkey patch Rx, we would have already done that, but we can't which is why I would like to have this PR, which adds almost no overhead, and does it the right way. 

> From what I understand about commonjs and ES6 Modules, Angular should be able to replace the Observable export in rxjs/Rx and rxjs/Observable with one that's aware of Zones, which would solve the RxFirebase issue.

Unfortunately that is not possible.

1. Interception the constructor is same as monkey patching the constructor, and that is not sufficient as explained before.
2. Not everyone uses configurable Modules, there is UMD, Closure Compiler, just loading it using script tags, and we need to support all of them.

src/Subscriber.ts
```diff
@@ -1,4 +1,5 @@
 import { isFunction } from './util/isFunction';
+import {getZone} from './util/getZone';
```
fxck: missing spaces, `{getZone}` => `{ getZone }`

blesh:
So, I think I'm on-board with the zone from the subscription carrying through the async calls... but it seems weird that the zone at the time `.map` was called is going to be captured. I think the concepts of Observable are being conflated with the concepts of Promises here. I can totally understand why you'd want to capture the zone for `then` with promises... but since observables are really just functions, `map` is more of a higher-order function. When you're calling operators you're building functions that get called when you `subscribe`. It's whatever zone you have at `subscribe` time that's probably important.

@mhevery I'd really like to see a solid example of a bug with zones this PR solves. I think it would help everyone understand the concerns and the need better than just discussing the academics of it.

To reiterate what I'm thinking, it seems like this should hold true:

```js
let source$;
zone1.run(() => {
  source$ = new Observable(observer => {
    console.log('init', Zone.current.name);
    observer.next(1);
    observer.complete();
  });
});

zone2.run(() => {
  source$ = source$.mergeMap(x => {
    console.log('mergeMap', Zone.current.name);
    return Observable.timer(1000).map(() => {
      console.log('map', Zone.current.name);
      return x;
    });
  });
});

zone3.run(() => {
    source$.subscribe(x => {
      console.log(Zone.current.name);
    });
});

/** output
  init zone3
  mergeMap zone3
  // delay here
  map zone3
  sub zone3
*/
```

trxcllnt:
@blesh for context, @mhevery @jayphelps and I got on a call to discuss this last night. Takeaways are that I'm going to try and put together an alternate PR this week/weekend that monkey-patches in Zone support (and would be imported as needed like operators, via `import 'rxjs/add/Zone'`), then do a perf bake-off between the two, as well as followup discussions on ergonomics.

Neither @jayphelps nor I want to dismiss a feature request that's critical to the success of the Angular team and the hundreds of thousands of engineers they support, and are OK merging this PR into core if we can't find an optimal alternative solution. Sound about right @jayphelps @mhevery?

blesh:
@trxcllnt that's fine. I'm more concerned that we're doing the wrong thing WRT to zones and Observables after talking with @jhusain about it. Given than both Zones and Observables are proposals before the TC39, I'd like to make sure we're doing the right thing.

mhevery:
> @blesh  So, I think I'm on-board with the zone from the subscription carrying through the async calls... but it seems weird that the zone at the time .map was called is going to be captured.

I have been trying to explain this, and I have been failing :-(. Let me try again. The behavior you describe is the current behavior. In other words the constructor runs in the same zone as that of the `subscribe` invocation. This behavior is not correct.

Imagine we have a subscriber which fires once per second like so:
```
var obs = new Observable((sub) => {
  var count = 0;
  var id = setInterval(() => next(count++), 1000);
  return () => clearInterval(id);
});
```

Now imagine you run it through a filter which will only let go even numbers: 
```
obs = obs.filter((v) => v%2 == 0);
```

And finally someone subscribes to it:

```
obs.subscribe(() => {
  // do some rendering side-effect which requires Angular
});
```

By your logic all functions should be running in the Angular zone, which implies that Angular would do Change Detection every 1 second. But that is not what we want. Because of the filter only time when Change Detection is needed is ever 2 seconds. 

Our goal is that only the side-effect should be run in Angular Zone and hence have Change Detection only when the `subscribe` callback is run. The way we do that is this:


```
var obs;
rootZone.run(() => {
  obs = new Observable((sub) => {
    var count = 0;
    var id = setInterval(() => next(count++), 1000);
    return () => clearInterval(id);
  });
obs = obs.filter((v) => v%2 == 0);
});

angularZone.run(() => {
  obs.subscribe(() => {
    // do some rendering side-effect which requires Angular
  });
});
```

In the above example we are now guaranteed, that the only time Change Detection will run is when the `subscribe` callback runs not every time the `setInterval` fires.

blesh:
> I have been trying to explain this, and I have been failing :-(. Let me try again.

You explained it to me (successfully) offline after I made that comment.

trxcllnt:
@mhevery do the tests in this PR run on your machine? I'm getting an error deep inside `node-zone.js` during mocha initialization. A mocha class instantiates an Error in its constructor, which gets intercepted by `ZoneAwareError`, then `ZoneAwareError` is throwing this error inside `assignAll`: `TypeError: Cannot assign to read only property 'toJSON' of object 'Error: done() called multiple times'`. Curiously, the call to `assignAll` doesn't exist in [zone.js master](https://github.com/angular/zone.js/blob/master/dist/zone-node.js#L719).

trxcllnt:
@mhevery also if I patch my local copy of `node-zone` to `try/catch` around initialization errors, I can get it to work. But then adding Zones globally breaks all our unit tests that use chai to validate instanceof Error (I get a lot of `expected { Object (stack, message, ...) } to be an error` failures). Is there a way to test zones without affecting the rest of our tests?

trxcllnt:
@mhevery I was able to move forward by pinning `zone.js` to v0.7.4. You can see the changes in https://github.com/trxcllnt/rxjs/commit/53eddb1cca1228fe6d81aa6bec0ee0a57c8f4664, or jump right into the action at [`rxjs/add/zone.js`](https://github.com/trxcllnt/rxjs/blob/53eddb1cca1228fe6d81aa6bec0ee0a57c8f4664/src/add/zone.ts).

Because zones patch the world, I created a separate `build_test_zone` step to compile and run the tests with the `zone.js` import and observable patch. Even if Zones aren't supposed to do anything if they're not used, I think it's best if we can keep running the tests as-is, then have a second step where they're run with zones.

I haven't done any perf tests yet. It will be a bit of a task to update our micro-perf-test-runner to run all permutations of "do/don't include zone.js + do/don't include the patch." In the meantime, feel free to comment on the implementation. I'm shooting somewhat in the dark here, so I'll need your guidance as to correct behavior.

mhevery:
@trxcllnt yes, we are aware of the zone v0.7.5 issue, should be releasing v0.7.6 soon.

mhevery:
@trxcllnt looks like the zone specs are passing, so I assume that the implementation is correct and it is doing what it is supposed to.

I do have to say that the patching is very hard for me to follow, even ofter sort up understanding what is going on.

What is the next step? I assume we need to get the perf numbers?

trxcllnt:
> @blesh Can we just use plain ol' prototype assignment in that code? Just for readability's sake. defineProperties might be slightly more performant in some environments, but I'd be surprised if it is to a degree it's worth the hit to readability.

We have to use Object.defineProperties to turn Observable's `source` and `_subscribe`, Subscription's `_unsubscribe`, and Subscriber's `destination` fields into getters/setters. Since these fields are set in the constructors (or in `lift`), we can capture the value of `Zone.current`.

If these fields aren't set (like when Observable/Subscription aren't called with subscribe/unsubscribe callbacks), we don't read Zone.current, which we only need when the CBs are provided.

The others yeah, I'd prefer proto patching. I originally wrote it that way in ES6, and I can't remember exactly right now, but I think Typescript yelled at me when I made it a TS file.

> we should get perf numbers for all scenarios.

👍 for sure. I've been really busy with work lately. The quick and dirty way (I _think_) would be to add another cmd line arg to `perf/micro/index.js` to require `rxjs/add/zone.js` up front. That'll be rough to do a side-by-side though, since we'd be testing this branch of Rx 5 vs. 4, then my patch branch of 5 vs. 4, and have to infer the perf diffs. It could even be totally bogus. Ideally we'd do something like https://github.com/ReactiveX/rxjs/issues/2198.

> Also, I have a distinct feeling that there are some operators that will slip under the radar for both of these.

I won't say it's not possible, but I'm pretty sure we got all the bases covered Observable API. The only thing I left out was patching the Schedulers' `schedule` method, because a. patching lift/Subscriber/Subscription will capture/switch to the right Zone.current, b. they weren't in this PR, and c. I forgot :-P

fxck:
Are there any updates? It this blocked by something?

benlesh:
@fxck just debating the best approaches and trying to find the time to vet each one.

zxsoft:
Please, let us know what's going on about this PR. I think it's quite important for future plans for the Angular  dev-team and users.

chaosmonster:
What's the status?

trxcllnt:
@chaosmonster aside from rebasing these PRs from master, the biggest hurdle is we need to work to get versions of the perf tests that allow us to run w/ zones enabled and disabled, then create a comparison matrix that looks something like this:

|  | Rx master | Rx w/ Zones in core | Rx w/ Zones patched |
| ------------------------ | ---------- | -------------------- | -------------------- |
| with `require('zone)`      | ✔ | ✔ | ✔ |
| without `require('zone)` | ✔ | ✔ | ✔ |

chaosmonster:
thx for the insides

philjones88:
There been any movement on this? We've just hit this and had to add workarounds.

jayphelps:
@philjones88 hey, can you clarify what exactly what you hit? AFAIK this is more a correctness thing to reduce some unnecessary diffing when nothing changed. I'm personally skeptical that an app's performance would make or break on this PR but if that's the case I'm interested to know how 👍

mleibman:
@jayphelps This isn't just about Angular though.  It's about making it work in a more sane way with Zone.js, which is super useful for all sorts of things (testing/conformance checks/performance instrumentation/etc.).  Angular's use of zones for change detection is just one aspect of it.

philjones88:
@jayphelps see my GH issue on the zone.js repo:

https://github.com/angular/zone.js/issues/830

I got redirected here after eventually figuring out it was an rxjs + zone.js issue rather than a browser specific problem.

ptitjes:
Sorry to bother but is there any news ?

trotyl:
It has already been fixed on Zone.js side.

meriturva:
@trotyl but not so perfectly: https://github.com/ReactiveX/RxJS/issues/1391

somombo:
If it's fixed then why is this issue still open

benlesh:
@somombo I'm waiting to hear back from the Angular team. As such, this PR is on hold.

somombo:
@benlesh thanks for the timely feedback, appreciate it

benlesh:
I think this PR is stale, and we all agree that adding Zones to RxJS wasn't what we wanted to do. I'm going to close this one.

benlesh:
(Waits to change teams, closes his boss's PR)

PR #2268, opened 2017-01-09
Add sample/sampleTime operator split

sergi:
**Description:**
When using a number interval, we must now use `sampleTime` instead of `sample`. This was not specified in the document.


jayphelps:
👍  thank you!

LGTM

PR #2271, opened 2017-01-11
fix(catch): update the catch operator to dispose inner subscriptions if the catch subscription is disposed

trxcllnt:
This is a non-breaking change to make track and dispose its subscription to the Observable returned from the catch selector if the catch subscriber is disposed early.

Without this fix, subscriptions to Observables returned by the catch selector don't get cleaned up, and will leak memory if the `catch` is used on a Subject, and the catch selector returns a sequence that uses the second `caught` Observable argument :-(


staltz:
LGTM

blesh:
LGTM

PR #2276, opened 2017-01-13
fix(combineLatest): Don't mutate array of observables passed to combineLatest

extremeandy:
**Description:**

`combineLatest` was mutating the array of observables passed to it. This is a very simple fix to clone the array before modifying it.

See this codepen for example of the problem: https://codepen.io/anon/pen/qRNBVJ?editors=1111


blesh:
@extremeandy ... good catch. This looks like a side-effect from the time we added the ability to do both `combineLatest(a, b, c)` and `combineLatest([a, b, c])` (originally it was just the former, which used rest arguments).

LGTM! :+1:

PR #2278, opened 2017-01-14
fix(windowTime): clean up closed window with timeSpanOnly

kwonoj:
**Description:**
This PR separates https://github.com/ReactiveX/rxjs/pull/2240, updates clear up closed instance of window to when invoked via timespan only.


blesh:
This change seems good. @kwonoj, can you please resolve the conflicts?

blesh:
Coveralls is complaining just because of the math of removing a couple of lines I think. I didn't see anything glaring to be concerned about.

Issue #2004, opened 2016-10-04
Subject.create vs AnonymousSubject.create


benlesh:
Related #2002, #2003, #1996, #1975

Currently, `Subject.create` is built to mimic it's behavior in v4 and under. That is that `Subject.create` actually returns an `AnonymousSubject`, and not a `Subject`. 
### Problem
- This flies in the face of other uses of `create` in this library, which is as an alternative to `new`... for example: `new Observable(fn)` vs `Observable.create(fn)`. This creates confusion for users (and experts, apparently)
- We can't move `AnonymousSubject` into it's own module, because there would be a circular reference with `Subject`, which has this create method that only creates an `AnonymousSubject`.
### Proposal:
- `Subject.create` becomes alternative version of `new Subject`
- `AnonymousSubject.create` does what `Subject.create` does now... which is just an alternative version to `new AnonymousSubject`.
- move `AnonymousSubject` to it's own module.

reduckted:
This is really confusing behaviour. I figured, since `new Observable()` and `Observable.create()` are basically the same, that `new Subject()` and `Subject.create()` would also be the same. How wrong I was!

zoechi:
#2002 was merged but that doesn't make `AnonymousSubject` public. For this it would need to be listed here as well https://github.com/ReactiveX/rxjs/blob/master/src/index.ts#L10.
Also `new AnonymouseSubject(...)` supports a `destination` and `source` parameter, while `new Subject()` or `Subject.create()` do not. Some comment on how to translate would be great when one is deprecated.

PR #2279, opened 2017-01-14
refactor(Subject): introduce interface to Subject

kwonoj:
**Description:**
This PR enhances type definition of `subject` separated from https://github.com/ReactiveX/rxjs/pull/2233, similar to https://github.com/ReactiveX/rxjs/pull/2249.

This change introduces new interface `ISubject<T>` have more clear type definition regarding its properties, and let static creation method returns interface instead of actual implementation type.

### DIFF
--- a/src/Subject.ts
+++ b/src/Subject.ts
@@ -16,40 +16,41 @@ export class SubjectSubscriber<T> extends Subscriber<T> {
   }
 }
 
+export interface ISubject<T> extends ISubscription, Observable<T>, Observer<T> {
+  readonly closed: boolean;
+  asObservable(): Observable<T>;
+}
+
 /**
  * @class Subject<T>
  */
-export class Subject<T> extends Observable<T> implements ISubscription {
+export class Subject<T> extends Observable<T> implements ISubject<T> {
 
   [$$rxSubscriber]() {
     return new SubjectSubscriber(this);
   }
 
-  observers: Observer<T>[] = [];
-
+  protected observers: Observer<T>[] = [];
+  protected isStopped = false;
+  protected hasError = false;
+  protected thrownError: any = null;
   closed = false;
-
-  isStopped = false;
-
-  hasError = false;
-
-  thrownError: any = null;
 
   constructor() {
     super();
   }
 
-  static create: Function = <T>(destination: Observer<T>, source: Observable<T>): AnonymousSubject<T> => {
+  static create: Function = <T>(destination: Observer<T>, source: Observable<T>): ISubject<T> => {
     return new AnonymousSubject<T>(destination, source);
   }
 
-  lift<R>(operator: Operator<T, R>): Observable<T> {
+  public lift<R>(operator: Operator<T, R>): Observable<T> {
     const subject = new AnonymousSubject(this, this);
     subject.operator = <any>operator;
     return <any>subject;
   }
 
-  next(value?: T) {
+  public next(value?: T): void {
     if (this.closed) {
       throw new ObjectUnsubscribedError();
     }
@@ -63,7 +64,7 @@ export class Subject<T> extends Observable<T> implements ISubscription {
     }
   }
 
-  error(err: any) {
+  public error(err: any): void {
     if (this.closed) {
       throw new ObjectUnsubscribedError();
     }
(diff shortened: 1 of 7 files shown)


jayphelps:
So this is only not a breaking change because TS is structurally typed?

kwonoj:
It is not breaking since this PR only extracts out interface shape from existing Subject implementation. It isn't being actively used elsewhere, so while it gives clarity who reads interface of Subject but actual type inferences related with subject won't benefit hugely. Previous PR was actually changing runtime behavior to prevent unintended interface access though.

jayphelps:
@kwonoj hmmm but isn't `Subject.create()` a public API?

```ts
// some user code that assumes Subject, not ISubject
function gimmieStuff(stuff: Subject) {}

let stuff = Subject.create(...);
gimmieStuff(stuff);
```

kwonoj:
@jayphelps , it is. My comment is based on most of subject creation is currently recommended via instantiation but not by static creation method. Anyway there are opened issue for definition of static subject creation method. https://github.com/ReactiveX/rxjs/issues/2004 too.

src/Subject.ts
```diff
@@ -16,40 +16,43 @@ export class SubjectSubscriber<T> extends Subscriber<T> {
   }
 }
 
+export interface ISubject<T> extends ISubscription, Observable<T> {
+  readonly observers: ReadonlyArray<Observer<T>>;
```
blesh: this is sort of an implementation detail of `Subject`.

src/Subject.ts
```diff
@@ -16,40 +16,43 @@ export class SubjectSubscriber<T> extends Subscriber<T> {
...
+  readonly closed: boolean;
+  readonly isStopped: boolean;
```
blesh: `isStopped` is also an implementation detail.

src/Subject.ts
```diff
@@ -16,40 +16,43 @@ export class SubjectSubscriber<T> extends Subscriber<T> {
   }
 }
 
+export interface ISubject<T> extends ISubscription, Observable<T> {
```
blesh: Why does `ISubject` not extend an `Observer` interface as well?

src/Subject.ts
```diff
@@ -16,40 +16,43 @@ export class SubjectSubscriber<T> extends Subscriber<T> {
...
  * @class Subject<T>
  */
-export class Subject<T> extends Observable<T> implements ISubscription {
+export class Subject<T> extends Observable<T> implements ISubject<T> {
 
   [$$rxSubscriber]() {
     return new SubjectSubscriber(this);
   }
 
   observers: Observer<T>[] = [];
```
blesh: should probably be protected.

src/Subject.ts
```diff
@@ -16,40 +16,43 @@ export class SubjectSubscriber<T> extends Subscriber<T> {
...
-
   closed = false;
-
   isStopped = false;
```
blesh: should probably be protected

blesh:
So where are we returning `ISubject`? or expecting `ISubject`? How does this typing benefit us?

Igorbek:
If an interface being extracted, why doesn't it follow the classic definition of the _Subject_ where it is just simply that:
```ts
interface ISubject<T> extends /*IObserver<T>*/ Observer<T>, /*IObservable<T>*/ Subscribable<T> { }
```
all other attributes would make the abstraction leaking (as do `Observer` and `Subscribable` currently).

blesh:
cc @kwonoj ... I'm generally in favor of this change... can you make the changes I've requested? Or rebutt them I guess? haha. :)

blesh:
Also... I *think* this is a MINOR version change, since it's not really breaking, but it does sort of add a feature. Opinions? @staltz? @trxcllnt?

kwonoj:
@blesh oh yes, I just forgot this one :/

and I think this can be definitely minor, as it doesn't breaking existing behavior in most cases.

kwonoj:
Updated PR as suggested. /cc @blesh

staltz:
Seems like a MINOR. Adds feature with backwards compat.

kwonoj:
So, making interface clear makes cannot build this, as some of internal implementation (SubjectSubscription, COnnectableObservable) was relying on incorrectly publicly-exposed ones. :/

benlesh:
@kwonoj.. this seems to have gone stale. Can we close it?

kwonoj:
I still believe this need to be landed in master in some way, but this PR itself has too diverged. Closing it and let me create other PR based on TOT.

PR #2286, opened 2017-01-16
docs(single): add missing param name

gkalpak:
**Description:**
Add missing param name (and fix casing).


staltz:
Looks good

gkalpak:
(See https://github.com/ReactiveX/rxjs/pull/2288#issuecomment-272955366.)

PR #2288, opened 2017-01-16
docs(*): several fixes

gkalpak:
**Description:**
Several docs fixes, including:

- Add missing params/param names.
- Remove non-eistent params.
- Fix typos.
- Consistent capitalization.
- ...

### DIFF
--- a/src/operator/skip.ts
+++ b/src/operator/skip.ts
@@ -4,18 +4,18 @@ import { Observable } from '../Observable';
 import { TeardownLogic } from '../Subscription';
 
 /**
- * Returns an Observable that skips `n` items emitted by an Observable.
+ * Returns an Observable that skips the first `count` items emitted by the source Observable.
  *
  * <img src="./img/skip.png" width="100%">
  *
- * @param {Number} the `n` of times, items emitted by source Observable should be skipped.
+ * @param {Number} count - The number of times, items emitted by source Observable should be skipped.
  * @return {Observable} an Observable that skips values emitted by the source Observable.
  *
  * @method skip
  * @owner Observable
  */
-export function skip<T>(this: Observable<T>, total: number): Observable<T> {
-  return this.lift(new SkipOperator(total));
+export function skip<T>(this: Observable<T>, count: number): Observable<T> {
+  return this.lift(new SkipOperator(count));
 }
 
 class SkipOperator<T> implements Operator<T, T> {
--- a/src/operator/skip.ts
+++ b/src/operator/skip.ts
@@ -9,7 +9,7 @@ import { TeardownLogic } from '../Subscription';
  * <img src="./img/skip.png" width="100%">
  *
  * @param {Number} count - The number of times, items emitted by source Observable should be skipped.
- * @return {Observable} an Observable that skips values emitted by the source Observable.
+ * @return {Observable} An Observable that skips values emitted by the source Observable.
  *
  * @method skip
  * @owner Observable
--- a/src/observable/BoundCallbackObservable.ts
+++ b/src/observable/BoundCallbackObservable.ts
@@ -74,7 +74,7 @@ export class BoundCallbackObservable<T> extends Observable<T> {
    *
    * @param {function} func Function with a callback as the last parameter.
    * @param {function} [selector] A function which takes the arguments from the
-   * callback and maps those a value to emit on the output Observable.
+   * callback and maps those to a value to emit on the output Observable.
    * @param {Scheduler} [scheduler] The scheduler on which to schedule the
    * callbacks.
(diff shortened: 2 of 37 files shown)


kwonoj:
👋 @gkalpak, appreciate for PR - would you mind aggregating `missing param name` cases into single PR, if you have additional changes ready? Each changes have same scope, so easier to have single PR to review. 

If there isn't further PR, it's ok to leave current one as-is.

Thanks!

gkalpak:
I am reading through the docs and submitting PRs with small typos as I come across them. I can aggregate `missing param name` into one if that works better for you. (I'll close #2286 and #2287 to include in the new PR too.)

Question:
I've seen that some operators have a dash between the param name and param description. Although not necessary, I find the dash much more readable (e.g. as in [here][1]). Do you want the dash or should I rather leave it off?

cc @staltz (see https://github.com/ReactiveX/rxjs/pull/2287#issuecomment-272952555)

[1]: https://github.com/ReactiveX/rxjs/pull/2287/files#diff-29f6e8df442d02c119249aa36614ca9fR11

gkalpak:
I ended up with quite a few more commits than I excepted. I've moved everything here (mostly grouped into 1 commit per operator), but if you prefer me to split it up differently, just say so 😃 

cc @kwonoj, @staltz

jayphelps:
Since docs are not usually something we need to revert one-by-one, separate commits for operators are not really neccesary. But you don't have to squash them, we can do it when we merge. There's a merge conflict btw, rebase plz

gkalpak:
Rebased

src/operator/skip.ts
```diff
@@ -8,8 +8,8 @@ import { TeardownLogic } from '../Subscription';
  *
  * <img src="./img/skip.png" width="100%">
  *
- * @param {Number} the `n` of times, items emitted by source Observable should be skipped.
- * @return {Observable} an Observable that skips values emitted by the source Observable.
+ * @param {Number} n - The number of times, items emitted by source Observable should be skipped.
```
martinsik: The `@param` needs to match the argument name in the function declaration so this should be:

@param {Number} total

gkalpak: The description reads: "Returns an Observable that skips `n` items emitted by an Observable."
If I change the param name to `total`, it will be confusing imo. Should I change the description as well?

martinsik: I'm refering to this https://esdoc.org/manual/usage/tags.html#-param. The `<name>` has to match a parameter name. Function `skip()` has one parameter called `total` not `n`. I think the variable should be actually called `count` to be consistent with [`take()`](https://github.com/ReactiveX/rxjs/blob/master/src/operator/take.ts#L9).

gkalpak: I know what you are referring to :smiley:

The thingis that the docs are generated based on the ESDoc tags (apparently), not the actual param names in code. So currently, people don't see `total` at all.

Here is what they see currently:

> public skip(the: Number): Observable
>
> ##
> Returns an Observable that skips `n` items emitted by an Observable.
> 
> Params:
>
> Name | Type | Attribute | Description
> -|-|-|-
> the | Number |  | `n` of times, items emitted by source Observable should be skipped.

Here is what they will see with this PR (in its current state):

> public skip(n: Number): Observable
>
> ##
> Returns an Observable that skips `n` items emitted by an Observable.
> 
> Params:
>
> Name | Type | Attribute | Description
> -|-|-|-
> n | Number |  | The number of times, items emitted by source Observable should be skipped.

Here is what they see with your suggestion:

> public skip(total: Number): Observable
>
> ##
> Returns an Observable that skips `n` items emitted by an Observable.
> 
> Params:
>
> Name | Type | Attribute | Description
> -|-|-|-
> total | Number |  | The number of times, items emitted by source Observable should be skipped.

In this last version, the description mentions `n`, which is not refernced anywhere else, so my question was: Should I update the description too, so that it doesn't mention `n` but `total`.

##
And now that I have written this long answer, it seems redundant - it would be simpler if I went ahead and did the change and wait for feedback :grin:
<sub>(So, that's what I'm gonna do...)</sub>

gkalpak: Done.

gkalpak:
More conflicts :scream: Feel free to ping me when/if you are ready to merge this, so I won't be rebasing over and over :smiley:

gkalpak:
Done.

jayphelps:
@kwonoj I think this is ready to be merged via squash, I was last to comment so didn't want to leave this hanging.

kwonoj:
@jayphelps I'll briefly go through and make it happen in today after getting back from meetings.

PR #2291, opened 2017-01-17
fix(reduce): forward index to accumulator

### LINKED ISSUE #2290: Reduce() overload invoked with index=undefined
**RxJS version:**
5.0.3
**Code to reproduce:**
```
Rx.Observable.from([1,2,3,4,5])
  .reduce((acc,curr,index) => `${acc}\ncurr:${curr}@index(${index})`, '')
  .subscribe(console.log);
```

**Expected behavior:**

```
curr:1@index(0)
curr:2@index(1)
curr:3@index(2)
curr:4@index(3)
curr:5@index(4)"
```
**Actual behavior:**

```
curr:1@index(undefined)
curr:2@index(undefined)
curr:3@index(undefined)
curr:4@index(undefined)
curr:5@index(undefined)"
```

**Additional information:**

The `seed` argument does not matter, results are the same

kwonoj:
**Description:**
This PR updates `reduce` to forward index of an element as same as native reduce operator does. 

- closes #2290

### DIFF
--- a/src/operator/reduce.ts
+++ b/src/operator/reduce.ts
@@ -44,7 +44,7 @@ export function reduce<T, R>(this: Observable<T>, accumulator: (acc: R, value: T
  * @see {@link mergeScan}
  * @see {@link scan}
  *
- * @param {function(acc: R, value: T): R} accumulator The accumulator function
+ * @param {function(acc: R, value: T, index: number): R} accumulator The accumulator function
  * called on each source value.
  * @param {R} [seed] The initial accumulation value.
  * @return {Observable<R>} An observable of the accumulated values.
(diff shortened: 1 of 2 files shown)


src/operator/reduce.ts
```diff
@@ -81,15 +81,20 @@ export class ReduceOperator<T, R> implements Operator<T, R> {
...
-              private accumulator: (acc: R, value: T) => R,
+              private accumulator: (acc: R, value: T, index?: number) => R,
               seed: R,
               private hasSeed: boolean) {
     super(destination);
     this.acc = seed;
+
+    if (!this.hasSeed) {
+      this.index++;
+    }
```
paulpdaniels: Why does index start at 1 if there is no seed?

kwonoj: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

> currentIndex The index of the current element being processed in the array. Starts at index 0, if an initialValue is provided, and at index 1 otherwise.

I tried to follow same behavior to native reduce, where it says start from 0 if seed exists, 1 otherwise.

kwonoj: maybe I've read spec incorrectly?

paulpdaniels: @kwonoj Nope, you read it correctly. TIL.

crunchie84: Maybe worth documenting this behaviour in the RxJs documentation?

PR #2292, opened 2017-01-17
docs(readme): Improve clarity of library installation and usage

### LINKED ISSUE #2269: Unnecessary README.md special case instructions for TypeScript
The current instructions direct TypeScript users to install using `npm install @reactivex/rxjs`.

However it works just fine to do it the simple way: `npm install rxjs` followed by e.g. `import { Observable } from 'rxjs';`

spirosikmd:
**Description:**
This PR updates README.md to clarify the installation and usage of the library for different systems CJS/ES6/AMD/TypeScript.

- closes #2269

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,8 @@ Observable::of(1,2,3)::map(x => x + '!!!'); // etc
 
 ### CommonJS via npm
 
+To install this library for CommonJS (CJS) usage, use the following command:
+
 ```sh
 npm install rxjs
 ```
@@ -111,6 +113,8 @@ To install this library via [npm](https://www.npmjs.org) **version 3**, use the 
 ```sh
 npm install @reactivex/rxjs
 ```
+
+This will include CJS/Global builds and can be used for all module types.
 
 If you are using npm **version 2** before this library has achieved a stable version, you need to specify the library version explicitly:
 


README.md
```diff
@@ -112,6 +114,8 @@ To install this library via [npm](https://www.npmjs.org) **version 3**, use the
 npm install @reactivex/rxjs
 ```
 
+This will include all builds for CJS/ES6/AMD/TypeScript.
```
kwonoj: current RxJS packages does not ship `es6` builds.

spirosikmd: So the output builds are CJS/AMD/TypeScript?

spirosikmd: @kwonoj does README actually need to mention also `npm install rxjs`? Doesn't `npm install @reactivex/rxjs` work for every bundling system? You can either use CommonJS, or ES6 and transpile it, or AMD, or TypeScript and compile it with either es5/es6 module option? Maybe `npm install rxjs` can be removed?

kwonoj: For now it's only `cjs/global`. I don't think instruction to use `rjxs` need to be removed though.

billba: I much prefer `npm install rxjs`. Works great for TypeScript. I guess this is a naive question, but why wouldn't we prefer the shorter version?

kwonoj: Well, I think this PR's crux is to resolve those question :) if you're using cjs only, there is no difference between two package. @reactivex scoped package is intended for someone who wants to have full builds other than cjs such as global (and potentially es6 once it's sorted out). If you prefer shorter package, then it's good to go. Is this answers for you?

kwonoj: And clarify, 'including all builds for Typecript' is not correct term and I mentioned it in above comment that scoped package contains cjs/global only. Important thing for TS user is both package is being shipped with type definition can be used for typescript codebases.

billba: I agree that's the crux of the problem. I think this is one of those situations where more words is better. It would be great to spell out why there are two different packages, and go into detail about who should use each one. It's frustrating to have a confusing/ambiguous decision point right at the very beginning of one's relationship with a library.

spirosikmd:
Hey @benlesh! Any update on this one?

PR #2305, opened 2017-01-20
refactor(TestScheduler): return hot and cold observables

Podlas29:
**Description:**

Return HotObservable and ColdObservable instead of simply Observable, in order to give users access to subscriptions parameter.

**Motivation:**

Previously getting subscriptions of observables for `expectSubscriptions` assertions was impossible without typecasting:
```javascript
const hot = scheduler.createHotObservable('--a---b--|');
hot.subscriptions // does not compile
```


kwonoj:
Looks good to me.

PR #2308, opened 2017-01-23
chore(*): correctly scope disabled `max-line-length` tslint rule

gkalpak:
**Description:**
The max line length is set to 150 in 'tslint.json'. In specific regions, it is desirable to allow longer lines, so these regions should be wrapped in comments like the following:

```js
// Max line length enforced here.

// Max line length NOT enforced here.
/* tslint:enable:max-line-length */   <-- CORRECT

// Max line length enforced here.
```

In many cases, the re-enabling comment incorrectly included `disable` instead of `enable` (as shown below), which essentially keeps the `max-line-length` rule **disabled** for the rest of the file:

```js
// Max line length enforced here.

// Max line length NOT enforced here.
/* tslint:disable:max-line-length */   <-- INCORRECT

// Max line length NOT enforced here.
```

This commit fixes these comments, so the `max-line-length` rule is properly enforced in regions that don't need longer lines.

(Luckily, there was only 1 violation :smiley:)


blesh:
@gkalpak can you better outline the issue that this fixes?

gkalpak:
@blesh, I have updated the commit message and PR description.

david-driscoll:
@blesh looks like a goof on my part, this change looks good from that perspective.

jayphelps:
Yeah, this is a whoopies!

PR #2309, opened 2017-01-26
Use onNext, onError, onCompleted in custom operator example

janmarthedal:


kwonoj:
It is expected to use v5 interfaces in the document. (https://github.com/ReactiveX/rxjs/blob/master/MIGRATION.md#observer-interface-changes-also-subjects)

PR #2316, opened 2017-01-30
feat(skipLast): add skipLast operator 

### LINKED ISSUE #1404: Are we missing skipLast operator?
We currently don't have skipLast implemented. Are we removing it intentionally or just forgot to build it?

Notice the discrepancy here:
- `skip`
- `skipUntil`
- `skipWhile`
- `take`
- `takeLast`
- `takeUntil`
- `takeWhile`

martinsik:
- [X] Add the operator to Rx
- [X] It must have a `-spec.ts` tests file covering the canonical corner cases, with marble diagram tests
- [X] If possible, write a `asDiagram` test case too, for PNG diagram generation purposes
- [X] The operator must be documented in JSDoc style in the implementation file, including also the PNG marble diagram image
- [X] The operator should be listed in `doc/operators.md` in a category of operators
- [X] It should also be inserted in the operator decision tree file `doc/decision-tree-widget/tree.yml`
- [X] You may need to update `MIGRATION.md` if the operator differs from the corresponding one in RxJS v4

**Description:**

Adds skipLast operator from RxJS 4. Its internals and tests are based on takeLast for better performance.

Closes #1404

### DIFF
--- a/doc/decision-tree-widget/tree.yml
+++ b/doc/decision-tree-widget/tree.yml
@@ -59,6 +59,11 @@ children:
         - label: based on custom logic
           children:
           - label: skipWhile
+      - label: from the end of the Observable
+        children:
+        - label: based on a given amount
+          children:
+          - label: skipLast
       - label: until another Observable emits a value
         children:
         - label: skipUntil
--- a/doc/decision-tree-widget/tree.yml
+++ b/doc/decision-tree-widget/tree.yml
@@ -61,9 +61,7 @@ children:
           - label: skipWhile
       - label: from the end of the Observable
         children:
-        - label: based on a given amount
-          children:
-          - label: skipLast
+        - label: skipLast
       - label: until another Observable emits a value
         children:
         - label: skipUntil
--- a/spec/operators/skipLast-spec.ts
+++ b/spec/operators/skipLast-spec.ts
@@ -1,0 +1,155 @@
+import {expect} from 'chai';
+import * as Rx from '../../dist/cjs/Rx';
+declare const {hot, cold, asDiagram, expectObservable, expectSubscriptions};
+
+const Observable = Rx.Observable;
+
+/** @test {takeLast} */
+describe('Observable.prototype.skipLast', () => {
+  asDiagram('skipLast(2)')('should skip two values of an observable with many values', () => {
+    const e1 =  cold('--a-----b----c---d--|');
+    const e1subs =   '^                   !';
+    const expected = '-------------a---b--|';
+
+    expectObservable(e1.skipLast(2)).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should skip last three values', () => {
+    const e1 =  cold('--a-----b----c---d--|');
+    const e1subs =   '^                   !';
+    const expected = '-----------------a--|';
+
+    expectObservable(e1.skipLast(3)).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should skip all values when trying to take larger then source', () => {
+    const e1 =  cold('--a-----b----c---d--|');
+    const e1subs =   '^                   !';
+    const expected = '--------------------|';
+
+    expectObservable(e1.skipLast(5)).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should skip all element when try to take exact', () => {
+    const e1 =  cold('--a-----b----c---d--|');
+    const e1subs =   '^                   !';
+    const expected = '--------------------|';
+
+    expectObservable(e1.skipLast(4)).toBe(expected);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should not skip any values', () => {
+    const e1 =  cold('--a-----b----c---d--|');
+    const e1subs =   '^                   !';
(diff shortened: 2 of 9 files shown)


martinsik:
I'm actually not sure what the last unchecked bullet point means:

> The spec file should have a type definition test at the end of the spec to verify type definition for various use cases

Is there any other operator that already tests this?

staltz:
You can ignore that task, it used to be needed.

doc/decision-tree-widget/tree.yml
```diff
@@ -59,6 +59,11 @@ children:
         - label: based on custom logic
           children:
           - label: skipWhile
+      - label: from the end of the Observable
+        children:
+        - label: based on a given amount
+          children:
+          - label: skipLast
```
staltz: Because there is only one child under "from the end of the Observable", I suggest you combine them into one. So:
```
- label: from the end of the Observable
  children:
  - label: skipLast
```

spec/operators/skipLast-spec.ts
```diff
@@ -0,0 +1,155 @@
...
+declare const {hot, cold, asDiagram, expectObservable, expectSubscriptions};
+
+const Observable = Rx.Observable;
+
+/** @test {takeLast} */
+describe('Observable.prototype.skipLast', () => {
+  asDiagram('skipLast(2)')('should skip two values of an observable with many values', () => {
+    const e1 =  cold('--a-----b----c---d--|');
+    const e1subs =   '^                   !';
+    const expected = '-------------a---b--|';
```
staltz: There is something really wrong about this marble diagram. The bottom `a` should *not* be emitted at the same time as the top `c`, because when `c` occurs, we don't actually yet know that it's the second-last emission. We can only know that when the `e1` completes. So the expected should be `(ab|)` when the source completes.

mbonneau: As soon as `c` is emitted, it is known that `a` should not be skipped because it is not one of the last 2 elements and can be emitted.

staltz: Hmm, funny. I may have misunderstood skipLast. Is this how RxJS 4 works?

mbonneau: It appears to be: https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/skiplast.js#L248 .

trxcllnt: Yeah, this looks correct. if the `skipLast(n)` buffer grows to `n + 1`, we know we can drain to `n` and emit.

src/operator/skipLast.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/**
+ * We need this JSDoc comment for affecting ESDoc.
+ * @ignore
+ * @extends {Ignored}
+ */
+class SkipLastSubscriber<T> extends Subscriber<T> {
+  private ring: T[] = [];
+  private count: number = 0;
```
blesh: A few things:

1. rename these to start with underscores. I know we're not doing that everywhere, but we should be with private and protected properties as a signal to non-TS users.
2. Rename `total` to `_skipCount` or `_numberToSkip`, something more descriptive.
3. We should be initializing the ring buffer in the ctor to be the same size as the `_skipCount`, it should perform better than constantly using `push`.
4. Instead of keeping a `_count`, we can keep a `_currentIndex` or `_index`, such that when we increment it we do:  `this._currentIndex = (_currentIndex + 1) % _skipCount`; That way it doesn't inflate past where it needs to be.

src/operator/skipLast.ts
```diff
@@ -0,0 +1,88 @@
...
+    const len = this.ring.length;
+
+    if (len < this.total) {
+      this.ring.push(value);
+      this.count++;
+    } else {
+      const idx = this.count++ % this.total;
+      const oldValue = this.ring[idx];
+
+      this.ring[idx] = value;
```
blesh: be sure if your accessing the same property more than once to pull it out into a `const`. That way we don't have the cost of looking it up over and over.

blesh:
Overall this looks good. I don't think `skipLast` is trivial to implement using existing operators, so this is a solid candidate to re-add to the library. I've just requested a few changes, WRT to performance and naming conventions. (Please do check the perf before and after the changes).

blesh:
Also, it seems the files added have 100% coverage, so I'm not worried about coveralls complaining.

src/operator/skipLast.ts
martinsik: I think I have to keep `_count` instead of just `_currentIndex` because I need to know when I should start remitting buffered items since the buffer is now fixed size.

martinsik:
After the changes performance is more or less the same:

Before:
```
                                         |                     RxJS 4.1.0 |                     RxJS 5.0.3 |          factor |      % improved
                    skiplast - immediate |                 3,154 (±1.23%) |                40,651 (±1.04%) |          12.89x |        1,189.0%
                                skiplast |                 1,442 (±2.35%) |                18,099 (±1.55%) |          12.55x |        1,155.2%
```

After:

```
                                         |                     RxJS 4.1.0 |                     RxJS 5.0.3 |          factor |      % improved
                    skiplast - immediate |                 3,218 (±2.27%) |                42,526 (±0.80%) |          13.21x |        1,221.4%
                                skiplast |                 1,497 (±2.31%) |                19,353 (±1.75%) |          12.92x |        1,192.5%
```

benlesh:
Thank you @martinsik!

PR #2318, opened 2017-01-31
fix(Subject): ensure subject properly throws ObjectUnsubscribedError …

blesh:
…when unsubscribed then resubscribed to

- Also changes one AsyncSubject test that seemed to have a typo where the author intended to dispose of a subscription, but disposed of the subject itself instead

### DIFF
--- a/spec/Subject-spec.ts
+++ b/spec/Subject-spec.ts
@@ -261,6 +261,9 @@ describe('Subject', () => {
     expect(() => {
       subject.subscribe(
         function (x) { results3.push(x); },
+        function (err) {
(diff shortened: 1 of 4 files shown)


blesh:
Basically, my last fix around errors thrown in the body of the observables initialization created a regression in subjects where instead of just throwing ObjectUnsubscribedError, it was sending it down the error chain. It also uncovered one bad test.

blesh:
Given that my fork build is passing (for the source of this PR), and we're waiting on Travis builds from RxSwift, I feel okay to merge this.

PR #2322, opened 2017-02-03
docs(bindCallback): add details about behaviour and compare to simila…

Podlas29:
**docs(bindCallback): add details about behaviour and compare to similar operators**

**Description:**

This is my attempt to make docs more in-depth.

I am myself on a quest to learn rx.js as well as possible, and docs currently are very general and leave many details untouched. I intend to go over every operator and describe it is as precisely as I can.

Please let me know if such an effort is welcome here. If not (perhaps docs are meant to be brief and to-the-point), I will make my own repository were such docs will land.

Also please note I am not english native speaker, so propositions how to phrase some things better will be very welcome.

Have a great day everyone!

### DIFF
--- a/src/observable/BoundCallbackObservable.ts
+++ b/src/observable/BoundCallbackObservable.ts
@@ -45,16 +45,60 @@ export class BoundCallbackObservable<T> extends Observable<T> {
    * last parameter must be a callback function that `func` calls when it is
    * done.
    *
-   * The output of `bindCallback` is a function that takes the same
-   * parameters as `func`, except the last one (the callback). When the output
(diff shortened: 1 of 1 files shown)


Podlas29:
Hi.

I added even more details and examples.

I described carefully how using `async` scheduler would affect behaviour of operator. Since I still do not fully grasp how schedulers operate, it would be best if someone proficient double checked if my claims are true. I checked examples and they seem to support what I have written.

I also described how Observable laziness affects when input function of `bindCallback` is really called.

blesh:
This is all solid work, @Podlas29! Thank you so much for your contributions.

PR #2326, opened 2017-02-04
docs(bindNodeCallback): add more details and examples

Podlas29:
**Description:**

Another attempt to make docs more in-depth and complete, with descriptions of all modes of usage and more examples.

More background on what and why I am doing is here: https://github.com/ReactiveX/rxjs/pull/2322


Podlas29:
Hi.

Just as in https://github.com/ReactiveX/rxjs/pull/2322 I have added notes about schedulers and how observable laziness affects when input function is called.

Also changed little bit how some things are described.

Issue #2254, opened 2017-01-06
bindNodeCallback returns incorrect result if undefined success argument in callback


stusherwin:
**RxJS version:**
5.0.0-beta.6 but bug is present in master as of f8ba77d8d933608caae7c2b25aaa8467eb9e8f3d

**Code to reproduce:**
````
let nodeFunction = function(param1: string, param2: string, callback: any) {
  // Bug only manifests when success parameter is implicitly undefined
  callback(null /*error*/); // no success parameter
}

Observable.bindNodeCallback(nodeFunction)('a', 'b')
  .subscribe(succ => console.log(succ));

````

**Expected behavior:**
`undefined`

**Actual behavior:**
`[]`

**Additional information:**
This is a bug IMO because it does not preserve the behaviour of the wrapped method. There are actual libraries in production that have this behaviour, e.g. [node-sqlite3](https://github.com/mapbox/node-sqlite3/wiki/API#databasegetsql-param--callback) (which is how I found it):

> **Database#get(sql, [param, ...], [callback])**
> 
> Runs the SQL query with the specified parameters and calls the callback with the first result row afterwards [...] The signature of the callback is function(err, row) {}. **If the result set is empty, the second parameter is undefined**, otherwise it is an object containing the values for the first row.

The problem appears to be in this line:
https://github.com/ReactiveX/rxjs/blob/master/src/observable/BoundNodeCallbackObservable.ts#L109

It could be fixed with something like this:
````
if(!innerArgs.length) {
  subject.next(undefined);
} else {
  subject.next(innerArgs.length === 1 ? innerArgs[0] : innerArgs);
}
````

But there's probably a more elegant way of doing it!

trxcllnt:
@stusherwin you're right, that's the correct fix. Happy to review if you want to submit a PR.

PR #2328, opened 2017-02-05
fix(bindCallback): emit undefined when callback is without arguments

Podlas29:
**Description**
Related to issue #2254 in `bindNodeCallback`. It turns out `bindCallback` has the same problem:

When wrapped function calls its callback without any arguments, resulting Observable will emit empty array, intead of undefined.

```javascript
function callback(cb) {
   cb();
}

Observable.bindCallback(callback)
.subscribe(value => console.log(value)); // value is []
```

In RxJS 4 in `fromCalllback` resulting Observable emits undefined as seen here:
https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/perf/operators/fromcallback.js#L20-L21
(when array is empty, calling `results[0]` results in `undefined`, which is emitted).

**Related to**
#2254

### DIFF
--- a/src/observable/BoundCallbackObservable.ts
+++ b/src/observable/BoundCallbackObservable.ts
@@ -15,6 +15,7 @@ export class BoundCallbackObservable<T> extends Observable<T> {
   subject: AsyncSubject<T>;
 
   /* tslint:disable:max-line-length */
+  static create(callbackFunc: (callback: () => any) => any, selector?: void, scheduler?: IScheduler): () => Observable<void>;
   static create<R>(callbackFunc: (callback: (result: R) => any) => any, selector?: void, scheduler?: IScheduler): () => Observable<R>;
   static create<T, R>(callbackFunc: (v1: T, callback: (result: R) => any) => any, selector?: void, scheduler?: IScheduler): (v1: T) => Observable<R>;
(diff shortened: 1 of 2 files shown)


src/observable/BoundCallbackObservable.ts
```diff
@@ -15,6 +15,7 @@ export class BoundCallbackObservable<T> extends Observable<T> {
   subject: AsyncSubject<T>;
 
   /* tslint:disable:max-line-length */
+  static create(callbackFunc: (callback: () => any) => any, selector?: void, scheduler?: IScheduler): () => Observable<void>;
```
kwonoj: It's not related to this PR, but I just noticed `bindNodeCallback` doesn't have corresponding overloading signatures. Maybe we need this? /cc @david-driscoll

Podlas29: Now that I think about it, I have seen some operators have type-only tests. Maybe I should add ones here?

kwonoj: No, that is intended for specific cases if current test case doesn't covers it. it is worth to create those and it'll be appreciated, but I'd like to suggest to separate with this PR since each has different scope.

PR #2329, opened 2017-02-05
fix(bindNodeCallback): emit undefined when callback has no success ar…

### LINKED ISSUE #2254: bindNodeCallback returns incorrect result if undefined success argument in callback

Podlas29:
**Description:**

Currently when callback is called without success arguments (second and following arguments),
`bindNodeCallback` emits empty array.

In RxJS 4 `undefined` is emitted instead as can be seen here:
https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/perf/operators/fromnodecallback.js#L23-L24
(when results array is empty `results[0]` returns `undefined` which is then emitted)

Related issue with `bindCallback` #2328 
Closes #2254

### DIFF
--- a/spec/observables/bindNodeCallback-spec.ts
+++ b/spec/observables/bindNodeCallback-spec.ts
@@ -8,6 +8,24 @@ const Observable = Rx.Observable;
 /** @test {bindNodeCallback} */
 describe('Observable.bindNodeCallback', () => {
   describe('when not scheduled', () => {
+    it('should emit undefined when callback is called without success arguments', () => {
+      function callback(cb) {
+        cb(null);
+      }
+
+      const boundCallback = Observable.bindNodeCallback(callback);
+      const results = [];
+
+      boundCallback()
+        .subscribe((x: any) => {
+          results.push(typeof x);
+        }, null, () => {
+          results.push('done');
+        });
+
+      expect(results).to.deep.equal(['undefined', 'done']);
+    });
+
     it('should emit one value from a callback', () => {
       function callback(datum, cb) {
         cb(null, datum);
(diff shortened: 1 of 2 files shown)


spec/observables/bindNodeCallback-spec.ts
```diff
@@ -8,6 +8,24 @@ const Observable = Rx.Observable;
...
+      function callback(cb) {
+        cb(null);
+      }
+
+      const boundCallback = Observable.bindNodeCallback(callback);
+      const results = [];
+
+      boundCallback()
+        .subscribe((x: any) => {
+          results.push(typeof x);
```
kwonoj: just curiosity, is there reason to push type instead of actual value to assert with `undefined` directly?

Podlas29: I guess just habit of "everyone can set undefined value, so never trust it". All JS resources used to be very paranoid when I was learning. ;)

PR #2332, opened 2017-02-06
Ensure withCredentials is set after open for IE10

bluespore:
**Description:**
IE10 throws an `InvalidStateError` due to `withCredentials` being set **before** `open()`.  This seems to be due to the release of IE10 stuck on an older spec [as described in this StackOverflow ticket](http://stackoverflow.com/questions/19666809/cors-withcredentials-support-limited)

None.

### DIFF
--- a/src/observable/dom/AjaxObservable.ts
+++ b/src/observable/dom/AjaxObservable.ts
@@ -25,11 +25,7 @@ export interface AjaxRequest {
 
 function getCORSRequest(this: AjaxRequest): XMLHttpRequest {
   if (root.XMLHttpRequest) {
-    const xhr = new root.XMLHttpRequest();
-    if ('withCredentials' in xhr) {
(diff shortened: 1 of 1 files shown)


bluespore:
I only briefly opened the project in Chrome to check that it was still okay and it seemed fine. Though there may be another browser with an edge case I'm not aware of? Some assistance from another tester would be valuable.

bluespore:
@kwonoj Sorry, I made this change in a rush whilst at work - Let me know if you need me to update the commit format, etc to get it merged. Happy to make changes if necessary.

kwonoj:
@bluespore It'd be appreciate if you can amend commit message, but also I'm curious about how to test this in general to avoid any possible regressions.

blesh:
From what I know after researching it, this change is correct and safe. LGTM

bluespore:
@blesh Awesome, thanks for this Ben.

PR #2346, opened 2017-02-09
test(all): enable type inference to marble diagram

kwonoj:
**Description:**
This PR is following changes from https://github.com/ReactiveX/rxjs/pull/2202 apply enabling type inferences for marble test method.

- relates to #1633


kwonoj:
@jayphelps I just went ahead and enable most of test cases's type inference to prevent further PR blocked by our github bot 🤖 . Since this is trivial change, I've cc'ed you for review but will check in myself around tomorrow.

Issue #2350, opened 2017-02-09
Schedulers - is their behaviour arbitrary per operator? Is it possible to use them meaningfully without knowing operator implementation details?


Podlas29:
Hi.

So currently I am learning about schedulers and writing some docs for operators at the same time.

All this time I was convinced that scheduler behaviour is very predictable in every operator and it is just me (and maybe lack of good tutorials/docs) that does not know these rules.

And although it is true on very high level / from general perspective, the more I read source code and go into details, the more I am convinced that only good way to describe what scheduler does in general is "It schedules... Something. Sometimes. You want to know what exactly? Read our extremally long operator doc entry with all gory details, edge cases etc. Or just read source."

I'll give two examples of what I mean:

**scheduling async subscriptions to Observables in combineLatest and merge static functions**

Both `combineLatest` and `merge` accept collection of observables and subscribe to them one by one. Both accept optional schedulers. When using async scheduler, in one operator every subscription to input Observable happens in separate "task" in event loop, while second operator will ignore scheduler and simply iterate over Observables synchronously and subscribe to each in that loop.

Can you guess which one is which? How about without looking into code?

**queueScheduler in of and combineLatest**

```ts
Observable.combineLatest(
   Observable.of(1, 2, 3, schedulerA),
   Observable.of(4, 5, 6, schedulerA),
   null,
   schedulerB
)
```
You have four combinations of queue Scheduler and immediate Scheduler (think - no scheduler at all):

schedulerA - queue, schedulerB - immediate
schedulerA - immediate, schedulerB - queue
schedulerA - queue, schedulerB - queue
schedulerA - immediate, schedulerB - immediate

What will be the outcome of that obervable in all this combinations? Why?

**My questions**

* Can you - after gaining some proficiency in Rx - use schedulers reliably without looking too much into actual implementations of operators?
* Can you combine/stack different schedulers in different operators to have very fine-grained control when things will happen in your app? Or are schedulers usable and predictable only in simplest scenarios and you should stick only to those, when coding?
* When you implement new operator, are there clear rules what should be scheduled and how?
* Is it even possible to write operator that works predictably with every possible scheduler?
* Can you schedule in operator differently, depending on which scheduler you were passed, so that behaviour of every scheduler in operator is always reasonable? (for example when you have queue, schedule something one way, but when you have async, schedule something differently)
[...]

PR #2351, opened 2017-02-09
docs(combineLatest): add details about static version of operator

Podlas29:
**Description:**

Yet another pr with more detailed documentation for operator.

I struggled to meaningfully describe what scheduler actually does with this operator. This is related to my issue here: https://github.com/ReactiveX/rxjs/issues/2350. 

Irony is that I know how this operator behaves with different schedulers, but I have no idea how to describe it without going into implementation details (calling out `ArrayObservable` implementation and the fact that `combineLatest` subscribes to input Observables in its `complete` function). I would really appreciate some help or pointers here.

**Other doc prs**
https://github.com/ReactiveX/rxjs/pull/2322 - here is more background what do I do and why
https://github.com/ReactiveX/rxjs/pull/2326

### DIFF
--- a/src/observable/combineLatest.ts
+++ b/src/observable/combineLatest.ts
@@ -39,14 +39,73 @@ export function combineLatest<R>(...observables: Array<ObservableInput<any> | ((
  * <img src="./img/combineLatest.png" width="100%">
  *
  * `combineLatest` combines the values from all the Observables passed as
- * arguments. This is done by subscribing to each Observable, in order, and
- * collecting an array of each of the most recent values any time any of the
- * input Observables emits, then either taking that array and passing it as
- * arguments to an optional `project` function and emitting the return value of
- * that, or just emitting the array of recent values directly if there is no
- * `project` function.
+ * arguments. This is done by subscribing to each Observable in order and,
+ * whenever any Observable emits, collecting an array of the most recent
+ * values from each Observable. So if you pass `n` Observables to operator,
+ * returned Observable will always emit an array of `n` values, in order
+ * corresponding to order of passed Observables (value from the first Observable
+ * on the first place and so on).
  *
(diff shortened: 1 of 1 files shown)


jayphelps:
I'm going review this shortly. btw you might reference the reactivex docs as they usually contain excellent operator descriptions since it's mostly a sum of all Rx implementations. http://reactivex.io/documentation/operators/combinelatest.html

Just note that, while rare, RxJS sometimes has differing semantics of certain operators.

src/observable/combineLatest.ts
```diff
@@ -39,14 +39,73 @@ export function combineLatest<R>(...observables: Array<ObservableInput<any> | ((
...
- *
- * @example <caption>Dynamically calculate the Body-Mass Index from an Observable of weight and one for height</caption>
+ * arguments. This is done by subscribing to each Observable in order and,
+ * whenever any Observable emits, collecting an array of the most recent
+ * values from each Observable. So if you pass `n` Observables to operator,
+ * returned Observable will always emit an array of `n` values, in order
+ * corresponding to order of passed Observables (value from the first Observable
+ * on the first place and so on).
+ *
+ * Static version of `combineLatest` accepts either an array of Observables
```
blesh: Shouldn't the static version of combineLatest be documented along side of that code? This file is for the member method.

Podlas29: Are you sure? This is `observable` directory, where are all functions that land on `Observable`. Unless we understand `static` differently?

staltz: @blesh this JSDoc is for the static, as you can see down there we have the tag `@static true` which instructs ESDoc when generating the site.

blesh: @staltz @Podlas29 right you are. 👍

PR #2357, opened 2017-02-10
feat(webSocket): Add binaryType to config object

### LINKED ISSUE #2353: Websocket binaryType setting
[Websockets](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) have a property called `binaryType` that takes a value of `blob` (default) or `arraybuffer`. In my app, I have to change it to `arraybuffer`, but there is no simple mechanism in `WebSocketSubject` to do this.

For now, I'm setting the property once the socket has been opened. For instance:

```
  let subject = Observable.webSocket({
    url: '',
    openObserver: {
      next: e => { e.currentTarget.binaryType = 'arraybuffer' }
    }
  })
```

This seems to work, but I'm worried there could be a race condition where data is sent or received before the `binaryType` is set. When this value is not set correctly, `TypeError: First argument to DataView constructor must be an ArrayBuffer` is thrown.

I propose that a `binaryType` property be accepted in the config object passed to `Observable.webSocket`. It will immediately set the property on the socket instance as soon as it is created:

```
        try {
            socket = this.protocol ?
                new WebSocketCtor(this.url, this.protocol) :
                new WebSocketCtor(this.url);
            this.socket = socket;
            if (this.binaryType) {
                this.socket.binaryType = this.binaryType
            }
        }
```

Thoughts?

Podlas29:
**Description**
Add `binaryType` to websocket config object, so that it is possible to set that parameter on underlying socket before any data emits happen.

**Related**
Closes #2353

### DIFF
--- a/spec/observables/dom/webSocket-spec.ts
+++ b/spec/observables/dom/webSocket-spec.ts
@@ -261,6 +261,20 @@ describe('Observable.webSocket', () => {
       subject.unsubscribe();
     });
 
+    it('should take a binaryType and set it properly on the web socket', () => {
+      const subject = Observable.webSocket({
+        url: 'ws://mysocket',
+        binaryType: 'blob'
+      });
+
+      subject.subscribe();
+
+      const socket = MockWebSocket.lastSocket;
+      expect(socket.binaryType).to.equal('blob');
+
+      subject.unsubscribe();
+    });
+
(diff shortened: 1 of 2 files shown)


jayphelps:
@Podlas29 are you on Twitter?

Issue #2355, opened 2017-02-10
ChildSubscriptions unsubscribed but not removed.


johncvrs:
In the <code>switch()</code> operator, you'll find this code used to add
subscriptions to the <code>Subscriptions._subscriptions[]</code> array:

    this.add(this.innerSubscription = subscribeToResult(this, value));

And you'll find this code to remove them:

    const innerSubscription = this.innerSubscription;
    if (innerSubscription) {
      innerSubscription.unsubscribe();
      this.remove(innerSubscription);
    }

But note behavior of Subscription.add(s) is not "add s to the array and return s"; 
behavior is "create a new ChildSubscription from s, add the ChildSubscription to the array,
and return the ChildSubscription."
The consequence?  The remove() does not match the add().  The remove() becomes a silent no-op.  
The unsubscribe() was successful (good - we get correct program behavior)
but the remove() fails (bad - cannot GC these silent ChildSubscriptions).

**RxJS version:**
v5.1.0

**Code to reproduce:**

    var parentTicker = Rx.Observable.interval(1000);
    var childTickers = parentTicker.map((x) => {
      return Rx.Observable.interval(100).take(2);
    });

    var switcher = childTickers.switch();
    var sub = switcher.subscribe(function(x){
            console.log('tick #' + x + ': subscriptions = ' +
              sub._subscriptions[0]._innerSub._subscriptions.length);
          });

http://jsbin.com/zotobo/edit?js,console

**Additional information:**

PRs with specs and fixes for two operators (switch(), throttle())
will follow this message.  Take a look and tell me if you agree 
that this is a legit problem and not failed understanding on my part.

( Because if you do agree, this problem also exists in many
operators.  Operators audit(), switchMap(), switchMapTo(), windowToggle(),
windowWhen() can be fixed as above, but fixing MergeMap() e.g., is going to
require stepping back and deciding if remove() responsibility should
even live in the operators, or if it should live in the Subscription class
or ChildSubscription class rolled into the unsubscribe methods. )

johncvrs:
Oh, should have included the fix in this message, not just the PR.  For the <code>switch</code> and <code>toggle</code> operators the fix is simply:

Before

    this.add(this.innerSubscription = subscribeToResult(this, value));

After
    
    this.innerSubscription = this.add(subscribeToResult(this, value));

And with this in place, this line

    this.remove(innerSubscription);

is not even needed (ChildSubscriptions remove themselves on unsubscribe).

[...]

PR #2358, opened 2017-02-10
fix(switch): stop leaking ChildSubscriptions

johncvrs:
**Description:**

For the switch operator, fix issue of ChildSubscriptions not removed, not GCd. (#2355)

### DIFF
--- a/spec/operators/switch-spec.ts
+++ b/spec/operators/switch-spec.ts
@@ -223,4 +223,44 @@ describe('Observable.prototype.switch', () => {
 
     expect(completed).to.be.true;
   });
+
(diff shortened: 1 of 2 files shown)


trxcllnt:
@johncvrs Now that we've merged #2360 and published v5.1.1, we've covered this case, and all other operators that are similarly affected. I'm going to close this so people don't think this is still an issue. Thanks again for reporting #2355 and taking time out to submit this PR.

johncvrs:
I'm still digesting, but I'm in agreement: the fixes have removed the memory leaks and the run-time behavior is correct, and issue should be closed.   Thanks for the quick turn-around.

PR #2359, opened 2017-02-11
perf(bufferCount): optimize bufferCount operator

martinsik:
The `bufferCount()` operator can be used as `bufferCount(5, 3)` or `bufferCount(5)` (= `bufferCount(5, 5)`.

My assumption is that the second use-case `bufferCount(5)` is very common (maybe even more common that the first one) and can be handled in an simpler way. This means instead of keeping an array of buffers `T[][]` we can have just a single buffer `T[]`. This simplifies the subscriber class significantly and runs faster.

For this reason I split the original subscriber into two classes `BufferSkipCountSubscriber` and `BufferCountSubscriber` each handling one use-case.

**Performance of `bufferCount(5)`:**

Before:

```
$ node perf/micro buffercount 
                                         |                     RxJS 4.1.0 |                     RxJS 5.1.0 |          factor |      % improved
                 buffercount - immediate |                13,277 (±1.39%) |               308,516 (±2.25%) |          23.24x |        2,223.6%
                             buffercount |                 9,046 (±2.52%) |               143,636 (±2.19%) |          15.88x |        1,487.9%
```

After:

```
$ node perf/micro buffercount 
                                         |                     RxJS 4.1.0 |                     RxJS 5.1.0 |          factor |      % improved
                 buffercount - immediate |                13,205 (±1.02%) |               525,996 (±1.84%) |          39.83x |        3,883.2%
                             buffercount |                 9,330 (±2.55%) |               204,174 (±2.12%) |          21.88x |        2,088.4%
```

This improves performance on the `buffercount - immediate` script roughly from `23.24x` to `39.83x` and `buffercount` from `15.88x` to `21.88x`.

Performance of `bufferCount(5, 3)` remains the same. I could just remove one additional condition that wasn't necessary any more when we have two separate classes now.

I also tried to make an implementations using only fixed size arrays for both use-cases but these always turned out to be less efficient than the current implementation. It required calculating indices on every `next()`, at least one additional `if` condition and calling `Array.split()` to clone the buffer when calling `destination.next(...)`. Using `Array.push()` without `Array.split()` was always faster.

### DIFF
--- a/src/operator/bufferCount.ts
+++ b/src/operator/bufferCount.ts
@@ -110,9 +110,10 @@ class BufferSkipCountSubscriber<T> extends Subscriber<T> {
   }
 
   protected _next(value: T): void {
-    const { bufferSize, startBufferEvery, buffers } = this;
+    const { bufferSize, startBufferEvery, buffers, count } = this;
 
-    if (this.count++ % startBufferEvery === 0) {
+    this.count++;
+    if (count % startBufferEvery === 0) {
       buffers.push([]);
     }
 
--- a/src/operator/bufferCount.ts
+++ b/src/operator/bufferCount.ts
@@ -48,11 +48,18 @@ export function bufferCount<T>(this: Observable<T>, bufferSize: number, startBuf
 }
 
 class BufferCountOperator<T> implements Operator<T, T[]> {
+  private subscriberClass: any;
+
   constructor(private bufferSize: number, private startBufferEvery: number) {
+    if (!startBufferEvery || bufferSize === startBufferEvery) {
+      this.subscriberClass = BufferCountSubscriber;
+    } else {
+      this.subscriberClass = BufferSkipCountSubscriber;
+    }
   }
 
   call(subscriber: Subscriber<T[]>, source: any): any {
-    return source.subscribe(new BufferCountSubscriber(subscriber, this.bufferSize, this.startBufferEvery));
+    return source.subscribe(new this.subscriberClass(subscriber, this.bufferSize, this.startBufferEvery));
   }
 }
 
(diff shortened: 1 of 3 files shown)


trxcllnt:
Yay! Great work, thanks a lot @martinsik

src/operator/bufferCount.ts
```diff
@@ -83,20 +121,21 @@ class BufferCountSubscriber<T> extends Subscriber<T> {
...
-  protected _complete() {
-    const destination = this.destination;
+  protected _complete(): void {
     const buffers = this.buffers;
+
     while (buffers.length > 0) {
       let buffer = buffers.shift();
       if (buffer.length > 0) {
-        destination.next(buffer);
+        this.destination.next(buffer);
```
jayphelps: `destination` lookup used to be cached above, so it doesn't need to happen every time. Seemed like a simple win for micro perf?

src/operator/bufferCount.ts
```diff
@@ -62,19 +70,49 @@ class BufferCountOperator<T> implements Operator<T, T[]> {
...
 
-  protected _next(value: T) {
-    const count = this.count++;
-    const { destination, bufferSize, startBufferEvery, buffers } = this;
-    const startOn = (startBufferEvery == null) ? bufferSize : startBufferEvery;
+  protected _next(value: T): void {
+    const { bufferSize, startBufferEvery, buffers } = this;
 
-    if (count % startOn === 0) {
+    if (this.count++ % startBufferEvery === 0) {
```
jayphelps: I would prefer if we don't increment inside a conditional. I understand it just fine, but I've seen it be unnecessarily confusing for many. Can we increment before we enter the conditional?

src/operator/bufferCount.ts
martinsik: @jayphelps I removed the `const` because `this.destination` is used just once in this method. Maybe I don't understand what you mean.

src/operator/bufferCount.ts
martinsik: @jayphelps done!

src/operator/bufferCount.ts
jayphelps: @martinsik this is inside a loop, is it not possible for the branch inside to call destination.next() multiple times? Seems like it would be.

blesh: @jayphelps is correct here, you'll want to access the property outside of the loop and put in in a `const`

benlesh:
Thank you @martinsik!

PR #2360, opened 2017-02-12
fix(Subscription): fold ChildSubscription logic into Subscriber to prevent operators from leaking ChildSubscriptions.

### LINKED ISSUE #2244: Memory Leak with Queue Scheduler and Subject

### LINKED ISSUE #2355: ChildSubscriptions unsubscribed but not removed.

trxcllnt:
The addition of ChildSubscription to fix #2244 accidentally introduced a different memory leak. Most operators that add and remove inner Subscriptions store the inner Subscriber instance, not the value returned by Subscription#add. When they try to remove the inner Subscription manually, nothing is removed, because the ChildSubscription wrapper instance is the one added to the subscriptions list.

Fixes #2355, and supersedes the fix in #2358. Thanks again to @johncvrs for reporting this.

### DIFF
--- a/src/Subscription.ts
+++ b/src/Subscription.ts
@@ -40,7 +40,9 @@ export class Subscription implements ISubscription {
    */
   public closed: boolean = false;
 
-  private _subscriptions: ISubscription[];
+  protected _parent: Subscription = null;
+  protected _parents: Subscription[] = null;
+  private _subscriptions: ISubscription[] = null;
 
   /**
    * @param {function(): void} [unsubscribe] A function describing how to
@@ -66,11 +68,26 @@ export class Subscription implements ISubscription {
       return;
     }
 
+    let { _parent, _parents, _unsubscribe, _subscriptions } = (<any> this);
+
     this.closed = true;
+    this._parent = null;
+    this._parents = null;
+    // null out _subscriptions first so any child subscriptions that attempt
+    // to remove themselves from this subscription will noop
+    this._subscriptions = null;
 
-    const { _unsubscribe, _subscriptions } = (<any> this);
+    let index = -1;
+    let len = _parents ? _parents.length : 0;
 
-    (<any> this)._subscriptions = null;
+    // if this._parent is null, then so is this._parents, and we
+    // don't have to remove ourselves from any parent subscriptions.
+    while (_parent) {
+      _parent.remove(this);
+      // if this._parents is null or index >= len,
+      // then _parent is set to null, and the loop exits
+      _parent = ++index < len && _parents[index] || null;
+    }
 
     if (isFunction(_unsubscribe)) {
       let trial = tryCatch(_unsubscribe).call(this);
@@ -85,8 +102,8 @@ export class Subscription implements ISubscription {
 
     if (isArray(_subscriptions)) {
 
-      let index = -1;
-      const len = _subscriptions.length;
+      index = -1;
+      len = _subscriptions.length;
 
       while (++index < len) {
         const sub = _subscriptions[index];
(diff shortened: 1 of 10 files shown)


src/Subscription.ts
```diff
@@ -66,11 +67,25 @@ export class Subscription implements ISubscription {
...
 
-    (<any> this)._subscriptions = null;
+    this.closed = true;
+    this._parent = null;
+    this._parents = null;
+    // null out _subscriptions first so any child subscriptions that attempt
+    // to remove themselves from this subscription will noop
+    this._subscriptions = null;
+
+    let index = -1, len = _parents && _parents.length || 0;
```
jayphelps: IMO we should prefer multiline declarations for clarity--that's just my opinion tho.

src/Subscription.ts
```diff
@@ -66,11 +67,25 @@ export class Subscription implements ISubscription {
...
+
+    // if this._parent is null, then so is this._parents, and we
+    // don't have to remove ourselves from any parent subscriptions.
+    while (_parent) {
+      _parent.remove(this);
+      // if this._parents is null or index >= len,
+      // then _parent is set to false, and the loop exits
+      _parent = ++index < len && _parents[index];
```
jayphelps: I think this is harder to grok than neccesary--shouldn't need a comment that describes what it does.

src/Subscription.ts
```diff
@@ -168,37 +189,25 @@ export class Subscription implements ISubscription {
...
 
-  _unsubscribe() {
-    const { _innerSub, _parent } = this;
-    _parent.remove(this);
-    _innerSub.unsubscribe();
+  private _addParent(parent: Subscription) {
+    let { _parent, _parents } = this;
+    if (!_parent) {
+      this._parent = parent;
+    } else if (_parents || (_parents = this._parents = [])) {
```
jayphelps: Can we move this assignment out of the conditional and into the block instead?

blesh:
Pending @jayphelps' requests, it LGTM

src/Subscription.ts
jayphelps: Something like
```js
 } else {
  if (!_parents) {
    _parents = this._parents = [];
  }
  _parents.push(this);
 }
```

I get that these checks are annoying, perhaps there's a better place to set defaults.

blesh:
The coveralls change looks like unfortunate math. This fix is too critical to block over a 0.01% coverage change. I'm going to go ahead and merge.

PR #2362, opened 2017-02-12
refactor(*): strict null checks compiler option

mpodlasin:
Let's make TS compiler even more annoying ;)

I made it as an experiment to see how much havoc in codebase this option would create. Since it turned out fairly good, I am submitting it for debate.

Some conclusions:
* `subscribeToResult` returns `null` when it does not recognize `result` as acceptable input. This has consequences in entire codebase. We should decide how operators which accept `ObservableInput` should behave when user passes something that can't be treated as observable.
* Whenever I was unable to justify that some variable is defined but I used `!` non-null assertion operator nevertheless, I left `TODO` comment. Sometimes someone who knows codebase better will be able to provide such justification, but sometimes it will mean that some refactor is needed. 
* It is up to debate if `!` operator should be (always/sometimes) accompanied with comment why we can safely use this operator. Some ot them are obvious, like in while loops where we check if an array is non empty and then we `shift` that array.

Closes: #2066


kwonoj:
I am marking this as blocked for now. Original discussion issue #2066 hasn't discussed in further and we haven't get concrete conclusion yet, especially not having guarantees for any regressions. In other codebases I already though, I'm feeling this still might need to be discussed more and figure out way to prevent regression before introduce this changes.

Podlas29:
Cool. 
Just as a side note however, if you go through PR, you will see that almost no external api definitions were changed.
It would be cool to test this PR on existing codebases and see what happend in practice. ;)

kwonoj:
>  almost no external api definitions were changed.

: yes, that was assumption in original issue and also assumed this'll be probably OK, but just giving extra cautious to avoid unexpected side effect just occurs with release. There was few actual cases in beta phase in-dev-built works fine but packaged, published one causes issues.

blesh:
I'm with @kwonoj on this one, it requires discussion.

saneyuki:
> I'm with @kwonoj on this one, it requires discussion.

@benlesh @kwonoj @mpodlasin  

TypeScript 2.3's strict null checks compiler option works well stably.
I seem we can do this change on `next` branch safely. It's a time to retake this.

kwonoj:
we're not talking about compiler stability but codewise regression.

saneyuki:
About codewise regression, I feel it would not be a problem if this pull request does not change a part of testcase.

If we'll find a regression after merging this, but there was not a unit test. It means we could not ensure/stabilize the correctness about its part until now for each of releases. It's hard to detect for us.

kwonoj:
>  I feel it would not be a problem if this pull request does not change a part of testcase.

No, our test case doesn't fully verify type definition of our codebase. We already experienced regression after publishing release breaks type which test cases couldn't caught it.

saneyuki:
@kwonoj 

> No, our test case doesn't fully verify type definition of our codebase.  We already experienced regression after publishing release breaks type which test cases couldn't caught it.

If you say so, I strongly think we should land this change to our _next_ branch and ship it as _alpha_ release. _I don't claim to land to master as stable release_.

We have a branch to experiment for future release to collect feedback. We have a great chance to use it.

staltz:
Thumbs up to strict null checks, but it doesn't need to be turned on for tests.

benlesh:
This PR is stale, and I believe we already have this done anyhow.

Sorry for the long wait to have this closed, @mpodlasin... I really appreciate your contributions. Closing this one.

PR #2373, opened 2017-02-14
docs(ObservableInput): add ObservableInput and SubscribableOrPromise descriptions

Podlas29:
**Description:**

Add `ObservableInput` and `SubscribableOrPromise` interface descriptions, as well as link these interfaces in type descriptions of operators, so that users always know what kind of parameters they can pass to used methods.

**Related issues:**
#2326 
#2344 
#2351

### DIFF
--- a/src/MiscJSDoc.ts
+++ b/src/MiscJSDoc.ts
@@ -128,3 +128,221 @@ export class ObserverDoc<T> {
     return void 0;
   }
 }
+
+/**
+ * `SubscribableOrPromise` interface describes values that behave like either
+ * Observables or Promises. Every operator that accepts arguments annotated
+ * with this interface, can be also used with parameters that are not necessarily
+ * RxJS Observables.
+ *
+ * Following types of values might be passed to operators expecting this interface:
+ *
+ * ## Observable
+ *
+ * RxJS {@link Observable} instance.
+ *
+ * ## Observable-like (Subscribable)
+ *
+ * This might be any object that has `Symbol.observable` method. This method,
+ * when called, should return object with `subscribe` method on it, which should
+ * behave the same as RxJS `Observable.subscribe`.
+ *
+ * `Symbol.observable` is part of https://github.com/tc39/proposal-observable proposal.
+ * Since currently it is not supported natively, and every symbol is equal only to itself,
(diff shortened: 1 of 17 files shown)


src/MiscJSDoc.ts
```diff
@@ -128,3 +128,197 @@ export class ObserverDoc<T> {
...
+ */
+export class SubscribableOrPromiseDoc<T> {
+
+}
+
+/**
+ * `ObservableInput` interface describes all values that are observables,
+ * or can be treated as such. Every operator that accepts arguments annotated
+ * with this interface, can be also used with parameters that are not necessarily
+ * RxJS Observables.
```
jayphelps: The two descriptions for `SubscribableOrPromise` and `ObservableInput` seem to start with the same statement but not entirely accurate.

>  X interface describes all values that are observables, or can be treated as such.

src/MiscJSDoc.ts
```diff
@@ -128,3 +128,197 @@ export class ObserverDoc<T> {
...
+ * [DOM NodeList](https://developer.mozilla.org/pl/docs/Web/API/NodeList),
+ * or, actually, any object that has `length` property (which is a number)
+ * and stores values under non-negative (zero and up) integers.
+ *
+ * ## ES6 Iterable
+ *
+ * Operators will accept both built-in and custom ES6 Iterables, by treating them as
+ * observables that emit all its values in order of iteration and then complete
+ * when iteration ends. Note that contrary to arrays, Iterables do not have to
+ * necessarily be finite, so creating Observables that never complete is possible as well.
```
jayphelps: The infinite iterator use case is not well known but very powerful and easy with generators. I think an example that uses generators would be very very useful.

```js
// infinite stream of incrementing numbers
const infinite = function* () {
  let i = 0;

  while (true) {
    yield i++;
  }
};

Rx.Observable.from(infinite())
  .take(3) // only take 3, cause this is infinite
  .subscribe(
    value => console.log(value),
    err => {},
    () => console.log('ta dam!')
  );
// 0
// 1
// 2
```

Podlas29:
@jayphelps I made the changes you requested.

PR #2377, opened 2017-02-16
chore(danger): warn only for non conventional changelog

kwonoj:
**Description:**

This PR lower failure checker for conventional change log commit message into warning, to avoid checker fails due to commit like merge commits.


kwonoj:
Checking in myself since this is trivial change.

Issue #2379, opened 2017-02-16
AnimationFrameScheduler - shouldn't it requestAnimationFrame also when scheduled with delay?


mpodlasin:
Hi.

Currently when scheduled with delay, `AnimationFrameScheduler` delegates work to `AsyncScheduler`. So for example if I use it with `Observable.interval`, `window.requestAnimationFrame` will not be used to schedule interval ticks.

But what if I want to make an animation where - for example - every second div changes position randomly? Wouldn't then using  `window.requestAnimationFrame` for delayed tasks make that animation more smooth? 

Shouldn't `AnimationFrameScheduler` delay given task asynchronously but then - when it is time to do that task - delay it a little bit more with  `window.requestAnimationFrame`?

Answers, links to answers and links to links to answers are all welcome.

Thanks in advance and have a great day.

trxcllnt:
@Podlas29 great questions! Let me start by saying the Schedulers are by no means sacred. Many decisions were made early on either because that's how they worked in <= v4, or to minimize surprises. The answer to your question is a mix of the two, and applies equally to the ASAP scheduler. As a side note, if you use `Observable.interval(0, Scheduler.animationFrame)`, the interval _will_ be scheduled via rAF, so you can build smooth animations.

Whenever possible, Schedulers should delegate as much of the timing control to the platform, as the platform will take into account variables the Schedulers can't know. This is why we've opted to use `setInterval` instead of `setTimeout`, even in the case the interval is canceled after the first execution. Assuming operators clean up after themselves, this leads to more consistent timer synchronization than we could get from concurrent pending setTimeouts.

With respect to the principle of least surprise, unlike setTimeout/setInterval, neither setImmediate nor requestAnimationFrame accept a due-time parameter. We relied on this omission as guidance, and opted not to implement custom timing logic on top of these methods, which would effectively redefine the behavior of these methods for Rx users. Implementing due times with serially scheduled next ticks/animation frames would necessitate scheduling as many ticks as necessary until the due time, which isn't 

I do see a potential alternative to what we have today, where scheduling on the asap/animationFrame scheduler with a due time > 0 would schedule a timeout for the due time, and follow-up with a setImmediate/rAF when the timeout resolves.

trxcllnt:
[...]

PR #2380, opened 2017-02-16
docs(Scheduler): add docs for async, asap, queue and animationFrame

Podlas29:
**Description:**

Schedulers are probably the most mysterious for newcomers (they were for me and in a way still are).

I believe it would help to show some examples of using schedulers without operators, so that users can
have good intuition what given scheduler can achieve.

I have put descriptions next to already created Schedulers, since these are the ones that should be used primarily.

**Problems:**

* I can not make `{@link async}` in docs, since there is already `async` property on `AjaxRequest` interface. 
* Schedulers and `root` are in the same html file, because they are all variables. Is there a known way to separate them?

**Related issues:**
https://github.com/ReactiveX/rxjs/issues/2379 - question about animationFrame scheduler

### DIFF
--- a/src/scheduler/animationFrame.ts
+++ b/src/scheduler/animationFrame.ts
@@ -1,4 +1,35 @@
 import { AnimationFrameAction } from './AnimationFrameAction';
 import { AnimationFrameScheduler } from './AnimationFrameScheduler';
 
+/**
+ *
+ * Animation Frame Scheduler
+ *
+ * <span class="informal">Perform task when `window.requestAnimationFrame` would fire</span>
+ *
+ * When `animationFrame` scheduler is used with delay, it will fall back to {@link async} scheduler
+ * behaviour.
+ *
+ * Without delay, `animationFrame` scheduler can be used to create smooth browser animations.
+ * It makes sure scheduled task will happen just before next browser content repaint,
+ * thus performing animations as efficiently as possible.
+ *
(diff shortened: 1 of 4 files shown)


staltz:
@Podlas29 Thank you.

To make it easier to predict how this will look like in the final generated website, could you please host the generated site from your local machine, perhaps to some easy service like http://surge.sh/ or https://zeit.co/now/ ?

Podlas29:
@staltz http://rxjs-docs-prototype.surge.sh/ here you go!

src/scheduler/animationFrame.ts
```diff
@@ -1,4 +1,32 @@
...
+ * Without delay, `animationFrame` scheduler can be used to create smooth browser animations.
+ * It makes sure scheduled task will happen just before next browser content repaint,
+ * thus performing animations as efficiently as possible.
+ *
+ * @example <caption>Schedule div height animation</caption>
+ * const div = document.querySelector('#some-div');
+ *
+ * Rx.Scheduler.schedule(function(height) {
+ *   div.style.height = height + "px"
+ *
```
blesh: Perhaps add a comment here explaining that `this` is a reference to the scheduled `Action`?

blesh:
Looks great, as always, @Podlas29! I've only made one request, but other than that, it looks really good!

Podlas29:
@blesh thanks. I have made changes you requested.

Issue #1776, opened 2016-06-16
Discuss: stop publishing system bundle 


jeffbcross:
The original reason I added the systemjs bundle was because angular 2 was publishing system bundles in our docs, and was distributing them in our npm package. Now we only publish umd bundles alongside our commonjs and es6 modules. Since maintaining the system bundles takes some work, I recommend stop publishing system bundles unless others in the community are depending on them.

blesh:
This sounds good to me. Any objections? I'll leave this open for a week or so to find out.

kwonoj:
Just curiosity, (since I'm not familiar with SystemJS) after stop publishing systemjs bundle, what would be needed if anyone would like to use Rx as SystemJS?

robwormald:
@kwonoj it would just mean using the modules individually, vs a big bundle blob of code... SystemJS will load basically any format (ES6, CJS, etc...)

kwonoj:
got it, thanks for explanation.

kwonoj:
Closing via #1781 /

Issue #1725, opened 2016-05-23
SystemJs Builder - Error calculate canonical name


meriturva:
We are trying to build our solution with SystemJs Builder without success.
RxJs version: 5.0.0-beta.6

We have problems and we had already some help from SystemJs guy: https://github.com/systemjs/builder/issues/587#issuecomment-220835913

We have developed a simple project to test build process:

https://github.com/meriturva/SystemJsBuilder-587

Just clone it and run `gulp build` and you will see problems related to RxJs package:

```
error { [Error: Error tracing ./node_modules/rxjs/Subject.js at file:///H:/Projects/github.com/SystemJsBuilder-587/node_modules/rxjs/Subject.js
        Loading ./node_modules/@angular/core/core.umd.js
        Loading app/main.js
        Unable to calculate canonical name to bundle file:///H:/Projects/github.com/SystemJsBuilder-587/node_modules/rxjs/Observable. Ensure that this module sits within the baseURL or a wildcard path config.]
  originalErr: [Error: Unable to calculate canonical name to bundle file:///H:/Projects/github.com/SystemJsBuilder-587/node_modules/rxjs/Observable. Ensure that this module sits within the baseURL or a wildcard path config.] }
```

We also would like to use umd version but we need WebSocket class that it is not already included https://github.com/ReactiveX/rxjs/issues/1506

Any ideas how to solve it?
Thanks.
Diego

blesh:
Related PR #2381

benlesh:
I don't think this is any longer an issue.

PR #2381, opened 2017-02-17
Link an externally available System bundle repackaging in the README

kylecordes:
**Description:**

In the past the RxJS project published a System bundle, but decided not to continue doing so.  Studying the discussion of that decision, there did not appear to be a strong motivation against the availability of such a bundle (and indeed it is very convenient for consumption via a Plunkr or any other contacts without a build process), avoiding the numerous separate requests that otherwise happen when using System.

Therefore, this adds links to a parallel package I have created to publish the System bundles.

https://github.com/ReactiveX/rxjs/issues/1776

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,9 +120,20 @@ npm install @reactivex/rxjs@5.0.0
 
 ### CDN
 
-For CDN, you can use [unpkg](https://unpkg.com/):
-  
+For CDN, you can use [unpkg](https://unpkg.com/) to retrieve the single-file UMD
+bundle:
+
 https://unpkg.com/rxjs/bundles/Rx.min.js
+
+Alternatively, the following package publishes a single-file System bundle for
+easy consumption via Plunkr. This bundle makes it possible to consume the
+various sub-modules with a single network request:
+
+https://www.npmjs.com/package/rxjs-system-bundle
+
+And is also available via unpkg:
+
+https://unpkg.com/rxjs-system-bundle@5.1.1/Rx.system.min.js
 
 #### Node.js Usage:
 


kwonoj:
So am I understanding correct that `rxjs-system-bundle` is published by you & maintained by you, and this documentation PR is refer it into upstream?

kylecordes:
@kwonoj  Yes, is correct. In order of preference, the situation for consumers of rxjs is somewhere on this spectrum:

1) The core project ships a system bundle. However, this has already been discussed in the issue tracker and the team has decided to remove it. So I felt that putting in a PR to add it back in would be rude.
2) An external project ships a matching system bundle, linked from the main project README.
3) An external project ships a matching system bundle, but it is not linked.
4) There is no publicly available system bundle at all, everyone who wants one has to bundle their own.

As far as I know, until today 4 was the status quo. I created this separate project, upgrading to situation 3. This pull request would create situation 2. 1 was already considered and rejected.

Obviously whether or not this PR is accepted will not affect me at all: I already know about my published bundle :-) but I think it would be helpful for other folks who occasionally express irritation at the lack of a published System bundle.

kylecordes:
@kwonoj Also, I would be completely thrilled to hand over ownership of this repo and package to the core team, if someone there wants to maintain it instead. You can read it in a minute or two, all it does is use system-builder to create the bundle, then a little machinery to make the versions match.

trxcllnt:
SystemJS should be a build target. It's not difficult to maintain, and makes people's lives easier if we do.

blesh:
I'm not comfortable linking externally maintained builds in the README. If this was something we maintained, I'd say okay.. but it's a :-1: from me.

kylecordes:
I am the author of this PR and I actually agree with @blesh - but I would love to have any outcome where the system build is available and discoverable from here.  Rather than just floating out there were someone would have to think to Google for it. Would love to have the rxjs team take this from me and run. Or for that matter I am happy to tinker with maintaining it and have the package owned by the rxjs team.  Or any variation thereof.

kwonoj:
So in my point of view, this PR contains few things, 

- Is it ok to refer externally maintained pkg in the documentation?
: For this matter, I'm on the same page it would better not to. It gives implicit assumption pkg is maintained same way as other pkgs.

- Is it ok to someone externally publish non-supported module packages?
: Maybe debatable, but I don't think this is an issue.

- Does RxJS need to publish SystemJS bundle as package?
: I believe this is crux of issue. Main reason we've stopped publishing some type of modules like native ES or SystemJS is there (looks) aren't sufficient numbers of module consumers but also it causes confusions between default CJS module packages, while it doesn't give huge benefits. It'd worth to rediscuss if we'd expand module publishing as same as discussing native ES module, with few questions like what is benefit for publishing new module based package? like, what's actual issue SystemJS user while they're using either CJS or UMD bundle? need to be clarified. I am not daily SystemJS user, so maybe I'm underestimated difficulties of SystemJS user experiencing with current packages.

blesh:
this issue seems related: https://github.com/ReactiveX/rxjs/issues/1725

blesh:
What would it take to have this build target and add it to our published package(s)?

kwonoj:
@kylecordes appreciate for this PR for taking effort for systemJS users. I don't think we'll going to direct external dependency in documentation, so I'm going to close this PR itself for now - and we can continue discussion in #1725 (or if separate issue needed, feel freely create one) to discuss if separate systemJS pkg is needed or either some configuration can resolve this.

kylecordes:
@kwonoj I'm not sure that #1725 from a year ago is about this at all, maybe a mis-paste by Ben?

I will make a fresh issue about the general question of shipping a system bundle.

kylecordes:
New item, more clearly about a desire for a System bundle: #2381

kwonoj:
@kylecordes no, I don't think #1725 is mispasted. It's issue regarding systemJS consumers, and not sure if it can't be solved by some configuration or either systemjs build need to be generated.

kylecordes:
@kwonoj Ah, I see. I would be happy to add some dock showing how to properly consume RxJS from SystemJS. The problem is, the ideal way to do that is with a system bundle, the subject of this item and the new one :-)

PR #2382, opened 2017-02-17
fix(reduce): fix overload ordering

### LINKED ISSUE #2338: Reduce logs as array, but erroring as object
**RxJS version:**
5.1.0

**Code to reproduce:**
```typescript
this.searchSubscription = this.filtersSubject
  .takeLast(2)
  .reduce((acc, cur) => {
    console.log(acc);
    acc.push(cur);
    return acc;
  }, [])
  .debounceTime(1000)
  .subscribe(x => this.emitFilters());
```

**Error:**
Code fails to compile because error claiming acc is an object (thus, no push method), even though the console.log shows acc to be an array.

**Additional information:**
I was told by [this StackOverflow answer](http://stackoverflow.com/questions/42080580/rxjs-reduce-logs-as-array-but-erroring-as-object#42080580) that this is a bug with RxJS? If I'm wrong, please ignore.

aluanhaddad:
**Description:**
Type inference for `Observable.prototype.reduce` currently behaves incorrectly, resulting in spurious type errors, in many common cases, when the type of `seed` is a different type than the value type of the Observable.

One of the declared overloads is intended to specifically cater to this scenario but fails to do so because of its ordering with respect to the other overload declared.

The overload in question takes two type arguments, `T` and `R`, making it more specific than the overloads declared before it.

Due to TypeScript's overload resolution algorithm, which picks the first matching overload, this overload does not have the intended effect.

This is documented here https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html in the section entitled **Function Overloads**, under the **Ordering** heading, which states

> Don’t put more general overloads before more specific overloads:

> Do sort overloads by putting the more general signatures after more specific signatures:

> Why: TypeScript chooses the first matching overload when resolving function calls. When an earlier overload is “more general” than a later one, the later one is effectively hidden and cannot be called.

This fixes #2338

### DIFF
--- a/spec/operators/reduce-spec.ts
+++ b/spec/operators/reduce-spec.ts
@@ -302,23 +302,84 @@ describe('Observable.prototype.reduce', () => {
 
   it('should accept T typed reducers', () => {
     type(() => {
-      let a: Rx.Observable<{ a?: number; b?: string }>;
-      a.reduce((acc, value) => {
+      let a: Rx.Observable<{ a: number; b: string }>;
+      const reduced = a.reduce((acc, value) => {
         value.a = acc.a;
         value.b = acc.b;
         return acc;
-      }, {});
+      });
+
+      reduced.subscribe(r => {
+        r.a.toExponential();
+        r.b.toLowerCase();
+      });
     });
   });
 
-  it('should accept R typed reducers', () => {
+  it('should accept R typed reducers when R is assignable to T', () => {
     type(() => {
-      let a: Rx.Observable<{ a: number; b: string }>;
-      a.reduce<{ a?: number; b?: string }>((acc, value) => {
+      let a: Rx.Observable<{ a?: number; b?: string }>;
+      const reduced = a.reduce((acc, value) => {
         value.a = acc.a;
         value.b = acc.b;
         return acc;
       }, {});
+
+      reduced.subscribe(r => {
+        r.a.toExponential();
+        r.b.toLowerCase();
+      });
+    });
+  });
+
+  it('should accept R typed reducers when R is not assignable to T', () => {
+    type(() => {
+      let a: Rx.Observable<{ a: number; b: string }>;
+      const seed = {
+        as: [1],
+        bs: ['a']
+      };
+      const reduced = a.reduce((acc, value) => {
+        acc.as.push(value.a);
+        acc.bs.push(value.b);
+        return acc;
+      }, seed);
+
+      reduced.subscribe(r => {
+        r.as[0].toExponential();
+        r.bs[0].toLowerCase();
+      });
+    });
+  });
+
+  it('should accept R typed reducers and reduce to type R', () => {
+    type(() => {
+      let a: Rx.Observable<{ a: number; b: string }>;
+      const reduced = a.reduce<{ a?: number; b?: string }>((acc, value) => {
+        value.a = acc.a;
+        value.b = acc.b;
+        return acc;
+      }, {});
+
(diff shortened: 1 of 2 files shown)


aluanhaddad:
@kwonoj Thank you, I most definitely will.

aluanhaddad:
It looks like there is more nuance here, the overload taking an array seems to be producing more issues than I had thought.

blesh:
@aluanhaddad it looks like there are build errors with the specs. Do you think you can fix them?

aluanhaddad:
Yes I've been very busy recently. I will try to fix them tonight and add some additional tests as well. When I looked at the specs, they seemed to rely on the seed being inferred to have the value type of the Observable. I need to double-check if that is the case.
Perhaps I misunderstand the behavior of RxJS or the intent of the Declarations but this doesn't seem like a reasonable assumption at face value.

aluanhaddad:
I am having some trouble replicating the test failures locally.
When I run
```bash
npm run build_spec && npm run test_mocha && node ./node_modules/markdown-doctest/bin/cmd.js
```
Any tips would be greatly appreciated.

aluanhaddad:
@kwonoj I added several new test including tests covering behavior that was an error before these changes.
@blesh I fixed the failing tests and also broke them down into multiple discrete tests, for assignability.

I added a local binding and a subsequent statement to each test I touched so that the resulting value is now used in such a way that it validates that the types are propagated.

Let me know if these tests are sufficient and also let me know if I should have added the same tests to `spec/operator/scan-spec`.

kwonoj:
I think this change is ready, re-approving. /cc @david-driscoll for second eye just in case.

spec/operators/reduce-spec.ts
```diff
@@ -302,23 +302,84 @@ describe('Observable.prototype.reduce', () => {
...
+      });
     });
   });
 
-  it('should accept R typed reducers', () => {
+  it('should accept R typed reducers and reduce to type R', () => {
     type(() => {
       let a: Rx.Observable<{ a: number; b: string }>;
-      a.reduce<{ a?: number; b?: string }>((acc, value) => {
+      const reduced = a.reduce<{ a?: number; b?: string }>((acc, value) => {
```
masaeedu: Is a separate test required for generic type argument inference of `R`?

masaeedu:
lgtm

spec/operators/reduce-spec.ts
aluanhaddad: I believe so. Failure to infer `R` as distinct from `T` and `T[]` was the original issue.

masaeedu: @aluanhaddad This test doesn't seem to be verifying correct inference, because it specifies the generic type arg explicitly. Probably for an inference test you'd repeat this without the explicit type argument, then try to assign `reduced` to a variable of the explicit type you expect to have been inferred.

masaeedu: My bad, just realized the "R is not assignable to T" test already covers inference. 👍

benlesh:
@david-driscoll can you review this please? <3

aluanhaddad:
Just to note, when I first looked at this, I believed the issue was simply due to the overload ordering but the root cause turned out to have as much to do with the optionality of the seed arguments as signature ordering. The now required seed arguments eliminate a previously applicable overload from the candidate set, but a minor order change was still ultimately necessary.

benlesh:
Thanks @aluanhaddad !!!

PR #2393, opened 2017-02-20
docs(*): add more descriptive @param definitions for callbacks

martinsik:
I updated a couple of `@param` tags to include more precise definitions for callback functions


blesh:
@martinsik, I believe if we do this it will break people using Closure compiler.

martinsik:
@blesh Ok, I thought it's the same as the already existing docblocks. For example here: https://github.com/ReactiveX/rxjs/blob/master/src/operator/mergeMap.ts#L52

benlesh:
cc @alexeagle  you might know... will this cause problems with Closure Compiler?

PR #2394, opened 2017-02-20
test(forkJoin): add tests with never and with unsubscribing after error

Podlas29:
* Add tests with Observable that never completes, so that it is clear how forkJoin operator behaves with it.
* Add test which shows that forkJoin unsubscribes from other Observables when one of them errors.

**Description:**

I am writing docs for `forkJoin` and felt bad making assertions about how it behaves, without having unit tests to back me up.

### DIFF
--- a/spec/observables/forkJoin-spec.ts
+++ b/spec/observables/forkJoin-spec.ts
@@ -194,6 +194,35 @@ describe('Observable.forkJoin', () => {
     expectObservable(e1).toBe(expected);
   });
 
+  it('should not complete when only source never completes', () => {
+    const e1 = Observable.forkJoin(
+      hot('--------------')
+    );
+    const expected = '-';
+
+    expectObservable(e1).toBe(expected);
+  });
+
+  it('should not complete when one of the sources never completes', () => {
+    const e1 = Observable.forkJoin(
+      hot('--------------'),
+      hot('-a---b--c--|')
+    );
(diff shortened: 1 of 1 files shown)


spec/observables/forkJoin-spec.ts
```diff
@@ -245,6 +274,15 @@ describe('Observable.forkJoin', () => {
     expectObservable(e1).toBe(expected);
   });
 
+  it('should raise error when any of source raises error with source that never completes', () => {
+    const e1 = Observable.forkJoin(
+      hot('------#'),
+      hot('----------'));
+    const expected = '------#';
+
```
blesh: Can we vertically align these marble strings? It's good to make the next developer see the visual alignment of the events.

spec/observables/forkJoin-spec.ts
```diff
@@ -194,6 +194,35 @@ describe('Observable.forkJoin', () => {
...
+    expectObservable(e1).toBe(expected);
+  });
+
+  it('should complete when one of the sources never completes but other completes without values', () => {
+    const e1 = Observable.forkJoin(
+      hot('--------------'),
+      hot('------|')
+    );
+    const expected = '------|';
```
blesh: Same as the other feedback... vertically align the marbles.

Podlas29:
@blesh done.

PR #2397, opened 2017-02-21
fix(ErrorObservable): remove type constraint to error value

### LINKED ISSUE #2395: Error Observable has incorrect signature
In master currently, ErrorObservable<T> has the following code: 

```
  static create<T>(error: T, scheduler?: IScheduler): ErrorObservable<T> {
    return new ErrorObservable(error, scheduler);
  }

  static dispatch(arg: DispatchArg) {
    const { error, subscriber } = arg;
    subscriber.error(error);
  }

  constructor(public error: T, private scheduler?: IScheduler) {
    super();
  }
```

I don't think any of the `error` parameters should be of type `T`.  They should be of type `Error` or most likely `any`.  The documentation even says this: 
`   * @param {any} error The particular Error to pass to the error notification.`

kwonoj:
**Description:**
This PR updates type signature of static creation method for `ErrorObservable` to no longer carry type constraint to value of error.

- closes #2395

### DIFF
--- a/src/observable/ErrorObservable.ts
+++ b/src/observable/ErrorObservable.ts
@@ -12,7 +12,7 @@ export interface DispatchArg {
  * @extends {Ignored}
  * @hide true
  */
-export class ErrorObservable<T> extends Observable<any> {
+export class ErrorObservable extends Observable<any> {
 
   /**
    * Creates an Observable that emits no items to the Observer and immediately
@@ -54,7 +54,7 @@ export class ErrorObservable<T> extends Observable<any> {
    * @name throw
    * @owner Observable
    */
-  static create<T>(error: T, scheduler?: IScheduler): ErrorObservable<T> {
+  static create(error: any, scheduler?: IScheduler): ErrorObservable {
     return new ErrorObservable(error, scheduler);
   }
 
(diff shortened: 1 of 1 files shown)


src/observable/ErrorObservable.ts
```diff
@@ -63,7 +63,7 @@ export class ErrorObservable<T> extends Observable<any> {
     subscriber.error(error);
   }
 
-  constructor(public error: T, private scheduler?: IScheduler) {
+  constructor(public readonly error: any, private readonly scheduler?: IScheduler) {
```
blesh: Is this not going to be a breaking change for some users? I feel like we've done this in the past, but I can't remember the verdict. @staltz? What are your thoughts?

For now, @kwonoj, can we exclude this change from this PR?

kwonoj: yes, we've done similar (https://github.com/ReactiveX/rxjs/pull/2249) and we've accepted to introduce those changes, since it brings explicit clarification for TS users.

kwonoj: Either case I've updated code as it can be easily created as separate PR.

rokerkony:
Hi guys,
after fresh `npm install` we have error as `Type 'ErrorObservable' is not generic.` with the code:
`protected handleError (error: Response): ErrorObservable<string> {`

Change is easy but should not be this marked as BC ? @benlesh

PR #2398, opened 2017-02-21
feat(Subject): implement hasObservers

kwonoj:
**Description:**
Related with https://github.com/ReactiveX/rxjs/pull/2279#discussion_r98346733.

I was originally under assumption of `Subject::observers` is publicly exposed as readonly to allow check Subject has observer or not, But since it's encapsulated, current implementation of subject does not have way to check if it has observers or not. This PR implements `Subject::hasObservers` corresponds to RxJS v4.

https://github.com/ReactiveX/rxjs/issues/2232

### DIFF
--- a/spec/Subject-spec.ts
+++ b/spec/Subject-spec.ts
@@ -473,6 +473,18 @@ describe('Subject', () => {
     expect(results).to.deep.equal(['a', error]);
   });
 
+  it('should return corresponding hasObserver value', () => {
+    const subject = new Rx.Subject();
+
+    expect(subject.hasObservers()).to.be.false;
+
+    subject.subscribe(() => {
+      //noop
+    });
+
+    expect (subject.hasObservers()).to.be.true;
+  });
+
   describe('asObservable', () => {
     it('should hide subject', () => {
       const subject = new Rx.Subject();
--- a/src/Subject.ts
+++ b/src/Subject.ts
@@ -43,6 +43,10 @@ export class Subject<T> extends Observable<T> implements ISubscription {
     return new AnonymousSubject<T>(destination, source);
   }
 
+  public hasObservers(): boolean {
+    return this.observers && this.observers.length > 0;
+  }
+
   lift<R>(operator: Operator<T, R>): Observable<T> {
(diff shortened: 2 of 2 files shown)


blesh:
@kwonoj ... I think for the purposes of *our* unit tests, we can just cast as <any>. I don't think this addition is necessary, as whether or not any Subject or Observable has observers is a very imperative API to expose that almost boils down to an implementation detail.

kwonoj:
@blesh so you're thinking v5 no longer need https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/subjects/subject.md#rxsubjectprototypehasobservers and shouldn't expose any details related with observers?

kwonoj:
To clarify: I'm indifferent of having this interfaces, but observed some cases ppl looking this in v5 since .net / Java / RxJS v4 binding has equivalent interfaces. (https://github.com/ReactiveX/rxjs/issues/2232, http://stackoverflow.com/questions/41410452/what-replaces-subject-prototype-hasobservers) and trying to make ppl to not to access observers directly for those purpose by lacks of interfaces. If there's recommended pattern (either replaces hasObservers or more better pattern do not require hasObservers) then I think it'll be fine. Just need to ensure `Subject::observers` is encapsulated and should not be accessed in general.

kwonoj:
@benlesh old doesn't mean this doesn't go away - v5, v6 both don't have equivalent of v4's interfaces, while it does seem like it's just not picked up and it's breaking changes. If we think those breaking change was intended we could close issue, but in that case we may need some documentation of how-to achieve same behavior without this interface. (or other reasoning why we dropped vice versa)

benlesh:
`hasObservers` was omitted for a few reasons: 

1. I'm not sure it existed at the time we started writing v5 (because then it wasn't "v5" it was really going to be "v3" haha).. 
2. Imperative examination of a Subject's state is a bit of a smell. 
3. It's not a feature anyone has asked for or missed that I'm aware of
4. I'm not sure what solid use cases it has.

@kwonoj can we close this?

kwonoj:
closing's fine but plz update migration docs at least.

PR #2399, opened 2017-02-21
fix(AjaxObservable): Use encodeURIComponent to encode request body

### LINKED ISSUE #2389: Ajax requests do not handle ampersand correctly
**RxJS version:** 5.1.0

**Code to reproduce:**
`Rx.ajax.post('my-api',{ url: 'bla&foo=bar' })`

**Expected behavior:**
Expected to send post request with a single parameter (*url*)

**Actual behavior:**
Sends post request with two parameters: *url* and *foo*

**Additional information:**
The problem could be resolved by setting `Content-Type` to `application/json` manually on each request, but presumably these requests should work out of the box.

ericponto:
**Description:**
The contentType 'application/x-www-form-urlencoded' was using `encodeURI` but should be using
`encodeURIComponent` on the request body.

closes #2389


blesh:
> The contentType 'application/x-www-form-urlencoded' was using encodeURI but should be using
encodeURIComponent on the request body.

@ericponto interesting. Can you please provide a link or additional information on why this change is required?

ericponto:
Sure! I looked at MDN before making the PR and it has pretty good info on the difference.

[encodeURI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI)

> Assumes that the URI is a complete URI, so does not encode reserved characters that have special meaning in the URI.

[encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)

> To avoid unexpected requests to the server, you should call encodeURIComponent on any user-entered parameters that will be passed as part of a URI. 

When encoding the request body params it is the latter use case since it is not a complete URI.

Additionally, I took at look at how [jQuery handles it](https://github.com/jquery/jquery/blob/master/src/serialize.js#L67) (since probably more ajax request have used jQuery than anything else), and it uses `$.param` on ajax data which uses `encodeURIComponent`.

gernsdorfer:
This PR is required because:
If a value of an Object contains a character like `&`, `encode` ignore this character.

Here i've the following problem:

BodyParameters: `{ name1: "value1&test" }`
Sended Parameters via http(post): 
`name1=value1`
`test=`

dfb1900:
Will this PR gonna be merged soon? I'm having the same issue.

benlesh:
I'll try to get this into 5.5.6

winghouchan:
This didn't make it into 5.5.6. Is there anything block this? If there are, is there anything that can be done? If there aren't, will it make it to the next release and will that be soon?

benlesh:
Something is really messed up with this PR. I'm going to close it and we can start another PR.

ericponto:
@benlesh I'll submit a new PR here in a bit.

sfishel18:
I'm still seeing this issue in 5.5.10.  Is there a way to know if/when it will end up in a 5.5.x release?  Thanks much!

PR #2400, opened 2017-02-22
docs(forkJoin): add documentation for the operator

mpodlasin:
**Description:**

* Add detailed description of forkJoin operator.
* Summarize what it accepts and returns.
* Provide links to similar operators.
* Provide examples of different ways it can be used.

### DIFF
--- a/src/observable/ForkJoinObservable.ts
+++ b/src/observable/ForkJoinObservable.ts
@@ -40,8 +40,98 @@ export class ForkJoinObservable<T> extends Observable<T> {
   /* tslint:enable:max-line-length */
 
   /**
-   * @param sources
-   * @return {any}
+   * Joins last values emitted by passed Observables.
+   *
+   * <span class="informal">Wait for Observables to complete and then combine last values they emitted.</span>
+   *
+   * <img src="./img/forkJoin.png" width="100%">
+   *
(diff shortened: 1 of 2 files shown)


src/observable/ForkJoinObservable.ts
```diff
@@ -40,8 +40,98 @@ export class ForkJoinObservable<T> extends Observable<T> {
   /* tslint:enable:max-line-length */
 
   /**
-   * @param sources
-   * @return {any}
+   * Joins last values emitted by passed Observables.
+   *
+   * <span class="informal">Wait for Observables to complete and then combine last values they emitted.</span>
+   *
+   * <img src="./img/forkJoin.png" width="100%">
```
staltz: forkJoin doesn't have a `asDiagram` test case in its spec file, so I recommend either making that test case or removing this image. Otherwise it would be shown as a broken image.

staltz: forkJoin needs an `asDiagram` test case in its spec file, or you should remove this line. Otherwise, it would render as a broken image element on the page.

Podlas29: @staltz thanks! I added `asDiagram` test case.

staltz:
Thanks @mpodlasin for the great PR, sorry for the delay in merging it.

Issue #2391, opened 2017-02-19
All window* operators emit Subjects instead of Observables


martinsik:
All `window*` operators emit Subjects instead of Observables.

**RxJS version:** 5.1.1

**Code to reproduce:**
```
const Observable = Rx.Observable;

Observable.timer(0, 200)
  .window(Observable.timer(0, 1000))
  .take(1)
  .subscribe(observable => console.log('window', observable.constructor.name));

Observable.timer(0, 200)
  .windowCount(5)
  .take(1)
  .subscribe(observable => console.log('windowCount', observable.constructor.name));

Observable.timer(0, 200)
  .windowTime(1000)
  .take(1)
  .subscribe(observable => console.log('windowTime', observable.constructor.name));

Observable.timer(0, 200)
  .windowToggle(Observable.timer(0, 1000), () => Observable.timer(1000, 1000))
  .take(1)
  .subscribe(observable => console.log('windowToggle', observable.constructor.name));

Observable.timer(0, 200)
  .windowWhen(() => Observable.of(1).delay(1000))
  .take(1)
  .subscribe(observable => console.log('windowWhen', observable.constructor.name));
```

Demo: https://jsbin.com/qiseles/3/edit?js,console

This prints to console:
```
window Subject
windowCount Subject
windowTime Subject
windowWhen Subject
windowToggle Subject
```

**Expected behavior:**
All `window*` operators should emit only Observables.
I checked RxJS 4 and all `window*`s emit `AnnonymousObservable`'s.

RxJS 4 Demo: https://jsbin.com/hehozin/2/edit?js,console

This prints to console:
```
window AnonymousObservable
windowCount AnonymousObservable
windowTime AnonymousObservable
windowWithTimeOrCount AnonymousObservable
```

**Actual behavior:**
All `window*` operators emit Subjects.

**Additional information:**

This would be obviously easy to fix with `asObservable()` but I'm wondering if this is intentional behavior. Since `window*` operators are going to be used probably with `switch()` then you don't have many ways to unintentionally misuse the Subjects. Even if I try to emit manually values to a Subject from `window*` it won't be reemitted because it's before the `switch()` subscribes to it. So this example won't produce unexpected results:

```
Observable.timer(0, 200)
  .window(Observable.timer(0, 1000))
  .do(subject => subject.next('ignore me'))
  .switch()
  .take(5)
  .subscribe(val => console.log('next', val));
```

[...]

PR #2408, opened 2017-02-23
fix(window*): always emit only Observables

martinsik:
**Description:**
This PR makes `window*` operators emit only Observables instead of Subjects.

#2391

### DIFF
--- a/spec/operators/window-spec.ts
+++ b/spec/operators/window-spec.ts
@@ -1,3 +1,4 @@
+import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
 import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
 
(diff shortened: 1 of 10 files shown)


blesh:
This is debatable. By wrapping the Subject in an Observable, it's going to incur a performance penalty for each nexted value through the windowed observables. On the other hand, giving a subject here does leak an implementation detail.

FWIW, the decision to send in a Subject here was intentional from the beginning for performance reasons. 

Given that:

1. It's fairly harmless in this instance to give people (who know) access to the full Subject API
2. We're doing what we can in most places (docs, typings, etc) to guide people to not use them as Subject.
3. It's a performance penalty and a few additional items in a call stack to add this.

... I'm afraid this is a :-1: from me.

I do understand where your coming from, and I suspect there will be similar PRs from others in the future, but I don't think this is something we want to do at the moment.

blesh:
I'm going to close this for now, but I'll request that @trxcllnt review it as well, and reopen if he disagrees.

mattpodwysocki:
@blesh please reconsider this one.  You closed without objectively measuring the impact on this, as we do not want to leak through abstractions like this, therefore leaving us open to breaking changes if someone relied upon it being a subject

martinsik:
I just noticed it's the same with `retryWhen()`: https://github.com/ReactiveX/rxjs/blob/master/src/operator/retryWhen.ts#L66-L67

and `repeatWhen()` as well: https://github.com/ReactiveX/rxjs/blob/master/src/operator/repeatWhen.ts#L112-L113

benlesh:
> @blesh please reconsider this one. You closed without objectively measuring the impact on this, as we do not want to leak through abstractions like this, therefore leaving us open to breaking changes if someone relied upon it being a subject

I'm always open to reconsidering PRs! :) I just closed for cleaner book keeping.

We've made a lot of trade-offs for performance reasons, trade-offs that sometimes come with abstraction leaks. In JavaScript this is not at all uncommon to do. I'll reopen this PR for now, and we can discuss at the next meeting.

staltz:
I would argue to return an Observable from **TypeScript**, not necessarily from JavaScript, and otherwise keep the implementation as it was, like Ben argued.

Issue #2388, opened 2017-02-18
add index to delayWhen selector


Dorus:
Most operators include the index of the current element in the selector function For example `lter(predicate: function(value: T, **index**: number)` and ` mergeMap(project: function(value: T, **?index**: number)`.

`delayWhen` however is missing the index:
```
public delayWhen(delayDurationSelector: function(value: T): Observable, subscriptionDelay: Observable): Observable
```
However, it would be especially useful if `delayDurationSelector` included the index. That would make it easy to create an incremental delay, for example when `delayWhen` is combined with `retryWhen`.

macklinu:
Was this resolved by #2473? Anything else left to do here?

cartant:
> Was this resolved by #2473?

That seems to be the case, yes.

PR #2409, opened 2017-02-23
feat(delayWhen): add index to the selector function

martinsik:
This PR adds current index to the selector function in `delayWhen`operator.

```
source.delayWhen((value, index) => {
  ...
});
```

#2388

### DIFF
--- a/spec/operators/delayWhen-spec.ts
+++ b/spec/operators/delayWhen-spec.ts
@@ -1,3 +1,4 @@
+import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
 import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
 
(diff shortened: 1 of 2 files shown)


kwonoj:
Change looks fine, but I assume TypeScript users who used selector before encounter compilation error by selector signature changes?

benlesh:
@kwonoj I don't think so... since arguments are optional in JavaScript, TypeScript *shouldn't* complain I don't think.

benlesh:
@kwonoj confirmed... `(a: A, b: B) => void` matches `(a: A) => void` just fine in TS.

kwonoj:
Thanks for confirmation. I'll check this in around today to tomorrow.

benlesh:
Oops.... @kwonoj, we didn't want to merge this new feature until after we patched the existing minor.

kwonoj:
@benlesh 😱  should this be reverted?

benlesh:
@kwonoj yes, please. What I'd like to do is merge all of the `minor` changes *after* I merge all of the patches. So we can publish 5.2.1 and 5.3.0 at the same time.

benlesh:
ugh.. so we're going to have to re-merge this later. :) @kwonoj can you open another PR with these changes? (no worries @martinsik, your github credit is already in the history :) )

kwonoj:
Yup, will do.

PR #2411, opened 2017-02-23
docs(fromEvent): add more details about the operator

mpodlasin:
* Add list of types of event targets fromEvent accepts.
* Add details and and example about optional parameters for addEventListener. 
* Add note about callback-oriented APIs and operators that might be more useful in that case.
* Describe when fromEventPattern would be more useful.
* Add note that only first argument from event handler will appear in Observable and describe how optional project function mitigates that problem.


staltz:
Great job @mpodlasin, sorry for the long delay in merging

PR #2413, opened 2017-02-24
feat(fromEventPattern): improve typings

mpodlasin:
**Description:**

Currently `handler` injected into users function has general type signature `Function`. Because of that, even when used with properly typed APIs, TypeScript will not derive the type of resulting Observable.

Thanks to this change, output Observable is derived properly in MOST (see comment below) cases.

**Caveats:**

When `handler` expected by API has type `() => void`, I wanted to return `Observable<void>` type, but was unable to (no order of overloaded operator signatures worked well). By default TS sets returned value to `Observable<{}>`. Should not be huge issue since user does not use that `{}` value anyways. That being said type is wrong. If that is a real issue for a user, he can enforce `void` manually by calling `fromEventPattern<void>(...)`.

### DIFF
--- a/spec/observables/fromEventPattern-spec.ts
+++ b/spec/observables/fromEventPattern-spec.ts
@@ -4,7 +4,7 @@ import * as Rx from '../../dist/cjs/Rx';
 import {noop} from '../../dist/cjs/util/noop';
 import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
 
-declare const { asDiagram };
+declare const { asDiagram, type };
 declare const expectObservable: typeof marbleTestingSignature.expectObservable;
 
 declare const rxTestScheduler: Rx.TestScheduler;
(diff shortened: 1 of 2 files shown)


kwonoj:
While I was suffering this cases as well, I'm bit concerned usecases for non-most supported usecase, some framework like `Electron` does have own signature for lot of custom eventemitter. Shouldn't there be at least worst-case fallback to `Function` as previous if there's none of matching cases? While I have gut feeling it won't play well with overloaded signatures, probably reason why we've sticked into `Function` originally as same as static creation method `create<T>`.

Podlas29:
I see @blesh implemented that one, so perhaps he might have some insight as well.

 However last message in his pr says "I have to move on", so my bet is it just slipped under the radar. ;)

Podlas29:
@kwonoj have in mind that API itself here is irrelevant. If in tests I switch 'addHandler` method name to `fffffff`and have the `handler` be third parameter instead of first, it will work just fine.

I'll play around a little bit more and try to test it against some bizarre API signatures to see how it handles them.

kwonoj:
overall I'm always in favor of having more declarative types. Just consider myself obsessively try to avoid possible regressions. :)

Podlas29:
Well... I actually found failing case (API that expects `handler` to return something else than `void`).

I added it to tests and added overload that falls back to `Function`, but at this point I am not sure about backwards compatibility myself. Perhaps major release?

david-driscoll:
@mpodlasin if you can resolve the conflicts we can merge this.

benlesh:
Given that these files have drastically changed, and we're deprecating the resultSelector, I think we can close this one.

Thank you as always, @mpodlasin!

PR #2416, opened 2017-02-25
docs(fromEventPattern): add more details about the operator

mpodlasin:
* Compare to fromEvent. 
* Describe what addHandler and removeHandler functions are called with and when they are called. 
* Describe how APIs that return cancellation token on handler registration can be handled and show example of that.
* Add note that only first parameter from event handler will appear in the output Observable. 
* Describe how project function can be used to get access to other parameters and show an example of that.


benlesh:
@mpodlasin can you resolve conflicts? Sorry this went so long.

benlesh:
Belated thanks, @mpodlasin

PR #2419, opened 2017-02-26
docs(if): create documentation for the operator

mpodlasin:
* Describe what if operator accepts and returns.
* Emphasize when condition function will be called. 
* Describe two basic use cases for that operator and show examples with that use cases.
* Compare to defer operator.


ladyleet:
thx @mpodlasin and apologies for the delay - we have had other pressing issues we have been working on. @benlesh or @kwonoj can we go ahead and merge this in?

ladyleet:
lol @mpodlasin bet you thought this would *never* get merged! thanks for your contribs and patience! :) <3

PR #2432, opened 2017-03-01
docs(merge): add more details about the operator

mpodlasin:
* Describe more when merge will complete and error with different input Observables. 
* Add paragraph about concurrent parameter. 
* Add note about scheduler and what it may be used for in the operator.


benlesh:
SInce we're deprecating the merge operator I think we can close this one.

PR #2435, opened 2017-03-03
Fix symbols to work with Babel UMD exporting

### LINKED ISSUE #2415: Consider rename observable symbol $$observable => observable
**Actual behavior:**
I use [transform-es2015-modules-umd](https://babeljs.io/docs/plugins/transform-es2015-modules-umd/)  Babel plugin. And [I need](https://github.com/angular/angular/pull/14722/files) to map `import {$$observable} from 'rxjs/symbol/observable';` 

Observable symbol is exported as `Rx.Symbol.observable` but exported symbol from `rxjs/symbol/observable` is called `$$observable`. Because symbols have different names I cannot just map it like `"rxjs/symbol/observable": "Rx.Symbol"`

**Expected behavior:**
I see 3 possible solutions:
1. Rename `$$observable` => `observable` so it will be possible to do `"rxjs/symbol/observable": "Rx.Symbol"`
2. export `$$observable` from `Rx` as is => `Rx.$$observable`
3. export `$$observable` from `Rx.Symbol` => `Rx.Symbol.$$observable`

I can also accomplish it by changing import `import {$$observable} from 'rxjs/symbol/observable';`  => `import {Symbol} from 'rxjs/Rx';`  but I don't want to import the whole lib :)

DzmitryShylovich:
Closes #2415

### DIFF
--- a/src/Rx.ts
+++ b/src/Rx.ts
@@ -171,7 +171,7 @@ import { QueueScheduler } from './scheduler/QueueScheduler';
 import { AnimationFrameScheduler } from './scheduler/AnimationFrameScheduler';
 import { $$rxSubscriber as rxSubscriber } from './symbol/rxSubscriber';
 import { $$iterator as iterator } from './symbol/iterator';
-import { $$observable as observable } from './symbol/observable';
+import { observable } from './symbol/observable';
 
 /* tslint:enable:no-unused-variable */
 
--- a/src/Rx.ts
+++ b/src/Rx.ts
@@ -169,7 +169,7 @@ import { AsapScheduler } from './scheduler/AsapScheduler';
 import { AsyncScheduler } from './scheduler/AsyncScheduler';
 import { QueueScheduler } from './scheduler/QueueScheduler';
 import { AnimationFrameScheduler } from './scheduler/AnimationFrameScheduler';
-import { $$rxSubscriber as rxSubscriber } from './symbol/rxSubscriber';
+import { rxSubscriber } from './symbol/rxSubscriber';
 import { iterator } from './symbol/iterator';
 import { observable } from './symbol/observable';
 
--- a/src/Rx.ts
+++ b/src/Rx.ts
@@ -170,7 +170,7 @@ import { AsyncScheduler } from './scheduler/AsyncScheduler';
 import { QueueScheduler } from './scheduler/QueueScheduler';
 import { AnimationFrameScheduler } from './scheduler/AnimationFrameScheduler';
 import { $$rxSubscriber as rxSubscriber } from './symbol/rxSubscriber';
-import { $$iterator as iterator } from './symbol/iterator';
+import { iterator } from './symbol/iterator';
 import { observable } from './symbol/observable';
 
 /* tslint:enable:no-unused-variable */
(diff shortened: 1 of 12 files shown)


benlesh:
@DzmitryShylovich apologies for taking my time to review this. I've had a busy couple of weeks

DzmitryShylovich:
@benlesh ok, np, I understand. new workplace etc :)

should I change iterator and scheduler symbols in the similar fashion?

src/Rx.ts
```diff
@@ -171,7 +171,7 @@ import { QueueScheduler } from './scheduler/QueueScheduler';
 import { AnimationFrameScheduler } from './scheduler/AnimationFrameScheduler';
 import { $$rxSubscriber as rxSubscriber } from './symbol/rxSubscriber';
 import { $$iterator as iterator } from './symbol/iterator';
-import { $$observable as observable } from './symbol/observable';
+import { observable } from './symbol/observable';
```
DzmitryShylovich: I can use `$$observable` as alias so only import will change `observable as $$observable`

benlesh:
> should I change iterator and scheduler symbols in the similar fashion?

Sure.. We'll just need to rename this PR to something like "Fix symbols to work with Babel UMD exporting".

Also: If you can please change the commit messages to be like `fix(symbol/observable): will be exported as \`observable\` to better support Babel UMD and others`. That way this will show up in our automated change logs.

I'm still not certain how we should announce this change, and (ideally) deprecate the `$$observable`-type exports by v7.  Maybe @IgorMinar has a recommendation?

benlesh:
Thank you for your help on this @DzmitryShylovich

DzmitryShylovich:
@benlesh done. PTAL

src/Subject.ts
```diff
@@ -5,7 +5,7 @@ import { Subscriber } from './Subscriber';
 import { ISubscription, Subscription, TeardownLogic } from './Subscription';
 import { ObjectUnsubscribedError } from './util/ObjectUnsubscribedError';
 import { SubjectSubscription } from './SubjectSubscription';
-import { $$rxSubscriber } from './symbol/rxSubscriber';
+import { rxSubscriber as Symbol_rxSubscriber } from './symbol/rxSubscriber';
```
benlesh: haha... GOTCHA! `rxSubscriber` doesn't exist on `Symbol`, really... (i.e. `Symbol.rxSubscriber` is not at thing)... so you can probably just leave this as `rxSubscriber` or `as rxSubscriberSymbol` or something)

benlesh: So close

DzmitryShylovich: > Symbol.rxSubscriber is not at thing

https://github.com/ReactiveX/rxjs/blob/master/src/Rx.ts#L212 ?

DzmitryShylovich: http://plnkr.co/edit/SS0t7wCUh2ANZiVGeTNi?p=preview I can import it. or what do you mean by `not a thing`? :)

benlesh: @DzmitryShylovich oh I meant that `Symbol.observable` -> `Symbol_observable`, `Symbol.iterator` -> `Symbol_iterator` ... those are polyfills.  `rxSubscriber` is just an RxJS library-specific symbol. It wouldn't exist on `window.Symbol` or the like.

DzmitryShylovich: hmm ok I'll just leave it as rxSubscriber for simplicity

DzmitryShylovich: done

Issue #2437, opened 2017-03-03
[PROPOSAL][AWAITING TS 2.1] Retry to move to @types/ type definitions


saneyuki:
1. Before to release v5, we tried to move to `@types/***` type definitions from `typings/`.
   - Worked in https://github.com/ReactiveX/rxjs/pull/2120
2. But we had faced to https://github.com/Microsoft/TypeScript/issues/11849, so we backout-ed the above change. 
3. For the future if we move to use TypeScript 2.1 or later, we can retry to do it.

kwonoj:
I've migrated into @types packagage in other codebases and generally in favor of this, but not sure how to verify regression.

alexeagle:
To verify regression, you can use a typings test like Angular:
https://github.com/angular/angular/tree/master/integration/typings_test_ts21

The formula:
- for each version of TypeScript you support users having
  - install that version
  - install the distribution you publish to npm
  - compile a file that exhaustively imports all the public APIs

if there's anything in your distro that would break users type-checking your lib, this should detect it.

kwonoj:
This is completed in master.

Issue #2436, opened 2017-03-03
[PROPOSAL][AWAITING TS 2.1] Use tsc's --importHelpers option to reduce the code size.


saneyuki:
- _This is a proposal for the future semver major._
- After TypeScript 2.1, we can use [external helper module](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#support-for-external-helpers-library-tslib) to emit a down-level transforming. By this option, we may have a chance to reduce our code size.

kwonoj:
As far as I know this requires consumer to import tslib for their own, isn't it?

saneyuki:
@kwonoj 

I seem we need to add tslib into our _dependencies_, but we don't have to require installing tslib for our user.

csvn:
In my project when using rollup, I end up with 103 `__extends` from `rxjs` alone (`__extends$1` => `__extends$103`). It would be great if this could be considered again. as @saneyuki says, it could simply be added as a dependency for `rxjs`. Externally it should not be noticed by users.

csvn:
@benlesh Is this something that can be considered for v6 with the other large restructuring?

With all imports used from `rxjs@alpha`, the `esm5` version is ~32KB larger (~55KB non-minified) than if `importHelpers` would be used. I made a small example repo here: https://github.com/csvn/rxjs-tslib-example.

As I see it, there are four possible options:

- Use `importHelpers` with `tslib` as peerDependency
- Use `importHelpers` with `tslib` as dependency
- Use rollup/webpack to combine all files into one file for each entry point with `es2015` module exports (e.g. `index.esm.js` + `operators.esm.js`)
- Do nothing, and keep repeated helpers

Using `tslib` as peerDependency would need to be done in i a major version.

benlesh:
@csvn, I think most people are using rollup/webpack to combine the files. This had fallen off my radar, and I'm not exactly sure what's best practice here now. Let me confer with a few people.

kwonoj:
I am concerned about making this as dep, as rx uses pinned version of dep it could cause 2 different tslib set (user code have latest tslib version a, while rx still uses old version b). Reason I haven't push forward this in v5. We could make as peerdep and open version instead of pinning, but that opens possibility breaking change in tslib (or incorrect combination between compiler) causes some unexpected behavior.

Also curious what's best practices to resolve those cases. If there is, I'd definitely push it for v6.

benlesh:
cc @jasonaden @IgorMinar

[...]

PR #2438, opened 2017-03-03
[DO NOT MERGE][MERGE INTO NEXT BRANCH] chore(TypeScript): Upgrade TypeScript to 2.2

### LINKED ISSUE #2178: Custom error object is no longer typeof Error from TypeScript 2.1

saneyuki:
- BREAKING CHANGE: This requires to use TypeScript 2.2 or later for
  users.
   - This will be effect to the next semver major release.
   - Not related for the next v5 release.
- This is the proposal for the next semver major release.
- After TypeScript 2.1, there are some breaking change which cannot
  support ~IE10. It is handled by
  https://github.com/ReactiveX/rxjs/issues/2178.
- This also sort tslib's version with tsc's version.
- By this change, we can retry to install type definitions from npm.
  - https://github.com/ReactiveX/rxjs/pull/2120

## Related Issues

- This requires to fix #2178 before merging this.
- https://github.com/ReactiveX/rxjs/issues/2437
- https://github.com/ReactiveX/rxjs/issues/2436


kwonoj:
Please refer https://github.com/ReactiveX/rxjs-core-notes/blob/master/2017-01/january-30.md, we decided to align with ng2's compiler upgrade plan and until then we probably won't visit compiler bump up.

saneyuki:
> we decided to align with ng2's compiler upgrade plan and until then we probably won't visit compiler 

I know it. This is still a proposal and I don't claim to merge this immediately :)

kwonoj:
Do we need pr opened for this? There is no specific eta for this and probably this PR will stale for quite long time, will make you wait for those time period.

saneyuki:
Okay. I'll close this.

kwonoj:
I'm also quite in favor this and would like to have it soon, thanks for bringing up. Let's take a revisit once we can introduce this.

PR #2439, opened 2017-03-03
typing(delay): clarify the typing of DelaySubscriber

saneyuki:
**Description:**

This just clean up an internal typing.

- This retry the part of https://github.com/ReactiveX/rxjs/pull/2125.


src/operator/delay.ts
```diff
@@ -134,7 +142,7 @@ class DelaySubscriber<T> extends Subscriber<T> {
 }
 
 class DelayMessage<T> {
-  constructor(private time: number,
-              private notification: any) {
+  constructor(public time: number,
```
kwonoj: `public readonly` maybe?

saneyuki: Okay. I fixed it.

PR #2441, opened 2017-03-04
docs(webSocket): fix node.js example

jsayol:
**Description:** Fixes the example showing the usage of `WebSocketSubject` on Node.js.

**Related issue (if exists):** none

### DIFF
--- a/src/observable/dom/WebSocketSubject.ts
+++ b/src/observable/dom/WebSocketSubject.ts
@@ -48,30 +48,32 @@ export class WebSocketSubject<T> extends AnonymousSubject<T> {
    *
    * @example <caption>Wraps browser WebSocket</caption>
    *
-   * let subject = Observable.webSocket('ws://localhost:8081');
-   * subject.subscribe(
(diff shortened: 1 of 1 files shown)


src/observable/dom/WebSocketSubject.ts
```diff
@@ -60,12 +60,11 @@ export class WebSocketSubject<T> extends AnonymousSubject<T> {
    *
    * import { w3cwebsocket } from 'websocket';
    *
-   * let socket = new WebSocketSubject({
+   * let subject = new WebSocketSubject({
```
benlesh: Actually, I think we should go with `socket$` everywhere, and use `Observable.webSocket` rather than `new WebSocketSubject`.

I'm glad you caught this, though. Would you mind making this change?

jsayol: Done.

jsayol:
Let me know if I should squash the commits, I couldn't find anything about it in the contributing guidelines :)

PR #2443, opened 2017-03-04
feat(shareReplay): adds `shareReplay` variant of `publishReplay`

benlesh:
`shareReplay` returns an observable that is the source multicasted over a `ReplaySubject`. That replay subject is recycled on error from the `source`, but not on completion of the source. This makes `shareReplay` ideal for handling things like caching AJAX results, as it's retryable. It's repeat behavior, however, differs from `share` in that it will not repeat the `source` observable, rather it will repeat the `source` observable's values.

related #2013, #453, #2043

### DIFF
--- a/spec/operators/shareReplay-spec.ts
+++ b/spec/operators/shareReplay-spec.ts
@@ -1,0 +1,167 @@
+import {expect} from 'chai';
+import * as Rx from '../../dist/cjs/Rx';
+import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
+
+declare const { asDiagram };
+declare const hot: typeof marbleTestingSignature.hot;
+declare const cold: typeof marbleTestingSignature.cold;
+declare const expectObservable: typeof marbleTestingSignature.expectObservable;
+declare const expectSubscriptions: typeof marbleTestingSignature.expectSubscriptions;
+
+const Observable = Rx.Observable;
+
+/** @test {shareReplay} */
+describe('Observable.prototype.shareReplay', () => {
+  it('should mirror a simple source Observable', () => {
+    const source = cold('--1-2---3-4--5-|');
+    const sourceSubs =  '^              !';
+    const published = source.shareReplay();
+    const expected =    '--1-2---3-4--5-|';
+
+    expectObservable(published).toBe(expected);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+  });
+
+  it('should do nothing if result is not subscribed', () => {
+    let subscribed = false;
+    const source = new Observable(() => {
+      subscribed = true;
+    });
+    source.shareReplay();
+    expect(subscribed).to.be.false;
+  });
+
+  it('should multicast the same values to multiple observers, bufferSize=1', () => {
+    const source =     cold('-1-2-3----4-|'); const shared = source.shareReplay(1);
+    const sourceSubs =      '^           !';
+    const subscriber1 = hot('a|           ').mergeMapTo(shared);
+    const expected1   =     '-1-2-3----4-|';
+    const subscriber2 = hot('    b|       ').mergeMapTo(shared);
+    const expected2   =     '    23----4-|';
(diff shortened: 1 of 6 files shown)


benlesh:
I suspect the 0.009% change in coverage isn't worth blocking this over, as it's pretty much a rounding error.

robwormald:
> It's repeat behavior, however, differs from share in that it will not repeat the source observable, rather it will repeat the source observable's values.

can you expand on the behavior differences? Perhaps with an example?

benlesh:
> can you expand on the behavior differences? Perhaps with an example?

The [tests show the behavior](https://github.com/ReactiveX/rxjs/pull/2443/files#diff-52f62da75a279650a9e6557aa988f9ceR130), but in case they aren't clear enough...

`repeat` will work _exactly_ the same way repeat works with `publishReplay().refCount()`.

assume that `source$` below is using `shareReplay()`...

```
--a--b--c--|             // source$
--a--b--c--(abcabc|)     // source$.repeat(3);
```

`retry` on the otherhand, will work exactly the same way `.multicast(() => new ReplaySubject()).refCount()` does.

```
--a--b--c--#                                      // source$
--a--b--c----a--b--c----a--b--c----a--b--c--#     // source$.retry(3)
```

... hopefully, that illustrates the behavior?

trxcllnt:
The only reservation I have is that the ReplaySubject isn't deallocated when the refCount Observable is disposed. By referencing the RS in the Observable definition, it'll only be GC'd when the `shareReplay` _Observable_ is dereferenced. None of the other operators work this way, and it seems this could introduce leaks when people unfamiliar with the implementation use Observables like normal, and expect they teardown when all the Subscriptions are disposed.

wardbell:
This seems promising. How do I use if for the following scenario.

I want to cache a list of customers from the server but periodically refresh that list, say, if it is more than an hour old. 
1. First subscriber triggers the first fetch (in the `source$`, yes?).
1. If a second subscriber asks for customers _before the first fetch succeeds_, it waits until that first fetch succeeds or errors. 
1. What does the second subscriber see if it errors and we retry? Should it wait? Is there a race?
1. How do I trigger the re-fetch after the cached value expires? That only works if the fetching source (unlike Angular `http`) _does not complete_, right? 

 In the absence of the `shareReplay` operator, I handle this scenario by managing both the fetching `Observable` and the `ReplaySubject` on my own in an imperative manner. Should I continue to do so?

Be gentle ... I'm a comparative noob.

spec/operators/shareReplay-spec.ts
```diff
@@ -0,0 +1,167 @@
...
+  it('should completely restart for subsequent subscriptions if source errors, bufferSize=2', () => {
+    const source =     cold('-1-2-----3-#               ');
+    const shared = source.shareReplay(2);
+    const sourceSubs1 =     '^          !               ';
+    const subscriber1 = hot('a|                         ').mergeMapTo(shared);
+    const expected1   =     '-1-2-----3-#               ';
+    const subscriber2 = hot('    b|                     ').mergeMapTo(shared);
+    const expected2   =     '    (12)-3-#               ';
+    const subscriber3 = hot('               (c|)        ').mergeMapTo(shared);
+    const expected3   =     '               -1-2-----3-#';
```
staltz: Shouldn't `expected3` be `(23)#`?

staltz: Ignore me. It's correct. (because the source errored, it didn't *complete*, so we create a new ReplaySubject for the next multicast)

staltz:
Apart from what has already been said (e.g. by Paul), this looks good to merge IMO.

benlesh:
@staltz, in discussing this with @trxcllnt on Slack, I pointed out that the memory behavior her is no different than how any of the publish operators, or `multicast(subject)` have worked since RxJS 2.

Thought I'd add it to the thread though, for visibility.

QuentinFchx:
Shall the [MIGRATION.md](https://github.com/ReactiveX/rxjs/blob/master/MIGRATION.md) be updated aswell? It might be confusing otherwise

Issue #2455, opened 2017-03-11
Unexpected behaviour of first operator together with switchMap


artall64:
**RxJS version:** 5.2.0

**Code to reproduce:** 
var x = Rx.Observable.interval(1000)
  .do( x=> console.log("One"))
  .first()
  .switchMap(x => Rx.Observable.interval(1000))
  .do( x=> console.log("Two"))
  .subscribe((x) => {})

**Expected behaviour:**
// In console you will see:
// One
// Two
// Two
// Two
// Two
// etc...

**Actual behaviour:**
// In console you will see:
// One
// One
// Two
// One
// Two
// One
// etc... 

**Additional information:**
I've expected that first() in this case should behave exactly the same as a take(1) but behaviour is quite different

trxcllnt:
This sounds like a bug in `first`

Podlas29:
@trxcllnt In a way it is `first`s fault and in a way it is not. I implemented forcing unsubscription logic when first value arrives to `first` - this is what `take(1)` does and it works, but for me feels extremaly hacky.

The problem is since every operator is implemented via `Subscriber` inheritance, it can override default unsubscription logic that happens when source completes or errors.

This is exactly what `switchMap` does - if inner Observable (the one that is returned by function passed to `switchMap`) is still emitting values (as it is in the case of `interval` which will never stop), `switchMap` holds off with unsubscribing from its source (result of `first` in that case).

My point is - should an operator have power to intercept unsubscription logic of source observable? When observable sends complete or error notification, shouldn't one central mechanism make sure that it is always unsubscribed, so that operators are not prone to such mistakes?

Podlas29:
As far as I understand, in tc39 spec, if I call `complete` on observer in my custom Observable, I do not have to trigger unsubscription logic by hand. I can trust that there exists a mechanism that will call it for me. Right now implementation of `take` (https://github.com/ReactiveX/rxjs/blob/master/src/operator/take.ts#L80) and my fix for `first` go against it, by not trusting that underlying mechanism will work properly and calling unsubscription by hand.

Podlas29:
In other words: shouldn't this: https://github.com/ReactiveX/rxjs/blob/master/spec/operators/switchMap-spec.ts#L57-L58 be absolutely illegal? We have source that completed, but we are still subscribing to it.

trxcllnt:
@Podlas29 ah, yes there is a deeper issue here. I'll comment on #2457 with more details.

[...]

PR #2457, opened 2017-03-12
[DO NOT MERGE][NEXT?] Unsubscribe source observable immediately after it completes or errors

mpodlasin:
**Disclaimer:**
For discussion. Either I am doing something very smart or something very dumb right now. ;)

**Problem:**
Currently many operators still subscribe for some time to source observables after they completed, and thus do not call teardown logic immediately after completion. These kind of marbles are in many tests:
```
---a-----b-----c-----|               <- source observable
^------------------------------!    <- subscription to that observable
```

It is a source of very subtle bugs, like: #2455 

It also seems that this goes against tc39 proposal, which guarantees that source will be unsubscribed whenever it sends `complete` or `error` notification to an observer: https://tc39.github.io/proposal-observable/#subscription-observer-objects

There are operators that have to manually trigger their unsubscription to normalilze their behaviour: https://github.com/ReactiveX/rxjs/blob/master/src/operator/take.ts#L80

### DIFF
--- a/spec/observables/onErrorResumeNext-spec.ts
+++ b/spec/observables/onErrorResumeNext-spec.ts
@@ -14,7 +14,7 @@ describe('Observable.onErrorResumeNext', () => {
     const next1  =  cold(        '--c--d--#');
     const next2  =  cold(                '--e--#');
     const next3  =  cold(                     '--f--g--|');
-    const subs =         '^                            !';
+    const subs =         '^       !';
     const expected =     '--a--b----c--d----e----f--g--|';
 
     expectObservable(Observable.onErrorResumeNext(source, next1, next2, next3)).toBe(expected);
@@ -22,12 +22,12 @@ describe('Observable.onErrorResumeNext', () => {
   });
 
   it('should continue array of observables', () => {
-    const source = hot('--a--b--#');
+    const source =  hot('--a--b--#');
     const next  = [ source,
                    cold(        '--c--d--#'),
                    cold(                '--e--#'),
                    cold(                     '--f--g--|')];
-    const subs =        '^                            !';
+    const subs =        '^       !';
     const expected =    '--a--b----c--d----e----f--g--|';
 
     expectObservable(Observable.onErrorResumeNext(next)).toBe(expected);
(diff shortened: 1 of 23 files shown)


kwonoj:
My personaly suggestion is starting for issue to discuss details first for these kind of cases, as it is not ceratin what desired behavior we want to have, this PR probably stall very long time and either be closed or required lot of updates in further.

kwonoj:
(But not saying this should be closed now since it's already prepped)

Podlas29:
@kwonoj Will have that in mind, but I wanted to see if my planned fix will work and won't break stuff it shouldn't.

kwonoj:
@Podlas29 if you don't mind, would you create issue? It's better to make discussion continues on issue, not on PR.

trxcllnt:
@Podlas29 the spirit of the PR seems right, but allocating an extra Subscription per Subscriber shouldn't be necessary. We essentially hook up the "parent" subscription [here](https://github.com/ReactiveX/rxjs/blob/master/src/Subscriber.ts#L72), so I have a hunch we can update this PR to use that directly.

Podlas29:
@kwonoj https://github.com/ReactiveX/rxjs/issues/2459

benlesh:
This PR will require a LOT of review. The behavior seems right-ish, on the surface, and the spirit of the PR seems solid. But there are a lot of questions and even more changes to review.

Most of our tests were modeled after the tests that existed in Rx4 and under. So many of these changes may incur breaking changes for the historical use of the library.

It also looks like (at a quick glance) that we're adding another Subscription reference per subscription, as well as a few new publicly available properties and methods.

Questions:

How does this affect performance?
How about memory pressure?
Who will this break?
Is this fixing a problem that many people are affected by? Is it really a problem?

There are just so many behavioral changes here, it's really going to be hard to tackle this and give it the proper attention it deserves in any reasonable amount of time.

benlesh:
As an aside: the format of your commit messages is incorrect.

Perhaps it would be better to file each bug you think is a bug separately with it's accompanying test from this PR, so each one could be discussed individually?

Podlas29:
@benlesh - please take a look at #2459 In following days I will try to flesh out the problem more, provide more examples as well as alternative plans for fixing it in a more controlled/gradual way.

benlesh:
Okay, I think this PR should be redone and added as a bug fix during 6.0. It's not really a "breaking" change to anything but a broken behavior, IMO.

benlesh:
Closing in favor of #3963, which is an attempt to pull this in.

Issue #2459, opened 2017-03-13
Many operators do not call teardown logic of source Observable immediately after it completes or errors


mpodlasin:
#2457 - general problem description and proposed solution.
#2455 - bug that made me discover an issue, directly caused by it. Contains much more in-depth explanation of what is happening in that concrete case.

**Consequences:**

1. every take-ish operator (one that completes earlier than its source Observable), will keep subscribing to source Observable when combined with operator that prolongs subscription (here `switchMap`).
```js
Rx.Observable.interval(1000)
.do(() => console.log('First just wanted one value, but I\'ll keep creating them forever'))
.first()
.switchMap(x => Rx.Observable.interval(1000))
.subscribe();
```

2. custom made Observable (third party & via `create`) will not have their teardown logic called when they complete, when combined with operator that prolongs subscription (here `switchMap`):
```js
Rx.Observable.create(obs => {
  obs.next(5);
  obs.complete();
  
  return () => console.log('I will never be called');
})
.switchMap(x => Rx.Observable.interval(1000))
.subscribe();
```

3. Some error handling operators (currently only `onErrorResumeNext`) might still hold subscription to an Observable that had an error in its operator chain:
```js
Rx.Observable.onErrorResumeNext(

  Rx.Observable.interval(1000)
  .do(x => console.log('a', x)) // this will be logging forever, even when map beneath errors
  .map(i => {
    if (i > 10) {
      throw new Error('oops!');
    }
  
    return i;
  }),

  Rx.Observable.interval(1000)

)
.do(x => console.log('b', x))
.subscribe();
```

**What operators it affects?**

Points 1. and 2. affect all operators that have following pattern in their unit tests:
```
----a----b----c----|                  <- Observable on which operator is used
^--------------------------!          <- subscriptions diagram for that Observable
```
```
---a-----b----c---|
^---------------------------
```
since `|` is the moment when Observable completes and `!` is the moment when unsubscription function is called (which happens too late or sometimes does not happen at all).

Point 3. from what I have seen affects only `onErrorResumeNext` but in general it might affect all operators with following pattern in their unit tests:
```
----a----b----c----#
^--------------------------!
```
```
---a-----b----c---#
^---------------------------
```
since `#` marks the moment when Observable errored and `!` marks the moment when unsubscription function is called.

**Alternative fix:**
[...]

PR #2463, opened 2017-03-15
fix(first): force unsubscription logic on complete and error

### LINKED ISSUE #2455: Unexpected behaviour of first operator together with switchMap

mpodlasin:
**Description**

Force unsubscription logic when operator completes or errors, so that source Observable is only subscribed as long as it has to, even when combined with operators that do not immediately unsubscribe
from 'first' when it completes or errors.

Introduces test-only `doNotUnsubscribe` operator, which simulates operators that are unreliable with managing subscriptions (like `switchMap` for example - see #2459).

Closes #2455

This issue is a concrete example of more general one: #2459

### DIFF
--- a/spec/helpers/doNotUnsubscribe.ts
+++ b/spec/helpers/doNotUnsubscribe.ts
@@ -1,0 +1,16 @@
+///<reference path='../../typings/index.d.ts'/>
+import * as Rx from '../../dist/cjs/Rx';
+
+export function doNotUnsubscribe<T>(ob: Rx.Observable<T>): Rx.Observable<T> {
+  return ob.lift(new DoNotUnsubscribeOperator());
+}
+
+class DoNotUnsubscribeOperator<T, R> implements Rx.Operator<T, R> {
+  call(subscriber: Rx.Subscriber<R>, source: any): any {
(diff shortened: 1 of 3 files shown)


benlesh:
While this is a fix, it also might be a breaking change for some. @staltz, @kwonoj what do you think? Next major version?

staltz:
Yeah it's major. If it could possibly break existing behavior for *anyone* out there, it's a breaking change. There's no shame in bumping the major num. In the changelog we can say "this will probably not affect you, but" type of message.

Issue #1625, opened 2016-04-14
trailing behavior on throttleTime


matthewwithanm:
It's possible that this could be considered a feature request, however, I think the implemented behavior is actually not the generally desired one. Underscore and lodash both default their throttles' `leading` and `trailing` options to true.

**RxJS version:** 5.0.0-beta.6

**Code to reproduce:**

``` js
Observable.interval(1000)
  .take(4)
  .merge(Observable.never())
  .do(x => console.log('<-', x))
  .throttleTime(2000)
  .subscribe(
    x => console.log('->', x),
    err => console.log('err', err),
    _ => console.log('complete')
  );
```

**Expected behavior:**

```
<- 0
-> 0
<- 1
<- 2
-> 2
<- 3
-> 3
```

**Actual behavior:**

```
<- 0
-> 0
<- 1
<- 2
-> 2
<- 3
```

**Additional information:**

I would argue that even if the source observable completes (i.e. the `.merge(Observable.never())` line is removed), the output should contain the final `-> 3`.

matthewwithanm:
Sorry, maybe I'm just confused about the purpose of `throttleTime` and I should be using `sampleTime` instead.

Though in Rx 4, that will include `-> 3` regardless of completion (my desired behavior) and in Rx 5 it doesn't include the final value if the source completes first. 😕

blesh:
Yeah, currently the behavior of throttleTime matches Rx4 exactly when comparing this code. Is it that you just want to add the last value to your stream?

```
let source = Observable.interval(1000).share();

source.take(4)
  .merge(Observable.never())
  .do(x => console.log('<-', x))
  .throttleTime(2000)
  .concat(source.last())
  .subscribe(
    x => console.log('->', x),
    err => console.log('err', err),
    _ => console.log('complete')
  );
```

blesh:
Currently the behavior matches the behavior shown in RxJS 4. Testing this organically without a virtual scheduler might yield more random results, but it was the same for me.

As for the change to `sample` I feel like this was something that was asked for by @staltz or @trxcllnt and discussed, but I could be wrong.

blesh:
... here is the relevant discussion about the sample operator. Incidentally, this is also the birthplace for the  idea of the marble tests: https://github.com/ReactiveX/rxjs/issues/150

matthewwithanm:
[...]

PR #2465, opened 2017-03-15
Leading/Trailing throttle and throttleTime configuration

benlesh:
related to #1625  see commits for details.

### DIFF
--- a/spec/operators/throttle-spec.ts
+++ b/spec/operators/throttle-spec.ts
@@ -357,7 +357,7 @@ describe('Observable.prototype.throttle', () =>  {
   });
 
   describe('throttle(fn, { leading: false, trailing: true })', () => {
-    asDiagram('throttle(fn, { leading: true, trailing: true })')('should immediately emit the first value in each time window', () =>  {
+    asDiagram('throttle(fn, { leading: false, trailing: true })')('should immediately emit the first value in each time window', () =>  {
       const e1 =   hot('-a-xy-----b--x--cxxx--|');
       const e1subs =   '^                     !';
       const e2 =  cold( '----|                 ');
--- a/spec/operators/throttle-spec.ts
+++ b/spec/operators/throttle-spec.ts
@@ -337,4 +337,40 @@ describe('Observable.prototype.throttle', () =>  {
       }
     );
   });
+
+  describe('throttle(fn, { leading: true, trailing: true })', () => {
+    asDiagram('throttle(fn, { leading: true, trailing: true })')('should immediately emit the first value in each time window', () =>  {
+      const e1 =   hot('-a-xy-----b--x--cxxx--|');
+      const e1subs =   '^                     !';
+      const e2 =  cold( '----|                 ');
+      const e2subs =  [' ^   !                 ',
+                       '          ^   !        ',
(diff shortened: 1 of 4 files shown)


spec/operators/throttleTime-spec.ts
```diff
@@ -141,4 +142,32 @@ describe('Observable.prototype.throttleTime', () => {
...
+
+      const result = e1.throttleTime(t, rxTestScheduler, { leading: true, trailing: true });
+
+      expectObservable(result).toBe(expected);
+      expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    });
+  });
+
+  describe('throttleTime(fn, { leading: false, trailing: true })', () => {
+    asDiagram('throttleTime(fn, { leading: true, trailing: true })')('should immediately emit the first value in each time window', () =>  {
```
david-driscoll: `{ leading: true, trailing: true }` should be `{ leading: false, trailing: true }`

spec/operators/throttle-spec.ts
```diff
@@ -337,4 +337,40 @@ describe('Observable.prototype.throttle', () =>  {
...
+      const result = e1.throttle(() =>  e2, { leading: true, trailing: true });
+
+      expectObservable(result).toBe(expected);
+      expectSubscriptions(e1.subscriptions).toBe(e1subs);
+      expectSubscriptions(e2.subscriptions).toBe(e2subs);
+    });
+  });
+
+  describe('throttle(fn, { leading: false, trailing: true })', () => {
+    asDiagram('throttle(fn, { leading: true, trailing: true })')('should immediately emit the first value in each time window', () =>  {
```
david-driscoll: `{ leading: true, trailing: true }` should be `{ leading: false, trailing: true }`

david-driscoll:
Is it fair to say...

* `x.throttleTime(100, { trailing: true })` is the same as `x.auditTime(100)`?
* `x.throttleTime(100, { leading: true, trailing: true })` is the same as `Observable.merge(x.throttleTime(100), x.auditTime(100))`?

Also would it be possible (perhaps as a separate PR!) to give the same treatment to `debounce` and `debounceTime`?

We have no operator that compares to [lodash's `debounce(x => {}, { leading: true})`](https://lodash.com/docs/4.17.4#debounce)

benlesh:
> Also would it be possible (perhaps as a separate PR!) to give the same treatment to debounce and debounceTime?

That's the idea.

benlesh:
@david-driscoll the changes are made. Good catch. I put them in a separate commit so I didn't have to pull everything apart and change two commits.

david-driscoll:
Looks good to me.  @kwonoj @trxcllnt @jayphelps any complaints?

benlesh:
The potential bug found in `throttle` (it existed prior, and this PR does NOT address it) I mention in my commit messages is listed here: #2466

connorwyatt:
Hi, I would be interested in knowing when you guys think this will be in a release? We have a workaround that we can remove once this goes in and I was wondering when you think this will be merged.

Thanks

lee-tunnicliffe:
I could also do with this change for my project. Any update would be greatly appreciated.

PR #2470, opened 2017-03-16
fix(takeWhile): force unsubscribe when it completes or errors

mpodlasin:
Force unsubscribe on `takeWhile` when it completes or errors, even when used with operator that does not unsubscribe reliably, so that source Observable is never subscribed longer than it needs to.

**Related issues:**

`takeWhile` suffered from the same problem as `first`: #2455 
It is a concrete case of more general problem: #2459

### DIFF
--- a/spec/helpers/doNotUnsubscribe.ts
+++ b/spec/helpers/doNotUnsubscribe.ts
@@ -1,0 +1,16 @@
+///<reference path='../../typings/index.d.ts'/>
+import * as Rx from '../../dist/cjs/Rx';
+
+export function doNotUnsubscribe<T>(ob: Rx.Observable<T>): Rx.Observable<T> {
+  return ob.lift(new DoNotUnsubscribeOperator());
+}
+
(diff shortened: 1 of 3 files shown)


benlesh:
Like #2463, I worry that this will be a breaking change for some people, and we might have to release this with the next major (which we should probably do pretty soon)... @staltz? @trxcllnt?

trxcllnt:
@benlesh wouldn't all bug fixes count as breaking changes then?

staltz:
> wouldn't all bug fixes count as breaking changes then?

Yup, and there's no shame in that. If people are unknowingly relying on buggy behavior, when the bug gets fixed, their app will behave differently, and who knows, that new behavior might look like a bug to their business logic.

kwonoj:
Interesting, to clarify, I don't object to make this breaking changes for major but just for curiosity - if some behavior was unintended / undocumented and ppl was using it by just found it, still we consider fixing it has major breaking changes? (this one looks in those way to me)

PR #2472, opened 2017-03-16
Revert "feat(delayWhen): add index to the selector function"

kwonoj:
Reverts ReactiveX/rxjs#2409


kwonoj:
@benlesh this is reverting PR as discussed in #2409 , /cc'ed you as reviewer.

PR #2474, opened 2017-03-17
fix(forEach): fix a temporal dead zone issue in forEach.

mprobst:
According to the ES6 spec and the implementation in V8, accessing a lexically scoped construct like const or let before it is assigned is a ReferenceError. Unlike var, it is not implicitly undefined. Because the closure handed to subscribe is immediately invoked and accesses `subscription` before it is assigned, this causes a ReferenceError in compliant engines.

The error is only triggered when running in plain ES6, i.e. without transpilation.

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -139,7 +139,10 @@ export class Observable<T> implements Subscribable<T> {
     }
 
     return new PromiseCtor<void>((resolve, reject) => {
-      const subscription = this.subscribe((value) => {
+      // Must be declared in a separate statement to avoid a RefernceError when
+      // accessing subscription below in the closure due to Temporal Dead Zone.
+      let subscription: Subscription;
+      subscription = this.subscribe((value) => {
         if (subscription) {
           // if there is a subscription, then we can surmise
           // the next handling is asynchronous. Any errors thrown


src/Observable.ts
```diff
@@ -139,7 +139,10 @@ export class Observable<T> implements Subscribable<T> {
     }
 
     return new PromiseCtor<void>((resolve, reject) => {
-      const subscription = this.subscribe((value) => {
+      // Must be explicitly assigned to avoid RefernceError when accessing
+      // subscription below in the closure due to Temporal Dead Zone.
+      let subscription: Subscription = undefined;
```
Brooooooklyn: `let subscription: Subscription;`

mprobst: Done.

kwonoj:
Curiousity question: how previous code can be executed without transpilation to hit those errorneus cases?

mprobst:
@kwonoj if you run the code in an ES6 VM, it'll fail. But you can repro this much more easily locally:

```js
> function call(c) { return c(); }
undefined
> function undef() { const foo = call(() => foo ? 1 : 2); }
undefined
> undef()
ReferenceError: foo is not defined
    at call (repl:1:55)
    at call (repl:1:38)
    at undef (repl:1:44)
    at repl:1:1
    at ContextifyScript.Script.runInThisContext (vm.js:23:33)
    at REPLServer.defaultEval (repl.js:340:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.onLine (repl.js:537:10)
    at emitOne (events.js:101:20)
> function undef() { let foo; foo = call(() => foo ? 1 : 2); }
undefined
> undef()
undefined
```

kwonoj:
Ah, question was since code is TS anyway, it should be transpiled via compiler so curious how did original could be executed. Just curiousity question though.

mprobst:
If you transpile TypeScript with target language ES6, you get ES6 :-)

kwonoj:
Oh thought compiler does something like hoisting automatically. Thx for clarification.

mprobst:
@kwonoj can you merge? I don't have write access.

kwonoj:
No, I intentionally left this to get second eye and release planning. /cc @benlesh

benlesh:
thanks @mprobst!

PR #2478, opened 2017-03-19
fix(catch): fix catch typings

ghetolay:
**Description:**

The catch operator is returning an observable with the type of the innerObservable (observable returned when error occur). 
It should be a union of the outerObservable's type and the innerObservable's type :

```ts
function _catch<T, R>(this: Observable<T>, selector: (err: any, caught: Observable<T>) => ObservableInput<R>): Observable<T | R>
```

### DIFF
--- a/src/operator/catch.ts
+++ b/src/operator/catch.ts
@@ -64,13 +64,13 @@ import { subscribeToResult } from '../util/subscribeToResult';
  * @name catch
  * @owner Observable
  */
-export function _catch<T, R>(this: Observable<T>, selector: (err: any, caught: Observable<T>) => ObservableInput<R>): Observable<R> {
+export function _catch<T, R>(this: Observable<T>, selector: (err: any, caught: Observable<T>) => ObservableInput<R>): Observable<T | R> {
   const operator = new CatchOperator(selector);
   const caught = this.lift(operator);
   return (operator.caught = caught);
 }
 
-class CatchOperator<T, R> implements Operator<T, R> {
+class CatchOperator<T, R> implements Operator<T, T | R> {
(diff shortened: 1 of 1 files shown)


kwonoj:
Would you mind share examples of `It should be a union of the outerObservable's type`?

david-driscoll:
Would be great to see the use case for this... `catch` allows you to handle an error and decide what to do.  There is nothing stopping you from doing something like...

```typescript
o.catch((e, c) => c);
```

Which would make `T` -> `R`.  With this change then the output would always be a special union of two potentially different types.

ghetolay:
Any example where you would return an observable with a different type on the catch would reveal the need of union but I think this apply for all case.

if we take the first example given on catch : 
```ts
Observable.of(1, 2, 3, 4, 5)
   .map(n => {
 	   if (n == 4) {
 	     throw 'four!';
     }
	   return n;
   })
   .catch(err => Observable.of('I', 'II', 'III', 'IV', 'V'))
```
`x` should be of type `number | string` and not just `number`.

And in the case you use the same type like :

```ts
obsv$ = Observable.of(1)
  .catch( err => Observable.of(0) );
``` 
the type of `obsv$` is `Observable<number>` but not because the catch returns that type, it's because the union `number | number` is equal to type `number`.

ghetolay:
> With this change then the output would always be a special union of two potentially different types.

@david-driscoll isn't it what exactly happens after a catch ? it's either the source observable of the one defined on catch, you can't know which one and thus can't know for sure the type of emitted item.

david-driscoll:
That is actually a fair point, I had forgotten about how that behavior would flow.  I have been corrected 👍

PR #2479, opened 2017-03-19
fix(mergeAll): add source subscription to composite before actually subscribing

### LINKED ISSUE #2476: IteratorObserable's never termines when within Observable.merge
**RxJS version:**
5.2.0

**Code to reproduce:**
unit tests added to IteratorObservable-spec.js
```
  it('should finalize generators when merged if the subscription ends', () => {
    const iterator1 = {
      finalized: false,
      next() {
        return { value: 'duck', done: false };
      },
      return() {
        this.finalized = true;
      }
    };

    const iterable1 = {
      [Rx.Symbol.iterator]() {
        return iterator1;
      }
    };

    const iterator2 = {
      finalized: false,
      next() {
        return { value: 'duck', done: false };
      },
      return() {
        this.finalized = true;
      }
    };

    const iterable2 = {
      [Rx.Symbol.iterator]() {
        return iterator2;
      }
    };

    const results = [];

    const i1 = IteratorObservable.create(iterable1)
    const i2 = IteratorObservable.create(iterable2)
    Rx.Observable.merge(i1, i2)
      .take(3)
      .subscribe(
        x => results.push(x),
        null,
        () => results.push('GOOSE!')
      );

    // never even get here
    expect(results).to.deep.equal(['duck', 'duck', 'duck', 'GOOSE!']);
    expect(iterator1.finalized).to.be.true;
    expect(iterator2.finalized).to.be.true;
  });
```

**Expected behavior:**

See test

**Actual behavior:**

The test never terminates (until V8 runs out of memory that is).

**A simpler example*
Given:
```
const myInfiniteIterator1 = ...
const myInfiniteIterator2 = ...
```
This terminates:
```
Rx.Observable.from(myInfiniteIterator1).take(3)
```
But this never terminates
```
Rx.Observable.merge(…Rx.Observable.from(myInfiniteIterator1, myInfiniteIterator2).take(3)
```

mpodlasin:
Add subscriptions for source Observables to mergeAll composite subscription before actually subscribing to any of these Observables, so that if source Observable emits synchronously and consumer of mergeAll unsubscribes at that moment (for example `take` operator), subscription to source is unsubscribed as well and Observable stops emitting.

Closes #2476

### DIFF
--- a/spec/operators/mergeAll-spec.ts
+++ b/spec/operators/mergeAll-spec.ts
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import { mergeAll, mergeMap } from 'rxjs/operators';
+import { mergeAll, mergeMap, take } from 'rxjs/operators';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 import { throwError, from, of, Observable } from 'rxjs';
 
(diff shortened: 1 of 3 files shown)


spec/operators/mergeAll-spec.ts
```diff
@@ -415,4 +415,34 @@ describe('Observable.prototype.mergeAll', () => {
...
+    const i1 = Rx.Observable.from(iterator1 as any);
+    Rx.Observable.merge(i1, i1)
+      .take(3)
+      .subscribe(
+        x => results.push(x),
+        null,
+        () => results.push('GOOSE!')
+      );
+
+    // never even get here
```
benlesh: This comment is confusing. Why wouldn't it get here?

Podlas29:
@benlesh done.

cartant:
@benlesh I've rebased this PR.

PR #2484, opened 2017-03-21
Remove redundant overload signature

masaeedu:
**Description:**
The first overload signature for `reduce` is redundant, since it is simply a special case of the `reduce<T, R>` overload. Its presence breaks type inference when the type of the source observable items and the accumulator differ.


kwonoj:
This seems related (or maybe dupe) to https://github.com/ReactiveX/rxjs/pull/2382.

masaeedu:
@kwonoj Yup, it's a dupe.

PR #2491, opened 2017-03-24
perf(pluck,bufferTime,asObservable): add performance tests for pluck(), bufferTime() and asObservable() operators

martinsik:
**Description:**

I've noticed that `pluck()`, `bufferTime()` and `asObservable()` operators don't have any performance tests so this PR adds them. All three of them have their RxJS 4 counterparts.

There're more operators without performance tests (for example `audit()` or `windowToggle()`) but these don't have any RxJS 4 alternative to compare them to and the current benchmark runner requires always two tests per suite. But this is probably for a separate another issue.


benlesh:
Thanks, @martinsik!

PR #2497, opened 2017-03-27
fix(IteratorObservable): get new iterator for each subscription

### LINKED ISSUE #2496: Second subscription to the same observable is not starting a sequence if the observable was built from a typed array
**RxJS version:**
5.2.0

**Code to reproduce:**
```
let x_source = [1,2];
let x_obs = Rx.Observable.from(x_source);
let x_obs1 = x_obs.map(v => `x_obs1: ${v}`).subscribe(console.log);
let x_obs2 = x_obs.map(v => `x_obs2: ${v}`).subscribe(console.log);

let y_source = new Int32Array([1,2]);
let y_obs = Rx.Observable.from(y_source);
let y_obs1 = y_obs.map(v => `Int32Array 1st observer: ${v}`).subscribe(console.log);
let y_obs2 = y_obs.map(v => `Int32Array 2nd observer: ${v}`).subscribe(console.log);
```

**Expected behavior:**
```
x_obs1: 1
x_obs1: 2
x_obs2: 1
x_obs2: 2
Int32Array 1st observer: 1
Int32Array 1st observer: 2
Int32Array 2nd observer: 1
Int32Array 2nd observer: 2
```

**Actual behavior:**
```
x_obs1: 1
x_obs1: 2
x_obs2: 1
x_obs2: 2
Int32Array 1st observer: 1
Int32Array 1st observer: 2
```

**Additional information:**
For some reason, subscribing to an observable, that is built from typed array (Int32Arrat, Float32Array, etc), in the 2nd time is not initiates a sequence. The 1st subscription is working well.
Observables built from classic arrays don't have such behavior.

kwonoj:
**Description:**
Our `IteratorObservable` gets an iterator from the object once only when it's constructed, so any subsequent subscription is not able to iterate since the first subscription will make iterator completes.

This PR updates behavior to get corresponding new iterator per each subscription, allows any subscription can correctly iterate object. I'm labeling this for next patch version, since I believe this isn't breaking changes.

- closes #2496

### DIFF
--- a/spec/observables/IteratorObservable-spec.ts
+++ b/spec/observables/IteratorObservable-spec.ts
@@ -47,6 +47,24 @@ describe('IteratorObservable', () => {
       );
   });
 
+  it('should get new iterator for each subscription', () => {
+    const expected = [
+      Rx.Notification.createNext(10),
+      Rx.Notification.createNext(20),
+      Rx.Notification.createComplete()
+    ];
+
+    const e1 = IteratorObservable.create<number>(new Int32Array([10, 20])).observeOn(rxTestScheduler);
+
+    let v1, v2: Array<Rx.Notification<any>>;
+    e1.materialize().toArray().subscribe((x) => v1 = x);
+    e1.materialize().toArray().subscribe((x) => v2 = x);
+
+    rxTestScheduler.flush();
+    expect(v1).to.deep.equal(expected);
+    expect(v2).to.deep.equal(expected);
+  });
+
   it('should finalize generators if the subscription ends', () => {
     const iterator = {
       finalized: false,
--- a/src/observable/IteratorObservable.ts
+++ b/src/observable/IteratorObservable.ts
@@ -11,8 +11,6 @@ import { Subscriber } from '../Subscriber';
  * @hide true
  */
 export class IteratorObservable<T> extends Observable<T> {
-  private iterator: any;
-
   static create<T>(iterator: any, scheduler?: IScheduler): IteratorObservable<T> {
     return new IteratorObservable(iterator, scheduler);
   }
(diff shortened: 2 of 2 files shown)


benlesh:
This definitely a bug fix. The issue is that it's going to be a breaking change for *somebody*. So we might want to put this in the next major (which we should publish soon). But it's debatable.

The problem, better outlined is this:

```ts
const data = new Int32Array([1, 2, 3, 4, 5]);

const source$ = Observable.from(data);

source$.toArray().subscribe(x => console.log(x)); // [1, 2, 3, 4, 5]

source$.toArray().subscribe(x => console.log(x)); // []
```

That's not what users would expect at all. It should behave like the standard Array implementation. However, we might have *one* crazy person relying on this behavior.

The work around to get that behavior again would be to pass the actual `iterator` itself:

```ts
const data = new Int32Array([1, 2, 3, 4, 5]);
const iterator = data[Symbol.iterator]();

// "new" appropriate behavior
const source1$ = Observable.from(data);
source1$.toArray().subscribe(x => console.log(x)); // [1, 2, 3, 4, 5]
source1$.toArray().subscribe(x => console.log(x)); // [1, 2, 3, 4, 5]

// "old" behavior, if desired
const source2$ = Observable.from(iterator);
source2$.toArray().subscribe(x => console.log(x)); // [1, 2, 3, 4, 5]
source2$.toArray().subscribe(x => console.log(x)); // []
```

staltz:
> The issue is that it's going to be a breaking change for somebody.

Pros of major: no one will be surprised, no code will break in production. Cons: *it will be a different number*. Pros of patch: *it will look like a harmless corner-casey number bump, like this is a corner case*. Cons of patch: there is a positive probability someone will be surprised, and that someone could be a user of an app in production.

I vote for major.

> Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API.

http://semver.org/#spec-item-8

It depends what we consider our public API. Usually its tests + docs. We don't have any test that asserts the buggy behavior as normal. In our docs we say

> Creates an Observable from an Array, an array-like object, a Promise, an iterable object, or an Observable-like object.

It could be understood that the **Iterator creation side effect** happens once, independent of multiple subscriptions, **like the promise constructor side effect does as well**.

jayphelps:
I think this is a major bump for the reasons outline by Andre.

kwonoj:
Removing labels for patch to avoid accidental check in.

benlesh:
@kwonoj can you resolve conflicts against `next` please?

kwonoj:
sure.

kwonoj:
@benlesh rebased to `next`.

kwonoj:
Now this can go into master as breaking change.

PR #2505, opened 2017-03-30
fix(PromiseObservable): fromPromise does not accept PromiseLike object

MidnightWonderer:
I can not use `Observable.fromPromise` with promises from [when](https://github.com/cujojs/when).  
The class is clearly documented that it should accept Promises/A+ spec compliant Promise.  
I think the type should be relaxed.

### DIFF
--- a/spec/observables/from-promise-spec.ts
+++ b/spec/observables/from-promise-spec.ts
@@ -72,6 +72,25 @@ describe('Observable.fromPromise', () => {
     });
   });
 
+  it('should accept PromiseLike object for interoperability', (done: MochaDone) => {
+    class CustomPromise<T> implements PromiseLike<T> {
+      constructor(private promise: PromiseLike<T>) {
+      }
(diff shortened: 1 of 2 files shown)


kwonoj:
And interestingly we didn't have test cases for this cases as well. @MidnightWonderer , would you mind to update PR to include one test cases for given changes?

MidnightWonderer:
Hi @kwonoj, I am not really sure how to test the changes because they are all about the type declaration itself and not the implementation details.  
  
How about a new class exposes only what defined in `PromiseLike` interface as in my new updates?  
If the test fails it will fail at the transpilation step, not the execution step.  
  
I can't see why the test affects code coverage in any way. (unless you count the coverage inside the spec itself)

kwonoj:
I guess this is for minor, labeled it for now.

staltz:
PR LGTM

PR #2506, opened 2017-03-30
Update FromEventObservable.ts

gregorym:
It took me forever to figure out that I also needed the removeEventListener function so I'm making it explicit.


kwonoj:
I don't think updated error message explains possible erroneous case correctly, cause `EventObservable` accepts different forms of eventsource which possibly doesn't have `add/remove` pair. For example, `JQueryStyleEventEmitter` has `on` and `off` and if consumer supplied jquery style emitter and hit error message about `add/remove` only, it can possibly make confused.

gregorym:
Yeah you're right. I didn't realize their could be multiple sources.

PR #2508, opened 2017-03-31
feat(smooth): Adds the smooth() operator, the cool operator

jayphelps:
Demo: http://jsbin.com/wufawuq/edit?js,output (turn your volume up)
Prior art: https://en.wikipedia.org/wiki/Smooth_Operator

This is mostly a port of this video to RxJS https://www.youtube.com/watch?v=4TYv2PhG89A

Credit: I can't remember who first came up with this. May have been @benlesh. Let's just say it was all of us.

PR #2512, opened 2017-04-03
feat(smooth): stop to ship `smooth()` as a part of default bundled operator

saneyuki:
BREAKING CHANGE: `smooth()` is not a part of `Observable.prototype` by default.
Please use `import 'rxjs/add/operator/smooth'`.

Motivation
------------

I love software's easter egg and april fool joke.
I don't have any objections about that adding joke codes like `smooth()`
into our repository.

However, I strongly think we should not ship it as a part of default
bundled operator (in other words, I think it should not be loaded by
default in `src/Rx.ts`).

Default bundling `smooth()` causes these bad effects

- *__Increasing the file size of user application by joke code by default__*
  - I don't have any objection to add an acutual useful code,
    and increase the file size of user application by it. It is user
    choise. But this is joke code. It's not essential for actual
    usecase.
  - This is not a code comment which we can remove by some minifiers.
    So this enforces to increasing file size all users who use
    `import  { Observable } from 'rxjs'` style.

Related issue (if exists)
------------

https://github.com/ReactiveX/rxjs/pull/2508


jayphelps:
This was just an april fools joke and will be removed. It was only available in the special `5.2.1-smooth` package, it won't be in the real `5.2.1`

jayphelps:
Ah ha! So I see now that npm has the `latest` tag set to `5.2.1-smooth` so if anyone does `npm install rxjs` (any version) they'll get `5.2.1-smooth` which isn't "ideal" but not really harmful AFAICT. Lmk if I'm missing something else. I'll try to get @benlesh to merge #2516 asap and publish a real `5.2.1`

benlesh:
Oh... didn't think about updating the `latest` tag in npm. :(  Welp... Not doing that again. I was only concerned about people with `^5.2.0` in there package.json... those people should be fine.

Issue #2509, opened 2017-04-01
Feature request: debug operator


Boshen:
Floating around the internet there is a mysterious debug operator that looks something like this:
```
Observable.prototype.debug = (message: string) => {
  return this.do(
    (next) => {
      console.log(message, next)
    },
    (err) => {
      console.error(message, err)
    },
    () => {
      console.info(message, 'completed')
    }
  );
};
```

I'd like to see it in the core. Save us some typing, and remove issues around people doing  `.do(console.log)`.

Cheers :smiley:

PR #2513, opened 2017-04-03
feat(debug): Adds the debug() operator

Boshen:
#2509

### DIFF
--- a/spec/operators/debug-spec.ts
+++ b/spec/operators/debug-spec.ts
@@ -1,0 +1,46 @@
+import {expect} from 'chai';
+import * as sinon from 'sinon';
+import * as Rx from '../../dist/cjs/Rx';
+
+const Observable = Rx.Observable;
+
+/** @test {debug} */
+describe('Observable.prototype.debug', () => {
+
+  let logNext, logError, logComplete
+
+  beforeEach(() => {
+    logNext = sinon.stub(console, 'log')
+    logError = sinon.stub(console, 'error')
+    logComplete = sinon.stub(console, 'info')
+  })
+
+  afterEach(() => {
+    logNext.restore()
+    logError.restore()
+    logComplete.restore()
+  })
+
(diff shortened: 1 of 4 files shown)


kwonoj:
Perspective of usefullness I agree this is kind of common pattern, but I don't think this is needed operator as core level. Given changes can be trivially implemented in user code, while making this into core does not provides addition feature level enhancement or improvement compare to use level. I'd rather recommend to something similar to 

```
const dbg = (o) => {
 if(prod) return o; //user code specific prod detection
 return o.do (logger.xxx...); //or console, depends on code
}

OtherObservable.let(dbg).otheroperator ()...
```

Achieves similar shorthanded expression.

Boshen:
Ha .. I've never noticed the `let` operator until you mention it. Your solution works for me since it's so trivial to add to a project. But I still think this is a nice thing to have in the core because it removes a lot of friction for all our developers :-)

Now I wonder whether there is separate repo for things like this, where a common pattern is abstracted, and it just combines operators from the core.

staltz:
Hi @Boshen , thank you for taking your time to send us this PR. I understand your desire for this operator (I myself built a stream library xstream that has [debug](https://github.com/staltz/xstream/#debug)), but I want to explain how adding operators to RxJS may bring problems.

Every operator added means more maintenance to be done, more analysis paralysis for a developer looking for core (essential) operators, and overall weight of this library. @kwonoj pointed out a good solution for this. It would be awesome to make an RxJS **ecosystem**, not an RxJS framework. You can publish the debug operator as a separate npm package, and then users can import that and use it with `let()`. Right now, it seems like there should be one more operator in RxJS core, but I want to help show how a lot of operators could simply be served as separate packages. We don't want this library to get 300+ operators, right?

Boshen:
Right. Thank you all for the responses @kwonoj and @staltz :-)

Issue #2488, opened 2017-03-23
suggestion: consider improving the type signature for Observable.catch


aluanhaddad:
**RxJS version: 5.2.0**

**Code to reproduce:**
**case 1:** `Observable<number>` decays to `Observable<any>`
```ts
import {Observable} from 'rxjs/Rx';

const o = Observable.of(1)
  .catch((err, caught) => Observable.throw(err))
```
**case 2:** `Observable<number>` decays to `Observable<{}>`
```ts
import {Observable} from 'rxjs/Rx';

const o = Observable.of(1)
  .catch((err, caught) => {
    throw Error(err);
  });
```

**Expected behavior:**
For `o` to have the type `Observable<number>` in both cases.

**Actual behavior:**
`o` has type `Observable<any>` in **case 1**.

`o` has type `Observable<{}>` in **case 2**.

This can mask serious bugs such as
```ts
function f() {
  const o = Observable.of(1)
    .map(() => {
      console.log(1); 
    }) // programmer likely meant to return something
    .catch((err, caught) => Observable.throw(err));
  return o;
}
```
Where `f` should have type `Observable<void>`, likely due to programmer error, but `any` absorbs even `void`.

**Additional information:**
For better or worse, the catch and rethrow idiom seems to have overtaken RxJS, sometimes even when they only rethrow (not even logging the error) developers insist on explicitly catching and rethrowing all exceptions. I suspect the Angular docs are the primary reason for this, but that is beside the point except to remark that clearer usage guidelines for these operators would be helpful. Catching and only rethrowing doesn't make much sense to me but I am very new to RxJS.

For reference here is what the idiom I see daily (at work and on Stack Overflow) looks like
```ts
import { Observable } from 'rxjs/Observable';
import 'rxjs/add/operator/map';
import 'rxjs/add/operator/catch';

export interface Something {
  name: string;
}

export class SomeService {
  getItems() {
    return this.http('api/something')
      .map(x => (x.json() as Something).name) // Observable<string>
      .catch((err, caught) => Observable.throw(err.json())); // Observable<any>
  }
}
```
The problem with this pattern, and the pattern really is pervasive in the Angular at least, is that chaining `catch` removes all type information unless the user actually uses catch to recover by supplying a new Observable. If they return the result of an `Observable.throw` call all further operations receive `Observable<any>`.

To fix the type issue, I suggest that the signatures of `catch` and `_catch` be changed from
```ts
export declare function _catch<T, R>(
[...]

PR #2520, opened 2017-04-03
feat(catch): make rethrow idiom type safe

aluanhaddad:
**Description:**
For better or worse, the catch and rethrow idiom seems to have overtaken RxJS. I suspect the Angular docs are the primary reason for this trend but that is beside the point except that it gives me the opportunity to remark that clearer usage guidelines for operators would be helpful.

For reference, this is what the code that I see daily, both at work and on Stack Overflow, looks like:
```ts
import 'rxjs/add/operator/map';
import 'rxjs/add/operator/catch';

export interface Something {
  name: string;
}

export class SomeService {
  getItems() {
    return this.http('api/something')
      .map(x => x.json() as Model) // Observable<Model>
      .catch((err, caught) => Observable.throw(err.json())); // Observable<any>
  }
}
```
The problem with this idiom, and it really is pervasive in Angular at least, is that chaining `catch` removes all type information unless the user actually uses it to recover by supplying another strongly Observable.

While recovering from an error seems like the primary purpose of the operator, the rethrow idiom is widespread.

If they return the result of an `Observable.throw` call all further operations receive `Observable<any>`.

This PR changes the type of `_catch` by introducing an overload which explicitly accepts a projection to `ErrorObservable` and preserves the original value type of the observable in that case.

I have also added a test verifying that `any` is no longer the value type of the chained observable by leveraging perhaps the only restriction on _any_, that it cannot be the object of an `extends` clause.

#2488

### DIFF
--- a/src/operator/catch.ts
+++ b/src/operator/catch.ts
@@ -4,7 +4,7 @@ import { Observable, ObservableInput } from '../Observable';
 
 import { OuterSubscriber } from '../OuterSubscriber';
 import { subscribeToResult } from '../util/subscribeToResult';
-import { ErrorObservable } from '../Observable/ErrorObservable';
+import { ErrorObservable } from '../observable/ErrorObservable';
 
 /**
  * Catches errors on the observable to be handled by returning a new observable or throwing an error.
--- a/src/operator/catch.ts
+++ b/src/operator/catch.ts
@@ -4,6 +4,7 @@ import { Observable, ObservableInput } from '../Observable';
 
 import { OuterSubscriber } from '../OuterSubscriber';
 import { subscribeToResult } from '../util/subscribeToResult';
+import { ErrorObservable } from '../Observable/ErrorObservable';
 
 /**
  * Catches errors on the observable to be handled by returning a new observable or throwing an error.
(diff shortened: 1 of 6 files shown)


aluanhaddad:
@kwonoj Any idea why this might be failing? The file is created correctly and the tests pass when I run a clean build locally.

src/operator/catch.ts
```diff
@@ -64,6 +65,8 @@ import { subscribeToResult } from '../util/subscribeToResult';
  * @name catch
  * @owner Observable
  */
+export function _catch<T>(this: Observable<T>, selector: (err: any, caught: Observable<T>) => ErrorObservable): Observable<T>;
+export function _catch<T, R>(this: Observable<T>, selector: (err: any, caught: Observable<T>) => ObservableInput<R>): Observable<T | R>;
```
paulpdaniels: I guess I don't understand enough about the typescript compiler. Why do you need to duplicate this signature? Since it is directly below?

aluanhaddad: When you overload a function _definition_, you need to declare an additional overload implementation as the implementation signature is no longer part of the type.

kwonoj:
@aluanhaddad check travis log, https://travis-ci.org/ReactiveX/rxjs/jobs/218251237#L260 seems you have incorrect capitcal case letter in path.

src/operator/catch.ts
kwonoj: https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Functions.md#overloads will gives some more details of signature overload in TS.

kwonoj:
I guess this is patch, as it doesn't break any current behavior?

aluanhaddad:
@kwonoj it would break someone who was incorrectly using the `Observable<any>` type that arose from previous uses of catch. That is to say it would cause them to suffer compilation errors.

For example, consider this snippet (assume Angular 2)
```ts
getAll(): Observable<Model> {
  return this.http('api/resource/')
    .catch(e => Observable.throw(e));
}
```
currently this does not cause a compiler error because, even though the implementer forgot to apply `.map(response => response.json())` to the result. This is because `.catch(...)` always changes the type of chain into `Observable<any>`.

With this change, the type of the returned expression becomes `Observable<Response>` and the error is caught at compile time.

> `Observable<Response>` is not assignable to `Observable<Model>`.

kwonoj:
🤔  unlabeld patch for now.

kwonoj:
Per discussion, I changed base to `next` and labeled for major version.

PR #2521, opened 2017-04-03
fix(takeUntil): complete an observable when notifier completes

### LINKED ISSUE #2160: takeUntil docs wrong.

DzmitryShylovich:
Closes #2160 

From the original issue:
> I think the correct behaviour would be to **terminate when the notifier terminates**, however at the very least the docs should be in line with the specs.

As for me it's much more intuitive.

Might be a breaking change though...

### DIFF
--- a/spec/operators/takeUntil-spec.ts
+++ b/spec/operators/takeUntil-spec.ts
@@ -35,12 +35,12 @@ describe('Observable.prototype.takeUntil', () => {
     expectSubscriptions(e2.subscriptions).toBe(e2subs);
   });
 
-  it('should take all values when notifier is empty', () => {
+  it('should take values until notifier completes', () => {
     const e1 =     hot('--a--b--c--d--e--f--g--|');
-    const e1subs =     '^                      !';
+    const e1subs =     '^            !          ';
     const e2 =     hot('-------------|          ');
     const e2subs =     '^            !          ';
-    const expected =   '--a--b--c--d--e--f--g--|';
+    const expected =   '--a--b--c--d-|          ';
 
     expectObservable(e1.takeUntil(e2)).toBe(expected);
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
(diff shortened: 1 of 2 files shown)


jbaudanza:
I'm very much in favor of this change. If this isn't the way RxJs wants to go, it would be nice to have the docs updated to reflect this.

benlesh:
I'm not sure what this breaking change buys us. Is there an example where one couldn't just compose a completed observable to necessarily `next` with `concat`?

`source$.takeUntil(maybeEmpty$.concat(Observable.of(true)))`

I'm going to close this for now, because there has been limited interest, and it's a breaking change from any existing version of Rx, with no discernible benefit.

benlesh:
@DzmitryShylovich I really appreciate the effort, and I'm sorry to close this PR. The referenced issue this was meant to close was pretty clear the issue was with documentation and not with the functionality of the operator.

jbaudanza:
@benlesh Fair enough. Would you accept a PR to fix discrepancy with the docs?

DzmitryShylovich:
@jbaudanza see https://github.com/ReactiveX/rxjs/pull/2418

Issue #2519, opened 2017-04-03
Reduce with array input defaults the return type to an array of arrays


PrimalZed:
**RxJS version:** 5.3.0

**Code to reproduce:**
```
Observable.of([0, 1, 2, 3],[4, 5, 6, 7])
.reduce((acc, value) => acc.concat(value), [])
.map(numbers => numbers.map(n => n++))
.subscribe(value => console.log(JSON.stringify(value)));
```

**Expected behavior:** Can take an input type T[] and return type T[] without specifying typing.

**Actual behavior:** On input type T[], defaults return type T[][].

**Additional information:**
This was introduced with the fix to #2382 
My current workaround is to simply specify the types, e.g. `.reduce<number[], number[]>(...)`

aluanhaddad:
@PrimalZed The issue you are experiencing comes from the type of `of`, not from `reduce`. The reduce issue, where all seed parameters were optional, with `Array<ObservableElementType>` taking precedence was probably hiding the underlying issue.
For reference, here is the error I receive.
![image](https://cloud.githubusercontent.com/assets/5034656/24635121/94c76098-189f-11e7-8907-d0e4516257ec.png)

PrimalZed:
@aluanhaddad the type from `of` was intentional to be `Observable<number[]>`.  The `reduce` is to combine all the arrays returned by the observable into a single array, with `Array.prototype.concat`.

That error is the problem I'm experiencing, yes.  The return type of `reduce` is `Observable<number[][]>` rather than the desired `Observable<number[]>`.

Again, it's easily resolved by specifying some typing somewhere, but it seems like it should be able to pick it up.  Maybe a type definition of `reduce<T>(this: Observable<T[]>, accumulator: (acc: T[], value: T[], index: number) => T[], seed: T[]): Observable<T[]>` should be added?  Or is this use case just too fringe to deserve a type definition to support it?

david-driscoll:
The overload was added so that common case of reduce to an array of `T` would be easily supported without explicit casting.  It looks like in this case however it falls over.

```typescript
// without the overload
Observable.of(0, 1, 2, 3)
  .reduce((acc, value) => acc.concat(value), []); // [] would be {}[] and would require a cast to T[] to work correctly.
```

[...]

PR #2523, opened 2017-04-04
fix(reduce): change order of reduce overload type definitions

PrimalZed:
**Description:** Update overload type definitions of operator reduce so returning type T takes precedence over returning type T[].

**Related issue (if exists):** #2519

### DIFF
--- a/spec/operators/reduce-spec.ts
+++ b/spec/operators/reduce-spec.ts
@@ -316,6 +316,33 @@ describe('Observable.prototype.reduce', () => {
     });
   });
 
+  it('should accept T typed reducers when T is an array', () => {
+    type(() => {
+      let a: Rx.Observable<number[]>;
+      const reduced = a.reduce((acc, value) => {
+        return acc.concat(value);
+      }, []);
+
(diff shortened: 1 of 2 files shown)


david-driscoll:
Yeah we should have two tests.  The ones from #2519 should be sufficient.  Make sure you assign them to a specific type so that the types match both sides.

```typescript
const o1: Observable<number[]> = Observable.of([0, 1, 2, 3], [4, 5, 6, 7])
    .reduce((acc, value) => acc.concat(value), [])
    .map(numbers => numbers.map(n => n++))

const o2: Observable<number[]> = Observable.of(0, 1, 2, 3, 4, 5, 6, 7)
    .reduce((acc, value) => { acc.push(value); return acc; }, [])
    .map(numbers => numbers.map(n => n++))
```

PrimalZed:
Sorry, I rarely do pull requests on github.  I'll check to see how testing works in other pull requests and get this updated.

kwonoj:
What'd be release for this? /cc @david-driscoll

Issue #2299, opened 2017-01-19
Template for creating new operators


mattpodwysocki:
I would like to see some sort of RFC process for adding new operators to Rx itself.  Currently, there is no defined standard for acceptance criteria for what is included and not included in the library itself.

Taking some hints from the [Rust RFC Process](https://github.com/rust-lang/rfcs), we can glean a bit on how others handle substantial changes to the language/runtime/etc.  Changes such as these are not to be taken lightly, as Rx can suffer from an API bloat as we cover a wide array of use cases already.  If a new solution can simply be made from existing operators, that should be preferred, over writing a new core operator.  Conversely, it could also remain in user land for the users who really need said operator to import it as needed from another repository.

Here is what I think the minimum bar should be in order to add a new operator to Rx:

- A thorough explanation of the operator and its use cases.  This also includes error handling, completion, early unsubscription, and infinite case semantics just as we define in our unit tests.

- An explanation of how current operators cannot accomplish this task, or if they can, but performance is not ideal.

- Complete documentation including marble diagrams for further discussion

From there, the core team should enter discussions on whether to proceed or not.

staltz:
> Currently, there is no defined standard for acceptance criteria for what is included and not included in the library itself.

There is something: https://raw.githubusercontent.com/ReactiveX/rxjs/master/.github/PULL_REQUEST_TEMPLATE.md

> If a new solution can simply be made from existing operators, that should be preferred, over writing a new core operator. 

Complicated to follow this rule, since we can do `.mergeMap(x => Observable.of(f(x)))` instead of `map(f)`. Should we drop map?

> - A thorough explanation of the operator and its use cases. This also includes error handling, completion, early unsubscription, and infinite case semantics just as we define in our unit tests.
> - An explanation of how current operators cannot accomplish this task, or if they can, but performance is not ideal.
> - Complete documentation including marble diagrams for further discussion

Agreed a lot.

mattpodwysocki:
@staltz 

> There is something: https://raw.githubusercontent.com/ReactiveX/rxjs/master/.github/PULL_REQUEST_TEMPLATE.md

But that's already for a pull request, and this discussion should happen before a pull request even happens, because we don't want to waste anyone's time in having them implement the full operator without a chance of it being accepted.

> Complicated to follow this rule, since we can do .mergeMap(x => Observable.of(f(x))) instead of map(f). Should we drop map?

No, because as I stated, if performance is not ideal, then it should be considered.

[...]

PR #2524, opened 2017-04-04
docs(repo): add guidelines for operator PRs

staltz:
For issue #2299. We need to educate people to publish operators as libraries, in this way fostering an RxJS ecosystem, instead of making this library thicker and harder to navigate and maintain.

### DIFF
--- a/doc/operator-creation.md
+++ b/doc/operator-creation.md
@@ -8,7 +8,6 @@ For how to develop a custom operator for *this* library, [see below](#advanced).
 
 ## DIY Custom Operators for End Users
 
-
 ### Guidelines
 
 In the most common case, users might like to create an operator to be used only by their app. These can be developed in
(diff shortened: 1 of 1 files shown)


kwonoj:
what about link `Publish your operator as a separate library` part in main README?

doc/operator-creation.md
```diff
@@ -98,9 +97,49 @@ Observable.prototype.mySimpleOperator = mySimpleOperator;
...
+Example implementation:
+
+```js
+function mySimpleOperator(someCallback) {
+  // notice that we return a function here
+  return function mySimpleOperatorImplementation(source) {
+    return Observable.create(subscriber => {
+      var subscription = source.subscribe(value => {
+        try {
+          subscriber.next(someCallback(value));
```
mattpodwysocki: Minor nit here as we want to catch the error from the `someCallback(value)` but not necessarily from the `subscriber.next()` call, so we if could move that out just even for this example.

staltz: Yeah I would have to change the previous example too. I can do that, but just saying

staltz:
@kwonoj the main readme refers to https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md which in turn refers to this md

doc/operator-creation.md
mattpodwysocki: @staltz fair, just wanted to steer people in the right direction to ensure they are only capturing the things under their control.

doc/operator-creation.md
```diff
@@ -98,9 +97,49 @@ Observable.prototype.mySimpleOperator = mySimpleOperator;
...
+
+This can now be used with the `let()` method on the Observable:
+
+```js
+const obs = someObservable.let(mySimpleOperator(x => x + '!'));
+```
+
+## Publish your operator as a separate library
+
+We strongly recommend you to publish your custom operator as an npm package, either as a prototype function to be patched, or as a pure function to be used with `let`. See the sections above. RxJS core already has over 100 operators, and we should not add more operators unless they are absolutely essential and add functionality not possible with existing operators.
```
benlesh: Given future plans to try to move away from prototype patching, I think we should guide them to make a lettable function.

Other than this, and what was pointed out above, I think this is good. I'll try to merge as soon as the changes are in. Sorry for the delayed review.

staltz:
@benlesh I updated it.

staltz:
😃

kwonoj:
I have rebased to latest master for danger checker failure.

staltz:
The CI fails because of markdown-doctest not passing for some older versions of Node.js. Seems harmless to me.

kwonoj:
yes, it's harmless but our branch is protected so can't merge this 😅 I'll probably disable doctest or make this one skipped and check in.

kwonoj:
Sorry for taking so long, I've updated one commit to suppress markdown doctest failing.

PR #2532, opened 2017-04-07
add exports for missing static operators

jayphelps:
Noticed there were a few static operators not exported at `rxjs/observable/*`

This PR currently fails. I can't figure out how to get these new files to be compiled into `dist/cjs/observable/`. I must be missing something cause I can't figure out how some of the other ones like `src/observable.forkJoin.ts` are being compiled but these aren't. Anyone?

If I have time, I'll add similar tests for the instance operators since it's a good idea to verify the public API is actually available.

### DIFF
--- a/spec/exports-spec.ts
+++ b/spec/exports-spec.ts
@@ -1,0 +1,124 @@
+import { expect } from 'chai';
+import { bindCallback } from '../dist/cjs/observable/bindCallback';
+import { bindNodeCallback } from '../dist/cjs/observable/bindNodeCallback';
+import { combineLatest } from '../dist/cjs/observable/combineLatest';
+import { concat } from '../dist/cjs/observable/concat';
+import { defer } from '../dist/cjs/observable/defer';
+import { empty } from '../dist/cjs/observable/empty';
+import { forkJoin } from '../dist/cjs/observable/forkJoin';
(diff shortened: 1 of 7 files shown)


kwonoj:
@jayphelps it's just our tsc compiler picks `Rx.ts` only and if added file's are not references anywhere in import chain, compiler won't pick it up. You may need specify those files along with rx.ts or either reference it somewhere in our codebase make compile works.

jayphelps:
@kwonoj I thought that was the case, but looked and didn't see files like `src/observable/forkJoin.ts` being included there BUT now that I look again, it looks like it's being included by `add/observable/forkJoin` so I think I get the complication and solution.

benlesh:
@jayphelps are you still working on this one?

jayphelps:
@benlesh I had moved on, but I'll try to find time to finish it up--if I can remember 😞

benlesh:
POKE @jayphelps :D

jayphelps:
@benlesh 

dontpokeme

FINE ASSHOLE I UPDATED THIS PR. PS LOVE YOU.

jayphelps:
Hi peeps. Anything left for me or can someone merge plz 😄

kwonoj:
I guess there isn't a specific opinion for changes further.

PR #2538, opened 2017-04-10
fix(find): emit empty observable when predicate find no match

### LINKED ISSUE #2515: find yields single value (undefined) when no match
**RxJS version:**

`5.2.0` / master

**Code to reproduce:**

    Observable.empty().find(_ => false)

**Expected behavior:**

Should not yield any result, as expected (and as per rxjs4 behavior and doc), cf unit test:
https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/find.js#L63

**Actual behavior:**

Yields a single value (undefined) before completion, as per:
https://github.com/ReactiveX/rxjs/blob/master/src/operator/find.ts#L104

**Additional information:**

I believe the cause is that both `find`/`findIndex` operators were implemented by same subscriber, and `findIndex` does emit a value (-1) when not found (also in rxjs4).

kwonoj:
**Description:**
It is my fault when I implemented `find` and `findIndex` operator, make emit `undefined` when there is no match to given predicate, while RxJS v4 states to emit empty observable when it does (https://github.com/ReactiveX/rxjs/issues/2515#issuecomment-291194777). 

This PR changes behavior to align with v4, when there isn't no match sequence will complete without emitting anything.

BREAKING CHANGES: if anyone used `undefined` emitted by unmatched predicate via find operator, it
won't work anymore

- closes #2515

### DIFF
--- a/spec/operators/find-spec.ts
+++ b/spec/operators/find-spec.ts
@@ -43,14 +43,14 @@ describe('Observable.prototype.find', () => {
     expectSubscriptions(source.subscriptions).toBe(subs);
   });
 
-  it('should return undefined if source is empty to match predicate', () => {
+  it('should return empty if source is empty to match predicate', () => {
     const source = cold('|');
     const subs =        '(^!)';
-    const expected =    '(x|)';
+    const expected =    '(|)';
 
     const result = (<any>source).find(truePredicate);
 
-    expectObservable(result).toBe(expected, {x: undefined});
+    expectObservable(result).toBe(expected);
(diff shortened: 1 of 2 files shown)


kwonoj:
I've marked this as `BREAKING CHANGE` at this moment as if anyone was relying on this behavior might be impacted, but there may be different opinions, will leave for review and do not labeled for now.

benlesh:
This is definitely a breaking change. I'm not sure what the proper behavior is here. I think this is right... cc @mattpodwysocki for additional verdict... either way, this is a next major release thing.

Issue #2539, opened 2017-04-10
Subject::lift has incorrect type signature


hearnden:
**RxJS version:** 5.3.0+

**Additional information:**

`lift` is defined on `Observable<T>` as:
```typescript
lift<R>(operator: Operator<T, R>): Observable<R>;
```

`Subject<T>` defines `lift` with the signature:
```typescript
lift<R>(operator: Operator<T, R>): Observable<T>
```
The return type should probably be `Observable<R>` rather than `Observable<T>`

hearnden:
https://github.com/ReactiveX/rxjs/pull/2540

saneyuki:
I confirm this problem affects to the environment using TypeScript _2.4.1_ as stable. (It's no problem with _2.4.0_ as beta version).

And we can workaround this problem with [`--skipLibCheck`](https://www.typescriptlang.org/docs/handbook/compiler-options.html) tsc option. However, its option is not listed by `tsc --help` now. I'm not sure about the status and the future for `--skipLibCheck`.

benlesh:
There are two solutions I can see:

1. Use v 6 alpha
2. We can type the return value as `Obsevable<any>` to solve the problem in a non-breaking way in the short term.

benlesh:
Related: https://github.com/Reactive-Extensions/RxJS/issues/1487#issuecomment-309589773

wulfsberg:
The `any` hack sounds very good to me. I'd love to move to TypeScript 2.4 at first opportunity, given its stronger type inference (as just evidenced :smile:).
And given that the signature *is* actually wrong currently, we're not losing anything with `any`.

foxel:
Will this be fixed for 5.x branch?

floodedcodeboy:
The sister issue from the TypeScript library: https://github.com/Microsoft/TypeScript/issues/16593

DanielRosenwasser:
You can get around this using the `--noStrictGenericChecks` flag in TypeScript 2.4.1.

ScottRFrost:
There's some confusion here with versions impacted.  The current 5.x version of rxjs works fine with TypeScript 2.4.**0**, but not 2.4.**1**.

Meligy:
2.4.0 acts like 2.3 sort of, which would explain why it works.

[...]

PR #2540, opened 2017-04-10
fix(subject): Fixes signature of Subject::lift

hearnden:
https://github.com/ReactiveX/rxjs/issues/2539

### DIFF
--- a/src/Subject.ts
+++ b/src/Subject.ts
@@ -43,7 +43,7 @@ export class Subject<T> extends Observable<T> implements ISubscription {
     return new AnonymousSubject<T>(destination, source);
   }
 
-  lift<R>(operator: Operator<T, R>): Observable<T> {
+  lift<R>(operator: Operator<T, R>): Observable<R> {
     const subject = new AnonymousSubject(this, this);
     subject.operator = <any>operator;
     return <any>subject;


benlesh:
@david-driscoll @kwonoj This *might* be breaking, so let's punt it to the next major. It's not a bug that should effect common uses cases, so we can wait a little bit.

staltz:
LGTM

Igorbek:
With `typescript@next` this signature mismatch is being caught by compiler that prevents using rxjs completely. So I would not say this is a breaking change.

benlesh:
@Igorbek right now 6.0.0-alpha.0 is targeting TS 2.3... I'll be updating that to TS 2.4 RC in short order. That should level this out for you. However, the stable branch will stay at 2.0, as there were breaking changes required to update to 2.1. It's better to not break anyone than make a few people happy.

karptonite:
FYI, TypeScript has released 2.4; this issue is now caught by the compiler, so it won't be possible for anyone using rxjs version 5 to upgrade to the latest TypeScript.

david-driscoll:
Can you check if turning on `skipLibCheck` in your `tsconfig.json` will stop this error from happening?  It should be a stop gap change until we can get 6.0 out.

karptonite:
@david-driscoll Unfortunately, that doesn't help. While that gets rid of the error the compiler finds in Subject directly, it does nothing for all of the errors in my code that use Subject.

saneyuki:
@karptonite `asObservable()` method might help you if your code try to assign `Subject<T>` to the variable/return value which expect `Observable<T>`.

karptonite:
@saneyuki Thanks! I think in this case, I'm better off just staying on ts 2.3.4 until the official fix comes.

benjamincombes:
It would be really great if it could be released as soon as possible, as it affects Angular users too that can't upgrade to TypeScipt 2.4: https://github.com/angular/angular/issues/17800

villanuevadani:
I managed to make it work with RxJS fixed at 5.4.1 and TS at 2.3.4

OliverJAsh:
For all TypeScript users who want to upgrade to 2.4.x but don't want to use an alpha version of TypeScript, and don't want to upgrade to the next major version, is it possible to apply this change to v5 and release a new patch version? Of course this may not be possible if it's considered a breaking change.

Igorbek:
My hack was to automatically patch rxjs typings on `postinstall` script to change `'lift<R>(operator: Operator<T, R>): Observable<T>` to `lift<R>(operator: Operator<T, R>): Observable<R>`.
ps. I still cannot understand why the bugfix considered a breaking change.

Meligy:
I also think this fix should not be considered a breaking change, and would make total sense as a bugfix release version not even minor.

JemiloII:
This should just be a bugfix patch fix for 5.x.x

fletchsod-developer:
fyi the fix already landed on RxJS 5.4.2 (<a href="https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md#542-2017-07-05">changelog</a>), you can either close this issue or leave it open.

PR #2546, opened 2017-04-11
fix(closure): don't throw at top-level scope

alexeagle:
Closure compiler does not allow this. We need this modification to use RxJS in Angular projects.

Addresses https://github.com/angular/tsickle/issues/420

**Description:**

### DIFF
--- a/src/util/root.ts
+++ b/src/util/root.ts
@@ -12,12 +12,18 @@ declare module NodeJS {
  * self: browser in WebWorker
  * global: Node.js/other
  */
-export const root: any = (
-     typeof window == 'object' && window.window === window && window
-  || typeof self == 'object' && self.self === self && self
-  || typeof global == 'object' && global.global === global && global
-);
-
-if (!root) {
-  throw new Error('RxJS could not find any global context (window, self, global)');
+export let root: any;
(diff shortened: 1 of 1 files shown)


alexeagle:
Note, this is the minimal fix. You could instead just avoid throwing an exception: it's unclear from module semantics whether such a throw will abort execution of this one module, or the loading of the whole program. You'd be just as well with a runtime error when attempting to access the `root` property if it doesn't match one of the conditions above. (suggested by @evmar)

alexeagle:
Hold on merging this please, it looks like there is still an issue with closure compiler in the newest release. I'll figure it out tomorrow...

alexeagle:
okay, fixed now @benlesh.
Exciting news: this is the only change needed to make Angular and Closure work, we don't need the ESM distro.

benlesh:
> Note, this is the minimal fix. You could instead just avoid throwing an exception: it's unclear from module semantics whether such a throw will abort execution of this one module, or the loading of the whole program

Yeah, overall this throwing seems odd (I probably added it). I've often wondered if we could get around this whole `root` module to begin with.

This LGTM tho, next patch version.

PR #2547, opened 2017-04-11
build(package): support es6 package generation

### LINKED ISSUE #2335: Angular/Closure users need an ESM distro of rxjs
See https://github.com/angular/angular/issues/8550#issuecomment-277498959
We are close to supporting Angular apps bundled by Closure compiler. It's way smaller than the alternatives and I expect a number of big apps / enterprises will want to evaluate this.

Currently the last blocker is that we need a custom build of RxJS to make this work. We also need one extra bit in there - we use `@angular/tsc-wrapped` (or just `ngc`) in place of `tsc` in order to work around a few closure limitations (no `/index` implicit imports, no `export *`, need complete JSDoc type annotations, etc).

I could send a PR but the bigger issue is the release engineering steps to support a second build, where to publish it (or just replace the current distro as a breaking change), etc. @jayphelps I know you were planning to revisit the topic of an ES6 distro after 5 final shipped. Can we discuss?

kwonoj:
**Description:**
This PR updates build script command to generate es6 module.

- closes #2335


kwonoj:
/cc @alexeagle for visibility.

This should be patch version and should not impact any behavior to existing pkg build. (I hope so :crossed_fingers:)

alexeagle:
What would be the release plan for this? Distribute in https://www.npmjs.com/package/rxjs-es ?

kwonoj:
@alexeagle that's my vote (and `@reactivex/rxjs` as well) but need @benlesh 's input for final decision.

david-driscoll:
Why wouldn't we just look at something like `package.module`?  It would mean that anyone using angular would have to reference `rxjs-es` where most of the docs strictly talk about `rxjs`.  I feel like it would an extra level of confusion for users that's unnecessary.

It may also lead to compiler problems, where I've seen before where we get an error like `Observable<T> is not assignable to Observable<T>`.  This is honestly my biggest fear.

kwonoj:
@david-driscoll I've already shared my opinion to not to use CJS module with module.package definition at https://github.com/ReactiveX/rxjs/issues/2335#issuecomment-285838446 - as we're not only considering consumers of ng (or any webapp has luxury of using bundler) but also have to consider consumers like electron, have to ship pkg in node_modules as-is.

kwonoj:
blocked by https://github.com/ReactiveX/rxjs/pull/2546 , may not needed depends on #2546 goes.

benlesh:
I think we can close this in favor of #2546... since it solves the issue the Angular team had with Closure compiler, and we can continue to punt on ES2015 generation until the community settles on a standard more broadly.

cc @alexeagle

alexeagle:
SGTM. Note that the Angular hello world app gets 0.9% smaller if we get RxJS as ES2015 sources. Let's hope sometime we've fixed all the >0.9% issues and return to that :)

PR #2548, opened 2017-04-11
fix(race): support array of observables other than rest param

kwonoj:
**Description:**
I just noticed `Observable.race([o1, o2])` is not working cause current definition is shaped for rest param args only (so `Observable.race(...[o1, o2])` works. This PR aligns type definition to support those forms to actual operator behavior.

### DIFF
--- a/spec/observables/race-spec.ts
+++ b/spec/observables/race-spec.ts
@@ -35,6 +35,20 @@ describe('Observable.race', () => {
     expectSubscriptions(e2.subscriptions).toBe(e2subs);
   });
 
+  it('should race with array of observable', () => {
+    const e1 =  cold('---a-----b-----c----|');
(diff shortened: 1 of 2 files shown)


kwonoj:
I guess this is patch as it won't breaking existing behavior? (...while I'm currently bad at picking right version for changes though)

kwonoj:
uh..one another thing -

```js
const o1 = Observable.of(1);
const b = Observable.race(o1, o1); //b is inferred to Observable<any> instead of Observable<number>?
```

benlesh:
Might want to change the commit message to be `fix(race): allow TypeScript support for ... ` ... the feature is there, the typings are not.

kwonoj:
Updated commit message.

PR #2550, opened 2017-04-12
fix(find): force unsubscribe when it completes or errors

Podlas29:
**Description:**

Force unsubscribe when resulting Observable completes or errors, even when following operator does not unsubscribe reliably, so that source Observable is not being subscribed unnecessarily.

General description of this type of issue - #2459 
The same fix for other operators - #2501 #2463 #2470

### DIFF
--- a/src/operator/find.ts
+++ b/src/operator/find.ts
@@ -85,6 +85,7 @@ export class FindValueSubscriber<T> extends Subscriber<T> {
 
     destination.next(value);
     destination.complete();
+    this.unsubscribe();
   }
 
   protected _next(value: T): void {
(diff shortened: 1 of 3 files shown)


src/operator/find.ts
```diff
@@ -85,6 +85,7 @@ export class FindValueSubscriber<T> extends Subscriber<T> {
 
     destination.next(value);
     destination.complete();
```
benlesh: This seems odd... calling `complete` on the destination, should end up calling `unsubscribe` on the destination, and subsequently up the chain to here. Although maybe I've got my head in a knot thinking about it. cc @trxcllnt

benlesh:
Thank you @Podlas29!!

Issue #2485, opened 2017-03-21
Observable.throw doesn't throw the error when used inside Observable.fromPromise().catch()


rrhvella:
**RxJS version:** 5.2.0

**Code to reproduce:**

Refer to the following two fiddles:
[RxJS 5.2.0](https://jsfiddle.net/48pxewrs/3/)
[RxJS 5.0.1](https://jsfiddle.net/b3pq14nb/2/)

**Expected behavior:**

In both examples I expect there to be an unhandled error. 

**Actual behavior:**

There is only an unhandled error if version 5.0.1 is used, subsequent versions do not seem to behave this way, including the latest version.

**Additional information:**

I have global error handlers in my javascript code

If I have code inside a promise, inside an observable, then perform some error handling using .catch and throw any errors I don't want to handle using Observable.throw, then the errors I throw again do not reach my global error handlers. It's as if an error wasn't thrown at all.

Actually throwing the error instead of using Observable.throw actually works, however this is not ideal, as I think Observable.throw is supposed to throw the error again. It certainly does when fromPromise isn't used, as show in [this fiddle](https://jsfiddle.net/6p62u57d/2/).

kwonoj:
Closing via #2552.

PR #2552, opened 2017-04-13
fix(ErrorObservable): Added and fixed failing test for Observable.catch

trshafer:
**Description:**

Added a failing test describing unexpected behavior between throw Error and Observable.throw(Error) when the source of an observable is a Promise. In my opinion this behavior is confusing as the catch handler needs to know if the source is from a Promise vs another source. Thanks for taking a look. 

Added fix by forcing `sync.syncErrorThrowable` on `ErrorObservable`. Let me know if this is a good fix and if I should add a unit test for ErrorObservable.

Related #2485.

### DIFF
--- a/spec/operators/catch-spec.ts
+++ b/spec/operators/catch-spec.ts
@@ -1,5 +1,6 @@
 import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
+import * as sinon from 'sinon';
 import {createObservableInputs} from '../helpers/test-helper';
 import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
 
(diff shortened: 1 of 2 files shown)


spec/operators/catch-spec.ts
```diff
@@ -278,4 +279,84 @@ describe('Observable.prototype.catch', () => {
...
+
+  context('fromPromise', () => {
+    type SetTimeout = (callback: (...args: any[]) => void, ms: number, ...args: any[]) => NodeJS.Timer;
+
+    let trueSetTimeout: SetTimeout;
+    let sandbox: sinon.SinonSandbox;
+    let timers: sinon.SinonFakeTimers;
+
+    let source: Rx.Observable<any>;
+    beforeEach(() => {
```
benlesh: These tests are a little too DRY for my liking. Prefer doing as much set up as possible in each test.

Also, I'll be honest, I'm not at all familar with sinon.sandbox. I'll need to research it to see if it's an approach we want to use. I presume it's some virtualized scheduling thing.

trshafer: Hey @benlesh, thanks for taking a look at this PR. I'll update the tests to be more explicit per test and get back to you. I used sinon.sandbox as I found other examples in the current test suite: https://github.com/ReactiveX/rxjs/search?utf8=%E2%9C%93&q=sinon.sandbox&type=

kwonoj: we have few cases of using it, but in general if it can be avoided try not to use it. except some obvious cases (like test scheduler), usually test for Rx can be written without sandbox.

trshafer: I removed the source observable from the setup block and into each test.

spec/operators/catch-spec.ts
```diff
@@ -278,4 +279,83 @@ describe('Observable.prototype.catch', () => {
       done();
     });
   });
+
+  context('fromPromise', () => {
```
benlesh: I know this is a nitpick... but everywhere else I think we use `describe`. I didn't even know `context` is a thing. Is there a reason this is better in this case than `describe`?

trshafer: Happy to change it to `describe`, didn't realize there are no other usages of `context` in the test suite. I personally have used `context` to mean "a different approach to call the same feature" where `describe` is a "different feature". There is no difference in functionality.

```typescript
describe('functionA', () => {
  context('success', () => {...});
  context('error', () => {...});
});
```

benlesh: it's fine I think.

benlesh:
Thanks, @trshafer!

PR #2559, opened 2017-04-16
fix(Subscriber): do not call complete with undefined value param

### LINKED ISSUE #2502: complete() always called with single, undefined argument (differs from prior Rx behavior)
**RxJS version:**
5.1.1

**Code to reproduce:**
```javascript
Rx.Observable
  .of(`hello`)
  .subscribe({
    next() { console.log(arguments); },
    error() { console.log(arguments); },
    complete() { console.log(arguments); }
  });
```

**Expected behavior:**
(When viewed in jsbin's console) should output:
```
[object Arguments] {
  0: 1
}
[object Arguments] { ... }
```

The second output line shows that ```complete()``` did not receive any arguments (as expected).

**Actual behavior:**
Outputs:
```
[object Arguments] {
  0: 1
}
[object Arguments] {
  0: undefined
}
```

The second output line shows that ```complete()``` unexpectedly received an argument.

**Additional information:**
* Previous versions of, for example, Rx 4.0.6, exhibit the expected behavior.
* In rxjs 5.x, the "bonus" argument passed to ```complete()``` is always the value ```undefined```, but that fact that that any arguments are present can conflict with certain use cases, leading to strange, non-obvious behavior.

kwonoj:
**Description:**
This PR makes to call observer's complete function without delivering one param as undefined value all times. While observer is requested to follow its contract interfaces, change seems trivial to avoid this behavior in code as well. 

I think this is non-breaking changes maybe?

- closes #2502

### DIFF
--- a/spec/Subscriber-spec.ts
+++ b/spec/Subscriber-spec.ts
@@ -85,7 +85,7 @@ describe('Subscriber', () => {
   });
 
   it('should not be closed when other subscriber with same observer instance completes', () => {
-    let observer = {
+    const observer = {
       next: function () { /*noop*/ }
     };
 
@@ -96,5 +96,20 @@ describe('Subscriber', () => {
 
     expect(sub1.closed).to.be.false;
     expect(sub2.closed).to.be.true;
+  });
+
+  it('should call complete observer without any arguments', () => {
+    let argument: Array<any> = null;
+
+    const observer = {
+      complete: (...args: Array<any>) => {
+        argument = args;
+      }
+    };
+
+    const sub1 = new Subscriber(observer);
+    sub1.complete();
+
+    expect(argument).to.have.lengthOf(0);
   });
 });
--- a/src/Subscriber.ts
+++ b/src/Subscriber.ts
@@ -234,11 +234,13 @@ class SafeSubscriber<T> extends Subscriber<T> {
     if (!this.isStopped) {
       const { _parentSubscriber } = this;
       if (this._complete) {
+        const wrappedComplete = () => this._complete.call(this._context);
+
         if (!_parentSubscriber.syncErrorThrowable) {
-          this.__tryOrUnsub(this._complete);
+          this.__tryOrUnsub(wrappedComplete);
           this.unsubscribe();
         } else {
-          this.__tryOrSetError(_parentSubscriber, this._complete);
+          this.__tryOrSetError(_parentSubscriber, wrappedComplete);
           this.unsubscribe();
         }
       } else {


src/Subscriber.ts
```diff
@@ -247,18 +247,18 @@ class SafeSubscriber<T> extends Subscriber<T> {
     }
   }
 
-  private __tryOrUnsub(fn: Function, value?: any): void {
+  private __tryOrUnsub(fn: Function, ...value: Array<any>): void {
     try {
-      fn.call(this._context, value);
+      fn.call(this._context, ...value);
```
trxcllnt: @kwonoj couldn't this be `fn.apply(this._context, value)`? And since this is a hot code path, can we avoid allocating the rest args Array with a third boolean argument as a flag?

kwonoj: mean like `__tryOrUnsub(fn: Function, value: Array<any>, passValue: boolean)`?

trxcllnt: yes, exactly

kwonoj: I can, it's eventually same. I just picked rest operator for simplicity of readability.

trxcllnt: @kwonoj yeah, but it transpiles down to an Array allocation + apply call, which is 20-25x slower than the equivalent conditional call [in this jsperf](https://jsperf.com/fn-apply-args-vs-cond-call/1) test.

kwonoj: ah yes, what I meant of `same` is the perspective of consequences only without perf differences. again, I just picked it up for readability at the moment.

src/Subscriber.ts
```diff
@@ -234,11 +234,13 @@ class SafeSubscriber<T> extends Subscriber<T> {
     if (!this.isStopped) {
       const { _parentSubscriber } = this;
       if (this._complete) {
+        const wrappedComplete = () => this._complete.call(this._context);
```
kwonoj: @trxcllnt what about this approach? create one fn closure for `complete` only, and let `__try...` avoid any branching (`if`, or rest operator...). There is overhead to create one additional closure, but only for completion and will not impact `next`.

trxcllnt: @kwonoj that looks great! I like how it removes the branching in the `next` case

spec/Subscriber-spec.ts
```diff
@@ -97,4 +97,15 @@ describe('Subscriber', () => {
...
+
+  it('should call complete observer without any arguments', () => {
+    const observer = {
+      complete: sinon.spy()
+    };
+
+    const sub1 = new Subscriber(observer);
+    sub1.complete();
+
+    expect(observer.complete.firstCall.args).to.empty;
```
benlesh: `.to.be.empty`

benlesh: Also, I don't know enough about sinon to know if this is how we check this. It seems like a more straight forward test just using a plain object and `expect(arguments.length).to.equal(0)` would do.

kwonoj: yeah, probably simple test like this could be achieved in those way. I'll update test.

kwonoj:
test cases are updated as suggested.

PR #2562, opened 2017-04-18
npmignore typings folder

### LINKED ISSUE #2561: Don't publish typings folder
The typings folder with installed TypeScript declarations from the typings definition manager is not .npmignored. The definitions in there are all global, which causes conflicts, because you cannot define the same global symbols twice. It is not possible to use own typings for node without compiler conflicts because rxjs has its own version bundled (`registry:env/node#6.0.0+20161105011511`). Environmental (ambient) typings should not be bundled. All other typings in the typings folder are devDependencies, so I would completely .npmignore the folder.

felixfbecker:
Closes #2561

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -3,3 +3,4 @@ spec
 .git
 yarn.lock
 coverage
+typings/


kwonoj:
Looks good other than commit messages.

PR #2563, opened 2017-04-18
chore(tests): add scripts for running browser tests against global builds

### LINKED ISSUE #2229: Run browser testing via global builds

mpodlasin:
**Description:**

Add scripts for building specs including global builds (Rx.js
and Rx.min.js) instead of cjs builds, so that these packages can
be regression tested in browsers.

Closes #2229

**Caveats:**
Running against global builds crashes few tests, because there are conflicts between global and cjs codebases. For example some tests check if returned values are `instanceof` some internal Observable (say, `GroupedObservable`) which is not exported in global build. Since cjs and global are two different codebases, such checks fail.

### DIFF
--- a/spec/observables/of-spec.ts
+++ b/spec/observables/of-spec.ts
@@ -3,19 +3,21 @@ import * as Rx from '../../dist/cjs/Rx';
 import {ArrayObservable} from '../../dist/cjs/observable/ArrayObservable';
 import {ScalarObservable} from '../../dist/cjs/observable/ScalarObservable';
 import {EmptyObservable} from '../../dist/cjs/observable/EmptyObservable';
+import { Observable } from '../../dist/cjs/Observable';
+import '../../dist/cjs/add/observable/of';
+import '../../dist/cjs/add/operator/concatAll';
 import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
 
 declare const { asDiagram };
 declare const expectObservable: typeof marbleTestingSignature.expectObservable;
 declare const rxTestScheduler: Rx.TestScheduler;
-const Observable = Rx.Observable;
 
 /** @test {of} */
 describe('Observable.of', () => {
   asDiagram('of(1, 2, 3)')('should create a cold observable that emits 1, 2, 3', () => {
-    const e1 = Observable.of(1, 2, 3)
+    const e1 = Rx.Observable.of(1, 2, 3)
       // for the purpose of making a nice diagram, spread out the synchronous emissions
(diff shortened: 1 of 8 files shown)


kwonoj:
few notes:
- travis config running saucelabs should point new browser testing, need check if it passes correctly
- existing browser testing may need to be removed
- `test_browser` script need to be updated

kwonoj:
I'm marking it for reviewing now, though change itself won't impact semver. May need to run couple of saucelabs test, etcs etcs.

spec/observables/of-spec.ts
```diff
@@ -3,19 +3,21 @@ import * as Rx from '../../dist/cjs/Rx';
...
 declare const { asDiagram };
 declare const expectObservable: typeof marbleTestingSignature.expectObservable;
 declare const rxTestScheduler: Rx.TestScheduler;
-const Observable = Rx.Observable;
 
 /** @test {of} */
 describe('Observable.of', () => {
   asDiagram('of(1, 2, 3)')('should create a cold observable that emits 1, 2, 3', () => {
-    const e1 = Observable.of(1, 2, 3)
+    const e1 = Rx.Observable.of(1, 2, 3)
```
benlesh: This seems wrong. We're importing `Observable` above, we should probably use that, no?

benlesh: Same for other similar changes below.

spec/operators/groupBy-spec.ts
```diff
@@ -208,28 +207,6 @@ describe('Observable.prototype.groupBy', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
-  it('should emit GroupObservables', () => {
```
benlesh: We probably shouldn't remove this whole test.

spec/operators/groupBy-spec.ts
```diff
@@ -208,28 +207,6 @@ describe('Observable.prototype.groupBy', () => {
...
-    const e1 = hot('-1--2--^-a-b----|', values);
-    const e1subs =        '^        !';
-    const expected =      '--g------|';
-    const expectedValues = { g: 'foo' };
-
-    const source = e1
-      .groupBy((val: string) => val.toLowerCase().trim())
-      .do((group: any) => {
-        expect(group.key).to.equal('foo');
-        expect(group instanceof GroupedObservable).to.be.true;
```
benlesh: an `instanceof Observable` test here is fine, too.

benlesh:
I think this is now really outdated, since testing strategies have changed since having to add rxjs-compat and others. We can probably just close it. But we do need to re-explore browser testing in the near future.

PR #2569, opened 2017-04-19
docs(css): enable momentum scrolling for iOS

nelsonjchen:
**Description:**

https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-overflow-scrolling

All other mobile and desktop platforms with built-in momentum scrolling
enable it by default. iOS requires an explicit CSS rule.


kwonoj:
I'm checking this in as this won't impact code packages.

PR #2576, opened 2017-04-24
fix(tests): add missing babel-polyfill to package.json

### LINKED ISSUE #2261: Fix incorrect setup causes test runner stops on browser
**RxJS version:**
master

**Code to reproduce:**
Run browser test

**Additional information:**
It seems some of test cases are having incorrect fixture setup. I can't even tell what is incorrect at this moment - it could be test case was composed incorrectly, or browser test setup might wrong. It creates intermittent test failure on browsers such as causing

- test completes too early without executing all suite
- some test causes test runner hangs without timeout, infinitely stays there

below screenshot illustrates first cases I could see, only 1/10 of tests are executed but marked as 100%.

image

Note this is very intermittent issue and sometimes it occurs quite easily but sometimes never occurs. to improve stability of tests, this need to be addressed and fixed. 

Q: What's the difference between https://github.com/ReactiveX/rxjs/issues/1660?
: #1660 is more about issue of connection to browser testing itself, which cannot connect or either disconnect quite frequently prevents test run reliably. I have tested `browserStack` and it indeed works more reliably, but due to issue in our test cases it now 'reliably fails' on several browsers.

Podlas29:
Add missing babel-polyfill to package.json, so that tests
run in the browsers that do not support ES6 features run
without issues.

Closes #2261

Dependency is used here: https://github.com/ReactiveX/rxjs/blob/master/spec/support/mocha-browser-runner.html#L11

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,6 +146,7 @@
   },
   "homepage": "https://github.com/ReactiveX/RxJS",
   "devDependencies": {
+    "babel-polyfill": "^6.23.0",
     "benchmark": "^2.1.0",
     "benchpress": "2.0.0-beta.1",
     "chai": "^3.5.0",


kwonoj:
I'll check this in right away once CI returns.

kwonoj:
Thanks for finding this out, this was long pain 👏

PR #2580, opened 2017-04-27
fix(AsyncAction): Make AsyncAction update its timeout if rescheduled before its executed

### LINKED ISSUE #2579: The timeout() operator throws error prematurely
**RxJS version:** 5.3.0

It seems like after the most [recent changes in 5.3.0](https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md#bug-fixes) the `timeout()` operator works differently. But I don't know whether this is intentional or not since it's not listed under either "Features" nor "Breaking Changes" in the change log.

**Code to reproduce:**

```
Observable
  .range(0, 10)
  .concatMap(val => Observable.of(val).delay(val >= 3 ? 5000 : 1000))
  .timeout(2000)
  .subscribe(
    val => console.log('next', val),
    err => console.log('error', err)
  );
```

First three items are delayed by 1 second while the fourth item is delayed by 5 seconds.

**Expected behavior:**

```
next 0
next 1
next 2
error { TimeoutError: Timeout has occurred
    at TimeoutError.Error (native)
...
  message: 'Timeout has occurred' }
```

**Actual behavior:**

```
next 0
error { TimeoutError: Timeout has occurred
    at TimeoutError.Error (native)
...
  message: 'Timeout has occurred' }
```

**Additional information:**

Only `5.3.0` is affected. I checked both `5.2.0` and `4.1.0` and in both cases it work as shown in Expected behavior. Also it looks like this situation is already tested: https://github.com/ReactiveX/rxjs/blob/master/spec/operators/timeout-spec.ts#L89

Demos:

RxJS 5.3.0: https://jsbin.com/wemuzuc/5/edit?html,js,console
RxJS 5.2.0: https://jsbin.com/nedagox/4/edit?html,js,console
RxJS 4.1.0: https://jsbin.com/ganofon/2/edit?html,js,console

trxcllnt:
Rescheduling an async action with the same due time before its executed should cause it to cancel
its current interval and schedule a new one.

Fixes #2579

### DIFF
--- a/src/scheduler/AsyncAction.ts
+++ b/src/scheduler/AsyncAction.ts
@@ -74,11 +74,11 @@ export class AsyncAction<T> extends Action<T> {
 
   protected recycleAsyncId(scheduler: AsyncScheduler, id: any, delay: number = 0): any {
     // If this action is rescheduled with the same delay time, don't clear the interval id.
-    if (delay !== null && this.delay === delay) {
+    if (delay !== null && this.delay === delay && this.pending === false) {
       return id;
     }
     // Otherwise, if the action's delay time is different from the current delay,
-    // clear the interval id
+    // or the action has been rescheduled before it's executed, clear the interval id
     return root.clearInterval(id) && undefined || undefined;
   }
 


kwonoj:
I guess this can be patch release?

trxcllnt:
@kwonoj I believe so, as it's a backwards-compatible bugfix

benlesh:
Looks good @trxcllnt ... Also points out how tricky this implementation is. haha.

:shipit:

Issue #2582, opened 2017-04-28
Bump up typescript compiler? 


kwonoj:
We're currently staying bit behind version of latest TypeScript compiler for some of compatibility reasons. now ng4.1 (http://angularjs.blogspot.com/2017/04/angular-410-now-available.html) is fully compatible with 2.3, our biggest concern seems resolve and maybe we can bump up compiler to use some of latest language features?

david-driscoll:
I say lets do it.  We can bump the compiler and then turn on `strict: true` as a separate PR.

cc @benlesh

kwonoj:
afaik only currently blocking issue is https://github.com/ReactiveX/rxjs/pull/2217 .

saneyuki:
From [rxjs-core-notes/2017-01/january-30.md](https://github.com/ReactiveX/rxjs-core-notes/blob/master/2017-01/january-30.md):

> Punt and work with ng2 team towards the same solution, since they have the problem too.

By this decision, we need to sort with angular.

From their code, I seem Angular have some pattern to construct an error:

1. https://github.com/angular/angular/blob/master/packages/router/src/events.ts#L71-L84
  - I don't know about angular framework well, however, I think we cannot follow this approach because it would be a compat problem with IE11. 
2. https://github.com/angular/angular/blob/master/packages/router/src/events.ts#L71-L84
    - This pattern don't drive from `Error` object.
3. https://github.com/angular/angular/blob/bebedfed24d6fbfa492e97f071e1d1b41e411280/packages/core/src/di/reflective_errors.ts#L39-L58
  - This pattern export a interface (`InjectionError`) and export a factory function which returns a new `InjectionError`.

 #2217's approach is different from them. 

I feel _pattern-3_ would reasonable way for us.

It does not have a complex exporting symbols like the approach I proposed in #2217. It make sense. The drawback point of _pattern-3_ is that we cannot create the error with `new` syntax. However, I don't feel it's serious problem because our user would not create rxjs' internal error except for testing.

---------------------------

@kwonoj @david-driscoll @benlesh 

How about this?

david-driscoll:
Option 3 seems reasonable to me.

saneyuki:
> The drawback point of pattern-3 is that we cannot create the error with new syntax. However, I don't feel it's serious problem because our user would not create rxjs' internal error except for testing.

Ugh, I missed more drawback points.

[...]

PR #2585, opened 2017-05-02
fix(AjaxObservable): fix AjaxObservable's XML responseType for TS 2.3

voithos:
**Description:**
This change fixes the type of the `responseType` property for TS 2.3, which introduced stricter typings, and changes the AjaxObservable code to look for the `document` responseType instead of `xml` (which isn't a valid responseType). See the MDN article on `.responseXML` usage:
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseXML

For reference, here is the relevant section of the TS 2.3 typings: https://github.com/Microsoft/TypeScript/blob/master/lib/lib.d.ts#L17605

Note that this doesn't tackle any of the other tasks necessary to upgrade to TS 2.3, which seem to be being tracked at #2582.

### DIFF
--- a/src/observable/dom/AjaxObservable.ts
+++ b/src/observable/dom/AjaxObservable.ts
@@ -20,7 +20,7 @@ export interface AjaxRequest {
   withCredentials?: boolean;
   createXHR?: () => XMLHttpRequest;
   progressSubscriber?: Subscriber<any>;
-  responseType?: string;
+  responseType?: typeof XMLHttpRequest.prototype.responseType;
 }
 
 function getCORSRequest(this: AjaxRequest): XMLHttpRequest {
@@ -401,8 +401,8 @@ export class AjaxResponse {
   /** @type {string} The raw responseText */
   responseText: string;
 
-  /** @type {string} The responseType (e.g. 'json', 'arraybuffer', or 'xml') */
-  responseType: string;
+  /** @type {string} The responseType (e.g. 'json', 'arraybuffer', or 'document') */
+  responseType: typeof XMLHttpRequest.prototype.responseType;
 
   constructor(public originalEvent: Event, public xhr: XMLHttpRequest, public request: AjaxRequest) {
     this.status = xhr.status;
@@ -417,7 +417,10 @@ export class AjaxResponse {
           this.response = JSON.parse(xhr.responseText || 'null');
         }
         break;
-      case 'xml':
+      // TODO: The 'xml' case exists for backwards compatibility, but is not a
+      // valid responseType and should be removed.
+      case 'xml' as any:
+      case 'document':
         this.response = xhr.responseXML;
         break;
       case 'text':


kwonoj:
I'm bit confused, while commit / PR says this is fix for TS 2.3, PR seems changing code behavior as well? (https://github.com/ReactiveX/rxjs/pull/2585/files#diff-f80ecdc9803b338b8f47819f25b3a721R420) in that case, shouldn't this falls into code fixes?

voithos:
Yes, sorry for not being clear - this really is just a code fix (+ a forwards-compatible type fix) that was exposed via the TS 2.3 changes. I'll change the commit/PR description to reflect that.

kwonoj:
looks ok to me, but not sure about releases.

src/observable/dom/AjaxObservable.ts
```diff
@@ -417,7 +417,7 @@ export class AjaxResponse {
           this.response = JSON.parse(xhr.responseText || 'null');
         }
         break;
-      case 'xml':
```
benlesh: why are we removing `case 'xml'`? This is a breaking change. If this is truly necessary: 1. We'll need demonstratable proof of the need to remove this, and 2. We'll need to wait until the next MAJOR release (RxJS 6) in order to get this into production (because it's a breaking change)

src/observable/dom/AjaxObservable.ts
```diff
@@ -417,7 +417,7 @@ export class AjaxResponse {
...
+      case 'document':
```
benlesh: Do we have test cases for this? Also, why `document` and not `xml`?

voithos: Because it seems that `xml` is not a valid responseType, both per-spec as well as per the TS 2.3 XHR typings (see the links I gave in the PR description). I don't think this code had test cases, unfortunately.

src/observable/dom/AjaxObservable.ts
voithos: Indeed, this may be a breaking change - but there are 2 issues here:
- `xml` is not a valid responseType per the spec (see the links I gave in the PR description). So it seems that any server that returned `xml` as the responseType would not be compliant to the spec.
- Thus, `xml` is not accepted as a valid responseType by the TS 2.3 compiler (because the typings were made stricter, to disallow invalid responseTypes).

One approach to preserve compatibility may be to handle both the `xml` and `document` cases explicitly (via a type cast, etc).

src/observable/dom/AjaxObservable.ts
benlesh: Okay... can we leave `'xml'` in there for now, with a // TODO: to remove it? That will keep compatability and make this a patch rather than a major revision change. We can remove `xml` in the next major version.

voithos: Sure, added the `'xml'` back in with a TODO.

voithos:
Friendly ping?

voithos:
@benlesh Let me know if you think anything else should be changed.

benlesh:
@voithos. Sorry, I've just been busy. This is definitely something we're going to merge. Just need time to align it with the TS update

voithos:
@benlesh Sure, no worries at all. Just wanted to confirm that this looks good as-is, or if more changes would be desired (it's no longer a breaking change, right?).

voithos:
So, it's been a while, and given that RxJS and Angular (and TS) have all advanced quite a bit, I'm guessing this PR is no longer relevant? Should we close this?

benlesh:
@voithos yes, thank you for closing this. This is resolved and what we have for version 6 builds against the latest version of TypeScript.

PR #2589, opened 2017-05-03
fix(delayWhen): correctly handle synchronous duration observable

### LINKED ISSUE #2587: Bug: delayWhen operator returns an Observable that does not complete.
**RxJS version:**
5.3.0

**Code to reproduce:**
```
Rx.Observable.of('foo')
  .delayWhen(function() {
    return Rx.Observable.of('bar');
  })
  .subscribe({
    next: function(x) {
      console.log('next');
    },
    complete: function() {
      console.log('complete');
    }
  });
```

**Expected behavior:**
Output
`'next'`
`'complete'`

**Actual behavior:**
Output
`'next'`

**Additional information:**
If you substitute `delayWhen(...)` for say `delay(1000)` in the above, it will emit complete as expected.

[Here is a jsFiddle](https://jsfiddle.net/b8sd30k3/) using [https://unpkg.com/rxjs/bundles/Rx.min.js](https://unpkg.com/rxjs/bundles/Rx.min.js) showcasing the issue.

kwonoj:
**Description:**
This PR fixes outersubscriber behavior of delayWhen operator, to complete it correctly when duration selector returns synchronous observable. Similar to https://github.com/ReactiveX/rxjs/pull/1490 in crux.

- closes #2587

### DIFF
--- a/src/operator/delayWhen.ts
+++ b/src/operator/delayWhen.ts
@@ -56,7 +56,7 @@ export function delayWhen<T>(this: Observable<T>, delayDurationSelector: (value:
                              subscriptionDelay?: Observable<any>): Observable<T> {
   if (subscriptionDelay) {
     return new SubscriptionDelayObservable(this, subscriptionDelay)
-            .lift(new DelayWhenOperator(delayDurationSelector));
+      .lift(new DelayWhenOperator(delayDurationSelector));
   }
   return this.lift(new DelayWhenOperator(delayDurationSelector));
 }
@@ -112,7 +112,7 @@ class DelayWhenSubscriber<T, R> extends OuterSubscriber<T, R> {
         this.tryDelay(delayNotifier, value);
       }
     } catch (err) {
-        this.destination.error(err);
+      this.destination.error(err);
     }
   }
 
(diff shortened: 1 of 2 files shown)


kwonoj:
This shouldn't cause any breaking changes.

src/operator/delayWhen.ts
```diff
@@ -56,7 +56,7 @@ export function delayWhen<T>(this: Observable<T>, delayDurationSelector: (value:
                              subscriptionDelay?: Observable<any>): Observable<T> {
   if (subscriptionDelay) {
     return new SubscriptionDelayObservable(this, subscriptionDelay)
-            .lift(new DelayWhenOperator(delayDurationSelector));
```
benlesh: In the future, these sorts of formatting corrections should be in a separate commit for easier review.

src/operator/delayWhen.ts
```diff
@@ -112,7 +112,7 @@ class DelayWhenSubscriber<T, R> extends OuterSubscriber<T, R> {
         this.tryDelay(delayNotifier, value);
       }
     } catch (err) {
-        this.destination.error(err);
```
benlesh: same as above.

PR #2592, opened 2017-05-05
refactor(root): make the root module more concise

benlesh:
Borrowed an implementation from Angular that's a bit more concise than what we have.


benlesh:
Oops... working off the wrong branch. :\

Closing, and trying again.

PR #2593, opened 2017-05-05
refactor(root): a slightly more concise definition borrowed from Angular

benlesh:
**Description:**

### DIFF
--- a/src/util/root.ts
+++ b/src/util/root.ts
@@ -1,29 +1,22 @@
-declare let global: NodeJS.Global;
-
-declare module NodeJS {
-  interface Global {
-    window: any;
-    global: any;
-  }
-}
+declare let global: any;
 
 /**
- * window: browser in DOM main thread
- * self: browser in WebWorker
- * global: Node.js/other
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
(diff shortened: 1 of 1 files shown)


benlesh:
> so we're not throwing anymore if root is not resolvable?

I'm not sure there is a situation where root isn't resolvable, honestly. That was just some defensive coding. I'm indifferent, I'll leave it to the community.

kwonoj:
I wouldn't expect this occurs in usual cases, but there were few exceptional cases patching globals around could be caught by thrown exception. So for me it's good to have, it allowed us to identify some unexpected behavior comes from it.

src/util/root.ts
```diff
@@ -1,29 +1,26 @@
...
+// the global "global" var for now.
+
+const __window = typeof window !== 'undefined' && window;
+const __self = typeof self !== 'undefined' && typeof WorkerGlobalScope !== 'undefined' &&
+    self instanceof WorkerGlobalScope && self;
+const __global = typeof global !== 'undefined' && global;
+const _root: any = __window || __global || __self;
+
+if (!_root) {
+  throw new Error('No window, global or root reference found');
```
kwonoj: I guess throw should be wrapped in IIFE as previous for closure compiler compatibility?

benlesh: Yeah.. changed it to match what used to be there exactly.

PR #2595, opened 2017-05-08
fix(audit): mirror source if durations are Observable.empty()

GrahamDennis:
**Description:**
The `audit` operator was suppressing all events if the duration was `Observable.empty()` or any observable that was already complete.

There was already a similar test for this, but it relied upon `cold('|')`. It might be worth considering either duplicating tests involving `cold('|')` to use `Observable.empty()` or special-casing it to return `Observable.empty()`

### DIFF
--- a/src/operator/audit.ts
+++ b/src/operator/audit.ts
@@ -86,6 +86,9 @@ class AuditSubscriber<T, R> extends OuterSubscriber<T, R> {
         this.destination.error(errorObject.e);
       } else {
         this.add(this.throttled = subscribeToResult(this, duration));
+        if (this.throttled.closed) {
+          this.clearThrottle();
+        }
       }
     }
   }
--- a/src/operator/audit.ts
+++ b/src/operator/audit.ts
@@ -85,9 +85,11 @@ class AuditSubscriber<T, R> extends OuterSubscriber<T, R> {
       if (duration === errorObject) {
         this.destination.error(errorObject.e);
       } else {
-        this.add(this.throttled = subscribeToResult(this, duration));
-        if (this.throttled.closed) {
+        const innerSubscription = subscribeToResult(this, duration);
+        if (innerSubscription.closed) {
           this.clearThrottle();
+        } else {
+          this.add(this.throttled = innerSubscription);
         }
       }
     }
--- a/spec/operators/audit-spec.ts
+++ b/spec/operators/audit-spec.ts
@@ -127,6 +127,18 @@ describe('Observable.prototype.audit', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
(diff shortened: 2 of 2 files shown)


src/operator/audit.ts
```diff
@@ -86,6 +86,9 @@ class AuditSubscriber<T, R> extends OuterSubscriber<T, R> {
         this.destination.error(errorObject.e);
       } else {
         this.add(this.throttled = subscribeToResult(this, duration));
```
kwonoj: shouldn't synchronized subscription to be added into subscription as well?

src/operator/audit.ts
```diff
@@ -86,6 +86,9 @@ class AuditSubscriber<T, R> extends OuterSubscriber<T, R> {
...
+        if (this.throttled.closed) {
```
benlesh: I'm curious about this change. Wouldn't this make it so a completed `Subject` would cause this behavior? That seems incorrect.

Is this really a subscription ordering thing? What do you think @trxcllnt?

kwonoj: I have assumed this change's similar to this -> https://github.com/ReactiveX/rxjs/pull/1490, where there are synchronous observables are supplied `subscribeToResult` completes immediately instead of scheduled.

GrahamDennis: When I ran this code under a debugger, that's exactly what happened: inside `subscribeToResult`, the observable completed, so by the time it is returned, it is already completed, so we shouldn't throttle subsequent results anymore.

I have two questions:
1. How should we solve this problem?
2. How do we avoid this problem elsewhere as the cold('|') observable used in tests doesn't complete immediately.

src/operator/audit.ts
GrahamDennis: I don't understand what you mean. Can you explain?

kwonoj: Refer this one - https://github.com/ReactiveX/rxjs/pull/1490/files#diff-4c9df56dc210f37759b96afb96e5fdd0R135

if `subscribeToResult` is synchronously subscribed, returned subscription is immediately unsubscribed already, so there's no meaning to add subscription for managing it (`this.add`). So in that case, check returned subscription and add only if it's asynchronous would be better.

GrahamDennis: I've updated the PR to address this.

GrahamDennis:
Who can merge this?

kwonoj:
I can, but checking with release cadences to make right version. Guess this is patch? /cc @benlesh

deadbeef84:
There's a problem with this commit, when `subscribeToResult` returns null it will crash. It should check if innerSubscription is null or closed.
```js
if (!innerSubscription || innerSubscription.closed) {
```

PR #2606, opened 2017-05-17
Update to TypeScript 2.1, Update Errors, Breaking Changes!

### LINKED ISSUE #2605: Stop duplicating TypeScript class helpers in every CJS file output
Currently, we're outputting the class helpers in ever CJS file we output, that means that anyone bundling them will end up having N copies of `__extend` methods in their bundle, where N in the number of our modules they import that use classes.

This is a particular problem for the Angular team.

cc/ @IgorMinar

benlesh:
### Breaking Changes!!

- Updates to TS 2.1
- Eliminates duplicate class helpers in every CJS output file 
- adds tslib
- No longer exporting any `Error` subclasses, rather there are utility functions to create Errors, and to test if an error is a specific type.
- Requires that `Symbol` is properly shimmed

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "build_test": "shx rm -rf ./dist/ && npm-run-all build_cjs clean_spec build_spec test_mocha",
     "build_cover": "shx rm -rf ./dist/ && npm-run-all build_cjs build_spec cover",
     "build_docs": "npm-run-all build_global build_es6_for_docs build_cjs clean_spec build_spec tests2png decision_tree_widget && esdoc -c esdoc.json && npm-run-all clean_dist_es6",
-    "build_spec": "tsc --project ./spec --pretty",
+    "build_spec": "tsc --project ./spec --lib es5,es2015.iterable,es2015.collection,es2015.promise,dom --pretty",
     "build_spec_browser": "webpack --config spec/support/webpack.mocha.config.js",
     "check_circular_dependencies": "madge ./dist/cjs --circular",
     "clean_spec": "shx rm -rf spec-js",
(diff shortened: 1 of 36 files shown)


benlesh:
I'm honestly not entirely sure I did this properly, but the tests are running okay, and I don't see the duplicate `__extends` methods in the build output.

cc @IgorMinar @kwonoj

kwonoj:
In my blur memory our global build did manually import tslib something similar, not sure if it still exists and need to be updated?

benlesh:
@kwonoj  are you talking about this? https://github.com/ReactiveX/rxjs/blob/cf88a208883f99f4fd0f368dcbaa0d0b646877aa/tools/make-umd-bundle.js#L20

I honestly don't remember what that's doing or why. :\ If no one remembers, I guess I'll have to dig in.

kwonoj:
@benlesh yes, these 3 are history of those .

https://github.com/ReactiveX/rxjs/pull/2064
https://github.com/ReactiveX/rxjs/pull/2093
https://github.com/ReactiveX/rxjs/pull/2121

src/Rx.ts
```diff
@@ -1,4 +1,6 @@
 /* tslint:disable:no-unused-variable */
+import 'tslib';
```
IgorMinar: is this just so that we bring in the helpers for the umd file?

tsconfig.json
```diff
@@ -1,5 +1,7 @@
 {
   "compilerOptions": {
+    "importHelpers": true,
+    "noEmitHelpers": true,
```
IgorMinar: I don't think  that you need `noEmitHelpers`, `importHelpers` should be sufficient.

package.json
```diff
@@ -75,10 +75,10 @@
...
-    "cover": "shx rm -rf dist/cjs && tsc src/Rx.ts src/add/observable/of.ts -m commonjs --lib es5,es2015.iterable,es2015.collection,es2015.promise,dom --outDir dist/cjs --sourceMap --target ES5 -d && nyc --reporter=lcov --reporter=html --exclude=spec/support/**/* --exclude=spec-js/**/* --exclude=node_modules mocha --opts spec/support/default.opts spec-js",
+    "compile_dist_cjs": "tsc ./dist/cjs/src/Rx.ts ./dist/cjs/src/add/observable/of.ts -m commonjs --lib es5,es2015.iterable,es2015.collection,es2015.promise,dom --noEmitHelpers --sourceMap --outDir ./dist/cjs --target ES5 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --noImplicitThis --suppressImplicitAnyIndexErrors --moduleResolution node",
+    "compile_module_es6": "tsc ./dist/es6/src/Rx.ts ./dist/es6/src/add/observable/of.ts -m es2015  --noEmitHelpers  --sourceMap --outDir ./dist/es6 --target ES5 -d --diagnostics --pretty --noImplicitAny --noImplicitReturns --noImplicitThis --suppressImplicitAnyIndexErrors --moduleResolution node --noEmitHelpers --lib es5,es2015.iterable,es2015.collection,es2015.promise,dom ",
```
IgorMinar: is this right? shouldn't this be `importHelpers`?

tsconfig.json
```diff
@@ -1,5 +1,7 @@
 {
   "compilerOptions": {
+    "importHelpers": true,
```
imcotton: `importHelpers` requires TypeScript [v2.1](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html#support-for-external-helpers-library-tslib)

package.json
benlesh: :man_shrugging: 

What I found online said `noEmitHelpers`. I'll try it the other way around.

tsconfig.json
benlesh: Noted, thanks... I'm going to attempt to update TS... 

Sadly, that makes this a breaking change. :(

staltz:
Don't fear the major number ;)

benlesh:
FWIW: I move that this can be the first RxJS 6 alpha or beta once merged, and we can start looking at merging some of the other changes for the next major sitting in our queue. As long as the breaking changes aren't too fast.

benlesh:
Closing in favor of #2614

PR #2610, opened 2017-05-24
chore(typescript): Cleanup usage of typings package.

alexeagle:
TypeScript now distributes typings via npm.

Note, `@types` is not semver, because patch releases may introduce syntax that requires a TypeScript upgrade. So the `package.json` pins these to versions (just as the `typings.json` did)


kwonoj:
related to https://github.com/ReactiveX/rxjs/issues/2437.

kwonoj:
danger checker seems off, don't mind it - I'll look into separately.

alexeagle:
From the other issue you linked, it's not clear to me if this PR causes some problem - do we need testing of user-land type-checking?

kwonoj:
> it's not clear to me if this PR causes some problem

: that's reason we were hesitating migrate into @types package. Surprisingly we had regressions by shipping pkg using @types pkg from typings pkg previously, and haven't introduced automatic way to verify those in our repo. It's not about testing 3rd party typing's validity, it was using those pkg TS compiler generated unexpected results. I believe that's no longer issue though.

alexeagle:
Sounds like you would benefit from the same typings test setup we use for
angular.

benlesh:
I've merged this into #2614. Closing in favor of that.

PR #2611, opened 2017-05-24
chore(travis): pin danger version

kwonoj:
**Description:**
Seems latest version caused regressions make checker failure, simply pin down to known good.


kwonoj:
It's not impacting codebase and will unblock other PR's checker failure - checking it in now.

PR #2614, opened 2017-05-25
Update TS: Alternative Error Fix

benlesh:
This is the same as #2606, but updates the custom error classes to use `setPrototypeOf`

### Pros:

- `instanceof` still works
- fewer breaking changes
- call stacks and other goodness preserved for Chrome DevTools

### Cons:

- IE10 users will need an `Object.setPrototypeOf` polyfill.
- `setPrototypeOf` creates a "slow object" in V8... deoptimized as a hashtable, to my understanding.

### Other things:

The `Object.setPrototypeOf` polyfill is trivial:

```ts
Object.setPrototypeOf || Object.setPrototypeOf = function(a, b) { a.__proto__ = b; return a; }
```

cc @trxcllnt


src/observable/dom/AjaxObservable.ts
```diff
@@ -5,6 +5,7 @@ import { Observable } from '../../Observable';
 import { Subscriber } from '../../Subscriber';
 import { TeardownLogic } from '../../Subscription';
 import { MapOperator } from '../../operator/map';
+import 'tslib';
```
Brooooooklyn: don't need this

src/Rx.ts
```diff
@@ -1,4 +1,6 @@
 /* tslint:disable:no-unused-variable */
+import 'tslib';
```
Brooooooklyn: don't need this either

Widdershin:
The failure from `markdown-doctest` appears to be a module issue.

It's failing when trying to `import Rx from 'rxjs/Rx'`.

This is the example from the README that's failing:

```js
import Rx from 'rxjs/Rx';

Rx.Observable.of(1,2,3)
```

After being compiled by babel this is how it looks:
```js
var _Rx = require('rxjs/Rx');

var _Rx2 = _interopRequireDefault(_Rx);

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

_Rx2.default.Observable.of(1, 2, 3);
```
This started breaking with TypeScript 2.3 because now `require('rxjs/Rx').__esModule` is `true` so `_interopRequireDefault` goes down a different branch in the ternary.

This line now appears at the top of the compiled output:

```js
Object.defineProperty(exports, "__esModule", { value: true });
```

Arguably this is reasonable behaviour, as `import Rx from 'rxjs/rx';` is importing a default, and the module in question has no exported default.

I'm unsure why `__esModule` is now being set to true, seems like this must have changed in TypeScript 2.3 or is configured somewhere I can't see.

One possible fix is to change `import Rx from 'rxjs/Rx'` to `import * as Rx from 'rxjs/Rx'`.

Another would be to export a default, but that seems like it would be a lot of duplication.

I think this failure is indicative of a problem that could impact real users. I know TypeScript is strict about importing using that syntax from modules with no default. I think in the past `babel` allows you to use that syntax for `commonjs` modules, but maybe it's always been based on `__esModule`.

benlesh:
Thanks @Widdershin ... I think we can actually just make that change. It's only in two spots I believe.

Issue #1905, opened 2016-08-25
publish<T>(selector?: function(Observable<T>): Observable<R> ): Observable<R> should be allow to get type changed.


Dorus:
The method signature for publish(selector) should be 

```
function publish<T>(selector?: (source: Observable<T>) => Observable<R>): Observable<R> | ConnectableObservable<R>
```

like `let` to allow type changes in the selector. Currently it is using 

```
function publish<T>(selector?: (source: Observable<T>) => Observable<T>): Observable<T> | ConnectableObservable<T>
```

https://github.com/ReactiveX/rxjs/blob/master/src/operator/publish.ts#L19

staltz:
Seems legit.

daixtrose:
I propose to provide 2 different methods here. One with and one without argument. In a strongly typed world it is surprising to have the return type depend on the availability of an argument.

- `publish()`
- `mapPublish(selector ...)`

Dorus:
I like that idea. Could this pattern be mimicked for other operators that return a ConnectableObservable?

`publishReplay` could have an `mapReplay` similar to this [`Replay`](https://msdn.microsoft.com/en-us/library/hh229288(v=vs.103).aspx) operator from Rx.Net

Same with:
`publishBehavior` => `mapBehavior`
`publishLast()` => `mapLast` (?)

There is also `multicast()` that seems to do something with `ConnectableObservable` but i do not quite understand it.

david-driscoll:
Publish should be changed, the main signature to update (until we move to 2.0 features) is [here](https://github.com/ReactiveX/rxjs/blob/master/src/operator/multicast.ts#L44-L47).

We probably also need to make changes to all the multicast observables, as [`multicast`](https://github.com/ReactiveX/rxjs/blob/master/src/operator/multicast.ts#L44-L47) has the same issue.

kwonoj:
I'm not sure where we at for this issue, probably still need to update type definitions? /cc @david-driscoll

Dorus:
Small update to the op: `ConnectableObservable<T>` should have been `ConnectableObservable<R>` too. I've changed that up here.

benlesh:
The typings and definitions have changed so much around this, that I'm pretty sure this thread is no longer relevant. Feel free to reopen if I'm in error.

Thanks!

PR #2616, opened 2017-05-25
Fix TS definition of publish & multicast

hermanbanken:
**Description:**
Change TS definition of publish & multicast to support selectors changing the type.

#1905

### DIFF
--- a/src/operator/multicast.ts
+++ b/src/operator/multicast.ts
@@ -6,7 +6,7 @@ import { ConnectableObservable, connectableObservableDescriptor } from '../obser
 
 /* tslint:disable:max-line-length */
 export function multicast<T>(this: Observable<T>, subjectOrSubjectFactory: factoryOrValue<Subject<T>>): ConnectableObservable<T>;
-export function multicast<T>(SubjectFactory: (this: Observable<T>) => Subject<T>, selector?: selector<T>): Observable<T>;
+export function multicast<T, R>(SubjectFactory: (this: Observable<T>) => Subject<T>, selector?: selector<T, R>): Observable<R>;
 /* tslint:enable:max-line-length */
 
 /**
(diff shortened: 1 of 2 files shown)


hermanbanken:
Currently working around this...

```typescript
declare module "rxjs/operator/publish" {
	export type selectors<T,R> = (source: Observable<T>) => Observable<R>;
	export function publish<T,R>(this: Observable<T>, selector: selectors<T,R>): Observable<R>;
}
```

hermanbanken:
Checks are failing due to some conflicts in the Travis <-> Github API communication after I fixed the commit message and force pushed... /cc @kwonoj

kwonoj:
I'm feeling this might be legit case of `<T, R=T>` with #2614 maybe.

kwonoj:
And second thought, isn't this PR causes breaking to existing consumers uses `publish<T>`?

hermanbanken:
@kwonoj the following two types exist:

```typescript
export function publish<T>(this: Observable<T>): ConnectableObservable<T>;
export function publish<T, R>(this: Observable<T>, selector: selector<T, R>): Observable<R>;
```
So:

- users of `publish<T>()` should not be affected. This will be the largest use case (e.g. `publish` & `connect`).
- users of `publish<T>(selector)` with a selector that does not change the type of the output would not be affected if they did not explicitly write `.publish<T>(` in their code. Typescript would automatically infer the new R type from the selector. If they did explicitly write `<T>` they will need to change that to `<T,R>`.
- users of `publish<T>(selector)` with a selector that does change the type can compile without errors/warnings, only by using my PR. Those people can remove any workarounds (e.g. `as any`).

benlesh:
I think this just needs refactored for v6

cartant:
@benlesh To me, it looks like this PR doesn't need to be merged. It adds overload signatures to `multicast` and to `publish` for selectors that change the observable's type. However, the current v6 codebase already includes said signatures - they appear to have been added in #2891.

benlesh:
Ugh! Oh no! Am I closing 2 of @hermanbanken's PRs in one day? :crying_cat_face: 

So sorry, @hermanbanken! :pray: 

@cartant is right here, this work is already done.

hermanbanken:
I’m just sad we didn’t get this in Rx 5 @benlesh, but since it’s in 6,
closing this is just fine 😉
Op do 31 mei 2018 om 20:34 schreef Ben Lesh <notifications@github.com>

PR #2624, opened 2017-05-29
docs(exhaust/exhaustMap): fix the example code of exhaust and exhaustMap

SangKa:
**Description:**
In the example code of exhaust and exhaustMap operators, the finite timer need a operator to work with `interval`, such as `take`. The current code that need to be fixed as following:
`var result = clicks.exhaustMap((ev) => Rx.Observable.interval(1000));`


staltz:
LGTM

PR #2629, opened 2017-05-30
fix: eliminate use of instanceof

### LINKED ISSUE #2628: Observables cannot be passed between window contexts.
**RxJS version:** 5.4

**Code to reproduce:**
```javascript
if (window.opener) {
  window.myObservable = window.opener.myObservable;
} else {
  window.myObservable = Rx.Observable.of(1, 2, 3);
  window.open(location.href);
}

window.myObservable
  .switchMap(function(x) {
    return Rx.Observable.of(x);
  })
  .subscribe(function(x) {
    console.log(x);
  });
```

**Expected behavior:**
Both parent and child window should print 1, 2, 3 to the console.

**Actual behavior:**
The child window throws an exception because `myObservable` from the parent window's context cannot be identified as an Observable in the child window's context. 

**Additional information:**
The issue is caused by a reliance on [`instanceof`](https://github.com/ReactiveX/rxjs/blob/5.4.0/src/util/subscribeToResult.ts#L27) for determining if an object is an Observable which [fails for objects passed between javascript contexts](https://github.com/baconjs/bacon.js/issues/296). Instead, a more robust solution would test for the shape of an Observable, perhaps using a type guard similar to [`isPromise`](https://github.com/ReactiveX/rxjs/blob/5.4.0/src/util/isPromise.ts)

goldsam:
**Description:**
Eliminate use of instanceof to support passing observables between javascript contexts

Closes #2628

### DIFF
--- a/package.json
+++ b/package.json
@@ -166,7 +166,7 @@
     "gzip-size": "^3.0.0",
     "http-server": "^0.9.0",
     "husky": "^0.13.3",
-    "lint-staged": "3.2.5",
+    "lint-staged": "3.5.1",
     "lodash": "^4.15.0",
     "madge": "^1.4.3",
     "markdown-doctest": "^0.9.1",
--- a/src/observable/FromObservable.ts
+++ b/src/observable/FromObservable.ts
@@ -12,6 +12,7 @@ import { Observable, ObservableInput } from '../Observable';
 import { Subscriber } from '../Subscriber';
 import { ObserveOnSubscriber } from '../operator/observeOn';
 import { observable as Symbol_observable } from '../symbol/observable';
+import { isObservable } from '../util/isObservable';
 
 /**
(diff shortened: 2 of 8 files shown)


kwonoj:
Leving comment for fyi: let's discussion continues on issue itself at #2628.

benlesh:
Given that the workaround for the corner case issue this is solving is a simple `Observable.from`, I'm opposed to this change unless we can prove there's no performance impact.

Adding a function call, 2 property lookups, 2 `typeof`'s and 3 assertions, instead of a single `instanceof` check everywhere we want to assert that it's *our* Observable instance? Seems extreme to solve a rare problem.

benlesh:
Given the lack of action on this, and my general disagreement, I'm going to close this issue for now.

PR #2630, opened 2017-05-30
docs(advanced): add advanced section

SangKa:
Hi, @benlesh and @kwonoj , this PR is completely a suggestion. If you guys think there is no need fot it, please close it directly.

**Description:**
The `doc/writing-marble-tests.md` and `doc/operator-creation.md` are both so amazing, they are incredible useful rather than the basic topic(eg. Observable, Subject, etc) for the mid-level developer of RxJS. But in the officel document site, there's no entry for them at all, what a pity!

Now I suggest add a entry on the `index.html` for these advanced topic. I configure the `manual.usage` for now, I know it's not exactly for our need, but `manual.advanced` is available in the esdoc 0.5.0. If one day update the esdoc version beyond 0.5.0, you can use `manual.advanced` replace `manual.usage`.

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -12,6 +12,8 @@ RxJS is a library for reactive programming using Observables, to make it easier 
 *Read detailed documentation on each operator*
 ### [» View examples](./manual/tutorial.html)
 *Read our tutorials on using RxJS in real applications*
+### [» Advanced](./manual/usage.html)
+*Read advanced topic about RxJS*
 
 - - -
 
--- a/doc/writing-marble-tests.md
+++ b/doc/writing-marble-tests.md
@@ -7,8 +7,8 @@ from many teachings and documents by people such as @jhusain, @headinthebox, @ma
 
 #### links
 
-- [Contribution](../CONTRIBUTING.md)
-- [Code of Conduct](../CODE_OF_CONDUCT.md)
+- [Contribution](https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md)
+- [Code of Conduct](https://github.com/ReactiveX/rxjs/blob/master/CODE_OF_CONDUCT.md)
 
 ## Basic methods
 
--- a/esdoc.json
+++ b/esdoc.json
@@ -38,6 +38,10 @@
       "./doc/tutorial/basics.md",
       "./doc/tutorial/applications.md",
       "./doc/external-references.md"
+    ],
+    "usage": [
+      "./doc/writing-marble-tests.md",
+      "./doc/operator-creation.md"
     ]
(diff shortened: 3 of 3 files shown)


doc/index.md
```diff
@@ -12,6 +12,8 @@ RxJS is a library for reactive programming using Observables, to make it easier
 *Read detailed documentation on each operator*
 ### [» View examples](./manual/tutorial.html)
 *Read our tutorials on using RxJS in real applications*
+### [» Advanced](./manual/usage.html)
+*Read advanced topic about RxJS*
```
staltz: Hi @SangKa. I'd recommend not adding this line, because this list is supposed to be very short and the 4 existing bullets are already quite a lot.

esdoc.json
```diff
@@ -38,6 +38,10 @@
       "./doc/tutorial/basics.md",
       "./doc/tutorial/applications.md",
       "./doc/external-references.md"
+    ],
+    "usage": [
+      "./doc/writing-marble-tests.md",
+      "./doc/operator-creation.md"
```
staltz: How would this `usage` section be rendered on the website? Can you upload the generated site to some easy link to surge.sh or now.sh or netlify? It would help us know how the site would look like after these changes.

doc/index.md
SangKa: Hi @staltz . I agree with you, it's too long. But It's just a form and the purpose is that add the entry for these awesome topic.

esdoc.json
SangKa: @staltz The screen shot is as below: 
image

staltz: The "usage" section may be misleading to beginners because typically (e.g. in readme files) the usage section explains how to use the library after installing, but in this case "usage" is just a place for advanced content. I would recommend placing it under Tutorial somewhere, instead.

SangKa: I agree with @staltz ,  put them in the tutorial section would be better. I will make the change later.

doc/writing-marble-tests.md
```diff
@@ -7,8 +7,8 @@ from many teachings and documents by people such as @jhusain, @headinthebox, @ma
 
 #### links
 
-- [Contribution](../CONTRIBUTING.md)
-- [Code of Conduct](../CODE_OF_CONDUCT.md)
+- [Contribution](https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md)
```
SangKa: Hi @benlesh . Great to meet you finally, you are too busy. 
I found this links in the output docs wiill not be link correctly if the md file out of the `doc` folder, so  I make this change.

PR #2635, opened 2017-06-03
docs(throttleTime): fix the description of `scheduler` param

SangKa:
**Description:**
Hi, @kwonoj , it's me again. 
In the params of throttleTime operator, there is a issues within the description of `scheduler` param:
`The IScheduler to use for managing the timers that handle the sampling.`
It should be `throttling` rather than `sampling`.


benlesh:
Thanks again, @SangKa!

PR #2636, opened 2017-06-03
docs(toPromise): fix the type and attribute of `PromiseCtor` param

SangKa:
This PR is need you guys to check because I'm not so sure the format and the type of the params whether it is correct. @kwonoj  @staltz @benlesh 

**Description:**
In the param of toPromise operator, now the `PromiseCtor` param is as below:
`@param PromiseCtor promise The constructor of the promise. If not provided, it will look for a constructor first in Rx.config.Promise then fall back to the native Promise constructor if available.`

First the `PromiseCtor` is optional, and secondly there is no type for it.


benlesh:
Thank you @SangKa!

PR #2643, opened 2017-06-05
docs(concatMap): remove the unused @return tag

SangKa:
**Description:**
In the docs of  concatMap operator, there are two `@return` tags, the first one is unused.

```
 * @return {Observable} An observable of values merged from the projected
 * Observables as they were subscribed to, one at a time. Optionally, these
 * values may have been projected from a passed `projectResult` argument.
 * @return {Observable} An Observable that emits the result of applying the
 * projection function (and the optional `resultSelector`) to each item emitted
 * by the source Observable and taking values from each projected inner
 * Observable sequentially.
```


benlesh:
Thanks for catching that, @SangKa!

Issue #2661, opened 2017-06-13
GroupBy GroupDurationSubscriber references retained - memory build up


crunchie84:
**RxJS version:**
5.4.0

**Code to reproduce:**

```
const Rx = require('rxjs/Rx');

Rx.Observable.interval(500)
  .groupBy(
    i => i,
    null,
    i => i.take(1)
  )
  .subscribe(
    null,
    err => console.log('err: ' + err),
    () => console.log('completed')
  );
```

1. Invoke code using `node --debug-brk --inspect index.js`
2. attach the chrome debugger
3. take heap snapshot before letting code run
4. after ~30 seconds take another heap snapshot
5. => 60 GroupDurationSubscriber's still around in memory

**Expected behaviour:**

GroupDurationSubscribers should be removed from the (`GroupBy`) parent subscription once completed.

**Actual behaviour:**

`GroupDurationSubscribers` are retained in memory.

<img width="1432" alt="screen shot 2017-06-13 at 14 31 57" src="https://user-images.githubusercontent.com/570645/27082460-1fa9d4c0-5045-11e7-8e4b-a9475c14ae24.png">

**Additional information:**

When the `GroupDurationSubscriber` is completed it invokes [`this.parent.removeGroup`](https://github.com/ReactiveX/rxjs/blob/master/src/operator/groupBy.ts#L245) which [removes the group](https://github.com/ReactiveX/rxjs/blob/master/src/operator/groupBy.ts#L207) but it never invokes `this.parent.remove(this)` to let the parent (`groupBy` operator) remove the reference to the `GroupDurationSubscriber` so it can be garbage collected. 

This behaviour was implemented in [rxjs4](https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/linq/observable/groupbyuntil.js#L62)

Only when the `GroupBy` operator completes are all the subscriptions let go off and garbage collected.

Issue #2660, opened 2017-06-13
GroupBy subscriptions not disposed in due time - durationSelector


crunchie84:
**RxJS version:**
5.4.0
**Code to reproduce:**

```
const ticker = Rx.Observable.interval(100)
  .map(tick => ({
    key: tick
  }))
;


ticker
  .groupBy(
    keySelector => keySelector.key % 2,
    null,
    durationSelector => ticker
      .filter(tick => tick.key > durationSelector.key + 10)
      .do(val => console.log(`[durationSelector] emitting value for group ${durationSelector.key}: ${val.key}`), null, () => console.log(`[durationSelector] completed for group ${durationSelector.key}`))
      .finally(() => console.log(`[durationSelector] unsubcribed for group ${durationSelector.key}`))
//    .take(1) // <--THIS FIXES IT
  )
  .take(3)
  .subscribe(
    val => console.log(`grouping: ${val.key}`),
err => console.error(err),
() => console.log('completed'));
```
Live version can be found at t his jsbin [https://jsbin.com/vilacos/16/edit?js,console](https://jsbin.com/vilacos/16/edit?js,console)

**Expected behavior:**

When the `durationSelector` emits a value the group is completed. This should make the `GroupBySubscriber` also unsubscribe from the `durationSelector` so it can be garbage collected.

We would expect the `finally()` to be invoked directly after the durationSelector has emitted a value because it completes the groupBy grouping thus creating the following console logs:

```
grouping: 0
grouping: 1
[durationSelector] emitting value for group 0: 11
[durationSelector] unsubcribed for group 0
```

but instead the unsubscribe from the durationSelectors are only emitted when the groupBy completes due to the `take(3)` in the outer stream.

Appending `.take(1)` to the durationSelector stream manually fixes this behaviour (note; this was implemented as such in [rxjs4 groupbyuntil](https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/linq/observable/groupbyuntil.js#L54).)

**Actual behavior:**

The `durationSelector` is not unsubscribed when the group has been completed. Only when the `GroupBy` is completed itself it releases all its captured durationSelectors. This results in hoarding of memory until OOM (our case) or parent stream is completed.

hermanbanken:
I added a test to verify the observables returned by the durationSelector where indeed not closed in 4bd3ea2aa. A simple fix is something like 1a03cc29.

PR #2662, opened 2017-06-13
fix(groupBy): unsubscribe & cleanup GroupDurationSubscriber

hermanbanken:
The `groupBy` operator has a `durationSelector` argument that we can use to specify additional closings for the produced groups. However, the durationSelector produces Observables that were never unsubscribed.  (#2660)

Furthermore even if the durationSelector Observable was auto-unsubscribed (due to for example a `take(1)` in that Observable) its subscription would not be removed from the `GroupBySubscriber`, which causes memory usage to grow and eventually causes a OOM exception in @crunchie84's application. (#2661)

This PR solves both problems by explicitly unsubscribing & removing the subscription from the GroupBySubscriber. I also added a test for the unsubscribing part, but could not easily create a test in the same style as the other groupBy-tests without poking into the subscription.

### DIFF
--- a/spec/operators/groupBy-spec.ts
+++ b/spec/operators/groupBy-spec.ts
@@ -903,6 +903,51 @@ describe('Observable.prototype.groupBy', () => {
     expectObservable(source, unsub).toBe(expected, expectedGroups);
   });
 
+  it('should dispose a durationSelector after closing the group',
+  () => {
+    let durationMarble = '-------s';
+    let durations = [];
+    const duration = () => {
+      let o = cold(durationMarble); // duration 7 frames, then stop
+      durations.push(o);
+      return o;
+    };
+    const e1 = hot( '-1-2-1-3-1-4-1-5------------|');
+    const sub =     '^                            !' ;
+    const gr =      '-a-k---l-b-m---n------------|' ;
+    const empty =   '-----------------------------' ;
+
+    let unsubs = [];
+    for (let i = 0; i < gr.length; i++) {
+      if (gr[i] !== '-' && gr[i] !== '|') {
+        unsubs.push(empty.slice(0, i) + '^' + durationMarble.slice(1, -1) + '!');
+      }
+    }
+
+    const a = cold(  '1---1--|'                     );
+    const b = cold(          '1---1--|'             );
+
+    const k = cold(    '2------|'                   );
+    const l = cold(        '3------|'               );
+    const m = cold(            '4------|'           );
+    const n = cold(                '5------|'       );
+
+    const expectedValues = { a, b, k, l, m, n };
+
+    const source = e1.groupBy(
+      (val: string) => val,
+      (val: string) => val,
+      (group: any) => duration()
+    );
+
+    expectObservable(source, sub).toBe(gr, expectedValues);
+    rxTestScheduler.schedule(() => {
+      durations.forEach((d, i) => {
+        expectSubscriptions(d.subscriptions).toBe(unsubs[i]);
+      });
+    }, 500);
+  });
+
   it('should allow using a durationSelector, but keySelector throws', () => {
     const values = {
       a: '  foo',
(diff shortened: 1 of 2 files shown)


src/operator/groupBy.ts
```diff
@@ -240,6 +240,8 @@ class GroupDurationSubscriber<K, T> extends Subscriber<T> {
       group.error(err);
     }
     this.parent.removeGroup(this.key);
+    this.unsubscribe();
+    this.parent.remove(this);
```
kwonoj: wouldn't it be better `GroupBySubscriber::removeGroup` manages subscription as well when it's specified to remove?

trxcllnt: @kwonoj if anything, the `GroupDurationSubscriber` should be a pass-through to the `group` Subject. If we did that, the base Subscriber `_error` and `_complete` implementations would call `unsubscribe`, and the Subscription's unsubscribe would automatically call `parent.remove()`

hermanbanken:
With the change of @trxcllnt we no longer need to do `this.parent.remove(this)` manually (as in the commit of @crunchie84 and me). 

It took me a while to understand why his change was enough. Let me refrase the explanation, maybe it will help anyone understand if it is not clear already:

During unsubscribe the `Subscription` class makes sure to [remove itself from the parents list of active subscriptions](https://github.com/hermanbanken/RxJS/blob/20f439cb16cc76ebf3eb85e5fd934d2856cfa107/src/Subscription.ts#L85-L90), which was already possible as `this._parent` of the GroupDurationSubscriber always correctly [linked to the GroupBySubscriber](https://github.com/ReactiveX/rxjs/blob/master/src/operator/groupBy.ts#L174), but due to overriding `complete` and `error` the `unsubscribe` that is normally called there was not called. This caused the `GroupDurationSubscribers` to build up in `GroupBySubscriber`. 

Previously

```typescript
Rx.Observable.range(0, 10000)
  .concat(Rx.Observable.never())
  .groupBy(x => x, x => x, group => group)
  .subscribe()
``` 

would result in 10000 `GroupDurationSubscribers` retained in `GroupBySubscriber` but with this PR that is no longer true.

arjenvanderende:
@crunchie84 and I have retested our application against these changes. We can happily confirm that it fixes both #2660 and #2661. Thanks for your work on this PR! 👍

spec/operators/groupBy-spec.ts
```diff
@@ -903,6 +903,51 @@ describe('Observable.prototype.groupBy', () => {
...
+    };
+    const e1 = hot( '-1-2-1-3-1-4-1-5------------|');
+    const sub =     '^                            !' ;
+    const gr =      '-a-k---l-b-m---n------------|' ;
+    const empty =   '-----------------------------' ;
+
+    let unsubs = [];
+    for (let i = 0; i < gr.length; i++) {
+      if (gr[i] !== '-' && gr[i] !== '|') {
+        unsubs.push(empty.slice(0, i) + '^' + durationMarble.slice(1, -1) + '!');
```
benlesh: This is arguably less readable than just using the strings:

```
const unsubs = [
  '^------!',
  '-^------!',
  '--^------!,
 /* etc */
];
```

It's more busy work, sure, but tests are about readability.

hermanbanken: Yes true. How about this [incoming].

spec/operators/groupBy-spec.ts
```diff
@@ -905,47 +905,35 @@ describe('Observable.prototype.groupBy', () => {
...
-    const l = cold(        '3------|'               );
-    const m = cold(            '4------|'           );
-    const n = cold(                '5------|'       );
+    const obs = hot('-0-1--------2-|');
+    const sub =     '^              !' ;
+    let unsubs = [
+                    '-^--!',
+                    '---^--!',
+                    '------------^-!',
+    ];
```
benlesh: haha... so much cleaner than the for loop! :smile:

PR #2664, opened 2017-06-14
Revert "fix(elementAt): force unsubscribe when it completes or errors"

benlesh:
Reverts ReactiveX/rxjs#2501

Damnit... this is the SECOND TIME I've done this. @mpodlasin, do you think you can rename your PRs that are similar to this from `fix()` to `NEXT MAJOR fix()`? So sorry. You'll need to resubmit the one for this and the one for `find`


kwonoj:
@benlesh I can make `danger` to block merge to master if PR's labeled as `major`. (by failing checker). If you're interested, will create PR.

benlesh:
@kwonoj sure, why not? Really, we should just always have those PRs target a branch for major... we have some refactoring to do WRT our publishing and branching practices.

kwonoj:
@benlesh yes, once we have branch policy that'd way to go.

PR #2667, opened 2017-06-15
WIP: Adding lettable operators

benlesh:
This is a WIP, do not merge.

Just planting a flag so people see progress being made.

The overall idea here is that inevitably you'll be able to do this:

```ts
import { map, filter } from 'rxjs/operators';

myObservable
  .let(filter(x => x % 2 === 0))
  .let(map(x => x + x))
```

I've got a lot of plans here, and there's clearly a lot of work to do. I'd just ask people to refrain from making PRs against this branch until they've chatted with me about it first, I'm still getting this idea to gel within this repository, and how we're going to migrate things toward a more T-Rx project.


benlesh:
Also of note: There should be NO BREAKING CHANGES in in this PR. This should be able to ship with the next minor version

felixfbecker:
Cool! Is this forward-compatible with the pipeline operator? https://github.com/gilbert/es-pipeline-operator

mikehaas763:
@felixfbecker that'd be awesome because I've been using the pipe operator in F# with Observable operators and have been waiting for this to end up in JS. How cool will this be?!

```typescript
import { map, filter } from 'rxjs/operators';

myObservable
  |> filter(x => x % 2 === 0)
  |> map(x => x + x)
  |> subscribe(x => console.log(x))
```

benlesh:
@felixfbecker ... no it's not:

1. `|>` requires the first argument to be the return value of the previous expression
2. The pipe operator has made ZERO traction in the TC39. 

Given this library was build on the optimistic hope that we'd have the bind `:: operator, which had *more* traction in the committee, but ultimately failed, that's a lesson learned. We'll just stick with good ol' FP, and we can adjust for new syntax when/if it arrives.

src/interfaces.ts
```diff
@@ -0,0 +1,9 @@
+import { Observable } from './Observable';
+
+export type UnaryFunction<T, R> = (source: T) => R;
+
+export type OperatorFunction<T, R> = UnaryFunction<Observable<T>, Observable<R>>;
+
+export type FactoryOrValue<T> = T | (() => T);
+
+export type MonoTypeOperatorFunction<T> = OperatorFunction<T, T>;
```
kwonoj: nit: we'll likely to have `OperatorFunction<T, R=T>` in `next` if we pick this up into, possible candidate for breaking changes.

jasonaden:
It looks like both `compose` and `pipe` are left-to-right implementations. I believe most functional implementations of `compose` are right-to-left (so the value ends up being the return of the first function) and `pipe` (or a `pipe` operator) is used for left-to-right. Should this just be a `pipe` function as well as the current `pipe` on `Observable.prototype`? This would be more consistent for those coming from other functional languages/libraries.

benlesh:
@jasonaden... you're exactly right... Ramda's compose is right to left. On it's own, it's not quite "pipe"... hmm

marcinnajder:
hi

when I saw the fallowing code `myObservable .let(filter(x => x % 2 === 0))`  first time, I was wondering, is really type inference in TypeScript so good that the type of variable `x` will be inferred correctly ? 

I did some investigation and it turned out that TypeScript version 2.4 is required. Does it mean that since RxJS 5.5 TS 2.4 will be required  ?

marcin

.

niieani:
Hi @benlesh. What you've done here is awesome!

I've started working on a project already using the lettable operators, however it's a bit of a hassle having to manually build `rxjs` on each computer that I checkout the source. Would you consider releasing a beta with what's currently on master, so that people could start playing with the new lettables?

Also, I take it all operators will eventually become lettables? I was missing some, like `startWith`, so I had to write my own for the time being, like this:

```js
import {startWith as startWithOnThis} from 'rxjs/dist/cjs/operator/startWith'
export const startWith = (...elements) => (observable) => startWithOnThis.apply(observable, elements)
```

PS
Love the "_This is a WIP, do not merge._" annotation next to a violet "MERGED" status. 🤣

benlesh:
@nileani ... This will all be in 5.5 as soon as we're done porting them over. It shouldn't be too long. Sorry for the inconvenience.

PR #2669, opened 2017-06-16
fix(test-helper): Add test-helper dependency to spec/support/default.opt

jooyunghan:
Since some specs are dependent on global `__root__` (e.g Observable-spec) without explicit `reuiqre/import', running those specs will fail.

To reproduce:

  > npm run build_spec
  > ./node_modules/.bin/mocha --opts spec/support/default.opts spec-js/Observable-spec.js

**Description:**

### DIFF
--- a/spec/support/default.opts
+++ b/spec/support/default.opts
@@ -1,5 +1,6 @@
 --require source-map-support/register
 --require spec-js/helpers/testScheduler-ui.js
+--require spec-js/helpers/test-helper.js
 --ui spec-js/helpers/testScheduler-ui.js
 
 --reporter dot


kwonoj:
I recall I intentionally take this out (https://github.com/jooyunghan/rxjs/commit/ab4dcfb4da7fb83137278258309539d36bd72bde#diff-c9a04f35a74a1d9b965ec30cf0e601e8) and our test suite is currently not intended to support run individual files. 

It's bit unfortunate due to our test fixture setup is quite complicated, but for now I don't plan to support to run individual file suite only.

PR #2673, opened 2017-06-16
[NEXT] chore(changelog): add missing entry about takeWhile

mpodlasin:
Add missing entry about unsubscriptions in takeWhile. At the same time remove doubled entry about first.

Related to #2470


benlesh:
Thanks for catching that @mpodlasin

PR #2676, opened 2017-06-17
refactor(build): share tsconfig.json for each build target

saneyuki:
**Description:**

From TypeScript 2.1, wen [can use `extends` for tsconfig.json](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#configuration-inheritance). This can make our build script simply.


tsconfig.cjs.json
```diff
@@ -0,0 +1,14 @@
...
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "module": "commonjs",
+    "outDir": "./dist/cjs",
+    "target": "ES5"
+  },
+  "include": [
+    "dist/cjs/src/**/*"
+  ],
+  // XXX: if we don't specify this, tsc adds the value of `outDir` to `exclude`
```
kwonoj: this seems bit odd. I remember it shouldn't be needed with correct config inherits. what brings this for our config?

tsconfig.json
```diff
@@ -1,30 +1,6 @@
 {
-  "compilerOptions": {
```
kwonoj: where do we use this config?

saneyuki: > where do we use this config?

This file exists to support some editors/IDE (e.g. vscode).

kwonoj: I'd rather make this config for base config either cjs or es2015 config and get rid one of them. don't feel we need unusued one.

tsconfig.cjs.json
saneyuki: > this seems bit odd. I remember it shouldn't be needed with correct config inherits. what brings this for our config?

If I remove this `exclude`, then tsc says: 

> error TS18003: No inputs were found in config file '/Users/tetsuharu/src/js/rxjs/tsconfig.cjs.json'. Specified 'include' paths were '["dist/cjs/src/**/*"]' and 'exclude' paths were '["./dist/cjs"]'.


`npm run copy_src_{cjs,es6}` requires that the source and the dist should be placed in the same file: https://github.com/ReactiveX/rxjs/blob/master/package.json#L75-L76

I feel we might have more refactoring chance to there process. But I also think it should be opened as an another pull request.

kwonoj: So if this comes from our custom config, I'd rather take PR correctly setup (probably we should simply copy tsconfig itself when we copy src) or either just maintain current setup. 

I am in favor of make config simpler but feeling this should be included PR to correclty amend with our build pipeline.

tsconfig.json
saneyuki: > I'd rather make this config for base config either cjs or es2015 config and get rid one of them. don't feel we need unusued one.

`src`/`include` in tsconfig.json would be a problem to do it. Our build system use `dist/{cjs,es6}/**/*` as build source, not `src/` directly.

If we change your proposed one, we also need to change the build source dir in other parts. It might be [a chance to fix the above problem](https://github.com/ReactiveX/rxjs/pull/2676#discussion_r122578326), but I'm not sure about that such refactoring would resolve them at once.

kwonoj: Then honestly I'd postpone this changes and see another chance for refactoring. This PR obviously makes config simpler, but functionwise everything should be same, and to make that introducing another workaround seems bit off to me. Our bulid pipeline already have sufficient workaround to achieve our goal and I'd like to move way to remove those.

saneyuki:
@kwonoj okay, I'll look into the way to use `src/` as our build source.

saneyuki:
@kwonoj 

I have updated my changeset. How do you think about these?

PR #2677, opened 2017-06-19
perf(ReplaySubject): slightly improved performance

martinsik:
I was going through the `ReplaySubject` class and I think in some situations it can be simplified. Most importantly when I don't set the `timeWindow` parameter the buffer trimming method can be avoided completely and I can store just the values without wrapping them with `ReplayEvent`.

Also, I tried some performance tests to see whether `splice()` or `shift()` would be better. It seems that running `shift()` in a for loop is faster than `splice()` when removing `< 25` items. However, `splice()` is more consistent even when removing a large portion of an array. See https://jsperf.com/splice-vs-shift-2

There were no benchmarks for Subjects in `perf` directory so I made a few myself but I'm not sure I'm testing it the right way. It shows that RxJS 4 implementation was faster than RxJS 5...

Before:

```
                                         |                     RxJS 4.1.0 |                     RxJS 5.4.1 |          factor |      % improved
    replaysubject_windowtime - immediate |                68,591 (±1.36%) |                17,151 (±2.57%) |           0.25x |          -75.0%
               replaysubject - immediate |                72,951 (±1.08%) |                17,099 (±2.69%) |           0.23x |          -76.6%
                replaysubject_windowtime |                39,104 (±0.85%) |                15,393 (±2.85%) |           0.39x |          -60.6%
                           replaysubject |                34,721 (±1.13%) |                15,393 (±2.30%) |           0.44x |          -55.7%
```

After:

```
                                        |                     RxJS 4.1.0 |                     RxJS 5.4.1 |          factor |      % improved
    replaysubject_windowtime - immediate |                63,306 (±1.14%) |                19,745 (±2.68%) |           0.31x |          -68.8%
               replaysubject - immediate |                63,870 (±3.35%) |                28,659 (±2.49%) |           0.45x |          -55.1%
                replaysubject_windowtime |                36,607 (±1.78%) |                16,646 (±2.83%) |           0.45x |          -54.5%
                           replaysubject |                36,058 (±1.29%) |                22,130 (±2.86%) |           0.61x |          -38.6%
```


kwonoj:
lol I need to update size diff output

src/ReplaySubject.ts
```diff
@@ -90,12 +116,13 @@ export class ReplaySubject<T> extends Subject<T> {
       spliceCount = Math.max(spliceCount, eventsCount - _bufferSize);
     }
 
-    if (spliceCount > 0) {
-      _events.splice(0, spliceCount);
+    for (let i = 0; i < spliceCount; i++) {
+      _events.shift();
     }
```
benlesh: I'm actually shocked this would be faster.  this section was O(1) now it's O(n).

benlesh:
@martinsik the change looks fantastic, just one thing I'd like you to verify if possible.

I suspect that `shift()` is faster for 1 or 2 items, but `splice` is faster for more than that. If that's the case, we just need to accommodate the 80% case and move on (with a comment to note why it's a loop and not a splice)

martinsik:
@benlesh I made this test to see which one is faster: https://jsperf.com/splice-vs-shift-2
It seems like even this:
```
for (var i = 0; i < 25; i++) {
    arr.shift();
}
```

is faster than this:

```
arr.splice(0, 25);
```

I don't know if there're any other aspects to consider. I guess size of the array shouldn't matter but maybe there are some specific cases when it does?

benlesh:
Yeah, it looks like what I suspected. Which is that for small arrays `shift()` will be more efficient, but for larger arrays `splice()` wins out (the `50` vs `50` tests show a 2x perf increase for `splice`)

While there's some serious wins for the smaller arrays, I think they die off quickly enough that it's worth looking at the more consistent (and better for a larger range of sizes) performance of `splice`.

I think we should move it back to `splice` to keep performance consistent.

src/ReplaySubject.ts
benlesh: After reviewing perf, i think we should return it to `splice` because of it's consistent performance, even if `shift` is faster on smaller arrays.

martinsik:
@benlesh I understand, I switched back to `splice()`.

I'm not sure about the performance test I made. It shows that RxJS 4 is faster than RxJS 5 but is this even possible?

martinsik:
@benlesh I just want to ask if this PR is still relevant or it needs updates for 5.5 or 6.0 :).

PR #2678, opened 2017-06-19
chore(danger): update size diff output

kwonoj:
**Description:**


benlesh:
Looks like we started working on the same problem at the same time. I added colors tho :stuck_out_tongue: 

Closing this for #2679 (also because @kwonoj said it was okay) :heart:

PR #2684, opened 2017-06-20
Update to TypeScript 2.4

david-driscoll:
**Description:**
Fixed code that was failing due to compiler upgrade

Several methods are now using default generic arguments, which were introduced in ts2.3


src/operator/catch.ts
```diff
@@ -64,16 +64,16 @@ import { subscribeToResult } from '../util/subscribeToResult';
...
+export function _catch<T, R = T>(this: Observable<T>, selector: (err: any, caught: Observable<R>) => ObservableInput<R>): Observable<R> {
   const operator = new CatchOperator(selector);
   const caught = this.lift(operator);
   return (operator.caught = caught);
 }
 
-class CatchOperator<T, R> implements Operator<T, T | R> {
-  caught: Observable<T>;
+class CatchOperator<T, R = T> implements Operator<T, R> {
+  caught: Observable<R>;
```
david-driscoll: This was failing without this change, I wasn't entirely sure it made sense, but looking at the code it should have always been `Observable<T | R>` and never `Observable<T>`

package.json
```diff
@@ -164,40 +168,40 @@
     "doctoc": "^1.0.0",
     "escape-string-regexp": "^1.0.5 ",
     "esdoc": "^0.4.7",
-    "eslint": "^3.8.0",
+    "eslint": "^4.0.0",
```
kwonoj: not relevant to this PR, but I'm feeling we just need to drop eslint and use tslint everywhere..

src/Observable.ts
```diff
@@ -15,7 +15,7 @@ export interface Subscribable<T> {
 }
 
 export type SubscribableOrPromise<T> = Subscribable<T> | PromiseLike<T>;
-export type ObservableInput<T> = SubscribableOrPromise<T> | ArrayLike<T>;
+export type ObservableInput<T> = Subscribable<T> | PromiseLike<T> | ArrayLike<T>;
```
kwonoj: what was reason for not using `SubscribableOrPromise` here?

tsconfig.json
```diff
@@ -4,11 +4,12 @@
     "removeComments": false,
     "preserveConstEnums": true,
     "sourceMap": true,
-    "declaration": true,
+    // "declaration": true,
```
kwonoj: commented?

tsconfig.json
```diff
@@ -24,7 +25,12 @@
     "indentSize": 2,
     "tabSize": 2
   },
-  "files": [
-    "src/Rx.ts"
+  "include": [
+    "src"
```
kwonoj: do we need this changes in this PR?

tsconfig.json
david-driscoll: not allowed with `allowJs`, going to back that part out (and leave the lower half)

I'll bring this back in another PR later.

kwonoj: as same as below Q, do we need to modify tsconfig for compiler bump up? (esp. allowJS?)

tsconfig.json
david-driscoll: I wouldn't use need, I can split it into another commit if needed.

kwonoj: let's split it. Especially current tsconfig.json is not being used actively (npm script runs command for its own).

package.json
```diff
@@ -138,6 +138,10 @@
     {
       "name": "Andre Staltz",
       "email": "andre@staltz.com"
+    },
+    {
+      "name": "David Driscoll",
```
kwonoj: 👏

Issue #2535, opened 2017-04-09
ObservableInput is missing Iterable


felixfbecker:
It's easy to convert an array to an Observable, but not an ES6 Map or Set or custom Iterable.
There is `IteratorObservable` though: https://github.com/ReactiveX/rxjs/blob/master/src/observable/IteratorObservable.ts
but it needs to be imported as `import { IteratorObservable } from '@reactivex/rxjs/dist/cjs/observable/IteratorObservable';` and the parameter is just typed as `any` instead of the `string | ArrayLike<T> | Iterable<T>` [it apparently accepts](https://github.com/ReactiveX/rxjs/blob/master/src/observable/IteratorObservable.ts#L121). Also weird it doesn't accept `Iterator<T>`, given the name.

If an Iterable is passed to `from()`, it should just call `Symbol.iterator()` on it and use an IteratorObservable. This would make working with ES6 collections so much easier and enable interop between sync, pull-style iterators and async, push-style Observables.

felixfbecker:
Found https://github.com/ReactiveX/rxjs/issues/2306

PR #2685, opened 2017-06-20
Bring back Iterable<T> and Iterator<T> interfaces

### LINKED ISSUE #2306: Observable.from(iterable) generates Typescript compiler error
**RxJS version:** 5.0.3

**Code to reproduce:**
```
function* generateDoubles(seed: number) {
  let i = seed;
  while (true) {
    yield i;
    i = 2 * i; // double it
  }
}

const iterator = generateDoubles(3);
const result = Rx.Observable.from<number>(iterator).take(10);
result.subscribe((x: number) => console.log(x));
```

**Expected behavior:**
Since this is the example given on  [reactivex.io](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-from)
It would be expected to compile in Typescript without an error.

**Actual behavior:**
```
error TS2345: Argument of type 'IterableIterator<number>' is not assignable to parameter of type 'ObservableInput<number>'.
  Type 'IterableIterator<number>' is not assignable to type 'PromiseLike<number>'.
    Property 'then' is missing in type 'IterableIterator<number>'.
```

**Additional information:**
This seems to be an issue with the definition of `ObservableInput<T>`:
```
export interface Subscribable<T> {
    subscribe(observerOrNext?: PartialObserver<T> | ((value: T) => void), error?: (error: any) => void, complete?: () => void): AnonymousSubscription;
}
export declare type SubscribableOrPromise<T> = Subscribable<T> | PromiseLike<T>;
export declare type ObservableInput<T> = SubscribableOrPromise<T> | ArrayLike<T>;
```

The iterable type is missing.

A fix might be to add:
```
export interface IterableLike<T> {
    [Symbol.iterator](): Iterator<T> | IterableIterator<T>;
}

// and fix ObservableInput<T>
export declare type ObservableInput<T> = SubscribableOrPromise<T> | ArrayLike<T> | IterableLike<T>;
```

**tsconfig.json**
```
{
	"compilerOptions": {
		"target": "es2015",
		"module": "commonjs",
		"moduleResolution": "node",
		"sourceMap": true,
		"alwaysStrict": true,
        "declaration": true,
		"noImplicitAny": true,
		"noImplicitThis": true,
		"strictNullChecks": true,
		"outDir": "dist",
		"lib": ["dom", "dom.iterable", "es2015", "es2015.iterable", "scripthost"]
	},
	"include": [
		"src"
	]
}
```

david-driscoll:
**Description:**
Now that TypeScript supports down level compilation, we can safely bring back `Iterable<T>` and `Iterator<T>`.

Fixes #2306 #2535

### DIFF
--- a/src/observable/IteratorObservable.ts
+++ b/src/observable/IteratorObservable.ts
@@ -1,9 +1,18 @@
 import { root } from '../util/root';
 import { IScheduler } from '../Scheduler';
-import { Observable } from '../Observable';
+import { Observable, ArrayOrIterable } from '../Observable';
 import { iterator as Symbol_iterator } from '../symbol/iterator';
 import { TeardownLogic } from '../Subscription';
 import { Subscriber } from '../Subscriber';
+import { isArrayLike } from '../util/isArrayLike';
+
+interface IteratorObservableState<T> {
+  subscriber: Subscriber<T>;
+  iterator: Iterator<T>;
+  index: number;
+  hasError?: boolean;
+  error?: any;
+}
 
 /**
  * We need this JSDoc comment for affecting ESDoc.
@@ -11,13 +20,13 @@ import { Subscriber } from '../Subscriber';
  * @hide true
  */
 export class IteratorObservable<T> extends Observable<T> {
-  private iterator: any;
+  private iterator: Iterator<T>;
 
-  static create<T>(iterator: any, scheduler?: IScheduler): IteratorObservable<T> {
-    return new IteratorObservable(iterator, scheduler);
+  static create<T>(iterator: ArrayOrIterable<T> | string, scheduler?: IScheduler): IteratorObservable<T> {
+    return new IteratorObservable<T>(iterator, scheduler);
   }
 
-  static dispatch(state: any) {
+  private static dispatch<T>(state: IteratorObservableState<T>) {
 
     const { index, hasError, iterator, subscriber } = state;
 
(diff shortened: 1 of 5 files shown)


david-driscoll:
cc @benlesh

src/observable/IteratorObservable.ts
```diff
@@ -1,23 +1,32 @@
 import { root } from '../util/root';
 import { IScheduler } from '../Scheduler';
-import { Observable } from '../Observable';
+import { Observable, ArrayOrIterable } from '../Observable';
 import { iterator as Symbol_iterator } from '../symbol/iterator';
 import { TeardownLogic } from '../Subscription';
 import { Subscriber } from '../Subscriber';
+import { isArrayLike } from '../util/isArrayLike';
+
+export interface IteratorObservableState<T> {
```
kwonoj: is this need to be exported?

src/util/isIterable.ts
```diff
@@ -0,0 +1,5 @@
+import { iterator as Symbol_iterator } from '../symbol/iterator';
+
+export function isIterable<T>(value: any): value is Iterable<T> {
```
kwonoj: generic param T is required?

david-driscoll: It helps flow the types through, `isPromise` and `isArrayLike` are also both generic.

src/observable/IteratorObservable.ts
david-driscoll: Yes, compiler complains otherwise (because dispatch is public).  If we made dispatch private it probably wouldn't need to be.

kwonoj: I think we may need, I recall some of dispatch is private and interfaces are hidden already. No need to expose unnecessary interface.

david-driscoll:
Updated

david-driscoll:
cc @benlesh I can merge at will right? 🚢 😄

HipsterZipster:
When is this getting released? I don't see it in RxJS@5.4.2

PR #2691, opened 2017-06-22
fix(never): Observable.never() Observable.empty() .ignoreElements () now return Observable<never>

### LINKED ISSUE #2640: Return type of .ignoreElements() and .empty() should be Observable<never>
**RxJS version:** 5.4.0

**Code to reproduce:**
```ts
/** Returns an Observable of files */
function fetchFiles(): Observable<string> {
  return Observable.of('foo', 'bar', 'baz')
}

const files = new Set<string>()

/** Returns an Observable that emits no items and completes when the task is done */
function updateFiles() {
  return fetchFiles()
    .do(file => {
      files.add(file)
    })
    .ignoreElements()
}
```

**Expected behavior:**
Inferred return type of `updateFiles` should be `Observable<never>`. The Observable never emits items, so the items don't have a type, similar to how a function that always throws has no return type - it `never` emits/returns.

**Actual behavior:**
Emitted return type is `Observable<string>`

jayphelps:
fixes #2640

BREAKING CHANGE: Previously, `Observable.never()` `Observable.empty()`
and the `.ignoreElements()` operator all returned `Observable<T>` which
was incorrect since they actually never emit anything. Now they all
return `Observable<never>` (`never` was added in TS 2.0 as a special
type)

### DIFF
--- a/src/observable/EmptyObservable.ts
+++ b/src/observable/EmptyObservable.ts
@@ -3,8 +3,8 @@ import { Subscriber } from '../Subscriber';
 import { Observable } from '../Observable';
 import { TeardownLogic } from '../Subscription';
 
-export interface DispatchArg<T> {
-  subscriber: Subscriber<T>;
+export interface DispatchArg {
+  subscriber: Subscriber<never>;
 }
 
 /**
@@ -12,7 +12,7 @@ export interface DispatchArg<T> {
  * @extends {Ignored}
  * @hide true
  */
-export class EmptyObservable<T> extends Observable<T> {
+export class EmptyObservable extends Observable<never> {
 
   /**
    * Creates an Observable that emits no items to the Observer and immediately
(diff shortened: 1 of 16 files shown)


src/observable/EmptyObservable.ts
```diff
@@ -57,11 +57,11 @@ export class EmptyObservable<T> extends Observable<T> {
...
    * @owner Observable
    */
-  static create<T>(scheduler?: IScheduler): Observable<T> {
-    return new EmptyObservable<T>(scheduler);
+  static create(scheduler?: IScheduler): Observable<never> {
+    return new EmptyObservable(scheduler);
   }
 
-  static dispatch<T>(arg: DispatchArg<T>) {
+  static dispatch(arg: DispatchArg<never>) {
```
kwonoj: wouldn't it better to make `DispatchArg` as non-generic in line 6? (also making this fn to private and not export dispatcharg would be preferred)

jayphelps: wasn't clear if DispatchArg was used elsewhere since it was exported, but certainly can 👍

jayphelps: it will basically just mean I make `subscriber: Subscriber<never>` instead of `subscriber: Subscriber<T>`

david-driscoll:
You should probably target the `next` branch instead of `master`.

kwonoj:
(I've changed base to `next`)

kwonoj:
ah, change itself LGTM

jayphelps:
@kwonoj issue is that next doesn't have everything from master yet, so my branch is bringing them along for the ride. I can prolly reset my branch to the same as `next` then cherry pick my commit in, but, or next can be updated to the latest master before this is merged. LMK what ya'll would like

jayphelps:
@kwonoj reset to `next` then cherry picked commit into it.

jayphelps:
Need anything else for merge?

PR #2692, opened 2017-06-22
feat(typings): propogate thisArgs using this syntax

david-driscoll:
**Description:**
updated all operators to use this context, if they supply a thisArg, (map, filter, etc)

### DIFF
--- a/spec/operators/every-spec.ts
+++ b/spec/operators/every-spec.ts
@@ -2,7 +2,7 @@ import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
 import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
 
-declare const { asDiagram };
+declare const asDiagram: Function;
 declare const hot: typeof marbleTestingSignature.hot;
 declare const cold: typeof marbleTestingSignature.cold;
 declare const expectObservable: typeof marbleTestingSignature.expectObservable;
@@ -12,11 +12,11 @@ const Observable = Rx.Observable;
 
 /** @test {every} */
 describe('Observable.prototype.every', () => {
-  function truePredicate(x) {
+  function truePredicate(x: never) {
     return true;
   }
 
-  function predicate(x) {
+  function predicate(x: number) {
     return x % 5 === 0;
   }
 
@@ -55,13 +55,13 @@ describe('Observable.prototype.every', () => {
       observer.next(1);
       observer.complete();
     })
-    .every(function (value: number, index: number) {
+    .every(function (this: typeof thisArg, value: number, index: number) {
       expect(this).to.deep.equal(thisArg);
     }, thisArg).subscribe();
   });
 
   it('should emit true if source is empty', () => {
-    const source = hot('-----|');
+    const source = hot('-----|', <{ [index: string]: number; }>{});
     const sourceSubs = '^    !';
     const expected =   '-----(x|)';
 
@@ -70,7 +70,7 @@ describe('Observable.prototype.every', () => {
   });
 
   it('should emit false if single source of element does not match with predicate', () => {
-    const source = hot('--a--|');
+    const source = hot('--a--|', <{ [index: string]: number; }>{});
     const sourceSubs = '^ !';
     const expected =   '--(x|)';
 
@@ -79,7 +79,7 @@ describe('Observable.prototype.every', () => {
   });
 
   it('should emit false if none of element does not match with predicate', () => {
-    const source = hot('--a--b--c--d--e--|');
+    const source = hot('--a--b--c--d--e--|', { a: 1, b: 2, c: 3, d: 4, e: 5 });
     const sourceSubs = '^ !';
     const expected =   '--(x|)';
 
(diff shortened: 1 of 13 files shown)


kwonoj:
I guess this isn't breaking changes?

david-driscoll:
It's not breaking, but it I don't want to risk bringing it down into `master`, and just leave it for `next`.

david-driscoll:
any input @kwonoj @benlesh ? 🕺 💃

spec/operators/every-spec.ts
```diff
@@ -12,12 +12,12 @@ const Observable = Rx.Observable;
...
 describe('Observable.prototype.every', () => {
-  function truePredicate(x) {
+  function truePredicate(x: never) {
     return true;
   }
 
-  function predicate(x) {
-    return x % 5 === 0;
+  function predicate(x: string | number) {
+    return (<any>x) % 5 === 0;
```
benlesh: This seems invalid to me... like predicate should only take a number.

spec/operators/every-spec.ts
```diff
@@ -167,7 +167,7 @@ describe('Observable.prototype.every', () => {
     const source = Observable.of(3);
     const expected = '#';
 
-    function faultyPredicate(x) {
+    function faultyPredicate(x: never) {
```
benlesh: Did you mean `number` here instead of `never`?

spec/operators/filter-spec.ts
```diff
@@ -12,11 +12,11 @@ const Observable = Rx.Observable;
 
 /** @test {filter} */
 describe('Observable.prototype.filter', () => {
-  function oddFilter(x) {
+  function oddFilter(x: number | string) {
```
benlesh: should probably just be `number`, same for `isPrime` below... if we have to change the tests, we should. Also I think we should probably use `Number(x)` instead of `+x` in spots because it's a little easier to read. (even tho I am a fan of that trick)

spec/operators/filter-spec.ts
```diff
@@ -178,18 +178,18 @@ describe('Observable.prototype.filter', () => {
 
     expectObservable(
       source
-        .filter((x: number) => x % 2 === 0)
-        .filter((x: number) => x % 3 === 0)
+        .filter((x) => parseInt(x) % 2 === 0)
```
benlesh: use `Number(x)` instead of `parseInt` or `parseFloat`

spec/operators/filter-spec.ts
```diff
@@ -178,18 +178,18 @@ describe('Observable.prototype.filter', () => {
...
 
   it('should be able to accept and use a thisArg', () => {
     const source = hot('-1--2--^-3-4-5-6--7-8--9--|');
     const expected =          '--------6----------|';
 
-    function Filterer() {
-      this.filter1 = (x: number) => x % 2 === 0;
-      this.filter2 = (x: number) => x % 3 === 0;
+    class Filterer {
+      filter1 = (x: string) => (+x) % 2 === 0;
```
benlesh: `Number(x)` instead.

spec/operators/filter-spec.ts
```diff
@@ -178,18 +178,18 @@ describe('Observable.prototype.filter', () => {
...
+      filter2 = (x: string) => (+x) % 3 === 0;
```
benlesh: `Number(x)` instead

spec/operators/find-spec.ts
```diff
@@ -13,7 +13,7 @@ const Observable = Rx.Observable;
 
 /** @test {find} */
 describe('Observable.prototype.find', () => {
-  function truePredicate(x) {
+  function truePredicate(x: never) {
```
benlesh: again, I'm not sure what `never` is.. :\ Perhaps I'm just dumb.

spec/operators/map-spec.ts
```diff
@@ -214,8 +214,8 @@ describe('Observable.prototype.map', () => {
     let invoked1 = 0;
     let invoked2 = 0;
     const r = a
-      .map((x: string) => { invoked1++; return parseInt(x) * 2; })
-      .map((x: number) => { invoked2++; return x / 2; })
+      .map((x) => { invoked1++; return parseInt(x) * 2; })
```
benlesh: use `Number(x)` rather than `parseInt`

spec/operators/map-spec.ts
```diff
@@ -231,9 +231,9 @@ describe('Observable.prototype.map', () => {
     const expected = '--a--b--c--d--|';
     const values = {a: 11, b: 14, c: 17, d: 20};
 
-    function Filterer() {
-      this.selector1 = (x: string) => parseInt(x) + 2;
-      this.selector2 = (x: string) => parseInt(x) * 3;
+    class Filterer {
+      selector1 = (x: any) => parseInt(x) + 2;
```
benlesh: use `Number(x)` rather than `parseInt`

spec/operators/map-spec.ts
```diff
@@ -231,9 +231,9 @@ describe('Observable.prototype.map', () => {
...
+      selector2 = (x: any) => parseInt(x) * 3;
```
benlesh: use `Number(x)` rather than `parseInt`

spec/operators/partition-spec.ts
```diff
@@ -25,7 +25,7 @@ describe('Observable.prototype.partition', () => {
     const expected = ['--1-----3---------5------|',
                     '----2----------4------6--|'];
 
-    const result = e1.partition((x: any) => x % 2 === 1);
+    const result = e1.partition((x) => parseInt(x) % 2 === 1);
```
benlesh: use `Number(x)` rather than `parseInt`

spec/operators/every-spec.ts
david-driscoll: I was just using `never` since the value is never used in the actual method.  Regardless either way works. 👍

david-driscoll:
Updated for feedback

kwonoj:
I think this seems ready 🤔

PR #2715, opened 2017-07-01
fix(Notification): Don't reference `this` from static methods.

anthonydmays:
**Description:**
Fixes bug caused by referencing `this` from static methods. 

### DIFF
--- a/src/Notification.ts
+++ b/src/Notification.ts
@@ -107,7 +107,7 @@ export class Notification<T> {
     if (typeof value !== 'undefined') {
       return new Notification('N', value);
     }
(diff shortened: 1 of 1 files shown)


anthonydmays:
Sorry for being a TypeScript noob, but just realized that "this" is legal in static methods. My issue apparently stems from the Google Closure JSCompiler not doing the right thing when transpiling down to ES5. This change would fix my issue, but wouldn't address the real problem.

benlesh:
This change is pretty innocuous. If it solves a problem I don't see any reason to not merge it. I'll leave it open for others to review.

cc @kwonoj

PR #2721, opened 2017-07-04
Fix (Observable.fromEvent): event listener with options is not removed

zy-serguei:
removeEventListener must be called with the same options used in addEventListener

**Description:** This problem is noticeable in combination with Angular2. When having subscribed to and then unsubscribed from an observable created with Observable.fromEvent with options, each time the event is triggered it causes Angular's change detection.


zy-serguei:
Is there a way to change the commit message via the web interface? I couldn't find it. Please feel free to change it if it's not much trouble for you.

kwonoj:
no, you have to change via local git `git --commit-amend` and update branch. Still, you may need to update commit to include test cases.

benlesh:
Really this just needs a test case and it's good. I can change the commit message if I have to.

benlesh:
This is already done in master.

PR #2722, opened 2017-07-04
fix(Subject): lift signature is now appropriate for stricter TS 2.4 c…

benlesh:
…hecks

Type safety wasn't really gauranteed with lift before because we survived for a long time with an incorrect type signature, TypeScript 2.4 introduces stricter type-checking, and all of a sudden `lift` on `Subject` was breaking builds for those that are riding the wave of the latest-and-greatest.

EDIT: Upon review, really not sure of any way this change could break anyone. So I'm going to push this through as a patch. It was basically building with an incorrect type signature in TS 2.3 and lower, so correcting it should not break anyone's builds.

### DIFF
--- a/src/Subject.ts
+++ b/src/Subject.ts
@@ -43,7 +43,7 @@ export class Subject<T> extends Observable<T> implements ISubscription {
     return new AnonymousSubject<T>(destination, source);
   }
 
-  lift<R>(operator: Operator<T, R>): Observable<T> {
+  lift<R>(operator: Operator<T, R>): Observable<R> {
     const subject = new AnonymousSubject(this, this);
     subject.operator = <any>operator;
     return <any>subject;


hansl:
LGTM :+1:

I think it's highly unlikely, unless someone uses `lift` incorrectly which should have undefined behaviour.

saneyuki:
@benlesh 

I feel we can close https://github.com/ReactiveX/rxjs/issues/2539 with "we has released hotfix once for this problem. please file a new issue if you find a corner case that v5.4.2 could not caught". What to do?

kgentes:
#2539 was closed, it appears, but because of CoC issues not due to it being tagged as resolved. Just for linking and clarity, it might be good to have it marked as fixed in the new release (unless that is not what this fix is claiming).  Thanks for the quick resolve on this!

fletchsod-developer:
https://github.com/ReactiveX/rxjs/pull/2540
https://github.com/Microsoft/TypeScript/issues/16593

GulinSS:
Seems not enough, did
```
npm install rxjs@5.4.2 --save
```

But still see same issue for typescript 2.4.1, need update `.d.ts` files also

Issue #2656, opened 2017-06-12
Documentation missing for operator / interface need to be updated


kwonoj:
**RxJS version:**

**Code to reproduce:**

**Expected behavior:**

**Actual behavior:**

**Additional information:**
This is umbrella issue to track documentation status. It is known there are missing documentations around operators, interfaces and website(reactivex.io) need to be updated.

`Subject.create` https://github.com/ReactiveX/rxjs/issues/2655
`Observable.if` https://github.com/ReactiveX/rxjs/issues/2648
`Timeinterval` https://github.com/ReactiveX/rxjs/issues/2627
`installation guide on http://reactivex.io/rxjs` https://github.com/ReactiveX/rxjs/issues/2482
`finally` https://github.com/ReactiveX/rxjs/issues/2256
https://github.com/ReactiveX/rxjs/issues/2656

SangKa:
I will supplement it:
 - publishBehavior
 - publishLast
 - publishReplay
 - timeout
 - timeoutWith
 - timeoutWith
 - timeoutWith
 - timeoutWith

wldcordeiro:
* `ajax`/`get/post/put/etc`

Since @kwonoj closed my issue I'd figured I'd add to the list.

mpodlasin:
I created `timeout` docs: https://github.com/ReactiveX/rxjs/pull/2724.
Also gentle reminder, that of those mentioned here, I have opened PRs for `if` (https://github.com/ReactiveX/rxjs/pull/2419) and `finally` (https://github.com/ReactiveX/rxjs/pull/2511). 
:)

benjamenweber:
It doesn't look like there is an issue to update the website.  It would be nice if at least some of the merged documentation updates made it there.  Ended up finding this from #2400 and #2449

JSchneidler:
`forkJoin` is also missing documentation. I would submit a PR for it if I was more experienced with RxJS.

rinfan:
@JSchneidler it's already there https://github.com/ReactiveX/rxjs/pull/2400/commits/5cdef625ed7ed312399aec4c2a2ebe62fb0a210b  there just hasnt been any updates to the website to use newly generated docs in who knows how long.

dahei:
Please update the website docs. It's very confusing for newcomers not to find the API-docs for e.g. "Observable.subscribe()"

[...]

PR #2732, opened 2017-07-10
docs(toArray): create documentation for operator

crunchie84:
**Description:**

Add documentation for operator including marble diagram, reference to buffer and code example.

https://github.com/ReactiveX/rxjs/issues/2656

### DIFF
--- a/src/operator/toArray.ts
+++ b/src/operator/toArray.ts
@@ -3,6 +3,24 @@ import { Subscriber } from '../Subscriber';
 import { Observable } from '../Observable';
 
 /**
+ * Creates an array containing all emissions of an Observable.
+ *
+ * <span class="informal">It's like {@link buffer } with the source Observable completion as closingNotifier.</span>
+ *
+ * <img src="./img/toArray.png" width="100%">
+ *
+ * `toArray` will wait until the Observable completes before emitting
+ * the array containing all emissions. If the Observable errors no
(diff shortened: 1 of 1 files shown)


src/operator/toArray.ts
```diff
@@ -3,6 +3,24 @@ import { Subscriber } from '../Subscriber';
 import { Observable } from '../Observable';
 
 /**
+ * Creates an array containing all emissions of an Observable.
```
staltz: @crunchie84 Thanks for this JSDoc.

I'd recommend that this first sentence emphasizes two key things: returns an Observable (not an array), when the source completes.

For example:

> Collects all source emissions and emits them as an array when the source completes.

Then, the informal description can be even more informal (try not mentioning closingNotifier because it assumes the reader knows the details of the buffer operator), try to imagine you're using conversational language with a co-worker:

> Emits all the values at once, as an array, when the source completes.

crunchie84: Seems sane to simplify, i'll update the wording a bit. 

One thing that i found odd; when compiling the documentation I get a marble diagram image in `/tmp/docs/img/toArray.png` but it only shows the marbles without text. Is that somehow post-processed when releasing the documentation?

staltz:
Can I merge this? No objections?

kwonoj:
@staltz yeah 🚢

PR #2740, opened 2017-07-11
docs(multicast): enhance the documentation

mpodlasin:
Make clear that multicast can be used in two separate modes. Describe two modes and provide examples for both. Add note how multicast relates to Subjects and several publish and share operators.


benlesh:
Thanks @mpodlasin!

PR #2743, opened 2017-07-12
fix(audit): fix crashes in case of the duration Observable completes

asnov:
**Description:**
fix the problem when subscribeToResult() returns null it will crash described by @deadbeef84
fixes #2595

### DIFF
--- a/spec/operators/audit-spec.ts
+++ b/spec/operators/audit-spec.ts
@@ -139,6 +139,18 @@ describe('Observable.prototype.audit', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
+  it('should mirror source if durations are Observable.of()', () => {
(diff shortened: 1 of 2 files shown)


spec/operators/audit-spec.ts
```diff
@@ -139,6 +139,18 @@ describe('Observable.prototype.audit', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
+  it('should mirror source if durations are Observable.of()', () => {
```
kwonoj: I'd rather name it for synchronous observable instead of specifically say `of` operator (this could happen for any synchronous observable).

kwonoj:
I'm labelling this as patch for now.

kwonoj:
@benlesh I think this can go with 5.5?

deadbeef84:
Why is this still not fixed? :(

PR #2745, opened 2017-07-12
docs: style problem in light theme

m98:
When the website  chosen theme is on the light mode, and you search something, mouse hovering on link will make it disappear because the `: hover` of the link is still white but there is not any background (same as dark-theme) for the search-result container. 
The simple style will resolve the problem

The image below show the problem, you will see a white space which is where I put the mouse on a link.

screenshot from 2017-07-12 17-45-34

### DIFF
--- a/doc/styles/main.css
+++ b/doc/styles/main.css
@@ -112,6 +112,11 @@ body.light-theme .search-result li.selected {
   color: #EC0C8E;
 }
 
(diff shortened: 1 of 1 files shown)


staltz:
Thanks @m98 for your first PR. This is tiny and seems safe to merge. Good job

Issue #2727, opened 2017-07-06
throttleTime(n, undefined, {leading: true, trailing: true}) ... double output


sod:
**RxJS version:** 6.5.1
**Code to reproduce (6.5.1):** https://stackblitz.com/edit/rxjs-tthiju
**Expected behavior:** One value initially, then defer next value for given `duration`
**Actual behavior:** One value initially, two values each `duration`
**Additional information:**

Test:

```javascript
import { interval } from 'rxjs'; 
import { throttleTime, take } from 'rxjs/operators';

interval(50).pipe(
  throttleTime(1234, undefined, {leading: true, trailing: true}),
  take(8),
).subscribe(x => console.log(x));

// actual output: 0----24-25----49-50----74-75----99|
// expected output: 0----24----49----74----99----124----149----174|
```

jscti:
Has anybody found a work around on this ?

tim-kuteev:
Hi **jscti**,

I think the work around for that might be to to give up on leading value.
```JavaScript
.throttleTime(1234, undefined, {leading: false, trailing: true})
```
Note that in this case you will be missing the first value. And there is also opened issue for single value #2859.

simeyla:
Would `distinctUntilChanged()` work after `throttleTime` - at least for some situations?

ubugnu:
> Would `distinctUntilChanged()` work after `throttleTime` - at least for some situations?

Yes it would work, for example:

```
distinctUntilChanged((prev, curr) => curr - prev < minWaitTime/1000)
```

sod:
Well, let me try to propose a fix again via https://github.com/ReactiveX/rxjs/pull/4864 :D

adolgoff:
Are there any plans to merge it?

sod:
Oh nice, thank you for fixing it ben 🥳

Issue #2466, opened 2017-03-15
Bug? Perhaps improper behavior of throttle(fn)


benlesh:
**RxJS version:** 5.0.0-alpha.* to 5.2.0

**Code to reproduce:**

```js
Observable.of(1, 2, 3).throttle(() => { throw new Error('oops'); })
  .subscribe(x => console.log(x), err => console.log(err.message));
```

**Expected behavior:**

logs:
```
1
"oops"
```

**Actual behavior:**

logs
```
"oops"
```

**Additional information:**

Apparently, the `durationSelector` function is called, and it's result is subscribed to _before_ the leading value is emitted. This seems unintuitive, as it's _that_ value that triggered the need to get the duration notifier from the `durationSelector` to begin with.

If we "fix" this, it is, indeed, a breaking change.

benlesh:
Closing as stale

PR #2749, opened 2017-07-14
fix(throttleTime): leading and trailing true should only emit a single value per time frame

sod:
Fixes https://github.com/ReactiveX/rxjs/issues/2727

### fix 1: `throttleTime` `{leading: true, trailing: true}`

Currently it behaves like
```typescript
'a1234b1234c1234d12e--|'
'a----b1----12----2e--|'
```

By not scheduling a new throttle when emitting a trailing value.

this PR changes it to:
```typescript
'a1234b1234c1234d12e--|'
'a----b----c----d----e|'
```

### fix 2: `throttleTime` `{leading: false, trailing: true}`

Currently it behaves like
```
'a-------------a-|'
'----------------|'
```

First and only value in each timeframe is dropped by not storing the first value when not yet throttled. So this code doesn't console.log anything:

```typescript
Rx.Observable.of(1)
    .throttleTime(100, undefined, {leading: false, trailing: true})
    .subscribe((value) => console.log(value))
```

This PR changes the behavior to:
```
'a-------------a-|'
'-----a-------------a|'
```

---
I added tests to cover those issues.

### DIFF
--- a/spec/operators/throttleTime-spec.ts
+++ b/spec/operators/throttleTime-spec.ts
@@ -49,99 +49,110 @@ describe('Observable.prototype.throttleTime', () => {
   it('should simply mirror the source if values are not emitted often enough', () => {
     const e1 =   hot('-a--------b-----c----|');
     const subs =     '^                    !';
+    const t =   time('-----|                ');
     const expected = '-a--------b-----c----|';
 
-    expectObservable(e1.throttleTime(50, rxTestScheduler)).toBe(expected);
+    expectObservable(e1.throttleTime(t, rxTestScheduler)).toBe(expected);
     expectSubscriptions(e1.subscriptions).toBe(subs);
   });
 
   it('should handle a busy producer emitting a regular repeating sequence with leading: true, trailing: false', () => {
     const e1 =   hot('a12345b12345c12345d123|');
     const subs =     '^                     !';
+    const t =   time('-----|                 ');
     const expected = 'a-----b-----c-----d---|';
 
-    expectObservable(e1.throttleTime(50, rxTestScheduler, {leading: true, trailing: false})).toBe(expected);
+    expectObservable(e1.throttleTime(t, rxTestScheduler, {leading: true, trailing: false})).toBe(expected);
     expectSubscriptions(e1.subscriptions).toBe(subs);
   });
 
   it('should handle a busy producer emitting a regular repeating sequence with leading: true, trailing: true', () => {
     const e1 =   hot('a1234b1234c1234d12e--|');
     const subs =     '^                    !';
+    const t =   time('-----|                ');
     const expected = 'a----b----c----d----e|';
 
-    expectObservable(e1.throttleTime(50, rxTestScheduler, {leading: true, trailing: true})).toBe(expected);
+    expectObservable(e1.throttleTime(t, rxTestScheduler, {leading: true, trailing: true})).toBe(expected);
     expectSubscriptions(e1.subscriptions).toBe(subs);
   });
 
   it('should not drop values if values are far apart with leading: true, trailing: true', () => {
     const e1 =   hot('ab-----------c|');
     const subs =     '^             !';
+    const t =   time('-----|         ');
     const expected = 'a----b-------c|';
 
-    expectObservable(e1.throttleTime(50, rxTestScheduler, {leading: true, trailing: true})).toBe(expected);
+    expectObservable(e1.throttleTime(t, rxTestScheduler, {leading: true, trailing: true})).toBe(expected);
     expectSubscriptions(e1.subscriptions).toBe(subs);
   });
 
   it('should handle a busy producer emitting a regular repeating sequence with leading: false, trailing: true', () => {
     const e1 =   hot('12345a1234b1234c12d--|');
     const subs =     '^                    !';
+    const t =   time('-----|                ');
     const expected = '-----a----b----c----d|';
 
-    expectObservable(e1.throttleTime(50, rxTestScheduler, {leading: false, trailing: true})).toBe(expected);
+    expectObservable(e1.throttleTime(t, rxTestScheduler, {leading: false, trailing: true})).toBe(expected);
     expectSubscriptions(e1.subscriptions).toBe(subs);
   });
 
   it('should only emit trailing values with leading: false, trailing: true', () => {
     const e1 =   hot('ab-----------c--d--|');
     const subs =     '^                  !';
+    const t =   time('-----|              ');
     const expected = '-----b------------d|';
 
-    expectObservable(e1.throttleTime(50, rxTestScheduler, {leading: false, trailing: true})).toBe(expected);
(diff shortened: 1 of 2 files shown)


sod:
@benlesh it's a followup to https://github.com/ReactiveX/rxjs/pull/2465 ... what do you think? :)

Parakleta:
I've been testing this, and the only hiccup is that the throttle call should come before the `.next()` emission because otherwise if the `.next()` is resolved synchronously it can really mess up the timing.

I suggest changing `_next()` to have something like:
```js
let emit = false;
if (!this.throttled && this.leading) {
    emit = true;
} else ...

if (!this.throttled) this.throttle();
if (emit) this.destination.next(value);
```

And also `clearThrottle` to something like:
```js
if (this.trailing && this._hasTrailingValue) {
    let value = this._trailingValue;
    this._trailingValue = null;
    this._hasTrailingValue = false;
    this.throttle();
    this.destination.next(value);
}
```

I have also recommended that this PR be replicated to solve #2466 but considering what I have discovered regarding the timing this may not be so simple.

sod:
How did you test the hiccup? `next` is not being called when a throttle is present, so I can't see how that happens.

```typescript
if (!this.throttled && this.leading) {
    this.destination.next(value);
} /* ...
```

Parakleta:
Later in the same function is the call to `this.throttle()`.  If `next()` takes 20ms to complete and our throttle is supposed to be 100ms then it will actually be 120ms between emissions because the setting of throttle doesn't start until after next completes.

sod:
Thanks, now I got it. I think just adding a second `this.throttle()` should be sufficient

```typescript
if (!this.throttled && this.leading) {
    this.throttle(); // <---
    this.destination.next(value);
} /* ...

**Comment by sod** [comment]
Thanks, now I understood the issue.

**Comment by sod** [comment]
rebased & added your feedback

File: src/operators/throttleTime.ts [hunk 1]
```diff
@@ -85,30 +85,38 @@ class ThrottleTimeSubscriber<T> extends Subscriber<T> {
...
+    if (!this.throttled && this.leading) {
+      this.throttle();
+      this.destination.next(value);
+    } else if (this.trailing) {
+      this._trailingValue = value;
+      this._hasTrailingValue = true;
+    }
+
+    if (!this.throttled) {
+      this.throttle();
```
benlesh: Can we refactor this a little to be:


```ts
if (!this.trottled) {
  this.throttle();
  if (this.leading) {
    this.destination.next(value);
  }
}

if (this.trailing) {
  this._trailingValue = value;
  this._hasTrailingValue = true;
}
```

I'm not sure in it's current configuration it will work as desired, because of that `else if`. If it's trailing, we always want to record the value, not only if we've already started throttling.

src/operators/throttleTime.ts
```diff
@@ -85,30 +85,38 @@ class ThrottleTimeSubscriber<T> extends Subscriber<T> {
...
+      throttled.unsubscribe();
+      this.remove(throttled);
+      this.throttled = null;
+
       if (this.trailing && this._hasTrailingValue) {
-        this.destination.next(this._trailingValue);
+        const trailingValue = this._trailingValue;
         this._trailingValue = null;
         this._hasTrailingValue = false;
+        this.throttle();
```
benlesh: this is unnecessary and changes the behavior of the operator. It's not supposed to start throttling again when the throttle is cleared.

src/operators/throttleTime.ts
```diff
@@ -85,30 +85,38 @@ class ThrottleTimeSubscriber<T> extends Subscriber<T> {
...
+  private throttle(): void {
+    this.add(this.throttled = this.scheduler.schedule(dispatchNext, this.duration, { subscriber: this }));
+  }
+
   clearThrottle() {
     const throttled = this.throttled;
+
     if (throttled) {
+      throttled.unsubscribe();
+      this.remove(throttled);
```
benlesh: this is unnecessary, the `unsubscribe()` will remove the inner subscription.

spec/operators/throttleTime-spec.ts
```diff
@@ -55,12 +55,48 @@ describe('Observable.prototype.throttleTime', () => {
...
-    const e1 =   hot('abcdefabcdefabcdefabcdefa|');
-    const subs =     '^                        !';
-    const expected = 'a-----a-----a-----a-----a|';
+  it('should handle a busy producer emitting a regular repeating sequence with leading: true, trailing: false', () => {
+    const e1 =   hot('a12345b12345c12345d123|');
+    const subs =     '^                     !';
+    const expected = 'a-----b-----c-----d---|';
 
-    expectObservable(e1.throttleTime(50, rxTestScheduler)).toBe(expected);
+    expectObservable(e1.throttleTime(50, rxTestScheduler, {leading: true, trailing: false})).toBe(expected);
```
benlesh: if you could use the `time` function here and represent the `time` as marbles above like:

```
const t1 = '----|';
```

and then use `time(t1)` which returns a number.

That would be ideal as the test could be more readable, you can even used commented out `----|` marblse to demonstrate where the throttles occur.

Remember, the whole purpose of these tests is to be more readable.

benlesh: Please do the same for the other tests.

src/operators/throttleTime.ts
sod: Then `throttleTime` would emit the same input twice on `{leading:true, trailing: true}`. I guess in the end it doesn't matter when throttleTime is used in conjunction with `distinctUntilChanged`, but it feels wrong regardless.

```typescript
Observable.of(1).throttleTime(100, {leading:true, trailing: true}).subscribe((n) => console.log(n));

// 1
// 1
```

vs. lodash
```typescript
fn = _.throttle((n) => console.log(n), 100, {leading:true, trailing: true});
fn(1);

// 1
```

src/operators/throttleTime.ts
sod: But this is exactly why I opened the PR. Because currently the operator emits 2 values each timeframe instead of 1.

If the doubleemit is intended, this PR can be closed and https://github.com/ReactiveX/rxjs/issues/2727 & https://github.com/ReactiveX/rxjs/issues/2859 rejected.

benlesh: hmm.. I think there's a disconnect here. `throttle()` actually starts the throttle measurement. The, with `leading` and `trailing` both `true`, it should emit the same as it would with `leading` behavior, but then emit the trailing value at the end.

benlesh: Oh, I okay, I actually see what you're saying now. This is to start the throttle again as soon as there's another emission (due to trailing behavior)

benlesh: Okay... the more I look at this, the more I think it will have to be a change that lands in v 6.0, because it will be a breaking change for someone. (Albeit a fix)

benlesh:
@sod can you look at what I've done with `throttle` in #2864 and try to mirror it as closely as possible to keep the code base consistent?

Issue #2754, opened 2017-07-18
WebSocketSubjectConfig's resultSelector is excepting an not necessary generic type T


fangpenlin:
**RxJS version:**

5.4.2

**Code to reproduce:**

Say if you want to make WebSocket subject emits raw MessageEvent, and you write

```typescript
import * as Rx from 'rxjs/Rx'

Rx.Observable.webSocket<MessageEvent>({
  url: 'ws://localhost:8080',
  resultSelector: (e: MessageEvent) => e
})
```

**Expected behavior:**

It should compiles

**Actual behavior:**

```
[ts]
Argument of type '{ url: string; resultSelector: (e: MessageEvent) => MessageEvent; }' is not assignable to parameter of type 'string | WebSocketSubjectConfig'.
  Type '{ url: string; resultSelector: (e: MessageEvent) => MessageEvent; }' is not assignable to type 'WebSocketSubjectConfig'.
    Types of property 'resultSelector' are incompatible.
      Type '(e: MessageEvent) => MessageEvent' is not assignable to type '<T>(e: MessageEvent) => T'.
        Type 'MessageEvent' is not assignable to type 'T'.
(property) resultSelector: (e: MessageEvent) => MessageEvent
```

**Additional information:**

I just upgraded to typescript 2.4.1, maybe it has something to do with [this change](https://github.com/Microsoft/TypeScript/pull/16072)

I can work around the issue either by returning `any` type from the `resultSelector` function, or make it a generic function like this

```typescript
import * as Rx from 'rxjs/Rx'

function resultSelector<T> (e: MessageEvent): T {
  return e as any as T
}

Rx.Observable.webSocket<MessageEvent>({
  url: 'ws://localhost:8080',
  resultSelector
})
```

But returning any is losing the type check, as we know exactly what type to be expected as it's given as an `Rx.Observable.webSocket` type argument. And for the second working around, which is even more awkward .

So, maybe a more ideal solution is to make `WebSocketSubjectConfig` also a generic type, and use the T as the return value from resultSelector?

kwonoj:
Yes, it seems legit and this seems oversight when we didn't have strict generic checks.

fangpenlin:
@kwonoj created a bugfix at #2755

jrmcdona:
I am still facing issues with this.

Here is the code for the WebSocketSubjectConfig.
```
   this.wsSubjectConfig = {
      url: url,
      closeObserver: {
        next: (e: CloseEvent) => {
[...]

PR #2755, opened 2017-07-18
fix: Make WebSocketSubjectConfig a generic type

fangpenlin:
**Description:**

Fix typescript type issue for websocket #2754, make `WebSocketSubjectConfig` a generic type, so that it will respect the type T from websocket.

In this way, the type can be inferred correctly from the resultSelector's result type, you can write something like

```typescript
const myWebSocket = webSocket({
  url: 'ws://localhost:8080',
  resultSelector: (e: MessageEvent) => e.data + '!'
})
```

the `myWebSocket` will automatically inferred as `WebSocketSubject<string>`

### DIFF
--- a/src/observable/dom/WebSocketSubject.ts
+++ b/src/observable/dom/WebSocketSubject.ts
@@ -10,10 +10,10 @@ import { tryCatch } from '../../util/tryCatch';
 import { errorObject } from '../../util/errorObject';
 import { assign } from '../../util/assign';
 
-export interface WebSocketSubjectConfig {
+export interface WebSocketSubjectConfig<T> {
   url: string;
   protocol?: string | Array<string>;
-  resultSelector?: <T>(e: MessageEvent) => T;
+  resultSelector?: (e: MessageEvent) => T;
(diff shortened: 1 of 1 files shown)


fangpenlin:
Hmmm, not sure what about default type

```typescript
const myWebSocket = webSocket({
  url: 'ws://localhost:8080'
})
```

it seems by default the T will get inferred to be `{}` strangely in this case. Maybe we should make the `T` default to `any`, as the default `resultSelector` is implementing as a JSON parser?

kwonoj:
I guess this should target `next` branch with default generic features in TS compiler.

fangpenlin:
@kwonoj 

sorry, maybe we should hold on a little bit on this PR, just tried, even with default generic type

```
export class WebSocketSubject<T = any> extends AnonymousSubject<T> {
```

Somehow we are still getting `{}` as the inferred type if `resultSelector` is not given :/

I am not sure how TypeScript works in terms of getting `{}` out of undefined `resultSelector`, that's really odd behavior

benlesh:
This is done in master.

PR #2758, opened 2017-07-18
feat(Subscription): allow subscription.add to accept promise of teardownlogic

joostverdoorn:
Currently, the following code is not allowed:

```typescript
const observable = new Observable(async subscriber => {
  const value = await getAsyncValue();
  subscriber.next(value);
});
```

and it will yield the following error:
```
Argument of type '(this: Observable<{}>, observer: Subscriber<{}>) => Promise<void>' is not assignable to parameter of type '(this: Observable<{}>, subscriber: Subscriber<{}>) => TeardownLogic'. Type 'Promise<void>' is not assignable to type 'TeardownLogic'. Type 'Promise<void>' is not assignable to type 'AnonymousSubscription'. Property 'unsubscribe' is missing in type 'Promise<void>'.  
```

This is because subscriptions don't allow teardownlogic to be provided through a promise.  This PR adds that functionality to subscriptions.

### DIFF
--- a/spec/Subscription-spec.ts
+++ b/spec/Subscription-spec.ts
@@ -112,6 +112,26 @@ describe('Subscription', () => {
       expect(isCalled).to.equal(true);
     });
 
+    it('Should return a new Subscription created with teardown function if it is passed a promise', (done: MochaDone) => {
+      const sub = new Subscription();
+
+      let resolvePromise;
+      let isCalled = false;
+
+      const promise = new Promise(resolve => {
+        resolvePromise = () => resolve(() => isCalled = true);
+      });
+      const ret = sub.add(promise);
+
+      promise.then(() => {
+        ret.unsubscribe();
+      }).then(() => {
+        expect(isCalled).to.equal(true);
+      }).then(() => done(), err => done(err));
+
+      resolvePromise();
+    });
+
     it('Should wrap the AnonymousSubscription and return a subscription that unsubscribes and removes it when unsubbed', () => {
       const sub: any = new Subscription();
       let called = false;
--- a/src/Subscription.ts
+++ b/src/Subscription.ts
@@ -1,6 +1,7 @@
 import { isArray } from './util/isArray';
 import { isObject } from './util/isObject';
 import { isFunction } from './util/isFunction';
+import { isPromise } from './util/isPromise';
 import { tryCatch } from './util/tryCatch';
 import { errorObject } from './util/errorObject';
 import { UnsubscriptionError } from './util/UnsubscriptionError';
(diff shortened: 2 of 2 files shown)


kwonoj:
I'm not sure if we want to allow asynchronous teardown logics to be added in subscription. @benlesh thoughts?

kwonoj:
I haven't verified so it's just gut feeling but this could possibly bring weird execution order.

david-driscoll:
I have the same feeling as @kwonoj it seems like this could cause some weird race conditions.  The biggest one would be what happens if you unsubscribe from the observable, while the promise is still in flight?  Since we can't cancel the promise, then we may see any weird number of things popup.

Is there a specific reason that you can't just do something like...

```typescript
Observable.defer(() => Observable.from(getAsyncValue()));
```

I'm just trying to find a compelling reason that we must introduce promises into the constructor in this way.

benlesh:
This is actually interesting, I'd like to read @jhusain's opinion on it.

The major difficulty I see is we'd need to return a `Promise<Subscription|Function|void>` from that internal function... however, it *should* be simple enough to wrap that in a `Subscription` synchronously.

I see this as totally doable, but the real question is whether or not we need to even have this if we make `Observable.from()` accept `async function *` (AsyncIterators). That's work that has already been prototyped, and I think I might even have a PR sitting around for it.

Also technically this would work:

```ts
const source = new Observable(observer => {
  (async () => {
    const value = await getAsyncValue();
    observer.next(value);
    const value2 = await getAsyncValue();
    observer.next(value2);
    observer.complete();
  }());
});
```

If you needed some sort of cancellation (which promises don't natively support, you'd have to role your own thing), you could do this:

```ts
const source = new Observable(observer => {
  (async () => {
    const token = new SomeCancelationTokenThing();
    subscription.add(() => token.cancel());
    const value = await getAsyncValue(token);
    observer.next(value);
    const value2 = await getAsyncValue(token);
    observer.next(value2);
    observer.complete();
  }());
  return subscription;
});
```

benlesh:
Yeah... the more I look at this, I think it's completely doable. It's just a matter of whether or not it's a good idea... We need to explore the corner cases.

Just adding support for `Promise<TeardownLogic>` should indeed do it, you're right @joostverdoorn

benlesh:
This is currently being discussed over in the TC39 proposal repo: https://github.com/tc39/proposal-observable/issues/163

benlesh:
SInce there hasn't been much traction on this idea, and it's too different from both the WHATWG and TC39 proposal. I'm going to close it for now.

matthargett:
This is something we think we'll need so we can do deep cancellation in one of our libraries that ultimately relies on a fetch polyfill on some platforms. We believe we need this PR and GitHub/fetch#592 merged to get the improved performance we're shooting for.

jhampton:
I've been following several PR's against the `fetch` polyfill - [this one seems most promising](https://github.com/github/fetch/pull/592) - that is driving some of our cancellation work.  In speaking with @jayphelps yesterday, we have a good solution for cancellation if we hand back an `abort-able` in our subscription.  In reading through this (and the TC39 Observable you referenced above), I'm wondering: was it ever decided that a Subscription _must_ be returned synchronously @benlesh ?

Thanks for the hard work and time, ladies and gentlemen – I'm enjoying digging-in with Rx (and RxJS in particular).

PR #2690, opened 2017-06-22
fix(typings): fix typings for error prone operators like concatAll

david-driscoll:
**Description:**
fixes concatAll, combineAll, exhaust, mergeAll, switch, zipAll.  Also made a few improvements to
how hot and cold observables are typed.  A few compiler errors were fixed due to this change.

**Related issue (if exists):**
fixes #2658 #2493 #2551

PR #2760, opened 2017-07-20
Fix typings for mergeAll and concatAll

lukaselmer:
```ts
// this is how the typings work
const numbers: number[] = Observable.of([1, 2, 3]).mergeAll();
// this would be correct (I think)
const numbers: Observable<number> = Observable.of([1, 2, 3]).mergeAll();
```

I believe that it is fixed in the next major release #2690


benlesh:
cc @david-driscoll

benlesh:
I'm unable to verify that there is an issue here, at least in master with pipeable `mergeAll`. Sorry, @lukaselmer, I'm going to close this for now.

Issue #2769, opened 2017-07-26
Could Subscription.add() method accept multiple tear down functions at once?


martinsik:
**RxJS version:** 5.*

**Actual behavior:**

I often find myself adding multiple `Subscription`s into a single `Subscription` so I can later unsubscribe with a single `unsubscribe()` call (for example in Angular).

This means I need to do for example the following:

```
const subscription = new Subscription();

const sub1 = Observable...subscribe();
const sub2 = Observable...subscribe();
const sub3 = Observable...subscribe();

subscription.add(sub1);
subscription.add(sub2);
subscription.add(sub3);

...

subscription.unsubscribe();
```

**Desired behavior:**

It would be helpful to me if I could add all tear down functions into `subscription` at once. For example similarly to `Array.push()` function that accepts multiple items:

```
subscription.add(sub1, sub2, sub3);
```

kwonoj:
This was discussed before https://github.com/ReactiveX/rxjs/issues/1583 , without concrete conclusion at the moment. Personally I think current interface mostly works, either via chaining (`sub.add().add()....`) for simply using array `[].forEach(sub.add)` - but will leave this opened to see if there's other opinion.

martinsik:
Using `sub.add().add()` is probably enough.
I think `[].forEach(sub.add)` won't work because it'll loose `this` context.

larrifax:
I recently discovered that ```sub.add(another)``` doesn't return ```sub```, but ```another```. This means that chaining can have adverse effects if one of the chained subscriptions ends before the root subscription does:

```
const sub = new Rx.Subscription();

sub
  .add(Rx.Observable.interval(1000).do(x => console.log("Interval 1", x)).first().subscribe())
  .add(Rx.Observable.interval(1000).do(x => console.log("Interval 2", x)).subscribe());

setTimeout(() => sub.unsubscribe(), 5000);
```
In this case, "Interval 2" never logs to the console, as "Interval 1" effectively unsubscribes "Interval 2" when it ends due to the ```first()``` operator.

benlesh:
In RxJS 5.X  `sub.add().add()` does NOT work like you'd expect. Returning a subscription from `add()` is a byproduct of legacy behavior, and should be avoided.

benlesh:
To answer this OP issue:

I don't think we'll be adding an `addTo` method because:

1. There isn't a reason for it in the library.
[...]

PR #2770, opened 2017-07-26
feat(Subscription): allow Subscription add mutiple teardown logics

Brooooooklyn:
**Description:**

Allow Subscription add multiple teardown logics

resolve https://github.com/ReactiveX/rxjs/issues/2769

### DIFF
--- a/spec/Subscription-spec.ts
+++ b/spec/Subscription-spec.ts
@@ -179,7 +179,7 @@ describe('Subscription', () => {
       const main = new Subscription();
       main.unsubscribe();
 
-      let isCalled = [];
+      let isCalled: boolean[] = [];
 
       const subscriptions = range(5)
         .map((_, i) => new Subscription(() => {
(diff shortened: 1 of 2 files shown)


kwonoj:
I'm marking this as blocked to get conclusion of parent issue #2769.

benlesh:
I like the spirit of this PR, and I've actually worked something like this into the experimental branch. We should probably refactor it for v6.

cartant:
@benesh I've rebased this PR.

benlesh:
Thanks @cartant .... Unfortunately, @Brooooooklyn, looking at how this is implemented it incurs a breaking change that we can't do until the next version of RxJS (where I already have it implemented in the experimental branch).

The problem is that currently, `subscription.add(child) === child` will return `true`. But this seems to change that.  Also, how would `subscription.add(a, b, c); subscription.remove(b)` work? I think we can't do this until the next major version because ideally, `add` returns void, and the second scenario above will work.

I'm going to close this, but just know that we are indeed doing something like this for the next version.

PR #2771, opened 2017-07-27
refactor(filter): remove extra predicate setting

bafolts:
**Description:**
I was looking at the generated javascript today and saw that `this.predicate` was being set twice. The declaration of the `private predicate` should suffice to set the property to the argument. The additional setting is unnecessary.


kwonoj:
Yes, I think it remain while refactor to use ctor params.

PR #2777, opened 2017-07-30
ErrorObservable changed to subtype of Observable<never>

### LINKED ISSUE #2776: ErrorObservable should be subtype of Observable<never>
RxJS version: 5.4.2

Code to reproduce:

```ts
/**
 * Asserts (by resulting in an errored stream) that the given stream produces
 * only true values.
 */
export function assertValid(streamToTest: Observable<boolean>): Observable<never> {
    return streamToTest
        .flatMap(test => {
            return test
                ? Observable.empty()
                : Observable.throw("Invalid value in stream.");
        });
}
```

Expected behaviour:

The above code should compile without errors since neither `Observable.empty` nor `Observable.throw` actually produce values, they should both return `Observable<never>`.

Actual behaviour:

The above code produces a compiler error:

```
Type 'Observable<any>' is not assignable to type 'Observable<never>'.
```

This is because the return type of `Observable.throw` is `Observable<any>`.

**Note**: In RxJS version 5.4.2, `Observable.empty` also produces `Observable<any>`, but this was fixed in #2691.  See #2640 for further justification of using `never` for streams that produce no values.

101100:
**Description:**

`ErrorObservable` was changed to be a subtype of `Observable<never>`.  Also fixed several build issues on the latest TS compiler so I could run the tests.

Fixes #2776

### DIFF
--- a/src/observable/ErrorObservable.ts
+++ b/src/observable/ErrorObservable.ts
@@ -13,7 +13,7 @@ export interface DispatchArg {
  * @extends {Ignored}
  * @hide true
  */
-export class ErrorObservable extends Observable<any> {
+export class ErrorObservable extends Observable<never> {
 
   /**
    * Creates an Observable that emits no items to the Observer and immediately


jayphelps:
I think this is a duplicate of #2640, which was fixed in #2691, but since it is a breaking change it's being held off until the next major release, v6

101100:
Actually, this is an extension of #2691.  That fixes the issue for empty observables.  This fixes the issue for error observables (which also do not produce a value).  The line that needs to be changed is here:

https://github.com/ReactiveX/rxjs/blob/27576f749ecac85cc4238eead8b1e54673e1e440/src/observable/ErrorObservable.ts#L16

`ErrorObservable` should extends `Observable<never>`.

I saw that #2691 was done against `next`, so I attempted it, but at that time, `next` was failing.  If it has been repaired, I can rebase this fix on the latest version of `next`.

jayphelps:
@101100 ah got it. Before my comment, your PR included fixes for empty/never as well, so I missed the error part 😝  sneaky sneaky.

You'll want to target the `next` branch since this is a breaking change.

101100:
@jayphelps I was trying to fix the build because travis was failing, but I took that out so you could see the main fix.  I was trying to get Travis to build, but now I see that it distracted from the PR purpose.

jayphelps:
@101100 right on! Sorry for the confusion. I'm not a maintainer any more, just trying to help where I can. I'll see if I can get `next` back into a good state so this PR will pass.

101100:
This was done in #3288.

PR #2781, opened 2017-08-04
Lazy Error.stack usage

hermanbanken:


kwonoj:
We have actually changed our custom error (targeted `next` for now) to no longer try to mutate those directly (https://github.com/ReactiveX/rxjs/pull/2614/files#diff-cd4b33df1ab129e23dc2bc5c12d8369bR8) 

which probably doesn't require this changes. This is interesting approach, but we could probably just cherry-pick those changes only from `next` instead (as I believe it won't be breaking?)

hermanbanken:
Yes, those setPrototypeOf calls are fine as well! Good alternative. I don't think it is breaking: the same properties are exposed, the errors don't change. 

There is some other stuff in there too though. A careful cherry-pick, including only the non-nexty changes, would be fine. 

I can rebase this PR to such a cherry-pick on Monday.

hermanbanken:
Never mind. The proposed changes in [next](https://github.com/ReactiveX/rxjs/pull/2614/files#diff-cd4b33df1ab129e23dc2bc5c12d8369bR8) are wrong: `stack === undefined` (at least when running mocha).

I added a test for this in the last commit. I had no idea Errors where so compilated in JavaScript. See for example https://gist.github.com/justmoon/15511f92e5216fa2624b#anti-patterns.

kwonoj:
If changes in `next` doesn't work as expected, I'd rather try to make legit changes for `next` and backport it into master instead of having 2 separate changes (as long as it's not breaking changes).

hermanbanken:
Agreed... I'm writing some tests as we speak.

Unfortunately the PR #2614 is a little messed up because next is merged in back instead of rebasing. Therefore cherry-picking that cleanly is not easy. That complicates the backporting part...

benlesh:
I suddenly have more time to invest in RxJS.. because it's actually my job now... 

I think this is something we're interested in doing. It does need more review, but it also needs refactored since files moved around.

cartant:
@benlesh I've had a look at this PR with a view to getting it into a state in which it can be applied to v6. However, I'm not convinced that it's required. In fact, I think the current, v6 implementation should be preferred.

Using TypeScript 2.8.3, the following code:

```ts
class SomeError extends Error {
    private err: Error;
    constructor() {
        const err: any = super("some error");
        this.err = err;
    }
    get stack(): string | undefined {
        return this.err.stack;
    }
}
const s = new SomeError();
console.log(`has stack = ${Boolean(s.stack)}`);
console.log(`instanceof Error = ${s instanceof Error}`);
console.log(`instanceof SomeError = ${s instanceof SomeError}`);
```

outputs this:

```
SomeError has stack = true
instanceof Error = true
instanceof SomeError = false
```

Whereas this code:

```ts
class AnotherError extends Error {
    constructor() {
        super("another error");
        (Object as any).setPrototypeOf(this, AnotherError.prototype);
    }
}
const a = new AnotherError();
console.log(`has stack = ${Boolean(a.stack)}`);
console.log(`instanceof Error = ${a instanceof Error}`);
console.log(`instanceof AnotherError = ${a instanceof AnotherError}`);
```

outputs this:

```
AnotherError has stack = true
instanceof Error = true
instanceof AnotherError = true
```

Note that the implementation that uses this PR's mechanism returns `false` for `instanceof SomeError` and that both implementations provide the `stack`. For me, the current implementation works fine in the browser (Chrome Headless) and in Node (v8.11.2).

[Here's a playground link](https://agentcooper.github.io/typescript-play/?target=1#code/MYGwhgzhAEDKD2BbApgUQE7vu6yAeALsgHYAmMGWOA3gFDQPQAO6AlgG5hG6YBc0lbAG56jYPGIQC6AK7AC2ABQBKaHUYbo4yQR7p+YYgE9oAXmgQZTZOkUAiCEmR7sd5SM2MCAC1YQAdDY45kEejAC+ogwA5si6UmDAANYq-FJsxNHQAD7QMmTIAGasxMikalGa6HEy6MTQPn6BmP4JyWEMkZHaUhZm0KUA7nBOgrbutD3wIMj+IPDRigAGCChj0N6QFgSJSf0AJNQAQvDTyIaKAW1JyuFLE1MzcwvLJQnEwMjwhQKY2AfUGBvHYfL4-MZ3B4SRxPeaLJbAwyfb4jNZ-YLQQ5AnRIsGotDoyEiSbgKDQACCxHgPhs63wRDIFHRFQ0PWkcgU4xZngsVhs9kM1O8Nhc6DcHU0igA8gAjABWyHk0C2hiMylacQAClgFAQjNYpYVFI0IAAaClUmnoMb+FjU6n65ATDRdSbQ3RgfpDC1C2nolQiR6zOHLSm+63MzYwa4Ak5nC5gVo7ZK3e6B6FnZ7wxGglHrcyHT055Hgwlpt2STMhhE43M-MNW-OY6hF2sln2NsvuIA) if you want to try it out.

spec/util/EmptyError-spec.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/** @test {ObjectUnsubscribedError} */
+describe('EmptyError', () => {
+  it('should contain a stacktrace', () => {
+    const observable = Observable.empty();
+    try {
+      observable.first().subscribe();
+    } catch (err) {
+      expect(err instanceof EmptyError).to.equal(true);
+      expect(err.stack).to.not.be.undefined;
```
benlesh: `.to.be.a('string')` would be a more specific check here.

spec/operators/timeout-spec.ts
```diff
@@ -33,6 +33,7 @@ describe('Observable.prototype.timeout', () => {
       throw new Error('this should not next');
     }, err => {
       expect(err).to.be.an.instanceof(Rx.TimeoutError);
+      expect(err.stack).to.not.be.undefined;
```
benlesh: Same as the other spot. `.to.be.a('string')`

spec/observables/dom/ajax-spec.ts
```diff
@@ -262,6 +262,7 @@ describe('Observable.ajax', () => {
     });
 
     expect(error instanceof Rx.AjaxError).to.be.true;
+    expect(error.stack).to.not.be.undefined;
```
benlesh: Same  `.to.be.a('string')`

src/observable/FromObservable.ts
```diff
@@ -91,7 +91,7 @@ export class FromObservable<T> extends Observable<T> {
         return new FromObservable<T>(ish, scheduler);
       } else if (isArray(ish)) {
         return new ArrayObservable<T>(ish, scheduler);
-      } else if (isPromise(ish)) {
+      } else if (isPromise<T>(ish)) {
```
benlesh: This change seems unrelated.

src/operator/catch.ts
```diff
@@ -66,7 +66,7 @@ import { subscribeToResult } from '../util/subscribeToResult';
  */
 export function _catch<T, R>(this: Observable<T>, selector: (err: any, caught: Observable<T>) => ObservableInput<R>): Observable<T | R> {
   const operator = new CatchOperator(selector);
-  const caught = this.lift(operator);
+  const caught = this.lift<T>(operator);
```
benlesh: This change seems unrelated.

src/util/TimeoutError.ts
```diff
@@ -1,15 +1,21 @@
...
+ *
+ * @class TimeoutError
+ */
+export class TimeoutError extends Error {
+  private err: Error;
+
+  constructor() {
+    const err: any = super('Timeout has occurred');
+    this.err = err;
+    (<any> this).name = err.name = 'TimeoutError';
```
benlesh: `this as any` .. we're trying to move away from this style of casing and move to `as whatever`

cartant:
@benlesh I'm suggesting that this PR should be closed as the current v6 implementation is better behaved. Do you not agree?

hermanbanken:
Does the new implementation “touch” the stack? Eg access the getter/setter of the stack property? Because that was the actual problem I described. If it uses inheritance and does not access that field ahead of time then we should be fine without this PR and I would also suggest closing it.

cartant:
@hermanbanken I don't believe it does, I did some checking of the `__extends` function that's in the emitted TypeScript implementation and I don't believe it does touch the `stack` property. I can double check, but I'm close to certain that it does not.

Issue #2784, opened 2017-08-05
zip loses type information with rest parameters


cartant:
**RxJS version:** 5.4.2

**Code to reproduce:**

```js
const others = [
  Rx.Observable.of(1),
  Rx.Observable.of(2)
];
const zipped: Rx.Observable<Date> = Rx.Observable
  .of(0)
  .zip(...others);
```

**Expected behavior:**

A TypeScript error similar to the following ought to be effected:

```
Type 'Observable<number[]>' is not assignable to type 'Observable<Date>'.
  Type 'number[]' is not assignable to type 'Date'.
    Property 'toDateString' is missing in type 'number[]'.
```

**Actual behavior:**

There is no TypeScript error.

**Additional information:**

There appears to be a problem with the TypeScript signature for the instance `zip` method that matches the rest parameters usage.

If no `project` function is specified, `R` cannot be inferred and the result is typed as `Observable<{}>` - so the value's type information is lost, rendering any strong typing ineffectual and hiding problems.

The signature that it matches is this one:

```ts
export declare function zipProto<T, R>(
  this: Observable<T>,
  ...observables: Array<ObservableInput<T> | ((...values: Array<T>) => R)>
): Observable<R>;
```

And the issue can be fixed by adding another signature before it:

```ts
export declare function zipProto<T, R>(
  this: Observable<T>,
  ...observables: Array<ObservableInput<T>>
): Observable<T[]>;
```

I've found similar problems with `combineLatest` and `withLatestFrom`. (And there are some signatures in `forkJoin` that don't look right to me either.)

I'm happy to submit a PR with the fixes, but I'm not sure to what extent - if any - the type declarations are tested. Other than the changed signatures, is there anything I should be including in the PR?

kwonoj:
I don't think proposed signature is correct one. In short, by supplying array via rest operator, output should be type of `T` instead of `Array<T>`.

```js
const values = [1,2,3]
zip(values); //this signature is zip(value: Array<number>)
zip(...values): //this sigature is zip(v1: number, v2: number, v3: number)
```

so in your example above, return type of `zip` is `Observable<number>` and should not be `Observable<Array<number>>`. 

Of course still current behavior is bit incomplete to infer correct output type, but there are some limitations of inferring types in rest parameters in TSC compiler at this moment - unless we're manually handcraft hugh numbers of combination, we can't create precise signature and reason why we provide `R` as output type to be specified explicitly.

cartant:
[...]

PR #2787, opened 2017-08-07
chore(typings): improve zip typings

cartant:
Add signature to zipProto so it's consistent with zipStatic.

**Description:** Add a rest parameters signature that does not include a `project` function.

At the moment, `zipProto` is not consistent with `zipStatic`, as the only rest parameter signature accounts for a project function, and if it's not specified, the result is of type `Observable<{}>` - losing the type information.

**Related issue (if exists):** #2784

### DIFF
--- a/spec/operators/zip-spec.ts
+++ b/spec/operators/zip-spec.ts
@@ -2,6 +2,7 @@ import {expect} from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
 import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
 
+declare const type;
 declare const hot: typeof marbleTestingSignature.hot;
 declare const cold: typeof marbleTestingSignature.cold;
 declare const expectObservable: typeof marbleTestingSignature.expectObservable;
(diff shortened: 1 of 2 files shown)


spec/operators/zip-spec.ts
```diff
@@ -595,4 +596,34 @@ describe('Observable.prototype.zip', () => {
     expectSubscriptions(a.subscriptions).toBe(asubs);
     expectSubscriptions(b.subscriptions).toBe(bsubs);
   });
+
+  it('should support rest parameter observables', () => {
+    type(() => {
```
kwonoj: `type` is actually have signature of `type(title: string, test: fn)` so doesn't require wrap via `it`. I see current codebases have this practices, which need to be updated as well.

cartant:
I see what you mean. It took a bit of digging - as `type` isn't the easiest term to search for - but I eventually found [its implementation](https://github.com/ReactiveX/rxjs/blob/5.4.3/spec/helpers/testScheduler-ui.ts#L91-L100).

cartant:
@kwonoj @david-driscoll I understand that this PR addresses an issue that is not high-priority and that reviewing PRs takes up your valuable time, but I'm curious as to whether there is a particular reason for this PR not being merged. Primarily, because I've discovered a number of similar issues and I'm wondering whether/when to submit further PRs.

kwonoj:
@cartant It's just waiting second approval from @david-driscoll . Each contributor's are quite busy, sometimes PR could take bit long to get reviewed.

david-driscoll:
Apologies slipped through the cracks of my emails.

PR #2789, opened 2017-08-09
fix(symbol.iterator): function name typo issue

brutalimp:
fix(symbol.iterator): function name typo issue


kwonoj:
Ponyfill is intended word - https://github.com/benlesh/symbol-observable/blob/master/readme.md

PR #2790, opened 2017-08-09
feat(AjaxError): Include the response on instances of AjaxError

ncphillips:
A non-2xx HTTP response may include a body to provide context to the error. Provide the ability to
retrieve the response from the AjaxError as you would from an AjaxResponse.


benlesh:
@ncphillips ... thank you for this PR, sorry it took me so long to review. Looks good. Merging this in before 5.5 (to be released soon)

Issue #2446, opened 2017-03-07
Rxjs not working in firefox webextensions


fouadbhatti:
**RxJS version:**
5.2.0 / Firefox version: 52
rxjs@5.2.0/bundles/Rx.min.js
**Code to reproduce:**
const subject = new Rx.Subject();
**Expected behavior:**
Rxjs should work
**Actual behavior:**
Getting the following error on loading rxjs
```XrayWrapper denied access to property "random" (reason: object is not safely Xrayable). See https://developer.mozilla.org/en-US/docs/Xray_vision for more information. Note that only the first denied property access from a given global object will be reported. Rx.min.js:214:186```

Firefox is throwing error a ```TypeError: a.Math.random is not a function  Rx.min.js:214:186```. Also when tried to create a subject throws this error ```TypeError: Rx.Subject is not a constructor```
**Additional information:**
Working perfectly fine in safari and chrome extensions. Only in firefox webextensions it throws the above mentioned errors and fails.

benlesh:
@fouadbhatti do you have the minimal code to reproduce the error? It's unlikely we'll be able to look into this without a "working" example that shows the behavior you're seeing.

fouadbhatti:
@benlesh Sure. Here is the minimal code. This is a basic firefox webextension in which i have added rxjs.

**RxJS Version**
5.2.0
rxjs@5.2.0/bundles/Rx.min.js

**Firefox Version**
Firefox version: 52

**How to install**
1. Open firefox and go to about:debugging, click on Load Temporary Addon and open the boderify folder and then select manifest.json and open. 

2. Now extension will load. Just click on debug for boderify on the about:debugging page. Accept the incoming request dialog. Keep it open, please don't close it for now.

3. Open any page now in firefox, and then go back to debugger window. You will see these errors throw.
` 
XrayWrapper denied access to property "random" (reason: object is not safely Xrayable). See https://developer.mozilla.org/en-US/docs/Xray_vision for more information. Note that only the first denied property access from a given global object will be reported.  Rx.min.js:214:167
TypeError: a.Math.random is not a function  Rx.min.js:214:186
TypeError: Rx.subject is not a function  borderify.js:6:15
TypeError: a.Math.random is not a function  Rx.min.js:214:186
TypeError: Rx.subject is not a function
`

**Code to reproduce**
borderify.js
`
document.body.style.border = "5px solid red";
var subject = Rx.subject();
console.log(subject);
`

**Attatched webextension**
[...]

PR #2791, opened 2017-08-09
fix(Immediate): Use global Math

rothfels-test:
**Description:**

This library does not work properly inside of a Firefox addon. See related issue.

https://github.com/ReactiveX/rxjs/issues/2446

### DIFF
--- a/src/util/Immediate.ts
+++ b/src/util/Immediate.ts
@@ -129,7 +129,7 @@ export class ImmediateDefinition {
     // * http://www.whatwg.org/specs/web-apps/current-work/multipage/comms.html#crossDocumentMessages
     const root = this.root;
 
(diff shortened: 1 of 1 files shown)


kwonoj:
And this change actually makes one of our test case fails, cause by not looking correct global context resolved via `root` module. I'm feeling it is quite exceptional cases some environment can resolve its own global but global function is not defined in there - rather ask code runs under such environment fulfills object correctly. But that's my 5 cent, I'll leave this opened bit to see other opinion. /cc @benlesh for review.

benlesh:
We can probably just change the test to temporarily monkey-patch Math.random.

benlesh:
Immediate is no longer a thing.

PR #2794, opened 2017-08-12
Added RxJs Logo to README, Similar to RxSwift README

lordKnighton:
**Description:**


kwonoj:
I think we are using `RxJS` elsewhere? (not sure if it's legit though)

lordKnighton:
So can we get this Merged?

kwonoj:
@lordKnighton I think it need to be updated with updating commit message (for changelog) / updating RxJs word, otherwise lgtm. I'm adding @benlesh for double confirm just in case.

PR #2796, opened 2017-08-14
fix(subscribeToResult): throw error in subscriber with inner observable

### LINKED ISSUE #2618: subscribe onError swallows any exception thrown
I have an Angular 4 application that use rxs.

The error handler when added to the subscribe is swallowing an exception if one is thrown from within the handler. In version 5.0.1, exceptions thrown from the error handler bubbled out to Angular's global error handler.

It appears this was broken with version 5.1, If I change the version in my plunker to 5.1, it behaves as it does in the latest version.

```js
.subscribe(
  result => {
    //Throwing an error here gets picked up by the global error handler
    throw new Error('break in on next');
    this.result = result.artists.items; },
  err => { 
    console.log('subscribe on error in on next');
    //This error does bubble out to the global angular error handler,
    // since it was thrown from the onNext
    throw new Error('Error in on next here'); 
  },
  () => { console.log('Completed'); }
);
```

I posted the following on stack overflow.
https://stackoverflow.com/questions/44187493/rxjs-angular-4-throw-error-in-subscribe-onerror-not-bubbling-out-to-custom-glo

**RxJS version:** 5.0.1

**Code to reproduce:**

where it works: 5.0.1 - https://plnkr.co/edit/ZMhTNbBfWaz4Y0TiI0jR?p=preview

where it is broken: latest version of rxjs - https://plnkr.co/edit/BI4njaMbtR7pfplQvgru?p=preview

**Expected behavior:**
Exception thrown from the onError of a subscription should be thrown out. In my example it would be caught by Angular's global error handler.

**Actual behavior:**
Exception is swallowed by the subscription and not thrown out.

kwonoj:
**Description:**
#2626 fixed error to be thrown when observable have sources, but it doesn't cover cases of inner observable with `subscribeToResult` as it doesn't have sources. This PR explicitly sets subscriber's `syncErrorThrowable` for inner observables, let exception in subscribers correctly thrown with inner observables as well.

- closes #2618

### DIFF
--- a/spec/util/subscribeToResult-spec.ts
+++ b/spec/util/subscribeToResult-spec.ts
@@ -1,9 +1,11 @@
-import {expect} from 'chai';
+import { expect } from 'chai';
 import * as Rx from '../../dist/cjs/Rx';
-import {subscribeToResult} from '../../dist/cjs/util/subscribeToResult';
-import {OuterSubscriber} from '../../dist/cjs/OuterSubscriber';
-import {$$iterator} from '../../dist/cjs/symbol/iterator';
+import { subscribeToResult } from '../../dist/cjs/util/subscribeToResult';
+import { OuterSubscriber } from '../../dist/cjs/OuterSubscriber';
+import { $$iterator } from '../../dist/cjs/symbol/iterator';
 import $$symbolObservable from 'symbol-observable';
+import { Observable } from '../../dist/cjs/Observable';
+import { Subject } from '../../dist/cjs/Subject';
 
 describe('subscribeToResult', () => {
   it('should synchronously complete when subscribe to scalarObservable', () => {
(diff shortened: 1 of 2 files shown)


kwonoj:
I believe this is patch as it fixes regressions.

jayphelps:
Do you have cycles to add a test case? Dunno if this is truly minimum but it reproduces it:

```js
const input$ = new Rx.Subject();

input$
  .mergeMap(() => Rx.Observable.of(1, 2))
  .subscribe(d => {
    throw new Error('some error');
  });

input$.next();
```

kwonoj:
@jayphelps I've used given snippet to verify behavior. Have to think a bit for test cases.

kwonoj:
@jayphelps test case updated.

PR #2804, opened 2017-08-21
refactor(package): change distribution to support main, module, es201…

jasonaden:
…5 keys in package.json

The structure created by the `build_all` in this PR is different from the current one. There's now directories by language/package format. Everything gets built into `dist` as it already did, but then instead of publishing the `cjs` folder, we will now publish the `package` folder.

In here you will find folders for `cjs`, `esm5`, and `esm15`. The `esmX` folders represent `ECMAScript Modules` and the language level (`es5` or `es2015`). `cjs` is exactly as it used to be (`CommonJS` format in `es5` language level).

There is also a `global` folder (same as before) holding the `UMD` bundles. And a `src` directory with the TypeScript sources.

In order to support deep imports in all language levels, every file in the package has it's own `package.json` file that's dynamically generated in the `.make-packages.js` file.


kwonoj:
We do have scoped package https://www.npmjs.com/package/@reactivex/rxjs for shipping all packges at once, while it currently doesn't have native ES module packages yet. (We did once and stopped, will do again with refined codes)

I don't think we'll going to change pkg publishing in `rxjs`. It was intentionally created for CJS only and it'll be.

jasonaden:
@kwonoj So the point on this PR is to distribute different formats, while keeping support for deep imports. This way tools like Webpack will pick up the ESM files and will produce significantly smaller application bundles. CJS format is great for Node obviously, but Webpack and Rollup have a hard time optimizing it, therefore this format will allow both to co-exist.

I probably should have given a bit more background. I've been working with @benlesh to come up with some way to distribute Rx where we don't break existing applications (deep imports), but also supports optimizations through ESM format. This seems to be the solution that works.

kwonoj:
thx for explanation. I'll leave this to @benlesh for primary review and will try to have some more context in core metings.

package.json
```diff
@@ -21,23 +21,26 @@
...
+    "build_cjs": "Build CJS package with clean up existing build",
+    "build_esm5": "Build ESM/ES5 package with clean up existing build",
+    "build_esm15": "Build ESM/ES2015 package with clean up existing build",
     "build_closure_core": "Minify Global core build using closure compiler",
     "build_global": "Build Global package, then minify build",
     "build_perf": "Build CJS & Global build, run macro performance test",
     "build_test": "Build CJS package & test spec, execute mocha test runner",
     "build_cover": "Run lint to current code, build CJS & test spec, execute test coverage",
-    "build_docs": "Build ES6 & global package, create documentation using it",
+    "build_docs": "Build ESM15 & global package, create documentation using it",
```
IgorMinar: should we just use "2015" instead of "15" throughout?

jasonaden: @benlesh We switched to `esm2015` on the Angular project. Would you like to use the same?

benlesh:
@kwonoj I've worked with @jasonaden to try to get this PR right. The outcome is the best we can hope for with current technology stacks/build systems, I think. People should experience some decent code size reductions if they're using Rollup or Webpack properly with this new way of distributing the code.

Merging this (dispite the dangerjs complaints) ahead of the 5.5 release.

kwonoj:
@benlesh yeah I'm fine, 👍

PR #2807, opened 2017-08-23
feat: `Observable.fromDOMObserver`: adapts MutationObserver and friends

wbinnssmith:
This adapts the following interface, which is becoming more and more popular in DOM apis nowadays, to RxJS Observables:

```js
type RecordCallback = (record: any, ...rest: any[]) => void;
interface DOMObserver {
  new (callback: RecordCallback, ...rest: any[]): DOMObserver;
  observe(...rest: any[]): void;
  disconnect(): void;
}
```

In practice, this means that it can adapt `MutationObserver`, `IntersectionObserver`, `ResizeObserver`, and `PerformanceObserver` to RxJS Observables.

While the DOMObserver yields lists of records (or an object that produces such a list), the returned RxJS Observable `next()`s each individual record at a time.

For example, here's a use of retrieving an observable for `MutationObserver`:

```js
var mutations = Rx.DOM.fromDOMObserver(
  MutationObserver,
  document.getElementById('foo'),
  { attributes: true, childList: true, characterData: true }
);

mutations.subscribe(record => console.log(record));
```

This `next()`s, and therefore logs, each individual [MutationRecord](https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord) on the returned Observable.

If there's a trivial way to create more integration-style tests with real examples from the dom, I'd be happy to provide these to add confidence.

This particular dom api appears to have no way of signaling error or completion (much like `addEventListener`), so there is no equivalent in the returned Observable.

### DIFF
--- a/spec/observables/dom/fromDOMObserver-spec.ts
+++ b/spec/observables/dom/fromDOMObserver-spec.ts
@@ -1,0 +1,116 @@
+import {expect} from 'chai';
+import * as Rx from '../../../dist/cjs/Rx';
+import marbleTestingSignature = require('../../helpers/marble-testing'); // tslint:disable-line:no-require-imports
+
+declare const { asDiagram };
+declare const expectObservable: typeof marbleTestingSignature.expectObservable;
+declare const rxTestScheduler: Rx.TestScheduler;
+
+const Observable = Rx.Observable;
+
+type RecordCallback = (record: any, ...rest: any[]) => void;
+
+/** @test {fromDOMObserver} */
+describe('Observable.fromDOMObserver', () => {
+  let observerInstance;
+  class MockDOMObserver {
+    protected _args: any[];
+    protected _connected: boolean;
+
+    constructor(protected _callback: RecordCallback) {
+      observerInstance = this;
+    }
+    observe(...args: any[]) {
+      this._args = args;
+      this._connected = true;
+    }
+    disconnect() {
+      this._connected = false;
+    }
+  }
+
+  beforeEach(() => {
+    observerInstance = null;
+  });
+
+  it('does not construct a DOM Observer until the Rx Observable is subscribed to', () => {
+    const o = Observable.fromDOMObserver(MockDOMObserver, 'some', 'additional', {args: true});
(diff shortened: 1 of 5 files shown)


kwonoj:
Appreciate for effort, but we do recommend to create custom operator or others as separate module instead (https://github.com/ReactiveX/rxjs/blob/master/doc/operator-creation.md#publish-your-operator-as-a-separate-library) and does not accept direct merge into core library.

benlesh:
This isn't really an operator, it's a creation method. I think it's open for debate. I think the sticky bit here is the DOM dependency.

kwonoj:
For me category of feature (either if it's static creation or it's operator) doesn't changes crux of our recommendation to make user level. Primary reason we recommend to start off user level module is to prevent api surface increase, also try to find out key-core-set of features only in core. While this is static creation, still implementation itself does not require deep internals of core library that can be seperated into - I'd vote those approach first.

wbinnssmith:
Apologies for not inquiring first before the PR. I figured if there's a creation method for WebSockets that there might be room for this 😛

We're using this style of Observers very heavily inside of Nuclide and Atom, though I'm not sure this represents the majority of RxJS.

What is the intended direction of the `dom/` subtree in the source? Will environment-specific adapters continue to live in this repo, and are they all bundled by default for browser builds?

The `RxJS-DOM` package isn't yet compatible with v5 but nicely stands own its own as a set of maintained contributions that make interfacing with the browser environment straightforward. There's a similar situation with `rx-node`. I assume these are basically irrelevant to RxJS 5.

To me it makes sense that modern tooling can allow users to pick and choose individual operators and creation methods without bundling every possible combination into the core of Rx.

Thanks for considering!

kwonoj:
as far as I remember we wanted to split dom into separate module, but I think it's just had not progressed yet.

benlesh:
I think for now, unless we see a lot of demand for it, we can leave this in the user-land. If it gets to the point where this feature is popular enough that there are competing impls or the like, we can pull one in as a standard (and make yet another competing impl! haha).

PR #2819, opened 2017-08-30
fix(partition): update signature to match docs and filter operator

manbearwiz:
**Description:**

This PR adds the index parameter to the predicate signature for the partition operator. This change matches the docs as well as the existing `filter` operator.

### DIFF
--- a/spec/operators/partition-spec.ts
+++ b/spec/operators/partition-spec.ts
@@ -45,6 +45,20 @@ describe('Observable.prototype.partition', () => {
     expectSubscriptions(e1.subscriptions).toBe([e1subs, e1subs]);
   });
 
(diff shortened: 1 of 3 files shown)


kwonoj:
Interesting, it seems like we had signature to support index but removed it for some reason, not sure why. 

Still, it already makes current public api surface as-is, means if we make `index` as non-optional param, it'll be breaking changes for next major instead of minor. (`next` branch.) For me it seems more feasible to make index to optional.

benlesh:
I'm not sure this is a breaking change, @kwonoj.`(x: T) => R` is compatable with `(x: T, i: number) => R` as an argument.

kwonoj:
yeah, I think I misread signatures.

PR #2820, opened 2017-08-30
add test for multiple unsubscribes

manbearwiz:
**Description:**

This PR adds a simple test to cover the case where the `unsubscribe` method is called multiple times. According to coveralls this was currently untested so we should see a tiny increase in coverage.

### DIFF
--- a/spec/subjects/ReplaySubject-spec.ts
+++ b/spec/subjects/ReplaySubject-spec.ts
@@ -100,6 +100,22 @@ describe('ReplaySubject', () => {
     });
   });
 
+  it('should gracefully handle multiple calls to unsubscribe', () => {
+    const subject = new ReplaySubject(1);
+
+    subject.next(1);
+
+    const subscription = subject.subscribe((x: number) => {
+      expect(x).to.equal(1);
+    });
+
+    subject.next(1);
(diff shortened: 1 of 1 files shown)


kwonoj:
seems coverage hasn't changed with PR?

manbearwiz:
Ya I will have to take another look at this

benlesh:
@manbearwiz ... unrelated to anything, one of my best friends IRL always has the online nickname of something about manbears or pandas, but he's not on GitHub, and I was really excited like "OMG, he's contributing!" haha

So... there's more manbears in the world than you'd think I guess.

manbearwiz:
Ha, @benlesh I was given this nickname as an modification of South Park's manbearpig so I'm sure there are a few.

On the topic of this issue though, I must have been looking at the wrong method when I stepped through. I'm pretty sure [this check in `SubjectSubscription`](https://github.com/ReactiveX/rxjs/blob/master/src/SubjectSubscription.ts#L18-L20) is unnecessary because this case is always guarded against with [this check in `Subscription`](https://github.com/ReactiveX/rxjs/blob/master/src/Subscription.ts#L67-L69) but its really not hurting anything. Closing this.

PR #2839, opened 2017-09-19
Fix umd bundling

jasonaden:
This should fix the bundling. The problem was in the past when we did an `es6` build, that was just for the modules. Which is now the same as the new `esm5` build. So I used the `esm5` sources to rollup the UMD.

However, when compiling the old `es6` build, we would pass the `--noEmitHelpers` flag. This was fine in the past because we never published the `es6` output.

Now that we publish both `es6` and `umd`, we need to re-compile with the `--noEmitHelpers` flag enabled.

This can go away with `v6` when we will use the `--importHelpers` flag.


benlesh:
LOL... our npm scripts are amazing now.

PR #2840, opened 2017-09-20
Add jsDelivr links

LukasDrgon:
[jsDelivr](https://www.jsdelivr.com/) is the [fastest opensource cdn](https://www.cdnperf.com/) available and built specifically for production usage. It can serve any project from npm with zero config just like unpkg. This PR adds it to the readme as an alternative to unpkg.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,8 +120,9 @@ npm install @reactivex/rxjs@5.0.0
 
 ### CDN
 
-For CDN, you can use [unpkg](https://unpkg.com/):
+For CDN, you can use [jsDelivr](https://www.jsdelivr.com/package/npm/rxjs) or [unpkg](https://unpkg.com/):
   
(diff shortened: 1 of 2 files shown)


LukasDrgon:
I updated the commit message to match the contribution guidelines.

benlesh:
I'm unfamiliar with jsdelivr or how wide-spread the usage is... It seems like another thing to maintain...

I'll wait for others to chime in on this one.

kwonoj:
I am not in great favor, we've used unpkg as just example to use umd from cdn, not to list available options. User might have their own preference.. 🤔

MartinKolarik:
@benlesh jsDelivr serves about 10 times more traffic than unpkg and there's no maintenance needed as everything is automated now.

benlesh:
Given the lack of interest, I'm closing this one for now.

Issue #2850, opened 2017-09-23
Coverage report checker's broken


kwonoj:
**RxJS version:**

**Code to reproduce:**

**Expected behavior:**

**Actual behavior:**

**Additional information:**

Seems build pipeline change (https://github.com/ReactiveX/rxjs/pull/2804) causes coverage checker no longer works.

kwonoj:
Closing for now, #2851 can be tracked separately.

PR #2851, opened 2017-09-23
Simplify test script pipeline

kwonoj:
**Description:**
This PR fixes https://github.com/ReactiveX/rxjs/issues/2850, also revises current build pipeline around unit tests. Our test setup is quite complex, by using compiled output from src for importing modules, then build each (src, spec) separately to run test cases. This also causes coverage remapping tricky, as recent change caused.

Instead, PR changes test cases to require sources directly - then run compile just once for specs and create coverage as well. It get rid of few npm scripts for testing purpose only, performance wise doesn't have much differences.

Few caveat is it requires to update some other scripts like doc generation but it can be dealt with seperately.

- closes https://github.com/ReactiveX/rxjs/issues/2850.


kwonoj:
@benlesh for reviewing. Mostly it doesn't touch functionality but also it touches huge number of files.

benlesh:
This looks great OJ, I'll try to pull this down and run it locally before I merge it very soon.

benlesh:
Apologies, @kwonoj, I dropped the ball on looking at this. We might want to wait until the 5.5 version is released and then come back to this, as the packaging/building scripts are in a little bit of flux right now.

kwonoj:
I'm fine with that, feel freely to do so.

benlesh:
Okay... so this looks okay... however a 4% drop in code coverage is HUGE. We should look into that before proceeding.

kwonoj:
@benlesh it's good now, it was just glob wasn't smart enough to pick all test files. coverage change is within the threshold.

PR #2853, opened 2017-09-23
docs: Fix case of Observable imports in lettables

intellix:
**Description:** The imports were wrong and this was stinging me when deploying to Heroku (linux) but didn't fail on OSX locally

**Related issue (if exists):** Casing is confusing as it causes issues across OSes


intellix:
Ok, it seems that this isn't actually correct either. Was truly bamboozled but it seems that the published NPM package has uppercased names like `Observable` and `Operator`.

Compare these two:
 - https://registry.npmjs.org/rxjs/-/rxjs-5.5.0-beta.0.tgz
 - https://github.com/ReactiveX/rxjs/tree/master/src

PR #2860, opened 2017-09-25
chore(typings): improve zip typings

cartant:
Add signature to the lettable zipProto so it's consistent with the
non-lettable zipProto.

**Description:**

This PR is closely related to #2787. In between that PR being submitted and its being merged there were WIP commits made re: lettable operators. Consequently, the overload signature that the PR added to the non-lettable `zipProto` was not included in the lettable `zipProto`.

This PR adds the relevant overload to the lettable `zipProto`.

**Related issue (if exists):** #2784


cartant:
Unfortunately, this PR falls victim to TypeScript matching arbitrary functions with `ArrayLike<any>` and, consequently, with `ObservableInput<any>`. That means that if `T` is `any` there is no sensible order for:

```ts
export function zip<T>(...observables: Array<ObservableInput<T>>): OperatorFunction<T, T[]>;
```

and:

```ts
export function zip<T, R>(...observables: Array<ObservableInput<T> | ((...values: Array<T>) => R)>): OperatorFunction<T, R>;
```

When `T` is `any` the whichever is first will be matched regardless of whether or not a projection function is specified.

This is a pity, as when `T` is not `any` it works fine. Such is life.

This also means that the signature added in PR #2787 is also unsafe when `T` is `any` and should probably be removed.

PR #2862, opened 2017-09-25
feat(global): export lettables as Rx.operators

### LINKED ISSUE #2861: Lettable operators are not exposed in global bundle.
**RxJS version:** 5.0.0-beta.0

**Code to reproduce:**
n/a

**Expected behavior:**
Users should be able to access lettable versions of operators in global build. Since the prototype versions use the lettable versions internally, there should be a minimal impact on size.

Recommended we do:  `Rx.operators.map` `Rx.operators.mergeMap`, etc... so that they can be destructured in modern browsers like so:  `const { map, filter, mergeMap } = Rx.operators`

**Actual behavior:**
n/a

**Additional information:**
n/a

benlesh:
fixes #2861

**Description:**

### DIFF
--- a/src/Rx.ts
+++ b/src/Rx.ts
@@ -176,6 +176,10 @@ import { rxSubscriber } from './symbol/rxSubscriber';
 import { iterator } from './symbol/iterator';
 import { observable } from './symbol/observable';
 
+import * as _operators from './operators';
+
+export const operators = _operators;
+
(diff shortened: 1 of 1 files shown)


benlesh:
This does add a lot of size to the global bundle, but the global bundle is really meant for playground code and/or code being used in node.js or the like.

Issue #2859, opened 2017-09-24
throttle and throttleTime broken for single values when trailing is true.


Parakleta:
**RxJS version:**

5.4.3

**Code to reproduce:**

```js
const after100 = ()=>Observable.interval(100).first();
Observable.interval(500).take(5)
    .throttleTime(100, undefined, {trailing;true})
    .subscribe(x=>console.log(x));
Observable.interval(500).take(5)
    .throttle(after100, {trailing:true})
    .subscribe(x=>console.log(x));
Observable.interval(500).take(5)
    .throttle(after100, {leading:true, trailing:true})
    .subscribe(x=>console.log(x));
```

**Expected behavior:**

```
0....1....2....3....4|   <- source stream
.0....1....2....3....4|
.0....1....2....3....4|
0....1....2....3....4|
```

**Actual behavior:**

```
.....................|
.....................|
00...11...22...33...4|
```

**Additional information:**

The changes introduced by #2465 were sorely needed but somewhat botched.  The handling of the trailing value is different between `throttle` and `throttleTime` and wrong in both cases.  In both versions if `leading` is disabled then the first value is discarded despite the fact that it might actually be the trailing value.  In the `throttle` case there is also the problem that for some reason if the leading and trailing values are both enabled it keeps the first value and emits this for both (duplicating the value).

I strongly recommend that #2749 is replicated to cover the `throttle` version as well and merged as soon as possible.  It resolves this issue, as well as #2727 which it was written to solve, and also #2466.

tim-kuteev:
> In both versions if `leading` is disabled then the first value is discarded despite the fact that it might actually be the trailing value.

Is there any chance that this one will be fixed the near future?

PR #2864, opened 2017-09-25
Throttle fix

benlesh:
To address #2859

### DIFF
--- a/spec/operators/throttle-spec.ts
+++ b/spec/operators/throttle-spec.ts
@@ -7,6 +7,8 @@ declare const hot: typeof marbleTestingSignature.hot;
 declare const cold: typeof marbleTestingSignature.cold;
 declare const expectObservable: typeof marbleTestingSignature.expectObservable;
 declare const expectSubscriptions: typeof marbleTestingSignature.expectSubscriptions;
+declare const time: typeof marbleTestingSignature.time;
+declare const rxTestScheduler: typeof marbleTestingSignature.rxTestScheduler;
 
 const Observable = Rx.Observable;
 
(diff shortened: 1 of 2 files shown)


sod:
Was something wrong with https://github.com/ReactiveX/rxjs/pull/2749?

benlesh:
@sod nope... I just didn't see it. sorry.

benlesh:
Okay, @sod I've reviewed your PR. there were a lot of changes I requested, if you think you can knock those out, that would be great. I'll just ween this PR down to only be the `throttle` operator, as your fixes were only to `throttleTime`.

Parakleta:
This has the same problem that #2749 had initially in that it defers the setup of the throttle observable until after emitting the value, and this can lead to drift in the throttling.  I think it needs to look something like:

```js
    this._hasValue = true;
    this._sendValue = value;

    if (!this._throttled) {
        this.throttle(this._leading);
    }
}
private throttle(doSend: boolean) {
    const { _hasValue, _sendValue } = this;
    if (!_hasValue) return;

    let duration, err;
    try {
        duration = this.durationSelector(_sendValue);
    } catch (e) {
        err = e;
    }

    if (doSend) {
        this.destination.next(_sendValue);
        this._hasValue = false;
        this._sendValue = null;
    }

    if (err != null) {
        this.destination.error(err);
    } else {
        this.add(this._throttled = subscribeToResult(this, duration));
    }
}
private throttlingDone() {
    const { _throttled, _trailing } = this;
    if (_throttled) {
        _throttled.unsubscribe();
    }
    this._throttled = null;

    if (_trailing) {
        this.throttle(true);
    }
}
```

The `throttle()` method feels untidy to me, but I'm not sure how to make it nicer.

micha149:
Whats up with this PR?

benlesh:
@micha149 

it needs to be rebased, and I haven't gotten back around to it.

micha149:
@benlesh thanks for the response. will it be possible to get this into Version 6.0.0? Can I support by rebasing your commit?

benlesh:
@micha149 Yeah, we're going to get this in. Feel free to rebase if you have the time.

micha149:
@benlesh I rebased the commit. As far as I know I'm not able to get it into this PR. You can pull from here: https://github.com/micha149/rxjs/commit/df689e5550bef58f534d8779b2c7c27d35440089

Parakleta:
This still doesn't deal with the issue that a slow `next()` emission will cause the throttling to drift (or have inconsistent duration) as discussed in #2749.  I think swapping lines https://github.com/micha149/rxjs/blob/df689e5550bef58f534d8779b2c7c27d35440089/src/internal/operators/throttle.ts#L113-L114 should fix it.

benlesh:
Changes coped to #3505

PR #2865, opened 2017-09-25
fix(Subscriber): creates subscriber throwable by default

### LINKED ISSUE #2813: Errors sometimes not rethrown
Been tracking down this tricky bug that was causing some real-world pain 😢

**RxJS version:** 5.3.1 & 5.4.3

**Code to reproduce:**

Runnable version [here](https://runkit.com/embed/8cwacxv1osjl).

```js
Observable.of(0, Scheduler.async)
  .mergeMap(() =>
    Observable.combineLatest(Observable.of(0))
  )
  .subscribe(() => { throw new Error('uh oh'); });
```

**Expected behavior:**

Error is re-thrown asynchronously and can be handled by normal uncaught exception means. (`window.onerror`, `uncaughtException`).

**Actual behavior:**

Error gets swallowed.

**Additional information:**

`combineLatest` seems to be involved here—if you get rid of it, the example behaves as expected.

kwonoj:
**Description:**

**I have gut feeling this might not be appropriate fix, feel free to close.**

This PR ensure subscriber created internally have nested parents (i.e CombineLatestSubscriber has Mergemap as parents, etcs) are having proper syncErrorThrown behavior. Previously it bypasses setting syncError* property as well as wrapping it into safesubscriber, allows some of cases errors are not thrown but just swallowed.

- closes #2813

### DIFF
--- a/src/Subscriber.ts
+++ b/src/Subscriber.ts
@@ -39,7 +39,7 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
 
   public syncErrorValue: any = null;
   public syncErrorThrown: boolean = false;
-  public syncErrorThrowable: boolean = false;
+  public syncErrorThrowable: boolean = true;
 
   protected isStopped: boolean = false;
   protected destination: PartialObserver<any>; // this `any` is the escape hatch to erase extra type param (e.g. R)
@@ -60,10 +60,12 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
     switch (arguments.length) {
       case 0:
         this.destination = emptyObserver;
+        this.syncErrorThrowable = false;
         break;
       case 1:
         if (!destinationOrNext) {
           this.destination = emptyObserver;
+          this.syncErrorThrowable = false;
           break;
         }
         if (typeof destinationOrNext === 'object') {
@@ -71,13 +73,11 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
             this.destination = (<Subscriber<any>> destinationOrNext);
             (<any> this.destination).add(this);
           } else {
-            this.syncErrorThrowable = true;
             this.destination = new SafeSubscriber<T>(this, <PartialObserver<any>> destinationOrNext);
           }
           break;
         }
       default:
-        this.syncErrorThrowable = true;
         this.destination = new SafeSubscriber<T>(this, <((value: T) => void)> destinationOrNext, error, complete);
         break;
     }


kwonoj:
lol this makes whole test cases fail, not sure why I couldn't caught in local test.

kwonoj:
even if this is accepted it'll be probably 5 only and not being cherrypicked into next branch as planned breaking changes.

benlesh:
I'm not sure about this one... it seems like it will make syncError tracking occur within operators, which I'm pretty sure isn't necessary (which is why the current code avoids it explicitly).

Either way, error rethrowing is going away in 6.0 to reach compliance with the proposal, and to prevent some really, really nasty errors like:


```ts
// imagine we're shipping this off to a few third parties
const source$ = Observable.interval(1000).share();

// third party 1
source$.map(x => x + x).subscribe(x => console.log(x));

// third party 2
source$.filter(x => {
  if (x > 3) throw new Error('haha');
  return x <= 3;
})

// third party 3
source$.mergeMap(x => Observable.timer(x))
```

... what's going to happen above is on the fifth emission, the number `4` hits "third party 2", which throws an error, the error propagates down to the final subscriber, is unhandled, then is rethrown. The rethrown error unwinds the stack back to the loop in the `share()` that is notifying everyone for the multicast. That loop breaks, stopping notifications from being sent to anyone else. This is producer interference.

There are other reasons too... Like how are you going to `try { } catch { }` an error that is thrown asynchronously?

In the end, in v6, we will have a notification channel for unhandled errors, and we'll need to remove the rethrowing behavior.

kwonoj:
> I'm not sure about this one

Yes, reason I mentioned in PR `I have gut feeling this might not be appropriate fix, feel free to close.` :)

>  it seems like it will make syncError tracking occur within operators, which I'm pretty sure isn't necessary

Most cases it's true, refer issue #2813 for some higher order observable cases with asynchronously scheduled. As I said already above I have feeling this might not appropriate fix, but we do currently have completely swalloing exception and doesn't deliver it into top level observer.

> In the end, in v6, we will have a notification channel for unhandled errors, and we'll need to remove the rethrowing behavior.

Absolutely, issue should be v5 targeted fix as mentioned `even if this is accepted it'll be probably 5 only and not being cherrypicked into next branch as planned breaking changes.`

benlesh:
I'm trying to decide if this solves #3183 or not.

kwonoj:
@benlesh worth peek https://github.com/ReactiveX/rxjs/pull/3161 as well, I have guess that might be more feasible to try out. This PR is too ad-hoc honestly.

jayphelps:
did https://github.com/ReactiveX/rxjs/pull/3161 solve the same problem?

benlesh:
Yes, #3161 solved the same problem. I think we can close this. @kwonoj, please verify, and open it up if I'm mistaken.

PR #2867, opened 2017-09-26
dont' -> don't

rgbkrk:
`let` me just say I'm super freakin' excited for the lettable operators, thanks for this


kwonoj:
@rgbkrk off-topic, how do you deal with module sizes in Electron? doing bundling via webpack or something else?

rgbkrk:
We bundle both the `main` and `renderer` process with webpack: https://github.com/nteract/nteract/blob/d84fcc3d2ac3b16987ba3b6548fdeda76feda91d/packages/desktop/webpack.config.js#L25-L44

By the way I'm more than pleased to tune into the bikeshedding on the operator names, as I think we're definitely going to want to adopt these now because of some current lerna + hoisting issues.

rgbkrk:
Oh -- one other sneaky trick we do @kwonoj -- since we do bundling with webpack we stick (most) of our dependencies in `devDependencies`: https://github.com/nteract/nteract/blob/d84fcc3d2ac3b16987ba3b6548fdeda76feda91d/packages/desktop/package.json#L103

kwonoj:
Oh that's interesting, I should take a peek :bow:

kwonoj:
@rgbkrk when you have some free time, mind if I poke you for chat with coffee around packaging (mostly bundling) strategy in nteract?

rgbkrk:
@kwonoj I would love that!

PR #2869, opened 2017-09-26
docs(lettable): mention TS >= 2.4 is required

### LINKED ISSUE #2856: New pipe operators in 5.5 fail in TS < 2.4 as they don't transfer type
**RxJS version:**

RxJS 5.5.0-beta.0
TypeScript 2.3

**Code to reproduce:**

Quite the same example as in [lettable operators doc](https://github.com/ReactiveX/rxjs/blob/master/doc/lettable-operators.md) : 

```typescript
const source$ = range(0, 10);

source$.pipe(
  filter(x => x % 2 === 0),
  map(x => x.toFixed())
)
```

**Expected behavior:**

It should work.

**Actual behavior:**

On TypeScript compilation, it fails with this error : 

```ERROR: Property 'toFixed' does not exist on type '{}'.```

Meaning the type (here `number`) is not transferred by the `pipe` operator. The `pipe` operator seems to default back to `{}` all the time.

**Additional information:**

It works starting TypeScript 2.4+.

If this is normal, it should at least be indicated in the CHANGELOG, as it's an unexpected (breaking ?) change for a minor update.

### PR DESCRIPTION
**Pull request by cyrilletuzi**
Fixes #2856

### DIFF
--- a/doc/lettable-operators.md
+++ b/doc/lettable-operators.md
@@ -4,6 +4,8 @@ Starting in version 5.5 we have shipped "lettable operators", which can be acces
 
 **NOTE**: During 5.5 beta we will be bikeshedding a few of the names for operators that we had to give new names to due to keyword restrictions in javascript. 
 
+**NOTE**: If you're using TypeScript, version >= 2.4 is required, otherwise lettable operators will lose the type and compilation will break. 
+
 These operators are:
 
 1. `do` -> `tap`

### CODE REVIEW

File: doc/lettable-operators.md [hunk 1]
```diff
@@ -4,6 +4,8 @@ Starting in version 5.5 we have shipped "lettable operators", which can be acces
 
 **NOTE**: During 5.5 beta we will be bikeshedding a few of the names for operators that we had to give new names to due to keyword restrictions in javascript. 
 
+**NOTE**: If you're using TypeScript, version >= 2.4 is required, otherwise lettable operators will lose the type and compilation will break. 
+
```
benlesh: This is a little more "doom and gloom" than I'd like.. It sounds like you can't use TypeScript 2.3 or under, which isn't true. It should just be `NOTE: in TypeScript 2.3 and under, typings will need to be added to functions passed to operators` or something like that.

cyrilletuzi: Even if it's true technically, in real world no one will do this. You really need to add the type to each functions, sometimes even twice, so that's clearly not usable. And in some way it's really an error, as lettable operators transfer the wrong type in TS < 2.4, then causing real compilation errors when you do any action on the data. But you're the owner, you choose what's best.

benlesh:
I've added a section to the docs to cover this in more detail. It can be [found here](https://github.com/ReactiveX/rxjs/blob/5dbad94e25d24b7a2ae5b77d2b73b26b51435a53/doc/lettable-operators.md#known-issues)

Thank you for pointing out that we needed this, @cyrilletuzi

PR #2871, opened 2017-09-26
fix(build): inline sources so map files work without the /src directory

jasonaden:


benlesh:
this is going to drastically increase the size of the package we out put... so I'm holding off on it for now. Ideally we find a better solution.

jasonaden:
Closing as this will bloat the distributed package, creating copies of the sources in each of the module formats we support.

PR #2875, opened 2017-09-27
fix(build): make build_test run the minimum to speed it up

jasonaden:
Changing the build slowed down the `npm run build_test` command significantly. This PR should improve speed by only building one copy of the sources. It still runs the same packaging code, skipping packaging of ESM5 and ESM2015.

### DIFF
--- a/.make-packages.js
+++ b/.make-packages.js
@@ -86,28 +86,34 @@ mkdirp.sync(PKG_ROOT);
 // Copy over the sources
 copySources('src/', PKG_ROOT + 'src/');
 copySources(CJS_ROOT, CJS_PKG);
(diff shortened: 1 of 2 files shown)


kwonoj:
I'd like to simply deprecate `build_test` on behalf of https://github.com/ReactiveX/rxjs/pull/2851. Except historical reason, there isn't lot need test cases imports compiled src output instead of import src directly. (also resolves perf issues when running test as well).

benlesh:
@kwonoj ... I still use build_test all the time. haha

kwonoj:
@benlesh yes, but is it needed? by removing reference to compiled src ref from test, you can just `build` or either `test` (and latter won't take significantly longer than current)

PR #2877, opened 2017-09-28
chore(typings): unwrap it-wrapped type tests

cartant:
**Description:**

This is a maintenance PR. In looking to add some tests for #2855, I found a bunch of typings tests - see the [`type`](https://github.com/ReactiveX/rxjs/blob/5.4.3/spec/helpers/testScheduler-ui.ts#L91-L100) function - that are wrapped in `it` calls. I've unwrapped them.

I've made the changes in three commits so that it should be a little easier to review what was changed. I'll squash them, if you're happy with the changes made in the PR.

**Related issue (if exists):** None.

### DIFF
--- a/spec/observables/combineLatest-spec.ts
+++ b/spec/observables/combineLatest-spec.ts
@@ -480,66 +480,66 @@ describe('Observable.combineLatest', () => {
     expectSubscriptions(e2.subscriptions).toBe(e2subs);
   });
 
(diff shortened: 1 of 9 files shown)


kwonoj:
I should've merge this long time ago.. @cartant would you mind rebase to `master`?

cartant:
Should be rebased, now.

Issue #2876, opened 2017-09-28
Inconsistent casing


Delagen:
**RxJS version:** 5.5.0-beta.2

**Code to reproduce:**

**Expected behavior:**
No errors

**Actual behavior:**
TS1149: File name '.../rxjs/observable/of.d.ts' differs from already included file name '.../rxjs/Observable/of.d.ts' only in casing.

**Additional information:**
I think it caused by compilation store files in lowercase folder, but typings compiled in uppercase folder

Delagen:
I think it windows specific.
But now error are gone. Strange.

Delagen:
It occurs when compiler use it default settings.

Delagen:
I use the following in build task, and have no error, but IDE tools reporting it
```
var ts = require("typescript");
ts.sys.useCaseSensitiveFileNames = true;
```

benlesh:
closed by #2878

PR #2878, opened 2017-09-28
fix(build): revert to 5.4.x build output for CJS & add configurable s…

jasonaden:
…upport for ESM

I had attempted to make a fully non-breaking change where we resolve modules through a `package.json` file (one `package.json` file for every module distributed). As it turns out, this doesn't work due to differences between case sensitive and insensitive file systems (and TypeScript configuration options to make case insensitive file systems resolve files with case sensitivity).

This PR makes the basic output the same as it was previously. But it adds distribution of `ESM5` and `ESM2015` build artifacts. It also dynamically generates a file that can be imported and used in Webpack to resolve to the `ESM` distribution.

Usage looks like this:

```
// webpack.config.js

const path = require('path');
const rxPaths = require('rxjs/_esm5/path-mapping');

var config = {
  resolve: {
    alias: rxPaths(path.resolve(__dirname, 'node_modules'))
  }
}
```

### DIFF
--- a/.make-packages.js
+++ b/.make-packages.js
@@ -16,7 +16,7 @@ const ESM2015_ROOT = ROOT + 'esm2015/';
 const UMD_ROOT = ROOT + 'global/';
 const TYPE_ROOT = ROOT + 'typings/';
 const PKG_ROOT = ROOT + 'package/';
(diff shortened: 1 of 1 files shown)


benlesh:
This should close #2876

PR #2879, opened 2017-09-28
refactor: map operators all the way to lettable version rather than r…

### LINKED ISSUE #2812: Some operators e.g. rxjs/operator/multicast pulls in code for *all* operators now
It looks like now that the older operator files import the lettable operators, they (unintentionally?) import _all_ operators.

https://github.com/ReactiveX/rxjs/blob/de9f2c8543381ae5f2fa31eb2c9968329ef4fbe0/src/operator/multicast.ts#L4

https://github.com/ReactiveX/rxjs/blob/de9f2c8543381ae5f2fa31eb2c9968329ef4fbe0/src/operator/buffer.ts#L3

https://github.com/ReactiveX/rxjs/blob/de9f2c8543381ae5f2fa31eb2c9968329ef4fbe0/src/operators/index.ts#L1-L59

***

I also see this pattern is used in [some pending PRs too](https://github.com/ReactiveX/rxjs/pull/2806/files#diff-d31027c07027614c655a42b473b41792R2)

I've been out of the loop, so it's entirely possible I'm confused or there is some magic that prevents it from being a problem--without _requiring_ people use treeshaking. Or maybe that's the plan?

jasonaden:
…e-export

Fixes #2812

### DIFF
--- a/src/operator/audit.ts
+++ b/src/operator/audit.ts
@@ -1,6 +1,6 @@
 
 import { Observable, SubscribableOrPromise } from '../Observable';
-import { audit as higherOrder } from '../operators';
+import { audit as higherOrder } from '../operators/audit';
 
 /**
  * Ignores source values for a duration determined by another Observable, then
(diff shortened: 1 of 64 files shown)


jayphelps:
@jasonaden 👏 👏 👏

ladyleet:
@jasonaden what @jayphelps said! 👏 👏 👏 👏

PR #2882, opened 2017-09-28
refactor: remove operators reexport

jasonaden:
Ensures users won't run into #2812

I think we should consider merging this before pushing 5.5.0. Due to tooling (Webpack, Uglify, etc), if a user does:

```
import {map, merge} from 'rxjs/operators';
```

Their bundle size will increase significantly. On a very small sample project, the above import (as opposed to importing each directly) makes about 120kb change in the final bundle (almost doubling the bundle).

In order to avoid this problem, we shouldn't publish the re-export until tooling properly supports importing through re-exports.


jasonaden:
This will not go green until #2879 is merged and rebased.

jasonaden:
Closing this PR. I think we can fix this issue by running the output package through the build optimizer (https://github.com/angular/ngo). The problem seems to be that all classes imported through the entry point are retained. `ngo` helps fix that by marking downleveled classes as `__PURE__` which allows Uglify to remove them.

jasonaden:
Re-opening this PR. I created a test build of rxjs after running it through the Angular Build Optimizer. While the optimizer helps in final application bundle size, it doesn't allow Uglify to get rid of all unused symbols. This results in about 40k of unminified bloat. It's better than the previous number (which was actually well north of 120k), but is still far too much to be justified.

Therefore I think we should remove the reexport until the problem of pulling in unused code can be solved. This may be solved by Webpack doing better module resolution.

benlesh:
So... webpack is why we can't have nice things?

benlesh:
I don't think this is a direction we want to go, honestly... The direction that we let inadequate (or possibly just hard to configure) tooling dictate how we structure the ergonomics of a project. Esp. when there's no reason it shouldn't work.

jasonaden:
@benlesh So pulling in the Build Optimizer from the Angular project (and making a few hacks to get it to run against the older version of TypeScript we're currently using), I can produce a build where the final application bundle size is really close between using direct deep imports and importing from `rxjs/operators`. I think the size will be acceptably close, and as the tools get better we will see an even further reduction.

Of note is this will rely on running through UglifyJS. Most projects do as it's the default for Webpack. Also, there are some configuration options that should be added to Uglify configuration in Webpack to get the best bundle size. I'll note all this in the PR with the updated build.

TheLarkInn:
@jasonaden webpacks module resolution is working just perfectly. The only flaw is that we implemented es module support via the spec which states that all modules even reexports that aren't used, must be evaluated in the situation of potential side effects. 

Luckily for you, webpack 4 will optimize this and prune unused. Please see our medium publication medium.com/webpack and check out our logbook for more details.

jasonaden:
@TheLarkInn Sorry... my explanation above was a bit brief.

Yes, Webpack 4 should solve some of these issues by pruning unused imports. But I feel like we need a solution that works for people now, rather than publishing something that will be fixed in the future, but add lots of bloat until users upgrade.

But we found a way to do this with the Angular Build Optimizer and adding all the needed /**__PURE__*/ annotations. This gets us almost all the way there, but Webpack 4 should get us the rest of the way.

TheLarkInn:
> But I feel like we need a solution that works for people now, rather than publishing something that will be fixed in the future, but add lots of bloat until users upgrade.

Contributions welcomed. Happy if you want to accelerate the release process we need testing and early tried against webpack @next branch.

aitboudad:
@jasonaden importing lettable operator from `rxjs/operators` produce a large bundle size even with `build-optimizer`, the new option `sideEffects` introduced in Webpack 4 removed the unused operators but only when importing from `_esm5`  folder

```patch
++ import { filter } from 'rxjs/_esm5/operators';
-- import { filter } from 'rxjs/operators';
```
Note: tested with `rxjs: 5.5.0-beta.7`

jasonaden:
@aitboudad Because by default (`rxjs/operators`), you will import CJS versions of these operators, you will get all of Rx by running this import. CJS cannot be treeshaken, so there will be a bunch of dead code in the bundle.

I'll be writing down the steps to get tree shaking with this new import, but if you import the `path-mapping` file as showing in [this gist](https://gist.github.com/jasonaden/9d5588d8d0c0b51b94eef96329d3fe9c), you should get tree shaking back.

I haven't written this up yet because we found a few environments where using this `path-mapping` doesn't seem to result in tree shaking, so still debugging these use cases.

cc @benlesh @IgorMinar

Issue #2844, opened 2017-09-22
Add selector to publishReplay


benlesh:
We need to add an optional selector argument to `publishReplay` similar to what `publish` has.

Related #2841

PR #2885, opened 2017-10-01
feat(publishReplay): add selector function to publishReplay

martinsik:
**Description:**

This PR adds selector function to the `publishReplay` operator similarly to `publish`.

I didn't modify any existing tests. However, this could be probably a breaking change because I added the parameter before `scheduler` (I'm not sure this is a problem or not).

#2844

### DIFF
--- a/spec/operators/publishReplay-spec.ts
+++ b/spec/operators/publishReplay-spec.ts
@@ -411,13 +411,73 @@ describe('Observable.prototype.publishReplay', () => {
   });
 
   it('should mirror a simple source Observable with selector', () => {
-    const selector = observable => observable.map(v => String.fromCharCode(96 + parseInt(v)));
+    const values = {a: 2, b: 4, c: 6, d: 8};
+    const selector = observable => observable.map(v => 2 * v);
     const source = cold('--1-2---3-4---|');
     const sourceSubs =  '^             !';
     const published = source.publishReplay(1, Number.POSITIVE_INFINITY, selector);
     const expected =    '--a-b---c-d---|';
 
+    expectObservable(published).toBe(expected, values);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+  });
+
+  it('should emit an error when the selector throws an exception', () => {
+    const error = "It's broken";
+    const selector = () => {
+      throw error;
+    };
+    const source = cold('--1-2---3-4---|');
+    const published = source.publishReplay(1, Number.POSITIVE_INFINITY, selector);
+
+    // The exception is thrown outside Rx chain (not as an error notification).
+    expect(() => published.subscribe()).to.throw(error);
+  });
+
+  it('should emit an error when the selector returns an Observable that emits an error', () => {
+    const error = "It's broken";
+    const innerObservable = cold('--5-6----#', undefined, error);
+    const selector = observable => observable.mergeMapTo(innerObservable);
+    const source = cold('--1--2---3---|');
+    const sourceSubs =  '^          !';
+    const published = source.publishReplay(1, Number.POSITIVE_INFINITY, selector);
+    const expected =    '----5-65-6-#';
+
+    expectObservable(published).toBe(expected, undefined, error);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+  });
+
+  it('should terminate immediately when the selector returns an empty Observable', () => {
+    const selector = () => Observable.empty();
+    const source = cold('--1--2---3---|');
+    const sourceSubs =  '(^!)';
+    const published = source.publishReplay(1, Number.POSITIVE_INFINITY, selector);
+    const expected =    '|';
+
     expectObservable(published).toBe(expected);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+  });
+
+  it('should not emit and should not complete/error when the selector returns never', () => {
+    const selector = () => Observable.never();
+    const source = cold('-');
+    const sourceSubs =  '^';
+    const published = source.publishReplay(1, Number.POSITIVE_INFINITY, selector);
+    const expected =    '-';
+
+    expectObservable(published).toBe(expected);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+  });
+
+  it('should emit error when the selector returns Observable.throw', () => {
+    const error = "It's broken";
+    const selector = () => Observable.throw(error);
+    const source = cold('--1--2---3---|');
+    const sourceSubs =  '(^!)';
+    const published = source.publishReplay(1, Number.POSITIVE_INFINITY, selector);
+    const expected =    '#';
+
+    expectObservable(published).toBe(expected, undefined, error);
     expectSubscriptions(source.subscriptions).toBe(sourceSubs);
   });
 });
--- a/spec/operators/publishReplay-spec.ts
+++ b/spec/operators/publishReplay-spec.ts
@@ -409,4 +409,15 @@ describe('Observable.prototype.publishReplay', () => {
 
     published.connect();
   });
+
+  it('should mirror a simple source Observable with selector', () => {
+    const selector = observable => observable.map(v => String.fromCharCode(96 + parseInt(v)));
+    const source = cold('--1-2---3-4---|');
+    const sourceSubs =  '^             !';
+    const published = source.publishReplay(1, Number.POSITIVE_INFINITY, selector);
+    const expected =    '--a-b---c-d---|';
+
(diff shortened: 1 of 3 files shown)


spec/operators/publishReplay-spec.ts
```diff
@@ -409,4 +409,15 @@ describe('Observable.prototype.publishReplay', () => {
...
+    const selector = observable => observable.map(v => String.fromCharCode(96 + parseInt(v)));
+    const source = cold('--1-2---3-4---|');
+    const sourceSubs =  '^             !';
+    const published = source.publishReplay(1, Number.POSITIVE_INFINITY, selector);
+    const expected =    '--a-b---c-d---|';
+
+    expectObservable(published).toBe(expected);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+  });
 });
```
benlesh: Need to add tests for:

- errors thrown in `selector` directly.
- when `selector` returns an errored Observable (`--a--b--#`)
- when `selector` returns empty Observable
- when `selector` returns never
- when `selector` returns Observable.throw

benlesh:
Also, @martinsik, if you can explain your cool pixel art profile pic, that would be good too. What is it?

src/operator/publishReplay.ts
```diff
@@ -6,13 +6,15 @@ import { publishReplay as higherOrder } from '../operators/publishReplay';
...
  * @param bufferSize
  * @param windowTime
+ * @param selector
  * @param scheduler
- * @return {ConnectableObservable<T>}
+ * @return {Observable<T> | ConnectableObservable<T>}
  * @method publishReplay
  * @owner Observable
  */
 export function publishReplay<T>(this: Observable<T>, bufferSize: number = Number.POSITIVE_INFINITY,
```
benlesh: Like below, we'll need to add some function overloads for TypeScript users to keep this change from being breaking. Here's what the `publish` counterpart to this file is doing...

https://github.com/ReactiveX/rxjs/blob/accbcd0c5f9fd5976be3f491d454c4a61f699c4b/src/operator/publish.ts#L7-L9

src/operators/publishReplay.ts
```diff
@@ -7,6 +7,8 @@ import { UnaryFunction } from '../interfaces';
 
 export function publishReplay<T>(bufferSize: number = Number.POSITIVE_INFINITY,
```
benlesh: I think we'll need some function signature overloads for this so we don't break compatability for people expecting `ConnectableObservable<T>` to be the only return type... 

See what `publish` is doing here for a better idea: https://github.com/ReactiveX/rxjs/blob/accbcd0c5f9fd5976be3f491d454c4a61f699c4b/src/operators/publish.ts#L6-L8

benlesh:
LGTM.... @kwonoj?

martinsik:
@benlesh Sorry, it's work in progress I just pushed it because I was too late yesterday :). I'm not sure about a few things there.

btw, my avatar is a screenshot from this very old game https://en.wikipedia.org/wiki/Moonstone:_A_Hard_Days_Knight

spec/operators/publishReplay-spec.ts
```diff
@@ -409,4 +409,75 @@ describe('Observable.prototype.publishReplay', () => {
...
+  it('should emit an error when the selector throws an exception', () => {
+    const error = "It's broken";
+    const selector = () => {
+      throw error;
+    };
+    const source = cold('--1-2---3-4---|');
+    const published = source.publishReplay(1, Number.POSITIVE_INFINITY, selector);
+
+    // The exception is thrown outside Rx chain (not as an error notification).
+    expect(() => published.subscribe()).to.throw(error);
```
martinsik: This exception is thrown outside the chain so I need to check it like this. This situations isn't tested in neither `publish` nor `multicast` (if I didn't miss anything) so I'm not sure I'm doing it right.

src/operator/publishReplay.ts
```diff
@@ -2,17 +2,27 @@ import { Observable } from '../Observable';
...
-export function publishReplay<T>(this: Observable<T>, bufferSize: number = Number.POSITIVE_INFINITY,
-                                 windowTime: number = Number.POSITIVE_INFINITY,
-                                 scheduler?: IScheduler): ConnectableObservable<T> {
-  return higherOrder(bufferSize, windowTime, scheduler)(this) as ConnectableObservable<T>;
+export function publishReplay<T, R>(this: Observable<T>, bufferSize?: number,
+                                    windowTime?: number,
+                                    selectorOrScheduler?: IScheduler | OperatorFunction<T, R>,
+                                    scheduler?: IScheduler): Observable<R> | ConnectableObservable<R> {
+
+  return higherOrder(bufferSize, windowTime, selectorOrScheduler as any, scheduler)(this);
```
martinsik: I couldn't find any other sane way to call `higherOrder`. TypeScript didn't allow me just using:

```
higherOrder(bufferSize, windowTime, selectorOrScheduler, scheduler)
```

.. but I think this should work.

src/operators/publishReplay.ts
```diff
@@ -3,10 +3,25 @@ import { ReplaySubject } from '../ReplaySubject';
...
 import { ConnectableObservable } from '../observable/ConnectableObservable';
-import { UnaryFunction } from '../interfaces';
+import { UnaryFunction, MonoTypeOperatorFunction, OperatorFunction } from '../interfaces';
 
-export function publishReplay<T>(bufferSize: number = Number.POSITIVE_INFINITY,
-                                 windowTime: number = Number.POSITIVE_INFINITY,
-                                 scheduler?: IScheduler): UnaryFunction<Observable<T>, ConnectableObservable<T>> {
-  return (source: Observable<T>) => multicast(new ReplaySubject<T>(bufferSize, windowTime, scheduler))(source) as ConnectableObservable<T>;
+/* tslint:disable:max-line-length */
+export function publishReplay<T>(bufferSize?: number, windowTime?: number, scheduler?: IScheduler): UnaryFunction<Observable<T>, ConnectableObservable<T>>;
```
martinsik: The `publish` operator defines `MonoTypeOperatorFunction` return type but I don't think that's correct because this results into `Observable<T>` while `publish` should return `ConnectableObservable<T>` I think.

https://github.com/ReactiveX/rxjs/blob/master/src/operators/publish.ts#L6-L8

src/operators/publishReplay.ts
```diff
@@ -3,10 +3,25 @@ import { ReplaySubject } from '../ReplaySubject';
...
+export function publishReplay<T>(bufferSize?: number, windowTime?: number, selector?: MonoTypeOperatorFunction<T>, scheduler?: IScheduler): MonoTypeOperatorFunction<T>;
+export function publishReplay<T, R>(bufferSize?: number, windowTime?: number, selector?: OperatorFunction<T, R>, scheduler?: IScheduler): OperatorFunction<T, R>;
+/* tslint:enable:max-line-length */
+
+export function publishReplay<T, R>(bufferSize?: number,
```
martinsik: I removed all default values here and left it up to `ReplaySubject` to decide.

spec/operators/publishReplay-spec.ts
benlesh: OH GEEZ! I missed this... So what *should* happen is you should catch the error and send it down the `error` path with `observer.error(err)`. There are all sorts of examples of how to handle this around teh library. 

Basically any user-supplied function needs to have it's errors caught and sent down the error channel of observation.

src/operator/publishReplay.ts
benlesh: This will do, the other thing you can do is duplicate the logic that figures out what was passed from the other method.

src/operators/publishReplay.ts
benlesh: This is fine for now... we might need to address this more generally across the library.

spec/operators/publishReplay-spec.ts
martinsik: But I would expect `multicast` to catch this error for me because that's where the selector is called: https://github.com/ReactiveX/rxjs/blob/master/src/operators/multicast.ts#L64

src/operator/publishReplay.ts
martinsik: I understand, I didn't want to write the same logic twice so I'll leave it as it is.

martinsik:
@benlesh I added a comment above that I'd expect `multicast` to catch the error because that's where the selector is called https://github.com/ReactiveX/rxjs/blob/master/src/operators/multicast.ts#L64.

benlesh:
Merged! Thanks @martinsik!

Issue #2889, opened 2017-10-02
`publish`'s `selector` doesn't have return type captured in generics in TypeScript


benlesh:
**RxJS version:** 5.4.x

**Code to reproduce:**

```ts
const result: Observable<string> = of<number>(1, 2, 3).publish(source => source.map(x => x + '!'))
                                                                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```
**Expected behavior:**

No errors, proper type inferrence

**Actual behavior:**
```
error TS2345: Argument of type '(source: Observable<number>) => Observable<string>' is not assignable to parameter of type 'selector<number>'.
  Type 'Observable<string>' is not assignable to type 'Observable<number>'
```

**Additional information:**
TS 2.4

PR #2891, opened 2017-10-03
fix(publish): fix selector typings

cartant:
**Description:**

Fixes the typings for `publish` and `multicast`, so that selectors that change the observable's type are supported.

Adds some typings tests for `publish` and `multicast`.

**Related issue (if exists):** #2889

### DIFF
--- a/src/operators/publish.ts
+++ b/src/operators/publish.ts
@@ -1,10 +1,11 @@
 import { Subject } from '../Subject';
 import { multicast } from './multicast';
-import { MonoTypeOperatorFunction } from '../interfaces';
+import { MonoTypeOperatorFunction, OperatorFunction } from '../interfaces';
 
 /* tslint:disable:max-line-length */
 export function publish<T>(): MonoTypeOperatorFunction<T>;
 export function publish<T>(selector: MonoTypeOperatorFunction<T>): MonoTypeOperatorFunction<T>;
+export function publish<T, R>(selector: OperatorFunction<T, R>): OperatorFunction<T, R>;
 /* tslint:enable:max-line-length */
 
 /**
@@ -20,7 +21,7 @@ export function publish<T>(selector: MonoTypeOperatorFunction<T>): MonoTypeOpera
  * @method publish
  * @owner Observable
  */
-export function publish<T>(selector?: MonoTypeOperatorFunction<T>): MonoTypeOperatorFunction<T> {
+export function publish<T, R>(selector?: OperatorFunction<T, R>): MonoTypeOperatorFunction<T> | OperatorFunction<T, R> {
   return selector ?
     multicast(() => new Subject<T>(), selector) :
     multicast(new Subject<T>());
--- a/spec/operators/multicast-spec.ts
+++ b/spec/operators/multicast-spec.ts
@@ -3,6 +3,7 @@ import * as Rx from '../../dist/package/Rx';
 import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
 
 declare const { asDiagram, time, rxTestScheduler };
+declare const type;
 declare const hot: typeof marbleTestingSignature.hot;
 declare const cold: typeof marbleTestingSignature.cold;
 declare const expectObservable: typeof marbleTestingSignature.expectObservable;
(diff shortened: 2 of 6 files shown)


cartant:
With the changes in this PR, the `selector` exported from [`operator/publish.ts`](https://github.com/ReactiveX/rxjs/blob/master/src/operator/publish.ts#L28) is no longer used (internally). Should it be removed?

kwonoj:
I recall @benlesh intentionally brought `MonoTypeOperatorFunction` 🤔

src/operators/publish.ts
```diff
@@ -20,7 +21,7 @@ export function publish<T>(selector: MonoTypeOperatorFunction<T>): MonoTypeOpera
  * @method publish
  * @owner Observable
  */
-export function publish<T>(selector?: MonoTypeOperatorFunction<T>): MonoTypeOperatorFunction<T> {
+export function publish<T, R>(selector?: any): any {
```
benlesh: This should be `publish<T, R>(selector?: OperatorFunction<T, R>): OperatorFunction<T, R>`

benlesh:
@kwonoj ... this is mostly right. The problem I encountered at work (now that I'm using RxJS with TypeScript all the time, haha) was   `Observable.of(1, 2, 3).publish(source => source.map(x + '!'))` ... would fail because typings didn't permit it. So you had to hack your way through it.

benlesh:
@cartant if you can try to get this in today I'll try to get it out in the next beta. Because I've had to code around it, and it's annoying. haha

src/operators/publish.ts
cartant: Using `OperatorFunction<T, R>` instead of `any` effects this error:

```
25   return selector ?
            ~~~~~~~~~~
26     multicast(() => new Subject<T>(), selector) :
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
27     multicast(new Subject<T>());
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

dist/src/operators/publish.ts(25,10): error TS2322: Type '((source: Observable<T>) => Observable<R>) | ((source: Observable<T>) => Observable<T>)' is not assignable to type '(source: Observable<T>) => Observable<R>'.
  Type '(source: Observable<T>) => Observable<T>' is not assignable to type '(source: Observable<T>) => Observable<R>'.
    Type 'Observable<T>' is not assignable to type 'Observable<R>'.
      Type 'T' is not assignable to type 'R'.
```

The problem is the returning of `multicast(new Subject<T>())` - that call sees the return type inferred as `OperatorFunction<T, T>` - hence the error.

The problem can be fix by using `any` here, instead of in the signature:

```
  return selector ?
    multicast(() => new Subject<T>(), selector) :
    multicast(new Subject<T>()) as any; // <-- here
```

Would you prefer that?

cartant:
@benlesh After a couple of iterations - each preceded by a cup of coffee - there are now no `any`s in the typings.

src/operators/publish.ts
```diff
@@ -20,7 +21,7 @@ export function publish<T>(selector: MonoTypeOperatorFunction<T>): MonoTypeOpera
  * @method publish
  * @owner Observable
  */
-export function publish<T>(selector?: MonoTypeOperatorFunction<T>): MonoTypeOperatorFunction<T> {
+export function publish<T, R>(selector?: OperatorFunction<T, R>): MonoTypeOperatorFunction<T> | OperatorFunction<T, R> {
   return selector ?
```
benlesh: I think you can go with just OperatorFunction<T, R> as the return type.

cartant: Nope. You can't. That will effect the error I mentioned in response to your initial change request.

cartant: It a necessity. I don't see it as a problem, as TypeScript uses only the overload signatures when matching methods calls. It won't match calls against the implementation signature - which I why I have the of whacking implementation signatures with the `any` hammer. That is, the implementation signature does not form part of the method's public contract; it only enforces the internal contract.

benlesh: Yup. I see why now... This works. Sorry for any confusion.

PR #2893, opened 2017-10-03
[Fix] Make overload signature compatible in FromEventObservable

gund:
**Description:** Fix compilation error in FromEventObservable.ts file:
```bash
static create<T>(target: EventTargetLike, eventName: string, selector: SelectorMethodSignature<T>): Observable<T>;
       ~~~~~~

dist/cjs/src/observable/FromEventObservable.ts(57,10): error TS2394: Overload signature is not compatible with function implem
entation.
```

This is my first contribution to this repo so if there is something missing please let me know =)
Thanks.

### DIFF
--- a/src/observable/FromEventObservable.ts
+++ b/src/observable/FromEventObservable.ts
@@ -101,11 +101,11 @@ export class FromEventObservable<T> extends Observable<T> {
    */
   static create<T>(target: EventTargetLike,
                    eventName: string,
-                   options?: EventListenerOptions,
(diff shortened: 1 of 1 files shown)


kwonoj:
I think we may need massive dependency update in next branch, by reading travis failure logs. :(

gund:
@kwonoj actually I just fixed those errors in unit tests and they are passing on my local machine.
Will submit that PR shortly

PR #2895, opened 2017-10-03
[Test] Fix types in WebSockerSubject and from-promise-spec

gund:
**Description:** This PR fixes remaining compiler errors in unit tests along with [this one](https://github.com/ReactiveX/rxjs/pull/2893).

Once both merged you should have all tests passing.

BTW:
There is one small change in the `WebSocketSubjectConfig` type and I just want to make sure that the change I made actually makes sense.
So basically there is a property named `resultSelector` which is a function - result of which was generic.
And I bound it's generic type to `WebSocketSubject` generic type (in order to fix compiler error) assuming that this is what user will get after this function will run.
And now basically `resultSelector` prop of `WebSocketSubjectConfig` (if provided) will drive inferred type of the whole `WebSocketSubject` for ex:
```ts
const sock$ = WebSocketSubject.create('ws://socket');
// sock$ type inferred to WebSocketSubject<{}> which was always the case in prev version

const sock2$ = WebSocketSubject.create({url: 'ws://socket', resultSelector: () => false});
// sock2$ type inferred to WebSocketSubject<boolean> because of the result type of 'resultSelector' function
```

Thanks

### DIFF
--- a/spec/observables/from-promise-spec.ts
+++ b/spec/observables/from-promise-spec.ts
@@ -76,7 +76,7 @@ describe('Observable.fromPromise', () => {
     class CustomPromise<T> implements PromiseLike<T> {
       constructor(private promise: PromiseLike<T>) {
       }
-      then(onFulfilled?, onRejected?): PromiseLike<T> {
+      then(onFulfilled?, onRejected?) {
         return new CustomPromise(this.promise.then(onFulfilled, onRejected));
       }
     }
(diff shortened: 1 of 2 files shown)


kwonoj:
I'll review current PR shortly, but if you're planning more changes to next branch I'd like to suggest bit hold off - there will be massive changes planned in next branch and it is possible either new change will bring merge conflict or got lost while picking changes.

gund:
@kwonoj sure, thanks for letting me know.

PR #2898, opened 2017-10-04
refactor: add @angular-devkit optimizations so the Rx bundle can be t…

jasonaden:
…ree-shaken

This change will modify the ESM5 output to add annotations used by Uglify in order to treeshake TypeScript downleveled classes.

### DIFF
--- a/.make-packages.js
+++ b/.make-packages.js
@@ -8,6 +8,14 @@ let klawSync = require('klaw-sync');
 let licenseTool = require('./tools/add-license-to-file');
 let addLicenseToFile = licenseTool.addLicenseToFile;
 let addLicenseTextToFile = licenseTool.addLicenseTextToFile;
+let bo = null;
+// Build Optimizer is not available on Node 4.x. Using a try/catch
+// here to make sure the build passes on Travis using Node 4, but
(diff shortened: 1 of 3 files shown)


benlesh:
LGTM other than the build error in Node 4.0. If you get can the build-optimizer fixed for Node 4.* then I think it's good to go.

cc @kwonoj for additional review.

kwonoj:
it's just destructuring `const { inputFilePath } = options;` doesn't work on node 4. quick fix should be ok to check this in.

jasonaden:
Talking to the build optimizer team, there are other things that likely will not work with Node 4. Therefore I changed the script so it doesn't error with Node 4. It's not actually needed, since we don't produce the Rx build with Node 4. The modification to `.make-packages.js` will allow Travis to pass by not running build-optimizer if Node can't `require` it.

cc @kwonoj @benlesh

kwonoj:
This seems falls under goal one of our long to-do list of auto publish from travis on specific job, meanwhile this looks fine for me.

kwonoj:
Just let anyone on local publishes package don't use node@4 😅

PR #2899, opened 2017-10-04
fix(subscribeOn): remove subscribeOn from reexport to support treesha…

jasonaden:
…kability

There is more research to be done (see #2900), but right now `subscribeOn` needs to be removed from `operators/index.ts`. Having it in there pulls in a tree totaling around 20k unminified. The tree that gets pulled in includes subscribeOn, schedulers/asap, AsapAction, ImmediateDefiniton, root, and a bit more.

### DIFF
--- a/src/operators/index.ts
+++ b/src/operators/index.ts
@@ -75,7 +75,13 @@ export { skipLast } from './skipLast';
 export { skipUntil } from './skipUntil';
 export { skipWhile } from './skipWhile';
 export { startWith } from './startWith';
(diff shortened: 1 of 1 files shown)


benlesh:
This look good... I think removing this ahead of a 5.5 release is better than including it and having it cause hard to explain bundle size problems for people. Especially given our top priority is decreasing deployment size for people. We can always readd it later and not break semver.

benlesh:
ping @kwonoj for second opinion.

PR #2901, opened 2017-10-04
[Fix] Pluck overload signature

gund:
**Description:** With new version of typescript and new keyword `keyof` it is possible now to add type safety to the signature of `pluck` operator.

But because nature of pluck is quite dynamic it is impossible to completely describe statically return type of this operator however for single argument version it's feasible.

This change yields the next result:
```ts
const obs = Observable.of({ a: 1, b: { c: '2' } });

const a = obs.pluck('a');
// 'a' has type Observable<number> <-- This is new safer behavior

const b = obs.pluck('b', 'c');
// 'b' has type Observable<{}> <-- 2 or more arguments version still works as before

const c = obs.pluck<any, string>('b', 'c');
// 'c' has type Observable<string> <-- And it is still possible to manually specify the return type
```

Thanks

### DIFF
--- a/src/operator/pluck.ts
+++ b/src/operator/pluck.ts
@@ -27,7 +27,9 @@ import { map } from './map';
  * @method pluck
  * @owner Observable
  */
-export function pluck<T, R>(this: Observable<T>, ...properties: string[]): Observable<R> {
+export function pluck<T, R extends keyof T>(this: Observable<T>, property: R): Observable<T[R]>;
+export function pluck<T, R>(this: Observable<T>, ...properties: string[]): Observable<R>;
+export function pluck<T, R, U extends keyof T>(this: Observable<T>, ...properties: string[]): Observable<T[U] | R> {
   const length = properties.length;
   if (length === 0) {
     throw new Error('list of properties cannot be empty.');


kwonoj:
As I suggested in https://github.com/ReactiveX/rxjs/pull/2895#issuecomment-333949400, I'd like to hold off changes to `next` branch for a moment to prevent diverge between 2 branches. It is already quite diverged and we expect huge number of merge conflict resolving. Do you feel this is urgently reviewed and checked in?

gund:
I just saw that you merged my previous PRs so thought that it's okay now to make new PR =)
Anyway this is not urgent so take your time.
Thanks

benlesh:
Yeah... @gund, we have a really nasty merge coming up that I'm setting aside some time to do. After that, we're going to move `next` to `master` and press onward. But first, we need to get 5.5 out

benlesh:
@gund ... sorry for the long delay. Would you like to update this PR (probably just create a new one) with the same changes to our pipeable version of `pluck`? (heh... "pipeable pluck" ... that's perfectly pleasant).

cc @david-driscoll can you help out if @gund isn't able?

gund:
Hey @benlesh, thanks for the follow up.

I will take a look into this in a few days as right now I do not have access to computer.

gund:
Just a quick question: should this change be done in next branch or any different?

Issue #2904, opened 2017-10-05
Use PromiseConstructorLike and PromiseLike for operators/toPromise


cartant:
The overload signatures for the new `operators/toPromise` function include:

```ts
export function toPromise<T>(PromiseCtor: typeof Promise): UnaryFunction<Observable<T>, Promise<T>>;
```

If the intention is to allow for the use of an arbitrary promise implementation, this seems overly specific.

`PromiseConstructorLike` and `PromiseLike` could be used instead:

```ts
export function toPromise<T>(PromiseCtor: PromiseConstructorLike): UnaryFunction<Observable<T>, PromiseLike<T>>;
```

**RxJS version:** 5.5.0-beta.3

**Code to reproduce:**

```ts
import { Promise as Bluebird } from "bluebird";
import { toPromise } from "rxjs/operators";

const func = toPromise(Bluebird);
```

**Expected behavior:**

No error.

**Actual behavior:**

An error is effected, as Bluebird does not include the `Symbol.species` property.

**Additional information:**

Since at least version 2.0.10, TypeScript's in-built typings have included `PromiseLike` and `PromiseConstructorLike`.

PR #2905, opened 2017-10-05
chore(typings): improve toPromise typings

cartant:
**Description:**

Add overload signatures that use TypeScript's built-in `PromiseConstructorLike` and
`PromiseLike` types.

**Related issue (if exists):** #2904


cartant:
@benlesh Done. Do you want me to replace `typeof Promise` with `PromiseConstructor`, as it's a little clearer and `Promise` has been declared [like this](https://github.com/Microsoft/TypeScript/blob/v2.0.2/lib/lib.es2015.promise.d.ts#L187) since TypeScript 2.0 (at least) was released?

PR #2914, opened 2017-10-06
To promise move

benlesh:
- removes lettable version of `toPromise`
- adds `toPromise` as a permanent member of Observable, since it's an interop point and means of subscription, and not really an "operator".

This enables:

```ts
source$.pipe(map(x => x + x), filter(x => x < 10)).toPromise()
```


src/operator/toPromise.ts
```diff
@@ -57,5 +56,5 @@ export function toPromise<T>(this: Observable<T>, PromiseCtor: PromiseConstructo
  * @owner Observable
  */
 export function toPromise<T>(this: Observable<T>, PromiseCtor?: PromiseConstructorLike): Promise<T> {
```
kwonoj: can't we now simplify signature as 
```
const toPromise : typeof Observable.prototype.toPromise = ...
```
and remove duplicated signature? (not sure this'll work current version of TSC we're using though)

benlesh:
@kwonoj give it another look.

PR #2924, opened 2017-10-09
fix(shareReplay): again uses `lift`

benlesh:
- adds a test to ensure lift it being used.
- uses lift.

NOTE: this is a first attempt also at using `lift` in a non-class-based (but also non-breaking) way.

### DIFF
--- a/src/operators/shareReplay.ts
+++ b/src/operators/shareReplay.ts
@@ -3,18 +3,24 @@ import { ReplaySubject } from '../ReplaySubject';
 import { IScheduler } from '../Scheduler';
 import { Subscription } from '../Subscription';
 import { MonoTypeOperatorFunction } from '../interfaces';
+import { Subscriber } from '../Subscriber';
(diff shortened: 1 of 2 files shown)


src/operators/shareReplay.ts
```diff
@@ -3,18 +3,24 @@ import { ReplaySubject } from '../ReplaySubject';
...
+
 /**
  * @method shareReplay
  * @owner Observable
  */
 export function shareReplay<T>(bufferSize?: number, windowTime?: number, scheduler?: IScheduler ): MonoTypeOperatorFunction<T> {
+  return (source: Observable<T>) => source.lift(shareReplayOperator(bufferSize, windowTime, scheduler));
+}
+
+export function shareReplayOperator<T>(bufferSize?: number, windowTime?: number, scheduler?: IScheduler) {
```
kwonoj: we don't need export this function.

PR #2925, opened 2017-10-09
refactor(map): use closure style operator

benlesh:
This refactors map to not use a class for the Operator definition. The result is an improvement in performance in Turbofan (tested in Node v 8.6.0) and a slight reduction in output size.

**master**

```
                                         |                     RxJS 4.1.0 |              RxJS 5.5.0-beta.5 |          factor |      % improved
                         map - immediate |                49,724 (±5.69%) |               291,789 (±3.86%) |           5.87x |          486.8%
                                     map |                26,662 (±0.99%) |               280,664 (±0.69%) |          10.53x |          952.7%
```

**this branch**

```
                                         |                     RxJS 4.1.0 |              RxJS 5.5.0-beta.5 |          factor |      % improved
                         map - immediate |                63,424 (±0.92%) |               462,842 (±1.03%) |           7.30x |          629.8%
                                     map |                27,536 (±0.92%) |               273,656 (±1.56%) |           9.94x |          893.8%

```

The notable part here is the 462k ops/sec vs the 291k ops/sec.

This PR was done more as an experiment to see what the perf impact was of both moving to closures for operator definitions and Turbofan. `map` seemed like a simple enough guinea pig.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,7 @@ environment:
 install:
   - ps: Install-Product node $env:nodejs_version
   - set PATH=%APPDATA%\npm;%PATH%
+  - npm install -g npm@4
   - node -v && npm -v
   - npm install
 


cartant:
Why is there a relatively large difference in the ops/sec for `map - immediate` in RxJS 4.1.0 between the two runs?

* 49,724 for master; and
* 63,424 for this branch

benlesh:
@cartant I'm not sure... but you'll notice that the accuracy of the Rx4 tests varies a bit more. I ran the tests several times, and the Rx 5 stuff was very consistent.

kwonoj:
off topic thought, but maybe we need switch perf tests compare against 5 to 5 now (latest rel, or master vs. PR) instead of 4 to 5?

src/operators/map.ts
```diff
@@ -41,17 +41,14 @@ export function map<T, R>(project: (value: T, index: number) => R, thisArg?: any
...
 }
 
-export class MapOperator<T, R> implements Operator<T, R> {
-  constructor(private project: (value: T, index: number) => R, private thisArg: any) {
-  }
-
-  call(subscriber: Subscriber<R>, source: any): any {
-    return source.subscribe(new MapSubscriber(subscriber, this.project, this.thisArg));
-  }
+export function mapOperator<T, R> (project: (value: T, index: number) => R, thisArg?: any) {
```
kwonoj: how much difference it'll be between `function` to property based, such as
````
const mapOperator = (project, thisArg) => (this, source) => source.subscribe(new MapSubscribe())
```?

I usually prefer it's shorthanded as long as it performs not huge different, prolly most difference is between arrow fn to normal.

benlesh:
@kwonoj it might make things easier... however, it might result in bad perf tests because parts of the second one to be run could get optimized. as it'll be almost hte same thing. I ran into that once even testing against v4. The order in which the tests were run affected the perf in a few one-off cases.

kwonoj:
Yup, perf is concern for taking those so asked out.

src/operators/map.ts
benlesh: There isn't much difference, however, since we have to use the `this` context here, I purposely avoided arrow functions.

Generally, I like arrow functions better also.

benlesh:
@kwonoj any idea why Appveyor is dying on this one? The error in Node 8 doesn't make any sense.

kwonoj:
@benlesh it should be good now.

benlesh:
This change was an experiment, and is mostly irrelevant, closing.

PR #2932, opened 2017-10-10
fix(parseMarbles): should handle group length correctly

chrmod:
This PR aim to fix a problem with `TestScheduler.parseMarbles`, but apparently I'm not getting something about this function behavior behavior. 

My expectation for this marble `--(ab)-c-|` is:
```
20 - a
20 - b
40 - c
60 - complete
```

but `parseMarbles` add up entire group as it would be individual values which result in:
```
20 - a
20 - b
70 - c
90 - complete
``` 

I've added the test that covers this example.

The fix for a test is commented out https://github.com/ReactiveX/rxjs/compare/master...chrmod:group-length?expand=1#diff-951c3a2803f2684f1f28e92e47607418R240 as it breaks lots of tests (practically every test with marble containing `(` character).

Please let me know what I've missed.

### DIFF
--- a/spec/schedulers/TestScheduler-spec.ts
+++ b/spec/schedulers/TestScheduler-spec.ts
@@ -73,6 +73,16 @@ describe('TestScheduler', () => {
         { frame: 30, notification: Notification.createNext('c') }
       ]);
     });
+
+    it('should handle group length', () => {
+      const result = TestScheduler.parseMarbles('--(ab)-c-|');
+      expect(result).deep.equal([
+        { frame: 20, notification: Notification.createNext('a') },
+        { frame: 20, notification: Notification.createNext('b') },
+        { frame: 40, notification: Notification.createNext('c') },
+        { frame: 60, notification: Notification.createComplete() }
+      ]);
+    });
   });
 
   describe('parseMarblesAsSubscriptions()', () => {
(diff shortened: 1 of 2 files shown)


kwonoj:
I *think* this is correct representation and I made it in those way in rxSandbox(https://github.com/kwonoj/rx-sandbox#observable-marble-diagram-token-description) as breaking changes - but, it is also *breaking change* as you could see whole test case fails. 

I'm bit hesitant to change this behavior, cause anyone who uses `TestScheduler` implementation for their test suite will experience huge migration effort by having breaking changes. Best cases next major, but I'd probably just leave it and even rather deprecate testScheduler in v6 but separate test suite.

benlesh:
@kwonoj I assume moving the rx sandbox stuff in here will fix this?

kwonoj:
@benlesh yes.

chrmod:
if there is no way for the fix to landm, the bug should at least be documented somewhere. Guess the PR can be closed.

benlesh:
I'm closing this in favor of @kwonoj's work to move rx-sandbox into rxjs proper.

benlesh:
Thank @chrmod for bringing this to our attention, I'm surprised I hadn't noticed in the past.

PR #2974, opened 2017-10-18
Simplify webpack alias

jasonaden:
Using `__dirname` directly in the `path-mapping.js` file should be more reliable and should always be accurate. This simplifies setting up the aliases for Webpack to map to the `esm` distribution.

Output in the `path-mapping.js` file will now look like this:

```

"use strict"

var path = require('path');

module.exports = function() {
  return {
    "rxjs/AsyncSubject": path.resolve(__dirname, "AsyncSubject.js"),
    "rxjs/add/observable/interval": path.resolve(__dirname, "add/observable/interval.js"),
    "rxjs/util/tryCatch": path.resolve(__dirname, "util/tryCatch.js"),
    // ...
};
}
```

### DIFF
--- a/.make-packages.js
+++ b/.make-packages.js
@@ -138,9 +138,10 @@ function createImportTargets(importTargets, targetName, targetDirectory) {
 "use strict"
 
 var path = require('path');
+var dir = path.resolve(__dirname);
 
-module.exports = function(PATH_REPLACEMENT) {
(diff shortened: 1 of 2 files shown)


kwonoj:
Change itself looks fine. I'd personally would capture __dirname once

```
function() {
  var dir = path.resolve(__dirname);
  return {
    "rxjs/AsyncSubject": path.join(dir, "AsyncSubject.js"),
}
```

like, but not sure how much gain it'll be honestly. probably not worth it?

jasonaden:
@kwonoj Sounds good. I made that change.

PR #2978, opened 2017-10-19
docs(lettable-operator): recommend ts-import-plugin for tree-shaking

Brooooooklyn:

### DIFF
--- a/doc/lettable-operators.md
+++ b/doc/lettable-operators.md
@@ -219,6 +219,65 @@ var config = {
 module.exports = config;
 ```
 
+**Use ts-import-plugin**
+You can use [ts-import-plugin](https://www.npmjs.com/package/ts-import-plugin) to transform your import clause with ts-loader.
+For example, it can transform
+
+```ts
+import { map } from 'rxjs/operators'
+```
+
+into
+
+```ts
+import { map } from 'rxjs/operators/map'
+```
+
+here is an example config:
+
+```ts
+const { resolve } = require('path')
+const tsImportPluginFactory = require('ts-import-plugin')
+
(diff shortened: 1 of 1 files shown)


kwonoj:
I'm not sure we'd like to encourage specific plugin in official readme. /cc @jasonaden for visibility and assigning @benlesh for review, but personally downvote for this.

doc/lettable-operators.md
```diff
@@ -219,6 +219,65 @@ var config = {
 module.exports = config;
 ```
 
+**Use ts-import-plugin**
```
jasonaden: Add another line break after this.

doc/lettable-operators.md
```diff
@@ -219,6 +219,65 @@ var config = {
...
+You can use [ts-import-plugin](https://www.npmjs.com/package/ts-import-plugin) to transform your import clause with ts-loader.
+For example, it can transform
```
jasonaden: And another line break here.

kwonoj:
My concern here is not avout usefulness of plugin but doc implies it as sort of de facto recommended in official repo. For same reason we haven't added other 3rd party repo in doc links.

For me, this is more feasible for rxjs-docs' awesome-list we are trying to list modules for ecosystem.

benlesh:
Yeah. I think this is something where it would be good to have some blog posts on approaches for tree-shaking that we can recommend in some sort of 'awesome list' or the like. But I don't want us to have to maintain an up-to-date list of the best approaches to tree-shaking. We have enough on our plate.

I agree with @kwonoj. We should probably close this one.

Brooooooklyn:
oh thanks，maybe I will write a blog about this later😄

Issue #2972, opened 2017-10-18
Pipe operator cannot infer return type as ConnectableObservable


kasiditi:
**RxJS version:** 5.5.0

**Code to reproduce:**

```ts
const obs = Observable.of(5);
const connectableObs = obs.pipe(
    publishReplay(1)
);
```

**Expected behavior:** The inferred type of `connectableObs` is `ConnectableObservable<number>`.

**Actual behavior:** The inferred type of `connectableObs` is `Observable<number>`.

**Additional information:**

TypeScript version: 2.5.3

My current workaround is to manually downcast the return value:

```ts
const obs = Observable.of(5);
const connectableObs = obs.pipe(
    publishReplay(1)
) as ConnectableObservable<number>;
```

Happy to send a PR, but not sure how should I fix this properly.

cartant:
The problem can be solved by adding additional overload signatures to `pipe`, allowing for the last operator to return a `ConnectableObservable`.

For example:

```ts
pipe<A>(op1: UnaryFunction<Observable<T>, ConnectableObservable<A>>): ConnectableObservable<A>;
pipe<A>(op1: OperatorFunction<T, A>): Observable<A>;
```

Using `ConnectableObservable` in the signtures for a method in `Observable` seems a little weird, as `ConnectableObservable` extends `Observable`, but this is only way I can see this working.

An import of `ConnectableObservable` would be necessary in `Observable.ts`, but as it's only used in the typings, the import is erased during transpilation.

The signatures for `publish` and `publishLast` would have to be changed, too, as it they do not specify `ConnectableObservable` in their return types.

Interestingly, adding such overload signatures to `pipe` seems to fix another problem with the inference of the `publishReplay`. Without them, the return type of the following:

```ts
const o = of(0).pipe(publishReplay(1));
```

is inferred as `Observable<any>` (note the `any`) and with them it's inferred as `ConnectableObservable<number>`.

cartant:
Actually, if the overload signatures for `pipe` are written like this:

```ts
pipe<A, OA extends Observable<A>>(op1: UnaryFunction<Observable<T>, OA>): OA;
pipe<A, B, OB extends Observable<B>>(op1: OperatorFunction<T, A>, op2: UnaryFunction<Observable<A>, OB>): OB;
// etc.
```

The problem can be solved without specific reference to `ConnectableObservable`.

This seems to be okay with the version of TypeScript (2.0) that RxJS uses, but I've not investigated this thoroughly.

[...]

PR #2983, opened 2017-10-20
chore(typings): improve publish typings

cartant:
Use `ConnectableObservable` in signatures that do not have selectors.

**Description:**

This PR updates the overload signatures for lettable/pipeable `publish`, `publishLast` and `multicast` operators so that `ConnectableObservable` is specified where appropriate.

<s>I've not included `type` tests, as it's not apparent where they should be placed. That is, the lettable operators appear to be tested via the prototype-patched operators in `spec/operators`. If you'd like the typings tests for the lettables added in the same location, please let me know and I'll add them.</s>

The PR does not completely resolve the issue below, but it is directly related to it. And these changes would need to be made as the first step in resolving the issue.

**Related issue (if exists):** #2972


benlesh:
Tested this and I seems to work fine in TS 2.5.

PR #2986, opened 2017-10-21
docs: operators contribution: add pipe example

0xClpz:
Add an example of pipe usage for operators as pure functions in the doc about contributing on operators.

**Description:**
The doc for the pipe is still missing but I still fill like this might be useful.


benlesh:
Actually, I think we should just tell people to use pipe, and not let. So I'm going to close this PR. Sorry, @0xClpz... thank you SO much for your contribution, and I hope you contribute again in the future!

0xClpz:
@benlesh things moved quite a bit since that PR so it's all good 👍

PR #2987, opened 2017-10-22
refactor: move re-export file mapping to bottom of mapping file

jasonaden:
Needed due to Webpack issue where sequence of the "alias" object keys matters:

https://github.com/webpack/webpack/issues/5870


hansl:
Fixes https://github.com/angular/angular-cli/issues/8165

Issue #2991, opened 2017-10-24
rxjs fails to compile on Typescript 2.6


sandersn:
In src/operator/publishReplay.ts, the delegating call to `higherOrder` (which is an alias for `publishReplay` in src/operators/publishReplay.ts), has arguments that don't provide enough information to infer type arguments. That's because the third argument is cast to `any`, which is because it's just a delegation; a call to `publishReplay` could delegate to any of the three overloads of `higherOrder`. The compiler has to choose one, however, and it chooses the wrong one in Typescript 2.6.

**RxJS version:**
master

**Code to reproduce:**
Update typescript to `typescript@next` then try to `npm run build_all`

**Expected behavior:**
No compile error

**Actual behavior:**
Type `Observable<{}>` is not assignable to type `Observable<R> | ConnectableObservable<R>`

PR #2992, opened 2017-10-24
operator/publishReplay.ts:give type arguments to higherOrder call

sandersn:
**Description:**
In src/operator/publishReplay.ts, the delegating call to higherOrder (which is an alias for publishReplay in src/operators/publishReplay.ts), has arguments that don't provide enough information to infer type arguments. That's because the third argument is cast to any, which is because it's just a delegation; a call to publishReplay could delegate to any of the three overloads of higherOrder. The compiler has to choose one, however, and it chooses the wrong one in Typescript 2.6.

This change provides the type arguments explicitly so that the most general overload gets called. This results in the the correct return type, `OperatorFunction<T, R>`.

#2991


sandersn:
I couldn't use the commit-hook because it failed to find git.exe on my ancient windows laptop I use at home. I can update the commit message or produce an alternate PR tomorrow when I get back to my linux desktop at work.

PR #2998, opened 2017-10-25
fix(build): fix make_packages to not error when running build_test

jasonaden:
When running `build_test` we don't create the ESM sources. `.make_packages.js` was looking for ESM sources in all cases. This change doesn't error when the ESM sources don't exist.


kwonoj:
We just need https://github.com/ReactiveX/rxjs/pull/2851 checked in and test no longer tries to invoke make_pkg or others. build_test simply can go away.

benlesh:
Closed in favor of #2851

PR #2999, opened 2017-10-25
fix(concatStatic): missing export creating breaking change

ghetolay:
**Description:**

Export of `concatStatic` from `operator/concat` was removed on 5.5.
While I could easily replace : 
```ts
import { concatStatic } from 'rxjs/operators/concat';
```
with
```ts
import { concat } from 'rxjs/observable/concat';
```
 this is still a breaking change that could affect other projects.

### DIFF
--- a/src/observable/merge.ts
+++ b/src/observable/merge.ts
@@ -1,3 +1,101 @@
-import {  mergeStatic  } from '../operator/merge';
+import { Observable, ObservableInput } from '../Observable';
+import { IScheduler } from '../Scheduler';
+import { ArrayObservable } from './ArrayObservable';
+import { isScheduler } from '../util/isScheduler';
+import { mergeAll } from '../operators/mergeAll';
 
-export const merge = mergeStatic;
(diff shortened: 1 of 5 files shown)


kwonoj:
This seems makes sense and we defer this change for v6 as breaking?

ghetolay:
I looked a bit more on the code and this add inconsistency between how mergeStatic and concatStatic are handled.
mergeStatic is defined on operators/merge and reexported on both operator/merge and observable/merge while here concatStatic is defined on observable/concat, reexported on operator/concat and nothing inside operators/concat.

I was about to defined both inside observable/* and only reexport on operator/* for breaking change. But now that merge was also exported on operators/* in 5.5, removing it would create a new breaking change :)
So I'll just define them both inside observable/* and reexport on both operator and operators.
Will do that tomorrow in need some sleep right now :sleeping: .

ghetolay:
I think this is good to go.
We'll just have to remove the exports on `operator/*` and `operators/*` on v6.

PR #3013, opened 2017-10-27
Removing tests that don't make any sense

benlesh:
We had a few slip past the radar on us that weren't really testing anything of value. See the commit messages for more detail.

### DIFF
--- a/spec/operators/partition-spec.ts
+++ b/spec/operators/partition-spec.ts
@@ -255,11 +255,6 @@ describe('Observable.prototype.partition', () => {
     expectSubscriptions(e1.subscriptions).toBe([e1subs, e1subs]);
   });
 
-  it('should throw without predicate', () => {
-    const e1 = hot('--a-b---a------d----');
-    expect(e1.partition).to.throw();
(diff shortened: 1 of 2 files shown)


spec/operators/partition-spec.ts
```diff
@@ -255,11 +255,6 @@ describe('Observable.prototype.partition', () => {
     expectSubscriptions(e1.subscriptions).toBe([e1subs, e1subs]);
   });
 
-  it('should throw without predicate', () => {
-    const e1 = hot('--a-b---a------d----');
-    expect(e1.partition).to.throw();
```
kwonoj: shouldn't we just correct it into `expect(() => e1.partition.subscribe)).to.throw?`

benlesh: I'm not sure it should throw... it doesn't throw now and none of our other methods throw... for example, if you don't provide a predicate to `filter`.

Perhaps they *should* throw... but it seems unrelated.

kwonoj:
🚢

PR #3019, opened 2017-10-30
build(tsconfig): introduce extended tsconfig for target

kwonoj:
**Description:**
This PR is effort to simplify our npm script, as first step instead of repeating whole same configuration for each build targes, utilizes latest tsc's config extending.


kwonoj:
Yes, this is targeting master only which for v6 and further. Ben branched off 5.5 stable for minor fixes.

benlesh:
This is SO MUCH BETTER.

PR #3022, opened 2017-10-30
fix(TSC): Fixing TSC errors. Fixes #3020

### LINKED ISSUE #3020: Resolve compiler errors in test cases by implicit any
**RxJS version:**

**Code to reproduce:**

**Expected behavior:**

**Actual behavior:**

**Additional information:**
With latest PR https://github.com/ReactiveX/rxjs/pull/3019, running `tsc` on top of repo with `tsconfig.json` will bring up whole bloat of compilation error by our spec has numbers of implicit any declaration. This is dept we need to pay off and enable those compilation in CI to prevent it happens again.

davidkpiano:
**Description:** This fixes all the warnings output by running `tsc` at the project's root by adding types for `node` and `mocha`, and also fixing a TS4023 issue.

**Related issue (if exists):** https://github.com/ReactiveX/rxjs/issues/3020

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -16,7 +16,8 @@
       "es2015.collection",
       "es2015.promise",
       "dom"
-    ]
+    ],
+    "types": ["mocha", "node"]
   },
   "formatCodeOptions": {
     "indentSize": 2,
--- a/spec/helpers/test-helper.ts
+++ b/spec/helpers/test-helper.ts
@@ -24,25 +24,27 @@ export function lowerCaseO<T>(...args: Array<any>): Rx.Observable<T> {
   return <any>o;
 };
 
-export const createObservableInputs = <T>(value: T) => Rx.Observable.of<ObservableInput<T>>(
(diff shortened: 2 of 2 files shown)


tsconfig.json
```diff
@@ -16,10 +16,11 @@
       "es2015.collection",
       "es2015.promise",
       "dom"
-    ]
+    ],
+    "types": ["mocha", "node"]
```
kwonoj: my other codebases does have types for these globals but didn't require this explicitly 🤔 any idea why it's needed?

davidkpiano: Probably because you were `declare`-ing things like `describe()` and `it()` somewhere else; but hey, this seems like the right thing to do and it fixes the errors so... yeah, LGTY?

kwonoj:
@davidkpiano I tried your PR with merged, and it actually doesn't work - I'm reverting currently merged one. You may able to try run tsc directly (not via npm script) on root of repo without any option to see what happens.

PR #3026, opened 2017-10-30
build: fix pointer to Rx.d.ts file

### LINKED ISSUE #3005: Could not find a declaration file for module '@reactivex/rxjs' (5.5.x)
**RxJS version:**
5.5.0, 5.5.1, 5.5.2

**Code to reproduce:**
*index.ts*
```typescript
import { Subject } from "@reactivex/rxjs";

const test = new Subject<boolean>();

test.next(true);
```

*package.json*
```json
{
  "dependencies": {
    "@reactivex/rxjs": "5.5.0",
    "typescript": "^2.5.3"
  }
}

*tsconfig.json*
```json
{
    "compilerOptions": {
        "noImplicitAny": true
    }
}
```

**Expected behavior:**
No errors.

**Actual behavior:**
> index.ts(1,25): error TS7016: Could not find a declaration file for module '@reactivex/rxjs'. 'C:/Users/mattd/Source/Examples/rxjs-bug/node_modules/@reactivex/rxjs/index.js' implicitly has an 'any' type.
>  Try `npm install @types/@reactivex/rxjs` if it exists or add a new declaration (.d.ts) file containing `declare module '@reactivex/rxjs';`

**Additional information:**
The `package.json` of rxjs contains `"typings": "./dist/package/typings/Rx.d.ts"`, but there is no `typings` folder in the `package` directory. I expect that it should be changed to `"typings": "./dist/typings/Rx.d.ts"`, because that appears to work.

jasonaden:
Fixes #3005

### DIFF
--- a/package.json
+++ b/package.json
@@ -216,7 +216,7 @@
   "engines": {
     "npm": ">=2.0.0"
   },
-  "typings": "./dist/package/typings/Rx.d.ts",
+  "typings": "./dist/package/Rx.d.ts",
   "dependencies": {
     "symbol-observable": "^1.0.1"
   }


benlesh:
Something crazy is going on with this build... @jasonaden, it looks like typings are off or something.

jasonaden:
@benlesh I'm pushing a new commit hash to get Travis to re-run. I can't replicate this issue locally. Looks like something was messed up with Lodash's typings at time of install on Travis.

jasonaden:
So for some reason when we run the build on Travis, we get a `node_modules/@types/lodash` directory. We don't get that directory when running locally.

Locally I can reproduce these errors if I do `npm install @types/lodash`, then run `npm run build_all` because TypeScript ends up pulling in the `node_modules/@types/lodash/index.d.ts` file. That file is incompatible with the version of TypeScript we end up on (`2.0.10`).

@benlesh @kwonoj Any ideas?

kwonoj:
@jasonaden seems it's just travis caching issue. I've busted cache and retriggered.

jasonaden:
Thanks @kwonoj.

@benlesh this should be good to go.

Issue #3034, opened 2017-11-01
Unsubscription doesn't work after shareReplay?!


ivan-kleshnin:
**RxJS version: 5.5.2**

**Code to reproduce:**

```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/rxjs/5.5.2/Rx.min.js"></script>

<button id="foo">Foo</button>
<button id="bar">Bar</button>

<script>
  let {Observable: O} = Rx

  let fooNode = document.querySelector("#foo")
  let barNode = document.querySelector("#bar")

  let fooClick$ = O.fromEvent(fooNode, "click")
    .do(() => { console.log("foo clicked!") })

  let barClick$ = O.fromEvent(barNode, "click")
    .do(() => { console.log("bar clicked!") }).shareReplay(1) // !!! 

  let fooSb = fooClick$.subscribe()
  let barSb = barClick$.subscribe() // !!!

  setTimeout(() => {
    fooSb.unsubscribe()
  }, 2000)

  setTimeout(() => {
    barSb.unsubscribe() // !!!
  }, 2000)
</script>
```

**Expected behavior:**

Both buttons stop working after two seconds.

**Actual behavior:**

Only the first button stop working after two seconds.

cartant:
I'm unable to reproduce this with the script that's in the issue.

Well after two seconds has elapsed, clicking on the `Bar` button effects the logging of `bar clicked!`. The subscription to the source observable does not appear to be unsubscribed when the subscription to the shared observable is unsubscribed.

ivan-kleshnin:
> Well after two seconds has elapsed, clicking on the Bar button effects the logging of bar clicked!. 

Yes, that's a problem.

> The subscription to the source observable does not appear to be unsubscribed when the subscription to the shared observable is unsubscribed.

Yes and I believe it's a bug because it produces unremovable event listener (which is inappropriate in any scenario).

Documentation on the `refCount` says:

> Returns an observable sequence that **stays connected to the source as long as there is at least one subscription to the observable sequence**.

*This is from v4 (can't find the same doc piece for v5, but that didn't change I guess).*

In my example we have 0 subscriptions after 2 second and `fromEvent` is still connected to the source.

cartant:
According to [this PR](https://github.com/ReactiveX/rxjs/pull/2910), what you are seeing is the intended behaviour.

ivan-kleshnin:
[...]

PR #3043, opened 2017-11-04
docs(MIGRATION): remove shareReplay

cartant:
**Description:**

Remove `shareReplay` from the **Operators Renamed or Removed** list in `MIGRATION.md`, as it was implemented in [5.4.0](https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md#540-2017-05-09) and its listing is causing confusion.

**Related issue (if exists):** #3034


benlesh:
Thank you @cartant

Issue #3042, opened 2017-11-03
Observable.interval in inactive tab work different than plain js setInterval()


nikshik:
Observable.interval in inactive tab work different than plain js setInterval()

Open http://jsbin.com/muvaqipaqa/1/edit?html,js,console,output
See 'i:0', 'j:0', ... in console. Switch to another browser tab for 30 seconds or so. Switch back to jsbin tab. See that i have much more different value than j

**RxJS version:**
5.5.2

**Code to reproduce:**
http://jsbin.com/muvaqipaqa/1/edit?html,js,console,output

**Expected behavior:**
i and j have same values
**Actual behavior:**
i and j have different values

kwonoj:
you can't expect interval observable works identical to setInterval as interval observable have scheduling to create inverval values.

this seems browser implementation does try to prevent js execution in background for inactive tabs. I'm not sure devtools allows to disable it - if it's possible, please check with it to confirm.

nikshik:
Rly? So if Observable.interval will be called after 1500 ms instead of expected 1000 ms this is fine?
You mean I should not expect Observable.interval to work correctly in browsers? 
I'm pretty sure this is an issue with rxjs implementation...

nikshik:
If you check it in jsbin with RxJs 5.0.3 it works as expected, in 5.5.2 it doesn't!
Please reopen issue

kwonoj:
hmm, yeah that sounds odd. I thought it's happening all time.

kwonoj:
Reopened.

benlesh:
You need to be careful relying on setInterval, setTimeout, etc in JSBin, I've seen some odd behavior there, and I think it's being patched or hooked in some way by JSBin.

nikshik:
Thanks for the hint! I will pay attention to it.
Anyway I initially found the issue on my local environment, just added example to jsbin.

PR #3044, opened 2017-11-04
fix(scheduler): prevent unwanted clearInterval

cartant:
**Description:**

In `AsyncAction`, `this.pending` was assigned before the call to `recycleAsyncId`. That prevented the interval from being re-used resulting in the interval being cleared and re-created for each notification.

<s>I cannot see how a test could be easily written for this, but the logic is straightforward.</s>

**Related issue (if exists):** #3042

### DIFF
--- a/spec/schedulers/AsapScheduler-spec.ts
+++ b/spec/schedulers/AsapScheduler-spec.ts
@@ -40,6 +40,57 @@ describe('Scheduler.asap', () => {
     sandbox.restore();
   });
 
+  it('should reuse the interval for recursively scheduled actions with the same delay', () => {
+    const sandbox = sinon.sandbox.create();
+    const fakeTimer = sandbox.useFakeTimers();
+    const stubSetInterval = sinon.stub(global, 'setInterval').callThrough();
+    function dispatch(state: any): void {
+      state.index += 1;
+      if (state.index < 3) {
+        (<any> this).schedule(state, state.period);
+      }
+    }
+    const period = 50;
+    const state = { index: 0, period };
+    asap.schedule(dispatch, period, state);
+    expect(state).to.have.property('index', 0);
+    expect(stubSetInterval).to.have.property('callCount', 1);
+    fakeTimer.tick(period);
+    expect(state).to.have.property('index', 1);
+    expect(stubSetInterval).to.have.property('callCount', 1);
+    fakeTimer.tick(period);
+    expect(state).to.have.property('index', 2);
+    expect(stubSetInterval).to.have.property('callCount', 1);
+    stubSetInterval.restore();
+    sandbox.restore();
+  });
+
+  it('should not reuse the interval for recursively scheduled actions with a different delay', () => {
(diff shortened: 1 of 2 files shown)


benlesh:
interesting. @cartant, can you please add a test?

cartant:
@benlesh Added two tests.

spec/schedulers/AsapScheduler-spec.ts
```diff
@@ -40,6 +40,57 @@ describe('Scheduler.asap', () => {
     sandbox.restore();
   });
 
+  it('should recycle the interval for recursively scheduled actions with the same delay', () => {
```
benlesh: Nit: This should probably say "reuse" rather than "recycle"... To me, "recycle" means you're tearing it down and creating another. Which is the opposite of the meaning here.

benlesh:
I'm going to have this target the 5.5 branch and cherry pick it over. We can do a patch.

benlesh:
Actually, @cartant, do you think you can rebase your changes against `stable` rather than `master`? I'd like to get this out in a patch to 5.5.

cartant:
@benlesh Rebased and tests renamed as per request.

benlesh:
Hmmm... it looks like there are some TS 2.0 incompatible lines in this already... I'm not sure I can merge this into the stable branch. :(

cartant:
Are you talking about this:

```
46     const stubSetInterval = sinon.stub(global, 'setInterval').callThrough();
                                                                 ~~~~~~~~~~~

spec/schedulers/AsapScheduler-spec.ts(46,63): error TS2339: Property 'callThrough' does not exist on type 'SinonStub'.


71     const stubSetInterval = sinon.stub(global, 'setInterval').callThrough();
                                                                 ~~~~~~~~~~~

spec/schedulers/AsapScheduler-spec.ts(71,63): error TS2339: Property 'callThrough' does not exist on type 'SinonStub'.
```

The `sinon` version in the `package.json` is the same for `stable` and `next`, but `stable` uses  `typings.json` - rather than `@types` and the typings in the `typings.json` for `sinon` look old:

```
"sinon": "registry:npm/sinon#1.16.0+20160723033700"
```

cartant:
I can make the tests build on a clean `node_modules` install for `stable` if I assert to `any` to avoid the missing TS declaration for `callThrough` in the old `sinon` typings used to `stable`:

```ts
const stubSetInterval = (<any> sinon.stub(global, 'setInterval')).callThrough();
```

If you're okay with that, I can update the PR.

cartant:
@benlesh I've made the change mentioned in the above comment and my branch's Travis build now passes - although, I had to clear the cache.

trxcllnt:
LGTM 👍

benlesh:
Yeah. I'm not sure what's going on with Travis, and I don't have time to take a look. @cartant or @kwonoj, can one of you figure this out?

kwonoj:
@benlesh should be good now.

PR #3045, opened 2017-11-04
refactor(interfaces): replace type aliases

cartant:
**Description:**

Regarding [this issue](https://github.com/Microsoft/TypeScript/issues/19198) - that was opened in the TypeScript repo - there is a simple resolution: replace the type aliases with interfaces.

The TypeScript documentation for [advanced types](https://www.typescriptlang.org/docs/handbook/advanced-types.html) includes these comments in the **Type Aliases** section:

> One difference is that interfaces create a new name that is used everywhere. Type aliases don’t create a new name — for instance, error messages won’t use the alias name. [...]
>
> A second more important difference is that type aliases cannot be extended or implemented from (nor can they extend/implement other types). Because an ideal property of software is being open to extension, you should always use an interface over a type alias if possible.

I can see that the issue has been labelled as a bug, but the referenced documentation suggests otherwise.

cc/ @jayphelps @benlesh 

**Related issue (if exists):** https://github.com/Microsoft/TypeScript/issues/19198

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -1,9 +1,9 @@
 import { Observable } from './Observable';
 
-export type UnaryFunction<T, R> = (source: T) => R;
+export interface UnaryFunction<T, R> { (source: T): R; }
 
-export type OperatorFunction<T, R> = UnaryFunction<Observable<T>, Observable<R>>;
+export interface OperatorFunction<T, R> extends UnaryFunction<Observable<T>, Observable<R>> {}
 
 export type FactoryOrValue<T> = T | (() => T);
 
-export type MonoTypeOperatorFunction<T> = OperatorFunction<T, T>;
+export interface MonoTypeOperatorFunction<T> extends OperatorFunction<T, T> {}


jayphelps:
Seems like a good solution at first glance.

cartant:
TIL: documentation can be out-of-date.

Anyway, with the interfaces, you could have it working now, instead of in January (maybe) with TS 2.7.

jayphelps:
Totes. Btw rereading my last comment it might have seemed like I meant “at first glance” to mean that it’s not a good solution after all. Whoops! Not what I meant. I just meant that I haven’t tested it or seen if it has any other impacts in real usage. Someone on the core team will review hopefully.

cartant:
No worries. I took it the way you intended it.

jayphelps:
I started to think this would be a breaking change because `OperatorFunction` is exposed publicly, but ~~unfortunately :trollface:~~ fortunately because TypeScript is structurally typed instead of nominally, it doesn't care:

[Playground Example](https://www.typescriptlang.org/play/index.html#src=declare%20class%20Observable%3CT%3E%20%7B%20%7D%0D%0A%0D%0Atype%20UnaryFunction_alias%3CT%2C%20R%3E%20%3D%20(source%3A%20T)%20%3D%3E%20R%3B%0D%0Atype%20OperatorFunction_alias%3CT%2C%20R%3E%20%3D%20UnaryFunction_alias%3CObservable%3CT%3E%2C%20Observable%3CR%3E%3E%3B%0D%0A%0D%0Ainterface%20UnaryFunction_interface%3CT%2C%20R%3E%20%7B%20(source%3A%20T)%3A%20R%3B%20%7D%0D%0Ainterface%20OperatorFunction_interface%3CT%2C%20R%3E%20extends%20UnaryFunction_interface%3CObservable%3CT%3E%2C%20Observable%3CR%3E%3E%20%7B%20%7D%0D%0A%0D%0Adeclare%20let%20usingAlias%3A%20OperatorFunction_alias%3Cvoid%2C%20void%3E%3B%0D%0Adeclare%20let%20usingInterface%3A%20OperatorFunction_interface%3Cvoid%2C%20void%3E%3B%0D%0A%0D%0A%2F%2F%20no%20type%20error%0D%0AusingAlias%20%3D%20usingInterface%3B%0D%0A%0D%0A)

```ts
// as an alias
type UnaryFunction_alias<T, R> = (source: T) => R;
type OperatorFunction_alias<T, R> = UnaryFunction_alias<Observable<T>, Observable<R>>;

// as an interface
interface UnaryFunction_interface<T, R> { (source: T): R; }
interface OperatorFunction_interface<T, R> extends UnaryFunction_interface<Observable<T>, Observable<R>> { }

declare let usingAlias: OperatorFunction_alias<void, void>;
declare let usingInterface: OperatorFunction_interface<void, void>;

// no type error
usingAlias = usingInterface;
```

It would have been _very_ unlikely anyone depended on it anyway.

benlesh:
Interestingly, this is how I first implemented this when working on it, but switched it because I thought it was more readable for other contributors. (interfaces for functions is pretty weird, IMO, but probably a personal preference thing)

If this solves a problem that's blocking you @jayphelps or the documentation efforts in general, I'll happily merge.

benlesh:
Also, if @jayphelps says this solves a problem for our docs, we should add a test to make sure we don't break it in the future.

cartant:
This is another one that could use a snippet test - to ensure the inferred type is what's expected.

Issue #3048, opened 2017-11-05
fromEventPattern does not pass multiple arguments from event callbacks


raycarter:
**RxJS version:**
5.5.2
**Code to reproduce:**
```javascript
const EventEmitter = require('events').EventEmitter;
const Observable = require('rxjs').Observable;

const emitter = new EventEmitter();

const obs = Observable.fromEventPattern(function(handler) {
  emitter.on('test', handler);
});

obs.subscribe(args => {
  console.log(args); // 'a', expected ['a', 'b', 'c']
});

emitter.emit('test', 'a', 'b', 'c');
```
**Expected behavior:**
A group of arguments is expected in `subscribe` (here expected ['a', 'b', 'c']).
**Actual behavior:**
Only the first argument is emtted (here 'a').
**Additional information:**
I think it could be common to create Observables from events and a lot of events pass more than one arguments in its callback functions. It is useful to let rxjs collect all of the arguments and pass them to `subscribe`. If I am wrong here about how to use rxjs in this situation, please give me a hint how I should do it correctly or which other operators should be used for events with multiple arguments. Thanks a lot!

kwonoj:
dupe of https://github.com/ReactiveX/rxjs/issues/2431.

Issue #2431, opened 2017-03-01
subscribe ignores second argument of event emitter


benjaminhon:
I am using redis, which exposes the event emitter as follows with 2 arguments, without rxjs, i can get both arguments, however with rxjs, the second argument is missing.

```
r.on 'message', (channel, message) ->
  console.log channel, message

# => Greet, Hello

Rx.Observable.fromEvent(r, 'message').subscribe (channel, message) ->
  console.log channel, message

# => Greet, undefined
```

is this supposed to be the case?

Podlas29:
That is how it works at the moment. You can get access to the second parameter by using project function:

```javascript
Rx.Observable.fromEvent(r, 'message', (channel, message) => [channel, message])
.subscribe(arr => {
  console.log(arr[0]); // logs value of channel
  console.log(arr[1]); // logs value of message
});
```

That being said it seems like a strong case to make `fromEvent` behave like `bindCallback` which emits an array if there are many arguments in callback.

Podlas29:
Does redis have an `off` function for canceling subscriptions? How do you stop listening to events?

benjaminhon:
yea it has,

```

sub.unsubscribe()

```

Podlas29:
When you unsubscribe `fromEvent` it will try to call `off` function on redis. Since (as I understand) it is not there, your code will crash.
Use `fromEventPattern` which allows you to customize how callback is registered and unregistered.

I have made prs for refined docs for these two operators, if you still have trouble figuring out which does what: https://github.com/ReactiveX/rxjs/pull/2411 
https://github.com/ReactiveX/rxjs/pull/2416
If these are helpful or if they still need some clarifications, please leave some comment in a pr. ;)

trxcllnt:
We should probably update `fromEvent` to work with EventEmitters too, as I constantly have this problem as well. From what I remember, EventEmitters have `on` and `removeListener` methods. Does that sound right?

kwonoj:
> , EventEmitters have on and removeListener methods. 

doesn't node `evemtEmitter` works already via `addListener`  (https://nodejs.org/api/events.html#events_emitter_addlistener_eventname_listener)  /`removeListener` ?(https://github.com/ReactiveX/rxjs/blob/7c41c08119cecb0c039ed40b2e1f3a0e6ac0d56e/src/observable/FromEventObservable.ts#L10-L13)

benjaminhon:
@Podlas29 that does the job

[...]

PR #3049, opened 2017-11-05
fix(fromEventPattern): pass multiple arguments of event emitters to subscribe

raycarter:
**Description:**
The situation is like the following code:
```javascript
const EventEmitter = require('events').EventEmitter;
const Observable = require('rxjs').Observable;

const emitter = new EventEmitter();

const obs = Observable.fromEventPattern(function(handler) {
  emitter.on('test', handler);
});

obs.subscribe(args => {
  console.log(args); // expected ['a', 'b', 'c']
});

emitter.emit('test', 'a', 'b', 'c');

```
If multiple arguments are emitted from an event, only the first will be passed to `subscribe`. This PR changes it to pass an array of arguments to `subscribe`. A test case is added to `spec.ts`.

**Related issue (#3048)**

### DIFF
--- a/spec/observables/fromEventPattern-spec.ts
+++ b/spec/observables/fromEventPattern-spec.ts
@@ -133,4 +133,28 @@ describe('Observable.fromEventPattern', () => {
 
     trigger('test');
   });
+
+  it('should accept and pass multiple parameters from an event emitter', (done: MochaDone) => {
+    let target: Function = () => '';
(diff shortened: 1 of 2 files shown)


kwonoj:
I'm going to close this PR for now as dupe of https://github.com/ReactiveX/rxjs/issues/2431, on discussion. I personally do not see huge need of this as selector argument allows this behavior.

Issue #3031, opened 2017-11-01
Error in project with TS 2.6 and strictFunctionTypes


jgoz:
**RxJS version:** 5.5.2

**Code to reproduce:**

In `tsconfig.json`:

```json
{
  "compilerOptions": {
    "skipLibCheck": false,
    "strictFunctionTypes": true
  }
}
```

```ts
import * as Rx from "rxjs";
```

**Expected behavior:**

Compiles successfully.

**Actual behavior:**

Compiler error:

```
../../node_modules/rxjs/scheduler/VirtualTimeScheduler.d.ts(22,22): error TS2415: Class 'VirtualAction<T>' incorrectly extends base class 'AsyncAction<T>'.
  Types of property 'work' are incompatible.
    Type '(this: VirtualAction<T>, state?: T | undefined) => void' is not assignable to type '(this: AsyncAction<T>, state?: T | undefined) => void'.
      The 'this' types of each signature are incompatible.
        Type 'AsyncAction<T>' is not assignable to type 'VirtualAction<T>'.
          Property 'index' is missing in type 'AsyncAction<T>'
```

**Additional information:**

Workaround: disable `strictFunctionTypes` OR enable `skipLibCheck`

I believe this is happening because `work` is not defined as a method (rather, it's defined as a property), so its parameters are checked contravariantly. If it was defined as a method, the parameters would be [checked covariantly](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#strict-function-types).

This could be fixed by defining `work` as a method in both classes, or defining it as `(this: AsyncAction<T>, state?: T) => void` in `VirtualAction`.

kentrino:
same issue

archanid:
I guess really we want `(this: AsyncAction<T>, state?: T | undefined) => void` ?

laurilehmijoki:
We're experiencing the problem as well.

brycesteinhoff:
I'm also running into this issue.

lednhatkhanh:
The same problem for me! Any updates on this?

kentrino:
the same issue.

amir-arad:
same issue

amir-arad:
also, even after using `"skipLibCheck": false`, when using `Observable.fromEvent()` the argument given, which extends node's event emitter, does not match `NodeStyleEventEmitter`...

MoAsmar:
same issue here, can you please tell me what are the steps to fix ?

I updraded to
[...]

PR #3050, opened 2017-11-05
refactor(types): enable strict fntype compiler option

kwonoj:
**Description:**
This PR enables `strictFnType` in master branch resolves build error with 2.6 version of tsc. (related with https://github.com/ReactiveX/rxjs/issues/3031) . due to our inheritance structure, there are some types are actually loosened (mostly around `Action`'s `work` types). 

This is targeting master intentionally due to scope of change is relatively large, include (probably) unnecessary part to close issue itself. Some change need to be cherrypicked into `stable` though.

### DIFF
--- a/src/internal/observable/BoundCallbackObservable.ts
+++ b/src/internal/observable/BoundCallbackObservable.ts
@@ -172,7 +172,7 @@ export class BoundCallbackObservable<T> extends Observable<T> {
   static create<T>(func: Function,
                    selector: Function | void = undefined,
                    scheduler?: IScheduler): (...args: any[]) => Observable<T> {
-    return function(this: any, ...args: any[]): Observable<T> {
+    return function (this: any, ...args: any[]): Observable<T> {
       return new BoundCallbackObservable<T>(func, <any>selector, args, this, scheduler);
     };
   }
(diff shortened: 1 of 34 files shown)


bondz:
I understand eveyone is busy, but can we get this reviewed early? I understand there's a workaround but its suboptimal.

kwonoj:
I forgot this :/ and I'm sure this'll require to resolve merge conflict with https://github.com/ReactiveX/rxjs/pull/3145.

Anyway, this is for v6 branch and not for stable, for 5.x probably we need small scoped change only resolve issue.

bondz:
Oh, okay, thanks for clarifying.

samosaara:
Well the new year is here, and as I see it the #3145 that is causing the conflicts only moves files around, so what is needed to do is move the files this PR changed into the new locations? I really agree this should be also back-ported into 5.x, this issue breaks rxjs completely in typescript 2.6.1

kwonoj:
uh-oh, this PR itself even have huge conflict due to recent changes. I need to rebase this first, then picks some into 5.5

benlesh:
:shipit: Good add, @kwonoj!

PR #3053, opened 2017-11-06
fix(typings): fix subscribe overloads

### LINKED ISSUE #3052: Incorrect type checking in catch operator
**RxJS version:**
RXJS Version: 5.1.0

**Code to reproduce:**
```ts
interface Action = {
    type: string;
}

function loadTasks(){
    return this.service.load()
      .map<any[],Action>(_ => ({type:"TASKS_LOADED"}))
      .catch<Action,Action>((e: any) => Observable.of({broken:"True"}))
}
```

**Expected behavior:**
Given code should throw an error as `{broken:"True"}` cannot be converted to an `Action`

**Actual behavior:**
No Errors, code compiles fine

**Additional information:**
If I change the code to:
```ts
function loadTasks(){
    return this.service.load()
      .map<any[],Action>(_ => ({type:"TASKS_LOADED"}))
      .catch<Action,Action>((e: any) => Observable.of<Action>({broken:"True"}))
}
```
I get the expected error.

cartant:
Remove the no-arg overload. If not removed, any Observable will be compatible with any ObservableInput regardless of type - as T does not appear in the no-arg overload.

Closes #3052

**Description:**

The overload signatures for `subscribe` include a signature that takes no parameters. This is a problem and results in any `Observable` being deemed compatible with any `ObservableInput`.

For example, this code will compile without error:

```ts
import { Observable, ObservableInput } from "rxjs/Observable";
import "rxjs/add/observable/of";

const numbers = Observable.of<number>(1);
const strings = Observable.of<string>("1");

let input: ObservableInput<string>;
input = numbers;
input = strings;
```

The problem is caused by `T` not appearing in the signature. The signature should be removed and the `observer` in the subsequent signature should be made optional. With that change, the compilation of the above code fails with this error:

```
error TS2322: Type 'Observable<number>' is not assignable to type 'ObservableInput<string>'.
  Type 'Observable<number>' is not assignable to type 'ArrayLike<string>'.
    Property 'length' is missing in type 'Observable<number>'.
```

**Related issue (if exists):** #3052

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -73,8 +73,7 @@ export class Observable<T> implements Subscribable<T> {
     return observable;
   }
 
-  subscribe(): Subscription;
-  subscribe(observer: PartialObserver<T>): Subscription;
+  subscribe(observer?: PartialObserver<T>): Subscription;
   subscribe(next?: (value: T) => void, error?: (error: any) => void, complete?: () => void): Subscription;
   /**
    * Invokes an execution of an Observable and registers Observer handlers for notifications it will emit.


benlesh:
Can we add typings tests for this?

cartant:
We can, but I don't think they'll test anything useful.

What would be much better would be to add some snippet-based testing that could test for failures. It really needs a test to ensure that an `Observable<string>` cannot be assigned to an `ObservableInput<number>`. That's not possible with the basic `type` test mechanism, as the entire compilation will fail.

Some snippet-based testing would be useful for testing other overloads, too. I discussed this with OJ in another issue - from [this comment](https://github.com/ReactiveX/rxjs/pull/2855#issuecomment-333751365) onwards.

For another example of what the snippet tests would look like, have a look at [`research-spec.ts`](https://github.com/cartant/ts-action/blob/master/source/research-spec.ts) in my `ts-action` repo.

benlesh:
Okay... LGTM. I think that we should have a longer conversation about type testing. Perhaps with the TypeScript folks. I'm very interested in your solution, but it seems like TypeScript should provide more in this area. cc @david-driscoll

benlesh:
@kwonoj do you think we should merge these into stable and cherry pick them over to master?

benlesh:
@cartant can you please rebase this against `stable` rather than `master` so we get get it out in the next patch?

cartant:
@benlesh Rebased.

cartant:
@benlesh
> it seems like TypeScript should provide more in this area

The library is a fairly thin wrapper around TypeScript's [Language Service API](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API#incremental-build-support-using-the-language-services). In fact, it's pretty much based on the example in the referenced documentation. There are other snippet compiler libraries/tools out there, but I wasn't able to find one that supported importing modules.

kwonoj:
@benlesh yeah, seems non breaking change to me go for stable.

benlesh:
Travis is failing for some reason on this can we figure out why? I'm going to rerun it. I'm looking at this from my phone so I couldn't really see the logs.

cartant:
It appears to be breaking because the the types that are installed for `lodash`:

```
4256             ...props: Array<Many<keyof T>>
                                            ~
node_modules/@types/lodash/index.d.ts(4256,40): error TS1005: '>' expected.
```

`keyof` was introduced in TypeScript 2.1 and the stable `package.json` uses TypeScript `"~2.0.6"`.

I was able to get the tests to pass on Travis on my branch by clearing Travis's cache. How it gets into the state that it's currently in, I'm not sure. But it does seem cache-related.

kwonoj:
flushed cache and retriggered build, also created https://github.com/ReactiveX/rxjs/pull/3132 for better module cache.

PR #3062, opened 2017-11-10
feat(Observable): unhandled errors are now reported to HostReportErrors

benlesh:
BREAKING CHANGE: Unhandled errors are no longer caught and rethrown, rather they are caught and scheduled to be thrown, which causes them to be reported to window.onerror or process.on('error'), depending on the environment. Consequently, teardown after a synchronous, unhandled, error will no longer occur, as the teardown would not exist, and producer interference cannot occur


kwonoj:
Will check this soon.

src/Subscriber.ts
```diff
@@ -199,80 +191,49 @@ class SafeSubscriber<T> extends Subscriber<T> {
...
     if (!this.isStopped && this._next) {
-      const { _parentSubscriber } = this;
-      if (!_parentSubscriber.syncErrorThrowable) {
-        this.__tryOrUnsub(this._next, value);
-      } else if (this.__tryOrSetError(_parentSubscriber, this._next, value)) {
+      try {
+        this._next.call(this._context, value);
+      } catch (err) {
         this.unsubscribe();
+        this._hostReportError(err);
```
kwonoj: not sure if this need to be aligned - cases like error / complete we do report error first then unsubscribe, for next we do unsubscribe first then report. Nitpick maybe?

src/operators/tap.ts
```diff
@@ -74,49 +76,62 @@ class DoOperator<T> implements Operator<T, T> {
...
-
-    const safeSubscriber = new Subscriber<T>(nextOrObserver, error, complete);
-    safeSubscriber.syncErrorThrowable = true;
-    this.add(safeSubscriber);
-    this.safeSubscriber = safeSubscriber;
-  }
+      super(destination);
+      this._tapError = error || noop;
+      this._tapComplete = complete || noop;
+      this._tapNext = noop;
```
kwonoj: 83 already assigned into noop, do we need double init?

kwonoj:
Except few nits I think change looks ok, probably biggest thing we'd like to get opinion is this breaking change itself I guess.

src/Subscriber.ts
benlesh: The error is reported asynchronously, so in this case it doesn't matter. However, I think it's probably supposed to be report error then unsubscribe. I'll fix it.

benlesh:
> opinion is this breaking change itself I guess.

The opinions have already been expressed [here](https://github.com/tc39/proposal-observable/issues/119)

It's part of the spec, and frankly I really am not interested in this discussion again. We want to align with the proposal.

src/operators/tap.ts
benlesh: Fixed.

src/Subscriber.ts
benlesh: Improved.

martinsik:
I feel it's hard for me to imagine what is this change going to cause (for example in Angular apps). I guess it shouldn't change basically anything for me because I don't rely on rethrowing errors but I'm not completely sure.

Maybe this change should be described in [MIGRATION.md](https://github.com/ReactiveX/rxjs/blob/master/MIGRATION.md) with some before/after example.

PR #3066, opened 2017-11-11
feat(start): start method added to observers and subjects

### LINKED ISSUE #3061: Add `start` method to Observers
The TC39 spec has a `start` method to prevent this scenario from being a problem:

```ts
const subscription = range(0, 10000000).subscribe({
  next(value) { 
    if (value > 3) {
      // try to unsubscribe here, I dare you.
      subscription.unsubscribe();
    }
  }
});
```

The solution was to add an optional `start` method to Observer:

```ts
const subscription = range(0, 10000000).subscribe({
  start(subs) { this.subscription = subs; },
  next(value) { 
    if (value > 3) {
      // this.subscription is literally the same reference you'd get back
      // from the subscribe call itself.
      this.subscription.unsubscribe();
    }
  }
});
```

benlesh:
In order to match the TC39 and WHATWG proposals, a `start` method has been added to the `Observer` interface. The start method is fired prior to the Observable initializing, and it is handed the same `Subscription` object that is returned from the subscribe call. This enables consumers to get a handle on the subscription and unsubscribe early in the event of a synchronous firehose that they want to stop.

resolves #3061

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -1,6 +1,7 @@
 import { expect } from 'chai';
 import * as sinon from 'sinon';
 import * as Rx from '../src/Rx';
+import { ISubscription } from '../src/Subscription';
 import { TeardownLogic } from '../src/Subscription';
 import marbleTestingSignature = require('./helpers/marble-testing'); // tslint:disable-line:no-require-imports
 import { map } from '../src/operators/map';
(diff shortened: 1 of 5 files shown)


benlesh:
Oops, I'm going to need to make sure that you can't call start after it's been called once. We don't need user-land code calling it.

benlesh:
I'm not sure this is the route we're going to go with this. Closing for now.

PR #3067, opened 2017-11-11
docs(AjaxObservable): fix esdoc generation

grumblerchester:
**Description:**

Fix esdoc generation for AjaxResponse and AjaxError:
1. Due to doc generation process member info provided not in
constructor is missed.
2. esdoc does not recognize desciption provided as a param for type
annotation.

Add documentation for constructor parameters.

N/A


benlesh:
this has a lot of conflicts since the type was refactored. We don't really want people using the ctor directly anymore, so the documentation here isn't really necessary. Thank you for your thoughful contribution and your time, sorry I didn't work out this time. Hopefully you continue to contribute!

PR #3073, opened 2017-11-14
fix(tap): make next optional

### LINKED ISSUE #2534: Default values for do() parameters
`promise.then()` and `subscribe()` can be called with passing `undefined` as the first parameter, e.g. to only handle errors or completions. `do()` requires to pass a noop function like `() => undefined`. It should just have this as default value.

`doOnComplete` and `doOnError` shorthands would also be nice.

felixfbecker:
This is possible at runtime, but the typings don't reflect it and it's annoying when only wanting to tap into error or complete.

closes #2534

### DIFF
--- a/src/operators/tap.ts
+++ b/src/operators/tap.ts
@@ -8,7 +8,7 @@ import { noop } from '../util/noop';
 import { isFunction } from '../util/isFunction';
 
 /* tslint:disable:max-line-length */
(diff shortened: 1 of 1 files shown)


martinsik:
You can also use this:

```
tap({ error: e => console.log(e) })
```

or:

```
tap({ complete: () => console.log('...') })
```

felixfbecker:
I know

benlesh:
This is a valid change... however I'd also advise @martinsik's method mentioned above, as it's more readable, IMO.

Issue #3069, opened 2017-11-14
‘distinct’ can't work as the description in tutorials 


liuyunwei:
the test code in http://reactivex.io/rxjs/manual/tutorial.html#producing-values

```javascript
// typing "hello world"
var input = Rx.Observable.fromEvent(document.querySelector('input'), 'input');
// Only pass unique values through
input.pluck('target', 'value').distinct()
  .subscribe(value => console.log(value)); // "helo wrd"
```
When I type "hello world", it prints "hello world", is there anything wrong?

**RxJS version:**
5.0.3
**Code to reproduce:**
copy the demo codes to make a online demo
http://reactivex.io/rxjs/manual/tutorial.html#producing-values
http://jsbin.com/kiwumisoja/edit?html,js,console,output
**Expected behavior:**
http://reactivex.io/rxjs/manual/tutorial.html#producing-values
print "helo wrd" in console tab
**Actual behavior:**
print "hello world"
**Additional information:**

martinsik:
@liuyunwei I think it works as expected. When you write for example "Hello" then clear the input and start writing the same "Hello" again it'll be ignored as expected. But if you wrote "Hello2" you'll see a new log in the console "Hello2" because you haven't written this term yet.

SangKa:
The operators are fine, but the tutorials would confuse some guy.
The following sample code is better for  illuminate `distinct` or `distinctUntilChanged`:
```
Rx.Observable
  .of('h', 'e', 'l', 'l', 'o', ' ', 'w', 'o', 'r', 'l', 'd')
  .distinct()
  .subscribe(x => console.log(x));  // h, e, l, o,  , w, r, d
```

dengwanc:
as far as i'm concerned, 
1. the doc is totally confuse new one
2. in this case simplely
```
 input.pluck('target', 'value')
  .map(value => distinct(value)) // distinct need implement
  .subscribe(value => console.log(value)); // "helo wrd"
```

RxJS is not doing every thing, focus on the import part __lodash for async__

benlesh:
This is outdated it seems.

PR #3120, opened 2017-11-15
update the "input" event's plucking,  event.target.value=> event.data

liuyunwei:
**Description:**
update the tutorials to make work as expected
#3069


benlesh:
I think the original example does a better job showing that you can "deep pluck". I don't think we're going to merge this one.

Thank you very much, @liuyunwei, I greatly appreciate the thought and effort. Apologies for the slow response, and sorry we can't merge this request. I sincerely hope you contribute again.

PR #3123, opened 2017-11-17
docs: correct broken Reactive Angular video link

aciccarello:
**Description:**
The parleys.com domain has expired. The new Reactive Angular link will go directly to YouTube


benlesh:
Thank you, @aciccarello!!

PR #3129, opened 2017-11-27
WIP: Lerna integration

wardpeet:
**Description:**
This is just a simple WIP to show of what we can do with lerna and how it can help in the future to  move to a monorepo. I haven't got time to create demo or even test if this code even works 😛 

to test you can run
`npm install` it will run lerna bootstrap on **postinstall**. Lerna bootstrap will install all packages that are needed for the projects inside the packages folder.
`npm run build` runs lerna run build under the hood to start creating an esm5 build for ajax and schedulers.


kwonoj:
Before discuss anything further, how did you resolve https://github.com/lerna/lerna/issues/901 for rx repo?

wardpeet:
really unsure what the issue is here. It's just like any other package. I have never encountered any problems with it. I don't really get what issue they are describing..

kwonoj:
@wardpeet check my comment in slack channel. I'm copy pasting it here: 

```
in Rxjs repo, it doesn’t publish directly under rxjs repo itself - actual `package.json` will be created under dist for build pkgs, and publish happens here
With correct setup of monorepo evetually we’ll have `core`, `ajax`, etcs on top level dir, and each of them will have `dist` which’ll have actual package structure to publish in
it’s not about moving src into other places and else, it’s packaging structures to support multiple build targets around which’ll probably remain even after moving src into actual lerna repo
to try out, under `rxjs` git repo just `npm pack` and see what package contains, and compare it run `cd dist && npm pack`. 
```

In short, when we have structures like this
-- rxjs
   - core
     - dist
   - ajax
     - dist
   lerna.config.xxx

when you do lerna publish, it has to traverse down each subdir instead of push on top of `core` or `ajax`. Is this explains issue?

kwonoj:
if it's already working and good, no worries about this. I just wanted to confirm it.

wardpeet:
I'll cook up a fix that fixes what you need :) I just have to set the correct paths in package.json so npm pack does the correct thing.

benlesh:
I know this is a long time coming, but thanks again for your work here Ward. We're still investigating where we want to go with this. There's a possibility of also moving the build to Bazel. But I'm not sure that's going to be doable... we'll see.

wardpeet:
@benlesh no worries i'm happy to give a try at bazel just need to find some free time to make a proof of concept

benlesh:
I think we're going to pass on Lerna for now. There's a bazel build spike in the works, and we still have a lot of work to do to even decide how we'd like to structure this mono repo. I will be using this PR as a reference, but I'm going to close it for now.

benlesh:
Thanks again, @wardpeet! <3 I really appreciate all of the effort you've put in here.

PR #3131, opened 2017-11-28
build(bazel): adds bazel build files to support building rxjs with bazel from dist

gregmagolan:
Adds `WORKSPACE` and `BUILD.bazel` files to `/src` folder so they end up in `/dist/package/src` and copies `/tsconfig.json` to `/dist/packages/src` for external bazel build support.

Bazel built projects that pull in rxjs from npm can build it with Bazel with the following configuration in their WORKSPACE file:

```
local_repository(
    name = "rxjs",
    path = "node_modules/rxjs/src",
)
```

and using the following dependency on targets that depend on rxjs:

```
"@rxjs//:rxjs"
```

An example of this can be found here: https://github.com/gregmagolan/webpack-typescript-benchmarking/tree/bazel-rxjs

### DIFF
--- a/.make-packages.js
+++ b/.make-packages.js
@@ -98,6 +98,8 @@ fs.copySync(TYPE_ROOT, TYPE_PKG);
 copySources(ESM5_ROOT, ESM5_PKG, true);
 copySources(ESM2015_ROOT, ESM2015_PKG, true);
 
+// Copy over tsconfig.json for bazel build support
+fs.copySync('./tsconfig.json', PKG_ROOT + 'src/tsconfig.json');
 
 fs.writeJsonSync(PKG_ROOT + 'package.json', rootPackageJson);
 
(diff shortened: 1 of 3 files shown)


alexeagle:
Context: this is to support building Angular apps with Bazel. It's much easier to build rxjs from source (as we do at google) so the module format matches the JS files built for Angular, and avoids the need for System.js as the loader, see

> slight challenge with RxJs
from http://www.syntaxsuccess.com/viewarticle/angular-bazel-build-with-dev-server

Also, as discussed with @benlesh in the past, this is a first tentative step to building rxjs with Bazel, which can give improvements in developer turnaround time.

kwonoj:
so this is for consumer who install rx as dep and run bazel uild for their own project?

kwonoj:
oh answer was already posted before my question posted :)

benlesh:
@alexeagle did you mean to target master (v6 alpha) or stable (5.5) with this?

kwonoj:
I think nothing blocks to check in this?

PR #3134, opened 2017-11-28
fix(Observable.toArray): Fix Observable.toArray with multiple subscriptions.

josketres:
**Description:**
The current implementation of `Observable.toArray` reuses the same array instance for multiple subscriptions.

This example clarifies why that's wrong.
```js
const Rx = require('rxjs/Rx');
const observable = Rx.Observable.from([1,2]).toArray();
observable.take(1).subscribe(console.log); // prints [ 1, 2 ]
observable.take(1).subscribe(console.log); // prints [ 1, 2, 1, 2 ]
```

### DIFF
--- a/spec/operators/toArray-spec.ts
+++ b/spec/operators/toArray-spec.ts
@@ -65,6 +65,17 @@ describe('Observable.prototype.toArray', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
+  it('should allow multiple subscriptions', () => {
+    const e1 = hot('-x-^--y--|');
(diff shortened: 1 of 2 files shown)


benlesh:
:grimacing: I did that :blush:

So, I think the real solution here is to call `reduce(toArrayReducer, undefined)` and then inside of of the `toArrayReducer` check `!arr` instead of `index`. That will allocate fewer arrays.

Can you do that, @josketres?

benlesh:
Also, please target this at the `stable` branch, so we can get it in the next `5.5` patch.

PR #3135, opened 2017-11-28
toArray refactor

benlesh:
Just reducing array allocations and adding a test.


benlesh:
FFS... this PR is jacked up... I'll try it again.

PR #3136, opened 2017-11-28
To array refactor

benlesh:
Reduces array allocations and adds a test to ensure different arrays are produced by each subscription.

### DIFF
--- a/spec/operators/toArray-spec.ts
+++ b/spec/operators/toArray-spec.ts
@@ -65,6 +65,17 @@ describe('Observable.prototype.toArray', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
+  it('should allow multiple subscriptions', () => {
+    const e1 = hot('-x-^--y--|');
+    const e1subs =    '^     !';
+    const expected =  '------(w|)';
+
+    const result = e1.toArray();
+    expectObservable(result).toBe(expected, { w: ['y'] });
+    expectObservable(result).toBe(expected, { w: ['y'] });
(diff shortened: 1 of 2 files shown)


josketres:
I used an empty array and the index because I'd expect Observable.toArray to reduce to an empty array when the source observable completes without emitting any value.

So with this refactoring I'd expect the following test to fail: https://github.com/benlesh/RxJS/blob/9208e1de495742fc202285c4f6a6789e55caff12/spec/operators/toArray-spec.ts#L33

kwonoj:
Please do not aggregate fix and breaking changes in single PR targeting stable version for patch. We may need discuss desired behavior in your PR is way to go for next major version if it changes behavior of interface.

benlesh:
Actually, I don't think we can get around this extra allocation and still use `reduce` to build `toArray`. i'm going to close this PR for now.

- `empty().pipe(toArray())` should return `of([])`.
- `reduce(fn, seed)` requires a seed in this case.

benlesh:
> Please do not aggregate fix and breaking changes in single PR targeting stable version for patch. We may need discuss desired behavior in your PR is way to go for next major version if it changes behavior of interface.

I'm not sure what you're talking about here. There was no intent to introduce breaking changes. It was just an ill-conceived PR. lol

kwonoj:
@benlesh I was on mobile and thought
```
 I'd expect Observable.toArray to reduce to an empty array when the source observable completes without emitting any value.

So with this refactoring I'd expect the following test to fail
``` this is intended change which made PR breaks on CI.

PR #3139, opened 2017-11-28
fix(AjaxObservable): notify with error if fails to parse json response

### LINKED ISSUE #3138: Unhandled exception is thrown if AjaxObservable fails to parse response
**RxJS version:** 5.5.2

**Code to reproduce:** 
Please find test cases in the PR which will follow up the issue.

**Expected behavior:**
AjaxObservable sends an error to subscribers which could be caught using `catchError` operator.

**Actual behavior:**
If `responseType` is `''` in response, but it was set to `json` in request and xhr's `response` or `responseText` is not valid JSON, the unhandled error "Invalid character" is thrown.

**Additional information:**
The error is reproducible with IE11 only. Once the resource is not found, there is a proxy on our server, which overtakes the response and returns HTML 404 page.

myklt:
Catch error when parsing json response and notify observer with thrown error.
JSON could be invalid if proxy overtakes the response and in IE responseType is empty.

closes #3138

**Description:**
Please find description under the issue. This is my first PR to rxjs, so any feedback is welcome.

**Related issue (if exists):** #3138

### DIFF
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -246,6 +246,34 @@ describe('ajax', () => {
     expect(complete).to.be.true;
   });
 
+  it('should fail if fails to parse response', () => {
+    let error: any;
+    const obj = {
+      url: '/flibbertyJibbet',
+      responseType: 'json',
+      method: ''
+    };
+
+    ajax(obj)
+      .subscribe((x: any) => {
+        throw 'should not next';
+      }, (err: any) => {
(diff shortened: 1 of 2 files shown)


OliverJAsh:
Any updates on this?

benlesh:
Given this PR is massively in conflict, I'm not sure it can be merged. But I think the work here is important. Labelling it for refactor.

cartant:
@benlesh I've rebased and refactored this PR.

PR #3140, opened 2017-11-29
fix(animationFrame.spec): spec description fix

gsans:
This test is not related to delay but the effect of unsubscribe on the resulting Subscription.


benlesh:
Thanks, @gsans!

PR #3145, opened 2017-12-01
The big reorganization PR for v6 (WIP)

benlesh:
See commit messages for more details.

- Exports everything from top level `rxjs`.
- Pipeable (lettable) operators `import { map, filter, scan, switchMap } from 'rxjs';`
- Observable creation helpers `import { timer, concat, fromEvent } from 'rxjs';`
- Removes operator versions of observable creation helpers (use `concat(a, b, c)` instead of `a.pipe(concat(b, c))`)
- Starts moving things people should not be importing directly into an "internal" directory, so it's more obvious you're doing something silly.

### DIFF
--- a/spec/schedulers/AsapScheduler-spec.ts
+++ b/spec/schedulers/AsapScheduler-spec.ts
@@ -40,59 +40,6 @@ describe('Scheduler.asap', () => {
     sandbox.restore();
   });
 
-  it('should reuse the interval for recursively scheduled actions with the same delay', () => {
-    const sandbox = sinon.sandbox.create();
-    const fakeTimer = sandbox.useFakeTimers();
-    // callThrough is missing from the declarations installed by the typings tool in stable
-    const stubSetInterval = (<any> sinon.stub(global, 'setInterval')).callThrough();
-    function dispatch(state: any): void {
-      state.index += 1;
-      if (state.index < 3) {
-        (<any> this).schedule(state, state.period);
-      }
-    }
-    const period = 50;
-    const state = { index: 0, period };
-    asap.schedule(dispatch, period, state);
-    expect(state).to.have.property('index', 0);
-    expect(stubSetInterval).to.have.property('callCount', 1);
-    fakeTimer.tick(period);
-    expect(state).to.have.property('index', 1);
-    expect(stubSetInterval).to.have.property('callCount', 1);
-    fakeTimer.tick(period);
-    expect(state).to.have.property('index', 2);
-    expect(stubSetInterval).to.have.property('callCount', 1);
-    stubSetInterval.restore();
-    sandbox.restore();
-  });
-
-  it('should not reuse the interval for recursively scheduled actions with a different delay', () => {
-    const sandbox = sinon.sandbox.create();
-    const fakeTimer = sandbox.useFakeTimers();
-    // callThrough is missing from the declarations installed by the typings tool in stable
-    const stubSetInterval = (<any> sinon.stub(global, 'setInterval')).callThrough();
-    function dispatch(state: any): void {
-      state.index += 1;
-      state.period -= 1;
-      if (state.index < 3) {
-        (<any> this).schedule(state, state.period);
-      }
-    }
-    const period = 50;
-    const state = { index: 0, period };
-    asap.schedule(dispatch, period, state);
-    expect(state).to.have.property('index', 0);
-    expect(stubSetInterval).to.have.property('callCount', 1);
-    fakeTimer.tick(period);
-    expect(state).to.have.property('index', 1);
-    expect(stubSetInterval).to.have.property('callCount', 2);
-    fakeTimer.tick(period);
-    expect(state).to.have.property('index', 2);
-    expect(stubSetInterval).to.have.property('callCount', 3);
-    stubSetInterval.restore();
-    sandbox.restore();
-  });
-
   it('should schedule an action to happen later', (done: MochaDone) => {
     let actionHappened = false;
     asap.schedule(() => {
--- a/src/internal/scheduler/AsyncAction.ts
+++ b/src/internal/scheduler/AsyncAction.ts
@@ -29,6 +29,10 @@ export class AsyncAction<T> extends Action<T> {
     // Always replace the current state with the new state.
     this.state = state;
 
+    // Set the pending flag indicating that this action has been scheduled, or
+    // has recursively rescheduled itself.
+    this.pending = true;
+
     const id = this.id;
     const scheduler = this.scheduler;
 
@@ -56,10 +60,6 @@ export class AsyncAction<T> extends Action<T> {
     if (id != null) {
       this.id = this.recycleAsyncId(scheduler, id, delay);
     }
-
-    // Set the pending flag indicating that this action has been scheduled, or
-    // has recursively rescheduled itself.
-    this.pending = true;
 
     this.delay = delay;
     // If this action has already an async Id, don't request a new one.


kwonoj:
Assigning myself to give a quick peek.

rgbkrk:
I thought the transition from v4 -> v5 was good, this is even better. So much easier to navigate what's being exported, the interfaces make so much sense, and I'm looking forward to introducing more people to Rx with how much simpler the imports read. I'm loving this switchover from v5 -> v6.

benlesh:
cc @IgorMinar for additional review

IgorMinar:
Sorry for the delays, here is my feedback:

### Moving all the internals into `internal` directory is a very good change.

This will make it more explicit what is and what is not public api surface. I wonder if `private` as a directory name would be even more error-proof. either is much better than the current state

### Removing concat and friends as lettable operator is also very good.

There should be just one way to do this stuff and the creation method is way better way to go about it.

### Unifying all the exports and using `rxjs` module id as the only public entry point is better than the current state of dozens of entry points to import stuff from but I feel like it's going a bit too far.

There is a danger that the api surface will remain confusing because now all of the symbols will be exported only via one place and it will be much harder to tell what is what.

#### Example: the concat creation method

You removed the concat as an (lettable) operator but you left it there as a factory/creation function. Now if I go and try to use autocompletion to insert a operator into the `pipe` call chain, I'll likely get `concat` as one of the possible options (especially if I use javascript rather than typescript). It will be difficult for me to tell without reading the docs that `concat` is different from `map`. And I bet that the runtime error will be very cryptic unless you go the extra way (and add extra bytes) to check if the function looks like an operator before calling it.

The two symbols looks the same on the surface, but they are fundamentally two different things. If you exported all operators via `rxjs/operators` and `Observable`, `concat` and other symbols from `rxjs` the difference would be more clear.

### Code splitting issues

I'd also like to point out that webpack v3 will not be able to split up a library that has just a single entry point into multiple chunks. All of rxjs will always end up in the first chunk that references rxjs (most likely the "main" chunk). This will not be a problem with rxjs v4 if the whole rxjs package is marked as side-effect-free, which would however mean that you'd need to move the `rxjs/add/*` code into a different npm package.

I wouldn't worry too much about webpack v3 as long as we ensure that v4 can codesplit rxjs correctly.

I also believe that google3 will have the same problem as webpack v3, but we should confirm that with people more familiar with MSS and AJD.


### Overall I think this is the step in right direction

I just worry that it's a step too far. Rather than a single entry point I suggest that you have two + support for legacy `rxjs/operator/*` and `rxjs/add/*` module ids that would just reexport stuff from a new package `rxjs-compat` (or something like that) which would have to be installed by folks that want to keep the old module ids working while they refactor their code.

If you can explain how merging all operator and non-operator symbols into a single namespace (module) won't be confusing, then I'm happy to be persuaded to just go with a single `rxjs` entry point (+ support for all the legacy once via the `rxjs-compat` package).

fahad19:
can we still import the modules individually from their own paths? 

like:

```
// main package
import { scan } from 'rxjs';

// top-level module in a package
import { scan } from 'rxjs/scan';
```

IgorMinar:
I hope not. Having an ability to import the same symbol from more than one path creates a ton of confusion and in some scenarios results in code that is not always portable.

benlesh:
So... @igorminar has made a compelling argument to keep operators under `rxjs/operators` and everything else under `rxjs`... That would give us two different import points though, which makes me a little sad. 

What is the community consensus here?

@kwonoj? @mattpodwysocki? @jayphelps? @david-driscoll?

dominique-mueller:
I'd always prefer one source of import, makes things less complicated for library consumers.

dominique-mueller:
I'd always prefer one source of import, makes things less complicated for library consumers.

dominique-mueller:
I'd always prefer one source of import, makes things less complicated for library consumers.

jayphelps:
I'd be stoked with either. If I had to vote, I would pick all imports from root or put all in categories similar to how they are today, `import { concat } from 'rxjs/observable'` `import { AsyncScheduler } from 'rxjs/schedulers'` etc.

rgbkrk:
Either one is cool with me. I lean a little bit toward having `rxjs/operators` as a path so that I can tell the difference between a creation method and an operator via the path.

kylecordes:
I'm also very skeptical of two import paths. One is easy to defend, and N (one for each type of thing) is defensible though it would keep some of today's import verbosity. I would not be thrilled though, to have to explain that most things are in one place, but just one category of things is special enough to have its own separate place.

IgorMinar:
Just to be clear, I'm cool with N paths (where N is a small (single digit)
number), as long as each symbol is accessible only from one path.

My preferential vote is:

1. N paths
2. rxjs + rxjs/operators
3. single path (rxjs)

I'll let Ben, the community and other contributors decide as any of these 3
options is better than the current state *as long as* each symbol is
exported only via single path.

benlesh:
Would it be helpful to have the creation methods named something like: `rxMerge`, `rxConcat`, `rxFrom`, `rxFromEvent`, etc? or `fromConcat`, `fromMerge`, `fromForkJoin` `fromRace`, etc?

And then still have a single import location?

felixfbecker:
I don't really see a problem with having the source of the operators be in many files, but then have an index.ts that does `export * from` for all of them. That's a very common pattern. Most people will import from the root, but some may not have a bundling process with tree shaking. I feel like the index.ts shouldn't contain all that code, but also deep imports shouldn't be a requirement for all users.

> Would it be helpful to have the creation methods named something like: rxMerge, rxConcat, rxFrom, rxFromEvent, etc? or fromConcat, fromMerge, fromForkJoin fromRace, etc? or obserfvableMerge, observableConcat, observableOf?

This is definitely a question I have asked myself. Just using `of` or `concat` in your code sounds very generic. At the same time, that's true for all modules and it shouldn't be the module's responsibility to prefix with the module name like `rx` - folks can always alias or do `import * as rx`.
In a codebase that heavily uses rxjs it definitely reads nicer to just `combineLatest` instead of `observableCombineLatestFrom`. If there is a naming conflict, folks can use import aliases for those instances.

wardbell:
My instincts lie with @igorminar. Renaming creation operators will make migration harder and still leave you with something to explain. 

How many kinds are there? Creators/Observables, Operators, Subscribers, and Schedulers as I recall. And then there are Utilities right? That’s five. 

So is n===5? I guess that’s livable. 

One big rxjs bucket is my least favorite because there are just too many kinds of things in there that are easy to confuse. 

Adding prefixes (such as “rx” or “from”)will make migration harder and is just another way of saying “These things belong in the same bucket.” If so, give them a bucket.

Bikeshedding perhaps. It’s a great direction however you decide.

IgorMinar:
I should also point out that the small N paths solution doesn't suffer from ergonomics issues of introducing a new import line, because most editors and IDEs do that for you automatically.

If you however use an editor that can't create imports for you via autocompletion then using a single path is obviously more ergonomic.

The question is if optimizing the api surface for code authoring using an editor that has no ES import support is the right goal. I don't think it is, but that's my personal opinion 😄 

I believe that this is an example of authoring workflow most rxjs developers deserve:



source: https://marketplace.visualstudio.com/items?itemName=steoates.autoimport (note webstorm and other editors, support this workflow as well)

IgorMinar:
I forgot to mention that you should notice how the autocompletion in the screencast above highlights the path the symbol is coming from. This is what helps me pick the right symbol and in case of rxjs would help me distinguish between creation methods and operators. Once I make the choice, I no longer see the path (except) in the import, which removes clutter (unlike the "rx" prefix which would remain in the code even after I pick the symbol)

felixfbecker:
@IgorMinar I get that argument and I personally use an editor that has that feature, but I would still prefer to not have the top of every file look like this (as atm):

image

So not adding 1 import per operator is a valid goal no matter of having an IDE or not

felixfbecker:
Btw, static operators like `concat` could still be used as an operator by currying it:
```ts
of(1, 2, 3).pipe(
  $ => concat($, of(4, 5, 6))
)
```

IgorMinar:
@felixfbecker most of those import lines are because each operator is imported separately. my proposal would reduce that snippet to two imports (`from 'rxjs'` for Observable and creation menthods, and `from 'rjxs/opeartors'` for the operators).

Schedulers and other symbols are used very rarely so 99% of the code should not need to import them.

acutmore:
Controversial:
Single import path & operators all start with a common prefix

I think the discoverability of operators via IDE auto complete is critical. Gives back what we lose moving away from proto and towards pipes.

```js
import { of, concat, $map, $filter } from ‘rxjs’;

of(1).pipe(
  $map(v => v + 1),
  $ => concat($, of(3),
  $filter(v => v > 2),
)
```

SanderElias:
I'm on the same path as @IgorMinar. 

Personally, I would love to rename the creation methods as this would make it way more explicit what they are doing. However, as that is breaking backwards compatibility too much I don't think its a good idea. Adding aliases would sadly only add to the confusion. 

When that's from the table, `from 'rxjs'` for everything except for the operators that will come `from 'rxjs/operators'` seems the most straightforward. 
As it separates what goes on the "inside" and "outside" of the pipe, it will relieve some of the confusion to new(ish) devs coming to RxJs.

myspivey:
I back @IgorMinar mostly as well although I would argue that as this is a major version, I think pulling the trigger and renaming is cleaner. I like the from* options as that is very descriptive to what you are doing and reduces confusion between the operators. Sure we can import from two places but that then leaves code readability up to how it was imported and can create collision and confusion. If I am adopting a new major version I know things are going to break. When I switched to lettables I had to go through and rename catch and do in many places, why would this be any different. That is just my own feelings personally. Overall amazing work and looking forward to seeing this in the wild.

JacobChang:
I think single import path will be fine for TypeScript developers, as the type will tell the difference between operators and create function, but for JS developers, especially for beginners, single import will create some confusion, so agree with @IgorMinar

wmaurer:
I'm for importing from either a single entry point (1st preference) or ~5 paths (2nd preference).

I would also welcome renaming creation operators to `rxMerge`, `rxConcat`, etc. I like the naming distinction from lettable operators.

I think renaming the creation operators has a relative minor impact on existing code bases (search/replace). The bigger impact is removing the lettable operator equivalents. I now use the creation methods exclusively now, but our older code does not. Changing from operators to create methods involves more shifting of code around.

benlesh:
So everyone wants to divide things up in terms of functionality I guess...

Commonly used:
- `rxjs` - for types like Subject, Observable, BehaviorSubject, etc
- `rxjs/create` - for `of`, `from`, `fromEvent`, `concat`, `combineLatest`, etc
- `rxjs/operators` - for `map`, `switchMap`, etc
- `rxjs/util` for utilities like `noop` and `pipe` (After typing "noop" I typed "poop" instead of "pipe" the first time)
- `rxjs/scheduler` for the schedulers
- `rxjs/symbol` for the symbols such as `Symbol.observable` 

I have to be honest, I *really* don't like it. I'd just assume have one entry point.

Is the "compromise" proposal to have:

- `rxjs/operators` - `map`, `switchMap`, etc
- `rxjs` - everything else

??


I'm going to finish up this PR today and ship. This has taken too long.

kylecordes:
@benlesh Your "one" notion has much wisdom, is very compatible with the future of bundling tools, is easy to consume, etc. The only downsides, discussed here, are that the list of things exported from that one place can get a little long (hard to look through it by hand), and names have to be deconflicted. Trust your heart :-)

benlesh:
Well the fact is, even if we split them up, there's going to be a problem with `concat`.

```ts
import { concat } from 'rxjs';
import { concat } from 'rxjs/operators';
```
... you'll need to alias regardless.  What to, though? `fromConcat`? If that's the case, should we just rename it? Are there compelling cases why: `a.pipe(concat(b, c))` is better than `concat(a, b, c)`? Perhaps `a.pipe(switchMap(fn), concat(b, c))` is better than `concat(a.pipe(switchMap(fn)), b, c)`?  Doesn't read nicely to me.

jayphelps:
@benlesh 

> I have to be honest, I really don't like it. I'd just assume have one entry point.

For me at least, the idea was primarily to have `rxjs/operators` and `rxjs/observables` (or whatever). The other categories we could do, or we could just put others in the root as they're so rarely imported that I doubt it makes much of a difference.

> Well the fact is, even if we split them up, there's going to be a problem with concat.

I thought the operator variants were going away?

> Removes operator versions of observable creation helpers (use concat(a, b, c) instead of a.pipe(concat(b, c)))

benlesh:
@jayphelps 

> I thought the operator variants were going away?

That's what I'd like, but I felt like there is a subtext here to keep them. I'm probably mistaken. If so, that's great.

kylecordes:
@benlesh It always seemed odd to have two ways to do `concat` and friend. Yet some people rely on them. I think it is justifiable on a major new version, to tighten up the core API, and move "pipeable concat" over to a secondary support package, to help out the folks would would miss it if it were gone.

benlesh:
So:

- `rxjs/operators` - `map`, `switchMap`, etc
- `rxjs/observable` - `from`, `of`, `fromEvent` and friends
- `rxjs` - junk drawer? Just classes? Should Schedulers go in here? What about utility functions like `identity`, `noop` and `pipe`?

rgbkrk:
I think it's cool to put `identity`, `noop`, and ~~`poop`~~ `pipe` at the top level.

kylecordes:
Speaking in favor of what Ben said up a ways (just one package): It seems silly to have a `rxjs/observable`. Observables are what RxJS *is all about*.

rgbkrk:
> Speaking in favor of what Ben said up a ways (just one package): It seems silly to have a rxjs/observable. Observables are what RxJS is all about.

Yeah, I liked the `rxjs/create` name.

amoerie:
If I can chime in here: I would vote for importing everything  from "RxJs". I don't agree that this would be very confusing, because there are types and documentation for everything. (Thanks btw!) If anything, I'm afraid that having multiple import paths deviates from the norm and would be worse complexity wise! When in doubt, go for simple..

Just my 2 cents..

benlesh:
Moving this pr to #3223 ... it's the same PR I just had a problem with my fork. (Sorry).

Issue #3008, opened 2017-10-27
Autopublish nightly package 


kwonoj:
**RxJS version:**

**Code to reproduce:**

**Expected behavior:**

**Actual behavior:**

**Additional information:**
We'd like to introduce `nightly` package publish, as canary version based on master (unstable) branch.

kwonoj:
in addition per-PR, if it's possible. like https://github.com/facebookincubator/create-react-app/pull/3124#issuecomment-329472767 .

kwonoj:
This was an idea we never got to execution. If we have proper action plan will create new issue.

PR #3159, opened 2017-12-04
build(publish): nightly publish script

kwonoj:
**Description:**

**THIS SCRIPT DOESN'T PUBLISH ANYTHING YET**
I've promised this for a while, and finally able to stab at very initial steps for this.
This PR works on our travis cron job runs daily on master, replaces version in package.json into specific nightly scheme and publish it. 

Since travis doesn't have npm auth token yet, all of this script is safe at this moment and does not publish anything - before start publish, need to conclude couple of things

- [x] what's semver safe version scheme for nightly? 
: using `--tag=nightly` as suggested
- [ ] is publish script in here is correct? (run `npm publish` under `dist/package`)

#3008


benlesh:
>  what's semver safe version scheme for nightly?

Just do something like `5.5.0-nightly.COMMIT_SHA_HERE` That should be fine. And make sure it's tagged with `nightly`.

benlesh:
Although, I'm not sure I want to junk up the npm versioning with 365 builds a year... Maybe the same version every time would be better? 

It's worth researching the best approach.

kwonoj:
@benlesh 

1. shouldn't it be `6.0`-nightly? this is intended for `master`.
2. are you referring creating git tags, or other else?

> I'm not sure I want to junk up the npm versioning with 365 builds a year... Maybe the same version every time would be better?

: makes sense. Maybe we can define one version only and stick for nightly.

kwonoj:
Now PR doesn't modify but publish with `--tag=nightly` as suggested.

kwonoj:
and we can actually turn this into general publish script for tagged, regular build as well 🤔  let's proceed separately, once after we verify nightly actually works.

kwonoj:
Marking as blocked until getting conclusion for version number strategy.

cartant:
With TypeScript, the maintainers use a concatenated year, month and day:

```
$ npm dist-tag ls typescript
beta: 2.0.0
insiders: 2.6.2-insiders.20171120
latest: 2.6.2
next: 2.7.0-dev.20171203
rc: 2.6.0-rc
```

If it's a nightly, using a date-derived number has the additional advantage of allowing devs to quickly see how old/recent the nightly is.

cartant:
> Maybe we can define one version only and stick for nightly.

http://blog.npmjs.org/post/77758351673/no-more-npm-publish-f

kwonoj:
> With TypeScript, the maintainers use a concatenated year, month and day:

Yeah, that was my initial take. @benlesh 's primary concern is stacking up new pkgs per each day afaik.

benlesh:
Sorry, I thought the plan was to always tag with nightly. ... my concern was over how to add versions if needed... like 5.5.0-nightly.SHA_HERE or something... and will that junk up the npm info for the project? That's what I was trying to figure out from @seldo on Twitter.

kwonoj:
I just did `npm info typescript` for checking how they does, and it is indeeded stacking up whole 😅  I haven't feel this is concern (probably me didn't use npm info at all) - what's main issue if it happens?

benlesh:
Meh. Whatever. Let's do it.

@kwonoj can you fix Travis?

kwonoj:
Assigning myself for further. I'll try to make dry run and will ask reviwew once again before actually push the button.

benlesh:
@kwonoj what are we thinking about this PR at this point? It's high on my priority list to get nightlies published.

benlesh:
Just chatted with @kwonoj, and we're going to kill this one for now.

PR #3160, opened 2017-12-04
fix(dependency): move symbol-observable into devdependency

kwonoj:
**Description:**
We don't use `symbol-observable` for dependency anywhere, moving it into devdep.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ matrix:
 
 before_install:
   - nvm install $NODE_VER
-  - npm install -g npm@4 && npm install -g npx && node -v && npm -v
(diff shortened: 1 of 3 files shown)


benlesh:
@kwonoj was this meant for stable or master?

kwonoj:
@benlesh master. I didn't want to bring any possible regressions in `stable`, though change looks harmless.

lrowe:
If rxjs removes the symbol-observable dependency how do other observable implementations interoperate with rxjs through the symbol-observable protocol? Does this continue to work because symbol-observable mutates the global Symbol object to add a Symbol.observable property?

kwonoj:
Rxjs has own symbol interop already. https://github.com/ReactiveX/rxjs/blob/master/src/symbol/observable.ts

PR #3167, opened 2017-12-05
build(bazel): adds bazel build files to support building rxjs with bazel from dist [STABLE]

gregmagolan:
Same as #3131 but onto the stable branch


gregmagolan:
Whoops

gregmagolan:
Switched base to stable

gregmagolan:
Something doesn't look right

PR #3173, opened 2017-12-08
refactor(elementAt): implement in terms of first

benlesh:
Reduces the size of the operator substantially.

NOTE: We might want to consider changing the error we expect from this and first to be something uniform to remove a step there is a small mismatch between EmptyError and ArgumentOutOfRangeError. I understand the semantics, but I think might be able to come up with an Error that suits both

**Description:**

### DIFF
--- a/src/operators/elementAt.ts
+++ b/src/operators/elementAt.ts
@@ -1,9 +1,12 @@
-import { Operator } from '../Operator';
-import { Subscriber } from '../Subscriber';
+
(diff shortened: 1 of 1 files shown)


src/operators/elementAt.ts
```diff
@@ -1,9 +1,12 @@
-import { Operator } from '../Operator';
-import { Subscriber } from '../Subscriber';
+
```
kwonoj: empty newline?

benlesh:
I think I can do better.

PR #3174, opened 2017-12-08
refactor(first): make first smaller

benlesh:
Reduces the size of first my implementing it in terms of other operators.

### DIFF
--- a/src/operators/first.ts
+++ b/src/operators/first.ts
@@ -1,20 +1,28 @@
 import { Observable } from '../Observable';
-import { Operator } from '../Operator';
-import { Subscriber } from '../Subscriber';
 import { EmptyError } from '../util/EmptyError';
 import { OperatorFunction, MonoTypeOperatorFunction } from '../interfaces';
+
+import { concat } from '../observable/concat';
+import { filter } from './filter';
+import { of } from '../observable/of';
+import { _throw } from '../observable/throw';
+import { identity } from '../util/identity';
+import { take } from './take';
+import { map } from './map';
+import { pipe } from '../util/pipe';
+
 /* tslint:disable:max-line-length */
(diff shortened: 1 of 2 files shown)


src/operators/first.ts
```diff
@@ -67,100 +75,34 @@ export function first<T>(predicate: (value: T, index: number, source: Observable
...
-  private _emitFinal(value: any) {
-    const destination = this.destination;
-    if (!this._emitted) {
-      this._emitted = true;
-      destination.next(value);
-      destination.complete();
-      this.hasCompleted = true;
-    }
-  }
+  const after = defaultValue ? of(defaultValue) : _throw(new EmptyError());
```
cartant: Testing whether or not `defaultValue` is falsy won't allow for default values of `0` or `false`, etc.

benlesh: Right. Don't code tired. Thanks

martinsik:
I tested this PR and it actually degrades performance compared to the previous implementation but I understand that this operator isn't used as much so maybe it doesn't matter (it's cca 10x slower).

Now:
```
                                         |                     RxJS 4.1.0 |                     RxJS 5.5.3 |          factor |      % improved
                       first - immediate |                34,519 (±4.93%) |               253,469 (±4.53%) |           7.34x |          634.3%
```

Before:

```
                                         |                     RxJS 4.1.0 |                     RxJS 5.5.3 |          factor |      % improved
                       first - immediate |                27,327 (±5.23%) |             2,053,011 (±3.78%) |          75.13x |        7,412.9%
```

benlesh:
@martinsik yeah.. I'm dubious about the change here myself. I'll look at a few other ways to reduce the size. This is more of an experiment.

PR #3177, opened 2017-12-08
refactor(timeInterval): implement in terms of scan

benlesh:
reducing the size of timeInterval. A non-hotpath operator.

### DIFF
--- a/src/internal/operators/timeInterval.ts
+++ b/src/internal/operators/timeInterval.ts
@@ -1,48 +1,24 @@
-import { Operator } from '../Operator';
+
 import { Observable } from '../Observable';
-import { Subscriber } from '../Subscriber';
 import { async } from '../scheduler/async';
-import { OperatorFunction, SchedulerLike, TimeInterval as TimeIntervalInterface } from '../types';
+import { SchedulerLike, OperatorFunction } from '../types';
+import { scan } from './scan';
+import { defer } from '../observable/defer';
+import { map } from './map';
 
 export function timeInterval<T>(scheduler: SchedulerLike = async): OperatorFunction<T, TimeInterval<T>> {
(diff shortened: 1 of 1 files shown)


src/operators/timeInterval.ts
```diff
@@ -1,49 +1,24 @@
...
 
 export function timeInterval<T>(scheduler: IScheduler = async): OperatorFunction<T, TimeInterval<T>> {
-  return (source: Observable<T>) => source.lift(new TimeIntervalOperator(scheduler));
+  return (source: Observable<T>) => defer(() => {
+    return source.pipe(
+      scan(
+        ({ current }, value) => ({ value, current: scheduler.now(), last: current }),
+        { current: scheduler.now() }
+      ) as any,
+      map(({ current, last, value }) => new TimeInterval(value, current - last)),
```
kwonoj: nit: will there be difference if we just map to object instead of class?

```
interface TimeInterval<T> {
readonly value: T;
readonly interval: number
}

....
map<TimeInterval>((...) => ({value, interval: current - last}));
```

benlesh: Not really. There's probably some negligible perf benefit to it being a class, butt I'm speculating I can change it. It'll probably further reduce size

benlesh: Not really. There's probably some negligible perf benefit to it being a class, butt I'm speculating I can change it. It'll probably further reduce size

kwonoj: If we don't lose anything (and I'm pretty sure transpiled sizewise object will be smaller than class) I'd like to suggest to go for object, as `TimeInterval` class really doesn't nothing in here.

benlesh:
@kwonoj, I could remove the class just yet, because it's being exported for backwards compat reasons.  We'll remove it soon.

PR #3181, opened 2017-12-12
Minor optimization in ignoreElements operator

fljot:
**Description:**
Minor optimization in `ignoreElements` operator.
We don't need to call any kind of `noop` function, simply do not call anything.
I mean, sure, it's not a bottleneck by no means, but better without, right?


fljot:
... will fix commit message later, if PR is needed at all.

kwonoj:
This doesn't harm anything and safe to check in, so.. 🚢

PR #3187, opened 2017-12-15
docs: use `pipe` method for showing usage of operators as pure functions

fahad19:
**Description:**

The operators creation guide was recommending the usage of `let` method in Observables.

But from v5.5, we have lettable operators. Better suggest new users to use `pipe` instead.


benlesh:
Thank you very much for your contribution!

fahad19:
@benlesh: that was very quick of you to take care of this! thanks! 🎉

PR #3200, opened 2017-12-27
Fix broken example in /src/scheduler/animationFrame.ts.

chiunhau:
**Description:**
Just fix a broken example in /src/scheduler/animationFrame.ts and now it works 👍


benlesh:
Thank you for catching this, @chiunhau!

PR #3201, opened 2017-12-28
feat: allow TestScheduler to accept comparison functions

tklever:
TestScheduler is an invaluable tool when testing observables, however, in practice, the singular reliance on deepEqual can make for a burdensome and rigid testing experience. This change adds the ability for users to specify their own comparison functions. These user supplied comparisons can open the door to testing conveniences like property-value assertions that developers may be accustomed to.

The first step to make this a reality was to step away from comparing observables as a whole and compare each individual emission. This allowed me to write up some (what I believe to be) more helpful error messages when tests fail, but the downside is test failures no longer provide "whole stream" context.

**The Good**
- User Supplied Comparison Functions (allow a developer to run propVal, greaterThan, hasMembers, etc. on an observable's values)
- Targeted specific error messages about failures, no more (potentially giant) deep equal monster for test failures 

**The Bad**
- TestScheduler needs to do a handful of comparisons now, which I'm doing with lodash, I could pass this function in, but that would require a BC break on the scheduler constructor, which I didn't want to do unless I got feedback to do so.
- I stole the error stack normalization and stringify functions from testscheduler-ui for use in this class, open to suggestions on how I should appropriately remove that duplication.

### DIFF
--- a/spec/helpers/marble-testing.ts
+++ b/spec/helpers/marble-testing.ts
@@ -2,7 +2,7 @@ import { Observable } from '../../src/Observable';
 import { SubscriptionLog } from '../../src/testing/SubscriptionLog';
 import { ColdObservable } from '../../src/testing/ColdObservable';
 import { HotObservable } from '../../src/testing/HotObservable';
-import { TestScheduler, observableToBeFn, subscriptionLogsToBeFn } from '../../src/testing/TestScheduler';
+import { TestScheduler, observableToBeFn, observableToFn, subscriptionLogsToBeFn } from '../../src/testing/TestScheduler';
 
 declare const global: any;
 
@@ -23,7 +23,7 @@ export function cold(marbles: string, values?: any, error?: any): ColdObservable
 }
 
 export function expectObservable(observable: Observable<any>,
-                                 unsubscriptionMarbles: string = null): ({ toBe: observableToBeFn }) {
+                                 unsubscriptionMarbles: string = null): ({ toBe: observableToBeFn, to: observableToFn }) {
   if (!global.rxTestScheduler) {
     throw 'tried to use expectObservable() in async test';
   }
--- a/spec/schedulers/TestScheduler-spec.ts
+++ b/spec/schedulers/TestScheduler-spec.ts
@@ -197,6 +197,10 @@ describe('TestScheduler', () => {
         expect(expectObservable(Rx.Observable.of(1)).toBe).to.be.a('function');
       });
 
+      it('should return an object with a to function', () => {
+        expect(expectObservable(Rx.Observable.of(1)).to).to.be.a('function');
+      });
+
       it('should append to flushTests array', () => {
         expectObservable(Rx.Observable.empty());
         expect((<any>rxTestScheduler).flushTests.length).to.equal(1);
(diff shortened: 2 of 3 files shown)


kwonoj:
I am actually more interested in interfaces around `taking out` Notification<T> values instead of inject comparison into assertion of test scheduler as similar to https://github.com/kwonoj/rx-sandbox#custom-assertion-for-marble-diagram (yes, it's my work), allows scheduler itself becomes test framework / assertion agnostic and anyone can use their own comparison. Core team has been thinking of improving / spliiting test related code from core and do have some additional changes (including breaking changes).

tklever:
@kwonoj, Certainly open to going down that road and think that it's probably a more robust long term solution. This change struck me a minimally invasive (for now) and I get assertion agnostic (although wrapped) out of the deal.

If that's what the core team wants to see, it's honestly not an impossible amount of work. The marble parsing would have to refactored out independently and on first glance some kind of "Observable Notifications" result object containing emissions and subscriptions would need to exist (similar concepts in your library)

kwonoj:
@tklever thanks for creating PR, now core team trying to make actual changes in core repo around test scheduler interfaces. Rough proof of concept can be found at https://github.com/ReactiveX/rxjs/pull/3266, which resolves most of pain points around current interface - plz check it out. Once we have high level agreement, there'll be following actual PR to migrate it.

jayphelps:
> The first step to make this a reality was to step away from comparing observables as a whole and compare each individual emission.

It's not clear how this works in practice as the same function would be called for every next/error/complete individual wouldn't it? How do you distinguish?

@benlesh appears to be a non-breaking change, so it could be introduced. I'm still a bit confused though on the design.

tklever:
In theory (and forgive me, it's been a bit since I worked on or thought about this), by being able to specify a variable comparison function, you could (in psudocode):

```typescript
const obs = hot('a-b-c-d-(e|)', { 
    a: { num: 1 },
    b: { num: 2 },
    c: { num: 3 },
    d: { num: 4 },
    e: { num: 5 },
})

expectObservableWithComparator(propValue.bind({}, 'num'), obs, '1-2-3-4-(5|)');
```

Basically allowing for testing of pieces of a emission as, in some circumstances, deepEqual in practice can force a significant amount of mocking. I think a change like this (or allowing for variable comparisons through some means, even if this PR is not good) could make testing observables more flexible and in some cases easier to read.

The example above is contrived and small, but imagine for a moment my emission is a large object with many properties, and I want to test that only one of them is changing, or that every emission is a number greater than zero. To do this now, you have to drop back to subscribe.

As you mentioned the function would be called every time, but informed by the provided marble diagram (just as deepEqual is today).

Hope that helps clear up the DESIRE of this PR, even if the implementation isn't worthy of inclusion

tklever:
@benlesh No worries. Certainly not off putting in the slightest. 

I'll keep an ear to the ground for @jayphelps' upcoming "improved test runner" work to see if that satisfies my needs or allows me a baseline to re-propose an updated and clarified version of this idea, I will.

Sorry I never made an articulate point here, thanks for taking the time to review it.

jayphelps:
The latest testing update was released as part of v6 https://github.com/ReactiveX/rxjs/blob/master/doc/marble-testing.md

PR #3202, opened 2017-12-29
docs(lettable-operators): example of own operators

JLarky:


doc/lettable-operators.md
```diff
@@ -95,11 +95,13 @@ const takeEveryNth = (n: number) => <T>(source: Observable<T>) =>
...
   });
 
+const takeEveryNthSimple = (n: number) => <T>(source: Observable<T>) =>
+  source.pipe(filter((value, index) => index % n === 0 ))
 
 interval(1000).pipe(
   takeEveryNth(2),
   map(x => x + x),
-  takeEveryNth(3),
+  takeEveryNthWithPipe(3),
```
rgbkrk: Did you mean to make this `takeEveryNthSimple`

JLarky: yes, I decided to rename it to Simple but forgot this place :)

JLarky:
@rgbkrk pushed amended version with takeEveryNthSimple in both places

JLarky:
Is there anything else I can do to increase chances of PR being merged? :)

doc/lettable-operators.md
```diff
@@ -95,11 +95,13 @@ const takeEveryNth = (n: number) => <T>(source: Observable<T>) =>
     })
   });
 
+const takeEveryNthSimple = (n: number) => <T>(source: Observable<T>) =>
```
benlesh: Interestingly, you could simplify this further like so:

```ts
const takeEveryNthSimplest = (n: number) => filter((_, index) => index % n === 0);
```

... the primary point here was to show the deepest example in the simplest way... But I suppose we can go into more detail.

benlesh: LGTM

benlesh:
Thanks @JLarky ... sorry for the slow review. I've been busy. ;)

Issue #3204, opened 2018-01-01
debounce() and throttle()’s selector argument has wrong return type


mernen:
Here’s the current signature for `debounce()`:

https://github.com/ReactiveX/rxjs/blob/c1721e3dd582c509064e141425c15622f68d60cb/src/operators/debounce.ts#L53

One thing that confused me is that it asks for your function to return a `SubscribableOrPromise<number>`. The documentation compounded my confusion ("A function (…) for computing the timeout duration for each source value”): is my selector supposed to return the number of milliseconds? Turns out the result of that observable/promise is completely discarded.

**RxJS version:** 5.5.6 (and, looking at the code, [the next branch](https://github.com/ReactiveX/rxjs/blob/27576f749ecac85cc4238eead8b1e54673e1e440/src/operator/debounce.ts#L52), too)

**Additional information:** `throttle()` has the same issue. Interestingly, `audit()` is correctly using `SubscribableOrPromise<any>`:

https://github.com/ReactiveX/rxjs/blob/c1721e3dd582c509064e141425c15622f68d60cb/src/operators/audit.ts#L52

cartant:
Regarding `debounce`, see #3164 and #3169. I will have a look at the `throttle` typings.

mernen:
Ah, that’s nice! Sorry, it completely forgot to search for past reports. My bad.

cartant:
No worries; you're right. `throttle` also uses `number` instead of `any`. I'll create another PR.

PR #3169, opened 2017-12-05
fix(typings): relax debounce selector type

cartant:
**Description:**

Changes the typings for the `debounce` method's selected observable to `any` - in line with the typings used for other methods that include selectors for notification observables. E.g. [`audit`](https://github.com/ReactiveX/rxjs/blob/5.5.4/src/operators/audit.ts#L52).

**Related issue (if exists):** #3164

PR #3205, opened 2018-01-02
fix(typings): relax throttle selector type

cartant:
**Description:**

Changes the typings for the `throttle` method's selected observable to `any` - in line with the typings used for other methods that include selectors for notification observables. E.g. [`audit`](https://github.com/ReactiveX/rxjs/blob/5.5.4/src/operators/audit.ts#L52).

**Related issue (if exists):** #3204 (also see PR #3169)

### DIFF
--- a/spec/operators/throttle-spec.ts
+++ b/spec/operators/throttle-spec.ts
@@ -2,6 +2,7 @@ import { expect } from 'chai';
 import * as Rx from '../../src/Rx';
 import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
 
+declare const type;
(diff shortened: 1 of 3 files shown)


kwonoj:
Let me try to see changes soon.

benlesh:
Can we get this PR updated and checked? @cartant @kwonoj 

It seems like a solid add.

cartant:
This PR appears to no longer be required. `number` seems to have been changed to `any` as part of the v6 restructure.

cartant:
Whoops. It seems that I'm wrong and Git is smarter than I thought it was.

cartant:
Should be rebased, now.

Issue #3122, opened 2017-11-16
Using AbortSignal and AbortController


benlesh:
The standards proposals are leaning towards using AbortSignal and AbortController (ala `fetch` cancellation) for cancellation with Observables. I generally find this favorable for a few reasons and I think I see a migration path for us.

### Proposal

1. Observable `subscribe` accepts `Observer` and an optional `AbortSignal`.
2. Observable `forEach` accepts an optional `AbortSignal`.
3. The subscriber function for `new Observable(subscriber)` gives both a safe `Observer` and an `AbortSignal`, but is expected to return void.
4. The `Observer` passed to subscribe will have an optional `abort` handler that will get the `AbortError` if the observable is aborted.
5. If an abort is triggered on a subscription that happened with `forEach`, the `AbortError` would go down the rejection path just like it does with fetch. Since the returned promise is a guarantee to a future where the producer either errors or completes, it makes sense to send the `AbortError` down the rejection path in this case.

### Advantages

- Synchronous observables cannot firehose and lock up an app, the cancellation semantic exists prioer to subscription.
- If multiple data producers are set up in a subscriber function (in the Observable constructor), you can add them to the abort signal one at a time, guaranteeing they can be torn down if a subsequent step fails. This wasn't possible before without some serious hackery.
- Observables all the way down. Since `AbortSignal` is an EventTarget, and EventTarget may soon have an `on` method that returns an observable. you could use `abortSignal.on('abort')` to compose teardown if you so choose.
- We will now have the ability to be notified, uniquely, of cancellation (without having to do trickery to figure out if it was really an error or a completion)

### Migration path
We'll start using `AbortSignal` internally throughout the library.
1. Add an overload to `subscribe` that accepts an `Observer` and an `AbortSignal`. If `AbortSignal` is passed, it does not return a `Subscription`.
2. Make `Subscription` into an `AbortController`
  - Add the ability to create an `AbortSignal`.
  - Alias `unsubscribe` as `abort`.
3. Have the subscriber function passed to the `Observable` constructor receive an `AbortSignal`.
4. Any function/Subscription returned from the subscriber function (from the constructor), will automatically be added to the `AbortSignal`.
5. Remove the observer `start` method I just added. :) lol It's no longer necessary.
6. Add the observer `abort` method that accepts `AbortError`.
7. Ensure that aborts arriving at the tail-end of a `forEach` get sent as rejections of type `AbortError` to the returned `Promise`.
[...]

PR #3206, opened 2018-01-02
split the method “subscribe” into two steps to get the subscription instance in handle function

ArthasModern:
**Description:**

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -19,7 +19,7 @@ export interface Subscribable<T> {
             error?: (error: any) => void,
             complete?: () => void): AnonymousSubscription;
 
-  startSubscription(sink?: AnonymousSubscription): void;
+  startSubscription(sink: AnonymousSubscription<T>): void;
 }
 
 export type SubscribableOrPromise<T> = Subscribable<T> | PromiseLike<T>;
@@ -252,7 +252,7 @@ export class Observable<T> implements Subscribable<T> {
     return toSubscriber(observerOrNext, error, complete);
   }
 
-  startSubscription(sink?: Subscription): void {
+  startSubscription(sink: Subscriber<T>): void {
 
     const { operator } = this;
 
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -14,6 +14,12 @@ export interface Subscribable<T> {
   subscribe(observerOrNext?: PartialObserver<T> | ((value: T) => void),
             error?: (error: any) => void,
             complete?: () => void): AnonymousSubscription;
+
+  createSubscription(observerOrNext?: PartialObserver<T> | ((value: T) => void),
+            error?: (error: any) => void,
+            complete?: () => void): AnonymousSubscription;
+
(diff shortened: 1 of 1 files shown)


ArthasModern:
Fuck, too complicated!

ArthasModern:
Fuck, too complicated to commit!

Brooooooklyn:
may I ask what's the intent of `split subscribe` ?

ArthasModern:
Hi @Brooooooklyn 
split the method “subscribe” into two steps
for this case: we need to get the subscription instance in the handle function

@example <caption>unsubscribe in the handle function</caption>

const subscription = Rx.Observable.of(1, 2, 3).subscribe(
    function(value) {
         // do handle here, then we need unsubscribe immediately!
        subscription.unsubscribe();
         // error: subscription undefined!!!
    },
    undefined,
    undefined
    );

ArthasModern:
@Brooooooklyn because the subscribe function do not return the subscription immediately, sometimes we need to get the subscription instance before the action start.

ArthasModern:
@Brooooooklyn do you agree with me？and you know how to submit it correctly？

paulpdaniels:
@ArthasModern Thanks for the PR, but IMO I don't think this is a good way to approach this problem. Explicit unsubscribe, especially in the `subscribe` block is an anti-pattern. If you encounter this then more than likely you want to be using something like `takeWhile` instead.  @benlesh wrote a good article on this as well: https://medium.com/@benlesh/rxjs-dont-unsubscribe-6753ed4fda87

Brooooooklyn:
I agree with @paulpdaniels , you should never `unsubscribe` you subscription in `subscriber`

ArthasModern:
@paulpdaniels Thank you very much for your answer! `take` `takeWhile` `takeUntil` etc. is very useful! Combined with `filter` , there are many ways to solve my problem. I just think it's not convenient enough! and there is no harm to split `subscribe` as an extension.

ArthasModern:
OK, it's just my personal opinion. Thank you very much for your advice!

ArthasModern:
OK, it's just my personal opinion. Thank you very much for your advice!
@Brooooooklyn @paulpdaniels

cartant:
See #3061 and #3122.

kwonoj:
As suffeciently described above there are discussions of controlling interfaces for observables and I don't think we'll pursue this proposed interfaces.

benlesh:
I don't think we can accept this PR. There are other designs around how to handle this.

One is the start event, added to the TC39 Proposal, and it gives you the Subscription just before the Observable body is executed. The other is to hand the Subscription to the next callback as a second argument, and that's being looked at in the WHATWG proposal. Both proposals are cooperating and will align at some point, but it's not a big issue. So until the proposals settle, we're going to hold off on adding anything. (I had added the start handler briefly in the past).

Thank you so much for your thought and effort, @ArthasModern! I know it's hard to have a PR closed. Hopefully you continue to contribute!

PR #3207, opened 2018-01-02
Updated year range to 2018

fahad19:
**Description:**

Updated year range to 2018 in `LICENSE` file. Quick PR 😬 


benlesh:
Thank you, @fahad19, for catching this!

PR #3210, opened 2018-01-02
fix(throw): Observable.throw() now returns Observable<never>

mernen:
**Description:** `Observable.never()`, `Observable.empty()` and `.ignoreElements()` were previously changed to `Observable<never>`; this patch complements it with `Observable.throw()`.

**Related issue (if exists):** #2640 and its PR #2691

### DIFF
--- a/src/observable/ErrorObservable.ts
+++ b/src/observable/ErrorObservable.ts
@@ -13,7 +13,7 @@ export interface DispatchArg {
  * @extends {Ignored}
  * @hide true
  */
-export class ErrorObservable extends Observable<any> {
+export class ErrorObservable extends Observable<never> {
 
   /**
    * Creates an Observable that emits no items to the Observer and immediately
@@ -68,7 +68,7 @@ export class ErrorObservable extends Observable<any> {
     super();
   }
 
-  protected _subscribe(subscriber: Subscriber<any>): TeardownLogic {
+  protected _subscribe(subscriber: Subscriber<never>): TeardownLogic {
     const error = this.error;
     const scheduler = this.scheduler;
 


benlesh:
I don't think `never` is the right type here, as it does have a value type.  If anything maybe in this case we actually want to give it a type of some sort, because we actually know what it is. But everywhere it could be handled, it's still `any` though, so it probably doesn't matter.

I think we're going to pass on this PR, unless I misunderstand the intent.

cc @kwonoj

mernen:
I believe `never` is the right parameter here, what it means (when the signature of `Observable<never>` is fully expanded) is that the method `next()` will never be called (while saying nothing about `complete()` or `error()`).

The main practical appeal of this change, and why I proposed it in the first place, is that `never` disappears when folded in a sum, while `any` “infects” the sum. For example, `concat(of(1, 2, 3), _throw(Error()))` currently is inferred as `Observable<any>`, while with my patch it becomes `Observable<number>`.

staltz:
As in #2640 I was opposed to the `never` until they informed me that `never` disappears when type summed with others, so this PR seems like the right way to go. A merge or a concat of `Observable<number>` with `Observable<any>` (from throw) would give `Observable<any>` which is inaccurate and the type system knows better. Merging or concatenating `Observable<number>` with `Observable<never>` gives `Observable<number>`.

kwonoj:
Sorry for late peeking, lemme try few things on my end too.

benlesh:
I'm convinced. I think this is right, too. The PR just needs fixed (and rebased) so the build isn't failing.

benlesh:
This change is now included in the `throwError` refactor. Closing this PR in favor of #3288

Issue #2586, opened 2017-05-02
Defer factory should return ObservableInput, not just Subscribable | PromiseLike


felixfbecker:
`Observable.defer()` takes a factory that takes `Subscribable | PromiseLike`. That is very annoying when the value is available synchronously and often results in a lot of boiler plate when using`concat()`. Simplified example:

```ts
const contents = new Map<string, string>();

getFilesInWorkspace()
  .mergeMap(uri =>
    // ensureContent emits no items, just completes once content is ensured
    ensureContent(uri)
      // After that, we can get the content synchronously
      // Need Observable.of() here because just returning an array is not allowed
      .concat(Observable.defer(() => Observable.of(contents.get(uri))))
  )
  .mergeMap(content => ...)
```

kwonoj:
what's type of `contents[uri]`? is it `Array<object>` or `Array<Observable<object>>`? given snippet above does not provide information to actually reproduce issues.

felixfbecker:
Updated example.

kwonoj:
Ok, so now I'm more confused. By subject, I guess `Observable.defer(() => Observable.of(contents.get(uri)))` is part you'd like to discuss, with `// Need Observable.of() here because just returning an array is not allowed` - is that correct?

but,
1. comment says defer factory function returns array but code returns factory function returns string
2. either case it doesn't work anyway cause defer expects function to return Observable object, so wrapping it via Observable.of is expected.

Would you clarify what's exact issue you're experiencing?

felixfbecker:
I would like `defer()` to accept `ObservableInput`, just like `mergeMap`, `catch` etc. do:
```ts
.concat(Observable.defer(() => [contents.get(uri)]))
```

kwonoj:
OK, I see now.

I personally don't agree with those as it makes defer have polymorphic behavior by allowing `ArrayLike<T>` other than Observable itself. As you've already said those can be achievable trivially via current operator / static creation method set which doesn't require huge code boilerplate or technical difficulties to do so. 

I guess I was confused by your subject stating `Defer factory should return...` makes me think some of incompatible type signatures are there around operators, but this seems more about feature suggestion. I'll update label and /cc @benlesh for second opinion.

felixfbecker:
[...]

PR #3211, opened 2018-01-04
fix(typings): the return type of factory of defer should be ObservableInput<T>

Brooooooklyn:
should be able to defer a `ArrayLike` result

```ts
Observable.defer(() => [1, 2, 3])
```


benlesh:
Thanks, @Brooooooklyn!

felixfbecker:
Closes https://github.com/ReactiveX/rxjs/issues/2586

PR #3215, opened 2018-01-05
Fix menu item hover and active in light theme

luketevl:
**Description:**
The selector in body.light-theme. layout-container doens't work, because the class light-theme is apply with layout-container in same element
Example Example


benlesh:
Thank you @luketevl !

luketevl:
It's my pleasure

PR #3218, opened 2018-01-08
fix(debounceTime): synchronous reentrancy of debounceTime no longer swallows the second value

### LINKED ISSUE #2748: `debounceTime` dropping values received in subscription
**RxJS version:** 5.4.2

**Code to reproduce:**

```javascript
const s = new Subject()
s.debounceTime(0).subscribe(n => {
    if (n === 1) {
        s.next(2)
    }
    console.log(n)
})
s.next(1)
```

**Expected behavior:**
Should output `-1-2`

**Actual behavior:**
Output: `-1`

**Additional information:**
`lastValue`/`hasValue` are cleared out after `_next` is called synchronously.
https://github.com/ReactiveX/rxjs/blob/master/src/operator/debounceTime.ts#L98-L102

jayphelps:
Currently the second value is swallowed because `hasValue` is set to `false` in between the scheduling of the second and its execution.

fixes #2748

```js
const results = [];
const source = new Rx.Subject();
const scheduler = new VirtualTimeScheduler();

source.debounceTime(0, scheduler).subscribe(value => {
  results.push(value);

  if (value === 1) {
    source.next(2);
  }
});
source.next(1);
scheduler.flush();

expect(results).to.deep.equal([1, 2]);
```

@kwonoj you had mentioned in #2748 that rxjs doesn't guarantee reentrancy, though I think in this case the fix is simple. Are there reasons to explicitly not do it in this case?

FWIW the regular `debounce` operator [already supports reentrancy](https://github.com/ReactiveX/rxjs/blob/master/src/operators/debounce.ts#L132-L134), so I added a test for it too. This all seems so familiar so perhaps we discussed this years ago and fixed only debounce lol

### DIFF
--- a/spec/operators/debounce-spec.ts
+++ b/spec/operators/debounce-spec.ts
@@ -399,4 +399,20 @@ describe('Observable.prototype.debounce', () => {
         done(new Error('should not be called'));
       });
   });
+
+  it('should debounce correctly when synchronously reentered', () => {
+    const results = [];
+    const source = new Rx.Subject();
+
+    source.debounce(() => Observable.of(null)).subscribe(value => {
+      results.push(value);
+
+      if (value === 1) {
+        source.next(2);
+      }
+    });
+    source.next(1);
+
+    expect(results).to.deep.equal([1, 2]);
+  });
 });
--- a/spec/operators/debounceTime-spec.ts
+++ b/spec/operators/debounceTime-spec.ts
@@ -1,5 +1,7 @@
+import { expect } from 'chai';
 import * as Rx from '../../src/Rx';
 import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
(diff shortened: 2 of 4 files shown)


kwonoj:
@jayphelps as long as it doesn't hurts current code execution, I don't think there's strong reason to explicitly block changes to allow reentrancy. Still in v5 without scheduler reentrancy is not _gauranteed_, means might work but mostly not work in general - not sure just fixing single operator gives huge value. If someone chains another operator doesn't work with reentrancy with fixed debouncetime, then consumer still will be blocked and have to reschedule it.

jayphelps:
Hmm Travis  can’t seem to find Node v4. Maybe they removed it? I googled a bit, can’t find confirmation

kwonoj:
🤔 weird, nvm seems failing to look it up but I also couldn't see anything related.

kwonoj:
seems it was one time glitch. restarting job makes it work.

benlesh:
LGTM

PR #3220, opened 2018-01-09
docs(lettable-operators): add missing quote

caroso1222:
**Description:**


benlesh:
Thank you for catching this, @caroso1222!

Issue #3225, opened 2018-01-13
Reinstate the "prevent unwanted clearInterval" PR


cartant:
I left [a comment](https://github.com/ReactiveX/rxjs/pull/3152#issuecomment-354638226) on the PR that reverted PR #3044. After Zone.js 0.8.19 was released I looked into the reported Angular problem and discovered that PR #3044 was not the cause.

The comment is repeated below.

---

FYI, the problem with 5.5.3 is **unrelated** to PR #3044 and it was **not** the reversion of that PR that solved the problem of Angular's routing module effecting an `EmptyError`.

Instead, the problem appears to have been the way in which the `.js` files in 5.5.3 were generated.

I can reproduce the effected error by taking the `util/EmptyError.js` file from 5.5.3 and overwriting the file in the 5.5.2 distribution.

In fact, if you look at the files that are incorporated into [the Plunk](http://plnkr.co/edit/LW7lfgL0BZmnYzsbMgsD), you will see that `scheduler/AsyncAction.js` - the file modified in PR #3044 - does not even form part of the build.

The 5.5.2 [`EmptyError.js` file](https://unpkg.com/rxjs@5.5.2/util/EmptyError.js) looks like this:

```js
"use strict";
var __extends = (this && this.__extends) || function (d, b) {
    for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
    function __() { this.constructor = d; }
    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
};
/**
 * An error thrown when an Observable or a sequence was queried but has no
 * elements.
 *
 * @see {@link first}
 * @see {@link last}
 * @see {@link single}
 *
 * @class EmptyError
 */
var EmptyError = (function (_super) {
    __extends(EmptyError, _super);
    function EmptyError() {
        var err = _super.call(this, 'no elements in sequence');
        this.name = err.name = 'EmptyError';
        this.stack = err.stack;
        this.message = err.message;
    }
    return EmptyError;
}(Error));
exports.EmptyError = EmptyError;
//# sourceMappingURL=EmptyError.js.map
```

Whereas the 5.5.3 [`EmptyError.js` file](https://unpkg.com/rxjs@5.5.3/util/EmptyError.js) looks like this:

```js
"use strict";
var __extends = (this && this.__extends) || (function () {
    var extendStatics = Object.setPrototypeOf ||
[...]

PR #3152, opened 2017-12-03
fix: Revert "fix(scheduler): prevent unwanted clearInterval (#3044)"

kwonoj:
This reverts commit 7d722d4d36debcf5bb79fa4006f455bd72caeee1.

**Description:**
fixing regressions.
**Related issue (if exists):**
- closes #3151

PR #3226, opened 2018-01-13
fix(scheduler): prevent unwanted clearInterval

cartant:
**Description:**

Reinstates the commits that were made in #3044 and reverted in #3152.

The problem that effected the reversion was not caused by PR #3044. See the related issue (#3225) for details.

**Related issue (if exists):** #3225 #3044


benlesh:
Thanks again, @cartant!

Issue #3229, opened 2018-01-16
Stack overflow in Subscriber.js


3n-mb:
**RxJS version:** 5.5.6 (used in node)

**Code to reproduce:**
```javascript
// we want to use a subject-backed shared observable
const subj = new rxjs.Subject();
const obs$ = subj.asObservable().share();
// want to wrap subscription into a function (library's external api)
const watch = (observer) => {
  const sub = obs$.subscribe(observer.next, observer.error, observer.complete);
  return () => sub.unsubscribe();
};
// using subscription function
const prom = (new rxjs.Observable(observer => watch(observer)))
.take(1).toPromise();

subj.next(5);  // <-- this throws up
prom.then(n => console.log(n));
```
**Expected behavior:** complete without an error, and print 5

**Actual behavior:** line `subj.next(5);` overflows the stack,  throwing the following error:
```
RangeError: Maximum call stack size exceeded
  at SafeSubscriber.Subscriber.next (.../node_modules/rxjs/Subscriber.js:88:42)
  at SafeSubscriber.Subscriber.next (.../node_modules/rxjs/Subscriber.js:90:18)
  at SafeSubscriber.Subscriber.next (.../node_modules/rxjs/Subscriber.js:90:18)
  at SafeSubscriber.Subscriber.next (.../node_modules/rxjs/Subscriber.js:90:18)
...
```

**Additional information:**
- If code is changed to not use subscription wrapping function, then all goes well:
```javascript
// we want to use a subject-backed shared observable
const subj = new rxjs.Subject();
const obs$ = subj.asObservable().share();
// using subscription function
const prom = obs$
.take(1).toPromise();

subj.next(5);  // <-- goes fine
prom.then(n => console.log(n));
```
- It is not obvious that function `watch` is creating a function call loop. Yet, it seems that a combination of `watch` and a subscription via creating a new observer `new Observer(...)` establishes a loop that throws up.

Let me know if code for `watch` function violates rxjs' api. And if not, this is a bug.

kwonoj:
🤔  this looks bit odd, it seems like we're keep trying to create safeSubscriber internally.

as an immediate workaround, you can pass observer directly
```
const watch = (observer) => {
  //const sub = obs$.subscribe(observer.next, observer.error, observer.complete);
  const sub = obs$.subscribe(observer);
  return () => sub.unsubscribe();
};
```

and stack won't blow up.

I may need dig what's happening internally still.

3n-mb:
[...]

Issue #2140, opened 2016-11-17
Maximum Call Stack Exceeded When Passing oberver.next


marwan-at-work:
**RxJS version:** "^5.0.0-rc.3"

If I want to create an observable that subscribes to another observable, I'd have to pass anonymous functions that calls the `observer.next` function. Instead, I'd like to just pass `observer.next` top level. I'm not sure why I get a `Maximum call stack size exceeded`. 

**Code to reproduce:**

Code that causes an error:

```js

import { Observable } from 'rxjs/Rx';

const stream$ = new Observable(observer => {
  Observable.fromPromise(fetch('www.example.com'))
  .subscribe(observer.next, observer.error, observer.complete)
})

stream$.subscribe(console.log, console.log, console.log)
```

Code that passes:

```js
import { Observable } from 'rxjs/Rx';

const stream$ = new Observable(observer => {
  Observable.fromPromise(fetch('www.example.com'))
  .subscribe((x) => {
    observer.next(x);
  }, (err) => {
    observer.error(err);
  }, () => {
    observer.complete()
  });
})

stream$.subscribe(console.log, console.log, console.log);
```

**Expected behavior:**
I should be able to pass `observer.next` as a function to the higher-order `subscribe`

**Actual behavior:**

```
RangeError: Maximum call stack size exceeded
    at SafeSubscriber.Subscriber.next (/node_modules/rxjs/Subscriber.js:87:42)
    at SafeSubscriber.Subscriber.next (/node_modules/rxjs/Subscriber.js:89:18)
    at SafeSubscriber.Subscriber.next (/node_modules/rxjs/Subscriber.js:89:18)
    at SafeSubscriber.Subscriber.next (/node_modules/rxjs/Subscriber.js:89:18)
    at SafeSubscriber.Subscriber.next (/node_modules/rxjs/Subscriber.js:89:18)
    at SafeSubscriber.Subscriber.next (/node_modules/rxjs/Subscriber.js:89:18)
    at SafeSubscriber.Subscriber.next (/node_modules/rxjs/Subscriber.js:89:18)
    at SafeSubscriber.Subscriber.next (/node_modules/rxjs/Subscriber.js:89:18)
    at SafeSubscriber.Subscriber.next (/node_modules/rxjs/Subscriber.js:89:18)
    at SafeSubscriber.Subscriber.next (/node_modules/rxjs/Subscriber.js:89:18)
```

**Additional information:**
Would love to know the reason if this is expected.

[...]

Issue #1981, opened 2016-09-27
observer.next does not allow point free style


rasmusvhansen:
**RxJS version:**
beta12

**Code to reproduce:**
http://jsbin.com/gicizoxefi/edit?js,console

**Expected behavior:**
I would expect 

    sayHi(observer.next);

to work, 

**Actual behavior:**
but it throws a cryptic error.
´´´
error_handler.js:45EXCEPTION: Cannot read property 'isStopped' of nullErrorHandler.handleError @ error_handler.js:45next @ application_ref.js:273schedulerFn @ async.js:82SafeSubscriber.__tryOrUnsub @ Subscriber.js:223SafeSubscriber.next @ Subscriber.js:172Subscriber._next @ Subscriber.js:125Subscriber.next @ Subscriber.js:89Subject.next @ Subject.js:55EventEmitter.emit @ async.js:74onError @ ng_zone.js:120onHandleError @ ng_zone_impl.js:64ZoneDelegate.handleError @ zone.js:207Zone.runGuarded @ zone.js:113(anonymous function) @ zone.js:86
´´´

If I do 

    sayHi(m => observer.next(m));

it works

**Additional information:**

trxcllnt:
JavaScript doesn't explicitly bind prototype methods to their instance, and neither do we. Additionally, the Observer notification methods aren't implemented as closures on the Observer for performance' sake.

rasmusvhansen:
Fair enough. Could the error message be improved maybe? By checking `this` inside the observer.next function and throw if it is not as expected.

trxcllnt:
@rasmusvhansen it's likely lead to a significant performance regression, which would be a non-starter for us. Minimizing operations inside `next` is critical to real-world performance metrics.

I believe there's work planned soon to create special internal Observers (even more sparse, but will be easier for JIT to inline), but enhance the public Observer types at the ends of each sequence. It might be possible to put more error condition checking after that's done cc: @blesh who knows more about that timeline

blesh:
> I believe there's work planned soon to create special internal Observers (even more sparse, but will be easier for JIT to inline), but enhance the public Observer types at the ends of each sequence. It might be possible to put more error condition checking after that's done cc: @blesh who knows more about that timeline

[...]

PR #3230, opened 2018-01-17
Removal of unsanctioned introduction of context

3n-mb:
**Description:**
Observing functions (callbacks) are given to `.subscribe(next, error, complete)` by an outside code. But `rxjs` dictates a context of execution, instead of leaving it `undefined`. It sort of disrespects separation between callbacks that come from outside, and its own internals on the public API of the library.
As a result, there are issues #3229, #1981, #1949, #2140, that happen due to `rxjs` mangling with execution context of externally given callbacks. Library behaves in an unexpected way.

This change fixes an unsanctioned introduction of context to callback.
Now, it is possible, that this weird behaviour was introduced to help some other parts of the lib. Tests should show this. And may be those places should change instead of keeping users of the library puzzled (number of issues times minimum 1-2 hours, plus unreported cases).

**Related issue (if exists):**  #3229, #1981, #1949, #2140

If this will be a no-fix, at least put comment near the place that is pointed in a stack overflow error, which people get. Put some indications for poor devs :cry:

### DIFF
--- a/src/internal/Subscriber.ts
+++ b/src/internal/Subscriber.ts
@@ -165,7 +165,7 @@ class SafeSubscriber<T> extends Subscriber<T> {
     super();
 
     let next: ((value: T) => void);
-    let context: any = this;
+    let context: any = undefined;
 
     if (isFunction(observerOrNext)) {
       next = (<((value: T) => void)> observerOrNext);


benlesh:
Can you please add a test that would fail should this change?

3n-mb:
@benlesh I don't think that a test for this tiny beautification of code failure is needed. I bet, that no-one would need to change this anyway. And if someone does, she will have reasons stronger than mere human error comprehension concern.
Initially, I thought that a reason for this code was something, but, judging from tests passing, original assignment was a way to satisfy typescript compiler (I was in similar places myself :smile: ).

benlesh:
The point was to add a test that is broken *before* the fix, so we can see that the fix works. I can't merge this without a test.

3n-mb:
@benlesh okey, then I need your help here.
- We may have a check that ensures that thrown error is `TypeError`. How to write this? There are no strict runtime types in JS (Java, Rust). So, how do you suggest this check be done? I never done it myself in JS.
- Where should file be placed? Or, into which file should a test case be added?
- What name should I give this test? What convention do you use here?
- Give me a link to some example test, to assimilate form, that is accepted by rxjs.

benlesh:
Really the test should come first. Basically any code that is broken before the change, and some assertions around it. Just keep in mind that it can't run for a long time, because it'll slow down all of our builds.

cartant:
Since [commenting](https://github.com/ReactiveX/rxjs/issues/3229#issuecomment-358156359) on the linked issue, I've seen [this tweet](https://twitter.com/trxcllnt/status/956695620202389504) from @trxcllnt:

> TBF the next fn should be a non-arrow fn & do “this.unsubscribe()” instead. Combining observer + disposable was to help cut down on the number of intermediate disposables allocated in hot paths, but we clearly didn’t always follow our own advice.

The tweet suggests that using the subscriber as the context was a design decision and that it should not be replaced with `undefined`.

Note that this PR does not fix an error. Rather, its change effects a `TypeError` instead of the `RangeError` (see [this explanation](https://github.com/ReactiveX/rxjs/issues/3229#issuecomment-358130109)) if an unbound method is passed like this:

```js
obs$.subscribe(
  observer.next,
  observer.error,
  observer.complete
);
```

Interestingly, the subscriber is also used as the default context for operator subscribers that use projection functions - see the implemetation of [`map`](https://github.com/ReactiveX/rxjs/blob/6.0.0-alpha.2/src/internal/operators/map.ts#L66-L84).

3n-mb:
@cartant Wow!
But the change in this PR didn't break tests. 🤔 
This leads to two options:
- there need to be a test ( @benlesh  ) that will flag current change as incompatible;
- design, described in a tweet, has long been changed, and nobody noticed, cause no tests failed, of cause.

If current design stays, do
- add documentation, and
- place a comment(s) in a place(s) from which errors are coming in  #3229, #1981, #1949, #2140.

3n-mb:
@cartant can you point to a code that illuminates quoted part from tweet:
>  the next fn should be a non-arrow fn & do “this.unsubscribe()” instead. ...

My callback should be responsible to call `unsubscribe`?
*`<rant>`Definitely, coding on tweeter with an artificial words limit isn't helping understanding.`</rant>`*

benlesh:
Yup. There's the breaking change. Although, IMO, Current behavior is also leaking implementation. 

There are other proposals to fix this. A start event, and one that passes the subscription to the next handler as the second argument.

This all requires more thought. But I'd like to eliminate the implementation leaks.

trxcllnt:
@cartant aye, good eye. The goal of context-rebinding in `SafeSubscriber` was to ensure anonymous Observers have same basic capabilities as you'd get by subclassing `Subscriber`, mostly access to the `next`, `error`, `complete`, and `unsubscribe` methods, and the `isUnsubscribed` flag. Later, this was extended to ensure Observable teardown functions are faithfully executed if an error is thrown while synchronously `next'ing` to a sink Observer. More context [here](https://github.com/ReactiveX/rxjs/issues/2525#issuecomment-292015076) and [here](https://github.com/ReactiveX/rxjs/issues/1948#issuecomment-247487466).

Combining Observer and Disposable into a single class was indeed to cut down on intermediate disposables, as well as optimizing the Observer chain by storing references to sink Observers as instance properties rather than closure state. The decision to use the `this` context to unsubscribe isn't my favorite, but was made in the absence of better (more thought out, battle-tested etc.) alternatives that realized the same performance numbers. As @benlesh mentions, this is an area of active research.

To be clear, the challenge here was to ensure Observers could unsubscribe from potentially infinite synchronous sources (again, in the absence of something like cancelation tokens, `Observer#start()`, or being passed the subscription as a second argument):

```js
Observable.range(0, Infinity).subscribe((function(x) {
  if (x > 100) { this.unsubscribe(); }
}));
```

A better way to do the above is to use `take(100)`, but we didn't want to make it _impossible_ to unsubscribe just because someone happens to be doing something differently than how I'd do it.

I know we do have [this test](https://github.com/ReactiveX/rxjs/blob/37d9ea9a8665d0c4b6593da4c237e037e98f66b5/spec/Observable-spec.ts#L401) that validates the context rebinding happens correctly on duck-typed anonymous Observer objects, but I can't seem to find any that validate the context is maintained for PartialObservers.

Also I seem to remember a conversation with @jayphelps a while back about `this.unsubscribe()` not actually working properly, which also may be why we don't have a test for it. I can't seem to find any issues or PRs about it, but that may be a thing we should look into fixing if it's still the case.

trxcllnt:
Also to note, we [bend over backwards](https://github.com/ReactiveX/rxjs/blob/master/src/internal/Subscriber.ts#L170) to ensure the `this` context works like I described above -- it would be way easier if we didn't have to do any of that, so I'd be very receptive to any suggestions that solve all these problems without all the monkeying about with `this` contexts.

3n-mb:
Looks like a rabbit hole goes deep.
As a pedestrian, who tripped over this pothole, can I ask you, guys, to
- put comment on `let context: any = this;`, like a flag near a pothole;
- comment in places that are referred in errors, seen by other pedestrians in #3229, #1981, #1949, #2140, so that folks are less surprised;
- add docs, so that members and owners of this very project are aware. And with more eyes on it, you may faster get a decent replacement, if it is possible, of cause, given constraints of js environment.

benlesh:
Seems like we've worked this out, I'm going to close it.

Issue #3166, opened 2017-12-05
TypeScript errors not reported in tests


cartant:
TypeScript errors are not being reported by the tests, so all of the tests within the `type()` tests are ineffectual.

**RxJS version:** current master

**Code to reproduce:**

Adding an incorrectly typed constant to `debounce-spec.ts`, say, like this:

```ts
import { expect } from 'chai';
import * as Rx from '../../src/Rx';
import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports

const n: number = "1"; // Obviously an error
```

should effect a TypeScript error, but it does not.

**Expected behavior:**

An error should be reported and the tests should fail.

**Actual behavior:**

No error is reported

**Additional information:** -

kwonoj:
I accidently close issue, it's https://github.com/ReactiveX/rxjs/issues/3020. basically we don't run compiler to specs yet.

cartant:
I see. No worries.

cartant:
Closing in favour of #3411.

PR #3234, opened 2018-01-18
fix: correct compilation warnings from missing types in tests

jasonaden:
This PR adds types needed to fix type errors/warnings in tests. It doesn't change any functionality.


kwonoj:
related with https://github.com/ReactiveX/rxjs/issues/3166. Actually there are huge numbers of syntax need to be updated still 😅

PR #3235, opened 2018-01-19
Big refactor to from and of to be "just functions"

benlesh:
- updates from tests
- updates `from` implementation to be just a function, not class-based
- updates `of` implementation to be just a function, not class-based
- removes `ScalarObservable` class in favor of internal function `scalar`
- removes `ArrayObservable` class in favor of `fromArray`
- removes `PromiseObservable` class in favor of `fromPromise`
- removes `IteratorObservable` (which was poorly named anyhow, it should have been IterableObservable, but whatever. LOL) in favor of `fromIterable`
- adds internal `fromObservable` function (for converting observable like)
- removes `ArrayLikeObservable` in favor of `fromArray` which accepts any array like, because the logic was the same.
- removes tests that no longer make sense (instanceof tests, etc)
- updates operators and creation functions that depended on ArrayObservable or ScalarObservable

I did my absolute best not to change tests other than to update them, or remove the assertions that no longer made sense.


kwonoj:
👁 👀

src/internal/observable/fromObservable.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (!scheduler) {
+    return new Observable<T>(subscriber => {
+      const observable: Subscribable<T> = input[Symbol_observable]();
+      const sub = observable.subscribe(subscriber);
+      return () => sub && sub.unsubscribe();
+    });
+  } else {
+    return new Observable<T>(subscriber => {
+      const sub = new Subscription();
+      sub.add(scheduler.schedule(() => {
```
kwonoj: do we need to create new subscription instead of returning subscription as-is from `Scheduler::schedule()`?

src/internal/observable/fromIterable.ts
```diff
@@ -0,0 +1,71 @@
...
+        sub.add(scheduler.schedule(function () {
+          if (subscriber.closed) {
+            return;
+          }
+          let value: T;
+          let done: boolean;
+          try {
+            const result = iterator.next();
+            value = result.value;
+            done = result.done;
```
kwonoj: do we need capture value and `next/complete` separately instead of just do 

```
try {
const {value, done} = iterator.next();
if (!done) { 
 next(value); 
this.schedule();
} else {
complete();
}

} catch {
 error();
}

Comment (benlesh) [reply src/internal/observable/fromIterable.ts]: Yes.. if `next()` synchronously throws, we're going to catch it in the wrong spot if we wrap it in the same try/catch

File: src/internal/observable/fromObservable.ts (hunk 1 above)
Comment (benlesh) [reply src/internal/observable/fromObservable.ts]: yes, because I'm closing over it to manage child subscriptions below. (line 21, for example)

File: src/internal/observable/fromPromise.ts [hunk 3]
```diff
@@ -1,3 +1,30 @@
...
+import { IScheduler } from '../Scheduler';
+import { Subscription } from '../Subscription';
 
-export const fromPromise = PromiseObservable.create;
\ No newline at end of file
+export function fromPromise<T>(input: Promise<T>, scheduler: IScheduler) {
+  if (!scheduler) {
+    return new Observable<T>(subscriber => {
+      input.then(value => {
+        subscriber.next(value);
```
acutmore: Any reason why there is no longer the optimisation of becoming a scalar observable?

Issue #3232, opened 2018-01-18
debounce does not handle scalar observables correctly


Medo42:
**RxJS version:**
5.5.2, but the offending code is unchanged in 5.5.6

**Code to reproduce:**
    
    // Not working
    Observable.interval(100).pipe(
        debounce(i => Observable.of(0))
    ).subscribe(console.log);

    // Working
    Observable.interval(100).pipe(
        debounce(i => Observable.timer(0))
    ).subscribe(console.log);

**Expected behavior:**

The `debounce` should act as a no-op.

**Actual behavior:**

There is an error:

    ERROR TypeError: Cannot read property 'closed' of null
        at DebounceSubscriber._tryNext (debounce.js:102)

The line number refers to 5.5.2, it is this line: https://github.com/ReactiveX/rxjs/blob/5.5.6/src/operators/debounce.ts#L108

**Additional information:**

The call to `subscribeToResult` returns null if no subscription is necessary because the value can be immediately obtained, as in this case of a simple "scalar" observable. `debounce` should handle this return value properly.

The example code is useless of course. My actual use case only returned such an Observable in one case, depending on the value passed in. It is for a loading bar which should only show after at least 200ms of activity, but should be hidden immediately once activity stops.

PR #3236, opened 2018-01-19
fix(debounce): support scalar selectors

cartant:
**Description:**

Adds a failing test and then fixes `debounce` so that selectors returning scalar observables are supported.

**Related issue (if exists):** #3232


cartant:
Coverage decreased? That seems weird.

kwonoj:
it sometimes occurs with those threshold, calculation is not always accurate.

benlesh:
Thanks again, @cartant!

Don't worry about a 0.03% drop in coverage, it's probably just some math that's not in your favor, not necessarily a missed execution path or something.

PR #3239, opened 2018-01-19
refactor(range): move range to be just a function

benlesh:
- updates tests without changing what they're testing
- removes RangeObservable

This is in the same vein as #3235


src/internal/observable/range.ts
```diff
@@ -1,3 +1,80 @@
...
+          break;
+        }
+      } while (true);
+    }
+
+    return undefined;
+  });
+}
+
+export function dispatch(this: Action<any>, state: any) {
```
kwonoj: Let's put 
```
/**
* @internal
*/
```

comment at least.

src/internal/observable/range.ts
```diff
@@ -1,3 +1,80 @@
...
+          break;
+        }
+        subscriber.next(start++);
+        if (subscriber.closed) {
+          break;
+        }
+      } while (true);
+    }
+
+    return undefined;
```
kwonoj: when this can execution path can be reached?

benlesh: This was added to make TS happy, it was warning that not all code paths returned a value.

benlesh:
@kwonoj I marked that dispatch function as `/** @internal */` :+1:

src/internal/observable/range.ts
kwonoj: sounds weird (that TS doesn't check branching correctly) 🤔 , but since it's non excutable anyway.. I'd suggest to change `throw new Error()` (and if anyone actually hit those, give user T-shirts for prize.)

benlesh: But you *can* hit it.

kwonoj: Uhhh yes you are right. I got confused above branch return early for all cases. Good as is then.

PR #3240, opened 2018-01-19
refactor(defer): make defer creation just a function

benlesh:
- Removes DeferObservable
- Updates tests

In the same vein as #3239 and #3235


src/internal/observable/defer.ts
```diff
@@ -1,3 +1,63 @@
...
+export function defer<T>(observableFactory: () => SubscribableOrPromise<T> | void): Observable<T> {
+  return new Observable(subscriber => {
+    let input: SubscribableOrPromise<T> | void;
+    try {
+      input = observableFactory();
+    } catch (err) {
+      subscriber.error(err);
+      return undefined;
+    }
+    const source = input ? from(input) : empty();
```
kwonoj: guess this can return `EMPTY` instead of `empty()` with other PR around empty?

benlesh: Yep.

PR #3247, opened 2018-01-22
feat(empty): empty() returns the same instance

benlesh:
- Adds more robust tests
- Ensures that all calls to empty without a scheduler return a single instance
- Removes EmptyObservable
- Updates other tests and code using EmptyObservable
EmptyObservable
BREAKING CHANGE: `empty()` without a scheduler will return the same
instance every time.
BREAKING CHANGE: In TypeScript, `empty()` no longer accepts a generic
argument, as it returns `Observable<never>`

Background: This is so we can optimize internal functions when they see an instance of `empty()`, as well as reducing the size of the library and improving the typings for TypeScript.


benlesh:
... speciifcally, the breaking changes are for people using typescript and doing `empty<number>()` or the like (which sucks, so I'm sure this is a welcome change), and for people that are, for whatever reason, getting an Observable from `empty()`, then altering it somehow and expecting every call to `empty()` to return a new instance.

benlesh:
@kwonoj ... the discussed change to export `EMPTY` has been added. Currently I'm exporting from `rxjs` rather than `rxjs/create`... it seems like constants should go there. But I'm all for bike-shedding.

This could be a case for merging `rxjs` and `rxjs/create`, I dunno. But if we do that it should be in a different PR.

spec/observables/empty-spec.ts
```diff
@@ -1,16 +1,42 @@
...
     const expected = '|';
-    const e1 = Observable.empty();
+    const e1 = empty();
     expectObservable(e1).toBe(expected);
   });
+
+  it('should return the same instance EMPTY', () => {
+    const s1 = empty();
+    const s2 = empty();
+    expect(s1).to.equal(s2);
```
kwonoj: since we export `EMPTY` constant, maybe `expect(s1).to.equal(s2).to.equal(EMPTY)`?

spec/observables/empty-spec.ts
```diff
@@ -1,16 +1,42 @@
...
+  it('should be synchronous by default', () => {
+    const source = empty();
+    let hit = false;
+    source.subscribe({
+      complete() { hit = true; }
+    });
+    expect(hit).to.be.true;
+  });
+
+  it('should take a scheduler', () => {
```
kwonoj: do we want to verify scheduled `empty` is different to `EMPTY`?

spec/index-spec.ts
```diff
@@ -32,4 +32,8 @@ describe('index', () => {
     expect(index.noop).to.exist;
     expect(index.identity).to.exist;
   });
+
+  it('should export constants', () => {
+    expect(index.EMPTY).to.exist;
```
kwonoj: what's expected usecases of exported `EMPTY` constant? (in other words, end user how do they use it?)

src/internal/observable/empty.ts
```diff
@@ -1,3 +1,59 @@
...
+ * notification.
+ * @static true
+ * @name empty
+ * @owner Observable
+ */
+export function empty(scheduler?: IScheduler) {
+  return scheduler ? emptyScheduled(scheduler) : EMPTY;
+}
+
+export function emptyScheduled(scheduler: IScheduler) {
```
kwonoj: 1. do we need export this? (seems not referenced elsewhere)
2. is this need to be separate fn instead of creating instance in fn directly?
```
empty = (scheduler?: IScheduler) => scheduler? new Observable(...) : EMPTY; 
```

spec/observables/empty-spec.ts
benlesh: Done.

spec/observables/empty-spec.ts
benlesh: This test, combined with the test I just added (`empty()` equals `EMPTY`) should do the trick.

spec/index-spec.ts
benlesh: Basically, users should use `EMPTY` instead of `empty()`... mostly because it gives them the idea that all the empties they're using are the same instance.

kwonoj: not needed in this specific PR, but feeling it's worth to mention in somewhere in docs later.

PR #3250, opened 2018-01-22
Refactor iif impl to be defer-based (plus some file moves/cleanup)

benlesh:
- renames if.ts to iif.ts
- removes IfObservable
- updates `iif` to use `defer` internally


src/internal/observable/iif.ts
```diff
@@ -0,0 +1,93 @@
...
+ * @param {Observable} [trueObservable] An Observable that will be subscribed if condition is true.
+ * @param {Observable} [falseObservable] An Observable that will be subscribed if condition is false.
+ * @return {Observable} Either first or second Observable, depending on condition.
+ * @static true
+ * @name iif
+ * @owner Observable
+ */
+export function iif<T, F>(
+  condition: () => boolean,
+  // TODO(benlesh): use EMPTY once #3247 lands
```
kwonoj: do we need this comment still?

benlesh: nope, thanks! :D

PR #3253, opened 2018-01-23
Fix timer and remove TimerObservable

### LINKED ISSUE #3252: BUG: timer with dueTime of Date will not behave properly if you subscribe to it twice on a delay
**RxJS version:** 5.x

**Code to reproduce:**

```ts
const due = new Date(Date.now() + 3000);
const source = timer(due);

source.subscribe(x => console.log(x))

setTimeout(() => source.subscribe(x => console.log(x)), 1000)
```

**Expected behavior:** Both subscriptions emit at the same time (set by the `due` Date)

**Actual behavior:** They emit one second appart

**Additional information:**
http://jsbin.com/diwagoduma/1/edit?html,js,console,output

This is because the Date calculation is done in the `TimerObservable` constructor and not in `_subscribe`

benlesh:
- fixes #3252 
- Removes  TimerObservable and makes `timer` just a function.
- Updates tests
- Adds test.


benlesh:
@kwonoj ... I don't think so. It's just broken so bad for this edge case I don't think anyone would be reasonably expecting this behavior.

PR #3263, opened 2018-01-26
Functional refactor of `from` and `subscribeToResult`

benlesh:
This is a multistep refactor, please review each commit separately to help you grok the refactor.

`subscribeToResult` was responsible for optimizations when subscribing to inner observables in operators like `mergeMap`, `switchMap`, et al.

`from` is... well... `from`.

This PR unifies the logic from `subscribeToResult` and `from`, because they were basically doing exactly the same thing. The former was taking some object and tying a Subscriber to it.. the later was returning an Observable, which in subscription, takes some object and ties a Subscriber to it.

Notes:

- Error messages are now also unified, so they have changed slightly for `from(wrongThingHere)`

### DIFF
--- a/src/internal/util/subscribeTo.ts
+++ b/src/internal/util/subscribeTo.ts
@@ -1,0 +1,38 @@
+import { Observable, ObservableInput } from '../Observable';
+import { subscribeToArray } from './subscribeToArray';
+import { subscribeToPromise } from './subscribeToPromise';
+import { subscribeToIterable } from './subscribeToIterable';
+import { subscribeToObservable } from './subscribeToObservable';
+import { isArrayLike } from './isArrayLike';
+import { isPromise } from './isPromise';
+import { isObject } from './isObject';
+import { iterator as Symbol_iterator } from '../symbol/iterator';
+import { observable as Symbol_observable } from '../symbol/observable';
+import { Subscriber } from '../Subscriber';
+
+export const subscribeTo = <T>(result: ObservableInput<T>) => {
+  if (result instanceof Observable) {
+    return (subscriber: Subscriber<T>) => {
+        if (result._isScalar) {
+        subscriber.next((result as any).value);
+        subscriber.complete();
+        return undefined;
+      } else {
+        return result.subscribe(subscriber);
+      }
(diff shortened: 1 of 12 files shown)


src/internal/util/subscribeTo.ts
```diff
@@ -0,0 +1,38 @@
...
+import { isArrayLike } from './isArrayLike';
+import { isPromise } from './isPromise';
+import { isObject } from './isObject';
+import { iterator as Symbol_iterator } from '../symbol/iterator';
+import { observable as Symbol_observable } from '../symbol/observable';
+import { Subscriber } from '../Subscriber';
+
+export const subscribeTo = <T>(result: ObservableInput<T>) => {
+  if (result instanceof Observable) {
+    return (subscriber: Subscriber<T>) => {
```
kwonoj: how much perf will hurt if we use `subscribeToObservable` in here instead, and let `subscribeToObservable` have checker to scalar? (like
```
if (obs[symbol.observable) {
 if (!!obs._isScalar) {
 }
 else {
...
 }
}
```

benlesh: instanceof checks are really fast . Especially compared to property lookups. What would be the win if we did change this?

kwonoj: not so much, it'll reduce some codes duplicated between symbolic observable to instance observable. Reason asking if it hurts perf - if it doesn't hurt much, maybe considerable.

benlesh: Oh... yeah, we don't want it to fall down to `subscribeToObservable`, that's meant to subscribe to "lowercase-o" observables where we can't assume anything about them. That means we have to do a call to `Symbol.observable` implementation on whatever the object is, and get out an unknown observable, which will only take an observer, and will only return a plain Subscription object, not what we expect inside of RxJS.

This `instanceof` check is basically saying "Hey, l know this implementation and I trust it, so do the fastest thing"... it's the 99% use case for this code.  I don't think I want to change it.

kwonoj: makes sense.

PR #3265, opened 2018-01-26
refactor(test): updated test helpers to be strongly typed

david-driscoll:
**Description:**
This allows tests to pick more on variation in type definitions when tests are being built

**Related issue (if exists):**

Issue #3254, opened 2018-01-23
Marble diagrams incorrectly including ) from value groups in tick counts


benlesh:
**RxJS version:** 5.x

**Code to reproduce:**

When using the test scheduler create a marble-based observable with a group

More details can be found in this PR: #2932, which we are closing in favor of @kwonoj's upcoming merging of rx-sandbox.

Basically marbles that look like this:

`---(ab)---(cd)--|`
**Expected behavior:**
```
{ time: 30, type: 'N', value: a },
{ time: 30, type: 'N',  value: b },
{ time: 70, type: 'N', value: c },
{ time: 70, type: 'N',  value: d },
{ time: 100, type: 'C' },
```
**Actual behavior:**
```
{ time: 30, type: 'N', value: a },
{ time: 30, type: 'N',  value: b },
{ time: 80, type: 'N', value: c },
{ time: 80, type: 'N',  value: d },
{ time: 110, type: 'C' },
```

**Additional information:**
The above structures are just pseudo code (clearly) lol.

See #2932 

Also use rx-sandbox for marble tests, if you even need them at all.

benlesh:
This is apparently handled by rx-sandbox, which we're going to move into RxJS proper.

Assigning to @kwonoj 

cc @jayphelps

kwonoj:
I'll propose small pr convert one of our tests using new module as devdeps. (eventually it could go in monorepo, or other ways around but as an immediate workaround.)

kazinov:
Hello,

I started writing marble tests for my application and I realised that I can't test observables like this:
```
const observable$ = of(1, 2).concat(of(3, 4).delay(20));
```
If I understood correctly, the current marbles syntax just doesn't allow me to do this. 
I guess it should be ```'(ab)-(cd)'```, but it's not (c and d emitted on frame 50, not 20).

What is the recommend way to work around it? 
Is there any plan to fix this problem or should I start using rx-sandbox? 

I also created a [stack overflow question](https://stackoverflow.com/questions/50406951/how-to-test-observables-which-emit-grouped-events-with-rxjs-marbles)

Thanks!

benlesh:
I believe this is covered by `TestScheduler.run` mode now, and we can close this.

Issue #3231, opened 2018-01-18
Test scheduler interface improvement


kwonoj:
I'm writing it as an issue based on discussions we had on core meeting - 

Testscheduler included in core is working as expected, but also lacks of some ergonomics around setups and assertions - reason there are couple of variant for 3rd party assertion suite out there. I'd like to self-advertise to my own implementation to adapt into current schedules, which satisfies most of complaints around current testscheduler:

- frametime adjustment (`1` by default, customizable)
- test framework agnostic
- long-timeframe support via extended marble syntax
- fine-grained time travelling (`advanceTo` interface as similar v4)
- visual friendly assertion helper function

It could be either make pkg itself as upstream core (once we have monorepo) or either cherrypicking necessary implementation only, or something else - notable thing is core (or separate monorepo-pkg from core) provides de-facto OOTO solution for testing to anyone can easily use.

My current implementation is available at https://github.com/kwonoj/rx-sandbox currently.

/cc @benlesh 
(and also @jayphelps , heard you are looking for something similar? not sure but pinging)

PR #3266, opened 2018-01-26
test(take): refactor test use rxsandbox

kwonoj:
**Description:**

**DONOTMERGE** (Test will fail anyway, so can't be merged)
related to #3254, #3231 and https://github.com/ReactiveX/rxjs/pull/3265. 

This PR is poc attempt to move our test assertion to use `rx-sandbox` instead, provides strong type safety around interfaces, as well as fixes some of known issues in current test assertion.

Unfortunately there is one nasty issue - since rx-sandbox has deps to Rx, there are 2 observable object (from source we'd like to assert, and one rx-sandbox uses internally) and it creates some conflict around scheduling, so couple of subscription tests are failing in here.

Not entirely sure how this can be resolved in quick manner - probably best thing is setting up monorepo and use linked pkg to source allows to have same object over, otherwise we may need to port changes of sandbox into core / or try to expose override rxsandbox to import source directly. Opening up as discussion, since @david-driscoll created PR around current test assertion - we may need to decide which way before make actions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -217,6 +217,7 @@
     "rollup-plugin-inject": "^2.0.0",
     "rollup-plugin-node-resolve": "^2.0.0",
     "rx": "latest",
+    "rx-sandbox": "0.0.10",
     "rxjs": "latest",
     "shx": "^0.2.2",
     "sinon": "^2.1.0",
--- a/spec/operators/take-spec.ts
+++ b/spec/operators/take-spec.ts
@@ -1,140 +1,154 @@
 import { expect } from 'chai';
-import * as Rx from '../../src/Rx';
-import marbleTestingSignature = require('../helpers/marble-testing'); // tslint:disable-line:no-require-imports
-
-declare const { asDiagram };
-declare const hot: typeof marbleTestingSignature.hot;
-declare const cold: typeof marbleTestingSignature.cold;
-declare const expectObservable: typeof marbleTestingSignature.expectObservable;
-declare const expectSubscriptions: typeof marbleTestingSignature.expectSubscriptions;
-
-const Subject = Rx.Subject;
-const Observable = Rx.Observable;
+import { hotObservable, coldObservable, getObservableMessage, expectedObservable, expectedSubscription, rxSandbox } from 'rx-sandbox';
+import { take, mergeMap } from '../../src/operators';
+import { ArgumentOutOfRangeError } from '../../src/internal/util/ArgumentOutOfRangeError';
+import { Observable } from '../../src/internal/Observable';
+import { Subject } from '../../src/internal/Subject';
+import { of } from '../../src/create';
+import { Observer } from '../../src/internal/Observer';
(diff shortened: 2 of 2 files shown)


kwonoj:
@david-driscoll plz check this out too.

benlesh:
So far it looks okay to me... @jayphelps, I know you have keen interest in this, what do you think?

@kwonoj: can we get the changes that exist in rx-sandbox implemented in core? What documentation exists for this? Did you happen to write down any design rationale?

kwonoj:
@benlesh 

1. porting changes over to core is possible, while it requires some of refactorings of course (mainly due to difference between current testscheduler)

2. I didn't write specific rationonals, most of api surface are documented in https://github.com/kwonoj/rx-sandbox though. Primary differences are

- scheduler flush is on its own instead of hookup test assertion (to make test framework agnostic)
- marble is 1 frame (configurable, no big deal)
- few additional features on marbles (as we've discussed before, like long-frame support, `advanceTo` support, etcs)

Mostly `rx-sandbox` is feature parity to TetsScheduler in core.

jayphelps:
@benlesh 

> So far it looks okay to me... @jayphelps, I know you have keen interest in this, what do you think?

I think we should pull them into core, rather than depend on rx-sandbox. If you go with this new API I'd prefer some bikesheddding around the name `getMessages` though off hand I'm not sure what to call it either haha. I really dig the intent of his changes though, so the dev can just do deep array comparisons.

It does feel a bit weird that it's mutable array that changes as virtual time advances. e.g. this example in the docs

```js
const e1 = hot('    --a--b--|');
const subs = s(`    ^       !`);
const messages = getMessages(e1.mapTo('x'));

//at this moment, messages are empty!
expect(messages).to.be.empty;

advanceTo(3);
const expected = e('--x------'); // we're flushing to frame 3 only, so rest of marbles are not constructed

//now values are available
expect(messages).to.deep.equal(expected);
//subscriptions are also available too
expect(e1.subscriptions).to.deep.equal(subs);
```

I'm not sure if this is weird enough for me to object, just was my initial reaction. It's probably fine or even ideal.

kwonoj:
> I think we should pull them into core, rather than depend on rx-sandbox.

Agreed, consider this if POC only. I don't think it's viable to make rx-sandbox as dep, first we'd go for core and eventually create separate pkg once we have core side monorepo.

> I'm not sure if this is weird enough for me to object, just was my initial reaction. It's probably fine or even ideal.

debatable, obviously. For me it's kinda-ok api, as `advanceTo` and `flush` is orthogonal that you can't `advance` against scheduler already flushed (opposite is allowed if I remember correctly, that you advance first then flush out rest). 

This design is based on intention of you're getting notification metadata array directly so you can assert as you want - while given array is mutable so over virtual time passes you get updated values. If there's better ergonomics, this can be discussed over via first PR migrating sandbox impl to core.

/cc @jayphelps ☝️

PR #3277, opened 2018-01-31
chore(package): update package-lock

### LINKED ISSUE #3284: Update dependency-cruiser
Once this lands: https://github.com/sverweij/dependency-cruiser/pull/31

We should try to update dependency-cruiser to the latest version.

kwonoj:
**Description:**
This PR pins all of our dependncies, also check in package-lock.json. Now package lock support becomes quite stable, it works on cross-platform without modifying dependency when just try to install dependencies. We may need bump up huge number of dependencies in here, but that's different story.

- closes #3284


kwonoj:
🤔 just pinning dependency makes test fails?

kwonoj:
pin just removed caret from version range. There should be some semver minor bumps with carets, but still making test fails seems bit odd.

cartant:
How did that test ever pass? The [implementation](https://github.com/ReactiveX/rxjs/blob/master/src/internal/observable/concat.ts#L112-L114) seems to explicitly **not** use the scheduler if there is only one observable?

```ts
if (observables.length === 1 || (observables.length === 2 && isScheduler(observables[1]))) {
  return from(<any>observables[0]);
}
```

kwonoj:
> The implementation seems to explicitly not use the scheduler if there is only one observable?

that's true, test case description is reading bit confusing. It concat's `one` observable + scheduler as param, which takes execution path for arguments length as > 1 .

kwonoj:
phew, bit larger work than I thought.

PR #3285, opened 2018-02-01
chore(BUILD): fix broken build

benlesh:
TypeScript seemed to have an issue with XHR all of a sudden, I presume it's some TS update, quick fix any for now to fix the build


benlesh:
If I had to post mortem this... I'd say it's because TS dictates that all XmlHttpRequest objects MUST have a `response` property on it, therefor the "else" cases in the two changed lines are (according to TS) never hit. Therefor `never`.

PR #3287, opened 2018-02-02
forkJoin as a function, refactor and documentation fixes

benlesh:
- removes `ForkJoinObservable`
- implements `forkJoin` as a function, with same `ForkJoinSubscriber` for optimizations
- Removes some cruft
- Improves typing
- Updates tests
- Updates docs to show modern imports.


src/internal/InnerSubscriber.ts
```diff
@@ -3,13 +3,14 @@ import { OuterSubscriber } from './OuterSubscriber';
...
+ * @internal
  * @ignore
  * @extends {Ignored}
  */
 export class InnerSubscriber<T, R> extends Subscriber<R> {
-  private index: number = 0;
+  private index = 0;
 
-  constructor(private parent: OuterSubscriber<T, R>, private outerValue: T, private outerIndex: number) {
+  constructor(private parent: OuterSubscriber<T, R>, public outerValue: T, public outerIndex: number) {
```
jayphelps: Double checking that going from private to public was desired. Was this changed for tests that aren't in this PR that stopped passing because they accessed these, but passed previously?

jayphelps: Double checking that going from private to public was desired. Was this changed for tests that aren't in this PR that stopped passing because they accessed these, but passed previously?

benlesh: Hmm... I remember making this change as I worked on this incrementally, because there were some ineffeciencies going on around passing these two things around. Given that `OuterSubscriber` is really an internal-only type, there's probably not an issue here. But if it's not being used in this PR, I can revert that part.

jayphelps: Agreed. Your call. 👍

PR #3288, opened 2018-02-02
throwError as just a function

benlesh:
- Removes `ErrorObservable`
- Updates tests
- Renames `throw` files to `throwError` files (impl, spec, etc)
- Updates type to return `Observable<never>`
- Updates docs for modern imports

NOTE: Breaking change for TypeScript users, legacy use `Observable.throw` will require a cast to `any`.


src/internal/observable/throwError.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @param {any} error The particular Error to pass to the error notification.
+ * @param {Scheduler} [scheduler] A {@link IScheduler} to use for scheduling
+ * the emission of the error notification.
+ * @return {Observable} An error Observable: emits only the error notification
+ * using the given error argument.
+ * @static true
+ * @name throw
+ * @owner Observable
+ */
+export function throwError(error: any, scheduler?: IScheduler): Observable<never> {
```
david-driscoll: To avoid a big break you could add a `type` `export type ErrorObservable = Observable<never>`, that will avoid an errors if someone has an explicit type reference (right-hand-side), but not if they have a value reference.

benlesh: That's an interesting idea! I'll have to think about it. I could probably do it elsewhere as well. 🤔

spec/exports-spec.ts
```diff
@@ -19,7 +19,7 @@ import { onErrorResumeNext } from '../src/internal/observable/onErrorResumeNext'
 import { pairs } from '../src/internal/observable/pairs';
 import { race } from '../src/internal/observable/race';
 import { range } from '../src/internal/observable/range';
-import { _throw } from '../src/internal/observable/throw';
+import { throwError } from '../src/internal/observable/throwError';
```
jayphelps: was just `error` considered?

benlesh: @jayphelps ... yes, but the roundness of `throwError` and `catchError` seemed to make sense. But we're in alpha, so it's definitely open to suggestion.

PR #3289, opened 2018-02-02
fix(onErrorResumeNext): no longer holds onto subscriptions too long

benlesh:
Reduces the size of onErrorResumeNext. 

Fixes an issue where `onErrorResumeNext` was holding onto all inner subscriptions for the lifetime of the resulting observable.

Updates documentation

related #2459
closes #3178

### DIFF
--- a/src/internal/observable/onErrorResumeNext.ts
+++ b/src/internal/observable/onErrorResumeNext.ts
@@ -22,40 +22,37 @@ export function onErrorResumeNext<R>(array: ObservableInput<any>[]): Observable<
  *
  * <img src="./img/onErrorResumeNext.png" width="100%">
  *
- * `onErrorResumeNext` is an operator that accepts a series of Observables, provided either directly as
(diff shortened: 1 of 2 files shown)


src/internal/observable/onErrorResumeNext.ts
```diff
@@ -68,67 +61,38 @@ export function onErrorResumeNext<R>(array: ObservableInput<any>[]): Observable<
...
   }
 
-  notifyComplete(innerSub: InnerSubscriber<T, any>): void {
-    this.subscribeToNextSource();
-  }
+  const [ first, ...remainder ] = sources;
 
-  protected _error(err: any): void {
-    this.subscribeToNextSource();
+  if (sources.length === 1 && isArray(first)) {
```
david-driscoll: Is there any perf win/loss if we avoid the above variable allocations by moving this first, and use `isArray(sources[0])`?

benlesh: It would probably be a slight perf loss, as you'd need to look up `first` by index twice.

PR #3292, opened 2018-02-02
fix(Observable): expose pipe rest parameter overload

clydin:
**Description:**
The spread overload of the `pipe` method was not present in the typings for `Observable`.  This PR corrects this.  This is required for TS 2.7 compatibility as it provides stricter parameter checking.


benlesh:
Thank you, @clydin!

PR #3295, opened 2018-02-03
refactor(retryWhen/repeatWhen): Refactor "fooling" the _unsubscribe method

martinsik:
**Description:**

I was going through the `retryWhen` operator code and I noticed that the `notifyNext` is kind of confusing because it was first assigning properties to `null` and then reassigning them back to their original values: https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/retryWhen.ts#L102-L116

```
const { errors, retries, retriesSubscription } = this;

this.errors = null;
this.retries = null;
this.retriesSubscription = null;

this._unsubscribeAndRecycle();

this.errors = errors;
this.retries = retries;
this.retriesSubscription = retriesSubscription;
```

It turned out that this is necessary to "fool" the [`_unsubscribe`](https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/retryWhen.ts#L89) method to do actually nothing (all conditions result to `false` because all properties are `null` and then just get reassigned back). Then I noticed the same code appears in `repeatWhen` as well.

So I thought it'll be easier to reassign only the `_unsubscribe` method itself instead of each property and thus make it a little simpler.

```
this._unsubscribe = null;
this._unsubscribeAndRecycle();
this._unsubscribe = _unsubscribe;
```

All tests are passing. This should also slightly increase performance but I guess it's not really important with this type of operators.

### DIFF
--- a/src/internal/operators/repeatWhen.ts
+++ b/src/internal/operators/repeatWhen.ts
@@ -99,14 +99,12 @@ class RepeatWhenSubscriber<T, R> extends OuterSubscriber<T, R> {
   }
 
   protected _unsubscribeAndRecycle(): Subscriber<T> {
-    const { notifications, retries, retriesSubscription } = this;
-    this.notifications = null;
-    this.retries = null;
-    this.retriesSubscription = null;
+    const { _unsubscribe } = this;
+
+    this._unsubscribe = null;
     super._unsubscribeAndRecycle();
-    this.notifications = notifications;
-    this.retries = retries;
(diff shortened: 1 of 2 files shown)


benlesh:
Nice! Yeah I was looking at this code today like "ugh", but I didn't have time to fix it. Glad I noticed this PR. (I finally have time to go through everything)

benlesh:
Thanks, @martinsik!

PR #3296, opened 2018-02-05
chore(doc): alphabetical order

Shyam-Chen:
**Description:**


benlesh:
Thanks @Shyam-Chen! And thanks @sumitarora  for the review as well.

PR #3297, opened 2018-02-05
Change for loop to for of

nairicode:
Hi guys, I have changes for loop to for of, it's very simple fix but I think for of loop good then for with array length and etc...


nairicode:
sorry for wrong pull request))

PR #3302, opened 2018-02-07
refactory: improve hack for XMLHttpRequest usage with TypeScript 2.7

chuckjaz:
The XMLHttpRequest type is defined to always have the `request`
property. This will lead 2.7 to infer `never` for the else
part of any test of the form `'request' in xhr` as this, as far
as the declarations are concerned, is always true and TypeScript
then infers the else part will `never` run.

This change prevents this inference wihtout removing the type
checking of the other members of xhr such as `responseText`.

**Description:**
See commit text above.

### DIFF
--- a/src/internal/observable/dom/AjaxObservable.ts
+++ b/src/internal/observable/dom/AjaxObservable.ts
@@ -461,12 +461,12 @@ export class AjaxError extends Error {
 function parseXhrResponse(responseType: string, xhr: XMLHttpRequest) {
   switch (responseType) {
     case 'json':
-        if ('response' in xhr) {
+        // HACK(benlesh): TypeScript shennanigans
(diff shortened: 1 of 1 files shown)


JulianWielga:
why not simplier?
 `
    case 'json':
        if (xhr.response) {
          //IE does not support json as responseType, parse it internally
          return xhr.responseType ? xhr.response : JSON.parse(xhr.response || xhr.responseText || 'null');
        } else {
          return JSON.parse(xhr.responseText || 'null');
        }
      case 'xml':
        return xhr.responseXML;
      case 'text':
      default:
        return xhr.response || xhr.responseText;
`
this way there is no 'any', no hacks, we are not loosing type in 'else'. I don't know if this condition covers all needed cases.

PR #3303, opened 2018-02-07
docs(single): Describe case where `undefined` is emitted

### LINKED ISSUE #3182: The Single operator returns undefined when no values match predicate
**RxJS version:**
5.x (all versions and in current master)

**Code to reproduce:**

```js
Rx.Observable.of(1).single(v => v === 2).subscribe(
  v => console.log('next', v),
  e => console.log('error', e),
  () => console.log('complete')
);
```

**Expected behavior:**
Logs an error.

**Actual behavior:**
Logs ('next' 'undefined) ('complete')

**Additional information:**

This behavior matches the rxjs5 spec:

https://github.com/ReactiveX/rxjs/blob/efcd922384116b2a45e7d7859e09cf6da34d7371/spec/operators/single-spec.ts#L143-L154

But does not match the jsdoc description or the behaviour in rxjs 4.

https://github.com/ReactiveX/rxjs/blob/efcd922384116b2a45e7d7859e09cf6da34d7371/src/operators/single.ts#L11-L18

This was an unexpected change while in process of updating an RxJs4 code base to RxJs5.

acutmore:
Expands docs for the Single operator to cover the behaviour where `undefined` is emitted.

Behaviour already has test coverage [test](https://github.com/ReactiveX/rxjs/blob/master/spec/operators/single-spec.ts#L143)

Closes #3182

### DIFF
--- a/src/internal/operators/single.ts
+++ b/src/internal/operators/single.ts
@@ -10,7 +10,8 @@ import { MonoTypeOperatorFunction } from '../../internal/types';
 /**
  * Returns an Observable that emits the single item emitted by the source Observable that matches a specified
  * predicate, if that Observable emits one such item. If the source Observable emits more than one such item or no
- * such items, notify of an IllegalArgumentException or NoSuchElementException respectively.
+ * items, notify of an IllegalArgumentException or NoSuchElementException respectively. If the source Observable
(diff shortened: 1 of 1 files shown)


benlesh:
thank you, @acutmore

Issue #2929, opened 2017-10-09
Remove secondary mappings for flattening operators in v6?


benlesh:
For operators like `mergeMap`, `switchMap`, `concatMap`, etc, we have a secondary selector. It's not used a lot in practice that I've seen. Originally, it added a lot of value in terms of perf because it prevented a closure... however, in practice it makes it harder to isolate errors from observation chains, and in a world with Turbofan, I'm not sure the performance benefit is there. We could eliminate a good amount of code in the library written to support this.

### Use and Alternatives

```ts
Observable.of(1, 2, 3).mergeMap(a => range(0, a), (a, b) => a + b);

/// which is mostly the same as

Observable.of(1, 2, 3).mergeMap(a => range(0, a).map(b => a + b))

// however if there's an error in that second map, you're forced to kill the whole observable...

Observable.of(1, 2, 3)
  .mergeMap(a => range(0, a), (a, b) => { throw new Error('lol') })
  .catch(err => empty())

// rather than this, which is much more common

Observable.of(1, 2, 3)
  .mergeMap(
    a => range(0, a)
      .map(b => a + b)
      .catch(err => empty())
  )


```

wmaurer:
I've not a lot to technically add to this discussion.
I can only say that I very often use the secondary selector, and my team has followed my practice. This change would cause us to have to touch a lot of files across quite a number of projects.
However according to your observations, we're exceptions to the rule ...

benlesh:
@wmaurer hi, Wayne... generally it's not a bad practice. Overall, it's a little better, performance-wise, for older browsers and runtimes because it eliminates some closure. However, that perf difference probably doesn't matter a ton in practice. The biggest thing about it is in situations where it might error (which is admittedly uncommon), you can't isolate the error to the inner observable. So in the two examples below one will stop, and the other will not:

```ts
// With this one, the interval will die on the first error
Observable.interval(1000)
  .mergeMap(
    x => Observable.of(x),
    y => {
      if (y % 2 === 0) throw new Error(y)
      return y
    }
  )
  .catch(err => {
    console.error(err);
    return Observable.empty();
  })
  .subscribe(x => console.log(x));

/// With this one, the interval will live on. (usually desired)
Observable.interval(1000)
  .mergeMap(
    x => Observable.of(x)
      .map(y => {
        if (y % 2 === 0) throw new Error(y)
        return y
      })
      .catch(err => {
        console.error(err);
        return Observable.empty();
      })
  )
[...]

PR #3304, opened 2018-02-07
Removing resultSelectors (WIP)

benlesh:
This is an experimental PR that is aiming to remove all superfluous resultSelector arguments from operators and observable creators.

To start with I've done `mergeMap`, `mergeMapTo`, `switchMap`, `switchMapTo`, `concatMap`, `concatMapTo` and `forkJoin`.

See commit messages for more detail.

This is aimed at smaller size, simplifying the API, simplifying the code base, and simplifying TypeScript typings. 

Just doing it makes it clear that it's a solid decision, as the code cleans up substantially everywhere it's done.

Relates to #2929

### DIFF
--- a/spec/operators/concatMap-spec.ts
+++ b/spec/operators/concatMap-spec.ts
@@ -538,44 +538,6 @@ describe('Observable.prototype.concatMap', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
-<<<<<<< HEAD
-  it('should concatMap many outer to inner arrays, using resultSelector', () => {
-    const e1 =   hot('2-----4--------3--------2-------|');
-    const e1subs =   '^                               !';
-    const expected = '(44)--(8888)---(666)----(44)----|';
-
-    const result = e1.concatMap((value) => arrayRepeat(value, +value),
-      (x, y) => String(parseInt(x) + parseInt(y)));
-
-    expectObservable(result).toBe(expected);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
-  });
-
-  it('should concatMap many outer to inner arrays, and outer throws', () => {
-    const e1 =   hot('2-----4--------3--------2-------#');
-    const e1subs =   '^                               !';
-    const expected = '(22)--(4444)---(333)----(22)----#';
-
-    const result = e1.concatMap((value) => arrayRepeat(value, +value));
-
-    expectObservable(result).toBe(expected);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
-  });
-
-  it('should concatMap many outer to inner arrays, resultSelector, outer throws', () => {
-    const e1 =   hot('2-----4--------3--------2-------#');
-    const e1subs =   '^                               !';
-    const expected = '(44)--(8888)---(666)----(44)----#';
-
-    const result = e1.concatMap((value) => arrayRepeat(value, +value),
-      (x, y) => String(parseInt(x) + parseInt(y)));
-
-    expectObservable(result).toBe(expected);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
-  });
-
-=======
->>>>>>> feat(mergeMap|concatMap|concatMapTo): simplified the signatures
   it('should mergeMap many outer to inner arrays, outer unsubscribed early', () => {
     const e1 =   hot('2-----4--------3--------2-------|');
     const e1subs =   '^            !                   ';
--- a/spec/operators/concatMap-spec.ts
+++ b/spec/operators/concatMap-spec.ts
@@ -538,6 +538,7 @@ describe('Observable.prototype.concatMap', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
+<<<<<<< HEAD
   it('should concatMap many outer to inner arrays, using resultSelector', () => {
     const e1 =   hot('2-----4--------3--------2-------|');
     const e1subs =   '^                               !';
@@ -573,6 +574,8 @@ describe('Observable.prototype.concatMap', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
+=======
+>>>>>>> feat(mergeMap|concatMap|concatMapTo): simplified the signatures
   it('should mergeMap many outer to inner arrays, outer unsubscribed early', () => {
     const e1 =   hot('2-----4--------3--------2-------|');
     const e1subs =   '^            !                   ';
@@ -580,19 +583,6 @@ describe('Observable.prototype.concatMap', () => {
     const expected = '(22)--(4444)--                   ';
 
     const result = e1.concatMap((value) => arrayRepeat(value, +value));
-
-    expectObservable(result, unsub).toBe(expected);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
-  });
-
-  it('should concatMap many outer to inner arrays, resultSelector, outer unsubscribed', () => {
-    const e1 =   hot('2-----4--------3--------2-------|');
-    const e1subs =   '^            !                   ';
-    const unsub =    '             !                   ';
-    const expected = '(44)--(8888)--                   ';
-
-    const result = e1.concatMap((value) => arrayRepeat(value, +value),
-      (x, y) => String(parseInt(x) + parseInt(y)));
 
     expectObservable(result, unsub).toBe(expected);
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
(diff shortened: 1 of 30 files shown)


kwonoj:
@benlesh before further refactoring in code, please consider https://github.com/ReactiveX/rxjs/pull/3050 to land prior. it has been resolved conflict several time and waiting review - while functionally it doesn't change anything, it's blocking for TS user would like to use `strictFn` option which need to be resolved eventually.

benlesh:
@kwonoj that's done.

david-driscoll:
So I wonder if there is a good way to deal with the result selector scenario other than just saying "tough luck".

Going from something like the below just feels worse from a consumers perspective.  I totally get it from the authors perspective.

```typescript
let a4: Rx.Observable<{ o: number; i: string; }> = o.mergeMap(x => [x.toString()], (o, i) => ({ o, i }), 3);
```
to

```typescript
let a6: Rx.Observable<{ o: number; i: string; }> = o.map(o => Observable.from([o.toString()]).map(i => ({ o, i }))).mergeAll(3);
```

I wonder if there is somewhere we can meet in the middle... I'll see if I can figure something out.

kwonoj:
all blocked by mono-repo roadmap, but what we discussed in general is probably considerable to have separate operator package for `legacy` behaviors for opt-in. It's aligned effort of making core slim, and opt in for other intended behaviors.

benlesh:
@david-driscoll the change would be more like:

```
o.mergeMap((o => Observable.from(a.toString()).map(i => ({o, i })), 3);
```

There's no need to use `mergeAll`

luillyfe:
why are they superfluous ? Actually I can find ResultSelector pretty useful, because I can get access to the index of the source observable and inner observable. Also I can make a custom merge of the observables since I can get access to each item produce in each observable. Is there any place where I can get more info about this topic ?

david-driscoll:
@luillyfe The reason they're seen as superfluous is due to the fact they complicate the inner workings of each operator they're running inside, where it causes problems unit testing all the parts of the operator.

the result selector can easily substituted for something like...

```typescript
.mergeMap((outer, outerIndex) => Observable.from(outer).map((inner, innerIndex) => ...));
```

However I agree that is pretty verbose and over the top, so I have a different PR where I'm proposing a helper method to help with the transition (and ensure consistency) #3332 

```typescript
// <type imports here>
import { map } from '../operators/map';
import { from } from '../observable/from';

export function over<O, I, R>(
  selector: (outer: O, index: number) => ObservableInput<I>,
  project: (outerValue: O, innerValue: I, outerIndex: number, innerIndex: number) => R): (outer: O, index: number) => Observable<R> {
  return (outer, outerIndex) => from(selector(outer, outerIndex))
    .pipe(map((inner, innerIndex) => project(outer, inner, outerIndex, innerIndex)));
}
```

the usage is pretty simple...
```typescript
.mergeMap(over((outer, inner, outerIndex, innerIndex) => { ... }));
```

The method is fairly simple but I'd hate to force everyone that wants to use it to define it, so I'll be bringing this up in the next meeting.

benlesh:
@luillyfe they're superfluous because the following is equivalent:


```ts
source$.pipe(
  mergeMap((a, i) => of(a + i), (a, b, i, ii) => a + b + i + ii),
)
```

is the same as:

```ts
source$.pipe(
  mergeMap((a, i) => of(a + i).pipe(map((b, ii) => a + b + i + ii)))
)
```

Only the latter is actually *better* in many cases, as it allows for catching errors before they propagate out of the mergeMap operation.

luillyfe:
Awesome, thank you guys @benlesh @david-driscoll for the clarification.

spec/operators/concatMap-spec.ts
```diff
@@ -538,6 +538,7 @@ describe('Observable.prototype.concatMap', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
+<<<<<<< HEAD
```
jayphelps: @benlesh whoops

spec/operators/concatMap-spec.ts
```diff
@@ -573,6 +574,8 @@ describe('Observable.prototype.concatMap', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
+=======
+>>>>>>> feat(mergeMap|concatMap|concatMapTo): simplified the signatures
```
jayphelps: Whoops

spec/operators/mergeMap-spec.ts
```diff
@@ -283,92 +229,6 @@ describe('Observable.prototype.mergeMap', () => {
...
-    const e1subs =     '^                                                            !';
-    const inner =  cold('----i---j---k---l---|                                        ', values);
-    const innersubs = [' ^                   !                                        ',
-                     '                     ^                   !                    ',
-                     '                                         ^                   !'];
-    const expected =   '-----i---j---k---l-------i---j---k---l-------i---j---k---l---|';
-
-    function project() { return inner; }
-    function resultSelector(oV: string, iV: string, oI: number, iI: number) { return iV; }
-    const result = e1.mergeMap(project, resultSelector, 1);
```
jayphelps: It looks like these tests shouldn't have been removed, instead just removing the usage of resultSelector

spec/operators/mergeMap-spec.ts
```diff
@@ -283,92 +229,6 @@ describe('Observable.prototype.mergeMap', () => {
...
-    const e1subs =     '^                                        !';
-    const inner =  cold('----i---j---k---l---|                    ', values);
-    const innersubs = [' ^                   !                    ',
-                     '         ^                   !            ',
-                     '                     ^                   !'];
-    const expected =   '-----i---j---(ki)(lj)k---(li)j---k---l---|';
-
-    function project() { return inner; }
-    function resultSelector(oV: string, iV: string, oI: number, iI: number) { return iV; }
-    const result = e1.mergeMap(project, resultSelector, 2);
```
jayphelps: ditto

spec/operators/mergeMap-spec.ts
```diff
@@ -283,92 +229,6 @@ describe('Observable.prototype.mergeMap', () => {
...
-    const hotC =   hot('x-xxxx---x-x-x-x-x-xx--x--x-x--x--xxxx-x-----i---j---k---l---|', values);
-    const asubs =      ' ^                   !                                        ';
-    const bsubs =      '                     ^                   !                    ';
-    const csubs =      '                                         ^                   !';
-    const expected =   '-----i---j---k---l-------i---j---k---l-------i---j---k---l---|';
-    const inners = { a: hotA, b: hotB, c: hotC };
-
-    function project(x: string) { return inners[x]; }
-    function resultSelector(oV: string, iV: string, oI: number, iI: number) { return iV; }
-    const result = e1.mergeMap(project, resultSelector, 1);
```
jayphelps: ditto

spec/operators/mergeMap-spec.ts
```diff
@@ -283,92 +229,6 @@ describe('Observable.prototype.mergeMap', () => {
...
-    const hotC =   hot('x-xxxx---x-x-x-x-x-xx----i---j---k---l---|', values);
-    const asubs =      ' ^                   !                    ';
-    const bsubs =      '         ^                   !            ';
-    const csubs =      '                     ^                   !';
-    const expected =   '-----i---j---(ki)(lj)k---(li)j---k---l---|';
-    const inners = { a: hotA, b: hotB, c: hotC };
-
-    function project(x: string) { return inners[x]; }
-    function resultSelector(oV: string, iV: string, oI: number, iI: number) { return iV; }
-    const result = e1.mergeMap(project, resultSelector, 2);
```
jayphelps: ditto

spec/operators/mergeMapTo-spec.ts
```diff
@@ -264,42 +214,6 @@ describe('Observable.prototype.mergeMapTo', () => {
...
-    const values = {i: 'foo', j: 'bar', k: 'baz', l: 'qux'};
-    const e1 =     hot('-a-------b-------c---|                                        ');
-    const e1subs =     '^                                                            !';
-    const inner =  cold('----i---j---k---l---|                                        ', values);
-    const innersubs = [' ^                   !                                        ',
-                     '                     ^                   !                    ',
-                     '                                         ^                   !'];
-    const expected =   '-----i---j---k---l-------i---j---k---l-------i---j---k---l---|';
-
-    function resultSelector(oV: string, iV: string, oI: number, iI: number) { return iV; }
```
jayphelps: ditto

spec/operators/mergeMapTo-spec.ts
```diff
@@ -264,42 +214,6 @@ describe('Observable.prototype.mergeMapTo', () => {
...
-    const values = {i: 'foo', j: 'bar', k: 'baz', l: 'qux'};
-    const e1 =     hot('-a-------b-------c---|                    ');
-    const e1subs =     '^                                        !';
-    const inner =  cold('----i---j---k---l---|                    ', values);
-    const innersubs = [' ^                   !                    ',
-                     '         ^                   !            ',
-                     '                     ^                   !'];
-    const expected =   '-----i---j---(ki)(lj)k---(li)j---k---l---|';
-
-    function resultSelector(oV: string, iV: string, oI: number, iI: number) { return iV; }
```
jayphelps: ditto

benlesh: Oh... yeah, there's the same test below it "without resultSelector"

spec/operators/mergeMapTo-spec.ts
benlesh: exists below "without resultSelector". I refactored to use the pipeable operator anyhow.

benlesh:
Thanks for the review @jayphelps!

Note to future viewers of this PR: The red "X" was only because code coverage dropped 0.03% (or something like that)... So it's acceptable and merged anyhow.

PR #3314, opened 2018-02-12
feat(testscheduler): migrate utility functions for new testscheduler (Stage 1/2)

kwonoj:
**Description:**
Actual work for https://github.com/ReactiveX/rxjs/pull/3266 .

This PR actually migrates new TestScheduler implementation from rx-sandbox, as first step migrates all utility functions it uses along with test cases. Once this PR is merged, will bring actual new testscheduler implementation as following PR to start migrations in existing test cases.

#3231


benlesh:
The hold up with this PR is that I'm waiting to see back from @jayphelps about his work on a new TestScheduler he and I did design work on. I'm not at all sure where the progress is there.

benlesh:
@kwonoj I just caught up with @jayphelps. Things are coming along well with his spike. Given what we've designed that he's working on, I think I would recommend closing this issue in favor of what he has once it lands. 

We should probably schedule a meeting sometime to go over the design with you and the rest of the core team. Might try to include some Angular folks as well.

PR #3320, opened 2018-02-14
feat(typings): fixed some cases where multicast and publish would not return a ConnectableObservable

david-driscoll:
**Description:**
* `multicast` now returns ConnectableObservable when not given a selector.
* `publish` had an issue with overload resolution
* Fixes typing errors in related spec files


benlesh:
More good things, @david-driscoll! Thanks again!

PR #3329, opened 2018-02-20
Fix typing issues in remaining unit tests.

david-driscoll:
### Huge commit ahead
I can break it up into smaller chunks, but I figured it would easier to just rip off the bandaid.

**Description:**
This fixes a few minor typing errors found while correcting all the compiler errors.

#### typescript "linting"
This also adds another lint step `lint_typescript` which just runs `tsc -p tsconfig.vscode.json`. I realize that it isn't really a lint step, but it's not really a build step either.  Lint seemed like the right place to put this.

This causes the typescript compiler to run over the entire project (but does not emit any files). Ensuring that we don't introduce any breaking changes into the unit tests or source files.  My hope is that this will help ensure that we can build the best rxjs typscript consumer experience possible.


david-driscoll:
Fixed the terribly named title (sorry!)

benlesh:
So far we've just been updating the tests as we touch them. I'd like to close this in favor of a more incremental approach _after_ v6 lands. Any objections, @david-driscoll?

benlesh:
Closing this in favor of a more incremental approach.

benlesh:
(Will still use this as a reference, though)

PR #3332, opened 2018-02-21
[wip] feat(helper): new helper method over to replace the resultSelector used by methods like mergeMap

david-driscoll:
**Description:**
See related, this method allows you to (more or less) get similar behavior to how the `resultSelector` works for `mergeMap`, `concatMap`, `switchMap`, etc.

Consider this is a proof of concept, some of the tests are still not working correctly and I need to figure out how to test things a little better.

#3304

### DIFF
--- a/src/internal/helpers/over.ts
+++ b/src/internal/helpers/over.ts
@@ -1,0 +1,13 @@
+import { Operator } from '../Operator';
+import { Subscriber } from '../Subscriber';
+import { Observable } from '../Observable';
+import { OperatorFunction, ObservableInput } from '../types';
+import { map } from '../operators/map';
+import { from } from '../observable/from';
+
+export function over<O, I, R>(
+  selector: (outer: O, index: number) => ObservableInput<I>,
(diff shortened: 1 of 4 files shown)


src/internal/helpers/over.ts
```diff
@@ -0,0 +1,13 @@
+import { Operator } from '../Operator';
+import { Subscriber } from '../Subscriber';
+import { Observable } from '../Observable';
+import { OperatorFunction, ObservableInput } from '../types';
+import { map } from '../operators/map';
+import { from } from '../observable/from';
+
+export function over<O, I, R>(
+  selector: (outer: O, index: number) => ObservableInput<I>,
+  project: (outerValue: O, innerValue: I, outerIndex: number, innerIndex: number) => R): (outer: O, index: number) => Observable<R> {
```
benlesh: While I'm still debating whether or not we want to add this directly to the core library or if it should be something standalone, I wanted to point out that the names are flipped as opposed to Rx 5.x... should be `project`, then `selector`.

benlesh:
Given that we've put the resultSelectors back as deprecated for now, I think we can close this one.

PR #3345, opened 2018-02-23
Update Rx.ts

valeriob:
Allow type definitions to be used in UMD projects, or with no module system.
https://github.com/zspitz/TypeScript-Handbook/blob/release-2.0/pages/Modules.md#umd-modules

### DIFF
--- a/src/Rx.ts
+++ b/src/Rx.ts
@@ -225,3 +225,5 @@ export {
     Scheduler,
     Symbol
 };
+
+export as namespace Rx;


kwonoj:
Afaik it doesn't fully allow umd type support for our codebases. We are waiting https://github.com/Microsoft/TypeScript/issues/4433 to get compiler side support - would you share how did you test this PR?

valeriob:
I'm sorry i'm not familiar with the code base, i just made that change on my local node_module folder and it works. I guess i'll stick with an older version :(

Issue #3349, opened 2018-02-26
fromEvent not removing event listener when bound using event options


crisbeto:
**RxJS version:**
5.5.6

**Code to reproduce:**
```ts
const subscription = fromEvent<KeyboardEvent>(document.body, 'keydown', true).subscribe(() => {});

// To check the bound keyboard listeners, use `getEventListeners(document.body).keydown` in the Chrome dev tools.

subscription.unsubscribe();

// Run `getEventListeners(document.body).keydown` again to check the amount of event listeners.
```

**Expected behavior:**
The `keydown` event should be removed.

**Actual behavior:**
The `keydown` event isn't being removed.

**Additional information:**
The issue only seems to happen when passing in the third argument to `fromEvent`. It passes the argument along to `addEventListener` correctly, however the same set of arguments have to be passed back to `removeEventListener` in order for the event to be removed when unsubscribing.

Related issue: https://github.com/angular/material2/issues/10143
Related PR: https://github.com/angular/material2/pull/10160

PR #3350, opened 2018-02-27
fix(fromEvent): pass options in unsubscribe

cartant:
**Description:**

Fixes the problem reported in #3349. In Chrome, at least, if options are passed in the call to `addEventListener`, but not in the call to `removeEventListener`, the listener is not removed.

The PR adds an expectation to an existing test and passes the options in the call to `removeEventListener`, too.

**Related issue (if exists):** #3349

### DIFF
--- a/spec/observables/fromEvent-spec.ts
+++ b/spec/observables/fromEvent-spec.ts
@@ -132,16 +132,17 @@ describe('fromEvent', () => {
     });
   });
 
-  it('should pass through options to addEventListener', () => {
-    let actualOptions;
+  it('should pass through options to addEventListener and removeEventListener', () => {
+    let onOptions;
+    let offOptions;
     const expectedOptions = { capture: true, passive: true };
 
     const obj = {
       addEventListener: (a: string, b: EventListenerOrEventListenerObject, c?: any) => {
-        actualOptions = c;
+        onOptions = c;
       },
       removeEventListener: (a: string, b: EventListenerOrEventListenerObject, c?: any) => {
-        //noop
+        offOptions = c;
       }
     };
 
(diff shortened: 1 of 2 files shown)


kwonoj:
So actually MDN describes this topic already: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener 
```
While addEventListener() will let you add the same listener more than once for the same type if the options are different, the only option removeEventListener() checks is the capture/useCapture flag. Its value must match for removeEventListener() to match, but the other values don't.
```

+
 
```
It's worth noting that some browser releases have been inconsistent on this, and unless you have specific reasons otherwise, it's probably wise to use the same values used for the call to addEventListener() when calling removeEventListener().
```

Change looks legit to me.

src/internal/observable/FromEventObservable.ts
```diff
@@ -206,7 +206,7 @@ export class FromEventObservable<T> extends Observable<T> {
     } else if (isEventTarget(sourceObj)) {
       const source = sourceObj;
       sourceObj.addEventListener(eventName, <EventListener>handler, <boolean>options);
-      unsubscribe = () => source.removeEventListener(eventName, <EventListener>handler);
+      unsubscribe = () => source.removeEventListener(eventName, <EventListener>handler, <boolean>options);
```
crisbeto: The `options` aren't necessarily a `boolean`, they can be an object as well.

cartant: I'm unaware of why the `boolean` assertion was used in the original code's call to the `addEventListener` method. Perhaps the typings in an earlier version of TypeScript were not representative of the API? The [current typings](https://github.com/Microsoft/TypeScript/blob/v2.7.2/lib/lib.dom.d.ts#L14638-L14645) seem okay, so I guess this could be changed. However, it is only a type assertion and has no runtime effect.

benlesh:
Thanks, @cartant! :)

cartant:
@benlesh Before I rebased this PR - to accommodate the refactorings in master - it would have be cherry-pickable into stable. Do you see any need for another PR to fix this in stable, too?

Issue #2916, opened 2017-10-06
Eliminate @reactivex/rxjs?


benlesh:
Currently we release two packages:

- `@reactivex/rxjs`
- `rxjs`

Right now, the vast, vast majority of users are using `rxjs`, not `@reactivex/rxjs`. The major difference between the two *used to be* that the `@reactivex/rxjs` one had more builds in it and the original source... *however* as of 5.5, we have all module build styles in `rxjs` to support bundlers, *and* we have the src to support source maps. So they're effectively not that much different anymore.

Should we deprecate `@reactivex/rxjs` in favor of `rxjs`?

kwonoj:
do not forget we have rxjs-es too: https://github.com/ReactiveX/rxjs/issues/2218

I'm voting to deprecate except main one.

cartant:
This is kinda related to #2913. That issue mentions package size as a motivator. Something to consider is that if you are going in the direction of multiple, related packages, having them all under the `@reactivex` scope might have some advantages.

rgbkrk:
With all the build styles available in the `rxjs` name, I'm in favor of deprecating `@reactivex/rxjs`. It's what I use today and teach others to use.

benlesh:
We want to do deprecate @reactivex/rxjs still. We can do that once the publishing/build changes land.

benlesh:
cc @kwonoj can you link your build-related PR to this?

kwonoj:
https://github.com/ReactiveX/rxjs/pull/5032

kwonoj:
I'm hoisting this as agenda item to check status quo before closing.

benlesh:
Core Team Meeting:

- We still need to support this for version 6.x.
- Change the NPM package README.md for just @reactivex/rxjs to have some big "deprecated" message.

Issue #3212, opened 2018-01-04
Make rxjs a side-effect free package (for webpack v4)


IgorMinar:
Webpack is going to introduce a new feature called side-effect free module optimization in v4 (currently in alpha).

This optimization requires packages to opt-in to it and promise to contain only side-effect free ES modules by adding `side-effects: false` flag into the package.json. More info: https://github.com/webpack/webpack/tree/next/examples/side-effects

This is mostly true for rxjs except for the special modules that monkey patch the prototype (specifically `rxjs`, `rxjs/add/operator/*`, `rxjs/add/observable/*`).  These modules are being deprecated anyway in v6 in favor of lettable operators and creation functions (observable factories), so we should just move that code into a separate package with reexports from the `rxjs` package to avoid breaking the world.

The redirects could look something like this:

`rxjs/add/operator/foo` in the rxjs package (marked as side-effect free)
```
   import from 'rxjs-deprecations/add/operator/foo';
```

`rxjs-deprecations/add/operator/foo` in the rxjs-deprecations package (not marked as side-effect free)
```
import {Observable} from 'rxjs'

Obeservable.prototype.foo = ....
```

This way webpack does it's magic for people that use the new module ids (see #3145) and for people that still require old paths due to legacy code, all they have to do is to install rxjs-deprecations into their app (we could also consider making it a dependency of rxjs and auto installing it but that could cause issues and would promote the bad side-effecty behavior which is dangerous, so for that reason alone I'd prefer the legacy package to be an opt-in)
  
Note: this issue came out of discussion in https://github.com/angular/angular-cli/issues/9069#issuecomment-355394912

IgorMinar:
btw the credit for the extra package and redirect idea goes to @jasonaden. I think it's a pretty rad idea.

benlesh:
Would this effect the Symbol.observable ponyfill we have in RxJS? It technically has a side-effect if it polyfills.

IgorMinar:
@benlesh can you point me to the code and explain how/when the polyfill does it's thing? I think we should be able to make it work, but it's good that you pointed it out.

kwonoj:
@IgorMinar I believe this is it : https://github.com/ReactiveX/rxjs/blob/master/src/symbol/observable.ts

IgorMinar:
thanks! in that case this is the only line with a side-effect:

[...]

PR #3356, opened 2018-03-01
fix(ESM): Add operators/package.json for ESM support, fixes #3227

### LINKED ISSUE #3227: Should libraries that depend on rxjs use path-mappings/tree shaking?
👋  This might at first glance seem like an obvious question, but let me explain:

By default in v6 if you import anything it's going to resolve the files from the project root. e.g. `import { Observable } from 'rxjs'` is going to come from `node_modules/rxjs/Rx.js` which re-exports it from `node_modules/rxjs/internal/Observable.js`.

If a library uses the [path-mappings and tree shaking](https://github.com/ReactiveX/rxjs/blob/master/doc/pipeable-operators.md#build-and-treeshaking) they would use custom resolve overrides. So in that case, Observable would come from `node_modules/rxjs/_esm5/index.js` which re-exports it from `node_modules/rxjs/_esm5/internal/Observable.js`. Note the fact that this is inside `rxjs/_esm5`.

Then let's say that someone imports my library (e.g. https://redux-observable.js.org) which had done this, but **the consumer does not set up their own webpack resolve overrides** so they'll get Observable from `node_modules/rxjs/internal/Observable.js` and now we have two copies of rxjs in the same project. `Observable1 instanceof Observable2 === false`. If they use pipeable operators they probably won't immediately run into errors, if ever, but their bundle size will be silently larger and might later get cases where `Observable1 instanceof Observable2 === false`, which is quite the hard thing to diagnose for most.

If the library doesn't use path-mappings, the same problem exists in reverse; if the library consumer do use it, they'll again get their own copy of RxJS.

***

This was brought up previously when we had rxjs-es (because a redux-observable user hit it) and discussed including the esm5 versions, but I don't recall a solution ever being found.

Yes, this sucks. AFAIK the only real solution would be have the esm5 version in its own package (or make it the default package and put the other stuff in another). Interestingly, if that was done I think people wouldn't need to have path-mappings and instead could just alias `"rxjs": require.resolve('rxjs-esm5')` or whatever it would be called. This could still give you two copies but to me feels like it reduces the likelihood as long as libs set their peerDep correctly.

🔥

jayphelps:
See #3227

I didn't do anything for `@reactivex/rxjs` package because I'm hopeful for its elimination (#2916) and I'm not entirely sure how it works lol.

### DIFF
--- a/.make-packages.js
+++ b/.make-packages.js
@@ -40,7 +40,9 @@ fs.removeSync(PKG_ROOT);
 let rootPackageJson = Object.assign({}, pkg, {
   name: 'rxjs',
   main: './index.js',
-  typings: './index.d.ts'
+  typings: './index.d.ts',
+  module: './_esm5/index.js',
+  es2015: './_esm2015/index.js'
 });
 
 // Get a list of the file names. Sort in reverse order so re-export files
@@ -102,6 +104,10 @@ copySources(ESM2015_ROOT, ESM2015_PKG, true);
 fs.copySync('./tsconfig.json', PKG_ROOT + 'src/tsconfig.json');
 
 fs.writeJsonSync(PKG_ROOT + 'package.json', rootPackageJson);
+fs.copySync('src/operators/package.json', PKG_ROOT + '/operators/package.json');
+fs.copySync('src/ajax/package.json', PKG_ROOT + '/ajax/package.json');
+fs.copySync('src/websocket/package.json', PKG_ROOT + '/websocket/package.json');
+fs.copySync('src/testing/package.json', PKG_ROOT + '/testing/package.json');
 
 if (fs.existsSync(UMD_ROOT)) {
   fs.copySync(UMD_ROOT, UMD_PKG);
--- a/src/ajax/package.json
+++ b/src/ajax/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "rxjs/ajax",
+  "typings": "./index.d.ts",
+  "main": "./index.js",
+  "module": "../_esm5/ajax/index.js",
+  "es2015": "../_esm2015/ajax/index.js"
+}
--- a/src/operators/package.json
+++ b/src/operators/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "rxjs/operators",
+  "typings": "./index.d.ts",
+  "main": "./index.js",
+  "module": "../_esm5/operators/index.js",
+  "es2015": "../_esm2015/operators/index.js"
+}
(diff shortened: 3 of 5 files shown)


.make-packages.js
```diff
@@ -40,7 +40,9 @@ fs.removeSync(PKG_ROOT);
 let rootPackageJson = Object.assign({}, pkg, {
   name: 'rxjs',
   main: './index.js',
-  typings: './index.d.ts'
+  typings: './index.d.ts',
+  module: './_esm5/index.js',
+  es2015: './_esm2015/index.js'
```
jayphelps: It's hard to keep everything straight, but AFAICT this also makes the whole path-mappings.js thing unnecessary if you're using a newish version of webpack or rollup.

Would be cool to still include instructions (and maybe path-mappings still) for people who are stuck on older build pipelines.

simonbuchan: Yes, if users only import from `"rxjs"` and `"rxjs/operators"`, importing `"rxjs/index"` or (more realistically) `"rxjs/Rx"` would not be mapped.

simonbuchan:
This says "fixes #3227", but it doesn't handle webpack native tree-shaking? I'd have to check the quality difference between having `"sideEffects": false` and not in webpack 4, I think with the new uglifyjs in there too it should be able to optimize without it, but it might be slower or something.

jayphelps:
>  it doesn't handle webpack native tree-shaking?

Can you clarify?

simonbuchan:
See [the example](https://github.com/webpack/webpack/tree/f6e366b4be1cfe2770251a890d93081824789209/examples/side-effects)
It's why I called out "webpack native" tree shaking and mentioned uglify should still work - but presumably the option exists for a reason.

jayphelps:
@simonbuchan oh yeah `sideEffects: false` might be able to be supported but that's being tracked in #3212 not #3227. Perhaps confusing the two? 😄

jasonaden:
Agreed on #3212 vs #3227. This PR should close #3227 (as long as a `package.json` gets added for other secondary entry points such as `ajax, websocket`, and `testing`).

#3212 will have another PR shortly (I've got that one).

simonbuchan:
@jayphelps Probably! But #3227 reads to me an umbrella issue of "make it just work™".

@jasonaden Is `"rxjs/Rx"` going away too?

jayphelps:
@jasonaden 

> I think we also need a similar package.json file in ajax, testing, and websocket, right?

Ohhh right! Thanks for catching that!

@simonbuchan sorry for the confusion! My primary intention with that ticket was how to prevent two copies of RxJS. 🕺

jayphelps:
@jasonaden I've updated to include package.json's for ajax/websocket/testing

PR #3362, opened 2018-03-02
Refactor fromEvent

benlesh:
- makes `fromEvent` just a function
- updates tests
- removes `resultSelector` from `fromEvent`
- `fromEvent` will now emit an array if the event callback has multiple arguments.
- updates docs


spalger:
Since resultSelector is deprecated, if people want a consistent result from `fromEvent()` you recommend they:

```js
.pipe(map(value => Array.isArray(value) ? value[0] : value))
```

This reminds me of https://github.com/petkaantonov/bluebird/issues/307

I personally think bluebird made the right decision to require users explicitly choose between single or multi-arg support. Sort of like how RxJS used to work... Feels like a tiny step back for RxJS.

j2L4e:
@benlesh Can you elaborate on why a single argument stays as-is instead of being wrapped into an array? If you get a single argument that is an array of its own, there seems to be no way to tell it apart from a multiple argument array.

benlesh:
@spaiger @j2L4e.. to be clear, `fromEventPattern` has had this behavior for years. Long before my involvement with RxJS. `fromEvent` in RxJS 5 accidentally had **no** support for multiple arguments. So we added it in to be consistent with `fromEventPattern` and other versions on RxJS. Nothing "changed" or "stepped backwards".

I'm happy to look at an issue and discuss with the larger community ways to improve this. 

We could, of course, undeprecate the result selector. There's a lot of options. 

File another issue

PR #3368, opened 2018-03-02
feat(throwIfEmpty): adds throwIfEmpty operator

benlesh:
This is a new, simple, operator that will emit an error if the source observable completes without emitting a value. This primitive operator can be used to compose other operators such as `first` and `last`, and is a good compliment for `defaultIfEmpty`.

This is a step toward refactoring a few other operators to be smaller.


benlesh:
Having this operator basically means that `first` would become:

`first()` => `pipe(take(1), throwIfEmpty())`
`first(predicate)` => `pipe(filter(predicate), take(1), throwIfEmpty())`
`first(undefined, defaultValue)` => `pipe(take(1), defaultIfEmpty(defaultValue))`
`first(predicate, defaultValue)` => `pipe(filter(predicate), take(1), defaultIfEmpty(defaultValue))`

`last` would be similar, only with `takeLast`

src/internal/operators/throwIfEmpty.ts
```diff
@@ -0,0 +1,41 @@
...
+ *   next() { console.log('The button was clicked'); },
+ *   error(err) { console.error(err); },
+ * });
+ * @param {Function} [errorFactory] A factory function called to produce the
+ * error to be thrown when the source observable completes without emitting a
+ * value.
+ */
+export const throwIfEmpty =
+  <T>(errorFactory: (() => any) = defaultErrorFactory) => tap<T>({
+    hasValue: false,
```
jayphelps: TIL this works

PR #3369, opened 2018-03-02
chore(BUILD): fix broken build (#3285)

alexeagle:
TypeScript seemed to have an issue with XHR all of a sudden, I presume it's some TS update, quick fix any for now to fix the build

cherry-pick of #3285 to stable branch

/cc @chuckjaz


alexeagle:
FYI @kwonoj if Angular 6 ships with RxJS 5.5 we'd need to be able to compile it with TS 2.7, so we'd like to cherry-pick Ben's fix to the stable branch

PR #3372, opened 2018-03-02
 test(Subject): add action after AsyncSubject completion

maratvmk:
test(Subject): add action after AsyncSubject completion


benlesh:
I thought it might be a redundant check, but it seems legit.

Thanks, @maratvmk

Issue #3374, opened 2018-03-02
fromPromise no longer exported in master (v6), the replacement from creation function is too heavy weight


IgorMinar:
**RxJS version:**
6.0.0-alpha.3

**Code to reproduce:**
```
import { fromPromise } from 'rxjs';
```

**Expected behavior:**
should work just as it did with v5.5 (but different location)

**Actual behavior:**
no longer exported, the only alternative is `from` but that operator is significantly [more heavier](https://github.com/ReactiveX/rxjs/blob/9684e4648f49f156e1a771f429c76e00d343d049/src/internal/observable/from.ts) because it supports all kinds of imports. I suggest that you export just the specific `fromXXX` creation methods and don't export `from` because it's too big.

Alternatively export both individual `fromXXX` functions as well as `from` but then you need to document that client side developers should not use `from`, which makes things very confusing. It's always better to have just one way to do a thing - and it should be the right way.

**Additional information:**

benlesh:
So you're aware, using `mergeMap`, `switchMap`, `concatMap`, etc, will pull in most of the logic from `from` as well. So I'm not sure this is a big win for 95% of apps.

IgorMinar:
ok. sounds like splitting up this is not worth it. we should just be
consistent and intentional about which `fromXXX` operators we expose. I
suggest that we just alias `fromPromise` to `from` in the compat library
and remove it from rxjs proper. Having two operators that accept the same
input and return the same output is confusing. There should be just one.
Your call.

bellizio:
It looks like this did not make the cut for the 6.0 release.

Should we use `import { fromPromise } from 'rxjs/internal/observable/fromPromise';` for now?

snowyu:
just use `from`  instead of `fromPromise`.

```js
import { from } from 'rxjs';
```

PR #3382, opened 2018-03-06
Export froms

benlesh:
In this spirit of #3374, if we're exporting `fromPromise`, we should export these as well.

HOWEVER: I still find the "smallness" of fromPromise to be dubious, as things like `defer`, `switchMap`, `concatMap`, `mergeMap`, `exhaustMap`, `forkJoin`, etc, all pull in the majority of `from`'s logic.

cc @IgorMinar @jasonaden @kwonoj 

... this is, of course, at the _expense_ of more API surface area.


jasonaden:
After further discussion, this should be closed and the `fromPromise` fix (#3375) should be reverted.

PR #3384, opened 2018-03-06
Side effect free imports & backwards compatibility APIs

jasonaden:
This PR will address a couple things at once:

* Make `rxjs` side-effect-free to take advantage of Webpack 4 and significantly decrease application bundle sizes
* Adds an `rxjs-compat` package to be separately distributed to NPM
* `rxjs-compat` holds imports that produce side-effects (any code that monkey patches through `add` operator will live here)
* Adds a `legacy-reexports` directory used at build time to make imports such as `rxjs/Observable` still work (though devs will need to install `rxjs-compat` to make these imports behave correctly)

Basically, this makes the default case for Rx side-effect free, but includes many breaking API changes. The `rxjs-compat` package mitigates these breaking changes.


simonbuchan:
So is `rxjs-compat` *because* of `"sideEffects": false`, or is that an *excuse* for it? Because with webpack 4.0.1 you could instead use something like `"sideEffects": ["add/**", ...]`...

jasonaden:
@simonbuchan Intriguing note there. I can't find that anywhere in Webpack docs, and [`WebpackOptions.json`](https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json) says that option is a `boolean`. Can you point me to what you're referring to?

jasonaden:
@simonbuchan Also, there is a reason for the `rxjs-compat` package. This is so users will have to take explicit action to get the old behavior, thus opting in.

simonbuchan:
Huh, is a very recent change, but it's not even mentioned in the release changelog. Here's the PR: https ://github.com/webpack/webpack/pull/6536
Perhaps deemed not ready for primetime?
Re. `rxjs-compat`, what I meant was would you want to make these breaking changes if you didn't have to for sideEffects? Sounds like you do, I'm mostly neutral so long as we have less duplicate import names confusing IDEs.

jasonaden:
Thanks for the reference commit. Yes, we would want to have the separate package either way, but good to know on the new feature.

PR #3387, opened 2018-03-06
feat(Symbol.observable): is no longer polyfilled

benlesh:
BREAKING CHANGE: RxJS will no longer be polyfilling Symbol.observable. That should be done by an actual polyfill library. This is to prevent duplication of code, and also to prevent having modules with side-effects in rxjs.


src/internal/symbol/observable.ts
```diff
@@ -1,32 +1,27 @@
...
-  return $$observable;
 }
 
-export const observable = getSymbolObservable(root);
+/* NOTE: Warning users that they don't have a Symbol.observable
+   polyfill. It's not something we'd throw on, because it's not
+   really *required*, however, they should be warned to give them a
+   clue as to why they might be seeing errors when trying to convert
+   ObservableLikes to Observables */
+if (!Symbol || !Symbol.observable) {
```
jayphelps: This will need to do `typeof Symbol` because otherwise it still will do `Symbol is not defined`. See #3394 for reference example

src/internal/symbol/observable.ts
```diff
@@ -1,32 +1,27 @@
...
+  console.warn('RxJS: Symbol.observable does not exist');
+}
 
 /**
- * @deprecated use observable instead
+ * @deprecated use {@link observable}
  */
-export const $$observable = observable;
+export const $$observable = Symbol && Symbol.observable || '@@observable';
```
jayphelps: Same as above, prolly easiest to wrap in a function like done in #3394 so only need to check once. Also will allow you to easily make sure the type is of `symbol` instead of `any` as it is in this case.

jayphelps:
btw I think these symbols are no longer publicly exported except for in the UMD build. Which is intended?

I'm not entirely sure if we _should_ be exporting any of them. If we did, seems like `Symbol.observable` would be the only one we have any business doing so but even then if someone wants to use one of them without native support, they should include a polyfill IMO and then rxjs will end up using it transparently anyway so no need to access the one rxjs creates.

I can see a possible argument about allowing libraries that use RxJS to add support for `Symbol.observable` without including the polyfill automatically (which would mutate Symbol, forcing it on their users). My gut says that in those cases they too should either follow our lead with `"@@observable"` string as a fallback, or just choose not to implement support for `Symbol.observable` when it's not defined. In fact, that's probably what _we_ should be doing too? I'm not sure when the `"@@observable"` approach is useful?

src/internal/symbol/observable.ts
benlesh: Wrapping it in a function will only make it easier to test. The module body itself should only be called once, no?

benlesh:
I don't disagree, @jayphelps, I think we should only be implementing/executing some logic if Symbol.observable is defined... this was just the quickest incremental step.

src/internal/symbol/observable.ts
jayphelps: @benlesh Partially, though it also makes it easier to only need to do the `typeof Symbol` check once.

PR #3389, opened 2018-03-07
feat(Symbol.iterator): no longer polyfilled

benlesh:
BREAKING CHANGE: We are no longer polyfilling `Symbol.iterator`. That would be done by a proper polyfilling library


benlesh:
NOTE: The only worry I have with this change is that we're no longer watching for the weirdness with older versions of Firefox.. however that should be covered by a proper polyfill.

cc @IgorMinar @jasonaden

src/internal/symbol/iterator.ts
```diff
@@ -1,38 +1,16 @@
...
-        }
-      }
-    }
-    return '@@iterator';
-  }
+/* NOTE: Warning users that they don't have a Symbol.iterator
+   polyfill. We don't want to throw on this, because it's not required
+   by the library. However it will provide clues to users on older
+   browsers why things like `from(iterable)` doesn't work. */
+if (!Symbol.iterator) {
```
jayphelps: I believe `Symbol.iterator` and `Symbol` itself were both defined in ES2015 so if `Symbol.iterator` isn't available it's likely that `Symbol` isn't either and this my error. 🤔

src/internal/symbol/iterator.ts
```diff
@@ -1,38 +1,16 @@
...
+  console.warn('RxJS: Symbol.observable does not exist');
 }
 
-export const iterator = symbolIteratorPonyfill(root);
+/** The native Symbol.iterator instance or a string */
+export const iterator = Symbol.iterator || '@@iterator';
```
jayphelps: Same issue here `Uncaught ReferenceError: Symbol is not defined`

src/internal/symbol/iterator.ts
benlesh: Ah! Good catch!

src/internal/symbol/iterator.ts
benlesh: :+1:

src/internal/symbol/iterator.ts
```diff
@@ -1,38 +1,16 @@
...
-        }
-      }
-    }
-    return '@@iterator';
-  }
+/* NOTE: Warning users that they don't have a Symbol.iterator
+   polyfill. We don't want to throw on this, because it's not required
+   by the library. However it will provide clues to users on older
+   browsers why things like `from(iterable)` doesn't work. */
+if (!Symbol || !Symbol.iterator) {
```
jayphelps: @benlesh This still has the "Symbol is not defined" problem. Needed to use typeof checks. Here's a follow up PR to fix: #3394

PR #3390, opened 2018-03-07
chore(rxjs): add new rexport file that will be used to build the UMD …

benlesh:
This is an incremental step towards a new global UMD output, such that we'll be registering `rxjs` instead of `Rx` on the global to provide symmetry with the rest of the library's usage:

```js
/* For example, the following will exist on `window` or `global` */
rxjs.Observable
rxjs.fromEvent
rxjs.operators.map
rxjs.operators.mergeMap
rxjs.Subject

/* such that this will work... */
const { Observable, Subject, fromEvent } = rxjs;
const { map, mergeMap } = rxjs.operators;
```

This makes it congruent with our modules usage:

```ts
import { Observable, Subject, fromEvent } from 'rxjs';
import { map, mergeMap } from 'rxjs/operators';
```

Additionally, it will keep `rxjs` from trampling older versions of `Rx` if someone really wanted to have both globals. (for better or worse)

### DIFF
--- a/src/internal/umd.ts
+++ b/src/internal/umd.ts
@@ -1,0 +1,22 @@
+/*
+  NOTE: This is the global export file for rxjs v6 and higher.
+ */
+
+/* rxjs */
+export * from '../';
+
+/* rxjs/operators */
+import * as _operators from '../operators';
+export const operators = _operators;
+
+/* rxjs/testing */
+import * as _testing from '../testing';
+export const testing = _testing;
+
+/* rxjs/ajax */
+import * as _ajax from '../ajax';
+export const ajax = _ajax;
+
+/* rxjs/websocket */
+import * as _websocket from '../websocket';
+export const websocket = _websocket;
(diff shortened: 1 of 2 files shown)


benlesh:
cc @IgorMinar @jasonaden

benlesh:
Again: This is just an incremental step and does not change the actual UMD output yet. That's another ball of wax because so many of our tests are using the UMD base file, `Rx.ts`.

src/internal/umd.ts
```diff
@@ -0,0 +1,22 @@
+/*
+  NOTE: This is the global export file for rxjs v6 and higher.
+ */
+
+/* rxjs */
+export * from '../';
+
+/* rxjs/operators */
+import * as _operators from '../operators';
+export const operators = _operators;
```
IgorMinar: why not just `export * from '../operators';`? (also all the once below)

IgorMinar: otherwise this change LGTM. I think consistency is important, and making umd/globals usage as close to ESM usage will make documentation and learning of RxJS so much easier.

PR #3392, opened 2018-03-07
export ConnectableObservable interface

benlesh:
exports the ConnectableObservable type in such a way that TypeScript users can use it, but it's not the actual implementation as no one should be creating `new ConnectableObservable()`s.


benlesh:
closed in favor #3380

PR #3394, opened 2018-03-07
fix(Symbol.iterator): correctly handle case where Symbol constructor itself is not defined

jayphelps:
Follow up to #3389. This also makes the type a real `symbol` instead of `any` (return type of `getSymbolIterator` does it)

#3387 also needs to be adjusted as well.


src/internal/symbol/iterator.ts
```diff
@@ -1,14 +1,19 @@
+import { root } from '../util/root';
 
-/* NOTE: Warning users that they don't have a Symbol.iterator
-   polyfill. We don't want to throw on this, because it's not required
-   by the library. However it will provide clues to users on older
-   browsers why things like `from(iterable)` doesn't work. */
-if (!Symbol || !Symbol.iterator) {
-  console.warn('RxJS: Symbol.observable does not exist');
```
jayphelps: I missed this before 🙂

src/internal/symbol/iterator.ts
```diff
@@ -1,14 +1,19 @@
...
+    console.warn('RxJS: Symbol.iterator does not exist, so things like from(iterable) won\'t work');
+    return '@@iterator' as any;
+  }
+
+  return root.Symbol.iterator;
 }
 
-/** The native Symbol.iterator instance or a string */
-export const iterator = Symbol && Symbol.iterator || '@@iterator';
+export const iterator = getSymbolIterator(root);
```
benlesh: We actually want to get rid of `root` at some point. It's caused more issues than it's fixed.

jayphelps: do you have alternatives for testing?

PR #3405, opened 2018-03-07
Prettier

benlesh:
The first commit sets up Prettier to run on all `.ts` files under `./src` before any `git commit`. We can't run them on `./spec` because it would ruin the offsets of the marble tests. :(

The second one does nothing but run prettier on everything under `./src`, so it looks crazy, but it's just a reformat.


benlesh:
Probably don't want to merge this until things settle down.

benlesh:
Closing this because of formatting issues with Prettier and pipeable operators. See https://github.com/prettier/prettier/issues/4172

PR #3410, opened 2018-03-08
refactor(ignoreElements): ignoreElements now returns Observable<never>

### LINKED ISSUE #3340: ignoreElements operator should return Observable<never>
As once concluded in #2640, the `ignoreElements` operator should return a never observable.

Currently, `ignoreElements<T>` returns `Observable<T>`.

claudiordgz:
Per #2640 ignoreElements should resolve to Observable<never>, #3340 asks for this explicitly, there was also another issue with tests2png that I fix here (not sure where that came from)

Closes #3340

**Description:**

  - Makes ignoreElements return `Observable<never>` instead of whatever you feed into it (for example if you use `Observable<string>` it returns `Observable<string>`, fixing #3340 and #2640 
  - `tests2png` got broken again, fixed that too (see `spec/support/tests2png.opts`)

#3340 and #2640

### DIFF
--- a/src/internal/operators/ignoreElements.ts
+++ b/src/internal/operators/ignoreElements.ts
@@ -1,7 +1,8 @@
 import { Observable } from '../Observable';
 import { Operator } from '../Operator';
 import { Subscriber } from '../Subscriber';
-import { MonoTypeOperatorFunction } from '../types';
+import { OperatorFunction } from '../types';
+import { NEVER } from '../observable/never';
 
 /**
  * Ignores all items emitted by the source Observable and only passes calls of `complete` or `error`.
@@ -13,7 +14,7 @@ import { MonoTypeOperatorFunction } from '../types';
  * @method ignoreElements
  * @owner Observable
  */
-export function ignoreElements<T>(): MonoTypeOperatorFunction<T> {
+export function ignoreElements<T>(): OperatorFunction<T, Observable<never> > {
   return function ignoreElementsOperatorFunction(source: Observable<T>) {
     return source.lift(new IgnoreElementsOperator());
   };
@@ -28,10 +29,9 @@ class IgnoreElementsOperator<T, R> implements Operator<T, R> {
 /**
  * We need this JSDoc comment for affecting ESDoc.
  * @ignore
- * @extends {Ignored}
  */
 class IgnoreElementsSubscriber<T> extends Subscriber<T> {
-  protected _next(unused: T): void {
-    // Do nothing
+  protected _next(unused: T): Observable<never> {
+    return NEVER;
   }
 }
--- a/src/internal/operators/ignoreElements.ts
+++ b/src/internal/operators/ignoreElements.ts
@@ -2,7 +2,6 @@ import { Observable } from '../Observable';
 import { Operator } from '../Operator';
 import { Subscriber } from '../Subscriber';
 import { OperatorFunction } from '../types';
-import { NEVER } from '../observable/never';
 
 /**
  * Ignores all items emitted by the source Observable and only passes calls of `complete` or `error`.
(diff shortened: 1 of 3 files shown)


src/internal/operators/ignoreElements.ts
```diff
@@ -28,10 +29,9 @@ class IgnoreElementsOperator<T, R> implements Operator<T, R> {
 /**
  * We need this JSDoc comment for affecting ESDoc.
  * @ignore
- * @extends {Ignored}
  */
 class IgnoreElementsSubscriber<T> extends Subscriber<T> {
-  protected _next(unused: T): void {
-    // Do nothing
+  protected _next(unused: T): Observable<never> {
+    return NEVER;
```
benlesh: it should continue to do nothing. `_next` should never return a value.

src/internal/operators/ignoreElements.ts
```diff
@@ -13,7 +14,7 @@ import { MonoTypeOperatorFunction } from '../types';
  * @method ignoreElements
  * @owner Observable
  */
-export function ignoreElements<T>(): MonoTypeOperatorFunction<T> {
+export function ignoreElements<T>(): OperatorFunction<T, Observable<never> > {
```
benlesh: An `OperatorFunction<A, B>` is a function that takes an `Observable<A>` and returns an `Observable<B>` like so: `(source: Observable<A>) => Observable<B>`. What you've defined here is `(source: Observable<T>) => Observable<Observable<never>>`, which isn't quite what we want.

It doesn't really matter what type of values are coming out of the source observable, so that can be `Observable<any>`, and we're returning an `Observable<never>`, so the function we want is: `(source: Observable<any>) => Observable<never>`. Given that, we want to return `OperatorFunction<any, never>` 😄 

This should really be:

```ts
export function ignoreElements(): OperatorFunction<any, never> {
}
```

src/internal/patching/operator/ignoreElements.ts
```diff
@@ -11,6 +11,6 @@ import { ignoreElements as higherOrder } from '../../operators/ignoreElements';
  * @method ignoreElements
  * @owner Observable
  */
-export function ignoreElements<T>(this: Observable<T>): Observable<T> {
-  return higherOrder()(this) as Observable<T>;
+export function ignoreElements<T>(this: Observable<T>): Observable<never> {
```
benlesh: And here we should have:

```ts
export function ignoreElements(this: Observable<any>): Observable<never> {
}
```

src/internal/patching/operator/ignoreElements.ts
```diff
@@ -11,6 +11,6 @@ import { ignoreElements as higherOrder } from '../../operators/ignoreElements';
...
+  return higherOrder()(this) as Observable<never>;
```
benlesh: Check to make sure this cast is necessary after the other changes.

src/internal/operators/ignoreElements.ts
```diff
@@ -1,7 +1,8 @@
 import { Observable } from '../Observable';
 import { Operator } from '../Operator';
 import { Subscriber } from '../Subscriber';
-import { MonoTypeOperatorFunction } from '../types';
+import { OperatorFunction } from '../types';
+import { NEVER } from '../observable/never';
```
benlesh: This won't be needed

claudiordgz:
@benlesh Thank you so much for taking time to review the PR

src/internal/operators/ignoreElements.ts
claudiordgz: reverted

src/internal/operators/ignoreElements.ts
claudiordgz: I didn't notice it was resolving to `Observable<Observable<never>>` :flushed: 

I'm not sure how to test for proper types yet

src/internal/patching/operator/ignoreElements.ts
claudiordgz: thx!

src/internal/operators/ignoreElements.ts
claudiordgz: removed

PR #3416, opened 2018-03-08
build: use tsc importHelpers flag to remove duplicate helpers from esm5 and esm2015 builds

IgorMinar:
I intentionally didn't use the flag for cjs builds because that would
make cjs builds incompatible with SystemJS without further configuration
on the client-side - this would be highly undesirable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -166,6 +166,9 @@
     "url": "https://github.com/ReactiveX/RxJS/issues"
   },
   "homepage": "https://github.com/ReactiveX/RxJS",
+  "dependencies": {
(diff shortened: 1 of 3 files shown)


kwonoj:
what was conclusion for https://github.com/ReactiveX/rxjs/issues/2436#issuecomment-366882257 ? no issues expected? @benlesh

IgorMinar:
@kwonoj I explained it here: https://github.com/ReactiveX/rxjs/issues/2436#issuecomment-371585945

PR #3417, opened 2018-03-08
fix(fromEvent): Defines toString to fix Closure compilations

bowenni:
**Description:**
Compiling RxJS with Closure currently fails with `rxjs/src/internal/observable/fromEvent.closure.js:204: ERROR - variable toString is undeclared`. This fix is copied from [here](https://github.com/ReactiveX/rxjs/blob/1d100d3c34ebee5d2484a3b664a56f8600cc1f2c/src/observable/FromEventObservable.ts#L8).


benlesh:
Thanks @bowenni!

PR #3422, opened 2018-03-09
refactor(Observable): remove internal flag on source and operator

benlesh:
These are normal to access inside of the `lift` method of an extended Observable. They should not be used anywhere else, this also adds some documentation to that effect.

### DIFF
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -20,9 +20,16 @@ export class Observable<T> implements Subscribable<T> {
   /** @internal */
   public _isScalar: boolean = false;
 
-  /** @internal */
+  /**
+   * The source observable to get values from. This should only be used when
+   * overriding {@link lift}.
+   */
   protected source: Observable<any>;
-  /** @internal */
+
+  /**
+   * The operator definution to use as part of the operator chain. This should
+   * only be used when overriding {@link lift}.
+   */
   protected operator: Operator<any, T>;
 
   /**


src/internal/Observable.ts
```diff
@@ -20,9 +20,16 @@ export class Observable<T> implements Subscribable<T> {
...
-  /** @internal */
+  /**
+   * The source observable to get values from. This should only be used when
+   * overriding {@link lift}.
+   */
   protected source: Observable<any>;
-  /** @internal */
+
+  /**
+   * The operator definution to use as part of the operator chain. This should
```
IgorMinar: Typo

IgorMinar: Definition

IgorMinar:
this change is not sufficient. I still see the following compilation errors when building apps against a build from this branch at 2dfe07d4df7837c452fe00fe6b738a73ef418015:

```
/node_modules/rxjs/internal/types.d.ts (38,13): Property 'observable' does not exist on type 'SymbolConstructor'.
/node_modules/rxjs/internal/OuterSubscriber.d.ts (2,10): Module '"/node_modules/rxjs/internal/InnerSubscriber"' has no exported member 'InnerSubscriber'.
/node_modules/rxjs/internal/observable/combineLatest.d.ts (7,10): Module '"/node_modules/rxjs/internal/InnerSubscriber"' has no exported member 'InnerSubscriber'.
/node_modules/rxjs/internal/observable/race.d.ts (6,10): Module '"/node_modules/rxjs/internal/InnerSubscriber"' has no exported member 'InnerSubscriber'.
```

IgorMinar:
This PR is now obsolete since we merged #3418 instead

PR #3426, opened 2018-03-13
feat(umd): UMD now mirrors export schema for ESM and CJS

benlesh:
For example, if you want to do the same thing as  but using the global file, you would use , If you wanted to do , that will be available at . Thus making the import/access points more predictable between the UMD and the ESM and CJS versions.

### DIFF
--- a/.make-packages.js
+++ b/.make-packages.js
@@ -112,10 +112,10 @@ fs.copySync('src/testing/package.json', PKG_ROOT + '/testing/package.json');
 if (fs.existsSync(UMD_ROOT)) {
   fs.copySync(UMD_ROOT, UMD_PKG);
   // Add licenses to tops of bundles
-  addLicenseToFile('LICENSE.txt', UMD_PKG + 'Rx.js');
-  addLicenseTextToFile(license, UMD_PKG + 'Rx.min.js');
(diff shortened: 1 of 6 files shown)


.make-packages.js
```diff
@@ -112,10 +112,10 @@ fs.copySync('src/testing/package.json', PKG_ROOT + '/testing/package.json');
 if (fs.existsSync(UMD_ROOT)) {
   fs.copySync(UMD_ROOT, UMD_PKG);
   // Add licenses to tops of bundles
-  addLicenseToFile('LICENSE.txt', UMD_PKG + 'Rx.js');
-  addLicenseTextToFile(license, UMD_PKG + 'Rx.min.js');
-  addLicenseToFile('LICENSE.txt', UMD_PKG + 'Rx.js');
-  addLicenseTextToFile(license, UMD_PKG + 'Rx.min.js');
+  addLicenseToFile('LICENSE.txt', UMD_PKG + 'rx.js');
```
IgorMinar: I suggest that all these files are called `rxjs.*`, i.e. `rxjs.umd.js` and `rxjs.umd.min.js`

IgorMinar: this is just to be consistent about the naming across all the namespaces.

benlesh:
@kwonoj I"m just going to disable that part of the script for now.

PR #3427, opened 2018-03-13
fix: isObservable should return true for objects marked with '@@obser…

IgorMinar:
…vable'


src/internal/util/isObservable.ts
```diff
@@ -3,5 +3,5 @@ import { observable as Symbol_observable } from '../symbol/observable';
 
 /** Identifies an input as being Observable (but not necessary an Rx Observable) */
 export function isObservable(input: any): input is ObservableLike<any> {
```
benlesh: There should definitely be a function here. The contract is that `Symbol.observable` is a function that returns an object with a `subscribe(observer: Observer<T>): SubscriptionLike` on it. obj[Symbol.observable] should never equal a string.

IgorMinar:
discussed. the problem is real. the fix is wrong :)

PR #3436, opened 2018-03-15
fix(forkJoin): fix forkJoin typings for forkJoin(Observable<any>[])

benlesh:
Solves an issue where we were hitting the wrong typings for forkJoin if passed an Array of observables.

### DIFF
--- a/src/internal/observable/forkJoin.ts
+++ b/src/internal/observable/forkJoin.ts
@@ -8,13 +8,22 @@ import { InnerSubscriber } from '../InnerSubscriber';
 import { Subscriber } from '../Subscriber';
 
 /* tslint:disable:max-line-length */
+// forkJoin([a$, b$, c$]);
+export function forkJoin<T>(sources: [ObservableInput<T>]): Observable<T[]>;
+export function forkJoin<T, T2>(sources: [ObservableInput<T>, ObservableInput<T2>]): Observable<[T, T2]>;
+export function forkJoin<T, T2, T3>(sources: [ObservableInput<T>, ObservableInput<T2>, ObservableInput<T3>]): Observable<[T, T2, T3]>;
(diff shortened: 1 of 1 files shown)


src/internal/observable/forkJoin.ts
```diff
@@ -13,9 +13,9 @@ export function forkJoin<T, T2, T3>(v1: ObservableInput<T>, v2: ObservableInput<
 export function forkJoin<T, T2, T3, T4>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>): Observable<[T, T2, T3, T4]>;
 export function forkJoin<T, T2, T3, T4, T5>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>): Observable<[T, T2, T3, T4, T5]>;
 export function forkJoin<T, T2, T3, T4, T5, T6>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>): Observable<[T, T2, T3, T4, T5, T6]>;
-export function forkJoin<T>(v1: ObservableInput<T>): Observable<T[]>;
 export function forkJoin<T>(sources: Array<ObservableInput<T>>): Observable<T[]>;
 export function forkJoin<T>(...sources: Array<ObservableInput<T>>): Observable<T[]>;
+export function forkJoin<T>(v1: ObservableInput<T>): Observable<T[]>;
```
kwonoj: not exactly sure, but prolly this could be go line 10 instead?

benlesh: Yeah, I agree.

benlesh:
It now also updates typings to be a little nicer for people passing an array in.

PR #3437, opened 2018-03-15
fix(publishReplay): type inference improved

### LINKED ISSUE #3260: Pipe operator publishReplay cannot infer generic type.
**RxJS version:**
5.5.0

**Code to reproduce:**
```ts
declare const obs: Observable<number>;
const temp: null = obs.pipe(publishReplay(1));
```
**Expected behavior:**
I'm expecting `Type 'Observable<number>' is not assignable to type 'null'.`

**Actual behavior:**
`Type 'Observable<any>' is not assignable to type 'null'.`

benlesh:
closes #3260


ivan7237d:
@benlesh shouldn't `publishReplay` return a `ConnectableObservable`? In my code I call `publishReplay` then `connect`, and now I have to manually cast the retval of `publishReplay`...

PR #3439, opened 2018-03-16
feat(error-handling): add deprecated sync error handling behind a flag

benlesh:
Given the LARGE number of people that seem to be using bad patterns like wrapping subscribe calls in a try catch, we're going to readd the synchronous error throwing behind a config flag during transition.


src/internal/util/hostReportError.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Throws an error on another job so that it's picked up by the runtime's
+ * uncaught error handling mechanism.
+ * @param err the error to throw
+ */
```
kwonoj: should add `@internal` in doc comment

src/internal/config.ts
```diff
@@ -7,5 +7,15 @@ export const config = {
...
-  Promise
+  Promise,
+
+  /**
+   * If true, turns on synchronous error rethrowing, which is a deprecated behavior
+   * in v6 and higher. This behavior enables bad patterns like wrapping a subscribe
+   * call in a try/catch block. It also enables producer interference, a nasty bug
+   * where a multicast can be broken for all observers by a downstream consumer with
+   * an unhandled error. DO NOT USE THIS FLAG UNLESS IT'S NEEDED TO BY TIME
+   * FOR MIGRATION REASONS.
```
kwonoj: honestly I'd rather left this as undocumented, but guess that'll bring another confusion too. At least let's add `//TODO` note to move this into new `compat` pkg in further.

src/internal/Subscriber.ts
```diff
@@ -33,9 +35,14 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
                    error?: (e?: any) => void,
                    complete?: () => void): Subscriber<T> {
     const subscriber = new Subscriber(next, error, complete);
+    subscriber.syncErrorThrowable = false;
     return subscriber;
   }
 
+  public syncErrorValue: any = null;
```
kwonoj: I'd rather make these as `private` and let other internal just cast and refer it, at least to not to be exposed as TS interfaces for strict indication these are internals. (including below property as well)

kwonoj:
minor nit comments only.

src/internal/config.ts
```diff
@@ -17,5 +19,17 @@ export const config = {
    * an unhandled error. DO NOT USE THIS FLAG UNLESS IT'S NEEDED TO BY TIME
    * FOR MIGRATION REASONS.
    */
-  useDeprecatedSynchronousErrorHandling: false,
+  set useDeprecatedSynchronousErrorHandling(value: boolean) {
+    if (value) {
+      const error = new Error();
+      console.warn('DEPRECATED! RxJS was set to use deprecated synchronous error handling behavior by code at: \n' + error.stack);
+    } else if (_enable_super_gross_mode_that_will_cause_bad_things) {
+      console.log('RxJS: Back to a better error behavior. Thank you. <3');
```
kwonoj: I don't think this log message is necessary. let's leave warn only.

src/internal/Subscriber.ts
benlesh: We can't as there are other parts of the library that need to access it as a property, but I can flag it as `/**@internal*/`.

src/internal/config.ts
benlesh: I think the compat package might be able to just turn this on. cc @jasonaden

benlesh: then again... this changes behavior, where the compat stuff just has some performance and payload size changes associated with using it. So it's probably better to have people explicitly opt into this behavior.

src/internal/Subscriber.ts
benlesh: Done.

src/internal/config.ts
benlesh: I added this in part so we can see that it's removed in the tests (which turn it on then off in the beforeEach and afterEach)

src/internal/config.ts
kwonoj: What I meant was moving this config itself into compat pkg later, so anyone want to opt in need compat instead of core directly exposes this. I believe eventually core shouldn't contain this _flag_ at least.

src/internal/config.ts
kwonoj: Since test itself does verify behavior of config, is it really needed to observe console logs for it?

PR #3440, opened 2018-03-16
fix(config): make sure that Promise config is undefined initially

benlesh:
We need to make sure promise is undefined initially because systems like Zones need to have a chance to pick up Promise and patch it, and we're doing a runtime check for configuration anyhow. This makes the runtime check worthwhile and solves the problem with Zones (without needing to reference Zones in anyway)

### DIFF
--- a/spec/config-spec.ts
+++ b/spec/config-spec.ts
@@ -1,0 +1,9 @@
+import { config } from '../src/internal/config';
+import { expect } from 'chai';
+
+describe('config', () => {
+  it('should have a Promise property that defaults to nothing', () => {
+    expect(config).to.have.property('Promise');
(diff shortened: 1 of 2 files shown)


spec/config-spec.ts
```diff
@@ -0,0 +1,9 @@
+import { config } from '../src/internal/config';
+import { expect } from 'chai';
+
+describe('config', () => {
+  it('should have a Promise property that defaults to nothing', () => {
+    expect('Promise' in config).to.be.true;
+    expect(config.Promise).to.equal(undefined);
```
kwonoj:
```
expect(config).to.have.property('Promise');
expect(config.Promise).to.be.undefined;
```

PR #3451, opened 2018-03-20
V6 forward compatibility

jasonaden:
This PR adds forward compatibility from the stable branch to master's import locations and symbols.

Need a discussion on how to publish this once merged.


benlesh:
Well, I'm not sure about this 5.6 idea. I understand what you want to do, but this introduces a breaking change, and anyone with `rxjs@^5.5.0` or `rxjs@5.*` in their package.json will automatically get it, thus breaking them.

Maybe, if the desire is to be incremental, we need to have a v6 publish for this, then v7 to stick with semver.

benlesh:
Alternatively, Igor pointed out we could do a 5.6.0-forward-compat release, which would not be picked up by anyone's semver resolution and would allow people to opt-in to rxjs 6 import practices today without the other breaking changes, so long as they weren't using `* as Rx` style imports.

IgorMinar:
notes related to this change from:

Versioning:
- we can't release a regular 5.6.0 because it would break anyone using star imports (mostly node.js users), but at the same time, we can't version this as next major because existing libraries on npm specify rxjs ^5 or ^5.5 as their peerDependency. This is why we are releasing this as a pre-release version so that it is an explicit opt-in for users that want it
-when releasing this version to npm we should ensure that this version is not marked as `@latest` in npm, the publishing should be done without a label, or with an explicit label that is other than `@latest`, e.g. `@5-6-forward-compat`
- the initial release will be versioned as `5.6.0-forward-compat.0` (the "`-`" (dash) in the version number turns this version into a pre-release as far as semver is concerned)
- should we need to release an update to this forward-compat version, it should be versioned using .<number> suffix, e.g. `5.6.0-forward-compat.1`, `5.6.0-forward-compat.2`, etc
- libraries that want to be RxJS v6 compatible, should specify the following range in their peerDependencies declaration to enable the forward-compatibility solution to work without any dependency/versioning-related warnings or errors:
`peerDependencies: { "rxjs": "^5.6.0-forward-compat.0 || ^6.0.0-beta.0" }`

src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/* Constants */
+export { EMPTY } from './observable/empty';
+export { NEVER } from './observable/never';
+
+/* Types */
+export * from './types';
+
+/* Config */
+// export { config } from './config';
```
jayphelps: ?

benlesh: Config didn't exist in stable/5.5... that was new from 6.0.

PR #3453, opened 2018-03-20
Export the interface for Scheduler

bowenni:
Discussed offline:
- The Scheduler interface should be exported
- The class implementations of any Schedulers should not


bowenni:
This is not needed.
The `SchedulerLike` interface is exported via `export * from './internal/types';`

PR #3454, opened 2018-03-20
build: fix BUILD.bazel specification by excluding all backwards compat code

IgorMinar:
we have no target setup for this, we should set it up or be ok with not having backwards compat
story for external bazel users. I'll leave it up to @alexeagle and @jasonaden to decide and
potentially fix in a followup PR.


IgorMinar:
can you relint?

Issue #3444, opened 2018-03-16
repeatWhen operator broken since 5.5.6


wwjhu:
**RxJS version:**
5.5.6, 5.5.7

**Code to reproduce:**
https://jsfiddle.net/jy5fj6r8/16/

**Expected behavior:**
No errors, see
https://jsfiddle.net/cm6abypb/3/

**Actual behavior:**
Throws error `Rx.min.js:234 Uncaught TypeError: Cannot read property 'next' of null`

**Additional information:**
It appears that changes in 5.5.6 broke the repeatWhen operator. The fiddle mentioned as "code to reproduce" uses 5.5.7 while the one in "expected behavior" uses 5.5.5.

When the unit test is slightly modified (as shown below) by wrapping the code in a try..catch, the error is also visibile in the tests.

```
  it('should retry when notified and complete on returned completion', (done: MochaDone) => {
    try {
      const expected = [1, 2, 1, 2];
    Observable.of(1, 2)
      .map((n: number) => {
        return n;
      })
      .repeatWhen((notifications: any) => Observable.empty())
      .subscribe((n: number) => {
        expect(n).to.equal(expected.shift());
      }, (err: any) => {
        done(new Error('should not be called'));
      }, () => {
        done();
      });
    }
    catch (e) {
      done(new Error('should not throw'));
    }
  });
```

benlesh:
Thank you for reporting this, @wwjhu!

cartant:
I've had a quick peek at this. The `repeatWhen` implementation in 5.5.6 is unchanged from that in 5.5.5 - so 5.5.5 has the error, too.

However, prior with the change made in [this PR](https://github.com/ReactiveX/rxjs/pull/3161), the error was swallowed.

The problem is that `_unsubscribeAndRecycle` [sets](https://github.com/ReactiveX/rxjs/blob/5.5.5/src/operators/repeatWhen.ts#L103) `this.notifications` to `null`, but `this.notifications.next()` [is called](https://github.com/ReactiveX/rxjs/blob/5.5.5/src/operators/repeatWhen.ts#L83-L84) immediately after `_unsubscribeAndRecycle`.

kemsky:
Does this error exist in the latest RxJS (6.0.0-beta) version?

cartant:
> Does this error exist in the latest RxJS (6.0.0-beta) version?

Yes. And no. There is still an error in the beta, but - as in 5.5.5 - the error is swallowed.

[...]

PR #3460, opened 2018-03-22
 fix(repeatWhen): support synchronous notifier

cartant:
**Description:**

This PR is an attempt at fixing #3444. However, in making the changes, I found some conflict between a test, the documentation and the current behaviour. My changes were made in favour of the docs and the behaviour. In particular, [this test](https://github.com/ReactiveX/rxjs/blob/master/spec/operators/repeatWhen-spec.ts#L67-L81) attempts to assert that the observable repeats when an empty notifier is used:

```ts
it('should retry when notified and complete on returned completion', (done: MochaDone) => {
  const expected = [1, 2, 1, 2];
  Observable.of(1, 2)
    .map((n: number) => {
      return n;
    })
    .repeatWhen((notifications: any) => Observable.empty())
    .subscribe((n: number) => {
      expect(n).to.equal(expected.shift());
    }, (err: any) => {
    }, () => {
      done();
    });
});
```

However, it does not repeat. The test still passes, as there is no expectation to assert that all four values are received. [The docs](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-repeatWhen) have this to say:

> If the source Observable calls `complete`, this method will emit to the Observable returned from `notifier`. If that Observable calls `complete` or `error`, then this method will call `complete` or `error` on the child subscription.

I read that as meaning that an empty notifier that completes should not effect a repeat.

The change itself is trivial, it just checks to see whether or not the subscriber `isStopped` as it's possible for it to become so during the subscription to the notifier. See the explanation in [this comment](https://github.com/ReactiveX/rxjs/issues/3444#issuecomment-375127242).

Given the conflict between the docs - at least by my reading of them - and the tests, it will be interesting to hear opinions as to whether or not this is a breaking change. Fun stuff. At any rate, the changes I've made to the tests should be useful, even if the solution is not appropriate.

**Related issue (if exists):** #3444

### DIFF
--- a/spec/operators/repeatWhen-spec.ts
+++ b/spec/operators/repeatWhen-spec.ts
@@ -36,7 +36,7 @@ describe('Observable.prototype.repeatWhen', () => {
     expectSubscriptions(source.subscriptions).toBe(subs);
   });
 
-  it('should retry when notified via returned notifier on complete', (done: MochaDone) => {
+  it('should repeat when notified via returned notifier on complete', (done: MochaDone) => {
     let retried = false;
     const expected = [1, 2, 1, 2];
     let i = 0;
@@ -64,7 +64,7 @@ describe('Observable.prototype.repeatWhen', () => {
     }
   });
 
-  it('should retry when notified and complete on returned completion', (done: MochaDone) => {
+  it('should repeat when notified and complete on returned completion', (done: MochaDone) => {
     const expected = [1, 2, 1, 2];
     Observable.of(1, 2)
       .map((n: number) => {
--- a/spec/operators/repeatWhen-spec.ts
+++ b/spec/operators/repeatWhen-spec.ts
@@ -64,8 +64,9 @@ describe('Observable.prototype.repeatWhen', () => {
     }
   });
 
-  it('should repeat when notified and complete on returned completion', (done: MochaDone) => {
-    const expected = [1, 2, 1, 2];
+  it('should not repeat when applying an empty notifier', (done: MochaDone) => {
+    const expected = [1, 2];
+    const nexted: number[] = [];
     Observable.of(1, 2)
       .map((n: number) => {
         return n;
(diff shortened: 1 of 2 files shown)


benlesh:
The test is wrong. It should only repeat when next is notified, complete should cause completion.

benlesh:
It seems like there's a breaking change here, so we want to make sure that the breaking change is documented in the commit message.

cartant:
I don't know that there is a breaking change - or at least, I can't see it - but fighting with sinon in an attempt to stub/spy on `Subscriber.prototype.error` has done my head in.

I felt that it would be a breaking change if the behaviour were to be aligned with the test, but if the test is wrong, perhaps it's not a breaking change?

I've not been able to find a scenario where a `complete` from the notifier effects a repetition, but I've not looked too hard.

Also, there seems to be a problem - in 5.5.7, I expect the error will be swallowed in the beta - with notifiers that emit a value and then complete synchronously, like this:

```
repeatWhen(() => Rx.Observable.of(1))
```

It explodes on [this line](https://github.com/ReactiveX/rxjs/blob/6.0.0-beta.1/src/internal/operators/repeatWhen.ts#L79) as `retriesSubscription` is `null`.

cartant:
@benlesh Appologies for changing the PR after you'd approved it, but I had another look at it this morning and I think this fix is better.

The problem was with synchronous notifiers - empty or otherwise.

The PR now includes a failing test for a non-empty, synchronous notifier.

With an empty notifier that was empty, it would explode because `this.notifications` would be null` and with a non-empty notifier, it would explode because `this.retriesSubscription` would be `undefined` - as the completion notification, etc. would occur before `subscribeToResult` returned the subscription.

If the test I mentioned in the PR's details is wrong, I don't think there are breaking changes - the PR just fixes errors that would have been thrown in 5.5.6 and 5.5.7 and swallowed in other versions (including the beta).

benlesh:
Thank you, @cartant!

PR #3465, opened 2018-03-22
chore: add @nocollapse to static members.

josephperrott:
**Description:**

Add @nocollapse to prevent closure compiler from dropping static members that are reexported
twice.  The static members are kept in the first export, but if that symbol is then exported
again the static members are not kept.


benlesh:
Thanks @josephperrott!

Issue #3461, opened 2018-03-22
Errors swallowed in stopped subscribers


cartant:
**RxJS version:** 6.0.0-beta.1

**Code to reproduce:** See #3444

**Expected behavior:**

Errors should not be swallowed.

**Actual behavior:**

Errors are swallowed.

**Additional information:**

The [`Subscriber.prototype.error` implementation](https://github.com/ReactiveX/rxjs/blob/6.0.0-beta.1/src/internal/Subscriber.ts#L109-L114):

```ts
error(err?: any): void {
  if (!this.isStopped) {
    this.isStopped = true;
    this._error(err);
  }
}
```

swallows any errors that are passed to a stopped subscriber.

See #3444 for a repro. The internal error is swallowed because the subscriber to which it is passed is stopped at the time the error occurs.

benlesh:
This is by design. If a Subscriber has already been notified of completion or error, we do not want it to be notified again in a reentrant way.

cartant:
But should the error be completely ignored? At the moment, the implementation of `Subscriber.prototype.error` hides the implementation error in #3444 - calling `next` on a `null` property.

cartant:
The problem is in the implementation of [`_trySubscribe`](https://github.com/ReactiveX/rxjs/blob/6.0.0-beta.1/src/internal/Observable.ts#L210-L220):

```ts
protected _trySubscribe(sink: Subscriber<T>): TeardownLogic {
  try {
    return this._subscribe(sink);
  } catch (err) {
    if (config.useDeprecatedSynchronousErrorHandling) {
      sink.syncErrorThrown = true;
      sink.syncErrorValue = err;
    }
    sink.error(err);
  }
}
```

If there is a subscriber in the list of sinks - the subscribers chained together using the `destination` property - that happens to be closed/stopped, the error will be swallowed.

The only way that I can see that this could be fixed would be to walk the chain of subscribers and if a closed/stopped subscriber is found, rethrow the error or call `hostReportError` depending upon the `useDeprecatedSynchronousErrorHandling` configuration setting.

benlesh:
So to clarifiy, you're saying that in this case:

```ts
new Observable(subscriber => {
  subscriber.error(new Error('first one'));
  subscriber.error(new Error('second one'));
})
.subscribe({
  error(err) { console.log(`handled ${err.message}`); }
});
```

We should see the first one log as handled, and the second one should be rethrown (in another callstack)?

cartant:
[...]

PR #3468, opened 2018-03-24
fix(Subscriber): report or throw internal errors

cartant:
**Description:**

#3444 showed that there are situations in which `error` can be called after `unsubscribe` - see [this comment](https://github.com/ReactiveX/rxjs/issues/3444#issuecomment-375127242) - and in those situations the subscriber's `isStopped` property will be `true` and the error will be swallowed.

With this PR, if `error` is called when the subscriber is stopped, the error will either be reported to the host or rethrown - depending on the `useDeprecatedSynchronousErrorHandling` configuration setting.

**Related issue (if exists):** #3461

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -396,6 +396,28 @@ describe('Observable', () => {
       );
     });
 
+    it('should report internal, post-unsubscribe errors to the host', () => {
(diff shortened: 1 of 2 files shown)


cartant:
This doesn't solve the problem, as the calls to `error` aren't made through the `SafeSubscriber`.

PR #3469, opened 2018-03-24
fix(subscribe): prevent swallowing of errors when stopped

cartant:
**Description:**

If `_trySubscribe` catches an error and the sink - or one of its destinations - is stopped and the sink's `error` method is called, the error is swallowed - as it cannot be reported via the observer's callback. This commit adds an internal `reportError` method to `Subscriber`. The method defers to the appropriate error reporting mechanism if a stopped subscriber is found.

For an example scenario of an error being swallowed due to a bug and a closed subscriber (that has not previously been notified of any error) see #3444.

**Related issue (if exists):** #3461

### DIFF
--- a/src/internal/Subscriber.ts
+++ b/src/internal/Subscriber.ts
@@ -181,7 +181,10 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
   [Symbol_reportError](err?: any): void {
     let observer: PartialObserver<T> = this;
     while (observer) {
-      if (observer instanceof Subscriber) {
+      // `observer` will either be a subscriber or an empty observer, but for
+      // the reasons outlined in the constructor, `instanceof Subscriber`
+      // cannot be used.
+      if (isTrustedSubscriber(observer)) {
         const { destination, isStopped } = observer;
         if (isStopped) {
           if (config.useDeprecatedSynchronousErrorHandling) {
--- a/src/internal/Subscriber.ts
+++ b/src/internal/Subscriber.ts
@@ -118,39 +118,6 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
   }
 
   /**
-   * @ignore
-   * @internal
-   * Synchronous errors thrown within `subscribe` and caught by `_trySubscribe`
-   * can only be reported using a subscriber's `error` method if the subscriber
-   * is not stopped. This method traverses the `destination` chain of
-   * subscribers and uses the appropriate unhandled mechanism to report the
-   * error if a stopped subscriber is found.
-   */
-  reportError(err?: any): void {
-    let observer: PartialObserver<T> = this;
-    while (observer) {
-      if (observer instanceof Subscriber) {
-        const { destination } = observer;
-        if (observer.isStopped) {
-          if (config.useDeprecatedSynchronousErrorHandling) {
-            throw err;
-          } else {
-            hostReportError(err);
-            return;
-          }
-        } else if (destination === emptyObserver) {
-          observer.error(err);
-          return;
-        }
-        observer = destination;
-      } else {
-        observer.error(err);
-        return;
-      }
-    }
-  }
-
-  /**
    * The {@link Observer} callback to receive a valueless notification of type
    * `complete` from the Observable. Notifies the Observer that the Observable
    * has finished sending push-based notifications.
@@ -195,6 +162,36 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
     this._parent = _parent;
     this._parents = _parents;
     return this;
+  }
+
+  /**
+   * @ignore
+   * @internal
+   * Synchronous errors thrown within `subscribe` and caught by `_trySubscribe`
+   * can only be reported using a subscriber's `error` method if the subscriber
+   * is not stopped. This method traverses the `destination` chain of
+   * subscribers and uses the appropriate unhandled mechanism to report the
+   * error if a stopped subscriber is found.
+   */
+  private _reportError(err?: any): void {
+    let observer: PartialObserver<T> = this;
+    while (observer) {
+      if (observer instanceof Subscriber) {
+        const { destination, isStopped } = observer;
+        if (isStopped) {
+          if (config.useDeprecatedSynchronousErrorHandling) {
+            throw err;
+          } else {
+            hostReportError(err);
+            return;
+          }
+        }
+        observer = destination;
+      } else {
+        this.error(err);
+        return;
+      }
+    }
   }
 }
 
--- a/src/internal/Subscriber.ts
+++ b/src/internal/Subscriber.ts
@@ -6,6 +6,11 @@ import { rxSubscriber as rxSubscriberSymbol } from '../internal/symbol/rxSubscri
 import { config } from './config';
 import { hostReportError } from './util/hostReportError';
 
+export const Symbol_reportError =
+  (typeof Symbol === 'function' && typeof Symbol.for === 'function')
+    ? Symbol.for('reportError')
+    : '@@reportError';
+
 /**
  * Implements the {@link Observer} interface and extends the
  * {@link Subscription} class. While the {@link Observer} is the public API for
@@ -173,7 +178,7 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
    * subscribers and uses the appropriate unhandled mechanism to report the
    * error if a stopped subscriber is found.
    */
-  private _reportError(err?: any): void {
+  [Symbol_reportError](err?: any): void {
     let observer: PartialObserver<T> = this;
     while (observer) {
       if (observer instanceof Subscriber) {
--- a/src/internal/Subscriber.ts
+++ b/src/internal/Subscriber.ts
@@ -118,6 +118,39 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
   }
 
   /**
+   * @ignore
+   * @internal
+   * Synchronous errors thrown within `subscribe` and caught by `_trySubscribe`
+   * can only be reported using a subscriber's `error` method if the subscriber
+   * is not stopped. This method traverses the `destination` chain of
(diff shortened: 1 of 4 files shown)


src/internal/Subscriber.ts
```diff
@@ -113,6 +113,39 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
...
+  /**
+   * @ignore
+   * @internal
+   * Synchronous errors thrown within `subscribe` and caught by `_trySubscribe`
+   * can only be reported using a subscriber's `error` method if the subscriber
+   * is not stopped. This method traverses the `destination` chain of
+   * subscribers and uses the appropriate unhandled mechanism to report the
+   * error if a stopped subscriber is found.
+   */
+  reportError(err?: any): void {
```
benlesh: This should be private or protected. And ideally named with `_` prefix to discourage external use.

src/internal/Subscriber.ts
```diff
@@ -113,6 +113,39 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
...
+    let observer: PartialObserver<T> = this;
+    while (observer) {
+      if (observer instanceof Subscriber) {
```
benlesh: I'm not sure about this check. When would `this` not be the Subscriber? If for some reason it gets wrapped in a `SafeSubscriber`?

src/internal/Subscriber.ts
```diff
@@ -113,6 +113,39 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
...
+        const { destination } = observer;
+        if (observer.isStopped) {
```
benlesh: It seems like this could pass if someone just calls `error` twice in a row on Subscriber, no?

src/internal/Subscriber.ts
cartant: Done. It's now private and has a leading `_`. However, its being private means that a cast to `any` is required in `_trySubscribe` and in the tests.

src/internal/Subscriber.ts
cartant: `this` will always be a `Subscriber`. It's just that `observer` is the iteration variable. `PartialObserver<T>` is the type of the `destination` property - which is assigned to `observer` during the iteration.

The iteration starts with `this` so that its `isStopped` property is checked. And it's all done in the one loop so that the `hostReportError` call, etc. is just in the one place.

src/internal/Subscriber.ts
cartant: The existing mechanism for guarding against re-enterant calls to `error` should not be compromised. `_reportError` is called only from within the `catch` inside `_trySubscribe`.

Even if a caller subverts the `private` qualification (added per request above) and calls `_reportError` instead of `error`, the `Subscriber` will still become stopped and further calls to either `_reportError` or `error` will not see the observer's error callback invoked.

cartant: Wait. I'm wrong. The implementation needs to set `isStopped` to `true` for the subscriber at the root of the destination chain. I'll do that and add another test.

cartant: I've updated the implementation so that `error` is called on the chain's root subscriber. This will see calls to `error` flow through the chain of subscribers and each will be marked as stopped.

Even if the subscriber is used as a destination for a subscriber that itself is not part of the chain, the observer's error callback will be invoked only once. So the existing mechanism for guarding against re-enterant calls to `error` should not be compromised.

`_reportError` is called only from within the `catch` inside `_trySubscribe`, but even if a caller subverts the `private` qualification (added per request above) and calls `_reportError` instead of `error`, all of the subscribers in the chain will still become stopped and further calls to either `_reportError` or `error` will not see the observer's error callback invoked.

cartant:
@benlesh I'm finished making changes that relate to the review's comments.

cartant:
It occurred to me that `reportError` could be made less visible by using a `Symbol` and the method would then be callable without an `as any` assertion in `_trySubscribe` and in the tests.

I've updated the PR to do this.

benlesh:
Actually, the `instanceof Subscriber` check is going to be problematic. I just spent a while tracking down a nasty bug in node related to an instanceof Subscriber check (See #3477 and #3476). So I'm not quite ready to merge this one. I'll need to think about it what the implications are a little more carefully.

cartant:
Yes, I see what you mean. The implementation in this PR would have to use the `isTrustedSubscriber` function that you added in #3476.

cartant:
My take on it is that `instanceof Subscriber` will fail with foreign `Subscriber` instances and that will see `error` called on the root subscriber - even if the foreign `Subscriber` (or a subscriber further along the chain) is stopped. And that would see the error swallowed and not reported.

Technically, using `isTrustedSubscriber` as a custom type guard is overkill - as it's really only necessary to discern whether it's shaped like a `Subscriber` or just a `PartialObserver` - but it's probably better than introducing a seemingly redundant custom type guard. At least this way, it's apparent that `instanceof Subscriber` is dangerous.

src/internal/Subscriber.ts
```diff
@@ -163,6 +168,39 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
...
+          } else {
+            hostReportError(err);
+            return;
+          }
+        }
+        observer = destination;
+      } else {
+        this.error(err);
+        return;
+      }
```
benlesh: it looks like this loop will continue to hit hostReportError over and over for every level at which a subscriber `isStopped`. So if it has this.destination.destination.destination and all are stopped, the same error will be reported 3 or 4 times.

Also, I'm trying to understand the bug this is solving. The `isStopped` flag is meant to guard against code calling `subscriber.complete(); subscriber.error(new Error());` or calling `error` twice, etc.

It seems like the real fix might be to figure out why the Subscriber is prematurely stopped in the first place. Sorry, I should have caught that in an earlier review. Perhaps I missed something in one of your earlier comments?

cartant: > Sorry, I should have caught that in an earlier review.

No worries. With PRs like this - where there isn't a prior, agreed-upon solution that was discussed in an issue - I'm aware that it could all be for nought. It could be the wrong approach or could just be plain wrong, etc. I probably should have mentioned that at the outset, as it's never fun having to tell someone you don't want to merge a PR.

Anyway, this PR is far from urgent.

> Also, I'm trying to understand the bug this is solving.

The swallowing of the error is problem that it solves. If an error is thrown after a subscriber becomes stopped - as in #3444 - `_trySubscribe` attempts to report it via which `error` is pointless. The motivation for the PR was the appearance of the bug in #3444 in 5.5.6 and its disappearance in the beta. I thought that if a change to the sync error handling was made in 5.5.6 - to better report errors - it would be preferrable it the beta didn't go back to swallowing them.

I guess swallowed errors are a pet peeve of mine.

> It seems like the real fix might be to figure out why the Subscriber is prematurely stopped in the first place.

In #3444, the reason the subscriber became stopped was due to `notifyComplete` being called by the inner subscriber. The `notifyComplete` implementation called `super.complete()` and stopped the subscriber. However, with a synchronous notifier there was more code that ran after that and there was a bug in that code.

It might be possible to re-arrange `repeatWhen` so that becoming stopped is guaranteed to be the very last thing that occurs, but situations in which errors are swallowed could re-arise if someone were to write a wierd, user-land repeat-ish operator.

I imagine that the types of errors that this PR would expose - that would otherwise be swallowed - are very rare. I doubt there are too many operators like `repeatWhen` that have to fudge with the forwarding of `complete` notifications.

> it looks like this loop will continue to hit `hostReportError` over and over for every level at which a subscriber `isStopped`

I don't believe that's the case; the `return` will exit the function and, as it doesn't re-throw, there's no error to be caught by any outer `subscribe` calls. Basically, there are three ways out of the loop:

* the old, sync error behaviour of re-throwing the error;
* calling `hostReportError` and returning;
* calling `error` and returning.

Of those three, I'm quite sure that the last two are correct. However, I'm less certain about the first: the `throw`. I think it's okay, but I will need to give it some more thought.

benlesh:
I think the core team will have to have a discussion about what Subscribers should do when `error` is called twice. Historically, we've always ignored the second call. There are a variety of reasons for this.

Until then I think we can either block or close this PR, as I don't think I can land it in v6 and it would most definitely be a breaking change after that.

cartant:
> I think the core team will have to have a discussion ...

Yeah, in light of #3487 - which is almost the same, but with a completed subject instead of a completed/stopped subscriber - it would make a lot of sense to close this PR and institute an across-the-board policy for the reporting of errors if an observer's `error` method cannot be called.

If the deprecated sync error handling is going to be removed in v7, calling `hostReportError` in all situations in which an observer's `error` method cannot be called would make a lot of sense. And it would be simple. Unlike this PR.

benlesh:
#3487 was a little different, as it has to do with the specific implementation of bindCallback, but reporting errors to hostReportErrors in situations where error couldn't be called would definitely help there, yes.

cartant:
I'm really only talking about errors caught with a `try`/`catch` and then reported by calling `error` on an observer.

The suggestion is that a specific function could be called in those situations and said function could decide whether or not to report via the observer or the host.

Something like this:

```ts
function reportError(err: any, observer: ErrorObserver<any>): void {
  const { closed, destination, isStopped } = observer as any;
  if (closed || isStopped) {
    hostReportError(err);
  } else if (destination) {
    reportError(err, destination);
  } else {
    observer.error(err);
  }
}
```

It'd need better type checking, but that's the basic concept.

Changing the behaviour of `error` itself would be simpler still, but I'd have to go and re-read the observable contract and the TC39 proposal to see if that violates any guarantees.

slashhuang:
have the same thought like @cartant 
observer error is swallowed and cannot be caught.
{
   next: () => {
     // throw some error here
  },
  error: (err) => {
   // won't trigger,  but unsubscribe parent observables
 },
}

benlesh:
Okay, going to close this one for now, pending further discussion.

PR #3476, opened 2018-03-26
Multi package issue

benlesh:
This resolves a very nasty issue that happens when more than one copy of rxjs exists in node_modules. This is something that can organically happen to users if they have a dependency that hard-depends on one version of rxjs, but they try to use another version of rxjs for their app. Sometimes even if they revert back, they end up with, say, 2 or 3 copies of `rxjs@5.5.7` in their node_modules

The problem is that the `Subscriber` constructor was doing an `instanceof` check to see if the `destination` was a known/trusted RxJS subscriber. When that fails, it causes RxJS to treat the destination subscriber as any other POJO would be. Due to optimizations involving lift, InnerSubscriber, and a behavior where we're setting the context of the `next`, `error` and `complete` functions manually, then calling `unsubscribe` for them synchronously in SafeSubscriber, this was causing some asynchronous subscriptions to be unsubscribed prematurely in error.

There will be another patch for 5.5.x stable branch as well.

cc/ @hansl

### DIFF
--- a/node-tests/run.sh
+++ b/node-tests/run.sh
@@ -1,0 +1,16 @@
+
+rm -rf node-tests/node_modules/
+
+mkdir -p node-tests/node_modules/rxjs
+mkdir -p node-tests/node_modules/rxjs1
+
+cp -R dist/package/. node-tests/node_modules/rxjs
+cp -R dist/package/. node-tests/node_modules/rxjs1
+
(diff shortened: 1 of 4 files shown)


filipesilva:
Awesome work @benlesh and @jasonaden!

PR #3477, opened 2018-03-26
fix(node): multno more silent failure for many copies of rxjs

### LINKED ISSUE #3475: Observable does not complete using Sync/Async between two instances of RxJS
Smallest reproduction: https://stackblitz.com/edit/rxjs-3475-reproduction?file=index.ts

This is an issue we found in the Angular DevKit when updating projects and RxJS does not get properly hoisted (either because different versions or no peer deps).

Basically:

1. create an observable using Library 1
1. pipe a sync operator from Library 2
1. pipe an async operator from Library 1 (or 2)
1. the original SafeSubscriber is stopped, but did not complete or error, therefore the new pipeline will never next or complete the first pipeline.

benlesh:
related to #3476 

closes #3475

### DIFF
--- a/node-tests/run.sh
+++ b/node-tests/run.sh
@@ -1,0 +1,15 @@
+rm -rf node-tests/node_modules/
+
+mkdir -p node-tests/node_modules/rxjs
+mkdir -p node-tests/node_modules/rxjs1
+
+cp -R dist/package/. node-tests/node_modules/rxjs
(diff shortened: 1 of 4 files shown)


jasonaden:
Also, we need a version of this for the `forward-compat` branch as well (or just merge this to both `stable` and `forward-compat`).

benlesh:
The PR title was wrong, but I fixed the commit.

PR #3480, opened 2018-03-27
fix(bindCallback): add better type overloads

benlesh:
This is currently blocked, as it doesn't seem that TypeScript can really handle inferring types here like I hoped it could

### DIFF
--- a/src/internal/observable/bindCallback.ts
+++ b/src/internal/observable/bindCallback.ts
@@ -4,15 +4,47 @@ import { AsyncSubject } from '../AsyncSubject';
 import { Subscriber } from '../Subscriber';
 
 // tslint:disable:max-line-length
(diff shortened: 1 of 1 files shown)


cartant:
Have you tried re-ordering the signatures so that those with the most parameters precede those with fewer parameters? The most specific signatures should come first.

I saw the issue you created and it looked to me like your approach would be fine.

benlesh:
Blocked by https://github.com/Microsoft/TypeScript/issues/22952

cartant:
@benlesh I think you can disregard the amendment I made to the comment above. It most likely is only the ordering that's the problem.

I messed up my minimal repro and bound a generic function:

```ts
declare function func<T>(callback: (t: T) => void): void;
declare function bindCallback<T>(f: (callback: (t: T) => void) => void): T;
const result = bindCallback(func); // inferred as {}
```

I'm such a doofus. :grimacing:

PR #3481, opened 2018-03-27
build: Fix AMD module names generated by bazel

alexeagle:
Downstream users expect to require('rxjs') and require('rxjs/operators')

### DIFF
--- a/src/BUILD.bazel
+++ b/src/BUILD.bazel
@@ -5,15 +5,36 @@ exports_files(["tsconfig.json"])
 load("@build_bazel_rules_typescript//:defs.bzl", "ts_library")
 
 ts_library(
-  name = "rxjs",
-  module_name = "rxjs",
-  # exclude all backwards compatibility code because we don't have a bazel target setup for that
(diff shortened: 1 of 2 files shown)


alexeagle:
part of fixing https://github.com/angular/angular/issues/23005

benlesh:
It looks like this is failing on something new from TypeScript? Obviously doesn't have to do anything with the PR.

alexeagle:
@benlesh do I need to get a green build here to merge? Note, this is needed before we can cut Angular 6 rc.1

Maybe the TypeScript 2.8 release got picked up somehow, despite the `package-lock.json`?

kwonoj:
I have put tsc version to latest to catch new release of ts breaks current compilation while we staying master on alpha / beta stage, so explains why pkglock doesn't work.

I think build issue need to be fixed but probably in separate PR cause this PR didn't cause failure.

alexeagle:
thanks @kwonoj - could you merge this one pls? I need to get the fix downstream.

IgorMinar:
the CI is failing because of the unpinned typescript dependency that autoupgraded rxjs to typescript 2.8 which contains breaking changes.

alexeagle:
Merge Me @benlesh

PR #3485, opened 2018-03-29
build(typescript): lock TypeScript to 2.7.x to avoid unexpected break…

jasonaden:
…ing changes

TypeScript should always be locked down because they do not follow semver and make breaking changes in what appears to be feature releases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -249,7 +249,7 @@
     "ts-node": "4.1.0",
     "tslint": "5.9.1",
     "tslint-no-unused-expression-chai": "0.0.3",
-    "typescript": "latest",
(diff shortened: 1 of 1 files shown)


kwonoj:
I set master to latest to catch those breaking changes in typedef, to avoid previous issues when end user bumps up their compiler and rx start to breaking.

To properly handle this, I think we need to setup additional job on ci to compile against with latest compiler accepts to fail to catch those failure early at least.

benlesh:
The issue is fixed in #3489 I think we should keep latest TypeScript until we're out of beta at least.

PR #3503, opened 2018-03-30
fix(AjaxObservable): Use encodeURIComponent to encode ajax url params

### LINKED ISSUE #2389: Ajax requests do not handle ampersand correctly

ericponto:
**Description:**
The contentType 'application/x-www-form-urlencoded' was using encodeURI but should be using encodeURIComponent.

closes #2389


ericponto:
Looks like these changes were already merge in another PR.

PR #3504, opened 2018-03-30
feat(takeUntil): no longer subscribes to source if notifier synchrono…

benlesh:
…usly emits

closes #2189

cc @trxcllnt


benlesh:
Given that this is a copy of #2189, which is already reviewed, I'm going to merge it once it's green.

PR #3505, opened 2018-03-30
fix(throttle): now properly trailing throttles for individual values

benlesh:
BREAKING CHANGES: This changes the behavior of throttle, in particular
throttling with both leading and trailing behaviors set to true, to more
closely match the throttling behavior of lodash and other libraries.
Throttling now starts immediately after any emission from the
observable, and values will not be double emitted for both leading and
trailing values

closes #2864


benlesh:
Given that this is a copy of #2864, and that work was approved, I'll merge this once it's green.

PR #3509, opened 2018-03-31
fix(smoosh): merge et al are now smoosh (smooshMap, smooshAll, etc)

jayphelps:
Due to the TC39 proposal for Array flatten/flatMap likely being changed to smoosh/smooshMap it was also decided that, to have name parity, Observables should use those names as well.

In the case of RxJS, there are a few more operators with merge in their name that this PR fixes:

* `merge` → `smoosh`
* `mergeMap` → `smooshMap`
* `mergeMapTo` → `smooshMapTo`
* `mergeAll` → `smooshAll`
* `mergeScan` → `smooshScan`

Here's some background on smoosh and the issues TC39 has had with the name "flatten":

https://github.com/tc39/proposal-flatMap/pull/56
http://logs.libuv.org/tc39/2018-03-07#01:04:06.355
https://github.com/staltz/prevent-smoosh

This is unfortunately a breaking change, but one that I think **everyone agrees with** so it shouldn't be a big deal.

:shipit:


compat/operator/concatMapTo.ts
```diff
@@ -38,12 +38,12 @@ import { concatMapTo as higherOrder } from 'rxjs/operators';
...
  * @see {@link concatAll}
  * @see {@link concatMap}
- * @see {@link mergeMapTo}
+ * @see {@link smooshMapTo}
  * @see {@link switchMapTo}
  *
  * @param {ObservableInput} innerObservable An Observable to replace each value from
  * the source Observable.
- * @return {Observable} An observable of values merged together by joining the
+ * @return {Observable} An observable of values smooshd together by joining the
```
jayphelps: looks like the find-and-replace didn't work right for past tense. Will fix.

jayphelps:
@rgbkrk fixed! (assuming all the build variants pass this time)

cartant:
These changes should be applied to the `stable` branch, too.

lrowe:
LGTM

PR #3512, opened 2018-04-02
fix skipUntil

benlesh:
Fixes skipUntil behavior to be correct.

- No longer waits for notifier to complete before completing
- No longer retains subscription to notifier after it notifies

### DIFF
--- a/spec/operators/skipUntil-spec.ts
+++ b/spec/operators/skipUntil-spec.ts
@@ -1,56 +1,55 @@
-import * as Rx from 'rxjs/Rx';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
+import { Observable, of, Subject } from 'rxjs';
+import { skipUntil, mergeMap } from 'rxjs/operators';
 
 declare function asDiagram(arg: string): Function;
 
-const Observable = Rx.Observable;
-
 /** @test {skipUntil} */
-describe('Observable.prototype.skipUntil', () => {
+describe('skipUntil', () => {
   asDiagram('skipUntil')('should skip values until another observable notifies', () => {
     const e1 =     hot('--a--b--c--d--e----|');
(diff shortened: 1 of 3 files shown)


src/internal/operators/skipUntil.ts
```diff
@@ -26,8 +27,12 @@ class SkipUntilOperator<T> implements Operator<T, T> {
   constructor(private notifier: Observable<any>) {
   }
 
-  call(subscriber: Subscriber<T>, source: any): TeardownLogic {
-    return source.subscribe(new SkipUntilSubscriber(subscriber, this.notifier));
+  call(destination: Subscriber<T>, source: any): TeardownLogic {
+    const { notifier } = this;
+    const innerSubscription = new Subscription();
```
cartant: This looks a little strange, to me. Is there a particular reason the `innerSubscription` cannot be created in the `SkipUntilSubscriber` constructor?

spec/operators/skipUntil-spec.ts
```diff
@@ -218,12 +206,12 @@ describe('Observable.prototype.skipUntil', () => {
     expectSubscriptions(skip.subscriptions).toBe(skipSubs);
   });
 
-  it('should skip all element when another observable unsubscribed early before emit', () => {
+  it('should skip skip all elements if notifier is unsubscribed explicitly before the notifier emits', () => {
```
cartant: Is this test really required? To me, it looks like it only tests that calling [`unsubscribe`](https://github.com/ReactiveX/rxjs/blob/6.0.0-rc.0/src/internal/Subject.ts#L97-L101) on the `Subject` sets its `closed` property to `true`.

benlesh: Yeah, this is one weird case where semantics are a little different. In almost every other case, the notifier can only complete, error or last forever. This is a case where it can stop notifying and not still be active.

PR #3523, opened 2018-04-03
fix(sourcemaps): fix mappings for source maps so they will work

### LINKED ISSUE #3479: Sourcemaps in v6
**RxJS version:** 6.0.0-beta.3

**Code to reproduce:** any webpack bundle

**Expected behavior:** no warnings

**Actual behavior:**
multiple warnings such as
(Emitted value instead of an instance of Error) Cannot find source file '../../../../../src/internal/observable/dom/WebSocketSubject.ts': Error: Can't resolve '../../../../../src/internal/observable/dom/WebSocketSubject.ts'

**Additional information:**
I think need to enable `inlineSources` in tsconfig.json

jasonaden:
This PR basically just adjusts the `sources` property of the `.js.map` files to point to the right places.

Fixes #3479

### DIFF
--- a/.make-compat-package.js
+++ b/.make-compat-package.js
@@ -6,6 +6,11 @@ let mkdirp = require('mkdirp');
 let path = require('path');
 let klawSync = require('klaw-sync');
 let licenseTool = require('./tools/add-license-to-file');
+let makePackages = require('./.make-helpers');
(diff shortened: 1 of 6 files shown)


benlesh:
Travis is sad. Looks like something is executing out of order and some files don't exist when the build gets to a particular step.

PR #3530, opened 2018-04-06
fix: fix type mismatch in `NodeStyleEventEmitter`

joemphilips:
**Description:**
There was a type mismatch in `observable/fromEvent.ts` which causes `tsc` compile error when I tried to compile the following.
```
Observable.fromEvent(new EventEmitter(), "foo")
```
Error I've got is below
```
src/my-library/index.ts:41:5 - error TS2352: Type 'EventEmitter' cannot be converted to type 'NodeStyleEventEmitter'.
  Types of property 'addListener' are incompatible.
    Type '(event: string | symbol, listener: (...args: any[]) => void) => EventEmitter' is not comparable to type '(eventName: string, handler: Function) => void'.
      Types of parameters 'listener' and 'handler' are incompatible.
        Type 'Function' is not comparable to type '(...args: any[]) => void'.
          Type 'Function' provides no match for the signature '(...args: any[]): void'.
```

NOTE: There was also a same problem in stable branch. Please tell me what should I do for it.
https://github.com/ReactiveX/rxjs/blob/stable/src/observable/FromEventObservable.ts

### DIFF
--- a/src/internal/observable/fromEvent.ts
+++ b/src/internal/observable/fromEvent.ts
@@ -6,10 +6,12 @@ import { map } from '../operators/map';
 
 const toString: Function = Object.prototype.toString;
 
-export type NodeStyleEventEmitter = {
-  addListener: (eventName: string, handler: Function) => void;
-  removeListener: (eventName: string, handler: Function) => void;
-};
+export interface NodeStyleEventEmitter {
+  addListener: (eventName: string | symbol, handler: NodeEventHandler) => this;
+  removeListener: (eventName: string | symbol, handler: NodeEventHandler) => this;
+}
+
+export type NodeEventHandler = (...args: any[]) => void;
 
 export type JQueryStyleEventEmitter = {
   on: (eventName: string, handler: Function) => void;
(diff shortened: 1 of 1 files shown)


benlesh:
Thank you, @joemphilips! ☺️

cartant:
@joemphilips Regarding your question about the `stable` branch. Now that your PR has been merged, you could create another branch (named `emitter-fix-stable` or whatever)  based on the `stable` branch:

```
git checkout -b emitter-fix-stable stable
```

Often, you can use `git cherry-pick <commit-sha>` to apply a commit made to `master` to your new branch. However, the file structures have changed radically between `stable` and `master` and Git appears to be unable to automatically apply this particular commit.

So you should make the required changes to [`FromEventObservable.ts`](https://github.com/ReactiveX/rxjs/blob/stable/src/observable/FromEventObservable.ts) in your new branch and then create another PR from it that targets the `stable` branch instead of `master` (the GitHub New-Pull-Request page lets you specify the base branch). Then your new PR can be merged with `stable`.

joemphilips:
@cartant Thanks a lot! I made a separate PR here #3536

PR #3532, opened 2018-04-06
fix(compat): export TeardownLogic

### LINKED ISSUE #3531: TeardownLogic not exported when using rxjs-compat
**RxJS version:** `rxjs@6.0.0-rc.0` and `rxjs-compat@6.0.0-rc.0`

**Code to reproduce:**

```ts
import { Subscription, TeardownLogic } from "rxjs/Subscription";
```

**Expected behavior:**

`TeardownLogic` should be importable as in 5.5.8.

**Actual behavior:**

An error is effected:

```
[ts] Module '".../rxjs-tslint-rules/fixtures/v6-compat/node_modules/rxjs/Subscription"' has no exported member 'TeardownLogic'.
```
**Additional information:**

I discovered this when adding support for v6 and v6+compat to `rxjs-tslint-rules`. It's no big deal for me, but I thought I'd point it out.

FYI the 5.5.8 export is [here](https://github.com/ReactiveX/rxjs/blob/5.5.8/src/Subscription.ts#L12).

cartant:
Closes #3531

**Description:**

Adds a `TeardownLogic` export to `compat/Subscription` for 5.5.8 compatibility - see [this export](https://github.com/ReactiveX/rxjs/blob/5.5.8/src/Subscription.ts#L12) in 5.5.8.

**Related issue (if exists):** #3531

### DIFF
--- a/compat/Subscription.ts
+++ b/compat/Subscription.ts
@@ -1,1 +1,1 @@
-export {Subscription, Unsubscribable as AnonymousSubscription, SubscriptionLike as ISubscription} from 'rxjs';
+export {Subscription, Unsubscribable as AnonymousSubscription, SubscriptionLike as ISubscription, TeardownLogic} from 'rxjs';


benlesh:
Good catch!

Issue #3525, opened 2018-04-04
[rxjs v6] fromPromise not exported from index.d.ts


mkurcius:
**RxJS version:**
6.0.0-rc.0

**Code to reproduce:**
`import { fromPromise } from 'rxjs';`

**Expected behavior:**
exported `fromPromise`

**Actual behavior:**
`Module '".../node_modules/rxjs/index"' has no exported member 'fromPromise'`

mdbootstrap:
RxJS:
6.0.0-rc.0

I have a similar problems with:
```
node_modules/rxjs-compat/add/operator/catch.d.ts(1,10): error TS2305: Module '"node_modules/rxjs/internal-compatibility/index"' has no exported member '_catch'.
node_modules/rxjs-compat/add/operator/combineLatest.d.ts(1,10): error TS2305: Module '"node_modules/rxjs/internal-compatibility/index"' has noexported member 'combineLatest'.
node_modules/rxjs-compat/add/operator/filter.d.ts(1,10): error TS2305: Module '"node_modules/rxjs/internal-compatibility/index"' has no exported member 'filter'.
node_modules/rxjs-compat/add/operator/map.d.ts(1,10): error TS2305: Module '"/node_modules/rxjs/internal-compatibility/index"' has no exported member 'map'.
node_modules/rxjs-compat/add/operator/mergeAll.d.ts(1,10): error TS2305: Module '"node_modules/rxjs/internal-compatibility/index"' has no exported member 'mergeAll'.
```
I'm able to fix everything from this by importing it from 'rxjs/internal/operators/', but for _catch there's no such a file. 

Any suggestion / fix / help / workaround would be very pleasant.

Bloodcast69:
RxJS:
6.0.0-rc.0

I have a similar problems with:
```
node_modules/rxjs-compat/add/operator/catch.d.ts(1,10): error TS2305: Module '"node_modules/rxjs/internal-compatibility/index"' has no exported member '_catch'.
node_modules/rxjs-compat/add/operator/combineLatest.d.ts(1,10): error TS2305: Module '"node_modules/rxjs/internal-compatibility/index"' has noexported member 'combineLatest'.
node_modules/rxjs-compat/add/operator/filter.d.ts(1,10): error TS2305: Module '"node_modules/rxjs/internal-compatibility/index"' has no exported member 'filter'.
[...]

PR #3535, opened 2018-04-08
Added fromPromise to index.js addresses #3525

SanderElias:
**Description:**
This PR adds the missing fromPromise into the main barrel.

addresses #3525


kwonoj:
Please refer https://github.com/ReactiveX/rxjs/pull/3391 , taking out top level export was intended. /cc @benlesh

SanderElias:
@kwonoj Thanks for the helpful link! I was not aware of that. I get the reasoning. However, this is a breaking change, and not mentioned n the change-log. 
So one of two things should be done.

1. put it in the changelog as a breaking change
2, put it back in as per this PR.

My vote would be on 1.

PR #3536, opened 2018-04-09
fix: fix a typing mismatch in FromEventObservable

joemphilips:
**Description:**
see #3530 

changes made here is exactly the same with #3530 except the file path.

### DIFF
--- a/src/observable/FromEventObservable.ts
+++ b/src/observable/FromEventObservable.ts
@@ -8,9 +8,12 @@ import { Subscriber } from '../Subscriber';
 const toString: Function = Object.prototype.toString;
 
 export type NodeStyleEventEmitter = {
-  addListener: (eventName: string, handler: Function) => void;
-  removeListener: (eventName: string, handler: Function) => void;
+  addListener: (eventName: string, handler: NodeEventHandler) => void;
+  removeListener: (eventName: string, handler: NodeEventHandler) => void;
 };
+
+export type NodeEventHandler = (...args: any[]) => void;
+
 function isNodeStyleEventEmitter(sourceObj: any): sourceObj is NodeStyleEventEmitter {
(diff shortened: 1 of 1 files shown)


benlesh:
I'm not sure about this one, as `stable` targets TS 2.0, and we don't want to break people. I'll need to confirm this change before I can merge. :\

cartant:
@benlesh To me, it looks like this should be fine, as this code is fine for TypeScript 2.0.10:

```ts
type EmitterTypeWithFunction = {
  addListener: (eventName: string, handler: Function) => void;
  removeListener: (eventName: string, handler: Function) => void;
}

type HandlerWithArgs = (...args: any[]) => void;
type EmitterTypeWithArgs = {
  addListener: (eventName: string, handler: HandlerWithArgs) => void;
  removeListener: (eventName: string, handler: HandlerWithArgs) => void;
}

const emitterWithFunction = {
  addListener(eventName: string, handler: Function) {},
  removeListener(eventName: string, handler: Function) {}
}

const emitterWithArgs = {
  addListener(eventName: string, handler: (...args: any[]) => void) {},
  removeListener(eventName: string, handler: (...args: any[]) => void) {}
}

const e1: EmitterTypeWithFunction = emitterWithFunction;
const e2: EmitterTypeWithArgs = emitterWithFunction;
const e3: EmitterTypeWithFunction = emitterWithArgs;
const e4: EmitterTypeWithArgs = emitterWithArgs;
```

`Function` will match `(...args: any[]) => void` and vice versa.

TypeScript 2.6 introduced the `strictFunctionTypes` compiler option. If that option - or the `strict` option - is set to `true`, the assignment to `e3` will effect this error:

```
Type 'Function' provides no match for the signature '(...args: any[]): void'.
```

So this PR fixes a problem for users of the stable release that are themselves using TypeScript 2.6+ with either `strict` or `strictFunctionTypes` enabled.

benlesh:
Thanks, @joemphilips! And thanks for the additional review @cartant!

PR #3538, opened 2018-04-09
feat(schematics): add migration schematics for schematics users

benlesh:


spec/migrations/update-6_0_0/index.ts
```diff
@@ -0,0 +1,87 @@
...
+
+const rxjsCompatVersion = '^6.0.0-rc.0';
+const rxjsTSLintVersion = '0.0.0';
+
+function addDependencies() {
+  return (tree: Tree, context: SchematicContext) => {
+      const pkgPath = '/package.json';
+      const buffer = tree.read(pkgPath);
+      if (buffer == null) {
+        throw new SchematicsException('Could not read package.json');
```
hansl: This is a runtime dependency.  Just throwing `Error` is fine though.

spec/migrations/update-6_0_0/index.ts
```diff
@@ -0,0 +1,87 @@
...
+      tree.overwrite(pkgPath, JSON.stringify(pkg, null, 2));
+      context.addTask(new NodePackageInstallTask());
+
+      return tree;
+  };
+}
+
+function runTsLint() {
+  return (tree: Tree, context: SchematicContext) => {
+    const workspace = getWorkspace(tree);
```
hansl: This limits the update schematic to angular CLI projects. I'm pretty sure you don't need this restriction; you should be able to run tslint over the root tsconfig (or all TS files). We have a Tslint task now.

spec/migrations/update-6_0_0/latestVersions.ts
```diff
@@ -0,0 +1,5 @@
+export const latestVersions = {
```
Brocco: This file does not seem to be used anywhere

spec/migrations/update-6_0_0/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      if (!pkg.dependencies) {
+        pkg.dependencies = {};
+      }
+      if (!pkg.devDependencies) {
+        pkg.devDependencies = {};
+      }
+
+      pkg.dependencies['rxjs-compat'] = rxjsCompatVersion;
+      pkg.devDependencies['rxjs-tslint'] = rxjsTSLintVersion;
```
IgorMinar: Can we skip this install for now? Since we 3ant people to primarily update their app to v6 we shouldn't bring deps they absolutely don't need. Because if we do they will not know when it's ok remove them.

IgorMinar:
Thinking about this more, I believe that the best solution would be to move the contents of rxjs-lint into this package.

benlesh:
@IgorMinar my only concern is that publishing the packages in this repository is already pretty complicated, and worse, it's still manual. I'm afraid I'm too dumb to complicate it anymore than it already is.

If publishing were more automated, then I'd completely agree. Right now I only mostly agree. lol

mgechev:
I can see all the pros of moving the migration rules to the `rxjs` package and I agree it sounds ideal. If this is going to complicate the publishing process, I think there's one more option which keeps the changes here to the minimum.

We can keep all TSLint migration and linting rules for RxJS separately (as `rxjs-tslint` is right now), and introduce a TSLint configuration preset for migration from version 5 to 6 as part of the rxjs package. This will look something like:

```
{
  "rules": {
    "update-rxjs-imports": true,
    "migrate-to-pipeable-operators": true,
    "collapse-rxjs-imports": true
  }
}
```
The pros of this solution that I see are:
- We keep the changes to the rxjs package & repository to the minimum
- RxJS 6 knows about the migration process from RxJS 5 (the tslint rule preset)

As for cons:
- The maintenance of two packages is still necessary

hansl:
How about make the linting rules part of the compat package?

IgorMinar:
I like it! That way once you are done updating you remove both the
rxjs-compat package and the tslint transforms.

PR #3539, opened 2018-04-09
Update using() to apply to any disposable resource type

darthtrevino:
**Description:**
The documentation of the _using_ function on reactivex.io indicates that it can be used to dispose of arbitrary resources. (http://reactivex.io/documentation/operators/using.html). The RxJs section has an image indicating that the client should be able to pass in a _disposeFunction_. The provided RxJs in the docs does not work for a couple of reasons:
   * The resource allocated by _using_ is only ever disposed of using _unsubscribe()_, which implies something about its type.
   * The resultant subscription does not have a dispose() method, but the implementation here looks like it provides that capability.

### DIFF
--- a/compat/observable/UsingObservable.ts
+++ b/compat/observable/UsingObservable.ts
@@ -1,8 +1,11 @@
-import { Observable, Unsubscribable, SubscribableOrPromise, using } from 'rxjs';
+import { Observable, ObservableInput, Unsubscribable, SubscribableOrPromise, using } from 'rxjs';
 
 export class UsingObservable<T> extends Observable<T> {
-  static create<T>(resourceFactory: () => Unsubscribable | void,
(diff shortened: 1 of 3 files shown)


benlesh:
@darthtrevino The `using` in the definition at that link is actually from RxJS v4 and under. Those docs are horrifically out of date.

I see the spirit of what you're doing here, but I'm not sure if we want to change the semantics of `using` just yet. That said, I personally rarely use `using` in my apps.

@mattpodwysocki, what do you think?

FWIW: I'm going to leave this PR open for discussion.

darthtrevino:
Sounds good. I was putting together a brown-bag for my team on RxJs, and as part of it I wanted to cover some more esoteric scenarios involving resource management with using and ref-counting.

darthtrevino:
So one option may be to make the _disposeFunction_ an optional argument, and if it's undefined then use the original logic of invoking `resource.unsubscribe()`

PR #3540, opened 2018-04-09
chore(TypeScript): locked to 2.8.*

benlesh:
Now that we're nearing release, locking TypeScript down to 2.8.x to prevent further breakage.

(For those of you who don't know, TypeScript does _not_ adhere to semver, and publishes breaking changes in minor releases)


kwonoj:
What cases do we want allow via `~` version cases vs. explicit pinned version? Technically TS can break any version up.

PR #3544, opened 2018-04-10
fix(node/TS): eliminate incompatable types to protected properties

benlesh:
Resolves an issue in Node and TypeScript where if a project had more than one copy of RxJS, even if it was the same version, types would be incompatable between say, Observable and Observable, due to protected fields in the dts files. As all of these fields are really implementation details of the library, and my be removed at any time do to architectural changes, it is prudent to mark them as deprecated. Marking them as public only serves to eliminate the type incompatibilities.


ORESoftware:
Yep hit this one today:
https://stackoverflow.com/questions/50109847/strange-transpile-error-with-rxjs5-subject-subscription

Issue #3550, opened 2018-04-11
Migration collection not in 6.0.0-terrific-rc.3 release


cexbrayat:
**RxJS version:**

6.0.0-terrific-rc.3

**Code to reproduce:**

Using CLI 6.0.0-rc.3
Bump RxJS to 6.0.0-terrific-rc.3
Then run `ng update rxjs --migrate-only --from=5.5.8`

**Expected behavior:**

It should run the migration schematic

**Actual behavior:**

It throws `Collection "rxjs/./migrations/collection.json" cannot be resolved.`

**Additional information:**

It's entirely possible that I am missing something, or that this is a CLI issue, but it looks like `node_modules/rxjs` does not contain the `migrations` folder. I know the support of migration is very experimental, but maybe there is a missing piece in the RxJS release to integrate the `migrations` folder...

cexbrayat:
I confirm that the migration script is now included

But, FYI, using CLI 6.0.0-rc.4, I'm now running into: `Cannot find module '/Users/cedric/Code/ninjasquad/ng2propack/frontend/node_modules/rxjs/migrations/update-6_0_0/index'`. 

Which is maybe more a CLI issue? Because the `update-6_0_0` folder does contain `index.ts`...

cexbrayat:
@benlesh 👍 Fixed in rc.7.

PR #3555, opened 2018-04-12
fix(migrations): build now properly copies migration into package

benlesh:
We (*I*) forgot to copy the migrations directory to the dist/package prior to publish.


benlesh:
addresses #3550

PR #3559, opened 2018-04-12
fix(throttleTime): emit throttled values when complete if trailing=true

### LINKED ISSUE #3351: throttleTime does not emit last value
`throttleTime` does not have a `_complete` handler which means that even if we specify `trailing: true` we might not get the last value of the stream.

Not sure if this is intentional, but it seems rather unintuitive. Also looking at the spec test the end of stream scenario isn't really defined so it's difficult to tell how it should work.

yozo1984:
**Description:**
BREAKING CHANGE: This changes the behavior of throttleTime, in particular when throttling with trailing set to true. Throttling now treats observable complete as a trailing edge in addition to the timer and emits the last throttled value when observable completes before the current timer is up.

closes #3351


benlesh:
Thank you @yozo1984

PR #3562, opened 2018-04-13
fix(isArrayLike): reject functions because functions have "length"

johnlindquist:
**Description:**
Added a `typeof x !== "function"` check to `isArrayLike`. `isArrayLike` was only checking for `length` and functions have a `length` property. This prevented functions from properly falling through to this case:

https://github.com/ReactiveX/rxjs/blob/6fd1fd7719e690d93ba3205db8b6b6e9e0990579/src/internal/util/subscribeTo.ts#L31

### DIFF
--- a/spec/observables/from-spec.ts
+++ b/spec/observables/from-spec.ts
@@ -1,6 +1,7 @@
 import { expect } from 'chai';
 import { TestScheduler } from 'rxjs/testing';
-import { asyncScheduler, of, from, Observable, asapScheduler, Observer } from 'rxjs';
+import { asyncScheduler, of, from, Observable, asapScheduler, Observer, observable, Subject, ObservableInput } from 'rxjs';
+import { first } from 'rxjs/operators';
 
 // tslint:disable:no-any
 declare const asDiagram: any;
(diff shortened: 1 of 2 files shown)


spec/observables/from-spec.ts
```diff
@@ -16,13 +17,13 @@ function getArguments<T>(...args: T[]) {
...
-      .concatMap((x, i) => of(x).delay(i === 0 ? 0 : 20, rxTestScheduler));
-    const expected = 'x-y-(z|)';
-    expectObservable(e1).toBe(expected, {x: 10, y: 20, z: 30});
-  });
+    ('should create an observable from an array', () => {
+      const e1 = from([10, 20, 30])
+        // for the purpose of making a nice diagram, spread out the synchronous emissions
+        .concatMap((x, i) => of(x).delay(i === 0 ? 0 : 20, rxTestScheduler));
+      const expected = 'x-y-(z|)';
+      expectObservable(e1).toBe(expected, { x: 10, y: 20, z: 30 });
```
benlesh: Can we undo the formatting changes?

spec/observables/from-spec.ts
```diff
@@ -118,5 +119,26 @@ describe('from', () => {
         );
       expect(nextInvoked).to.equal(false);
     });
+    it(`should accept a function`, (done) => {
```
benlesh: `from` should not accept a function.

`from((...args) => subject.next(...args))` should not work and should error.

The change to `isArrayLike` I can understand, I'm uncertain about this test.

johnlindquist: if `from` accepts a `Promise`, shouldn't it be able to accept a `handler`?

spec/observables/from-spec.ts
johnlindquist: done

spec/observables/from-spec.ts
benlesh: I see, the function has `Symbol.observable` on it. I missed that. Apologies.

benlesh:
Thanks @johnlindquist <3

PR #3565, opened 2018-04-13
fix(build): Update test script to not require build_all step (#3563)

### LINKED ISSUE #3563: Slow Development/Test Cycle
The instructions in README.md indicate that the way to start working in this project is to use the `build_all` and `test` scripts. These scripts appear to clean and retranspile the entire project - that makes sense for CI, but it's a very slow feedback loop for experimenting with ideas locally.  Even the `watch` script just thunks down to this process, and it takes quite a while to re-run.. 

It may make sense to either update the watch script or add a new develop script that runs `tsc` in watch mode to improve the feedback loop time.

darthtrevino:
Add `test:fast` and `watch:fast` scripts to improve local development cycle.

* Remove `--require ts-node/register` from the test script, since that was included in the mocha opts file.
* Fast testing uses the typescript paths feature to use the source typescript directly instead of the fully compiled library.

Fixes #3563


kwonoj:
How much difference actually is with options? from ts-node 4, fast is by default (https://github.com/TypeStrong/ts-node/releases/tag/v4.0.0) and don't need specify it - bit surprising it can  have differences.

darthtrevino:
@kwonoj That makes sense - in this latest commit I've removed TS_NODE_FAST from the tests2png and test:cover scripts. Those scripts don't appear to be working though. I'll see if I can figure out what's up.

darthtrevino:
@kwonoj Those scripts look good on my end now

kwonoj:
I'm feeling actual gain is just by not generating coverage - what numbers actually gained here compare to
1. coverage enabled. vs not in current
2. current without coverage vs. new `fast` script?

esp. around 2.

darthtrevino:
Let me grab some meausurements. The main benefit of the `test:fast` script is that you don't have to run `build_all`.

kwonoj:
> The main benefit of the test:fast script is that you don't have to run build_all.

That should be achieved with current test script directly. Historical reason, we were relying on import from build results of core and it no longer does (as far as I remember correctly) made test script steps bit complex to build first then run test itself. If there's remaining pieces of incorrect import, we just need to fix it and let test script runs to transpile test code and its dependent imports directly, no longer need to call build script separately.

Not sure current test script does explicitly triggers build script currently, I vaguely recall it doesn't.

darthtrevino:
> That should be achieved with current test script directly.

I don't think that's true - according to README.md, build_all is required, and the main test script executes them out of the spec-build directory.

So on my Surface Book in Powershell:

test - 10.729s
test:cover - 22.456s
test:fast - 10.158s
build_all - 3.2 minutes

darthtrevino:
Sure, that makes sense. If you want I can replace the test script with what's in test:fast. I thought maybe you were performing a build step to intentionally verify a built-out version of RxJs.

kwonoj:
Just reading readme once again https://github.com/ReactiveX/rxjs#buildingtesting , those two sections are orthogonal. `test` script should work without build all, build_all is still needed since test shortcuts type checking. So this PR just need to be ensuer current behavior of test script and fix as necessary.

darthtrevino:
@kwonoj Let's talk about the watch script briefly - currently it does a fresh `build_all` and `test` whenever **src/** changes, but this is a super long feedback loop and there's not a lot of benefit you get out of running build_all in a watch loop. 

I have a `watch:test` script that just executes the tests whenever **src/** or **spec/** changes, and if you guys are okay with it, I can just replace the `watch` script with the `watch:test` script.

kwonoj:
I honestly do not see lot of values for `watch:*` script. It is too basic and nearly not used. Personally, rather take out all of `watch:*` and reinvent something if it's really needed.

darthtrevino:
Since you're running Wallaby I can  see that argument, but it makes sense to have some kind of watch mode for driving tests. The watch defined in package.json is much faster than Mocha's watch mode.

darthtrevino:
@kwonoj It's simplified down to a single watch task, I think that makes sense

kwonoj:
> Since you're running Wallaby I can see that argument

No, I'm not saying in perspective of those - it's more about current tool we use for watch is too primitive and doesn't provide usefulness.

>  but it makes sense to have some kind of watch mode for driving tests. 

I already clarified we may need, but somewhat better way. Not sure if it can be included in this PR, rather think separately.

kwonoj:
> watch script is much improved by taking the build_all time out of it altogether

This confuses me. Per discussion we had above, test script should orthogonal to build script now and watch is independent to test.

Also, we are now repeating same thing - discussion's about current watch script is not useful in practical dev flow, so amending it won't helpful as well. Even before we had build orocess taking time, watch script doesn't play well along with compiler already.

I would like to suggest to separate PR for each scope. Per original discussion, pr can be updated without any watch script related changea and it can be dealt separately.

darthtrevino:
> This confuses me. Per discussion we had above, test script should orthogonal to build script now and watch is independent to test.
 
Correct, and it is. However, the watch script on master runs `build_all` and `test` on change.

>  discussion's about current watch script is not useful in practical dev flow, so amending it won't helpful as well.

Maybe that's true for maintainers - I personally like to have a watch script running when I'm developing to have TDD.

> I would like to suggest to separate PR for each scope.
Since the scope of this PR is decoupling `build_all` from `test`, it also makes sense to decouple it from `watch`. 

Any further improvements like concurrent or incremental builds be in other PRs.

kwonoj:
> Maybe that's true for maintainers - I personally like to have a watch script running when I'm developing to have TDD.

Why we are repeating same thing over? It's matter of script is not generally working, regardless of maintainer or not.

I think I have suffeciently shared my opinion and came to feel existence of current watch script itself makes some confusions around our setups. 

I'm going to take out watch script from master and create issues for tracking separately other than this PR.

Issue #3558, opened 2018-04-12
RxJS 6 requires "dom" lib when using TypeScript


dwieeb:
**RxJS version:** 6.0.0-beta.4

**Code to reproduce:**

`index.ts`
```
import { Observable } from 'rxjs';
```

`tsconfig.json`
```
{
  "compilerOptions": {
    "target": "es6",
    "module": "commonjs",
    "moduleResolution": "node",
    "strict": true,
    "lib": [
      "es2015"
    ]
  },
  "files": [
    "index.ts"
  ]
}
```

**Expected behavior:** Successful compilation.

**Actual behavior:** Without `"dom"` in `"lib"`, the following compilation error occurs:

```
❯ npx tsc --project tsconfig.json
node_modules/rxjs/internal/observable/fromEvent.d.ts(10,39): error TS2304: Cannot find name 'EventTarget'.
node_modules/rxjs/internal/observable/fromEvent.d.ts(10,103): error TS2304: Cannot find name 'NodeList'.
node_modules/rxjs/internal/observable/fromEvent.d.ts(10,114): error TS2304: Cannot find name 'HTMLCollection'.
```

cartant:
This is down to having all of the factory functions now exported from within `"rxjs/index.d.ts"`.

The distribution now relies upon a bundler to 'tree-shake' any unused exports. However, TypeScript will still see all of the exports - whether used or not. That means that when using RxJS v6 in a Node application, it's still necessary to include TypeScript's `dom` library.

This has the unfortunate effect of introducing a whole bunch of types and globals that are not appropriate for a Node target.

dwieeb:
You could ship two sets of `.d.ts` files: browser and Node. This would offer a workaround for Node users using TypeScript's `paths` compiler option:

```json
{
  "compilerOptions": {
    "baseUrl": ".",
    "target": "es6",
    "module": "commonjs",
    "moduleResolution": "node",
    "strict": true,
    "paths": {
      "rxjs": ["node_modules/rxjs/node/index.d.ts"]
    },
    "lib": [
      "es2015"
    ]
  },
  "files": [
    "index.ts"
  ]
}

```

jasonaden:
You can also add the option `skipLibCheck`. Without this, you're re-typechecking `d.ts` files that were compiled during the rxjs build process. You don't need to re-typecheck these since they were type checked when produced.

[...]

PR #3566, opened 2018-04-13
fix(rxjs): no longer requires `dom` lib

benlesh:
Adds interfaces to cover what is used from DOM in `fromEvent` so that types are carried through, but `dom` lib isn not required by default even if you are not using `fromEvent`

fixes: #3558

### DIFF
--- a/src/internal/observable/fromEvent.ts
+++ b/src/internal/observable/fromEvent.ts
@@ -13,26 +13,38 @@ export interface NodeStyleEventEmitter {
 
 export type NodeEventHandler = (...args: any[]) => void;
 
-export type JQueryStyleEventEmitter = {
+export interface JQueryStyleEventEmitter {
   on: (eventName: string, handler: Function) => void;
   off: (eventName: string, handler: Function) => void;
-};
+}
 
-export type EventTargetLike = EventTarget | NodeStyleEventEmitter | JQueryStyleEventEmitter | NodeList | HTMLCollection;
+export interface HasEventTargetAddRemove<E> {
+  addEventListener(type: string, listener: ((evt: E) => void) | null, options?: boolean | AddEventListenerOptions): void;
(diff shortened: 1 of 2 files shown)


src/internal/observable/fromEvent.ts
```diff
@@ -177,18 +189,18 @@ export function fromEvent<T>(
...
 
-function setupSubscription<T>(sourceObj: EventTargetLike, eventName: string,
-                              handler: Function, subscriber: Subscriber<T>,
+function setupSubscription<T>(sourceObj: FromEventTarget<T>, eventName: string,
+                              handler: (...args: any[]) => void, subscriber: Subscriber<T>,
                               options?: EventListenerOptions) {
   let unsubscribe: () => void;
-  if (isNodeList(sourceObj) || isHTMLCollection(sourceObj)) {
-    for (let i = 0, len = sourceObj.length; i < len; i++) {
+  if (sourceObj && (sourceObj as any).length) {
```
dwieeb: You can use `isArrayLike` here, right? https://github.com/ReactiveX/rxjs/blob/master/src/internal/util/isArrayLike.ts

specialkk:
Hey, this has some major issues when you want to do `fromEvent(window, 'message')`. `window` isn't a NodeList or HTMLCollection but [it does have a `.length` property](https://developer.mozilla.org/en-US/docs/Web/API/Window/length) which refers to the number of frames in the window. As-is, this will attempt to add event listeners to all frames on the window and cause an error of the form
```
Uncaught DOMException: Blocked a frame with origin "..." from accessing a cross-origin frame.
```
when subscribing.

cartant:
@specialkk Thanks for pointing out a problem.

Could you please create an issue and, from it, reference this PR?

I'm not a maintainer of this repo, but I think I speak for all maintainers in saying that it's preferable to have new issues opened, than to have comments made on closed issues or merged PRs.

PR #3590, opened 2018-04-20
refactor(webSocket): rename back to webSocket ala 5.0

benlesh:
BREAKING CHANGE: UNBREAKING websocket to be named `webSocket` again, just like it was in 5.0. Now you should import from `rxjs/webSocket`


benlesh:
Got it, I didn't update the tsconfig files

kwonoj:
> Error: ENOENT: no such file or directory, lstat '/home/travis/build/ReactiveX/rxjs/src/websocket/package.json'

Build still seems failing, looks related with case sensitivity.

PR #3591, opened 2018-04-20
chore(publish): 6.0.0

benlesh:
This is it.

Once this is merged, I'll publish the release.


benlesh:
I think we're going to shore up the upgrade/migration documentation a little more before proceeding.

Issue #3578, opened 2018-04-17
Issues creating Observable with fromEvent and window since RC 7


specialkk:
**RxJS version:** 6.0.0-uncanny-rc.7

**Code to reproduce:**
```javascript
import { fromEvent } from 'rxjs';

// on a page with multiple frames loaded
fromEvent(window, 'message').subscribe(console.log);
window.postMessage('boop', '*');
```
**Expected behavior:**
This should subscribe to window `'message'` events and log them.

**Actual behavior:**
`window` is incorrectly treated as a NodeList or HTMLCollection-type object since [it has a `.length` property](https://developer.mozilla.org/en-US/docs/Web/API/Window/length) and attempts to set up on the sub-frames rather than the actual window object.  
(When some of the these sub-frames have a different origin, the error
```
Uncaught DOMException: Blocked a frame with origin "..." from accessing a cross-origin frame.
```
is often thrown when subscribing.)

**Additional information:**
This was introduced with the changes in #3566

alvachien:
Meet the same issue.

mchmielarski:
The problem is here: https://github.com/ReactiveX/rxjs/blob/d267b3896b0d7e1392a79c7d8077edf430072a36/src/internal/observable/fromEvent.ts#L196

because `sourceObj.length` is true for window with least one iframe

specialkk:
That's a different issue since that's an issue with type definitions. What version of TypeScript are you using? I can't reproduce your issue.

https://stackblitz.com/edit/typescript-j92cbp

The snippet seems to work fine in my TS 2.7 projects. I also don't believe anything has changed with `Window` in 2.8 either.

adrienverge:
Hi @specialkk, thanks for this fast answer!

I've done some tests and the problem happens with TypeScript 2.5.3 and 2.6.2, only with `"strict": true` in `tsconfig.json`.

I did not manage to reproduce with a fresh Angular 6 `ng new project` + TypeScript 2.7 install (even with `"strict": true`. So I'll dig further into my own project (which still uses TypeScript 2.5.3) to find the exact problem!

Thanks again.

Adrien

PR #3594, opened 2018-04-22
 fix(fromEvent): don't enumerate valid sources

cartant:
**Description:**

This PR adds a failing test and moves the check for a `length` property on the event source to last place - after the checks for DOM, JQuery and Node event sources.

It fixes the problem with using `fromEvent` on `window` - which has a `length` property.

**Related issue (if exists):** #3578


benlesh:
Nice! Thanks, @cartant!

Issue #3604, opened 2018-04-26
Delay operator will leak memory


ubnt-michals:
**RxJS version:** 5 and 6

**Code to reproduce:**
5
```.js
const { Observable } = require('rxjs');

const setup = delay => Observable.timer(1)
  .repeatWhen(notifications => notifications.delay(delay))
  .subscribe();

let instances = 100;
while (instances--) {
  setup(100);
}
```

6
```.js
const { timer } = require('rxjs');
const { repeatWhen, delay } = require('rxjs/operators');

const setup = d => timer(1)
  .pipe(
    repeatWhen(notifications => notifications.pipe(delay(d)))
  )
  .subscribe();

let instances = 100;
while (instances--) {
  setup(100);
}
```

**Expected behavior:**
Memory shouldn't grow

**Actual behavior:**
Memory grows indefinitely

**Additional information:**
Have fix and will provide PR

benlesh:
I'm not sure this is a problem with delay so much as it might be a problem with repeatWhen. The recursive nature of what you're doing here is suspect.

ubnt-michals:
Where is the recursion? I running asynchronous action and resubscribing it with delay. 

In synchronous code it would be something like
```
while(true) {
  doStuff()
  sleep(1)
}
```

The code above when replaced with
```.js
timer(1)
  .pipe(
    repeatWhen(notifications => notifications.pipe(mergeMapTo(timer(d))))
  )
  .subscribe();
```
is not leaking.

ubnt-michals:
Fixed in #3605

PR #3605, opened 2018-04-26
fix(delay): fix memory leak

ubnt-michals:
**Description:**
Hopefully fix leaking memory for delay operator

EDIT:
I had to bend the VirtualTimeScheduler to not `shift` the last action in order to properly test this. This could be potential breaking change if left as it is. I am not sure how to test the case without the change though. Any help is appreciated.

#3604

### DIFF
--- a/spec/operators/delay-spec.ts
+++ b/spec/operators/delay-spec.ts
@@ -1,5 +1,5 @@
-import { Observable, timer } from 'rxjs';
-import { delay, repeatWhen, skip, take, tap, finalize } from 'rxjs/operators';
+import { Observable } from 'rxjs';
+import { delay, repeatWhen, skip, take, tap } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import * as sinon from 'sinon';
 import { expect } from 'chai';
@@ -147,12 +147,12 @@ describe('Observable.prototype.delay', () => {
     expectObservable(result).toBe(expected);
   });
 
-  it('should unsubscribe scheduled action when result is unsubscribed explicitly', () => {
+  it('should unsubscribe scheduled actions after execution', () => {
     let subscribeSpy: any = null;
     const counts: number[] = [];
 
     const e1 =       cold('a|');
-    const expected =     '--a-(a|)';
+    const expected =      '--a-(a|)';
     const duration = time('-|');
     const result = e1.pipe(
       repeatWhen(notifications => {
--- a/spec/operators/delay-spec.ts
+++ b/spec/operators/delay-spec.ts
@@ -1,9 +1,12 @@
-import * as Rx from 'rxjs/Rx';
+import { Observable, timer } from 'rxjs';
+import { delay, repeatWhen, skip, take, tap, finalize } from 'rxjs/operators';
+import { TestScheduler } from 'rxjs/testing';
+import * as sinon from 'sinon';
+import { expect } from 'chai';
 import { hot, cold, expectObservable, expectSubscriptions, time } from '../helpers/marble-testing';
 
 declare const asDiagram: Function;
-declare const rxTestScheduler: Rx.TestScheduler;
-const Observable = Rx.Observable;
+declare const rxTestScheduler: TestScheduler;
 
 /** @test {delay} */
 describe('Observable.prototype.delay', () => {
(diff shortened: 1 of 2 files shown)


src/scheduler/VirtualTimeScheduler.ts
```diff
@@ -24,7 +24,15 @@ export class VirtualTimeScheduler extends AsyncScheduler {
     const {actions, maxFrames} = this;
     let error: any, action: AsyncAction<any>;
 
-    while ((action = actions.shift()) && (this.frame = action.delay) <= maxFrames) {
+    while (actions.length > 0) {
```
benlesh: To fix delay, we shouldn't have done anything to the VirtualTimeScheduler, changes to the virtual scheduler will affect _all marble tests_.

src/operators/delay.ts
```diff
@@ -79,7 +79,7 @@ interface DelayState<T> {
  */
 class DelaySubscriber<T> extends Subscriber<T> {
   private queue: Array<DelayMessage<T>> = [];
-  private active: boolean = false;
+  private nextAction: Subscription = null;
```
benlesh: i'm suspicious this will have issues in situations where more than one delayed action is dispatched synchronously. For example `of(1, 2, 3).pipe(delay(10))` or the like. You'd have to keep more than one subscription for each scheduled action.

src/scheduler/VirtualTimeScheduler.ts
ubnt-michals: This change has nothing to do with the fix. It's just I wanted to test that the last action is correctly unsubscribed and when calling scheduler.flush() it removed the action that wasn't executed 
```
actions.length === 1
maxFrames === 60
action.delay === 80 
(action = actions.shift()) && (this.frame = action.delay) <= maxFrames
true && false
// no execution
actions.length === 0 // same as if action would execute
```

src/operators/delay.ts
ubnt-michals: The original code would only schedule one action. `this.active` would be set to `true` and additional messages would be buffered until dispatch. On dispatch the queue would be processed... etc. I hope I am interpreting it correctly...

I think this code does the same except it doesn't register unsubscription logic that is the cause of the leak.

cartant:
@benlesh I'll review this later today. If `sinon` is used, it should be testable without modifying the scheduler.

src/operators/delay.ts
```diff
@@ -96,7 +96,8 @@ class DelaySubscriber<T> extends Subscriber<T> {
       const delay = Math.max(0, queue[0].time - scheduler.now());
       this.schedule(state, delay);
     } else {
-      source.active = false;
```
cartant: I think the fix is a one-liner and `source.remove(this);` should be added here.

ubnt-michals: Good point, I've totally missed that. Will fix the PR. Thanks.

benlesh:
Thank you, @cartant, for taking the time to review this in more detail.

src/internal/operators/delay.ts
```diff
@@ -92,6 +92,7 @@ class DelaySubscriber<T> extends Subscriber<T> {
       const delay = Math.max(0, queue[0].time - scheduler.now());
       this.schedule(state, delay);
     } else {
+      source.remove(this);
```
cartant: I've changed my mind with this. I think `this.unsubscribe()` instead of `source.remove(this)` would be better. The effected behaviour will be identical, but the former is more idiomatic.

spec/operators/delay-spec.ts
```diff
@@ -143,4 +146,33 @@ describe('Observable.prototype.delay', () => {
 
     expectObservable(result).toBe(expected);
   });
+
+  it('should unsubscribe scheduled action when result is unsubscribed explicitly', () => {
```
cartant: I minor nit: the description doesn't actually describe the problem. The leak happens because the actions are not removed/unsubscribed from the subscriber's internal subscriptions after the action is executed. I think a more accurate description would be something like `should unsubscribe scheduled actions after execution`.

The actions will be unsubscribed when the subscriber unsubscribes; it's just that they will accumulate up until that happens.

ubnt-michals:
@cartant All should be fixed now. I've learned a lot finding and fixing this bug. Thanks for your time.

benlesh:
Thank you, @ubnt-michals! (and thanks for the review @cartant )

Issue #3603, opened 2018-04-26
Should the race operator be deprecated?


cartant:
**RxJS version:** 6.0.0

**Code to reproduce:**

```ts
const winner = of(1).pipe(race(of(2)));
```

**Expected behavior:**

I expected a TSLint deprecation warning, as the `race` operator has a namesake observable factory function.

**Actual behavior:** 

No warning is effected.

**Additional information:**

The other operators that have namesake observable factory functions (`concat`, `merge`, etc.) are deprecated, but [`race` is not](https://github.com/ReactiveX/rxjs/blob/6.0.0/src/internal/operators/race.ts#L6-L11).

benlesh:
Yes. It should be deprecated. Thanks for catching that.

natmegs:
I'll work on a PR for this!

cartant:
Closed by #3611.

PR #3611, opened 2018-04-26
refactor(race): deprecate race operator in favor of static race

natmegs:
deprecate race operator in favor of static race, similar to other deprecated operators concat,
merge, etc

#3603

### DIFF
--- a/src/internal/operators/race.ts
+++ b/src/internal/operators/race.ts
@@ -4,19 +4,18 @@ import { MonoTypeOperatorFunction, OperatorFunction } from '../types';
 import { race as raceStatic } from '../observable/race';
 
 /* tslint:disable:max-line-length */
+/** @deprecated Deprecated in favor of static race. */
 export function race<T>(observables: Array<Observable<T>>): MonoTypeOperatorFunction<T>;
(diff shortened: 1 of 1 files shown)


src/internal/operators/race.ts
```diff
@@ -4,19 +4,18 @@ import { MonoTypeOperatorFunction, OperatorFunction } from '../types';
...
+/** @deprecated Deprecated in favor of static race. */
 export function race<T, R>(observables: Array<Observable<T>>): OperatorFunction<T, R>;
+/** @deprecated Deprecated in favor of static race. */
 export function race<T>(...observables: Array<Observable<T> | Array<Observable<T>>>): MonoTypeOperatorFunction<T>;
+/** @deprecated Deprecated in favor of static race. */
 export function race<T, R>(...observables: Array<Observable<any> | Array<Observable<any>>>): OperatorFunction<T, R>;
 /* tslint:enable:max-line-length */
 
 /**
- * Returns an Observable that mirrors the first source Observable to emit an item
```
benlesh: Can leave the documentation here, just mark it deprecated.

Issue #3609, opened 2018-04-26
mergeMap/concatMap with limited concurrency will blow the stack


ubnt-michals:
**RxJS version:** 5.5.10

**Code to reproduce:**

```.js
const { Observable, Subject } = require('rxjs');

const signal = new Subject();
const terminate$ = signal.take(1).publishLast();

terminate$.connect();

const create = () => Observable.timer(1000).takeUntil(terminate$);

Observable.range(1, 1000)
  .mergeMap(() => create(), 10) // or concatMap
  .subscribe();

setTimeout(() => {
  signal.next();
}, 500);
```

**Expected behavior:**
I would expect it to not blow the stack

**Actual behavior:**
RangeError: Maximum call stack size exceeded

**Additional information:**
I haven't tested 6.0.0

cartant:
This is fixable. It's a re-enterancy issue that's due to synchronous `complete` notifications from buffered observables. I'll create a PR with a fix sometime soon. However, I'm curious: did you come across this problem in a real-world situation?

ubnt-michals:
@cartant This and #3604 is from production. We maintain a lot of long running WebSocket connections - this particular case is an edge case that happened probably because of a network outage when all connections disconnected at once.

cartant:
The solution to the problem is to use the `QueueScheduler`:

```js
const { Observable, Scheduler, Subject } = require('rxjs');

const signal = new Subject();
const terminate$ = signal.take(1).publishLast();

terminate$.connect();

const create = () => Observable.timer(1000)
  .takeUntil(terminate$)
  .observeOn(Scheduler.queue); // HERE

Observable.range(1, 1000)
  .mergeMap(() => create(), 10)
  .subscribe();

setTimeout(() => {
  signal.next();
}, 500);
```

The re-enterant behaviour is [by design](https://github.com/ReactiveX/rxjs/pull/3621#issuecomment-386356358) and [using `observeOn`](https://github.com/ReactiveX/rxjs/pull/3621#issuecomment-386472093) is the mechanism to effect the non-re-enterant behaviour.

ubnt-michals:
> The re-enterant behaviour is by design and using observeOn is the mechanism to effect the non-re-enterant behaviour.

This is not a good design then... It was non trivial to find this issue in our codebase, especially when there are a lot of operators involved. 

Also this essentially means that every Observable I use in concatMap&friends has to be using `QueueScheduler` otherwise SO is possible if someone adds cancellation, right?

[...]

PR #3621, opened 2018-04-29
fix(mergeMap): avoid re-enterancy/stack overflow

cartant:
**Description:**

When `mergeMap` is used with limited concurrency and when the buffered observables complete synchronously, the current implementation sees the `notifyComplete` method called in a re-enterant manner.

That is, a complete notification effects the removal of the next observable from the buffer and it's then subscribed to. If that observable complete synchronously, `notifyComplete` is called again and if there is a sufficient number of buffered observables, a stack overflow error will be effected.

The linked issue has a repro, but I'll include another (minimal) repro here:

```ts
import { EMPTY, range, timer } from "rxjs";
import { concatMap } from "rxjs/operators";

range(0, 10000).pipe(
    concatMap(value => (value === 0) ? timer(0) : EMPTY)
).subscribe();
```

If the buffered observables are more complicated and are composed from more operators, the number of buffered observables required to effect an overflow will be significantly lower than the 10,000 used in the repro.

This PR fixes the problem (in a synchronous manner) by queuing the observables shifted from the buffer.

**Related issue (if exists):** #3609

### DIFF
--- a/spec/operators/mergeMap-spec.ts
+++ b/spec/operators/mergeMap-spec.ts
@@ -1,10 +1,12 @@
 import { expect } from 'chai';
 import { mergeMap, map } from 'rxjs/operators';
-import { Observable, from, of } from 'rxjs';
+import { Observable, Observer, from, of, range, timer } from 'rxjs';
+import { TestScheduler } from 'rxjs/testing';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 
 declare const type: Function;
 declare const asDiagram: Function;
+declare const rxTestScheduler: TestScheduler;
 
 /** @test {mergeMap} */
 describe('mergeMap', () => {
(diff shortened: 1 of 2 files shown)


benlesh:
@cartant, this is actually by design. giving people the option of depth-first or breadth first execution. If someone runs into this situation, they should use the `queueScheduler`. I'm afraid I can't merge this one.

benlesh:
I really appreciate the thought and effort that went into this one, though.

cartant:
Unfortunately, this cannot be avoided using the `QueueScheduler`, as it's not possible to pass a scheduler to `mergeMap` and the problem occurs when `mergeMap` processes the notifications that have accumulated in its internal buffer.

So, as long as there is a something that effects the accumulation, there is no way of avoiding the problem. For example, if I adding the `QueueScheduler` to the repro and using a non-zero duration for the `timer` does not help:

```ts
import { EMPTY, queueScheduler, range, timer } from "rxjs";
import { concatMap } from "rxjs/operators";

range(0, 10000, queueScheduler).pipe(
    concatMap(value => (value === 0) ? timer(1, queueScheduler) : EMPTY)
).subscribe();
```

cartant:
Whoops. Scratch that; I'm wrong.

It can be solved by using `observeOn` with the inner observables:

```ts
import { EMPTY, queueScheduler, range, timer } from "rxjs";
import { concatMap, observeOn } from "rxjs/operators";

range(0, 10000).pipe(
    concatMap(value => (value === 0) ? timer(0) : EMPTY.pipe(observeOn(queueScheduler)))
).subscribe();
```

/cc @ubnt-michals

ubnt-michals:
There should be some kind of warning in the docs that this could happen... We have this issue all over the codebase and the thing is it's not obvious! - the tests won't show it, because I expected that if Observable chain works for ten items it's going to work for ten thousands.

As I wrote in  #3609 it's a bad design IMHO.

PR #3623, opened 2018-04-30
test: empty Observable.toPromise

NickLydon:
**Description:**
show that an empty Observable results in a promise resolved with a
value of undefined


benlesh:
Thank you for the thoughtful contribution, @NickLydon!

PR #3628, opened 2018-05-01
feat(isObservable): adds recommended method for checking objects to see if they're RxJS 5+ Observables

benlesh:
### Overview

There are a lot of ways to check to see of something is an Observable, none of them are great. This is an attempt to provide a recommended way to see of an Observable is, in fact, an RxJS (5+) Observable or not.

- `instanceof` doesn't always work in node, as sometimes people have Observables from other copies of RxJS.
- `typeof o.subscribe === 'function'` doesn't always work, because older versions of RxJS (<4) have different APIs that still use `subscribe`
- I've seen `typeof o === 'object'` sort of checks to see if something is an Observable, which has obvious problems.

### The goals of this addition:

- Provide a library-blessed way to identify compatible observables
- Provide a future-proof way to identify observables, as implementation details are likely to change in upcoming versions

### Other things to note:

the type `ObservableLike<T>` is deprecated in favor of the name `InteropObservable<T>`.

### DIFF
--- a/src/internal/types.ts
+++ b/src/internal/types.ts
@@ -34,7 +34,7 @@ export interface SubscriptionLike extends Unsubscribable {
   readonly closed: boolean;
 }
 
-export type SubscribableOrPromise<T> = Subscribable<T> | Subscribable<never> | PromiseLike<T> | ObservableLike<T>;
+export type SubscribableOrPromise<T> = Subscribable<T> | Subscribable<never> | PromiseLike<T> | InteropObservable<T>;
 
 /** OBSERVABLE INTERFACES */
 
@@ -45,7 +45,11 @@ export interface Subscribable<T> {
 }
 
 export type ObservableInput<T> = SubscribableOrPromise<T> | ArrayLike<T> | Iterable<T>;
-export type ObservableLike<T> = { [Symbol.observable]: () => Subscribable<T>; };
+
+/** @deprecated use {@link InteropObservable } */
+export type ObservableLike<T> = InteropObservable<T>;
+
+export type InteropObservable<T> = { [Symbol.observable]: () => Subscribable<T>; };
 
 /** OBSERVER INTERFACES */
 
--- a/src/internal/util/isObservable.ts
+++ b/src/internal/util/isObservable.ts
@@ -1,0 +1,10 @@
+import { Observable } from '../Observable';
+import { ObservableInput } from '../types';
+
+/**
+ * Tests to see if the object is an RxJS {@link Observable}
+ * @param obj the object to test
+ */
+export function isObservable<T>(obj: any): obj is Observable<T> {
+  return obj && obj instanceof Observable || (typeof obj.lift === 'function' && typeof obj.subscribe === 'function');
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,10 +10,10 @@
       "integrity": "sha512-q6pUXxb8gZMguuh00ATWHItLw38YT2RE353LptCh38uQQ16bGx/8YxJVJZ0IsLK7mEhd2PtkyRzoyM3ttrYhag==",
       "dev": true,
       "requires": {
-        "loader-utils": "^1.1.0",
-        "source-map": "^0.5.6",
-        "typescript": "~2.7.2",
-        "webpack-sources": "^1.1.0"
+        "loader-utils": "1.1.0",
+        "source-map": "0.5.7",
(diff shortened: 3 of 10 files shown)


kwonoj:
before reading any code, why this PR contains massive changes in pkg lock? codes doesn't seem to have any pkg changes.

src/internal/util/isObservable.ts
```diff
@@ -1,7 +1,10 @@
...
 
-/** Identifies an input as being Observable (but not necessary an Rx Observable) */
-export function isObservable(input: any): input is ObservableLike<any> {
-  return input && typeof input[Symbol_observable] === 'function';
+/**
+ * Tests to see if the object is an RxJS {@link Observable}
+ * @param obj the object to test
+ */
+export function isObservable<T>(obj: any): obj is Observable<T> {
+  return obj && obj instanceof Observable || (typeof obj.lift === 'function' && typeof obj.subscribe === 'function');
```
kwonoj: I recall discussions around supporting interop between multiple installations, but using `instanceof` doesn't breaks it?

benlesh: If you'll notice, that's just a short circuit.

If it's `instanceof` it won't do the property lookups immediately after.

benlesh:
There's a package-lock, because I did an npm install and it generated a new package lock. Seems okay to me.

kwonoj:
🤔 pkg-lock supposed to not change when do simple npm install, as it should serve as lock. not sure what caused it, but maybe ok as we don't deal with deps but mostly it's devdeps.

cartant:
@kwonoj The lock file isn't supposed to change, but I see this all the time with the RxJS repo.

npm's seemigly inexplicable lock-file-related behaviour was one of the reasons I switched to yarn. I just gave up trying to understand what npm was doing.

kwonoj:
What version of npm you (@benlesh also) are using?

cartant:
```
$ npm --version
5.5.1
```

I just rebased my `master` and ran `npm install` and, sure enough, `package-lock.json` has been updated.

kwonoj:
Try 6, or either at least 5.6.0.

src/internal/types.ts
```diff
@@ -45,7 +45,11 @@ export interface Subscribable<T> {
 }
 
 export type ObservableInput<T> = SubscribableOrPromise<T> | ArrayLike<T> | Iterable<T>;
-export type ObservableLike<T> = { [Symbol.observable]: () => Subscribable<T>; };
+
+/** @deprecated use {@link InteropObservable } */
+export type ObservableLike<T> = InteropObservable<T>;
+
+export type InteropObservable<T> = { [Symbol.observable]: () => Subscribable<T>; };
```
alfaproject: Just curious, but why the new name?

cartant:
I've upgraded to 5.6 and I've re-synched with `master`. Running `npm install` still effects changes to `package-lock.json`.

Scrolling through the diff, all of the differences seem to involve the removal of `~` and `^` characters from the versions and some of the versions themselves have changed, too.

Maybe I just don't understand how this works, but my expectation was that upon running `npm install` the content of `node_modules` would be altered to reflect what's in the lock file. That's how yarn behaves and that's one of the reasons I switched. The other being yarn's support for OS-specific, optional dependencies - which I believe npm now supports, too.

These days, with the RxJS repo, I just do an `npm install` and then do `git checkout -f` to dump the changed lock file. It's entirely possible that my doing so is part of the problem, but I'm just so over it.

alfaproject:
@cartant just do `npm ci` and that will behave the same as `yarn`. Also much faster to run.

cartant:
@alfaproject Thanks. I will try that when I next work with the RxJS repo.

benlesh:
Yeah, I'm on npm v 6. I'll try `npm ci` next time.

src/internal/util/isObservable.ts
AviVahl: @benlesh the function will throw if it gets `null` or `undefined`. Could add extra parenthesis to fix:

`return obj && (obj instanceof Observable || (typeof obj.lift === 'function' && typeof obj.subscribe === 'function'));`
(but it should probably also be tested)

src/internal/types.ts
cartant: @alfaproject I guess it's because `Observable` is not assignable to `ObservableLike`. Every other type with a `Like` suffix that I can think of is assignable to its namesake type that does not have the suffix.

PR #3630, opened 2018-05-02
Fix quote and unnecessary tag in MIGRATION.md

yamachu:
**Description:**

This PR fixes charactors that cannot be displayed and removes unnecessary html tag.


benlesh:
Thank you, @yamachu! :)

PR #3640, opened 2018-05-04
refactor(audit-spec): update audit operator tests to use pipeable ope…

natmegs:
…rators

update audit spec to use updated import sites and pipeable operators


benlesh:
Nice! Thanks, @natmegs!

PR #3642, opened 2018-05-04
refactor(bufferCount-spec): update bufferCount operator tests to use pipeable operators

natmegs:
update bufferCount spec to use updated import sites and pipeable operators


benlesh:
So awesome, thanks again, @natmegs!

PR #3645, opened 2018-05-05
refactor(bufferToggle-spec): update bufferToggle operator tests to use pipeable operators

natmegs:
update bufferToggle spec to use updated import sites and pipeable operators


benlesh:
Thanks again, @natmegs!

PR #3648, opened 2018-05-05
refactor(catch-spec): update catchError operator tests to use pipeable operators

natmegs:
update catch spec to use updated import sites and pipeable operators


benlesh:
And... thanks, again, @natmegs!!

PR #3652, opened 2018-05-06
refactor(concatAll-spec): update concatAll operator tests to use pipeable operators

natmegs:
update concatAll spec to use updated import sites and pipeable operators


cartant:
This is great, @natmegs. I think I speak for everyone in saying that your doing this is appreciated. Thanks!

natmegs:
@cartant happy to be able to help in a small way :)

Issue #3606, opened 2018-04-26
Ajax response with status 0 must throw an error


eltiodelsofa:
If ajax observable has parameter timeout configured, with a slow request, ajax is emitting a next instead of an error.

**RxJS version:**
 6.0.0

**Code to reproduce:**
```js
ajax({
  url: 'https://reqres.in/api/users?delay=5',
  timeout: 500,
}).subscribe({
  next: v => console.log('next', v),
  error: err => console.log('err', err),
});
```

**Expected behavior:**

Emit an error event with an AjaxError object

**Actual behavior:**

Emit a next event with ajax null response and status 0

felixfbecker:
Workaround:
```ts
tap(response => {
            if (response.status === 0) {
                throw Object.assign(new Error('Ajax status 0'), response)
            }
        }),
```

adriensamson:
I believe #3653 also fixes the issue with down network: moving the next() logic from `onreadystatechange` to `onload` allows `onerror` to trigger first.

felixfbecker:
@bbonnet Was this fixed by https://github.com/ReactiveX/rxjs/pull/3653?

goenning:
Tested and working on 6.4.*

![rxjs-ajax](https://user-images.githubusercontent.com/94755/57067684-95db9c00-6cc7-11e9-8f7c-6c303f005093.PNG)

ndlarsen:
In which version was this actually fixed?

cartant:
Closed by #3653

PR #3653, opened 2018-05-06
fix(ajax): Handle timeouts as errors

bbonnet:
**Description:**
XHR Timeouts are not being handled properly.

It turns out the XHR timeouts fire the `readystatechange` event (with status=0, readyState=4) and then the `timeout` event.  This is causing `next` to be called before `error`.

Confirmed this behavior in Chrome, added a test for timeouts and fixed the behavior by pointing the main event handler to fire on `load`.  Definitely recommend confirming this behavior and compatibility in other browsers.

Fixed up our `MockXMLHttpRequest` to cause timeouts after the `send` method is called.  
Overall, I want to recommend moving away from this `MockXMLHttpRequest` so that we aren't re-implementing browser functionality in the tests.  I saw that the current browser tests are broken, wanted to take a stab at either fixing those or porting over some integration tests to Karma.  Totally open to suggestions there. 

This is my first contribution to rxjs, really open to feedback and suggestions! 

https://github.com/ReactiveX/rxjs/issues/3606

### DIFF
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -380,6 +380,38 @@ describe('Observable.ajax', () => {
     });
   });
 
+  it('should error on timeout of asynchronous request', (done) => {
+    const obj: Rx.AjaxRequest = {
+      url: '/flibbertyJibbet',
+      responseType: 'text',
+      timeout: 10
+    };
+
+    Rx.Observable.ajax(obj)
+      .subscribe((x: any) => {
+        throw 'should not have been called';
+      }, (e) => {
+        expect(e.status).to.equal(0);
+        expect(e.xhr.method).to.equal('GET');
+        expect(e.xhr.async).to.equal(true);
+        expect(e.xhr.timeout).to.equal(10);
+        expect(e.xhr.responseType).to.equal('text');
+      });
+
+    const request = MockXMLHttpRequest.mostRecent;
+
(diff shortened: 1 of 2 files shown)


spec/observables/dom/ajax-spec.ts
```diff
@@ -380,6 +380,38 @@ describe('Observable.ajax', () => {
...
+        expect(e.xhr.async).to.equal(true);
+        expect(e.xhr.timeout).to.equal(10);
+        expect(e.xhr.responseType).to.equal('text');
+      });
+
+    const request = MockXMLHttpRequest.mostRecent;
+
+    expect(request.url).to.equal('/flibbertyJibbet');
+
+    setTimeout(() => {
```
benlesh: We need a way to virtualize this, perhaps with the TestScheduler, so the test doesn't take longer than a second to run. It will increase the time it takes to test everything by up to 50% if we leave this timeout in here.

benlesh: The test scheduler is available in all tests, to schedule something in "virtual time" you'd do the following:

```ts
rxTestScheduler.schedule(() => { /* thing to do */ }, 1000)

// later flush it to run all scheduled things synchronously
rxTestScheduler.flush();
```

bbonnet: Great, updated to use the test scheduler instead.

benlesh:
Thanks, @bbonnet!

Issue #3602, opened 2018-04-25
RxJS v6 TimeoutError is missing name property


felixfbecker:
**RxJS version:** 6.0.0

**Code to reproduce:**

```ts
export const browserExtensionInstalled = fromEvent(document, 'custom-event').pipe(
    mapTo(true),
    timeout(500),
    catchError(err => {
        if (err.name === 'TimeoutError') {
            return [false]
        }
        throw err
    }),
)
```
**Expected behavior:**

Should emit `false` after 500ms.

**Actual behavior:**

Throws an error `Error: Timeout has occurred`. (worked in RxJS 5)

**Additional information:**

This broke in commit https://github.com/ReactiveX/rxjs/commit/2f395da56189090215462bb88cfd2014c842be57#diff-cd4b33df1ab129e23dc2bc5c12d8369b

`UnsubscriptionError` still has the `name` set, but `TimeoutError` doesn't, and it's not mentioned in the release notes, so I assume this is an unintentional bug (in any case it can be reintroduced in a semver-patch).

https://github.com/ReactiveX/rxjs/blob/b948e651949eb7b4c1103d7013fc43f49ac1b870/src/internal/util/UnsubscriptionError.ts#L5-L13

https://github.com/ReactiveX/rxjs/blob/b948e651949eb7b4c1103d7013fc43f49ac1b870/src/internal/util/TimeoutError.ts#L8-L13

_Why not use `instanceof`?_ `instanceof` breaks when multiple RxJS versions are in play, or even just the same version had to be installed multiple times because deduping wasn't possible. It asserts the implementation, `err.name` asserts an interface. All browser errors use it (e.g. `AbortError`).

cartant:
This might be the first issue that'll effect a 6.0.1 release. I wonder if you get a prize? :thinking:

martinsik:
This seems to be fixed in https://github.com/ReactiveX/rxjs/pull/3656.

PR #3656, opened 2018-05-07
fix(ajax): RxJS v6 TimeoutError is missing name property

bbonnet:
**Description:**
Update the `name` property of the `AjaxTimeoutError` to be "AjaxTimeoutError" instead of "AjaxError".

@felixfbecker mentioned the following in the original issue: 
> Why not use instanceof? instanceof breaks when multiple RxJS versions are in play, or even just the same version had to be installed multiple times because deduping wasn't possible. It asserts the implementation, err.name asserts an interface. All browser errors use it (e.g. AbortError).

This still will cause problems if the implementation name changes.  Open to feedback for a more permanent solution.

https://github.com/ReactiveX/rxjs/issues/3602

### DIFF
--- a/src/internal/util/TimeoutError.ts
+++ b/src/internal/util/TimeoutError.ts
@@ -8,7 +8,7 @@
 export class TimeoutError extends Error {
   constructor() {
     super('Timeout has occurred');
-
+    this.name = 'TimeoutError';
     (Object as any).setPrototypeOf(this, TimeoutError.prototype);
   }
 }
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -273,6 +273,7 @@ describe('Observable.ajax', () => {
     });
 
     expect(error instanceof Rx.AjaxError).to.be.true;
+    expect(error.name).to.equal('AjaxError');
     expect(error.message).to.equal('ajax error 404');
     expect(error.status).to.equal(404);
   });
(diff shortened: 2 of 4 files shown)


felixfbecker:
Note this doesn't fix #3602, as that uses [a different error](https://github.com/ReactiveX/rxjs/blob/b948e651949eb7b4c1103d7013fc43f49ac1b870/src/internal/util/TimeoutError.ts#L8-L13).

bbonnet:
Ah I see, that is my mistake.  I will look into a better fix that will cover the `TimeoutError` as well

bbonnet:
Updated to fix `TimeoutError`.  

Also noticed that some of the tests that are implemented like the following are not hitting some of the expectations:.  This test will still pass. 
```ts
   result.subscribe(() => {
      throw new Error('this should not next');
    }, err => {
      // Never actually getting tested
      expect(err).to.be.an.instanceof(Rx.TimeoutError);
      expect(err.name).to.equal('BlahBlahBlah');
    }, () => {
      throw new Error('this should not complete');
    });
```

I updated as follows to get the assertions to actually hit.  Seen this pattern elsewhere in the tests. 
```ts
    let error;
    result.subscribe(() => {
      throw new Error('this should not next');
    }, err => {
      error = err;
    }, () => {
      throw new Error('this should not complete');
    });
    rxTestScheduler.flush();
    expect(error).to.be.an.instanceof(Rx.TimeoutError);
    expect(error.name).to.equal('TimeoutError');
```

src/internal/util/TimeoutError.ts
```diff
@@ -1,14 +1,14 @@
...
+ *
+ * @class TimeoutError
+ */
+export class TimeoutError extends Error {
+  constructor() {
+    super('Timeout has occurred');
+    this.name = 'TimeoutError';
+    (Object as any).setPrototypeOf(this, TimeoutError.prototype);
+  }
+}
```
felixfbecker: did line endings change in this file?

benlesh:
Thank you, @bbonnet!

src/internal/util/TimeoutError.ts
```diff
@@ -8,7 +8,7 @@
 export class TimeoutError extends Error {
   constructor() {
     super('Timeout has occurred');
-
+    this.name = 'TimeoutError';
```
felixfbecker: it would be nice if this was declared as a property so it is a visible part of the API contract:
```ts
public readonly name: 'TimeoutError' = 'TimeoutError'
```

PR #3667, opened 2018-05-08
fix(delay): fix memory leak (in stable)

cartant:
**Description:**

This PR cherry picks PR #3605 into the `stable` branch. Given that it fixes a memory leak, I think there's something to be said for its application to v5, too.

**Related issue (if exists):** #3604

### DIFF
--- a/spec/operators/delay-spec.ts
+++ b/spec/operators/delay-spec.ts
@@ -1,4 +1,7 @@
+import { expect } from 'chai';
+import * as sinon from 'sinon';
 import * as Rx from '../../dist/package/Rx';
+import { delay, repeatWhen, skip, take, tap } from '../../dist/package/operators';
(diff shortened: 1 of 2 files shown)


benlesh:
@cartant ... I'm torn on this, if we patch 5.x, we'll also need to patch 5.6.0-forward-compat. At some point we're going to need to move on (unless there's some glaring security issue).

benlesh:
cc @kwonoj

cartant:
@benlesh Yeah, I get that not all fixes should be backported, but memory leaks are especially insidious. And the `delay` operator is not uncommon.

I guess it depends upon how many apps using v5 would benefit from the fix. To benefit, they'd have to be using `delay` on a long-lived stream.

That fact that it hasn't been a reported as an error is reasonable argument for not backporting. But memory leaks are hard to find.

I don't have a strong opinion either way, but I'd favour backporting.

PR #3675, opened 2018-05-09
refactor(delay-spec): update delay operator tests to use pipeable operators

natmegs:
update delay spec to use updated import sites and pipeable operators


benlesh:
Thanks again, @natmegs!

PR #3678, opened 2018-05-10
chore(Error): Make error name part of visible api contract

bbonnet:
**Description:**
Add `Error.name` to all custom errors.  Add tests for name and description. 
 
https://github.com/ReactiveX/rxjs/issues/3602
https://github.com/ReactiveX/rxjs/pull/3656

### DIFF
--- a/src/internal/util/EmptyError.ts
+++ b/src/internal/util/EmptyError.ts
@@ -10,7 +10,7 @@
  */
 export class EmptyError extends Error {
 
-  public readonly name: 'EmptyError' = 'EmptyError';
+  public readonly name = 'EmptyError';
 
   constructor() {
     super('no elements in sequence');
--- a/src/internal/util/EmptyError.ts
+++ b/src/internal/util/EmptyError.ts
@@ -9,9 +9,11 @@
  * @class EmptyError
  */
 export class EmptyError extends Error {
+
+  public readonly name: 'EmptyError' = 'EmptyError';
+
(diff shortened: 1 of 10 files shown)


src/internal/util/EmptyError.ts
```diff
@@ -9,9 +9,11 @@
  * @class EmptyError
  */
 export class EmptyError extends Error {
+
+  public readonly name: 'EmptyError' = 'EmptyError';
```
martinsik: I'm surprised why is this notation `name: 'EmptyError' = 'EmptyError';` required? It think using just:

    public readonly name = 'EmptyError';

should be enough for TypeScript's to type guard to enforce only string declared as `readonly` anyway. [Here the same thing is defined](https://github.com/ReactiveX/rxjs/pull/3678/files#diff-5446941a3fbe64a64353c2fa505ded36R493) as `name: string = 'AjaxTimeoutError';`.

bbonnet: @martinsik You're right, we only need to do `public readonly name = 'EmptyError';` to enforce.  

The `AjaxTimeoutError` was implementing the parent class `AjaxError`'s `name` property so that was why it was a string.  I added a new intersection type `type AjaxErrors = 'AjaxError' | 'AjaxTimeoutError';` Let me know what you think.

benlesh:
Thanks, @bbonnet!

Issue #1709, opened 2016-05-13
Add `endWith`


benlesh:
Frequently while developing with our redux middleware, we find the need to have an operator `source.endWith('foo')`. Generally, this can be accomplished with a `source.concat(Observable.of('foo'))`, but it's just more verbose.

`endWith` would be a nice to have for our use-cases at Netflix.

cc/ @jayphelps

jayphelps:
RxJava calls this `concatWith`

https://github.com/ReactiveX/RxJava/issues/1176

Personally, I think `endWith` is more elegant since it's opposite of `startWith` and more self-documenting, even though Rx gurus _should_ know what concat means, if you're not one it is easier to grok.

blesh:
> Rx gurus

Wherever _those_ are. lol.

benjchristensen:
We called it `concatWith` since it is literally just an instance method for `concat`. It is one of many we used this pattern for where we added `*With` instance methods as siblings to their static counterparts, such as `ambWith`, `concatWith`, `mergeWith`, `zipWith` allowing chained usage, rather than requiring a new expression with static usage. 

In other words, if `a.concatWith(b)` is done statically, it is `concat(a, b)`: https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Observable.java#L4083

However, `startWith` is *also* just an instance method for `concat`, just the order is reversed: https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Observable.java#L8022

So, I could see you going either way with the naming. It is odd though if multiple are chained, such as: `a.endWith(b).endWith(c)` vs `a.concatWith(b).concatWith(c)`. In practice though this is somewhat inefficient to do and should just be turned into `concat(a, b, c)` so only one operation instead of two has to occur.

So, even if `endWith` was added to the API, I would still want `concatWith` (the two are aliases of each other) for the common pattern of `zipWith`, `mergeWith`, `concatWith`, `ambWith`, etc.

jayphelps:
@benjchristensen super great points. What happens in RxJava if you chain starts? `a.startWith(b).startWith(c)`? isn't that just as odd as chaining ends?

jayphelps:
[...]

PR #3679, opened 2018-05-10
feat(endWith): add new operator endWith

natmegs:
endWith operator returns an Observable that emits given items last after items emitted by the source
Observable


Airblader:
#1709

benlesh:
Nice! Thanks @natmegs!  This one has been a hole in the API, especially in use cases around redux-observable, ngrx effects and animations. This will be well-used I expect.

Issue #3411, opened 2018-03-08
Test suite doesn't tsc type-check


IgorMinar:
The current test suite under `spec/` is being transpiled to JS and executed, but no type-checking is currently being enforced. As a result of that, lots of typing errors have been introduced into the test suite.

Repro:
```
yarn tsc -p spec/tsconfig.json
```

Expected: no type errors
Actual: lots and lots of errors

Once this problem is resolved, the tsc typecheck should be added to the CI suite so that we can avoid introduction of new errors in future commits.

kwonoj:
@david-driscoll is working on this. https://github.com/ReactiveX/rxjs/pull/3329 . I've accidentally closed https://github.com/ReactiveX/rxjs/issues/3020 .

kwonoj:
I think we can use this one until it's resolved, #3020 doesn't have additional detail specifically worth to reopen.

cartant:
I can't cope any longer with the test suite's not being type checked. I see that #3329 has been closed:

> ... in favor of a more incremental approach.

I'm going to start work on an incremental approach now. Unless anyone has objections, I plan on doing the following:

* Create a temporary `spec/tsconfig.check.json` file.
* Specify an `include` in said file so that files can be incrementally checked and fixed. E.g. `'include': ['spec/operators/a*.ts']`
* Create a temporary `script` named `test_check` (or whatever) in the `package.json` to run a no-emit  `tsc` build that uses `spec/tsconfig.check.json`.
* Include `npm test_check` in the `.travis.yml` configuration file so that the CI build runs it.
* Incrementally update the `-spec.ts` files so that they pass the check. That is, update them with individual PRs or, at least, in very small batches.
* When all `-spec.ts` files pass, remove the temporary scripts from the `package.json` and the `.travis.yml` files, delete the `spec/tsconfig.check.json` file and set the [`TS_NODE_TYPE_CHECK` environment variable](https://github.com/TypeStrong/ts-node/blob/v6.0.3/src/index.ts#L100) in the `test` script so that `ts-node` will throw an error if TypeScript errors are introduced into the tests.

Once the test suite is again type checked, I'd like to extend the type tests to include snippet-based testing, as  discussed [here](https://github.com/ReactiveX/rxjs/pull/3053#issuecomment-342885852) and as done [here](https://github.com/cartant/react-streams/blob/automate-test-args/src/index.test.ts).

david-driscoll:
[...]

PR #3680, opened 2018-05-10
chore(test): start fixing test types

cartant:
**Description:**

At the moment, the `test` script runs the specs using mocha and uses `ts-node/register` to transpile the TypeScript. However, the configuration does not enable type checking and there are many, many errors.

This PR adds some temporary changes to the infrastructure to facilitate the incremental fixing of these type errors. Specifically it:

* Creates a temporary `spec/tsconfig.check.json` file.
* Specifies an `include` in said file so that files can be incrementally checked and fixed.
* Creates a temporary `script` named `test_check` in the `package.json` to run a no-emit  `tsc` build that uses `spec/tsconfig.check.json`.
* Includes `npm run test_check` in the `.travis.yml` configuration file so that the CI build runs it.

With these changes it's possible to incrementally fix specific spec files - by adding them to the `includes` property in the `spec/tsconfig.check.json` file. This PR includes fixes for several files. However, fixes are still outstanding for all of the spec files in:

* `spec/observables`
* `spec/operators`
* `spec/schedulers`
* `spec/subjects`

The intention is for PRs to include individual commits for specific files. For example, add `./operators/buffer-spec.ts` to the `include` property, fix any type errors, then include those fixes in a commit.

When all `-spec.ts` files pass, the temporary scripts can be removed from the `package.json` and the `.travis.yml` files and the `spec/tsconfig.check.json` file can be deleted.

At that stage, the [`TS_NODE_TYPE_CHECK` environment variable](https://github.com/TypeStrong/ts-node/blob/v6.0.3/src/index.ts#L100) can be added to the `test` script so that `ts-node` will throw an error if a TypeScript error is introduced into the tests.

**Related issue (if exists):** #3411


benlesh:
Thanks, @cartant! This is a solid add.

PR #3682, opened 2018-05-10
refactor(delayWhen-spec): update delayWhen operator tests to use pipeable operators

natmegs:
update delayWhen spec to use updated import sites and pipeable operators


benlesh:
@Airblader, that's great, but I think it would be better if we did this in two steps. I'll merge this one, then you'll need to rebase yours, is that okay?

benlesh:
thanks again, @natmegs!

Issue #3685, opened 2018-05-11
skipUntil - v6.1.0 - throws an exception 'unsubscribe' of undefined


claylaut:
**RxJS version:** 6.1.0 / 6.0.0 (tested on both)

**Code to reproduce:**
https://stackblitz.com/edit/rxjs-skipuntil-issue

**Expected behaviour:**
it should work without throwing an exception

**Actual behaviour:**
Throwing an exception:
Error in rxjs@6.1.0/internal/operators/skipUntil.js (60:32)
Cannot read property 'unsubscribe' of undefined

**Additional information:**
This issue is related to this change:
https://github.com/ReactiveX/rxjs/pull/3512/files#diff-fb87e1b2c0661f48fca4bf43fb5a19a3R60

**PR including Fix**
https://github.com/ReactiveX/rxjs/pull/3686

This works perfectly in 5.5.10 and lower versions.

Stephanemw:
We're also seeing this using a BehaviorSubject that initially contains undefined (for deferred population).

```typescript
deferredBehaviorSubject = new BehaviorSubject<any>(undefined);
// ... 
obs.pipe(skipUntil(this.deferredBehaviorSubject))
````

Hacking skipUntilSubscriber to add a check works around this, but I don't know enough about rxjs internals to judge if that's the proper approach:
````typescript
if (this.innerSubscription) {
  this.innerSubscription.unsubscribe();
}
````

PR #3686, opened 2018-05-11
fix(skipUntil): fix skipUntil when innerSubscription is null

claylaut:
**Description:** Fix a minor issue with skipUntil

**Related issue:**
https://github.com/ReactiveX/rxjs/issues/3685

### DIFF
--- a/src/internal/operators/skipUntil.ts
+++ b/src/internal/operators/skipUntil.ts
@@ -57,7 +57,7 @@ class SkipUntilSubscriber<T, R> extends OuterSubscriber<T, R> {
              outerIndex: number, innerIndex: number,
              innerSub: InnerSubscriber<T, R>): void {
     this.hasValue = true;
-    if(this.innerSubscription) {
+    if (this.innerSubscription) {
       this.innerSubscription.unsubscribe();
     }
   }
--- a/src/internal/operators/skipUntil.ts
+++ b/src/internal/operators/skipUntil.ts
@@ -57,7 +57,9 @@ class SkipUntilSubscriber<T, R> extends OuterSubscriber<T, R> {
              outerIndex: number, innerIndex: number,
              innerSub: InnerSubscriber<T, R>): void {
     this.hasValue = true;
-    this.innerSubscription.unsubscribe();
+    if(this.innerSubscription) {
+      this.innerSubscription.unsubscribe();
+    }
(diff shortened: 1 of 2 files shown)


cartant:
Your fix looks appropriate, to me, but, ideally, the PR should also include a test that's based on code that reproduces the error.

The problem is effected because the implementation doesn't account for the synchronous completion of the notifier. You don't need to use a `BehaviorSubject` to effect the error; a simple, synchronous observable will do:

```ts
import { of } from "rxjs";
import { skipUntil } from "rxjs/operators";

of(1).pipe(
  skipUntil(of(2))
```

claylaut:
@cartant I've just simplified my usage example and opened an issue. 
I've tested the fix by modifying the node modules and it worked fine. Unfortunately, without this fix, I have to revert to 5.5.10.

I m unsure how to do the test if you can guide me I will appreciate it.

claylaut:
@cartant thanks a lot for your help. 
I've updated the tests and they passed successfully.
PR Updated.

src/internal/operators/skipUntil.ts
```diff
@@ -57,7 +57,9 @@ class SkipUntilSubscriber<T, R> extends OuterSubscriber<T, R> {
              outerIndex: number, innerIndex: number,
              innerSub: InnerSubscriber<T, R>): void {
     this.hasValue = true;
-    this.innerSubscription.unsubscribe();
+    if(this.innerSubscription) {
```
martinsik: This is just nitpicking but in all code around here there's always space between `if` and the starting parenthesis `if (this.innerSubscription)`.

claylaut: fixed

cartant:
@claylaut The repository's maintainers will eventually get around to reviewing and, likely, merging your PR. You will just need to be a little patient, as they're busy people with other priorities. In the interim, you could use a scheduler for a temporary workaround:

```ts
import { asapScheduler, of } from "rxjs";
import { skipUntil } from "rxjs/operators";

of(1).pipe(
  skipUntil(of(2, asapScheduler))
```

benlesh:
Thanks, @claylaut!

claylaut:
@benlesh thanks for merge.

@cartant thanks for the workaround 👍

Issue #3687, opened 2018-05-11
isObservable(null) throws an exception


demenskiy:
**RxJS version:** 6.1.0

**Code to reproduce:**

```ts
import {isObservable} from 'rxjs';

console.log(isObservable(null));
```

**Expected behavior:**

```js
false
```

**Actual behavior:**

```
Error: Cannot read property 'lift' of null
```

bbonnet:
@demenskiy https://github.com/ReactiveX/rxjs/pull/3688 should fix it!

PR #3688, opened 2018-05-11
fix(isObservable): Fix throwing error when testing isObservable(null)

bbonnet:
**Description:**
Fix `isObservable` to test to make sure the test subject is a function or object before attempting to access properties on it.

https://github.com/ReactiveX/rxjs/issues/3687

### DIFF
--- a/src/internal/util/isObservable.ts
+++ b/src/internal/util/isObservable.ts
@@ -1,10 +1,13 @@
 import { Observable } from '../Observable';
 import { ObservableInput } from '../types';
+import { isObject  } from './isObject';
+import { isFunction } from './isFunction';
 
 /**
  * Tests to see if the object is an RxJS {@link Observable}
  * @param obj the object to test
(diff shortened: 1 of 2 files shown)


bbonnet:
Above gave suggestion gave me `AssertionError: expected null to be false` for the `isObservable(null)` case so added `!!obj` for coercion to boolean

src/internal/util/isObservable.ts
```diff
@@ -1,10 +1,12 @@
 import { Observable } from '../Observable';
 import { ObservableInput } from '../types';
+import { isObject  } from './isObject';
+import { isFunction } from './isFunction';
```
demenskiy: Remove unused imports.

bbonnet: Good catch, updated.

benlesh:
thanks @bbonnet!

PR #3690, opened 2018-05-11
refactor(dematerialize-spec): update dematerialize operator tests to use pipeable operators

natmegs:
update dematerialize spec to use updated import sites and pipeable operators

### DIFF
--- a/spec/operators/dematerialize-spec.ts
+++ b/spec/operators/dematerialize-spec.ts
@@ -4,6 +4,8 @@ import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/mar
 
 declare function asDiagram(arg: string): Function;
 
+const NO_VALUES: { [key: string]: Notification<any> } = {};
+
 /** @test {dematerialize} */
 describe('dematerialize operator', () => {
   asDiagram('dematerialize')('should dematerialize an Observable', () => {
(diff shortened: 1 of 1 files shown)


natmegs:
Would love to get some input on a typing issue I ran into while refactoring these tests. For hot observables that never emit a value, their default type is `HotObservable<string>` (same for cold observables). So when an observable is setup in a test (ex: `e1 = hot(‘-----’)`), and I want to test the dematerialize operator with it using `e1.pipe(dematerialize())` (which expects a Notification type as its source), there is a type error. 
I got around this here by adding `hot(‘-----’, { a: Notification.createNext('a') } )`, but is there a more elegant way to indicate the type of the emissions of a hot/cold testing observable? 
I tried `hot(‘-----’) as HotObservable<Notification<any>>` but got the error `Type 'HotObservable<string>' cannot be converted to type 'HotObservable<Notification<any>>'`
I also tried `hot(‘-----’) as Observable<Notification<any>>` (since I saw that HotObservable is no longer exported directly), which resulted in the same error, and additionally expectSubscriptions tests break because there’s no subscriptions property on regular Observables.

cartant:
Typically, you'd specify an explicit type parameter, like this:

```ts
hot<Notification<any>>('---', {})
```

But that'd mess up the alignments and would also introduce a fair bit of clutter. I'd create a constant at the top of the file:

```ts
const NO_VALUES: { [key: string]: Notification<any> } = {};
```

and would use that in the tests:

```ts
const e1 =   hot('------', NO_VALUES);
const e1subs =   '^';
const expected = '-';
```

natmegs:
@cartant That's perfect - thank you!!

benlesh:
Perfect, thanks, @natmegs (and for the assist @cartant)

Issue #3692, opened 2018-05-12
ReferenceError: WebSocket is not defined (6.1.0)


krismrp:
**RxJS version:** 6.1.0

**Code to reproduce:**
```javascript
const { webSocket } = require("rxjs/websocket");
const w3cwebsocket = require("websocket").w3cwebsocket;

webSocket({
  url: 'ws://foobar',
  WebSocketCtor: w3cwebsocket
});
```

**Expected behavior:**
Should not error when given a WebSocket constructor.

**Actual behavior:**
```
            config.WebSocketCtor = WebSocket;
                                   ^

ReferenceError: WebSocket is not defined
    at new WebSocketSubject (E:\CODE\Github\[REMOVED]\node_modules\rxjs\internal\observable\dom\WebSocketSubject.js:49:36)
    at webSocket (E:\CODE\Github\[REMOVED]\node_modules\rxjs\internal\observable\dom\webSocket.js:43:12)
    at Object.<anonymous> (E:\CODE\Github\[REMOVED]\index.js:4:1)
    at Module._compile (module.js:624:30)
    at Object.Module._extensions..js (module.js:635:10)
    at Module.load (module.js:545:32)
    at tryModuleLoad (module.js:508:12)
    at Function.Module._load (module.js:500:3)
    at Function.Module.runMain (module.js:665:10)
    at startup (bootstrap_node.js:201:16)
```

waitingsong:
try
`const websocket = require('rxjs/websocket');`

krismrp:
I still get the same error message, can you confirm that the above code works for you?

waitingsong:
yes. it works .

krismrp:
Hm, still the same issue. Didn't have any problems with 5.x version.

bbonnet:
I think that should fix the issue, can someone help verify? https://github.com/ReactiveX/rxjs/pull/3694

krismrp:
@bbonnet I'll gladly do that, but how exactly? I was able to get install your pull request with Yarn, but everything went downhill from there. 😕

bbonnet:
@krismrp Did you get a specific error when everything went downhill? lol I think you should just be able to pin to that branch and pull it down into your app but I'm not 100% sure

PR #3694, opened 2018-05-14
fix(WebSocketSubject): Check if WebSocket exists before using it

bbonnet:
**Description:**
Check to make sure that the `WebSocket` constructor is in the root before using it.  This is currently breaking in Node.

https://github.com/ReactiveX/rxjs/issues/3692

### DIFF
--- a/src/internal/observable/dom/WebSocketSubject.ts
+++ b/src/internal/observable/dom/WebSocketSubject.ts
@@ -80,7 +80,6 @@ export class WebSocketSubject<T> extends AnonymousSubject<T> {
       this.source = urlConfigOrSource as Observable<T>;
     } else {
       const config = this._config = { ...DEFAULT_WEBSOCKET_CONFIG };
-      config.WebSocketCtor = WebSocket;
       this._output = new Subject<T>();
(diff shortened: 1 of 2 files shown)


src/internal/observable/dom/WebSocketSubject.ts
```diff
@@ -7,6 +7,7 @@ import { ReplaySubject } from '../../ReplaySubject';
 import { Observer, NextObserver } from '../../types';
 import { tryCatch } from '../../util/tryCatch';
 import { errorObject } from '../../util/errorObject';
+import { root } from '../../util/root';
```
benlesh: We want to get away from using `root`.

bbonnet: @benlesh Updated this to remove `root` and just see if `WebSocket` exists in the global scope.  Is there a specific reason we're moving away from using `root`?

benlesh: It causes weird issues with some bundling setups, and it's just unnecessary code for the most part.

benlesh:
thanks, @bbonnet!

Issue #3684, opened 2018-05-10
Documentation is actively misleading after release of v6


b-gran:
All of the documentation seems to indicate that operators are available on the prototype of Observables. There are dozens or hundreds of examples that all use this model. Here's an example [chosen at random](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-concatAll). This example uses operators on the prototype.

With the release of v6, it looks like all operators are now static functions, which breaks **all of the official documentation and practically every third-party tutorial**, of which there are thousands.

To make matters more confusing, the migration guide doesn't state this very consequential change in straightforward language. In order to figure out what was going on, I had to write some test code and manually examine the prototype of an Observable.

This documentation/API mismatch is serious enough to drive away many new users. Imagine: you read some blog post about why `rxjs` is great, but none of the code works and there isn't even a straightforward migration to make the example work in v6.

#### Proposed actions

At minimum, we should document in the `MIGRATION` file the consequences of the changes to operators, in clear and highly visible language.

We should also be thinking about how the community will be affected by this. Google searches for "do X in `rxjs`" will now turn up non-working code that can't be easily migrated in all cases. 

I've also opened a PR that makes the changed more prominent: #3700

ltoshev:
Completely agree. I had spent hours trying to figure out something, where it just pops up to old documentation and samples! Complete madness!

mxhllnn:
This is my first week using rxjs. I appreciate all of the hard work that has gone into the 6.0 release. However, I am now switching to baconjs because I am losing too much time trying to translate the 5.x documentation to the 6.0 version. It's strange to find a major version, especially one involving significant revision of the API, that has been released without updated documentation. I hope this doesn't sound like piling on. I just want to emphasize that docs are an essential part of the user experience.

benlesh:
The rxjs documentation is in a bit of flux at the moment. We have a new documentation site we're trying to publish this week. Everything on reactivex.io is not to be trust WRT RxJS>

benlesh:
Okay, so for now, I have http://reactivex.io/rxjs redirecting to the newer docs (with an HTML redirect unfortunately, until we have the new domain setup)

Hopefully this will help limit confusion.

PR #3700, opened 2018-05-15
Highlight important changes in migration script

b-gran:
**Description:**
Makes the change to chainable operators very prominent by adding a new section with bright colors to the top of the file.

Links to the more detailed explanation of the change.

**Related issue:**
Discuseed in #3684


benlesh:
While I like the addition of the diff section (I didn't know GIthub markdown supported that!), I disagree with the placement, the verbiage and the actual example code given here.

RxJS 6 is moving to pipeable operators. The rxjs-compat library keeps from breaking code targeted at older versions.  

I like the spirit of this PR, but I'm not sure about the content.  We don't really recommend `import operators as * from 'rxjs/operators'`, for example.

It would have been better to discuss where the migration doc is failing and how to improve it, rather than just wing it in a PR, I think.

Thank you for the time and thought, @b-gran, but I don't think we can accept this PR at this time.

b-gran:
Thanks for the response @benlesh 

I have an open issue to discuss the `MIGRATION` file and this change here: #3684 

My intention with this PR was exactly what you described: to discuss how to improve the `MIGRATION` doc. Is there a better way to have that discussion?

Appreciate that there are probably issues with the initial version of this PR (and in fact I never imagined this would be merged as-is), but would it be possible to leave this open while we continue the discussion on the linked issue? That way, anyone who finds the issue can also comment on the proposed changes to the `MIGRATION` file.

PR #3704, opened 2018-05-17
 fix(TestScheduler): restore run changes upon error

cartant:
**Description:**

This PR adds a failing test and a `finally` block to the `TestScheduler`'s `run` method so that `runMode` changes are restored upon an error being thrown by a test.

Without the `finally` block, the `TestScheduler` is left in `runMode` and that breaks tests that depend upon the previous, deprecated behaviour. This is a problem when converting a suite of tests to use the new behaviour: a failing `runMode` test will cause all subsequent non-`runMode` tests to fail with useless messages, as their marble diagrams will be interpreted incorrectly.

/cc @jayphelps


benlesh:
Thanks @cartant! Good catch!

PR #3708, opened 2018-05-18
chore(imports): import only once from `./types`

mprobst:
Somewhat confusingly, `Observable` used to import from `./types` once via the
child path, and once via `../internal/types`, which does resolve to the same file.


benlesh:
If I had to guess, I'd say that one of those was magically added by VSCode while I was working. :\

Thanks, @mprobst!

PR #3715, opened 2018-05-21
chore(test): fix operator test types a-d

cartant:
**Description:**

Continues from PR #3680 and fixes operators tests for operators starting with `a` through to `d`.

**Related issue (if exists):** #3411 #3680


benlesh:
Great! Thanks, @cartant

Issue #3699, opened 2018-05-15
Typing of isNumeric function is incorrect


AsamK:
The `isNumeric` function is implemented as a typeguard that returns true for `value is number`.
The problem is that the function returns true also if the `value` is a string, that can be parsed to a number.
The typescript type `number` means strictly a number, not a parseable string.

https://github.com/ReactiveX/rxjs/blob/92dd3d135d3fcf7fa98526a2ba0493754407a94a/src/internal/util/isNumeric.ts

**RxJS version:** 6.1.0

**Code to reproduce:**
```
const a = '10'
isNumeric(a) === true
typeof a === 'string'

if (isNumeric(a)) {
    const b: number = a
}
```

**Expected behavior:**
ts compiler should complain, because `a` (`'10'`) is not assignable to `number`

**Actual behavior:**
ts compiler accepts the code, because `isNumeric` makes an incorrect type guarantee.

**Additional information:**

benlesh:
This is a bug in the typings, however, `isNumeric` is internal to the library and not meant for public consumption, really.

dkosasih:
@benlesh  I am trying to have a look at this.
is this as simple as putting a check on like `typeof value !== 'string'` ?

cartant:
I don't have a problem with it being fixed, as it's incorrectly typed. And I'd favour just changing the type within the type predicate and leaving it as a user-defined type guard. So if you want to submit a PR, sure.

But, as Ben has stated, it's an internal function, so if you are changing it to address an issue in your code base, you should stop importing the function.

dkosasih:
No I didn't. Just try to be useful to the community and try to pick something simple to start with

AsamK:
@dkosasih thanks for fixing this :+1:

PR #3721, opened 2018-05-21
fix(isNumeric): handle numeric string case

akoserwal:
**Description:**
Added a string check before , to eliminate numeric string cases.
 #3699


benlesh:
Hi, @akoserwal! This is an api for internal use only. Is there something breaking in the external API that this is fixing?

akoserwal:
@benlesh: It is not fixing anything in the external API. I find many people are using this internal API and discussing about this particular case which I had covered here.

akoserwal:
@benlesh : In that case, I can close this PR. Although it is meant to be used internally. I have seen some discussion where it is been used as external consumption by some other projects. It's optional, if we want to handle string input.

PR #3724, opened 2018-05-22
refactor(distinct-spec): update distinct operator tests to use pipeable operators

natmegs:
update distinct spec to use updated import sites and pipeable operators


benlesh:
Thanks again, @natmegs!

PR #3730, opened 2018-05-22
refactor(distinctUntilChanged-spec): update distinctUntilChanged operator tests to use pipeable operators

natmegs:
update distinctUntilChanged spec to use updated import sites and pipeable operators


benlesh:
Thanks, @natmegs!

PR #3734, opened 2018-05-23
fix(typings): allow bufferCreationInterval null for bufferTime

### LINKED ISSUE #3728: bufferTime typings do not accept bufferCreationInterval null
**RxJS version:**
6.1.0

**Code to reproduce:**

```javascript
import {from} from 'rxjs'
import {bufferTime} from 'rxjs/operators'

from([1,2,3]).pipe(bufferTime(1000, null, 2)).subscribe(n => console.log(n))
```
**Expected behavior:**
Typescript code should compile the code correctly **with strictNullChecks**.

**Actual behavior:**
```
index.ts(4,37): error TS2345: Argument of type 'null' is not assignable to parameter of type 'number'.
```

**Additional information:**

The tests for bufferTime in spec/operators/bufferTime-spec.ts test that it works to send in null as bufferCreationInterval (https://github.com/ReactiveX/rxjs/blob/master/spec/operators/bufferTime-spec.ts#L45), but the typings for bufferTime specifies that bufferCreationInterval is a number only. The reason that your tests compile is that you have strictNullChecks set to false in tsconfig.json.

mattiash:
**Description:**

The tests for bufferTime verifies that it is possible to set
bufferCreationInterval to null, but the typings did not allow it
when strictNullChecks were set to true.

Fixes #3728

### DIFF
--- a/compat/operator/bufferTime.ts
+++ b/compat/operator/bufferTime.ts
@@ -4,8 +4,8 @@ import { bufferTime as higherOrder } from 'rxjs/operators';
 
 /* tslint:disable:max-line-length */
 export function bufferTime<T>(this: Observable<T>, bufferTimeSpan: number, scheduler?: SchedulerLike): Observable<T[]>;
-export function bufferTime<T>(this: Observable<T>, bufferTimeSpan: number, bufferCreationInterval: number | null, scheduler?: SchedulerLike): Observable<T[]>;
-export function bufferTime<T>(this: Observable<T>, bufferTimeSpan: number, bufferCreationInterval: number | null, maxBufferSize: number, scheduler?: SchedulerLike): Observable<T[]>;
+export function bufferTime<T>(this: Observable<T>, bufferTimeSpan: number, bufferCreationInterval: number | null | undefined, scheduler?: SchedulerLike): Observable<T[]>;
+export function bufferTime<T>(this: Observable<T>, bufferTimeSpan: number, bufferCreationInterval: number | null | undefined, maxBufferSize: number, scheduler?: SchedulerLike): Observable<T[]>;
 /* tslint:enable:max-line-length */
 
 /**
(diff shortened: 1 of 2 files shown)


cartant:
My two cents: Is this consistent with other parts of the API? Are there other places where `null` is used rather than `undefined`? I can't recall an API method to which `null` is passed to indicate that a parameter is not specified in a call; I can only recall those to which `undefined` is passed.

For example, `undefined` is used to to `subscribe`. This is okay:

```ts
of(1).subscribe(undefined);
```

Whereas this is not:

```ts
of(1).subscribe(null); // [ts] Argument of type 'null' is not assignable to parameter of type '((value: number) => void) | undefined'.
```

For consistency, I think is should be `undefined`, rather than `null`. Or all of the typings should be changed to accept either `null` or `undefined` for parameters that are to be ignored.

mattiash:
The tests pass in null for bufferCreationInterval: https://github.com/ReactiveX/rxjs/blob/master/spec/operators/bufferTime-spec.ts#L24

The code sets bufferCreationInterval to null if the parameter is not included:
https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/bufferTime.ts#L67

and it later checks if it is null:
https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/bufferTime.ts#L129

Changing it to undefined is a much bigger change, but I have nothing against it.

cartant:
> Changing it to undefined is a much bigger change ...

Yeah, you are right. It would be a bigger change and smaller changes are always more appealing.

Regarding the check against `null`, the check [uses `==`](https://github.com/ReactiveX/rxjs/blob/6.2.0/src/internal/operators/bufferTime.ts#L129) which will return `true` for either `undefined` or `null`. That's the mechanism used throughout the Angular codebase, IIRC.

Anyway, others can decide, I just thought I'd point out the consistency issue.

Thanks for the PR, BTW. :+1:

benlesh:
Thanks, @mattiash!

PR #3750, opened 2018-05-26
chore(test): update operator tests d-m to use pipeable operators

natmegs:
update operators with names starting with d-m inclusive to use pipeable operators and updated import sites

### DIFF
--- a/spec/operators/min-spec.ts
+++ b/spec/operators/min-spec.ts
@@ -1,19 +1,18 @@
 import { expect } from 'chai';
-import * as Rx from 'rxjs/Rx';
+import { min, skip, take, mergeMap } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


cartant:
@natmegs I've had a quick look at the PR. There are some conflicts, as some of my changes to the specs - to fix TypeScript errors - appear to have been merged after you've made your changes, so you'll need to rebase against `master`.

When you do that, you will likely see errors regarding some of the `Notification`-related changes. `Notification` takes a type parameter and your changes have removed it. It will need to be added back.

natmegs:
@cartant Good catch, thanks for taking a look! I've rebased and checked for tests where Notification was being used to make sure there's the type parameter (materialize-spec). If you notice that I've missed anything please let me know :)

Issue #3753, opened 2018-05-28
The `rxjs` package is bloated


sindresorhus:
**RxJS version:** 6.2.0

The package is more than 6 MB. That is a lot for a utility library.

Why is `src` included in the package? Source maps and TS definition files are already included for each JS file, so I don't see the need to include the TS source files too.

There's also a lot of bloat related to browser support: `bundles`, `_esm5`, `_esm2015` (3.8 MB in total). Would you consider publishing a separate package for Node.js usage without all the bloat? For example, `rxjs-node`.

![screen shot 2018-05-28 at 11 40 13](https://user-images.githubusercontent.com/170270/40597474-e6bb81d8-626b-11e8-935d-438eb5d3c847.png)

benlesh:
I agree. :man_shrugging: 

IIRC, `src` is included because of source map needs. I'm curious as to why you say it's not necessary. The build process is something I've been a little detached from recently.

Most of the bloat is us trying to support everyone's module styles needs without confusing the crap out of users. If we publish `rxjs` and `rxjs-node`, then why not `rxjs-esm` and `rxjs-umd` and so on? And then all of the examples publicly will be weird and confusing, some will `import { fromEvent } from 'rxjs-node';` and others will `improt { fromEvent } from 'rxjs';` 

I'm open to any and all proposed solutions (especially from people that want to help). I know @kwonoj would kill to get this smaller, because they use it with an Electron app over at SlackHQ, and they're including the whole package.

kwonoj:
just fyi around Electron side, I made decisions around gave up to deal with node_modules size / loading time and moving toward to bundling (even in main nodejs process, yes) to dodge several cases we needed to handcraft some solutions per each cases.

csvn:
@benlesh I suggest to set `removeComments` to `true` in `tsconfig.base.json`. Having comments in `d.ts` files is enough for e.g. VSCode to show tooltips and more. There is a lot of comments that are used to generate documentation, and removing them from `.`, `_es5` and `es2015` would reduce total size by a lot.

benlesh:
FWIW: There's an opportunity to try to change this for v7, as what I'm looking at doing there is currently in our [experimental branch](https://github.com/ReactiveX/rxjs/tree/experimental)

kwonoj:
@csvn's proposal is actually good point, we may able to have as non-breaking changes?

csvn:
[...]

PR #3760, opened 2018-05-29
perf: remove comments from js-files

csvn:
**Description:**

Strip comments from all `.js` files (except `legacy-reexports` and `compat`) while retaining them for `.d.ts` files. The `legacy-reexport` does not have any comments and will be removed with `compat` in next major version.

```
npx get-folder-size --folder=dist/package
// 5.88 Mb (with comments)
// 4.63 Mb (without comments)
```

Roughly 1.25 Mb is saved by removing comments.

**Related issue (if exists):** #3753


csvn:
If you are wondering why `build_types` is slightly out of order in `build_all`, it's because it has to come before `build_legacy_reexports`, since the latter uses `./dist/typings`.

benlesh:
LGTM... @cartant second opinion?

benlesh:
Thanks @csvn!

PR #3768, opened 2018-05-31
update jsdoc examples (audit* through distinct*)

moniuch:
**Description:**
Update operators' (audit* through distinct*) jsdoc examples to use pipe()

### DIFF
--- a/src/internal/operators/audit.ts
+++ b/src/internal/operators/audit.ts
@@ -31,8 +31,8 @@ import { subscribeToResult } from '../util/subscribeToResult';
  * repeats for the next source value.
  *
  * @example <caption>Emit clicks at a rate of at most one click per second</caption>
- * var clicks = Rx.Observable.fromEvent(document, 'click');
- * var result = clicks.audit(ev => Rx.Observable.interval(1000));
+ * const clicks = fromEvent(document, 'click');
+ * const result = clicks.pipe(audit(ev => Rx.Observable.interval(1000)));
  * result.subscribe(x => console.log(x));
  *
  * @see {@link auditTime}
--- a/src/internal/operators/auditTime.ts
+++ b/src/internal/operators/auditTime.ts
@@ -25,8 +25,8 @@ import { MonoTypeOperatorFunction, SchedulerLike } from '../types';
  * Optionally takes a {@link IScheduler} for managing timers.
  *
  * @example <caption>Emit clicks at a rate of at most one click per second</caption>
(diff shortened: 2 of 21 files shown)


cartant:
Thanks, @moniuch. I will have a look at this soon.

I noticed that you closed #3767. FYI, if the only reason for that was to add more commits, you could have just pushed more commits (or changed commits) to your branch using `git push -f` and that would have updated the PR.

moniuch:
Hi @cartant 
Thanks, actaully I also wanted the branch name to correspond with the contents so I decided to make a new one instead.
Not sure whether you want so granular commits in PRs.
I will try to push more updates tomorrow after you have looked through this PR and after we have agreed on the format.

moniuch:
Cool, thanks for your patience.

I have noticed today purely accidentally that jsdoc need update in operator files, but I don't know the project structure well enough to tell if this is all we need to update. I will also be happy to update other docs if you point me to a specific folder.

cartant:
Regarding other examples that need to be updated to v6, some of the factory functions in the `src/internal/observable` directory also include examples that use `pipe` and v5 globals like `Rx.Observable.of`. For example, [`empty.ts`](https://github.com/ReactiveX/rxjs/blob/master/src/internal/observable/empty.ts#L23-L38).

moniuch:
Regarding trailing commas: should I remove them also in the object arrays used throughout examples?

cartant:
It's up to you, I guess. Personally, I'd prefer it if they weren't there, but, equally, I prefer changes to be as small as possible. I mentioned the trailing commas within `pipe`, as they were introduced in these changes and I think the examples are easier to read without them. This is, however, a close-to-trivial issue and it isn't really a big deal whether they are there or not.

benlesh:
> I think the trailing commas that are included - in some of the examples - after the last operator passed to pipe, should be removed.

I disagree... for one reason: Google uses clang format, and if you want your pipe operators to look all pretty on their own lines, you need to add a trailing comma so clang format formats it that way.

:)

Also it helps for version control if you add another operation to the end of your pipe.

This is :+1: from me, @moniuch, thanks for this PR!! Docs are _so_ important right now.

moniuch:
Thanks Ben,

Hanging commas are something I use on an everyday basis and I really like them for two good reasons: (1) meaningful git diffs when lines get added/removed
(2) easier managable lines (adding, duplicating, deleting)
(3) visual consistency.

So - I am glad we have agreed on this - this is how I will proceed from now on with the remaining examples.

PR #3771, opened 2018-06-01
ci(travis): expand matrix to test multiple tsc

kwonoj:
**Description:**
This PR expands our test matrix setup as previously discussed - in addition to current test setups, it builds against `typescript@latest` and `typescript@next` always to detect forward-breaking changes early. Those builds are specified as `allow_failures`, so does not block as long as default job passes.

example: 
image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,8 @@ addons:
   sauce_connect:
     # Update SAUCE_USERNAME / SAUCE_ACCESS_KEY in travis if necessary
     # https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings
-    sauce_connect: true
+    # TODO: reenable saucelabs once we have update browser tests
+    # sauce_connect: true
 
 cache:
   timeout: 1000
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -17,23 +18,38 @@ env:
...
   - npm install -g yarn
-  - if [ "$FULL_VALIDATE" == "true" ]; then npm install --no-save grunt@0.4.1 grunt-cli grunt-contrib-connect grunt-run; fi
 
 install:
   - npm ci
   - if [ "$FULL_VALIDATE" == "true" ]; then npm run lint && npm run test:circular; fi
+  - if [ "${TSC_VERSION}" ]; then npm install --no-save typescript@$TSC_VERSION && ./node_modules/.bin/tsc --version && npm run build_cjs ; fi
 
 script:
+  - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
```
kwonoj: `$TRAVIS_BRANCH` points head of PR (most cases `master`), so https://github.com/ReactiveX/rxjs/pull/3771/files#diff-354f30a63fb0907d4ad57269548329e3L46 tends to be triggered for each PR.

cartant:
When I've used yarn with Travis, I've not installed it explicitly. My understanding is that Travis will install it automatically if the project contains a `yarn.lock` file.

kwonoj:
That's not our case. We do install yarn currently (https://github.com/ReactiveX/rxjs/pull/3771/files#diff-354f30a63fb0907d4ad57269548329e3R44), and this PR fails by yarn try to install some deps while my PR doesn't touches it.

kwonoj:
Question is more about: `why we do use` _yarn_ in mixture between npm?

cartant:
I have no idea why yarn would need to be installed if npm is being used. That doesn't seem to make much sense.

cartant:
The docs app seems to use it: https://github.com/ReactiveX/rxjs/blob/10823c9cf4f101190362fe66eaf04df298e1bb86/docs_app/aio-builds-setup/scripts/test.sh

kwonoj:
Just raised q in doc channel to get clarification. This PR is in no hurry so I'll wait if any other PR fails for same reason, if so will take out to unblock PR contributions.

benlesh:
I personally don't care if we switch to yarn or not. Whatever works 🤷‍♂️

kwonoj:
Me too, I don't mind if it doesn't break PR.

cartant:
FWIW, I switched to using yarn largely because of the issues that npm had with optional, OS-dependent dependencies (which, I believe, have since been fixed).

Yesterday, I was talking with another dev about some article he'd sent me about devs moving back to npm from yarn. With yarn working for me, I'm in no hurry to go back, but what came out of our discussion - well, my part was more of a rant - is that neither is perfect.

Moving to yarn might or might not be totally painless. The problems that I've had with yarn have had to do with:

* re-installing or updating the global yarn sometimes (always?) sees any globally installed packages lost.
* without adding some extra bits to the `.travis.yml` file, yarn didn't play nice with Greenkeeper (but they had documentation about what needed to be done to make it work).

Travis installs/uses its own global yarn if it sees a `yarn.lock` file, so the first problem isn't really relevant for Travis. Can be annoying locally though, but globally installed packages should probably be avoided anyway.

kwonoj:
> I switched to using yarn largely because of the issues that npm had with optional, OS-dependent dependencies (which, I believe, have since been fixed).

for this, it's no longer an issue from npm 5 (our team uses npm 5 resolves those issues)

My most strongest objection around yarn is only limited in this particular repo - we were having single source pkg mgr until now and now it's 2, which is quite confusing for anyone who'll try to clone repo. If we were using yarn and second was npm, I'd raise similar opinions. Also things get quite changed recently afaik, but yarn on windows was pretty much painful to deal with. 

Anyway, both are not huge blocker as long as it works - if it doesn't break PR, I don't hardly object on it. Still surprised we just made this changes without any discussions.

cartant:
> My most strongest objection around yarn is only limited in this particular repo - we were having single source pkg mgr until now and now it's 2, which is quite confusing

I agree completely.

JWO719:
Sorry for adding yarn as dependency for the docs. It came from the angular docs. They use it for package management. Our first goal was to integrate it into this repository to be able to stick to the current development of rxjs. Changing the tool chain was more or less a minor issue. But you are totally right it is confusing. It was on my todo list to remove yarn as dependency too. If you need any kind of help with this, just ping me.

benlesh:
Well, recently on the Angular team we had some sort of issue with Yarn because npm changed how it was hosting some packages, and it messed up yarn (Or some such thing, I didn't follow it closely). There was just an issue with our CI because of how yarn was resolving a karma package for a day or so. It seems like a one-off, I'm sure yarn is fine.

FWIW, I think yarn did a more than adequate job forcing npm to get better than it was. I feel like I still prefer npm a little, but that's mostly because I'm used to it.

I really, really don't care which one we use, but we should be consistent.

We can take a vote in the rxjs-core channel in Slack or something.

benlesh:
FYI: @JWO719 @kwonoj ... (OJ knows this already) but the core team took a vote and we decided to stay with NPM. So all traces of Yarn should be removed. We can do that in a separate PR if need be.

JWO719:
Ok Thanks for informing. So how to proceed with this yarn stuff. I can take care of it. Or is there someone else who wants to do that?

kwonoj:
Go for it in new PR, probably most work is replace yarn to npm and remove checkers around.

JWO719:
Yeah I think that too. I'll link the pr when I am ready.

PR #3774, opened 2018-06-01
docs: update jsdoc examples (operators e through z)

moniuch:
**Description:** Continuation of #3768 - jsdoc examples: new pipe syntax (remaining operators)
**Related issue (if exists):** #3768

Noticed and not sure about:
- observeOn - `observeOn(Rx.Scheduler.animationFrame),` - is `animationFrame` a creation method, just like `fromEvent`? never used it before.
- some operators are still missing their examples (eg. `endsWith`, `repeat`, `race`)
- do/tap operator - is the do.png link still valid?
- share.ts: `.multicast(() => new Subject()).refCount()` - should this be changed to:
  `.pipe(multicast(() => new Subject()), refCount())` ?
- withLatestFrom - uses chaining + result selector in the description. The description might require a complete makeover though: * `a.withLatestFrom(b, c, (a1, b1, c1) => a1 + b1 + c1)`).

Code style:
- callbacks - single untyped parameters might do without parens around it - example would gain on readability where there is a lot of parens floating around (eg. `groupBy`)
- some examples (eg. `groupBy`) use Finnish notation for streams. Even though it is my personal preference as well in apps, I think it is not necessary in short examples. Unless there is a deeper reason behind that I am unaware of.

### DIFF
--- a/src/internal/operators/elementAt.ts
+++ b/src/internal/operators/elementAt.ts
@@ -25,8 +25,8 @@ import { take } from './take';
  * ## Example
  * Emit only the third click event
  * ```javascript
- * var clicks = Rx.Observable.fromEvent(document, 'click');
- * var result = clicks.elementAt(2);
+ * const clicks = fromEvent(document, 'click');
+ * const result = clicks.pipe(elementAt(2));
  * result.subscribe(x => console.log(x));
  *
  * // Results in:
--- a/src/internal/operators/every.ts
+++ b/src/internal/operators/every.ts
@@ -9,9 +9,10 @@ import { Observer, OperatorFunction } from '../types';
  * ## Example
  * A simple example emitting true if all elements are less than 5, false otherwise
  * ```javascript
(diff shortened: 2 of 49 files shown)


moniuch:
>> is the do.png link still valid?
> Yep. It seems to be.
Well, it is not broken, but the png describes the old `do`. Saw a PR on generating the docs - will ask the author to perhaps touch that problem as well.

> Sure, seems reasonable to me. If you're changing the dot-chaining examples to use pipe and see dot chaining elsewhere in the docs, change them to use pipe.

OK, I changes those to pipeable syntax. Wasn't sure what is the current status of result selectors right now, but as long as they appear as @params they should be documented correctly.

benlesh:
@moniuch so sorry, there was a conflict with a recent change by @JWO719 I just merged... can you rebase?

moniuch:
@benlesh checks in progress, I hope I didn't screw anything up during rebase ;-)

benlesh:
Thanks, @moniuch  :rocket:

PR #3777, opened 2018-06-01
build(docs): remove docs app from published node modules

JWO719:
Due to the publish.sh the docs-app shouldn't published within the rxjs node module. But to ensure that I think it is a good idea to add the docs-app folder to the .npmignore file.


benlesh:
Awesome, thanks!

PR #3781, opened 2018-06-02
docs(content): add migration guide

JWO719:
This pr moves the migration.md file to the docs app. Additionally I added a "redirect" for the old file in for already existing links to this file. The md file will be rendered inside of the docs app and looks like this:

image


benlesh:
Awesome!

JigneshPatel04:
I have used
import { Observable ,of,throwError } from 'rxjs';
import { flatMap,catchError } from 'rxjs/operators';

I got errors 
Property 'of' does not exist on type 'typeof Observable'.
Property 'flatMap' does not exist on type 'Observable<Response>'.

my example is

       return this.http.request(url_, options_).flatMap((response_ : any) => {
            return this.getTimezones(response_);
        }).catch((response_: any) => {
            if (response_ instanceof Response) {
                try {
                    return this.GetTimezones(<any>response_);
                } catch (e) {
                    return <Observable<ListResultDto>><any>Observable.throw(e);
                }
            } else
                return <Observable<ListResultDto>><any>Observable.throw(response_);
        });

PR #3782, opened 2018-06-03
docs(contributing): add guideline not to overwrite commit history

moniuch:
**Description:**
Based on a conversation with @cartant during my first PR, it turned out that squashing commits containing the requested changes - even though it has the purpose of making the git history clean - makes the github review flow unclear. The original commits appear with annotation of being "outdated", and the PR author is not confident whether the comments they see pertain to the original or updated commits.

My understanding then is that the practice should be to append new commits, thus keeping the history of the feature branch linear and the flow clear. Therefore, I am updating the guidelines in CONTRIBUTING.md with an appropriate info.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,6 +80,8 @@ Before you submit your Pull Request (PR) consider the following guidelines:
     git push -f
     ```
 
+  * When updating your feature branch with the requested changes, please do not overwrite the commit history, but rather contain the changes in new commits. This is for the sake of a clearer and easier review process.
+
 That's it! Thank you for your contribution!
 
 


benlesh:
This makes sense. We do, however, often "Squash and Merge" changes as they're merged into master. That's where we keep the commit history as clean as possible.

benlesh:
Thanks @moniuch

moniuch:
@benlesh Well, now I can see that what I wrote in .md file is not quite accurate. During a review-changes-review-changes drill we don't want to overwrite history, but it is actually unavoidable when PR needs a rebase.

I actually contacted github regarding the flow and whether they plan to show individual "updates" (= result of `git push -f`) just the way TFS does it. They had no plans for it but took a note of it. And now that they are Microsoft - who knows..  ;-)

PR #3783, opened 2018-06-04
docs(pipable-operators.md): update for RxJS 6

martinsik:
**Description:**

I've noticed some people on StackOverflow referring to this document and being surprised why it doesn't work with RxJS 6 (mostly when importing "creation" methods).

### DIFF
--- a/doc/pipeable-operators.md
+++ b/doc/pipeable-operators.md
@@ -1,8 +1,8 @@
 # Pipeable Operators
 
-Starting in version 5.5 we have shipped "pipeable operators", which can be accessed in `rxjs/operators` (notice the pluralized "operators"). These are meant to be a better approach for pulling in just the operators you need than the "patch" operators found in `rxjs/add/operator/*`.
+Starting in version 5.5 we have shipped "pipeable operators", which can be accessed in `rxjs/operators` (notice the pluralized "operators"). These are meant to be a better approach for pulling in just the operators you need than the "patch" operators found in `rxjs-compat` package.
 
-**NOTE**: Using `rxjs/operators` without making changes to your build process can result in larger bundles. See [Known Issues](#known-issues) section below.
+**NOTE**: Using `rxjs` or `rxjs/operators` without making changes to your build process can result in larger bundles. See [Known Issues](#known-issues) section below.
 
 **Renamed Operators**
 
@@ -23,16 +23,6 @@ The former `toPromise()` "operator" has been removed
 because an operator returns an `Observable`,
 not a `Promise`.
 There is now an `Observable.toPromise()`instance method.
-
-Because `throw` is a key word you could use `_throw` after `import { _throw } from 'rxjs/observable/throw'`.
-
-If the leading `_` bothers you (because a leading `_` typically means "_Internal - Do not use_"), you can do as follows:
-```
-import { ErrorObservable } from 'rxjs/observable/ErrorObservable';
-...
-const e = ErrorObservable.create(new Error('My bad'));
-const e2 = new ErrorObservable(new Error('My bad too'));
-```
 
 ## Why?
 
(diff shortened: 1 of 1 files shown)


doc/pipeable-operators.md
```diff
@@ -24,16 +24,6 @@ because an operator returns an `Observable`,
...
 
-Because `throw` is a key word you could use `_throw` after `import { _throw } from 'rxjs/observable/throw'`.
-
-If the leading `_` bothers you (because a leading `_` typically means "_Internal - Do not use_"), you can do as follows:
-```
-import { ErrorObservable } from 'rxjs/observable/ErrorObservable';
-...
-const e = ErrorObservable.create(new Error('My bad'));
-const e2 = new ErrorObservable(new Error('My bad too'));
-```
```
martinsik: I think this whole part isn't necessary since in RxJS 6 you import one of these:
```
import { catchError } from 'rxjs/operators';
import { throwError } from 'rxjs';
```

JWO719:
Hey, great idea to update the pipeable-operator docs, but would you like to update the new docs? There is a folder docs_app. This one is used for the new docs (see: https://rxjs-dev.firebaseapp.com/). I would really like to see that becoming part of the new documentation. If you have any question to that, don't hesitate to ask me.

doc/pipeable-operators.md
```diff
@@ -265,7 +255,7 @@ import { map, filter, reduce } from 'rxjs/operators';
 becomes:
 
 ```ts
-import { map } from 'rxjs/operators/map';
-import { filter } from 'rxjs/operators/filter';
-import { reduce } from 'rxjs/operators/reduce';
+import { map } from 'rxjs/internal/operators/map';
+import { filter } from 'rxjs/internal/operators/filter';
+import { reduce } from 'rxjs/internal/operators/reduce';
```
benlesh: These should all be imported from `rxjs/operators` ... importing anything under `rxjs/internal/**` is a no-no.

martinsik: I think that's intentional because the paragraph above it says:

> #### No Control over Build Process
> If you have no control over your build process (or are unable to upgrade to Webpack 3+), the above solution will not work. Therefore importing from rxjs/operators will likely make your application bundle larger. However, there's still a way you can use pipeable operators. You will have to use deep imports, similar to how you import prior to version 5.5 and pipeable operators:

benlesh: Hmmm... I'm not sure about this. People may want to use the `rxjs-compat` library instead. I'd hate to give people the idea that using anything under `internal` was anything but a pathway to upgrade pain whenever v7 comes up. :thinking:

jayphelps: I'm pretty against recommending people reach into `rxjs/internal/` as it should be used as a signal that the API is not public and subject to change without notice. If we want to expose the individual operators as paths it should be in a non-internal path.

I do sympathize with the desire to import these directly for people who don't have access to tree shaking, e.g. react-native folks don't get shaking yet if you use Metro bundler. 

> People may want to use the rxjs-compat library instead. 

I don't believe rxjs-compat exposes pipeable operators with individual paths, does it?

benlesh: > I don't believe rxjs-compat exposes pipeable operators through individual paths, does it?

It does, in fact... it basically re-enables imports from RxJS 5 locations:  `rxjs/operators/map` for example.

benlesh: @jayphelps is absolutely right, anything under internal can change without notice. So we can't recommend anyone uses that stuff.

benlesh:
I guess, for now at least, we can remove documentation about importing them individually. There are a few ways, but we don't want people to import from locations under `internal`.

@martinsik can you do that, please?

doc/pipeable-operators.md
jayphelps: Hmm oh sorry I recalled that rxjs/operators/map wasn’t pipeable but expected `this`. I’ll look again when I get a sec. 👍👍

martinsik:
@benlesh Ok, I removed that paragraph.

benlesh:
@JWO719 ... can we land this, then just move the file to the new docs app?

benlesh:
Thanks, @martinsik!

PR #3784, opened 2018-06-04
fix(docs): fix code examples

JWO719:
Dgeni is actually not capable to transform @example jsdoc notation. For this purpose I changed all examples to the md notation. For results see attached images. The only exception is generate.ts. the function overloading makes the example looks kinda weird. Will make a seperate pull request for generate.ts

windowwhen_before
windowwhen_after


benlesh:
Awesome! Thanks @JWO719

PR #3785, opened 2018-06-04
refactor(multicast-spec): update multicast operator tests to use pipeable operators

natmegs:
Update multicast spec to use updated import sites and pipeable operators.

Ran into a typing issue in this test. Not sure if it would be considered a separate issue (or an issue at all really?) but would appreciate some input. With the following example:
```
const expected = [1, 2, 3, 4];

const connectable = of(1, 2, 3, 4).pipe(multicast((new Subject<number>())));

connectable.subscribe((x) => { expect(x).to.equal(expected.shift()); },
   (x) => {
    }, () => {
      done();
    });

connectable.connect();
```
I get a type error on the last line: `[ts] Property 'connect' does not exist on type 'Observable<number>'`
This looks to be because the `pipe` function return value is typed as `Observable<number>` when in this case we really need it to be `ConnectableObservable<number>`

To get around this I added 
```
const connectable = of(1, 2, 3, 4).pipe(multicast((new Subject<number>()))) as ConnectableObservable<number>;
```
but this is verbose and also cheats the typings test at the end of the file (line 708) where it should infer the type.

### DIFF
--- a/spec/operators/multicast-spec.ts
+++ b/spec/operators/multicast-spec.ts
@@ -1,17 +1,16 @@
 import { expect } from 'chai';
-import * as Rx from 'rxjs/Rx';
+import { multicast, tap, mergeMapTo, takeLast, mergeMap, refCount, retry, repeat, switchMap, map } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


benlesh:
Thanks again, @natmegs!

PR #3787, opened 2018-06-05
fix(from): Changed the priority of observable detection

### LINKED ISSUE #3779: Observable.from prioritizes array over observable symbol
## Bug Report

**Current Behavior**
When an object that is "array like" and has Observable Symbol, Observable.from uses `array like` Observable.from mechanism before the Observable Symbol interop. 

This is a problem in Microstates because the Observable Symbol provides the proper mechanism to make a Microstate observable. Treating it like an array does something very funky.

**Reproduction**
- REPL or Repo link:
https://runkit.com/taras/creating-an-observable-from-array-not-working

```js
const assert = require('assert');
const { from } = require('rxjs');
const Microstates = require('microstates').default;

let ms = Microstates.from([{hello: 'world'}]);

from(ms).subscribe(next => assert(next[0], 'Should have an array with one item'))
```

**Expected behavior**
I would expect the Observable Symbol interop to be used before any other. 

**Environment**
- Runtime: [Node & Browsers]
- RxJS version:
- 6.1

**Possible Solution**
Check for Observable Symbol interop before all other options.

Related https://github.com/microstates/microstates.js/issues/130

taras:
**Description:**

This PR ensures that `Observable.from` uses Observable Symbol interop mechanism before attempting any others. 

Closes #3779

### DIFF
--- a/spec/observables/from-spec.ts
+++ b/spec/observables/from-spec.ts
@@ -59,6 +59,21 @@ describe('from', () => {
     })
   });
 
+  const fakeArrayObservable = <T>(...values: T[]) => {
+    let arr = ['bad array!'];
+    arr[Symbol.observable] = () =>  {
+      return {
+        subscribe: (observer: Observer<T>) => {
+          for (const value of values) {
+            observer.next(value);
+          }
+          observer.complete();
+        }
+      };
+    };
+    return arr;
+  };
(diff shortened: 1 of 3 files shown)


benlesh:
Thanks @taras!

Issue #3766, opened 2018-05-31
Mixed types in pipes with mergeMap and more than 8 parameters yield typescript error


akehir:
In the below code, foo1() is ok, but foo2() is not. In other words, in pipes with more than 8 parameters in the stream, the type definition assumes that all the observables of the stream return the same types.

**RxJS version:**
```6.2.0```

**Code to reproduce:**
Stackblitz: https://stackblitz.com/edit/typescript-e8dcm5?file=index.ts

```typescript
import {
  of,
  Observable,
} from 'rxjs';

import {
  tap,
  mergeMap,
} from 'rxjs/operators';

export class Foo {

  public foo1(): Observable<boolean> {
    return of(true).pipe(
      tap(), // 1
      tap(), // 2
      tap(), // 3
      tap(), // 4
      tap(), // 5
      tap(), // 6
      tap(), // 7
      mergeMap(() => {
        return of('');
      }),
      mergeMap(() => {
        return of(true);
      }),
    );
  }

  public foo2(): Observable<boolean> {
    return of(true).pipe(
      tap(), // 1
      tap(), // 2
      tap(), // 3
      tap(), // 4
      tap(), // 5
      tap(), // 6
      tap(), // 7
      tap(), // 8
      mergeMap(() => {
        return of('');
      }),
      mergeMap(() => {
        return of(true);
      }),
    );
  }
}
```

**Expected behavior:**
No typescript compilation error, as mergeMap should be able to change the type of the Observable, and only the last mergeMap should be relevant for the resulting type.

**Actual behavior:**
```
ERROR in index.ts(41,7): error TS2345: Argument of type 'OperatorFunction<boolean, string>' is not assignable to parameter of type 'OperatorFunction<boolean, boolean>'.
  Type 'string' is not assignable to type 'boolean'.
```

**Additional information:**
The issue seems to be with Observable.ts, line 292, and mergeMap.ts, line 13:
```typescript
pipe<R>(...operations: OperatorFunction<T, R>[]): Observable<R>;
```
```typescript
export function mergeMap<T, R>(project: (value: T, index: number) => ObservableInput<R>, concurrent?: number): OperatorFunction<T, R>;
```
With more than 8 parameters, apparently all the OperatorFunction<T, R> need to return the same type <R>.

cartant:
That particular signature for `pipe` is incorrect. It most likely should be:

```ts
pipe<R>(...operations: OperatorFunction<any, any>[]): Observable<R>;
```

[...]

PR #3789, opened 2018-06-05
fix(Observable): Update typings of pipe

akehir:
**Description:**
Added <any, any>[] as tyings for the generic type of operators provided to the pipe, as
OperatorFunctions can transform the type.

The fix is as suggested by @cartant  here: https://github.com/ReactiveX/rxjs/issues/3766#issuecomment-393447462

Travis CI passes for typescript current, but not for next (not related to my change): https://travis-ci.org/akehir/rxjs/builds/388179712

#3766


benlesh:
LGTM... Thanks, @akehir!

PR #3794, opened 2018-06-05
docs(jsdoc): add examples for race()

moniuch:
**Description:**
Add two jsdoc examples for `race()` factory method.


JWO719:
Yeah you're probably right, but I liked the first one too :D

PR #3795, opened 2018-06-05
docs(jsdoc): add examples for endWith operator

moniuch:
**Description:**
Add jsdoc examples for `endWith` operator.

### DIFF
--- a/src/internal/operators/endWith.ts
+++ b/src/internal/operators/endWith.ts
@@ -21,6 +21,44 @@ export function endWith<T>(...array: Array<T | SchedulerLike>): MonoTypeOperator
  * Returns an Observable that emits the items you specify as arguments after it finishes emitting
  * items emitted by the source Observable.
  *
+ * ## Examples
+ * ### Emits a value before completing (appends an emission to a series of emissions).
+ *
+ * ```javascript
+ * from(['hi', 'how are you?', 'sorry, I have to go now']).pipe(
+ *   endWith('goodbye!'),
+ * )
+ * .subscribe(word => console.log(word));
+ * // result:
+ * // 'hi'
+ * // 'how are you?'
+ * // 'sorry, I have to go now'
+ * // 'goodbye!'
+ * ```
+ *
+ * ### Emits a value before completing (with no prior emissions)
+ * ```javascript
+ * empty().pipe(
+ *   endWith('goodbye!'),
+ * )
+ * .subscribe(word => console.log(word));
+ * // result:
+ * // 'goodbye!'
+ * ```
+ *
+ * ### Does not work with abnormal termination
+ * ```javascript
+ * throwError('any error').pipe(
+ *   endWith('goodbye!'),
+ * )
+ * .subscribe(
+ *   word => console.log(word),
+ *   err => console.log('sorry, I choked'),
+ * );
+ * // result:
+ * // 'sorry, I choked' - endWith won't emit values when an error is thrown
+ * ```
+ *
  * @param {...T} values - Items you want the modified Observable to emit last.
  * @param {Scheduler} [scheduler] - A {@link IScheduler} to use for scheduling
  * the emissions of the `next` notifications.
--- a/src/internal/operators/endWith.ts
+++ b/src/internal/operators/endWith.ts
@@ -22,10 +22,10 @@ export function endWith<T>(...array: Array<T | SchedulerLike>): MonoTypeOperator
  * items emitted by the source Observable.
  *
  * ## Examples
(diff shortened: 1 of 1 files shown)


moniuch:
What do we do when adding comments decreases the coverage?

JWO719:
Hi @moniuch . I think it isn't a problem with your code. Maybe the tool had some problems. One of my prs had exactly the same problem. I'll take a look at it this afternoon, maybe I find something. Don't panic :D

moniuch:
Just `KEEP CALM AND CONTRIBUTE`  :)

@JWO719 Would you mind if I asked you for help?... After I `npm install`-ed, I want to make sure everything works ok locally for me. Do we have any "take-off" checklist to run? I really want to contribute with some real code now (first operator, even if it is ends up not being accepted in PR).

I feel like this kind of checklist could make it to CONTRIBUTING.md.

JWO719:
Of course you can ask me anytime for help :D But I actually don't understand what you are looking for?

JWO719:
Ok now I get it. IMO it makes sense to add stuff like this to the CONTRIBUTING.md. So in general I would suggest running all tests makes sense due to possible "side effects". And in general tests should be executed with `npm run test`. Did that helped you? But anyhow updating the CONTIBUTING.md by making a proposal is definetly a good idea.

moniuch:
Thanks, will give it a shot and see how it goes. The main point is - how do I test my new operator locally in a playground, just like I can test published release on stackblitz? Is there a nice flow for that?
I sure can create a ticket for updating CONTRIBUTING.md, but I don't think I would be the right person to take it :)

JWO719:
maybe @benlesh or @cartant have some tipps for you

cartant:
@moniuch If you are looking at writing an operator, you should start with a user-land operator. That is, you should write an operator that is not part of the RxJS package.

Some information from OJ:

> Our general recommendation is if implementation is trivial / and doesn't require modification in core, we do recommend for user-land first. We want to gather usecases around how much common it is before we include into core, since adding new api surface is easy but deprecating / changing behavior is exceptionally hard, as well as we don't want to core api surface it too broad (it is alreay, and we are trying to shrink it down).

Regarding a playground, it's up to you. All you really need is the UMD bundle and a browser (see the operator in [this Stack Overflow answer](https://stackoverflow.com/a/50654079/6680611).

It should be pretty easy to use either StackBlitz or CodeSandbox (via the vanilla starting point) to get an in-browser, TypeScript environment happening. Just import RxJS and get started.

moniuch:
@cartant Thanks for the info. So, by userland you mean just any type of fiddle, not any specific platform. Would this be ok? https://stackblitz.com/edit/ts-rxjs-some-userland?file=index.ts

cartant:
@moniuch A user-land operator is just an operator that's separate from the RxJS package.

For example, my [`rxjs-etc`](https://github.com/cartant/rxjs-etc) package has a bunch of operators in it. They are user-land operators, as they're not part of RxJS proper.

moniuch:
@cartant Oh, you got an entire repo of those! Nice, thanks - this gives me a much better idea of how to go about it.

src/internal/operators/endWith.ts
```diff
@@ -21,6 +21,44 @@ export function endWith<T>(...array: Array<T | SchedulerLike>): MonoTypeOperator
  * Returns an Observable that emits the items you specify as arguments after it finishes emitting
  * items emitted by the source Observable.
  *
+ * ## Examples
+ * ### Emits a value before completing (appends an emission to a series of emissions).
+ *
+ * ```javascript
+ * from(['hi', 'how are you?', 'sorry, I have to go now']).pipe(
```
benlesh: I think use `of` here instead of `from`, makes the code a *little* simpler.

moniuch:
@benlesh Agreed - this is exactly what I did for `ignoreElements` - somehow didn't think of applying similar changes to `endWith`. Fixed.

src/internal/operators/endWith.ts
```diff
@@ -21,11 +21,11 @@ export function endWith<T>(...array: Array<T | SchedulerLike>): MonoTypeOperator
  * Returns an Observable that emits the items you specify as arguments after it finishes emitting
  * items emitted by the source Observable.
  *
- * ## Examples
- * ### Emits a value before completing (appends an emission to a series of emissions).
+ * ## Example
+ * ### After the source observable completes, appends an emission and then completes too.
```
benlesh: one or more emissions, actually...  but the most common use case is one. :)

benlesh:
Thanks, @moniuch

PR #3804, opened 2018-06-06
Don't comment on locked issues

rgbkrk:
If you set up lock bot to _only_ lock, it won't make notifications on old threads. It does mean that users won't get told why the issue is locked though, so it's a bit of a tradeoff. For ye old issue reporters and PR makers (like myself), I want to get notifications on relevant threads I'm in rather than knowing about every locked one.

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -10,10 +10,7 @@ exemptLabels: []
 lockLabel: false
 
 # Comment to post before locking. Set to `false` to disable
-lockComment: >
-  This thread has been automatically locked since there has not been
-  any recent activity after it was closed. Please open a new issue for
-  related bugs.
+lockComment: false
 
 # Limit to only `issues` or `pulls`
 # only: issues


rgbkrk:
> commit message does not follows conventional change log

ya caught me bottie

rgbkrk:
Rebased to appease the robot overlords.

jayphelps:
ummm holy cow yes please.

BIG YES

***

If someone is worried perhaps we could add support to lockbot to edit the OP's comment at the top with the declaimer, assuming lockbot can have permissions to do that like admins of a project can? Seems possible: https://developer.github.com/v3/issues/comments/#edit-a-comment

jayphelps:
For clarify, @benlesh could you add your thoughts around why lockbot was added? I'm guessin gyou prefer people don't comment on old stale issues and instead create new ones?

jayphelps:
Makes sense 🙂

benlesh:
I get way too many comments on old, dead issues. I'd rather people file new issues and maybe reference the old one if it's relevant.

I see the point about the lockbot notifications, though... they're just as annoying. I think I can turn those comments off. ... but then people won't know why it's locked, I guess. But most people are smart, they'll figure it out.

benlesh:
Thanks, @rgbkrk

rgbkrk:
Thank you! Now back to my regularly scheduled RxJS lurking.

PR #3806, opened 2018-06-06
migration.md: Typo

felixrabe:
**Description:**


benlesh:
@JWO719 it's okay in this case, not a critical fix we need to tell people about in the changelog, really.

benlesh:
Thanks, @felixrabe!

PR #3808, opened 2018-06-06
docs(teams): add alumn entry and add nicholas jamieson to core team

JWO719:
as requested by @ladyleet .
@cartant please check if the deets are correct

image


ladyleet:
Thanks @JWO719 and @cartant and @benlesh

PR #3810, opened 2018-06-07
docs(contributors): add contributors entry on team page

JWO719:


benlesh:
I'm okay with this as long as @natmegs, @cedricss, and @jasonaden are okay with it. :+1:

benlesh:
To be clear, these are "community members"? We might need to reoganize how we do this. I'm with @ladyleet, I like what Vue has done: https://vuejs.org/v2/guide/team.html

Sorta flat with a few key people at the top, but it includes community members and people that are good to follow.

I think this PR is a good start, however.

benlesh:
I'm super excited to give people who support us so well some additional attention, though! So this PR is great.

cedricss:
I'm okay with this❣️

PR #3821, opened 2018-06-10
fix(fromEvent): Support React Native and node-compatible event sources.

### LINKED ISSUE #3809: Support React Native EventEmitter type with fromEvent
## Feature Request
React Native defines its own EventEmitter type that it claims is a subset of the Node.js EventEmitter type, but it is a bit different: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native/index.d.ts#L143

Specifically, `removeListener` returns `void`. The return type of `addListener` and parameters are also different, but this doesn't cause a typing problem.

**Is your feature request related to a problem? Please describe.**
If you attempt to use `fromEvent` with a React Native EventEmitter implementation such as https://github.com/urbanairship/react-native-module you will get a type error:

```ts
fromEvent(UrbanAirship, 'pushReceived');

Argument of type 'typeof UrbanAirship' is not assignable to parameter of type 'FromEventTarget<{}>'.
  Type 'typeof UrbanAirship' is not assignable to type 'ArrayLike<EventTargetLike<{}>>'.
    Index signature is missing in type 'typeof UrbanAirship'.
```

There are at least two solutions:

1. Add a `ReactNativeEventEmitter` type.
2. Allow `addListener` and `removeListener` for `NodeStyleEventEmitter` to return `any`.

The `isNodeStyleEventEmitter` check can be used for both, but unfortunately I don't think there is a way to know whether the provided EventEmitter is Node.js style or React Native except based on type. However, the return types for `addListener` and `removeListener` are not used, so being able to distinguish between them at runtime won't matter at this time.

You can of course do `fromEvent(UrbanAirship as any...` or use `fromEventPattern`, but these are more verbose and you lose type safety you could get from the former.

ajcrites:
React Native implements a similar interface to Node.js event listeners -- specifically, its Event Emitters have addListener and removeListener methods. These are compatible with the Node interface, but they are not identical to the Node.js *style* since the addListener method returns a subscription rather than the event emitter itself, and the removeListener method returns nothing.

These return values are not actually used by fromEvent, so creating an Observable from an Event Emitter where addListener returns something different and removeListener returns nothing should still work to create a compatible Observable that emits with the corresponding events.

Closes #3809

### DIFF
--- a/spec/observables/fromEvent-spec.ts
+++ b/spec/observables/fromEvent-spec.ts
@@ -1,8 +1,11 @@
 import { expect } from 'chai';
 import { expectObservable } from '../helpers/marble-testing';
-import { fromEvent, NEVER, timer, pipe } from 'rxjs';
+import { Observable, fromEvent, NEVER, timer, pipe } from 'rxjs';
+import { NodeStyleEventEmitter, NodeCompatibleEventEmitter, NodeEventHandler } from 'rxjs/internal/observable/fromEvent';
 import { mapTo, take, concat } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
+
+declare const type: Function;
 
 declare function asDiagram(arg: string): Function;
 declare const rxTestScheduler: TestScheduler;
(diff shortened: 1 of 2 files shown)


benlesh:
Thanks, @ajcrites!

PR #3822, opened 2018-06-10
chore(test): enable TS_NODE_TYPE_CHECK

cartant:
**Description:**

This PR completes the changes to the test suite outlined in #3680.

The `test` script in `package.json` now sets the `TS_NODE_TYPE_CHECK` environment variable so that `ts-node` performs type checking.

/cc @kwonoj @darthtrevino I've also added a `test_no_type_check` script which is somewhat faster. Based on your comments in #3565, you might want to use it in your watched builds.

**Related issue (if exists):** #3411


kwonoj:
If we bump up ts-node to latest ( > 6.0) https://github.com/TypeStrong/ts-node/releases/tag/v6.0.0 it went back to enable type checking by default, does not require setting env variable.

cartant:
Upgraded to `ts-node` 6, renamed `test_no_type_check` to `test_transpile_only` and relied upon type checking being enabled by default.

PR #3832, opened 2018-06-12
fix(AjaxObservable): switch to encodeURIComponent to properly encode body form data

### LINKED ISSUE #3824: Ajax does not fully escape data in body object
## Bug Report

**Current Behavior**
A clear and concise description of the behavior.
Given I pass in a JS object to the body parameter in the rxjs ajax body containing the following string: "vudFamJHuC%oXe&jFA0AM1IzVpt2daZv".This string would be sent as "vudFamJHuC%25oXe&jFA0AM1IzVpt2daZv". In this case the percent sign has been URL encoded, but the ampersand has not. In the case of form data, this would split that item and result in the request not being as expected.

**Reproduction**
```TypeScript
import { ajax } from 'rxjs/observable/dom/ajax';
import { first } from 'rxjs/operators';

// Just send a test request
ajax({
    url: 'https://example.com',
    method: 'post',
    body: {
        test: 'vudFamJHuC%oXe&jFA0AM1IzVpt2daZv'
    },
    crossDomain: true,
    headers: {
            'Content-Type': 'application/x-www-form-urlencoded; charset=UTF-8'
        },
}).pipe(first()).subscribe(() => { return; });
```

**Expected behavior**
The 'test' value should be encoded like: 'vudFamJHuC%25oXe%26jFA0AM1IzVpt2daZv'

**Environment**
- Runtime: Chrome 67.0.3396.79 (Official Build) (64-bit)
- RxJS version: 5.5.10

NickGeek:
**Description:**
Switch to encodeURIComponent so characters like & and + are properly encoded. This fix is in RxJS 6
but is not in RxJS 5.

**Related issue:**
Fixes #3824

### DIFF
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -643,6 +643,40 @@ describe('Observable.ajax', () => {
       expect(complete).to.be.true;
     });
 
+    it('properly encode full URLs passed', () => {
+      const expected = { test: 'https://google.com/search?q=encodeURI+vs+encodeURIComponent' };
+      let result: Rx.AjaxResponse;
+      let complete = false;
+
+      Rx.Observable
+        .ajax.post('/flibbertyJibbet', expected)
+        .subscribe(x => {
+          result = x;
+        }, null, () => {
+          complete = true;
+        });
+
+      const request = MockXMLHttpRequest.mostRecent;
+
+      expect(request.method).to.equal('POST');
+      expect(request.url).to.equal('/flibbertyJibbet');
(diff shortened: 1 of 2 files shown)


NickGeek:
Hmm, they all pass [here](https://travis-ci.com/NickGeek/rxjs/builds/76004903)

benlesh:
@NickGeek this looks like an issue with the dangerfile stuff. @kwonoj

kwonoj:
rebase to TOT of stable (https://github.com/ReactiveX/rxjs/pull/3843) will resolve danger failure issue.

benlesh:
Thanks, @NickGeek

PR #3835, opened 2018-06-13
feat(operators): add average and sum operators

jafaircl:
`average` operator returns the average of values emitted by an Observable. Nested values can be accessed with a predicate function. `sum` operator returns the sum of values emitted by an Observable. Nested values cam be accessed with a predicate function. Both operators already exist in other ReactiveX libraries.

**Description:**


kwonoj:
before any other code changes, why this PR have huge changes in `package-lock` without any package changes?

jafaircl:
Honestly I didn’t see that the package-lock came in with it. I ran npm install to run the tests and it must’ve updated package-lock. I can close and submit another pull request if that’s easiest.

kwonoj:
you don't need to create new PR, just delete it and can force push commit.

benlesh:
A few things before this PR continues.

Both of these proposed operators will only function on `Observable<number>`, presumably? I didn't see any actual code files here. Honestly, this could be done just as easily with a third party library of reducer functions that weren't RxJS specific:  `source$.pipe(reduce(sum))` for example. 

Given that it's specific to `Observable<number>`, it's trivial to implement, and it might be better off in a more utilitarian library of common reducer functions, I'm not sure this is a worthy inclusion to RxJS core.

I'm interested in what my peers think, though. @kwonoj? @david-driscoll? @cartant?

jafaircl:
I noticed that the operator and spec files didn’t transfer over. I’ll add those as soon as I get back to my computer. You’re right that they’re mostly for `Observable<number>` but they can also take a predicate function. I wrote the operators separately then noticed they were in RxJava and RxPHP but not RxJS. No worries if you don’t want to have them in core. I was just reading the docs and noticed a hole that I had already filled.

kwonoj:
I admit I'm not in great favor of adding new one in general. 

> noticed they were in RxJava and RxPHP but not RxJS

This is obvious case I mentioned time to time, we don't carry over operator 1:1 between different language bindings.

Other judgement call I usually use are
- How hard will it be if chain existing operator? Is it impossible / too verbose?
- How hard will it be if new operator serves as user land module? any dependency to core implementation requires new operator should be in core?

cartant:
Given that `average` and `sum` were [removed in the port from RxJS v4 to v5](https://github.com/ReactiveX/rxjs/blob/stable/MIGRATION.md#operators-renamed-or-removed) my question is: what's changed that would warrant their being reinstated?

I'm not in favour of the reinstatement. I don't think anyone who needs to average or sum a stream of numbers would be hard pressed to implement the calculation.

Also, if they are to be merged, there is an implementation problem in `average` - there is internal state within the reducer that's not per-subscription.

jafaircl:
Ah I didn’t see any references in this repo to them ever being in the library so didn’t realize they had been removed. Sorry about that. I can close if everyone is opposed to bringing them back or leave it open if you’d like to discuss it further.

benlesh:
@jafaircl I think that would be best.  We haven't had much demand for this, and it's something that individuals can hand-roll pretty easily.

benlesh:
Thank you so much, @jafaircl, for your thoughtful contribution! I hope that the closing of this PR doesn't put you off from contributing again!

Issue #3844, opened 2018-06-17
Docs PWA Logo is wrong


JWO719:
I noticed that the PWA logo of the rxjs docs is wrong. It shows the angular logo. This should be fixed as soon as possible.     The splash screen is wrong as well.

martinsik:
I think you can close this one. It should be fixed in #3846.

Issue #3837, opened 2018-06-13
Docs fix migration guide 


JWO719:
The markdown file of the migration guide seems to be wrongly interpreted. We have to look if there is some error in the interpretation of if this is just a small syntactic typo.

JWO719:
closed by #3846

Issue #3818, opened 2018-06-08
Remove internals from API docs


JWO719:
Everything from 'rxjs/internal/' shouldn't be visible from the docs website. This should prevent usage of this package. But we shouldn't remove the jsdocs comment from the code. It just shouldn't be visible on the website.

JWO719:
closed by #3846

Issue #3817, opened 2018-06-08
Malformed "See Also" in Doc Reference


JWO719:
Sometimes the @see leads to a wrong formatted Doc Reference.

E.g. The docs of auditTime prints "@see audit @see debounceTime @see delay @see sampleTime @see throttleTime". But also sometimes for example with ArgumentOutOfRangeError, there it prints a seperate "See Also" chapter, how it is should be.

We have consistently create the "See Also" Chapter and find out what causes the error to be aware of that in future prs.

martinsik:
I think you can close this one. It should be fixed in #3846.

JWO719:
closed by #3846

PR #3846, opened 2018-06-18
docs(*): fix broken cross-reference doc links and missing marble diagrams

martinsik:
**Description:**

While doing https://github.com/ReactiveX/rxjs/issues/3805 I fixed a few other issues along the way:

- Running `npm run docs` under `docs_app` was throwing some `checkUnbalancedBackTicks` warnings because in some files there was odd number of backticks ` ``` `.
- `@nocollapse` were unknown to the `parseTagsProcessor ` so these were throwing warnings as well.
- Some exports from `index.ts` overrode each other on case insensitive filesystems (eg. OS X but I guess Windows would have the same problem as well). This means that eg. `EMPTY.json` and `empty.json` were both saved into the same file. The same applies to `Observable` class and `observable` symbol. This is why I moved `const`s into a subdir called `const`.
This doesn't affect the structure on `/api` page. The only difference is that there are eg. `/api/index/Observable` and `/api/index/const/observable` pages.
- When fixing doc links (`@link xyz`) I removed all references to internal APIs because these don't have docs pages anyway. For example I changed `{@link Action}` to only `Action` because `Action` isn't exported. The same goes for `create` that should be replaced with `new Observable`.  #3635 #3818.
- There were missing marble diagrams for all operators. I added `embedMarbleDiagramsPostProcessor` that copies them from `tmp/docs/img` into `generated/images/marbles`. This also means that `` tags don't need to specify that actual path which is added during post processing. For example this `` is now only this `` where the actual path, `alt` and `width` attributes are both added later. The `alt` attribute is filled automatically when empty #3179
- There was also some weird problem that `` tags in markdown when processed by dgeni made the rest of the docblock not parsed (there were dangling `@see` and `@param` at the end of the document). #3817
- Also added missing `endWith` marble diagram. #3796
- Updated `pwa-manifest.json` with new icons (generated them here `https://app-manifest.firebaseapp.com/`). #3844

#3179
#3635
#3796
#3805
#3817
#3818
#3837 - seems to be fine here.
#3844


docs_app/tools/transforms/angular-api-package/index.js
```diff
@@ -87,15 +90,15 @@ module.exports = new Package('angular-api', [basePackage, typeScriptPackage])
 
   // Additional jsdoc config (for RxJS source)
   .config(function(parseTagsProcessor) {
-    parseTagsProcessor.tagDefinitions.push({ name: 'internal' });
```
JWO719: Why did you removed the internal tag? I think this one is used to hide internal api which we will need in future?

martinsik: I though that when something is marked as `@internal` it shouldn't be part of the public API exported by neither `index.ts` nor `operators/index.ts` so if I remove this tag definition it's going to throw warnings  about an unknown tag when running `npm run docs` and thus telling me that I'm exporting something that's supposed to be internal.

JWO719: Ah ok good point. Let's stick to your approach. We have to think about it again when we start hiding the internal api.

petebacondarwin: We use the internal flag in Angular docs to indicate that a member of a public class is not to be exposed in the docs. Also `@internal` is used by some tooling (TypeScript??) I think. So unless you do not expect this tag to appear in your source at all, I would leave it in.

martinsik: I'm checking it again and the `@internal` tag is actually already included:

https://github.com/ReactiveX/rxjs/blob/master/docs_app/tools/transforms/angular-api-package/tag-defs/internal.js

https://github.com/ReactiveX/rxjs/blob/master/docs_app/tools/transforms/angular-api-package/index.js#L81-L86

So there no need to add it again with `parseTagsProcessor.tagDefinitions.push({ name: 'internal' });`

docs_app/tools/transforms/angular-api-package/index.js
```diff
@@ -162,17 +165,33 @@ module.exports = new Package('angular-api', [basePackage, typeScriptPackage])
       pathTemplate: '${moduleDoc.moduleFolder}/${name}',
       outputPathTemplate: '${moduleDoc.moduleFolder}/${name}.json',
     });
+    // Override path for symbols that would otherwise collide with other names (index/observable vs. index/Observable)
+    // on case insensitive filesystems (eg. OS X).
+    computePathsProcessor.pathTemplates.push({
+      docTypes: ['const'],
+      getPath: (doc) => {
```
JWO719: @petebacondarwin suggested to always use the type in the url and therefore be more generic. And to be honest I would really like this  :D

src/internal/operators/auditTime.ts
```diff
@@ -22,7 +22,7 @@ import { MonoTypeOperatorFunction, SchedulerLike } from '../types';
  * the time unit determined internally by the optional `scheduler`) has passed,
  * the timer is disabled, then the most recent source value is emitted on the
  * output Observable, and this process repeats for the next source value.
- * Optionally takes a {@link IScheduler} for managing timers.
```
benlesh: These can change to `SchedulerLike`, which is a public interface.

benlesh: (there's more throughout this commit where `Scheduler` could be `SchedulerLike`)

src/internal/observable/range.ts
```diff
@@ -25,7 +25,7 @@ import { Observable } from '../Observable';
  *
  * @param {number} [start=0] The value of the first integer in the sequence.
  * @param {number} [count=0] The number of sequential integers to generate.
- * @param {Scheduler} [scheduler] A `Scheduler` to use for scheduling
+ * @param {Scheduler} [scheduler] A {@link SchedulerLike} to use for scheduling
```
benlesh: The `{Scheduler}` here should be `{SchedulerLike}` . (same elsewhere in this commit)

src/internal/observable/combineLatest.ts
```diff
@@ -50,7 +50,7 @@ export function combineLatest<R>(...observables: Array<ObservableInput<any> | ((
  * computes a formula using the latest values from all the inputs, then emits
  * the output of that formula.</span>
  *
- * <img src="combineLatest.png">
+ * ![](combineLatest.png)
```
benlesh: Can we add text in the `[]` part of these? It should set the title of the image tag and help with a11y issues.  `combineLatest marble diagram`

src/internal/observable/range.ts
petebacondarwin: The types of the params should not be necessary. I believe that the dgeni processing will extract this info from the function declaration below. You should try it and see.

src/internal/observable/combineLatest.ts
martinsik: @benlesh The alt attributes are generated by the post-processor because I didn't want to copy&paste them everywhere so when there's no `alt` it'll be filled automatically here: 

https://github.com/martinsik/rxjs/blob/4731b753e6de81900e1f029c2a61d699b9ce6881/docs_app/tools/transforms/angular-api-package/post-processors/embedMarbleDiagrams.js#L34-L36

src/internal/operators/auditTime.ts
martinsik: I replaced all occurrences where I thought it makes sense. Sometimes it looked more like the text is referring to arguments so I left it with just `scheduler`.

src/internal/observable/range.ts
martinsik: It looks like when listing overloads it reads function declarations and prints proper `"Type: xyz"` but "Parameters" section at the top of the page usually doesn't mention any type. I'm not sure why is that.

benlesh:
Thank you so much, @martinsik!!! This was an awesome PR.

PR #3855, opened 2018-06-23
chore(build): Build a umd bundle of rxjs-compat

### LINKED ISSUE #3626: Minified compat build
Currently it is impossible for rxjs-compat to use the minified build of rxjs without the consumer taking on the responsibility of minifying the result.

In Nuclide, for example, we rely on using the minified build of rxjs as its impact on startup time is significant, but doesn't warrant minifying the rest of Nuclide.

There could be, for example, a build of rxjs-compat that bundles all of Rx and rxjs-compat in a single minified file.

If this is outside the scope of the project I understand. I'm happy to take a look at implementing this otherwise.

wbinnssmith:
Resolves #3626

This introduces a build step for creating a umd bundle (and minified umd bundle)  of rxjs-compat.

This involved creating a new tsconfig for esm5 for rollup (no helpers), a new rollup build script for producing the bundle, and a closure build script for producing the minified bundle. Each was derived from their rxjs counterpart.

This also creates a `umd.ts` entrypoint in `rxjs-compat` which exposes `testing`, `operators`, etc. just like the rxjs umd entry.

Tested by using this bundle in place of the rxjs 5 bundle in [Nuclide](https://github.com/facebook/nuclide/)

Tested the performance impact of the unminified bundle by loading it in <=20ms with `for i in {1..5}; do node -p "s=Date.now();require('./dist-compat/global/rxjs-compat.umd');Date.now()-s"; done`. The commonjs entrypoint to `rxjs-compat` loaded in 130ms by comparison. Note this improvement over the [rxjs 5 bundle](https://github.com/ReactiveX/rxjs/pull/2064) which loads in 26ms -- I verified I could reproduce nearly @zertosh's exact values on the hardware I tested 6 on.

cc @zertosh @matthewwithanm @captbaritone


wbinnssmith:
@jasonaden @benlesh anything I can do to help this move forward?

benlesh:
Thanks, @wbinnssmith

PR #3867, opened 2018-06-27
chore(test): udpate all tests in spec/observables

natmegs:
update all tests in spec/observables/ to use pipeable operators and updated import sites


benlesh:
Awesome! Thanks, @natmegs!

PR #3876, opened 2018-06-29
docs(throwIfEmpty): add missing marble diagram

martinsik:
**Description:**

I've noticed that the new `throwIfEmpty` operator doesn't have a marble diagram so this PR adds `asDiagram` test and also adds the image into its docblock.

This is what it looks like:
throwifempty


benlesh:
Awesome! Thanks, @martinsik!

PR #3881, opened 2018-07-02
docs(build): deploy docs to firebase on publish

JWO719:
This script should deploy the docs if the rxjs node module is published.
I am not completly sure if it is working completly because I couldn't test it :D
Mind that you have to enter the credentials for the firebase account each time.


benlesh:
We'll give it a try next publish :). Thanks, @JWO719!

Issue #3681, opened 2018-05-10
Document ISubscription to SubscriptionLike breaking change


closedLoop:
**RxJS version:**
6.1.0

**Code to reproduce:**

    import { ISubscription } from 'rxjs/Subscription';

**Expected behavior:**

    import { SubscriptionLike } from 'rxjs';

**Actual behavior:**

   No documentation for this change is mentioned in MIGRATION.md#import-paths

**Additional information:**
 
  Please include an update about this breaking change https://github.com/ReactiveX/rxjs/blob/master/MIGRATION.md#import-paths

BioPhoton:
@benlesh 
As `SubscriptionLike` is now mentioned in  https://github.com/ReactiveX/rxjs/blob/master/docs_app/content/guide/v6/migration.md#import-paths we can close this issue.

benlesh:
Hi @closedLoop — thank you for taking the time to open this issue and for sharing your thoughtful feedback with the RxJS project. I’m sorry it took us so long to respond. Given how much time has passed and the lack of recent activity, I’m going to close this for now so the tracker reflects the work we’re likely to act on. If this is still relevant on a current RxJS release, please feel welcome to open a new issue with updated context and a reproduction. Thank you again for your contribution and patience.

— Ben

PR #3884, opened 2018-07-03
docs(migration): mention util interfaces in migration import paths

martinsik:
**Description:**

I added mentions of `SubscriptionLike` and `PartialObserver` into "Import paths" migration guide: https://github.com/ReactiveX/rxjs/blob/master/docs_app/content/guide/v6/migration.md#import-paths

The first bullet point already says:

> 1. rxjs: Creation methods, types, schedulers and utilities

so I only added `SubscriptionLike` and `PartialObserver` because these two are most common util interfaces I think.

#3681


benlesh:
Makes sense, thanks @martinsik!

Issue #3898, opened 2018-07-05
TestScheduler can miss tests when flushing


TPReal:
## Bug Report

**Current Behavior**
Calling `expectObservable` adds tests to `flushTests` field, and calling `toBe` on the expect, marks those tests as ready. Then calling `flush` should go over `flushTests`, and execute and remove those that are ready at this point. The logic of removing is broken, and different tests get removed than should be. As a result, some tests might never be checked, and some are checked multiple times.

The problem is that `TestScheduler.flush` is broken. When removing items from `flushTests`, it uses the index of the item from before removal, so for example if there are 3 tests, `[ready1, ready2, notReady]`, then `flush()` will execute `ready1`, then remove at index 0, so `ready1`. Then execute `ready2`, then remove at index 1, but now it's `notReady` that is present at index 1. So after `flush()`, `flushTests` should be `[notReady]`, but in fact it is `[ready2]`.

**Reproduction**
https://stackblitz.com/edit/rxjs-testscheduler-flush-bug

```js
  expectObservable(cold('-x')).toBe('-x');
  expectObservable(cold('-y')).toBe('-y');
  const expectation=expectObservable(cold('-z')).toBe.bind(null,'-q');
  flush(); // Causes the check z vs q not to be checked at all.
  expectation();
```

**Expected behavior**
The expectation in the code above should be executed, and the test should fail, because `'z'` is not equal to `'q'`. The test of `'y'` should be executed once, while it is executed twice.

**Environment**
- RxJS version: 6.2.1

**Possible Solution**
Change the relevant part of `TestScheduler.flush` to:
```js
  this.flushTests = this.flushTests.filter(test => {
    if (test.ready) {
      this.assertDeepEqual(test.actual, test.expected);
      return false;
    }
    return true;
  });
```

**Additional context**
I'm testing a service in an Angular application, calling also a material dialog. For some reason, I need to call `flush()` at some point, between subscribing to an observable (by calling `expectObservable`) and actually comparing its result with the expectation.

ladyleet:
@cartant 👍 to close?

cartant:
@ladyleet This is the issue for https://github.com/ReactiveX/rxjs/pull/3901. It'll be closed automatically when that's merged. However, I guess there's no reason it couldn't be closed right now.

PR #3901, opened 2018-07-07
 fix(TestScheduler): flush expectations correctly

cartant:
**Description:**

Adds a failing test for the `TestScheduler` and applies the fix from #3898.

**Related issue (if exists):** #3898

### DIFF
--- a/spec/schedulers/TestScheduler-spec.ts
+++ b/spec/schedulers/TestScheduler-spec.ts
@@ -462,5 +462,18 @@ describe('TestScheduler', () => {
       expect(testScheduler['runMode']).to.equal(runMode);
       expect(AsyncScheduler.delegate).to.equal(delegate);
     });
+
+    it('should flush expectations correctly', () => {
+      expect(() => {
+        const testScheduler = new TestScheduler(assertDeepEquals);
+        testScheduler.run(({ cold, expectObservable, flush }) => {
+          expectObservable(cold('-x')).toBe('-x');
+          expectObservable(cold('-y')).toBe('-y');
+          const expectation = expectObservable(cold('-z'));
+          flush();
+          expectation.toBe('-q');
+        });
+      }).to.throw();
+    });
   });
 });
--- a/src/internal/testing/TestScheduler.ts
+++ b/src/internal/testing/TestScheduler.ts
@@ -141,17 +141,14 @@ export class TestScheduler extends VirtualTimeScheduler {
     }
 
     super.flush();
-    const { flushTests } = this;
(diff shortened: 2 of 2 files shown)


TPReal:
There is one thing I'm not sure about. The original change that introduced the bug (https://github.com/ReactiveX/rxjs/commit/2d5b3b284f6520c130f2dc0d8928fe4d7db9ee46#diff-a65b432aff1fe2954db903fd7fe6b322R141) changed the code so that `flushTests` was only modified in place (so it could be readonly, but still wasn't). I don't see why this would be important not to change the reference, but on the other hand what would be any other reason to modify `flush()` at all??

It would be possible to fix the issue without modifying the `flushTests` reference, but the fix that I proposed also seems to work, and the code is nicer. So I don't know, just dumped my thoughts here.

cartant:
@TPReal After I created the PR, I looked to see if you had any commits in this or other repos from which I could obtain your email address - so that I could credit you with the PR's commits - but I couldn't find any. However, I just noticed that your email address is on your GitHub profile page, so I'll make you the author in a short while - as it's all your work.

jayphelps:
> There is one thing I'm not sure about. The original change that introduced the bug (2d5b3b2#diff-a65b432aff1fe2954db903fd7fe6b322R141) changed the code so that flushTests was only modified in place (so it could be readonly, but still wasn't). I don't see why this would be important not to change the reference, but on the other hand what would be any other reason to modify flush() at all??

I don't recall specifics unfortunately. Presumably there was an issue with the previous implementation, but it's not clear why. You solution appears fine, and if not, someone will report and we can fix appropriately.

TPReal:
Thanks for the quick fix, and for the credit!

cartant:
Interestingly, this change broke a test that I had in `rxjs-marbles`. [This particular test](https://github.com/cartant/rxjs-marbles/blob/76ceb5097dd8cf9372a636592caa9d2a27a9a545/fixtures/mocha/issues-spec.ts#L15-L21) explicitly calls `flush` and, because the test uses `TestScheduler.run`, `flush` is called again at the end of the test.

Prior to this change, that wasn't a problem, as each ready test was removed from the array before the assertion was performed. So the additional call to `flush` did nothing, as the ready test had been removed. Now, the array is filtered, so if an assertion fails, the array is unchanged and the same assertion will fail whenever `flush` is again called.

However, I doubt this is going to be an issue for any situation in which the `TestScheduler` is not itself being tested - which is, basically, what I was doing in the above-referenced test. I don't think there's anything that needs to be fixed. I'm just commenting to make a note of this, as I've had to work around it, today.

Issue #3717, opened 2018-05-21
Redundant tests for the first/last operators


cartant:
Whilst fixing the typing problems in the test suite, I found a [seemingly redundant test](https://github.com/ReactiveX/rxjs/blob/6.1.0/spec/operators/first-spec.ts#L173-L230) for the `first` operator:

```ts
it('should support type guards without breaking previous behavior', () => {
```

`first` does not accept a used-defined type guard, so the test does nothing. Indeed, it will fail once type checking is enabled for the test suite.

Whether or not the test should be removed or the signature of `first` should be modified to accept a user-defined type guard, I don't know. Hence the opening of this issue.

cartant:
There is a similar problem with `last`.

benlesh:
Yeah, it just looks poorly written. We do need to make sure that type guards work with operators like `first`, `last`, `filter`, etc.

cartant:
I'll come back and fix these up, then, once the test suite is type-checked. Some overload signatures will need to be added for the type guards to effect narrowed observable types.

PR #3902, opened 2018-07-08
chore(typings): add type guard overloads for first/last

cartant:
**Description:**

Uncomments the tests that were skipped when enabling type checking for the test suite. And adds overload signatures with type guards to `first` and `last`.

Note that because the predicates are optional, additional overload signatures that match calls made with no arguments or with a `null` or `undefined` predicate are required. Otherwise, the `S` type parameter is inferred as `{}` and the type of the output observable becomes `Observable<{}>`.

**Related issue (if exists):** #3717

### DIFF
--- a/compat/operator/first.ts
+++ b/compat/operator/first.ts
@@ -1,8 +1,12 @@
 import { Observable } from 'rxjs';
 import { first as higherOrder } from 'rxjs/operators';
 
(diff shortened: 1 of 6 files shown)


benlesh:
:+1:

benlesh:
Oops... just noticed after I merged this... next time this should be labelled a "fix" so it shows up in the change log, since it's an improvement.

PR #3904, opened 2018-07-09
Made scan description more accurate and linked to ref docs

jonahbron:
**Description:**

Fixing an incorrect description

### DIFF
--- a/doc/introduction.md
+++ b/doc/introduction.md
@@ -64,7 +64,7 @@ fromEvent(button, 'click')
   .subscribe(count => console.log(`Clicked ${count} times`));
 ```
 
-The **scan** operator works just like **reduce** for arrays. It takes a value which is exposed to a callback. The returned value of the callback will then become the next value exposed the next time the callback runs.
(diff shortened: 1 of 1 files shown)


doc/introduction.md
```diff
@@ -64,7 +64,7 @@ fromEvent(button, 'click')
   .subscribe(count => console.log(`Clicked ${count} times`));
 ```
 
-The **scan** operator works just like **reduce** for arrays. It takes a value which is exposed to a callback. The returned value of the callback will then become the next value exposed the next time the callback runs.
+The [**scan**](https://rxjs-dev.firebaseapp.com/api/operators/scan) operator works similarly to [**reduce**](https://rxjs-dev.firebaseapp.com/api/operators/reduce) for arrays, except it emits every accumulated value instead of only the final one. It takes a value which is exposed to a callback. The returned value of the callback will then become the next value exposed the next time the callback runs.
```
benlesh: I'm not sure about this one... @JWO719  is this the best way to link things? Seems like it would require a lot of updating when we finally get the domain name in place.

jonahbron: @benlesh Good point, I can adjust to make them relative to the domain root, if you guys don't expect the path to change.

jonahbron:
@benlesh Updated the URLs.

PR #3907, opened 2018-07-09
docs(of): add of operator docs

rarmatei:
**Description:**

add description and examples for the rxjs/of observable creator

### DIFF
--- a/src/internal/observable/of.ts
+++ b/src/internal/observable/of.ts
@@ -47,9 +47,9 @@ import { Observable } from '../Observable';
  * ```
  *
  * @see {@link from}
+ * @see {@link range}
  *
- * @param {T} args A comma separated list of arguments to be converted
- * into next notifications
+ * @param {...T} values A comma separated list of arguments you want to be emitted
  * @return {Observable} An Observable that emits the arguments
  * described above and then completes.
  * @method of
--- a/src/internal/observable/of.ts
+++ b/src/internal/observable/of.ts
@@ -5,6 +5,57 @@ import { empty } from './empty';
 import { scalar } from './scalar';
 import { Observable } from '../Observable';
 
+/**
+ * Converts the arguments to an observable sequence.
+ *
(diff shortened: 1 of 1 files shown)


rarmatei:
I tried following other examples, like `ignoreElements` or `count` when writing - But when I run it locally, for some reason, all my additions go under the `Overloads` section, and not at the top.

If someone could help me understand why, I'd be happy to push a fix asap. Thanks so much in advance!

src/internal/observable/of.ts
```diff
@@ -5,6 +5,57 @@ import { empty } from './empty';
 import { scalar } from './scalar';
 import { Observable } from '../Observable';
 
+/**
+ * Converts the arguments to an observable sequence.
+ *
+ * <span class="informal">Each argument becomes a `next` notification.</span>
+ *
+ * ![](of.png)
```
rarmatei: I realize I'm commenting on my own PR - but where do these files come from?

JWO719: Could you elaborate a bit more in detailed what's the problem? I actually don't get it sorry

rarmatei: Sure - the images on the [docs site](https://rxjs-dev.firebaseapp.com/api/operators/mergeMap) don't seem to be loading:

image

image

Is it still WIP?

JWO719: I am not sure when the last version was deployed, but there was a change that the images have to be generated. To do that you have to call `npm run tests2png` (there is a documentation about that, but I actually don't find it :D). Maybe this was forgotten last time, and it definetly has to be added to the deployment script. But you also have to do it locally. If you have any problems with that, just ping me

benlesh:
Thanks, @rarmatei

src/internal/observable/of.ts
martinsik: @JWO719 Did you create an issue for this? Maybe this could be optional or at least ignored when fails. I don't know if you should be forced to install imagick if you only want to test eg. docs typos.

Issue #3900, opened 2018-07-06
The decision tree widget uses older operator/function names


mtaran-google:
### Documentation Related To Component:

https://github.com/ReactiveX/rxjs/blob/master/doc/decision-tree-widget/tree.yml

### Please check those that apply

- [ ] typo
- [ ] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

### Description Of The Issue

See above. This uses some older syntax (e.g. `Observable.throw` instead of `throwError`). I also couldn't discover any version of it on google (aside from the github source) by searching for some of the included strings, so it seems to not be very discoverable. This is a bit of a shame, since it could be a great resource. What would be the process for either improving its visibility or removing it since it's yet another thing to keep up to date with API changes?

cartant:
This seems to be a dupe of https://github.com/ReactiveX/rxjs/issues/3866 and see also https://github.com/ReactiveX/rxjs-docs/issues/251.

mtaran-google:
Thanks! I guess I didn't look hard enough to find that name for it.

Then perhaps we can leave the question of discoverability to those bugs. Instead I'd like to focus on how best to update it -- would there be any concerns if I sent a PR renaming the legacy creation operator usages to their modern equivalents?

cartant:
Sure. Focusing on updating it sounds good to me, as I'd like to see it reinstated. I'd welcome a PR.

ladyleet:
@cartant should we close this?

cartant:
@ladyleet This is the issue for https://github.com/ReactiveX/rxjs/pull/3912. That PR has not been merged - which, incidentally, contains [some discussion](https://github.com/ReactiveX/rxjs/pull/3912#issuecomment-403732688) about the `/doc` directory and what its purpose is supposed to be.

I guess this could be closed. I don't know whether or not there's a particular policy regarding issues remaining open until their associated PR is closed. I tend to leave them open until the PR is merged; typically by including, for example, `Closes #3900` in a commit message.

PR #3912, opened 2018-07-09
doc(decision-tree): update operator/function names

mtaran-google:
**Description:** updates to the operator decision tree widget's content

**Related issue (if exists):** #3900

### DIFF
--- a/doc/decision-tree-widget/tree.yml
+++ b/doc/decision-tree-widget/tree.yml
@@ -15,7 +15,7 @@ children:
       - label: pluck
     - label: I want to spy the values being emitted without affecting them
       children:
-      - label: do
+      - label: tap
     - label: I want to allow some values to pass
       children:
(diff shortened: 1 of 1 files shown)


JWO719:
In my opinion these changes shouldn't be part of the old docs. I thought we use the /doc folder for documenting rxjs < 6? I'm not sure if we want to mix it up

cartant:
@JWO719 I'm not sure what's supposed to be where, but there is already some v6-specific documentation in `/doc` directory: [`doc/marble-testing.md`](https://github.com/ReactiveX/rxjs/blob/master/doc/marble-testing.md).

JWO719:
@cartant I totally share your opinion on that, but we should agree on an intention of the `/doc` folder

benlesh:
Thanks, @mtaran-google

PR #3914, opened 2018-07-09
doc: explicitly note that resultSelector is deprecated

mtaran-google:
**Description:** makes it clearer for readers of the generated docs that resultSelector is deprecated

**Related issue (if exists):** n/a

### DIFF
--- a/src/internal/operators/switchMapTo.ts
+++ b/src/internal/operators/switchMapTo.ts
@@ -47,7 +47,7 @@ export function switchMapTo<T, I, R>(observable: ObservableInput<I>, resultSelec
  * @param {ObservableInput} innerObservable An Observable to replace each value from
  * the source Observable.
  * @return {Observable} An Observable that emits items from the given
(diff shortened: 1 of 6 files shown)


src/internal/operators/switchMapTo.ts
```diff
@@ -47,8 +47,8 @@ export function switchMapTo<T, I, R>(observable: ObservableInput<I>, resultSelec
  * @param {ObservableInput} innerObservable An Observable to replace each value from
  * the source Observable.
  * @return {Observable} An Observable that emits items from the given
- * `innerObservable` (and optionally transformed through `resultSelector`) every
- * time a value is emitted on the source Observable, and taking only the values
+ * `innerObservable` (and optionally transformed through the deprecated `resultSelector`)
```
cartant: According to the linter rules, this line has trailing whitespace that has to be removed:

```
ERROR: src/internal/operators/switchMapTo.ts[50, 90]: trailing whitespace
```

ATM, the linter is effecting a failure and that breaks the full-validation CI build.

PR #3923, opened 2018-07-14
build(docs): add image generation to doc generation

JWO719:


martinsik:
@JWO719 I'm not sure how the build process works but `npm run tests2png` needs to be called before building the docs app because it checks whether the images exist. https://github.com/ReactiveX/rxjs/blob/master/docs_app/tools/transforms/angular-api-package/post-processors/embedMarbleDiagrams.js#L24

JWO719:
yeah I know this one is used if Ben is releasing a new version of rxjs. In that case this script is called and it will first generate the images and afterwards build the docs and deploy it in firebase.

JWO719:
That's a really good hint. So basically it is working, I think. But I faced some windows related problems, as always.

benlesh:
Thanks @JWO719

PR #3932, opened 2018-07-17
fix: convert @internal comment to JSDoc

mprobst:
`@internal` and similar tags are only parsed in JSDoc.


mprobst:
/CC @benlesh minor change, but unbreaks a Closure Compiler interop.

benlesh:
Thank you @mprobst!

Issue #3823, opened 2018-06-10
Use dtslint for better type expectations


cartant:
## Feature Request

To better test TypeScript expectations, we should be able to utilise Microsoft's dtslint tool.

I've had a look at how it could be integrated into the repo and I've created a branch that adds a `dtslint` script and adds the `dtslint`-related configuration and type-only tests to the `type-specs` directory:

[cartant/rxjs/tree/dtslint/type-specs](https://github.com/cartant/rxjs/tree/dtslint/type-specs)

The `dtslint` script can be added to the Travis `$FULL_VALIDATE` configuration.

This looks to me to be a great tool. Thoughts?

I was thinking that the `type` tests currently used in the test suite could be pretty easily be brought across to `dtslint` - in which the expectations would be more robustly tested.

**Additional context**

* [Microsoft/dtslint](https://github.com/Microsoft/dtslint)
* [Testing Types: An Introduction to dtslint](https://medium.com/@danvdk/testing-types-an-introduction-to-dtslint-b178f9b18ac8)

PR #3944, opened 2018-07-23
chore(test): add dtslint files and script

cartant:
**Description:**

This PR adds [dtslint](https://github.com/Microsoft/dtslint) to the full-validation build. The dtslint-related files are in the `spec-dtslint` directory.

The `type` tests for the `zip` factory function and the `zip` operator have been included so that they are tested using dtslint's expectations. The original `type` tests have not been removed.

**Related issue (if exists):** #3823


cartant:
Hmm. I hadn't noticed that npm had made such a mess of the lock file. I will try to fix it up a little.

kwonoj:
though lockfile itself seems ok-change?

cartant:
I'm just going to re-apply the commits and do a single install of dtslint using the current lock file from master. I'm not going to manually mess with the lock file.

The initial commits and install of dtslint were from some time ago and I had to re-install dtslint to get an updated dependency - as I'm creating this PR on a different computer - and I really don't think that the lock file should have as many changes in it as it appears to have. I'd like to be sure that what was in the `node_modules` directory on this computer hasn't messed with the lock file. npm infuriates me sometimes.

Issue #3841, opened 2018-06-14
switchAll, mergeAll etc are not type safe


felixfbecker:
## Bug Report

**Current Behavior**
```ts
const input: Observable<number> = of(1, 2, 3)
const output = input.pipe(switchAll())
output.subscribe(value => console.log(value))
```
No type error, but runtime error:
```
You provided '1' where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.
```

**Reproduction**
- REPL or Repo link: https://stackblitz.com/edit/typescript-zgtnbp

**Expected behavior**
This should give a type error like `number is not assignable to ObservableInput<T>`

**Environment**
- RxJS version: 6.2.1

**Possible Solution**
I tried a few things, but it seems like type inference just fails here. TypeScript needs to infer a type based on the _usage_ of the return value here.

**Workaround**
Use `switchMap(x => x)` instead (`switchMap(identity)` also breaks inference)

cartant:
The problem seems to have been introduced in 6.2.1 with [this commit](https://github.com/ReactiveX/rxjs/commit/ad010ea2029f06113406865427486cee0655d440) - which uses the signature that I suggested.

Prior to that commit, the snippet would have effected this error:

```
[ts]
Argument of type 'OperatorFunction<ObservableInput<{}>, {}>' is not assignable to parameter of type 'OperatorFunction<number, {}>'.
  Type 'number' is not assignable to type 'ObservableInput<{}>'.
```

The commit fixed an error with the `pipe` signature, but it could be improved. A more refined rest-parameters signature for `pipe` would be:

```ts
pipe<R>(op1: OperatorFunction<T, any>, ...operations: OperatorFunction<any, any>[]): Observable<R>;
```

Such a signature would avoid an matching the first parameter with `any`, but, unfortunately, that's as far as this approach could be taken. It would not solve the problem if `switchAll` where, say, the second of two parameters. For that situation, a signature like this would be required:

```ts
pipe<A, R>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, any>, ...operations: OperatorFunction<any, any>[]): Observable<R>;
```

However, such a signature requires the inclusion of another type parameter. And given that the caller of such a rest-parameter signature will want to explicitly specify `R`, the signature isn't really usable.

I don't know if it's going to be possible to support a rest-parameters signature with which the caller can specify only the result type (`R`) and proper type safety for the parameters.

---

[...]

PR #3945, opened 2018-07-24
fix(pipe): replace rest parameters overload [depends on PR #3944]

cartant:
**Description:**

This PR depends upon the dtslint additions in #3944.

It fixes a problem that was introduced in [this commit](https://github.com/ReactiveX/rxjs/commit/ad010ea2029f06113406865427486cee0655d440). Prior to that commit, the `pipe` overload signature for more than 9 arguments looked like this:

```ts
pipe<R>(...operations: OperatorFunction<T, R>[]): Observable<R>;
```

Which is incorrect, as consecutive arguments have to have matching output/input types - every `OperatorFunction` passed cannot accept a `T` and return an `R`.

Unfortunately, fixing that by changing the signature to the following introduced another problem:

```ts
pipe<R>(...operations: OperatorFunction<any, any>[]): Observable<R>;
```

With that signature, any usage with 9 or fewer arguments that does not match a signature due to type errors will now match the rest parameters signature. So strong typing is subverted and that's the problem highlighted in #3841.

In [a comment in this TypeScript issue](https://github.com/Microsoft/TypeScript/issues/25660#issuecomment-405336498), Anders Hejlsberg has pointed out the output/input relationship between arguments means that the overloads cannot be avoided. So to restore strong typing, this PR replaces the rest parameters overload signature with a signature that also includes the preceding 9 type-parameter-specified arguments, like this:

```ts
pipe<A, B, C, D, E, F, G, H, I, R>(
  op1: OperatorFunction<T, A>,
  op2: OperatorFunction<A, B>,
  op3: OperatorFunction<B, C>,
  op4: OperatorFunction<C, D>,
  op5: OperatorFunction<D, E>,
  op6: OperatorFunction<E, F>,
  op7: OperatorFunction<F, G>,
  op8: OperatorFunction<G, H>,
  op9: OperatorFunction<H, I>,
  ...operations: OperatorFunction<any, any>[]
): Observable<R>;
```

The effect of this is that it will no longer be possible to specify a single `R` type parameter for `pipe` calls that involve more than 9 arguments. Instead, an assertion would need to be used like this:

```ts
const result = of("T").pipe(
    mapTo("A"),
    mapTo("B"),
    mapTo("C"),
    mapTo("D"),
    mapTo("E"),
    mapTo("F"),
    mapTo("G"),
    mapTo("H"),
    mapTo("I"),
    mapTo("R")
) as Observable<"R">; // Otherwise inferred to be Observable<{}>
```

This PR also adds dtslint tests for `pipe` and the static `pipe` utility function that include expectations for type inference and errors.

**Related issue (if exists):** #3841


cartant:
Hmm. This PR is uncovering a bunch of typing problems in the test suite.

type-specs/Observable-spec.ts
```diff
@@ -0,0 +1,94 @@
+import { of, OperatorFunction } from 'rxjs';
+import { mapTo } from 'rxjs/operators';
+
+function a<I extends string, O extends string>(input: I, output: O): OperatorFunction<I, O>;
+function a<I, O extends string>(output: O): OperatorFunction<I, O>;
+function a<I, O extends string>(inputOrOutput: I | O, output?: O): OperatorFunction<I, O> {
+  return mapTo<I, O>(output === undefined ? inputOrOutput as O : output);
+}
+
```
benlesh: we actually just hit an issue where the following broke in one of our apps here at Google:

```ts
const customOperator = () => <T>(a: Observable<T>) => a;

of('string').pipe(customOperator());
```

NOTE: `customOperator` is `() => <T>(source: Observable<T>) => Observable<TT>` NOT `<T>() => (source: Observable<T>) => Observable<T>`, which seems to work fine.

Can we add a test for that here too?

rkirov: I narrowed it down to our use of 'noStrictGenerics' flag, which we should be using anyways. Anyways, good idea to add a test for it, though I expect it to just pass in your setup.

src/internal/Observable.ts
```diff
@@ -318,7 +318,7 @@ export class Observable<T> implements Subscribable<T> {
    *   .subscribe(x => console.log(x))
    * ```
    */
-  pipe<R>(...operations: OperatorFunction<T, R>[]): Observable<R> {
+  pipe<R>(...operations: OperatorFunction<any, any>[]): Observable<R> {
```
rkirov: One thing to consider here is `pipe<R=any>`, so that you don't over-rely on return type for inference.

cartant: @rkirov Can the inference really bubble out of the implementation? I'll take your word for it, as I'm unclear on the actual inference rules. (If you know of any resources - other than the TypeScript source - that explain the workings of TypeScript's type inference, I'd really appreciate if you could direct me to them.)

Anyway, I'm not sure that `R` is even required here, shouldn't we use something like this?

```ts
pipe(...operations: OperatorFunction<any, any>[]): Observable<any>
```

With the `any` type parameters already being specified in the rest-parameters type and with it being the implementation of an overloaded function, does `R` really offer any value whatsoever?

rkirov: You are right about "bubbling out", my comment was pertaining to line 299, but I got mixed up. The types here are only relevant for 'implementation', not for the consumer.

About learning how type inference works - TS doesn't maintain a spec any longer, so the best one can do if they want to learn the rules is keep testing with small examples (what I do) or read the source code (haven't done that for type inference).

What makes matters hard here is that this is mixing generics and function overloads. Separately each one has odd corners, and together things are twice as hard.

Historically, we (team maintaining TS in google) have been having a lot of issues with generics that are only inferred through the return type of a function - simplest example being `f<T>(): T`. TS will try to use an context type to pick T, but sometimes there will be no such type to pick, so it will fallback to `{}`. 

What happens is that in simple cases everything works fine:
```typescript
const x: string = f();  // T is inferred to be string
const y = f();  // y is {}, which is ok, its an opaque object.
```

But what we see the inferred `{}` can be propagating too much, and into another loosely typed API.
```typescript
function g() {
  return f();
}
function someAPI(x: any) {...}

someAPI(g());
```

Now, if the maintainer tries to change `someAPI` to be stricter say 'x: string', they are confronted with a `{}` that flowed through without anybody noticing, which makes it hard to change.  That's why we are considering whether it would have been better if the fallback for when TS cannot infer the generic type should have been `any` instead of the default `{}` (technically it is the lower bound, i.e whatever is after `extends`). Paradoxically, even though `any` is less safe, it can help making other type change stricted.

Maybe it would have been better if TS forced people to write a type like `f<Something>`, instead of fallback on `{}`.

We are still figuring out what is most ergonomic for ppl, would love to hear your opinion too.

cartant: > Maybe it would have been better if TS forced people to write a type like `f<Something>`, instead of fallback on `{}`.

Yes. I agree wholeheartedly with this.

src/internal/Observable.ts
```diff
@@ -296,7 +296,7 @@ export class Observable<T> implements Subscribable<T> {
...
   pipe<A, B, C, D, E, F, G, H>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>, op5: OperatorFunction<D, E>, op6: OperatorFunction<E, F>, op7: OperatorFunction<F, G>, op8: OperatorFunction<G, H>): Observable<H>;
   pipe<A, B, C, D, E, F, G, H, I>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>, op5: OperatorFunction<D, E>, op6: OperatorFunction<E, F>, op7: OperatorFunction<F, G>, op8: OperatorFunction<G, H>, op9: OperatorFunction<H, I>): Observable<I>;
-  pipe<R>(...operations: OperatorFunction<any, any>[]): Observable<R>;
+  pipe<A, B, C, D, E, F, G, H, I, R>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>, op5: OperatorFunction<D, E>, op6: OperatorFunction<E, F>, op7: OperatorFunction<F, G>, op8: OperatorFunction<G, H>, op9: OperatorFunction<H, I>, ...operations: OperatorFunction<any, any>[]): Observable<R>;
```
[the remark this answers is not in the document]
rkirov: @evmar and I were just chatting about the options here. Here is the total space of options (simplified to only talk about the return type).

```typescript
// using {} (could be changed to unknown post TS 3.0)
declare function f1<T>(): T;
declare function f2(): {};
declare function f3<T={}>(): T;

// using any
declare function g1(): any;
declare function g2<T=any>(): T;
```

He convinced me that picking up 'any' without ever writing it explicitly in their user code is a bad idea. Also, f1 and f3 are equivalent (at least I haven't gotten an example where they can be separated). So what's left is f1 and f2. Evan's point (and I tend to agree) is that f2 is more explicit because:

```typescript
let x: string = f1(); // ok
let y: string = f2(); // error, needs an 'as' cast
let z = f2() as string;  // ok
```

So our suggestion is f2(), i.e. inline `Observable<{}>` and drop the R. The downside is the syntactic connivence of `f<string>()` vs `f() as Observable<string>`. Thankfully, the 'as' casts are expressions like any other, so they can be used anywhere, and we are just talking about syntactic preference.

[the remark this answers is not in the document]
cartant: Thanks. That makes sense.

cartant:
@benlesh As discussed with @rkirov, the `R` type parameter has been removed so that `OperatorFunction<T, {}>` is returned and an assertion is required. And a test has been added for this.

Note that this makes it even more likely that there will be typing issues exposed in users' code when/if a release including this PR's changes is published.

Issue #3951, opened 2018-07-25
first/last typings are incorrect


cartant:
## Bug Report

**Current Behavior**

The implementation signature is not compatible with all of the overload signatures, as it does not support a `null` predicate. However, an error isn't effected with the particular version of TypeScript used in the project.

Also, the signatures that accept type guards allow defaults to be specified. However, the defaults are of type `T` - which makes no sense, as the type guards are supposed to ensure that only values of type `S` will be passed.

PR #3952, opened 2018-07-25
 fix(first/last): fix type-guard signatures

cartant:
**Description:**

Fixes problems with the implementation and type-guard signatures for the `first` and `last` operators.

**Related issue (if exists):** #3951

### DIFF
--- a/src/internal/operators/first.ts
+++ b/src/internal/operators/first.ts
@@ -16,7 +16,7 @@ export function first<T>(
 ): MonoTypeOperatorFunction<T>;
 export function first<T, S extends T>(
   predicate: (value: T, index: number, source: Observable<T>) => value is S,
-  defaultValue?: T
+  defaultValue?: S
 ): OperatorFunction<T, S>;
 export function first<T>(
   predicate: (value: T, index: number, source: Observable<T>) => boolean,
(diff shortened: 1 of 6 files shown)


src/internal/operators/first.ts
```diff
@@ -72,10 +72,10 @@ export function first<T>(
  * @method first
  * @owner Observable
  */
-export function first<T>(
+export function first<T, S extends T>(
   predicate?: ((value: T, index: number, source: Observable<T>) => boolean) | null,
-  defaultValue?: T
-): MonoTypeOperatorFunction<T> {
+  defaultValue?: S
+): OperatorFunction<T, S> {
```
benlesh: this doesn't seem right... the default value doesn't have to extend `T`, it could be anything.

`first<T, D> ` and return `:OperatorFunction<T, T|D>` seems more accurate, then type overloads for the more common cases.

cartant:
@benlesh I've updated the signatures to separate `D` from `T` and I've added dtslint-based tests for the variations.

cartant:
FYI, I looked at separating `D` from `S` in the user-defined type guard signature, but doing so causes problems. With this signature:

```ts
export function first<T, S extends T, D = S>(
  predicate: (value: T, index: number, source: Observable<T>) => value is S,
  defaultValue?: S
): OperatorFunction<T, S | D>;
```

this usage:

```ts
const o = of('foo').pipe(first(isFooBar, 'bar'));
```

sees `D` inferred as `string` (for whatever reason) and the return type then becomes `Operator<string, string>` instead of `Operator<string, "foo" | "bar">` - which renders the type guard ineffectual.

This might be addressable by splitting the signature into two - one with the default and one without - but I have a suspicion that the signature that includes the default will still render the type guard ineffectual and the overall behaviour will be the same as with the current signatures.

benlesh:
Nice!

(It looks like appveyor flaked on this one, merging anyhow)

PR #3956, opened 2018-07-25
unused variables

henjun:
An unused variable seems to exist for the _ execute function.

**Description:**
Couldn't unused variables affect readability?


benlesh:
@henjun ... while you're right and these variables are unused, for type safety reasons, TypeScript will not build with this change. I appreciate the effort, but I'm going to need to close this PR.

Thank you for your time, @henjun! I hope this doesn't put you off from contributing again!

PR #3959, opened 2018-07-27
chore(package): update lock file

cartant:
**Description:**

This PR updated the lock file using these commands (with npm v6.2.0):

```
npm uninstall --save-dev dtslint
npm install --save-dev --save-exact dtslint
```


benlesh:
Thanks, I started doing this and got pulled into something else. :)

Issue #3879, opened 2018-07-01
scan() definition


Karql:
Hi!

I've just started my journey with rxjs ;)
Reading overview I've gone to scan() documentation.
After some lecture I thought "why seed is type of T|R"

I checked the source code:
https://github.com/ReactiveX/rxjs/blob/6.2.1/src/internal/operators/scan.ts

Ok seed is type T|R in case of use first value of source when is not passed.
But imo this is correct behavior when T == R

When T != R it can case errors like this:

![scan](https://user-images.githubusercontent.com/173098/42131959-485e3d42-7d0e-11e8-997c-b83025607966.png)

Imo to avoid this seed should be mandatory when T != R.
It imiples on change in definition from:
`export function scan<T, R>(accumulator: (acc: R, value: T, index: number) => R, seed?: R): OperatorFunction<T, R>;`
to:
`export function scan<T, R>(accumulator: (acc: R, value: T, index: number) => R, seed: R): OperatorFunction<T, R>;`

What do you think? Should I create PR for this?

bafolts:
Requiring users to pass the initial value of their seed seems like the safest route. If not requiring users to provide a value for the first seed the final scan really gets `T | R`.

```
export function scan<T, R>(accumulator: (acc: T | R, value: T, index: number) => R, seed?: R): OperatorFunction<T, R>;
```

If not requiring an initial value for the seed `undefined` for the original value would make sense. It is strange to me that the seed becomes of type `T` in this final case.

BioPhoton:
@cartant I believe we can close this issue please correct me if not and in the best case suggest a way to proceed with this

Issue #3150, opened 2017-12-01
types inferring problem with scan operator


ghetolay:
I'm having a problem when using `scan` with 2 different type for item and accumulator/seed on ts 2.5+ (didn't try older versions).
You can see it here : https://stackblitz.com/edit/ts-fn-overload-bug?file=test.ts

```ts
of('0').pipe(
  scan((acc, str) => acc + parseInt(str), 0)
);
```
gives the following error : 
> Argument of type 'UnaryFunction<Observable<number>, Observable<any>>' is not assignable to parameter of type 'UnaryFunction<Observable<string>, Observable<any>>'.
  Types of parameters 'source' and 'source' are incompatible.
    Type 'Observable<string>' is not assignable to type 'Observable<number>'.
      Type 'string' is not assignable to type 'number'.

Even if scan has a signature matching that use case (the last one) : 
```ts
function scan<T>(accumulator: (acc: T, value: T, index: number) => T, seed?: T): MonoTypeOperatorFunction<T>;
function scan<T>(accumulator: (acc: T[], value: T, index: number) => T[], seed?: T[]): OperatorFunction<T, T[]>;
function scan<T, R>(accumulator: (acc: R, value: T, index: number) => R, seed?: R): OperatorFunction<T, R>;
```

As it's shown on the stackblitz, order seems to matter so it's most likely a ts bug.
But looking closely on those types, @NaridaL pointed out that the last signature should be enough to cover all cases. Indeed `R` can totally be inferred with the same type as `T` or `T[]`.

So the question is : is there any reason for having those *subtype* signatures ? especially since now it seems to create a bug ?

P.S: Couldn't find a corresponding ts issue but didn't do much digging
P.S2: I think the same applies for `reduce`.

cartant:
Probably related to #2897. See discussion therein.

ghetolay:
Oh sorry, I missed that. 

But now there is a game changer as it's causing errors. I'll dig into typescripts issues see if I can find something.

ghetolay:
I thought the problem was because I update my ts but it's not, it's because I refactored my code to use pipe. 
I can go as low as 2.0 with that code and I got the error while the *non-pipeable* version, with `this` on the signature, works fine.

I found this new issue https://github.com/Microsoft/TypeScript/issues/20305 but I'm not sure it's exactly the same problem.

benlesh:
This has been addressed to the best of our ability in RxJS 7.

PR #3960, opened 2018-07-27
Update scan.ts

domoritz:
Otherwise this breaks

```ts
$brush.pipe(
    scan(
      (acc, curr) => acc.set(curr.dimension, curr.brush),
      new Map<D, Interval<number>>()
    )
  );
```

### DIFF
--- a/src/internal/operators/scan.ts
+++ b/src/internal/operators/scan.ts
@@ -4,9 +4,9 @@ import { Subscriber } from '../Subscriber';
 import { OperatorFunction, MonoTypeOperatorFunction } from '../types';
 
 /* tslint:disable:max-line-length */
+export function scan<T, R>(accumulator: (acc: R, value: T, index: number) => R, seed?: R): OperatorFunction<T, R>;
 export function scan<T>(accumulator: (acc: T, value: T, index: number) => T, seed?: T): MonoTypeOperatorFunction<T>;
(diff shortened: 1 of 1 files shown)


cartant:
Thanks for opening the PR.

Having a another look at the typings for `scan` and `reduce` is on my list of things to do. There are a number of open issues that related to the `scan` typings: #3150 and #3879. Any changes to the `scan` typings should take those into account.

The change in this PR appears to be the same as the change suggested in #2897 - which was not merged. See the conversation in that PR for the reasons. I'm not saying that those reasons still apply - and I've not re-read the conversation recently - but it does provide some context.

IMO, now that the testing infrastructure supports dtslint, I don't think we should change the typings until we have dtslint-based tests that document/describe the current behaviour.

Also any changes to the `scan` typings will likely need to be made to the `reduce` typings, too.

domoritz:
Thanks for getting back to me so quickly. I'm going to close this PR then. 

I just started using rxjs yesterday and tripped over this issue. I'll just set the types manually for now.

cartant:
Well, hopefully, I've not discouraged you from contributing, as that wasn't my intention.

Anyway, I'll bump the priority of my writing some dtslint tests for `scan` and `reduce`, so that any changes we make to the typings need not be so scary.

domoritz:
Oh no, I'm not discouraged not at all. I know how open source development goes and I'm impressed how quickly you responded. Thanks for taking my request seriously!

PR #3967, opened 2018-07-28
fix(IE10): Remove dependency on Object.setPrototypeOf

### LINKED ISSUE #3966: Object.setPrototypeOf not available in IE10
As of RxJS 6, partners like Angular are having a hard time with IE10 because we are using Object.setPrototypeOf in our Error classes.

Related: https://github.com/angular/angular/issues/24488

benlesh:
This reimplements all Error types as ES5 classes, and types them with TypeScript such that
we no longer need to rely on Object.setPrototypeOf, which was causing issues for users who had not
polyfilled it properly in IE10.

fixes #3966


croso-df:
Has this been released already? I can still see setPrototypeOf in 6.2.2

PR #3968, opened 2018-07-28
fix(find): unsubscribe from source when found

### LINKED ISSUE #3934: Operator find does not unsubscribe on match
## Bug Report

**Current Behavior**
The `find` operator does not unsubscribe from the source observable when a match is found, but rather only after the result observable is unsubscribed from. This means that `find(()=>true)` behaves differently from `first()`, and the behaviour of `first()` is better.

**Reproduction**
https://rxviz.com/v/1o7zEN08
https://stackblitz.com/edit/typescript-ghjbsy

```js
// Test 1:
let test1='t1';
interval(200).pipe(
  take(10),
  // This happens only once as the interval is unsubscribed from after first() matches.
  tap(()=>test1+='-'),
  first(),
  delay(900),
  map(()=>test1),
).subscribe(v=>console.log(v));

// Test 2:
let test2='t2';
interval(200).pipe(
  take(10),
  // This happens many times because find does not unsubscribe on match.
  tap(()=>test2+='-'),
  find(()=>true),
  delay(900),
  map(()=>test2),
).subscribe(v=>console.log(v));
```

**Expected behavior**
Find should unsubscribe from the source when it matches the first element. It is not justified to hold the subscription as we know that no more elements from the source observable will be useful for us.

**Environment**
- Runtime: Chrome 67.0.3396.87
- RxJS version: 6.2.2

cartant:
**Description:**

Fixes #3934 and adds unsubscribe-when-matched tests for `find`, `findIndex` and `first`.

**Related issue (if exists):** #3934

### DIFF
--- a/spec/operators/find-spec.ts
+++ b/spec/operators/find-spec.ts
@@ -1,9 +1,12 @@
 import { expect } from 'chai';
-import { find, mergeMap } from 'rxjs/operators';
+import { find, mergeMap, delay } from 'rxjs/operators';
+import { TestScheduler } from 'rxjs/testing';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 import { of, Observable, from } from 'rxjs';
 
 declare function asDiagram(arg: string): Function;
+
+declare const rxTestScheduler: TestScheduler;
 
 /** @test {find} */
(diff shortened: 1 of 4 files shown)


benlesh:
Awesome! Thanks!

Issue #3971, opened 2018-07-29
Why is file size so big?


bcherny:
I'm using RxJS6 in [Undux](https://github.com/bcherny/undux), via [typed-rx-emitter](https://github.com/bcherny/typed-rx-emitter). I use it by importing `Observable`, and calling `Observable.create`. I don't use any other Rx methods.

Bundling Undux (+typed-rx-emitter, +rxjs) in as part of [undux-todomvc](https://github.com/bcherny/undux-todomvc/blob/master/dist/bundle.js), I see all sorts of things included in the bundle that I'm not using. Eg:

- CombineLatestOperator
- MergeMapOperator
- MapOperator
- RaceOperator
- ZipOperator
- AsyncSubject
- BehaviorSubject
- Subject
- ReplayEvent
- AsapScheduler
- VirtualTimeScheduler
- ForkJoinSubscriber
- ZipBufferIterator

All in all, that's ~80kb (unmangled). Even after running the code through Uglify's tree shaker, these are included. Are they core parts of `Rx.Observable.create`? Is there a way to remove them? Am I doing something wrong when bundling?

Repro:

Generated [bundle.js](https://github.com/bcherny/undux-todomvc/blob/5c92be22dfead43435f884a201f0c24002cbe1ff/dist/bundle.js) ([after Uglify](https://github.com/bcherny/undux-todomvc/blob/602fb5a48709f5e58df4dd70636da9cb651ccecc/dist/bundle.js))

Uglify options:

```
new UglifyJsPlugin({
  cache: true,
  parallel: true,
  sourceMap: true,
  uglifyOptions: {
    keep_classnames: true,
    keep_fnames: true,
    mangle: false
  }
})
```

kwonoj:
You are importing everything in rxjs surface api via https://github.com/bcherny/undux/blob/master/src/index.ts#L1, If codes are targeting v5 import each individual via corresponding subimport, if it's v6 follow guide for optimization. https://github.com/ReactiveX/rxjs/blob/master/doc/pipeable-operators.md#build-and-treeshaking

bcherny:
> You are importing everything in rxjs

No, I'm only importing the type. It's compiled away by TSC before it reaches Webpack. Confirmed by [avoiding the wildcard import](https://github.com/bcherny/undux/blob/ac434772d130ef422c7141fbe045c1f26a4250e2/src/index.ts).
> if it's v6 follow guide for optimization

[...]

PR #3974, opened 2018-07-29
fix(Observable): use more granular Observable exports in compat mode

bcherny:
**Description:** Use more granular exports in compat/Observable.ts to enable tree shaking optimization even if the consumer is compiling to CommonJS (instead of ES2015) modules.

**Related issue (if exists):** https://github.com/ReactiveX/rxjs/issues/3971

Importing `Observable` and calling `Observable.create`, with `"module": "commonjs"` vs `"module": "es2015"` in tsconfig.json:

### Before patch

| | `"module": "commonjs"` | `"module": "es2015"` |
|-|-|-|
| rxjs | 79.6 KiB | 14.7 KiB |
| rxjs-compat | 80 KiB | 80.1 KiB

### After patch

| | `"module": "commonjs"` | `"module": "es2015"` |
|-|-|-|
| rxjs | 79.6 KiB | 14.7 KiB |
| rxjs-compat | 18.6 KiB | 18.7 KiB

### DIFF
--- a/compat/Observable.ts
+++ b/compat/Observable.ts
@@ -1,1 +1,2 @@
-export {Observable, Subscribable, SubscribableOrPromise, ObservableInput} from 'rxjs';
+export {Observable} from 'rxjs/internal/Observable';
+export {Subscribable, SubscribableOrPromise, ObservableInput} from 'rxjs/internal/types';


benlesh:
Great find, @bcherny!

Issue #3969, opened 2018-07-28
Typings for find don't include undefined


cartant:
## Bug Report

**Current Behavior**

The [overload signatures for `find`](https://github.com/ReactiveX/rxjs/blob/6.2.2/src/internal/operators/find.ts#L6-L13) don't reflect the operator's behaviour.

The return types are `OperatorFunction<T, S>` and `MonoTypeOperatorFunction<T>`.

However, the operator will emit `undefined` if no match is found. See [this test](https://github.com/ReactiveX/rxjs/blob/6.2.2/spec/operators/find-spec.ts#L41-L50).

**Possible Solution**

The return types should be `OperatorFunction<T, S | undefined>` and `OperatorFunction<T, T | undefined>`.

PR #3975, opened 2018-07-29
chore(find): fix compat signatures

cartant:
**Description:**

I forgot to update the compat signatures in #3970.

**Related issue (if exists):** #3969


cartant:
PS. This isn't a campaign to prevent the number of open PRs from reaching zero. :smile:

PR #3977, opened 2018-07-30
Cancellable forEach

benlesh:
- Adds cancellation semantic to `forEach`, by allowing a `Subscription` to be passed as a sort of cancellation token.
- Adds documentation
- Fixes some bad tests
- **Deprecation** deprecates passing PromiseCtor to `forEach`. (That can be configured globally, and I'd like to remove it for v7)

### Behavior Described

You can read the docs in the PR for this, but basically:

If you call `forEach` with a second argument that is a `Subscription`, that subscription will act as a cancellation token, such that if you `unsubscribe` from it, the forEach's returned promise will reject with an `Error` with the message `Observable forEach unsubscribed`. This is done, so there's a way to gracefully handle cancellations in async-await code, and I think provides a better alternative to `takeUntil` in those use cases.

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -1,18 +1,16 @@
 import { expect } from 'chai';
 import * as sinon from 'sinon';
-import * as Rx from 'rxjs/Rx';
 import { Observer, TeardownLogic } from '../src/internal/types';
 import { cold, expectObservable, expectSubscriptions } from './helpers/marble-testing';
 import { map } from '../src/internal/operators/map';
 import * as HostReportErrorModule from '../src/internal/util/hostReportError';
 import { noop } from '../src/internal/util/noop';
+import { Subscription, interval, timer, Observable, Operator, config, Subscriber, Subject } from 'rxjs';
 
 //tslint:disable-next-line
 require('./helpers/test-helper');
 
 declare const asDiagram: any, rxTestScheduler: any;
-
-const Observable = Rx.Observable;
 
 declare const __root__: any;
 
@@ -82,7 +80,7 @@ describe('Observable', () => {
     it('should allow Promise to be globally configured', (done) => {
       let wasCalled = false;
 
-      Rx.config.Promise = function MyPromise(callback: any) {
+      config.Promise = function MyPromise(callback: any) {
         wasCalled = true;
         return new Promise<number>(callback);
       } as any;
@@ -114,7 +112,7 @@ describe('Observable', () => {
       });
     });
 
-    it('should handle a synchronous throw from the next handler', () => {
+    it('should handle a synchronous throw from the next handler', (done: MochaDone) => {
       const expected = new Error('I told, you Bobby Boucher, threes are the debil!');
       const syncObservable = new Observable<number>((observer) => {
         observer.next(1);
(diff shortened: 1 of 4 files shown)


src/internal/Observable.ts
```diff
@@ -230,30 +231,120 @@ export class Observable<T> implements Subscribable<T> {
...
-   * @param {PromiseConstructor} [promiseCtor] a constructor function used to instantiate the Promise
+   * @param {PromiseConstructorLike|Subscription} [promiseCtorOrSubscription] a constructor function used to instantiate the Promise
    * @return {Promise} a promise that either resolves on observable completion or
    *  rejects with the handled error
    */
-  forEach(next: (value: T) => void, promiseCtor?: PromiseConstructorLike): Promise<void> {
+  forEach(next: (value: T) => void, promiseCtorOrSubscription?: PromiseConstructorLike|Subscription): Promise<void> {
+    let promiseCtor: PromiseConstructorLike;
+    let subs = isSubscription(promiseCtorOrSubscription) ? promiseCtorOrSubscription : undefined;
     promiseCtor = getPromiseCtor(promiseCtor);
```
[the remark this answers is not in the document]
benlesh: updated.

src/internal/util/isSubscription.ts
```diff
@@ -0,0 +1,9 @@
+import { Subscription } from '../Subscription';
+
+/**
+ * Tests to see if a value is an RxJS {@link Subscription}.
+ * @param x the value to test
+ */
+export function isSubscription(x: any): x is Subscription {
```
jsonberry: 👍

spec/Observable-spec.ts
```diff
@@ -125,21 +123,22 @@ describe('Observable', () => {
...
+      .then(
         () => {
-          throw new Error('should not be called');
-        },
-        (err) => {
-          results.push(err);
           // Since the consuming code can no longer interfere with the synchronous
           // producer, the remaining results are nexted.
-          expect(results).to.deep.equal([1, 2, 3, 4, expected]);
+          expect(results).to.deep.equal([1, 2, 3, expected]);
```
cartant: The change to this test's expectation suggests that the behaviour of the previous, non-cancellable implementation was actually a bug - as `next` was called despite an error being effected. The new behaviour makes more sense to me, as an error effected from `next` sees the observer unsubscribe.

Could the comment be removed? It's no longer applicable, as the remaining results are not nexted.

benlesh: Yup. I'll make sure to add the fix to the changelog too.

benlesh:
@cartant ... I'm having it reject with an `ObjectUnsubscribedError`... it seemed fitting and it already existed.

felixfbecker:
What I am worried about is forward compatibility with whatever cancellation primitive makes it into the spec. We already have AbortSignal in the DOM spec, and there is ongoing work to get a primitive into ES, which I would love love love to see supported seamlessly in rxjs.

benlesh:
> What I am worried about is forward compatibility with whatever cancellation primitive makes it into the spec

I'm not worried about this in the slightest. We can deprecate and pivot if we need to, but honestly, I have yet to see a proposal or cancellation type that isn't compatible with what we have here. The committees are slow, and unlikely to pass anything related to any of this. CancellationToken failed once. Observable is stuck at stage 1 forever.

Issue #3852, opened 2018-06-21
Add a deprecations page to the docs site


benlesh:
Basically, it would be great if we had a page of what is going to change in the (as of right now unplanned) upcoming major release of RxJS. The idea would be to find all `@deprecated` markers in the source and display information about what's being deprecated and why (we would be able to glean that from the deprecation message)

In a pinch, we could do this manually.

JWO719:
I'll take a look at it!

ladyleet:
@JWO719 any progress on this one?

JWO719:
Unfortunately not  yet. I have generated a JSON file with all the deprecated this, but I have to add a template to render some content

ladyleet:
@JWO719 👍

JWO719:
closed by #3981.... it could be improved to also display deprecated instance or overloaded methods

PR #3981, opened 2018-08-01
docs(deprecations): add deprecation page

JWO719:
This PR is related to #3852

That's how it looks like:
image

I faced some problems with members or overloaded functions that are deprecated. For now we don't even mention this in the dedicated doc page. I think changing this and adding it to the deprecation page could maybe done in an additional pr. But maybe we want to discuss that in detail.

Additionally I changed some of the methods that were marked as deprecated, but just because they were for internal usage, as this. This makes it not be visible in the documentation. That seems to be more logical for me. 

Generally the deprecation text could be improved.


docs_app/tools/transforms/angular-api-package/processors/generateDeprecationsListDoc.js
```diff
@@ -0,0 +1,28 @@
...
+    outputFolder: null,
+    $process: function(docs) {
+      docs.push({
+        docType: 'deprecation',
+        template: 'json-doc.template.json',
+        path: this.outputFolder + '/deprecations',
+        outputPath: this.outputFolder + '/deprecations.json',
+        data: docs
+          .filter(doc => {
+            console.log(doc.members);
```
benlesh: left over console.log

src/internal/AsyncSubject.ts
```diff
@@ -10,7 +10,7 @@ export class AsyncSubject<T> extends Subject<T> {
   private hasNext: boolean = false;
   private hasCompleted: boolean = false;
 
-  /** @deprecated This is an internal implementation detail, do not use. */
+  /** @internal This is an internal implementation detail, do not use. */
```
benlesh: I feel like we moved from internal to deprecated for some reason related to closure compiler or the like... cc @jasonaden

rkirov: TypeScript --stripInternal will remove @internal properties from the .d.ts which can affect downstream typechecking.

https://www.typescriptlang.org/docs/handbook/compiler-options.html

JWO719: ha interesting. Didn't knew about that. I'll fix that. Thanks for elaborating

Issue #3962, opened 2018-07-27
[Docs] Websocket is missing


JWO719:
For some reason the Websocket Observable isn't listed in the documentation. After a brief look into the source code I think there is some api documentation which could be used. I don't know if this might need some updates, but making it available again should be high prio.

Propably related to #3961

mattyao1984:
@JWO719 I am pushing a PR for this issue.

JWO719:
closed by #3984

Issue #3961, opened 2018-07-27
[Docs] Ajax is missing


JWO719:
For some reason the ajax Observable isn't listed in the documentation. After a brief look into the source code I think there is some api documentation which could be used. I don't know if this might need some updates, but making it available again should be high prio.

Propably related to #3962

JWO719:
I really appreciate your help. So basically there are already some api documentation in the source code but for some reasons it is just not shown in the docs website. At the moment I have no clue what could lead to that behaviour. I just could imagine that in the jsdoc comments there is something like @internal or something like that. But I really don't know at the moment.

mattyao1984:
I am pushing a PR for this issue.

JWO719:
closed by #3984

PR #3984, opened 2018-08-03
docs(ajax): added ajax and webSocket docs back to the list

mattyao1984:
**Description:**
- Added ajax docs back to the list
- Added webSocket docs back to the list

https://github.com/ReactiveX/rxjs/issues/3961
https://github.com/ReactiveX/rxjs/issues/3962


mattyao1984:
@cartant Thanks for the review! It seems the docs site only picks up the items from the primary location. I probably need to look at the other places to include them, if this is not the right place.

cartant:
@mattyao1984 No worries. There is one other import location, too: `'rxjs/testing'`

mattyao1984:
@cartant I updated by adding those sections so they can be generated into the `api-list.json` and get rendered into the template.  Also, I added a new example because there was none.

PR #3987, opened 2018-08-03
Revert "Cancellable forEach"

benlesh:
Reverts ReactiveX/rxjs#3977

Currently, there's an issue with what is in there if you unsubscribe from a synchronous observable, synchronously, but after the observable completes:

```ts
const subs = new Subscription();

of('test').forEach(x => console.log(x), subs)
  .then(() => console.log('this should not be called'));

subs.unsubscribe();

// Logs
// "test"
// "this should not be called"
```

Solutions around this aren't too bad, but it adds a lot of bloat, mostly due to the fact we're accepting PromiseConstructorLike as an argument, and we're beholden to use it.

I want to revert this new feature until:

1. We deprecate and remove all Promise Constructor configuration (This is legacy, and users should use a proper polyfill)
2. We deprecate and update the behavior of `forEach` to call it's `next` handler on the next microtask, as per the proposals. (We can't do this until v7 because it's a breaking change)

I still want to use Subscriptions as a cancellation here, but it's going to have to wait.

Sorry folks.

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -1,16 +1,18 @@
 import { expect } from 'chai';
 import * as sinon from 'sinon';
+import * as Rx from 'rxjs/Rx';
 import { Observer, TeardownLogic } from '../src/internal/types';
 import { cold, expectObservable, expectSubscriptions } from './helpers/marble-testing';
 import { map } from '../src/internal/operators/map';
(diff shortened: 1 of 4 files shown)


benlesh:
NOTE: We'll still need to redo the fix to `forEach` to make sure additional emissions aren't still happening after an error is thrown.

PR #3996, opened 2018-08-07
[WIP] chore(build): use importHelpers from TS instead of rollupInject

imcotton:
**Description:**

This follows up my [previous PR](https://github.com/ReactiveX/rxjs/pull/2121#issuecomment-260616205) on this aspect, now it's time to finish the second half.

- #2121

### DIFF
--- a/integration/systemjs/systemjs-compatibility-spec.js
+++ b/integration/systemjs/systemjs-compatibility-spec.js
@@ -2,7 +2,10 @@ var System = require('systemjs');
 var path = require('path');
 
 System.config({
-  map: { 'rxjs': path.join(__dirname, '..', '..', 'dist', 'package', '/') },
+  map: {
+    'rxjs': path.join(__dirname, '..', '..', 'dist', 'package', '/'),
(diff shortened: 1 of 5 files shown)


imcotton:
Now the build & test worked, but changes seem affect the **SystemJS** user, which requires to have `tslib` mapped in the config.

I'm not familiar with **SystemJS**, so if that's the case, shall this be treat as breaking change in next major release (v7)?

benlesh:
@jasonaden or @robwormald can you please have a look at this PR. You're needed for your Rollup and SystemJS expertise, thanks.

benlesh:
After some review and discussion, I think we need to close this. The UMD bundle is designed for standalone use, mostly for prototyping or for developers that don't care that they're including the world anyhow. This would be a breaking change, so we definitely can't do it in 6, and frankly I'm not sure we would do it in 7 either, given what the use case is for the umd bundle.

Thank you so much for the time you've put into this, @imcotton. I just don't think we're going to do this one. Sorry.

PR #4007, opened 2018-08-10
chore(TS): moves to use TypeScript@latest

benlesh:
Since we're doing CI against `typescript@latest` and `@next` anyhow, building against 2.8 locally doesn't make any sense.


cartant:
Bumping the TS version seems okay, but I think the Travis config will have to be updated to explicitly test against 2.8, as v6 was published as supporting that version.

Using an `@latest`-only feature would be a breaking change for many people.

PR #4012, opened 2018-08-13
chore(dtslint): disable no-unnecessary-generics

cartant:
**Description:**

Disables the [`no-unnecessary-generics`](https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md) TSLint rule that's applied to the dtslint tests. There are tests that deliberately involve generics that the rule thinks are unnecessary (in a `.d.ts` file context), but that's not the case.


benlesh:
Thanks, @cartant

Issue #4020, opened 2018-08-15
Migrate Timer Documentation from rxjs-docs repo


JWO719:
The Documentation for the timer function in the rxjs-docs repo has content, that should be migrated to the new docs in this repository.

I'll take care of it, I just created this issue to keep this in mind :D

cartant:
Closed by #4021, I guess.

PR #4021, opened 2018-08-15
docs(timer): migrate old docs repo for timer function

JWO719:
This merge request should update  the timer documentation, copy & pasted over from the old rxjs docs repo.

see #4020


src/internal/observable/timer.ts
```diff
@@ -38,14 +38,14 @@ import { Subscriber } from '../Subscriber';
  * @see {@link index/interval}
  * @see {@link delay}
  *
- * @param {number|Date} [dueTime] The initial delay time to wait before
- * emitting the first value of `0`.
+ * @param {number|Date} [dueTime] The initial delay time specified as a Date object or as an integer denoting
+      milliseconds between emissions to wait before emitting the first value of 0`.
```
cartant: I think "between emissions" should be removed, here.

JWO719: yep I think so too.

cartant: BTW, congratulations on joining the core team :tada:

JWO719: Thank you very much. I really appreciate this :smile:

src/internal/observable/timer.ts
```diff
@@ -38,60 +38,62 @@ import { Subscriber } from '../Subscriber';
...
+    } else if (isScheduler(periodOrScheduler)) {
+        scheduler = periodOrScheduler as any;
+    }
 
-  if (!isScheduler(scheduler)) {
-    scheduler = async;
-  }
+    if (!isScheduler(scheduler)) {
+        scheduler = async;
+    }
```
cartant: Hmm. In addition to deleting `"between emissions"` his last commit seems to have a whole bunch of whitespace/formatting changes. In particular, the indent size appears to have been changed from 2 spaces to 4, so whichever editor was used, it didn't pick up the `.editorconfig`.

JWO719: thanks for pointing, for some reasons VSCode is ignoring the editorconfig. I redo the changes and check it afterwards

cartant: Support for `.editorconfig` isn't built-in. VS Code needs an extension added to see it supported. Asking, there are still a bunch of other whitespace changes and there are also parentheses added. That's not necessarily a problem, but if these are automated changes, I'd like to figure out why they are happening and prevent it from happening again - if only because it complicates reviews. Small, isolated changes would be preferable.

cartant:
Support for `.editorconfig` isn't built-in. VS Code needs an extension added to see it supported. Also, there are still a bunch of other whitespace changes and there are expressions to which parentheses have been added. That's not necessarily a problem, but if these are automated changes, I'd like to figure out why they are happening and prevent it from happening again - if only because it complicates reviews. Small, isolated changes would be preferable.

JWO719:
You're totally right. I am currently switching over from webstorm to vscode, and for some reasons prettier was used for code formatting with some default settings. I will check today what's the problem here 😂 thanks again

JWO719:
So now it looks much better. Thanks again for pointing @cartant

PR #4022, opened 2018-08-15
docs(teams): revert deleting teams page

JWO719:
It seems like I deleted the teams page with my latest changes. So here's the revert for this


pertrai1:
👍

Issue #4027, opened 2018-08-16
Migrate from Operator Documentation from RxJS Docs Repo


JWO719:
In the rxjs-docs repo there is an improved documentation of the from operator (see: 
https://github.com/ReactiveX/rxjs-docs/blob/master/src/operator-docs/creation/from.ts).

Please migrate this to the new documentation by updating the JSDoc comments.

JWO719:
closed by #4031

PR #4031, opened 2018-08-17
docs(from): Convert docs for from

pertrai1:
**Description:**

Update the new docs for `from` with the older docs and add live StackBlitz example.

### DIFF
--- a/src/internal/observable/from.ts
+++ b/src/internal/observable/from.ts
@@ -12,6 +12,90 @@ import { ObservableInput, SchedulerLike } from '../types';
 
 export function from<T>(input: ObservableInput<T>, scheduler?: SchedulerLike): Observable<T>;
 export function from<T>(input: ObservableInput<ObservableInput<T>>, scheduler?: SchedulerLike): Observable<Observable<T>>;
+
+/**
+ * Creates an Observable from an Array, an array-like object, a Promise, an iterable object, or an Observable-like object.
+ *
+ * <span class="informal">Converts almost anything to an Observable.</span>
+ *
+ * ![](from.png)
+ *
+ * `from` converts various other objects and data types into Observables. It also converts a Promise, an array-like, or an
+ * <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#iterable" target="_blank">iterable</a>
+ * object into an Observable that emits the items in that promise, array, or iterable. A String, in this context, is treated
+ * as an array of characters. Observable-like objects (contains a function named with the ES2015 Symbol for Observable) can also be
+ * converted through this operator.
+ *
+ * ## Examples
+ * ### Converts an array to an Observable
+ * ```javascript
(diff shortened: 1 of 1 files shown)


src/internal/observable/from.ts
```diff
@@ -34,3 +34,74 @@ export function from<T>(input: ObservableInput<T>, scheduler?: SchedulerLike): O
 
   throw new TypeError((input !== null && typeof input || input) + ' is not observable');
 }
+
+/**
```
JWO719: Hi Rob, basically this is an awesome pr. Thanks for all the work you put into it. But this part has to be above the function declaration to be valid JSDoc. Maybe take a look at another operator like `fromEvent` then you will see what I mean.

src/internal/observable/from.ts
```diff
@@ -34,3 +34,74 @@ export function from<T>(input: ObservableInput<T>, scheduler?: SchedulerLike): O
...
+ * ```javascript
+ * import { from } from 'rxjs/observable/from';
+ *
+ * const array = [10, 20, 30];
+ * const result = from(array);
+ * result.subscribe(x => console.log(x));
+ * // Logs:
+ * // 10 20 30
+ * ```
+ * <a href="https://stackblitz.com/edit/js-tfvhag" target+"_blank">Stackblitz example</a>
```
JWO719: I really appreciate adding stackblitz, but first I think it doesn't really adds value for the small api examples. I think it would be more helpful for the tutorial, we want to add soon. Additionally this messes up the outline in the docs and it isn't conform with accessibility standards. Really thanks for this, but I think this makes sense to add in a seperate pull request

src/internal/observable/from.ts
```diff
@@ -34,3 +34,74 @@ export function from<T>(input: ObservableInput<T>, scheduler?: SchedulerLike): O
...
+ * result.subscribe(x => console.log(x));
+ *
+ * console.log('end');
+ *
+ * // Logs:
+ * // start end 10 20 30
+ * ```
+ *
+ * <a href="https://stackblitz.com/edit/js-kghdtb" target="_blank">StackBlitz example</a>
+ */
```
JWO719: could you add param describtion for the method signature and return value

JWO719:
BTW: This will close #4027

src/internal/observable/from.ts
```diff
@@ -12,6 +12,90 @@ import { ObservableInput, SchedulerLike } from '../types';
...
+ *
+ * // Logs:
+ * // start end 10 20 30
+ * ```
+ *
+ * @see {@link fromEvent}
+ * @see {@link fromEventPattern}
+ * @see {@link fromPromise}
+ *
+ * @param {ArrayLike<T> | ObservableInput<T>} A subscription object, a Promise, an Observable-like,
```
JWO719: ObserableInput is a type of `SubscribableOrPromise<T> | ArrayLike<T> | Iterable<T>` therefore adding `ArrayLike<T>` isn't really needed

PR #4034, opened 2018-08-18
fix(startWith): add empty signature

### LINKED ISSUE #3947: startWith missing an 'empty' signature
## Bug Report

The `startWith` operator includes a test for its being [called with only a scheduler](https://github.com/ReactiveX/rxjs/blob/6.2.2/spec/operators/startWith-spec.ts#L134-L141):

```ts
it('should start with empty if given value is not specified', () => {
  const e1 =   hot('-a-|');
  const e1subs =   '^  !';
  const expected = '-a-|';

  expectObservable(e1.pipe(startWith(rxTestScheduler))).toBe(expected);
});
```

However there is no associated [overload signature](https://github.com/ReactiveX/rxjs/blob/6.2.2/src/internal/operators/startWith.ts#L9-L17):

```ts
export function startWith<T>(v1: T, scheduler?: SchedulerLike): MonoTypeOperatorFunction<T>;
export function startWith<T>(v1: T, v2: T, scheduler?: SchedulerLike): MonoTypeOperatorFunction<T>;
export function startWith<T>(v1: T, v2: T, v3: T, scheduler?: SchedulerLike): MonoTypeOperatorFunction<T>;
export function startWith<T>(v1: T, v2: T, v3: T, v4: T, scheduler?: SchedulerLike): MonoTypeOperatorFunction<T>;
export function startWith<T>(v1: T, v2: T, v3: T, v4: T, v5: T, scheduler?: SchedulerLike): MonoTypeOperatorFunction<T>;
export function startWith<T>(v1: T, v2: T, v3: T, v4: T, v5: T, v6: T, scheduler?: SchedulerLike): MonoTypeOperatorFunction<T>;
export function startWith<T>(...array: Array<T | SchedulerLike>): MonoTypeOperatorFunction<T>;
```

**Possible Solution**

Add an overload signature that takes only a scheduler.

timdeschryver:
**Description:**
This PR adds an overload to `startWith` that only takes a scheduler.

Closes #3947

### DIFF
--- a/compat/operator/startWith.ts
+++ b/compat/operator/startWith.ts
@@ -2,13 +2,14 @@ import { Observable, SchedulerLike } from 'rxjs';
 import { startWith as higherOrder } from 'rxjs/operators';
 
 /* tslint:disable:max-line-length */
-export function startWith<T>(this: Observable<T>, v1: T, scheduler?: SchedulerLike): Observable<T>;
-export function startWith<T>(this: Observable<T>, v1: T, v2: T, scheduler?: SchedulerLike): Observable<T>;
-export function startWith<T>(this: Observable<T>, v1: T, v2: T, v3: T, scheduler?: SchedulerLike): Observable<T>;
-export function startWith<T>(this: Observable<T>, v1: T, v2: T, v3: T, v4: T, scheduler?: SchedulerLike): Observable<T>;
-export function startWith<T>(this: Observable<T>, v1: T, v2: T, v3: T, v4: T, v5: T, scheduler?: SchedulerLike): Observable<T>;
-export function startWith<T>(this: Observable<T>, v1: T, v2: T, v3: T, v4: T, v5: T, v6: T, scheduler?: SchedulerLike): Observable<T>;
-export function startWith<T>(this: Observable<T>, ...array: Array<T | SchedulerLike>): Observable<T>;
+export function startWith<T>(this: Observable<T>, scheduler?: SchedulerLike): Observable<T>;
+export function startWith<T, D = T>(this: Observable<T>, v1: D, scheduler?: SchedulerLike): Observable<T | D>;
+export function startWith<T, D = T, E = T>(this: Observable<T>, v1: D, v2: E, scheduler?: SchedulerLike): Observable<T | D | E>;
+export function startWith<T, D = T, E = T, F = T>(this: Observable<T>, v1: D, v2: E, v3: F, scheduler?: SchedulerLike): Observable<T | D | E | F>;
+export function startWith<T, D = T, E = T, F = T, G = T>(this: Observable<T>, v1: D, v2:  E, v3: F, v4: G, scheduler?: SchedulerLike): Observable<T | D | E | F | G>;
+export function startWith<T, D = T, E = T, F = T, G = T, H = T>(this: Observable<T>, v1: D, v2: E, v3: F, v4: G, v5: H, scheduler?: SchedulerLike): Observable<T | D | E | F | G | H>;
+export function startWith<T, D = T, E = T, F = T, G = T, H = T, I = T>(this: Observable<T>, v1: D, v2: E, v3: F, v4: G, v5: H, v6: I, scheduler?: SchedulerLike): Observable<T | D | E | F | G | H | I>;
+export function startWith<T, D = T>(this: Observable<T>, ...array: Array<D | SchedulerLike>): Observable<T | D>;
 /* tslint:enable:max-line-length */
 
 /**
@@ -25,6 +26,6 @@ export function startWith<T>(this: Observable<T>, ...array: Array<T | SchedulerL
  * @method startWith
  * @owner Observable
  */
-export function startWith<T>(this: Observable<T>, ...array: Array<T | SchedulerLike>): Observable<T> {
-  return higherOrder(...array)(this);
+export function startWith<T, D>(this: Observable<T>, ...array: Array<D | SchedulerLike>): Observable<T |D> {
+  return higherOrder<T, D>(...array)(this);
 }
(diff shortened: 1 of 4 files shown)


timdeschryver:
I totally missed those, thanks for pointing them out Nicholas 😅 .
You can expect a commit later today/tomorrow.

timdeschryver:
Something like this this, right?
Should I add a test for each type definition?

cartant:
Yeah, that's great.

If you want, you can add a test for each signature, but as all of the value types are of type `T`, I don't think it's that big a deal. `pipe` is different, as the parameters chain together.

If you want to add another, maybe you could add one to show that a scheduler can be passed as the last parameter?

spec-dtslint/operators/startWith-spec.ts
```diff
@@ -0,0 +1,30 @@
...
+it('should infer correctly with a value and a scheduler', () => {
+  const o = of(1, 2, 3).pipe(startWith(5, asyncScheduler)); // $ExpectType Observable<number>
+});
+
+it('should infer correctly with only a scheduler', () => {
+  const o = of(1, 2, 3).pipe(startWith(asyncScheduler)); // $ExpectType Observable<number>
+});
+
+it('should enforce types with one value', () => {
+  const o = of(1, 2, 3).pipe(startWith('foo')); // $ExpectError
```
benlesh: I feel like this should be `// $ExpectType Observable<number|string>`

spec-dtslint/operators/startWith-spec.ts
```diff
@@ -0,0 +1,30 @@
...
+});
+
+it('should enforce types with multiple value', () => {
+  const o = of(1, 2, 3).pipe(startWith(4, 'foo')); // $ExpectError
```
benlesh: Same as above, although this might not be possible in this case without the overloads getting out of control.

src/internal/operators/startWith.ts
```diff
@@ -7,6 +7,7 @@ import { isScheduler } from '../util/isScheduler';
 import { MonoTypeOperatorFunction, SchedulerLike } from '../types';
 
 /* tslint:disable:max-line-length */
+export function startWith<T>(scheduler?: SchedulerLike): MonoTypeOperatorFunction<T>;
```
benlesh: I think we're going to need to differentiate between what's being passed into the `startWith` and what the source observable is to really get this right. :thinking: 

```ts
export function startWith<T, R>(...args: R[], scheduler?: SchedulerLike): OperatorFunction<T, T|R>;
```

cartant: @benlesh Yeah, you are right. This is very similar to the changes that were made regarding the default parameters in `first` and `last` - in https://github.com/ReactiveX/rxjs/pull/3952/commits/04f5f0b3499e0c3fdf4088d5fbe8f90d534c6480 - that effect a union type if a non-`T` default is specified.

timdeschryver: AFAICT this won't be work because it's only possible to define a rest parameter as the last argument in a function. 
I think it should be doable if we solve it the same way as in the commit @cartant mentioned.

cartant: @timdeschryver You'd need to use a union type as the rest parameter type. Have a look at [`concat`](https://github.com/ReactiveX/rxjs/blob/6.2.2/src/internal/observable/concat.ts#L15).

cartant: I just noticed that there is [already a rest parameters signature](https://github.com/ReactiveX/rxjs/blob/6.2.2/src/internal/operators/startWith.ts#L16). I think the `T` in the signatures should be replaced with an `R` - where `R` defaults to `T`, as in the referenced commit. And the return type should be `OperatorFunction<T, T | R>` which will collapse to `OperatorFunction<T, T>` when `R` is `T`.

timdeschryver: Yea, I agree that's what I have right now.
The only difference is that I have a type for each argument, I'll commit and push so you can take a look and give your opinions on it.

cartant: Yeah, I guess if `R` doesn't have to be `T`, having one type parameter per parameter makes sense.

src/internal/operators/startWith.ts
```diff
@@ -49,7 +49,7 @@ export function startWith<T>(...array: Array<T | SchedulerLike>): MonoTypeOperat
  * @method startWith
  * @owner Observable
  */
-export function startWith<T>(...array: Array<T | SchedulerLike>): MonoTypeOperatorFunction<T> {
+export function startWith<T, D>(...array: Array<T | SchedulerLike>): OperatorFunction<T, T | D> {
```
cartant: I think the rest parameter signature should be made the same as the implementation signature. And I think both should be:

```ts
export function startWith<T, D = T>(...array: Array<D | SchedulerLike>): OperatorFunction<T, T | D>
```

timdeschryver: Yea I'll add it, but I didn't had the time to adjust and test it. Currently a bit busy, but I'll add it soon.

compat/operator/startWith.ts
```diff
@@ -26,5 +26,5 @@ export function startWith<T>(this: Observable<T>, ...array: Array<T | SchedulerL
  * @owner Observable
  */
 export function startWith<T>(this: Observable<T>, ...array: Array<T | SchedulerLike>): Observable<T> {
-  return higherOrder(...array)(this);
+  return higherOrder<T, T>(...array)(this);
```
timdeschryver: Is this OK 😅 ?

cartant: I seem to always forget the `compat` signatures. They should be updated, too.

I need to make myself a checklist of things to do upon making a typings change.

timdeschryver: Done, thanks for your help!

felixfbecker:
I love that this finally allows starting with a different type than `T`. I often need to write code like
```ts
values.pipe(
  switchMap(value =>
    fetchSuggestions(value).pipe(
      startWith('LOADING')
      catchError(err => [err])
    )
  )
)
```
Which should be `Observable<'LOADING' | Suggestions | Error>`, but it was not possible to do before with `startWith` (only with `concat()`).

Issue #4029, opened 2018-08-16
Migrate groupBy Documentation from RxJS Docs Repository


JWO719:
In the rxjs-docs repo there is an improved documentation of the groupBy operator (see: https://github.com/ReactiveX/rxjs-docs/blob/master/src/operator-docs/transformation/groupBy.ts).

Please migrate this to the new documentation by updating the JSDoc comments.

JWO719:
I think a real assignment isn't needed. But glad that you take care of it. Thank you very much

pertrai1:
@JWO719 is the JSDoc comments here the same as what you are referring to? https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/groupBy.ts or am I looking in the wrong place?

JWO719:
That's the right file? Why are there any problems`?

JWO719:
closed by #4036

PR #4036, opened 2018-08-18
docs(groupBy): Update to groupBy docs

pertrai1:
**Description:**

Modifications that update the docs for `groupBy`  from the older docs to the newer.

related issue: https://github.com/ReactiveX/rxjs/issues/4029

### DIFF
--- a/src/internal/operators/groupBy.ts
+++ b/src/internal/operators/groupBy.ts
@@ -19,23 +19,34 @@ export function groupBy<T, K, R>(keySelector: (value: T) => K, elementSelector?:
  *
  * ![](groupBy.png)
  *
- * ##Examples
- * Group objects by id and return as array
- * ```typescript
+ * When the Observable emits an item, a key is computed for this item with the keySelector function.
+ *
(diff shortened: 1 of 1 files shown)


src/internal/operators/groupBy.ts
```diff
@@ -19,53 +19,70 @@ export function groupBy<T, K, R>(keySelector: (value: T) => K, elementSelector?:
...
  *   groupBy(p => p.id),
  *   mergeMap((group$) => group$.pipe(reduce((acc, cur) => [...acc, cur], []))),
  * )
  * .subscribe(p => console.log(p));
  *
  * // displays:
- * // [ { id: 1, name: 'aze1' },
- * //   { id: 1, name: 'erg1' },
- * //   { id: 1, name: 'df1' } ]
+ * // [ { id: 1, name: 'javascript'},
```
JWO719: We try to have short api examples, so that you get the way this operator works as fast as possible. Maybe you can shorten it a bit

pertrai1: @JWO719 I have shorted and fixed the first example as well.

Issue #4025, opened 2018-08-16
`defer` operator inner-observable execute after the related outer-observable has been unsubscribed from


peaBerberian:
## Bug Report

**Current Behavior**

Deferred inner-observables continue to execute even when the outer observable it depends on has no subscription left. This was seen only with a non-asynchronous Scheduler.
 
Please forgive me if it's not an issue in your opinion, but after giving it some thought I don't really see any advantage to this behavior.

**Reproduction**

Running:
```js
const {
  Subject,
  of: obsOf,
  concat: obsConcat,
  defer,
} = require("rxjs");
const {
  finalize,
  mergeMap,
  tap,
  takeUntil,
} = require("rxjs/operators");

const subject = new Subject();

obsOf(null).pipe(
  mergeMap(() =>
    obsConcat(
      defer(() => {
        console.log("side-effect 1");
        return obsOf(1);
      }),
      defer(() => {
        console.log("side-effect 2");
        return obsOf(2);
      }),
      defer(() => { // This should never be run
        console.log("side-effect 3");
        return obsOf(3);
      })
    )
  ),
  finalize(() => {
    console.log("finalized");
  })
)
.pipe(
  takeUntil(subject),
  tap((i) => {
    if (i === 2) { // Stop subscribing when `2` is emitted
      subject.next();
    }
  })
).subscribe(
  (i) => { console.log("next", i); },
  (e) => { console.log("error", e); },
  () => { console.log("complete"); },
);
```
Outputs:
```sh
side-effect 1
next 1
side-effect 2
complete
finalized
side-effect 3
```
As you can see, the `side-effect 3` line is displayed even after the outer-observable has no subscription left.

**Expected behavior**

I would expect that any deferred observable running after unsubscription would not be run.

That is, obtaining the following output from the previous code:
```sh
side-effect 1
next 1
side-effect 2
complete
finalized
```
_(no `side-effect 3` line)_

As the item(s) emitted from those observables would never be used, I don't see any advantage to "running" them.

Moreover, using something like an AsapScheduler do not run the deferred observable. 

**Environment**
- Runtime: Node v10.8.0
- RxJS version: 6.2.2

cartant:
It's possible that this could have been fixed in https://github.com/ReactiveX/rxjs/pull/3963. Would you be able to investigate whether or not the problem is reproducible with the current `master`?

peaBerberian:
It seems to have been fixed on master, yes.
Thanks a lot!

[...]

PR #4037, opened 2018-08-19
Fix teardown issues with synchronous inner-observables

peaBerberian:
This PR fixes teardown issues that arised when unsubscribing from an observable emitting values from inner-observable(s) synchronously after subscription, like reported in #4025. An even simpler example is available [here](https://stackblitz.com/edit/typescript-ecmb6g?file=index.ts&devtoolsheight=50).
 Basically, the problem is that even after the unsubscription, the inner-observable(s) will continue to run until an asynchronous step is reached.

 This was simply because most ``OuterSubscriber``s first call ``subscribeToResult`` on an inner-Observable and THEN call ``this.add`` on the returned Subscription. This sounds fair, but when we're dealing with an unsubscription happening synchronously after the subscription, ``this.add`` will not have been called yet, and so the teardown will happen without unsubscribing from those inner-observables.

 A fix has already been brought in https://github.com/ReactiveX/rxjs/commit/40852ff71, but only for the ``MergeMapSubscriber``. In this fix, an ``InnerSubscriber`` is created, registered through ``this.add`` and then given as a last argument to ``subscribeToResult``, which will use it instead of creating one.
 It looks like it worked well for ``mergeMap`` and other related operators for some time now, so I brought the same fix for the following ``OuterSubscriber``s (I linked an example showing problematic real use-cases for each of them):
   - [CatchSubscriber](https://stackblitz.com/edit/typescript-p9ygrd?file=index.ts&devtoolsheight=50)
   - [ExhaustMapSubscriber](https://stackblitz.com/edit/typescript-oesxqp?file=index.ts&devtoolsheight=50)
   - [MergeScanSubscriber](https://stackblitz.com/edit/typescript-mbp98x?file=index.ts&devtoolsheight=50)
   - [OnErrorResumeNextSubscriber](https://stackblitz.com/edit/typescript-hpfqqz?file=index.ts&devtoolsheight=50)
   - [SkipUntilSubscriber](https://stackblitz.com/edit/typescript-etcev1?file=index.ts&devtoolsheight=50)
   - [SwitchMapSubscriber](https://stackblitz.com/edit/typescript-ecmb6g?file=index.ts&devtoolsheight=50)

 I was hesitant to also bring the same fix to other OuterSubscribers with the same problems.

 For example, with combineLatest the following code:
 ```js
 combineLatest(
   of(1), // should be subscribed to
   throwError(new Error()), // error! -> Teardown
   defer(() => { // should not run
     console.log("side-effect 1");
     return of(2).pipe(
       tap(() => console.log("side-effect 2")) // just to insist a little
     );
   })
 ).subscribe(
   (i) => { console.log("next", i); },
   (e) => { console.log("error!"); },
   () => { console.log("complete"); },
 );
 ```
 outputs:
 ```sh
 error
 side-effect 1
 side-effect 2
 ```
 instead of just:
 ```sh
 error
 ```

 But here, it's not that problematic, considering that we can just think that every Observable given to a CombineLatest run at the same time (or at least, that's a mental model I have when I'm using it).
 Still, code run for nothing there and the code logic looks wrong. What do you think we should do?

 The same problematic logic ("adding" the return of a ``subscribeToResult``) is also written for many other ``OuterSubscriber``s.
 Namely for the ``BufferSubscriber``, ``BufferToggleSubscriber``, ``DistinctSubscriber`` (for its ``flushes`` argument, but unsubscribing/erroring from that observable just completes/errors the observable its applied on), ``ExpandSubscriber``, ``SwitchFirstSubscriber``, ``ThrottleSubscriber``, ``WithLatestFromSubscriber``, ``WindowSubscriber`` and ``WindowToggleSubscriber`` but I did not yet succeed to reproduce a real use-case with any of them where that would be problematic. For this reason, I did not update the logic there.

### DIFF
--- a/spec/operators/catch-spec.ts
+++ b/spec/operators/catch-spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
-import { concat, Observable, of, throwError, EMPTY, from } from 'rxjs';
-import { catchError, map, mergeMap } from 'rxjs/operators';
+import { concat, defer, Observable, of, throwError, EMPTY, from } from 'rxjs';
+import { catchError, map, mergeMap, takeWhile } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import * as sinon from 'sinon';
 import { createObservableInputs } from '../helpers/test-helper';
@@ -119,6 +119,31 @@ describe('catchError operator', () => {
     expectObservable(result, unsub).toBe(expected);
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
     expectSubscriptions(e2.subscriptions).toBe(e2subs);
+  });
+
+  it('should stop listening to a synchronous observable when unsubscribed', () => {
+    const sideEffects: number[] = [];
+    const synchronousObservable = concat(
+      defer(() => {
+        sideEffects.push(1);
+        return of(1);
+      }),
+      defer(() => {
+        sideEffects.push(2);
+        return of(2);
+      }),
+      defer(() => {
+        sideEffects.push(3);
+        return of(3);
+      })
+    );
+
+    throwError(new Error('Some error')).pipe(
+      catchError(() => synchronousObservable),
+      takeWhile((x) => x != 2) // unsubscribe at the second side-effect
+    ).subscribe(() => { /* noop */ });
+
+    expect(sideEffects).to.deep.equal([1, 2]);
   });
 
   it('should catch error and replace it with a hot Observable', () => {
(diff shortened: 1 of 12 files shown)


cartant:
Thanks for the PR. This looks good. I'll review it tomorrow.

benlesh:
This LGTM. I'll wait for a second opinion from @cartant or @kwonoj before I merge.

kwonoj:
I recall few places I added checker for return value of `subscribeToResult` for synchronous, as sync case it can complete and does not return any values for it. Do we possibly need audit same to check / guard around those for new changes? or it's logically prevented?

benlesh:
An important PR. Thanks, @peaBerberian

PR #4040, opened 2018-08-20
By example with array extends it with '...'

markstein:
**Description:**


kwonoj:
Would you mind elaborate bit more? if use spread, it doesn't match with the comment `// note that array is passed` since it no longer passes array.

pertrai1:
@markstein please have a look at the documentation for pull requests in regards to commit messages: https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md#commit-message-guidelines

markstein:
I'm not sure if that change is correct. But this doc example don't work for me.

Issue #3895, opened 2018-07-05
[Discussion] Add Blog to Documentation


JWO719:
I think it would be useful to add a blog-like page to the docs with rxjs related blogs.

I see 2 possible solutions for this:

1. Like angular.io we could link to a medium account. 
pro: 
              -  it is a well known blog platform with a lot of features like comments
              - not that much work to setup
con:
              - not that flexible for using code
              - share account details
              - I don't know if it is possible to add a review process before an article is published

2. We could add a new site to the docs, where we reference pages which could be inside the docs_app or across the internet.
pro:
               - more or less easy to prepare
               - kinda flexible with code and other stuff
               - easy to find content to start with
               - review could be done easily via github
cons:
              - no(not easily) blog-like features like comments
              - linked pages could be remove and therefore we reference a dead page or the content changes in a way not compliant to the rxjs code of  conduct.

Any opinions or other solutions?

pertrai1:
@JWO719 is there any way that we can make use of https://github.com/ReactiveX/reactivex.github.io ?

JWO719:
Could you please elaborate what you have in mind, I actually don't get it

pertrai1:
Seeing that it used Jekyll to create the site, I thought we could look over the structure of how the content is being delivered. Once that is done, it could have a section devoted strictly towards RxJS content. One of the pros that this checks off is the use of code reviews for any content that would be added. I think it is very flexible yet also know that it is not the docs site right now. 

Throwing out ideas as they come to me.

JWO719:
I think we can do pretty much the same with dgeni what we are currently using. Therefore I don't see a benefit in using jekyll, or maybe I just don't get it. Theoretically authors could just file a pr with an md in the content folder and this would generate an html page in it. Then we would have the content, but I don't think that many authors will do that. Maybe it would rather be an option for rxjs contributors to use that way. But I'd still prefer a json file with the references. Something like: 
```json
[{
  "title": "some awesome title",
  "author": "some awesome author",
  "level": "beginner",
  "link": "....",
  "tags": ["rxjs", "testing", "dunno", ....],
  "releasedAt": "some date"
}]
```
Those have to be added manually, but I think it would be pretty straightforward to do so and also we could provide search and filter functions easily. Having a component which loads and renders this json file would come with easy unit tests and so on...

pertrai1:
[...]

PR #4045, opened 2018-08-21
docs(resources): Add New Section in Docs for External Resources

pertrai1:
**Description:**

This adds a new section on the left hand navigation for External Resources. It is a starting collection of resources that people can use to find more content available on the web. Hopes are that it grows over time to be the one place people can come and find everything they need for information regarding RxJS.

### DIFF
--- a/docs_app/content/external-resources.md
+++ b/docs_app/content/external-resources.md
@@ -1,0 +1,13 @@
+# External Resources
+
+This is a collection of resources that can be found to help with the learning process of RxJS. The list is ever growing which will make it a place for you to come to for finding helpful information.
+
+---
+
+* [https://www.learnrxjs.io/](https://www.learnrxjs.io/)
+* [https://www.academind.com/learn/javascript/rxjs-6-what-changed/]()
+* [Introducing RxJS! - Ben Lesh - YouTube](https://www.youtube.com/watch?v=JCXZhe6KsxQ)
+* [Learn RxJS in 60 Minutes for Beginners - YouTube](https://www.youtube.com/watch?v=PhggNGsSQyg)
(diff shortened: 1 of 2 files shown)


mattyao1984:
I guess this is a new feature to the doc website and it looks awesome. My suggestion is to have several categories for the resources such as videos, tutorials, blogs etc.

pertrai1:
@mattyao1984 that sounds like a great idea. going off of the categories, which ones do you see being needed for this to be a good resource to start with? I want to get this off the ground and then have smaller PR's towards it. If that does not sound good, I can work on it for a bit more until I find enough resources

JWO719:
Hi @pertrai1 basically this is an awesome PR, thank you very much. Personally I would prefer something similar to the way we render the teams page. There we have a json file for the data and an angular component loading the json file and rendering the content. I think this is far more complex, but it also comes with some benefits. It would be easier to add resources and functions like searching or filtering. Also it would be testable and maybe easier to understand for other contributors. I know that this would be need more effort and therefore I would really like to discuss on that. What do you think of that?

docs_app/content/external-resources.md
```diff
@@ -0,0 +1,13 @@
+# External Resources
+
+This is a collection of resources that can be found to help with the learning process of RxJS. The list is ever growing which will make it a place for you to come to for finding helpful information.
```
JWO719: Maybe add a hint here, that RxJS isn't responsible for the content of the sites or something like that

pertrai1:
@JWO719 I like your idea and I will take a look at how Teams is constructed. How about closing this PR and we can take up discussion on how best to put together a library of resources for users?

JWO719:
maybe let's discuss this here -> #3895

PR #4047, opened 2018-08-21
docs(interval): more practical example

pertrai1:
**Description:**

Updating the example to add import(s) into the example as is being done in other docs, as well as a cleaner usage.

### DIFF
--- a/src/internal/observable/interval.ts
+++ b/src/internal/observable/interval.ts
@@ -21,10 +21,22 @@ import { Subscriber } from '../Subscriber';
  * {@link SchedulerLike} to it.
  *
  * ## Example
- * Emits ascending numbers, one every second (1000ms)
+ * Emits ascending numbers, one every second (1000ms) up to the number 3
  * ```javascript
+ * import { interval } from 'rxjs';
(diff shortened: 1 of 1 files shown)


pertrai1:
LOL. Just found out that the example is pretty much the same that is being used for `take`. https://rxjs-dev.firebaseapp.com/api/operators/take

src/internal/observable/interval.ts
```diff
@@ -21,10 +21,22 @@ import { Subscriber } from '../Subscriber';
...
- * Emits ascending numbers, one every second (1000ms)
+ * Emits ascending numbers, one every second (1000ms) up to the number 3
  * ```javascript
+ * import { interval } from 'rxjs';
+ * import { take } from 'rxjs/operators';
+ * 
  * const numbers = interval(1000);
- * numbers.subscribe(x => console.log(x));
+ *
+ * const takeThreeNumbers = numbers.pipe(take(4));
```
JWO719: isn't it `takeFourNumbers` :smile:

pertrai1: LOL. good catch

PR #4050, opened 2018-08-23
fix(subscribable): make subscribe() signature match Observable

### LINKED ISSUE #3891: Observable<never> is not assignable to Subscribable<T>
## Bug Report

**Current Behavior**

Kinda related to #3890, when trying implement interop by accepting `Subscribable` as input, you cannot pass `Observable<never>` into any `Subscribable<T>`.
This is because `Subscribable`'s signature for `subscribe` doesn't match that of `Observable`:
```ts
    subscribe(observerOrNext?: PartialObserver<T> | ((value: T) => void), error?: (error: any) => void, complete?: () => void): Unsubscribable;
```
vs
```ts
    subscribe(observer?: PartialObserver<T>): Subscription;
    subscribe(next?: (value: T) => void, error?: (error: any) => void, complete?: () => void): Subscription;
```

So you get this error:
```
Argument of type 'Observable<never>' is not assignable to parameter of type 'Subscribable<number>'.
  Types of property 'subscribe' are incompatible.
    Type '{ (observer?: NextObserver<never> | ErrorObserver<never> | CompletionObserver<never> | undefined)...' is not assignable to type '(observerOrNext?: NextObserver<number> | ErrorObserver<number> | CompletionObserver<number> | ((v...'.
      Types of parameters 'observer' and 'observerOrNext' are incompatible.
        Type 'NextObserver<number> | ErrorObserver<number> | CompletionObserver<number> | ((value: number) => v...' is not assignable to type 'NextObserver<never> | ErrorObserver<never> | CompletionObserver<never> | undefined'.
          Type '(value: number) => void' is not assignable to type 'NextObserver<never> | ErrorObserver<never> | CompletionObserver<never> | undefined'.
            Type '(value: number) => void' is not assignable to type 'CompletionObserver<never>'.
              Property 'complete' is missing in type '(value: number) => void'.
```

**Reproduction**
```ts
import { Observable, Subscribable, NEVER } from 'rxjs'

function foo(input: Subscribable<number>) {}

foo(NEVER)
```

**Expected behavior**

An Observable that `never` emits should be assignable to any `Observable` of any type, since the type doesn't matter if it never emits. This is the semantic of `never`.

**Environment**
- Runtime: any
- RxJS version: 6.2.1

**Possible Solution**
Change the signature of `Subscribable.subscribe` to use an overload, which in my eyes more accurately represents the actual interface. 
With the current signature, this is possible, which it shouldn't:
```ts
subscribable.subscribe({ error: err => console.log('err 1') }, err => console.log('err 2'))
```
What would get logged?

felixfbecker:
**Description:**
The signature of `Subscribable.subscribe()` did not match that of `Observable.subscribe()`, which made the two types not compatible if `T=never` (see linked issue).

It also was wrong, because it allowed passing both an Observer and functions at the same time. 

This makes it match the signature of `Observable.subscribe()` exactly, which solves the type error in the issue.

It is not breaking (unless you passed both observer and functions, which is extremely unlikely and did not work at runtime anyway as `SafeSubscriber` checks if the first argument is a function and if yes, discards the other arguments, so this would only highlight a bug).

The alternative discussed in the issue was removing the 3-arg overload, but the discussion in the spec does not seem resolved yet, it is not clear that the interface is supposed to represent the spec, it would be a breaking change and require way more changes in rxjs (it did not compile when I tried).

Fixes #3891

cc @cartant

### DIFF
--- a/spec/operators/zipAll-spec.ts
+++ b/spec/operators/zipAll-spec.ts
@@ -39,7 +39,7 @@ describe('zipAll operator', () => {
       of('a', 'b', 'c'),
       of(1, 2, 3)
     )
-    .pipe(zipAll((a, b) => a + b))
+    .pipe(zipAll((a, b) => String(a) + String(b)))
     .subscribe((x) => {
       expect(x).to.equal(expected[i++]);
     }, null, done);
@@ -378,7 +378,7 @@ describe('zipAll operator', () => {
       of('a', 'b', 'c'),
       of(1, 2)
     )
-    .pipe(zipAll((a, b) => a + b))
+    .pipe(zipAll((a, b) => String(a) + String(b)))
     .subscribe((x) => {
       expect(x).to.equal(expected[i++]);
     }, null, done);
--- a/spec/operators/zipAll-spec.ts
+++ b/spec/operators/zipAll-spec.ts
@@ -231,7 +231,7 @@ describe('zipAll operator', () => {
       const b = [4, 5, 6];
       const expected = '---x--#';
 
-      const selector = function (x: number, y: number) {
+      const selector = function (x: string, y: number) {
         if (y === 5) {
           throw new Error('too bad');
         } else {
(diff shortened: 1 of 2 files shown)


felixfbecker:
I wish rxjs could just get rid of all the unneeded `project` functions

PR #4051, opened 2018-08-23
Pass AbortSignal to callbacks passed to defer(), switchMap() etc

felixfbecker:
**Description:**
This is a small PoC for my proposal outlined in https://github.com/ReactiveX/rxjs/issues/3122#issuecomment-382096053
Initially for `defer()` and `switchMap()`.
`AbortControllers` only get instantiated if available in the environment so nothing breaks if they are not polyfilled.

Note that I am creating `AbortController`s in the operator implementations - ideally those would be passed to the callback of the `Observable` constructor, but that was more effort to implement. My goal was just to create a PoC and keep the discussion going as this has been an ongoing pain point.

https://github.com/ReactiveX/rxjs/issues/3122#issuecomment-382096053


felixfbecker:
> Considering right now AbortSignal isn't widely used or supported

Are you referring to browser support? `AbortSignal` is supported by all modern browsers: 

image

And it's already used by `fetch` and the WebAuthentication API (https://github.com/w3c/webauthn/pull/544)

It's not supported in NodeJS of course, but I would take the same stance as with Promises - you can polyfill if you want to use it.

For the future, Node has shipped DOM APIs before - e.g. WHATWG `URL`

> given that Subscription is basically both an AbortController and an AbortSignal.

I think this ascpect makes Subscription basically unusable for a general-purpose cancellation signal. If I pass a signal to a function, that function should have the ability to fork the token and cancel sub-operations, but it should never be able to cancel the operations of the caller.

> In the experimental branch, I'm pasing Subscription as a cancellation token of sorts to methods like forEach and toPromise. I hadn't thought about defer, etc.

For me, `defer`, `switchMap` etc. are actually the more important use cases, because I see a lot that people want to do single-result Promise operations for every event in an Observable chain (e.g., call the `fetch` API or an API wrapper that returns a single result in reaction to events). Only rarely do you need to use a (multi-value) Observable from within an async function.

PR #4054, opened 2018-08-24
fix(3699): Amend isNumeric typings and add unit test

dkosasih:
**Description:** isNumber type predicate should narrow to nnumber | string when return true

**Related issue (if exists):** #3699


dkosasih:
@cartant I have changed the commit message to chore.
please advice if there is anything else I need to be aligned.

PR #4056, opened 2018-08-27
test(dtslint): add switchMap

timdeschryver:
**Description:**
I'm on a quest to learn the insides of RxJS and its operators. 
While doing so I could add these kind of dtslint tests as I go, at least if this is OK for you 😄 .

### DIFF
--- a/spec-dtslint/operators/switchMap-spec.ts
+++ b/spec-dtslint/operators/switchMap-spec.ts
@@ -1,0 +1,38 @@
+import { of } from 'rxjs';
+import { switchMap } from 'rxjs/operators';
+
+it('should infer correctly', () => {
+  const o = of(1, 2, 3).pipe(switchMap(p => of(Boolean(p)))); // $ExpectType Observable<boolean>
+});
+
+it('should support a projector that takes an index', () => {
+  const o = of(1, 2, 3).pipe(switchMap((p, index) => of(Boolean(p)))); // $ExpectType Observable<boolean>
+});
+
+it('should infer correctly by using the resultSelector first parameter', () => {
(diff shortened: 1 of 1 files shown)


spec-dtslint/operators/switchMap-spec.ts
```diff
@@ -0,0 +1,34 @@
...
+  const o = of(1, 2, 3).pipe(switchMap(p => of(Boolean(p)), (a, b, innnerIndex, outerIndex) => a)); // $ExpectType Observable<number>
+});
+
+it('should support an undefined resultSelector', () => {
+  const o = of(1, 2, 3).pipe(switchMap(p => of(Boolean(p)), undefined)); // $ExpectType Observable<boolean>
+});
+
+it('should enforce types', () => {
+  const o = of(1, 2, 3).pipe(switchMap()); // $ExpectError
+});
```
cartant: Maybe add another to ensure that the value returned from the callback has to be an `ObservableInput`:

it('should enforce the return type', () => {
  const o = of(1, 2, 3).pipe(switchMap(n => n)); // $ExpectError
});

timdeschryver:
Thanks for the info @cartant !
I've added the extra check.

cartant:
Thanks.

This is a small PR, so it's easy to see what you've changed since I commented, but something I mentioned to another contributor, earlier today, also applies here - so I'm copying and pasting:

> Reviews are much easier if instead of amending a commit, you just add another commit to the PR to address whatever changes were requested. If you do that, reviewers need only look at the post-request changes. The commits will be squashed when the PR is merged, so you can add as many as you like.

Thanks again for the PR.

PR #4057, opened 2018-08-27
docs(external resources): First version of external resources

pertrai1:
**Description:**

This is a starting boilerplate to introduce external resources that will
be useful for users when looking for other places to learn. The
boilerplate is rough draft with the goal of figuring out the best way to
manage an up to date list of resources.

I don't want to sit on this yet at the same time have to be careful of
making sure that it is maintained and kept up to date. It can very
quickly become out of date if the content no longer is relevant to the
current version of RxJS.

Thoughts are welcomed. There have been talks already started about
creating a more dynamic resource, similar to Teams. I would like to get
information in front of people in quicker iteration and build that out
on a separate branch using the data that will continue to populate this
markdown file.

References #3895


docs_app/content/navigation.json
```diff
@@ -54,7 +54,12 @@
 			"url": "code-of-conduct",
 			"title": "Code of Conduct",
 			"tooltip": "Code of Conduct"
-		}
+    },
+    {
+      "url": "external-resources",
```
JWO719: I think resources would be a cleaner url

docs_app/content/navigation.json
```diff
@@ -54,7 +54,12 @@
 			"url": "code-of-conduct",
 			"title": "Code of Conduct",
 			"tooltip": "Code of Conduct"
-		}
+    },
```
JWO719: And could you switch this with the code of conduct. I would like to have it at the very bottom

docs_app/content/external-resources.md
```diff
@@ -0,0 +1,44 @@
...
+
+**Note**: RxJS is not responsible for the content of the sites listed below unless directly from the RxJS site.
+
+---
+
+## Blogs
+
+### 2018
+
+* Angular 6: Upgrading API calls to RxJS 6 | [Metal Toad](https://www.metaltoad.com/blog/angular-6-upgrading-api-calls-rxjs-6) | May 23, 2018
```
JWO719: Could you add https://blog.strongbrew.io/the-sip-principle/ here

docs_app/content/navigation.json
```diff
@@ -50,6 +50,11 @@
 				}
 			]
 		},
+		{
+      		"url": "resources",
+			"title": "External Resources",
```
JWO719: Sorry to claim about that again, but the indention seems to be wrong now

pertrai1:
@JWO719 good catch. That was weird how it was indented. fixed when you have a chance to take a look.

benlesh:
I haven't vetted any of the linked material. I know that there is a lot of other material out there though. 

I'm holding off on merging this until I have a few core team members view the material for quality. We need to make sure we're not just advertising someone's courses or something like that.

cc @ladyleet @cartant @kwonoj

pertrai1:
@benlesh  - Sure. This can even be held off if more discussion is needed. This might not even be needed.

cartant:
Of the linked resources, the only ones with which I'm familiar are the StrongBrew SIP blog article and Brian's learnrxjs.io.

I've not spent much time with the latter, but I did like the StrongBrew article.

I'm not sure what would be involved in vetting these or the others, as I don't know what standards should be applied re: self-promotion, ads, etc. And, TBH, I'm not super keen on reading through or watching all of the material.

pertrai1:
I do like @cartant idea of the separate awesome list of RxJS repo.

JWO719:
I think pointing to an separate repository would be even more volatile. For sure we have to check the linked resources roughly but I think as long as we make clear that we aren't responsible for the content on the linked resources this is the best approach to go. Because then the docs become some kind of single source of truth to make developer getting into rxjs easier

pertrai1:
I think it comes down to 2 options:

1. There is a link that can lead people to a curated list of RxJS resources that is being maintained by another repository owner. - not in favor of this per se - example https://github.com/ichpuchtli/awesome-rxjs
2. Create a new Repo under the ReactiveX umbrella that would become a library of resources.

Of course, if neither of these are ideal, it can always be left to Google :-)

JWO719:
@benlesh is this a thing we should discuss on the next core team meeting?

benlesh:
@JWO719 I'm approving this one, I think we can add the link to "edit on Github" (as discussed in the core team meeting) after it's part of the repo a little easier, rather than guessing at what it should be.

pertrai1:
@JWO719 @benlesh should the "Edit on GitHub" be added to this or do you have another method you want to use?

JWO719:
I would suggest adding it later on to have a better overview what is done and what needs to be done. As soon this pr is merged I can create a new follow up issue for this edit feature. I would appreciate it if you continue supporting and taking care of it :blush:

benlesh:
Can someone please investigate the build errors here? Perhaps it just needs rebased?

Issue #3827, opened 2018-06-11
Rxjs has no default export


RobbyDmz:
## Bug Report

**Current Behavior**
The documented method to import the entire library does not work with webpack/ES6 modules.
From the current docs:
````javascript
import Rx from 'rxjs/Rx';
Rx.Observable.of(1, 2, 3);
````
As the Rxjs/Rx package has no default export defined this can't work.
   
**Expected behavior**
Either the documentation should be updated or the package should have a default export.

**Environment**
- Runtime: node v8.11.1
- RxJS version: 6.2.0
- Tested with react-create-app's defaults
- Tested with own minimal webpack v4.12.0 + ts-loader v4.4.1 setup

**Possible Solution**
Update documentation to 
````javascript
import * as Rx from 'rxjs/Rx'
Rx.Observable.of(1, 2, 3);
````

**Comment by r3h0** [comment]
I believe the preferred way is now this:

```
import { of } from 'rxjs';
of(1, 2, 3);
```

syndicatedshannon:
Note that operators currently require a different import syntax for UMD, as described in #3850

Issue #3943, opened 2018-07-23
UMD module name changed from "Rx" to "rxjs"


martinsik:
### Documentation Related To Component:

CHANGELOG.md

### Please check those that apply

- [ ] typo
- [X] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

### Description Of The Issue

Everyone who has been using RxJS by importing it with `<script>` for example is used to prefix everything with `Rx` (eg. `Rx.Observable.of`).

This commit https://github.com/ReactiveX/rxjs/commit/556c904ea61a8424e5d24f170b20eadbc05d01f0#diff-6d2911fe563068b8126098588db98a84R28 renamed the module name to `rxjs` but this change is not mentioned in `CHANGELOG.md`.

I don't know if this is intentional or it was just forgotten.

cartant:
FWIW, my understanding is that it was renamed so that the usage can be similar to what would be done with imports, e.g.:

```js
const { of } = rxjs;
const { map } = rxjs.operators;
/* ... etc. */
```

Whether or not this should be in the CHANGELOG or elsewhere, I don't know.

benlesh:
It should probably be added to the `MIGRATION.md`, maybe?

macklinu:
Was this resolved by the `MIGRATION.md` update in #4063?

BioPhoton:
@benlesh I guess we can close this one. If not please let me know whats missing

PR #4063, opened 2018-08-28
docs(installation): update imports to use rxjs 6+ convention (#3827)

dkosasih:
**Description:** Update sample in Installation section to use Rxjs 6+ convention.

**Related issue (if exists):** #3827

### DIFF
--- a/docs_app/content/guide/installation.md
+++ b/docs_app/content/guide/installation.md
@@ -16,7 +16,7 @@ import * as rxjs from 'rxjs';
 rxjs.of(1, 2, 3);
 ```
 
-To import only what you need by patching (this is useful for size-sensitive bundling):
+To import only what you need using pipeable operators:
 
 ```js
 import { of } from 'rxjs';
@@ -24,6 +24,7 @@ import { map } from 'rxjs/operators';
 
 of(1,2,3).pipe(map(x => x + '!!!')); // etc
 ```
+* See [Pipeable Operator Documentation](https://github.com/ReactiveX/rxjs/blob/91088dae1df097be2370c73300ffa11b27fd0100/doc/pipeable-operators.md) for more information about pipeable operator.
 
 To use with globally imported bundle
 ```js
--- a/docs_app/content/guide/installation.md
+++ b/docs_app/content/guide/installation.md
@@ -11,30 +11,18 @@ npm install rxjs
 To import the entire core set of functionality:
 
 ```js
-import Rx from 'rxjs/Rx';
+import * as rxjs from 'rxjs';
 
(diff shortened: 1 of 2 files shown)


docs_app/content/guide/installation.md
```diff
@@ -11,30 +11,28 @@ npm install rxjs
...
 ```
 
 To import only what you need by patching (this is useful for size-sensitive bundling):
 
 ```js
-import ❴ Observable ❵ from 'rxjs/Observable';
-import 'rxjs/add/observable/of';
-import 'rxjs/add/operator/map';
+import { of } 'rxjs';
+import { map }'rxjs/operators';
```
cartant: Both of these imports are missing `from`.

docs_app/content/guide/installation.md
```diff
@@ -11,30 +11,28 @@ npm install rxjs
...
 
 ```js
-import ❴ Observable ❵ from 'rxjs/Observable';
-import ❴ of ❵ from 'rxjs/observable/of';
-import ❴ map ❵ from 'rxjs/operator/map';
+import ❴ of ❵ from 'rxjs';
+import ❴ map ❵ from 'rxjs/operators';
 
-Observable::of(1,2,3)::map(x => x + '!!!'); // etc
+of(1,2,3)|>map(x => x + '!!!'); // etc
```
cartant: This is the proposed pipe operator, not the proposed [bind operator](https://github.com/tc39/proposal-bind-operator) - which, AFAICT, is not going to be accepted. It might be better to just remove this section altogether.

cartant:
Also, it might be worth incorporating [this issue](https://github.com/ReactiveX/rxjs/issues/3943) into the PR.

dkosasih:
@cartant Thanks for looking at it so quickly.
I have made the changes. 

> Also, it might be worth incorporating this issue into the PR.

Could you advise on this? Should I add the documentation on Migration.md?

cartant:
> Could you advise on this?

You could add, to both places, some docs that show how to import it from a global when using a bundle. I think new users won't be looking at the MIGRATION document, but might still want to know how bundle imports would work.

If you want to add some global import docs in this PR and then add something to the MIGRATION doc in a separate PR, that's fine.

dkosasih:
> PS. Regarding my looking at this so quickly: I have to do something whilst ts-node rebuilds its cache. 😄
As long as it's quick 👍 

I have done another commit to document the issue #3943

dkosasih:
@benlesh I have made the changes. Please check if they are good. 
Thanks.

benlesh:
Awesome! Thanks @dkosasih!

Issue #3948, opened 2018-07-24
Disable ts-node caching for tests


cartant:
## Feature Request

By default, `ts-node` uses a per-user cache of compiled files. Files are re-compiled when the original source file is determined to have changed. However, that can hide typing problems and the current test infrastructure is reliant upon `ts-node`'s compilation of the TypeScript files to detect type-related errors.

For example if `npm run test` is used to run the test suite, then everything is compiled and cached. If, say, a change is made to a core function like `pipe` and the `npm run test` is run again, only the file containing the `pipe` is recompiled. And if the change uncovers errors in other test files, those errors are hidden and won't be seen unless said test files are changed to effect recompilation.

Today, not seeing the failures on a local computer meant that I didn't know there was a problem until the commits were pushed and the Travis CI build failed - as it had no cached, compiled files.

**Describe the solution you'd like**

I'd like the test script to disable the `ts-node` cache - using either the `--no-cache` CLI option or the `TS_NODE_CACHE` environment variable.

**Describe alternatives you've considered**

Alternatively, another test script that disables the cache could be added, say `test_no_cache` or something.

cartant:
Once you know what's going on, it's easy to set `TS_NODE_CACHE` on the command line, but having it disabled in the first place would avoid the problem.

Or having an obviously named script - like `test_no_cache` - would be a pretty good hint regarding what's going on for the next person to wonder why tests pass on their machine and not on Travis.

PR #4065, opened 2018-08-29
feat(test): Add npm script to test with ts-node not caching

pertrai1:
**Description:**

There is a npm script that can run - `npm run test_no_cache`, that will
run the test with `TS_NODE_CACHE` set to false.

#3948

### DIFF
--- a/spec/migration/update-6_0_0/index-spec.ts
+++ b/spec/migration/update-6_0_0/index-spec.ts
@@ -13,11 +13,12 @@ describe('Migration Schematic', () => {
     tree.create('/package.json', `{}`);
   });
 
-  it('adds missing dependencies', () => {
+  it('adds missing dependencies', (done) => {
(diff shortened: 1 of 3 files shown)


spec/migration/update-6_0_0/index-spec.ts
```diff
@@ -13,11 +13,12 @@ describe('Migration Schematic', () => {
     tree.create('/package.json', `{}`);
   });
 
-  it('adds missing dependencies', () => {
+  it('adds missing dependencies', (done) => {
```
cartant: Is the `done` here really needed? Is there a particular reason for adding it?  It seems superfluous.

pertrai1: @cartant Sorry, should have commented on that. When I added the script and ran `npm run test_no_cache`, the test failed with a timeout without having the done.

benlesh:
Thanks @pertrai1

PR #4067, opened 2018-08-30
docs(UI): Reintroduce Docs Search Capability

pertrai1:
**Description:**

Uncomment the search box for quicker document lookup.

screen shot 2018-08-30 at 3 39 06 am


JWO719:
We commented that out, because it displayed angular related content. It might be possible that there is none now, after  we have removed most of the angular specific content. But I would like to check it manually

pertrai1:
@JWO719 do you think this will be something that will be good to re-introduce?

JWO719:
Definetly !!! I just couldn't check it yet. Sorry for that. I hope to take a look at it this weekend.

pertrai1:
@JWO719 thanks. I fixed the conflict.

PR #4068, opened 2018-08-30
docs(UI): Mobile Header Logo

pertrai1:
**Description:**

The header is now showing the correct mobile sized logo rather than an outdated
link to one that no longer exists.

screen shot 2018-08-30 at 3 59 15 am


benlesh:
Thank you, @pertrai1

Issue #4071, opened 2018-08-30
nested concatMaps no longer next values to destination in 6.3.0


benlesh:
## Bug Report

**Current Behavior**
Currently if you nest a `concatMap` inside of a `concatMap`, and the source completes, the inner most concatMap will not forward along values.

**Reproduction**
```ts
import { timer, from } from 'rxjs';
import { concatMap, tap } from 'rxjs/operators';

function wait(ms){
  return new Promise(res => {
    setTimeout(() => res('poop'), ms);
  })
}

timer(10).pipe(
  concatMap(() => from(wait(20)).pipe(
    concatMap(() => {
      console.log('here');
      return from(wait(20)).pipe(
        tap(x => console.log('>', x))
      );
    })
  ))
)
.subscribe(x => console.log('>', x));
```

**Expected behavior**
It should work like it did in 6.2

**Environment**
- Runtime: all
- RxJS version: 6.3.0

**Possible Solution**

None

benlesh:
cc @hansl @IgorMinar

benlesh:
This seems to specifically have to do with promises, FYI:

```ts
    const results: any[] = [];

    of(1).pipe(
      concatMap(() =>
        from(Promise.resolve(2)).pipe(
          concatMap(() => Promise.resolve(3))
        )
      )
    )
    .subscribe({
      next(value) { results.push(value); },
      complete() { results.push('done'); }
    });

    setTimeout(() => {
      expect(results).to.deep.equal([3, 'done']);
      done();
    }, 0);
```

cartant:
Specifically, it seems to be due to the outermost promise.

If you comment out the `from`, replacing it with `of(2, asapScheduler)`, it works fine:

```ts
of(1).pipe(
  concatMap(() =>
    from(Promise.resolve(2))
    // of(2, asapScheduler)
    .pipe(
      concatMap(() => defer(() => {
        console.log('sub');
        return of(3, asapScheduler).pipe(
          tap(() => console.log('next')),
          finalize(() => console.log('unsub'))
        );
      }))
    )
  )
)
.subscribe({
  next(value: any) { results.push(value); },
  complete() { results.push('done'); }
});
```

When the promise is used, it appears to subscribe to the innermost observable and then immediately unsubscribe. And then it seemingly doesn't complete, as there's no `done` in `results`.

I can look further into this later, if you've not already resolved the regression in the interim.

cartant:
This fails, too, and does not use a promise:

[...]

PR #4072, opened 2018-08-31
fix(mergeMap): fix nested mergeMaps

cartant:
**Description:**

This PR should fix the nested `concatMap`/`mergeMap` regression. The problem seems to have been related to a subscriber being added to itself as a parent.

**Related issue (if exists):** #4071

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@reactivex/rxjs",
-  "version": "6.2.2",
+  "version": "6.3.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


filipesilva:
The Appveyor CI is failing on the new test:
```

  1) mergeMap
       should support nested merges:
      Uncaught AssertionError: expected [] to deeply equal [ 3, 'done' ]
      + expected - actual
      -[]
      +[
      +  3
      +  "done"
      +]
      
      at Proxy.assertEqual (node_modules\chai\lib\chai\core\assertions.js:1020:19)
      at Proxy.methodWrapper (node_modules\chai\lib\chai\utils\addMethod.js:57:25)
      at Timeout._onTimeout (spec\operators\mergeMap-spec.ts:737:31)
```

Odd it didn't fail on other platforms. Might be a race condition.

cartant:
> Might be a race condition.

Most likely. I've changed that test to use the micro task scheduler, so the `setTimeout` should always happen last, now.

benlesh:
Thanks @cartant! You saved me some work this morning for sure. That was first on my agenda today. <3

PR #4078, opened 2018-08-31
fix(node): will no longer error mixing RxJS 6.3 and 6.2

### LINKED ISSUE #4077: Errors when RxJS 6.3 and 6.2 exist in the same node project.
It's complaining because it passed the `isTrustedSubscriber` test when it shouldn't. There's a new `_addParentTeardownLogic` method that is required in 6.3

benlesh:
- updates `isTrustedSubscriber` to check for `_addParentTeardownLogic`

resolves #4077

### DIFF
--- a/spec/Subscriber-spec.ts
+++ b/spec/Subscriber-spec.ts
@@ -1,7 +1,7 @@
 import { expect } from 'chai';
-import * as Rx from 'rxjs/Rx';
-
-const Subscriber = Rx.Subscriber;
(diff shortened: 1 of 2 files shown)


kwonoj:
we don't need check for both for prior version?

benlesh:
@kwonoj not really, if `_addParentTeardownLogic` is present, the other one will be there. Unless someone tries to write their own clever, forked version of RxJS and mix it with 6.3.

kwonoj:
gotcha. Go for it 🚢

Issue #3982, opened 2018-08-02
Deprecate `create` static methods


benlesh:
The `create` static methods, like `Observable.create` and `Subject.create` are rife with problems.

While it's GREAT to have a functional way to create these things without the `new` keyword... it should be noted that:

1. TypeScript hates what we're doing here.
2. `Subject.create` doesn't actually create a `Subject`, but rather an `AnonymousSubject`, which I would REALLY like to rename as `FrankenSubject` because that describes what it is, you basically glue an Observer to an Observable and call it a "Subject".
3. None of the Subjects have a `create` method that makes sense, they all get the one from `Subject` (see 2)
4. Neither does `ConnectableObservable`... it just gets the one from `Observable`.

### Proposal

Deprecate all `create` methods.

Then either:

1. Do nothing - no other library deals with wrapping constructors in functions for people, afaik.
2. Supply lowercase functional versions of each:  `subject`, `behaviorSubject`, `connectableObservable`
3. Make each constructor callable without `new` (the experimental version can do this, but the current version here can't really because we're using ES2015 classes). So like `Observable(fn)` is the same as `new Observable(fn)`

Thoughts?

jayphelps:
I'm for removing `create` methods, then doing nothing for now. Is there an argument for "functional" versions that don't require `new`, other than disliking the keyword?

benlesh:
> Is there an argument for "functional" versions that don't require new, other than disliking the keyword?

1. You can't call ES2015 class constructors properly with `call` or `apply`. 
2. You can pass a constructor to a function that accepts a function and have it realize it needs to be called with `new`.  e.g. `function execute(fn) {  fn(); }`... (UNLESS You try/catch and handle the `TypeError: Class constructor Test cannot be invoked without 'new'` and then call it appropriately, I guess)

Those are really the only solid arguments.

cartant:
> no other library deals with wrapping constructors in functions for people, afaik

FWIW, Leaflet does this. It provides classes and [lower-case factory functions](https://github.com/Leaflet/Leaflet/blob/v1.3.3/src/layer/tile/TileLayer.js#L242-L247).

cartant:
> Make each constructor callable without `new`

Changing `Observable`, etc. to not be a class would be a breaking change, so I don't see how, in v6, the `create` functions could be deprecated and replaced with ctors that are callable without `new`.

My vote would be to deprecate and then do nothing.

benlesh:
[...]

Issue #4079, opened 2018-09-01
Consider a temporary node_modules installation for dtslint


cartant:
## Bug Report

**Current Behavior**

There is a problem that sometimes occurs with `dtslint`. In particular, it usually occurs on Travis and usually occurs around the time of a TypeScript release (AFAICT).

When `dtslint` runs, it installs a bunch of TypeScript versions underneath its own directory in `node_modules`. It seems that using a lock file sometimes cause errors like this:

```
> @reactivex/rxjs@6.3.1 dtslint /home/travis/build/ReactiveX/rxjs
> dtslint spec-dtslint
module.js:549
    throw err;
    ^
Error: Cannot find module 'definitelytyped-header-parser'
    at Function.Module._resolveFilename (module.js:547:15)
    at Function.Module._load (module.js:474:25)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/travis/build/ReactiveX/rxjs/node_modules/dtslint/bin/index.js:12:41)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
```

It seems that the error is effected because whatever `dtslint` downloads expects to be run in the presence of dependencies that are up-to-date and not pinned by a lock file.

My workaround for this is to update the lock file using:

```
npm uninstall dtslint --save-dev ; npm install dtslint --save-dev --save-exact
```

And to then push the updated lock file in a commit. I've used this procedure to fix the error on several occasions.

**Possible Solution**

Given that `dtslint` downloads modules when it runs and doesn't expect its dependencies to be pinned, it might be better to use a separate, temporary `node_modules` installation for `dtslint` itself.

PR #4080, opened 2018-09-01
chore(Observable): deprecate create method (#3982)

dkosasih:
**Description:** deprecate all the create method, use `new` instead

**Related issue (if exists):** #3982

### DIFF
--- a/src/internal/Subject.ts
+++ b/src/internal/Subject.ts
@@ -39,7 +39,9 @@ export class Subject<T> extends Observable<T> implements SubscriptionLike {
     super();
   }
 
-  /**@nocollapse */
+  /**@nocollapse
+   * @deprecated use new Subject() instead
+  */
   static create: Function = <T>(destination: Observer<T>, source: Observable<T>): AnonymousSubject<T> => {
     return new AnonymousSubject<T>(destination, source);
   }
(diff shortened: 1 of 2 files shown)


dkosasih:
@cartant I am not sure if the commit message type should be chore - do you want this changes to make it to changelog?

cartant:
The error isn't related to your change; it's a problem with `dtslint`. See https://github.com/ReactiveX/rxjs/issues/4079

dkosasih:
I was trying to figure out why is this failing. I tried dtslint on my local machine and it was fine. 
Thanks for letting me know. Happy Monday.

src/internal/Subject.ts
```diff
@@ -39,7 +39,9 @@ export class Subject<T> extends Observable<T> implements SubscriptionLike {
     super();
   }
 
-  /**@nocollapse */
+  /**@nocollapse
+   * @deprecated use new Subject() instead
```
cartant: Hmm. I'm not sure about this deprecation message, as this `create` method does things that cannot be done with the `Subject` constructor alone. Your thoughts, @benlesh ?

dkosasih: I was not very sure about this one too because the constructor signature is different. Observable one is more straight forward.

benlesh: I think it's fine. The only thing that `create` did that `new Observable` couldn't was allow `apply` and `call`, which people can get from a simple factory function.

benlesh:
@dkosasih can you rebase, please?

dkosasih:
Done 😊

Issue #4004, opened 2018-08-10
Please put old docs back, as too much is missing in the new ones


dietergeerts:
### Documentation Related To Component:

Several ones

### Please check those that apply

- [ ] typo
- [x] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

### Description Of The Issue

Several components are missing documentation, like switchAll. + There is no possibility to see the old docs from it, which makes it hard to check them out. That's why the old docs should still be accessible!!!

JWO719:
So basically the old docs are still available -> http://reactivex.io/rxjs/manual/index.html
You just can't go to the home page, there you will be redirected to the new docs. But besides that, could you maybe elaborate what else is missing. SwitchAll is poorly documented absolutely right. But if you help me with that kind of feedback, I can improve the new docs, to make them more useful for everyone.

001123:
I wish have new docs for rxjs 6 above 🙏

dietergeerts:
But why redirect when the new docs aren't complete yet? Better to give the option then, no?

ahahn95:
Do we know when the new docs will be available? I'm pretty excited about rxjs6, but it might as well be DOA for me without docs..

pertrai1:
@ahahn95 you can find the new docs for version 6 at http://reactivex.io/rxjs. Each day more and more of the docs are getting updated. If there is any documentation you find missing that would help you, please share in this thread.

ahahn95:
Sure, I'm specifically looking for the ajax docs. (importing ajax from 'rxjs/ajax' etc etc)

pertrai1:
@ahahn95 the docs are being built from JSDocs from each of the files. I am not sure why this is not added, and I am in the process of finding out. In the meantime, you can take a look at the file and see the JSDocs for the ajax.ts file: https://github.com/ReactiveX/rxjs/blob/master/src/internal/observable/dom/ajax.ts

Will find out about this for you in regards to showing in the actual docs.

cartant:
@pertrai1 Regarding the `ajax` docs inclusion in the build, see this recent PR: https://github.com/ReactiveX/rxjs/pull/3984

ahahn95:
`"It creates an observable for an Ajax request with either a request object with url, headers..."`

So I assume it would look like

    ajax({
        url: 'http://myurl.com',
        headers: {
            ...
        }
    })

alphashuro:
[...]

PR #4082, opened 2018-09-03
chore(docs): revert gh-pages with banner to v6 firebase

pertrai1:
**Description:**

Possible solution to those who are able to find more helpful information from the old docs while still having a clearly visible callout to the newer docs that are being developed.

screen shot 2018-09-03 at 2 42 16 am

#4004


benlesh:
The problem is the old docs have outdated information on them. I'm not sure it's going to make things less confusing to have two sets of documentation on them.

We should identify what exactly people are missing from the old docs and port that, rather than have two sets of documentation. It took _years_ for people to start finding the RxJS 5/6 docs instead of always going to the v4 docs in the old repository.

benlesh:
If we *did* go with two sets of docs (which I'm _not_ in favor of) I think it would be better to have the new docs be the primary, since most people are on v6.

pertrai1:
So how about we go with the docs being as they are with the redirect and then a link to take someone to previous docs if they need them?

JWO719:
Unfortunately the docs couldn't be deployed recently due to some mistakes I made with the deployment script. We are currently trying to find the problem. But I agree with Ben it would be mostly confusing for people. And we anyway wanted to migrate the stuff of the old docs. Maybe we should focus on this?

pertrai1:
@JWO719 are you saying that we should not worry about having a way for others to find the older documentation?

JWO719:
No you slightly misunderstood me. It's not about being worried or not, I just think that this is not the best approach to solve this problem. The problem with the old docs is that it is sometimes rxjs 5 related and sometimes v6. I would prefer having one strict version and a proper documentation of that. If one  has a proper v 6 documentation and really needs infos about v5, that step backwards is not that complicated. Therefore I would prefer improving the new docs with the stuff that is needed by the community and maybe add a chapter about things changed in v6 from v5.

pertrai1:
Ok, sounds good. Best to go ahead and close this PR.

JWO719:
Don't get me wrong, I really welcome any suggestions and your initiativ

benlesh:
Thank you for your work, @pertrai1. I'm going to close this one for now.

dietergeerts:
I also think one place for docs is better, BUT almost everything is missing..... No marble diagrams, no examples, descriptions gone from a LOT of methods etc....  I wonder why this happened? Are the docs not manually tested to see if the docs written actually end up on the docs site?

examples:
- https://rxjs-dev.firebaseapp.com/api/index/function/from (only typings, you can't even click through to find out what all the types are)
- shareReplay, same
- ...

PR #4085, opened 2018-09-03
Overload take while to accept a type guard

### LINKED ISSUE #4081: overload takeWhile with type guard
## Feature Request

Hi, I would like to propose overloading the type definition of the takeWhile operator with a signature similar to the one used by the filter operator.

I think of something similar to this:

```typescript
function takeWhile<T, S extends T>(predicate: (value: T, index: number) => value is S): OperatorFunction<T, S>;
```

This would help with the following scenario:

```typescript
of('hello', 'world', null)
  .pipe(
    takeWhile((word): word is string => (word !== null)),
    map((word) => word.toUpperCase())
  );
```

Currently TypeScript will complain about this code, as it doesn't know that only strings can possibly reach the map operator.

The example is of course contrived but I run often into situations like that when selecting something from a redux store.

It is currently possible to add explicit type casts or even an additional filter operator to achieve the same behavior, but I think overloading takeWhile would be the cleaner solution.

In case you agree, I'm happy to submit a PR for that.

chrisguttandin:
**Description:**

This pull request overloads the takeWhile operator with an additional signature to accept a type guard.

I tried to keep the code as similar as possible to the corresponding overload signature of the filter operator. Please let me know, if I should change something.

I wasn't sure if the JSDoc should be changed as well. The return type of the filter operator is just `Observable` (without a generic) whereas the one of the takeWhile operator is `Observable<T>`.

This PR fixes #4081.

### DIFF
--- a/spec-dtslint/operators/takeWhile-spec.ts
+++ b/spec-dtslint/operators/takeWhile-spec.ts
@@ -1,0 +1,10 @@
+import { of } from 'rxjs';
+import { takeWhile } from 'rxjs/operators';
+
+it('should support a user-defined type guard', () => {
+  const o = of('foo').pipe(takeWhile((s): s is 'foo' => true)); // $ExpectType Observable<"foo">
+});
+
+it('should support a predicate', () => {
+  const o = of('foo').pipe(takeWhile(s => true)); // $ExpectType Observable<string>
+});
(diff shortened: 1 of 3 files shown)


cartant:
Thanks. Looks good.

Could I get you to add some `dtslint` tests, too? For testing typings, they're much less cumbersome. And I'd like typings changes to be accompanied by `dtslint` tests so that we can build out coverage of the typings.

Have a look at the `find` tests for examples:

https://github.com/ReactiveX/rxjs/blob/master/spec-dtslint/operators/find-spec.ts

To run the tests, use this command:

```
npm run dtslint
```

benlesh:
Thanks, @chrisguttandin!

PR #4088, opened 2018-09-03
chore(dtslint): use npx

cartant:
**Description:**

<s>This PR removes the running of `dtslint` from the full-validation build.</s>

<s>Until #4079 is sorted and `dtslint` can be installed without pinned dependency versions, it'll be less painful to just switch it off. Having every PR fail is too annoying and having to update the lock file all the time is absurd.</s>

<s>Its running can be re-instated once #4079 fixes the problem.</s>

This PR disables a new `dtslint` rule that's not needed and removes `dtslint` from the project's `package.json`. Instead, it's <s>installed into `spec-dtslint-build` (which is a git-ignored directory)</s> run using `npx`, which installs a temporary copy. This should mean that `dtslint` installs (on Travis, at least) are no longer pinned by the project's lock file - something that has been causing problems.

**Related issue (if exists):** #4079


cartant:
I'm not familiar with `npx`. If it were to be used, how/where would `dtslint` be installed? Globally?

Also, regarding a different `tsc` version, `dtslint` does something weird and installs/downloads a bunch of different versions of TypeScript (under its own `node_modules/dtslint` directory) and uses them when it runs. That is, it already runs against a bunch of different TypeScript versions.

I think it's downloading stuff at runtime is somehow related to the problems that are effected by pinning versions in the `package.json`. But I cannot say that I fully understand what it's doing. It's somewhat infuriating.

kwonoj:
Npx resolves to its temp (npm cache) location if package is not specified / installed under node_modules. Theoritically it shoudln't affected by local module installation due to its behavior.

cartant:
Interesting. I'll have a look at it.

cartant:
Well, it works, but it installs a fresh copy into its cache each time. So it's kinda slow.

kwonoj:
yeah, but it's ci so I wouldn't be much worried for some slowness. if someone need to run it on local frequently, may need to think better workaround though.

cartant:
The `spec-dtslint-build` script fails on Linux due to a quote issue, so I'm going with your `npx` solution. I can workaround it differently - locally - to avoid the slowness problem.

Thanks.

pertrai1:
🚢

Issue #3803, opened 2018-06-06
Report synchronous errors thrown after subscriber completion


cartant:
## Feature Request

**Is your feature request related to a problem? Please describe.**

The issue is a re-write (as discussed in the last core team meeting) of issue #3461 - which is not as clearly described as it could have been.

There have been two recent bugs in which errors thrown synchronously - after subscriber completion - were not reported:

* in #3444, the `repeatWhen` swallowed an error that was thrown due to a re-enterant to `complete` (described [here](https://github.com/ReactiveX/rxjs/issues/3444#issuecomment-375127242)).
* in #3487, a call to `bindCallback` swallowed an error thrown synchronously after the callback was invoked.

The errors were swallowed because the implementation of `subscribe` attempts to report them in [`_trySubscribe`](https://github.com/ReactiveX/rxjs/blob/6.2.0/src/internal/Observable.ts#L213-L223) by calling the subscriber's `error` method:

```ts
_trySubscribe(sink: Subscriber<T>): TeardownLogic {
  try {
    return this._subscribe(sink);
  } catch (err) {
    if (config.useDeprecatedSynchronousErrorHandling) {
      sink.syncErrorThrown = true;
      sink.syncErrorValue = err;
    }
    sink.error(err);
  }
}
```

However, in both the `repeatWhen` and `bindCallback` issues, the subscriber's `complete` method had already been called. That means the call to `error` was ineffectual and no error was reported because the [`SafeSubscriber`](https://github.com/ReactiveX/rxjs/blob/6.2.0/src/internal/Subscriber.ts#L219-L247) was stopped:

```ts
error(err?: any): void {
  if (!this.isStopped) {
    // ...
  }
}
```

**Describe the solution you'd like**

The errors should not be swallowed; they need to be reported somehow.

Ideally, any errors that cannot be reported via the subscriber's `error` method should be reported via `hostReportError`.

However, doing so has the potential to break client code, as throwing errors that were previously swallowed would be a breaking change.

Using the mechanism described in #3469, it's possible to navigate the chain of subscribers to determine if any are in fact stopped. If none is stopped, the error can be reported by calling the subscriber's `error` function. If a stopped subscriber is encountered, the error cannot be reported to the subscriber and should instead be reported to the console.

Something like this implementation of `reportError`, which could be called in `_trySubscribe`:

```ts
export function reportError(err: any, observer: ErrorObserver<any>): void {
  if (canReportError(observer)) {
    observer.error(err);
  } else {
[...]

Issue #3848, opened 2018-06-21
Mocking fromEvent in rxjs6 


rsirs:
### Spy on fromEvent haven't been called

```js
spyOn(Observable, 'fromEvent)
expect(Observable.fromEvent).toHaveBeenCalled();
```

Above mentioned piece of code is not working for rxjs6 version but working on rxjs5 

## Work arounds to make this work but failed

```js
import {fromEvent} from 'rxjs'

let Obj = { observableFromEvent: fromEvent }
spyOn(Obj, 'observableFromEvent')
expect(observableFromEvent).toHaveBeenCalled();

/* jasmine test case failed with spy hasn't been called */
```

```js
import * as rxjs form 'rxjs'

spyOn(rxjs, 'fromEvent')
expect(rxjs.fromEvent).toHaveBeenCalled();

/* Test case failed with not setter or non writable fromEvent */

<!-- Descrbe the issue in detail here -->

**Comment by benlesh** [comment]
So you want to spy on `fromEvent`?

You should be able to just import everything, and spy on that:

```ts
import * as rxjs from 'rxjs';

spyOn(rxjs, 'fromEvent')
```

rsirs:
@benlesh I already did  but, Jasmine is complaining [method_name] is not declared writable or has no setter since es6 imports are constants.

aniruddhadas9:
After using @benlesh idea, I am getting below 

```
	Error: <spyOn> : fromEvent is not declared writable or has no setter
	Usage: spyOn(<object>, <methodName>)
	    at Object.<anonymous> src/app/app.component.spec.ts:215:7)
	    at TestBed.push../node_modules/@angular/core/fesm5/testing.js.TestBed.execute node_modules/@angular/core/fesm5/testing.js:1074:1)
	    at Object.<anonymous> node_modules/@angular/core/fesm5/testing.js:1225:29)
	    at Object.<anonymous> node_modules/zone.js/dist/fake-async-test.js:593:1)
	    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invoke node_modules/zone.js/dist/zone.js:388:1)
	    at ProxyZoneSpec.push../node_modules/zone.js/dist/proxy.js.ProxyZoneSpec.onInvoke node_modules/zone.js/dist/proxy.js:128:1)
[...]

Issue #3487, opened 2018-03-29
bindCallback swallows error if it was thrown after the callback was executed, and the entire function is sycnhronous


olessavluk:
**RxJS version:**
5.5.8

**Code to reproduce:**
same for node and browser

```javascript
const fn = (datum, cb) => { 
  cb(datum);  
  throw new Error("callback error");
};

Rx.Observable.bindCallback(fn)(42)
  .subscribe(
    v => console.log(v), 
    e => console.error(e),
    c => console.log("completed"),
  );
```

**Expected behavior:**
see 42 and then "callback error" in the console

**Actual behavior:**
console output:
```
42
completed
```

**Additional information:**

In the real world, I've captured this issue with `ReactDOM.render` method. If you run it with "bad" JSX it will call passed callback first and then throw an error.

I've tried to reproduce this case as a test, but it worked well in the test environment. So maybe there is something with test scheduler or production build
<details>

```js
    it('should error if callback throws', () => {
      const expected = new Error('haha no callback for you');
      function callback(datum, cb) {
        cb(datum);
        throw expected;
      }
      const boundCallback = Observable.bindCallback(callback, null, rxTestScheduler);

      boundCallback(42)
        .subscribe((x: number) => {
          throw 'should not next';
        }, (err: any) => {
          expect(err).to.equal(expected);
        }, () => {
          throw 'should not complete';
        });

      rxTestScheduler.flush();
    });
```
</detail>

benlesh:
I've updated the title of this issue, because it's specific to completely synchronous functions with callbacks.

benlesh:
In the short term, the fix for this is to pass a scheduler to the bindCallback function:  `bindCallback(fn, asapScheduler)` for example.

cartant:
Closing this because it's addressed in #4089.

PR #4089, opened 2018-09-04
fix(subscribe): don't swallow internal errors

cartant:
**Description:**

This PR reports errors that would otherwise be swallowed - because of a closed subscriber - to the console. (This behaviour is something that bit me again, today, when I was investigating a snippet that effected a stack overflow error in v5, but didn't effect the error in v6. The error was still occurring, of course, it just wasn't being thrown.)

As discussed in the linked issue, calling `hostReportError` would be preferred - rather than writing to the console - but doing so would be a breaking change.

**Related issue (if exists):** #3803


pertrai1:
`kaboom` - 👍 LGTM

src/internal/util/reportError.ts
```diff
@@ -0,0 +1,34 @@
+import { Subscriber } from '../Subscriber';
```
benlesh: Since this is only used in one spot, you can just inline everything in that one spot. It'll perform better, and will be more readable (if a little harder to test)

src/internal/util/reportError.ts
```diff
@@ -0,0 +1,34 @@
...
+  const { closed, destination, isStopped } = observer as any;
+  if (closed || isStopped) {
+    return false;
+  } else if (destination instanceof Subscriber || destination[rxSubscriberSymbol]) {
+    return canReportError(destination);
+  }
+  return true;
+}
+
+function consoleReportError(err: any): void {
```
benlesh: `consoleWarn`, this is too confusing considering we have a `hostReportError` function in the library, which is different.

src/internal/util/reportError.ts
benlesh: For tests, I'd rather just have more integration/functional tests of this.

cartant:
@benlesh I've made a bunch of changes inline with your requests:

* Removed `reportError` and used `canReportError` for readability.
* Renamed to `consoleWarn`.
* The `canReportError` error tests now just test the logic - not the reporting.
* I've fixed a logic error in the `canReportError` implementation.
* `canReportError` is no longer recursive.
* `canReportError` error is also used in `bindCallback` and `bindNodeCallback` - see #3487. (Your mentioning that it's only used in one place reminded me of errors that were swallowed elsewhere.)
* There are integration tests for `Observable.subscribe`, `bindCallback` and `bindNodeCallback` that stub `console.warn`.

src/internal/util/consoleWarn.ts
```diff
@@ -0,0 +1,7 @@
+export function consoleWarn(...args: any[]): void {
+  if (console.warn) {
```
benlesh: We shouldn't have to check this every time the function is called.

```ts
export const consoleWarn = console.warn ? console.warn.bind(console) : console.log.bind(console);
```

benlesh:
Looking good, just a couple of nits.

cartant:
@benlesh I've made changes inline with your requests:

* `isTrustedSubscriber` has been used.
* `ErrorObserver<any>` has been changed to `Subscriber<any> | Subject<any>`.

However, I've not changed `consoleWarn` to test `console.warn` only once for these reasons:

* Doing so means that stubbing `console.warn` won't stub `consoleWarn`.
* There are [issues](https://github.com/ReactiveX/rxjs/issues/3848) stubbing exported functions.
* The actual logging to the console will be so much slower that it will make next to no difference, IMO. And it's only going to be in an error condition.

cartant:
`console.warn` is now called directly. It's called directly elsewhere in the codebase, anyway.

aniruddhadas9:
which version of RxJs will this be included?

Issue #4075, opened 2018-08-31
How is rxjs-compat published?


wbinnssmith:
I'm trying to figure out why the artifacts of https://github.com/ReactiveX/rxjs/pull/3855 weren't included in yesterday's release.

I see rxjs's publish script in the repo, but no indication of how rxjs-compat gets published.

I'd like to address this myself but need a bit of direction. Thanks!

benlesh:
it's manual...


First, build the root and publish `@reactivex/rxjs` (soon to be deprecated)
- run `npm publish` from root

then for `rxjs-compat`:
- run `npm publish` from `dist-compat/package`

PR #4091, opened 2018-09-04
chore(build): Copy compat bundle output into package

wbinnssmith:
**Description:**
Previously the bundles were left behind in `dist-compat` but never copied into `dist-compat/package` prior to publish. This ensures the bundles are part of the compat release.

#4075


benlesh:
Thank you @wbinnssmith!

PR #4105, opened 2018-09-06
AjaxObservable: complete subscribers on abort

OliverJAsh:
**Description:**
It is possible for an XHR to be aborted externally, for example when the machine loses connectivity.

Currently, when the XHR is aborted in this way, the subscription remains.

In this case, I think we would want to complete the observable? Although it might also be useful to emit something to indicate it was aborted.

### DIFF
--- a/src/internal/observable/dom/AjaxObservable.ts
+++ b/src/internal/observable/dom/AjaxObservable.ts
@@ -411,6 +411,13 @@ export class AjaxSubscriber<T> extends Subscriber<Event> {
       }
     }
     xhr.onload = xhrLoad;
+    xhr.onabort = () => {
+      const { subscriber, progressSubscriber } = (<any>xhrLoad);
+      if (progressSubscriber) {
+        progressSubscriber.complete();
+      }
+      subscriber.complete();
+    }
     (<any>xhrLoad).subscriber = this;
     (<any>xhrLoad).progressSubscriber = progressSubscriber;
     (<any>xhrLoad).request = request;


cartant:
Are you sure that the `onabort` event fires for anything other than an `abort` call? My reading of the [whatwg spec](https://xhr.spec.whatwg.org/) is that the aborted flag can only be set via a call to the `abort` method.

OliverJAsh:
I tested it and the event was called when abort occurred due to sleeping my computer.

cartant:
I'm wondering if the behaviour differs between browsers and also it differs depending on whether a connection is severed or cannot be established in the first place.

In any case, if `onabort` is to be wired up, I think it should see the observable error, rather than complete.

That's my two cents. I'll be interested in what others think about this.

kwonoj:
Interesting. it doesn't looks like ng/http handles onAbort as well: https://github.com/angular/angular/blob/master/packages/common/http/src/xhr.ts so far I give a quick peek. 

What other libs does? if it's promise based, does it reject / or not handling, etcs? My immediate feeling is also this maybe error cases.

OliverJAsh:
> if it's promise based, does it reject / or not handling, etcs?

I just attempted the same thing with `fetch`.

http://output.jsbin.com/kelayinizo/quiet makes a request that takes 10 seconds to complete.

If I sleep my computer in that time, here is what I find when I return:

```
Uncaught (in promise) TypeError: Failed to fetch
```

benlesh:
I would think the `error` event should fire on the XHR with `readyState` `0`. Is that not the case?

src/internal/observable/dom/AjaxObservable.ts
```diff
@@ -411,6 +411,13 @@ export class AjaxSubscriber<T> extends Subscriber<Event> {
       }
     }
     xhr.onload = xhrLoad;
+    xhr.onabort = () => {
+      const { subscriber, progressSubscriber } = (<any>xhrLoad);
+      if (progressSubscriber) {
+        progressSubscriber.complete();
```
benlesh: `complete` is a signal of successful completion. This isn't correct. If it's aborted, it did not complete successfully.

benlesh:
I can't approve this one, because I don't think it's the right approach.

OliverJAsh:
Thanks for the review @benlesh, I agree it should emit `error` instead of `complete`, like `fetch` does.

> I would think the error event should fire on the XHR with readyState 0. Is that not the case?

In my testing, the observable does not emit anything, so I guess no.

I'll update this PR to emit an error when the `abort` event is received?

OliverJAsh:
@benlesh This is still a bug. Could you reply to my previous comment? I'm happy to contribute

OliverJAsh:
Moved to https://github.com/ReactiveX/rxjs/issues/4251

Issue #4095, opened 2018-09-05
Observable behaves differently when wrapped


OliverJAsh:
## Bug Report

**Current Behavior**
Given this code:

``` ts
import { Observable, of, OperatorFunction, timer } from 'rxjs';
import { catchError, concat, mergeMap, mergeMapTo, startWith } from 'rxjs/operators';

enum State {
    loading = 'loading',
    error = 'error',
}

const ajax$ = new Observable<State>(subscriber => {
    setTimeout(() => {
        subscriber.error('foo');
        subscriber.complete();
    }, 3000);
}).pipe(
    catchError(_error => of(State.error)),
    startWith(State.loading),
);

const delayedRepeatWhenError = (): OperatorFunction<State, State> => state$ =>
    state$.pipe(
        mergeMap(state => {
            const currentState$ = of(state);
            return state === State.error
                ? currentState$.pipe(concat(timer(3000).pipe(mergeMapTo(state$))))
                : currentState$;
        }),
    );

ajax$.pipe(delayedRepeatWhenError()).subscribe(console.log);
```

The output is as expected:

```
loading
error
loading
error
```

However, if the observable is wrapped and then subscribed to, the output is different and not as expected.

Code:

``` ts
const wrappedAjax$ = new Observable<State>(subscriber => {
    const subscription = ajax$.subscribe(subscriber);
    return () => {
        subscription.unsubscribe();
    };
});

wrappedAjax$.pipe(delayedRepeatWhenError()).subscribe(console.log);
```

Output:

```
loading
error
```

This seems to be a regression from one of the releases after 6.2.2, as that's what I just upgraded from.

**Reproduction**
- REPL or Repo link: https://stackblitz.com/edit/typescript-bcrr1f?file=index.ts

cartant:
Thanks for the repro. I've confirmed that this is due to https://github.com/ReactiveX/rxjs/pull/3963.

cartant:
I have a solution for this. I will submit a PR tomorrow.

PR #3963, opened 2018-07-27
Refactor #2457 for v6

cartant:
**Description:**

This PR refactors #2457 for v6 and changes a few identifiers so that they are prefixed with underscores.

/cc @jayphelps

<s>I did try to use [git-blame-someone-else](https://github.com/jayphelps/git-blame-someone-else) to credit the PR's author, but the command left the branch in an indeterminate state, so I just aborted the rebase. If someone else wants to have a crack at updating the PR - to credit the author - the author's details are:</s>

`Mateusz Podlasin <Podlas29@gmail.com>`

**UPDATE**: I've edited the commits manually to credit the author.

**Related issue (if exists):** #2457 #2459

PR #4106, opened 2018-09-06
fix(subscriber): don't unsubscribe self

cartant:
**Description:**

This PR fixes a problem that was introduced in #3963 and was partially fixed in #4072.

The problem was introduced with this change - to `Subscriber` - in #3963 where a parent subscription is unsubscribed upon a `complete` notification:

```diff
complete(): void {
  if (!this.isStopped) {
    this.isStopped = true;
    this._complete();
+   this._unsubscribeParentSubscription();
  }
}
```

The parent subscription is added to the subscriber in `Observable.subscribe`:

```ts
const sink = toSubscriber(observerOrNext, error, complete);

if (operator) {
  operator.call(sink, this.source);
} else {
  sink._addParentTeardownLogic(
    this.source || (config.useDeprecatedSynchronousErrorHandling && !sink.syncErrorThrowable) ?
    this._subscribe(sink) :
    this._trySubscribe(sink)
  );
}
```

In numerous situations, the subscription passed to `_addParentTeardownLogic` will be the same subscriber. That is, it will be `sink`.

Adding `sink` to itself, as a parent, introduces a bug, as `sink` will be unsubscribed from within the `complete` method - which differs from the pre-#3963 mechanism.

#4072 introduced a check to ensure that a specific subscriber isn't added to itself as a parent. However, it's still possible for a subscriber that is a *child* of the specific subscriber to be added as a *parent*. And doing so will see the specific subscriber's `unsubscribe` called within its `complete` method - which can effect unwanted behaviour. See the failing test that's included in this PR.

This PR solves the problem by adding a `_keepAliveCount` that's incremented before unsubscribing the parent subscription and is decremented afterwards. In `Subscription.unsubscribe`, if `_keepAliveCount` is greater than zero, the subscription is not unsubscribed.

**Related issue (if exists):** #4095

### DIFF
--- a/spec/operators/mergeMap-spec.ts
+++ b/spec/operators/mergeMap-spec.ts
@@ -7,7 +7,7 @@ declare const type: Function;
 declare const asDiagram: Function;
 
 /** @test {mergeMap} */
-describe('mergeMap', () => {
+describe.only('mergeMap', () => {
   asDiagram('mergeMap(i => 10*i\u2014\u201410*i\u2014\u201410*i\u2014| )')
   ('should map-and-flatten each item to an Observable', () => {
     const e1 =    hot('--1-----3--5-------|');
--- a/spec/operators/mergeMap-spec.ts
+++ b/spec/operators/mergeMap-spec.ts
@@ -7,7 +7,7 @@ declare const type: Function;
 declare const asDiagram: Function;
 
 /** @test {mergeMap} */
-describe.only('mergeMap', () => {
+describe('mergeMap', () => {
   asDiagram('mergeMap(i => 10*i\u2014\u201410*i\u2014\u201410*i\u2014| )')
   ('should map-and-flatten each item to an Observable', () => {
     const e1 =    hot('--1-----3--5-------|');
--- a/spec/operators/mergeMap-spec.ts
+++ b/spec/operators/mergeMap-spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
-import { mergeMap, map, mapTo } from 'rxjs/operators';
-import { asapScheduler, concat, defer, Observable, from, of, timer } from 'rxjs';
+import { mergeMap, map } from 'rxjs/operators';
+import { asapScheduler, defer, Observable, from, of, timer } from 'rxjs';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 
 declare const type: Function;
@@ -772,18 +772,18 @@ describe('mergeMap', () => {
     const results: (number | string)[] = [];
 
     const wrapped = new Observable<number>(subscriber => {
-        const subscription = timer(0, asapScheduler).pipe(mapTo(42)).subscribe(subscriber);
+        const subscription = timer(0, asapScheduler).subscribe(subscriber);
         return () => subscription.unsubscribe();
     });
     wrapped.pipe(
-      mergeMap(value => concat(of(value), timer(0, asapScheduler).pipe(mapTo(value))))
+      mergeMap(() => timer(0, asapScheduler))
     ).subscribe({
       next(value) { results.push(value); },
       complete() { results.push('done'); }
     });
 
     setTimeout(() => {
-      expect(results).to.deep.equal([42, 42, 'done']);
+      expect(results).to.deep.equal([0, 'done']);
       done();
     }, 0);
   });
--- a/spec/operators/mergeMap-spec.ts
+++ b/spec/operators/mergeMap-spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
-import { mergeMap, map } from 'rxjs/operators';
-import { asapScheduler, defer, Observable, from, of } from 'rxjs';
+import { mergeMap, map, mapTo } from 'rxjs/operators';
+import { asapScheduler, concat, defer, Observable, from, of, timer } from 'rxjs';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 
 declare const type: Function;
@@ -717,7 +717,7 @@ describe('mergeMap', () => {
     // Added as a failing test when investigating:
     // https://github.com/ReactiveX/rxjs/issues/4071
 
-    const results: any[] = [];
+    const results: (number | string)[] = [];
 
     of(1).pipe(
       mergeMap(() => defer(() =>
@@ -744,7 +744,7 @@ describe('mergeMap', () => {
     // Added as a failing test when investigating:
     // https://github.com/ReactiveX/rxjs/issues/4071
 
-    const results: any[] = [];
+    const results: (number | string)[] = [];
 
     of(1).pipe(
       mergeMap(() =>
@@ -760,6 +760,30 @@ describe('mergeMap', () => {
 
     setTimeout(() => {
       expect(results).to.deep.equal([3, 'done']);
+      done();
+    }, 0);
+  });
+
+  it('should support wrapped sources', (done: MochaDone) => {
+
+    // Added as a failing test when investigating:
+    // https://github.com/ReactiveX/rxjs/issues/4095
+
+    const results: (number | string)[] = [];
+
+    const wrapped = new Observable<number>(subscriber => {
+        const subscription = timer(0, asapScheduler).pipe(mapTo(42)).subscribe(subscriber);
+        return () => subscription.unsubscribe();
+    });
+    wrapped.pipe(
+      mergeMap(value => concat(of(value), timer(0, asapScheduler).pipe(mapTo(value))))
+    ).subscribe({
+      next(value) { results.push(value); },
+      complete() { results.push('done'); }
+    });
+
+    setTimeout(() => {
+      expect(results).to.deep.equal([42, 42, 'done']);
       done();
     }, 0);
   });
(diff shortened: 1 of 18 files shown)


benlesh:
Thank you for bringing this issue to my attention. Taking a step back, I think that `_addParentTeardownLogic` is probably a smell, and we likely just wanted to `this.add(trustedSubscriber)` on [this line](https://github.com/ReactiveX/rxjs/blob/cea71968446b3dd6b11a23e4d51b4ddac81b0bae/src/internal/Subscriber.ts#L81)

cartant:
@benlesh This PR is now a WIP, as I've made some changes to make things simpler, but the changes need to be applied to many of the operator-specific subscribers and I've only changed `MergeMapSubscriber`, so far.

I've removed all of the parent-teardown business and the `_keepAliveCount`.

The `_addParentTeardownLogic` function is still there - as a noop - as its used in `isTrustedSubscriber` and I'm not sure that you'll want to do with that if `_addParentTeardownLogic` is no longer needed.

Anyway, the important parts are in `MergeMapSubscriber`'s [`_innerSub` function](https://github.com/ReactiveX/rxjs/blob/041cd2f86a56de18f8bd1e986d361861e797ae84/src/internal/operators/mergeMap.ts#L143-L144), where the inner subscription is added to the destination (which should always be a trusted subscriber). And in its [`_complete` function](https://github.com/ReactiveX/rxjs/blob/041cd2f86a56de18f8bd1e986d361861e797ae84/src/internal/operators/mergeMap.ts#L153), where `this.unsubscribe()` is called.

To me this is simpler than dealing with children, parents and destinations. In the original implementation - that preceded the PR that sought to address the not-unsubscribing-from-source problem - a subscription was added to its destination so that unsubscribing the destination unsubscribes the subscription. That behaviour is now restored. And because the subscription's destination is not added to the subscription, calling `this.unsubscribe()` in `_complete` does not unsubscribe the destination - or any inner observables that were added to it.

If you think this approach is a better way to go, I can apply similar changes to the other subscribers. (It's not just the flattening subscribers. E.g. the subscribers for `delay` and `delayWhen` will need to be changed, too.)

We would need to decide the best way of adding subscriptions to the destination, as the type assertions are rather ugly:

```ts
const destination = this.destination as any as Subscription;
```

spec/operators/mergeMap-spec.ts
```diff
@@ -1,13 +1,13 @@
...
+import { mergeMap, map, mapTo } from 'rxjs/operators';
+import { asapScheduler, concat, defer, Observable, from, of, timer } from 'rxjs';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 
 declare const type: Function;
 declare const asDiagram: Function;
 
 /** @test {mergeMap} */
-describe('mergeMap', () => {
+describe.only('mergeMap', () => {
```
benlesh: I know this is a WIP, but don't forget to remove this `only`.

src/internal/operators/mergeMap.ts
```diff
@@ -140,7 +140,8 @@ export class MergeMapSubscriber<T, R> extends OuterSubscriber<T, R> {
 
   private _innerSub(ish: ObservableInput<R>, value: T, index: number): void {
     const innerSubscriber = new InnerSubscriber(this, undefined, undefined);
-    this.add(innerSubscriber);
+    const destination = this.destination as any as Subscription;
```
benlesh: destination is a `Subscriber`, which extends `Subscription`... I'm not sure why any casting is necessary here... :thinking:

src/internal/Subscriber.ts
```diff
@@ -158,16 +156,7 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
 
   /** @deprecated This is an internal implementation detail, do not use. */
   _addParentTeardownLogic(parentTeardownLogic: TeardownLogic) {
```
benlesh: We can probably just remove _addParentTeardownLogic entirely.

1. It's prefixed with `_` which is a universal signal for "do not use".
2. It's been marked as "deprecated do not use" since the moment it was added.

src/internal/operators/mergeMap.ts
cartant: Nope, it's a `PartialObserver`:

```ts
protected destination: PartialObserver<any>;
```

src/internal/Subscriber.ts
cartant: My concern is that the name has been used in [the `isTrustedSubscriber` test](https://github.com/ReactiveX/rxjs/blob/6.3.2/src/internal/Subscriber.ts#L328-L330) and that's been released. I wasn't entirely sure of what the affect of it's removal in a new release would be. I suppose it means a trusted subscriber could end up being re-wrapped in a using-6.3.3-with-6.3.2 scenario.

src/internal/operators/mergeMap.ts
cartant: Changing `destination` to be:

```ts
protected destination: PartialObserver<any> | Subscriber<any>;
```

means the type assertion becomes:

```ts
const destination = this.destination as Subscription;
```

cartant:
@benlesh This should be done, now.

benlesh:
@cartant I think we should schedule a time to walk through this one.

cartant:
What follows are notes on the history of changes made in this PR.

#3963 (which was a refactoring of #2457) made changes to the way in which subscribers were wired up. The changes were made so that teardown logic for sources was called as soon as sources completed or errored.

Its changes were:

* Instead of just adding a subscriber to its destination, it was added using a new method: `_addParentTeardownLogic`.
* In `Observable.prototype.subscribe`, the result from `_trySubscribe` was also added using `_addParentTeardownLogic`.
* When `complete` or `error` notifications occurred, subscribers would unsubscribe the teardown login added via `_addParentTeardownLogic`.

The major problem with this approach was that a subscriber could be added to itself, as `_trySubscribe` can return the subscriber that's passed to it.

This PR added a [test](https://github.com/cartant/rxjs/blob/b817f539b9696bd7afd0dfab8aeb1ee0b5278c44/spec/operators/mergeMap-spec.ts#L774-L783) that included this code:

```ts
const wrapped = new Observable<number>(subscriber => {
    const subscription = timer(0, asapScheduler).subscribe(subscriber);
    return () => subscription.unsubscribe();
});
wrapped.pipe(
  mergeMap(() => timer(0, asapScheduler))
).subscribe({
  next(value) { results.push(value); },
  complete() { results.push('done'); }
});
```

When run with the changes in #3963, the sequence of subscribers/subscriptions was:

```
construct Subscription@1
construct Subscriber@2
construct SafeSubscriber@3
construct MergeMapSubscriber@4
add MergeMapSubscriber@4 to Subscriber@2
construct AsapAction@5
add AsapAction@1 to MergeMapSubscriber@4
construct Subscription@6
add Subscription@6 to MergeMapSubscriber@4
next MergeMapSubscriber@4
construct InnerSubscriber@7
add InnerSubscriber@7 to MergeMapSubscriber@4 <--
construct AsapAction@8
add AsapAction@2 to InnerSubscriber@7
complete MergeMapSubscriber@4
unsubscribe Subscription@6
unsubscribe MergeMapSubscriber@4
unsubscribe AsapAction@1
unsubscribe InnerSubscriber@7
unsubscribe AsapAction@2
```

Note that the `InnerSubscriber` was added to the `MergeMapSubscriber`. That means that the `InnerSubscriber` was unsubscribed when the `MergeMapSubscriber` was unsubscribed. And that was happen when the source - the wrapped observable - completed.

This PR reverted the changes in #3963 and made the following changes to a number of operators:

* Inner subscribers are now added to the destination subscriber.
* Outer subscribers now unsubscribe upon `complete` and `error`, with their inner subscribers remaining subscribed due to their having been added to the destination.

With these changes, the test now passes and the sequence of subscribers/subscriptions is:

```
construct Subscription@1
construct Subscriber@2
construct SafeSubscriber@3
construct MergeMapSubscriber@4
add MergeMapSubscriber@4 to Subscriber@2
construct AsapAction@5
add AsapAction@1 to MergeMapSubscriber@4
construct Subscription@6
add Subscription@6 to MergeMapSubscriber@4
next MergeMapSubscriber@4
construct InnerSubscriber@7
add InnerSubscriber@7 to Subscriber@2 <--
construct AsapAction@8
add AsapAction@2 to InnerSubscriber@7
complete MergeMapSubscriber@4
unsubscribe MergeMapSubscriber@4
unsubscribe AsapAction@1
unsubscribe Subscription@6
next InnerSubscriber@7
next Subscriber@2
complete InnerSubscriber@7
complete Subscriber@2
unsubscribe SafeSubscriber@3
unsubscribe Subscriber@2
unsubscribe InnerSubscriber@7
unsubscribe AsapAction@2
```

Note that the `InnerSubscriber` is now added to the destination `Subscriber` and the is not unsubscribed upon completion of the source. And is able to `next` its value, etc.

The changes in this PR affected a number of behaviour tests in [`observeOn-spec.ts`](https://github.com/cartant/rxjs/blob/b817f539b9696bd7afd0dfab8aeb1ee0b5278c44/spec/operators/observeOn-spec.ts#L86-L126) and in [`switch-spec.ts`](https://github.com/cartant/rxjs/blob/b817f539b9696bd7afd0dfab8aeb1ee0b5278c44/spec/operators/switch-spec.ts#L222-L265) because adding the inner subscibers to the destination changed the subscription counts that those tests were checking.

Issue #3553, opened 2018-04-12
Documentation isn't clear about the differences between Observable / Subject


SetTrend:
From the documentation it doesn't get clear what the repercussions of prefering a `Subject` over an `Observable` are.
<br/>

## (a) Compare these two explanations:
<hr/>

### [ReactiveX/rxjs/doc/observable.md](https://github.com/ReactiveX/rxjs/blob/e91d113e49acc9b0e12e2524040fa84cf3fa2c78/doc/observable.md)

You can write the same behavior above, but with Observables:

```js
var foo = Rx.Observable.create(function (observer) {
  console.log('Hello');
  observer.next(42);
});

foo.subscribe(function (x) {
  console.log(x);
});
foo.subscribe(function (y) {
  console.log(y);
});
```

And the output is the same:

```none
"Hello"
42
"Hello"
42
```
<hr/>

### [ReactiveX/rxjs/doc/subject.md](https://github.com/ReactiveX/rxjs/blob/e91d113e49acc9b0e12e2524040fa84cf3fa2c78/doc/subject.md)

In the example below, we have two Observers attached to a Subject, and we feed some values to the Subject:

```js
var subject = new Rx.Subject();

subject.subscribe({
  next: (v) => console.log('observerA: ' + v)
});
subject.subscribe({
  next: (v) => console.log('observerB: ' + v)
});

subject.next(1);
subject.next(2);
```

With the following output on the console:

```none
observerA: 1
observerB: 1
observerA: 2
observerB: 2
```
<hr/>

These explanations basically show the same behaviour.

<br/>

## (b) The reference page on `Subject` doesn't have any comment/remark on the peculiarities of that class:
![subject](https://user-images.githubusercontent.com/9283914/38668135-b936b0c6-3e43-11e8-8da6-41e0162ef58b.png)
<hr/>

<br/>

I feel the documentation should further elaborate on this part for learning and offline reading.

MichaelSolati:
@benlesh It's not marked as a `status: PRs welcome` but I wouldn't mind doing some clarification of the docs. This as well as other bits as/if they show up.

benlesh:
@MichaelSolati feel free. :+1:

cartant:
Closing this as the associated PR has been merged.

PR #4107, opened 2018-09-07
docs(Subject): Clarity for Subject Implementations

ajcrites:
This is at least partially #3553.

**Description:** This includes some basic descriptions about the various `Subject` classes. The documentation was acquired from doc/subject.md and contextually rewritten.

**Related issue (if exists):** #3553

### DIFF
--- a/src/internal/Subject.ts
+++ b/src/internal/Subject.ts
@@ -8,6 +8,12 @@ import { SubjectSubscription } from './SubjectSubscription';
 import { rxSubscriber as rxSubscriberSymbol } from '../internal/symbol/rxSubscriber';
 
 /**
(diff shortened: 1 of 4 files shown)


src/internal/Subject.ts
```diff
@@ -8,6 +8,12 @@ import { SubjectSubscription } from './SubjectSubscription';
 import { rxSubscriber as rxSubscriberSymbol } from '../internal/symbol/rxSubscriber';
 
 /**
+ * A Subject is a special type of Observable that allows values to be
```
JWO719: I think here is a minor error. This part belongs to the Subject class, directly below the SubjectSubscriber

ajcrites:
Updated, silly mistake!

Issue #3953, opened 2018-07-25
Doc pages missing / misleading directions in README?


robations:
### Documentation Related To Component:

RxJS

### Please check those that apply

- [ ] typo
- [x] documentation doesn't exist OR
- [x] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

### Description Of The Issue

In the doc/README.md page it says:

> These files are not meant for reading directly in GitHub, they are just source code for generating the official page. You should find the docs at http://reactivex.io/rxjs/, containing all the documentation.

However, either most pages in the doc/ directory are missing from the reactivex.io site, or they're really hard to find. For example, I was looking for a page to match "writing-marble-tests.md" but can only see: Overview, Installation, Migration and the auto-generated API docs.

Would guess either:

- README.md is wrong — you are better off reading directly in GitHub at the moment;
- Some work needed to add these files to the docs site; or
- The docs are there somewhere and navigation can be improved.

Sorry if this is in the wrong place or a duplicate, I spent a few minutes looking through existing bugs open and closed but didn't see anything similar.

P.S. It seems like the doc/ folder is in the process of being phased out in favour of the work in docs_app/ ? Might suggest a note is added to clarify.

mattyao1984:
@JWO719 please check this issue and let me know anything I can help with.

JWO719:
Hi @robations, so basically we are currently reworking the documentation. And yes you are right, docs_app is what we label "new docs". We don't want to migrate all the stuff, because many of those are outdated, or not easy to understand and stuff like that. But at this moment it would be really helpfull if you could say what you are struggeling with/ searching for so that we can cover those parts in some kind of tutorial or something likewise. Is there anything else the new docs "http://rxjs-dev.firebaseapp.com/" could help you with this issue?
Regarding the reactivex.io page. I think the root will redirect to the "new docs" but if you directly try to open some of the urls this should work.

JWO719:
So personally improving the developer experience is my main goal, therefore I try to improve the documentation :D So thank you very much for this valuable feedback, I'll take care of it

ajcrites:
@JWO719 are there plans to migrate particular pages? The current docs are missing a lot of the useful "Overview" pages including the Observable and Subject (e.g. https://github.com/ReactiveX/rxjs/blob/master/doc/subject.md). I would like to help to update some of these documentation pages to new examples, but I'm not sure of the progress on that or whether there are plans to do away with some of these pages.

[...]

PR #4108, opened 2018-09-07
docs(subject): Subject documentation for new docs app.

ajcrites:
**Description:**

* Updated subject.md from old docs for consistency with the new API
* Copied and linked this to the new docs app

I'm not sure if we want to remove things that were copied over from `doc`. I'm also not sure whether the migration / update was done properly. I'm happy to address any feedback.

**Related issue (if exists):** #3953 (partial)


pertrai1:
This looks good. One note is that the newer examples are showing the output in the same code example markdown rather than a newer one below. Example: https://github.com/ReactiveX/rxjs/blob/master/src/internal/observable/forkJoin.ts#L85

docs_app/content/guide/subject.md
```diff
@@ -0,0 +1,364 @@
+# Subject
+
+**What is a Subject?** An RxJS Subject is a special type of Observable that allows values to be multicasted to many Observers. While plain Observables are unicast (each subscribed Observer owns an independent execution of the Observable), Subjects are multicast.
+
+<span class="informal">A Subject is like an Observable, but can multicast to many Observers. Subjects are like EventEmitters: they maintain a registry of many listeners.</span>
```
pertrai1: I have wondered what is the purpose of this span class used for?

docs_app/content/guide/subject.md
```diff
@@ -0,0 +1,364 @@
...
+  next: (v) => console.log(`observerB: ${v}`)
+});
+
+subject.next(1);
+subject.next(2);
+```
+
+With the following output on the console:
+
+```none
```
pertrai1: I would combine this with the above code example just by showing the log below:

```
// Logs:
// observerA: 1
// observerB: 1
// observerA: 2
// observerB: 2
```

docs_app/content/guide/subject.md
pertrai1: I found it. It is supposed to be used as the blockquote but the new docs does not have the css in place to show it that way.

ajcrites:
Comments addressed. I didn't really know where to put the style. I also changed it to just italicize the text instead of use the cursive font since it seems like Kalam isn't included. The cursive is a little _too_ informal.

docs_app/content/navigation.json
```diff
@@ -24,7 +24,13 @@
 		{
 			"url": "guide/overview",
 			"title": "Overview",
-			"tooltip": "RxJS Overview"
+			"tooltip": "RxJS Overview",
+			"children": [
+				{
+					"url": "guide/subject",
+					"title": "Subjects in RxJS"
```
JWO719: 'Subjects' would be totally fine for the title. I think it is totally clear that it is about rxjs :D

JWO719: 'Subjects' would be totally fine for the title. I think it is totally clear that it is about rxjs :D

docs_app/content/guide/subject.md
```diff
@@ -0,0 +1,343 @@
...
+
+subject.next(5);
+subject.complete();
+
+// Logs:
+// observerA: 5
+// observerB: 5
+```
+
+The AsyncSubject is similar to the [`last()`](../class/es6/Observable.js~Observable.html#instance-method-last) operator, in that it waits for the `complete` notification in order to deliver a single value.
```
JWO719: I would link to last with: `["last"](/api/operators/last)`

docs_app/content/navigation.json
ajcrites: I was just trying to clarify that this was "Subjects, the RxJS construct" vs. "Subjects (e.g. topics) about RxJS)," mostly for newer readers.  I can just change it to "Subjects" though -- that may be clear enough.

ajcrites:
Updated for requested changes.

Issue #4109, opened 2018-09-07
static pipe operator is not the same as Observable.pipe


dkosasih:
## Bug Report

**Current Behavior**
Static pipe function in the recently merged fixed (#3945) for Observable pipe function signature overload is not in sync. 

**Reproduction**
Rest parameters overload like below show also exists in static pipe function.
```js
pipe<A, B, C, D, E, F, G, H, I>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>, op5: OperatorFunction<D, E>, op6: OperatorFunction<E, F>, op7: OperatorFunction<F, G>, op8: OperatorFunction<G, H>, op9: OperatorFunction<H, I>, ...operations: OperatorFunction<any, any>[]): Observable<{}>;
```

**Expected behavior**
It should behave the same as Observable pipe function and be able to compose static pipe functions separately and pass on to Observable pipe.

**Environment**
- RxJS version: 6.3.2

cartant:
Closing this, as it has been addressed in a PR.

PR #4112, opened 2018-09-07
fix(pipe): align static pipe to Observable pipe rest parameters overl…

dkosasih:
**Description:**
Bring static pipe rest parameters overload to be in line with Observable.pipe 

#4109


dkosasih:
OK. I will work on that. Thanks 😊

dkosasih:
@cartant  I have added new commit for the tests.

dkosasih:
That was quick. Thanks 👍

jgbpercy:
pipeFromArray could also drop the type params now that no one calls it with anything other than <any, any> ?

cartant:
@jgbpercy Yes, `pipeFromArray` has a signature that needs to be fixed, too, but I'd prefer that to be done in a separate PR. You're welcome to open one.

benlesh:
currently, `pipeFromArray` is an internal API, however.

Issue #4000, opened 2018-08-09
improve takeUntil and takeWhile to have overload to includeLastValue = true


ianadavies:
## Feature Request

takeUntil and takeWhile does not include last value.

Please see this post, there are links in the comments to other users having similar problems. 
https://stackoverflow.com/questions/51767204/rxjs-interval-with-takeuntil-to-publish-last-value

**Describe the solution you'd like**
Need option to get last value of takeUntil and takeWhile

**Describe alternatives you've considered**
Workaround is in stack overflow post but is not clean.

simeyla:
I would really value this too. I very often need something 'sticky' where I basically want the stream to end when a certain value is reached. When you know the last value you can do the following:

  public showBall$ = this.showBall$.pipe(takeWhile(x => x == false), endWith(true));

So here we keep checking if we should show the ball, until it is true and then we complete with an explicit value of true so the ball always shows.

Currently the `takeUntil(X)` operator takes an observable. There is no overload with a predicate. Therefore it's pretty unambiguous (to me at least) that you won't get the 'last' value emitted from the source stream when you use this. This definitely shouldn't change - take until X happens (something outside our stream) is quite clear.

So what you're really asking for here is `takeUntil(predicate)` which emits the final value. However there's a problem with this. If we added that we'd have `takeUntil(predicate)` that *did* emit the final (non predicate matching) value, and `takeUntil(obs)` that didn't. 

So it would need to be named differently:

`takeUntil(predicate, includeLastValue)` seems clumsy
`takeUntilInclusive(predicate)` also seems clumsy
`takeWhile(x => x != FOO), endWith(FOO)` is something I'd like to get away from

My best suggestion so far would be a new operator:

`endWhen(predicate)` which would be very 'googlable' and explicit in what it does

benlesh:
Given that I have people inside of Google making a custom operator to suit this need, I think it's time to discuss implementing it.

### Proposal

Add a second argument boolean `inclusive`, that will include the first value to return false for `takeWhile`.

### Workaround

Until then, there are super-gross workarounds that use current operators like:

```ts
const takeWhileInclusive = (predicate) => (source) =>
  source.pipe(
    concatMap((value, i) => prediciate(value, i) ? of({ value }) : of({ value }, { kill: true })),
    takeWhile(({ kill }) => !kill),
    map(({ value }) => value),
  );

// usage

source$.pipe(takeWhileInclusive(predicate));
```

idosela:
[...]

PR #4115, opened 2018-09-07
feat(takeWhile): add `inclusive` option to the operator which causes it to emit the final value

idosela:
Description:
By default, the value that causes the predicate to return `false` is not emitted.

This change adds an `inclusive` option which changes the behavior of the operator to include the final value which made the predicate return `false`.
Typical use case: Polling an API until the response contains a certain value, and then doing something with the response.

Related issue: #4000

### DIFF
--- a/src/internal/operators/takeWhile.ts
+++ b/src/internal/operators/takeWhile.ts
@@ -4,7 +4,8 @@ import { Subscriber } from '../Subscriber';
 import { OperatorFunction, MonoTypeOperatorFunction, TeardownLogic } from '../types';
 
 export function takeWhile<T, S extends T>(predicate: (value: T, index: number) => value is S): OperatorFunction<T, S>;
-export function takeWhile<T>(predicate: (value: T, index: number) => boolean): MonoTypeOperatorFunction<T>;
+export function takeWhile<T, S extends T>(predicate: (value: T, index: number) => value is S, inclusive: false): OperatorFunction<T, S>;
+export function takeWhile<T>(predicate: (value: T, index: number) => boolean, inclusive?: boolean): MonoTypeOperatorFunction<T>;
 
 /**
  * Emits values emitted by the source Observable so long as each value satisfies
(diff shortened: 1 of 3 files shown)


idosela:
@benlesh I keep getting this error:
https://travis-ci.org/ReactiveX/rxjs/jobs/426812416

As far as I can tell I updated all of the required files, and all test are passing when I run them on my machine. I'm probably missing something super-obvious, but I can't figure it out. Any help would be appreciated.

compat/operator/takeWhile.ts
```diff
@@ -34,9 +34,13 @@ import { takeWhile as higherOrder } from 'rxjs/operators';
...
+ * `predicate` to return `false` will also be emitted.
  * @method takeWhile
  * @owner Observable
  */
-export function takeWhile<T>(this: Observable<T>, predicate: (value: T, index: number) => boolean): Observable<T> {
-  return higherOrder(predicate)(this);
+export function takeWhile<T>(this: Observable<T>,
+                             predicate: (value: T, index: number) => boolean,
+                             inclusive?: boolean): Observable<T> {
+  return higherOrder<T>(predicate, inclusive)(this);
```
benlesh: Because this is a new feature, we don't need to update the backward compatibility code. You can revert this file.

timdeschryver:
I think these changes should also be added to the [dtslint](https://github.com/ReactiveX/rxjs/blob/master/spec-dtslint/operators/takeWhile-spec.ts) test specs, @cartant ?

src/internal/operators/takeWhile.ts
```diff
@@ -36,22 +36,29 @@ import { MonoTypeOperatorFunction, TeardownLogic } from '../types';
...
  * Observable so long as each value satisfies the condition defined by the
  * `predicate`, then completes.
  * @method takeWhile
  * @owner Observable
  */
-export function takeWhile<T>(predicate: (value: T, index: number) => boolean): MonoTypeOperatorFunction<T> {
-  return (source: Observable<T>) => source.lift(new TakeWhileOperator(predicate));
+export function takeWhile<T>(
+    predicate: (value: T, index: number) => boolean,
+    inclusive = false): MonoTypeOperatorFunction<T> {
```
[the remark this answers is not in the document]
idosela: Rebased. Sorry about that. That also explains the issues I had with the build. 
Added the dtslint tests as requested.

idosela:
@cartant I've made the required fixes. Can you please take a look and let me know if there are any other issues I need to address?

src/internal/operators/takeWhile.ts
```diff
@@ -3,8 +3,8 @@ import { Observable } from '../Observable';
 import { Subscriber } from '../Subscriber';
 import { OperatorFunction, MonoTypeOperatorFunction, TeardownLogic } from '../types';
 
-export function takeWhile<T, S extends T>(predicate: (value: T, index: number) => value is S): OperatorFunction<T, S>;
-export function takeWhile<T>(predicate: (value: T, index: number) => boolean): MonoTypeOperatorFunction<T>;
+export function takeWhile<T, S extends T>(predicate: (value: T, index: number) => value is S, inclusive?: boolean): OperatorFunction<T, S>;
```
cartant: Cool, but there'll need to be a few more signatures. The problem is that if `inclusive` can be `true`, the type guard has no meaning.

For example, consider a stream that is emitting carrots and potatoes and a type guard that accepts only carrots. If `inclusive` is `true`, the potato that completes the stream will be emitted, so the type guard should not narrow the emitted types. (Thanks go to @benlesh for the vegetable analogy.)

You can solve that problem by narrowing with a type guard only if `inclusive` is either omitted or is known to be `false`, like this:

```ts
export function takeWhile<T, S extends T>(predicate: (value: T, index: number) => value is S): OperatorFunction<T, S>;
export function takeWhile<T, S extends T>(predicate: (value: T, index: number) => value is S, inclusive: false): OperatorFunction<T, S>;
export function takeWhile<T>(predicate: (value: T, index: number) => boolean, inclusive: boolean): MonoTypeOperatorFunction<T>;
```

Thanks for the effort you've put into this PR.

cartant: And if you could add another dtslint in which you explicitly pass `false` for `inclusive` - to make sure the signature with the `false` literal type returns a narrowed observable - that would be great.

idosela: Makes sense. Thanks for the detailed explanation. I made the requested changes.

idosela:
@cartant Can you please take another look and let me know if there is anything else I need to fix?

cartant:
@idosela Sure. I'll look at it again, later today. Sorry for forgetting about it.

felixfbecker:
In need of this too :)

Issue #4093, opened 2018-09-05
Increase dtslint test coverage


cartant:
## Chore

https://github.com/ReactiveX/rxjs/pull/3944 added [`dtslint`](https://github.com/Microsoft/dtslint) to the Travis `$FULL_VALIDATE` configuration. However, the coverage of the codebase is far from complete.

PRs that contribute to increasing the coverage are welcome.

The purpose of the `dtslint` tests is to ensure that the types are inferred as expected and to ensure that type errors are effected for invalid types. At the moment, there are some `type` tests in the codebase. These are able to do the former, but not the latter - as type errors would break the build.

For an example of the `type` tests, see [these for the `zip` observable](https://github.com/ReactiveX/rxjs/blob/6.3.0/spec/observables/zip-spec.ts#L580-L627) and [these for the `zip` operator](https://github.com/ReactiveX/rxjs/blob/master/spec/operators/zip-spec.ts#L594-L616). The `type` tests are not executed when the tests run. Rather, they're tested when the tests are built. And if the tests build, the `type` tests have passed. Not all operators and operators have `type` tests. In fact, many don't.

The `dtslint` tests reside in the `spec-dtslint` directory. They are written as Mocha-style `it` tests, but they are not run. Instead, TSLint runs rules that test the inferred types against expectations that are expression in comments. For example:

```ts
it('should support observables', () => {
  const a = of(1); // $ExpectType Observable<number>
  const b = of('foo'); // $ExpectType Observable<string>
  const c = of(true); // $ExpectType Observable<boolean>
  const o1 = zip(a, b, c); // $ExpectType Observable<[number, string, boolean]>
});
```

You can find the `type` tests that have been converted to `dtslint` tests for the `zip` observable [here](https://github.com/ReactiveX/rxjs/blob/6.3.0/spec-dtslint/observables/zip-spec.ts) and for the `zip` operator [here](https://github.com/ReactiveX/rxjs/blob/6.3.0/spec-dtslint/operators/zip-spec.ts).

Apart from converting the `type` tests, we want to ensure that there are `dtslint` tests for each overload signature for each observable and operator.

Submitting separate PRs - each with one or only a few observables or operators - would be the best way to approach this, as PRs with a limited number of observables or operators should be straightforward to review.

**Running the tests**

The tests will run on the Travis CI build, but you can run them locally using the `dtslint` script:

```
npm run dtslint
```

[...]

PR #4120, opened 2018-09-09
test(dtslint): add throttle

timdeschryver:
**Description:**
This PR adds dtslint tests for `throttle`.

### DIFF
--- a/spec-dtslint/operators/throttle-spec.ts
+++ b/spec-dtslint/operators/throttle-spec.ts
@@ -1,0 +1,25 @@
+import { of, timer } from 'rxjs';
+import { throttle } from 'rxjs/operators';
+
+it('should infer correctly', () => {
+  const o = of(1, 2, 3).pipe(throttle(() => timer(47))); // $ExpectType Observable<number>
+});
+
+it('should infer correctly with a Promise', () => {
+  const o = of(1, 2, 3).pipe(throttle(() => new Promise<boolean>(() => {}))); // $ExpectType Observable<number>
+});
+
+it('should support a config', () => {
+  const o = of(1, 2, 3).pipe(throttle(() => timer(47), { leading: true, trailing: true })); // $ExpectType Observable<number>
+});
+
+it('should enforce types', () => {
+  const o = of(1, 2, 3).pipe(throttle()); // $ExpectError
+  const p = of(1, 2, 3).pipe(throttle(() => {})); // $ExpectError
+});
+
+it('should enforce config types', () => {
+  const o = of(1, 2, 3).pipe(throttle(() => timer(47), { x: 1 })); // $ExpectError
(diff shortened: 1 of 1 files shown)


spec-dtslint/operators/throttle-spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+it('should enforce types', () => {
+  const o = of(1, 2, 3).pipe(throttle()); // $ExpectError
+  const p = of(1, 2, 3).pipe(throttle(() => {})); // $ExpectError
+});
+
+it('should enforce config types', () => {
+  const o = of(1, 2, 3).pipe(throttle(() => timer(47), { x: 1 })); // $ExpectError
+  const p = of(1, 2, 3).pipe(throttle(() => timer(47), { leading: 1, trailing: 1 })); // $ExpectError
+  const q = of(1, 2, 3).pipe(throttle(() => timer(47), null)); // $ExpectError
```
benlesh: Hmm... this seems like a bug. Should be `Observable<number>`. We need to improve our checks in this operator.

timdeschryver: Would this mean that we would default to the default config when `null` is passed? 
I might be wrong, but I think most of the operators accepting a config parameter have the same behavior of "throwing" an error e.g. `throttleTime`

timdeschryver: Ah, it seems that only `throttle` and `throttleTime` has a config parameter, I'll update both operators.
Just to be sure, is this the kind of behavior we want?
> Would this mean that we would default to the default config when null is passed?

cartant: Both [`throttle`](https://github.com/ReactiveX/rxjs/blob/6.3.2/src/internal/operators/throttle.ts#L66) and [`throttleTime`](https://github.com/ReactiveX/rxjs/blob/6.3.2/src/internal/operators/throttleTime.ts#L55) use a parameter default for the `config` object, so passing `null` should be an error. For the default to be used either the parameter would have to be omitted or `undefined` would have to be passed. That's my understanding, anyway, so I think the tests are okay.

cartant: To accept a `null` the implementation will have to change. There is also a separate issue with some parts of the RxJS API being typed as accepting `null` parameters and others as accepting `undefined`. Both `null` and `undefined` should be acceptable.

cartant:
@benlesh Not merging this - or #4121 - because of [this comment](https://github.com/ReactiveX/rxjs/pull/4120#discussion_r216427027). I think the tests are okay. With the current implementation, passing `null` won't effect the use of the default configuration. Changing this would be a separate issue that would necessitate changes to the implementation.

benlesh:
Thanks, @timdeschryver

Issue #3272, opened 2018-01-30
Add information on README about documentation effort


benlesh:
We should be helping people find and contribute to the docs, since more people put in docs PRs than anything else. 

@ladyleet can you have someone put together a .md file on how to contribute to the docs? Or link anyhting you have that already exists in the rxjs master README?

PR #4126, opened 2018-09-10
docs(readme): improve the readme to point to the new docs

JWO719:
I hope that this pr makes the different usages of the `docs` and the `docs_app` folder clearer


JWO719:
Maybe we can close #3272 afterwards

Issue #4110, opened 2018-09-07
Static pipe not able to return generic type


dkosasih:
## Bug Report

**Current Behavior**
Using static pipe to compose operators will lose the return type.

**Reproduction**
```js
const pipes = pipe(debounceTime(500), delay(100));
const obs = timer(0, 1000).pipe(take(5));
obs.pipe(pipes).subscribe((res)=> console.log(res)); // Typing for res is lost
```

**Expected behavior**
Should be able to detect number

**Environment**
- RxJS version: 6.3.2

PR #4129, opened 2018-09-10
fix(pipe): change static pipe overload to reflect source observable t…

dkosasih:
**Description:** add overloads to static pipes functions to enable typing to reflect the correct type from source Observable.

**Related issue (if exists):**  #4110

### DIFF
--- a/spec-dtslint/util/pipe-spec.ts
+++ b/spec-dtslint/util/pipe-spec.ts
@@ -1,4 +1,4 @@
-import { pipe, UnaryFunction } from 'rxjs';
+import { pipe, UnaryFunction, of, Observable } from 'rxjs';
 
 function a<I extends string, O extends string>(input: I, output: O): UnaryFunction<I, O> {
   return i => output;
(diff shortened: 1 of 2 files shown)


spec-dtslint/util/pipe-spec.ts
```diff
@@ -75,3 +75,10 @@ it('should enforce types for the 8th argument', () => {
 it('should enforce types for the 9th argument', () => {
   const o = pipe(a('0', '1'), a('1', '2'), a('2', '3'), a('3', '4'), a('4', '5'), a('5', '6'), a('6', '7'), a('7', '8'), a('#', '9')); // $ExpectError
 });
+
+it('should inherit typings from source Observable ', () => {
```
cartant: I wouldn't use the word 'inherit' here - it has a specific meaning. I'd say that it should return a generic `OperatorFunction`.

dkosasih:
@cartant  what would you suggest about the commit type? I personally don't think it's a chore, not sure of it's feat.

cartant:
`fix` is fine, but the explanation needs to be like the related issue, IMO:

```
fix(pipe): static pipe should return generic functions when possible
```

spec-dtslint/util/pipe-spec.ts
```diff
@@ -75,3 +75,15 @@ it('should enforce types for the 8th argument', () => {
...
+it('should return generic OperatorFunction', () => {
+  const customOperator = <T>() => (a: Observable<T>) => a;
+
+  const staticPipe = pipe(customOperator());
+  const o = of('foo').pipe(staticPipe); // $ExpectType Observable<string>
+});
+
+it('should return generic function', () => {
+  const func = pipe((value: string) => value);
+  const value = func("foo"); // ExpectType "foo"
```
benlesh: `$ExpectType`

dkosasih: oppps... but, running dtslint does not throw error. I wonder if there is a way to prevent this kind of typo.

cartant: No error, because it's just seen as a normal comment. At some stage, I'll write a TSLint rule to pick this up, but dealing with comments in the AST is tedious.

PR #3872, opened 2018-06-28
docs(test): Added docs and examples for shareReplay operator

mattyao1984:
**Description:**
Added docs and code examples for shareReplay operator.

**Related issue (if exists):**
https://github.com/ReactiveX/rxjs/issues/3819

PR #4130, opened 2018-09-11
perf(shareReplay): use correct default paramter

nename0:
**Description:**
shareReplay() uses a wrong default value for `windowTime` parameter since #3872. To disable the time function of the internal ReplaySubject the `windowTime` must be `Number.POSITIVE_INFINITY` and not `Number.MAX_VALUE`.
This makes shareReplay() faster because the ReplaySubject will not call `scheduler.now()`, when it is not needed.

This PR also adds a test, whichs checks if the ReplaySubject is in `_infiniteTimeWindow` mode, when no `windowTime` is given to shareReplay().

This will need merging with #4059 whatever comes first.

### DIFF
--- a/spec/operators/shareReplay-spec.ts
+++ b/spec/operators/shareReplay-spec.ts
@@ -1,8 +1,9 @@
 import { expect } from 'chai';
+import * as sinon from 'sinon';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 import { shareReplay, mergeMapTo, retry, take } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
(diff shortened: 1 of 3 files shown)


nename0:
I also added a test to ReplaySubject to get coverage back up.

cartant:
Thanks for changing the test to use `sinon`. However, something appears to have happened to the alignment of the other marble diagrams in `spec/operators/shareReplay-spec.ts`. Do you have some sort of automated formatting enabled?

Regarding the coverage, it won't have been anything you did. The coverage seems to move around +/- 0.2% seemingly on a whim.

I think I'll have to look at the additional test when I'm less tired. I can't get my head around it, at the moment.

Thanks again for the PR.

PR #4132, opened 2018-09-12
chore(dtslint): add rule to catch missing $

cartant:
**Description:**

This PR adds a TSLint rule that's included in the dtslint configuration to catch dtslint expectation comments that are missing a `$` character - as in [this PR](https://github.com/ReactiveX/rxjs/pull/4129#discussion_r216817502).

The rule is really simple and the TS source - and tests - are [here](https://github.com/cartant/tslint-etc/blob/master/source/rules/noMissingDollarExpectRule.ts). I've just copied the TypeScript output into the `rules` directory. And added a reference to the MIT licence. Doing so was much less complicated than dealing with a `node_modules`-based install, etc.


kwonoj:
we can't install `tslint-etc` as devdep?

cartant:
I guess we can do that, if that's what you'd prefer. Any rules in it should be included if the `spec-dtslint/tslint.json` extends `tslint-etc`.

If you want me to change it to a `devDependency`, just let me know.

kwonoj:
I'm generally fine with devdependency, we don't have lot and it doesn't impact installation time much.

benlesh:
@cartant can you rebase this one please?

cartant:
Okay, it looks like this should now be fixed.

PR #4137, opened 2018-09-14
docs(observable): update observable docs real quick

benlesh:
This is just a quick update to the existing observable documentation to make it more accurate. I plan on rewriting a good portion of this in the near future, not because it's wrong, but just to refresh the material.

:+1:

**Description:**

### DIFF
--- a/doc/observable.md
+++ b/doc/observable.md
@@ -12,13 +12,13 @@ Observables are lazy Push collections of multiple values. They fill the missing 
 ```ts
 import { Observable } from 'rxjs';
 
-const observable = Observable.create(function (observer) {
-  observer.next(1);
-  observer.next(2);
-  observer.next(3);
+const observable = new Observable(subscriber => {
+  subscriber.next(1);
+  subscriber.next(2);
+  subscriber.next(3);
   setTimeout(() => {
-    observer.next(4);
-    observer.complete();
+    subscriber.next(4);
+    subscriber.complete();
   }, 1000);
 });
 ```
@@ -28,21 +28,21 @@ To invoke the Observable and see these values, we need to *subscribe* to it:
 ```ts
 import { Observable } from 'rxjs';
 
-const observable = Observable.create(function (observer) {
-  observer.next(1);
-  observer.next(2);
-  observer.next(3);
+const observable = new Observable(subscriber => {
+  subscriber.next(1);
+  subscriber.next(2);
+  subscriber.next(3);
   setTimeout(() => {
-    observer.next(4);
-    observer.complete();
+    subscriber.next(4);
+    subscriber.complete();
   }, 1000);
 });
 
 console.log('just before subscribe');
 observable.subscribe({
-  next: x => console.log('got value ' + x),
-  error: err => console.error('something wrong occurred: ' + err),
-  complete: () => console.log('done'),
+  next(x) { console.log('got value ' + x); },
+  error(err) { console.error('something wrong occurred: ' + err); },
+  complete() { console.log('done'); },
 });
 console.log('just after subscribe');
 ```
@@ -120,12 +120,12 @@ You can write the same behavior above, but with Observables:
 ```ts
 import { Observable } from 'rxjs';
 
-const foo = Observable.create(function (observer) {
+const foo = new Observable(subscriber => {
   console.log('Hello');
-  observer.next(42);
+  subscriber.next(42);
 });
 
-foo.subscribe(function (x) {
+foo.subscribe(x => {
   console.log(x);
 });
 foo.subscribe(function (y) {
@@ -167,7 +167,7 @@ And this is the same behavior with Observables:
 
 ```js
 console.log('before');
-foo.subscribe(function (x) {
+foo.subscribe(x => {
   console.log(x);
 });
 console.log('after');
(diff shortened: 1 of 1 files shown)


benlesh:
@JWO719 please look over this and make sure it reads fine. I did some find and replace on some of what's there, I might have missed something.

doc/observable.md
```diff
@@ -267,7 +267,7 @@ Conclusion:
 
 ## Anatomy of an Observable
 
-Observables are **created** using `Observable.create` or a creation operator, are **subscribed** to with an Observer, **execute** to deliver `next` / `error` / `complete` notifications to the Observer, and their execution may be **disposed**. These four aspects are all encoded in an Observable instance, but some of these aspects are related to other types, like Observer and Subscription.
+Observables are **created** using `new Observable` or a creation operator, are **subscribed** to with an Observer, **execute** to deliver `next` / `error` / `complete` notifications to the Observer, and their execution may be **disposed**. These four aspects are all encoded in an Observable instance, but some of these aspects are related to other types, like Observer and Subscription.
```
pertrai1: Do you think this might be easier to read if it was in list format?

1. .....
2. .....
3. .....
4. .....

That way it is easier to identify the 4 aspects.

doc/observable.md
```diff
@@ -277,21 +277,21 @@ Core Observable concerns:
 
 ### Creating Observables
 
-`Observable.create` is an alias for the `Observable` constructor, and it takes one argument: the `subscribe` function.
+The `Observable` constructor takes one argument: the `subscribe` function.
 
-The following example creates an Observable to emit the string `'hi'` every second to an Observer.
+The following example creates an Observable to emit the string `'hi'` every second to an subscriber.
```
pertrai1: an subscriber -> a subscriber

doc/observable.md
```diff
@@ -277,21 +277,21 @@ Core Observable concerns:
...
+const observable = new Observable(function subscribe(subscriber) {
   const id = setInterval(() => {
-    observer.next('hi')
+    subscriber.next('hi')
   }, 1000);
 });
 ```
 
-<span class="informal">Observables can be created with `create`, but usually we use the so-called [creation operators](./overview.html#creation-operators), like `of`, `from`, `interval`, etc.</span>
+<span class="informal">Observables can be created with `new Observable`, but usually we use the so-called [creation operators](./overview.html#creation-operators), like `of`, `from`, `interval`, etc.</span>
```
pertrai1: might want to stay away from the use of 'but' simply because it negates anything that is said before it and there is importance in what is said at first

doc/observable.md
```diff
@@ -315,15 +315,15 @@ A `subscribe` call is simply a way to start an "Observable execution" and delive
...
+The code inside `new Observable(function subscribe(subscriber) {...})` represents an "Observable execution", a lazy computation that only happens for each Observer that subscribes. The execution produces multiple values over time, either synchronously or asynchronously.
 
 There are three types of values an Observable Execution can deliver:
 
 - "Next" notification: sends a value such as a Number, a String, an Object, etc.
 - "Error" notification: sends a JavaScript Error or exception.
 - "Complete" notification: does not send a value.
 
-Next notifications are the most important and most common type: they represent actual data being delivered to an Observer. Error and Complete notifications may happen only once during the Observable Execution, and there can only be either one of them.
+Next notifications are the most important and most common type: they represent actual data being delivered to an subscriber. Error and Complete notifications may happen only once during the Observable Execution, and there can only be either one of them.
```
pertrai1: might want to put an emphasis or quotes around "Next" so it does not look like it is a word that is leading to the next piece of information.

doc/observable.md
ajcrites: Should this be "subscribed to with a Subscriber," etc.?

ajcrites:
Do these changes also need to be applied to the Subject docs?

> **Every Subject is an Observer.**

doc/observable.md
maktarsis: What about taking out `subscribe` function declaration on line 287 and put this above as arrow function?

```ts
const observable = new Observable(function subscribe(subscriber) {
  const id = setInterval(() => {
    subscriber.next('hi')
  }, 1000);
 ...
});
```

```ts
const subscribe = (subscriber) => {
  const id = setInterval(() => subscriber.next('hi'), 1000);
 ...
} 
const observable = new Observable(subscribe);
```

maktarsis: It would be more readable if you taking out every callback function in these changes and declare a const as expression of arrow function 

```ts
const subscribe = (subscriber) => ...
const observer = new Observable(subscribe);
```

jsonberry:
Just a comment from the peanut gallery over here... I don't know how much easier it'll make understanding the `new Observable` with renaming `observer` to `subscriber`. RxJS was my first introduction to reactive programming and the Observer pattern. When I went looking for more information about these patterns, Subjects and Observers are what gets discussed, not subscribers.

With a doc refresh, as long as there was a good distinct connection made between `observer` and `subscriber`, I think this would be fine and not cause additional confusion.

... unless I'm totally off and we're renaming all the things ᕕ( ᐛ )ᕗ

benlesh:
@jsonberry ... one problem is that it's not *just* an `Observer`. it's got a property on it called `closed`, and that's tied to the subscription. `Observer` is an interface, which it does implement.

I do see your point, though.

benlesh:
I'm not sure when I'm going to get a chance to finish updating this. Maybe in the next couple of days.

doc/observable.md
benlesh: I'm going to rewrite this whole page soon, so I don't want to get into the minutia of changes right now.

benlesh: Actually no, the subscriber is the "safe" version of the Observer that's been provided by the user, that's also been attached to the subscription. It's only available in the body of the observable function.

benlesh:
I'd like to rewrite this entire document. For now the changes made are just to make it a bit more accurate than it was.

PR #4139, opened 2018-09-15
test(dtslint): add dtslint test on count operator (#4093)

dkosasih:
**Description:** Add count dtslint test

### DIFF
--- a/spec-dtslint/operators/count-spec.ts
+++ b/spec-dtslint/operators/count-spec.ts
@@ -1,0 +1,31 @@
+import { of, Observable } from 'rxjs';
+import { count, buffer } from 'rxjs/operators';
+
+it('should always infer number', () => {
+  const o = of(1, 2, 3).pipe(count(x => x > 1)); // $ExpectType Observable<number>
+  const j = of('a', 'b', 'c').pipe(count(x => x === 'a')); // $ExpectType Observable<number>
+});
+
+it('should accept empty parameter', () => {
+  const o = of(1, 2, 3).pipe(count()); // $ExpectType Observable<number>
+});
+
+it('should infer source observable type in parameter', () => {
+  const o = of(1, 2, 3).pipe(count((x, i, source: Observable<string>) => x === 3)); // $ExpectError
+});
+
+it('should enforce value type of source type', () => {
+  const o = of(1, 2, 3).pipe(count((x, i, source) => x === '3')); // $ExpectError
+});
+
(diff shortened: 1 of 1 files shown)


spec-dtslint/operators/count-spec.ts
```diff
@@ -0,0 +1,14 @@
+import { of } from 'rxjs';
+import { count } from 'rxjs/operators';
+
+it('should always infer number', () => {
+  const o = of(1, 2, 3).pipe(count(x => x > 1)); // $ExpectType Observable<number>
```
timdeschryver: I believe we should also test the predicate overloads.

```ts
predicate: (value: T, index: number, source: Observable<T>)
```

spec-dtslint/operators/count-spec.ts
```diff
@@ -0,0 +1,14 @@
...
+});
+
+it('should accept empty parameter', () => {
+  const o = of(1, 2, 3).pipe(count()); // $ExpectType Observable<number>
+});
+
+it('should expect function parameter', () => {
+  const o = of(1, 2, 3).pipe(count(9)); // $ExpectError
```
timdeschryver: We should also test if the value is the same type as the source.
For example 

```ts
const o = of(1, 2, 3).pipe(count((value: string) => value !== ''); // $ExpectError
```

spec-dtslint/operators/count-spec.ts
dkosasih: Added few more tests around the predicate overloads types. 
Please let me know if they are of a good value. 
Thanks

spec-dtslint/operators/count-spec.ts
```diff
@@ -0,0 +1,26 @@
+import { of } from 'rxjs';
+import { count, buffer } from 'rxjs/operators';
+
+it('should always infer number', () => {
+  const o = of(1, 2, 3).pipe(count(x => x > 1)); // $ExpectType Observable<number>
```
cartant: With `count`, the return type of the piped observable will always be `Observable<number>`, so I think that for the tests, the source observable type should be something other than `number` - to make sure that the output type comes from the `count` operator and not the source. E.g.:

```ts
const o = of('1', '2', '3').pipe(count(x => x !== '1')); // $ExpectType Observable<number>
```

dkosasih: @cartant I have added another row to both number or string will always return number

PR #4147, opened 2018-09-18
test(dtslint): add dtslint test on delay operator (#4093)

dkosasih:
**Description:** add dtsLint test on delay operator

### DIFF
--- a/spec-dtslint/operators/delay-spec.ts
+++ b/spec-dtslint/operators/delay-spec.ts
@@ -1,0 +1,20 @@
+import { of, asyncScheduler } from 'rxjs';
+import { delay } from 'rxjs/operators';
+
+it('should infer correctly', () => {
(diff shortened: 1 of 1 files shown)


spec-dtslint/operators/delay-spec.ts
```diff
@@ -0,0 +1,20 @@
+import { of, asyncScheduler } from 'rxjs';
+import { delay } from 'rxjs/operators';
+
+it('should infer correctly', () => {
+  const o = of(1, 2, 3).pipe(delay(100)); // $ExpectType Observable<number>
+});
+
+it('should support date parameter', () => {
+  const o = of(1, 2, 3).pipe(delay(Date.now())); // $ExpectType Observable<number>
```
timdeschryver: [`Date.now()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now) returns a number, it would be better to do `new Date(2018, 08, 18)`.

PR #4149, opened 2018-09-18
docs(accessibility): minor tweaks to improve accessibility

timdeschryver:
**Description:**
I reviewed the accessibility of the new website with aXe and pa11y.
These changes addresses some of the flaws.

I also change the notification style because it had the Angular styles applied (and also triggered the accessibility color contrast warning).

**Related issue (if exists):** /

### DIFF
--- a/docs_app/src/index.html
+++ b/docs_app/src/index.html
@@ -141,7 +141,7 @@
     <div class="background-sky hero"></div>
     <section id="intro">
       <div class="hero-logo">
-        <img src="assets/images/logos/angular/angular.svg" width="250" height="250" alt="RxJS logo">
+        <img src="assets/images/logos/angular/angular.svg" width="250" height="250" alt="Angular logo">
       </div>
       <div class="homepage-container">
         <div class="hero-headline">One framework.<br>Mobile &amp; desktop.</div>
--- a/docs_app/src/index.html
+++ b/docs_app/src/index.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html>
+<html lang="en">
 <head>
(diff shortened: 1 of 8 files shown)


docs_app/src/index.html
```diff
@@ -141,7 +141,7 @@
     <div class="background-sky hero"></div>
     <section id="intro">
       <div class="hero-logo">
-        <img src="assets/images/logos/angular/angular.svg" width="250" height="250">
+        <img src="assets/images/logos/angular/angular.svg" width="250" height="250" alt="RxJS logo">
```
JWO719: This seems to be angular related, none of your business, but a thing we should take care of at some point

timdeschryver: 🤦‍♂️ didn't notice it was the angular logo. I'll change it to `Angular logo`, so if we do address this at some point we don't forget to change the alt attribute.

JWO719: Sorry but you slightly misunderstood me. This part seems to be old stuff of the aio web page we used as basis for the new rxjs docs therefore we should create new content for rxjs in case js is disabled in a browser

timdeschryver: Yea, this is indeed what I have understood.
I changed the `alt` attribute to not forget it when we change the `noscript` part because I thought we wouldn't want to change this content at this point.

But perhaps you prefer to remove the `alt` attribute? Or to already change the logo to the RxJS logo?

JWO719: Ah ok get it:D I would suggest leave it as it is right now, and we think about a proper noscript content later on!

PR #4151, opened 2018-09-19
Operator guide for new docs.

ajcrites:
**Description:**
This includes updates to the operators guide for consistency with the latest documentation and RxJS v6.

I removed the decision tree widget which may be good to add later.


docs_app/content/navigation.json
```diff
@@ -26,17 +26,21 @@
 			"title": "Overview",
 			"tooltip": "RxJS Overview",
 			"children": [
-                {
+								{
```
JWO719: this is wrongly indented

docs_app/content/navigation.json
```diff
@@ -26,17 +26,21 @@
...
+				{
 					"url": "guide/subscription",
 					"title": "Subscription"
 				},
-                {
+				{
 					"url": "guide/subject",
 					"title": "Subjects"
+				},
+				{
```
JWO719: could you maybe change the order. I think I would put it below observables.

docs_app/src/assets/images/examples/marble-diagram-anatomy.svg
```diff
@@ -0,0 +1,140 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
JWO719: could you put this in a folder `guide` instead of examples. Then it's clear that it belongs together

docs_app/content/guide/operators.md
```diff
@@ -0,0 +1,219 @@
...
+```none
+10
+20
+30
+40
+```
+
+Notice that a subscribe to `output` will cause the `input` Observable to be subscribed. We call this an "operator subscription chain".
+
+You can also create custom operators by wrapping calls to existing operators. We can reimplement our `multiplyByTen` operator by wrapping the `map` operators which will project our multiplicatoin function.
```
JWO719: here's a typo 'multiplicatoin' it should be 'multiplication'

docs_app/content/guide/operators.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+## Creation Operators
+
+**What are creation operators?** Besides pipeable operators, creation operators are functions that can be used to create an Observable with some common predefined behavior or by combining other Observables.
+
+A typical example of a creation operator would be the `interval` function. It takes a number (not an Observable) as input argument, and produces an Observable as output:
+
+```js
+const observable = interval(1000 /* number of milliseconds */);
```
JWO719: could you add the import for interval here

docs_app/content/guide/operators.md
```diff
@@ -0,0 +1,219 @@
...
+
+Below you can see the anatomy of a marble diagram.
+
+<img src="assets/images/examples/marble-diagram-anatomy.svg">
+
+Throughout this documentation site, we extensively use marble diagrams to explain how operators work. They may be really useful in other contexts too, like on a whiteboard or even in our unit tests (as ASCII diagrams).
+
+## Categories of operators
+
+There are operators for different purposes, and they may be categorized as: creation, transformation, filtering, combination, multicasting, error handling, utility, etc. In the following list you will find all the operators organized in categories.
```
JWO719: I really like this overview, but I fear that we could forget updating this. Therefore I would recommend adding a sentence like 'For a complete api overview, see references page' something like that

docs_app/content/guide/operators.md
pertrai1: should the 'operators' after `map` be singular?

docs_app/content/navigation.json
ajcrites: I'm happy to change it, but I was trying to retain the order of the old docs http://reactivex.io/rxjs/manual/overview.html

ajcrites:
I had also updated the Subject page earlier, so I have a question --

Do we want to mirror these changes in the old `doc`, or do we want to update the old docs to link to the new ones as with https://github.com/ReactiveX/rxjs/blob/master/MIGRATION.md? Or just leave the old doc alone.

JWO719:
@ajcrites Do you need any help with the requested changes? I would really like to see this in the new docs. When there is anything I can help you with, just poke me

ajcrites:
@JWO719 I have made all of the requested changes (to my knowledge), but I didn't put a comment about that because I thought it would be noise. In the future, I'll make sure to leave a comment. If you have any further comments, I'll be able to address them today. I went through your feedback, and I attempted to follow through on Ben's feedback with 98883363e35cfd1f3d2ea9d4fb19e7e203b825c3 (although I think some of the distinctions he brought up were already in the doc though maybe not as clear as they could be). Hope you enjoy!

ajcrites:
I've resolved conflicts with the navigation menu, so hopefully this is ready to merge.

ajcrites:
Checking up on this. Anything that I can do to help get this merged?

docs_app/content/guide/operators.md
```diff
@@ -0,0 +1,237 @@
...
+
+const observable = interval(1000 /* number of milliseconds */);
+```
+
+See the list of [all static creation operators here](#creation-operators).
+
+Some *Join Operators* such as `merge`, `combineLatest`, `concat`, etc. take *multiple* Observables as input, not just one, for instance:
+
+```js
+const observable1 = interval(1000);
```
JWO719: could you add the imports here

ajcrites:
I have made the requested changes. Please let me know if other changes are needed.

PR #4160, opened 2018-09-20
test(dtslint): add tap operators dtslint test (#4093)

dkosasih:
**Description:** Add dtslint test for tap operators

### DIFF
--- a/spec-dtslint/operators/tap-spec.ts
+++ b/spec-dtslint/operators/tap-spec.ts
@@ -1,0 +1,24 @@
+import { of } from 'rxjs';
+import { tap } from 'rxjs/operators';
+
+it('should infer correctly', () => {
+  const a = of(1, 2, 3).pipe(tap()); // $ExpectType Observable<number>
+});
+
+it('should accept partial observer', () => {
+  const a = of(1, 2, 3).pipe(tap({ next: (x: number) => { } })); // $ExpectType Observable<number>
+  const b = of(1, 2, 3).pipe(tap({ error: (x: any) => { } })); // $ExpectType Observable<number>
+  const c = of(1, 2, 3).pipe(tap({ complete: () => { } })); // $ExpectType Observable<number>
+});
+
+it('should not accept empty observer', () => {
+  const a = of(1, 2, 3).pipe(tap({})); // $ExpectError
+});
+
+it('should enforce type for next observer function', () => {
+  const a = of(1, 2, 3).pipe(tap({ next: (x: string) => { })); // $ExpectError
+});
+
+it('should accept any type for error observer function', () => {
+  const a = of(1, 2, 3).pipe(tap({ error: (x: any) => { } })); // $ExpectType Observable<number>
+});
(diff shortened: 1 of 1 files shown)


spec-dtslint/operators/tap-spec.ts
```diff
@@ -0,0 +1,20 @@
+import { of } from 'rxjs';
+import { tap } from 'rxjs/operators';
+
+it('should infer correctly', () => {
+  const a = of(1, 2, 3).pipe(tap()); // $ExpectType Observable<number>
+});
+
+it('should accept partial observer', () => {
+  const a = of(1, 2, 3).pipe(tap({ next: (x: number) => x })); // $ExpectType Observable<number>
+  const b = of(1, 2, 3).pipe(tap({ error: (x: number) => x })); // $ExpectType Observable<number>
```
cartant: The `error` callbacks should be typed as `any`. There is no guarantee that the callback will be passed a particular type. Also, the error callback should not return a value. Use `error: (err: any) => {}`.

spec-dtslint/operators/tap-spec.ts
```diff
@@ -0,0 +1,20 @@
...
+  const c = of(1, 2, 3).pipe(tap({ complete: () => 1 })); // $ExpectType Observable<number>
+});
+
+it('should enforce type for next observer function', () => {
+  const a = of(1, 2, 3).pipe(tap({ next: (x: string) => x })); // $ExpectError
+});
+
+it('should accept any type for error observer function', () => {
+  const a = of(1, 2, 3).pipe(tap({ error: (x: string) => x })); // $ExpectType Observable<number>
```
cartant: Use `error: (err: any) => {}`. The error callback should accept `any` and should not return. TypeScript is okay with it, but the test should reflect the intended usage.

spec-dtslint/operators/tap-spec.ts
```diff
@@ -0,0 +1,20 @@
+import { of } from 'rxjs';
+import { tap } from 'rxjs/operators';
+
+it('should infer correctly', () => {
+  const a = of(1, 2, 3).pipe(tap()); // $ExpectType Observable<number>
+});
+
+it('should accept partial observer', () => {
+  const a = of(1, 2, 3).pipe(tap({ next: (x: number) => x })); // $ExpectType Observable<number>
```
cartant: Don't return a value from the `next` handler. TypeScript is okay with this, but the test should reflect the intended usage.

dkosasih:
@cartant 
I did the changes. Could you have a look, please?

spec-dtslint/operators/tap-spec.ts
```diff
@@ -0,0 +1,24 @@
...
+
+it('should not accept empty observer', () => {
+  const a = of(1, 2, 3).pipe(tap({})); // $ExpectError
+});
+
+it('should enforce type for next observer function', () => {
+  const a = of(1, 2, 3).pipe(tap({ next: (x: string) => { })); // $ExpectError
+});
+
+it('should accept any type for error observer function', () => {
```
cartant: This test can be removed, as it's the same as the one in the `should accept partial observer` test.

PR #4163, opened 2018-09-20
[DONOTMERGE] build(package): project references proposal

kwonoj:
**Description:**

We are aware of our build script is quite verbose due to multiple typescript configurations for several targets. TS 3.0 have new features (https://www.typescriptlang.org/docs/handbook/project-references.html) which we could leverage to reduce some of those steps.

In short, it allows to `reference` between multiple configurations, like `compat` and our `src` or something else, and one build (`tsc -b ...projects`) will chain those projects and build necessary changed files only. What we currently do is manually rebuild each project, which compiler doesn't aware changes between if whole rebuild retrigger is required or not.

I tried small poc (`src` + `compat` for cjs only) as poc script in `npm run build` here, for anyone give it a try and see how does it feels. Still it's not even close to replace existing build, there are number of steps need to figure out if we'd like to move in this way:

- [ ] setting up references chain correct to enable all builds successfully
- [ ] setting up output configuration correctly to replicate current output does
- [ ] experiments multiple target config (same rootdir i.e `src` for cjs / esm etcs) works with project references, or should it be separate script as same as currently
- vice versa

I feel this is somewhat better improvement than what we currently have, would like to push forward if we got agreement to go. PR raised before actually dig down deeper.


kwonoj:
I'm going to create new PR for actual effort.

PR #4175, opened 2018-09-24
test(dtslint): add dtslint test for endWith operator (#4093)

dkosasih:
**Description:** add dtslint test for endWith operator


dkosasih:
Added a couple of tests to cover you scenario. Empty parameter is accepted.

cartant:
This looks good to me, but it's highlighted that `endWith` needs to be brought into like with `startWith` in that it should be possible to pass values of any type and the type should be widened.

I think you made those changes to `startWith`, so would you like to change `endWith`, too?

dkosasih:
@cartant it wasn't me, I think @timdeschryver did. 
I can create a new issue and you can assign it to me. sounds OK?

Issue #4176, opened 2018-09-24
Enable declarationMap in typings


kwonoj:
## Feature Request

**Additional context**
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html

> Enabling --declarationMap alongside --declaration causes the compiler to emit .d.ts.map files alongside the output .d.ts files. Language Services can also now understand these map files, and uses them to map declaration-file based definition locations to their original source, when available.

> In other words, hitting go-to-definition on a declaration from a .d.ts file generated with --declarationMap will take you to the source file (.ts) location where that declaration was defined, and not to the .d.ts.

Since our package includes source anyway, enabling declaration map will allow more easier navigation experiences.

kwonoj:
what we need to resolve:

 our package layout places src other than resolvable in package: try this for check

```
npm run build_all
cd dist/packages/typings
open one of .map
```

check src reference,  👉 is part need to be resolved cause our `packages` flattens our src into different location.

felixfbecker:
Would love to see this because it enables cross-repository jump-to-definition on sourcegraph.com

benlesh:
This no longer seems to be an issue. Closing for now.

PR #4179, opened 2018-09-24
chore(build): Enable declarationMap in typings (#4176)

rajmayank:
**Description:**
Since our package includes source anyway, enabling declaration map will allow more easier navigation experiences.

#4176

### DIFF
--- a/tsconfig/compat/tsconfig.cjs.json
+++ b/tsconfig/compat/tsconfig.cjs.json
@@ -6,6 +6,7 @@
     "removeComments": false,
     "declaration": true,
     "declarationDir": "../../dist-compat/typings",
+    "declarationMap": true,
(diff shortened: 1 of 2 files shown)


kwonoj:
Thanks, but unfortunately this is not suffecient as our package layout places src other than resolvable in package: try this for check

```
npm run build_all
cd dist/packages/typings
open one of .map
```

check src reference,  👉 is part need to be resolved cause our `packages` flattens our src into different location.

benlesh:
I'm with @kwonoj on this one, sorry, @rajmayank. I really appreciate the thought and effort here, but I'm afraid I'll need to close this PR. 

Thank you, and I hope this doesn't dissuade you from contributing again in the future!

PR #4182, opened 2018-09-25
fix(Subscriber): use only local Subscriber instances

benlesh:
This removes any checks for `rxSubscriber` symbol or . That was a micro-optimization that proved to be problematic for node projects with more than one instance of RxJS.

This means that all local `Subject` instances will be wrapped by
`SafeSubscriber`, but prior to that, they were wrapped in
`SubjectSubscriber` anyhow, so the difference is minimal.

Leaves behind `SubjectSubscriber` and implementations of `rxSubscriber`,
which are still functioning, but now cruft.


benlesh:
> Instanceof check doesn't work across different version instances, right? Mean this change will makes changing across versions? (Or was it already not working?)

That's 100% correct.

Basically, the problem is we were trying to be overly clever here. This is the reason that over the last couple of years on at least two occasions (that I can remember) we had issues where having multiple instances of RxJS in a node app would throw runtime errors.

In a nutshell, we were trying to reuse Subscribers from "trusted" sources rather than wrapping them in a `SafeSubscriber`. Fact is, it just wasn't worth the small optimization for that edge case.

benlesh:
@cartant that's exactly right, although I'm not sure that some of the type narrowing there made sense.

Unfortunately, the type signatures I moved I had to to make TS@next happy. So that's a bridge we're going to have to cross regardless. I'm experimenting with better solutions.

benlesh:
The following typing was incorrect

```ts
export function pipe<T>(...operators: UnaryFunction<T, T>[]): <R extends T>(source: R) => R
```

Specifically the return type. Consider this:

```ts
const fn = pipe((x: string) => x, (x: string) => x + x);

const r = fn('poo'); // The type of r here is "foo", where the value is actually "poopoo"
```

The last commit here should fix that.

cartant:
The value might be `'poopoo'`, but the type should be `string`, so I'm not sure how that's a problem. I can look at it in more detail later; on the train ATM.

BTW, the reasoning behind the signatures is here: https://medium.com/p/81146fbb14b6

benlesh:
@cartant

[This outlines the problem a little more concisely, I think](https://www.typescriptlang.org/play/#src=%0D%0Afunction%20pipe%3CA%3E(...fns%3A%20Array%3C(input%3A%20A)%20%3D%3E%20A%3E)%3A%20%3CR%20extends%20A%3E(input%3A%20R)%20%3D%3E%20R%20%7B%0D%0A%20%20return%20null!%3B%0D%0A%7D%0D%0A%0D%0Aconst%20fn%20%3D%20pipe((x%3A%20string)%20%3D%3E%20x)%3B%0D%0A%0D%0Aconst%20x%20%3D%20fn('foo')%3B%0D%0A%0D%0A%0D%0A%2F%2F%20Here%20is%20the%20problem.%0D%0Aconst%20fn2%20%3D%20pipe((x%3A%20string)%20%3D%3E%20x%20%2B%20x)%3B%0D%0A%0D%0A%2F%2F%20x2's%20value%20is%20%22foofoo%22%2C%20but%20it's%20type%3F%20%22foo%22.%20%3A(%0D%0Aconst%20x2%20%3D%20fn('foo')%3B%0D%0A%0D%0A%2F*%0D%0A%20%20This%20is%20because%20the%20result%20type%20of%0D%0A%20%20pipe%20above%20%60(input%3A%20R)%20%3D%3E%20R%60%20is%20really%20just%20tricking%0D%0A%20%20TS%20into%20narrowing%20the%20type%20to%20whatever%20the%20input%20was%0D%0A*%2F%0D%0A)

cartant:
I see. In `next` they've started narrowing the types wherever possible. I'll give it some thought. If the signatures cannot be changed - I think I might be able to solve it - they might as well be removed.

Right now, you might want to just comment them out - along with the dtslint tests - to get the CI builds working. I can look at this later.

Issue #4183, opened 2018-09-25
endWith should widen the type if necessary


cartant:
## Feature Request

`startWith` can be passed values other than `T` and if that's done, the operator's return type is widened. See #4034.

`endWith` needs to be treated similarly and needs an additional dtslint test, etc.

timdeschryver:
Feel free to ping me if you need to verify something @dkosasih

cartant:
A PR for this has been merged.

PR #4185, opened 2018-09-25
fix(endWith): ability to endWith different types (#4183)

dkosasih:
**Description:** add ability for endWith operator to end the Observable with different types.

**Related issue (if exists):** #4183


benlesh:
Thanks, @dkosasih!

PR #4186, opened 2018-09-25
chore(dtslint): comment helper functions

cartant:
**Description:**

This PR adds comments explaining the purpose of the `a` helper functions that are used in the dtslint tests for `Observable.prototype.pipe` and the static `pipe` function.

Without the helper functions the tests would involve much more code, but what the functions actually do isn't immediately obvious, as they depend upon the inference and compatibility/incompatibility of string literal types.


spec-dtslint/Observable-spec.ts
```diff
@@ -1,6 +1,22 @@
 import { Observable, of, OperatorFunction } from 'rxjs';
 import { mapTo } from 'rxjs/operators';
 
+// The `a` (for argument) function is used to keep the tests uncluttered.
```
benlesh: If you do a `/** Info here */` style doc-comment, IDEs like VS Code will pick them up. :)

cartant: Done.

benlesh:
@cartant looks like Travis is failing on this one because of a bad type in the comments.

cartant:
@benlesh This should be fixed. I've disabled some picky TSLint rules that defaulted to being enabled in the dtslint configuration.

PR #4191, opened 2018-09-26
docs(marbles): Fix for Marble Diagrams

pertrai1:
@JWO719 @benlesh - Not sure if this is completely correct, but it does output the diagrams as you will see with the buffer.ts test if you were to run `npm run tests2png`

buffer-diagram

**Description:**


benlesh:
thanks, @pertrai1!

Issue #4159, opened 2018-09-19
Refine subscribe and tap signatures


benlesh:
Currently we have a variety of ways you can use `subscribe` and tap:

```ts
o.subscribe(fn, fn, fn);
o.subscribe(fn, fn);
o.subscribe(fn);
o.subscribe(void, void, fn);
o.subscribe(void, fn, fn);
o.subscribe(void, fn);
o.subscribe(observer);


o.pipe(tap(fn, fn, fn));
o.pipe(tap(fn, fn));
o.pipe(tap(fn));
o.pipe(tap(void, void, fn));
o.pipe(tap(void, fn, fn));
o.pipe(tap(void, fn));
o.pipe(tap(observer));
```

It's been my personnel recommendation that people stick to the following two signatures:

```ts
o.subscribe(fn);
o.subscribe(observer);

// or likewise with tap:
o.pipe(tap(fn));
o.pipe(tap(observer));
```

The reason being is:

```ts
// POOR readability (I have to count arguments?)
o.subscribe(null, null, () => {
  // side effect here
});

// BETTER readability
o.subscribe({
  complete() {
    // side effect here
  }
});

// GOOD readability
o.subscribe(x => {
  // side effect here
});

// GOOD readability (if unnecessarily verbose)
o.subscribe({
   next(x) { 
      // side effect here
   }
});
```

### Proposal

I'd like to deprecate the other signatures to `subscribe` and `tap`, and narrow down the signature to just those two prescribed above for the next major version

**Pros**

- Simplify subscription code slightly.
- Simpler typing for TypeScript.
- Guides people into writing cleaner, more readable code

**Cons**

- The behavior around handling errors at the point of subscription may get a slightly harder to control, with only one way of adding error handling at subscription time. (NOTE: errors should really be handled by a `catchError` or `retry`-type operator, rather than in the subscribe.)

### Other thoughts

It would also be good to add some lint rules around this to rxjs-tslint and the like.

benlesh:
> but do not allow any other forms? or it is still allowed

It would be *allowed* in the short term, but in the long term (next major version or some future major version), it would be a hard API change.

benlesh:
And if anyone is wondering "why not just `subscribe(observer)` and `forEach(fn)`?"... Well, `forEach` is *supposed* to have the semantic of nexting on a microtask, *and* `forEach` returns a promise and is currently non-cancellable.

I'd expect that `.subscribe(fn)` is probably the 90% use case for people.

[...]

PR #4202, opened 2018-09-28
fix(subscribe): Deprecate null starting parameter signatures for subscribe

ajcrites:
**Description:**
Per discussion on #4159, this deprecates the signatures for `.subscribe` and `.tap` that are `(null, null, complete)`, `(null, error, complete)`, and `(next, null, complete)`.

#4159

### DIFF
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -70,12 +70,12 @@ export class Observable<T> implements Subscribable<T> {
   }
 
   subscribe(observer?: PartialObserver<T>): Subscription;
-  /** @deprecated */
-  subscribe(next: null, error: null, complete: () => void): Subscription;
-  /** @deprecated */
-  subscribe(next: null, error: (error: any) => void, complete?: () => void): Subscription;
-  /** @deprecated */
-  subscribe(next: (value: T) => void, error: null, complete: () => void): Subscription;
+  /** @deprecated Use an observer instead of a complete callback */
+  subscribe(next: null | undefined, error: null | undefined, complete: () => void): Subscription;
+  /** @deprecated Use an observer instead of an error callback */
+  subscribe(next: null | undefined, error: (error: any) => void, complete?: () => void): Subscription;
+  /** @deprecated Use an observer instead of a complete callback */
+  subscribe(next: (value: T) => void, error: null | undefined, complete: () => void): Subscription;
   subscribe(next?: (value: T) => void, error?: (error: any) => void, complete?: () => void): Subscription;
   /**
    * Invokes an execution of an Observable and registers Observer handlers for notifications it will emit.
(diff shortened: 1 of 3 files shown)


src/internal/Observable.ts
```diff
@@ -70,6 +70,12 @@ export class Observable<T> implements Subscribable<T> {
   }
 
   subscribe(observer?: PartialObserver<T>): Subscription;
+  /** @deprecated */
+  subscribe(next: null, error: null, complete: () => void): Subscription;
```
cartant: I think `undefined` would be the way to go. At the moment, it's not possible to call `subscribe` and pass `null` as an argument without TypeScript complaining, so unless the existing signatures are altered to accept `null` arguments, I the deprecations will be unnecessary.

cartant: There'd be no harm in using `null | undefined`, though.

cartant:
I'm actually thinking about that right now, as I'm writing a TSLint rule to do the same. I'll let you know what I come up with.

cartant:
Regarding the coveralls. It seems to +/-0.2% changes in a seemingly non-deterministic way. I was speaking to someone on Monday, at a meetup, and he was of the opinion that it was due to the PR's branch not being completely up-to-date with `master`.

As an experiment, could you ensure your `master` is up-to-date and then rebase your PR and push it again? Like this:

```
git checkout master
git fetch upstream
git rebase upstream/master
git checkout deprecate-subscribe-fn
git rebase master
git push -f
```

I'd be really interested in whether or not coveralls passes after that. This +/-0.2% variation has been an annoyance for a long time.

kwonoj:
regarding above, I'm curious if generated coverage have problem or the way coveralls translate it have some problems. ..brutal way, maybe we can setup codecov also and 1:1 compare for a while per each PRs?

cartant:
@kwonoj I don't really know enough about Coveralls or what it's doing. If we can establish what the problem is, we might be able to get a decent solution.

ATM, if you cannot reliably see whether the coverage goes up or down, it's not especially useful.

TBH, though, I'm not super keen on getting into the coveralls nitty gritty. :grimacing:

kwonoj:
I am usually just ignoring those minor deets, so on same boat. Only reason I care is most new contributor got highly confused like 'I only changed doc and coverage dropped, maybe I did something wrong'.

cartant:
@ajcrites I'd suggest something like:

```ts
/** @deprecated Use an observer instead of an error callback */
```

ajcrites:
We're in the midst of updating the terminology "observer" to "subscriber," at least in some documentation. Should I stick with "observer" for this?

cartant:
Sorry about the closure. On mobile with fat fingers.

My understanding is that the documentation is changing re: `new Observable(subscriber => {})` - where the parameter really is a `Subscriber` - and that's not the case here. The argument is not a subscriber. I could be wrong, but that's my take. In fact, the type that corresponds to the argument, here, is actually `PartialObserver`, as not all methods are required.

benlesh:
@ajcrites ... subscribe accepts a `PartialObserver`, which is any subset of `Observer`... `Subscriber` is a superset of `Observer` in that it has a `closed` property on it (as well as other things in RxJS 5 and 6).

PR #4201, opened 2018-09-28
build(docs_app): upgrade to latest SW and other fixes

gkalpak:
**Description:**
Includes the following changes to the docs app:

1. Remove unused files/scripts (that were unnecessarily copied over from angular.io).

2. Upgrade `@angular/service-worker` to 6.1.9.
    This essentially includes the (relevant) commits from angular/angular#25997 (plus a couple minor ones).
    The current `ngsw-config.json` seems reasonable, but you might want to tweak it.

3. Remove unused images (mainly to avoid fetching them from the SW and unnecessarily consuming bandwidth).

4. Replace Angular images and headline with RxJS ones:
    **a.** In the 404 page:
        _Before:_
        rxjs-404-before
        _After:_
        rxjs-404-after
        **NOTE:**
        I hacked together an RxJS sad logo (similar to the Angular one), but you probably want to provide your own.
        (Still, it is better than showing the Angular logo :grin:)
    **b.** In the "no JavaScript" page (shown when JavaScript is disabled):
        _Before:_
        rxjs-nojs-before
        _After:_
        rxjs-nojs-after

**Related issue (if exists):**
Fixes #4178.

PR #4204, opened 2018-09-28
docs(docs_app): fix commands in `README.md`

gkalpak:
In angular.io (where the README was originally taken from), we use `yarn`, so the commands need minor modification when translating to `npm`.
(This is a follow-up to #4201.)


benlesh:
Thanks, @gkalpak!

PR #4207, opened 2018-09-29
Ignore this PR; it's just a Coveralls experiment

cartant:
See PR title.

### DIFF
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -70,6 +70,12 @@ export class Observable<T> implements Subscribable<T> {
   }
 
   subscribe(observer?: PartialObserver<T>): Subscription;
+  /** @deprecated */
+  subscribe(next: null, error: null, complete: () => void): Subscription;
+  /** @deprecated */
+  subscribe(next: null, error: (error: any) => void, complete?: () => void): Subscription;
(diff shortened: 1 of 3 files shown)


cartant:
@kwonoj I grabbed Andrew's branch and created this PR. After rebasing it, Coveralls is back in sync with `master`. When rebasing, there were two commits added to this branch from `master`. I'm inclined to believe the guy I was speaking with at the meetup. The -0.2% difference comes from the small amount of code that was in `master` that was not in the PR branch. What do you think?

kwonoj:
> the -0.2% difference comes from the small amount of code that was in master that was not in the PR branch.

Welp. I think we can't enforce user to rebase into master all times. 🤔 

OK, I did this: 

image

Set failure minimum to 96 (which should be sufficient cause our's current about 96.8) and ignore -0.2% as allowed threshold. Guess.. this'll work? at least it won't mark as checker fail.

cartant:
I can see myself ending up looking at the Coveralls documentation. :grimacing:

What would be nice would be for the PR's coverage to be compared against the coverage of the commit in the PR's branch that precedes the PR's first commit - rather than being compared against master.

Whether or not that's possible, I dunno.

PR #4208, opened 2018-09-29
feat(Operators): add minStream and maxStream

PxyUp:
Good day. Two operators were added to obtain the minimum value and the maximum for each operation.


kwonoj:
Appreciate for taking time to implement, but our general recommendation is if implementation is trivial / and doesn't require modification in core, we do recommend for user-land first. We want to gather usecases around how much common it is before we include into core, since adding new api surface is easy but deprecating / changing behavior is exceptionally hard, as well as we don't want to core api surface it too broad (it is alreay, and we are trying to shrink it down).

Looking implementation, I'd like to recommend to publish as userlevel module first.

PxyUp:
Thank you, will do it

PxyUp:
@kwonoj 
I thought and want to offer you an idea. If you enter the operator minStream / maxStream, then this makes the stream more universal. Then the operator min / max turn into a combination of 
```javascript
.pipe(
    minStream(), // maxStream
    last()
)
```
This solution will increase performance because earlier in the `min / max` operators the `reduce` was used, and each time it counted everything, now you can only compare the last value.

```javascript
var rxjsMinMax = require('./dist/index.js');
var rxjs = require('rxjs');
var last = require('rxjs/operators').last;
var min = require('rxjs/operators').min;

const n = 100000;
const array = Array.from({ length: n }, () => Math.floor(Math.random() * n));

console.time('minStream');
rxjs.from(array).pipe(rxjsMinMax.minStream(), last()).subscribe((x) => console.log(x));
console.timeEnd('minStream');

console.time('min');
rxjs.from(array).pipe(min()).subscribe((x) => console.log(x));
console.timeEnd('min');

// 0
// minStream: 11.276ms
// 0
// min: 40.306ms
```
When the number of values> 10,000 minutes begins to lose in performance

PR #4209, opened 2018-10-01
test(dtslint): add dtslint test for min operator (#4093)

dkosasih:
**Description:** add dtslint test for Min operator

### DIFF
--- a/spec-dtslint/operators/min-spec.ts
+++ b/spec-dtslint/operators/min-spec.ts
@@ -1,0 +1,17 @@
+import { of } from 'rxjs';
+import { max } from 'rxjs/operators';
+
+it('should infer correctly', () => {
+    const a = of(1, 2, 3).pipe(max()); // $ExpectType Observable<number>
+    const b = of('abc', 'bcd', 'def').pipe(max()); // $ExpectType Observable<string>
+});
+
+it(' should except empty comparer', () => {
+    const a = of(1, 2, 3).pipe(max()); // $ExpectType Observable<number>
+});
+
+it('should infer on comparer types', () => {
+    const a = of(1, 2, 3).pipe(max((a: number, b: number) => a <= b ? -1 : 1)); // $ExpectType Observable<number>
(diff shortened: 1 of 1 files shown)


spec-dtslint/operators/min-spec.ts
```diff
@@ -0,0 +1,17 @@
...
+it('should infer correctly', () => {
+    const a = of(1, 2, 3).pipe(max()); // $ExpectType Observable<number>
+    const b = of('abc', 'bcd', 'def').pipe(max()); // $ExpectType Observable<string>
+});
+
+it(' should except empty comparer', () => {
+    const a = of(1, 2, 3).pipe(max()); // $ExpectType Observable<number>
+});
+
+it('should infer on comparer types', () => {
```
cartant: This isn't so much inferring as it is enforcing, so I'd have used `'should enforce comparer types'`.

spec-dtslint/operators/min-spec.ts
```diff
@@ -0,0 +1,17 @@
...
+    const a = of(1, 2, 3).pipe(max((a: number, b: number) => a <= b ? -1 : 1)); // $ExpectType Observable<number>
```
cartant: You can write numeric comparers like this: `(a: number, b: number) => a - b`

spec-dtslint/operators/min-spec.ts
```diff
@@ -0,0 +1,17 @@
...
+    const b = of(1, 2, 3).pipe(max((a: number, b: string) => a <= b ? -1 : 1)); // $ExpectError
```
cartant: You cannot subtract a `string` from a `number`, so just return `0` to be sure the error is effected from the parameters and not the subtraction: `(a: number, b: string) => 0`

The comparer implementation doesn't have to make sense; it just shouldn't effect an error of its own.

dkosasih:
@cartant I have made the changes advised. That also applied to Max operator.
Thank you.

PR #4211, opened 2018-10-01
fix(build): disable tsetse rule ban-expect-truthy-promise

gregmagolan:
**Description:**

Fixes Bazel build of rxjs with latest rules_typescript


kwonoj:
Would you elaborate bit more details around changes?

Issue #4131, opened 2018-09-11
Type inference for pluck operator


michaeljota:
## Feature Request

**Is your feature request related to a problem? Please describe.**
Currently pluck operator does not have type inference in the returning object.

**Describe the solution you'd like**
Actually, this was discussed before in #2295. I think is the only way so far to do this properly.

**Describe alternatives you've considered**
Maybe no too many overloads (?)

**Additional context**
Back then the original OP was closed because this project wasn't yet using TS 2.1. But now that it uses TS3.0 it would be a great feature to have.

kwonoj:
Yeah, I think we can try with new compiler features + dtslint verification.

michaeljota:
Looking at the source code, I think this could be done just with regular overloads, but I don't know if you feel good about it.

cartant:
I think you'll have to use overloads - like [this](https://github.com/cartant/rxjs-etc/blob/master/source/operators/pluck.ts) - as the current release has to work with 2.8. And I'm not sure that the tuple parameter extraction in 3.0 will be of any use, anyway, as the parameters chain - with each depending upon the previous, etc.

cartant:
Closing as #4232 has been merged.

PR #4215, opened 2018-10-03
test(dtslint): add dtslint test for pluck operator(#4093)

dkosasih:
**Description:** Add dtslint test for Pluck operator

### DIFF
--- a/spec-dtslint/operators/pluck-spec.ts
+++ b/spec-dtslint/operators/pluck-spec.ts
@@ -1,0 +1,18 @@
+import { of } from 'rxjs';
+import { pluck } from 'rxjs/operators';
+
+it('should infer empty interface when not specified', () => {
+  const a = of([{ name: 'abc', id: 123 }, { name: 'def', id: 256 }]).pipe(pluck('name')); // $ExpectType Observable<{}>
+});
+
+it('should infer generic type', () => {
+  const a = of([{ name: 'abc', id: 123 }, { name: 'def', id: 256 }]).pipe<string>(pluck('name')); // $ExpectType Observable<string>
+});
+
+it('should support nested object', () => {
+  const a = of([{ name: 'abc', id: 123, address: { postcode: 1 } }, { name: 'def', id: 256, address: { postcode: 2 } }]).pipe(pluck('address', 'postcode')); // $ExpectType Observable<{}
+});
+
+it('should accept string only', () => {
+  const a = of([{ name: 'abc', id: 123}, { name: 'def', id: 256}]).pipe(pluck(1)); // $ExpectError
+});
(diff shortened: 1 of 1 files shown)


spec-dtslint/operators/pluck-spec.ts
```diff
@@ -0,0 +1,18 @@
...
+it('should infer generic type', () => {
+  const a = of([{ name: 'abc', id: 123 }, { name: 'def', id: 256 }]).pipe<string>(pluck('name')); // $ExpectType Observable<string>
+});
+
+it('should support nested object', () => {
+  const a = of([{ name: 'abc', id: 123, address: { postcode: 1 } }, { name: 'def', id: 256, address: { postcode: 2 } }]).pipe(pluck('address', 'postcode')); // $ExpectType Observable<{}
+});
+
+it('should accept string only', () => {
+  const a = of([{ name: 'abc', id: 123}, { name: 'def', id: 256}]).pipe(pluck(1)); // $ExpectError
```
timdeschryver: Should we also add a test that enforces a parameter?

```ts
const a = of([{ name: 'abc', id: 123}, { name: 'def', id: 256}]).pipe(pluck()); // $ExpectError
```

dkosasih:
Should I port that over?
I don't mind doing a PR for that.

cartant:
@dkosasih Sure, if you want. There is an issue here: #4131.

It will also need a rest parameters overload like the one in `pipe` - i.e. with trailing rest parameters.

cartant:
Actually, using a trailing rest parameters signature would be a breaking change. The typed signatures should be added and its current signature should be kept.

dkosasih:
@cartant I think I lost you there :smile:

dkosasih:
I have change the tests for this operator - this should only be be merge if the pull request for #4131 is approved.
Thanks.

spec-dtslint/operators/pluck-spec.ts
```diff
@@ -0,0 +1,112 @@
...
+              aa: 1
+            }
+          }
+        }
+      }
+    }
+  }).pipe(pluck('address', 'postcode', 'x', 'y', 'z', 'aa')); // $ExpectType Observable<number>
+});
+
+it('should support nested object of 6 layer depth', () => {
```
cartant: Isn't this the same as the test above?

spec-dtslint/operators/pluck-spec.ts
```diff
@@ -0,0 +1,112 @@
...
+          }
+        }
+      }
+    }
+  }).pipe(pluck('address', 'postcode', 'x', 'y', 'z', 'aa', 'ab')); // $ExpectType Observable<{}>
+});
+
+it('should accept string only', () => {
+  const a = of({ name: 'abc', id: 123}).pipe(pluck(1)); // $ExpectError
+});
```
cartant: Another test that shows that if a string that is not a key of the type is specified,  `{}` should be inferred, as it should fall through to the rest parameters signature.

dkosasih:
@cartant Do you think we could merge this test already since the PR for #4131 has been merged?

cartant:
You will need to rebase this PR against `master`, first, so that the tests pass.

Issue #4009, opened 2018-08-11
switchAll Operator is not documented


JWO719:
The SwitchAll operator is visible in the new Docs, but it is more or less empty. A detailed explanation how this operator is supposed to work and what it is useful for, would be really nice.

see #4004

luillyfe:
@JWO719 I'll take care of it.

cartant:
Closing this as it's addressed in https://github.com/ReactiveX/rxjs/pull/4242

PR #4242, opened 2018-10-10
docs($browser): Created examples for using switchAll operator

luillyfe:
Added description for switchAll, 2 examples of how to use it and links to related operators

it will fix this issue, https://github.com/ReactiveX/rxjs/issues/4009

### DIFF
--- a/src/internal/operators/switchAll.ts
+++ b/src/internal/operators/switchAll.ts
@@ -1,9 +1,82 @@
-import { OperatorFunction, ObservableInput } from '../types';
+import {OperatorFunction, ObservableInput, MonoTypeOperatorFunction} from '../types';
 import { switchMap } from './switchMap';
 import { identity } from '../util/identity';
 
 export function switchAll<T>(): OperatorFunction<ObservableInput<T>, T>;
 export function switchAll<R>(): OperatorFunction<any, R>;
+
+/**
+ * Converts a higher-order Observable into a first-order Observable
+ * producing values only from the most recent observable sequence
+ *
+ * <span class="informal">Flattens an Observable-of-Observables.</span>
+ *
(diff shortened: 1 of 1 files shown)


src/internal/operators/switchAll.ts
```diff
@@ -1,10 +1,75 @@
...
+ * import { fromEvent } from "rxjs";
+ * import { switchAll, map } from "rxjs/operators";
+ *
+ * const clicks = fromEvent(document, 'click');
+ * const higherOrder = clicks.pipe(map((ev) => interval(1000)));
+ * const firstOrder = higherOrder.pipe(switchAll());
+ * firstOrder.subscribe(x => console.log(x));
+ * ```
+ *
+ * Hit on the screen and it will let you track the last 10 mouse position.
```
pertrai1: This would be a good candidate to have a stackblitz example to go along with.

luillyfe: This already have a stackbliz example but since we do not have support from it (stackbliz) I could not added it. https://stackblitz.com/edit/switchall-rxjs

benlesh:
@luillyfe it appears there are just some linting issues with this PR if you can fix them. Trailing whitespaces it seems.

luillyfe:
@benlesh sure I'll take care of it.

src/internal/operators/switchAll.ts
```diff
@@ -1,10 +1,65 @@
-import { OperatorFunction, ObservableInput } from '../types';
+import {OperatorFunction, ObservableInput, MonoTypeOperatorFunction} from '../types';
```
cartant: This PR needlessly imports `MonoTypeOperatorFunction` which is not used.

Issue #4052, opened 2018-08-23
docs: add example to publish operator


pertrai1:
### Documentation Related To Component:

https://rxjs-dev.firebaseapp.com/api/operators/publish

### Please check those that apply

- [x] needs example

### Description Of The Issue

The new docs do not have a working example of how publish could be used in a real world scenario

luillyfe:
I'll take care of it.

cartant:
Closing this as `publish` now appears to have an example.

PR #4243, opened 2018-10-10
docs($browser): It adds a tiny example of how to use publish operator.

luillyfe:
https://github.com/ReactiveX/rxjs/issues/4052

### DIFF
--- a/src/internal/operators/publish.ts
+++ b/src/internal/operators/publish.ts
@@ -19,14 +19,11 @@ export function publish<T>(selector: MonoTypeOperatorFunction<T>): MonoTypeOpera
  * ![](publish.png)
  *
  * ## Examples
- * No one of connectable$ observables start emitting values until the screen gets click (calling connect method of
- * the observable) . After clicked on the screen it will output values every two second before each one
- * (connectable$) gets complete.
- * ```javascript
- * import { of, zip, interval, fromEvent } from "rxjs";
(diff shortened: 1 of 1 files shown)


src/internal/operators/publish.ts
```diff
@@ -14,14 +14,42 @@ export function publish<T>(selector: MonoTypeOperatorFunction<T>): MonoTypeOpera
  * Returns a ConnectableObservable, which is a variety of Observable that waits until its connect method is called
  * before it begins emitting items to those Observers that have subscribed to it.
  *
+ * <span class="informal">Share the same source, prevent it from emitting values until calling connect.</span>
```
luillyfe: @benlesh  Because the example you suggest, I think this does not longer make sense. should I delete it?

src/internal/operators/publish.ts
```diff
@@ -14,14 +14,42 @@ export function publish<T>(selector: MonoTypeOperatorFunction<T>): MonoTypeOpera
...
+ *
  * ![](publish.png)
  *
+ * ## Examples
```
pertrai1: With examples, we have been adding the log output at the end so someone can easily see what that would be and compare it against the example. Would be nice to show the log here as well.

luillyfe: Awesome 👍

luillyfe:
@benlesh, changes you requested are already done.

PR #4252, opened 2018-10-12
test(dtslint): add dtslint test for publish operator (#4093)

dkosasih:
**Description:** Add dtslint test for Publish operator

### DIFF
--- a/spec-dtslint/operators/publish-spec.ts
+++ b/spec-dtslint/operators/publish-spec.ts
@@ -6,9 +6,13 @@ it('should support empty parameter', () => {
   const a = of(1, 2, 3).pipe(publish()); // $ExpectType Observable<any>
 });
 
+it('should infer when type is specified', () => {
+  const a = of(1, 2, 3).pipe<number>(publish()); // $ExpectType Observable<number>
+});
+
 it('should infer correctly with parameter', () => {
-  const a = of(1, 2, 3).pipe(publish((x: Observable<number>) => x)); // $ExpectType Observable<number>
-  const b = of('a', 'b', 'c').pipe(publish((x: Observable<string>) => x)); // $ExpectType Observable<string>
+  const a = of(1, 2, 3).pipe(publish(x => x)); // $ExpectType Observable<number>
+  const b = of('a', 'b', 'c').pipe(publish(x => x)); // $ExpectType Observable<string>
 });
 
 it('should enforce type on selector', () => {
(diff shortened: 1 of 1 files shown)


spec-dtslint/operators/publish-spec.ts
```diff
@@ -0,0 +1,16 @@
+import { of, Observable } from 'rxjs';
+import { publish } from 'rxjs/operators';
+
+it('should support empty parameter', () => {
+  // Typescript limitations - inference only happen in one direction from operator to pipe
+  const a = of(1, 2, 3).pipe(publish()); // $ExpectType Observable<any>
```
cartant: I think you should also include const `a = of(1, 2, 3).pipe(publish<number>());` to show that `Observable<number>` is inferred if the type parameter is specified. This shows that inference works in one direction, despite `publish` returning `UnaryFunction<Observable<T>, ConnectableObserable<T>>` instead of  `UnaryFunction<Observable<T>, Obserable<T>>`.

spec-dtslint/operators/publish-spec.ts
```diff
@@ -0,0 +1,16 @@
...
+});
+
+it('should infer correctly with parameter', () => {
+  const a = of(1, 2, 3).pipe(publish((x: Observable<number>) => x)); // $ExpectType Observable<number>
+  const b = of('a', 'b', 'c').pipe(publish((x: Observable<string>) => x)); // $ExpectType Observable<string>
```
cartant: I don't think the parameter types should be specified here, leave them to be inferred: `publish(x => x)`.

dkosasih: @cartant I have removed the types for selectors. Thanks :)

spec-dtslint/operators/publish-spec.ts
dkosasih: done :)

cartant:
The problem is that the `next` version of TypeScript **is** able to infer `Observable<number>`, so the expectation fails when dtslint tests with that version.

At the moment, it's not possible to [specify an upper bound for the TypeScript version used by dtslint](https://github.com/Microsoft/dtslint/issues/117), so the test's expectation is going to have to be removed:

```ts
it('should support empty parameter', () => {
  // Here, TypeScript versions 3.1 and earlier infer Observable<any>. However,
  // the next version infers Observable<number>. It's not possible to specify
  // an upper bound for the TypeScript version used by dtslint, so an
  // expectation cannot be applied.
  const a = of(1, 2, 3).pipe(publish());
});
```

dkosasih:
should I just amend the test to exactly like above until we upgrade to typescript version that support the inference?

cartant:
Yep. You can add a TODO if you want, but won't be testable until 3.2 is the *minimum* supported version of TS - which means not before RxJS v7 at the earliest.

dkosasih:
@cartant I have changed the test, could you please review? Thank you!

PR #4256, opened 2018-10-13
chore(marketing): Give Logo a Hair Cut

pertrai1:
**Description:**

The outline of the logo now does not have a pixelated white border
around the edges. Removing this allows it to be used on other color
backgrounds besides white.

### Comparison

Left side is current - Right side is hair cut - (Click on image to see it in full size)

rxjs-side-by-side-example


pertrai1:
Thanks for catching that @cartant. I was using the border to check for any out of balance pixels that might have been missed. It is now removed.

rgbkrk:
This is great, thank you!

JWO719:
I also like it very much!

benlesh:
cartant, is that .svg in the repo? I feel like it's been floating around for a while. Or different people have an SVG, but we haven't put one officially anywhere. Perhaps we should keep the assets here somewhere (but make sure we aren't publishing them with a package).

benlesh:
Thanks, @pertrai1

PR #4261, opened 2018-10-16
docs(playground): Ability to use console for playground

pertrai1:
**Description:**

This adds rxjs to the global namespace in the docs app so that anyone can use the console as a playground in learning and testing while looking at the docs. An example would be:

```
rxjs.interval(4000).pipe(rxjs.operators.take(4)).subscribe(console.log);
```

### DIFF
--- a/docs_app/src/index.html
+++ b/docs_app/src/index.html
@@ -151,6 +151,7 @@
       <b><i>This website requires JavaScript.</i></b>
     </h2>
   </noscript>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/rxjs/6.3.3/rxjs.umd.min.js" async></script>
   <script src="assets/js/devtools-welcome.js"></script>
 </body>
 </html>
--- a/docs_app/src/assets/js/devtools-welcome.js
+++ b/docs_app/src/assets/js/devtools-welcome.js
@@ -4,11 +4,9 @@ var welcomeText = (
   '| |_) |\\ \\/ /  | \\___ \\  \n' +
   '|  _ <  >  < |_| |___) |    \n' +
   '|_| \\_\\/_/\\_\\___/|____/ \n' +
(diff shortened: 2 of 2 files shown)


JWO719:
Basically I like the idea. Thanks for coming up with it, but I would be interested how much the bundle size is affected by it?! I'm not sure if many users will make use of that feature?

pertrai1:
Hey, I was going based off of an issue, https://github.com/ReactiveX/reactivex.github.io/issues/374 , that I saw and thought it would be good to have that as an option if someone wants to get a quick example working. I guess someone will have to put a twitter poll out and see what the response is :-)

pertrai1:
Also, what bundle size are you referring to? This is using cdn so it should not have anything to do with bundle size of this end.....at least I don't think

JWO719:
I was referring to the initial download volume when you initially visit the website. What's the difference in kb between loading the website with this playground and without?

pertrai1:
@JWO719 Hey bro, thanks for clarifying. I will take a look and get back to you

pertrai1:
@benlesh changed to `docs()`

docs_app/src/index.html
```diff
@@ -151,6 +151,7 @@ <h2 style="color: red; text-align: center;">
       <b><i>This website requires JavaScript.</i></b>
     </h2>
   </noscript>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/rxjs/6.3.3/rxjs.umd.min.js"></script>
```
benlesh: This can be loaded asynchronously, to prevent it from blocking the main application loading. Please add the async attribute `<script src="" async></script>`. This will allow content loaded events to fire before that script is done loading.

JWO719:
@pertrai1 can I support you in some way to solve the requested changes? I would really like to merge this one!

rsimpson2:
@JWO719 @benlesh I have made the quick change asked for.

PR #4278, opened 2018-10-21
Update

Pranjal-Soni:
**Description:**


pertrai1:
@Pranjal-Soni thanks for the PR. Can you explain what this is needed for? It looks like it is removing a whitespace that the rest of the rules have.

cartant:
This PR has a meaningless title, has no explanation whatsoever and removes a single space from `.eslintrc` - making the content bid that file more difficult to read.

PR #4282, opened 2018-10-25
test(dtslint): add dtslint test for reduce operator (#4093)

dkosasih:
**Description:** Add dtslint for reduce operators

### DIFF
--- a/spec-dtslint/operators/reduce-spec.ts
+++ b/spec-dtslint/operators/reduce-spec.ts
@@ -13,12 +13,7 @@ it('should infer correctly for accumulator of type array', () => {
   const a = of(1, 2, 3).pipe(reduce((x: number[], y: number, i: number) => x, [])); // $ExpectType Observable<number[]>
 });
 
-it('should infer correctly for accumulator that change type', () => {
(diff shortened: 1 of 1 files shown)


spec-dtslint/operators/reduce-spec.ts
```diff
@@ -0,0 +1,24 @@
...
+
+it('should infer correctly ', () => {
+  const a = of(1, 2, 3).pipe(reduce((x, y, z) => x + 1)); // $ExpectType Observable<number>
+});
+
+it('should infer correctly for accumulator of type array', () => {
+  const a = of(1, 2, 3).pipe(reduce((x: number[], y: number, i: number) => x, [])); // $ExpectType Observable<number[]>
+});
+
+it('should infer correctly for accumulator that change type', () => {
```
cartant: I think you should remove this test. It's actually incorrect and it only passes because of a bug in the typings.

`reduce` should not allow this. If the `seed` is not specified, the first value becomes the seed and the function is called only for the second and subsequent values, so it should not be possible to specify an accumulator that has type that's different to that of the source observable's values unless a seed is specified.

The typings for `reduce` and `scan` need some work. Doing it is on my TODO list.

dkosasih: the irrelevant test case removed. Thanks @cartant

PR #4283, opened 2018-10-26
docs(throwError): mistake on description for demo

NirvanaChow:
the demo shows that when 2 is emitted, an error throws, but description says 13. Maybe it's an previous version -.-

**Description:**


benlesh:
Thanks @NirvanaChow

PR #4290, opened 2018-10-27
docs(install): Enfoced correct node version for docs_app

tobmaster:
**Description:**
To prevent confusion I fixed the required node version in package.json to the required version 10 and enforced it in npmrc to prevent an errornous install

### DIFF
--- a/package.json
+++ b/package.json
@@ -239,7 +239,8 @@
     "xmlhttprequest": "1.8.0"
   },
   "engines": {
-    "npm": ">=2.0.0"
+    "node": ">=10.0.0",
+    "npm": ">=6.0.0"
   },
   "typings": "./dist/package/Rx.d.ts",
   "ng-update": {


docs_app/package.json
```diff
@@ -38,7 +38,7 @@
     "~~clean-generated": "node --eval \"require('shelljs').rm('-rf', 'src/generated')\""
   },
   "engines": {
-    "node": ">=8.9.1 <9.0.0",
+    "node": ">=10.0.0",
     "npm": ">=2.0.0"
```
JWO719: Sorry I didn't saw that but could you also set npm to >= 6?

JWO719:
For travis build you have to remove the strict thingy in the npmrc

tobmaster:
Hmmm but the docs don't build with <10. So Travis shouldn't build it with lower version as well?

benlesh:
@JWO719 where does this PR stand?

JWO719:
It doesn't seem that there isn't an issue at this moment. Please re-open a PR if you can take care of it

PR #4292, opened 2018-10-27
chore(typings): fix typo in api-list.component.spec.ts

arthur-melo:
**Description:**
Small typo fix in api-list.component.spec.ts.


arthur-melo:
I've created a new PR (Previous #4255) because the original commit wasn't in a separate branch, as required by the guidelines.

Issue #4148, opened 2018-09-18
"Getting Started" should point to Introduction


benlesh:
The landing page is a sore point right now.

- The `Getting Started` link that is so prominent needs to point to the "Introduction" content.

- A link to the API documentation should also be shown prominently on the main page, but not in "getting started"

JWO719:
Getting started link is already fixed, just the api link needs to be improved

dzhavat:
This issue can be closed now. The described changes have been made and are already live.

cartant:
Thanks for the heads up, @dzhavat

PR #4301, opened 2018-10-30
docs(content): add 'API DOCS' CTA link to main page (#4148)

sunny-b:
**Description: This adds a "API Docs" CTA link to the main page of the marketing site. This addresses the most recent comment made by @JWO719 in #4148**

**Related issue (if exists): #4148**

image

### DIFF
--- a/docs_app/src/styles/1-layouts/_marketing-layout.scss
+++ b/docs_app/src/styles/1-layouts/_marketing-layout.scss
@@ -58,6 +58,10 @@ section#intro {
     justify-content: center;
     align-items: center;
 
+    .hero-cta + .hero-cta {
+      margin-top: 15px;
+    }
+
     @media  (max-width: 780px) {
       display: flex;
       flex-direction: column;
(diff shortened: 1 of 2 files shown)


docs_app/src/styles/1-layouts/_marketing-layout.scss
```diff
@@ -58,6 +58,10 @@ section#intro {
     justify-content: center;
     align-items: center;
 
+    .hero-cta + .hero-cta {
```
pertrai1: Does it not work just to use `.hero-cta` without the sibling?

sunny-b: It would work both ways. I was specifically trying to target the second `.hero-cta` to give it some space from the first. Removing the sibling selector would cause the "Reactive Extensions Library for JavaScript" to get pushed upwards.

Sibling Selector:
image

No sibling selector:
image

There isn't a big difference. If it's preferrable to apply the styling to both buttons I can certainly remove the sibling selector.

JWO719: Nah it's fine like it is, thanks a lot!

PR #4309, opened 2018-10-31
Build bazel umds

alan-agius4:
**Description:**
This PR adds AMD ids in UMD bundles, sourceMapUrl in UMD bundle, and remove Bazel files

//cc @alexeagle & @gregmagolan 

Ps: I can split this into multiple PR, if you'd like.

### DIFF
--- a/tools/rollup-bundle.js
+++ b/tools/rollup-bundle.js
@@ -31,6 +31,9 @@ module.exports = function rollupBundle(options) {
     return bundle.generate({
       format: 'umd',
       name: 'rxjs',
+      amd: {
+        id: 'rxjs'
+      },
       sourcemap: true,
     });
   }).then(function (result) {
--- a/tools/rollup-bundle.js
+++ b/tools/rollup-bundle.js
@@ -1,0 +1,47 @@
+var _ = require('lodash');
+
+var rollup = require('rollup');
+var rollupAlias = require('rollup-plugin-alias');
+var rollupInject = require('rollup-plugin-inject');
+var rollupNodeResolve = require('rollup-plugin-node-resolve');
+
+var fs = require('fs');
+var tslib = require('tslib');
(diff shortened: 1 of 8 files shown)


tools/rollup-bundle.js
```diff
@@ -0,0 +1,50 @@
...
+          return ['tslib', key];
+        }),
+      }),
+    ],
+  }).then(function (bundle) {
+    return bundle.generate({
+      format: 'umd',
+      name: 'rxjs',
+      amd: {
+        id: 'rxjs'
```
alexeagle: this is the only line I recognize as being load-bearing. Why is the rest of the change so large? is there some refactoring going on at the same time?

alan-agius4: As far as I remember, the current rollup version used by RXJS didn’t support AMD ids. 

Since then Rollup API has changed, which caused a change in the build, and preferred the extract it into a common function.

Re the bazel files, Greg informed me that we should probably remove these files.

alexeagle: I see - depending on what the rxjs maintainers prefer, it might be easier to land in two PRs: update rollup version, and separately change the id and remove Bazel files

alan-agius4:
I am open for both as well. Depends on what they prefer. I can split them
in several Pr if it’s preferred.

alan-agius4:
@pertrai1, there are 2 commits, and thus there is no difference in the changelog, apart from that these are `build` related stuff, thus they won't be in the changelog.

alexeagle:
Thinking a bit more: removing the Bazel files is a breaking change and should probably land later?

alexeagle:
At least the rollup upgrade should be it's own commit

alexeagle:
@benlesh is out, @kwonoj can you help us get this merged? thx

PR #4310, opened 2018-10-31
fix(bundle): don't export `operators` twice

alan-agius4:
At the moment operators are exported twice for UMDs.

https://github.com/ReactiveX/rxjs/blob/master/src/internal/Rx.ts#L181-L183

and

https://github.com/ReactiveX/rxjs/blob/master/src/internal/umd.ts#L9-L10

Follow up of https://github.com/ReactiveX/rxjs/pull/4309

### DIFF
--- a/compat/umd.ts
+++ b/compat/umd.ts
@@ -2,12 +2,8 @@
   NOTE: This is the global export file for rxjs v6 and higher.
  */
 
-/* rxjs */
+/* rxjs and rxjs.operators */
 export * from './Rx';
-
-/* rxjs.operators */
-import * as _operators from 'rxjs/operators';
-export const operators = _operators;
 
 /* rxjs.testing */
 import * as _testing from 'rxjs/testing';


src/internal/umd.ts
```diff
@@ -2,13 +2,9 @@
   NOTE: This is the global export file for rxjs v6 and higher.
  */
 
-/* rxjs */
+/* rxjs and rxjs.operators */
 export * from '../index';
```
cartant: I cannot see that an `operators` export is effected by the above export declaration, so IMO the export declaration below is needed.

alan-agius4: What is effected is that you will get warnings during build time.

When I updated rollup in my other PR

cartant: Warnings or not, if you cannot point out where the `operators` export will be effected - if it's removed here - this change is unsafe.

alan-agius4: Had a look, the UMD output is the same. As rollup itself will not add the second export to the exports list. So they will only contain a single `exports.operators`

https://github.com/rollup/rollup/blob/273ac735b155ea63a6e8bf1d0862c872cb3e94b4/src/Graph.ts#L583-L598

cartant: It's only the non-`compat` change that I think is unsafe. I can see where it's exported twice for the `compat` bundle and the change to the file in the `compat` directory is fine by me.

alan-agius4: Agreed. Updated

benlesh:
Thanks, @alan-agius4

PR #4312, opened 2018-10-31
build(bundles): inline sourcesmap and stop shipping sourcecode

alan-agius4:
- Source codes are not needed for sourcemaps when having `inlineSource`
- Yesterday it was discussed that we should probably make `rxjs` stop publishing `src/` while still having sourceMaps to work.

With the other PR https://github.com/ReactiveX/rxjs/pull/4309 downstream users shall no longer require to compile RxJs code in order to get it working in Bazel thus there is no need to ship it and as a matter of fact we're remove Bazel.

//cc @alexeagle & @gregmagolan

### DIFF
--- a/.make-packages.js
+++ b/.make-packages.js
@@ -112,11 +112,6 @@ createImportTargets(importTargets, "_esm2015/", ESM2015_PKG);
 // Make the distribution folder
 mkdirp.sync(PKG_ROOT);
 
-// Copy over the sources
-copySources('src/', SRC_ROOT_PKG);
-// Copy legacy-reexport sources
-copySources('legacy-reexport/', SRC_ROOT_PKG);
-
 copySources(CJS_ROOT, CJS_PKG);
 fs.copySync(LEGACY_REEXPORT_ROOT, CJS_PKG, {overwrite: false, errorOnExist: true});
 
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -3,6 +3,7 @@
     "removeComments": true,
     "preserveConstEnums": true,
     "sourceMap": true,
+    "inlineSources": true,
     "strictFunctionTypes": true,
     "noImplicitAny": true,
     "noImplicitReturns": true,


kwonoj:
Woule you elaborate for this PR's reason / what to solve? Plus with declarationmap in further, I think we will still need to include src codes anyway.

alan-agius4:
Updated description

kwonoj:
I see reasoning, but I do plan to enable declarationMap option in latest compiler which will need ship source still. If we enable inline sources & sources, wouldn't it be duplicate? / and is there any compiler issues by shipping source but not inlining?

alan-agius4:
I think in that case in makes sense to keep the source. 

@alexeagle what to do think?

alexeagle:
My take on this is that no npm users expect to be able to compile typescript sources they fetch in the package, because doing so requires knowing how the typescript compiler should be invoked (maybe just the right tsconfig.json file but maybe some directories are compiled separately or maybe there are command line args to tsc that aren't in the tsconfig, etc)

So Angular team is not shipping any TS files to npm and we are stopping that practice with Bazel as well.

Obviously you can ship .js.map and .d.ts.map files that have the references to original TS line numbers but not ship the sources. That seems like the smallest package you can ship and doesn't harm your users?

kwonoj:
> .d.ts.map files that have the references to original TS line numbers but not ship the sources.

@alexeagle so far I know `d.ts.map` doesn't inline sources and source need to be placed to allow navigate (i.e in code editors `go to definition`) etcs, reason I'd preserve to ship source in pkg , totally agreed it's not for direct building on user's end.

does `d.ts.map` works without shipping sources? then shipping source obviously not required.

kwonoj:
I was following https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#new---declarationmap `Language Services can also now understand these map files, and uses them to map declaration-file based definition locations to their original source, when available.` and so far when I tried with my pkgs, it seems source was needed.

benlesh:
Currently, RxJS must support CJS, ES5 and ES2015 users. If we suddenly switch to inline source maps, it will inflate the overall size of the bundle considerably for no real gain, and in fact, there may be people using the TS source directly. :grimacing:

alexeagle:
I don't think anyone cares that strongly about this one.

PR #4322, opened 2018-11-06
docs(ci): try autodeployment

JWO719:
the firebase token was removed because I put it in the travis settings. Might make changes to it easier.
Additionally I have added the tests2png script to generate the marble diagrams for the docs.

Not completly sure if docs deployment now works automatically maybe we have to take another round :D


JWO719:
FYI: I will change the travis file to deploy this version of the docs. When everything's working I will restrict it again to only deploy master

PR #4323, opened 2018-11-07
test(dtslint): add dtslint test for sample operator (#4093)

dkosasih:
**Description:** Add dtslint test for Sample operator

### DIFF
--- a/spec-dtslint/operators/sample-spec.ts
+++ b/spec-dtslint/operators/sample-spec.ts
@@ -1,0 +1,10 @@
+import { of, interval } from 'rxjs';
+import { sample, delay } from 'rxjs/operators';
+
+it('should enforce parameter', () => {
+  const a = of(1, 2, 3).pipe(sample()); // $ExpectError
+});
+
+it('should accept observable as notifier parameter', () => {
+  const a = interval(1000).pipe(sample(of(1).pipe(delay(2000)))); // $ExpectType Observable<number>
+});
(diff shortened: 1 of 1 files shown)


spec-dtslint/operators/sample-spec.ts
```diff
@@ -0,0 +1,10 @@
+import { of, interval } from 'rxjs';
+import { sample, delay } from 'rxjs/operators';
+
+it('should enforce parameter', () => {
+  const a = of(1, 2, 3).pipe(sample()); // $ExpectError
+});
+
+it('should accept observable as notifier parameter', () => {
+  const a = interval(1000).pipe(sample(of(1).pipe(delay(2000)))); // $ExpectType Observable<number>
```
timdeschryver: What about doing "just" the following here, to not make it "too complex"?

```
const a = of(1, 2, 3).pipe(sample(of(4, 5, 6))); // $ExpectType Observable<number>
const b = of(1, 2, 3).pipe(sample(of('a', 'b', 'c'))); // $ExpectType Observable<number>
```

cartant: Yeah, I agree with Tim. The tests aren't executed, so we're only interested in the types.

dkosasih: OK. Will change that.
I was writing the test as simple as what Tim has pointed out. But, I thought might as well writing a sample that works.

PR #4325, opened 2018-11-08
create UMD bundle with Rollup

### LINKED ISSUE #4324: HELP WANTED: Need UMD global builds for RxJS 7 (experimental branch)
## WANTED

Help creating the global umd files for people that want to use script tags and use everything...

Needs to output:

- package/bundles/rxjs.umd.js
- package/bundles/rxjs.umd.js.map
- package/bundles/rxjs.umd.min.js
- package/bundles/rxjs.umd.min.js.map

We don't want

- Weird `webpack://` urls in the source maps file. source map paths must be relative to the bundle like `../src/*`.
- Minified versions that don't actually set `rxjs` in global scope (This seems to be the default with webpack :man_shrugging: )

### NOTES

- There's already an npm script called `build:umd`. It should work there.
- The general idea of what exists is to copy sources to `package/src` and build from that, since that directory needs to be there anyhow.
- Needs to work as part of `npm run build`. We don't want to change anything else about the `build` script, it's building everything else (seemingly) just fine. 
- Can't additionally alter the structure of `package/`. Everything that's being put there in the build process needs to be there.
- Just needs to put bundles and their sourcemaps under `package/bundles`
- We don't want or need a whole bunch of new npm scripts
- Needs to work on Windows.

Experimental branch can be found at: https://github.com/reactivex/rxjs/tree/experimental

Rich-Harris:
This seems to work — fixes #4324


benlesh:
Thanks @Rich-Harris! This is an awesome start for me (just a couple things off, but easy to fix). I started all of this with Rollup first (it's what we were using in v6, but it was a really old setup and I couldn't get it working). This is much cleaner and neater than everything I tried.

benlesh:
It's also much smaller than whatever I was doing with Webpack!  Nice.

PR #4337, opened 2018-11-12
test(dtslint): add expand

timdeschryver:
Description:
This PR adds dtslint tests for `expand`.

### DIFF
--- a/spec-dtslint/operators/expand-spec.ts
+++ b/spec-dtslint/operators/expand-spec.ts
@@ -1,0 +1,47 @@
+import { of, asyncScheduler } from 'rxjs';
+import { expand } from 'rxjs/operators';
+
+it('should infer correctly', () => {
+  const o = of(1, 2, 3).pipe(expand((value, index) => of(value))); // $ExpectType Observable<number>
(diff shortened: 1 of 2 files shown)


spec-dtslint/operators/expand-spec.ts
```diff
@@ -0,0 +1,47 @@
+import { of, asyncScheduler } from 'rxjs';
+import { expand } from 'rxjs/operators';
+
+it('should infer correctly', () => {
+  const o = of(1, 2, 3).pipe(expand((value, index) => of(value))); // $ExpectType Observable<number>
```
timdeschryver: Couldn't write this as the other tests because of the following error:

```
ERROR: 5:37  no-unnecessary-callback-wrapper  No need to wrap 'of' in another function. Just use it directly.
```

cartant: Feel free to switch that rule off in the `tslint.json` in the `spec-dtslint` dir.

timdeschryver: ✔️ Done

PR #4346, opened 2018-11-17
chore(sampleTime): implement sampleTime

cartant:
**Description:**

Adds `sampleTime` to the `experimental` branch.

### DIFF
--- a/src/internal/operators/derived/exhaust-spec.ts
+++ b/src/internal/operators/derived/exhaust-spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
 import { exhaust, mergeMap } from 'rxjs/operators';
-import { of, Observable } from 'rxjs';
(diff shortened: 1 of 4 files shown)


paulpdaniels:
Just curious, what is the use case for this to go into core? The implementation seems trivial, isn't it better suited for userland or a recipes? Is it used that much?

benlesh:
Awesome! Thanks @cartant

benlesh:
@paulpdaniels, this is just one of `sample`'s overloads from RxJS v4 and under. In RxJS 5 we tried to split heavily overloaded operators into different operators to reduce complexity and size when bundled.

paulpdaniels:
@benlesh Yeah I wasn't against it, since I still remember those days fondly. I had just noticed that other operators got rejected from core for their trivial implementations so I was curios what made `sampleTime` special 😝 .

PR #4356, opened 2018-11-18
test(dtslint): add onErrorResumeNext

timdeschryver:
Description:
This PR adds dtslint tests for `onErrorResumeNext`.

Could it be that the following type is missing?

```ts
export function onErrorResumeNext<T, T2, R>(v: ObservableInput<T2>): OperatorFunction<T, R>;
```


cartant:
It's not an operator I've ever used. TBH, its signatures look wrong, to me. I think a lot of those signatures should return observables with a union type. E.g. `Observable<T, T2 | T3>`.

timdeschryver:
I thought the same way while writing the tests.
I already had rewritten the types to see how it looked, what do you think?

cartant:
LGTM.

I noticed that the existing typings (for `onErrorResumeNext` and `onErrorResumeNextStatic`) include a union in some of the signatures:

`...observables: Array<ObservableInput<any> | ((...values: Array<any>) => R)`

The `((...values: Array<any>) => R)` looks wrong to me. How can it make sense for a rest parameters type to be a non-array? I've had a quick look through the implementation and cannot see why it's needed. If you could double check that and remove it if you, too, think it should not be there, that would be great.

timdeschryver:
At first sight, it seems like you're right.
The operators `combineLatest`, `zip` and `withLatestFrom` are also using the same signature tho...

Since I've changed the signatures, I should also update the compat signatures (https://github.com/ReactiveX/rxjs/blob/master/compat/operator/onErrorResumeNext.ts), right?

cartant:
Those other operators have a result selector - ATM, at least - but `onErrorResumeNext` doesn't have one. It looks to me like a copy/paste mistake that happened once.

timdeschryver:
I removed the type, I'll try to update the `compat` typings tomorrow or else during this week.

src/internal/operators/onErrorResumeNext.ts
```diff
@@ -17,8 +17,8 @@ export function onErrorResumeNext<T, T2, T3, T4>(v: ObservableInput<T2>, v2: Obs
 export function onErrorResumeNext<T, T2, T3, T4, T5>(v: ObservableInput<T2>, v2: ObservableInput<T3>, v3: ObservableInput<T4>, v4: ObservableInput<T5>): OperatorFunction<T, T | T2 | T3 | T4 | T5>;
 export function onErrorResumeNext<T, T2, T3, T4, T5, T6>(v: ObservableInput<T2>, v2: ObservableInput<T3>, v3: ObservableInput<T4>, v4: ObservableInput<T5>, v5: ObservableInput<T6>): OperatorFunction<T, T | T2 | T3 | T4 | T5 | T6>;
 export function onErrorResumeNext<T, T2, T3, T4, T5, T6, T7>(v: ObservableInput<T2>, v2: ObservableInput<T3>, v3: ObservableInput<T4>, v4: ObservableInput<T5>, v5: ObservableInput<T6>, v6: ObservableInput<T7>): OperatorFunction<T, T | T2 | T3 | T4 | T5 | T6 | T7>;
-export function onErrorResumeNext<T, R>(...observables: Array<ObservableInput<any>>): OperatorFunction<T, R>;
-export function onErrorResumeNext<T, R>(array: ObservableInput<any>[]): OperatorFunction<T, R>;
+export function onErrorResumeNext<T, R>(...observables: Array<ObservableInput<any>>): OperatorFunction<T, T | R>;
```
timdeschryver: Do you think this typing and the next one make sense?

timdeschryver:
👍 , thanks for the update @cartant !

cartant:
@benlesh FYI, I've reviewed and approved this, but I've not merged it, as it includes code (type signature) changes, that require your approval. But, IMO, it's good to go.

PR #4360, opened 2018-11-19
feat(range): accept one argument

arpadvas:
**Description:**

### DIFF
--- a/spec/observables/range-spec.ts
+++ b/spec/observables/range-spec.ts
@@ -73,6 +73,21 @@ describe('range', () => {
     });
 
   });
+
+  it('should accept only one argument where count is argument and start is zero', () => {
+    const e1 = range(5)
+      .pipe(concatMap((x, i) => of(x).pipe(delay(i === 0 ? 0 : 20, rxTestScheduler))));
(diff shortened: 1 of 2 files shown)


benlesh:
Tentatively, I'm okay with this addition, but:

1. It needs to be discussed with the core team
2. It will have to wait for a minor release. (not a patch).

benlesh:
NOTE: This is a possible breaking change for JavaScript users doing something like:  `range(-5)` and relying on counting up to "zero" which is `undefined`. (I haven't tested this, but it seems very corner-casey anyway)

arpadvas:
> NOTE: This is a possible breaking change for JavaScript users doing something like: `range(-5)` and relying on counting up to "zero" which is `undefined`. (I haven't tested this, but it seems very corner-casey anyway)

True. range(-5) doesn't work. Shall I update the PR so when range(-5) is applied it'd count up to zero (-4, -3, -2, -1, 0)?

cartant:
> Shall I update the PR

Hmm. But `range(-5)` counting up to zero would be nothing like lodash - AFAICT - so my vote would be to not have it count up to zero.

PR #4371, opened 2018-11-22
test(dtslint): add dtslint test for combineLatest operator (#4093)

dkosasih:
**Description:** Add dtslint test for combineLatest operator

### DIFF
--- a/spec-dtslint/operators/combineLatest-spec.ts
+++ b/spec-dtslint/operators/combineLatest-spec.ts
@@ -5,52 +5,52 @@ describe('combineLatest', () => {
   describe('without project parameter', () => {
     it('should infer correctly with 1 param', () => {
       const a = of(1, 2, 3);
-      const b = ['a', 'b', 'c'];
+      const b = of('a', 'b', 'c');
       const res = a.pipe(combineLatest(b)); // $ExpectType Observable<[number, string]>
     });
 
     it('should infer correctly with 2 params', () => {
       const a = of(1, 2, 3);
-      const b = ['a', 'b', 'c'];
-      const c = ['d', 'e', 'f'];
+      const b = of('a', 'b', 'c');
+      const c = of('d', 'e', 'f');
       const res = a.pipe(combineLatest(b, c)); // $ExpectType Observable<[number, string, string]>
     });
 
     it('should infer correctly with 3 params', () => {
       const a = of(1, 2, 3);
-      const b = ['a', 'b', 'c'];
-      const c = ['d', 'e', 'f'];
-      const d = ['g', 'h', 'i'];
+      const b = of('a', 'b', 'c');
+      const c = of('d', 'e', 'f');
+      const d = of('g', 'h', 'i');
       const res = a.pipe(combineLatest(b, c, d)); // $ExpectType Observable<[number, string, string, string]>
     });
 
     it('should infer correctly with 4 params', () => {
       const a = of(1, 2, 3);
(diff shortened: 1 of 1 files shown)


spec-dtslint/operators/combineLatest-spec.ts
```diff
@@ -0,0 +1,110 @@
...
+      const c = ['d', 'e', 'f'];
+      const d = ['g', 'h', 'i'];
+      const e = ['j', 'k', 'l'];
+      const f = ['m', 'n', 'o'];
+      const res = a.pipe(combineLatest(b, c, d, e, f, (a, b, c, d, e, f) => b + c)); // $ExpectType Observable<string>
+    });
+
+    it('should infer correctly with array param', () => {
+      const a = of(1, 2, 3);
+      const b = ['a', 'b', 'c'];
```
cartant: Given that you are testing the acceptance of an array here, I think all other arrays in the other tests should be replaced with `of`:

```ts
const a = of(1, 2, 3);
const b = of('a', 'b', 'c');
const res = a.pipe(combineLatest(b)); // $ExpectType Observable<[number, string]>
```

Those tests use arrays as `ObservableInput`s - as does this test.

The more important array-related signatures are the ones that take [an array of `ObservableInput`s](https://github.com/ReactiveX/rxjs/blob/6.3.3/src/internal/operators/combineLatest.ts#L35-L38). And it should be tested with something like this:

```ts
it('should infer correctly with array param', () => {
  const a = of(1, 2, 3);
  const b = [of('x')];
  const res = a.pipe(combineLatest(b, (a, b) => b)); // $ExpectType Observable<string>
});
```

Note that `a.pipe(combineLatest([of('x'), of('y')]))` should be equivalent to `a.pipe(combineLatest(of('x'), of('y')))` and `a.pipe(combineLatest(...[of('x'), of('y')]))`.

Also, looking at the order of the signatures, I wouldn't be surprised if this test exposes a bug.

spec-dtslint/operators/combineLatest-spec.ts
```diff
@@ -0,0 +1,110 @@
...
+      const d = of('g', 'h', 'i');
+      const e = of('j', 'k', 'l');
+      const f = of('m', 'n', 'o');
+      const res = a.pipe(combineLatest(b, c, d, e, f, (a, b, c, d, e, f) => b + c)); // $ExpectType Observable<string>
+    });
+
+    it('should infer correctly with array param', () => {
+      const a = of(1, 2, 3);
+      const b = [of('a', 'b', 'c')];
+      const res = a.pipe(combineLatest(b, (a, b) => b)); // $ExpectType Observable<Observable<string>>
```
dkosasih: @cartant - This is the one I was asking you about.
I would expect the return would be `Observable<[number, string, string]>`
Instead, the typing when I hover over res was `Observable<Observable<string>>`.
However, when I put this into Stackblitz and console log the result it does return me what I expected.
Please advise.

cartant: The problem is that the typings of the `combineLatest` operator and the `combineLatest` creator function don't match the implementation's behaviour. If a single array is passed, it's assumed that the array contains the observables that are to be combined - that is, the array itself is not treated as an `ObservableInput`.

From here, I think you should comment out this test and add a TODO that references this comment. And, if you are keen, create an issue that references both. I will look at fixing this later.

Thanks for the work you've put into writing the tests.

PR #4396, opened 2018-12-02
test(dtslint): add zipAll

timdeschryver:
Description:
This PR adds dtslint tests for `zipAll`.

### DIFF
--- a/spec-dtslint/operators/zipAll-spec.ts
+++ b/spec-dtslint/operators/zipAll-spec.ts
@@ -1,0 +1,25 @@
+import { of } from 'rxjs';
+import { zipAll } from 'rxjs/operators';
+
+it('should infer correctly', () => {
+  const o = of(of(1, 2, 3)).pipe(zipAll()); // $ExpectType Observable<number[]>
+});
+
+it('should support projecting values', () => {
+  const o = of(of(1, 2, 3)).pipe(zipAll(value => String(value))); // $ExpectType Observable<string>
+});
+
+it('should support custom typings on the projector', () => {
+  const o = of(of(1, 2, 3)).pipe(zipAll<string>(value => value)); // $ExpectType Observable<string>
+  const p = of(of(1, 2, 3)).pipe(zipAll<string>((value: string) => value)); // $ExpectType Observable<string>
+});
+
+it('should enforce types', () => {
+  const o = of(1, 2, 3).pipe(zipAll()); // $ExpectError
+});
+
+it('should enforce projector types', () => {
+  const o = of(of(1, 2, 3)).pipe(zipAll('')); // $ExpectError
+  const p = of(of(1, 2, 3)).pipe(zipAll((value: string) => value)); // $ExpectError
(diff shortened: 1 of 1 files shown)


spec-dtslint/operators/zipAll-spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  const o = of(of(1, 2, 3)).pipe(zipAll()); // $ExpectType Observable<number[]>
+});
+
+it('should support projecting values', () => {
+  const o = of(of(1, 2, 3)).pipe(zipAll(value => String(value))); // $ExpectType Observable<string>
+});
+
+it('should support custom typings on the projector', () => {
+  const o = of(of(1, 2, 3)).pipe(zipAll<string>(value => value)); // $ExpectType Observable<string>
+  const p = of(of(1, 2, 3)).pipe(zipAll<string>((value: string) => value)); // $ExpectType Observable<string>
```
cartant: If this test passes, it looks to me like there must be an error in the typings - there is no `string` in the source observable.

cartant:
Yeah, that looks wrong, too. I'll have a look at this later - after I've had a coffee. Perhaps the signatures are there due to issues that arise with early versions of TypeScript? Anyway, I think the `any` signatures should be removed.

spec-dtslint/operators/zipAll-spec.ts
timdeschryver: I sure can 😄 !

timdeschryver:
Woops, seems like this wasn't as simple as I thought 😅 
I removed the `any` typings and the existing tests started to fail.
I reverted my last commit and will take a look at it later.

spec-dtslint/operators/zipAll-spec.ts
benlesh: this might actually be a problem with `pipe` typings, not `zipAll`. :thinking:  I think we can just remove these tests for now and file a separate issue.

timdeschryver:
Is this good to merge? Or is there something I should address?

spec-dtslint/operators/zipAll-spec.ts
```diff
@@ -0,0 +1,24 @@
...
+it('should infer correctly', () => {
+  const o = of(of(1, 2, 3)).pipe(zipAll()); // $ExpectType Observable<number[]>
+});
+
+it('should support projecting values', () => {
+  const o = of(of(1, 2, 3)).pipe(zipAll(value => String(value))); // $ExpectType Observable<string>
+});
+
+it('should enforce types', () => {
+  const o = of(1, 2, 3).pipe(zipAll()); // $ExpectError
```
benlesh: Can we add a couple more here for other types of Observable inputs? Iterators, Promises, Arrays, etc.

timdeschryver: Forgot about this, here you go 😅 
I had to remove the following check

```ts
const p = of(of(1, 2, 3)).pipe(zipAll((value: string) => value)); // $ExpectError
```

PR #4398, opened 2018-12-02
docs(observable): Update Observable docs

dzhavat:
**Description:**

Update Observable in `docs_app` to match the style of Observable in `docs`.

* @benlesh [updated](https://github.com/ReactiveX/rxjs/commit/b256ee170625a0c323852bbb97b25cee7cce613a) old docs related to Observable more than a month ago. This update is just his changes transferred to the Observable in `docs_app`.

* There are also a few small fixes for Observable in `docs`.

### DIFF
--- a/docs_app/content/guide/observable.md
+++ b/docs_app/content/guide/observable.md
@@ -12,13 +12,13 @@ Observables are lazy Push collections of multiple values. They fill the missing 
 ```ts
 import { Observable } from 'rxjs';
 
-const observable = Observable.create(function (observer) {
-  observer.next(1);
-  observer.next(2);
-  observer.next(3);
+const observable = new Observable(subscriber => {
+  subscriber.next(1);
+  subscriber.next(2);
+  subscriber.next(3);
   setTimeout(() => {
-    observer.next(4);
-    observer.complete();
+    subscriber.next(4);
+    subscriber.complete();
   }, 1000);
 });
 ```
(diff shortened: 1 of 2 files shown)


docs_app/content/guide/observable.md
```diff
@@ -28,21 +28,21 @@ To invoke the Observable and see these values, we need to *subscribe* to it:
...
+    subscriber.complete();
   }, 1000);
 });
 
 console.log('just before subscribe');
 observable.subscribe({
-  next: x => console.log('got value ' + x),
-  error: err => console.error('something wrong occurred: ' + err),
-  complete: () => console.log('done'),
+  next(x) { console.log('got value ' + x); },
```
pertrai1: what do you think about changing to using template strings?

docs_app/content/guide/observable.md
```diff
@@ -267,7 +267,7 @@ Conclusion:
 
 ## Anatomy of an Observable
 
-Observables are **created** using `Observable.create` or a creation operator, are **subscribed** to with an Observer, **execute** to deliver `next` / `error` / `complete` notifications to the Observer, and their execution may be **disposed**. These four aspects are all encoded in an Observable instance, but some of these aspects are related to other types, like Observer and Subscription.
+Observables are **created** using `new Observable` or a creation operator, are **subscribed** to with an Observer, **execute** to deliver `next` / `error` / `complete` notifications to the Observer, and their execution may be **disposed**. These four aspects are all encoded in an Observable instance, but some of these aspects are related to other types, like Observer and Subscription.
```
pertrai1: note sure the semantics of this sentence. Might be worth breaking it up after 'creation operator. The sentence starts with are and then after the comma continues with are. Looks to be a run on sentence of sorts

pertrai1: I think that the first part of the sentence, 'These four aspects....' is important, so when 'but' is used, it is negating that first part.

docs_app/content/guide/observable.md
dzhavat: It makes sense to use it. Will make the example a bit more clear.

docs_app/content/guide/observable.md
dzhavat: Will it make it easier to understand if was something like: 

> There are four important aspects when we talk about an Observable instance.
> 
> Observables:
>   * are **created** using `new Observable` or a creation operator.
>   * are **subscribed** to with an Observer.
>   * **execute** to deliver `next` / `error` / `complete` notifications to the Observer.
>   * may be **disposed**.
>
> While all of these aspects are encoded in an Observable instance, some of them are also related to other types, like Observer and Subscription.

dzhavat: @pertrai1 any comments on my suggestions above? :)

PR #4401, opened 2018-12-04
docs(index): fix typo

shoffing:
**Description:**
Simple documentation fix for `zip`

`If the latest parameter is a function, ...`

should be changed to

`If the last parameter is a function, ...`

Parameters are ordinal, not temporal. So "last" makes more sense than "latest".


pertrai1:
👍

PR #4405, opened 2018-12-05
style(Notification): Use enum to safely type the notification kinds

zakhenry:
**Description:**
Rather than using magic strings which are easily use incorrectly, this PR introduces a `enum NotificationKind`. 

This is not a breaking change as all clients that happen to already test the kind against a string will still have the same behaviour, both at compiletime and runtime, with the exception that if they use a notification kind letter that does not exist, they will get a compile time error.

### DIFF
--- a/spec/Notification-spec.ts
+++ b/spec/Notification-spec.ts
@@ -12,7 +12,7 @@ describe('Notification', () => {
   });
 
   it('should not allow convert to observable if given kind is unknown', () => {
-    const n = new Notification('x');
+    const n = new Notification('x' as any);
     expect(() => n.toObservable()).to.throw();
   });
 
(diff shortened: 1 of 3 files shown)


kwonoj:
This is technically breaking changes since string enum only works > 2.4 typescript compiler.

zakhenry:
@kwonoj hmm true, is there a stated minimum supported version of typescript for rxjs?

kwonoj:
we don't explicitly and I feel we may need, but that's also breaking changes 😅 for next major.

kwonoj:
True, but I would prefer `const enum` if possible.

benlesh:
I think this change looks great. Just need to discuss this with the team, and it'll have to wait for the next minor release (likely soon), since it's a new feature.

kwonoj:
I like this but proposing next major and we explicitly declared supported ts version from next major.

Issue #3722, opened 2018-05-21
Proposal: new operator similar to shareReplay()


voithos:
This is related to the apparent current mis-behavior of `shareReplay()` as described in #3354, specifically my comment [here](https://github.com/ReactiveX/rxjs/pull/3354#issuecomment-378690222), parts of which I have included below.

---

This issue proposes adding an operator similar to `shareReplay()`, but one which does not re-subscribe to the source Observable if a previous subscription had completed. As an example, consider an Observable that represents an HTTP request. Ideally:

- It would be cold until subscribed to, because there's no need to execute the expensive operation (in this case, an HTTP request) unless something needs the result. So we cannot use `connect()`.
- After completing, it would not re-subscribe to the source but instead would use something like a `ReplaySubject(1)` to retain the last value (or multiple values, if it's chunked). Thus, `publishReplay(1).refCount()` or the old `shareReplay(1)` behavior isn't ideal because it re-subscribes every time the refCount goes from 0 -> 1.
- If the source subscription gets cancelled before completion, then re-subscribing is ok since we never got a value anyway and thus can't use the underlying `ReplaySubject`.

This new operator would behave similarly to `publishReplay(), refCount()` or the old `shareReplay()` implementation, except it would just keep the `ReplaySubject()` around, even after subscribers have reached 0, _if the source observable completes_. This would allow unsubscribing as normal (which is a problem with the current `shareReplay()` implementation, see #3354), but would ensure that the source observable's emissions are reused if it completes.

Would something like this be desired / useful? From a cursory search, I haven't found any prior art for such an operator.

As for naming, some ideas include `shareOnce()`, `shareReplayOnce()`, `safeShare()`.

Alternatively, it may (?) be possible to do this by having a specialized version of `refCount()` that can be used alongside a normal `publishReplay()`, but I haven't looked into this.

benlesh:
> This issue proposes adding an operator similar to shareReplay(), but one which does not re-subscribe to the source Observable if a previous subscription had completed. 

`shareReplay` doesn't resubscribe to the source if the source completes, it caches the values and the completion and replays them on new subscriptions.

Either way, I'd recommend implementing a user-land operator and publishing it in this case. If it becomes popular or widely used, we can adopt it into the core library.

PR #4411, opened 2018-12-07
feat(shareReplay): added 'disconnect' feature

Malvolio:
**Description:**
Updated shareReplay to support 'disconnect'.  That is, when the refCount drops to 0, the buffer can (optionally) be dropped, and then, upon the next subscription, re-subscribes to the source.

Rather than adding *another* parameter,  a config-object interface has been added. For compatibility, the old parameters still work,  but the disconnect option can only be reached with the config option, so `shareReplay(3)` is equivalent to `shareReplay({ bufferSize: 3})`, but you might instead want `shareReplay({ bufferSize: 3, disconnect: true})`.

This is in response to [issue 3722](https://github.com/ReactiveX/rxjs/issues/3722).

### DIFF
--- a/docs_app/content/guide/operators.md
+++ b/docs_app/content/guide/operators.md
@@ -1,0 +1,141 @@
+# Operators
+
+An operator is just a function that takes an Observable as its input and returns another Observable.  For example, there is an operator called `map()`, that is equivalent to the Array method of the same name. Just as `[1, 2, 3].map(x => x * x)` will yield `[1, 4, 9]`, the Observable created like this:
+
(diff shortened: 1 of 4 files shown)


Malvolio:
> The problem with `shareReplay` has been addressed in #4059 - but that PR has not yet been merged.

Haha, I was talking to Ben Lesh and he suggested I make this change.  I guess he forgot about that PR.

cartant:
Closing this in favour of https://github.com/ReactiveX/rxjs/pull/4059

PR #4414, opened 2018-12-09
test(dtslint): add dtslint test for forkjoin obsrevable (#4093)

dkosasih:
**Description:** Add dtslint test for forkJoin observable static method creation

### DIFF
--- a/spec-dtslint/observables/forkJoin-spec.ts
+++ b/spec-dtslint/observables/forkJoin-spec.ts
@@ -1,0 +1,92 @@
+import { of, forkJoin } from 'rxjs';
+
+it('should infer correctly with 1 parameter', () => {
+  const a = of(1, 2, 3);
+  const res = forkJoin(a); // $ExpectType Observable<number[]>
+});
+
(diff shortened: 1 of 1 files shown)


spec-dtslint/observables/forkJoin-spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  const g = of(1, 2, 3);
+  const res = forkJoin(a, b, c, d, e, f, g); // $ExpectType Observable<any>
+});
+
+it('should infer correctly for array of 1 observable', () => {
+  const a = [of(1, 2, 3)];
+  const res = forkJoin(a); // $ExpectType Observable<number[]>
+});
+
+it('should infer correctly for array of 2 observables', () => {
```
cartant: This test - and the tests following it - should be commented out. They have exposed a bug that will have to be addressed in a separate PR. This one should be inferring:

```ts
// $ExpectType Observable<[string, number]>
```

The tests describe the current behaviour, but that behaviour is not what's intended and it's not how the implementation behaves.

Issue #4424, opened 2018-12-16
Expose isPromise similarly to how isObservable was exposed


ChrisDahmCS:
## Feature Request

isPromise is currently an internal method, but isObservable was recently (6.1) exposed out of internal.

**Describe the solution you'd like**
Similarly to isObservable, isPromise could be exposed instead of being kept internal. 

**Describe alternatives you've considered**
Other teams/groups have basically copied the code from this method to check for isPromise [like the Angular team](https://github.com/angular/angular/blob/master/packages/core/src/util/lang.ts#L14-L18).

benlesh:
As described in my comments on the PR, we don't want to expose `isPromise` and be on the hook for maintaining that for all users. Closing this one. Thanks you for the suggestion!

PR #4425, opened 2018-12-16
feat(isPromise): change isPromise to return bool, add tests

ChrisDahmCS:
isPromise returns a boolean like isObservable instead of a "falsy" value. isPromise is available for more than internal use since Observable has toPromise. Add tests to validate isPromise.

**Description:**
isObservable was exposed from the internal util methods in [6.1.0](https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md#610-2018-05-03) and had the null/undefined errors fixed in [6.2.0](https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md#610-2018-05-03). Exposing isPromise from the internal util methods and updating it to also return a boolean instead of a "truthy" / "falsy" seems consistent.

Related to #4424

### DIFF
--- a/src/internal/util/isPromise.ts
+++ b/src/internal/util/isPromise.ts
@@ -1,3 +1,8 @@
+/**
+ * Tests to see if the object is an ES2015 (ES6) Promise
+ * @see {@link https://www.ecma-international.org/ecma-262/6.0/#sec-promise-objects}
+ * @param value the object to test
+ */
 export function isPromise(value: any): value is PromiseLike<any> {
-  return value && typeof (<any>value).subscribe !== 'function' && typeof (value as any).then === 'function';
+  return !!value && typeof (<any>value).subscribe !== 'function' && typeof (value as any).then === 'function';
 }
--- a/spec/util/isPromise-spec.ts
+++ b/spec/util/isPromise-spec.ts
@@ -1,0 +1,38 @@
+import { of } from 'rxjs';
+import { expect } from 'chai';
+import { isPromise } from 'rxjs/util/isPromise';
+
+describe('isPromise', () => {
+  it('should return true for new Promise', () => {
+    const o = new Promise<any>(() => null);
+    expect(isPromise(o)).to.be.true;
+  });
+
+  it('should return true for a Promise that comes from an Observable', () => {
+    const o: any = of(null).toPromise();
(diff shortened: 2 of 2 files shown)


src/internal/util/isPromise.ts
```diff
@@ -1,3 +1,8 @@
+/**
+ * Tests to see if the object is an ES2015 (ES6) Promise
+ * @see {@link https://www.ecma-international.org/ecma-262/6.0/#sec-promise-objects}
+ * @param value the object to test
+ */
 export function isPromise(value: any): value is PromiseLike<any> {
```
okeefem3: I believe the return type on this function should be a `boolean`, though I see that this matches the code in for `isObservable` [here](https://github.com/ReactiveX/rxjs/blob/master/src/internal/util/isObservable.ts#L8). I could be totally missing something, but this seems incorrect... Could someone on the team confirm??

okeefem2: I believe the return type on this function should be a `boolean`, though I see that this matches the code in for `isObservable` [here](https://github.com/ReactiveX/rxjs/blob/master/src/internal/util/isObservable.ts#L8). I could be totally missing something, but this seems incorrect... Could someone on the team confirm??

kwonoj: No, current return type definition is desired.

ChrisDahmCS: I also wasn't familiar with this, but had left it because isObservable also did not have the boolean return type -
Thanks for the clarification @kozlown and the docs find @okeefem2 !

src/index.ts
```diff
@@ -31,6 +31,7 @@ export { pipe } from './internal/util/pipe';
 export { noop } from './internal/util/noop';
 export { identity } from './internal/util/identity';
 export { isObservable } from './internal/util/isObservable';
+export { isPromise } from './internal/util/isPromise';
```
benlesh: So sorry, we don't want RxJS to be on the hook for supporting a utility function like this. I don't think we should be exporting thing.

I think that the tests and the other changes are valid, however, and I'd like to merge them.

ChrisDahmCS: Makes sense, thank you. I'll get that change for you quick!

benlesh:
Thank you, @ChrisDahmCS!

PR #4430, opened 2018-12-19
docs(meta): add meta tags to docs

JWO719:
As discussed in the Core Team Meeting. I have added some Meta Tags but I have not a single clue about seo and this stuff in particular.

That's the way it is looking (This shows the mergeMap Operators Page)
image

Don't know if we need any additional meta tags?!

### DIFF
--- a/docs_app/src/app/layout/doc-viewer/doc-viewer.component.ts
+++ b/docs_app/src/app/layout/doc-viewer/doc-viewer.component.ts
@@ -93,6 +93,7 @@ export class DocViewerComponent implements OnDestroy {
    * Return a function to actually set them.
    */
   protected prepareTitleAndToc(targetElem: HTMLElement, docId: string): () => void {
+    const descriptionEl = targetElem.querySelector('.api-body > p:nth-child(2)');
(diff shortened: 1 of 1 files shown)


docs_app/src/app/layout/doc-viewer/doc-viewer.component.ts
```diff
@@ -241,4 +247,17 @@ export class DocViewerComponent implements OnDestroy {
         }),
     );
   }
+
+  private addDocumentMetaTags(title: string, description: string | null): void {
+    this.metaService.updateTag({ name: 'twitter:title', content: title });
+    this.metaService.updateTag({ name: 'twitter:card', content: 'summary' });
+    this.metaService.updateTag({ property: 'og:title', content: title });
```
cartant: There are also `twitter:site` and `twitter:image` tags. What the image ought to be, I dunno - maybe the pink eel? - but the site is useful, as it's the URL.

benlesh: @cartant can you please create a separate issue to track this and add @JWO719  as the owner?

PR #4432, opened 2018-12-21
test(dtslint): add dtslint test for from static function (#4093)

dkosasih:
**Description:** Add dtslint test for from observable static function 


dkosasih:
I removed some of my tests because master has had some tests on this as well and it's very similar to what I  had.

PR #4433, opened 2018-12-21
fix(VirtualTimeScheduler): rework flush so it won't lose actions

baizulin:
**Description:**
Previously VirtualTimeScheduler.flush would lose the action that was on a verge of maxFrames limit which rendered testing of observables that are ticking indefinitely impossible. After the fix the user can set maxFrames, flush, make assertions and repeat the process as many times as needed.

**Related issue:**
None

### DIFF
--- a/spec/schedulers/VirtualTimeScheduler-spec.ts
+++ b/spec/schedulers/VirtualTimeScheduler-spec.ts
@@ -1,6 +1,5 @@
 import { expect } from 'chai';
-import * as Rx from 'rxjs/Rx';
-import { SchedulerAction, Subscription, VirtualAction, VirtualTimeScheduler } from 'rxjs';
+import { SchedulerAction, VirtualAction, VirtualTimeScheduler } from 'rxjs';
 
 /** @test {VirtualTimeScheduler} */
 describe('VirtualTimeScheduler', () => {
@@ -67,7 +66,7 @@ describe('VirtualTimeScheduler', () => {
     let count = 0;
     const expected = [100, 200, 300];
 
-    v.schedule<string>(function(this: SchedulerAction<string>, state: string) {
+    v.schedule<string>(function (this: SchedulerAction<string>, state: string) {
       if (++count === 3) {
         return;
       }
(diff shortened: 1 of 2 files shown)


cartant:
Thanks for the PR. I will have a look at this tomorrow. (I also need to look at what's causing the `dtslint` run with TypeScript `next` to effect the errors in the Travis log - something that's unrelated to this PR.)

baizulin:
@cartant Hi! Did you have a chance to take a look?

cartant:
I looked at it briefly.

My concern is that `maxFrames` is no longer relevant given that its [set to `POSITIVE_INFINITY`](https://github.com/ReactiveX/rxjs/blob/6.3.3/src/internal/testing/TestScheduler.ts#L378-L381) in `TestScheduler#run`. And that is the recommended mechanism for writing marble tests.

I'll wait to see what other's have to say about it.

src/internal/scheduler/VirtualTimeScheduler.ts
```diff
@@ -25,7 +25,10 @@ export class VirtualTimeScheduler extends AsyncScheduler {
     const {actions, maxFrames} = this;
     let error: any, action: AsyncAction<any>;
 
-    while ((action = actions.shift()) && (this.frame = action.delay) <= maxFrames) {
+    while ((action = actions[0]) && action.delay <= maxFrames) {
```
benlesh: Seems like this should do the trick:

```
while ((action = actions.shift()) && action.delay <= maxFrames) {
   this.frame = action.delay;
```

spec/schedulers/VirtualTimeScheduler-spec.ts
```diff
@@ -82,12 +81,60 @@ describe('VirtualTimeScheduler', () => {
...
+  });
+
+  it('should pick up actions execution where it left off after reaching previous maxFrames limit', function () {
+    const MAX_FRAMES = 50;
+    const v = new VirtualTimeScheduler(VirtualAction, MAX_FRAMES);
+    const messages: string[] = ['first message', 'second message', 'third message'];
+
+    const actualMessages: string[] = [];
+
+    messages.forEach(function (message: string, index: number): void {
```
benlesh:
```suggestion
    messages.forEach((message, index) => {
```

spec/schedulers/VirtualTimeScheduler-spec.ts
```diff
@@ -82,12 +81,60 @@ describe('VirtualTimeScheduler', () => {
...
   });
+
+  it('should execute only those virtual actions that fall into the maxFrames timespan', function () {
+    const MAX_FRAMES = 50;
+    const v = new VirtualTimeScheduler(VirtualAction, MAX_FRAMES);
+    const messages: string[] = ['first message', 'second message', 'third message'];
+
+    const actualMessages: string[] = [];
+
+    messages.forEach(function (message: string, index: number): void {
```
benlesh:
```suggestion
    messages.forEach((message, index) => {
```

spec/schedulers/VirtualTimeScheduler-spec.ts
```diff
@@ -82,12 +81,60 @@ describe('VirtualTimeScheduler', () => {
...
+      v.schedule(
+        (state: string) => actualMessages.push(state),
+        index * MAX_FRAMES,
+        message
+      );
+    });
+
+    v.flush();
+
+    expect(actualMessages).to.deep.equal(messages.slice(0, 2));
```
benlesh:
```suggestion
    expect(actualMessages).to.deep.equal(['first message', 'second message']);
```

benlesh:
```suggestion
    expect(actualMessages).to.deep.equal(['third message']);
```

benlesh:
@cartant ... `VirtualTimeScheduler` may be used on it's own, and this seems like a valid fix/bug to me.

src/internal/scheduler/VirtualTimeScheduler.ts
baizulin: nope, this changes nothing.
the problem was in `.shift()` mutating the `actions` array **before** actual checks

spec/schedulers/VirtualTimeScheduler-spec.ts
baizulin: yeah, thought about it

spec/schedulers/VirtualTimeScheduler-spec.ts
baizulin: any particular reason?

src/internal/scheduler/VirtualTimeScheduler.ts
benlesh: I don't see the issue you're describing. I know the array is being mutated, but you've moved the mutation one line, and there's no checks to the array in that space.

```ts
const arr = [1, 2, 3, 4, 5];
while(arr.shift()) {
  console.log(arr);
}
```

has the same output as

```ts
const arr = [1, 2, 3, 4, 5];
while (arr.shift()) {
  console.log(arr);
}
```

baizulin:
@cartant @benlesh guys let's fix this bug already, it's been over a month

benlesh:
@baizulin ... I appreciate that you put in the effort to submit this PR, I understand it's been a while and that can be frustrating. Please try to be respectful of our team members and our time, at this time all RxJS work is done by volunteers in our spare time.

src/internal/scheduler/VirtualTimeScheduler.ts
benlesh: I see. You're right, I missed that. Thank you.

Issue #4441, opened 2018-12-29
`mergeScan()`'s accumulator doesn't receive any `index` parameter


martinsik:
## Bug Report

**Current Behavior**
When using `mergeScan()` its `accumulator` function never receives `index` like all other operators.

**Reproduction**
- REPL or Repo link: https://stackblitz.com/edit/rxjs-u27r7g

```
mergeScan((acc, v, index) => {
  // index will be always `undefined`
}, null)
```

**Expected behavior**
The `accumulator` function should receive `index` parameter with ever increasing numbers for every emission from source like `scan` or `mergeMap` do.

**Environment**
- Runtime: all
- RxJS version: 6.3.3

Issue #4348, opened 2018-11-17
`scan` operator description might be confusing when not using any seed


martinsik:
### Documentation Related To Component:

scan

### Please check those that apply

- [ ] typo
- [ ] documentation doesn't exist
- [X] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

### Description Of The Issue

The last paragraph in `scan`'s docblock says (emphasis is mine):

> Returns an Observable that applies a specified `accumulator` function to **each item emitted by the source** Observable. If a `seed` value is specified, then that value will be used as the initial value for the accumulator. If no seed value is specified, **the first item of the source is used as the seed**.

However, the actual `scan` behavior isn't exactly like this:

> ...applies a specified accumulator function to each item emitted by the source Observable

It doesn't apply the `accumulator` function to every item if I don't set any seed. In that case it just passes the first emission through without calling `accumulator`:

```
from(['a', 'b', 'c']).pipe(
  scan<string, number>((acc, value, index) => {
    console.log(value, index);
    return 0;
  }),
).subscribe();
```

This will make the following output even when the first item was `"a"`:

```
b 0
c 1
```

This is the relevant code:
https://github.com/ReactiveX/rxjs/blob/de9d2e4cef759b358c6c5d53b85e337dbf1c9c45/src/internal/operators/scan.ts#L98-L105

Demo: https://stackblitz.com/edit/rxjs6-demo-ynfnan

So even though the sequence started with `"a"` the `accumulator` function was called for the first time for `"b"`. Also the `index` passed started at `0` for the second item (does `index` count how many times the function was called or does it count the number of items that went through this operator?).

>  the first item of the source is used as the seed.

From the last sentence I would guess that the `accumulator` function will be called even for the first value like this order (even though it is a little weird):

```
accumulator('a', 'a', 0);
accumulator(..., 'b', 1);
accumulator(..., 'c', 2);
```

... while in fact it's called like this:

```
accumulator('a', 'b', 0);
accumulator(..., 'c', 1);
```

So maybe I'm just dumb but the description and actual behavior without setting any seed really confused me.

cartant:
You are correct and the documentation is wrong. The `scan` and `reduce` operators behave like `Array.prototype.reduce` when a seed is not specified.

That also means that the typing is incorrect. Fixing the types has been on my TODO list for some time.

martinsik:
[...]

PR #4442, opened 2018-12-30
feat(switchScan): add switchScan() operator

### LINKED ISSUE #2931: Proposal: add a switchScan operator
It will be similar as the [mergeScan][rxjs-mergescan] but instead of merging all events of inner observables, if the outer source emitted a new event, it [switch][rxjs-switch] to the recently accumulated inner observable.

[rxjs-mergescan]: http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-mergeScan
[rxjs-switch]: http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-switch

martinsik:
- [x] Add the operator to Rx
- [x] It must have a `-spec.ts` tests file covering the canonical corner cases, with marble diagram tests
- [x] If possible, write a `asDiagram` test case too, for PNG diagram generation purposes
- [x] The spec file should have a type definition test at the end of the spec to verify type definition for various use cases
- [x] The operator must be documented in JSDoc style in the implementation file, including also the PNG marble diagram image
- [x] The operator should be listed in `doc/operators.md` in a category of operators
- [x] The operator should also be documented. See [Documentation Guidelines](../CONTRIBUTING.md).
- [x] It should also be inserted in the operator decision tree file `doc/decision-tree-widget/tree.yml`
~You may need to update `MIGRATION.md` if the operator differs from the corresponding one in RxJS v4~

**Description:**

This PR adds `switchScan` operator as described here #2931 and reopened by @benlesh some time ago. I was told on Slack that it might get refused but I'm ok with that (no harsh feelings :)).

It's internally just using `switchMap` and `tap` because I first started implementing it myself but then realised I'm doing almost the same thing as `switchMap` is doing already so I just used combination of those two operators (I think it's been recommended to rather reuse existing operators than creating them from scratch). Tests are a combination of existing tests for `scan` and `mergeScan`.

I made it to work similarly to `scan` because that's more close to `Array.reduce()` even though `mergeScan` behaves a little differently. I tried to describe it here https://github.com/ReactiveX/rxjs/issues/4348#issuecomment-450561459 and here #4441 (the way it handles the first emission from source without seed and that it passes `index` to the accumulator function).

I also didn't add this operator to `rxjs-compat` because it didn't exist previously.

Closes #2931


benlesh:
Very interesting, @martinsik... I know this is something @trxcllnt wanted to add back in the day. I've marked this for core team discussion. Thank you!

src/internal/operators/switchScan.ts
```diff
@@ -0,0 +1,63 @@
...
+  // The same reason as described in `scan` https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/scan.ts#L54-L58
+  if (arguments.length >= 2) {
+    hasSeed = true;
+  }
+
+  return (source: Observable<T>) => source.lift(new SwitchScanOperator(accumulator, seed, hasSeed));
+}
+
+class SwitchScanOperator<T, R> implements Operator<T, R> {
+  constructor(private accumulator: (acc: R, value: T, index: number) => ObservableInput<R>, private seed: R, private hasSeed: boolean = false) { }
```
cartant: The seed is mutated, so it cannot be stored in a property that belongs to the operator instance. The operator instance forms part of the composed observable chain. Stored here, the seed will not be per-subscriber, so the subscribers will interfere with one another. The `source.pipe` call could be made within a `defer` call and the seed could be stored within the callback passed to `defer`.

src/internal/operators/switchScan.ts
```diff
@@ -0,0 +1,63 @@
+import { Observable } from '../Observable';
+import { ObservableInput, OperatorFunction, MonoTypeOperatorFunction } from '../types';
+import { Operator } from '../Operator';
+import { Subscriber } from '../Subscriber';
+import { of } from '../observable/of';
+import { switchMap } from './switchMap';
+import { tap } from './tap';
+
+export function switchScan<T>(accumulator: (acc: T, value: T, index: number) => ObservableInput<T>, seed?: T): MonoTypeOperatorFunction<T>;
+export function switchScan<T, R>(accumulator: (acc: R, value: T, index: number) => ObservableInput<R>, seed?: R): OperatorFunction<T, R>;
```
cartant: I'm not sure why the seed is optional here, when it's not optional in `mergeScan`. Is there a particular reason for this? Also, the overload signatures here seem to have repeated a problem that the `scan` signatures have, IIRC.

If the accumulator is something other than `T`, a seed will be required, as without it, there will be no initial, non-`T` value that can be passed.

cartant (dismissed):
This isn't a complete review. I'm just pointing out some things I've noticed whilst perusing the changes. I'll have a closer look at them later.

src/internal/operators/switchScan.ts
martinsik: Yeah, you're right. It really was overriding the same seed among multiple subscribers. I think I don't even need `defer` and just use scoped variables. I fixed it and added a test that would fail otherwise.

src/internal/operators/switchScan.ts
martinsik: I actually don't know what is the expected behavior here. I mentioned here https://github.com/ReactiveX/rxjs/issues/4348#issuecomment-450561459 that this is different even between `scan` and `mergeScan` so I made it closer to [`Array.reduce` that allows no seed value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce#Syntax) as well.

src/internal/operators/switchScan.ts
cartant: Yeah, `call` is called per-subscriber, so shoving it in there as a variable is fine.

src/internal/operators/switchScan.ts
martinsik: I understand now. If the `seed` param is mandatory then the `accumulator` function will always be called so this `T` and `R` mismatch won't happen.

cartant: Yeah. There are a number of ways in which it could be resolved - including using `T | R` for the accumulator in the `R` signature without a seed and using `R` for the accumulator in a separate `R` signature with a seed. We should really have a look at the typings for `Array.prototype.reduce` that are within the TypeScript lib to see what's done there.

cartant: https://github.com/Microsoft/TypeScript/blob/v3.3.1/lib/lib.es5.d.ts#L1168-L1180

spec/operators/switchScan-spec.ts
```diff
@@ -0,0 +1,401 @@
...
+describe('switchScan', () => {
+  asDiagram('switchScan(i => 10*i\u2014\u201410*i\u2014\u201410*i\u2014|, 0)')
+  ('should map-and-flatten each item to an Observable while passing the seed', () => {
+    const e1 =    hot('--1-----3--5-------|');
+    const e1subs =    '^                  !';
+    const e2 =   cold('x-x-x|              ', {x: 10});
+    const expected =  '--x-x-x-y-yz-z-z---|';
+    const values = {x: 10, y: 40, z: 90};
+
+    const result = e1.pipe(switchScan((acc: number, x: string) => e2.pipe(map(i => i * Number(x) + acc)), 0));
```
benlesh: Can we remove the explicit typings here?

spec/operators/switchScan-spec.ts
```diff
@@ -0,0 +1,401 @@
...
+  it('should switch inner cold observables', () => {
+    const x =   cold(         '--a--b--c--d--e--|           ');
+    const xsubs =    '         ^         !                  ';
+    const y =   cold(                   '---f---g---h---i--|');
+    const ysubs =    '                   ^                 !';
+    const e1 =   hot('---------x---------y---------|        ');
+    const e1subs =   '^                            !        ';
+    const expected = '-----------a--b--c----f---g---h---i--|';
+
+    const observableLookup = { x: x, y: y };
```
benlesh: Just `{ x, y }` here should be fine. (Same throughout the file)

spec-dtslint/operators/switchScan-spec.ts
```diff
@@ -0,0 +1,38 @@
...
+it('should infer correctly when using a single type', () => {
+  const o = of(1, 2, 3).pipe(switchScan((acc, v) => of(acc + v), 0)); // $ExpectType Observable<number>
+});
+
+it('should infer correctly when using a seed', () => {
+  const o = of(1, 2, 3).pipe(switchScan((acc, v) => of(acc + v), 0)); // $ExpectType Observable<number>
+});
+
+it('should infer correctly when using seed of a different type', () => {
+  const o = of(1, 2, 3).pipe(switchScan((acc: string, v: number) => of(acc + v), '0')); // $ExpectType Observable<string>
```
benlesh: Explicit typings here shouldn't be necessary.

spec-dtslint/operators/switchScan-spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  const o = of(1, 2, 3).pipe(switchScan(p => p)); // $ExpectError
+});
+
+it('should enforce seed and the return type from accumulator', () => {
+  const o = of(1, 2, 3).pipe(switchScan(p => of(1), [])); // $ExpectError
+});
+
+it('should enforce seed and accumulator to have the same type', () => {
+  const o = of(1, 2, 3).pipe(switchScan((acc, p) => of([...acc, p]))); // $ExpectError
+});
```
benlesh: Add some tests for when `accumulator` returns a union type like `Observable<string> | Observable<number>`

For example:

```ts
const o = of(1, 2, 3).pipe(switchScan((acc, p) => Math.random() > 0.5 ? of('abc') : of(1232)); // $ExpectType Observable<string | number>
```

benlesh: `switchMap`, et al, have examples of this.

benlesh (dismissed):
Apologies for the delayed review. Let's get this in!

doc/decision-tree-widget/tree.yml
```diff
@@ -117,6 +117,9 @@ children:
       - label: and output the computed values as a nested Observable when the source emits a value
         children:
         - label: mergeScan
+      - label: and output the computed values as a nested Observable when the source emits a value and cancel the previous nested Observable
```
cartant: Since you opened this, the decision tree widget has moved to `docs_app\tools\decision-tree-generator\src\tree.yml`, so this change should be applied to that file.

spec-dtslint/operators/switchScan-spec.ts
```diff
@@ -0,0 +1,38 @@
+import { of } from 'rxjs';
+import { switchScan } from 'rxjs/operators';
+
+it('should infer correctly', () => {
+  const o = of(1, 2, 3).pipe(switchScan((acc: boolean, v: number) => of(Boolean(v)), false)); // $ExpectType Observable<boolean>
+});
+
+it('should infer correctly when using a single type', () => {
+  const o = of(1, 2, 3).pipe(switchScan((acc, v) => of(acc + v), 0)); // $ExpectType Observable<number>
```
cartant: This looks identical to the test below in which a seed is specified.

martinsik:
@benlesh @cartant Hey, I updated this PR with everything you mentioned above.

BioPhoton:
I stumble over this open PR and see that all requested changes are implemented.
The failed build on node v9 is related to some non-code related things which I guess could be ignored.
@benlesh may I kindly ask you for a status update to this PR= :)

tonivj5:
Any update on this PR? :+1:

martinsik:
Rebased to master and fixed some types in spec.

cartant:
Thanks. I'll re-review it.

benlesh:
Oh man... this old thing. Sigh. Casualty of burnout mostly. I'll mark it as an agenda item and see if the team still wants to land it, or if it would be better in rxjs-misc.

cc @cartant

benlesh:
Resurrected! 🧟

benlesh (approved):

spec/operators/switchScan-spec.ts
```diff
@@ -0,0 +1,398 @@
...
+    const expected =  '--x-x-x-y-yz-z-z---|';
+    const values = {x: 10, y: 40, z: 90};
+
+    const result = e1.pipe(switchScan((acc, x) => e2.pipe(map(i => i * Number(x) + acc)), 0));
+
+    expectObservable(result).toBe(expected, values);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should not pass seed when undefined', () => {
```
cartant, line 21: nitpick: This test description doesn't make much sense. The seed passed is zero and the expectation tests the accumulator  not the seed.

spec/operators/switchScan-spec.ts
```diff
@@ -0,0 +1,398 @@
...
+
+    const result = e1.pipe(switchScan((acc, value) => observableLookup[value], null));
+
+    expectObservable(result).toBe(expected);
+    expectSubscriptions(x.subscriptions).toBe(xsubs);
+    expectSubscriptions(y.subscriptions).toBe(ysubs);
+    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  });
+
+  it('should handle a synchronous switch to the second inner observable', () => {
```
cartant, line 171: thought: This looks like one of those tests that purports to test something that's synchronous, but isn't really synchronous-within-`subscribe`. Nothing to do here, this is just an observation related to the sync-within-`subscribe` marble syntax that we talked about.

src/internal/operators/switchScan.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Applies an accumulator function over the source Observable where the
+ * accumulator function itself returns an Observable, emitting values
+ * only from the most recently returned Observable.
+ *
+ * <span class="informal">It's like {@link scan}, but only the most recent
+ * Observable returned by the accumulator is merged into the outer Observable.</span>
+ *
+ * ![](switchScan.png)
```
cartant, line 14: question: @niklas-wortmann With the addition of a new operator, is there some config that needs to updated so that a marble diagram for it is generated?

src/internal/operators/switchScan.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @return An observable of the accumulated values.
+ */
+export function switchScan<T, R, O extends ObservableInput<any>>(
+  accumulator: (acc: R, value: T, index: number) => O,
+  seed: R
+): OperatorFunction<T, ObservedValueOf<O>> {
+  return operate((source, subscriber) => {
+    // The state we will keep up to date to pass into our
+    // accumulator function at each new value from the source.
+    let state: any = seed;
```
cartant, line 32: nitpick: Not sure why this has be an `any` when the `seed` is known to be an `R`.

src/internal/operators/switchScan.ts
```diff
@@ -0,0 +1,51 @@
...
+      // our previous state, the value and the index.
+      (value: T, index) => accumulator(state, value, index),
+      // Using the deprecated result selector here as a dirty trick
+      // to update our state with the flattened value.
+      (_, innerValue) => ((state = innerValue), innerValue)
+    )(source).subscribe(subscriber);
+
+    return () => {
+      // Release state on teardown
+      state = null!;
```
cartant, line 48: nitpick: If `state` is going to be `any`, the non-null assertion (`!`) isn't necessary.

cartant (approved):
LGTM, just a bunch of nitpicks.

spec/operators/switchScan-spec.ts
benlesh: Yeah, now that I look at it, the test is overly complicated as well. I've simplified it a bit.

spec/operators/switchScan-spec.ts
benlesh: Yeah, now that I look at it, the test is overly complicated as well. I've simplified it a bit.

src/internal/operators/switchScan.ts
benlesh: fixed

src/internal/operators/switchScan.ts
benlesh: state is now `R`, so it has to stay.

src/internal/operators/switchScan.ts
benlesh: Added a TODO around this.

PR #4444, opened 2018-12-31
fix(delay): emit complete notification as soon as possible

### LINKED ISSUE #4249: Behavior of `delay` inconsistent with completions.
Currently, `delay` will delay the completion notification, but *only if* there is no nexted value (the source is "empty").

This just doesn't make any sense. It should either always delay notification of complete, or it shouldn't.

Personally, I think it should not delay notification of complete, given that it's a composable behavior that way. The other way it's not, it's hard to remove the delay once it's added, because it's more obscured from downstream consumers (outside of some crazy mulitcasting scenario)

martinsik:
**Description:**
This is an attempt to fix #4249. `delay` will now emit complete notification as soon as possible. This means that if there are no values queued it'll emit immediately. If there are queued values then it'll emit complete right after the last scheduled value.

Since complete shouldn't be delayed I'm not even adding it into the queue which means complete might be sent in the same frame as the last value. For this reason I had to change this test https://github.com/ReactiveX/rxjs/compare/master...martinsik:fix-4249?expand=1#diff-dd8d0ea30be1698896e447581ec91fffR379

This is a breaking change although probably not many people rely on this behavior.
Closes #4249

### DIFF
--- a/spec/operators/delay-spec.ts
+++ b/spec/operators/delay-spec.ts
@@ -11,10 +11,10 @@ declare const rxTestScheduler: TestScheduler;
 /** @test {delay} */
 describe('delay operator', () => {
   asDiagram('delay(20)')('should delay by specified timeframe', () => {
-    const e1 =   hot('---a--b--|  ');
-    const t =   time(   '--|      ');
-    const expected = '-----a--b--|';
-    const subs =     '^        !  ';
+    const e1 =   hot('---a--b--|');
+    const t =   time(   '--|    ');
+    const expected = '-----a--b|';
+    const subs =     '^        !';
 
     const result = e1.pipe(delay(t, rxTestScheduler));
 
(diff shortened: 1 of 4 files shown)


benlesh:
Okay... I approve of this one... but it's a *little* special, in that it might be a breaking change for some people. SO, I'm going to wait until I have a chance to run this through google3 to see what the possible public impact will be. It's definitely a bug fix, but there may be people relying on the behavior of this bug.

We also need to provide a workaround for people who want the behavior.

martinsik:
Maybe I could add another test to make sure the previous behavior is still composable:

```
const e1 =   hot('---a--b--|  ');
const t =   time(   '--|      ');
const expected = '-----a--b--|';
const subs =     '^        !  ';

const result = concat(
  e1.pipe(delay(t, rxTestScheduler)),
  of(undefined).pipe(delay(t, rxTestScheduler), ignoreElements()),
);
```

benlesh:
@martinsik ... sorry for the delay in reviewing this one again... If you could please add the additional test you've suggested, I think that would be good.

martinsik:
@benlesh Test added

benlesh:
Looks good. But will have to wait for the next major version, because it's a breaking change.

PR #4446, opened 2019-01-02
fix(WebSocketSubject): fix subject failing to close socket

rsikdar:
**Description:**
Fix WebSocketSubject failing to close the underlying socket connection when it is unsubscribed
before the socket is in the open state. Currently, when unsubscribe is called on the subject or
when a subscription is unsubscribed there are checks on whether the socket is open (readyState = 1)
before it closes the connection. If unsubscribe is called before the connection is open, it will not
close the socket but will reset the state of the subject (_socket set to null), and it becomes
impossible to close the socket.

**Related issue (if exists):** #4447

### DIFF
--- a/spec/observables/dom/webSocket-spec.ts
+++ b/spec/observables/dom/webSocket-spec.ts
@@ -143,6 +143,34 @@ describe('webSocket', () => {
       (<any>socket.close).restore();
     });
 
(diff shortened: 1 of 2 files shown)


benlesh:
Thank you, @rsikdar!

PR #4450, opened 2019-01-03
clear tryCatchTarget to release reference of function for gc

chenxm:
**Description:**


benlesh:
I agree. @cartant, I made the changes, and also added cleaning of the `errorObject.e` reference on subsequent `tryCatcher` calls. Please review.

src/internal/util/tryCatch.ts
```diff
@@ -3,11 +3,15 @@ import { errorObject } from './errorObject';
 let tryCatchTarget: Function;
 
 function tryCatcher(this: any): any {
+  let result;
```
cartant: With the changes made, `result` is no longer needed.

PR #4453, opened 2019-01-07
fix(ajax): Fix case-insensitive headers in HTTP request

DostalTomas:
**Description:**
In RFC 7230 and RFC 7540 is written, that HTTP headers is case-insensitive, so this is fix for correct serializing request body based on HTTP headers.

### DIFF
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -600,6 +600,26 @@ describe('ajax', () => {
       expect(MockXMLHttpRequest.mostRecent.data).to.equal('{"🌟":"🚀"}');
     });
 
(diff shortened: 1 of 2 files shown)


DostalTomas:
Thank You for your response.
Test has been added and branch squashed.
I get this issue, when I set `content-type: application/json; charset=UTF-8` header and on server I got URL-encoded data. I know, that this is edge case, but it happens.

```javascript
const reqSettings = {
   url: 'some/url',
   headers: {
      'content-type': 'application/json; charset=UTF-8'
   },
   body: {
      hello: 'world'
   }
}

ajax(reqSettings).subscribe();
```

benlesh:
Thanks, @DostalTomas!

DostalTomas:
Thank You too ;)

PR #4458, opened 2019-01-08
feat(mergeScan): Add index to the accumulator function

### LINKED ISSUE #4441: `mergeScan()`'s accumulator doesn't receive any `index` parameter

martinsik:
**Description:**
This PR adds `index` parameter to the accumulator function used by `mergeScan` just like in most other operators.

```
of('a', 'b', 'c', 'd').pipe(
  mergeScan((acc, x, index) => {
    console.log(index); // prints: 0, 1, 2, 3
    return of(x);
  }, 0)
).subscribe();
```

Closes #4441

### DIFF
--- a/spec-dtslint/operators/mergeScan-spec.ts
+++ b/spec-dtslint/operators/mergeScan-spec.ts
@@ -21,6 +21,10 @@ it('should support a currency', () => {
   const o = of(1, 2, 3).pipe(mergeScan((acc, value) => of(acc + value), '', 47)); // $ExpectType Observable<string>
 });
 
+it('should support an index parameter', () => {
+  const o = of(1, 2, 3).pipe(mergeScan((acc, value, index) => of(index), 0)); // $ExpectType Observable<number>
+});
+
 it('should enforce types', () => {
   const o = of(1, 2, 3).pipe(mergeScan()); // $ExpectError
 });
--- a/spec/operators/mergeScan-spec.ts
+++ b/spec/operators/mergeScan-spec.ts
@@ -419,4 +419,17 @@ describe('mergeScan', () => {
     expectSubscriptions(inner[1].subscriptions).toBe(ysubs);
     expectSubscriptions(inner[2].subscriptions).toBe(zsubs);
   });
+
+  it('should pass current index to accumulator', () => {
+    const recorded: number[] = [];
+    const expected = [0, 1, 2, 3];
+    const e1 = of('a', 'b', 'c', 'd');
+
+    e1.pipe(mergeScan((acc, x, index) => {
+      recorded.push(index);
+      return of(x);
+    }, 0)).subscribe();
+
+    expect(recorded).to.deep.equal(expected);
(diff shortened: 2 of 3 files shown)


benlesh:
@cartant I think the test is fine given it's just asserting the index is being passed to the function.

spec-dtslint/operators/mergeScan-spec.ts
```diff
@@ -21,6 +21,10 @@ it('should support a currency', () => {
   const o = of(1, 2, 3).pipe(mergeScan((acc, value) => of(acc + value), '', 47)); // $ExpectType Observable<string>
 });
 
+it('should support an index parameter', () => {
+  const o = of(1, 2, 3).pipe(mergeScan((acc, value, index) => of(index), 0)); // $ExpectType Observable<number>
+});
```
benlesh: this new test is unnecessary. :)

spec/operators/mergeScan-spec.ts
```diff
@@ -419,4 +419,17 @@ describe('mergeScan', () => {
...
+    const recorded: number[] = [];
+    const expected = [0, 1, 2, 3];
+    const e1 = of('a', 'b', 'c', 'd');
+
+    e1.pipe(mergeScan((acc, x, index) => {
+      recorded.push(index);
+      return of(x);
+    }, 0)).subscribe();
+
+    expect(recorded).to.deep.equal(expected);
```
benlesh: Can you please move expected inside of `equal` here? It will make the test moderately more readable, IMO.

spec-dtslint/operators/mergeScan-spec.ts
cartant: I'm pretty sure this test will fail if the index is removed from the signature, so it guards against a regression, IMO.

martinsik: I removed this test. I was looking at how is this thing tested elsewhere and it seemed to be the same usecase as here https://github.com/ReactiveX/rxjs/blob/master/spec-dtslint/operators/map-spec.ts#L12-L14 or here https://github.com/ReactiveX/rxjs/blob/master/spec-dtslint/operators/findIndex-spec.ts#L8-L10

cartant: Personally, I'd favour the dtslint typings tests being written without regard to what's tested in the specs - even if that means there are so excited redundant tests.

PR #4459, opened 2019-01-09
comparor ->comparer

xp44mm:
replace comparor with comparer

**Description:**

mistake spell comparer to comparor


benlesh:
Thank you, @xp44mm!

PR #4462, opened 2019-01-09
test(dtslint): add dtslint test for race observable static function (…

dkosasih:
**Description:** Add dtslint test for `race` observable static function


benlesh:
Looks like there are some typings errors in `race`... I'll correct those in a separate PR.

dkosasih:
Or you can assign it to me if you want

PR #4463, opened 2019-01-10
docs(operators): variable interval rename to i

xp44mm:
in example the const interval(1000); will result in error. so rename interval to i

**Description:**


benlesh:
Thanks for pointing this one out, @xp44mm!

PR #4465, opened 2019-01-11
fix(typescript): Update typings for race to support proper return types

benlesh:
The types returned by race were incorrect. This ought to resolve that.

### DIFF
--- a/src/internal/observable/race.ts
+++ b/src/internal/observable/race.ts
@@ -16,9 +16,9 @@ export function race<A, B, C, D>(a: Observable<A>, b: Observable<B>, c: Observab
 export function race<A, B, C, D, E>(a: Observable<A>, b: Observable<B>, c: Observable<C>, d: Observable<D>, e: Observable<E>): Observable<A> | Observable<B> | Observable<C> | Observable<D> | Observable<E>;
 // tslint:enable:max-line-length
 
-export function race<T>(observables: Array<Observable<T>>): Observable<T>;
-export function race<T>(observables: Array<Observable<any>>): Observable<T>;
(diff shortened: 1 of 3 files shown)


src/internal/observable/race.ts
```diff
@@ -9,9 +9,18 @@ import { OuterSubscriber } from '../OuterSubscriber';
...
+// tslint:disable:max-line-length
+export function race<A, B>(a: Observable<A>, b: Observable<B>): Observable<A> | Observable<B>;
+export function race<A, B, C>(a: Observable<A>, b: Observable<B>, c: Observable<C>): Observable<A> | Observable<B> | Observable<C>;
+export function race<A, B, C, D>(a: Observable<A>, b: Observable<B>, c: Observable<C>, d: Observable<D>): Observable<A> | Observable<B> | Observable<C> | Observable<D>;
+export function race<A, B, C, D, E>(a: Observable<A>, b: Observable<B>, c: Observable<C>, d: Observable<D>, e: Observable<E>): Observable<A> | Observable<B> | Observable<C> | Observable<D> | Observable<E>;
+// tslint:enable:max-line-length
+
 export function race<T>(observables: Array<Observable<T>>): Observable<T>;
 export function race<T>(observables: Array<Observable<any>>): Observable<T>;
 export function race<T>(...observables: Array<Observable<T> | Array<Observable<T>>>): Observable<T>;
```
cartant: I don't understand why
```
export function race<T>(...observables: Array<Observable<T> | Array<Observable<T>>>): Observable<T>;
```
needs to be an overload signature. I can see why it is the implementation signature, but why it needs to be an overload baffles me. Don't the others handle the scenarios that are bundled into this signature? LGTM, otherwise.

PR #4467, opened 2019-01-12
`async` decorator for observable-yielding generators

### LINKED ISSUE #1497: Implement Observable.spawn(function* (){})
It seems that the spawn function is not implemented in RxJS5. As such,  I cannot yield to Observables.
Our use case is to retrieve paginated data from a web-server. This is my how I can do it in RxJS4:

```
Rx.Observable.spawn(function* () {
  const limit = 2;
  try {
    const aprxjs = Rx.Observable.fromCallback(api);
    const lazyArr = [];
    yield aprxjs(0)
      .expand(v => v < limit ? aprxjs(v) : Rx.Observable.empty())
      .do(v => lazyArr.push(v));
  }
  catch (e) { console.log('er', e.stack); }
}).subscribe();
```

thejohnfreeman:
A [longer version][] of the justification for this change is available online. In brief, we're missing the observable equivalent of async functions. Chaining operators like `map`, `filter`, and `catchError` with observables that yield a single value is as painful as chaining callbacks on promises. Async functions let us write asynchronous code in a synchronous style, which is easier to read, write, and comprehend. Short of native syntax like async/await, we can offer a decorator for generator functions.

[longer version]: https://medium.com/@thejohnfreeman/escaping-pipeline-hell-38d962f66d31

Note: A [limitation in TypeScript](https://github.com/Microsoft/TypeScript/issues/2983) prevents us from more accurately constraining the types of iterators/generators.

I started on the path of this contribution before reading the [suggestion](https://github.com/ReactiveX/rxjs/blob/master/doc/operator-creation.md) to create a separate library. My mistake. I figured it's worth a shot anyhow, and I'll make the effort to move it to a separate library if it is rejected here. If it is rejected, I would still appreciate any feedback you can offer. I've left comments in the code asking for specific feedback. I'd love to hear what kind of tests you'd like to see.

Fixes #1497

### DIFF
--- a/spec/observables/async-spec.ts
+++ b/spec/observables/async-spec.ts
@@ -1,0 +1,22 @@
+import { expect } from 'chai';
+import { of, async } from 'rxjs';
+
+// REVIEWER: Before I invest in writing more tests, I want to make sure I'm on
+// the right path to an accepted contribution.
+
+/** @test {async} */
+describe('async', () => {
+  it('should step through yielded observables', (done) => {
+    const f = async(function* (d) {
+      const a = yield ['a'];
+      const b = yield of('b');
+      const c = yield Promise.resolve('c');
+      return a + b + c + d;
+    });
+    const observable = f('d');
+    observable.subscribe((x) => {
+      expect(x).to.equal('abcd');
+      done();
+    });
+  });
+});
--- a/spec/observables/spawn-spec.ts
+++ b/spec/observables/spawn-spec.ts
@@ -1,0 +1,18 @@
+import { expect } from 'chai';
+import { of, spawn } from 'rxjs';
+
+/** @test {spawn} */
+describe('spawn', () => {
+  it('should step through yielded observables', (done) => {
+    const spawned = spawn(function* () {
+      const a = yield ['a'];
+      const b = yield of('b');
+      const c = yield Promise.resolve('c');
+      return a + b + c;
+    });
+    spawned.subscribe((x) => {
+      expect(x).to.equal('abc');
+      done();
+    });
+  });
+});
--- a/src/index.ts
+++ b/src/index.ts
@@ -40,6 +40,7 @@ export { UnsubscriptionError } from './internal/util/UnsubscriptionError';
 export { TimeoutError } from './internal/util/TimeoutError';
 
 /* Static observable creation exports */
+export { async } from './internal/observable/async';
 export { bindCallback } from './internal/observable/bindCallback';
 export { bindNodeCallback } from './internal/observable/bindNodeCallback';
 export { combineLatest } from './internal/observable/combineLatest';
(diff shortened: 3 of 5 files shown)


benlesh:
@thejohnfreeman, this is pretty interesting and a lot to review. It will definitely need to be discussed at the core team meeting.

*My* initial gut reactions:

1. Exporting a function named `async` seems risky since that's a JavaScript keyword.
2. I'm not entirely sure this needs to be a part of the core library, but it is _really_ cool/interesting.
3. I'd like to see more use cases.

kwonoj:
I may not be able to attend core meeting this time so will leave opinions here, while I think other core team might have opinions as well.

In short, I do not strongly feel this would needed in core library. This approach brings interesting approach to subscribe mechanism, but I feel core lib should be sufficient to present de-facto way to subscribe (namely, `subscribe` itself) and other mechanisms can be represented in userlevel and anyone can opt into. Maybe I'm not having full visibility, but I don't see the specific reason it has to be in core lib.

Not related with above but few nits also
- I agree with Ben for naming `async` should be avoided - even in our core lib we have one known issue by using reserved keyword as variable name (`var async`). Using ECMA keyword as user-level is defintely no-go.
- I don't think I agree with term `pipeline hell` in explainer from beginning honestly. I do agree proposed solution provides imperative way to subscribe into observables, but that doesn't mean current pipe, or operator chain creates `hell`.

Lastly super nit
- At a glimpse, I feel this is more suitable IxJS's `asyncIterable` and actually some of source producing just can be done in IxJS instead of Observable trying to create generator? not sure before going through in detail though.

benlesh:
After discussion with the core team, we all agreed that this is a pretty cool thing, but not something we want to support as a feature for RxJS. It's best to have this in a user-land library, and we'd be happy to support you there in whatever capacity we can.

Sorry we aren't letting it into the core library, I understand that may be disappointing, but I'm not kidding that we unanimously thought it was very cool. I strongly encourage you to publish this as a standalone library. Let me know if you do, and I'm happy to promote it.

Thank you so much, @thejohnfreeman, for your hard work and effort with this PR. I deeply appreciate it.

thejohnfreeman:
No problem, I completely understand. I'll definitely be publishing this is a library then. I'll report back when I have it on npm. Thanks for the support!

Atry:
You can also have a look at https://github.com/pelotom/burrido before publishing yours.

thejohnfreeman:
Burrido is a good implementation of a good idea, but two reasons I won't be able to use it:

1. Like `spawn`, it does not let you pass arguments to the generator function. It assumes you have a parameter-less generator function.
2. By using [`immutagen`](https://github.com/pelotom/immutagen), it limits itself to side-effect free functions, which means no effectful requests to backend services, the primary motivation that led me down this road.

I learned a lot today, though. Thanks for the pointer!

thejohnfreeman:
Here it is! [`@thejohnfreeman/await-observable`](https://github.com/thejohnfreeman/await-observable) Feedback and requests welcome.

PR #4471, opened 2019-01-15
test(dtslint): add dtslint test for throwError observable static func…

dkosasih:
**Description:** Add dtslint test for `throwError` observable static function


benlesh:
Thank you @dkosasih

PR #4475, opened 2019-01-16
test(dtslint): add dtslint test for timer observable static function …

dkosasih:
**Description:** Add dtslint test for `timer` observable static function


benlesh:
Thank you, @dkosasih!

PR #4478, opened 2019-01-19
test(dtslint): add dtslint test for iif observable static function (#…

dkosasih:
**Description:** Add dtslint test for `iif` observable static funtion

### DIFF
--- a/spec-dtslint/observables/iif-spec.ts
+++ b/spec-dtslint/observables/iif-spec.ts
@@ -1,0 +1,21 @@
+import { iif, of } from 'rxjs';
+
+it('should accept function as first parameter', () => {
+  // Typescript limitation: cannot infer default value in params
+  // e.g.: const a = iif(() => false, EMPTY, EMPTY) // $ExpectType Observable<never>
+  // However, test below does not infer Observable<never> if no params is passed
+  // although the default value for second and third param is EMPTY observable
+  const a = iif(() => false); // $ExpectType Observable<{}>
+});
+
+it('should infer correctly with 2 parameters', () => {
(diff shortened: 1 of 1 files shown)


spec-dtslint/observables/iif-spec.ts
```diff
@@ -0,0 +1,21 @@
+import { iif, of } from 'rxjs';
+
+it('should accept function as first parameter', () => {
+  // Typescript limitation: cannot infer default value in params
+  // e.g.: const a = iif(() => false, EMPTY, EMPTY) // $ExpectType Observable<never>
+  // However, test below does not infer Observable<never> if no params is passed
+  // although the default value for second and third param is EMPTY observable
+  const a = iif(() => false); // $ExpectType Observable<{}>
```
benlesh: This seems like a bug... we could add a type overload to cover this.

spec-dtslint/observables/iif-spec.ts
```diff
@@ -0,0 +1,21 @@
...
+});
+
+it('should infer correctly with 2 parameters', () => {
+  const a = iif(() => false, of(1)); // $ExpectType Observable<number | {}>
```
benlesh: should be `Observable<number | never>` ... which I think results in `Observable<number>` regardless?

Add a type overload to account for this.

spec-dtslint/observables/iif-spec.ts
dkosasih: The comments should be able to remove when #4494 is merged.

cartant:
Closed in favour of #5176 - which cherry-picks this PR's commit.

Issue #4476, opened 2019-01-16
Converting all examples to Stackblitz?


ladyleet:
### Description Of The Issue

We are potentially thinking about using converting Stackblitz.

#### Benefits: 

Interactive examples. 

#### Negatives: 

Most operators are one or two line examples so it might be overkill. 
If someone edits the example we don't want them to edit the example that everyone is seeing. 
It should not be linked to someone else's Stackblitz.

#### Questions: 

Are they embeddable?
Who has time for this?

#### IF YOU ARE GOING TO PROCEED:

Only do one conversion so we have time to discuss as a team the PR before spending too much time on it!

ladyleet:
@benlesh @JWO719 @cartant anything I missed here?

sulco:
Hey @ladyleet, I believe I can prepare a proposition of the solution.

I'd propose to look for inspiration in the Angular Material docs
(the button opens an example in a new tab): 
![image](https://user-images.githubusercontent.com/1511906/51288191-57152500-19fb-11e9-9cd7-12f2955569bd.png)
https://material.angular.io/components/snack-bar/overview

- it's simple, unintrusive and, as the Material example shows, a pattern somewhat known/used in the community

Have you seen these? What are your thoughts on that?

ladyleet:
@sulco that is so lovely of you! :) @benlesh i know you had a lot of thoughts about this. and @JWO719 as well. can y'all provide some guidance to @sulco ?

Only1MrAnderson:
I am happy to help as well. Just simple like this? https://stackblitz.com/edit/rxjs-zjwhup

ladyleet:
@Only1MrAnderson thank you so much!!! :) wait for their comments. :)

JoshDSommer:
@ladyleet, @Kajmeh  Here is the Async Subject, example, it seems almost too easy. 

https://stackblitz.com/edit/rxjs-uek8lm

I went to stack blitz. created a new RXJS project. then copied the example exactly from the existing demo here: https://rxjs.dev/guide/subject

ladyleet:
@JoshDSommer isn't it lovely that it's so easy??? <3

ladyleet:
@Kajmeh looking forward to your contributions too! :) super stoked folks are excited about helping with this! the core team is so busy, so it seems small, but is sooooooo awesome.

sulco:
[...]

PR #4479, opened 2019-01-20
Stackblitz docs

sulco:
**Description:**

This PR adds a possibility to view code examples in StackBlitz.

- First commit adds a button that displays next to the existing "Copy code snippet" button. It's implemented in the `code.component.ts` (`<aio-code>`) component and uses the same exact code value that is being displayed / available to copy, to generate the configuration for the StackBlitz' SDK call. Therefore the demo which the user runs is generated **on the fly from the sources**.

- Second commit adds JavaScript imports necessary for the code to be actually runnable. 

In one instance (`animationFrameScheduler`) an html needed to be defined as well. The most straightforward way to do that was to add it as a comment (`// html: <some>html code</some>`).
Other solutions would require placing the js & html code in separate files -> keeping them in sync with sources etc. which, for such _uncommon_ (and simple: a line of html) scenarios seems like an overkill.

**Related issue #4476**


docs_app/src/app/custom-elements/code/code.component.ts
```diff
@@ -147,6 +158,16 @@ export class CodeComponent implements OnChanges {
     return this.codeContainer.nativeElement.textContent;
   }
 
+  /** Extracts html placed in the `// html: ` comment in the code.  */
+  private getHtmlFromCode(code: string): string {
+    const pattern = new RegExp('// html: (.*)');
```
gkalpak: Just a note: This assumes there is only one `// html` occurrence in the code and that it does not span multiple lines.

docs_app/src/app/custom-elements/code/code.component.ts
```diff
@@ -161,6 +182,17 @@ export class CodeComponent implements OnChanges {
     }
   }
 
+  editInStackBlitz() {
+    this.stackblitz.openProject({
+      code: this.codeText,
+      language: this.language,
+      dependencies: {
+        rxjs: '6'
```
gkalpak: In the final version, this shouldn't be hard-coded, I assume.

docs_app/src/app/custom-elements/code/code.component.ts
```diff
@@ -38,16 +39,24 @@ const DEFAULT_LINE_NUMS_COUNT = 10;
     <pre class="prettyprint lang-{{language}}">
       <button *ngIf="!hideCopy" class="material-icons copy-button no-print"
         title="Copy code snippet"
-        [attr.aria-label]="ariaLabel"
+        [attr.aria-label]="ariaLabelCopy"
         (click)="doCopy()">
         <span aria-hidden="true">content_copy</span>
       </button>
+      <button *ngIf="!hideCopy" class="material-icons edit-button no-print"
+        style="right: 0px"
```
gkalpak: This shouldn't be necessary.

docs_app/src/app/custom-elements/code/code.component.ts
gkalpak: I mean, it should be somehow populated (either at build time or at runtime) from the current version.

JWO719:
This is a really awesome PR. Function-wise it's really a joy to use this feature. It's such a big improvement. I totally love it. 
I noticed here and there some compilation errors within the generated examples (syntax errors, missing imports, etc.), but I think we should merge this and improve the examples seperatly because this merge request is already slightly big and the examples used to be "wrong" all the time :D. I'd appreciate if we could implement the things @gkalpak mentioned, besides the html thingy AFAIK there aren't many examples including html code. Most of the time we just print some stuff to the console. Thank you very much @sulco !

sulco:
Hey @benlesh, just letting you know that I've made those requested changes. Cheers!

benlesh:
Thank you, @sulco! This was an awesome PR!

sulco:
Awesome, thanks! Glad to be of help :-)

PR #4483, opened 2019-01-22
docs: annotate deprecated and experimental APIs on list page

abraham:
**Description:**

I frequently find myself selecting deprecated items from the API list page and having to return to the list page to select a non-deprecated API. This marks deprecated APIs with a `<s>` and experimental APIs with `<i>`. (I "created" a fake `mergeJoin` operator to illustrate)

screenshot from 2019-01-21 17-31-15

### DIFF
--- a/docs_app/src/app/custom-elements/api/api-list.component.html
+++ b/docs_app/src/app/custom-elements/api/api-list.component.html
@@ -25,7 +25,10 @@
     <ul class="api-list">
       <ng-container *ngFor="let item of section.items">
         <li *ngIf="item.show" class="api-item">
-          <a [href]="item.path"><span class="symbol {{item.docType}}"></span> {{item.title}}</a>
+          <a [href]="item.path">
+            <span class="symbol {{item.docType}}"></span>
+            <span class="stability {{item.stability}}">{{item.title}} {{!item.stability || item.stability === 'stable' ? '' : '(' + item.stability + ')'}}</span>
+          </a>
         </li>
       </ng-container>
     </ul>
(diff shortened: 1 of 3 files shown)


docs_app/src/app/custom-elements/api/api-list.component.html
```diff
@@ -25,7 +25,14 @@ <h2>{{section.title}}</h2>
...
       <ng-container *ngFor="let item of section.items">
         <li *ngIf="item.show" class="api-item">
-          <a [href]="item.path"><span class="symbol {{item.docType}}"></span> {{item.title}}</a>
+          <a [href]="item.path">
+            <span class="symbol {{item.docType}}"></span>
+            <span [ngSwitch]="item.stability">
+              <s *ngSwitchCase="'deprecated'" title="deprecated">{{item.title}}</s>
+              <i *ngSwitchCase="'experimental'" title="experimental">{{item.title}}</i>
+              <span *ngSwitchDefault title="stable">{{item.title}}</span>
+            </span>
```
benlesh: This is a great idea!

I couple of things:

1. To make this more accessible, can we just add the text "(deprecated)" or "(experimental)" next to the name? e.g. `mergeJoin (experimental)` or `merge (deprecated)`. But keep the styling, I like it.
2. Can we do this with a classes instead of markup? It should be a little less overhead, and we're already doing that for the symbol icon.

benlesh:
Hi, @abraham, just letting you know that I haven't forgotten about this awesome PR, and I was hoping that you'd have time to address the changes I requested soon so I can get it in the next release. :)

abraham:
Updated the implementation. `stabiliity: ''` (empty string) is the stable value for the API response but it looks like the disabled status filter  is going to rely on on `stabiliity: 'stable'` I display the stability after the API name if it's present and not `stable`.

screenshot from 2019-01-29 22-03-27

PR #4485, opened 2019-01-22
test(dtslint): add dtslint test for merge observable static function …

dkosasih:
**Description:** Add dtslint test for `merge` observable static function


dkosasih:
It looks like Ben's PR already included a dtslint test for `merge`. Should I abandon this one?

Issue #4427, opened 2018-12-17
rxjs(umd file) will pollute global/window variable


Aqours:
## Bug Report

**Current Behavior**
UMD rxjs file will inject global variable `$jscomp` to global/window.

**Reproduction**

test code

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Variable Leaked</title>
    <style>
        #app {
            margin: 2em;
            padding: 1em;
            font-family: "Roboto Mono", "Source Code Pro", monospace;
            font-weight: 500;
            font-size: 1.125rem;
            line-height: 1.6;
        }
    </style>
</head>
<body>

<div id="app"></div>

<script>
    document.a = Object.keys(window);
</script>

<script src="https://unpkg.com/rxjs@6.3.3/bundles/rxjs.umd.min.js"></script>

<script>
    document.b = Object.keys(window);
    document.c = {};
    document.d = [];
    document.a.forEach(x => {
        document.c[x] = true;
    });
    document.b.forEach(x => {
        if (!document.c[x]) {
            document.d.push(x);
        }
    });

    (function(leaked, fragment) {
        console.log(leaked);
        leaked.reduce((p, c) => {
            p.appendChild(document.createTextNode(c));
            p.appendChild(document.createElement('br'));
            return p;
        }, fragment);
        self.app.appendChild(fragment);
    })(document.d, document.createDocumentFragment());
</script>

</body>
</html>

```

**Expected behavior**
UMD rxjs file will only inject global variable `rxjs` to global/window.

**Environment**
- Runtime: n/a
- RxJS version: 6.3.3
- (If bug is related) Loader, build configuration: [e.g webpack, angular-cli version, config]

**Possible Solution**

Using `--isolation_mode="IIFE"` or `--output_wrapper="(function(){%output%)).call(this);"` flag to Google Closure Compiler which will not preserve previous IIFE. (See [Flags-and-Options](https://github.com/google/closure-compiler/wiki/Flags-and-Options))

**Additional context/Screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.

PR #4487, opened 2019-01-23
fix(bundle): closure to not rewrite polyfills for minification

imcotton:
**Description:**
Sorry I've never been using umd bundle in a ES5 environment w/o external polyfill (i.e. `core-js`) already existed, am I right?
 
fixes: #4427

### DIFF
--- a/tools/make-closure.js
+++ b/tools/make-closure.js
@@ -9,6 +9,7 @@ module.exports = function makeClosure(sourcePath) {
     jsCode: [{src: source}],
     languageIn: 'ES2015',
     createSourceMap: true,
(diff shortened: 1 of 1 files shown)


kwonoj:
will this removes global variable exports? flag description just says about how does closure handles polyfills.

imcotton:
I think it won't, as mentioned [here](https://github.com/google/closure-compiler/issues/3104#issuecomment-434939549), this flag currently does no _rewritten_ as it should be.

Same changing goes in facebook/react#12618.

benlesh:
Thank you @imcotton!

PR #4489, opened 2019-01-24
perf(internal): optimize Subscription#add() for the common case

bmeurer:
This removes unnecessary checks on the `teardown` parameter upfront, and
instead moves these checks into the appropriate branches guarded by the
type check. This way we get the same behavior, but with fewer checks and
smaller code size.

Also optimize adding the child `subscription` to `this._subscriptions`
for the common case that there were no previous subscriptions.

### DIFF
--- a/src/internal/Subscription.ts
+++ b/src/internal/Subscription.ts
@@ -143,21 +143,13 @@ export class Subscription implements SubscriptionLike {
    * list.
    */
   add(teardown: TeardownLogic): Subscription {
-    if (!teardown || (teardown === Subscription.EMPTY)) {
-      return Subscription.EMPTY;
-    }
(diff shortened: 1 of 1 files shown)


src/internal/Subscription.ts
```diff
@@ -168,13 +160,21 @@ export class Subscription implements SubscriptionLike {
           subscription._subscriptions = [tmp];
         }
         break;
-      default:
+      default: {
+        if (!(<any>teardown)) {
```
bmeurer: Is this just leftover defensive programming? Or is this on purpose?

benlesh: Seems like it's leftover.

PR #4491, opened 2019-01-24
fix(throttleTime): fix trailing

devel112:
**Description:**
Emitting a trailing value should be counted as emitting a value.

[Commit that introduced bug](https://github.com/ReactiveX/rxjs/commit/bb0738f3e6df643158c4a7d1e38f035978b42882)

[#2859](https://github.com/ReactiveX/rxjs/issues/2859)


devel112:
Having trouble making a test for { leading: true, trailing true } because an active task needs to be flushed on _next, i think. Maybe the scheduled task is queued for execution after _next in the tests.

MatthiasKunnen:
This PR can be closed since the issue has been addressed in PR #4564.

PR #4492, opened 2019-01-24
refactor: remove uses of `tryCatch` helper

bmeurer:
The `tryCatch` helper is a leftover from the CrankshaftScript days and
is considered bad practice for modern JavaScript engines. Same for the
`errorObject` trick.

### DIFF
--- a/src/internal/Subscription.ts
+++ b/src/internal/Subscription.ts
@@ -1,8 +1,6 @@
 import { isArray } from './util/isArray';
 import { isObject } from './util/isObject';
 import { isFunction } from './util/isFunction';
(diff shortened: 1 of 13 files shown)


cartant:
If this PR removes 'most' of the uses, can you list those that are not yet removed? how many are there and what's the reason for not removing them all and removing the `tryCatch.ts` and `errorObject.ts` files, too?

bmeurer:
@benlesh Interesting, was wondering a bit about that already. I updated the PR, please take another look.

benlesh:
Awesome! Thanks again, @bmeurer!

PR #4493, opened 2019-01-24
perf(internal): use strict equality for isObject()

bmeurer:
In `isObject(x)` use triple equal instead of double equal comparison
with `null`, since `undefined` will be ruled out via the `typeof`, and
therefore it's sufficient to just check for `null` initially.


benlesh:
Thank you @bmeurer!

PR #4496, opened 2019-01-25
fix(fromEventPattern): improve typings for fromEventPattern

evertbouw:
The handler in fromEventPattern is typed as a Function, meaning it takes no arguments. Overwriting this requires a pretty awkward cast. By applying the same change as in https://github.com/ReactiveX/rxjs/pull/3530 this becomes a lot easier.


benlesh:
Thank you, @evertbouw!

Issue #4513, opened 2019-01-30
NotificationKind not exported


cartant:
## Bug Report

**Current Behavior**

6.4.0 introduced a const enum: [`NotificationKind`](https://github.com/ReactiveX/rxjs/blob/6.4.0/src/internal/Notification.ts#L7-L11). However, that enum is [not exported](https://github.com/ReactiveX/rxjs/blob/6.4.0/src/index.ts#L26-L27):

```ts
/* Notification */
export { Notification } from './internal/Notification';
```

That means that it's not possible to create a `Notification` instance (without resorting to an `as any` assertion).

This was a breaking change for me, as the tests for `rxjs-spy` need to be able to create `Notification` instances. (It's also a breaking change because current calls to the constructor - like `new Notification("N", "alice")` - are now invalid.)

PR #4514, opened 2019-01-30
fix(index): export NotificationKind

cartant:
**Description:**

Exports `NotificationKind`.

**Related issue (if exists):** #4513


cartant:
I've had to update a dtslint spec. It seems that a version must have been updated somewhere, as one of the inferred unions is now in an unexpected order.

Hopefully, this is something that won't happen often. Otherwise, the tests will be pretty fragile.

PR #4516, opened 2019-01-30
test(from): comment out test that dtslint is confused about

benlesh:
There's one dtslint test that is failing no matter what you do, and it really shouldn't be failing. This just comments it out until it's resolved upstream (hopefully).

### DIFF
--- a/spec-dtslint/observables/from-spec.ts
+++ b/spec-dtslint/observables/from-spec.ts
@@ -49,7 +49,7 @@ it('should accept an array of Observables', () => {
 //     yield 42;
 //   }());
 
-//   const o = from([of(1), ['test'], iterable]); // $ExpectType Observable<IterableIterator<number> | Observable<number> | string[]>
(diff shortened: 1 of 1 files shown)


spec-dtslint/observables/from-spec.ts
```diff
@@ -36,13 +36,21 @@ it('should accept an array of Observables', () => {
...
+// and if you switch it to that, it wants
+// Observable<Observable<number> | IterableIterator<number> | string[]>
+// and vica versa.
+
+// it('should accept an array of Inputs', () => {
+//   const iterable = (function*() {
+//     yield 42;
+//   }());
+
+//   const o = from([of(1), ['test'], iterable]); // $ExpectType Observable<IterableIterator<number> | Observable<number> | string[]>
```
cartant: IIRC, dtslint will still see `$ExpectType` and will complain about not being able to find a missing node. You have to change it to `$_TODO_ExpectType` or something.

cartant: If different TS versions report the union type in different orders, this is going to be really annoying/painful. There's a related dtslint PR (https://github.com/Microsoft/dtslint/pull/61) that was abandoned.

Issue #4494, opened 2019-01-25
iif Observable static function does not infer EMPTY Observable when no parameters is passed


dkosasih:
## Bug Report

**Current Behavior**
`const a = iif(()=> true) `
The above code infer `Observable<{}>`

**Expected behavior**
The sample code in Current Behaviour section should infer `Observable<never>`

**Environment**
- RxJS version: 6.3.3

OliverJAsh:
Just ran into this! The PRs for it seem to have stalled. What's currently blocking the fix? Anything I can help with?

cartant:
@OliverJAsh I would imagine they were stalled because the changes and the tests were in separate PRs. I've combined them: #5176

PR #4520, opened 2019-01-31
chore(typings): fix iif observable typings for optional parameters (#…

dkosasih:
**Description:**
Fix typings for `iif` observable inference error when optional parameters are not supplied.

**Related issue (if exists):** #4494

### DIFF
--- a/src/internal/observable/iif.ts
+++ b/src/internal/observable/iif.ts
@@ -90,7 +90,10 @@ import { SubscribableOrPromise } from '../types';
  * @static true
  * @name iif
  * @owner Observable
- */
+*/
+/* tslint:disable:max-line-length */
+export function iif<T = never, F = never>(condition: () => boolean, trueResult?: SubscribableOrPromise<T>, falseResult?: SubscribableOrPromise<F>): Observable<T | F>;
+/* tslint:enable:max-line-length */
 export function iif<T, F>(
   condition: () => boolean,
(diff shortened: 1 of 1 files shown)


src/internal/observable/iif.ts
```diff
@@ -90,7 +94,7 @@ import { SubscribableOrPromise } from '../types';
  * @static true
  * @name iif
  * @owner Observable
- */
+*/
 export function iif<T, F>(
```
benlesh: seems like you could just put the default generics down here, and you wouldn't need the line up top.

src/internal/observable/iif.ts
```diff
@@ -90,7 +90,10 @@ import { SubscribableOrPromise } from '../types';
  * @static true
  * @name iif
  * @owner Observable
- */
+*/
+/* tslint:disable:max-line-length */
+export function iif<T = never, F = never>(condition: () => boolean, trueResult?: SubscribableOrPromise<T>, falseResult?: SubscribableOrPromise<F>): Observable<T | F>;
```
dkosasih: Is this OK?

benlesh: Honestly, we should probably only make these changes coupled with the dtslint unit tests that clearly show the problem it's solving, and that the tests pass in CI.

cartant:
Closed in favour of #5176 - which cherry-picks this PR's commit.

PR #4526, opened 2019-02-01
perf(Subscription): improve parent management

bmeurer:
Merge the `_parent` and `_parents` fields on the `Subscription` class
into a single `_parentOrParents` field, which can hold either `null`
(in case of no parents), a `Subscription` instance (in case of a
single parent), or an array of `Subscription`s (in case of multiple
parents). This not only shrinks the size of `Subscription` (and
subclass) instances, but more importantly reduces the number of
megamorphic property access sites in the hot `Subscription` methods,
especially in `unsubscribe()` and `add()`.

Also inline the `Subscription#_addParent()` method into
`Subscription#add()`, as it's the only call site anyways and
this removes another hot megamorphic `LoadIC` that was only needed
to lookup `subscription._addParent` in the call site.

Finally remove the `hasErrors` variable from `unsubscribe()` method
and instead just check `errors` variable directly.

### DIFF
--- a/src/internal/Subscriber.ts
+++ b/src/internal/Subscriber.ts
@@ -151,14 +151,12 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
 
   /** @deprecated This is an internal implementation detail, do not use. */
   _unsubscribeAndRecycle(): Subscriber<T> {
-    const { _parent, _parents } = this;
-    this._parent = null;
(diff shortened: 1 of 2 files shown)


benlesh:
Thanks again, @bmeurer! I really appreciate the work you're doing.

bmeurer:
Thanks 👍

PR #4527, opened 2019-02-01
chore(typings): add mapped tuple return type to combineLatest

Goobles:
**Description:**
This makes it possible to have type safety for infinite amount of input arguments without having to define every single case.
Makes use of typescript 3.1 new "Mapped types on tuples and arrays" feature.
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html


src/internal/observable/combineLatest.ts
```diff
@@ -11,6 +11,8 @@ import { fromArray } from './fromArray';
 
 const NONE = {};
 
+type ObservableOutputs<T> = { [K in keyof T]: ObservedValueOf<T[K]> };
```
benlesh: Move this to `types.ts`.

benlesh:
This is a solid add, but it'll be blocked on getting the library updated to TS 3.2 and using `typesVersions` to output types for TS 2.7, 2.8, etc.

src/internal/types.ts
```diff
@@ -99,3 +99,5 @@ export interface SchedulerAction<T> extends Subscription {
 }
 
 export type ObservedValueOf<O> = O extends ObservableInput<infer T> ? T : never;
+
+export type ObservableOutputs<T> = { [K in keyof T]: ObservedValueOf<T[K]> };
```
benlesh:
```suggestion
export type ObservedValueLookup<T> = { [K in keyof T]: ObservedValueOf<T[K]> };
```

cartant:
Closing this in favour of #5847. Sorry, @Goobles but with the variadic tuples feature introduced with TS 4 we are going with the approach outlined in #5857. Thanks for the work you put into this PR.

Issue #4502, opened 2019-01-29
of Observable cannot accept more than 9 parameters


dkosasih:
## Bug Report

**Current Behavior**
Currently, `of` observable more than 9 parameters would throw error.

```js
const a = of(1, 2, 3, 4, 5, 6, 7, 8, 9, 0); // $ExpectType Observable<{}>
```

**Expected behavior**
The above code should infer `Observable<{}>`, instead, it throws error.

**Environment**
- RxJS version: 6.3.3
- (If bug is related) Loader, build configuration: [e.g webpack, angular-cli version, config]

benlesh:
This has been resolved for a bit.

PR #4445, opened 2019-01-01
test(dtslint): add dtslint test for of observable static function

dkosasih:
**Description:** Add dtslint test for `of` observable static function

**Related issue (if exists):** #4093

PR #4528, opened 2019-02-01
chore(typings): fix of observable static function inference for 9+ pa…

dkosasih:
**Description:** Fix typings for of observable static function so that it infer empty interface instead of error

**Related issue (if exists):** #4502

### DIFF
--- a/src/internal/observable/of.ts
+++ b/src/internal/observable/of.ts
@@ -19,6 +19,7 @@ export function of<T, T2, T3, T4, T5, T6, T7, T8>(a: T, b: T2, c: T3, d: T4, e: 
 export function of<T, T2, T3, T4, T5, T6, T7, T8, T9>(a: T, b: T2, c: T3, d: T4, e: T5, f: T6, g: T7, h: T8, i: T9, scheduler?: SchedulerLike):
   Observable<T | T2 | T3 | T4 | T5 | T6 | T7 | T8 | T9>;
 export function of<T>(...args: Array<T | SchedulerLike>): Observable<T>;
+export function of<T>(...args: Array<unknown | SchedulerLike>): Observable<T>;
 /* tslint:enable:max-line-length */
 
 /**


src/internal/observable/of.ts
```diff
@@ -19,6 +19,7 @@ export function of<T, T2, T3, T4, T5, T6, T7, T8>(a: T, b: T2, c: T3, d: T4, e:
 export function of<T, T2, T3, T4, T5, T6, T7, T8, T9>(a: T, b: T2, c: T3, d: T4, e: T5, f: T6, g: T7, h: T8, i: T9, scheduler?: SchedulerLike):
   Observable<T | T2 | T3 | T4 | T5 | T6 | T7 | T8 | T9>;
 export function of<T>(...args: Array<T | SchedulerLike>): Observable<T>;
+export function of<T>(...args: Array<unknown | SchedulerLike>): Observable<T>;
```
benlesh: We would need a dtslint test to cover this.

But this *looks* like you're trying to get the user to force a generic in the signature like so:  `of<number>(1,2,3,4,5,6,7,'whatever',9,10,11)`, that's an antipattern, really,  I think we should just return `Observable<{}>` or the like, and force the user to cast themselves. `Observable<unknown>` might be better.

dkosasih: Sorry @benlesh , I don't quite follow. 
what I try to solve here is to fix [this test](https://github.com/ReactiveX/rxjs/pull/4445/files#diff-d0ec8bfb3126e86588f3e57115b9a6feR56) that I commented out because it does not behave as it should be

benlesh:
Looks like we won't be able to merge this until we start RxJS 7 and we're targeting TS 3.0+. 🤔

BioPhoton:
@benlesh as we are now on typescript@^3.0.1 we can either consider going on with this PR 
or close it as well as the related issue.

benlesh:
This has been taken care of for some time now. Thanks for your contribution, sorry this one didn't land!

Issue #2193, opened 2016-12-13
Documentation (in particular Observable.webSocket)


timjacobi:
Hi Guys,

First of all congrats on releasing the final version, I appreciate the work everyone has put into this project as I've been closely following. Now that the final version is released I was hoping to find some more complete API documentation. In this case I was looking to create an observable from web sockets. 

First I went to the [API docs](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-webSocket) which tell me I need to pass `urlConfigOrSource` but I could not figure out how to set it up, am I missing something? 
Moving on `Observable.webSocket` returns a `WebSocketSubject` so I tried to find out how to use it. All the API docs tell me is _We need this JSDoc comment for affecting ESDoc._

Next I attempted to find some docs in this repo so I went [here](https://github.com/ReactiveX/rxjs/blob/master/doc/index.md) and clicked on [Full Reference](» Full reference) but got a 404.

I would really like to fully start using RxJS 5 when we build our next product but the lack of documentation prevents us from doing so. If we were to use it then I'm sure we would also contribute if we find bugs as we go along.

Am I missing an important piece here?
Are there any examples elsewhere?
How can I get involved to help improving the docs?

Again thanks for this great library.

kwonoj:
Yes, while it is released version it still lacks some of the documentation in detail, need to be improved over time. If you're using TypeScript in codebases, most of type comes with the inference that you can get it immediately (i.e `config` interface you mentioned: https://github.com/ReactiveX/rxjs/blob/master/src/observable/dom/WebSocketSubject.ts#L13) but that isn't applicable for all cases since it's TypeScript only.

For now, if you find some of docs are missing and looking for concise examples of code snippet, unit test in repo provides immediate workarounds. For example, you can find several approach of constructing websocketsubject in test case (https://github.com/ReactiveX/rxjs/blob/master/spec/observables/dom/webSocket-spec.ts#L226) with its usages. 

It is hugely appreciated to have contribution to improve document - first you can check general contribution guideline here (https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md) and use other documented codebase as reference (https://github.com/ReactiveX/rxjs/blob/master/src/operator/filter.ts for example).

blesh:
The bottom line here is that we need to add documentation around WebSocketSubject. Thanks @timjacobi for bringing this back to our attention. I think we should prioritize documentation as well.

[...]

PR #4529, opened 2019-02-02
docs(webSocket): added more info for multiplex() method

inf3cti0n95:
**Description:**
Removed ambiguous explaination of the messageFilter function of the multiplex() method.

#2193


benlesh:
Thank you, @inf3cti0n95!

PR #4543, opened 2019-02-08
docs(from): update examples

### LINKED ISSUE #4542: `from` examples

[from](https://rxjs.dev/api/index/function/from)

- [ ] typo
- [x] error(s) in example
- [ ] needs example

### Description Of The Issue

I was looking at the docs about `from` and noticed that the examples can be improved a bit. 

E.g. `from` is imported from `rxjs/observable/from`.

I thought of making a PR but then noticed that in the third example, `async` is imported from a weird place. Is this still correct?

```js
import { async } from 'rxjs/scheduler/async';
```

Shouldn't it be just `import { async } from 'rxjs';`? 

But then looking at the docs about `async`, I noticed that the examples there import `asyncScheduler` and use it in a different way? This confused me a bit.

So I'm up for making a PR but can you please clarify the third example for me.

dzhavat:
**Description:**
Update examples so they work standalone.

Closes #4542

### DIFF
--- a/src/internal/observable/from.ts
+++ b/src/internal/observable/from.ts
@@ -26,9 +26,11 @@ export function from<O extends ObservableInput<any>>(input: O, scheduler?: Sched
  * converted through this operator.
  *
  * ## Examples
+ * 
  * ### Converts an array to an Observable
+ * 
  * ```javascript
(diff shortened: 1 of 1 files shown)


benlesh:
Thank you, @dzhavat!

PR #4548, opened 2019-02-09
docs(windowWhen): fix example

tanem:
Fixes the `windowWhen` example imports.


benlesh:
Thanks again, @tanem!

PR #4549, opened 2019-02-09
docs(windowToggle): fix example

tanem:
The example was missing an import for `empty()`, but since that function is deprecated I figured it should be using `EMPTY` anyway? 🤔


JWO719:
Thanks a lot for all the fixes. very appreciated!

tanem:
No problem @JWO719, it's helping me get more familiar with the API too :eyes:

benlesh:
Awesome, thanks @tanem!

PR #4551, opened 2019-02-09
docs(buffer): fix example

tanem:
Fixes a duplicate declaration error in the `buffer` example.


benlesh:
Thanks again, @tanem!

PR #4553, opened 2019-02-09
Tidy style & docs in catchError

tanem:
Keeping things consistent with the other files I've been looking at 💄


benlesh:
Thank you, @tanem

PR #4554, opened 2019-02-10
docs(groupBy): fix examples

tanem:
Fixes a few glitches so the examples are runnable.

This also includes stripping the type definitions since the example code blocks specify `javascript`, and the other examples I've seen so far don't seem to include these.

### DIFF
--- a/src/internal/operators/groupBy.ts
+++ b/src/internal/operators/groupBy.ts
@@ -38,12 +38,7 @@ export function groupBy<T, K, R>(keySelector: (value: T) => K, elementSelector?:
  * import { of } from 'rxjs';
  * import { mergeMap, groupBy, reduce } from 'rxjs/operators';
  *
- * interface Obj {
- *    id: number,
- *    name: string,
- * }
- *
- * of<Obj>(
+ * of(
  *   {id: 1, name: 'JavaScript'},
  *   {id: 2, name: 'Parcel'},
  *   {id: 2, name: 'webpack'},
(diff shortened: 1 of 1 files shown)


tanem:
Oops, just realised #4554 addresses some of these issues but not all. Happy to re-jig my changes with that other PR once it's merged, if that's easier.

JWO719:
Unfortunately there are some merge conflicts in the meantime. @tanem could you resolve the conflicts?

tanem:
Ah ok, no worries @JWO719, will get onto that ASAP 💯

tanem:
Hey @JWO719, have resolved the merge conflicts :+1:

src/internal/operators/groupBy.ts
```diff
@@ -71,16 +66,16 @@ export function groupBy<T, K, R>(keySelector: (value: T) => K, elementSelector?:
...
+ * of(
  *   {id: 1, name: 'JavaScript'},
  *   {id: 2, name: 'Parcel'},
  *   {id: 2, name: 'webpack'},
- *   {id: 1, name: 'TypeScript'}
+ *   {id: 1, name: 'TypeScript'},
  *   {id: 3, name: 'TSLint'}
  * ).pipe(
  *   groupBy(p => p.id, p => p.name),
  *   mergeMap( (group$) => group$.pipe(reduce((acc, cur) => [...acc, cur], ["" + group$.key]))),
```
dzhavat: Not really part of the PR but I noticed this extra space `mergeMap( (group$)...`. Can we get it removed?

And one more thing. Is there any reason why `group$.key` is send as a string and later parsed to an integer? If the string is required, does it make sense then to make it ``[`${group$.key}`]`` instead?

tanem: Thanks for the feedback @dzhavat, made both of those changes :+1:

dzhavat: Looks better 👍 

I'm actually not part of the RxJS team, so I commented more as a fellow contributor rather than someone who is responsible for reviewing the PR :)

Issue #4538, opened 2019-02-07
Ambient const enums are not allowed when the '--isolatedModules' flag is provided (with typescript)


steabert:
## Bug Report

**Current Behavior**
When compiling a typescript file that imports something from `rxjs` with the `isolatedModules` option for typescript set to true, the following error occurs:

```
node_modules/rxjs/internal/Notification.d.ts:3:27 - error TS1209: Ambient const enums are not allowed when the '--isolatedModules' flag is provided.

3 export declare const enum NotificationKind {
```

**Reproduction**

Create minimal project
```shell
yarn add --dev typescript rxjs
ts 
```
Create `test.ts`:
```ts
import {Observable} from 'rxjs'
```

Run Typescript compiler
```shell
yarn tsc --lib esnext --isolatedModules test.ts
```

**Expected behavior**
No error.

**Environment**
- Runtime: Node v10.15.0
- RxJS version: 6.4.0
- (If bug is related) Loader, build configuration: typescript 3.3.1

**Possible Solution**

**Additional context/Screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.

cartant:
I'm not familiar with the error, but I have a hunch that it might be addressed by the enum being exported in this PR: https://github.com/ReactiveX/rxjs/pull/4514

BrianMitchL:
This also causes compilation errors with @babel/plugin-transform-typescript (the TypeScript compiler used in create-react-app), which [does not support const enums](https://babeljs.io/docs/en/babel-plugin-transform-typescript#caveats).

cartant:
@BrianMitchL Although it also pertains to const enums, your issue is unrelated to this one. Please open an issue and fill out the template - including all relevant version information.

IIRC, TypeScript 2.4 is the version in which const enums were introduced, so I'd be more than a little surprised if the current version of the transform does not support them.

Anyway, please move all discussion to a new issue.

am17torres:
@steabert I just ran into this issue after running a `yarn upgrade`

```
./node_modules/rxjs/internal/Notification.d.ts
Type error: Ambient const enums are not allowed when the '--isolatedModules' flag is provided.  TS1209

    1 | import { PartialObserver } from './types';
    2 | import { Observable } from './Observable';
  > 3 | export declare const enum NotificationKind {
      |                           ^
    4 |     NEXT = "N",
    5 |     ERROR = "E",
    6 |     COMPLETE = "C"
```
[...]

PR #4556, opened 2019-02-11
fix(Notification): replace const enum

cartant:
**Description:**

This PR replaces the `NotiticationKind` const enum that was introduced in 6.4.0 with a string-literal union type. The latter is type-safe and compatible with single-module transpilation. Const enums are not, as the transpiler cannot know what should be used as a replacement for a reference to a const enum without using type information.

The `NotiticationKind` const enum is a problem for any projects using the `--isolatedModules` flag or for projects using the Babel TypeScript transform - which means any TypeScript projects generated with `create-react-app`.

**Related issue (if exists):** #4538

### DIFF
--- a/src/internal/Notification.ts
+++ b/src/internal/Notification.ts
@@ -3,8 +3,17 @@ import { Observable } from './Observable';
 import { empty } from './observable/empty';
 import { of } from './observable/of';
 import { throwError } from './observable/throwError';
+import { deprecate } from 'util';
 
-export type NotificationKind = 'N' | 'E' | 'C';
+// TODO: When this enum is removed, replace it with a type alias. See #4556.
+/**
+ * @deprecated NotificationKind is deprecated as const enums are not compatible with isolated modules. Use a string literal instead.
+ */
+export enum NotificationKind {
+  NEXT = 'N',
+  ERROR = 'E',
+  COMPLETE = 'C',
+}
 
 /**
  * Represents a push-based event or value that an {@link Observable} can emit.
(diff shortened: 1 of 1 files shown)


cartant:
BTW, I attempted to add a `transpile` script to reproduce the error - and guard against it in the future - but I was unable to get it to work, as TypeScript seems to only perform the check that effects that error for packages consumed from `node_modules`. Consuming them from `dist` effects a truckload of other errors about re-exports also not being compatible with `--isolatedModules`.

mrmckeb:
This is great, and will allow us to move to v6.4. Are there any concerns about merging this in?

benlesh:
Unfortunately, this is a breaking change.  Anyone using the `NotificationKind` enum like `NotificationKind.NEXT` will be broken with this change.

benlesh:
I think if we just leave the enum where it is, and convert everything else to just use `'N'|'E'|'C'` we should be able to resolve the bug.

We can then `@deprecate` `NotificationKind`.

cartant:
@benlesh Unfortunately, this has to be a breaking change. It's the export that is the problem - regardless of whether or not `NotificationKind` is used:

```
node_modules/rxjs/internal/Notification.d.ts:3:27 - error TS1209: Ambient const enums are not allowed when the '--isolatedModules' flag is provided.
3 export declare const enum NotificationKind {
```

The introduction of the const enum was a breaking change for many people - myself included - so I see no reason the fix cannot also be a breaking change.

kwonoj:
Maybe just `const enum` to `enum`?

cartant:
@benlesh I've updated the PR to use a literal union in the `Notification` signature with `NotificationKind` being kept as a **non-const** enum. A const enum cannot be exported without effecting an error in situations in which isolated modules are used.

I've added a deprecation - with a message - to the enum. In the next major version, the enum could be ditched and `NotificationKind` could be declared as a type alias for the literal union.

mrmckeb:
It looks like this is actually passing, the Travis error seems to be unrelated. Any chance we can get this in now?

PR #4580, opened 2019-02-19
docs(operators): Added docs/examples for isEmpty

jonahbron:
**Description:**

Added documentation for `isEmpty` operator

### DIFF
--- a/src/internal/operators/isEmpty.ts
+++ b/src/internal/operators/isEmpty.ts
@@ -3,6 +3,66 @@ import { Subscriber } from '../Subscriber';
 import { Observable } from '../Observable';
 import { OperatorFunction } from '../types';
 
+/**
+ * Emits false if the input observable emits any values, or emits true if the
+ * input observable completes without emitting any values.
+ *
+ * <span class="informal">Tells whether any values are emitted by an observable</span>
+ *
+ * ![](isEmpty.png)
+ *
+ * `isEmpty` transforms an Observable that emits values into an Observable that
+ * emits a single boolean value representing whether or not any values were
+ * emitted by the source Observable. As soon as the source Observable emits a
+ * value, `isEmpty` will emit a `false` and complete.  If the source Observable
+ * completes having not emitted anything, `isEmpty` will emit a `true` and
+ * complete.
+ *
+ * A similar effect could be achieved with {@link count}, but `isEmpty` can emit
+ * a `false` value sooner.
+ *
+ * ## Examples
+ *
+ * Emit `false` for a non-empty Observable
+ * ```javascript
+ * import { Subject } from 'rxjs';
+ * import { isEmpty } from 'rxjs/operators';
+ *
+ * const source = new Subject<string>();
+ * const result = source.pipe(isEmpty());
+ * source.subscribe(x => console.log(x));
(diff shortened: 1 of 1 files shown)


src/internal/operators/isEmpty.ts
```diff
@@ -3,6 +3,60 @@ import { Subscriber } from '../Subscriber';
 import { Observable } from '../Observable';
 import { OperatorFunction } from '../types';
 
+/**
+ * Emits true if the input observable emits any values, or emits false if the
```
martinsik: I think this textual description says the exact opposite of what this operator does:

> Emits true if the input observable emits any values...

... and then ...

> As soon as the source Observable emits a value, `isEmpty` will emit a `true`...

But `isEmpty` emits `false` if the source emits any value. https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/isEmpty.ts#L33

src/internal/operators/isEmpty.ts
```diff
@@ -3,6 +3,60 @@ import { Subscriber } from '../Subscriber';
...
+ *
+ * const values = from(['a', 'b', 'c']).pipe(shareReplay());
+ * const result = values.pipe(isEmpty());
+ * values.subscribe(x => console.log(x));
+ * result.subscribe(x => console.log(x));
+ * // Results in:
+ * // a
+ * // false
+ * // b
+ * // c
```
martinsik: The result of this example is different (maybe this one comes from RxJS 4 examples?). It first prints letters `a`, `b` and `c` and then prints `false`.

src/internal/operators/isEmpty.ts
```diff
@@ -3,6 +3,60 @@ import { Subscriber } from '../Subscriber';
...
+ * input observable completes without emitting any values.
+ *
+ * <span class="informal">Tells whether any values are emitted by an observable</span>
+ *
+ * ![](isEmpty.png)
+ *
+ * `isEmpty` transforms an Observable that emits values into an Observable that
+ * emits a single boolean value representing whether or not any values were
+ * emitted by the source Observable. As soon as the source Observable emits a
+ * value, `isEmpty` will emit a `true`.  If the source Observable completes having
```
martinsik: I think it would be good if this mentioned that this operator completes right after emitting `false` because it seems like it keeps re emitting source:

> As soon as the source Observable emits a value, `isEmpty` will emit a `false` and complete.

instead of this

> As soon as the source Observable emits a value, `isEmpty` will emit a `true`

src/internal/operators/isEmpty.ts
jonahbron: Oh oops, I got them on backwards.  Will fix.

src/internal/operators/isEmpty.ts
jonahbron: Fixed the example.

src/internal/operators/isEmpty.ts
jonahbron: Changed as recommended

jonahbron:
@martinsik all of your comments have been resolved, thanks for reviewing 😃

jonahbron:
Fixed the commit message warning.

PR #4583, opened 2019-02-22
RxJS v7 Scheduler Work

benlesh:
## This work is for v7 and should not be merged until we've split off a patch branch for v6

Thus far:

- Removes dependencies on `SchedulerAction` from all operators relying on them for recursive scheduling
- Rewrite of `VirtualTimeScheduler` and `TestScheduler`
- `TestScheduler` now defaults to 1ms frames, and no max frame count (formerly 10s and 750 frames)
- All schedulers rewritten to be smaller and simpler (esp asyncScheduler, which is the most commonly used)
- Removed some `interval` tests that frankly didn't make sense.
- `animationFrameScheduler`, when called with a delay > 0, will now delay, then schedule an animationFrame, rather than [just delay](https://github.com/ReactiveX/rxjs/blob/c4a84cc116837bc1616a3e5b501efe4d41801288/src/internal/scheduler/AnimationFrameAction.ts#L20).


benlesh:
After discussing this with the core team, in particular @trxcllnt today, I was reminded of a few things:

### Semantics 

 The timing implications between scheduling things with `setInterval` and scheduling things with `setTimeout` recursively.

### Subscription allocation

(Okay, I knew this one going into it, but it's still worth considering) There's also an additional impact in this PR of allocating (and executing methods of) one subscription per each recursive loop, rather than just using a single Subscription for all recursive loops.

### Abilty to alter delays in rescheduled Actions

This seems to have been added solely to support `timer(100, 2000)` type scenarios, IIRC. We decided it probably wasn't a huge loss to allocate a second subscription in this case.

## Alternate designs to what's going on in this PR

All of the below eliminate `Action`, lower complexity, and reduce the amount of `Subscription` inheritance we have going on in the library. Also all of the solutions below will limit the number of Subscriptions required to accomplish scheduling to one subscription, even in recursive/repeated cases.

### Passing a reschedule function to `work`:

The idea here is we provide a function as a second argument to the `work` callback provided to `schedule`. That function, when called, will reschedule the action.

```ts
scheduler.schedule((state: S, reschedule: (nextState: S) => void) => {
   // in order to schedule recursively, you'd call reschedule with the next state
   if (shouldRecurse) {
     state.counter++;
     reschedule(state);
   }
});
```
- **Pros**
  - Fairly simple API.
  - Somewhat backward compatible (but not completely)
- **Cons**
  - Implementation could be tricky, because we have this function that could be called at any point during `work`'s execution, and executing it synchronously might have weird impacts on schedulers that are synchronous (this is an extreme edge case though).

### Utilizing the return value of `work`:

The idea here is that we use the return value of `work` to decide whether or not it should be rescheduled.

```ts
scheduler.schedule((state: S) => {
   // in order to schedule recursively, you'd call reschedule with the next state
   if (shouldRecurse) {
     state.counter++;
     return state;
   } else {
     return;
   }
});
```
- **Pros**
  - Fairly simple API.
  - Somewhat backward compatible (but not completely)
  - rescheduling is forced to wait until `work` completes, making it a little easier to reason about.
- **Cons**
  - What if someone wants a state of `undefined`?
  - a hair more verbose than the idea above

### Having two functions, one for `once` and another to `repeat`

Here, we'd change the `SchedulerLike` API to have two methods, `once` for scheduling something to happen one time, and `repeat` to schedule something to happen over and over. 

```ts
scheduler.once((state: S) => { /* do a thing */ });

scheduler.repeat((state: S) => { /* do a thing over and over */ });
```
- **Pros**
  - Fairly simple API.
- **Cons**
  - Completely different API that previous schedulers.
  - Implementing a scheduler is a little more complicated now.
  - In this case, we'd need to provide a mechanism to `repeat`'s `work` function to stop the repetition. Maybe just the `Subscription` instance? (For edge-case synchronous schedulers)

src/internal/scheduler/animationFrame.ts
```diff
@@ -50,18 +56,25 @@ export const animationFrame: SchedulerLike = {
          }, delay)
```
trxcllnt: @benlesh I'm very pro-this-change :-). This is how it should have always worked, definitely an oversight on my part initially.

src/internal/scheduler/animationFrame.ts
```diff
@@ -50,18 +56,25 @@ export const animationFrame: SchedulerLike = {
...
+
+      // Set up the teardown
       subscription.add(() => {
-        const index = _queue.indexOf(subscription);
+        // Remove the task from the queue if it's still in there
+        // It won't be in there if it's been executed.
+        const index = _currentQueue.indexOf(task);
         if (index >= 0) {
-          _queue.splice(index - 2, 3);
+          _currentQueue.splice(index, 1);
```
trxcllnt: @benlesh could we eliminate these `indexOf()` and `splice()` calls if we check `task.subscription.closed` inside `flush()`?

src/internal/scheduler/animationFrame.ts
```diff
@@ -50,18 +56,25 @@ export const animationFrame: SchedulerLike = {
...
         }
-        if (_queue.length === 0) {
-          _animationFrameId = -1;
-          cancelAnimationFrame(_animationFrameId);
+
+        // If the current animation frame queue is empty, AND the overall queue is empty, stop the animation frame loop
+        if (_currentQueue.length === 0 && _nextQueue.length === 0) {
```
trxcllnt: (continuing from the comment above :point_up_2:) if we aren't splicing things out of the queues, we'll probably need to track the following values along with the queues (starting from line 12 above), and update them in `flush()` also:
```ts
let _currentQueue: Task[] = [];
let currentQueueLength = 0;
let currentQueueDisposedCount = 0;
let _nextQueue: Task[] = [];
let nextQueueLength = 0;
let nextQueueDisposedCount = 0;
```

src/internal/scheduler/animationFrame.ts
```diff
@@ -71,27 +84,60 @@ export const animationFrame: SchedulerLike = {
...
+
+    // Move to the next queue
+    _currentQueue = _nextQueue;
+
+    // Start a new queue for newly scheduled or rescheduled items
+    _nextQueue = [];
+
+    // process the tasks on this queue FIFO.
+    while (_currentQueue.length > 0) {
+      const task = _currentQueue.shift();
```
trxcllnt: (continuing the comment thread) could this be a non-destructive `for (let i = 0; i++ < currentQueueLength;) {}` loop rather than `shift()`?

src/internal/testing/TestScheduler.ts
```diff
@@ -24,378 +44,508 @@ interface FlushableTest {
...
 export type subscriptionLogsToBeFn = (marbles: string | string[]) => void;
 
 export class TestScheduler extends VirtualTimeScheduler {
-  public readonly hotObservables: HotObservable<any>[] = [];
-  public readonly coldObservables: ColdObservable<any>[] = [];
-  private flushTests: FlushableTest[] = [];
-  private runMode = false;
+  private _runMode = false;
+  private _flushTests: FlushableTest[] = [];
+  // private _patches = DEFAULT_SCHEDULER_PATCHES;
```
cartant: If this - and the associated interface - isn't used, can we get rid of it? Personally, I'm not a fan of commented-out code. Either that or mark it with a TODO comment.

src/internal/scheduler/animationFrame.ts
```diff
@@ -36,5 +46,99 @@ import { AnimationFrameScheduler } from './AnimationFrameScheduler';
  * @name animationFrame
  * @owner Scheduler
  */
+export const animationFrame: SchedulerLike = {
+  schedule<S>(work: (state: S, reschedule: (nextState: S) => void) => void, delay = 0, state?: S): Subscription {
```
cartant: The previous implementation of `AnimationFrameScheduler` extended `AsyncScheduler` and got it's `schedule` implementation from the base class, so it supported `__rx_scheduler_overrides__`. The current implementation disregards `__rx_scheduler_overrides__` so it will behave differently under test - i.e. it'll fail, as it'll be async.

src/internal/operators/delay.ts
```diff
@@ -74,88 +73,41 @@ class DelayOperator<T> implements Operator<T, T> {
...
-
-    const scheduler = this.scheduler;
-    const message = new DelayMessage(scheduler.now() + this.delay, notification);
-    this.queue.push(message);
-
-    if (this.active === false) {
-      this._schedule(scheduler);
-    }
+    const destination = this.destination as Subscriber<T>;
+    this.scheduler.schedule<DelayState<T>>(dispatch, this.delay, { notification, destination });
```
cartant: What happens with the subscription that's returned here? What if the subscriber to the observable returned by `delay` unsubscribes? Shouldn't any pending notifications be unsubscribed in that situation?

src/internal/observable/generate.ts
```diff
@@ -281,99 +281,99 @@ export function generate<T, S>(initialStateOrOptions: S | GenerateOptions<T, S>,
...
+          return;
+        }
+        if (subscriber.closed) {
+          return;
+        }
+        subscriber.next(value);
+        if (subscriber.closed) {
+          return;
+        }
+        subscriber.add(scheduler.schedule(work, 0, state));
```
cartant: Could/should this use `reschedule`?

spec/schedulers/VirtualTimeScheduler-spec.ts
```diff
@@ -61,43 +61,9 @@ describe('VirtualTimeScheduler', () => {
...
-        return;
-      }
-      const virtualAction = this as VirtualAction<string>;
-      expect(virtualAction.delay).to.equal(expected.shift());
-      this.schedule(state, virtualAction.delay);
-    }, 100, 'test');
-
-    v.flush();
-    expect(count).to.equal(3);
-  });
```
cartant: Was this removed before the `rescheduled` parameter was added? Should rescheduling be supported here?

spec/schedulers/AsapScheduler-spec.ts
```diff
@@ -1,103 +1,53 @@
...
-    expect(stubSetInterval).to.have.property('callCount', 1);
-    fakeTimer.tick(period);
-    expect(state).to.have.property('index', 1);
-    expect(stubSetInterval).to.have.property('callCount', 1);
-    fakeTimer.tick(period);
-    expect(state).to.have.property('index', 2);
-    expect(stubSetInterval).to.have.property('callCount', 1);
-    stubSetInterval.restore();
-    sandbox.restore();
-  });
```
cartant: Now that there is a `rescheduler` parameter, should this test be kept?

spec/schedulers/AsapScheduler-spec.ts
```diff
@@ -1,103 +1,53 @@
...
-    expect(stubSetInterval).to.have.property('callCount', 1);
-    fakeTimer.tick(period);
-    expect(state).to.have.property('index', 1);
-    expect(stubSetInterval).to.have.property('callCount', 2);
-    fakeTimer.tick(period);
-    expect(state).to.have.property('index', 2);
-    expect(stubSetInterval).to.have.property('callCount', 3);
-    stubSetInterval.restore();
-    sandbox.restore();
   });
```
cartant: Now that there is a `rescheduler` parameter, should this test be kept?

src/internal/observable/generate.ts
benlesh: I think I'd like to refactor these individually after this PR. It's already doing so much.

src/internal/scheduler/animationFrame.ts
benlesh: I was able to accomplish this with a single counter that is incremented when we schedule, and decremented during unsubscription (teardown).

benlesh:
I think we should table this for now, there's been a few discussions and designs around this since then, and I think for v7, we want to limit breaking changes here. We may just try to move them to their own import site and have a few deprecations at this point.

PR #4585, opened 2019-02-23
fix(pluck): key union type strictness

imcotton:
**Description:**
Fixing the overloading type define ~~off-by-one~~ being widen


imcotton:
Wait... does the loosely typed key inputs actually desired behavior? I mean it's valid JavaScript runtime code but shall the TS add the extra checking over here?

benlesh:
this change is likely to break a few people, so I'm going to target it at the next major version

imcotton:
> target it at the next major version

Good call.

_BTW, I've made rebase & squash for easy picking up later on._

PR #4589, opened 2019-02-25
Fixes error "Property 'observable' does not exist on type 'SymbolCons…

nmpribeiro:
Fixes error "Property 'observable' does not exist on type 'SymbolConstructor'".

**Description:**

Fixes error "Property 'observable' does not exist on type 'SymbolConstructor'" by importing Symbol.observable from file

Not sure if it's possible to simplify it somehow, as I am not fully aware of how the Symbol get's imported in the first place.

This error was happening on a new Angular v7 project, where I have libraries - that live in an external location to the root folder - importing 'rxjs' with an alias: 
in `tsconfig.js`:
```
"paths": {
     ...
      "rxjs": [
        "node_modules/rxjs/index.js"
      ],
      "rxjs/*": [
        "node_modules/rxjs/*"
      ],
      ...
}
```

There's no topic with issue, but in a nutshell, this was the output of `ng serve`:
```
ERROR in node_modules/rxjs/internal/types.d.ts(48,13): error TS2339: Property 'observable' does not exist on type 'SymbolConstructor'.

ℹ ｢wdm｣: Failed to compile.
```

While with the inclusion of the file, it compiles without error.

### DIFF
--- a/src/internal/types.ts
+++ b/src/internal/types.ts
@@ -1,3 +1,4 @@
+import 'rxjs/internal/symbol/observable';
 import { Observable } from './Observable';
 import { Subscription } from './Subscription';
 


benlesh:
This is interesting. Do you have a simple repro you can share? This is the first time I've seen this as an issue.

nmpribeiro:
Hi @benlesh ,

By setting up the repo, I actually figured it won't break if I import Observable from rxjs:
https://github.com/nmpribeiro/test_alias_rxjs/blob/master/ng-test/src/app/app.component.ts#L4

However, is this your intended behavior? Because I feel we should be able to import it selectively.

Anyway, have a look here: https://github.com/nmpribeiro/test_alias_rxjs

benlesh:
@nmpribeiro Yeah, you should never import from anything under `rxjs/internal` .. that's "internal" :wink:

benlesh:
This seems unnecessary to me, @cartant can you give this a second set of eyes?

cartant:
I suspect that the issue might be that `type` is imported before `symbol/observable` [here](https://github.com/ReactiveX/rxjs/blob/c4f44b9ac0b19d7425dcf84e5b0446df09fa70c7/src/internal/Observable.ts).

Or, maybe, because the `symbol/observable` import has an unnecessary `../internal` - it's already in the `internal` directory.

Whatever the reaseon, as Ben has pointed out, if users avoid importing from `internal`, this isn't really an issue.

cartant:
^ @benlesh

benlesh:
Yeah. Sorry @nmpribeiro, I'm inclined to close this PR as an unnecessary change. Thanks for the additional review, @cartant.

Issue #4565, opened 2019-02-14
throwIfEmpty doesn't throw on retry


bgotink:
## Bug Report

**Current Behavior**

`throwIfEmpty` doesn't throw when the source observable completes without any value after it's re-subscribed to.

**Reproduction**
- https://stackblitz.com/edit/rxjs-2nbh1q?file=index.ts

```ts
source$.pipe(
  throwIfEmpty(),
  someOperatorThatMightThrow(),
  retry(),
).subscribe(value => console.log('got value', value));
```

If `source$` emits a value, the `someOperatorThatMightThrow()` operator throws, and `source$` completes on the second subscription without emitting any value, the `throwIfEmpty()` doesn't throw.

**Expected behavior**

I'd expect the `throwIfEmpty` to result in a failure.

**Environment**
- Runtime: Chrome v71
- RxJS version: 6.4.x

**Additional context/Screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.

benlesh:
This is resolve in the latest 6.x and 7.x releases.

PR #4592, opened 2019-02-25
feat(tap): reset throwIfEmpty hasValue on retry

arpadvas:
**Description:**

https://github.com/ReactiveX/rxjs/issues/4565

### DIFF
--- a/src/internal/operators/tap.ts
+++ b/src/internal/operators/tap.ts
@@ -78,6 +78,12 @@ class DoOperator<T> implements Operator<T, T> {
               private complete?: () => void) {
   }
   call(subscriber: Subscriber<T>, source: any): TeardownLogic {
+    /* make sure throwIfEmpty hasValue is set to false on re-subscribe */
+    if (this.nextOrObserver) {
+      if (this.nextOrObserver['hasValue']) {
+        this.nextOrObserver['hasValue'] = false;
+      }
+    }
     return source.subscribe(new TapSubscriber(subscriber, this.nextOrObserver, this.error, this.complete));
   }
 }
--- a/src/internal/operators/tap.ts
+++ b/src/internal/operators/tap.ts
@@ -78,12 +78,6 @@ class DoOperator<T> implements Operator<T, T> {
               private complete?: () => void) {
   }
   call(subscriber: Subscriber<T>, source: any): TeardownLogic {
-    /* make sure throwIfEmpty hasValue is set to false on re-subscribe */
-    if (this.nextOrObserver) {
-      if (this.nextOrObserver['hasValue']) {
-        this.nextOrObserver['hasValue'] = false;
-      }
-    }
     return source.subscribe(new TapSubscriber(subscriber, this.nextOrObserver, this.error, this.complete));
   }
 }
--- a/src/internal/operators/throwIfEmpty.ts
+++ b/src/internal/operators/throwIfEmpty.ts
@@ -2,7 +2,7 @@ import { EmptyError } from '../util/EmptyError';
 import { Observable } from '../Observable';
 import { Operator } from '../Operator';
 import { Subscriber } from '../Subscriber';
-import { TeardownLogic } from '../types';
+import { TeardownLogic, MonoTypeOperatorFunction } from '../types';
 
 /**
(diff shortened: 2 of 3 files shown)


kwonoj:
bit unsure if we allowed mutate values in subscribers, will let other reviewers give a peek.

src/internal/operators/tap.ts
```diff
@@ -78,6 +78,12 @@ class DoOperator<T> implements Operator<T, T> {
               private complete?: () => void) {
   }
   call(subscriber: Subscriber<T>, source: any): TeardownLogic {
+    /* make sure throwIfEmpty hasValue is set to false on re-subscribe */
+    if (this.nextOrObserver) {
+      if (this.nextOrObserver['hasValue']) {
```
benlesh: We can't do this here. Anyone that passes any object with a `hasValue` on it will have that mutated. It's too unpredictable.

More likely, we have a bug in the `throwIfEmpty` implementation that we need to clear up.

benlesh:
I do see the issue, thank you for bringing it to my attention, @arpadvas... But this isn't the right fix. Really, we need to update the `throwIfEmpty` implementation to return something that will create a new object each time.

A quick fix would be to wrap it in a defer:

```ts
export const throwIfEmpty =
  <T>(errorFactory: (() => any) = defaultErrorFactory) => (source: Observable<T>) => defer(() => source.pipe(tap<T>({
    hasValue: false,
    next() { this.hasValue = true; },
    complete() {
      if (!this.hasValue) {
        throw errorFactory();
      }
    }
  } as any));
```

from there, we can probably reintroduce a small, custom implementation at a later point.

benlesh:
I'm marking this as "WIP" and "bug" because it does outline a bug we have, however, it doesn't resolve the issue appropriately. You don't have to fix it, @arpadvas, if you don't want to, but I'm leaving this open so someone remembers to.

Thanks for your hard work!

arpadvas:
Thanks for the feedback @benlesh. I see your point... I will look into trying solving it as you suggested above.

src/internal/operators/throwIfEmpty.ts
```diff
@@ -16,32 +18,57 @@ import { MonoTypeOperatorFunction } from '../types';
...
+  }
+
+  protected _next(value: T): void {
+    this.hasValue = true;
+    this.destination.next(value);
+  }
+
+  protected _complete() {
+    if (!this.hasValue) {
+        this.destination.error(this.errorFactory());
```
benlesh: this is going to sound funny, but since `errorFactory` is a _user provided function_, we need to wrap it in a `try/catch`, and send the error along to `destination.error` if there is one. And you can't do it along with the other `destination.error` call, or any error that happens in there will unwind back to your try catch... so it has to be like:

```ts
if (!this.hasValue) {
  let err: any;
  try {
    err = this.errorFactory();
  } catch (e) {
    err = e;
  }
  this.destination.error(err);
} else {
```

src/internal/operators/throwIfEmpty.ts
```diff
@@ -16,32 +18,57 @@ import { MonoTypeOperatorFunction } from '../types';
...
  *   ),
  * )
  * .subscribe({
  *   next() { console.log('The button was clicked'); },
- *   error(err) { console.error(err); },
+ *   error(err) { console.error(err); }
  * });
  * ```
  *
  * @param {Function} [errorFactory] A factory function called to produce the
```
benlesh:
```suggestion
 * @param errorFactory A factory function called to produce the
```

We don't actually need this additional typing information now, it's just from legacy docs. (not your doing, just adding that change in as I notice it)

src/internal/operators/throwIfEmpty.ts
arpadvas: Makes sense... I will fix it. Thanks for the review!

benlesh:
Great catch and a great fix, @arpadvas! (I broke this in a bit of late-night cleverness, it seems :laughing:)

Thank you again for all of your hard work!

arpadvas:
Thank you @benlesh! My pleasure to work on such a great project! 😊

PR #4595, opened 2019-02-26
feat(scheduled): Adds scheduled creation method

benlesh:
The `scheduled` creation method is meant to create an observable with a scheduler. The goal of this PR is to set us up to remove scheduling from very common code paths like `from` and `of`, amongst others.

This deprecates versions of `from`, `concat`, `merge`, `startWith` and `endWith` that allow passing a scheduler.

This PR also simplifies a few common code paths.

Put another way, this PR:

- Splits off the scheduled code paths for various branches of `from` into their own functions
- Updates a bunch of scheduled versions of functions as *deprecated*
- Cleans up some scheduling-related code


src/internal/observable/empty.ts
```diff
@@ -53,19 +53,16 @@ export const EMPTY = new Observable<never>(subscriber => subscriber.complete());
...
- * @owner Observable
- * @deprecated Deprecated in favor of using {@link index/EMPTY} constant.
+ * @deprecated Deprecated in favor of using {@link EMPTY} constant, or {@link scheduled} (e.g. `scheduled([], scheduler)`)
  */
 export function empty(scheduler?: SchedulerLike) {
   return scheduler ? emptyScheduled(scheduler) : EMPTY;
 }
 
-export function emptyScheduled(scheduler: SchedulerLike) {
+function emptyScheduled(scheduler: SchedulerLike) {
```
martinsik: I'm just curious how is this supposed to be used by users? `empty()` is deprecated in favor of `EMPTY` and `emptyScheduled` is not exported

benlesh: `EMPTY.pipe(observeOn(scheduler))`.  

It's unergonomic because it's not really something people should be doing that often, and we really need to reduce the API surface area.

DaSchTour:
Is there some documentation how migration from `of`, `from` to `scheduled` looks like?

Issue #4086, opened 2018-09-03
reduce operator typing gets lost


bridzius:
## Bug Report

**Current Behavior**
When using reduce/scan, typings get lost on 6.3.1 version of rxjs
```
﻿﻿index.ts:5:27 - error TS2345: Argument of type 'MonoTypeOperatorFunction<any[]>' is not assignable to parameter of type 'OperatorFunction<string, any[]>'.
  Types of parameters 'source' and 'source' are incompatible.
    Type 'Observable<string>' is not assignable to type 'Observable<any[]>'.
      Type 'string' is not assignable to type 'any[]'.

5   return from(items).pipe(reduce((acc, item) => acc.concat([item]), []));
```

**Reproduction**
- REPL or Repo link:
https://stackblitz.com/edit/typescript-osxh4v

```js
import { from, Observable } from "rxjs";
import { reduce } from "rxjs/operators";

function load(items: string[]): Observable<string[]> {
  return from(items).pipe(reduce((acc, item) => acc.concat([item]), []));
}
```

**Expected behavior**
It should not throw the typing error when trying to reduce.

**Environment**
- Runtime: NodeJS 8.11
- RxJS version: 6.3.1
- Angular 6.2, but reproducible without as well.

jgbpercy:
tl;dr: I'm pretty sure this isn't a bug, and I think your code was getting around typechecking due to a bug in rxjs 6.2.1. I believe you need to be explicit about the types passed to reduce, and the following should type check and work correctly under rxjs 6.3.x+:
```
function load(items: string[]): Observable<string[]> {
  return from(items).pipe(reduce((acc, item: string) => acc.concat([item]), []));
}
```
(change is `item: string` instead of just `item`)

----
Full explanation (disclaimer - I don't work on rxjs, I just enjoy figuring this stuff out and I could be wrong):

I believe your code was getting around correct typing using the funky `pipe<R>(...operations: OperatorFunction<any, any>[]): Observable<R>;` pipe overload added in https://github.com/ReactiveX/rxjs/pull/3789 and removed in https://github.com/ReactiveX/rxjs/pull/3945 .

If you take your code under rxjs 6.2.2 and change it to:
```
function load(items: string[]): Observable<string[]> {
  const thingToReturn = from(items).pipe(reduce((acc, item) => acc.concat([item]), []));
  return thingToReturn;
}
```
[...]

PR #4598, opened 2019-02-26
fix(scan): fixed declarations to properly support different return types

david-driscoll:
**Description:**
This solves an issue where if you would reduce over `T` you can get `T` out reliably, but you could
not get out `R` in anyway without manually typing the reduce method call

#4086

cc @cartant

### DIFF
--- a/compat/operator/reduce.ts
+++ b/compat/operator/reduce.ts
@@ -2,9 +2,9 @@ import { Observable } from 'rxjs';
 import { reduce as higherOrderReduce } from 'rxjs/operators';
 
 /* tslint:disable:max-line-length */
-export function reduce<T>(this: Observable<T>, accumulator: (acc: T, value: T, index: number) => T, seed?: T): Observable<T>;
-export function reduce<T>(this: Observable<T>, accumulator: (acc: T[], value: T, index: number) => T[], seed: T[]): Observable<T[]>;
 export function reduce<T, R>(this: Observable<T>, accumulator: (acc: R, value: T, index: number) => R, seed: R): Observable<R>;
+export function reduce<T>(this: Observable<T>, accumulator: (acc: T, value: T, index: number) => T, seed?: T): Observable<T>;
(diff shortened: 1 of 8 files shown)


compat/operator/reduce.ts
```diff
@@ -2,9 +2,8 @@ import { Observable } from 'rxjs';
 import { reduce as higherOrderReduce } from 'rxjs/operators';
 
 /* tslint:disable:max-line-length */
+export function reduce<T, R>(this: Observable<T>, accumulator: (acc: R, value: T, index: number) => R, seed?: R): Observable<R>;
```
cartant: I don't think the seed can be optional if the accumulator has a different type the the source elements. If a seed is not specified, my understanding is that the first element will be used, instead. That means that the initial accumulator value will be a `T` and not an `R` as indicator by the call signature.

src/internal/operators/reduce.ts
```diff
@@ -6,9 +6,9 @@ import { OperatorFunction, MonoTypeOperatorFunction } from '../types';
 import { pipe } from '../util/pipe';
 
 /* tslint:disable:max-line-length */
+export function reduce<T, R>(accumulator: (acc: R, value: T, index: number) => R, seed: R): OperatorFunction<T, R>;
```
benlesh: So there's this messed up scenario I think we need to test for:

`of(1).pipe(reduce((acc, value, i) => i === 0 ? [] : [..value], { seed: 'LOL' }))` The problem is that `acc` gets typed wrong.

I'm sure that `scan` has the same issue.

david-driscoll: So we assume that `seed` and `ReturnType<typeof accumulator>` are the same type, if they differ we get an error because accumulator returns an invalid value.  Thoughts?

compat/operator/reduce.ts
david-driscoll: You're correct... adding tests and updating.

cah-dunn:
Would love to see this merged as it is creating quite the nuisance while migrating from `rxjs-compat`.

PR #4602, opened 2019-02-28
test(dtslint): add fromEvent dtslint test (#4093)

dkosasih:
**Description:** Add dtslint test for fromEvent 


spec-dtslint/observables/fromEvent-spec.ts
```diff
@@ -0,0 +1,20 @@
...
+const nodeStyleEventEmitter: NodeStyleEventEmitter = {
+  addListener: (eventName, handler) => nodeStyleEventEmitter ,
+  removeListener: (eventName, handler) => nodeStyleEventEmitter
+};
+
+it('should infer correctly with HasEventTargetAddRemove event type', () => { 
+  const a = fromEvent(hasEventTargetAddRemove, 'foo'); // $ExpectType Observable<string>
+});
+
+it('should infer correctly with NodeStyleEventEmitter event type', () => {
```
dkosasih: Hi @cartant,

Am I already in the right direction for this test?

dkosasih: Sorry to ping you again @cartant. 
I was just wondering if this is a need-to-fix typing or I am doing something wrong here to test the inference?'
or maybe @benlesh could help explaning?

benlesh: I think this looks okay... However, we probably want to test it against "real world" things... Try creating a DOMElement type like:   `const div: HTMLDivElement = !null;`  and seeing what you get back from that. You could do the same for node things like `process`, for example. Remember, it doesn't have to work at run time, it just needs to infer the correct type.

dkosasih:
Hello @benlesh ,
I've been getting this above build errror after rebase with master. 
```
...
Error: Errors in typescript@local for external dependencies:
../src/internal/Observable.ts(75,3): error TS2394: Overload signature is not compatible with function implementation.
../src/internal/observable/fromEvent.ts(191,44): error TS2345: Argument of type 'EventListenerOptions | undefined' is not assignable to parameter of type 'EventListenerOptions'.
  Type 'undefined' is not assignable to type 'EventListenerOptions'.
...
```
Is there anything new that I should be aware of to get this working? 
Dtslint works when I remove the fromEvent test. 
I will try to dig deeper meanwhile.

dkosasih:
I figured that the problem is because of the import from relative path to `fromEvent` classes and types like `HasEventTargetAddRemove`. 
I am not sure if I could import that from any public api that RxJS expose.
That raise error like `expect  TypeScript@local compile error`.

When I copy the interfaces into the spec file, the tests actually passing.

Any idea how we can resolve this? I don't think copying the interface into the spec file is ideal.

Issue #4616, opened 2019-03-04
[docs]: Operator `single` is missing the description, and examples


BioPhoton:
### Please check those that apply

- [ ] typo
- [x] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [x] needs example

### Description Of The Issue
The description and example section is missing for the `single` operator

weitalu:
I would like to help with this issue, created a local commit, 

but I could not submit the PR because I don't have enough access to create a remote branch.

What should I do ?

(Was following this guideline https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md#submitting-a-pull-request-pr)

JWO719:
@weitalu  I think the confusing point here is (also not mentioned in the document), you need to fork the rxjs repo, push to a branch of your private forked repository and afterwards github will propose you to create a merge request for that branch to the rxjs repo. This should basically work. When there is anything unclear or still not working, feel free to contact me! Thanks for helping out with this!

niklas-wortmann:
closed by #4623

PR #4623, opened 2019-03-07
docs(single): Updated docs for single with description and example (#…

weitalu:
**Description:**
Updated docs for single with description and example
#4616

### DIFF
--- a/src/internal/operators/single.ts
+++ b/src/internal/operators/single.ts
@@ -11,7 +11,33 @@ import { Observer, MonoTypeOperatorFunction, TeardownLogic } from '../types';
  * items, notify of an IllegalArgumentException or NoSuchElementException respectively. If the source Observable
  * emits items but none match the specified predicate then `undefined` is emitted.
  *
+ * <span class="informal">Like {@link first}, but emit with error notification if there is more than one value.</span>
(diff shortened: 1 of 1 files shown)


src/internal/operators/single.ts
```diff
@@ -11,8 +11,34 @@ import { Observer, MonoTypeOperatorFunction, TeardownLogic } from '../types';
...
+ * import { single } from 'rxjs/operators';
+ *
+ * const numbers = range(1,5).pipe(single());
+ * numbers.subscribe(x => console.log('never get called'), e => console.log('error'));
+ * // result
+ * // 'error'
+ * ```
+ *
+ * emits 'undefined'
+ * ```javascript
```
JWO719: could you change the examples to ts

weitalu: do you mean that change the word 'javascript' to 'typescript',

or use some feature of typescript in the example like interface or type ?

dzhavat: What @JWO719 meant was changing

```
```javascript
```

to

```
```ts
```

weitalu: Thank you for the hint !

PR #4625, opened 2019-03-08
docs(exhaustMap): fix example

tanem:
Add missing import 🦅

### DIFF
--- a/src/internal/operators/exhaustMap.ts
+++ b/src/internal/operators/exhaustMap.ts
@@ -38,12 +38,12 @@ export function exhaustMap<T, I, R>(project: (value: T, index: number) => Observ
  * ## Example
  * Run a finite timer for each click, only if there is no currently active timer
  * ```ts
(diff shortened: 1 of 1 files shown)


stck:
I think it better to avoid remove of line breaks. Your thoughts?

tanem:
> I think it better to avoid remove of line breaks. Your thoughts?

That was due to Prettier reformatting. Since there doesn't appear to be a consistent code style applied to the docs examples though, I'm happy to change it back in order to respect the existing style in this example.

tanem:
Thanks @JWO719, no problem at all 🙂

JWO719:
@tanem Could you report the merge conflicts?

tanem:
> @tanem Could you report the merge conflicts?

@JWO719 sure, all done :+1:

JWO719:
Thank you very much for all the work you put into the docs! I really appreciate that

PR #4636, opened 2019-03-14
docs(operators): Add a simple example in doc string

xuemind:
Add much simpler example to help understand what `switchMap` can do. It is easy to understand for document reader.

**Description:**

### DIFF
--- a/src/internal/operators/switchMap.ts
+++ b/src/internal/operators/switchMap.ts
@@ -36,6 +36,16 @@ export function switchMap<T, R, O extends ObservableInput<any>>(project: (value:
  * subsequent inner Observables.
  *
  * ## Example
+ * Generate new Observable according to source Observable values
+ * ```javascript
+ * import { of } from 'rxjs';
+ * import { switchMap } from 'rxjs/operators';
+ *
+ * const switched = of(1,2,3).pipe(switchMap((x) => of( x, x ** 2, x ** 3)));
+ * switched.subscribe(x => console.log(x));
+ * ```
+ * 
+ *
  * Rerun an interval Observable on every click event
  * ```javascript
  * import { fromEvent, interval } from 'rxjs';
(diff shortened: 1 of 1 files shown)


src/internal/operators/switchMap.ts
```diff
@@ -36,6 +36,16 @@ export function switchMap<T, R, O extends ObservableInput<any>>(project: (value:
  * subsequent inner Observables.
  *
  * ## Example
+ * Generate new Observable according to source Observable values
+ * ```javascript
```
dzhavat: Can you mark the snippet as `ts` instead? This will make it open in the right file format when the "Edit in StackBlitz" button is clicked.

src/internal/operators/switchMap.ts
```diff
@@ -36,6 +36,16 @@ export function switchMap<T, R, O extends ObservableInput<any>>(project: (value:
...
+ * import { of } from 'rxjs';
+ * import { switchMap } from 'rxjs/operators';
+ *
+ * const switched = of(1,2,3).pipe(switchMap((x: number) => of( x, x ** 2, x ** 3)));
```
dzhavat: Can you add space between the numbers `of(1,2,3)`? Not that it's wrong as it is right now but formatting it looks better.

Also there's an extra space here `of( x, ...`. Again, the example is great but this small change will make it look better.

src/internal/operators/switchMap.ts
```diff
@@ -36,6 +36,16 @@ export function switchMap<T, R, O extends ObservableInput<any>>(project: (value:
...
+ * switched.subscribe(x => console.log(x));
+ * ```
+ *
```
dzhavat: Have you considered adding what the result would look like? That'd be quite valuable. It can be in the format:

```
Outputs
1
1
1
2
4
8
... and so on
```

src/internal/operators/switchMap.ts
xuemind: resolved in new commit

src/internal/operators/switchMap.ts
xuemind: added the outputs in new commit

src/internal/operators/switchMap.ts
xuemind: add the space in `of(1, 2, 3)`

xuemind:
thanks very much~

PR #4637, opened 2019-03-14
refactor: remove BUILD.bazel file

manekinekko:
**Description:**
rename `build_bazel_rules_typescript` to `npm_bazel_typescript` according to https://github.com/bazelbuild/rules_typescript/commit/3e92736af9d6bcae1991a6a6947a20f4bfd8f0de

cc @alexeagle @gregmagolan should this change be considered as a BREAKING CHANGE?

UPDATE: see discussion below.


alexeagle:
I think the BUILD files should just be removed?
We no longer expect bazel users to build rxjs from source, and the rxjs team doesn't use Bazel either.
/cc @gregmagolan

gregmagolan:
> I think the BUILD files should just be removed?
> We no longer expect bazel users to build rxjs from source, and the rxjs team doesn't use Bazel either.
> /cc @gregmagolan

Agreed

manekinekko:
OK. I will update my PR and remove the BUILD files.

PR #4640, opened 2019-03-15
feat(forkJoin): accepts a dictionary of sources

benlesh:
- Adds `forkJoin({ foo, bar })` such that it will return an `Observable<{ foo, bar }>` containing the values emitted from the sources held at the properties of the passed object.
- DEPRECATED: `forkJoin(a, b, c, d)` should no longer be used, rather, just pass an array such as `forkJoin([a, b, c, d])`.
- Reduces the size of the fork join code a bit
Been wanting to do this one for a while. Going to try to get `combineLatest` also. (Inspired to get to it by @cartant :heart:)

### DIFF
--- a/spec/observables/forkJoin-spec.ts
+++ b/spec/observables/forkJoin-spec.ts
@@ -2,6 +2,7 @@ import { expect } from 'chai';
 import { Observable, forkJoin, of } from 'rxjs';
 import { lowerCaseO } from '../helpers/test-helper';
 import { hot, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
+import { AssertionError } from 'assert';
 
 declare const type: Function;
 declare const asDiagram: Function;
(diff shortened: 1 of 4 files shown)


spec/observables/forkJoin-spec.ts
```diff
@@ -67,284 +68,441 @@ describe('forkJoin', () => {
...
-  type('should support observables', () => {
-    /* tslint:disable:no-unused-variable */
-    let a: Observable<number>;
-    let b: Observable<string>;
-    let c: Observable<boolean>;
-    let o1: Observable<[number, string, boolean]> = forkJoin(a, b, c);
-    let o2: Observable<boolean> = forkJoin(a, b, c, (aa: number, bb: string, cc: boolean) => !!aa && !!bb && cc);
-    /* tslint:enable:no-unused-variable */
-  });
+    it('should accept array of observable contains single', () => {
```
cartant: This test description doesn't appear to describe what's tested.

spec/observables/forkJoin-spec.ts
```diff
@@ -67,284 +68,441 @@ describe('forkJoin', () => {
...
+        bar: Promise.resolve(2)
+      });
+
+      e1.subscribe({
+        next: x => expect(x).to.deep.equal({ foo: 1, bar: 2 }),
+        complete: done,
+      });
+    });
+
+    it('should accept array of observables', () => {
```
cartant: Mismatched test description.

PR #4665, opened 2019-03-26
test(all): modernize all tests

### LINKED ISSUE #4633: Identify all tests that have not been updated to proper imports and "run mode"
# Just a list, not the fixes.

The work here is simple: Create a list or spreadsheet we can edit that tracks all of the tests in our suite that need to be updated in two different ways:

1. Imports need to be "modernized"
2. TestScheduler usage needs to be using "run mode".

### Identify tests without "modernized" imports

A lot of work was already done here, but there's still some work that needs to be done.  Basically, we need to update tests to make sure they are all importing from `rxjs`, `rxjs/operators`, `rxjs/testing`, etc, and **NOT** importing like `import * as Rx from 'rxjs'`. 

We also need to make sure all operator usage is "pipeable". There are still a few tests where they're being "dot-chained" which isn't what we want.

### Identify tests that need to use "run mode"

Run mode is testing with `TestScheduler.prototype.run` like so:

```ts
let testScheduler: TestScheduler;

beforeEach(() => {
  // "defaultObservableComparar" isn't a real thing, I forgot what it's called :P haha.
  testScheduler = new TestScheduler(defaultObservableComparer); 
});

it('should be a test like this', () => {
  testScheduler.run(({ hot, cold, expectObservable }) => {
     const source = hot('---a--b--c--d--|');
     const subs = '     -----^----------!';
     const expected = '      --b--c--d--|';
     expect(source, subs).toBe(expected);
  });
});
```

### Expected result

A list we can use to track progress on updating these files.

### Impact

It will drive a *LOT* of work for people, particularly people new to contributing to the code base. This work will enable us to more cleanly move to RxJS v 7 and beyond!

Moving to using "run mode" will enable us to use a code formatter (such as Prettier or Clang format) on the library, because we won't have to worry about jacking up marble diagrams anymore (because we can use spaces in the strings more effectively). This will help others contribute to the codebase by making it more readable.

benlesh:
- disables tests that relied on polluting the global  instance, thus causing all tests that were relying on legacy dot-chaining or patched creation methods to fail to build.
- updates all tests to use pipeable operators and appropriate creation functions.

resolves #4633


spec/operators/publish-spec.ts
```diff
@@ -175,7 +175,7 @@ describe('publish operator', () => {
     it('should disconnect when last subscriber unsubscribes', () => {
       const source =     cold(   '-1-2-3----4-|');
       const sourceSubs =      '   ^        !   ';
-      const replayed = source.publish().refCount();
+      const replayed = (source.pipe(publish()) as ConnectableObservable<string>).refCount();
```
cartant: Is there are particular reason for not using the `refCount` operator?

benlesh: I just missed it, honestly. It's a bit of a special case, as you can see.

benlesh:
```suggestion
      const replayed = source.pipe(publish(), refCount());
```

PR #4677, opened 2019-03-30
chore(lint): add no-const-enum rule

cartant:
**Description:**

This PR enables the `no-const-enum` rule that's in `tslint-etc`.

ATM, `tslint-etc` is used in the dtslint tests to catch malformed type expectations. This PR bumps the `tslint-etc` version and enables the above-mentioned rule in the project's `tslint.json` file.

The reason const enums cannot be used is outlined in #4556. Basically, using a const enum prohibits isolated module transpilation, as it's necessary to refer to another file to determine the literal that should replace a reference to a const enum member. That means that versions of RxJS that export const enums cannot be used with `create-react-app`-generated projects or any other projects that build using isolated modules.

As configured, the rule forbids all use of const enums. The rule does have an `allowLocal` option, but I don't see the use of local (i.e. within a single module) const enums to be a likely use case.

The main reason for adding the rule is to prevent the sort of breaking change that happened in 6.4.0 from happening again. In #4556, I mentioned that I tried to include a test that compiled the project using isolated modules, but was unable to do so (the reason is given in the linked PR).

The linting in this PR will fail until #4556 is merged.

**Related issue (if exists):** #4556

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -283,6 +283,26 @@
         "url-template": "^2.0.8"
       }
     },
+    "@phenomnomnominal/tsquery": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/@phenomnomnominal/tsquery/-/tsquery-3.0.0.tgz",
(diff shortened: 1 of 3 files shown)


cartant:
Aarrgghh! I rebased and force pushed the branch that I had on this notebook and managed to wipe my commits, as this notebook must have had a branch with no work in it. Or something. Will have to fix this later, when I get home. Hopefully, the work will be on my other notebook.

Issue #4675, opened 2019-03-29
[docs]: Operator `repeat` is missing documentation, example, and see also


BioPhoton:
### Please check those that apply

- [ ] typo
- [x] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [x] needs example

### Description Of The Issue

This operator needs documentation, example and the see Alos section.
We might want to include it.

For See Also we could include `repeatWhen`

JWO719:
done by #4681

PR #4681, opened 2019-04-02
docs(repeat): Add repeat documentation, example, and see also

honzalo:
**Description:**
Updated documentation for `repeat` operator.

#4675

### DIFF
--- a/src/internal/operators/repeat.ts
+++ b/src/internal/operators/repeat.ts
@@ -7,7 +7,51 @@ import { MonoTypeOperatorFunction, TeardownLogic } from '../types';
 /**
  * Returns an Observable that will resubscribe to the source stream when the source stream completes, at most count times.
  *
+ * <span class="informal">Repeats all values emitted on the source. It's like {@link retry}, but for non error cases.</span>
+ *
  * ![](repeat.png)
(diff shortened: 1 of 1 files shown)


src/internal/operators/repeat.ts
```diff
@@ -7,8 +7,41 @@ import { MonoTypeOperatorFunction, TeardownLogic } from '../types';
...
+ * ## Example
+ * Repeat a message stream
+ * ```ts
+ * import { of } from 'rxjs';
+ * import { repeat, delay } from 'rxjs/operators';
+ *
+ * const source = of('Repeat message');
+ * const example = source.pipe(repeat(3));
+ * example.subscribe(x => console.log(x));
+ * // Results: Repeat message Repeat message Repeat message
```
dzhavat: First of all, great contribution! 

I'd like to suggest one tiny improvement. Don't want to sound picky but wouldn't it be better if each output was on its own line? Because in reality that is what is happening.

honzalo: Yes, agree. I have pushed a change to have them on its own line and reduce the second example so that it takes less lines but is still comprehensible.

PR #4683, opened 2019-04-03
feat(retry): add config to reset error count on successful emission

omermizr:
This PR adds the ability to reset the error counter on successful emissions using the `retry`
operator. 

**Description:**
The current behavior for `retry(n)` is to call error if n errors occurred, regardless of
whether or not they were consecutive. Now one would be able to use `retry(n, true)` to have the
count reset so that only n consecutive errors will cause the observable to fail.


benlesh:
@cartant said we can have this. :+1:

omermizr:
Something to consider (breaking changes aside ): I think that `resetOnSuccess` should default to `true`. 
I mainly use `retry` for polling mechanisms, and I suspect that's a pretty common usage. In that use case, I would most likely want to throw an error only if the stream fails consistently and consecutively (because that probably means there's some issue that isn't transient) - @benlesh WDYT?

omermizr:
@benlesh ping

package.json
```diff
@@ -157,7 +157,7 @@
     "@angular-devkit/schematics": "^7.3.9",
     "@types/chai": "4.1.2",
     "@types/lodash": "4.14.102",
-    "@types/mocha": "2.2.48",
+    "@types/mocha": "5.0.0",
```
benlesh: Probably don't need to update mocha as part of this change.

spec/Observable-spec.ts
```diff
@@ -18,7 +18,10 @@ function expectFullObserver(val: any) {
 /** @test {Observable} */
 describe('Observable', () => {
   let originalConfigPromise: any;
-  before(() => originalConfigPromise = config.Promise);
+  before(() => {
```
benlesh: This change is unnecessary for the PR.

package.json
omermizr: It's only a **types** update. 
The old types didn't allow passing a message to the `done` function indicating what caused the failure (not to mention the version mismatch we previously had).
Any reason not to make this change (other than separation of changes)?

spec/Observable-spec.ts
omermizr: I had to make this change because of the mocha types update.
The `before` function should return `void`, but here it returns a promise so it failed my build.
If we decide to leave the types change out, then this is indeed unnecessary and the previous comment's resolution will also be this one's.

omermizr:
@benlesh ping

cartant:
Closed in favour of #5280 - into which your commits have been cherry picked, @omermizr.

Issue #4631, opened 2019-03-13
[docs]: Operator `skipUntil` is missing the description and example


BioPhoton:
### Please check those that apply

- [ ] typo
- [x] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [x] needs example

### Description Of The Issue
 We should include a proper description as well as an example

JWO719:
closed by #4696

PR #4696, opened 2019-04-08
docs(skipUntil): add description and example with an interval waiting for click

ngfelixl:
**Description:**

The example contains two base observables. An interval and a click listener. The resulting observable is the interval but skipping all the values, until the click event is fired.

#4631

### DIFF
--- a/src/internal/operators/skipUntil.ts
+++ b/src/internal/operators/skipUntil.ts
@@ -10,7 +10,34 @@ import { Subscription } from '../Subscription';
 /**
  * Returns an Observable that skips items emitted by the source Observable until a second Observable emits an item.
  *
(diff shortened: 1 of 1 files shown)


src/internal/operators/skipUntil.ts
```diff
@@ -10,8 +10,35 @@ import { Subscription } from '../Subscription';
 /**
  * Returns an Observable that skips items emitted by the source Observable until a second Observable emits an item.
  *
+ * The `skipUntil` operator causes the observable stream to skip the emission of values ​​until the passed in observable returns the first value.
```
JWO719: `observable returns the first value.` I think it would be clearer if it says `observable emits the first value`

ngfelixl: Thanks @JWO719. I've implemented your requested change.

Issue #4647, opened 2019-03-16
Operator `throttleTime` is missing an example for the throttleConfig param


BioPhoton:
### Please check those that apply

- [ ] typo
- [ ] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [x] needs example

### Description Of The Issue

We may want to include an example for throttleConfig param.

kosich:
Would it make sense to also mention that configurability in the description?

Current description totally disregards that:

> Emits a value from the source Observable, then ignores subsequent source values for duration milliseconds, then repeats this process.

We could add, e.g.

> Can be configured to ignore the first value and|or also emit the last value within given duration.

I personally missed that option, probably due to rather strict description of the behavior.

Nosfistis:
The new `throttleTime` behavior with `{ leading: true, trailing: true }` is not documented, neither in the operator page (https://rxjs.dev/api/operators/throttleTime) nor in the v7 breaking changes page. The operator marble image also still shows the old behavior, where both the last and the first event is emitted, even if the time between them is smaller than the throttle time. I noticed the new behavior through unit testing.

CroMarmot:
Same issue, the graph shows 
![image](https://github.com/ReactiveX/rxjs/assets/24691835/37c9362b-d193-40b5-bc4c-74756d0b167b)

while the api is `throttleTime<T>(duration: number, scheduler: SchedulerLike = asyncScheduler, config?: ThrottleConfig): MonoTypeOperatorFunction<T>`

https://rxjs.dev/api/operators/throttleTime

benlesh:
Hi @BioPhoton — thank you for taking the time to open this issue and for sharing your thoughtful feedback with the RxJS project. I’m sorry it took us so long to respond. Given how much time has passed and the lack of recent activity, I’m going to close this for now so the tracker reflects the work we’re likely to act on. If this is still relevant on a current RxJS release, please feel welcome to open a new issue with updated context and a reproduction. Thank you again for your contribution and patience.

— Ben

PR #4698, opened 2019-04-08
docs(throttleTime): add double click example making use of throttle-config

ngfelixl:
**Description:**

The example contains the simulation of double-clicks that happen within a time-range of 400ms.

#4647

### DIFF
--- a/src/internal/operators/throttleTime.ts
+++ b/src/internal/operators/throttleTime.ts
@@ -24,7 +24,10 @@ import { MonoTypeOperatorFunction, SchedulerLike, TeardownLogic } from '../types
  * and this process repeats for the next source value. Optionally takes a
  * {@link SchedulerLike} for managing timers.
  *
- * ## Example
+ * ## Examples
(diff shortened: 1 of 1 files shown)


JWO719:
The pipeline failed due to dtslint, but that's not your fault. We take care of it and let you know. Thanks for your contribution!

benlesh:
This is a cool contribution. I think my only concern is that people should probably be using the `dblclick` event which is supported by all browsers, TMK.

ngfelixl:
@benlesh you are right. Do you think I should add a hint for it like

```ts
/*
 * [...]
 * #### Double Click
 *
 * For common double click tasks use the build in `dbclick` event.
 * The following example only emits clicks which happen within a subsequent
 * delay of 400ms of the previous click. This for example can emulate a double
 * click. It makes use of the `trailing` parameter of the throttle configuration.
 * 
 * [...]
 */
```

or would it be better to think about a completely new example?

Issue #4534, opened 2019-02-06
Rewrite ajax module to use abortable `fetch`


OliverJAsh:
## Feature Request

The ajax module is currently built on top of `XHR`. IIUC, this is because `fetch` did not initially support cancellation (and maybe also for browser support?). However, [`fetch` does not support cancellation](https://developers.google.com/web/updates/2017/09/abortable-fetch).

There might be benefits to using `fetch` under the hood instead of `XHR`. One that I've just stumbled across is, when using `XHR`, there is [no way to reliably detect network errors](https://github.com/ReactiveX/rxjs/issues/1926#issuecomment-246131172), whereas [with `fetch` there is](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Checking_that_the_fetch_was_successful).

Is this something we should consider? Or perhaps it should exist in user land.

kwonoj:
Dupe of https://github.com/ReactiveX/rxjs/issues/4128. We have long term plan but not able to polyfill abortcontroller makes whole rewrite will take time to happen. My userland module https://github.com/kwonoj/oxid/issues/71 plan to have it in userlans first.

OliverJAsh:
Related: `abortable-rx` allows you to use `fetch` inside `switchMap` https://github.com/felixfbecker/abortable-rx/tree/0b60bfc80e61904e720f99e9ca8c662c72dc4f51#using-fetch-inside-switchmap

kwonoj:
This is achieved, closing. Each issue may need to be filed separately if found.

PR #4702, opened 2019-04-10
feat(fetch): Making fetch happen

benlesh:
- Adds `fromFetch` implementation that uses native `fetch`
- Adds tests and basic documentation
- DOES NOT polyfill fetch

### Overview

I think, long term, we should deprecate the `ajax` methods and try to move people towards fetch now that both `fetch` and `AbortController` are universally implemented (everywhere but IE) and polyfills are available.

Maintaining our own ajax implementation is challenging, and this is lot more lightweight and full featured.

### DIFF
--- a/spec/observables/dom/fetch-spec.ts
+++ b/spec/observables/dom/fetch-spec.ts
@@ -1,0 +1,195 @@
+import { fromFetch } from 'rxjs/fetch';
+import { expect } from 'chai';
+import { root } from '../../../src/internal/util/root';
+
+const OK_RESPONSE = {
+  ok: true,
+} as Response;
+
+function mockFetchImpl(input: string | Request, init?: RequestInit): Promise<Response> {
+  (mockFetchImpl as MockFetch).calls.push({ input, init });
+  return new Promise<any>((resolve, reject) => {
+    if (init.signal) {
+      init.signal.addEventListener('abort', () => {
+        console.log('triggered');
+        reject(new MockDOMException());
+      });
+    }
+    return Promise.resolve(null).then(() => {
+      console.log('resolved');
+      resolve((mockFetchImpl as any).respondWith);
+    });
+  });
+}
+(mockFetchImpl as MockFetch).reset = function (this: any) {
+  this.calls = [] as any[];
+  this.respondWith = OK_RESPONSE;
+};
+(mockFetchImpl as MockFetch).reset();
+
+const mockFetch: MockFetch = mockFetchImpl as MockFetch;
+
+class MockDOMException {}
+
+class MockAbortController {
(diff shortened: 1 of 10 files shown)


benlesh:
### From Core Team Meeting

- Node concerns: Node users are probably just going to use whatever they're currently using
- AbortController isn't something that can be easily polyfilled, but exists in all modern browers.
- We should add warnings to the documentation that parts of the fetch API are experimental.
- It is published under `rxjs/fetch` which separates it from the other functionality a bit. Also document that this is a browser-only API.
- Perhaps test it with `node-fetch` just to see/make sure it works?

OliverJAsh:
🎉 I think this will close https://github.com/ReactiveX/rxjs/issues/4534

spec/observables/dom/fetch-spec.ts
```diff
@@ -0,0 +1,195 @@
...
+    expect(mockFetch.calls[0].init.signal.aborted).to.be.false;
+
+    subscription.unsubscribe();
+    expect(mockFetch.calls[0].init.signal.aborted).to.be.true;
+  });
+
+  it('should allow passing of init object', () => {
+    const myInit = {};
+    const fetch$ = fromFetch('/foo', myInit);
+    fetch$.subscribe();
```
cartant: Unlike the previous test - which calls `unsubscribe` - this one will have code that continues to execute after the test returns - as a `done` callback isn't used. That won't matter unless something gets broken in the implementation, but if it does, an error from this test will appear to come from someplace else.

I would pass a `done` callback here for `error` and `complete`.

src/internal/observable/dom/fetch.ts
```diff
@@ -0,0 +1,61 @@
...
+    } else {
+      init = { signal };
+    }
+
+    fetch(input, init).then(response => {
+      if (response.ok) {
+        subscriber.next(response);
+        subscriber.complete();
+      } else {
+        subscriber.error(new Error(`RxJS Fetch HTTP Error\n\nStatus:\n${response.status}\n\nBody:\n${response.body}\n`));
```
cartant: IMO, it would be a mistake to error on non-OK responses. This is not what `fetch` does; `fetch` only rejects if a network error occurs.

I guess the question is: should this align with `fetch`'s non-OK response handling or should it be like Angular and turn non-OK responses into errors.

I'm more inclined to go with the former, but I can see why others might favour the latter. However, if it is going to use the latter approach, IMO, it *must* make the error response available via the error.

benlesh: Thinking about this more, I think this should mimic `fetch` as much as possible. If that is going to be the defacto standard going forward.

PR #4704, opened 2019-04-10
feat(Subscription): update logics of subscription relationship

githubxiaowen:
**Description:**
use `_parents` and `_subscriptions` to handle the management of subscriptions


cartant:
I'm not at all sure that this is necessary. Benedikt Meurer reviews this code in his PR: https://github.com/ReactiveX/rxjs/pull/4526

benlesh:
Yeah. This is going to result in additional array allocations. I like where this is going, but I'm not sure it's necessary either.

Issue #4697, opened 2019-04-08
concat example does not work with array


joriyann:
### Documentation Related To Component:

concat static operator in `src/internal/observable/concat.ts`

### Please check those that apply

- [ ] typo
- [ ] documentation doesn't exist
- [ ] documentation needs clarification
- [x] error(s) in example
- [ ] needs example

### Description Of The Issue

The exemple `Concatenate an array of 3 Observables` in the concat documentation page at https://rxjs.dev/api/index/function/concat does not output the expected result.

#### Expected result
```javascript
// results in the following:
// (Prints to console sequentially)
// -1000ms-> 0 -1000ms-> 1 -1000ms-> ... 9
// -2000ms-> 0 -2000ms-> 1 -2000ms-> ... 5
// -500ms-> 0 -500ms-> 1 -500ms-> ... 9
```

#### Current result in Stackblitz
```javascript
// Observable {_isScalar: false, source: {…}, operator: {…}}
// Observable {_isScalar: false, source: {…}, operator: {…}}
// Observable {_isScalar: false, source: {…}, operator: {…}}
```

#### Additional information
Passing the array with the rest operator instead of the array itself output the expected result.
```javascript
const result = concat([timer1, timer2, timer3]); // note that array is passed // does not work
const result = concat(...[timer1, timer2, timer3]); // note that array is passed // works
```

dzhavat:
Thanks for reporting it. There's already an older issue (#4203) that is kind of the same.

dzhavat:
Sure, you can close it or, if you're up to it, you can open a PR with a fix. What do you think about it?

alex-shamshurin:
The same issue is here.

Polyterative:
I'm fighting with this too. I'm trying to make an array of network request in series and I'm not finding any way to do it

robertbjarum:
Thanks for reporting this issue. I ran into the same problem and finally found this and could fix my code using the rest operator.

PR #4713, opened 2019-04-13
docs(concat): clarify array as input

joriyann:
**Description:**
Since an array is itself an ObservableInput and so considered as an observable source, this specifies the use of spread operator to pass an array of observables.

#4697

### DIFF
--- a/src/internal/observable/concat.ts
+++ b/src/internal/observable/concat.ts
@@ -67,7 +67,7 @@ export function concat<R>(...observables: (ObservableInput<any> | SchedulerLike)
  * // 0 -1000ms-> 1 -1000ms-> 2 -1000ms-> 3 -immediate-> 1 ... 10
  * ```
  *
- * ### Concatenate an array of 3 Observables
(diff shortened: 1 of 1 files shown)


src/internal/observable/concat.ts
```diff
@@ -75,7 +75,10 @@ export function concat<R>(...observables: (ObservableInput<any> | SchedulerLike)
  * const timer1 = interval(1000).pipe(take(10));
  * const timer2 = interval(2000).pipe(take(6));
  * const timer3 = interval(500).pipe(take(10));
- * const result = concat([timer1, timer2, timer3]); // note that array is passed
+ *
+ * // an array itself being considered an observable source,
+ * // use the spread operator to observe its values instead of itself
+ * const result = concat(...[timer1, timer2, timer3]); // note that array is spread
```
benlesh: Honestly, this makes me wonder if there was a regression at some point where we used to take a single array. 🤔 

Either way, for now, it's probably best just to pass them in without spreading them from an array, because it doesn't make sense.

PR #4720, opened 2019-04-23
chore(refCount): remove unnecessary reference of object

githubxiaowen:
**Description:**
Since the `source` and the `connectableObservable` is the same thing in RefCountOperator, there is no need to hold the variable `connectable` int the RefCountOperator
NONE

### DIFF
--- a/src/internal/observable/ConnectableObservable.ts
+++ b/src/internal/observable/ConnectableObservable.ts
@@ -12,8 +12,8 @@ import { refCount as higherOrderRefCount } from '../operators/refCount';
 export class ConnectableObservable<T> extends Observable<T> {
 
   protected _subject: Subject<T>;
-  private _refCount: number = 0;
-  private _connection: Subscription;
+  protected _refCount: number = 0;
+  protected _connection: Subscription;
   /** @internal */
   _isComplete = false;
 
--- a/src/internal/observable/ConnectableObservable.ts
+++ b/src/internal/observable/ConnectableObservable.ts
@@ -12,8 +12,8 @@ import { refCount as higherOrderRefCount } from '../operators/refCount';
 export class ConnectableObservable<T> extends Observable<T> {
 
   protected _subject: Subject<T>;
-  protected _refCount: number = 0;
-  protected _connection: Subscription;
+  private _refCount: number = 0;
+  private _connection: Subscription;
   /** @internal */
   _isComplete = false;
 
--- a/src/internal/operators/refCount.ts
+++ b/src/internal/operators/refCount.ts
@@ -60,20 +60,18 @@ import { Observable } from '../Observable';
  */
 export function refCount<T>(): MonoTypeOperatorFunction<T> {
   return function refCountOperatorFunction(source: ConnectableObservable<T>): Observable<T> {
-    return source.lift(new RefCountOperator(source));
+    return source.lift(new RefCountOperator());
   } as MonoTypeOperatorFunction<T>;
 }
 
 class RefCountOperator<T> implements Operator<T, T> {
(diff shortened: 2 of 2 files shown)


benlesh:
I'm not sure how I feel about exposing all of these properties as public. They really aren't meant to be consumed by other users.

src/internal/observable/ConnectableObservable.ts
```diff
@@ -12,8 +12,8 @@ import { refCount as higherOrderRefCount } from '../operators/refCount';
 export class ConnectableObservable<T> extends Observable<T> {
 
   protected _subject: Subject<T>;
-  protected _refCount: number = 0;
-  protected _connection: Subscription;
+  public _refCount: number = 0;
+  public _connection: Subscription;
```
benlesh: We don't want to make these public APIs really.

src/internal/operators/refCount.ts
```diff
@@ -32,7 +30,7 @@ class RefCountOperator<T> implements Operator<T, T> {
 
 class RefCountSubscriber<T> extends Subscriber<T> {
 
-  private connection: Subscription;
+  public connection: Subscription;
```
benlesh: Revert, please.

src/internal/observable/ConnectableObservable.ts
githubxiaowen: already changed

src/internal/operators/refCount.ts
githubxiaowen: already changed

src/internal/observable/ConnectableObservable.ts
```diff
@@ -12,8 +12,8 @@ import { refCount as higherOrderRefCount } from '../operators/refCount';
 export class ConnectableObservable<T> extends Observable<T> {
 
   protected _subject: Subject<T>;
-  protected _refCount: number = 0;
-  protected _connection: Subscription;
+  private _refCount: number = 0;
+  private _connection: Subscription;
```
benlesh: This is technically a breaking change for TypeScript users.

src/internal/operators/refCount.ts
```diff
@@ -60,20 +60,18 @@ import { Observable } from '../Observable';
  */
 export function refCount<T>(): MonoTypeOperatorFunction<T> {
   return function refCountOperatorFunction(source: ConnectableObservable<T>): Observable<T> {
-    return source.lift(new RefCountOperator(source));
+    return source.lift(new RefCountOperator());
   } as MonoTypeOperatorFunction<T>;
 }
 
 class RefCountOperator<T> implements Operator<T, T> {
-  constructor(private connectable: ConnectableObservable<T>) {
```
benlesh: Removing the constructor is also a breaking change.

src/internal/observable/ConnectableObservable.ts
cartant: Yeah, as this would be a breaking change, I don't see any advantage to changing these to `private`. Especially as this is a 'perf' PR. This is unrelated to perf.

src/internal/operators/refCount.ts
cartant: I'm fine with this. AFAICT, this is not a breaking change as `RefCountOperator` is not exported. And if someone has reached into the internals to get it and this breaks them, then that's on them for being naughty.

src/internal/observable/ConnectableObservable.ts
githubxiaowen: reverted to protected

githubxiaowen:
Yeah，i gave it a second thought. This commit indeed is just removing  the unused reference of `source`， nothing about performance.

Issue #4723, opened 2019-04-23
deprecation warning when using of as intended


evertbouw:
## Bug Report

**Current Behavior**

I've got the tslint rule `"deprecation": true` enabled and it is now throwing warnings when using `of` without a scheduler.

**Reproduction**
```ts
import { of } from "rxjs"

of(4) // of is deprecated: use {@link scheduled} instead `scheduled([a, b, c], scheduler)` (deprecation)
```

**Expected behavior**
No warning

**Environment**
- Runtime: Node v11.13.0
- RxJS version: 6.5.1
- tslint version 5.16.0

dmytro-gokun:
As described here, only the version of "of" which accepts a scheduler should be deprecated: https://github.com/ReactiveX/rxjs/pull/4595 So, it's definitely a bug.

evertbouw:
I think these two lines should be moved to the bottom of the declaration file

```ts
/** @deprecated use {@link scheduled} instead `scheduled([a, b, c], scheduler)` */
export declare function of<T>(...args: (T | SchedulerLike)[]): Observable<T>;
```

cartant:
@evertbouw Yes, they need to be placed below:

```ts
export function of<T>(...args: T[]): Observable<T>;
```

cartant:
@evertbouw If you would like to open a PR, please let me know. Otherwise, I'll open one myself.

evertbouw:
@cartant I can do it

nlaplante:
I see the same problem with `startWith`

cartant:
@nlaplante If there is a problem with `startWith` it would be better to open another issue. This one will be closed when the PR is merged. Anyway, if there is a problem with that operator's deprecations, it appears to be to be different to this issue, as the signature is [already in the correct place](https://github.com/ReactiveX/rxjs/blob/1dbbafad6abe4be5e0ff992014217c98389b300b/src/internal/operators/startWith.ts#L29-L30).

imransilvake:
I have the same problem.

imransilvake:
> I see the same problem with `startWith`

I had the probem with `startWith` when I was using `startWith(void 0)`. I changed it to `startWith(0)` and now it works without showing any deprecation warning.

cartant:
@imransilvake As mentioned above, please open a new issue and fill out the issue template.

yogeshgadge:
@nlaplante  Temporary workaround - use `startWith`  as below:-

```
startWith<string>(null)
[...]

PR #4724, opened 2019-04-23
fix(of): deprecation warning when using of as intended

evertbouw:
**Description:**

Typescript picks the deprecated signature when using `of`. ~~By moving the signature to the bottom it will pick the correct one.~~ Removed the deprecation comment instead

closes https://github.com/ReactiveX/rxjs/issues/4723

### DIFF
--- a/src/internal/observable/of.ts
+++ b/src/internal/observable/of.ts
@@ -26,7 +26,6 @@ export function of<T, T2, T3, T4, T5, T6, T7, T8>(a: T, b: T2, c: T3, d: T4, e: 
 /** @deprecated use {@link scheduled} instead `scheduled([a, b, c], scheduler)` */
 export function of<T, T2, T3, T4, T5, T6, T7, T8, T9>(a: T, b: T2, c: T3, d: T4, e: T5, f: T6, g: T7, h: T8, i: T9, scheduler: SchedulerLike):
   Observable<T | T2 | T3 | T4 | T5 | T6 | T7 | T8 | T9>;
-/** @deprecated use {@link scheduled} instead `scheduled([a, b, c], scheduler)` */
 export function of<T>(...args: (T | SchedulerLike)[]): Observable<T>;
 
 // TODO(benlesh): Update the typings for this when we can switch to TS 3.x


src/internal/observable/of.ts
```diff
@@ -45,6 +43,9 @@ export function of<T, T2, T3, T4, T5, T6, T7, T8, T9>(a: T, b: T2, c: T3, d: T4,
 export function of<T>(...args: T[]): Observable<T>;
 /* tslint:enable:max-line-length */
 
+/** @deprecated use {@link scheduled} instead `scheduled([a, b, c], scheduler)` */
+export function of<T>(...args: (T | SchedulerLike)[]): Observable<T>;
```
cartant: I would move it to just above the `/* tslint:enable:max-line-length */` line. Throughout the codebase, those lint flags wrap the overload signatures, so this file should stick with that convention.

cartant:
@mtraynham If there is a specific problem with the operators you are talking about, please open another issue and fill out the template with the specifics. As it stands, I don't have all of the information and I don't want to spend anytime looking into this based on what I've assumed you are talking about.

Note that the `concat` and `merge` *operators* are deprecated for a different reason.

src/internal/observable/of.ts
```diff
@@ -43,6 +41,9 @@ export function of<T, T2, T3, T4, T5, T6, T7, T8>(a: T, b: T2, c: T3, d: T4, e:
 export function of<T, T2, T3, T4, T5, T6, T7, T8, T9>(a: T, b: T2, c: T3, d: T4, e: T5, f: T6, g: T7, h: T8, i: T9):
   Observable<T | T2 | T3 | T4 | T5 | T6 | T7 | T8 | T9>;
 export function of<T>(...args: T[]): Observable<T>;
+
+/** @deprecated use {@link scheduled} instead `scheduled([a, b, c], scheduler)` */
```
cartant: Actually, I've had another look at this and I don't think the signature can be moved. If it is, usage of a scheduler will match the signature above it and `T` will be inferred to be a union that includes the scheduler's type.

I think the signature will have to be left in its original location and that the deprecation for this particular signature should be removed instead.

KurtGokhan: @cartant how about using Typescript Exclude operator in non-deprecated overloads? i.e. line 43 would look like this:

`export function of<T extends Exclude<any, SchedulerLike>>(...args: T[]): Observable<T>;`

KurtGokhan: You are right, seems like Negated types are coming in Typescript 3.5.

PR #4727, opened 2019-04-23
fix(throttleTime): double emission with leading and trailing enabled

### LINKED ISSUE #2466: Bug? Perhaps improper behavior of throttle(fn)

### LINKED ISSUE #2727: throttleTime(n, undefined, {leading: true, trailing: true}) ... double output

MatthiasKunnen:
This PR fixes an issue with throttleTime emitting both leading and trailing values in the same time window.
    
Double emission problem:
```
source:   a123b12-c-23d-2-ef---
expected: a---b---c---d---e---f
actual:   a---b1---c2---2-e---f
```
    
Closes #2466 and closes #2727.
Follows #2749 and #2864.

### DIFF
--- a/spec/operators/throttleTime-spec.ts
+++ b/spec/operators/throttleTime-spec.ts
@@ -141,16 +141,44 @@ describe('throttleTime operator', () => {
   });
 
   describe('throttleTime(fn, { leading: true, trailing: true })', () => {
-    asDiagram('throttleTime(fn, { leading: true, trailing: true })')('should immediately emit the first and last values in each time window', () =>  {
-      const e1 =   hot('-a-xy-----b--x--cxxx--|');
-      const e1subs =   '^                     !';
-      const t =  time( '----|                  ');
-      const expected = '-a---y----b---x-c---x-|';
+    asDiagram('throttleTime(fn, { leading: true, trailing: true })')('should immediately emit the first and last values in each time window', () => {
+      const e1 =   hot('a123b12-c-23d-2-ef---|');
+      const t =   time('----|                 ');
+      const expected = 'a---b---c---d---e---f|';
 
       const result = e1.pipe(throttleTime(t, rxTestScheduler, { leading: true, trailing: true }));
 
       expectObservable(result).toBe(expected);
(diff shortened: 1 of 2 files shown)


benlesh:
I'm not sure this is the correct behavior. Even if it was, this would be a breaking change for someone.

I'll wait for other core team members to weigh in... @JWO719? @cartant? @kwonoj?

MatthiasKunnen:
I think you've already had a discussion about whether the behavior is intended in #2749. See https://github.com/ReactiveX/rxjs/pull/2749#discussion_r141075604.

That PR was closed by the author. Perhaps they didn't have the time.

MatthiasKunnen:
I've fixed the fact that slow user code could influence timings.

I understand that time is finite and this PR might not be anywhere on the priority list but would a review be possible?

I've also rebased the PR to be up-to-date.

adolgoff:
PR is definitely fixing unexpected behavior

PR #4722, opened 2019-04-23
docs(fromFetch): typo in docs

elie29:
**Description:**
Fixed two typos in doc blocs

**Related issue (if exists):**

PR #4734, opened 2019-04-26
docs(fetch): typofix

revolunet:
Just a typofix in an example


cartant:
Thanks, but this has been dealt with in PR https://github.com/ReactiveX/rxjs/pull/4722

Issue #4731, opened 2019-04-24
Is this implementation of endWith correct


CarstenLeue:
## Bug Report

rxjs 6.5.1 implements endWith as follows: 

```
export function endWith<T>(...array: Array<T | SchedulerLike>): MonoTypeOperatorFunction<T> {
  return (source: Observable<T>) => concat(source, ...(array as any[])) as Observable<T>;
}
```

Is this really correct given that `concat` expects `observables` as input, but `array` carries the individual items?

daka1510:
I've also hit issues with the `endWith` operator. Here's a simple jest test case that works fine with 6.4.0 but fails with 6.5.1.

```
import { of } from 'rxjs';
import { endWith, tap } from 'rxjs/operators';

it('endWith operator: sample failing test', () => {
  return of('1')
    .pipe(
      endWith({}),
      tap((value) => console.log(value))
    )
    .toPromise();
});
```

Here's the console output when running this test:
```
 console.log src/endWith.spec.ts:8
1


TypeError: You provided an invalid object where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.

at Object.<anonymous>.exports.subscribeTo (/node_modules/rxjs/src/internal/util/subscribeTo.ts:27:11)
at Object.subscribeToResult (/node_modules/rxjs/src/internal/util/subscribeToResult.ts:28:10)
at MergeMapSubscriber.Object.<anonymous>.MergeMapSubscriber._innerSub (/node_modules/rxjs/src/internal/operators/mergeMap.ts:148:5)
at MergeMapSubscriber.Object.<anonymous>.MergeMapSubscriber._tryNext (/node_modules/rxjs/src/internal/operators/mergeMap.ts:141:10)
at MergeMapSubscriber.Object.<anonymous>.MergeMapSubscriber._next (/node_modules/rxjs/src/internal/operators/mergeMap.ts:125:12)
at MergeMapSubscriber.Object.<anonymous>.Subscriber.next (/node_modules/rxjs/src/internal/Subscriber.ts:99:12)
at Observable._subscribe (/node_modules/rxjs/src/internal/util/subscribeToArray.ts:9:16)
at Observable.Object.<anonymous>.Observable._trySubscribe (/node_modules/rxjs/src/internal/Observable.ts:238:19)
at Observable.Object.<anonymous>.Observable.subscribe (/node_modules/rxjs/src/internal/Observable.ts:219:14)
[...]

PR #4735, opened 2019-04-26
fix(endWith): wrap args - they are not observables - in of before concatenating

cartant:
**Description:**

This PR spreads the arguments passed to `endWith` into a call to `of` which is then passed to `concat.

The current implementation spreads the args directly into `concat` which is incorrect. The tests pass because strings are iterable and are therefore valid observable inputs. And the strings in the tests each contain only a single character. The test added in this PR uses numbers for values, and their use effects an error from the current, incorrect `endWith` implementation.

**Related issue (if exists):** #4731

### DIFF
--- a/src/internal/operators/endWith.ts
+++ b/src/internal/operators/endWith.ts
@@ -1,5 +1,6 @@
 import { Observable } from '../Observable';
 import { concat } from '../observable/concat';
+import { of } from '../observable/of';
 import { MonoTypeOperatorFunction, SchedulerLike, OperatorFunction } from '../types';
 
 /* tslint:disable:max-line-length */
@@ -62,5 +63,5 @@ export function endWith<T, Z = T>(...array: Array<Z | SchedulerLike>): OperatorF
  * @owner Observable
  */
 export function endWith<T>(...array: Array<T | SchedulerLike>): MonoTypeOperatorFunction<T> {
-  return (source: Observable<T>) => concat(source, ...(array as any[])) as Observable<T>;
+  return (source: Observable<T>) => concat(source, of(...(array as any[]))) as Observable<T>;
 }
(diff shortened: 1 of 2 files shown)


src/internal/operators/endWith.ts
```diff
@@ -62,5 +63,5 @@ export function endWith<T, Z = T>(...array: Array<Z | SchedulerLike>): OperatorF
  * @owner Observable
  */
 export function endWith<T>(...array: Array<T | SchedulerLike>): MonoTypeOperatorFunction<T> {
-  return (source: Observable<T>) => concat(source, ...(array as any[])) as Observable<T>;
+  return (source: Observable<T>) => concat(source, of(...(array as any[]))) as Observable<T>;
```
felixfbecker: Is the `as any[]` completely unavoidable? Without the cast to `any` this would have been caught at compile time

cartant: The assertion is not necessary. I've removed it. Thanks.

benlesh:
Oh wow! What a cool bug that made it through all of those tests! Since `'x'` is an iterable, `concat(obs$, 'x')` is the _same thing_ as `concat(obs$, of('x'))`. 🤣

AlexAegis:
In `v6.5.1` this worked:
```typescript
endWith(of(undefined))
```

In `v6.5.2` this works, just as the changelog says:
```typescript
endWith(undefined)
```

But right now in `v6.5.2` `endWith(undefined)` is marked as deprecaded. Am I doing something wrong or is this a bug?

cartant:
@alex-wilmer It's matching the signature with the scheduler in it. Most likely because you do not have `strictNullChecks` enabled. See [this comment](https://github.com/ReactiveX/rxjs/issues/4776#issuecomment-491559945) and [this comment](https://github.com/ReactiveX/rxjs/issues/4772#issuecomment-491553380).

As mentioned in the latter comment, IMO, the deprecation in the TSDoc should not be removed to avoid the match in situations in which `strictNullChecks` are disabled. Override the lint failure using the mechanism mentioned in that comment.

Issue #4739, opened 2019-04-26
fromFetch throws AbortError


strout:
## Bug Report

**Current Behavior**
`fromFetch` seems to fail with `AbortError` when I would expect it to succeed.

**Reproduction**
- REPL or Repo link:
[Here's an example using fromFetch(...) and from(fetch(...)) that I would expect to have the same result.](https://codepen.io/strout/pen/jRdrQq)

This code throws an AbortError:

```js
fromFetch('https://jsonplaceholder.typicode.com/todos/1')
  .pipe(switchMap(resp => resp.json()))
  .subscribe(json => console.log(json));
```

**Expected behavior**
The snippit above should complete and log the response to the console.

**Environment**
- Runtime: Chrome 74, Firefox 66
- RxJS version: 6.5.1

**Possible Solution**
In the definition of `fromFetch()`, don't call `controller.abort()` if it has already emitted a value. Perhaps set a flag (reuse/rename the `unsubscribed` flag) before calling `subscriber.next`.

felixfbecker:
@cartant how does this happen? Shouldn't the AbortError be ignored if the Observable unsubscribed?

cartant:
It aborts the promise returned by `resp.json()`, IIRC.

felixfbecker:
So in that code sample provided in the issue description, why would that matter?

cartant:
Try the StackBlitz and you'll see. Sometimes it works and sometimes it fails. Bottom line is that `abort` should not be called upon the normal unsubscription that follows completion. See also https://github.com/ReactiveX/rxjs/issues/4744.

PR #4742, opened 2019-04-26
 fix(fromFetch): don't abort if fetch resolves

cartant:
**Description:**

This PR rearranges an existing test so that the expectations are performed after the complete - and subsequent unsubscribe - are finished. The rearranged test fails.

It also adds a flag to the `fromFetch` implementation so that `abort` is not called in the teardown if the promise returned by `fetch` has resolved.

This fixes the problem, but it also highlights a problem with the implementation: streamed data obtained using `fromFetch` cannot be aborted, as the `AbortController` passed to fetch also aborts the promises returned from `text` and `json`, etc.

**Related issue (if exists):** #4739

### DIFF
--- a/spec/observables/dom/fetch-spec.ts
+++ b/spec/observables/dom/fetch-spec.ts
@@ -115,14 +115,19 @@ describe('fromFetch', () => {
         expect(response).to.equal(OK_RESPONSE);
       },
       error: done,
-      complete: done,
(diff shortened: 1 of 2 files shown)


benlesh:
Thanks for picking this one up @cartant!

topaxi:
Just ran into this issue, any chance for a release?

PR #4743, opened 2019-04-26
fix(pairwise): make it recursion-proof

thorn0:
**Description:** `pairwise` calls `next` before saving the current value as the previous one, which makes it not recursively re-enterable. If its subscriber synchronously triggers the input observable, `pairwise` erroneously emits a pair with the wrong previous value.

**Related issue:** https://github.com/reduxjs/redux/issues/3404

### DIFF
--- a/spec/operators/pairwise-spec.ts
+++ b/spec/operators/pairwise-spec.ts
@@ -1,5 +1,7 @@
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
-import { pairwise } from 'rxjs/operators';
+import { pairwise, take } from 'rxjs/operators';
+import { Subject } from 'rxjs';
+import { expect } from 'chai';
 
 declare function asDiagram(arg: string): Function;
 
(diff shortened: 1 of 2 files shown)


spec/operators/pairwise-spec.ts
```diff
@@ -103,4 +106,28 @@ describe('pairwise operator', () => {
...
+        take(3)
+      )
+      .subscribe(() => {
+        emit('b');
+        spy();
+      });
+
+    emit('b');
+
+    expect(spy).to.have.been.calledOnce;
```
cartant: I agree that there is a bug here, but I'm not sure why the spy is expected to have been called only once. And I'd like to see some expectations regarding the values that are emitted.

Basically, I'd like the test to be easier to understand. It could be made a little simpler using a `Subject`. Here's a snippet that reproduces the bug using a `Subject`:

```ts
import { Subject } from "rxjs";
import { filter, pairwise, take } from "rxjs/operators";

const subject = new Subject<string>();
subject.pipe(
  pairwise(),
  filter(pair => pair[0] !== pair[1]),
  take(3)
)
.subscribe(value => {
  console.log(value);
  subject.next("c");
});
subject.next("a");
subject.next("b");
console.log("done");
```

Which outputs:

```
[ 'a', 'b' ]
[ 'a', 'c' ]
[ 'a', 'c' ]
done
```

thorn0: Please have a look now.

PR #4763, opened 2019-05-09
fix(scan-operator): Improve typings using newer TS features

cah-dunn:
**Description:**
The `scan` operator does not properly infer types when the output Observable type differs from the input Observable type (i.e. `T != R`). This issue was discussed at length a few years back in https://github.com/ReactiveX/rxjs/pull/2897. That PR was ultimately closed because there were unavoidable limitations due to the minimum supported TypeScript at that time. The current minimum TS version should allow these types to be improved.

Using issue https://github.com/ReactiveX/rxjs/issues/3150 as an example:

```
of('0').pipe(
  scan((acc, str) => acc + parseInt(str), 0)
);
```

The current types cause both `acc` and `str` to be inferred as a `string` type, but `acc` should be inferred to be a `number` from the seed value `0`, which is what my change fixes.

https://github.com/ReactiveX/rxjs/issues/3150

### DIFF
--- a/src/internal/operators/scan.ts
+++ b/src/internal/operators/scan.ts
@@ -1,13 +1,7 @@
 import { Operator } from '../Operator';
 import { Observable } from '../Observable';
 import { Subscriber } from '../Subscriber';
-import { OperatorFunction, MonoTypeOperatorFunction } from '../types';
-
-/* tslint:disable:max-line-length */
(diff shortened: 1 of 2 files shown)


cah-dunn:
Closing in favor of https://github.com/ReactiveX/rxjs/pull/4598

PR #4764, opened 2019-05-10
feat(TestScheduler): now supports new expectSubscriptionsTo helper

benlesh:
The old `expectSubscriptions` helper in run mode required users to manually flush the test scheduler in order to run the assertions. This new helper just queues the assertions up to be run, in the same way that `expectObservable` does.

This is a requirement so we can port all of our tests over to use `TestScheduler#run`.

### DIFF
--- a/src/internal/testing/TestScheduler.ts
+++ b/src/internal/testing/TestScheduler.ts
@@ -15,6 +15,7 @@ export interface RunHelpers {
   hot: typeof TestScheduler.prototype.createHotObservable;
   flush: typeof TestScheduler.prototype.flush;
   expectObservable: typeof TestScheduler.prototype.expectObservable;
+  expectSubscriptionsTo: typeof TestScheduler.prototype.expectSubscriptionsTo;
   expectSubscriptions: typeof TestScheduler.prototype.expectSubscriptions;
 }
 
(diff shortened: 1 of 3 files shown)


src/internal/testing/TestScheduler.ts
```diff
@@ -131,6 +132,30 @@ export class TestScheduler extends VirtualTimeScheduler {
...
+    return {
+      toBe: (marbles: string | string[]) => {
+        const marblesArray: string[] = (typeof marbles === 'string') ? [marbles] : marbles;
+        flushTest.ready = true;
+        flushTest.expected = marblesArray.map(marbles =>
+          TestScheduler.parseMarblesAsSubscriptions(marbles, this.runMode)
+        );
+      }
+    };
+  }
```
cartant: I don't see how this performs an expectation. Can a test that deliberately fails - due to an invalid subscriptions expectation - be added to show that this works and isn't just a noop?

benlesh: Added. Good idea.

benlesh:
Ah... You're right @cartant! I think what happened is when I rewrote the TestScheduler for the experimental branch, I ran into this problem, and I guess I figured the same issue existed before. Bad assumption. I should have checked. Fortunately, this wasn't much work.

I'll close this.

PR #4765, opened 2019-05-10
Error rework

benlesh:
The first step in getting rid of custom errors to help reduce bundle sizes. Please [see design doc here](https://hackmd.io/juTxaVkLSiGwkbMK5yZKjA).

### DIFF
--- a/spec/operators/elementAt-spec.ts
+++ b/spec/operators/elementAt-spec.ts
@@ -1,7 +1,8 @@
 import { expect } from 'chai';
 import { elementAt, mergeMap } from 'rxjs/operators';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
-import { ArgumentOutOfRangeError, of, range } from 'rxjs';
+import { ArgumentOutOfRangeError, of, range, isOutOfRangeError } from 'rxjs';
+import { createOutOfRangeError } from 'rxjs/internal/util/ArgumentOutOfRangeError';
 
 declare function asDiagram(arg: string): Function;
 
@@ -12,7 +13,7 @@ describe('elementAt operator', () => {
     const subs =       '^       !      ';
     const expected =   '--------(c|)   ';
 
-    expectObservable((<any>source).pipe(elementAt(2))).toBe(expected);
+    expectObservable(source.pipe(elementAt(2))).toBe(expected);
     expectSubscriptions(source.subscriptions).toBe(subs);
   });
 
@@ -21,7 +22,7 @@ describe('elementAt operator', () => {
     const subs =       '^ !';
     const expected =   '--(a|)';
 
-    expectObservable((<any>source).pipe(elementAt(0))).toBe(expected);
+    expectObservable(source.pipe(elementAt(0))).toBe(expected);
     expectSubscriptions(source.subscriptions).toBe(subs);
   });
 
@@ -30,7 +31,7 @@ describe('elementAt operator', () => {
     const subs =       '^          !';
     const expected =   '-----------(d|)';
 
-    expectObservable((<any>source).pipe(elementAt(3))).toBe(expected);
+    expectObservable(source.pipe(elementAt(3))).toBe(expected);
     expectSubscriptions(source.subscriptions).toBe(subs);
   });
 
@@ -39,7 +40,7 @@ describe('elementAt operator', () => {
     const subs =       '^       !';
     const expected =   '--------(c|)';
 
-    expectObservable((<any>source).pipe(elementAt(2))).toBe(expected);
+    expectObservable(source.pipe(elementAt(2))).toBe(expected);
     expectSubscriptions(source.subscriptions).toBe(subs);
   });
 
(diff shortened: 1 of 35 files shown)


benlesh:
There's something crazy going on with Travis... It's installing node 0.10. Not sure why it started doing that... I hate CI configuration. :roll_eyes:

spec/operators/elementAt-spec.ts
```diff
@@ -1,7 +1,8 @@
 import { expect } from 'chai';
 import { elementAt, mergeMap } from 'rxjs/operators';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
-import { ArgumentOutOfRangeError, of, range } from 'rxjs';
+import { ArgumentOutOfRangeError, of, range, isOutOfRangeError } from 'rxjs';
```
cartant: Don't import `ArgumentOutOfRangeError`?

spec/operators/elementAt-spec.ts
```diff
@@ -48,7 +49,7 @@ describe('elementAt operator', () => {
     const subs =        '(^!)';
     const expected =    '#';
 
-    expectObservable((<any>source).pipe(elementAt(0))).toBe(expected, undefined, new ArgumentOutOfRangeError());
+    expectObservable(source.pipe(elementAt(0))).toBe(expected, undefined, new ArgumentOutOfRangeError());
```
cartant: Use `createOutOfRangeError`?

spec/util/UnsubscriptionError-spec.ts
```diff
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import { UnsubscriptionError, Observable, timer, merge } from 'rxjs';
+import { UnsubscriptionError, Observable, timer, merge, isTeardownError } from 'rxjs';
```
cartant: Don't import `UnsubscriptionError`?

src/internal/Subscription.ts
```diff
@@ -1,8 +1,9 @@
 import { isArray } from './util/isArray';
 import { isObject } from './util/isObject';
 import { isFunction } from './util/isFunction';
-import { UnsubscriptionError } from './util/UnsubscriptionError';
+import { UnsubscriptionError, createTeardownError } from './util/UnsubscriptionError';
```
cartant: Don't import `UnsubscriptionError`?

src/internal/util/UnsubscriptionError.ts
```diff
@@ -21,5 +23,17 @@ UnsubscriptionErrorImpl.prototype = Object.create(Error.prototype);
...
+ *
+ * @deprecated (gone in v8) for `instanceof` checks, instead use {@link isTeardownError}
  */
-export const UnsubscriptionError: UnsubscriptionErrorCtor = UnsubscriptionErrorImpl as any;
\ No newline at end of file
+export const UnsubscriptionError: UnsubscriptionErrorCtor = UnsubscriptionErrorImpl as any;
+
+export function createTeardownError(errors: any[]) {
+  const error: any = new UnsubscriptionError(errors);
+  error.__rxjsErrorCode = RxErrorCode.Teardown;
```
cartant: Is there a reason for not using `createRxError` here?

src/internal/util/errors.ts
```diff
@@ -0,0 +1,68 @@
...
+  Timeout = 3,
+  Teardown = 4,
+}
+
+/**
+ * Creates an error and decorates it with the appropriate error code for identification later.
+ */
+export function createRxError(message: string, code: RxErrorCode, ErrorType: any = Error) {
+  const result = new ErrorType('RxJS: ' + message);
+  (result as any).__rxjsErrorCode = code;
```
cartant: Would there be any advantage in using a symbol instead of a string key for the `__rxjsErrorCode` property?

src/internal/util/UnsubscriptionError.ts
benlesh: Yes. Temporarily, I needed to leverage the existing one, because of the behavior it has in the constructor. Adding a TODO.

src/internal/util/errors.ts
benlesh: There would be one disadvantage, which is it wouldn't work in IE. that's really it.

src/internal/util/errors.ts
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Used to identify different types of RxJS errors.
+ */
+export const enum RxErrorCode {
+  Empty = 0,
+  OutOfRange = 1,
+  ObjectUnsubscribed = 2,
+  Timeout = 3,
+  Teardown = 4,
+}
```
cartant: We cannot use const enums. It will break with isolated modules. They cannot even be exported within internal modules, IIRC.

cartant: IMO, we should get this - or something very much like it - merged: https://github.com/ReactiveX/rxjs/pull/4677

benlesh:
I'm going to close this for now. We still want to do something like this, but I'm not sure 7.0 is the time to do it

Issue #4656, opened 2019-03-19
Creation function `race` is missing description


BioPhoton:
### Please check those that apply

- [ ] typo
- [x] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

### Description Of The Issue

Creation function `race` is missing the description. We might want to implement it.

mkazlauskas:
@waseemahmad31 

There is a description [here](https://rxjs.dev/api/index/function/race), but it's inaccurate:

```Returns an Observable that mirrors the first source Observable to emit an item.```

Instead it appears to be doing what the deprecated operator function description says:

```Returns an Observable that mirrors the first source Observable to emit a next, error or complete notification from the combination of this Observable and supplied Observables.```

OliverJAsh:
Fixed by https://github.com/ReactiveX/rxjs/pull/4773?

benlesh:
Hi @BioPhoton — thank you for taking the time to open this issue and for sharing your thoughtful feedback with the RxJS project. I’m sorry it took us so long to respond. Given how much time has passed and the lack of recent activity, I’m going to close this for now so the tracker reflects the work we’re likely to act on. If this is still relevant on a current RxJS release, please feel welcome to open a new issue with updated context and a reproduction. Thank you again for your contribution and patience.

— Ben

PR #4773, opened 2019-05-11
docs(race):  added description for race

waseemahmad31:
**Description:** added description for race function

#4656


src/internal/observable/race.ts
```diff
@@ -31,7 +31,18 @@ export function race<T>(...observables: ObservableInput<T>[]): Observable<T>;
...
 
 /**
- * Returns an Observable that mirrors the first source Observable to emit an item.
+ * Returns an observable that mirrors the first source observable to emit an item.
+ *
+ * ![](race.png)
+ *
+ * `race` returns an observable, that when subscribed to, subscribes to all source observables immediately.
+ * As soon as one of the source observables emits a value, the result unsubscribes from the other sources.
+ * The resulting observable will forward all emissions, including error and completion, from the "winning"
```
BioPhoton: I would suggest using the wording notification instead of emission, which is more close to the internal naming conventions.

src/internal/observable/race.ts
```diff
@@ -31,7 +31,18 @@ export function race<T>(...observables: ObservableInput<T>[]): Observable<T>;
...
+ * source observable.
+ *
```
BioPhoton: One thing that I would add it the situation of errors of source observables. 
If one of the used sources errors before a first notification the race operator also errors no matter if another source could potentially win the race.

src/internal/observable/race.ts
waseemahmad31: Thanks for suggession. I have made the changes as suggested.

src/internal/observable/race.ts
waseemahmad31: Thanks for suggession. I have made the changes as suggested.

Issue #4613, opened 2019-03-03
[docs]: Operator `toArray` is missing documentation


BioPhoton:
- [ ] typo
- [x] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [x] needs example

### Description Of The Issue

There is no documentation for toArray. This is a duplicate from #3111 #3113. As both of them are closed without any changes to the docs I resubmitt this issue

niklas-wortmann:
closed by #4775

PR #4775, opened 2019-05-12
docs(toArray): added description and example for toArray

waseemahmad31:
**Description:** Added description and example for `toArray`

**Related issue (if exists):** #4613


benlesh:
Thank you @waseemahmad31!

Issue #4614, opened 2019-03-03
[docs]: Operator `every` is missing the description


BioPhoton:
### Please check those that apply

- [ ] typo
- [x] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

### Description Of The Issue

The description block of the `every` operator is missing. We might provide some description to fix this.

hardikpthv:
I belive, this can be closed.

jakovljevic-mladen:
I believe #4777 fixed the problem described in the issue. Closing for now.

PR #4777, opened 2019-05-12
docs(every): Added description for `every`

waseemahmad31:
**Description:** Added description for `every`

**Related issue (if exists):** #4614

### DIFF
--- a/src/internal/operators/every.ts
+++ b/src/internal/operators/every.ts
@@ -6,6 +6,11 @@ import { Observer, OperatorFunction } from '../types';
 /**
  * Returns an Observable that emits whether or not every item of the source satisfies the condition specified.
  *
+ * <span class="informal">If all values pass predicate before completion emit true,
+ * else false.</span>
+ *
+ * ![](every.png)
+ *
  * ## Example
  * A simple example emitting true if all elements are less than 5, false otherwise
  * ```ts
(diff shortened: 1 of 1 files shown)


src/internal/operators/every.ts
```diff
@@ -6,6 +6,11 @@ import { Observer, OperatorFunction } from '../types';
 /**
  * Returns an Observable that emits whether or not every item of the source satisfies the condition specified.
  *
+ * <span class="informal">If all values pass predicate before completion emit true,
```
benlesh:
```suggestion
 * <span class="informal">If all values pass predicate before the source completes, emits true before completion,
```

src/internal/operators/every.ts
```diff
@@ -6,6 +6,11 @@ import { Observer, OperatorFunction } from '../types';
...
+ * else false.</span>
```
benlesh:
```suggestion
 * otherwise emit false, then complete.</span>
```

src/internal/operators/every.ts
waseemahmad31: Thanks @benlesh for the suggesstion.

src/internal/operators/every.ts
waseemahmad31: Thanks @benlesh for the suggesstion.

PR #4780, opened 2019-05-14
feat(filterByLatestFrom): add filterByLatestFrom operator

OlaviSau:
This operator allows to use other observables as filters.

This is doable with other operators, but is prone to error and complicated to write.

## Feature Request

This should either be a core operator or addition of good documentation about the implementation and pitfalls.

This can be implemented as a user land operator.

**(If this is new operator request) describe reason it should be core operator**
There should be a clear answer to filtering using input from an additional observable.

**Additional context**
The most common issue is that it is very simple to think that moving the filter into the inner observable is safe. Example in a comment. There is an issue that RxJS does not have a canonical extensions library like lodash. Many operators in RxJS do not need to be in the core, but they still are. I would gladly add this to the canonical extensions library, if it were available.

### DIFF
--- a/doc/decision-tree-widget/tree.yml
+++ b/doc/decision-tree-widget/tree.yml
@@ -21,6 +21,9 @@ children:
       - label: based on custom logic
         children:
         - label: filter
+      - label: based on another obvervable
+        children:
+        - label: filterByLatestFrom
       - label: if they are at the start of the Observable
         children:
         - label: and only the first value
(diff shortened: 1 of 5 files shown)


OlaviSau:
This can be implemented via other operators -
```
withLatestFrom(...),
filter(([, ...)] => ...),
map(([...]) => ...)
```
The idiomatic way skips the map. The issue with this is that it is fairly easy for the developer to think that they can do - 
```
withLatestFrom(observable.pipe(filter(...))),
map(([...]) => ...)
```
The latest value will never be updated with a value that does not pass the filter, thus creating a hard to spot issue where a `filter` becomes a `skipUntil` instead.

cartant:
I'm inclined to think that this could be implemented using `windowToggle` pretty easily.

OlaviSau:
Something like this?
```
windowToggle(observableFilter, boolean ? timer(0) : EMPTY)
```

OlaviSau:
@kwonoj Added the issue template.

kwonoj:
Thanks, but following comments here I am bit hard to see why this should be in core. We do strongly recommend to create operator in user level module, and think this one is same case too.

OlaviSau:
@kwonoj Thank you for the response :)

PR #4782, opened 2019-05-14
fix(migrations): use latest version of package when adding compat

clydin:
Update and cleanup the RxJS migration schematic.  

//cc @benlesh


benlesh:
@clydin if you rebase, CI should work properly now. Sorry about that.

PR #4785, opened 2019-05-15
Misc fixes

manuelaljibesrosas:
**Description:**

Just a couple of minor updates to improve readability:

* Remove unused Teardown import.

* Merge `__tryOrUnSub` and `__tryOrSetError` in a single method called `_try` that's easier to understand
and eliminates repetition.

* Remove the redundant `!this.isStopped` checks inside the SafeSubscriber's
handlers, the parent Subscriber is already doing that for us. 

* Remove the `observerOrNext !== emptyObserver` check inside the SafeSubscriber constructor, the marginal (if any) performance gain doesn't justify the cognitive overhead of a three-layered conditional.

* Change the format of arguments to align in the same indentation as the methods bodies, this way the eye doesn't have to jump around diagonally between arguments and the body and has the collateral benefit or reducing line length.


cartant:
Thanks for the effort you've put into this. However, IMO, these changes cannot be merged. Readability is not the most important metric for the code in this particular file. The code in that file was recently reviewed and changed by [Benedikt Meurer](https://github.com/ReactiveX/rxjs/commits?author=bmeurer) with a view to improving performance.

And, on a general note, if some portion of the PR addresses only reformatting, those changes should be separated into individual commits - to make reviewing easier.

ATM, I am reluctant to review this. In future, you might want to open an issue check whether or not the changes you are proposing are the sorts of changes that are likely to be merged.

Again, thanks for the effort you've put into this. I hope this won't discourage you from contributing.

manuelaljibesrosas:
Understandable, although these changes are related to form only, the code is functionally the same except for the case of `emptyObserver` in the `SafeSubscriber` constructor, which in itself shouldn't even be there since `SafeSubscriber` never receives an `emptyObserver` in the first place, and even if a user went through the hassle of importing it from the Subscriber file, it wouldn't make sense to do so since observers are automatically converted to instances of `SafeSubscriber` internally.

This file in particular is filled with defensive checks like this one that made it really hard for me as an outsider to even understand what exactly are the cases it covers, even though the underlying logic is really simple; so, would you reconsider reviewing the proposed `_try` method if I break up the reasoning behind it like I just explained the reasons behind the removal of the `emptyObserver` case?

cartant:
Regarding the use of `arguments.length`, I would be reluctant of accept a change of that nature, as the effected behaviour will be essentially unchanged. And the performance impact of using `arguments.length` is something I would need to research. Although, I suspect others would already know the answer to this.

I don't want to discourage you from contributing; I want the opposite, really. And I'm pleased to see that you have taken such an interested in the inner workings of RxJS.

I do have some suggestions:

* Please separate any changes into individual PRs - unless the changes are directly related.

    Doing so makes reviewing and merging much simpler. I would rather see half a dozen small PRs than one large one. It's entirely possible that some changes might be welcomed and others not.

* Ideally, PRs should be accompanied by issues. Opening an issue first provides the opportunity to discuss the bug that has been found or the changes that are being proposed. And talking about it first is generally preferable. It can avoid work being done on a change that is unlikely to be merged and it avoids the situation in which a maintainer has to tell a contributor that the change they have proposed isn't wanted. No one likes doing that.

The other thing that I am considering - that you might or might not be aware of - is that the `Subscriber` implementation is going to be changing. So the code for which you are proposing changes is not going to be around in the long term.

manuelaljibesrosas:
I see, thanks for the feedback I appreciate it, and yes, I read about `MutabeSubscriber` in your Rxjs 8 design document but still didn't want to miss the chance of discussing the current implementation with the maintainers, anyway, I'll open up an issue to discuss other aspects about the current Subscriber that are related to behavior.

PR #4787, opened 2019-05-16
chore: remove useless import

cartant:
**Description:**

This PR removes a useless import - that appears to have been added by VS Code's autocomplete/autoimport feature.


kwonoj:
Thought we had tslint or compiler for unused import.

cartant:
@kwonoj The in-built TSLint rule for unused imports is hugely problematic. There is a rule in `tslint-etc` that does a better job. If you want that enabled, I can submit a PR.

kwonoj:
This kind of things can benefit from tools obviously, so I'm in for it if you're ok.

PR #4789, opened 2019-05-16
docs(operators): fix marble-diagram image src

jmonterroso:
Updating to proper marble diagram image src property

**Description:**
Fixing src property of marble diagram, in RxJS buffer operator documentation. 

N/A

### DIFF
--- a/src/internal/operators/buffer.ts
+++ b/src/internal/operators/buffer.ts
@@ -12,7 +12,7 @@ import { OperatorFunction } from '../types';
  * <span class="informal">Collects values from the past as an array, and emits
  * that array only when another Observable emits.</span>
  *
- * ![](content/img/buffer.png)
+ * ![](/assets/images/marble-diagrams/buffer.png)
  *
  * Buffers the incoming Observable values until the given `closingNotifier`
  * Observable emits a value, at which point it emits the buffer on the output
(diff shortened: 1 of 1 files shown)


src/internal/operators/buffer.ts
```diff
@@ -12,7 +12,7 @@ import { OperatorFunction } from '../types';
  * <span class="informal">Collects values from the past as an array, and emits
  * that array only when another Observable emits.</span>
  *
- * ![](content/img/buffer.png)
+ * ![](/assets/images/marble-diagrams/buffer.png)
```
JWO719: This will propably work, but we append the relative path anyway, for reference see the count operator. It would be more consistent if we just use `buffer.png`

jmonterroso: Thanks for your feedback @JWO719 I fixed the path based on your comment.

pertrai1:
@jmonterroso You will probably also want to look and see how the diagrams are being created and where they are being stored:

https://github.com/ReactiveX/rxjs/blob/master/spec/helpers/tests2png/diagram-test-runner.ts#L99
https://github.com/pertrai1/rxjs/blob/70c33f968ac2f82dd02f56f0f7d9a72d31f0d855/package.json#L99

cc: @JWO719

PR #4790, opened 2019-05-17
chore(lint): add no-unused-declaration rule

cartant:
**Description:**

This PR bumps the version of `tslint-etc` and enables the `no-unused-declaration` rule to effect failures for unused imports - as discussed in [this PR](https://github.com/ReactiveX/rxjs/pull/4787#issuecomment-492899988). There is a built-in TSLint rule named `no-unused-variable`, but it is buggy and has caused [problems for me in the past](https://github.com/cartant/tslint-etc#rules).

Note that the rule requires type information, so the `tsconfig.json` is included in the lint commands.

**Related issue (if exists):** None

PR #4791, opened 2019-05-17
chore: simplify dtslint tooling

cartant:
**Description:**

[The PR that I opened](https://github.com/microsoft/dtslint/pull/229) to have the `localTs` path resolved has been merged and a new version of dtslint has been published, so there is no longer a need to pass an absolute path via the command line.

This PR removes that script that was used to pass an absolute path to the local TypeScript install and passes a relative path, instead.


benlesh:
@cartant Looks like we just need the conflict resolved and this is good to go.

cartant:
@benlesh Conflicts in this should be resolved. I've not resolved those in #4790 because - AFAICT - it will conflict with the changes in this PR. I'll resolve #4790 after this PR is merged.

Issue #4808, opened 2019-05-25
race continues to subscribe when won via complete or error


cartant:
## Bug Report

This is a re-write of the bug found in #4806.

**Current Behavior**

When a `race` is won by a source observable that completes or errors synchronously, the implementation continues to subscribe to subsequent source observables.

This does not happen if the source observable that wins the race does so by emitting a value notification.

**Reproduction**

This code:

```ts
import { defer, EMPTY, of, race, throwError } from "rxjs";
import { delay } from "rxjs/operators";

race(
  defer(() => {
    console.log("subscribed to 1");
    return of(1).pipe(delay(100));
  }),
  defer(() => {
    console.log("subscribed to EMPTY");
    return EMPTY;
  }),
  defer(() => {
    console.log("subscribed to 3");
    return of(3).pipe(delay(300));
  })
).subscribe({
  complete: () => console.log("won"),
  error: () => console.log("won with error")
});
```

will output:

```
subscribed to 1
subscribed to EMPTY
won
subscribed to 3
```

and this code:

```ts
race(
  defer(() => {
    console.log("subscribed to 1");
    return of(1).pipe(delay(100));
  }),
  defer(() => {
    console.log("subscribed to error");
    return throwError(new Error("kaboom"));
  }),
  defer(() => {
    console.log("subscribed to 3");
    return of(3).pipe(delay(300));
  })
).subscribe({
  complete: () => console.log("won"),
  error: () => console.log("won with error")
});
```

will output:

```
subscribed to 1
subscribed to error
won with error
subscribed to 3
```

**Expected behavior**

For races won via completion or error, the behaviour should be like this code and subscriptions to subsequent source observables should not be made:

```ts
race(
  defer(() => {
    console.log("subscribed to 1");
    return of(1).pipe(delay(100));
  }),
  defer(() => {
    console.log("subscribed to 2");
    return of(2);
  }),
  defer(() => {
    console.log("subscribed to 3");
    return of(3).pipe(delay(300));
  })
).subscribe({
  complete: () => console.log("won"),
  error: () => console.log("won with error")
});
```

will output:

```
subscribed to 1
subscribed to 2
won
```

**Environment**
- Runtime: Node
- RxJS version: 6.5.2

PR #4809, opened 2019-05-25
fix(race): ignore latter sources after first complete or error

cartant:
**Description:**

The implementation of `race` ignores latter sources if an earlier source emits a `next` notification. This PR fixes the implementation so that latter sources are also ignored if an earlier source wins the race by emitting either a `complete` or an `error` notification.

This PR adds failing tests that are fixed by the change.

For more information, see the related issue - it contains code snippets that highlight the inconsistent behaviour.

**Related issue (if exists):** #4808


benlesh:
Because subscribing to Observables can have side effects, this will be a breaking change, and we can't land it in 6.x. However we could land it in master.

PR #4815, opened 2019-05-28
docs(operators): skipWhile() docs to improve

TinTin83:
I took a look at the marble diagram, usually they guide me well, but this one is a bit tricky. Because I misunderstood with filter(), thinking that the condition will be always checked for skipWhile(), which is true, UNTIL we reach one time the condition equals false.
 I didn't see that UNTIL :/

Is it possible in the marble to add these values ==> 2 / 3 / 4 (condition becomes false) / 5 / 6 / 1 / 2 / 3 (so adding  the last 1 / 2 / 3) to show people that the value will be emitted.

Thanks,
Martin

**Description:**


benlesh:
Should probably file an issue instead of a PR.

PR #4817, opened 2019-05-29
Marble testing docs update

tomwhite007:
**Description:**

When I first read this document original, I had difficulty understanding the following:

1. How to create a `testScheduler.run()` clause because the example had a different variable name - which sounds simple, but when you're scanning help pages, requires you to stop and think.
2. How to 'wait' in virtual time for the hot or cold Observables to complete before making an assertion.
3. Why `flush()` is described as 'Not often used', when my Angular Observable subscription code calls actions or updates to variables _almost_ as often as they just `map` to another Observable.

Therefore, my main reason for submitting this update is to make it easier for newbies to Marble Testing  like myself can:

a) Googling on 'marbles testing wait', 'marbles testing pause', 'marbles testing delay' 'marbles testing timer', 'marbles testing observable complete', and 'synchronous assertion' will take you to this document - which I would have benefited from.
b) A `flush()` example for synchronous assertion exists within the document, which would have previously saved me an hour or two.

### DIFF
--- a/docs_app/content/guide/testing/marble-testing.md
+++ b/docs_app/content/guide/testing/marble-testing.md
@@ -205,47 +205,27 @@ it('should repeat forever', () => {
 
 ## Synchronous Assertion
 
-Sometimes, we need to assert changes in state after an observable stream has completed - such as when our subscription doesn’t map to another observable but instead just updates a basic variable. Outside of Marbles testing with TestScheduler, we might think of this as creating a delay or setting a timer before making our assertion.
+Sometimes, we need to assert changes in state _after_ an observable stream has completed - such as when a side effect like `tap` updates a variable. Outside of Marbles testing with TestScheduler, we might think of this as creating a delay or waiting before making our assertion.
 
 For example:
 
 ```ts
-  let myVar = 0;
-  let myStream$: Observable<string>;
-  const myFunctionUnderTest = () => {
-    myStream$
-      .pipe(
-        tap(() => {
-          myVar++;
-        }),
-      )
-      .subscribe();
-  };
+  let eventCount = 0;
 
-  it('should set a local variable', () => {
-    const testScheduler = new TestScheduler(() => {});
+  const s1 = cold('--a--b|', { a: 'x', b: 'y' });
 
-    testScheduler.run(helpers => {
-      const { cold, flush } = helpers;
+  // side effect using 'tap' updates a variable
+  const result = s1.pipe(tap(() => eventCount++));
+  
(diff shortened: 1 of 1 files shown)


docs_app/content/guide/testing/marble-testing.md
```diff
@@ -202,7 +203,46 @@ it('should repeat forever', () => {
...
+      .subscribe();
+
+    // trigger testScheduler 'virtual time' to complete all outstanding hot or cold observables
+    // (or, 'Pause' in virtual time until the observable completes)
+    flush();
+
+    expect(myVar).toBe(2);
+  });
+});
+```
```
cartant: If an example of a test that explicitly calls `flush` is to be included, it will have to, IMO, be a better test than this. This test doesn't test anything that could not be better tested with `expectObservable`.

tomwhite007: Hi @cartant 
Thanks for reviewing my PR so quickly.

1. I had previously tried to keep my example test as short/simple as possible, but now agree with you that it needs to be clearer. I've submitted an update for your review.
2. Can I take it that you're ok with my suggested text changes, and all I need to do is iterate on an example test to get your approval?

Thanks
Tom

cartant: I've only skimmed the text changes. For me, the example test is more important, as it's, IMO, important that it's idiomatic and encourages good testing practice.

I'm still not convinced that the example is realistic or that it represents testing best practice. I would not want to encourage people to test using state outside of the `it` call.

I think a better example would be something like what's done in the library's tests. Something like this:

https://github.com/ReactiveX/rxjs/blob/01a09789a0a9484c368b7bd6ed37f94d25490a00/spec/operators/finalize-spec.ts#L79-L88

tomwhite007:
Hello @cartant 
I've rehashed the example test to follow the example patterns you sent me, and simplified the text description. Hope this is what you want?

docs_app/content/guide/testing/marble-testing.md
benlesh: @cartant ALL of our testing documentation should really be guiding people to use "run mode" tests, what we're doing in the library here really needs to be refactored to do the same as soon as possible.

So it should look like:

```ts
const rxTest = new TestScheduler(assertDeepEquals);
rxTest.run(({ hot, cold, expectObservable }) => {
  /// tests here.
});
```

benlesh: Ah, nevermind, I see what you were saying. You weren't suggesting globals.

benlesh:
Thanks @tomwhite007!

Issue #4804, opened 2019-05-22
defer type doesn't handle void correctly


mhverbakel:
## Bug Report

**Current Behavior**
When using `defer` with a `() => void` factory, the return type will be `Observable<any>`.

**Reproduction**
https://stackblitz.com/edit/rxjs-gm1wy5

```ts
import { defer } from 'rxjs';

const example = defer(() => {});

// Type of example is Observable<any> but should be Observable<never>.
```

**Expected behavior**
When using `defer` with a `() => void` factory, the return type will be `Observable<never>`.

**Possible Solution**
Adding the following to the declaration:
```ts
function defer(observableFactory: () => void): Observable<never>;
```

cartant:
I don't agree with the expected behaviour. `defer` should return an `ObservableInput` and `void` is not an `ObservableInput`.

mhverbakel:
Fine with me, but then `| void` should be removed from https://github.com/ReactiveX/rxjs/blob/6.5.2/src/internal/observable/defer.ts#L55.

cartant:
Yikes. I didn't even realise that was there. This should definitely have two signatures.

PR #4835, opened 2019-06-05
 fix(defer): restrict allowed factory types

cartant:
**Description:**

This PR fixes a problem with the `defer` type signatures that was introduced in #4810 - the PR that ensured that `Observable<never>` was inferred for factory functions that return `void`.

The problem was that `() => void` will match functions returning any value - as the return value is ignored - which meant that factories could return anything - e.g. a `number`.

See the discussion on [this commit](https://github.com/ReactiveX/rxjs/commit/362d1d4fc1d7f15e7168453ea640079d0ca19dba#commitcomment-33814932).

The PR includes a failing dtslint test that is fixed by the change.

**Related issue (if exists):** #4804 #4810

### DIFF
--- a/src/internal/observable/defer.ts
+++ b/src/internal/observable/defer.ts
@@ -52,18 +52,16 @@ import { empty } from './empty';
  * @name defer
  * @owner Observable
  */
-export function defer<O extends ObservableInput<any>>(observableFactory: () => O): Observable<ObservedValueOf<O>>;
-export function defer(observableFactory: () => void): Observable<never>;
-export function defer<O extends ObservableInput<any>>(observableFactory: () => O | void): Observable<ObservedValueOf<O>> {
-  return new Observable<ObservedValueOf<O>>(subscriber => {
-    let input: O | void;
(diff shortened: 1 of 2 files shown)


src/internal/observable/defer.ts
```diff
@@ -52,18 +52,16 @@ import { empty } from './empty';
  * @name defer
  * @owner Observable
  */
-export function defer<O extends ObservableInput<any>>(observableFactory: () => O): Observable<ObservedValueOf<O>>;
-export function defer(observableFactory: () => void): Observable<never>;
-export function defer<O extends ObservableInput<any>>(observableFactory: () => O | void): Observable<ObservedValueOf<O>> {
-  return new Observable<ObservedValueOf<O>>(subscriber => {
-    let input: O | void;
+export function defer<R extends ObservableInput<any> | void>(observableFactory: () => R): Observable<ObservedValueOf<R>> {
```
benlesh: `R extends void`

:fried_egg: <-- your brain on TypeScript

cartant:
Added the missing test.

Issue #3798, opened 2018-06-05
Undeprecate `concat` operator as `concatWith`


benlesh:
While the use cases for the `merge`, `zip`, and `combineLatest` operators, et al, remain dubious, I've several times seen cases where the `concat` _operator_ might be preferred to the concat static method.

Most of these use cases are around animations or action dispatching with `redux-observable` or `ngrx`.

It could be used to compose `endWith`.

The name `concatWith` has some precedent. I believe it's used in RxJava.

benlesh:
This is done #4988

PR #4836, opened 2019-06-06
Adding static publish/multicast variants, and deprecating operators that return ConnectableObservable

benlesh:
Work in progress....

### DIFF
--- a/src/internal/observable/publishReplay.ts
+++ b/src/internal/observable/publishReplay.ts
@@ -1,0 +1,12 @@
+import { Observable } from '../Observable';
+import { ReplaySubject } from '../ReplaySubject';
+import { multicast } from './multicast';
+import { ConnectableObservable } from '../observable/ConnectableObservable';
+
+export function publishReplay<T>(
+  source: Observable<T>,
+  bufferSize = Number.POSITIVE_INFINITY,
+  windowTime = Number.POSITIVE_INFINITY
+): ConnectableObservable<T> {
+  return multicast(source, new ReplaySubject<T>(bufferSize, windowTime));
+}
--- a/src/internal/observable/multicast.ts
+++ b/src/internal/observable/multicast.ts
@@ -1,0 +1,8 @@
+import { Observable } from '../Observable';
+import { Subject } from '../Subject';
+import { ConnectableObservable } from '../observable/ConnectableObservable';
+
+export function multicast<T>(source: Observable<T>, subjectOrSubjectFactory?: (() => Subject<T>) | Subject<T>) {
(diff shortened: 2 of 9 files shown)


JWO719:
In general I like the idea of splitting that up due to it's differnet behavior. But one think that came to my mind, isn't it confusing for api users to have a static operator and a pipeable operator with the same name but a different behavior? I don't know if it would be a good idea to give one of those two a different name in that case.

cartant:
In #3833, the operator names had a `With` suffix - e.g. `publishWith` - to distinguish them from the static creators. I quite liked that idea. Is there a reason for not going in that direction? Perhaps because the usage is a little different to the `concatWith` in #3798? Is that issue still a thing?

benlesh:
>  Is there a reason for not going in that direction?

This is still a WIP, :), hence the label. I just didn't get to the aliasing/deprecations yet.

>  But one think that came to my mind, isn't it confusing for api users to have a static operator and a pipeable operator with the same name but a different behavior?

Perhaps... although right now we have one operator that behaves two different ways depending on how you call it.  I think once we have `publishWith` or `publishAs` or whatever we want to call it, it will make more sense.

BioPhoton:
> we have one operator that behaves two different ways depending on how you call it

This is also the case with a selector parameter in `multicast` and `publish`. Here the output is completely different.
`multicast` itself should be at least 2 different operator as well as static methods I believe.

Also, shareReplay has completely different behavior if refCount true is used.

In my opinion, the whole multicasting section should be checked. There are a lot of historical edge cases etc included which could get cleaned up or deprecated.

src/internal/observable/publishReplay.ts
```diff
@@ -0,0 +1,12 @@
+import { Observable } from '../Observable';
+import { ReplaySubject } from '../ReplaySubject';
+import { multicast } from './multicast';
+import { ConnectableObservable } from '../observable/ConnectableObservable';
+
+export function publishReplay<T>(
+  source: Observable<T>,
+  bufferSize = Number.POSITIVE_INFINITY,
```
LayZeeDK: `Number.MAX_SAFE_INTEGER`?

src/internal/observable/publishReplay.ts
```diff
@@ -0,0 +1,12 @@
...
+  windowTime = Number.POSITIVE_INFINITY
```
LayZeeDK: `Number.MAX_SAFE_INTEGER`?

benlesh:
Closing this older PR in favor of work (and renewed discussion) here: #5432

PR #4839, opened 2019-06-06
feat(rxjs-compat): removed for v7

benlesh:
Removes all files that were solely related to v6 backward compatibility with v5, and/or related to schematic migration from v5 to v6.

Note that there were some changes required to tsconfig files based off of changes to the build process (in order to stop building compatibility related files)

BREAKING CHANGE: `rxjs/Rx` is no longer a valid import site.
BREAKING CHANGE: `rxjs-compat` is not published for v7 (yet)


package.json
```diff
@@ -222,8 +198,5 @@
   "engines": {
     "npm": ">=2.0.0"
   },
-  "typings": "./dist/package/Rx.d.ts",
```
cartant: If `rxjs/Rx` is no longer a valid import, is `./dist/package/Rx.d.ts` going to be were the types are?

benlesh: Hmmm... that's a good question. I think we're going to need to take another pass through the module settings in package.json to ensure we're going about this right. We have have a little time, we just need to sort it out before the first alpha release.

PR #4841, opened 2019-06-07
docs(page): add breaking changes page

### LINKED ISSUE #4512: Issues with RxJS 6.4.0 and Angular CLI 6.1.x
## Issue
TL;DR: Angular CLI used TS 2.7 and pulled in RxJS ^6.0.0, RxJS 6 has always used TS 2.8, RxJS 6.4 introduced the first feature that required TS 2.8. ^6.0.0 will update to 6.4, and break you if you're still using TS 2.7.

### Summary 
As summarized by @cartant and @kwonoj :

- RxJS 6.0.0 was released on 25 Apr 2018
- TypeScript 2.8 was released on 28 Mar 2018
- Angular CLI 6.1.0 was released on 26 Jul 2018, depending upon RxJS ^6.0.0 and TypeScript ~2.7.2
- It seems that Angular CLI chose RxJS v6 and an at-the-time older version of TS
- Since the RxJS v6 release, my understanding has been that we were targeting the TS 2.8. That’s what’s used in Travis, etc.
- The PR that introduced `ObservedValueOf` introduced a conditional type
- Conditional types are not supported in TS 2.7 - they were introduced in TS 2.8
- Until conditional types were introduced, it seems that - by chance - we’d not used any TS features that were in 2.8, but not in 2.7
- Angular CLI 6.2.0 was released on 7 Sep 2018, depending upon RxJS ~6.2.0 and TypeScript ~2.9.2, so this is an Angular CLI 6.1.x problem only

### Fix (TBD)

~~The current plan to fix this is actually to update RxJS to TS 3.2, as TS 3.1 supports building .d.ts files that target multiple versions of TS. So we should be able to fix this in a patch release, and, even better, put RxJS in a better place to use the newest TS features that could really help us without us needing to do major release.~~

**(EDIT: TS 3.1 does NOT support "building" .d.ts files for multiple versions of TS, we'd need to maintain those by hand, given the limited development bandwidth we have, I'm not sure what we can do here)**

### Workarounds

- You could update your project to TS 2.8
- You can pin your project to rxjs@6.3.3

### EDIT: The core team is currently investigating the impact of this issue and options to fix it

niklas-wortmann:
As discussed in #4512 . This pr adds a static page describing the actual bug occured with version 6.4 and how to solve it. It can also be used for any further breaking changes to add some dedicated documentation for it.

closes #4512 

The Page actually looks like that:
image


docs_app/content/navigation.json
```diff
@@ -90,6 +90,10 @@
 					"url": "api/deprecations",
 					"title": "Deprecations",
 					"tooltip": "List of Deprecations"
+        },
+        {
+					"url": "guide/v6/breaking-changes",
+					"title": "Breaking Changes"
```
benlesh: Does this need a tooltip for a11y reasons?

benlesh:
The more I think about this, the more I'm not sure we should do this. The `CHANGELOG` file should be the source of truth for breaking changes. They've all been documented there. Perhaps we need to be better about going back and updating it when we find something breaking after a release that we can't unbreak, for fear that would break even more people.

JWO719:
To sum up yesterday's discussion: I will add the Changelog to the docs as well and we will keep this breaking changes pages as additional information besides the changelog

benlesh:
@niklas-wortmann is this still relevant? Or can it be closed?

niklas-wortmann:
good call. close in favor of #5426

PR #4852, opened 2019-06-11
Fix pluck typings

benlesh:
We have an issue in Google where `pluck('foo', ...otherArgs)` is getting compilation errors, this should correct that. Also starts moving to using `unknown` instead of `{}` when possible.

### DIFF
--- a/spec-dtslint/operators/pluck-spec.ts
+++ b/spec-dtslint/operators/pluck-spec.ts
@@ -26,11 +26,11 @@ it('should support nested object of 6 layer depth', () => {
 });
 
 it('should support nested object of more than 6 layer depth', () => {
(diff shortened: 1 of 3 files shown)


spec-dtslint/operators/pluck-spec.ts
```diff
@@ -30,13 +30,29 @@ it('should support nested object of more than 6 layer depth', () => {
 });
 
 it('should accept existing keys only', () => {
-  const a = of({ name: 'abc' }).pipe(pluck('xyz')); // $ExpectError
+  const a = of({ name: 'abc' }).pipe(pluck('xyz')); // $ExpectType Observable<unknown>
 });
 
 it('should not accept empty parameter', () => {
-  const a = of({ name: 'abc' }).pipe(pluck()); // $ExpectError
+  const a = of({ name: 'abc' }).pipe(pluck()); // $ExpectType<unknown>
```
cartant: This doesn't look like a valid dtslint type expectation. Should it be `$ExpectType Observable<unknown>`?

Issue #4851, opened 2019-06-11
takeLast emits undefined it applied with no params


kievsash:
## Bug Report

**Current Behavior**
 takeLast() operator when applied without any param  returns undefined.

**Reproduction**
https://codepen.io/kievsash/pen/dEgyGX?editors=0010

```js
let Rx = window.Rx = window['rxjs'];
let {forkJoin, Subject, merge, of} = Rx;
let {ajax} = Rx.ajax;
let {map, finalize, tap, takeLast, scan} = Rx.operators;
console.clear();


const m$ = ajax('https://jsonplaceholder.typicode.com/comments/1').pipe(takeLast()); // incorrect

//const m$ = ajax('https://jsonplaceholder.typicode.com/comments/1').pipe(takeLast(1)); // correct

m$.subscribe(console.warn)
```

**Expected behavior**
WEll, should be discussed I guess. I propose that takeLast() to work like takeLast(1)  -

**Environment**
- Runtime: Chrome Version 74.0.3729.169 (Official Build) (64-bit)
- RxJS version: https://cdnjs.cloudflare.com/ajax/libs/rxjs/6.5.2/rxjs.umd.min.js

**Possible Solution**
 To set count=1 by default here: https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/takeLast.ts#L50

**Additional context/Screenshots**

kwonoj:
Since our interface strictly requires param, it's more like an edge case. If I were to amend behavior, I'd rather throw argumentRangeError as same as non-acceptable count param instead of implicitly fallback to other behavior.

kievsash:
Ok, If it is final decision I can try to provide pull request with expected behavior.

kievsash:
@cartant  Kindly take a look: https://github.com/ReactiveX/rxjs/pull/4854

kwonoj:
https://github.com/ReactiveX/rxjs/pull/5326

PR #4854, opened 2019-06-12
fix(takeLast): no arguments lead to undefined emission - changed to A…

kievsash:
…rgumentOutOfRangeError #4851

**Description:**
takeLast should throw argumentRangeError if applied with no argument 

https://github.com/ReactiveX/rxjs/issues/4851

### DIFF
--- a/spec/operators/takeLast-spec.ts
+++ b/spec/operators/takeLast-spec.ts
@@ -146,6 +146,11 @@ describe('takeLast operator', () => {
       .to.throw(ArgumentOutOfRangeError);
   });
 
+  it('should throw if total is not provided (takeLast()) or undefined', () => {
+    expect(() => { range(0, 10).pipe(takeLast(undefined)); })
+        .to.throw(ArgumentOutOfRangeError);
+  });
+
   it('should not break unsubscription chain when unsubscribed explicitly', () => {
     const e1 = hot('---^--a--b-----c--d--e--|');
     const unsub =     '         !            ';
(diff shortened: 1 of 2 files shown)


kievsash:
Thank you for review, @cartant!

benlesh:
This seems like the wrong solution. `takeLast()` with no arguments should be the same as `takeLast(1)`.

kievsash:
We discussed it in respective issue. But implementing Ben's solution is even easier - just to provide counter=1 as default value

benlesh:
Well... We'll discuss it at the team meeting I guess

benlesh:
Notes on this:

1. It's weird that `takeLast()` emits `undefined`... 
2. It's even weirder that `takeLast(-1)` doesn't complete.

The best choice may be to add runtime checks and error. Same for `take`.

kievsash:
If we check count param in constructor - we cannot complete source since we don't have it yet.
Possibly to move count logic check to 'call'

class TakeOperator {
  constructor(private total: number) {
   // if (this.total < 0) {
   //   throw new ArgumentOutOfRangeError;
   // }
  }

  call(subscriber, source) {
     if (this.total < 0 or this.total === undefined) {
        source.complete()
        throw new ArgumentOutOfRangeError;
    }
    return source.subscribe(new TakeSubscriber(subscriber, this.total));
  }
}

What do you think?

cartant:
Sinks don't complete sources. Rather, they would unsubscribe and send a complete notification to the destination.

Regardless of that, this issue was discussed in a team meeting and a more comprehensive change was decided upon. See https://github.com/ReactiveX/rxjs/pull/5326

This PR should be closed in favour of the linked PR.

benlesh:
Thank you for this contribution, @kievsash! Sorry we went another direction, but this PR was helpful in getting us to notice and discuss it!

PR #4858, opened 2019-06-12
fix(scan/reduce): Typings correct for mixed seed/value types

benlesh:
- Adds dtslint tests to cover various mixtures of seeds, accumulator results, and value types
- Refactors scan a little bit, as types needed to be updated in the implementation
- Resolves a performance issue where scan was calling next on the destination Subscriber when it had already errored


src/internal/operators/scan.ts
```diff
@@ -45,14 +45,14 @@ export function scan<T, R>(accumulator: (acc: R, value: T, index: number) => R):
  * @see {@link mergeScan}
  * @see {@link reduce}
  *
- * @param {function(acc: R, value: T, index: number): R} accumulator
+ * @param {function(acc: A, value: V, index: number): A} accumulator
  * The accumulator function called on each source value.
- * @param {T|R} [seed] The initial accumulation value.
- * @return {Observable<R>} An observable of the accumulated values.
+ * @param {V|A} [seed] The initial accumulation value.
```
kwonoj: https://github.com/ReactiveX/rxjs/pull/4858/files#diff-bcface33933d00ae116bcf19e16b2b7cR55 specifies seed as `S` but it doesn't extends `V|A` as comment says - maybe this needs to be S?

kwonoj: uh welp maybe above overloading does its job

src/internal/operators/reduce.ts
```diff
@@ -54,28 +54,32 @@ export function reduce<T, R>(accumulator: (acc: R, value: T, index: number) => R
...
- * @param {R} [seed] The initial accumulation value.
- * @return {Observable<R>} An Observable that emits a single value that is the
+ * @param {A} [seed] The initial accumulation value.
+ * @return {Observable<A>} An Observable that emits a single value that is the
  * result of accumulating the values emitted by the source Observable.
  * @method reduce
  * @owner Observable
  */
-export function reduce<T, R>(accumulator: (acc: T | R, value: T, index?: number) => T | R, seed?: T | R): OperatorFunction<T, T | R> {
+export function reduce<V, A>(accumulator: (acc: V | A, value: V, index?: number) => A, seed?: any): OperatorFunction<V, V | A> {
```
cartant: Does `seed` really need to be `any`? It seems to be `S` in the implementation signature for `scan`.

benlesh: Given the hoops you need to jump through in the implementation if it's `S`, I just made it `any`. This is the implementation, not the type signature.

src/internal/operators/scan.ts
benlesh: Yeah. These docs need an overhaul, probably.

PR #4861, opened 2019-06-15
build: refactor build script

filipesilva:
Following a conversation with @benlesh, I'm taking a look at the tooling in RxJS to see where things can be simplified or removed.

I wanted to put this PR while it's still in progress to see if I'm missing important things or changing stuff I really shouldn't.

With the current changes, the `dist/` folder has a slightly different layout:

```
/dist
├── /ajax
|  ├── LICENSE.txt
|  └── package.json
├── /fetch
|  ├── LICENSE.txt
|  └── package.json
├── /operators
|  ├── LICENSE.txt
|  └── package.json
├── /testing
|  ├── LICENSE.txt
|  └── package.json
├──/webSocket
|  ├── LICENSE.txt
|  └── package.json
├── /formats
|  ├── /cjs
|  ├── /esm2015
|  ├── /esm5
|  └── /typings
├── /bundles
|  ├── rxjs.umd.js
|  └── rxjs.umd.min.js
├── LICENSE.txt
├── package.json
└── README.md
```

The `ajax`, `fetch`, `operators`, `testing`, and `webSocket` folders remain mostly the same. They are secondary entry points and just point to the format specific files. There's no `index.js` in them anymore though.
```
{
  "name": "rxjs/ajax",
  "main": "../formats/cjs/ajax/index.js",
  "module": "../formats/esm5/ajax/index.js",
  "es2015": "../formats/esm2015/ajax/index.js",
  "typings": "../formats/typings/ajax/index.d.ts",
  "sideEffects": false
}
```

The `formats` folder contains all the different supported formats: `cjs`, `esm2015`, `esm5`, and `typings`. The UMD bundles are inside `bundles`. Maybe they should also be inside `formats` but since users do deep imports for them today, I left them in the same place.

Then there's the license, package.json and readme in the top level.

I think this format still caters to all the same consumers are before, but is simpler and requires moving less stuff around. Along the way I tried to remove scripts that weren't needed anymore. `.make-packages.js` is now much smaller and does all the build/packaging tasks.

I still have more scripts to look at and try to simplify but this is a start. 

@benlesh (and others) please let me know what you think.


.make-packages.js
```diff
@@ -1,136 +1,62 @@
...
-  es2015: './_esm2015/index.js'
+const { writeFileSync } = require('fs');
+const { cp, exec, ls, mkdir, rm } = require('shelljs');
+const { buildOptimizer } = require('@angular-devkit/build-optimizer');
+const pkg = require('./package.json');
+
+// TODO: 
+// - what about sourcemaps? 
+// Angular itself has inline sourcemaps. I don't see the point, since you don't have sources
+// on the package.
```
filipesilva: I'm not sure what to do about sourcemaps. I think the current ones are a bit weird and work only because the TS sourcefiles are bundled. We could inline sourcemaps but I'm not sure if that impacts parsing performance.

.make-packages.js
```diff
@@ -1,136 +1,62 @@
...
+// - what about shipping TS for bazel support? 
+// I think this isn't needed anymore?
```
filipesilva: Does RxJS still need to ship TS for Bazel support? I thought this wasn't needed anymore.

.make-packages.js
```diff
@@ -1,136 +1,62 @@
...
+rm('-rf', './dist');
+
+// Run TS to produce cjs, esm5, esm2015, and typings.
+console.log('# Compiling TypeScript sources.');
+exec('npm run tsc -- --build ./build-configuration/tsconfig.project.json', { silent: true });
+
+// Run BO over esm5 and esm2015 to remove side effects.
+console.log('# Removing side effects from esm5 and esm2015.');
+const buildOptimizerOptions = { emitSourceMap: false, isSideEffectFree: true };
+ls('./dist/formats/{esm5,esm2015}/**/*.js').forEach(inputFilePath => {
```
filipesilva: Running Build Optimizer over both ESM5 and ESM2015 should enable tree shaking for both (and prevent https://github.com/rollup/rollup/issues/2415).

benlesh:
Hey, @filipesilva, this is old, and probably fell through the cracks when I left Google. I appreciate the effort, but I think we've got other solutions in place for now. Is it okay if we close this one? Or do you still see opportunity to improve what we have?

filipesilva:
SGTM Ben, no worries at all.

Issue #3712, opened 2018-05-20
Invalid behavior of ThrottleTime operator


kzaher:
**RxJS version:**
5.5.8
6.1.0
rxjs-compat 6.1.0

**Code to reproduce:**

This is RxJs behavior

5.5.8
```js
const rx = require('rxjs');

const a = rx.Observable.timer(0, 700).throttleTime(1000, rx.Scheduler.async, { leading: true, trailing: true });
let lastDate = new Date();  
let d = a.subscribe(x => { 
    const now = new Date(); 
    console.log('\n' + x + ' ' + (now.getTime() - lastDate.getTime())); 
    lastDate = now; 
});
```

6.1.0
```js
const rx = require('rxjs');
require('rxjs/operators');
require('rxjs-compat/add/operator/throttleTime');

const a = rx.timer(0, 700).throttleTime(1000, rx.Scheduler.async, { leading: true, trailing: true });
let lastDate = new Date();  
let d = a.subscribe(x => { 
    const now = new Date(); 
    console.log('\n' + x + ' ' + (now.getTime() - lastDate.getTime())); 
    lastDate = now; 
});
```

**Expected behavior:**

This is underscore:

```js
const underscore = require('underscore');
const rx = require('rxjs');

let lastDate = new Date();  
d= rx.Observable.timer(0, 700).subscribe(underscore.throttle(x => { 
    const now = new Date(); 
    console.log('\n' + x + ' ' + (now.getTime() - lastDate.getTime())); 
    lastDate = now; 
}, 
1000, 
{ leading: true, trailing: true }))
```

```
11 1003
12 1002
/// missing 13, all intervals above 1 sec
14 1001
```

This is RxSwift behavior

```swift
var last = Date();
        _ = Observable<Int>.timer(0, period: 0.7, scheduler: MainScheduler.instance)
            .throttle(1.0, latest: true, scheduler: MainScheduler.instance)
            .subscribe({ next in
                let now = Date();
                print("\(next) \(now.timeIntervalSince(last))");
                last = now;
            })
```

```
next(11) 1.00001406669617
next(12) 1.00061988830566
// missing 13, all intervals above 1 sec
next(14) 1.00011909008026
```

**Actual behavior:**

```
11 1000
12 407 // <--- time interval less then 1 sec
13 1004 // <--- has 13
14 402
```

**Additional information:**

[...]

PR #4864, opened 2019-06-17
fix: evenly distribute values on throttleTime(fn, {leading: true, trailing: true})

sod:
fixes https://github.com/ReactiveX/rxjs/issues/2727
fixes https://github.com/ReactiveX/rxjs/issues/3712

reproduction (rxjs 6.5.1) https://stackblitz.com/edit/rxjs-tthiju

Look at the new test `it('should handle a busy producer emitting a regular repeating sequence'`

```typescript
      const e1 =   hot('abcdeabcdeabcdeabcdea|');
      const expected = 'a----a----a----a----a|';
```

That should be the expected behavior for `{leading: true, trailing: true}`. Thats the same behavior of https://lodash.com/docs/4.17.11#throttle and https://underscorejs.org/#throttle as well.

Right now throttleTime does it more like:
```typescript
      const e1 =   hot('abcdefabcdefabcdefabcdefa|');
      const expected = 'a----fa----fa----fa----fa|';
```

### DIFF
--- a/spec/operators/throttleTime-spec.ts
+++ b/spec/operators/throttleTime-spec.ts
@@ -141,16 +141,25 @@ describe('throttleTime operator', () => {
   });
 
   describe('throttleTime(fn, { leading: true, trailing: true })', () => {
-    asDiagram('throttleTime(fn, { leading: true, trailing: true })')('should immediately emit the first and last values in each time window', () =>  {
-      const e1 =   hot('-a-xy-----b--x--cxxx--|');
-      const e1subs =   '^                     !';
+    asDiagram('throttleTime(fn, { leading: true, trailing: true })')('should emit only one value in given time window', () =>  {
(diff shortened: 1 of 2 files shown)


sod:
@benlesh I seem to always come at the worst time ... when you prepare the next major version :D. But was there the possibility to include this in rxjs7?

MatthiasKunnen:
Is this not a duplicate of #4727?

sod:
Oh indeed it is. Haven't seen it. But https://github.com/ReactiveX/rxjs/pull/4727 schedules the timer after emitting the value. IMO it's important that scheduling always happens upfront, so user code (if slow) doesn't mess with the timings.

sod:
i give up again, I hope @MatthiasKunnen has more stamina

MatthiasKunnen:
I got you @sod. I wanna see this through.

PR #4865, opened 2019-06-17
feat(flow): add functional version pipe(name flow)

s97712:
**Description:**
Add functional version pipe for increase readability

**define:**
```ts
function flow(obs, ...operators)
function flow<T>(obs: Observable<T>): Observable<T>;
function flow<T, A>(obs: Observable<T>, op1: OperatorFunction<T, A>): Observable<A>;
function flow<T, A, B>(obs: Observable<T>, op1: OperatorFunction < T, A >, op2: OperatorFunction<A, B>): Observable<B>;
...
```

**usage:**
```ts

CreateEventObs<MouseEvent>(elm, "click").pipe(
  flatMap(evt => request(url, {
   ...
  }).pipe(
    catchError(() => ({error:"response error"}))
  ))
)

// replace with
flow(
  CreateEventObs<MouseEvent>(elm, "click"),
  flatMap(evt => flow(
    request(url, {
      ... 
    }),
    catchError(() => ({error:"response error"}))
  ))
)
```

Obviously, the function version is more readable.

### DIFF
--- a/spec/index-spec.ts
+++ b/spec/index-spec.ts
@@ -38,6 +38,7 @@ describe('index', () => {
     expect(index.pipe).to.exist;
     expect(index.noop).to.exist;
     expect(index.identity).to.exist;
+    expect(index.flow).to.exist;
   });
 
   it('should export error types', () => {
--- a/spec/util/flow-spec.ts
+++ b/spec/util/flow-spec.ts
@@ -1,0 +1,31 @@
+import { map, toArray} from 'rxjs/operators';
+import { flow, of } from 'rxjs';
+import { expect } from 'chai';
+
+declare const asDiagram: any;
+
+/** @test {flow */
+describe('flow', () => {
+  it('should be equal obs.pipe(...observable)', (done) => {
+    const e1 = of(1);
+    const obs_pipe = e1.pipe(
+      map(v => v + 1),
+      map(v => v.toString())
+    ).pipe(
+      toArray()
+    );
+
+    const obs_flow = flow(
+      e1,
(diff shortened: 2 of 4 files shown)


kwonoj:
I don't see strong reason this should goes in core as this is utility wraps existing pipe operator - this could be in userland.

s97712:
> I don't see strong reason this should goes in core as this is utility wraps existing pipe operator - this could be in userland if it's agreed to provide usefulness. Personally, I don't see this is obvious readable also - I'd expect start chain of pipe is static creation which gives me more explicit clarity.

The function version can display structure very clarity.

benlesh:
I toyed with this idea when `pipe` was added to RxJS back in 5.5. I'm generally with you, however, I thought at the time, and still think, it's a bit too much of a change for most people.

At the time, my idea was to name it `rx` for terseness like:

```ts
rx(
  source$,
  map(x => x + x),
  filter(x => x > 20),
)
```

However, the readability can suffer here.

```ts
const users = getUsers();
const namedBen = filter(x => x.name === 'Ben');
const loadData = concatMap(user => getUserData(user.id));

flow(
  users,
  namedBen,
  loadData
)

// vs

users.pipe(
  namedBen,
  loadData,
)
```

In the latter, it's clear to see we're acting upon `users` with two operator instances. In the former, it's a little more ambiguous.

I think I'm with @kwonoj that we can pass on this one for now, even though I generally like the pattern. It's probably something that can easily live in user land. If it becomes a widespread pattern, we can totally adopt it.

PR #4875, opened 2019-06-19
fix(pluck): param type

kirjahylly:
Fix: The parameter type of pluck does not accept numbers as array indices while the operator itself already supports arrays.

### DIFF
--- a/spec-dtslint/operators/pluck-spec.ts
+++ b/spec-dtslint/operators/pluck-spec.ts
@@ -37,6 +37,10 @@ it('should not accept empty parameter', () => {
   const a = of({ name: 'abc' }).pipe(pluck()); // $ExpectError
 });
 
-it('should accept string only', () => {
+it('should not accept a number when plucking an object', () => {
   const a = of({ name: 'abc' }).pipe(pluck(1)); // $ExpectError
 });
+
+it('should support arrays', () => {
+  const a = of(['abc']).pipe(pluck(0)) // ExpectType Observable<string>
+})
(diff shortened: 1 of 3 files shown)


src/internal/operators/pluck.ts
```diff
@@ -39,20 +39,20 @@ export function pluck<T, K1 extends keyof T, K2 extends keyof T[K1], K3 extends
...
+export function pluck<T, R>(...properties: Array<string | number>): OperatorFunction<T, R> {
   const length = properties.length;
   if (length === 0) {
     throw new Error('list of properties cannot be empty.');
   }
   return (source: Observable<T>) => map(plucker(properties, length))(source as any);
 }
 
-function plucker(props: string[], length: number): (x: string) => any {
+function plucker(props: Array<string | number>, length: number): (x: string) => any {
```
cartant: I realize that this isn't code you wrote, but `string` is not the correct type for `x` here. Could you change it to `any`?

spec-dtslint/operators/pluck-spec.ts
```diff
@@ -37,6 +37,15 @@ it('should not accept empty parameter', () => {
   const a = of({ name: 'abc' }).pipe(pluck()); // $ExpectError
 });
 
-it('should accept string only', () => {
+it('should not accept a number when plucking an object', () => {
   const a = of({ name: 'abc' }).pipe(pluck(1)); // $ExpectError
 });
+
+it('should support arrays', () => {
+  const a = of(['abc']).pipe(pluck(0)) // ExpectType Observable<string>
```
MateuszLeo: Missing `$`:

```suggestion
  const a = of(['abc']).pipe(pluck(0)) // $ExpectType Observable<string>
```

Issue #4847, opened 2019-06-08
windowTime race condition


Velenir:
## Bug Report

**Current Behavior**
`windowTime` for some parameter values  throws

```
Error in /turbo_modules/rxjs@6.5.2/internal/operators/windowTime.js (105:27)
Cannot read property 'closed' of undefined
```

Couldn't figure out the exact relation between parameters, but basically it reliably broke for me with `windowTimeSpan > windowCreationInterval && maxWindowSize > 2`,

**Reproduction**
https://stackblitz.com/edit/rxjs-gajdpw?devtoolsheight=60

```js
import { interval } from 'rxjs';
import { windowTime, mergeAll } from 'rxjs/operators';

const timer = interval(200)
const result = timer.pipe(
  windowTime(1000, 600, 4),
  mergeAll(),
);
result.subscribe(x => console.log(x));
```

**Expected behavior**
Not breaking. Not sure as to exact console output, though. Same as with `bufferTime`? I am pretty new to RxJS, just experimenting with operators.

**Environment**
- Runtime: Chrome 76.0.3806.1 (Official Build) dev (64-bit)
- RxJS version: 6.5.2

**Additional context/Screenshots**
I'm assuming `windowTime` tries to close a window stream before it creates one, so a race condition?

githubxiaowen:
check the [rxViz](https://rxviz.com/v/9J9zyRG8) here.

Frame:0
windows: [w1]
w1: [0]

Frame:1
windows: [w1]
w1:[0,1]

Frame: 2
windows: [w1,w2]
w1: [0,1,2]
w2:[2]

Frame: 3
w1:[0,1,2,3]
w2: [ 2, 3]

``` ts
  protected _next(value: T): void {
    const windows = this.windows;
    const len = windows.length;
    for (let i = 0; i < len; i++) {
      const window = windows[i];
      if (!window.closed) {
        window.next(value);
        if (window.numberOfNextedValues >= this.maxWindowSize) {
          this.closeWindow(window);
        }
      }
    }
  }
```
so , after w1 emitted the value 3, it will do the action closeWindow(window)， and remove the w1 window from windows。

``` ts
  public closeWindow(window: CountedSubject<T>): void {
    window.complete();
    const windows = this.windows;
    windows.splice(windows.indexOf(window), 1);
```
now back the `_next` method， since the windows changed to [w2]， look at our code

```ts
// i ==> 1
const window = windows[i]
```
[...]

PR #4883, opened 2019-06-24
fix(windowTime): fix windowTime bug

githubxiaowen:
**Description:**

this line of code cause the bug
```js
const windows = this.windows
```

#4847


githubxiaowen:
leave a comment in case for being locked

cartant:
Could you add a failing test that reproduces the problem and shows that this change fixes it? Ideally, IMO, bug fix PRs should consist of two commits: one that adds a failing test; and another that includes the change that fixes the bug and makes the test pass.

cartant:
Closed by #5245 - which added a test.

Issue #4890, opened 2019-06-28
docs(resultSelector) description for resultSelector doesn't exist


Eddie-CooRo:
### Documentation Related To Components:
[bindCallback](https://rxjs-dev.firebaseapp.com/api/index/function/bindCallback), [bindNodeCallback](https://rxjs-dev.firebaseapp.com/api/index/function/bindNodeCallback), [fromEvent](https://rxjs-dev.firebaseapp.com/api/index/function/fromEvent) and [fromEventPattern](https://rxjs-dev.firebaseapp.com/api/index/function/fromEventPattern)

### Please check those that apply

- [ ] typo
- [x] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

### Description Of The Issue
There is no description, nor example nor deprecation note exists for resultSelector argument in any of the above components.
The only component that has a description for the resultSelector argument is [generate](https://rxjs-dev.firebaseapp.com/api/index/function/generate), But its document is broken. (#4889)
I can copy this description to every place that is needed and open a PR for that if you want.

Eddie-CooRo:
Going to close this.
Check https://github.com/ReactiveX/rxjs/pull/4892#issuecomment-506882156

PR #4892, opened 2019-06-28
docs(combineLatest) Use project function instead of map operator

Eddie-CooRo:
**Description:**
"Use project function to dynamically calculate the Body-Mass Index" example in the combineLatest documentations was using "map" operator instead of the project function.

None

### DIFF
--- a/src/internal/observable/combineLatest.ts
+++ b/src/internal/observable/combineLatest.ts
@@ -196,9 +196,7 @@ export function combineLatest<R>(...observables: Array<ObservableInput<any> | ((
  *
  * const weight = of(70, 72, 76, 79, 75);
  * const height = of(1.76, 1.77, 1.78);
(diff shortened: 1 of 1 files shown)


cartant:
Thanks. This is a problem, but it's the text that needs changing - not the example - as passing a result selector is deprecated:

https://github.com/ReactiveX/rxjs/blob/a9fa9d421d69e6e07aec0fa835b273283f8a034c/src/internal/observable/combineLatest.ts#L32-L44

Eddie-CooRo:
Oops!
So we should close #4890 too?

cartant:
> So we should close #4890 too?

Yeah, I think so. BTW, an explanation for the deprecations is here: https://github.com/ReactiveX/rxjs/blob/master/docs_app/content/guide/v6/migration.md#result-selectors-removed-or-deprecated

And thanks again for the work you are doing.

PR #4893, opened 2019-06-28
Gracefully handle null schedulers in `delay` operator

bjohnson-va:
### Issue
Passing a `null` `TestScheduler` to `delay` causes unhandled errors.

### Background

As I see it, there are two rational outcomes when a null value is passed
to the delay function for the scheduler argument.

1) Throw a clear exception stating that null values are not allowed
2) Accept null, but treat it as `undefined`, using the default `async`
   in its stead.

The current behaviour is neither.  It throws the unhandled error:
```
TypeError: Cannot read property 'now' of undefined
```

### Solution
I'm proposing that we opt for option 2.

### DIFF
--- a/spec/operators/delay-spec.ts
+++ b/spec/operators/delay-spec.ts
@@ -176,4 +176,10 @@ describe('delay operator', () => {
 
     expectObservable(result).toBe(expected);
   });
+  it('should not throw when provided scheduler is null', () => {
+    const input = cold('a|');
+    const sched: TestScheduler = null;
+    const observable = input.pipe(delay(10, sched));
+    expect(() => observable.subscribe()).to.not.throw();
+  });
 });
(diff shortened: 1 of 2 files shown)


benlesh:
Core Team Notes: This is just how optional arguments and default arguments work in JavaScript, and we should not be trying to support `null | undefined`. It's easy enough to convert to `undefined` and let defaults work the way they're supposed to work in JavaScript.

Thank you so much for your contribution, @bjohnson-va ... we're going to close this one though.

bjohnson-va:
@benlesh My main concern is that an internal error is returned from the delay function. As a user of the library, when I see `Cannot read property 'now' of undefined`. It is not immediately obvious what the actual problem is without knowing intimate details about the internal workings of RXJS.

Would you consider my other suggestion?
> Throw a clear exception stating that null values are not allowed

This way, the library is not *supporting* null, but it is providing runtime feedback that makes it easier to use.

Issue #4908, opened 2019-07-12
Marbles Testing: Missing create TestScheduler assertion causes silent fail of .run() tests


tomwhite007:
## Bug Report

**Current Behavior**
A boilerplate assertion is required inside the callback from a creation of a TestScheduler object. 
```ts
  const testScheduler = new TestScheduler((actual, expected) => {
    expect(actual).toEqual(expected);
  });
```
If this is missed out by someone looking to remove unnecessary boilerplate (like me), Marbles Testing assertions inside testScheduler.run() pass **or fail** silently because the external test framework (in my case Jasmine), doesn't wrap them as part of its tests. They become orphan functions that just run and report nothing.

**Reproduction**

```ts
  it('test should fail but actually passes', () => {
    const testScheduler = new TestScheduler(() => {});

    testScheduler.run(({ expectObservable, cold }) => {
      const obs$ = cold('--a--b|', { a: 'x', b: 'y' });

      expectObservable(obs$).toBe('-|');
    });
  });
```

**Expected behavior**
Ideally an error would be thrown if this assertion was missed, but I assume that's not possible. So, couldn't there be a lint rule to do it instead?

**Environment**
- Test framework: Jasmine (but I'm pretty sure others behave the same)
- RxJS version: 6.5.2

cartant:
I don't see how this is a bug. This implementation is deliberate; it's like this to avoid a dependency upon any particular test framework. And the documentation - IIRC - is clear on this point.

You've chosen to pass a noop.

tomwhite007:
Hi @cartant 
You're right it isn't strictly a bug, but it is a vulnerability of sorts.
Just wanted to raise the topic as a conversation to help awareness.
Happy to close this off.

PR #4909, opened 2019-07-12
Marbles testing docs - patch syntax error

tomwhite007:
**Description:**

There is a syntax error in a marbles string in my previously accepted PR - that means the test would fail.

This PR is to correct that, and add a comment in another code example that I believe would have saved this from happening (see related issue)

**Related issue:**

[Marbles Testing: Missing create TestScheduler assertion causes silent fail of .run() tests #4908](https://github.com/ReactiveX/rxjs/issues/4908)


docs_app/content/guide/testing/marble-testing.md
```diff
@@ -212,12 +213,12 @@ For example:
 ```ts
   let eventCount = 0;
 
-  const s1 = cold('--a--b|', { a: 'x', b: 'y' });
+  const s1 = cold('--0--1|', { 0: 'x', 1: 'y' });
```
benlesh: revert this.

docs_app/content/guide/testing/marble-testing.md
```diff
@@ -212,12 +213,12 @@ For example:
...
 
   // side effect using 'tap' updates a variable
   const result = s1.pipe(tap(() => eventCount++));
-  
-  expectObservable(result).toBe('--a--b|', ['x', 'y']);
+
+  expectObservable(result).toBe('--0--1|', ['x', 'y']);
```
benlesh:
```suggestion
  expectObservable(result).toBe('--a--b|', { a: 'x', b: 'y' });
```

tomwhite007: @benlesh thanks for your comments. Happy to do it that way.
I've added all the changes you requested as one commit.

Issue #4889, opened 2019-06-28
Docs(generate) Overloads section is broken


Eddie-CooRo:
### Please check those that apply

- [ ] typo
- [ ] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

### Description Of The Issue
The overloads section for the "[generate](https://rxjs-dev.firebaseapp.com/api/index/function/generate)" function is broken.

benlesh:
Hi @eddiecooro — thank you for taking the time to open this issue and for sharing your thoughtful feedback with the RxJS project. I’m sorry it took us so long to respond. Given how much time has passed and the lack of recent activity, I’m going to close this for now so the tracker reflects the work we’re likely to act on. If this is still relevant on a current RxJS release, please feel welcome to open a new issue with updated context and a reproduction. Thank you again for your contribution and patience.

— Ben

Issue #4672, opened 2019-03-29
[docs]: Operator `generate` need a total over whole of the documentation


BioPhoton:
### Please check those that apply

- [ ] typo
- [x] documentation doesn't exist
- [x] documentation needs clarification
- [x] error(s) in example
- [ ] needs example

### Description Of The Issue

The whole documentation of [generate](https://rxjs.dev/api/index/function/generate) is .... a mess :-p

I guess because it is not yet refactored to the new docs style. 

I would love to file a PR of this. Is there a good example of what to do where to start what structure or in which order all the different chapters should ideally be?

If this not exist yet I will create a doc for refactoring the docs :)

Please let me know how/where to get my missing info.

JWO719:
For now there isn't such a document, but I think it would be really helpful for "first timers". I'm not sure what's the easiest way to add such a document and provide you with all the needed informations, but I would love to support you with this!

niklas-wortmann:
closed by #5006

PR #4911, opened 2019-07-12
docs(<Observable>): brought 'generate' docs to new docs format

Ore4444:
'generate' docs looking good.
Only thing, I couldn't get the "description" heading to appear like in all the other pages.
it's the only page that needs to maintain separate docblocks for different overloads. maybe the template needs to support this use case?

### DIFF
--- a/src/internal/observable/generate.ts
+++ b/src/internal/observable/generate.ts
@@ -47,27 +47,12 @@ export interface GenerateOptions<T, S> extends GenerateBaseOptions<S> {
 }
 
 /**
- * Generates an observable sequence by running a state-driven loop
- * producing the sequence's elements, using the specified scheduler
- * to send out observer messages.
- *
- * ![](generate.png)
- *
- * @example <caption>Produces sequence of 0, 1, 2, ... 9, then completes.</caption>
- * const res = generate(0, x => x < 10, x => x + 1, x => x);
- *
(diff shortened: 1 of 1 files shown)


Ore4444:
Closes:
#4889 
#4672

src/internal/observable/generate.ts
```diff
@@ -47,27 +47,12 @@ export interface GenerateOptions<T, S> extends GenerateBaseOptions<S> {
...
- * @example <caption>Produces sequence of 0, 1, 2, ... 9, then completes.</caption>
- * const res = generate(0, x => x < 10, x => x + 1, x => x);
- *
- * @example <caption>Using asap scheduler, produces sequence of 2, 3, 5, then completes.</caption>
- * const res = generate(1, x => x < 5, x => x * 2, x => x + 1, asap);
- *
- * @see {@link from}
- * @see {@link Observable}
- *
  * @param {S} initialState Initial state.
```
benlesh: We can actually remove all the type information like `{S}` as well. We're gleaning that stuff from TypeScript now.

Ore4444: remove just the type part of the overload docblocks? (e.g. `{S}`)
or remove all overload docblocks (e.g. `@param {S} initialState Initial state.`)?

JWO719: just the {s}, sorry for the super late reply

Ore4444:
Solved already by #5006 
Great job!

PR #4912, opened 2019-07-13
test(Subscription): Update subscription tests

NothingEverHappens:
* Cleaned up wording
* Moved positive cases to the top
* Grouped unsubscribe() tests together


benlesh:
Looks like there's a problem with lint, @NothingEverHappens. Otherwise I think this is okay.

NothingEverHappens:
Thanks, fixed

PR #4919, opened 2019-07-19
Add subscription via Observable overload to Observable.subscribe

DanielMSchmidt:
This works implementation wise but is not allowed based from the types.

**Description:**

I noticed that a valid way to subscribe to an Observable is to run `Observable.subscribe(otherObservable)`. While this works in the JS world, the Typescript compiler errors due to a missing type overload. I added it, I hope I did not overlook any deprecation notice.

### DIFF
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -71,6 +71,7 @@ export class Observable<T> implements Subscribable<T> {
   }
 
   subscribe(observer?: PartialObserver<T>): Subscription;
+  subscribe(observable?: Observable<T>): Subscription;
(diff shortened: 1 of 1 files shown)


cartant:
Passing an `Observable` to `subscribe` is not semantically valid. You can pass a `Subject` - which is both an `Observable` and an `Observer` - but you cannot pass any old `Observable`.

kwonoj:
@cartant is correct. Even passing observable works, that may comes with internal implementation and it is noy expected to gauranteed to work. Subscriber should have observer.

PR #4929, opened 2019-07-24
chore(testing): Make testing easier

MatthiasKunnen:
**Description:**
While working on operators I've had difficulty with running tests. I was missing the ability to autorun tests on change (watch) and run only specific tests to avoid long run times. This PR adds this functionality.


kwonoj:
I don't have issue with crux of PR, but bit uncomfortable having another js script for spawning test + having 2 more npm script to support this.

MatthiasKunnen:
The extra npm script is just to use the same base as testing all and be DRY.

About the JS script, it's the only way I've found to create a readable specific test command. Are you thinking of another approach?

kwonoj:
> The extra npm script is just to use the same base as testing all and be DRY.

I am aware, but also given current number of npm scripts we are trying to not to grow if unless absolutely necessary.

I don't have immediate proposal: just first sight I can see those two and would hope to avoid it to push this forward.

MatthiasKunnen:
I can also remove test_base but duplicate the command partially

MatthiasKunnen:
@kwonoj, which other maintainers can look at this PR? Maybe they have a better approach

kwonoj:
I wrote most of script & test setups.

MatthiasKunnen:
The thing is, without a script to run the tests, `npm run test_specific -- throttleTime,auditTime` turns into `npm run test_base -- "spec/**/throttleTime-spec.ts" "spec/**/auditTime-spec.ts"` which is very elaborate. So this could be accomplished without script but it would be cumbersome.

I can remove the `test_specific` npm script and make devs execute `node ./tools/run-specific-tests.js throttleTime,auditTime` instead of `npm run test_specific -- throttleTime,auditTime`. That means one less npm script but a bit more cumbersome execution. It will also have a negative effect on readability/conventions since devs will look for test commands in the scripts section of `package.json`.

kwonoj:
I do agree some dev might want different dev workflow than current, but for now I value to not increase our script / or other as much as I can especially we are preparing major overhaul around whole build / test script for upcoming major versions.

kwonoj:
so if we could achieve same even if it's bit cumbersome without adding up current script, I'd pursue it.

benlesh:
@MatthiasKunnen thank you so much for your effort here. Very sorry we let this sit. I ended up adding a couple of watch modes recently. IIRC our build process was in a bit of flux at the time. Thanks again.

PR #4934, opened 2019-07-26
docs(operators): fixed references to marble diagrams

claasahl:
**Description:**
Noticed that the documentation for [`tap`](https://rxjs.dev/api/operators/tap) references a missing marble diagram.

n/a

### DIFF
--- a/src/internal/observable/pairs.ts
+++ b/src/internal/observable/pairs.ts
@@ -8,7 +8,7 @@ import { Subscription } from '../Subscription';
  *
  * <span class="informal">Turn entries of an object into a stream.</span>
  *
- * <img src="./img/pairs.png" width="100%">
+ * <img src="/assets/images/marble-diagrams/pairs.png" width="100%">
  *
  * `pairs` takes an arbitrary object and returns an Observable that emits arrays. Each
  * emitted array has exactly two elements - the first is a key from the object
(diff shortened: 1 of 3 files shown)


src/internal/observable/pairs.ts
```diff
@@ -8,7 +8,7 @@ import { Subscription } from '../Subscription';
  *
  * <span class="informal">Turn entries of an object into a stream.</span>
  *
- * <img src="./img/pairs.png" width="100%">
+ * <img src="/assets/images/marble-diagrams/pairs.png" width="100%">
```
JWO719: This should be changed to just 'pairs', similar to like you did for `tap`

claasahl: done

src/internal/observable/pairs.ts
```diff
@@ -8,7 +8,7 @@ import { Subscription } from '../Subscription';
  *
  * <span class="informal">Turn entries of an object into a stream.</span>
  *
- * <img src="./img/pairs.png" width="100%">
+ * <img src="pairs.png" width="100%">
```
JWO719: I'm really sorry, I didn't expressed well.
It should be like this:

claasahl: makes sense. done.

claasahl: I came across another img-tag in the `generate` operator

PR #4936, opened 2019-07-27
fix(pipe): Special handling for 0-arg case.

### LINKED ISSUE #4933: Nitpick: Shouldn't Rx.pipe with no arguments return identity instead of noop?
https://github.com/ReactiveX/rxjs/blob/6.5.2/src/internal/util/pipe.ts#L25
```ts
if (!fns) {
    return noop as UnaryFunction<any, any>;
}
```

The composition of zero things should be the identity under composition, which is `Rx.identity`, not `Rx.noop`. That allows `pipe(pipe(), pipe(f))` to be equivalent to `pipe([].concat([], [f])`, which is equivalent to `f`.

I'm new to RxJS, so I don't know if people rely on this behavior.

leewz:
Optimize calling `pipe` with 0 args.

Closes #4933

There is an `if (!fns)` check in the `pipeFromArray` helper, and it results in `noop`. The test for `pipe` says the 0-arg case returns a no-op, but it checks for `identity`. The test is probably meant to test for that check, but the check is for the lack of an args array, so I added a check for that test to test and fixed the test's label.

The `if (!fns)` check might not be necessary at all. It is unreachable from client code, since `pipeFromArray` is internal, and is only used from `pipe`, which will never pass null/undefined.

### DIFF
--- a/src/internal/util/pipe.ts
+++ b/src/internal/util/pipe.ts
@@ -1,4 +1,5 @@
 import { noop } from './noop';
+import { identity } from './identity';
 import { UnaryFunction } from '../types';
 
 /* tslint:disable:max-line-length */
@@ -23,6 +24,10 @@ export function pipe(...fns: Array<UnaryFunction<any, any>>): UnaryFunction<any,
 export function pipeFromArray<T, R>(fns: Array<UnaryFunction<T, R>>): UnaryFunction<T, R> {
   if (!fns) {
     return noop as UnaryFunction<any, any>;
+  }
+
+  if (fns.length === 0) {
+    return identity as UnaryFunction<any, any>;
   }
 
   if (fns.length === 1) {
--- a/src/internal/util/pipe.ts
+++ b/src/internal/util/pipe.ts
@@ -22,10 +22,6 @@ export function pipe(...fns: Array<UnaryFunction<any, any>>): UnaryFunction<any,
 
 /** @internal */
 export function pipeFromArray<T, R>(fns: Array<UnaryFunction<T, R>>): UnaryFunction<T, R> {
-  if (!fns) {
-    return noop as UnaryFunction<any, any>;
-  }
-
   if (fns.length === 0) {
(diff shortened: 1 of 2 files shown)


src/internal/util/pipe.ts
```diff
@@ -25,6 +26,10 @@ export function pipeFromArray<T, R>(fns: Array<UnaryFunction<T, R>>): UnaryFunct
     return noop as UnaryFunction<any, any>;
   }
 
+  if (fns.length === 0) {
+    return identity as UnaryFunction<any, any>;
+  }
+
```
cartant: IMO, the previous test should be included in this `if` statement and both should return `identity`:

```
  if (!fns || fns.length === 0) {
    return identity as UnaryFunction<any, any>;
  }
```
I can see no reason whatsoever for returning a no-op. Having different behaviour for no-array and empty-array scenarios makes little sense.

I think you should do what it was that you mentioned in the issue you raised and return `identity`. The code might not be reachable ATM, but if the `pipeFromArray` is used elsewhere in the future, it should behave in a predictable way.

leewz:
The right predictable behavior may be to throw an error. The sum of an
empty collection of numbers is 0, but what is the sum of a null of numbers?

>

cartant:
> The right predictable behavior may be to throw an error.

I'm fine with it throwing an error. I just don't think it should return a no-op.

Given that the type signature requires an array to be passed, I think removing the entire `if (!fns)` statement. If `fns` is `undefined` or `null`, it will throw when `fns.length` is evaluated. And that's fine by me.

leewz:
Removed the check for nonexistent array argument.

src/internal/util/pipe.ts
benlesh: fixed it myself.

benlesh:
Oh, I see, we wanted to keep the error on the `fn.length` call in that case ... I'll revert my change. It's a weird corner case anyhow.

PR #4944, opened 2019-08-04
4842 suggestions for debounce documentation improvements

### LINKED ISSUE #4842: [docs]: Suggestions for `debounce` documentation improvements

The `debounce` operator: 
https://github.com/ReactiveX/rxjs/blob/9026a36f6fd188a7b2620ad7b8fd0acea5921f18/src/internal/operators/debounce.ts

- [ ] typo
- [X] error(s)/adoption/updates in example
- [ ] needs example

### Description Of The Issue

The description should include also cases for complete and error situations as they are not trivial in their impact as well as pretty different from other rate-limiting operators. 
Also, the example could be adapted to show a more dynamic creation of the `durationSelector`  and the see also section is missing some references.

We might want to improve this to serve a better more detailed description for the community.

#### Description
What is missing is the case for completion of the source Observable. Here we will receive the latest cached value occurred in the `durationSelector`.
In error case, we get returned the error and drop the cached value.

#### Example
We could show that the `durationSelector` and therefor it's debounce time can change on every input.

#### See Also
In the see also section we miss the other rate-limiting operators.

The current list looks like this:
https://github.com/ReactiveX/rxjs/blob/9026a36f6fd188a7b2620ad7b8fd0acea5921f18/src/internal/operators/debounce.ts#L45-L48

A complete list for rate-limiting operators could look like this:
- audit
- auditTime
- debounce
- debounceTime
- sample
- sampleTime
- throttle
- throttleTime

I would suggest removing timing `delayWhen` operator from the list as it belongs to another group. (even if debounce, audit and sample delay the output of values)

Otherwise, we should also list `delay` and maybe other timing related operators like `timer`, `interval` etc.

BioPhoton:
**Description:**
Improve documentation for debounce

Closes #4842


JWO719:
Something is weird here, there isn't any change? @BioPhoton could you have a look?

JWO719:
Because this is an empty pr (for some reason) I am going to close this pr. If there is any need to improve the docs for `debounce` I am happy to review any new pr on that.

Issue #4382, opened 2018-11-25
[docs]: Create a design system for marble diagrams - a standard for marble diagrams


BioPhoton:
# (UPDATE): I started to collect all outcome and questions here:    
# [Rx-Marble-Design-System](https://github.com/BioPhoton/Rx-Marble-Design-System) 

- [ X ] documentation doesn't exist
- [ X ] documentation needs clarification

### Description Of The Issue

Why is this important?

Marble diagrams where ever since a very practical and easy way to explain the behavior of streams in a visual way. There are a lot of diagrams out there, but they are not fully explaining the complex behavior or more important internal workflows. Also, there are different scenarios for some operators that are only possible to explain in diagrams if we would have internal workflows visible.
Furthermore, there is a lack of a standard for these diagrams. 

What I want to create is:
A public open source standard of drawing marble diagrams
An easy and more important quick way for everyone to create marble diagrams 
If the standard is done I will create diagrams for all operators and their different scenarios

Here is where I started with: https://docs.google.com/presentation/d/1LlVmTEcoc1LRV00Tp-rDp4kopI-NobC9S2w-w5pC_JA/edit?usp=sharing

As you can see there are a lot of questions to answer.
What I would need is people that help me to discuss several edge cases to generate this standard.
Also if the standard is finished I would need a second pair of eyes that help me to double check the drawings for all the operators.

kwonoj:
This is quite interesting but I am bit unsure if we can define something. We define semantics of marbles, but does not define how does it visualized: i.e even in ascii marble we allow all chars as marble tokens vice versa. Also even if we have something not sure if we can enforce to everyone use those instead of each user's page design.

JWO719:
Hey there, 
I really love that idea, especially for the docs web page. Ideally it comes with consistency and clarity. 

@kwonoj Maybe I get you toally wrong, and for sure we can't enforce anyone to use this kind of semantic, but is your point really just about semantic? I think we can define some kind of standard. I think we already did in some way (see http://reactivex.io/rxjs/manual/overview.html#marble-diagrams)

Anyway I definetly think, that this is a topic we should discuss at the next core team meeting!

kwonoj:
[...]

PR #4955, opened 2019-08-07
Update all diagrams for transformation operators

BioPhoton:
**Description:**
Update marble diagrams for all transform operators.

Update marble diagrams for all transform operators.
All images are in the same format as the original images. All file names are identical with the original one.

Included operators are:
- map
- mapTo
- pluck
- timestamp
- timeInterval
- pairs
- materialize
- dematerialize

#4382


benlesh:
I noticed that these are missing the completion information that was conveyed in the old diagrams with a `-|->`.

BioPhoton:
Excellent feedback! I will check all other diagrams and align them with the original information.
If I change something I will comment it.

niklas-wortmann:
IIRC we put these marble diagrams on hold for the docs, therefore, I close all the related prs. If I got this wrong, feel free to reopen

PR #4961, opened 2019-08-11
fix(types): add Boolean signature to filter

### LINKED ISSUE #4959: [TS] `filter(Boolean)` loses type information, TypeScript 3.5 makes it worse
## Bug Report
More of a heads up, than a bug report, but this has been a major PITA for upgrading Google internal to TS 3.5, so I wanted to have a discussion here too.

**Current Behavior**
With TS3.4 `pipe(filter(Boolean))` loses type information, everything after it is of type `any`. See stackblitz below:

```ts
const source = of('World').pipe(
  map(x => `Hello ${x}!`),
  filter(Boolean),
  map(y => y.madeup)  // no type error
);
```

The type of Boolean in `lib.d.ts` is
```ts
interface BooleanConstructor {
    new (value?: any): Boolean;
    (value?: any): boolean;
    readonly prototype: Boolean;
}
```

With TS3.5 Boolean is typed as 

```ts
interface BooleanConstructor {
    new (value?: any): Boolean;
    <T>(value?: T): boolean;
    readonly prototype: Boolean;
}
```
https://github.com/microsoft/TypeScript/blob/master/lib/lib.es5.d.ts#L536

However, in most cases it appears `<T>` is not inferrable (I don't blame TS, this is really round-about), so TS3.5 pick the new default `unknown`, which in turn makes the type for the next operation `unknown` from previously `any` causing many compilation failures.

**Reproduction**
https://stackblitz.com/edit/rxjs-erah82

**Possible Solution**
I have been recommending replacing `filter(Boolean)` with the following more explicit workaround `filter((x): x is OutputType => !!x)` where `OutputType` is explicitly spelled out the expected type in the next operation in the chain.

Or maybe once https://github.com/microsoft/TypeScript/issues/16655#issuecomment-494907057 lands `filter(Boolean)` will just work for Rxjs.

cartant:
Closes #4959

**Description:**

This PR adds another signature overload to `filter` to match `Boolean` predicates. This solves the problem identified in #4959 and is sufficiently specific that I cannot foresee it effecting any problems.

I've added a conventional test and a dtslint test.

Other operators that have a similar issue are: `find`, `first`, `last` and `takeWhile`. If this fix is approved, I'll have a look at what can be done for those (`first` and `last` are a little more complicated, as they support a default value).

@benlesh @rkirov

**Related issue (if exists):** #4959

### DIFF
--- a/spec-dtslint/operators/filter-spec.ts
+++ b/spec-dtslint/operators/filter-spec.ts
@@ -38,3 +38,9 @@ it('should enforce user-defined type guard types', () => {
   const o = of(1, 2, 3).pipe(filter((value: string): value is '1' => value < '3')); // $ExpectError
   const p = of(1, 2, 3).pipe(filter((value: number, index): value is 1 => index < '3')); // $ExpectError
 });
+
+it('should support Boolean as a predicate', () => {
+  const o = of(1, 2, 3).pipe(filter(Boolean)); // $ExpectType Observable<number>
+  const p = of(1, null, undefined).pipe(filter(Boolean)); // $ExpectType Observable<number>
+  const q = of(null, undefined).pipe(filter(Boolean)); // $ExpectType Observable<never>
+});
--- a/spec/operators/filter-spec.ts
+++ b/spec/operators/filter-spec.ts
@@ -330,4 +330,13 @@ describe('filter operator', () => {
 
     // tslint:disable enable
   });
+
+  it('should support Boolean as a predicate', () => {
+    const source = hot('-t--f--^-t-f-t-f--t-f--f--|', { t: 1, f: 0 });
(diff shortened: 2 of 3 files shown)


rkirov:
Looks good to me, I will downstream this to google internal to unblock the TS3.5 migration.

cartant:
> Should we also do this for other operators with predicates? Like `first`, `last`, etc?

Yeah. We should use this approach for `find`, `first`, `last` and `takeWhile` - those are the only other operators that accept a user-defined type guard. I was waiting to hear back from Rado, as some of those other operators are a little more complicated - as they support default values - and didn't want to prep type changes if they weren't gonna do the job. But if the Google code base is now synced, I guess we should look at change 'em, too.

Issue #4879, opened 2019-06-20
forkJoin incorrect result in marble diagram image


mathew-pigram:
### Documentation Related To Component:

forkJoin - https://rxjs.dev/api/index/function/forkJoin

### Please check those that apply

- [X ] typo
- [ ] documentation doesn't exist
- [ ] documentation needs clarification
- [X ] error(s) in example
- [ ] needs example

### Description Of The Issue
Result marble, in the marble diagram image in the description section, has the wrong value.

In the description section, a marble diagram image is displayed.
The three input Observables end with values e, j, and 4 respectively.
The result would be an array [e, j, 4], but the image shows [e, 4, j]

![forkJoin](https://user-images.githubusercontent.com/29947725/59845454-9bd32c80-939c-11e9-818a-fbfab0f735be.png)

PR #4962, opened 2019-08-11
refactor(forkJoin): remove redundant forkJoin spec

martinsik:
This is an addition to #4879. I've noticed there's some forgotten [`/spec/operators/forkJoin-spec.ts`](https://github.com/ReactiveX/rxjs/blob/master/spec/operators/forkJoin-spec.ts) that was overriding the marble diagram from the correct `forkJoin` spec here [`/spec/observables/forkJoin-spec.ts`](https://github.com/ReactiveX/rxjs/blob/master/spec/observables/forkJoin-spec.ts). But the old diagram is probably more descriptive (this is the one use at https://rxjs.dev/api/index/function/forkJoin right now):

```
  asDiagram('forkJoin')('should emit the last emitted value from each observable', () => {
    const e1 =   hot('-a--b-----c-d-e-|');
    const e2 =   cold('--1--2-3-4---|   ');
    const e3 =   hot('--------f--g-h-i--j-|');
    const expected = '--------------------(z|)';
    const expectedValue = {
      z: ['e', '4', 'j']
    };
    const result = forkJoin(e1, e2, e3);
    expectObservable(result).toBe(expected, expectedValue);
  });
```

So I pasted this initial test over the current one in [`/spec/observables/forkJoin-spec.ts`](https://github.com/ReactiveX/rxjs/blob/master/spec/observables/forkJoin-spec.ts) and just switched the order of source Observables because "hot" Observable as always listed first so the marble diagram will match the code. https://github.com/ReactiveX/rxjs/blob/master/spec/helpers/tests2png/diagram-test-runner.ts#L15

Related issue (if exists): #4879


spec/observables/forkJoin-spec.ts
```diff
@@ -11,13 +11,14 @@ declare const asDiagram: Function;
 describe('forkJoin', () => {
   asDiagram('forkJoin')
     ('should join the last values of the provided observables into an array', () => {
-      const e1 = forkJoin(
-                   hot('---a---b---c---d---|'),
-                   hot('-1---2---3---|')
-      );
-      const expected = '-------------------(x|)';
+      const e1 = forkJoin([
```
benlesh: please be sure the marble diagrams line up. Until we get our tests updated to use run mode, that means simply adding (gross looking) spaces to make sure the first single quotes align in the strings.

martinsik: Fixed

cartant:
Could you please rebase this onto `master` so that the Node types are updated, the TypeScript version 3 tests pass and this can be merged?

martinsik:
@cartant I messed the rebase so I had to force push it but it should be fine now

cartant:
No worries. Thanks for fixing it up.

PR #4967, opened 2019-08-13
docs(EMPTY): fix broken link, updated description

nikagar4:
**Description:**

At some point we’d need to move `empty()` examples to `EMPTY` so I just moved it now,
since I already moved my codebase to `EMPTY` and there’s still no description in official docs

### DIFF
--- a/src/internal/observable/empty.ts
+++ b/src/internal/observable/empty.ts
@@ -4,6 +4,36 @@ import { SchedulerLike } from '../types';
 /**
  * The same Observable instance returned by any call to {@link empty} without a
  * `scheduler`. It is preferrable to use this over `empty()`.
+ *
+ * ## Examples
+ * ### Emit the number 7, then complete
+ * ```ts
+ * import { EMPTY } from 'rxjs';
+ * import { startWith } from 'rxjs/operators';
+ *
+ * const result = EMPTY.pipe(startWith(7));
+ * result.subscribe(x => console.log(x));
+ * ```
+ *
+ * ### Map and flatten only odd numbers to the sequence 'a', 'b', 'c'
+ * ```ts
+ * import { EMPTY, interval, of } from 'rxjs';
+ * import { mergeMap } from 'rxjs/operators';
+ *
+ * const interval$ = interval(1000);
+ *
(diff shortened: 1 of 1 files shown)


src/internal/observable/empty.ts
```diff
@@ -57,7 +87,7 @@ export const EMPTY = new Observable<never>(subscriber => subscriber.complete());
  * the emission of the complete notification.
  * @return An "empty" Observable: emits only the complete
  * notification.
- * @deprecated Deprecated in favor of using {@link EMPTY} constant, or {@link scheduled} (e.g. `scheduled([], scheduler)`)
+ * @deprecated Deprecated in favor of using {@link index/EMPTY} constant, or {@link scheduled} (e.g. `scheduled([], scheduler)`)
```
nikagar4: If there's better way to refer to right link feel free to correct me.

Didn't have much time to dig into dgeni to fix this issue

benlesh: cc @JWO719

src/internal/observable/empty.ts
```diff
@@ -4,6 +4,36 @@ import { SchedulerLike } from '../types';
 /**
  * The same Observable instance returned by any call to {@link empty} without a
  * `scheduler`. It is preferrable to use this over `empty()`.
+ *
+ * ## Examples
```
benlesh: I find both of these examples confusing... but I do want to thank you for pointing out that we need good examples of when to use `EMPTY`. You can remove these examples, and we'll get the fix to the link merged. (Unless @JWO719 can come up with a better fix for the links)

nikagar4: Agreed but couldn’t think of something small and clear, at least this got your attention.

Tried to come up with clear real life examples but when I wrote it I realized I introduced lots of other operators which confuses people who actually need to read these

src/internal/observable/empty.ts
nikagar4: Looks like he’s not answering

nikagar4:
> LGTM, thanks for the fix!

Always a pleasure :)

PR #4973, opened 2019-08-20
Reduce scheduler footprint for default usage of shareReplay, timeInterval, and timestamp

benlesh:
Refactors `ReplaySubject`, `timeInterval`, and `timestamp` to remove reliance on Schedulers that were not necessary. This should help drastically reduce the amount of code pulled in for default `shareReplay` usage or `timestamp` usage.

See commit messages for more details.

NOTE: Breaking change in that `ReplaySubject` will no longer have inherent `observeOn` behavior if provided a scheduler.

### DIFF
--- a/src/internal/ReplaySubject.ts
+++ b/src/internal/ReplaySubject.ts
@@ -1,17 +1,39 @@
 import { Subject } from './Subject';
-import { SchedulerLike } from './types';
-import { queue } from './scheduler/queue';
+import { TimestampProvider } from './types';
 import { Subscriber } from './Subscriber';
 import { Subscription } from './Subscription';
-import { ObserveOnSubscriber } from './operators/observeOn';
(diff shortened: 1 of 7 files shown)


src/internal/ReplaySubject.ts
```diff
@@ -94,8 +117,9 @@ export class ReplaySubject<T> extends Subject<T> {
     return subscription;
   }
 
-  _getNow(): number {
-    return (this.scheduler || queue).now();
+  private _getNow(): number {
+    const { timestampProvider: scheduler } = this;
+    return scheduler ? scheduler.now() : Date.now();
```
cartant: Is there anything to be gained by keeping the scheduler terminology in here? I mean, it could be a scheduler, but it's entirely possible it could be something else. Why not just use `timestampProvider` as the name for the local const?

PR #4976, opened 2019-08-20
fix(of): Fix compilation failure in TS 3.5+ for `of()`

benlesh:
I've also heard, that great commit summaries contain fewer than 50 characters. Thanks, Github! <3

### DIFF
--- a/src/internal/observable/of.ts
+++ b/src/internal/observable/of.ts
@@ -29,6 +29,7 @@ export function of<T, T2, T3, T4, T5, T6, T7, T8, T9>(a: T, b: T2, c: T3, d: T4,
 export function of<T>(...args: (T | SchedulerLike)[]): Observable<T>;
 
 // TODO(benlesh): Update the typings for this when we can switch to TS 3.x
+export function of(): Observable<never>;
 export function of<T>(a: T): Observable<T>;
 export function of<T, T2>(a: T, b: T2): Observable<T | T2>;
 export function of<T, T2, T3>(a: T, b: T2, c: T3): Observable<T | T2 | T3>;


src/internal/observable/of.ts
```diff
@@ -29,6 +29,7 @@ export function of<T, T2, T3, T4, T5, T6, T7, T8, T9>(a: T, b: T2, c: T3, d: T4,
 export function of<T>(...args: (T | SchedulerLike)[]): Observable<T>;
 
 // TODO(benlesh): Update the typings for this when we can switch to TS 3.x
+export function of(): Observable<never>;
```
cartant: I think there will be a problem with the ordering of the signatures. The signature above this one - the signature that takes a `T | SchedulerLike` rest parameter - will match calls with no arguments, so the inferred return type will be, AFAICT, `Observable<unknown>`:

```ts
declare function of<T>(...args: (T | SchedulerLike)[]): Observable<T>;
declare function of(): Observable<never>;
declare function of<T>(...args: T[]): Observable<T>;
const r = of(); // Observable<unknown>
```

cartant: Looked at this again, this morning, and I think the rest-parameter signature with the `(T | SchedulerLike)[]` in it should be moved to just above the rest-parameter signature that just has `T[]` in it. This was [touched on](https://github.com/ReactiveX/rxjs/pull/4724#discussion_r282212650) in #4724 but the fix in that PR suggested moving the `(T | SchedulerLike)[]` signature all the way to the bottom.

In that PR, I thought it would be best to just remove the deprecation comment and not risk breaking anything. But in this PR, some reordering is needed.

PR #4977, opened 2019-08-22
feat(TestScheduler): expose `frameTimeFactor` property

benlesh:
- Makes `frameTimeFactor` public so people do not have to cast in TypeScript
- Deprecates `frameTimeFactor` on `VirtualTimeScheduler` as it is not used in that class, it
is really only intended to be used for `TestScheduler` marble diagrams
- Deprecates `index` on `VirtualTimeScheduler` because it is a leaked implementation detail
- Adds documentation


benlesh:
cc @kolodny

PR #4979, opened 2019-08-22
Add concatMap overload to include changes from Google

kolodny:
This PR includes downstream changes in Google that were needed to keep the build green for a vast amount of projects when updating to v7-alpha

### DIFF
--- a/spec-dtslint/operators/concatMap-spec.ts
+++ b/spec-dtslint/operators/concatMap-spec.ts
@@ -40,3 +40,7 @@ it('should enforce types', () => {
 it('should enforce the return type', () => {
   const o = of(1, 2, 3).pipe(concatMap(p => p)); // $ExpectError
 });
+
+it('should produce `Observable<never>` when concating an empty stream', () => {
+  const o = of(1, 2, 3).pipe(concatMap(n => of())); // $ExpectType Observable<never>
+});
--- a/spec-dtslint/operators/concatMap-spec.ts
+++ b/spec-dtslint/operators/concatMap-spec.ts
@@ -33,14 +33,14 @@ it('should support union-type projections', () => {
   const o = of(Math.random()).pipe(concatMap(n => n > 0.5 ? of('life') : of(42))); // $ExpectType Observable<string | number>
 });
 
+it('should support union-type projections with empty streams', () => {
+  const o: Observable<number> = of(1, 2, 3).pipe(concatMap(n => Math.random() < 0.5 ? of(123) : of()));
+});
+
 it('should enforce types', () => {
   const o = of(1, 2, 3).pipe(concatMap()); // $ExpectError
 });
 
 it('should enforce the return type', () => {
   const o = of(1, 2, 3).pipe(concatMap(p => p)); // $ExpectError
-});
-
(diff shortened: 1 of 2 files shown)


kolodny:
Done

cartant:
A similar signature should be added to the other flattening operators, too, IMO. I'll open a PR later.

kolodny:
@cartant You're right, the test I gave was a poor one. I created a test which more closely matches what exactly is breaking internally and why this change was needed. Note that this is only breaking in one place (albeit a whole bunch of targets rely on this breaking helper function) so I can fix it in this internally and there wouldn't be a need for the PR, but it does seem that this PR is introducing better types overall

spec-dtslint/operators/concatMap-spec.ts
```diff
@@ -33,6 +33,10 @@ it('should support union-type projections', () => {
   const o = of(Math.random()).pipe(concatMap(n => n > 0.5 ? of('life') : of(42))); // $ExpectType Observable<string | number>
 });
 
+it('should support union-type projections with empty streams', () => {
+  const o: Observable<number> = of(1, 2, 3).pipe(concatMap(n => Math.random() < 0.5 ? of(123) : of()));
```
cartant: This needs to be written as:

```ts
const o = of(1, 2, 3).pipe(concatMap(n => Math.random() < 0.5 ? of(123) : of())); // $ExpectType Observable<number>
```

dtslint doesn't care about or report compilation errors. An expectation is required, otherwise dtslint would pass this even if the inferred return type wasn't assignable to `Observable<number>`.

cartant:
I don't think we can make the change in this PR. Not to `concatMap` nor to any of the other flattening operators.

In #4983, I made the change to the other flattening operators and it broke a test - see the log in that PR. For whatever reason, the `mergeMap` operator in the

```ts
it('should not break unsubscription chains when unsubscribed explicitly', () => {
```

in `combineAll-spec.ts` infers matches the signature and infers `Observable<never>`. Why this is the case, IDK, but we cannot make this change without breaking people's code, IMO.

kolodny:
I'm not 100% clear on why that tests would start failing with changing `combineAll` TBH. Do note that this PR is not needed for our use case once https://github.com/ReactiveX/rxjs/pull/4976 lands since that ensures `of()` emits a `never` and `never`s disappear in unions. I'd recommend just waiting for that PR to land and close this as obsolete once that happens.

cartant:
IDK either. Specifically, it fails because the `mergeMap` on [this line](https://github.com/ReactiveX/rxjs/blob/b0fe2866989bff76431304ee71a8299d4cc341a6/spec/operators/combineAll-spec.ts#L174) infers `Observable<never>`. You should be able to switch that out - temporarily - with `concatMap` in your own PR to see the compilation failure.

The observables in that are merged/concatenated in that test are higher-order observables, but that should make no difference. It's entirely possible that this is a TypeScript bug, but if we cannot determine why the signature is matched - and the a compilation error is effected for valid code - I don't see how we can merge this change.

kolodny:
This PR isn't needed anymore and in fact will break the build if it's included. closing this PR

PR #4983, opened 2019-08-24
test: add never dtslint tests for flattening operators

cartant:
**Description:**

<s>This PR adds tests for flattening operators to ensure that `Observable<never>` is inferred whenever the inner source has the type `ObservableInput<never>`. These tests all pass without changes to the types being necessary, so I'm unsure that the change in #4979 is actually required.</s>

The signature changes in this PR have been reverted - the problems they fixed were addressed in a change in another PR (#4985). This PR now contains only the dtslint tests that were added along with the initial changes.

### DIFF
--- a/src/internal/operators/mergeMapTo.ts
+++ b/src/internal/operators/mergeMapTo.ts
@@ -3,6 +3,7 @@ import { mergeMap } from './mergeMap';
 import { ObservableInput } from '../types';
 
 /* tslint:disable:max-line-length */
+export function mergeMapTo(observable: ObservableInput<never>, concurrent?: number): OperatorFunction<any, never>;
 export function mergeMapTo<O extends ObservableInput<any>>(innerObservable: O, concurrent?: number): OperatorFunction<any, ObservedValueOf<O>>;
 /** @deprecated */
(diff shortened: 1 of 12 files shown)


src/internal/operators/mergeMapTo.ts
```diff
@@ -3,7 +3,7 @@ import { mergeMap } from './mergeMap';
 import { ObservableInput } from '../types';
 
 /* tslint:disable:max-line-length */
-export function mergeMapTo<O extends ObservableInput<any>>(innerObservable: O, concurrent?: number): OperatorFunction<any, ObservedValueOf<O>>;
+export function mergeMapTo<T, O extends ObservableInput<any>>(innerObservable: O, concurrent?: number): OperatorFunction<any, ObservedValueOf<O>>;
```
benlesh, line 6:
```suggestion
export function mergeMapTo<T, O extends ObservableInput<any>>(innerObservable: O, concurrent?: number): OperatorFunction<T, ObservedValueOf<O>>;
```

Should this not be `T` here also?

cartant: I thought I'd reverted the signature changes and included only the tests that were added. These changes to the signatures should not be there. The signatures they are replacing are correct. Lemme see if I can clean this PR up a bit.

PR #4984, opened 2019-08-25
refactor: remove observers for loop in subject class

AllanEd:
**Description:**


AllanEd:
I did not add a description, because it is only a refactoring of the subject class.

kwonoj:
I believe for loop in our code is micro optimization and we won't likely refactor for internal readability: 

This PR requires a performance benchmark / proof that this change wouldn't regress in hot path performance.

kwonoj:
I'm closing this for now as we didn't hear back around performance proof. Appreciate for taking effort for PR and sorry for not landing this.

PR #4985, opened 2019-08-27
Update to TypeScript 3.5.3

benlesh:
- Updates places that were inferring `Observable<{}>` to infer `Observable<unknown>` instead.
- Updates to proper types for TS 3.5.3 for certain dtslint tests
- Updates types for `of`.
  - BREAKING CHANGE: generic signature of `of` has changed. Users should allow this type to infer properly or use `as`.  There is deprecated handling for the common cases of `of<T>()` and `of<T>(x)`, but multiple specified generic arguments (like `of<A, B>(a, b)`) will yield incorrect types, possibly build errors, and should not be used.
  - BREAKING CHANGE: uses of `of` with >9 arguments punctuated with a `SchedulerLike` will no longer properly infer, and will need to be cast. Scheduled `of` is deprecated anyhow, and developers should be using `scheduled` from `rxjs`. (Before it would infer `Observable<{}>`, which is no longer possible, and now it will infer `Observable<A | B | C | D | E | F | G | H | I>`, which isn't correct. Again, use `scheduled` instead.


src/internal/types.ts
```diff
@@ -101,3 +101,5 @@ export interface SchedulerAction<T> extends Subscription {
 export type ObservedValueOf<O> = O extends ObservableInput<infer T> ? T : never;
 
 export type ObservedValuesFromArray<X> = X extends Array<ObservableInput<infer T>> ? T : never;
+
+export type ArrayValueOf<A> = A extends Array<infer T> ? T : never;
```
cartant: Given that the above type is named `ObservedValuesFromArray` and extracts the observed values from an array and this one extracts the value from an array, should it be called `ValueFromArray`?

src/internal/observable/concat.ts
```diff
@@ -139,5 +139,5 @@ export function concat<R>(...observables: (ObservableInput<any> | SchedulerLike)
  * @owner Observable
  */
 export function concat<O extends ObservableInput<any>, R>(...observables: Array<O | SchedulerLike>): Observable<ObservedValueOf<O> | R> {
```
cartant: Do we really need this `R` here? It seems that it's only present so that folk can explicitly provide it as a type parameter and do we really want to encourage that?

spec-dtslint/operators/scan-spec.ts
```diff
@@ -31,9 +31,8 @@ it('should accept seed parameter of a different type', () => {
 });
 
 it('should act appropriately with no seed', () => {
-  // Because an observable of one value will just pass that value directly through the reducer,
-  // the below could be a number or a string.
-  const a = of(1, 2, 3).pipe(scan((a: any, v) => '' + v)); // $ExpectType Observable<string | number>
+  // Starting in TS 3.5, the return type is inferred from the accumulator's type if it's provided without a seed.
+  const a = of(1, 2, 3).pipe(scan((a: any, v) => '' + v)); // $ExpectType Observable<any>
```
cartant: Maybe type the accumulator as a `string` and expect `Observable<string>`? Is there anything to be gained by using `any` here?

spec-dtslint/operators/reduce-spec.ts
```diff
@@ -31,9 +31,8 @@ it('should accept seed parameter of a different type', () => {
 });
 
 it('should act appropriately with no seed', () => {
-  // Because an observable of one value will just pass that value directly through the reducer,
-  // the below could be a number or a string.
-  const a = of(1, 2, 3).pipe(reduce((a: any, v) => '' + v)); // $ExpectType Observable<string | number>
+  // Starting in TS 3.5, the return type is inferred from the accumulator's type if it's provided without a seed.
+  const a = of(1, 2, 3).pipe(reduce((a: any, v) => '' + v)); // $ExpectType Observable<any>
```
cartant: As for `scan`, maybe use `string` as the accumulator type?

spec-dtslint/operators/scan-spec.ts
cartant: I see. Do you think it would be worth including another test that explicitly types the `value` - to make sure that works, too? I assume that `value` will also be `any` here - for the reason you gave above? FWIW, in application code I'd type it like this:

```ts
(scan((a: string, v: number) => '' + v))
```

So maybe we need a test to make sure this is okay?

spec-dtslint/observables/concat-spec.ts
```diff
@@ -28,8 +28,8 @@ it('should accept more than 6 params', () => {
   const o = concat(of(1), of(2), of(3), of(4), of(5), of(6), of(7), of(8), of(9)); // $ExpectType Observable<number>
 });
 
-it('should return Observable<{}> for more than 6 different types of params', () => {
-  const o = concat(of(1), of('a'), of(2), of(true), of(3), of([1, 2, 3]), of(4)); // $ExpectType Observable<{}>
+it('should return Observable<unknown> for more than 6 different types of params', () => {
```
cartant: This description no longer matches the test's expectation.

src/internal/observable/of.ts
```diff
@@ -1,47 +1,42 @@
-import { SchedulerLike } from '../types';
+import { SchedulerLike, ValueFromArray } from '../types';
 import { isScheduler } from '../util/isScheduler';
 import { fromArray } from './fromArray';
 import { Observable } from '../Observable';
 import { scheduleArray } from '../scheduled/scheduleArray';
+import { never } from './never';
```
cartant: This isn't needed. It looks like VS Code's auto-import has added this.

PR #4986, opened 2019-08-27
Using Boolean as predicates.

benlesh:
See commits.

In particular, note the comments in the addition to the `takeWhile` dtslint tests.

Blocked by #4985


src/internal/operators/first.ts
```diff
@@ -8,18 +8,10 @@ import { throwIfEmpty } from './throwIfEmpty';
...
-  predicate: (value: T, index: number, source: Observable<T>) => value is S,
-  defaultValue?: S
-): OperatorFunction<T, S>;
-export function first<T, D = T>(
-  predicate: (value: T, index: number, source: Observable<T>) => boolean,
-  defaultValue?: D
-): OperatorFunction<T, T | D>;
+export function first<T, D = T>(predicate: BooleanConstructor, defaultValue?: D): OperatorFunction<T, T>;
+export function first<T, D = T>(predicate?: null, defaultValue?: D): OperatorFunction<T, T | D>;
+export function first<T, S extends T>(predicate: (value: T, index: number, source: Observable<T>) => value is S, defaultValue?: S): OperatorFunction<T, S>;
```
cartant: Does the default value have to be an `S`? Why can't it be a `D` with the return type being `OperatorFunction<T, S | D>`?

src/internal/operators/last.ts
```diff
@@ -8,18 +8,10 @@ import { defaultIfEmpty } from './defaultIfEmpty';
...
-  predicate: (value: T, index: number, source: Observable<T>) => value is S,
-  defaultValue?: S
-): OperatorFunction<T, S>;
-export function last<T, D = T>(
-  predicate: (value: T, index: number, source: Observable<T>) => boolean,
-  defaultValue?: D
-): OperatorFunction<T, T | D>;
+export function last<T, D = T>(predicate: BooleanConstructor, defaultValue?: D): OperatorFunction<T, T | D>;
+export function last<T, D = T>(predicate?: null, defaultValue?: D): OperatorFunction<T, T | D>;
+export function last<T, S extends T>(predicate: (value: T, index: number, source: Observable<T>) => value is S, defaultValue?: S): OperatorFunction<T, S>;
```
cartant: Does the default value have to be an `S`? Why can't it be a `D` with the return type being `OperatorFunction<T, S | D>`?

src/internal/operators/takeWhile.ts
```diff
@@ -3,10 +3,10 @@ import { Observable } from '../Observable';
 import { Subscriber } from '../Subscriber';
 import { OperatorFunction, MonoTypeOperatorFunction, TeardownLogic } from '../types';
 
-export function takeWhile<T, S extends T>(predicate: (value: T, index: number) => value is S): OperatorFunction<T, S>;
-export function takeWhile<T, S extends T>(predicate: (value: T, index: number) => value is S, inclusive: false): OperatorFunction<T, S>;
-export function takeWhile<T>(predicate: (value: T, index: number) => boolean, inclusive?: boolean): MonoTypeOperatorFunction<T>;
-
+export function takeWhile<T>(predicate: BooleanConstructor, inclusive?: boolean): OperatorFunction<T, T>;
+export function takeWhile<T, S extends T>(predicate: (value: T, index: number) => value is S, inclusive: true): OperatorFunction<T, T>;
+export function takeWhile<T, S extends T>(predicate: (value: T, index: number) => value is S, inclusive?: boolean): OperatorFunction<T, S>;
```
cartant: I think this is the wrong way around. If a boolean variable is passed - instead of a literal - it will match the second of the highlighted signatures and the return type will be `OperatorFunction<T, S>`, but if the variable is `true` at runtime, a non-`S` type could be emitted.

I think that a user-defined type guard can be used as the predicate **only** when `inclusive` is known to be `false` at compile time. So the signature with the user-defined type guard should be:

```
export function takeWhile<T, S extends T>(predicate: (value: T, index: number) => value is S, inclusive: false): OperatorFunction<T, S>;
```

And the other should use a garden-variety predicate and not a user-defined type guard.

benlesh:
This is no longer an issue with TS 3.8.

Issue #4630, opened 2019-03-13
[docs]: Replace JQuery to be library agnostic


BioPhoton:
### Please check those that apply

- [ ] typo
- [ ] documentation doesn't exist
- [x] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

### Description Of The Issue

In the docs of the bindCallback is an example for JQuery. 
https://github.com/ReactiveX/rxjs/blob/master/src/internal/observable/bindCallback.ts#L113

We should replace it with a more modern lib. i.e. axios

BioPhoton:
Ok, I realized that the example is for bindCallback axios will not work. 
Most of the modern technologies are at least promise based. 

I'm happy to implement any suggestion. Just don't know what I should aim for.

benlesh:
This issue is no longer relevant.

PR #5010, opened 2019-09-12
docs(bindCallback): remove jQuery example #4630

jabiinfante:
**Description:** I think using _navigator.geolocation.getCurrentPosition_ callback with **bindCallback**, is a 
simple sample, since no external library is needed.

**Related issue (if exists):** #4630

### DIFF
--- a/src/internal/observable/bindCallback.ts
+++ b/src/internal/observable/bindCallback.ts
@@ -110,15 +110,20 @@ export function bindCallback(callbackFunc: Function, scheduler?: SchedulerLike):
  *
  * ## Examples
  *
- * ### Convert jQuery's getJSON to an Observable API
+ * ### Convert navigator.geolocation.getCurrentPosition to an Observable API
  * ```ts
  * import { bindCallback } from 'rxjs';
- * import * as jQuery from 'jquery';
  *
- * // Suppose we have jQuery.getJSON('/my/url', callback)
- * const getJSONAsObservable = bindCallback(jQuery.getJSON);
- * const result = getJSONAsObservable('/my/url');
- * result.subscribe(x => console.log(x), e => console.error(e));
+ * // Assuming navigator.geolocation is available on the browser.
+ * const getCurrentPosition = bindCallback(navigator.geolocation.getCurrentPosition);
+ *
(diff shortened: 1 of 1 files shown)


src/internal/observable/bindCallback.ts
```diff
@@ -110,15 +110,20 @@ export function bindCallback(callbackFunc: Function, scheduler?: SchedulerLike):
...
  * ```ts
  * import { bindCallback } from 'rxjs';
- * import * as jQuery from 'jquery';
  *
- * // Suppose we have jQuery.getJSON('/my/url', callback)
- * const getJSONAsObservable = bindCallback(jQuery.getJSON);
- * const result = getJSONAsObservable('/my/url');
- * result.subscribe(x => console.log(x), e => console.error(e));
+ * // Assuming navigator.geolocation is available on the browser.
+ * const getCurrentPosition = bindCallback(navigator.geolocation.getCurrentPosition);
```
cartant: This doesn't really work as an example, as `getCurrentPosition` takes the position callback as the first parameter and `bindCallback` wants the caller to:

> Give it a function f of type f(x, callback) and it will return a function g that when called as g(x) will output an Observable.

Also, `getCurrentPosition` also accepts an error callback.

jabiinfante: Thanks for your feedback... I see your point.

I guess my propossed example works just because _navigator.geolocation.getCurrentPosition_ luckyly works with no parameters. am I right?

Also _navigator.geolocation.getCurrentPosition_ might be continuosly emitting values... so certanly won't match bindCallback purpouse...

So, sorry but i can't think of any other "native" callback which could replace the jquery example :confused:

cartant: I can't think of one either. Outside of the DOM there is Node's [`fs.exists`](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback), but it is deprecated (and rightly so). I'm happy for this to stay as a jQuery example.

Even if it 'works', there are other issues with `getCurrentPosition`. One of those is that any example would have to make some sense with the TypeScript typing of the bound function and `getCurrentPosition` doesn't make sense, as the callback is not the last parameter.

PR #5016, opened 2019-09-14
Fix not cleared storage task handlers from setImmediate

mazulatas:
**Description:**
This PR adds cleaning up handle tasks in Immediate


stylesam:
@cartant Why did the build fall?

cartant:
> Why did the build fall?

It's nothing you've done The version of `@types/node` is too old for TypeScript 3.6.3. That can be dealt with elsewhere.

I'll review your PR a little later. It looks good; there's just one or two things I'd like changed. Thanks.

stylesam:
@cartant Thanks for review, we are waiting you

src/internal/util/Immediate.ts
```diff
@@ -21,3 +22,9 @@ export const Immediate = {
     delete tasksByHandle[handle];
   },
 };
+
+export const ForTests = {
```
cartant: I'd suggest renaming `ForTests` to `TestTools` and renaming `numberOfTasksToHandle` to `pending`.

mazulatas: I agree

benlesh:
Thank you @mazulatas for catching this one and bringing it to my attention. I've further refactored the code to make it a little more efficient than it was, and I've kept your test and test util.

(I did the code editing in the browser, so we'll see if CI passes, haha)

PR #5020, opened 2019-09-16
docs(pipe) Update piping example

spyros-uk:
Updates example operation number


cartant:
I only just noticed that this change was addressed in another PR that I've just merged. Sorry, but I'm going to have to close this. Thank you for putting in the effort to open this PR, though. It's appreciated.

Issue #5002, opened 2019-09-06
New Creation Operator for Mutation Observer


niklas-wortmann:
I think it would be very valuable to have a creation operator for an Observable of a Mutation Observer. It used to be in rxjs-dom already but wasn't ported. I think it allows really cool and reactive ui architectures. 
Rough idea: The observable is is for example triggered as soon as there is a child element added to the dom. With the trigger within the observable one could initiate a loading indicator or do some optimistic ui update, while fetching some data from an HTTP server or whatever. Thinks like endless scrolling would be amazing to implement with such an approach. 

From an api perspective, I'm unsure whether it would be better to have a creation operator that accepts a mutation observer instance or just the needed parameter.

What do you think of that idea?

A-Babin:
[MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) shares a mostly similar interface with [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver) and [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver). The latter two are currently at the working and editors draft stage respectively and don't have full browser support yet, but it may be worthwhile to have an creation operator that works with all three. I've already used my own Mutation Observable to great effect in Chrome extensions, but I expect that Intersection and Resize will end up seeing more general use. 

They all (unfortunately) require the handler on construction so I don't know how creating with instances would work in any case. Perhaps something like this:

```DOMObservables.ts
type DOMObservableType = "mutation" | "resize" | "intersection";

type DOMObservableOptions = {
  mutation : MutationObserverInit,
  resize : {box ?: "content-box" | "border-box"},
  intersection: {
    root ?: Element,
    rootMargin ?: string,
    threshold ?: number
  }
}

type DOMObservableEntry = {
  mutation : MutationRecord[],
  resize : ResizeObserverEntry[] // Not in DOM type library yet
  intersection : IntersectionObserverEntry[]
}

function fromDOM<T extends DOMObservableType>(
    type : T, 
    target : Element, // Iterable<Element> ? 
    options ?: DOMObservableOptions[T]
  ) : Observable<DOMObservableEntry[T]>{ // Maybe pass the observer instance too like the native callbacks do
  //...
}
```

[...]

PR #5021, opened 2019-09-16
feat(animationFrames): Adds an observable of animationFrames

benlesh:
- Also adds tests and test harness for requestAnimationFrame stubbing with sinon
- Updates TypeScript lib to use ES2018 (so we can use `findIndex` on `Array`).

## This is a proposed API, we haven't decided on anything yet

Example 1: Just a simple use.
```ts
import { animationFrames } from 'rxjs';

// Log the elapsed milliseconds at each animation frame.
animationFrames().subscribe(elapsed => console.log(elapsed));
```

Example 2: Tweening animation function
```ts
 import { animationFrames } from 'rxjs';
 import { map, takeWhile, endWith } from 'rxjs/operators';
 
 function tween(start: number, end: number, duration: number) {
   const diff = end - start;
   return animationFrames().pipe(
     // Figure out what percentage of time has passed
     map(elapsed => elapsed / duration),
     // Take the vector while less than 100%
     takeWhile(v => v < 1),
     // Finish with 100%
     endWith(1),
     // Calculate the distance traveled between start and end
     map(v => v * diff + start)
   );
 }
 
 // Setup a div for us to move around
 const div = document.createElement('div');
 document.body.appendChild(div);
 div.style.position = 'absolute';
 div.style.width = '40px';
 div.style.height = '40px';
 div.style.backgroundColor = 'lime';
 div.style.transform = 'translate3d(10px, 0, 0)';
 
 tween(10, 200, 4000).subscribe(x => {
   div.style.transform = `translate3d(${x}px, 0, 0)`;
 });
```

See documentation in code for more information.

### Highlights:

- Starts a single animation loop for N subscriptions.
- If no more subscribers are subscribed to the animation loop, it will terminate the animation loop.
- Replacement for odd patterns like `range(0, animationFrameScheduler)` or `interval(0, animationFrameScheduler)`
- Outputs more useful values for animations than above recipes.

### TBD:

- Do we want this at all?
- Do we want to export this from `rxjs`, or do we add a new import site for `rxjs/animation`?  If the latter, does it need to be `dom-animation`?
- Recommendations for testing usage of this to be included in docs? 
- Do we find this approach agreeable? 
  - Is `Observable<void>` preferrable, and make users compose to get the elapsed time? 
  - Do we keep it as-is?
  - Do we want each subscription to this to start a different animation loop? (Seems inefficient)

### DIFF
--- a/src/internal/observable/dom/animationFrames.ts
+++ b/src/internal/observable/dom/animationFrames.ts
@@ -1,5 +1,4 @@
 import { Observable } from '../../Observable';
-import { Subscriber } from '../../Subscriber';
 
 // TODO: move to types.ts
 export interface TimestampProvider {
@@ -7,37 +6,17 @@ export interface TimestampProvider {
 }
 
 /**
- * A list of subscribers to notify, the timestamp providers used to get values, and the start times
- * from their subscriptions.
- *
- * The structure is as follows:
- *
- * `i`: `Subscriber<number>` - The subscriber to notify.
- * `i + 1`: `timestampProvider` - the object used to get each new timestamp
- * `i + 2`: `number` - the start timestamp
- */
-const subscriberState: (Subscriber<number> | TimestampProvider | number)[] = [];
-
-/**
- * A list of subscribers to remove at the end of the run loop.
- *
- * This exists because during the run loop, when `next` is called on each subscriber,
- * a side effect could occur that would cause unsubscription. In order to prevent us
- * from looping over a currently mutating array of subscribers, we buffer the subscribers
- * we want to remove until after the run loop is done, then remove them.
- */
-const subscribersToRemoveAfterRun: Subscriber<number>[] = [];
-
-/**
  * An observable of animation frames
  *
  * Emits the the amount of time elapsed since subscription on each animation frame. Defaults to elapsed
  * milliseconds. Does not end on its own.
  *
- * Will schedule a shared animation frame loop and notify all subscribers from that one loop. As
- * an implementation detail, this means that it does not necessary call `requestAnimationFrame` at
- * the time of subscription, rather it shares a single `requestAnimationFrame` loop between many
- * subscriptions.
+ * Every subscription will start a separate animation loop. Since animation frames are always scheduled
+ * by the browser to occur directly before a repaint, scheduling more than one animation frame synchronously
+ * should not be much different or have more overhead than looping over an array of events during
+ * a single animation frame. However, if for some reason the developer would like to ensure the
+ * execution of animation-related handlers are all executed during the same task by the engine,
+ * the `share` operator can be used.
  *
  * This is useful for setting up animations with RxJS.
  *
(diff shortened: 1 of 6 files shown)


src/internal/observable/dom/animationFrames.ts
```diff
@@ -0,0 +1,214 @@
...
+  if (scheduledAnimationId === 0) {
+    scheduledAnimationId = requestAnimationFrame(animate);
+  }
+}
+
+/**
+ * Executes notification of all subscribers, then reschedules itself.
+ * Do not call directly. This is the "run loop".
+ */
+function animate() {
```
cartant: This is variously referred to as the `run` loop and the `animate` loop. Maybe pick one term for it and use that term only?

JWO719:
I think this is a good idea to add to the core library. Anyway, I'd vote for an import site `rxjs/dom`, because this would fit quite well with #5002 & we used to have rxjs-dom, but didn't migrate it.

basham:
Somewhat related to this conversation, I've published a package ([`conduit-rxjs-react`](https://github.com/indiana-university/conduit/tree/master/packages/conduit-rxjs-react)) that connects RxJS observables to React components. As an optimization, I limit the number of state updates using `requestAnimationFrame` to reduce thrashing. If there are multiple state updates while waiting for the animation frame callback, only the most recent update is kept. The following is an abbreviated version of the code to demonstrate this implementation. Maybe this could be rewritten slightly to take advantage of an `animationFrames` observable.

```js
import { Component } from 'react'
import { bindCallback } from 'rxjs'
import { audit } from 'rxjs/operators'

class StreamComponent extends Component {
  constructor () {
    state$.pipe(
      audit(bindCallback((x, callback) =>
        window.requestAnimationFrame(callback)
      ))
    ).subscribe((state) => {
      this.setState(state)
    })
  }
}
```

benlesh:
> Presumably, this is done to avoid lots of requestAnimationFrame calls per frame when there are lots of subscribers?

@cartant That's exactly right. There's no reason to schedule multiple animation frames, when sharing one is more efficient, at least in the common case. IMO, it's also surpising and confusing for developers that they can call `requestAnimationFrame(updateBox)` and `requestAnimationFrame(updateBall)`, and the box will be updated at a separate time than the ball. I guess the method isn't called `scheduleForNextAnimationFrame(fn)`, but I know I found it surprising the callbacks didn't land at the same time.

benlesh:
Actually, I'm on the fence about whether or not to have each one start it's own animation loop.

If you call `requestAnimationFrame` 3 times in a row, synchronously, it will indeed fire 3 separate tasks, but they also all occur _directly_ before the repaint. So having them all in loop probably isn't as big of a win as I think it is.

I think I'm going to make each one start it's own loop, and if users want to share that loop they can compose that behavior with `share`.

benlesh:
Okay, the change is made, and I like this better. It now creates a different animation loop per subscriber, and if a developer wants to share that animation loop, they just use `share`, although i'm not at all sure anymore that it's a big win.

cartant:
Regarding the import location, I'd vote for keeping it as-is - `import { animationFrames } from "rxjs"` - for now, as that's import location for `animationFrameScheduler`.

Moving to an `"rxjs/dom"` import location could be something that we could look at later - as part of a restructuring. Doing so would allow us to do some tricky things with the TypeScript's `dom` library typings so that `fromEvent` could be 'smart' and correctly infer the event object that's associated with mouse events, etc.

jayphelps:
> It now creates a different animation loop per subscriber, and if a developer wants to share that animation loop, they just use share, although i'm not at all sure anymore that it's a big win.

FWIW I came here to request this

PR #5024, opened 2019-09-19
fix(filter): Fix overload order for filter to support inferring the generic type

kolodny:
The provided test will create an `Observable<never>` with the order unchanged.

### DIFF
--- a/spec-dtslint/operators/filter-spec.ts
+++ b/spec-dtslint/operators/filter-spec.ts
@@ -65,8 +65,5 @@ it('should support inference from a generic return type of the predicate', () =>
     };
   }
 
-  // $ExpectType Observable<string>
-  const source = of('World', 'test', null, 'ok').pipe(
-    filter(isDefined()),
-  );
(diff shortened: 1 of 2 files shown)


spec-dtslint/operators/filter-spec.ts
```diff
@@ -56,4 +56,17 @@ it('should support inference from a return type with Boolean as a predicate', ()
...
+
+it('should support inference from a generic return type of the predicate', () => {
+  function isDefined<T>() {
+    return (value: T|undefined|null): value is T => {
+      return value !== undefined && value !== null;
+    };
+  }
+
+  // $ExpectType Observable<string>
+  const source = of('World', 'test', null, 'ok').pipe(
```
benlesh: Does this work with more than one type?

```
const o$ = of(1, true, 'Nick Cage').pipe(filter(isDefined())) // $ExpectType Observable<number | boolean | string>
```

kolodny: Yup, updated the test to the following to reflect that:

```ts
  const o$ = of(1, null, {foo: 'bar'}, true, undefined, 'Nick Cage').pipe(filter(isDefined())); // $ExpectType Observable<string | number | boolean | { foo: string; }>

```

PR #5027, opened 2019-09-20
Update of.ts

kolodny:
Remove deprecation warning for valid use of `of`

### DIFF
--- a/src/internal/observable/of.ts
+++ b/src/internal/observable/of.ts
@@ -34,7 +34,6 @@ export function of<T, T2, T3, T4, T5, T6, T7, T8, T9>(a: T, b: T2, c: T3, d: T4,
 export function of(): Observable<never>;
 /** @deprecated remove in v8. Do not use generic arguments directly, allow inference or cast with `as` */
 export function of<T>(): Observable<T>;
-/** @deprecated remove in v8. Do not use generic arguments directly, allow inference or cast with `as` */
 export function of<T>(value: T): Observable<T>;
 export function of<A extends Array<any>>(...args: A): Observable<ValueFromArray<A>>;
 /* tslint:enable:max-line-length */
--- a/src/internal/observable/of.ts
+++ b/src/internal/observable/of.ts
@@ -32,10 +32,11 @@ export function of<T, T2, T3, T4, T5, T6, T7, T8, T9>(a: T, b: T2, c: T3, d: T4,
 
 // TODO(benlesh): Update the typings for this when we can switch to TS 3.x
 export function of(): Observable<never>;
+export function of<A extends Array<any>>(...args: A): Observable<ValueFromArray<A>>;
 /** @deprecated remove in v8. Do not use generic arguments directly, allow inference or cast with `as` */
 export function of<T>(): Observable<T>;
+/** @deprecated remove in v8. Do not use generic arguments directly, allow inference or cast with `as` */
 export function of<T>(value: T): Observable<T>;
-export function of<A extends Array<any>>(...args: A): Observable<ValueFromArray<A>>;
 /* tslint:enable:max-line-length */
 
 /**
--- a/src/internal/observable/of.ts
+++ b/src/internal/observable/of.ts
@@ -32,11 +32,12 @@ export function of<T, T2, T3, T4, T5, T6, T7, T8, T9>(a: T, b: T2, c: T3, d: T4,
 
 // TODO(benlesh): Update the typings for this when we can switch to TS 3.x
 export function of(): Observable<never>;
-export function of<A extends Array<any>>(...args: A): Observable<ValueFromArray<A>>;
 /** @deprecated remove in v8. Do not use generic arguments directly, allow inference or cast with `as` */
 export function of<T>(): Observable<T>;
-/** @deprecated remove in v8. Do not use generic arguments directly, allow inference or cast with `as` */
 export function of<T>(value: T): Observable<T>;
+export function of<T, U>(value1: T, value2: U): Observable<T|U>;
+export function of<T, U, V>(value1: T, value2: U, value3: V): Observable<T|U|V>;
+export function of<A extends Array<any>>(...args: A): Observable<ValueFromArray<A>>;
 /* tslint:enable:max-line-length */
 
 /**


kolodny:
Actually, the more I think about it, we probably just need to switch the order of the overloads but keep the deprecation in for the `of<T>(t: T)` case. That way folks doing `of(new Foo())` won't get a deprecation warning but doing `of<Foo|null>(null)` would warn, which is what we really want.

cartant:
> ...switch the order of the overloads but keep the deprecation...

Yeah. I agree. Might as well move both of the deprecated signatures so that they follow the rest-parameter signature. They'll only be matched if an explicit type argument is specified.

kolodny:
Done, I wish there was a way to add a test for this. The closest I can do is mock the state here and link to a [**playground where it works**](https://www.typescriptlang.org/play/#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXx0QAoBKALngHIB5AaUoG4BYAKFElgRXWzwMQA8AQXggAHhhCpgAZ3hCYMKAE8BUVMoB8mogDp9sAOYyKQ8lTqNW7aHCRpMufIQEAVHecqgADnDBRJYCs2cFsuB15nQXciADcoCGQQCldPHz8AkCDmFlZCUngAXmKLehz8gEYSIpKaMrziCoAaKkkZDEpq4sLS4JdUZABbACMQGA8anq8QX3BM7IaBAZGxnSrJqnS5wOCGgu7NmYydhngAejP4GXUsDGV4MAALcABrFoxHrDkILFQEL-gUHgd28CDGMBwMF0rCAA)

```ts
declare function of(): 'OK';
declare function of<A extends Array<any>>(...args: A): 'OK';
declare function of<T>(): 'deprecated';
declare function of<T>(value: T): 'deprecated';

of() === 'OK';
of(1) === 'OK';
of(1, 'test') === 'OK';
of<number>() === 'deprecated';
of<number>(1) === 'deprecated';

of() === 'deprecated'; // sanity check, this line is a type error.
```

kolodny:
> ...Although, could that break anyone doing `of<string[]>(["foo", "bar"])` as it will match the rest-parameter signature? Does that matter?

`of<string[]>(["foo", "bar"]) === 'deprecated'` in the example above which is what we're after, they would need to do `of<string[][]>(["foo", "bar"]) === 'OK'`. I'm good with that tradeoff

cartant:
> ... I wish there was a way to add a test for this ...

Yeah. I'm inclined to look into adding another assertion to dtslint. I guess it should be possible to add `$ExpectDeprecation`. If it's not too hard, I think we should look into it before the v7 release.

kolodny:
This PR may not be ready due to https://github.com/microsoft/TypeScript/issues/33597

We may need to keep the original order and just remove the deprecation warning for `of<T>(t: T)`

cartant:
@kolodny @benlesh Looking at this again, I'm not sure why we even need `ValueFromArray`. In `of`, it's always going to be an array and the `ValueFromArray` type is not used elsewhere. I think the signature should be:

```ts
export function of<T>(...args: T[]): Observable<T>;
```

And all of the other signatures can be re-ordered and deprecated.

Regarding https://github.com/microsoft/TypeScript/issues/33597, if this signature is used the behaviour won't be a problem with `of` and AFAICT it won't be a problem elsewhere as the widening only happens with literal types (and elsewhere we are are dealing with observable inputs, etc. and not literal types).

kolodny:
I believe this PR is ready to go. This will be a slight breaking change to folks doing

<a href="https://www.typescriptlang.org/play/?ssl=15&ssc=125&pln=11&pc=1#code/C4TwDgpgBAaghgGwK4QGICcD2BbAgu9OEAHgBUoIAPYCAOwBMBnKOWkAbQF0A+KAXijkqNBs3yESAS1oAzCOigBVXgH4lUAFxRaEAG7yAUAYDGCOI2YB5AEaN5uuNYQQyvAN4GoXqA+QQAhFqkBgC+RvQQpnDo0DJItMbAkpi0UJgyxLgU1HRMUOJExKwg3NwAFJ5QAHQ10QDmjFq4BgCUWjZ26A5OLvB+GDgFJLilANxG0jToMnDG0ACyILiJyakeXqCQWgDky0kpAIzbUAA+ULsrKQBM26FGximMwGnWACTttvaOzsSLe6u8ATpMpuTYQHb-Q7bEItUZQAD08ME4Gg2ye6GkdWOkmYtEwz3MjEkdVo32gwEwUDB5wARJDaAcaacoHTLrQrjTtlUgA">

```
interface MyAction {
  type: 'Action1' | 'Action2'
}

// Error: Type 'string' is not assignable to type '"Action1" | "Action2"'.
const ob$: Observable<MyAction> = of({type: 'Action1'});
```

<a/>

But other than that, there shouldn't be much impact. This issue only affected 30 out of ~100k targets within Google

cartant:
> I believe this PR is ready to go.

Given that TS doesn't behave particularly nicely with rest parameters and literals, I'm not convinced that we should strive to remove the overload signatures for `of`. Particularly, as there is no scheduler-related shenanigans here. I think a handful of non-deprecated, fixed-argument overloads followed by a rest-parameter overload to catch everything else would be a better way to go.

E.g.

```ts
export function of(): Observable<never>;
export function of<T>(): Observable<T>;
export function of<T>(value: T): Observable<T>;
export function of<T1, T2>(value1: T1, value2: T2): Observable<T1 | T2>;
// maybe up to T3?
export function of<A extends Array<any>>(...args: A): Observable<ValueFromArray<A>>;
```

This seems a small price to pay to get it to work properly in 95% of case - and there are only a few RxJS functions that take values like this. IMO, the TS behaviour is not going to change anytime soon.

@benlesh

kolodny:
I'm OK with that, however I'd suggest we get creative with overloads to make the deprecated usage of `of<number>(1)` properly warn the user by doing something like this:

```ts
declare function of<_T, T extends _T>(t: T): 'OK';
declare function of<T>(t: T): 'deprecated';

const ok = of(1);
ok; // 'OK'

const deprecated = of<number>(1);
deprecated; // 'deprecated'
```

Thoughts?

cartant:
> Thoughts?

I'd prefer to leave it to a lint rule. Shenanigans with type parameters that don't directly relate to actual parameters makes me a little nervous.

kolodny:
Done, we should probably have a story around moving folks away from `of<number>(1)` though, but that's out of scope for this PR

cartant:
> ... we should probably have a story around moving folks away from `of<number>(1)`

Sure. I have an item in my TODO list - from a discussion with @JWO719 - about creating an issue to deal with brining the RxJS linting rules into this repo. I'll make a note that another issue - relating to recommended rules, etc. - should reference this issue.

PR #5031, opened 2019-09-23
use more exact return type in Notification#do

jokester:
**Description:**

This prevents a return type of `any` when the type can be inferred from callback types.

```

// Before: return type is always `any`

// ============

// After:

// when all 3 callbacks are specified: return type becomes (union of their return types)
const a = notification.do(
  _ => 'str',
  _ => 123,
  _ => ({a: 1})
);  // => a: string | number | {a: number}

// when not all of callbacks are specified:
// return type becomes (undefined | union of specified return types)
const b = notification.do(
  _ => 'str'
); // => b: string | undefined

```

### DIFF
--- a/src/internal/Notification.ts
+++ b/src/internal/Notification.ts
@@ -53,13 +53,13 @@ export class Notification<T> {
 
   /**
    * Given some {@link Observer} callbacks, deliver the value represented by the
-   * current Notification to the correctly corresponding callback.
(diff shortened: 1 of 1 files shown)


benlesh:
Closing this in favor of #5478

Thank you for your effort, @jokester ... it's helpful and it drew our attention to the issue.

PR #5038, opened 2019-09-24
test(take operator): use TestScheduler for tests

rall:
This switches the `take` operator tests over to use the `TestScheduler#run` method.

It looks like the subscription marbles syntax is changed for `TestScheduler` - ticks are denoted by a `-` and not a space - so those expectations needed to be updated too. Other than that the content of the tests should be the same as before.

### DIFF
--- a/spec/operators/take-spec.ts
+++ b/spec/operators/take-spec.ts
@@ -16,8 +16,8 @@ describe('take operator', () => {
 
   asDiagram('take(2)')('should take two values of an observable with many values', () => {
     testScheduler.run(({ cold, expectObservable, expectSubscriptions }) => {
-      const e1 =  cold('--a-----b----c---d--|');
-      const e1subs =   '^-------!------------';
+      const e1 = cold(' --a-----b----c---d--|');
+      const e1subs = '  ^-------!------------';
       const expected = '--a-----(b|)         ';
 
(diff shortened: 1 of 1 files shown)


spec/operators/take-spec.ts
```diff
@@ -1,110 +1,139 @@
...
+  });
+
   asDiagram('take(2)')('should take two values of an observable with many values', () => {
-    const e1 =  cold('--a-----b----c---d--|');
-    const e1subs =   '^       !            ';
-    const expected = '--a-----(b|)         ';
+    testScheduler.run(({ cold, expectObservable, expectSubscriptions }) => {
+      const e1 =  cold('--a-----b----c---d--|');
+      const e1subs =   '^-------!------------';
+      const expected = '--a-----(b|)         ';
```
cartant: We'd like to be able to use Prettier with the codebase and that means the alignment will need to happen within the marble strings, like this:

```
const e1 = cold('  --a-----b----c---d--|');
const e1subs = '   ^-------!------------';
const expected = ' --a-----(b|)         ';
```

As Prettier will collapse multiple adjacent spaces outside of the string literals.

rall: ok! I'll do that right now

Issue #5034, opened 2019-09-24
shareReplay memory leak


gogakoreli:
## Bug Report

**Current Behavior**
After component gets destroyed, its reference stays alive, because the fact that I subscribe to bigger scoped Observable (for example route.params) and switch it (via switchMap) to the Observable piped with shareReplay operator.

After taking heap snapshots via chrome devtools, I noticed that reference retainers mentioned shareReplay operator. If I use publishReplay and refCount instead of shareReplay problem is fixed, that's why I think shareReplay has an issue.

Please have a look at code on GitHub and screen shots for better understanding the issue. (Code explains issue better than description). You can see code sample below.

Event after reviewing #3336, issue still remains with shareReplay.

**Reproduction**
- Repo link: Please have a look at [GitHub repo](https://github.com/gogakoreli/rxjs-share-replay-leak)
- StackBlitz: Same repo source code imported from GitHub [Here](https://stackblitz.com/github/gogakoreli/rxjs-share-replay-leak)

```js
export class PersonComponent implements OnInit, OnDestroy {

  public person: any;

  private unsubscribe$ = new Subject();

  constructor(private api: ApiService, private route: ActivatedRoute) { }

  ngOnInit() {
    this.route.params.pipe(
      switchMap(_ => this.api.find()),
      takeUntil(this.unsubscribe$),
    ).subscribe({
      next: (person) => {
        this.person = person;
        console.log(person);
      },
      complete: () => console.log('subscription completed'),
    });
  }

  ngOnDestroy() {
    this.unsubscribe$.next();
    this.unsubscribe$.complete();
    console.log('person component destroyed')
  }
}

@Injectable({ providedIn: 'root' })
export class ApiService {
  private cache = {};

  constructor(private http: HttpClient) {
  }

  public find(id = 1): Observable<any> {
    let res = this.cache[id];
    if (!res) {
      res = this.http.get(`https://swapi.co/api/people/${id}`).pipe(
        shareReplay({ bufferSize: 1, refCount: true }),
      )
      this.cache[id] = res;
    }
    return res;
  }
}
```

**Expected behavior**
After person component gets destroyed (for example, via navigation) it should be picked up by garbage collector and no reference of it should exist in the application.

**Environment**
- Runtime: Chrome latest
- RxJS version: 6

**Possible Solution**
To fix the issue instead of shareReplay use
```
 publishReplay(1),
 refCount(),
```

**Additional context/Screenshots (Heap Snapshots)**

[...]

PR #5044, opened 2019-09-28
fix: clear subscription on shareReplay completion

cartant:
**Description:**

In `shareReplay`, the subscription needs to be cleared to prevent the implementation holding a reference to the completed source. In Angular, not clearing the reference can lead to components not being garbage collected.

Using the DevTools, I've verified that the fix in this PR is required, but I have no idea how a test could be written for this sort of thing.

**Related issue (if exists):** #5034


gogakoreli:
Looks good, thanks for the help.

Issue #4968, opened 2019-08-15
TypeScript config not strict enough


cartant:
The TypeScript configuration used in this repo is insufficiently strict. The configuration should use `"strict": true` and should relax individual compiler options only if absolutely necessary.

For background, I have attempted to write a failing test to exercise the behvaiour described in [this comment](https://github.com/ReactiveX/rxjs/issues/4959#issuecomment-520629091) - that relates to a problem with the typing of the `filter` operator.

Unfortunately, I'm unable to do so as `dtslint` doesn't provide a mechanism for the types of errors involved in the snippet and because the compiler options used for the conventional tests are not strict enough - i.e. the snippet compiles without error.

cartant:
Closed by #5244.

PR #5045, opened 2019-09-28
chore: use strict TS option for dtslint

cartant:
**Description:**

As mentioned in #4968, the TypeScript compiler options could be more strict. Changing this is going to involve code changes. However, there's no reason we cannot enable `strict` for the dtslint tests right now. Doing so will allow us to use `$ExpectError` in more situations in our dtslint tests - the `'strictNullChecks': false` in the current configuration will prevent some types of errors from being reported. 

**Related issue (if exists):** #4968


cartant:
@benlesh It would be nice to get this merged, too.

PR #5052, opened 2019-10-02
test(audit operator): execute tests with TestScheduler#run

rall:
Use TestScheduler#run for the audit operator tests

As with #5038 and #5040 , run the specs with TestScheduler: 

- align the marbles with inner spaces
- uses '-' for ticks in subscription marbles

The alignment was tricky in a few cases: I'll highlight two examples of what I'm concerned about

### DIFF
--- a/spec/operators/audit-spec.ts
+++ b/spec/operators/audit-spec.ts
@@ -1,315 +1,388 @@
 import { expect } from 'chai';
+import { TestScheduler } from 'rxjs/testing';
 import { of, interval, EMPTY } from 'rxjs';
 import { audit, take, mergeMap } from 'rxjs/operators';
-import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
+import { observableMatcher } from '../helpers/observableMatcher';
 
 declare function asDiagram(arg: string): Function;
 
 /** @test {audit} */
 describe('audit operator', () => {
+  let testScheduler: TestScheduler;
+
+  beforeEach(() => {
+    testScheduler = new TestScheduler(observableMatcher);
+  });
+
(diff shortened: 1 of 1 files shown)


spec/operators/audit-spec.ts
```diff
@@ -1,315 +1,388 @@
...
-    expectSubscriptions(e2.subscriptions).toBe(e2subs);
+    testScheduler.run(({ hot, cold, expectObservable, expectSubscriptions }) => {
+      const e1 = hot('    -a-xy-----b--x--cxxx-|');
+      const e1subs = '    ^--------------------!';
+      const e2 = cold('    ----|                ');
+      const e2subs = [
+        '                 -^---!                ',
+        '                ----------^---!        ',
+        '                ----------------^---!  '
+      ];
```
rall: Should the `e2subs[1]` and `e2subs[2]` start a tick before all the other marbles like this? (actually i think previously it was 2 ticks - so I'll fix one way or the other)

spec/operators/audit-spec.ts
```diff
@@ -1,315 +1,388 @@
...
-    expectSubscriptions(e2.subscriptions).toBe(e2subs);
+    testScheduler.run(({ hot, cold, expectObservable, expectSubscriptions }) => {
+      const e1 = hot('  -a--------b-----c----|');
+      const e1subs = '  ^--------------------!';
+      const e2 = cold('  ----|                ');
+      const e2subs = [
+        '               -^---!                ',
+        '               ----------^---!       ',
+        '               ----------------^---! '
+      ];
```
rall: Or should the e2subs[1] and e2subs[2] starting ticks be aligned as in this example?

Issue #5051, opened 2019-10-02
Observable from webpack library doesn't unsubscribe


samal84:
## Bug Report

**Current Behavior**
The source observable doesn't unsubscribe when unsubscribing an observable that came from a library that returns rxjs observables.

The libray was bundled with Webpack.

**Reproduction**
https://github.com/samal84/rxjs-import-from-library-test

**Environment**
- Runtime: node 10.16.3, Chrome 77, Windows 10
- RxJS version: 6.5.3
- Webpack: 4.41.0

BioPhoton:
Hi @samal84. 

I run your repository to reproduce it. It's truly not working. I also changed you test observable to use Interval to even better demonstrate the problem. 
PRs are also sent to your repo with the changes.

In your application part, I copied the same code that you compile to compare it. This works as expected.

So the problem lies in your compilation process and can IMHO not be fixed in the library.
I suggest looking for how rxjs is compiled and what the difference is.

If @kwonoj or @cartant agrees, I would suggest closing this one.

kwonoj:
In example first rxjs is bundled in `lib` via webpack, and then *another* rxjs is bundled in `application`. This creates 2 sets of rxjs in application which doesn't passes instance check. Application should only have 1 set of observable.

samal84:
Thanks BioPhoton for the pull requests. I merged them.

I also added a passing test case without pipe or switchMap:
```
test$().pipe(
	take(1),
)
.subscribe(getObserver('compiled direct no pipe and switchMap'));
```

@kwonoj Does this mean that returning rxjs observable instances from libraries is not supported?

kwonoj:
> rxjs observable instances from libraries is not supported

no, library definitely can return observable as it needs, but in your application example you have **2** sets of rx (more than 1). That's problem need to be fixed.

samal84:
> > rxjs observable instances from libraries is not supported
> 
> no, library definitely can return observable as it needs, but in your application example you have **2** sets of rx (more than 1). That's problem need to be fixed.

The library in this case is using rxjs internally and then returning rxjs observables to callers. So it needs rxjs to be bundled along with it, right? Or else it cannot use rxjs internally.

The application is also using rxjs itself, so it also needs rxjs bundled with it for the same reason.

So this means that the lib and application get bundled separately, and get rxjs bundled into them separately.

How would we have only one rxjs when using rxjs inside both the lib and the app?

samal84:
[...]

PR #5059, opened 2019-10-05
fix: chain subscriptions from interop observables

cartant:
**Description:**

This PR fixes a bug in which subscriptions returned from interop observables are ignored - instead of being composed into the subscription chain. Prior to this fix, when an interop observable was returned to a flattening operator - like `mergeMap` or `switchMap` - the interop observable would not be unsubscribed upon explicit unsubscription from the flattening operator.

The bug was introduced in the changes made in https://github.com/ReactiveX/rxjs/pull/2479 and https://github.com/ReactiveX/rxjs/pull/4037.

This PR:

* adds a test helper for simulating an interop observable;
* copies existing, chaining-related tests and uses an interop observable within them;
* ensures subscriptions returned from interop observables are added to the subscription chain;
* fixes another bug in which `subscribeToResult` was passed outer values and indices when they should have been passed to the `InnerSubscriber` (the bug would have manifested itself in the now-deprecated result selectors for the flattening operators); and
* tweaks the `subscribeToResult` signature to make it safer (if an `InnerSubscriber` is passed, `value` and `index` must be `undefined` and must be passed via the `InnerSubscriber` instead).

**Related issue (if exists):** #5051

### DIFF
--- a/src/internal/operators/mergeMap.ts
+++ b/src/internal/operators/mergeMap.ts
@@ -146,6 +146,9 @@ export class MergeMapSubscriber<T, R> extends OuterSubscriber<T, R> {
     const destination = this.destination as Subscription;
     destination.add(innerSubscriber);
     const innerSubscription = subscribeToResult<T, R>(this, ish, undefined, undefined, innerSubscriber);
+    // The returned subscription will usually be the subscriber that was
+    // passed. However, interop subscribers will be wrapped and for
+    // unsubscriptions to chain correctly, the wrapper needs to be added, too.
     if (innerSubscription !== innerSubscriber) {
       destination.add(innerSubscription);
     }
--- a/src/internal/operators/mergeMap.ts
+++ b/src/internal/operators/mergeMap.ts
@@ -142,10 +142,13 @@ export class MergeMapSubscriber<T, R> extends OuterSubscriber<T, R> {
   }
 
   private _innerSub(ish: ObservableInput<R>, value: T, index: number): void {
-    const innerSubscriber = new InnerSubscriber(this, undefined, undefined);
+    const innerSubscriber = new InnerSubscriber(this, value, index);
     const destination = this.destination as Subscription;
     destination.add(innerSubscriber);
-    subscribeToResult<T, R>(this, ish, value, index, innerSubscriber);
+    const innerSubscription = subscribeToResult<T, R>(this, ish, undefined, undefined, innerSubscriber);
+    if (innerSubscription !== innerSubscriber) {
+      destination.add(innerSubscription);
+    }
   }
 
   protected _complete(): void {
--- a/src/internal/operators/mergeScan.ts
+++ b/src/internal/operators/mergeScan.ts
@@ -103,10 +103,13 @@ export class MergeScanSubscriber<T, R> extends OuterSubscriber<T, R> {
   }
 
   private _innerSub(ish: any, value: T, index: number): void {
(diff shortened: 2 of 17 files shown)


src/internal/operators/switchMap.ts
```diff
@@ -133,10 +133,13 @@ class SwitchMapSubscriber<T, R> extends OuterSubscriber<T, R> {
     if (innerSubscription) {
       innerSubscription.unsubscribe();
     }
-    const innerSubscriber = new InnerSubscriber(this, undefined, undefined);
+    const innerSubscriber = new InnerSubscriber(this, value, index);
     const destination = this.destination as Subscription;
     destination.add(innerSubscriber);
-    this.innerSubscription = subscribeToResult(this, result, value, index, innerSubscriber);
+    this.innerSubscription = subscribeToResult(this, result, undefined, undefined, innerSubscriber);
+    if (this.innerSubscription !== innerSubscriber) {
```
benlesh: Can we add comments in this section explaining why we're doing what we're doing? Otherwise, there's a lot of mental math for the next developer.

src/internal/operators/skipUntil.ts
```diff
@@ -74,7 +74,11 @@ class SkipUntilSubscriber<T, R> extends OuterSubscriber<T, R> {
     const innerSubscriber = new InnerSubscriber(this, undefined, undefined);
     this.add(innerSubscriber);
     this.innerSubscription = innerSubscriber;
-    subscribeToResult(this, notifier, undefined, undefined, innerSubscriber);
+    const innerSubscription = subscribeToResult(this, notifier, undefined, undefined, innerSubscriber);
+    if (innerSubscription !== innerSubscriber) {
```
benlesh: Add some comments explaining the "whys"

src/internal/operators/mergeScan.ts
```diff
@@ -103,10 +103,13 @@ export class MergeScanSubscriber<T, R> extends OuterSubscriber<T, R> {
   }
 
   private _innerSub(ish: any, value: T, index: number): void {
-    const innerSubscriber = new InnerSubscriber(this, undefined, undefined);
+    const innerSubscriber = new InnerSubscriber(this, value, index);
     const destination = this.destination as Subscription;
     destination.add(innerSubscriber);
-    subscribeToResult<T, R>(this, ish, value, index, innerSubscriber);
+    const innerSubscription = subscribeToResult<T, R>(this, ish, undefined, undefined, innerSubscriber);
```
benlesh: Same... add comments explaining the "whys"

src/internal/operators/mergeMap.ts
```diff
@@ -142,10 +142,13 @@ export class MergeMapSubscriber<T, R> extends OuterSubscriber<T, R> {
   }
 
   private _innerSub(ish: ObservableInput<R>, value: T, index: number): void {
-    const innerSubscriber = new InnerSubscriber(this, undefined, undefined);
+    const innerSubscriber = new InnerSubscriber(this, value, index);
     const destination = this.destination as Subscription;
     destination.add(innerSubscriber);
-    subscribeToResult<T, R>(this, ish, value, index, innerSubscriber);
+    const innerSubscription = subscribeToResult<T, R>(this, ish, undefined, undefined, innerSubscriber);
```
benlesh: same.

cartant:
Added comments.

benlesh:
@cartant: Unfortunately, the patch branch will need a separate PR. Can you tackle that? The problem is that the tests have changed between the two branches.

Issue #4578, opened 2019-02-19
finalize docs are incomplete


cartant:
### Documentation Related To Component:

The `finalize` operator.

### Please check those that apply

- [ ] typo
- [ ] documentation doesn't exist
- [x] documentation needs clarification
- [ ] error(s) in example
- [x] needs example

### Description Of The Issue

The docs state:

> Returns an Observable that mirrors the source Observable, but will call a specified function when the source terminates on complete or error.

`finalize` will also call its callback if the subscriber explicitly unsubscribes. See [this test](https://twitter.com/1StevieKilner/status/1097228842194948096).

The documentation needs an example, too.

---

Okay. I clearly pasted the wrong link. It's [this test](https://github.com/ReactiveX/rxjs/blob/6.4.0/spec/operators/finalize-spec.ts#L43-L52).

dzhavat:
@cartant Hilarious :D The `this test` link points to a tweet of a dog holding two boiled eggs in its mouth 😂

cartant:
@dzhavat I've added the correct link. And I've left the mistakenly-pasted link for the amusement of others.)

lukk84:
Here is another reproduction of this issue
https://stackblitz.com/edit/rxjs-bsv8sg

avrim:
HI, can we confirm that indeed finalize calls it's method in case we unsubscribe manually, before observable's execution reached complete  / error ? 
This is sad that I discovered this issue here :(

BioPhoton:
@cartant thanks a lot for the link with the dog. It was really nice having this in between all the issue stuff. Haha

Anyway, i will put it on my todo list.

ClementVidal:
Same behavior observed and discussed here:
https://stackoverflow.com/questions/57696249/rxjs-switchmap-and-tap-issue

IMHO this should deserve attention as it somehow introduce a new hookable "moment" in the lifetime of an observable which is not described yet .

PDavid:
Hi, I'd like to help with this issue. 
I have some experience with RxJS with developing Angular applications at my job.
Maybe this is a silly question, but to which branch should the PR of the documentation improvement be targeted? Should it be master or 6.x branch?

cartant:
@PDavid Please target `master`. Thanks.

PDavid:
@cartant Thanks, then I'll redo my changes and open a new PR to master.

cartant:
[...]

PR #5065, opened 2019-10-09
docs(finalize): examples, improvement #4578

PDavid:
**Description:**
- Added the fact that to the documentation of finalize operator that it will also call its callback if the subscriber explicitly unsubscribes. (This was missing from the documentation before.)
- Added example for finalize basic usage and when the subscriber explicitly unsubscribes.

I targeted 6.x branch, not sure if this is correct. Should it be master?

I'm not sure if this is the correct approach to submit a PR and I'm open to any feedback.
Thanks and Regards,
Dávid

#4578


PDavid:
I'll redo this change targeting master and open a new PR.

PR #5067, opened 2019-10-10
docs(finalize): examples, improvement #4578

PDavid:
**Description:**
- Added the fact that to the documentation of finalize operator that it will also call its callback if the subscriber explicitly unsubscribes. (This was missing from the documentation before.)
- Added example for finalize basic usage and when the subscriber explicitly unsubscribes.

I'm not sure if this is the correct approach to submit a PR and I'm open to any feedback.
Thanks and Regards,
Dávid

#4578

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 integration/side-effects/snapshots/
+docs_app/src/ie-polyfills.js
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,1 @@
 integration/side-effects/snapshots/
-docs_app/src/ie-polyfills.js
--- a/src/internal/operators/finalize.ts
+++ b/src/internal/operators/finalize.ts
@@ -18,13 +18,19 @@ import { MonoTypeOperatorFunction, TeardownLogic } from '../types';
  *
  * // emit value in sequence every 1 second
  * const source = interval(1000);
- * // 0,1,2,3,4,5....
  * const example = source.pipe(
  *   take(5), //take only the first 5 values
  *   finalize(() => console.log('Sequence complete')) // Execute when the observable completes
  * )
  * const subscribe = example.subscribe(val => console.log(val));
- * // 0,1,2,3,4,Sequence complete
+ *
+ * // results:
+ * //   0
+ * //   1
+ * //   2
+ * //   3
+ * //   4
+ * //   'Sequence complete'
  * ```
  *
  * Execute callback function when the subscriber explicitly unsubscribes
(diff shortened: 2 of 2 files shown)


src/internal/operators/finalize.ts
```diff
@@ -7,6 +7,43 @@ import { MonoTypeOperatorFunction, TeardownLogic } from '../types';
...
+ * ```ts
+ * import { interval, timer, noop } from 'rxjs';
+ * import { finalize, tap } from 'rxjs/operators';
+ *
+ * const source = interval(100).pipe(
+ *   finalize(() => console.log('[finalize] Called')),
+ *   tap(noop, noop, () => console.log('[tap] Not called')),
+ * );
+ *
+ * const sub = source.subscribe(x => console.log(x), noop, () => console.log('[complete] Not called'));
```
cartant: It's not particularly important in this example, but I would suggest passing `undefined` for the error callback. Passing a `noop` will swallow the error. Passing `undefined` will see any error thrown asynchronously as unhandled. I mention this because I think it would be good to avoid situations where a newcomer to RxJS might see this example and take up the pattern of passing `noop` and inadvertently introduce the swallowing of errors into their code.

Even better, though, pass an observer, like this:

```
...subscribe({
  next: value => /* whatever */,
  complete: () => /* whatever */
);
```

src/internal/operators/finalize.ts
```diff
@@ -7,6 +7,43 @@ import { MonoTypeOperatorFunction, TeardownLogic } from '../types';
...
+ *
+ * timer(150).subscribe(() => sub.unsubscribe());
+ * // 0,[finalize] Called
```
cartant: Console output on separate lines, please.

.eslintignore
```diff
@@ -1 +1,2 @@
-integration/side-effects/snapshots/
\ No newline at end of file
+integration/side-effects/snapshots/
+docs_app/src/ie-polyfills.js
```
cartant: Please remove this change so that configuration/build environment changes are not included with documentation changes. AFAICT, ESLint is only run over `.js` files that are included in commits and there are no such files in this PR.

PDavid: This is kind of interesting because I did not changed this file but when I rebased my branch onto master and pushed, this change appeared. I also checked the git log on this file and I do not have any commit on it.

JWO719: I think this is just part in the stable branch, we might want to exclude eslint for the docs app completly. So for now I would appreciate if you could revert this change, so that we can have a proper decision how to proceed with eslint and the docs app

PDavid: OK thanks, I reverted the eslint related change.

PDavid:
Hi @JWO719,
Is there still something which I'd need to do change in this PR?
I think this should be OK to merge.

JWO719:
sorry for coming back super late regarding that pr. Many thanks for your contribution. I'm happy to merge it finally

PDavid:
> 
> 
> sorry for coming back super late regarding that pr. Many thanks for your contribution. I'm happy to merge it finally

Of course no problem at all @JWO719. :)
I'd like to thank you and @cartant for reviewing my PR and providing feedback.  👍

PR #5071, opened 2019-10-11
test(combineAll, combineLatest operators): execute tests with TestScheduler#run

rall:
Use TestScheduler#run for the combineAll and combineLatest operator tests

### DIFF
--- a/spec/operators/combineAll-spec.ts
+++ b/spec/operators/combineAll-spec.ts
@@ -78,10 +78,10 @@ describe('combineAll operator', () => {
 
   it('should work with empty and empty', () => {
     testScheduler.run(({ cold, expectObservable, expectSubscriptions }) => {
-      const e1 = cold('|');
-      const e1subs = ' (^!)';
-      const e2 = cold('|');
-      const e2subs = ' (^!)';
+      const e1 = cold(' |');
+      const e1subs = '  (^!)';
+      const e2 = cold(' |');
+      const e2subs = '  (^!)';
       const expected = '|';
 
       const result = of(e1, e2).pipe(combineAll((x, y) => x + y));
--- a/spec/operators/combineAll-spec.ts
+++ b/spec/operators/combineAll-spec.ts
@@ -489,7 +489,7 @@ describe('combineAll operator', () => {
 
   it('should handle two consecutive hot observables with error right', () => {
     testScheduler.run(({ hot, expectObservable, expectSubscriptions }) => {
-      const left =  hot('--a--^--b--c--|');
+      const left = hot(' --a--^--b--c--|');
       const leftSubs = '      ^--------!';
       const right = hot('-----^----------d--e--f--#', undefined, 'dun dun dun');
       const rightSubs = '     ^-------------------!';
--- a/spec/operators/combineAll-spec.ts
+++ b/spec/operators/combineAll-spec.ts
@@ -1,453 +1,522 @@
 import { expect } from 'chai';
 import { queueScheduler, of, Observable } from 'rxjs';
 import { combineAll, mergeMap } from 'rxjs/operators';
-import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
+import { TestScheduler } from 'rxjs/testing';
+import { observableMatcher } from '../helpers/observableMatcher';
 
 declare function asDiagram(arg: string): Function;
 declare const type: Function;
 
 /** @test {combineAll} */
 describe('combineAll operator', () => {
+  let testScheduler: TestScheduler;
+
+  beforeEach(() => {
+    testScheduler = new TestScheduler(observableMatcher);
+  });
+
   asDiagram('combineAll')('should combine events from two observables', () => {
-    const x =    cold(               '-a-----b---|');
-    const y =    cold(               '--1-2-|     ');
-    const outer = hot('-x----y--------|           ', { x: x, y: y });
-    const expected =  '-----------------A-B--C---|';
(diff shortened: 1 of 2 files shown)


spec/operators/combineAll-spec.ts
```diff
@@ -1,453 +1,522 @@
...
-
-    expectObservable(result).toBe(expected);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
-    expectSubscriptions(e2.subscriptions).toBe(e2subs);
+    testScheduler.run(({ cold, expectObservable, expectSubscriptions }) => {
+      const e1 = cold('|');
+      const e1subs = ' (^!)';
+      const e2 = cold('|');
+      const e2subs = ' (^!)';
+      const expected = '|';
```
cartant: I think `expected` was misaligned in the previous version. It should line up with the other bars.

spec/operators/combineAll-spec.ts
```diff
@@ -1,453 +1,522 @@
...
-
-    expectObservable(result).toBe(expected);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
-    expectSubscriptions(e2.subscriptions).toBe(e2subs);
+    testScheduler.run(({ hot, expectObservable, expectSubscriptions }) => {
+      const e1 = hot('-a-^-|');
+      const e1subs = '   ^-!';
+      const e2 = hot('------'); //never
+      const e2subs = '   ^--';
+      const expected = ' -'; //never
```
cartant: If you want, you can extend the `-` in `expected` to line up with the end of `e2` - like in the test that follows this one. If the marble diagram has nothing but `-` in it, where is ends doesn't matter.

spec/operators/combineAll-spec.ts
```diff
@@ -1,453 +1,522 @@
...
-    const rightSubs =      '^                   !';
-    const expected =       '-----------x--y--z--#';
-
-    const result = of(left, right).pipe(combineAll((x, y) => x + y));
-
-    expectObservable(result).toBe(expected, { x: 'cd', y: 'ce', z: 'cf' }, 'dun dun dun');
-    expectSubscriptions(left.subscriptions).toBe(leftSubs);
-    expectSubscriptions(right.subscriptions).toBe(rightSubs);
+    testScheduler.run(({ hot, expectObservable, expectSubscriptions }) => {
+      const left =  hot('--a--^--b--c--|');
```
cartant: There are two spaces in `=  hot`, so Prettier will effect a misalignment here.

spec/operators/combineLatest-spec.ts
```diff
@@ -1,480 +1,551 @@
...
+  });
+
   asDiagram('combineLatest')('should combine events from two cold observables', () => {
-    const e1 =   cold('-a--b-----c-d-e-|');
-    const e2 =   cold('--1--2-3-4---|   ');
-    const expected = '--A-BC-D-EF-G-H-|';
+    testScheduler.run(({ cold, expectObservable }) => {
+      const e1 = cold('  -a--b-----c-d-e-|');
+      const e2 = cold('  --1--2-3-4---|   ');
+      const expected = '--A-BC-D-EF-G-H-|';
```
cartant: I think the original version was misaligned (you will find a bit of that in the code base). All of the snippets should start at the same time.

spec/operators/combineLatest-spec.ts
```diff
@@ -1,480 +1,551 @@
...
-
-    expectObservable(result).toBe(expected);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
-    expectSubscriptions(e2.subscriptions).toBe(e2subs);
+    testScheduler.run(({ cold, expectObservable, expectSubscriptions }) => {
+      const e1 = cold('|');
+      const e1subs = ' (^!)';
+      const e2 = cold('|');
+      const e2subs = ' (^!)';
+      const expected = '|';
```
cartant: `expected` in the original version was misaligned.

spec/operators/combineLatest-spec.ts
```diff
@@ -1,480 +1,551 @@
...
-    expectSubscriptions(e2.subscriptions).toBe(e2subs);
+    testScheduler.run(({ hot, expectObservable, expectSubscriptions }) => {
+      const values = {
+        a: 1
+      };
+      const e1 = hot('-a-^-|', values);
+      const e1subs = '   ^-!';
+      const e2 = hot('------', values); //never
+      const e2subs = '   ^--';
+      const expected = ' -'; //never
```
cartant: As in the next test, `expected` can end where `e2` ends, but it was no impact on the test.

rall:
ah, sorry - I should have checked my work more carefully. I'll make these fixes now

spec/operators/combineAll-spec.ts
rall: I'd think we would want to align on the ^ in`(^!)`, but it looks like we align on the opening parenthesis in other examples so I'll do that here - is that right?

rall:
All set I think. I'll look out for these issues in the other operator specs as I get them

spec/operators/combineAll-spec.ts
cartant: The `(^!)` is a subscription marble: `^` represents subscription and `!` represents unsubscription. The parentheses mean that the contents occur in the frame at which the `(` is placed, so, yeah, it needs to align with the `(`.

Note that `^` has a different meaning when it's used in an observable's marble diagram. There it represents the zero frame.

benlesh:
YES! So good!

Issue #4196, opened 2018-09-27
Typing of toPromise is wrong


plol:
## Bug Report

**Current Behavior**
toPromise has the wrong type

**Reproduction**
```js
const p = of<string>().toPromise() // Type is Promise<string>
```

**Expected behavior**
p will in fact resolve to undefined and should have the type Promise<string | undefined>

cartant:
This is a fairly subtle problem.

The promise will resolve when the observable completes - resolving with the last emitted value or with `undefined` if a value wasn't emitted.

However, the type declarations cannot express whether or not an observable will emita value, so it could be argued that the type for the promise should **always** be `Promise<T | undefined>`.

That would more correctly describe the behaviour, but it would be a breaking change.

benlesh:
This was fixed in #5072

PR #5072, opened 2019-10-11
fix(toPromise): correct toPromise return type

jchv:
**Description:**
As described in #4196, the typing of `toPromise` does not match reality as it may return `undefined` in the case that no value is ever emitted. We have hit this bug in practice and believe it would be very useful to fix the actual typing.

(This does not implement the `default` argument as suggested in #4196 - while this seems like a reasonable addition, there are cases today where returning undefined is in fact desired, and it would be nice if the typing was correct for this case.)

*    #4196 - Typing of toPromise is wrong

(cc @sadarby)

### DIFF
--- a/spec-dtslint/Observable-spec.ts
+++ b/spec-dtslint/Observable-spec.ts
@@ -126,4 +126,8 @@ describe('pipe', () => {
     const customOperator = () => <T>(a: Observable<T>) => a;
     const o = of('foo').pipe(customOperator()); // $ExpectType Observable<string>
   });
+
+  it('should have proper return type for toPromise', () => {
+    const o = of('foo').toPromise(); // $ExpectType Promise<string | undefined>
+  });
 });
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -347,9 +347,9 @@ export class Observable<T> implements Subscribable<T> {
   }
 
   /* tslint:disable:max-line-length */
-  toPromise<T>(this: Observable<T>): Promise<T>;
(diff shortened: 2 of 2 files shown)


cartant:
LGTM. Do you think you could add some dtslint tests? They would need to go into this file: https://github.com/ReactiveX/rxjs/blob/master/spec-dtslint/Observable-spec.ts

Also, whilst I agree with this change, there will need to be some discussion about it before it's merged. Although it fixes a problem, it is a breaking change. Whether or not it's okay for v7 or will have to wait until v8 will have to be determined.

jchv:
Thanks, done. As a note, I had to run `mkdir "$HOME/.dts"` for `dtslint` to succeed, I am not sure why or if that's intended.

>Also, whilst I agree with this change, there will need to be some discussion about it before it's merged. Although it fixes a problem, it is a breaking change. Whether or not it's okay for v7 or will have to wait until v8 will have to be determined.

Understood. There's definitely other approaches with varying pros/cons, but I think this one is pretty straightforward so I hope it or a compatible variant of it ends up working out.

cartant:
> As a note, I had to run mkdir "$HOME/.dts" for dtslint to succeed, I am not sure why or if that's intended.

Yeah, that's happened to me, too. dtslint has been an ongoing source of frustration. It's a really useful tool and has helped improve the codebase, but some of the things it does are so annoying.

benlesh:
I think the real fix for `toPromise` is that it should reject if no value arrives with a known error.

I think we should discuss this at the next meeting.

jchv:
That also SGTM.

benlesh:
Approved, but as a BREAKING CHANGE... (from Core Team meeting)

PR #5076, opened 2019-10-15
feat(lastValue): Adds lastValue method to Observable

### LINKED ISSUE #5075: toPromise should reject if no value is emitted.
### Current Behavior

`toPromise` will resolve with `undefined` if the source observable simply completes without emitting a value. 

```
EMPTY.toPromise().then(x => console.log(x)); // logs "undefined"
```

### The problem

Promises are a guarantee to a future value. With `toPromise()` we should be saying "this is the last value we've gotten prior to completion". But what we're actually saying is "this *might* be the last value we've gotten prior to completion". It's semantically unsound, IMO.

If the developer wants a guaranteed value from a source, and that value *might* be `undefined`, there's currently *no way* for that developer to discern whether or not the `undefined` that `toPromise` resolved with was the `undefined` from the source, or just happenstance from completion without a value.

```ts
const promiseTheresAValue1 = of(1, 2, 3, undefined).toPromise();

const promiseTheresAValue2 = EMPTY.toPromise(); // Oops, there wasn't actually a value
```

### Proposal

Have `toPromise` reject with an `EmptyError` if no value is emitted prior to completion of the source.

```ts
EMPTY.toPromise()
  .then(() => console.log('not hit'))
  .catch(err => console.log(err instanceof EmptyError)); // logs "true"
```

**Optionally** we could introduce a `last()` or `lastValue()` method to `Observable` and deprecate `toPromise` to reduce confusion while migrating to this change, while also being able ot introduce this better semantic sooner.

Related: https://github.com/ReactiveX/rxjs/issues/4993

benlesh:
**PROPOSED CHANGE**

This is meant to be a more well-behaved, a better-named, replacement for `toPromise`.

Resolves #5075

### DIFF
--- a/spec-dtslint/Observable-spec.ts
+++ b/spec-dtslint/Observable-spec.ts
@@ -1,4 +1,4 @@
-import { Observable, of, OperatorFunction } from 'rxjs';
+import { Observable, of, OperatorFunction, EMPTY, NEVER } from 'rxjs';
 import { mapTo } from 'rxjs/operators';
 
 function a<I extends string, O extends string>(input: I, output: O): OperatorFunction<I, O>;
@@ -11,7 +11,7 @@ function a<I, O extends string>(output: O): OperatorFunction<I, O>;
  * That is, `a('0', '1')` returns `OperatorFunction<'0', '1'>`.
  * That means that the `a` function can be used to create consecutive
  * arguments that are either compatible or incompatible.
- * 
+ *
  * ```javascript
  * a('0', '1'), a('1', '2') // OK
  * a('0', '1'), a('#', '2') // Error '1' is not compatible with '#'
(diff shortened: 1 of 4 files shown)


ncjamieson:
@benlesh Could this be refactored into the static method that was discussed in the core team meeting?

```ts
import { lastValue, of } from "rxjs";

function example () {
  const value = await lastValue(of(42));
}
```

IIRC, I might have previously had reservations about the name. If that was the case, I've changed my mind. I think `lastValue` is nice and is a big improvement on `toPromise`. (The name describes the behaviour and the types will tell people that it returns a promise.)

cartant:
Closing this in favour of https://github.com/ReactiveX/rxjs/pull/5295

PR #5080, opened 2019-10-15
feat(ajax): new ajax implementation

kwonoj:
**Description:**
This PR is discussion for implementing new ajax potentially replaces existing ajax implementation. Majority of implementation is coming from ng's `http`, could be used without dependency to ng & its dependency injection. Notable diff to existing ajax are

- `Ajax` class itself is no longer observable, but its method returns observable.
- explicit rejection on node.js unless custom adapter provided (due to some dependency issue, node.js adapter impl won't be included)
- through test cases
- request / response handling behavior resolves most of known issues 

Since this implementation is radically different to existing interfaces, it's exported with `__unstable` prefix anyone can opt in to try this.


benlesh:
We should talk with @IgorMinar about maybe helping them make their HTTP implementation Angular-agnostic (which is basically what this is) and available to any project that wants to leverage RxJS.

benlesh:
@kwonoj, after the many, many improvements to the `ajax` function in v7, I think we can close this, but I'll let you do that. Ideally you can link to the library you wrote here.

Issue #5066, opened 2019-10-10
Type signature review progress for v7 beta


cartant:
This issue is for the tracking of the review of the TypeScript signatures - that involve either an `Observable` or `ObservableInput` rest parameter or a callback that returns `ObservableInput` or have open, type-related issues, etc. - as per the discussion in today's core-team meeting.

Deprecations for the scheduler-related changes are not in this list. My understanding is that those deprecations will remain in place, with the breaking changes being made in version 8.

@benlesh As I've review individual source files, I'll check them off in this list if they're okay and will link to a comment (or issue) if changes are required or attention is warranted.

In addition to what's shown below, I think we should look at how `null` and `undefined` are used to ensure that passing either- to indicate the absence of an argument - is supported.

* [x] `AsyncSubject.ts`
* [x] `BehaviorSubject.ts`
* [x] `Notification.ts` - see #5478, #5031
* [x] `Observable.ts` - see https://github.com/ReactiveX/rxjs/issues/3890, https://github.com/ReactiveX/rxjs/issues/4159, https://github.com/ReactiveX/rxjs/issues/4415 and https://github.com/ReactiveX/rxjs/pull/5072
* [x] `Observer.ts`
* [x] `Operator.ts`
* [x] `ReplaySubject.ts`
* [x] `Scheduler.ts`
* [x] `Subject.ts` - see [this comment](https://github.com/ReactiveX/rxjs/issues/5066#issuecomment-540305826)
* [x] `Subscriber.ts`
* [x] `Subscription.ts`
* [ ] `types.ts` - see https://github.com/ReactiveX/rxjs/issues/4532 and [this comment](https://github.com/ReactiveX/rxjs/issues/5066#issuecomment-540387259) and [this comment](https://github.com/ReactiveX/rxjs/issues/5066#issuecomment-586541767)
* [ ] `observable/bindCallback.ts` - needs N-args signature for values? Should fall back to `unknown` instead of `any`. Also, see [this comment](https://github.com/ReactiveX/rxjs/pull/5781#pullrequestreview-499090474).
* [ ] `observable/bindNodeCallback.ts` - needs N-args signature for values? Should fall back to `unknown` instead of `any`. Also, see [this comment](https://github.com/ReactiveX/rxjs/pull/5781#pullrequestreview-499090474).
* [x] `observable/combineLatest.ts` - needs N-args signature for observable inputs? See also #5209
* [x] `observable/concat.ts`
* [x] `observable/defer.ts`
[...]

PR #5084, opened 2019-10-16
fix: use `unknown` type for errors

OliverJAsh:
**Description:**
Errors can be of any type. `unknown` is preferable to `any`, as `any` completely disables type checking:

```ts
declare const myError: any;
myError.some.property.that.does.not.exist // runtime error!
```

### DIFF
--- a/src/internal/types.ts
+++ b/src/internal/types.ts
@@ -61,21 +61,21 @@ export type InteropObservable<T> = { [Symbol.observable]: () => Subscribable<T>;
 export interface NextObserver<T> {
   closed?: boolean;
   next: (value: T) => void;
-  error?: (err: any) => void;
+  error?: (err: unknown) => void;
   complete?: () => void;
 }
 
 export interface ErrorObserver<T> {
   closed?: boolean;
   next?: (value: T) => void;
-  error: (err: any) => void;
+  error: (err: unknown) => void;
   complete?: () => void;
 }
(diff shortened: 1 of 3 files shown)


src/internal/types.ts
```diff
@@ -61,21 +61,21 @@ export type InteropObservable<T> = { [Symbol.observable]: () => Subscribable<T>;
 export interface NextObserver<T> {
   closed?: boolean;
   next: (value: T) => void;
-  error?: (err: any) => void;
+  error?: (err: unknown) => void;
```
OliverJAsh: There might be other places to update this.

benlesh:
Yeah, this would be a breaking change. And I'm not sure it's correct. Sure, the type of error is "unknown", but the error could also literally be anything. This change would force everyone to explicitly define the type of error they think they're going to get, and would be a pretty massive breaking change for people.

We can discuss this at the next team meeting, but I'm going to say it's unlikely we'll be willing to introduce such a massive breakage for TS users in this next version.

cartant:
Related: https://github.com/ReactiveX/rxjs/issues/5066#issuecomment-540387259

Basically, I don't think we should make changes to this until TypeScript does, too - e.g. for `Promise`.

benlesh:
We've decided to leave as-is, until TS does something similar with Promise's error path.

PR #5089, opened 2019-10-17
feat(Subject): SnapShotReplaySubject

mrudangit:
SnapShotReplay Subject will collect items based on some key
if new value with same key is published it will replace the old value

**Description:**


kwonoj:
I'm closing this for now as stated above we recommend to publish userland operator for custom operator behavior in most of cases.

PR #5092, opened 2019-10-18
docs(readme): update readme file

dzhavat:
Update the main readme file

**Description:**
Since there's no `stable` branch, I thought using the `6.x` will be more correct.

And I've also copied a line from `5.x` branch about making PRs. I've seen at least one person asking for this in a comment. Another option could be adding that line to the PR template.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Reactive Extensions Library for JavaScript. This is a rewrite of [Reactive-Exten
 - [master](https://github.com/ReactiveX/rxjs/commits/master) - This is all of the current, unreleased work, which is against v7 of RxJS right now
 - [stable](https://github.com/ReactiveX/rxjs/tree/6.x) - This is the branch for the latest version you'd get if you do `npm install rxjs`
 
-Most PRs should be made to **master**, unless you know it is a breaking change.
+Most PRs should be made to **master**.
 
 ## Important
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,8 +22,10 @@ Reactive Extensions Library for JavaScript. This is a rewrite of [Reactive-Exten
 
 ## Versions In This Repository
 
-- [master](https://github.com/ReactiveX/rxjs/commits/master) - This is all of the current, unreleased work, which is against v6 of RxJS right now
-- [stable](https://github.com/ReactiveX/rxjs/commits/stable) - This is the branch for the latest version you'd get if you do `npm install rxjs`
+- [master](https://github.com/ReactiveX/rxjs/commits/master) - This is all of the current, unreleased work, which is against v7 of RxJS right now
+- [stable](https://github.com/ReactiveX/rxjs/tree/6.x) - This is the branch for the latest version you'd get if you do `npm install rxjs`
+
+Most PRs should be made to **master**, unless you know it is a breaking change.
```
cartant: I think we can remove the "unless you know it is a breaking change".

TBH, that makes no sense to me: there are two branches listed here and this implies that if the submitter knows it's a breaking change it should target ... `stable`? 🤷‍♂

IMO, it should just say "Most PRs should be made to **master**."

My understanding is that we change the branch to which `master` refers to manage this. E.g. until fairly recently, `master` was the `6.x` branch, but now that the version 7 release is imminent, we want PRs to target version 7, so that's the `master` branch. Contributors should always target `master` because we configure `master` to be the branch to which PRs should be targeted.

dzhavat: Makes sense. Just made the change :)

Issue #5090, opened 2019-10-17
Amendment to `mapTo` operator type


alaboudi:
## Bug Report
This issue is not a "bug" per se but more of an amendment suggestion to the `mapTo` operators's type. Didn't know what best to label this as.

**Current Behavior**
`mapTo`'s current type signature is `mapTo<T, R>(value: R): OperatorFunction<T, R>` but there is absolutely no need for the `T` generic for this operator. The whole point of the operator is to start a completely new stream of values with no dependency on the source stream value.

**Possible Solution**
I think `mapTo<R>(value: R): OperatorFunction<any, R>` may better represent the intention of using the `mapTo` operator.

kwonoj:
Sounds ok, presume this won't be breaking?

alaboudi:
@kwonoj If we want to not make this a breaking change, then I think we can overload the type definition of this function to the current two generic type along with the new suggested one generic type. Otherwise, this is a breaking change. Since I see @Cartant referencing this suggestion in the v7 rollout, is there a point of making this a non breaking introduction when the current implementation doesn't capture the intent of the operator?

cartant:
> I think we can overload the type definition of this function to the current two generic type along with the new suggested one generic type...

Yeah. ATM, this is a breaking change: it will break anyone who explicitly specifies the `T` type argument. I agree with the change, as `T` serves no purpose. We could do the same thing as with `of` and provide the current signature - that includes the `T` - as a deprecated signature.

alaboudi:
I like @cartant 's suggestion to add the new signature and mark the current one as a deprecated signature. It won't be a breaking change but will push people towards the correct direction till its phased. If there are no objections, I'll put up a PR soon to address this issue.

alaboudi:
@cartant I have put up a PR 3 days ago, please have a look.

PR #5102, opened 2019-10-21
refactor(mapTo): update function signature

alaboudi:
**Description:**
Reduce the number of generics used in the 'mapTo' function to 1 generic parameter. Mark the 2 generic variant as a deprecated variant

issue #5090


benlesh:
Thank you @alaboudi!

PR #5110, opened 2019-10-26
fix(fromFetch): abort signal ignored if already aborted

bgotink:
**Description:**

Fix two issues with abort signal handling in `fromFetch`:

- The `signal` property on the init object was modified, which could lead to issues if the init object is reused for multiple requests.
- The `fromFetch` function basically ignored the incoming signal if it is already aborted.

### DIFF
--- a/spec/observables/dom/fetch-spec.ts
+++ b/spec/observables/dom/fetch-spec.ts
@@ -173,13 +173,23 @@ describe('fromFetch', () => {
   });
 
   it('should allow passing of init object', done => {
-    const myInit = {};
-    const fetch$ = fromFetch('/foo', myInit);
+    const fetch$ = fromFetch('/foo', {method: 'HEAD'});
     fetch$.subscribe({
       error: done,
       complete: done,
     });
-    expect(mockFetch.calls[0].init).to.equal(myInit);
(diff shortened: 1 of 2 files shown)


spec/observables/dom/fetch-spec.ts
```diff
@@ -173,13 +177,13 @@ describe('fromFetch', () => {
...
+    const myInit = {myInit: true};
+    const fetch$ = fromFetch('/foo', myInit as {});
     fetch$.subscribe({
       error: done,
       complete: done,
     });
-    expect(mockFetch.calls[0].init).to.equal(myInit);
+    expect((mockFetch.calls[0].init as any).myInit).to.be.true;
     expect(mockFetch.calls[0].init.signal).not.to.be.undefined;
   });
```
cartant: I would like to see the test description changed to specifically state that the `init` object should not be mutated.

Also, I'd like to see `expect(mockFetch.calls[0].init).to.not.equal(myInit);` added as an expectation.

And I would like to see one of the actual `RequestInit` properties used instead of the invalid `myInit` property. For example: `const myInit = { method: 'HEAD' }`

bgotink: Thanks for the feedback! I've updated the tests

Issue #5047, opened 2019-09-30
TypeScript 3.6 compatibility


mprobst:
RxJS does not compile with TypeScript 3.6.

Here's a list of issues and (presumed) problems that need to be fixed:

* [ ] NodeJS typings are out of date, contain conflicting type for `IteratorResult`

```
node_modules/@types/node/index.d.ts:74:11 - error TS2300: Duplicate identifier 'IteratorResult'.

74 interface IteratorResult<T> { }
             ~~~~~~~~~~~~~~

  node_modules/typescript/lib/lib.es2015.iterable.d.ts:41:6
    41 type IteratorResult<T, TReturn = any> = IteratorYieldResult<T> | IteratorReturnResult<TReturn>;
            ~~~~~~~~~~~~~~
    'IteratorResult' was also declared here.
...
```

* [ ] `from(ObservableInput<T> val)` in TypeScript 3.6 no longer resolves to `Observable<T>`, but to `Observable<unknown>`. I don't know why exactly, but it looks as if TS resolves the call to the overload `export function from<O extends ObservableInput<T>, T=any>(input: O): Observable<ObservedValueOf<O>>;`, then resolving `type ObservedValueOf<T> = O extends ObservableInput<infer T> ? T : never` fails.

```
src/internal/observable/using.ts:55:43 - error TS2769: No overload matches this call.
  Overload 1 of 5, '(observer?: PartialObserver<unknown>): Subscription', gave the following error.
    Argument of type 'Subscriber<T>' is not assignable to parameter of type 'PartialObserver<unknown>'.
      Type 'Subscriber<T>' is not assignable to type 'CompletionObserver<unknown>'.
        Types of property 'next' are incompatible.
          Type '(value?: T) => void' is not assignable to type '(value: unknown) => void'.
            Types of parameters 'value' and 'value' are incompatible.
              Type 'unknown' is not assignable to type 'T'.
                'unknown' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint '{}'.
  Overload 2 of 5, '(next?: (value: unknown) => void, error?: (error: any) => void, complete?: () => void): Subscription', gave the following error.
    Argument of type 'Subscriber<T>' is not assignable to parameter of type '(value: unknown) => void'.
      Type 'Subscriber<T>' provides no match for the signature '(value: unknown): void'.

55     const subscription = source.subscribe(subscriber);
                                             ~~~~~~~~~~
```

[...]

PR #5124, opened 2019-11-04
Upgrade RxJS to TS-3.7

kolodny:
This should unblock Google from compiling RxJS with TS3.6, the other issues "should" be able to be fixed internally.

This should make a dent on #5047


benlesh:
@kolodny looks like there's one build error that needs to be resolved before I can merge this.

kolodny:
> @kolodny looks like there's one build error that needs to be resolved before I can merge this.

Resolved.

PR #5126, opened 2019-11-05
2 new interfaces improve rxjs's composability

alshdavid:
**Description:**

rxjs is an indispensable part of my development toolkit. This pull requests seeks to extend rxjs's composability by introducing 2 new interfaces to describe commonly used behaviours.

The introduced types enable the ability to take advantage of composition and interface segregation. My objective is to enable greater flexibility in how we work with basic observable types.

**Explanation:**

_I apologise in advance to those already familiar with the concepts_

```typescript
import { ValueGetter, Subscribable } from 'rxjs'

type ValueGetterSubscriber<T> = (
  ValueGetter<T> &
  Subscribable<T>
)

function doSomething(v: ValueGetterSubscriber<string>) {
  console.log(v.getValue())
  v.subscribe(console.log)
}
```

In the above example, a developer could pass in anything that implicitly matches the composite type signature. 

An example of such a type would be the `BehaviorSubject`. Defining types like this, with the expectation of implicit interface satisfaction allows for low coupling yet yields high cohesion.

This can be extended to enabling the construction of interesting objects from smaller basic functionalities. 

Imagine an rxjs-powered wrapper for something like Redux. 

Rather than using a concrete reference like `redux.Store`, developers would describe the behavior of the dependency and implicitly satisfy it. 

This makes testing fantastic and because developers have the ability to pick out what methods are used from that dependency (interface segregation), it reduces the chances someone will touch something they shouldn't.  

```typescript
import { ValueGetter, Subscribable as ValueSubscriber } from 'rxjs'

interface Dispatcher {
  dispatch(action: { type: string, payload: any }): void
}

type Storer<T> = (
  ValueGetter<T> &
  ValueSubscriber<T> &
  Dispatcher
)

function doSomething(s: Storer<number>) {
  console.log(s.getValue())
  s.subscribe(console.log)
  s.dispatch({ type: 'Marco', payload: 'Polo' })
}
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -28,8 +28,7 @@ export { Subscriber } from './internal/Subscriber';
 export { Notification, NotificationKind } from './internal/Notification';
 
 /* Utils */
-export { pipe, Piper } from './internal/util/pipe';
-export { ValueGetter } from './internal/util/ValueGetter';
+export { pipe } from './internal/util/pipe';
 export { noop } from './internal/util/noop';
 export { identity } from './internal/util/identity';
 export { isObservable } from './internal/util/isObservable';
--- a/src/index.ts
+++ b/src/index.ts
@@ -28,7 +28,8 @@ export { Subscriber } from './internal/Subscriber';
 export { Notification, NotificationKind } from './internal/Notification';
 
 /* Utils */
-export { pipe } from './internal/util/pipe';
+export { pipe, Piper } from './internal/util/pipe';
+export { ValueGetter } from './internal/util/ValueGetter';
 export { noop } from './internal/util/noop';
 export { identity } from './internal/util/identity';
 export { isObservable } from './internal/util/isObservable';
--- a/src/internal/types.ts
+++ b/src/internal/types.ts
@@ -119,3 +119,23 @@ export type ObservedValuesFromArray<X> = X extends Array<ObservableInput<infer T
  * `ValueFromArray<T>` will return the actual type of `string`.
  */
 export type ValueFromArray<A> = A extends Array<infer T> ? T : never;
+
+export interface Pipable<T> {
+  /* tslint:disable:max-line-length */
+  pipe(): Observable<T>;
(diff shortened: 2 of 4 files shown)


kwonoj:
I do not believe core design of observable expects to encourage imperative value access, some subject types are special nature to expose those interfaces and I would disagree to expand it as general interfaces. I'll leave this open for another core team member to chime in opinon.

alshdavid:
I'd argue that rxjs is two projects which represent multiple paradigms. 

The first is a standard mechanism to handle reactive elements (streams, listeners, emitters), the basic types like `Observable`, `Subject`, etc. 

The second is a suite of extensions which are strung together to describe control flow declaratively.

This is also symbolised by the division of imports - `rxjs` `rxjs/operators`

Providing types which aid in the composability of the aforementioned basic types does not contradict the goal of the extensions. It does however serve to help in scenarios where people wish to build out their own custom basic types without the introduction of inheritance.

cartant:
I'm not convinced that it's up to this package to group types into a particular taxonomy.

TypeScript doesn't have nominal types (yet) - only the shape matters - so you can choose to define and group them in whichever way suits you and can publish them in a separate package.

Generally - I think I speak for the team here -  we would prefer to see ideas put forth in a userland package before being considered for inclusion in the core.

alshdavid:
What are the motivations behind the `Subscribable` and `Unsubscribable` interfaces located in `types.ts`?

kwonoj:
With this current proposal, core team has not reached agreement this should be inculded in core. Same as custome operator & other logics, we'd like to recommend to publish this as user level module and see its usage before we consider to discuss to include this in core.

alshdavid:
Another case for this is during testing.

both `Subject` and `Observable` conform to type `Subscribable & Pipeable`. Which means developers can apply dependency inversion on the type of stream-able being dependency injected.

Say during testing I am using a subject but in operation I am using an observable. Currently I have to convert the subject to an observable before accepting it.

PR #5138, opened 2019-11-11
fix(google): Subclassing TestScheduler needs RunHelpers

kolodny:
Until now, folks subclassing `TestScheduler` needed to `run(callback: (helpers: RunHelpers...` method. This was super brittle so we need to properly export this so it can be used more sanely


kolodny:
This PR is now ready for review

kolodny:
Yay, some mod somewhere in the bowels of google3 can now be removed

PR #5143, opened 2019-11-15
test(operator-concat): using run mode tests

ZackDeRose:
**Description:** 

Adjusts to run mode tests for concat operator.

### DIFF
--- a/spec/helpers/observableMatcher.ts
+++ b/spec/helpers/observableMatcher.ts
@@ -1,4 +1,5 @@
 import * as _ from 'lodash';
+import * as chai from 'chai';
 
 function stringify(x: any): string {
   return JSON.stringify(x, function (key: string, value: any) {
--- a/spec/operators/concat-spec.ts
+++ b/spec/operators/concat-spec.ts
@@ -18,30 +18,32 @@ describe('concat operator', () => {
 
   asDiagram('concat')('should concatenate two cold observables', () => {
     testScheduler.run(({ cold, expectObservable }) => {
-      const e1 =   cold('--a--b-|');
-      const e2 =   cold(       '--x---y--|');
-      const expected =  '--a--b---x---y--|';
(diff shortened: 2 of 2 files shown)


spec/helpers/observableMatcher.ts
```diff
@@ -1,4 +1,5 @@
 import * as _ from 'lodash';
+import * as chai from 'chai';
```
ZackDeRose: I'm not sure why this fixes the issue (or rather how other tests are working without this), but without this, all tests for the concat operator fail with `chai is not defined` error.

If there is a better solution, please let me know!

cartant: I was looking through some not-yet-merged PRs - 'cause other contributors have already converted some tests to use `run` and I wanted to make sure your efforts weren't going to be wasted, with the same tests converted twice - and I found a PR that makes this change: https://github.com/ReactiveX/rxjs/pull/5037 I've merged that, so you can rebase this PR.

spec/operators/concat-spec.ts
```diff
@@ -2,210 +2,245 @@ import { expect } from 'chai';
...
+  });
+
   asDiagram('concat')('should concatenate two cold observables', () => {
-    const e1 =   cold('--a--b-|');
-    const e2 =   cold(       '--x---y--|');
-    const expected =  '--a--b---x---y--|';
+    testScheduler.run(({ cold, expectObservable }) => {
+      const e1 =   cold('--a--b-|');
+      const e2 =   cold(       '--x---y--|');
+      const expected =  '--a--b---x---y--|';
```
cartant: When converting to `run`, we need to move the whitespace padding into the actual strings, so that we can use Prettier to format the source once the conversions are done. Prettier will collapse any whitespace outside of the string literals to a single space character and that will mess up the alignment.

Thanks for doing this, Zack.

PR #5151, opened 2019-11-18
docs(combineLatest): changed examples of code to not use deprecated code

vector67:
In a recent commit [https://github.com/ReactiveX/rxjs/commit/6661c796ea6f2918d4e6c0b68dfbecd4fbb7da74](https://github.com/ReactiveX/rxjs/commit/6661c796ea6f2918d4e6c0b68dfbecd4fbb7da74) a whole bunch of combineLatest calling forms were deprecated, but the docs weren't updated. I've now updated the docs for combineLatest to remove any references to the old way.

### DIFF
--- a/src/internal/observable/combineLatest.ts
+++ b/src/internal/observable/combineLatest.ts
@@ -111,37 +111,36 @@ export function combineLatest<R>(...observables: Array<ObservableInput<any> | ((
  *
  * ![](combineLatest.png)
  *
- * `combineLatest` combines the values from all the Observables passed as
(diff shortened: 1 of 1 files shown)


vector67:
I just realized that force pushing may not have been the right choice. Is that what I'm supposed to do on the forked repo?

cartant:
> Is that what I'm supposed to do on the forked repo?

Yep. That's fine. That's what's expected when you rebase and resolve the conflicts. Thanks.

PR #5153, opened 2019-11-19
fix(ReplaySubject): allow subclasses to override next method

kingcody:
Previously subclass `next` method would be overwriten by the ReplaySubject constructor.
This change also allows subclasses to call `super.next(value)` and
invoke the ReplaySubject `next` method instead of the Subject `next`
method.

### DIFF
--- a/spec/subjects/ReplaySubject-spec.ts
+++ b/spec/subjects/ReplaySubject-spec.ts
@@ -13,6 +13,38 @@ describe('ReplaySubject', () => {
     expect(subject).to.be.instanceof(Subject);
   });
 
+  describe('inheritance', () => {
+    it('should allow subclasses to override the next method', () => {
+      let wasMethodCalled = false;
+      class SubClassReplaySubject<T> extends ReplaySubject<T> {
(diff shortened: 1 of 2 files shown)


benlesh:
Adding this as an Agenda Item for the next team meeting, because we should probably discuss guidance around subclassing.

JWO719:
We roughly discussed that in an (incomplete) core team meeting:
 - from our point of view, we need to define some kind of recommendations regarding that topic
 - additional we need to think about proper placement within the docs (e.g. api docs)

gmfun:
@benlesh @kingcody is there any workaround available for this issue

benlesh:
@gmfun

As a workaround, you could overwrite it _again_ in the subclass constructor

cartant:
Closing this because the the implementation of `ReplaySubject` has changed and `next` is no longer assigned within the ctor:

https://github.com/ReactiveX/rxjs/blob/bc1b379c9cd0d79cf2178f8f27f23ec33875e20f/src/internal/ReplaySubject.ts#L47-L66

PR #5157, opened 2019-11-22
fix(startWith): fix incorrect types

thefliik:
**Description:**

I *think* (not sure) that the function definition of `startWith` is incorrectly typed. Because `startWith` has function overloads, this incorrect typing has no direct user impact. Except! this incorrect function signature is shown in the documentation for `startWith` (at the top), creating a source of confusion: https://rxjs-dev.firebaseapp.com/api/operators/startWith

This change can be thought of as a documentation fix, though it is also updates (fixes) the internal typing of the `startWith` operator function.

### DIFF
--- a/src/internal/operators/startWith.ts
+++ b/src/internal/operators/startWith.ts
@@ -65,13 +65,13 @@ export function startWith<T, D = T>(...array: Array<D | SchedulerLike>): Operato
  * @method startWith
  * @owner Observable
  */
-export function startWith<T, D>(...array: Array<T | SchedulerLike>): OperatorFunction<T, T | D> {
+export function startWith<T, D>(...array: Array<D | SchedulerLike>): OperatorFunction<T, T | D> {
   const scheduler = array[array.length - 1] as SchedulerLike;
   if (isScheduler(scheduler)) {
     // deprecated path
     array.pop();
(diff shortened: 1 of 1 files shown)


cartant:
Nope. IMO, it is correctly typed. The `D` refers to the type of the argument passed to the operator and the `T` refers to the type of the value received from the source.

thefliik:
@cartant what you said sounds correct, but I don't that `startWith` currently has that function signature... ?

Currently `startWith` looks like:

```ts
startWith<T, D>(...array: Array<T | SchedulerLike>)
```

But shouldn't it look like?

```ts
startWith<T, D>(...array: Array<D | SchedulerLike>)
```

After all,

> `D` refers to the type of the argument passed to the operator

thefliik:
As a point of comparison, the type signature of the `startWith` function itself appears to be incompatible with its second function overload.

i.e. this is the first function overload:

```ts
function startWith<T, D>(v1: D, scheduler: SchedulerLike): OperatorFunction<T, T | D>;
```

and this is the type signature of the function:

```ts
function startWith<T, D>(...array: Array<T | SchedulerLike>): OperatorFunction<T, T | D>
```

cartant:
Yeah. You're right. I read the diff backwards.

PR #5158, opened 2019-11-22
chore: use only tsc-generated files in tests

cartant:
**Description:**

Mocha is [configured to use `ts-node`](https://github.com/ReactiveX/rxjs/blob/d2f68db3a49079adffff52c5fbb950aa6fcb19ff/spec/support/default.opts), but it uses the JS test files that are written to the `dist` directory.

That seems at odds with using the [paths in the `tsconfig.json`](https://github.com/ReactiveX/rxjs/blob/d2f68db3a49079adffff52c5fbb950aa6fcb19ff/tsconfig.json) that reference the TypeScript files in the `src` directory.

<s>This PR changes the `test` script so that it consumes the TypeScript files, rather than the JavaScript files.</s>

This PR:

* removes `ts-node` from the Mocha configuration;
* removes `ts-node` as a dependency;
* uses a `spec/support/tsconfig.json` file to configure the `paths` to the `dist` files for `--require tsconfig-paths/register`; and
* fixes two imports that should have been relative paths.

With the removal of `ts-node`, the tests seem to compile faster. Thanks, @kwonoj !


kwonoj:
If I remember correctly I did it to benefit incremental build perf from tsc: ts-node currently doesn't use incremental build meaningfully, so each time test runs compiles everything again.

kwonoj:
still using ts-node is weird, I should've taken it off :/

cartant:
No worries. I'll remove `ts-node`, will put the compile back, and will change the `paths` to refer to the `dist` directory. The everything in the tests will be JavaScript from `dist`.

cartant:
@kwonoj I've updated the PR (and its description) to remove `ts-node`.

package.json
```diff
@@ -34,7 +34,7 @@
     "prepublishOnly": "npm run build:package",
     "postpublish": "./docs_app/scripts/publish-docs.sh",
     "publish_docs": "./publish_docs.sh",
-    "test": "npm run compile && mocha --opts spec/support/default.opts \"dist/spec/**/*-spec.js\"",
+    "test": "npm run compile && cross-env TS_NODE_PROJECT=spec/support/tsconfig.json mocha --opts spec/support/default.opts \"dist/spec/**/*-spec.js\"",
```
kwonoj: we removed dependency but specifying env var for ts-node?

spec/support/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
```
kwonoj: is this config needed? we have spec/tsconfig.json already

package.json
cartant: It's used to specify the aliases for the import paths `'rxjs/operators'`, etc. The `tsconfig-paths` `require` redirector needs it.

In the compile, the redirect need to be to the `src` directory, but here they need to be to the `dist/src` directory - because it's consuming JS.

We could use some other redirector - as that's all that's needed here - but I just left it using `tsconfig-paths`.

spec/support/tsconfig.json
cartant: Same answer as above. Paths are `dist/src` for the tests and only for the tests.

package.json
cartant: Happy to use something else, if you can suggest another redirection package.

kwonoj: Oh this reminds me reason I left ts-node, I used it for path mappings only. Need think little bit more - do we have some problem with current setup without this change?

cartant: Yeah, without the change, I think there is a problem. Essentially, I think we need two sets of path mappings:

* When `tsc` builds the projects the path mappings must point it to the TypeScript files in `src` - the specs use import paths like `'rxjs/operators'`.
* When Mocha runs the tests, it's running the JavaScript files within `dist/spec`, so another set of path mappings are needed and they have to point to the files in `dist/src`.

cartant: I removed `ts-node` because - with the change - it's not used. We only need the path mappings.

kwonoj: > set of path mappings are needed and they have to point to the files

yes, this is something I am aware and agree eventually need to be revisited. my q was do we have some noticable / practical failure due to having this changes.

cartant: There are no failures before the change and none after the change. The change doesn't fix anything that's broken. Instead, it makes sure that when the tests run, the run using the `tsc`-generated files that are in the `dist` directory.

Prior to the change, Mocha was running the JavaScript spec files in the `dist` directory, but was using `ts-node` to dynamically compile the source under test from the TypeScript files in the `src` directory. To me, this just seems so weird.

With the change, the tests and the source under test are loaded from the `dist` directory. AFAICT, this is faster - as you mentioned above "ts-node currently doesn't use incremental build meaningfully" - and seems much less strange.

I'm tempted to have a look for another path remapper, because the one that's used for the tests really has nothing to do with `ts-node` or with TypeScript.

kwonoj: yeah, sounds like having better path remapping seems right way to go with cutting ts-node properly.

cartant:
@kwonoj I've pushed a change that removes the `tsconfig-paths`-based path mapping and replaces it with a custom mapping that uses a mapping derived from the `paths` setting in the project's root `tsconfig.json` - so we only have to specify the mapping in one place.

cartant:
@kwonoj Added some comments.

kwonoj:
approved, feel free to check in anytime.

Issue #5042, opened 2019-09-27
Discussion: Hidden tuples to track Observable size and type


kolodny:
I'd like to suggest RxJS using hidden tuples to have stronger types which will also catch some logic errors at compile time

The general idea is that an `Observable` will have an internal tuple type of the elements of the Observable in situations when it's known. For example a call to  
`of(1, 2, 'test')` will produce `Observable<number|string, [number, number, string]>`  
while at the same time  
`let els = [1, 2, 'test']; of(...els)` will produce  
`Observable<number|string, (number|string)[]>`

This will allow things like:

```ts
of(1, 2, 3, 'foo', 'bar', false).pipe(take(2)).subscribe((thisIsOnlyANumber) => {
  // thisIsOnlyANumber is number and not number | string | boolean
})
```

It also empowers us to disallow invalid operators invocations like

```ts
of(1, 2).pipe(take(3))
                   ~ Argument of type '3' is not assignable to parameter of type '1 | 2'.
```

This technique would also give us a [`Single`](http://reactivex.io/documentation/single.html) for free since `fromFetch()`, `from(promise)`, and generated API could be set this hidden tuple to a single element. This can also remove the need for folks to do things like `api.fetchData().pipe(take(1))`.

I haven't quite worked out the correct typings yet but here's a rough approximation of what this would look like

```ts
type FromArray<T extends any[]> = T extends Array<infer U> ? U : never;

declare function of<T extends any[]>(...t: T): Observable<FromArray<T>, T>;

interface MonoTypeOperatorFunction<T, Elements extends T[] = T[]> extends OperatorFunction<T, T, Elements> {}
interface UnaryFunction<T, R> { (source: T): R; }
interface OperatorFunction<T, R, Elements extends T[] = T[]> extends UnaryFunction<Observable<T, Elements>, Observable<R>> {}

class Observable<T, Elements extends T[] = T[]> {
  subscribe(callback: (t: T) => void) {}

  pipe(): Observable<T>;
  pipe<A, AS extends A[]>(op1: OperatorFunction<T, A, Elements>): Observable<A, AS>;
  pipe<A, AS extends A[], B, BS extends B[]>(op1: OperatorFunction<T, A, Elements>, op2: OperatorFunction<A, B, AS>): Observable<B>;
  pipe<A, AS extends A[], B, C>(op1: OperatorFunction<T, A, Elements>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>): Observable<C>;
[...]

PR #5160, opened 2019-11-25
Hidden tuples Proof of Concept

kolodny:
!!DO NOT MERGE!!

This is a proof of concept for #5042 

This PR correctly catches the following cases

```ts
of(1, 2, 3).pipe(take(5)) // Error: Argument of type '5' is not assignable to parameter of type '0 | 1 | 2 | 3'.

const mapped = of(2, 'x', true, false).pipe(take(2), map(elementValue => elementValue.toString())) // `elementValue` has type `string | number`;
// mapped has type Observable<string, [string, string]>

interface User { name: string; }
declare let myApiResponse$: Observable<User>;
const firstResponseOnly = myApiResponse$.pipe(take(1)); // $ExpectType Observable<User, [User]>
```

This PR only wired up `of`, `from`, `map`, and `take` but can already catch a whole slew of bugs. Adding support for `isAsync`, `isClosed`, `isScalar` and other meta info that can help us catch logic errors at compile time can be added incrementally.

Getting dtslint working seems to be the wrong approach since it's too strict on how it does type matching. What this means is that making this work as is would require changing every single `// $ExpectType Observable<number>` to be `// $ExpectType Observable<number, number[]>` but if we were to add or change the type of the hidden type (for example to have an `isAsync` boolean), we would need to update every single instance every time.

Getting this working with tsd (or something similar) is going to be the bulk of the work since this means we need to completely revamp the TS type checks.

https://github.com/ReactiveX/rxjs/pull/5160/commits/84982c181cf648b2d064ae4f0d21d20b3b6e4d7d contains the main work needed to get this working, the rest of this PR diff is just moving around dtslint code to get TS typings to work


src/internal/types.ts
```diff
@@ -118,4 +118,123 @@ export type ObservedValuesFromArray<X> = X extends Array<ObservableInput<infer T
...
+  N extends  2 ? [T[0], T[1]] :
+  N extends  3 ? [T[0], T[1], T[2]] :
+  N extends  4 ? [T[0], T[1], T[2], T[3]] :
+  N extends  5 ? [T[0], T[1], T[2], T[3], T[4]] :
+  N extends  6 ? [T[0], T[1], T[2], T[3], T[4], T[5]] :
+  N extends  7 ? [T[0], T[1], T[2], T[3], T[4], T[5], T[6]] :
+  N extends  8 ? [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7]] :
+  N extends  9 ? [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7], T[8]] :
+  N extends 10 ? [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7], T[8], T[9]] :
+  T[number][];
```
benlesh: LMAO, I'm not saying this is wrong, but I am saying I'm lost with this TS usage. I'm not even really sure what this would match... `blah[number][]`? What does that match? `blah[1][2]`? wouldn't that be the same with `blah[][]`?

kolodny: Oh yeah, I can def make a tech talk on just these magic types.

Basically `tuple[number]` just creates a union of all the elements in the tuple

```ts
type MyTuple = [1, 2, 3];

let myThing: MyTuple[number];
myThing = 1;
myThing = 2;
myThing = 3;
// myThing = 4; Error
```

[demo](http://www.typescriptlang.org/play/?ssl=7&ssc=22&pln=1&pc=1#code/C4TwDgpgBAsiAqBXMAbaBeKBtAjAGigCYCBmAXQG4BYAKFrWCgFsEALASwDsBzALlgTI0WToiYAjCACdKtFvA48omHNRrzF3ZUTUauWzCTUB6Y8zb7tAFgpQAolKkB7KbVpA)

What these last cases are, are just the safe defaults to use when we can't do type magicking. In this case if we have an Observable of elements `[1, 2, 3, 4, boolean, '', 1, 2, 3, 4, 5, 6]`...

The type of this result would be `Observable<1|2|3|4|5|6|''|boolean, Array<1|2|3|4|5|6|''|boolean>>`

kolodny: Also https://github.com/kolodny/tuplizer/blob/master/index.test-d.ts has more examples of usage of these types

acutmore:
Cool idea. For the Rx projects I’ve been working on I can’t say I have ran into the bugs this would prevent* so not sure exactly how much value this approach adds. Maybe other projects would benefit more.

*these projects are almost entirely Observables of infinite length or length 1. In Java they solve this by having Observables and Singles.

src/internal/types.ts
TrejGun: Guys sorry but my eyes are bleeding so I can't understand the reason and limitation of this.

does it mean i can only emit 10 different types from observable?
like when I want to concat 10 streams into one? what will happen if i need 11?

kolodny: @TrejGun Not at all. These changes are completely backward compatible so anything you could do before this change you should be able to do after. The only thing this PR is trying to do is catch type errors on things that are probably incorrect like trying to take `6` elements on an Observable that will only ever emit 3 items. Even things like that are still possible but you need to tell the type system to ignore the number of elements in the Observable. I wish I had time to do a proper writeup the inner workings of this PR, how and why this works, how every `IsTuple<T> extends false ? ...` is a compile short circuit to speed up time the TS compiler spends on that long ternary chain, ...etc cc @benlesh on the last point

TrejGun: @kolodny thanks for explanation. it would be owesome to see longer doc someday

kolodny:
I just found out about https://pirix-gh.github.io/ts-toolbelt/, it looks like this would be a better type helper library to use than rolling our own. Any objections to using that instead of what I have in types.ts?

kwonoj:
In general we want to avoid to have dependencies. If we can work with devdependencies it would be ok.

benlesh:
Hey! I'm going to close this for now so it stops showing up in my reviews list. This was REALLY cool work. Definitely keep the branch for reference later.

PR #5167, opened 2019-11-30
chore(jsdoc): fix jsdoc comment

supperchong:
**Description:**


benlesh:
I think this change is stale now.

PR #5170, opened 2019-12-01
chore: remove useless type assertion

danranVm:
**Description:**
add generics to help type inference

### DIFF
--- a/src/internal/util/isPromise.ts
+++ b/src/internal/util/isPromise.ts
@@ -3,6 +3,6 @@
  * @see {@link https://www.ecma-international.org/ecma-262/6.0/#sec-promise-objects}
  * @param value the object to test
  */
-export function isPromise(value: any): value is PromiseLike<any> {
-  return !!value && typeof (<any>value).subscribe !== 'function' && typeof (value as any).then === 'function';
+export function isPromise<T>(value: any): value is PromiseLike<T> {
+  return !!value && typeof value.subscribe !== 'function' && typeof value.then === 'function';
 }
--- a/src/internal/util/isPromise.ts
+++ b/src/internal/util/isPromise.ts
@@ -3,6 +3,6 @@
  * @see {@link https://www.ecma-international.org/ecma-262/6.0/#sec-promise-objects}
  * @param value the object to test
  */
-export function isPromise<T = any>(value: any): value is PromiseLike<T> {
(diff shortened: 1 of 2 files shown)


src/internal/util/isPromise.ts
```diff
@@ -3,6 +3,6 @@
  * @see {@link https://www.ecma-international.org/ecma-262/6.0/#sec-promise-objects}
  * @param value the object to test
  */
-export function isPromise(value: any): value is PromiseLike<any> {
-  return !!value && typeof (<any>value).subscribe !== 'function' && typeof (value as any).then === 'function';
+export function isPromise<T = any>(value: any): value is PromiseLike<T> {
+  return !!value && typeof value.subscribe !== 'function' && typeof value.then === 'function';
```
cartant: IMO, this change is unnecessary and is incorrect. Essentially, it uses a type parameter as a type assertion and I think that should be avoid. The current types accurately describe what this user-defined type-guard function does.

src/internal/util/subscribeTo.ts
```diff
@@ -17,7 +17,7 @@ export const subscribeTo = <T>(result: ObservableInput<T>): (subscriber: Subscri
   } else if (isArrayLike(result)) {
     return subscribeToArray(result);
   } else if (isPromise(result)) {
-    return subscribeToPromise(result as Promise<any>);
+    return subscribeToPromise(result);
```
cartant: This change is fine, though. The type assertion in the current codebase is unnecessary and can be removed.

src/internal/util/isPromise.ts
danranVm: - this situation cannot be deduced correctly
image

- I hope so
image

cartant: If `isPromise` is used in that way in this package, the usage should be changed - `Promise<string> | any` is equivalent to `any`, so there is no inference that can be drawn from that type.

And `isPromise` is an internal function, so it should not use used outside of this package - it is not exported, AFAICT.

And the the snippet you have above illustrates my point. It isn't inferring anything; it's a type parameter acting as a type assertion. IMO, it is better to be explicit with a type assertion.

danranVm: Okay, I didn't notice `isPromise` is an internal function.

Issue #5171, opened 2019-12-01
`@deprecated` annotation used for `@internal` intent


edbzn:
Hi, 

I'm wondering if there is a good reason to use the `@depracated` doc annotation instead of `@internal` for internal code members usage. For example:

```ts
  /** @deprecated This is an internal implementation detail, do not use. */
  _subscribe(subscriber: Subscriber<T>) {
    return this.getSubject().subscribe(subscriber);
  }
```

Moreover using `@internal` in combination with:

```json
{
    "compilerOptions": {
        "stripInternal": true
    }
}
```
The annotated member is removed from generated `d.ts` file and consequently prevents package users from using internal implementation details.

cartant:
I remember asking this, but I cannot recall the answer. I thought it might have had something to do with the Closure compiler, but [that doesn't seem to be the case](https://github.com/google/closure-compiler/blob/faa896e0de117ece80e96d64c68dcacb7217da03/src/com/google/javascript/jscomp/parsing/Annotation.java#L22-L83).

edbzn:
I close the issue since v8 will solve this.

PR #5172, opened 2019-12-02
chore: use `@internal` annotation

edbzn:
Hi, 

This PR replaces the `@deprecated` annotation by `@internal` in combination with the `stripInternal` tsc option to remove internal members from generated `d.ts` file.

This will prevent TypeScript users from using internal implementation.

Related issue: #5171

### DIFF
--- a/src/internal/AsyncSubject.ts
+++ b/src/internal/AsyncSubject.ts
@@ -13,7 +13,7 @@ export class AsyncSubject<T> extends Subject<T> {
   private hasNext: boolean = false;
   private hasCompleted: boolean = false;
 
-  /** @deprecated This is an internal implementation detail, do not use. */
+  /** @internal This is an internal implementation detail, do not use. */
   _subscribe(subscriber: Subscriber<any>): Subscription {
     if (this.hasError) {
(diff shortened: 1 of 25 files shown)


Edouardbozon:
It seems that `stripInternal` option causes many problems and was undocumented https://github.com/Microsoft/TypeScript-Handbook/pull/856 :confused:

cartant:
If these are marked as `internal`, is there any warning given by TypeScript if they are used? If not and a lint rule is needed, I would suspect that such a rule - if it exists - would not be as widely used as the depreciation rule.

Edouardbozon:
There is no warning directly given by TypeScript, the only protection I know is a visual feedback  like the `@deprecated` annotation does in IDEs like Webstorm.

benlesh:
I seem to recall these are marked as "deprecated" rather than "internal" because of some issue with Angular builds. But I can't remember what that issue was. Either way, we're going to be removing these sorts of things in v8, so I'm not sure that this change is worth proceeding with.

edbzn:
@benlesh Understood, I close the PR.

Issue #5188, opened 2019-12-20
TypeScript: Pluck has a type safety detonator in it


Lonelind:
## Bug Report

**Current Behavior with reproduction example**
The `pluck` operator is working poorly with some cases in terms of type safety.

We use a rig to generate props for React components from Rx Observables. The function uses `Partial` of the Component Props type and expect this type to be returned. In general we have something like this:
```tsx
interface MyPropType = {
    id: number;
}
interface ComponentProps = {
    myProp: MyPropType;
}

const Component: React.FC<ComponentProps> = ({ myProp }) => (
    <div>{myProp.id}</div>
);

// It's the general idea, not the actual implementation
const Container = withRx(Component)(({ myStream$ }) => ({
    myProp: myStream$.pipe(pluck('someField')),
}));
```
So, it's pretty fine with that in terms of type safety. Let's assume we have the following in `myStream$`:

```ts
interface MyData = {
    anyField: { id: number };
}

const myStream$: Observable<MyData> = /* ... */
```
Type inference for `pluck` will work in an unexpected way. If you look carefully, you'll see that if `someField` is absent from the expected object type than `keyof` can't be inferred and we have a general overload that gets some `R` from what it is expected to be:

```ts
// pluck.d.ts
/* ... */
export declare function pluck<T, R>(...properties: string[]): OperatorFunction<T, R>;
```
The problem in this case is that we have a type for `myProp` field and the `R` is inferred from it. So, we have a false expectation that there should be `MyPropType` returned from `pluck`. But suddenly, there is an `undefined` there, in runtime.

**Expected behavior**
It's expected to have any kind of type check error if we pass a key that's absent in the object type we have in the `pluck` input regardless of the expectations.

**Environment**
- TypeScript version: 3.5.2
- RxJS version: 6.5.2

cartant:
Yeah, this could be better. It's inferred as `unknown`:

```ts
const source = of({ foo: "bar" });
const plucked = source.pipe(pluck("goo")); // Observable<unknown>
```

Because, as you've pointed out, it matches the catch-all, rest parameter signature - but it should effect a compile-time error. The catch-all should include a rest parameter *after* the other parameters - as in [this `pipe` signature](https://github.com/ReactiveX/rxjs/blob/b90b397b0ab3250ac56dd22de5e2d4ed69fdf57f/src/internal/Observable.ts#L318).

Lonelind:
And if you are more specific, it has specific type, not the `unknown`.

[...]

PR #5192, opened 2019-12-23
fix(pluck): fix pluck's catch-all signature for better type safety

Lonelind:
**Description:** Change pluck's catch-all signature that is causing troubles when the
passed string is not a key in the argument object.

**Related issue (if exists):** https://github.com/ReactiveX/rxjs/issues/5188

### DIFF
--- a/spec-dtslint/operators/pluck-spec.ts
+++ b/spec-dtslint/operators/pluck-spec.ts
@@ -1,4 +1,4 @@
-import { of } from 'rxjs';
+import { of, Observable } from 'rxjs';
 import { pluck } from 'rxjs/operators';
 
 it('should infer correctly', () => {
@@ -39,6 +39,10 @@ it('should not accept empty parameter', () => {
 
 it('should accept string only', () => {
   const a = of({ name: 'abc' }).pipe(pluck(1)); // $ExpectError
+});
+
+it('should not infer type from the variable if key doesn\'t exist', () => {
+  const a: Observable<number> = of({ name: 'abc' }).pipe(pluck('xyz')); // $ExpectError
 });
 
 it('should accept a spread of arguments', () => {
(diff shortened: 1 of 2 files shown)


Lonelind:
Just happened to see that there is a line that was added 7 months ago that does virtually the same but with one core difference: if you use `R` there, it will fail the same way as it was already but it will happen with the seventh argument passed. The `pluck` operator will resolve type by itself returning what was specified in result (it will fall down to `unknown` only if it was `any` there). Passing `unknown` every time is safer. You will need to check the type, yes but it will not say that everything is okay even if it isn't.

Lonelind:
Also, I've added a test case that checks type inference:

```ts
it('should not infer type from the variable if key doesn\'t exist', () => {
  const a: Observable<number> = of({ name: 'abc' }).pipe(pluck('xyz')); // $ExpectError
});
```

With the old implementation there's no error emitting and the `pluck` signature gets `number` as the `R`. This fix will returns `unknown` there, and the types don't match.

spec-dtslint/operators/pluck-spec.ts
```diff
@@ -41,6 +41,10 @@ it('should accept string only', () => {
   const a = of({ name: 'abc' }).pipe(pluck(1)); // $ExpectError
 });
 
+it('should not infer type from the variable if key doesn\'t exist', () => {
+  const a: Observable<number> = of({ name: 'abc' }).pipe(pluck('xyz')); // $ExpectError
```
gioragutt:
```suggestion
  const a: Observable<never> = of({ name: 'abc' }).pipe(pluck('xyz'));
```

Wouldn't that be a better test?

Lonelind: No, this test is meant to show that predefined type doesn't match to the `unknown` that is calculated from the `pluck` when the key doesn't exist.

In current implementation this test will fail. The `pluck('xyz')` return type is `number` and the key doesn't exist. It should be `unknown` but it's inferring from the predefined type of the variable/constant.

Lonelind: This test should be a positive failure, and the correction you are suggesting has already been checked in `should accept existing keys only` test case.

Consider this example:

```ts
interface SomeObject {
    a: string;
}

interface Input {
    b: SomeObject;
}

interface Output {
    b: Observable<SomeObject>;
}

const stream$ = of<Input>({
    b: { a: 'abc' },
});

const positiveCase$ = stream$.pipe(pluck('b')); // will be of type Observable<SomeObject>

const negativeCase$ = stream$.pipe(pluck('xyz')); // should return Observable<unknown> but the current implementation will infer the type and the type will be Observable<SomeObject>

const positiveOutput: Output = {
    b: positiveCase$, // no error here
}

const negativeOutput: Output = {
    b: negativeCase$, // will throw type error with my implementation and will be silent using current one
}
```

benlesh:
This can't go into patch, because it will break people's builds if they're violating the new constraint. However, it's a good addition to master.

PR #5199, opened 2019-12-27
interval to timer

xp44mm:
bufferWhen(), instead of taking an observable to trigger the start of each new buffer, accepts a closing selector method that's re-invoked every time the buffer is closed, and the resulting observable is used to determine when the next buffer should close.

**Description:**

### DIFF
--- a/src/internal/operators/bufferWhen.ts
+++ b/src/internal/operators/bufferWhen.ts
@@ -26,12 +26,12 @@ import { OperatorFunction } from '../types';
  * Emit an array of the last clicks every [1-5] random seconds
  *
  * ```ts
- * import { fromEvent, interval } from 'rxjs';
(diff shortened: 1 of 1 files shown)


Qquanwei:
I think that `interval` and `timer` do the same thing.  there no difference

gioragutt:
@Qquanwei They are similar, but don't do the same thing.

`timer`, without a second parameter, will only invoke once.
`interval` will invoke indefinitely.

xp44mm:
the code before modified same as after modified, but `timer` better than `interval` to emphisis `bufferWhen()` **re-invoked** every time the buffer is closed,

niklas-wortmann:
To clean the pr list up a bit, I will close this issue, if you want to update your changes, feel free to submit a new pr

PR #5211, opened 2020-01-04
docs(operators): add note about first() vs take(1)

fmalcher:
**Description:**
In the docs for the `first` operator add a note that `first()` is different from `take(1)`. First throws `EmptyError` if there are no elements in the source.

see also https://twitter.com/fmalcher01/status/1213417124036784128


martinsik:
This seems to be a common problem :) https://stackoverflow.com/questions/42345969/take1-vs-first

JWO719:
@fmalcher could you maybe rebase master into that branch. I hope CI is working afterwards, otherwise I'll investigate

fmalcher:
@JWO719 Done 👍

PR #5212, opened 2020-01-05
Expect observable

xp44mm:
write expected marbles is difficultly. tell whether result of pipeline is right is simpler than the figure out it's result. so suggest add `actual` prop that is testmessage[] into expectedObservable() function's return object.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,5 @@ npm-debug.log
 # The check-side-effects package generates and deletes this file.
 # If the process is killed, it will be left behind.
 check-side-effects.tmp-input.js
+/.vs/rxjs/config/applicationhost.config
+/.vs
--- a/src/internal/testing/TestScheduler.ts
+++ b/src/internal/testing/TestScheduler.ts
@@ -121,7 +121,7 @@ export class TestScheduler extends VirtualTimeScheduler {
   }
 
   expectObservable(observable: Observable<any>,
-    subscriptionMarbles: string = null): ({ toBe: observableToBeFn, actual: TestMessage[] }) {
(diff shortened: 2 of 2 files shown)


xp44mm:
i want to get expected `testmessage[]` from `expectedObservable()` function. and it just is `actual`. it is useful to write expected marbles.

cartant:
AFAICT, `TestMessage` is a private type. IMO, we should not be exposing this information, so I'm t in favour of the change in this PR.

xp44mm:
@cartant yes! may i get some similar `MarbleObservable` class to provide me with marble or messages?

cartant:
No, the representation is private. If it were public, we would have to support it and would not be able to change it.

The representation is passed to the assertion function:

https://github.com/ReactiveX/rxjs/blob/56e7cebd211d93b9958e1ec508052beb8927604a/src/internal/testing/TestScheduler.ts#L65-L67

But it is passed as `any` and it's internal representation is private and could be changed at any time. The assertion function only gets to decide whether the two representations are deep-equal. It does not get to inspect a documented - and supported - data structure.

Other tools - like https://github.com/kwonoj/rx-sandbox - that perform marble tests have - AFAICT - there own internal representations, they don't utilize a data structure that's exposed by this package.

xp44mm:
@cartant actually, i just want to get the marble, the string, for example `'---a---b---|'`. it is don't need public TestMessage type.

cartant:
The implementation does not - and probably cannot, AFAICT - render the actual result as a marble string. That's just not how it works.

niklas-wortmann:
To clean the pr list up a bit, I will close this issue, if you want to update your changes, feel free to submit a new pr

Issue #5078, opened 2019-10-15
Migrate to eslint


kwonoj:
TSlint is now planned to deprecate (https://github.com/palantir/tslint/issues/4534) and we'd like to migrate our lint config to use eslint instead.

We want

- Preserve existing rules
- Apply additional rules if necessary
- If there's no corresponding rules, audit & decide what to do.

cartant:
This resource might be useful for locating equivalent rules: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md

lambainsaan:
I would like to take it up. :)

kwonoj:
If anyone interested in this, feel free to take up. PR's welcome.

Jefiozie:
Hi @kwonoj and @cartant, I made a [branch](https://github.com/Jefiozie/rxjs/commit/a9b64ad23b3950b39b46bcc15d5e65e2daa4e070
) where I started the work needed for this. Could you have a look at it and help me with the rules? If I run the linting rules atm it will change +/- 120 files based on this ruleset. This is because of spacing and line lengths.


Looking forward on your reply.

benlesh:
Hi @kwonoj — thank you for taking the time to open this issue and for sharing your thoughtful feedback with the RxJS project. I’m sorry it took us so long to respond. Given how much time has passed and the lack of recent activity, I’m going to close this for now so the tracker reflects the work we’re likely to act on. If this is still relevant on a current RxJS release, please feel welcome to open a new issue with updated context and a reproduction. Thank you again for your contribution and patience.

— Ben

PR #5222, opened 2020-01-07
fix: add migration for tslint to eslint

Jefiozie:
**Description:**
Migration from tslint to eslint.

This is the first step to migrate to eslint, I've not yet removed tslint but will do this after some feedback. I've used the typescript-eslint to preserve existing rules.

Looking forward on your feedback.

#5078 

p.s. probably some CI steps will fail with this change.

cc: @kwonoj @cartant


package.json
```diff
@@ -27,8 +27,10 @@
     "commitmsg": "validate-commit-msg",
     "build:spec:browser": "echo \"Browser test is not working currently\" && exit -1 && webpack --config spec/support/webpack.mocha.config.js",
     "decision_tree_widget": "cd doc/decision-tree-widget && npm run build && cd ../..",
-    "lint_spec": "tslint -c tslint.json \"spec/**/*.ts\"",
-    "lint_src": "tslint -c tslint.json \"src/**/*.ts\"",
+    "lint-spec": "tslint -c tslint.json \"spec/**/*.ts\"",
+    "lint_eslint_spec": "eslint --ext=ts spec/**/*  -c .eslintrc",
```
kwonoj: I'd like to suggest to follow `:` for script convention (using `-` was old way we did), like `lint:eslint:spec`.

btw, maybe with eslint we can just make single lit to run all spec / src at once? I think it was splitted as tslint glob behaves weird on windows.

tslint-to-eslint-config.log
```diff
@@ -0,0 +1,6 @@
+import-spacing does not yet have an ESLint equivalent.
```
kwonoj: do we need log file to be checked in?

package.json
Jefiozie: Actually the tslint run scripts will be removed so I will do this in my next commit to give a good overview on what the change is.

tslint-to-eslint-config.log
Jefiozie: Nope, good catch

package.json
kwonoj: Comment was for new script `eslint`.

Jefiozie:
So I made the changes that where mentioned in the comments. I have two questions remaining the migration.

1. Should we change the dtslint rule set also?
2. The lint rule fails atm however the settings are the same. Would you like me to fix them once by running the fix command?

benlesh:
@kwonoj @cartant where are we at with this one? _I_ think it looks good, but I don't have a ton of experience with eslint.

Jefiozie:
Hi @carant, any updates on this PR?

cartant:
> Hi @cartant, any updates on this PR?

Sorry. I've not had time to look over it; I've had other priorities. Which is not to say that this isn't appreciated. It is and thanks for the time you've put into it. I'll bump it up a bit in my backlog and will try to make some time for it on the weekend.

Jefiozie:
>Sorry. I've not had time to look over it; I've had other priorities. Which is not to say that this isn't appreciated. It is and thanks for the time you've put into it. I'll bump it up a bit in my backlog and will try to make some time for it on the weekend.

HI @cartant, thanks for your reply. Take your time I was just curious what the status was of this PR and If some of the team already had a look.

Jefiozie:
I'm closing im going to close this one and look if I can retry it with the new changes.

PR #5227, opened 2020-01-10
Removed Extra Filters

parasharrajat:
As this document is inspired by Angular docs. There were some extra types in the filters So I think they are not necessary.

**Description:**


dzhavat:
Why not deleting them instead?

parasharrajat:
@dzhavat This is my first PR . So I was afraid to do so.

PR #5229, opened 2020-01-11
docs(fetch): example code return different type

xp44mm:
if and else return different type's data, whether is this a bug?

**Description:**


kwonoj:
switchmap requires to return observablelike, `response.json` returns `Promise<T>` (https://developer.mozilla.org/en-US/docs/Web/API/Body/json) and returning observable for other execution path is correct.

PR #5231, opened 2020-01-14
docs(index): fix docs Rx.Scheduler.async is deprecated

luckyscript:
**Description:**
fix the docs error:
Rx.Scheduler.async is deprecated, change it to Rx.asyncScheduler

### DIFF
--- a/src/internal/observable/bindCallback.ts
+++ b/src/internal/observable/bindCallback.ts
@@ -144,7 +144,7 @@ export function bindCallback(callbackFunc: Function, scheduler?: SchedulerLike):
  * }
  *
  * const boundSyncFn = bindCallback(iCallMyCallbackSynchronously);
- * const boundAsyncFn = bindCallback(iCallMyCallbackSynchronously, null, Rx.Scheduler.async);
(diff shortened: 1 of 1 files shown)


src/internal/observable/bindCallback.ts
```diff
@@ -144,7 +144,7 @@ export function bindCallback(callbackFunc: Function, scheduler?: SchedulerLike):
  * }
  *
  * const boundSyncFn = bindCallback(iCallMyCallbackSynchronously);
- * const boundAsyncFn = bindCallback(iCallMyCallbackSynchronously, null, Rx.Scheduler.async);
+ * const boundAsyncFn = bindCallback(iCallMyCallbackSynchronously, null, Rx.asyncScheduler);
```
cartant: The `Rx` should be removed and `asyncScheduler` should to be added to the import:

```ts
import { bindCallback, asyncScheduler } from 'rxjs';
```

benlesh: I did it myself. LGTM otherwise.

Issue #5237, opened 2020-01-16
finalize not working as expected with streams from dependencies


cdanielw:
## Bug Report

**Current Behavior**
Given project `a` and `b`. When `a` subscribe to a stream from `b`, the `finalize` operator doesn't get called.

The below script outputs:
```
a finalized c
```

**Reproduction**
https://github.com/cdanielw/rxjs-finalize

```js
const {finalize, first} = require('rxjs/operators')
const anotherModule$ = require('b')
const sameModule$ = require('./c')

anotherModule$.pipe(
        // Stream from another module doesn't finalize
	finalize(() => console.log('a finalized b')),
	first()
).subscribe()

sameModule$.pipe(
        // Stream from same module finalize
	finalize(() => console.log('a finalized c')),
	first()
).subscribe()

```

**Expected behavior**
Script should output :
```
a finalized b
a finalized c
```

**Environment**
- Runtime: Node v10.14.2
- RxJS version: 6.5.4

cdanielw:
I managed to workaround this by this:

```js
of(true).pipe(
    switchMap(() => anotherModule$),
    finalize(() => console.log('1. a finalized b')),
    first()
).subscribe()
```

But it's easy to forget to do this, leading to subtle bugs.

cartant:
`from(anotherModule$)` should be used. I'll write an explanation when I'm less tired. Not a bug, IMO.

cdanielw:
Having different code depending on how/where code has been required makes it very difficult to pull RxJS code into shared libraries. This is where my issue is coming from. 

Even if this wouldn't be considered a bug, I'd really like you to consider removing any differences in behaviour based on where streams/code creating streams have been required from.

cartant:
Yeah. I was wrong; this is definitely a bug.

I guess I really was tired, last night, as I recently fixed a problem that was closely related to this - see https://github.com/ReactiveX/rxjs/pull/5059 and the associated cherry pick into v6.

That PR includes testing infrastructure that'll allow me to reproduce your problem in a failing test. After which, I'll fix it.

Thanks for opening the issue.

PR #5239, opened 2020-01-17
fix: chain subscriptions for interop with finalize

cartant:
**Description:**

The change in this PR is very similar to the changes that were made in #5059: when an interop subscriber is created within `finalize`, it'll be wrapped when a subscription is made to the source observable in `FinallyOperator#call`. The interop subscriber needs to be added to the subscription so that unsubscription chains property and the finalize callback is invoked.

The adding of the interop subscription is now handled in `subscribeWith` - which was added in https://github.com/ReactiveX/rxjs/pull/5333.

This PR also includes a failing test that is fixed by the above change.

**Related issue (if exists):** #5237

### DIFF
--- a/src/internal/operators/finalize.ts
+++ b/src/internal/operators/finalize.ts
@@ -68,7 +68,15 @@ class FinallyOperator<T> implements Operator<T, T> {
   }
 
   call(subscriber: Subscriber<T>, source: any): TeardownLogic {
-    return source.subscribe(new FinallySubscriber(subscriber, this.callback));
+    // The returned subscription will usually be the FinallySubscriber.
+    // However, interop subscribers will be wrapped and for
+    // unsubscriptions to chain correctly, the wrapper needs to be added, too.
+    const finallySubscriber = new FinallySubscriber(subscriber, this.callback);
+    const subscription = source.subscribe(finallySubscriber);
+    if (subscription !== finallySubscriber) {
+      subscription.add(finallySubscriber);
+    }
+    return subscription;
   }
 }
 
--- a/src/internal/operators/finalize.ts
+++ b/src/internal/operators/finalize.ts
@@ -3,6 +3,7 @@ import { Subscriber } from '../Subscriber';
 import { Subscription } from '../Subscription';
 import { Observable } from '../Observable';
 import { MonoTypeOperatorFunction, TeardownLogic } from '../types';
+import { subscribeWith } from '../util/subscribeWith';
 
 /**
  * Returns an Observable that mirrors the source Observable, but will call a specified function when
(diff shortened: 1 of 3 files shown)


src/internal/operators/finalize.ts
```diff
@@ -69,7 +69,15 @@ class FinallyOperator<T> implements Operator<T, T> {
   }
 
   call(subscriber: Subscriber<T>, source: any): TeardownLogic {
-    return source.subscribe(new FinallySubscriber(subscriber, this.callback));
+    // The returned subscription will usually be the FinallySubscriber.
+    // However, interop subscribers will be wrapped and for
+    // unsubscriptions to chain correctly, the wrapper needs to be added, too.
+    const finallySubscriber = new FinallySubscriber(subscriber, this.callback);
+    const subscription = source.subscribe(finallySubscriber);
+    if (subscription !== finallySubscriber) {
```
benlesh: I suppose this is to get around edge cases with node where people have more than one version of RxJS?

cartant: Yep, but the versions can be the same - just different modules. In the related issue, operators from one RxJS module were used with observables from another.

There is another PR that I opened - for discussion - as this is pretty much a general problem: https://github.com/ReactiveX/rxjs/pull/5243

There are a bunch of operators that need `unsubscribe` chained.

benlesh: This is highly debatable. On one hand, I'd like to be ergonomic... On the other hand, it feels like we're just enabling people to build apps with garbage dependency graphs.

On yet ANOTHER hand, Observable is a primitive, like a promise, and people should feel free to export this primitive from their libraries. Ahhh...

cartant: It's a real problem with Node. The issues opened for this sort of thing are almost always Node-related.

benlesh: Is this something we should just be doing in `Observable` subscribe?

cartant: I did look at that, but, no. If it's done in `Observable#subscribe`, some tests will fail. I can have a closer look at that if you want an explanation.

cartant:
> Is this something we should just be doing in `Observable` subscribe?

@benlesh I had a quick look at this and the issue with putting this functionality into `subscribe` is that there are some subscribers upon which `unsubscribe` should not be chained: subjects. Calling `unsubscribe` on a subject closes/stops it. And if the change in this PR is added to `subscribe`, a bunch of `multicast` and `multicast`-related tests fail.

cartant:
Closed by https://github.com/ReactiveX/rxjs/pull/5333#pullrequestreview-398628074

cartant:
Nope, https://github.com/ReactiveX/rxjs/pull/5333 doesn't close this. Rather, this PR needs to be refactored to use `subscribeWith`.

PR #5243, opened 2020-01-18
[WIP] fix: ensure all interop operators are chained correctly

cartant:
**WIP:**

https://github.com/ReactiveX/rxjs/pull/5333 is likely a better solution to the problem.

---

**Description:**

This PR refactors the fix in #5239 so that it can be applied to all operators - or at least to those that need unsubscription to be chained.

The problem is outlined in the linked issue, but AFAICT the problem is not specific to the `finalize` operator. For example, the `buffer` operator's subscriber adds a subscription that depends upon unsubscription being chained correctly:

https://github.com/ReactiveX/rxjs/blob/56e7cebd211d93b9958e1ec508052beb8927604a/src/internal/operators/buffer.ts#L73-L76

And `windowTime` adds scheduled actions:

https://github.com/ReactiveX/rxjs/blob/56e7cebd211d93b9958e1ec508052beb8927604a/src/internal/operators/windowTime.ts#L189-L206

IMO, all operator subscriptions made within an operator's `call` method should be made using `subscribeAndChainOperator`.

ATM, this PR's branch is based on #5239, so it includes that PR's commits. The last two commits are the changes introduced in this PR.

**Related issue (if exists):** #5239

### DIFF
--- a/spec/helpers/interop-helper.ts
+++ b/spec/helpers/interop-helper.ts
@@ -1,4 +1,4 @@
-import { Observable, Subscriber, Subscription } from 'rxjs';
+import { Observable, Operator, Subscriber, Subscription } from 'rxjs';
 import { rxSubscriber as symbolSubscriber } from 'rxjs/internal/symbol/rxSubscriber';
 
 /**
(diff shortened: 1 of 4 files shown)


cartant:
Closed by https://github.com/ReactiveX/rxjs/pull/5333

Issue #5240, opened 2020-01-17
rxjs is not compatible with `strictBindCallApply`


mprobst:
## Bug Report

**Current Behavior**

If you change rxjs' compiler options to include the recommended `strictBindCallApply` (i.e. get closer to work with just `--strict`), rxjs fails to compile. See the error messages included in the patch below for a possible fix.

**Reproduction**
Turn on `strictBindCallApply`.

**Expected behavior**

Compiles.

**Possible Solution**

```
==== rxjs/src/internal/observable/fromEvent.ts#4 - new/rxjs/src/internal/observable/fromEvent.ts ====
--- old/rxjs/src/internal/observable/fromEvent.ts	2019-09-20 20:35:42.000000000 +0200
+++ new/rxjs/src/internal/observable/fromEvent.ts	2020-01-10 20:22:01.000000000 +0100
@@ -214,7 +214,11 @@
   return new Observable<T>(subscriber => {
     function handler(e: T) {
       if (arguments.length > 1) {
-        subscriber.next(Array.prototype.slice.call(arguments));
+        // TODO: Argument of type 'any[]' is not assignable to
+        // parameter of type 'T'.
+        subscriber.next(
+            Array.prototype.slice.call(arguments) as
+            any);
       } else {
         subscriber.next(e);
       }
==== rxjs/src/internal/operators/combineLatest.ts#4 - new/rxjs/src/internal/operators/combineLatest.ts ====
--- old/rxjs/src/internal/operators/combineLatest.ts	2019-08-20 17:44:12.000000000 +0200
+++ new/rxjs/src/internal/operators/combineLatest.ts	2020-01-10 20:22:01.000000000 +0100
@@ -73,5 +73,10 @@
     observables = (<any>observables[0]).slice();
   }
 
-  return (source: Observable<T>) => source.lift.call(from([source, ...observables]), new CombineLatestOperator(project));
+  // TODO: Type '(source: Observable<T>) => Observable<unknown>'
+  // is not assignable to type 'OperatorFunction<T, R>'.
+  return ((source: Observable<T>) => source.lift.call(
+              from([source, ...observables]),
+              new CombineLatestOperator(project))) as
+      any;
 }
==== rxjs/src/internal/operators/concat.ts#6 - new/rxjs/src/internal/operators/concat.ts ====
--- old/rxjs/src/internal/operators/concat.ts	2019-11-11 21:57:36.000000000 +0100
[...]

PR #5244, opened 2020-01-18
chore: enable fully strict TypeScript option

cartant:
**Description:**

This PR enables the currently disabled sub-options under TypeScript's `strict` option. It's based on the branch from the #5241 PR. I'll rebase this once that PR is merged.

The changes in this PR are almost entirely type-related. There were only a few places in which I had to add `if` statements so that calls to wrapped operators were mapped to the appropriate signatures.

There are a lot of files, but the changes are small, so hopefully reviewing this will not be too painful. In any case, these aren't changes that can be done in a piecemeal fashion.

These changes are not ideal, as that was not my aim. I was primarily concerned with getting the `strict` option enabled without having to add additional logic - `undefined` or `null` guards, etc. Changes to make code more easily typechecked can be made when things are refactored/reimplemented in v8.

**Related issue (if exists):** #5240

### DIFF
--- a/spec-dtslint/tsconfig.json
+++ b/spec-dtslint/tsconfig.json
@@ -4,8 +4,6 @@
     "lib": ["es2015", "dom"],
     "module": "commonjs",
     "noEmit": true,
-    "noImplicitAny": true,
-    "noImplicitThis": true,
     "paths": {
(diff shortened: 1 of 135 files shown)


benlesh:
LGTM... but there are merge conflicts, @cartant

PR #5248, opened 2020-01-21
chore: disable suppressImplicitAnyIndexErrors

cartant:
**Description:**

This PR disables the `suppressImplicitAnyIndexErrors` TypeScript option - which is another escape hatch for non-strict code.

**Related issue (if exists):** #5240 (somewhat related)


benlesh:
nice!

PR #5249, opened 2020-01-21
add zipWith

benlesh:
- updates tests
- updates docs
- updates deprecation messages
- moves tests for legacy selector behaviors to their own file

PR #5254, opened 2020-01-22
feat: add ObservedValueUnion/TupleFromArray

cartant:
**Description:**

This PR:

* Renames the `ObservedValuesFromArray` type to `ObservedValueUnionFromArray` and deprecates the former.
* Adds the `ObservedValueTupleFromArray` type.

The differences between the two are best shown by example:

```ts
type S = Observable<string>;
type N = Observable<number>;
ObservedValueUnionFromArray<[S, N]> // string | number
ObservedValueTupleFromArray<[S, N]> // [string, number]
```

It also adds an `Unshift` type - which adds a type to the beginning of a tuple - that should be useful in typing the `combineLatestWith` and `zipWith` operators.

These types should be sufficient to allow for N-args support in those operators:

```ts
export function zipWith<T, A extends ObservableInput<any>[]>(
  ...otherSources: A
): OperatorFunction<T, Unshift<ObservedValueTupleFromArray<A>, T>>;
```

**Related <s>issue</s> PRs:** #5249; #5251

### DIFF
--- a/src/internal/types.ts
+++ b/src/internal/types.ts
@@ -108,12 +108,40 @@ export interface SchedulerAction<T> extends Subscription {
 export type ObservedValueOf<O> = O extends ObservableInput<infer T> ? T : never;
 
 /**
- * Extracts the generic type from an `ObservableInput<any>[]`.
+ * Extracts a union of element types from an `ObservableInput<any>[]`.
  * If you have `O extends ObservableInput<any>[]` and you pass in
  * `Observable<string>[]` or `Promise<string>[]` you would get
- * back a type of `string`
+ * back a type of `string`.
+ * If you pass in `[Observable<string>, Observable<number>]` you would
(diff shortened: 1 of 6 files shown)


src/internal/types.ts
```diff
@@ -106,12 +106,40 @@ export interface SchedulerAction<T> extends Subscription {
...
+export type ObservedValueTupleFromArray<X> =
+  X extends Array<ObservableInput<any>>
+    ? { [K in keyof X]: ObservedValueOf<X[K]> }
+    : never;
+
+/**
+ * Adds a type to the beginning of a tuple.
+ * If you pass in `Unshift<[B, C], A>` you will get back `[A, B, C]`.
+ */
+export type Unshift<X extends any[], Y> =
```
benlesh: Nice! I wouldn't have thought to use extending a function to pull out the argument types. Very cool

PR #5256, opened 2020-01-22
chore: fix build errors

benlesh:
I'm getting build errors since our updates to strictness. They revolved around the call signatures of SchedulerLike and SchedulerAction in a few operators. This resolves that by being a little more explicit.


benlesh:
It was in this package. After I cleared `node_modules` and rebuilt locally I was getting errors (valid ones) in operators that used scheduling like `windowTime`.

PR #5258, opened 2020-01-23
docs(defer): mention analogy to function calls

robstoll:
... and that the creation behaviour is important in case of Subject creation

**Description:**

### DIFF
--- a/src/internal/observable/defer.ts
+++ b/src/internal/observable/defer.ts
@@ -19,7 +19,10 @@ import { empty } from './empty';
  * typically with an Observable factory function. It does this afresh for each
  * subscriber, so although each subscriber may think it is subscribing to the
  * same Observable, in fact each subscriber gets its own individual
- * Observable.
+ * Observable. This is not different to other creation operators -- i.e. 
+ * a `subscribe` call can be seen as a function call -- but plays an 
+ * important role when one wants to return a Subject (or a subtype of Subject)
+ * as they are then not shared amongst subscribers neither.
  *
  * ## Example
  * ### Subscribe to either an Observable of clicks or an Observable of interval, at random


robstoll:
Thanks for your quick review. The problem I see with
> so although each subscriber may think it is subscribing to the same Observable, in fact each subscriber gets its own individual Observable.

is that it sounds a bit like other creation operators have a shared behaviour, i.e. that they behave like EventEmitters even though that's not the default case. Do you see what I mean?

kwonoj:
>  other creation operators have a shared behaviour, i.e. that they behave like EventEmitters even though that's not the default case

this doesn't sound like true. What creation operator do you mean?

robstoll:
for instance `of(1, 2, 3)`

kwonoj:
`of` doesn't share anything between subscribers.

robstoll:
Exactly, but the above mentioned text implies it a bit IMO

kwonoj:
if you are referring existing doc comment as `above text`, it doesn't imply anything about sharing subscribers as far as I understand it.

robstoll:
Hm... hard to explain apparently. I agree that it does not state explicitly something about sharing.
But then again, why emphasise 
> It does this afresh for each subscriber so although each subscriber may think it is subscribing to the same Observable, in fact each subscriber gets its own individual Observable.

This sounds like it's something special and very important to consider. And I have the feeling that one can quickly confuse `subscribing to the same Observable` with `normally when I subscribe to the same Observable it has a shared behaviour`. 

I don't know, maybe I don't get my point over to you. Thus let's have a look at it differently

>  It waits until an Observer subscribes to it, and then it generates an Observable, typically with an Observable factory function. It does this afresh for each subscriber so although each subscriber may think it is subscribing to the same Observable, in fact each subscriber gets its own individual Observable.

Looking at it closely we probably all have to agree that it is a simplification and actually wrong.
1. defer is returning one Observable and each subscriber is subscribing to this one Observable
2. it depends on the factory function to what Observable the subscriber is subscribed and whether this is a new Observable or not (e.g. returning `EMPTY` would always be the same Observable)

So maybe it would be better to rephrase the whole comment to something like:

>  It waits until an Observer subscribes to it, calls the given factory function to get an Observable -- where a factory function typically generates a new Observable -- and subscribes the Observer to this Observable. In case the factory function returns a falsy value, then `EMPTY` is used as Observable instead. Last but not least, an exception during the factory function call is transferred to the Observer by calling `error`

Thoughts?

cartant:
@robstoll I think the last quoted section in your last comment is an improvement on this PR's current change and on the docs. I'd be happy to merge it, once you've updated the PR.

cartant:
Also, note that there are some lint errors in this PR's current change:

```
ERROR: src/internal/observable/defer.ts[22, 73]: trailing whitespace
ERROR: src/internal/observable/defer.ts[23, 69]: trailing whitespace
```

robstoll:
Closing this in favour of #5266

PR #5283, opened 2020-01-30
chore(npm): fix `serve-and-sync` command

dzhavat:
**Description:**
Fix `serve-and-sync` command fails when working with the docs locally. It happens only after saving files that are located in the `content` folder. Looks like the reason for this is the RegExp that matches the path to the files. 

**Related issue (if exists):**
Close #5282

PR #5284, opened 2020-01-30
fix(concatAll): fix path to file

dzhavat:
**Description:**
Just noticed that #5283 fails because `npm test` command fails due to `Error: Cannot find module 'spec/helpers/observableMatcher'`. Looks like the path to the file was wrong. This fixes that.


cartant:
AFAICT, this should have been fixed in https://github.com/ReactiveX/rxjs/pull/5278

dzhavat:
That's fantastic. Didn't see the PR. :)

Issue #5276, opened 2020-01-29
CI often doesn't kick off for new PRs


cartant:
## Bug Report

Numerous PRs - at least three - that I opened last week did not kick off CI. And to make that happen, I had to push empty commits to the PR's branch.

kwonoj:
gh actions are in, let's see if it works for us.

cartant:
AFAICT, the GitHub actions work and Circle CI just sits there waiting to be kicked. 🤦‍♂ 

This was a PR that I opened about 40 minutes ago:

<img width="523" alt="CI" src="https://user-images.githubusercontent.com/3878593/73577792-99f05080-44c9-11ea-8b62-30dc6c7590ef.PNG">

Your change LGTM. Maybe we can get Circle CI switched off?

In the interim, I'll push some more empty commits.

kwonoj:
Let's make circleci as non required to unblock check in and see few days if gh works reliablt?

kwonoj:
Ok, swapped required check. If something looks odd, feel free to change in settings.

PR #5285, opened 2020-01-31
build(actions): setup initial gh actions script

kwonoj:
As

**Description:**

As discussed in https://github.com/ReactiveX/rxjs/issues/5276, this PR try to setup ci using github actions to see if we can deprecate circleci.


kwonoj:
I think so, though might need confirm by running this action few times.

PR #5291, opened 2020-02-02
Expose the `isPromise` utility function to the public API

BioPhoton:
This could help people that deal with converting asynchronous primitives to observables.
In such cases, the `isObservable` function is used a lot and the check for a `Promise has to get implemented manually.

A common snippet could look like that:

```typescript
type potentialObservableValue<T> = Promise<T> | Observable<T> | undefined | null;
function isObservableValue<T>(nextValue: potentialObservableValue<T>): Observable<T> {
  if (nextValue === null || nextValue === undefined) {
    return of(undefined);
  }
  else if (isObservable(nextValue)) {
    return nextValue;
  } 
  else if (isPromise(nextValue)) {
    return from(nextValue);
  } 
  else {
    throw new Error('Wrong type!');
  }
}
```

Exposing `isPromise`  to the public API can help a lot and come in very handy.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -32,6 +32,7 @@ export { pipe } from './internal/util/pipe';
 export { noop } from './internal/util/noop';
 export { identity } from './internal/util/identity';
 export { isObservable } from './internal/util/isObservable';
+export { isPromise } from './internal/util/isPromise';
 
 /* Error types */
 export { ArgumentOutOfRangeError } from './internal/util/ArgumentOutOfRangeError';


BioPhoton:
This PR is related to https://github.com/ReactiveX/rxjs/issues/5292

Rush:
I agree it would be very useful. Already importing it in several util / reactive libraries. Having it in public API would reduce a risk of the API suddenly changing.

cartant:
https://github.com/ReactiveX/rxjs/issues/4424#issuecomment-456602346

Rush:
I think the point is to be able to assert if a value can be processed by RxJS. At least that's what I care about when I use `isPromise`. If it's promise-like, it's good enough!

cartant:
> ... At least that's what I care about ...

But that's just you. The package has a lot of users. IMO, you should pass the value to `from` which will throw if it cannot be used as an observable source.

kwonoj:
I tend to agree @cartant here, our lib's api suraface do not aim to provide support other than observable. Isobservable is exposed for those purpose as rxjs should gaurantee observable contract implemented in lib. Ispromise is internal detail what we do in core for some interop, doesn't reflect real world promise check. We don't want to make implementation to reflect real world promise (as we only need few interop point).

BioPhoton:
>  `isObservable` is exposed for those purpose as rxjs should gaurantee observable contract implemented in lib. 
I see your point! Also agreee on making sure the contract is possible!

Regarding:
> `isPromise` is internal detail what we do in core for some interop, **doesn't reflect real world promise check.**

And:
> We don't want to make implementation to reflect real world promise (as  **we only need few interop point **).

The [few points for interopt](https://github.com/ReactiveX/rxjs/blob/master/src/internal/util/isPromise.ts#L7) are if it has a `then` method and no `subscribe` method.

Furthermore [`from`](https://github.com/ReactiveX/rxjs/blob/master/src/internal/observable/from.ts#L11) provides a way and contract to create an `Observable` from an `Array`, an array-like object, a `Promise`, an iterable object, or an Observable-like object.

Under the hood, it checks for ` SubscribableOrPromise<T> | ArrayLike<T> | Iterable<T>`
used as `ObservableInput<T>`.  ` SubscribableOrPromise<T>` inturn uses `PromiseLike<T>`

So the contract here can hold if it is  `PromiseLike`, or in other words something like a `Promise` :).

It doesn't have to be a full grade ECMAScript 2015 `Promise`  here. An the consumer is fine with that. 
The check should help to get an `Observable`, not to ensure a `Promise`.

Exposing the  `isPromise`  function helps to ensure the contract. This helps the lib consumer to check before using `from` and don't handle get an error thrown by the creation operator.

What about calling it `isPromiseLike` and exposing it?
```typescript
export { isPromise as isPromiseLike} from './internal/util/isPromise';
```

cartant:
> ... check before using `from` and don't handle get an error thrown ...

What would you do in the check - having determined that it's not a promise - that you couldn't also  do by catching the error?

BioPhoton:
> What would you do in the check - having determined that it's not a promise - that you couldn't also do by catching the error?

Ignoring it maybe. 

`Observable<any> => Observable<ObservableInput<T>>`

Maybe this is a special case, also maybe we drift into discussing special cases.
I guess the contract thing is good enough to stick with and the `isPromiseLike`.

Speaking from my own useage but even more form people out there I see a lot of use for the people.

kwonoj:
I'm seeing usecase of `isPromise` itself seems like special case : if you want to interop promise to observable, as discussed above `from` provides. What's exact usecase isPromise is required other than use promise into observables inside of rxjs library consumer point of view?

BioPhoton:
It seems there are use cases like @Rush mentioned.
Even if things could be done other or the things are done are nor 100% correct I can tell there people use it. 
I work a lot with RxJS and also with people using RxJS and see it often. 
It would definitely help people in the context of RxJS.

If the other suggested things and the last one here are no options I guess we can close this issue to keep the number of open issues low.

cartant:
> It seems there are use cases ...

I'd like to see a use case for `isPromise` being used in an RxJS conext that wouldn't be better handled using something like `from` for obtaining an observable source.

> Even if things ... are nor 100% correct ...

Why should something be encouraged if there are better alternatives?

BioPhoton:
Continue the discussion in closed issue

BioPhoton:
On the way to convince my self. ATM I'm a bit tired so excuse any mistakes. I guess the idea is clear.

What about this?

**Previoue example:**
```typescript
type potentialObservableValue<T> = Promise<T> | Observable<T> | undefined | null;
function isObservableValue<T>(nextValue: potentialObservableValue<T>): Observable<T> {
  if (nextValue === null || nextValue === undefined) {
    return of(undefined);
  }
  else if (isObservable(nextValue)) {
    return nextValue;
  } 
  else if (isPromise(nextValue)) {
    return from(nextValue);
  } 
  else {
    throw new WrongTypeError('Wrong type!');
  }
}

// =====

const anyValue: unknown=  getAnyValue();
const o$: Observable<T | undefined | null> = of(anyValue)
  .pipe(
    isObservableValue(),
    tap({error: e => {
      // potential place to observe custom error
   }}),
  catchError({error: e => {
      // potential place to handle custom error
      return throwError(e);
   }})
  )
 
o$.subscribe({error: e => {
      // potential place to handle custom error
   }})
``` 

**New example:**
```typescript
const anyValue: T | unknown =  getAnyValue();
const o$: Observable<T | undefined | null> = of(anyValue)
  .pipe(
   map(nextValue => ( nextValue === null || nextValue === undefined) ? of(undefined) : nextValue )
   nextValue => from(nextValue);
   tap({error: e => {
      // potential place to observe RxJS error
   }}),
  catchError({error: e => {
     // potential place to rethrow with costom error
     return throwError(WrongTypeError(e));
   }}),
  tap({error: e => {
        // potential place to handle custom error
   }})
  );

o$.subscribe({error: e => {
      // potential place to handle RxJS or if rethrow custom error
   }})
```

PR #5297, opened 2020-02-06
RFC: AsyncIterable Support

benlesh:
This adds support for the following:

`for await` loops consuming RxJS observables:

```ts
const ticker = interval(1000).pipe(take(100));

async function test() {
  for await (const number of ticker) {
    console.log(number);
  }
}

test();
```

RxJS consuming async iterables:
```ts
function snooze(ms) {
  return new Promise(resolve => setTimeout(snooze, ms));
}

async function* values() {
  for (let i = 0; i < 3; i++) {
    await snooze(1000);
    yield i;
  }
}

from(values()).subscribe(x => console.log(x));
of(1).pipe(mergeMap(() => values())).subscribe(x => console.log(x));
```

### Support

Async Iterables are currently supported in all major browsers, but not IE. They are also supported in Node 10 and higher.

The goal of the implementation is as such that it simply won't work in runtimes that do not support it. But as we drop support for older runtimes, possibly by version 8, it would "just work" and be in there.

### Motivation

As time goes on, using Observables with `for await` loops may prove ergonomic and useful.

### Negatives

If the observable emits at a pace faster than the loop completes, there will be a memory build up as the buffer gets more full. We could provide other methods that use different strategies (e.g. just the most recent value, etc), but leave this as the default.  Note that the loop itself may have several `await`s in it, that exacerbate the problem.

Using `for await` to subscribe to an observable forgoes true cancellation. However you can throw an error in the loop and that will kill the observable subscription.

### Other Thoughts

We could just expose `asyncIterableFrom` instead, but all that would do is hurt the ergonomics of the idea more than anything.

`for await`'ing an observable is fundamentally different from `await`-ing the observable (by making `Observable` a "thennable"), for a couple of reasons: 

1. `for await` will instaniate an async iterator from an async iterable (this means it could have side effects), where awaiting a promise is generally expected _not_ to have side effects.

2. There's no question about which value to return, we just yield nexted values... where as with awaiting a thennable observable, we would need to make a decision about what value it returned, if any.

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -3,7 +3,7 @@ import * as sinon from 'sinon';
 import { Observer, TeardownLogic } from '../src/internal/types';
 import { cold, expectObservable, expectSubscriptions } from './helpers/marble-testing';
 import { Observable, config, Subscription, noop, Subscriber, Operator, NEVER, Subject, of, throwError, empty, interval } from 'rxjs';
-import { map, multicast, refCount, filter, count, tap, combineLatest, concat, merge, race, zip, take } from 'rxjs/operators';
+import { map, multicast, refCount, filter, count, tap, combineLatest, concat, merge, race, zip, take, finalize } from 'rxjs/operators';
 
 declare const asDiagram: any, rxTestScheduler: any;
 
@@ -979,6 +979,24 @@ if (Symbol && Symbol.asyncIterator) {
         results.push(value);
       }
       expect(results).to.deep.equal([0, 1, 2]);
+    });
+
+    it('should do something clever if the loop exits', async () => {
+      let finalized = false;
+      const source = interval(10).pipe(take(10), finalize(() => finalized = true));
+      const results: number[] = [];
+      try {
+        for await (const value of source) {
+          results.push(value);
+          if (value === 1) {
+            throw new Error('bad');
+          }
+        }
+      } catch (err) {
+        // ignore
+      }
+      expect(results).to.deep.equal([0, 1]);
+      expect(finalized).to.be.true;
     });
   });
 }
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -2,8 +2,8 @@ import { expect } from 'chai';
 import * as sinon from 'sinon';
 import { Observer, TeardownLogic } from '../src/internal/types';
 import { cold, expectObservable, expectSubscriptions } from './helpers/marble-testing';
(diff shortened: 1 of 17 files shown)


benlesh:
Another thought I have around this: While it's really interesting to have observables direclty supported by `for await`, it might be better for users to be able to choose backpressure strategies here.

Maybe we default the the "all values" behavior above... but we provide a function to only use the latest value or even buffered values:

```ts
// This is the default behavior... all values, in order, one at a time
for await (const value of source$) { }
// or
for await (const value of eachValueFrom(source$)) {
}

// only gives you whatever the last received value was from the observable
// if it doesn't have a value yet, it'll wait for one.
for await (const value of latestValueFrom(source$)) {
}

// gives you all of the values that arrived since the last loop through
// but in an array buffer. If there's nothing in the array, it'll wait for something
// to arrive then give you that in a new buffer (or perhaps you can configure it
// to just give you empty arrays and continue looping?)
for await (const value of bufferedValuesFrom(source$)) {
}
```

spec/Observable-spec.ts
```diff
@@ -946,3 +946,57 @@ describe('Observable.lift', () => {
...
+
+if (Symbol && Symbol.asyncIterator) {
+  describe('async iterator support', () => {
+    it('should work for sync observables', async () => {
+      const source = of(1, 2, 3);
+      const results: number[] = [];
+      for await (const value of source) {
+        results.push(value);
+      }
+      expect(results).to.deep.equal([1, 2, 3]);
```
kolodny: By definition of usage of the `await` keyword, this runs on the next microtask tick, right? Is that worth highlighting?

kolodny:
> However you can throw an error in the loop and that will kill the observable subscription.

Does it need to be that drastic, wouldn't `return`ing in the generator do the same thing?

~The initial comment looks pretty cool, however the followup is already possible today and is arguably more intuitive since the user can choose their strategy off the bat. Would it make sense to expose some asyncInterator methods to use instead of~ On further thought, I agree the proposed solution is better with regards to ergonomics

Should we be deprecating the `ob$.forEach` method with this change?

JWO719:
I really like this change, but I think if we add `for await` it's a logical step to also add `await`, even though there is a semantic difference that is mentioned above. I think adding these two features would be good to lower the barrier of getting into RxJS.

matthewjh:
Love it -- the potential of this enormous. @benlesh, any update (noting you removed the Agenda Item label)?

Issue #4744, opened 2019-04-27
fromFetch cannot abort all responses


cartant:
## <s>Bug Report</s> Feature Request

**Current Behavior**

`fromFetch` returns an observable that emits the request's `Response` object. When `json()` or `text()` is called on the `Response`, a promise is returned. My understanding is that the `AbortController` used in the `fetch` call also aborts promises returned by the aforementioned methods.

In #4742, a bug was fixed: `fromFetch` was calling `abort` as part of the implicit unsubscribe that occurs after completion. This would abort any still pending responses. When `fetch` resolves, a flag is now set to prevent `abort` being called in the unsubscribe teardown.

However, that means there is no way of aborting the promises returned by calls to `Response` methods.

**Possible Solution**

Add another signature to `fromFetch` so that a selector can be passed:

```ts
export function fromFetch<R>(
  input: string | Request,
  init: RequestInit | undefined,
  selector: (response: Response) => Observable<R>,
): Observable<R>;
```

Much like the selector passed to `multicast`, observables composed within the selector will be 'contained' within the observable returned by `fromFetch` and unsubscribing from said observable will facilitate the aborting of any promises returned by `Response` methods within the selector.

If a selector is not passed, `fromFetch` should continue to behave as in #4742 and not abort upon unsubscription if the promise returned by `fetch` has resolved.

**Additional Information**

Note that the changes introduced in https://github.com/ReactiveX/rxjs/pull/3963/files will ensure that unsubscription from the observable returned by `fromFetch` will occur ASAP if said observable serves as the source for, say, `concatMap`. So extending the 'life' of the `AbortController` for an emitted `Response` - if the selector is not used - would be a non-trivial undertaking.

benlesh:
This is resolved in 6.5.2

cartant:
Nope. This isn't resolved.

kambing86:
I have the same issue and found one very interesting finding about this

I can abort the fromFetch if I change the code in `fromFetch.js` from

```js
    fetch(input, init).then(response => {
      abortable = false;
      subscriber.next(response);
      subscriber.complete();
    }).catch(err => {
      abortable = false;
      if (!unsubscribed) {
        // Only forward the error if it wasn't an abort.
        subscriber.error(err);
      }
    });
```

to async/await
```js
    (async () => {
      try {
        const response = await promise;
        abortable = false;
        subscriber.next(response);
        subscriber.complete();
      } catch (err) {
        abortable = false;
[...]

PR #5306, opened 2020-02-09
feat(fetch): add selector

cartant:
**Description:**

This PR adds an optional selector  to `fromFetch` so that it's better able to support chunked transfer encoding.

When chunked transfer encoding is used, the headers are received and the `Response` is resolved before the entire content is received. Without a using selector, it's not possible to abort the promises returned from the `Response`'s methods - `text`, `json`, `blob`, et al.

```ts
const mapped = fromFetch('/foo').pipe(
  mergeMap(res => res.text())
);
```

In the above snippet, once the headers are received, the `Response` will be emitted and mapped to the `text()` promise. Whilst the chunked content is being received, the `fromFetch` will complete, so explicit unsubscription from the `mapped` observable will not abort the `text()` promise and all of the content will be retrieved.

```ts
const selected = fromFetch('/foo', undefined, res => res.text());
```

When a selector is used - as above - explicit unsubscription from the `selected` observable will abort the `text()` promise and the browser will stop the request and will not download further data.

For more information, see the linked issue - #4744 - and I've created a harness here if anyone wants to play around with this stuff: https://github.com/cartant/rxjs-issue-4744

<s>Also, this PR includes some of the changes made in #5305. I'll rebase this once/if that PR is merged.</s>

**Related issue:** #4744

### DIFF
--- a/src/internal/observable/dom/fetch.ts
+++ b/src/internal/observable/dom/fetch.ts
@@ -87,7 +87,7 @@ export function fromFetch(
  *  complete: () => console.log('done')
  * });
  * ```
- * 
+ *
  * @param input The resource you would like to fetch. Can be a url or a request object.
  * @param init A configuration object for the fetch.
  * [See MDN for more details](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Parameters)
--- a/src/internal/observable/dom/fetch.ts
+++ b/src/internal/observable/dom/fetch.ts
@@ -1,5 +1,18 @@
 import { Observable } from '../../Observable';
 import { Subscription } from '../../Subscription';
+import { from } from '../../observable/from';
+import { ObservableInput } from '../../types';
+
+export function fromFetch(
+  input: string | Request,
+  init?: RequestInit
+): Observable<Response>;
+
+export function fromFetch<T>(
+  input: string | Request,
+  init: RequestInit | undefined,
+  selector: (response: Response) => ObservableInput<T>
+): Observable<T>;
 
 /**
  * Uses [the Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) to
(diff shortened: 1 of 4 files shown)


cartant:
BTW, this actually fixes a bug, IMO. Unless a selector is used, a signal that's passed in - via the `init` object - won't always abort the `text()` (or `json()`, et al.) promises for chunked responses.

felixfbecker:
We faced the bug in Sourcegraph too and I can confirm this PR fixes it and everything else still works. Any chance we got get this merged soon? 🙂

src/internal/observable/dom/fetch.ts
```diff
@@ -1,4 +1,18 @@
...
+
+export function fromFetch(
+  input: string | Request,
+  init?: RequestInit
+): Observable<Response>;
+
+export function fromFetch<T>(
+  input: string | Request,
+  init: RequestInit | undefined,
+  selector: (response: Response) => ObservableInput<T>
```
felixfbecker: Looking at most of our uses, an overload for `fromFetch(input, selector)` without `init` would be useful (but not necessary)

felixfbecker:
The upgrade 6.5.3 -> 6.5.4 was actually failing tests for us, but they pass after this PR

src/internal/observable/dom/fetch.ts
cartant: Done

cartant:
@benlesh

> Perhaps just add a `selector` property to `RequestInit`?

Done. And it's a nicer API and implementation, too. Thanks.

benlesh:
Action Items

- [ ] Documentation for chunky bodies :p (just a blurb or something with advice and the "why", maybe?)
- [ ] dtslint test for `selector` case.

cartant:
@benlesh Done. See new commits.

src/internal/observable/dom/fetch.ts
```diff
@@ -51,8 +94,14 @@ import { Subscription } from '../../Subscription';
...
  */
-export function fromFetch(input: string | Request, init?: RequestInit): Observable<Response> {
-  return new Observable<Response>(subscriber => {
+export function fromFetch<T>(
+  input: string | Request,
+  initWithSelector: RequestInit & {
+    selector?: (response: Response) => ObservableInput<T>
+  } = {}
+): Observable<Response | T> {
+  const { selector, ...init } = initWithSelector;
```
iamandrewluca, line 103: A observation. Now that `init` is created using spread, this means that `init` always will be an object, even it has no fields. And below wrap with `if(init)` has no value

https://github.com/ReactiveX/rxjs/pull/5306/files#diff-717ea4ff7feb7c6009e9048143f3ef0fc9ddc6cdeb16b1effcb19d18f3a9c875R119

cartant: @iamandrewluca 👍 Yep. Feel free to open a PR to simplify it.

PR #5307, opened 2020-02-10
Stricter type definition of Subject

hmil:
**Description:**

Make type definition of `Subject` more strict to avoid common bugs when using Subject.

This issue been discussed extensively in #2852 and #5066 .

**Sensitive points**

As discussed in the issues linked above, it is possible to omit the argument to `next` when the Subject is of type `Subject<void>`.
When the template type is anything other than `void`, then an argument for `next` is required.

The primary case where this can be a problem is when people instantiate `Subject<any>` and then use `subject.next()` with no arguments. This will break.

The present change makes the default template type of Subject `void` instead of `any`. This means that when you write `new Subject()`, you get a `Subject<void>` (instead of `Subject<any>`) and are therefore able to write `subject.next()` or `subject.next(value)` if `value` is either `any` or `undefined`.

Other good candidates for default template type are `unknown` and `any`. We can discuss whether `void` is the best choice here, I am not completely sure yet.

### DIFF
--- a/doc/subject.md
+++ b/doc/subject.md
@@ -341,3 +341,39 @@ subject.complete();
 ```
 
 The AsyncSubject is similar to the [`last()`](../class/es6/Observable.js~Observable.html#instance-method-last) operator, in that it waits for the `complete` notification in order to deliver a single value.
+
+## Void subject
+
+Sometimes the emitted value doesn't matter as much as the fact that a value was emitted.
+
+For instance, the code below signals that one second has passed.
+
+```ts
+const subject = new Subject<string>();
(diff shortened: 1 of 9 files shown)


hmil:
@cartant I added tests and some documentation to explain the breaking change and the motivation for choosing `void` as a default type.

benlesh:
The only issue I see with this is that it will break usages like `subject.next()`, users will be forced to pass in `undefined`, which isn't very ergonomic. I'm not sure I want to merge this one. We should ping the TS team on what the best solution is here.

Issue #2675, opened 2017-06-17
SafeSubscriber does not unsubscribe itself


czterocyty:
When subscribing to Observable with Partial Observer, rxjs creates SafeSubscriber, but on unsubscribe it does not unsubscribe itself. Moreover if a closure is passed to next/error callback of partial observer, this leads to memory leaks, as consumers cannot be freed.

See the code (pure JS). 

**RxJS version:** 5.4.1

**Code to reproduce:**

via gist: https://gist.github.com/czterocyty/d4ad5dea55316c9a734c64f378975f04

**Expected behavior:**

I think SafeSubscriber#unsubscribe should be called too upon unsubscription.

Test should pass:

```typescript

  it('upon subscription, the destination SafeSubscriber should be unsubscribed too', () => {
    const observer = {
      next: function () { /*noop*/ },
      error: function (e) { throw e; }
    };

    const sub = new Subscriber(observer);

    const destination = sub['destination'];
    const destUnsubscribe = destination['unsubscribe'] = sinon.spy();
    const destUnsubscribeProtected = destination['_unsubscribe'] = sinon.spy();

    sub.unsubscribe();

    expect(sub.closed).to.eq(true);
    expect(destination.closed).to.eq(true);

    expect(destUnsubscribe).have.been.called;
    expect(destUnsubscribeProtected).have.been.called;
  });
```

**Actual behavior:**

It is not when passing partial observer into Subscriber.

**Additional information:**

After playing around with buttons, check JS heap and not-freed Foo instance due to handing closures in SafeSubscriber. See https://drive.google.com/open?id=0B4JH51FD7JBZcUJTTVRjRXJ2a00. There is Subscriber which is closed (good), but destination as SafeSubscriber is still not closed.

jayphelps:
@trxcllnt is this the same `unsubscribe` issue as what you and I were discussing on Slack? I think so.

* its called on error https://jsbin.com/zudivez/edit?js,console
* called on complete https://jsbin.com/nazixix/edit?js,console
* but not called on unsubscribe https://jsbin.com/lerilub/edit?js,console

martin-the-geek:
Any work being done on this. Just noticed a memory leak in my angular application because of this.

volser:
Any news? I have the same issue (

Medo42:
[...]

PR #5311, opened 2020-02-13
fix(SafeSubscriber): Propagate disposal to terminal Subscriber

trxcllnt:
**Description:**
Ensure the internal SafeSubscriber is disposed when an Observable is subscribed with a Subscriber or Subscriber-like object.

#2675

:heart:

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -599,6 +599,45 @@ describe('Observable', () => {
         console.log = _log;
       });
     });
+
+    it('should fully propogate disposal when subscribed with a Subscriber', () => {
+      let observableUnsubscribed = false;
+      let subscriberUnsubscribed = false;
+      let subscriptionUnsubscribed = false;
+
+      const subscriber = new Subscriber<any>();
+      // verify unsubscribe is called on the Subscriber
+      subscriber.add(() => subscriberUnsubscribed = true);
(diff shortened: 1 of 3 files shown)


cartant:
Sorry for my tardiness in reviewing this. I'll do it tomorrow. It LGTM, but there is one situation with interop observables I want to look into.

spec/Observable-spec.ts
```diff
@@ -599,6 +599,45 @@ describe('Observable', () => {
         console.log = _log;
       });
     });
+
+    it('should fully propogate disposal when subscribed with a Subscriber', () => {
```
jakovljevic-mladen: Typo in word propagate. The other `it` has it also.

benlesh:
There are some concerns here around `Subject` and `Subject`s originating from other copies of RxJS (in Node).

trxcllnt:
I _think_ foreign Subjects should be caught by these lines here, and not be wrapped in a Subscriber:
https://github.com/ReactiveX/rxjs/blob/5cf26e7f567dcc482d37a447eaa61c00d4786f46/src/internal/util/toSubscriber.ts#L16-L18

cartant:
Blocking this because of #5472

trxcllnt:
I think this was superseded by https://github.com/ReactiveX/rxjs/pull/5472. Closing.

PR #5322, opened 2020-02-20
docs(forkJoin): add example of empty array

robstoll:
**Description:**


benlesh:
This is an interesting suggestion, however, I'm not sure if it adds value. I'm 50/50 on this PR. I'll let someone else weigh in.

benlesh:
Also, I'm not sure why lint is failing here.

cartant:
Lint failure from the CI logs:

```
ERROR: src/internal/observable/forkJoin.ts[50, 1]: Exceeds maximum line length of 150
```

Issue #5320, opened 2020-02-18
single(()=>false) should have the same behavior on empty and non-empty sequences as first(()=>false) does.


ghuser:
## Bug Report

In the documentation of [single](https://rxjs.dev/api/operators/single) I read:

> Like first, but emit with error notification if there is more than one value.

The above statement is reasonable but it is not the case in the implementation.

**Current Behavior**

When using `single(()=>false)` on an **empty sequence** there is an error correctly thrown notifying about `no elements in sequence`. But when using `single(()=>false)` on a single-element or multiple-element sequence there is one `undefined` value emitted (unlike `first(()=>false)` behaviour ).

**Reproduction**

    from([]).pipe(single(()=>false)).pipe(tap((v)=>console.log(v))).subscribe();
    //Error [EmptyError]: no elements in sequence]

    from(['x']).pipe(single(()=>false)).pipe(tap((v)=>console.log(v))).subscribe();
    //undefined

    from([]).pipe(first(()=>false)).pipe(tap((v)=>console.log(v))).subscribe();
    //[Error [EmptyError]: no elements in sequence]

    from(['x']).pipe(first(()=>false)).pipe(tap((v)=>console.log(v))).subscribe();
    //[Error [EmptyError]: no elements in sequence]

**Expected behavior**

In my opinion, the same or some other error should be thrown on the latter case of `single` also, instead of an `undefined` value, as it happens with the case of `first(()=>false)`.

**Environment**
- Runtime: [e.g. Node v12.13.1]
- RxJS version: `6.5.4`

**Possible Solution**

**Additional context/Screenshots**

benlesh:
It seems like we're missing an error type for something like "Element Not Found" or whatever. It's not really that the sequence was empty so much as it was nothing matched.

benlesh:
It looks like `single` is wrong in a lot of cases.

If it finds two matches, it just completes and does not error, which is incorrect. It should throw. This needs fixed for v7.

cartant:
Closed by #5325 AFAICT.

PR #5325, opened 2020-02-26
Single operator fixes

benlesh:
Per our discussion of #5320, this is a set of fixes to the behavior of single, which was wildly incorrect. Modernizes and updates tests. Rewrites the operator in the newer functional style.

BREAKING CHANGES:

- `single()` will now throw if more than one value comes from the source. Previously it would not emit.
- `single(predicate)`, in cases where nothing matches the predicate, will now throw a `NotFoundError` if nothing matched, and an `EmptyError` if the source simply did not emit any values.
- `single(predicate)` on an empty source will now emit an `EmptyError`. Previously it would erroneously emit `undefined`.

### DIFF
--- a/spec/operators/single-spec.ts
+++ b/spec/operators/single-spec.ts
@@ -78,8 +78,8 @@ describe('single operator', () => {
   it('should raise error from empty predicate if observable emits error', () => {
     rxTest.run(({ hot, expectObservable, expectSubscriptions }) => {
       const e1 = hot('  --a--b^--#');
-      const e1subs = '  ^--!';
-      const expected = '---#';
+      const e1subs = '        ^--!';
+      const expected = '      ---#';
 
       expectObservable(e1.pipe(single())).toBe(expected);
       expectSubscriptions(e1.subscriptions).toBe(e1subs);
--- a/spec/operators/single-spec.ts
+++ b/spec/operators/single-spec.ts
@@ -1,173 +1,188 @@
 import { expect } from 'chai';
-import { hot, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 import { single, mergeMap, tap } from 'rxjs/operators';
 import { of, EmptyError } from 'rxjs';
-
-declare function asDiagram(arg: string): Function;
+import { TestScheduler } from 'rxjs/testing';
+import { assertDeepEquals } from '../helpers/test-helper';
 
 /** @test {single} */
 describe('single operator', () => {
-  asDiagram('single')('should raise error from empty predicate if observable emits multiple time', () => {
-    const e1 =    hot('--a--b--c--|');
-    const e1subs =    '^    !      ';
-    const expected =  '-----#      ';
-    const errorMsg = 'Sequence contains more than one element';
+  let rxTest: TestScheduler;
 
-    expectObservable(e1.pipe(single())).toBe(expected, null, errorMsg);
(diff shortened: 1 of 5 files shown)


src/internal/operators/single.ts
```diff
@@ -1,40 +1,77 @@
...
+ *  { name: 'Tracy' },
+ *  { name: 'Lily' },
+ *  { name: 'Lincoln' }
+ * );
+ *
+ * source3.pipe(
+ *   single(x => x.name.startsWith('B'))
+ * )
+ * .subscribe(x => console.log(x));
+ * // Error emitted: SequenceError('No values match')
```
cartant: According to the description and the implementation this looks like it should be a `NotFoundError`.

src/internal/operators/single.ts
```diff
@@ -44,80 +81,55 @@ import { Observer, MonoTypeOperatorFunction, TeardownLogic } from '../types';
  *
  * @throws {EmptyError} Delivers an EmptyError to the Observer's `error`
  * callback if the Observable completes before any `next` notification was sent.
+ * @throws {SequenceError} Delivers a SequenceError if more than one value is emitted that matches the
+ * provided predicate. If no predicate is provided, will deliver a SequenceError if more
+ * that one value comes from the source
```
cartant: Could throw `NotFoundError`, too.

spec/operators/single-spec.ts
```diff
@@ -1,173 +1,331 @@
 import { expect } from 'chai';
-import { hot, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 import { single, mergeMap, tap } from 'rxjs/operators';
-import { of, EmptyError } from 'rxjs';
-
-declare function asDiagram(arg: string): Function;
```
cartant: IDK what the consequences of removing this `asDiagram` business are? Will it break website images when the docs are re-created?

spec/operators/single-spec.ts
```diff
@@ -1,173 +1,331 @@
...
 
-    const result = e1.pipe(
-      mergeMap((x: string) => of(x)),
-      single(),
-      mergeMap((x: string) => of(x))
-    );
+  it('should raise error from empty predicate if observable emits error', () => {
+    rxTest.run(({ hot, expectObservable, expectSubscriptions }) => {
+      const e1 = hot('  --a--b^--#');
+      const e1subs = '  ^--!';
```
cartant: These carets (and the implicit carent in the expected marbles) aren't aligned.

spec/operators/single-spec.ts
benlesh: From what I remember, we stop generating them like this, and we're just using old, static images. @JWO719, can you confirm this? I've been removing these `asDiagram` calls when I see them.

JWO719: Lol I never noticed. So it's kinda true, that we are using a static old version of the generated marbleDiagrams, but there's a pr and several issues to improve this mechanism and IIRC it's still using this  `asDiagram` function, so it would be good to keep it for now. I can go through the history to find the deleted images.

This explains now why there is always a diff when I try to generate the new images :D

benlesh: @niklas-wortmann ... can you please set up a separate issue? I've been removing these for the last year at least, because at one point we stopped using them. It's always the first test in the file that had it before. It would be nice if this was something that was imported and used, rather than a global variable, too.

PR #5326, opened 2020-02-26
Fix take and takeLast

benlesh:
Add runtime assertions for JavaScript usage, where invalid argument may be passed causing strange behavior.

BREAKING CHANGE: `take` will throw a `TypeError` if called as `take()` or with an isNaN argument passed.

BREAKING CHANGE: `takeLast` will throw a `TypeError` if called as `takeLast()` or with an isNaN argument passed.

### DIFF
--- a/spec/operators/take-spec.ts
+++ b/spec/operators/take-spec.ts
@@ -4,8 +4,6 @@ import { range, ArgumentOutOfRangeError, of, Observable, Subject } from 'rxjs';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
 
-declare function asDiagram(arg: string): Function;
-
 /** @test {take} */
 describe('take operator', () => {
   let testScheduler: TestScheduler;
@@ -14,7 +12,28 @@ describe('take operator', () => {
     testScheduler = new TestScheduler(observableMatcher);
   });
 
-  asDiagram('take(2)')('should take two values of an observable with many values', () => {
+  it('should error when a non-number is passed to it, or when no argument is passed (Non-TS case)', () => {
+    expect(() => {
+      of(1, 2, 3).pipe(
+        (take as any)()
+      );
+    }).to.throw(TypeError, `'count' is not a number`);
+
+    expect(() => {
+      of(1, 2, 3).pipe(
+        (take as any)('banana')
+      );
+    }).to.throw(TypeError, `'count' is not a number`);
+
+    // Standard type coersion behavior in JS.
+    expect(() => {
+      of(1, 2, 3).pipe(
+        (take as any)('1')
+      );
+    }).not.to.throw();
+  });
+
+  it('should take two values of an observable with many values', () => {
     testScheduler.run(({ cold, expectObservable, expectSubscriptions }) => {
       const e1 = cold(' --a-----b----c---d--|');
       const e1subs = '  ^-------!------------';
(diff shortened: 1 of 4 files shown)


spec/operators/take-spec.ts
```diff
@@ -14,7 +12,28 @@ describe('take operator', () => {
     testScheduler = new TestScheduler(observableMatcher);
   });
 
-  asDiagram('take(2)')('should take two values of an observable with many values', () => {
```
cartant: IDK what the implications of removing this are.

src/internal/operators/take.ts
```diff
@@ -43,51 +43,42 @@ import { MonoTypeOperatorFunction, TeardownLogic } from '../types';
...
  */
 export function take<T>(count: number): MonoTypeOperatorFunction<T> {
-  return (source: Observable<T>) => {
-    if (count === 0) {
-      return empty();
-    } else {
-      return source.lift(new TakeOperator(count));
-    }
-  };
+  if (isNaN(count)) {
```
cartant: I think we should use `Number.isNaN` here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

src/internal/operators/takeLast.ts
```diff
@@ -39,18 +39,24 @@ import { MonoTypeOperatorFunction, TeardownLogic } from '../types';
...
+ * @param count The maximum number of values to emit from the end of
  * the sequence of values emitted by the source Observable.
- * @return {Observable<T>} An Observable that emits at most the last count
+ * @return An Observable that emits at most the last count
  * values emitted by the source Observable.
- * @method takeLast
- * @owner Observable
  */
 export function takeLast<T>(count: number): MonoTypeOperatorFunction<T> {
+  if (isNaN(count)) {
```
cartant: I think we should use `Number.isNaN` here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

spec/operators/takeLast-spec.ts
```diff
@@ -1,164 +1,216 @@
...
 import { takeLast, mergeMap } from 'rxjs/operators';
 import { range, ArgumentOutOfRangeError, of } from 'rxjs';
-
-declare function asDiagram(arg: string): Function;
+import { TestScheduler } from 'rxjs/testing';
+import { assertDeepEquals } from '../helpers/test-helper';
 
 /** @test {takeLast} */
 describe('takeLast operator', () => {
-  asDiagram('takeLast(2)')('should take two values of an observable with many values', () => {
```
cartant: `asDiagram` again ... IDK.

kievsash:
is it expected behavior not to complete if the operator throws because it is used inappropriately?

spec/operators/take-spec.ts
benlesh: Nothing, really. I can be put back, but as of right now, we don't run the diagram generation anymore.

src/internal/operators/takeLast.ts
benlesh: IIRC, that would only return true if it's *actually* `NaN`. People could pass in 'blah' there and it would be `false` if we used `Number.isNaN`.

spec/operators/take-spec.ts
benlesh: We'll tackle these in another PR, cc @niklas-wortmann

src/internal/operators/take.ts
benlesh: Well, not really, `Number.isNaN` would only return true if it were *actually* `NaN`. We want to account for strings that are valid numbers as well. There are people that use RxJS with plain JavaScript (like psychopaths!) and this should be valid for them:

```js
const input = document.querySelector('input[type="text"]');

of(1, 2, 3, 4).pipe(take(input.value)).subscribe(x => console.log(x));
```

DOES THAT CODE SUCK? Yes. Yes it does. But JS has coercion as a feature. 🤷‍♂

PR #5334, opened 2020-03-01
add dtslint bindCallback

artem-galas:
**Description:**
This PR adds dtslint tests for `bindCallback`.

### DIFF
--- a/spec-dtslint/observables/bindCallback-spec.ts
+++ b/spec-dtslint/observables/bindCallback-spec.ts
@@ -1,0 +1,271 @@
+import { bindCallback } from 'rxjs';
+import { a,  b,  c,  d,  e,  f,  g, A, B, C, D, E, F, G } from '../helpers';
+import { SchedulerLike } from '../../src';
+
+describe('callbackFunc', () => {
+  const f0 = (cb: () => any) => {
+    cb();
+  };
+
+  const f1 = (cb: (res1: A) => any) => {
+    cb(a);
+  };
+
+  const f2 = (cb: (res1: A, res2: B) => any) => {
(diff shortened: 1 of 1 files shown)


artem-galas:
I haven't added tests for two overloads:

```
export function bindCallback<A, R>(callbackFunc: (...args: Array<A | ((result: R) => any)>) => any, scheduler?: SchedulerLike): (...args: A[]) => Observable<R>;
export function bindCallback<A, R>(callbackFunc: (...args: Array<A | ((...results: R[]) => any)>) => any, scheduler?: SchedulerLike): (...args: A[]) => Observable<R[]>;
```

Could you please help me to figure out which types should be on `callbackFunction`

Thanks in advance :)

spec-dtslint/observables/bindCallback-spec.ts
```diff
@@ -0,0 +1,271 @@
...
+  it('should enforce function parameter', () => {
+    const o = bindCallback() // $ExpectError
+  });
+
+  it('should accept cb 0 param', () => {
+    const o = bindCallback(f0) // $ExpectType () => Observable<void>
+  });
+
+  it('should accept cb 1 param', () => {
+    const o = bindCallback<A>(f1) // $ExpectType () => Observable<A>
```
cartant, line 35: Specifying type parameters explicitly should not be necessary. Usage should be like this:

```ts
const o = bindCallback(f1) // $ExpectType () => Observable<A>
```

artem-galas: Hm.. I didn't notice that, Thanks.

Issue #5296, opened 2020-02-06
Rename window to windowWith


benlesh:
We need to deprecate `window`, because it was a really dumb name, and rename it to `windowWith` or the like. I suggest `windowWith` just out of the spirit of `concatWith` and `mergeWith` renames... But I'm open to suggestions.

Related: https://github.com/ReactiveX/rxjs/issues/3788

benlesh:
Similarly, we might want to rename `buffer` to `bufferWith` because they're closely related.

BioPhoton:
Is there a concept for a counter part?
if 'With' is used to relate to a operator.
E. g. mergeWith. merge would be the counter part.
this obviously doesn't work for window.

The real problem is the basenaming window.

I suggest take a new name.
windows? slice? span?

benlesh:
I don't even like the name I had here anymore.

Core Team Notes: Ben is dumb. This isn't a good name, and he shouldn't name things.

raymonddavis:
Why not remove it? It appears to me that windowWhen does the same thing? Is there a difference other than typing?

shees-usman:
I would suggest either using `partitionWith` or `sliceWith` since I assume `with` is used to identify that this operation is with another observable emission. Partition already is used to make one stream into two based on some condition, partition could be used to partition it into multiple streams with the trigger being the emitting stream.

yunuseon:
@shees-usman I like both suggestions however I see a small issue with both:

- `partition`: RxJS used to have `partition` both as an operator and a creation function. Both of these referred to partitioning streams into two parts. This could be confusing both for people that still remember the old functions and people ending up reading documentation for deprecated operators by mistake.
- `slice`: In principal I like slice but I think it does not imply a continuous operator but one similiar to `count` which only is done on completion of the stream. 

I would suggest using `segment`. It would fit every current `window` operator:

- `window` -> `segment`
- `windowCount` -> `segmentCount`
- `windowTime` -> `segmentTime`
- `windowToggle` -> `segmentToggle`
- `windowWhen` -> `segmentWhen`

PR #5335, opened 2020-03-01
feat(windowWith): rename window to windowWith

artem-galas:
**Description:**
Since window might have conflict with global window object, rename `window` to `windowWith`

#5296


benlesh:
After the Core Team meeting: I'm not sure I like this name anymore, haha sorry @artem-galas... we're going to punt on this one for a bit and revisit it later. Not closing, just punting.

Issue #4087, opened 2018-09-03
WebsocketSubject close with code and reason


Fruko:
## Feature Request

With WebsocketSubject wrapping native WebSocket object the only way of turning off connection from browser is calling an unsubscribe method. It would be nice to make native close method accessible so we could manually set code and reason while closing the connection

**Is your feature request related to a problem? Please describe.**
We're seeing issue when we call unsubscribe since the event passed to closeObserver looks like this
_wasClean: false, code: 1006_ in Chrome while in Firefox _wasClean: true, code: 1005_

**Describe the solution you'd like**
A new method to pass code and reason to websocket close method same way native WebSocket object supports

luillyfe:
When creating the `webSocketSubject` we are allow to pass a plain object (a.k.a `webSocketSubjectConfig`) that we can use to set code and reason while closing the connection. [From docs](https://rxjs-dev.firebaseapp.com/api/webSocket/WebSocketSubjectConfig#closeObserver).

```ts
import { webSocket } from "rxjs/webSocket";

const webSocketSubjectConfig = {
    closeObserver: {
        next(closeEvent) {
            const customError = { code: 6666, reason: "Custom reason message" }
            console.log(`code: ${customError.code}, reason: ${customError.reason}`);
        }
    }
};

const wsSubject = webSocket(options);
```

jackyyuenahsay:
Any update? The lack of this feature makes us not to upgrade Rxjs6 as we can access the primitive web socket in Rxjs5 such that we can call the close method manually.

If we cannot close with code and reason, is it possible to expose the primitive web socket in Rxjs6?

luillyfe:
@jackyyuenahsay what do you mean by, **but this does not set the native CloseEvent?**.
Since `CloseEvent.code` and `CloseEvent.reason` are `read-only` properties ([check more on this](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent)).

Please share how would you do it with the **native Websocket API**.

Fruko:
@luillyfe when I was creating this feature request. I had in mind this https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close

Where I believe when by using native WebSocket you can manually close and define your own code and reason 

Or am I mistaken?

jackyyuenahsay:
[...]

PR #5346, opened 2020-03-07
feat(webSocketSubject): adding support for closing the socket using a…

luillyfe:
Adding support for closing internal socket with custom error code and message:
```js
import { webSocket } from "rxjs/webSocket";

const webSocketSubjectConfig = {
    closeObserver: {
        next(closeEvent) {
            // Printing custom error and code
            console.log(closeEvent);
        }
    }
};

const wsSubject = webSocket(webSocketSubjectConfig);
wsSubject.close(3000, "custom evil reason!");
```

**Motivation**
Currently we have no way to close the inner socket with a custom error or message that provides to our users a better understanding about what could be possible went wrong with the connection.

 Issue related #4087

### DIFF
--- a/src/internal/observable/dom/WebSocketSubject.ts
+++ b/src/internal/observable/dom/WebSocketSubject.ts
@@ -379,11 +379,16 @@ export class WebSocketSubject<T> extends AnonymousSubject<T> {
   }
 
   unsubscribe() {
-    const { _socket } = this;
-    if (_socket && _socket.readyState === 1) {
-      _socket.close();
-    }
+    this.closeSocket();
     this._resetState();
     super.unsubscribe();
+  }
+
+  closeSocket(code = 1000, message =  '') {
+    const { _socket } = this;
+
+    if (_socket && _socket.readyState === 1) {
+      _socket.close(code, message);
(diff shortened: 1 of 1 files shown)


benlesh:
Hmm... interesting. It makes me wonder if it would be better to have the subject accept actions. like

```ts
webSocket.next(new CloseAction(code, reason));
```

Pros:

1. It would be a better "reactive" flow, because you could compose it.
2. It's a non-breaking change, because internally we could do an `instanceof` check (or a symbol check or something for node's multiple copies of the same package problems, I guess).
3. It doesn't involve adding a new method that people would have to know about.

benlesh:
Turns out this isn't necessary, although we clearly need to update the documentation. If you want to close the connection with a code and a reason, you just have to call `wsSubject.error()` appropriately.. [see code here](https://github.com/ReactiveX/rxjs/blob/28e317e4b1c050a76abb57e798ab419fcbe8ed47/src/internal/observable/dom/WebSocketSubject.ts#L310)

```ts
wsSubject.error({ code: 3000, reason: 'Custom evil reason!' });
```

PR #5352, opened 2020-03-11
docs(subscribe): update `subscribe` documentation

### LINKED ISSUE #5339: [docs]: Replace deprecated subscribe() methods in documentation

[Observable](https://rxjs.dev/api/index/class/Observable)

- [ ] typo
- [ ] error(s) in example
- [ ] needs example
- [x] needs update

### Description Of The Issue

Observable's `subscribe()` method has some examples how to pass parameters, but there are some issues there:
- [*Subscribe with an Observer*](https://rxjs.dev/api/index/class/Observable#subscribe-with-an-observer) example is probably fine.
- [*Subscribe with functions*](https://rxjs.dev/api/index/class/Observable#subscribe-with-functions) uses [deprecated](https://github.com/ReactiveX/rxjs/pull/4202) subscribe method. Should a deprecated example be in the docs? It's OK to show users all (or many) possible ways to subscribe, but I think that deprecated methods should be noted or that examples with deprecated methods should not be used. This example (of summing 1, 2 and 3) is probably bad since it waits for complete event and having `complete` handler is not deprecated only if used with observer. I'd suggest changing an example (only to use `next` handler), but I'd add a note that adding other handlers should be probably done by passing an `observer` object. Is that OK?
- [Cancel a subscription](https://rxjs.dev/api/index/class/Observable#cancel-a-subscription) example should use observer as parameter with no `error` handler at all.

jakovljevic-mladen:
* docs(subscribe): add comma after "In particular"

* docs(subscribe): add notice about asynchronously thrown errors

* docs(subscribe): add notice about deprecated usages of `subscribe` method (relates to #4159)

* docs(subscribe): add notice about using `subscribe` with no params

* docs(subscribe): fix example not to use deprecated `subscribe`

* docs(subscribe): add example with deprecated `subscribe` with notice that it's deprecated

* docs(subscribe): fix unsubscribe example replacing deprecated `subscribe` usage with Observer

* docs(subscribe): add notice that unprovided error handlers will cause throwing errors asynchronously

* docs(subscribe): fix return type

I'm certainly not a native English speaker, but there's a `span` tag in `Observable.subscribe` documentation that says this:

> Use it when you have all these Observables, but still nothing is happening.

This is a bit weird sentence and I don't quite understand it. Is there a way to formulate it a bit batter but still keep the same meaning?

Closes #5339

### DIFF
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -93,20 +93,17 @@ export class Observable<T> implements Subscribable<T> {
    * not attempt to use any RxJS implementation details to create Observers - you don't need them. Remember also
    * that your object does not have to implement all methods. If you find yourself creating a method that doesn't
    * do anything, you can simply omit it. Note however, if the `error` method is not provided and an error happens,
-   * it will be asynchronously thrown. There is no way to catch asynchronously thrown errors by using try/catch as
-   * they are caught by the global error handler (`window.onerror` or `process.on('error)`, depending on the used
-   * runtime). So it's probably the best solution to provide `error` handler to avoid missing thrown errors.
+   * it will be thrown asynchronously. Errors thrown asynchronously cannot be caught using `try`/`catch`. Instead,
+   * use the {@link onUnhandledError} configuration option or use a runtime handler (like `window.onerror` or
+   * `process.on('error)`) to be notified of unhandled errors. Because of this, it's recommended that you provide
+   * an `error` method to avoid missing thrown errors.
    *
    * The second way is to give up on Observer object altogether and simply provide callback functions in place of its methods.
    * This means you can provide three functions as arguments to `subscribe`, where the first function is equivalent
    * of a `next` method, the second of an `error` method and the third of a `complete` method. Just as in case of Observer,
-   * if you do not need to listen for something, you can omit a function. If you plan to omit a function, please keep in mind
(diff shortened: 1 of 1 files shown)


src/internal/Observable.ts
```diff
@@ -95,18 +95,27 @@ export class Observable<T> implements Subscribable<T> {
...
    * defined by that interface, but note that it should be just a regular JavaScript object, which you can create
-   * yourself in any way you want (ES6 class, classic function constructor, object literal etc.). In particular do
+   * yourself in any way you want (ES6 class, classic function constructor, object literal etc.). In particular, do
    * not attempt to use any RxJS implementation details to create Observers - you don't need them. Remember also
    * that your object does not have to implement all methods. If you find yourself creating a method that doesn't
-   * do anything, you can simply omit it. Note however, if the `error` method is not provided, all errors will
-   * be left uncaught.
+   * do anything, you can simply omit it. Note however, if the `error` method is not provided and an error happens,
+   * it will be asynchronously thrown. There is no way to catch asynchronously thrown errors by using try/catch as
+   * they are caught by the global error handler (`window.onerror` or `process.on('error)`, depending on the used
```
jakovljevic-mladen: Taken from Ben's talk [about RxJS 6](https://youtu.be/JCXZhe6KsxQ?t=724).

cartant (approved):
Merging this 'cause the changes improve the docs, IMO, and 'cause this PR makes no changes to code whatsoever.

Issue #5362, opened 2020-03-22
Object arguments for combineLatest and forkJoin


rraziel:
## Feature Request

**Is your feature request related to a problem? Please describe.**
When extensively using `combineLatest` or `forkJoin`, especially with many arguments, it can be cumbersome to have the result as an array (e.g. you can't `pluck`) and I typically end up having a single:

```
map(([a,b,c]) => ({a,b,c}))
```

Except it can be with many arguments and their names are obviously longer.

Moreover, the result of a `combineLatest(a,b,c)` (with 7+ arguments) is a `[typeof(a)|typeof(b)|typeof(c)]`, which is not very type-safe, e.g.:

```
combineLatest([
  of(true),
  of('hello world'),
  of(42)
]).pipe(
  map(([a, b, c]) => {
    // a is boolean|string|number
    // b is boolean|string|number
    // c is boolean|string|number
  })
);
```

**Describe the solution you'd like**
On top of taking an array, `combineLatest` and `forkJoin` could take an object where each field is an `Observable`. It would then return an object with what the observable emitted, e.g:

```
combineLatest({
  a: of(true),
  b: of('hello world'),
  c: of(42)
}).pipe(
  map(({a, b, c}) => ({
    // a is boolean
    // b is string
    // c is number
  })
);
```

**Describe alternatives you've considered**
Writing a pipable operator, however due to TypeScript limitations there is not much that can be done to turn the array into an object, especially after `combineLatest` and `forkJoin` already lost most of the type information for each array element.

kwonoj:
I believe this is dupe of https://github.com/ReactiveX/rxjs/issues/5022 .

Issue #5022, opened 2019-09-17
Support dictionary in combineLatest


Akxe:
## Feature Request

`combineLatest` should accept dictionary/object same as `forkJoin`. Implementation in `forkJoin` #4640 

```
combineLatest({
  width: this.width$,
  height: this.height$,
}).pipe(
  tap(({width, height}) => console.log(`dimensions: ${width}x${height}`))
)
```

### Reasoning
It is much easier to mess up when deconstructing array than deconstructiong object. Thus this change would enable developer to name their streams, reducing the potential for error.

benlesh:
This is something we're planning, but I'm not sure we can do until the deprecated versions of `combineLatest` are removed in version 8.

intellix:
This would be awesome in Angular when transitioning to observable values the view can get real messy. Would be nice to just use a single async pipe at the top and use values like they're not async

amakhrov:
Currently a workaround is to use array-based `combineLatest` followed by `map()`:
```
const abc$ = combineLatest([a$, b$, c$]).pipe(
  map(([a, b, c]) => ({a, b, c}))
)
```
It's not too bad, but it results it repeating code (also error-prone, as @Akxe has pointed out)

Also it's probably not too hard to create a custom wrapper to implement that in your project.

rraziel:
Is there any plan on implementing this, or at least would a pull request be considered for merging?

Goobles:
You can create your own operator functions in the meantime as a workaround:
https://gist.github.com/Goobles/45f0b3027d31884dba0b164ee2cc50ea

Akxe:
@rraziel I belive this can now be closed, right?

rraziel:
@Akxe this is available in rxjs 7 yes

tmtron:
[RxJs7 docs: Combine a dictionary of Observables](https://rxjs.dev/api/index/function/combineLatest#combine-a-dictionary-of-observables)

PR #5363, opened 2020-03-23
feat(combineLatest): support for observable dictionaries (#5022)

rraziel:
**Description:**
Support for passing a dictionary of observables to `combineLatest`, removing the need to pass an array then directly using `map` to turn it back into an object.

I mostly based this on the similar work that was done on `forkJoin`.

#5022 and #5362

### DIFF
--- a/spec-dtslint/observables/combineLatest-spec.ts
+++ b/spec-dtslint/observables/combineLatest-spec.ts
@@ -123,3 +123,13 @@ it('should accept 6 params and a result selector', () => {
 it('should accept 7 or more params and a result selector', () => {
   const o = combineLatest([a$, b$, c$, d$, e$, f$, g$, g$, g$], (a: any, b: any, c: any, d: any, e: any, f: any, g1: any, g2: any, g3: any) => new A()); // $ExpectType Observable<A>
 });
+
+describe('combineLatest({})', () => {
+  it('should properly type empty objects', () => {
+    const res = combineLatest({}); // $ExpectType Observable<never>
+  });
+
+  it('should work for the simple case', () => {
+    const res = combineLatest({ foo: a$, bar: b$, baz: c$ }); // $ExpectType Observable<{ foo: A; bar: B; baz: C; }>
+  });
+});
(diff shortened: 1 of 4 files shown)


rraziel:
I updated the commit after an `npm run test:side-effects:update`, hopefully that was the right thing to do?

rraziel:
Anything missing or that I need to fix in this PR?

joshribakoff:
@rraziel not sure if you saw the merge conflicts?

rraziel:
Yes, they appeared few weeks after my PR, I was kind of hoping for someone to actually answer my questions before doing a rebase 😅

rraziel:
@joshribakoff alright, rebased and fixed the side-effect weirdness

rraziel:
I'll resolve the conflicts when someone decides there is any point in having this PR, since there hasn't been any comment since the initial PR over 3 months ago.

ladyleet:
@benlesh @cartant @niklas-wortmann any thoughts on this?

rraziel:
Rebased + api_guard files updated.

cartant:
> Rebased + api_guard files updated.

Thanks. I'll review this later today.

spec-dtslint/observables/combineLatest-spec.ts
```diff
@@ -123,3 +123,13 @@ it('should accept 6 params and a result selector', () => {
...
   const o = combineLatest([a$, b$, c$, d$, e$, f$, g$, g$, g$], (a: any, b: any, c: any, d: any, e: any, f: any, g1: any, g2: any, g3: any) => new A()); // $ExpectType Observable<A>
 });
+
+describe('combineLatest({})', () => {
+  it('should properly type empty objects', () => {
+    const res = combineLatest({}); // $ExpectType Observable<never>
+  });
+
+  it('should work for the simple case', () => {
+    const res = combineLatest({ foo: of(1), bar: of('two'), baz: of(false) }); // $ExpectType Observable<{ foo: number; bar: string; baz: boolean; }>
```
cartant, line 133: I would prefer this test to use the helpers - `a$`, etc. - instead of `of` with literal values.

I can see that what's here is based on the `forkJoin` tests - which use `of` and literal values - but that's because those tests have not been refactored to use the helpers. Going forward, we should favour using the helpers.

rraziel: I see the helpers are - so far - only used in `spec-dtslint/operators/*.spec.ts` and not in `spec/observables/*.spec.ts`.
Can I just import those same helpers in `spec/`? (as they are in `spec-dtslint/helpers`)

cartant: Nope. They're only for use - at least ATM - in the dtslint tests. They offer a somewhat more disciplined approach regarding the types - there are only a few different literal types, but we can add as many distinct 'helper' types as is necessary. With the tests in `spec`, we're less interested in the types and more interested in the notifications and their values and strings, etc. would be fine there.

benlesh:
Thank you, @rraziel! Very nice!

PR #5364, opened 2020-03-23
Safe Observable#subscribe

yogurt1:
```js
const observable = new Observable(subscriber => {
  subscriber.next(1);
});

const subscription = observable.subscribe(value => console.log(value)); // prints 1

subscription.next(2); // prints 2
```
Subscribe method returns Subscriber object while Subscription expected
I fixed UB


yogurt1:
Please delete this one
Wrong branch 😢

PR #5365, opened 2020-03-23
Safe Observable#subscribe

yogurt1:
```js
const observable = new Observable(subscriber => {
  subscriber.next(1);
});

const subscription = observable.subscribe(value => console.log(value)); // prints 1

subscription.next(2); // prints 2
```
Subscribe method returns Subscriber object while Subscription expected
I fixed UB

### DIFF
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -229,7 +229,9 @@ export class Observable<T> implements Subscribable<T> {
       }
     }
 
-    return sink;
+    const subscription = new Subscription();
+    subscription.add(sink);
+    return subscription;
   }
 
   /** @deprecated This is an internal implementation detail, do not use. */


benlesh:
Yeah. This is a known issue as a leaked implementation detail. I'm not sure if we should tackle it now, or simply wait to migrate to AbortSignal as it becomes ubiquitous as a cancelation mechanism.

src/internal/Observable.ts
```diff
@@ -229,7 +229,9 @@ export class Observable<T> implements Subscribable<T> {
       }
     }
 
-    return sink;
+    const subscription = new Subscription();
+    subscription.add(sink);
+    return subscription;
```
benlesh, line 234: There are performance implications do us doing this with every subscribe call. I'm not sure this is the *exact* approach we can take. Ideally, we'd figure how where the last subscription was and do it there... but there are other things that are being discussed that will solve this issue.

benlesh:
After some thinking. I think we have to close this for now. While it's definitely a leaked implementation detail, I think the fact that it will add overhead to execution means it might be something we're willing to accept for the time being. I can see version 8 or 9 bringing some positive changes to this area.

I will add this to the core team agenda for further discussion. But I'm going to say "closed for now".

Thank you for your thought and effort, @yogurt1! It's very much appreciated.

PR #5384, opened 2020-04-09
feat(ajax): implement url params inside ajax body

OmarioHasan:
implement url query params to make it easy for user to pass it dynamically as an object and not to make it by himself and joining it with the url

### DIFF
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -676,6 +676,35 @@ describe('ajax', () => {
       expect(complete).to.be.true;
     });
 
+    it('request for url and params should succeed on 200', () => {
+      const expected = { foo: 'bar' };
+      let result;
+      let complete = false;
+      const obj: AjaxRequest = {
+        url: 'http://talk-to-me-goose',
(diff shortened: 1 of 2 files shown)


OmarioHasan:
@benlesh  I think about this change it's a good change because i use Ajax alwayes with qParams more than HttpClient in Angular.
Did you see that it missed something ?

benlesh:
We've decided to hold off on adding this feature for now. We'll wait to see if it's requested in issues as time goes on.

OmarioHasan:
No problem
thanks for reply and your good efforts

benlesh:
I'm still not totally sure we're going to add this feature, but I've added a new PR at #6174, and I'm closing this one in favor of that one. They're both similar but there's a few things that are different about the updated PR.

One thing is that this tries to use `new URL`, which breaks if the user passes a relative url like `/this`. The other difference is that the new PR is more permissive with what it allows the user to pass as `param`, including `URLSearchParams`, a dictionary, an array of tuples, or a query string directly.

Thank you so much again for your hard work. If we land the new feature hopefully you find it useful.

PR #5389, opened 2020-04-13
fix(Subject): next value argument optional in TS again

benlesh:
- After core team discussion about how much was broken by properly typing Subject next, we decided it was prudent to return the typing fof the next value argument to being optoinal.

Although it is incorrect, fixing the typing broke a huge number of TypeScript builds within Google and therefore would break a large number of builds externally. Therefore, we have decided to introduce a lint rule to help people gradually move toward the proper typings for Subject.

- Also marks a few Subject internal-only APIs as deprecated, so people can stop using them ahead of the next major.


benlesh:
In a core team meeting, we've decided to leave this as "correct". It was the whole point of this major, to get the typings right.

Issue #3861, opened 2018-06-25
Subject.pipe() returns an AnonymousSubject even though the type says it's an plain Observable


szymonbultrowicz:
## Bug Report

**Current Behavior**
Calling `pipe()` method on a `Subject` returns an `AnonymousSubject` but according to the types it should be a plain `Observable`. Firstly, it allows a consumer to cast the object as a `Subject` and access the `next` function. Secondly, it prevents a developer to avoid such problems by calling `asObservable`, because TypeScript doesn't allow to call it on an object of type `Observable`.

**Reproduction**
https://stackblitz.com/edit/rxjs-pipe-anonymous-subject

```typescript
const piped = new Subject<string>().pipe(
  map((el1) => el1),
);
piped.subscribe(value => console.log('Got value', value))
(piped as Subject<string>).next('banana');
// Output: Got value banana

new Subject<string>().pipe().asObservable() // not allowed
```

**Expected behavior**
The output of `pipe` function should be a plain `Observable` or expose a method `asObservable`.

**Environment**
- RxJS version: 6.2.1

PR #5390, opened 2020-04-13
feat(FrankenSubject): adds type (DO NOT MERGE YET)

benlesh:
- updates deprecation message for `Subject.create`, as it was incorrect.

- Adds some documentation for `FrankenSubject`

NOTE: No tests yet. Requires discussion. This is something that could be used to create better types for WebSocketSubject, etc.

### DIFF
--- a/src/internal/FrankenSubject.ts
+++ b/src/internal/FrankenSubject.ts
@@ -1,0 +1,53 @@
+import { Observable } from './Observable';
+import { Observer, PartialObserver, ObservableInput } from './types';
+import { hostReportError } from './util/hostReportError';
+import { from } from './observable/from';
+
+/**
+ * Glues any observer to any observable to make a "subject".
+ *
+ * Assembles the parts like Frankenstein to make your own monster.
+ *
+ * Developer's note: `FrankenSubject` does NOT extend `Subject`, therefor
+ * `instanceof Subject` checks will not work.
+ */
+export class FrankenSubject<In, Out> extends Observable<Out> implements Observer<In> {
+  private _closed = false;
+
+  get closed() {
+    return this._closed;
+  }
+
+  constructor(private observer: PartialObserver<In>, source: ObservableInput<Out>) {
+    super();
+    this.source = from(source);
+  }
+
+  next(value: In): void {
+    if (!this.closed) {
+      this.observer.next?.(value);
+    }
+  }
+
+  error(err: any): void {
+    if (!this._closed) {
+      this._closed = true;
+      if (typeof this.observer.error === 'function') {
+        this.observer.error(err);
+      } else {
+        hostReportError(err);
+      }
+    }
+  }
+
+  complete(): void {
+    if (!this._closed) {
+      this._closed = true;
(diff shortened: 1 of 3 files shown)


src/internal/FrankenSubject.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Glues any observer to any observable to make a "subject".
+ *
+ * Assembles the parts like Frankenstein to make your own monster.
+ *
+ * Developer's note: `FrankenSubject` does NOT extend `Subject`, therefor
+ * `instanceof Subject` checks will not work.
+ */
+export class FrankenSubject<In, Out> extends Observable<Out> implements Observer<In> {
```
niklas-wortmann, line 14: Just thinking out loud, wouldn't it be a good idea to override the pipe method here to fix the types specially for subjects.

Today if you call `pipe` on a Subject it will still return a Subject even though the typing say it is an observable ...

niklas-wortmann, line 14: *Sorry made that comment already before, but deleted it unintentionally*

Would this be a good idea to tackle the typings of the `pipe` method for subjects? I did locally a small POC by overriding `pipe` within the Subject base class. It was just one test failing, but I can imagine any side effects I am not aware of right now?
I think it might be confusing that the typing says it returns an observable though it's a subject instance.

niklas-wortmann: I saw that this was already raised in #3861, but actually I don't understand the explanation @cartant  :D

cartant: ATM, `lift` is pretty much an implementation detail. Expressing that in the types means that `lift` itself will need to be explained and would make any re-implementation of RxJS that doesn't involve `lift` even more of a breaking change. AFAICT, the type changes would only be necessary for use cases that should, IMO, be discouraged, so I see no wins here. Only a documentation and type-maintenance burden. ATM, it's trivial for someone who wants to rely upon subjects being lifted to perform a type assertion. IMO, way more than 95% of people would never need this.

benlesh:
Maybe what we need is a new way to create Subjects, and not this crazy town type. (from core team meeting)

cartant (commented):
IMO this needs a different name. I don't think it should be called a subject.

AFAICT, in this and other Rx implementations, the values that are `next`-ed into a subject are the values that are received by the subject's subscribers. That is the opposite of what's happening here.

IMO, the name should reflect the fact that the `next`-ed values are sent someplace other than the source to which subscribers are subscribed.

`AnonymousSubject` is used in the implementation of `Subject#lift` and is used as the base class for `WebSocketSubject`. If the `FrankenSubject` is used to replace `AnonymousSubject` - is that the plan? - lifting will be broken, AFAICT, unless `FrankenSubject` implements `lift`.

---

If the word 'subject' is going to be used, I think the name should represent what it does - not just that it's weird. Maybe `IndirectSubject` or `RedirectSubject` or something. IDK.

benlesh:
I'm not sure I want to do this anymore, TBH. Closing for now.

PR #5396, opened 2020-04-18
docs: Fix simple typo, stoted -> stored

timgates42:
There is a small typo in docs_app/tools/transforms/README.md.

Should read `stored` rather than `stoted`.


cartant:
Thanks.

PR #5402, opened 2020-04-23
feat(dequeueReplay): add DequeueReplaySubject and dequeueReplay operator

DanielSchaffer:
**Description:**
Add a new Subject subtype, DequeueReplaySubject, which can replay events like ReplaySubject, but also remove events from its internal buffer using a "dequeue trigger" Observable.

### DIFF
--- a/docs_app/content/guide/operators.md
+++ b/docs_app/content/guide/operators.md
@@ -222,6 +222,7 @@ Also see the [Join Creation Operators](#join-creation-operators) section above.
 
 ### Multicasting Operators
 
+- [`dequeueReplay`](/api/operators/dequeueReplay)
 - [`multicast`](/api/operators/multicast)
 - [`publish`](/api/operators/publish)
 - [`publishBehavior`](/api/operators/publishBehavior)
--- a/spec/operators/dequeueReplay-spec.ts
+++ b/spec/operators/dequeueReplay-spec.ts
@@ -1,0 +1,183 @@
+import { EMPTY, NEVER, Observable } from 'rxjs';
+import { dequeueReplay } from 'rxjs/operators';
+import { cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
+
+declare function asDiagram(arg: string): Function;
+
+/** @test {dequeueReplay} */
+describe.only('dequeueReplay operator', () => {
+
+  asDiagram('dequeueReplay')('should emit only values that have not been dequeued for' +
+    ' subsequent subscribers', () => {
+    const source = cold( '-a-b--------');
+    const dequeue = cold( '----b');
+    const dequeueSub =    '-^';
+    const sub1 =         '^---';
+    const sub1Expected = '-a-b';
+    const sub2 =         '------^-';
+    const sub2Expected = '------a';
+
+    const dequeueTest = source.pipe(dequeueReplay(dequeue));
+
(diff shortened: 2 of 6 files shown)


DanielSchaffer:
I'm having a couple issues following the [contributing guidelines](https://github.com/ReactiveX/rxjs/blob/master/CONTRIBUTING.md):

- Performance tests and scripts don't exist where it says they do
- Generating the png for my marble diagram fails with `gm convert: Non-conforming drawing primitive definition (line) [Result too large].`
- Try to run any "watch" script for the docs app fails with this error on any change: `Error generating docs TypeError: Cannot read property 'name' of undefined`

I also have a question about unit tests for my subject - if the operator tests cover the Subject, are separate tests just for the subject also required? I had a spec for it, but it ended up being nearly identical to the operator tests, so I left it out for now.

kwonoj:
Appreciate for PR! before getting into those detail, we do ask some questions prior to those for any new features, mainly if it should be core package. We do strongly recommend to publish new features as userland npm modules before land in core due to couple of reasons, primarily remaining number of core API surface, as well as introduce new api as stable manner. 

For those reason, I recommend to publish this as userland module as well. If you believe it should be core operator directly would like to have some detailed explanations for reasons.

DanielSchaffer:
@kwonoj okay, that makes sense. Admittedly, this functionality probably doesn't need to be in core. I've been using RxJS for a couple years now, and only ran into the need for it recently trying to write logic for a game.

To clarify: by "userland module", you mean just publishing my own npm module, right?

Also, do you have any suggestions for that error I'm running into with `tests2png`?

kwonoj:
> To clarify: by "userland module", you mean just publishing my own npm module, right?

yes.

> Also, do you have any suggestions for that error I'm running into with tests2png?

`test2png` is somewhat broken state now so probably not worth to invest much time in my opinion.

DanielSchaffer:
Okay, is there an alternative method for generating the marble diagram images then?

kwonoj:
not much afaik except manually drawing.

cartant:
@DanielSchaffer FWIW, years ago, I used this for some simple marble diagrams, in a blog article: https://bitbucket.org/achary/rx-marbles/src/master/

martinsik:
I was able to use the official generator from this repo to create marble diagrams for my custom operators. https://github.com/martinsik/rxjs-extra/tree/master/doc/marble-diagrams

However, it's not nice. I had to clone this repo and so on. I described the way I did it here https://github.com/martinsik/rxjs-extra#generating-marble-diagrams with some extra notes for Windows. This is important as well https://github.com/martinsik/rxjs-extra/blob/master/spec/tests2png.opts.

This reminds me that I wanted to post an issue regarding the generator and that it could be exported and more configurable.

cartant:
@DanielSchaffer Today, I had a need to generate a marble diagram and I found this:

https://github.com/timdp/swirly

It's pretty cool.

tonivj5:
@cartant, out of curiosity, there's a PR (https://github.com/ReactiveX/rxjs/pull/5019) proposing that generator by its creator @timdp

benlesh:
My understanding is that there's agreement that this should be in a third-party library and not here? I'll close this for now, if I'm incorrect, let me know and I'll reopen and mark it as an agenda item so it can be discussed by the team.

PR #5403, opened 2020-04-23
Deprecate toPromise.

leggechr:
Link to firstValueFrom or lastValueFrom instead.

### DIFF
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -348,8 +348,11 @@ export class Observable<T> implements Subscribable<T> {
   }
 
   /* tslint:disable:max-line-length */
+  /** @deprecated Deprecated use {@link firstValueFrom} or {@link lastValueFrom} instead */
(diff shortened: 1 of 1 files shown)


src/internal/Observable.ts
```diff
@@ -364,6 +367,7 @@ export class Observable<T> implements Subscribable<T> {
    * rejects on an error. If there were no emissions, Promise
    * resolves with undefined.
    */
+  /** @deprecated Deprecated use {@link firstValueFrom} or {@link lastValueFrom} instead */
```
jakovljevic-mladen, line 370: Hi there. This note is added in the new comment block. Is this intended? I'm not sure, but this may hide the first comment block for the documentation generator. Maybe this `@deprecated` part should be moved to the first comment block?

leggechr: Fixed!

benlesh:
thanks, @leggechr!

PR #5419, opened 2020-05-08
docs(general): clarify introduction section

niklas-wortmann:
I started to go over the general information section to remove some technical bla bla. It aims to make it easier for the reader to understand the important concepts and focus on the relevant parts!

Additionally, I slightly changed the sidebar and split the guides in an "Essential" and an "Advanced" section. We could add a small paragraph somewhere describing this structure to give an overview, but I think for now I can imagine this to help people to get into this by focusing on the essentials. At the same time, I split some larger sections into multiple smaller one and put them in the related category. E.g. the former `operators` guide is split into `operators`, `marble diagrams` & `higher-order Observables`

### DIFF
--- a/docs_app/content/guide/installation.md
+++ b/docs_app/content/guide/installation.md
@@ -22,7 +22,7 @@ To import only what you need using pipeable operators:
 import { of } from 'rxjs';
 import { map } from 'rxjs/operators';
 
-of(1,2,3).pipe(map(x => x + '!!!')); 
+of(1,2,3).pipe(map(x => x + '!!!'));
 ```
 * See [Pipeable Operator Documentation](https://github.com/ReactiveX/rxjs/blob/91088dae1df097be2370c73300ffa11b27fd0100/doc/pipeable-operators.md) for more information about pipeable operator.
 
--- a/docs_app/content/guide/installation.md
+++ b/docs_app/content/guide/installation.md
@@ -22,7 +22,7 @@ To import only what you need using pipeable operators:
 import { of } from 'rxjs';
 import { map } from 'rxjs/operators';
 
-of(1,2,3).pipe(map(x => x + '!!!')); // etc
+of(1,2,3).pipe(map(x => x + '!!!')); 
 ```
 * See [Pipeable Operator Documentation](https://github.com/ReactiveX/rxjs/blob/91088dae1df097be2370c73300ffa11b27fd0100/doc/pipeable-operators.md) for more information about pipeable operator.
 
@@ -32,7 +32,7 @@ To use with globally imported bundle:
 const { of } = rxjs;
 const { map } = rxjs.operators;
 
-of(1,2,3).pipe(map(x => x + '!!!')); // etc
+of(1,2,3).pipe(map(x => x + '!!!'));
 ```
 
 ## CommonJS via npm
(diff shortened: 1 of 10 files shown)


docs_app/content/guide/installation.md
```diff
@@ -22,7 +22,7 @@ To import only what you need using pipeable operators:
 import { of } from 'rxjs';
 import { map } from 'rxjs/operators';
 
-of(1,2,3).pipe(map(x => x + '!!!')); // etc
+of(1,2,3).pipe(map(x => x + '!!!'));
```
jakovljevic-mladen, line 25: Hi there. There is an empty space left after `;` which might be removed. The other one couple of lines below is removed.

docs_app/content/guide/overview.md
```diff
@@ -1,127 +1,50 @@
...
 
-fromEvent(document, 'click').subscribe(() => console.log('Clicked!'));
+console.log(c);
 ```
 
-### Purity
+Generally, considering usual programming paradigms the code above will log `20`. In a fully reactive approach the code would consider the change in the variable `a` and therefore log `30` instead. 
 
-What makes RxJS powerful is its ability to produce values using pure functions. That means your code is less prone to errors.
+This is only an example demonstrating the behavior of reactive programming.
```
jakovljevic-mladen, line 16: This file has also couple of empty spaces at the end of some lines. If this is not important, please let me know, I won't report it anymore.

niklas-wortmann: I don't have really strong opinions about that. IMO it helps maintaining the markdown file

docs_app/content/guide/installation.md
niklas-wortmann: thanks for spotting this

docs_app/content/guide/overview.md
jakovljevic-mladen: I kinda don't like empty spaces at the end of lines, they look a bit dirty to me. But if they can help, sure, keep having them.

docs_app/content/guide/observer.md
```diff
@@ -29,13 +29,13 @@ const observer = {
 };
 ```
 
-When subscribing to an Observable, you may also just provide the callbacks as arguments, without being attached to an Observer object, for instance like this:
+When subscribing to an Observable, you can also provide the callbacks as anonymous function, without wrapping it in an object:
```
joshribakoff, line 32: Whether or not you pass an anonymous function to the argument, does not matter, i don't think.  https://rxjs-ftnxyh.stackblitz.io

Could we consider rewording to "pass a references to callback functions as the arguments"?

niklas-wortmann: good point. Thanks a lot for your input on that. Happy to use this wording

benlesh:
@niklas-wortmann  we need to think of a way to merge this work with the work I'm doing over on #5592

niklas-wortmann:
@benlesh as #5592 is going in master and will therefore be the updated docs for v7 and upcoming I would suggest moving this to the v6 branch to add some clarifications there.

benlesh:
Core Team: this goes to v6 docs.

benlesh:
Looks like this needs rebased.

niklas-wortmann:
closed by #5871

Issue #5100, opened 2019-10-20
tests2png is defeated by TestScheduler#run()


timdp:
## Bug Report

**Current Behavior**

As part of #5019, I discovered that the `tests2png` script currently (as of 1b2021f) renders the following PNG images as only the operator and no observables:

- audit
- auditTime
- buffer
- bufferCount
- bufferTime
- bufferToggle
- bufferWhen
- catch
- combineAll
- combineLatest
- delay

**Example**

![audit](https://user-images.githubusercontent.com/201034/67158253-240b1380-f336-11e9-85cd-c8eef6d89d18.png)

**Diagnosis**

I did a `git bisect` and found out that the issue is introduced by multiple commits. For `audit`, the behavior started in c41d00b. For `auditTime`, it's broken as of bc91ba0.

This suggests that the problem is the introduction of `TestScheduler#run()`.

**Environment**

- Runtime: Node v12.11.0
- RxJS version: 1b2021f

timdp:
Still seeing this on master as of d2f68db3.

Pinging @JWO719 for visibility.

timdp:
Still happening with d7f70782.

Additionally, I had to `npm install ts-node tsconfig-paths -D` first.

I guess the diagrams haven't been receiving a lot of love recently?

JWO719:
The diagrams definitely didn't received a lot of love in the recent time, that's true. Sorry for coming back this late to you. To be honest, I kinda lost track on the issues with the marble diagrams in general. Is there anything I can support you with, because I would really love to have the improved marble diagrams you worked on?

timdp:
I won't be able to take a look for another couple of weeks due to medical reasons, but I think it would definitely help to have some form of documentation on how the testing APIs changed. Perhaps the tests' authors could share an explanation?

niklas-wortmann:
This topic kinda poped up again. Sorry I didn't respond in the meantime. I'll take a look and when I am stuck connect to the people that might can help here

niklas-wortmann:
Just to be totally clear, so besides the operator listed at the top, it is working properly in the current state?

timdp:
With 02f458d, in addition to the operators in my original message, the following now also render empty:

* concat
* concatAll
* concatMap
* take

I also still had to `npm install ts-node tsconfig-paths -D` on a clean repo before I was able to run tests2png.

niklas-wortmann:
Ok great I will have a look at this tomorrow and come back to you!

[...]

PR #5422, opened 2020-05-10
Remove 'asDiagram' infrastructure

niklas-wortmann:
This cleans up the old way of generating marble diagrams. It also hides the internal testing document from the docs, we can still refer to that but it's not visible from the menu anymore.

That's how the new diagrams (inlined as svg) do kinda look like:
image


timdp:
Guessing this relates to #5019, #5100, and #5402, so just cross-linking for visibility...

marbles/scripts/index.ts
```diff
@@ -0,0 +1,44 @@
+import { renderMarbleDiagram } from 'swirly-renderer-node';
+import { readdir, readFileSync, writeFileSync } from 'fs';
+import { join } from 'path';
+import { parseMarbleDiagramSpec } from 'swirly-parser';
+import { styles as defaultLight } from 'swirly-theme-default-light';
```
timdp, line 5: It's fine if you don't provide all styles. Swirly will extend the default light theme with whatever you do provide.

marbles/scripts/index.ts
```diff
@@ -0,0 +1,44 @@
...
+};
+
+const styles: DiagramStyles = {
+    ...defaultLight,
+    frame_width: 20,
+    completion_height: 20,
+    higher_order_angle: 30,
+    arrow_fill_color: 'black',
+    background_color: 'rgba(255, 255, 255, 0.0)',
+    operator_fill_color: 'rgba(255, 255, 255, 0.0)'
```
timdp, line 43: These two already default to white.

marbles/scripts/index.ts
```diff
@@ -0,0 +1,44 @@
...
+const renderMarble = (path: string, fileName: string): Promise<true> => {
+    const file = readFileSync(join(path, fileName));
+    const diagramSpec = parseMarbleDiagramSpec(file.toString());
+    const { xml: unoptXml, width, height } = renderMarbleDiagram(diagramSpec, { styles });
+    const optimizedSVGPromise = optimizeXml(unoptXml);
+    return optimizedSVGPromise.then((svgXML) => {
+        const svgFileName = fileName.split('.')[0] + '.svg';
+        const svgPath = join(process.cwd(), 'docs_app', 'assets', 'images', 'svgs', svgFileName);
+        writeFileSync(svgPath, svgXML, { encoding: 'utf-8', flag: 'w' });
+        return true;
```
timdp, line 26: I don't think this does anything.

timdp:
Note that Swirly currently introduces a pseudo-circular dependency. In order to parse the ASCII marble diagrams, it depends on the stable `rxjs` to use `TestScheduler.parseMarbles()`. I assume that's not even a public API? Maybe it'd be good to extract it into a separate package, so that both `rxjs` and `swirly` can depend on it?

marbles/scripts/index.ts
niklas-wortmann: oh that's good to know. Thanks

marbles/scripts/index.ts
niklas-wortmann: I use them to make the svg transparent. The color doesn't really matter

marbles/scripts/index.ts
niklas-wortmann: yeah I prefered returning something instead of void, but the value is not really used, that's true

niklas-wortmann:
Good call regarding the circular dependency. I don't know about the `TestScheduler.parseMarbles()` API, but I think as quick fix I could just move the marble diagram generation over to the `docs_app`.

niklas-wortmann:
https://github.com/timdp/swirly/issues/2 Is kinda related to this pr as the generated SVGs could be improved in terms of accessibility

marbles/scripts/index.ts
timdp: Oh, sorry, I missed the alpha channel. SVG actually accepts `transparent` as a color, I think.

timdp:
I've also introduced a workaround for the circular-ish dependency. As of v0.13.6, instead of depending on RxJS directly, the parser now uses a tree-shaken copy of `TreeScheduler.parseMarbles()`. So it's a rudimentary way of extracting the logic into a minimal package, at the expense of duplicating it.

benlesh:
LGTM... @niklas-wortmann ... I'll let you merge this one. You probably want to cherry pick it over to 6.x as well, so I recommend doing a squash and merge.

timdp:
Little bug: in package.json, you're depending on the `swirly` package directly, and pulling in all its dependencies (including Puppeteer) as a result.

niklas-wortmann:
Oh I see, good call. Thanks yeah now that makes sense. I will fix that up

niklas-wortmann:
@benlesh I don't think I will cherry pick it in V6. There's still a lot of work needed for that and I don't want to deal with the hassle :D Also it makes the V7 release even better having lots of great docs stuff too :D

timdp:
I'm a big fan of `depcheck` to avoid stuff like this. Could be something for a follow-up PR. :-)

PR #5426, opened 2020-05-12
[docs]: Create Deprecation page per deprecation group

niklas-wortmann:
This is a rough POC to see if it works for our purpose to improve the deprecation messages.

Text might need some improvement/additional information
@benlesh @cartant

### DIFF
--- a/docs_app/content/deprecations/scheduler-argument.md
+++ b/docs_app/content/deprecations/scheduler-argument.md
@@ -1,0 +1,101 @@
+# Scheduler Argument
+
+To limit the API surface of some operators, but also prepare for a [major refactoring in V8](https://github.com/ReactiveX/rxjs/pull/4583), we
+agreed on deprecating the `scheduler` argument from many operators. It solely deprecates those methods where this argument is barely used. So `time` related
+operators, like [`interval`](https://rxjs.dev/api/index/function/interval) are not affected by this deprecation.
+
+To support this transition the [scheduled creation function](/api/index/function/scheduled) was added.
+
+<div class="alert is-important">
+    <span>
+        This deprecation was introduced in **RxJS 6.5** and will become breaking with **RxJS 8**.
+    </span>
+</div>
+
+## Operators affected by this Change
+
+- [from](/api/index/function/from)
+- [of](/api/index/function/of)
+- [merge](/api/index/function/merge)
+- [concat](/api/index/function/concat)
+- [startWith](/api/operators/startWith)
+- [endWith](/api/operators/endWith)
+- [combineLates](/api/index/function/combineLatest)
+
+## How to Refactor
+
+If you use any other operator from the list above and using the `scheduler` argument, you have to three potential refactoring options.
+
+### Refactoring of `of` and `from`
+
+`scheduled` is kinda copying the behavior of `from`. Therefore if you used `from`  with a `scheduler` argument, you can just replace them.
+
+For the `of` creation function you need to this Observable with `scheduled` and instead of passing the `scheduler` argument to `of` pass it to `scheduled`.
+Following code example demonstrate this process.
+
+```ts
+import { of, asyncScheduler, scheduled } from 'rxjs'; 
+
+// Deprecated approach
+of([1,2,3], asyncScheduler).subscribe(x => console.log(x));
+// suggested approach
+scheduled(of([1,2,3]), asyncScheduler).subscribe(x => console.log(x));
+```
+
+### Refactoring of `merge`, `concat`, `startWith` and `endWith`
+
+In case you used to pass a scheduler argument to one of these operators you propably had code like this:
+
+```ts
+import { concat, of, asyncScheduler } from 'rxjs'; 
+
+concat(
+  of('hello '),
+  of('World'),
+  asyncScheduler
+).subscribe(x => console.log(x));
+```
+
+To work around this deprecation you can leverage the [`scheduled`](/api/index/function/scheduled) function.
+
+```ts
+import { scheduled, of, asyncScheduler } from 'rxjs'; 
+import { concatAll } from 'rxjs/operators'
+
+scheduled(
+  [of('hello '), of('World')],
+  asyncScheduler
+).pipe(
+  concatAll()
+).subscribe(x => console.log(x));
+```
+
+You can apply this pattern to refactor deprecated usage of `concat`, `startWith` and `endWith` but do notice that you want to use [mergeAll](/api/operators/mergeAll) to refactor the deprecated usage of `merge`.
+
+### Refactoring `combineLatest`
+
+Unfortunately, there's no point in using `scheduled` with `combineLatest`. You rather want to leverage [`subscribeOn`](/api/operators/subscribeOn) or [`observeOn`](/api/operators/observeOn) depending on your goals. 
+
+E.g. code that used to look like this:
+
+```ts
+import { combineLatest, timer, asyncScheduler } from 'rxjs'; 
+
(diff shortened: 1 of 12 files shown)


src/internal/observable/combineLatest.ts
```diff
@@ -44,19 +44,19 @@ export function combineLatest<O1 extends ObservableInput<any>, O2 extends Observ
 export function combineLatest<O1 extends ObservableInput<any>, O2 extends ObservableInput<any>, O3 extends ObservableInput<any>, O4 extends ObservableInput<any>, O5 extends ObservableInput<any>, O6 extends ObservableInput<any>, R>(v1: O1, v2: O2, v3: O3, v4: O4, v5: O5, v6: O6, resultSelector: (v1: ObservedValueOf<O1>, v2: ObservedValueOf<O2>, v3: ObservedValueOf<O3>, v4: ObservedValueOf<O4>, v5: ObservedValueOf<O5>, v6: ObservedValueOf<O6>) => R, scheduler?: SchedulerLike): Observable<R>;
 
 // With a scheduler (deprecated)
-/** @deprecated Passing a scheduler here is deprecated, use {@link subscribeOn} and/or {@link observeOn} instead */
+/** @deprecated The scheduler argument is deprecated, use {@link subscribeOn} and/or {@link observeOn}. Details: {@link https://rxjs.dev/deprecations/scheduler-argument} */
```
cartant, line 47: Is the content of the `deprecated` tag shown anywhere in the docs? AFAICT, it isn't. If that is the case, can we please remove this `{@link ... }` business, as it just makes the messages harder to read in linter reports and in the VS Code hints.

niklas-wortmann: I changed it for now, but in the long run I would like to have this in the docs as well. 
In the long run I would like to have this in the docs and IntelliJ is supporting @link and @see, so it's mainly just vscode that has several open issues for that:
https://github.com/microsoft/TypeScript/issues/35524
https://github.com/microsoft/TypeScript/issues/16498

As soon as this is supported by VsCode I'd like to change it to @link etc again

docs_app/content/guide/higher-order-observables.md
```diff
@@ -0,0 +1,28 @@
...
+
+```ts
+const fileObservable = urlObservable.pipe(
+   map(url => http.get(url)),
+);
+```
+
+`http.get()` returns an Observable for each URL. Now you have an Observable *of* Observables, a higher-order Observable.
+
+But how do you work with a higher-order Observable? Typically, by _flattening_: by converting a higher-order Obser vable into an ordinary Observable. For example:
```
jakovljevic-mladen, line 13: Hi. There's an extra space: `Obser vable` -> `Observable`

docs_app/content/guide/higher-order-observables.md
```diff
@@ -0,0 +1,28 @@
...
+
+```ts
+const fileObservable = urlObservable.pipe(
+   concatMap(url => http.get(url)),
+);
+```
+
+The Observable returned in the `concatMap` function is usually referred to as a so-called "inner" Observable, while in this context the `urlObservable` is the so-called "outer" Observable.
+
+The [`concatMap()`](/api/operators/concatMap) operator subscribes to each "inner" Observable, buffers all further emissions of the "outer" Observable, and copies all the emitted values until the inner Observable completes, and continues processing the values of the "outer Observable". All of the values are in that way concatenated.  Other useful flattening operators are
```
jakovljevic-mladen, line 23: I'm sorry for bothering as this is probably not important since browsers clear more than one space, but there are two spaces between `concatenated.` and `Other` words.

docs_app/content/guide/higher-order-observables.md
```diff
@@ -0,0 +1,28 @@
...
+
+* [`mergeMap()`](/api/operators/mergeMap) — subscribes to each inner Observable as it arrives, then emits each value as it arrives
```
jakovljevic-mladen, line 25: This bullet is missing the dot at the end of sentence (unlike other two below which are not). Or maybe, since these bullets are continuation of the sentence `Other useful flattening operators are`, a comma `,` or semicolon `;` could be used for `switchMap` and `switchMap`.

docs_app/content/deprecations/scheduler-argument.md
```diff
@@ -0,0 +1,101 @@
...
+
+// Deprecated approach
+of([1,2,3], asyncScheduler).subscribe(x => console.log(x));
+// suggested approach
+scheduled(of([1,2,3]), asyncScheduler).subscribe(x => console.log(x));
+```
+
+### Refactoring of `merge`, `concat`, `startWith` and `endWith`
+
+In case you used to pass a scheduler argument to one of these operators you propably had code like this:
```
jakovljevic-mladen, line 47: `propably` -> `probably`.

cartant:
Fixed some typos that were pointed out above and rebased onto master.

There are some deprecation messages that still include the `{@ ... }` markup that's discussed [here](https://github.com/ReactiveX/rxjs/pull/5426#discussion_r425513689). IMO, it would be best to remove these ATM, as they clutter the output in VS code and on the command line. Apart from that, this LGTM and, IMO, improvements regarding the deprecations need to be incorporated into the current v6 docs ASAP.

cartant:
FWIW, we really should try to get the deprecation messages sorted before this lands:

https://mobile.twitter.com/WenluWang/status/1275458404132073473

We should really fix them in v6 too, as people are going to be using that for a while in Angular, etc.

src/internal/observable/from.ts
```diff
@@ -4,7 +4,7 @@ import { ObservableInput, SchedulerLike, ObservedValueOf } from '../types';
 import { scheduled } from '../scheduled/scheduled';
 
 export function from<O extends ObservableInput<any>>(input: O): Observable<ObservedValueOf<O>>;
-/** @deprecated use {@link scheduled} instead. */
+/** @deprecated The scheduler argument is deprecated, use scheduled. Details: https://rxjs.dev/deprecations/scheduler-argument */
```
benlesh, line 7: Do backticks not work here?

benlesh: I think my concern is this reads like we're saying "Don't use a scheduler, use scheduled" and to people who don't yet know about `scheduled` it might read like "don't schedule, schedule!" haha.

niklas-wortmann: backticks and general md support is IDE specific, we could just rephrase it to 'use the scheduled operator' and I think this way it's pretty clear

docs_app/content/deprecations/scheduler-argument.md
```diff
@@ -0,0 +1,101 @@
...
+```ts
+import { combineLatest, timer, asyncScheduler } from 'rxjs'; 
+import { subscribeOn } from 'rxjs/operators'
+
+combineLatest(
+  [timer(0, 1000), timer(500, 500)],
+).pipe(
+  subscribeOn(asyncScheduler)
+).subscribe(console.log)
+```
```
cartant, line 101: Just realised that there's a `combineAll` operator that's the higher-order version of `combineLatest`, so the equivalent ought to be:

```ts
import { scheduled, of, asyncScheduler } from 'rxjs'; 
import { combineAll } from 'rxjs/operators'
scheduled(
  [timer(0, 1000), timer(500, 500)],
  asyncScheduler
).pipe(
  combineAll()
```

And maybe this example should use the `[of('hello '), of('World')]` sources that were used above? Was there a particular reason for using the timers?

cartant (approved):

PR #5427, opened 2020-05-13
fix: remove noEmit from ts compiler options

ajitsinghkaler:
Typescript compiler does not work well with both no emit and incremental options. Changed no Emit option with emit declarations only so that we can emit the lest amount with the incremantal option.

**Description:**
Typescript compiler does not work well with both noEmit and incremental options so changed it with emitting declarations only so that we generate the least code and take advantage of incremental build

Details: https://github.com/microsoft/TypeScript/issues/33809#issuecomment-589446027 

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,7 +8,7 @@
     "noImplicitReturns": true,
     "moduleResolution": "node",
     "stripInternal": false,
-    "noEmit": true,
+    "emitDeclarationOnly": true,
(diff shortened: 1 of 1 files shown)


kwonoj:
we do not use noemit for actual compilation, base tsconfig is for editor and elses only. https://github.com/ReactiveX/rxjs/blob/master/src/tsconfig.base.json#L4

Issue #5431, opened 2020-05-14
Stop composing custom observable instances in version 8


benlesh:
We are currently jumping through a lot of hoops in order to preserve the type through the operator chain, when that is no longer as relevant. As a hold-over from the dot-chained days, there was an attempt, starting in v5, to maintain the same type through operators by leveraging `lift`. This was done such that this would still work:

```ts
class CustomObservable<T> extends Observable<T> {
  specialMethod(): void { }
}

const custom = new CustomObservable<T>;

custom.filter(fn).map(fn2).mergeMap(fn3).specialMethod()
```

Over the years this has proven to have limited value, and in fact, is flawed in some ways. For example, look at the code we've had for years in order to support "lifting" publish operators in such a way that `connect` composes through.

Fact is, the current implementation doesn't support what it claims, because if you have more than one `publish` operator in the chain, it breaks `connect`. Try this:

This doesn't work:

```ts
interval(100).pipe(
  publish(),
  publish(),
  refCount()
).subscribe(x => console.log(x));
```

And this doesn't work:
```ts
const source = interval(100).pipe(publish(), publish());
source.subscribe(x => console.log(x));
source.connect();
```

This the byproduct of a few things, but chief among them are:

1. `publish()` et al, should not be "pipeable operators".
2. Composing custom types (or even Subject) through the operators via `lift` was a cool experiment, but ultimately was of limited value and required some strange workarounds in all cases.

### For version 8, I'd like to eliminate the composition of types like `ConnectableObservable`, `Subject` and "custom observables" through operators as a goal.

benlesh:
Core Team Meeting Notes: Do it.

fan-tom:
Yeah, recently I was manages to map `BehaviorSubject` to `BehaviorSubject` and it required some boilerplate, because if you just use `pipe`, you get `Observable` back, not `BehaviorSubject`

voliva:
I've been probably late on the `lift` train, but I still don't fully understand what was lift for.

From OP it looks like it was added to keep methods from the prototype chain when applying operators, but as of now, this code (using just JS):

```js
class CustomObservable extends Observable {
    specialMethod() { }
}
const custom = new CustomObservable();
console.log(
  map((x) => x)(custom).specialMethod
);
```

logs undefined, which means that even though `map` is using lift internally, it doesn't seem like it's keeping the original   functions from `CustomObservable` 🤔

PR #5432, opened 2020-05-14
WIP: Add static multicasters

benlesh:
This is another go at adding proposed static multicast functions. Returning ConnectableObservable from operators is/was a mistake.

### TODO:

- [ ] Settle on names (currently, `multicastFrom` (static), `multicastWith` (operator), `publishFrom`, etc)
- [x] Deprecate `publish` operator variants that do not have a supplied selector
- [ ] Add tests
- [ ] Add more docs

### NOTE:

Also adds a test showing a weakness of our current approach with regards to composing custom observable types through operator chains. This could probably get added at another time, but I wanted to be sure to capture it while I was thinking of it, and seeing some of the `ConnectableObservable` code reminded me of the issue.

### To discuss:

- [ ] Names? `xWith` makes sense, but what about `xFrom`?
- [ ] Should we have a static variant for *every* permutation, or just for `multicast`?
- [ ] How frequently do we use these things?

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -759,6 +759,38 @@ describe('Observable.lift', () => {
       });
   });
 
+  // TODO: This test shows a limitation of the library as it is currently implemented
+  // This test should never pass, and it was probably not a great goal to have the one
+  // above this pass.
+  // See issue here: https://github.com/ReactiveX/rxjs/issues/5431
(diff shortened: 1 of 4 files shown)


benlesh:
Related #5431

jgbpercy:
Related #3833

benlesh:
Closing in favor of #5634

Issue #5372, opened 2020-04-02
Non-type-related issues for v7 beta


cartant:
This issue is for tracking non-type-related issues - type-related issues are in https://github.com/ReactiveX/rxjs/issues/5066 - that could/should be addressed in the v7 beta.

* [x] Use `hostReportError` for swallowed errors - https://github.com/ReactiveX/rxjs/issues/5370 (From @benlesh: Here we opted to add a skipped test that we can "unskip" when we get to the larger refactor in v8)
* [x] `console.warn` still seems to be used instead of `hostReportError` - see [this comment](https://github.com/ReactiveX/rxjs/pull/5676#pullrequestreview-474139791)
* [x] Several issues related to the behaviour of notification observables that complete - https://github.com/ReactiveX/rxjs/issues/3665, https://github.com/ReactiveX/rxjs/issues/3853, https://github.com/ReactiveX/rxjs/issues/5379 - these (and the documentation) should be made consistent. I suspect that `audit` has this problem, too. <s>All operators that have a notifier need to be checked. (the only remaining PR on this one is here: https://github.com/ReactiveX/rxjs/pull/5769)</s> Nope, there are a whole bunch of operators that treat `next` and `complete` identically in notifiers/selectors. See #5838 and #5853 (the PR)
* [x] `mergeScan` has strange behaviour upon the completion of the inner observable - [see here](https://github.com/ReactiveX/rxjs/blob/674e540747f65febe8b87bc05f7ed764f0f33245/src/internal/operators/mergeScan.ts#L65) - this should be changed. (PR here: https://github.com/ReactiveX/rxjs/pull/5805)
* [x] `animationFrames` should use the API's `DOMHighResTimeStamp` - https://github.com/ReactiveX/rxjs/issues/5194
* [x] `finalize` is dodgy AF - see https://github.com/ReactiveX/rxjs/issues/5357 and issues linked therein.
* [x] `Subscribable` sources without `Symbol.observable` - see https://github.com/ReactiveX/rxjs/issues/4532
* [x] Scheduler exports - see https://github.com/ReactiveX/rxjs/issues/5421
* [ ] `@internal` - see #5768
[...]

PR #5433, opened 2020-05-14
fix: finalize after unsubscription from source

cartant:
**Description:**

This PR changes the implementation of `finalize` so that the callback passed to the operator is not called until *after* the subscriber unsubscribes from its source. The effect of this is that - with this change - finalization happens in same order as completion and unsubscription.

Specifically, with unsubscription, we can be sure that when a subscription's `unsubscribe` method returns, every source observable in the chain will have been guaranteed to have been unsubscribed.

Without this change, `finalize`'s behaviour is very different to that of unsubsciption.

The problem is that when `finalize` calls its callback, the observable to which it is applied is still subscribed to any sources and any `finalize` operators applied to sources further up the composed chain will not have called their finalization callbacks.

Aside from this behaviour being surprising, it makes `finalize` pretty much useless for resource management. For example, if there are observables that use and release some resource:

```ts
const a = /* whatever */.pipe(finalize(() => /* release some sub-resource */));
const b = /* whatever */.pipe(finalize(() => /* release some sub-resource */));
```

and are composed into another observable:

```ts
const c = merge(a, b).pipe(finalize(() => /* release the resource */);
```

the `finalize` calls will not do what you would expect, as the calls to release the sub-resources in `a` and `b` will happen *after* the release of the resource itself occurs in `c` and - depending upon the resource - that can effect an error.

**Why the call order is weird without this change:**

The order in which the callbacks are called is weird because the `finalize` operator adds the callback to the `Subscription`:

https://github.com/ReactiveX/rxjs/blob/dfa239d41b97504312fa95e13f4d593d95b49c4b/src/internal/operators/finalize.ts#L80-L85

It's added in the `Subscription` constructor, so the callback is the **first** teardown in the subscription. The subscription to the source is going to be added **after** that:

https://github.com/ReactiveX/rxjs/blob/dfa239d41b97504312fa95e13f4d593d95b49c4b/src/internal/operators/finalize.ts#L70-L72

https://github.com/ReactiveX/rxjs/blob/dfa239d41b97504312fa95e13f4d593d95b49c4b/src/internal/Observable.ts#L212-L222

**It's a breaking change:**

Although the re-ordering of the `finalize` calls broke none of the existing tests, this is a breaking change, but it's a change for the better, IMO.

**Related issue (if exists):** See #5372 and issues linked therein.

### DIFF
--- a/spec/operators/finalize-spec.ts
+++ b/spec/operators/finalize-spec.ts
@@ -2,7 +2,7 @@ import { expect } from 'chai';
 import { finalize, map, share } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
-import { of, timer, interval, NEVER } from 'rxjs';
+import { of, timer, interval, NEVER, Observable } from 'rxjs';
 import { asInteropObservable } from '../helpers/interop-helper';
 
 declare const type: Function;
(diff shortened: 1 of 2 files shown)


cartant:
I added [this test](https://github.com/ReactiveX/rxjs/pull/5433/commits/0ba30b85bbc64cfaf323c44d2964bec5005be892) as another example. I doesn't make much sense to me for the `finalize` callback to be called before the source observable is torn down.

cartant:
@benlesh

> The fix for this should be as simple as moving the addition of the callback to the operator call instead of in the constructor.

Yeah, that's much better. Thanks.

PR #5434, opened 2020-05-14
chore: update to TS 3.9

ajitsinghkaler:
typecript 3.9 was released, updated the repo typescript version to 3.9

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,7 +144,7 @@
     "tslint-etc": "1.9.2",
     "tslint-no-toplevel-property-access": "0.0.2",
     "tslint-no-unused-expression-chai": "0.0.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -144,7 +144,7 @@
     "tslint-etc": "1.9.2",
     "tslint-no-toplevel-property-access": "0.0.2",
     "tslint-no-unused-expression-chai": "0.0.3",
-    "typescript": "~3.8.3",
+    "typescript": "^3.9.2",
```
cartant, line 147: This needs to remain a tilde semver range. It should not be changed to a caret range.

ajitsinghkaler: @cartant made the changes please have a look

Issue #5421, opened 2020-05-10
Make schedulers compatible with babel-plugin-transform-imports


hlehmann:
## Feature Request

All methods are compatible with `babel-plugin-transform-imports` except schedulers due to their renaming.

```
# index.js
export { asap as asapScheduler } from './internal/scheduler/asap';
export { async as asyncScheduler } from './internal/scheduler/async';
export { queue as queueScheduler } from './internal/scheduler/queue';
export { animationFrame as animationFrameScheduler } from './internal/scheduler/animationFrame';
``` 

`babel-plugin-transform-imports` allow some build/dev time improvements but require to keep the same name for reexported variables (or to use `default`) and that's the case for almost all exports.

Would it be possible in future version to avoid this renaming ?

cartant:
Sounds reasonable to me.

cartant:
> Would it be possible in future version to avoid this renaming ?

In version 7, both names are exported:

https://github.com/ReactiveX/rxjs/blob/b50926274d748c563b8371fa46e80f68250b762e/src/internal/scheduler/asap.ts#L39-L44

https://github.com/ReactiveX/rxjs/blob/b50926274d748c563b8371fa46e80f68250b762e/src/index.ts#L16

I'm presuming that doing so will address this issue, so I'm closing it.

PR #5435, opened 2020-05-14
Add ts-api-guardian and golden files

benlesh:
- Had to remove aliases, as `ts-api-guardian` does not support those.
- generates first golden files
- Adds testing and ts-api-guardian check to prepublish

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7171,9 +7171,9 @@
       "dev": true
     },
     "typescript": {
-      "version": "3.8.3",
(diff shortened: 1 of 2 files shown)


benlesh:
To get a diff with v6, this work will need to be replicated on the v6 branch. But I want to make sure this is right first.

benlesh:
This should close #5421 as well. Because I had to remove aliases.

benlesh:
@cartant that was my mistake. I meant to just check in the api_guard directory. :(

benlesh:
Moved: https://github.com/ReactiveX/rxjs/pull/5493

PR #5438, opened 2020-05-15
fix(animationFrames): emit the timestamp from the rAF's callback

### LINKED ISSUE #5194: animationFrames should emit the timestamp from requestAnimationFrame's callback
## Bug Report
Opening a bug report, per @cartant [here](https://github.com/ReactiveX/rxjs/issues/2935#issuecomment-568164054).

**Current Behavior**
Why doesn't the `animationFrames` Observable -- added in #5021 -- pass the DOMHighResTimeStamp that the requestAnimationFrame callback receives, to its subscribers? According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame):

> When callbacks queued by requestAnimationFrame() begin to fire multiple callbacks in a single frame, each receives the same timestamp even though time has passed during the computation of every previous callback's workload.

This is critical when running multiple rAF loops to keep them in sync. The current default implementation, using `Date.now()`, will emit slightly different values for each loop and to each subscriber.

**Reproduction**
Comparison of using two Observables with the rAF timestamp, then two uses of Date in animationFrames. The two timestamps always show the same elapsed ms, while the two Dates will (randomly) show different elapsed ms.
https://codepen.io/kirk-l/pen/BayZQyK

**Expected behavior**
Pass the timestamp from `requestAnimationFrame`'s callback to subscribers.

**Environment**
- Runtime: browser
- RxJS version: 7.0.0-alpha

**Possible Solution**
Here's the implementation I use:

```javascript
const animationFrames = new Observable(subscriber => {
    let requestId = window.requestAnimationFrame(function step(timestamp) {
        subscriber.next(timestamp);
        if (!subscriber.closed) {
            requestId = window.requestAnimationFrame(step);
        }
    });
    return () => window.cancelAnimationFrame(requestId);
});

// And to get the elapsed milliseconds:
const elapsedMs = defer(() => {
    let start = null;
    return animationFrames.pipe(
        tap(timestamp => start === null && (start = timestamp)),
        map(timestamp => timestamp - start),
    );
});
```

I think this would be a better animationFrames Observable.

tmair:
**Description:**
Emit the timestamp provided by rAF to align better with the rAF
specification. Multiple scheduled rAF callbacks within the same frame
will be called with the same timestamp.

This fixes #5194 

**Further notes**
I kept the possibility to supply a custom `TimestampProvider` to the `animationFrames` function.

### DIFF
--- a/src/internal/observable/dom/animationFrames.ts
+++ b/src/internal/observable/dom/animationFrames.ts
@@ -71,7 +71,7 @@ import { requestAnimationFrameProvider } from '../../scheduler/requestAnimationF
  * const source$ = animationFrames(customTSProvider);
  *
  * // Log increasing numbers 0...1...2... on every animation frame.
- * source$.subscribe(({elapsed}) => console.log(elapsed));
+ * source$.subscribe(({ elapsed }) => console.log(elapsed));
  * ```
  *
  * @param timestampProvider An object with a `now` method that provides a numeric timestamp
@@ -86,15 +86,30 @@ export function animationFrames(timestampProvider?: TimestampProvider) {
  */
 function animationFramesFactory(timestampProvider?: TimestampProvider) {
   const { schedule } = requestAnimationFrameProvider;
-  return new Observable<{timestamp: number, elapsed: number}>(subscriber => {
+  return new Observable<{ timestamp: number, elapsed: number }>(subscriber => {
     let subscription: Subscription;
-    let start: DOMHighResTimeStamp | number | null = null;
+    // If no timestamp provider is specified, use performance.now() - as it
+    // will return timestamps 'compatible' with those passed to the run
+    // callback and won't be affected by NTP adjustments, etc.
+    const provider = timestampProvider || performance;
+    // Capture the start time upon subscription, as the run callback can remain
+    // queued for a considerable period of time and the elapsed time should
+    // represent the time elapsed since subscription - not the time since the
+    // first rendered animation frame.
+    const start = provider.now();
     const run = (timestamp: DOMHighResTimeStamp | number) => {
-      const currentTimestamp = timestampProvider ? timestampProvider.now() : timestamp;
-      if (start === null) {
-        start = currentTimestamp;
-      }
-      subscriber.next({timestamp: currentTimestamp, elapsed: currentTimestamp - start});
+      // Use the provider's timestamp to calculate the elapsed time. Note that
+      // this means - if the caller hasn't passed a provider - that
+      // performance.now() will be used instead of the timestamp that was
+      // passed to the run callback. The reason for this is that the timestamp
+      // passed to the callback can be earlier than the start time, as it
+      // represents the time at which the browser decided it would render any
+      // queued frames - and that time can be earlier the captured start time.
+      const now = provider.now();
+      subscriber.next({
+        timestamp: timestampProvider ? now : timestamp,
+        elapsed: now - start
+      });
       if (!subscriber.closed) {
         subscription = schedule(run);
       }
@@ -110,4 +125,4 @@ function animationFramesFactory(timestampProvider?: TimestampProvider) {
  * In the common case, where the timestamp provided by the rAF API is used,
  * we use this shared observable to reduce overhead.
  */
-const DEFAULT_ANIMATION_FRAMES = animationFramesFactory(dateTimestampProvider);
+const DEFAULT_ANIMATION_FRAMES = animationFramesFactory();
--- a/src/internal/observable/dom/animationFrames.ts
+++ b/src/internal/observable/dom/animationFrames.ts
@@ -76,7 +76,7 @@ import { requestAnimationFrameProvider } from '../../scheduler/requestAnimationF
  *
  * @param timestampProvider An object with a `now` method that provides a numeric timestamp
  */
-export function animationFrames(timestampProvider: TimestampProvider = dateTimestampProvider) {
+export function animationFrames(timestampProvider?: TimestampProvider) {
   return timestampProvider === dateTimestampProvider ? DEFAULT_ANIMATION_FRAMES : animationFramesFactory(timestampProvider);
 }
 
--- a/src/internal/observable/dom/animationFrames.ts
+++ b/src/internal/observable/dom/animationFrames.ts
@@ -77,7 +77,7 @@ import { requestAnimationFrameProvider } from '../../scheduler/requestAnimationF
  * @param timestampProvider An object with a `now` method that provides a numeric timestamp
  */
 export function animationFrames(timestampProvider?: TimestampProvider) {
-  return timestampProvider === dateTimestampProvider ? DEFAULT_ANIMATION_FRAMES : animationFramesFactory(timestampProvider);
+  return timestampProvider ? animationFramesFactory(timestampProvider) : DEFAULT_ANIMATION_FRAMES;
 }
 
 /**
@@ -103,7 +103,7 @@ function animationFramesFactory(timestampProvider: TimestampProvider) {
 }
 
 /**
- * In the common case, where `Date` is passed to `animationFrames` as the default,
+ * In the common case, where the timestamp provided by the rAF API is used,
  * we use this shared observable to reduce overhead.
  */
 const DEFAULT_ANIMATION_FRAMES = animationFramesFactory(dateTimestampProvider);
--- a/src/internal/observable/dom/animationFrames.ts
+++ b/src/internal/observable/dom/animationFrames.ts
@@ -1,7 +1,7 @@
 import { Observable } from '../../Observable';
 import { Subscription } from '../../Subscription';
 import { TimestampProvider } from "../../types";
-import { dateTimestampProvider } from '../../scheduler/dateTimestampProvider';
+import { performanceTimestampProvider } from '../../scheduler/performanceTimestampProvider';
 import { requestAnimationFrameProvider } from '../../scheduler/requestAnimationFrameProvider';
 
 /**
@@ -91,7 +91,7 @@ function animationFramesFactory(timestampProvider?: TimestampProvider) {
     // If no timestamp provider is specified, use performance.now() - as it
     // will return timestamps 'compatible' with those passed to the run
     // callback and won't be affected by NTP adjustments, etc.
-    const provider = timestampProvider || performance;
+    const provider = timestampProvider || performanceTimestampProvider;
     // Capture the start time upon subscription, as the run callback can remain
     // queued for a considerable period of time and the elapsed time should
     // represent the time elapsed since subscription - not the time since the
--- a/src/internal/observable/dom/animationFrames.ts
+++ b/src/internal/observable/dom/animationFrames.ts
@@ -7,8 +7,8 @@ import { requestAnimationFrameProvider } from '../../scheduler/requestAnimationF
 /**
  * An observable of animation frames
  *
- * Emits the the amount of time elapsed since subscription on each animation frame. Defaults to elapsed
- * milliseconds. Does not end on its own.
+ * Emits the the amount of time elapsed since subscription and the timestamp on each animation frame.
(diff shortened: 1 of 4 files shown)


src/internal/observable/dom/animationFrames.ts
```diff
@@ -77,20 +77,23 @@ export interface TimestampProvider {
  *
  * @param timestampProvider An object with a `now` method that provides a numeric timestamp
  */
-export function animationFrames(timestampProvider: TimestampProvider = Date) {
+export function animationFrames(timestampProvider?: TimestampProvider) {
   return timestampProvider === Date ? DEFAULT_ANIMATION_FRAMES : animationFramesFactory(timestampProvider);
```
cartant, line 81: The default should be changed to **not** use `Date` as the timestamp provider, IMO. The purpose of the default is to avoid creating observable instances when `animationFrames` is called with the expected/default timestamp provider and the default is to use the timestamp provided by the API.

src/internal/observable/dom/animationFrames.ts
```diff
@@ -77,20 +77,23 @@ export interface TimestampProvider {
...
     let id: number;
-    const start = timestampProvider.now();
-    const run = () => {
-      subscriber.next(timestampProvider.now() - start);
+    let start: DOMHighResTimeStamp | number | null = null;
+    const run = (timestamp: DOMHighResTimeStamp | number) => {
+      if (start === null) {
+        start = timestampProvider ? timestampProvider.now() : timestamp;
+      }
+      subscriber.next((timestampProvider ? timestampProvider.now() : timestamp) - start);
```
cartant, line 96: If the timestamps provided by the API are being used, IMO they should be passed through without modification.

That is, the `start` should only be subtracted when using a caller-specified timestamp provider.

AFAICT, the point of using the API-provider timestamp is to synchronize animations and if an elapsed time is calculated from the `start` time, synchronization would only be possible between animations that started on the same animation frame. This seems like an unnecessary restriction, to me.

@benlesh What is your opinion on this? When should the operator emit an elapsed time: always, never or only if a caller-specified timestamp provider is specified?

benlesh: I see. So you'd have users calculate the elapsed time themselves?

The problem I see is they would then need to capture the initial time themselves. Maybe we can provide both?  `{ timestamp: number, elapsed: number }`

benlesh: Then again, if they already have the timestamp provider, it's trivial to get the current timestamp within whatever operation you're performing. But less trivial to know the recipe to get the elapsed time, which is more useful in animations.

cartant: I think providing both - `{ timestamp: number, elapsed: number }` - would be the best. It'll cover all bases and is essentially self-documenting. Thanks.

src/internal/observable/dom/animationFrames.ts
tmair: Yes, I did forget to change the default here.

benlesh:
I'm not convinced that this is a solid change. With this change, we are back to world where people need to know a recipe in order to get the time that has elapsed since the beginning of the observable. Which is what you usually need when you're doing animations. That, in the face of the fact that it's as simple as calling `Date.now()` anywhere in your code to get the current timestamp.

I suppose as a compromise we could add an operator provided the elapsed time. But that's just one more thing, right? I'm not sure I like that either.

I'm not saying I'm against this change. I'm just saying that I'm not convinced.

cartant:
> With this change, we are back to world where people need to know a recipe in order to get the time that has elapsed since the beginning of the observable.

I'm not sure what's meant be 'recipe' here. If the value emitted from the rAF observable has both the `timestamp` and `elapsed` properties, isn't the latter obvious and immediately available in the `next` callback?

benlesh:
Ah. You know what, I was reviewing this from my phone. I missed that. My mistake. I'll think about it some more.

src/internal/observable/dom/animationFrames.ts
```diff
@@ -72,25 +72,29 @@ export interface TimestampProvider {
...
+  return new Observable<{timestamp: number, elapsed: number}>(subscriber => {
     let id: number;
-    const start = timestampProvider.now();
-    const run = () => {
-      subscriber.next(timestampProvider.now() - start);
+    let start: DOMHighResTimeStamp | number | null = null;
+    const run = (timestamp: DOMHighResTimeStamp | number) => {
+      const currentTimestamp = timestampProvider ? timestampProvider.now() : timestamp;
+      if (start === null) {
+        start = currentTimestamp;
```
benlesh, line 95: Okay, the only issue I have with this now is that the `start` time is now non-deterministic. I think collection of `start` should be moved back to where it was. This is no longer the "start" it's the first animation frame.

Consider the following scenario: A user triggers an animation that is supposed to take 30 seconds, then, before the animation frame fires they switch tabs. If the start isn't collected until the first animationFrame, when the user tabs back, they'll be forced to watch the full animation.

If the `start` is collected when the observable is subscribed to, when they come back, the animation will be done as expected.

I understand that maybe this change probably arose from the confusion about it being a DOMHighResTimeStamp, and the code was previously using `Date` as default timestamp provider. Instead, it's safe to use `performance` as the default TimestampProvider, as it is supported by all browsers and platforms we currently support.

benlesh:
@cartant We might want to consider updating `TestScheduler.run` to temporarily monkey patch known TimestampProviders like Date and Performance. I think we're going to end up breaking people's tests, including maybe our own. I'll open an issue.

benlesh:
We might just have to block on this one, until we iron on #5475 ... Sorry again, @tmair! I'll add this to the next team meeting agenda so we can discuss it.

benlesh:
(Really, really sorry, haha... I promise I'm not intentionally trying to be a PITA on this one)

tmair:
> (Really, really sorry, haha... I promise I'm not intentionally trying to be a PITA on this one)

No worries. I totally understand why these things are problematic. I will try to update the PR soon.

cartant:
@tmair I don't think there's anything that really needs to be updated in this PR, ATM. The `animationFrame` observable itself is blocked because it's not testable in user code. What's being looked into, ATM, is the best way of addressing that issue. Once that's resolved, we can look again at this PR.

joshribakoff:
Keep in mind:

> The callback method is passed a single argument, a DOMHighResTimeStamp, which indicates the current time (based on the number of milliseconds since time origin). When callbacks queued by requestAnimationFrame() begin to fire multiple callbacks in a single frame, each receives the same timestamp even though time has passed during the computation of every previous callback's workload. This timestamp is a decimal number, in milliseconds, but with a minimal precision of 1ms (1000 µs).

Sampling the time with performance.now beats using Date.now, but can still cause animations to be out of sync from each other in a give animation frame, which can affect the animation & it make it look not smooth. Unless the time is sampled once per animation frame, which RAF's callback provided DOMHighResTime already handles. If we're sampling a timestampProvider, we should ideally add the same affordance the native RAF callback has.

src/internal/observable/dom/animationFrames.ts
```diff
@@ -86,15 +86,30 @@ export function animationFrames(timestampProvider?: TimestampProvider) {
...
+      // Use the provider's timestamp to calculate the elapsed time. Note that
+      // this means - if the caller hasn't passed a provider - that
+      // performance.now() will be used instead of the timestamp that was
+      // passed to the run callback. The reason for this is that the timestamp
+      // passed to the callback can be earlier than the start time, as it
+      // represents the time at which the browser decided it would render any
+      // queued frames - and that time can be earlier the captured start time.
+      const now = provider.now();
+      subscriber.next({
+        timestamp: timestampProvider ? now : timestamp,
```
benlesh, line 110: could just call the provider here, to avoid calling it when you didn't have to.

src/internal/observable/dom/animationFrames.ts
```diff
@@ -86,15 +86,30 @@ export function animationFrames(timestampProvider?: TimestampProvider) {
...
+    // If no timestamp provider is specified, use performance.now() - as it
+    // will return timestamps 'compatible' with those passed to the run
+    // callback and won't be affected by NTP adjustments, etc.
+    const provider = timestampProvider || performance;
+    // Capture the start time upon subscription, as the run callback can remain
+    // queued for a considerable period of time and the elapsed time should
+    // represent the time elapsed since subscription - not the time since the
+    // first rendered animation frame.
+    const start = provider.now();
     const run = (timestamp: DOMHighResTimeStamp | number) => {
```
benlesh, line 100: Note that this still doesn't really solve the problem that this is untestable in TestScheduler run mode, which automatically patches our schedulers.

benlesh: I think the solution here is going to be creating a lightweight wrapper around rAF, such that we can override it in `run` mode. That's what the animationFrameScheduler is _supposed_ to be, but it's inadequate, because it doesn't provide the timestamp, and our schedulers in general are perhaps a bit overkill for this use case.

Something like this should suffice:

```ts
export const rAF = {
   schedule(callback: (timestamp: number) => void): Subscription {
     const id = this.impl(callback);
     return new Subscription(() => cancelAnimationFrame(id));
   },
   impl: requestAnimationFrame,
}
```

Then, in our `run` mode, we just need to pull that in and patch the `impl` to override it's behavior where we are using it.

We could also just monkey patch rAF directly, but that has other implications I'm not keen on. 

We also need some sort of way to declare when animation frames should be fired in a test. Probably another, idempotent helper that we add to the context passed to `TestScheduler.run`. like:

```
rxTest.run(({ cold, expectObservable, animationFrames }) => {
  animationFrames('-----x----x----x----x---x-----');
  // rest of the test here
});
```

src/internal/observable/dom/animationFrames.ts
cartant: You can't actually do that without effecting negative durations, as the passed timestamp can be before the captured start time.

cartant:
@benlesh To separate it from the annoying weirdness of the `rAF` callback timestamp business, I'm going to put the run-mode-patchable providers into a separate PR - referenced above - that targets [this issue](https://github.com/ReactiveX/rxjs/issues/5475). I'll come back to this once the providers have been reviewed and merged.

src/internal/observable/dom/animationFrames.ts
benlesh: you mean to say that this can produce negative numbers?

```
const start = performance.now();
requestAnimationFrame((ts) => console.log(ts - start)); ?
```

benlesh: Or are you saying that _this_ will produce a negative number?

```ts
const start = performance.now();
requestAnimationFrame(() => console.log(performance.now() - start));
```

benlesh: Nevermind, I missed that you were using it below in `elapsed` as well.

benlesh:
Okay, LGTM, in that I think it's a good iteration, but we have failing tests now that need cleared up.

benlesh:
@cartant Just because I wasn't being clear. I'm good with these changes, the only thing stopping the merge is getting the tests working.

cartant:
@benlesh AFAICT, there are no test failures here. The problem is that the API guardian is failing because of the `1 | 0 | -1` business. I will have another look, but that was what appeared to be the problem and that prompted me to open #5608 - which I don't understand (in light of OJ's comment) and don't know how to fix.

cartant:
@benlesh FWIW, making these tests pass - before #5607 is merged - requires too much effort. `performance.now()` is delegated and stubbed in #5607 and it isn't here. Basically, it will be easier, IMO, to fix the tests in this PR once #5607 is merged.

benlesh:
I guess just mark the tests as skipped for now

cartant:
@benlesh

> I guess just mark the tests as skipped for now

Donesies

Issue #5439, opened 2020-05-15
groupBy should support typeguards


benlesh:
I'm not even sure this is possible, but it would be great if this "just worked":

```ts
function isNumber(input: any): input is number {
   return typeof input === 'number';
}

of('a', 1, 'b', 2).pipe(
   groupBy(isNumber),
   mergeMap(group => {
     if (group.key) {
         return group; // Observable<number>
     } else {
         return group; // Observable<string>
     }
   });
)
```

cartant:
This sounds like `splitBy` - see #3807 - which I implemented [here](https://github.com/cartant/rxjs-etc/blob/master/source/operators/splitBy.ts) but I didn't open a PR for it - I guess I wasn't entirely happy with it. I can take another look.

cartant:
Hmm, this is a little weird. It would seem that adding this signature should do it:

```ts
  export function groupBy<T, K extends T>(
    keySelector: (value: T) => value is K
  ): OperatorFunction<T, GroupedObservable<true, K> | GroupedObservable<false, T>>;
```

And the `key` property needs to be made `const`:

https://github.com/ReactiveX/rxjs/blob/f5c6332b6098ec5c302f8034ad2273106e301275/src/internal/operators/groupBy.ts#L281

However, the type is only narrowed within the `if` and not the `else` - which is weird:

```ts
of("a", 1, "b", 2).pipe(
  groupBy(isNumber),
  mergeMap(group => {
    if (group.key) {
      return group; // GroupedObservable<true, number>
    } else {
      return group; // GroupedObservable<true, number> | GroupedObservable<false, any>
    }
  })
);
```

PR #5441, opened 2020-05-17
feat(groupBy): Add typeguards support for groupBy

kolodny:
A couple of notes here. I couldn't figure out how to test this since it doesn't work in oldish versions of TS like 2.8 which is where the dtslint tests run. If there was some way to enable/disable for TS versions that would be one solution.

Note that in the example in #5439 you need to test `key === true` in order to get the type checking to work for both the if and else blocks. This may be related to https://github.com/Microsoft/TypeScript/issues/26852

Another item and this is probably a TS issue is that while all 4 of the following usages work, the 5th one, using an inline type guard with no explicit arg type, doesn't work. I'm guessing there's some TS weirdness going on that's probably a bug but I can't explain what is making it behave that way

```ts
function isNumberFromAny(input: any): input is number {
  return typeof input === 'number';
}

function isNumberFromStringOrNumber(input: string | number): input is number {
  return typeof input === 'number';
}

// Works!
of('a', 1, 'b', 2).pipe(
  groupBy((x: any): x is number => typeof x === 'number'),
  mergeMap(group => {
    if (group.key === true) {
        group.subscribe(x => x.toExponential(1)); // x is number here
    } else {
        group.subscribe(x => x.toLowerCase()); // x is string here
    }
    return group;
  })
);

// Works!
of('a', 1, 'b', 2).pipe(
  groupBy((x: string | number): x is number => typeof x === 'number'),
  mergeMap(group => {
    if (group.key === true) {
        group.subscribe(x => x.toExponential(1)); // x is number here
    } else {
        group.subscribe(x => x.toLowerCase()); // x is string here
    }
    return group;
  })
);

// Works!
of('a', 1, 'b', 2).pipe(
  groupBy(isNumberFromAny),
  mergeMap(group => {
    if (group.key === true) {
        group.subscribe(x => x.toExponential(1)); // x is number here
    } else {
        group.subscribe(x => x.toLowerCase()); // x is string here
    }
    return group;
  })
);


// Works!
of('a', 1, 'b', 2).pipe(
  groupBy(isNumberFromStringOrNumber),
  mergeMap(group => {
    if (group.key === true) {
        group.subscribe(x => x.toExponential(1)); // x is number here
    } else {
        group.subscribe(x => x.toLowerCase()); // x is string here
    }
    return group;
  })
);

// THIS IS BUSTED!
of('a', 1, 'b', 2).pipe(
  groupBy((x): x is number => typeof x === 'number'),
  mergeMap(group => {
    if (group.key === true) {
        // group is GroupedObservable<true, string | number>
        group.subscribe(x => x.toExponential(1))
    } else {
        // group is GroupedObservable<false, never>
        group.subscribe(x => x.toLowerCase())
    }
    return group;
  })
);
```

### DIFF
--- a/spec-dtslint/operators/groupBy-spec.ts
+++ b/spec-dtslint/operators/groupBy-spec.ts
@@ -1,5 +1,5 @@
 import { of, Subject, GroupedObservable } from 'rxjs';
-import { groupBy } from 'rxjs/operators';
+import { groupBy, mergeMap } from 'rxjs/operators';
 
 it('should infer correctly', () => {
   const o = of(1, 2, 3).pipe(groupBy(value => value.toString())); // $ExpectType Observable<GroupedObservable<string, number>>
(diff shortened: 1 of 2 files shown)


cartant:
The dtslint tests should be using the version of TS that's installed in `node_modules`. If that's not the case, the config is broken. FWIW, you need to run it using `npm dtslint`.

IDK what happened when I looked quickly at this, on Saturday morning - before I noticed that it has been reassigned - but this now works, for me, in VS Code. That is, hovering over the statements within the `if` and `else` show the expected types. It does not seem to need `group.key === true` and the inline user-defined type guard seems fine.

```ts
import { of, OperatorFunction } from "rxjs";
import { groupBy, mergeMap } from "rxjs/operators";

function isNumber(input: any): input is number {
  return typeof input === "number";
}

declare module "rxjs/internal/operators/groupBy" {
  export function groupBy<T, K extends T>(
    keySelector: (value: T) => value is K
  ): OperatorFunction<
    T,
    GroupedObservable<true, K> | GroupedObservable<false, Exclude<T, K>>
  >;
}

of("a", 1, "b", 2).pipe(
  groupBy(isNumber),
  mergeMap((group) => {
    if (group.key) {
      return group; // GroupedObservable<true, number>
    } else {
      return group; // GroupedObservable<false, any>
    }
  })
);

of("a", 1, "b", 2).pipe(
  groupBy((value): value is number => typeof value === "number"),
  mergeMap((group) => {
    if (group.key) {
      return group; // GroupedObservable<true, number>
    } else {
      return group; // GroupedObservable<false, string>
    }
  })
);
```

PS. the `Exclude` is nice! ❤️

benlesh:
Nice work, guys!

PR #5442, opened 2020-05-17
feat(fetch): add selector (cherry pick #5306 for v6)

cartant:
* feat(fetch): add selector

**Description:**

This PR cherry picks the fix in #5306 for v6.

**Related issue (if exists):** #5306


benlesh:
Looks like the side-effects check just needs to be updated.

cartant:
@benlesh

> Looks like the side-effects check just needs to be updated.

Should be done.

PR #5448, opened 2020-05-20
feat(fromEvent): improve type inference on event name

kevmo314:
This change updates the type of the event name to be covariant with string. This permits using listeners that are defined as:

```
addEventListener(event: 'foo'|'bar', listener: ...)
```

Without this change, the following error is produced because the typings are too tight around `string`:

```
Argument of type 'BrowserWindow' is not assignable to parameter of type 'FromEventTarget<unknown>'.
  Type 'BrowserWindow' is not assignable to type 'JQueryStyleEventEmitter'.
    Types of property 'on' are incompatible.
      Type '{ (event: "always-on-top-changed", listener: (event: Event, isAlwaysOnTop: boolean) => void): BrowserWindow; (event: "app-command", listener: (event: Event, command: string) => void): BrowserWindow; (event: "blur", listener: Function): BrowserWindow; (event: "close", listener: (event: Event) => void): BrowserWindow;...' is not assignable to type '(eventName: string, handler: Function) => void'.
        Types of parameters 'event' and 'eventName' are incompatible.
          Type 'string' is not assignable to type '"always-on-top-changed"'.ts(2345)
```

### DIFF
--- a/src/internal/observable/fromEvent.ts
+++ b/src/internal/observable/fromEvent.ts
@@ -4,9 +4,9 @@ import { isFunction } from '../util/isFunction';
 import { Subscriber } from '../Subscriber';
 import { map } from '../operators/map';
 
-export interface NodeStyleEventEmitter {
-  addListener: (eventName: string | symbol, handler: NodeEventHandler) => this;
-  removeListener: (eventName: string | symbol, handler: NodeEventHandler) => this;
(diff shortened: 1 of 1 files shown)


kevmo314:
Hrm okay, nevermind. It seems that making this change will break a lot of existing interfaces and backwards-compatibility means that the type parameters will be ordered awkwardly...

PR #5449, opened 2020-05-20
fix(defer) Disallow passing () => undefined to observableFactory

### LINKED ISSUE #5446: defer(() => undefined) should be a TS error
Currently the following results in `Observable<never>`, which is very incorrect:

```ts
defer(() => undefined);
```

I would expect a TypeScript error there. As the developer should be returning some form of `ObservableInput` from that function.

kolodny:
fixes #5446

### DIFF
--- a/spec-dtslint/observables/defer-spec.ts
+++ b/spec-dtslint/observables/defer-spec.ts
@@ -44,5 +44,5 @@ it('should infer correctly with function that sometimes error', () => {
 });
 
 it('should infer correctly with functions that sometimes do not return an ObservableInput', () => {
-  const a = defer(() => { if (Math.random() < 0.5) { return of(42); } }); // $ExpectType Observable<number>
+  const a = defer(() => { if (Math.random() < 0.5) { return of(42); } }); // $ExpectError
 });
--- a/spec-dtslint/observables/defer-spec.ts
+++ b/spec-dtslint/observables/defer-spec.ts
@@ -32,6 +32,17 @@ it('should error if function returns never', () => {
   const a = defer(() => { throw new Error(); }); // $ExpectError
 });
 
+
+it('should infer correctly with function that sometimes error', () => {
+  // $ExpectType Observable<number>
+  defer(() => {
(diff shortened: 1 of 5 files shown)


spec-dtslint/observables/defer-spec.ts
```diff
@@ -28,6 +28,10 @@ it('should error if function returns undefined', () => {
   const a = defer(() => undefined); // $ExpectError
 });
 
+it('should error if function returns never', () => {
+  const a = defer(() => { throw new Error(); }); // $ExpectError
+});
```
benlesh, line 33: Does this work?

```ts
defer(() => {
   if (Math.random() > 0.5) {
      throw new Error();
   }
   return of(1, 2, 3);
});
```

I mean, I assume it does, but It would be nice to have a test there.

kolodny: Added

src/internal/observable/defer.ts
```diff
@@ -52,7 +52,9 @@ import { EMPTY } from './empty';
  * @name defer
  * @owner Observable
  */
-export function defer<R extends ObservableInput<any> | void>(observableFactory: () => R): Observable<ObservedValueOf<R>> {
+export function defer<R extends ObservableInput<any> | void>(
+  observableFactory: [R] extends [undefined] | [void] ? never : () => R
+): Observable<ObservedValueOf<R>> {
```
cartant, line 57: IMO, if we are going to fix this footgun, we should only be accepting factories that return a valid `ObservableInput`. I think the fix should be the removal of the `void`:

```ts
export function defer<R extends ObservableInput<any>>(observableFactory: () => R): Observable<ObservedValueOf<R>> { ...
```

And I think the implementation should be changed, too. If it's not, the footgun is still there for non-TS usage.

benlesh: I agree. It's weird that we ever had this work.

kolodny: Done, I've updated the tests and dstlint spec to reflect the changes

Issue #5455, opened 2020-05-26
shareReplay with refCount doesn't restart due to unsubscriptions if the source completes first


josepot:
## Bug Report

**Current Behavior**
`shareReplay` doesn't restart due to unsubscriptions when `refCount` is `true` if the source completes before the subscribers.

**Reproduction**
- [REPL](https://stackblitz.com/edit/typescript-jtwfyk)

```ts
import {interval} from 'rxjs';
import {finalize, shareReplay, take} from 'rxjs/operators'

const sharedSource = interval(100).pipe(
  take(5),
  finalize(() => {
    console.log('The source has completed!')
  }),
  shareReplay({refCount: true, bufferSize: 1}),
);

const s1 = sharedSource.subscribe(x => console.log('subscription 1', x));
const s2 = sharedSource.subscribe(x => console.log('subscription 2', x));
console.log('Two active subscriptions...')

setTimeout(() => {
  s1.unsubscribe();
  s2.unsubscribe();
  console.log('Both subscriptions have been unsubscribed (refCount === 0)')
}, 550)

setTimeout(() => {
  console.log('Creating a new subscription...')
  sharedSource.subscribe(x => console.log('subscription 3', x))
}, 600)
```

**Expected behavior**
I would expect to restart the subscription to the source after the `refCount` has gone down to zero.

**Environment**
- Runtime: Node v12.4.0,
- RxJS version: 6.5.5

cartant:
What you are seeing is the expected behaviour. Once the source completes, that's it. The completion is replayed.

josepot:
Hi @cartant and thanks for your quick reply.

> What you are seeing is the expected behaviour. Once the source completes, that's it. The completion is replayed. If you don't want this behaviour, use share instead.

I _really_ thought that this was a bug. Please notice that in the [PR that I've opened](#5456 ) all the existing tests are passing and I also added a new one to cover this case.

One of the reasons why I thought that this was a bug is that currently there is a test for `shareReplay` that states: 

> should restart due to unsubscriptions if refCount is true

I don't understand why that shouldn't be the case if the source has completed and all the subscribers have unsubscribed. I thought that was the point of using `refCount`. If that's not the expected behavior perhaps would it be worth it to add a test that makes sure that once the source has completed then the `refCount` is ignored?

Also, I know how to create my own operator with this behavior, but is there a way that I can get this behavior using `share` or `publishReplay`... because I can't think of a way to get that.

Thanks a lot for looking into this.

[...]

PR #5457, opened 2020-05-26
test(shareReplay): don't restart when refCount if the source has completed

josepot:
**Description:**
As discussed in #5455: it would be useful to have a test that removes the ambiguity on what should happen with the `shareReplay` operator if `refCount` is `true` once the source completes.

I've also added a second commit that gets rid of the redundant variable `isComplete`. This variable became unnecessary after [this change](https://github.com/josepot/rxjs/commit/35e600fb031ba2814d65b586ca0664af85b179ba#diff-a3e23feeea8613de90d70844292403e3). I guess that this second commit is a "nit pick", so I don't mind removing it from this PR. Although, it may marginally reduce the bundle size?

**Related issue**
#5455

### DIFF
--- a/spec/operators/shareReplay-spec.ts
+++ b/spec/operators/shareReplay-spec.ts
@@ -175,6 +175,7 @@ describe('shareReplay operator', () => {
 
   it('should not restart due to unsubscriptions if refCount is false', () => {
     const source = cold('a-b-c-d-e-f-g-h-i-j');
+    const sourceSubs =  '^------------------';
     const sub1 =        '^------!';
     const expected1 =   'a-b-c-d-';
     const sub2 =        '-----------^-------';
@@ -184,10 +185,14 @@ describe('shareReplay operator', () => {
 
     expectObservable(shared, sub1).toBe(expected1);
     expectObservable(shared, sub2).toBe(expected2);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
   });
 
   it('should restart due to unsubscriptions if refCount is true', () => {
+    const sourceSubs = [];
     const source = cold('a-b-c-d-e-f-g-h-i-j');
+    sourceSubs.push(    '^------!----------------------');
+    sourceSubs.push(    '-----------^------------------');
     const sub1 =        '^------!';
     const expected1 =   'a-b-c-d-';
     const sub2 =        '-----------^------------------';
@@ -197,6 +202,7 @@ describe('shareReplay operator', () => {
 
     expectObservable(shared, sub1).toBe(expected1);
     expectObservable(shared, sub2).toBe(expected2);
+    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
   });
 
   it('should not restart due to unsubscriptions if refCount is true when the source has completed', () => {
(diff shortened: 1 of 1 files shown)


spec/operators/shareReplay-spec.ts
```diff
@@ -199,6 +199,19 @@ describe('shareReplay operator', () => {
...
+    const source = cold('a-(b|)         ');
+    const sub1 =        '^------!       ';
+    const expected1 =   'a-(b|)         ';
+    const sub2 =        '-----------^!  ';
+    const expected2 =   '-----------(b|)';
+
+    const shared = source.pipe(shareReplay({ bufferSize: 1, refCount: true }));
+
+    expectObservable(shared, sub1).toBe(expected1);
+    expectObservable(shared, sub2).toBe(expected2);
```
cartant, line 212: We also need to test that there is only one subscription to the `source`. That can be done with `expectSubscriptions` which takes an expected subscription marble diagram - or an array of diagrams if multiple subscriptions are expected.

Something like this:

```ts
    const source     = cold('a-(b|)         ');
    const sourceSubs =      '^-!            ';
    /* ... */
```

If there multiple subscriptions to the source are made, this expectation would fail, as `sourceSubs` would have to be an array of subscriptions.

josepot: Yep. That makes a lot of sense. I've amended that commit with this expectation, thanks!

I've also added another commit that does the same thing on [other tests that were not doing that](https://github.com/ReactiveX/rxjs/pull/5457/commits/2863c9d792859dd01b68498afbe9c353dc32eaf0). I do understand that for those tests that expectation may not be as important, because the current expectations are "implicitly" checking for that already. So, I don't mind dropping this commit from the PR. I thought that maybe for consistency it could be worth it to have those expectations?

benlesh:
What's not immediately clear to me from this PR is whether or not this is a breaking change. It doesn't _look_ like it is, but the only way to know that for sure would have been to land the passing test, _then_ land the change proving it didn't break the test.

cartant:
@benlesh There should be no behavioural changes in this whatsoever. The PR adds a test that describes behaviour that was not fully characterised by the docs. Namely that the completion should be replayed regardless of the `refCount` option. The removal of the unnecessary `isComplete` is an incidental clean up, so to speak.

The original PR - which was closed - sought to change the behaviour.

cartant:
@josepot Maybe you should remove the commit that removes the `isComplete` variable from this PR and then open another with that change?

josepot:
Yep, no worries. I've dropped the commit that removes the `isComplete` variable. @benlesh has a point, for sure. I mean, if I had a way to run the CI on the first and the last commit, then I guess that it would have been _ok_ to squish the commit that removed the `isComplete` varialbe in the middle

cartant:
@benlesh I merged this, as the code modification - the removal of the unnecessary variable - was removed, so this PR now only adds a test and some some additional expectations.

cartant:
@josepot Feel free to open another PR that removes that unnecessary variable. If it's not needed and the code's simpler without it, it might as well be deleted.

PR #5465, opened 2020-06-01
null out _unsubscribe after unsubscription

### LINKED ISSUE #5464: Subscription callback is not nulled out when Subscription is unsubscribed
## Bug Report

**Current Behavior**
When instantiating a Subscription and passing a callback, that callback is saved as `this._unsubscribe` of the `Subscription` object:

https://github.com/ReactiveX/rxjs/blob/d608cf33791cceb5f330270da96b0b2805223208/src/internal/Subscription.ts#L43

When unsubscribing that Subscription, the function is called, but not nulled out afterwards:

https://github.com/ReactiveX/rxjs/blob/d608cf33791cceb5f330270da96b0b2805223208/src/internal/Subscription.ts#L79

This means that if it did some manual resource cleanup like closing a MessagePort, that MessagePort still can't be garbage-collected, because the callback, and therefor the Subscription, still holds a reference to it.

This is different from Subscriptions added via `add()`, which are correctly de-referenced:

https://github.com/ReactiveX/rxjs/blob/d608cf33791cceb5f330270da96b0b2805223208/src/internal/Subscription.ts#L66

**Reproduction**
- REPL or Repo link:
(you can use https://stackblitz.com/ to create one to attach here)

```js
const { port1 } = new MessageChannel()
const subscription = new Subscription(() => {
  port1.close()
})
subscription.unsubscribe()
console.log(subscription._unsubscribe)
```

**Expected behavior**
After calling the callback, `this._unsubscribe` should be nulled out so the callback and its references can be garbage-collected. Since a Subscription can only be unsubscribed once, there is no need to keep the reference.

**Environment**
- Runtime: All
- RxJS version: 6.5.5

**Possible Solution**
Set `this._unsubscribe` to `null` when `unsubscribe()` is called.

felixfbecker:
**Description:**
`null` out the `unsubscribe` callback after unsubscribing.

The `any` is kind of ugly but that is also how it's assigned in the constructor.

**Related issue (if exists):** Fixes #5464

### DIFF
--- a/src/internal/Subscription.ts
+++ b/src/internal/Subscription.ts
@@ -39,16 +39,7 @@ export class Subscription implements SubscriptionLike {
    */
   constructor(unsubscribe?: () => void) {
     if (unsubscribe) {
-      // Wrap the unsubscribe teardown in a function so that the argument can
-      // be nulled. It's not possible to null the _unsubscribe member as there
-      // are many classes that are derived from Subscriber (which derives from
-      // Subscription) that implement an _unsubscribe method as a mechanism for
-      // obtaining unsubscription notifications and some of those subscribers
-      // are recycled.
-      (<any> this)._unsubscribe = () => {
-        unsubscribe!();
-        unsubscribe = undefined;
-      };
+      (this as any)._unsubscribe = unsubscribe;
     }
   }
 
@@ -65,13 +56,20 @@ export class Subscription implements SubscriptionLike {
       return;
     }
 
-    let { _parentOrParents, _unsubscribe, _subscriptions } = (<any> this);
+    let { _parentOrParents, _unsubscribe, _subscriptions } = (this as any);
 
     this.closed = true;
     this._parentOrParents = null;
     // null out _subscriptions first so any child subscriptions that attempt
     // to remove themselves from this subscription will noop
     this._subscriptions = null;
+    // It's not possible to null the _unsubscribe member as there are many
+    // classes that are derived from Subscriber (which derives from
+    // Subscription) that implement an _unsubscribe method as a mechanism for
+    // obtaining unsubscription notifications and some of those subscribers are
+    // recycled. Deleting the member will release the reference to any teardown
+    // functions passed in the constructor and will leave any methods intact.
+    delete (this as any)._unsubscribe;
 
     if (_parentOrParents instanceof Subscription) {
       _parentOrParents.remove(this);
@@ -139,7 +137,7 @@ export class Subscription implements SubscriptionLike {
   add(teardown: TeardownLogic): Subscription {
     let subscription = (<Subscription>teardown);
 
-    if (!(<any>teardown)) {
+    if (!teardown) {
       return Subscription.EMPTY;
     }
 
--- a/src/internal/Subscription.ts
+++ b/src/internal/Subscription.ts
@@ -39,7 +39,16 @@ export class Subscription implements SubscriptionLike {
    */
   constructor(unsubscribe?: () => void) {
     if (unsubscribe) {
-      (<any> this)._unsubscribe = unsubscribe;
+      // Wrap the unsubscribe teardown in a function so that the argument can
+      // be nulled. It's not possible to null the _unsubscribe member as there
+      // are many classes that are derived from Subscriber (which derives from
+      // Subscription) that implement an _unsubscribe method as a mechanism for
+      // obtaining unsubscription notifications and some of those subscribers
+      // are recycled.
+      (<any> this)._unsubscribe = () => {
+        unsubscribe!();
+        unsubscribe = undefined;
+      };
     }
   }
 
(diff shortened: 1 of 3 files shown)


felixfbecker:
Okay this seems to break tests, don't know enough to proceed. I'd appreciate if a maintainer could look at this bug 🙇

cartant:
Could you add a test to show/prove that multiple `unsubscribe` calls are okay and that overwriting the [`_unsubscribe` method in `Subscriber`](https://github.com/ReactiveX/rxjs/blob/d608cf33791cceb5f330270da96b0b2805223208/src/internal/Subscriber.ts#L296) with `null` - which is what this change will end up doing - behaves as expected?

cartant:
> Okay this seems to break tests, don't know enough to proceed. I'd appreciate if a maintainer could look at this bug.

It's likely related to my previous comment. I'll have a look at it. Thanks for the PR and the bug report.

felixfbecker:
@benlesh would you mind giving this a review when you have the chance? 🙌

src/internal/Subscription.ts
```diff
@@ -75,6 +75,14 @@ export class Subscription implements SubscriptionLike {
...
 
     if (isFunction(_unsubscribe)) {
+      // It's not possible to null the _unsubscribe member as there are many
+      // classes that are derived from Subscriber (which derives from
+      // Subscription) that implement an _unsubscribe method as a mechanism for
+      // obtaining unsubscription notifications and some of those subscribers
+      // are recycled. Deleting the member will release the reference to any
+      // teardown functions passed in the constructor and will leave any
+      // methods intact.
+      delete (this as any)._unsubscribe;
```
felixfbecker, line 85: Won't this cause a de-opt in V8 though, i.e. put the entire `Subscription` instance in dictionary mode?

cartant: Yeah. We should really implement an `_unsubscribe` method that's always present and store the teardown that's passed to the ctor in a property that's actually writable.

felixfbecker: A quick fix could also be to make the constructor just call `this.add(teardown)` instead of assigning it to `_unsubscribe`

cartant: Yeah. I considered that, but, IIRC, the implementation of `add` creates another `Subscription` - passing the teardown to the ctor.

cartant: I wasn't too happy about creating a closure in the ctor, either. I think we should just bit the bullet and add the method.

felixfbecker: Sounds good!

cartant: Given that the `Subscription` implementation does so much other clean up, I suppose it make some sense to release the `_unsubscribe` teardown, too, but - I'm curious - is there some particular use case in which you want to continue to hold the subscription after you've called unsubscribe? I mean, if it's inside - i.e. was added to - another longer-lived subscription, the unsubscription of the shorter-lived subscription should remove it from the longer-lived one, so I'm wondering about the use case and whether or not there's a bug someplace else.

felixfbecker: I have a fairly large codebase that uses web workers and `MessagePort`s, which are notorious for needing manual memory management. This code is also running inside a browser extension in a background page (i.e. a long-living context) and will create new ports very often for new tabs, proxying objects within that tab, etc. Given these attributes, I want to do everything in my power to make sure there is no memory leaking. If `_unsubscribe` is not nulled out, I can't guarantee that, because it is very easy to accidentally break out of the nice parent-cleaning tree and there's no way to ensure this at scale. For example, somewhere in the code someone may have written
```ts
subscription.add(() => otherSubscription.unsubscribe())
```
instead of
```ts
subscription.add(otherSubscription)
```
and the nice parent-cleaning chain would be broken, and the parent would hold on to the callback. This is a simple example, but this could also happen with more layers in between, e.g. a class that has an internal `Subscription` bag, but exposes an `unsubscribe` or `close` method itself, that is then added to a `Subscription`. It should be possible to rely on at least the low-level to clean things up properly, so that we may hold a reference to an empty  `Subscription` longer than needed, but at least the underlying `MessagePort` and heavier resources like the web worker etc are cleaned up.

cartant:
@benlesh I've requested your review on this. It's a small change to `Subscription` to ensure that the reference to the teardown passed to the ctor is released/nulled upon unsubscription.

Felix's explanation for this is [here](https://github.com/ReactiveX/rxjs/pull/5465#discussion_r444155818). Given that the subscription implementation goes to so much effort to remove closed subscriptions, this kinda makes sense, to me.

FWIW, various attempts at implementing this change ran into the inheritance weirdness/abuse that you mentioned, recently. There is a comment regarding this where `_unsubscribe` is nulled.

Also, the tests that are added in this PR don't fail without the PR's change to the `Subscription` class. They were added when I was trying to figure out why nulling `_unsubscribe` was effecting failures in other tests. I couldn't see specific tests for `unsubscribe` idempotency, so I figured remain in the PR for inclusion in the test suite.

PR #5473, opened 2020-06-08
chore(readme): add ultimate rxjs course reference

btroncone:
**Description:**

Add Ultimate RxJS course reference to references page. :)

### DIFF
--- a/docs_app/content/resources.md
+++ b/docs_app/content/resources.md
@@ -37,3 +37,9 @@ This is a collection of resources that can be found to help with the learning pr
 * RxJS 6 Changes - Overview |  [academind.com](https://www.academind.com/learn/javascript/rxjs-6-what-changed/) | Last Updated: May 3, 2018
 * Introducing RxJS! | [ng-conf 2018](https://www.youtube.com/watch?v=JCXZhe6KsxQ) | April 20, 2018
 * Learn RxJS in 60 Minutes for Beginners | [YouTube](https://www.youtube.com/watch?v=PhggNGsSQyg) | March 30, 2018
+
+## Courses
+
+### 2020
+
+* RxJS Basics + RxJS Masterclass Courses | [ultimatecourses.com/courses/rxjs](https://ultimatecourses.com/courses/rxjs)


niklas-wortmann:
Honestly I very mixed feelings about this. I kind would prefer not to recommend paid content. By thinking about that I also noticed that we have thinkster as a reference with a paid course and I am also not super happy about this. 

I will think about that and come back to you.

btroncone:
Hey @niklas-wortmann! My thought is it would be nice to have all available resources on the site to make the barrier to entry smaller, paid or unpaid. Ultimately it's each developers decision whether they want to invest in training or use the free resources, but I know when I was getting started it would have been nice to have a comprehensive list to choose from to at least speed up my research. This is similar to the [Angular.io education section](https://angular.io/resources?category=education) .

Let me know what you think, thanks!

benlesh:
@niklas-wortmann I think I agree with you. At least that it makes me a little uncomfortable.  My understanding was that the Thinkster stuff did have some free content. Perhaps we should ask that everyone always have at least _some_ free content available in order to keep their link? (If that's the case, we need to document this somewhere and link it from that list)

On one hand, more resources are a good thing, paid or unpaid. On the other hand, I'd hate for the barrier to entry for RxJS end up being financial. That's not good.

I'll mark this as an Agenda Item and we can discuss it at a team meeting.

benlesh:
There's also a strange boundary around this. I've never used the site to advertise a workshop or paid material I've done. However, we *did* use it to promote rxjs.live, which was paid. But that was also the first ever "community event". 

We probably just need to establish and agree on some policies here.

cartant:
> However, we did use it to promote rxjs.live, which was paid.

But if there hadn't been technical issues with the recordings, all of the talks would have been available for free afterwards.

joshribakoff:
I'm also against promotion of paid courses in OSS docs. Authors of paid courses are benefiting from the hard work of contributors. It goes against the spirit of open source, imo, which is the pursuit of knowledge to advance humanity. Making a living educating is great, but I don't feel docs are a good venue to promote it. As a potential contributor I'd feel like I'm working to make someone else rich, rather than the pursuit of knowledge to advance humanity /rant

niklas-wortmann:
We discussed the whole situation and the core team agreed on deleting the resource page entirely. I am really sorry that it took so long to come up with a decision, but we don't have enough time to maintain the resource page properly. Feel free to reach out to me, if you need further elaborations, but therefore I am going to close this pr.

PR #5477, opened 2020-06-09
chore: deprecate old config options

benlesh:
We need to remove these configuration options for v8, to help streamline the library and reduce size. Adding deprecations messages that give a little bit about why they are deprecated and some workaround information

Also adds some additional blurbs to documentation for promise-related methods on Observable.

Also adds some documentation to forEach with an example.

### DIFF
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -251,12 +251,64 @@ export class Observable<T> implements Subscribable<T> {
   }
 
   /**
-   * @method forEach
+   * Used as a NON-CANCELLABLE means of subscribing to an observable, for use with
+   * APIs that expect promises, like `async/await`. You cannot unsubscribe from this.
+   *
+   * **WARNING**: Only use this with observables you *know* will complete. If the source
+   * observable does not complete, you will end up with a promise that is hung up, and
+   * potentially all of the state of an async function hanging out in memory. To avoid
+   * this situation, look into adding something like {@link timeout}, {@link take},
+   * {@link takeWhile}, or {@link takeUntil} amongst others.
+   *
+   * ### Example:
+   *
+   * ```ts
+   * import { interval } from 'rxjs';
+   * import { take } from 'rxjs/operators';
+   *
+   * const source$ = interval(1000).pipe(take(4));
+   *
+   * async function getTotal() {
+   *    let total = 0;
+   *
+   *    await source$.forEach(value => {
+   *      total += value;
+   *      console.log('observable -> ', value);
+   *    });
+   *
+   *    return total;
+   * }
+   *
+   * getTotal().then(
+   *    total => console.log('Total:', total)
+   * )
+   *
+   * // Expected:
+   * // "observable -> 0"
+   * // "observable -> 1"
+   * // "observable -> 2"
+   * // "observable -> 3"
+   * // "Total: 6"
+   * ```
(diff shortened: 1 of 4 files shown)


src/internal/config.ts
```diff
@@ -18,6 +22,10 @@ export const config = {
    * where a multicast can be broken for all observers by a downstream consumer with
    * an unhandled error. DO NOT USE THIS FLAG UNLESS IT'S NEEDED TO BY TIME
    * FOR MIGRATION REASONS.
+   *
+   * @deprecated remove in v8. As of version 8, RxJS will no longer support synchronous throwing
```
kolodny, line 26: Good catch, I'm surprised this wasn't already "here be dragons"ed already

src/internal/Observable.ts
```diff
@@ -251,12 +251,64 @@ export class Observable<T> implements Subscribable<T> {
   }
 
   /**
-   * @method forEach
+   * Used as a NON-CANCELLABLE means of subscribing to an observable, for use with
```
kolodny, line 254: Is it worth mentioning [rxjs-for-await](https://www.npmjs.com/package/rxjs-for-await)? I'm 50/50

src/internal/lastValueFrom.ts
```diff
@@ -54,4 +60,4 @@ export function lastValueFrom<T>(source: Observable<T>) {
       },
     });
   });
-}
+}
```
kolodny, line 63: We should have an autoformatter that catches things like this

src/internal/Observable.ts
benlesh: I'm not sure. Debatable. I'll leave it for now.

src/internal/firstValueFrom.ts
```diff
@@ -13,6 +13,12 @@ import { Subscription } from './Subscription';
  * If the observable stream emits an error, the returned promise will reject
  * with that error.
  *
+ * **WARNING**: Only use this with observables you *know* will complete. If the source
```
jakovljevic-mladen, line 16: Hi there. Would this sentence be entirely true for `firstValueFrom`? There is a case where the source stream doesn't have to ever complete, it just have to emit at least one value. I would add ...`*know* will complete or emit at least once.`. Does this makes sense?

src/internal/Observable.ts
```diff
@@ -251,12 +251,64 @@ export class Observable<T> implements Subscribable<T> {
...
+   * @param promiseCtor a constructor function used to instantiate the Promise
    * @return a promise that either resolves on observable completion or
    *  rejects with the handled error
+   * @deprecated remove in v8. Passing a Promise constructor will no longer be available
+   * in upcoming versions of RxJS. This is because it adds weight to the library, for very
+   * little benefit. If you need this functionality, it is recommended that you either
+   * polyfill Promise, or you create an adapter to convert the returned native promise
+   * to whatever promise implementation you wanted.
    */
+  forEach(next: (value: T) => void, promiseCtor: PromiseConstructorLike): Promise<void>;
```
cartant, line 310: Why is the `promiseCtor` no longer optional? Adding this signature means that the implementation signature won't be seen, so the fact that the parameter is option in the implementation signature would be meaningless.

cartant: Never mind. I've misread things.

PR #5478, opened 2020-06-09
fix: fix up notification types

benlesh:
- Ensures Notification methods only accept appropriate arguments
- Ensures Arbitrary objects of the proper shape my be passed through dematerialized
- Utilizes hot path functions to create lightweight objects for internal notification use
- Deprecates the Notification class
- Adds a deprecation message for materialize to notify users that soon the emitted object type will change to not have all of the same methods as Notification
- Updates a few tests that were relying on the shape of Notification instances to pass, as we are now using POJOs internally in the TestScheduler
- Adds dtslint tests for notifications
- Adds dtslint tests to enforce types on dematerialize

BREAKING CHANGE: Notification.createNext(undefined) will no longer return the exact same reference everytime.

BREAKING CHANGE: Type signatures tightened up around Notification and dematerialize

### DIFF
--- a/spec-dtslint/Notification-spec.ts
+++ b/spec-dtslint/Notification-spec.ts
@@ -1,0 +1,26 @@
+import { Notification } from "../dist/types/internal/Notification";
+
+describe('Notification', () => {
+  // Basic method tests
+  const nextNotification = Notification.createNext('a'); // $ExpectType Notification<string> & NextNotification<string>
+  nextNotification.do((value: number) => {}); // $ExpectError
+  const r = nextNotification.do((value: string) => {}); // $ExpectType void
+  const r1 = nextNotification.observe({ next: value => { } }); // $ExpectType void
+  const r2 = nextNotification.observe({ next: (value: number) => { } }); // $ExpectError
+  const r3 = nextNotification.toObservable(); // $ExpectType Observable<string>
+  const k1 = nextNotification.kind; // $ExpectType "N"
+
+  const completeNotification = Notification.createComplete(); // $ExpectType Notification<never> & CompleteNotification
+  const r4 = completeNotification.do((value: string) => {}); // $ExpectType void
+  const r5 = completeNotification.observe({ next: value => { } }); // $ExpectType void
+  const r6 = completeNotification.observe({ next: (value: number) => { } }); // $ExpectType void
+  const r7 = completeNotification.toObservable(); // $ExpectType Observable<never>
+  const k2 = completeNotification.kind; // $ExpectType "C"
+
+  const errorNotification = Notification.createError(); // $ExpectType Notification<never> & ErrorNotification
+  const r8 = errorNotification.do((value: string) => {}); // $ExpectType void
+  const r9 = errorNotification.observe({ next: value => { } }); // $ExpectType void
+  const r10 = errorNotification.observe({ next: (value: number) => { } }); // $ExpectType void
+  const r11 = errorNotification.toObservable(); // $ExpectType Observable<never>
+  const k3 = errorNotification.kind; // $ExpectType "E"
+});
--- a/spec-dtslint/Notification-spec.ts
+++ b/spec-dtslint/Notification-spec.ts
@@ -1,4 +1,4 @@
-import { Notification } from "../dist/types/internal/Notification";
+import { Notification } from 'rxjs';
 
 describe('Notification', () => {
   // Basic method tests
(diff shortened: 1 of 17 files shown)


src/internal/operators/dematerialize.ts
```diff
@@ -66,12 +68,18 @@ class DeMaterializeOperator<T extends Notification<any>, R> implements Operator<
...
   }
 
-  protected _next(value: T) {
-    value.observe(this.destination);
+  protected _next(notification: N) {
+    observeNotification(notification, this.destination);
   }
 }
+
+type GetNotificationValueOf<T> = T extends { kind: 'N'|'E'|'C' } ?
```
cartant, line 81: The naming of this type differs from the convention used elsewhere. Other similar types are not prefixed with `Get` - see https://github.com/ReactiveX/rxjs/blob/e1d35dc258edea0237ef49a31f7b34c058755969/src/internal/types.ts#L143

Also, I think this will need to be exported, as it's in the public API.

src/internal/Notification.ts
```diff
@@ -109,22 +193,20 @@ export class Notification<T> {
...
+   * and use them. Rather, try to create POJOs matching the signature outlined in {@link ObservableNotification}.
+   * For example: `{ kind: 'N', value: 1 }`, `{kind: 'E', error: new Error('bad') }`, or `{ kind: 'C' }`.
    */
-  static createNext<T>(value: T): Notification<T> {
-    if (typeof value !== 'undefined') {
-      return new Notification('N', value);
-    }
-    return Notification.undefinedValueNotification;
+  static createNext<T>(value: T): Notification<T> & NextNotification<T> {
+    return new Notification('N', value) as any;
```
cartant, line 209: Consider using an assertion instead of `any` and, maybe, an implicit return type:

```ts
  static createNext<T>(value: T) {
    return new Notification('N', value) as Notification<T> & NextNotification<T>;
}
```

src/internal/Notification.ts
```diff
@@ -134,17 +216,86 @@ export class Notification<T> {
...
    * argument.
    * @nocollapse
+   * @deprecated remove in v8. It is NOT recommended to create instances of `Notification` directly
+   * and use them. Rather, try to create POJOs matching the signature outlined in {@link ObservableNotification}.
+   * For example: `{ kind: 'N', value: 1 }`, `{kind: 'E', error: new Error('bad') }`, or `{ kind: 'C' }`.
    */
-  static createError<T>(err?: any): Notification<T> {
-    return new Notification('E', undefined, err);
+  static createError(err?: any): Notification<never> & ErrorNotification {
+    return new Notification('E', undefined, err) as any;
```
cartant, line 224: Consider using an assertion, etc., instead of `any`.

src/internal/Notification.ts
```diff
@@ -134,17 +216,86 @@ export class Notification<T> {
...
    * A shortcut to create a Notification instance of the type `complete`.
    * @return {Notification<any>} The valueless "complete" Notification.
    * @nocollapse
+   * @deprecated remove in v8. It is NOT recommended to create instances of `Notification` directly
+   * and use them. Rather, try to create POJOs matching the signature outlined in {@link ObservableNotification}.
+   * For example: `{ kind: 'N', value: 1 }`, `{kind: 'E', error: new Error('bad') }`, or `{ kind: 'C' }`.
    */
-  static createComplete(): Notification<any> {
+  static createComplete(): Notification<never> & CompleteNotification {
     return Notification.completeNotification;
```
cartant, line 236: Consider using an assertion, etc., instead of `any`.

spec-dtslint/Notification-spec.ts
```diff
@@ -0,0 +1,26 @@
+import { Notification } from "../dist/types/internal/Notification";
```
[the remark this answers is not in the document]
benlesh: UGH... VS Code gets me with that all the time.

src/internal/operators/dematerialize.ts
benlesh: Renamed to `ValueFromNotification` to match things like `ValueFromArray`.

src/internal/Notification.ts
```diff
@@ -134,17 +216,86 @@ export class Notification<T> {
...
+ */
+export const COMPLETE_NOTIFICATION = (() => createNotification('C', undefined, undefined) as CompleteNotification)();
+
+/**
+ * Internal use only. Creates an optimized error notification that is the same "shape"
+ * as other notifications.
+ * @internal
+ */
+export function errorNotification(error: any): ErrorNotification {
+  return createNotification('E', undefined, error) as any;
```
cartant, line 277: There's one `any` left that could instead be an assertion (and the function could have an implicit return type).

PR #5479, opened 2020-06-10
feat(expand): Scheduler argument removed.

benlesh:
Turns out it never, ever worked properly and no one ever complained about it. So I want to remove it.

Adds better examples and documentation.

BREAKING CHANGE: expand no longer accepts a scheduler argument. Because it never worked, and no one complained. Good riddance.


benlesh:
Closed in favor of #5480  (Thanks, @trxcllnt)

PR #5480, opened 2020-06-10
operator(expand): fix expand with async/asap schedulers

trxcllnt:
**Description:**
Using `expand()` on synchronous outer source/inner sources with the async and asap schedulers causes expand to terminate prematurely. The inner active counter is incremented when the scheduled task fires, not when it's enqueued. This PR fixes that.


cartant:
> People may be working around this bug by using the same scheduler for the outer and inner source.

If there's any way the use of a scheduler could have worked, I now think it should be fixed. Removal would only be an option if it was completely broken and would have not worked under any circumstances.

Issue #5209, opened 2020-01-02
forkJoin should emit empty array or object given empty array or object as input


aquark:
I'm reopening https://github.com/ReactiveX/rxjs/issues/2816 because the response wasn't satisfactory.

**RxJS version:** 6.5.2

**Code to reproduce:**

```
forkJoin([]).pipe(
  tap(values => console.log(values)),
).subscribe()
```

**Expected behavior:**

Console logs `[]`

**Actual behavior:**

Nothing is logged.

**Additional information:**

Compare the behavior of `Promise.all`:

```
Promise.all([]).then(values => console.log(values))
// logs []
```

Or `Array.every`:

```
[].every(() => false) // returns true
```

This behavior is important for reasoning inductively and for composability. For example, say I have a list of search terms to expand to search results and then I want to concatenate all the results. The natural way to write this is:

```
forkJoin(searchTerms.map(term => getSearchResults(term))).pipe(
  map(allResults => allResults.flat()),
)
```

You would expect an empty list of search terms to produce an empty list of results, however with current `forkJoin` behavior you need to handle that with a special case:

```
forkJoin(searchTerms.map(term => getSearchResults(term))).pipe(
  map(allResults => allResults.flat()),
  defaultIfEmpty(new Array<SearchResult>()),
)
```

I believe that in every use case where the length of the input to `forkJoin` is unknown at compile time, it's simpler for the consuming code if `forkJoin` returns empty output on empty input, so the consuming code doesn't need to treat empty input as a special case; it can iterate over the output value the same way regardless. This is an example of the [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern), where for example it's better for an operation which produces a list to produce an empty list in case there is no output rather than a `null` value which must be handled specially.

In the original issue https://github.com/ReactiveX/rxjs/issues/2816 it was argued that "Forkjoin doesn't consider any other case except all sources emits value, so empty array is also short-curcuit to completion immediately." However in case there are no sources, it is logically true that "all sources emit value", so one would therefore expect forkJoin to emit a value.

This reasoning also applies to `forkJoin({})`, which should emit an empty object `{}`.

startswithaj:
Yeah, this is a recurring problem for us when using `combineLatest`. We've implemented our own wrapper function. 

For details on our use case see: 

https://stackblitz.com/edit/rxjs-playground-royl-vxadk4?file=index.ts

cartant:
`zip` needs the same consideration, too.

[...]

Issue #4410, opened 2018-12-07
Use generic rest parameters for `combineLatest`, `zip`, `merge`, `concat`, `withLatestFrom`


ggradnig:
## Use generic rest parameters for `combineLatest`, `zip`, `merge`, `concat`, `withLatestFrom` (more than 6 input Observables)

**Is your feature request related to a problem? Please describe.**
There are currently no type signatures for the mentioned operators/creation functions that allow more than 6 Observables while still keeping the correct types for the subscriber. 

**Describe the solution you'd like**
Add a signature that makes use of generic rest parameters ([documented here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#generic-rest-parameters)) and mapped types. Currently, the signatures look like this:

```typescript
export function combineLatest<T, T2, T3, T4, T5, T6>(v1: ObservableInput<T>, 
    v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, 
    v5: ObservableInput<T5>, v6: ObservableInput<T6>, scheduler?: SchedulerLike)
    : Observable<[T, T2, T3, T4, T5, T6]>;
```

This could for example be replaced with:

```typescript
type CombinedObservable<T> = Observable<{
  [P in keyof T] : T[P] extends ObservableInput<infer U> ? U : never
}>

combineLatest<T extends (ObservableInput<any> | SchedulerLike)[]> (...obs: T) 
    : CombinedObservable<T>
```

cartant:
Using tuples and rest parameters - necessitating a minimum of TS 3.0 - would be a breaking change for many projects, so we'll need to wait for v7 to do this. The good news is that v7 is getting close to an alpha. Also, we now have a large portion of the API surface area covered with `dtslint`-based typings tests - so making changes to the typings ought to be less worrying. When we do get around to taking advantage of what TS 3.0, any contributions you're able to make would definitely be appreciated.

ggradnig:
Okay, thanks for the detailed reasoning and update on the development status. I'll stick around for v7 and see what I can contribute :-)

PR #5488, opened 2020-06-13
feat(combineLatest): add N-args signature for observable inputs

ggradnig:
**Description:**
Adds a type signature to provide an array with more than six elements to `combineLatest` without losing the order of the input observable types. We recently had this problem at a client project and needed to resort to nesting `combineLatest`. 

The solution makes use of TypeScript readonly tuples. They allow for an array `[a, b]` to have the type `[A, B]` instead of `(A | B)[]`. With this type, we can make use of `ObservedValueTupleFromArray` to get all the observed types of the tuple in the correct order.

This comes with the "drawback" that the input array must be declared `as const`, else TypeScript assumes a union type per default. It would work with spreading syntax though like in `combineLatestWith`, but this signature was deprecated for `combineLatest`.

For this to work, the conditional type `ObservedValueTupleFromArray` must check for the narrow `readonly T[]` type, but this shouldn't be an issue as `T[]` extends it.

**Related issue:**
https://github.com/ReactiveX/rxjs/issues/5066
> observable/combineLatest.ts - needs N-args signature for observable inputs? See also #5209

https://github.com/ReactiveX/rxjs/issues/4410


rraziel:
@ggradnig this one could be useful to you as well: #5363

ggradnig:
Yeah thanks for the hint, but this pretty much already exists in https://github.com/cartant/rxjs-etc
This PR just adds a type signature without any runtime changes.

benlesh:
Thank you @ggradnig!

PR #5496, opened 2020-06-15
chore: type repeatWhen notifier as void

cartant:
**Description:**

This PR makes a trivial change to the type of the `notifications` observable passed to the `notifier` in `repeatWhen`. The observable emits when the source completes and it does not emit a value - there is an internal `Subject<void>` that's used for notifications.

This PR types the `notifications` as `Observable<void>` instead of `Observable<any>`.

No dtslint test is added because there is no inference involved here.

### DIFF
--- a/src/internal/operators/repeatWhen.ts
+++ b/src/internal/operators/repeatWhen.ts
@@ -39,12 +39,12 @@ import { MonoTypeOperatorFunction, TeardownLogic } from '../types';
  * @return {Observable} The source Observable modified with repeat logic.
  * @name repeatWhen
  */
-export function repeatWhen<T>(notifier: (notifications: Observable<any>) => Observable<any>): MonoTypeOperatorFunction<T> {
+export function repeatWhen<T>(notifier: (notifications: Observable<void>) => Observable<any>): MonoTypeOperatorFunction<T> {
   return (source: Observable<T>) => source.lift(new RepeatWhenOperator(notifier));
 }
 
 class RepeatWhenOperator<T> implements Operator<T, T> {
-  constructor(protected notifier: (notifications: Observable<any>) => Observable<any>) {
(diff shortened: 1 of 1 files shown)


benlesh:
Thanks, @cartant!

leggechr:
I'm seeing an issue in Google that I think is a result of this PR. Could I get some help with a fix please?
The code is as follows: 
```ts
 this.updateSubscription =
        this.alertService.getAlerts()
            .pipe(repeatWhen(
                (observable: Observable<GetAlertDescriptionResponse>) =>
                    observable.pipe(concatMap(() => timer(5000)))))
            .subscribe((response: GetAlertDescriptionResponse) => {
              this.response = response;
            });
```

And I get the following error:
```
Argument of type '(observable: Observable<GetAlertDescriptionResponse>) => Observable<number>' is not assignable to parameter of type '(notifications: Observable<void>) => Observable<any>'.
  Types of parameters 'observable' and 'notifications' are incompatible.
    Type 'Observable<void>' is not assignable to type 'Observable<GetAlertDescriptionResponse>'.
      Type 'void' is not assignable to type 'GetAlertDescriptionResponse'.

38                 (observable: Observable<GetAlertDescriptionResponse>) =>
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
39                     observable.pipe(concatMap(() => timer(5000)))))
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

cartant:
Remove the `Observable<GetAlertDescriptionResponse>` annotation. It's incorrect. The notifier will not emit a value.

PR #5498, opened 2020-06-15
ci(circle): deprecate circle ci

kwonoj:
**Description:**

I think gh actions working well and #5276 is proven enough, we can take out circleci now.


kwonoj:
womp, just removing config doesn't seem to disable check.

PR #5501, opened 2020-06-16
Timer fix unhappy numbers

benlesh:
Fixes for `TestScheduler` and `VirtualTimeScheduler`:
- Resolves an issue where scheduling at non-finite numbers would result in marbles showing up in tests that had a `frame` of `null`.

Fixes for `timer`:
- Adds tests for cases where `Infinity` could be passed
- Adds a test for dueTime `Date` objects in the past
- Completely rewrites and improves documentation and updates examples
- Renames internal utility function `isDate` to `isValidDate` to better convey what it is doing and why

This is just part of my march through the code to fix up things that were using `isNumeric`, which I find highly suspect.


benlesh:
@niklas-wortmann please check the docs changes here... since I pretty much rewrote the docs for timer.

PR #5506, opened 2020-06-19
test(import): introduce import site integration test

kwonoj:
**Description:**

This PR introduces new integration test for checking `import` works across environments.
Supporting cjs is not a huge concern for us, but as we are approaching to support more env (native esm, webpack's tree-shaking, etc) regressions like https://medium.com/javascript-in-plain-english/is-promise-post-mortem-cab807f18dcc could possibly occur. This test intends to test run script with specific import site, validate all exports are included compare to src code when pkg's created.

flowwise this test does

- create local pkg
- each test fixture installs local pkg
- import each export site (`rxjs`, `rxjs/operators`) in its way then validate imports are identical to exported in srcs.

Currently there's only cjs testing, we'll likely add esm once it we able to support native esm / and webpack's esmodule too. 


integration/import/runner.ts
```diff
@@ -0,0 +1,60 @@
...
+  // create snapShot from existing src's export site
+  writeFileAsync(path.join(fixturePath, 'rx.json'), JSON.stringify(Object.keys(rx_export_src)));
+  writeFileAsync(path.join(fixturePath, 'operators.json'), JSON.stringify(Object.keys(operators_export_src)));
+
+  await execPromise(`npm install ${pkgPath} --no-save`, fixturePath);
+};
+
+const main = async () => {
+  // create local pkgs
+  await execPromise('npm run build:package');
```
benlesh, line 41: If this is running in CI, couldn't it just use what was already built rather than building it again?

integration/import/runner.ts
```diff
@@ -0,0 +1,60 @@
...
+  await execPromise('npm pack');
+
+  const pkgPath = path.join(projectRoot, `rxjs-${version}.tgz`);
+
+  // install local pkg to each fixture path, run test scripts
+  for (const key in IMPORT_TYPE) {
+    const value = (IMPORT_TYPE as any)[key];
```
benlesh, line 48: Seems like a simple array might make more sense here than the `enum`.

integration/import/runner.ts
kwonoj: it's precaution to maintain clean slate. our CI turnaround is not slow and thought this would be better.

integration/import/runner.ts
```diff
@@ -0,0 +1,60 @@
...
+import * as fs from 'fs';
+
+import * as rx_export_src from '../../src/index';
+import * as operators_export_src from '../../src/operators/index';
+
+const writeFileAsync = promisify(fs.writeFile);
+const projectRoot = process.cwd();
+const { version } = require(path.join(projectRoot, 'package.json'));
+
+const execPromise = (cmd: string, cwd = projectRoot) =>
```
benlesh, line 13: nit: use named functions at top level

benlesh:
I've added a few comments on _the code_... but I'm still wrapping my head around the issue it is solving. (I didn't have time to read the whole article). It sounds like we're trying to optimize to make sure we don't change how we're exporting things in such a way it breaks the way some bundlers handle imports?

kwonoj:
This test doesn't optimize anything: it validates we properly exports so consumer won't hit breaking changes. In current state this test doesn't reveal much, since cjs alone our pkg doesn't have major difference. This'll be more clear when we trying to support native esm - depends on pkg manifest, whole cjs / esm / old-esm (which used `module` property for tree-shaking in webpack) behaves entirely different.

Refer attached ref https://medium.com/javascript-in-plain-english/is-promise-post-mortem-cab807f18dcc which explains this cases - i.e 

> When adding the"exports" field to package.jsonas a way to tell newer versions of node.js about the ES Modules version, I had assumed the paths worked the same as "main" but instead they require the ./ prefix.
> I had not understood that the"exports" field to package.jsonrestricts not just what you can import, but how you can import it. So even though my change allowed you to import the index.js file, it would break your code if you were doing require('is-promise/index') or require('is-promise/index.js').

It is totally possible we can break entire cjs import by setting pkg to native esm (realistically it won't, but theoretically). Test trying to prevent those regressions.

benlesh:
Core Team meeting: General approval.

benlesh:
@kwonoj ... just bumping this. It's only waiting for additional comments to be added about what and why, and it also needs to be rebased. <3

kwonoj:
yeah sorry, bit swamped with personal issues. Trying to get back soon.

kwonoj:
@benlesh PR's rebased and added top-level comment for quick explanations.

PR #5507, opened 2020-06-19
build(ci): add node.js 14 into matrix

kwonoj:
This PR adds node.js 14 into ci matrix.

### DIFF
--- a/.github/workflows/ci_main.yml
+++ b/.github/workflows/ci_main.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


kwonoj:
This currently fails, need to dig bit.

kwonoj:
That's easy thing, I recall reasoning behind was at least running all major even if it's not LTS cause it could have different issues around - and running all of them is not expensive as our CI turnaround time is quite quick.

kwonoj:
ugh, I checked this in - I'll check LTS matrix and prep separate PR for drop off some version if we need it.

PR #5513, opened 2020-06-21
fix(defer) allow passing () => any to observableFactory

### LINKED ISSUE #5511: defer typings incorrect
In `[R] extends [undefined] | [void] ? never : () => R`, if a user passes a function to `defer` shaped like `() => any`, `any` matches `[R] extends [undefined] | [void]`, and therefor it returns `never`, and it doesn't compile.

kolodny:
Fixes #5511

### DIFF
--- a/src/internal/observable/defer.ts
+++ b/src/internal/observable/defer.ts
@@ -52,7 +52,14 @@ import { from } from './from'; // lol
  * @owner Observable
  */
 export function defer<R extends ObservableInput<any>>(
-  observableFactory: [R] extends [undefined] | [void] ? never : () => R
+  observableFactory:
+    [R] extends [never] ?
+      never :
+      [R] extends [any] ?
+        () => R :
+        [R] extends [undefined] | [void] ?
+          never :
+          () => R
 ): Observable<ObservedValueOf<R>> {
   return new Observable<ObservedValueOf<R>>(subscriber => {
     let input: R;
--- a/src/internal/observable/defer.ts
+++ b/src/internal/observable/defer.ts
@@ -51,16 +51,7 @@ import { from } from './from'; // lol
  * @name defer
  * @owner Observable
  */
-export function defer<R extends ObservableInput<any>>(
-  observableFactory:
-    [R] extends [never] ?
-      never :
-      [R] extends [any] ?
-        () => R :
-        [R] extends [undefined] | [void] ?
-          never :
-          () => R
-): Observable<ObservedValueOf<R>> {
+export function defer<R extends ObservableInput<any>>(observableFactory: () => R): Observable<ObservedValueOf<R>> {
   return new Observable<ObservedValueOf<R>>(subscriber => {
     let input: R;
     try {
--- a/spec-dtslint/observables/defer-spec.ts
+++ b/spec-dtslint/observables/defer-spec.ts
@@ -32,8 +32,8 @@ it('should error if function returns undefined', () => {
   const a = defer(() => undefined); // $ExpectError
 });
 
-it('should error if function returns never', () => {
-  const a = defer(() => { throw new Error(); }); // $ExpectError
+it('should infer if function returns never', () => {
+  const a = defer(() => { throw new Error(); }); // $ExpectType Observable<never>
 });
 
 
--- a/spec-dtslint/observables/defer-spec.ts
+++ b/spec-dtslint/observables/defer-spec.ts
@@ -16,6 +16,10 @@ it('should support union type returns', () => {
   const a = defer(() => Math.random() > 0.5 ? of(123) : of('abc')); // $ExpectType Observable<string | number>
 });
 
+it('should infer correctly with function return any', () => {
+  const a = defer(() => 3 as any); // $ExpectType Observable<unknown>
+});
+
 it('should error with void functions', () => {
   const a = defer(() => {}); // $ExpectError
 });
@@ -43,6 +47,6 @@ it('should infer correctly with function that sometimes error', () => {
   });
 });
 
-it('should infer correctly with functions that sometimes do not return an ObservableInput', () => {
+it('should error with functions that sometimes do not return an ObservableInput', () => {
   const a = defer(() => { if (Math.random() < 0.5) { return of(42); } }); // $ExpectError
 });


cartant:
I'm not sure why the change that introduced these types is in #5449 the codebase. I though it was agreed - see this https://github.com/ReactiveX/rxjs/pull/5449#discussion_r428276385 comment - that the removal of [the `void` overload and the `void` in the return type's union](https://github.com/ReactiveX/rxjs/blob/362d1d4fc1d7f15e7168453ea640079d0ca19dba/src/internal/observable/defer.ts#L56-L57) ought to be sufficient. Is that not the case?

I'm a little uneasy about having to resort to a ternary conditional type for something that should be straightforward. I mean, rejecting functions that return `void` is pretty common. And functions that return `undefined` should be no problem, as `undefined` is not a valid `ObservableInput`.

kolodny:
I guess I misunderstood that comment then. I tried the changes in https://github.com/ReactiveX/rxjs/blob/362d1d4fc1d7f15e7168453ea640079d0ca19dba/src/internal/observable/defer.ts#L56-L57 but that causes a slew of other errors such as

```ts
it('should error with void functions', () => {
  const a = defer(() => {}); // $ExpectError
});
```

> And functions that return undefined should be no problem, as undefined is not a valid ObservableInput.

I thought we wanted to reject functions that don't return an observablelike value

Are the current dtslint tests incorrect? Otherwise I don't see how else we can get the expected behavior without nested conditionals.

cartant:
> I thought we wanted to reject functions that don't return an observablelike value

Yeah, we do. I meant that rejecting functions return `undefined` should be no problem, as `undefined` is not a valid `ObservableInput`. (Sorry for lack of clarity in my comment.)

The current dtslint tests and those in this PR look okay, to me. I don't see why they should not pass if the requirement is that the factory returns a valid `ObservableInput`.

src/internal/observable/defer.ts
```diff
@@ -52,7 +52,14 @@ import { from } from './from'; // lol
...
 export function defer<R extends ObservableInput<any>>(
-  observableFactory: [R] extends [undefined] | [void] ? never : () => R
+  observableFactory:
+    [R] extends [never] ?
+      never :
+      [R] extends [any] ?
+        () => R :
+        [R] extends [undefined] | [void] ?
+          never :
+          () => R
```
benlesh, line 62: I'm always suspicious that the simplest thing is probably the correct thing (however I'm frequently wrong, but this looks suspicious to me)...  Just to be sure we're clear here: What this is doing is saying "This has to be an observable input, such that: `void|undefined|never` are errors, and `any` will result in `Observable<unknown>`? Is that correct?

Is there any way to make a utility type out of this? Do we need this anywhere else? 🤔 

The returning of `any` resulting in an `Observable<unknown>` seems useful, however it sort of breaks `any`, right? IMO, `any` is the escape hatch people use when they just give up.  Let's say someone has a hard time figuring out the types to a defer that would otherwise return valid things:

```ts
defer(() => {
   return getSomethingAuthorIsTooLazyToType() as any
}).pipe(
   map((thing) => thing.methodThatIsAlwaysThere())
)
```

If this is `Observable<unknown>` we just broke `any` for them, and they have to kick the `any` down the way...

And _maybe that is what we want to do_... I'm just discussing at this point, not asking for a change.

@cartant ? @kwonoj ?

I'm truly on the fence about this one.

cartant: @benlesh I thought the original problem was that `defer` accepted functions that returned `void` and I thought we'd agreed that the solution was to remove the overload signature that allowed that. I was surprised to see that the code this PR fixes was merged. See [my comment above](https://github.com/ReactiveX/rxjs/pull/5513#issuecomment-647073283).

benlesh: Yeah... so that was the original issue. Then, with the new typings that fixed that _original issue_, we caused issues with factories that returned `any`, which should be an escape hatch for developers and "just work"... because `any`. Which is why I filed the issue that this is trying to solve.

I'm just not sure returning `Observable<unknown>` is right here. Really, truly, on the fence about it.

benlesh: I think maybe there was a misunderstanding with your comment in the other PR, at least on my end. I thought you were talking about the runtime behavior, which was indeed resolved, but you were generally approving of the typings. Which looked fine to me. But those typings caused issues with `() => any` , which is what we're trying to resolve here.

cartant: I'm saying that I don't think [this fix](https://github.com/ReactiveX/rxjs/pull/5449) was the best way of fixing the problem. And that this fix exacerbates it. It seems too complicated, IMO.

I think the fix should have been what's in [this comment](https://github.com/ReactiveX/rxjs/pull/5449#discussion_r428276385).

benlesh: Yeah. I agree.

Funny enough, the recommended fix has the same outcome. Was talking with @jayphelps about this offline, and he pointed this out.

```ts
declare function foo<T>(cb: () => T[]): T[] { }

const result = foo(() => 1 as any); // unknown[]
```

cartant: I think is still needs the `ObservedValueOf` business to handle factories with multiple return paths of different types, but, yeah, I think the types and - the implementation - should be changed to require a factory that actually returns an observable input. That's easily the 95% use case, IMO. If someone wants to return `EMPTY`, they can do so explicitly.

kolodny:
> We should simplify the types here to just be:
> 
> ```ts
> function defer<T>(callback: () => ObservableInput<T>): Observable<T>
> ```

Just getting back to this now, is looks like this causes some TS weirdness with union types

```ts
export function defer<T>(observableFactory: () => ObservableInput<T>): Observable<T> {
  return new Observable<T>(subscriber => {
    let input: ObservableInput<T>;
    try {
      input = observableFactory();
    } catch (err) {
      subscriber.error(err);
      return undefined;
    }
    const source = from(input);
    return source.subscribe(subscriber);
  });
}

defer(() => true ? of('3') : of(1)) // Type 'number' is not assignable to type 'string'
```

This also breaks https://github.com/ReactiveX/rxjs/blob/master/src/internal/observable/iif.ts#L99
image


Changing iif.ts to use `ObservableInput` or changing defer.ts to use `SubscribableOrPromise` doesn't fix the issue either.

Also this definition has the same issue as well:

```ts
export function defer<R>(observableFactory: () => ObservableInput<R>): Observable<ObservedValueOf<R>> {
```

I need to spend a bit more time on this to figure this out, it's possible I'm missing something obvious (and if so please point it out)

cartant:
I think you should go back to #5449 and look at what was there before that PR's changes. In particular, this line seems to have been changed:

```ts
const source = input ? from(input as ObservableInput<ObservedValueOf<R>>) : EMPTY;
```

And the assertion was removed. If that doesn't help, I can look at this tomorrow, but, IMO, the gist of it is to go back to what it was before #5449,  but remove the `void` business.

kolodny:
> I think you should go back to #5449 and look at what was there before that PR's changes. In particular, this line seems to have been changed:
> 
> ```ts
> const source = input ? from(input as ObservableInput<ObservedValueOf<R>>) : EMPTY;
> ```
> 
> And the assertion was removed. If that doesn't help, I can look at this tomorrow, but, IMO, the gist of it is to go back to what it was before #5449, but remove the `void` business.

Got it, does that mean that this is no longer a type error?:

```ts
const a = defer(() => { throw new Error(); });
```

cartant:
> ... does that mean that this is no longer a type error?

Yes, that should not be a type error. For example, [this is not an error](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZgLhgCgJQwLwD4ZgK4C2ARgKYBOmK62MA3jFABZkgDuMAomS2cgEQAhECAIBCPqgDcMAL6SgA):

```ts
const f: () => number = () => { throw Error("Boom!"); };
```

benlesh:
Okay, just to clear things up here:

Passing any function to `defer` that doesn't return a `ObservableInput<T>` type, should be a TS compilation error.

At runtime, (in plain JS), if the function doesn't return an ObservableInput, that should also be an error (propagated down the error path of the observable).

The weird edge cases that are discussed above are mostly covered, I think, but for clarity:   `defer(() => { throw new Error() })` should be a TS compilation error, and `defer((): Observable<any> => { throw new Error(); })` should not be a ts compilation error. This is because the former clearly is `() => never`, and is likely a bug. The later is perceived to be intentional (with the explicit return type).

cartant:
@benlesh IDK that this last bit clears things up:

> The weird edge cases that are discussed above are mostly covered, I think, but for clarity: `defer(() => { throw new Error() })` should be a TS compilation error, and `defer((): Observable<any> => { throw new Error(); })` should not be a ts compilation error. This is because the former clearly is `() => never`, and is likely a bug. The later is perceived to be intentional (with the explicit return type).

I disagree completely with this. This isn't how the type system works and to get the outlined behaviour, the type will be unnecessarily complicated. Functions that always throw are assignable to functions that expect a return type - as shown in [this comment](https://github.com/ReactiveX/rxjs/pull/5513#issuecomment-650823070).

IMO, there is no value in changing this for `defer` - i.e. deliberately effecting a compilation error. `Observable<never>` is assignable to a non-`never` `Observable` and adding special handling to `defer` will just make things complicated and the 'problem' will exist in other places, anyway:

```ts
declare let valNever: never;
declare let valNumber: number;
valNever = valNumber; // Error: Type 'number' is not assignable to type 'never'
valNumber = valNever; // No error

declare let obsNever: Observable<never>;
declare let obsNumber: Observable<number>;
obsNever = obsNumber; // Error: Type 'number' is not assignable to type 'never'
obsNumber = obsNever; // No error

// No error
const source: Observable<number> = new Observable<number>(() => {
  throw new Error("Boom!");
});

// No error
const answer: Observable<number> = of(42).pipe(
  map(() => { throw new Error("54"); })
);
```

benlesh:
To clarify my position, I want to make the types as simple as possible. I thought that what I was saying was the default behavior of the type system. It's entirely likely that I'm mistaken.

I think you and I are violently in agreement. LOL

kolodny:
OK, I think I've got it. Let me know if this is what you had in mind

Issue #5520, opened 2020-06-23
shareReplay: it skips values on initial subscription if the source is sync


josepot:
## Bug Report

**Current Behavior**
`shareReplay` only emits the latest `bufferSize` values on the initial subscription when the source is synchronous.

**Reproduction**
- REPL or Repo link:
https://stackblitz.com/edit/rxjs-qc2bsv?devtoolsheight=60

```js
import { from } from 'rxjs'; 
import { shareReplay } from 'rxjs/operators';

const source = from([1, 2, 3, 4]).pipe(
  shareReplay(1)
);

source.subscribe(x => console.log(x));
// the console only logs 4
```

**Expected behavior**
IMO the first subscription should log all the values (`1`, `2`, `3`, `4`).

**Environment**
- Runtime: all
- RxJS version: 6.5.5

**Possible Solution**
Change [these lines](https://github.com/ReactiveX/rxjs/blob/9efc3d50f67ac1dac8c0440b28d0f6bae07bd5f7/src/internal/operators/shareReplay.ts#L158-L174) to:

```ts
    let innerSub: Subscription;
    if (!subject || hasError) {
      hasError = false;
      subject = new ReplaySubject<T>(bufferSize, windowTime, scheduler);
      innerSub = subject.subscribe(this);
      subscription = source.subscribe({
        next(value) { subject!.next(value); },
        error(err) {
          hasError = true;
          subject!.error(err);
        },
        complete() {
          subscription = undefined;
          subject!.complete();
        },
      });
    } else {
      innerSub = subject.subscribe(this);
    }
```

I'm having a hard time coming up with a marble-test that tests this :thinking:

cartant:
Yeah, definitely a bug. The subscriber to the operator should subscribe to the subject **before** the subject subscribes to the source.

BTW, I was just perusing the source for `re-rxjs`. Very interesting.

josepot:
> Yeah, definitely a bug. The subscriber to the operator should subscribe to the subject before the subject subscribes to the source.

Np. I will send a PR, thanks!

> BTW, I was just perusing the source for re-rxjs. Very interesting.

Thanks, that means a lot to me! How did you find re-rxjs?

Actually we found this bug [thanks to re-rxjs](https://github.com/re-rxjs/re-rxjs/pull/26) :sweat_smile: 

[...]

PR #5521, opened 2020-06-23
fix(shareReplay): subscribe to subject before source subscription starts

josepot:
**Description:**

Addresses #5520

### DIFF
--- a/spec/operators/shareReplay-spec.ts
+++ b/spec/operators/shareReplay-spec.ts
@@ -3,7 +3,7 @@ import * as sinon from 'sinon';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 import { shareReplay, mergeMapTo, retry } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
-import { Observable, Operator, Observer, of } from 'rxjs';
+import { Observable, Operator, Observer, of, from } from 'rxjs';
 
 declare const rxTestScheduler: TestScheduler;
 
@@ -267,7 +267,15 @@ describe('shareReplay operator', () => {
   });
 
   it('should not skip values on a sync source', () => {
-    const a = cold(  '(abcd)');
+    const a = from(['a', 'b', 'c', 'd']);
+    // We would like for the previous line to read like this:
+    //
+    // const a = cold('(abcd|)');
+    //
+    // However, that would synchronously emit multiple values at frame 0,
+    // but it's not synchronous upon-subscription.
+    // TODO: revisit once https://github.com/ReactiveX/rxjs/issues/5523 is fixed
+
     const x = cold(  'x-------x');
     const expected = '(abcd)--d';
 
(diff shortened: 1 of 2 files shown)


josepot:
> TBH, I preferred the previous test - the one with the cold, synchronous marbles source.

The problem with using `cold` is that if we run it against master the test should fail, but it passes :disappointed: I think that the scheduler does something behind the scenes that doesn't make the observable completely cold? :man_shrugging: @voliva was the one that came up with the idea of using `from` instead of `cold` and it did the trick (fails on master and passes after the fix)

cartant:
My original, suggested test had an error in it: `!` instead of `|`. A `|` character would represent completion. Does it fail as you'd expect if the marble source completes?

```
(abcd|)
```

If it doesn't, using `from` is fine, but I guess there's a bug if that's the case.

voliva:
> My original, suggested test had an error in it: `!` instead of `|`. A `|` character would represent completion. Does it fail as you'd expect if the marble source completes?

Unfortunately it doesn't - It seems like the "frame 0" doesn't get emitted synchronously when the subscription is made.

josepot:
> My original, suggested test had an error in it: ! instead of |. A | character would represent completion. Does it fail as you'd expect if the marble source completes?

When I try to run this test:
```ts
it('should not skip values on a sync source', () => {
  const a = cold(  '(abcd!)');                       
  const x = cold(  'x-------x');                     
  const expected = '(abcd)--d';                      
                                                     
  const shared = a.pipe(shareReplay(1));             
  const result = x.pipe(mergeMapTo(shared));         
});                            
```
I get the following error:

```bash
Error: cold observable cannot have unsubscription marker "!"
```

cartant:
Yeah, that's the error I made. The source should have had `|` in it, not `!`. It should have been:

```
const a = cold(  '(abcd|)');
```

cartant:
> Unfortunately it doesn't - It seems like the "frame 0" on a cold() doesn't get emitted synchronously when the subscription is made.

Yeah, I guess that would be what's happening. Using `from` is fine. Thanks.

josepot:
yep, sorry for the mess @cartant . I've opened #5522 to demonstrate that the test using `cold` for some reason does pass on `master`.

As much as I struggle with marble tests I do appreciate the value that they add by visually showing what's happening. However, in this particular case it looks like we can't use `cold` to test this

cartant:
Could you add a comment that mentions the use of `from` is deliberate as `cold("(abcd|)")` synchronously emits at frame 0, but is not synchronous-upon-subcription? - so that the `from` source is not unwittingly replaced with a marble at a later date.

I'll have to go through the other tests later, too, as I suspect there are other places in which a source is supposed to be synchronous-upon-subscription and not just the synchronous emission of multiple notifications - e.g. [here](https://github.com/ReactiveX/rxjs/blob/d608cf33791cceb5f330270da96b0b2805223208/spec/operators/share-spec.ts#L209).

josepot:
@cartant please let me know if [this is good enough](https://github.com/ReactiveX/rxjs/pull/5521/files#diff-52f62da75a279650a9e6557aa988f9ceR271-R278), or if you would like for me to change/rephrase that. (English is not my first language and I'm always trying to improve, so please don't hesitate to suggest changes to make it more understandable.

PR #5524, opened 2020-06-23
fix(pluck): operator breaks with null/undefined inputs.

### LINKED ISSUE #5505: Pluck operator doesn't fully check for null or undefined
If I have a source that emits a nullish value and try to use `pluck`, I get an error.

```typescript
const source = of(null).pipe(pluck('someProp'));
source.subscribe();
// ERROR TypeError: Cannot read property 'someProp' of null
```

I think it would be nice if `pluck` checked for this and returned `undefined`.

```typescript
const source = of(null).pipe(pluck('someProp'));
// logs: undefined
```

[The description of this operator even says](https://github.com/ReactiveX/rxjs/blob/78032157f5c1655436829017bbda787565b48c30/src/internal/operators/pluck.ts#L25), "If a property can't be resolved, it will return `undefined` for that value." Seems to make sense that it would do the same for the value whose property we're checking as well.

This would be especially helpful when trying to get more deeply nested values.

```typescript
interface Person {
  name: Name | null;
  pet: {
    name: Name | null;
  }
}
interface Name {
  first: string,
  last: string
}
const person: Person = { name: null, pet: { name: null } };
const petFirstName = of(person).pipe(pluck('pet', 'name', 'first'));
// logs: undefined
```

---

Looking at the source code, [I see this](https://github.com/ReactiveX/rxjs/blob/78032157f5c1655436829017bbda787565b48c30/src/internal/operators/pluck.ts#L59):
```typescript
const mapper = (x: string) => {
  let currentProp = x;
  for (let i = 0; i < length; i++) {
    const p = currentProp[props[i]];
    if (typeof p !== 'undefined') {
      currentProp = p;
    } else {
      return undefined;
    }
  }
  return currentProp;
};
```

Seems like it could be as simple as changing
```typescript
const p = currentProp[props[i]];
```
to
```typescript
const p = currentProp && currentProp[props[i]];
```

**P.S.** Why in the `mapper` function is `x` of type `string`? Wouldn't it be an object?

anirudhvarma12:
**Description:**
Add a fix to prevent the pluck operator from breaking on null or undefined inputs. 

https://github.com/ReactiveX/rxjs/issues/5505

### DIFF
--- a/src/internal/operators/pluck.ts
+++ b/src/internal/operators/pluck.ts
@@ -51,7 +51,7 @@ export function pluck<T, R>(...properties: Array<string | number | symbol>): Ope
   return map((x) => {
     let currentProp: any = x;
     for (let i = 0; i < length; i++) {
-      const p = currentProp?.[properties[i]] ?? undefined;
+      const p = currentProp?.[properties[i]];
       if (typeof p !== 'undefined') {
         currentProp = p;
       } else {
--- a/src/internal/operators/pluck.ts
+++ b/src/internal/operators/pluck.ts
@@ -51,7 +51,7 @@ export function pluck<T, R>(...properties: Array<string | number | symbol>): Ope
   return map((x) => {
     let currentProp: any = x;
     for (let i = 0; i < length; i++) {
-      const p = currentProp[properties[i]];
+      const p = currentProp?.[properties[i]] ?? undefined;
       if (typeof p !== 'undefined') {
         currentProp = p;
       } else {
--- a/spec/operators/pluck-spec.ts
+++ b/spec/operators/pluck-spec.ts
@@ -196,4 +196,14 @@ describe('pluck operator', () => {
     expectObservable(r).toBe(expected, {y: 'abc'});
     expectSubscriptions(a.subscriptions).toBe(asubs);
   });
+
+  it('should not break on null values', () => {
+    const a =   cold('--x--|', {x: null});
+    const asubs =    '^    !';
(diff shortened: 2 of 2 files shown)


src/internal/operators/pluck.ts
```diff
@@ -51,7 +51,7 @@ export function pluck<T, R>(...properties: Array<string | number | symbol>): Ope
   return map((x) => {
     let currentProp: any = x;
     for (let i = 0; i < length; i++) {
-      const p = currentProp[properties[i]];
+      const p = currentProp?.[properties[i]] ?? undefined;
```
cartant, line 54: I don't think the nullish coalescing operator (`??`) should be used here. It will convert values that are explicitly `null` to `undefined`. I think the optional chaining operator (`?.`) is sufficient.

anirudhvarma12: Right, made the change.

cartant: It does mean that if the value is `null`, the operator will emit `null` instead of `undefined`, but I'm okay with that. We'll see what the others think. LGTM.

anirudhvarma12:
@cartant , is there a process to PR getting merged after approval? (Sorry if this is already documented somewhere, I could not find it)

cartant:
PRs that update actual library code - i.e. not just tests or docs - need to be approved and merged by Ben.

benlesh:
Merged also into 6.x, although with a slight alteration, as we don't compile with TS 3.9 there.

PR #5529, opened 2020-06-24
docs(subscription): Improving subscription.add() documentation

TheMcMurder:
**Description:**
Currently the documentation claims that `subscription.add()` will call teardown logic during unsubscribe, this is only part of the story as teardown logic is also called when the subscription terminates for any reason.

### DIFF
--- a/src/internal/Subscription.ts
+++ b/src/internal/Subscription.ts
@@ -108,8 +108,9 @@ export class Subscription implements SubscriptionLike {
   }
 
   /**
-   * Adds a tear down to be called during the unsubscribe() of this
-   * Subscription. Can also be used to add a child subscription.
+   * Adds a tear down to be called when the subscription terminates
+   * (on complete, error, or during unsubscribe()). Can also be used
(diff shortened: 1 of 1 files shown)


TheMcMurder:
Can I get a review on this MR? It's pretty small and I was hoping it would improve the documentation. If it's not a valuable documentation change I'd be happy to tweak it to make it more valuable.

benlesh:
@TheMcMurder Hey, it looks like your work here coincided with other work I was doing on the Subscriptions, and this was already updated in master.

Also, the text your adding is slightly debatable, as Subscriptions can stand on their own outside of an Observable life cycle (although they rarely do). It might make sense to add something that frames this behavior a little more accurately, as you're suggesting. 

I'm also working on a better guide here, #5592, if you'd like to take a look at what I have written over there about subscriptions.

At this time, I think I'm going to let the new docs that were added here simmer for a bit. But I see what you're saying and you've got me thinking about adding more detail here. I'm just not sure this is the right way to phrase it.

Issue #5528, opened 2020-06-23
bufferTime operator is impure


joshribakoff:
## Bug Report

**Current Behavior**

```
import { interval, Subject } from 'rxjs'; 
import { bufferTime, takeWhile, tap } from 'rxjs/operators';


const subject = new Subject<number>()

interval(100).pipe(tap(n => subject.next(n))).subscribe();

const log = batch => {
  // setTimeout(()=>{
    subject.next(99)
  // })
  console.log(batch)
}

const source = subject.pipe(
  takeWhile(x => x < 100),
  bufferTime(1000),
  tap(batch => log(batch))
);

source.subscribe();
```

`99` shows up at end of each array that bufferTime emits, it should not be possible for an operator that comes later to modify a value a previous operator emits. It should show up at the start of the 2nd buffer onwards.

Discovered while unit testing code that pushes onto a buffer in a jest mock callback that was being used in a side effect, after a bufferTime().

**Environment**
- Runtime: both node / chrome
- RxJS version: 6.5.4

**Possible Solution**
don't mutate

cartant:
Yeah, this is a bug. `buffer` doesn't have this problem, as it assigns a new array to the subscriber's property before nexting:

https://github.com/ReactiveX/rxjs/blob/82bcc2179b8aefc08342a419ccc549080c358a64/src/internal/operators/buffer.ts#L84-L86

`bufferTime` does not. This code needs to be rearranged:

https://github.com/ReactiveX/rxjs/blob/82bcc2179b8aefc08342a419ccc549080c358a64/src/internal/operators/bufferTime.ts#L212-L218

joshribakoff:
Could I volunteer to tackle the fix & create the PR?

cartant:
> Could I volunteer to tackle the fix & create the PR?

Sure. If you could create a PR that contains two commits - a failing test that reproduces the problem; and the fix - that would be great. Thanks.

joshribakoff:
This is done. It was a bit more complex than just swapping the order of statements in the code snippet you posted above, unfortunately. I can see some refactoring I'd like to discuss separately after I get my feet more wet, will post more follow up Qs on the PR.

PR #5532, opened 2020-06-26
fix bufferTime impurity

joshribakoff:
**Description:**
Fix bufferTime impurity issue

https://github.com/ReactiveX/rxjs/issues/5528

### DIFF
--- a/spec/operators/bufferTime-spec.ts
+++ b/spec/operators/bufferTime-spec.ts
@@ -1,5 +1,5 @@
 import { of, throwError, interval } from 'rxjs';
-import { bufferTime, mergeMap, take } from 'rxjs/operators';
+import { bufferTime, mergeMap, take, tap } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
 
@@ -38,11 +38,11 @@ describe('bufferTime operator', () => {
       const expected = '--------------------------------x-------------------------------y---(z|)';
       const values = {
         x: ['a', 'b', 'c'],
-        y: ['d', 'e', 'g'],
-        z: [] as string[]
+        y: ['9', 'd', 'e', 'g'],
+        z: ['9']
       };
 
-      const result = e1.pipe(bufferTime(t, null, Infinity, testScheduler));
+      const result = e1.pipe(bufferTime(t, null, Infinity, testScheduler), tap(_value => e1.next('9')));
 
       expectObservable(result).toBe(expected, values);
     });
--- a/spec/operators/bufferTime-spec.ts
+++ b/spec/operators/bufferTime-spec.ts
@@ -88,6 +88,29 @@ describe('bufferTime operator', () => {
     });
   });
 
+  it('should emit buffers at intervals or when the buffer is full, when nexting in subsequent operator', () => {
+    testScheduler.run(({ hot, time, expectObservable, expectSubscriptions }) => {
+      const e1 = hot('  ---a---b---c---d---e---f---g-----|   ');
+      const subs = '    ^--------------------------------!   ';
+      const t = time('  ----------|                          ');
+      const expected = '-------u---v---w---x---y---z-----(c|)';
+      const values = {
+        u: ['a', 'b'],
+        v: ['9', 'c'],
+        w: ['9', 'd'],
+        x: ['9', 'e'],
+        y: ['9', 'f'],
+        z: ['9', 'g'],
+        c: ['9']
+      };
+
+      const result = e1.pipe(bufferTime(t, null, 2, testScheduler), tap(_value => e1.next('9')));
+
+      expectObservable(result).toBe(expected, values);
+      expectSubscriptions(e1.subscriptions).toBe(subs);
+    });
+  });
+
   it('should emit buffers that have been created at intervals and close after the specified delay', () => {
     testScheduler.run(({ hot, time, expectObservable }) => {
       const e1 = hot('       ---a---b---c----d----e----f----g----h----i----(k|)');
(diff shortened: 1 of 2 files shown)


src/internal/operators/bufferTime.ts
```diff
@@ -144,8 +146,14 @@ class BufferTimeSubscriber<T> extends Subscriber<T> {
       const timeSpanOnlyState = { subscriber: this, context, bufferTimeSpan };
       this.add(context.closeAction = scheduler.schedule(dispatchBufferTimeSpanOnly, bufferTimeSpan, timeSpanOnlyState));
     } else {
-      const closeState = { subscriber: this, context };
-      const creationState: DispatchCreateArg<T> = { bufferTimeSpan, bufferCreationInterval, subscriber: this, scheduler };
+      const closeState = { subscriber: this, context, destination: this.destination };
+      const creationState: DispatchCreateArg<T> = {
```
joshribakoff, line 150: Why aren't we using idiomatic JS like closure state? Passing around the explicit state does not seem ideal.

src/internal/operators/bufferTime.ts
```diff
@@ -144,8 +146,14 @@ class BufferTimeSubscriber<T> extends Subscriber<T> {
...
+        bufferTimeSpan,
+        bufferCreationInterval,
+        subscriber: this,
+        scheduler,
+        destination: this.destination
```
joshribakoff, line 155: Unfortunately, because of this pattern of passing explicit state into the scheduler's callback as an argument like this, it means we need to copy in the private member here, so it can "next" onto it in the close action. If there is a better way, let me know. My idea to improve this would be to close over state instead of passing it in.

src/internal/operators/bufferTime.ts
```diff
@@ -209,9 +219,7 @@ class BufferTimeSubscriber<T> extends Subscriber<T> {
   }
 
   closeContext(context: Context<T>) {
-    this.destination.next(context.buffer);
     const contexts = this.contexts;
```
joshribakoff, line 222: No offense, but I've never seen this pattern before. Is there a dogma with the `this` keyword in this project or something? This aliasing seems to be done more out of superstition than logic, and makes the code a bit inscrutable.  I tried to follow existing patterns, but some of them make no sense, and I would like to discuss changing things in future PRs as I cut my teeth w/ this code base.

src/internal/operators/bufferTime.ts
```diff
@@ -231,23 +239,26 @@ function dispatchBufferTimeSpanOnly(this: SchedulerAction<any>, state: any) {
     state.context = subscriber.openContext();
     state.context.closeAction = this.schedule(state, state.bufferTimeSpan);
   }
+
+  state.subscriber.destination.next(prevContext.buffer);
 }
 
 function dispatchBufferCreation<T>(this: SchedulerAction<DispatchCreateArg<T>>, state: DispatchCreateArg<T>) {
```
joshribakoff, line 246: If these were members of the class instead of functions that have been bifurcated, they could also access the private `destination` member of the instance & wouldn't need it passed in.

I feel like there's some good reason why the scheduler was written this way, but it feels like a code smell to me, being naive about what those reasons may be.

src/internal/operators/bufferTime.ts
joshribakoff, line 150: I got a lint error about long line length. It's very distracting seeing red when you're trying to type, and it's just formatting. I am used to automatic code formatting. I hope this manual formatting I chose is ok. It seems odd to me in this day & age that we're not using something like Prettier to avoid subjective & tedious formatting decisions.

src/internal/operators/bufferTime.ts
```diff
@@ -231,23 +239,26 @@ function dispatchBufferTimeSpanOnly(this: SchedulerAction<any>, state: any) {
...
     ));
     action.schedule(state, bufferCreationInterval!);
   }
 }
 
 function dispatchBufferClose<T>(arg: DispatchCloseArg<T>) {
-  const { subscriber, context } = arg;
+  const { subscriber, context, destination } = arg;
   subscriber.closeContext(context);
+  destination.next(context.buffer);
```
joshribakoff, line 263: I don't like that the logic is scattered around this big file, lots of indirection & my changes make it worse. One idea for future improvement to this operator would be to rewrite if from the ground up, reactively, by building on the `buffer(), delay(), count()` operators or something. Right now the implementation is very imperative which makes it not easy to reason about.

spec/operators/bufferTime-spec.ts
```diff
@@ -88,6 +88,29 @@ describe('bufferTime operator', () => {
...
+        w: ['9', 'd'],
+        x: ['9', 'e'],
+        y: ['9', 'f'],
+        z: ['9', 'g'],
+        c: ['9']
+      };
+
+      const result = e1.pipe(bufferTime(t, null, 2, testScheduler), tap(_value => e1.next('9')));
+
+      expectObservable(result).toBe(expected, values);
```
joshribakoff, line 109: As someone working in the codebase for the first time, I "get" marbles & love the idea, but these failure messages are quite obfuscated. I'm sure it's been discussed before, but a more visual output would be nice.

Screen Shot 2020-06-27 at 11 22 34 AM

benlesh:
This is now massively out of date. Closing.

joshribakoff:
@benlesh it does not feel good to put in hard work and have my PR ignored for a year and closed without so much as a thanks or an opportunity to bring it up to date. Presumably the underlying issue still exists too.

benlesh:
I'm sorry if closing this bothered you, I know having PRs closed can be tough. Thank you for your effort. Normally I add a thank you, I guess I was in a hurry.

PR #5540, opened 2020-06-28
Update resources.md w/ my wip yt playlist

joshribakoff:
Adds my YT playlist.

My playlist is still very much a work in progress, and I'm taking a somewhat controversial approach with introducing subjects before observables, but I think for various reasons some people may learn better this way, especially if they think like me. I do discuss all the tradeoffs there of. I've yet to make the videos about higher order observables, error handling, control flow, and walk-throughs of operators, so if there is any objection to adding my work in progress playlist, I'm happy to close this PR / let it wait until the playlist is more complete - not sure if ya'll review the technical content for correctness before linking to it from here, so I understand if you want to close this PR for now, and hopefully we can revisit it later, but I think the content I've recorded so far could be super helpful to some.

**Description:**

### DIFF
--- a/docs_app/content/resources.md
+++ b/docs_app/content/resources.md
@@ -37,3 +37,4 @@ This is a collection of resources that can be found to help with the learning pr
 * RxJS 6 Changes - Overview |  [academind.com](https://www.academind.com/learn/javascript/rxjs-6-what-changed/) | Last Updated: May 3, 2018
 * Introducing RxJS! | [ng-conf 2018](https://www.youtube.com/watch?v=JCXZhe6KsxQ) | April 20, 2018
 * Learn RxJS in 60 Minutes for Beginners | [YouTube](https://www.youtube.com/watch?v=PhggNGsSQyg) | March 30, 2018
+* RxJs & Reactive Programming  | [Youtube](https://www.youtube.com/watch?v=iEJ_Rhkg3_o&list=PLj2oFNVaxfJ8nRFUA2CLyt8TymA0_vQux&index=3&t=0s)


benlesh:
Hi, @joshribakoff, we're still reviewing our process for how/what/why/when we add new things to our community resources. We want to agree upon it, and document it as a team before we start adding more material there.

That said, I _personally_ would be very hesitant to add a YouTube playlist to the rxjs.dev site controlled by someone that isn't a core team member.

benlesh:
I really do appreciate the effort that you've put in here, though. And I hope that many people can benefit from your hard work.

joshribakoff:
Makes sense, I would tend to agree, but I do think my video fills a gap where the RxJs core team has no official playlist yet, and I'm biased, but I'd hope to see my free playlist merge in before some other non core team member's paid for profit playlist, but I totally understand/respect whatever direction ya'll choose. I know maintaining a large OSS project like RxJS has a lot of demands, and any wrong info could cause more tickets/issues for ya'll to deal with - on that note, I'd welcome any constructive feedback too :)

benlesh:
The paid content that currently is there wasn't always paid, we're going to have to put a policy in place.

joshribakoff:
Ah I meant others who opened PRs to add paid content. but yes I also noticed there is existing free content, which promote paid content, as part of putting a policy in place considering if this content is somehow "grandfathered" in may be worth considering. One example of a library that has lots of external links is Redux, for better or worse. There's definitely pros & cons :)

niklas-wortmann:
As described in #5778 we are going to delete the external resources, sorry that it took so long to come up with the decision, but in short we don't have the resources to maintain such a page properly. Feel free to reach out to me if you need further elaboration, but therefore I will close this for now.

Issue #5523, opened 2020-06-23
Marbles cannot represent synchronous-upon-subscription notifications


cartant:
## Bug Report

This is more of a quirk than a bug, but it does mean that some tests in our test suite might be suspect and might not be testing quite what's expected.

In [this PR](https://github.com/ReactiveX/rxjs/pull/5521), it wasn't possible to reproduce a failing test for the bug because a marble source like this:

```
cold("(abcd|)")
```

emits multiple, synchronous notifications at frame zero, but - although the subscription occurs at frame zero - the notifications are not synchronous-upon-subscription.

At some stage, someone - me, I guess - is going to have to look through the tests to see whether or not tests that uses sources like this are testing what they are intended to test - e.g. is [this test](https://github.com/ReactiveX/rxjs/blob/d608cf33791cceb5f330270da96b0b2805223208/spec/operators/share-spec.ts#L209) testing what it's supposed to be testing?

benlesh:
Hi @cartant — thank you for taking the time to open this issue and for sharing your thoughtful feedback with the RxJS project. I’m sorry it took us so long to respond. Given how much time has passed and the lack of recent activity, I’m going to close this for now so the tracker reflects the work we’re likely to act on. If this is still relevant on a current RxJS release, please feel welcome to open a new issue with updated context and a reproduction. Thank you again for your contribution and patience.

— Ben

PR #5549, opened 2020-07-01
fix: revert #5458

cartant:
**Description:**

This PR adds a failing test and reverts #5458 - which removed the `isComplete` variable from `shareReplay`. With the reversion, the test passes. There was an existing test that attempted to test what's tested by the new test, but there was a subtle difference because of this problem: https://github.com/ReactiveX/rxjs/issues/5523

**Related issue (if exists):** #5458

### DIFF
--- a/spec/operators/shareReplay-spec.ts
+++ b/spec/operators/shareReplay-spec.ts
@@ -3,7 +3,7 @@ import * as sinon from 'sinon';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 import { shareReplay, mergeMapTo, retry } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
-import { Observable, Operator, Observer, of, from } from 'rxjs';
+import { Observable, Operator, Observer, of, from, defer } from 'rxjs';
 
 declare const rxTestScheduler: TestScheduler;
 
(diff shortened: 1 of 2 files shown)


josepot:
FYI @cartant @benlesh #5550 also adds the test and it prevents a possible edge-case that I explained in [this comment](https://github.com/ReactiveX/rxjs/pull/5458#issuecomment-652544364).

spec/operators/shareReplay-spec.ts
```diff
@@ -219,6 +219,21 @@ describe('shareReplay operator', () => {
     expectSubscriptions(source.subscriptions).toBe(sourceSubs);
   });
 
+  it('should not restart a synchronous source due to unsubscriptions if refCount is true when the source has completed', () => {
+    // The test above this one doesn't actually test completely synchronous
```
josepot, line 223: I'm 99.9% sure that the test above doesn't have anything to do with the issue that this revert is trying to fix, because the issue with that revert only happens when `refCount` is set to `true`. So, I'm pretty sure that the test above does its job correctly.

josepot:
> There was an existing test that attempted to test what's tested by the new test

I don't think so. Because the issue that this PR is solving only happens when the source completes synchronously upon subscription. AFAIK the current test:

> 'should not restart due to unsubscriptions if refCount is true when the source has completed

doesn't complete synchronously on its first subscription.

Issue #5548, opened 2020-07-01
shareReplay with refCount restarts with synchronous sources


josepot:
## Bug Report

**Current Behavior**
shareReplay with refCount restarts with synchronous sources

**Reproduction**
- REPL or Repo link:

I can't use stackblitz because since the latest release doesn't have #5521 then that bug masks the other one, but with the current code on master, this would happen:

```js
const foo$ = from([1, 2, 3, 4, 5]).pipe(
  shareReplay({ refCount: true, bufferSize: 1})
)

foo$.subscribe(x => {
  console.log(x);
})
// 1
// 2
// 3
// 4
// 5

foo$.subscribe(x => {
  console.log(x);
})
// 1
// 2
// 3
// 4
// 5
```

**Expected behavior**
The second subscribe should only emit `5`

**Possible Solution**
In [this line](https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/shareReplay.ts#L181) also check that the subscription isn't closed:
```ts
if (subscription && !subscription.closed && useRefCount && refCount === 0) {
```

**Additional context**
See [this comment](https://github.com/ReactiveX/rxjs/pull/5458#issuecomment-652466377) from @leggechr and the comments afterwords.

josepot:
I'm closing this issue because it was addressed by #5549

PR #5550, opened 2020-07-01
Fix/share replay ref count sync

josepot:
**Description:**
Bug fix: shareReplay with refCount sync sources shouldn't restart

It addresses #5548


cartant:
For the moment, we've just reverted the change. There is a little more to this, as the reason `subscription` was assigned `undefined` was to 'null out' the reference to the source subscriber (the subscription), but the problem the reverted PR caused exposed a separate issue that probably needs to be looked at: sources that complete synchronously won't end up with an `undefined` subscription and that wasn't the intention. In short, this reference-releasing behaviour needs more thought, so we've reverted it for now.

PR #5551, opened 2020-07-01
Run mode updates 1

benlesh:
Going through and updating more tests to run mode. (Code also formatted)


cartant:
I'm looking forward to all of the tests being converted to `run` mode.

Mostly so that I don't have to review any more of these PRs. 😅

PR #5552, opened 2020-07-01
Prettier

benlesh:
Adds prettier to RxJS, such that:

1. It is run by a pre-commit hook.
2. It *only* runs on files with the `/** @prettier */` pragma at the top, to allow us to slowly migrate over time.
3. It reflects my evil, evil, totalitarian opinions on right and wrong in JavaScript and shall not be contested by anyone for the rest of time because I'm a huge, rotten jerkface! MWUHAHAHAHA!

(No, really I'm just tired of manually formatting things and this is a way forward)

### Updates

- Updates tslint
- Updates tslint config to use `tslint-config-prettier`
- Rebuilds tslint config from the ground up to "work" with what we have in our code base

NOTE: We should probably go through after this and remove almost all of the tslint rule overrides and clean things up, because it makes me really sad. haha.


package.json
```diff
@@ -13,17 +13,16 @@
...
-      "git add"
+    "src/**/*.ts": [
+      "prettier --write"
     ],
+    "*.js": "eslint --cache --fix",
     "*.ts": [
-      "tslint --fix",
-      "git add"
-    ]
+      "tslint --fix"
```
kwonoj, line 21: shouldn't we preserve `git add`? otherwise re-formatted file need to be added manually.

.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "trailingComma": "es5",
+  "singleQuote": true,
+  "requirePragma": true
+}
```
kolodny, line 5: .             | .
:-------------------------:|:-------------------------:
  |

package.json
kolodny: I think the latest lint-staged does this at the end by default

cartant: IIRC, this depends upon the version of `lint-staged` with recent versions not requiring it and warning if it's present. The version in the repo looks pretty old, though.

src/internal/Observable.ts
```diff
@@ -359,15 +402,78 @@ export class Observable<T> implements Subscribable<T> {
...
-  pipe<A, B, C, D, E, F, G, H>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>, op5: OperatorFunction<D, E>, op6: OperatorFunction<E, F>, op7: OperatorFunction<F, G>, op8: OperatorFunction<G, H>): Observable<H>;
-  pipe<A, B, C, D, E, F, G, H, I>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>, op5: OperatorFunction<D, E>, op6: OperatorFunction<E, F>, op7: OperatorFunction<F, G>, op8: OperatorFunction<G, H>, op9: OperatorFunction<H, I>): Observable<I>;
-  pipe<A, B, C, D, E, F, G, H, I>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>, op5: OperatorFunction<D, E>, op6: OperatorFunction<E, F>, op7: OperatorFunction<F, G>, op8: OperatorFunction<G, H>, op9: OperatorFunction<H, I>, ...operations: OperatorFunction<any, any>[]): Observable<unknown>;
+  pipe<A, B>(
```
kolodny, line 405: image

package.json
kwonoj: Yes, either we bump up pkg or need it back.

src/internal/Observable.ts
cartant: And Ben just bought an extra wide monitor! 😂

package.json
```diff
@@ -122,15 +121,17 @@
     "fs-extra": "^8.1.0",
     "glob": "7.1.2",
     "google-closure-compiler-js": "20170218.0.0",
+    "husky": "^4.2.5",
     "klaw-sync": "3.0.2",
-    "lint-staged": "^8.1.6",
+    "lint-staged": "^10.2.11",
```
benlesh, line 126: @kwonoj I updated to the latest here, we don't need to git add anymore.

package.json
benlesh: @kwonoj I added a comment where I updated lint-staged. The git add is no longer required.

src/internal/Observable.ts
jakovljevic-mladen: This looks really ugly IMHO. It doesn't contribute to the code readability at all.

benlesh: This is just the default. I wonder if we can tweak this at all.

It's also possible we could use Clang format. It's what we used at Google, and there you could dictate wrapping by a trailing comma or not.

benlesh:
Apparently, this is at odds with our lint rules. haha.

src/internal/Observable.ts
benlesh: So, digging into it, we *can* make it wider, but honestly... I'm not sure it's a huge win for these cases, it's less readable even on a big screen, because it's harder to count the number of arguments, but it is ... ugly. haha.

PR #5554, opened 2020-07-02
refactor(shareReplay): avoid memory leaks

josepot:
**Description:**
This PR is an attempt to address[ the following comment](https://github.com/ReactiveX/rxjs/pull/5550#issuecomment-652729169) from @cartant on #5550 

> the problem the reverted PR caused exposed a separate issue that probably needs to be looked at: sources that complete synchronously won't end up with an undefined subscription and that wasn't the intention. In short, this reference-releasing behaviour needs more thought, so we've reverted it for now.

Somewhat related to this issue #5548

### DIFF
--- a/src/internal/operators/shareReplay.ts
+++ b/src/internal/operators/shareReplay.ts
@@ -151,28 +151,32 @@ function shareReplayOperator<T>({
   let subject: ReplaySubject<T> | undefined;
   let refCount = 0;
   let subscription: Subscription | undefined;
-  let hasError = false;
-  let isComplete = false;
 
   return function shareReplayOperation(this: Subscriber<T>, source: Observable<T>) {
     refCount++;
     let innerSub: Subscription;
-    if (!subject || hasError) {
-      hasError = false;
+    if (!subject) {
       subject = new ReplaySubject<T>(bufferSize, windowTime, scheduler);
       innerSub = subject.subscribe(this);
       subscription = source.subscribe({
         next(value) { subject!.next(value); },
         error(err) {
-          hasError = true;
-          subject!.error(err);
+          const dest = subject;
+          subscription = undefined;
+          subject = undefined;
+          dest!.error(err);
         },
         complete() {
-          isComplete = true;
           subscription = undefined;
           subject!.complete();
         },
       });
+      // The following condition is needed because source can complete synchronously
+      // upon subscription. When that happens `subscription` is first set to `undefined`
+      // and right after is set to the "closed subscription" returned by `subscribe`
+      if (subscription.closed) {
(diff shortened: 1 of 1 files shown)


src/internal/operators/shareReplay.ts
```diff
@@ -168,19 +167,24 @@ function shareReplayOperator<T>({
...
         complete() {
-          isComplete = true;
           subscription = undefined;
           subject!.complete();
         },
       });
+      // The following condition is needed because source can complete synchronously
+      // upon subscription. When that happens `subscription` is first set to `undefined`
+      // and right after is set to the "closed subscription" returned by `subscribe`
+      if (subscription.closed && !hasError) {
```
cartant, line 177: It shouldn't be necessary to check `hasError`, here. Not nulling-out the `subscription` in the error handler was an oversight. AFAICT, there's no reason the `subscription` should not be set to `undefined` when an error occurs.

```ts
        error(err) {
          hasError = true;
          subscription = undefined;
          subject!.error(err);
        },
```

josepot: You are absolutely right. In fact I also wanted to release the subject in there too, that way we would avoid a possible memory-leak and also the `hasError` flag would no longer be needed... But I thought: "touch as little as possible, because the last time you already broke it".

On the other hand, thanks to the fact that I "broke" it, we became aware of this annoyance, so this time I've decided to be a bit braver and I've done the refactor that I think that this operator needed.

josepot: @cartant have you had a chance to re-review the latest changes? Also, it goes without saying that I don't mind closing this PR if you rather not touch this file unless there is an actual need/bug.

cartant: I'll get to it eventually.

src/internal/operators/shareReplay.ts
```diff
@@ -151,36 +151,40 @@ function shareReplayOperator<T>({
...
       innerSub = subject.subscribe(this);
       subscription = source.subscribe({
         next(value) { subject!.next(value); },
         error(err) {
-          hasError = true;
-          subject!.error(err);
+          const subjectError = subject!.error.bind(subject);
+          subscription = undefined;
+          subject = undefined;
+          subjectError(err);
```
cartant, line 167: This would be simpler:

```ts
const dest = subject;
subscription = undefined;
subject = undefined;
dest!.error(err);
```

josepot: very good point! Thanks!

josepot:
> This looks right, to me, but the `error` callback could be simplified.

💯  done!

PR #5560, opened 2020-07-02
refactor(root): Update to use globalThis

### LINKED ISSUE #5388: Use `globalThis` for next version
It's too late to land this in v7, IMO... but I think we should move to using `globalThis` in our next major, instead of that silly `root` module we have now. It would require a polyfill for `IE`.

benlesh:
- Updates to use globalThis, self, and global, which covers all of our target runtimes
- Fixes code in ajax to have proper types and better runtime checks
- Updates a few tests for types related to root use

resolves #5388

### DIFF
--- a/src/internal/util/root.ts
+++ b/src/internal/util/root.ts
@@ -1,31 +1,21 @@
-declare let global: any;
-
 /**
- * @license
- * Copyright Google Inc. All Rights Reserved.
- *
- * Use of this source code is governed by an MIT-style license that can be
- * found in the LICENSE file at https://angular.io/license
+ * @prettier
  */
 
-declare var WorkerGlobalScope: any;
-// CommonJS / Node have global context exposed as "global" variable.
-// We don't want to include the whole node.d.ts this this compilation unit so we'll just fake
-// the global "global" var for now.
-
-const __window = typeof window !== 'undefined' && window;
-const __self = typeof self !== 'undefined' && typeof WorkerGlobalScope !== 'undefined' &&
-    self instanceof WorkerGlobalScope && self;
-const __global = typeof global !== 'undefined' && global;
-const _root: any = __window || __global || __self;
-
-// Workaround Closure Compiler restriction: The body of a goog.module cannot use throw.
-// This is needed when used with angular/tsickle which inserts a goog.module statement.
-// Wrap in IIFE
-(function () {
-  if (!_root) {
-    throw new Error('RxJS could not find any global context (window, self, global)');
+/**
+ * The global root we can use, given different runtimes
+ */
+export const root: (typeof globalThis | Window & typeof globalThis | NodeJS.Global) = (() => {
+  if (typeof globalThis !== 'undefined') {
+    return globalThis;
   }
+  if (typeof self !== 'undefined') {
+    return self;
+  }
+  if (typeof global !== 'undefined') {
+    return global;
+  }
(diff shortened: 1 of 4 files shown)


kwonoj:
We didn't deprecate ie11 officially yet, so this should mark breaking change for it.

src/internal/util/root.ts
```diff
@@ -1,31 +1,21 @@
...
-// Workaround Closure Compiler restriction: The body of a goog.module cannot use throw.
-// This is needed when used with angular/tsickle which inserts a goog.module statement.
-// Wrap in IIFE
-(function () {
-  if (!_root) {
-    throw new Error('RxJS could not find any global context (window, self, global)');
+/**
+ * The global root we can use, given different runtimes
+ */
+export const root: (typeof globalThis | Window & typeof globalThis | NodeJS.Global) = (() => {
```
cartant, line 8: I don't think we can use these types, here. AFAICT, if we use `Window`, we will force Node users to use the `dom` TypeScript library and if we use `NodeJS.Global` we will force developer's targeting the browser to install `@types/node`.

benlesh: Hmmm... you might be right here. However, I think it will all depend on what they pull in, wouldn't it? I'm not sure the best way to verify this TBH. But I'm not really thinking too hard at the moment. haha.

benlesh:
> We didn't deprecate ie11 officially yet, so this should mark breaking change for it.

`self` works in ie11, so this should work for ie11.

src/internal/util/root.ts
benlesh: A lot of this is just usage in the Ajax Library. All in all, I think it would be good to just get rid of the `root` module entirely, and I'm sure I've filed issues to that effect. Perhaps we should handle these global access issues on a case-by-case basis in individual modules instead of having one module to rule them all. Globals, after-all, by nature, "global", so we don't really need to handle this in a centralized place.

I'll think on it.

benlesh: Okay, so just using the types around `XMLHttpRequest` and `fetch` _alone_ will require node users to pull in the `dom` stuff for TypeScript. I'm not sure what to do about that. Perhaps all of that stuff should be moved to another package?

benlesh: (In other words, I think this is a pre-existing condition to our library with regards to TS)

cartant: > Okay, so just using the types around XMLHttpRequest and fetch alone will require node users to pull in the dom stuff for TypeScript.

I'm not sure what this means. AFAICT, there is no current requirement to use the `"dom"` lib in Node when using APIs other than the `XMLHttpRequest` stuff and `fetch`.

Having to use the `"dom"` lib in conjunction with RxJS APIs that use the `XMLHttpRequest` stuff and `fetch` would be expected, IMO, but it shouldn't be needed if an application does not use said APIs.

If the `root` business is used only with `XMLHttpRequest` stuff and `fetch`, that's fine, but I don't see why the Node types would be necessary. What's in the `"dom"` lib should suffice.

benlesh: Yes, currently, it's only used within `fromFetch` and `ajax`. Maybe at some point we should just start publishing that as a separate package to avoid confusion, I'm not sure.

benlesh:
I'm closing this in favor of #5579

PR #5564, opened 2020-07-03
docs: Add the start of a glossary

benlesh:
This is an effort on my part to clearly define the language we use (and should be using) in our documentation and our articles. This is an important step to improving our resources in my opinion, and a long time coming.

### DIFF
--- a/docs_app/content/guide/glossary-and-semantics.md
+++ b/docs_app/content/guide/glossary-and-semantics.md
@@ -62,7 +62,7 @@ The [producer](#Producer) has encountered a problem and is notifying the [consum
 
 ### Complete
 
-The [producer](#Producer) is notifying the [consumer](#Consumer) that it is done [nexting](#Next) values, without error, will send no more values, and intends to [teardown](#Teardown). [Completion](#Complete) cannot occur after an [error](#Error), or [unsubscribe](#Unsubscription). [Complete](#Complete) cannot be called twice. Complete will always happen before [teardown](#Teardown).
+The [producer](#Producer) is notifying the [consumer](#Consumer) that it is done [nexting](#Next) values, without error, will send no more values, and it will [teardown](#Teardown). [Completion](#Complete) cannot occur after an [error](#Error), or [unsubscribe](#Unsubscription). [Complete](#Complete) cannot be called twice. [Complete](#Complete), if it occurs, will always happen before [teardown](#Teardown).
 
 ### Notification
 
@@ -78,7 +78,7 @@ The act of one [producer](#Producer) being [observed](#Observation) by **many** 
 
 ### Unicast
 
-The act of one [producer](#Producer) being [observed](#Observation) **only one** [consumer](#Consumer). An observable is "unicast" when it only connects one [producer](#Producer) to one [consumer](#consumer). Unicast doesn't necessarily mean ["cold"]([#3254](https://github.com/ReactiveX/rxjs/issues/3254)
+The act of one [producer](#Producer) being [observed](#Observation) **only one** [consumer](#Consumer). An observable is "unicast" when it only connects one [producer](#Producer) to one [consumer](#consumer). Unicast doesn't necessarily mean ["cold"](#Cold).
 
 ### Cold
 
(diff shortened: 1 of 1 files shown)


kwonoj:
This is great :+1:

docs_app/content/guide/glossary-and-semantics.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Inner Source
+
+One, of possibly many [sources](#Source), which are [subscribed](#Subscribe) to automatically within a single [subscription](#Subscription) to another observable. Examples of an "inner source" include the [observable inputs](#Observable_Inputs) returned by the mapping function in a [mergeMap](/api/operators/mergeMap) [operator](#Operator). (e.g. `source.pipe(mergeMap(value => createInnerSource(value))))`, were `createInnerSource` returns any valid [observable input](#Observable_Input)).
+
+### Partial Observer
+
+An [observer](#Observer) that lacks all necessary [notification](#Notification) handlers. Generally these are supplied by user-land [consumer](#Consumer) code. A "full observer" or "observer" would simply be an observer than had all [notification](#Notification) handlers.
```
cartant, line 122: > An [observer](#Observer) that lacks all necessary [notification](#Notification) handlers.

An [observer](#Observer) that lacks *some* [necessary] [notification](#Notification) handlers.

I would also favour omitting *necessary* as they are not really necessary, if they don't *have* to be provided.

docs_app/content/guide/glossary-and-semantics.md
```diff
@@ -0,0 +1,136 @@
...
+
+Some of what we discuss is conceptual. These are mostly common traits of behaviors that can manifest in observables or in push-based reactive systems.
+
+### Multicast
+
+The act of one [producer](#Producer) being [observed](#Observation) by **many** [consumers](#Consumer).
+
+### Unicast
+
+The act of one [producer](#Producer) being [observed](#Observation) **only one** [consumer](#Consumer). An observable is "unicast" when it only connects one [producer](#Producer) to one [consumer](#consumer). Unicast doesn't necessarily mean ["cold"]([#3254](https://github.com/ReactiveX/rxjs/issues/3254)
```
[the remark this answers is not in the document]
jnizet: Missing **by** before "only one consumer".
The link at the end doesn't look good:
Capture d’écran 2020-07-03 à 08 12 15

docs_app/content/guide/glossary-and-semantics.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Cold
+
+An observable is "cold" when it creates a new [producer](#Producer) during [subscribe](#Subscribe) for every new [subscription](#Subscription). As a result, a "cold" observables are _always_ [unicast](#Unicast), being one [producer](#Producer) [observed](#Observation) by one [consumer](#Consumer). Cold observables can be made [hot](#Hot) but not the other way around.
```
jnizet, line 85: The "a" should be removed in "a cold observables are always"

kwonoj:
>  what term are we going to use for functions like timer that create and return an observable?

I usually used `(observable / static) creation method`, though it's a bit verbose.

docs_app/content/guide/glossary-and-semantics.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Operator Function
+
+A function that takes an [observable](#Observable), and maps it to a new [observable](#Observable). Nothing more, nothing less. Operator functions are created by [operators](#Operator). If you were to call an rxjs operator like [map](/api/operators/map) and put the return value in a variable, the returned value would be an operator function.
+
+### Operation
+
+An action taken while handling a [notification](#Notification), as set up by an [operator](#Operator) and/or [operator function](#Operator_Function). In RxJS, a developer can chain several [operator functions](#Operator_Function) together by calling [operators](#Operator) and passing them to [pipe](/api/index/class/Observable#pipe), which results in a new [observable](#Observable). During [subscription](#Subscription) to that observable, operations are performed in an order dictated by the [observation chain](#Observation_Chain).
```
jnizet, line 103: "passing them to pipe": the phrasing is confusing: you don't pass them to pipe. You pass the operator functions that they have created. So maybe rephrase to

> In RxJS, a developer can chain several operator functions together by calling operators, and passing the created operator functions to pipe.

docs_app/content/guide/glossary-and-semantics.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Source
+A [observable](#Observable) or [valid observable input](#Observable_Inputs) having been converted to an observable, that will supply values to another [observable](#Observable), either as the result of an [operator](#Operator) or other function that creates one observable as another. This [source](#Source), will be the [producer](#Producer) for the resulting [observable](#Observable) and all of its [subscriptions](#Subscriptions). Sources may generally be any type of observable.
```
jnizet, line 106: This section is confusing to me. 
I guess "that creates one observable as another" should be "that creates one observable **from** another". And since that's the difinition of an operator function, I guess that "as the result of an operator" should be "as the result of an operator function".

docs_app/content/guide/glossary-and-semantics.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Observable Inputs
+
+A "observable input" ([defined as a type here](/api/index/type-alias/ObservableInput)), is any type that can easily converted to an [Observable](#Observable). Observable Inputs may sometimes be referred to as "valid observable sources".
```
jnizet, line 110: that can easily **be** converted

cartant:
> ... static creation methods ...

They're not methods, though. They are functions. A method is called on an object. From the [ECMAScript spec](https://www.ecma-international.org/ecma-262/5.1/#sec-4.3.27):

> method
> function that is the value of a property
> ... When a function is called as a method of an object, the object is passed to the function as its this value.

kolodny:
This is amazing!! One small suggestion I had to make this more approachable for most folks (even if we're targeting "power" users: Can we include a giant snippet (or many small snippets) where we label things while giving concrete examples. I'm thinking of some code like this

```ts
import { Observable } from 'rxjs'; 

const thisIsTheObservable = new Observable(function thisIsTheProducer(thisIsTheObserver) {
  thisIsTheObserver.next('this value is being nexted');
  setTimeout(() => {
    thisIsTheObserver.complete();
  }, 1000);
  return function thisIsTheTeardown() {
    console.log('tearing down');
  }
})

const thisIsThePartialObserver = {
  next(value) { console.log('got:', value); },
  complete() { console.log('finished') }
}
const thisIsTheSubsription = thisIsTheObservable.subscribe(thisIsThePartialObserver);

const doneWithObservable = Math.random() > 0.5;

if (doneWithObservable) {
  thisIsTheSubsription.unsubscribe();
}
```

I'm sure some of the terms are not going to fit into this format but a large chunk will and will make those terms less esoteric.

docs_app/content/guide/glossary-and-semantics.md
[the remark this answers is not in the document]
benlesh: No, that's a crazy typo. I'm not sure how that even got in there. I must have pivoted to do something else and accidentally copy-pasted in there or something. It was supposed to be `[cold](#Cold)`

benlesh:
@kolodny That's a good idea. However, this glossary is largely to cover concepts and terms that don't necessarily map 1:1 with specific variables, methods, or functions in the library. It's more to cover terms that will be used to describe those things. As the docs are improved, we will end up linking both directions, from the API docs to here, and from here to the API docs. 

If I can think of more generic examples, I'll add them.

benlesh:
I'm going to merge this so I can move on with docs improvements that link here. It seems we have a quorum, or at the very least, general approval, and iterating on docs is much easier than iterating on the library itself.

PR #5575, opened 2020-07-05
refactor(buffer): rename buffer to bufferWith

tylersaunders:
**Description:** renames the buffer operator to bufferWith and marks buffer as deprecated

**Related issue (if exists):** #5296

### DIFF
--- a/docs_app/content/guide/operators.md
+++ b/docs_app/content/guide/operators.md
@@ -152,11 +152,11 @@ These are Observable creation operators that also have join functionality -- emi
 
 ### Transformation Operators
 
-- [`buffer`](/api/operators/buffer)
 - [`bufferCount`](/api/operators/bufferCount)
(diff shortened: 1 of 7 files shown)


tylersaunders:
Looks like `api_guardian` is failing, but I'm not sure how to resolve this. Hoping the owners can point me in the right direction, as I haven't been able to find any documentation around it, and it's not mentioned in `COLLABORATORS.md`

benlesh:
I think this one is a near-miss, unfortunately.

The ones we are marking "___With" are the operators that have a static creation function counterpart.  `concat` and `concatWith`, or `merge` and `mergeWith`. Unfortunately, there's no static creation function for `buffer`.

benlesh:
Thank you so much for your effort, @tylersaunders. It is appreciated!

PR #5576, opened 2020-07-05
Fix broken Docs build

niklas-wortmann:
#5422 introduced a subtle bug in the docs build process. This pr mainly reverts the async rehype plugin process which was introduced to load the marble svgs at build time and inline them in the html ast. For some reason this affects the build that the navigation.json and api-list.json are not generated anymore. Therefore everything used to work if the file was existing before. 

Now svg files are not inlined anymore but used with a common img tag, which is also a11y a little bit easier at least.   

Additionally, I did some minor fixes:
 - fix missing h1 in operators guide
 - create nav entry for glossary
 - remove toc from glossary as we can leverage the existing outline

Sorry I know I should have created multiple commits but this build error was super frustrating so I became lazy :D


docs_app/content/license.md
```diff
@@ -1,23 +0,0 @@
-@title
-@description
-The MIT License
-
-Copyright (c) 2014-2018 Google, Inc.
```
benlesh, line 5: I'm not sure we can remove this. We're still using parts of the documentation app that Google built, and therefor the license applies. However, we can probably change this to include "And Contributors" since other contributors have added code to this, and we're also MIT licensed. Or even `Google, Inc., RxJS Team Members and Contributors` or whatever you like.

benlesh:
Also, rebasing will resolve your build issues.

niklas-wortmann:
this branch is up to date with master and the api_guardian tests also fail locally in master for me

benlesh:
Okay. I'm not worried about it. The failure is unrelated to your PR. I can fix it separately.

PR #5579, opened 2020-07-06
refactor: remove root

benlesh:
removes the root module entirely.


src/internal/observable/dom/AjaxObservable.ts
```diff
@@ -344,7 +351,7 @@ export class AjaxSubscriber<T> extends Subscriber<Event> {
           const { progressSubscriber } = (<any>xhrProgress);
           progressSubscriber.next(e);
         };
-        if (root.XDomainRequest) {
+        if (XDomainRequest) {
```
kwonoj, line 354: will this work node.js uses polyfills? it seems like node 14 at least doesn't seem to like it 
image

we've been guiding user to use polyfills on node.js for ajaxobservable, I believe numbers will be quite small though.

src/internal/observable/dom/AjaxObservable.ts
```diff
@@ -21,34 +20,42 @@ export interface AjaxRequest {
...
+function isFormData(body: any): body is FormData {
+  return typeof FormData !== 'undefined' && body instanceof FormData;
+}
+
 function getCORSRequest(): XMLHttpRequest {
-  if (root.XMLHttpRequest) {
-    return new root.XMLHttpRequest();
-  } else if (!!root.XDomainRequest) {
-    return new root.XDomainRequest();
+  if (XMLHttpRequest) {
```
cartant, line 32: With these no longer being property accesses, I think tests like `if (typeof XMLHttpRequest === 'undefined`) {` will be necessary to avoid errors being thrown.

src/internal/observable/dom/AjaxObservable.ts
benlesh: If the polyfill puts something on `global`, it _should_ work. I suppose I could make it a little less volatile though

src/internal/observable/dom/AjaxObservable.ts
benlesh: added

kwonoj:
ugh autorebase doesn't work for forked :/

PR #5584, opened 2020-07-10
fix(fromEvent): improve typings for JQueryStyleEventEmitter

evertbouw:
**Description:**

The handler in JQueryStyleEventEmitter is typed as a Function, meaning it takes no arguments. Overwriting this requires a pretty awkward cast. 

https://github.com/ReactiveX/rxjs/pull/4496

### DIFF
--- a/spec/operators/mergeMap-spec.ts
+++ b/spec/operators/mergeMap-spec.ts
@@ -1,137 +1,140 @@
+/** @prettier */
 import { expect } from 'chai';
 import { mergeMap, map, delay, take } from 'rxjs/operators';
 import { asapScheduler, defer, Observable, from, of, timer } from 'rxjs';
-import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 import { asInteropObservable } from '../helpers/interop-helper';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
 
-declare const asDiagram: Function;
-
 /** @test {mergeMap} */
 describe('mergeMap', () => {
-  let rxTest: TestScheduler;
+  let testScheduler: TestScheduler;
 
-  // TODO: Convert the rest of these tests to use run mode!
   beforeEach(() => {
-    rxTest = new TestScheduler(observableMatcher);
+    testScheduler = new TestScheduler(observableMatcher);
   });
 
   it('should map-and-flatten each item to an Observable', () => {
-    const e1 =    hot('--1-----3--5-------|');
-    const e1subs =    '^                  !';
-    const e2 =   cold('x-x-x|              ', {x: 10});
-    const expected =  '--x-x-x-y-yzyz-z---|';
-    const values = {x: 10, y: 30, z: 50};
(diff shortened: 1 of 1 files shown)


src/internal/observable/fromEvent.ts
```diff
@@ -20,8 +20,8 @@ export interface NodeCompatibleEventEmitter {
 }
 
 export interface JQueryStyleEventEmitter {
-  on: (eventName: string, handler: Function) => void;
-  off: (eventName: string, handler: Function) => void;
+  on: (eventName: string, handler: NodeEventHandler) => void;
```
benlesh, line 23: `NodeEventHandler` appears to be from `@typings/node`. @cartant, am I wrong in saying that this would make us dependent on `@typings/node`?

evertbouw: Since that was already used in this file I thought I'd use it again. The important part is that it's `(...args: any[]) => void`, could create a local type for that.

cartant: > ... this would make us dependent on `@typings/node`?

Nah, this looks okay. The `NodeEventHandler` is declared in this file. I suppose this is okay, but I though jQuery events used a single object parameter - more like the DOM event handlers - rather than multiple parameters. It has been some time since I've used jQuery, so IDK.

Whatever the situation, `Function` is wrong, here, as it's pretty much totally useless, these days, being declared like this:

```ts
interface Function {
    /**
     * Returns the name of the function. Function names are read-only and can not be changed.
     */
    readonly name: string;
}
```

benlesh:
I'd feel a whole lot better about this change with some sort of dtslint tests or something.

evertbouw:
I've not written those before, does this work?

cartant:
@evertbouw  Regarding the dtslint tests, once https://github.com/ReactiveX/rxjs/pull/5724 is merged, you'll find a `spec-dtslint/observables/fromEvent-spec.ts` file. It contains the type tests. Note that the code is not executed; the dtslint expectations - `$ExpectType ...` - are performed using a TSLint rule. To run the tests, use the `npm run dtslint` command.

cartant:
As I mentioned [here](https://github.com/ReactiveX/rxjs/pull/5584/files#r459852071), I think the jQuery handler type could/should be changed to better reflect the actual jQuery typings - i.e. be more like `EventHandlerBase`, but without the `this`/context.

Relevant types: [`EventHandlerBase`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/847731ba1d7fa6db6b911c0e43aa0afe596e7723/types/jquery/misc.d.ts#L6395-L6408) and [`on`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/847731ba1d7fa6db6b911c0e43aa0afe596e7723/types/jquery/JQuery.d.ts#L7621-L7627).

However, this is jQuery, so it's not something about which I'm especially passionate.

evertbouw:
Rebased and uncommented the test

cartant:
It seems these changes were somehow (inadvertently?) included in this PR: https://github.com/ReactiveX/rxjs/pull/5783

Issue #5411, opened 2020-04-28
reusing publish or publishReplay in a pipe can subscribe to the wrong data source


e-davidson:
## Bug Report

**Current Behavior**

When you reuse a pipe with publish or publishReplay all uses of the pipe share the same instance of ConnectableObservable.
Which causes the connection to connect when the other one connected and gets data from the other source.

```
const pipeLine = pipe(
  map((x: number) => x * x),
  publish(),
);

const obs1 = src1.pipe(pipeLine);
const obs2 = src2.pipe(pipeLine);

obs1.subscribe(d => console.log('sub 1', d));
obs2.subscribe(d => console.log('I should not be running', d));

obs1.connect();
```

This is what you get on the console 

```
sub 1 1
I should not be running 1
sub 1 4
I should not be running 4
sub 1 9
I should not be running 9
sub 1 16
I should not be running 16
sub 1 25
I should not be running 25
```
`obs2` which was never connected has been connected and it's also connected to the wrong source.

**Reproduction**

https://stackblitz.com/edit/rxjs-nrcyfu

**Expected behavior**

reusing pipes should not effect which source connects to which subscription.

Only the 1st subscription should run as the 2nd one hasn't been connected yet.
Also the 2nd subscription should not be subscribed to the first observable.

**Workaround**

Use a factory method instead.

```
    const pipeLine = () =>  pipe(
      map((x: number) => x * x),
      publish(),
    );
```

**Possible Solution**

for publishReplay
https://github.com/ReactiveX/rxjs/blob/259e5cd2e8149bc5db14cc0216afec7131e8aa11/src/internal/operators/publishReplay.ts

Instead of creating the replay subject on line 22 instead it can be created in the call back on line 24. That way each source observable would have it's own replay subject.

for publish

https://github.com/ReactiveX/rxjs/blob/259e5cd2e8149bc5db14cc0216afec7131e8aa11/src/internal/operators/publish.ts 

On line 61 we can return a factory that takes the source instead of calling multicast directly.

cartant:
Yeah, this looks like a bug to me. Thanks.

kawazoe:
I have just stumbled on this issue as well in a project I am working on.

[...]

PR #5585, opened 2020-07-10
fix(publish,publishReplay) resolve sharing Subject when subscribing to different source observables

e-davidson:
**Description:**

Fixes the issue when using the publish or publishReplay operator in a pipe that's shared with different source observables.

For more details see #5411 

- change publish operator to use factory
- change publishReplay operator to not share ReplaySubject

- fixes issue #5411

### DIFF
--- a/spec/operators/publish-spec.ts
+++ b/spec/operators/publish-spec.ts
@@ -1,7 +1,7 @@
 import { expect } from 'chai';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 import { publish, zip, mergeMapTo, mergeMap, tap, refCount, retry, repeat } from 'rxjs/operators';
-import { ConnectableObservable, of, Subscription, Observable } from 'rxjs';
+import { ConnectableObservable, of, Subscription, Observable, pipe } from 'rxjs';
 
 /** @test {publish} */
 describe('publish operator', () => {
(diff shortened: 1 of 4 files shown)


cartant:
This behaviour is by design and these changes would be breaking.

e-davidson:
I'd be happy to make it backwards compatible. It's unclear where the breaking change would occur since the tests didn’t break. Would you be able to point me in the right direction?

cartant:
The change to `src/internal/operators/publishReplay.ts` looks to me to be breaking. Changing it to use a factory will effect different behaviour, AFAICT.

In any case, I'm these APIs are likely to be deprecated and replaced. See #5432, #5431 and other issues/PRs linked therein.

cartant:
I shouldn't have blocked this, @e-davidson

It does significantly change behaviour, but it's not behaviour that anyone could reasonably be relying upon - so it's more of a bug fix than a breaking change. At least that's what I'm thinking ATM, so we should be able to introduce this fix into v7 (and maybe backport it to v6).

I've cherry-picked your commits into a separate PR. I've resolved the conflicts and I've made similar changes to other operators that have the same issue. Thanks for your work. And your patience.

Closing this to address it in #6410 - which includes your work.

e-davidson:
@cartant 
Thank you for moving this forward and I'm glad I was able to assist.

PR #5589, opened 2020-07-13
refactor: Remove rxSubscriber symbol

benlesh:
- Gets rid of `SubjectSubscriber`, as it was a confusing bit of dancing around inheritence quirks
- Adds basically the same thing as `SubjectSubscriber`, only with a better name (`FullObserverSubscriber`) in a spot that makes more sense, and with documentation.
- Changes `ConnectableObservableSubscriber` to just override `destination` directly instead of going through the old `SubjectSubscriber`
- Fixes code in `toSubscriber` to ensure that anything that is a valid subscriber will just get used directly.


src/internal/util/toSubscriber.ts
```diff
@@ -24,3 +25,29 @@ export function toSubscriber<T>(
...
+ */
+class FullObserverSubscriber<T> extends Subscriber<T> {
+  /**
+   * Creates an instance of a subscriber
+   *
+   * NOTE: The `protected destination` is a tricky override of the
+   * destination on the original `Subscriber` class. This whole
+   * class may look unnecessary, but it's doing this trick to get around
+   * some of the ugly inheritance stuff we have going on in the library.
+   */
```
cartant, line 49: So the purpose of `destination`, here, is to ensure the instance's `destination` property is the full observer that was passed to the ctor, bypassing the `SafeSubscriber` wrapping that would happen within the `Subscriber` implementation?

benlesh: That's right. There was a weird round-about way of doing this before through the `SubjectSubscriber`.

Issue #5545, opened 2020-06-29
Input Wanted: Move to using AbortSignal rather than Subscriptions


benlesh:
Now that it looks like Node [will be getting AbortController/AbortSignal](https://docs.google.com/document/d/19lEj6h1xDe1I2lEvW_HSoyYG1OfAJO2PLe4dNE7gJ0w/edit?fbclid=IwAR0O3_XAFeOLbQ0crRD_DhYGGNr5heif9JVvzKpvzWXFnXo7wGf2JmZ5T0w) I think it's time we seriously start planning moving RxJS toward using `AbortSignal`.

I have done some research and work on a naive implementation here: https://github.com/benlesh/trex

### CONS:

- This will be a huge breaking change that requires a lot of lead time and long planning to execute.
- `AbortSignal` uses `EventTarget` (`addEventListener` and `removeEventListener`), which is really gross.
- The word "Abort" and its variations everywhere is a little unpalatable, IMO, but we didn't really have a say in the naming of the thing, I guess.

### PROS:

- We can drastically simplify internal code and APIs
- We no longer have to ship a Subscription type
- We can also simplify external APIs
- We are being good citizens with the rest of the JS ecosystem

### Proposed (Future) API:

Here is where I'd like to go with this:

1. `Observable` `subscribe` and `forEach` both accept an optional `AbortSignal` argument.
2. `Observable`'s initialization function will have a second argument to it that is the `AbortSignal`.
3. The (very) slow phase out of `Subscription`.

### Observable initialization function changes:

The idea here would be to pass the `signal` into the initialization function to allow users to check it directly for whether or not
it had been aborted, and also to allow them to register teardown for the producer.

In the asynchronous case:

```ts
const onePerSecond = new Observable((subscriber) => {
  let n = 0;
  const id = setInterval(() => subscriber.next(n++), 1000);
  return () => {
    clearInterval(id);
  }
});

// becomes

const onePerSecond = new Observable((subscriber, signal) => {
  let n = 0;
  const id = setInterval(() => subscriber.next(n++), 1000);
  signal.addEventHandler('abort', () => clearInterval(id));
});
```

In the synchronous case:
```ts
const range1To100 = new Observable((subscriber) => {
  for (let i = 1; i <= 100 && !subscriber.closed; i++) {
    subscriber.next(i);
  }
  subscriber.complete();
});

// becomes

const range1To100 = new Observable((subscriber, signal) => {
[...]

PR #5591, opened 2020-07-15
Feature: AbortSignal support

benlesh:
This is a flag in the ground for adding support for `AbortSignal` to RxJS.

This is a non-breaking change that adds support for passing AbortSignal to both `subscribe` and `forEach` calls of Observable. See commits.

Related #5545 and #3122

### DIFF
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -71,6 +71,7 @@ export class Observable<T> implements Subscribable<T> {
   }
 
   subscribe(observer?: PartialObserver<T>): Subscription;
+  subscribe(observer: PartialObserver<T> | null | undefined, signal: AbortSignal): Subscription;
   /** @deprecated Use an observer instead of a complete callback */
   subscribe(next: null | undefined, error: null | undefined, complete: () => void): Subscription;
   /** @deprecated Use an observer instead of an error callback */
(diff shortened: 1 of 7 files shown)


benlesh:
NOTE: This _does not yet_ add passing of an `AbortSignal` as an argument to the initialization function of `Observable`. That would definitely be planned, as it would improve `fetch` support and support for other APIs using `AbortSignal`, but I didn't add it at this time, as I was going for the minimal change to get cancellation working.

benlesh:
We really need to get the circle CI build stuff off of here. As it's no longer relevant. haha

src/internal/Observable.ts
```diff
@@ -307,28 +358,57 @@ export class Observable<T> implements Subscribable<T> {
...
+            // says there is supposed to be an AbortError, but I haven't found any
+            // implementations in any runtimes. For now, we can document this
+            // as the users needing to check `err.name === 'AbortError'` and
+            // adjust over time.
+            const err = new Error('Abort exception');
+            err.name = 'AbortError';
+            reject(err);
+          },
+          { once: true }
+        );
```
voliva, line 390: Maybe I'm missing something, but doesn't this need cleanup? `once: true` will clean the event listener if it fires, but it seems that if the observable naturally completes, then this would keep listening to the AbortSignal, as there's nothing removing it.

benlesh: You're absolutely right! Thank you for catching that!

src/internal/Observable.ts
```diff
@@ -10,6 +10,7 @@ import { toSubscriber } from './util/toSubscriber';
 import { observable as Symbol_observable } from './symbol/observable';
 import { pipeFromArray } from './util/pipe';
 import { config } from './config';
+import { values } from 'lodash';
```
cartant, line 13: WAT? `lodash`? This looks like an auto-import artefact.

benlesh:
Related #5649

benlesh:
This is outdated, but remains on our radar. I'll close it for now.

PR #5595, opened 2020-07-16
Modified package.json to use shorter repository location syntax

dhrumilrana:
Like mentioned in the npm package.json specification [here](https://docs.npmjs.com/files/package.json): 
`For GitHub, GitHub gist, Bitbucket, or GitLab repositories you can use the same shortcut syntax you use for npm install`

This also resolves a problem when installing via CITGM.


benlesh:
> This also resolves a problem when installing via CITGM.

What's the issue there? I'm curious to know what problem we're solving.

Otherwise, I can't imagine why we'd want to have a less specific repository setting (the URL is pretty specific, and by definition, "universal")

MylesBorins:
CITGM explicitly looks for `repository.url`, so while I think CITGM will still work after this change I want to point out that any CITGM failures will be unrelated to this change

https://github.com/nodejs/citgm/blob/master/lib/lookup.js#L46

cartant:
Closing because the comments above suggest that this is unnecessary.

Issue #4164, opened 2018-09-21
`toArray` operator like that does not wait for the stream to be complete before emitting


maxime1992:
## Feature Request

**(If this is new operator request) describe reason it should be core operator**
`reduce` waits for the observable to finish before emitting the result.
`scan` does the same job but passes values as they come.

`toArray` waits for the observable to finish before emitting the result.
I think there should be an operator similar to `toArray` that passes the array as it keeps receiving values.

To go around that issue, we can use scan an push within a new array but it feels like a core operator might be welcome.

kwonoj:
As you already mentioned scan is verasatile operator and it should be trivial to implement behavior as you suggested. Is there specific reason you think 1. Current operator chain can't solve 2. Requires separate api surface for core?

maxime1992:
I understand that adding a new operator might be seen as something "bad" because rxjs already has a lot of them. But thinking of `reduce` and `scan` I would've expect to have something similar for `toArray` and couldn't find it. So just raising the question.

Otherwise I guess we could also juste remove `reduce` and do `scan` + `last` for example.

kwonoj:
For new operators, we do like to get elaborations :

1. how trivially it can be achieved with existing operators? is it requires lot of complex, or near impossible?
2. if so, can it be made as user level operator module, or it should be dealt with some of internals that should be in core operator?

What we trying to achieve / recommend is having more user level api surface anyone can opt into, instead of core library contains all operators which unnecessary burdens users for having all operators. This isn't about bytesize, more about general approachability anyone need to look into core api surface.

For that reason, we don't pair all operator set, sometimes even deprecate some while semantically it makes more sense to having it, also from v5 number of operators already being deprecated from v4.

In short, I can understand motivation but not sure given suggestion meets direction we'd like to encourage. I'd like to get some more elaborations if it can fit above 2 conditions. Otherwise, I'd suggest to publish as user level module. (https://github.com/ReactiveX/rxjs/blob/master/doc/operator-creation.md#publish-your-operator-as-a-separate-library)

maxime1992:
Replying to both points:

1) how trivially it can be achieved with existing operators? 
```
obs$.pipe(
  scan((acc, value) => {
    acc.push(value);
    return acc;
}, [])
```

so it's not really complex IMO.

2) can it be made as user level operator module

Certainly.

It's mostly for consistency that I'm raising that issue and I totally understand your point of view :+1:

kwonoj:
[...]

Issue #3585, opened 2018-04-17
Create Observable from Promise function


ajcrites:
It is a very common pattern -- especially when interoperating with libraries that use promises -- to depend on some Promise/async functionality to create an Observable. For example, you might need to get some value from a storage library that implements Promises:

```ts
const loadData = async () => {
  const token = await storage.get('token');
  return ajax({ url: '/data', headers: { token } });
}
```

`loadData` returns a promise, so you have a Promise that returns an Observable. If you want to use this with other Observables you will have to use `fromPromise`. You could also do something like:

```ts
// or `fromPromise` with an IIFE
const loadData = () => defer(async () => {
  const token = await storage.get('token');
  return ajax({ url: '/data', { headers: { token } }));
}).pipe(switchAll());

// or ... but if you have a lot of async data to pass through it becomes uglier
const loadData = () => fromPromise(storage.get('token')).pipe(
  switchMap(token => ajax({ url: '/data', { headers: { token } }))
);
```

Since this is such a common pattern, I think it would make sense to have an Observable creation function that creates an Observable returned from a Promise. I think that using `defer` would be cool, but that would also probably overload a lot of `defer`'s functionality and may not be possible with the current implementation, so the best name I can come up with is `deferAsync`.

```ts
const loadData = () => deferAsync(async () => {
  const token = await storage.get('token');
  const token2 = await storage.getFromToken('token2', token);
  const token3 = await storage.getFromToken('token3', token2);

  return ajax({ url: '/data', { headers: { token3 } }));
});
```

This would effectively return the inner `ajax` Observable once the promise resolved. It doesn't necessarily have to function like `defer` where it returns a new Observable on each subscription, but I'm not sure how easy / hard it would be to do either way and `deferAsync` probably isn't the best name for it.

I think that the existing solutions to this problem as stated in this issue are okay ... but I think that this additional method can simplify some of the interop between typical Promise-based workflows and libraries and ease the transition to using Observable patterns for some developers while keeping things a bit cleaner.

cartant:
Using your first solution, a user-land `deferAsync` function is simple to implement:

```ts
const deferAsync = <T>(factory: () => Promise<Observable<T>>) => defer(factory).pipe(mergeAll());
```

`mergeAll` can be used instead of `switchAll`, as the promise will resolve only once, so there'll never be a need to switch.

ajcrites:
[...]

PR #5601, opened 2020-07-22
feat(filterExists): Create the filterExists operator

norato:
**Description:**

It creates a new filter operator, that only emits a value from the source if it is not nullable


kwonoj:
Appreciate for taking time to create changes, but as suggested in our issue template https://github.com/ReactiveX/rxjs/blob/master/.github/ISSUE_TEMPLATE/feature_request.md and refs like https://github.com/ReactiveX/rxjs/issues/3585#issuecomment-382571620 / https://github.com/ReactiveX/rxjs/issues/4164#issuecomment-423603887 we strongly recommend to create userland module for operator before discuss to add into core api surface. 

Reading operator implementation I believe this should be able to implement without requiring core's internal and would encourage to publish it as userland module.

norato:
@kwonoj Thanks, for the information
I'll gonna create the userland module, but I trying to find this link you put at the linked comment and it is broken since version 5.
Should we put back this file?

https://github.com/ReactiveX/rxjs/blob/5.x/doc/operator-creation.md

benlesh:
@norato this was a _ton_ of work. And I think it's super awesome that you're contributing and thinking about these things. I really hope you publish this useful operator to npm and people get a lot of use out of it. If it's something that catches on, we'll definitely consider adding it to the core library.

For now, I'm very sorry that we'll need to pass on it though. I truly hope that this doesn't dissuade you from contributing in the future!

Thank you so much again!

PR #5603, opened 2020-07-23
refactor(tap): rename DoOperator to TapOperator

jakovljevic-mladen:
Renaming `DoOperator` to `TapOperator` to follow the rest of the `tap` operator stuff: the `tap` itself is no longer `do` operator and the `TapSubscriber` is [no longer](https://github.com/ReactiveX/rxjs/commit/cd9626a4f93cac6f631d5a97dd9c9b2aa8e4b5db) `DoSubscriber`.

This is a leftover that some may love to keep, but if possible, it would be nice to have this merged so that `tap` is now completely renamed.


benlesh:
Thanks!

Issue #5475, opened 2020-06-09
TimestampProviders and TestScheduler "run mode"


benlesh:
Currently, `TestScheduler` "run mode" will patch all schedulers by default. This means that everywhere we were only using schedulers as `TimestampProvider`s, will use the test scheduler's currently time when `now()` is called.

However... we're going to be breaking people's tests with a move to use `Date` or `performance` as default `TimestampProviders`.

I see two solutions:

1. Have `TestScheduler` temporarily monkey patch `Date.now` and `performance.now` to accommodate bare uses of things like `animationFrames`, etc.
2. Create _very, very light_ wrappers around `Date.now` and `Performance.now` that are used internally (but not exported, yet) that we monkey patch when we are in run mode.

I'm inclined to say that #2 is the best answer, and the least likely to break others.

cartant:
Pretty sure this is closed by #5607.

PR #5607, opened 2020-07-26
feat(TestScheduler): add an animate "run mode" helper

cartant:
**Description:**

This PR:

* Adds a `Date` timestamp provider.
* Adds a `performance` timestamp provider.
* Adds a `requestAnimationFrame` provider.
* All providers have `delegate` properties that are `undefined` by default but are used - by `TestScheduler#run` - to override the default behaviour. This mirrors the `delegate` implementation detail that's used with the schedulers.
* Removes the arrow function from the static `Scheduler.now` - it's unnecessary, as the `Date.now` call now occurs within the `dateTimestampProvider`.
* The `animationFrames` observable now uses the `requestAnimationFramesProvider`.
* `TestScheduler#run` sets the delegates for the providers, with the `requestAnimationFramesProvider`'s delegate working in conjunction with a `animate()` marble that's added to the run helpers:

    ```ts
    testScheduler.run(({ animate }) => {
      animate('---x---x---x---x---x');
      /* ... */
    });
    ```

    Each time an animation/repaint occurs, all queued `rAF` callbacks will be executed - in the order in which they were scheduled - and each will be passed the same timestamp - the `TestScheduler`'s `now` value - which is also what will be returned from the timestamp providers - as the `TestScheduler` itself is the delegate.

* Adds tests for the `animate` `run` helper.
* Updates API guardian for the inclusion of the `animate` `run` helper.
* Reimplements the `animationFrames` tests to use the `run` helper and the provider - which is spied upon - and removes the `rAF` mock.
* Fixes the `rAF` provider so that cancellation of already fulfilled requests (upon unsubscription) is not attempted.

**Related issue (if exists):** #5438 and #5475

### DIFF
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -618,7 +618,7 @@ export declare class VirtualAction<T> extends AsyncAction<T> {
     protected recycleAsyncId(scheduler: VirtualTimeScheduler, id?: any, delay?: number): any;
     protected requestAsyncId(scheduler: VirtualTimeScheduler, id?: any, delay?: number): any;
     schedule(state?: T, delay?: number): Subscription;
-    static sortActions<T>(a: VirtualAction<T>, b: VirtualAction<T>): 1 | 0 | -1;
+    static sortActions<T>(a: VirtualAction<T>, b: VirtualAction<T>): 0 | 1 | -1;
 }
 
 export declare class VirtualTimeScheduler extends AsyncScheduler {
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -618,7 +618,7 @@ export declare class VirtualAction<T> extends AsyncAction<T> {
     protected recycleAsyncId(scheduler: VirtualTimeScheduler, id?: any, delay?: number): any;
     protected requestAsyncId(scheduler: VirtualTimeScheduler, id?: any, delay?: number): any;
     schedule(state?: T, delay?: number): Subscription;
-    static sortActions<T>(a: VirtualAction<T>, b: VirtualAction<T>): 0 | 1 | -1;
+    static sortActions<T>(a: VirtualAction<T>, b: VirtualAction<T>): 1 | 0 | -1;
 }
 
 export declare class VirtualTimeScheduler extends AsyncScheduler {
--- a/src/internal/testing/TestScheduler.ts
+++ b/src/internal/testing/TestScheduler.ts
@@ -405,22 +405,25 @@ export class TestScheduler extends VirtualTimeScheduler {
     return testMessages;
   }
 
-  run<T>(callback: (helpers: RunHelpers) => T): T {
-    const prevFrameTimeFactor = TestScheduler.frameTimeFactor;
-    const prevMaxFrames = this.maxFrames;
+  private createAnimator() {
+    if (!this.runMode) {
+      throw new Error('animate() must only be used in run mode');
+    }
 
-    TestScheduler.frameTimeFactor = 1;
-    this.maxFrames = Infinity;
-    this.runMode = true;
-    AsyncScheduler.delegate = this;
-    dateTimestampProvider.delegate = this;
-    performanceTimestampProvider.delegate = this;
+    // The TestScheduler assigns a delegate to the provider that's used for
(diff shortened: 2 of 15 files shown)


api_guard/dist/types/index.d.ts
```diff
@@ -618,7 +618,7 @@ export declare class VirtualAction<T> extends AsyncAction<T> {
     protected recycleAsyncId(scheduler: VirtualTimeScheduler, id?: any, delay?: number): any;
     protected requestAsyncId(scheduler: VirtualTimeScheduler, id?: any, delay?: number): any;
     schedule(state?: T, delay?: number): Subscription;
-    static sortActions<T>(a: VirtualAction<T>, b: VirtualAction<T>): 1 | 0 | -1;
+    static sortActions<T>(a: VirtualAction<T>, b: VirtualAction<T>): 0 | 1 | -1;
```
benlesh, line 621: LOL.

src/internal/testing/TestScheduler.ts
```diff
@@ -409,14 +413,62 @@ export class TestScheduler extends VirtualTimeScheduler {
     this.maxFrames = Infinity;
     this.runMode = true;
     AsyncScheduler.delegate = this;
+    dateTimestampProvider.delegate = this;
+    performanceTimestampProvider.delegate = this;
 
-    const helpers = {
+    let animationFramesHandle = 0;
+    let animationFramesQueue: Map<number, FrameRequestCallback> | undefined;
+
```
benlesh, line 421: Be sure to add comments as to the "why" around here. We don't do that enough in our library, I think, and while it might be obvious to us right now, maybe it won't be 5 years from now.

src/internal/testing/TestScheduler.ts
```diff
@@ -409,14 +413,62 @@ export class TestScheduler extends VirtualTimeScheduler {
...
+        return handle;
+      },
+      cancelAnimationFrame(handle) {
+        if (!animationFramesQueue) {
+          throw new Error("repaints() was not called within run()");
+        }
+        animationFramesQueue.delete(handle);
+      }
+    };
+    const repaints = (marbles: string) => {
```
benlesh, line 438: I can easily have my mind changed on this... but I think we should probably move this out to have it in the same place as all the other helpers. I know that makes it available for "non-run-mode" use, and maybe we want to avoid that.. but I do like consistency.

benlesh, line 438: this also is going to need documentation comments.

cartant:
@benlesh I'll make those changes and will then add some docs - I was holding off on documenting it until it had been reviewed.

src/internal/testing/TestScheduler.ts
benlesh: Oh... on more thing, I'm not sure I like `repaints`, as it's not readily obvious that it is associated to "animation frames". Perhaps a different name here would be better?  `animates`? `animationFrames`? (the latter might be too confusing though).

cartant: > Perhaps a different name here would be better? 

How about `animate`? It doesn't return anything, so I suppose it doesn't need the plural noun that would usually be used for an observable.

cartant:
@benlesh I've made some changes:

* I've changed the name of the `repaints` run helper to `animate`.
* I've moved the delegate and the run helper implementation into a `createAnimator` method. This is more consistent with the current implementation.
* I've added a TSDoc comment to `run` - there was no existing comment/documentation - and I've linked it to an updated marble testing guide.

ATM, the documentation in the guide is a little light, but I think it's enough, for now. Once this PR is merged, I am going to:

* rebase and fix up the PR that adds the `timestamp` and `elapsed` properties to the `animationFrames` observable's value.
* look at removing/refactoring the `delegate` that's in `AsyncScheduler`, as I think to ought to be possible to add `setInterval` and `setImmediate` providers (and delegates). Doing that would allow for tests to actually use all of the schedulers within 'run mode'. ATM, they are bypassed and there is no way to write a test in which `AsyncScheduler` tasks are executed *after* subsequently-scheduled `AsapScheduler` tasks - i.e. there is no differentiation between virtual micro and macro tasks in the `TestScheduler`.

If the docs need more work, I can look at them after I've done the above. Or someone else can do it.

benlesh:
Nice work, @cartant! I'm very pleased to see this land!

PR #5610, opened 2020-07-30
refactor: Reduce memory consumption and simplify inner and outer subscription

benlesh:
- Reduces memory pressure by no longer capturing outer values where they are not needed. For example, every inner subscription in mergeMap was retaining the outer value, if the outer value was a large array, that could add up
- Makes subscribeToResult less ridiculous, I am not sure what I was thinking there 🤦 
- Leaves a few operators with more complex inner subscriptions, to be refactored later.
- Removes some redundant code that removed inner subscriptions manually, this was necessary in the early days of RxJS 5, but has not been necessary in a long time

### Explainer

So the old `OuterSubscription` is now called `ComplexOuterSubscription`, and `InnerSubscription` is `ComplexInnerSubscription`.

Basically, back when we had all sorts of additional "resultSelectors" (pre v6), those result selectors relied on data that we captured with the `InnerSubscription`. After removing resultSelectors, that was no longer needed in the majority of cases, BUT it was still there to keep this "one inner subscription process to rule them all" sort of thing we had going.

Unfortunately, those trapped outer values can be costly.  Imagine this scenario:

```ts
import { of, timer } from 'rxjs';
import { mergeMap } from 'rxjs';

const data = new Array(1000000).fill('test');

of(data).pipe(
  mergeMap(d => timer(1000000))
)
.subscribe(console.log);
```

In the above scenario, the subscription to `of(data)`, as a _source_ begins and ends synchronously. **BUT!!!** _even though we don't even use that huge array_ in the inner observable we return in the `mergeMap`, that array is _retained_ in the `InnerSubscriber`'s `outerValue` property, which is created by the merge map... and held, in this case, for 1,000 seconds! Not great.

This PR solves that by no longer retaining the `outerValue` unless necessary.

There's a long tail of one-offs that we'll need to address. But they are commented on the `ComplexOuterSubscriber` implementation and its methods.


benlesh:
So the old `OuterSubscription` is now called `ComplexOuterSubscription`, and `InnerSubscription` is `ComplexInnerSubscription`.

Basically, back when we had all sorts of additional "resultSelectors" (pre v6), those result selectors relied on data that we captured with the `InnerSubscription`. After removing resultSelectors, that was no longer needed in the majority of cases, BUT it was still there to keep this "one inner subscription process to rule them all" sort of thing we had going.

Unfortunately, those trapped outer values can be costly.  Imagine this scenario:

```ts
import { of, timer } from 'rxjs';
import { mergeMap } from 'rxjs';

const data = new Array(1000000).fill('test');

of(data).pipe(
  mergeMap(d => timer(1000000))
)
.subscribe(console.log);
```

In the above scenario, the subscription to `of(data)`, as a _source_ begins and ends synchronously. **BUT!!!** _even though we don't even use that huge array_ in the inner observable we return in the `mergeMap`, that array is _retained_ in the `InnerSubscriber`'s `outerValue` property, which is created by the merge map... and held, in this case, for 1,000 seconds! Not great.

This PR solves that by no longer retaining the `outerValue` unless necessary.

There's a long tail of one-offs that we'll need to address. But they are commented on the `ComplexOuterSubscriber` implementation and its methods.

src/internal/innies-and-outies.ts
```diff
@@ -0,0 +1,114 @@
+/** @prettier */
```
cartant, line 1: IMO, this file should be named something else, as "innies and outies" isn't a comparison that's made only between belly buttons and this has the potential to make some people uncomfortable.

src/internal/innies-and-outies.ts
```diff
@@ -0,0 +1,114 @@
...
+import { Subscription } from './Subscription';
+import { Subscriber } from './Subscriber';
+import { Observable } from './Observable';
+import { subscribeTo } from './util/subscribeTo';
+
+export interface ISimpleOuterSubscriber<T> {
```
cartant, line 7: I trivial nit - as the interface is used only in this file - but I'd go with the TypeScript convention of using a `Like` suffix over an `I` prefix for interfaces/shapes. IIRC, we do this for the library's exported types. Whether or not it's done here doesn't really matter, though.

src/internal/innies-and-outies.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  /**
+   * @param _innerSub Used by: race, bufferToggle, delayWhen, windowToggle, windowWhen
+   */
+  notifyComplete(_innerSub: ComplexInnerSubscriber<T, R>): void {
+    this.destination.complete();
+  }
+}
+
+export function subscribeToResult2(result: any, innerSubscriber: Subscriber<any>): Subscription | undefined {
```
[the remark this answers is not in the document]
benlesh: Haha... I was tired and forgot that... I had two versions while I was refactoring. I've renamed it to `innerSubscribe`, since it should only be used for inner subscriptions.

src/internal/operators/exhaustMap.ts
```diff
@@ -92,65 +90,51 @@ class ExhaustMapOperator<T, R> implements Operator<T, R> {
...
  */
-class ExhaustMapSubscriber<T, R> extends OuterSubscriber<T, R> {
-  private hasSubscription = false;
+class ExhaustMapSubscriber<T, R> extends SimpleOuterSubscriber<T, R> {
+  private innerSubscription?: Subscription;
   private hasCompleted = false;
   private index = 0;
 
-  constructor(destination: Subscriber<R>,
+  constructor(protected destination: Subscriber<R>,
```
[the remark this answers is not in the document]
benlesh: You're right, but we were already doing this in several spots in the library, and the runtime implications are innocuous. I think we can fix this everywhere in another PR.

src/internal/innies-and-outies.ts
benlesh: Yeah, I don't like the `I`, much either, I guess.

src/internal/innies-and-outies.ts
benlesh: I'm not sure I want to be friends with people who are uncomfortable about bellybuttons.

But I'll change it.

PR #5613, opened 2020-07-30
refactor: Improve memory pressure

benlesh:
this is the duplicate of #5610, it is refactoring to ensure outer values are not retained when they do not have to be. It needs to be done in a separate PR because the branches diverge just enough to require it. This PR also has some mild, internal type fixes.

### DIFF
--- a/src/internal/innerSubscribe.ts
+++ b/src/internal/innerSubscribe.ts
@@ -1,0 +1,114 @@
+/** @prettier */
+import { Subscription } from './Subscription';
+import { Subscriber } from './Subscriber';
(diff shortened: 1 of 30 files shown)


benlesh:
Note that the changes are not 1:1 with the other PR. I have to do things like leave `InnerSubscriber` and `OuterSubscriber` in the same spot with the same names for rxjs-compat. I could have refactored and aliased, but it's not really worth the effort and extra risk.

benlesh:
The TS3 test isn't really supposed to pass for 6.x. I'm not sure why we're hinging on it.

Issue #5614, opened 2020-07-31
Wrong typings with `tap` operator (for the observer overload)


josepot:
## Bug Report

**Current Behavior**
The typings of the observer overload of the `tap` operator can incorrectly return partial observable

**Reproduction**
- REPL or Repo link:
https://codesandbox.io/s/nice-varahamihira-vuf19?file=/src/index.ts

```ts
const test$ = new Subject<number>();                             
const one$: Observable<number> = of(1);                          
const somethingIsWrong$ = one$.pipe(tap(test$));
```
the inferred type of `somethingIsWrong$` is `Observable<number | undefined>`

**Expected behavior**
~If the observer is not a "PartialObserver", then the returned Observable shouldn't be partial.~
`tap` should always return the same type as its source.

**Environment**
- Runtime: All
- RxJS version: 6.6.0

josepot:
As @cartant pointed out in [this comment](https://github.com/ReactiveX/rxjs/pull/5615#issuecomment-666817460) this issue happens because of an issue with the types of `Subject`.

For instance, in v6 when defining a `new Subject<number>()` the types of `next` look like this `Subject<number>.next(value?: number | undefined): void`, which it's something that has been addressed on v7.

However, the `tap` operator should always return the same type as its source and that's an issue that also affects v7. For instance consider [the following code](https://codesandbox.io/s/busy-hawking-r27c5?file=/src/index.ts:0-268):

```ts
import { Observable, of } from "rxjs";
import { tap } from "rxjs/operators";

const one$: Observable<number> = of(1);

const result$: Observable<number> = one$.pipe(
  tap({
    next(value: string | number) {
      console.log(value);
    }
  })
);
result$.subscribe()
```

`result$`'s type should be `Observable<number>` rather than `Observable<string | number>`.

Therefore, we have 2 different issues:
- the types of `tap` which affects v6 and v7
- the types of `Subject`.

After thinking a bit more about it, I think that the actual fix for solving the issue with the types of the `tap` operator should be to change [this line](https://github.com/ReactiveX/rxjs/blob/3db2eaa5d6bfe0188b0c4da3626123c8eefedf98/src/internal/operators/tap.ts#L17) like this:

```ts
export function tap<T, P extends T>(observer: PartialObserver<P>): MonoTypeOperatorFunction<T>;
```

I will try to send a PR with that change before the end of the weekend.

cartant:
[...]

PR #5615, opened 2020-07-31
fix(tap): typings for observer overload of tap

josepot:
**Description:**
An attempt to fix the issue described on #5614 

 #5614

### DIFF
--- a/src/internal/operators/tap.ts
+++ b/src/internal/operators/tap.ts
@@ -1,7 +1,7 @@
 import { Operator } from '../Operator';
 import { Subscriber } from '../Subscriber';
 import { Observable } from '../Observable';
-import { MonoTypeOperatorFunction, PartialObserver, TeardownLogic } from '../types';
+import { MonoTypeOperatorFunction, Observer, PartialObserver, TeardownLogic } from '../types';
 import { noop } from '../util/noop';
 import { isFunction } from '../util/isFunction';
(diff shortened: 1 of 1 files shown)


cartant:
FYI, this should include a dtslint test - ideally in a separate commit that's fixed by a change included in a subsequent commit.

josepot:
> FYI, this should include a dtslint test - ideally in a separate commit that's fixed by a change included in a subsequent commit.

Yep, sorry! Also, I just realized that this is not the correct fix... Because, I'm pretty sure that `tap` should never alter the type of the source, right?

So, I will close this PR, I will wait for the bug to be confirmed and I will try to come up with a better solution if the issue actually gets confirmed. Thanks!

cartant:
I suspect this is v6-only and relates to this signature messing up the inference:

https://github.com/ReactiveX/rxjs/blob/075af28f2567188d6feb0ee26b4a91a5dcded3ed/src/internal/Subject.ts#L61

That signature has changed in v7 - i.e. `master`.

josepot:
> I suspect this is v6-only and relates to this signature messing up the inference:

Makes sense! Also, I tested it and you are correct, this issue is fixed on [`7.0.0-beta.3`](https://codesandbox.io/s/jolly-mclean-oe95e?file=/src/index.ts)

PR #5616, opened 2020-07-31
fix(tap): typings for observer overload of tap

### LINKED ISSUE #5614: Wrong typings with `tap` operator (for the observer overload)

josepot:
**Description:**
It fixes #5614

### DIFF
--- a/spec-dtslint/operators/tap-spec.ts
+++ b/spec-dtslint/operators/tap-spec.ts
@@ -11,6 +11,12 @@ it('should accept partial observer', () => {
   const c = of(1, 2, 3).pipe(tap({ complete: () => { } })); // $ExpectType Observable<number>
 });
 
+it('should not alter the type of the source', () => {
+  const a = of(1).pipe(tap({next: (x: number | string) => {}})); // $ExpectType Observable<number>
+  const b = of(1).pipe(tap((x: number | string) => {})); // $ExpectType Observable<number>
+  const c = of(1).pipe(tap((x: number | string) => {}, (e: any) => {}, () => {})); // $ExpectType Observable<number>
+})
+
 it('should not accept empty observer', () => {
   const a = of(1, 2, 3).pipe(tap({})); // $ExpectError
 });
--- a/src/internal/operators/tap.ts
+++ b/src/internal/operators/tap.ts
@@ -12,9 +12,9 @@ export function tap<T>(next: null | undefined, error: null | undefined, complete
 /** @deprecated Use an observer instead of an error callback */
 export function tap<T>(next: null | undefined, error: (error: any) => void, complete?: () => void): MonoTypeOperatorFunction<T>;
 /** @deprecated Use an observer instead of a complete callback */
-export function tap<T>(next: (value: T) => void, error: null | undefined, complete: () => void): MonoTypeOperatorFunction<T>;
-export function tap<T>(next?: (x: T) => void, error?: (e: any) => void, complete?: () => void): MonoTypeOperatorFunction<T>;
-export function tap<T>(observer: PartialObserver<T>): MonoTypeOperatorFunction<T>;
+export function tap<T, P extends T>(next: (value: P) => void, error: null | undefined, complete: () => void): MonoTypeOperatorFunction<T>;
+export function tap<T, P extends T>(next?: (x: P) => void, error?: (e: any) => void, complete?: () => void): MonoTypeOperatorFunction<T>;
+export function tap<T, P extends T>(observer: PartialObserver<P>): MonoTypeOperatorFunction<T>;
 /* tslint:enable:max-line-length */
 
 /**


benlesh:
This is actually no longer an issue in version 7. The real issue happened to be with the typing of `Subject`'s `next` method... which was `next(value?: T)` and should have been `next(value: T)`. That's where it's getting the `undefined | T` from.

This PR would actually make the types of `tap` _less_ correct, at least in `master`. I'm not sure of the implications of this fix if we were to apply it in 6.x. I suspect in 6.x, the fix might just be to provide a special case for `tap<T>(subject: Subject<T>): MonoTypeOperatorFunction<T>` or something to that effect.

josepot:
Hi @benlesh ,

Thanks for the quick reply. I made [this comment](https://github.com/ReactiveX/rxjs/issues/5614#issuecomment-667193616) on #5614 explaining why I think that this fix also applies to v7 (master).

> This PR would actually make the types of tap less correct, at least in master.

Maybe I'm missing something, but I think that the types of the `tap` observer should not alter the type of the source observable. Could you please provide an example where this would make the types less correct?

cartant:
> Without these changes the following test would fail:
>
> ```ts
> it('should not alter the type of the source', () => {
>   const a = of(1).pipe(tap({next: (x: number | string) => {}})); // $ExpectType Observable<number>
>   const b = of(1).pipe(tap((x: number | string) => {})); // $ExpectType Observable<number>
>   const c = of(1).pipe(tap((x: number | string) => {}, (e: any) => {}, () => {})); // $ExpectType Observable<number>
> })
> ```

IMO, the code in this snippet is incorrect - or, at least, not good. It deliberately specifies a type that's different - wider! - than the source type. In general, IMO, it is good practice to avoid explicit types wherever possible and it is avoidable here. If the explicit types are omitted, inference behaves as you would expect.

josepot:
> IMO, the code in this snippet is incorrect - or, at least, not good. It deliberately specifies a type that's different - wider! - than the source type. In general, IMO, it is good practice to avoid explicit types wherever possible and it is avoidable here. If the explicit types are omitted, inference behaves as you would expect.

I agree that in general it's a good practice to avoid these situations. However, from the point of view of the types, I think that this makes them more correct.

Also, when using `tap` there are instances when it's very convenient to use wider types. For example a logger for debugging that looks like this:

```ts
const logId = (tag: string) => (value: {id: string}) => {
  console.log(`${tag} - id: ${id}`);
}
```
if we piped `tap(logId('user'))` then the returning observable would be of type `{id: string}`, which would be very unfortunate.

cartant:
IMO, the most important thing is for the RxJS types to be as simple as possible and for them to be pragmatic. I would rather them be simple and cover 95% of the use cases. Also, IMO, your using-`tag`-for-debugging example would be better implemented as a user-land operator - to which you only passed the tag.

cartant:
An alternative solution for your second example is to make the `log` function generic, so that it can participate in the inference, rather than drive it:

```ts
import { of } from "rxjs";
import { tap } from "rxjs/operators";

function log<T>(value: T) {
  /* whatever */
}
const answer = of(42).pipe(tap(value));
```

Fundamentally, the explicit types in what's passed to `tap` are what's effecting/driving the inference and that's just the way TS behaves. I am not in favour of complicating the types solely to effect different inference behaviour. From past experience, this is something we should simply accept and live with.

josepot:
I think that I have failed at explaining the reason why this PR can only make things better and never worse.

One last example before I close the PR:
```ts
const logCorrelationId = (value: {correlationId: string}) => {
  console.log(`${value.correlationId}: ${JSON.strigify(value, null, 2)});
}

someObservableWithACorrelationId$.pipe(
  tap(logCorrelationId),
  switchMap(requestData) // => currently this would fail, with these changes it would be just fine
)
```

> Fundamentally, the explicit types in what's passed to tap are what's effecting/driving the inference and that's just the way TS behaves.

Because of the current TS definitions, not because those are the real types of `tap`. In reality the `tap` operator never alters the type of the source observable, which is what this PR ensures. If nowadays TS provides the means to have better type inference, I don't quite understand why we choose not to benefit from that.

>  I am not in favour of complicating the types solely to effect different inference behaviour.

I don't think that these changes complicate them, I think that they make them more accurate.

> From past experience, this is something we should simply accept and live with.

I understand your position, but I respectfully disagree :slightly_smiling_face:

All that being said, you are the maintainers and I really appreciate you taking the time to review this and to get back to me. I'm sorry if I made myself "too pushy" about something this small.

I will leave this PR open for a couple more days in case that you change your mind. Once again, thanks a lot for your time and patience.

cartant:
My previous comments have dealt with what it is you have been trying to do and why I am not in favour. Further to those comments - and this touches on what Ben mentioned - the changes are demonstrably incorrect, for example:

```ts
import { of, OperatorFunction } from "rxjs";
import { tap } from "rxjs/operators";

declare module "rxjs/internal/operators/tap" {
  export function tap<T, P extends T>(
    next: (value: P) => void
  ): OperatorFunction<T, T>;
}

function f<T>(t: T) {
  /* whatever */
}
const answer = of(42).pipe(tap((value: 54) => { /* whatever */ }));
```

This should not be valid - as there is no guarantee that value will be `54` - but no TS compilation error is effected. In this snippet, `P` is `54` which extends - i.e. narrows - `number`.

josepot:
> This should not be valid - as there is no guarantee that value will be 54 - but no TS compilation error is effected. In this snippet, P is 54 which extends - i.e. narrows - number.

Fair enough. Thanks for showing me that, I didn't think about that case.

Once again, thanks a lot for taking the time to explain this.

PR #5620, opened 2020-08-03
fix(repeat): Ensure teardown happens between repeated synchronous obs…

benlesh:
…ervables

- Simplifies repeat code
- Removes reliance on unsubscribeAndRecycle
- Gets rid of strange, undocumented behavior where numbers less than zero were treated like positive infinity
- Fixes an issue where `finalize` would be called N times after the completion of the resulting observable if the source was synchronous.
- Fixes an issue where N unsubscriptions would always occur after the total completion.

BREAKING CHANGE: Removed an undocumented behavior where passing a negative count argument to `repeat` would result in an observable that repeats forever.

### DIFF
--- a/src/internal/operators/repeat.ts
+++ b/src/internal/operators/repeat.ts
@@ -67,7 +67,8 @@ export function repeat<T>(count = Infinity): MonoTypeOperatorFunction<T> {
   return (source: Observable<T>) =>
     count <= 0
       ? EMPTY
-      : new Observable<T>((subscriber) => {
+      : lift(source, function (this: Subscriber<T>, source: Observable<T>) {
+          const subscriber = this;
           let soFar = 0;
           const subscription = new Subscription();
           let innerSub: Subscription | null;
--- a/src/internal/operators/repeat.ts
+++ b/src/internal/operators/repeat.ts
@@ -72,7 +72,7 @@ export function repeat<T>(count = Infinity): MonoTypeOperatorFunction<T> {
           let soFar = 0;
           const subscription = new Subscription();
           let innerSub: Subscription | null;
-          const subscribeNext = () => {
+          const subscribeForRepeat = () => {
             let syncUnsub = false;
             innerSub = source.subscribe({
               next: (value) => subscriber.next(value),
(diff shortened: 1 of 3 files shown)


benlesh:
I seriously considered removing `lift` with this one... but we can wait for v8.

I suspect that `retry` has the same issue... but I have not checked yet.

src/internal/operators/repeat.ts
```diff
@@ -60,47 +62,44 @@ import { lift } from '../util/lift';
...
+export function repeat<T>(count = Infinity): MonoTypeOperatorFunction<T> {
+  return (source: Observable<T>) =>
+    count <= 0
+      ? EMPTY
+      : lift(source, function (this: Subscriber<T>, source: Observable<T>) {
+          const subscriber = this;
+          let soFar = 0;
+          const subscription = new Subscription();
+          let innerSub: Subscription | null;
+          const subscribeNext = () => {
```
cartant, line 75: You mention elsewhere, IIRC, that the stack traces are weird. Would giving this function a name that's more closely related to the operator be helpful? Something like `subscribeForRepeat`? Or `repeatSubscribeImpl`? Or something like that?

src/internal/operators/repeat.ts
```diff
@@ -60,47 +62,44 @@ import { lift } from '../util/lift';
...
+            let syncUnsub = false;
+            innerSub = source.subscribe({
+              next: (value) => subscriber.next(value),
+              error: (err) => subscriber.error(err),
+              complete: () => {
+                if (++soFar < count) {
+                  if (innerSub) {
+                    subscription.remove(innerSub);
+                    innerSub.unsubscribe();
```
cartant, line 84: Is there a reason `innerSub` is not nulled here?

I think that if it isn't, there will be a problem if the source is async on the first subscription and sync on a subsequent subscription - as `innerSub` won't be `null` for the subsequent completion, despite that completion being synchronous.

src/internal/operators/repeat.ts
benlesh: Good idea.

src/internal/operators/repeat.ts
benlesh: nope... just didn't think of it. Also, that `remove` is superfluous, teh `unsubscribe` will do that...

PR #5621, opened 2020-08-03
fix(retry): Ensure teardown happens before resubscription with synchr…

benlesh:
…onous observables

Related: #5620

- Resolves an issue where all teardowns would not execute until the result observable was complete if the source was synchronous

BREAKING CHANGE: Removed an undocumented behavior where passing a negative count argument to `retry` would result in an observable that repeats forever.

### DIFF
--- a/src/internal/operators/retry.ts
+++ b/src/internal/operators/retry.ts
@@ -1,9 +1,10 @@
-import { Operator } from '../Operator';
 import { Subscriber } from '../Subscriber';
 import { Observable } from '../Observable';
 
-import { MonoTypeOperatorFunction, TeardownLogic } from '../types';
+import { MonoTypeOperatorFunction } from '../types';
 import { lift } from '../util/lift';
+import { Subscription } from '../Subscription';
+import { EMPTY } from '../observable/empty';
 
 export interface RetryConfig {
   count: number;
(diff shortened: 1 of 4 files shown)


benlesh:
In fact, I'd bet there's an issue like this with `retryWhen`, `repeatWhen` and `catchError` as well.

src/internal/operators/retry.ts
```diff
@@ -58,62 +59,55 @@ export interface RetryConfig {
...
-      if (count === 0) {
-        return super.error(err);
-      } else if (count > -1) {
-        this.count = count - 1;
+  return (source: Observable<T>) => count <= 0 ? EMPTY: lift(source, function (this: Subscriber<T>, source: Observable<T>) {
+    const subscriber = this;
+    let soFar = 0;
+    const subscription = new Subscription();
+    let innerSub: Subscription | null;
+    const subscribeNext = () => {
```
cartant, line 78: As with `repeat`, this could have a more operator-specific name - for more obvious stack traces.

src/internal/operators/retry.ts
```diff
@@ -58,62 +59,55 @@ export interface RetryConfig {
...
+        next: (value) => {
+          if (resetOnSuccess) {
+            soFar = 0;
+          }
+          subscriber.next(value);
+        },
+        error: (err) => {
+          if (soFar++ < count) {
+            if (innerSub) {
+              subscription.remove(innerSub);
```
cartant, line 90: As you've pointed out, the `remove` redundant.

src/internal/operators/retry.ts
```diff
@@ -58,62 +59,55 @@ export interface RetryConfig {
...
+              innerSub.unsubscribe();
```
cartant, line 91: I think this will need to be nulled here - for the same reason mentioned in the `repeat` review.

benlesh:
Addressed comments.

Also removed `_unsubscribeAndRecycle`! 🎉 🎈 🍰

benlesh:
🎉

PR #5623, opened 2020-08-03
fix(retryWhen): Ensure subscription tears down between retries

benlesh:
Fixes an issue where subscriptions would not tear down until all retries were exhausted in the synchronous case.

### DIFF
--- a/src/internal/operators/retryWhen.ts
+++ b/src/internal/operators/retryWhen.ts
@@ -1,13 +1,11 @@
-import { Operator } from '../Operator';
+/** @prettier */
 import { Subscriber } from '../Subscriber';
 import { Observable } from '../Observable';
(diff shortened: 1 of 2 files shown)


src/internal/operators/retryWhen.ts
```diff
@@ -62,88 +60,84 @@ import { SimpleOuterSubscriber, innerSubscribe, SimpleInnerSubscriber } from '..
...
-    }
-    this.retries = null;
-  }
-
-  notifyNext(): void {
-    const { _unsubscribe } = this;
+        return errors$;
+      };
+
+      const subscribeNext = () => {
```
cartant, line 110: An operator-specific name for more obvious stack traces?

PR #5625, opened 2020-08-03
fix(repeatWhen): Ensure teardown happens between repeat subscriptions

benlesh:
Resolves an issue where teardowns would wait until the resulting observable ended before being executed.


src/internal/operators/repeatWhen.ts
```diff
@@ -38,98 +38,96 @@ import { SimpleOuterSubscriber, innerSubscribe, SimpleInnerSubscriber } from '..
...
+            },
+          })
+        );
       }
-      if (!this.retriesSubscription || this.retriesSubscription.closed) {
-        return super.complete();
+      return completions$;
+    };
+
+    const subscribeNext = () => {
```
cartant, line 94: An operator-specific name for more obvious stack traces?

PR #5634, opened 2020-08-05
Reworking Multicasting: share, connect, and makeConnectable

benlesh:
I'd like to do this instead of what I was doing over at #5432 ..

## Overview

1. Make `share` configurable, such that it can be used to create things like `shareReplay` or people can configure the behavior however they like. If they want a `shareBehavior` or `shareLast`, they can easily make it work however they like.
2. Add a new`connect` operator that does the "selector" version of `multicast`, `publish`, et al. (this also fixes a broken error behavior in `publishReplay`). 
3. Add a new `makeConnectable` function that creates a `ConnectableObservableLike` (instead of `ConnectableObservable`).

### We'd have
- `share` (NOTE: I'll leave `shareReplay` as just a wrapper around this, as seen in this PR)
- `connect`
- `connectable`

### Instead of
- `share`
- `multicast`
- `publish`
- `publishLast`
- `publishReplay`
- `publishBehavior`
- `refCount`
- `shareReplay`
- `ConnectableObservable`

All "operators" at that point would actually only return `Observable` and never `ConnectableObservable`, nor would any require `ConnectableObservable` as a source (such as with `refCount`).

## TODO:

- [ ] Additional documentation
- [ ] Additional tests
- [ ] Add deprecation details to deprecated operators and paths

## Other thoughts

It's possible that a better API for `makeConnectable` might be to return a tuple with a `connect` function. Like so:

```ts
const [ result$, connect ] = makeConnectable(source);
result$.subscribe(console.log);
connect();

// instead of
const result$ = makeConnectable(source);
result$.subscribe(console.log);
result$.connect();
```

## How this "replaces" certain things

Well, the I find the prevalence of `publishReplay(1), refCount()` _very_ disturbing. For one thing, I doubt many folks using this are aware of what this does completely. It's not retryable, it's not replayable, etc.

```ts
a$.pipe(publishReplay(1), refCount())

// is

a$.pipe(
  share({
    connector: () => new ReplaySubject(1),
    resetOnError: false,
    resetOnComplete: false,
    resetOnUnsubscribe: false
  });
)
```

"But Ben, that's more code!"... yes, and it's explicit code that tells you _exactly_ what you're doing there. That's not retryable, it's not repeatable... it will never reset internally. It also allows the author more control without having to know implementation details of `publishReplay`. Needing to know implementation details of our multicasting operators has been a sore point of this library, IMO.

**NO WORRIES**: `a$.pipe(share())` will work _exactly_ as it does now. (Notice those tests didn't change)

### DIFF
--- a/spec/operators/share-spec.ts
+++ b/spec/operators/share-spec.ts
@@ -1,331 +1,466 @@
 import { expect } from 'chai';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
-import { share, retry, mergeMapTo, mergeMap, tap, repeat, take } from 'rxjs/operators';
-import { Observable, EMPTY, NEVER, of } from 'rxjs';
+import { share, retry, mergeMapTo, mergeMap, tap, repeat, take, takeUntil, takeWhile, materialize } from 'rxjs/operators';
+import { Observable, EMPTY, NEVER, of, Subject, Observer, from } from 'rxjs';
+import { TestScheduler } from 'rxjs/testing';
+import { observableMatcher } from '../helpers/observableMatcher';
+import sinon = require('sinon');
 
 /** @test {share} */
-describe('share operator', () => {
-  it('should mirror a simple source Observable', () => {
-    const source = cold('--1-2---3-4--5-|');
-    const sourceSubs =  '^              !';
-    const expected =    '--1-2---3-4--5-|';
+describe('share', () => {
+  describe('share()', () => {
+    it('should mirror a simple source Observable', () => {
+      const source = cold('--1-2---3-4--5-|');
+      const sourceSubs =  '^              !';
+      const expected =    '--1-2---3-4--5-|';
 
-    const shared = source.pipe(share());
+      const shared = source.pipe(share());
 
-    expectObservable(shared).toBe(expected);
-    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
-  });
-
-  it('should share a single subscription', () => {
-    let subscriptionCount = 0;
-    const obs = new Observable<never>(observer => {
-      subscriptionCount++;
+      expectObservable(shared).toBe(expected);
+      expectSubscriptions(source.subscriptions).toBe(sourceSubs);
     });
 
-    const source = obs.pipe(share());
+    it('should share a single subscription', () => {
+      let subscriptionCount = 0;
+      const obs = new Observable<never>(observer => {
+        subscriptionCount++;
+      });
 
-    expect(subscriptionCount).to.equal(0);
+      const source = obs.pipe(share());
 
-    source.subscribe();
-    source.subscribe();
+      expect(subscriptionCount).to.equal(0);
 
-    expect(subscriptionCount).to.equal(1);
-  });
+      source.subscribe();
+      source.subscribe();
 
-  it('should not change the output of the observable when error', () => {
-    const e1 = hot('---a--^--b--c--d--e--#');
-    const e1subs =       '^              !';
-    const expected =     '---b--c--d--e--#';
-
-    expectObservable(e1.pipe(share())).toBe(expected);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
-  });
-
-  it('should not change the output of the observable when successful with cold observable', () => {
-    const e1 =  cold('---a--b--c--d--e--|');
-    const e1subs =   '^                 !';
-    const expected = '---a--b--c--d--e--|';
-
-    expectObservable(e1.pipe(share())).toBe(expected);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
-  });
-
-  it('should not change the output of the observable when error with cold observable', () => {
-    const e1 =  cold('---a--b--c--d--e--#');
-    const e1subs =   '^                 !';
-    const expected = '---a--b--c--d--e--#';
-
-    expectObservable(e1.pipe(share())).toBe(expected);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
-  });
-
-  it('should retry just fine', () => {
-    const e1 =  cold('---a--b--c--d--e--#');
-    const e1subs =  ['^                 !                  ',
-                   '                  ^                 !'];
-    const expected = '---a--b--c--d--e-----a--b--c--d--e--#';
-
-    expectObservable(e1.pipe(share(), retry(1))).toBe(expected);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
-  });
-
-  it('should share the same values to multiple observers', () => {
-    const source =     cold('-1-2-3----4-|');
-    const sourceSubs =      '^           !';
-    const shared = source.pipe(share());
-    const subscriber1 = hot('a|           ').pipe(mergeMapTo(shared));
-    const expected1   =     '-1-2-3----4-|';
-    const subscriber2 = hot('    b|       ').pipe(mergeMapTo(shared));
-    const expected2   =     '    -3----4-|';
-    const subscriber3 = hot('        c|   ').pipe(mergeMapTo(shared));
-    const expected3   =     '        --4-|';
-
-    expectObservable(subscriber1).toBe(expected1);
-    expectObservable(subscriber2).toBe(expected2);
-    expectObservable(subscriber3).toBe(expected3);
-    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
-  });
-
-  it('should share an error from the source to multiple observers', () => {
-    const source =     cold('-1-2-3----4-#');
-    const sourceSubs =      '^           !';
-    const shared = source.pipe(share());
-    const subscriber1 = hot('a|           ').pipe(mergeMapTo(shared));
-    const expected1   =     '-1-2-3----4-#';
-    const subscriber2 = hot('    b|       ').pipe(mergeMapTo(shared));
-    const expected2   =     '    -3----4-#';
-    const subscriber3 = hot('        c|   ').pipe(mergeMapTo(shared));
-    const expected3   =     '        --4-#';
-
-    expectObservable(subscriber1).toBe(expected1);
-    expectObservable(subscriber2).toBe(expected2);
-    expectObservable(subscriber3).toBe(expected3);
-    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
-  });
-
-  it('should share the same values to multiple observers, ' +
-  'but is unsubscribed explicitly and early', () => {
-    const source =     cold('-1-2-3----4-|');
-    const sourceSubs =      '^        !   ';
-    const shared = source.pipe(share());
-    const unsub =           '         !   ';
-    const subscriber1 = hot('a|           ').pipe(mergeMapTo(shared));
-    const expected1   =     '-1-2-3----   ';
-    const subscriber2 = hot('    b|       ').pipe(mergeMapTo(shared));
-    const expected2   =     '    -3----   ';
-    const subscriber3 = hot('        c|   ').pipe(mergeMapTo(shared));
-    const expected3   =     '        --   ';
-
-    expectObservable(subscriber1, unsub).toBe(expected1);
-    expectObservable(subscriber2, unsub).toBe(expected2);
-    expectObservable(subscriber3, unsub).toBe(expected3);
-    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
-  });
-
-  it('should share an empty source', () => {
-    const source = cold('|');
-    const sourceSubs =  '(^!)';
-    const shared = source.pipe(share());
-    const expected =    '|';
-
-    expectObservable(shared).toBe(expected);
-    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
-  });
-
-  it('should share a never source', () => {
-    const source = cold('-');
-    const sourceSubs =  '^';
-    const shared = source.pipe(share());
-    const expected =    '-';
-
-    expectObservable(shared).toBe(expected);
-    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
-  });
-
-  it('should share a throw source', () => {
-    const source = cold('#');
-    const sourceSubs =  '(^!)';
-    const shared = source.pipe(share());
-    const expected =    '#';
-
-    expectObservable(shared).toBe(expected);
-    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
-  });
-
-  it('should connect when first subscriber subscribes', () => {
-    const source = cold(       '-1-2-3----4-|');
-    const sourceSubs =      '   ^           !';
-    const shared = source.pipe(share());
-    const subscriber1 = hot('   a|           ').pipe(mergeMapTo(shared));
-    const expected1 =       '   -1-2-3----4-|';
-    const subscriber2 = hot('       b|       ').pipe(mergeMapTo(shared));
-    const expected2 =       '       -3----4-|';
-    const subscriber3 = hot('           c|   ').pipe(mergeMapTo(shared));
-    const expected3 =       '           --4-|';
-
-    expectObservable(subscriber1).toBe(expected1);
-    expectObservable(subscriber2).toBe(expected2);
-    expectObservable(subscriber3).toBe(expected3);
-    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
-  });
-
-  it('should disconnect when last subscriber unsubscribes', () => {
-    const source =     cold(   '-1-2-3----4-|');
-    const sourceSubs =      '   ^        !   ';
-    const shared = source.pipe(share());
-    const subscriber1 = hot('   a|           ').pipe(mergeMapTo(shared));
-    const unsub1 =          '          !     ';
-    const expected1   =     '   -1-2-3--     ';
-    const subscriber2 = hot('       b|       ').pipe(mergeMapTo(shared));
-    const unsub2 =          '            !   ';
-    const expected2   =     '       -3----   ';
-
-    expectObservable(subscriber1, unsub1).toBe(expected1);
-    expectObservable(subscriber2, unsub2).toBe(expected2);
-    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
-  });
-
-  it('should not break unsubscription chain when last subscriber unsubscribes', () => {
-    const source =     cold(   '-1-2-3----4-|');
-    const sourceSubs =      '   ^        !   ';
-    const shared = source.pipe(
-      mergeMap((x: string) => of(x)),
-      share(),
-      mergeMap((x: string) => of(x))
-    );
-    const subscriber1 = hot('   a|           ').pipe(mergeMapTo(shared));
-    const unsub1 =          '          !     ';
-    const expected1   =     '   -1-2-3--     ';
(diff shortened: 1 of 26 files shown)


benlesh:
Related: #3833

benlesh:
Rebased

benlesh:
Core team to review... revisit next meeting.

benlesh:
This is blocked on #5729

src/internal/operators/connect.ts
```diff
@@ -0,0 +1,44 @@
...
+        result = from(setup(subject.asObservable()));
+      } catch (err) {
+        subscriber.error(err);
+        return;
+      }
+
+      const subscription = result.subscribe(subscriber);
+      subscription.add(source.subscribe(subject));
+      return subscription;
+    });
```
cartant, line 42: nitpick: This PR predates the smallification, so it should be updated to use `operate`. And the `source.subscribe(subject)` subscription should be added to the `subscriber`, etc.

src/internal/operators/connect.ts
```diff
@@ -0,0 +1,44 @@
...
+
+function defaultConnector<T>() {
+  return new Subject<T>();
+}
+export function connect<T, R>({
+  connector = defaultConnector,
+  setup,
+}: {
+  connector?: () => Subject<T>;
+  setup: (shared: Observable<T>) => ObservableInput<R>;
```
cartant, line 18: nitpick: I'm not a fan of the `setup` property's name. TBH, I preferred `selector`. My main beef with `setup` is that, IMO, it doesn't really explain what the parameter does. [Elsewhere in the API](https://github.com/ReactiveX/rxjs/blob/8ef4fee2b5f7fb68426661ffcc47b50ab81b145e/src/internal/operators/catchError.ts#L12) functions that are used in this manner are called selectors and I am a fan of consistency and precedent.

src/internal/observable/makeConnectable.ts
```diff
@@ -0,0 +1,43 @@
...
+export function makeConnectable<T>(source: ObservableInput<T>, connector: Subject<T> = new Subject<T>()): ConnectableObservableLike<T> {
+  let connection: Subscription | null = null;
+
+  const result: any = new Observable<T>((subscriber) => {
+    return connector.subscribe(subscriber);
+  });
+
+  result.connect = function () {
+    if (!connection) {
+      connection = defer(() => source).subscribe(connector);
```
cartant, line 37: question: Is there a reason for using `defer` here instead of `from`? `source` is a parameter and is never reassigned, so I don't see why this needs to be lazy.

src/internal/observable/makeConnectable.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Creates an observable that multicasts once `connect()` is called on it.
+ *
+ * @param source The observable source to make connectable.
+ * @param connector The subject to used to multicast the source observable to all subscribers.
+ * Defaults to a new {@link Subject}.
+ * @returns A "connectable" observable, that has a `connect()` method, that you must call to
+ * connect the source to all consumers through the subject provided as the connector.
+ */
+export function makeConnectable<T>(source: ObservableInput<T>, connector: Subject<T> = new Subject<T>()): ConnectableObservableLike<T> {
```
cartant, line 28: nitpick: Not a fan of the `makeConnectable` name. IMO, this seems odd, when looking at the other parts of the API. I mean, it's `timer`, not `makeTimer` or `createTimer`. My preference would be for it to be named `connectable`.

src/internal/operators/share.ts
```diff
@@ -1,14 +1,51 @@
...
-import { MonoTypeOperatorFunction } from '../types';
+import { MonoTypeOperatorFunction, OperatorFunction } from '../types';
+import { Subscription } from '../Subscription';
+import { from } from '../observable/from';
+import { lift } from '../util/lift';
+import { Subscriber } from '../Subscriber';
 
-function shareSubjectFactory() {
-  return new Subject<any>();
+interface ShareOptions<T, R> {
```
cartant, line 11: (very minor) nitpick: In [other parts of the API](https://github.com/ReactiveX/rxjs/blob/8ef4fee2b5f7fb68426661ffcc47b50ab81b145e/src/internal/operators/throttle.ts#L9), these types/parameters are referred to as 'config'. I think it should be one or the other and should be consistent throughout the API, but - again - this is very minor.

src/internal/operators/share.ts
```diff
@@ -50,13 +87,65 @@ function shareSubjectFactory() {
...
+    let refCount = 0;
+    let hasCompleted = false;
+    let hasErrored = false;
+
+    const reset = () => {
+      connection = subject = null;
+      hasCompleted = hasErrored = false;
+    };
+
+    return lift(source, function (this: Subscriber<T | R>, source: Observable<T>) {
```
cartant, line 109: nitpick: With the changes made in the smallification, this should now used `operate`.

src/internal/operators/share.ts
```diff
@@ -50,13 +87,65 @@ function shareSubjectFactory() {
...
+      const subscriber = this;
+      refCount++;
+      if (!subject) {
+        subject = connector!();
+      }
+
+      const castSubscription = subject.subscribe(subscriber);
```
cartant, line 116: nitpick: This seems like a weird variable name, IMO. Where is the cast?

cartant (approved):
LGTM. Just some questions, a bunch of nitpicks and some names of which I am not a fan. 😅

Also, after this is rebased, the skipped firehose tests should be unskipped and made to pass. A contributor opened a PR to address the skipped tests, but I blocked the PR, as the changes were based on the old `multicast` infrastructure - see https://github.com/ReactiveX/rxjs/pull/5834

src/internal/operators/connect.ts
backbone87: how about `project`?

src/internal/operators/share.ts
```diff
@@ -50,13 +87,65 @@ function shareSubjectFactory() {
...
  */
-export function share<T>(): MonoTypeOperatorFunction<T> {
-  return (source: Observable<T>) => refCount()(multicast(shareSubjectFactory)(source)) as Observable<T>;
+export function share<T, R>(options?: ShareOptions<T, R>): OperatorFunction<T, T | R> {
+  return (source: Observable<T>) => {
+    options = options || {};
+    let { connector, resetOnComplete = true, resetOnError = true, resetOnUnsubscribe = true } = options;
+    if (!connector) {
+      connector = () => new Subject<T>();
+    }
```
backbone87, line 97: isnt this more consistent?
```
const {
  connector = () => new Subject<T>(),
  resetOnComplete = true,
  resetOnError = true,
  resetOnUnsubscribe = true,
} = options || {};
```

also shouldnt this be moved into the `share` functions block insteadof the operator function block? maybe even:
```
export function share<T, R>({
  connector = () => new Subject<T>(),
  resetOnComplete = true,
  resetOnError = true,
  resetOnUnsubscribe = true,
}: ShareOptions<T, R> = {}): OperatorFunction<T, T | R> {
  // ...
}
```

src/internal/operators/share.ts
```diff
@@ -1,14 +1,51 @@
...
+  resetOnComplete?: boolean;
+  /**
+   * If true, when the number of subscribers to the resulting observable reaches zero due to those subscribers unsubscribing, the
+   * internal state will be reset and the resulting observable will return to a "cold" state. This means that the next
+   * time the resulting observable is subscribed to, a new subject will be created and the source will be subscribed to
+   * again.
+   * If false, when the number of subscribers to the resulting observable reaches zero due to unsubscription, the subject
+   * will remain connected to the source, and new subscriptions to the result will be connected through that same subject.
+   */
+  resetOnUnsubscribe?: boolean;
```
backbone87, line 42: i was under the assumption that a complete/error will always result in an implicit `unsubscribe` afterwards. its not reflected in this description that an unsubscribe that follows a complete/error, will **not** result in a reset.

backbone87: > This means that the next time the resulting observable is subscribed to, a new subject will be created and the source will be subscribed to again.

This should also be added to the `resetOnError`/`resetOnComplete` descriptions. Maybe its worth to describe a `reset` independantly of these 3 configs in the share operator itself, so that the description of the reset params can be reduced to actual differences between them and therefore causing less mental friction.

src/internal/operators/share.ts
backbone87, line 11: is it worth to add a `resetNotifier$` in order to have more fine grained control over when resets happen?

backbone87:
> It's possible that a better API for makeConnectable might be to return a tuple with a connect function. Like so:

just throwing in other alternatives:
```
const notifier = new Subject();

makeConnectable(source, notifier);
notifier.subscribe(connects(source));
```

src/internal/operators/connect.ts
benlesh: `selector`, IMO, also doesn't explain what the parameter does, either. I mean, this sets up the multicast as much as it selects the output. Happy to bikeshed the name, but overall, I guess I'm just not sure what to name this. `project` is used elsewhere to mean a synchronous map-type function.

benlesh: Maybe `cast`? or `multicast`? or??

benlesh: (I'm open to moving to back to the `selector`, though). I just don't like that it doesn't completely imply what's going on here. `catchError` is a little different, because it's only called when an error happens, and you have to "select" what to observe after that error shuts things down.

This function is required to setup the multicast. It also selects what sort of values are emitted from the result. I guess I'm now unhappy with either `setup` or `select` haha. But I do know I want it to be terse.

benlesh:
# NOTE: Lots of work being done here, still a WIP.

benlesh:
Okay, I've made a few changes to this.

Per @cartant's request, `makeConnectable` is now `connectable`.

### OTHER VERY IMPORTANT THINGS:

1. I've added a new TestScheduler helper:  `expectObservable(a$).toEqual(b$)`. I needed this so I could test that the deprecation advice I was giving people in the deprecation messages was proven, not just theoretical. It's publicly exposed, and I think we should ship it. It's useful for this exact scenario and it follows the `jest` and `jasmine` pattern of `toBe` and `toEqual`, IMO.
2. I've fixed the `Observer` interface to remove the optional `closed` property, which shouldn't have been there. (That may be breaking for TypeScript users).
3. I've fixed the `Subscribable` interface to just be what is valid as a return value of `[Symbol.observable]()`. That's because I was using with `SubjectLike`.
4. I'm now exporting a new type `SubjectLike`, it's exactly what it sounds like.

### TODO

1. We need to discuss `SubjectLike`. We might want it to be `SubjectLike<In, Out>` and not just `SubjectLike<T>`. This relates to things like `WebSocketSubject`, etc.
2. We need to bikeshed the `setup` function as discussed above.

cc @cartant @kwonoj

src/internal/operators/share.ts
benlesh: You could technically do that by composing `takeUntil` 

```ts
source$.pipe(
  takeUntil(resetNotifier$),
  share(),
  repeat(),
)
```

src/internal/operators/connect.ts
cartant: > This function is required to setup the multicast. It also selects what sort of values are emitted from the result.

Well, the function is "required to setup" the multicast, but it doesn't actually set it up. It receives the shared/multicast observable and then "selects what sort of values are emitted from the result".

That said, `cast` or `multicast` are okay, I suppose, but my preference is still for `selector`.

src/internal/operators/share.ts
```diff
@@ -1,14 +1,50 @@
...
 
-import { MonoTypeOperatorFunction } from '../types';
+import { MonoTypeOperatorFunction, OperatorFunction, SubjectLike } from '../types';
+import { Subscription } from '../Subscription';
+import { from } from '../observable/from';
+import { operate } from '../util/lift';
 
-function shareSubjectFactory() {
-  return new Subject<any>();
+interface ShareOptions<T, R> {
```
cartant, line 10: issue: My understanding is that typing this using `interface` and not exporting it will be a problem for consumers of the lib. Types declared using `type` are anonymous, but types declared with `interface` are not. See [this playground snippet](selects what sort of values are emitted from the result) - from [this blog post](https://ncjamieson.com/prefer-interfaces/) - that simulates the situation using an IIFE. However, I don't understand why this isn't effecting problems with the tests. 🤷‍♂️ Still a concern, though.

spec/schedulers/TestScheduler-spec.ts
```diff
@@ -393,8 +393,9 @@ describe('TestScheduler', () => {
         const expected = ' -abcdef|';
 
         expectObservable(output).toBe(expected);
-        expectSubscriptions(obs1.subscriptions).toBe('^-----!');
-        expectSubscriptions(obs2.subscriptions).toBe('^------!');
+        expectObservable(output).toEqual(cold(expected));
+        expectSubscriptions(obs1.subscriptions).toBe(['^-----!', '^-----!']);
+        expectSubscriptions(obs2.subscriptions).toBe(['^------!', '^------!']);
```
cartant, line 398: issue: It's not at all clear - to me, at least - why `obs1` and `obs2` are subscribed to twice. I think these expectations warrant an explanation. And if this is somehow related to `toEqual`, I don't think that helper should be made public.

src/internal/observable/connectable.ts
```diff
@@ -0,0 +1,47 @@
+/** @prettier */
+
+import { ObservableInput } from '../types';
+import { Subject } from '../Subject';
+import { Subscription } from '../Subscription';
+import { Observable } from '../Observable';
+import { defer } from './defer';
+
+export type ConnectableObservableLike<T> = Observable<T> & {
```
cartant, line 9: nitpick: this could be an interface as per [TS team recommendation](https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections):

```ts
export interface ConnectableObservableLike<T> extends Observable<T> {
  connect(): Subscription;
}
```

src/internal/operators/multicast.ts
```diff
@@ -3,54 +3,83 @@ import { Subject } from '../Subject';
...
- * as many times as needed, without causing multiple subscriptions to the source stream.
- * Subscribers to the given source will receive all notifications of the source from the
- * time of the subscription forward.
- * @return {Observable} An Observable that emits the results of invoking the selector
- * on the items emitted by a `ConnectableObservable` that shares a single subscription to
- * the underlying stream.
+ * @param subjectFactory A factory that creates the subject used to multicast.
+ * @param selector A setup function to setup the multicast
+ * @deprecated To be removed in version 8. Please use the new {@link connect} operator.
+ * `multicast(fn1, fn2)` is equivalent to `connect({ connector: fn1, setup: fn2 })`.
```
cartant, line 63: nitpick/reminder: post-bikeshedding, the `setup` prop in this comment will likely need renaming. It shows up in a few other places in docs comments, too.

spec/operators/share-spec.ts
```diff
@@ -1,331 +1,466 @@
...
-  it.skip('should stop listening to a synchronous observable when unsubscribed', () => {
-    const sideEffects: number[] = [];
-    const synchronousObservable = new Observable<number>(subscriber => {
-      // This will check to see if the subscriber was closed on each loop
-      // when the unsubscribe hits (from the `take`), it should be closed
-      for (let i = 0; !subscriber.closed && i < 10; i++) {
-        sideEffects.push(i);
-        subscriber.next(i);
-      }
+  describe('share(config)', () => {
```
cartant, line 337: nitpick: In a previous review I mentioned that the parameter type is named `ShareOptions` whereas elsewhere type names like `XxxConfig` is used. Can we rename it to config? 🙏 I mean, it's referred to here as  config - not options. I am a fan of consistency, however trivial.

cartant (commented):
Some things that might or might not be issues and some nitpicks and comments/thoughts, etc. No major issues, AFAICT.

spec/schedulers/TestScheduler-spec.ts
benlesh: They're merged. And the merged observable is subscribed to on line 395 and again on line 396.

```ts
expectObservable(output).toBe(expected);
expectObservable(output).toEqual(cold(expected));
```

benlesh: I can add a comment I guess.

src/internal/operators/multicast.ts
benlesh: Discussed offline: Moving to `connect(fn, config)`.

src/internal/types.ts
```diff
@@ -87,14 +87,7 @@ export type SubscribableOrPromise<T> = Subscribable<T> | Subscribable<never> | P
...
 export interface Subscribable<T> {
-  subscribe(observer?: PartialObserver<T>): Unsubscribable;
-  /** @deprecated Use an observer instead of a complete callback */
-  subscribe(next: null | undefined, error: null | undefined, complete: () => void): Unsubscribable;
-  /** @deprecated Use an observer instead of an error callback */
-  subscribe(next: null | undefined, error: (error: any) => void, complete?: () => void): Unsubscribable;
-  /** @deprecated Use an observer instead of a complete callback */
-  subscribe(next: (value: T) => void, error: null | undefined, complete: () => void): Unsubscribable;
-  subscribe(next?: (value: T) => void, error?: (error: any) => void, complete?: () => void): Unsubscribable;
+  subscribe(observer: Observer<T>): Unsubscribable;
```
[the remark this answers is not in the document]
benlesh: Well, I guess that makes subscribables harder to develop, as the onus will be on their implementations to figure out what handlers were called. But I supposed that's not too bad now with `?.`. I'll make it `Partial`.

benlesh:
Requesting rereview.

Note that the change to make `Subscribable` accept `Partial<Observer<T>>` also meant that the type for `Observable` `subscribe` needed to take `Partial<Observer<T>>` instead of `PartialObserver<T>`, the latter being a type that forced at least one callback (for better or worse). I'm totally fine with this, and I'm not sure that bit was helpful anyhow. If anything, I'd bet it was confusing for users at times.

cartant (approved):
LGTM ❤

src/internal/operators/publishLast.ts
```diff
@@ -51,14 +50,17 @@ import { UnaryFunction } from '../types';
...
- *
- * @return {ConnectableObservable} An observable sequence that contains the elements of a
+ * @return A connectable observable sequence that contains the elements of a
  * sequence produced by multicasting the source sequence.
+ * @deprecated To be removed in version 8. If you're trying to create a connectable observable
+ * with an {@link AsyncSubject} under the hood, please use the new {@link connectable} creation function.
+ * `source.pipe(publishLast())` is equivalent to `connectable(source, () => new AsyncSubject())`.
+ * If you're using {@link refCount} on the result of `publishLast`, you can use the updated {@link share}
+ * operator, which is now highly configurable. `source.pipe(publishLast(), refCount())`
+ * is equivalent to `source.pipe(share({ connector: () => new AsyncSubject(), resetOnError: false, resetOnComplete: false, resetOnRefCountZero: false }))`.
```
backbone87, line 60: maybe i missunderstood something, but shouldnt be `resetOnRefCountZero: true` here?

benlesh: :) `publishLast` keeps the same subject instance under the hood even after the refCount recycles.

cartant:
🎉

PR #5643, opened 2020-08-12
chore: bump TypeScript to 4.0.1-rc

cartant:
**Description:**

This PR bumps TypeScript to `4.0.1-rc`. Almost no changes were needed:

* `npm run lint` effects no failures.
* `npm run test` sees all tests pass with no compile errors.
* `npm run build:package` builds with no compile errors.
* `npm run dtslint` effected a single failure that was due to named tuple elements being a thing in TS4. (It seems the file with this change had CRLF line endings. They're now LF, hence the large diff. The change is on line 71.)
* And I guess we'll soon find out what happens in CI ... It seems CI is happy 🎉

IMO, we might as well bump this now. Given that it will allow us to remove so many overloads, I think we should be choosing TS4 as the minimum supported version for v7.

Note that this change will affect the Google merge, as [they're using TypeScript 3.8.2](https://github.com/ReactiveX/rxjs/pull/5609#issuecomment-670680772). 😬

### DIFF
--- a/spec-dtslint/types-spec.ts
+++ b/spec-dtslint/types-spec.ts
@@ -68,7 +68,7 @@ describe('Cons', () => {
 
   it('should support rest tuples', () => {
     let explicit: Cons<A, B[]>;
-    let inferred = explicit!; // $ExpectType [A, ...B[]]
+    let inferred = explicit!; // $ExpectType [arg: A, ...rest: B[]]
   });
 });
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "rxjs",
-  "version": "7.0.0-beta.3",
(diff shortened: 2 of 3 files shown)


spec-dtslint/types-spec.ts
```diff
@@ -1,87 +1,87 @@
...
+
+describe('Cons', () => {
+  it('should construct a tuple with the specified type at the head', () => {
+    let explicit: Cons<A, [B, C]>;
+    let inferred = explicit!; // $ExpectType [A, B, C]
+  });
+
+  it('should support rest tuples', () => {
+    let explicit: Cons<A, B[]>;
+    let inferred = explicit!; // $ExpectType [arg: A, ...rest: B[]]
```
benlesh, line 71: 🤯

PR #5646, opened 2020-08-15
perf(SafeSubscriber): avoid using `Object.create`

JoostK:
In profiling sessions, the `SafeSubscriber` constructor was showing as hot
function for code that creates and tear downs Observables very frequently.
The usage of `Object.create` attributes to the slowness; avoiding using it
completely removes any overhead of `SafeSubscriber`.

The `Object.create` result was used to call anonymous subscribers in their
own context, i.e. the `this` pointer in `next`/`error`/`complete` is the
anonymous object itself. This commit implements an alternative to achieve
the same effect; binding the subscriber functions to the observer object.

# Timings
The below timings are with NodeJS 14.8.0 in a full Angular CLI build of https://github.com/JoostK/ng9-aot-build-times/tree/extra.

**Before**:
Before

**After**:
After

### DIFF
--- a/src/internal/Subscriber.ts
+++ b/src/internal/Subscriber.ts
@@ -162,15 +162,22 @@ export class SafeSubscriber<T> extends Subscriber<T> {
     let next: ((value: T) => void) | undefined;
 
     if (isFunction(observerOrNext)) {
-      next = (<((value: T) => void)> observerOrNext);
+      next = observerOrNext;
     } else if (observerOrNext) {
-      next = (<PartialObserver<T>> observerOrNext).next;
-      error = (<PartialObserver<T>> observerOrNext).error;
-      complete = (<PartialObserver<T>> observerOrNext).complete;
+      next = observerOrNext.next;
+      error = observerOrNext.error;
+      complete = observerOrNext.complete;
       if (observerOrNext !== emptyObserver) {
-        next = next && next.bind(observerOrNext);
-        error = error && error.bind(observerOrNext);
-        complete = complete && complete.bind(observerOrNext);
+        let context: any;
+        if (config.useDeprecatedNextContext) {
+          context = Object.create(observerOrNext);
+          context.unsubscribe = this.unsubscribe.bind(this);
+        } else {
+          context = observerOrNext;
+        }
+        next = next && next.bind(context);
+        error = error && error.bind(context);
+        complete = complete && complete.bind(context);
         if (isSubscription(observerOrNext)) {
           observerOrNext.add(this.unsubscribe.bind(this));
         }
@@ -244,7 +251,7 @@ export class SafeSubscriber<T> extends Subscriber<T> {
 
   private __tryOrUnsub(fn: Function, value?: any): void {
     try {
-      fn.call(this, value);
+      fn(value);
     } catch (err) {
       this.unsubscribe();
       if (config.useDeprecatedSynchronousErrorHandling) {
@@ -260,7 +267,7 @@ export class SafeSubscriber<T> extends Subscriber<T> {
       throw new Error('bad call');
     }
     try {
-      fn.call(this, value);
+      fn(value);
     } catch (err) {
       if (config.useDeprecatedSynchronousErrorHandling) {
         parent.syncErrorValue = err;
--- a/src/internal/Subscriber.ts
+++ b/src/internal/Subscriber.ts
@@ -153,8 +153,6 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
  */
 export class SafeSubscriber<T> extends Subscriber<T> {
 
-  private _context: any;
-
   constructor(private _parentSubscriber: Subscriber<T>,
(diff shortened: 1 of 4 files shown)


src/internal/Subscriber.ts
```diff
@@ -171,15 +168,15 @@ export class SafeSubscriber<T> extends Subscriber<T> {
...
       complete = (<PartialObserver<T>> observerOrNext).complete;
       if (observerOrNext !== emptyObserver) {
-        context = Object.create(observerOrNext);
+        next = next && next.bind(observerOrNext);
+        error = error && error.bind(observerOrNext);
+        complete = complete && complete.bind(observerOrNext);
         if (isSubscription(observerOrNext)) {
           observerOrNext.add(this.unsubscribe.bind(this));
         }
-        context.unsubscribe = this.unsubscribe.bind(this);
```
JoostK, line 178: I was able to remove this statement without affecting any tests; this makes me a bit unsure on what this was trying to achieve and if it's safe to remove it, although I can't think of what this was supposed to do.

JoostK: And I realized after opening this that without this assignment, it becomes possible to just use `context = observerOrNext` as an alternative to binding the subscriber callbacks. If that is indeed possible (again, no tests start to fail) then I think switching to use that approach makes sense.

cartant:
> I was able to remove this statement without affecting any tests; this makes me a bit unsure on what this was trying to achieve and if it's safe to remove it, although I can't think of what this was supposed to do.

It adds `unsubscribe` to the context so that it's possible to effect unsubscription from a synchronous source within a handler. See [this comment](https://github.com/ReactiveX/rxjs/pull/3230#issuecomment-361113247) and the links therein. Notwithstanding the lack of tests/docs, this would be a breaking change.

When the implementation moves to an [`AbortSignal`-based approach](https://github.com/ReactiveX/rxjs/pull/5591), this context shenanigans will no longer be necessary.

JoostK:
Thanks for the additional context @cartant, that gives me exactly the information I was missing before. The `AbortSignal` does indeed look like a viable alternative that should allow for avoiding this performance penalty sometime in the future.

I opened a PR in the Angular CLI to switch the anonymous observer to use separate callbacks instead, also avoiding this penalty: angular/angular-cli#18549. I wouldn't normally recommend using that syntax, but this occurs in a hot code path where the performance penalty starts to make a real difference.

benlesh:
😩 I really want to get rid of this. But... one thing at a time.

benlesh:
@cartant Do you know if we have any tests or documentation that covers the edge case outlined in the comment you linked to? I suspect we do not. Also, the promise of all of this sort of fails, as outlined by those firehose tests that I had you add last week.

benlesh:
What I'm looking for is a way to make this change non-breaking.. OR a way to introduce the change as breaking with some sort of backwards compatibility package. It seems like the later would have to be it, honestly. I'm going to guess that this change would break fewer than 0.001% of projects using RxJS, as I'm all but sure that this is undocumented. (And it's also relatively impossible to type with TypeScript).

benlesh:
At this point, I'm sort of in favor of just making this breaking change. If people gripe during the beta, or even after, we can reenable it behind a flag in config.ts or something. I suspect it will not cause many, if any, issues.

cartant:
> Do you know if we have any tests or documentation that covers the edge case outlined in the comment you linked to?

Nope. Only the comment of Paul's. I recall this coming up a couple of times in issues. Paul's comment is the simplest to find in a search and I've referred to it a few times. This was all well before my time, so I'm unaware of the history re: no tests/docs.

> At this point, I'm sort of in favor of just making this breaking change.

Adding a `useDeprecatedNextContext` to the [`config`](https://github.com/ReactiveX/rxjs/blob/b03c70ce5a2ebc2d324123ce9ecc26922ebaa6b0/src/internal/config.ts) - and defaulting it to `false` - would be reasonable, IMO.

benlesh:
@cartant  I've added the flag, as well as tests, and a little documentation about the breaking change. I've also rebased as there was a conflict in the spec file (since we just recently added more tests, and I also added more tests, no actual tests changed).

src/internal/Subscriber.ts
```diff
@@ -230,7 +234,7 @@ export class SafeSubscriber<T> extends Subscriber<T> {
     if (!this.isStopped) {
       const { _parentSubscriber } = this;
       if (this._complete) {
-        const wrappedComplete = () => this._complete.call(this._context);
+        const wrappedComplete = () => this._complete.call(this);
```
cartant, line 237: suggestion: This `wrappedComplete` business is not necessary. The `_complete` will have already been bound to the callers context - if an object was passed. `_complete` can be called directly.

cartant (approved):
@benlesh LGTM, but `_complete` is unnecessarily wrapped.

src/internal/Subscriber.ts
benlesh: I am about to make a pass through all of this. I'm going to merge this as-is, and address the rest in another PR.

PR #5647, opened 2020-08-16
feat(throwError): now accepts a factory to create the error

### LINKED ISSUE #5617: Feature: allow `throwError` to accept a factory function
The thought is this: Frequently, you don't want to create an `Error` object until the moment the error happens. However, currently, `throwError` only accepts the instance of the error it is going to throw as an argument. 

I'm wondering if we could allow it to optionally take a factory function that will create the error at the moment of subscription.

This would be a breaking change, however, as right now, it's possible to emit a function as an error by using `throwError`:

### Current behavior when you provide a function
```ts
throwError(new Error('bad'  + Date.now())).subscribe({
  error: console.error // logs the error above, but the call stack is from the moment the `throwError` was called, not from during the subscription.
})

throwError(() => new Error('bad' + Date.now())).subscribe({
  error: console.error // logs a function reference
})
```

### Proposed change
```ts
throwError(new Error('bad' + Date.now()); // This is unchanged

// NEW:
throwError(() => new Error('bad' + Date.now())).subscribe({
  error: console.error // logs the error created above, with a call stack from that place
})
```

benlesh:
- Adds feature to allow the error instance the consumer is notified with to be created at the moment of notification.
- Updates tests to use "run mode".
- Improves documentation, including recommendations on when `throwError` is not necessary.

BREAKING CHANGE: In an extreme corner case for usage, `throwError` is no longer able to emit a function as an error directly. If you need to push a function as an error, you will have to use the factory function to return the function like so: `throwError(() => functionToEmit)`, in other words `throwError(() => () => console.log('called later'))`.

resolves #5617

### DIFF
--- a/src/internal/observable/throwError.ts
+++ b/src/internal/observable/throwError.ts
@@ -48,7 +48,7 @@ import { Subscriber } from '../Subscriber';
  *
  * Using `throwError` inside of an operator or creation function
  * with a callback, is usually not necessary:
- * 
+ *
  * ```ts
  * import { throwError, timer, of } from 'rxjs';
  * import { concatMap } from 'rxjs/operators';
@@ -70,9 +70,9 @@ import { Subscriber } from '../Subscriber';
  *    error: console.error
  * });
  * ```
- * 
+ *
  * You can just throw the error instead:
- * 
+ *
  * ```ts
  * import { throwError, timer, of } from 'rxjs';
  * import { concatMap } from 'rxjs/operators';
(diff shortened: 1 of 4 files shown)


src/internal/observable/throwError.ts
```diff
@@ -66,11 +183,17 @@ import { Subscriber } from '../Subscriber';
  * @name throwError
  * @owner Observable
  */
```
cartant, line 185: nitpick: Do we really need to have a TSDoc comment on the implementation? It's not directly callable; only the overload signatures are callable. If there's anything in here that's not in the non-deprecated overload signature, can it be moved there?

PR #5650, opened 2020-08-17
fix(onErrorResumeNext): observables always finalized before moving to next source

benlesh:
- Fixes an issue where finalize would be called after all sources were complete
- Removes superfluous onErrorResumeNext implementation
- Ensures that invalid sources handed to onErrorResumeNext do not cause an error, and instead resume to the next source
- Updates types to infer N arguments
- Removes invalid type signature that allowed a function for no reason
- Adds dtslint tests for static creation function
- Adds finalize test to ensure order of finalize calls (as a sanity check)

### DIFF
--- a/src/internal/observable/onErrorResumeNext.ts
+++ b/src/internal/observable/onErrorResumeNext.ts
@@ -1,18 +1,15 @@
+/** @prettier */
 import { Observable } from '../Observable';
-import { ObservableInput } from '../types';
-import { from } from './from';
-import { isArray } from '../util/isArray';
+import { ObservableInput, ObservedValueUnionFromArray, ObservedValueOf } from '../types';
 import { EMPTY } from './empty';
+import { onErrorResumeNext as onErrorResumeNextWith } from '../operators/onErrorResumeNext';
+import { isArray } from 'util';
 
 /* tslint:disable:max-line-length */
-export function onErrorResumeNext<R>(v: ObservableInput<R>): Observable<R>;
-export function onErrorResumeNext<T2, T3, R>(v2: ObservableInput<T2>, v3: ObservableInput<T3>): Observable<R>;
-export function onErrorResumeNext<T2, T3, T4, R>(v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>): Observable<R>;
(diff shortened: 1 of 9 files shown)


src/internal/observable/onErrorResumeNext.ts
```diff
@@ -1,18 +1,15 @@
+/** @prettier */
 import { Observable } from '../Observable';
-import { ObservableInput } from '../types';
-import { from } from './from';
-import { isArray } from '../util/isArray';
+import { ObservableInput, ObservedValueUnionFromArray, ObservedValueOf } from '../types';
 import { EMPTY } from './empty';
+import { onErrorResumeNext as onErrorResumeNextWith } from '../operators/onErrorResumeNext';
+import { isArray } from 'util';
```
cartant, line 6: issue: This looks like a VS Code auto-import-effected problem. AFAICT, we don't have a `util` package dependency.

src/internal/operators/onErrorResumeNext.ts
```diff
@@ -87,81 +80,53 @@ export function onErrorResumeNext<T, R>(array: ObservableInput<any>[]): Operator
...
-export function onErrorResumeNextStatic<R>(...observables: Array<ObservableInput<any> | ((...values: Array<any>) => R)>): Observable<R>;
-export function onErrorResumeNextStatic<R>(array: ObservableInput<any>[]): Observable<R>;
-/* tslint:enable:max-line-length */
-
-export function onErrorResumeNextStatic<T, R>(...nextSources: Array<ObservableInput<any> |
-  Array<ObservableInput<any>> |
-  ((...values: Array<any>) => R)>): Observable<R> {
-  let source: ObservableInput<any> | null = null;
-
+export function onErrorResumeNext<T, R>(...nextSources: ObservableInput<any>[]): OperatorFunction<T, R> {
```
cartant, line 83: suggestion: This doesn't relate to the changes made in this PR - it was pre-existing here and occurs elsewhere in the codebase, too - but I think the `R` type parameter should be removed and the return type should be `OperationFunction<T, unknown>`, 'cause that's what it is, here. The actual inferences are made in the overload signatures; the implementation has no idea what the output type should be. Type parameters that are not used serve no purpose and are inferred to be `unknown` anyway.

I think the uses of `any` within the implementation should also be replaced with `unknown`.

With those changes, there are not lint failures or compilation issues that I can see.

benlesh:
Rebased to resolve conflict.

cartant (approved):
LGTM. Seems to need an API guardian update, though.

PR #5654, opened 2020-08-18
fix(buffer): Ensure notifier is subscribed after source

### LINKED ISSUE #1754: Bug in buffer or bufferCount operation when used with subject

benlesh:
- Resolves an issue where a multicast observable could not adequately be used to notify a buffer on itself
- Corrects a regression that was introduced by a bad merge back in 6.0. This was originally corrected in PR #2195.

fixes #1754

### DIFF
--- a/spec/operators/buffer-spec.ts
+++ b/spec/operators/buffer-spec.ts
@@ -1,7 +1,8 @@
 import { buffer, mergeMap, take } from 'rxjs/operators';
-import { EMPTY, NEVER, throwError, of } from 'rxjs';
+import { EMPTY, NEVER, throwError, of, Subject } from 'rxjs';
 import { TestScheduler } from 'rxjs/testing';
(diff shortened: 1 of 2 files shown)


niklas-wortmann (approved):
LGTM

PR #5655, opened 2020-08-18
fix(catchError): inner synchronous observables will properly terminate

benlesh:
Related to #5652

### DIFF
--- a/src/internal/operators/catchError.ts
+++ b/src/internal/operators/catchError.ts
@@ -125,9 +125,8 @@ export function catchError<T, O extends ObservableInput<any>>(
         }
       };
 
-      innerSub = source.subscribe({
-        next: (value) => subscriber.next(value),
-        error: (err) => {
+      innerSub = source.subscribe(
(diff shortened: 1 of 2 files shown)


src/internal/operators/catchError.ts
```diff
@@ -153,3 +151,21 @@ export function catchError<T, O extends ObservableInput<any>>(
...
     });
 }
+
+/**
+ * This must exist to ensure that the `closed` state of the inner subscriber is set at
+ * the proper time to ensure operators like `take` can stop the inner subscription if
+ * it is a synchronous firehose.
+ */
+class CatchErrorSubscriber<T, C> extends Subscriber<T> {
+  ready = false;
```
cartant, line 161: issue: I cannot see where this property is used. Do we really need it?

cartant (approved):
LGTM, the only issue seems to be an extraneous property.

src/internal/operators/catchError.ts
benlesh: oops. That's cruft.

PR #5656, opened 2020-08-18
feat(Subscription): `add` no longer returns unnecessary Subscription …

benlesh:
…reference

- Updates a few locations where this feature was being used incorrectly.
- Eliminates confusion where people were trying to "chain" `add` calls.

BREAKING CHANGE: `add` no longer returns an unnecessary Subscription reference. This means that if you are calling `add` with a _function_ and not a `Subscription` (e.g. `subscription.add(() => { /* teardown */ })`), you will not be able to remove that teardown _function_ with `remove`. The fix for this is to wrap your function with a `Subscription` like so:  `const childSub = new Subscription(() => { /* teardown */ }); subscription.add(childSub);`. Then you will be able to remove it via `subscription.remove(childSub);`. Bear in mind that is it an edge case to need to manually remove a child subscription from a parent subscription. **All subscriptions that have been added to other subscriptions will remove themselves from the parent subscription(s) automatically when they are unsubscribed.**.

### DIFF
--- a/src/internal/Subscription.ts
+++ b/src/internal/Subscription.ts
@@ -140,11 +140,11 @@ export class Subscription implements SubscriptionLike {
    * `remove()` to remove the passed teardown logic from the inner subscriptions
    * list.
    */
-  add(teardown: TeardownLogic): Subscription {
+  add(teardown: TeardownLogic): void {
     let subscription = (<Subscription>teardown);
 
     if (!teardown) {
-      return Subscription.EMPTY;
+      return;
     }
 
     switch (typeof teardown) {
@@ -153,10 +153,9 @@ export class Subscription implements SubscriptionLike {
       case 'object':
         if (subscription === this || subscription.closed || typeof subscription.unsubscribe !== 'function') {
           // This also covers the case where `subscription` is `Subscription.EMPTY`, which is always in `closed` state.
-          return subscription;
+          return;
         } else if (this.closed) {
           subscription.unsubscribe();
-          return subscription;
         } else if (!(subscription instanceof Subscription)) {
           const tmp = subscription;
           subscription = new Subscription();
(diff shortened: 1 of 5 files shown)


benlesh:
This is an important step toward fixing some critical internal pieces of our library.

src/internal/Subscription.ts
```diff
@@ -170,14 +169,14 @@ export class Subscription implements SubscriptionLike {
 
     // Add `this` as parent of `subscription` if that's not already the case.
     let { _parentOrParents } = subscription;
-    if (_parentOrParents === null) {
+    if (_parentOrParents == null) {
```
cartant, line 172: thought: Yikes. I wonder if there's a lint rule to catch this sort of thing?

cartant (approved):
LGTM. IDK why ts-api-guardian does funky stuff with the union orderings. I might investigate it sometime. Although, Craig Spence tweeted about it, last night, so maybe he'll figure it out. 🤞

src/internal/Subscription.ts
benlesh: This is intentional.  `== null` matches `null` or `undefined`.

cartant: Yeah. I know, your change fixes a potential bug. I'm wondering if comparisons like this could be elsewhere. I know there is a lint rule that relaxes `===` when comparing `null`, but IDK that there's one that enforces/recommends the use of `==` when comparing against `null`. With the varied use of `null` and `undefined` in this codebase - callers can pass either - it would be interesting - to me, anyway - to see if this occurs elsewhere. A search could find it, but a lint rule could enforce it.

cartant: It seems that there are about half a dozen other uses of either `=== null` or `=== undefined` throughout the codebase (ignoring the docs app).

PR #5657, opened 2020-08-18
refactor(VirtualTimeScheduler): remove sortActions from public API

benlesh:
BREAKING CHANGE: The static `sortActions` method on `VirtualTimeScheduler` is no longer publicly exposed by our TS types.


cartant (approved):
LGTM 🎉

PR #5659, opened 2020-08-20
feat(Subscription): remove will now remove any teardown by reference

benlesh:
Previously, only `Subscription` instances could be removed from a `Subscription`. Now any valid `TeardownLogic` that has been added to a `Subscription` may be removed by passing the same instance to `add`.

For example:

```ts
const teardown = () => console.log('called');

subscription.add(teardown);
subscription.remove(teardown);
subscription.unsubscribe();

// Will log nothing
```

Similarly with "unsubscribables":

```ts
const unsubscribable = {
    unsubscribe() {
        console.log('called');
    }
};

subscription.add(unsubscribable);
subscription.remove(unsubscribable);
subscription.unsubscribe();

// Will log nothing
```

- Cleans up and refactors Subscription substantially.
- Adds a few additional tests around Subscription and checking for teardown registries and unregistries

**Description:**

### DIFF
--- a/spec/Subscriber-spec.ts
+++ b/spec/Subscriber-spec.ts
@@ -2,6 +2,7 @@ import { expect } from 'chai';
 import { SafeSubscriber } from 'rxjs/internal/Subscriber';
 import { Subscriber, Observable } from 'rxjs';
 import { asInteropSubscriber } from './helpers/interop-helper';
+import { getRegisteredTeardowns } from './helpers/subscription';
 
 /** @test {Subscriber} */
 describe('Subscriber', () => {
@@ -128,5 +129,38 @@ describe('Subscriber', () => {
 
     subscriber.unsubscribe();
     expect(count).to.equal(1);
+  });
+
+  it('should close, unsubscribe, and unregister all teardowns after complete', () => {
+    let isUnsubscribed = false;
+    const subscriber = new Subscriber();
+    subscriber.add(() => isUnsubscribed = true);
+    subscriber.complete();
+    expect(isUnsubscribed).to.be.true;
+    expect(subscriber.closed).to.be.true;
+    expect(getRegisteredTeardowns(subscriber).length).to.equal(0);
+  });
+
+  it('should close, unsubscribe, and unregister all teardowns after error', () => {
+    let isUnsubscribed = false;
+    const subscriber = new Subscriber({
+      error: () => {
+        // Mischief managed!
+        // Adding this handler here to prevent the call to error from 
+        // throwing, since it will have an error handler now.
+      }
+    });
+    subscriber.add(() => isUnsubscribed = true);
+    subscriber.error(new Error('test'));
+    expect(isUnsubscribed).to.be.true;
+    expect(subscriber.closed).to.be.true;
+    expect(getRegisteredTeardowns(subscriber).length).to.equal(0);
+  });
+
+  
+  it('should unregister all teardowns after complete', () => {
+    const subscriber = new Subscriber();
+    subscriber.complete();
+    expect(getRegisteredTeardowns(subscriber).length).to.equal(0);
   });
 });
--- a/src/internal/Subscription.ts
+++ b/src/internal/Subscription.ts
@@ -1,3 +1,4 @@
+/** @prettier */
 import { isArray } from './util/isArray';
 import { isObject } from './util/isObject';
 import { isFunction } from './util/isFunction';
@@ -18,20 +19,27 @@ import { SubscriptionLike, TeardownLogic } from './types';
  */
 export class Subscription implements SubscriptionLike {
   /** @nocollapse */
-  public static EMPTY: Subscription = (function(empty: any) {
+  public static EMPTY: Subscription = (function (empty: any) {
     empty.closed = true;
     return empty;
-  }(new Subscription()));
+  })(new Subscription());
 
   /**
    * A flag to indicate whether this Subscription has already been unsubscribed.
    */
   public closed = false;
 
-  /** @internal */
-  protected _parentOrParents: Subscription | Subscription[] | null = null;
-  /** @internal */
-  private _subscriptions: SubscriptionLike[] | null = null;
+  /** If this subscription has been added to one parent, it will show up here */
+  private _singleParent?: Subscription;
+
+  /** If this subscription has been added to more than one parent, they will show up here. */
+  private _parents?: Subscription[];
+
+  /**
+   * The list of registered teardowns to execute upon unsubscription. Adding and removing from this
+   * list occurs in the {@link add} and {@link remove} methods.
+   */
+  private _teardowns?: Exclude<TeardownLogic, void>[];
 
   /**
    * @param {function(): void} [unsubscribe] A function describing how to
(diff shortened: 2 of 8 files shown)


src/internal/Subscription.ts
```diff
@@ -18,20 +19,27 @@ import { SubscriptionLike, TeardownLogic } from './types';
...
+  private _singleParent?: Subscription;
+
+  /** If this subscription has been added to more than one parent, they will show up here. */
+  private _parents?: Subscription[];
+
+  /**
+   * The list of registered teardowns to execute upon unsubscription. Adding and removing from this
+   * list occurs in the {@link add} and {@link remove} methods.
+   */
+  private _teardowns?: Exclude<TeardownLogic, void>[];
```
cartant, line 42: question: Is there a particular reason for not initializing these members to `undefined`. Won't this cause a de-opt if/when the members added outside of the ctor? And won't this code from `unsubscribe` involve a walking the prototype chain if the members are not present?

```ts
const { _parents, _singleParent, _teardowns } = this;
```

src/internal/Subscription.ts
```diff
@@ -53,180 +61,188 @@ export class Subscription implements SubscriptionLike {
...
 
-    this.closed = true;
-    this._parentOrParents = null;
-    // null out _subscriptions first so any child subscriptions that attempt
-    // to remove themselves from this subscription will noop
-    this._subscriptions = null;
+      const { _parents, _singleParent, _teardowns } = this;
+      this._parents = this._singleParent = this._teardowns = undefined;
+      let { _ctorUnsubscribe, _unsubscribe } = this as any;
+      // (this as any)._unsubscribe = (this as any)._ctorUnsubscribe = undefined;
```
cartant, line 70: nitpick: Can we remove this commented-out code?

src/internal/Subscription.ts
```diff
@@ -53,180 +61,188 @@ export class Subscription implements SubscriptionLike {
...
+    // If we already have one or more parents, we need
+    // to add to the list of parents.
+    if ((this._parents || this._singleParent) && this._singleParent !== parent) {
+      if (this._singleParent) {
+        // We are adding our second parent.
+        // Move the single parent over to the parents list.
+        this._parents = [this._singleParent];
+        this._singleParent = undefined;
+      }
+      if (!this._parents!.includes(parent)) {
```
cartant, line 189: issue: There is no need to check this here, AFAICT. `_hasParent` is called before `_addParent` is called - it's called in only one place - so this check means that the perf penalty of scanning the array is paid twice.

src/internal/Subscription.ts
```diff
@@ -53,180 +61,188 @@ export class Subscription implements SubscriptionLike {
...
-      // the `this`, which is the common case that we optimize for.
-      subscription._parentOrParents = this;
-    } else if (_parentOrParents instanceof Subscription) {
-      if (_parentOrParents === this) {
-        // The `subscription` already has `this` as a parent.
-        return;
+  private _addParent(parent: Subscription) {
+    // If we already have one or more parents, we need
+    // to add to the list of parents.
+    if ((this._parents || this._singleParent) && this._singleParent !== parent) {
```
cartant, line 182: nitpick: For me, the parentheses in this condition make it more difficult to understand. To me, it would be easier to understand as `this._parents || (this._singleParent && this._singleParent !== parent)`. Prettier will likely remove the parentheses, as they are not necessary because of operator precedence, but putting them around `this._parents || this._singleParent` doesn't make this easier to understand, IMO.

src/internal/Subscription.ts
```diff
@@ -53,180 +61,188 @@ export class Subscription implements SubscriptionLike {
...
-    let { _parentOrParents } = subscription;
-    if (_parentOrParents == null) {
-      // If we don't have a parent, then set `subscription._parents` to
-      // the `this`, which is the common case that we optimize for.
-      subscription._parentOrParents = this;
-    } else if (_parentOrParents instanceof Subscription) {
-      if (_parentOrParents === this) {
-        // The `subscription` already has `this` as a parent.
-        return;
+  private _addParent(parent: Subscription) {
```
cartant, line 179: question: The implementation makes numerous 'reads' of members via `this`. Would it be worthwhile - from a perf perspective - to read them once - using the destructuring syntax - and then access them via `this` only when assigning to the members?

cartant (changes requested):

src/internal/Subscription.ts
benlesh: answer: I was tired. :P

src/internal/Subscription.ts
benlesh: I must have been tired. I don't like any of of this method now... Refactored completely to be more readable and less nonsense. Those `_parents!` checks were a smell.

src/internal/Subscription.ts
benlesh: Ah! Right you are! I was under the impression that TS would initialize them to `undefined` for us, but nope. Although, I wonder if "deopt"-wise if `null` is the proper choice. 🤔

spec/Subscriber-spec.ts
```diff
@@ -129,4 +130,37 @@ describe('Subscriber', () => {
...
+    expect(isUnsubscribed).to.be.true;
+    expect(subscriber.closed).to.be.true;
+    expect(getRegisteredTeardowns(subscriber).length).to.equal(0);
+  });
+
+  
+  it('should unregister all teardowns after complete', () => {
+    const subscriber = new Subscriber();
+    subscriber.complete();
+    expect(getRegisteredTeardowns(subscriber).length).to.equal(0);
```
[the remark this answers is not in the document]
benlesh: Yikes. Fixing that test.

src/internal/Subscription.ts
```diff
@@ -53,180 +59,201 @@ export class Subscription implements SubscriptionLike {
...
-      // from a prototype method to an instance property - see notifyNext in
-      // RetryWhenSubscriber.
-      if (_ctorUnsubscribe) {
-        (this as any)._unsubscribe = undefined;
-      }
-      try {
-        _unsubscribe.call(this);
-      } catch (e) {
-        errors = e instanceof UnsubscriptionError ? flattenUnsubscriptionErrors(e.errors) : [e];
+      if (_parents) {
```
cartant, line 75: nitpick: This could be an `else if` as the implementation is guaranteeing that `_parents` will be `null` if `_singleParent` is non-`null`. IMO, this is significant not just for perf, but also to aid a reader's understanding: if the guarantee is made, IMO, the code paths in the implementation should reflect that - rather than suggest the guarantee isn't here.

src/internal/Subscription.ts
```diff
@@ -53,180 +59,201 @@ export class Subscription implements SubscriptionLike {
...
+    if (this._singleParent) {
+      if (this._singleParent === parent) {
+        this._singleParent = null;
+      }
+    } else if (this._parents) {
+      const index = this._parents.indexOf(parent);
+      if (index >= 0) {
+        this._parents.splice(index, 1);
+      }
+    }
```
cartant, line 218: nitpick: There's a fair bit of reading via `this` in here that could be done using destructured variables.

PR #5660, opened 2020-08-20
fix: use != and == to compare null/undefined delays

Zcating:
The type checking shows that parameter `delay` will never be null.

### DIFF
--- a/src/internal/scheduler/QueueAction.ts
+++ b/src/internal/scheduler/QueueAction.ts
@@ -35,7 +35,7 @@ export class QueueAction<T> extends AsyncAction<T> {
     // If delay exists and is greater than 0, or if the delay is null (the
     // action wasn't rescheduled) but was originally scheduled as an async
     // action, then recycle as an async action.
-    if ((delay !== null && delay > 0) || (delay === null && this.delay > 0)) {
+    if ((delay !== undefined && delay > 0) || (delay === undefined && this.delay > 0)) {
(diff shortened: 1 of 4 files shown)


src/internal/scheduler/QueueAction.ts
```diff
@@ -35,7 +35,7 @@ export class QueueAction<T> extends AsyncAction<T> {
     // If delay exists and is greater than 0, or if the delay is null (the
     // action wasn't rescheduled) but was originally scheduled as an async
     // action, then recycle as an async action.
-    if ((delay !== null && delay > 0) || (delay === null && this.delay > 0)) {
+    if ((delay !== undefined && delay > 0) || (delay === undefined && this.delay > 0)) {
```
cartant, line 38: You're right that these tests are a problem, but they should really be:

```ts
if ((delay != null && delay > 0) || (delay == null && this.delay > 0)) {
```

As it's possible that the `delay` argument could be either `null` or `undefined`. The implementation has to support JS and TS callers and JS callers could pass `null`. The `==` and `!=` comparisons will 'work' with either `null` or `undefined`.

Zcating: Got it. This means `delay` can be null or undefined in JS.

Zcating:
You're welcome.

cartant:
This will need to stay open until Ben has reviewed and merged it.

Zcating:
Sorry, Will remember it.

PR #5661, opened 2020-08-20
refactor(ajax): Move to simple Observable implementation. Fix errors.

### LINKED ISSUE #2801: Observable Ajax POST with wrong Content-Type after PUT
I'm trying to upload an image but I found that, every now and then, the server says that the binary image was not sent.
After some tests I found that the POST calls stops working after a PUT call and that is because the "Content-Type" of the PUT remains in the Request Headers. Those calls are not mapped, chained or consecutive, simply different user actions.

I've made some changes to AjaxObservable.ts but don't know if it is the right solution.

**RxJS version:**
5.4.3

**Code to reproduce:**
The **POST** code goes something like this:
```javascript
...
const formData = new FormData();
formData.append('picture', action.payload.image);
return ajax.post(
  `${URL_API}/action/picture`,
  formData,
  AUTH_HEADERS,
)
...
```
The **PUT** code goes something like this:
```javascript
...
const newSettings = {
  ...
};
return ajax.put(
  `${URL_API}/action/info`,
  newSettings,
  AUTH_HEADERS,
)
...
```

**Expected behavior:**
I expected the Content-Type header to change according to the method, like.
```
1st call POST -> Content-Type: multipart/form-data; boundary=... (has an attachment)
2nd call PUT  -> Content-Type: application/x-www-form-urlencoded; charset=UTF-8
3rd call POST -> Content-Type: multipart/form-data; boundary=... (has an attachment)
```
**Actual behavior:**
The 3rd call has the binary data but the wrong Content-Type so the server won't get it.
```
1st call POST -> Content-Type: multipart/form-data; boundary=... (has an attachment)
2nd call PUT  -> Content-Type: application/x-www-form-urlencoded; charset=UTF-8
3rd call POST -> Content-Type: application/x-www-form-urlencoded; charset=UTF-8 (has an attachment)
```

**My actual solution:**
I fount that the header is not correctly unset in the .../observable/dom/AjaxObservable.ts:203 when the body has an FormData.

This is the actual code.
```javascript
    // ensure content type is set
    if (!('Content-Type' in headers) && !(root.FormData && request.body instanceof root.FormData) && typeof request.body !== 'undefined') {
      headers['Content-Type'] = 'application/x-www-form-urlencoded; charset=UTF-8';
    }

    // properly serialize body
    request.body = this.serializeBody(request.body, request.headers['Content-Type']);
```
My solution code.
```javascript
    // force Content-Type to unset
    if (root_1.root.FormData && request.body instanceof root_1.root.FormData) {
        delete headers['Content-Type'];
    }
    // ensure content type is set
    if (!('Content-Type' in headers) && !(root.FormData && request.body instanceof root.FormData) && typeof request.body !== 'undefined') {
      headers['Content-Type'] = 'application/x-www-form-urlencoded; charset=UTF-8';
    }

    // properly serialize body
    request.body = this.serializeBody(request.body, headers['Content-Type']);
```
The issue was solved, maybe this helps some one or you can guide me to the right solution.

### LINKED ISSUE #2837: AJAX Content-Type Handling.
When sending data using the AJAX methods the existing code attempts to serialise the data using its own rules which do not match the XHR serialisation rules.  For my use notably it doesn't automatically handle XML Document objects which XHR does.  I'm not sure why the decision was made to make 'application/x-www-form-urlencoded' the default content type, and while the serialisation code is helpful I feel it should be opt in rather than the default behaviour.

Additionally, when checking to see if the user has supplied a 'Content-Type' header to override the default behaviour it only does a case-sensitive check despite the fact that HTTP headers are case insensitive.  It should instead either normalise user provided headers, or else do case insensitive checks for relevant headers, or else it should be clearly documented that it will interrogate specific headers in a case insensitive fashion.  See: https://github.com/ReactiveX/rxjs/blob/7bb8280181bf1d5b0a2d3338ede8bea5e30f64c7/src/observable/dom/AjaxObservable.ts#L201-L211

At the very least clearly documenting the default serialisation behaviour and how to use 'Content-Type' to override the default behaviour is required.

benlesh:
- Refactors `ajax` to use a simple `Observable` implementation.
- Moves code to a new ajax-specific folder.
- Adds a lot more documentation to classes and support classes.
- Differentiates between the configuration passed to `ajax` (`AjaxConfig`) and the request values used to make the HTTP request (`AjaxRequest`), as the latter has more required values.
- Ensures that no configuration values are mutated while making the request. (resolves #2801)
- Ensures there is at least one valid test for `ajax.patch`. The old one was sketchy.
- Adds better comments throughout the code.
- Adds better typing to `ajax` functions.
- Also ensures x-requested-with header is set properly
- Adds comments
- Code clean up
- Normalize headers to be lowercase, since they are case insensitive.
- Updates tests
- Remove superfluous checks in unsubscribe
- Ajax will now allow default behavior as defined in the specs:
  - https://xhr.spec.whatwg.org/#the-send()-method
  - https://fetch.spec.whatwg.org/#concept-bodyinit-extract
- if the `body` is set to: `Blob`, `ArrayBuffer`, any array buffer view (like a byte sequence, e.g. `Uint8Array`, etc), `FormData`, `URLSearchParams`, `string`, or `ReadableStream`, default handling is use. If the `body` is otherwise `typeof` `"object"`, then it will be converted to JSON via `JSON.stringify`, and the `Content-Type` header will be set to `application/json;charset=utf-8`. All other types will emit an error. (resolves #2837)
- Puts the behavior in-line with Axios, et al.

BREAKING CHANGE: For TypeScript users, `AjaxRequest` is no longer the type that should be explicitly used to create an `ajax`. It is now `AjaxConfig`, although the two types are compatible, only `AjaxConfig` has `progressSubscriber` and `createXHR`.

BREAKING CHANGE: `ajax` body serialization will now use default XHR behavior in all cases. If the body is a `Blob`, `ArrayBuffer`, any array buffer view (like a byte sequence, e.g. `Uint8Array`, etc), `FormData`, `URLSearchParams`, `string`, or `ReadableStream`, default handling is use. If the `body` is otherwise `typeof` `"object"`, then it will be converted to JSON via `JSON.stringify`, and the `Content-Type` header will be set to `application/json;charset=utf-8`. All other types will emit an error.

BREAKING CHANGE: The `Content-Type` header passed to `ajax` configuration no longer has any effect on the serialization behavior of the AJAX request.


src/internal/observable/dom/AjaxObservable.ts
```diff
@@ -2,14 +2,15 @@
 import { Observable } from '../../Observable';
 import { Subscriber } from '../../Subscriber';
 import { TeardownLogic, PartialObserver } from '../../types';
+import { reduce } from 'lodash';
```
cartant, line 5: issue: VS Code auto imports are the worst.

src/internal/observable/dom/AjaxObservable.ts
```diff
@@ -140,7 +161,11 @@ export class AjaxSubscriber<T> extends Subscriber<Event> {
       }
 
       // set headers
-      this.setHeaders(xhr, headers!);
+      for (const key in headers) {
+        if (headers.hasOwnProperty(key)) {
+          xhr.setRequestHeader(key, (headers as any)[key]);
```
cartant, line 166: nitpick: with `headers` having a `Record` type, the `any` should not be necessary.

src/internal/observable/dom/AjaxObservable.ts
```diff
@@ -71,28 +56,63 @@ export class AjaxObservable<T> extends Observable<T> {
...
+      // Override with passed user values
+      ...request,
+
+      // Set values we ensured above
+      headers,
+      body,
+    };
+
+    // Create our XHR so we can get started.
+    this.xhr = request.createXHR ? request.createXHR() : new XMLHttpRequest();
```
cartant, line 115: question: With this moved out of the `try`/`catch` - in `send` - what will happen with any errors effect here? What should happen?

cartant (changes requested):

src/internal/ajax/ajax.ts
```diff
@@ -262,3 +268,189 @@ export const ajax: AjaxCreationMethod = (() => {
...
+        xhr.send(body);
+      } else {
+        xhr.send();
+      }
+    } catch (err) {
+      destination.error(err);
+    }
+
+    return () => {
+      if (!done && xhr) {
```
tonivj5, line 427: I think `done` could be removed? https://github.com/angular/angular/pull/37367/files#diff-26683cfb34fa97883213282055aef1e7R342

```suggestion
      if (xhr && xhr.readyState !== xhr.DONE) {
```

josepot: Either that or move `done` outside of the `if` statement, right? Otherwise, I think that it will try to abort requests that have completed with 4xx and 5xx codes.

benlesh: Good point. Thank you.

benlesh:
As I'm going through the backlog of issues, I keep finding stuff related to ajax, so I've added a couple more commits to fix another (nasty) issue.

cartant (approved):

cartant:
This also seems to need an API guardian update.

benlesh:
The API guardian is killing me. It seems like there's an issue where you have to rebase, THEN update api_guardian before it actually works. This is the third PR where I've had issues. The thing it's complaining about is unrelated to the PR.

Issue #5668, opened 2020-08-22
Change homepage setting in package.json to the official docs site


BioPhoton:
### Documentation Related To Component:

The `package.json` file.

### Please check those that apply

- [x] typo
- [ ] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

### Description Of The Issue

The `package.json` file contains a `homepage` property that points to the github docs. IMO this should point to `https://rxjs.dev`

cartant:
Closed by https://github.com/ReactiveX/rxjs/pull/5669

PR #5669, opened 2020-08-22
fix(package.json): change homepage setting to official docs site.

BioPhoton:
**Description:**
Change the homepage setting in package.json to the official docs site.

#5668


benlesh:
Thanks, @BioPhoton!

PR #5670, opened 2020-08-23
refactor(Subscription): Get rid of _unsubscribe weirdness

benlesh:
This is just another step toward getting what we're doing with inheritance to be a little more sane, ultimately, this may be refactored further, as I am not happy with some of these subscribers adding their own teardowns, but this is a start. The goal was to get rid of the weirdness around `_unsubscribe` in our code base.

- Moves to calling `_unsubscribe` a new internal name in `Subscription`: `initialTeardown`, because it's slightly more descriptive
- Removes `_unsubscribe` usage from the entire code base.
- Formatted a few files just for my sanity.

### DIFF
--- a/src/internal/Subscriber.ts
+++ b/src/internal/Subscriber.ts
@@ -160,6 +160,7 @@ export class SafeSubscriber<T> extends Subscriber<T> {
               error?: ((e?: any) => void) | null,
               complete?: (() => void) | null) {
     super();
+    this.add(this._teardown);
 
     let next: ((value: T) => void) | undefined;
     let context: any = this;
@@ -277,8 +278,7 @@ export class SafeSubscriber<T> extends Subscriber<T> {
     return false;
   }
 
-  /** @internal This is an internal implementation detail, do not use. */
-  _unsubscribe(): void {
+  private _teardown = () => {
     const { _parentSubscriber } = this;
     this._context = null;
     this._parentSubscriber = null!;
--- a/src/internal/Subscription.ts
+++ b/src/internal/Subscription.ts
@@ -71,7 +71,7 @@ export class Subscription implements SubscriptionLike {
         }
       }
 
-      const initialTeardown = this.initialTeardown;
+      const { initialTeardown } = this;
       if (isFunction(initialTeardown)) {
         try {
           initialTeardown();
--- a/src/internal/Subscription.ts
+++ b/src/internal/Subscription.ts
@@ -40,15 +40,10 @@ export class Subscription implements SubscriptionLike {
   private _teardowns: Exclude<TeardownLogic, void>[] | null = null;
 
   /**
-   * @param {function(): void} [unsubscribe] A function describing how to
-   * perform the disposal of resources when the `unsubscribe` method is called.
+   * @param initialTeardown A function executed first as part of the teardown
+   * process that is kicked off when {@link unsubscribe} is called.
    */
-  constructor(unsubscribe?: () => void) {
-    if (unsubscribe) {
-      (this as any)._ctorUnsubscribe = true;
-      (this as any)._unsubscribe = unsubscribe;
-    }
-  }
+  constructor(private initialTeardown?: () => void) {}
 
   /**
    * Disposes the resources held by the subscription. May, for instance, cancel
@@ -76,22 +71,10 @@ export class Subscription implements SubscriptionLike {
         }
       }
 
-      const _unsubscribe = (this as any)._unsubscribe;
-      if (isFunction(_unsubscribe)) {
-        // It's only possible to null _unsubscribe - to release the reference to
-        // any teardown function passed in the constructor - if the property was
-        // actually assigned in the constructor, as there are some classes that
-        // are derived from Subscriber (which derives from Subscription) that
-        // implement an _unsubscribe method as a mechanism for obtaining
-        // unsubscription notifications and some of those subscribers are
-        // recycled. Also, in some of those subscribers, _unsubscribe switches
-        // from a prototype method to an instance property - see notifyNext in
-        // RetryWhenSubscriber.
-        if ((this as any)._ctorUnsubscribe) {
-          (this as any)._unsubscribe = undefined;
-        }
+      const initialTeardown = this.initialTeardown;
+      if (isFunction(initialTeardown)) {
         try {
-          _unsubscribe.call(this);
+          initialTeardown();
         } catch (e) {
           errors = e instanceof UnsubscriptionError ? e.errors : [e];
         }
--- a/src/internal/observable/ConnectableObservable.ts
+++ b/src/internal/observable/ConnectableObservable.ts
@@ -74,18 +74,19 @@ class ConnectableSubscriber<T> extends Subscriber<T> {
   constructor(protected destination: Subject<T>,
               private connectable: ConnectableObservable<T>) {
     super();
+    this.add(this._teardown);
   }
   protected _error(err: any): void {
-    this._unsubscribe();
+    this._teardown();
     super._error(err);
   }
   protected _complete(): void {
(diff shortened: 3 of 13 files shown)


benlesh:
In retrospect, I might be able to just have a `private intialTeardown = () =>` property on those that are doing `this.add(this._teardown)`... but again, one thing at a time. I'm _trying_ to be incremental, but I could do that in this PR, I suppose. After an initial review.

src/internal/Subscriber.ts
```diff
@@ -160,6 +160,7 @@ export class SafeSubscriber<T> extends Subscriber<T> {
               error?: ((e?: any) => void) | null,
               complete?: (() => void) | null) {
     super();
+    this.add(this._teardown);
```
cartant, line 163: suggestion: Use `this._teardown.bind(this)` here rather than a per-subscriber `_teardown` property. That is, make `_teardown` a method rather than an instance property. Are there perf benefits with using the arrow function? Won't allocating the function per-subscriber be more expensive than the `bind`?

src/internal/Subscription.ts
```diff
@@ -40,15 +40,10 @@ export class Subscription implements SubscriptionLike {
...
+   * @param initialTeardown A function executed first as part of the teardown
+   * process that is kicked off when {@link unsubscribe} is called.
    */
-  constructor(unsubscribe?: () => void) {
-    if (unsubscribe) {
-      (this as any)._ctorUnsubscribe = true;
-      (this as any)._unsubscribe = unsubscribe;
-    }
-  }
+  constructor(private initialTeardown?: () => void) {}
```
cartant, line 46: 🎉

src/internal/Subscription.ts
```diff
@@ -76,22 +71,10 @@ export class Subscription implements SubscriptionLike {
...
-        // are derived from Subscriber (which derives from Subscription) that
-        // implement an _unsubscribe method as a mechanism for obtaining
-        // unsubscription notifications and some of those subscribers are
-        // recycled. Also, in some of those subscribers, _unsubscribe switches
-        // from a prototype method to an instance property - see notifyNext in
-        // RetryWhenSubscriber.
-        if ((this as any)._ctorUnsubscribe) {
-          (this as any)._unsubscribe = undefined;
-        }
+      const initialTeardown = this.initialTeardown;
```
cartant, line 74: nitpick: `const { initialTeardown } = this;`

src/internal/observable/ConnectableObservable.ts
```diff
@@ -74,18 +74,19 @@ class ConnectableSubscriber<T> extends Subscriber<T> {
   constructor(protected destination: Subject<T>,
               private connectable: ConnectableObservable<T>) {
     super();
+    this.add(this._teardown);
```
cartant, line 77: suggestion: (as above) `this.add(this._teardown.bind(this))`?

src/internal/observable/ConnectableObservable.ts
```diff
@@ -125,9 +126,10 @@ class RefCountSubscriber<T> extends Subscriber<T> {
   constructor(destination: Subscriber<T>,
               private connectable: ConnectableObservable<T>) {
     super(destination);
+    this.add(this._teardown);
```
cartant, line 129: suggestion: (as above) this.add(`this._teardown.bind(this))`?

src/internal/operators/bufferTime.ts
```diff
@@ -139,6 +139,7 @@ class BufferTimeSubscriber<T> extends Subscriber<T> {
               private maxBufferSize: number,
               private scheduler: SchedulerLike) {
     super(destination);
+    this.add(this._teardown);
```
cartant, line 142: suggestion: (as above) this.add(`this._teardown.bind(this))`?

src/internal/operators/bufferWhen.ts
```diff
@@ -75,6 +75,7 @@ class BufferWhenSubscriber<T> extends SimpleOuterSubscriber<T, any> {
 
   constructor(destination: Subscriber<T[]>, private closingSelector: () => Observable<any>) {
     super(destination);
+    this.add(this._teardown);
```
cartant, line 78: suggestion: (as above) this.add(`this._teardown.bind(this))`?

src/internal/operators/groupBy.ts
```diff
@@ -251,17 +251,17 @@ class GroupBySubscriber<T, K, R> extends Subscriber<T> implements RefCountSubscr
  */
 class GroupDurationSubscriber<K, T> extends Subscriber<T> {
   constructor(private key: K,
-              private group: Subject<T>,
+              group: Subject<T>,
               private parent: GroupBySubscriber<any, K, T | any>) {
     super(group);
+    this.add(this._teardown);
```
cartant, line 257: suggestion: (as above) this.add(`this._teardown.bind(this))`?

src/internal/operators/refCount.ts
```diff
@@ -89,9 +89,10 @@ class RefCountSubscriber<T> extends Subscriber<T> {
   constructor(destination: Subscriber<T>,
               private connectable: ConnectableObservable<T>) {
     super(destination);
+    this.add(this._teardown);
```
cartant, line 92: suggestion: (as above) this.add(`this._teardown.bind(this))`?

src/internal/operators/window.ts
```diff
@@ -80,6 +80,7 @@ class WindowSubscriber<T> extends SimpleOuterSubscriber<T, any> {
 
   constructor(destination: Subscriber<Observable<T>>) {
     super(destination);
+    this.add(this._teardown);
```
cartant, line 83: suggestion: (as above) this.add(`this._teardown.bind(this))`?

src/internal/operators/windowCount.ts
```diff
@@ -98,6 +98,7 @@ class WindowCountSubscriber<T> extends Subscriber<T> {
               private windowSize: number,
               private startWindowEvery: number) {
     super(destination);
+    this.add(this._teardown);
```
cartant, line 101: suggestion: (as above) this.add(`this._teardown.bind(this))`?

src/internal/operators/windowToggle.ts
```diff
@@ -90,6 +90,7 @@ class WindowToggleSubscriber<T, O> extends ComplexOuterSubscriber<T, any> {
               private openings: Observable<O>,
               private closingSelector: (openValue: O) => Observable<any>) {
     super(destination);
+    this.add(this._teardown);
```
cartant, line 93: suggestion: (as above) this.add(`this._teardown.bind(this))`?

src/internal/scheduler/AsyncAction.ts
```diff
@@ -20,6 +20,7 @@ export class AsyncAction<T> extends Action<T> {
   constructor(protected scheduler: AsyncScheduler,
               protected work: (this: SchedulerAction<T>, state?: T) => void) {
     super(scheduler, work);
+    this.add(this._teardown);
```
cartant, line 23: suggestion: (as above) this.add(`this._teardown.bind(this))`?

cartant (approved):
One suggestion (in multiple places) and one nitpick.

src/internal/Subscriber.ts
benlesh: Using `bind` also [allocates a new function per subscriber](https://github.com/v8/v8/blob/4b9b23521e6fd42373ebbcb20ebe03bf445494f9/src/builtins/builtins-function.cc#L199-L201).

In most previous iterations of V8, it was also slower, overall. I think that's still the case, just by a narrower margin. In general, IMO, we should avoid using `bind` where possible. Arrow functions, simple closures, call, and apply are better alternatives.

Issue #5174, opened 2019-12-03
Unexpected synchronous behavior of map/switchMap


ivan7237d:
## Bug Report

**Current Behavior**
The following code outputs 1, then 0.

```js
const subject = new BehaviorSubject(0);
subject
  .pipe(
    map(value => {
      if (value === 0) {
        subject.next(1);
      }
      return value;
    }),
  )
  .subscribe(console.log);
```

**Reproduction**
https://stackblitz.com/edit/rxjs-mbgssc?devtoolsheight=60

**Expected behavior**
The expected behavior is that the order would be 0 then 1. 

**Environment**
- RxJS version: 6.5.2

**Additional context**
The actual, a little bit more complicated issue that I've run into IRL is reproduced by the following Stackblitz: https://stackblitz.com/edit/rxjs-ylzbfn?devtoolsheight=60. There, the switchMap operator is subscribed to two inner observables at the same time (console outputs 2 values every second, but should output only 1 value every second). While investigating that issue, I've run into the above-described unexpected behavior of the map() operator, so I thought I'd like to clear that one up first.

kwonoj:
Coincidentally same q has asked in same crux : https://github.com/ReactiveX/rxjs/issues/5173

This is due to you're doing reentrant synchronosly and expected behavior. Generally you need to handle properly via scheduling / escaping if you want to control order of emission via reentrant.

ivan7237d:
@kwonoj thanks for the answer, but how can it be expected behavior that the switchMap operator is subscribed simultaneously to two inner observables (as in the second Stackblitz)? 

Also, maybe it would be helpful to describe what actually happened in a real-life app. There is a switchMap operator that converts a URL into what the user sees on screen. Normally all works fine, except when the user opens a homepage URL, because instead of mapping that URL to some UI, we redirect the user to some other page (like the last previously opened document) by setting the URL to a different value via History API. So in this case the browser's `location` is the equivalent of a Subject, and the whole thing is also synchronous. The result? Even though I haven't done anything obviously criminal, I see my switchMap operator emitting values corresponding to both the current URL and some previous URL, and the UI shown to the user flickers between two different views. The problem also was that while debugging this, I automatically assumed that switchMap can only be subscribed to one inner observable at a time, so it was the last line of investigation I tried.

kwonoj:
[...]

Issue #5173, opened 2019-12-03
Wired Subject behavior


Brooooooklyn:
## Bug Report

**Current Behavior**
```js
const rxjs = require('rxjs')

const subject = new rxjs.Subject()

subject.subscribe({
  next: (v) => {
    if (v === 1) {
      subject.next(2)
    }
  }
})

subject.subscribe({
  next: (v) => {
    console.log(v)
  }
})

subject.next(1)

```

log 2
log 1

**Reproduction**
- REPL or Repo link:
https://stackblitz.com/edit/typescript-pw115w

```js
var your => (code) => here;
```

**Expected behavior**
log 1
log 2

**Environment**
- RxJS version: 6.5.3

cartant:
It might seems weird, but this the expected behaviour. Subscribers are notified in the order in which they subscribe and your first subscriber effects another next notification - essentially re-entering the notification if the subscribers.

kwonoj:
This is typical reentrant example and answer is well explained.

Issue #1759, opened 2016-06-10
First operator is not completing the observable.


onlyann:
**RxJS version: 5.0.0-beta.6

**Code to reproduce:
http://codepen.io/anon/pen/zBrPzj?editors=1111 

**Expected behavior:**
Subcribe should only be called once when using the `first`operator.

**Actual behavior:**
It doesn't complete the subscription.

**Additional information**

Replacing `first` with `take(1)` works as expected.

babeal:
To add some detail:

The call to the BehaviorSubject.next inside of the subscribe method seems to be the issue. Since first() calls complete after emitting the value in the _emitValue function, calling next on the behavior subject synchronously will cause an infinite loop.  This doesn't occur with take() since the recursive call will complete when the number of emitted items is greater than the count.  If you make the "next" call async, then the first operator behaves correctly.
 
```
  .subscribe((x) => {
      console.log("I have been called " + x);
      Rx.Observable.timer(1).map(_ => {
        console.log("hello");
        store$.next({a: x[0] + 1, b: x[1] + 1});
      }).subscribe();

    }
```

Dorus:
It's worth nothing that you are not allowed to call `store$.next` concurrent. This is one of the serialization properties of Rx, callers need to ensure they implement a proper locking mechanism, so that downstream operators do not have to run extensive overhead to protect them self against multiple concurrent calls.

blesh:
@onlyann yup, that's a bug. 

It appears to be an issue with the synchronous nexting inside of the observation. Here's a more minimal example showing the bug:

```js
var store$ = new Rx.Subject();

store$.first().subscribe((x) => {
  console.log("I have been called " + x);
 store$.next({a: x[0] + 1, b: x[1] + 1});
});

store$.next({ a: 1, b: 2 })
```

But if you use a `setTimeout` it works fine:

```js
var store$ = new Rx.Subject();

store$.first().subscribe((x) => {
  console.log("I have been called " + x);
  setTimeout(() => store$.next({a: x[0] + 1, b: x[1] + 1}));
});

store$.next({ a: 1, b: 2 })
```

kwonoj:
Is it expected to allow to source emitting value in subscription in case of synchronous observable instead of should be scheduled differently?

jkyle:
`.take(1)` also does not complete when called with synchronous `next`ing. It behaves as expected in the example, but doesn't complete:
http://codepen.io/anon/pen/GqNjra?editors=0011

[...]

PR #5682, opened 2020-08-26
fix: handle synchronous reentrant during subscribing the inner sub

naiweizheng:
**Description:**

Some operators create inner subscriptions on `next()`. But during subscribing to the inner sub, some side effects may trigger another `next()` on this operator synchronously, i.e. synchronous reentrant, leading to a new inner subscription to be created. However the previous inner subscription is not yet set on the operator, which would cause unexpected behavior for some operators. It's clear to us that `switchMap` and `exhaustMap` behaves unexpectedly under this setting, because their correct behaviors are defined by the order of messages being received.

This PR is trying to address these unexpected behaviors so that:

- `switchMap` should unsubscribe previous inner sub when getting synchronously reentrance during subscribing the inner sub
- `exhaustMap` should ignore subsequent synchronous reentrances during subscribing the inner sub

It's yet to be determined if other operators' behaviors under this setting should be considered unexpected.

### DIFF
--- a/src/internal/operators/exhaustMap.ts
+++ b/src/internal/operators/exhaustMap.ts
@@ -113,7 +113,8 @@ class ExhaustMapSubscriber<T, R> extends SimpleOuterSubscriber<T, R> {
       const innerSubscriber = new SimpleInnerSubscriber(this);
       const destination = this.destination;
       destination.add(innerSubscriber);
-      this.innerSubscription = innerSubscribe(result, innerSubscriber);
+      this.innerSubscription = innerSubscriber;
+      innerSubscribe(result, innerSubscriber);
     }
   }
 
--- a/src/internal/operators/switchMap.ts
+++ b/src/internal/operators/switchMap.ts
@@ -129,7 +129,8 @@ class SwitchMapSubscriber<T, R> extends SimpleOuterSubscriber<T, R> {
     }
     const innerSubscriber = new SimpleInnerSubscriber(this);
     this.destination.add(innerSubscriber);
-    this.innerSubscription = innerSubscribe(result, innerSubscriber);
+    this.innerSubscription = innerSubscriber;
+    innerSubscribe(result, innerSubscriber);
   }
 
   protected _complete(): void {
--- a/spec/operators/exhaustMap-spec.ts
+++ b/spec/operators/exhaustMap-spec.ts
@@ -1,5 +1,5 @@
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
-import { concat, defer, Observable, of } from 'rxjs';
+import { concat, defer, Observable, of, BehaviorSubject } from 'rxjs';
 import { exhaustMap, mergeMap, takeWhile, map, take } from 'rxjs/operators';
 import { expect } from 'chai';
 import { asInteropObservable } from '../helpers/interop-helper';
@@ -450,5 +450,20 @@ describe('exhaustMap', () => {
     ).subscribe(() => { /* noop */ });
 
     expect(sideEffects).to.deep.equal([0, 1, 2]);
+  });
+
+  it('should ignore subsequent synchronous reentrances during subscribing the inner sub', () => {
+    const e = new BehaviorSubject(1);
+    const results: Array<number> = [];
+
+    e.pipe(
+      take(3),
+      exhaustMap(value => new Observable<number>(subscriber => {
+        e.next(value+1);
+        subscriber.next(value);
+      })),
+    ).subscribe(value => results.push(value));
+
+    expect(results).to.deep.equal([1]);
   });
 });
--- a/spec/operators/switchMap-spec.ts
+++ b/spec/operators/switchMap-spec.ts
@@ -1,7 +1,7 @@
 import { expect } from 'chai';
 import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 import { switchMap, mergeMap, map, takeWhile, take } from 'rxjs/operators';
-import { concat, defer, of, Observable } from 'rxjs';
+import { concat, defer, of, Observable, BehaviorSubject } from 'rxjs';
 import { asInteropObservable } from '../helpers/interop-helper';
 
 /** @test {switchMap} */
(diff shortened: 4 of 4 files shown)


kwonoj:
For me added test cases

```
const e = new BehaviorSubject(1);
const results: Array<number> = [];

e.pipe(
      take(3),
      exhaustMap(value => new Observable<number>(subscriber => {
        e.next(value+1);
```

are a direct example of synchronous reentrant, which is known to not supported and caller should schedule appropriately. Any specific reason do you think this is a bug to be fixed?

kwonoj:
ref: https://github.com/ReactiveX/rxjs/issues/5174#issuecomment-561303690 / https://github.com/ReactiveX/rxjs/issues/5173 etcs.

naiweizheng:
@kwonoj From the definition of switchMap for example:

> emitting values only from the most recently projected Observable.

It's not limited to asynchronous values, or has any restrictions on synchronous reentrant. The order of messages received doesn't have anything to do with async or not. Unless specified, we should expect it to be supported, otherwise it's an exception to the definition and should be well documented.

There are legitimate use case for synchronous reentrant. `switchMap` for example can be used for the central building block for a state machine where state transition is done by switch-mapping to different state Observables. In this case, a state Observable can decide to synchronously switching to another state upon subscription. This is where the unexpected behavior can be considered a bug. Of course you can say a "workaround" is to schedule the state transition asynchronously, but just trying to understand why it failed in the first place can cost a lot. And adhering to a workaround is adding another learning barrier and "gotcha" to developers in the team.

kwonoj:
1. I agree we may need some documentation about reentrant behavior. This was brought up couple of times.

2. If this behavior can happen without reentrant, I'd like to see and example and I believe this should be included as test case instead.

3. Whole rxjs v5, and beyond by design to not honor synchronous reentrant. I do not agree with core we should able to support all possible cases of synchronous reentrant. Scheduling reentrant is not a `workaround`, it is intended per current design of core behaves. Moreover, core team generally recommend avoiding reentrant itself: not only for this synchronous behavior, there are a lot of possible catch around reentrant behavior can cause unexpected behavior: it's due to several reasons that rx core does not support some / does not define some / and possible some unexpected bug lastly.

kwonoj:
for 3, not specifically for this operator but https://github.com/ReactiveX/rxjs/issues/1759#issuecomment-227575178 explains early decision why we did it: 

```
Protecting against re-entrency without throwing errors or dropping values would necessitate introducing unbounded buffers, which was a conscious design trade-off we made in the early days.
```

Until (if) we make other breaking design decisions, I doubt we'll offer official support for reentrant for all edge cases.

naiweizheng:
It's not like "if we support **any** reentrant edge cases our future plans would be blocked" right? I think the core team should view it case by case, breaking down the cost and benefits of supporting **some** reentrant usages. Of course, if core thinks a proposal is too intrusive to the design such that some future planning is affected, we can discuss about that in more extends. But to dismiss it just because not supporting any reentrant edge case is just unfortunate.

kwonoj:
As you said we may visit per each cases: but if fix is limited to synchronous reentrant, it's likely not able to be picked up. That's reason I was sugesting in `2` if test case can be replaced doesn't involve synchronous reentrant.

benlesh:
@naiweizheng is right, this is a fix, and it actually fixes a regression. This worked as expected in past versions and was broken by recent changes: https://codesandbox.io/s/gracious-visvesvaraya-3503j?file=/src/index.ts

I'm not sure about the method for fixing this yet... I still need to review it and think about it. But this bug is legit.

src/internal/operators/exhaustMap.ts
```diff
@@ -113,6 +113,7 @@ class ExhaustMapSubscriber<T, R> extends SimpleOuterSubscriber<T, R> {
       const innerSubscriber = new SimpleInnerSubscriber(this);
       const destination = this.destination;
       destination.add(innerSubscriber);
+      this.innerSubscription = innerSubscriber;
```
benlesh, line 116: I think that, in this case, we don't need line 117, because the return value is === innerSubscriber. we can actually call it an ignore the returned value. Perhaps adding a comment as to why.

src/internal/operators/switchMap.ts
```diff
@@ -123,13 +123,17 @@ class SwitchMapSubscriber<T, R> extends SimpleOuterSubscriber<T, R> {
...
+    let innerSubscription = this.innerSubscription;
     if (innerSubscription) {
       innerSubscription.unsubscribe();
     }
     const innerSubscriber = new SimpleInnerSubscriber(this);
     this.destination.add(innerSubscriber);
-    this.innerSubscription = innerSubscribe(result, innerSubscriber);
+    this.innerSubscription = innerSubscriber;
+    innerSubscription = innerSubscribe(result, innerSubscriber);
+    if (this.innerSubscription === innerSubscriber) {
```
benlesh, line 134: I think this is the same as the comment above. But please do check. I'm pretty sure that `innerSubscriber === innerSubscription`, and therefor also `=== this.innerSubscription`. So I think we just need to set `this.innerSubscription = innerSubscriber`, then call `innerSubscribe` and let the return value drop.

src/internal/operators/exhaustMap.ts
benlesh: I mean, double check, perhaps I'm wrong... But if memory serves me, they should be equal.

benlesh:
@naiweizheng thank you for catching this, and MOST importantly, acting on it and adding tests!

src/internal/operators/exhaustMap.ts
naiweizheng: `innerSubscribe()` can actually return `undefined`, at least for `Array` and `AsyncIterableIterator`. Considering that, it's actually incorrect to set the return value of `innerSubscribe()` to `this.innerSubscription`. In fact just setting `this.innerSubscription  = innerSubscriber`, and dropping the return value should be enough, and is indeed the correct way.

src/internal/operators/switchMap.ts
naiweizheng: Same as above, changed to just setting `this.innerSubscription = innerSubscriber`, and dropping the return value.

benlesh:
Nice! Thank you for your efforts, @naiweizheng!

PR #5686, opened 2020-09-01
fix(errors): Custom RxJS errors now all have a call stack

### LINKED ISSUE #4250: Errors do not get stack traces in 6.3.x
## Bug Report

**Current Behavior**

Stack traces are no longer recorded on rxjs errors as of 6.3.0.

**Reproduction**

6.3.3 (`e.stack` is `undefined`):

https://stackblitz.com/edit/rxjs-yabcir?file=index.ts

6.2.2 (`e.stack` is defined):

https://stackblitz.com/edit/rxjs-8y9kt4?file=index.ts

My guess is that https://github.com/ReactiveX/rxjs/pull/3967 is what broke the traces.

**Expected behavior**
Errors have stack traces

**Environment**
- Runtime: Chrome
- RxJS version: 6.3.x

**Possible Solution**

Possibly do something like:

```js
    if (!Error.captureStackTrace)
      this.stack = (new Error()).stack;
    else
      Error.captureStackTrace(this, this.constructor);
```

benlesh:
NOTE: Because we have to workaround compilation-to-ES5 inadequacies until call stack locations will show as being inside of the constructor implementation at the top level. This will go away as we are able to move the community to ES2015+

BREAKING CHANGE: Tests that are written with naive expectations against errors may fail now that errors have a proper `stack` property. In some testing frameworks, a deep equality check on two error instances will check the values in `stack`, which could be different.

fixes #4250

### DIFF
--- a/spec/operators/concatWith-spec.ts
+++ b/spec/operators/concatWith-spec.ts
@@ -2,14 +2,15 @@ import { expect } from 'chai';
 import { of, Observable } from 'rxjs';
 import { concatWith, mergeMap, take } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
-import { assertDeepEquals, NO_SUBS } from '../helpers/test-helper';
+import { NO_SUBS } from '../helpers/test-helper';
+import { observableMatcher } from '../helpers/observableMatcher';
 
 /** @test {concat} */
 describe('concat operator', () => {
   let rxTest: TestScheduler;
 
   beforeEach(() => {
-    rxTest = new TestScheduler(assertDeepEquals);
+    rxTest = new TestScheduler(observableMatcher);
   });
 
   it('should concatenate two cold observables', () => {
--- a/spec/util/createErrorClass-spec.ts
+++ b/spec/util/createErrorClass-spec.ts
@@ -1,0 +1,28 @@
+/** @prettier */
+import { createErrorClass } from 'rxjs/internal/util/createErrorClass';
+import { expect } from 'chai';
+
+describe('createErrorClass', () => {
+  it('should create a class that subclasses error and has the right properties', () => {
+    const MySpecialError: any = createErrorClass(
+      (_super) =>
+        function MySpecialError(this: any, arg1: number, arg2: string) {
+          _super(this);
+          this.message = 'Super special error!';
+          this.arg1 = arg1;
+          this.arg2 = arg2;
+        }
+    );
+
+    expect(MySpecialError).to.be.a('function');
+    const err = new MySpecialError(123, 'Test');
+    expect(err).to.be.an.instanceOf(Error);
+    expect(err).to.be.an.instanceOf(MySpecialError);
+    expect(err.name).to.equal('MySpecialError');
+    expect(err.constructor).to.equal(MySpecialError);
+    expect(err.stack).to.be.a('string');
+    expect(err.message).to.equal('Super special error!');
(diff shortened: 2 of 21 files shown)


src/internal/util/createErrorClass.ts
```diff
@@ -0,0 +1,15 @@
+/** @prettier */
+
+export function createErrorClass<T>(name: string, setup: (this: Error, ...args: any[]) => void): T {
```
benlesh, line 3: I think I'm going to add a doc comment about _why_ this exists.

spec/operators/concatWith-spec.ts
```diff
@@ -2,14 +2,15 @@ import { expect } from 'chai';
...
+import { NO_SUBS } from '../helpers/test-helper';
+import { observableMatcher } from '../helpers/observableMatcher';
 
 /** @test {concat} */
 describe('concat operator', () => {
   let rxTest: TestScheduler;
 
   beforeEach(() => {
-    rxTest = new TestScheduler(assertDeepEquals);
+    rxTest = new TestScheduler(observableMatcher);
```
benlesh, line 13: NOTE: These had to be changed, because a simple deep-equals check fails for `expect(new CustomError()).to.deep.equal(new CustomError())` once call stacks are in the mix.

spec/util/createErrorClass-spec.ts
```diff
@@ -0,0 +1,28 @@
+/** @prettier */
+import { createErrorClass } from 'rxjs/internal/util/createErrorClass';
+import { expect } from 'chai';
+
+describe('createErrorClass', () => {
+  it('should create a class that subclasses error and has the right properties', () => {
+    const MySpecialError: any = createErrorClass(
+      (_super) =>
+        function MySpecialError(this: any, arg1: number, arg2: string) {
+          _super(this);
```
kolodny, line 10: tiny nit: would it be cleaner to call `_super(this, 'Super special error')` instead?

src/internal/observable/dom/AjaxObservable.ts
```diff
@@ -353,7 +332,25 @@ const AjaxErrorImpl = (() => {
...
+      this.status = xhr.status;
+      this.responseType = xhr.responseType;
+      let response: any;
+      try {
+        response = getXHRResponse(xhr);
+      } catch (err) {
+        response = xhr.responseText;
+      }
+      this.response = response;
+      return this;
```
kolodny, line 351: nit: I don't think you need to keep returning `this`

src/internal/util/UnsubscriptionError.ts
```diff
@@ -1,29 +1,28 @@
...
+export const UnsubscriptionError: UnsubscriptionErrorCtor = createErrorClass(
+  (_super) =>
+    function UnsubscriptionError(this: any, errors: (Error | string)[]) {
+      _super(this);
+      this.message = errors
+        ? `${errors.length} errors occurred during unsubscription:
+${errors.map((err, i) => `${i + 1}) ${err.toString()}`).join('\n  ')}`
+        : '';
+      this.name = 'UnsubscriptionError';
+      (this as any).errors = errors;
```
kolodny, line 25: `this` already is `any`

spec/util/createErrorClass-spec.ts
benlesh: _super is probably a bad name, I'm just not sure what else to call it. :\

Issue #4429, opened 2018-12-18
Reconsider inspectTime


pcurrivan:
## Feature Request

Of the five *Time operators, it seems to me you guys removed the most useful one.  inspectTime was the only operator that behaved predictably, stayed up to date, and did not make you wait.

Here's a break down of the five *Time operators:
```
- - - 1 2 3 - - - - - - - - - - - 4 5 6 - - - 7 8 - - - - - - - - 9 - - - -  input sequence

- - - 1 - - - - - - - - - - - - - 4 - - - - - 7 - - - - - - - - - 9 - - - -  throttleTime(4) - does not stay up to date
- - - 0 1 2 3 - - - - - - - - - - 0 1 2 3 - - 0 1 2 3 - - - - - - 0 1 2 3 -  (timer) - emit on 0

- - - - - - - 3 - - - - - - - - - - - - - 6 - - - - - 8 - - - - - - - - - 9  auditTime(4) - makes you wait
- - - 0 1 2 3 4 - - - - - - - - - 0 1 2 3 4 - 0 1 2 3 4 - - - - - 0 1 2 3 4  (timer) - emit on 4

- - - - - - - - - 3 - - - - - - - - - - - - - - - - - - 8 - - - - - - - - 9  debounceTime(4) - makes you wait even more, skips entirely bursts that are close together
- - - 0 0 0 1 2 3 4 - - - - - - - 0 0 0 1 2 3 0 0 1 2 3 4 - - - - 0 1 2 3 4  (timer) - emit on 4, input refreshes timer

- - - - 2 - - - 3 - - - - - - - - - - - 6 - - - 8 - - - - - - - - - - - 9 -  sampleTime(4) - *erratically* makes you wait
0 1 2 3 0 1 2 3 0 1 2 3 0 1 2 3 0 1 2 3 0 1 2 3 0 1 2 3 0 1 2 3 0 1 2 3 0 1  (timer) - emit on 0 if input since last emission, timer restarts automatically

- - - 1 - - - 3 - - - - - - - - - 4 - - - 6 - - - 8 - - - - - - - 9 - - - -  inspectTime(4) - simply limits the frequency.  REMOVED from RxJS https://github.com/ReactiveX/RxJS/commit/17341a4
- - - 0 1 2 3 0 1 2 3 - - - - - - 0 1 2 3 0 1 2 3 0 1 2 3 - - - - 0 1 2 3 -  (timer) - emit on 0 and 4
```

[...]

PR #5687, opened 2020-09-02
fix(throttleTime): ensure the spacing between throttles is always at least the throttled amount

### LINKED ISSUE #2727: throttleTime(n, undefined, {leading: true, trailing: true}) ... double output

### LINKED ISSUE #3712: Invalid behavior of ThrottleTime operator

benlesh:
Works to align the behavior with expectations set by lodash's throttle

- Updates tests
- Ensures trailing throttle will wait to notify and then complete
- Ensures that every time we emit a value a new throttle period starts

fixes #3712
related #4864
fixes #2727
closes #4727
related #4429

### DIFF
--- a/spec/operators/throttleTime-spec.ts
+++ b/spec/operators/throttleTime-spec.ts
@@ -226,11 +226,11 @@ describe('throttleTime operator', () => {
 
     it('should emit the last throttled value when complete', () => {
       rxTest.run(({ hot, time, expectObservable, expectSubscriptions }) => {
-        const e1 = hot('  -a-xy-----b--x--cxx|');
-        const e1subs = '  ^------------------!';
-        const t = time('   ----|              ');
+        const e1 = hot('  -a-xy-----b--x--cxx-|');
+        const e1subs = '  ^-------------------!';
+        const t = time('   ----|               ');
         //                 ----|---|----|---|---|
-        const expected = '-----y--------x---x---|';
+        const expected = '-----y--------x---x-|';
 
         const result = e1.pipe(throttleTime(t, rxTest, { leading: false, trailing: true }));
 
--- a/spec/operators/throttleTime-spec.ts
+++ b/spec/operators/throttleTime-spec.ts
@@ -1,200 +1,254 @@
+/** @prettier */
 import { expect } from 'chai';
-import { hot, cold, expectObservable, expectSubscriptions, time } from '../helpers/marble-testing';
 import { throttleTime, take, map, mergeMap } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { of, concat, timer } from 'rxjs';
-
-declare const rxTestScheduler: TestScheduler;
+import { observableMatcher } from '../helpers/observableMatcher';
 
 /** @test {throttleTime} */
 describe('throttleTime operator', () => {
-  it('should immediately emit the first value in each time window', () => {
-    const e1 =   hot('-a-x-y----b---x-cx---|');
-    const subs =     '^                    !';
-    const expected = '-a--------b-----c----|';
+  let rxTest: TestScheduler;
 
-    const result = e1.pipe(throttleTime(50, rxTestScheduler));
-
-    expectObservable(result).toBe(expected);
-    expectSubscriptions(e1.subscriptions).toBe(subs);
+  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


spec/operators/throttleTime-spec.ts
```diff
@@ -1,200 +1,254 @@
...
-      const e1 =   hot('-a-xy-----b--x--cxx|');
-      const e1subs =   '^                  !';
-      const t =   time('----|               ');
-      const expected = '-----y--------x----(x|)';
+      rxTest.run(({ hot, time, expectObservable, expectSubscriptions }) => {
+        const e1 = hot('  -a-xy-----b--x--cxx|');
+        const e1subs = '  ^------------------!';
+        const t = time('   ----|              ');
+        //                 ----|---|----|---|---|
+        const expected = '-----y--------x---x---|';
```
cartant, line 233: question/nitpick: Should this really delay/throttle the completion of the source? This isn't the behaviour that I would have expected. Whatever the case, this test's description doesn't seem to describe what it's actually testing, IMO. If this is how it's supposed to behave, I think the description should explicitly state that complete notifications that are received whilst throttling will be delayed.

cartant (approved):
LGTM, but the behaviour surprised me a little. See comment below.

spec/operators/throttleTime-spec.ts
benlesh: Ah... yeah, I think if it has to trailing value we should probably complete it.. I'll make that change and merge.

PR #5688, opened 2020-09-02
fix(zip): zip operators and functions are now able to zip all iterable sources

### LINKED ISSUE #4304: zip does not do anything if passed all iterators
Ages ago, I had this amazing idea to give `zip` special handling for iterables, such that, as it was "zipping" it would attempt to only next out values from the iterable as needed.

Unfortunately, this has introduced a lot of complexity in that operator as well as a bug... 

```ts
zip('abc', 'def').subscribe(x => console.log); // this does nothing.
```

It should yield:

```
['a', 'd']
['b', 'e']
['c', 'f']
```

But it does nothing.

While it seemed like a clever idea at the time, I now think it deviates from the "push only" nature of RxJS, and this behavior should be removed in the upcoming v7.

benlesh:
Resolves an issue I outlined in #4304. Basically, once upon a time, I had a "brilliant" idea to have iterators iterated "as needed" during a zip operation. There are a few consequences to this, but the nastiest one is that you could not use zip with all-iterable sources. It also doesn't jive with the push-based nature of this library. This PR fixes all of that.

- Simplifies zip implementations
- Updates tests as appropriate

BREAKING CHANGE: `zip` operators will no longer iterate provided iterables "as needed", instead the iterables will be treated as push-streams just like they would be everywhere else in RxJS. This means that passing an endless iterable will result in the thread locking up, as it will endlessly try to read from that iterable. This puts us in-line with all other Rx implementations. To work around this, it is probably best to use `map` or some combination of `map` and `zip`. For example, `zip(source$, iterator)` could be `source$.pipe(map(value => [value, iterator.next().value]))`.

fixes #4304


cartant (approved):
LGTM

PR #5691, opened 2020-09-03
Update packages with vulnerabilities

benlesh:
Updates all packages that had vulnerabilities with the exception of `check-side-effects`. But I put in a PR for that here: https://github.com/filipesilva/check-side-effects/pull/6

So whenever @filipesilva gets around to reviewing/merging/correcting that, we should be able to update that one too.


benlesh:
Given what this is... I'm just going to merge it. Happy to roll it back if anyone has an issue with it.

PR #5699, opened 2020-09-07
Fix of incorrect condition

lopatnov:
**Description:**

Fix of incorrect condition.

```ts
let _parents: Subscription[] | null;
// ....
if ((_parents = this._parents)) // <-- So _parents = this._parents 
// It doesn't mean _parents === this._parents, It means that _parents doesn't have any sense. 
// The same for const { _singleParent } = this;. My suggestion that better to write this._singleParent.
// It will be shortly and clearly.
```

Doesn't exists

### DIFF
--- a/src/internal/Subscription.ts
+++ b/src/internal/Subscription.ts
@@ -170,15 +170,14 @@ export class Subscription implements SubscriptionLike {
    */
   private _addParent(parent: Subscription) {
     const { _singleParent } = this;
-    let _parents: Subscription[] | null;
     if (_singleParent) {
       // We already have one parent so we'll need to expand
       // to use an array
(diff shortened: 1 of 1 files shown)


cartant:
Thanks, but the author has done this deliberately - to reduce the number of property accesses via `this`. It's more important for this code to be efficient than readable.

lopatnov:
Why author didn't assign value on the 172 line, where _singleParent? So it will be more clear.

cartant:
> Why author didn't assign value on the 172 line, where _singleParent?

Because `this._parents` is not needed if `this._singleParent` is non-null, so the property is read only if/when it needs to be read.

Issue #4003, opened 2018-08-10
Ajax with PUT and CORS


Agraphie:
## Bug Report

**Current Behavior**
Ajax PUT is not sending the cookie in the header.

**Reproduction**
- Create a simple request and have a cookie in your store
- Try ajax PUT

```js
update(id, data) {
        return ajax({
            url: `/test`,
            method: "PUT",
            body: JSON.stringify(data),
            crossDomain: true,
            withCredentials: true
        })
    }
```

**Expected behavior**
When exchanging "PUT" with "POST" a cookie is sent. PUT should behave the same way.

**Environment**
- Runtime: Firefox 61.0.2
- RxJS version: 6.2.2

**Additional context/Screenshots**
I'm not sure if I'm doing something incorrectly here, but when using "POST" everything works as expected. A cookie is sent and I can use my IP. But when using "PUT" I can see that the header does not include any cookie. Any help is appreciated!

DCtheTall:
I'd like to take a crack at this, probably not over the long weekend but some time in the next week or so.

DCtheTall:
@benlesh I have a PR with the general implementation set up. I will get to unit testing it at some point in the next couple days. Let me know if this is what you had in mind 😄

PR #5702, opened 2020-09-08
feat(ajax.ts): send XSRF cookies in a custom header

DCtheTall:
**Description:**

Add the ability to send XSRF cookies in a custom header. This implementation is based off of [Axios' own implementation of the same feature](https://github.com/axios/axios/blob/4b3947aa59aaa3c0a6187ef20d1b9dddb9bbf066/lib/adapters/xhr.js#L113-L122).

This fixes [#4003](https://github.com/ReactiveX/rxjs/issues/4003).


src/internal/ajax/ajax.ts
```diff
@@ -295,6 +295,21 @@ export function fromAjax<T>(config: AjaxConfig): Observable<AjaxResponse<T>> {
       headers['x-requested-with'] = 'XMLHttpRequest';
     }
 
+    // Allow users to provide their XSRF cookie name and the name of a custom header to use to
+    // send the cookie. This
+    if (config.withCredentials && config.xsrfCookieName && config.xsrfHeaderName) {
+      try {
+        const match = document.cookie.match(new RegExp(`(^|;\\s*)(${name})=([^;]*)`));
```
benlesh, line 302: Instead of the try/catch here, we should check to see if `document` exists at all. No reason to throw really, we can probably just noop this if it happens to be a non-browser environment. As long as whatever we're doing aligns with contemporaries like axios, et al.

We can put a cookie-getter function together based off of the function [here](https://stackoverflow.com/a/25490531/135786)

And do something like:

```ts
const { withCredentials, xsrfCookieName, xsrfHeaderName } = config;
if (withCredentials && xsrfCookiename && xsrfHeaderName) {
  headers[xsrfHeadername] = getCookie(xsrfCookiename);
}
```

and below something like:

```ts
function getCookie(name: string): string {
  if (typeof document !== 'undefined') {
    return document.cookie.match(`(^|;)\\s*${name}\\s*=\\s*([^;]+)`)?.pop() ?? '';
  }
}
```

Then we can ditch the error handling below.

Otherwise, we'll be introducing a non-isometric behavior that actually causes errors at runtime. It's still not isometric, but at least it won't error and it's relatively understandable.

benlesh (commented):
I know this is just a draft, and it looks okay. Nonetheless, I'm leaving comments.

As much as possible, I would look to other ajax libraries like axios for guidance here. We should probably set the same defaults with regards to this.

In particular, you'll notice that [axios will check to see not only if `withCredentials` is `true`, but also to see if the URL is from the same origin](https://github.com/axios/axios/blob/4b3947aa59aaa3c0a6187ef20d1b9dddb9bbf066/lib/adapters/xhr.js#L115).. I think that's expected for this sort of functionality.

src/internal/ajax/ajax.ts
DCtheTall: > We can put a cookie-getter function together

Done.

> Then we can ditch the error handling

Actually your function would still throw a `ReferenceError` in Node. This can be avoided by optional chaining the `globalThis` object!

```javascript
globalThis?.document.cookie.match(...)?.pop() ?? '';
```

DCtheTall:
> I know this is just a draft, and it looks okay. Nonetheless, I'm leaving comments.

They're appreciated 😄 

> In particular, you'll notice that axios will check to see not only if withCredentials is true, but also to see if the URL is from the same origin. I think that's expected for this sort of functionality.

Done.

DCtheTall:
@benlesh I am working through unit tests, but if you get to this before they're up feel free to leave any comments.

src/internal/ajax/ajax.ts
DCtheTall: Another option is we just rely on `config.crossDomain` instead of determining if the request is the same origin at runtime, whichever you prefer sgtm.

DCtheTall: Turns out `globalThis` is not supported in Node 11, so in order to get all tests to pass we have to do some checking for `ReferenceError`s 😢

DCtheTall:
@benlesh a bit confused by the failing checks since they don't seem to have to do with this change?

benlesh:
The API guardian tests are to show if there's been any changes to the public API. which there were in this.

If you run `npm run api_guardian:update`, it will update the appropriate files.

DCtheTall:
> If you run npm run api_guardian:update, it will update the appropriate files.

Done.

I am not sure why the expected `throttleTime` operator type needed to change for this change to pass checks, is that expected?

DCtheTall:
@benlesh friendly ping

benlesh:
Sorry, big conference day today. I will get back to this soon.

src/internal/ajax/ajax.ts
```diff
@@ -295,6 +295,16 @@ export function fromAjax<T>(config: AjaxConfig): Observable<AjaxResponse<T>> {
       headers['x-requested-with'] = 'XMLHttpRequest';
     }
 
+    // Allow users to provide their XSRF cookie name and the name of a custom header to use to
+    // send the cookie.
+    const { withCredentials, xsrfCookieName, xsrfHeaderName } = config;
+    if ((withCredentials || !config.crossDomain) && xsrfCookieName && xsrfHeaderName) {
+      const xsrfCookie = readCookie(xsrfCookieName);
```
benlesh, line 302: I think we can inline the entire `readCookie` call here to save on size. And since we'd just be rethrowing the error anyhow, that means this would just be:

```ts
const { withCredentials, xsrfCookieName, crossDomain, xsrfHeaderName } = config;
((withCredentials || !crossDomain) && xsrfCookieName && xsrfHeaderName) && (headers[xsrfHeaderName] = document?.cookie?.match(new RegExp(`(^|;\\s*)(${name})=([^;]*)`))?.pop() ?? '');
```

I know that *looks* ridiculous, but we're trying to be size-sensitive with our library because so many people use it.

src/internal/ajax/ajax.ts
```diff
@@ -507,3 +517,14 @@ function isURLSearchParams(body: any): body is URLSearchParams {
 function isReadableStream(body: any): body is ReadableStream {
   return typeof ReadableStream !== 'undefined' && body instanceof ReadableStream;
 }
+
+function readCookie(name: string): string {
+  try {
+    return document?.cookie.match(new RegExp(`(^|;\\s*)(${name})=([^;]*)`))?.pop() ?? '';
+  } catch (err) {
+    if (err instanceof ReferenceError) {
```
benlesh, line 525: Further explanation of the in-lined stuff above. The `try/catch` isn't necessary because it's really just here to catch the `ReferenceError` and then rethrow. If we check `cookie` inline like `document?.cookie?.match`, then there's no possibility of a `ReferenceError` and the `try/catch` has no purpose (because we're just rethrowing).

benlesh (changes requested):

benlesh:
So so close. Just one last set of changes to help reduce the size impact of the feature.

src/internal/ajax/ajax.ts
DCtheTall: Done.

src/internal/ajax/ajax.ts
DCtheTall: Optional chaining defends against `TypeError`s that get thrown due to accessing properties of `null` or `undefined`.

In Node.js, the `document` variable is not even initialized, so accessing the variable `document` itself (even without accessing properties on that object!) throws a `ReferenceError` that optional chaining will not prevent.

That being said, I have thought about this some and decided that I am probably being overly-defensive. I think that it is reasonable to leave it to the user to only use `xsrf*Name` options in the DOM, and to allow the `ReferenceError` to throw if they try to use it in Node. If someone _really_ wants it in Node then the library can always change later.

DCtheTall:
@benlesh I addressed your comments and left a bit of explanation for my reasoning for the code I had before, hope this is what you had in mind 😄

DCtheTall:
@benlesh friendly ping

benlesh (approved):

benlesh:
Merged! Thanks for all of your hard work, @DCtheTall! We'll have this in the next 7 beta release!

🎉 🎉 🎉

src/internal/ajax/ajax.ts
benlesh: I see what you're getting at. Thank you for the explanation. I'll try to revisit this in another pass. I suspect there are some places where we need to just do a sanity check on references for environments like node and older runtimes. And I also suspect that there's a compact way to do this.

PR #5703, opened 2020-09-08
refactor(Observable): simplify subscription and subscriber code

benlesh:
- Gets rid of weirdness with syncErrorThrowable, et al
- General code golf in a few spots.

### DIFF
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -208,37 +208,25 @@ export class Observable<T> implements Subscribable<T> {
     error?: ((error: any) => void) | null,
     complete?: (() => void) | null
   ): Subscription {
-    const { operator } = this;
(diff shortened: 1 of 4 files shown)


src/internal/Observable.ts
```diff
@@ -506,15 +488,7 @@ export class Observable<T> implements Subscribable<T> {
...
-  if (!promiseCtor) {
-    promiseCtor = config.Promise || Promise;
-  }
-
-  if (!promiseCtor) {
-    throw new Error('no Promise impl found');
-  }
-
-  return promiseCtor;
+  return promiseCtor ?? config.Promise ?? Promise;
```
benlesh, line 491: This was changed to make it lighter weight.

The old code served no purpose that wouldn't be served by `ReferenceError: Promise is not defined`.

cartant (approved):
LGTM ❤

PR #5704, opened 2020-09-09
refactor(tap): slight size reduction

benlesh:
- Also ensures that unsubscribe is called in error and complete callbacks by calling `super._error` etc.
- Rewrites docs and adds more examples (LOC will look like more, probably)

Before: 913B

tap-before

After: 549B (60% of original size)

tap-after

### DIFF
--- a/src/internal/operators/tap.ts
+++ b/src/internal/operators/tap.ts
@@ -1,3 +1,4 @@
+/** @prettier */
 import { Operator } from '../Operator';
 import { Subscriber } from '../Subscriber';
 import { Observable } from '../Observable';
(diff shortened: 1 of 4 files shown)


src/internal/operators/tap.ts
```diff
@@ -18,131 +19,165 @@ export function tap<T>(observer: PartialObserver<T>): MonoTypeOperatorFunction<T
...
+       */
+      const wrap = (fn: any) => (arg?: any) => {
+        try {
+          fn(arg);
+        } catch (err) {
+          subscriber.error(err);
+        }
+      };
+      let onError = wrap(error ?? noop);
+      let onComplete = wrap(complete ?? noop);
```
cartant, line 132: question: Is there a reason the calls to `wrap` for `error` and `complete` are not inside the `if (!nextOrObserver || typeof nextOrObserver === 'function') { ... }` block? AFAICT, they will pointlessly wrap `noop`s when an observer is passed.

cartant (approved):
LGTM, but AFAICT some calls could/should be moved into a specific conditional branch.

src/internal/operators/tap.ts
benlesh: Probably playing too much code golf.

benlesh: Ironically, correcting these goofs resulted in an even smaller operator. Added to the PR. I'll just merge after this passes CI.

Now: 549B

tap-after

PR #5705, opened 2020-09-09
Add cast to configOrCount to fix types in Google

leggechr:
**Description:**
Without this case I kept the following compilation error: 

```
Type 'number | RetryConfig' is not assignable to type 'number'.
  Type 'RetryConfig' is not assignable to type 'number'.

85     config = {count: configOrCount};
                 ~~~~~

  third_party/javascript/rxjs/src/internal/operators/retry.ts:27:3
    27   count: number;
         ~~~~~
    The expected type comes from property 'count' which is declared here on type 'RetryConfig'

```

### DIFF
--- a/src/internal/operators/retry.ts
+++ b/src/internal/operators/retry.ts
@@ -65,7 +65,7 @@ export function retry<T>(configOrCount: number | RetryConfig = Infinity): MonoTy
     config = configOrCount;
   } else {
     config = {
-      count: configOrCount
+      count: configOrCount as number
     };
   }
   const { count, resetOnSuccess = false } = config;


benlesh:
That's super weird! TypeScript isn't inferring that properly for you? Does this happen with every usage of this in Google? I don't think this has changed since your last sync. Have you checked with the TS team?

tmair:
I found out what causes the issue. The error only occurs when the `strictNullChecks` compiler setting is set to `false`. 

When the condition is switched the compiler does not issue the error.

See also this minimal [example ]https://www.typescriptlang.org/play?strictNullChecks=false&target=7&ts=4.0.2#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgJWBigE8BhCJdBAczgG8AoOObCAVxQC44l2BbAEZoA3MzhRgAZyIB5JAGV22XFKkB+HoIgQANsExIxAX0aN0nbDASU4qKTACMAClaVqNWVAqcYPQsTk7rRwAD68AsJQAJQMZvrwbFS0-kSkFMk0YiwI6HDOMCRgwBB5SR5ePihwALx1cADkfEJoDbFMLCzlITUMbL483Z7eHCjG2XDGcMC6MnGdbpm1ixUjvhOm4pIw7FBIK7QmZuaW1rb2MABMrkOVo34EaUFL4c1R7fFEBzSpgRkeE1y+Vua2qADIwXYiiUysFhlV4HVeg0IIIAFbAKxteadIbLEEIjbTWZ4DoLPG9ej9bjfO6+cbiTYsba7fZDI5AA) in the TS Playground.

cartant:
The Google codebase has `strictNullChecks` set to `false`? 😮 If that is what's effecting the problem, do we really want to support the complication of RxJS source using TypeScript configurations that differ so radically?

leggechr:
Yes it is set to false. Let me see if I can change it to true. Otherwise I can add a patch that makes this change for Google.

cartant:
We spoke about this again at this week's core team meeting and we agreed that we do not want to support the _compilation_ of the RxJS source with options that differ from those that _we_ use to compile it. Specifically, we don't want to support the compilation of the RxJS source with `strictNullChecks` set to `false`.

PR #5707, opened 2020-09-09
refactor(Subscription): minor size reduction

benlesh:
- Reduces the size of subscription a little
- Moves back to a single property for single/many parents scenarios
- Updates the type of `Teardown` to not be so loose.  does not may any sense there.


cartant (approved):
LGTM. I can see that this restores one of Benedikt's perf changes. FWIW, I'm planning on taking a closer look at some perf-related aspects of the codebase - but not until after I've written the next week's talk. I built the V8 source, yesterday, so that I can use `d8`. It's been a long time since I last dealt with a C++ project that takes as long to build as that one does.

PR #5709, opened 2020-09-09
Refactor isArray usage

benlesh:
- Gets rid of legacy `isArray` implementation.
- Centralizes the common tasks involving arrays of arguments that `isArray` was being used for into a few useful utilities.
- Additional LOC caused by formatting with prettier, not actual code addition. Should be less code when bundled.

### DIFF
--- a/src/internal/observable/bindCallback.ts
+++ b/src/internal/observable/bindCallback.ts
@@ -1,7 +1,7 @@
 import { SchedulerLike } from '../types';
 import { Observable } from '../Observable';
-import { map } from '../operators/map';
 import { isScheduler } from '../util/isScheduler';
+import { mapOneOrManyArgs } from '../util/mapOneOrManyArgs';
 
 // tslint:disable:max-line-length
 /** @deprecated resultSelector is no longer supported, use a mapping function. */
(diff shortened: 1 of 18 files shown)


src/internal/observable/combineLatest.ts
```diff
@@ -237,41 +495,26 @@ export function combineLatest<T, K extends keyof T>(sourcesObject: T): Observabl
...
-      observables = first;
-    }
-    // if the first and only argument is an object, assume it's been called with
-    // `combineLatest({})`
-    if (isObject(first) && Object.getPrototypeOf(first) === Object.prototype) {
-      keys = Object.keys(first);
-      observables = keys.map(key => first[key]);
-    }
-  }
+  const { args: observables, keys } = argsArgArrayOrObject(args);
```
benlesh, line 511: Note, the actual change was here. The rest was formatting.

src/internal/observable/bindCallback.ts
```diff
@@ -173,18 +172,19 @@ export function bindCallback(callbackFunc: Function, scheduler?: SchedulerLike):
...
  */
-export function bindCallback<T>(
-  callbackFunc: Function,
-  resultSelector?: Function|SchedulerLike,
+export function bindCallback<T, R>(
+  callbackFunc: any,
+  resultSelector?: any,
   scheduler?: SchedulerLike
-): (...args: any[]) => Observable<T> {
+): (...args: any[]) => Observable<T | R> {
```
cartant, line 179: nitpick: If the `R` type parameter is used only in the return type, it serves no real purpose and it should be possible to replace it with `unknown` (in the return type), 'cause that's what it will be anyway. The actual return inference is made in the overload signature; not the implementation sig.

src/internal/util/argsArgArrayOrObject.ts
```diff
@@ -0,0 +1,32 @@
...
+const { isArray } = Array;
+const { getPrototypeOf, prototype: objectProto, keys: getKeys } = Object;
+
+/**
+ * Used in functions where either a list of arguments, a single array of arguments, or a
+ * dictionary of arguments can be returned. Returns an object with an `args` property with
+ * the arguments in an array, if it is a dictionary, it will also return the `keys` in another
+ * property.
+ */
+export function argsArgArrayOrObject<T, O extends { [key: string]: T }>(args: T[] | [O] | [T[]]): { args: T[]; keys: string[] | null } {
```
cartant, line 12: nitpick: I'd use `Record<string, T>` instead of `{ [key: string]: T }`. It's no big deal; it's just that I use the built-in types whenever possible.

src/internal/operators/raceWith.ts
```diff
@@ -23,14 +22,8 @@ export function race<T, R>(...observables: Array<Observable<any> | Array<Observa
...
  */
-export function race<T, R>(...observables: ObservableInput<R>[]): OperatorFunction<T, (T|R)[]> {
-  // if the only argument is an array, it was most likely called with
-  // `pair([obs1, obs2, ...])`
-  if (observables.length === 1 && isArray(observables[0])) {
-    observables = observables[0];
-  }
-
-  return raceWith(...observables) as any;
+export function race<T, R>(...args: any[]): OperatorFunction<T, R> {
```
cartant, line 25: nitpick: `R` could be replaced with `unknown`.

cartant (approved):
LGTM, just a bunch of nitpicks.

src/internal/util/argsArgArrayOrObject.ts
benlesh: Normally I do too, TBH... but in this case, I feel like it communicated a little more about the use, I don't know. I guess `Record<string, T>` is more terse. I think I'll change it.

PR #5711, opened 2020-09-10
refactor(combineLatest): remove stankyLift, make smaller

benlesh:
- Refactors one test to make it easier to debug
- Removes the use of `stankyLift` for `combineLatestWith`.
- Builds `combineAll` from `toArray` and `combineLatest`. I think this has the added benefit if making it more obvious what the operator does for those that read the source.
- Refactors implementation of static `combineLatest` to be smaller.

I'm hopeful that this refactor will serve as the basis for refactoring other operators with similar issues.

combineLatest static judged by including a build of this PR into a new angular app, built in prod mode and looked at in `source-map-explorer`:

Before: 1.21KB

PR-5711-before

After: 698B

PR-5711-after

### DIFF
--- a/src/internal/observable/combineLatest.ts
+++ b/src/internal/observable/combineLatest.ts
@@ -2,14 +2,13 @@
 import { Observable } from '../Observable';
 import { ObservableInput, SchedulerLike, ObservedValueOf, ObservedValueTupleFromArray } from '../types';
 import { isScheduler } from '../util/isScheduler';
-import { Subscriber } from '../Subscriber';
-import { ComplexOuterSubscriber, ComplexInnerSubscriber, innerSubscribe } from '../innerSubscribe';
-import { Operator } from '../Operator';
-import { fromArray } from './fromArray';
-import { lift } from '../util/lift';
 import { argsArgArrayOrObject } from '../util/argsArgArrayOrObject';
-
-const NONE = {};
+import { Subscriber } from '../Subscriber';
+import { isObject } from '../util/isObject';
+import { from } from './from';
+import { identity } from '../util/identity';
+import { map } from '../operators/map';
+import { Subscription } from '../Subscription';
 
 /* tslint:disable:max-line-length */
 
@@ -496,7 +495,7 @@ export function combineLatest<T, K extends keyof T>(sourcesObject: T): Observabl
  */
 export function combineLatest<O extends ObservableInput<any>, R>(
   ...args: (O | ((...values: ObservedValueOf<O>[]) => R) | SchedulerLike)[]
-): Observable<R> {
+): Observable<R> | Observable<ObservedValueOf<O>[]> {
   let resultSelector: ((...values: Array<any>) => R) | undefined = undefined;
   let scheduler: SchedulerLike | undefined = undefined;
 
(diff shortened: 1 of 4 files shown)


src/internal/operators/combineLatestWith.ts
```diff
@@ -80,13 +81,19 @@ export function combineLatest<T, TOther, R>(
 /**
  * @deprecated Deprecated, use {@link combineLatestWith} or static {@link combineLatest}
  */
-export function combineLatest<T, R>(...args: (ObservableInput<any> | ((...values: any[]) => R))[]): OperatorFunction<T, R> {
+export function combineLatest<T, R>(...args: (ObservableInput<any> | ((...values: any[]) => R))[]): OperatorFunction<T, unknown> {
```
cartant, line 84: nitpick: `R` is used in this signature's parameter types, so it could remain in the return type, but if using `unknown` instead of `R` solves some problem, I guess that's fine. The general rule is that type parameters [have to be used twice](https://effectivetypescript.com/2020/08/12/generics-golden-rule/) - to avoid their being pointless.

src/internal/operators/combineAll.ts
```diff
@@ -53,6 +56,10 @@ export function combineAll<R>(project: (...values: Array<any>) => R): OperatorFu
  * @return {Observable<T>}
  * @name combineAll
  */
-export function combineAll<T, R>(project?: (...values: Array<any>) => R): OperatorFunction<T, R> {
-  return (source: Observable<T>) => lift(source, new CombineLatestOperator(project, null));
+export function combineAll<T, R>(project?: (...values: Array<any>) => R): OperatorFunction<ObservableInput<T>, R|T[]> {
+  return (source: Observable<ObservableInput<T>>) => source.pipe(
+    toArray(),
+    concatMap((sources) => combineLatest(sources)),
+    project ? map((args) => args.length === 1 ? project(args) : project(...args)) : identity as any
```
cartant, line 63: question: can `mapOneOrManyArgs` be used here?

src/internal/observable/combineLatest.ts
```diff
@@ -510,87 +510,108 @@ export function combineLatest<O extends ObservableInput<any>, R>(
...
+      scheduler,
+      keys
+        ? // A handler for scrubbing the array of args into a dictionary.
+          (args: any[]) => {
+            const value: any = {};
+            for (let i = 0; i < args.length; i++) {
+              value[keys![i]] = args[i];
+            }
+            return value;
+          }
```
cartant, line 525: question: Will something like this be needed for `forkJoin`, too? Should it be extracted into another args-related util?

cartant (approved):
LGTM. Some questions and a nitpick.

src/internal/operators/combineLatestWith.ts
benlesh: It's `unknown` because it all depends on whether or not the user passed the deprecated resultSelector or not. And I'm too lazy to dance around the types and cast etc for this internal bit. I didn't think it added anything to do so.

src/internal/observable/combineLatest.ts
benlesh: Probably. I figured I'd cross the bridge when I came to it (after this is merged)

src/internal/operators/combineAll.ts
benlesh: Yup! Good catch.

PR #5712, opened 2020-09-10
refactor(bufferTime): reduce the size of the implementation

benlesh:
- Complete rewrite of `bufferTime`
- Adds some comments
- Formats a couple of tests for readability during debugging.

Before: 2.24KB

bufferTime-before

After: 823B

bufferTime-after

### DIFF
--- a/spec/operators/bufferTime-spec.ts
+++ b/spec/operators/bufferTime-spec.ts
@@ -53,6 +53,9 @@ describe('bufferTime operator', () => {
       const e1 = hot('  ---a---b---c---d---e---f---g-----|   ');
       const subs = '    ^--------------------------------!   ';
       const t = time('  ----------|                          ');
+      //                       ----------|
+      //                               ----------|
+      //                                       ----------|
       const expected = '-------w-------x-------y---------(z|)';
       const values = {
         w: ['a', 'b'],
@@ -73,6 +76,7 @@ describe('bufferTime operator', () => {
       const e1 = hot('  ---a---b---c---d---e---f---g-----|   ');
       const subs = '    ^--------------------------------!   ';
       const t = time('  ----------|                          ');
+      //                          ---------|---------|---------|
       const expected = '----------w--------x---------y---(z|)';
       const values = {
         w: ['a', 'b'],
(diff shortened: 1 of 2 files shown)


spec/operators/bufferTime-spec.ts
```diff
@@ -53,6 +53,9 @@ describe('bufferTime operator', () => {
       const e1 = hot('  ---a---b---c---d---e---f---g-----|   ');
       const subs = '    ^--------------------------------!   ';
       const t = time('  ----------|                          ');
+      //                       ----------|
+      //                               ----------|
+      //                                       ----------|
```
kolodny, line 58: What are these comments for?

src/internal/operators/bufferTime.ts
```diff
@@ -69,189 +79,128 @@ export function bufferTime<T>(bufferTimeSpan: number, bufferCreationInterval: nu
...
-    bufferCreationInterval = arguments[1];
+  if (isScheduler(otherArgs[otherArgs.length - 1])) {
+    scheduler = otherArgs.pop() as SchedulerLike;
   }
 
-  let maxBufferSize: number = Infinity;
-  if (length >= 3) {
-    maxBufferSize = arguments[2];
-  }
+  const bufferCreationInterval = (otherArgs[0] as number) || null;
```
kolodny, line 89: Can we use `??` in this case? If `otherArgs[0]` is 0 then this becomes null. Not a big deal in this case but can bite us when we miss it

src/internal/operators/bufferTime.ts
```diff
@@ -69,189 +79,128 @@ export function bufferTime<T>(bufferTimeSpan: number, bufferCreationInterval: nu
...
+  const maxBufferSize = (otherArgs[1] as number) || Infinity;
```
kolodny, line 90: This is breaking since before if `length >= 3` it would be set to `0` but here it's set to `Infinity` in that case. `??` may help here

kolodny (approved):
LGTM with some comments. I'd probably lean towards getting another LG (your call) since this is a major refactor but overall looks great.

src/internal/operators/bufferTime.ts
benlesh: I don't quite follow. [It appears that we were setting it to `Infinity` by default](https://github.com/ReactiveX/rxjs/blob/6416935ab03b38d7048d0641c8826ba5311eae8c/src/internal/operators/bufferTime.ts#L86-L89), and pulling the number out of the arguments if it was there. This is equivalent, AFAICT.

src/internal/operators/bufferTime.ts
benlesh: Changed that, and also added a test around scheduling creation intervals at zero -- like some sort of psychopath.

spec/operators/bufferTime-spec.ts
benlesh: They were just to help me visualize/estimate where values should be. A sanity test for the marbles in that test.

PR #5714, opened 2020-09-10
refactor(throttleTime): reduce the size of the implementation

benlesh:
- Just a rewrite of the implementation and some comments.

Before: 1.11KB

throttleTime-before

After: 472B

throttleTime-after

### DIFF
--- a/src/internal/operators/throttleTime.ts
+++ b/src/internal/operators/throttleTime.ts
@@ -90,105 +90,113 @@ export function throttleTime<T>(
   scheduler: SchedulerLike = asyncScheduler,
   config: ThrottleConfig = defaultThrottleConfig
 ): MonoTypeOperatorFunction<T> {
-  return (source: Observable<T>) => lift(source, new ThrottleTimeOperator(duration, scheduler, !!config.leading, !!config.trailing));
-}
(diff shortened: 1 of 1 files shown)


src/internal/operators/throttleTime.ts
```diff
@@ -90,105 +89,113 @@ export function throttleTime<T>(
...
-        // we need to set the trailing value so once the throttle
-        // is done we can emit it. We don't do this if we're leading
-        // as well, because that would result in a double emission.
-        this.trailingValue = value;
-        this.hasTrailingValue = true;
-      }
+  return (source: Observable<T>) =>
+    lift(source, function (this: Subscriber<T>, source: Observable<T>) {
+      const subscriber = this;
+      const { leading = false, trailing = false } = config;
```
cartant, line 95: nitpick: `const { leading = false, trailing = false } = config;` could be moved to the up to the enclosing scope - i.e. above the `return (source: ...) =>` - as the config only needs to be read once.

cartant (approved):
LGTM. Just one nitpick.

src/internal/operators/throttleTime.ts
benlesh: Good catch

PR #5715, opened 2020-09-10
refactor(mergeMap): reduce the size of the implementation

benlesh:
- Refactor with comments

Before: 1.1KB

mergeMap-before

After: 639B (58% of the original size)

mergeMap-after

### DIFF
--- a/src/internal/operators/mergeMap.ts
+++ b/src/internal/operators/mergeMap.ts
@@ -1,3 +1,4 @@
+/** @prettier */
 import { Observable } from '../Observable';
 import { Operator } from '../Operator';
 import { Subscriber } from '../Subscriber';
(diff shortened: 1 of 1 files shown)


src/internal/operators/mergeMap.ts
```diff
@@ -74,93 +86,125 @@ export function mergeMap<T, R, O extends ObservableInput<any>>(
...
+            }
+            // Be sure to teardown the outer subscription ASAP, in any case.
+            outerSubs?.unsubscribe();
+          }
+        )
+      );
+    });
 }
 
+// TODO(benlesh): This may end up being so common that we can centralize on one Subscriber for a few operators.
```
cartant, line 196: Yes. This is one of the perf-related issues I'll be looking at in a week or so. If we can eliminate derived subscribers, a whole class of performance issues will simply disappear.

cartant (approved):
LGTM.

PR #5716, opened 2020-09-11
refactor(switchMap): reduce the size of the implementation

benlesh:
- Refactored and commented.
- It's a little known fact that switchMap is really about a map for finding witches whose names start with "S".

Before: 875B

switchMap-before

After: 511B (58% of the original size)

switchMap-after

### DIFF
--- a/src/internal/operators/switchMap.ts
+++ b/src/internal/operators/switchMap.ts
@@ -1,3 +1,4 @@
+/** @prettier */
 import { Operator } from '../Operator';
 import { Observable } from '../Observable';
 import { Subscriber } from '../Subscriber';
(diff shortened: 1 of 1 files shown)


src/internal/operators/switchMap.ts
```diff
@@ -9,11 +10,19 @@ import { lift } from '../util/lift';
 import { SimpleInnerSubscriber, innerSubscribe, SimpleOuterSubscriber } from '../innerSubscribe';
 
 /* tslint:disable:max-line-length */
```
un33k, line 12: I think this PR has addressed the max-length now, so the tslint max-line can be removed.

benlesh:
Closing this in favor of #5729

PR #5719, opened 2020-09-11
fix(delay): Now properly handles Date and negative numbers

benlesh:
- Resolves an issue where passing a Date would cause every value to be scheduled incorrectly, instead of delaying all values until the absolute date -- as documented.
- Resolves an issue where negative numbers were treated as though they were positive numbers
- Reduces operator sizes
- Adds comments and formatting
- Updates tests to reflect better behavior
- Adds test for Dates in the past.
- `delay` now based on `delayWhen` and `timer`.

### DIFF
--- a/spec/operators/delay-spec.ts
+++ b/spec/operators/delay-spec.ts
@@ -6,7 +6,7 @@ import { expect } from 'chai';
 import { observableMatcher } from '../helpers/observableMatcher';
 
 /** @test {delay} */
-describe('delay operator', () => {
+describe('delay', () => {
   let testScheduler: TestScheduler;
 
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


benlesh:
`timer` looks like it needs a little work, but I'll put that in another PR.

spec/operators/delay-spec.ts
```diff
@@ -27,12 +27,26 @@ describe('delay operator', () => {
     });
   });
 
+  it('should not delay at all if the delay number is negative', () => {
+    testScheduler.run(({ hot, expectObservable, expectSubscriptions }) => {
+      const e1 =   hot('---a--b--|');
+      const t = -1; //      --|
```
cartant, line 33: nitpick: The comment serves no purpose.

cartant (approved):
LGTM, just a really minor nitpick.

PR #5720, opened 2020-09-11
refactor(map): reduce the size of the implementation

benlesh:
- Refactor and add comments

Before: 525B

map-before

After: 345B (65% of original size)

map-after

BREAKING CHANGE: `thisArg` will now default to `undefined`. The previous default of `MapSubscriber` never made any sense. This will only affect code that calls map with a `function` and references `this` like so: `source.pipe(map(function () { console.log(this); }))`. There wasn't anything useful about doing this, so the breakage is expected to be very minimal. If anything we're no longer leaking an implementation detail.

### DIFF
--- a/src/internal/operators/map.ts
+++ b/src/internal/operators/map.ts
@@ -1,4 +1,3 @@
-import { Operator } from '../Operator';
 import { Subscriber } from '../Subscriber';
 import { Observable } from '../Observable';
 import { OperatorFunction } from '../types';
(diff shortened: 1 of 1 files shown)


kwonoj (approved):

src/internal/operators/map.ts
```diff
@@ -43,49 +42,39 @@ import { lift } from '../util/lift';
  * @name map
  */
 export function map<T, R>(project: (value: T, index: number) => R, thisArg?: any): OperatorFunction<T, R> {
+
   return function mapOperation(source: Observable<T>): Observable<R> {
     if (typeof project !== 'function') {
       throw new TypeError('argument is not a function. Are you looking for `mapTo()`?');
     }
```
SanderElias, line 49: Question. I assume this is reoccurring construct. Would an assaerFunction utility function not be easier?
```
  function asserFunction(target,errorMsg=`"${target} is not a function") {
    if (typeof target !== 'function') {
       throw new typeError(errorMsg)
   }
}
```
This will however the overall size. The downside, the call stack is a bit deeper, but even that is fixable.

cartant (approved):

src/internal/operators/map.ts
benlesh: In general, I think we need to make a pass and either decide to check all of these things at runtime, or walk away from doing that and let TypeScript do its thing and let bad calls just fail to reduce bundle size. The problem is these runtime assertions are "death by 1,000 cuts" with a library like this, but on the other hand, non-TypeScript users may need them.

cartant: > ... or walk away from doing that and let TypeScript do its thing.

That would get my vote.

SanderElias: TS can't catch a thing runtime. I would say, give the dev a useful error during runtime too. 
Also, a lot of TS users will slap `:any` on anything they don't understand, nullifying all compile-time safety.
So, it's not only JS users you are serving with this. it might prevent devs from turning away from RX because they don't understand why it's erroring out.
(I'm pretty sure if you don't catch this, the error would be less then helpful)

benlesh: The scale of this is the problem, though. We'll be forcing billions of bytes over the wire for all RxJS users to accommodate developers with nicer errors when they skip type safety.

SanderElias: I do realize that. But not all RxJs consumers are using TS.  Removing those assertions will add to their pain.
I see where you are coming from, and I get the point. My personal preference would really be to handle it during runtime. 
But, luckily, it is not up to me to make this decision. It's a hard one.

PR #5722, opened 2020-09-11
fix(scan): proper indexes when seed is not supplied

### LINKED ISSUE #3879: scan() definition

### LINKED ISSUE #4348: `scan` operator description might be confusing when not using any seed

benlesh:
- Smaller implementation
- Improved docs
- Adds comments
- Adds a test
- Fixes weird adjustment in `reduce`.

Size differences...

Before: 676 B

scan-before

After: 248 B (36% of original size)

scan-after

Documentation updates close a couple of issues:

Closes #4348
Closes #3879

### DIFF
--- a/spec/operators/scan-spec.ts
+++ b/spec/operators/scan-spec.ts
@@ -42,6 +42,16 @@ describe('scan operator', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
+  it('should provide the proper index if seed is skipped', () => {
+    const expected = [1, 2];
+    of(3, 3, 3).pipe(
+      scan((_: any, __, i) => {
+        expect(i).to.equal(expected.shift());
+        return null;
+      })
+    ).subscribe();
+  });
+
   it('should scan with a seed of undefined', () => {
     const e1 = hot('--a--^--b--c--d--e--f--g--|');
     const e1subs =      '^                    !';
--- a/src/internal/operators/reduce.ts
+++ b/src/internal/operators/reduce.ts
@@ -78,7 +78,7 @@ export function reduce<V, A>(accumulator: (acc: V | A, value: V, index: number) 
   }
   return function reduceOperatorFunction(source: Observable<V>): Observable<V | A> {
     return pipe(
-      scan<V, V | A>((acc, value, index) => accumulator(acc, value, index + 1)),
+      scan<V, V | A>((acc, value, index) => accumulator(acc, value, index)),
       takeLast(1),
     )(source);
   };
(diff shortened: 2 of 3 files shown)


benlesh:
Closing in favor of #5729 (merged this into that PR)

benlesh:
blocked on #5729 (I'm not sure I merged that work in there, I'll have to double check when the dust settles.)

PR #5723, opened 2020-09-11
feat(fromEvent): support lazy event targets

troch:
**Description:**

Hello 👋, first thanks for this great library that I've been using for some time now!

Like other observables, `fromEvent` will lazily consume resources when an observer will subscribe to it. However in the case of DOM elements, its current API requires the target to exist when creating an observable, despite a target only been required at subscription time. With this PR I propose to support event target getters, so that event targets can be retrieved lazily at the time of subscription.

```ts
const clickEvents$ = fromEvent(() => document.getElementById('id'), 'click')
```

Or even in React:

```ts
const ref = useRef()
const clickEvents$ = useMemo(
  () => fromEvent(() => ref.current, 'click'),
  []
)
```

Note that it is possible to achieve with `fromEventPattern`, but it is more verbose.

### DIFF
--- a/spec/observables/fromEvent-spec.ts
+++ b/spec/observables/fromEvent-spec.ts
@@ -392,4 +392,28 @@ describe('fromEvent', () => {
     }).to.not.throw(TypeError);
   });
 
+  it('should allow to subscribe to lazy targets', () => {
+    let onHandler;
+    let offHandler;
+
+    const event$ = fromEvent(() => target as any, 'click')
+    const target = {
+      addEventListener: (a: string, b: EventListener) => {
+        onHandler = b;
+      },
+      removeEventListener: (a: string, b: EventListener) => {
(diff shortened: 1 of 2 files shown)


benlesh:
Thank you for the PR, @troch! So you know, this is exactly the sort of thing that `defer` is designed for:

```ts
import { defer, fromEvent } from 'rxjs';

const clickEvents$ = defer(() => fromEvent(document.getElementById('id'), 'click'));
```

`defer` is very lightweight, and is useful in a number of other scenarios. As such, I don't think we need this feature at this time. Happy to reopen for discussion if others disagree.

Thank you again for your idea and your contribution! I hope that this doesn't put you off from contributing further.

troch:
@benlesh Thanks for your reply. I didn't know about `defer`, which is actually a lot more useful and generic!

PR #5727, opened 2020-09-13
refactor(groupBy): reduce the size of the implementation

benlesh:
- Refactor to make it smaller
- Replaced a faulty test with a test that tests the same thing. Verified the behavior has not changed since 6.x.

BEFORE: 1.99KB

groupBy-before

AFTER: 1012B (50% of the original size)

groupBy-after

### DIFF
--- a/spec/operators/groupBy-spec.ts
+++ b/spec/operators/groupBy-spec.ts
@@ -1,7 +1,7 @@
 import { expect } from 'chai';
-import { groupBy, delay, tap, map, take, mergeMap, materialize, skip } from 'rxjs/operators';
+import { groupBy, delay, tap, map, take, mergeMap, materialize, skip, ignoreElements } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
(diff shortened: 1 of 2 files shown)


spec/operators/groupBy-spec.ts
```diff
@@ -1371,54 +1371,35 @@ describe('groupBy operator', () => {
...
-
-        rxTestScheduler.schedule(() => {
-          group.pipe(
-            phonyMarbelize()
-          ).subscribe((value: any) => {
-            arr.push(value);
-          });
-        }, innerSubscriptionFrame - rxTestScheduler.frame);
-
-        return arr;
```
benlesh, line 1415: Note, this bit here caused the test to be faulty. It was "asynchronous" because of the scheduling above... but because of the faux-nature of the asynchronicity, was able to get an array values out here, which it should not have been able to do under "real scenarios".

I tested this with real code in a stackblitz and confirmed what the actual behavior was in 6.6.2 and again locally with the new code just to be sure. Then I refactored the test to be something that works a bit better, with comments.

benlesh:
Closing in favor of #5729 ... (Actually merging these commits into that one then working on them some more)

PR #5729, opened 2020-09-13
WIP: Refactor/smallify all the things

benlesh:
Going through, operator by operator, and making them smaller. Centralizing on a new type called `OperatorSubscriber` that enables most behaviors fairly cleanly and concisely.
WIP

Will show stats when finished.

### Progress Update 9/17

BEFORE: 93KB

image

~~AFTER: 59KB (63% of the original size)~~
AFTER: 57KB (61% of the original size)

image

### Review Checklist:

Reviewed files:

_Check only files that have been reviewed and have no issues. If there are issues with a review, add a link to the review to the listed file._

- [x] api_guard/dist/types/index.d.ts
- [x] api_guard/dist/types/operators/index.d.ts
- [x] spec/observables/IteratorObservable-spec.ts
- [x] spec/observables/fromEvent-spec.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-489350297
- [x] spec/operators/debounce-spec.ts
- [x] spec/operators/delay-spec.ts
- [x] spec/operators/expand-spec.ts
- [x] spec/operators/find-spec.ts
- [x] spec/operators/groupBy-spec.ts
- [x] spec/operators/map-spec.ts
- [x] spec/operators/observeOn-spec.ts
- [x] spec/operators/scan-spec.ts
- [x] spec/operators/skipLast-spec.ts
- [x] spec/operators/take-spec.ts
- [x] spec/operators/takeLast-spec.ts
- [x] spec/operators/takeWhile-spec.ts
- [x] spec/operators/windowTime-spec.ts
- [x] spec/subjects/AsyncSubject-spec.ts
- [x] src/internal/AsyncSubject.ts
- [x] src/internal/BehaviorSubject.ts
- [x] src/internal/Notification.ts
- [x] src/internal/Observable.ts
- [x] src/internal/ReplaySubject.ts
- [x] src/internal/Subject.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493970267
- [x] src/internal/SubjectSubscription.ts
- [x] src/internal/Subscriber.ts
- [x] src/internal/Subscription.ts
- [x] src/internal/innerSubscribe.ts
- [x] src/internal/observable/ConnectableObservable.ts
- [x] src/internal/observable/concat.ts
- [x] src/internal/observable/from.ts
- [x] src/internal/observable/fromEvent.ts
- [x] src/internal/observable/fromIterable.ts
- [x] src/internal/observable/fromObservable.ts
- [x] src/internal/observable/fromPromise.ts
- [x] src/internal/observable/generate.ts
- [x] src/internal/observable/merge.ts
- [x] src/internal/observable/partition.ts
- [x] src/internal/observable/race.ts
- [x] src/internal/operators/OperatorSubscriber.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-492197181
- [x] src/internal/operators/audit.ts
- [x] src/internal/operators/buffer.ts
- [x] src/internal/operators/bufferCount.ts
- [x] src/internal/operators/bufferTime.ts
- [x] src/internal/operators/bufferToggle.ts
- [x] src/internal/operators/bufferWhen.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-492171889
- [x] src/internal/operators/catchError.ts
- [x] src/internal/operators/concat.ts
- [x] src/internal/operators/concatWith.ts
- [x] src/internal/operators/count.ts
- [x] src/internal/operators/debounce.ts
- [x] src/internal/operators/debounceTime.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493388673
- [x] src/internal/operators/defaultIfEmpty.ts
- [x] src/internal/operators/delay.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493392532
- [x] src/internal/operators/delayWhen.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493394509
- [x] src/internal/operators/dematerialize.ts
- [x] src/internal/operators/distinct.ts
- [x] src/internal/operators/distinctUntilChanged.ts
- [x] src/internal/operators/every.ts
- [x] src/internal/operators/exhaust.ts
- [x] src/internal/operators/exhaustMap.ts
- [x] src/internal/operators/expand.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493188990
- [x] src/internal/operators/filter.ts
- [x] src/internal/operators/finalize.ts
- [x] src/internal/operators/find.ts
- [x] src/internal/operators/findIndex.ts
- [x] src/internal/operators/groupBy.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493844452
- [x] src/internal/operators/ignoreElements.ts
- [x] src/internal/operators/index.ts
- [x] src/internal/operators/isEmpty.ts
- [x] src/internal/operators/map.ts
- [x] src/internal/operators/mapTo.ts
- [x] src/internal/operators/materialize.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-492183688
- [x] src/internal/operators/mergeMap.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493162881
- [x] src/internal/operators/mergeScan.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493413986
- [x] src/internal/operators/mergeWith.ts
- [x] src/internal/operators/multicast.ts
- [x] src/internal/operators/observeOn.ts
- [x] src/internal/operators/pairwise.ts
- [x] src/internal/operators/partition.ts
- [x] src/internal/operators/raceWith.ts
- [x] src/internal/operators/reduce.ts
- [x] src/internal/operators/refCount.ts
- [x] src/internal/operators/repeat.ts
- [x] src/internal/operators/repeatWhen.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493828658
- [x] src/internal/operators/retryWhen.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493831163
- [x] src/internal/operators/sample.ts
- [x] src/internal/operators/sampleTime.ts
- [x] src/internal/operators/scan.ts
- [x] src/internal/operators/sequenceEqual.ts
- [x] src/internal/operators/single.ts
- [x] src/internal/operators/skip.ts
- [x] src/internal/operators/skipLast.ts
- [x] src/internal/operators/skipUntil.ts
- [x] src/internal/operators/skipWhile.ts
- [x] src/internal/operators/subscribeOn.ts
- [x] src/internal/operators/switchMap.ts
- [x] src/internal/operators/take.ts
- [x] src/internal/operators/takeLast.ts
- [x] src/internal/operators/takeUntil.ts
- [x] src/internal/operators/takeWhile.ts
- [x] src/internal/operators/throttle.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-492359092
- [x] src/internal/operators/throttleTime.ts
- [x] src/internal/operators/throwIfEmpty.ts
- [x] src/internal/operators/timeout.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493087683
- [x] src/internal/operators/window.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-492308225
- [x] src/internal/operators/windowCount.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493089050
- [x] src/internal/operators/windowTime.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493089648
- [x] src/internal/operators/windowToggle.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493091769
- [x] src/internal/operators/windowWhen.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-493092340
- [x] src/internal/operators/withLatestFrom.ts https://github.com/ReactiveX/rxjs/pull/5729#pullrequestreview-492316214
- [x] src/internal/scheduler/AsyncAction.ts
- [x] src/internal/util/arrRemove.ts
- [x] src/internal/util/lift.ts
- [x] src/internal/util/not.ts
- [x] src/internal/util/subscribeTo.ts
- [x] src/operators/index.ts


cartant (commented):
The following changed files LGTM:

```
api_guard/dist/types/index.d.ts
api_guard/dist/types/operators/index.d.ts
spec/operators/debounce-spec.ts
spec/operators/delay-spec.ts
spec/operators/expand-spec.ts
spec/operators/find-spec.ts
spec/operators/groupBy-spec.ts
spec/operators/map-spec.ts
spec/operators/observeOn-spec.ts
spec/operators/skipLast-spec.ts
spec/operators/take-spec.ts
spec/operators/takeWhile-spec.ts
spec/operators/windowTime-spec.ts
src/internal/Subscriber.ts
src/internal/operators/OperatorSubscriber.ts
```

I'll work through the reviews of the operator changes gradually.

cartant:
Reviewed files:

- [ ] api_guard/dist/types/index.d.ts
- [ ] api_guard/dist/types/operators/index.d.ts
- [ ] spec/observables/IteratorObservable-spec.ts
- [ ] spec/observables/fromEvent-spec.ts
- [ ] spec/operators/debounce-spec.ts
- [ ] spec/operators/delay-spec.ts
- [ ] spec/operators/expand-spec.ts
- [ ] spec/operators/find-spec.ts
- [ ] spec/operators/groupBy-spec.ts
- [ ] spec/operators/map-spec.ts
- [ ] spec/operators/observeOn-spec.ts
- [ ] spec/operators/scan-spec.ts
- [ ] spec/operators/skipLast-spec.ts
- [ ] spec/operators/take-spec.ts
- [ ] spec/operators/takeLast-spec.ts
- [ ] spec/operators/takeWhile-spec.ts
- [ ] spec/operators/windowTime-spec.ts
- [ ] src/internal/Notification.ts
- [ ] src/internal/Subject.ts
- [ ] src/internal/Subscriber.ts
- [ ] src/internal/Subscription.ts
- [ ] src/internal/innerSubscribe.ts
- [ ] src/internal/observable/ConnectableObservable.ts
- [ ] src/internal/observable/concat.ts
- [ ] src/internal/observable/from.ts
- [ ] src/internal/observable/fromEvent.ts
- [ ] src/internal/observable/fromIterable.ts
- [ ] src/internal/observable/fromObservable.ts
- [ ] src/internal/observable/fromPromise.ts
- [ ] src/internal/observable/generate.ts
- [ ] src/internal/observable/merge.ts
- [ ] src/internal/observable/partition.ts
- [ ] src/internal/observable/race.ts
- [ ] src/internal/operators/OperatorSubscriber.ts
- [ ] src/internal/operators/audit.ts
- [ ] src/internal/operators/buffer.ts
- [ ] src/internal/operators/bufferCount.ts
- [ ] src/internal/operators/bufferTime.ts
- [ ] src/internal/operators/bufferToggle.ts
- [ ] src/internal/operators/bufferWhen.ts
- [ ] src/internal/operators/catchError.ts
- [ ] src/internal/operators/concat.ts
- [ ] src/internal/operators/concatWith.ts
- [ ] src/internal/operators/count.ts
- [ ] src/internal/operators/debounce.ts
- [ ] src/internal/operators/debounceTime.ts
- [ ] src/internal/operators/defaultIfEmpty.ts
- [ ] src/internal/operators/delay.ts
- [ ] src/internal/operators/delayWhen.ts
- [ ] src/internal/operators/dematerialize.ts
- [ ] src/internal/operators/distinct.ts
- [ ] src/internal/operators/distinctUntilChanged.ts
- [ ] src/internal/operators/every.ts
- [ ] src/internal/operators/exhaust.ts
- [ ] src/internal/operators/exhaustMap.ts
- [ ] src/internal/operators/expand.ts
- [ ] src/internal/operators/filter.ts
- [ ] src/internal/operators/finalize.ts
- [ ] src/internal/operators/find.ts
- [ ] src/internal/operators/findIndex.ts
- [ ] src/internal/operators/groupBy.ts
- [ ] src/internal/operators/ignoreElements.ts
- [ ] src/internal/operators/index.ts
- [ ] src/internal/operators/isEmpty.ts
- [ ] src/internal/operators/map.ts
- [ ] src/internal/operators/mapTo.ts
- [ ] src/internal/operators/materialize.ts
- [ ] src/internal/operators/mergeMap.ts
- [ ] src/internal/operators/mergeScan.ts
- [ ] src/internal/operators/mergeWith.ts
- [ ] src/internal/operators/multicast.ts
- [ ] src/internal/operators/observeOn.ts
- [ ] src/internal/operators/pairwise.ts
- [ ] src/internal/operators/raceWith.ts
- [ ] src/internal/operators/reduce.ts
- [ ] src/internal/operators/refCount.ts
- [ ] src/internal/operators/repeat.ts
- [ ] src/internal/operators/repeatWhen.ts
- [ ] src/internal/operators/retryWhen.ts
- [ ] src/internal/operators/sample.ts
- [ ] src/internal/operators/sampleTime.ts
- [ ] src/internal/operators/scan.ts
- [ ] src/internal/operators/sequenceEqual.ts
- [ ] src/internal/operators/single.ts
- [ ] src/internal/operators/skip.ts
- [ ] src/internal/operators/skipLast.ts
- [ ] src/internal/operators/skipUntil.ts
- [ ] src/internal/operators/skipWhile.ts
- [ ] src/internal/operators/subscribeOn.ts
- [ ] src/internal/operators/switchMap.ts
- [ ] src/internal/operators/take.ts
- [ ] src/internal/operators/takeLast.ts
- [ ] src/internal/operators/takeUntil.ts
- [ ] src/internal/operators/takeWhile.ts
- [ ] src/internal/operators/throttle.ts
- [ ] src/internal/operators/throttleTime.ts
- [ ] src/internal/operators/throwIfEmpty.ts
- [ ] src/internal/operators/timeout.ts
- [ ] src/internal/operators/window.ts
- [ ] src/internal/operators/windowCount.ts
- [ ] src/internal/operators/windowTime.ts
- [ ] src/internal/operators/windowToggle.ts
- [ ] src/internal/operators/windowWhen.ts
- [ ] src/internal/operators/withLatestFrom.ts
- [ ] src/internal/util/lift.ts
- [ ] src/internal/util/subscribeTo.ts
- [ ] src/operators/index.ts

spec/observables/fromEvent-spec.ts
```diff
@@ -3,6 +3,7 @@ import { expectObservable } from '../helpers/marble-testing';
 import { fromEvent, NEVER, timer } from 'rxjs';
 import { mapTo, take, concat } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
+import sinon = require('sinon');
```
43081j, line 6: you should be able to do a regular import these days with sinon FYI (`import * as sinon`)

src/internal/Observable.ts
```diff
@@ -500,13 +494,9 @@ function getPromiseCtor(promiseCtor: PromiseConstructorLike | undefined) {
...
   while (subscriber) {
     const { closed, destination, isStopped } = subscriber as any;
-    if (closed || isStopped) {
-      return false;
-    } else if (destination && destination instanceof Subscriber) {
-      subscriber = destination;
-    } else {
-      subscriber = null!;
-    }
+    return closed || isStopped
```
JoostK, line 497: This return changes makes the `while` obsolete; I'm wondering if this change is correct as it changes the semantics?

I think it could be this:

```ts
for (;;) {
  const { closed, destination, isStopped } = subscriber as any;
  if (closed || isStopped) {
    return false;
  }
  if (destination instanceof Subscriber) {
    subscriber = destination;
  } else {
    break;
  }
}
```

benlesh: Well, the guy that did that was tired. 😛 

Haha... thanks for catching that @JoostK .. although it's disconcerting that wasn't found in tests. Updated.

spec/observables/fromEvent-spec.ts
benlesh: Ironically, the import wasn't even being used.

src/internal/operators/bufferToggle.ts
```diff
@@ -56,123 +58,56 @@ export function bufferToggle<T, O>(
...
+              arrRemove(buffers, buffer);
+              subscriber.next(buffer);
+              closingSubscription.unsubscribe();
+            };
+
+            // The line below will add the subscription to the parent subscriber *and* the closing subscription.
+            closingSubscription.add(from(closingSelector(openValue)).subscribe(new OperatorSubscriber(subscriber, emit, undefined, emit)));
+          },
+          undefined,
+          noop
```
cartant, line 88: nitpick/question: why is a `noop` needed here? The `onError` and `onComplete` parameter are optional, AFAICT, so why not omit them?

cartant (commented):
`bufferToggle` LGTM. <s>Just a nitpick/question.</s>

src/internal/operators/bufferWhen.ts
```diff
@@ -48,89 +47,53 @@ import { SimpleOuterSubscriber, innerSubscribe, SimpleInnerSubscriber } from '..
...
+          return;
+        }
 
-    const buffer = this.buffer;
-    if (this.buffer) {
-      this.destination.next(buffer);
-    }
+        closingNotifier.subscribe(
+          (closingSubscriber = new OperatorSubscriber(subscriber, openBuffer, undefined, () => {
+            isComplete ? subscriber.complete() : openBuffer();
```
cartant, line 75: issue: The `subscriber.complete` call here seems unnecessary. `isComplete` is only set to `true` in the other `OperatorSubscriber` and immediately after that `subscriber.complete` is called. If there's a reason for this, maybe it warrants a comment?

cartant (commented):
`bufferWhen` LGTM, but there's either an unnecessary condition in it or a comment is warranted, IMO.

src/internal/operators/materialize.ts
```diff
@@ -60,40 +61,23 @@ import { lift } from '../util/lift';
...
+          subscriber,
+          (value) => {
+            subscriber.next(Notification.createNext(value));
+          },
+          (err) => {
+            subscriber.next(Notification.createError(err));
+            subscriber.complete();
+          },
+          () => {
+            subscriber.next(Notification.createComplete()), subscriber.complete();
```
cartant, line 78: nitpick: This might as well be a `;` - instead of a `,` - IMO.

src/internal/operators/OperatorSubscriber.ts
```diff
@@ -0,0 +1,48 @@
...
+    onNext?: (value: T) => void,
+    onError?: (err: any) => void,
+    onComplete?: () => void,
+    private onUnsubscribe?: () => void
+  ) {
+    super(destination);
+    if (onNext) {
+      this._next = function (value: T) {
+        try {
+          onNext?.(value);
```
cartant, line 16: issue: This `?.` should not be necessary. If, for some reason, TypeScript complains, a non-null assertion - `onNext!(value)` - would be a better option.

43081j:
Just a question about this in general.

Most of your large savings are refactors, often by reusing some other functionality we already had. These are great.

But a fair few of them are also near enough "manual minification". Like collapsing switches, conditions, etc into ternaries. This does save you a whole lot of characters, but isn't that what a minifier is for? One could argue a lot of this code was far more readable before this PR.

I'm not aware how rx is distributed to NPM but it seems running terser would achieve the same results for those "minifying" changes? (with mangling off etc)

Is there a reason we didn't go down that route?

src/internal/operators/materialize.ts
benlesh: Yeah, I think that was just a typo. Good call.

src/internal/operators/OperatorSubscriber.ts
benlesh: I think the reason it complains it there's a concern (on the part of the static analyzer, not me) the reference could be reset to undefined before `_next` is called. But you're right, at runtime `!` would be lighter wait.

benlesh: Oh, interestingly it no longer complains. I must have changed something since I added that and not removed it.

src/internal/operators/bufferWhen.ts
benlesh: Good find.

benlesh:
@43081j in many cases, you're right. However, there's no way to _know_ for sure the output will be terse unless you write it in a terse way yourself and ensure it works. Otherwise it's really educated guessing.

In some cases, the move to nested ternaries actually changes the semantics a little. Terser or Closure wouldn't do that.

As for readability, sure it's a concern, but that is why I try to add comments. In >6 years, we've had plenty of verbose "readable" code, and --in the grand scheme of things-- not very many contributors. So I'm not too worried about the barrier to entry being raised because of some ternaries.

43081j:
That's fair enough, it helps to understand the choices so thanks for explaining.

I do think its worthwhile taking into account the readability still, even if very few contribute. So it might be worth avoiding changes which drastically reduce readability for not much saving. Though I have had a quick read through and haven't noticed any such changes at least (most are still pretty readable).

aboyton:
While I've never pushed to RxJS (my one PR was closed without being merged) I actually read through the existing operators semi-regularly when writing my own custom ones. I've always struggled to copy the code directly as there's so much abstraction, helper functions and types, but mainly the abstraction using classes has always confused me. These changes have made reading the code for operators so much simpler that I'm not at all bothered by a few ternaries here and there (especially as they often have nice comments above them).

What I would love to see, if this is the recommended way of writing custom operators, if we could export `OperatorSubscriber` (and maybe `lift` although I know you're wanting to remove that one). Being able to use the same tools externally as internally would be nice (and it might fix strange scheduler issues I've been having with custom operators I've written). That said, it's a nice to have, definitely not required.

Just one data point. Do with it what you will. I am loving seeing all the hard work you've and the team have been putting to RxJS both recently and over the years.

src/internal/operators/window.ts
```diff
@@ -50,77 +49,37 @@ import { SimpleOuterSubscriber, innerSubscribe, SimpleInnerSubscriber } from '..
...
+              window.error(err);
+              subscriber.error(err);
+            },
+            () => {
+              window.complete();
+              subscriber.complete();
+            },
+            () => {
+              window?.unsubscribe();
+              window = null!;
```
cartant, line 74: issue: IMO, this usage means that `window` should be typed as `Subject<T> | null`. If optional chaining is needed hear, AFAICT, it will also be needed in the `error` and `complete` handlers and type current typing - as `Subject<T>` - is hiding what appears to be an error.

src/internal/operators/withLatestFrom.ts
```diff
@@ -66,88 +147,68 @@ export function withLatestFrom<T, R>(array: ObservableInput<any>[], project: (..
...
-  private values: any[];
-  private toRespond: number[] = [];
+    return lift(source, function (this: Subscriber<R | T[]>, source: Observable<T>) {
+      const subscriber = this;
+      const len = inputs.length;
+      const otherValues = new Array(len);
+      // An array of whether or not the other sources have emitted. Matched with them by index.
+      // TODO: At somepoint, we should investigate the performance implications here, and look
+      // into using a `Set()` and checking the `size` to see if we're ready.
+      let hasValue: Record<string, any> = inputs.map(() => false);
```
cartant, line 164: issue: IMO, this should be typed as `boolean[]` 'cause that more accurately represents what it is. The fact that it's typed as `Record<string, any>` - with emphasis on the **any** is why `every` can be called on it, despite `every` not being a `Record` method.

src/internal/operators/throttle.ts
```diff
@@ -63,92 +63,43 @@ export const defaultThrottleConfig: ThrottleConfig = {
...
+    lift(source, function (this: Subscriber<T>, source: Observable<T>) {
+      const subscriber = this;
+      let hasValue = false;
+      let sendValue: T | null = null;
+      let throttled: Subscription | null = null;
+
+      const throttlingDone = () => {
+        throttled?.unsubscribe();
+        throttled = null;
+        trailing && send();
```
cartant, line 80: issue: I think this should also include `!subscriber.closed` in the condition - `trailing && !subscriber.closed && send()` - as `throttlingDone` is called upon unsubscription and it calls `send()`, which subscribes to the duration selector and that's pointless if the subscription has been torn down.

src/internal/operators/window.ts
benlesh: you know... it just occurred to me that I might change the name of that variable too. Just so nothing bad happens where someone deletes `let`. (The operator itself should be changed as well, obviously, but we've never really agreed on what that should be)

benlesh: Okay, I've renamed the variable, and I've moved that teardown so it wasn't called twice (that way the conditional was unnecessary)

src/internal/operators/withLatestFrom.ts
benlesh: Removed the incorrect typing to let inference do it's thing

src/internal/operators/throttle.ts
benlesh: you'll notice that `!this.closed` is checked before this unsubscribe handler is called.

benlesh:
@cartant I wanted to make sure you've noticed that I moved the checklist to the top of the PR, since it was getting collapsed by GitHub.

cartant:
@benlesh I'd added a link to it at the very top of the PR description, but I'll switch to filling out what was copied into the description.

benlesh:
Added some additional comments to `OperatorSubscriber` to make sure it was clearer what is going on. Not just now, but for our future selves.

src/internal/AsyncSubject.ts
```diff
@@ -10,40 +10,28 @@ import { Subscription } from './Subscription';
...
     }
   }
 
   complete(): void {
-    this.hasCompleted = true;
-    if (this.hasNext) {
-      super.next(this.value!);
-    }
+    const { hasValue, value } = this;
+    hasValue && super.next(value!);
```
jayphelps, line 34: Since `this.isStopped` (which used to be `hasCompleted`) is no longer set before doing this `super.next(value!)` is it possible this might trip something up in a reentrancy scenario since it will be incorrectly still `false` until `super.complete()` is called.

src/internal/BehaviorSubject.ts
```diff
@@ -21,25 +19,22 @@ export class BehaviorSubject<T> extends Subject<T> {
   }
 
   /** @deprecated This is an internal implementation detail, do not use. */
-  _subscribe(subscriber: Subscriber<T>): Subscription {
+  protected _subscribe(subscriber: Subscriber<T>): Subscription {
     const subscription = super._subscribe(subscriber);
-    if (subscription && !(<SubscriptionLike>subscription).closed) {
-      subscriber.next(this._value);
-    }
+    !subscription.closed && subscriber.next(this._value);
```
jayphelps, line 24: Minifiers usually do this compression for us. I'm not personally a fan of hand writing conditional invocations like this since it makes it non-obvious, to me, when scanning the code. YTMMV

src/internal/Notification.ts
```diff
@@ -181,16 +154,16 @@ export class Notification<T> {
...
+    const { kind, value, error } = this;
+    return kind === 'N'
+      ? of(value!)
+      : kind === 'E'
+      ? throwError(error)
+      : kind === 'C'
+      ? EMPTY
+      : (() => {
+          throw new TypeError(`Unexpected notification kind ${kind}`);
+        })();
```
jayphelps, line 166: IMO this is taking things too far sacrificing readability for a few bytes, especially after gzip.

src/internal/Notification.ts
```diff
@@ -245,20 +218,11 @@ export class Notification<T> {
...
-      observer.next?.(notification.value!);
-      break;
-    case 'E':
-      observer.error?.(notification.error);
-      break;
-    case 'C':
-      observer.complete?.();
-      break;
-  }
+  kind === 'N' ? observer.next?.(value!) : kind === 'E' ? observer.error?.(error) : observer.complete?.();
```
jayphelps, line 225: If kept as ternaries this needs newline formatting otherwise it's _really_ tough to understand. Has the team considered using prettier maybe?

src/internal/Observable.ts
```diff
@@ -237,13 +237,9 @@ export class Observable<T> implements Subscribable<T> {
...
-        if (canReportError(sink)) {
-          sink.error(err);
-        } else {
-          // If an error is thrown during subscribe, but our subscriber is closed, so we cannot notify via the
-          // subscription "error" channel, it is an unhandled error and we need to report it appropriately.
-          reportUnhandledError(err);
-        }
+        // If an error is thrown during subscribe, but our subscriber is closed, so we cannot notify via the
+        // subscription "error" channel, it is an unhandled error and we need to report it appropriately.
+        canReportError(sink) ? sink.error(err) : reportUnhandledError(err);
```
jayphelps, line 242: Kill me.

jayphelps (changes requested):
I've only been able to get half way through so far, but wanted to submit what I have thus far.

Overall good, but I personally do not like the ultra-hand-minified code approach as I would be afraid it will lead to bugs, as well as increasing the barrier for contribution.

e.g.

```ts
needIterate ? (state = iterate!(state)) : (needIterate = true);
condition && !condition(state) ? subscriber.complete() : subscriber.next(resultSelector(state));
```

If this has already been discussed at length, feel free to ignore my concerns.

Edit:

Okay I see [it's been discussed](https://github.com/ReactiveX/rxjs/pull/5729#issuecomment-695796160). I still disagree, but it's your ballgame.

src/internal/Notification.ts
benlesh: I tend to agree. We're using prettier, however, I think maybe the line length is too long.

benlesh: I think we'll probably just change the line length to the default in another PR. That should correct this issue.

src/internal/BehaviorSubject.ts
benlesh: I opted to specifically "self minify" here, just to _be sure_ that this would be the minified outcome, rather than best-guessing what the tooling would do.

src/internal/AsyncSubject.ts
benlesh: Good catch. I've fixed that and added a test.

api_guard/dist/types/index.d.ts
```diff
@@ -21,16 +21,15 @@ export declare const async: AsyncScheduler;
 export declare const asyncScheduler: AsyncScheduler;
 
 export declare class AsyncSubject<T> extends Subject<T> {
-    _subscribe(subscriber: Subscriber<any>): Subscription;
+    protected checkFinalizedStatuses(subscriber: Subscriber<T>): void;
```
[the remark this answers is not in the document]
benlesh: Good idea.

api_guard/dist/types/index.d.ts
```diff
@@ -310,17 +307,17 @@ export declare function merge<T, T2, T3, T4, T5>(v1: ObservableInput<T>, v2: Obs
 export declare function merge<T, T2, T3, T4, T5, T6>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>, scheduler: SchedulerLike): Observable<T | T2 | T3 | T4 | T5 | T6>;
 export declare function merge<T, T2, T3, T4, T5, T6>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>, concurrent: number, scheduler: SchedulerLike): Observable<T | T2 | T3 | T4 | T5 | T6>;
 export declare function merge<T>(v1: ObservableInput<T>): Observable<T>;
-export declare function merge<T>(v1: ObservableInput<T>, concurrent?: number): Observable<T>;
+export declare function merge<T>(v1: ObservableInput<T>, concurrent: number): Observable<T>;
```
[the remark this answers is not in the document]
benlesh: Because it only served to allow  `merge(source$, undefined)`, which is weird.  `merge(source$)` was caught by the one above it.

api_guard/dist/types/index.d.ts
```diff
@@ -310,17 +307,17 @@ export declare function merge<T, T2, T3, T4, T5>(v1: ObservableInput<T>, v2: Obs
...
+export declare function merge<T, T2, T3, T4, T5>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, concurrent: number): Observable<T | T2 | T3 | T4 | T5>;
 export declare function merge<T, T2, T3, T4, T5, T6>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>): Observable<T | T2 | T3 | T4 | T5 | T6>;
-export declare function merge<T, T2, T3, T4, T5, T6>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>, concurrent?: number): Observable<T | T2 | T3 | T4 | T5 | T6>;
+export declare function merge<T, T2, T3, T4, T5, T6>(v1: ObservableInput<T>, v2: ObservableInput<T2>, v3: ObservableInput<T3>, v4: ObservableInput<T4>, v5: ObservableInput<T5>, v6: ObservableInput<T6>, concurrent: number): Observable<T | T2 | T3 | T4 | T5 | T6>;
```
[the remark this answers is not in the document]
benlesh: We're on TS 4. Just no one has done the work to fix all of our signatures. 😿

spec/operators/find-spec.ts
```diff
@@ -24,12 +24,6 @@ describe('find operator', () => {
     expectSubscriptions(source.subscriptions).toBe(subs);
   });
 
-  it('should throw if not provided a function', () => {
-    expect(() => {
-      of('yut', 'yee', 'sam').pipe(find('yee' as any));
-    }).to.throw(TypeError, 'predicate is not a function');
-  });
-
```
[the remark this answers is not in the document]
benlesh: That's correct. The way we were doing these checks was completely incoherent. We'll have a separate discussion about adding these checks at runtime.

One option I thought of was to actually have a second package that wrapped everything with checks like:

```ts
import { map } from 'rxjs-with-runtime-checks/operators'; 
```

(Or something like that, I'm not married to it). But it's something that adds a lot of bloat and not a huge amount of benefit.

Otherwise, we just need to have a single, coherent approach and type check _everything_ in `rxjs` and people will just have to pay for it.

spec/operators/observeOn-spec.ts
```diff
@@ -81,48 +81,6 @@ describe('observeOn operator', () => {
     expectSubscriptions(e1.subscriptions).toBe(sub);
   });
 
-  it('should clean up subscriptions created by async scheduling (prevent memory leaks #2244)', (done) => {
```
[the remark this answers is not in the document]
benlesh: This was doing deep introspection into something that we no longer have access to. I wasn't quite sure how I could come up with a comparable assertion here, so I decided to remove it and press on. I think the simplified code helps eliminate the possibility of the issues that caused the original problem here.

spec/operators/skipLast-spec.ts
```diff
@@ -132,11 +132,6 @@ describe('skipLast operator', () => {
     expectSubscriptions(e1.subscriptions).toBe(e1subs);
   });
 
-  it('should throw if total is less than zero', () => {
```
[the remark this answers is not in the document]
benlesh: I believe in this case I said that skipping "-1" would be the same as skipping 0. In that it would just mirror the source. Rendering the check useless.

I suppose that should be listed as a breaking change, though.

src/internal/Notification.ts
benlesh: Probably. I've refactored it a little and added comments to help.

benlesh:
Given a more than one person has expressed concern about the hand-minified code: Just to reiterate... **I get it**. However, I wanted to be _sure_ that the code was minified well, and not just guess about what the tooling would do. At somepoint, when we have better checks in place for code size (built into our CI), I think that I'd be happy to go through and make certain things more readable.

However, given the minimal contributions over the last 6 years, I'm unsure if ternaries or inline conditional expressions are the largest barrier to entry. A more valid concern, raised by @jayphelps, is that these patterns might be the source of bugs in the future, where a contributor is tripped up by a switch statement replaced with nested ternaries.

Let's make a point to revisit this concern after this lands and we have some sort of automated size test in our CI pipeline.

src/internal/operators/timeout.ts
```diff
@@ -332,62 +325,71 @@ export function timeout<T, R, M>(
...
+        (originalSourceSubscription = source.subscribe(
+          new OperatorSubscriber(
+            subscriber,
+            (value) => {
+              // clear the timer so we can emit and start another one.
+              timerSubscription.unsubscribe();
+              seen++;
+              // Emit
+              subscriber.next((lastValue = value));
+              each && each > 0 && startTimer(each);
```
cartant, line 375: nitpick: neither `null` nor `undefined` are `> 0`, so the initial `each &&` could be removed.

src/internal/operators/windowCount.ts
```diff
@@ -69,79 +70,66 @@ import { lift } from '../util/lift';
...
-  protected _error(err: any) {
-    const windows = this.windows;
-    if (windows) {
-      while (windows.length > 0 && !this.closed) {
-        windows.shift()!.error(err);
-      }
-    }
-    this.destination.error(err);
-  }
+      const outerSubscriber = new OperatorSubscriber(
```
cartant, line 85: nitpick: This variable is used in only one place: the subscribe call. Other operator implementations pass the `new OperatorSubscriber` directly to `subscribe`.

src/internal/operators/windowTime.ts
```diff
@@ -97,188 +101,100 @@ export function windowTime<T>(windowTimeSpan: number,
...
+       * in the error and complete handlers.
+       */
+      const terminate = (cb: (consumer: Observer<any>) => void) => {
+        loop(({ window }) => cb(window));
+        cb(subscriber);
+        windowRecords = null!;
+        subscriber.unsubscribe();
+      };
+
+      const windowTimeSubscriber = new OperatorSubscriber(
```
cartant, line 179: nitpick: as with `windowCount`, this doesn't really need to be a variable.

src/internal/operators/windowToggle.ts
```diff
@@ -56,152 +58,90 @@ import { lift } from '../util/lift';
...
+          (openValue) => {
+            const window = new Subject<T>();
+            windows.push(window);
+            const closingSubscription = new Subscription();
+            const closeWindow = () => {
+              arrRemove(windows, window);
+              window.complete();
+              closingSubscription.unsubscribe();
+            };
+            const closingSubscriber = new OperatorSubscriber(subscriber, closeWindow, handleError, closeWindow);
```
cartant, line 95: nitpick: Creating the `new OperatorSubscriber` here - rather than inline - doesn't make this more readable, IMO, as the `subscribe` call is quite a bit further down. And the variable isn't used elsewhere.

src/internal/operators/windowWhen.ts
```diff
@@ -50,95 +51,76 @@ import { lift } from '../util/lift';
...
+        // we only cared about the first next or complete notification.
+        closingSubscriber?.unsubscribe();
 
-  protected _complete(): void {
-    this.window!.complete();
-    this.destination.complete();
-    this.unsubscribeClosingNotification();
-  }
+        // Close our window before starting a new one.
+        window?.complete();
```
cartant, line 82: issue: AFAICT, this optional chaining is unnecessary. I can't see that `window` is nulled out anyplace. Maybe it should be? And if it is, the type should reflect that, IMO.

src/internal/operators/mergeMap.ts
```diff
@@ -106,6 +105,12 @@ export function mergeMap<T, R, O extends ObservableInput<any>>(
       // The buffered values from the source (used for concurrency)
       let buffer: T[] = [];
```
cartant, line 106: question: AFAICT, there's nothing here that removes values from the buffer upon explicit unsubscription. If a concurrency is specified, it's entirely possible that there could be a large number of buffered values (or a small number of values that are themselves large). In some other operator implementations, some consideration is given to releasing resources ASAP, but that does not seem to happen here.

src/internal/operators/expand.ts
```diff
@@ -61,114 +70,49 @@ export function expand<T>(project: (value: T, index: number) => ObservableInput<
...
-          destination.add(this.scheduler.schedule<DispatchArg<T, R>>(
-            ExpandSubscriber.dispatch as any,
-            0,
-            state
-          ));
+  return (source: Observable<T>) =>
+    lift(source, function (this: Subscriber<any>, source: Observable<T>) {
+      const subscriber = this;
+      let active = 0;
+      const buffer: (T | R)[] = [];
```
cartant, line 84: question: As with `mergeMap`, a limited concurrency could mean that the buffer could be non-empty upon explicit unsubscription. Maybe ensure the buffer is emptied in those circumstances?

src/internal/operators/debounceTime.ts
```diff
@@ -64,74 +65,42 @@ import { lift } from '../util/lift';
...
+          },
+          undefined,
+          () => {
+            if (hasValue) {
+              subscriber.next(lastValue!);
+              lastValue = null;
+            }
+            subscriber.complete();
+          }
+        )
```
cartant, line 103: question: In `debounce`, AFAICT, a teardown is used to null out the `lastValue` upon explicit unsubscription. Should that be done here, too?

src/internal/operators/delay.ts
```diff
@@ -59,96 +55,75 @@ import { lift } from '../util/lift';
...
+                  checkComplete();
+                }, delay as number)
+              );
+            }
+          },
+          undefined,
+          () => {
+            isComplete = true;
+            checkComplete();
+          }
```
cartant, line 125: question: I guess this is another place that might warrant a teardown that nulls out the array. Thoughts?

src/internal/operators/delayWhen.ts
```diff
@@ -1,15 +1,24 @@
...
 import { Subscriber } from '../Subscriber';
 import { Observable } from '../Observable';
-import { Subscription } from '../Subscription';
-import { ComplexOuterSubscriber, ComplexInnerSubscriber, innerSubscribe } from '../innerSubscribe';
-import { MonoTypeOperatorFunction, TeardownLogic } from '../types';
+import { MonoTypeOperatorFunction } from '../types';
 import { lift } from '../util/lift';
+import { OperatorSubscriber } from './OperatorSubscriber';
+import { concat } from '../observable/concat';
+import { take } from './take';
```
cartant, line 8: issue: This `take` import isn't used. We really should re-enable the `unused-import` lint rule.

The `concat` and `ignoreElements` imports aren't used, either.

src/internal/operators/delayWhen.ts
```diff
@@ -71,151 +80,50 @@ export function delayWhen<T>(delayDurationSelector: (value: T, index: number) =>
...
+      const outerSubscriber = new OperatorSubscriber(
+        subscriber,
+        (value: T) => {
+          const durationNotifier = delayDurationSelector(value, index++);
+          active++;
+          let closed = false;
+          const notify = () => {
+            subscriber.next(value);
+            durationSubscriber?.unsubscribe();
+            !closed && ((closed = true), --active === 0) && isComplete && subscriber.complete();
```
cartant, line 108: issue: I think this piece of terseness might warrant a comment, TBH. Or maybe an apology? 😅

src/internal/operators/delayWhen.ts
```diff
@@ -71,151 +80,50 @@ export function delayWhen<T>(delayDurationSelector: (value: T, index: number) =>
...
+            // A completion should not be a notification. Note the deprecation above
+            notify
+          );
+          durationNotifier.subscribe(durationSubscriber);
+        },
+        undefined,
+        () => {
+          isComplete = true;
+          active === 0 && subscriber.complete();
+          outerSubscriber?.unsubscribe();
```
cartant, line 124: issue: I don't understand why `outerSubscriber?.unsubscribe` needs to be called here. And I cannot see how it could ever require the optional chaining.

src/internal/operators/mergeScan.ts
```diff
@@ -43,96 +44,78 @@ import { SimpleInnerSubscriber, SimpleOuterSubscriber, innerSubscribe } from '..
...
-      }
-      this.destination.complete();
-    }
-  }
+      source.subscribe(
+        new OperatorSubscriber(subscriber, nextSourceValue, undefined, () => {
+          // Outer completed, make a note of it, and check to see if we can complete everything.
+          isComplete = true;
+          checkComplete();
+        })
```
cartant, line 118: question: This operator has a concurrency limit and buffer, too. Should it ensure the buffer is nulled/emptied upon explicit unsubscription? Should there be a general policy on this?

src/internal/operators/mergeMap.ts
benlesh: Good thinking. I can add a clean up to set `buffer` to `null` to ensure that array is GC'ed.

src/internal/operators/expand.ts
benlesh: Yup, I should null that out and release the values.

src/internal/operators/debounceTime.ts
benlesh: I've unified the approach to emitting here, and added more comments.

src/internal/operators/delay.ts
benlesh: Good catch! Added. This PR has been really good for code health, IMO.

src/internal/operators/delayWhen.ts
benlesh: It actually is used for the deprecated path. :) But you're right, I think we need to reenable that.

src/internal/operators/expand.ts
benlesh: I've also made this one a little more efficient in this commit, such that it doesn't always push onto the buffer when it doesn't need to.

src/internal/operators/delayWhen.ts
benlesh: Cleaned up. Added comments.

src/internal/operators/mergeScan.ts
benlesh: >  Should there be a general policy/practice for this?

Yeah, I'm trying to standardize that in these new commits.

src/internal/operators/windowTime.ts
benlesh: Done... Also added a bit to ensure the buffer here is released.

benlesh:
@cartant and @jayphelps ... I think I've addressed all of your comments.

I've unified on an approach for freeing up resources like buffered values, I think... basically, a returned function from the lift is a terse way to add a teardown to the destination subscriber. This ensures that the resource will be freed regardless of source (or inner) completion or error, or destination unsubscription. It could _delay_ the release of those resources in some cases, so things will still have to be addressed on a case-by-case basis. But over all it looks like this:

```ts
return (source: Observable<T>) => lift(source, function(this: Subscriber<R>, source: Observable<T>) {
   const subscriber = this;
   /* stuff here */

   // Additional teardown
   return () => {
      buffer = null!
   };
});
```

This is equivalent to making the last line:  `subscriber.add(() => buffer = null!)`.

src/internal/operators/delayWhen.ts
benlesh: Yeah, that's unnecessary. Removed.

src/internal/operators/repeatWhen.ts
```diff
@@ -38,96 +38,94 @@ import { SimpleOuterSubscriber, innerSubscribe, SimpleInnerSubscriber } from '..
...
-    const subscription = new Subscription();
-    let innerSub: Subscription | null;
-    let syncResub = false;
-    let completions$: Subject<void>;
-    let isNotifierComplete = false;
-    let isMainComplete = false;
+  return (source: Observable<T>) =>
+    lift(source, function (this: Subscriber<T>, source: Observable<T>) {
+      const subscriber = this;
+      const subscription = new Subscription();
```
cartant, line 44: issue: It's not clear to me why this `Subscription` instance is necessary. I mean, as it's returned, it's added to the subscriber - within `Observable#subscribe` - but the subscriber is the destination for both `OperatorSubscribers` in this implementation. If there's a reason for this - given that this is so unlike each of the other operators - I think it warrants a comment.

src/internal/operators/retryWhen.ts
```diff
@@ -61,64 +62,37 @@ import { lift } from '../util/lift';
  */
 export function retryWhen<T>(notifier: (errors: Observable<any>) => Observable<any>): MonoTypeOperatorFunction<T> {
   return (source: Observable<T>) =>
-    lift(source, function (this: Subscriber<T>, source: Observable<T>) {
-      const subscriber = this;
+    wrappedLift(source, (subscriber: Subscriber<T>, source: Observable<T>) => {
       const subscription = new Subscription();
```
cartant, line 66: issue: As with `repeatWhen`, it's not clear to me why this `Subscription` instance is necessary. Maybe I need a coffee?

src/internal/operators/groupBy.ts
```diff
@@ -127,214 +124,139 @@ export function groupBy<T, K, R>(
...
-      try {
-        element = this.elementSelector(value);
-      } catch (err) {
-        this.error(err);
-      }
-    } else {
-      element = value as any;
-    }
+      // Subscribe to the source
+      return source.subscribe(groupBySourceSubscriber);
```
cartant, line 202: issue: AFAICT, this `return` is not necessary. Won't it just try to add the `groupBySourceSubscriber` to the destination `subscriber`? IIRC, other operators don't do this - although, I'm inclined to check them, again, for unnecessary returns.

src/internal/operators/repeatWhen.ts
benlesh: Ah... yeah. I didn't actually touch this one much other than to change it to use `OperatorSubscriber`... but I can rework it a little to see what I can do there.

src/internal/Subject.ts
```diff
@@ -41,94 +42,91 @@ export class Subject<T> extends Observable<T> implements SubscriptionLike {
...
   lift<R>(operator: Operator<T, R>): Observable<R> {
     const subject = new AnonymousSubject(this, this);
-    subject.operator = <any>operator;
-    return <any>subject;
+    subject.operator = operator as any;
+    return subject as any;
   }
 
-  next(value: T) {
+  protected throwIfClosed() {
```
cartant, line 49: issue: I think this protected method should have a leading underscore, too. I would not want this to be seen as part of the public API. I'd like us to reserve the right to remove it and replace it if there is an alternative that offers better perf - as this is, AFAICT, only a code-size-related change.

cartant: Also, whilst in `Subject.ts`, it really irks me that each of these members is implicitly public:

https://github.com/ReactiveX/rxjs/blob/480b7c83c733af35aaa402735bb832652cb9e45d/src/internal/Subject.ts#L17-L25

How about making only `closed` (explicitly) public (for the class's implementation of `SubscriptionLike`) and each of the others either protected or private as appropriate?

src/internal/operators/repeatWhen.ts
benlesh: Updated.

src/internal/operators/retryWhen.ts
benlesh: It's not. Just another one I didn't really touch as I was going through other that updating the `OperatorSubscriber`... will update.

src/internal/operators/groupBy.ts
benlesh: That's correct. This isn't necessary.

cartant (approved):
🚢

Issue #5732, opened 2020-09-15
bug(throttle): Does not emit last value on source complete


benlesh:
`throttleTime(n, s, config)` should be equivalent to `throttle(() => timer(n, s), config)`... But it's not. The issue is that `throttle` does not behave correctly when the `source` completes.

If the `source` completes, we're `trailing`, and we have a trailing value already, the result should wait for the throttle duration to end, then emit the last value, then complete the result. If the `source` completes and we don't have a trailing value -- or we're not trailing -- the result should complete.

alexsherekin:
I would like to work on this issue. It looks like a good first task to contribute to RxJS.

alexsherekin:
After the refactoring the initial bug is fixed. I created a PR with additional tests.

OliverJAsh:
Until v7 is released, I'm using this workaround:

```ts
/**
 * Workaround for https://github.com/ReactiveX/rxjs/issues/5732
 * Fixed in v7
 */
export const throttleWithFix: typeof RxJSOperators.throttle = (...params) => (ob$) =>
  pipe(
    ob$,
    RxJSOperators.publish((published$) => {
      const last$ = pipe(published$, RxJSOperators.last());
      const throttled$ = pipe(published$, RxJSOperators.throttle(...params));
      return RxJS.merge(throttled$, last$);
    }),
  );

```

akaltar:
Is there an option to fix this for a 6.6.7 release?
I'd be happy to try and backport the fix and test from the v7 alpha.

gewoonwoutje:
Closed?

PR #5736, opened 2020-09-19
fix(throttle): Does not emit last value on source complete

alexsherekin:
**Description:**
Fixes a bug when throttle operator did not emit last value on source complete.

https://github.com/ReactiveX/rxjs/issues/5732


benlesh:
Thanks!

I feel like I might have fixed that in #5729 ... please check. Either way, this would blocked until that is in... and we'd just add the tests.

PR #5742, opened 2020-09-23
Issue/5739 smaller operator creation

benlesh:
- Adds new `operate` function for creating operators
- Replaces all `lift` and `wrappedLift` helpers.
- Fixes an issue with `publishReplay` et al, where an error thrown in the selector function was being thrown externally instead of send to the subscriber via an error notification.

Further size reduction.

~~57K -> 55K~~
57K -> 54K approximately

image

### DIFF
--- a/spec/operators/publishReplay-spec.ts
+++ b/spec/operators/publishReplay-spec.ts
@@ -438,8 +438,7 @@ describe('publishReplay operator', () => {
     const source = cold('--1-2---3-4---|');
     const published = source.pipe(publishReplay(1, Infinity, selector));
 
-    // The exception is thrown outside Rx chain (not as an error notification).
-    expect(() => published.subscribe()).to.throw(error);
+    expectObservable(published).toBe('#', undefined, "It's broken");
   });
 
   it('should emit an error when the selector returns an Observable that emits an error', () => {
--- a/src/internal/operators/combineLatestWith.ts
+++ b/src/internal/operators/combineLatestWith.ts
@@ -1,11 +1,10 @@
 /** @prettier */
 import { combineLatestInit } from '../observable/combineLatest';
-import { Observable } from '../Observable';
 import { ObservableInput, OperatorFunction, ObservedValueTupleFromArray, Cons } from '../types';
-import { lift } from '../util/lift';
-import { Subscriber } from '../Subscriber';
+import { operate } from '../util/lift';
 import { argsOrArgArray } from '../util/argsOrArgArray';
 import { mapOneOrManyArgs } from '../util/mapOneOrManyArgs';
+import { pipe } from '../util/pipe';
 
 /* tslint:disable:max-line-length */
 /** @deprecated use {@link combineLatestWith} */
@@ -83,17 +82,15 @@ export function combineLatest<T, TOther, R>(
  */
 export function combineLatest<T, R>(...args: (ObservableInput<any> | ((...values: any[]) => R))[]): OperatorFunction<T, unknown> {
   let project: ((...values: Array<any>) => R) | undefined = undefined;
+
   if (typeof args[args.length - 1] === 'function') {
     project = args.pop() as (...values: any[]) => R;
+    return pipe(combineLatest(...args), mapOneOrManyArgs(project));
   }
 
-  return (source: Observable<T>) => {
-    const result = lift(source, function (this: Subscriber<any>, source: Observable<T>) {
-      return combineLatestInit([source, ...argsOrArgArray(args)])(this);
-    });
-
-    return project ? result.pipe(mapOneOrManyArgs(project)) : result;
-  };
+  return operate((source, subscriber) => {
+    return combineLatestInit([source, ...argsOrArgArray(args)])(subscriber);
+  });
 }
 
 /**
--- a/src/internal/operators/map.ts
+++ b/src/internal/operators/map.ts
@@ -1,8 +1,6 @@
 /** @prettier */
-import { Subscriber } from '../Subscriber';
-import { Observable } from '../Observable';
 import { OperatorFunction } from '../types';
-import { lift } from '../util/lift';
+import { operate } from '../util/lift';
 import { OperatorSubscriber } from './OperatorSubscriber';
 
 /**
@@ -44,19 +42,17 @@ import { OperatorSubscriber } from './OperatorSubscriber';
  * @name map
  */
 export function map<T, R>(project: (value: T, index: number) => R, thisArg?: any): OperatorFunction<T, R> {
-  return (source: Observable<T>) =>
-    lift(source, function (this: Subscriber<R>, source: Observable<T>) {
-      const subscriber = this;
-      // The index of the value from the source. Used with projection.
-      let index = 0;
-      // Subscribe to the source, all errors and completions are sent along
-      // to the consumer.
-      source.subscribe(
-        new OperatorSubscriber(subscriber, (value: T) => {
-          // Call the projection function with the appropriate this context,
-          // and send the resulting value to the consumer.
-          subscriber.next(project.call(thisArg, value, index++));
-        })
-      );
-    });
+  return operate((source, subscriber) => {
+    // The index of the value from the source. Used with projection.
+    let index = 0;
+    // Subscribe to the source, all errors and completions are sent along
+    // to the consumer.
+    source.subscribe(
+      new OperatorSubscriber(subscriber, (value: T) => {
+        // Call the projection function with the appropriate this context,
+        // and send the resulting value to the consumer.
+        subscriber.next(project.call(thisArg, value, index++));
+      })
+    );
+  });
 }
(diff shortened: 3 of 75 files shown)


src/internal/util/lift.ts
```diff
@@ -1,49 +1,30 @@
...
-export function wrappedLift<T, R>(
-  source: Observable<T>,
-  wrappedOperator: (subscriber: Subscriber<R>, liftedSource: Observable<T>) => TeardownLogic
-): Observable<R> {
-  return lift(source, function (this: Subscriber<R>, liftedSource: Observable<T>) {
-    try {
-      wrappedOperator(this, liftedSource);
-    } catch (err) {
-      this.error(err);
+export function operate<T, R>(init: (liftedSource: Observable<T>, subscriber: Subscriber<R>) => TeardownLogic): OperatorFunction<T, R> {
```
benlesh, line 17: This is the most important part of this PR. Please notice it's an amalgamation of the `lift` and `wrappedLift` functions.

src/internal/operators/map.ts
```diff
@@ -44,19 +42,17 @@ import { OperatorSubscriber } from './OperatorSubscriber';
  * @name map
  */
 export function map<T, R>(project: (value: T, index: number) => R, thisArg?: any): OperatorFunction<T, R> {
-  return (source: Observable<T>) =>
-    lift(source, function (this: Subscriber<R>, source: Observable<T>) {
-      const subscriber = this;
```
benlesh, line 49: This best outlines what changed in each file....

```ts
return (source: Observable<T>) =>	 
    lift(source, function (this: Subscriber<R>, source: Observable<T>) {	
      const subscriber = this;
```

becomes:

```ts
```

The same work is done, it's just more concise.

spec/operators/publishReplay-spec.ts
```diff
@@ -438,8 +438,7 @@ describe('publishReplay operator', () => {
     const source = cold('--1-2---3-4---|');
     const published = source.pipe(publishReplay(1, Infinity, selector));
 
-    // The exception is thrown outside Rx chain (not as an error notification).
-    expect(() => published.subscribe()).to.throw(error);
+    expectObservable(published).toBe('#', undefined, "It's broken");
```
benlesh, line 441: Here, there's a strange, old test asserting that errors thrown in a selector function provided to `publishReplay` -- and therefor, by proxy, `multicast`, `publish`, et al -- are simply "thrown" into the ether. Which is obviously wrong.

When `multicast` got the unified operator creation treatment with the new `operate` function, any error thrown during the initialization of the subscription created within the multicast operation was caught and sent to the consumer via an error notification. As we always should have been doing.

src/internal/operators/multicast.ts
```diff
@@ -42,21 +40,21 @@ export function multicast<T, R>(
...
-        // Intentionally terse code: Subscribe to the result of the selector,
-        // then immediately connect the source through the subject, adding
-        // that to the resulting subscription. The act of subscribing with `this`,
-        // the primary destination subscriber, will automatically add the subcription
-        // to the result.
-        selector(subject).subscribe(this).add(source.subscribe(subject));
-      });
-    }
+  if (typeof selector === 'function') {
+    return operate((source, subscriber) => {
```
benlesh, line 46: This change fixed the test in `publishReplay`.

src/internal/operators/withLatestFrom.ts
```diff
@@ -147,68 +146,65 @@ export function withLatestFrom<T, R>(array: ObservableInput<any>[], project: (..
...
-      // TODO: At somepoint, we should investigate the performance implications here, and look
-      // into using a `Set()` and checking the `size` to see if we're ready.
-      let hasValue = inputs.map(() => false);
-      // Flipped true when we have at least one value from all other sources and
-      // we are ready to start emitting values.
-      let ready = false;
+export function withLatestFrom<T, R>(...inputs: any[]): OperatorFunction<T, R | any[]> {
+  let project: (...values: any[]) => R;
+  if (typeof inputs[inputs.length - 1] === 'function') {
+    project = inputs.pop();
```
benlesh, line 152: this fixes an issue that actually wasn't a bug _purely by coincidence_. Before this change, the arguments array was being mutated on every subscription to the result (via `pop()`). However, we're just lucky that the majority of the time, only the very last argument can be a function, so it would not mutate the second time around. (It's possible that this wouldn't be true if someone made a function that was also thennable or implemented `Symbol.observable` or something... but that's a pretty extreme corner case)... Either way, it was resolved when I moved the argument handling out of the initialization to accommodate the new `operate` function.

cartant (changes requested):
LGTM, but there are a bunch of `return` statements in the operators that I think aren't needed. And I have a few nitpicks.

src/internal/util/lift.ts
```diff
@@ -1,49 +1,30 @@
...
- * is related to issues around the "stanky" lifting of static creation functions as operators, see below.
+ * Used to determine if an object is an Observable with a lift function.
  */
-export function lift<T, R>(source: Observable<T>, operator?: Operator<T, R>): Observable<R> {
-  if (hasLift(source)) {
-    return source.lift(operator);
-  }
-  throw new TypeError('Unable to lift unknown Observable type');
+export function hasLift(source: any): source is { lift: InstanceType<typeof Observable>['lift'] } {
+  return source && typeof source.lift === 'function';
```
tonivj5, line 10:
```suggestion
  return typeof source?.lift === 'function';
```

src/internal/util/lift.ts
jakovljevic-mladen: Does it make sense to rename file name as well? To `operate.ts`?

src/internal/operators/combineLatestWith.ts
```diff
@@ -83,17 +82,15 @@ export function combineLatest<T, TOther, R>(
...
 
-  return (source: Observable<T>) => {
-    const result = lift(source, function (this: Subscriber<any>, source: Observable<T>) {
-      return combineLatestInit([source, ...argsOrArgArray(args)])(this);
-    });
-
-    return project ? result.pipe(mapOneOrManyArgs(project)) : result;
-  };
+  return operate((source, subscriber) => {
+    return combineLatestInit([source, ...argsOrArgArray(args)])(subscriber);
```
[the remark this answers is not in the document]
benlesh: I've addressed a host of these in newer commits.

src/internal/operators/raceWith.ts
```diff
@@ -58,7 +58,7 @@ export function race<T>(...args: any[]): OperatorFunction<T, unknown> {
 export function raceWith<T, A extends ObservableInput<any>[]>(
   ...otherSources: A
 ): OperatorFunction<T, T | ObservedValueUnionFromArray<A>> {
-  return (source: Observable<T>) => (!otherSources.length) ? source : lift(source, function(this: Subscriber<any>, source: Observable<T>) {
-    return raceInit([source, ...otherSources])(this);
+  return !otherSources.length ? identity : operate((source, subscriber) => {
```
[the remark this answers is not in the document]
benlesh: I think that we _should_ use the curlies, because it results in a faster noop in `Subscription#add` if it gets `undefined`.

src/internal/operators/zipWith.ts
```diff
@@ -1,48 +1,84 @@
...
 /**
  * @deprecated Deprecated. Use {@link zipWith}.
  */
-export function zip<T, R>(...sources: Array<ObservableInput<any> | ((...values: Array<any>) => R)>): OperatorFunction<T, R> {
-  return (source: Observable<T>) => lift(source, function (this: Subscriber<any>, source: Observable<T>) {
-    const args = [source, ...sources];
-    return zipStatic(...args).subscribe(this);
-  })
+export function zip<T, R>(...sources: Array<ObservableInput<any> | ((...values: Array<any>) => R)>): OperatorFunction<T, any> {
+  return operate((source, subscriber) => zipStatic(source, ...sources).subscribe(subscriber));
```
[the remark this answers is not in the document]
benlesh: I've updated `repeat` and `retry` to "modernize" them to our new way of doing things. They were an oversight in #5729. 

I've also updated the types in `operate` so you can only return `(() => void) | void`. This will help force us to do the right thing, and it caught several others that we didn't see. However, if we push this API public, I suspect there will be a lot of people that try to return `return () => subscription.unsubscribe()` there... which isn't ideal.

src/internal/util/lift.ts
benlesh: Oops... didn't see this and already did it per @cartant's suggestion.  Thanks @tonivj5!

benlesh:
HA! While fixing the stuff @cartant pointed out, I updated `onErrorResumeNext`, `retry` and `repeat`, and knocked us down another 1K almost.

src/internal/util/lift.ts
tonivj5: No problem!, I dind't see @cartant was faster than me.. hahah

cartant (approved):

PR #5745, opened 2020-09-23
Refactor: Simplify the entire subscription process. (Also: smaller tap)

benlesh:
I started off trying to unify the logic between `SafeSubscriber` and `tap` that figures out if it's three functions or an observer, but I ended up with a refactor that updated each one independently. The tap commit is actually unrelated to the other commits at this point, other than they were part of one journey.

- refactor(SafeSubscriber): Smaller implementation
    
    - Also removes unnecessary check to see if the observer was a Subscriber. Subscribers are never passed to SafeSubscriber, so it was unnecessary.

- refactor(tap): Smaller implementation
    
    - Also optimizes dynamic calls of tap that might not pass any handlers or observers.

- refactor(Subscriber): Massively untangle Subscriber and SafeSubscriber
    
    1. `Subscriber` is just a `Subscriber` now. It doesn't do anything magical to wrap the destination it is handed in a `SafeSubscriber`.
    2. `Subscriber.create` returns `SafeSubscriber`.
    3. Simplifies logic in `Subscriber` significantly.
    4. Adds comments outlining code to be removed when the deprecated next context code is removed.
    5. Completely removes `toSubscriber` and related complexity.
    
 
BREAKING CHANGE: `new Subscriber` no longer takes 0-3 arguments. To create a `Subscriber` with 0-3 arguments, use `Subscriber.create`. However, please note that there is little to no reason that you should be creating `Subscriber` references directly, and `Subscriber.create` and `new Subscriber` are both deprecated.


src/internal/Subscriber.ts
```diff
@@ -167,64 +171,14 @@ export class SafeSubscriber<T> extends Subscriber<T> {
...
-        try {
-          this._complete();
-        } catch (err) {
-          this._throw(err);
-          return;
-        }
-      }
-      this.unsubscribe();
-    }
+    this.destination = {
```
benlesh, line 177: Here I'm just leveraging the code that already exists in `Subscriber` to pass calls on to the `destination`, only providing some defaults if we're missing handlers.

src/internal/Subscriber.ts
```diff
@@ -235,4 +189,12 @@ export class SafeSubscriber<T> extends Subscriber<T> {
...
-}
\ No newline at end of file
+}
+
+function handleError(err: any) {
+  if (config.useDeprecatedSynchronousErrorHandling) {
+    throw err;
+  } else {
+    reportUnhandledError(err);
+  }
```
tonivj5, line 199:
```suggestion
  if (config.useDeprecatedSynchronousErrorHandling) {
    throw err;
  }

  reportUnhandledError(err);
```

benlesh: I saw this too late and had already updated the code. Good find! I've added it manually.

cartant (approved):
LGTM. Just a few minor issues with `tap`.

src/internal/operators/tap.ts
```diff
@@ -104,83 +104,54 @@ export function tap<T>(observer: PartialObserver<T>): MonoTypeOperatorFunction<T
...
+    ? (source: Observable<T>): Observable<T> =>
+        lift(source, function (this: Subscriber<T>, source: Observable<T>) {
+          const subscriber = this;
+          // We have to check to see not only if next is a function,
+          // but if error or complete were passed. This is because someone
+          // could technically call tap like `tap(null, fn)` or `tap(null, null, fn)`.
+          const tapObserver =
+            isFunction(observerOrNext) || error || complete
+              ? { next: observerOrNext as (value: T) => void, error, complete }
+              : observerOrNext ?? {};
```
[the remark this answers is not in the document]
benlesh: Oops, that's all just cruft that needed to be removed. 😬

src/internal/operators/tap.ts
```diff
@@ -104,83 +104,54 @@ export function tap<T>(observer: PartialObserver<T>): MonoTypeOperatorFunction<T
...
-  protected _error(err: any) {
-    this.onError(err);
-    super._error(err);
-  }
-
-  protected _complete() {
-    this.onComplete();
-    super._complete();
-  }
+          return source.subscribe(
```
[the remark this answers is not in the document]
benlesh: Yup, using `operate` caught that immediately! 🎉

benlesh:
🔥 🔥 🔥 

image

Issue #5658, opened 2020-08-19
Fix operators that fail the synchronous firehose test


cartant:
The following operators currently fail tests that were introduced in #5652 and those failing tests are currently skipped:

- [x] `catchError` - see #5655
- [x] `exhaust` - see #5743
- [x] `exhaustMap` - see #5743
- [ ] `multicast`
- [x] `raceWith` - see #5743
- [ ] `refCount`
- [x] `repeat` - see #5742
- [x] `repeatWhen` - see #5742
- [x] `retry` - see #5742
- [x] `retryWhen` - see #5743
- [ ] `share`
- [ ] `shareReplay`
- [x] `single` - see #5743
- [x] `skipLast` - see #5749
- [x] `throttle` - see #5749
- [x] `timeout` - see #5748
- [x] `timeoutWith` - see #5748

The tests are marked with `// TODO: fix firehose unsubscription` comments.

Update: [it's not going to be possible to make the `multicast`-based tests pass](https://github.com/ReactiveX/rxjs/pull/5749#pullrequestreview-495185071).

josepot:
:wave: @kobelb 

This is the behavior that I would expect. Because `exhaustMap` will ignore any emissions from its source until the inner-observable hasn't completed. The code of the first example generates an emission into the source, before the inner observable has had a chance to complete. That's why I think that this should be the expected behavior.

> the inner observable isn't marked as completed when the next callback is called

Keep in mind that once an observable has completed, then it can no longer emit any values. So, it's not possible to complete before emitting the value. First it emits and after it completes :slightly_smiling_face:

One way to address that would be to observe the resulting observable of `exhaustMap` through the `asapScheduler`, like this:

```ts
const subject = new BehaviorSubject<number>(0);
const obs$ = subject.pipe(
  tap(() => console.log('before')),
  exhaustMap((val) => {
    return Promise.resolve(val * 100);
  }),
  observeOn(asapScheduler),
  tap(() => console.log('after')),
);

let i = 0;
obs$.subscribe(x => { 
  i++;
  console.log(x)
  if (i === 1) {
    subject.next(i); // the `exhaustMap` causes this to have no effect
  }
});
```

> It feels like these two code-paths should behave the same

I disagree. In the second snipped, the inner observable of `switchMap` creates a micro-task that ensures that when the resulting observable emits the source of `exhaustMap` has completed. The code that should be equivalent (and it is equivalent), if the inner-observable of `switchMap` was synchronous, like this:

[...]

PR #5749, opened 2020-09-24
test: enable remaining synchronous firehose tests

cartant:
**Description:**

This PR enables the remaining synchronous firehose tests and fixes operators to ensure they pass. Several operators have tests that now pass due to changes made in other refactoring PRs.

**Related issue (if exists):** #5658

### DIFF
--- a/spec/operators/multicast-spec.ts
+++ b/spec/operators/multicast-spec.ts
@@ -702,6 +702,22 @@ describe('multicast operator', () => {
   });
 
   // TODO: fix firehose unsubscription
+  // AFAICT, it's not possible for multicast observables to support ASAP
+  // unsubscription from synchronous firehose sources. The problem is that the
+  // chaining of the closed 'signal' is broken by the subject. For example,
+  // here:
+  //
+  // https://github.com/ReactiveX/rxjs/blob/2d5e4d5bd7b684a912485e1c1583ba3d41c8308e/src/internal/operators/multicast.ts#L53
+  //
+  // The subject is passed to subscribe. However, in the subscribe
+  // implementation a SafeSubcriber is created with the subject as the
+  // observer:
+  //
+  // https://github.com/ReactiveX/rxjs/blob/2d5e4d5bd7b684a912485e1c1583ba3d41c8308e/src/internal/Observable.ts#L210
+  //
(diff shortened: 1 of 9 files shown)


spec/operators/multicast-spec.ts
```diff
@@ -702,6 +702,22 @@ describe('multicast operator', () => {
...
+  //
+  // The subject is passed to subscribe. However, in the subscribe
+  // implementation a SafeSubcriber is created with the subject as the
+  // observer:
+  //
+  // https://github.com/ReactiveX/rxjs/blob/2d5e4d5bd7b684a912485e1c1583ba3d41c8308e/src/internal/Observable.ts#L210
+  //
+  // That breaks the chaining of closed - i.e. even if the unsubscribe is
+  // called on the subject, closing it, the SafeSubscriber's closed property
+  // won't reflect that.
```
cartant, line 720: FYI, `multicast`-based operators are not going to pass the synchronous firehose test with the current implementation of the `closed` 'signal'.

src/internal/operators/throttle.ts
```diff
@@ -92,7 +92,10 @@ export function throttle<T>(
       new OperatorSubscriber(subscriber, (value) => {
         hasValue = true;
         sendValue = value;
-        !throttled && (leading ? send() : throttle(value));
+        // If a synchronous duration selector is specified - weird, but legal -
+        // an already-closed subscription will be assigned to throttled, so the
+        // subscription's closed property needs to be checked, too.
+        (!throttled || throttled.closed) && (leading ? send() : throttle(value));
```
cartant, line 98: This is the only significant change/fix in the PR.

spec/operators/multicast-spec.ts
cartant, line 720: FYI, `multicast`-based operators are not going to pass the synchronous firehose test with the current implementation of the `closed` 'signal'.

cartant: I suppose this could be fixed by bringing back the [`SubjectSubscriber`](https://github.com/ReactiveX/rxjs/blob/5c68e208512dd0f3ba1aafbf3e592592ec13db9b/src/internal/Subject.ts#L13-L17).

PR #5750, opened 2020-09-24
feat: stopped notification handler

cartant:
**Description:**

@benlesh

This PR is a WIP representation of what I would like to see implemented as a replacement for the `canReportError` shenanigans - which I would delete as part of this being merged.

The fundamental problem is that errors reported to stopped subscribers are swallowed and no mechanism is offered to RxJS developers to let them alter this behaviour. ATM, there is one place - in the `try`/`catch` that surrounds the `subscribe` within `Observable` - at which the chain of subscribers is traversed to determine whether or not any are stopped/closed and whether the error can be reported. This is pretty crappy - I feel comfortable saying this, because I wrote the code. 😅

I would prefer to ditch the traversal entirely and, instead, offer a configuration point that can be used to register a handler for notifications sent to stopped observers. This would be more general than the traversal that currently happens - I can see it being useful in the DevTools - and would certainly be simpler. IMO, this shouldn't really impact anyone. ATM, in v6, errors that cannot be reported are written to the console; it's only in v7 that an error is actually thrown. Also, to my knowledge, this error swallowing has only ever affected me - when investigating two bugs in the RxJS core.

What are your thoughts on this? AFAICT, it the wiring up of the stopped-notification handler only needs to happen in `Subscriber#next`/`error`/`complete`. If you want me to go ahead with this, I can add some tests and can remove `canReportError`, etc.

IMO, this doesn't have to be perfect; it just has to be better than the traversal - and that shouldn't be too hard. 😅


src/internal/config.ts
```diff
@@ -16,6 +18,20 @@ export const config = {
...
+   * and error notifications sent to stopped subscribers are noops. However, sometimes callers
+   * might want a different behavior. For example, with sources that attempt to report errors
+   * to stopped subscribers, a caller can configure RxJS to throw an unhandled error instead.
+   * This will _always_ be called asynchronously on another job in the runtime. This is because
+   * we do not want errors thrown in this user-configured handler to interfere with the
+   * behavior of the library.
+   */
+  onStoppedNotification: undefined as
+    | ((notification: { kind: 'N'; value: any } | { kind: 'E'; error: any } | { kind: 'C' }, subscriber: Subscriber<any>) => void)
+    | undefined,
```
cartant, line 33: Registration point added to `config` for notifications sent to stopped subscribers.

src/internal/Subscriber.ts
```diff
@@ -62,7 +62,9 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
    * @return {void}
    */
   next(value?: T): void {
-    if (!this.isStopped) {
+    if (this.isStopped) {
+      handleStoppedNotification({ kind: 'N', value }, this);
+    } else {
       this._next(value!);
     }
```
cartant, line 69: Any registered function would be called (asynchronously) within the `next`, `error` and `complete` methods of `Subscriber`.

src/internal/Subscriber.ts
```diff
@@ -74,10 +76,12 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
    * @param {any} [err] The `error` exception.
    * @return {void}
    */
-  error(err?: any): void {
-    if (!this.isStopped) {
+  error(error?: any): void {
+    if (this.isStopped) {
+      handleStoppedNotification({ kind: 'E', error }, this);
```
benlesh, line 81: These could use `errorNotification(error)` or `nextNotification(value)`, etc. It will minify nicer.

src/internal/Subscriber.ts
```diff
@@ -186,6 +192,18 @@ function defaultErrorHandler(err: any) {
...
+/**
+ * A handler for notifications that cannot be sent to a stopped subscriber.
+ * @param notification The notification being sent
+ * @param subscriber The stopped subscriber
+ */
+function handleStoppedNotification(
+  notification: { kind: 'N'; value: any } | { kind: 'E'; error: any } | { kind: 'C' },
+  subscriber: Subscriber<any>
+) {
+  setTimeout(() => config.onStoppedNotification?.(notification, subscriber));
```
benlesh, line 204: We should probably check to see if `onStoppedNotification` is undefined before scheduling the setTimeout:

```ts
const { onStoppedNotification } = config;
onStoppedNotification && setTimeout(() => onStoppedNotification(notification, subscriber))
```

src/internal/config.ts
```diff
@@ -16,6 +18,20 @@ export const config = {
...
+   * A registration point for notifications that cannot be sent to subscribers because they
+   * have completed, errored or have been explicitly unsubscribed. By default, next, complete
+   * and error notifications sent to stopped subscribers are noops. However, sometimes callers
+   * might want a different behavior. For example, with sources that attempt to report errors
+   * to stopped subscribers, a caller can configure RxJS to throw an unhandled error instead.
+   * This will _always_ be called asynchronously on another job in the runtime. This is because
+   * we do not want errors thrown in this user-configured handler to interfere with the
+   * behavior of the library.
+   */
+  onStoppedNotification: undefined as
```
benlesh, line 31: nit: I'd default this to `null`. It's the most common thing that such APIs do, and it's what we're doing with `onUnhandledError`.

src/internal/Notification.ts
```diff
@@ -238,41 +238,4 @@ export function observeNotification<T>(notification: ObservableNotification<T>,
...
- * @internal
- */
-export function createNotification(kind: 'N' | 'E' | 'C', value: any, error: any) {
-  return {
-    kind,
-    value,
-    error,
-  };
-}
+export * from './NotificationFactories';
```
benlesh, line 241: Q: Do we need to export this here? Or can we just import these things directly where they are needed?

benlesh: Actually, given that they're marked `@internal`. I think we should not be exporting these from here. We should just import them directly where they are need.

cartant: It's fine to import them where they're needed. I had to move them because of a circular dependency - that was introduced with their being consumed within `Subscriber` - and I exported them here 'cause this is where they used to live. I can change the import locations instead, if you like.

src/internal/Subscriber.ts
```diff
@@ -186,6 +193,19 @@ function defaultErrorHandler(err: any) {
...
+ * A handler for notifications that cannot be sent to a stopped subscriber.
+ * @param notification The notification being sent
+ * @param subscriber The stopped subscriber
+ */
+function handleStoppedNotification(
+  notification: { kind: 'N'; value: any } | { kind: 'E'; error: any } | { kind: 'C' },
+  subscriber: Subscriber<any>
+) {
+  const { onStoppedNotification } = config;
+  onStoppedNotification && setTimeout(() => onStoppedNotification(notification, subscriber));
```
benlesh, line 206: Q: Should we use a "setTimeoutProvider" here? Do we want this to be testable with the `TestScheduler`? (I guess the same could go for `hostReportError` etc)

src/internal/config.ts
```diff
@@ -16,6 +18,20 @@ export const config = {
    */
   onUnhandledError: null as ((err: any) => void) | null,
 
+  /**
```
benlesh, line 21: Nice doc comment!

benlesh (changes requested):
Really just that re-export needs to be changed, IMO.

And I've left an open question about the `setTimeout` we might want to discuss (but I don't think it should block a merge)

src/internal/config.ts
benlesh: LMAO... thanks GitHub.. Made this look sarcastic. I really did like the comment below this line.

benlesh:
Also needs a rebase it seems.

src/internal/Subscriber.ts
cartant: I've changed the two implementations to use the `immediateProvider`, as we don't have a provider for `setTimeout` - the scheduling implementation uses `setInterval` and using `setImmediate` here didn't involve any keeping-the-return-value-around shenanigans.

The tests still perform their waiting using `setTimeout` 'cause macro tasks happen after micro tasks and 'cause I'm lazy. I can change those, too, if you want.

api_guard/dist/types/index.d.ts
```diff
@@ -189,7 +189,16 @@ export declare function concat<A extends ObservableInput<any>[]>(...observables:
 
 export declare const config: {
     onUnhandledError: ((err: any) => void) | null;
-    Promise: PromiseConstructorLike;
+    onStoppedNotification: ((notification: {
+        kind: 'N';
```
benlesh, line 193: Question: is `ObservableNotification<any>` not adequate here as a type?

src/internal/Subscriber.ts
```diff
@@ -186,6 +194,19 @@ function defaultErrorHandler(err: any) {
   reportUnhandledError(err);
 }
 
+/**
+ * A handler for notifications that cannot be sent to a stopped subscriber.
+ * @param notification The notification being sent
+ * @param subscriber The stopped subscriber
+ */
+function handleStoppedNotification(
+  notification: { kind: 'N'; value: any } | { kind: 'E'; error: any } | { kind: 'C' },
```
benlesh, line 203: Same question as above: Can we use `ObservableNotification<any>` here?

src/internal/config.ts
```diff
@@ -16,6 +18,20 @@ export const config = {
...
+   * have completed, errored or have been explicitly unsubscribed. By default, next, complete
+   * and error notifications sent to stopped subscribers are noops. However, sometimes callers
+   * might want a different behavior. For example, with sources that attempt to report errors
+   * to stopped subscribers, a caller can configure RxJS to throw an unhandled error instead.
+   * This will _always_ be called asynchronously on another job in the runtime. This is because
+   * we do not want errors thrown in this user-configured handler to interfere with the
+   * behavior of the library.
+   */
+  onStoppedNotification: null as
+    | ((notification: { kind: 'N'; value: any } | { kind: 'E'; error: any } | { kind: 'C' }, subscriber: Subscriber<any>) => void)
```
benlesh, line 32: Same question about `ObservableNotification<any>`

src/internal/util/reportUnhandledError.ts
```diff
@@ -11,7 +12,7 @@ import { config } from '../config';
  * @param err the error to report
  */
 export function reportUnhandledError(err: any) {
-  setTimeout(() => {
+  immediateProvider.setImmediate(() => {
```
benlesh, line 15: issue: Errors thrown in this user-provided function (`onUnhandledError`) will be unhandled promise rejections, because `setImmediate` uses a resolved promise to schedule by default. We should probably do this on `setTimeout`, as it will result in errors being sent to `window.onerror` or `process.on('error', fn)` as I think more people will expect.

benlesh (changes requested):
Some questions, but one easy to fix issue.

src/internal/util/reportUnhandledError.ts
cartant: Added `timeoutProvider`, etc.

cartant:
Addressed comments.

src/internal/testing/TestScheduler.ts
```diff
@@ -564,7 +572,29 @@ export class TestScheduler extends VirtualTimeScheduler {
       }
     };
 
-    return { immediate, interval };
+    const timeout = {
```
benlesh, line 575: This could probably use a comment.

benlesh (approved):
There's one spot that could probably use a comment or two... Otherwise LGTM. Squash and merge whenever you're ready.

PR #5753, opened 2020-09-24
Refactor/unify arg popping

benlesh:
fix(from): objects that are thennable that happen to have a subscribe method will no longer error.

refactor: Unify argument parsing and runtime type checks
    
- Ensures that all runtime checks that check to see if a value is a function are done via `isFunction`.
- Unifies a common pattern of inspecting the last argument and popping it off as `popScheduler` and `popResultSelector`.

~ 2KB in savings

(Down to the ~52K range)


src/internal/scheduled/scheduleIterable.ts
```diff
@@ -1,45 +1,49 @@
...
-  return new Observable<T>(subscriber => {
+  return new Observable<T>((subscriber) => {
     const sub = new Subscription();
     let iterator: Iterator<T>;
     sub.add(() => {
       // Finalize generators
-      if (iterator && typeof iterator.return === 'function') {
-        iterator.return();
-      }
+      isFunction(iterator.return) && iterator.return();
```
cartant, line 17: issue: The replaced code tests `iterator`, so optional chaining should be used, IMO.
```
isFunction(iterator?.return)
```

src/internal/observable/interval.ts
```diff
@@ -52,20 +54,17 @@ import { Subscriber } from '../Subscriber';
...
+    scheduler = asyncScheduler;
   }
 
-  return new Observable<number>(subscriber => {
-    subscriber.add(
-      scheduler.schedule(dispatch as any, period, { subscriber, counter: 0, period })
-    );
+  return new Observable<number>((subscriber) => {
+    subscriber.add(scheduler.schedule(dispatch as any, period, { subscriber, counter: 0, period }));
     return subscriber;
```
cartant, line 68: issue: Unrelated to the change, but I don't think it's necessary to return the subscriber here.

src/internal/observable/from.ts
```diff
@@ -130,17 +131,16 @@ export function from<T>(input: ObservableInput<T>, scheduler?: SchedulerLike): O
 }
 
 function subscribeTo<T>(result: ObservableInput<T>): (subscriber: Subscriber<T>) => Subscription | void {
-  if (result && typeof (result as any)[Symbol_observable] === 'function') {
+  if (isFunction((result as any)[Symbol_observable])) {
```
cartant, line 134: issue: This isn't quite the same. `result` is checked, so I think optional chaining is needed to effect exactly the same behaviour:
```
isFunction((result as any)?.[Symbol_observable])
```

src/internal/observable/from.ts
```diff
@@ -130,17 +131,16 @@ export function from<T>(input: ObservableInput<T>, scheduler?: SchedulerLike): O
...
     return subscribeToObservable(result as any);
   } else if (isArrayLike(result)) {
     return subscribeToArray(result);
   } else if (isPromise(result)) {
     return subscribeToPromise(result);
-  } else if (result && typeof (result as any)[Symbol_iterator] === 'function') {
+  } else if (isFunction((result as any)[Symbol_iterator])) {
```
cartant, line 140: issue: Optional chaining again.

cartant (changes requested):
AFAICT, if this PR is to exactly reproduce the checks of the existing implementation, there are a bunch of places where optional chaining needs to be used.

src/internal/scheduled/scheduled.ts
```diff
@@ -29,11 +31,11 @@ export function scheduled<T>(input: ObservableInput<T>, scheduler: SchedulerLike
       return schedulePromise(input, scheduler);
     } else if (isArrayLike(input)) {
       return scheduleArray(input, scheduler);
-    }  else if (isIterable(input) || typeof input === 'string') {
+    } else if (isIterable(input) || typeof input === 'string') {
       return scheduleIterable(input, scheduler);
-    } else if (Symbol && Symbol.asyncIterator && typeof (input as any)[Symbol.asyncIterator] === 'function') {
+    } else if (isFunction((input as any)[Symbol.asyncIterator])) {
```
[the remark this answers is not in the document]
benlesh: Yeah, we can probably do that. IE is the last problem, and people supporting that should have about 5 years of experience adding polyfills.

src/internal/observable/from.ts
benlesh: I've opted instead to just check `result != null` at the start. As it would be more efficient.

benlesh:
A little further down a rabbit hole while addressing your comments: I've refactored `from` and `scheduled` quite a bit, and there's more to do in a follow-up PR as noted in some todos.

From the commit message:

refactor: `from` and `scheduled` use the same methodology
    
- Ensures that `from` and `scheduled` do checks in the same order. Note that I realize that scheduled has no check for `input instanceof Observable`, however, the interop check right after is about the same thing, and I'm not terribly worried about the performance here for the moment. Any optimizations in this area can come once we fully remove the deprecated `from(x, scheduler)` pathway, so we don't bloat that hot-path.
- Unifies the error thrown when a non-observable argument is passed to either `from` or `scheduled`.
- Unifies checks for `isAsyncIterable` and `isAsyncIterable`.
- Removes extra boilerplate around `subscribeTo` logic, as we don't use it inside of any other operators, it was just unnecessary overhead.
- Moves functions closer to actual usage.
- Adds a couple of TODOs around `innerFrom`. Since we're using `from` throughout the library now to convert our returned `ObservableInputs` for inner subscriptions, we are now going to inadvertently include `scheduled` in many operators. We need to switch those operators to use `innerFrom` where possible, as it will omit the bit `scheduled` logic, because it's unnecessary.

benlesh:
Rebased and resolved conflicts.

src/internal/observable/from.ts
```diff
@@ -119,34 +119,130 @@ export function from<O extends ObservableInput<any>>(input: O, scheduler: Schedu
...
+    // Loop over the array and emit each value. Note two things here:
+    // 1. We're making sure that the subscriber is not closed on each loop.
+    //    This is so we don't continue looping over a very large array after
+    //    something like a `take`, `takeWhile`, or other synchronous unsubscription
+    //    has already unsubscribed.
+    // 2. In this form, reentrant code can alter that array we're looping over.
+    //    This is a known issue, but considered an edge case. The alternative would
+    //    be to copy the array before executing the loop, but this has
+    //    performance implications.
+    for (let i = 0; i < array.length && !subscriber.closed; i++) {
```
cartant, line 187: question: Would there be any advantage in using `for ... of` here? 

```ts
for (const value of array) {
  if (subscriber.closed) break;
  subscriber.next(value);
}
```

cartant (approved):
LGTM. Just a speculative question.

src/internal/observable/from.ts
```diff
@@ -119,34 +119,130 @@ export function from<O extends ObservableInput<any>>(input: O, scheduler: Schedu
...
+ */
+function fromInteropObservable<T>(obj: any) {
+  return new Observable((subscriber: Subscriber<T>) => {
+    const obs = (obj as any)[Symbol_observable]();
+    if (typeof obs.subscribe !== 'function') {
+      // Should be caught by observable subscribe function error handling.
+      throw new TypeError('Provided object does not correctly implement Symbol.observable');
+    } else {
+      return obs.subscribe(subscriber);
+    }
```
tonivj5, line 165:
```suggestion
    const obs = obj[Symbol_observable]();
    if (isFunction(obs.subscribe)) {
      return obs.subscribe(subscriber);
    }
    // Should be caught by observable subscribe function error handling.
    throw new TypeError('Provided object does not correctly implement Symbol.observable');
```

`obj` is `any` already, isn't it?

src/internal/util/isObservable.ts
```diff
@@ -1,9 +1,11 @@
...
 import { Observable } from '../Observable';
+import { isFunction } from './isFunction';
 
 /**
  * Tests to see if the object is an RxJS {@link Observable}
  * @param obj the object to test
  */
 export function isObservable<T>(obj: any): obj is Observable<T> {
-  return !!obj && (obj instanceof Observable || (typeof obj.lift === 'function' && typeof obj.subscribe === 'function'));
+  return !!obj && (obj instanceof Observable || (isFunction(obj.lift) && isFunction(obj.subscribe)));
```
tonivj5, line 10:
```suggestion
  return obj && (obj instanceof Observable || (isFunction(obj.lift) && isFunction(obj.subscribe)));
```

src/internal/scheduled/scheduled.ts
```diff
@@ -25,15 +28,19 @@ export function scheduled<T>(input: ObservableInput<T>, scheduler: SchedulerLike
...
+    }
+    if (isPromise(input)) {
+      return schedulePromise(input, scheduler);
+    }
+    if (isIterable(input)) {
       return scheduleIterable(input, scheduler);
-    } else if (Symbol && Symbol.asyncIterator && typeof (input as any)[Symbol.asyncIterator] === 'function') {
+    }
+    if (isAsyncIterable(input)) {
       return scheduleAsyncIterable(input as any, scheduler);
```
tonivj5, line 42:
```suggestion
      return scheduleAsyncIterable(input, scheduler);
```

Using `isAsyncIterable` as type guard, `input` shouldn't be `AsyncIterable` already?

src/internal/observable/from.ts
```diff
@@ -119,34 +119,130 @@ export function from<O extends ObservableInput<any>>(input: O, scheduler: Schedu
...
+      return fromArrayLike(input);
+    }
+    if (isPromise(input)) {
+      return fromPromise(input);
+    }
+    if (isIterable(input)) {
+      return fromIterable(input as any);
+    }
+    if (isAsyncIterable(input)) {
+      return fromAsyncIterable(input as any);
```
tonivj5, line 146: Same `fromAsyncIterable` and `fromIterable`?

src/internal/observable/from.ts
benlesh: The primary reason is that this works for any ArrayLike. `for...of` will not.

Also, `for...of`, last I checked, was notably slower than the `for(;;)` loop, as it uses the iterator instead of accessing the array by index. However, there's an argument to be made for condensing the `fromIterable` and `fromArray` code into one thing, but again, that doesn't work for every ArrayLike. We'd have to do an `Array.from()` first, and then iterate it, which would cause double iteration and slow us down.

src/internal/observable/from.ts
benlesh: Yeah... although looking at those, I'm going to need to reverse their order I think. 🤔

src/internal/util/isObservable.ts
benlesh: So this change actually illustrates why I'm not always a fan of the inline comment changes. I looked and thought "yeah, that's reasonable" and merged it, and it broke a test :)

I've reverted that change and added a comment. But basically, this is a publicly exposed API, and we want to make sure it returns a boolean.... but if someone passes `null`, it would return `null` and not `false` after this change.

PR #5754, opened 2020-09-24
refactor(mergeMap/mergeScan): unify logic

benlesh:
Creates a new `mergeInternals` function that can be used to create a merge-strategy operator, with extensiblity points for accommodating the nuances of each operator.

In theory, we could use this in `concatMap`, etc, directly.

### DIFF
--- a/src/internal/operators/mergeInternals.ts
+++ b/src/internal/operators/mergeInternals.ts
@@ -1,0 +1,109 @@
+/** @prettier */
+import { Observable } from '../Observable';
+import { from } from '../observable/from';
+import { Subscriber } from '../Subscriber';
+import { ObservableInput } from '../types';
+import { OperatorSubscriber } from './OperatorSubscriber';
+
+/**
+ * A process embodying the general "merge" strategy. This is used in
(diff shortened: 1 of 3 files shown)


src/internal/operators/mergeInternals.ts
```diff
@@ -0,0 +1,105 @@
...
+          active--;
+          // If we have more values in the buffer, try to process those
+          // Note that this call will increment `active` ahead of the
+          // next conditional, if there were any more inner subscriptions
+          // to start.
+          if (buffer.length) {
+            while (buffer.length && active < concurrent) {
+              doInnerSub(buffer.shift()!);
+            }
+          }
```
tonivj5, line 75:
```suggestion
          while (buffer.length && active < concurrent) {
            doInnerSub(buffer.shift()!);
          }
```

cartant (approved):
LGTM for a merge once [the comment from @tonivj5](https://github.com/ReactiveX/rxjs/pull/5754/files#r494810245) is addressed.

src/internal/operators/mergeInternals.ts
benlesh: Whoops. haha.

PR #5756, opened 2020-09-25
docs(partition): add an overload for better typing

tech6hutch:
**Description:**

Just adds a new overload to `partition` to give more useful types, similar to the one for `filter`. I found myself wanting this feature, and It seemed easy enough to implement.

### DIFF
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -321,10 +321,10 @@ export declare function pairs(n: number | bigint | boolean | ((...args: any[]) =
 
 export declare type PartialObserver<T> = NextObserver<T> | ErrorObserver<T> | CompletionObserver<T>;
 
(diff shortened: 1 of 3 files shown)


benlesh (changes requested):
Thank you so much for this contribution! Looks like we need some dtslint tests to prove that the typings are indeed better.

(I'm not sure if we actually have any for this one yet, TBH)

tech6hutch:
> Thank you so much for this contribution! Looks like we need some dtslint tests to prove that the typings are indeed better.
> 
> (I'm not sure if we actually have any for this one yet, TBH)

I uh, don't know how to do that, tbh. 😅

benlesh:
@tech6hutch Well, if you're willing to give it a go, all you'd have to do is look at some of the files under `spec-dtslint` and try adding one for `partition`. To run those tests, you just run `npm run dtslint`. 

If you're not, just let me know, and I'll try to find time to do it.

cartant (approved):
@benlesh I've added some dtslint tests and I've updated the PR author's changes to include the `thisArg` changes.

benlesh (approved):

tech6hutch:
I forgot about this - thanks for finishing it!

PR #5757, opened 2020-09-25
Unify approaches to reduce, scan, combineAll and zipAll

benlesh:
- refactor: `combineAll` and `zipAll` use shared logic

- fix(zip): zip now accepts an array of arguments like its counterparts
  - `forkJoin` and `combineLatest` both allow an array of arguments to be passed to it like so: `forkJoin([a$, b$, c$])`. Now `zip` does the same: `zip([a$, b$, c$])`
  
  BREAKING CHANGE: Zipping a single array will now have a different result. This is an extreme corner-case, because it is very unlikely that anyone would want to zip an array with nothing at all.

- refactor(toArray): reduce complexity a little.

  There was a situation where the `toArray` implementation was relying on a quirk of `reduce` and a trick with the `index` in order to ensure that two subscriptions to the same `toArray`-ed observable would not mutate the same underlying array.

  This change gets around that problem in a more straight-forward manner, if at the cost of a few bytes.

- refactor(count): Base off of `reduce`.

  BREAKING CHANGE: No longer passes `source` observable as a third argument to the predicate. That feature was rarely used, and of limited value. The workaround is to simply close over the source inside of the function if you need to access it in there.

- refactor: Unify code that runs `scan` and `reduce`, make reduce a little smaller and more efficient
  - Creates `scanInternals`, which is a shared operator setup between both operators.

### DIFF
--- a/src/internal/observable/zip.ts
+++ b/src/internal/observable/zip.ts
@@ -3,6 +3,9 @@ import { Observable } from '../Observable';
 import { ObservableInput, ObservedValueOf } from '../types';
 import { Subscription } from '../Subscription';
 import { from } from './from';
+import { argsOrArgArray } from '../util/argsOrArgArray';
+import { EMPTY } from './empty';
+import { OperatorSubscriber } from '../operators/OperatorSubscriber';
 
 /* tslint:disable:max-line-length */
 /** @deprecated resultSelector is no longer supported, pipe to map instead */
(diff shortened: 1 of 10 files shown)


src/internal/observable/zip.ts
```diff
@@ -183,47 +186,38 @@ export function zip<O extends ObservableInput<any>, R>(
...
+              },
+              undefined,
+              () => {
+                completed[i] = true;
+                !buffers[i].length && subscriber.complete();
+              }
+            )
+          );
+        }
+        return subscription;
```
cartant, line 220: issue: `subscription` isn't used for anything, AFAICT, it's just created and returned.

cartant (approved):
LGTM, but there's a pre-refactor `Subscription` that doesn't seem to be needed anymore.

src/internal/observable/zip.ts
benlesh: Removed... Added comments and also added a bit to make sure the `buffer` and `completed` arrays were released.

PR #5765, opened 2020-09-28
Removing more unnecessary try catches

benlesh:
Goes through the library removing try/catch blocks that didn't need to be there (because another try/catch lower in the stack does the same thing).

Also refactors `OperatorSubscriber` so it doesn't have a strange special accommodation for `groupBy`, which means that part of `groupBy` had to be wrapped in it's own try/catch. 

(In general, `groupBy` is a convoluted operator that we should revisit at some point)


src/internal/ajax/ajax.ts
```diff
@@ -335,99 +328,95 @@ export function fromAjax<T>(config: AjaxConfig): Observable<AjaxResponse<T>> {
 
     let xhr: XMLHttpRequest;
 
-    try {
```
benlesh, line 338: Here the big `try/catch` that was around everything was removed. That caused the code to shift, but was the only change.

src/internal/ajax/ajax.ts
```diff
@@ -335,99 +328,95 @@ export function fromAjax<T>(config: AjaxConfig): Observable<AjaxResponse<T>> {
...
       }
+    }
 
-      // finally send the request
-      if (body) {
-        xhr.send(body);
-      } else {
-        xhr.send();
-      }
-    } catch (err) {
```
benlesh, line 429: The other end of the big try/catch that was removed.

src/internal/observable/fromEventPattern.ts
```diff
@@ -1,12 +1,20 @@
+/** @prettier */
```
benlesh, line 1: Formatting changes in this file mostly.

src/internal/observable/fromEventPattern.ts
```diff
@@ -137,32 +145,19 @@ export function fromEventPattern<T>(addHandler: (handler: NodeEventHandler) => a
...
-    );
+    return fromEventPattern<T>(addHandler, removeHandler).pipe(mapOneOrManyArgs(resultSelector));
   }
 
-  return new Observable<T | T[]>(subscriber => {
+  return new Observable<T | T[]>((subscriber) => {
     const handler = (...e: T[]) => subscriber.next(e.length === 1 ? e[0] : e);
-
-    let retValue: any;
-    try {
```
benlesh, line 155: Actual try/catch that was removed. The rest of the changes are formatting changes.

src/internal/operators/OperatorSubscriber.ts
```diff
@@ -31,13 +31,7 @@ export class OperatorSubscriber<T> extends Subscriber<T> {
...
           onNext(value);
         } catch (err) {
-          // NOTE: At some point we may want to refactor this to send to
-          // `destination.error(err)`. Currently, this is the way it is *only* to
-          // accommodate `groupBy`, with minimal ill effects to other operators, but
-          // it does mean that additional logic is being fired at each step during
-          // an error call. Which since it is by definition an exceptional state, probably
-          // isn't a big deal. Just making a note of this here so context isn't lost.
-          this.error(err);
+          this.destination.error(err);
```
benlesh, line 34: Here we're going to a more expected pattern, as I fixed `groupBy` to not rely on a quirky implementation.

src/internal/operators/bufferWhen.ts
```diff
@@ -49,38 +48,49 @@ import { innerFrom } from '../observable/from';
...
+      // about one notification.
       closingSubscriber?.unsubscribe();
-
+      // emit the buffer if we have one, and start a new buffer.
       const b = buffer;
       buffer = [];
       b && subscriber.next(b);
 
-      let closingNotifier: Observable<any>;
-      try {
```
benlesh, line 63: Removed this try/catch... the rest was just adding comments.

src/internal/operators/groupBy.ts
```diff
@@ -143,49 +147,59 @@ export function groupBy<T, K, R>(
     const groupBySourceSubscriber = new GroupBySubscriber(
       subscriber,
       (value: T) => {
-        const key = keySelector(value);
+        // Because we have to notify all groups of any errors that occur in here,
+        // we have to add our own try/catch to ensure that those errors are propagated.
+        // OperatorSubscriber will only send the error to the main subscriber.
+        try {
```
benlesh, line 153: here, I added this try/catch to accommodate the fact that any error thrown in this particular `next` handler had to go to the `handleError` function, which previously was the function that we passed to `OperatorSubscriber`, before I changed the quirky behavior there.

benlesh:
Added some comments here to assist in the review process.

cartant (approved):
LGTM

PR #5767, opened 2020-09-29
Refactor some scheduled jobs with try/catch

benlesh:
Unifies a few areas where we are scheduling work that requires a try/catch that sends to `subscriber`.

Some other minor clean up of schedule-related calls.

### DIFF
--- a/src/internal/operators/timeout.ts
+++ b/src/internal/operators/timeout.ts
@@ -357,31 +357,29 @@ export function timeout<T, R, M>(config: number | Date | TimeoutConfig<T, R, M>,
       );
     };
 
-    subscriber.add(
-      (originalSourceSubscription = source.subscribe(
-        new OperatorSubscriber(
-          subscriber,
-          (value) => {
-            // clear the timer so we can emit and start another one.
+    originalSourceSubscription = source.subscribe(
+      new OperatorSubscriber(
+        subscriber,
+        (value: T) => {
+          // clear the timer so we can emit and start another one.
+          timerSubscription?.unsubscribe();
+          seen++;
+          // Emit
+          subscriber.next((lastValue = value));
+          // null | undefined are both < 0. Thanks, JavaScript.
+          each! > 0 && startTimer(each!);
+        },
+        undefined,
+        undefined,
+        () => {
+          if (!timerSubscription?.closed) {
             timerSubscription?.unsubscribe();
-            seen++;
-            // Emit
-            subscriber.next((lastValue = value));
-            // null | undefined are both < 0. Thanks, JavaScript.
-            each! > 0 && startTimer(each!);
-          },
-          undefined,
-          undefined,
-          () => {
-            if (!timerSubscription?.closed) {
-              timerSubscription?.unsubscribe();
-            }
-            // Be sure not to hold the last value in memory after unsubscription
-            // it could be quite large.
-            lastValue = null;
           }
-        )
-      ))
+          // Be sure not to hold the last value in memory after unsubscription
+          // it could be quite large.
+          lastValue = null;
+        }
+      )
     );
 
     // Intentionally terse code.
--- a/src/internal/operators/timeout.ts
+++ b/src/internal/operators/timeout.ts
@@ -7,6 +7,7 @@ import { operate } from '../util/lift';
 import { Observable } from '../Observable';
 import { innerFrom } from '../observable/from';
 import { createErrorClass } from '../util/createErrorClass';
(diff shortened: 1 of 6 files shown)


src/internal/scheduled/schedulePromise.ts
```diff
@@ -1,21 +1,23 @@
...
-          sub.add(scheduler.schedule(() => subscriber.complete()));
-        }));
-      },
-      err => {
-        sub.add(scheduler.schedule(() => subscriber.error(err)));
-      }
-    )));
-    return sub;
+  return new Observable<T>((subscriber) => {
+    return scheduler.schedule(() =>
```
cartant, line 7: nitpick: The scheduled action's subscription has already been added to the subscriber, so returning it isn't necessary.

src/internal/scheduled/scheduleIterable.ts
```diff
@@ -22,36 +23,29 @@ export function scheduleIterable<T>(input: Iterable<T>, scheduler: SchedulerLike
...
-                // The iterable is not done, emit the value.
-                subscriber.next(value);
-                // Reschedule. This will cause this function to be
-                // called again on the same scheduled delay.
-                this.schedule();
-              }
+        caughtSchedule(subscriber, scheduler, function () {
+          // Check to make sure teardown was not triggered
+          // by the consumer since the last iteration.
+          if (!subscriber.closed) {
```
cartant, line 29: nitpick: This isn't part of the refactor - it's in the original code - but I'm not sure this `subscriber.closed` check is needed here. If the subscriber has been explicitly unsubscribed, so too will have been the scheduled action that was added to it. So the work function should never be called in this situation.

FWIW, the refactor in `src/internal/observable/generate.ts` removes the outer and inner checks of `subscriber.closed` - which I think is the correct thing to do, as the rescheduling `schedule` method checks `closed` internally - so I'm not sure why the outer check is needed here.

cartant (approved):
LGTM, just some nitpicks regarding code that I think can be removed.

src/internal/scheduled/scheduleIterable.ts
```diff
@@ -22,36 +23,29 @@ export function scheduleIterable<T>(input: Iterable<T>, scheduler: SchedulerLike
...
+            let value: T;
+            let done: boolean | undefined;
+            // Pull the value out of the iterator
+            ({ value, done } = iterator.next());
```
tonivj5, line 33: I think here destructuring could be used (and inferred type is okay?)
```ts
            // Pull the value out of the iterator
            let { value, done } = iterator.next();
```

src/internal/operators/timeout.ts
```diff
@@ -336,52 +337,46 @@ export function timeout<T, R, M>(config: number | Date | TimeoutConfig<T, R, M>,
...
-              })
-            );
-          } catch (err) {
-            subscriber.error(err);
-            return;
-          }
+      timerSubscription = caughtSchedule(
+        subscriber,
+        scheduler,
+        () => {
```
tonivj5, line 343: In the rest of places this param has been a traditional anonymous function but here an arrow function is used, is it intentional?

src/internal/scheduled/scheduleIterable.ts
cartant: AFAICT, the inference in this suggestion would work if the `iterator` declaration were changed to:

```ts
let iterator: Iterator<T, T>;
```

The second type parameter is necessary - otherwise, it defaults to `any` and the union in `IteratorResult` effects `any` as the inferred type for `value`.

Also, `const` could be used instead of `let`, if the suggestion is accepted.

src/internal/scheduled/schedulePromise.ts
benlesh: In this case, I don't think that's true. This bit was just removing an unnecessary `Subscription`.

src/internal/scheduled/scheduleIterable.ts
benlesh: change made.

src/internal/scheduled/schedulePromise.ts
cartant: 🤦‍♂️

src/internal/operators/timeout.ts
benlesh: Yeah. We didn't need the `this` context in this particular case. We only use `function () {}` if we need to reschedule with `this.schedule()`.

Issue #3665, opened 2018-05-07
delayWhen should use an indefinite delay for notifiers that complete without emitting


cartant:
This issue's title has been changed, as the implementation of `delayWhen` is not correct - see the discussion [below](https://github.com/ReactiveX/rxjs/issues/3665#issuecomment-387758861).

The original issue's description follows.

**RxJS version:** 6.1.0

**Code to reproduce:**

```ts
import { EMPTY, of } from "rxjs";
import { delay, delayWhen } from "rxjs/operators";

of(0).pipe(
    delayWhen(() => EMPTY.pipe(delay(0)))
).subscribe(console.log);
```

**Expected behavior:**

`0` should be emitted and logged to the console.

**Actual behavior:**

Nothing is emitted and logged to the console.

If the `of` function is passed `1` instead of `0`, `1` is logged to the console.

**Additional information:**

Looking at `delayWhen`, I can see another problem (other than #3663) with the implementation. If the value is falsy, and the notifier completes - rather than nexts - the value won't be emitted.

benlesh:
So the semantics here seem off... The notifier completes without notifying, therefor the delay should last indefinitely. This isn't a bug with the library so much as it is with the usage.

Airblader:
@benlesh But this is explicitly documented behavior (which is broken).

Airblader:
@benlesh The docs explicitly state that the notifier completing should cause the value to be emitted.

benlesh:
🤦‍♂️

benlesh:
Okay, so we need to fix the issue, but deprecate the behavior.

Hyumiris:
I have recently run into the same issue\(s\).

#### Regarding the deprecation
Even though there is a deprecation warning for Observable\<never\> this only works if the generated Observable is known to be of type never at compile time.
In every other case the deprecation warning isn't shown and the [official documentation](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-delayWhen) still doesn't mention anything about the depreciation.
#### Regarding the bug
The bug that the behaviour of the delayWhen operator depends on the truthiness of the source value persists as well.

Since the initial reproduction code had some oddities here is a more concise version:
```
import { of, EMPTY } from 'rxjs'; 
import { delayWhen } from 'rxjs/operators';

of(true, false, 0, 1).pipe(
  delayWhen(() => EMPTY)
).subscribe(console.log);

// this logs:
// >> true
// >> 1
```

[...]

PR #5769, opened 2020-09-29
fix(delayWhen): no longer emits if duration selector is empty

benlesh:
- Resolves an issue where a duration selector that never emitted a value would cause the value to be emitted, even though there wasn't a "notification".
- `delayWhen` is now based on `mergeMap`. I started off with trying to use `mergeInternals`, but quickly realized it was doing the same thing as `mergeMap`, and I could just use that directly.

I'm not sure if this size optimization is "taking it too far", but I feel that `delayWhen` isn't exactly "hot path" code, so this optimization is fine, IMO, given the bundle savings it should yield.

### DIFF
--- a/spec/operators/delayWhen-spec.ts
+++ b/spec/operators/delayWhen-spec.ts
@@ -91,13 +91,13 @@ describe('delayWhen operator', () => {
     expectSubscriptions(selector.subscriptions).toBe(selectorSubs);
   });
 
-  it('should delay by selector completes if selector does not emits', () => {
+  it('should delay, but not emit if the selector never emits a notification', () => {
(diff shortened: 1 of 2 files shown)


niklas-wortmann:
I am not sure if this is really a fix. I mean I get your point, but we had a test for the behavior when the delaySelector just completes. This could break people, right?

benlesh:
This will resolve #3665.

benlesh:
@niklas-wortmann Yes, this is a fix, which would be a breaking change, but it's still a bug fix and we want to get it into v7.

benlesh:
Updated commit message.

cartant (approved):

PR #5770, opened 2020-09-29
refactor(expand): use mergeInternals

benlesh:
- Adds a small amount of complexity to `mergeInternals` to support `expand`.
- `expand` now based off of the same internals as `mergeMap` and `mergeScan`.
- `deprecates` the scheduler argument to `expand` with an explanation.

NOTE: While I'm not fond of the number of arguments to `mergeInternals`, it's an _internal_ API, and given that it's not public, I'd like to refrain from creating named parameters via a configuration object just to make our internal code more readable, especially when this is relatively hot code. Instead, I've opted to make sure everything is well commented.


benlesh:
Probably the craziest part of this, is we can get `expandScan` done in a few lines of code.

benlesh:
(I'm thinking of you, @trxcllnt) <3 lol

cartant (approved):
LGTM

PR #5772, opened 2020-09-29
docs(refCount): update description

### LINKED ISSUE #4799: [docs]: The operator refCount needs an adoption of the description

- [ ] typo
- [x] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

### Description Of The Issue

The docs say: 
[Note that using the share operator is exactly the same as using the publish operator (making the observable hot) and the refCount operator in a sequence.](https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/refCount.ts#L18-L19)
 
But in fact, this is not right. If we take a look a the source code `share` internally uses `multicast` with a subjetFactory and `refCount`.
[refCount()(multicast(shareSubjectFactory)(source))](https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/share.ts#L25)

So it is NOT "exactly the same as using the publish operator" because share, as well as shareReplay, always use a subjectFactory where publish uses depending on the optional selector either a static subject or a subjectFactory.

Also, the marble diagram of refCount does not serve any value.

### Suggested behavior:
replace:
Note that using the share operator is exactly the same as using the publish operator (making the observable hot) and the refCount operator in a sequence.
with:
Note that using the share operator is exactly the same as using the multicast(() => new Subject()) operator (making the observable hot) and the refCount operator in a sequence.

hardikpthv:
**Description:**
- Update description for the `refCount` operator as suggested in #4799
- Do I also need to remove the marble image?

- Closes #4799

### DIFF
--- a/src/internal/operators/refCount.ts
+++ b/src/internal/operators/refCount.ts
@@ -15,7 +15,7 @@ import { OperatorSubscriber } from './OperatorSubscriber';
  * refCount has only a single subscription independently of the number of subscribers to the target
  * observable.
  *
- * Note that using the {@link share} operator is exactly the same as using the *publish* operator
+ * Note that using the {@link share} operator is exactly the same as using the `multicast(() => new Subject())` operator 
  * (making the observable hot) and the *refCount* operator in a sequence.
  *
  * ![](refCount.png)


niklas-wortmann (approved):
LGTM

PR #5774, opened 2020-09-29
docs(last): add description and example for the operator

### LINKED ISSUE #4870: [docs]: Suggestions for `last` documentation improvements

The `last` operator: 
[last.ts](https://github.com/ReactiveX/rxjs/blob/b24d3ea85d33b893e4c020fb4fa9f6fa4d05dc0b/src/internal/operators/last.ts)

- [ ] typo
- [ ] error(s)/adoption/updates in example
- [ ] needs example

### Description Of The Issue

The detailed description is missing. Also, see also section is not present.

#### Description
The detailed description is missing. Here we should include that:
- it will error on if on the notification is sent but the source completes
- it returns last value or if a predicate is used last value that matches the predicate
- it returns the given default value if no notification is emitted or matches the predicate

#### See Also
We miss the see also section completely.

A suggested list for related operators could look like this:
```typescript
 * @see {@link skip}
 * @see {@link skipUntil}
 * @see {@link skipLast}
 * @see {@link skipWhile}
```

In case we also want to reference take operators we also might want to include the following operators:
```typescript
 * @see {@link first}
 * @see {@link take}
 * @see {@link takeWhile}
 * @see {@link takeUntil}
 * @see {@link takeLast}
```

hardikpthv:
**Description:**
- Add description, example, and See Also section for the `skip` operator

- Closes #4870

### DIFF
--- a/src/internal/operators/last.ts
+++ b/src/internal/operators/last.ts
@@ -30,8 +30,8 @@ export function last<T, D = T>(
  *
  * ![](last.png)
  * 
- * It will throw an error if the notification is sent but the source is completed. It returns the
- * last value or if a predicate is provided last value that matches the predicate. It returns the 
+ * It will throw an error if the source completes without notification or one that matches the predicate. It
+ * returns the last value or if a predicate is provided last value that matches the predicate. It returns the 
  * given default value if no notification is emitted or matches the predicate.
  * 
  * ## Example
--- a/src/internal/operators/last.ts
+++ b/src/internal/operators/last.ts
@@ -29,6 +29,38 @@ export function last<T, D = T>(
  * from the source Observable that satisfies the predicate.
  *
  * ![](last.png)
+ * 
+ * It will throw an error if the notification is sent but the source is completed. It returns the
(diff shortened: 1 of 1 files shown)


src/internal/operators/last.ts
```diff
@@ -29,6 +29,38 @@ export function last<T, D = T>(
  * from the source Observable that satisfies the predicate.
  *
  * ![](last.png)
+ * 
+ * It will throw an error if the notification is sent but the source is completed. It returns the
```
niklas-wortmann, line 33: I think this should be a little bit clearer, e.g. 
"It will throw an error if the source completes without notification or one that matches the predicate"

niklas-wortmann (changes requested):
Hi thanks a lot for filling this gap in the documentation. I would just like to see some minor wording change to make things a little bit clearer.

src/internal/operators/last.ts
hardikpthv: Just updated.

niklas-wortmann (approved):
LGTM

PR #5775, opened 2020-09-29
refactor(pairs): Based off of `from` and `Object.entries`

benlesh:
BREAKING CHANGE: `pairs` will no longer function in IE without a polyfill for `Object.entries`. `pairs` itself is also deprecated in favor of users just using `from(Object.entries(obj))`.


niklas-wortmann (approved):
LGTM

cartant (approved):

PR #5776, opened 2020-09-29
refactor(generate): Use an actual generator

benlesh:
- This seems obvious.

NOTE: There's some weirdness with the types here, and it seems like we need to figure out how to resolve it in either `defer` or `ObservableInput`? @kolodny @cartant

### DIFF
--- a/src/internal/observable/generate.ts
+++ b/src/internal/observable/generate.ts
@@ -1,7 +1,7 @@
 /** @prettier */
 import { Observable } from '../Observable';
 import { identity } from '../util/identity';
-import { SchedulerLike } from '../types';
+import { ObservableInput, SchedulerLike } from '../types';
 import { isScheduler } from '../util/isScheduler';
 import { defer } from './defer';
 import { scheduleIterable } from '../scheduled/scheduleIterable';
@@ -347,7 +347,13 @@ export function generate<T, S>(
   if (arguments.length == 1) {
     // If we only have one argument, we can assume it is a configuration object.
     // Note that folks not using TypeScript may trip over this.
-    ({ initialState, condition, iterate, resultSelector = identity as any, scheduler } = initialStateOrOptions as GenerateOptions<T, S>);
+    ({
+      initialState,
+      condition,
+      iterate,
+      resultSelector = identity as ResultFunc<S, T>,
+      scheduler,
+    } = initialStateOrOptions as GenerateOptions<T, S>);
   } else {
     // Deprecated arguments path. Figure out what the user
     // passed and set it here.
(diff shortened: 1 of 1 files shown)


src/internal/observable/generate.ts
```diff
@@ -360,24 +360,23 @@ export function generate<T, S>(
...
+  // We use `defer` because we want to defer the creation of the iterator from the iterable.
+  return defer(
+    scheduler
+      ? // If a scheduler was provided, use `scheduleIterable` to ensure that iteration/generation
+        // happens on the scheduler.
+        () => scheduleIterable(gen(), scheduler!)
+      : // Otherwise, if there's no scheduler, we can just use the generator function directly in
+        // `defer` and executing it will return the generator (which is iterable).
+        (gen as any)
+  ) as any;
```
cartant, line 381: issue: The problem with the types is that the ternary introduces a union type for the callback received by `defer`. more simply:

```ts
let b = false;
defer(b ? () => Promise.resolve(1) : [2]);
```

Here, the type of the callback is `(() => Promise<number) | (() => [number])` and that union defeats `defer`'s type signature's inference.

You could solve the problem with a narrower assertion (than `any`), like this:

```ts
  return defer(
    (scheduler
      ? // If a scheduler was provided, use `scheduleIterable` to ensure that iteration/generation
        // happens on the scheduler.
        () => scheduleIterable(gen(), scheduler!)
      : // Otherwise, if there's no scheduler, we can just use the generator function directly in
        // `defer` and executing it will return the generator (which is iterable).
        gen) as () => ObservableInput<T>
  );
```

Or could move `defer` into the ternary, like this:

```ts
  return scheduler
    ? // If a scheduler was provided, use `scheduleIterable` to ensure that iteration/generation
      // happens on the scheduler.
      defer(() => scheduleIterable(gen(), scheduler!))
    : // Otherwise, if there's no scheduler, we can just use the generator function directly in
      // `defer` and executing it will return the generator (which is iterable).
      defer(gen)
  );
```

I would opt for the former - the narrower assertion.

src/internal/observable/generate.ts
```diff
@@ -344,13 +345,12 @@ export function generate<T, S>(
...
   if (arguments.length == 1) {
-    const options = initialStateOrOptions as GenerateOptions<T, S>;
-    initialState = options.initialState;
-    condition = options.condition;
-    iterate = options.iterate;
-    resultSelector = options.resultSelector || (identity as ResultFunc<S, T>);
-    scheduler = options.scheduler;
+    // If we only have one argument, we can assume it is a configuration object.
+    // Note that folks not using TypeScript may trip over this.
+    ({ initialState, condition, iterate, resultSelector = identity as any, scheduler } = initialStateOrOptions as GenerateOptions<T, S>);
```
cartant, line 350: nitpick: I think this assertion could be narrower than `any`. Using `identity as ResultFunc<S, T>` should work. If not, `identity as ResultFunc<any, T>`

cartant (approved):
LGTM. Suggested some alternatives for the `any` assertions.

PR #5779, opened 2020-09-29
refactor(fromEvent): smaller implementation

benlesh:
- Adds a lot of comments.
- Adds one `TODO` that notes we may not want to be throwing the `TypeError` so late, it might be better for everyone if we throw that error as soon as we know about it, as it would provide a better call stack.

### DIFF
--- a/src/internal/observable/fromEvent.ts
+++ b/src/internal/observable/fromEvent.ts
@@ -228,7 +228,7 @@ export function fromEvent<T>(
   // event registry points, and we'd rather delegate to that when possible.
   if (!add) {
     if (isArrayLike(target)) {
-      return mergeMap((target: any) => fromEvent(target, eventName, options as EventListenerOptions))(
+      return mergeMap((subTarget: any) => fromEvent(subTarget, eventName, options as EventListenerOptions))(
         internalFromArray(target)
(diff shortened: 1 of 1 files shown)


niklas-wortmann (approved):

src/internal/observable/fromEvent.ts
```diff
@@ -200,43 +203,91 @@ export function fromEvent<T>(
...
+  // Check to see if it is an ArrayLike, because if it is, we want to
+  // try to apply fromEvent to all of it's items. We do this check last,
+  // because there are may be some types that are both ArrayLike *and* implement
+  // event registry points, and we'd rather delegate to that when possible.
+  if (!add) {
     if (isArrayLike(target)) {
-      return (mergeMap((subTarget: any) => fromEvent(subTarget, eventName, options as any))(internalFromArray(target)) as Observable<
-        T
-      >).subscribe(subscriber);
+      return mergeMap((target: any) => fromEvent(target, eventName, options as EventListenerOptions))(
```
cartant, line 231: issue: `target` is effecting a lint failure, as it shadows a variable in an outer scope. FWIW, I have little patience for that rule and don't use it.

cartant (approved):

PR #5780, opened 2020-09-29
fix(bindNodeCallback): ensure underlying function is not called twice…

benlesh:
… during subscription

- Fixes an issue where `bindNodeCallback` observables would call the underlying function twice if it has not called back yet, and more than one subscription occurred.
- Reduces the size of the implementation
- Adds more comments to the file.


benlesh:
Node that `bindCallback` will have exactly the same issue.

cartant (approved):

PR #5781, opened 2020-09-30
refactor: `bindCallback` and `bindNodeCallback` are based on the same…

benlesh:
… code

- Add `bindCallbackInternals`, a base implementation for both creation methods
- Add a missing test to `bindNodeCallback` to catch an issue where the initial call to the provided function was not scheduled.
- Updates comments

This work effectively just copies and pastes the work from #5780 into a new function called `bindCallbackInternals`, and alters it a bit to account for the differences in the call patterns. Then it adds the use of `subscribeOn` to the scheduled code path, because `bindCallback` had a test that caught something that the tests for `bindNodeCallback` did not. Also adds that test for `bindNodeCallback`.

### DIFF
--- a/src/internal/observable/bindNodeCallback.ts
+++ b/src/internal/observable/bindNodeCallback.ts
@@ -1,10 +1,7 @@
 /** @prettier */
-import { AsyncSubject } from '../AsyncSubject';
 import { Observable } from '../Observable';
-import { observeOn } from '../operators/observeOn';
 import { SchedulerLike } from '../types';
-import { isScheduler } from '../util/isScheduler';
-import { mapOneOrManyArgs } from '../util/mapOneOrManyArgs';
+import { bindCallbackInternals } from './bindCallbackInternals';
 
 /** @deprecated resultSelector is deprecated, pipe to map instead */
 export function bindNodeCallback(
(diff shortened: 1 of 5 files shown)


src/internal/observable/bindNodeCallback.ts
```diff
@@ -255,98 +252,10 @@ export function bindNodeCallback(callbackFunc: Function, scheduler?: SchedulerLi
  * deliver.
  * @name bindNodeCallback
  */
-export function bindNodeCallback<T>(
+export function bindNodeCallback(
   callbackFunc: Function,
```
cartant, line 256: issue: Not related to the refactor, but can we get rid of the `Function` usage while we're here? It's a completely useless type. IIRC, it's something like `{ name: string }`. (I think I'll write a lint rule, now, to ban it.) `(...args: any[]) => any` would be better. Maybe we should have an internal type for an arbitrary function?

I can see that the `bindCallback` and `bindNodeCallback` implementation also make liberal use of `any` instead of an arbitrary function type. They should be fixed up, too, IMO. Maybe in another commit.

cartant (approved):
Looks okay to me. Quite a few `any`s - and the `Function` type is also less than ideal - but those could be addressed in a separate PR. They're in the existing code, rather than introduced in this refactor, so ... 🤷‍♂️

src/internal/observable/bindNodeCallback.ts
benlesh: Sure. I think it's just there from "way back when" TBH. I was pretty new to TypeScript whenever that was written.

benlesh: I'll leave it for another PR though.

PR #5782, opened 2020-09-30
refactor(fromFetch): some clean up on the implementation

benlesh:
Just what it says. A little smaller. Removes an unnecessary subscription reference. Adds comments, etc.

### DIFF
--- a/src/internal/observable/dom/fetch.ts
+++ b/src/internal/observable/dom/fetch.ts
@@ -1,6 +1,6 @@
 /** @prettier */
+import { OperatorSubscriber } from '../../operators/OperatorSubscriber';
 import { Observable } from '../../Observable';
-import { Subscription } from '../../Subscription';
 import { innerFrom } from '../../observable/from';
 import { ObservableInput } from '../../types';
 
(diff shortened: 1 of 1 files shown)


niklas-wortmann (approved):

src/internal/observable/dom/fetch.ts
```diff
@@ -100,56 +100,70 @@ export function fromFetch<T>(
...
+    // Any externally provided AbortSignal will have to call
+    // abort on this controller when signaled, because the
+    // signal from this controller is what is being passed to `fetch`.
     const controller = new AbortController();
-    const signal = controller.signal;
+    const { signal } = controller;
+    // This flag exists to make sure we don't `abort()` the fetch
+    // before we code is done using the response.
+    // In particular in the case where a `selector` is provided.
     let abortable = true;
```
cartant, line 112: issue: This comment isn't quite right, IMO. The fetch is abortable for as long as the Response's lifetime is bound by this observable's lifetime. If the Response is emitted in a `next` notification, `abortable` needs to be set to `false`, as unsubscription from this observable would abort the promise from a subsequent `json()` - or whatever - call that could be made in a flattening operator.

I'd suggest a comment like this:

```ts
// This flag exists to make sure we don't `abort()` the fetch upon tearing down
// this observable after emitting a Response. Aborting in such circumstances
// would also abort subsequent methods - like `json()` - that could be called
// on the Response.
```

cartant (changes requested):

src/internal/observable/dom/fetch.ts
benlesh: That's right. Thanks, @cartant

cartant (approved):

PR #5783, opened 2020-10-01
Fix tslint rules (plus a fix for `every`)

benlesh:
- Removes every tslint override that (IMO) was bad except one
- Adds a `TODO` for the `unified-signatures` override, because I felt addressing that would be a minefield of type changes.
- fixes an issue with `every` that was uncovered by the removal of the `prefer-const` override, where we were not incrementing the index being passed to the predicate. Adds a test for that one.

Consequently, this change also removes all usage of `Function`, which I know @cartant hated passionately.

Other thoughts:

- We should move to `eslint` soon.
- Many of the linting issues were actually fixed during the refactor.
- I couldn't get tslint to set up `Array` usage in a way that _I personally_ agreed with, so that is still on
- the `variable-names` bit is still on because it conflicts with the unused variables override of prefixing with an underscore.

**Note that commits labelled as `chore` were mostly non-changes. Just deleting a rule or an irrelevant bit of code or test.**

### DIFF
--- a/src/internal/observable/combineLatest.ts
+++ b/src/internal/observable/combineLatest.ts
@@ -359,7 +359,7 @@ export function combineLatest<R>(
 
 // combineLatest({})
 export function combineLatest(sourcesObject: {}): Observable<never>;
-export function combineLatest<T, K extends keyof T>(sourcesObject: T): Observable<{ [K in keyof T]: ObservedValueOf<T[K]> }>;
+export function combineLatest<T>(sourcesObject: T): Observable<{ [K in keyof T]: ObservedValueOf<T[K]> }>;
 
 /* tslint:enable:max-line-length */
 
@@ -504,10 +504,10 @@ export function combineLatest<O extends ObservableInput<any>, R>(...args: any[])
       scheduler,
       keys
         ? // A handler for scrubbing the array of args into a dictionary.
-          (args: any[]) => {
+          (values: any[]) => {
             const value: any = {};
-            for (let i = 0; i < args.length; i++) {
-              value[keys![i]] = args[i];
+            for (let i = 0; i < values.length; i++) {
+              value[keys![i]] = values[i];
             }
             return value;
           }
--- a/src/internal/observable/combineLatest.ts
+++ b/src/internal/observable/combineLatest.ts
@@ -545,7 +545,7 @@ class CombineLatestSubscriber<T> extends Subscriber<T> {
 
 export function combineLatestInit(
   observables: ObservableInput<any>[],
-  scheduler: SchedulerLike | undefined = undefined,
+  scheduler: SchedulerLike | undefined,
   valueTransform: (values: any[]) => any = identity
 ) {
   return (subscriber: Subscriber<any>) => {
--- a/src/internal/observable/combineLatest.ts
+++ b/src/internal/observable/combineLatest.ts
@@ -545,7 +545,7 @@ class CombineLatestSubscriber<T> extends Subscriber<T> {
 
 export function combineLatestInit(
   observables: ObservableInput<any>[],
-  scheduler: SchedulerLike | undefined,
+  scheduler?: SchedulerLike | undefined,
   valueTransform: (values: any[]) => any = identity
 ) {
   return (subscriber: Subscriber<any>) => {
--- a/src/internal/observable/combineLatest.ts
+++ b/src/internal/observable/combineLatest.ts
@@ -545,7 +545,7 @@ class CombineLatestSubscriber<T> extends Subscriber<T> {
 
 export function combineLatestInit(
   observables: ObservableInput<any>[],
-  scheduler?: SchedulerLike | undefined,
+  scheduler?: SchedulerLike,
   valueTransform: (values: any[]) => any = identity
 ) {
   return (subscriber: Subscriber<any>) => {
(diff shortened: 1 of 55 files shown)


tslint.json
```diff
@@ -8,39 +8,20 @@
...
     "object-literal-sort-keys": [false],
     "no-this-assignment": [false],
-    "one-variable-per-declaration": [false],
     "no-conditional-assignment": [false],
-    "no-unnecessary-initializer": [false],
-    "max-classes-per-file": [true, 10],
+    "max-classes-per-file": [false, 1000],
+    // TODO: unified signatures not be disabled, but it will
+    // be a lot of work to verify the fixes don't mess something up.
     "unified-signatures": [false],
```
cartant, line 23: thought: TBH, I don't agree with this rule anyway. Sometimes separate signatures are clearer.

src/internal/util/argsArgArrayOrObject.ts
```diff
@@ -27,6 +27,6 @@ export function argsArgArrayOrObject<T, O extends Record<string, T>>(args: T[] |
   return { args: args as T[], keys: null };
 }
 
-function isPOJO(obj: any): obj is Object {
+function isPOJO(obj: any): obj is {} {
```
cartant, line 30: issue: IMO this should be `object` - note the lower case - rather than `{}`.

`{}` is wider than `object` and will match any value that's neither `null` nor `undefined`. The test within the implementation implies that the result is very much an object.

FWIW, the default `ban-types` configuration in the TypeScript _ESLint_ rule bans `object`, but I disagree with that. TypeScript's own types - e.g. for `WeakMap` - make use of `object`.

src/internal/scheduler/animationFrameProvider.ts
```diff
@@ -1,7 +1,7 @@
 /** @prettier */
 import { Subscription } from '../Subscription';
 
-type AnimationFrameProvider = {
+interface AnimationFrameProvider {
```
cartant, line 4: thought: If this change from a `type` to an `interface` was effected by a lint rule, I would switch it off. Increasingly, types are able to due more and more. Personally, I would favour using interfaces only when they offer a feature that's not offered by a type. If it's a rule, I think it's a useless rule. Mostly because TypeScript has changed so much in the last few years.

src/internal/observable/pairs.ts
```diff
@@ -56,6 +56,6 @@ import { from } from './from';
  * [key, value] pairs from the object.
  * @deprecated To be removed in version 8. Use `from(Object.entries(obj))` instead.
  */
-export function pairs<T>(obj: Object, scheduler?: SchedulerLike): Observable<[string, T]> {
+export function pairs<T>(obj: { [s: string]: T } | ArrayLike<T>, scheduler?: SchedulerLike): Observable<[string, T]> {
```
cartant, line 59: nitpick: I'd use `Record<string, T> | ArrayLike<T>` here.

src/internal/observable/fromEvent.ts
```diff
@@ -22,8 +22,8 @@ export interface NodeCompatibleEventEmitter {
 }
 
 export interface JQueryStyleEventEmitter {
-  on: (eventName: string, handler: Function) => void;
-  off: (eventName: string, handler: Function) => void;
+  on: (eventName: string, handler: (event: any) => any) => void;
+  off: (eventName: string, handler: (event: any) => any) => void;
```
cartant, line 26: issue: jQuery listener signatures look like [this](https://github.com/ReactiveX/rxjs/blob/9e9064b7513aec4c365e792beac211f9d12bd7b9/spec-dtslint/observables/fromEvent-spec.ts#L17-L23). See also #5584 - which is gonna need some rebasing, now.

src/internal/observable/combineLatest.ts
```diff
@@ -545,7 +545,7 @@ class CombineLatestSubscriber<T> extends Subscriber<T> {
 
 export function combineLatestInit(
   observables: ObservableInput<any>[],
-  scheduler: SchedulerLike | undefined = undefined,
+  scheduler?: SchedulerLike | undefined,
```
cartant, line 548: nitpick: If `scheduler` is optional, the `undefined` is implied and doesn't have to be included in the union.

cartant (approved):
LGTM, but I'm pretty sure the changes to the jQuery signatures aren't quite right.

src/internal/util/argsArgArrayOrObject.ts
benlesh: Given what we're checking here and why we're checking it, maybe `is Record<string, T>` or `is Record<string, any>` is probably what I wanted?

benlesh: `is object` seems to pass, so I'll leave it as you suggest.

src/internal/scheduler/animationFrameProvider.ts
benlesh: Hmmm... for me this was about just homogenizing the approach. FWIW: I think if we're just defining an interface (like one for a provider) than it should probably be an interface. However, if we're defining some sort of argument type that could be more than one thing, or a function shape, then I guess I'd prefer a `type`. But I suppose it's all personal preference.

src/internal/observable/fromEvent.ts
benlesh: Must've been mangled in my other work. Thanks for catching that.

benlesh: Oh, I see, that's in the dtslint test.

I've made the fix and uncommented the test, updating it a little.

PR #5784, opened 2020-10-01
refactor(skipLast): rename skipLast's parameter to `count`

jakovljevic-mladen:
I renamed `skipLast`'s parameter from `skipCount` to `count` for two reasons:
- the docs in the first sentence is still referring to this parameter named as `count`;
- to have to same name as the other operators (e.g.`skip`, `take` or `takeLast`).


benlesh:
I'm not sure this makes the code any better. If anything, it's less readable now with `count`, `counter` and `currentCount`. Admittedly it wasn't great before, but at least it was obvious which one came from the argument.

benlesh:
@jakovljevic-mladen I've added a related PR, and I'd be happy if you reviewed it: https://github.com/ReactiveX/rxjs/pull/5788

PR #5785, opened 2020-10-02
docs(skipWhile): add description and example for the operator

### LINKED ISSUE #4869: [docs]: Suggestions for `skipWhile` documentation improvements

The `skipWhile` operator: 
[skipWhile.ts](https://github.com/ReactiveX/rxjs/blob/b24d3ea85d33b893e4c020fb4fa9f6fa4d05dc0b/src/internal/operators/skipWhile.ts)

- [ ] typo
- [X] documentation doesn't exist
- [ ] error(s)/adoption/updates in example
- [X] needs example

### Description Of The Issue

The detailed description is missing. Also, the example as well as seel also links are missing.
We might want to improve this to serve a better more detailed description for the community.

#### Description
The detailed description is missing. Here we should include that:
- it skips all notifications  with a true or truthy predicate 
- and don't skip notifications  with a false or falsy predicate
- you can skip notifications  using the index
- once predicate is true it will never get called again

#### Example
We at least provide a basic example.

#### See Also
We miss the see also section completely.

A suggested list for related operators could look like this:
```typescript
 * @see {@link last}
 * @see {@link skip}
 * @see {@link skipUntil}
 * @see {@link skipLast}
```

In case we also want to reference take operators we also might want to include the following operators:
```typescript
 * @see {@link first}
 * @see {@link take}
 * @see {@link takeWhile}
 * @see {@link takeUntil}
 * @see {@link takeLast}
```

hardikpthv:
**Description:**
- Add description, example, and See Also section for the `skipWhile` operator

- Closes #4869

### DIFF
--- a/src/internal/operators/skipWhile.ts
+++ b/src/internal/operators/skipWhile.ts
@@ -9,6 +9,37 @@ import { OperatorSubscriber } from './OperatorSubscriber';
  *
  * ![](skipWhile.png)
  *
+ * Skips all the notifications with a truthy predicate. It will not skip the notifications when the predicate is falsy.
+ * It can also be skipped using index. Once the predicate is true, it will not be called again.
+ *
+ * ## Example
+ * Using Value: Skip some super heroes
+ * ```ts
+ * import { from } from 'rxjs';
+ * import { skipWhile } from 'rxjs/operators';
+ *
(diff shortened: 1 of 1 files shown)


hardikpthv:
Could you take a look @niklas-wortmann? 😄

niklas-wortmann (approved):
LGTM. Appreciate your contributions!

PR #5791, opened 2020-10-03
WIP: Draft (concept of smaller operators still)

benlesh:
- Removes crazy `source` arguments from predicates and other callbacks, as they really don't make sense.
- Adds common `index++` logic to `OperatorSubscriber`.

Rough idea in place.


benlesh:
This is outdated, but I'm still thinking about implementing this concept. I'm going to close this for now to reduce noise.

PR #5793, opened 2020-10-03
docs(versioning): add versioning to docs website

niklas-wortmann:


docs_app/.firebaserc
```diff
@@ -0,0 +1,11 @@
+{
+  "targets": {
+    "rxjs-dev": {
+      "hosting": {
+        "stable": [
+          "rxjs-dev"
```
benlesh, line 6: I don't actually know much about this configuration. Can you explain what this is doing?

niklas-wortmann: sure, so we have one firebase project called rxjs-dev, we only use the hosting feature of firebase. Within firebase though we host now one "site" per version and those get an alias. So if you do `firebase deploy --only hosting:stable` it will deploy the current dist folder in the stable site. It's kinda like an alias for the site that is used within this branch

benlesh (approved):
LGTM. Rebase and merge when you're ready.

I think the failure on the "latest" build is expected and we need to figure out how to make it optional on the 6.x branch. 🤔

PR #5796, opened 2020-10-05
Update skip and skipLast tests to run mode

jakovljevic-mladen:
~`take`, `takeLast` and `skipLast` are not throwing errors anymore when provided with negative values, so I updated docs and added tests for such cases. Also, I changed `skip` and `skipLast` tests to use `TestScheduler` in run mode.~

[Updated](https://github.com/ReactiveX/rxjs/pull/5796#issuecomment-705690602).

### DIFF
--- a/spec/operators/skip-spec.ts
+++ b/spec/operators/skip-spec.ts
@@ -1,151 +1,188 @@
 import { expect } from 'chai';
-import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
 import { skip, mergeMap, take } from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


jakovljevic-mladen:
I updated this PR. With previous version, I wanted to add multiple commits which would certainly add confusion. Now, I have removed commits that were removing `@throws` stuff from operators that are no longer throwing - I saw that there are other PRs that solve this issue. Also, I removed (and saved for another PR) commits that are adding new tests that cover cases for sending negative values to `take`, `takeLast` and `skipLast`.

cartant (approved):
LGTM. Thanks.

PR #5797, opened 2020-10-05
fix(delayWhen): only deprecates when subscriptionDelay presents

imcotton:
Noticed this after **vscode** starts [showing][vsc] the deprecation from jsdoc, not sure how to test this change tho.

[vsc]: https://code.visualstudio.com/updates/v1_49#_deprecated-tag-support-for-javascript-and-typescript

### DIFF
--- a/api_guard/dist/types/operators/index.d.ts
+++ b/api_guard/dist/types/operators/index.d.ts
@@ -72,7 +72,8 @@ export declare function defaultIfEmpty<T, R = T>(defaultValue?: R): OperatorFunc
 export declare function delay<T>(due: number | Date, scheduler?: SchedulerLike): MonoTypeOperatorFunction<T>;
 
 export declare function delayWhen<T>(delayDurationSelector: (value: T, index: number) => Observable<never>, subscriptionDelay?: Observable<any>): MonoTypeOperatorFunction<T>;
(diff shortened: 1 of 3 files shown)


imcotton:
`api_guardian` is failing, not exactly have fixing in mind, could use some help.

cartant:
Thanks for the PR.

> ... not sure how to test this change tho.

We should add some dtslint tests to make sure that the deprecations work as expected. ATM, there aren't many examples of this, as the `ExpectDeprecation` assertion was something that I added in anticipation of fixing up these sorts of false positives - and I've not yet got around to that. For an example, have a look at `of`:

https://github.com/ReactiveX/rxjs/blob/338b1746e11247bbdc6d617b634fab900717288f/spec-dtslint/observables/of-spec.ts#L122-L139

To run the dtslint tests, use `npm run dtslint`.

Note that there is also an `ExpectNoDeprecation` assertion.

> `api_guardian` is failing

You should be able to use `npm run api_guardian:update` to update its files.

imcotton:
Updated, have `spec-dtslint` and `api_guardian` included.

benlesh (approved):

benlesh:
Thank you @imcotton

Issue #5730, opened 2020-09-15
bug(audit): Does not end audit period and emit when source completes


benlesh:
I would expect that with `audit`, if the source closes before the duration ends, we'd get the last value it captured. For whatever reason we decided not to do that ages ago, and I can't imagine why.

When the source completes, I think it should either:

1. If it has seen a value, wait for the duration to end, then emit the value and complete the result.

OR

2. Immediately emit the last value it saw (if any), then complete the result.

I'm inclined to say #1 makes the most sense. It's also the behavior of throttle.

alexsherekin:
> 1. If it has seen a value, wait for the duration to end, then emit the value and complete the result.

This would mean, that we never emit and never complete if the duration is `never` or just neither emits nor completes. Is it the desired behavior?

PR #5799, opened 2020-10-05
fix(audit, auditTime): audit and auditTime emit last value after source completes

alexsherekin:
**Description:**
Adjust audit operator to emit the last value when source completes

#5730

### DIFF
--- a/spec/operators/audit-spec.ts
+++ b/spec/operators/audit-spec.ts
@@ -120,17 +120,17 @@ describe('audit operator', () => {
 
   it('should handle a busy producer emitting a regular repeating sequence', () => {
     testScheduler.run(({ hot, cold, expectObservable, expectSubscriptions }) => {
-      const e1 = hot('  abcdefabcdefabcdefabcdefa|');
-      const e1subs = '  ^------------------------!';
-      const e2 = cold(' -----|                    ');
+      const e1 = hot('  abcdefabcdefabcdefabcdefa|    ');
+      const e1subs = '  ^------------------------!    ';
+      const e2 = cold(' -----|                        ');
       const e2subs = [
-        '               ^----!                    ',
-        '               ------^----!              ',
-        '               ------------^----!        ',
-        '               ------------------^----!  ',
-        '               ------------------------^!'
+        '               ^----!                        ',
+        '               ------^----!                  ',
+        '               ------------^----!            ',
+        '               ------------------^----!      ',
+        '               ------------------------^----!'
       ];
-      const expected = '-----f-----f-----f-----f-|';
+      const expected = '-----f-----f-----f-----f-----(a|)';
 
       const result = e1.pipe(audit(() => e2));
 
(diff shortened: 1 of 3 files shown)


spec/operators/audit-spec.ts
```diff
@@ -168,13 +168,13 @@ describe('audit operator', () => {
...
-  it('should emit no values if duration is a never', () => {
+  it('should emit no values and never complete if duration is a never', () => {
     testScheduler.run(({ hot, cold, expectObservable, expectSubscriptions }) => {
       const e1 = hot('  ----abcdefabcdefabcdefabcdefa|');
       const e1subs = '  ^----------------------------!';
       const e2 = cold(' -');
-      const e2subs = '  ----^------------------------!';
-      const expected = '-----------------------------|';
+      const e2subs = '  ----^-------------------------';
+      const expected = '------------------------------';
```
benlesh, line 177: Hmmm... I'll need to think about this. I think this makes the most sense. Given that there are items to audit, and we're waiting for a duration that just never arrives... But I want a second opinion: @cartant

src/internal/operators/audit.ts
```diff
@@ -56,6 +56,7 @@ export function audit<T>(durationSelector: (value: T) => SubscribableOrPromise<a
     let hasValue = false;
     let lastValue: T | null = null;
     let durationSubscriber: Subscriber<any> | null = null;
+    let isComplete = false;
 
     const endDuration = () => {
```
benlesh, line 61: @cartant, do we want to fix the duration notifier treating completion like a notification here? Or shall we get that at a later time? The whole thing is weird. Like `[]` "Here's your value, in this array!". My gut tells me it should be a `noop`, and I think we all agreed on that.

benlesh (changes requested):
Just a few things, and some open questions for other core team members.

spec/operators/audit-spec.ts
cartant: Yeah, I think this makes sense, too. Above all, the behaviour needs to be consistent between operators and, AFAICT, `delayWhen` doesn't complete if its notifier is `NEVER`: https://github.com/ReactiveX/rxjs/blob/0d3094f29b5f6f2cc4d2829f8a0e20addd2106a3/spec/operators/delayWhen-spec.ts#L131-L144

So this LGTM.

src/internal/operators/audit.ts
cartant: Yeah, we should fix it. Whether it's fixed here or someplace else, I don't mind. I intend to do another pass through all of the operators that take a notifier to make sure their behaviour is consistent, etc.

cartant (approved):
LGTM. @benlesh, I added responses to your comments.

benlesh (approved):

benlesh:
Thanks, @alexsherekin!

elepner:
Hello! This change broke our code after update. We use it to detect long press gesture and relied on the fact that audit does not emit last value if observable completes during audit window. Is it possible to add the same configuration as ``throttle`` or ``takeWhile`` operators do?

PR #5803, opened 2020-10-07
Bind callback ts4

kolodny:
**Description:**

TS4-ified the `bindCallback` and `bindNodeCallback` functions. 

Note: I also changed all `=> any` to `=> void` since that's more "correct". LMK if I should revert that


api_guard/dist/types/index.d.ts
```diff
@@ -34,73 +34,15 @@ export declare class BehaviorSubject<T> extends Subject<T> {
...
-export declare function bindNodeCallback<A1, A2, A3, A4, A5, R1>(callbackFunc: (arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5, callback: (err: any, res1: R1) => any) => any, scheduler?: SchedulerLike): (arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5) => Observable<R1>;
-export declare function bindNodeCallback<A1, A2, A3, A4, A5>(callbackFunc: (arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5, callback: (err: any) => any) => any, scheduler?: SchedulerLike): (arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5) => Observable<void>;
-export declare function bindNodeCallback(callbackFunc: (...args: any[]) => any, scheduler?: SchedulerLike): (...args: any[]) => Observable<any[]>;
+export declare function bindCallback(callbackFunc: (...args: any[]) => void, resultSelector: (...args: any[]) => void): (...args: any[]) => Observable<any>;
+export declare function bindCallback<A extends readonly unknown[], R extends readonly unknown[]>(callbackFunc: (...args: [...A, (...res: R) => void]) => void, schedulerLike?: SchedulerLike): (...arg: A) => Observable<R extends [] ? void : R extends [any] ? R[0] : R>;
```
benlesh, line 38: nit: Can we add comments about why the types exist? Like what scenarios does it cover? (Just `//` comments, not doc comments)

benlesh: Just for example, it's unclear to me why `readonly unknown[]` is there? I _think_ I know why, but I also know that I'm probably wrong. lmao

benlesh (approved):
Approved with a small nit about adding some explanatory comments.

benlesh:
@kolodny please see my comments squash and merge when you're ready.

api_guard/dist/types/index.d.ts
kolodny: I imagine you mean to add this to the source, this file is generated so any comments I make here will get wiped away. Regarding why I used `readonly unknown[]` is because it's more "correct" when specifying a tuple type that will get narrowed down. I added some comments at https://github.com/ReactiveX/rxjs/pull/5803/files#diff-475e2e1db68a71816e10b4d64b41ea8bR9 but I can flesh them out a bit more.

src/internal/observable/bindCallback.ts
```diff
@@ -4,48 +4,15 @@ import { bindCallbackInternals } from './bindCallbackInternals';
 
 // tslint:disable:max-line-length
 /** @deprecated resultSelector is no longer supported, use a mapping function. */
-export function bindCallback(callbackFunc: (...args: any[]) => any, resultSelector: (...args: any[]) => any, scheduler?: SchedulerLike): (...args: any[]) => Observable<any>;
+export function bindCallback(callbackFunc: (...args: any[]) => void, resultSelector: (...args: any[]) => void): (...args: any[]) => Observable<any>;
```
cartant, line 7: issue: This change also remove the scheduler parameter. That seems wrong.

nitpick: The return type for `resultSelector` really should be `any`, as the internals use the return value. With `callbackFunc`, using `void` is fine, as we ignore any return value, but with `resultSelector` then return value should reflect its usage. A result selector that returns `void` makes no sense.

src/internal/observable/bindCallback.ts
```diff
@@ -4,48 +4,15 @@ import { bindCallbackInternals } from './bindCallbackInternals';
...
 
-export function bindCallback<A1, R1, R2, R3, R4>(callbackFunc: (arg1: A1, callback: (res1: R1, res2: R2, res3: R3, res4: R4, ...args: any[]) => any) => any, scheduler?: SchedulerLike): (arg1: A1) => Observable<any[]>;
-export function bindCallback<A1, R1, R2, R3>(callbackFunc: (arg1: A1, callback: (res1: R1, res2: R2, res3: R3) => any) => any, scheduler?: SchedulerLike): (arg1: A1) => Observable<[R1, R2, R3]>;
-export function bindCallback<A1, R1, R2>(callbackFunc: (arg1: A1, callback: (res1: R1, res2: R2) => any) => any, scheduler?: SchedulerLike): (arg1: A1) => Observable<[R1, R2]>;
-export function bindCallback<A1, R1>(callbackFunc: (arg1: A1, callback: (res1: R1) => any) => any, scheduler?: SchedulerLike): (arg1: A1) => Observable<R1>;
-export function bindCallback<A1>(callbackFunc: (arg1: A1, callback: () => any) => any, scheduler?: SchedulerLike): (arg1: A1) => Observable<void>;
+export function bindCallback<A, R>(callbackFunc: (...args: Array<A | ((result: R) => void)>) => void): (...args: A[]) => Observable<R>;
+export function bindCallback<A, R>(callbackFunc: (...args: Array<A | ((...results: R[]) => void)>) => void): (...args: A[]) => Observable<R[]>;
```
cartant, line 13: issue: IDK why these are needed. The tests and dtslint tests pass without them, AFAICT. If they are needed for some reason, I think we should have dtslint tests that show why.

src/internal/observable/bindCallback.ts
```diff
@@ -4,48 +4,15 @@ import { bindCallbackInternals } from './bindCallbackInternals';
...
-export function bindCallback<A1, A2, A3, A4, A5, R1>(callbackFunc: (arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5, callback: (res1: R1) => any) => any, scheduler?: SchedulerLike): (arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5) => Observable<R1>;
-export function bindCallback<A1, A2, A3, A4, A5>(callbackFunc: (arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5, callback: () => any) => any, scheduler?: SchedulerLike): (arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5) => Observable<void>;
-
-export function bindCallback<A, R>(callbackFunc: (...args: Array<A | ((result: R) => any)>) => any, scheduler?: SchedulerLike): (...args: A[]) => Observable<R>;
-export function bindCallback<A, R>(callbackFunc: (...args: Array<A | ((...results: R[]) => any)>) => any, scheduler?: SchedulerLike): (...args: A[]) => Observable<R[]>;
-
-export function bindCallback(callbackFunc: (...args: any[]) => any, scheduler?: SchedulerLike): (...args: any[]) => Observable<any>;
+export function bindCallback(callbackFunc: (...args: any[]) => void): (...args: any[]) => Observable<any>;
```
cartant, line 15: issue: As with the two signatures above, this doesn't - AFAICT - seem to be needed.

src/internal/observable/bindNodeCallback.ts
```diff
@@ -5,148 +5,18 @@ import { bindCallbackInternals } from './bindCallbackInternals';
 
 /** @deprecated resultSelector is deprecated, pipe to map instead */
 export function bindNodeCallback(
-  callbackFunc: (...args: any[]) => any,
-  resultSelector: (...args: any[]) => any,
+  callbackFunc: (...args: any[]) => void,
+  resultSelector: (...args: any[]) => void,
```
cartant, line 9: nitpick: `resultSelector` should return `any`.

src/internal/observable/bindNodeCallback.ts
```diff
@@ -5,148 +5,18 @@ import { bindCallbackInternals } from './bindCallbackInternals';
...
-  scheduler?: SchedulerLike
-): (arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5) => Observable<void>;
+// args is the arguments array and we push the callback on the rest tuple since the rest parameter must be last (only item) in a parameter list
+export function bindNodeCallback<A extends readonly unknown[], R extends readonly unknown[]>(
+  callbackFunc: (...args: [...A, (err: any, ...res: R) => void]) => void,
+  schedulerLike?: SchedulerLike
+): (...arg: A) => Observable<R extends [] ? void : R extends [any] ? R[0] : R>;
 
-export function bindNodeCallback(callbackFunc: (...args: any[]) => any, scheduler?: SchedulerLike): (...args: any[]) => Observable<any[]>;
+export function bindNodeCallback(callbackFunc: (...args: any[]) => void, scheduler?: SchedulerLike): (...args: any[]) => Observable<any[]>;
```
cartant, line 19: issue: AFAICT, the tests and dtslint tests pass without this signature. If it's needed for something, we should add some dtslint tests to express why.

cartant (changes requested):

src/internal/observable/bindCallback.ts
kolodny: Fixed

src/internal/observable/bindCallback.ts
kolodny: Removed, hopefully this doesn't inadvertently break someone

src/internal/observable/bindCallback.ts
kolodny: Removed

src/internal/observable/bindNodeCallback.ts
kolodny: Good catch

src/internal/observable/bindNodeCallback.ts
kolodny: Done

kolodny:
> Approved with a small nit about adding some explanatory comments.

Did you still want comments? I'm still not 100% what they should address? I can try to explain the magic tuple rest thing. I just want to make sure I'm not missing anything

cartant (approved):

benlesh:
@kolodny can you rebase this and squash/merge?

PR #5805, opened 2020-10-07
fix(mergeScan): No longer emits state again upon completion.

benlesh:
Fixes an issue were accumulated state would be emitted a second time upon completion.

- Also ensures `state` is cleared on teardown for `mergeScan`.
- Minor refactoring to `mergeInternals` related to this change
- Refactors `mergeMap` to be less verbose in the deprecated path.

BREAKING CHANGE: `mergeScan` will no longer emit its inner state again upon completion.

related #5372


cartant (approved):
LGTM. I pushed a commit to remove an unused import that was effecting a lint failure.

benlesh:
Thanks, @cartant! I was hustling to get this done before doing "dad duties" for the evening. haha.

PR #5806, opened 2020-10-07
Fix Issue 4532 subscribe no longer takes subscribable, forkJoin has N-arguments typings.

benlesh:
Refactors to remove bad usage of `Subscribable` and updates the `InputObservable` type.

In the process I ended up needing to fix `mergeAll`, `concatAll` and `forkJoin`.

`bufferToggle` also had one bad dtslint test.

Related, I filed this in TS trying to figure this out: https://github.com/microsoft/TypeScript/issues/41091

### DIFF
--- a/api_guard/dist/types/operators/index.d.ts
+++ b/api_guard/dist/types/operators/index.d.ts
@@ -1,4 +1,4 @@
-export declare function audit<T>(durationSelector: (value: T) => SubscribableOrPromise<any>): MonoTypeOperatorFunction<T>;
+export declare function audit<T>(durationSelector: (value: T) => ObservableInput<any>): MonoTypeOperatorFunction<T>;
 
 export declare function auditTime<T>(duration: number, scheduler?: SchedulerLike): MonoTypeOperatorFunction<T>;
 
@@ -10,7 +10,7 @@ export declare function bufferTime<T>(bufferTimeSpan: number, scheduler?: Schedu
 export declare function bufferTime<T>(bufferTimeSpan: number, bufferCreationInterval: number | null | undefined, scheduler?: SchedulerLike): OperatorFunction<T, T[]>;
 export declare function bufferTime<T>(bufferTimeSpan: number, bufferCreationInterval: number | null | undefined, maxBufferSize: number, scheduler?: SchedulerLike): OperatorFunction<T, T[]>;
 
-export declare function bufferToggle<T, O>(openings: SubscribableOrPromise<O>, closingSelector: (value: O) => SubscribableOrPromise<any>): OperatorFunction<T, T[]>;
+export declare function bufferToggle<T, O>(openings: ObservableInput<O>, closingSelector: (value: O) => ObservableInput<any>): OperatorFunction<T, T[]>;
 
 export declare function bufferWhen<T>(closingSelector: () => ObservableInput<any>): OperatorFunction<T, T[]>;
 
@@ -63,7 +63,7 @@ export declare function concatWith<T, A extends ObservableInput<any>[]>(...other
 
 export declare function count<T>(predicate?: (value: T, index: number) => boolean): OperatorFunction<T, number>;
 
-export declare function debounce<T>(durationSelector: (value: T) => SubscribableOrPromise<any>): MonoTypeOperatorFunction<T>;
+export declare function debounce<T>(durationSelector: (value: T) => ObservableInput<any>): MonoTypeOperatorFunction<T>;
 
 export declare function debounceTime<T>(dueTime: number, scheduler?: SchedulerLike): MonoTypeOperatorFunction<T>;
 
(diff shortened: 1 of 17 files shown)


spec-dtslint/observables/forkJoin-spec.ts
```diff
@@ -99,6 +99,6 @@ describe('forkJoin([])', () => {
   });
 
   it('should force user cast for array of 6+ observables', () => {
-    const res = forkJoin([of(1, 2, 3), of('a', 'b', 'c'), of(1, 2, 3), of(1, 2, 3), of(1, 2, 3), of(1, 2, 3), of(1, 2, 3)]); // $ExpectType Observable<(string | number)[]>
+    const res = forkJoin([of(1, 2, 3), of('a', 'b', 'c'), of(1, 2, 3), of(1, 2, 3), of(1, 2, 3), of(1, 2, 3), of(1, 2, 3)]); // $ExpectType Observable<[number, string, number, number, number, number, number]>
```
kolodny, line 102: That's so cool!

api_guard/dist/types/operators/index.d.ts
```diff
@@ -169,7 +169,7 @@ export declare function merge<T, T2, T3, T4, T5, T6>(v2: ObservableInput<T2>, v3
 export declare function merge<T>(...observables: Array<ObservableInput<T> | SchedulerLike | number>): MonoTypeOperatorFunction<T>;
 export declare function merge<T, R>(...observables: Array<ObservableInput<any> | SchedulerLike | number>): OperatorFunction<T, R>;
 
-export declare function mergeAll<T>(concurrent?: number): OperatorFunction<ObservableInput<T>, T>;
+export declare function mergeAll<O extends ObservableInput<any>>(concurrent?: number): OperatorFunction<O, ObservedValueOf<O>>;
```
kolodny, line 172: Does this do the correct tuple thing?

spec-dtslint/operators/bufferToggle-spec.ts
```diff
@@ -23,7 +23,8 @@ it('should enforce types', () => {
 });
 
 it('should enforce type of openings', () => {
-  const o = of(1, 2, 3).pipe(bufferToggle('a', () => of('a', 'b', 'c'))); // $ExpectError
+  const o = of(1, 2, 3).pipe(bufferToggle('a', () => of('a', 'b', 'c'))); // $ExpectType Observable<number[]>
+  const o2 = of(1, 2, 3).pipe(bufferToggle('a', (x: number) => of('a', 'b', 'c'))); // $ExpectError
```
kolodny, line 27: Is this a regression or was it always supposed to word but limited by TS?

src/internal/observable/forkJoin.ts
```diff
@@ -1,68 +1,26 @@
...
-export function forkJoin<A, B, C, D, E, F>(
-  sources: [ObservableInput<A>, ObservableInput<B>, ObservableInput<C>, ObservableInput<D>, ObservableInput<E>, ObservableInput<F>]
-): Observable<[A, B, C, D, E, F]>;
-export function forkJoin<A extends ObservableInput<any>[]>(sources: A): Observable<ObservedValueUnionFromArray<A>[]>;
+// forkJoin(a, b, c)
+/** @deprecated Use the version that takes an array of Observables instead */
+export function forkJoin<A extends readonly unknown[]>(...sources: [...ObservableInputTuple<A>]): Observable<A>;
 
-// forkJoin({})
+// forkJoin({a, b, c})
```
kolodny, line 17: Does this type correctly? Maybe add a dtslint for this?

src/internal/observable/iif.ts
```diff
@@ -90,8 +90,8 @@ import { SubscribableOrPromise } from '../types';
  */
 export function iif<T = never, F = never>(
```
kolodny, line 91: Can you remove the ` = never` bit?

src/internal/operators/concatAll.ts
```diff
@@ -61,6 +61,6 @@ export function concatAll<R>(): OperatorFunction<any, R>;
  * @return {Observable} An Observable emitting values from all the inner
  * Observables concatenated.
  */
-export function concatAll<T>(): OperatorFunction<ObservableInput<T>, T> {
-  return mergeAll<T>(1);
+export function concatAll<O extends ObservableInput<any>>(): OperatorFunction<O, ObservedValueOf<O>> {
```
kolodny, line 64: Make sure to circle back with the Google folks to see how safe it is to change the generic here. Maybe keep this in mind for the next core meeting

kolodny (approved):

api_guard/dist/types/operators/index.d.ts
benlesh: Well, there's no tuples needed in the output... so it gets the proper union type out of the other side, yeah.

spec-dtslint/operators/bufferToggle-spec.ts
benlesh: Before `bufferToggle` was only taking `SubscribableOrPromise`, which isn't accurate. It should have been `ObservableInput`. Once that was in, `string`, which is an `Iterable`, was valid, making the above test inaccurate.

Then I added the bit to make sure the type passed to the closing selector was accurate.

src/internal/observable/iif.ts
benlesh: Let's tackle this in another pass. There's a lot of about `iif`'s type that I think is wrong.

src/internal/operators/concatAll.ts
benlesh: That's sort of the point of this major, right? To get all of the types correct. That's the breaking change.

PR #5814, opened 2020-10-09
docs(version-6): delete version 6 dedicated docs

niklas-wortmann:
This deletes the "migration" information from version 5 to 6. This is just affecting the version 7 docs. I also added firebase redirects pointing to a potential subdomain v6.rxjs.dev. Might need to be revisited when subdomains are in place. All the still relevant information should be in the general guide section or the WIP deprecations pages


cartant (approved):

benlesh:
@niklas-wortmann looks like this needs to be rebased.

niklas-wortmann:
rebased

PR #5816, opened 2020-10-09
Negative tests for skip, skipLast, take and takeLast

jakovljevic-mladen:
Adding tests for `skip`, `skipLast`, `take` and `takeLast` when provided with negative `count` values.


cartant (approved):
LGTM. Thanks.

PR #5819, opened 2020-10-11
Update debounceTime tests to run mode

jakovljevic-mladen:
Updated `debounceTime` tests to use run mode.

### DIFF
--- a/spec/operators/debounceTime-spec.ts
+++ b/spec/operators/debounceTime-spec.ts
@@ -1,155 +1,192 @@
+/** @prettier */
 import { expect } from 'chai';
 import { of, Subject } from 'rxjs';
 import { debounceTime, mergeMap } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
-import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
+import { observableMatcher } from '../helpers/observableMatcher';
 import { VirtualTimeScheduler } from '../../src/internal/scheduler/VirtualTimeScheduler';
 
-declare const rxTestScheduler: TestScheduler;
-
 /** @test {debounceTime} */
-describe('debounceTime operator', () => {
-  it('should debounce values by 20 time units', () => {
-    const e1 =   hot('-a--bc--d---|');
-    const expected = '---a---c--d-|';
+describe('debounceTime', () => {
+  let testScheduler: TestScheduler;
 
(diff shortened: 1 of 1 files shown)


spec/operators/debounceTime-spec.ts
```diff
@@ -1,163 +1,200 @@
...
+  });
 
-    expectObservable(e1.pipe(debounceTime(20, rxTestScheduler))).toBe(expected);
+  it('should debounce values by 2 time units', () => {
+    testScheduler.run(({ hot, expectObservable, expectSubscriptions }) => {
+      const e1 = hot('  -a--bc--d---|');
+      const e1subs = '  ^-----------!';
+      const expected = '---a---c--d-|';
+
+      expectObservable(e1.pipe(debounceTime(2, testScheduler))).toBe(expected);
```
tmair, line 23: I think the time units could also be specified by using the `time` helper function:
```suggestion
    testScheduler.run(({ hot, expectObservable, expectSubscriptions, time }) => {
      const e1 = hot('  -a--bc--d---|');
      const e1subs = '  ^-----------!';
      const expected = '---a---c--d-|';
      const timeUnit = '--|';

      expectObservable(e1.pipe(debounceTime(timeUnit, testScheduler))).toBe(expected);
```

jakovljevic-mladen: Oh, cool, I didn't know about this new property. When was it added?
I'll update PR, it makes sense to put these changes in the same PR.

tmair: I don't know when it was added. 
I just spotted a mistake in my suggestion. It should be `const timeUnit = time('--|');` instead of `const timeUnit = '--|';`

jakovljevic-mladen: I searched git and found that `time` is an alias for `createTime` which was added almost 5 years ago cb8cf6be 😄
`time` was added with commit 54d5a811 last year. We should probably have some docs related to `time` helper function. I'll create an issue.

> I just spotted a mistake in my suggestion. It should be const timeUnit = time('--|'); instead of const timeUnit = '--|';

Yeah, I saw that. I will refer to other tests to see how they solved issue, no worries. Thanks a lot for pointing this out to me.

benlesh (approved):

benlesh:
Awesome! Thanks, @jakovljevic-mladen!

PR #5821, opened 2020-10-11
Deprecation notes

### LINKED ISSUE #4776: Create documentation for deprecations and breaking changes and add URLs messages
My understanding is that in v6 and v7 there are going to be API deprecations. TSDoc decprecations have been already been added for some of these and they are causing some confusion.

For example, in [this issue](https://github.com/ReactiveX/rxjs/issues/4772) users of the package gained the mistaken impression that `of` was deprecated and in [this issue](https://github.com/ReactiveX/rxjs/issues/4772), the user was unsure of what was deprecated with `startWith(null)`. In both of these situations, the deprecation messages were due to subtleties with TypeScript's signature matching mechanism.

For contributors, it might be obvious what is and what is not deprecated, but it's understandable that users of the package might be confused by deprecation messages that are reported after a patch upgrade - particular when the deprecation message seems to related to 'normal' use.

We could make this somewhat clearer by creating a simple markdown document for each deprecation that lists what is deprecated and why and could add a link to that document in the deprecation message itself. Doing so should not require too much effort and should go some way to curtailing the opening of deprecation-related issues.

@benlesh @JWO719

niklas-wortmann:
This are the important deprecation notes from my point of view. We introduced some breaking changes recently, but I personally would keep these information in the changelog, those feel very edgy to me.

closes #4776 ?

### DIFF
--- a/docs_app/content/deprecations/resultSelector.md
+++ b/docs_app/content/deprecations/resultSelector.md
@@ -1,0 +1,70 @@
+# ResultSelector Parameter
+
+Some operator supported a resultSelector argument that acted as mapping function on the result of that operator.
+The same behavior can be reproduced with the `map` operator, therefore this argument became deprecated.
+
+<div class="alert is-important">
+    <span>
+        This deprecation was introduced in RxJS 6.0 and will become breaking with RxJS 8.
+    </span>
+</div>
+
+## Operators affected by this Change
+
+- [bindCallback](/api/index/function/bindCallback)
+- [bindNodeCallback](/api/index/function/bindNodeCallback)
+- [combineLatest](/api/index/function/combineLatest)
+- [forkJoin](/api/index/function/forkJoin)
+- [fromEvent](/api/index/function/fromEvent)
+- [fromEventPattern](/api/index/function/fromEventPattern)
+- [zip](/api/index/function/zip)
+- [fromEventPattern](/api/index/function/fromEventPattern)
+- [concatMap](/api/operators/concatMap)
+- [concatMapTo](/api/operators/concatMapTo)
+- [exhaustMap](/api/operators/exhaustMap)
+- [mergeMap](/api/operators/mergeMap)
+- [mergeMapTo](/api/operators/mergeMapTo)
+- [switchMap](/api/operators/switchMap)
+- [swithMapTo](/api/operators/swithMapTo)
+
+## How to Refactor
+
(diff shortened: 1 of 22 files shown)


docs_app/content/deprecations/resultSelector.md
```diff
@@ -0,0 +1,70 @@
+# ResultSelector Parameter
+
+Some operator supported a resultSelector argument that acted as mapping function on the result of that operator.
+The same behavior can be reproduced with the `map` operator, therefore this argument became deprecated.
+
```
benlesh, line 5: FYI: The reason this deprecation was added is because while having a resultSelector was good for performance during secondary mapping, in cases where secondary mapping wasn't necessary, it caused us to have to retain the outer value in memory longer than necessary and increased memory pressure.

It's also possible that we might reverse the decision on some of these where that isn't necessarily true.. `fromEvent` or `combineLatest` for example.

docs_app/content/deprecations/resultSelector.md
```diff
@@ -0,0 +1,70 @@
...
+## Operators affected by this Change
+
+- [bindCallback](/api/index/function/bindCallback)
+- [bindNodeCallback](/api/index/function/bindNodeCallback)
+- [combineLatest](/api/index/function/combineLatest)
+- [forkJoin](/api/index/function/forkJoin)
+- [fromEvent](/api/index/function/fromEvent)
+- [fromEventPattern](/api/index/function/fromEventPattern)
+- [zip](/api/index/function/zip)
+- [fromEventPattern](/api/index/function/fromEventPattern)
```
benlesh, line 21: this is in here twice

benlesh (changes requested):
Looks good other than one duplicated bit.

benlesh:
Related to discussion here: #5824

niklas-wortmann:
I updated this pr to align with #5824 and also I created a page for the breaking changes

cartant (approved):
I'm merging this 'cause it's a docs-only change, 'cause the deprecations effect an inordinate amount of confusion and need to be sorted out and because keeping a PR that's been open for 6 months in sync with the changes that accrue in the branch it targets is a PITA.

cartant:
FWIW, over the next few days, I'll be working on fixing some issues with deprecations and that will be much easier now that this PR has been merged.

PR #5825, opened 2020-10-15
TestScheduler's createTime bug fix

jakovljevic-mladen:
Hi everyone,

I found about `TestScheduler`'s `time` helper function couple of [days ago](5819) and I wanted to know when it was added and how to use it. And I think I have found a bug.

Since `time` is an alias for  `TestScheduler`'s `createTime`, I don't know what implications will this fix have, since I'm not sure how much people were using `createTime` function. It was added many years ago and I think this bug was introduced with [this PR](5060).

What's the problem? `createTime` trims marbles passed to it and I think that this trimming should happen only if the scheduler is in run mode because when in run mode, we're also ignoring spaces. For example, much like

```ts
const subs = '^   !';
```
advances time the same way this does (when not in run mode)
```ts
const subs = '^---!';
```

time progression syntax should behave the same when used with `createTime`:

```ts
const t = '  |';
```
```ts
const t = '--|';
```

But, this is not the case. The first time progression syntax gets trimmed and spaces do not progress time causing tests to fail. And I proved this by creating two commits and editing the only three tests we have that use `createTime` function (all of them are not using run mode). In the first commit, I just extracted time progression marbles aligning them with the rest of other marbles. In the second commit, I replaced minuses (`--`) with spaces (`  `) causing those three tests to fail. The last commit adds a fix by checking if the scheduler is using run mode.

@cartant, can you please review this since you were reviewing PR that introduced marbles trimming in `createTime`.

If everything looks good and if this PR gets approved, I will add some more unit tests for `createTime`.

### DIFF
--- a/spec/operators/find-spec.ts
+++ b/spec/operators/find-spec.ts
@@ -131,8 +131,9 @@ describe('find operator', () => {
     const source = hot('--a--b---c-|');
     const subs =       '^    !';
     const expected =   '-------(b|)';
+    const t =               '--|';
 
-    const duration = rxTestScheduler.createTime('--|');
+    const duration = rxTestScheduler.createTime(t);
 
     expectObservable(source.pipe(
       find((value: string) => value === 'b'),
--- a/spec/operators/find-spec.ts
+++ b/spec/operators/find-spec.ts
@@ -131,7 +131,7 @@ describe('find operator', () => {
     const source = hot('--a--b---c-|');
     const subs =       '^    !';
     const expected =   '-------(b|)';
-    const t =               '--|';
+    const t =               '  |';
 
     const duration = rxTestScheduler.createTime(t);
 
--- a/spec/operators/find-spec.ts
+++ b/spec/operators/find-spec.ts
@@ -131,7 +131,7 @@ describe('find operator', () => {
     const source = hot('--a--b---c-|');
     const subs =       '^    !';
     const expected =   '-------(b|)';
-    const t =               '  |';
+    const t =               '--|';
 
     const duration = rxTestScheduler.createTime(t);
 
--- a/spec/operators/findIndex-spec.ts
+++ b/spec/operators/findIndex-spec.ts
@@ -131,8 +131,9 @@ describe('findIndex operator', () => {
     const source = hot('--a--b---c-|');
     const subs =       '^    !';
     const expected =   '-------(x|)';
+    const t =               '--|';
 
-    const duration = rxTestScheduler.createTime('--|');
+    const duration = rxTestScheduler.createTime(t);
 
     expectObservable(source.pipe(
       findIndex((value: string) => value === 'b'),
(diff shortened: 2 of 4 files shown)


cartant:
Yeah, this is definitely a bug fix. The implementation of `createTime` in v6 doesn't trim the diagram:

https://github.com/ReactiveX/rxjs/blob/e968d930729671c7c9369a07e9c343cb61f2a58c/src/internal/testing/TestScheduler.ts#L41

So the always trimming - that was introduced in [this commit](https://github.com/ReactiveX/rxjs/commit/54d5a81158a377c5f1114fb111f76860e3cdf6c7) - would be a breaking change. The detection of run mode that's done in this PR is what's needed.

:+1:

spec/operators/find-spec.ts
```diff
@@ -131,8 +131,9 @@ describe('find operator', () => {
     const source = hot('--a--b---c-|');
     const subs =       '^    !';
     const expected =   '-------(b|)';
+    const t =               '  |';
```
cartant, line 134: nitpick: It's now technically valid for the times to be expressed with spaces, but they're easier to read with dashes. IMO, we should uses dashes here.

spec/operators/findIndex-spec.ts
```diff
@@ -131,8 +131,9 @@ describe('findIndex operator', () => {
     const source = hot('--a--b---c-|');
     const subs =       '^    !';
     const expected =   '-------(x|)';
+    const t =               '  |';
```
cartant, line 134: nitpick: Use dashes.

spec/operators/first-spec.ts
```diff
@@ -106,8 +106,9 @@ describe('Observable.prototype.first', () => {
     const source = hot('--a--b---c-|');
     const subs =       '^ !';
     const expected =   '----(a|)';
+    const t =            '  |';
```
cartant, line 109: nitpick: Use dashes.

cartant (changes requested):
LGTM, but I think the times are much easier to read if dashes are used. Save the spaces for alignment only uses. If you want to add a test that shows that spaces are counted when not in run mode, add a test to [spec/schedulers/TestScheduler-spec.ts](https://github.com/ReactiveX/rxjs/blob/master/spec/schedulers/TestScheduler-spec.ts).

spec/operators/find-spec.ts
jakovljevic-mladen: I would agree, I'll revert to dashes. Anyway, I will convert these tests to run mode.

cartant (approved):

jakovljevic-mladen:
@cartant, thanks a lot for review. I updated PR. I will add `createTime` tests in another PR, if you're OK with that?

cartant:
> I will add createTime tests in another PR, if you're OK with that?

Yeah, that's fine.

PR #5830, opened 2020-10-15
Refactor: combineLatest and forkJoin types for n-args

benlesh:
- Resolves an issue where `forkJoin([a, b, c] as const)` didn't infer properly.
- Updates `combineLatest` with new approach for n-args.

### DIFF
--- a/src/internal/observable/combineLatest.ts
+++ b/src/internal/observable/combineLatest.ts
@@ -1,6 +1,6 @@
 /** @prettier */
 import { Observable } from '../Observable';
-import { ObservableInput, SchedulerLike, ObservedValueOf, ObservedValueTupleFromArray } from '../types';
+import { ObservableInput, SchedulerLike, ObservedValueOf, ObservableInputTuple } from '../types';
 import { argsArgArrayOrObject } from '../util/argsArgArrayOrObject';
 import { Subscriber } from '../Subscriber';
 import { from } from './from';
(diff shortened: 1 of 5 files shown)


src/internal/observable/combineLatest.ts
```diff
@@ -9,7 +9,17 @@ import { Subscription } from '../Subscription';
...
+// combineLatest([a, b, c])
+export function combineLatest(sources: []): Observable<never>;
+export function combineLatest<A extends readonly unknown[]>(sources: readonly [...ObservableInputTuple<A>]): Observable<A>;
+
+// combineLatest(a, b, c)
+/** @deprecated Use the version that takes an array of Observables instead */
+export function combineLatest<A extends readonly unknown[]>(...sources: [...ObservableInputTuple<A>]): Observable<A>;
+
+// combineLatest({a, b, c})
+export function combineLatest(sourcesObject: {}): Observable<never>;
```
cartant, line 21: issue: `{}` is not the type for an empty object. It will match anything that's neither `null` nor `undefined`. Instead, this should be:

```ts
export function combineLatest(sourcesObject: { [K in any]: never }): Observable<never>;
```

This should be fixed in the `forkJoin` signature, too. You might even want to declare an `EmptyObject` type for this, so that things are a little clearer.

cartant (changes requested):
LGTM, but `{}` - as a type - does not represent an empty object.

cartant (approved):
Actually, I'll make the `{}`-related changes in a separate PR, as the `{}` signature wasn't introduced in this PR.

PR #5831, opened 2020-10-15
feat(filter): improve type inference for filter(Boolean)

josepot:
I think that the changes on this PR improve the types of the `BooleanConstructor` overload of `filter`

EDIT: I force-pushed after @cartant approved the PR because I realized that I forgot to add `-0` to the list of falsy types. Sorry about that! :grimacing:

### DIFF
--- a/spec-dtslint/operators/filter-spec.ts
+++ b/spec-dtslint/operators/filter-spec.ts
@@ -43,6 +43,12 @@ it('should support Boolean as a predicate', () => {
   const o = of(1, 2, 3).pipe(filter(Boolean)); // $ExpectType Observable<number>
   const p = of(1, null, undefined).pipe(filter(Boolean)); // $ExpectType Observable<number>
   const q = of(null, undefined).pipe(filter(Boolean)); // $ExpectType Observable<never>
+  const r = of(true).pipe(filter(Boolean)); // $ExpectType Observable<true>
+  const s = of(false as const).pipe(filter(Boolean)); // $ExpectType Observable<never>
+  const t = of(0 as const, -0 as const, 1 as const).pipe(filter(Boolean)); // $ExpectType Observable<1>
+  const u = of(0 as const, -0 as const).pipe(filter(Boolean)); // $ExpectType Observable<never>
+  const v = of('' as const, "foo" as const, "bar" as const).pipe(filter(Boolean)); // $ExpectType Observable<"foo" | "bar">
+  const w = of('' as const).pipe(filter(Boolean)); // $ExpectType Observable<never>
 });
 
 // I've not been able to effect a failing dtslint test for this situation and a
--- a/src/internal/operators/filter.ts
+++ b/src/internal/operators/filter.ts
@@ -5,8 +5,9 @@ import { OperatorSubscriber } from './OperatorSubscriber';
 
 /* tslint:disable:max-line-length */
 export function filter<T, S extends T>(predicate: (value: T, index: number) => value is S, thisArg?: any): OperatorFunction<T, S>;
-// NOTE(benlesh): T|null|undefined solves the issue discussed here: https://github.com/ReactiveX/rxjs/issues/4959#issuecomment-520629091
-export function filter<T>(predicate: BooleanConstructor): OperatorFunction<T | null | undefined, NonNullable<T>>;
+export function filter<T>(
+  predicate: BooleanConstructor
+): OperatorFunction<T, T extends null | undefined | false | 0 | -0 | 0n | '' ? never : T>;
 export function filter<T>(predicate: (value: T, index: number) => boolean, thisArg?: any): MonoTypeOperatorFunction<T>;
 /* tslint:enable:max-line-length */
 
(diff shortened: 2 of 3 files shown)


cartant (approved):
LGTM

cartant (changes requested):
Looking at this again, I think a dtstlint test needs to be added to ensure that the problem mentioned in this comment does not occur: https://github.com/ReactiveX/rxjs/issues/4959#issuecomment-520629091

See this comment which refers to the type signature that this PR changes:

https://github.com/ReactiveX/rxjs/blob/adbe65e659bbf17f6ab20a9b30fcca0e4d76af9a/src/internal/operators/filter.ts#L8-L9

josepot:
> Looking at this again, I think a dtstlint test needs to be added to ensure that the problem mentioned in this comment does not occur: [#4959 (comment)](https://github.com/ReactiveX/rxjs/issues/4959#issuecomment-520629091)

I was under the impression that [the current dtslint test](https://github.com/ReactiveX/rxjs/pull/5831/files#diff-d73e554c19d79db496b63c7915ffdfb73a9c83e9d7af9a599b670b51446f78c8R48-R59) already does that.

spec-dtslint/operators/filter-spec.ts
```diff
@@ -51,7 +51,7 @@ it('should support Boolean as a predicate', () => {
 // https://github.com/ReactiveX/rxjs/issues/4959#issuecomment-520629091
 it('should support inference from a return type with Boolean as a predicate', () => {
   interface I {
-    a: string | null;
+    a: string | null | undefined | false | 0 | -0 | 0n | '';
```
cartant, line 54: I think it would be best to not make **any changes** to this test. It's specifically about a comment that Rado made regarding an inference quirk - that no longer seems to happen, so it could have been a TypeScript bug.

Instead, add more expectations - for the different falsy types - to the test that's above this one.

cartant (changes requested):

spec-dtslint/operators/filter-spec.ts
josepot: > Instead, add more expectations - for the different falsy types - to the test that's above this one.

Yep. That makes a lot of sense. I will do that, for sure. Thanks!

josepot: Done, thanks!

cartant (approved):

src/internal/operators/filter.ts
```diff
@@ -5,8 +5,9 @@ import { OperatorSubscriber } from './OperatorSubscriber';
 
 /* tslint:disable:max-line-length */
 export function filter<T, S extends T>(predicate: (value: T, index: number) => value is S, thisArg?: any): OperatorFunction<T, S>;
-// NOTE(benlesh): T|null|undefined solves the issue discussed here: https://github.com/ReactiveX/rxjs/issues/4959#issuecomment-520629091
-export function filter<T>(predicate: BooleanConstructor): OperatorFunction<T | null | undefined, NonNullable<T>>;
+export function filter<T>(
+  predicate: BooleanConstructor
+): OperatorFunction<T, T extends null | undefined | false | 0 | -0 | 0n | '' ? never : T>;
```
benlesh, line 10: I feel like this is something that must exist elsewhere. `first()` for example. We probably need a type like:

```ts
type Truthy<T>  = T extends null | undefined | false | 0 | -0 | 0n | '' ? never : T
```

benlesh (approved):

src/internal/operators/filter.ts
josepot: What about?

```ts
type Falsy = null | undefined | false | 0 | -0 | 0n | "";

type NonFalsy<T> = T extends Falsy ? never : T;
type NonTruthy<T> = T extends Falsy ? T : never;
```

Would that be helpful anywhere else? I wouldn't mind working on that PR.

benlesh:
OH NO! haha...  This fails and worse, it's typed as `Observable<true>`... and it's obvious why once you think about it.

```ts
  const x = of(false, false, false, false).pipe(filter(Boolean)); // $ExpectType Observable<never>
```

I'm trying to work on an improvement right now.

josepot:
right, because `false` type is `boolean`, if it had been `of(false as const).pipe(filter(Boolean))` then the expected type would have been `Observable<never>`... I don't know, I won't feel bad if you revert the commit, but IMO that kinda makes sense :shrug:

benlesh:
I'm still thinking about this, really.

benlesh:
I mean... look at how bananas this is:

```ts
type Test<T>  = T extends true ? 'F' : 'T';

function test<T>(value: readonly T[]): Test<T> {
  return null!;
}

const o = test([false, false]); // $ExpectType "F" | "T"   (no, really). 
```

Same for `extends false`. I'm sure there's some principled reason for this, but it's baffling to me at the moment.

josepot:
Yup, fair enough. I agree. Please revert and let's pretend that this never happened :sweat_smile: Thanks!

benlesh:
@josepot Naw.. it's a good change... and it spurred me to do this:

https://github.com/ReactiveX/rxjs/pull/5842

PR #5833, opened 2020-10-16
test(TestScheduler): add createTime tests

jakovljevic-mladen:
Added `TestScheduler.createTime` tests for both run mode and regular mode.

### DIFF
--- a/spec/schedulers/TestScheduler-spec.ts
+++ b/spec/schedulers/TestScheduler-spec.ts
@@ -159,12 +159,6 @@ describe('TestScheduler', () => {
         scheduler.createTime('-a-b-#');
       }).to.throw();
     });
-
-    it('should ignore second stop', () => {
-      const scheduler = new TestScheduler(null!);
-      const time = scheduler.createTime('-----|--|');
-      expect(time).to.equal(50);
-    });
   });
 
   describe('createColdObservable()', () => {
@@ -770,15 +764,6 @@ describe('TestScheduler', () => {
           expect(() => {
             time('-a-b-#');
           }).to.throw();
-        });
-      });
-
-      it('should ignore second stop', () => {
-        const testScheduler = new TestScheduler(assertDeepEquals);
-
-        testScheduler.run(({ time }) => {
-          const t = time('-----|--|');
-          expect(t).to.equal(5);
         });
       });
     });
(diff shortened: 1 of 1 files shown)


jakovljevic-mladen:
@cartant, tests are related to #5825.

spec/schedulers/TestScheduler-spec.ts
```diff
@@ -140,12 +140,31 @@ describe('TestScheduler', () => {
...
       expect(() => {
         scheduler.createTime('-a-b-#');
       }).to.throw();
     });
+
+    it('should ignore second stop', () => {
+      const scheduler = new TestScheduler(null!);
+      const time = scheduler.createTime('-----|--|');
+      expect(time).to.equal(50);
+    });
```
cartant, line 167: issue: I don't think we should include a test like this. I mean, this isn't a valid marble diagram, so it should really throw. It just so happens that the `indexOf`-based implementation will pass this. It would also pass with `^#^#^|` - another completely invalid diagram. IMO, this could confuse someone reading the tests, so I think we'd be better off without it.

spec/schedulers/TestScheduler-spec.ts
```diff
@@ -724,5 +743,44 @@ describe('TestScheduler', () => {
...
+
+      it('should ignore second stop', () => {
+        const testScheduler = new TestScheduler(assertDeepEquals);
+
+        testScheduler.run(({ time }) => {
+          const t = time('-----|--|');
+          expect(t).to.equal(5);
+        });
+      });
+    });
```
cartant, line 784: issue: IMO, this isn't needed/desirable.

cartant (changes requested):
LGTM, but I don't see any value in the `-----|--|` tests.

spec/schedulers/TestScheduler-spec.ts
jakovljevic-mladen: I think I wanted to show that it will not throw if found another `!` sign like it throws here:
https://github.com/ReactiveX/rxjs/blob/adbe65e659bbf17f6ab20a9b30fcca0e4d76af9a/src/internal/testing/TestScheduler.ts#L246-L247
I agree they might confuse people, I will remove these tests.

cartant (approved):

PR #5839, opened 2020-10-20
Use `unknown` instead of `any` to catch more mistakes

OliverJAsh:
**Description:**

I recently shipped a bug to production which TS should have caught. It looked something like this: https://stackblitz.com/edit/rxjs-jlgx8q?devtoolsheight=60

```ts
import { of } from "rxjs";
import { mergeMapTo } from "rxjs/operators";

const source = of(10);

const fn = () => {};
// Expected type error, but got none.
source.pipe(mergeMapTo(fn)).subscribe();
```

The reason TS doesn't catch this is because:

- `mergeMapTo` uses `ObservableInput<any>`
- `ObservableInput<any>` uses `ArrayLike<any>`
- TS has a "bug" whereby `ArrayLike<any>` is assignable to a function https://github.com/microsoft/TypeScript/issues/18757

We can workaround this by replacing `ObservableInput<any>` with `ObservableInput<unknown>`, as @cartant suggested on Twitter: https://twitter.com/ncjamieson/status/1318501262162239489.

I had a quick go at replacing all usages across the codebase, but I ended up with a lot of failing type tests. Perhaps we will only be able to switch to `unknown` in certain places like this. WDYT?

### DIFF
--- a/api_guard/dist/types/operators/index.d.ts
+++ b/api_guard/dist/types/operators/index.d.ts
@@ -175,8 +175,8 @@ export declare function mergeMap<T, O extends ObservableInput<any>>(project: (va
 export declare function mergeMap<T, O extends ObservableInput<any>>(project: (value: T, index: number) => O, resultSelector: undefined, concurrent?: number): OperatorFunction<T, ObservedValueOf<O>>;
 export declare function mergeMap<T, R, O extends ObservableInput<any>>(project: (value: T, index: number) => O, resultSelector: (outerValue: T, innerValue: ObservedValueOf<O>, outerIndex: number, innerIndex: number) => R, concurrent?: number): OperatorFunction<T, R>;
 
-export declare function mergeMapTo<O extends ObservableInput<any>>(innerObservable: O, concurrent?: number): OperatorFunction<any, ObservedValueOf<O>>;
-export declare function mergeMapTo<T, R, O extends ObservableInput<any>>(innerObservable: O, resultSelector: (outerValue: T, innerValue: ObservedValueOf<O>, outerIndex: number, innerIndex: number) => R, concurrent?: number): OperatorFunction<T, R>;
+export declare function mergeMapTo<O extends ObservableInput<unknown>>(innerObservable: O, concurrent?: number): OperatorFunction<any, ObservedValueOf<O>>;
+export declare function mergeMapTo<T, R, O extends ObservableInput<unknown>>(innerObservable: O, resultSelector: (outerValue: T, innerValue: ObservedValueOf<O>, outerIndex: number, innerIndex: number) => R, concurrent?: number): OperatorFunction<T, R>;
 
 export declare function mergeScan<T, R>(accumulator: (acc: R, value: T, index: number) => ObservableInput<R>, seed: R, concurrent?: number): OperatorFunction<T, R>;
 
--- a/spec-dtslint/operators/mergeMapTo-spec.ts
+++ b/spec-dtslint/operators/mergeMapTo-spec.ts
@@ -54,6 +54,11 @@ it('should enforce types', () => {
   const o = of(1, 2, 3).pipe(mergeMapTo()); // $ExpectError
 });
 
+it('should enforce types of the observable parameter', () => {
+  const fn = () => {}
+  const o = of(1, 2, 3).pipe(mergeMapTo(fn)); // $ExpectError
+});
+
 it('should enforce the return type', () => {
   const o = of(1, 2, 3).pipe(mergeMapTo(p => p)); // $ExpectError
   const p = of(1, 2, 3).pipe(mergeMapTo(4)); // $ExpectError
--- a/src/internal/operators/mergeMapTo.ts
+++ b/src/internal/operators/mergeMapTo.ts
@@ -4,9 +4,9 @@ import { ObservableInput } from '../types';
 import { isFunction } from '../util/isFunction';
 
 /* tslint:disable:max-line-length */
-export function mergeMapTo<O extends ObservableInput<any>>(innerObservable: O, concurrent?: number): OperatorFunction<any, ObservedValueOf<O>>;
+export function mergeMapTo<O extends ObservableInput<unknown>>(innerObservable: O, concurrent?: number): OperatorFunction<any, ObservedValueOf<O>>;
 /** @deprecated */
-export function mergeMapTo<T, R, O extends ObservableInput<any>>(innerObservable: O, resultSelector: (outerValue: T, innerValue: ObservedValueOf<O>, outerIndex: number, innerIndex: number) => R, concurrent?: number): OperatorFunction<T, R>;
+export function mergeMapTo<T, R, O extends ObservableInput<unknown>>(innerObservable: O, resultSelector: (outerValue: T, innerValue: ObservedValueOf<O>, outerIndex: number, innerIndex: number) => R, concurrent?: number): OperatorFunction<T, R>;
 /* tslint:enable:max-line-length */
 
 /**
@@ -47,7 +47,7 @@ export function mergeMapTo<T, R, O extends ObservableInput<any>>(innerObservable
  * @return {Observable} An Observable that emits items from the given
  * `innerObservable`
  */
-export function mergeMapTo<T, R, O extends ObservableInput<any>>(
+export function mergeMapTo<T, R, O extends ObservableInput<unknown>>(
   innerObservable: O,
   resultSelector?: ((outerValue: T, innerValue: ObservedValueOf<O>, outerIndex: number, innerIndex: number) => R) | number,
   concurrent: number = Infinity


cartant:
AFAICT, the problem with the some of the other places is that there are some assertions that need to be added within the implementations. For example, in `defer`, the inferred type for what's returned by `innerFrom` is `Observable<unknown>` instead of `Observable<any>`:

https://github.com/ReactiveX/rxjs/blob/8ef4fee2b5f7fb68426661ffcc47b50ab81b145e/src/internal/observable/defer.ts#L54

So an `as Observable<ObservedValueOf<R>` is needed before the `subscribe` call.

I'm happy to approve this if you want to mark it as ready for review. Other places can be changed in separate PRs.

cartant (approved):
LGTM. Thanks for the PR.

OliverJAsh:
> So an `as Observable<ObservedValueOf<R>` is needed before the `subscribe` call.

Alternatively, we could do this, so we don't need the assertion:

```ts
export function defer<E, R extends ObservableInput<E>>(observableFactory: () => R): Observable<E> {
  return new Observable<E>((subscriber) => {
    innerFrom(observableFactory()).subscribe(subscriber);
  });
}
```

WDYT?

benlesh:
Okay, so the root of this is that `() => {}` passes for `ArrayLike<any>` because of `length` and the fact it can be indexed (even if the indices mean nothing). It's interesting to me that this is the "fix" here, because it certainly doesn't fix the larger problem.

For example:

`from(() => {})` is invalid at runtime and still passes type checks.

However, I'm not okay with walking away from `ArrayLike`, as we've always been able to handle other things. `arguments`, `NodeList`, etc. I think the real fix here is to do something where we allow "ArrayLike" but somehow exclude functions? Otherwise we need to narrow `ObservableInput` to `Array` only, which sucks a little, but it's arguably nothing people aren't used to.

cartant:
> I think the real fix here is to do something where we allow "ArrayLike" but somehow exclude functions?

I disagree. There should be no reason why we cannot use `unknown` - the real top type - in the signatures instead of `any`. If that's done, the problem goes away.

benlesh:
Oh. I'm not saying this isn't a valid fix. Just noticing the other problem.

cartant:
> Just noticing the other problem.

Yeah. I'm not sure about the [solution(s) above](https://github.com/ReactiveX/rxjs/pull/5839#issuecomment-712849752) for `defer`. I mean, it might be okay for `defer`, but the approach that uses the `ObservedValueOf` type is needed elsewhere. And I'm also very cautious when it comes to using `extends` constraints with type parameters. I'll have a look at it outside the scope of this PR, but, TBH, these parts of TS give me the heebie jeebies.

OliverJAsh:
> the approach that uses the `ObservedValueOf` type is needed elsewhere

Can you elaborate a little? If you could tell me the name of an operator that needs it, I can have a look and see what I come up with.

> I'll have a look at it outside the scope of this PR

I'm happy to help! I would like to get this issue fixed for all other operators, but I'm a bit uncomfortable sending a PR which adds a lot of type assertions. It would be good if we could find a way to avoid the need for them.

cartant:
> the approach that uses the `ObservedValueOf` type is needed elsewhere

The issue is that `ObservableInput` needs to be used as an element type for tuples, like this:

```ts
export declare function merge<Sources extends readonly ObservableInput<unknown>[]>(...sources: Sources): Observable<ObservedValueUnionFromArray<Sources>>;
```

in [this PR](https://github.com/ReactiveX/rxjs/pull/5844/files). AFAICT, there is no possibility of splitting out the element type, as done in the `defer` example. And, TBH, I am not sure about the first of the two `defer` examples. And I can't recall why we switched the second of the two `defer` examples - I'm pretty sure it would have originally looked like that.

I mentioned to Ben that I was planning on coming back to this once the N-args business is done.

> ... I'm a bit uncomfortable sending a PR which adds a lot of type assertions.

TBH, ATM, I'm more comfortable with the type assertion that I mentioned above than the between-two-type-parameters-constraint that's in the first of the two `defer` examples above. If you could point to any 'official' TypeScript typings that use this approach - or guidance from the TS team - I'd be happy to change my mind, but I have bad memories of things getting messed up with 'tricky' type parameter usage.

There is a very good changes that there are a whole bunch of operators that can have `any` constraints replaced with `unknown` and I'd be happy to approve those individually. I'm sure they'd be merged promptly, too.

cartant:
Regarding `defer`, it used to support factories which returned `void` - which made no sense: https://github.com/ReactiveX/rxjs/blob/a0bc359ebfed3005fac007f121ad9bf6a7ff6f59/src/internal/observable/defer.ts#L55-L67

(That types in that implementation are ... 🙈)

It's now a straightforward `ObservableInput`-returning factory, so we should be able to use the second of your suggestions:

https://github.com/ReactiveX/rxjs/blob/2337342948100d8620480cd4d045e67860dc4162/src/internal/observable/defer.ts#L52

If you want to make that change, that would be fine by me.

OliverJAsh:
> in [this PR](https://github.com/ReactiveX/rxjs/pull/5844/files). AFAICT, there is no possibility of splitting out the element type, as done in the `defer` example.

How about this?

```ts
type Observable<T> = { t: T };

declare function merge<Sources extends readonly unknown[]>(
    ...sources: { [K in keyof Sources]: Observable<Sources[K]> }
): Observable<Sources>;

declare const ob1: Observable<number>;
declare const ob2: Observable<string>;
// Observable<[number, string]>
const result = merge(ob1, ob2);
```

OliverJAsh:
> there is no possibility of splitting out the element type, as done in the `defer` example

Hmm, we might actually be able to do this. How about this pseudo code?

```ts
type Observable<T> = { t: T };

declare function merge<Sources extends readonly unknown[]>(
    ...sources: { [K in keyof Sources]: Observable<Sources[K]> }
): Observable<Sources[number]>;

declare const ob1: Observable<number>;
declare const ob2: Observable<string>;
// Observable<string | number>
const result = merge(ob1, ob2);
```

cartant:
Nice! It works with [variadic tuples, too](https://www.typescriptlang.org/play?ts=4.1.0-dev.20201015#code/C4TwDgpgBA8gRgZwgJwG4EM4BsIB4AqAfFALxQDeUwAXFPlAL4DcAUKJFAMoDGAFhABMArjmSkKUAHYB7AO60AFAEpSxACLpg0ZixYCI3LOmTQAZkMndgAS2mSoAWxQBzPJ2lDk3CAigQAHlqSAr4m6AJ2WCBQFgDWMrKSANoAuoQKLFBZUAB0eQgeXj60lEkA0lDW9rEQINKmXIXeCCm08EhomDi47p7N5WmMLEptiCgY2G5NPkmSQg5wKGms+obGZhZWtvZOyK49074BQSFQYRGSUTGS8XLJaRnZufmHtEl5OaUVVVA1dQ29IotUYdCbdQH9MqDBgAGi4fEEIiWwxB4y6Uz6MzmCyWhFYegMRhMUG4dgQwCg0jgAEZUZ1JrhsYtkHiCWtiaTJOTKXAAEx0sF4cnIKrOVmrInQTnchAI4SiWg8fjylD46UUkwIETAaniXauBRU6lwqm8pRMKAAektsDG9O6wtFUAAPlJ5szCCx1WcfNreXqXBBDTSTXy4bLlUjkOarTb2miGY7JM4XW6cSygA)

OliverJAsh:
Where does this leave us then? Maybe we could update all operators to do something like my 2nd `defer` example?

cartant:
> Where does this leave us then?

What we need to do is:

1. Figure out whether or not `defer` can be simplified and if it can, we should address any other operators that don't involve multiple inputs (i.e. tuples). There are a whole bunch. Searching I find `catchError`, `concatMap`, etc.

However, I recall now that one of the reasons `ObservedValueOf` was introduced was to handle situations like this:

https://github.com/ReactiveX/rxjs/blob/1e6092e98d988a300823b1e6e45a8940fff62b03/spec-dtslint/observables/defer-spec.ts#L15-L17

It's entirely possible that we _won't_ be able to replace it in all situations. Or maybe even at all.

2. Regarding multiple inputs (i.e. tuples) I've asked Ben to take a look, as it relates to his open PR.

cartant:
FWIW, I've just noticed/remembered that the mechanism you've pointed out was introduced recently. See [`combineLatest`](https://github.com/ReactiveX/rxjs/blob/1cc7a5dc035f895b08f2792bfacd0f6b7b7a27fe/src/internal/observable/combineLatest.ts#L14), which uses [`ObservableInputTuple`](https://github.com/ReactiveX/rxjs/blob/1e6092e98d988a300823b1e6e45a8940fff62b03/src/internal/types.ts#L218-L225) instead of `ObservedValueOf`. So the types declarations for observables and operators that use multiple sources should eventually be converted to use the former instead of relying upon the latter.

IMO, we will still need `ObservedValueOf` for functions that could return `Observable` unions that need be converted to an `Observable` with a union element type, as mentioned above.

PR #5842, opened 2020-10-20
refactor: Predicate types

benlesh:
Update operators with predicates to properly handle typings with `Boolean` constructor and predicates that return either always `true` or always `false`, depending on the operator.

- Updates dtslint tests.

Related #5831


src/internal/operators/last.ts
```diff
@@ -1,26 +1,26 @@
...
 import { identity } from '../util/identity';
 
-/* tslint:disable:max-line-length */
-export function last<T, D = T>(
-  predicate?: null,
-  defaultValue?: D
-): OperatorFunction<T, T | D>;
+export function last<T>(predicate: BooleanConstructor): OperatorFunction<T, TruthyTypesOf<T>>;
+export function last<T, D>(predicate: BooleanConstructor, defaultValue: D): OperatorFunction<T, TruthyTypesOf<T> | D>;
+export function last<T, D = T>(predicate?: null, defaultValue?: D): OperatorFunction<T, T | D>;
```
josepot, line 13: What about?

```suggestion
export function last<T, D = TruthyTypesOf<T>>(predicate?: null, defaultValue?: D): OperatorFunction<T, TruthyTypesOf<T> | D>;
```

I'm not sure, though :thinking:

benlesh: `last` without a predicate would just give you the last value (or the default if it's empty). And the last value could be `false` or `''`.

josepot: Ups sorry! I misread the implementation... I thought that it was doing `filter(identity)` when the predicate was not provided (which wouldn't make any sense, ofc) and it's actually doing just `identity`. Sorry about that!

src/internal/operators/takeWhile.ts
```diff
@@ -1,8 +1,13 @@
...
-import { OperatorFunction, MonoTypeOperatorFunction } from '../types';
+import { OperatorFunction, MonoTypeOperatorFunction, Falsy } from '../types';
 import { operate } from '../util/lift';
 import { OperatorSubscriber } from './OperatorSubscriber';
 
+export function takeWhile<T>(predicate: (value: T, index: number) => false, inclusive?: boolean): OperatorFunction<T, never>;
+export function takeWhile<T>(
+  predicate: BooleanConstructor,
+  inclusive?: boolean
+): OperatorFunction<T, Exclude<T, Falsy> extends never ? never : T>;
```
cartant, line 10: issue: Pretty sure that the return type can only be narrowed if `inclusive` is `false`.

cartant (changes requested):

spec-dtslint/operators/takeWhile-spec.ts
```diff
@@ -16,3 +16,15 @@ it('should support a predicate', () => {
 it('should support a predicate with inclusive option', () => {
   const o = of('foo').pipe(takeWhile(s => true, true)); // $ExpectType Observable<string>
 });
+
+it('should properly support Boolean constructor', () => {
+  const a = of(false as const, 0 as const, -0 as const, 0n as const, '' as const, null, undefined).pipe(takeWhile(Boolean)); // $ExpectType Observable<never>
+  const b = of(false as const, 0 as const, -0 as const, 0n as const, '' as const, null, undefined).pipe(takeWhile(Boolean, true)); // $ExpectType Observable<never>
```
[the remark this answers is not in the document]
benlesh: Good catch! Thanks!

src/internal/operators/every.ts
```diff
@@ -1,9 +1,19 @@
...
-import { OperatorFunction } from '../types';
+import { OperatorFunction, TruthyTypesOf } from '../types';
 import { operate } from '../util/lift';
 import { OperatorSubscriber } from './OperatorSubscriber';
 
+export function every<T, S extends T>(
+  predicate: (value: T, index: number, source: Observable<T>) => value is S,
+  thisArg?: any
+): OperatorFunction<T, S>;
+export function every<T>(predicate: BooleanConstructor, thisArg?: any): OperatorFunction<T, TruthyTypesOf<T>>;
```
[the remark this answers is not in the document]
benlesh: LOL... I blame type system brain damage.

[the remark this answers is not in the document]
benlesh: Okay, now it will infer correctly using `Boolean` for `Observable<Falsy>` -> `Observable<false>`. Unfortunately, we cannot do the opposite. This is because of a few things:  1. `NaN` isn't typable, but it's "Falsy". and 2. TMK there's no way to create a "Truthy" type. The best you can do is verify that nothing appears to be Falsy, but that's no guarantee, because `0` is a `number` and only `0 as const` is `0` as far as the type system is concerned. 🤷

src/internal/operators/takeWhile.ts
```diff
@@ -1,8 +1,15 @@
 /** @prettier */
-import { OperatorFunction, MonoTypeOperatorFunction } from '../types';
+import { OperatorFunction, MonoTypeOperatorFunction, Falsy } from '../types';
 import { operate } from '../util/lift';
 import { OperatorSubscriber } from './OperatorSubscriber';
 
+export function takeWhile<T>(predicate: (value: T, index: number) => false, inclusive?: boolean): OperatorFunction<T, never>;
```
cartant, line 6: issue: I don't think this can have a return type of `Operator<T, never>` unless `inclusive` is `false` or not specified:

```ts
export function takeWhile<T>(predicate: (value: T, index: number) => false, inclusive?: false): OperatorFunction<T, never>;
```

Even though the predicate always returns `false`, the operator is going to emit the value the caused the predicate to do so if `inclusive` is `true`.

cartant (changes requested):

src/internal/operators/takeWhile.ts
benlesh: That's correct. Good catch. Updated and added a test.

benlesh:
Rebased as well.

cartant (approved):

Issue #4891, opened 2019-06-28
Proposal: Statically typed `fromEvent`


devanshj:
**TLDR:** I have implemented a `fromEvent` alternative that is type-safe and has some additional features. And I want to know if I can open a PR. :)

---

So, as you people might know `fromEvent`'s types are not really great which is fair considering when they were written it wasn't possible to statically type it. But now we can have better types for `fromEvent`.

So in this light, I implemented rxjs-from-emitter. I recommend checking out the [readme](https://github.com/devanshj/rxjs-from-emitter/blob/master/README.md) to see all it's features, but in short these are it's two major features -
1. Correct `Observable` inferences corresponding to listener's arguments.
2. Static checking if an event exists or not.

**If this implementation is in alignment with RxJS's goals and interests, I would be more than happy to send a PR :)** I have answers to all your questions. Also don't get distracted with the change in API, it's the types that are important. To put it in the perspective runtime code is ~100 LOC where as type-level code is ~600 LOC.

Also additional to the type-safety, according to me there are some (respectfully) design-level flaws of `fromEvent` that rxjs-from-emitter solves which I have documented [here](https://github.com/devanshj/rxjs-from-emitter/blob/master/docs/solving-some-from-event-flaws.md).

PS: In case this comes across a publicity stunt to promote my package or whatever, it's not. I love RxJS and I genuinely want to make it better :)
Also I will take this opportunity to thank everyone that constantly works on RxJS to make it better!

devanshj:
@benlesh Is this any good? 🤔😅

benlesh:
I think this is really cool. I'm not sure I agree with all of the design choices, but I like the spirit of it. If we did have a design we agreed on, we'd need to have a migration plan. How do we get millions of users there gracefully over a series of major releases? What does the deprecation track look like? Are there code transformations we can do? Etc. RxJS is unfortunately way too popular to make a big breaking changes to API design without careful considerations like that.

devanshj:
I'm so happy that you find it cool! :D And I really appreciate that you asked these questions instead of just dismissing the proposal.

I think we can at least agree upon the fact that if we want type-safety `fromEvent` needs to go because it's just too dynamic. I mean consider this &ndash;

[...]

PR #5844, opened 2020-10-21
refactor(merge): Create better typings

benlesh:
- Now refers n-arguments in all cases, including arguments with tailing concurrency and Scheduler
- Adds dtslint tests

Please note the "unhappy path" in the dtslint tests. It's not ideal, but everything is _much_ better than it was before. I just struggled with how to ensure `merge(a$, 23, b$)` didn't match the signature for (pseudo-code here:) `merge(...args: ObservableInput<T>, concurrency: number)`

### DIFF
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -358,7 +358,7 @@ export declare function of<T, U>(value1: T, value2: U): Observable<T | U>;
 export declare function of<T, U, V>(value1: T, value2: U, value3: V): Observable<T | U | V>;
 export declare function of<A extends Array<any>>(...args: A): Observable<ValueFromArray<A>>;
 
-export declare type OneOrMoreUnknownObservableInputs = readonly [ObservableInput<unknown>, ...unknown[]];
+export declare type OneOrMoreUnknownObservableInputs = readonly [ObservableInput<unknown>, ...ObservableInput<unknown>[]];
 
 export declare function onErrorResumeNext(): Observable<never>;
 export declare function onErrorResumeNext<O extends ObservableInput<any>>(arrayOfSources: O[]): Observable<ObservedValueOf<O>>;
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -113,6 +113,8 @@ export declare function concat<O1 extends ObservableInput<any>, O2 extends Obser
 export declare function concat<O1 extends ObservableInput<any>, O2 extends ObservableInput<any>, O3 extends ObservableInput<any>, O4 extends ObservableInput<any>, O5 extends ObservableInput<any>, O6 extends ObservableInput<any>>(v1: O1, v2: O2, v3: O3, v4: O4, v5: O5, v6: O6, scheduler: SchedulerLike): Observable<ObservedValueOf<O1> | ObservedValueOf<O2> | ObservedValueOf<O3> | ObservedValueOf<O4> | ObservedValueOf<O5> | ObservedValueOf<O6>>;
 export declare function concat<A extends ObservableInput<any>[]>(...observables: A): Observable<ObservedValueUnionFromArray<A>>;
 
+export declare type Concat<T extends readonly unknown[], E extends readonly unknown[]> = readonly [...T, ...E];
+
 export declare const config: {
     onUnhandledError: ((err: any) => void) | null;
     Promise: PromiseConstructorLike;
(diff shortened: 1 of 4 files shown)


src/internal/types.ts
```diff
@@ -259,3 +259,8 @@ export type ValueFromNotification<T> = T extends { kind: 'N' | 'E' | 'C' }
       : undefined
     : never
   : never;
+
+/**
+ * Used to help deal with rest arguments of many observable inputs that are tailed by schedulers, etc.
+ */
+export type OneOrMoreUnknownObservableInputs = readonly [ObservableInput<unknown>, ...unknown[]];
```
cartant, line 266: issue: AFAICT, this makes things worse. The `...unknown[]` will match anything, so these don't effect errors:

```ts
const o5 = merge(a$, 1, 2, 3, 4, 5); // $ExpectError
const o6 = merge(a$, 1, 2, 3, 4, 5, asapScheduler); // $ExpectError
```

And, changing `...unknown[]` to `...ObservbleInput<unknown>[]` just makes things worse. I'll have another look at this after I've had a coffee, but ATM ... 🤷‍♂️

cartant (changes requested):

src/internal/types.ts
cartant: I don't have a decent, actionable suggestion for this problem. I still think it's a TS bug. If it weren't for the concurrency parameter, we could keep the ugly overloads for the deprecated scheduler sigs and use variadic tuples for the non-deprecated sigs. But we have the trailing concurrency, so ... 🤷‍♂️

api_guard/dist/types/index.d.ts
```diff
@@ -210,34 +210,17 @@ export declare function isObservable<T>(obj: any): obj is Observable<T>;
...
-export declare function merge<T>(...observables: (ObservableInput<T> | number)[]): Observable<T>;
-export declare function merge<T>(...observables: (ObservableInput<T> | SchedulerLike | number)[]): Observable<T>;
-export declare function merge<T, R>(...observables: (ObservableInput<any> | number)[]): Observable<R>;
-export declare function merge<T, R>(...observables: (ObservableInput<any> | SchedulerLike | number)[]): Observable<R>;
+export declare function merge<O extends ObservableInput<unknown>>(sources: O[], concurrent: number, scheduler: SchedulerLike): Observable<ObservedValueOf<O>>;
+export declare function merge<O extends ObservableInput<unknown>>(sources: O[], scheduler: SchedulerLike): Observable<ObservedValueOf<O>>;
+export declare function merge<O extends ObservableInput<unknown>>(sources: O[], concurrent: number): Observable<ObservedValueOf<O>>;
+export declare function merge<O extends ObservableInput<unknown>>(sources: O[]): Observable<ObservedValueOf<O>>;
```
cartant, line 216: thought: We need a consistent approach to this. [Elsewhere](https://github.com/ReactiveX/rxjs/blob/1cc7a5dc035f895b08f2792bfacd0f6b7b7a27fe/src/internal/observable/combineLatest.ts#L14), [`ObservableInputTuple`](https://github.com/ReactiveX/rxjs/blob/1e6092e98d988a300823b1e6e45a8940fff62b03/src/internal/types.ts#L218-L225) is used instead of `ObservedValueOf`.

benlesh: I think these are meant to solve different issues? One is to get `A  | B | C` out of `ObservableInput<A | B | C>`, and the other is used to get `[A, B, C]` out of `[ObservableInput<A>, ObservableInput<B>, ObservableInput<C>]`. Unless I'm mistaken?

src/internal/types.ts
benlesh: Ugh... you're right. I tried `...ObservableInput<unknown>[]` _first_, but it didn't work and I couldn't see why.

@DanielRossenwasser, do you have any suggestions?

benlesh: Actually, I've fixed your example, @cartant, [here](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtVIB4BleEADwxFWAGd44pg8IBPeAbRoxi1QHMANPAB0orj37sAulIB8ACgBQ8EWO68+NAILUAcsgC2AIxAwAXPRCNmbdqOFEhqQyZhTFASgt3Rj+M+NTKQBuRUUwPC54KHgAXiR5ACIoRKEARg9QiNQooziE5NT4RKMijKzIjHgwfMR5cvDK+GBa+qEAJkz4AHpu+AB3DEQAfiA)

```ts
declare function f<S extends readonly [string, ...string[]]>(
  ...stringsAndNumber: readonly [...S, number]
): [...S, number];

const a = f("a", 1);
const b = f("a", "b", 1);
const c = f(1);
const d = f(1, 2); // works fine now (with an error)
```

benlesh: Oh... but it still doesn't work in our case, once there's a generic inside of that declaration... farts.

benlesh: See [here](https://www.typescriptlang.org/play?#code/MYGwhgzhAEBC4AsA8AVAfNA3gKGtYA9gHYQAuATgK7CkHkAUADpQEYgCWw05ApmACbEQAT2hswCAFzQUASiwBfbEuz8eoML2gAzSkRrtiOpAGVoPAB6keRfjF4ChogNrwJSPQGsiBAO5E0ABpoADowt2QvH380ZwBdOLR6XFCwsnJ2IgBzCABBWwA5SgBbFh5yaQdBIhFoZzCQk2CiErLyOOxZaXqwpugW0vK4gG5sMcISUjFEAEZoAF5+nl84RHoAcjB12VGJsmmJACYFpZWI+nid8eJ9sBPtenEEGeCZq72plnvH2eCnw9e7xuUy4iweb12wOg-G+L2ghyBk3M3yecP+r2CAOgAGYdtAAPT46C+UjaAD8QA)

```ts
class Blah<T> {
  constructor(public readonly blah: T) {}
}

declare function f<S extends readonly [Blah<unknown>, ...Blah<unknown>[]]>(
  ...stringsAndNumber: readonly [...S, number]
): [...S, number];


const blah1 = new Blah('a');
const blah2 = new Blah([]);

const a = f(blah1, 1);
const b = f(blah1, blah2, 1);
const c = f(1);
const d = f(1, 2);
const e = f(blah1, blah2, 1, 2, 3); // wtf?
```

benlesh: Okay, I've pushed an update with a commit that _should work_ (I think), but doesn't. Marking this as blocked until we hear from the TS team about whether they're going to support that or not. It _looks_ like a simple bug, but maybe there's some other approach they want people to take.

benlesh:
Blocked on https://github.com/microsoft/TypeScript/issues/39595

src/internal/types.ts
cartant: The TypeScript playground link that I added above was for what is, IMO, the fundamental bug.

The solution you have [above](https://github.com/ReactiveX/rxjs/pull/5844#discussion_r511007616) will work if the tuple of type parameters to `Blah` is extracted and mapped, like this:

```ts
class Blah<T> {
  constructor(public readonly blah: T) {}
}

type Blahs<T extends readonly unknown[]> = {
  [K in keyof T]: Blah<T[K]>;
};

declare function f<S extends readonly [unknown, ...unknown[]]>(
  ...stringsAndNumber: readonly [...Blahs<S>, number]
): [...Blahs<S>, number];
````

See this [playground snippet](https://www.typescriptlang.org/play?#code/MYGwhgzhAEBC4AsA8AVAfNA3gKGtYA9gHYQAuATgK7CkHkAUADpQEYgCWw05ApmACbEQAT2hswCAFzQUASiwBfbEuylhjHnEQRU0HgA9SPIvxi8BQ0ZSIBrIgQDuRANoBdDAF4suaM4DS0OxE0DY8wgQAZjKu0vASqP7uANzKKdj8PKBgvNAR1jTsxLlIAMp6hsam3HyCRCK+1naORAA00AB0nY32Tm7u9D6d7WTkQQDmEACCJgBylAC2LDzk0ua19c5DcQg6JWhtRAtL5K7YstKbndu7+9CHi8uuadiEJKRiiACM0F5EPA5aCT0ADkYGBshSrzIHwkACYfnd-oCEPQ3BDsC9iNCwAiIvRxAhPm1Puioe8WLj8V82gTYcTSVj3lwvHiSZDGdB+JSidBYQy3npKQSebTiW06dAAMwQ6AAell0AcpAiAH4gA)

Whether that will work in the codebase, IDK. I'm pretty fed up with TS, TBH.

**Comment by devanshj** [comment]
Twitter brought me here. If I understand correctly y'all basically want to achieve something like `f(...a: [string, ...string[], number])`, here's a way you can do it. Notice the errors are pretty friendly as you would want them to be. If you like it, I can go deep and explain how it works or answer questions if y'all want.

[Playground](https://www.typescriptlang.org/play?ts=4.0.3#code/CYUwxgNghgTiAEAzArgOzAFwJYHtVIB4BBeEADwxFWAGd4BhKGjAgSVFWwwE8CBlDDCyoA5gKGiaAOWQBbAEYgYxAHwqANPCipuAbQC6agBQA6M1ABc8IgEorRANwAoJ4iMByedAAWARneavjbObp4+-pphUN4ATAHwQSFGia7JmjHBqVF+8dlxgemaAMyZbqUeUO7l7pWRleW+dVXOTjwADgjiwmKC3dJyisok5JTUdNp6hvAAvE7w8AASIFDAqqQUVLTwzBIic-PwAPzwACpQWBBrI5t0qANK+wcHxwAKOG383liILERq66MtjtugZHk9wcciGDwQcrLolitVJozCYupICG8Pnwvj9VGpNGcLqp9NDYadzpc-gCbtteqJ4KSYa9kDRvMRNHcFEoVIzwVYXiy2ZikfBOYMeU8rAjVlTrmN4GhQIhhCBgLzjrpgaIOfcYCSYfMrABVVCs74sbHmkVakQqFrtTp0kQ0YjU+UTAwqGb7AAyUAAXrwiJpGMwCABveC6ADS8GE8AA1iBuDhENZ9FYbfAAL6aD2GHmtbgdRbLGVurb5r3TeB+wPs6wVugGI4K6ggZWoVXwexNqPCRBKRbIswDocAfX0rYWPdFIAAbtznA7yUThht3TpPTNawGg5p14Dm1PjoqOyrgLPDzTdCixzB4OPNPfTifTrOu4uYHaiyXLbjr03SZq13etg0bOUthbDUp17SDmxfJ94DvVBBwfPg3z4D8FyXX8EBNM0AL7fNNAADRAut9yjUiRxMIhDGXYsEGFQDKy3KYa0ohtWOPVsWzgjcoJQtD4BeZ9UInN8Xmwr8fxXAVWVdeCtHYsiKL3BtbzMcDSIYvCGCYFgTk0I0QJOPsjVbczjUYksuJ4lTJgJED8z7EhjlcqwTlshB2CoLheBOEDIxjON8CTFM0xODNThjKdsycIA)

```typescript
declare function f<A extends Cast<Identity<StringStringsNumber<A>>, any[]>>(...a: A): A;

f('blah1', 1);
f('blah1', 'blah2', 1);
f(1);
f(1, 2);
f('blah1', 'blah2', 1, 2, 3);
f();
f('a');
f('a', 'a');
f(1, 'a');

type StringStringsNumber<A extends any[]> =
  Head<A> extends string
    ? Tail<A> extends number
        ? Pop<Shift<A>> extends string[]
            ? A
            : [Head<A>, ...Strings<Pop<Shift<A>>>, Tail<A>]
        : Tail<A> extends string 
            ? Push<A, number>
            : Push<Pop<A>, number>
    : Head<A> extends undefined
        ? [string, number]
        : Unshift<Shift<A>, string>;

type Strings<A extends any[]> =
  Lazy<A, Cast<{ [K in keyof A]: string }, any[]>>

type Head<A extends any[]> = Lazy<A, A extends [] ? undefined : A extends [infer H, ...infer _] ? H : never>;
type Tail<A extends any[]> = Lazy<A, A extends [] ? undefined : A extends [...infer _, infer T] ? T : never>;
type Shift<A extends any[]> = Lazy<A, A extends [] ? [] : A extends [infer _, ...infer S] ? S : never>;
type Unshift<A extends any[], X> = Lazy<A, [X, ...A]>;
type Pop<A extends any[]> = Lazy<A, A extends [] ? [] : A extends [...infer P, infer _] ? P : never>;
type Push<A extends any[], X> = Lazy<A, [...A, X]>;
type Cast<T, U> = T extends U ? T : U;
type Lazy<A extends any[], T> = any[] extends A ? any[] : T;
type Identity<T> = { [K in keyof T]: T[K] }
```

Not to mention no pressure to go ahead with this, I can understand it might come across as "hacky" or whatever (altho it's not imo).

Also please feel free to ping me henceforth if you want some help with some really tricky TypeScript problem, I'll try my best! :) Not sure if y'all can tell but I'm also the guy behind [this](https://github.com/ReactiveX/rxjs/issues/4891) :P

benlesh:
Closed in favor of #5859

PR #5848, opened 2020-10-23
build(package): use `types` instead of `typings` in generated alias

falsyvalues:
**Description:**

This aligns `types` prop usage in main `package.json` and generated ones via `generate-alias.js`.


cartant (approved):
Thanks for the PR. I'm fine with this change, as `"types"` is what's in [the documentation](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package), these days.

I'm curious as to what led to the PR, though: was the use of the older `"typings"` key causing problems someplace?

falsyvalues:
Hi @cartant!

I was going thru building process and how aliases being built and I found this script. Since TS docs mostly use types I wanted to align this to avoid future confusion.

benlesh:
Thanks, @falsyvalues

Issue #5838, opened 2020-10-20
Complete notifications deemed to be signals


cartant:
## Bug Report

With numerous operators, the `next` and `complete` notifications received from notifier (or selector) observables are treated in the same way. As commented [here](https://github.com/ReactiveX/rxjs/blob/8ef4fee2b5f7fb68426661ffcc47b50ab81b145e/src/internal/operators/delayWhen.ts#L141-L142), that is arguably incorrect behaviour (at the very least, IIRC, this behaviour is inconsistent throughout the API):

```ts
// TODO(benlesh): I'm inclined to say this is _incorrect_ behavior.
// A completion should not be a notification. Note the deprecation above
```

AFAICT, the following operators effect this incorrect behaviour:

- [x] [`audit`](https://github.com/ReactiveX/rxjs/blob/8ef4fee2b5f7fb68426661ffcc47b50ab81b145e/src/internal/operators/audit.ts#L80)
- [x] [`bufferToggle`](https://github.com/ReactiveX/rxjs/blob/8ef4fee2b5f7fb68426661ffcc47b50ab81b145e/src/internal/operators/bufferToggle.ts#L83)
- [x] [`bufferWhen`](https://github.com/ReactiveX/rxjs/blob/8ef4fee2b5f7fb68426661ffcc47b50ab81b145e/src/internal/operators/bufferWhen.ts#L72)
- [x] [`debounce`](https://github.com/ReactiveX/rxjs/blob/8ef4fee2b5f7fb68426661ffcc47b50ab81b145e/src/internal/operators/debounce.ts#L100)
- [x] [`delayWhen`](https://github.com/ReactiveX/rxjs/blob/8ef4fee2b5f7fb68426661ffcc47b50ab81b145e/src/internal/operators/delayWhen.ts#L136-L144) (Nope, `delayWhen` is fine. It seems to have been refactored in another PR.)
- [x] [`sample`](https://github.com/ReactiveX/rxjs/blob/8ef4fee2b5f7fb68426661ffcc47b50ab81b145e/src/internal/operators/sample.ts#L64)
- [x] [`throttle`](https://github.com/ReactiveX/rxjs/blob/8ef4fee2b5f7fb68426661ffcc47b50ab81b145e/src/internal/operators/throttle.ts#L84)
[...]

PR #5853, opened 2020-10-25
fix: don't treat notifier completion as a signal

cartant:
**Description:**

This PR changes the behaviour of `audit`, `bufferToggle`, `bufferWhen`, `debounce`, `sample`, `throttle` and `windowToggle` so the that the `complete` notifications received from notifier/duration-selector observables are no longer treated as signals.

BREAKING CHANGES: the behaviour of the above-mentioned operators has changed for empty notifiers/selectors.

**Related issue (if exists):** #5838

### DIFF
--- a/src/internal/operators/audit.ts
+++ b/src/internal/operators/audit.ts
@@ -23,7 +23,7 @@ import { OperatorSubscriber } from './OperatorSubscriber';
  * Initially, the timer is disabled. As soon as the first source value arrives,
  * the timer is enabled by calling the `durationSelector` function with the
  * source value, which returns the "duration" Observable. When the duration
- * Observable emits a value or completes, the timer is disabled, then the most
+ * Observable emits a value, the timer is disabled, then the most
  * recent source value is emitted on the output Observable, and this process
  * repeats for the next source value.
  *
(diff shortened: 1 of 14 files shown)


src/internal/operators/audit.ts
```diff
@@ -57,7 +57,7 @@ export function audit<T>(durationSelector: (value: T) => ObservableInput<any>):
     let durationSubscriber: Subscriber<any> | null = null;
     let isComplete = false;
 
-    const endDuration = () => {
+    const signalDuration = () => {
```
benlesh, line 60: Nit: I'm not sure this name change is better. This is to signal the end of the duration. "signalDuration" could be signalling the start?

benlesh (approved):
Approved. This looks great.

The only thing I'd really like you to do @cartant is to clean up the commits, flattening out the "chore" type commits and breaking it into just the relevant fixes, such that each fix with a breaking change has a `BREAKING CHANGE: ...` comment in the commit message. Then just rebase and merge here on GitHub. 

This way, it will show up in the CHANGELOG when I generate it at release time.

cartant:
@benlesh Done. I updated the docs, too. I'd missed those and they mentioned complete notifications.

cartant:
@benlesh I'm not sure about the change I made to `bufferToggle`. If the closing notifier completes, the buffer is discarded. I don't know whether it should do that or should keep the buffer open - accumulating values - to be emitted when/if the source completes. That's what happens with `windowToggle` - unclosed windows stay open until the source completes. What do you think should happen?

benlesh:
@cartant `bufferX` should always be the same as `windowX` with a `toArray`... `windowX(), mergeMap(toArray())` basically.

cartant:
@benlesh Done.

benlesh (approved):

benlesh:
Nice. Thanks!

PR #5857, opened 2020-10-26
refactor(concat): support N-args with scheduler

cartant:
**Description:**

This PR refactors the static `concat` to use N-args in the signature that accepts a trailing scheduler. And adds a bunch of this-usage-should-fail dtslint tests.

For whatever reason, the problems that occurred in the PRs that were opened for the static `merge` and `combineLastest` functions do not seem to occur here. A lot was going on in those PRs - and the conversations that took place in those PRs aren't especially easy to follow - but I would not be surprised if there is some small difference that effects different behaviour. For example, temporarily adding a signature that included a concurrency (for the purposes of experimentation) seemed to break things, so ... 🤷‍♂️ this seems fine here, but we might still have problems elsewhere. 🤞

### DIFF
--- a/src/internal/observable/concat.ts
+++ b/src/internal/observable/concat.ts
@@ -4,8 +4,8 @@ import { concatAll } from '../operators/concatAll';
 import { internalFromArray } from './fromArray';
 import { popScheduler } from '../util/args';
 
-export function concat<T extends readonly unknown[]>(...inputsAndScheduler: [...ObservableInputTuple<T>, SchedulerLike]): Observable<T[number]>;
 export function concat<T extends readonly unknown[]>(...inputs: [...ObservableInputTuple<T>]): Observable<T[number]>;
+export function concat<T extends readonly unknown[]>(...inputsAndScheduler: [...ObservableInputTuple<T>, SchedulerLike]): Observable<T[number]>;
 
 /**
  * Creates an output Observable which sequentially emits all values from the first given
--- a/src/internal/observable/concat.ts
+++ b/src/internal/observable/concat.ts
@@ -4,7 +4,7 @@ import { concatAll } from '../operators/concatAll';
 import { internalFromArray } from './fromArray';
 import { popScheduler } from '../util/args';
 
-export function concat<T extends [unknown, ...unknown[]]>(...inputsAndScheduler: [...ObservableInputTuple<T>, SchedulerLike]): Observable<T[number]>;
+export function concat<T extends readonly unknown[]>(...inputsAndScheduler: [...ObservableInputTuple<T>, SchedulerLike]): Observable<T[number]>;
 export function concat<T extends readonly unknown[]>(...inputs: [...ObservableInputTuple<T>]): Observable<T[number]>;
 
 /**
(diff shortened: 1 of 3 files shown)


src/internal/observable/concat.ts
```diff
@@ -1,26 +1,12 @@
...
-/** @deprecated The scheduler argument is deprecated, use scheduled and concatAll. Details: https://rxjs.dev/deprecations/scheduler-argument */
-export function concat<O1 extends ObservableInput<any>, O2 extends ObservableInput<any>, O3 extends ObservableInput<any>, O4 extends ObservableInput<any>, O5 extends ObservableInput<any>, O6 extends ObservableInput<any>>(v1: O1, v2: O2, v3: O3, v4: O4, v5: O5, v6: O6, scheduler: SchedulerLike): Observable<ObservedValueOf<O1> | ObservedValueOf<O2> | ObservedValueOf<O3> | ObservedValueOf<O4> | ObservedValueOf<O5> | ObservedValueOf<O6>>;
+export function concat<T extends readonly [unknown, ...unknown[]]>(...inputsAndScheduler: readonly [...ObservableInputTuple<T>, SchedulerLike]): Observable<T[number]>;
+export function concat<T extends readonly unknown[]>(...inputs: readonly [...ObservableInputTuple<T>]): Observable<T[number]>;
```
kolodny, line 8: This seems to work for me in my very basic testing. I may be missing some edge cases I didn't consider

```ts
export function concat<T extends readonly unknown[]>(...inputsAndScheduler: readonly [...ObservableInputTuple<T>, SchedulerLike]): Observable<T[number]>;
export function concat<T extends readonly unknown[]>(...inputsAndScheduler: readonly [...ObservableInputTuple<T>]): Observable<T[number]>;
```

cartant: You'll need to make sure that it effects errors properly without the `[unknown, ...unknown]` that's the main reason that it was introduced. See [this comment](https://github.com/ReactiveX/rxjs/pull/5844#discussion_r510489262) and the TypeScript playground linked therein. IMO, it's a TS bug.

kolodny: I'm not quite sure I understand how it breaks in the snippet above but it sounds like we want to disallow an empty inferred tuple. If so we can change it to

```ts
export function concat<T extends readonly unknown[]>(...inputsAndScheduler: [] extends T ? never : readonly [...ObservableInputTuple<T>, SchedulerLike]): Observable<T[number]>;
export function concat<T extends readonly unknown[]>(...inputsAndScheduler: [] extends T ? never : readonly [...ObservableInputTuple<T>]): Observable<T[number]>;
```

But I may be missing the boat completely, lol

benlesh (approved):

cartant:
@benlesh FYI, I've pushed some post-approval commits to re-order the signatures so that the signature with the trailing scheduler comes last. Here, it doesn't matter, but if there are more than two trailing arguments, it matters a lot. I've re-ordered these to, hopefully, not give anyone else the wrong idea regarding the ordering of the sigs if this is used as the basis/example for other refactors, etc.

benlesh (approved):

PR #5860, opened 2020-10-27
chore: update every tests to run mode

jakovljevic-mladen:
Updated `every` tests to use run mode.

### DIFF
--- a/spec/operators/every-spec.ts
+++ b/spec/operators/every-spec.ts
@@ -1,274 +1,365 @@
+/** @prettier */
 import { expect } from 'chai';
 import { every, mergeMap } from 'rxjs/operators';
-import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
+import { TestScheduler } from 'rxjs/testing';
 import { of, Observable, Observer } from 'rxjs';
+import { observableMatcher } from '../helpers/observableMatcher';
 
 /** @test {every} */
-describe('every operator', () => {
+describe('every', () => {
+  let testScheduler: TestScheduler;
+
+  beforeEach(() => {
+    testScheduler = new TestScheduler(observableMatcher);
+  });
+
   function truePredicate(x: number | string) {
     return true;
   }
 
   function predicate(x: number | string) {
-    return (+x) % 5 === 0;
+    return +x % 5 === 0;
   }
 
-  it('should return false if only some of element matches with predicate', () => {
-    const source = hot('--a--b--c--d--e--|', {a: 5, b: 10, c: 15, d: 18, e: 20});
-    const sourceSubs = '^          !      ';
-    const expected =   '-----------(F|)   ';
+  it('should return false if only one of elements does not match with predicate', () => {
+    testScheduler.run(({ hot, expectObservable, expectSubscriptions }) => {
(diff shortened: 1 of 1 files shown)


cartant:
I saw the PR title and thought you'd updated every operator in a single PR 😲😂

spec/operators/every-spec.ts
```diff
@@ -1,274 +1,365 @@
...
 
-    expectObservable(source.pipe(every(predicate))).toBe(expected, {x: true});
-    expectSubscriptions(source.subscriptions).toBe(sourceSubs);
+      expectObservable(e1.pipe(every(predicate))).toBe(expected, { x: true });
+      expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    });
   });
 
-  it('should emit true if Scalar source matches with predicate', () => {
-    const source = of(5);
```
cartant, line 160: issue: Probably best to not change any sources that are currently declared using `of()` to `cold()`, as there is a subtle difference - see #5760 and its associated issue for background.

Is this something that you've done in other PRs? I'm wondering whether or not I missed it in any of the other reviews.

cartant (changes requested):
This is all okay, but I don't think we should be changing `of()` to `cold()`. See my comment below.

spec/operators/every-spec.ts
jakovljevic-mladen: OK, I will change this, but I have been changing this in my other PRs as well. Should I revert from other tests?

jakovljevic-mladen: Oh, didn't see that you already asked... Yes, I was changing in other PRs as well. I'll make a revert PR.

jakovljevic-mladen: @cartant I started reading #5760 and I found it to be quite large to me. I'll give it try next time, but for now, out of curiosity, I'd like to know what would e.g. `cold('a')` represent? This:
`new Observable(subscriber => subscriber.next('a'))`
or this:
`new Observable(subscriber => setTimeout(() => subscriber.next('a'), 0))`
Is #5760 trying change behavior from one to another maybe?

cartant: `cold('a')` essentially represents `new Observable(subscriber => asapScheduler.schedule(() => subscriber.next('a')))`.

jakovljevic-mladen: Updated PR, I think I reverted everything. Other reverts coming...
I didn't know about this subtle distinction. Thanks a lot for explanations Nicholas. Any chance to explain why `asapScheduler` is used for completion notifications? If easier, you can post a link to an issue or to a source code.
Is completion notification scheduled on `asapScheduler` only for frame 0 or for every frame?

cartant: Thanks for the update.

Actually, what I said above was wrong. [All notifications are emitted on the test scheduler](https://github.com/ReactiveX/rxjs/blob/1f021fe280e707a6bfbd36dcc8880d17c3e64fd2/src/internal/testing/ColdObservable.ts#L35-L50), so it's kinda equivalent to:

```ts
  asapScheduler.schedule(() => subscriber.next('a'));
  asapScheduler.schedule(() => subscriber.complete());
})
```

cartant (approved):
LGTM. Thanks.

PR #5865, opened 2020-10-30
fix(dependencies): update symbol-observable

EtDu:
Latest version respects environments where `Symbol` may be frozen

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -160,9 +160,9 @@
       "dev": true
     },
     "@types/node": {
-      "version": "9.4.5",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-9.4.5.tgz",
-      "integrity": "sha512-DvC7bzO5797bkApgukxouHmkOdYN2D0yL5olw0RncDpXUa6n39qTVsUi/5g2QJjPgl8qn4zh+4h0sofNoWGLRg==",
+      "version": "14.6.0",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-14.6.0.tgz",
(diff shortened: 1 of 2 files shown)


kumavis:
heres the CI failure
```
Error: node_modules/@types/node/index.d.ts(74,11): error TS2300: Duplicate identifier 'IteratorResult'.
Error: node_modules/typescript/lib/lib.es2015.iterable.d.ts(41,6): error TS2300: Duplicate identifier 'IteratorResult'.
```

kumavis:
this issue suggests updating `@types/node` https://github.com/microsoft/TypeScript/issues/32333

benlesh:
This is all very tricky and requires a bit of thought and some known test cases.

`@types/node@14.14.3` and higher will no longer define `Symbol.observable`, which is a good thing.

`rxjs@7` (not what this targets) and `symbol-observable@3.+` correctly define `Symbol.observable` as a `unique symbol`.

however, `rxjs@6` is incompatible with `@types/node` at certain versions (as this PR is trying to fix that), but if we fix it, we'll potentially break other `rxjs` users that are using older versions of `@types/node` with patch release of `6.x`.

benlesh:
Okay... the only change here I'm worried about is the update to the `tslib` dependency... which, while it _should_ be fine, might constitute a breaking change for someone. (This is the problem with libraries that are used by a lot of people).

kumavis:
@benlesh would you consider merging and reverting if it becomes an issue?

benlesh:
The dependency on `symbol-observable` was removed in https://github.com/ReactiveX/rxjs/commit/97ee666482a7f4e9d1c91072c2f0497d01f2abb9

benlesh:
Okay, revisiting this. The update to the `symbol-observable` dev dependency doesn't add a lot of value here, and the update to the runtime dependency may constitute a breaking change. Given than 6.x is stable/support mode, we can't merge this. Sorry, @EtDu... and sorry to take so long on this decision.

PR #5866, opened 2020-10-30
chore: update exhaust tests to run mode

jakovljevic-mladen:
Updated `exhaust` tests to use run mode.

@cartant, I've got two issues/questions here:

- The first test `should handle a hot observable of hot observables` is exactly the same as test `should handle a hot observable of observables` with the exception of `expectSubscriptions` calls. I was usually adding `expectSubscriptions` calls for those tests that didn't have them and wanted to do the same for this test, but then later I found out that they're the same. What I'd suggest is to transform this test to the one which would really satisfy what it says in the title (now, it doesn't handle observable of _hot_ observables, but rather _cold_ ones). I'd add `expectSubscriptions` where possible. Would you agree? If yes, is it OK to do it with this PR?
- I'd inline empty arrays. E.g. instead of:

```ts
const y = cold('---d--e---f---|    ');
const ysubs: string[] = [];
const z = cold('     ---g--h--i---|');
expectSubscriptions(y.subscriptions).toBe(ysubs);
```
use like this:
```ts
const y = cold('---d--e---f---|    ');
const z = cold('     ---g--h--i---|');
expectSubscriptions(y.subscriptions).toBe([]);
```

I think that users could have more troubles wanting to navigate between adjacent marbles without having enough whitespace after the `const ysubs` line ends. What do you think about inlining them? Or, if inline is not an option (for whatever reason), I'd at least put them after all marbles strings.

### DIFF
--- a/spec/operators/exhaust-spec.ts
+++ b/spec/operators/exhaust-spec.ts
@@ -1,221 +1,269 @@
+/** @prettier */
 import { expect } from 'chai';
 import { exhaust, mergeMap, take } from 'rxjs/operators';
-import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
+import { TestScheduler } from 'rxjs/testing';
 import { of, Observable } from 'rxjs';
+import { observableMatcher } from '../helpers/observableMatcher';
 
 /** @test {exhaust} */
-describe('exhaust operator', () => {
-  it('should handle a hot observable of hot observables', () => {
-    const x =   cold(      '--a---b---c--|               ');
-    const y =   cold(              '---d--e---f---|      ');
-    const z =   cold(                    '---g--h---i---|');
-    const e1 = hot(  '------x-------y-----z-------------|', { x: x, y: y, z: z });
-    const expected = '--------a---b---c------g--h---i---|';
+describe('exhaust', () => {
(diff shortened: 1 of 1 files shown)


benlesh:
The first test was likely used to generate a marble diagram with an automated tool we used to use.

benlesh:
I'm relatively indifferent on the location of the empty arrays for subscription. On one hand, we usually keep the subscription marbles next to the observable marbles they are related to, so putting the empty array there makes sense. It says "no subscriptions". But I guess if we wanted to be that explicit, we could do: `const subs = ['------------------------------']` or something.

On the other hand, I don't think it's any more or less readable either way.

I guess I'd just say make sure it's consistent with the rest of the tests.

benlesh (approved):

benlesh:
Overall, I think this is good to merge as is.

benlesh:
Thanks again, @jakovljevic-mladen! Very much appreciated!

jakovljevic-mladen:
> The first test was likely used to generate a marble diagram with an automated tool we used to use.

Yes, most probably. Many tests suites don't `expectSubscriptions` in the first test, I was wandering why. I guess this is why 😃

Thank you Ben for the review.

PR #5869, opened 2020-10-30
refactor(pairs): Fix TS types to be more accurate

benlesh:
Types now match runtime behavior more closely.

### DIFF
--- a/src/internal/observable/pairs.ts
+++ b/src/internal/observable/pairs.ts
@@ -6,7 +6,7 @@ import { from } from './from';
 /**
  * @deprecated To be removed in version 8. Use `from(Object.entries(obj))` instead.
  */
-export function pairs<T>(arr: T[], scheduler?: SchedulerLike): Observable<[string, T]>;
+export function pairs<T>(arr: readonly T[], scheduler?: SchedulerLike): Observable<[string, T]>;
 /**
  * @deprecated To be removed in version 8. Use `from(Object.entries(obj))` instead.
  */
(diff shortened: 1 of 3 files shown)


src/internal/observable/pairs.ts
```diff
@@ -3,6 +3,23 @@ import { Observable } from '../Observable';
 import { SchedulerLike } from '../types';
 import { from } from './from';
 
+/**
+ * @deprecated To be removed in version 8. Use `from(Object.entries(obj))` instead.
+ */
+export function pairs<T>(arr: T[], scheduler?: SchedulerLike): Observable<[string, T]>;
```
cartant, line 9: nitpick: This should probably be `arr: readonly T[]` so that callers can pass read-only arrays.

src/internal/observable/pairs.ts
```diff
@@ -3,6 +3,23 @@ import { Observable } from '../Observable';
...
+ */
+export function pairs<O extends Record<string, unknown>>(obj: O, scheduler?: SchedulerLike): Observable<[keyof O, O[keyof O]]>;
+/**
+ * @deprecated To be removed in version 8. Use `from(Object.entries(obj))` instead.
+ */
+export function pairs<T>(iterable: Iterable<T>, scheduler?: SchedulerLike): Observable<[string, T]>;
+/**
+ * @deprecated To be removed in version 8. Use `from(Object.entries(obj))` instead.
+ */
+export function pairs(n: number | boolean | ((...args: any[]) => any) | symbol, scheduler?: SchedulerLike): Observable<[never, never]>;
```
cartant, line 21: nitpick: Given that TS 4.x is going to be our minimum, maybe include `bigint` here, too? Pretty sure that `0n` is used in the `filter` predicate business, so our types already use a `bigint` literal type in some places.

cartant (approved):
LGTM, just some very minor nitpicks.

PR #5873, opened 2020-11-02
fix(tap): Generic handle to not change return type

imcotton:


josepot:
This change was already proposed on #5616 and it was disregarded based on the reasons exposed on the discussion of that PR.

imcotton:
Oh, didn't realize that, after reading the discussion thinking I could live with that.

Thank you for the heads up, closing.

PR #5874, opened 2020-11-02
fix(Symbol.observable): properly defined as a `unique symbol`.

### LINKED ISSUE #4415: RxJS incorrectly defines observable symbol
## Bug Report
It would appear that the current way RxJS detects other observables is not working as expected, while other libraries can adapt from each other.

**Current Behavior**
The `from` function in RxJS 6.3.3 does not seem able to adapt `xstream` and `most` while the older RxJS 5 `Observable.from` was able to. The following error is reproduced in both cases:

> TypeError: You provided an invalid object where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.

Both `xstream` and `most` provide `Symbol` polyfills:
- https://github.com/staltz/xstream/blob/0cadc46b4cff9ab6deea964db10706709a10f113/dist/xstream.js#L1250
- https://github.com/cujojs/most/blob/7843b1ade293caf61889cd55409055c5319c47e6/dist/most.js#L584

Additional testing has confirmed:
- This issue does not occur in `webpack` but it does in `node`, `jest`, and `browserify`.
- The order of import seems to matter.

**Reproduction**
- From `xstream` to `rxjs`, does not work:
```js
const { from: observableFrom } = require('rxjs')
const { isInteropObservable } = require('rxjs/internal/util/isInteropObservable')
const { default: xs } = require('xstream')

const num$ = xs.from([ 1, 2, 3, 4 ])
console.log('Is num$ an observable?', isInteropObservable(num$)) // false

const adapt$ = observableFrom(num$) // error here
adapt$.subscribe(console.log)
```
- From `xstream` to `rxjs`, working after importing `xstream` first:
```js
const { default: xs } = require('xstream')
const { from: observableFrom } = require('rxjs')
const { isInteropObservable } = require('rxjs/internal/util/isInteropObservable')

const num$ = xs.from([ 1, 2, 3, 4 ])
console.log('Is num$ an observable?', isInteropObservable(num$)) // true

const adapt$ = observableFrom(num$)
adapt$.subscribe(console.log)
```
- The same behaviour in `most`, not working:
```js
const { from: observableFrom } = require('rxjs')
const { isInteropObservable } = require('rxjs/internal/util/isInteropObservable')
const most = require('most')

const num$ = most.from([ 1, 2, 3, 4 ])
console.log('Is num$ an observable?', isInteropObservable(num$)) // false

const adapt$ = observableFrom(num$) // error here
adapt$.subscribe(console.log)
```
- We cannot convert `rxjs` to `xstream` either unless we import `xstream` first:
```js
const { from: observableFrom } = require('rxjs')
const { default: xs } = require('xstream')

const num$ = observableFrom([ 1, 2, 3, 4 ])
const adapt$ = xs.from(num$)

adapt$.addListener({
  next: console.log
})
```

**Expected behavior**
The order of import should not matter, and the `jest` environment would work like `webpack`.

**Environment**
- Runtime: Node v11.3.0
- RxJS version: 6.3.3
- xstream version: 11.7.0
- most version: 1.7.3

**Possible Solution**
- The file ` rxjs/src/internal/symbol/observable.ts` could be defined the same way the other libraries do it:
```js
Symbol('observable')
```
- The `isInteropObservable` could be patched to check for the same symbol `xstream` and `most` use?
- A similar polyfill to `xstream` or `most` could be included?

- This issue came about while working on a `cyclejs` project with `rxjs`.
- Adding the following to `jest.setup.js` seems to have fixed the issue for now:
```js
import xs from 'xstream' // eslint-disable-line
```

Thank you and sorry in advance if I've diagnosed the wrong source :bowing_man:

### LINKED ISSUE #5861: Node v14: internal/types.d.ts(45,13): Property 'observable' does not exist on type 'SymbolConstructor'.
## Bug Report

**Current Behavior**
Issue #3674 is resurrected with @types/node version 14.

Looks like it was removed in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40927.

**Environment**
- Runtime: Node v14
- RxJS version: 6.6.3

benlesh:
Resolves #5861
Resolves #4415

This is also related to issues here: https://github.com/benlesh/symbol-observable/issues/41

I don't think we can publish this in 6.x, as it will break people using older versions of `@types/node`, however, we do need to set an upper bounds for `@types/node` and version 6.x.

### DIFF
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -298,7 +298,7 @@ export interface ObjectUnsubscribedError extends Error {
 
 export declare const ObjectUnsubscribedError: ObjectUnsubscribedErrorCtor;
 
-export declare const observable: string | symbol;
+export declare const observable: string | SymbolConstructor["observable"];
 
 export declare class Observable<T> implements Subscribable<T> {
     protected operator: Operator<any, T> | undefined;
--- a/package-lock.json
+++ b/package-lock.json
@@ -231,9 +231,9 @@
       "dev": true
     },
     "@types/node": {
-      "version": "12.12.7",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-12.12.7.tgz",
-      "integrity": "sha512-E6Zn0rffhgd130zbCbAr/JdXfXkoOUFAKNs/rF8qnafSJ8KYaA/j3oz7dcwal+lYjLA7xvdd5J4wdYpCTlP8+w==",
+      "version": "14.14.6",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-14.14.6.tgz",
+      "integrity": "sha512-6QlRuqsQ/Ox/aJEQWBEJG7A9+u7oSYl3mem/K8IzxXG/kAGbV1YPD9Bg9Zw3vyxC/YP+zONKwy8hGkSt1jxFMw==",
       "dev": true
     },
     "@types/parse-json": {
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
     "@types/chai": "^4.2.11",
     "@types/lodash": "4.14.102",
     "@types/mocha": "^7.0.2",
-    "@types/node": "^12.12.5",
+    "@types/node": "^14.14.6",
     "@types/shelljs": "^0.8.8",
     "@types/sinon": "4.1.3",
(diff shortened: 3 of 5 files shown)


kwonoj (approved):

benlesh:
Added breaking change note to show in up in the CHANGELOG.

PR #5875, opened 2020-11-02
chore: update delay tests

jakovljevic-mladen:
**Description:**

- Enabled Prettier.
- Stopped passing scheduler to `delay` since run mode is already used.
- Used `time` helper to convert time units where possible (only negative time units were left out).
- Ordered variables so that they, where possible, have this order: `e1`, `e1subs` (where needed, renamed from `subs`), `t` and then `expected` since this is the most common ordering in many other tests.

None


cartant (approved):
LGTM. Thanks.

PR #5878, opened 2020-11-06
perf: ensure same hidden class for OperatorSubscriber

cartant:
**Description:**

See the comments within the change. This PR should help realise one of the advantages of replacing operator-specific subscribers with a general-purpose subscriber - reducing/eliminating megamorphic property accesses within `Subscriber`.

I've not done any micro optimization/analysis, yet. I'll look at doing that after the version 7 release. However, I think that what's in this PR makes sense and shouldn't be controversial. 🤞

### DIFF
--- a/src/internal/operators/OperatorSubscriber.ts
+++ b/src/internal/operators/OperatorSubscriber.ts
@@ -25,40 +25,48 @@ export class OperatorSubscriber<T> extends Subscriber<T> {
     onComplete?: () => void,
     private onUnsubscribe?: () => void
   ) {
+    // It's important - for performance reasons - that all of this class's
+    // members are initialized and that they are always initialized in the same
+    // order. This will ensure that all OperatorSubscriber instances have the
+    // same hidden class in V8. This, in turn, will help keep the number of
+    // hidden classes involved in property accesses within the base class as
+    // low as possible. If the number of hidden classes involved exceeds four,
+    // the property accesses will become megamorphic and performance penalties
+    // will be incurred - i.e. inline caches won't be used.
     super(destination);
-    if (onNext) {
-      this._next = function (value: T) {
-        try {
-          onNext(value);
-        } catch (err) {
-          this.destination.error(err);
+    this._next = onNext
+      ? function (this: OperatorSubscriber<T>, value: T) {
+          try {
+            onNext(value);
+          } catch (err) {
+            this.destination.error(err);
+          }
         }
-      };
-    }
-    if (onError) {
-      this._error = function (err) {
-        try {
-          onError(err);
-        } catch (err) {
-          // Send any errors that occur down stream.
-          this.destination.error(err);
+      : super._next;
+    this._error = onError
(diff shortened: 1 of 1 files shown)


cartant:
I don't understand this CI failure. It seems to have something to do with side effects? I ran `test:side-effects:update`, but it seemed to generate a surprising number of changes, so ... 🤷‍♂️

src/internal/operators/OperatorSubscriber.ts
```diff
@@ -1,6 +1,16 @@
 /** @prettier */
 import { Subscriber } from '../Subscriber';
 
+// Ignore TypeScript's errors about the following methods being protected.
+// They're stored here to avoid their needing to be looked up in the
+// constructor.
+// @ts-ignore
+const defaultOnNext = Subscriber.prototype._next;
```
benlesh, line 8: I think you can just use `super._next` below. :)

benlesh (changes requested):

benlesh:
So the CI errors are from the side-effects checker. However, I think if you implement the change I'm suggesting (which is to use `super._next`, `super._error`, etc), then you won't have those declarations at the top level, and the side-effects checker will be happy again.

src/internal/operators/OperatorSubscriber.ts
cartant: `super` cannot be used in a parameter declaration; it can only be used after `super()` has been called.

cartant:
> So the CI errors are from the side-effects checker.

Yeah, I can see that it's related to that: the checker has no way of knowing (with absolute certainty) that the property accesses are not calls to getters. I was mostly concerned about the diff after running the update script. It seemed to be unnecessarily large. I'll have another look at it after #5898 is merged/addressed, as that problem is now breaking the build.

src/internal/operators/OperatorSubscriber.ts
cartant: I really shouldn't correspond until I've had m morning coffee. A previous iteration to what's in this commit was using default parameter and I forgot that I'd changed it. 😅 Sorry.

As it is, do we really want to use `super` with a property lookup just to appease two build-time tools? TypeScript  - because of the `protected` business - and the `check-side-effects` - which freaks out when it sees property accesses. Why pay a runtime cost - however small - to make those tools happy? As it is, this class is already doing weird stuff by overloading prototype methods with instance properties.

benlesh: 🤷‍♂️ The other way, there's an additional closure which is also a cost. All I know it's it's what `super` is for. Seems like we should use it there. If it's really down to performance, we should be profiling this to make the decision, I guess.

cartant: I don't want to be argumentative, but where is the additional closure?

benlesh (approved):

benlesh:
Sorry, this one almost fell off my radar, @cartant. Thanks for the work!

PR #5880, opened 2020-11-07
refactor(withLatestFrom): support N-args with project function

tmair:
**Description:**
This PR refactors the signatures of the `withLatestFrom` operator to use N-args in the signature.

I also added a few more dtslint tests and altered the current ones to better reflect the intent of the typings.

There is one issue with the typings I could not get to work correctly.
```typescript
const a = of(1, 2, 3);
const b = [a, a, a];
const res = a.pipe(withLatestFrom(...b, (a, b, c, d) => d); // Should be Observable<number> but is infered as Observable<unknown>
```

#5066

### DIFF
--- a/spec-dtslint/operators/withLatestFrom-spec.ts
+++ b/spec-dtslint/operators/withLatestFrom-spec.ts
@@ -12,46 +12,66 @@ describe('withLatestFrom', () => {
     it('should infer correctly with 2 params', () => {
       const a = of(1, 2, 3);
       const b = of('a', 'b', 'c');
-      const c = of('d', 'e', 'f');
-      const res = a.pipe(withLatestFrom(b, c)); // $ExpectType Observable<[number, string, string]>
+      const c = of(1, 2, 3);
+      const res = a.pipe(withLatestFrom(b, c)); // $ExpectType Observable<[number, string, number]>
     });
 
     it('should infer correctly with 3 params', () => {
       const a = of(1, 2, 3);
       const b = of('a', 'b', 'c');
-      const c = of('d', 'e', 'f');
+      const c = of(1, 2, 3);
       const d = of('g', 'h', 'i');
-      const res = a.pipe(withLatestFrom(b, c, d)); // $ExpectType Observable<[number, string, string, string]>
+      const res = a.pipe(withLatestFrom(b, c, d)); // $ExpectType Observable<[number, string, number, string]>
     });
 
     it('should infer correctly with 4 params', () => {
       const a = of(1, 2, 3);
       const b = of('a', 'b', 'c');
-      const c = of('d', 'e', 'f');
+      const c = of(1, 2, 3);
       const d = of('g', 'h', 'i');
-      const e = of('j', 'k', 'l');
-      const res = a.pipe(withLatestFrom(b, c, d, e)); // $ExpectType Observable<[number, string, string, string, string]>
+      const e = of(4, 5, 6);
+      const res = a.pipe(withLatestFrom(b, c, d, e)); // $ExpectType Observable<[number, string, number, string, number]>
     });
 
     it('should infer correctly with 5 params', () => {
       const a = of(1, 2, 3);
       const b = of('a', 'b', 'c');
-      const c = of('d', 'e', 'f');
+      const c = of(1, 2, 3);
       const d = of('g', 'h', 'i');
-      const e = of('j', 'k', 'l');
+      const e = of(4, 5, 6);
       const f = of('m', 'n', 'o');
-      const res = a.pipe(withLatestFrom(b, c, d, e, f)); // $ExpectType Observable<[number, string, string, string, string, string]>
+      const res = a.pipe(withLatestFrom(b, c, d, e, f)); // $ExpectType Observable<[number, string, number, string, number, string]>
     });
 
-    it('should only accept maximum params of 5', () => {
+    it('should infer correctly with 6 params', () => {
       const a = of(1, 2, 3);
       const b = of('a', 'b', 'c');
-      const c = of('d', 'e', 'f');
+      const c = of(1, 2, 3);
       const d = of('g', 'h', 'i');
-      const e = of('j', 'k', 'l');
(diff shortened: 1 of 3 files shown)


spec-dtslint/operators/withLatestFrom-spec.ts
```diff
@@ -70,35 +101,102 @@ describe('withLatestFrom', () => {
...
+
+    it('should infer correctly with array parameter', () => {
+      const a = of(1, 2, 3);
+      const b = of('a', 'b', 'c');
+      const c = of(1, 2, 3);
+      const d = of('g', 'h', 'i');
+      const e = of(4, 5, 6);
+      const f = of('m', 'n', 'o');
+      const g = of(7, 8, 9);
+      const res = a.pipe(withLatestFrom([b, c, d, e, f, g], (a, b) => b)); // $ExpectType Observable<string>
```
cartant, line 154: issue: This inference is not correct. AFAICT, there the implementation does not support an array of observables. It always expects sources to be supplied via a rest parameter. It looks to me like the resultant observable would be `Observable<number | Observable<string | number>>` - as the sources in the array would not be flattened by the implementation.

src/internal/operators/withLatestFrom.ts
```diff
@@ -1,107 +1,24 @@
...
-    v5: ObservedValueOf<O5>,
-    v6: ObservedValueOf<O6>
-  ) => R
+export function withLatestFrom<T, O extends unknown[]>(inputs: [...ObservableInputTuple<O>]): OperatorFunction<T, [T, ...O]>;
+
+export function withLatestFrom<T, O extends unknown[]>(...inputs: [...ObservableInputTuple<O>]): OperatorFunction<T, [T, ...O]>;
+
+export function withLatestFrom<T, O extends unknown[], R>(
+  inputs: [...ObservableInputTuple<O>],
+  project: (...value: [T, ...O]) => R
```
cartant, line 16: issue: The project needs to be moved into the same tuple that's used for the inputs - [as done here](https://github.com/ReactiveX/rxjs/blob/516166ed8dc389d57bf64dd536cb845dbb23402d/src/internal/operators/mergeWith.ts#L11-L14). It should look like this:

```ts
export function withLatestFrom<T, O extends unknown[], R>(
  inputsAndProject: [...ObservableInputTuple<O>, (...value: [T, ...O]) => R]
): OperatorFunction<T, R>;
```

src/internal/operators/withLatestFrom.ts
```diff
@@ -1,107 +1,24 @@
...
-  v6: O6
-): OperatorFunction<T, [T, ObservedValueOf<O2>, ObservedValueOf<O3>, ObservedValueOf<O4>, ObservedValueOf<O5>, ObservedValueOf<O6>]>;
-export function withLatestFrom<T, R>(...observables: Array<ObservableInput<any> | ((...values: Array<any>) => R)>): OperatorFunction<T, R>;
-export function withLatestFrom<T, R>(array: ObservableInput<any>[]): OperatorFunction<T, R>;
-export function withLatestFrom<T, R>(array: ObservableInput<any>[], project: (...values: Array<any>) => R): OperatorFunction<T, R>;
 
-/* tslint:enable:max-line-length */
+export function withLatestFrom<T, O extends ObservableInput<unknown>[], R>(
+  ...inputs: [...O, (...value: [T, ...ObservedValueTupleFromArray<O>]) => R]
+): OperatorFunction<T, R>;
```
cartant, line 21: issue: It should be possible to delete this. After the N-args refactors, we don't want any type parameter constraints to be using `ObservableInput`. That type should be used only for selector return types.

cartant (changes requested):
Thanks for the PR. What's here is very close. There are just a few things that I think need to be changed.

spec-dtslint/operators/withLatestFrom-spec.ts
tmair: You are right. I removed the signature and also the corresponding test.

src/internal/operators/withLatestFrom.ts
tmair: The test mentioned in the description works now correctly.

src/internal/operators/withLatestFrom.ts
tmair: Yep, I refactored the other signatures so this is no longer necessary.

tmair:
I refactored the code to reflect your comments. There is one issue I was not able to tackle. I don't know if it is a blocker for this PR, so I will let you decide if it is or not.

The following code does not typecheck:
```typescript
      const a = of(1, 2, 3);
      const b = of('a', 'b', 'c');
      const c = of(1, 2, 3);
      const res = a.pipe(withLatestFrom(b, c, (a, b) => b)); // $ExpectType Observable<string>
```

The error is the following: 
```
Argument of type '(a: number, b: string) => string' is not assignable to parameter of type 'ObservableInput<unknown>'.
  Property '[Symbol.iterator]' is missing in type '(a: number, b: string) => string' but required in type 'Iterable<unknown>'.
```

I am a little out of ideas what is causing the compiler to only typecheck the code correctly when the number of arguments supplied to the project function matches the number of observable inputs. If there are cases where this should typecheck we probably need to find some way to get the compiler relax the number of arguments of the projection function.

cartant:
> I refactored the code to reflect your comments. There is one issue I was not able to tackle. I don't know if it is a blocker for this PR, so I will let you decide if it is or not.
>
> The following code does not typecheck:

IMO, that is a TypeScript bug. Also, result selectors that specify fewer parameters than they receive are pretty unlikely and the workaround is straightforward - callers can specify the parameter.

I think you should comment out the test - noting that it looks like a TS bug.

With that done, I think this is good to merge.

tmair:
> With that done, I think this is good to merge. I'll review it again once the test is commented out.

I've updated the PR with your comments. The build is good now.

cartant (approved):
LGTM. Thanks.

benlesh (approved):

PR #5881, opened 2020-11-07
refactor: use ObservableInputTuple in xWith operators

cartant:
**Description:**

This PR refactors the `xWith` operators - `concatWith`, et al. - to use `ObservableInputTuple` with a type-parameter constraint of `readonly unknown[]` - instead of `ObservableInput<any>[]`. The former is simpler, more idiomatic and doesn't rely on the `ObservedValueOf` type.

Interestingly, this change uncovered two tests that use result selectors. I have no idea how those tests compiled without effecting TypeScript compilation errors. I've split the tests out into separate files `combineLatest-spec.ts` and `zip-spec.ts`, as the non-`With` operators are deprecated, but still exported. I guess this shows that the `ObservableInputTuple` approach has some advantages.

The refactor also removed some unnecessary overload signatures.

This PR deliberately does not refactor the deprecated, non-`With` signatures. Those aren't yet N-args and their refactor will involve changes to the tests. I wanted to keep this PR as simple as possible. I'll do the non-`With` operators in separate PRs.

### DIFF
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -192,7 +192,7 @@ export interface GroupedObservable<K, T> extends Observable<T> {
     readonly key: K;
 }
 
-export declare type Head<X extends readonly any[]> = ((...args: X) => any) extends (arg: infer U, ...rest: readonly any[]) => any ? U : never;
+export declare type Head<X extends readonly any[]> = ((...args: X) => any) extends (arg: infer U, ...rest: any[]) => any ? U : never;
 
 export declare function identity<T>(x: T): T;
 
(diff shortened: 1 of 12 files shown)


api_guard/dist/types/index.d.ts
```diff
@@ -132,7 +132,7 @@ export declare class ConnectableObservable<T> extends Observable<T> {
     refCount(): Observable<T>;
 }
 
-export declare type Cons<X, Y extends any[]> = ((arg: X, ...rest: Y) => any) extends (...args: infer U) => any ? U : never;
+export declare type Cons<X, Y extends readonly any[]> = ((arg: X, ...rest: Y) => any) extends (...args: infer U) => any ? U : never;
```
benlesh, line 135: I'll be honest, the name `Cons` confuses me when I see it. It would be nice if it was called like `PrependTuple` or `UnshiftTuple` or something? (We don't have to do that in this PR, obviously, I just don't understand the name `Cons`)

benlesh (approved):

PR #5884, opened 2020-11-08
refactor(zip): support N-args with result selector

cartant:
**Description:**

Updates the static `zip` to use `ObservableInputTuple`, etc.

### DIFF
--- a/spec-dtslint/observables/zip-spec.ts
+++ b/spec-dtslint/observables/zip-spec.ts
@@ -29,12 +29,12 @@ it('should support arrays of observables', () => {
 
 it('should return Array<T> when given a single promise', () => {
   const a = Promise.resolve(1); // $ExpectType Promise<number>
-  const o1 = zip(a); // $ExpectType Observable<number[]>
+  const o1 = zip(a); // $ExpectType Observable<[number]>
 });
 
 it('should return Array<T> when given a single observable', () => {
   const a = of(1); // $ExpectType Observable<number>
-  const o1 = zip(a); // $ExpectType Observable<number[]>
+  const o1 = zip(a); // $ExpectType Observable<[number]>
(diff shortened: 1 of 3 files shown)


spec-dtslint/observables/zip-spec.ts
```diff
@@ -1,49 +1,49 @@
...
-  const u2 = Math.random() > 0.5 ? of(true) : of([1, 2, 3]);
-  const o = zip(u, u2); // $ExpectType Observable<[string | number, boolean | number[]]>
-});
+import { Observable, of, zip } from 'rxjs';
+
+it('should support observables', () => {
+  const a = of(1); // $ExpectType Observable<number>
+  const b = of('foo'); // $ExpectType Observable<string>
+  const c = of(true); // $ExpectType Observable<boolean>
+  const o1 = zip(a, b, c); // $ExpectType Observable<[number, string, boolean]>
```
kolodny, line 7: nit: can you add a test for multiple const values?

```ts
const a = from([1, 2] as const);
const b = from(['foo', 'bar'] as const);
const c = from([true, false] as const);
const o1 = zip(a, b, c); // $ExpectType Observable<[1, 'foo', true] | [2, 'bar', false]>
```

Does that work?

kolodny (approved):

benlesh (changes requested):
This seems to be missing dtslint tests for result selector

benlesh (approved):

benlesh:
Merging and moving on.

cartant:
@benlesh

> Merging and moving on.

I'll come back to the missing dtslint tests later in the upcoming week. Or I'll add them to the list of things for Jan-Niklas to do whist he's in quarantine. 🙂

PR #5885, opened 2020-11-08
refactor(forkJoin): remove any-based signature

cartant:
**Description:**

This PR tweaks the overload signatures for `forkJoin` to remove an `any`-based signature, replacing it with an N-args versions that accept (deprecated) result selectors.

<s>The PR also adds `readonly` qualifiers for the `[...ObservableInputTuple<A>]` types. I swear I tested `readonly` qualifiers with mapped tuple types for a blog post and found that they weren't necessary. However, without them, tests fail. IDK whether or not this is a TypeScript-version issue. I will look into it later and separately.</s>


kolodny (approved):

benlesh (changes requested):
It looks like we're missing dtslint tests for result selector here.

PR #5896, opened 2020-11-17
docs: remove dead link

cartant:
**Description:**

This PR removes a dead link from the comments/docs. It's a link to one of my blog posts, but the domain - which was once used for a Medium publication - seems to have been allowed to die, so lets just rid of the link altogether.


niklas-wortmann (approved):
LGTM

PR #5898, opened 2020-11-18
chore(onErrorResumeNext): work around TS 4.1 RC bug

cartant:
**Description:**

After upgrading to TS 4.1 RC, there's an incorrect inference in the implementation of `onErrorResumeNext`. The problem didn't occur in the refactor of that operator - which was done under TS 4.0 - and the inference is clearly incorrect. This PR adds an explicit assertion - and a comment - as a workaround.

### DIFF
--- a/src/internal/operators/onErrorResumeNext.ts
+++ b/src/internal/operators/onErrorResumeNext.ts
@@ -87,7 +87,7 @@ export function onErrorResumeNext<T, A extends readonly unknown[]>(
   // result to be `A[number][]` - completely dropping the ObservableInput part
   // of the type. This makes no sense whatsoever. As a workaround, the type is
   // asserted explicitly.
-  const nextSources = (argsOrArgArray(sources) as unknown) as [[...ObservableInputTuple<A>]];
+  const nextSources = (argsOrArgArray(sources) as unknown) as ObservableInputTuple<A>;
 
   return operate((source, subscriber) => {
     const remaining = [source, ...nextSources];
--- a/src/internal/operators/onErrorResumeNext.ts
+++ b/src/internal/operators/onErrorResumeNext.ts
@@ -83,7 +83,11 @@ export function onErrorResumeNext<T, A extends readonly unknown[]>(
 export function onErrorResumeNext<T, A extends readonly unknown[]>(
   ...sources: [[...ObservableInputTuple<A>]] | [...ObservableInputTuple<A>]
 ): OperatorFunction<T, T | A[number]> {
-  const nextSources = argsOrArgArray(sources);
+  // For some reason, TS 4.1 RC gets the inference wrong here and infers the
+  // result to be `A[number][]` - completely dropping the ObservableInput part
+  // of the type. This makes no sense whatsoever. As a workaround, the type is
+  // asserted explicitly.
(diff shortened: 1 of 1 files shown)


src/internal/operators/onErrorResumeNext.ts
```diff
@@ -83,7 +83,11 @@ export function onErrorResumeNext<T, A extends readonly unknown[]>(
 export function onErrorResumeNext<T, A extends readonly unknown[]>(
   ...sources: [[...ObservableInputTuple<A>]] | [...ObservableInputTuple<A>]
 ): OperatorFunction<T, T | A[number]> {
-  const nextSources = argsOrArgArray(sources);
+  // For some reason, TS 4.1 RC gets the inference wrong here and infers the
+  // result to be `A[number][]` - completely dropping the ObservableInput part
+  // of the type. This makes no sense whatsoever. As a workaround, the type is
+  // asserted explicitly.
+  const nextSources = (argsOrArgArray(sources) as unknown) as [[...ObservableInputTuple<A>]];
```
kolodny, line 90: nit: Can this also be `as [ObservableInputTuple<A>]`?

src/internal/operators/onErrorResumeNext.ts
```diff
@@ -83,7 +83,11 @@ export function onErrorResumeNext<T, A extends readonly unknown[]>(
 export function onErrorResumeNext<T, A extends readonly unknown[]>(
   ...sources: [[...ObservableInputTuple<A>]] | [...ObservableInputTuple<A>]
 ): OperatorFunction<T, T | A[number]> {
-  const nextSources = argsOrArgArray(sources);
+  // For some reason, TS 4.1 RC gets the inference wrong here and infers the
+  // result to be `A[number][]` - completely dropping the ObservableInput part
+  // of the type. This makes no sense whatsoever. As a workaround, the type is
+  // asserted explicitly.
```
kolodny, line 89: Is it worth opening an issue with an MVP repro in the TS repo?

kolodny (approved):

src/internal/operators/onErrorResumeNext.ts
cartant: If you can create a minimal reproduction, sure, go for it. I won't be; I've found interactions in the TS repo to be sufficiently discouraging that I no longer bother.

cartant: You'll need to reproduce it in the nightly, BTW.

src/internal/operators/onErrorResumeNext.ts
cartant: Done. I think it should have been `ObservableInputTuple<A>`, actually.

cartant:
FWIW, this workaround is still necessary in the 4.1.2 release.

src/internal/operators/onErrorResumeNext.ts
benlesh: > I've found interactions in the TS repo to be sufficiently discouraging that I no longer bother.

FWIW: I haven't had that experience. We've always had a really strong relationship with the TS team, and they've pushed a lot of really cool stuff through for us in the past.

benlesh (approved):

PR #5902, opened 2020-11-21
chore: update filter tests to run mode

jakovljevic-mladen:
**Description:**
Updated `filter` tests to use run mode.

None


cartant (approved):
LGTM. Thanks.

PR #5905, opened 2020-11-24
chore: update finalize tests to run mode

jakovljevic-mladen:
**Description:**
Updated `finalize` tests to use run mode.

One note: I didn't want to add `testScheduler` and `beforeEach` to the `describe` callback (as it's usually done in other tests) because many of the tests in this suite don't use `testScheduler` instance, so I was creating them where they were needed.

None


cartant (approved):
LGTM. Thanks.

PR #3857, opened 2018-06-23
feat(TestScheduler): Support use of emoji in marble syntax

jiayihu:
TestScheduler.parseMarbles should correcty read an emoji as single value

**Description:** Currently `TestScheduler` doesn't correctly support emojis in marble syntax because the string is read using `marbles[i]` and it doesn't consider that emojis are encoded as [surrogate pairs](https://mathiasbynens.be/notes/javascript-unicode#counting-symbols).

Using `toArray` from `lodash` the string is correctly splitted into an equivalent array, where each emoji is considered as single character in `parseMarbles`.

All unit tests pass but the build is failing with Rollup and I'm not familiar with it, how could I resolve?

<details>
<summary>Build error</summary>
<pre><code>
> @reactivex/rxjs@6.2.1 clean_dist_global /Users/jiayi/Desktop/Repo/rxjs
> shx rm -rf ./dist/global

node_modules/lodash/lodash.js (17097:8) The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten. See https://github.com/rollup/rollup/wiki/Troubleshooting#this-is-undefined for more information
(node:3276) UnhandledPromiseRejectionWarning: Error: 'toArray' is not exported by node_modules/lodash/lodash.js (imported by dist/esm5_for_rollup/internal/testing/TestScheduler.js). For help fixing this error see https://github.com/rollup/rollup/wiki/Troubleshooting#name-is-not-exported-by-module
    at Module.trace (/Users/jiayi/Desktop/Repo/rxjs/node_modules/rollup/dist/rollup.js:7739:29)
    at ModuleScope.findDeclaration (/Users/jiayi/Desktop/Repo/rxjs/node_modules/rollup/dist/rollup.js:7361:22)
    at Scope.findDeclaration (/Users/jiayi/Desktop/Repo/rxjs/node_modules/rollup/dist/rollup.js:5361:39)
    at Scope.findDeclaration (/Users/jiayi/Desktop/Repo/rxjs/node_modules/rollup/dist/rollup.js:5361:39)
    at Node.bind (/Users/jiayi/Desktop/Repo/rxjs/node_modules/rollup/dist/rollup.js:5846:28)
    at /Users/jiayi/Desktop/Repo/rxjs/node_modules/rollup/dist/rollup.js:5155:50
    at Node.eachChild (/Users/jiayi/Desktop/Repo/rxjs/node_modules/rollup/dist/rollup.js:5172:5)
[...]

PR #5907, opened 2020-11-27
feat: support emoji in marble diagrams

cartant:
**Description:**

This PR adds support for emoji characters in marble diagrams by leveraging ES2015's built-in emoji support for string iteration.

An attempt to add this feature was made before - in #3857 - but that attempt sought to use `toArray` from `lodash`. This PR relies upon the built-in behaviour (and downleveled iteration for the ES5 builds):

```ts
const string = '🙈🙉🙊';
const { length } = string; // 6
const characters = [...string]; // ['🙈', '🙉', '🙊']
```

And, let's face it, this is something that everyone <s>wants</s> needs, right?

### DIFF
--- a/spec/schedulers/TestScheduler-spec.ts
+++ b/spec/schedulers/TestScheduler-spec.ts
@@ -88,7 +88,7 @@ describe('TestScheduler', () => {
       ]);
     });
 
-    it('should suppport time progression syntax when runMode=true', () => {
+    it('should support time progression syntax when runMode=true', () => {
       const runMode = true;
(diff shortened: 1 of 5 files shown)


spec/schedulers/TestScheduler-spec.ts
```diff
@@ -88,7 +88,7 @@ describe('TestScheduler', () => {
       ]);
     });
 
-    it('should suppport time progression syntax when runMode=true', () => {
+    it('should support time progression syntax when runMode=true', () => {
```
benlesh, line 91: supppppppppppppppppppport

benlesh (approved):

PR #5917, opened 2020-12-07
refactor(tap): remove deprecated signatures

aboyton:
This removes the deprecated signatures, simplifies the code a little, makes `next` a required parameter, and updates all the unit tests to pass.

Ref #4159

**Description:**
This is my best attempt at understanding what is desired by #4159.

### DIFF
--- a/spec/operators/delayWhen-spec.ts
+++ b/spec/operators/delayWhen-spec.ts
@@ -333,8 +333,10 @@ describe('delayWhen', () => {
 
       expectObservable(
         result.pipe(
-          tap(null, null, () => {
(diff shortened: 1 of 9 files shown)


cartant:
Blocking this for the moment, as I'll be looking at `tap` as part of getting v7 finalized. I'll have a closer look at this PR when I do that.

cartant:
There have been some more recent changes to the `subscribe` signature - to make it simpler - and I think that `tap` should get those changes - and they are a little different to what's in this PR. Also, I don't think we should be deleting tests just because the signature is deprecated. Thanks for the effort you put into this PR, but I'm going to close it. Sorry.

PR #5922, opened 2020-12-08
Modified behavior on empty inputs for forkJoin, combineLatest and zip

### LINKED ISSUE #5209: forkJoin should emit empty array or object given empty array or object as input

rraziel:
**Description:**

The `forkJoin`, `combineLatest` and `zip` functions have been modified so that:

```ts
combineLatest();   // Observable<never>: behavior kept for compatibility as the non-array version is deprecated
combineLatest([]); // Observable<[]>: now emits an empty array before completing
combineLatest({}); // Observable<{}>: now emits an empty object before completing

forkJoin();   // Observable<never>: behavior kept for compatibility as the non-array version is deprecated
forkJoin([]); // Observable<[]>: now emits an empty array before completing
forkJoin({}); // Observable<{}>, now emits an empty object before completing

zip();     // Observable<never>, behavior kept for compatibility as the non-array version is now deprecated
zip(a,b)   // /!\ has been marked as deprecated in this PR /!\
zip([]);   // Observable<[]>: now emits an empty array before completing
zip({a,b}) // Observable<{a: A, b: B}>: new behavior to align with the other ones
           // /!\ new behavior to align with the other ones /!\
zip({});   // Observable<{}>: emits an empty object then completes

// side effects
zipAll() // now emits [] for empty observables
```

Regarding the signature (#5066), in particular `needs N-args signature for observable inputs`:
All 3 functions seem to have the correct `<A extends readonly unknown[]>(sources: [...ObservableInputTuple<A>]): Observable<A>`

Closes #5209

### DIFF
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -40,12 +40,9 @@ export declare function bindCallback<A extends readonly unknown[], R extends rea
 export declare function bindNodeCallback(callbackFunc: (...args: any[]) => void, resultSelector: (...args: any[]) => any, scheduler?: SchedulerLike): (...args: any[]) => Observable<any>;
 export declare function bindNodeCallback<A extends readonly unknown[], R extends readonly unknown[]>(callbackFunc: (...args: [...A, (err: any, ...res: R) => void]) => void, schedulerLike?: SchedulerLike): (...arg: A) => Observable<R extends [] ? void : R extends [any] ? R[0] : R>;
 
-export declare function combineLatest(sources: []): Observable<never>;
 export declare function combineLatest<A extends readonly unknown[]>(sources: readonly [...ObservableInputTuple<A>]): Observable<A>;
+export declare function combineLatest(): Observable<never>;
 export declare function combineLatest<A extends readonly unknown[]>(...sources: [...ObservableInputTuple<A>]): Observable<A>;
-export declare function combineLatest(sourcesObject: {
-    [K in any]: never;
-}): Observable<never>;
 export declare function combineLatest<T>(sourcesObject: T): Observable<{
     [K in keyof T]: ObservedValueOf<T[K]>;
 }>;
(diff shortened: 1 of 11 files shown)


rraziel:
Note: if the deprecated signatures can be removed because of the major version bump (except for the ones I just added on `zip` I suppose), I can update the PR; let me know

rraziel:
@benlesh The fix in 7.0.0-beta.10 contradicts this PR and #5209 's comments, should this PR be dropped?

benlesh:
I really appreciate the effort that went into here, but I don't think we can accept this PR.

With these creation functions, the supplied object or array represents a dictionary or list of observable sources, respectively. If those are empty (`{}` or `[]`), then there are no observables to subscribe to, and therefor, nothing to emit values to emit as an array of values or a dictionary of values... So the behavior should be that it emits an `EMPTY` which is `Observable<never>`.

I hope that makes sense.

Thanks again for your work!

rraziel:
@benlesh No worries, I never got an answer on the comments from #5209 so it's not like anyone told me "go for it" (would have been nice to get a "don't go for it" though 😅)
Should I still create a PR for the `zip({})` code that was also in there (same as `forkJoin({})` and `combineLatest({})`)?

Issue #5931, opened 2020-12-13
Memory leaks caused by shareReplay when shared observable is synchronous or continuous


MaximSagan:
## Bug Report

**Current Behavior**
Under certain common conditions, use of the `shareReplay` operator will cause memory leaks. This is similar to what was documented in #5034, but although that issue was marked as resolved by #5044, it can be observed in the linked example that the problem persists to a considerable degree.

**Reproduction**
https://stackblitz.com/edit/angular-a27w1c
https://angular-a27w1c.stackblitz.io/
1. Open up https://angular-a27w1c.stackblitz.io/ in Chrome
2. Use the checkboxes to create and destroy each of the components. 
3. Once they are all destroyed, use the devtools Memory tab to take a heap snapshot, and search for the word "Component".
4. Observe that while the `AsyncSingleComponent` was indeed destroyed, the `SyncComponent` and the `AsyncContinuousComponent` have stuck around.

**Expected behavior**
Each of the three components should have their class instances garbage-collected.

**Environment**
- See Stackblitz above
- RxJS version: 6.6.3

josepot:
Hi @MaximSagan,

I just spent some time looking into this. The only issue is that the last time that I haven't touched Angular since AngularJS 😬 , so forgive me if I'm missing something here.

Unless I'm missing something, I think that the only thing that seems problematic is the fact that the `SyncComponent` is not being destroyed.

I think that it's understandable that `AsyncContinuousComponent` is not being destroyed, given the fact that `shareReplay(1)` does not (by default) use a `refcount`. Therefore, I think that the `AsyncContinuousComponent` should be destroyed if its underlying stream was enhanced using `shareReplay({bufferSize: 1, refCount: true})`, instead? Again, I'm not familiar with Angular, so I could be missing something here.

I just opened #5932, which I think that should solve (at least) the issue with the memory leaks produced by streams that complete synchronously.

MaximSagan:
Hey @josepot, I really appreciate your quick response and PR.

> I think that it's understandable that AsyncContinuousComponent is not being destroyed, given the fact that shareReplay(1) does not (by default) use a refcount.

[...]

PR #5932, opened 2020-12-14
fix(shareReplay): handle possible memory leaks

josepot:
**Description:**
Fixes possible memory leaks caused by the `shareReplay` operator

#5931

### DIFF
--- a/spec/operators/shareReplay-spec.ts
+++ b/spec/operators/shareReplay-spec.ts
@@ -262,4 +262,26 @@ describe('shareReplay operator', () => {
     expectObservable(result).toBe(expected);
   });
 
+  const FinalizationRegistry = (global as any).FinalizationRegistry;
+  if (FinalizationRegistry) {
+
+    it('should not leak the subscriber for sync sources', (done) => {
+      const registry = new FinalizationRegistry((value: any) => {
+        expect(value).to.equal('callback');
+        done();
+      });
+      let callback: (() => void) | undefined = () => { /* noop */ };
+      registry.register(callback, 'callback');
+
+      const shared = of(42).pipe(shareReplay(1));
+      shared.subscribe(callback);
+
+      callback = undefined;
+      global.gc();
+    });
+
+  } else {
+    console.warn(`No support for FinalizationRegistry in Node ${process.version}`);
+  }
+
 });
--- a/src/internal/operators/shareReplay.ts
+++ b/src/internal/operators/shareReplay.ts
@@ -1,8 +1,8 @@
-import { Observable } from "../Observable";
-import { ReplaySubject } from "../ReplaySubject";
-import { Subscription } from "../Subscription";
-import { MonoTypeOperatorFunction, SchedulerLike } from "../types";
-import { Subscriber } from "../Subscriber";
+import { Observable } from '../Observable';
+import { ReplaySubject } from '../ReplaySubject';
+import { Subscription } from '../Subscription';
+import { MonoTypeOperatorFunction, SchedulerLike } from '../types';
+import { Subscriber } from '../Subscriber';
 
 export interface ShareReplayConfig {
   bufferSize?: number;
@@ -70,7 +70,7 @@ export function shareReplay<T>(
   scheduler?: SchedulerLike
 ): MonoTypeOperatorFunction<T> {
   let config: ShareReplayConfig;
-  if (configOrBufferSize && typeof configOrBufferSize === "object") {
+  if (configOrBufferSize && typeof configOrBufferSize === 'object') {
     config = configOrBufferSize as ShareReplayConfig;
   } else {
     config = {
--- a/src/internal/operators/shareReplay.ts
+++ b/src/internal/operators/shareReplay.ts
@@ -1,8 +1,8 @@
-import { Observable } from '../Observable';
-import { ReplaySubject } from '../ReplaySubject';
-import { Subscription } from '../Subscription';
-import { MonoTypeOperatorFunction, SchedulerLike } from '../types';
-import { Subscriber } from '../Subscriber';
+import { Observable } from "../Observable";
+import { ReplaySubject } from "../ReplaySubject";
+import { Subscription } from "../Subscription";
+import { MonoTypeOperatorFunction, SchedulerLike } from "../types";
+import { Subscriber } from "../Subscriber";
 
 export interface ShareReplayConfig {
   bufferSize?: number;
@@ -56,22 +56,28 @@ export interface ShareReplayConfig {
  * @method shareReplay
  * @owner Observable
  */
-export function shareReplay<T>(config: ShareReplayConfig): MonoTypeOperatorFunction<T>;
-export function shareReplay<T>(bufferSize?: number, windowTime?: number, scheduler?: SchedulerLike): MonoTypeOperatorFunction<T>;
+export function shareReplay<T>(
+  config: ShareReplayConfig
+): MonoTypeOperatorFunction<T>;
+export function shareReplay<T>(
+  bufferSize?: number,
+  windowTime?: number,
+  scheduler?: SchedulerLike
+): MonoTypeOperatorFunction<T>;
 export function shareReplay<T>(
   configOrBufferSize?: ShareReplayConfig | number,
   windowTime?: number,
   scheduler?: SchedulerLike
 ): MonoTypeOperatorFunction<T> {
   let config: ShareReplayConfig;
-  if (configOrBufferSize && typeof configOrBufferSize === 'object') {
+  if (configOrBufferSize && typeof configOrBufferSize === "object") {
     config = configOrBufferSize as ShareReplayConfig;
   } else {
     config = {
       bufferSize: configOrBufferSize as number | undefined,
       windowTime,
       refCount: false,
-      scheduler
+      scheduler,
     };
   }
   return (source: Observable<T>) => source.lift(shareReplayOperator(config));
@@ -81,7 +87,7 @@ function shareReplayOperator<T>({
   bufferSize = Number.POSITIVE_INFINITY,
   windowTime = Number.POSITIVE_INFINITY,
   refCount: useRefCount,
-  scheduler
+  scheduler,
 }: ShareReplayConfig) {
   let subject: ReplaySubject<T> | undefined;
   let refCount = 0;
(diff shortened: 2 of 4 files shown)


josepot:
I tested this locally, and the changes in this PR don't fix the memory leak reported on #5931, so I'm closing it for now. I wil re-open it if I find a solution.

josepot:
Actually, sorry, [it *does* fix the memory leak](https://stackblitz.com/edit/angular-7jmpmo?file=src%2Fapp%2Fvalue.service.ts)

MaximSagan:
I believe this PR should be merged as there is no reason to keep those subscription objects around any longer than necessary, but I think it doesn't get to the root of the problem here.

That is, we have to ask why an instance of `Subscriber` (which is what `Observable.prototype.subscribe(observer)` returns) is maintaining any reference at all to its `Observer` after unsubscription (including complete & error).

With that in mind, I have made this related PR, so that this kind of thing will (hopefully) never be a problem again: #5934

src/internal/operators/shareReplay.ts
```diff
@@ -108,13 +108,17 @@ function shareReplayOperator<T>({
...
+        subscription = undefined;
+      }
     } else {
       innerSub = subject.subscribe(this);
     }
 
     this.add(() => {
       refCount--;
       innerSub.unsubscribe();
+      innerSub = undefined;
```
benlesh, line 121: This isn't necessary, really. As there would be no more references to `innerSub` held after this, and it would get GC'ed.

src/internal/operators/shareReplay.ts
```diff
@@ -108,13 +108,17 @@ function shareReplayOperator<T>({
           subject.complete();
         },
       });
+      if (isComplete) {
+        subscription = undefined;
+      }
```
benlesh, line 113: I'm not sure how this solves a memory leak. Do we have tests or proof for this?

The *only* way this conditional passes is if the `source` completed synchronously. Which, if you [look right here](https://github.com/ReactiveX/rxjs/blob/1e1e022ed9fdc4a4bfed63cf151887031ac57398/src/internal/operators/shareReplay.ts#L106-L107), the `isComplete` is set to `true` and the `subscription` is set to `undefined` during that very same completion. 

In other words, this work has already been done by the time that `isComplete` would ever be `true`. So we can remove this statement entirely.

benlesh (commented):
These changes don't at all affect the semantics of what's going on in this operator, and wouldn't really effect the memory usage. 

As I stated in the related issue, it seems the problem is confusion around the default behavior of `shareReplay`. Which, IMO, has been an ongoing problem.

Hopefully the new operators in version 7 will gain more adoption and we can leave these confusing days behind us.

benlesh:
I'm going to close this PR because the changes don't really do anything (as I outlined above). If there's proof that this changes the memory usage, I'll be happy to take another look.

src/internal/operators/shareReplay.ts
josepot: > I'm not sure how this solves a memory leak. Do we have tests or proof for this?

Yes, we do, in [this link](https://stackblitz.com/edit/angular-7jmpmo?file=src%2Fapp%2Fvalue.service.ts). If you follow the reproduction steps of [this comment](https://github.com/ReactiveX/rxjs/issues/5931#issue-765399108), then you will notice that the memory leak is not present with these changes.

> The only way this conditional passes is if the source completed synchronously. 

Exactly!

> Which, if you look right here, the isComplete is set to true and the subscription is set to undefined during that very same completion.

That's what it _looks_ like. However, that's not what it actually happens... The thing is that when it completes synchronously, then in that moment `subscription` is `undefined`, because its value has not been assigned yet. It will be assigned after the callback gets executed. Therefore, the `if` statement is necessary to prevent the memory leak.

josepot:
I'm sorry if I'm being annoying, but @benlesh or @cartant could you please have a look at [this codesandbox](https://codesandbox.io/s/beautiful-wood-m2h9o?file=/src/index.ts) :pray:. It's simply doing a `of(1).pipe(shareReplay()).subscribe()` while using the current version of `shareReplay` on v6 with 2 `console.log`s that illustrate why the `if (isComplete)` statement that I suggested in this PR is necessary in order to avoid the memory leak.

Also, thanks a lot for the amazing work that you are doing on v7. I've been using it on a project of mine and so far things are looking great! There is just one little thing that's annoying me: which is that I'm seeing code in a chunk that I really think that it should have been "tree shaken" and it wasn't. I'm still investigating, I will open an issue or a discussion in the coming days to explain this better and I will make sure that when I open the issue I can give you as much info as I can. Thanks again!

cartant:
@josepot I'll have a look at this in a bit. I'll push a test to this branch to clear things up.

benlesh:
Derp. @josepot is correct. I overlooked that synchronously, `subscription` is set *after* it's set to `undefined` when the `subscribe` call returns. Reopening this.

src/internal/operators/shareReplay.ts
benlesh: Yeah. I was wrong. Sorry about that, I should have looked more carefully. Thank you for being diligent.

Here's what would have avoided my mistake: A detailed comment above this check like:

```ts
// Here we need to check to see if the source synchronously completed. Although
// we're setting `subscription = undefined` in the completion handler, if the source
// is synchronous, that will happen *before* subscription is set by the return of
// the `subscribe` call.
```

benlesh (changes requested):
1. Add a comment, as suggested, above the new conditional.

2. Add a test that proves this resolves the issue.

RELATED: we'll need to create a second PR that adds the same test to `master`.

src/internal/operators/shareReplay.ts
josepot: No worries! I should have added a comment in the first place, since this is very counter intuitive.

josepot:
> Add a comment, as suggested, above the new conditional.

:heavy_check_mark: 

> Add a test that proves this resolves the issue.

Is that possible without testing the internals? :thinking: I mean, the only way that I can think off, involves exposing an implementation detail... which IMO wouldn't be worth it. I'm more than open to suggestions.

cartant:
> Is that possible without testing the internals? 🤔

Yes. I'll add one soon.

cartant:
@josepot @benlesh I've added a leak test for `shareReplay` to effect a failure, you will have to comment out the nulling of `innerSub` and `subscription` (although, the latter is, IMO, a smaller leak that is not related to the problem demonstrated in the Angular StackBlitz repro). Note that `FinalizationRegistry` is only in Node 14 and later. I'll add a similar test to `master` soon.

cartant:
IDK what these build failures are. I hate CI so much.

josepot:
@cartant hats off to you for that test!! :clap: :clap: :clap: TIL about [`FinalizationRegistry`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry), amazing! Thanks a lot!

PS @voliva [check this out](https://github.com/ReactiveX/rxjs/pull/5932/files#diff-3b05f28653d60d8c6acca2c70bee58c9c8dad00078f05a7e363e6f9253dd792d) we should start using this on `@react-rxjs/core` ASAP.

cartant:
@josepot IDK whether or not you can see the details of the CI failure, but this is what's breaking:

```
Error: node_modules/@types/node/index.d.ts(74,11): error TS2300: Duplicate identifier 'IteratorResult'.
Error: node_modules/typescript/lib/lib.es2015.iterable.d.ts(41,6): error TS2300: Duplicate identifier 'IteratorResult'.
npm ERR! code 2
npm ERR! path /home/runner/work/rxjs/rxjs
npm ERR! command failed
npm ERR! command sh -c tsc -p ./tsconfig/compat/tsconfig.cjs.json
```

🤷‍♂️ Maybe there is was a breaking change in the node types that match whatever semver is in this branch's `package.json`. IDK.

josepot:
> @josepot IDK whether or not you can see the details of the CI failure, but this is what's breaking:
> 
> ```
> Error: node_modules/@types/node/index.d.ts(74,11): error TS2300: Duplicate identifier 'IteratorResult'.
> Error: node_modules/typescript/lib/lib.es2015.iterable.d.ts(41,6): error TS2300: Duplicate identifier 'IteratorResult'.
> npm ERR! code 2
> npm ERR! path /home/runner/work/rxjs/rxjs
> npm ERR! command failed
> npm ERR! command sh -c tsc -p ./tsconfig/compat/tsconfig.cjs.json
> ```
> 
>  Maybe there is was a breaking change in the node types that match whatever semver is in this branch's `package.json`. IDK.

Thanks a lot @cartant ! Yep, I can see that. There were 2 different checks failing, one has been fixed by my latest commit (a linting issue, aparently). And this one, I'm investigating... The first thing that I tried was rebasing the branch from 6.x, just in case, but that didn't help... And yeah, based on a quick google search that error normally indicates an issue with the node types. I will investigate a bit more later. Thanks again.

josepot:
> man_shrugging Maybe there is was a breaking change in the node types that match whatever semver is in this branch's package.json. IDK.

Yep, I think that's the issue... Because -according to the logs- the task that fails is running under node@13.14.0:

```sh
/usr/bin/tar xz --warning=no-unknown-keyword -C /home/runner/work/_temp/7bd7433d-ba81-4237-ac22-1ea42ffa579b -f /home/runner/work/_temp/1e3c4cf6-32b1-405a-90d3-3bb7230d395f
/opt/hostedtoolcache/node/13.14.0/x64/bin/node --version
v13.14.0
```

And the types that are defined on the `package-lock.json` file are `9.4.5`.

Question: would it make sense to add the `skipLibCheck` for this task  `tsc -p ./tsconfig/compat/tsconfig.cjs.json`?

cartant (approved):

spec/operators/shareReplay-spec.ts
```diff
@@ -262,4 +262,26 @@ describe('shareReplay operator', () => {
     expectObservable(result).toBe(expected);
   });
 
+  const FinalizationRegistry = (global as any).FinalizationRegistry;
```
benlesh, line 265: This is a clever use of `FinalizationRegistry`.

benlesh (approved):

benlesh:
Great work, @josepot and @cartant. Thanks for seeing this through!

PR #5933, opened 2020-12-14
refactor(fromFetch): remove redundant if wrap of init

iamandrewluca:
**Description:**

`init` object is created using rest operator
this means that `init` always will be an object which is a truthy value
this commit removes redundant `init` check

Related: https://github.com/ReactiveX/rxjs/pull/5306/files#r542228847

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "rxjs",
-  "version": "7.0.0-beta.8",
+  "version": "7.0.0-beta.9",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/internal/observable/dom/fetch.ts
+++ b/src/internal/observable/dom/fetch.ts
@@ -87,7 +87,7 @@ export function fromFetch(input: string | Request, init?: RequestInit): Observab
  * ```
  *
  * @param input The resource you would like to fetch. Can be a url or a request object.
- * @param init A configuration object for the fetch.
+ * @param initWithSelector A configuration object for the fetch.
  * [See MDN for more details](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Parameters)
  * @returns An Observable, that when subscribed to performs an HTTP request using the native `fetch`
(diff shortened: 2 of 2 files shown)


iamandrewluca:
`package-lock.json` file was modified automatically when I executed `npm install`

kwonoj:
> this means that init always will be an object which is a truthy value

is it true? `RequestInit`'s property is all optional so this can happen: 

```
interface RequestInit {
    /**
     * A BodyInit object or null to set request's body.
     */
    body?: BodyInit | null;
    /**
     * A string indicating how the request will interact with the browser's cache to set request's cache.
     */
    cache?: RequestCache;
    /**
     * A string indicating whether credentials will be sent with the request always, never, or only when sent to a same-origin URL. Sets request's credentials.
     */
    credentials?: RequestCredentials;
    /**
     * A Headers object, an object literal, or an array of two-item arrays to set request's headers.
     */
    headers?: HeadersInit;
    /**
     * A cryptographic hash of the resource to be fetched by request. Sets request's integrity.
     */
    integrity?: string;
    /**
     * A boolean to set request's keepalive.
     */
    keepalive?: boolean;
    /**
     * A string to set request's method.
     */
    method?: string;
    /**
     * A string to indicate whether the request will use CORS, or will be restricted to same-origin URLs. Sets request's mode.
     */
    mode?: RequestMode;
    /**
     * A string indicating whether request follows redirects, results in an error upon encountering a redirect, or returns the redirect (in an opaque fashion). Sets request's redirect.
     */
    redirect?: RequestRedirect;
    /**
     * A string whose value is a same-origin URL, "about:client", or the empty string, to set request's referrer.
     */
    referrer?: string;
    /**
     * A referrer policy to set request's referrerPolicy.
     */
    referrerPolicy?: ReferrerPolicy;
    /**
     * An AbortSignal to set request's signal.
     */
    signal?: AbortSignal | null;
    /**
     * Can only be null. Used to disassociate request from any Window.
     */
    window?: any;
}
```

image

cartant (approved):

benlesh (approved):

benlesh:
thanks, @iamandrewluca

PR #5934, opened 2020-12-15
fix(Subscriber): handle memory leaks

MaximSagan:
**Description:**
Fixes memory leaks occurring when instances of `Subscriber` are kept around in memory even after unsubscription (which includes `complete()` and `error()`).

#5931

### DIFF
--- a/spec/operators/bufferWhen-spec.ts
+++ b/spec/operators/bufferWhen-spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
-import { of, EMPTY, Observable } from 'rxjs';
-import { bufferWhen, mergeMap, takeWhile, take } from 'rxjs/operators';
+import { of, throwError } from 'rxjs';
+import { bufferWhen, mergeMap, takeWhile } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
 
@@ -198,7 +198,7 @@ describe('bufferWhen operator', () => {
       const result = e1.pipe(
         bufferWhen(() => {
           if (i === 1) {
-            throw 'error';
+            return throwError(() => 'error');
           }
           return closings[i++];
         })
--- a/src/internal/Subscriber.ts
+++ b/src/internal/Subscriber.ts
@@ -106,6 +106,7 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
     if (!this.closed) {
       this.isStopped = true;
       super.unsubscribe();
+      this.destination = NOOP_OBSERVER;
     }
   }
 
@@ -208,5 +209,15 @@ export const EMPTY_OBSERVER: Readonly<Observer<any>> & { closed: true } = {
   closed: true,
   next: noop,
   error: defaultErrorHandler,
+  complete: noop,
+};
+
+/**
+ * The observer used as a stub for subscriptions that have already been closed.
+ */
+export const NOOP_OBSERVER: Readonly<Observer<any>> & { closed: true } = {
+  closed: true,
+  next: noop,
+  error: noop,
   complete: noop,
 };
--- a/src/internal/Subscription.ts
+++ b/src/internal/Subscription.ts
@@ -66,6 +66,7 @@ export class Subscription implements SubscriptionLike {
       }
 
       const { initialTeardown } = this;
+      this.initialTeardown = undefined;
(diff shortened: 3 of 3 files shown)


MaximSagan:
@benlesh Hi.

spec/operators/bufferWhen-spec.ts
```diff
@@ -198,7 +198,7 @@ describe('bufferWhen operator', () => {
       const result = e1.pipe(
         bufferWhen(() => {
           if (i === 1) {
-            throw 'error';
+            return throwError(() => 'error');
```
benlesh, line 201: This complete changes the test. Revert.

src/internal/Subscriber.ts
```diff
@@ -106,6 +106,7 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
     if (!this.closed) {
       this.isStopped = true;
       super.unsubscribe();
+      this.destination = NOOP_OBSERVER;
```
benlesh, line 109: at the point we've unsubscribed, it seems like we could just set `this.destination = null!` here, although I haven't tried that and tested.

src/internal/Subscriber.ts
```diff
@@ -210,3 +211,13 @@ export const EMPTY_OBSERVER: Readonly<Observer<any>> & { closed: true } = {
   error: defaultErrorHandler,
   complete: noop,
 };
+
+/**
+ * The observer used as a stub for subscriptions that have already been closed.
+ */
+export const NOOP_OBSERVER: Readonly<Observer<any>> & { closed: true } = {
```
benlesh, line 218: We can probably remove this if we're just nulling out `destination`.

src/internal/Subscription.ts
```diff
@@ -66,6 +66,7 @@ export class Subscription implements SubscriptionLike {
       }
 
       const { initialTeardown } = this;
+      this.initialTeardown = undefined;
```
benlesh, line 69: I'd use `null!` here, I think (not sold on it), just because `null` tells me that _we_ set it, rather than it just never being set.

benlesh (changes requested):
One thing is for sure, we don't want to change the semantics of the test I commented on. Overall I see what's being done here and it makes sense.

I think we really want to test `destination` to `null` though, and then maybe even have a test that confirms that is the case via introspection? The test wouldn't be super valuable, other than to serve as a reminder that we want to be mindful of that.

benlesh:
Okay, I'm looking more into this. It would be great to have a reproduction that didn't involve Angular. But I'm totally certain that the code here isn't the right solution. Using the "noop observer" that this adds causes errors we'd like to propagate _somehow_ to simply be swallowed.

Again, I _think_ I see what you're getting at, but it's hard without a minimal reproduction, and it's also plausible that this is something that only has to do with the `shareReplay` implementation (which indeed intentionally keeps a subscription around by default by design, for better or worse)

cc @cartant

benlesh:
I'm going to close this for now, because I don't want to send the wrong signal to the author, @MaximSagan, to continue working on it. What we _really_ need right now is a _minimal_, RxJS only, reproduction that definitively shows the leak. I'll try to figure that out too.

benlesh:
Also to be clear: I'm not saying that there isn't a leak. I'm saying that we need to more definitively identify the cause rather than blindly address the symptom.

spec/operators/bufferWhen-spec.ts
MaximSagan: @benlesh Yeah I assumed this wasn't the right way to go about it, but did it on the off-chance that the test itself was previously incorrect (as I figured that `bufferWhen`'s `closingSelector` was supposed to return an observable).

src/internal/Subscriber.ts
MaximSagan: I believe simply nulling it out caused several tests to fail.

MaximSagan:
@benlesh Thanks for looking into this. Not offended that you're not going with my solution, haha. Really just wanted to get this issue on the radar, and I figured a PR might help.
I'll prepare an RxJS-only reproduction shortly.

benlesh:
@MaximSagan one thing to look for when trying to figure out what is or isn't cleaning up is tests with an `expectSubscriptions` check, especially WRT inner observables. If you think you see a memory leak, it's easiest to prove that, within RxJS by adding checks to make sure inner subscriptions are unsubscribed at the proper time.

PR #5937, opened 2020-12-16
Update filter.ts

leggechr:
**Description:**
Remove helpful type at the beginning because it actually was incorrectly returning type never in some cases.


cartant (commented):
IMO, this change should be accompanied by a failing test that's fixed by the removal of the signature - i.e. something that actually demonstrates the problem.

MaximSagan:
I would think that if your predicate unconditionally returns `false`, `never` would be the correct type

cartant:
I suspect that predicates that return `any` will match the first signature, but I'd like to see a failing test that verifies this - especially, as these types of signatures are used elsewhere in the lib.

cartant:
@leggechr I've added a test that fails if the signature is not removed. If this isn't representative of the code that effected the bug, could you please add a test for that use case, too?

@benlesh This is going to be a problem with any other signatures that also rely upon matching predicates that return `false`. Any such signatures will need to be removed.

leggechr:
Hey! Yes that is the correct use case. Thanks for updating.

benlesh:
@cartant it looks like this problem may exist with `first`, `last`, `single`, `takeWhile`, and `iif`

benlesh:
Related: https://github.com/ReactiveX/rxjs/pull/5987

benlesh (approved):

PR #5939, opened 2020-12-19
feat: Undoable Subject

KevLeong:
**Description:**
Hope the official can support undo/redo, because it is very common in editor scenarios.

**React Demo**
```jsx
import React from 'react';
import { UndoableSubject } from 'rxjs';
import { useObservable } from 'react-use';

const Demo = () => {
  const input$ = React.useMemo(() => new UndoableSubject(''), []);
  const value = useObservable(input$);
  return (
    <React.Fragment>
      <button disabled={!value.past.length} onClick={input$.undo}>undo</button>
      <button disabled={!value.future.length} onClick={input$.redo}>redo</button>
      <input onChange={event => input$.next(event.target.value)} value={value.present} />
    </React.Fragment>
  )
};
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -9,6 +9,7 @@ export { animationFrames } from './internal/observable/dom/animationFrames';
 /* Subjects */
 export { Subject } from './internal/Subject';
 export { BehaviorSubject } from './internal/BehaviorSubject';
+export { UndoableSubject } from './internal/UndoableSubject';
 export { ReplaySubject } from './internal/ReplaySubject';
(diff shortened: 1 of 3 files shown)


benlesh:
This is certainly cool. However we're not looking to add any features like this at the moment. It belongs in a user-land library. If it ends up being very popular in the userland, we can discuss perhaps adding it into the core library.

The issue with adding any and all new features into the core library is that we're then on the hook for maintaining them, for better or worse.

Thank you for the very interesting PR. I really do appreciate it, @KevLeong!

PR #5943, opened 2020-12-25
Fix errors and warnings when building docs + fix links

jakovljevic-mladen:
**Description:**
By commits:

- Removed URL from navigation pointing to an already removed page that was mistakenly reverted back to `navigation.json` in #5814. The error when building `docs`:
image
is now gone.
- Fixed links on `Subscription` docs.
- Fixed link to `asObservable` in `publishReply`.
- Fixed `@return` statement in `windowTime` what had bad formatting producing warning when building docs.
- Adding new disambiguator that fixed links to `EMPTY` and `timestamp` because there were multiple documents pointing to them. New disambiguator does not prefer members. Taken from [Angular](https://github.com/angular/angular/pull/24000). Examples:
  - `operators/timeout` vs `ajax/AjaxRequest.timeout`,
  - `EMPTY` vs `Subscription.EMPTY`.

None

### DIFF
--- a/src/internal/operators/zipAll.ts
+++ b/src/internal/operators/zipAll.ts
@@ -8,8 +8,8 @@ import { joinAllInternals } from './joinAllInternals';
  * it will subscribe to all inner sources, combining their values by index and emitting
  * them.
  *
- * {@see zipWith}
- * {@see zip}
+ * @see {@link zipWith}
+ * @see {@link zip}
  */
 export function zipAll<T>(): OperatorFunction<ObservableInput<T>, T[]>;
 export function zipAll<T>(): OperatorFunction<any, T[]>;
--- a/docs_app/content/navigation.json
+++ b/docs_app/content/navigation.json
@@ -80,11 +80,6 @@
       ]
     },
     {
-      "url": "resources",
-      "title": "External Resources",
-      "tooltip": "External Resources"
-    },
-    {
       "url": "code-of-conduct",
       "title": "Code of Conduct",
       "tooltip": "Code of Conduct"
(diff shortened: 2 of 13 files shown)


jakovljevic-mladen:
cc @niklas-wortmann

src/internal/operators/zipAll.ts
```diff
@@ -8,8 +8,8 @@ import { joinAllInternals } from './joinAllInternals';
  * it will subscribe to all inner sources, combining their values by index and emitting
  * them.
  *
- * {@see zipWith}
- * {@see zip}
+ * @see {@link zipWith}
```
benlesh, line 11: Interesting. Is this required for this to work?

jakovljevic-mladen: Yeah, the `docs` script from `docs_app/package.json` produces warning:

image

And this is what it looks like in the docs (without this change):

- zipAll
image
- ReplySubject
image

I have now checked, there are some `@see` tags without `@link`, which I'll have to address as well.

jakovljevic-mladen: Done ☝️

benlesh:
non-Christmas ping to @niklas-wortmann

niklas-wortmann (approved):
Wow that's awesome. Thanks a lot for the fixes! LGTM

benlesh:
Thanks again, @jakovljevic-mladen!

jakovljevic-mladen:
@benlesh You're welcome Ben, I'm happy that I can help.

Could you maybe upload the latest docs from master to next.rxjs.dev? It appears that the latest pushed commit on the "next" was from October last year. Is this link official anyway or not yet?

PR #5945, opened 2020-12-29
fix(fromEvent): fixed HasEventTargetAddRemove to support EventTarget types

### LINKED ISSUE #5944: `fromEvent` / `FromEventTarget` signatures don't work  with basic event targets
## Bug Report

**Current Behavior**
```ts
const eventTarget = undefined as any as EventTarget;
const x: FromEventTarget<any> = eventTarget;
```

```
Type 'EventTarget' is not assignable to type 'FromEventTarget<any>'.
  Type 'EventTarget' is not assignable to type 'HasEventTargetAddRemove<any>'.
    Types of property 'removeEventListener' are incompatible.
      Type '(type: string, callback: EventListener | EventListenerObject | null, options?: boolean | EventListenerOptions | undefined) => void' is not assignable to type '(type: string, listener?: ((evt: any) => void) | null | undefined, options?: boolean | EventListenerOptions | undefined) => void'.
        Types of parameters 'callback' and 'listener' are incompatible.
          Type '((evt: any) => void) | null | undefined' is not assignable to type 'EventListener | EventListenerObject | null'.
            Type 'undefined' is not assignable to type 'EventListener | EventListenerObject | null'.
```

Note: this seems to be only for the `removeEventListener` signature, not `addEventListener`

**Expected behavior**

```ts
fromEvent(someEventTarget, 'some-event'); // compiles
```

**Environment**

- RxJS version: 7.0.0-beta.8
- TypeScript version: 4.1.3

**Possible Solution**

Changing the signature to be able to match with lib.dom.d.ts' `EventTarget.removeEventListener`

rraziel:
**Description:**

Makes it possible to use `fromEvent()` with a simple `EventTarget`.

Closes #5944

### DIFF
--- a/spec-dtslint/observables/fromEvent-spec.ts
+++ b/spec-dtslint/observables/fromEvent-spec.ts
@@ -1,30 +1,35 @@
 import { fromEvent } from 'rxjs';
 import { JQueryStyleEventEmitter } from '../../src/internal/observable/fromEvent';
-import { A, B } from "../helpers";
+import { A, B } from '../helpers';
 
-declare const eventTargetSource: HTMLDocument;
-
-interface NodeStyleSource {
-  addListener: (eventName: string | symbol, handler: (...args: any[]) => void) => this;
-  removeListener: (eventName: string | symbol, handler: (...args: any[]) => void) => this;
-};
-declare const nodeStyleSource : NodeStyleSource;
-
-declare const nodeCompatibleSource: {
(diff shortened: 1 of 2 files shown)


cartant (approved):
LGTM, thanks!

benlesh (approved):

benlesh:
Thanks, @rraziel

PR #5963, opened 2021-01-18
fix(combineLatest): Ensure `EMPTY` is returned if no observables are …

benlesh:
…passed.

This resolves an issue found by testing Angular router. There was code that expected this behavior (rightfully so), and it was broken during a refactor in v7.


cartant (approved):
LGTM. Do any other operators - like `zip` - have this issue or lack tests like these?

benlesh:
@cartant that's worth a look. The issue I found was code that was like `combineLatest(arrObservablesBuiltElsewhere)`.

Either way, this PR solve's Angular's issue.

PR #5980, opened 2021-01-25
fix: Resolve issues with deprecated synchronous error handling and chained operators

### LINKED ISSUE #5979: Errors are squeltched when thrown in `next`, after any operator, while useDeprecatedSynchronousErrorHandling = true;
In RxJS 7, if you set `useDeprecatedSynchronousErrorHandling = true`, and you have a source that uses any operator, when you throw an error in the next handler, it's not rethrown. The first example was with `take`, but I've confirmed it also happens with `map`, `concatMap`, and others.

https://stackblitz.com/edit/rxjs-fuauce?devtoolsheight=60

```ts
import { of, config } from "rxjs";
import { take } from "rxjs/operators";

console.clear();

try {
  of("test")
    .pipe(take(1))
    .subscribe({
      next: value => {
        console.log("value: ", value);
        throw new Error("test");
      }
    });
} catch (err) {
  // This is never hit in v7, but it is hit in v6.
  console.log("Caught!", err.message);
}
```

We'll need to check `takeUntil`, `takeWhile`, and `first` for the same issues.

cc @leggechr

benlesh:
Adds some logic to handle the deprecated case. This is unfortunate, but it's necessary to properly support people who are still using this feature until we can remove it. Note that there is potentially a severe performance penalty for people using this feature. But they're being warned at every turn, and the migration path to get away from this is easy: Stop relying on synchronously thrown errors.

fixes #5979

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -608,6 +608,27 @@ describe('Observable', () => {
         }).to.throw('error!');
       });
 
+
+      // From issue: https://github.com/ReactiveX/rxjs/issues/5979
+      it('should still rethrow synchronous errors from next handlers on synchronous observables', () => {
+        expect(() => {
+          of('test').pipe(
+            // Any operators here
+            map(x => x + '!!!'),
+            map(x => x + x),
+            map(x => x + x),
+            map(x => x + x),
+          ).subscribe({
+            next: () => {
+              throw new Error(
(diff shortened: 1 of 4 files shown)


kwonoj (approved):
Ugh.

cartant (approved):
Looking forward to this being deleted.

PR #5984, opened 2021-01-26
fix: useDeprecatedSynchronousErrorThrowing honored for flattened sync sources

### LINKED ISSUE #5983: Super gross mode no longer sync throws flattened errors
Given this code: 

```ts
import { throwError, of, config } from "rxjs";
import { concatMap } from "rxjs/operators";

console.clear();

try {
  of(1)
    .pipe(concatMap(() => throwError(new Error("test"))))
    .subscribe(console.log);
} catch (err) {
  console.log("Caught!", err);
}
```

It should log `"Caught!"` and the error, but it does not.

Related: #5979 

cc @leggechr

benlesh:
fixes #5983

### DIFF
--- a/src/internal/Subscriber.ts
+++ b/src/internal/Subscriber.ts
@@ -132,46 +132,40 @@ export class SafeSubscriber<T> extends Subscriber<T> {
   ) {
     super();
 
-    // If we don't have arguments, or the observer passed is already EMPTY_OBSERVER,
-    // use EMPTY_OBSERVER. This is just to save a little on object allocations.
-    this.destination = EMPTY_OBSERVER;
-    if ((observerOrNext || error || complete) && observerOrNext !== EMPTY_OBSERVER) {
-      // We've got either functions or an observer to deal with
-      // let's figure that out here.
-
-      let next: ((value: T) => void) | undefined;
-      if (isFunction(observerOrNext)) {
-        next = observerOrNext;
-      } else if (observerOrNext) {
-        // Even if it's an observer, we have to pull the handlers off and
-        // capture the owner object as the context. That is because we're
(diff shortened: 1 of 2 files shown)


src/internal/Subscriber.ts
```diff
@@ -132,46 +132,40 @@ export class SafeSubscriber<T> extends Subscriber<T> {
...
+      next = next?.bind(context);
+      error = error?.bind(context);
+      complete = complete?.bind(context);
     }
+
+    // Once we set the destination, the superclass `Subscriber` will
+    // do it's magic in the `_next`, `_error`, and `_complete` methods.
+    this.destination = {
+      next: next ? maybeWrapForDeprecatedSyncErrorHandling(next, this) : noop,
+      error: maybeWrapForDeprecatedSyncErrorHandling(error ? error : defaultErrorHandler, this),
```
benlesh, line 166: The meat of the change is here.

src/internal/Subscriber.ts
```diff
@@ -132,46 +132,40 @@ export class SafeSubscriber<T> extends Subscriber<T> {
   ) {
     super();
 
-    // If we don't have arguments, or the observer passed is already EMPTY_OBSERVER,
-    // use EMPTY_OBSERVER. This is just to save a little on object allocations.
-    this.destination = EMPTY_OBSERVER;
-    if ((observerOrNext || error || complete) && observerOrNext !== EMPTY_OBSERVER) {
```
benlesh, line 138: After some thought, I realized that this check, and setting `this.destination = EMPTY_OBSERVER` was pointless, so that got removed.

It might be better to review this with "ignore whitespace" on.

kwonoj (approved):
ugh(2).

PR #5987, opened 2021-01-26
fix: predicates that return `any` will now behave property in TS

### LINKED ISSUE #5986: Some operators return the wrong type if a predicate returns `any` in TypeScript.
From [this comment](https://github.com/ReactiveX/rxjs/pull/5937#issuecomment-767727907) in #5937.

Basically, if a predicate returns `any`, it will match the first `true` or `false` return value, even if it's always returning the opposite value. This results in the incorrect type being inferred with current type overloads.

benlesh:
Resolves #5986


niklas-wortmann (approved):
LGTM

PR #5989, opened 2021-01-28
docs: update marble testing documentation

jakovljevic-mladen:
**Description:**

Updated `marble-testing.md` file so that:
- an example `'generates the stream correctly'` is fully converted to run mode: moved spaces inside marble strings, used `time` helper function to visually align marbles, added import for an operator, removed passing a scheduler to the operator.
- removed ASCII reference since marble diagrams are not ASCII anymore, right? Since #5907?
- added missing `time` and `animate` to the destructuring assignment (they were both explained below, but were never added). Also, I reordered detailed explanation so that it follows ordering in the destructuring assignment.
- other minor fixes like adding links, fixing typos and so on (some of the fixes were done by Prettier after the commit, like adding parenthesis around callback arguments,and like changing `*` to `-` or `*not*` to `_not_` or rearranged code lines from the examples).

Removed `internal-marble-tests.md` file: the reference to this file was removed from navigation a while back with [5422](5422/files#diff-e6656b105dad26983f3dd8960952d27e954cf2f0d8b87d9cc6892376aa1e5053L60). Since it was removed, I thought removing the file itself and the references to it would be a good idea. Also, since internal unit test are now (slowly) being migrated to run mode, tests that use TestScheduler without run mode might become deprecated, so there is no need for this file also. Please let me know if this file should be kept any longer.

Fixed font weight so that stuff between `<strong></strong>` tags becomes **bolded**.

None


docs_app/content/guide/testing/marble-testing.md
```diff
@@ -278,14 +272,14 @@ On a related note, you also can't currently assert delays of zero, even with `As
...
+The `TestScheduler` has been around since v5, but was actually intended for testing RxJS itself by the maintainers, rather than for use in regular user apps. Because of this, some of the default behaviors and features of the TestScheduler did not work well (or at all) for users. In v6 we introduced the `testScheduler.run(callback)` method which allowed us to provide new defaults and features in a non-breaking way, but it's still possible to use the TestScheduler outside of `testScheduler.run(callback)`. It's important to note that if you do so, there are some major differences in how it will behave.
```
cartant, line 275: I think we should keep the link to the old documentation, but change it to a blob-specific URL - https://github.com/ReactiveX/rxjs/blob/7113ae4b451dd8463fae71b68edab96079d089df/docs_app/content/guide/testing/internal-marble-tests.md - so that the file can still be deleted in this PR.

docs_app/content/guide/testing/marble-testing.md
```diff
@@ -139,36 +137,32 @@ const expected = '-- 9ms a 9ms b 9ms (c|)';
 // or
 // const expected = '-----------a 9ms b 9ms (c|)';
 
-const result = cold(input).pipe(
-  concatMap(d => of(d).pipe(
-    delay(10)
-  ))
-);
+const result = cold(input).pipe(concatMap((d) => of(d).pipe(delay(10))));
```
cartant, line 140: Prettier makes this example more difficult to read. IMO, we should leave the formatting as-is. If you put `<!-- prettier-ignore -->` on the line before the snippet, Prettier should ignore it.

cartant (commented):
A couple of suggestions.

jakovljevic-mladen:
@cartant, thanks for suggestions, completely agreed on both of them. Thanks for pointing the proper Prettier ignore syntax for these files, I didn't know about it. One note: Prettier added parenthesis to callback function parameters in almost all examples, so I added parenthesis manually to the example that you have commented: `concatMap(d =>` -> `concatMap((d) =>` to be consistent with other examples in the file.
IMO, adding these parenthesis is a really annoying Prettier feature.

cartant (approved):
LGTM. Approving this, but I'll hold off on merging it until someone else okays the removal of `docs_app/content/guide/testing/internal-marble-tests.md`. AFAICT, it's not referenced anywhere else, so the removal looks fine to me, but we'll see.

benlesh (approved):

benlesh:
Thank you, @jakovljevic-mladen! This is very helpful!

PR #6001, opened 2021-02-07
feat(ajax): Add option for streaming progress

### LINKED ISSUE #2833: Rx.Observable.ajax onProgress only set for Upload
Hi there,

I hadn't been able to make `progressSubscriber.next` to fire on a **Rx.Observable.ajax** instance, when downloading. Checking the code and the log I found that you set `onprogress` for the upload but you removed the one for download.

As `window.XDomainRequest` is only available for [IE 8 & 9](https://developer.mozilla.org/es/docs/Web/API/XDomainRequest), then on modern browsers `onprogress` is not assigned.

https://github.com/ReactiveX/rxjs/blob/7bb8280181bf1d5b0a2d3338ede8bea5e30f64c7/src/observable/dom/AjaxObservable.ts#L332-L336

Did you avoid setting `xhr.onprogress = xhrProgress` for any specific reason? Or, Should it be there next to `xhr.upload.onprogress = xhrProgress` ?

Thanks
SntsDev

benlesh:
## Edit: I've moved the spec changes to another PR that should be reviewed first here: https://github.com/ReactiveX/rxjs/pull/6002

- Fixes a bunch of tests that were slightly off or had bad assumptions
- Adds two new features: `includeUploadProgress` and `includeDownloadProgress` that will add additional events to the output stream before the event that is usually emitted.

Resolves #2833

NOTE: We still need to add an example of usage.

### DIFF
--- a/api_guard/dist/types/ajax/index.d.ts
+++ b/api_guard/dist/types/ajax/index.d.ts
@@ -6,6 +6,8 @@ export interface AjaxConfig {
     createXHR?: () => XMLHttpRequest;
     crossDomain?: boolean;
     headers?: Readonly<Record<string, any>>;
+    includeDownloadProgress?: boolean;
+    includeUploadProgress?: boolean;
     method?: string;
     password?: string;
     progressSubscriber?: PartialObserver<ProgressEvent>;
@@ -43,13 +45,20 @@ export interface AjaxRequest {
 }
 
 export declare class AjaxResponse<T> {
-    readonly originalEvent: Event;
+    readonly loaded: number;
+    readonly originalEvent: ProgressEvent;
     readonly request: AjaxRequest;
     readonly response: T;
     readonly responseType: XMLHttpRequestResponseType;
     readonly status: number;
+    readonly total: number;
+    readonly type: string;
     readonly xhr: XMLHttpRequest;
-    constructor(originalEvent: Event, xhr: XMLHttpRequest, request: AjaxRequest);
+    constructor(
+    originalEvent: ProgressEvent,
+    xhr: XMLHttpRequest,
+    request: AjaxRequest,
+    type?: string);
 }
 
 export interface AjaxTimeoutError extends AjaxError {
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -863,7 +863,7 @@ describe('ajax', () => {
         { uploadProgressTimes: 3 }
       );
 
-      expect(spy).to.be.called.callCount(3);
+      expect(spy).to.be.called.callCount(4);
     });
 
     it('should emit progress event when progressSubscriber is specified', function () {
@@ -894,7 +894,7 @@ describe('ajax', () => {
         { uploadProgressTimes: 3 }
       );
 
-      expect(spy).to.be.called.callCount(3);
+      expect(spy).to.be.called.callCount(4);
     });
   });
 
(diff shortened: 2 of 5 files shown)


benlesh:
There will be a follow up refactor to make this smaller as well.

benlesh:
Blocked on merging https://github.com/ReactiveX/rxjs/pull/6002

src/internal/ajax/ajax.ts
```diff
@@ -268,8 +268,13 @@ export const ajax: AjaxCreationMethod = (() => {
...
 })();
 
+const DOWNLOAD = 'download';
+const LOADSTART = 'loadstart';
+const PROGRESS = 'progress';
+const LOAD = 'load';
+
 export function fromAjax<T>(config: AjaxConfig): Observable<AjaxResponse<T>> {
-  return new Observable<AjaxResponse<T>>((destination) => {
+  return new Observable((destination) => {
```
kolodny, line 277: Go TypeScript! Curious is anyone took a pass at trying to remove all inferable generics? cc @cartant

spec/observables/dom/ajax-spec.ts
```diff
@@ -1124,10 +1032,365 @@ describe('ajax', () => {
       });
     });
   });
+
+  describe('with includeDownloadProgress', () => {
```
kolodny, line 1036: Are there any tests that have a binary type? Not for this PR but may be something to consider adding

src/internal/ajax/ajax.ts
cartant: Nope. Not yet.

spec/observables/dom/ajax-spec.ts
```diff
@@ -1124,10 +1032,365 @@ describe('ajax', () => {
...
+        method: 'GET',
+        responseType: '',
+        timeout: 0,
+        url: '/flibbertyJibbett',
+        withCredentials: false,
+      };
+
+      expect(results).to.deep.equal([
+        {
+          type: 'download_loadstart',
```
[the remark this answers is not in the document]
benlesh: I thought about it. I think that we have to do `enum` and not `const enum`, and therefore we'd end up exporting an object. That doesn't really mesh with how events work elsewhere. I'll consider this a vote in favor of an enum, though.

api_guard/dist/types/ajax/index.d.ts
```diff
@@ -43,13 +45,20 @@ export interface AjaxRequest {
 }
 
 export declare class AjaxResponse<T> {
-    readonly originalEvent: Event;
+    readonly loaded: number;
```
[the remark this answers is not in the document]
benlesh: :) I did. You're looking at the golden file for api_guardian.

FWIW: It's the bytes loaded. It maps to the `ProgressEvent`'s `loaded` property.

spec/observables/dom/ajax-spec.ts
```diff
@@ -948,85 +948,6 @@ describe('ajax', () => {
     delete root.XMLHttpRequest.prototype.ontimeout;
   });
 
-  it('should work fine when XMLHttpRequest onprogress property is monkey patched', function () {
```
benlesh, line 951: These tests are no longer relevant, as they were only for when we were using the property method versions of event handlers like `xhr.onprogress`. Now we're using `xhr.addEventListener('progress', ...)`, so it's a non-issue. Since Zone.js is patching `EventTarget`, this should "just work". Also, hopefully in the not-so-distant future, Zone.js goes away.

spec/observables/dom/ajax-spec.ts
benlesh: Binary, as in streaming data? XHR doesn't really support that. Fetch does, though.

If you mean "binary" as in a binary data type like a `"Content-Type": "image/png"` or something... All of that is handled directly by the XHR we're using under the hood, and I'm not sure how I'd test that without an end-to-end test of some sort, which we aren't quite set up for. Right now we're mocking the XHR. I could test that within the confines of our mocks though, I guess.

benlesh: All  of that said, it should "just work", because in the event of binary data like that, the XHR in the browser will have a `responseType` of `"arraybuffer"`, and the `response` itself would be something like `Uint8Array`, which is used directly in our `AjaxResponse`, without any additional processing.

spec/observables/dom/ajax-spec.ts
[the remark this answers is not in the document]
cartant: FWIW, my vote would be against exporting an enum. If something is to be exported, I'd vote for a type that's a union of literals, rather than an enum.

src/internal/ajax/types.ts
```diff
@@ -166,12 +166,33 @@ export interface AjaxConfig {
...
+  includeDownloadProgress?: boolean;
+
+  /**
+   * If `true`, will emit all upload progress and load complete events as {@link AjaxUploadProgressEvents}
+   * from the observable. The final download event will also be emitted as a {@link AjaxDownloadCompleteEvent}
+   *
+   * If both this and {@link includeDownloadProgress} are `false`, then only the {@link AjaxResponse} will
+   * be emitted from the resulting observable.
+   */
+  includeUploadProgress?: boolean;
```
OliverJAsh, line 197: When this option is `true`, shouldn't `ajax` return `Observable< AjaxUploadProgressEvents | AjaxDownloadCompleteEvent | AjaxResponse<T>>` instead of `Observable<AjaxResponse<T>>`? Likewise for `includeDownloadProgress` and `AjaxProgressEvents`. /cc @benlesh

OliverJAsh: https://github.com/ReactiveX/rxjs/issues/6145

PR #6002, opened 2021-02-08
test: Fix/update the ajax tests to be a little more correct

benlesh:
Moved the spec updates to a different PR than #6001 so they can be reviewed separately. That way it's perhaps easier to tell they're doing the same thing.

### DIFF
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -6,12 +6,10 @@ import { TestScheduler } from 'rxjs/testing';
 import { noop } from 'rxjs';
 import * as nodeFormData from 'form-data';
 
-
-
 const root: any = (typeof globalThis !== 'undefined' && globalThis) || (typeof self !== 'undefined' && self) || global;
 
 if (typeof root.FormData === 'undefined') {
-  root.FormData = nodeFormData as any;  
+  root.FormData = nodeFormData as any;
 }
 
 /** @test {ajax} */
@@ -121,7 +119,7 @@ describe('ajax', () => {
     });
 
     afterEach(() => {
-      delete (global as any).document; 
+      delete (global as any).document;
     });
 
     it('should send the cookie with a custom header to the same domain', () => {
@@ -224,23 +222,22 @@ describe('ajax', () => {
 
   it('should error if createXHR throws', () => {
     let error;
-    const obj = {
+
+    ajax({
       url: '/flibbertyJibbet',
       responseType: 'text',
       createXHR: () => {
         throw new Error('wokka wokka');
       },
-    };
-
-    ajax(<any>obj).subscribe(
+    }).subscribe(
       () => {
-        throw 'should not next';
+        throw new Error('should not next');
       },
       (err: any) => {
         error = err;
       },
       () => {
-        throw 'should not complete';
+        throw new Error('should not complete');
       }
     );
 
@@ -282,7 +279,6 @@ describe('ajax', () => {
 
     ajax({
       url: '/flibbertyJibbet',
-      responseType: 'text',
       method: '',
     }).subscribe(
       (x: any) => {
@@ -298,16 +294,16 @@ describe('ajax', () => {
 
     MockXMLHttpRequest.mostRecent.respondWith({
       status: 200,
-      contentType: 'application/json',
+      responseType: 'json',
       responseText: JSON.stringify(expected),
     });
 
     expect(result!.xhr).exist;
-    expect(result!.response).to.deep.equal(JSON.stringify({ foo: 'bar' }));
+    expect(result!.response).to.deep.equal({ foo: 'bar' });
     expect(complete).to.be.true;
   });
 
-  it('should fail if fails to parse response', () => {
+  it('should fail if fails to parse response in older IE', () => {
     let error: any;
     const obj: AjaxConfig = {
       url: '/flibbertyJibbet',
@@ -315,22 +311,24 @@ describe('ajax', () => {
       method: '',
     };
 
+    // No `response` property on the object (for older IE).
+    MockXMLHttpRequest.noResponseProp = true;
+
     ajax(obj).subscribe(
       () => {
-        throw 'should not next';
+        throw new Error('should not next');
       },
       (err: any) => {
         error = err;
       },
       () => {
-        throw 'should not complete';
+        throw new Error('should not complete');
       }
     );
 
     MockXMLHttpRequest.mostRecent.respondWith({
       status: 207,
-      contentType: '',
-      responseType: '',
+      responseType: 'json',
       responseText: 'Wee! I am text, but should be valid JSON!',
     });
 
@@ -348,13 +346,13 @@ describe('ajax', () => {
 
     ajax(obj).subscribe(
       () => {
-        throw 'should not next';
+        throw new Error('should not next');
       },
       (err: any) => {
         error = err;
       },
       () => {
-        throw 'should not complete';
+        throw new Error('should not complete');
       }
     );
 
@@ -362,7 +360,7 @@ describe('ajax', () => {
 
     MockXMLHttpRequest.mostRecent.respondWith({
       status: 404,
-      contentType: 'text/plain',
+      responseType: 'text',
       responseText: 'Wee! I am text!',
     });
 
@@ -395,7 +393,7 @@ describe('ajax', () => {
 
     MockXMLHttpRequest.mostRecent.respondWith({
       status: 300,
-      contentType: 'text/plain',
+      responseType: 'text',
       responseText: 'Wee! I am text!',
     });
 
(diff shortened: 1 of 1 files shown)


spec/observables/dom/ajax-spec.ts
```diff
@@ -298,39 +294,41 @@ describe('ajax', () => {
 
     MockXMLHttpRequest.mostRecent.respondWith({
       status: 200,
-      contentType: 'application/json',
```
benlesh, line 301: `contentType` does nothing for this test. Here we assume that the browser is working as expected, and we deal with the `responseType`, which the browser sets for us. Since all of that process is mocked, switching to using `responseType` only in our tests.

spec/observables/dom/ajax-spec.ts
```diff
@@ -298,39 +294,41 @@ describe('ajax', () => {
...
+  it('should fail if fails to parse response in older IE', () => {
     let error: any;
     const obj: AjaxConfig = {
       url: '/flibbertyJibbet',
       responseType: 'json',
       method: '',
     };
 
+    // No `response` property on the object (for older IE).
+    MockXMLHttpRequest.noResponseProp = true;
```
benlesh, line 315: This is a new feature to the Mock that will strip the `response` property from the `xhr` to simulate an older browser that lacks that feature.

spec/observables/dom/ajax-spec.ts
```diff
@@ -298,39 +294,41 @@ describe('ajax', () => {
...
+
     ajax(obj).subscribe(
       () => {
-        throw 'should not next';
+        throw new Error('should not next');
```
benlesh, line 319: Just making sure that errors thrown are recognized and processed properly by the test framework.

spec/observables/dom/ajax-spec.ts
```diff
@@ -558,26 +555,17 @@ describe('ajax', () => {
       async: false,
     };
 
-    ajax(obj).subscribe(
-      (x: any) => {
```
benlesh, line 562: Nothing in here mattered, accept the check of `x.xhr.async`, but it's unremarkable that it's in the `subscribe` next callback. So I've cleaned this a bit to make the test simpler.

spec/observables/dom/ajax-spec.ts
```diff
@@ -558,26 +555,17 @@ describe('ajax', () => {
...
-    );
+    ajax(obj).subscribe();
 
-    const request = MockXMLHttpRequest.mostRecent;
+    const mockXHR = MockXMLHttpRequest.mostRecent;
 
-    expect(request.url).to.equal('/flibbertyJibbet');
+    expect(mockXHR.url).to.equal('/flibbertyJibbet');
+    // Open was called with async `false`.
+    expect(mockXHR.async).to.be.false;
```
benlesh, line 564: Here's the most important part of this test. This is only set when the XHR's `open` method is called, and it defaults to `true`.

spec/observables/dom/ajax-spec.ts
```diff
@@ -728,7 +716,6 @@ describe('ajax', () => {
     });
 
     it('should succeed on 204 No Content', () => {
-      const expected: null = null;
```
benlesh, line 731: `null` didn't make any sense here. a real `XHR` would never have `null` in the `responseText`, or the `response`.

spec/observables/dom/ajax-spec.ts
```diff
@@ -748,11 +735,11 @@ describe('ajax', () => {
 
       request.respondWith({
         status: 204,
-        contentType: 'application/json',
-        responseText: expected,
+        responseType: '',
+        responseText: '',
```
benlesh, line 739: this is more akin to what you'd see in a real call like this.

spec/observables/dom/ajax-spec.ts
```diff
@@ -823,11 +810,10 @@ describe('ajax', () => {
     });
 
     it('should succeed on 204 No Content', () => {
-      const expected: null = null;
       let result: AjaxResponse<any>;
       let complete = false;
 
-      ajax.post('/flibbertyJibbet', expected).subscribe(
```
benlesh, line 830: no idea why I chose to post a body of `null` here, seems weird. But this whole test was really just to check that 204 status passed okay.

spec/observables/dom/ajax-spec.ts
```diff
@@ -872,13 +857,13 @@ describe('ajax', () => {
       request.respondWith(
         {
           status: 200,
-          contentType: 'application/json',
+          responseType: 'json',
           responseText: JSON.stringify({}),
         },
-        3
+        { uploadProgressTimes: 3 }
```
benlesh, line 863: Just a change to configuration for these.

spec/observables/dom/ajax-spec.ts
```diff
@@ -1127,7 +1114,13 @@ describe('ajax', () => {
 });
 
 class MockXMLHttpRequest {
-  public static readonly DONE = 4;
+  static readonly DONE = 4;
+
+  /**
+   * Set to `true` to test IE code paths.
+   */
+  static noResponseProp = false;
```
benlesh, line 1122: This is that new property that is used to simulate older browsers that would lack a `response` property on their XHR.

spec/observables/dom/ajax-spec.ts
```diff
@@ -1140,19 +1133,23 @@ class MockXMLHttpRequest {
   }
 
   static clearRequest(): void {
+    MockXMLHttpRequest.noResponseProp = false;
```
benlesh, line 1136: Making sure we reset it here.

spec/observables/dom/ajax-spec.ts
```diff
@@ -1140,19 +1133,23 @@ class MockXMLHttpRequest {
...
   }
 
   protected responseType: string = '';
   private readyState: number = 0;
 
-  private async: boolean = true;
+  /**
+   * Used to test if `open` was called with `async` true or false.
+   */
+  public async: boolean = true;
```
benlesh, line 1147: Made this public so we can test it. (self nit: I should have removed the `public` and just let it be public, but we should probably use a lint rule for that)

spec/observables/dom/ajax-spec.ts
```diff
@@ -1173,6 +1170,9 @@ class MockXMLHttpRequest {
   constructor() {
     MockXMLHttpRequest.recentRequest = this;
     MockXMLHttpRequest.requests.push(this);
+    if (MockXMLHttpRequest.noResponseProp) {
+      delete this['response'];
+    }
```
benlesh, line 1175: Above here is _part_ of the magic for removing the `response` property from the XHR to simulate older browsers.

spec/observables/dom/ajax-spec.ts
```diff
@@ -1214,45 +1214,70 @@ class MockXMLHttpRequest {
...
-      this.response = this.responseText;
+  respondWith(
+    response: {
+      status?: number;
+      responseText?: string | undefined;
+      responseType: XMLHttpRequestResponseType;
+      total?: number;
+      loaded?: number;
+    },
+    config?: { uploadProgressTimes?: number; downloadProgressTimes?: number }
```
benlesh, line 1225: New configuration options for triggering the upload and download progress events (including `loadstart` for each)

spec/observables/dom/ajax-spec.ts
```diff
@@ -1214,45 +1214,70 @@ class MockXMLHttpRequest {
...
     }
+
+    // Set the readyState to 4.
     this.readyState = 4;
+
+    // Default to OK 200.
     this.status = response.status || 200;
     this.responseText = response.responseText;
-    const responseType = response.responseType !== undefined ? response.responseType : this.responseType;
-    if (!('response' in response)) {
```
benlesh, line 1241: I could have left this I suppose, since I'm removing the property, but I think this `in` check was kind of confusing, so I opted to just let this `switch` process. And remove the `response` property again after.

spec/observables/dom/ajax-spec.ts
```diff
@@ -1214,45 +1214,70 @@ class MockXMLHttpRequest {
...
-      switch (responseType) {
-        case 'json':
-          this.jsonResponseValue(response);
```
benlesh, line 1244: this got inlined.

spec/observables/dom/ajax-spec.ts
```diff
@@ -1214,45 +1214,70 @@ class MockXMLHttpRequest {
...
-    // TODO: pass better event to onload.
-    this.triggerEvent('load');
-    this.triggerEvent('readystatechange');
+
+    // We're testing old IE, forget all of that response property stuff.
+    if (MockXMLHttpRequest.noResponseProp) {
+      delete this['response'];
+    }
+
+    this.triggerEvent('load', { type: 'load', total: response.total ?? 0, loaded: response.loaded ?? 0 });
```
benlesh, line 1279: Added a duck-typed load event here.

cartant (approved):

PR #6003, opened 2021-02-08
fix: publish variants returning ConnectableObservable not properly utilizing lift

benlesh:
- Adds tests to show fixed issues
- Adds tests to show issues that simply can't be fixed, and support a case against removing operators that return ConnectableObservable, as well as possibly a case against lifting Subject.
- Moves logic that patched lift for ConnectableObservable to the constructor so it is used by all multicast operators

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -2,7 +2,7 @@ import { expect } from 'chai';
 import * as sinon from 'sinon';
 import { Observer, TeardownLogic } from '../src/internal/types';
 import { Observable, config, Subscription, noop, Subscriber, Operator, NEVER, Subject, of, throwError, empty } from 'rxjs';
-import { map, multicast, refCount, filter, count, tap, combineLatest, concat, merge, race, zip, catchError, concatMap, switchMap } from 'rxjs/operators';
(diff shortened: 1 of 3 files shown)


spec/Observable-spec.ts
```diff
@@ -810,6 +810,246 @@ describe('Observable.lift', () => {
...
+    result.next(20);
+    result.next(30);
+    
+    expect(emitted).to.deep.equal([100, 200, 300]);
+  });
+
+  /**
+   * Seriously, never do this. It's probably bad that we've allowed this. Fortunately, it's not
+   * a common practice, so maybe we can remove it?
+   */
```
cartant, line 924: OMG, don't give those naughty devs more ideas! 😬 🙈

spec/Observable-spec.ts
```diff
@@ -810,6 +810,246 @@ describe('Observable.lift', () => {
...
+    expect(emitted1).to.deep.equal([100, 200, 300]);
+    expect(emitted2).to.deep.equal([0, 10, 20]);
+  });
+
+  /**
+   * This section outlines one of the reasons that we need to get rid of operators that return
+   * Connectable observable. Likewise it also reveals a slight design flaw in `lift`. It
+   * probably should have never tried to compose through the Subject's observer methods.
+   * If you're a user and you're reading this... NEVER try to use this feature, it's likely
+   * to go away at some point.
```
cartant, line 961: 🙈

cartant (approved):

PR #6010, opened 2021-02-09
fix: concat/merge operators will finalize inners before moving to the next

### LINKED ISSUE #3338: concatAll() unexpectedly overlaps inner Observable subscriptions
EDIT: Note that the following repro code always produces the expected output with RxJS 4.x.  The output with RxJS 5.x is unintuitive (IMO) and can actually vary depending on the formulation of the inner observables (see further discussion of such cases in subsequent comments).

---

**RxJS version:** 5.5.0 (and all previous 5.x versions, seemingly)

**Code to reproduce:**

```javascript
const { Observable } = require(`rxjs`);

function Task(name) {
  return (
    Observable
      .defer(() => {
        console.log(`begin task ${ name }`);
        return (
          Observable
            .of(`simulated work`)
            .delay(100)
            .finally(() => {
              console.log(`end task ${ name }`);
            })
          );
      })
  );
}

Observable
  .of(
    Task(`1`),
    Task(`2`),
    Task(`3`)
  )
  .concatAll()
  .subscribe();
```

**Expected behavior:**
```
begin task 1
end task 1
begin task 2
end task 2
begin task 3
end task 3
```

**Actual behavior:**
```
begin task 1
begin task 2
end task 1
begin task 3
end task 2
end task 3
```

**Additional information:**
I discovered this behavior when using ```concatAll()``` as the basis for a serialized "task queue" for executing tasks (represented as ```Observables```) one at a time, much like my (simplified) sample code above.  Some of these tasks would allocate resources in ```defer()``` (or similar manner) and release them in ```finally()``` (or similar manner).

My task queue implementation all seemed to make logical sense, yet I was running into errors caused by concurrent access to resources that were only supposed to be accessed in a mutually-exclusive manner, but somehow weren't.  I was stumped, I couldn't see the the problem in my task queue or my tasks that was leading to this unexpected concurrent access.

After much head-scratching and debugging, I eventually tracked the problem to the ```InnerSubscriber``` implementation within rxjs, which contains logic that subscribes to the next sequence *before* unsubscribing from the previous sequence:

```javascript
    InnerSubscriber.prototype._error = function (error) {
        this.parent.notifyError(error, this); // <== subscribes to the next sequence
        this.unsubscribe();
    };
    InnerSubscriber.prototype._complete = function () {
        this.parent.notifyComplete(this); // <== subscribes to the next sequence
        this.unsubscribe();
    };
```
Ah-ha.

This causes the "creation" logic (```defer()``` callback or other invocation) of the next sequence to execute before the "disposal" logic (```finally()``` or other invocation) of the previous sequence has executed, making it impossible for the inner observables managed by ```concatAll()``` to behave like serialized tasks -- since rxjs actually "overlaps" the tasks at this critical moment.

As an experiment, I simply reversed the logic:
```javascript
    InnerSubscriber.prototype._error = function (error) {
        this.unsubscribe(); // <==  unsubscribe from previous sequence first!
        this.parent.notifyError(error, this);
    };
    InnerSubscriber.prototype._complete = function () {
        this.unsubscribe(); // <== unsubscribe from previous sequence first!
        this.parent.notifyComplete(this);
    };
```
Then, using the original "Code to reproduce" above, I actually see the expected behavior.

I don't know, however, what other ramifications this change might have on other use cases.  In the case of ```concatAll()``` (and variants) however, it would definitely seem "more correct" and "less surprising" for rxjs to always unsubscribe from the previous sequence before subscribing to the next.

benlesh:
Resolves an issue where inner observables would not finalize before the next inner observable got subscribed to. This happened in concat variants, and merge variants with concurrency limits, and could be surprising behavior to some users.

fixes #3338

### DIFF
--- a/spec/operators/concatAll-spec.ts
+++ b/spec/operators/concatAll-spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
-import { from, throwError, of, Observable } from 'rxjs';
-import { concatAll, take, mergeMap } from 'rxjs/operators';
+import { from, throwError, of, Observable, defer } from 'rxjs';
+import { concatAll, take, mergeMap, finalize, delay } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
 
@@ -57,6 +57,40 @@ describe('concatAll operator', () => {
       }
     );
   });
+
+    it('should finalize before moving to the next observable', () => {
+      const results: any[] = [];
+
+      const create = (n: number) => defer(() => {
+        results.push(`init ${n}`);
+        return of(`next ${n}`).pipe(
+          delay(100, testScheduler),
+          finalize(() => {
+            results.push(`finalized ${n}`)
+          })
+        );
+      });
+
+      of(create(1), create(2), create(3)).pipe(
+        concatAll()
+      ).subscribe({
+        next: value => results.push(value),
+      });
+
+      testScheduler.flush();
+
+      expect(results).to.deep.equal([
+        'init 1',
+        'next 1',
+        'finalized 1',
+        'init 2',
+        'next 2',
+        'finalized 2',
+        'init 3',
+        'next 3',
+        'finalized 3'
+      ]);
+    });
 
   it('should concat and raise error from promise', function(done) {
     this.timeout(2000);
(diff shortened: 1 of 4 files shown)


benlesh:
This PR is also an example of why I don't really want to release `OperatorSubscriber` yet. I had to change a feature of it to make this work properly.

cartant:
> This PR is also an example of why I don't really want to release OperatorSubscriber yet. I had to change a feature of it to make this work properly.
It's a beta, so it can be changed, IMO.

cartant (approved):

PR #6014, opened 2021-02-10
fix(distinctUntilChanged): Ensure reentrant code is compared properly

benlesh:
- Fixes an issue where reentrant code would not be compared properly against previous emissions
- Adds better code comments
- Updates documentation with better text and examples

Related to #6013

### DIFF
--- a/src/internal/operators/distinctUntilChanged.ts
+++ b/src/internal/operators/distinctUntilChanged.ts
@@ -1,78 +1,181 @@
 import { MonoTypeOperatorFunction } from '../types';
+import { identity } from '../util/identity';
 import { operate } from '../util/lift';
 import { OperatorSubscriber } from './OperatorSubscriber';
 
-/* tslint:disable:max-line-length */
-export function distinctUntilChanged<T>(compare?: (x: T, y: T) => boolean): MonoTypeOperatorFunction<T>;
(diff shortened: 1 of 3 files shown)


src/internal/operators/distinctUntilChanged.ts
```diff
@@ -1,78 +1,181 @@
...
+          // need to do this if it's the first value, or if the
+          // key we're tracking in previous needs to change.
+          first = false;
+          previousKey = currentKey;
+
+          // Emit the value!
           subscriber.next(value);
         }
-        prev = key;
-        first = false;
```
benlesh, line 75: These can't happen _after_ we next, it has to happen before. Otherwise the added test will fail.

src/internal/operators/distinctUntilChanged.ts
```diff
@@ -1,78 +1,181 @@
...
+): MonoTypeOperatorFunction<T>;
+
+export function distinctUntilChanged<T, K>(
+  comparator?: (previous: K, current: K) => boolean,
+  keySelector: (value: T) => K = identity as (value: T) => K
+): MonoTypeOperatorFunction<T> {
+  // We've been allowing `null` do be passed as the `compare`, so we can't do
+  // a default value for the parameter, because that will only work
+  // for `undefined`.
+  comparator = comparator ?? defaultCompare;
```
benlesh, line 150: Added this comment, because I tried to use a default value parameter here and it was sad. We should probably deprecate this sort of thing at some point, and make this a configuration object, I suppose.

cartant (approved):

PR #6020, opened 2021-02-11
refactor(forkJoin): clean up logic - unify approach with combineLatest

benlesh:
- cleans up unnecessary helper function
- creates `createObject` helper function to use in both
- uses `mapOneOrManyArgs` helper.
- switches to decrement instead of increment for internal counters for lighter weight conditionals

### DIFF
--- a/src/internal/observable/combineLatest.ts
+++ b/src/internal/observable/combineLatest.ts
@@ -7,6 +7,7 @@ import { identity } from '../util/identity';
 import { Subscription } from '../Subscription';
 import { mapOneOrManyArgs } from '../util/mapOneOrManyArgs';
 import { popResultSelector, popScheduler } from '../util/args';
+import { createObject } from '../util/createObject';
 
 // combineLatest([a, b, c])
 export function combineLatest(sources: []): Observable<never>;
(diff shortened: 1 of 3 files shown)


cartant (approved):

src/internal/observable/combineLatest.ts
```diff
@@ -475,24 +476,13 @@ export function combineLatest<O extends ObservableInput<any>, R>(...args: any[])
...
       keys
         ? // A handler for scrubbing the array of args into a dictionary.
-          (values: any[]) => {
-            const value: any = {};
-            for (let i = 0; i < values.length; i++) {
-              value[keys![i]] = values[i];
-            }
-            return value;
-          }
+          (values) => createObject(keys, values)
```
benlesh, line 479: We didn't need to be doing this two different ways in the same library.

src/internal/observable/forkJoin.ts
```diff
@@ -124,51 +126,40 @@ export function forkJoin<T>(sourcesObject: T): Observable<{ [K in keyof T]: Obse
...
       return;
     }
-    const values = new Array(len);
-    let completed = 0;
-    let emitted = 0;
-    for (let sourceIndex = 0; sourceIndex < len; sourceIndex++) {
-      const source = innerFrom(sources[sourceIndex]);
+    const values = new Array(length);
+    let remainingCompletions = length;
+    let remainingEmissions = length;
```
benlesh, line 138: Changed these to a "countdown" instead of a counter. It makes the checks below a little more terse, and I think it's easier to reason about if you were to step into the code "Oh, there's 1 left" instead of having to do (very simple) math in your head to figure out how many more need to complete.

PR #6021, opened 2021-02-12
refactor(combineLatest): small efficiency changes

benlesh:
- no longer use an array to track first value emission
- inline a few things than can be inlined
- adds more comments

### DIFF
--- a/src/internal/observable/combineLatest.ts
+++ b/src/internal/observable/combineLatest.ts
@@ -8,6 +8,7 @@ import { Subscription } from '../Subscription';
 import { mapOneOrManyArgs } from '../util/mapOneOrManyArgs';
 import { popResultSelector, popScheduler } from '../util/args';
 import { createObject } from '../util/createObject';
+import { OperatorSubscriber } from '../operators/OperatorSubscriber';
 
 // combineLatest([a, b, c])
 export function combineLatest(sources: []): Observable<never>;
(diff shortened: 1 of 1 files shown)


src/internal/observable/combineLatest.ts
```diff
@@ -495,25 +496,6 @@ export function combineLatest<O extends ObservableInput<any>, R>(...args: any[])
   return result;
 }
 
-/**
- * Because of the current architecture, we need to use a subclassed Subscriber in order to ensure
- * inner firehose observables can teardown in the event of a `take` or the like.
- * @internal
- */
-class CombineLatestSubscriber<T> extends Subscriber<T> {
```
benlesh, line 503: We don't need this, we can use `OperatorSubscriber`.

src/internal/observable/combineLatest.ts
```diff
@@ -522,44 +504,61 @@ export function combineLatestInit(
...
+        const { length } = observables;
+        // A store for the values each observable has emitted so far. We match observable to value on index.
+        const values = new Array(length);
+        // The number of currently active subscriptions, as they complete, we decrement this number to see if
+        // we are all done combining values, so we can complete the result.
+        let active = length;
+        // The number of inner sources that still haven't emitted the first value
+        // We need to track this because all sources need to emit one value in order
+        // to start emitting values.
+        let remainingFirstValues = length;
```
benlesh, line 519: Here I'm using a countdown, instead of an array, to track whether or not each source has emitted at least one value.

cartant (approved):

PR #6029, opened 2021-02-16
fix: add test from #5932 and plug leak in share

cartant:
**Description:**

This PR brings in the test from #5932 and fixes a leak that it exposed in the `share` implementation.

Maybe we should add the test to the specs for `share`, too, but this is all I have time for, ATM.

**Related PR:** #5932

### DIFF
--- a/src/internal/operators/share.ts
+++ b/src/internal/operators/share.ts
@@ -1,6 +1,6 @@
 import { Subject } from '../Subject';
 
-import { MonoTypeOperatorFunction, OperatorFunction, SubjectLike } from '../types';
+import { MonoTypeOperatorFunction, OperatorFunction, SubjectLike, Unsubscribable } from '../types';
 import { Subscription } from '../Subscription';
 import { from } from '../observable/from';
 import { operate } from '../util/lift';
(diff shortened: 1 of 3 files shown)


src/internal/operators/share.ts
```diff
@@ -132,7 +132,8 @@ export function share<T>(options?: ShareConfig<T>): OperatorFunction<T, T> {
 
     return () => {
       refCount--;
-      castSubscription.unsubscribe();
+      castSubscription!.unsubscribe();
+      castSubscription = null;
```
benlesh, line 136: If I uncomment this line, the added test still passes. Not sure what I'm missing.

benlesh:
@cartant I've added another commit that more appropriately fixes the issue.

The problem was we were accidentally returning a teardown that was basically adding a closure over a subscriber to itself.

Effectively:

```ts
  return () => {
    subscriber.unsubscribe();
  };
});
```

It's not super obvious when you look at it, but that's what it's doing.  `const castSubscription = subject.subscribe(subscriber);` was the confusing bit I think... because technically, `castSubscription === subscriber`. :)

Calling `subscribe` with our own `Subscriber` will return the same instance of `Subscriber`.

cartant:
@benlesh LGTM. I'd approve it, but I can't approve my own PR. 🙂

benlesh (approved):

PR #6038, opened 2021-02-21
chore: update materialize tests to run mode

jakovljevic-mladen:
**Description:**
Updated `materialize` tests to use run mode.

I have removed one of the tests (named `should materialize stream never completes`) because it was completely the same as the test before, it only used `cold` instead of `hot`, but emissions and the subscriptions points were the same.

None

### DIFF
--- a/spec/operators/materialize-spec.ts
+++ b/spec/operators/materialize-spec.ts
@@ -1,145 +1,164 @@
+/** @prettier */
 import { expect } from 'chai';
 import { materialize, map, mergeMap, take } from 'rxjs/operators';
 import { Notification, of, Observable } from 'rxjs';
-import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
+import { TestScheduler } from 'rxjs/testing';
+import { observableMatcher } from '../helpers/observableMatcher';
 
 /** @test {materialize} */
-describe('materialize operator', () => {
+describe('materialize', () => {
+  let testScheduler: TestScheduler;
+
+  beforeEach(() => {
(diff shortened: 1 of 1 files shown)


jakovljevic-mladen:
The error from the pipelines does not look like it was related to this PR.

At first, I wasn't able to reproduce it locally, but then I saw a change from #6033 where this line was introduced: `npm install --no-save typescript@latest tslib@latest @types/node@latest`. After I run it and after I run `npm run compile`, I managed to reproduce the same results locally, but then my IDE started complaining that I should run `npm install` again since `tslib` versions do not match.
image
@cartant, maybe you could help here? Is it possible to ignore this error for this PR or is there something that should be done regarding `tslib` version?

cartant:
I'll have a look at it.

> ... but then my IDE started complaining

You won't want to run exactly what happens in CI locally. As you've noticed, there is a CI task that forcibly installs the latest TS - so that we get notice of any breakage (like the one you've encountered) - and that also means that it needs to install the latest `tslib` and Node.js types. Forcibly installing the latest in the CI task is no big deal, as everything is thrown away afterwards. If that task is run locally, you'll end up with installed modules that don't match the `package.json` semver.

jakovljevic-mladen:
> If that task is run locally, you'll end up with installed modules that don't match the package.json semver.

Yeah, I had to reinstall everything from the ground after installing the latest stuff, but by doing so, I just wanted to make sure I can reproduce an error locally.

>I'll have a look at it.

Thanks 👍

cartant (approved):
LGTM. The CI issues should be fixed in #6039.

PR #6040, opened 2021-02-22
chore: update max tests to run mode

jakovljevic-mladen:
**Description:**
Updated `max` tests to use run mode.

_Please note_: in test `should max the values of an ongoing hot observable`, I changed one of the values to be less then the value emitted by the hot observable _before_ the subscription (changed from `666` to just `6` as it is less than `42`), so it is clear that the value emitted before subscription is not taken into consideration.

None


cartant (approved):
LGTM. Thanks.

Issue #6030, opened 2021-02-17
Ensure we're properly supporting ESM in Node


benlesh:
See the write up by @Rich-Harris here: 

https://github.com/d3/d3/issues/3469

benlesh:
(cries internally)

tomastrajan:
It would be great (and compliant) if rxjs shipped

`dist/esm/*.mjs`

instead if 

`dist/esm/*.js` 

(ie use `.mjs` extension, but only in esm folders)

The reson for this is that some tooling assumes (correctly) that `.js` means it's `.cjs` (in the absence of `"type": "module"` in the nearest `package.json`) and breaks and there is no way to override this behavior?

dynst:
Dual-packaging properly is basically impossible just using `tsc` as the transpiler, because it always outputs `.js` extensions no matter what. The new `observable` package uses `tshy`, so maybe it's worth switching to using that.

Because it looks like the problem this old PR was trying to fix still exists (all ESM files need to use the `.mjs` extension, but they currently don't): https://github.com/ReactiveX/rxjs/pull/7130

https://nodejs.org/api/packages.html#determining-module-system

The reason no one's noticed is the ESM files aren't actually exported in Node: #7418

dynst:
9.0 is going to be ESM-only and gives up dual-packaging entirely. For 9.0, can `tshy` just be removed and replaced with plain `tsc` then?

But it looks like the [`7.x`](https://github.com/ReactiveX/rxjs/commits/7.x/) branch hasn't changed and remains totally broken on this front.

Issue #5486, opened 2020-06-12
Support rxjs for deno


mtwzim:
## Feature Request

Hello guys, recently I am working on a version of rxjs for [deno](https://deno.land). Well the version I modified is already working, but when talking to Ben, he guided me to create an issue so we can fix this,

**Describe the solution you'd like**
When talking to ben, we can have a code for the node and the node, is that possible? My version of rxjs for deno is here: https://github.com/DenoBRComunitty/rxjs

What can we do about it?

Sorry for my bad issue.
Bye

benlesh:
Well, I think first we need to understand what all does not work in Deno. Do you happen to have a comprehensive list of what isn't compatible?

mtwzim:
In fact, it is very easy to migrate to deno (in the case of RxJS, you can see by the small amount of commits that I had to make in [my repository](https://github.com/DenoBRComunitty/rxjs)). Well nothing at NodeJS api https://nodejs.org/api/ works on Deno, this includes the file system ("fs"). https://doc.deno.land/https/github.com/denoland/deno/releases/latest/download/lib.deno.d.ts this is an API that should be used.
Changing the RxJS itself is very easy, in 30 minutes this is done, the real problem is the tests on node and builds, this is not yet available on Deno and has to be done on the node, although it is possible to use tools like CircleCI and TravisCI to do the tests, it works and I already tried.
In addition, deno uses another means of import, the import / export of files must have the extension explicit, example:
`export { Observable } from './internal/Observable';`
in Deno is:
`export { Observable } from './internal/Observable.ts';`
I think that's just it.

kwonoj:
As described above, biggest blocker is import path interop between two different runtime : esp to support node.js esm (which also requires proper extension) maintain compatibility / non breaking change it is something we need carefully evalute. This is primary reason I'm going to focus esm support first withou breaking existing published pkg (this is whole another tricky story) then consider new runtime's path support.

mtwzim:
[...]

Issue #5350, opened 2020-03-10
Deno support


FrancescoBorzi:
This has already been asked here: https://github.com/ReactiveX/rxjs/issues/4937

and the answer from @kwonoj was:

> We will eventually consider deno as support runtime, but will not have actual implementation until we see stable version of deno released.

Now, [it has been announced](https://github.com/denoland/deno/issues/2473#issue-453631188) that Deno 1.0 will be released the 13rd of May 2020.

So maybe it's time to reconsider this?

kwonoj:
Answer's pretty same as before: consideration does not mean we'll release support when deno have its stable release. Please note we even can't support native esm on node.js due to several blocker's we are seeing at this moment - and supporting deno's module resolution won't happen in day 1. 

It's on our radar but there's no specific date we can provide at this moment.

benlesh:
I think we should discuss these issues at the next meeting. Deno is a run time with growing usage, and I think we should support it.

kwonoj:
@benlesh we had discussed in prev core meeting (when you couldn't attend). Tl:dr story is import site support to conform both (node / deno) is not trivial, and supporting node's esm is higher priority at this point but even supporting esm is not viable in short timeframe. I agree we will / may support deno one day but not something we'll actively plan for now.

benlesh:
Well, given that the popularity of Deno has risen quite a bit, maybe we should discuss again what it entails at the next meeting.

kwonoj:
We have dupe (https://github.com/ReactiveX/rxjs/discussions/5971) and since both requests are exactly same, don't think we need open both. We _will_ evaluate this eventually, but other priorities like esm is much higher and we couldn't solve that problem yet even.

PR #6043, opened 2021-02-22
refactor: node native esm 

calebboyd:
**Description:**

This PR changes the package to a `"type": "module"` package. Adding an `exports` entry to the package.json which going forward will do the work of the generate-alias helper

- Appends .js to each import
- Fixes up some double quotes.
- Moves .make-helpers to a tool/make-helpers
- Keeps tools as commonjs
- Keeps dist/spec dist/cjs and dist/src as commonjs (aligning with typescript emit)

#6030 #4416 #2858 #3858

General beginning for deno support via compatibility layer #5350 #5486

## Breaking Changes

- When a user is on node > 12.7 They will no longer be able to use deep import paths, and can only use what is specified in `package.json#exports`.
  - Eg. if a user is accessing `require('rxjs/dist/cjs')` Node will throw
a `ERR_PACKAGE_PATH_NOT_EXPORTED` error.

- ~A user on early EOLd node 13 using `--experimental-modules` will be importing a commonjs module as esm since conditional entries are not supported.~ node 13 compat dropped in #6192


kwonoj:
Thanks for tackling this - we have few internal attempt and failed to make things work to satisfy for some conditions.  

Please ensure this should pass 

- source code itself, intellisense and others should able to navigate to proper source code 
- existing cjs / esm export should work (since we can't make breaking changes for v7). If we were to deprecate existing esm2015, this should be major breaking and goes for v8 or beyond
- import test suite (https://github.com/ReactiveX/rxjs/tree/master/integration/import) is very important to ensure it's not breaking. Currently it's cjs only, but with esm it should have corresponding `esm` suite and both should able to pass test. if you open up existing cjs suite, it checks export via `require` - same goes for `esm`, it should check export via `esm` and both should pass
- unit test (mocha) should able to run cjs / esm as separate suite to validate changes.

kwonoj:
I'm marking as blocked for now since this is in-progress.

kwonoj:
Personally, I'm very hesitant to postfix `.js` extension to all of our source codes, but if that makes all things work may be feasible. Still prefer there's some postprocess to handle those instead of all src keeps it manually. /cc @cartant @benlesh

package.json
```diff
@@ -36,7 +63,7 @@
     "test:systemjs": "node integration/systemjs/systemjs-compatibility-spec.js",
     "test:side-effects": "check-side-effects --test integration/side-effects/side-effects.json",
     "test:side-effects:update": "npm run test:side-effects -- --update",
-    "test:import": "ts-node ./integration/import/runner.ts",
+    "test:import": "node --loader ts-node/esm ./integration/import/runner.ts",
```
kwonoj, line 66: this is incorrect: test:import is SHOULD NOT use any interop but always use native node.js resolution to test proper import resolution. if we use third party, there's no meaning to run this test.

kwonoj (changes requested):

ncjamieson:
> Personally, I'm very hesitant to postfix .js extension to all of our source codes, but if that makes all things work may be feasible.

FWIW, my understanding is that this is the correct and recommended approach as far as TypeScript is concerned.

kwonoj:
> FWIW, my understanding is that this is the correct and recommended approach as far as TypeScript is concerned.

Back in the days I recall there's some problem around intellisense or other ts service things, is it now working?

tsconfig.json
```diff
@@ -7,6 +7,8 @@
     "strict": true,
     "noImplicitReturns": true,
     "moduleResolution": "node",
+    "allowSyntheticDefaultImports": true,
```
kwonoj, line 10: this should not be included: we intentionally avoid to allow synthetic default import

kwonoj (changes requested):

ncjamieson:
> ...  around intellisense or other ts service things

TBH, I'd be hesitant to say that the entire TS ecosystem will be working with it, despite it being the recommended approach. I'll have a look at my notes, this evening, for some links to 'official' TS statements. At work, ATM, hence the non-cartant GitHub account.

kwonoj:
> TBH, I'd be hesitant to say that the entire TS ecosystem will be working with it, despite it being the recommended approach. I'll have a look at my notes, this evening, for some links to 'official' TS statements. 

I peeked few ts issues back in time and afaik official recommendation is bit behind of supporting things. (we have to double confirm but) if recommended change requires some churns around navigating code, or intellisense or vice versa I'd say it's not something we can easily adapt.

src/internal/operators/mergeMap.ts
```diff
@@ -1,9 +1,9 @@
-import { ObservableInput, OperatorFunction, ObservedValueOf } from '../types';
-import { map } from './map';
-import { innerFrom } from '../observable/from';
-import { operate } from '../util/lift';
-import { mergeInternals } from './mergeInternals';
-import { isFunction } from '../util/isFunction';
+import { ObservableInput, OperatorFunction, ObservedValueOf } from '../types.js';
```
kwonoj, line 1: for type only imports it's still need to ref to .js while it's empty? (https://unpkg.com/browse/rxjs@7.0.0-beta.9/dist/cjs/internal/types.js)

tsconfig.json
calebboyd: This was added to support compile time typings (no emit) - only for `import pkg from 'shelljs'` which doesn't support. If I convert the integration tests setup to commonjs then I can likely use ts-node again.

package.json
calebboyd: This was an effort to replace the existing interop (ts-node bin).
Since the package is now defined as a module the native nodejs resolution won't work with ts-node.
I think I can address this by treating the integration folder as commonjs and then both of your concerns would be addressed

kwonoj:
> Emits package "type": "commonjs" to cjs dist folder (might need to do this in all non esm distros too)

mind elaborate this? we have single package.json for published pkg (i.e https://unpkg.com/browse/rxjs@7.0.0-beta.9/package.json) and don't have per-dist specific.

tsconfig.json
kwonoj: you still can import pkgs without this option. This is global option affects all srcs, so we'd like to avoid it.

kwonoj (changes requested):
I would like to recommend to split commits between other changes to appending `.js` extensions, otherwise reviewers are not possible to go through changes easily.

calebboyd:
> mind elaborate this?

The top level package is `type: module` If the runtime is supporting that it will follow the exports map accordingly for esm. 

Otherwise it will revert to exports[*].require, (or main and aliases folder structure if sufficiently old).
To invoke the cjs loader within a esm package, we have to include a package.json#type=commonjs.

https://github.com/ReactiveX/rxjs/pull/6043/files#diff-f9b4e2ee34781b16f6c4ae4801b021ad200776de33c06d9a1f96a2443ff2c471R32

kwonoj:
> Otherwise it will revert to exports[*].require, (or main and aliases folder structure if sufficiently old).
To invoke the cjs loader within a esm package, we have to include a package.json#type=commonjs.

oh, I mean node.js docs / spec reference I can read about.

tsconfig.json
calebboyd: Makes sense, However I couldn't avoid it when importing commonjs into esm

kwonoj: yes, agreed. that's what I encountered several time https://twitter.com/_ojkwon/status/1362187177975631873. Since we should not break any existing ergonomics, unlike greenfield we have lot of constraints to keep it. tsconfig is something falls into those category, especially we had some history with default import / export and explicitly banned out options around default import so we want to keep it while making changes.

calebboyd:
> oh, I mean node.js docs / spec reference I can read about.

This is a good place to start https://nodejs.org/dist/latest-v15.x/docs/api/packages.html#packages_modules_packages

kwonoj:
> This is a good place to start https://nodejs.org/dist/latest-v15.x/docs/api/packages.html#packages_modules_packages

I am aware of those docs. I think I missed part you specifically mentioned about cjs interop and would like to read those part.

calebboyd:
Gotcha, Maybe this: (I'm using conditional exports to support dual mode) https://nodejs.org/dist/latest-v15.x/docs/api/packages.html#packages_dual_commonjs_es_module_packages

tsconfig.json
calebboyd: Changing this specific option doesn't have any effect on code emit - its only effect is the type checking that occurs within this project.

kwonoj: Theoritically yes, but root option compiles down things to src shouldn't change due to other places. I personally against to emable this option in this repo, but if we can't avoid it'd better to separate config where it's being used.

cartant:
Regarding the `.js` extension business, [this is the comment](https://github.com/microsoft/TypeScript/issues/16577#issuecomment-309169829) which I have linked in my ES module-related notes, but there is a [more-recent comment](https://github.com/microsoft/TypeScript/issues/16577#issuecomment-754941937) that was made before that issue was closed.

calebboyd:
>I would like to recommend to split commits between other changes to appending .js extensions, otherwise reviewers are not possible to go through changes easily.

Split commits added! This is ready for review.
If there are any other tests or things you'd like to have added or clarified please let me know.

integration/import/package.json
```diff
@@ -0,0 +1,11 @@
+{
```
kwonoj, line 1: where's package.json for esm?

calebboyd: I'm forcing esm with the `.mjs` extension in these tests.

kwonoj: I think I'm bit confused. Purpose of this integration test is trying to mimc actual user's package.json with various config, so we can't enforce something since we can't ask user to adjust package.json. That's reason there was `cjs/package.json`, and I expect esm have same package.json have `type:module` to test esm import when user decides to go with native esm. So in total, we expect at least 3 tests - `cjs` without any `type` (for node.js doesn't support `type` property) / `cjs` with `type:cjs` / esm with `type:module`. 

User can choose either use mjs, or decide `type:module` and treat all js as esm.

kwonoj: Mind elaborate how current test covers above cases?

calebboyd: My thought was that by invoking cjs and mjs explicitly with file extensions, I'm exacting the codepaths in node that load the code with the correct loader. Whether that loader option is invoked via a package.json or a type field is a node implementation detail that may expand in the future. IMHO adding a separate package.json is just another way node gives you to chose a loader but I don't think it needs to be a focus of these tests.

Right now the assertions being made ensure asserting that the module exports are equivalent when run via cjs and esm.

kwonoj: > IMHO adding a separate package.json is just another way node gives you to chose a loader but I don't think it needs to be a focus of these tests.

as long as your test can represent real user's package.json usecase to test, it should be ok. But I'm bit hard to understand if it's covered. So current way can verify flows like

1. user uses old / latest node.js, package.json doesn't have type field and want to use cjs
2. user uses latest node.js, package.json have type field to cjs, want to use cjs
3. user uses latest node.js package.json have type field to esm, want to use esm
4. (need to be added) webpack / other bundle picks up esm correctly

> Whether that loader option is invoked via a package.json or a type field is a node implementation detail that may expand in the future

This isn't about node.js internal, As stated this test try to mimic current real world package.json structures as user load their module in various different way. Regardless of how node.js treats, we should support existing ergonomics user uses.

calebboyd: For sure, For the flows you outlined - the answers are declared in the package.json

1) `main` field is discovered -> dist/cjs is loaded
2 & 3) `exports` field is discovered -> import ? -> dist/esm : dist/cjs
4) Webpack implements resolution logic based on [these fields](https://github.com/webpack/enhanced-resolve#resolver-options) as well

I don't know if there is value to adding more tests here since the existing tests now validate that both esm and cjs can be requested and contain a consistent API.

integration/import/esm.mjs
```diff
@@ -0,0 +1,16 @@
+import { strict as assert } from 'assert'
+import { readFileSync } from 'fs'
+import { join, dirname } from 'path'
+import { fileURLToPath } from 'url'
+import * as rx from 'rxjs'
+import * as operators from 'rxjs/operators'
```
calebboyd, line 6: This is highlighting the new feature that is added in this PR -- currently these lines of code are not possible in 7.0.0-beta.10

calebboyd:
> Thanks for tackling this - we have few internal attempt and failed to make things work to satisfy for some conditions.
> 
> Please ensure this should pass at least
> 

* [x] ~source code itself, intellisense and others should able to navigate to proper source code~

* [x] ~existing cjs / esm export should work (since we can't make breaking changes for v7). If we were to deprecate existing esm2015, this should be major breaking and goes for v8 or beyond~
* [x] ~import test suite ([https://github.com/ReactiveX/rxjs/tree/master/integration/import](https://github.com/ReactiveX/rxjs/tree/master/integration/import?rgh-link-date=2021-02-22T23%3A06%3A24Z)) is very important to ensure it's not breaking. Currently it's cjs only, but with esm it should have corresponding `esm` suite and both should able to pass test. if you open up existing cjs suite, it checks export via `require` - same goes for `esm`, it should check export via `esm` and both should pass~

* [ ] unit test (mocha) should able to run ~cjs~ / esm as separate suite to validate changes.
	
* [ ] sourcemap should work for cjs / esm both, for node.js and browers (no regressions) 

@kwonoj - currently the specs are compiled to commonjs (dist/spec + dist/src) Should I add an additional dist/spec-esm and dist/src-esm to run mocha using both loader environments?

calebboyd:
If there is anything I can do to answer any questions or address any concerns with this please let me know. I've been keeping this rebased and have added bundler integration tests for rollup, webpack4, and webpack5.

benlesh:
@calebboyd, I'm keeping an eye on this. I'm not sure what to do with it until after we talk about it at the Core Team meeting.

calebboyd:
@benlesh Thanks thats good to know 🙂. I'll continue to rebase it as needed.

Some points that could be discussed from my limited point of view:
- Supported node versions for v7? (dropping unsupported (13) could make this config simpler)
- I made an assumption that module state wasn't a concern - Should a [wrapper approach](https://www.npmjs.com/package/gen-esm-wrapper) be taken to avoid multiple instances of the module (cjs+esm) being loaded into a v8 context.

benlesh:
Core Team Meeting Notes: We've decided we want to do this at some point, but version 7 isn't the right time to do so. We'd like to let this ecosystem simmer for a bit and address this maybe in version 8.

calebboyd:
Closing for now then, thanks for the update

Issue #6045, opened 2021-02-23
debounceTime performance when having large amounts of updates


voliva:
**RxJS version:**
6 and 7

**Code to reproduce:**
https://codesandbox.io/s/practical-poincare-gm0h6?file=/src/index.ts

It presents a stream with 100.000 updates, and two buttons to try two posibilities: use `debounceTime` from `rxjs` or use a custom (not finished) implementation that works with just one `setTimeout`

**Expected behavior:**
RxJS test takes a constant time, just like the custom implementation

**Actual behavior:**
RxJS test takes a linear time based on number of updates

**Additional information:**

There's a bit of discussion around debounce in #2834 and I can see that the implementation is not as straightforward as in the custom one, as there's an scheduler.

However, it should be posible to not cancel the task and re-create it on every message received, but use something similar to the custom implementation so that when the last task evaluates, it checks whether it needs to set up a new task or it can emit the last value straight away.

I think it's important to decrease the overhead of ignored messages, as one of the main points of `debounceTime` is to prevent processing quick succession of events.

I'm happy to contribute on this if you need some help.

cartant:
Closed by #6049

PR #6049, opened 2021-02-24
fix(debounceTime): improves performance on quick succession of emits

voliva:
Addresses #6045 

This turned out to be similar to `debounce`, but with the optimisation on the scheduling bit. Tests unchanged are green, I'm not sure how can I add a test for performance.

This optimization can't be used in the `debounce` operator, as that one takes in a stream, where only way of getting the "time window" is by unsubscribing from the previous stream and subscribing on the current one. With `debounceTime`, as the dueTime is always the same we can optimise.

In case this PR moves forward, would it be worth also having the same optimisation on RxJS@^6' debounceTime?

### DIFF
--- a/src/internal/operators/debounceTime.ts
+++ b/src/internal/operators/debounceTime.ts
@@ -1,7 +1,8 @@
 import { asyncScheduler } from '../scheduler/async';
-import { MonoTypeOperatorFunction, SchedulerLike } from '../types';
-import { debounce } from './debounce';
-import { timer } from '../observable/timer';
+import { Subscription } from '../Subscription';
+import { MonoTypeOperatorFunction, SchedulerAction, SchedulerLike } from '../types';
+import { operate } from '../util/lift';
+import { OperatorSubscriber } from './OperatorSubscriber';
 
 /**
  * Emits a notification from the source Observable only after a particular time span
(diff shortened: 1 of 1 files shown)


josepot:
Thanks a lot for this @voliva !

I would like to explain a situation that I found myself pretty recently that I think that highlights the need for this change:

I was trying to understand why there was a task that was chocking for about ~400ms every time that I was receiving a message from the websocket. At first I thought that it was a problem with the parser of a library, but then I realized that the issue was a piece of code that looked more or less like this:

```ts
const gePricesHistory$ = (symbol: string) =>
  concat(
    getRemoteProcedureCall$<Price[], string>(
      "priceHistory",
      "getPriceHistory",
      symbol,
    ).pipe(mergeAll()),
    getStream$<Price, Request>("pricing", "getPriceUpdates", { symbol }),
  ).pipe(
    scan((acc, price) => {
      const result = acc.concat(price)
      return result.length <= HISTORY_SIZE ? result : result.slice(1)
    }, [] as HistoryPrice[]),
    debounceTime(0),
  )

```

The problem was that in some instances the RPC call that returns the initial history would have thousands of entries, and that meant that the JS thread would "freeze" for ~400ms becaue `debounceTime(0)` was not behaving how I thought it would behave, once I refactored the code to something like this:

```ts
const gePricesHistory$ = (symbol: string) =>
  mergeWithKey(
    {
      init: getRemoteProcedureCall$<Price[], string>(
        "priceHistory",
        "getPriceHistory",
        symbol,
      ).pipe(
        map((history) => history.slice(Math.max(0, x.length - HISTORY_SIZE))),
      ),
      update: getStream$<Price, Request>("pricing", "getPriceUpdates", {
        symbol,
      }),
    },
    1,
  ).pipe(
    scan((acc, event) => {
      if (event.type === "init") return event.payload
      const result = acc.concat(event.payload)
      return result.length <= HISTORY_SIZE ? result : result.slice(1)
    }, [] as HistoryPrice[]),
    debounceTime(0),
  )
```

then the performance issue was instantly gone.

I know that the second version will always perform better than the first one, but I would have never anticipated for a `debouceTime(0)` to have that kind of performance overhead.

benlesh (approved):
Extremely well done!

I have some very minor nits around stylistic and comment type stuff, but nothing even worth mentioning. Very impressed.

PR #6050, opened 2021-02-25
chore: use workaround type instead of any

cartant:
**Description:**

This PR replaces the workaround - for what seems to be a TS 4.2 problem - with an explicit type instead of an `any`.

**Related PR:** #6048

### DIFF
--- a/src/internal/observable/bindCallback.ts
+++ b/src/internal/observable/bindCallback.ts
@@ -2,6 +2,7 @@
 import { SchedulerLike } from '../types';
 import { Observable } from '../Observable';
 import { bindCallbackInternals } from './bindCallbackInternals';
+import { TS_IncompatibleSignature } from '../util/workarounds';
 
 // tslint:disable:max-line-length
 /** @deprecated resultSelector is no longer supported, use a mapping function. */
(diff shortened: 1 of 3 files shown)


DanielRosenwasser:
Can someone explain what exactly broke/is not working anymore?

cartant:
TypeScript Playground repro [here](https://www.typescriptlang.org/play?ts=4.2.0-beta#code/GYVwdgxgLglg9mABAIxmAJgYQIYBtfLYQDWAFAFCKIR4FHEBi4EAXIqQHRfYBOA5gGc22MAE8A2gF0AlIgC8APkQA3ODHTlpbVeoDc5cqEiwEKNFlqESAHgCCiAKYAPKA4wDEPB9nQJcoxHBiMDgAdzApABpEACVHFzd0Dy8fPwCgkPCpBQoqGnwrRmY2Tm5+IURxLg5baNKOLwqY2UUVNXQZeSUdDS02vQMjaHgkVAwcAvpc6kt6JkgS6t5BYTEpFu72zUQAbwBfIA).

In the playground's 4.2.0-dev, this error is effected:

```
This overload signature is not compatible with its implementation signature.(2394)
```

No such error is effected in 4.1.

benlesh (approved):

DanielRosenwasser:
I believe this was catching a legit arity issue. Maybe try the following:

```ts
function bindCallback<A extends readonly unknown[], R extends readonly unknown[]>(
  callbackFunc: (...args: [...A, (...res: R) => void]) => void
): void;

function bindCallback(
  callbackFunc: (...args: [unknown[], (...res: any) => void]) => void
) {}

```

cartant:
Using `(...args: [unknown[], (...res: any) => void]) => void` [doesn't effect the error](), but that doesn't make a great deal of sense, to me, as one of the [overload signatures](https://github.com/ReactiveX/rxjs/blob/9da638a70d5abb862439ab4ee6a55368228811b0/src/internal/observable/bindCallback.ts#L9-L13) has `(...args: any[]) => void` as the callback type (I omitted it from the repro to make the repro minimal, but I've put it back). How can it be that it's okay to have a more specific implementation signature?

PR #6057, opened 2021-02-26
Feat/ajax response headers

benlesh:
Adds a feature that allows the user to examine response headers more easily (based loosely off of prior art from contemporary libraries):

```ts
ajax(url).subscribe(response => {
   console.log(response.responseHeaders); 
});
   /**
    * logs:
    * { "content-type": "application/json", "x-custom-header": "etc" }
    */
```

### DIFF
--- a/src/internal/ajax/AjaxResponse.ts
+++ b/src/internal/ajax/AjaxResponse.ts
@@ -97,14 +97,15 @@ export class AjaxResponse<T> {
     // other-header-here: some, other, values, or, whatever
     const allHeaders = xhr.getAllResponseHeaders();
     this.responseHeaders = allHeaders
-      ? Object.fromEntries(
-          // Split the header text into lines
-          allHeaders.split('\n').map((line) => {
-            // Split the lines on the first ": ". The value could technically have a ": " in it.
-            const index = line.indexOf(': ');
-            return [line.slice(0, index), line.slice(index + 2)];
-          })
-        )
+      ? // Split the header text into lines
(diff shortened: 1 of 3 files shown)


benlesh:
Ugh... `Object.fromEntries` is too new 😿

benlesh:
No biggy... it's probably more efficient as a reduce :)

src/internal/ajax/AjaxResponse.ts
```diff
@@ -76,6 +87,27 @@ export class AjaxResponse<T> {
...
+      ? // Split the header text into lines
+        allHeaders.split('\n').reduce((headers: Record<string, string>, line) => {
+          // Split the lines on the first ": " as
+          // "key: value". Note that the value could
+          // technically have a ": " in it.
+          const index = line.indexOf(': ');
+          headers[line.slice(0, index)] = line.slice(index + 2);
+          return headers;
+        }, {})
+      : {};
```
cartant, line 109: Maybe we could implement `responseHeaders` as a getter so that the headers don't have to be parsed for every response?

cartant (approved):
LGTM, but maybe use a getter?

src/internal/ajax/AjaxResponse.ts
benlesh: I did that initially, but then I had a add a private property, tests got messy, I got lazy because it was 2am, etc. Happy to take another pass at that in a follow up PR, though.

PR #6062, opened 2021-02-26
fix: finalization/teardown fixes for AsyncGenerators and take

### LINKED ISSUE #5998: from(AsyncIterable) doesn't finalize async generators
## Bug Report

**Current Behavior**
When using `from()` with an async generator, the generator is never finalized.

```ts
import { async, concat, from, Observable, Subscriber } from "rxjs";
import { take, tap } from "rxjs/operators";

async function* gen() {
  try {
    let i = 0;
    while (true) {
      yield i++;
      // added or Chrome hangs
      await new Promise((resolve) => setTimeout(resolve, 100));
    }
  } finally {
    console.log("Finalizing generator");
  }
}

const source = from(gen()).pipe(
  take(3),
  tap({ complete: () => console.log("Done") })
);

```
yields:
```
0
1
2
Done
```

**Expected behavior**
I would expect for the async generator to be finalized, just as is done for the non-async generators.

The above code should yield:
```
0
1
2
Done
Finalizing generator
```

**Reproduction**

https://stackblitz.com/edit/from-async-generator?file=index.ts

**Environment**
- Runtime: both Node v12 and Chrome v88
- RxJS version: 7.0.0-beta.10

**Possible Solution**
The `fromAsyncIterable()` function in `rxjs/src/internal/observable/from.ts` should handle this case, as does `fromIterable()`.

Something along the lines:
```ts
function fromAsyncIterable<T>(asyncIterable: AsyncIterable<T> | AsyncGenerator<T>) {
  return new Observable((subscriber: Subscriber<T>) => {
    process(asyncIterable, subscriber).catch(err => subscriber.error(err));
    return () => {
      if (isFunction(asyncIterable?.return)) asyncIterable.return(null);
    };
  });
}

async function process<T>(
  asyncIterable: AsyncIterable<T> | AsyncGenerator<T>,
  subscriber: Subscriber<T>
) {
  for await (const value of asyncIterable) {
    subscriber.next(value);
  }
  subscriber.complete();
}
```

benlesh:
- Resolves an issue where AsyncGenerators were not properly finalized when an observable result was unsubscribed.
- Resolves another nearly impossible to test scenario that this happens to cover: In the event of an operator like `take()` completing the resulting observable, if an error is thrown in the terminal completion handler, teardown is not invoked, as it would not continue to the next line to call the unsubscribe method of the Subscriber instance. This was resolved by ensuring synchronous calls before the unsubscribe call are called in a try/finally block.

Fixes #5998

### DIFF
--- a/spec/observables/from-spec.ts
+++ b/spec/observables/from-spec.ts
@@ -1,7 +1,7 @@
 import { expect } from 'chai';
 import { TestScheduler } from 'rxjs/testing';
 import { asyncScheduler, of, from, Observer, observable, Subject, noop } from 'rxjs';
-import { first, concatMap, delay } from 'rxjs/operators';
+import { first, concatMap, delay, take, map, tap } from 'rxjs/operators';
 
 // tslint:disable:no-any
 declare const expectObservable: any;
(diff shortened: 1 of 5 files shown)


spec/observables/from-spec.ts
```diff
@@ -34,6 +34,114 @@ describe('from', () => {
...
+      next: value => results.push(value),
+      complete: () => {
+        results.push('done');
+        setTimeout(() => {
+          try {
+            expect(sideEffects).to.deep.equal([0, 1, 2]);
+            expect(results).to.deep.equal([0, 1, 2, 'done', 'finalized generator']);
+          } finally {
+            done();
+          }
```
cartant, line 68: nitpick: I don't see why a `finally` is needed here. Calling `done` before allowing a failed expectation to bubble out seems unnecessary. AFAICT, with a failed expectation, the test will fail as you'd expect.

spec/observables/from-spec.ts
```diff
@@ -34,6 +34,114 @@ describe('from', () => {
...
+      next: value => results.push(value),
+      error: () => {
+        results.push('in error');
+        setTimeout(() => {
+          try {
+            expect(sideEffects).to.deep.equal([0, 1, 2]);
+            expect(results).to.deep.equal([0, 1, 'in error', 'finalized generator']);
+          } finally {
+            done();
+          }
```
cartant, line 111: nitpick: Or here.

cartant (approved):
LGTM

spec/observables/from-spec.ts
benlesh: It was behaving strangely without that try/finally, IIRC. The test would hang when the expectation failed and it wouldn't tell me what failed. I can try removing it and forcing a failure just to confirm.

benlesh: Well, no idea why I added that now. Removed. 🤷‍♂️

PR #6063, opened 2021-02-27
chore: remove bindCallback workaround

cartant:
**Description:**

This PR removes the workaround type that was added for `bindCallback` in #6050. TypeScript is correct that the signatures aren't compatible because callback arguments are contravariant.

**Related PR:** #6050

### DIFF
--- a/src/internal/observable/bindCallback.ts
+++ b/src/internal/observable/bindCallback.ts
@@ -2,7 +2,6 @@
 import { SchedulerLike } from '../types';
 import { Observable } from '../Observable';
 import { bindCallbackInternals } from './bindCallbackInternals';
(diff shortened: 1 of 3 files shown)


cartant:
BTW, this relies upon a 4.2 feature, so TS 4.2 is our minimum supported version. For 4.1 to be our minimum, we'd have to re-instate the workaround, as 4.2 deems our previous signatures to be incompatible.

benlesh (approved):

benlesh:
> BTW, this relies upon a 4.2 feature, so TS 4.2 will need to be our minimum supported version.

It seems like the output .d.ts files would be okay for 4.1 though, no? Either way, this seems like what we want to land in 7. So... :+1:

PR #6070, opened 2021-03-01
Fix/6064 default if empty and related fixes

### LINKED ISSUE #6064: the `defaultIfEmpty` operator has erroneous & inconvenient typing
## Bug Report

**Current Behavior**
```ts
const a = of(true).pipe(defaultIfEmpty()); // a is Observable<boolean>, but should be Observable<boolean | null>
const b = of(undefined).pipe(defaultIfEmpty(undefined)); // b is Observable<undefined>, but should be Observable<undefined | null>

const c = of(true).pipe(defaultIfEmpty(null));
const d: Observable<boolean | null> = of(true).pipe(defaultIfEmpty(null));
// c & d both report the same error at the defaultIfEmpty operator:
/*
Argument of type 'MonoTypeOperatorFunction<null>' is not assignable to parameter of type 'OperatorFunction<boolean, null>'.
  Types of parameters 'source' and 'source' are incompatible.
    Type 'Observable<boolean>' is not assignable to type 'Observable<null>'.
      Type 'boolean' is not assignable to type 'null'.ts(2345)
*/
```

**Expected behavior**
```ts
const a = of(true).pipe(defaultIfEmpty()); // a is Observable<boolean | null>
const b = of(undefined).pipe(defaultIfEmpty(undefined)); // b is Observable<undefined | null>

const c = of(true).pipe(defaultIfEmpty(null)); // no error, c is Observable<boolean | null>
const d: Observable<boolean | null> = of(true).pipe(defaultIfEmpty(null)); // no error, d is Observable<boolean | null>
```

**Reproduction**
Creating a repo with stackblitz or codesandbox didnt work as changing the TS version/config doesnt seem to affect their editor.
You an just take the code above and drop it into any TS ^4 strict project.

**Environment**
- RxJS version: ^6
- TypeScript: ^4

**Possible Solution**
change the type signatur of `defaultIfEmpty` to something like:
`<T, R>(value?: R) => OperatorFunction<T, T | R>`

I would also deprecate the default of value and require an argument:
`<T, R>(value: R) => OperatorFunction<T, T | R>`
this is because `defaultIfEmpty(undefined)` results actually in the same as `defaultIfEmpty(null)`, which is unexpected.

benlesh:
- fix(defaultIfEmpty): Allow `undefined` as an argument, require an argument
   
   **BREAKING CHANGE:** `defaultIfEmpty` requires a value be passed. Will no longer convert `undefined` to `null` for no good reason.
    
   Resolves #6064
    
- fix(last): Allow `defaultValue` of `undefined`.
- fix(first): Allow `defaultValue` of `undefined`.
- fix(elementAt): Allow `defaultValue` of `undefined`.

### DIFF
--- a/src/internal/operators/defaultIfEmpty.ts
+++ b/src/internal/operators/defaultIfEmpty.ts
@@ -2,10 +2,6 @@ import { OperatorFunction } from '../types';
 import { operate } from '../util/lift';
 import { OperatorSubscriber } from './OperatorSubscriber';
 
-/* tslint:disable:max-line-length */
-export function defaultIfEmpty<T, R = T>(defaultValue?: R): OperatorFunction<T, T | R>;
-/* tslint:enable:max-line-length */
-
 /**
  * Emits a given value if the source Observable completes without emitting any
  * `next` value, otherwise mirrors the source Observable.
@@ -34,13 +30,13 @@ export function defaultIfEmpty<T, R = T>(defaultValue?: R): OperatorFunction<T, 
  * @see {@link empty}
  * @see {@link last}
  *
- * @param {any} [defaultValue=null] The default value used if the source
+ * @param defaultValue The default value used if the source
  * Observable is empty.
- * @return {Observable} An Observable that emits either the specified
+ * @return An Observable that emits either the specified
  * `defaultValue` if the source Observable emits no items, or the values emitted
  * by the source Observable.
  */
-export function defaultIfEmpty<T, R>(defaultValue: R | null = null): OperatorFunction<T, T | R> {
+export function defaultIfEmpty<T, R>(defaultValue: R): OperatorFunction<T, T | R> {
   return operate((source, subscriber) => {
     let hasValue = false;
     source.subscribe(
--- a/src/internal/operators/elementAt.ts
+++ b/src/internal/operators/elementAt.ts
@@ -6,7 +6,7 @@ import { throwIfEmpty } from './throwIfEmpty';
 import { defaultIfEmpty } from './defaultIfEmpty';
 import { take } from './take';
 
-export function elementAt<T, D>(index: number): MonoTypeOperatorFunction<T>;
+export function elementAt<T>(index: number): MonoTypeOperatorFunction<T>;
 export function elementAt<T, D>(index: number, defaultValue: D): OperatorFunction<T, T | D>;
 
 /**
--- a/src/internal/operators/elementAt.ts
+++ b/src/internal/operators/elementAt.ts
@@ -1,10 +1,13 @@
 import { ArgumentOutOfRangeError } from '../util/ArgumentOutOfRangeError';
 import { Observable } from '../Observable';
-import { MonoTypeOperatorFunction } from '../types';
+import { MonoTypeOperatorFunction, OperatorFunction } from '../types';
 import { filter } from './filter';
 import { throwIfEmpty } from './throwIfEmpty';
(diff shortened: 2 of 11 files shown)


src/internal/operators/defaultIfEmpty.ts
```diff
@@ -34,13 +30,13 @@ export function defaultIfEmpty<T, R = T>(defaultValue?: R): OperatorFunction<T,
...
  *
- * @param {any} [defaultValue=null] The default value used if the source
+ * @param defaultValue The default value used if the source
  * Observable is empty.
- * @return {Observable} An Observable that emits either the specified
+ * @return An Observable that emits either the specified
  * `defaultValue` if the source Observable emits no items, or the values emitted
  * by the source Observable.
  */
-export function defaultIfEmpty<T, R>(defaultValue: R | null = null): OperatorFunction<T, T | R> {
```
benlesh, line 43: The problem was here. This made it so you couldn't pass `undefined` as an argument. That cascaded through other operators that were composed using this one.

src/internal/operators/elementAt.ts
```diff
@@ -1,11 +1,14 @@
 import { ArgumentOutOfRangeError } from '../util/ArgumentOutOfRangeError';
 import { Observable } from '../Observable';
-import { MonoTypeOperatorFunction } from '../types';
+import { MonoTypeOperatorFunction, OperatorFunction } from '../types';
 import { filter } from './filter';
 import { throwIfEmpty } from './throwIfEmpty';
 import { defaultIfEmpty } from './defaultIfEmpty';
 import { take } from './take';
 
+export function elementAt<T, D>(index: number): MonoTypeOperatorFunction<T>;
```
backbone87, line 9: nit: unused type variable `D`

src/internal/operators/elementAt.ts
```diff
@@ -1,11 +1,14 @@
...
-import { MonoTypeOperatorFunction } from '../types';
+import { MonoTypeOperatorFunction, OperatorFunction } from '../types';
 import { filter } from './filter';
 import { throwIfEmpty } from './throwIfEmpty';
 import { defaultIfEmpty } from './defaultIfEmpty';
 import { take } from './take';
 
+export function elementAt<T>(index: number): MonoTypeOperatorFunction<T>;
+export function elementAt<T, D>(index: number, defaultValue: D): OperatorFunction<T, T | D>;
+
```
cartant, line 11: nitpick: Why are these overloads necessary? They don't seem to do anything other than make `defaultValue` option - which it already is in the implementation signature.

cartant (approved):
LGTM. Just a nitpick.

src/internal/operators/elementAt.ts
kolodny: I can see removing this breaking folks specifying one generic and not using defaultValue. But we don't care about that use case anymore IIRC

kolodny (approved):

src/internal/operators/elementAt.ts
benlesh: It's required. Otherwise this fails, and you get `Observable<unknown>` because `D` is `unknown`, I guess.

```ts
  const o = of('foo').pipe(elementAt(47)); // $ExpectType Observable<string>
```

cartant: https://github.com/ReactiveX/rxjs/blob/c6ba9790ed42e4781278822aff96bce0863e7917/src/internal/operators/last.ts#L12

PR #6072, opened 2021-03-02
feat(bindNodeCallback): Works on overloaded functions

### LINKED ISSUE #5942: TS Typing issue with bindNodeCallback and readdir
## Bug Report

**Current Behavior**
Type of Observable created by bindNodeCallback(readdir) is not properly advertised.
causes VSCode to issue following complaint:
` Type 'Observable<unknown>' is not assignable to type 'Observable<string[]>'.`

**Reproduction**
Paste following code into a VSCode editor.
```ts
import {readdir} from "fs";
import {Observable,bindNodeCallback} from "rxjs";

function readFileObs(pathname: string): Observable<string[]> {
  const readFileAsObservable = bindNodeCallback(readdir);
  return readFileAsObservable(pathname);
}
```

**Expected behavior**
bindNodeCallback(readdir) should advertise Observable<string[]> as return type.

**Environment**
- Runtime: node: v15.2.1
- RxJS version: 6.6.3
- (If bug is related) Loader, build configuration:  TS: 4.1.3, VSCode: 1.52.1 (MacOS) 

node typings correctly advertises readdir, but bindNodeCallback somehow fails to advertise the correct type.
image

kolodny:
I thought this PR would be a lot simpler 😅 

Fixes #5942

### DIFF
--- a/src/internal/types.ts
+++ b/src/internal/types.ts
@@ -276,3 +276,99 @@ export type ValueFromNotification<T> = T extends { kind: 'N' | 'E' | 'C' }
 export type Falsy = null | undefined | false | 0 | -0 | 0n | '';
 
 export type TruthyTypesOf<T> = T extends Falsy ? never : T;
+
+/**
+ * One of those magic types that can extract types from an
+ * overloaded function. See https://github.com/ReactiveX/rxjs/issues/5942
+ * for more details
+ */
+type OverloadedArgumentsAndReturnType<T> = T extends {
+  (...args: infer A1): infer R;
+  (...args: infer A2): infer R;
+  (...args: infer A3): infer R;
+  (...args: infer A4): infer R;
+  (...args: infer A5): infer R;
+  (...args: infer A6): infer R;
+  (...args: infer A7): infer R;
+  (...args: infer A8): infer R;
+  (...args: infer A9): infer R;
+  (...args: infer A10): infer R;
+}
+  ? [A1 | A2 | A3 | A4 | A5 | A6 | A7 | A8 | A9 | A10, R]
+  : T extends {
+      (...args: infer A1): infer R;
+      (...args: infer A2): infer R;
+      (...args: infer A3): infer R;
+      (...args: infer A4): infer R;
+      (...args: infer A5): infer R;
+      (...args: infer A6): infer R;
+      (...args: infer A7): infer R;
+      (...args: infer A8): infer R;
+      (...args: infer A9): infer R;
+    }
+  ? [A1 | A2 | A3 | A4 | A5 | A6 | A7 | A8 | A9, R]
+  : T extends {
+      (...args: infer A1): infer R;
+      (...args: infer A2): infer R;
+      (...args: infer A3): infer R;
+      (...args: infer A4): infer R;
+      (...args: infer A5): infer R;
+      (...args: infer A6): infer R;
+      (...args: infer A7): infer R;
+      (...args: infer A8): infer R;
+    }
+  ? [A1 | A2 | A3 | A4 | A5 | A6 | A7 | A8, R]
+  : T extends {
+      (...args: infer A1): infer R;
+      (...args: infer A2): infer R;
+      (...args: infer A3): infer R;
+      (...args: infer A4): infer R;
+      (...args: infer A5): infer R;
+      (...args: infer A6): infer R;
+      (...args: infer A7): infer R;
+    }
+  ? [A1 | A2 | A3 | A4 | A5 | A6 | A7, R]
+  : T extends {
+      (...args: infer A1): infer R;
+      (...args: infer A2): infer R;
+      (...args: infer A3): infer R;
+      (...args: infer A4): infer R;
+      (...args: infer A5): infer R;
+      (...args: infer A6): infer R;
+    }
+  ? [A1 | A2 | A3 | A4 | A5 | A6, R]
+  : T extends {
+      (...args: infer A1): infer R;
+      (...args: infer A2): infer R;
+      (...args: infer A3): infer R;
+      (...args: infer A4): infer R;
+      (...args: infer A5): infer R;
+    }
+  ? [A1 | A2 | A3 | A4 | A5, R]
+  : T extends {
+      (...args: infer A1): infer R;
+      (...args: infer A2): infer R;
+      (...args: infer A3): infer R;
+      (...args: infer A4): infer R;
+    }
+  ? [A1 | A2 | A3 | A4, R]
+  : T extends {
+      (...args: infer A1): infer R;
+      (...args: infer A2): infer R;
+      (...args: infer A3): infer R;
+    }
+  ? [A1 | A2 | A3, R]
+  : T extends {
+      (...args: infer A1): infer R;
+      (...args: infer A2): infer R;
+    }
+  ? [A1 | A2, R]
+  : T extends (...args: infer A) => infer R
+  ? [A, R]
+  : never;
+
+export type OverloadedParameters<T> = OverloadedArgumentsAndReturnType<T>[0];
+export type OverloadedReturnType<T> = OverloadedArgumentsAndReturnType<T>[1];
+
+export type AllButLast<T extends any[]> = T extends T ? (T extends [...(infer H), any] ? H : never) : never;
+export type Last<T extends any[]> = [any, ...T][T['length']];
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -1,3 +1,5 @@
+export declare type AllButLast<T extends any[]> = T extends T ? T extends [...infer H, any] ? H : never : never;
+
 export declare const animationFrame: AnimationFrameScheduler;
 
 export declare function animationFrames(timestampProvider?: TimestampProvider): Observable<{
@@ -35,10 +37,10 @@ export declare class BehaviorSubject<T> extends Subject<T> {
 }
 
 export declare function bindCallback(callbackFunc: (...args: any[]) => void, resultSelector: (...args: any[]) => any, scheduler?: SchedulerLike): (...args: any[]) => Observable<any>;
-export declare function bindCallback<A extends readonly unknown[], R extends readonly unknown[]>(callbackFunc: (...args: [...A, (...res: R) => void]) => void, schedulerLike?: SchedulerLike): (...arg: A) => Observable<R extends [] ? void : R extends [any] ? R[0] : R>;
+export declare function bindCallback<Fn extends (...args: any[]) => any>(callbackFunc: Fn, schedulerLike?: SchedulerLike): (...arg: AllButLast<OverloadedParameters<Fn>>) => Observable<Last<Parameters<Fn>> extends () => any ? void : Last<Parameters<Fn>> extends (arg: infer R) => any ? R : Last<Parameters<Fn>> extends (...args: infer R) => any ? R : never>;
 
 export declare function bindNodeCallback(callbackFunc: (...args: any[]) => void, resultSelector: (...args: any[]) => any, scheduler?: SchedulerLike): (...args: any[]) => Observable<any>;
-export declare function bindNodeCallback<A extends readonly unknown[], R extends readonly unknown[]>(callbackFunc: (...args: [...A, (err: any, ...res: R) => void]) => void, schedulerLike?: SchedulerLike): (...arg: A) => Observable<R extends [] ? void : R extends [any] ? R[0] : R>;
+export declare function bindNodeCallback<Fn extends (...args: any[]) => any>(callbackFunc: Fn, schedulerLike?: SchedulerLike): (...arg: AllButLast<OverloadedParameters<Fn>>) => Observable<Last<OverloadedParameters<Fn>> extends (err: any) => any ? void : Last<Parameters<Fn>> extends (err: any, arg: infer R) => any ? R : Last<Parameters<Fn>> extends (err: any, ...args: infer R) => any ? R : never>;
 
 export declare function combineLatest(sources: []): Observable<never>;
 export declare function combineLatest<A extends readonly unknown[]>(sources: readonly [...ObservableInputTuple<A>]): Observable<A>;
@@ -211,6 +213,8 @@ export declare function interval(period?: number, scheduler?: SchedulerLike): Ob
 
 export declare function isObservable<T>(obj: any): obj is Observable<T>;
 
+export declare type Last<T extends any[]> = [any, ...T][T['length']];
+
 export declare function lastValueFrom<T>(source: Observable<T>): Promise<T>;
 
 export declare function merge<A extends readonly unknown[]>(...args: [...ObservableInputTuple<A>]): Observable<A[number]>;
@@ -360,6 +364,10 @@ export interface Operator<T, R> {
 
 export interface OperatorFunction<T, R> extends UnaryFunction<Observable<T>, Observable<R>> {
 }
+
+export declare type OverloadedParameters<T> = OverloadedArgumentsAndReturnType<T>[0];
+
+export declare type OverloadedReturnType<T> = OverloadedArgumentsAndReturnType<T>[1];
 
 export declare function pairs<T>(arr: readonly T[], scheduler?: SchedulerLike): Observable<[string, T]>;
 export declare function pairs<O extends Record<string, unknown>>(obj: O, scheduler?: SchedulerLike): Observable<[keyof O, O[keyof O]]>;
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -1,4 +1,4 @@
-export declare type AllButLast<T extends any[]> = T extends T ? T extends [...infer H, any] ? H : never : never;
+export declare type AllButLast<T extends any[]> = T extends T ? (T extends [...(infer H), any] ? H : never) : never;
 
 export declare const animationFrame: AnimationFrameScheduler;
 
--- a/spec-dtslint/observables/bindCallback-spec.ts
+++ b/spec-dtslint/observables/bindCallback-spec.ts
@@ -1,6 +1,5 @@
 import { bindCallback } from 'rxjs';
 import { a,  b,  c,  d,  e,  f,  g, A, B, C, D, E, F, G } from '../helpers';
-import { SchedulerLike } from '../../src';
 
 describe('callbackFunc', () => {
   const f0 = (cb: () => void) => {
@@ -268,4 +267,23 @@ describe('callbackFunc overkill' , () => {
     };
     const o = bindCallback(fa10cb5) // $ExpectType (_1: 1, _2: 2, _3: 3, _4: 4, _5: 5, _6: 6, _7: 7, _8: 8, _9: 9, _10: 10) => Observable<[_11: 11, _12: 12, _13: 13, _14: 14, _15: 15]>
   });
+});
+
+describe('works with overloads' , () => {
+  function foo(num: number, cb: (result: boolean) => void): void;
+  function foo(num: string, cb: (result: boolean) => void): void;
+  function foo(num: string, flag: boolean, cb: (result: boolean) => void): void;
+  function foo(...args: any[]) {}
+
+  const o = bindCallback(foo);
+  o(1).subscribe(v => {
+    v // $ExpectType boolean
+  });
+  o('1').subscribe(v => {
+    v // $ExpectType boolean
+  });
+  o('1', true).subscribe(v => {
+    v // $ExpectType boolean
+  });
+  o(1, true); // $ExpectError
 });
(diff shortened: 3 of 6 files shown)


src/internal/types.ts
```diff
@@ -276,3 +276,99 @@ export type ValueFromNotification<T> = T extends { kind: 'N' | 'E' | 'C' }
 export type Falsy = null | undefined | false | 0 | -0 | 0n | '';
```
kolodny, line 276: @benlesh Completely off topic, but shouldn't this include `NaN`?

kolodny:
Note, the only instance where this fails is if the overload has a different shape. For some reason changing `Parameters<` to `OverloadedParameters<` breaks the typings. Maybe @cartant knows why

cartant:
> Maybe @cartant knows why

TBH, my fantasy is that once v7 is released I'll no longer need deal with TypeScript and can ride off into the sunset and lead a simpler life.

src/internal/types.ts
```diff
@@ -276,3 +276,99 @@ export type ValueFromNotification<T> = T extends { kind: 'N' | 'E' | 'C' }
...
 
 export type TruthyTypesOf<T> = T extends Falsy ? never : T;
+
+/**
+ * One of those magic types that can extract types from an
+ * overloaded function. See https://github.com/ReactiveX/rxjs/issues/5942
+ * for more details
+ */
+type OverloadedArgumentsAndReturnType<T> = T extends {
```
benlesh, line 285: I don't understand this at all. haha. T extends a function that.. looks the same in 10 cases? But... isn't the same? But it's the same... but different.

same-but-different

src/internal/types.ts
```diff
@@ -276,3 +276,99 @@ export type ValueFromNotification<T> = T extends { kind: 'N' | 'E' | 'C' }
...
+  : T extends {
+      (...args: infer A1): infer R;
+      (...args: infer A2): infer R;
+    }
+  ? [A1 | A2, R]
+  : T extends (...args: infer A) => infer R
+  ? [A, R]
+  : never;
+
+export type OverloadedParameters<T> = OverloadedArgumentsAndReturnType<T>[0];
```
benlesh, line 370: Yeah. We need comments for all of this, because I literally have no idea what's going on here and I don't think I'm *that* bad at TypeScript.

src/internal/types.ts
benlesh: Wait... this is a type that is a "better `Parameters<T>`"? I still don't quite understand how it works.

src/internal/types.ts
benlesh: Yeah, probably.

cartant: `Nan` is not a type; it's a value, AFAICT; `number` is the type.

cartant:
> Note, the only instance where this fails is if the overload callback has a different shape. For some reason changing `Parameters<` to `OverloadedParameters<` breaks the typings.

I have a few questions about this:

- What does the failure look like?
- And how hard will it be for a caller to figure out what's happening?
- How would a caller work around the problem?
- Are there many/any callback APIs that have differing callback shapes?

src/internal/types.ts
kolodny: Yeah this is the magical overloaded function inference technique described at https://stackoverflow.com/questions/52760509/typescript-returntype-of-overloaded-function/52761156#52761156

kolodny:
> What does the failure look like?
It falls through to the never case
> And how hard will it be for a caller to figure out what's happening?
Anyone working with overloading functions should understand they're pretty busted from a TS inference POV
> How would a caller work around the problem?
There really isn't any way to specify which overload they're interested in
> Are there many/any callback APIs that have differing callback shapes?
I imagine this is pretty uncommon so I'm not too worried about it. The only place I see it being used is for encoding on fs.readFile which returns a string instead of a buffer when encoding is used

Note that the current PR is loads better then the default TS behavior of "let's pretend we only know about the last overload" approach

I can spend a bit more time on this but this problem is pretty in the weeds already

kolodny:
Hmm, can't make much progress here. If anyone wants to try their hand at it, I isolated the issue to a playground:

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAKjgQwM5wGbo1CI4DkWBA3AFBkD0lcAEgKZT0pNwwAWzAJsBho-QB28CADdGAGwjIumAK6CAxjGARB2aHCYyAYsAn0AXBSwA6bVz0GAFEQgQCAGjjXGUZwCM5GAJRwAvAB8cADeZHARcF4Yph4AnjD0APJ8qPTw1HAAQt78UGQAvj7kZGYWVvS2GPZOoQXOrlDuUd5+QaHhkdGxCcmp6XCZOXyMhcUmqOb0uvqVdg7OIUKKEDyCAOaGhGiKwMAE9S5unq0BwWGRLTEwEACqYGCMAMJolX6ZqDBQwBtjJVQ0AAq7GA6FBbE4cFQIGQsDgYFwIFBvDi8iUKjUEOQ8FQ7AgcgksgA7tAANZwInADhwMSSaRceiyDAKZSqdRwCTAUnMMrTSyzYylFkYwTwxHIjBxAA8OlF9AAHolBFx0NZTOrYetUFtkII4gBtAC6bWCurigWsGEEW1lPi2ao1UC1WwAghIJDkYAAZNAwKVJcRQKQyRkABVhyBA6UYqBlgkCgRNcFD4rSUs6ER9nyl4agkejUFjsoTcAVSpVRyaOr1ziYqAJMC2PzycAASkmzXAAPxtuBbQT0QNkc4ZrTpORQUUO0ya1BJgdE5OpyrWOsQCTiWv0ABW9GUSYul0iK3U8EUHgCLC1pkgYGs4yPkSt09nDTc1biW-rEkbKD1B9HR8Il4SsoD8JgYAnUUmF3ZRGjA8ggKPNcN0qOsGwfICilHIpyAKCgT0+McZgMF1UDiJRLwRPAJTiS1JnKWZxkY0jyKUWxEk+Ah3hoJIAGkyBY+gyIoxQOPoLjFiKQZeIEoSRPYghOJgWoliUVYfk2bZUF2fZpMyfiKHktixKUiSVJ8UwOCEawL3aDwekSFIsHSHi4EM4zRPEyS6ks6zBFss4okcvoXJgNyPL5CoFNM5TVOWDSNi2Agdj2A4-M4AK7OCBz4ic-pwpkuAAEJSrgABJAA5J4klbVsAFEnkBLZgXBXF8UJKJmGQKEvk0krSooIaKBgOJHncwNgwZLgXSdOQo2EVAXWVVtx0nQExvoKVAWCfw4EBUtFSECtD0uF8nW1OBm0YOAXQARjtK7BBbVtEKPc7nSelsXQAJke66oDbN6zvVGcLqbZ6bpdABmf7IcB17AJcUHZwh76ABY4Ze4G4HwiIe31e64AAH1un6Sdu6GKZddHnFbQ1Ohaw7y3QU7Ig+y6Aduh60ZuxHHw53nAd+rG+ZxiJBa+qHYaFoGcM6AmidJ37qehumGYiJmy2O1mkclrn7tFhHxeRx1PoNv7Zf5yI8e7OBCbu6mfvVxn9uZnXTbB834dupMudbBX7ZdF3NbgAdAxKBVIDhUbxoDOkQy4XN80SQttt2iaE+m2b1nmoQYCWla1sEDbHnT-UAAZDXIKPoHgWPmHjoN6UZVbIPWzb08vJupsZHO88W5auDbqDS62nb9Tu6uKFrmPNtu91PSzP0Du15V0DNI0M9Xo717dntrB3ln7dB6wudoHxnDNQ07doPsw8HRg-H7R+oBr+Vo-r+fl+292983w0Gd9RmmcKDQEhp9SAn1AQAwGwOAEENNPIAA

benlesh:
I think that TypeScript just can't solve this issue, TBH.

kolodny:
> I think that TypeScript just can't solve this issue, TBH.

In that case this PR should be good to go.

cartant:
IDK why `Last<OverloadedParameters<...>>` has an `undefined` in it. How can a union of three tuples - `FP` - effect a union of four types? That makes little sense and that's what the callback parameters are extracted from, so that might be part of the problem:

```ts
declare function f(a: "a"): "a";
declare function f(a: "a", b: "b"): "ab";
declare function f(a: "a", b: "b", c: "c"): "abc";
type FP = OverloadedParameters<typeof f>; // = [a: "a"] | [a: "a", b: "b"] | [a: "a", b: "b", c: "c"]
type FR = OverloadedReturnType<typeof f>; // = "a" | "ab" | "abc"
type FABL = AllButLast<FP>; // = [] | [a: "a"] | [a: "a", b: "b"]
type FL = Last<FP>; // = "a" | "b" | "c" | undefined
```

I'm not sure this problem is worth the effort, TBH.

kolodny:
Interesting, this definition of `Last` works as expected:

```ts
type Last<T> = T extends [...any, infer Last] ? Last : never;
```

kolodny:
It's possible we're union-ing the arg types too soon, we may need them to flow as distinct tuples as long as possible. Honestly this is so technical that it's completely beyond what RxJS should be providing. If anyone wants to try their hand at it purely for the academic challenge. This is pushing the envelope a bit too far IMO.

Anyway I believe [keeping the arg like this _may_ be a solution](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAKjgQwM5wGbo1CI4DkWBA3AFAUUA2ApvMlAOaoBccA8gG41RUTIATGgICCTAK4gaAOxioR0gQCU64qNIAqATzA0APDB00IGTKgB0UGoIBiwWgD5yDZnAC8cANoBvMnH9wIgCMbJ4EBAA0cN4AvlEAFDxQUVaoAJTuDtGp5gBGWjA0ADIyjDAAFjEAuhF+ASIATKHhUd4yAMYQAsDSjGwEaO3AwARxcIlQyXCpGW5Z3jkwEACqYLpQAMJoNNW1AYEAzM2R0WMTUzOZ2TTobneEAELiGBg8cAA+cKgwUD2McOUeDRRjU6v4RAAWY4JJIpG6zeY5fKFEq9Cq7MhjbxKEJwcSKGgYHrCKJKJp4glE6QkuBKI4UoRUmlKKEMwnEgTVciUMiGXQcbi8fhCUQSKSyeSKFQwNSaIx6DRZDwaOA0AAehUU6F8+388XMBpcrDgPVeUECQTSbFNbxx5F14wN5iN1ukZsCDStJrdtoa9t1+sNTGNNvNIgOXtDtIO-v2gedwdd7shkZ95pZsYC8Zd3uTAFZU+6lHnMzE6gB+LzeYIsYJRRq1hr1o7h+tQyH1vNsER5rE4lg4iJkgdNukDg5DqEs0ld4vVOpsFXqzUCbVgrNOnNR4KF21BTN6zeJ3NvRq79N+9eHoPMJOniN39Mxq+Om8htOBCHn2kQ0sVqs1nWHqNs2tYTp+tYQn2ISDsOZJDkcdKTgOUFVAucBLhqMirtEL7Zse26Wo+tL7nhR63ieYaesRZIHq+CYUduD6UdGf7+JWPiAUE9ZNI0oHhtBA7cXBo6IQc87+IuqpYVquEOvhjEfjuNGkfJ5Hvsm1EsbR65luxAEhEBDaNIJsFNGSElwFJy7YegCkaaeCLaf+nG1n2A6WWw1KCty6qQLAcB8jQAo8HwgjCAACgwyBSIUUCoAqSohUK4WiowkgyHICjKKo6jaLoiWeAADFU5B+dA8BBclYUitKsr5foiruNVwrCGI6XillUq5XKBWKp4QSlRQ5UBVVIhUFQTwwEUaAwAq0krugyDSFonhVElmGLRhcCVvEm22V4TrxFGAASaRRMtWhVDtcAnVZcDeTwGReTQPm8kYcAzd8iXNftsmeJuK1RFGX0wNdlag-dj1QOQQA)

cartant:
> This is pushing the envelope a bit too far IMO.

I think one of the tests that we should apply to this library's types could be framed as a (rhetorical) question to the TypeScript maintainers: "did you expect these TypeScript features to be used like this?"

And if the answer is (likely to be) "no", we probably shouldn't do it and library users will just have to live with it and implement their own workaround or take the `any` escape hatch. TypeScript has become an increasing source of frustration for me, and I think that there are limits to which a language that is not statically typed can be tamed with an after-the-fact type system.

cartant:
I've had another peek at this. I think the fundamental problem isn't extracting the parameters and the return types of the overloads, it's the disconnect between the all-but-first and last parameters in what's returned the call to `bindCallback` or `bindNodeCallback`, here:

https://github.com/ReactiveX/rxjs/blob/2a501b2de7af6f5f61754fa2862b917a8d04ade9/src/internal/observable/bindCallback.ts#L17-L20

I don't see how you can describe the relationship between them such that all-but-last and last correspond the the same overload signature. I don't think this is possible. And I think if we cannot make it work reliably with `readFile`, we should not do it.

cartant:
I kinda hate myself for this, but a solution popped into my head whilst I was out for a drive. It involves using [specific types for the `bindCallback` and `bindNodeCallback` results](https://www.typescriptlang.org/play?ts=4.2.0-beta#code/JYWwDg9gTgLgBAbzlApgQwCYDFgBsVwC+cAZlBCHAEQkDOVA3AFCiSyJwDyARrSlADc03fEVLlKVKAA8AVvWZMYATzAEAQsAB2GAHIQMKAMJpcubmgDGAawBKKWgFdcMADwAVAHxwAvHHdwKNIwKDq0iExwcAAUAHTxaFAA5rQAXHDaJPxwAIIAjACU6WhaysxRcQnJaRlaWVC5AExFcCVlkTHxsYkp6ZnZOQDMLW3lnVW9tfW5ACwjpczEAPwRFV09NTlm6o4wADJotG75ni08fILC+K6aOvqGJmYWNgAKiWggKCFQtK4HR64Tp5PGNKt1quktrgdvtDsdGqd0ud+EIRCgbto9AZjKZzFZrG8oB8vvxfv94cCQR0wRtIdtduTAYNEVxeCirujbliHrjngT3p9vmS4UzKaD1hDcvTYQCcjMWcjLmiMXdsY88a8BSSfn8RXKxUxiOktCgBPxFCo1HAufccU98YTiUKPN4-AEgiEwuNwZM2gBtAC6BV83jaHRW7jQeFcjsFpJdng6xtN5qYTCCbHgloIUJhjPdwVCGHC-oDrv8gULXoCK2iBc9xbgfq60X6DQAEgUADStUoBuArdtwZNmqDBkep7NwfPl+tF8LNhKlHtt6f9lZ7YdwE2j5hTyPRry+Cse+dNto9rqrvbr6dbneppiGSy4RIESwQLRHZAOZwwdI2mqvL4vYTguK42YQCQP6YDg+BUh+X7wFAeTHqgYEwNEVBgGgMAABZUAUzCId+UCNGhv4uFhOH4VQPZIKEH4YNoSTpFQhyWMAwBUEQRFAA) instead of the general-purpose `OverloadedArgumentsAndReturnType`.

If you look at the snipped, you'll see that the default and encoding overloads return the expected types. It would need more work done, as, IIRC, the `bindCallback` and `bindNodeCallback` implementations unpack arrays containing a single value and emit the value instead.

I still don't know how I feel about this solution, but I do know how I feel about TypeScript and I want it out of my head.

cartant:
What we really want is a mapped type that can be applied to [function types](https://www.typescriptlang.org/docs/handbook/interfaces.html#function-types), but, as far as I'm aware, mapped types are only applied to object types (with keys) and to tuples. The signatures in function types have no keys so ... 🤷‍♂️

cartant (changes requested):
If we're going to refine the types for `bindCallback` and `bindNodeCallback`, I think we'll need to use the approach I mentioned in [this comment](https://github.com/ReactiveX/rxjs/pull/6072#issuecomment-791884424). That is, I think we'll have to dispense with the more general `OverloadedArgumentsAndReturnType` and use types that are specific to `bindCallback` and `bindNodeCallback`.

What we really want is a TypeScript mechanism for mapping function signatures, but there is no such thing.

cartant:
Blocking this 'cause it's not going to make the cut for v7, IMO.

benlesh:
Core team meeting: Most people don't realize the function returned from `bindCallback` and `bindNodeCallback` is a "one-and-done" sort of thing. You can call it to get an observable, but that observable uses the same underlying `AsyncSubject` every time, and it's never reset/refreshed.

Maybe we need to deprecate this API and come up with something better?

kolodny:
> I kinda hate myself for this, but a solution popped into my head whilst I was out for a drive. It involves using [specific types for the `bindCallback` and `bindNodeCallback` results](https://www.typescriptlang.org/play?ts=4.2.0-beta#code/JYWwDg9gTgLgBAbzlApgQwCYDFgBsVwC+cAZlBCHAEQkDOVA3AFCiSyJwDyARrSlADc03fEVLlKVKAA8AVvWZMYATzAEAQsAB2GAHIQMKAMJpcubmgDGAawBKKWgFdcMADwAVAHxwAvHHdwKNIwKDq0iExwcAAUAHTxaFAA5rQAXHDaJPxwAIIAjACU6WhaysxRcQnJaRlaWVC5AExFcCVlkTHxsYkp6ZnZOQDMLW3lnVW9tfW5ACwjpczEAPwRFV09NTlm6o4wADJotG75ni08fILC+K6aOvqGJmYWNgAKiWggKCFQtK4HR64Tp5PGNKt1quktrgdvtDsdGqd0ud+EIRCgbto9AZjKZzFZrG8oB8vvxfv94cCQR0wRtIdtduTAYNEVxeCirujbliHrjngT3p9vmS4UzKaD1hDcvTYQCcjMWcjLmiMXdsY88a8BSSfn8RXKxUxiOktCgBPxFCo1HAufccU98YTiUKPN4-AEgiEwuNwZM2gBtAC6BV83jaHRW7jQeFcjsFpJdng6xtN5qYTCCbHgloIUJhjPdwVCGHC-oDrv8gULXoCK2iBc9xbgfq60X6DQAEgUADStUoBuArdtwZNmqDBkep7NwfPl+tF8LNhKlHtt6f9lZ7YdwE2j5hTyPRry+Cse+dNto9rqrvbr6dbneppiGSy4RIESwQLRHZAOZwwdI2mqvL4vYTguK42YQCQP6YDg+BUh+X7wFAeTHqgYEwNEVBgGgMAABZUAUzCId+UCNGhv4uFhOH4VQPZIKEH4YNoSTpFQhyWMAwBUEQRFAA) instead of the general-purpose `OverloadedArgumentsAndReturnType`.
> 
> If you look at the snippet, you'll see that the default and encoding overloads return the expected types. It would need more work done, as, IIRC, the `bindCallback` and `bindNodeCallback` implementations unpack arrays containing a single value and emit the value instead.
> 
> I still don't know how I feel about this solution, but I do know how I feel about TypeScript and I want it out of my head.

So this has the downside that there's still no way to glue the args with the return type. I think I'll need to change the `function` into a fat arrow function in order to assign types as a single atom [**like this**](https://www.typescriptlang.org/play?ts=4.2.0-beta#code/JYWwDg9gTgLgBAbzlApgQwCYDFgBsVwC+cAZlBCHAEQkDOVA3AFADGuattcA8gEa0ooANzS98AHgAqAPkRM4cGAEIAXHEnNCTJjACeYAgCFgAOwwA5CBhQBhNLly80LANYAlFLQCuuGFNkAvOpwKAAeMChmXAjycAAUAHRJaFAA5rRqpiSCcACCAIwAlGpoJrrMConJaRlwWTm5AEzFcKXlsVUJKemZJtlQeQDMLW0V8UldNb39eQAsI2WacAD8cpUT3bW5DoZeMAAyHH4F0i18AsKiEsZmltZ2Dk6uAAopaCAoEVC04oe0x-lpECxp1NmptrhdgcjuImqc1OdBCIxChxDcLFZbPZHM4XK8oO9PoIfn9jo0gdIQRspnkdntSbDBvCePwkVdUei7ljHrj8YSviSYbkmcCOtSerTIfShbNmYjLii0aYMfdsU88W8PgLfjKKUs1CYUEJBMwdPojMquQ8cS9NUTvv44EFJCFwpEMFxQTS2gBtAC6hSdsjasVWkjQeHEfK1xP80liBqNJu0YUgsEU5slUIZLrCESirTK-sCwTz7q4LtWcVzboLPomcXqAwAEoUADSF3R+lZwZtwRPGqCBgfJvQGOA5ks1-MeuD15JlDtNifd1b7ftwQ2D5hjgjhyMyJ2l2uzn1tDsTZf7VcTjdb5OsCAmf5wXiWzHW9VHqRt6RxGBqJILSch+aq4h43i+I6ASyEgqAwF4UAmIgxAcJ2RDaCwT4vrwEBeGYR5vrcoE8q4cSoJgOD4IUzBYc+8BQPkhF4WYcRUGAaAwAAFlQNGPvRyCNMx+EYGxHHcVQHZIJEWEYKYqRqFQHAsMAwBUEQfFAA)

Is this something we want to do? Also am I missing some magical way to keep it as a `function` (technically it'll be a breaking change unless we want to use this as a hack to force the type to be correct but still use an underlying function) while still having the args overload flavor agree with the return type?

cartant:
Looking at the types for `promisify`, it's clear that the authors cheat:

benlesh:
kill-it-with-fire

PR #6080, opened 2021-03-05
chore: move aliased/deprecated operators into separate files

cartant:
**Description:**

This PR moves some deprecated or aliased operators around so that each is in a separate file. The changes are purely internal to the package.

IMO this:

- makes it a little easier to find things;
- makes it clear that the new, shiny operators like `concatWith` have no icky overload signatures;
- might make `flatMap` easier to find so that people will stop asking about it.


benlesh (approved):

benlesh:
LGTM... but needs rebased after the other PR was merged.

PR #6084, opened 2021-03-05
refactor: remove elementAt overload signatures

cartant:
**Description:**

This PR removes the overload signatures for `elementAt` and uses a default type parameter so that it's possible for (naughty) devs to explicitly specify a single type parameter. (This was the approach used with the related signatures in `first` and `last`.)


benlesh (approved):

benlesh:
Naughty dev! 😮

PR #6086, opened 2021-03-05
chore: update mergeMap tests to run mode 

jakovljevic-mladen:
**Description:**
Updated `mergeMap` tests to use run mode.

The first commit is removing duplicated tests. They were completely the same, character after character.

Many tests have never had `expectSubscriptions` assertions, so I added them where possible. Also, some tests were using some random `values` which were never used anywhere else later in the test, so I removed them (example is `should mergeMap many outer to many inner, inner never completes`).

None


cartant (approved):
LGTM. Thanks for separating the changes into two commits.

PR #6092, opened 2021-03-06
refactor: use n-args for all static combineLatest signatures

cartant:
**Description:**

This PR uses the n-args mechanism for _all_ of the static `combineLatest`'s overload signatures.

As mentioned in a couple of other PRs: _I realize that these sigs are deprecated, but [we already have deprecated sigs in operators](https://github.com/ReactiveX/rxjs/blob/a7a04d1110666606a1f2ca6fd38642e6ca74f287/src/internal/operators/mergeWith.ts#L8-L17) that use n-args, so, IMO, things are less confusing if n-args is useds wherever possible._

Further to that:

- The sheer number of overloads removed makes this source much easier to read and reason with.
- It will make it easier to manage the deprecation comments - after the breaking changes are done and v7 is ready, IMO, getting the deprecations sorted/fixed and relating them to the docs is something that needs to be done, as the deprecations confuse _many_ devs.
- There has been some discussion about undeprecating some result selectors and - IMO - this is one API where the result selector is useful.
- Having the types behave differently for depreated and non-deprecated APIs is likely to confuse people.
- The `Scheduler`-related APIs that are deprecated are unlikely to be widely used.
- The `ObservableInput`-based type parameters are pretty unlikely to be specified explicitly anyway.
- If you're still not convinced, let me know and I'll come up with some more reasons, 'cause I really wanna get rid of those overload sigs. 🙂


benlesh (approved):

benlesh:
Approved... but needs rebased after the other PR was merged.

PR #6093, opened 2021-03-06
fix/refactor: add type parameter for thisArg in every

cartant:
**Description:**

This PR adds an overload signature with a type parameter for the `thisArg` parameter in `every` so that if `this` is used in the predicate, it it correctly typed. If an arrow function is used, it behaves as before - regardless of whether a `thisArg` is specified or not.

### DIFF
--- a/src/internal/operators/every.ts
+++ b/src/internal/operators/every.ts
@@ -7,9 +7,9 @@ export function every<T>(
   predicate: BooleanConstructor,
   thisArg?: any
 ): OperatorFunction<T, Exclude<T, Falsy> extends never ? false : boolean>;
-export function every<T>(
-  predicate: (value: T, index: number, source: Observable<T>) => boolean,
-  thisArg?: any
+export function every<T, A = any>(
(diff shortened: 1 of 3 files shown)


src/internal/operators/every.ts
```diff
@@ -7,10 +7,11 @@ export function every<T>(
   predicate: BooleanConstructor,
   thisArg?: any
 ): OperatorFunction<T, Exclude<T, Falsy> extends never ? false : boolean>;
-export function every<T>(
-  predicate: (value: T, index: number, source: Observable<T>) => boolean,
-  thisArg?: any
+export function every<T, A>(
+  predicate: (this: A, value: T, index: number, source: Observable<T>) => boolean,
+  thisArg: A
```
benlesh, line 12: As an aside, I'd like to deprecate everywhere we're doing `thisArg`. Because it's a goofy pattern, and `.bind()` is more obvious: `every(fn, obj)` vs `every(fn.bind(obj))`. (Also closures are probably more readable given how most people would use this, if it's even used)

benlesh (approved):

PR #6095, opened 2021-03-07
fix: forkJoin POJO signature should match only ObservableInput values

cartant:
**Description:**

This PR tightens the `forkJoin` POJO signature so that only POJOs with `ObservableInput` values are matched. The implementation throws if the received POJO contains a value that is not an `ObservableInput`, so the types should reflect that.

### DIFF
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -165,6 +165,7 @@ export declare type Falsy = null | undefined | false | 0 | -0 | 0n | '';
 
 export declare function firstValueFrom<T>(source: Observable<T>): Promise<T>;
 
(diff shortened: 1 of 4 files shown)


api_guard/dist/types/index.d.ts
```diff
@@ -173,7 +174,7 @@ export declare function forkJoin<A extends readonly unknown[], R>(...sourcesAndR
 export declare function forkJoin(sourcesObject: {
     [K in any]: never;
 }): Observable<never>;
-export declare function forkJoin<T>(sourcesObject: T): Observable<{
+export declare function forkJoin<T extends Record<any, ObservableInput<any>>>(sourcesObject: T): Observable<{
```
benlesh, line 177: Can this really be `Record<any`? I feel like the internal implementation will only work with strings.

PR #6097, opened 2021-03-07
fix: predicates that return `any` will now behave property with findIndex

cartant:
**Description:**

The changes in #5987 missed a `findIndex` signature that matched predicates that always returned `false` and was not compatible with predicates typed as returning `any`.

**Related issue (if exists):** #5986 and #5987


benlesh:
😭  this one was my fault. Fortunately it only lived during the beta.

benlesh (approved):

PR #6098, opened 2021-03-07
chore: update min tests to run mode

tmair:
Update `min` tests to run mode.

### DIFF
--- a/spec/operators/min-spec.ts
+++ b/spec/operators/min-spec.ts
@@ -1,255 +1,302 @@
+/** @prettier */
 import { expect } from 'chai';
 import { min, skip, take, mergeMap } from 'rxjs/operators';
-import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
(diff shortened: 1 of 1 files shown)


cartant (approved):
LGTM. Thank you for this; it's much appreciated.

IDK whether it's your intention to update any of the the other tests to run mode, but if it is, you might want to coordinate your efforts with @jakovljevic-mladen, who has also been updating tests to run mode. If you are planning on updating more, I would hate to see either of you discouraged should your efforts be unnecessarily duplicated. But if that isn't something you're planning, then thank you for updating this one. ❤️

jakovljevic-mladen:
@cartant, thanks a lot for linking us together. @tmair contacted me, he wants to help as well. We have agreed that I will continue from where we're currently by converting tests for operators that come in an alphabetical order, while Thomas will start working on converting tests for operators from the end of alphabet (zip and gang). So we eventually meet in the middle one day 😁

PR #6112, opened 2021-03-10
refactor: use n-args for race operator

cartant:
**Description:**

This PR refactors the `race` operator's signatures to use the n-args approach and deletes tests that are no-longer-relevant with the n-args implementation.

_I realize that these sigs are deprecated, but [we already have deprecated sigs in other operators](https://github.com/ReactiveX/rxjs/blob/a7a04d1110666606a1f2ca6fd38642e6ca74f287/src/internal/operators/mergeWith.ts#L8-L17) that use n-args, so, IMO, things are less confusing if n-args is used wherever possible._

### DIFF
--- a/spec-dtslint/operators/race-spec.ts
+++ b/spec-dtslint/operators/race-spec.ts
@@ -15,22 +15,6 @@ it('should allow an array of observables', () => {
   const p = of('a', 'b', 'c').pipe(race([of('x', 'y', 'z'), of('t', 'i', 'm')])); // $ExpectType Observable<string>
 });
 
-it('should be possible to provide a return type', () => {
-  const o = of('a', 'b', 'c').pipe(race<string, number>([of(1, 2, 3)])); // $ExpectType Observable<number>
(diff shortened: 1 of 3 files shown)


spec-dtslint/operators/race-spec.ts
```diff
@@ -1,36 +1,20 @@
...
+});
+
+it('should allow observables', () => {
+  const o = of('a', 'b', 'c').pipe(race(of('x', 'y', 'z'))); // $ExpectType Observable<string>
+  const p = of('a', 'b', 'c').pipe(race(of('x', 'y', 'z'), of('t', 'i', 'm'))); // $ExpectType Observable<string>
+});
+
+it('should allow an array of observables', () => {
+  const o = of('a', 'b', 'c').pipe(race([of('x', 'y', 'z')])); // $ExpectType Observable<string>
+  const p = of('a', 'b', 'c').pipe(race([of('x', 'y', 'z'), of('t', 'i', 'm')])); // $ExpectType Observable<string>
```
benlesh, line 15: LOL... who is Tim?

benlesh (approved):

PR #6116, opened 2021-03-12
fix(Subscriber): don't leak destination

cartant:
**Description:**

This PR adds a failing test that shows that subscriptions - i.e. subscribers - returned from `subscribe` leak their destinations - i.e. their observers or callbacks.

The PR changes `Subscriber#unsubscribe` to null the `destination` after `super.unsubscribe` is called.

This change broke three operators - `bufferWhen` was one of them - because if unsubscription was effected within `onNext` and if an error was thrown from within `onNext` (after the unsubscription) it'd be caught in the `OperatorSubscriber` and an attempt would be made to call `this.destination.error`, but `this.destination` would be `null`.

Rather than capture `this.destination` before the `try/catch` around `onNext`, `onComplete`, and `onError`, I just used the `destination` that was passed in to the `OperatorSubscriber` constructor - AFAICT, we can be sure that it's a `Subscriber`, because `OperatorSubscriber` is an internal, impementation detail.

**Related issue (if exists):** Nope, but there is a discussion: https://github.com/ReactiveX/rxjs/discussions/6115

### DIFF
--- a/spec/Subscriber-spec.ts
+++ b/spec/Subscriber-spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
 import { SafeSubscriber } from 'rxjs/internal/Subscriber';
-import { Subscriber, Observable, config, of } from 'rxjs';
+import { Subscriber, Observable, config, of, Observer } from 'rxjs';
 import { asInteropSubscriber } from './helpers/interop-helper';
 import { getRegisteredTeardowns } from './helpers/subscription';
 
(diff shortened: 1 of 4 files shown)


benlesh (commented):
I *think* this is okay... I'd just want to see if this works okay when you chain together a few operators. We altered `OperatorSubscriber`, but there's no new tests for operators related to that, so it has me a little concerned.

cartant:
> but there's no new tests for operators related to that

I'm not sure what additional tests you might be looking for. There are no explicit tests for `OperatorSubscriber`, AFAICT; it's tested via the operator tests themselves. And regarding operators being chained, pretty much every operator has a don't-break-subscription-chains test with something like this in it - i.e. several operators chained together:

https://github.com/ReactiveX/rxjs/blob/3ab3f6fe22816da7be25402f94ddcd7207cd4bb0/spec/operators/mergeMap-spec.ts#L307-L311

Regarding the `OperatorSubscriber` change, these are the tests that _fail_ without it. And they all fail because the subscribers to the selector-returned observables _unsubscribe themselves_ and if a error is thrown from within the selector and is caught in `OperatorSubscriber`, `this.destination` will be `null`, etc.:

https://github.com/ReactiveX/rxjs/blob/3ab3f6fe22816da7be25402f94ddcd7207cd4bb0/spec/operators/bufferWhen-spec.ts#L184-L211

https://github.com/ReactiveX/rxjs/blob/3ab3f6fe22816da7be25402f94ddcd7207cd4bb0/spec/operators/zipAll-spec.ts#L345-L365

https://github.com/ReactiveX/rxjs/blob/3ab3f6fe22816da7be25402f94ddcd7207cd4bb0/spec/operators/zipAll-spec.ts#L222-L240

So I'm not sure what else needs to be tested.

benlesh:
That makes sense, thanks @cartant

benlesh (approved):

PR #6118, opened 2021-03-12
fix: unhandled errors in observers correctly scheduled

benlesh:
- Resolves an issue where errors thrown in a handler provided as a plain function, or as part of a POJO observer, to a `subscribe` call were not being handled appropriately. They will no longer synchronously throw unless `useDeprecatedSynchronousErrorHandling` is configured to `true`.
- Removes two tests that were poorly written, and were not passing (and should never have been passing). The ground they covered is covered adequately by other tests.

### DIFF
--- a/spec/operators/throttle-spec.ts
+++ b/spec/operators/throttle-spec.ts
@@ -381,62 +381,6 @@ describe('throttle', () => {
     });
   });
 
-  it('should throttle by promise resolves', () => {
-    testScheduler.run(() => {
-      const e1 = concat(of(1), timer(10).pipe(mapTo(2)), timer(10).pipe(mapTo(3)), timer(50).pipe(mapTo(4)));
-      const expected = [1, 2, 3, 4];
-
-      e1.pipe(
-        throttle(() => {
-          return new Promise((resolve: any) => {
-            resolve(42);
-          });
-        })
-      ).subscribe(
-        (x: number) => {
-          expect(x).to.equal(expected.shift());
-        },
-        () => {
-          throw new Error('should not be called');
-        },
-        () => {
-          expect(expected.length).to.equal(0);
-        }
-      );
-    });
-  });
-
-  it('should raise error when promise rejects', () => {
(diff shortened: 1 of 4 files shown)


spec/operators/throttle-spec.ts
```diff
@@ -381,62 +381,6 @@ describe('throttle', () => {
     });
   });
 
-  it('should throttle by promise resolves', () => {
```
benlesh, line 384: After the changes, this test started failing. Reading the test, it should have *never* passed (as it's mixing a synchronous `TestScheduler` run with `Promise` resolution). Handling of promises (via converting `ObservableInput`s) is well-tested elsewhere, and we can just remove this test, and the test after it, as they are poorly written.

It's a bit more concerning to me that this test was passing before, and I'm unsure what made it finally behave correctly.

src/internal/Subscriber.ts
```diff
@@ -167,9 +167,9 @@ export class SafeSubscriber<T> extends Subscriber<T> {
     // Once we set the destination, the superclass `Subscriber` will
     // do it's magic in the `_next`, `_error`, and `_complete` methods.
     this.destination = {
```
benlesh, line 169: Ultimately, the issue is from https://github.com/ReactiveX/rxjs/commit/1b07686a1f79bb5b994ea85795fe3af868f07a4b,
When that was refactored, we removed the pieces that caught errors in user-provided handler functions, and then handled them appropriately.

cartant (approved):
LGTM. I have no idea how that test passed either.

spec/operators/throttle-spec.ts
cartant: @benlesh I had a quick look and this and for your peace of mind:

-   The test - as you mentioned - has an entirely synchronous source and test's expectation fails and throws.
-   The failures occurs in the call to the `complete` callback passed to `subscribe`, but the error bubbles all of the way back to the `throttle` operator's `onComplete` callback - which has a `try`/`catch` around it.
-   The `try`/`catch` around `onComplete` catches the error.
-   However, the `destination` is stopped, so the error is swallowed - well, the notification will be passed to the `onStoppedNotification` handler if there is one.

In the new implementation, the error is caught in the `SafeSubscriber` and it's passed to the `onUnhandledError`, instead.

Issue #4921, opened 2019-07-23
[docs]: Suggestions for `delay` documentation improvements


BioPhoton:
### Documentation Related To Component:

The `delay` operator: 
[delay.ts](https://github.com/ReactiveX/rxjs/blob/b24d3ea85d33b893e4c020fb4fa9f6fa4d05dc0b/src/internal/operators/delay.ts)

### Please check those that apply

- [ ] typo
- [] documentation doesn't exist
- [X] documentation needs clarification
- [ ] error(s)/adoption/updates in example
- [] needs example

### Description Of The Issue

The "Seel also" links are inconsistent in including similar operators.
We might want to improve this to serve a better more detailed description for the community.

#### See Also
We refer to `delayWhen` and `debounceTime`. `delay` is just delaying a value while `debounceTime` is swallowing values. Here I would suggest to include eighter all or on operator from the rate-limiting group.

A list including also all ratelimiting operators could look like this:
```typescript
 * @see {@link delayWhen}
 * @see {@link throttle}
 * @see {@link throttleTime}
 * @see {@link debounce}
 * @see {@link debouceTime}
 * @see {@link sample}
 * @see {@link sampleTime}
 * @see {@link audit}
 * @see {@link auditTime}
```

A list including none of the rate-limiting operators could look like this:
```typescript
 * @see {@link delayWhen}
```

PR #6123, opened 2021-03-13
Delay docs

Armenvardanyan95:
A minor doc change proposed by @BioPhoton in #4921 

(Also automatic linter fixes from the post-commit hook)

Related issue: https://github.com/ReactiveX/rxjs/issues/4921

### DIFF
--- a/src/internal/operators/delay.ts
+++ b/src/internal/operators/delay.ts
@@ -41,8 +41,15 @@ import { timer } from '../observable/timer';
  * delayedClicks.subscribe(x => console.log(x));
  * ```
  *
- * @see {@link debounceTime}
  * @see {@link delayWhen}
+ * @see {@link throttle}
+ * @see {@link throttleTime}
+ * @see {@link debounce}
+ * @see {@link debounceTime}
+ * @see {@link sample}
+ * @see {@link sampleTime}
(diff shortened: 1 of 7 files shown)


src/internal/operators/delay.ts
```diff
@@ -41,8 +41,15 @@ import { timer } from '../observable/timer';
  * delayedClicks.subscribe(x => console.log(x));
  * ```
  *
- * @see {@link debounceTime}
  * @see {@link delayWhen}
+ * @see {@link throttle}
```
niklas-wortmann, line 45: there are a lot of formatation changes  in this pr that are entirely unrelated to this, I'd appreciate it if you could revert those

Armenvardanyan95: The post-commit hook does that for me automatically, I did not want to interfere. Is there a way I can disable it? I ill revert those

Armenvardanyan95: It got a bit messy, sorry for that. I will submit another PR shortly

Armenvardanyan95: Here is the new PR, containing only the documentation change: https://github.com/ReactiveX/rxjs/pull/6156

Issue #5418, opened 2020-05-08
[docs] flatMap page almost completely refers to mergeMap


zrev2220:
### Documentation Related To Component:

[flatMap](https://rxjs-dev.firebaseapp.com/api/operators/flatMap), [mergeMap](https://rxjs-dev.firebaseapp.com/api/operators/mergeMap)

### Please check those that apply

- [ ] typo
- [ ] documentation doesn't exist
- [ ] documentation needs clarification
- [X] error(s) in example
- [ ] needs example

### Description Of The Issue

The documentation page for the `flatMap` operator is almost 100% identical to the page for `mergeMap`—the only differences I can tell are the **title** and the **function signature name** at the top of the page (on `flatMap`'s page it correctly says "flatMap", and on the `mergeMap` page they say "mergeMap").

Besides this, the entire page for `flatMap` consistently refers to `mergeMap`, including in the:
* [Marble diagram](https://rxjs-dev.firebaseapp.com/api/operators/flatMap#description)
* [Code example](https://rxjs-dev.firebaseapp.com/api/operators/flatMap#example)
* [Overload signatures](https://rxjs-dev.firebaseapp.com/api/operators/flatMap#overloads)

The [See Also](https://rxjs-dev.firebaseapp.com/api/operators/flatMap#see-also) section is even identical.

It seems the docs for `mergeMap` were copied with the intent to edit to describe `flatMap`, but very little actually got changed.

cartant:
FWIW, `flatMap` is an alias for `mergeMap`: https://github.com/ReactiveX/rxjs/blob/6.5.5/src/internal/operators/index.ts#L46

The docs are generated from the TSDoc comments and there are none for `flatMap`.

robations:
Related: in my IDE it flags the `flatMap` operator as deprecated, but in the docs https://rxjs.dev/api/operators/flatMap (assuming they are up to date) it makes no mention of this, which is confusing. Maybe aliases and deprecated functions should be flagged more clearly in online help?

This doesn't seem to make its way into the documentation:

```typescript
/**
 * @deprecated renamed. Use {@link mergeMap}
 */
export const flatMap = mergeMap;
```

cartant:
Closing this, as the v7 docs now show:

> Renamed to `mergeMap`. Will be removed in v8.

PR #6124, opened 2021-03-13
chore: undeprecate flatMap

cartant:
**Description:**

Does what it says in the title. Given that VS Code now uses the strikethrough text style for deprecated APIs, IMO we should err on the side of not deprecating unless we have a plan for removing something. Devs have enough to deal with without having to wonder why things are struck out.

As Ben pointed out on Twitter, according to GitHub, the RxJS use count is over 5 million. And with `flatMap` being such an established term for that particular feature, I don't see why we'd want to remove the alias and break people.

IMO: specify a version in which it will be removed or undeprecate it. Clearly, my vote is for the latter. (We can always deprecate it again, when we have a plan - maybe with code mods - for its removal.)

**Related issue (if exists):** https://github.com/ReactiveX/rxjs/issues/5418#issuecomment-719735322

### DIFF
--- a/src/internal/operators/flatMap.ts
+++ b/src/internal/operators/flatMap.ts
@@ -1,6 +1,3 @@
 import { mergeMap } from './mergeMap';
 
-/**
- * @deprecated renamed. Use {@link mergeMap}.
- */
 export const flatMap = mergeMap;


cartant:
Some more observations (🙂) regarding this deprecation:

- It doesn't actually do anything. Deprecating (some) result selectors has efficiency gains. Deprecating multi-arg `subscribe` signatures will allow for an `AbortSignal` parameter. Deprecating operators that have the same name as a creator means the `operators` import location could eventually be removed. Deprecating/removing the `flatMap` alias does nothing like this.
- It might avoid people being confused - "what does `flatMap` do" - but it's also likely that just as many people will say: "huh? RxJS has no `flatMap`? wut?"

josepot:
I'm torn on this one. I understand why @cartant wants to undeprecate it on v7, but these are some other things to take into account:

- I _think_ that RxJS is the only ReactiveX implementation where `flatMap` is just an alias to another operator.
- There is already one ReactiveX implementation that doesn't have `flatMap`: RxNET.
- As a developer, when I upgrade one of my dependencies to its next major version: I do understand that some things will "break", and that some things will be flagged as deprecated. What's important for me is that the library provides a clear migration path and/or tools to make the migration easier. In that sense, if the deprecation warning ends up staying, I volunteer for creating a codemod for this (`flatMap` -> `mergeMap`), if it doesn't already exist :sweat_smile:.
- I understand @cartant 's point when he says that "It doesn't actually do anything". However, I do think that it does: IMO it makes learning the API less confusing for new-comers. It's true that someone who is already familiar with other ReactiveX implementations could find this confusing, but I worry less about those developers, because they are already familiar with the core concepts.

benlesh:
So one of the reasons I have it deprecated is the advent of `Array.prototype.flatMap`, which, to the layperson looks like our `concatMap`, not our `mergeMap`. And the truth is, `mergeMap` is probably the least usable of all of our flattening operators: `concatMap`, `exhaustMap`, `switchMap`, IMO, are all generally more useful in normal application development. Yet we have `mergeMap` holding a psuedo-title as _the_ flattening operator "`flatMap`". 

We've had it aliased for years now, and I'm not sure it's brought a lot of value. Other than initially, as existing RxJS <4 users came onboard. 

I think the point is... *someday* we're going to need to deprecate it and remove it. I don't see a great reason to leave it there indefinitely.

cartant:
> And the truth is, `mergeMap` is probably the least usable of all of our flattening operators: `concatMap`, `exhaustMap`, `switchMap`, IMO, are all generally more useful in normal application development. Yet we have `mergeMap` holding a psuedo-title as _the_ flattening operator "`flatMap`".

Yeah, that's a good point. I'm going to close this. I think the bigger issue is that all of the deprecations need to have a docs page - which is what I'll be focusing on next - and a nominated will-be-removed-in-this-version. I'm also going to remove the label.

PR #6125, opened 2021-03-13
refactor: align tap sigs with subscribe

cartant:
**Description:**

Does what it says in the title.

### DIFF
--- a/src/internal/operators/tap.ts
+++ b/src/internal/operators/tap.ts
@@ -4,7 +4,6 @@ import { operate } from '../util/lift';
 import { OperatorSubscriber } from './OperatorSubscriber';
 import { identity } from '../util/identity';
 
-/* tslint:disable:max-line-length */
 export function tap<T>(observer?: Partial<Observer<T>>): MonoTypeOperatorFunction<T>;
 /** @deprecated Use an observer instead of a complete callback, Details: https://rxjs.dev/deprecations/subscribe-arguments */
 export function tap<T>(next: null | undefined, error: null | undefined, complete: () => void): MonoTypeOperatorFunction<T>;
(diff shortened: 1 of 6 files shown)


src/internal/operators/tap.ts
```diff
@@ -1,19 +1,19 @@
-import { MonoTypeOperatorFunction, PartialObserver } from '../types';
+import { MonoTypeOperatorFunction, Observer } from '../types';
 import { isFunction } from '../util/isFunction';
 import { operate } from '../util/lift';
 import { OperatorSubscriber } from './OperatorSubscriber';
 import { identity } from '../util/identity';
 
 /* tslint:disable:max-line-length */
```
benlesh, line 7: probably don't need these `tslint` comments anymore, I guess.

src/internal/operators/tap.ts
```diff
@@ -1,19 +1,19 @@
...
 /** @deprecated Use an observer instead of a complete callback, Details: https://rxjs.dev/deprecations/subscribe-arguments */
 export function tap<T>(next: null | undefined, error: null | undefined, complete: () => void): MonoTypeOperatorFunction<T>;
 /** @deprecated Use an observer instead of an error callback, Details: https://rxjs.dev/deprecations/subscribe-arguments */
 export function tap<T>(next: null | undefined, error: (error: any) => void, complete?: () => void): MonoTypeOperatorFunction<T>;
 /** @deprecated Use an observer instead of a complete callback, Details: https://rxjs.dev/deprecations/subscribe-arguments */
 export function tap<T>(next: (value: T) => void, error: null | undefined, complete: () => void): MonoTypeOperatorFunction<T>;
-export function tap<T>(next?: (x: T) => void, error?: (e: any) => void, complete?: () => void): MonoTypeOperatorFunction<T>;
-export function tap<T>(observer: PartialObserver<T>): MonoTypeOperatorFunction<T>;
-/* tslint:enable:max-line-length */
+export function tap<T>(next?: (value: T) => void, error?: (error: any) => void, complete?: () => void): MonoTypeOperatorFunction<T>;
```
benlesh, line 15: This signature should be here, but it should also be deprecated, because we're trying to move towards `Observer | fn`

benlesh (changes requested):

src/internal/operators/tap.ts
cartant: Done.

src/internal/operators/tap.ts
cartant: Done. Simplified the signatures for `subscribe` and `tap` so that only a single, deprecated multi-arg signature is needed for each. And added dtslint tests for the deprecations.

benlesh (approved):

benlesh:
Thanks (for everything) @cartant!

PR #6129, opened 2021-03-14
chore: remove a too-wide signature from combineLatestAll

cartant:
**Description:**

This PR removes a signature that will match `any`-thing as long as a result selector is specified - and the values received by the selector are `any`. This shenanigans has been removed from other places, so we should remove it here, too.


cartant:
Nope. Has to stay. I'll add a dtslint test in another PR to show why.

PR #6137, opened 2021-03-14
Zip legacy formatting

tmair:
**Description:**
This PR introduces prettier formatting in `zip` Legacy tests. One use of the deprecated subscribe signature is replaced and anonymous functions are converted to arrow functions.

None


cartant (approved):
LGTM. Thanks for splitting the changes into separate commits. Very easy to review.

Issue #6144, opened 2021-03-16
[v7] share throws an exception when using observables with circular references


voliva:
## Bug Report

**Current Behavior**
When you have an observable with circular references, something like:

```ts
const deferredCount = defer(() => source).pipe(
   startWith(0)
);

const source: Observable<string> = from([1, 2, 3]).pipe(
  withLatestFrom(deferredCount),
  map(([a, b]) => String(Number(a) + Number(b))),
  share()
);
```

(Sorry for the contrived example, this could be easily done with a scan, but shows the essential shape of a circular reference in observables)

Share will throw an error. I've seen two different errors depending on the environment I'm running this test, but both of them look similar. When testing with rxjs-marbles:
```
    TypeError: Cannot read property 'complete' of null

      at Object.complete (node_modules/rxjs/src/internal/operators/share.ts:138:10)
      at Object.complete (node_modules/rxjs/src/internal/Subscriber.ts:188:14)
      at SafeSubscriber.Object.<anonymous>.Subscriber._complete (node_modules/rxjs/src/internal/Subscriber.ts:126:24)
      at SafeSubscriber.Object.<anonymous>.Subscriber.complete (node_modules/rxjs/src/internal/Subscriber.ts:100:12)
      at OperatorSubscriber.Object.<anonymous>.Subscriber._complete (node_modules/rxjs/src/internal/Subscriber.ts:126:24)
      at OperatorSubscriber.Object.<anonymous>.Subscriber.complete (node_modules/rxjs/src/internal/Subscriber.ts:100:12)
      at OperatorSubscriber.Object.<anonymous>.Subscriber._complete (node_modules/rxjs/src/internal/Subscriber.ts:126:24)
      at OperatorSubscriber.Object.<anonymous>.Subscriber.complete (node_modules/rxjs/src/internal/Subscriber.ts:100:12)
      at Object.observeNotification (node_modules/rxjs/src/internal/Notification.ts:238:102)
      at VirtualAction.subscriber.add.scheduler.schedule.message (node_modules/rxjs/src/internal/testing/ColdObservable.ts:43:13)
      at VirtualAction.Object.<anonymous>.AsyncAction._execute (node_modules/rxjs/src/internal/scheduler/AsyncAction.ts:116:12)
      at VirtualAction.Object.<anonymous>.VirtualAction._execute (node_modules/rxjs/src/internal/scheduler/VirtualTimeScheduler.ts:111:28)
[...]

PR #6151, opened 2021-03-17
fix(share): handle reentrant subscribes

cartant:
**Description:**

This PR replaces the observer-based subscription that's made within the `share` implementation with a `Subcriber`-based subscription - so that the `connection` variable is assigned _before_ the `subscribe` call is made - to support reentrant calls to `subscribe`.

**Related issue (if exists):** #6144

### DIFF
--- a/src/internal/operators/share.ts
+++ b/src/internal/operators/share.ts
@@ -1,6 +1,6 @@
 import { Subject } from '../Subject';
 import { MonoTypeOperatorFunction, OperatorFunction, SubjectLike } from '../types';
-import { Subscriber } from '../Subscriber';
+import { SafeSubscriber } from '../Subscriber';
 import { from } from '../observable/from';
 import { operate } from '../util/lift';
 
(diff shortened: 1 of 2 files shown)


src/internal/operators/share.ts
```diff
@@ -118,9 +117,14 @@ export function share<T>(options?: ShareConfig<T>): OperatorFunction<T, T> {
...
 
     if (!connection) {
-      connection = from(source).subscribe({
-        next: (value) => subject!.next(value),
-        error: (err) => {
+      // We need to create a subscriber here - rather than pass an observer and
+      // assign the returned subscription to connection - because it's possible
+      // for reentrant subscriptions to the shared observable to occur and in
+      // those situations we don't want connection to be already-assigned so
+      // that we don't create another connection to the source.
```
backbone87, line 124: > in those situations we don't want connection to be already-assigned so that we don't create another connection to the source.

dont you mean this?
"in those situations we ~~don't~~ want connection to be already-assigned so that we don't create another connection to the source."

benlesh (approved):

Issue #5185, opened 2019-12-18
[Docs]: add best practice guide


niklas-wortmann:
In the last core meeting, we came up, that we want to have an official recommendation within the official docs. It should include:

 - Don't sub-class
 - handling subscribe + unsubscribe
 - mention linting rules
 - ...

 I'm very open to any kind of suggestion, so please let me know about anything you could think of.

thekiba:
@JWO719 is wonderful!

I suggest adding the article https://medium.com/p/92502d5639d0 to explain handling subscribe-unsubscribe

BioPhoton:
The section should clarify what .subscribe does. It ends composability. 

Subscribe starts a process and/or is here to perform a side effect.

If people want to compose the values further they should not subscribe.

Even if it a very essential basic practice shows this is not clear.

BioPhoton:
Error handling should also be implemented

thekiba:
Build own operators (i know that [rxjs docs already have it](https://rxjs-dev.firebaseapp.com/guide/v6/pipeable-operators#build-your-own-operators-easily) but that could be a good point for best practice guide)

The general rule is for takeUntil to be placed last ([Avoiding takeUntil Leaks](https://medium.com/p/fb5182d047ef))

stefanolepera:
Cover testing as well please, the marble tests can be tricky sometimes. I'd like to see more examples of more complex scenarios.

dzhavat:
I think @cartant 's [rxjs-tslint-rules](https://github.com/cartant/rxjs-tslint-rules) should be mentioned as well.

lamisChebbi:
Hi @JWO719 
I think it should include also performance tips and common mistakes that can lead to memory leaks.
Thanks

Armenvardanyan95:
As far as I know there is no official best practices page in the docs, and I would love to create one, using both my experience with the library (I've written several articles on both good approaches and incorporating RxJS in Angular, see: [this](https://indepth.dev/posts/1292/rxjs-the-unexpected), [this](https://medium.com/angular-in-depth/rx-js-best-practices-6a3b095ffb04) and [this one](https://medium.com/angular-in-depth/rx-js-best-practices-6a3b095ffb04)) and the advice shared by contributors in this thread. Is there anything that could block me? What is a good starting point? Any insight on how that page will be located and what it could look like, also, how much creative freedom can I have if we choose to go forward with it?

lamisChebbi:
[...]

PR #6157, opened 2021-03-19
docs: add a best practices document

Armenvardanyan95:
This is an initial version of a document describing best practices and antipatterns to avoid when using RxJS. This is a more finalized version of my draft, I tried to bring together important points mentioned by everyone in the related issue thread (#5185). Will be glad to hear how this can be improved

### DIFF
--- a/docs_app/content/guide/best-practices.md
+++ b/docs_app/content/guide/best-practices.md
@@ -1,0 +1,100 @@
+### RxJS best practices
+
+RxJS is very large and contains lots of functions, operators and other stuff which makes it a complex ecosystem, but for lots of applications it can be as simple as knowing what `Observables` are and what can be done with them with some `operators`. But for larger codebases it is important to have code written in a way that allows for reusability, readability and composability. This guide may help overcome common ba practices and adopt a reactive mindset necessary for righting concise RxJS code.
+
+## .subscribe ends composability
+When we call `.subscribe` on any `Observable` what we do is get rid of said `Observable` and then only focus on the values that it produces. When we call `.subscribe` we can no longer continue composing operators (`.subscribe` returns a `Subscription`, on which we can no longer call `.pipe` and use RxJS operators), and composability is **the** greatest feature of RxJS.  Obviously, without subscribing to most `Observables`, we won't have a working application. In practice, frameworks like Angular offer solutions that allow to use `Observable` values directly without explicitly subscribing to them, a great example is the `async` pipe. 
+So while subscribing is obviously not directly discouraged, we should be generally very careful when doing it, as to understand that the stream in question is no longer to be modified and can only be used as a source of values, nothing more.
+
+## Using imperative logic in .subscribe instead of operators
+
+Another problem that developers face when using `.subscribe` is performing complex operations on values from a source `Observable`. Especially bad scenarios involve using conditional operators and loops inside `.subscribe` callbacks, instead of composing `Observables` using pipes. Here are some common mistakes:
+
+### Using an if statement when subscribing
+
+In this piece of code, an operations is performed on a value emitted from a source observable if some condition is true:
+
+    of(1, 2, 3, 4, 5).subscribe(
+        value => {
+            if (value % 2 === 0) {
+                doSomething(value);
+            }
+        }
+    );
+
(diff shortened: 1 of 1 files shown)


Armenvardanyan95:
Important questions to be addressed:

- Do we need any more sections?
- Is the reason why subclassing is bad explained enough?
- Are the code examples okay?
- I did not update the navigation.json file, as I did not really know where should this go, if anyone directs me, I will also update that

WIll happily listen to everyone's thoughts

docs_app/content/guide/best-practices.md
```diff
@@ -0,0 +1,100 @@
...
+
+    interval(1000).pipe(
+        takeUntil(notifier$),
+    ).subscribe(console.log)
+
+In this example just looking at the `Observable` is enough to know exactly when it will terminate: when the `notifier$` `Observable` emits. Also, this is the widely accepted approach in the Angular community.
+
+### Be careful to avoid takeUntil leaks
+
+When `takeUntil` is not used correctly, it can lead to memory leaks too. As a rule of thumb, always place `takeUntil` as the last operator in the `pipe`.
```
niklas-wortmann, line 91: can you elaborate on this a little bit to make it easier to understand for newcomers

docs_app/content/guide/best-practices.md
```diff
@@ -0,0 +1,100 @@
...
+
+In this example just looking at the `Observable` is enough to know exactly when it will terminate: when the `notifier$` `Observable` emits. Also, this is the widely accepted approach in the Angular community.
+
+### Be careful to avoid takeUntil leaks
+
+When `takeUntil` is not used correctly, it can lead to memory leaks too. As a rule of thumb, always place `takeUntil` as the last operator in the `pipe`. 
+
+## Don't subclass
+
+Sometimes, it can be tempting to build our own version of an `Observable` using existing classes. In particular, some developers might extend the existing `Observable` class to add some functionality and then use the "better" version throughout the app. But problems can arise in the future, for example, if the base classes are updated with the new versions of the RxJS library, as those classes are tightly coupled. Also, implementing a bae class that can work requires the developer to implement the `lift` method, which in turn makes the consumer application aware of the internal implementation details of the library. This [thread](https://github.com/ReactiveX/rxjs/issues/5431) contains a detailed discussion of the matter.
```
niklas-wortmann, line 95: there's a typo in `bae`

niklas-wortmann (changes requested):
Hi sorry I wasn't able to review this earlier. First of all thanks for taking the time for this contribution, I do think that this is a gap in the docs we have to fill. From my point of view your suggestions have a couple personal opinions, that I don't necessarily share and I am not sure how we should incorporate those properly in the docs. Maybe other core team members have an opinion as well, @cartant, @benlesh, @kwonoj, @kolodny

cartant:
> Maybe other core team members have an opinion as well

I don't think this could be merged without discussion. I don't agree with everything in this PR and, FWIW, I'm reluctant to declare a list of official best practices.

CMCDragonkai:
Regarding the don't subclass section. Can you provide an alternative to this kind of usecase?

Suppose you had a domain class like `KeyRing`. The `KeyRing` is a class with properties and methods. You want the properties to be observable, and that's easy enough with third party observable decorators, but what if want the whole `KeyRing` object to be observable too?

Would that mean some sort of `BehaviourSubject` for the `KeyRing` object, so that downstream consumers can subscribe to the entire `KeyRing` and should get the pointer to the `KeyRing` object, and access all the changed properties atm.

One issue I'm thinking is that the `KeyRing` is a singleton object, it's not like new `KeyRing` is created on each change. That could meant it doesn't make sense for it to be a reactive class itself, but I'm not sure. There's not much information on whether it makes sense to subclass Observable with your own custom class, in which you want the whole domain object itself to be observable.

benlesh:
Hi @Armenvardanyan95 — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

PR #6163, opened 2021-03-21
feat(ReadableStreams): Support for ReadableStreams e.g. `from(readableStream)`

### LINKED ISSUE #6006: Feat: Support ReadableStream as an ObservableInput
Although RxJS gets a LOT of use in Node, RxJS's primary space has been in the browser. Based off of discussion [in this Twitter thread about fetch streaming](https://twitter.com/BenLesh/status/1358874655076077569?s=20)

We might want to support duck-typing `ReadableStream<T>` as an `ObservableInput<T>`.

If we did this, it might be beneficial, especially in the `fetch` API. But it's probably worth investigating what all use cases might exist.

I definitely think if it's added to `from`, `schedule`, and internally, `innerFrom`, it would need to be the absolute _last_ thing checked. The presence of `asyncIterator` or `iterator` should definitely beat out a check for `typeof o.getReader === 'function'`. 

This would be a lightweight and powerful add, IMO.

@jayphelps generously volunteered to do the work.

jayphelps:
**Description:**
Support ReadableStream and other similar objects by checking for `getReader()` existence. In theory this isn't foolproof, but in practice I can't see this ever actually causing issues in practice. `instanceof` checks don't work with cross-Realm instances and neither ReadableStream nor ReadableStreamDefaultReader implement Symbol.asyncIterator yet (which is why this custom support is needed)

Closes #6006

### DIFF
--- a/spec/observables/from-spec.ts
+++ b/spec/observables/from-spec.ts
@@ -1,7 +1,8 @@
 import { expect } from 'chai';
 import { TestScheduler } from 'rxjs/testing';
 import { asyncScheduler, of, from, Observer, observable, Subject, noop } from 'rxjs';
-import { first, concatMap, delay, take, map, tap } from 'rxjs/operators';
+import { first, concatMap, delay, take, tap } from 'rxjs/operators';
+import { ReadableStream } from 'web-streams-polyfill';
 
 // tslint:disable:no-any
 declare const expectObservable: any;
@@ -108,7 +109,7 @@ describe('from', () => {
     });
   });
 
-  
+
   it('should finalize a generator', () => {
     const results: any[] = [];
 
@@ -175,16 +176,24 @@ describe('from', () => {
   });
 
   // tslint:disable-next-line:no-any it's silly to define all of these types.
-  const sources: Array<{ name: string, value: any }> = [
-    { name: 'observable', value: of('x') },
-    { name: 'observable-like', value: fakervable('x') },
-    { name: 'observable-like-array', value: fakeArrayObservable('x') },
-    { name: 'array', value: ['x'] },
-    { name: 'promise', value: Promise.resolve('x') },
-    { name: 'iterator', value: fakerator('x') },
-    { name: 'array-like', value: { [0]: 'x', length: 1 }},
-    { name: 'string', value: 'x'},
-    { name: 'arguments', value: getArguments('x') },
+  const sources: Array<{ name: string, createValue: () => any }> = [
+    { name: 'observable', createValue: () => of('x') },
+    { name: 'observable-like', createValue: () => fakervable('x') },
+    { name: 'observable-like-array', createValue: () => fakeArrayObservable('x') },
+    { name: 'array', createValue: () => ['x'] },
+    { name: 'promise', createValue: () => Promise.resolve('x') },
+    { name: 'iterator', createValue: () => fakerator('x') },
+    { name: 'array-like', createValue: () => ({ [0]: 'x', length: 1 }) },
+    // ReadableStreams are not lazy, so we have to have this createValue() thunk
+    // so that each tests gets a new one.
+    { name: 'readable-stream-like', createValue: () => new ReadableStream({
+      pull(controller) {
+        controller.enqueue('x');
+        controller.close();
+      },
+    })},
+    { name: 'string', createValue: () => 'x'},
+    { name: 'arguments', createValue: () => getArguments('x') },
   ];
 
   if (Symbol && Symbol.asyncIterator) {
@@ -211,14 +220,14 @@ describe('from', () => {
 
     sources.push({
       name: 'async-iterator',
-      value: fakeAsyncIterator('x')
+      createValue: () => fakeAsyncIterator('x')
     });
   }
 
   for (const source of sources) {
     it(`should accept ${source.name}`, (done) => {
       let nextInvoked = false;
-      from(source.value)
+      from(source.createValue())
         .subscribe(
           (x) => {
             nextInvoked = true;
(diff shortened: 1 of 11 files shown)


spec/observables/from-spec.ts
```diff
@@ -108,7 +109,7 @@ describe('from', () => {
     });
   });
 
-
```
jayphelps, line 111: vscode removed this trailing whitespace on save.

src/internal/types.ts
```diff
@@ -91,7 +92,14 @@ export interface Subscribable<T> {
...
  */
-export type ObservableInput<T> = Observable<T> | InteropObservable<T> | AsyncIterable<T> | PromiseLike<T> | ArrayLike<T> | Iterable<T>;
+export type ObservableInput<T> =
+  | Observable<T>
+  | InteropObservable<T>
+  | AsyncIterable<T>
+  | PromiseLike<T>
+  | ArrayLike<T>
+  | Iterable<T>
+  | ReadableStreamLike<T>;
```
jayphelps, line 102: I'm happy to change this and the other places to have a consistent order, if there's a preference.

src/internal/util/isReadableStreamLike.ts
```diff
@@ -0,0 +1,24 @@
+import { isFunction } from './isFunction';
+
+export interface ReadableStreamLike<T> {
+  getReader(): ReadableStreamDefaultReader<T>;
```
cartant, line 4: I'll give this a proper review later, but I don't think we can rely on - or explicitly reference using a `///` comment - this type: `ReadableStreamDefaultReader<T>`

jayphelps: Ah great catch. I'll stub it out too since it's easy.

jayphelps: Done.

src/internal/types.ts
benlesh: I guess the order of this should match the runtime order.

src/internal/util/isReadableStreamLike.ts
```diff
@@ -0,0 +1,34 @@
...
+  read(): IteratorResult<T>;
+  releaseLock(): void;
+}
+
+// This is the only part that matters to us.
+export interface ReadableStreamLike<T> {
+  getReader(): ReadableStreamDefaultReaderLike<T>;
+}
+
+export async function* readableStreamLikeToAsyncGenerator<T>(readableStream: ReadableStreamLike<T>): AsyncGenerator<T> {
```
benlesh, line 15: Just thinking... that while this is _definitely_ the coolest way. The lame way might be more efficient, something like:

```ts
export function readableStreamLikeToAsyncGenerator(readableStream: ReadableStreamLike<T>): AsyncGenerator<T> {
  const reader = readableStream.getReader();
  
  return {
     [Symbol.asyncIterator]() { return this; },
     async next() { 
       return reader.read();
     },
     async throw(err) {
       reader.releaseLock();
       throw err;
     },
     async return() {
       reader.releaseLock();
       return { done: true };
     }
  }
}
```

But maybe I'm totally wrong. Just something I was thinking about. I guess if there was an error during the execution of `next` above, the lock wouldn't get released unless you had a try/catch in there, so maybe this isn't the best idea.

benlesh:
General approval from me. Pending @cartant's approval.

cartant (approved):
LGTM

cartant:
I had a look at adding a dtslint test for a `ReadableStreamLike` source to `spec-dtslint/observables/from-spec.ts` and found some things we might want to address in this PR.

IMO, we should export `ReadableStreamLike` - primarily because it's unused in the union type `ObservableInput` which is exported.

In the dtslint test I had to import it like this:

```ts
import { from, of, animationFrameScheduler } from 'rxjs';
import { ReadableStream } from 'web-streams-polyfill';
import { ReadableStreamLike } from '../../src/internal/util/isReadableStreamLike';
```

In the test, `ReadableStream` seemed to be match the `AsyncIterable` type in the `ObservableInput` and the inference wasn't what I expected:

```ts
it('should accept a ReadableStream', () => {
  const stream = new ReadableStream<string>({
    pull(controller) {
      controller.enqueue('x');
      controller.close();
    },
  });
  // Nope, this expectation fails 'cause Observable<any> is inferred
  const o = from(stream); // $ExpectType Observable<string>
});
```

However, trying to assign it to an explicitly-typed `ReadableStreamLike` variable - to test the inference using that type - like this:

```ts
it('should accept a ReadableStream', () => {
  const stream: ReadableStreamLike<string> = new ReadableStream<string>({
    pull(controller) {
      controller.enqueue('x');
      controller.close();
    },
  });
  const o = from(stream); // $ExpectType Observable<string>
});
```

fails with this TypeScript error:

```
Type 'ReadableStream<string>' is not assignable to type 'ReadableStreamLike<string>'.
  Types of property 'getReader' are incompatible.
    Type '{ ({ mode }: { mode: "byob"; }): ReadableStreamBYOBReader; (): ReadableStreamDefaultReader<string>; }' is not assignable to type '() => ReadableStreamDefaultReaderLike<string>'.ts(2322)
```

TypeScript is such a party pooper.

benlesh:
@cartant @jayphelps I've added a couple of commits (it should have been one, but I rushed the first one and forgot some important bits) to:

1. Export the `ReadableStreamLike` type with some documentation around what it is. 
2. Add a dtslint test around it.
3. Make the `ReadableStreamLike` type not have compilation errors with the default `ReadableStream` impl's types (They're apparently more exact with how they define the iterator results for that, and it's a little annoying it doesn't pass, but whatever)

cartant (approved):
Noice.

PR #6166, opened 2021-03-21
chore(toPromise): replace deprecated types and use prettier

tmair:
**Description:**
This PR converts`toPromise` operator tests to run mode. I did a few intermediate commits formatting the code before introducing `TestScheduler` as I was not sure if these tests should really be converted to run mode.
Also `toPromise` is not an operator, so the tests should not be located with the other operators. I am not sure where the tests should be moved to instead. Maybe it is just fine to leave them where they are since `toPromise` is deprecated and will be removed sometime in the future.

None

### DIFF
--- a/spec/operators/toPromise-spec.ts
+++ b/spec/operators/toPromise-spec.ts
@@ -1,13 +1,16 @@
+/** @prettier */
 import { expect } from 'chai';
 import { of, EMPTY, throwError, config } from 'rxjs';
 
 /** @test {toPromise} */
 describe('Observable.toPromise', () => {
   it('should convert an Observable to a promise of its last value', (done: Mocha.Done) => {
-    of(1, 2, 3).toPromise(Promise).then(x => {
-      expect(x).to.equal(3);
-      done();
-    });
+    of(1, 2, 3)
+      .toPromise(Promise)
+      .then((x) => {
+        expect(x).to.equal(3);
+        done();
+      });
   });
 
   it('should convert an empty Observable to a promise of undefined', (done: Mocha.Done) => {
@@ -18,12 +21,17 @@ describe('Observable.toPromise', () => {
   });
 
   it('should handle errors properly', (done: Mocha.Done) => {
-    throwError('bad').toPromise(Promise).then(() => {
-      done(new Error('should not be called'));
-    }, (err: any) => {
-      expect(err).to.equal('bad');
-      done();
-    });
+    throwError('bad')
+      .toPromise(Promise)
+      .then(
+        () => {
+          done(new Error('should not be called'));
+        },
+        (err: any) => {
+          expect(err).to.equal('bad');
+          done();
+        }
+      );
   });
 
   it('should allow for global config via config.Promise', (done: Mocha.Done) => {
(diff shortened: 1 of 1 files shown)


spec/operators/toPromise-spec.ts
```diff
@@ -1,42 +1,79 @@
...
+  let rxTestScheduler: TestScheduler;
+
+  beforeEach(() => {
+    rxTestScheduler = new TestScheduler(observableMatcher);
+  });
+
+  it('should convert an Observable to a promise of its last value', (done: Mocha.Done) => {
+    rxTestScheduler.run(({ cold, expectSubscriptions }) => {
+      const source = cold('12(3|)');
+      const subs = '       ^-!';
```
cartant, line 18: We need to be very careful replacing non-marble sources with marble sources, as they're sometimes not the same. One problem is that with marble sources, it's not possible to represent completion _within_ subscribe - see https://github.com/ReactiveX/rxjs/issues/5523.

So, to be sure that we are not changing tests, I don't think we should replace _any_ non-marble sources with marble sources.

spec/operators/toPromise-spec.ts
```diff
@@ -1,42 +1,79 @@
...
   });
 
-  it('should convert an empty Observable to a promise of undefined', (done: MochaDone) => {
-    EMPTY.toPromise(Promise).then((x) => {
-      expect(x).to.be.undefined;
-      done();
+  it('should convert an empty Observable to a promise of undefined', (done: Mocha.Done) => {
+    rxTestScheduler.run(({ cold, expectSubscriptions }) => {
+      const source = cold('|  ');
+      const unsub = '      (^!)';
```
cartant, line 31: Leave as `EMPTY` non-marble source.

spec/operators/toPromise-spec.ts
```diff
@@ -1,42 +1,79 @@
...
-  it('should handle errors properly', (done: MochaDone) => {
-    throwError('bad').toPromise(Promise).then(() => {
-      done(new Error('should not be called'));
-    }, (err: any) => {
-      expect(err).to.equal('bad');
-      done();
+  it('should handle errors properly', (done: Mocha.Done) => {
+    rxTestScheduler.run(({ cold, expectSubscriptions }) => {
+      const source = cold('--#', undefined, 'bad');
+      const subs = '       ^-!';
```
cartant, line 44: Leave as `throwError` non-marble source.

spec/operators/toPromise-spec.ts
```diff
@@ -1,42 +1,79 @@
...
+  it('should allow for global config via config.Promise', (done: Mocha.Done) => {
+    rxTestScheduler.run(({ cold, expectSubscriptions }) => {
+      let wasCalled = false;
+      config.Promise = function MyPromise(callback: Function) {
+        wasCalled = true;
+        return new Promise(callback as any);
+      } as any;
+
+      const source = cold('---n-|', { n: 42 });
+      const unsub = '      ^----!';
```
cartant, line 68: Leave as non-marble source.

cartant (changes requested):

spec/operators/toPromise-spec.ts
tmair: Ok, I was suspecting that this is something that might lead to some issues.

Looking at the tests it seems that none of the `toPromise` tests should be converted to marble tests. Therefore also the introduction of run mode does not make sense IMO. You could just remove the last commit of the PR (that is the one that introduces the marbles) or if you would like I could do that an rephrase the PR description.

cartant: I'll just remove the last commit, I think. Thanks.

cartant (approved):

PR #6168, opened 2021-03-22
fix(partition): add exclusion when using type guard

Boudewijn26:
**Description:**
Add exclusion to partition when using a type guard.

[Previously support for a user defined type guard was added to partition](https://github.com/ReactiveX/rxjs/pull/5756). This meant that Typescript could understand the first item in the resulting tuple to match a certain type. What was missing was Typescript understanding the second item in the tuple to _not_ be of that type. As an example:
```typescript
type AMessage = { a: number };
type BMessage = { b: number };
type Message = AMessage | BMessage;
const aMessage: AMessage = { a: 1 };
const bMessage: BMessage = { b: 2 };
const message$: Observable<Message> = from([aMessage, bMessage]);

function isAMessage(msg: Message): is AMessage { 
  return msg.a;
}
const [aMessage$, bMessage$] = partition(message$, isAMessage);
bMessage$.subscribe((msg) => console.log(msg.b)); // no longer fails with this change
```

Without this change Typescript would consider `bMessage$` to be the same type as `message$`: `Observable<AMessage | BMessage>`. Whereas we are sure `bMessage$` only emits `BMessage`, because all `AMessage` will be put in `aMessage$`.

### DIFF
--- a/src/internal/observable/partition.ts
+++ b/src/internal/observable/partition.ts
@@ -8,11 +8,11 @@ export function partition<T, U extends T, A>(
   source: ObservableInput<T>,
   predicate: (this: A, value: T, index: number) => value is U,
   thisArg: A
-): [Observable<U>, Observable<T>];
+): [Observable<U>, Observable<Exclude<T, U>>];
 export function partition<T, U extends T>(
   source: ObservableInput<T>,
   predicate: (value: T, index: number) => value is U
-): [Observable<U>, Observable<T>];
+): [Observable<U>, Observable<Exclude<T, U>>];
 
 export function partition<T, A>(
   source: ObservableInput<T>,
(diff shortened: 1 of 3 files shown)


src/internal/observable/partition.ts
```diff
@@ -4,15 +4,15 @@ import { ObservableInput } from '../types';
...
   predicate: (this: A, value: T, index: number) => value is U,
   thisArg: A
-): [Observable<U>, Observable<T>];
-export function partition<T, U extends T>(
+): [Observable<U>, Observable<S>];
+export function partition<T, U extends T, S extends Exclude<T, U>>(
   source: ObservableInput<T>,
   predicate: (value: T, index: number) => value is U
-): [Observable<U>, Observable<T>];
+): [Observable<U>, Observable<S>];
```
cartant, line 15: AFAICT, `S` doesn't have to be a type parameter. Instead, `Observable<Exclude<T, U>>` could be used in the return type.

cartant (changes requested):

cartant (approved):
LGTM. Thanks.

Boudewijn26:
Thank you for the quick response, happy to help!

benlesh (approved):

Issue #4033, opened 2018-08-17
`refCount` with delay


OliverJAsh:
## Feature Request

**Is your feature request related to a problem? Please describe.**
I have a `Observable<Array<Observable<T>>>` which I want to map to `Observable<Array<T>>`.

When a new array is emitted, the inner observables should unsubscribe/subscribe as follows:

- If `Observable` exists in previous array and the new/current array, *retain* pre-existing subscription
- If `Observable` did not exist in previous array but does exist in new/current array, *create new subscription*
- If `Observable` existed in previous array but does not exist in new/current array, *unsubscribe* from pre-existing subscription

I hoped to achieve this using `switchMap` on the outer observable and then passing `Array<Observable<T>>` into `combineLatest`. However, `switchMap` will unsubscribe from its previous inner `Observable` before subscribing to the new inner `Observable`, which means inner subscriptions are not retained as desired.

Example:
- https://stackblitz.com/edit/typescript-b4wgr1
- https://stackoverflow.com/questions/51875557/combinelatest-switchmap-and-retaining-inner-subscriptions/51901191

**Describe the solution you'd like**
To offset the fact that `switchMap` unsubscribes to the previous inner observable before subscribing to the new inner observable, I would like the `refCount`ing to be delayed, thereby ensuring that subscriptions to the inner observables are retained.

This is the solution suggested in [this article](https://medium.com/@volkeron/rxjs-unsubscribe-delay-218a9ab2672e
). An alternative `refCount` with delay operator can be found [here](https://gist.github.com/marinho/3637210b13c0f298e1692a0b7b104e64).

**Describe alternatives you've considered**

**(If this is new operator request) describe reason it should be core operator**
From my research this seems like something people want but it's not easy to do. I discovered issues/articles/examples on the internet, but there doesn't seem to be anything in the core to help with this.

**Additional context**
- Related discussion: https://github.com/ReactiveX/rxjs/issues/171
- Issue which needed this: https://stackoverflow.com/questions/51875557/combinelatest-switchmap-and-retaining-inner-subscriptions/51901191#51901191
- Example: `refCountWithDelay` implementation: https://gist.github.com/marinho/3637210b13c0f298e1692a0b7b104e64
- Example: `delayedRefCount` implementation: https://medium.com/@volkeron/rxjs-unsubscribe-delay-218a9ab2672e

aaronjensen:
[...]

PR #6169, opened 2021-03-23
feat(share): use another observable to control resets

backbone87:
This is a PoC. I will add tests, if the approach is agreed up on.

**Description:**
Extends the share operator to allow passing an observable factory to control the reset behavior and enable reset delays.
```ts
declare function createObservableWithExpensiveInitialization(): Observable<number>;

const number$ = createObservableWithExpensiveInitialization().pipe(share({ resetOnRefCountZero: () => timer(5000) }));
concat(number$.take(1), number$.take(1)).subscribe(); // this wont unsubscribe (unless your process is REALLY busy, one could say "stuck")
```

closes https://github.com/ReactiveX/rxjs/issues/4033

### DIFF
--- a/spec/operators/share-spec.ts
+++ b/spec/operators/share-spec.ts
@@ -1,23 +1,42 @@
 /** @prettier */
 import { expect } from 'chai';
+import { asapScheduler, concat, config, defer, EMPTY, NEVER, Observable, of, scheduled, Subject, throwError } from 'rxjs';
 import {
-  share,
-  retry,
-  mergeMapTo,
+  map,
   mergeMap,
-  tap,
+  mergeMapTo,
+  onErrorResumeNext,
   repeat,
+  retry,
+  share,
+  startWith,
   take,
   takeUntil,
   takeWhile,
-  map,
-  startWith,
+  tap,
+  toArray,
   withLatestFrom,
 } from 'rxjs/operators';
-import { Observable, EMPTY, NEVER, of, Subject, defer } from 'rxjs';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
-import sinon = require('sinon');
+import { SinonSpy, spy } from 'sinon';
+
+const syncNotify = of(1);
+const asapNotify = scheduled(syncNotify, asapScheduler);
+const syncError = throwError(() => new Error());
+
+function spyOnUnhandledError(fn: (spy: SinonSpy) => void): void {
+  const prevOnUnhandledError = config.onUnhandledError;
+
+  try {
+    const onUnhandledError = spy();
+    config.onUnhandledError = onUnhandledError;
+
+    fn(onUnhandledError);
+  } finally {
+    config.onUnhandledError = prevOnUnhandledError;
+  }
+}
 
 /** @test {share} */
 describe('share', () => {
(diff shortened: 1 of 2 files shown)


backbone87:
i have a modified version, which passes the existing test cases (and is more concise), while having the exact same public API.

the current approach resulted from a previous attempt, where i used mergeAll insteadof switchAll, to allow "concurrently" running resets. after putting some thought into it, i came to the conclusion that this is not needed (and would involve much more complicated logic beyond just using mergeMap), because the error/complete reset notifiers should always overrule/replace the refCount reset notifiers.

src/internal/operators/share.ts
```diff
@@ -19,15 +24,15 @@ export interface ShareConfig<T> {
    * or resubscriptions will resubscribe to that same subject. In all cases, RxJS subjects will emit the same error again, however
    * {@link ReplaySubject} will also push its buffered values before pushing the error.
    */
-  resetOnError?: boolean;
+  resetOnError?: boolean | ((error: any) => Observable<any>);
```
benlesh, line 27: I'm not sure this makes sense for `resetOnError` and `resetOnComplete`. Semantically, those are "done", so it's not like something were we want to "keep something open" before we check to see if it "unerrors" or "uncompletes". Whereas with refCounting, we're waiting to see if the refCount goes back up from zero.

backbone87: Resets on error/complete are not canceled. This notifier allows conditional/delayed resets (e.g. "cooldown" after error).

src/internal/operators/share.ts
```diff
@@ -90,25 +95,44 @@ export function share<T>(options: ShareConfig<T>): MonoTypeOperatorFunction<T>;
...
+    // We need to capture the connection before
+    // we reset (if we need to reset).
+    const conn = connection;
+    reset();
+    conn?.unsubscribe();
+  };
+
+  const resetOnComplete = createResetNotifierFactory(reset, options.resetOnComplete);
+  const resetOnError = createResetNotifierFactory(reset, options.resetOnError);
+  const resetOnRefCountZero = createResetNotifierFactory(resetAndUnsubscribe, options.resetOnRefCountZero);
```
backbone87, line 129: Question: do we want to pay the normalization cost once up front or should I change this and normalize when the reset happens?

backbone87:
what can i do, to push this forward?

cartant:
> what can i do, to push this forward?

Nothing. ATM, I have higher v7 priorities. This would not be a breaking change, so we can implement this after v7 is done/released. It's not a v7 blocker, so it's not near the top of my things-to-do list.

backbone87:
i actually have the need for this in one of my projects. it is not critical, but would improve behavior. i tried to find a work around by combining other operators (v6), but failed to achieve this behavior, except for having a permanent subscription so it never resets.

cartant:
> i actually have the need for this in one of my projects.

Unfortunately, that doesn't change my priorities.

cartant:
In general, I think this is okay, but it's going to need a whole bunch of tests. And it's going to need to be updated to account for  the change in https://github.com/ReactiveX/rxjs/pull/6151 - presumably, once it's no longer a draft, GitHub will show that there's a conflict that needs to be resolved. Maybe start with the latter.

backbone87:
- rebased onto latest master
- added docs
- switched to lazy options normalization

src/internal/operators/share.ts
```diff
@@ -84,30 +93,72 @@ export function share<T>(options: ShareConfig<T>): MonoTypeOperatorFunction<T>;
...
+ *
  * @see {@link api/index/function/interval}
  * @see {@link map}
  *
  * @return A function that returns an Observable that mirrors the source.
  */
-export function share<T>(options?: ShareConfig<T>): OperatorFunction<T, T> {
-  options = options || {};
-  const { connector = () => new Subject<T>(), resetOnComplete = true, resetOnError = true, resetOnRefCountZero = true } = options;
+export function share<T>(options: ShareConfig<T> = {}): MonoTypeOperatorFunction<T> {
```
backbone87, line 128: technically moving from `options = options || {}` to `options: ShareConfig<T> = {}` is a BC break, if you don't adhere to the typings. what is the stance towards this?

src/internal/operators/share.ts
backbone87: i have moved the code to lazy normalization

src/internal/operators/share.ts
cartant: I have no issue with this and I prefer the default that you've introduced. If someone is passing `null`, `0` or `""`, they've already violated our contract - expressed in the types and the docs that are derived from them - so if the behaviour of their code changes that's fine and and the change is non-breaking, IMO. Usage that violates the contract has undefined behaviour.

cartant:
This PR is going to need to be rebased again, to account for the change introduced in https://github.com/ReactiveX/rxjs/pull/6370

Also, I'd prefer this call to `cancelReset` to be removed. If there's a situation that cannot happen, that should be expressed in the tests and not by calling something that should always be ineffectual:

https://github.com/ReactiveX/rxjs/blob/f32f79eba7cbdff5939d00f4861d3e8985d57ed7/src/internal/operators/share.ts#L207-L210

After you've done the rebase, I'll review this and will let you know what needs to be tested. If you're unfamiliar with marble tests, I'll write one to show you what we need to do. This is going to need a lot of tests.

backbone87:
- rebased against master
- removed paranoia reset
- added general documentation about reset behavior

cartant (commented):
I've read through this and I have no further suggestions. It LGTM. I'll prepare a list of what I think we need to test and I'll post it here sometime tomorrow.

backbone87:
i have added some tests already
- execute all existing tests with equivalent sync reset notifiers
- tests for async/defered reset on ref count zero notifiers

spec/operators/share-spec.ts
```diff
@@ -27,535 +27,549 @@ describe('share', () => {
...
+  }
+
+  for (const { title, resetOnError, resetOnComplete, resetOnRefCountZero } of [
+    { title: 'share(config)', resetOnError: false, resetOnComplete: false, resetOnRefCountZero: false },
+    {
+      title: 'share(config) using sync reset notifier equivalents',
+      resetOnError: () => EMPTY,
+      resetOnComplete: () => EMPTY,
+      resetOnRefCountZero: () => EMPTY,
+    },
```
cartant, line 483: Nice. We should include another parameterised run here that uses `NEVER`.

cartant (commented):
IDK that I'm going to get around to creating that list, today. But one of the things we definitely need to test is that when a reset due to an error or a complete is initiated, any factory passed via the `resetOnRefCountZero` is _not_ called.

backbone87:
- more sync reset notifier equivalents:
  - NEVER after notify for resetting tests
  - error after notify for resetting tests
  - NEVER for non-resetting tests
  - sync error for non-resetting test -> **these fail the tests - different failures for the same tests depending on `useDeprecatedSynchronousErrorHandling`** -> what should the behavior be?
- added "should not reset on refCount 0 after reset notifier errored?" -> **fails** -> what should the behavior be?

backbone87:
added some cleanups to the impl:
- optimize internal naming
- remove unnecessary typings
- avoid non-null type assertation

cartant:
> `useDeprecatedSynchronousErrorHandling`

I would not add tests for this. It's going to be removed.

Regarding the handling of errors from notifiers, IMO we should make no attempt to handle and report them. There is really no avenue for their being reported anyway. Leave them fall through to `config.onUnhandledError` - which you can use in the tests.

https://github.com/ReactiveX/rxjs/blob/c23e567ccaee4a9aa2a7046519e7cf46a7191a3f/src/internal/config.ts#L22-L31

> should not reset on refCount 0 after reset notifier errored?

It should not reset. The only signal for any of the reset is a next notification. If once isn't received, there should be no reset. Errors should be reported via `config.onUnhandledError` as above.

> In general, i would say it is an anti-pattern to use unconditional NEVERs as notifiers (in any operator).

Yeah, but we still need to test for it. This is done with other operators, IIRC. It will leak, but that's expected.

backbone87:
- should contain all necessary tests now, if something is missing let me know
- squashed commits

backbone87:
in regards to notifiers emitting errors: we could attach a noop error handler when subscribing to notifiers, if the user wants to handle/react errors himself he could use `tap` or `catchError`

cartant:
> in regards to notifiers emitting errors: we could attach a noop error handler ...

I think it should be kept as simple as possible. The user is already providing a factory that returns an observable. If a potential error is something they thing they can handle, they already have the opportunity to compose that into the observable they are giving us.

benlesh:
The list I followed was more of a matrix. You pick all of the ones that make sense from this, for each observable involved:

| | sync | async | mixed |
|--:|:-----:|:------:|:-------:|
| single value | | | |
| multi value |  | | |
| empty |  | | |
| never |  | | |
| error |  | | |
| values and error | | | |
| values and never |  | | |

(clearly, by these standards, none of our operators has 100% coverage, probably, but like I said, pick the ones that make the most sense.)

backbone87:
i dont fully understand your table. but i guess you mean test coverage based on notifier characteristics? basically, the handling of the notifier observable is tied to `take(1)` semantics, but I try to answer:

| |sync | async | mixed|should reset?
|-|-----|------|--------|------
single value | x | x ||yes
multi value | same as value + error| ||yes
empty|x|same as sync||no
never|x|same as sync||no
error||x||no
value + error|x |||yes
value + never|x|||yes

(what do you mean with mixed? a sync emit + something async after?)

backbone87:
added some more test cases around sync multi value notifier, as well as firehose observable + sync notifier and explicit test for sync resubscribe with sync notifier.

src/internal/operators/share.ts
```diff
@@ -84,33 +101,75 @@ export function share<T>(options: ShareConfig<T>): MonoTypeOperatorFunction<T>;
...
 
   return operate((source, subscriber) => {
     refCount++;
+    if (!hasErrored && !hasCompleted) {
+      cancelReset();
+    }
 
     // Create the subject if we don't have one yet.
-    subject = subject ?? connector();
+    const dest = (subject = subject ?? connector());
```
cartant, line 172: The refactor that avoids the non-null assertions is fine, but I'd like to see a brief comment that explains why we have two variables here: `dest` and `subject`. I think it's warranted, as it's not immediately obvious. The comment just needs to point out that `dest` is needed for the notification after `subject` is (potentially) reset.

cartant (approved):
This is great. I'm more than happy with the tests. Thanks, @backbone87. I've suggested adding a comment to your `dest` declaration refactor, but approving this now. Thanks.

src/internal/operators/share.ts
backbone87: done

benlesh:
Okay. I finally had time to review this again. 

This looks really good. The only thing I'm not sure about is that we need to delay reconnect for `complete` and `error`. While I think it "rounds out" the API, I'm not sure that I can think of a use case for that.

In the case of _complete_ and _error_ the source is signaling that it's done and can no longer send anything else. What is the purpose of delaying the reseting of the connector at that point? The source can't possibly send anything else. Delaying isn't going to do anything.

I can _DEFINITELY_ think of use cases for delaying the  reset on _ref count_. I can think of plenty of use cases there, because that's something downstream. It's not the source saying it can't send anything else.

@cartant, since you approved this, maybe you have other thoughts, but it seems like we should only be supporting this for `resetOnRefCountZero`. I think that's tremendously useful. But in the other cases, if we enable it, we'll end up having to support people who use it that way. I'm concerned that those usages will be really weird, possibly wrong-headed, and we'll be on the hook for supporting them.

Otherwise, I generally approve of this entire PR. I just think we want to ratchet it back to only the `resetOnRefCountZero` option.

benlesh:
After some discussion with @cartant, the feature makes sense. He pointed out that it could be used to cache a completed source for a specified amount of time. That's a reasonable use case for completion. Still unsure about `error`, however, at the point that two of them accept it, it's probably confusing if they all don't accept it.

benlesh:
Thank you, @backbone87, for this great contribution!

PR #6170, opened 2021-03-23
chore(toArray): convert toArray tests to run mode

tmair:
**Description:**
This PR converts the `toArray` operator tests to run mode.

None

### DIFF
--- a/spec/operators/toArray-spec.ts
+++ b/spec/operators/toArray-spec.ts
@@ -1,115 +1,145 @@
-import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
+/** @prettier */
 import { toArray, mergeMap } from 'rxjs/operators';
 import { of } from 'rxjs';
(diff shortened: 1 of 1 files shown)


spec/operators/toArray-spec.ts
```diff
@@ -1,115 +1,145 @@
...
+      const e1 = cold(' -');
+      const e1subs = '  ^';
+      const expected = '-';
+
+      expectObservable(e1.pipe(toArray())).toBe(expected);
+      expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    });
   });
 
   it('should be never when source is empty', () => {
```
cartant, line 37: We might was well fix this while we're here:

```suggestion
  it('should be empty when source is empty', () => {
```

cartant (changes requested):

cartant (approved):
LGTM. Thanks.

PR #6171, opened 2021-03-23
chore(docs): upgrade Swirly to latest version

timdp:
Also upgrades the lockfile for the docs app to v2 by using npm@7. This was already the case for the repo-root-level one.


timdp:
CI failed but it was due to a temporary socket-level glitch. I temporarily pushed an empty commit to validate and [that worked](https://github.com/ReactiveX/rxjs/actions/runs/680161512), so I removed the empty commit again. If you retry the original run, it should pass.

benlesh (approved):

benlesh:
@niklas-wortmann note that this didn't get merged into 6.x, as that doesn't seem to use Swirly.

PR #6173, opened 2021-03-24
chore: add basic watch mode

benlesh:
- Adds `npm run watch` to watch files and test files and run `test`
- Adds `npm run watch:dtslint` to do the same while iterating on types
- Fixes two tests that were polluting global state

We probably should have done this forever ago. I mean... we had something once upon a time, but we removed it, I'm not sure why, and never put it back. haha.


benlesh:
It's worth noting that I _very_ specifically installed `nodemon@1.19.2` because the latest version of `nodemon` will not install on any OS but Mac OSX for whatever reason. I'm on Ubuntu (via WSL, but I don't think that makes a difference given the bug reports I've seen)

cartant (approved):

PR #6174, opened 2021-03-24
feat(ajax): add `params` query string parameter configuration

benlesh:
Adds a feature that will accept a wide variety of possible inputs for `params` to set the query string parameters for ajax.

Note that this feature will require a polyfill for `URLSearchParams` in IE, but this is not a breaking change for IE, as if they do not use this feature `URLSearchParams` is never accessed. Adds documentation and relatively thorough tests.

Related to #5384

### DIFF
--- a/integration/side-effects/snapshots/esm/ajax.js
+++ b/integration/side-effects/snapshots/esm/ajax.js
@@ -1,1 +1,1 @@
-
+import "tslib";
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -1429,7 +1429,7 @@ x-headers-are-fun: <whatever/> {"weird": "things"}`);
       ajax({
         method: 'GET',
         url: '/whatever',
-        params: '?foo=bar&whatever=123',
+        params,
       }).subscribe();
 
       const mockXHR = MockXMLHttpRequest.mostRecent;
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -1366,6 +1366,173 @@ describe('ajax', () => {
 x-custom-header: test
 x-headers-are-fun: <whatever/> {"weird": "things"}`);
   });
+
+  describe('with params', () => {
+    it('should allow passing of search params as a dictionary', () => {
+      ajax({
+        method: 'GET',
+        url: '/whatever',
+        params: { foo: 'bar', whatever: '123' },
+      }).subscribe();
+
+      const mockXHR = MockXMLHttpRequest.mostRecent;
+
+      mockXHR.respondWith({
+        status: 200,
+        responseText: JSON.stringify({ whatever: 'I want' }),
+      });
+
+      expect(mockXHR.url).to.equal('/whatever?foo=bar&whatever=123');
+    });
+
+    it('should allow passing of search params as an entries array', () => {
+      ajax({
+        method: 'GET',
+        url: '/whatever',
+        params: [
+          ['foo', 'bar'],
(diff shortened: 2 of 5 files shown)


src/internal/ajax/ajax.ts
```diff
@@ -275,15 +275,53 @@ const LOAD = 'load';
...
+    // that we don't want to add to the request information we're
+    // passing around.
+    const { params, body: configuredBody, headers: configuredHeaders, ...remainingConfig } = config;
+
+    let { url } = remainingConfig;
+    if (!url) {
+      throw new TypeError('url is required');
+    }
+
+    if (params) {
```
benlesh, line 288: Everything inside of this conditional is the bulk of what was added. Please notice that it's mostly comments, and not actually a lot of code. Also worth noticing that `URLSearchParams` is only ever accessed inside of this conditional.

integration/side-effects/snapshots/esm/ajax.js
```diff
@@ -1 +1 @@
-
+import 'tslib';
```
benlesh, line 1: It seems like using `URLSearchParams` pulls in `tslib` here. I was unaware that `tslib` polyfilled that. It might be the case that this "just works" in IE, which is weird, and almost unfortunate, because I'd rather not depend on that and not have it bundled in with this (to keep it smaller).

I guess this is something to discuss at the core team meeting then.

@cartant @kwonoj  do you any thoughts on this?

benlesh: (this comment isn't outdated, I just had to update the file after prettier mangled it and ruined the test)

cartant: IDK anything about the tool that's generating this side-effects business, but my guess would be that it has more to do with the object rest destructuring that was added here:

https://github.com/ReactiveX/rxjs/blob/fa9415e9c25d3e7819c238b5d6da5e825434811b/src/internal/ajax/ajax.ts#L281

benlesh: Ah. That makes sense.

spec/observables/dom/ajax-spec.ts
```diff
@@ -1366,6 +1366,173 @@ describe('ajax', () => {
...
+
+    it('should allow passing of search params as a URLSearchParams object', () => {
+      const params = new URLSearchParams();
+      params.set('foo', 'bar');
+      params.set('whatever', '123');
+      ajax({
+        method: 'GET',
+        url: '/whatever',
+        params: '?foo=bar&whatever=123',
+      }).subscribe();
```
cartant, line 1433: issue: AFAICT, this isn't passing the `params` object; it's passing a string.

cartant (approved):
LGTM. Just a nitpick and what looks like a copy/paste error in a test. Approving this 'cause the latter needs no discussion.

spec/observables/dom/ajax-spec.ts
benlesh: 🤦‍♂️

jayphelps:
Late to the party here, but note that “params” can also mean (and IMO more often refers to) route params a la Express ‘/products/:productId’ Express’s API calls these ‘params’. Angular and many others do as well.

I suggest naming it query, queryParams, search, or searchParams, to follow naming conventions used elsewhere. My gut says queryParams is probably the right choice since that’s what Angular uses: https://angular.io/api/router/NavigationExtras

cartant:
@benlesh +1 from me for `queryParams` ☝

benlesh:
I'm out this morning, but if someone puts a PR in to change it, I'm down.

PR #6176, opened 2021-03-26
chore: update mergeScan tests to run mode

jakovljevic-mladen:
**Description:**
Updated `mergeScan` tests to use run mode.

None

### DIFF
--- a/spec/operators/mergeScan-spec.ts
+++ b/spec/operators/mergeScan-spec.ts
@@ -1,140 +1,163 @@
-import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
+/** @prettier */
 import { TestScheduler } from 'rxjs/testing';
 import { of, defer, EMPTY, NEVER, concat, throwError, Observable } from 'rxjs';
 import { mergeScan, delay, mergeMap, takeWhile, startWith, take } from 'rxjs/operators';
 import { expect } from 'chai';
+import { observableMatcher } from '../helpers/observableMatcher';
 
-declare const rxTestScheduler: TestScheduler;
 /** @test {mergeScan} */
 describe('mergeScan', () => {
+  let testScheduler: TestScheduler;
+
+  beforeEach(() => {
+    testScheduler = new TestScheduler(observableMatcher);
+  });
+
   it('should mergeScan things', () => {
-    const e1 = hot('--a--^--b--c--d--e--f--g--|');
-    const e1subs =      '^                    !';
-    const expected =    '---u--v--w--x--y--z--|';
+    testScheduler.run(({ hot, expectObservable, expectSubscriptions }) => {
+      const e1 = hot('--a--^--b--c--d--e--f--g--|');
+      const e1subs = '     ^--------------------!';
+      const expected = '   ---u--v--w--x--y--z--|';
 
-    const values = {
-      u: ['b'],
-      v: ['b', 'c'],
-      w: ['b', 'c', 'd'],
-      x: ['b', 'c', 'd', 'e'],
-      y: ['b', 'c', 'd', 'e', 'f'],
-      z: ['b', 'c', 'd', 'e', 'f', 'g']
-    };
+      const values = {
+        u: ['b'],
+        v: ['b', 'c'],
+        w: ['b', 'c', 'd'],
+        x: ['b', 'c', 'd', 'e'],
+        y: ['b', 'c', 'd', 'e', 'f'],
+        z: ['b', 'c', 'd', 'e', 'f', 'g'],
+      };
 
-    const source = e1.pipe(mergeScan((acc, x) => of(acc.concat(x)), [] as string[]));
+      const result = e1.pipe(mergeScan((acc, x) => of(acc.concat(x)), [] as string[]));
 
-    expectObservable(source).toBe(expected, values);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+      expectObservable(result).toBe(expected, values);
+      expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    });
   });
 
   it('should handle errors', () => {
-    const e1 = hot('--a--^--b--c--d--#');
-    const e1subs =      '^           !';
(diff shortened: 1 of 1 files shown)


spec/operators/mergeScan-spec.ts
```diff
@@ -154,286 +177,294 @@ describe('mergeScan', () => {
...
+      expectSubscriptions(inner[1].subscriptions).toBe(ysubs);
+      expectSubscriptions(inner[2].subscriptions).toBe(zsubs);
+    });
   });
 
-  it('should emit accumulator if inner completes without value', () => {
-    const e1 = hot('--a--^--b--c--d--e--f--g--|');
-    const e1subs =      '^                    !';
-    const expected =    '---------------------|';
+  it('should not emit accumulator if inner completes without value', () => {
```
jakovljevic-mladen, line 342: Adding `not` here (as of #5805).

spec/operators/mergeScan-spec.ts
```diff
@@ -154,286 +177,294 @@ describe('mergeScan', () => {
...
-    const expected =    '-----------------------|';
-    const inner = cold(                   '-----|');
-
-    const source = e1.pipe(
-      mergeScan(() => inner, '1')
-    );
-
-    expectObservable(source).toBe(expected);
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  it('should not emit accumulator if inner completes without value after source completes', () => {
```
jakovljevic-mladen, line 355: And here.

spec/operators/mergeScan-spec.ts
```diff
@@ -1,140 +1,163 @@
...
+
+      const result = e1.pipe(
         mergeMap((x) => of(x)),
-        mergeScan((acc, x) =>
-          of([...acc, x]).pipe(
-            delay(50, rxTestScheduler)
-          )
-        , [] as string[]),
-        mergeMap(function (x) { return of(x); })
+        mergeScan((acc, x) => of(acc.concat(x)).pipe(delay(t)), [] as string[]),
```
jakovljevic-mladen, line 154: Should I replace this `of` call with the `cold`?

jakovljevic-mladen (commented):
A couple of notes and a question.

spec/operators/mergeScan-spec.ts
cartant: Nope, I don't think we should replace any non-`TestScheduler` observables with either `hot` or `cold` because there are some scenarios that they cannot represent. Let's just play it safe.

jakovljevic-mladen: Yeah, smells like #5523. But, how safe was this: https://github.com/ReactiveX/rxjs/pull/5805/files#diff-a76406e2b4935afce22a4ba7e0b415b12350464291e64e81daf12a3b180c7606L326 ?

jakovljevic-mladen: I guess that `EMPTY` doesn't notifies anything upon subscription, so this was safe enough...

jakovljevic-mladen: But, I'm still not sure that in the case of `of(...).pipe(delay(...))` we could have something like what's described in the #5523. The `delay` should not allow synchronous-upon-subscription notifications.
I don't think we _have to_ change anything here now, but I'd just like to be sure that this is not the synchronous-upon-subscription case and that it _should_ be safe to convert this to `cold`/`hot`.
cc @cartant

cartant: > I guess that, because `EMPTY` doesn't notify anything upon subscription, this was safe enough

`EMPTY` does emit a `complete` notification upon subscription, so it cannot be represented with a `TestScheduler` observable. In the test that you've linked - and as you've mentioned - the use of `delay` - means that the effect will be the same - whether a marble observable or `of` is used. There are going to be cases where the replacement it safe, I just don't want replacements to be made in PRs that have descriptions like 'convert X tests to run mode'.

spec/operators/mergeScan-spec.ts
```diff
@@ -1,140 +1,163 @@
...
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    testScheduler.run(({ hot, time, expectObservable, expectSubscriptions }) => {
+      const e1 = hot('--a--^--b--c--d--e--f--g--|');
+      const e1subs = '     ^--------------------!';
+      const t = time('        --|                '); // t = 2
+      //                         --|
+      //                            --|
+      //                               --|
+      //                                  --|
+      //                                     --|
```
cartant, line 66: 👍

spec/operators/mergeScan-spec.ts
```diff
@@ -1,140 +1,163 @@
...
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    testScheduler.run(({ hot, time, expectObservable, expectSubscriptions }) => {
+      const e1 = hot('--a--^--b--c--d--e--f--g--|     ');
+      const e1subs = '     ^--------------------!     ';
+      const t = time('        -----|'); //          acc = []; x = 'b'; acc.concat(x) = ['b']; t = 5
+      //                         -----|             acc = []; x = 'c'; acc.concat(x) = ['c']
+      //                            -----|          acc = ['b']; x = 'd'; acc.concat(x) = ['b', 'd']
+      //                               -----|       acc = ['c']; x = 'e'; acc.concat(x) = ['c', 'e']
+      //                                  -----|    acc = ['b', 'd']; x = 'f'; acc.concat(x) = ['b', 'd', 'f']
+      //                                     -----| acc = ['c', 'e']; x = 'g'; acc.concat(x) = ['c', 'e', 'g']
```
cartant, line 94: 👍

cartant (approved):
LGTM. Thanks.

spec/operators/mergeScan-spec.ts
jakovljevic-mladen: Yes, I understand. That's why I didn't do it in this PR in the first place, but I asked a question should (and could) I replace it later. I won't create any additional PRs for now.

PR #6188, opened 2021-03-28
docs(mergeScan): add parameter docs

jakovljevic-mladen:
**Description:**
While working on converting `mergeScan` tests to run mode, I struggled with finding out how `acc` parameter of the `accumulator` function is being set, so I had to read the code to figure it out. Now, I'm adding some documentation about `mergeScan` parameters and `mergeScan` internals.

None


cartant (approved):
LGTM. Thanks.

PR #6192, opened 2021-03-29
Add exports within package.json to enable scoped package loading.

samccone:
In order for module resolution to work with .mjs or package: module code
we need to utilize the conditional export feature of node.

> The current specifier resolution does not support all default behavior of the CommonJS loader. One of the behavior differences is automatic resolution of file extensions and the ability to import directories that have an index file.

https://nodejs.org/api/esm.html#esm_customizing_esm_specifier_resolution_algorithm
https://nodejs.org/api/packages.html#packages_conditional_exports

This directly enables rxjs to work with mjs code and commonjs code since
mjs does not support loading bare folder paths.

This is a fix to:
https://github.com/sveltejs/kit/issues/612

and is directly related to the conversation in this issue within node
core https://github.com/nodejs/node/issues/27408

-----

#### Test case

```
import {Observable} from 'rxjs';
import {scan} from 'rxjs/operators';

console.log(Observable, scan);
```

```
{
  "type": "module",
  "dependencies": {
    "rxjs": "^7.0.0-beta.13" // With the above modification
  }
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,33 @@
   "es2015": "./dist/esm/index.js",
   "types": "./dist/types/index.d.ts",
   "sideEffects": false,
+  "exports": {
+    ".": {
+      "node": "./dist/cjs/index.js",
+      "default": "./dist/esm5/index.js"
+    },
+    "./ajax": {
+      "node": "./dist/cjs/ajax/index.js",
+      "default": "./dist/dist/esm5/ajax/index.js"
+    },
+    "./fetch": {
+      "node": "./dist/cjs/fetch/index.js",
+      "default": "./dist/dist/esm5/fetch/index.js"
+    },
+    "./operators": {
+      "node": "./dist/cjs/operators/index.js",
+      "default": "./dist/dist/esm5/operators/index.js"
+    },
+    "./testing": {
+      "node": "./dist/cjs/testing/index.js",
+      "default": "./dist/dist/esm5/testing/index.js"
+    },
+    "./webSocket": {
+      "node": "./dist/cjs/webSocket/index.js",
+      "default": "./dist/dist/esm5/webSocket/index.js"
+    },
+    "./package.json": "./package.json"
+  },
   "config": {
     "commitizen": {
       "path": "cz-conventional-changelog"
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "prepublishOnly": "npm run build:package && npm run lint && npm run test && npm run test:circular && npm run dtslint && npm run test:side-effects && npm run api_guardian",
     "publish_docs": "./publish_docs.sh",
     "test": "npm run compile && mocha --config spec/support/.mocharc.js \"dist/spec/**/*-spec.js\"",
+    "test:esm": "node spec/module-test-spec.mjs",
     "test:browser": "echo \"Browser test is not working currently\" && exit -1 && npm-run-all build:spec:browser && opn spec/support/mocha-browser-runner.html",
     "test:circular": "dependency-cruiser --validate .dependency-cruiser.json -x \"^node_modules\" dist/esm5",
     "test:systemjs": "node integration/systemjs/systemjs-compatibility-spec.js",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -7,6 +7,16 @@
...
+  "exports": {
+    ".": {
+      "node": "./dist/cjs/index.js",
+      "default": "./dist/esm5/index.js"
+    },
+    "./operators": {
+      "node": "./dist/cjs/operators/index.js",
+      "default": "./dist/dist/esm5/operators/index.js"
+    }
+  },
```
MylesBorins, line 19: FYI this is a breaking change as it locks down the public interface of accessible deep imports. At the very least you probably want an explicit entry for the `package.json`

```suggestion
    },
    "./package.json": "./package.json"
  },
```

package.json
```diff
@@ -7,6 +7,16 @@
...
   "sideEffects": false,
+  "exports": {
+    ".": {
+      "node": "./dist/cjs/index.js",
+      "default": "./dist/esm5/index.js"
+    },
+    "./operators": {
+      "node": "./dist/cjs/operators/index.js",
+      "default": "./dist/dist/esm5/operators/index.js"
+    }
```
benlesh: There would also need to be entries for:

- rxjs/testing
- rxjs/webSocket
- rxjs/fetch
- rxjs/ajax

benlesh (changes requested):
I'm not totally sure of the implications here. But I can see that we're missing a few spots.

benlesh:
I'm really glad you're looking into this, @samccone.

Also, I guess we'd need to be running some sort of simple test that shows the problem that this is fixing, and that it's working now.

Thank you, @MylesBorins for helping to review this, BTW.

samccone:
As for a test @benlesh, the test is actually quite simple!


#### test_file.mjs
```
import {Observable} from 'rxjs';
import * as o from 'rxjs/operators';
import * as a from 'rxjs/ajax';
import * as f from 'rxjs/fetch';
import * as t from 'rxjs/testing';
import * as w from 'rxjs/webSocket';

console.log(Observable, o, a, f, t, w);
```

In node 14,15 run `node test_file.mjs` against the built package. Do you have any similar tests to this currently?

benlesh:
> Do you have any similar tests to this currently?

We have some import tests that @kwonoj set up, but I'm not sure they cover this. That test is so simple that it seems like just running that would be fine, IMO.

samccone:
ok @benlesh test case added and new github workflow config has been modified to run the test.

cartant (approved):
LGTM.

Widespread support for `exports` in `package.json` is something I've been looking forward to for a long time, so this is particularly encouraging.

One thing we might want to look at - not necessarily in this PR, but it's something I think we should discuss before v7 is released - is whether we really want to be downleveling our ES modules to ES5. It's 2021 and distributing ES5 makes me cry. 😢

MylesBorins:
@cartant moving to native ESM will be an even bigger lift as Node.js doesn't support File Extension Resolution or Folder Resolution by default... So every internal import will need to be updated in the source (or modified in transpiled output). There are creative ways to continue to use extensionless imports internally without requiring any flags (Specifically [Subpath Pattern](https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_subpath_patterns)), but they do require using a non-relative import specifier and I don't believe that TS supports this yet.

benlesh:
> is whether we really want to be downleveling our ES modules to ES5. It's 2021 and distributing ES5 makes me cry. 😢

Same... I want to get rid of those. Let's add it the the agenda.

> @cartant moving to native ESM will be an even bigger lift as Node.js doesn't support File Extension Resolution or Folder Resolution by default... So every internal import will need to be updated in the source (or modified in transpiled output). There are creative ways to continue to use extensionless imports internally without requiring any flags (Specifically Subpath Pattern), but they do require using a non-relative import specifier and I don't believe that TS supports this yet.

@MylesBorins Can you summarize what we need to do to be the best stewards of the community here? What should we ideally be outputting? How can we test it?

benlesh (approved):

benlesh:
Actually, @MylesBorins, if you have the time, I'd love it if you started a new issue for this with some checkboxes of what we need to have, and how we could test it? No worries if you don't have the time. We're all volunteers here. 

I want to merge this, and try to get the conversation somewhere it makes sense. <3

PR #6195, opened 2021-03-31
chore(ajax): rename `params` query string to `queryParams`

josepot:
**Description:**
As @jayphelps  suggested in [this comment](https://github.com/ReactiveX/rxjs/pull/6174#issuecomment-811036029) `queryParams` is a better name.

### DIFF
--- a/api_guard/dist/types/ajax/index.d.ts
+++ b/api_guard/dist/types/ajax/index.d.ts
@@ -9,9 +9,9 @@ export interface AjaxConfig {
     includeDownloadProgress?: boolean;
     includeUploadProgress?: boolean;
     method?: string;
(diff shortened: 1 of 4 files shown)


jayphelps:
Thanks for jumping in here.

benlesh:
@josepot you just need to run `npm run api_guardian:update` to update the golden files for the API changes.  Commit as `chore: update golden files` or something. It'll get squashed out anyhow.

benlesh (approved):

josepot:
> @josepot you just need to run npm run api_guardian:update to update the golden files for the API changes.

Thanks! I was wondering what was going on. I had edited it manually, but the order didn't match. Now I get it.

> Commit as chore: update golden files or something. It'll get squashed out anyhow.

I amended the current commit and I force pushed, that's fine too, right?

cartant (approved):

PR #6199, opened 2021-04-02
chore: use descriptive parameter names

cartant:
**Description:**

This PR changes the parameter names of some tuple-based signatures to use descriptive names - e.g. `sourcesAndScheduler` instead of `args`.


benlesh (approved):
Fully agree. Nice.

Issue #4866, opened 2019-06-17
TypeScript: `startWith`: multiple generics


OliverJAsh:
## Bug Report

**Reproduction**

```ts
declare const foo$: Observable<{ foo: number }>
// Expected type: { foo: number }
// Actual type: { foo: number } | { foo: number }
const bar$ = foo$.pipe(startWith({ foo: 1 }))
```

Although the expected/actual types are compatible, it's slightly irritating to deal with a union type `T | T` instead of a consolidated type `T`, as it makes type inspection in VS Code much more difficult (especially when the types are large or deeply nested).

Is there a way to tell `startWith` to consolidate its generics if they are the same? I tried adding an overload with just one generic, but then other tests started to fail.

cartant:
I've confirmed that this dtslint expectation fails:

```ts
const o = of<{ foo: number }>().pipe(startWith({ foo: 1 })); // $ExpectType Observable<{ foo: number; }>
```
```
ERROR: 5:3  expect  Expected type to be:
  Observable<{ foo: number; }>
got:
  Observable<{ foo: number; } | { foo: number; }>
```

And this one passes:

```ts
const v = { foo: 1 };
const o = of<typeof v>().pipe(startWith(v)); // $ExpectType Observable<{ foo: number; }>
```

However, I'd suggest that this is a TypeScript issue and for that reason I'd be reluctant to implement - what would be, IMO - a workaround.

I cannot see the reason for TypeScript's not consolidating a union of identical shapes.

cartant:
I guess performance would be a good-enough reason, but, if that's the case, I'd still be reluctant to implement a workaround. Particularly, as this mechanism is used extensively throughout RxJS's types.

OliverJAsh:
If you're interested, this overload (added as the first) provides a workaround:

```ts
export function startWith<T, D extends T>(v1: D): OperatorFunction<T, T | D>;
```

> I cannot see the reason for TypeScript's not consolidating a union of identical shapes.

I can file an issue and ask.

cartant:
> If you're interested, this overload (added as the first) provides a workaround ...

Hmm. Maybe. But to 'fix' it for all of the signatures will involve a lot of typing (pun intended).

Also, I'm not sure how this will affect the typings that will be coming in v7 that allow for the inference of N-arguments, etc. Maybe the `D extends T` approach can be used - or something similar can be incorporated into a conditional type. IDK. Will have to think about it.

[...]

PR #6204, opened 2021-04-03
feat: add (optional) defaultValue parameters to firstValueFrom and lastValueFrom

cartant:
**Description:**

This PR adds an optional `defaultValue` parameter to `firstValueFrom` and `lastValueFrom`. The parameters are added using an overload to [avoid this issue](https://github.com/ReactiveX/rxjs/issues/4866) (for some usage, at least).

The `first` and `last` operators both support defaults and I see no reason why `firstValueFrom` and `lastValueFrom` should not do the same. Sure, you could apply `defaultIfEmpty` to the source, but you say the same for `first` and `last`. And, IMO, including `defaultValue` in the signatures serves as a reminder that the functions expect a value and that not having an available value is an unexpected/error situation.

Also, supporting default values would make implementing [a user-land equivalent that resolves with an `Option`](https://github.com/ReactiveX/rxjs/issues/5075#issuecomment-757908538) straightforward. (IMO, introducing a `Option` isn't something we should do while there are bunch of APIs that throw for empty sources.)

### DIFF
--- a/src/internal/lastValueFrom.ts
+++ b/src/internal/lastValueFrom.ts
@@ -51,7 +51,7 @@ export function lastValueFrom<T>(source: Observable<T>): Promise<T>;
  * @param config a configuration object to define the `defaultValue` to use if the source completes without emitting a value
  */
 export function lastValueFrom<T, D>(source: Observable<T>, config?: LastValueFromConfig<D>) {
-  const hasConfig = arguments.length >= 2;
+  const hasConfig = typeof config === 'object';
   return new Promise<T | D>((resolve, reject) => {
     let _hasValue = false;
     let _value: T;
--- a/src/internal/lastValueFrom.ts
+++ b/src/internal/lastValueFrom.ts
@@ -1,7 +1,11 @@
 import { Observable } from './Observable';
 import { EmptyError } from './util/EmptyError';
 
-export function lastValueFrom<T, D>(source: Observable<T>, defaultValue: D): Promise<T | D>;
+export interface LastValueFromConfig<T> {
+  defaultValue: T;
+}
+
+export function lastValueFrom<T, D>(source: Observable<T>, config: LastValueFromConfig<D>): Promise<T | D>;
 export function lastValueFrom<T>(source: Observable<T>): Promise<T>;
 
 /**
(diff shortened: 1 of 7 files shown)


benlesh:
@cartant I think this should probably be in a config/options object. I say this because there's a strong possibility that this will support `AbortSignal` at some point:

```ts
const ac = new AbortController();

firstValueFrom(source$, { defaultValue: 2, signal: ac.signal });
```

benlesh (changes requested):
@cartant I think this should probably be in a config/options object. I say this because there's a strong possibility that this will support `AbortSignal` at some point:

```ts
const ac = new AbortController();

firstValueFrom(source$, { defaultValue: 2, signal: ac.signal });
```

cartant:
@benlesh

> I think this should probably be in a config/options object.

Nice catch. Done.

src/internal/lastValueFrom.ts
```diff
@@ -40,24 +48,28 @@ import { EmptyError } from './util/EmptyError';
  * ```
  *
  * @param source the observable to convert to a promise
+ * @param config a configuration object to define the `defaultValue` to use if the source completes without emitting a value
  */
-export function lastValueFrom<T>(source: Observable<T>) {
-  return new Promise<T>((resolve, reject) => {
+export function lastValueFrom<T, D>(source: Observable<T>, config?: LastValueFromConfig<D>) {
+  const hasConfig = arguments.length >= 2;
```
benlesh, line 54: For non-TS users, (or bad TS users) this can cause runtime errors: `lastValueFrom(promise, undefined)`

It doesn't really `hasConfig` unless `typeof config === 'object'`, I guess.

benlesh (changes requested):

src/internal/lastValueFrom.ts
benlesh: More realistically, I can see people doing things like:

```
const maybeConfig = externalConfig?.defaultValueConfig // maybe undefined

lastValueFrom(promise, maybeConfig);
```

It would be annoying to need to do this:

```
if (maybeConfig) {
  lastValueFrom(promise, maybeConfig);
} else {
  lastValueFrom(promise);
}
```

cartant: Nice catch. Done.

benlesh (approved):

PR #6207, opened 2021-04-03
chore(switchAll): convert switchAll tests to run mode

tmair:
**Description:**
This PR converts `switchAll` tests to run mode. It also replaces deprecated `subscribe` and `of` signatures within the tests. The deprecation removal has been done in its own commit to ease the review process.

None


cartant (approved):
LGTM. Thanks.

Issue #5824, opened 2020-10-13
Reinstating *some* of the resultSelectors


benlesh:
Currently we have deprecated resultSelectors on the following:

- bindCallback
- bindNodeCallback
- combineLatest
- forkJoin
- fromEvent
- fromEventPattern
- zip
- concatMap
- concatMapTo
- exhaustMap
- mergeMap
- mergeMapTo
- switchMap
- swithMapTo

The reasons for this were two fold:

1. It added size to every operator.
2. It meant that, in some cases, "outer values" from "outer sources" would need to be retained in memory for the lifetime of inner subscriptions. Causing memory pressure.

Given that I think recently the cost of 1 has gotten to be substantially less, I think we may want to revisit the operators where 2 is not the case (basically anything that isn't a map-and-flatten operator).

I'd propose that we can probably "undeprecate" the resultSelector argument for these creation functions:

- bindCallback
- bindNodeCallback
- combineLatest
- forkJoin
- fromEvent
- fromEventPattern
- zip

This is providing that they are not encumbered by issues 1 and 2 above. (I'm sure that 2 is not an issue, but 1 may or may or may not be true). I think this is especially okay if we're getting rid of "n-args" for sources for `forkJoin`, `combineLatest`, and (probably) `zip`.

benlesh:
Also worth noting: The secondary function/mapping argument is a little confusing to read, IMO: `mergeMap(fn, fn2)`.  If we can eliminate the confusing UX there, and mitigate 1 and 2, I might be okay with undeprecating all resultSelectors. Just a thought. 

While we have until v8 to make this decision, I think we should try to decide quickly so we can get our messaging across sooner rather than later.

niklas-wortmann:
I remember that @cartant wanted to discuss some of the deprecations as well, this is probably a good time to do this

PR #6208, opened 2021-04-04
fix(fromEvent): match targets properly; fix result selector type

cartant:
**Description:**

This PR fixes some problems with `fromEvent`:

- The interfaces used to match targets need to use methods instead of props. Otherwise, [`extends` behaves in an unintended manner](https://www.typescriptlang.org/play?ts=4.2.0-beta#code/JYOwLgpgTgZghgYwgAgPIgIIGcAKUD2ADsgN4BQyy+IAXMgBQhwC2EdWYUoA5sgD7IsAT2YAjfABsAlMgC8APmQA3fMAAmAbjIBfMmVCRYiFOmwBZCGAAW+NaQpUQjFm0Gce-QSPHS6K9Vq6+uDQ8EjIePhIWFj2lNTOrHQARBAAHsBgyVJ+qpo6emBChCgAommQIGq4BMSyEQTRsemV1WiYNUTIAPzIyQCaEITJyCkAckQQyVpFJcjlrVgW1rZyDVEQMcgtEFWxpkuWNna9A0Mj45PTQA) and the types won't match.
- There is a signature that takes only the target and the event name, so the optional qualifiers on the other signatures are redundant.
- If the `resultSelector` returns `T`, the target's type should not be `FromEventTarget<T>`.

The typing of `fromEvent` could be improved, but I think the first step is to fix these problems. The next would be to replace the `FromEventTarget` type with separate signatures.

### DIFF
--- a/src/internal/observable/fromEvent.ts
+++ b/src/internal/observable/fromEvent.ts
@@ -70,11 +70,11 @@ export interface AddEventListenerOptions extends EventListenerOptions {
 
 export function fromEvent<T>(target: FromEventTarget<T>, eventName: string): Observable<T>;
 /** @deprecated resultSelector no longer supported, pipe to map instead */
-export function fromEvent<T>(target: FromEventTarget<T>, eventName: string, resultSelector: (...args: any[]) => T): Observable<T>;
+export function fromEvent<T>(target: FromEventTarget<any>, eventName: string, resultSelector: (...args: any[]) => T): Observable<T>;
 export function fromEvent<T>(target: FromEventTarget<T>, eventName: string, options: EventListenerOptions): Observable<T>;
 /** @deprecated resultSelector no longer supported, pipe to map instead */
 export function fromEvent<T>(
-  target: FromEventTarget<T>,
+  target: FromEventTarget<any>,
   eventName: string,
   options: EventListenerOptions,
   resultSelector: (...args: any[]) => T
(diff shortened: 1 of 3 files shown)


benlesh:
> The interfaces used to match targets need to use methods instead of props. Otherwise, extends behaves in an unintended manner and the types won't match.

WTF. 🙄 lol

src/internal/observable/fromEvent.ts
```diff
@@ -70,11 +70,11 @@ export interface AddEventListenerOptions extends EventListenerOptions {
 
 export function fromEvent<T>(target: FromEventTarget<T>, eventName: string): Observable<T>;
 /** @deprecated resultSelector no longer supported, pipe to map instead */
-export function fromEvent<T>(target: FromEventTarget<T>, eventName: string, resultSelector?: (...args: any[]) => T): Observable<T>;
-export function fromEvent<T>(target: FromEventTarget<T>, eventName: string, options?: EventListenerOptions): Observable<T>;
+export function fromEvent<T>(target: FromEventTarget<any>, eventName: string, resultSelector: (...args: any[]) => T): Observable<T>;
```
benlesh, line 73: This seems to make sense. We're loosing up the type here, because really it's the `T` that matters. However, part of me feels like we should undeprecate this signature (as discussed in #5824), and in that case, we should try to get the types correct that are being passed to the result selector. I suppose we can do that in another pass though.

benlesh (approved):

src/internal/observable/fromEvent.ts
cartant: Yeah. That's something that can be done after it's split into separate signatures. ATM, `FromEventTarget` makes the function look simpler than it really is and the `T` is really only applicable to _some_ of the signatures.

PR #6211, opened 2021-04-08
fix(merge): single array is not an array of sources

cartant:
**Description:**

Prior to this change, the static `merge` behaved in a manner that was contrary to its types: it treated a single array as an array of sources. The static `concat` did not do this, so changing a `concat` to a `merge` or vice versa would effect wildly different behaviour if a single array was passed.

This PR adds a failing test and removes the call to `argsOrArgArray` - which did the sniffing and unpacking.

Regarding v7, there are a bunch of inconsistencies between `concat`, `merge`, `race`, `combineLatest`, `forkJoin` and `zip`, but most can be addressed through the application of deprecations. This change, however, addresses an inconsistency in the implementation.

### DIFF
--- a/spec/observables/merge-spec.ts
+++ b/spec/observables/merge-spec.ts
@@ -282,4 +282,10 @@ describe('merge(...observables, Scheduler, number)', () => {
 
     expect(e1Subscribed).to.be.false;
   });
+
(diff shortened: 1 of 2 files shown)


benlesh (approved):
Whoops.

Issue #5967, opened 2021-01-19
`observers` should be private on Subjects.


benlesh:
It was never meant as a means of introspecting or mutating what was going on within a Subject. IIRC, it was left `public` >6 years ago so we could access it within our own tests. That it was ever public is unfortunate.

Related #5961

benlesh:
After talking with @leggechr, I think so many people are using `observers` that we can only deprecate it for now. The use cases seem to mostly be around `if (subject.observers.length > 0) {  }`, or like "do you have observers? then do x, otherwise do y".

KamilSocha91:
@benlesh  @leggechr Is it already taken issue, I can create pull request and mark is as `deprecated`.

backbone87:
I would imagine the introduction of an `isObserved` property/getter/method could be a replacement.

dmitrii334:
I'm using following construction to check whether I can close an observed connection via sse (user could have multiple browser tabs opened and connected).

```
complete(id: string, completeAll = false) {
    const subject = this.subjectMap.get(id);

    if (subject && (completeAll || subject.observers.length < 2)) {
      subject.complete();
      this.subjectMap.delete(id);
    }
}
```

Seems like isObserved wouldn't do it for me. Are there any workarounds for such a case?

Jonathan002:
I find this property useful for unit test to check the subscriber count. Is there another recommended way to check subscriber count for test if this is deprecated?

rgolea:
I just found out this is going to be killed in the future. I was trying to build an operator that only emits to the last registered observer. Is there another way to do it?

This would be the exact example of it:

```
import { Observable, Subject, interval, take } from 'rxjs';

const triggerOnlyToLast = <T>() => {
  return (source$: Observable<T>) => {
    const subject$ = new Subject();

    const subscription = source$.subscribe({
      ...subject$,
      next: (val) => {
        if (!subject$.observed) {
          subscription.unsubscribe();
          return;
        }

        const last = subject$.observers.at(-1);
        last.next(val);
      },
    });

    const last = subject$.observers.at(-1);

    return subject$;
  };
};

const obs$ = interval(1000).pipe(take(5), triggerOnlyToLast());

obs$.subscribe((val) => console.log('From one:', val));
const sub = obs$.pipe(take(3)).subscribe((val) => {
  console.log('From two:', val);
});

```

sam-s4s:
[...]

PR #6215, opened 2021-04-11
chore: use @internal and _ for internals; remove protected modifiers

cartant:
**Description:**

Update: this PR now does what's listed in [this comment](https://github.com/ReactiveX/rxjs/pull/6215#issuecomment-818632903).

---

This PR removes the @deprecated tags from public/protected internals in favour of using @internal tags along with a underscore prefix - the latter was already used extensively throughout the codebase, so this just makes things more consistent, IMO, and means that people looking at runtime objects in the console or debugger will see properties and methods that are adorned with underscores that scream "THESE ARE INTERNAL".

Personally, I find that using the @deprecated tag for internals just makes the RxJS source harder to work with - as many things are shown struck out in VS Code. IMO, using @internal - there is an [ESLint rule](https://github.com/cartant/eslint-plugin-etc/blob/main/docs/rules/no-internal.md) that'll use that tag - and an underscore is more than enough to indicate to developers that something should not be relied upon or messed with.

I've not removed @deprecated from the `lift` API, that used to be public, albeit undocumented. And I've not removed it from the exported `Operator` and `Scheduler` types.

This PR also removes the `protected` modifiers that were re-introduced in the v7 refactor/rewrite. This avoids the problem that was addressed in #3544 - which _removed_ the `protected` modifiers. (TS Playground repro of the problem addressed in that issue [here](https://www.typescriptlang.org/play?ts=4.2.0-beta#code/MYGwhgzhAECC0G8BQ1rAPYDsIBcBOArsDungBQAOBARiAJbDSZgC2ApgFzS552YDmASkQBfJGNCQYAIUQo0WHkRLkqtBk1adu+PkNHikkqNADCc1Bmz5lpSnnQ42xNgBNN7Ljz3CEYieAmACIWCtaExHYUDk4u7syeOrwCvv5GijjQYFzwALxMbADu0NJkAETUZYIA3NAA9HUKLBRgOHS0bOnWaFzm+ZhF0EHlrlW1DUyOTS1tHUA).) I've not removed them from types that are not exported - like the scheduler actions.

Whether or not the removal of `protected` is the what ought to be done, IDK. I've removed it in case it was added without what was done in #3544 being considered. However, IMO, I think removing it - inline with what was done in #3544 - will be the least problematic option. ATM, in v7, the `lift` API is `protected` and I think if it's left like that, problems are going to be effected, as `Observable` is a type that's likely to be exported by user-land packages.

**Related issue (if exists):** #5967 (kinda)

### DIFF
--- a/src/internal/Subject.ts
+++ b/src/internal/Subject.ts
@@ -14,15 +14,15 @@ import { arrRemove } from './util/arrRemove';
  * Subject, and you can call next to feed values as well as error and complete.
  */
 export class Subject<T> extends Observable<T> implements SubscriptionLike {
-  observers: Observer<T>[] = [];
-
   closed = false;
-
-  isStopped = false;
-
-  hasError = false;
-
-  thrownError: any = null;
+  /** @internal */
+  _observers: Observer<T>[] = [];
+  /** @internal */
+  _isStopped = false;
+  /** @internal */
+  _hasError = false;
+  /** @internal */
+  _thrownError: any = null;
 
   /**
    * Creates a "subject" by basically gluing an observer to an observable.
@@ -39,12 +39,14 @@ export class Subject<T> extends Observable<T> implements SubscriptionLike {
     super();
   }
 
+  /** @internal */
   lift<R>(operator: Operator<T, R>): Observable<R> {
     const subject = new AnonymousSubject(this, this);
     subject.operator = operator as any;
     return subject as any;
   }
 
+  /** @internal */
   protected _throwIfClosed() {
     if (this.closed) {
       throw new ObjectUnsubscribedError();
@@ -53,8 +55,8 @@ export class Subject<T> extends Observable<T> implements SubscriptionLike {
 
   next(value: T) {
     this._throwIfClosed();
-    if (!this.isStopped) {
-      const copy = this.observers.slice();
+    if (!this._isStopped) {
+      const copy = this._observers.slice();
       for (const observer of copy) {
         observer.next(value);
       }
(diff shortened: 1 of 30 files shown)


benlesh:
Ugghhhhh.... This sucks. I see what you're doing, I agree with what you're doing. And I hate it. All at the same time.

Shouldn't the `@internal` annotation prevent these from being emitted, though? What does it look like in the IDE? I rather liked that people got an annoying "slash" through things they shouldn't be using. I'd even more rather it be a type error, TBH (hence `protected`). 

Also, #3544 was a long time ago. Are we sure this is still an issue?

cartant:
> ... prevent these from being emitted, though?

There is a TS thing that relates to not emitting internals, but it's not supposed to be used and using causes problems.

> Are we sure this is still an issue?

Yeah, the TS Playground shows that it will be.

src/internal/Subject.ts
```diff
@@ -14,15 +14,15 @@ import { arrRemove } from './util/arrRemove';
...
-
   closed = false;
-
-  isStopped = false;
-
-  hasError = false;
-
-  thrownError: any = null;
+  /** @internal */
+  _observers: Observer<T>[] = [];
```
benlesh, line 19: There are a **LOT** of people that are checking `subject.observers.length === 0` or `> 0`, for better or worse. It's rare that people are checking anything other than whether or not the subject is being observed. I don't think we can rename `observers` to `_observers` without deprecating it first.

We may want to add a feature like `isObserved` or something like that on our Subjects, given how often I've seen code that needed to do that, because whether I agree with the need or not, other people have found the need for this.

benlesh (changes requested):
Along with my other comment about `observers`, specifically, we really need to think about this.

The problem with what I see here is we're renaming things (rather than just affecting their TypeScript visibility), which is a run-time breaking change.

For any JavaScript (non-TS) user, the API that they thought was theirs to use, `isStopped` for example, is now suddenly `_isStopped`.  Which, to some degree is _fine_ but we're going to need to document it.

Some part of me really wants to make sure that #3544 is even an issue in 2021, given that it was dealt with back in 2018. On the surface, this PR looks like we're moving the wrong direction, if we don't have that context.

benlesh:
> Yeah, the TS Playground shows that it will be.

I'm stupid, where is the playground? I don't see it.

benlesh:
I feel like what I've learned from this is that `protected` is a joke when it comes to writing libraries, and shouldn't be used.

cartant:
> (TS Playground repro of the problem addressed in that issue [here](https://www.typescriptlang.org/play?ts=4.2.0-beta#code/MYGwhgzhAECC0G8BQ1rAPYDsIBcBOArsDungBQAOBARiAJbDSZgC2ApgFzS552YDmASkQBfJGNCQYAIUQo0WHkRLkqtBk1adu+PkNHikkqNADCc1Bmz5lpSnnQ42xNgBNN7Ljz3CEYieAmACIWCtaExHYUDk4u7syeOrwCvv5GijjQYFzwALxMbADu0NJkAETUZYIA3NAA9HUKLBRgOHS0bOnWaFzm+ZhF0EHlrlW1DUyOTS1tHUA).)

benlesh:
As a more moderate maneuver, could we only rename the properties we had deprecated in 6.x, and deprecate the properties we hadn't renamed yet for v8? Then we can remove/rename those in v8 without as many issues as doing it right now.

cartant:
> As a more moderate maneuver, could we only rename the properties we had deprecated in 6.x, and deprecate the properties we hadn't renamed yet for v8?

Sure. This PR rolls up several ideas. We can pick and choose.

cartant:
> I feel like what I've learned from this is that protected is a joke when it comes to writing libraries, and shouldn't be used.

Just go all the way: OO is a joke. 😅

benlesh:
Another alternative: What if we keep the properties `protected`, but use an `interface` of the public properties anywhere we're accepting these things? It seems like that's the better thing to do? I'm probably wrong, but I want to know why I'm wrong.

cartant:
@benlesh
 
> As a more moderate maneuver, could we only rename the properties we had deprecated in 6.x, and deprecate the properties we hadn't renamed yet for v8?

I found [the comment](https://github.com/ReactiveX/rxjs/pull/3981#discussion_r206964114) that I'd remembered and, AFAICT, we should do this:

- Tag previously deprecated, internal APIs as `@internal`. And ensure they're prefixed with underscores (already done in this PR) as that is the convention that api-extractor insists upon.
- Use [`--stripInternal`](https://www.typescriptlang.org/tsconfig/#stripInternal) to remove them from our `.d.ts` files. (Unfortunately, we cannot use api-extractor because of [this limitation](https://api-extractor.com/pages/setup/configure_rollup/#an-important-limitation).)
- Revert the removal of `protected` from APIs that are tagged with `@internal`, but any non-`@internal` APIs that were protected prior to this PR will have to remain public.

I'll look at making these changes.

src/internal/observable/ConnectableObservable.ts
```diff
@@ -37,6 +37,7 @@ export class ConnectableObservable<T> extends Observable<T> {
     }
   }
 
+  /** @internal */
```
cartant, line 40: I tagged `_subscribe` as internal 'cause it was tagged in many other places and this was done as part of a search/replace thing. I've ignored the other protected members in this class 'cause the entire class is deprecated.

cartant:
I wanted to double check that the `protected` thing has been addressed, so I had a quick peak at the `api_guardian`-generated files and, AFAICT, the only remaining uses of `protected` are in `ConnectableObservable`, `Subscriber` (the ctor of which is deprecated), and `VirtualAction`. So we should have no `protected`-related, multi-package, type interop issues now that `@internal` has been used.

benlesh (approved):

src/internal/Subject.ts
Gillardo: What is the way around this now, as i cannot seem to find an isObserved method?

PR #6217, opened 2021-04-11
docs: fix instruction demos in tap operator

willin:
**Description:**

there are several errors in the origin documentation, and i fixed them, already tested.

### DIFF
--- a/src/internal/operators/tap.ts
+++ b/src/internal/operators/tap.ts
@@ -88,9 +88,9 @@ export function tap<T>(
  *  concatMap(n => interval(1000).pipe(
  *    take(Math.round(Math.random() * 10)),
  *    map(() => 'X'),
- *    tap(n => ({
+ *    tap({
  *      complete: () => console.log(`Done with ${n}`)
- *    }))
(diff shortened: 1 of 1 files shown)


src/internal/operators/tap.ts
```diff
@@ -81,16 +81,16 @@ export function tap<T>(
...
  * of(1, 2, 3).pipe(
- *  concatMap(n => interval.pipe(
+ *  concatMap(n => interval(1000).pipe(
  *    take(Math.round(Math.random() * 10)),
  *    map(() => 'X'),
- *    tap({
+ *    tap(n => ({
  *      complete: () => console.log(`Done with ${n}`)
- *    })
+ *    }))
```
cartant, line 93: This last bit isn't an error. An observer with only a `complete` method is being passed.

cartant (changes requested):
Thanks. You've fixed a bunch of problems, but one of the changes is unnecessary.

cartant (approved):
LGTM. Thanks.

PR #6218, opened 2021-04-11
chore(skipWhile): convert skipWhile tests to run mode

tmair:
**Description:**
This PR converts `skipWhile` tests to run mode. It also refactors a deprecated `tap` signature.

None


cartant (approved):
LGTM. Thanks.

PR #6219, opened 2021-04-11
docs: remove unwanted word here

santoshyadavdev:
**Description:**


benlesh:
The original verbage was correct. It's a link to a location. Like [find Google here](https://www.google.com). 

Thank you for the thought, but we'll keep it as-is.

PR #6220, opened 2021-04-11
docs: only here should redirect

santoshyadavdev:
**Description:**


benlesh:
This seems to be duplicated in #6222.

PR #6222, opened 2021-04-11
docs: highlight the delay operator

santoshyadavdev:
**Description:**

### DIFF
--- a/docs_app/content/guide/operators.md
+++ b/docs_app/content/guide/operators.md
@@ -68,7 +68,7 @@ import { interval } from 'rxjs';
 const observable = interval(1000 /* number of milliseconds */);
 ```
 
-See the list of [all static creation operators here](#creation-operators-list).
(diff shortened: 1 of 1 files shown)


benlesh:
Hi @santoshyadavdev, while I appreciate the effort. I'm unsure this PR provides a ton of value. I'll leave it open for a bit to see what other core team members think.

santoshyadavdev:
Ahh no issues, I was reading the doc and found it somewhat incorrect, feel free to close if it's not needed, don't want to spam 😊

cartant (approved):
Approving this 'cause it does no harm. However, it doesn't really add anything and I agree that PRs that make changes like this aren't really something we want to encourage. Really, it takes more time trying to decide whether or not a change like this should be accepted than it does for a change the fixes a clear grammar, link or spelling error in the docs.

PR #6223, opened 2021-04-12
chore: move Marble Testing docs one folder up

jakovljevic-mladen:
**Description:**
Since Marble Testing docs are now the only document in the `/guide/testing` folder (as of #5989), I wanted to move it one folder up. Basically, it will go to `/guide` folder where all other documents are based. I also updated navigation.

I have tried to set up Firebase redirect so that old URLs keep working (there are a lot of sites pointing to https://rxjs.dev/guide/testing/marble-testing), but I don't know how to test this. Also, I'm not sure how all of this will work in combination with Angular because Angular app sends a request to `https://rxjs.dev/generated/docs/guide/testing/marble-testing.json`, so this `/generated` part and this `.json` part do not seem to be covered by this rule.

Anyway, I would like to test if this could work in any way so that we don't have `/testing` part in the URL, but that old URL is still doing some redirect to the new route.

None


jakovljevic-mladen:
I'm thinking about how this works and I think this setup should work fine. Let me try to break down what will possibly happen if this gets merged:
- When a users clicks a link or a bookmark that goes to https://rxjs.dev/guide/testing/marble-testing, the browser sends a request to the Firebase server;
- when Firebase receives a request, it should return Angular's `index.html`, but since there is a rule set up in the `firebase.json` file that defines a redirect from `/guide/testing/marble-testing` route to `/guide/marble-testing`, Firebase will return 301 with `Location` header pointing to `https://rxjs.dev/guide/marble-testing`
- browser will then send another request to Firebase to https://rxjs.dev/guide/marble-testing
- Firebase will now return the `index.html` and Angular will now know how to get the proper `.json` file as it will be located at `https://rxjs.dev/generated/docs/guide/marble-testing.json`

We just have to make sure that there is no active link inside Angular app that is pointing to `/guide/testing/marble-testing` and I think I removed all of them with this PR.

benlesh:
I'm not sure the work incurred is worth the value provided, in this case. There's a strong possibility we'll want to add more documentation around testing in the future, and if that's the case, we'd just end up moving it back. While I agree with the spirit of this PR, I don't think we should merge it, TBH.

@niklas-wortmann can reopen, of course, if he disagrees.

jakovljevic-mladen:
> There's a strong possibility we'll want to add more documentation around testing in the future

If this is the case, I agree this should stay. I never thought that might be the case. Thanks!

PR #6224, opened 2021-04-12
fix(symbol): revert unique symbol in #5874

### LINKED ISSUE #5919: 7.0.0-beta.9 conflicts symbol.observable to redux (and other have old symbol-observable version)
## Bug Report

**Current Behavior**
A clear and concise description of the behavior.

repro at https://github.com/kwonoj/rxjs-symbol-test, `npm i && npm run build` will raise compilation error.

image

https://github.com/ReactiveX/rxjs/pull/5874 introduces changes to symbol, which also updated in `symbol-observable` to support this (https://github.com/benlesh/symbol-observable/blob/master/CHANGELOG.md#300-2020-11-02). But some of packages, for examble redux (https://github.com/reduxjs/redux/blob/v4.0.5/package.json#L53) relies on older version of `symbol-observable` conflicts with latest rxjs changes.

This is particularly tricky as there's no easy workaround instead of trying to mutate module resolution somehow.

**Expected behavior**
A clear and concise description of what you expected to happen (or code).

**Environment**
- Runtime: [e.g. Node v${x}, Chrome v${x}]
- RxJS version:
- (If bug is related) Loader, build configuration: [e.g webpack, angular-cli version, config]

**Possible Solution**

Add any other context about the problem here. If applicable, add screenshots to help explain.

kwonoj:
- closes #5919

This is manual revert to #5874. 

Core team have discussed around #5919 for a while, and at this moment pretty much agreed

- there is no good suggestion to resolve all of problems we are facing
- current changes (#5784) is too much breaking especially we don't have good workaround to suggest.

At least, providing as-is experience between v6 to v7 is something we'd like to have instead of making huge breaking changes. This doesn't mean we'll stay with this forever, it's just hard enough find right solution (and doesn't break whole world).

I'm making this PR to v7 not slip this accidentally when we publish official release.

**Description:**

### DIFF
--- a/src/internal/symbol/observable.ts
+++ b/src/internal/symbol/observable.ts
@@ -1,2 +1,11 @@
+/** Symbol.observable addition */
+/* Note: This will add Symbol.observable globally for all TypeScript users,
+  however, we are no longer polyfilling Symbol.observable */
+declare global {
+  interface SymbolConstructor {
+    readonly observable: symbol;
+  }
+}
+
 /** Symbol.observable or a string "@@observable". Used for interop */
 export const observable = (() => (typeof Symbol === 'function' && Symbol.observable) || '@@observable')();
--- a/src/internal/symbol/observable.ts
+++ b/src/internal/symbol/observable.ts
@@ -1,11 +1,2 @@
-/** Symbol.observable addition */
-/* Note: This will add Symbol.observable globally for all TypeScript users,
-  however, we are no longer polyfilling Symbol.observable */
-declare global {
-  interface SymbolConstructor {
-    readonly observable: symbol;
-  }
-}
-
 /** Symbol.observable or a string "@@observable". Used for interop */
 export const observable = (() => (typeof Symbol === 'function' && Symbol.observable) || '@@observable')();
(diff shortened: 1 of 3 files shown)


src/internal/symbol/observable.ts
```diff
@@ -1,2 +1,11 @@
+/** Symbol.observable addition */
+/* Note: This will add Symbol.observable globally for all TypeScript users,
+  however, we are no longer polyfilling Symbol.observable */
+declare global {
+  interface SymbolConstructor {
+    readonly observable: symbol;
+  }
+}
+
```
cartant, line 9: I don't think we should revert the moving of this declaration, as its being moved solved a problem. See this PR that made the same change to the 6.x branch and the issue that it references: https://github.com/ReactiveX/rxjs/pull/6178

benlesh:
Oh... no, I'm just stupid. Thanks, @cartant and @kwonoj

benlesh (changes requested):
@kwonoj if you can move what you have back into `types.ts` I think this is GTG.

src/internal/symbol/observable.ts
kwonoj: I cherrypicked specific PR https://github.com/ReactiveX/rxjs/pull/6224/commits/ed34f007a7ba76ea6d1383ca50a6f62aa61e29f4

cartant (approved):

benlesh (approved):

PR #6227, opened 2021-04-14
fix: forkJoin/combineLatest return Observable<unknown> if passed any

### LINKED ISSUE #6226: Types: `forkJoin(any)` and `combineLatest(any)` return the wrong thing
```ts
// FAIL: Gives us Observable<{ [key: string]: unknown }>
const source1 = forkJoin([of(1), of(2)] as any); // $ExpectType Observable<unknown>

// FAIL: Gives us Observable<{ [key: string]: unknown }>
const source2 = combineLatest([of(1), of(2)] as any); // $ExpectType Observable<unknown>
```

Basically, if we get an `any` to either one of these, we have no way of knowing if we're getting back an array or an object, so we need to return `unknown`.

benlesh:
Resolves #6226

### DIFF
--- a/src/internal/AnyCatcher.ts
+++ b/src/internal/AnyCatcher.ts
@@ -1,3 +1,3 @@
 export interface AnyCatcher {
-  '@@@@@@CATCH_ANY_WITH_THIS@@@@@@@@': 'wat';
+  readonly ' @@@@@@CATCH_ANY_WITH_THIS@@@@@@@@': 'wat';
 }
--- a/src/internal/AnyCatcher.ts
+++ b/src/internal/AnyCatcher.ts
@@ -1,0 +1,3 @@
+export interface AnyCatcher {
+  '@@@@@@CATCH_ANY_WITH_THIS@@@@@@@@': 'wat';
+}
--- a/src/internal/AnyCatcher.ts
+++ b/src/internal/AnyCatcher.ts
@@ -6,6 +6,5 @@
  *
  * @internal
  */
-export interface AnyCatcher {
-  readonly ' @@@@@@CATCH_ANY_WITH_THIS@@@@@@': false;
-}
+declare const anyCatcherSymbol: unique symbol;
+export type AnyCatcher = typeof anyCatcherSymbol;
--- a/src/internal/AnyCatcher.ts
+++ b/src/internal/AnyCatcher.ts
@@ -1,3 +1,11 @@
+/**
+ * This is just a type that we're using to identify `any` being passed to
+ * function overloads. This is used because of situations like {@link forkJoin},
+ * where it could return an `Observable<T[]>` or an `Observable<{ [key: K]: T }>`,
+ * so `forkJoin(any)` would mean we need to return `Observable<unknown>`.
+ *
+ * @internal
+ */
 export interface AnyCatcher {
-  readonly ' @@@@@@CATCH_ANY_WITH_THIS@@@@@@@@': 'wat';
+  readonly ' @@@@@@CATCH_ANY_WITH_THIS@@@@@@': false;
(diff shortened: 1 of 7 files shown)


src/internal/AnyCatcher.ts
```diff
@@ -0,0 +1,3 @@
+export interface AnyCatcher {
```
benlesh, line 1: LOLOLOLOLOLOLOLOL 🤣 🎉

src/internal/AnyCatcher.ts
```diff
@@ -0,0 +1,3 @@
...
+  '@@@@@@CATCH_ANY_WITH_THIS@@@@@@@@': 'wat';
```
DanielRosenwasser, line 2:
```suggestion
  readonly ' @@@@@@CATCH_ANY_WITH_THIS@@@@@@@@': 'wat';
```

src/internal/observable/combineLatest.ts
```diff
@@ -47,6 +48,12 @@ export function combineLatest<A extends readonly unknown[]>(
 
 // combineLatest({a, b, c})
 export function combineLatest(sourcesObject: { [K in any]: never }): Observable<never>;
+/**
+ * You have passed `any` here, we can't figure out if it is
+ * an array or an object, so you're getting `unknown`. Use better types.
+ * @param arg Something typed as `any`
+ */
+export function combineLatest<T extends AnyCatcher>(arg: T): Observable<unknown>;
```
cartant, line 56: suggestion: I would make this the first signature. IDK why `any` skips all of the signatures above this one and I would not like to rely upon that behaviour never changing. There should be no reason this signature cannot be the first.

src/internal/observable/forkJoin.ts
```diff
@@ -27,6 +28,12 @@ export function forkJoin<A extends readonly unknown[], R>(
 
 // forkJoin({a, b, c})
 export function forkJoin(sourcesObject: { [K in any]: never }): Observable<never>;
+/**
+ * You have passed `any` here, we can't figure out if it is
+ * an array or an object, so you're getting `unknown`. Use better types.
+ * @param arg Something typed as `any`
+ */
+export function forkJoin<T extends AnyCatcher>(arg: T): Observable<unknown>;
```
cartant, line 36: suggestion: Make this the first overload signature.

src/internal/AnyCatcher.ts
cartant: suggestion: This could be something like this, instead:

```ts
declare const anyCatcherSymbol: unique symbol;
type AnyCatcher = typeof anyCatcherSymbol;
```

[TypeScript Playground demo](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF3lFBPAwlBmABYgwDKuAtgEZIQBc8ArigJYCOLCaN9EANwBYAFAZcABwQBBPIWJkY8ALzwJ0pADNs8oqXJU6DEaLGhIsBFrZgM7VPC0AeACrwQADwwgUwNPByBPpKAHwAFLAA5syuAJTMbADWKEgA7iimFtBwTrb2ji6uEdGxCfCupmLI6FhQqk7hAERQTdgBOLhxgvAA9L2sKCnpKNWomPC0DVrNtE3dfQNNc0A)

benlesh: Yeah, that's what I was saying on Slack, but it's SO BORING. Haha. Okay fine.

src/internal/observable/combineLatest.ts
benlesh: That's a really good idea!

cartant (approved):

PR #6228, opened 2021-04-15
fix: Symbol.observable type no longer `unique symbol`.

benlesh:
This is to fix compatibility issues with our type definitions and those from other projects. Of note: `symbol-observable@4` as also reverted this typing to just `symbol` from `unique symbol`.


cartant (approved):
Does this supercede https://github.com/ReactiveX/rxjs/pull/6224 ?

kwonoj:
It sounds like so?

benlesh:
@cartant no no... I'm stupid. sorry. 🤦 

I forgot about @kwonoj's PR because I'm absent minded. 😅

Issue #6067, opened 2021-03-01
[v7] TS automatic imports in VSCode default to wrong path


voliva:
## Bug Report

**Current Behavior**
When using operators in VSCode, the suggested import defaults to `rxjs/dist/types/operators`, failing in runtime.

This wasn't happening in v6, starts happening after upgrading to v7.

**Expected behavior**
VSCode imports from the correct `rxjs/operators` path

**Reproduction**

* create a simple playground
```sh
mkdir rxjs-playground
cd rxjs-playground
npm init -y
npm i rxjs@next
```

* open that folder in VSCode
* create a file foo.ts

```ts
interval(1000).pipe( // Correctly suggests `import { interval } from 'rxjs'`
   map(v => v*2) // Doesn't suggest any valid import, that's fine - it used to do that. Manually import it from `rxjs/operators`
)
```

* create a file bar.ts and type in:

```ts
interval(1000).pipe( // Correctly suggests `import { interval } from 'rxjs'`
   map(v => v*2) // Wrongly suggests `import { map } from 'rxjs/dist/types/operators'`
)
```

**Environment**
- RxJS version: 7.0.0-beta.12
- VSCode: 1.53.2
- Node: 12.18.3
- Typescript: 4.1.5

**Additional context/Screenshots**

![image](https://user-images.githubusercontent.com/5365487/109515914-eaa8f300-7aa7-11eb-8faf-73dd16ba54f4.png)

Note that after correcting it to `rxjs/operators` it correctly suggests adding the import from the right place

![image](https://user-images.githubusercontent.com/5365487/109516077-15934700-7aa8-11eb-9214-8738f6b82eb2.png)

benlesh:
This issue in TypeScript seems related: https://github.com/microsoft/TypeScript/issues/30472

I'm not really sure what the resolution to this is. If it's on our end, or if it's in your tsconfig.

@cartant @RyanCavanaugh any ideas?

benlesh:
@voliva do you have to same issues with RxJS 6.x?

benlesh:
One primary difference I see between 6.x and 7.x is in 6.x we have the CJS root and typings root right next to each other. Where they are not next to each other in 7.x:

6.x package.json:

```json
{
"main": "./index.js",
"typings": "./index.d.ts",
"module": "./_esm5/index.js",
"es2015": "./_esm2015/index.js"
}
```

7.x package.json:

```json
{
"main": "./dist/cjs/index.js",
[...]

PR #6229, opened 2021-04-16
chore: Add typesVersions to package.json

benlesh:
This is an attempt to fix issues with VS Code auto-import, as described here in this issue: https://github.com/microsoft/TypeScript/issues/43034\#issuecomment-820621280

Related #6067

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,9 @@
   "module": "./dist/esm5/index.js",
   "es2015": "./dist/esm/index.js",
   "types": "./dist/types/index.d.ts",
+  "typesVersions": {
+    ">=4.0": { "*": [ "dist/types/*" ] }
+  },
   "sideEffects": false,
   "exports": {
     ".": {


package.json
```diff
@@ -6,6 +6,9 @@
   "module": "./dist/esm5/index.js",
   "es2015": "./dist/esm/index.js",
   "types": "./dist/types/index.d.ts",
+  "typesVersions": {
+    ">=4.0": { "*": [ "dist/types/*" ] }
```
cartant, line 10: question: Could we use this to express that the minimum supported version is 4.2? I mean, we're not going to support anything older than that.

cartant (approved):

package.json
benlesh: Probably?

benlesh:
This didn't appear to work at all. We can probably revert it. :\

benlesh:
Shit. I wanted to get this into the RC :\

PR #6237, opened 2021-04-19
types(ajax): add HTTP Request methods

cenguidanos:
**Description:**
Adds types to AjaxRequest and AjaxConfig interfaces that uses http **method** property.

Noop

### DIFF
--- a/api_guard/dist/types/ajax/index.d.ts
+++ b/api_guard/dist/types/ajax/index.d.ts
@@ -33,6 +33,8 @@ export interface AjaxError extends Error {
 
 export declare const AjaxError: AjaxErrorCtor;
 
+export declare type AjaxHttpMethod = 'GET' | 'HEAD' | 'POST' | 'PUT' | 'DELETE' | 'CONNECT' | 'OPTIONS' | 'TRACE' | 'PATCH' | '';
+
 export interface AjaxRequest {
     async: boolean;
     body?: any;
--- a/api_guard/dist/types/ajax/index.d.ts
+++ b/api_guard/dist/types/ajax/index.d.ts
@@ -8,7 +8,7 @@ export interface AjaxConfig {
     headers?: Readonly<Record<string, any>>;
     includeDownloadProgress?: boolean;
     includeUploadProgress?: boolean;
-    method?: string;
+    method?: AjaxHttpMethod;
     password?: string;
     progressSubscriber?: PartialObserver<ProgressEvent>;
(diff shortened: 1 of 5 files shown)


api_guard/dist/types/ajax/index.d.ts
```diff
@@ -33,12 +33,14 @@ export interface AjaxError extends Error {
 
 export declare const AjaxError: AjaxErrorCtor;
 
+export declare type AjaxHttpMethod = 'GET' | 'HEAD' | 'POST' | 'PUT' | 'DELETE' | 'CONNECT' | 'OPTIONS' | 'TRACE' | 'PATCH' | '';
```
benlesh, line 36:
```suggestion
/**
 * A valid HTTP method as defined in the registry in RFC-7231 by the IETF, Section 8.1.3
 * https://tools.ietf.org/html/rfc7231#section-8.1.3
 */
export declare type HttpMethod = 'GET' | 'HEAD' | 'POST' | 'PUT' | 'DELETE' | 'CONNECT' | 'OPTIONS' | 'TRACE' | 'PATCH';
```

I think we probably shouldn't allow `''` here. Also I wanted to add a comment that points to the specification so people know we're not just pulling these out of our butts.

benlesh (changes requested):

benlesh:
@cartant @kwonoj This is _technically_ a breaking change, and we're in RC. It's breaking because `'get'` or `'post'` would start failing in TS builds. I'm sort of on the fence here. I sort of doubt people need guide rails in this case, so we might be able to just not merge this one for v7 but consider it for the next major.

src/internal/ajax/types.ts
```diff
@@ -1,5 +1,12 @@
 import { PartialObserver } from '../types';
 
+/**
```
cenguidanos, line 3: Hi @benlesh, the comment is here, on src/internal/ajax/types.ts

api_guard/dist/types/ajax/index.d.ts
cenguidanos: You are right, i added [here](https://github.com/ReactiveX/rxjs/pull/6237/files/886a69c674ef783a682bf3a2d7553622ecb46435#r616369336)

benlesh:
Okay, after some thought, I don't think we can accept this PR. While you're _technically_ correct that these are the only "allowed" methods. People are actually able to send any arbitrary method they want, and they're equally able to handle those arbitrary methods on their server. So given that, and given that *most* people aren't going to accidentally type the wrong thing here, and given that lowercase is still totally valid, I don't think we can accept this PR.

Thank you so much for your time and effort, @cenguidanos!

PR #6238, opened 2021-04-19
feat: use optional config.Promise in firstValueFrom and lastValueFrom

jakovljevic-mladen:
**Description:**
This PR enables `firstValueFrom` and `lastValueFrom` to construct a Promise if `config.Promise` is defined.

None


benlesh:
This is a thoughtful PR, but we'd like to get away from globally configured Promise. That is a legacy feature from 6-7 years ago. We're planning on removing it in v8, and I don't want to add it to new features just to remove it.

Thank you for your effort, @jakovljevic-mladen, but we can't accept this one.

PR #6233, opened 2021-04-18
types(ajax): improve Ajax types and docs

jakovljevic-mladen:
**Description:**
This PR improves `ajax` docs, introduces `AjaxDirection` type and fixes multiple typos.

Also, some `npm run docs` failures are fixed with this PR, like:
- `No handler provided for inline tag "{@see {@link AjaxConfig}"` or
- `Error in {@link AjaxProgressEvents} - Invalid link (does not match any doc)`.

**Related issue (if exists):**
None

PR #6247, opened 2021-04-21
types(config): add GlobalConfig interface

jakovljevic-mladen:
**Description:**
The reason why I wanted to add an interface for `config` object is so that we can have a proper documentation page for it. Constant documentation pages don't have a list of properties like interface documentation pages do.
Also, by introducing this interface, we will fix an error while building the docs (`npm run docs`) from this line:

https://github.com/ReactiveX/rxjs/blob/23bc7fdc16acd76dd71a4faf57b8b949684c3249/src/internal/Observable.ts#L101

because a link to `onUnhandledError` will now be available.

I was thinking about naming this interface with just as `Config`, but I chose `GlobalConfig` in the end. If needed, I can change to `Config` if it feels more appropriate.

None

### DIFF
--- a/src/internal/config.ts
+++ b/src/internal/config.ts
@@ -1,71 +1,18 @@
 import { Subscriber } from './Subscriber';
-import { ObservableNotification } from './types';
+import { ObservableNotification, GlobalConfig } from './types';
 
 /**
- * The global configuration object for RxJS, used to configure things
- * like what Promise constructor should used to create Promises
+ * The {@link GlobalConfig} object for RxJS. It is used to configure things
+ * like how to react on unhandled errors.
  */
-export const config = {
-  /**
-   * A registration point for unhandled errors from RxJS. These are errors that
-   * cannot were not handled by consuming code in the usual subscription path. For
-   * example, if you have this configured, and you subscribe to an observable without
-   * providing an error handler, errors from that subscription will end up here. This
-   * will _always_ be called asynchronously on another job in the runtime. This is because
-   * we do not want errors thrown in this user-configured handler to interfere with the
-   * behavior of the library.
-   */
+export const config: GlobalConfig = {
   onUnhandledError: null as ((err: any) => void) | null,
 
-  /**
-   * A registration point for notifications that cannot be sent to subscribers because they
-   * have completed, errored or have been explicitly unsubscribed. By default, next, complete
-   * and error notifications sent to stopped subscribers are noops. However, sometimes callers
-   * might want a different behavior. For example, with sources that attempt to report errors
-   * to stopped subscribers, a caller can configure RxJS to throw an unhandled error instead.
-   * This will _always_ be called asynchronously on another job in the runtime. This is because
-   * we do not want errors thrown in this user-configured handler to interfere with the
-   * behavior of the library.
-   */
   onStoppedNotification: null as ((notification: ObservableNotification<any>, subscriber: Subscriber<any>) => void) | null,
 
-  /**
-   * The promise constructor used by default for methods such as
-   * {@link toPromise} and {@link forEach}
-   *
-   * @deprecated remove in v8. RxJS will no longer support this sort of injection of a
-   * Promise constructor. If you need a Promise implementation other than native promises,
-   * please polyfill/patch Promises as you see appropriate.
-   */
   Promise: undefined as PromiseConstructorLike | undefined,
 
-  /**
-   * If true, turns on synchronous error rethrowing, which is a deprecated behavior
-   * in v6 and higher. This behavior enables bad patterns like wrapping a subscribe
-   * call in a try/catch block. It also enables producer interference, a nasty bug
-   * where a multicast can be broken for all observers by a downstream consumer with
-   * an unhandled error. DO NOT USE THIS FLAG UNLESS IT'S NEEDED TO BUY TIME
-   * FOR MIGRATION REASONS.
-   *
-   * @deprecated remove in v8. As of version 8, RxJS will no longer support synchronous throwing
-   * of unhandled errors. All errors will be thrown on a separate call stack to prevent bad
-   * behaviors described above.
-   */
   useDeprecatedSynchronousErrorHandling: false,
 
-  /**
-   * If true, enables an as-of-yet undocumented feature from v5: The ability to access
-   * `unsubscribe()` via `this` context in `next` functions created in observers passed
-   * to `subscribe`.
-   *
-   * This is being removed because the performance was severely problematic, and it could also cause
-   * issues when types other than POJOs are passed to subscribe as subscribers, as they will likely have
-   * their `this` context overwritten.
-   *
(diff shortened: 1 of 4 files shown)


benlesh:
@niklas-wortmann it seems odd that we have to introduce a type we're going to publish to the masses just to get documentation here. Is there a better way to fix this?

benlesh:
I guess a counter argument is that we're introducing a type either way. 🤔 I'm on the fence here. @cartant? @kwonoj?

cartant:
I think this is fine, but I don't see why it needs to be in types and I don't see why it needs to be exported. Does the docs generation really require the latter?

jakovljevic-mladen:
> Does the docs generation really require the latter?

@cartant I think so. When I was working on #6233, I couldn't make `AjaxDirection` show on the list up until I [exported it](https://github.com/ReactiveX/rxjs/pull/6233/files#diff-4ccc4965fd6e88c24e8a5ecd7dbebe16b728c06d69998c5ef245b0aa494fa74aR4).

cartant:
> ... it seems odd that we have to introduce a type we're going to publish to the masses just to get documentation here.

I'm more surprised that the type has to be exported, TBH.

jakovljevic-mladen:
If I move `GlobalConfig` to `config.ts` file (with or without exporting the interface), I get this error: `Error in {@link GlobalConfig} - Invalid link (does not match any doc): "GlobalConfig" - doc "index/config"`. This is basically [this line](https://github.com/ReactiveX/rxjs/pull/6247/files#diff-1efe38be0ed095b15c7a06573eef319a4485140095bfd54d951c28eb5b46f187R5) failing.

cartant:
My preference would be to try to figure out why the docs generation requires this, as, IIRC, it's legal to use parameters (for example) in an exported API that have a `type` that is not exported - the type alias is expanded.  TBH, I think this is done in several places, so I guess those doc's don't work either?

jakovljevic-mladen:
I could try playing with the docs builder to see why it fails to collect stuff that are not exported from `internal/types.ts`. Maybe it just needs to be exported, but if we export it in another file (and not export it publicly), it might work... Not sure though...

However, there _could_ be another solution to this, but it would take time to be handled and I'm not even sure it could be handled at all. I _could try_ to create [a new template](https://github.com/ReactiveX/rxjs/tree/master/docs_app/tools/transforms/templates/api) just for this case where that new template would have a list of properties much like [a template for interfaces has](https://github.com/ReactiveX/rxjs/blob/23bc7fdc16acd76dd71a4faf57b8b949684c3249/docs_app/tools/transforms/templates/api/interface.template.html#L9).

Would that be a good idea? 🤔

BTW, this is how this looks like now (on master, without changes from this PR):
image
At least I managed to add some types instead of `any`.

jakovljevic-mladen:
> My preference would be to try to figure out why the docs generation requires this

I'll try.

> IIRC, it's legal to use parameters (for example) in an exported API that have a type that is not exported - the type alias is expanded. TBH, I think this is done in several places, so I guess those doc's don't work either?

Do you have an example of this?

cartant:
> Do you have an example of this?

Pretty much all of the config types, IIRC:

https://github.com/ReactiveX/rxjs/blob/7.0.0-rc.2/src/internal/operators/share.ts#L7

jakovljevic-mladen:
`ShareConfig` doesn't seem to be shown anywhere:
- image
- image

cartant:
> `ShareConfig` doesn't seem to be shown anywhere

Yeah, that's why I think we should try to fix the docs generation. (FWIW, I thought TS had an issue with interfaces - rather than type aliases - being used without being exported, but I guess not.)

jakovljevic-mladen:
> Yeah, that's why I think we should try to fix the docs generation

I'll do my best, but I think we'll need help from @niklas-wortmann for this one.

cartant:
I'm inclined to change my mind. I think introducing and exporting the type is a small price to pay and we should make sure that all of the other types that are used for parameters, etc. are also exported, too.

src/internal/types.ts
```diff
@@ -302,3 +303,73 @@ interface ReadableStreamDefaultReaderLike<T> {
 export interface ReadableStreamLike<T> {
   getReader(): ReadableStreamDefaultReaderLike<T>;
 }
+
+/**
+ * The global configuration object for RxJS, used to configure things
+ * like how to react on unhandled errors. Accessible via {@link config}
+ * object.
+ */
+export interface GlobalConfig {
```
cartant, line 312: Can the interface declaration/docs be move back into the `src/internal/config.ts` file? And if it needs to be exported to be included in the docs, can the export be added here?

https://github.com/ReactiveX/rxjs/blob/c1c15db2448f8458de100a8d36c725c9c6bb0c14/src/index.ts#L86

benlesh: Yeah. It would be preferable if it were in the same file. It's not something that gets used all over the library like say... `ObservableInput` or something.

jakovljevic-mladen: Yes, this works. I'll move.

src/internal/config.ts
```diff
@@ -1,11 +1,28 @@
...
+/**
+ * The {@link GlobalConfig} object for RxJS. It is used to configure things
+ * like how to react on unhandled errors.
+ */
+export const config: GlobalConfig = {
+  onUnhandledError: null as ((err: any) => void) | null,
+
+  onStoppedNotification: null as ((notification: ObservableNotification<any>, subscriber: Subscriber<any>) => void) | null,
+
+  Promise: undefined as PromiseConstructorLike | undefined,
```
cartant, line 13: nitpick: now that we're using a type, we can remove these type assertions:
```suggestion
  onUnhandledError: null,
  onStoppedNotification: null,
  Promise: undefined,
```

cartant (approved):
LGTM. Just a nitpick.

src/internal/config.ts
jakovljevic-mladen: Yeah, noticed myself. Removed.

jakovljevic-mladen: Spaces between properties also?

cartant: The empty lines? Yeah, they were really only there because of the TSDoc. They serve no real purpose, but no big deal.

jakovljevic-mladen: I'll fix it anyway 🙂 Thanks!

jakovljevic-mladen:
> I think introducing and exporting the type is a small price to pay and we should make sure that all of the other types that are used for parameters, etc. are also exported, too.

I'll try to collect all other non-exported types and make a PR.

jakovljevic-mladen:
There is also one more thing that can be fixed (that is already fixed in the Angular docs) that I'll _try_ to work on: basically enable links on the types. In the next image from Angular docs:

image

This is actually a link that opens a type, while in RxJS docs these are not links at all. That would be a handy addition to the docs.

benlesh (approved):

benlesh:
Thank you, @jakovljevic-mladen!

PR #6251, opened 2021-04-22
fix: finalize behaves well with useDeprecatedSynchronousErrorHandling

### LINKED ISSUE #6250: finalize not firing in super gross mode
When using `useDeprecatedSynchronousErrorHandling`, `finalize` handlers are not called after a synchronous error is thrown

This is nefarious, because it means that maybe we're not calling any teardowns at all.

```ts
import { throwError, config } from "rxjs";
import { finalize } from "rxjs/operators";
console.clear();

try {
  throwError(new Error("test"))
    .pipe(
      finalize(() => {
       // This is never called.
        console.log("finalized");
      })
    )
    .subscribe(() => {
      console.log("nexted");
    });
} catch (err) {
  console.log("here");
}
```

benlesh:
Adds tests and ensures a few more scenarios that were hit in Google because they use the deprecated synchronous error handling.

fixes #6250

cc @leggechr

### DIFF
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -215,41 +215,54 @@ export class Observable<T> implements Subscribable<T> {
   ): Subscription {
     const subscriber = isSubscriber(observerOrNext) ? observerOrNext : new SafeSubscriber(observerOrNext, error, complete);
 
-    const { operator, source } = this;
     if (config.useDeprecatedSynchronousErrorHandling) {
-      let dest: any = subscriber;
-      dest._syncErrorHack_isSubscribing = true;
-
-      if (operator) {
-        subscriber.add(operator.call(subscriber, source));
-      } else {
-        try {
-          this._subscribe(subscriber);
-        } catch (err) {
-          dest.__syncError = err;
-        }
-      }
-
-      // In the case of the deprecated sync error handling,
-      // we need to crawl forward through our subscriber chain and
-      // look to see if there's any synchronously thrown errors.
-      // Does this suck for perf? Yes. So stop using the deprecated sync
-      // error handling already. We're removing this in v8.
-      while (dest) {
-        if ('__syncError' in dest) {
-          try {
-            throw dest.__syncError;
-          } finally {
-            subscriber.unsubscribe();
-          }
-        }
-        dest = dest.destination;
-      }
-      dest._syncErrorHack_isSubscribing = false;
+      this._deprecatedSyncErrorSubscribe(subscriber);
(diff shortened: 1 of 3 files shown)


src/internal/Observable.ts
```diff
@@ -215,43 +215,67 @@ export class Observable<T> implements Subscribable<T> {
...
+            this._subscribe(subscriber)
+          : // In all other cases, we're likely wrapping a user-provided initializer
+            // function, so we need to catch errors and handle them appropriately.
+            this._trySubscribe(subscriber)
+      );
     }
+    return subscriber;
+  }
 
-    subscriber.add(
```
benlesh, line 232: I split this up, and added comments. Just for organizational reasons more than anything.

src/internal/Observable.ts
```diff
@@ -215,43 +215,67 @@ export class Observable<T> implements Subscribable<T> {
...
+    let dest: any = subscriber;
+    dest._syncErrorHack_isSubscribing = true;
+    const { operator } = this;
+    if (operator) {
+      // We don't need to try/catch on operators, as they
+      // are doing their own try/catching, and will
+      // properly decorate the subscriber with `__syncError`.
+      subscriber.add(operator.call(subscriber, this.source));
+    } else {
+      try {
```
benlesh, line 253: the primary "fix" is here for the cases where teardown wasn't being called after it was registered if a sync error was thrown.

src/internal/operators/finalize.ts
```diff
@@ -58,7 +58,12 @@ import { operate } from '../util/lift';
  */
 export function finalize<T>(callback: () => void): MonoTypeOperatorFunction<T> {
   return operate((source, subscriber) => {
-    source.subscribe(subscriber);
-    subscriber.add(callback);
+    // TODO: This try/finally was only added for `useDeprecatedSynchronousErrorHandling`.
+    // REMOVE THIS WHEN THAT HOT GARBAGE IS REMOVED IN V8.
+    try {
```
benlesh, line 63: The "fix" for the finalize issue in the sync case is here. Note that adding a callback to a `closed` subscriber will cause it to be executed immediately.

cartant (approved):
🤢

cartant:
Do we have tests that are the equivalent of these but are for the normal error handling - not the deprecated synchronous error handling. Seems like something worth testing if we're not already doing that.

benlesh:
@cartant I'm unsure. I'll have a look after I merge this.

PR #6254, opened 2021-04-24
Fix types for configOrCount and configOrBufferSize to fix types in Google

rainshen49:
**Description:**

The current code would fail typescript check in non-strict-null mode. The proposed change does not change any behavior, but has better TypeScript inference so it also compiles in non-strict-null mode. 

This is a second attempt of https://github.com/ReactiveX/rxjs/pull/5705

### DIFF
--- a/src/internal/operators/shareReplay.ts
+++ b/src/internal/operators/shareReplay.ts
@@ -126,18 +126,20 @@ export function shareReplay<T>(
 ): MonoTypeOperatorFunction<T> {
   let bufferSize: number;
   let refCount = false;
-  if (configOrBufferSize && typeof configOrBufferSize === 'object') {
+  if (!configOrBufferSize) {
+    bufferSize = Infinity;
+  } else if (typeof configOrBufferSize === 'number') {
+    bufferSize = configOrBufferSize;
+  } else {
     bufferSize = configOrBufferSize.bufferSize ?? Infinity;
     windowTime = configOrBufferSize.windowTime ?? Infinity;
     refCount = !!configOrBufferSize.refCount;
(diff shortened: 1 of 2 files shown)


src/internal/operators/shareReplay.ts
```diff
@@ -126,18 +126,20 @@ export function shareReplay<T>(
...
     scheduler = configOrBufferSize.scheduler;
-  } else {
-    bufferSize = configOrBufferSize ?? Infinity;
   }
   return share<T>({
     connector: () => new ReplaySubject(bufferSize, windowTime, scheduler),
     resetOnError: true,
     resetOnComplete: false,
-    resetOnRefCountZero: refCount
+    resetOnRefCountZero: refCount,
```
rainshen49, line 143: I didn't do this. I think some commit hook did this.

src/internal/operators/shareReplay.ts
```diff
@@ -126,18 +126,20 @@ export function shareReplay<T>(
 ): MonoTypeOperatorFunction<T> {
   let bufferSize: number;
   let refCount = false;
-  if (configOrBufferSize && typeof configOrBufferSize === 'object') {
+  if (!configOrBufferSize) {
+    bufferSize = Infinity;
```
cartant, line 130: AFAICT, this will effect different behaviour, as passing a buffer size of zero will result in an infinite buffer.

benlesh:
We've rejected this in the past. Sorry. We can't make our build less strict/correct to accommodate another build process. 

https://github.com/ReactiveX/rxjs/pull/5705

If you're a googler, you should chat with @leggechr.

PR #6257, opened 2021-04-24
refactor: remove type annotations MochaDone & Mocha.Done for done parameter from all unit tests

nitinmalave:
**Description:**
- Assigned type as Mocha.Done to done parameter of tests arrow function in order to keep it consistent with other tests in repository.
- type MochaDone is deprecating and we can use Mocha.Done as a type in every unit test. Please take a look into it.

### DIFF
--- a/spec/observables/defer-spec.ts
+++ b/spec/observables/defer-spec.ts
@@ -132,7 +132,7 @@ describe('defer', () => {
       const source = hot('--a--b--c--|');
       const sourceSubs = '^-----!     ';
       const expected = '  --a--b-     ';
-      const unsub = '------!     ';
+      const unsub = '     ------!     ';
 
       const e1 = defer(() => source);
 
--- a/spec/observables/defer-spec.ts
+++ b/spec/observables/defer-spec.ts
@@ -47,7 +47,7 @@ describe('defer', () => {
     });
   });
 
-  it('should accept factory returns promise resolves', (done: MochaDone) => {
+  it('should accept factory returns promise resolves', (done) => {
     const expected = 42;
     const e1 = defer(() => {
       return new Promise<number>((resolve: any) => {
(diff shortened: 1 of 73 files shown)


cartant (changes requested):
Thanks for the PR, but I'm a bit confused. AFAICT, the changes in this PR don't seem to replace any uses of `MochaDone` with `Mocha.Done`. Rather, the changes seem to only add `Mocha.Done` to `done` parameters that didn't have type annotations. This isn't really necessary, as the parameter type can be inferred.

My preference would be to remove the type annotations for the `done` parameters altogether - as `Mocha.Done` is what TypeScript infers if no annotation is specified for the arrow function passed to `it`. If you are able to update this PR to remove the type annotations for all of the `done` parameters - those with `MochaDone` or `Mocha.Done` - I'd be happy to merge it.

nitinmalave:
Thanks @cartant for your feedback. I'll make those changes from all unit tests and raise the PR again.

nitinmalave:
Hey @cartant, I have made the changes as per your suggestion in all tests wherever i can do and pushed them on PR. Please take a look at it.

spec/observables/defer-spec.ts
```diff
@@ -132,7 +132,7 @@ describe('defer', () => {
       const source = hot('--a--b--c--|');
       const sourceSubs = '^-----!     ';
       const expected = '  --a--b-     ';
-      const unsub =      '------!     ';
+      const unsub = '------!     ';
```
cartant, line 135: This marble diagram wasn't aligned with spaces within the literal, so Prettier messed up its alignment:
```suggestion
      const unsub = '     ------!     ';
```

cartant (approved):
LGTM. Thanks.

PR #6261, opened 2021-04-25
refactor: refactor throwError operator to use factory function in unit tests and .ts file

nitinmalave:
…

**Description:**
- As per docs, throwError accepts factory functions to emit an error and no longer able to accept an error directly. That's why i went ahead and made changes in all unit tests wherever throwError is getting used to accept factory function.

Noop

### DIFF
--- a/spec-dtslint/observables/throwError-spec.ts
+++ b/spec-dtslint/observables/throwError-spec.ts
@@ -1,12 +1,23 @@
 import { throwError, animationFrameScheduler } from 'rxjs';
 
 it('should accept any type and return never observable', () => {
-  const a = throwError(() => (1)); // $ExpectType Observable<never>
-  const b = throwError(() => ('a')); // $ExpectType Observable<never>
-  const c = throwError(() => ({a: 1})); // $ExpectType Observable<never>
+  const a = throwError(1); // $ExpectType Observable<never>
+  const b = throwError('a'); // $ExpectType Observable<never>
+  const c = throwError({ a: 1 }); // $ExpectType Observable<never>
   const d = throwError(() => ({ a: 2 })); // $ExpectType Observable<never>
 });
 
-it('should support scheduler', () => {
-  const a = throwError(() => (1, animationFrameScheduler)); // $ExpectType Observable<never>
+it('should support an error value and a scheduler', () => {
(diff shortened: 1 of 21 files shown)


cartant (changes requested):
I'm okay with the wholesale replacement in the implementation tests, but I think the dtslint tests - spec-dtslint/observables/throwError-spec.ts - need to include both. So, rather than replace the calls in the dtslint tests, copy them and then replace them.

Also, this PR seems to have some conflicts with what's currently in master.

nitinmalave:
Hi @cartant, i am little confuse here. Can you please help me with what i understood is correct or not? 
As per your suggestion, 
- changes which needs to be done are in 'dtslint/observables/throwError-spec.ts' file only, right?
- 'dtslint/observables/throwError-spec.ts' should have 2 versions of throwError and is should look like this : 
 
`it('should accept any type and return never observable', () => {
  const a = throwError(1); // $ExpectType Observable<never>
  const a1 = throwError(() => (1)); // $ExpectType Observable<never>
  const b = throwError('a'); // $ExpectType Observable<never>
  const b1 = throwError(() => ('a')); // $ExpectType Observable<never>
  const c = throwError({a: 1}); // $ExpectType Observable<never>
  const c1 = throwError(() => ({ a: 1 })); // $ExpectType Observable<never>
  const d = throwError(() => ({ a: 2 })); // $ExpectType Observable<never>
});

it('should support scheduler', () => {
  const a = throwError(1, animationFrameScheduler); // $ExpectType Observable<never>
  const a1 = throwError(() => (1, animationFrameScheduler)); // $ExpectType Observable<never>
});`
- should i make this change in all of the files?

Please correct me if i am wrong.

cartant:
I'd create separate `it` functions to make it clearer:

```ts
it('should support an error value and a scheduler', () => {
  const a = throwError(1, animationFrameScheduler); // $ExpectType Observable<never>
});`

it('should support a factory and a scheduler', () => {
  const a = throwError(() => 1, animationFrameScheduler); // $ExpectType Observable<never>
});`
```

nitinmalave:
Got it. Thanks @cartant for quick support. I have committed this changes. Please take a look at it.

cartant (approved):
LGTM. Thanks.

PR #6262, opened 2021-04-26
docs: Document differences between 6.x and 7.x golden files

benlesh:
Adds a document outlining all discovered differences between version 6 and version 7 as they can be found while diffing the api_guardian snapshots. This document should be an exhaustive list of changes and we should modify it ongoing if we think of anything else that has changed that cannot be found here yet.

@cartant @kwonoj @kolodny please review.

@niklas-wortmann We're going to need some place to link this in the site, ideally to something on the home page.

ALL: I recommend we get this merged and out in the public as soon as possible, and simply iterate on it, adding anything we missed or changing verbiage. I don't want to get stuck reviewing this PR for eternity, I think it's better to get the information out and iterate, as the RC is already live.

### DIFF
--- a/docs_app/content/6-to-7-change-summary.md
+++ b/docs_app/content/6-to-7-change-summary.md
@@ -1,3 +1,7 @@
+# RxJS 6.x to 7.x Detailed Change List
+
+This document contains a detailed list of changes between RxJS 6.x and RxJS 7.x, presented in the order they can be found when diffing the TypeScript APIs in various module files.
+
 # module `rxjs`
 
 ## Breaking changes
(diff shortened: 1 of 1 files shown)


benlesh:
Also worth noting: I built this in such a way I figured that the docs site would generate an outline for it, @niklas-wortmann. Although I'm not sure if that's true or not, so I'd appreciate if you can confirm that.

cartant (approved):
Approving this. Will edit it later, once it's merged. There are some things in here that I think are now incorrect or are indicative of a problem - e.g. `_subscribe` is now tagged as `@internal` and should not be showing up in the public API as `protected`.

Yeah, it's no longer in the public API:

https://github.com/ReactiveX/rxjs/blob/04ceaa57b0cf93564dee01e38ddcfb208aaa2b08/api_guard/dist/types/index.d.ts#L252-L277

PR #6267, opened 2021-04-27
feat: add config object to connectable

cartant:
**Description:**

This PR changes the API of the new-to-version-7 `connectable` function. The changes introduce a config object (much like the one used in the `connect` and `share` operators) that:

- allows a `connector` subject factory to be passed;
- allows a `resetOnDisconnect` option to be specified.

Without these changes, it wasn't possible to refactor all use cases of `ConnectableObservable` to use `connectable` because `ConnectableObservable` would reset its subject upon disconnection - that is, on calling `unsubscribe` on the `Subscription` returned from the `connect` method.

This PR adds a few tests (there were none) for `connectable`.

### DIFF
--- a/src/internal/observable/connectable.ts
+++ b/src/internal/observable/connectable.ts
@@ -28,7 +28,7 @@ export interface ConnectableConfig<T> {
    * If true, the resulting observable will reset internal state upon disconnetion
    * and return to a "cold" state. This allows the resulting observable to be
    * reconnected.
-   * If false, upon disconnection, the connecting subject will will remain the
+   * If false, upon disconnection, the connecting subject will remain the
    * connecting subject, meaning the resulting observable will not go "cold" again,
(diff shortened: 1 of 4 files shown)


src/internal/observable/connectable.ts
```diff
@@ -18,29 +18,58 @@ export interface ConnectableObservableLike<T> extends Observable<T> {
...
+  /**
+   * A factory function used to create the Subject through which the source
+   * is multicast. By default this creates a {@link Subject}.
+   */
+  connector: () => SubjectLike<T>;
+  /**
+   * If true, the resulting observable will reset internal state upon disconnetion
+   * and return to a "cold" state. This allows the resulting observable to be
+   * reconnected.
+   * If false, upon disconnection, the connecting subject will will remain the
```
benlesh, line 31:
```suggestion
   * If false, upon disconnection, the connecting subject will remain the
```

benlesh (approved):

PR #6272, opened 2021-04-27
fix: resolve run-time errors when using deprecated sync error handling

### LINKED ISSUE #6271: 100% chance of runtime error if `config.useDeprecatedSyncErrorHandling` is enabled
Look at this dumbness: 

https://github.com/ReactiveX/rxjs/blob/7be67fa90f2f328906c3c0b8c8a3ee73d6b3941c/src/internal/Observable.ts#L278

I'm not sure how I wrote that and thought "okay, this is perfect", but clearly, we are missing a test or something. haha.

Fortunately, 99.9999% of people aren't using this mode. BECAUSE IT'S BAD.

benlesh:
Fixes a should-be-obvious issue where we were making sure `dest` was `undefined`, then trying to set a property on it.

fixes #6271

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -799,6 +799,17 @@ describe('Observable', () => {
         expect(results).to.deep.equal([1, 2]);
       });
 
+      // https://github.com/ReactiveX/rxjs/issues/6271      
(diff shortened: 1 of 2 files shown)


leggechr:
Tests are passing in google with this fix. LGTM

leggechr (approved):

PR #6276, opened 2021-04-28
fix: VS code will now properly import operators, et al

### LINKED ISSUE #6067: [v7] TS automatic imports in VSCode default to wrong path

benlesh:
- Adds typesVersions, enforcing TypeScript >= 4.2.
- Adds some superfluous imports required for TS and VS Code to find proper import locations for our strange deep import sites.
- Independantly tested by building the package and installing it locally in another project, then testing in VS code.

Fixes #6067
Related https://github.com/microsoft/TypeScript/issues/43034

NOTE: I'm worried about the `check-side-effects` run here, because it can't seem to figure out what to make of it. However, tested against an Angular build locally, this branch didn't have any problems tree-shaking or removing dead code. Given the importance of people getting their imports right, and the DX around this, we might want to disable `check-side-effects` if it causes us issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,12 @@
   "main": "./dist/cjs/index.js",
   "module": "./dist/esm5/index.js",
   "es2015": "./dist/esm/index.js",
-  "types": "./dist/types/index.d.ts",
+  "types": "index.d.ts",
+  "typesVersions": {
+    ">=4.2": {
+      "*": ["dist/types/*"]
+    }
+  },
   "sideEffects": false,
   "exports": {
     ".": {
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,18 @@
+//////////////////////////////////////////////////////////
+// Here we need to reference our other deep imports
+// so VS code will figure out where they are
+// see conversation here:
+// https://github.com/microsoft/TypeScript/issues/43034
+//////////////////////////////////////////////////////////
+
+// tslint:disable: no-reference
+/// <reference path="./ajax/index.ts" />
+/// <reference path="./fetch/index.ts" />
+/// <reference path="./operators/index.ts" />
+/// <reference path="./testing/index.ts" />
+/// <reference path="./webSocket/index.ts" />
+// tslint:enable: no-reference
+
 /* Observable */
 export { Observable } from './internal/Observable';
 export { ConnectableObservable } from './internal/observable/ConnectableObservable';


benlesh:
attn @filipesilva ... we're getting errors from `check-side-effects`, however, there's no real change to the tree-shakability of what rxjs is outputting here from any of my tests (with angular-cli).

image

benlesh:
It does have me a little concerned that `rollup` will have issues with this. Good grief, I had modules. @RichHarris, should rollup not like what we're doing here? It's admittedly bananas, but it seems to be working otherwise.

benlesh:
Yeah, I don't think we can get this fix into version 7 unless we're totally sure it's in the clear with everyone's bundlers. This is _only_ to fix auto imports in VS Code... as crazy as that is. So just waiting to have some experts weigh in. 

Hopefully we don't need to restructure our entire build to get things working in VS Code. :\

cartant:
FWIW, I did a small test with Rollup and the inclusion of the imports seems to make no difference to its output and tree shaking seems to work just fine.

Rich-Harris:
Unfortunately I got different results. I added a `tmp/index.js` file with a single line:

```js
import '../dist/esm5_for_rollup/index.js';
```

Running the following on `master` after `npm run compile`...

```bash
npx rollup@latest -i tmp/index.js -o tmp/master.js -f esm
```

...yielded a 48kb (unminified) file. 

I switched to this branch and it failed at first, because `dist/esm5_for_rollup/index.js` contained the new directory imports...

```js
import './ajax';
import './fetch';
import './operators';
import './testing';
import './webSocket';
```

...which Rollup chafes at (since native loaders don't handle directory imports). Appending `/index.js` to each of the imports allowed the command to run, but it yielded a 91kb file.

benlesh:
Thank you so much, @Rich-Harris, for taking the time to look at this.

---

So given this feedback, I think this PR is a no-go, and I'm going to close it. We're going to have to come up with another solution.

benlesh:
All right, I've reopened this because there was another approach proposed by the TypeScript team, and it seems to be working! cc @cartant

benlesh:
I've tested this with the methodology recommended by @Rich-Harris, and the output is identical between this branch and the master branch. So I think this is good to go.

kwonoj (approved):

PR #6279, opened 2021-04-28
feat(AjaxResponse): add stricter `type` (`AjaxResponseType`)

OliverJAsh:
**Description:** currently `AjaxResponse['type']` is typed as `string`, but we can use a stricter type for this since we know which strings will be used. This will help users `switch` over the various types.

### DIFF
--- a/src/internal/ajax/types.ts
+++ b/src/internal/ajax/types.ts
@@ -7,6 +7,10 @@ import { PartialObserver } from '../types';
  */
 export type AjaxDirection = 'upload' | 'download';
 
+export type ProgressEventType = 'loadstart' | 'progress' | 'load';
+
+export type AjaxResponseType = `${AjaxDirection}_${ProgressEventType}`;
+
 /**
  * The object containing values RxJS used to make the HTTP request.
  *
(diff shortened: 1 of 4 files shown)


src/internal/ajax/types.ts
```diff
@@ -7,6 +7,10 @@ import { PartialObserver } from '../types';
  */
 export type AjaxDirection = 'upload' | 'download';
 
+export type ProgressEventType = 'loadstart' | 'progress' | 'load';
+
+export type AjaxResponseType = `${AjaxDirection}_${ProgressEventType}`;
```
OliverJAsh, line 12: Template literal types require 4.1. I wasn't sure what the minimum supported TS version was. If needed I can swap this out for a regular union:

```suggestion
type AjaxResponseType =
  | 'upload_loadstart'
  | 'upload_progress'
  | 'upload_load'
  | 'download_loadstart'
  | 'download_progress'
  | 'download_load';
```

cartant: 4.2 is the minimum supported version, so template literal types are fine.

cartant: You'll need to do an `npm run api_guardian:update`, too.

cartant (approved):
LGTM. Thanks.

benlesh (approved):

benlesh:
Thanks, @OliverJAsh!

(Sorry I accidentally hit the close button there, haha 😅 )

Issue #6243, opened 2021-04-20
Problems testing observables that emit arrays or objects with other observables.


fcostarodrigo:
## Bug Report

**Current Behavior**
When using expectObservable to check an observable that emits and object with some observable the test fails.

**Expected behavior**
The test should pass.

**Reproduction**
That can be 100% reproduced using the code below.

- REPL or Repo link:
https://replit.com/@fcostarodrigo/testSchedulerHighOrderObservables#index.test.js

```js
const highOrderObj = of({ a: of('a') });

it('generate the stream correctly', () => {
  testScheduler.run(({ cold, expectObservable }) => {
    expectObservable(highOrderObj).toBe('(a|)', { a: { a: cold('(a|)') } });
  });
});
```

**Environment**
- Runtime: node v12.22.1
- RxJS version: 6.6.7

voliva:
I'm not sure this is to be considered a bug, but maybe a feature request?

`highOrderObj` is a stream that emits objects - In a way I would expect for that to just deeply compare both objects (the expected and the result), but now one of the property of that object is an observable.

I'd like to have this feature, I think it makes sense. But I wouldn't consider it a bug

fcostarodrigo:
I was not sure as well if it was a bug or a feature. Since it works for `of(of('a'))` I was thinking it would be a bug. But working for `of(of('a'))` could be a coincidence.

PR #6289, opened 2021-04-28
feat(testScheduler): support nested observables within objects

voliva:
**Description:**

Adds deep unwrapping of observables in marble tests, so that we can write tests such as:

```ts
expectObservable(
  cold('---a-|'),
  {
    a: {
      nestedObservable: cold('-1-2-3')
    }
  }
).toBe(
  '---a-|',
  {
    a: {
      nestedObservable: cold('-1-2-3')
    }
  }
)
```

Previously we had a similar feature, but would only work for just observables of observables (and only 1-level deep).

**Related issue (if exists):** #6243 

I'm sorry I went ahead and implemented this before any member of the core team had the chance to express if it's a feature you even want to support, but I found it interesting as an exercise. I also understand v7 is around the corner and there are things probably more important than this.

If you want this PR to move forward, I'll need to add a few more tests and tweak edge cases (circular references), but so far I have a proof-of-concept.

### DIFF
--- a/spec/schedulers/TestScheduler-spec.ts
+++ b/spec/schedulers/TestScheduler-spec.ts
@@ -281,6 +281,16 @@ describe('TestScheduler', () => {
         expectObservable(NEVER).toBe('---', {});
       });
 
+      it('handles observables of observables', () => {
+        expectObservable(cold('a', { a: cold('b') }))
+                        .toBe('a', { a: cold('b') })
+      })
+      
+      it('handles observables of nested observables', () => {
+        expectObservable(cold('-a-b', { a: {a: cold('--b'), b: 3} }))
+                        .toBe('-a-b', { a: {a: cold('--b'), b: 3}})
+      })
+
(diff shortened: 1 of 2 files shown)


cartant:
> If you want this PR to move forward, I'll need to add a few more tests and tweak edge cases...

Yeah, I'd suggest you hold off on that. I'm not sure that this is something that we'd want to add. It would need to be discussed. IMO, it's not something we'd want to have to support.

benlesh:
Hi @voliva — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

PR #6293, opened 2021-04-29
fix: correct path to types in package.json

brandonroberts:
**Description:**

VSCode shows rxjs being imported as `any` due to missing types. This fixes  the path to the `index.d.ts` types file so the type declarations are resolved correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "./dist/cjs/index.js",
   "module": "./dist/esm5/index.js",
   "es2015": "./dist/esm/index.js",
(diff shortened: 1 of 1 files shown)


cartant:
Believe it or not, that path is correct. If `typesVersions` is specified, the path is relative.

https://github.com/microsoft/TypeScript/issues/43034#issuecomment-825749864

What TS version are you using?

brandonroberts:
Gotcha. We're on 4.2.4. Here is the branch in the NgR xrepo https://github.com/ngrx/platform/tree/rxjs7

PR #6294, opened 2021-04-30
docs: remove beta tag since v7 is now official

ghiscoding:
**Description:**
A simple change in the readme to remove the (beta) tag since v7 is now official

nope


cartant (approved):
Thanks.

Issue #5270, opened 2020-01-26
[Discussion] adjust transpilation target above es5


kwonoj:
**This is definitely semver major breaking changes discussion**.

With Windows 7 EOL, IE11 is also considered as EOL. There are still in-use customers like few enterprise companies, but it won't get a major update in features at all.

Currently, rxjs targets transpilation target's set to `es5`, conservative runtime output to support most of browsers. This issue is to discuss

- potentially raise transpilation target above es5, to reduce transpilation runtime
- declare few browser support deprecation

1. What target should we aim for?

This depends on major browser usage across consumers, but probably best is set to ES2015 to aim conservative target (again). 

2. What about old browsers (IE11)?

Consumer aims to make application to use those browser may need additional configuration on their bundler to downlevel transpilation to rxjs.

3. Should we do this now? why?

Maybe not, just open for discussion (in related with native esm support changes I'm trying to tackle).

kwonoj:
Core team has reached general agreement to pursue this. Current planned roadmap as written below, still subject to change depends on ecosystem / user feedback.

- 7.x : no transpilation target adjustment
- 8.x : cjs / esm will adjust transpile target to 2015 as semver breaking major
- beyond 9: potential exploration to taget esm as native (current esm is equivalent to cjs transpiled output but only preserves es import syntax). this is TBD.

JeanMeche:
Looks like right now, v8 will ship `ES2022` code right ?

kwonoj:
Yes, that's the plan so far.

benlesh:
Hi @kwonoj — thank you for taking the time to open this issue and for sharing your thoughtful feedback with the RxJS project. I’m sorry it took us so long to respond. Given how much time has passed and the lack of recent activity, I’m going to close this for now so the tracker reflects the work we’re likely to act on. If this is still relevant on a current RxJS release, please feel welcome to open a new issue with updated context and a reproduction. Thank you again for your contribution and patience.

— Ben

PR #6295, opened 2021-04-30
build(package): update transpilation target

kwonoj:
**Description:**

https://github.com/ReactiveX/rxjs/issues/5270

We decided to adjust transpilation target higher and drops IE11 support, it needed 7 release to introduce major breaking in 8. PR also drops tslib which doesn't seem to be helpful lot in terms of tree shaking. We can bring it back later if needed.


benlesh:
I think we decided we can close this a long time ago.

PR #6302, opened 2021-04-30
docs(NEVER): fix marble diagram image URL

jakovljevic-mladen:
**Description:**
This change fixes [NEVER](https://rxjs.dev/api/index/const/NEVER) marble diagram image. This change is basically renaming the file name so that it matches this line:
https://github.com/ReactiveX/rxjs/blob/f132d0d74cb059db74748025d35e22b2b0570f42/src/internal/observable/never.ts#L7

None


cartant:
I was under the impression that those diagrams were generated using a tool and a script. If that is the case, I think we might be better of changing the TSDoc in case the file gets regenerated as `NEVER.png`. Thoughts?

jakovljevic-mladen:
I was under the same impression. I would rather fix the tool/script than do this, but I think this is safe.

Here's what I did and what's my understanding of what's going on:
- When I saw that `NEVER` is missing an image, I went investigating and saw that there is this line: ` `.
- I wanted to try to find a file named `never.png` but I stumbled upon two such files:
  - `docs_app/assets/images/marble-diagrams/never.png`
  - `docs_app/src/assets/images/marble-diagrams/NEVER.png`
- Then, I wanted to know why there two identical files in two folders (`assets/images/marble-diagrams` is the same for both, only the root is different: `docs_app` vs `docs_app/src`).
- Then, I went to delete the whole `marble-diagrams` from the `docs_app` root (because Angular usually *knows* about `assets` under `src` folder).
- I did that and run the `npm run docs` script only to find out that the `docs_app/assets/images/marble-diagrams` has been recreated yet again.
- I went investigating and I think this line https://github.com/ReactiveX/rxjs/blob/f132d0d74cb059db74748025d35e22b2b0570f42/docs_app/tools/transforms/angular-api-package/post-processors/embedMarbleDiagrams.js#L29 was doing that copy.
- Then, I went to delete both `marble-diagrams` folders from both roots: `docs_app` and `docs_app/src` which didn't generate anything in the `docs_app` or `docs_app/src` folders.
- I went to conclusion that, for some reason, `npm run docs` script is copying files from `docs_app/src/...` to `docs_app/...`. I don't yet understand files in the `docs_app/...` are doing here and how they are being used.
- One funny thing about this is that it looks like `NEVER.png` (from `docs_app/src/...`) got copied to `docs_app/...` as `never.jpg`.

I think marble images generation is not happening at all. I also discovered that Swirly package that was recently added is having only one template file: `docs_app/tools/marbles/diagrams/concatAll.txt` and that only `concatAll` docs are using `.svg` as image format instead of `.png` which means that this docs is using the generated file, while all others don't seem to be using this.

OFC, I may be highly wrong about all of this as I'm quite new to discovering how all of this works. Maybe @niklas-wortmann or @timdp could help here.

cartant:
I see. That makes sense and, yeah, your change looks like the best approach. Thanks for the detailed explanation; it's much appreciated.

cartant (approved):

timdp:
I don't have a lot to add since I only maintain Swirly, not the docs themselves. However, if there's anything I can do to actually get Swirly diagrams into the docs, let me know.

Issue #6297, opened 2021-04-30
Unable to import in TS project without TS `DOM` lib


OliverJAsh:
In `7.0.0-rc.3` and prior it was possible to import `rxjs` in a TS project that omitted the `DOM` `lib`. For example:

`tsconfig.json`:
```json
{
    "compilerOptions": {
        "lib": ["ES2015"]
    }
}
```

However in `7.0.0` this is no longer possible because of this change: https://github.com/ReactiveX/rxjs/pull/6276/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R9-R13

`index.ts`:
```ts
import * as RxJS from 'rxjs'
RxJS;
```

When I run `tsc --noEmit` I get a bunch of errors:

```sh-session
$ tsc --noEmit
node_modules/rxjs/dist/types/internal/ajax/AjaxResponse.d.ts:20:29 - error TS2304: Cannot find name 'ProgressEvent'.

20     readonly originalEvent: ProgressEvent;
                               ~~~~~~~~~~~~~

node_modules/rxjs/dist/types/internal/ajax/AjaxResponse.d.ts:26:19 - error TS2304: Cannot find name 'XMLHttpRequest'.

26     readonly xhr: XMLHttpRequest;
                     ~~~~~~~~~~~~~~

node_modules/rxjs/dist/types/internal/ajax/AjaxResponse.d.ts:56:28 - error TS2304: Cannot find name 'XMLHttpRequestResponseType'.

56     readonly responseType: XMLHttpRequestResponseType;
                              ~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/rxjs/dist/types/internal/ajax/AjaxResponse.d.ts:89:20 - error TS2304: Cannot find name 'ProgressEvent'.

89     originalEvent: ProgressEvent,
                      ~~~~~~~~~~~~~

node_modules/rxjs/dist/types/internal/ajax/AjaxResponse.d.ts:95:10 - error TS2304: Cannot find name 'XMLHttpRequest'.

95     xhr: XMLHttpRequest,
            ~~~~~~~~~~~~~~

node_modules/rxjs/dist/types/internal/ajax/errors.d.ts:13:10 - error TS2304: Cannot find name 'XMLHttpRequest'.

13     xhr: XMLHttpRequest;
            ~~~~~~~~~~~~~~

[...]

PR #6305, opened 2021-04-30
fix: do not reference DOM-related imports

cartant:
**Description:**

This PR removes all DOM-related imports from the references that were added to help VS Code choose the correct import locations for its automatic imports feature.

I've confirmed - using https://github.com/OliverJAsh/rxjs-v7-no-dom-lib-test - that this resolves the problem.

**Related issue (if exists):** #6297

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,11 +6,10 @@
 //////////////////////////////////////////////////////////
 
 // tslint:disable: no-reference
-/// <reference path="./ajax/index.ts" />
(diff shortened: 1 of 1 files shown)


cartant:
Also, the [no-internal](https://github.com/cartant/eslint-plugin-rxjs/blob/main/docs/rules/no-internal.md) ESLint rule now has a fixer.

benlesh (approved):

Issue #6301, opened 2021-04-30
Typing of 3 param version of `fromEvent` in v7 is not compatible with `EventEmitter` of node.js


thynson:
## Bug Report

**Current Behavior**
Typing of `fromEvent` in v7 is not compatible with `EventEmitter` of node.js, while its declaration claims it's accepted.

**Expected behavior** 
It shall be compatible with v6.

**Reproduction**
```typescript
fromEvent(process, 'SIGINT', {once: true}); // Does not compile
```

**Environment**
- Runtime: Node v14
- RxJS version: 7

Actually, it's Typescript that does not resolve function overload properly when the function signature contains union type, but at least Typescript is still able to resolve overload by the number of arguments, which is why declaration in v6 works.

cartant:
Needs a minimal reproduction. This seems to be fine, for me with the versions you mentioned:

```ts
import { fromEvent } from "rxjs";
const beforeExit = fromEvent(process, "beforeExit");
```

thynson:
Sorry, I imported an EventEmitter that is not from 'node' when trying to craft a minimal repo.
But 
```
fromEvent(process, 'SIGINT', {once: true});
```
does not compile.

thynson:
Umm, the original intent in my codebase was to subscribe to the event for once, so the `{once: true}` is a misuse and shall be fixed by `.pipe(first())`.

thynson:
And the 3-param overload of `fromEvent` for node's `EventEmitter` is invalid and shall be removed.

PR #6306, opened 2021-05-01
docs: mention tighter fromEvent types

cartant:
**Description:**

This PR mentions, in the changes doc, that prior to separating the `fromEvent` signatures by target type, it was possible to pass options to APIs that didn't support them. (The options are ignored in the `fromEvent` implementation for targets that don't support them.) Attempting to do this in v7 will effect a TypeScript compilation error.

**Related issue (if exists):** #6301


cartant:
Closed in favour of #6313.

PR #6318, opened 2021-05-02
chore(sequenceEqual): convert sequenceEqual tests to run mode

tmair:
**Description:**
This PR converts `sequenceEqual` tests to run mode. There are two more things that i believe where not correct which are also adressed in this PR.
* Two of the tests are using `values` instead of the default string values. They where defined however not used with the correct marble source.
* The same tests where modifying the `booleans` constant by using `_.assign`. This side effect was removed.

I also had to modify the last test in the test file from a subscription within `scheduler.schedule` to subscription marbles. While I believe that this change is semantically equal I am not totally sure if that is the case.

None

### DIFF
--- a/spec/operators/sequenceEqual-spec.ts
+++ b/spec/operators/sequenceEqual-spec.ts
@@ -1,4 +1,3 @@
-import * as _ from 'lodash';
 import { hot, cold, expectObservable, expectSubscriptions, time } from '../helpers/marble-testing';
 import { sequenceEqual } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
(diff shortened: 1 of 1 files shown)


spec/operators/sequenceEqual-spec.ts
```diff
@@ -1,323 +1,376 @@
...
-    const delay =        '-------------------|';
-    const s3 = cold(                        '-f---g---h---i---j---|');
-    const expected2 =    '                   ---------------------(T|)';
-    const s3subs =       '                   ^                    !';
-
-    const test1 = s1.pipe(sequenceEqual(s2));
-    const test2 = s1.pipe(sequenceEqual(s3));
-
-    expectObservable(test1).toBe(expected1, booleans);
-    rxTestScheduler.schedule(() => expectObservable(test2).toBe(expected2, booleans), time(delay));
```
tmair, line 319: This is the test where i removed the subscription wihin the `schedule` call.

cartant (approved):
LGTM. Thanks.

PR #6323, opened 2021-05-03
chore(sampleTime): convert sampleTime tests to run mode

tmair:
**Description:**
This PR converts `sampleTime` tests to run mode.

I did encounter an issue with the `never` test. It seems that in run mode there is no maximum frame count, so I had to add an explicit unsubscription to the expectation.

None

### DIFF
--- a/spec/operators/sampleTime-spec.ts
+++ b/spec/operators/sampleTime-spec.ts
@@ -1,114 +1,150 @@
-import { hot, cold, expectObservable, expectSubscriptions } from '../helpers/marble-testing';
+/** @prettier */
 import { sampleTime, mergeMap } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { of } from 'rxjs';
-
-declare const rxTestScheduler: TestScheduler;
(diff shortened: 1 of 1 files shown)


spec/operators/sampleTime-spec.ts
```diff
@@ -1,114 +1,150 @@
...
-    expectSubscriptions(e1.subscriptions).toBe(e1subs);
+  it('should not complete if source does not complete', () => {
+    rxTest.run(({ cold, expectObservable, expectSubscriptions, time }) => {
+      const e1 = cold('    --------');
+      const e1subs = '     ^------!';
+      const expected = '   --------';
+      const period = time('-----|  ');
+      const e1unsbs = '    -------!';
+
+      expectObservable(e1.pipe(sampleTime(period, rxTest)), e1unsbs).toBe(expected);
```
tmair, line 146: This is where an explicit unsubscription was added.

benlesh: Makes sense.

benlesh (approved):

benlesh:
Nicely done! Thanks, @tmair!

Issue #6351, opened 2021-05-04
subscription doesn't remove itself from parent subscriptions' _teardowns as expected when unsubscribed


panchao5:
## Bug Report

excuse any English errors

in `subscription.unsubscribe`, mutating `_parentage` while for-looping it results in some items being skipped. relative lines are:

```js
for (const parent of _parentage) {
   parent.remove(this); // will call `_removeParent` on `this`
}
```

**Reproduction**
[codesandbox](https://codesandbox.io/s/charming-surf-5723f?file=/src/index.js)

PR #6352, opened 2021-05-04
fix(Subscription): null _parentage on unsubscribe

cartant:
**Description:**

This PR fixed the bug reported in #6351 by nulling `this._parentage` before the array is iterated. This was an oversight. It should have been nulled as is `this._teardowns`, immediately after it is destructured:

https://github.com/ReactiveX/rxjs/blob/7b30546717ab8bb271217064318ef9e0c0c40ecc/src/internal/Subscription.ts#L76-L78

If `this._parentage` is not nulled, it's mutated as it's iterated and parents can be left holding onto child subscriptions that have been closed.

The PR also adds - in a separate commit - an `if` statement around the handling of `_parentage`, so that the implementation is more like what's done with `_teardowns`. If there's some reason for not using the `if` statement, we can just revert that commit.

**Related issue (if exists):** #6351

### DIFF
--- a/spec/Subscription-spec.ts
+++ b/spec/Subscription-spec.ts
@@ -116,7 +116,7 @@ describe('Subscription', () => {
   });
 
   describe('unsubscribe()', () => {
-    it('Should unsubscribe from all subscriptions, when some of them throw', (done) => {
+    it('should unsubscribe from all subscriptions, when some of them throw', (done) => {
       const tearDowns: number[] = [];
 
       const source1 = new Observable(() => {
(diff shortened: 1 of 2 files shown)


spec/Subscription-spec.ts
```diff
@@ -200,5 +200,28 @@ describe('Subscription', () => {
...
+      // separate property.
+      expect((a as any)._teardowns).to.have.length(1);
+      expect((b as any)._teardowns).to.have.length(1);
+      expect((c as any)._teardowns).to.have.length(1);
+      d.unsubscribe();
+      // When d is unsubscribed, it should remove itself from each of its
+      // parents.
+      expect((a as any)._teardowns).to.have.length(0);
+      expect((b as any)._teardowns).to.have.length(0);
+      expect((c as any)._teardowns).to.have.length(0);
```
cartant, line 224: I could re-write this test as a memory leak test - using `FinalizationRegistry` instead of the private `_teardown` property - but IDK. It's fairly straightforward ATM.

benlesh (approved):

Issue #6143, opened 2021-03-15
Ensure signatures with `thisArg` or `source` observable are deprecated in predicates and projection functions


benlesh:
Just what it says. `filter(fn, obj)` is more readable as `filter(fn.bind(obj))`, because the former doesn't really tell us what `obj` is for. This exists in various parts of the library. We also have signatures where the `source` observable is provided in the predicate. That is supremely silly, as I can't even think of a use case there, and a simple closure would have worked (for example, `every` has both an `thisArg` and a `source`).

OliverJAsh:
+100. I just shipped a bug to production because I accidentally passed an operator as `thisArg`. Very easy mistake to make:

```ts
import { Observable } from 'rxjs';
import { map, shareReplay } from 'rxjs/operators';

declare const ob$: Observable<number>;
// Oops, shareReplay is in the wrong place:
ob$.pipe(map((x) => x + 1, shareReplay()));
// Should have been:
ob$.pipe(map((x) => x + 1), shareReplay());
```

benlesh:
Core team approval.

PR #6361, opened 2021-05-05
chore: deprecate thisArg signatures

cartant:
**Description:**

This PR deprecates the `thisArg` signatures.

Unfortunately, the deprecation of predicates that accept a `source` parameter isn't possible. When the non-deprecated signatures - i.e. those with predicates that _do not_ accept `source` parameters - are added, the inference breaks and calls that pass source-accepting predicates won't compile - for the dtslint tests - because the parameters are inferred as `any` and implicit `any` is disallowed. I can guess at what's happening internally - 'cause, normally, you would not have separate signatures - but the inference failure is still weird.

The removal of the `source` parameters will have to be conveyed in documentation instead. Or left as a breaking change surprise.

**Related issue (if exists):** #6143


benlesh:
Hmm... it would be great if this worked perfectly... but part of me thinks that it's also okay if we just document it if we can't get this working just right.

benlesh (commented):
Question: Are we able to test these deprecations with dtslint?

benlesh (approved):

PR #6365, opened 2021-05-05
fix: Observable teardowns now properly called if `useDeprecatedSynchronousErrorHandling` is `true`.

### LINKED ISSUE #6364: useDeprecatedSynchronousErrorHandling not calling teardown if synchronously unsubscribed
Found in google3 (of course, haha) cc @leggechr 

```ts
import { Observable, config } from "rxjs";

let called = false;
const observable = new Observable(obs => () => (called = true));
const subscription = observable.subscribe();
subscription.unsubscribe();

console.log("called: ", called); // logs called: false
```

https://stackblitz.com/edit/rxjs-4stiyw?devtoolsheight=60

benlesh:
Resolves an issue where teardowns returned by the Observable initializer were not being registered with the subscriber.

Fixes #6364

cc @leggechr

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -810,6 +810,24 @@ describe('Observable', () => {
         }).not.to.throw();
       });
 
+      it('should call teardown if sync unsubscribed', () => {
+        let called = false;
+        const observable = new Observable(() => () => (called = true));
+        const subscription = observable.subscribe();
+        subscription.unsubscribe();
+
+        expect(called).to.be.true;
+      });
+
+      it('should call registered teardowns if sync unsubscribed', () => {
+        let called = false;
+        const observable = new Observable((subscriber) => subscriber.add(() => called = true));
+        const subscription = observable.subscribe();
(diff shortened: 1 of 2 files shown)


spec/Observable-spec.ts
```diff
@@ -810,6 +810,24 @@ describe('Observable', () => {
         }).not.to.throw();
       });
 
+      it('should call teardown if sync unsubscribed', () => {
```
benlesh, line 813: This is the only test that was failing. I just added the other one to cover the other scenario.

src/internal/Observable.ts
```diff
@@ -252,7 +252,7 @@ export class Observable<T> implements Subscribable<T> {
       subscriber.add(operator.call(subscriber, this.source));
     } else {
       try {
-        this._subscribe(subscriber);
+        subscriber.add(this._subscribe(subscriber));
```
benlesh, line 255: 🤦

cartant (approved):

PR #5834, opened 2020-10-16
fix(multicast,share,refCount,shareReplay): enable synchronous firehose

miginmrs:
- use known subscriber when possible instead of returned subscription
- remove unused unsubscribe from refCount Operator
- add teardown logic before calling subscribe when possible
- enable remaining synchronous firehose tests

fixes #5658

PR #6370, opened 2021-05-06
fix(share): propagate closed to firehose sources

cartant:
**Description:**

This PR fixes `share` so that it wires up the teardown _before_ subscribing to the source - ensuring the assignment to the subscriber's `closed` property is propagated to the firehose source.

The skipped firehose test has been enabled and the firehose tests that related to now-deprecated operators have been removed.

**Related issue (if exists):** #5834 (I've closed this because it related to the now-deprecated APIs)

### DIFF
--- a/spec/operators/refCount-spec.ts
+++ b/spec/operators/refCount-spec.ts
@@ -114,25 +114,4 @@ describe('refCount', () => {
     expect(arr[0]).to.equal('the number one');
     expect(arr[1]).to.equal('the number two');
   });
-
-  // TODO: fix firehose unsubscription
-  it.skip('should stop listening to a synchronous observable when unsubscribed', () => {
-    const sideEffects: number[] = [];
-    const synchronousObservable = new Observable<number>(subscriber => {
-      // This will check to see if the subscriber was closed on each loop
-      // when the unsubscribe hits (from the `take`), it should be closed
-      for (let i = 0; !subscriber.closed && i < 10; i++) {
-        sideEffects.push(i);
-        subscriber.next(i);
-      }
-    });
-
-    synchronousObservable.pipe(
-      multicast(() => new Subject<number>()),
-      refCount(),
-      take(3),
-    ).subscribe(() => { /* noop */ });
-
-    expect(sideEffects).to.deep.equal([0, 1, 2]);
-  });
 });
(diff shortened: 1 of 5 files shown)


src/internal/operators/share.ts
```diff
@@ -112,6 +112,26 @@ export function share<T>(options?: ShareConfig<T>): OperatorFunction<T, T> {
     // Create the subject if we don't have one yet.
     subject = subject ?? connector();
 
+    // Add the teardown directly to the subscriber - instead of returning it -
+    // so that the handling of the subscriber's unsubscription will be wired
+    // up _before_ the subscription to the source occurs. This is done so that
+    // the assignment to the source connection's `closed` property will be seen
+    // by synchronous firehose sources.
+    subscriber.add(() => {
```
benlesh, line 120: 🏆

spec/operators/shareReplay-spec.ts
```diff
@@ -347,25 +347,6 @@ describe('shareReplay', () => {
...
-    const synchronousObservable = new Observable<number>((subscriber) => {
-      // This will check to see if the subscriber was closed on each loop
-      // when the unsubscribe hits (from the `take`), it should be closed
-      for (let i = 0; !subscriber.closed && i < 10; i++) {
-        sideEffects.push(i);
-        subscriber.next(i);
-      }
-    });
-
-    synchronousObservable.pipe(shareReplay(), take(3)).subscribe(() => {
```
benlesh, line 362: It seems like we'd still want this test for `refCount: true`.

spec/operators/refCount-spec.ts
```diff
@@ -114,25 +114,4 @@ describe('refCount', () => {
     expect(arr[0]).to.equal('the number one');
     expect(arr[1]).to.equal('the number two');
   });
-
-  // TODO: fix firehose unsubscription
-  it.skip('should stop listening to a synchronous observable when unsubscribed', () => {
```
benlesh, line 119: Not sure why this one was just removed. Presumably because we're only fixing the modern operators?

benlesh (approved):
Approved. However, I think we may want to add the test back for `shareReplay` in the `refCount: true` case.

PR #6373, opened 2021-05-06
build(package): support internal subpath import

kwonoj:
**Description:**

This PR re-enables subpath import under `internal` in v7. 

Previously subpath import was not restricted in v6. With change of import maps node.js only allows explicitly exposed subpath import, ends up any subpath import to `internal` is prohibited. This is not a breaking change techinically - we never officially supported internal subpath and never will. But for some cases, user have usecases with `at your own risk` and this PR will allow it again.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -106,6 +106,10 @@
         "hot": true,
         "cold": true
     },
+    "parserOptions": {
+        "sourceType": "module",
+        "ecmaVersion": 2015
+    },
     "env": {
         "browser": true,
         "node": true,
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ spec-build/
 npm-debug.log
 .DS_STORE
 *.tgz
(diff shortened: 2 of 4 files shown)


cartant (approved):

cartant:
I doesn't change my mind about approving or merging this - I think this is fine - but do we know whether or not this takes us back so square one with the VS Code imports? That is, if `internal` is in the `exports` configuration, is VS Code likely to start defaulting to the internal location for some imports? @benlesh

kwonoj:
So far I think  it's not, since mapping is explicit to `./internal` and root path mapping is still preserved: https://github.com/ReactiveX/rxjs/pull/6373/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R18-R20 but bit hard to 100% sure.

cartant:
🤞

Issue #5404, opened 2020-04-24
About consistency, deprecations and docs


UserGalileo:
### Documentation Related To Component:

/

### Please check those that apply

- [ ] typo
- [ ] documentation doesn't exist
- [x] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

### Description Of The Issue

I'd like to talk about some topics which are honestly driving me nuts lately and I think they should be clarified. I found different issues for some of them, but some are quite old and it's not clear what's reliable or not, so forgive me if some of this is already present in other issues, but I think all of this is correlated.

Let's start with deprecations:

## Deprecations and misleading docs
The deprecation of **resultSelectors** was made clear in the codebase, in blog posts and in conferences. But the docs (v6.5.5) are incomplete or misleading: take for example _switchMap_/_mergeMap_, the **only** place it says "depcrecated" is in the "_Returns_" paragraph of the docs. But why? Why shouldn't we put a visible label on the parameter itself? In order to know what's deprecated or not, I'm doing cmd+f in every page looking for "deprecated" hoping to find something. I know there's a dedicated "Migration" page with this info, but I shouldn't be looking at it for every operator API that I check! Is it going to change in v7 docs?

PS. Little rant (forgive me): in version 6.x resultSelectors were deprecated and everywhere it was said that with the next major version they were going away. It appears that they're still around in v7. Why? I get it, fewer breaking changes, I just found these info to be misleading. And just to make it clear: I **love** resultSelectors and frankly I don't understand why there was the decision to get rid of them. Don't get me wrong, I understand that they make some "noise" in your codebase, but this noise is just moved in our own projects, where an additional _pipe+map_ for every Inner Observable (I'm talking about flattening operators here) makes the code more difficult to read. I've used them whenever I needed to and so did a lot of my colleagues, and we've always found them useful. Are we sure "almost nobody uses them"? / End of rant :D 

Moreover, it appears that with version 6.5 (almost one year ago?) the signatures for **forkJoin** and **combineLatest** which don't accept an array have been deprecated. For some reason my IDE never highlighted it, I just thought there was just one more way to use these operators (which would be absolutely fine in my opinion). In the docs, the deprecated signatures (without arrays) are used **just everywhere**.

## Consistency
[...]

PR #6379, opened 2021-05-07
docs: add deprecation notes to documentation pages

tmair:
**Description:**
This PR adds deprecation notes to the documentation for single method and function overrides. Additionally a strikethrough is also added to the method signature. The source of the changes was the angular.io project, although only the pieces necessary for the deprecations where included.

There are still some places that are a little odd:
* if all function overloads are deprecated, the deprecation note is shown for the function and all of its overloads
* if all function overloads are marked as deprecated the function is *not* marked with a deprecated badge
* the deprecation notes contain literal links to http://rxjs.dev/... Their purpose is probalby to accomodate IDEs to show the correct link

I am planing on working on some of the issues above in follow up PRs to keep this one simple as I think that this PR already adds value to the documentation.

**Examples:**
`toPromise` method
grafik
grafik

`pluck` function
grafik
grafik

`publishReplay` function
grafik
grafik

#5404

### DIFF
--- a/docs_app/src/styles/2-modules/_api-pages.scss
+++ b/docs_app/src/styles/2-modules/_api-pages.scss
@@ -106,3 +106,8 @@
     color: $pink;
   }
 }
+
+
+.deprecated-api-item {
+  text-decoration: line-through;
+}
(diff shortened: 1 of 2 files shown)


benlesh:
This looks awesome!

cc @niklas-wortmann

benlesh (approved):

niklas-wortmann (approved):
that's an awesome contribution. Many thanks for that

PR #6380, opened 2021-05-07
docs: fix broken links in documentation

tmair:
**Description:**
This PR fixes broken links within the documentation that are reported by the build of the documentation.

Two examples of the broken links can be found on the [Global Config](https://rxjs.dev/api/index/interface/GlobalConfig) documentation page.

None

### DIFF
--- a/src/internal/Notification.ts
+++ b/src/internal/Notification.ts
@@ -44,7 +44,7 @@ export class Notification<T> {
    * Creates a "Next" notification object.
    * @param kind Always `'N'`
    * @param value The value to notify with if observed.
-   * @deprecated Internal implementation detail. Use {@link createNext} instead.
+   * @deprecated Internal implementation detail. Use {@link Notification#createNext createNext} instead.
    */
   constructor(kind: 'N', value?: T);
   /**
@@ -52,13 +52,13 @@ export class Notification<T> {
    * @param kind Always `'E'`
    * @param value Always `undefined`
    * @param error The error to notify with if observed.
-   * @deprecated Internal implementation detail. Use {@link createError} instead.
+   * @deprecated Internal implementation detail. Use {@link Notification#createError createError} instead.
    */
   constructor(kind: 'E', value: undefined, error: any);
   /**
    * Creates a "completion" notification object.
    * @param kind Always `'C'`
-   * @deprecated Internal implementation detail. Use {@link createComplete} instead.
+   * @deprecated Internal implementation detail. Use {@link Notification#createComplete createComplete} instead.
    */
   constructor(kind: 'C');
   constructor(public readonly kind: 'N' | 'E' | 'C', public readonly value?: T, public readonly error?: any) {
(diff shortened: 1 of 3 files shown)


niklas-wortmann (approved):
thanks a lot for fixing this

src/internal/Notification.ts
```diff
@@ -44,21 +44,21 @@ export class Notification<T> {
    * Creates a "Next" notification object.
    * @param kind Always `'N'`
    * @param value The value to notify with if observed.
-   * @deprecated Internal implementation detail. Use {@link createNext} instead.
+   * @deprecated Internal implementation detail. Use {@link Notification#createNext createNext} instead.
```
cartant, line 47: Do links declared like this render correctly in the app? My understanding is that custom link text had to be placed after a `|`, like this:

```
{@link Notification#createNext | createNext}
```

https://tsdoc.org/pages/tags/link/

benlesh: @niklas-wortmann @tmair thoughts?

tmair: > Do links declared like this render correctly in the app?
Yes they do. And the reason why is that documentation generation does not depend on `tsdoc` but on the `dgeni` and `degeni-packages` packages. For `@link` extraction the following two files are important: 
* https://github.com/ReactiveX/rxjs/blob/86cce743334cf1b05f2a9f2ad182452aa3785a6e/docs_app/tools/transforms/links-package/inline-tag-defs/link.js
* https://github.com/ReactiveX/rxjs/blob/86cce743334cf1b05f2a9f2ad182452aa3785a6e/docs_app/tools/transforms/links-package/services/getLinkInfo.js

There is a RegExp in the `link.js` file https://github.com/ReactiveX/rxjs/blob/86cce743334cf1b05f2a9f2ad182452aa3785a6e/docs_app/tools/transforms/links-package/inline-tag-defs/link.js#L1 for parsing `@link` content which is basically of the form `${url} ${link name}`. So that is the reason why this works for the documentation home page.


There is however another issue that is related to how `@links` are parsed and shown in IDEs. As there is no standard for JS/TS documentation each IDE or Editor seems to ship its own slightly different version for `@link` processing and parsing. These are the ones I know of but this list is not exhaustive:
* VSCode: There has been a recent addition to better support `@link` processing and display in TypeScript which then also benefits VSCode. It should be included in TS 4.3 https://github.com/microsoft/TypeScript/pull/41877 
* TSDoc: TSDoc is an effort to add a standard for TS documentation. The `@link` proposal is still in discussion (There are a number of tricky issues regarding this simple tag https://github.com/microsoft/tsdoc/issues/9). Currently it seems that TSDoc is used by APIExtractor to  generate documentation (similar to dgeni)

So long story short: I would like to look into the recent addition of `@link` parsing to TS and its compatibility or incompatibility with TSDoc a little closer. So for now I am fine if this PR still stays open a little longer. I don't exactly know when I will have time to dig a little deeper into all of this as I am not too familiar with `dgeni` or the TSDoc project.

jakovljevic-mladen: Actually, Angular has so many examples of these links:

https://github.com/angular/angular/blob/cd252b99fe04bcde2a31e2314f9eb920b2d38e49/packages/forms/src/directives/reactive_directives/form_group_name.ts#L47

This is how it looks like in the docs:
https://angular.io/api/forms/FormGroupName#access-individual-controls-in-the-group

cartant: @tmair Thanks for the super-thorough comment. 👍

PR #6384, opened 2021-05-08
chore: remove Node v11 from CI

cartant:
**Description:**

This PR removes Node v11 from the CI configuration. Node v12 is no longer active LTS - it's now maintenance - so IMO a version that's older than that and wasn't an LTS version isn't something we need to continue to support. And it doesn't support `globalThis`, so it's annoying.

### DIFF
--- a/.github/workflows/ci_main.yml
+++ b/.github/workflows/ci_main.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [ '11', '12', '13', '14', '15' ]
+        node: [ '12', '13', '14', '15' ]
 
     name: Node ${{ matrix.node }} build
 


kwonoj (approved):
> it doesn't support globalThis, so it's annoying

V7 is binded by ie11 so regardless of this change we'll suffer still?

cartant:
Yeah. On closer inspection, `globalThis` is used only in [some specs](https://github.com/ReactiveX/rxjs/search?q=globalThis) and only in guards. It's not used in the library code itself. I still think we should drop Node v11 though. Do we actually indicate the minimum required Node version anywhere?

kwonoj:
We once did attempt globalthis and didn't land for the same reason: https://github.com/ReactiveX/rxjs/pull/5560

We don't explicit document version we support especially around node. I think it's worth to note (or use engines property in pkg.json?) Removing 11 on ci itself isn't breaking change and since we have fixed transpilation target from tsc I think it's fine for now.

cartant:
It kinda seems like a problem if it's expecting something for 11, 'cause that ain't gonna happen:

Capture

kwonoj:
I turned off checker. fyi, it's under settings / branch protection for later if you need to adjust.

Issue #6382, opened 2021-05-08
`debounceTime` does not emit with `jasmine.clock.tick` in v7


dpurp:
## Bug Report
\cc @leggechr 

**Current Behavior**
In v7, `debounceTime` does not emit any values when installing the jasmine clock and using `tick`. (This works in v6)

**Expected behavior**
rxjs is compatible with jasmine.clock, or there is a documented workaround.

**Reproduction**
Note the reproductions below contain examples using the jasmine clock (which fails in v7) and the native clock (which succeeds).

v7: https://stackblitz.com/edit/rxjs-v7-debouncetime?file=test.ts
v6: https://stackblitz.com/edit/rxjs-v6-debouncetime?file=test.ts

```ts
describe('failing time tests with jasmine clock', () => {
  beforeEach(() => {
    jasmine.clock().install();
  });

  afterEach(() => {
    jasmine.clock().uninstall();
  });

  it('debounces time', async () => {
    const duration = 10;
    const source = new Subject<string>();
    const debouncedSource$ = source.pipe(
      debounceTime(duration),
      shareReplay(1)
    );
    debouncedSource$.subscribe();

    source.next('should not be emitted');
    source.next('expected value');
    jasmine.clock().tick(duration + 1); // Wait for debounce to happen.

    expect(await firstValueFrom(debouncedSource$)).toEqual('expected value');
  });
});
```

cartant:
Thanks. This is a bug - 'cause it's an inadvertent breaking change - as `setInterval` is captured in the interval provider:

https://github.com/ReactiveX/rxjs/blob/53ef49459b03819022c2c81e57650c1cdc42fedc/src/internal/scheduler/intervalProvider.ts#L18-L25

Instead, the provider should obtain the `setInterval` and `clearInterval` functions via `globalThis`.

cartant:
Nope. That's not the problem. The `debounceTime` implementation now uses the scheduler's `now` function:

https://github.com/ReactiveX/rxjs/blob/53ef49459b03819022c2c81e57650c1cdc42fedc/src/internal/operators/debounceTime.ts#L84-L86

and

https://github.com/ReactiveX/rxjs/blob/53ef49459b03819022c2c81e57650c1cdc42fedc/src/internal/operators/debounceTime.ts#L99-L100

So the issue is that Jasmine purports to mock the clock, but it does not patch `Date.now`.

@benlesh

cartant:
[...]

PR #6386, opened 2021-05-08
docs: mention jasmine.clock().mockDate() re: debounceTime

cartant:
**Description:**

This PR adds mention of `jasmine.clock().mockDate()` to the v6/v7 changes doc. Mostly 'cause Jasmine is so widely used in the Angular community, 'cause it does not mock `Date.now()` by default and 'cause `debounceTime` is a frequently-used operator.

It's not really a breaking change, but if people's tests suddenly fail - 'cause Jasmine is ... Jasmine - it might as well be.

🤬

**Related issue (if exists):** #6382


benlesh (approved):

benlesh:
Good call!

Issue #5731, opened 2020-09-15
bug(partition): Subscribes to the source twice, when it should probably multicast


benlesh:
Currently `partition` will subscribe to whatever observable you get twice. Given that it is supposed to "partition", which sounds like splitting one thing into several parts, I'd think it should multicast the source.

ALSO: The `partition` _operator_ just needs to be removed in favor of `groupBy`, as it doesn't make any sense as an operator with `pipe`.

cartant:
Related: https://github.com/ReactiveX/rxjs/issues/3807

And there is a `splitBy` in `rxjs-etc` that I mentioned in [this comment](https://github.com/ReactiveX/rxjs/issues/3807#issuecomment-563187338).

benlesh:
This is still an issue with the existing static `partition`, and we should address it sometime. It's debatable as to whether it's a breaking change.

Ayudh:
@benlesh can I work on this?

Issue #4419, opened 2018-12-12
Partition operator example throws typescript error


luckylooke:
### Documentation Related To Component:

partition operator
https://www.learnrxjs.io/operators/transformation/partition.html
https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/partition.ts

### Please check those that apply

- [ ] typo
- [ ] documentation doesn't exist
- [ ] documentation needs clarification
- [x] error(s) in example
- [ ] needs example

### Description Of The Issue

Opening the [example in Stackblitz](https://stackblitz.com/edit/typescript-gr3ljs?file=index.ts&devtoolsheight=100) we are getting typescript error:
```
Argument of type 'UnaryFunction<Observable<number>, [Observable<number>, Observable<number>]>' is not assignable to parameter of type 'OperatorFunction<any, any>'.
  Type '[Observable<number>, Observable<number>]' is not assignable to type 'Observable<any>'.
    Property '_isScalar' is missing in type '[Observable<number>, Observable<number>]'.
```
typescript is not complaining about this usage:
```
const [evens, odds] = partition((val: number) => val % 2 === 0)(source);
```
Result is the same.

I did not check other example or other online editors but IMHO it would be the same.

I do not have more time for making PR, also I am not sure about whether implementation should be fixed not to confuse typescript or syntax in examples should follow the usage I have provided here.
Sorry.

Thank you for your time and happy coding :)

EDIT:
I have just check jsBin and jsFiddle. I have find out that they are not relevant as they does not support typescript errors. I have tested `const num: Number = 'text';` and they both did not complain at all.

jgbpercy:
That error is the "correct" error. Partition isn't usable within pipe because it's not really an operator, for the reason I explained in 4119:

Partition returns a UnaryFunction, not an OperatorFunction, and pipe only takes OperatorFunctions as arguments. It seems like partition was broken (at least when used in pipe) with the move to pipeable operators, but the clean up work to deprecate/replace it hasn't been priority as partition isn't that widely used. See these issues:

https://github.com/ReactiveX/rxjs/issues/3797
https://github.com/ReactiveX/rxjs/issues/2995
https://github.com/ReactiveX/rxjs/issues/3807

So TS is right to complain about partition being used in pipe, but you can still use partition as a function in its own right, I believe.

luckylooke:
Ok, so then examples should be fixed as they use partition as pipe now:
```
[...]

Issue #3807, opened 2018-06-06
Deprecate partition, add splitBy method that multicasts the source sequence


trxcllnt:
Related to #3797, the ideal replacement would be a method that mutlicasts the source sequence and returns two Observable "lanes", and sends values down one or the other lane based on the predicate. This is conceptually similar to `groupBy` with only two groups, except it returns an Array of the grouped Observables instead of an Observable of the groups since the groups can be known ahead of time.

lmanzke:
Seconded. To my mind, the partition operator is fully useful yet: There needs to be another one that does not subscribe twice.

For trivial examples like in the docs, it works totally fine. Behind the scenes however, it subscribes to the observable behind partition twice which might double the work the code does.

In my concrete case, I followed the docs to split success and error cases. However, instead of calculating if a number is odd or even, I did an API call (with flatMap / catchError in case anything goes wrong). In my opinion, this is quite a common use case. If you now partition this by success / failure and subscribe to both partitioned streams, the API call is actually performed twice because there are two subscriptions that are merged back into one. This might be undesirable because it could double the load on the server (if there is no caching in place).

For this scenario, I might be missing a better operator. In this case, I am sorry. However, the splitBy that is described here seems like a perfect match for the situation I'm facing. As a workaround, i am using a custom operator that uses two subjects and passes elements on correctly, but I doubt this is production-ready yet.

cartant:
FWIW, I implemented this, but never got around to discussing it or opening a PR: https://github.com/cartant/rxjs-etc/blob/master/source/operators/splitBy.ts

lmanzke:
That looks very promising, thanks. Maybe it would be a good idea to merge it back 😄

nathggns:
Would be good to get splitBy into rxjs.

benlesh:
Core Team Meeting: Have @cartant move this into core, whenever he gets the chance. It seems like there's a good deal of interest.

benlesh:
Okay, after a PR and some careful deliberation, we decided we aren't going to pursue getting this into core right now. For reasons that are outlined here: https://github.com/ReactiveX/rxjs/pull/6388#issuecomment-846094188

PR #6388, opened 2021-05-08
buckets and splitBy operators

backbone87:
- [x] Add the operator to Rx
- [x] It must have a `-spec.ts` tests file covering the canonical corner cases, with marble diagram tests
- [x] The spec file should have a type definition test at the end of the spec to verify type definition for various use cases -- *is the type test in splitBy-spec sufficient?*
- [ ] The operator must be documented in JSDoc style in the implementation file, including also the PNG marble diagram image -- *how to make the marble diagram?*
- [x] The operator should be listed in `docs_app/content/guide/operators.md` in a category of operators
- [x] The operator should also be documented. See [Documentation Guidelines](../CONTRIBUTING.md).

**Description:**

Introduces 2 new operators `buckets` and `splitBy`.

`buckets` uses a hash function to partition values across multiple multicasted (hot) observables. This behaves similar to `groupBy`, but the groups are known immediately.

`splitBy` is the special case of `buckets` where the count of buckets is 2. This is similar to `partition`, but has a usage ergonomy more similar to `groupBy`.

implementation is based upon `bucketBy` and `splitBy` in `rxjs-etc` by @cartant 

**Related issue (if exists):** https://github.com/ReactiveX/rxjs/issues/3807 https://github.com/ReactiveX/rxjs/issues/4419 https://github.com/ReactiveX/rxjs/issues/5731

### DIFF
--- a/spec/operators/splitBy-spec.ts
+++ b/spec/operators/splitBy-spec.ts
@@ -1,0 +1,27 @@
+/** @prettier */
+import { mergeMap, splitBy } from 'rxjs/operators';
+import { TestScheduler } from 'rxjs/testing';
+import { observableMatcher } from '../helpers/observableMatcher';
+
+describe('splitBy()', () => {
+  let rxTest: TestScheduler;
+
+  beforeEach(() => {
+    rxTest = new TestScheduler(observableMatcher);
+  });
+
+  it('should split values', () => {
+    rxTest.run(({ cold, expectObservable }) => {
+      const source = cold(' --a-b-c-a-b-c-|');
+      const x = cold('      --a-----a-----|');
+      const y = cold('      ----b-c---b-c-|');
+      const expected = '    (xy)----------|';
+
+      const split = source.pipe(
+        splitBy((value) => value === 'a'),
+        mergeMap((splits) => splits)
+      );
+      expectObservable(split).toBe(expected, { x, y });
+    });
+  });
+});
--- a/src/internal/operators/buckets.ts
+++ b/src/internal/operators/buckets.ts
@@ -1,0 +1,135 @@
+import { Observable } from '../Observable';
+import { Subject } from '../Subject';
+import { OperatorFunction } from '../types';
+import { operate } from '../util/lift';
+import { OperatorSubscriber } from './OperatorSubscriber';
+
+export interface BucketsOptions<T> {
+  /**
+   * The hash function used to partition the values into their buckets. The
+   * default hash function converts the values into numbers, if they are not
+   * already.
+   */
+  hashFn?: (value: T) => number;
+  /**
+   * The factory used to create the subject that will connect the source observable to
+   * multicast consumers.
+   */
+  connector?: (bucketIndex: number) => Subject<T>;
+}
+
+/**
+ * Maps a source observable into an observable that does a single immediate and
+ * synchronous emit that contains an array of multicasted ("hot") observables
+ * - the buckets - into which the source values are partitioned into according
+ * to a given hash function applied over each source value.
+ *
+ * This enables similar behavior to the {@link groupBy} operator with the
+ * difference that all possible groups are available immediately when using
+ * `buckets`.
+ *
+ * ## Example
+ *
+ * ```ts
+ * import { interval } from 'rxjs';
+ * import { buckets, mergeMap, mapTo, mergeAll, groupBy } from 'rxjs/operators';
+ *
+ * interval(1000).pipe(
+ *   buckets(3),
+ *   mergeAll(),
+ *   mergeMap((bucket, index) => {
+ *     console.log(`bucket ${index} discovered`);
+ *
+ *     return bucket.pipe(mapTo(`in bucket ${index}`));
+ *   }),
+ * ).subscribe(console.log);
+ *
+ * // Result
+ * // "bucket 0 discovered"
+ * // "bucket 1 discovered"
+ * // "bucket 2 discovered"
+ * // "in bucket 0"
+ * // "in bucket 1"
+ * // "in bucket 2"
+ * // "in bucket 0"
+ * // "in bucket 1"
+ * // "in bucket 2"
+ * // "in bucket 0"
+ * // ...
+ *
+ * // groupBy is similar but different
+ *
+ * interval(1000).pipe(
+ *   groupBy((value) => value % 3),
+ *   mergeMap((group) => {
+ *     console.log(`group ${group.key} discovered`);
+ *
+ *     return group.pipe(mapTo(`in group ${group.key}`));
+ *   }),
+ * ).subscribe(console.log);
+ *
+ * // Result
+ * // "group 0 discovered"
+ * // "in group 0"
+ * // "group 1 discovered"
+ * // "in group 1"
+ * // "group 2 discovered"
+ * // "in group 2"
+ * // "in group 0"
+ * // "in group 1"
+ * // "in group 2"
+ * // "in group 0"
+ * // ...
+ *
+ * ```
+ *
+ * @param bucketCount The number of buckets
+ * @return A function that returns an Observable that does a single emit
+ *  containing the array of buckets
+ *
+ * @see {@link groupBy}
+ * @see {@link partition}
+ * @see {@link splitBy}
+ */
+export function buckets<T>(bucketCount: number, options: BucketsOptions<T> = {}): OperatorFunction<T, Observable<T>[]> {
+  const { hashFn = Number, connector = () => new Subject() } = options;
+
+  return operate((source, subscriber) => {
+    const { subjects, bucketFor, observables } = indexedBuckets(bucketCount, hashFn, connector);
+    const observers = [...subjects, subscriber];
+    const connection = new OperatorSubscriber<T>(
+      subscriber,
+      (value) => {
+        try {
+          bucketFor(value).next(value);
+        } catch (e) {
+          observers.forEach((observer) => observer.error(e));
+        }
+      },
+      () => void observers.forEach((observer) => observer.complete()),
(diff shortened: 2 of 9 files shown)


spec/operators/splitBy-spec.ts
```diff
@@ -0,0 +1,35 @@
...
+      expectObservable(split).toBe(expected, { x, y });
+    });
+  });
+
+  it('should infer partition types when using a type guard predicate', () => {
+    of(5, 'a', []).pipe(
+      splitBy((value): value is number => typeof value === 'number'),
+      mergeMap(([numbers, strings]) => merge(numbers.pipe(map((value) => value * value)), strings.pipe(map((value) => value.length))))
+    );
+  });
```
cartant, line 34: We favour using dtslint for type-related tests:

https://github.com/ReactiveX/rxjs/blob/c23e567ccaee4a9aa2a7046519e7cf46a7191a3f/spec-dtslint/operators/filter-spec.ts#L37-L40

The tests are never executed, instead there is a lint rule that runs evaluates the types and dtslint expectations. To run them, use `npm run dtslint`.

src/internal/operators/buckets.ts
```diff
@@ -0,0 +1,135 @@
...
+import { operate } from '../util/lift';
+import { OperatorSubscriber } from './OperatorSubscriber';
+
+export interface BucketsOptions<T> {
+  /**
+   * The hash function used to partition the values into their buckets. The
+   * default hash function converts the values into numbers, if they are not
+   * already.
+   */
+  hashFn?: (value: T) => number;
```
cartant, line 13: Personally, I'm not a fan of abbreviations. I'd prefer `hasher`, instead, but this can be bikeshedded later.

src/internal/operators/splitBy.ts
```diff
@@ -0,0 +1,99 @@
...
+ *  containing the two observables for matching and non-matching values
+ *
+ * @see {@link buckets}
+ * @see {@link groupBy}
+ * @see {@link partition}
+ */
+export function splitBy<T>(
+  predicate: (value: T) => boolean,
+  options: SplitByOptions<T> = {}
+): OperatorFunction<T, [Observable<T>, Observable<T>]> {
```
cartant, line 92: This function has overload signatures and they hide the implementation signature. So you'll need this to be an overload signature in its own right - even though the implementation is exactly the same. That's my understanding, at least.

cartant (commented):
Some initial comments. Will review it properly, later. Also, for PRs that change the public API, you will need to use `npm run api_guardian:update`.

src/internal/operators/buckets.ts
backbone87: i also just picked this up from somewhere, bikeshed away ;)

src/internal/operators/splitBy.ts
backbone87: actually the options are just missing from the other signatures. typescript only shows the overloads to the callers not the impl signature. the only requirement for the impl signature is that is has to be assignable to each overload signature.

will fix the missing argument

backbone87:
in regards to naming: is `intoBuckets` a better choice?

cartant:
I think it could probably be a single operator - `splitBy` - with two signatures with different option types - `SplitByPredicateOptions` and `SplitByHashOptions` - with the predicate and count parameters moved into the respective options types. IDK. Much bikeshedding is possible.

backbone87:
personally, i dont like merging them into a single operator. too overloaded from user & dev perspective. but that's just my feeling. `splitInto(3)`, `splitBy(({ name }) => name === 'rxjs')` just reads/sounds nice (not a native speaker though)

cartant:
TBH, I was not planning on bringing `bucketBy` - or whatever - into this codebase. I was only going to bring in `splitBy`. I don't anticipate there being much appetite for an additional operator. Especially one that hasn't been requested. Expect this to be bikeshedded, too.

benlesh:
So looking over this, at first blush. `buckets` looks like a specialized `groupBy`. and `splitBy` is a specialized `buckets`. So it makes me wonder if we need either of these. 🤔  Just thinking out loud here.

```ts
const buckets = (count) => groupBy((value, i) => i % count)
// or
const buckets = ({ hasher }) => groupBy(hasher);
// or
const splitBy = (predicate) => groupBy(predicate);
```

Am I off-base? What am I missing?

benlesh:
To be clear @backbone87 ... this is awesome work! And I really appreciate it. However, we're going to be on the hook to support this API long-term, so I want to be sure we're doing the right thing.

benlesh:
Another thing to consider... And `groupBy` handles this... but there's a scenario with operators like this, where the consumer can take the observable out of the result and subscribe to it, then early-unsub from the source. In those cases, the expectation would be that the created "buckets" or "groups" would still work until they were unsubbed. At least that's true for `groupBy`, so I guess I would expect the same behavior here.

```ts
const source = interval(1000);
const result = source.pipe(buckets(3));

const capturedBucketSubs = [];

const outerSubs = result.subscribe(bucket => {
   capturedBucketSubs.push(capturedBucketSubs.subscribe(console.log));
});

// unsubbing outer first shouldn't kill existing inners
setTimeout(() => {
  outerSubs.unsubscribe();
}, 5000);

setTimeout(() => {
  for (const subs of capturedBucketSubs) {
    subs.unsubscribe();
  }
}, 10000);
```

I'm willing to agree this behavior is SUPER debatable, but there it is, and that's how groupBy has worked for years, for better or worse.

benlesh:
Sorry, too, @cartant, if I'm just springing this on you. When we talked about this at the core team meeting, I guess I was thinking `splitBy` did something different, but I'm not sure now, I guess, what I thought it did that `groupBy` couldn't do.

backbone87:
`buckets` is not a special case of `groupBy`. check the examples in the doc comments which demonstrate the difference in behavior. the buckets are also subjects, so unsubbing from them should not change anything, but i can add a test for this.

cartant:
@benlesh As I mentioned above, I was planning on ignoring `bucketBy`/`buckets` when I eventually got around to dealing with #3807. I don't think it adds anything that's particularly useful. It's just how `splitBy` was implemented in `rxjs-etc`. And adding random operators to that package does not come with the implications of adding operators to the core.

FWIW, I kinda agree with you on dumping this in favour of `groupBy` and a `partition` deprecation that links to a `deprecations/partition.md` document that explains how to use `groupBy` instead.

When I added `splitBy` to `rxjs-etc`, `groupBy` didn't have the typings that it does now. 12 months or so ago [Moshe added typings that made it play nice with user-defined type guards](https://github.com/ReactiveX/rxjs/pull/5441), so it composes pretty nicely:

```ts
dogsAndCats.pipe(
  groupBy(isDog),
  mergeMap(grouped => grouped.key
    ? /* Observable<Dog> */ grouped.pipe(tap(() => console.log('woof')))
    : /* Observable<Cat> */ grouped.pipe(tap(() => console.log('meow')))
)
```

I don't know that there is any advantage in using `splitBy` instead and being presented with the two observables up front - this is the difference mentioned in [the original issue](https://github.com/ReactiveX/rxjs/issues/3807):

> This is conceptually similar to groupBy with only two groups, except it returns an Array of the grouped Observables instead of an Observable of the groups since the groups can be known ahead of time.

Rather, I think `splitBy` will be more difficult for peeps to grok, as the two known-ahead-of-time observables will need to be combined into a single observable:

```ts
dogsAndCats.pipe(
  groupBy(isDog),
  mergeMap(([dog$, cat$])=> merge(
    dog$.pipe(tap(() => console.log('woof'))),
    cat$.pipe(tap(() => console.log('meow')))
  )
)
```

**TL/DR: I don't see a compelling case for `splitBy` either, so close this - and the related issues - and document the `partition` deprecation with examples on the docs site.**

benlesh:
@cartant maybe if we get everything in a monorepo, `rxjs-etc` might be a solid candidate to join this repository, so we can direct things over there? Obviously debatable, but it might open the door for moving lesser used core operators over, as well. Just something to think about.

backbone87:
no hard feelings here on my side, a little bit of disappointment, but I can handle this, if it doesn't fit, then leave this out.

personally, I have used partition only on very few occasions, but when I did, I definitely wasn't happy with its behavior as it is right now and would have preferred the splitBy behavior. also, these situations could have most likely also been solved with groupBy.

benlesh:
After discussing this PR in the Core Team meeting, I'm so sorry to say that I don't think we're going to go ahead with it at this time, and we should probably close the related issue. That issue, TMK, was filed at a time when `groupBy` was a little less capable than it is now. While `splitBy` _is indeed_ a little different, we've decided that the difference isn't sufficient enough to justify supporting the API as part of the library for years to come.

Thank you so much for your hard work, @backbone87, we all really do appreciate it. I'm sorry that this isn't something we can merge into core for now. Especially given that it was just trying to implement something that was requested by a core team member years ago.

Thank you again, @backbone87, and I hope this doesn't put you off from further contributions.

backbone87:
hm ok, no problem. i suspected it.
i have a few questions:
- what is the way forward with `partition`? the operator is already deprecated, but the behavior of the factory is still to consume twice instead of multicasting
- is there a way to achieve the bucket/splitBy behavior with another operator or pipeline of operators?
- would it be an alternative to add options to groupBy to enable a similar behavior to bucket/splitBy. something like `groupBy(({ key }) => key, { seed: () => of('key1', 'key2') })` where `seed` emits keys for which groups are eagerly opened and when it completes the `groupBy` completes to downstream, but still receives values from upstream and routes them into existing groups. unrouteable upstream values will trigger a stopped notification hook and `seed` defaults to `() => NEVER`, which means the `groupBy` does not create any groups eagerly and completes with the source.

Polyterative:
I'm also interested in a way to replicate the `splitBy` with classic operators

PR #6389, opened 2021-05-09
update Observable.md

xp44mm:
subscriber -> observer


cartant (commented):
I'm not sure about this change. It actually _is_ a `Subscriber` that's passed to the function:

https://github.com/ReactiveX/rxjs/blob/c23e567ccaee4a9aa2a7046519e7cf46a7191a3f/src/internal/Observable.ts#L37

One of the reasons for this is that `Subscriber` has a `closed` property which is used to signal explicit unsubscription. For example:

https://github.com/ReactiveX/rxjs/blob/c23e567ccaee4a9aa2a7046519e7cf46a7191a3f/spec/operators/share-spec.ts#L432-L439

At the moment, the documentation is accurate when it refers to a subscriber, rather than an observer.

xp44mm:
There are no subscribers in Glossary.

benlesh:
A long time ago I switched the names there from `observer` to `subscriber`, because truly it's not _just_ an observer. For example, you can call `unsubscribe` on it. Or you can add teardown logic to the subscription. It's both an "observer" and a "subscription". It's a "subscriber".

Thank you for the thought and effort, but I don't think we want to make this change.

PR #6391, opened 2021-05-10
docs: include additional interfaces to docs

jakovljevic-mladen:
**Description:**
These types are exported so they can be included to the docs. I was thinking about where to include them, to `src/index.ts` or `src/operators/index.ts`, but I chose `src/index.ts`. The reason is that on the API list page, we want to add exported interface to the **index** section, rather than to the **operators** section, like this:

image

Another note: I added `TimeoutInfo` to exports because `TimeoutConfig` is now being exported and it's got [with](https://github.com/ReactiveX/rxjs/blob/86cce743334cf1b05f2a9f2ad182452aa3785a6e/src/internal/operators/timeout.ts#L34) property that uses `TimeoutInfo` as a callback property.

Somewhat #6247 and a [related comment](https://github.com/ReactiveX/rxjs/pull/6247#issuecomment-825470866).


src/internal/operators/throttle.ts
```diff
@@ -61,9 +61,10 @@ export const defaultThrottleConfig: ThrottleConfig = {
  */
 export function throttle<T>(
   durationSelector: (value: T) => ObservableInput<any>,
-  { leading, trailing }: ThrottleConfig = defaultThrottleConfig
+  config: ThrottleConfig = defaultThrottleConfig
```
jakovljevic-mladen, line 64: This change is added just because we could have a better documentation. Instead of [this](https://rxjs.dev/api/operators/throttle):

image


we can have this:

image

src/index.ts
```diff
@@ -95,6 +95,12 @@ export { NEVER } from './internal/observable/never';
 
 /* Types */
 export * from './internal/types';
+export { ConnectConfig } from './internal/operators/connect';
```
benlesh, line 98: While I think it makes sense that you put these here, it's probably better just to have them next to the operators they're with. Just to help show the relationship a little more clearly.

src/internal/operators/throttle.ts
```diff
@@ -61,9 +61,10 @@ export const defaultThrottleConfig: ThrottleConfig = {
...
 ): MonoTypeOperatorFunction<T> {
   return operate((source, subscriber) => {
+    const { leading, trailing } = config;
```
benlesh, line 67: interesting. Was this change necessary for documentation reasons? That's sort of bananas.

jakovljevic-mladen:
Included new `groupBy` interfaces. They had to be exported though... IMO, it makes sense to export them, that way they could be reused:
```ts
const options: GroupByOptionsWithElement<string, string, string> = {
  element: value => value.trim()
};
const grouped$ = groupBy(val => val, options);
const groupedLowerCase$ = groupBy(val => val.toLowerCase(), options);
```

jakovljevic-mladen:
Sorry for this unintended commit, it should've been placed to another branch.

src/index.ts
jakovljevic-mladen: @benlesh, since #6488 has been merged to master, I think it would be wise to move exported interfaces back to this file and keep them exported only here.
The first reason is to avoid duplication - we don't want to export stuff in two interfaces (also, dgeni tools don't like duplicated stuff; also duplication is hard to maintain - what if a new interface is added, would a creator of that interface know that it should be added to both files?).
The second reason, if they would remain in `/src/operators/index.ts` file: that file will most probably be removed, so we want to have all exported interfaces in one place before removal so we don't remove exported interfaces accidentally.
Please let me know your thoughts on this. Thanks.

PR #6393, opened 2021-05-10
refactor(mapTo): smaller implementation

jakovljevic-mladen:
**Description:**
Smaller `mapTo` implementation.

None

### DIFF
--- a/src/internal/operators/mapTo.ts
+++ b/src/internal/operators/mapTo.ts
@@ -1,6 +1,5 @@
 import { OperatorFunction } from '../types';
-import { operate } from '../util/lift';
-import { OperatorSubscriber } from './OperatorSubscriber';
+import { map } from './map';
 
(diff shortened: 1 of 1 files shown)


kwonoj:
My memory's very vague - didn't we try something similar and back it off before? was it different operator? /cc @benlesh

benlesh:
Honestly, we talked about doing this at some point (I'm not sure if it was in an issue or a discussion). I mean, it's _obvious_.

However, it was decided at the time that `mapTo` was so popular that we should allow it to be as optimized as possible. That said, I'm not entirely sure how much better it is the way it was over the way this PR is proposed.

In fact, I'm inclined to deprecate all `xMapTo` operators, just because they're silly when they're just hiding a closure. `mergeMapTo(x)` and `mergeMap(() => x)` no matter how we'd reasonably write them are pretty much the same thing. Just one does the closure for you.

Issue #5487, opened 2020-06-13
first/tap operators should unsubscribe before emit next value


Enuvid:
## Bug Report

**Current Behavior**

Currently, we are unsubscribing from source observables after emit the last message next to the chain. In some cases, it can raise incorrect and unexpected behavior (it seems to me).
In example below, we are still have processing new messages second observable in `merge` after `first` have emitted the first message. But expected behavior is that `first` operator must unsubscribe from all sources on receiving the first message.

Also, we have strange behavior: we can emit error from second observable and this exception will go through `first` operator and handle (in my example)  in `catchError`, but if we try to pass just a value through `first`, this operator (actually `tap` operator inside of `first`) will omit this message.

**Reproduction**

```js
const notificator = new Subject();

merge(
  timer(1)
    .pipe(
      tap((m) => console.log('value from first: ', m)),
    ),
  notificator.pipe(
    tap(() => console.log('value from notificator')),
    switchMap(() => throwError(new Error('Some error'))),
))
  .pipe(
    first(),
    catchError((err) => {
      console.log('Catching error: ', err)
      return throwError('Continue receive error after first value');
    }),
    tap((value) => {
      console.log('value: ', value);
      notificator.next(null);
    })
  ).subscribe(
    (value) => console.log('Next value: ', value),
    (err) => console.log('Error: ', err),
  );
```
Output:
```
value from first: 0
value: 0
value from notificator
Catching error: Error {}
Error: Continue receive error after first value
```

You can try it [here](https://stackblitz.com/edit/rxjs-etniuk).

**Expected behavior**

As I told before, I think that we shouldn't get any effects (include errors) from source observables after the limit of allowed values in stream reached. And more, we should unsubscribe from them, to prevent any calculations inside source chain, we should unsubscribe before passing the last value next.

**Environment**
- Runtime: Chrome/83.0.4103.97
- RxJS version: 6.5.2

**Possible Solution**

In `take` operator we overriding next handler to: 

https://github.com/ReactiveX/rxjs/blob/3f3c614d8b86c68d271a1db995ff11c61842e757/src/internal/operators/take.ts#L79-L89

I suggest changing them to make unsubscribe stuff before emitting next value. Also, this will allow us to omit ```if (count <= total)``` this strange check, because if we already unsubscribed from source that we will no longer receive a message.

If this change can affect something (backward compatibility), I think we should at least override `_error` handler to prevent passing errors. But I think, more correct just unsubscribe before next.

[...]

PR #6394, opened 2021-05-10
fix(take): notifies errors after limit is reached

backbone87:
**Description:**
provides a failing test case for an edge case of take and dependent operators as presented in https://github.com/ReactiveX/rxjs/issues/5487

Will add a fix, if a change in behavior is agreed upon.

**Related issue (if exists):** fixes https://github.com/ReactiveX/rxjs/issues/5487

### DIFF
--- a/spec/operators/take-spec.ts
+++ b/spec/operators/take-spec.ts
@@ -1,7 +1,7 @@
 /** @prettier */
 import { expect } from 'chai';
-import { take, mergeMap } from 'rxjs/operators';
-import { of, Observable, Subject } from 'rxjs';
+import { merge, Observable, of, Subject } from 'rxjs';
+import { mergeMap, take, tap } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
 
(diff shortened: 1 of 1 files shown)


cartant:
The behaviour reported in that issue is a bug. It's supposed to stop listening.

IMO, I would deem this a bug fix - pretty much any bug fix could be deemed a breaking change - but could be persuaded that it's a break and should go into 8.x. Either way, this has to be fixed. Broadly similar to the fix in https://github.com/ReactiveX/rxjs/pull/6151.

I have this on my of things to do/fix, but you - or anyone else - are welcome to do it. I also have noted that `elementAt`, `find` and `takeWhile` have - or might have? - a similar issue.

benlesh:
Hmm... I just closed that issue as "not a bug". It seems more like a misunderstanding in the order of execution. I'll reopen because I'm interested in other opinions.

backbone87:
while checking out take, i noticed that it also doesn't handle non-ints very well either.

imho, it should either check for `Number.MAX_SAFE_INTEGER` and floor the count or try to deal with all numbers (the latter can be problematic because of incrementing doesn't work past `Number.MAX_SAFE_INTEGER`

benlesh:
@backbone87 ... probably a separate issue to address.

I do have a fix for take, though. Let's add a `skip` to this test and get it merged, and I'll have a look.

spec/operators/take-spec.ts
```diff
@@ -217,4 +217,21 @@ describe('take', () => {
       expectSubscriptions(e1.subscriptions).toBe(e1subs);
     });
   });
+
+  it('should unsubscribe from the source when it reaches the limit before a recursive synchronous upstream error is caught', () => {
```
benlesh, line 221:
```suggestion
  it.skip('should unsubscribe from the source when it reaches the limit before a recursive synchronous upstream error is caught', () => {
```

backbone87:
ah we pushed at the same time

backbone87:
feel free to edit around the PR & branch, I wont push anymore now to avoid any more conflicts since @benlesh is working on it

benlesh:
@cartant yeah, we're already handling reentrancy in the `next` case, just not in the `error` and `complete` cases. This fix is simple enough.

benlesh:
Thanks for this, @backbone87, very helpful!

benlesh:
Follow up PR is here: https://github.com/ReactiveX/rxjs/pull/6396

PR #6396, opened 2021-05-10
Fix/5487 reentrancy in take, takeWhile, and first

### LINKED ISSUE #5487: first/tap operators should unsubscribe before emit next value

benlesh:
- fix(first/take): behave properly with reentrant errors and completions
    
    - Resolves an issue with both `first` and `take` where an error emitted by a reentrant source at the moment the last value it taken would result in superceding the expected emitted value and completion.
    - Resolves a similar issue where a reentrant completion would supercede the expected last value and completion

    Fixes #5487

- fix(takeWhile): reentrant errors and completions behave properly
    
    - Resolves an issue where a reentrant error notification would short circuit the completion.
    - Adds additional tests.

    Related #5487

- chore: Add reentrant error test for takeUntil

### DIFF
--- a/spec/operators/take-spec.ts
+++ b/spec/operators/take-spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
-import { merge, Observable, of, Subject } from 'rxjs';
-import { mergeMap, take, tap } from 'rxjs/operators';
+import { config, merge, Observable, of, Subject } from 'rxjs';
+import { finalize, mergeMap, take, tap } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
 
@@ -250,4 +250,30 @@ describe('take', () => {
       expectSubscriptions(e1.subscriptions).toBe(e1subs);
     });
   });
+
+  it('should properly handle reentrancy when taking many values', () => {
+    const results: any[] = [];
+    const subject = new Subject<number>();
+
+    config.onStoppedNotification = (x) => results.push(x);
+
+    merge(subject, of(1))
+      .pipe(
+        finalize(() => results.push('finalize source')),
+        take(5),
+        tap((n) => subject.next(n + 1)),
+        finalize(() => results.push('finalize result'))
+      )
+      .subscribe({
+        next: (n) => results.push(n),
+        complete: () => results.push('done'),
+      });
+
+    // Because this is reentrant, the last "taken" value of 5 is followed by `complete()`
+    // which shuts off all value emissions. Otherwise, this would be
+    // [5, 'done', 4, 3, 2, 1, 'finalize source', 'finalize result']
+    expect(results).to.deep.equal([5, 'done', 'finalize source', 'finalize result']);
+  });
+
+  afterEach(() => (config.onStoppedNotification = null));
 });
--- a/spec/operators/take-spec.ts
+++ b/spec/operators/take-spec.ts
@@ -217,9 +217,7 @@ describe('take', () => {
     });
   });
 
-  // This is related to a PR with discussion here: https://github.com/ReactiveX/rxjs/pull/6396
-  // We can't fix this until version 8.
-  it.skip('should unsubscribe from the source when it reaches the limit before a recursive synchronous upstream error is notified', () => {
+  it('should unsubscribe from the source when it reaches the limit before a recursive synchronous upstream error is notified', () => {
     testScheduler.run(({ cold, expectObservable, expectSubscriptions }) => {
       const subject = new Subject();
       const e1 = cold(' (a|)');
@@ -229,6 +227,23 @@ describe('take', () => {
       const result = merge(e1, subject).pipe(
         take(1),
         tap(() => subject.error('error'))
+      );
+
+      expectObservable(result).toBe(expected);
+      expectSubscriptions(e1.subscriptions).toBe(e1subs);
+    });
+  });
+
+  it('should unsubscribe from the source when it reaches the limit before a recursive synchronous upstream completion is notified', () => {
+    testScheduler.run(({ cold, expectObservable, expectSubscriptions }) => {
+      const subject = new Subject();
+      const e1 = cold(' (a|)');
+      const e1subs = '  (^!)';
+      const expected = '(a|)';
+
+      const result = merge(e1, subject).pipe(
+        take(1),
+        tap(() => subject.complete())
       );
 
       expectObservable(result).toBe(expected);
--- a/spec/operators/takeWhile-spec.ts
+++ b/spec/operators/takeWhile-spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
 import { takeWhile, tap, mergeMap } from 'rxjs/operators';
-import { of, Observable, from } from 'rxjs';
+import { of, Observable, from, Subject, merge } from 'rxjs';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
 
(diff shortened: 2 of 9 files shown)


benlesh:
Note: also fixes `elementAt` (because it uses `take`).

src/internal/operators/take.ts
```diff
@@ -51,20 +51,41 @@ export function take<T>(count: number): MonoTypeOperatorFunction<T> {
...
               }
+            },
+            (err) => {
+              // If seen === count or higher, then we've already taken all of
+              // the values we were supposed to, and the error we're getting here
+              // is from reentrant code racing our `complete` above. We want to stop
+              // that here.
+              if (seen < count) {
+                subscriber.error(err);
+              }
```
backbone87, line 86: out of interest: is there any particular reason, why blocking notifications over unsubscribing before the last next notification was used?

(besides the failing delayWhen test)

src/internal/operators/takeWhile.ts
```diff
@@ -57,12 +57,34 @@ export function takeWhile<T>(predicate: (value: T, index: number) => boolean, in
...
+            // we found the "last" value, then we send the value along.
+            subscriber.next(value);
+          }
+          if (!taking) {
+            // We're no longer taking values after calling the predicate above
+            // time to complete the result.
+            subscriber.complete();
+          }
+        },
+        undefined,
```
backbone87, line 78: we should probably also block complete notifications here, as was done in take

spec/operators/takeWhile-spec.ts
```diff
@@ -364,4 +364,88 @@ describe('takeWhile', () => {
 
     expect(sideEffects).to.deep.equal([0, 1, 2]);
   });
+
+  it('should not emit errors sent from the source *after* it found the first value in reentrant scenarios', () => {
```
backbone87, line 368: if i revert the change in takeWhile, this test case isn't failing. to make it fail, i had to use:
```ts
        tap({
          next: (x) => {
            if (x === 'b') {
              subject.error(new Error('reentrant shennanigans'));
            }
          },
        })
```

Enuvid (commented):
@cartant @benlesh @backbone87

Also i am not sure that fix of this problem right now will not affect already written code which used this behaviour. Seems that is a Breaking Change.

I mean do 'unsubscribe source before emit last value next'.

benlesh:
Honestly, I'm inclined to say at some point we can't accommodate every crazy synchronous thing that someone wants to do with the library. It seems like we're going to have to introduce a good amount of complexity to the library in order to accommodate what is honestly a corner case, IMO. At a certain point, the simpler "fix" for this would be to give up and schedule everything.

Originally when I looked at the issue I closed it because I thought "well, this is expected behavior, honestly, it's just how things work". But then I saw we can code to protect against reentrancy in the `next` handler for `take`, and I reconsidered.

Fundamentally to really "solve" this in a way that satifisfies the requirements laid out here, we're going to need to expose some sort of "preteardown" or "stopping" method or flag on the `subscriber` to the world. I doubt the payoff is worth the confusion, but I'm willing to be convinced otherwise.

benlesh:
```
finalize 1
1
done
finalize 2
```

It is completely unexpected. `1` and `done` should happen _before_ `finalize 1`.  The source emits 1, then completes, _then_ is `finalized`.

benlesh:
Yeah... hmm..  in RxJS 6 and 7 (not sure about others), this is all kinds of weird/wrong:

```ts
let n = 0;

const subject = new Subject<number>();

merge(subject, of(n++))
  .pipe(
    finalize(() => {
      console.log('finalized');
    }),
    take(10),
    tap(() => subject.next(n++))
  )
  .subscribe({
    next: console.log,
    complete: () => console.log('done')
  });
  
// outputs 
// 9
// done
// finalized
```

benlesh:
Okay... after some thought, and discussion with @cartant, I think the correct thing to do is to have it `unsubscribe` as soon as it knows it's done.

The principle behind this is simply that:

1. All operators create observables that are the consumers of a source.
2. The consumer should unsubscribe as soon as it knows it no longer wants any values from the source.

That said... this is a **BREAKING CHANGE** and we cannot have it until version 8.

spec/operators/take-spec.ts
```diff
@@ -234,4 +234,47 @@ describe('take', () => {
...
+      )
+      .subscribe({
+        next: (n) => results.push(n),
+        complete: () => results.push('done'),
+      });
+
+    // Because this is reentrant, the last "taken" value of 5 is followed by `complete()`
+    // which shuts off all value emissions. Otherwise, this would be
+    // [5, 'done', 4, 3, 2, 1, 'finalize source', 'finalize result']
+    expect(results).to.deep.equal([5, 'done', 'finalize source', 'finalize result']);
```
backbone87, line 276: this one is weird. is this behavior we would be ok with? guess the solution would be to `subject.pipe(observeOn(asapScheduler))`

spec/operators/takeWhile-spec.ts
```diff
@@ -364,4 +364,130 @@ describe('takeWhile', () => {
...
+          },
+        })
+      );
+
+      expectObservable(result).toBe(expected);
+      expectSubscriptions(source.subscriptions).toBe(subs);
+    });
+  });
+
+  it('should not emit completions sent from the source *after* it found the first value in reentrant scenarios', () => {
```
backbone87, line 473: does not fail on master either, but the same question as above: how to even test this?

spec/operators/takeWhile-spec.ts
```diff
@@ -364,4 +364,130 @@ describe('takeWhile', () => {
...
+          }
+        })
+      );
+
+      expectObservable(result).toBe(expected);
+      expectSubscriptions(source.subscriptions).toBe(subs);
+    });
+  });
+
+  it('should not emit errors sent from the source *after* it found the first value in reentrant scenarios', () => {
```
backbone87, line 431: does not fail on master, but idk how to even test this one

spec/operators/takeWhile-spec.ts
```diff
@@ -364,4 +364,130 @@ describe('takeWhile', () => {
...
+          },
+        })
+      );
+
+      expectObservable(result).toBe(expected);
+      expectSubscriptions(source.subscriptions).toBe(subs);
+    });
+  });
+
+  it('should not emit completions sent from the source *after* it found the first value in reentrant scenarios', () => {
```
backbone87, line 452: does not fail on master, had to use
```ts
        tap((x) => {
          if (x === 'b') {
            subject.complete();
          }
        })
```

src/internal/operators/take.ts
benlesh: Sorry for the late response. I'm not sure I understand your question here.

src/internal/operators/takeWhile.ts
benlesh: This has been added.

spec/operators/takeWhile-spec.ts
benlesh: Seems this was fixed by a later commit (sorry, revisiting this)

spec/operators/takeWhile-spec.ts
benlesh: This would not be testing the same thing, exactly.

Ultimately, this and the test below it are really just exercising the subscriber guarantees. But nonetheless I think we should keep them so someone doesn't implement something in the future that breaks this contract.

jakovljevic-mladen (approved):
LGTM

PR #6401, opened 2021-05-12
feat(Subscription): idempotent add and remove of teardowns

### LINKED ISSUE #6400: Switch to using `Set` for Subscription teardowns
Related to this issue here: #5638

1. Using an array of teardowns in our Subscriptions isn't ideal, performance-wise, for removing subscriptions.
2. Using a Set, rather than an Array would more closely mimic the behavior of most other common event registries, like `EventTarget`. Consider the following:

```ts
function handler() {
   console.log('handled');
}

const target = new EventTarget();
target.addEventListener('unsubscribe', handler);
target.addEventListener('unsubscribe', handler);
target.addEventListener('unsubscribe', handler);

target.dispatchEvent(new Event('unsubscribe')); // Only fires `handler` once!

target.removeEventListener('unsubscribe', handler);

target.dispatchEvent(new Event('unsubscribe')); // handler not fired (as it was removed)
```

This is in drastic contrast to our Subscription implementation, where the same subscription may be registered `n` times (which doesn't make sense, because it can only be torn down once), and the same function can also be registered `n` times, where it will execute `n` times. 

Now, switching to using a `Set` internally would be a breaking change, as it would change how people can register and remove functions, more than anything. It's doubtful that it would be a breaking change for Subscriptions, because unsubscribe is idempotent. 

Therefore we will not be able to do this until version 8.x.

benlesh:
- Adding the same instance of a teardown more than once is the same as adding it once
- Removing a teardown with `Subscription.prototype.remove` will remove it much faster

BREAKING CHANGE: Adding the same function instance to a subscription as a teardown multiple times will now result in that function being executed only once on teardown. This brings us inline with the behavior of EventTarget, and also makes removing teardowns faster. The workaround is to make sure you are adding a new function instance to the Subscription each time if you need the same effect.

Resolves #6400

### DIFF
--- a/src/internal/Subscription.ts
+++ b/src/internal/Subscription.ts
@@ -34,7 +34,7 @@ export class Subscription implements SubscriptionLike {
    * The list of registered teardowns to execute upon unsubscription. Adding and removing from this
    * list occurs in the {@link #add} and {@link #remove} methods.
    */
-  private _teardowns: Exclude<TeardownLogic, void>[] | null = null;
+  private _teardowns: Set<Exclude<TeardownLogic, void>> | null = null;
 
   /**
    * @param initialTeardown A function executed first as part of the teardown
@@ -134,7 +134,7 @@ export class Subscription implements SubscriptionLike {
           }
           teardown._addParent(this);
         }
-        (this._teardowns = this._teardowns ?? []).push(teardown);
+        (this._teardowns = this._teardowns ?? new Set()).add(teardown);
       }
     }
   }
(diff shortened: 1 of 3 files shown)


cartant (approved):

src/internal/Subscription.ts
```diff
@@ -190,7 +190,9 @@ export class Subscription implements SubscriptionLike {
    */
   remove(teardown: Exclude<TeardownLogic, void>): void {
     const { _teardowns } = this;
-    _teardowns && arrRemove(_teardowns, teardown);
+    if (_teardowns) {
+      _teardowns.delete(teardown);
+    }
```
backbone87, line 195: very nit-pick:
```ts
const { _teardowns } = this;
if (_teardowns) {
  _teardowns.delete(teardown);
}
// can be replaced with
this._teardowns?.delete(teardown);
```

benlesh:
Changed the base to `8.x` branch, since that's where this should be going.

PR #6403, opened 2021-05-13
Docs: multicast deprecation

backbone87:
**Description:**
This adds deprecated tags to the primary signatures of deprecated multicast, publish & publishReplay, so they are properly marked as such in the [API reference](https://rxjs.dev/api).


cartant (approved):

cartant:
Thanks.

PR #6405, opened 2021-05-15
feat(Subject): add isObserved() api

bbarry:
**Description:**

This PR adds an api `subject.isObserved()` to handle the common use case of the deprecated `subject.observers` member 

* https://stackoverflow.com/questions/44555615/rxjs-detect-when-observable-has-been-subscribed-to
* https://stackoverflow.com/questions/33441393/is-there-a-way-to-check-for-output-wire-up-from-within-a-component-in-angular-2

Relates to: #5967

### DIFF
--- a/spec/Subject-spec.ts
+++ b/spec/Subject-spec.ts
@@ -430,29 +430,28 @@ describe('Subject', () => {
     done();
   });
 
-  it('should expose observed status', (done) => {
+  it('should expose observed status', () => {
     const subject = new Subject();
 
-    expect(subject.isObeserved()).to.equal(false);
+    expect(subject.observed()).to.equal(false);
 
     const sub1 = subject.subscribe(function (x) {
       //noop
     });
 
-    expect(subject.isObeserved()).to.equal(true);
+    expect(subject.observed()).to.equal(true);
 
     const sub2 = subject.subscribe(function (x) {
       //noop
     });
 
-    expect(subject.isObeserved()).to.equal(true);
+    expect(subject.observed()).to.equal(true);
     sub1.unsubscribe();
-    expect(subject.isObeserved()).to.equal(true);
+    expect(subject.observed()).to.equal(true);
     sub2.unsubscribe();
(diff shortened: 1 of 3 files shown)


src/internal/Subject.ts
```diff
@@ -91,6 +91,10 @@ export class Subject<T> extends Observable<T> implements SubscriptionLike {
     this.observers = null!;
   }
 
+  isObeserved() {
+    return this.observers?.length > 0;
+  }
+
```
benlesh, line 97: given we have a property `closed`, I think it's more inline with other parts of RxJS to not prefix with `is` and to use a property. Also `Obeserved` is a typo. :)

```ts
get observed() {
  return 0 < this.observers?.length;
}
```

benlesh (changes requested):

benlesh:
I'm generally in favor of this addition. However, it requires core team agreement.

spec/Subject-spec.ts
```diff
@@ -430,6 +430,31 @@ describe('Subject', () => {
...
+    });
+
+    expect(subject.isObeserved()).to.equal(true);
+    sub1.unsubscribe();
+    expect(subject.isObeserved()).to.equal(true);
+    sub2.unsubscribe();
+    expect(subject.isObeserved()).to.equal(false);
+    subject.unsubscribe();
+    expect(subject.isObeserved()).to.equal(false);
+    done();
```
backbone87, line 455: nit: I think the `done()` is not needed since this is a fully synchronous test?

bbarry: Fair, the one above it that I copied has it but others don't.  I've removed it.

src/internal/Subject.ts
bbarry: fixed

benlesh:
Notes from core team meeting: Approved, but it should be a readonly property `observed`, not a method.

benlesh (approved):

benlesh:
@bbarry thanks for this important contribution! It's now out in version 7.1.0

PR #6410, opened 2021-05-19
fix: make share, publish, et al. operator calls referentially transparent

cartant:
**Description:**

This PR incorporates the fixes and tests in #5585 and also fixes and tests the `share`, `shareReplay`, `publishBehavior` and `publishLast` operators.

The problem that is fixed can be described as follows:

- Calls to **operators** - like `share` - _must be referentially transparent_. Otherwise, those operators cannot be used in calls to the static `pipe` function (to create partial pipelines).
- It's the calls to the **operator functions** - the functions that are returned from **operator** calls - that _do not have to be referentially transparent_ (and can therefore implement the publishing and sharing). Those calls are made when the complete pipeline is composed via a call to a source observable's `pipe` method.

IMO, the changes are a bug fix, rather than a breaking change. The behaviour is changed. However, the behaviour prior to this change was not sensible and I cannot see how anyone could possibly be relying upon it. Any callers using a partial pipeline composed using the static `pipe` function and one of the affected operators would see wildly different behaviour - subscriptions to completely unexpected sources - depending upon the order in which subscriptions were made.

**Related issue (if exists):** #5411 #5585

### DIFF
--- a/src/internal/operators/share.ts
+++ b/src/internal/operators/share.ts
@@ -134,6 +134,7 @@ export function share<T>(options: ShareConfig<T>): MonoTypeOperatorFunction<T>;
  * @return A function that returns an Observable that mirrors the source.
  */
 export function share<T>(options: ShareConfig<T> = {}): MonoTypeOperatorFunction<T> {
+  const { connector = () => new Subject<T>(), resetOnError = true, resetOnComplete = true, resetOnRefCountZero = true } = options;
   // It's necessary to use a wrapper here, as the _operator_ must be
   // referentially transparent. Otherwise, it cannot be used in calls to the
   // static `pipe` function - to create a reusable pipeline.
@@ -143,8 +144,6 @@ export function share<T>(options: ShareConfig<T> = {}): MonoTypeOperatorFunction
   // _operator function_ is called when the complete pipeline is composed - not
   // when the static `pipe` function is called.
   return (wrapperSource) => {
-    const { connector = () => new Subject<T>(), resetOnError = true, resetOnComplete = true, resetOnRefCountZero = true } = options;
-
     let connection: SafeSubscriber<T> | null = null;
     let resetConnection: Subscription | null = null;
     let subject: SubjectLike<T> | null = null;
--- a/src/internal/operators/share.ts
+++ b/src/internal/operators/share.ts
@@ -134,90 +134,92 @@ export function share<T>(options: ShareConfig<T>): MonoTypeOperatorFunction<T>;
  * @return A function that returns an Observable that mirrors the source.
  */
 export function share<T>(options: ShareConfig<T> = {}): MonoTypeOperatorFunction<T> {
-  const { connector = () => new Subject<T>(), resetOnError = true, resetOnComplete = true, resetOnRefCountZero = true } = options;
+  return (observable) => {
+    const { connector = () => new Subject<T>(), resetOnError = true, resetOnComplete = true, resetOnRefCountZero = true } = options;
 
-  let connection: SafeSubscriber<T> | null = null;
-  let resetConnection: Subscription | null = null;
-  let subject: SubjectLike<T> | null = null;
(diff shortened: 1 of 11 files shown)


benlesh:
I'm generally in favor of this change. It seems like a bug that may have been introduced at some point. Can we prove that out? Is there a version of RxJS 5 or 6 where these new tests would pass?

cartant:
> It seems like a bug that may have been introduced at some point

Looking at the operators in HEAD of the `5.x` branch:

- [`share`](https://github.com/ReactiveX/rxjs/blob/5.x/src/operators/share.ts) **is** referentially transparent
- [`publish`](https://github.com/ReactiveX/rxjs/blob/5.x/src/operators/publish.ts) **is not** referentially transparent
- [`publishBehavior`](https://github.com/ReactiveX/rxjs/blob/5.x/src/operators/publishBehavior.ts) **is** referentially transparent
- [`publishLast`](https://github.com/ReactiveX/rxjs/blob/5.x/src/operators/publishLast.ts) **is** referentially transparent
- [`publishReplay`](https://github.com/ReactiveX/rxjs/blob/5.x/src/operators/publishReplay.ts) **is not** referentially transparent

src/internal/operators/share.ts
```diff
@@ -84,88 +101,151 @@ export function share<T>(options: ShareConfig<T>): MonoTypeOperatorFunction<T>;
...
+  // It's necessary to use a wrapper here, as the _operator_ must be
+  // referentially transparent. Otherwise, it cannot be used in calls to the
+  // static `pipe` function - to create a reusable pipeline.
+  //
+  // The _operator function_ - the function returned by the _operator_ - will
+  // not be referentially transparent - as it shares its source - but the
+  // _operator function_ is called when the complete pipeline is composed - not
+  // when the static `pipe` function is called.
+  return (wrapperSource) => {
+    const { connector = () => new Subject<T>(), resetOnError = true, resetOnComplete = true, resetOnRefCountZero = true } = options;
```
backbone87, line 146: the destructering of options can and probably should happen when the operator is invoked and not later when the operator function is. besides saving some "duplicated" work, its main purpose would be defend against modification of the options object and therefore "mutating" the operator function

src/internal/operators/share.ts
```diff
@@ -84,88 +101,151 @@ export function share<T>(options: ShareConfig<T>): MonoTypeOperatorFunction<T>;
...
+          complete: () => {
+            hasCompleted = true;
+            cancelReset();
+            resetConnection = handleReset(reset, resetOnComplete);
+            dest.complete();
+          },
+        });
+        from(source).subscribe(connection);
+      }
+    })(wrapperSource);
```
backbone87, line 229: is it worth introducing a helper function to make this more clear?

something like:
```ts
export function share<T>(options: ShareConfig<T> = {}): MonoTypeOperatorFunction<T> {
  // init operator

  return prepareAndOperate(() => {
    // init pipeline

    return (source, subscriber) => {
      // init subscription
    };
  });
}
```

cartant: I don't think that a helper function is warranted, as four of these operators are going to be removed in version 8. After that, AFAICT, `share` will be the only operator that returns an operator function that's not referentially transperant.

src/internal/operators/share.ts
cartant: Done. Thanks for spotting that.

benlesh:
So this is a regression in `publishReplay` introduced somewhere, and arguably a bug that has existed in `publish` since 5.x.

benlesh:
I guess the workaround for this if you really WANTED this behavior in, say, `publish` would be:

```ts
const olderPublish = () => {
  return connect({
    connector: () => subject
  })
}
```

Perhaps we should document that? Or maybe the comment here will be enough. I doubt it will be an issue, TBH.

benlesh (approved):

PR #6411, opened 2021-05-20
docs(operators): single

victororlyk:
mismatch in result

**Description:**

Mistake in docs
https://rxjs.dev/api/operators/single

### DIFF
--- a/src/internal/operators/single.ts
+++ b/src/internal/operators/single.ts
@@ -73,7 +73,7 @@ export function single<T>(predicate?: (value: T, index: number, source: Observab
  *   single(x => x.name.startsWith('B'))
  * )
  * .subscribe(x => console.log(x));
- * // Error emitted: NotFoundError('No values match')
+ * // Emits undefined
  * ```
  *
  * @see {@link first}


cartant (commented):
In version 7, it should throw if the predicate is unmatched:

https://github.com/ReactiveX/rxjs/blob/190e4d00990e43a10a3c4c9d792dbd089aa267ad/src/internal/operators/single.ts#L110-L117

cartant:
Closing this because v7 throws if there is no match:

https://github.com/ReactiveX/rxjs/blob/12c3716cecbf01f353c980488bf18845177b37b6/spec/operators/single-spec.ts#L153-L162

If that doesn't occur, it's a bug. It's not an oversight in the documentation.

I did just notice that the test description is misleading. I'll remedy that in a separate PR.

victororlyk:
Then it would be nice if there would be a version shown somewhere on the site, because this can improve readability really well

cartant:
> Then it would be nice if there would be a version shown somewhere on the site, because this can improve readability really well

Capture

The version is shown in the menu and it's possible to toggle between the docs for v6 and v7.

PR #6418, opened 2021-05-21
docs(operators): sequenceEqual

victororlyk:
probably a mistake, because not 'rxjs' nor 'rxjs/operators' has `innerFrom` exported


cartant (approved):
Yep, that was an error. Thanks.

PR #6419, opened 2021-05-23
docs: "timer" accepts milliseconds, not seconds

lazarljubenovic:
**Description:** A minor mistake in the JSDoc description of the parameter threw me off when I was using a quick lookup of the parameter meaning in my IDE. The example and earlier explanation correctly stated milliseconds, but the parameter description was wrong.

**Related issue (if exists):** Didn't really look, just created a quick edit from the GitHub editor.


cartant (approved):
Thanks for the correction.

PR #6421, opened 2021-05-24
feat(retry): Add configurable delay

benlesh:
- Adds a `delay` configuration that allows the user to create simpler exponential backoff, simple retry delays, and/or other functionality.

Basically, this will allow a user to write an exponential backoff or simple delay more easily and intuitively than `retryWhen` could. In fact, we might want to deprecate `retryWhen` in favor of this, and also add this pattern to `repeat` (if we decide it's a good idea).

**Retry every 5 seconds, a maximum of 100 times:**

```ts
source.pipe(
  retry({
     // max retries (count is an unfortunate existing name)
     count: 100, 
     // retry 5 seconds after any error
     delay: 5000, // retry after 5 seconds
  })
)
```

**Exponential backoff example**:
```ts
source.pipe(
  retry({
    // max 100 retries
    count: 100,
    // backoff starting at 2 seconds, exponentially up to 1 minute.
    // retryCount starts a 1. 
    delay: (_err, retryCount) => timer(Math.min(60000, 1000 * (2 ** retryCount)))
  })
)
```

### Notes

Honestly, I'm okay even with adding configuration for the above backoff scenario, given it's such a common thing I've had to implement. Even if all we did was provide a helper function that was easy to tree shake like:  `retry({ delay: exponentialBackoff({ startDelay: 1000, maxDelay: 60000 }) })` or the like.

### DIFF
--- a/src/internal/operators/retry.ts
+++ b/src/internal/operators/retry.ts
@@ -1,11 +1,28 @@
-import { MonoTypeOperatorFunction } from '../types';
+import { MonoTypeOperatorFunction, ObservableInput } from '../types';
 import { operate } from '../util/lift';
 import { Subscription } from '../Subscription';
 import { OperatorSubscriber } from './OperatorSubscriber';
 import { identity } from '../util/identity';
+import { timer } from '../observable/timer';
+import { innerFrom } from '../observable/from';
 
 export interface RetryConfig {
-  count: number;
+  /**
+   * The maximum number of times to retry.
+   */
+  count?: number;
+  /**
+   * The number of milliseconds to delay before retrying, OR a function to
+   * return a notifier for delaying. If a function is returned, that function should
+   * return a notifier that, when it emits will retry the source. If the notifier
+   * completes _without_ emitting, the resulting observable will complete without error,
+   * if the notifier errors, the error will be pushed to the result.
+   */
+  delay?: number | ((error: any, retryCount: number) => ObservableInput<any>);
+  /**
+   * Whether or not to reset the retry counter on success.
+   * Defaults to false.
+   */
   resetOnSuccess?: boolean;
 }
 
@@ -56,6 +73,14 @@ export interface RetryConfig {
  * source stream when the source stream errors, at most `count` times.
  */
 export function retry<T>(count?: number): MonoTypeOperatorFunction<T>;
+
+/**
+ * A more configurable means of retrying a source.
+ *
+ * If `delay` is provided as a `number`, after the source errors, the result will wait `delay` milliseconds,
+ * then retry the source. If `delay` is a function, th
+ * @param config The retry configuration
+ */
 export function retry<T>(config: RetryConfig): MonoTypeOperatorFunction<T>;
 export function retry<T>(configOrCount: number | RetryConfig = Infinity): MonoTypeOperatorFunction<T> {
   let config: RetryConfig;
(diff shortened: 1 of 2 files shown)


src/internal/operators/retry.ts
```diff
@@ -56,6 +73,14 @@ export interface RetryConfig {
  * source stream when the source stream errors, at most `count` times.
  */
 export function retry<T>(count?: number): MonoTypeOperatorFunction<T>;
+
+/**
+ * A more configurable means of retrying a source.
+ *
+ * If `delay` is provided as a `number`, after the source errors, the result will wait `delay` milliseconds,
+ * then retry the source. If `delay` is a function, th
```
jakovljevic-mladen, line 81: Seems like this sentence is not finished.

backbone87:
isnt `retryWhen` the more general operator which can be improved by providing premade notifier factories?

```ts
// like `retry`
retryWhen(count(n));
retryWhen(count({ count: n, resetOnSuccess: true }));

// exponential backoff
retryWhen(exponentialBackoff({ startDelay: d, maxDelay: m }));

// custom, like current `retryWhen`
retryWhen(createMyCustomNotifier);
```

cartant:
> isnt `retryWhen` the more general operator which can be improved by providing premade notifier factories?

I somewhat agree with this, but AFAICT the following would not be possible with `retryWhen` as it has no mechanism for conveying 'success' to the notifier:

```ts
retryWhen(count({ count: n, resetOnSuccess: true }));
```

I'm more curious as to whether this PR hints at a general approach for reducing the number of core operators by making them more 'configurable' - as was done with `share` and the multicasting API.

backbone87:
> general approach for reducing the number of core operators

i think the current state of `share` & `shareReplay` are a good example. a "highly configurable" base operator with the most common configuration as its default and if necessary additional operators for other very common configurations that are implemented in terms of the "base" operator.
imho the benefits of this approach is less code duplication, less tests required and most importantly no "slightly different behavior under equal configuration" for otherwise very similar operators.
from the idea to provide "option factories" (`exponentialBackoff`): should it then also be `share(replay(n))`?

src/internal/operators/retry.ts
benlesh: This is so weird. I feel like this is the third or forth time this has happened to me, but only with doc comments :\

benlesh:
Notes from core team: General approval.

cartant (commented):
I think this is fine, but I'm not sure about the `resetOnSuccess` name:

```ts
  /**
   * Whether or not to reset the retry counter on success.
   * Defaults to false.
   */
  resetOnSuccess?: boolean;
```

What is success? The accompanying doc/comment doesn't make this clear. The other `reset` options in the API are, IMO, a little clearer: `resetOnError`, `resetOnComplete`, `resetOnDisconnect`, `resetOnRefCountZero`.

I mean, from `resetOnSuccess` a caller could infer that it could only mean that it'll reset when a `next` notification is received - because resetting on complete would be meaningless - but should the caller really need to even think about this?

Maybe `resetOnNext`?

benlesh:
@cartant I've changed the one setting to `resetOnFirstValue`. Since `resetOnSuccess` doesn't mean much, and `resetOnNext` is ambiguous.

benlesh:
@cartant Just realized that `resetOnSuccess` was pre-existing. Probably should change that back.

cartant (approved):

PR #6425, opened 2021-05-25
feat(groupBy): add seed option to eagerly create groups

backbone87:
- [ ] It must have a `-spec.ts` tests file covering the canonical corner cases, with marble diagram tests
- [ ] The spec file should have a type definition test at the end of the spec to verify type definition for various use cases
- [ ] The operator should also be documented. See [Documentation Guidelines](../CONTRIBUTING.md).

**Description:**

adds a `seed` option to the `groupBy` operator which allows opening groups eagerly.

if the approach is agreed upon I will provide more tests.

https://github.com/ReactiveX/rxjs/pull/6388#issuecomment-846192168


cartant:
I would not be in favour of this, especially as `seed` is `() => ObservableInput<K>`. That doesn't fit with this description of the feature:

> opening groups eagerly

I'm also unconvinced there is a compelling use case for this.

backbone87:
the use case is what buckets/splitBy would have enabled:
```ts
number$.pipe(
  groupBy(isEven, { seed: booleans }),
  toArray(),
  mergeMap(([evenNumber$, oddNumber$]) =>
    merge(
      evenNumber$.pipe(mapTo('I am even')),
      oddNumber$.pipe(mapTo('I am odd')),
    )
  ),
);

benlesh:
I guess the important bit that @cartant is bringing up is the use cases this fulfills. One of the issues we _already have_ with this library is we have a lot of APIs that simply don't have a ton of use cases. `groupBy` borders on that already depending on who you ask. 

The thing to keep in mind is: Every new feature we introduce we have to maintain for _years_. So we need to be sure there's value to it.

backbone87:
> Every new feature we introduce we have to MAINTAIN FOR YEARS. So we need to be sure there's value to it.

**Absolutely** no problem with this. I just want the proposal to be given fair consideration and a bit of discussion. If the conclusion is to not include the PR, so it be.

> `groupBy` borders on that already depending on who you ask.

As probably one of the few who use rxjs not in the frontend, but in the backend for stream processing, I can tell that `groupBy` is a main workhorse for me. Sure that is anecdotal, but still.

spoilerdo:
any news on this?? I need to use it

PR #6431, opened 2021-05-28
build(package): bump up typescript@4.3

kwonoj:
**Description:**

Looks like tsc gets bit more strict for tuple types. Those 2 are marked as deprecated but instead of trying to change public surfaces bene internals to workaround.

**BREAKING CHANGE:** 


kwonoj:
CI fails with api_gurdian probably due to 4.3 changes

``` 
+export declare const observable: string | typeof Symbol.observable;
```

This seems bit.. concerning? Curious if we'll hit some regression by this.

cartant:
I don't think we should bump TS without some changes to CI to ensure that at least the tests (the dtslint tests especially) are run using our minimum-supported version of 4.2. In fact, given that that's our minimum, I'm not sure we should bump at all. IIRC, we did not do this for v6.

The code fixes in this LGTM. I'd prefer them to be separated from the bumping.

Also, I think we should always use a tildle version range with TS as it does not follow semver and minors can be breaking.

kwonoj:
ah that's good point.

src/internal/operators/zip.ts
```diff
@@ -21,6 +21,6 @@ export function zip<T, A extends readonly unknown[], R>(
  */
 export function zip<T, R>(...sources: Array<ObservableInput<any> | ((...values: Array<any>) => R)>): OperatorFunction<T, any> {
   return operate((source, subscriber) => {
-    zipStatic(source, ...sources).subscribe(subscriber);
+    zipStatic(source as ObservableInput<any>, ...(sources as Array<ObservableInput<any>>)).subscribe(subscriber);
```
benlesh, line 24:
```suggestion
    // Casting here as zipStatic accepts the result selector as well.
    zipStatic(source, ...(sources as any[])).subscribe(subscriber);
```

src/internal/operators/combineLatest.ts
```diff
@@ -27,7 +27,7 @@ export function combineLatest<T, A extends readonly unknown[], R>(...sources: [.
 export function combineLatest<T, R>(...args: (ObservableInput<any> | ((...values: any[]) => R))[]): OperatorFunction<T, unknown> {
   const resultSelector = popResultSelector(args);
   return resultSelector
-    ? pipe(combineLatest(...args), mapOneOrManyArgs(resultSelector))
+    ? pipe(combineLatest(...(args as Array<ObservableInput<any>>)), mapOneOrManyArgs(resultSelector))
```
benlesh, line 30:
```suggestion
    ? pipe(combineLatest(...(args as ObservableInput<any>[])), mapOneOrManyArgs(resultSelector))
```

package.json
```diff
@@ -192,7 +192,7 @@
     "tslint-no-toplevel-property-access": "0.0.2",
     "tslint-no-unused-expression-chai": "0.0.3",
     "typedoc": "^0.17.8",
-    "typescript": "~4.2.2",
+    "typescript": "^4.3.2",
```
benlesh, line 195: Since 4.2 is our minimum supported version, I don't think we can/should update this. While it's certainly important to update it while fixing this issue, and I'm glad CI caught it, I wouldn't want to miss a regression in 4.2 we could cause later.

benlesh (changes requested):

package.json
benlesh: Also, I suspect the api_guardian issues will go away if you revert this.

benlesh, line 195:
```suggestion
    "typescript": "~4.2.2",
```

benlesh:
@cartant @kwonoj I've reverted the change to the package.json, so we're only using 4.2 for the project. I think it's fine to temporarily, locally update, but we need to support 4.2 for the duration of version 7. 

We can discuss updating and leveraging tests to ensure 4.2 support another time. Right now we just need to fix CI.

benlesh (approved):

PR #6434, opened 2021-05-28
Core semantics

benlesh:
This is a starting place to put together a document that outlines the semantics of functionality provided by the core library. This will be used as a reference for the "correct" behaviors of the library starting in version 8. We probably can't do that in version 7 without breaking _someone_, but it's definitely something we want to get defined, as it keeps coming.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,23 +5,24 @@ it still applies to you. Ignorance is not an exemption.
 
 Contents
 
-- [Submitting a Pull Request (PR)](#submitting-a-pull-request-pr)
-  - [After your pull request is merged](#after-your-pull-request-is-merged)
-- [Coding Style Guidelines](#coding-style-guidelines)
-- [Documentation](#documentation)
-- [Unit Tests](#unit-tests)
-  - [CI Tests](#ci-tests)
-- [Performance Tests](#performance-tests)
-  - [Macro](#macro)
-  - [Micro](#micro)
-- [Commit Message Guidelines](#commit-message-guidelines)
-  - [Commit Message Format](#commit-message-format)
-  - [Revert](#revert)
-  - [Type](#type)
-  - [Scope](#scope)
-  - [Subject](#subject)
-  - [Body](#body)
-  - [Footer](#footer)
+- [Contributing to RxJS](#contributing-to-rxjs)
+  - [Submitting a Pull Request (PR)](#submitting-a-pull-request-pr)
+    - [After your pull request is merged](#after-your-pull-request-is-merged)
+  - [Coding Style Guidelines](#coding-style-guidelines)
+  - [Documentation](#documentation)
+  - [Unit Tests](#unit-tests)
+    - [CI Tests](#ci-tests)
+  - [Performance Tests](#performance-tests)
+    - [Macro](#macro)
+    - [Micro](#micro)
+  - [Commit Message Guidelines](#commit-message-guidelines)
+    - [Commit Message Format](#commit-message-format)
+    - [Revert](#revert)
+    - [Type](#type)
+    - [Scope](#scope)
+    - [Subject](#subject)
+    - [Body](#body)
+    - [Footer](#footer)
 
 ---
 
@@ -47,8 +48,6 @@ Before you submit your Pull Request (PR) consider the following guidelines:
 
 - Create your patch, following [code style guidelines](#coding-style-guidelines), and **including appropriate test cases**.
 - Run the full test suite and ensure that all tests pass.
-- Run the micro and macro performance tests against your feature branch and compare against master
-  to ensure performance wasn't changed for the worse.
 - Commit your changes using a descriptive commit message that follows our
   [commit message guidelines](#commit-message-guidelines). Adherence to these conventions
   is necessary because release notes are automatically generated from these messages.
--- a/docs_app/content/guide/core-semantics.md
+++ b/docs_app/content/guide/core-semantics.md
@@ -4,22 +4,30 @@ Starting in version 8, all RxJS operators that are provided in the core library 
 
 ## Purpose
 
-The purpose of these semantics is provide predictable behavior for the users of our library, and to ensure consistent behavior between our many different operators.
+The purpose of these semantics is provide predictable behavior for the users of our library, and to ensure consistent behavior between our many different operators. It should be noted that at the time of this writing, we don't always adhere to these semantic guidelines. This document is to serve as a goalpost for upcoming changes and work as much as it is to help describe the library. This is also a "living document" and is subject to change.
+
+## General Design Guidelines
+
+**Functions such as operators, constructors, and creation functions, should use named parameters in cases where there is more than 1 argument, and arguments after the first are non-obvious.** The primary use case should be streamlined to work without configuration. For example, `fakeFlattenMap(n => of(n))` is fine, but `fakeFlattenMap(n => of(n), 1)` is less readable than `fakeFlattenMap(n => of(n), { maxConcurrent: 1 })`. Other things, like `of(1, 2, 3)` are obvious enough that named parameters don't make sense.
 
 ## Operators
 
-    - MUST be a function that returns an [operator function](https://rxjs.dev/api/index/interface/OperatorFunction). That is `(source: Observable<In>) => Observable<Out>`.
-    - The observable returned by the operator function MUST subscribe to the source.
-    - If the operation performed by the operator can tell it not change anything about the output of the source, it MUST return the reference to the source. For example `take(Infinity)` or `skip(0)`.
-    - Operators that accept a "notifier", that is another observable source that is used to trigger some behavior, must accept any type that can be converted to an `Observable` with `from`. For example `takeUntil`.
-    - "Notifiers" that are provided to operators are expected to emit a value in order to trigger notification. Completion does not count as a notification.
-    - "Notifiers" provided directly to the operator MUST be subscribed to *before* the source is subscribed to. "Notifiers" created via factory function provided to the operator SHOULD be subscribed to at the earliest possible moment.
-    - The observable returned by the operator function is considered to be the "consumer" of the source. As such, the consumer MUST unsubscribe from the source as soon as it knows it no longer needs values before proceeding to do _any_ action.
-    - Events that happen after the completion of a source should happen after the source finalizes. This is to ensure that finalization always happens in a predictable time frame relative to the event.
-    - `Error` objects should never be retained longer than necessary. This is a possible source of memory pressure.
-    - `Promise` references should never be retained longer than necessary. This is a possible source of memory pressure.
-    - Operators that split a source `Observable<T>` into many child observables `Observable<Observable<T>>` should emit child observables that do not stop because the original consumer subscription is unsubscribed. This is because those child observables may be consumed outside of that subscription lifecycle. For example, a user could capture a grouped (child) observable emitted from `groupBy` and subscribe to it elsewhere. The purpose of the `groupBy` was to create observables, not dictate their lifespan.
+- MUST be a function that returns an [operator function](https://rxjs.dev/api/index/interface/OperatorFunction). That is `(source: Observable<In>) => Observable<Out>`.
+- The returned operator function MUST be [referentially transparent](https://en.wikipedia.org/wiki/Referential_transparency). That is to say, that if you capture the return value of the operator (e.g. `const double => map(x => x + x)`), you can use that value to operate on any many observables as you like without changing any underlying state in the operator reference. (e.g. `a$.pipe(double)` and `b$.pipe(double)`).
+- The observable returned by the operator function MUST subscribe to the source.
+- If the operation performed by the operator can tell it not change anything about the output of the source, it MUST return the reference to the source. For example `take(Infinity)` or `skip(0)`.
+- Operators that accept a "notifier", that is another observable source that is used to trigger some behavior, must accept any type that can be converted to an `Observable` with `from`. For example `takeUntil`.
+- Operators that accept "notifiers" (as described above), MUST ONLY recognized next values from the notifier as "notifications". Emitted completions may not be used a source of notification.
+- "Notifiers" provided directly to the operator MUST be subscribed to _before_ the source is subscribed to. "Notifiers" created via factory function provided to the operator SHOULD be subscribed to at the earliest possible moment.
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -47,8 +48,6 @@ Before you submit your Pull Request (PR) consider the following guidelines:
 
 - Create your patch, following [code style guidelines](#coding-style-guidelines), and **including appropriate test cases**.
 - Run the full test suite and ensure that all tests pass.
-- Run the micro and macro performance tests against your feature branch and compare against master
```
benlesh, line 50: This was unrelated, but I saw it and it needs to be removed.

CONTRIBUTING.md
```diff
@@ -5,23 +5,24 @@ it still applies to you. Ignorance is not an exemption.
...
-  - [Micro](#micro)
-- [Commit Message Guidelines](#commit-message-guidelines)
-  - [Commit Message Format](#commit-message-format)
-  - [Revert](#revert)
-  - [Type](#type)
-  - [Scope](#scope)
-  - [Subject](#subject)
-  - [Body](#body)
-  - [Footer](#footer)
+- [Contributing to RxJS](#contributing-to-rxjs)
```
benlesh, line 8: This was automatic.

docs_app/content/guide/core-semantics.md
```diff
@@ -0,0 +1,25 @@
...
+    - Events that happen after the completion of a source should happen after the source finalizes. This is to ensure that finalization always happens in a predictable time frame relative to the event.
+    - `Error` objects should never be retained longer than necessary. This is a possible source of memory pressure.
+    - `Promise` references should never be retained longer than necessary. This is a possible source of memory pressure.
+    - Operators that split an source `Observable<T>` into many child observables `Observable<Observable<T>>` should emit child observables that do not stop simply because the original consumer subscription is unsubscribed. This is because those child observables may be consumed outside of that subscription life cycle. For example, a user could capture a grouped (child) observable emitted from `groupBy` and subscribe to it elsewhere. The purpose of the `groupBy` was to create observables, not dictate their lifespan.
```
benlesh, line 21:
```suggestion
    - Operators that split a source `Observable<T>` into many child observables `Observable<Observable<T>>` should emit child observables that do not stop because the original consumer subscription is unsubscribed. This is because those child observables may be consumed outside of that subscription lifecycle. For example, a user could capture a grouped (child) observable emitted from `groupBy` and subscribe to it elsewhere. The purpose of the `groupBy` was to create observables, not dictate their lifespan.
```

docs_app/content/guide/core-semantics.md
```diff
@@ -0,0 +1,25 @@
...
+    - The observable returned by the operator function is considered to be the "consumer" of the source. As such, the consumer MUST unsubscribe from the source as soon as it knows it no longer needs values before proceeding to do _any_ action.
+    - Events that happen after the completion of a source should happen after the source finalizes. This is to ensure that finalization always happens in a predictable time frame relative to the event.
+    - `Error` objects should never be retained longer than necessary. This is a possible source of memory pressure.
+    - `Promise` references should never be retained longer than necessary. This is a possible source of memory pressure.
+    - Operators that split a source `Observable<T>` into many child observables `Observable<Observable<T>>` should emit child observables that do not stop because the original consumer subscription is unsubscribed. This is because those child observables may be consumed outside of that subscription lifecycle. For example, a user could capture a grouped (child) observable emitted from `groupBy` and subscribe to it elsewhere. The purpose of the `groupBy` was to create observables, not dictate their lifespan.
```
backbone87, line 21: it would be helpful to explain when to ultimately unsubscribe from the source and how to "detect" this case.

docs_app/content/guide/core-semantics.md
```diff
@@ -0,0 +1,25 @@
...
+    - Operators that accept a "notifier", that is another observable source that is used to trigger some behavior, must accept any type that can be converted to an `Observable` with `from`. For example `takeUntil`.
+    - "Notifiers" that are provided to operators are expected to emit a value in order to trigger notification. Completion does not count as a notification.
+    - "Notifiers" provided directly to the operator MUST be subscribed to *before* the source is subscribed to. "Notifiers" created via factory function provided to the operator SHOULD be subscribed to at the earliest possible moment.
+    - The observable returned by the operator function is considered to be the "consumer" of the source. As such, the consumer MUST unsubscribe from the source as soon as it knows it no longer needs values before proceeding to do _any_ action.
+    - Events that happen after the completion of a source should happen after the source finalizes. This is to ensure that finalization always happens in a predictable time frame relative to the event.
```
backbone87, line 18: "event" is a confusing term here. is general behavior meant, like cleanup or downstream notifications?

docs_app/content/guide/core-semantics.md
backbone87: just working on a `window` style operator and while looking at the `window` source code I noticed that this operator family unsubscribed the nested observables immediately when the primary subscription is unsubscribed from:
https://stackblitz.com/edit/qsuarg?devtoolsheight=33&file=index.ts

backbone87: The more i dig into the window/groupBy use cases the more I think that it is much more intuitive if the group lifecycles are tied to the top-level subscription and consuming them outside of its context is an anti-pattern, if we talk about window/groupBy as operators. is there some widely used pattern that requires this?

i also tried to compare it to the refactorings done to the multicasting operators, where connect/share operators are also entirely handled within the primary subscription(s), while the connectable observable factory was a part of the multicast operator that was taken out of pipeline/operators into an observable factory so the `connect` method can be handled outside of the scope of a primary subscription.

docs_app/content/guide/core-semantics.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Purpose
+
+The purpose of these semantics is provide predictable behavior for the users of our library, and to ensure consistent behavior between our many different operators.
+
+## Operators
+
+    - MUST be a function that returns an [operator function](https://rxjs.dev/api/index/interface/OperatorFunction). That is `(source: Observable<In>) => Observable<Out>`.
+    - The observable returned by the operator function MUST subscribe to the source.
+    - If the operation performed by the operator can tell it not change anything about the output of the source, it MUST return the reference to the source. For example `take(Infinity)` or `skip(0)`.
```
kolodny, line 13: nit: ...can tell it **does** not change... I had trouble parsing this sentence without that

docs_app/content/guide/core-semantics.md
```diff
@@ -0,0 +1,25 @@
...
+    - "Notifiers" that are provided to operators are expected to emit a value in order to trigger notification. Completion does not count as a notification.
+    - "Notifiers" provided directly to the operator MUST be subscribed to *before* the source is subscribed to. "Notifiers" created via factory function provided to the operator SHOULD be subscribed to at the earliest possible moment.
+    - The observable returned by the operator function is considered to be the "consumer" of the source. As such, the consumer MUST unsubscribe from the source as soon as it knows it no longer needs values before proceeding to do _any_ action.
+    - Events that happen after the completion of a source should happen after the source finalizes. This is to ensure that finalization always happens in a predictable time frame relative to the event.
+    - `Error` objects should never be retained longer than necessary. This is a possible source of memory pressure.
+    - `Promise` references should never be retained longer than necessary. This is a possible source of memory pressure.
```
kolodny, line 20: Can you give examples or some guidance on what "longer than necessary" means? If I ever plan on sending the error to the user then, by definition, it will be necessary.

benlesh: I'm not sure what else to put here. Basically like creating an `Error` object well in advance of some async throw. They should be created on-demand if at all possible.

docs_app/content/guide/core-semantics.md
benlesh: > how to "detect" this case.

If the operator takes a source of `Observable<A>` returns `Observable<Observable<B>>`, (windowX, groupBy) that's how you'd know. There's nothing to really "detect" in this case.

> The more i dig into the window/groupBy use cases the more I think that it is much more intuitive...

Yeah, I don't disagree. I'm putting this in here as a starting/discussion point. A lot of people don't even realize that this is a thing. So, I've added it to the core semantics document so the core team can agree on a behavior and move that way over the course of 7 -> 8 -> 9 versions. It very well might be the correct decision to stop `groupBy` from behaving this way. Or maybe we need to add some sort of "eject" functionality... either way, it's a point of discussion, and I'm glad you're picking up on it.

benlesh: cc @cartant  @kwonoj @kolodny @niklas-wortmann ☝️

docs_app/content/guide/core-semantics.md
benlesh: We should probably link terms to the glossary.

benlesh:
@ReactiveX/rxjs-core I think we should try to get this landed and iterate on it. Otherwise it's going to flounder as a PR like most of our documentation PRs do. haha. I think the main thing I want to settle though is around the "splitting operators" mentioned above.

docs_app/content/guide/core-semantics.md
backbone87: > There's nothing to really "detect" in this case.

by "detect" I meant, how does the operator find out when to unsubscribe from the source. I know how groupBy does it and window* doesn't do it, but I would expect a general guideline here on what the correct "detection mechanism" is.

----
After giving this some more thought on my side, I have to justify my original opinion on this topic.
Lets first talk about window*:
```ts
const sub1 = of(1, 2, 3)
  .pipe(
    windowCount(1),
    take(1),
    mergeAll()
  ).subscribe()
  // I would expect this to emit `1` (it currently does not).

let w: Observable<number>;
const sub2 = of(1, 2, 3)
  .pipe(
    windowCount(1),
    take(1),
  )
  .subscribe((v) => void (w = v));
w.subscribe();
// I would expect this to be UB,
// it will never emit in a sync case and may or may not work in the async case

let w: Observable<number>;
const sub3 = of(1, 2, 3)
  .pipe(
    windowCount(1),
    take(1),
  )
  .subscribe((v) => void v.subscribe());
  // this could work and emit `1` (it currently does not),
  // but I would consider this bad practice and should be discouraged.
  // if it eases implementation then UB would be fine too IMHO.
```

why should the first example work? the reason is the semantics of the inner observable. it has a clear definition of its lifecycle: the first window starts before the first source emission, emits the following 1 next notification, and then completes. since windowCount only operates on the sequence of notifications (and not the contents of notifications) and/or external notifiers (on other variants of the window operator) it is "predictable" what the inner observables will emit and I can imagine uses cases for chaining it with `take` et al.

this is also still in line with what I originally said: the observable is consumed within its "primary" subscription (referring to `sub1` here), as the nested/inner subscriptions are "invisible" to the user. this is not the case for the second example.

in the groupBy case it's basically the same, but I just don't see any use case for `pipe(groupBy(selector), take(n))` without having the possibility to control group openings. this was proposed in https://github.com/ReactiveX/rxjs/pull/6425. currently this can be worked around with something like `pipe(startWith(x, y, z), groupBy(selector), map(skip(1)), take(n))`, but it does not feel very nice.

kolodny (approved):
Looks good, I'm very happy with having a doc like this to get everything on the same mindset. Agree that we can iterate on the harder splitting operators.

docs_app/content/guide/core-semantics.md
```diff
@@ -0,0 +1,33 @@
...
+
+## General Design Guidelines
+
+**Functions such as operators, constructors, and creation functions, should use named parameters in cases where there is more than 1 argument, and arguments after the first are non-obvious.** The primary use case should be streamlined to work without configuration. For example, `fakeFlattenMap(n => of(n))` is fine, but `fakeFlattenMap(n => of(n), 1)` is less readable than `fakeFlattenMap(n => of(n), { maxConcurrent: 1 })`. Other things, like `of(1, 2, 3)` are obvious enough that named parameters don't make sense.
+
+## Operators
+
+- MUST be a function that returns an [operator function](https://rxjs.dev/api/index/interface/OperatorFunction). That is `(source: Observable<In>) => Observable<Out>`.
+- The returned operator function MUST be [referentially transparent](https://en.wikipedia.org/wiki/Referential_transparency). That is to say, that if you capture the return value of the operator (e.g. `const double => map(x => x + x)`), you can use that value to operate on any many observables as you like without changing any underlying state in the operator reference. (e.g. `a$.pipe(double)` and `b$.pipe(double)`).
```
cartant, line 16: It's the _operator_ that must be referentially transparent; not the _operator function_. There should be no requirement for the _returned function_ to be referentially transparent.

Calls to `share` must be referentially transparent, but calls to the operator function - that `share` returns - are _not_.

My understanding of referential transparency is that a function is referentially transparent if you can replace its being called with the value that it returns, so it's the _operator_ that needs to be referentially transparent for this to make sense:

> That is to say, that if you capture the return value of the operator...

docs_app/content/guide/core-semantics.md
```diff
@@ -0,0 +1,33 @@
...
+- The observable returned by the operator function MUST subscribe to the source.
```
cartant, line 17: Should this really be a 'MUST'. There will be no subscription to the source for a `take(0)`, for example. Is that something that is going to have to change?

cartant (approved):
Approving this, but IMO the referentially-transparent bit isn't quite right.

docs_app/content/guide/core-semantics.md
trxcllnt: And if you [create a rejected Promise](https://github.com/ReactiveX/IxJS/pull/328/files) and don't call `.catch()` before the next turn of the event loop, node will throw an `unhandledRejection` event.

docs_app/content/guide/core-semantics.md
trxcllnt: Why not just multicast the `groupBy()` Observable and manually connect it?

benlesh: @trxcllnt LMAO. I swear I didn't write `groupBy` to behave like this, I thought you did. I agree it's weird. I only recorded this semantic because I wanted to draw attention to it. I'd be much happier if it did not have that behavior.

benlesh:
Okay... I'm going to merge this as-is, and we can make subsequent PRs to clean up the minutiae. I will be VERY happy to fix the semantics of `groupBy`, which IMO are a nightmare. I'm glad @trxcllnt pointed out that he didn't agree with it, either. haha.

PR #6436, opened 2021-05-29
Chore/config handler spies test helpers

backbone87:
**Description:**

I introduced a helper function in https://github.com/ReactiveX/rxjs/pull/6169/files#diff-9e451a38948f76e3294d51d3a213e0e9d0170c3d327a871e1435a457a8268125R28 to more easily test the global config handlers.

With this PR I want to extract it into a general helper function in `helpers/test-helper` and convert existing config tests (or other tests involving the global config handlers) to use the helper. This also makes the config tests more readable imho, because the code order now matches the execution order.

I also removed a duplicated test in `Observable-spec.ts` which had the same test code as one of the config-spec tests.

**BREAKING CHANGE:**
While implementing the helper and converting the config-spec tests, I noticed a small discrepancy between how `onUnhandledError` and `onStoppedNotification` are handled. In the last commit, I adjusted `onUnhandledError` handling to match that of `onStoppedNotification` and added appropriate tests. this might be considered a BC break in the following case:
```ts
// current behavior
config.onUnhandledError = () => console.log('I am not called.');

new Observable((subscriber) => {
  subscriber.error('bad');
}).subscribe();

config.onUnhandledError = () => console.log('But I am!');

// behavior after the change
config.onUnhandledError = () => console.log('I am called!');

new Observable((subscriber) => {
  subscriber.error('bad');
}).subscribe();

config.onUnhandledError = () => console.log('But I am not.');
```

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -577,7 +577,7 @@ describe('Observable', () => {
           });
       });
     });
-    
+
     it('should teardown even with a synchronous thrown error', () => {
       let called = false;
       const badObservable = new Observable((subscriber) => {
@@ -595,7 +595,7 @@ describe('Observable', () => {
       expect(called).to.be.true;
     });
 
-    
+
     it('should handle empty string sync errors', () => {
       const badObservable = new Observable(() => {
         throw '';
@@ -610,7 +610,7 @@ describe('Observable', () => {
       });
       expect(caught).to.be.true;
     });
-      
+
 
     describe('if config.useDeprecatedSynchronousErrorHandling === true', () => {
       beforeEach(() => {
@@ -705,7 +705,7 @@ describe('Observable', () => {
         expect(called).to.be.true;
       });
 
-      
+
       it('should handle empty string sync errors', () => {
         const badObservable = new Observable(() => {
           throw '';
@@ -738,7 +738,7 @@ describe('Observable', () => {
         }
         expect(called).to.be.true;
       });
-      
+
       it('should execute finalize even with a sync thrown error', () => {
         let called = false;
         const badObservable = new Observable(() => {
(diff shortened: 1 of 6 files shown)


backbone87:
added one more test case for the execution order of onUnhandledError & onStoppedNotification

benlesh:
Thank you for this pr @backbone87. I'm going to add this to the Core Team agenda so we can decide which behavior is appropriate. While it might be a breaking change for someone, I think I personally see this as a bug fix, and it might be worth introducing in a patch. These features aren't very broadly used yet, to my knowledge. Particularly in this edge case.

backbone87:
rebased against master

benlesh:
Core Team meeting again: Punting. Not enough for quorum

benlesh:
Core Team: This is a bug fix and not a breaking change. We don't want people to "know" that the error is being handling asynchronously if it was synchronously thrown, the change to the handler shouldn't be honored. This may also help set us up for `reportError` usage. Related: #6795

spec/Observable-spec.ts
```diff
@@ -862,24 +862,6 @@ describe('Observable', () => {
     });
   });
 
-  it('should not swallow internal errors', (done) => {
```
benlesh, line 865: I don't think this test should be removed?

benlesh (changes requested):

spec/Observable-spec.ts
backbone87: this is a duplicated test from here https://github.com/ReactiveX/rxjs/pull/6436/files#diff-554ac4d3a829f8491d378d79ce9937ca42984e12d35b9b2ee454dca704f00d83R191

benlesh:
I'm so sorry, @backbone87 ... but this PR has a lot going on in it, and it's been so long since I looked at it, that I'm going to close it for now, in favor handling the identified bug in another PR.

PR #6437, opened 2021-05-30
chore(docs): update multicasting in deprecations

lolimay:
**Description:**

The `share` operator should be imported from 'rxjs/operators'

none


cartant (approved):
Thanks.

PR #6447, opened 2021-06-03
fix: fromEvent() type with resultSelector

rocwang:
**Description:**

This PR simply fixes the return type of one overloaded form of `fromEvent()` that accepts a `resultSelector`.

With a `resultSelector: (event: T) => R`, `fromEvent<T, R>` should return a `Observable<R>`, instead of `Observable<T>`.

**Related issue (if exists):** N/A


cartant (approved):

cartant:
This change affects the public API, so the `api_guardian` files will need to be updated by running `npm run api_guardian:update`.

rocwang:
> This change affects the public API, so the `api_guardian` files will need to be updated by running `npm run api_guardian:update`.

@cartant Thanks for the reminder! The `api_guardian` files are updated now.

cartant:
I've had a closer look at the dtslint tests to see why this wasn't picked up. We should add two tests similar to these and should include the options argument so that those signatures are tested:

https://github.com/ReactiveX/rxjs/blob/7.1.0/spec-dtslint/observables/fromEvent-spec.ts#L23-L30

rocwang:
Good call. I've added 4 more dtslint tests there for `fromEvent`. Now all overload forms should be covered.

cartant (approved):
LGTM. Thanks.

rocwang:
[Just fixed a typo](https://github.com/ReactiveX/rxjs/pull/6447/commits/0455cc5e6a0b23ef397e4cfa49d81db2213cf379)

benlesh (approved):

Issue #6455, opened 2021-06-06
Could not find a declaration file for module 'rxjs'


eliasm307:
## Bug Report

**Current Behavior**
Trying to build a Typescript project that uses Rxjs and I get the following error:
![image](https://user-images.githubusercontent.com/13420359/120911217-4e315680-c67d-11eb-9711-f821dd9c5f8f.png)

**Expected behavior**
Should build without type errors as these are built into the rxjs package.

**Reproduction**

Try to build a project that imports types from Rxjs

**Environment**
- Runtime:  Node v15.0.1
- RxJS version: 7.1.0 
- 
**Possible Solution**
What fixed it for me was to change [this line](https://github.com/ReactiveX/rxjs/blob/c90242c66ea9c0f2ed0e7f0f0d7dcd7cf011c694/package.json#L8)
![image](https://user-images.githubusercontent.com/13420359/120911296-3a3a2480-c67e-11eb-8322-59dfb658bc0e.png)

to

![image](https://user-images.githubusercontent.com/13420359/120911353-cc422d00-c67e-11eb-9ba9-a730c111ea5a.png)

ie: "types": "./dist/types/index.d.ts", 

I am not sure what the intent was but the index.d.ts referenced for the typings doesn't seem to exist and/or the "typesVersions" field doesn't seem to produce intended results if its meant to allow types to be specified by the package version.

kwonoj:
What's the version of typescript you're using?

PR #6456, opened 2021-06-06
Updated package.json types path

eliasm307:
**Description:**

Updates typings path in package.json to point to the actual location of type declarations.

Minor update and tests seem to pass, some are just 11 are shown as pending but I am not sure if this is an issue with the tests, there are no explicitly failing tests however: 
image

Addresses #6455

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "./dist/cjs/index.js",
   "module": "./dist/esm5/index.js",
   "es2015": "./dist/esm/index.js",
-  "types": "index.d.ts",
+  "types": "./dist/types/index.d.ts",
   "typesVersions": {
     ">=4.2": {
       "*": [


kwonoj:
Dupe of https://github.com/ReactiveX/rxjs/pull/6293 , please confirm if you're on correct typescript version.

eliasm307:
@kwonoj see below the version I am on, what is the correct version?
image

kwonoj:
@eliasm307 please create reproducible repo and comment under issue you have filed: also we do not support nightly / dev version of typescript - it may work or not but we cannot ensure it always work.

eliasm307:
@kwonoj the issue turned out to be a typescript version issue as you suspected. My global ts version is 4.3.0 but the project I was working on has this defined locally as 4.1.5, the error is gone after updgrading. Thanks for your help!

PR #6464, opened 2021-06-09
fix(debounceTime): unschedule dangling task on unsubscribe before complete

backbone87:
**Description:**

Since `debounceTime` is not based on `debounce` anymore (https://github.com/ReactiveX/rxjs/pull/6049), the "notifier subscription" is not managed within the parent subscription anymore, but scheduled "raw". We need to make sure we dont leave dangling tasks behind.

I am not sure how to test this.

edit: this also prevents stopped notifications, if the source errors, because that could also leave dangling tasks behind

### DIFF
--- a/spec/operators/debounceTime-spec.ts
+++ b/spec/operators/debounceTime-spec.ts
@@ -1,10 +1,12 @@
 /** @prettier */
 import { expect } from 'chai';
-import { of, Subject } from 'rxjs';
-import { debounceTime, mergeMap } from 'rxjs/operators';
+import { NEVER, of, Subject } from 'rxjs';
+import { AnimationFrameAction } from 'rxjs/internal/scheduler/AnimationFrameAction';
+import { AnimationFrameScheduler } from 'rxjs/internal/scheduler/AnimationFrameScheduler';
+import { debounceTime, mergeMap, startWith } from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


cartant:
I think it would be 'more correct' to add the scheduled action - `activeTask` - to `subscriber`. As to how this should be tested I'll have a think about it.

backbone87:
didnt notice that this was possible, i pushed a version, where the activeTask subscription is managed within the outer subscriber

cartant:
It ought to be possible to test this using the `asyncScheduler`'s `actions` property, as is done here (with the `animationFrameScheduler`):

https://github.com/ReactiveX/rxjs/blob/3d251be5383dd738f2dec436d28af066568489fe/spec/schedulers/AnimationFrameScheduler-spec.ts#L107-L110

cartant (commented):
This change is fine, but this needs [a test](https://github.com/ReactiveX/rxjs/pull/6464#issuecomment-859257770) that's fixed by the change.

backbone87:
test added (and verified that it is failing without the fix)

cartant (approved):
LGTM. Thanks.

benlesh (approved):

benlesh:
Thank you, @backbone87

PR #6467, opened 2021-06-10
docs(operators): Rerun is changed to Return

matwming:
**Description:** -

**Related issue (if exists):** -

### DIFF
--- a/src/internal/operators/switchMap.ts
+++ b/src/internal/operators/switchMap.ts
@@ -56,7 +56,7 @@ export function switchMap<T, R, O extends ObservableInput<any>>(
  * // ... and so on
  * ```
  *
- * Return an interval Observable on every click event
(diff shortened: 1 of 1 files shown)


src/internal/operators/switchMap.ts
```diff
@@ -56,7 +56,7 @@ export function switchMap<T, R, O extends ObservableInput<any>>(
  * // ... and so on
  * ```
  *
- * Rerun an interval Observable on every click event
+ * Return an interval Observable on every click event
```
cartant, line 59: This probably isn't a typo and the author likely intended to use 'rerun', but your opening of this PR kinda shows that it's not as clear as it could be. Let's use 'restart' instead:
```suggestion
 * Restart an interval Observable on every click event
```

cartant (approved):

PR #6472, opened 2021-06-11
fix(useDeprecatedSynchronousErrorHandling): now throws properly when notifying a Subject

### LINKED ISSUE #6471: Super Gross Mode @ Google: Subjects no longer throw while nexting
version: 7.x

```ts

const source = subject.pipe(switchMap(() => throwError('bad')));

try {
  subject.next('bha');
} catch (err) {
  console.log('here');
}
```

cc @leggechr

benlesh:
Resolves #6471

### DIFF
--- a/src/internal/Subscriber.ts
+++ b/src/internal/Subscriber.ts
@@ -110,6 +110,10 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
     if (!this.closed) {
       this.isStopped = true;
       super.unsubscribe();
+      if (config.useDeprecatedSynchronousErrorHandling && (this.destination as any).__syncErrorHack_hasThrown) {
+        (this as any).__syncErrorHack_hasThrown = true;
+        (this as any).__syncError = (this.destination as any).__syncError;
+      }
       this.destination = null!;
     }
   }
@@ -199,6 +203,8 @@ function wrapForErrorHandling(handler: (arg?: any) => void, instance: SafeSubscr
         if ((instance as any)._syncErrorHack_isSubscribing) {
           (instance as any).__syncError = err;
         } else {
+          (instance as any).__syncErrorHack_hasThrown = true;
+          (instance as any).__syncError = err;
           // We're not currently subscribing, but we're in super-gross mode,
           // so throw it immediately.
           throw err;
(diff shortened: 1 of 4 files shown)


benlesh:
cc @leggechr

src/internal/Subscriber.ts
```diff
@@ -84,6 +84,9 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
    */
   error(err?: any): void {
     if (this.isStopped) {
+      if (config.useDeprecatedSynchronousErrorHandling) {
+        throw err;
+      }
```
benlesh, line 89: @leggechr the fix is this bit here.

src/internal/Subscriber.ts
```diff
@@ -83,7 +83,7 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
    * @return {void}
    */
   error(err?: any): void {
-    if (this.isStopped) {
+    if (this.isStopped && !config.useDeprecatedSynchronousErrorHandling) {
```
benlesh, line 86: @leggechr this is the fix.

api_guard/dist/types/index.d.ts
```diff
@@ -249,7 +249,7 @@ export interface ObjectUnsubscribedError extends Error {
 
 export declare const ObjectUnsubscribedError: ObjectUnsubscribedErrorCtor;
 
-export declare const observable: string | symbol;
+export declare const observable: string | typeof Symbol.observable;
```
cartant, line 253: Why has this changed? The last time this happened, IIRC, it was in that PR that sought to bump to TS 4.3. Are you sure you have 4.2 installed?

src/internal/Subscriber.ts
cartant: Surely this is going to cause other problems? This renders stopped meaningless for error notifications if that config is used? How did this ever work in v6? If it didn't, should we really change this? FWIW, I don't particularly care, but I thought I'd point it out.

cartant (approved):

cartant (changes requested):
Actually, the api_guardian thing needs to be addressed.

leggechr:
This caused tons more failing tests in google. The error is: `TypeError: Cannot read property 'error' of null` at this line: https://github.com/ReactiveX/rxjs/blob/master/src/internal/Subscriber.ts#L123

benlesh:
Completely redone... very hacky for now... only one issue left though (marked with `it.skip` in the tests). cc @cartant @leggechr

benlesh:
Closing in favor of #6478

PR #6478, opened 2021-06-16
refactor: More elegantly handle super gross mode

benlesh:
As an alternative to #6472 ... which is SHENANIGANS

this adds a cleaner solution for handling super-gross mode. It shouldn't have much impact on regular use.


benlesh:
cc @leggechr @cartant

cartant (approved):
LGTM. Nice. Gross, but nice.

benlesh:
Yeah, I can't wait to remove all of this junk.

Issue #5381, opened 2020-04-06
WebSocketSubject may have different input and output types.


benlesh:
Just what the title says... The `AnonymousSubject` (aka "frankensubject") probably also needs some typings love.

Basically, a nicer interface would be:

```ts
class WebSocketSubject<In, Out> extends Observable<Out> implements Observer<In> {
  // ...stuff
}
```

benlesh:
Note the work done here: #5390

raymonddavis:
@benlesh I would like to do this if help is still wanted. Not sure if I should just make a PR, unsure of the appropriate etiquette in the open-source world.

PR #6479, opened 2021-06-17
fix(WebSocketSubject): add different input and output types

raymonddavis:
**Description:**
add different input and output types to WebSocketSubject

#5381

### DIFF
--- a/api_guard/dist/types/webSocket/index.d.ts
+++ b/api_guard/dist/types/webSocket/index.d.ts
@@ -1,23 +1,27 @@
-export declare function webSocket<T>(urlConfigOrSource: string | WebSocketSubjectConfig<T>): WebSocketSubject<T>;
+export declare function webSocket<In, Out = In>(urlConfigOrSource: string | WebSocketSubjectConfig<In, Out>): WebSocketSubject<In, Out>;
 
-export declare class WebSocketSubject<T> extends AnonymousSubject<T> {
-    constructor(urlConfigOrSource: string | WebSocketSubjectConfig<T> | Observable<T>, destination?: Observer<T>);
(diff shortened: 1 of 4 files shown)


benlesh:
My guess is that this is a breaking change for _somebody_. While it wouldn't breaking explicitly typed usages of `webSocket`, I do think it may break implicitly typed versions of it. I don't have the time at the moment to figure out what those would be. It would be helpful if we had tslint tests around this _before_ we merged this change. So we can assert that this change doesn't break any of those.

raymonddavis:
@benlesh I forgot to run `api_guardian:update`. The files are updated now.

I can create a separate pr for dtslint tests for `webSocket`, I do not see any for `webSocket` currently in the project (looked for them here https://github.com/ReactiveX/rxjs/tree/7.1.0/spec-dtslint/observables/dom).

raymonddavis:
Is the failing check a result of something I changed? Seems odd it says
```
npm ERR! Could not resolve dependency:
npm ERR! peer typescript@"^3" from @phenomnomnominal/tsquery@4.0.0
```

raymonddavis:
I rebased from master to hopefully fix the failing check due to npm package dependency.

benlesh:
Hi @raymonddavis — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

Issue #6474, opened 2021-06-14
fromEvent incorrect arg order causes infinite recursion


ronag:
Doing:

```js
fromEvent('abort', signal)
```

Causes:

```js
    err: {
      "type": "RangeError",
      "message": "Maximum call stack size exceeded",
      "stack":
          RangeError: Maximum call stack size exceeded
              at isEventTarget (/usr/src/app/node_modules/rxjs/internal/observable/fromEvent.js:63:5)
              at setupSubscription (/usr/src/app/node_modules/rxjs/internal/observable/fromEvent.js:31:9)
              at setupSubscription (/usr/src/app/node_modules/rxjs/internal/observable/fromEvent.js:48:13)
              at setupSubscription (/usr/src/app/node_modules/rxjs/internal/observable/fromEvent.js:48:13)
              at setupSubscription (/usr/src/app/node_modules/rxjs/internal/observable/fromEvent.js:48:13)
              at setupSubscription (/usr/src/app/node_modules/rxjs/internal/observable/fromEvent.js:48:13)
              at setupSubscription (/usr/src/app/node_modules/rxjs/internal/observable/fromEvent.js:48:13)
              at setupSubscription (/usr/src/app/node_modules/rxjs/internal/observable/fromEvent.js:48:13)
              at setupSubscription (/usr/src/app/node_modules/rxjs/internal/observable/fromEvent.js:48:13)
              at setupSubscription (/usr/src/app/node_modules/rxjs/internal/observable/fromEvent.js:48:13)
    }
```

It should be

```js
fromEvent(signal, 'abort')
```

However, a bit better error handling might be appropriate here? Maybe at least check if the sourceObj is an array?

mgred:
This is because the given `string` is identified as an [_ArrayLike_](https://github.com/ReactiveX/rxjs/blob/master/src/internal/util/isArrayLike.ts). I will create a PR to fix the funtion to not let a `string` pass.

hgaleh:
How can we participate?

PR #6487, opened 2021-06-19
fix(isArrayLike): function will now return false when a string is given

mgred:
**Description:**

For now `isArrayLike` returns true when a `string` is given since a `string` has a `length` property of type `number`. This fix prevents the function from returning `true` in this case.

This PR also adds a test for `isArrayLike`.

#6474

### DIFF
--- a/spec/util/isArrayLike-spec.ts
+++ b/spec/util/isArrayLike-spec.ts
@@ -1,0 +1,38 @@
+import { expect } from 'chai';
+import { isArrayLike } from 'rxjs/internal/util/isArrayLike';
+
+describe('isArrayLike', () => {
+  it('should return true for new Array', () => {
+    const a = new Array();
+    expect(isArrayLike(a)).to.be.true;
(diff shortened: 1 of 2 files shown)


cartant:
`isArrayLike` is an internal function. It's not exported. If this is going to be changed, it really needs to be accompanied by a failing test that shows how this addresses a problem that can affect people using only the public API. If this change is going to be made,, whatever's added to the CHANGLOG needs to explain how it affects the public API; `isArrayLike` is an implementation detail.

benlesh:
Even if it were a public API (which it isn't), the fact is, `String` is "array like". `("test")[0] === "t"` or `("test").length === 4'. I don't think we can accept this PR, sorry. Thank you for the effort and thought!

mgred:
@cartant @benlesh
I understand, thanks for your feedback.

mgred:
@cartant @benlesh
Since the Test for `isArrayLike` is missing completly, do you see the need of adding this?

Issue #6242, opened 2021-04-20
Export all core APIs from `rxjs` module directly.


benlesh:
Given the issue #6067 seems to be fairly difficult to solve, and given that we have renamed everything such that there are no conflicts between things like `concat` the creation function, and `concat` the operator, etc. Maybe it's time we talk about exporting all primary APIs from `rxjs` proper.

1. It's a non-breaking change.
2. It's the direction we've all discussed going _many_ times.
3. It's easier to deal with, overall.

I'm talking about:

```ts
import { of, interval, take, concatWith, map } from 'rxjs';

interval(1000).pipe(
  take(3),
  concatWith(of('done')),
  map(x => `Tick: ${x}`)
)
```

benlesh:
Oh, and to be clear, I'm not talking about `rxjs/ajax` or `rxjs/fetch` or `rxjs/webSocket` or `rxjs/testing`. Overtime maybe those should be their own modules I'm not sure. For now, though, I wouldn't include them in this. I'm just talking about creation functions, subjects, operators, and schedulers.

benlesh:
@kolodny we'd only move the newer name of the `concat` operator, which is `concatWith`.

LayZeeDK:
Any notable effect on tree-shaking or overall TypeScript compilation speed?

benlesh:
Core Team Meeting: General core team approval

tmair:
Just as a side note. This change could also enable the use of [API Extractor](https://api-extractor.com/) at least for the `rxjs` import location. The tool might help to automate detection of changes to the API surface of the library.

PR #6488, opened 2021-06-21
feat: operators are all exported at the top level, from "rxjs".

benlesh:
This only moves the operators to the top level. Other things like `ajax`, `webSocket`, `testing`, and `fromFetch` all require more debate.

Related #6242

(Ironically, this will cause tools like Bundlephobia to be like "Whoa! This is bigger!" because they excluded the "deep" imports we had, like `rxjs/operators`)


api_guard/dist/types/index.d.ts
```diff
@@ -249,7 +411,7 @@ export interface ObjectUnsubscribedError extends Error {
 
 export declare const ObjectUnsubscribedError: ObjectUnsubscribedErrorCtor;
 
-export declare const observable: string | symbol;
```
benlesh, line 252: I have no idea why this keeps happening. :\

benlesh: Apparently it's TS 4.3 inferring the type locally. I'll fix that here too.

cartant (approved):

PR #6501, opened 2021-07-02
feat: add generic type parameter to using for the resource

jennings:
This overload alleviates the observableFactory from needing to cast its parameter to a more specific type.

**Description:**

Previously, a TypeScript user has to cast the resource to the appropriate type in order to do anything with it:

```
const o: Observable<MyResource> = using(
  () => new MyResource(),
  r => of(r as MyResource)   // t is `Unsubscribable | void`, so a cast is needed
)
```

Adding a type parameter for the resource lets TypeScript infer the type of `r` more precisely.

I asked in [discussion][1] about this and there didn't seem to be a good reason not to implement it.

**BREAKING CHANGE (maybe):** @cartant suggested this might be a breaking change. There's no functionality change, but there's a worry that it will cause existing code to no longer type-check. The existing signature is in place, so I think at _worst_ it will remove the `void` from the first parameter's type. That doesn't seem breaking to me, but the maintainers here have definitely seen more weird cases than I have.

* [Discussion][1]

[1]: https://github.com/ReactiveX/rxjs/discussions/6497

### DIFF
--- a/spec-dtslint/observables/using-spec.ts
+++ b/spec-dtslint/observables/using-spec.ts
@@ -1,4 +1,4 @@
-import { using } from 'rxjs';
+import { using, Subscription, of} from 'rxjs';
 import { a$, b$ } from '../helpers';
 
 it('should infer with a simple factory', () => {
@@ -7,4 +7,8 @@ it('should infer with a simple factory', () => {
 
 it('should infer with a factory that returns a union', () => {
   const o = using(() => {}, () => Math.random() < 0.5 ? a$ : b$); // $ExpectType Observable<A | B>
+});
+
+it('should infer resource types', () => {
+  const o = using(() => new Subscription(), (sub) => of(sub)); // $ExpectType Observable<Subscription>
 });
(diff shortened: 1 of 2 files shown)


benlesh:
You'll need to run `npm run api_guardian:update` and commit the new golden files to fix the CI errors.

jennings:
@benlesh Thanks, I added that to my commit and pushed.

benlesh:
@jennings So sorry this is such a late response. I don't have a way to force the checks to rerun without getting into your branch and pushing up an empty commit or something. Github isn't letting me approve the checks to be run. If you can, please rebase and push again so I can approve the checks to be run. Otherwise, I guess we can open a new PR with this work.

jennings:
> If you can, please rebase and push again so I can approve the checks to be run.

Done.

bman654:
I've encountered this as well.  My workaround is to add this typings file to my project


```typescript
// src/@types/rxjs-using.d.ts

import type { Observable, ObservableInput, ObservedValueOf, Unsubscribable } from "rxjs"

declare module "rxjs" {
  // fix type declaration of using.
  export function using<T extends ObservableInput<any>, R extends Unsubscribable | void>(
    resourceFactory: () => R,
    observableFactory: (resource: R) => T | void
  ): Observable<ObservedValueOf<T>>
}
```

PR #6503, opened 2021-07-05
chore: fix TypeScript latest breakage in CI

cartant:
**Description:**

CI seems to be failing with the latest TypeScript and Node types. This PR should fix the failures.

### DIFF
--- a/spec/Subscriber-spec.ts
+++ b/spec/Subscriber-spec.ts
@@ -244,7 +244,7 @@ describe('Subscriber', () => {
   });
 
   const FinalizationRegistry = (global as any).FinalizationRegistry;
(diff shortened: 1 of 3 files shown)


spec/Subscriber-spec.ts
```diff
@@ -262,7 +262,7 @@ describe('Subscriber', () => {
       const subscription = of(42).subscribe(observer);
 
       observer = undefined;
-      global.gc();
+      global.gc?.();
```
cartant, line 265: The `?.` is necessary despite the inclusion of `global.gc` in the `if` statement's expression, as the invocation is made within a callback, etc.

PR #6514, opened 2021-07-08
fix(share): Minor fix of subscription name in example

samal-rasmussen:
A bit hard to see the that there are two subscriptions if they have the same name.


cartant (approved):
LGTM. Thanks.

PR #6522, opened 2021-07-14
fix(AsyncSubject): properly emits values during reentrant subscriptions

### LINKED ISSUE #6520: Regression from 6 to 7 with subscribing to a subject from within a next call
## Bug Report

**Current Behavior**
With an AsyncSubject, previously if you subscribed to the same subject from within a next call, your inner subscription would also receive the next call. As of RxJs 7, that no longer happens. The inner subscription does not receive the next. I am aware this is not the recommended behavior but it did cause a regression in our system.

**Expected behavior**
When subscribing to an AsyncSubject, I would expect to receive the next value immediately regardless of the context from which that subscription was wired up.

**Reproduction**

Version 6: https://jsfiddle.net/4e3tdabg/6/ 
Version 7: https://jsfiddle.net/6px4nb3a/

```js
const sub = new rxjs.AsyncSubject();

sub.subscribe({
   next: x => {
      sub.subscribe({
         next: xx => console.log('inner next', xx + xx),
         complete: () => console.log('inner complete'),
      });

      console.log('outer next', x);
   },
   complete: () => console.log('outer complete')
});

sub.next(3);
sub.complete();
```

**Environment**
- Runtime: Latest Chrome, Latest Firefox
- RxJS version: 7

Expected
image

Actual
image

benlesh:
Resolves #6520

### DIFF
--- a/spec/subjects/AsyncSubject-spec.ts
+++ b/spec/subjects/AsyncSubject-spec.ts
@@ -211,7 +211,7 @@ describe('AsyncSubject', () => {
     expect(calls).to.equal(1);
   });
 
-   it('should not be reentrant via next', () => {
+  it('should not be reentrant via next', () => {
     const subject = new AsyncSubject<number>();
     let calls = 0;
     subject.subscribe({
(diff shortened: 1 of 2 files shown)


spec/subjects/AsyncSubject-spec.ts
```diff
@@ -211,7 +211,7 @@ describe('AsyncSubject', () => {
     expect(calls).to.equal(1);
   });
 
-   it('should not be reentrant via next', () => {
+  it('should not be reentrant via next', () => {
```
benlesh, line 214: We need to get prettier going on this file. This was just moving a space that was bothering me. haha.

src/internal/AsyncSubject.ts
```diff
@@ -14,10 +14,10 @@ export class AsyncSubject<T> extends Subject<T> {
 
   /** @internal */
   protected _checkFinalizedStatuses(subscriber: Subscriber<T>) {
-    const { hasError, _hasValue, _value, thrownError, isStopped } = this;
+    const { hasError, _hasValue, _value, thrownError, isStopped, _isComplete } = this;
     if (hasError) {
       subscriber.error(thrownError);
-    } else if (isStopped) {
+    } else if (isStopped || _isComplete) {
```
benlesh, line 20: Basically, there's a temporal space where we're `_isComplete`, but not `isStopped` while we're nexting to consumers in our `complete()` method. If someone subscribes during that, we end up hitting this code, and if we're only checking `isStopped` it's not "stopped" yet, so we didn't notify the consumer.

cartant (approved):

PR #6526, opened 2021-07-16
chore: update pluck tests to run mode

jakovljevic-mladen:
**Description:**
Updated `pluck` tests to use run mode.

There are two commits that I can exclude if needed (and possibly move to another PR), please let me know if I should.

The first one removes unused constant (its never been altered anywhere, it can be safely removed).

The second one adds firehose test for `pluck`, can this go with this PR or should I move it to another one?

None


benlesh (approved):
So much, better. Thank you, @jakovljevic-mladen

PR #6527, opened 2021-07-16
feat(tap): Adds subscribe, unsubscribe, finalize handlers

benlesh:
This adds a common request/task for RxJS users, which are three new handlers:

- `subscribe`: fires on subscription to the source
- `unsubscribe`: fires when the subscription to the result is unsubscribed from, but NOT if the source completes or errors
- `finalize`: always fires on finalization, (basically equivalent to `finalize`)

This is useful for situations where you'd like to know that your consumer has decided to unsubscribe (and you're not terminating because of a complete or an error).

```ts
let subject = new Subject();

const sub1 = subject.pipe(
   tap({ 
      subscribe: () => console.log('subscribed'),
      complete: () => console.log('complete'),
      unsubscribe: () => console.log('unsubbed'),
      finalize: () => console.log('finalize')
   })
).subscribe();
// logs "subscribed"

sub1.unsubscribe();
// logs "unsubbed"
// logs "finalized"

const sub1 = subject.pipe(
   tap({ 
      complete: () => console.log('complete'),
      unsubscribe: () => console.log('unsubbed')
   })
).subscribe();
// logs "subscribed"

subject.complete();
// logs "complete"
// logs "finalized"
```

See included tests for more information.

### DIFF
--- a/src/internal/operators/tap.ts
+++ b/src/internal/operators/tap.ts
@@ -125,7 +125,6 @@ export function tap<T>(
         ({ next: observerOrNext as Exclude<typeof observerOrNext, Partial<TapObserver<T>>>, error, complete } as Partial<TapObserver<T>>)
       : observerOrNext;
 
-  // TODO: Use `operate` function once this PR lands: https://github.com/ReactiveX/rxjs/pull/5742
   return tapObserver
     ? operate((source, subscriber) => {
         tapObserver.subscribe?.();
--- a/src/internal/operators/tap.ts
+++ b/src/internal/operators/tap.ts
@@ -4,7 +4,13 @@ import { operate } from '../util/lift';
 import { OperatorSubscriber } from './OperatorSubscriber';
 import { identity } from '../util/identity';
 
-export function tap<T>(observer?: Partial<Observer<T>>): MonoTypeOperatorFunction<T>;
+export interface TapObserver<T> extends Observer<T> {
+  subscribe: () => void;
+  unsubscribe: () => void;
+  finalize: () => void;
+}
+
+export function tap<T>(observer?: Partial<TapObserver<T>>): MonoTypeOperatorFunction<T>;
 export function tap<T>(next: (value: T) => void): MonoTypeOperatorFunction<T>;
 /** @deprecated Instead of passing separate callback arguments, use an observer argument. Signatures taking separate callback arguments will be removed in v8. Details: https://rxjs.dev/deprecations/subscribe-arguments */
 export function tap<T>(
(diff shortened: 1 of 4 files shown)


src/internal/operators/tap.ts
```diff
@@ -106,19 +112,24 @@ export function tap<T>(
...
   // but if error or complete were passed. This is because someone
   // could technically call tap like `tap(null, fn)` or `tap(null, null, fn)`.
   const tapObserver =
-    isFunction(observerOrNext) || error || complete ? { next: observerOrNext as (value: T) => void, error, complete } : observerOrNext;
+    isFunction(observerOrNext) || error || complete
+      ? // tslint:disable-next-line: no-object-literal-type-assertion
+        ({ next: observerOrNext as Exclude<typeof observerOrNext, Partial<TapObserver<T>>>, error, complete } as Partial<TapObserver<T>>)
+      : observerOrNext;
 
   // TODO: Use `operate` function once this PR lands: https://github.com/ReactiveX/rxjs/pull/5742
```
backbone87, line 128: nit: i think this comment can be removed

backbone87 (commented):
what is the reasoning behind doing this only for the tap operator and not for Subscribable.subscribe?

kwonoj:
Just checking: did we discuss this in core meeting?

benlesh:
@backbone87 the main reasoning is we're not attempting to change the semantics of observable, (which `Subscribable` is really just a super type of), this is about side-effects and observation on the part of the _producer_, which `tap` is defining. 

```ts
import { interval, share, tap, take } from 'rxjs';

const producer = interval(1000).pipe(
  share(),
  take(100),
  tap({ 
    unsubscribe: () => console.log('a user has unsubscribed'),
  })
);

// At this point `producer` could be handed off to anyone.
// `tap` is concerned with side effects that _producer_ needs. Not the consumer.


// Here we have a consumer
const subscription = producer.subscribe({
   next: console.log,
   // (hypothetical)
   unsubscribe: () => console.log('you unsubscribed yourself')
});

// Why didn't you just console log here? :shrug:
subscription.unsubscribe(); 
```

the `subscribe` handler on the consumer end is honestly just as silly:

```ts
producer.subscribe({
  // Hypothetical:
  subscribe: () => console.log('you subscribed, lol');
});
```

Finalize *might* have merit however, and that's possibly worth discussing. Interestingly, it can be handled now like so (it's just non-obvious):

```ts
const subscription = producer.subscribe(console.log);

subscription.add(() => console.log('finalized');
```

In the future if we're using AbortSignal, and we land on not minding async completion (That's probably a non-starter), we could leverage `forEach` and `Promise.prototype.finally`:

```ts
// Hypothetical
const controller = new AbortControlelr();
const { signal } = controller;
producer.forEach(console.log, { signal }).finally(() => console.log('finalized'))
```

You can obviously do this now with `forEach`, it's just not cancellable. :)

benlesh:
@kwonoj 

> Just checking: did we discuss this in core meeting?

Yes, but it's been a long time since we did.

cartant (approved):

kolodny (approved):

PR #6538, opened 2021-07-27
docs(operators): clarify that `find` can emit `undefined`

lazarljubenovic:
**Description:** Was bitten by this; I expected no emission to happen if nothing is found. Thought it would be good to explicitly state the edge case in the docs, since just “it won't throw” don't really say anything about what _will_ happen instead (nothing vs emits undefined).


voliva:
Note that the commit message (and PR title) is wrong - it's referencing `filter`, but the operator docs that's fixing is `find`.

lazarljubenovic:
My bad, was trying to multitask too much 😄 It should be easy for the merger to change the `filter` to `find` from the commit message in GitHub UI (easier than me pulling and amending).

benlesh (approved):

benlesh:
Thank you @lazarljubenovic

Issue #5370, opened 2020-04-01
Exception not thrown when maximum call stack is exceeded


reline:
## Bug Report

**Current Behavior**
When subscribing to a stream that emits errors indefinitely and returning the original source when catching said error, eventually the maximum call stack is exceeded. I expect an exception to be thrown in this case, but it is currently swallowed and printed to the console.

**Reproduction**
```js
import { of } from 'rxjs';
import { map, catchError } from 'rxjs/operators';
import { TestScheduler } from 'rxjs/testing';

const observable = of(4).pipe(
  map(n => {
      if (n === 4) {
        throw 'four!';
    }
    return n;
  }),
  catchError((err, source) => source),
);
 
const testScheduler = new TestScheduler((actual, expected) => {
  // using jest
  expect(actual).toMatchObject(expected);
});
 
// This test should fail due to Maximum call stack size exceeded
it('should fail', () => {
  testScheduler.run(helpers => {
    const { expectObservable } = helpers;
    expectObservable(observable).toBe('-');
  });
});
```
```
 PASS  __tests__/callstack-test.ts
  ✓ should fail (14ms)

  console.warn node_modules/rxjs/internal/Observable.js:55
    RangeError: Maximum call stack size exceeded
        at new SourceMapConsumer (/dev/node_modules/source-map-support/node_modules/source-map/lib/source-map-consumer.js:17:22)
        at mapSourcePosition (/dev/node_modules/source-map-support/source-map-support.js:190:14)
        at wrapCallSite (/dev/node_modules/source-map-support/source-map-support.js:367:20)
        at Function.prepareStackTrace (/dev/node_modules/source-map-support/source-map-support.js:416:39)
        at prepareStackTrace (internal/errors.js:37:29)
        at formatError (internal/util/inspect.js:907:19)
        at formatRaw (internal/util/inspect.js:703:14)
        at formatValue (internal/util/inspect.js:591:10)
        at inspect (internal/util/inspect.js:221:10)
        at formatWithOptions (internal/util/inspect.js:1693:40)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        1.302s, estimated 2s
```

**Expected behavior**
I would like the exception to be thrown and for this test to fail.

**Environment**
- Runtime: Node v12.16.1
- RxJS version: 6.5.4

**Additional context/Screenshots**
[...]

PR #6540, opened 2021-07-29
chore: add disambiguateByNonOperator disambiguator

### LINKED ISSUE #6539: Create new disambiguator to remove warning when building docs
Since operators are now exported from top level module (as of #6488), when building docs, a multiple warnings are produced. For example:

```
warn:    Error in {@link raceWith} - Ambiguous link: "raceWith".
Matching docs:
  "index/raceWith" (function) : (api/index/function/raceWith / internal/operators/raceWith.ts)
  "operators/raceWith" (function) : (api/operators/raceWith / internal/operators/raceWith.ts) - doc "api/deprecations" (deprecation)
```

The reason is that there are multiple documents that exist in two modules: `index` module and `operators` module. For this reason, `{@link raceWith}` won't work since package that builds links can't decide which one to take. Thus, a new disambiguator should be created to remove operators from `operators` module.

jakovljevic-mladen:
This PR fixes all errors when building docs. It add new disambiguator that removes duplicates when building docs. Since most of the operators are now exported in both root (`'rxjs'`) and `/operators` paths, a lot of duplicates were created. With this disambiguator, precedence is given to those from `'index'` module (`'rxjs'` import site).

Also, existing `disambiguateByDeprecated` disambiguator is changed for this reason: many duplicated documents that were deprecated are filtering all existing documents with this disambiguator  which lead to yet another error saying that no documents are found.

Side note: for couple of months now, running `npm run docs` has always slightly modified `/concatAll.svg` file, which I committed now - this is totally unrelated to the issue solved with this PR. If not needed, please let me know so I can remove it.

Closes #6539


jakovljevic-mladen:
I'm not sure why this test is failing. It doesn't seem to be related to the changes from this PR. Seems related to #5370 and #4089 (@cartant).

cartant:
> I'm not sure why this test is failing.

Ben encountered the same failing test - in CI - recently and was unable to get it to fail locally. Rerunning the CI tests 'fixed' the problem there and here, too.

jakovljevic-mladen:
@cartant thanks!

benlesh (approved):

benlesh:
Thank you @jakovljevic-mladen!

PR #6542, opened 2021-07-30
test: add more test cases to isPromise spec

ftonato:
Hello,

**Description:**
I'm just ensuring that some other data types, which aren't also of the "isPromise" type, don't generate false negatives.

### DIFF
--- a/spec/util/isPromise-spec.ts
+++ b/spec/util/isPromise-spec.ts
@@ -34,5 +34,18 @@ describe('isPromise', () => {
   it('should return false for a string', () => {
     expect(isPromise('1')).to.be.false;
   });
-
+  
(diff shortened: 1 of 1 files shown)


benlesh:
Hi @ftonato! I really appreciate the thought and effort here. But honestly, I don't think `isPromise` really needs a stronger battery of unit tests. Instead, it's most important that the places where `isPromise` is used are sufficiently integration tested. (`from` and `scheduled`). In other words, we want to make sure that promises are handled correctly in those two spots, and that non-observable things aren't handled incorrectly.

Sorry, but I'm going to close this one for now.

Thank you so much for your contribution, and I hope me closing this issue doesn't dissuade you from contributing in the future!

PR #6543, opened 2021-07-30
refactor: import only what is being used from the lodash library

ftonato:
Hello,

**Description:**
I decided to rewrite the way you are using the _lodash_ import so that it only loads what is needed into each file.

Most of the changes have been made to files that are only used for project documentation, so there probably won't be any performance gains in the library (:sad-parrot-emoji:)...


kwonoj:
Thanks for the suggestion, but I don't think we'll need this change. These are codepaths does not being bundled or import-time critical, so having all lodash operator is not a concern but also we can manage our package.json compact instead of listing each dependencies we need for the future.

PR #6545, opened 2021-08-02
chore: deprecate observable symbol export

### LINKED ISSUE #6489: Export `observable` (the symbol) under a different name or location.
I think this is unfortunate, because we might want to export something else as `observable`, and it's a bit confusing for people at times I imagine:

```ts
import { observable } from 'rxjs';

console.log(observable); // symbol
```

We're currently exporting our copy of `Symbol.observable` as `observable` from the main module. I think we should rename it, remove it, or move it under something else. 

Options:

1. Export it as `observableSymbol` or `OBSERVABLE_SYMBOL`
2. Don't export it at all, and advise people to use a package like `symbol-observable`
3. Export it as a static property of `Observable` or something. like `Observable.SYMBOL`. (minimal overhead there, it's more of stylistic thing, I'm not saying I like it. Just an option)
4. ???? Something else.

**In all cases we'd have to leave the `observable` export for the duration of 7.x, and simply deprecate it**.

benlesh:
resolves #6489

### DIFF
--- a/src/internal/symbol/observable.ts
+++ b/src/internal/symbol/observable.ts
@@ -1,2 +1,7 @@
-/** Symbol.observable or a string "@@observable". Used for interop */
+/**
+ * Symbol.observable or a string "@@observable". Used for interop
+ *
+ * @deprecated We will no longer be exporting this symbol in upcoming versions of RxJS.
+ * Instead polyfill and use Symbol.observable directly *or* use https://www.npmjs.com/package/symbol-observable
+ */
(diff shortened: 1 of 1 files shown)


benlesh:
Core Team: 

+ Yes to deprecation.
+ Probably should rename it. 

@cartant will have a look.

benlesh:
Ping @cartant

benlesh:
Core team: Ping @cartant <3

PR #6548, opened 2021-08-06
docs: introduce importing document

jakovljevic-mladen:
**Description:**
As of RxJS version 7.2.0, a new way of importing operators is introduced. This PR adds a document that covers this addition.

None

### DIFF
--- a/docs_app/content/guide/importing.md
+++ b/docs_app/content/guide/importing.md
@@ -1,0 +1,163 @@
+# Importing instructions
+
+There are different ways you can {@link guide/installation install} RxJS. Using/importing RxJS depends on
+the used RxJS version, but also depends on the used installation method.
+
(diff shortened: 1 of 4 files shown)


benlesh:
@jakovljevic-mladen this is a very detailed document. I think one thing I would like to see if that _directly at the top_ as bold as possible, we state very simply and plainly as possible that as of 7.2, no one should be importing things from `rxjs/operators`, and should import those things from `rxjs` instead.

Like, I wouldn't make people scroll or read too far at all before they saw that. It should be bold and obvious.

niklas-wortmann:
We do have the `<span class="informal">CONTENT GOES HERE</span>` that we use in several places to give a TLDR. I think that would be good to highlight that you don't need `rxjs/operators` anymore

PR #6556, opened 2021-08-11
update umd rxjs version used by https://rxjs.dev

pgherveou:
**Description:**
Update the RxJS version used by https://rxjs.dev to the latest version, so we can test the latest features directly from the website console


benlesh (approved):

benlesh:
I believe we actually want to remove this to prevent it impacting load times. Most people use the Stackblitz links now.

PR #6557, opened 2021-08-11
Update switchScan doc

pgherveou:
**Description:**

Update the informal doc to refer to mergeScan instead of scan. lmk if I am wrong!


cartant (approved):
Thanks.

PR #6562, opened 2021-08-13
fix: scheduling with Rx-provided schedulers will no longer leak action references

### LINKED ISSUE #6561: (Likely) Scheduled action leaks
It seems likely to me that some - if not all - of the following `scheduler.schedule` calls will leak actions that are added to another subscription - similar to the leak that was fixed in #6559:

https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/src/internal/observable/combineLatest.ts#L296

https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/src/internal/operators/bufferTime.ts#L120

https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/src/internal/operators/bufferTime.ts#L128-L133

https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/src/internal/operators/mergeInternals.ts#L117

https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/src/internal/operators/windowTime.ts#L139

https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/src/internal/operators/windowTime.ts#L147-L152

https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/src/internal/scheduled/scheduleAsyncIterable.ts#L11-L25

https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/src/internal/scheduled/scheduleObservable.ts#L9-L16

https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/src/internal/scheduled/schedulePromise.ts#L9-L14

https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/src/internal/util/caughtSchedule.ts#L11-L18

Some of these might leak only a few actions, but any leaked actions will have a closure that could be referencing a resource.

@benlesh

benlesh:
Resolves #6561

Note that I haven't quite figured out how I want to test this yet.


src/internal/util/executeSchedule.ts
```diff
@@ -0,0 +1,44 @@
...
+): void;
+export function executeSchedule(
+  parentSubscription: Subscription,
+  scheduler: SchedulerLike,
+  work: () => void,
+  delay?: number,
+  repeat?: false
+): Subscription;
+
+export function executeSchedule(
```
benlesh, line 19: The magic is here.

src/internal/util/executeSchedule.ts
```diff
@@ -0,0 +1,44 @@
+import { Subscription } from '../Subscription';
+import { SchedulerAction, SchedulerLike } from '../types';
+
+export function executeSchedule(
```
benlesh, line 4: I've added two type overloads for this internal function, to make sure we don't use this in a way that would leak with poorly written user-implemented schedulers.

benlesh:
Good grief, I hate schedulers.

kwonoj:
Change looks ok, but looks like there's a new circular dependency?

benlesh:
@kwonoj I've eliminated the circular dependencies as well as eliminating some more code that didn't really make sense.

src/internal/util/executeSchedule.ts
```diff
@@ -0,0 +1,44 @@
...
+  parentSubscription: Subscription,
+  scheduler: SchedulerLike,
+  work: () => void,
+  delay = 0,
+  repeat = false
+): Subscription | void {
+  const scheduleSubscription = scheduler.schedule(function (this: SchedulerAction<any>) {
+    work();
+    if (repeat) {
+      parentSubscription.add(this.schedule(null, delay));
```
cartant, line 29: I can't see why it's necessary to add the action to the parent subscription if it repeats. It should already have been added and it won't be removed because it's repeating and the `unsubscribe` call - in the other branch of this `if` statement - is going to be skipped.

benlesh: Because someone can create their own Scheduler/Action implementations right now. (I hate exporting classes). And there's no guarantee they're properly reusing actions.

cartant: Can check for that? Internal actions should be returning themselves as the subscription.

```ts
const repeatSubscription = this.schedule(null, delay);
if (repeatSubscription !== this) {
  parentSubscription.add(repeatSubscription);
}
```

cartant (approved):

kwonoj (approved):

Issue #6544, opened 2021-08-01
Docs around new operators export site


jakovljevic-mladen:
Hi,

I have two questions and I would like to check if it's worth investing my time for creating PRs.

As of recently, operators are exported from a top level export site (`'rxjs'`). However, the old export site has still remained (`'rxjs/operators'`) and that is now creating "problems" because many documentation docs are now being duplicated. For example, `map` operator documentation can be accessed from http://rxjs.dev/api/operators/map and from http://rxjs.dev/api/index/function/map (this link doesn't work because docs site is still using old version, but once rxjs.dev is updated, it will start working). The first URL is the standard one that has existed ever since `'rxjs/operators'` export site has been introduced. The later is the new one (as of RxJS v7.2.0).

Well, what's the difference? There is really no difference in the generated documentation content. However, documentation links that look like this `{@link map}` can't point to both URLs, but a new [disambiguator had to be created](https://github.com/ReactiveX/rxjs/pull/6540) to exclude those that come from `operators` module. So, if mentioned PR is approved and merged, all links that look like `{@link map}` will point to http://rxjs.dev/api/index/function/map. Using links that have this syntax is good because a warning is raised when building docs - when something happens to the original document, for example, when one of the deprecated APIs gets removed in V8, all links that point to the removed class/interface/function will stop working and warnings will start appearing. This is how such links will be discovered quite easily because of this.

In the end, `'rxjs/operators'` export site will be removed eventually, thus, an update to the documentation will have to be made. For now, many `.md` documents don't use `{@link map}` syntax to create links, but rather use standard markup syntax, that goes like this: [`[map](/api/operators/map)`](https://github.com/ReactiveX/rxjs/blob/6f0e853a86e6db1ab321ff077c95e6b1d88fd3df/docs_app/content/guide/operators.md#L161). For this reason, once `'rxjs/operators'` export site is removed, `[map](/api/operators/map)` link will stop working - it will produce 404 error because there won't be a document at that path. My question is: shall I create a PR to replace links that look like `[map](/api/operators/map)` to the ones that look like `{@link map}`?

Also, since operators are just functions, the new URL does not indicate that `/api/index/function/map` is an operator. There is no distinction between `map` and for example `isObservable`, both are functions. I wanted to add a label to top, next to other labels. Something like this:
![image](https://user-images.githubusercontent.com/28087049/127749040-af527aab-f59e-43fe-8536-9b860db8479c.png)
[...]

PR #6563, opened 2021-08-13
docs: add operator flag to operator functions

jakovljevic-mladen:
**Description:**
This PR adds a label to all functions that are considered as operator functions. That is, any function that is exported from `src/internal/operators` will have this label.

The processor that was created to determine if a document `isOperator` determines this by checking if a document has `originalModule` and if `originalModule` starts with `'internal/operators'`. This means that any exported function that is not operator that may in future come from `'internal/operators'` will be marked as an operator - for now, there are no such functions. But if some util (non-operator) function gets exported in future from `'internal/operators'`, it should be moved to another place so that it doesn't get marked as operator.

This is a handy addition, especially after RxJS v7.2.0 since operators are exported from `'rxjs'` and many operators are now displayed on two pages (for example: http://rxjs.dev/api/operators/map and http://rxjs.dev/api/index/function/map).

Some examples how this looks like (please notice URL):

image

image

image

`zip` creation function, exported from `'rxjs'`:
image

Deprecated `zip` operator, exported from `'rxjs/operators'`:
image

Note: almost all (if not all) [processors](https://github.com/ReactiveX/rxjs/tree/5d0e41ee46337f56dba6d9a1ef938002b261ccac/docs_app/tools/transforms/angular-base-package/processors) come with a `.spec.js` file. I didn't add one with this PR as so many tests actually fail when [tests are run](https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/docs_app/package.json#L30). For this reason, I don't think those tests are really essential. If they ever become essential, I will write some.

Somewhat #6544.

### DIFF
--- a/docs_app/src/styles/2-modules/_label.scss
+++ b/docs_app/src/styles/2-modules/_label.scss
@@ -56,6 +56,10 @@ label.raised, .api-header label {
             &.impure-pipe {
                 background-color: $brightred;
             }
+
+            &.operator {
+                background-color: $brightred;
+            }
         }
 
         &.api-type-label {
--- a/docs_app/tools/transforms/templates/api/base.template.html
+++ b/docs_app/tools/transforms/templates/api/base.template.html
@@ -26,6 +26,7 @@
     {% if doc.experimental !== undefined %}<label class="api-status-label experimental">experimental</label>{% endif %}
     {% if doc.stable !== undefined %}<label class="api-status-label stable">stable</label>{% endif %}
     {% if doc.pipeOptions.pure === 'false' %}<label class="api-status-label impure-pipe">impure</label>{% endif %}
+    {% if doc.isOperator %}<label class="api-status-label operator">operator</label>{% endif %}
   </header>
   <aio-toc class="embedded"></aio-toc>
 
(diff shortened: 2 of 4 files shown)


docs_app/src/styles/2-modules/_label.scss
```diff
@@ -56,6 +56,10 @@ label.raised, .api-header label {
             &.impure-pipe {
                 background-color: $brightred;
             }
+
+            &.operator {
+                background-color: $brightred;
```
jakovljevic-mladen, line 61: I reused the same color as in `impure-pipe` class - there are no impure pipes in RxJS 🙂

docs_app/tools/transforms/templates/api/base.template.html
```diff
@@ -26,6 +26,7 @@ <h1>{$ doc.name $}</h1>
     {% if doc.experimental !== undefined %}<label class="api-status-label experimental">experimental</label>{% endif %}
     {% if doc.stable !== undefined %}<label class="api-status-label stable">stable</label>{% endif %}
     {% if doc.pipeOptions.pure === 'false' %}<label class="api-status-label impure-pipe">impure</label>{% endif %}
```
jakovljevic-mladen, line 28: I think I will remove all Angular references like this one - there are so many of them that are so specific to Angular that they could be removed quite safely.

jakovljevic-mladen:
@cartant, could you please help one more time here by re-running the pipeline? Thanks

benlesh (approved):

benlesh:
This LGTM. I was sort of waiting for @niklas-wortmann's review though, since he owns the docs app.

niklas-wortmann:
LGTM, this is a great idea and I like it a lot!

PR #6564, opened 2021-08-14
docs(operators): fix code example of concatWith

diyews:
**Description:**
Add used operators to the `import` statement.


cartant (approved):
Thanks.

PR #6575, opened 2021-08-19
Change exhaust to exhaustAll

FraiZj:
Change exhaust to exhaustAll as first one is deprecated.
[https://rxjs.dev/guide/operators](url)
Screenshot 2021-08-20 121537


cartant (approved):
Thanks.

PR #6577, opened 2021-08-22
docs(operators): fix code example

diyews:
**Description:**

Going to fix code examples so that can run in stackblitz.

### DIFF
--- a/src/internal/operators/endWith.ts
+++ b/src/internal/operators/endWith.ts
@@ -45,9 +45,7 @@ export function endWith<T, A extends unknown[] = T[]>(...values: A): OperatorFun
  *   takeUntil(documentClicks$),
  *   endWith('interval ended by click'),
  * )
- * .subscribe(
- *   x => console.log(x)
- * )
+ * .subscribe(x => console.log(x));
  *
  * // Result (assuming a user clicks after 15 seconds)
  * // "interval started"
--- a/src/internal/operators/endWith.ts
+++ b/src/internal/operators/endWith.ts
@@ -46,7 +46,7 @@ export function endWith<T, A extends unknown[] = T[]>(...values: A): OperatorFun
  *   endWith('interval ended by click'),
  * )
  * .subscribe(
- *   x = console.log(x);
+ *   x => console.log(x)
  * )
  *
  * // Result (assuming a user clicks after 15 seconds)
--- a/src/internal/operators/filter.ts
+++ b/src/internal/operators/filter.ts
@@ -30,6 +30,10 @@ export function filter<T>(predicate: (value: T, index: number) => boolean): Mono
  * import { fromEvent } from 'rxjs';
  * import { filter } from 'rxjs/operators';
  *
+ * const div = document.createElement('div');
+ * div.style.cssText = `width: 200px;height: 200px;background: #09c;`;
+ * document.body.appendChild(div);
+ *
  * const clicks = fromEvent(document, 'click');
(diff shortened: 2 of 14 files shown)


src/internal/operators/endWith.ts
```diff
@@ -46,7 +46,7 @@ export function endWith<T, A extends unknown[] = T[]>(...values: A): OperatorFun
  *   endWith('interval ended by click'),
  * )
  * .subscribe(
- *   x = console.log(x);
+ *   x => console.log(x)
  * )
```
jakovljevic-mladen, line 50: `;` could be added here since almost all other examples use `;`. This file uses it as well. Maybe the whole `subscribe` call could be written in a single line...

diyews: `;` here is a syntax error. As you mentioned, it will be better to make it a single line, thx!

src/internal/operators/filter.ts
```diff
@@ -30,6 +30,10 @@ export function filter<T>(predicate: (value: T, index: number) => boolean): Mono
  * import { fromEvent } from 'rxjs';
  * import { filter } from 'rxjs/operators';
  *
+ * const div = document.createElement('div');
+ * div.style.cssText = `width: 200px;height: 200px;background: #09c;`;
+ * document.body.appendChild(div);
+ *
  * const clicks = fromEvent(document, 'click');
```
diyews, line 37: BTW, `find, findIndex, first` all need a div element to run the example. Or maybe we can refactor to use Typescript only, eg: 
```ts
range(1, 10)
  .pipe(filter(x => x % 2 === 1))
```

src/internal/operators/find.ts
```diff
@@ -40,6 +40,10 @@ export function find<T>(predicate: (value: T, index: number, source: Observable<
  * import { fromEvent } from 'rxjs';
  * import { find } from 'rxjs/operators';
  *
+ * const div = document.createElement('div');
+ * div.style.cssText = `width: 200px;height: 200px;background: #09c;`;
```
benlesh, line 44:
```suggestion
 * div.style.cssText = 'width: 200px; height: 200px; background: #09c;';
```

src/internal/operators/findIndex.ts
```diff
@@ -34,6 +34,10 @@ export function findIndex<T>(predicate: (value: T, index: number, source: Observ
  * import { fromEvent } from 'rxjs';
  * import { findIndex } from 'rxjs/operators';
  *
+ * const div = document.createElement('div');
+ * div.style.cssText = `width: 200px;height: 200px;background: #09c;`;
```
benlesh, line 38:
```suggestion
 * div.style.cssText = 'width: 200px; height: 200px; background: #09c;';
```

src/internal/operators/first.ts
```diff
@@ -53,6 +53,10 @@ export function first<T, D = T>(
  * import { fromEvent } from 'rxjs';
  * import { first } from 'rxjs/operators';
  *
+ * const div = document.createElement('div');
+ * div.style.cssText = `width: 200px;height: 200px;background: #09c;`;
```
benlesh, line 57:
```suggestion
 * div.style.cssText = 'width: 200px; height: 200px; background: #09c;';
```

benlesh (approved):
Thank you, @diyews!

diyews:
😢 My mistake, the `connect` example works but just missing some imports. Add a commit to rollback.
Just some little things I can do while learning the operators. No thanks.

PR #6590, opened 2021-09-12
Update pipe.ts

dscotese:
**Description:**
While examining the docs at https://rxjs.dev/api/index/function/pipe, I saw that `interval` has a useful description, but `pipe` does not, so I added one.  If it's inaccurate, please correct my errors.

### DIFF
--- a/src/internal/util/pipe.ts
+++ b/src/internal/util/pipe.ts
@@ -70,9 +70,10 @@ export function pipe<T, A, B, C, D, E, F, G, H, I>(
 ): UnaryFunction<T, unknown>;
 
 /**
- * pipe() can be called on one or more functions, each of which can take one argument ("UnaryFunction"), to get back
- * a function that passes the result of each subsequent UnaryFunction to the next one.  It's an implementation of
- * what is described at https://typeofnan.dev/how-to-use-pipe-the-pipeline-operator-in-javascript/
+ * pipe() can be called on one or more functions, each of which can take one argument ("UnaryFunction")
+ * and uses it to return a value.
+ * It returns a function that takes one argument, passes it to the first UnaryFunction, and then
+ * passes the result to the next one, passes that result to the next one, and so on.  
  */
 export function pipe(...fns: Array<UnaryFunction<any, any>>): UnaryFunction<any, any> {
   return pipeFromArray(fns);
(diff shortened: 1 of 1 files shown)


src/internal/util/pipe.ts
```diff
@@ -1,6 +1,11 @@
 import { identity } from './identity';
 import { UnaryFunction } from '../types';
 
+/**
+ * pipe() can be called on one or more functions, each of which can take one argument ("UnaryFunction"), to get back
+ * a function that passes the result of each subsequent UnaryFunction to the next one.  It's an implementation of
+ * what is described at https://typeofnan.dev/how-to-use-pipe-the-pipeline-operator-in-javascript/
+ */
```
benlesh, line 8: This is great! Really though, I think we need to move the doc comment to where it's above the implementation. Otherwise, I'm not sure the docs app will pick it up properly.

benlesh (changes requested):

dscotese:
Ahh the learning process.  I didn't realize what I was doing and ended up adding two commits instead of just the one.  The second commit is just pulling in other changes from master. I understand correctly, those changes have already been merged into master, so it won't have any effect.  If it does mess things up, please let me know what I should do to fix it.

src/internal/util/pipe.ts
```diff
@@ -69,6 +69,11 @@ export function pipe<T, A, B, C, D, E, F, G, H, I>(
   ...fns: UnaryFunction<any, any>[]
 ): UnaryFunction<T, unknown>;
 
+/**
+ * pipe() can be called on one or more functions, each of which can take one argument ("UnaryFunction"), to get back
+ * a function that passes the result of each subsequent UnaryFunction to the next one.  It's an implementation of
+ * what is described at https://typeofnan.dev/how-to-use-pipe-the-pipeline-operator-in-javascript/
```
benlesh, line 75: We really don't want to link to third party sites that are beyond our control. Can you please remove this link? 

I think a better add might be to show the primary use case for this utility with RxJS, which is to compose an operator from more than one other operator.

benlesh (changes requested):

src/internal/util/pipe.ts
dscotese: I removed the link and improved the description.  I agree that your suggestion is better than the external link, and if you feel strongly about showing the primary use case, it should be an example, rather than in the description.  In the description, it might be confusing in relation to the fundamental behavior of pipe().

PR #6592, opened 2021-09-14
test: 3 Failing tests to reproduce forkJoin bug (but ignored)

wokier:
**Description:**
Tests to reproduce a bug on forkJoin

**BREAKING CHANGE:**
no

**Related issue:**
https://github.com/ReactiveX/rxjs/issues/6593

### DIFF
--- a/spec/observables/forkJoin-spec.ts
+++ b/spec/observables/forkJoin-spec.ts
@@ -1,6 +1,22 @@
 /** @prettier */
 import { expect } from 'chai';
-import { finalize, forkJoin, map, of, timer } from 'rxjs';
+import {
+  combineAll,
+  concat,
+  concatMap,
+  finalize,
+  first,
+  forkJoin,
+  last,
+  map,
+  mergeAll,
+  Observable,
+  of,
+  pipe,
+  take,
+  timer,
+  zipAll,
+} from 'rxjs';
 import { lowerCaseO } from '../helpers/test-helper';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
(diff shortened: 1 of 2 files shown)


spec/observables/forkJoin-spec.ts
```diff
@@ -595,5 +611,164 @@ describe('forkJoin', () => {
...
       });
     });
+
+    it('concat reference', () => {
+      const results: Array<number | string> = [];
+      concat(of(1), of(2))
+        .pipe(last())
+        .subscribe({
+          next(value) {
+            results.push(+value);
```
benlesh, line 621: Interesting. Why do we need to convert `value` to a number here? It should already be a number?

spec/observables/forkJoin-spec.ts
```diff
@@ -595,5 +611,164 @@ describe('forkJoin', () => {
...
+    it('should work with concatMap cross types mergeAll', () => {
+      const results: Array<number | string> = [];
+      forkJoin([
+        of('a').pipe(concatMap((val: string) => of(1))) as Observable<number>,
+        of('b').pipe(concatMap((val: string) => of(2))) as Observable<number>,
+      ])
+        .pipe(mergeAll())
+        .subscribe({
+          next(value) {
+            results.push(+value);
```
benlesh, line 695: Here we're pushing `+[1, 2]` onto the array. Basically converting the array to `NaN`. So this is incorrect, and we'd have to push both values into the array, or stop converting it to a number, and push the whole array in, then change the expectation to be `[[1, 2], 'done']` below.

benlesh (commented):
Hi, @wokier! I **really** appreciate the effort here. You thought you saw a bug, and you submitted a PR with tests to prove the bug exists. So you _absolutely did the right thing_. Unfortunately, the tests in this PR have some issues that I've commented on.

I took a quick look in [a stackblitz](https://stackblitz.com/edit/rxjs-n6ygx4?file=index.ts), and it seemed everything was working okay. So I'm going to go ahead and close this PR.

Thank you so much again for the time and effort you put in, I appreciate it!

PR #6594, opened 2021-09-14
fix(Schedulers): Throwing a falsy error in a scheduled function no longer results in strange error objects.

benlesh:
Because TS 4.4 now infers `unknown` for `catch` blocks, we had to make some adjustments.

This also helped us catch a type problem where we were maybe creating `new Error(false)` or `new Error('')`, etc, in some cases.

### DIFF
--- a/spec/schedulers/QueueScheduler-spec.ts
+++ b/spec/schedulers/QueueScheduler-spec.ts
@@ -63,7 +63,7 @@ describe('Scheduler.queue', () => {
     const actions: Subscription[] = [];
     let action2Exec = false;
     let action3Exec = false;
-    let errorValue = undefined;
+    let errorValue: any = undefined;
     try {
       queue.schedule(() => {
         actions.push(
--- a/spec/util/UnsubscriptionError-spec.ts
+++ b/spec/util/UnsubscriptionError-spec.ts
@@ -16,10 +16,13 @@ describe('UnsubscriptionError', () => {
     try {
       subscription.unsubscribe();
     } catch (err) {
-      expect(err instanceof UnsubscriptionError).to.equal(true);
-      expect(err.errors).to.deep.equal([err1, err2]);
-      expect(err.name).to.equal('UnsubscriptionError');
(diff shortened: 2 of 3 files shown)


spec/schedulers/QueueScheduler-spec.ts
```diff
@@ -63,7 +63,7 @@ describe('Scheduler.queue', () => {
     const actions: Subscription[] = [];
     let action2Exec = false;
     let action3Exec = false;
-    let errorValue = undefined;
+    let errorValue: any = undefined;
```
benlesh, line 66: To fix CI ts@latest

spec/util/UnsubscriptionError-spec.ts
```diff
@@ -16,10 +16,13 @@ describe('UnsubscriptionError', () => {
     try {
       subscription.unsubscribe();
     } catch (err) {
-      expect(err instanceof UnsubscriptionError).to.equal(true);
-      expect(err.errors).to.deep.equal([err1, err2]);
-      expect(err.name).to.equal('UnsubscriptionError');
-      expect(err.stack).to.be.a('string');
+      if (err instanceof UnsubscriptionError) {
```
benlesh, line 19: To fix CI ts@latest

src/internal/scheduler/AsyncAction.ts
```diff
@@ -116,7 +116,10 @@ export class AsyncAction<T> extends Action<T> {
       this.work(state);
     } catch (e) {
       errored = true;
-      errorValue = (!!e && e) || new Error(e);
+      // HACK: Since code elsewhere is relying on the "truthiness" of the
+      // return here, we can't have it return "" or 0 or false.
+      // TODO: Clean this up when we refactor schedulers mid-version-8 or so.
+      errorValue = e ? e : new Error('Falsy action error');
```
benlesh, line 122: To fix CI ts@latest *and* correct a weird bit of code that TS 4.4 helped catch.

kwonoj (approved):

PR #6595, opened 2021-09-14
feat: Remove "super gross mode" once and for all! `useDeprecatedSynchronousErrorHandling` is removed

benlesh:
**Notice the target branch is 8.x**

BREAKING CHANGE: Removed `useDeprecatedSynchronousErrorHandling`. No workarounds for this.


src/internal/Subscriber.ts
```diff
@@ -193,13 +192,9 @@ function wrapForErrorHandling(handler: (arg?: any) => void, instance: SafeSubscr
...
     } catch (err) {
-      if (config.useDeprecatedSynchronousErrorHandling) {
-        captureError(err);
-      } else {
-        // Ideal path, we report this as an unhandled error,
-        // which is thrown on a new call stack.
-        reportUnhandledError(err);
-      }
+      // Ideal path, we report this as an unhandled error,
+      // which is thrown on a new call stack.
```
cartant, line 196: nitpick: There is only one path, now. 🎉

cartant (approved):

PR #6600, opened 2021-09-16
docs: fix glossary anchor links

mauromattos00:
**Description:**

As I was reading the docs, I found out that the anchor links in the [Glossary and Semantics page](https://rxjs.dev/guide/glossary-and-semantics) were not working.

After a bit of research, I realized that the anchor links were not in the correct way. So I have opened this PR to fix this problem 😁

## Before

antes

## After

depois


benlesh (approved):

benlesh:
Awesome thank you!

PR #6601, opened 2021-09-17
refactor: change teardown to finalizer/finalization in docs, private api

### LINKED ISSUE #6528: Change language from "teardown" to "finalizer"/"finalization"
Throughout our codebase and our documentation, we talk about "teardown", yet we have an operator called `finalize` and really, "teardown" isn't as descriptive in broader JavaScript-related terminology as "finalize" or "finalization". I think it would be good if we consolidated our language on this.

- [ ] update glossary
- [ ] update docs
- [ ] update variable/param names

sangxxh:
**Description:**

- Update glossary
- Update docs
- Update variable/param/type names and deprecate old type names (with jsdoc `@deprecated`) for backward compatibility

**Related issue (if exists):** Closes #6528


src/internal/types.ts
```diff
@@ -66,7 +66,11 @@ export interface Unsubscribable {
   unsubscribe(): void;
 }
 
-export type TeardownLogic = Subscription | Unsubscribable | (() => void) | void;
+export type FinalizerLogic = Subscription | Unsubscribable | (() => void) | void;
```
benlesh, line 69: The core team have to bike shed this one a bit. 

My gut tells me that `Finalizer` is enough. But I'd need to look at usage. It's really the observable's init return that this was created for.

That's one sort of problem with a big-bang PR like this.

src/internal/operators/catchError.ts
```diff
@@ -128,8 +128,8 @@ export function catchError<T, O extends ObservableInput<any>>(
 
     if (syncUnsub) {
       // We have a synchronous error, we need to make sure to
-      // teardown right away. This ensures that `finalize` is called
-      // at the right time, and that teardown occurs at the expected
+      // finalize right away. This ensures that `finalize` is called
```
benlesh, line 131:
```suggestion
      // finalize right away. This ensures that callbacks in the `finalize` operator are called
```

benlesh (changes requested):
I absolutely love the initiative here. I think we may need to dial back the _code_ changes and keep the documentation/comment changes for now. There's going to be a lot of churn when we move to version 8, and changing some of the names of types and protected properties/methods might not be worth doing until a later time.

I'll mark this as an agenda item for the next core team meeting, but I think that for now we should probably undo any changes to developer-facing APIs.

benlesh:
Core team meeting: For APIs, go ahead and deprecate and update names. However, we're not going to remove anything newly deprecated until version 9 at this point. To give people a chance to adapt.

Also: `FinalizerLogic` should just be `Finalizer`.

sangxxh:
Hi @benlesh, any update for this PR?

benlesh (approved):

benlesh:
@sangxxh Sorry for the delay. This looks good. Can you please resolve the conflicts so I can get this merged? Thanks.

PR #6603, opened 2021-09-19
chore: Skip test that sporadically fails

benlesh:
This test works sometimes and not other times. I stopped skipping it a while back and it seemed to be fixed. I think what's really happening is that it's _not_ fixed, rather the test scheduler is saving us sometimes by bailing out at a max time limit just before the call stack is exceeded... but that's a wild guess. Basically it's a bad test and we need to make a better test, then fix the issue seperately.

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -881,7 +881,7 @@ describe('Observable', () => {
   });
 
   // Discussion here: https://github.com/ReactiveX/rxjs/issues/5370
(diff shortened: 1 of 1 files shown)


cartant (approved):
Should we create an issue - or a roadmap entry - to remind us to write a better test?

kwonoj:
Created issue without looking this PR -  https://github.com/ReactiveX/rxjs/issues/6630

Issue #6321, opened 2021-05-02
ESM version is never used


akanass:
## Bug Report

**Current Behavior**
`ES5` version is always used in **client project** when we build it with `rollup` or `webpack` even if we set `mainFields` to `es2015` to take `package.json` attribute like in previous version.

`CJS` version is always used in **node project** even if we want to use module in latest node runtime.

This is due to the `exports` attribute in `package.json` which only declares `node` and `default` values.

**Expected behavior**
`ESM` version should be used when **client project** is built and we are using `rxjs` with `import` in `javascript` or `typescript` code and we want to build a script compatible with `HTML <script type="module"> in HTML`

`ESM` version should be used in **node project** if in `package.json`  we set `type:module` attribute and we used it in latest runtime.

**Reproduction**
In a client project,

```typescript
import { ajax, AjaxConfig, AjaxResponse } from 'rxjs/ajax';
import { Observable } from 'rxjs';
```

After that, you have to build the **client project** and you will see that only `ES5` version is taken and not `ESM` version even if we set `mainFields` to `es2015` in `webpack` or `rollup` config.

In a Node project,

```typescript
import { Observable } from 'rxjs';
```

Build your project and in `package.json` set `type:module` and you will see that only `CJS` version is taken.

**Environment**
- RxJS version: `7.0.0`

**Possible Solution**

You can see all possibilities [here](https://webpack.js.org/guides/package-exports/#reference-syntax) and we have to make a proposition to change `package.json:exports` attribute

You can see examples in the link provided before and they are explaining how to manage `node` and `module`  version.

If we don't change the `exports` attribute, we should allow build system to take the good version we want using `main`, `module` or `es2015` attributes from `package.json` like it was possible in previous version especially for client project.

benlesh:
I'm looking through the solutions here and none of them seem ideal: https://webpack.js.org/guides/package-exports/#target-environment-independent-packages

One thing I know comes up is situations where people are doing `instanceof` checks. The "stateless" solution for Node seems like it would cause issues. One gotcha is RxJS Observables are often used as a return type from third party packages. Some may use ESM, and other CJS, if that's the future there. 

In other news: God, I hate module systems.

benlesh:
Core Team Meeting: We probably can't act on this during version 7, because it would incur a breaking change for some folks. But we definitely want to move to modern ESM in version 8, probably ES2018+.

shrinktofit:
[...]

PR #6613, opened 2021-09-23
Add exports to package.json on 6.x branch

crisbeto:
This is along the same lines as #6192, but for the 6.x branch.

On Angular we're exploring changing our packaging to be based around ES modules and rxjs is currently a blocker, because it doesn't specify the `exports` field on the `package.json` in the 6.x branch like it does in 7.x. These changes add an identical configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,44 @@
   "description": "Reactive Extensions for modern JavaScript",
   "main": "index.js",
   "sideEffects": false,
+  "exports": {
+    ".": {
+      "node": "./index.js",
+      "es2015": "./_esm2015/index.js",
+      "default": "./_esm5/index.js"
+    },
+    "./ajax": {
+      "node": "./ajax/index.js",
+      "es2015": "./_esm2015/ajax/index.js",
+      "default": "./_esm5/ajax/index.js"
+    },
+    "./fetch": {
+      "node": "./fetch/index.js",
+      "es2015": "./_esm2015/fetch/index.js",
+      "default": "./_esm5/fetch/index.js"
+    },
+    "./operators": {
+      "node": "./operators/index.js",
+      "es2015": "./_esm2015/operators/index.js",
+      "default": "./_esm5/operators/index.js"
+    },
+    "./testing": {
+      "node": "./testing/index.js",
+      "es2015": "./_esm2015/testing/index.js",
+      "default": "./_esm5/testing/index.js"
+    },
+    "./webSocket": {
+      "node": "./webSocket/index.js",
+      "es2015": "./_esm2015/webSocket/index.js",
+      "default": "./_esm5/webSocket/index.js"
+    },
+    "./internal/*": {
+      "node": "./internal/*.js",
+      "es2015": "./_esm2015/internal/*.js",
+      "default": "./_esm5/internal/*.js"
+    },
+    "./package.json": "./package.json"
+  },
   "config": {
     "commitizen": {
       "path": "cz-conventional-changelog"


package.json
```diff
@@ -4,6 +4,37 @@
   "description": "Reactive Extensions for modern JavaScript",
   "main": "index.js",
   "sideEffects": false,
+  "exports": {
+    ".": {
+      "node": "./index.js",
+      "default": "./_esm5/index.js"
```
IgorMinar, line 10: can you please also include `es2015` version here and in all the other entry-points?
```suggestion
      "default": "./_esm5/index.js",
      "es2015": "./_esm2015/index.js"
```

IgorMinar (approved):
otherwise lgtm! thank you @crisbeto! 🚀 ❤️

package.json
IgorMinar: we should also forward-port this change to rxjs@7, @crisbeto could you please send a separate PR for that? thank you

IgorMinar: This change would fix https://github.com/ReactiveX/rxjs/issues/6321 as well.

crisbeto: Done. Will send a PR for the v7 branch ASAP.

crisbeto:
Feedback has been addressed.

benlesh (approved):

benlesh:
This is an apparent oversight on our part. I'm surprised it went this long without anyone reporting it (Unless we missed the report). Thanks, @crisbeto

benlesh:
The changes look okay, but I'm still double checking what/how this might break folks. It's not readily apparent to me that someone's build won't go belly-up if they're running `^6`, given that this is going to effect bundlers in ways I'm not quite familiar with.

benlesh:
NOTE: The ts@latest CI failure can be ignored. It's untyped try/catch blocks we still need to update in the 6.x branch.

package.json
crisbeto: https://github.com/ReactiveX/rxjs/pull/6614

benlesh:
After some discussion:

- @cartant pointed out that this change will probably break users of `rxjs-compat`, which 6.x needs to support.
- @trxcllnt noted that this will possibly break node users, because they may be requiring things in differently than bundled apps, and node won't allow you to require anything that isn't explicitly in the "exports" map. @kwonoj pointed out that we've defined `/internal/*` so that's probably not an issue.
- Several unanswered questions about the `es2015` conditional and who/what uses it.

@crisbeto @IgorMinar Can you address any of these concerns? 

IIRC, when this was added by @samccone, there was some discussion about whether or not it should go into the 6.x branch as well, and there was general agreement that it would constitute a breaking change so we opted to not do that. However I'm not able to find the conversation.  It might be that since we added the special handling for `internal/*`, it's not going to break anyone. There just seems to be some uncertainty about that right now. `rxjs-compat` adds a new angle on that I hadn't considered, and I'm glad that @cartant pointed it out.

As it stands, I think we'll want to wait until the next RxJS core team meeting so we can reach consensus on this. Right now it seems like the majority of RxJS core team members are leaning towards "unsure" or "no, because it's a breaking change". 

It may be that we just need to decide what amount of breakage is acceptable as well. (That's the sort of decision people make when they fix bugs for even patch releases, after all) If it's very low risk, it's probably okay to release this as a minor, IMO &mdash; but again, I think the team needs to reach consensus before we act

IgorMinar:
I've discussed this with the team, and @clydin is going to respond with more details, but we understand the hesitation and support you doing the right thing for everyone — whatever that ends up being.

clydin:
For usage with the Angular CLI, we are now able to apply a temporary workaround for rxjs 6 usage by internally enabling a Webpack option that disables ESM compliant behavior for a package or module.  While this allows Webpack-based bundling to succeed, it is not an ideal solution as it is Webpack specific and does not necessarily apply to other bundlers nor Node.js itself.  For the Angular CLI, accumulating package specific workarounds is also, unfortunately, not an ideal or sustainable long-term solution.  For these reasons, if possible, it would be preferred that the package was made ESM compliant via this or a similar PR.  However, we do understand there is a balance that must be achieved when considering the potential for breaking changes.

In regards to mitigating breaking changes, there is also the option to add support for full deep importing of the package via a final subpath pattern (`"./*": “./*”`). This effectively disables the encapsulation normally provided by the `exports` field.

benlesh:
This seems unnecessary at this point, so I'm going to close it.

PR #6620, opened 2021-09-29
fix: fix types file path cannot be resolved

zthxxx:
will cause an error with [`eslint-import-resolver-typescript`](https://github.com/alexgorbatchev/eslint-import-resolver-typescript/blob/v2.5.0/src/index.ts#L125), due to missing types files.

image

image


kwonoj:
Dupe of https://github.com/ReactiveX/rxjs/pull/6293 .

zthxxx:
@kwonoj  why not just remove typesVersions field? It's not supported ts version < 4.2 at all.

Issue #6466, opened 2021-06-10
"types" in package.json does not resolve to a file by default


orta:
## Bug Report

[7.1.0 Package JSON](https://www.unpkg.com/rxjs@7.1.0/package.json)

```
  "types": "index.d.ts",
  "typesVersions": {
    ">=4.2": {
      "*": [
        "dist/types/*"
      ]
    }
  },

```

Goes to [`index.d.ts`](https://www.unpkg.com/rxjs@7.1.0/index.d.ts) which doesn't exist.

**Current Behavior**
I haven't implemented the `typesVersion` resolving in the playground, and so it NOOPs ( https://github.com/microsoft/TypeScript-Website/issues/1316 ). Perhaps if you plan on supporting TS 4.2 and below you could use [downlevel-dts](https://github.com/sandersn/downlevel-dts).

cartant:
The value used for the `types` property is what was recommended [here](https://github.com/microsoft/TypeScript/issues/43034#issuecomment-823423227). Also, there are no plans to support TypeScript versions below 4.2, as the types were massively refactored using recent TypeScript features.

orta:
Ouch TIL'd - I guess the playground would need to support that resolution path too 👍🏻

PR #6621, opened 2021-09-30
fix: fix types file path cannot be resolved, remove typesVersions field

zthxxx:
Will cause an error with [`eslint-import-resolver-typescript`](https://github.com/alexgorbatchev/eslint-import-resolver-typescript/blob/v2.5.0/src/index.ts#L125), due to missing types files.

And make the field `typesVersions` is meaningless, because it doesn't support ts version < 4.2 at rxjs@7. 

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,14 +5,7 @@
   "main": "./dist/cjs/index.js",
   "module": "./dist/esm5/index.js",
   "es2015": "./dist/esm/index.js",
-  "types": "index.d.ts",
-  "typesVersions": {
-    ">=4.2": {
-      "*": [
-        "dist/types/*"
-      ]
-    }
-  },
+  "types": "./dist/types/index.d.ts",
   "sideEffects": false,
   "exports": {
     ".": {


zthxxx:
@kwonoj

cartant:
https://github.com/ReactiveX/rxjs/issues/6466#issuecomment-859186944

cartant:
I'm closing this because the resolution implementation to which you've linked in `eslint-import-resolver-typescript` is incorrect - see [this explanation](https://github.com/microsoft/TypeScript/issues/43034#issuecomment-823423227) from a TypeScript contributor. As outlined in the `package.json`, RxJS v7 requires TypeScript 4.2 or greater. The incorrect resolution in `eslint-import-resolver-typescript` is not something for which RxJS is responsible.

zthxxx:
@cartant  May I ask why need `typesVersions` field, and what's the use of `"dist/types/*"`? What's helpful of it rather than `"types": "./dist/types/index.d.ts"`?

cartant:
Two reasons:

- it's a mechanism through which it can be made clear that 4.2 is the minimum-supported version of TypeScript for RxJS; and
- it's a mechanism that we could use to update the types - without breaking people - to take advantage of any useful features offered by newer versions of TypeScript. That is, we could add an additional entry in `typesVersions`. In v7, the types were simplified quite a bit, but there are still some that are complicated and I like to hope that future versions of TS will offer features that could be used to further simplify them.

PR #6622, opened 2021-09-30
build(test): run mocha directly without separate transpilation

kwonoj:
We've been running mocha with separate transpilation step via tsc as we've been experiencing transpilation performance via ts-node (or other mocha's transpiler registration) was not good enough. The current approach is generally fine for the perf, but as it requires 2 stage it makes bit tricky to setup things by pointing /dist, also some tooling side support like watch mode is not cleanly working.

This PR goes back to transpilation registration via mocha directly. But the difference is now this uses swc as transpilation backend instead of typescript itself. I've been trying to use swc for few places and so far our limited usecase (test) it seems working well sufficiently. If not, we can make upstream fix as well. 

You may notice this uses `ts-node` still instead of directly using `@swc/register`. I choose this to let ts-node interops our tsconfig into swc instead of we maintain another swcrc config.

Please give it a go local run for the performance. So far on my local machine this works quite well.

**Description:**

**BREAKING CHANGE:** 


kwonoj:
Note this is only for the test, nothing changes regarding actual build output: we'll always use tsc for the actual compilation since we need type check anyway.

kwonoj:
Interestingly one test case fails on CI, while locally passes. Will leave it for now while discussing if we want to this change or not - if we want to go, will try to debug further.

benlesh (approved):

kwonoj:
I think so, this is an improvement to test. Rebased.

PR #6632, opened 2021-10-06
Add clearBuffer method to ReplaySubject

crossan007:
**Description:**
Adds a `clearBuffer()` method on `ReplaySubject`.  This allows consumers to clear the buffer of events which have already been delivered to the `ReplaySubject`.

This addresses an edge case where a developer may want to avoid disrupting existing subscriptions to a `ReplaySubject` and prevent new subscribers from receiving the whole history before a point in time (while maintaining the behavior which facilitates new subscribers to receive history)

This _should be_ a preferred solution over the various proposed workarounds at:

- https://stackoverflow.com/questions/58812252/angular-8-replaysubject-clear-delete-history-and-cache
- https://stackoverflow.com/questions/51145664/resetting-replaysubject-in-rxjs-6

An alternative workaround may be to extend `ReplaySubject` as `ClearableReplaySubject` and adjust the visibility modifier on the `events[]` from `private` to `protected` so that this method can be implemented on the child class

**BREAKING CHANGE:** 

### DIFF
--- a/src/internal/ReplaySubject.ts
+++ b/src/internal/ReplaySubject.ts
@@ -65,6 +65,12 @@ export class ReplaySubject<T> extends Subject<T> {
     super.next(value);
   }
 
+  public clearBuffer() {
+    // clears the event buffer so that new subscribers do not receive events which have been 
+    // added to the buffer before this method was called.
+    this._events = [];
+  }
+  
   /** @internal */
   protected _subscribe(subscriber: Subscriber<T>): Subscription {
     this._throwIfClosed();


benlesh:
I'll leave it to the core team, and we'll discuss it, but In general, I'm inclined to say that we're probably not going to implement this feature for a few reasons:

1. It introduces another vector for mutability to `ReplaySubject`.
2. It has no tests
3. The use cases aren't clearly defined.

As a workaround, someone could take a more reactive approach, particularly with the new `share` features:

```ts
const shared = source$.pipe(
  // If some stream of resets notifies, error with a known error (like "RESET")
  mergeWith(reset$.pipe(tap(() => { throw 'RESET' })),
  // multicast through a ReplaySubject. `share` will reset this
  // ReplaySubject on error by default.
  share({
    connector: () => new ReplaySubject(),
  }),
  retry({
    // If the error is the known error of "RESET", then reconnect everyone
    // if it's any other error, just throw it.
    delay: (err) => {
       if (err === 'RESET') {
         return of(true);
       } else {
         throw error;
       }
    }
  })
)
```

benlesh:
Core Team: We don't want to move forward with this one. 

Thank you for your effort, @crossan007

crossan007:
Okay, thanks for the feedback.

I'm using this implementation currently to consume AngularFirestore observables in various locations in my app, in a way that I can clear the replay buffer when AngularFireauth context changes

douglascayers:
In this helps anyone help, I've put @benlesh's [example](https://github.com/ReactiveX/rxjs/pull/6632#issuecomment-1010057068) into a `ResettableReplaySubject` class.

Source code with example available at https://stackblitz.com/edit/rxjs-x4n1rt?file=index.ts

PR #6633, opened 2021-10-07
fix: takeWhile Boolean constructor types

josepot:
I think that the typings for the Boolean constructor of `takeWhile` are off... For instance, I think that the following dtslint test:

```ts
const c = of(
  false as const,
  0 as const,
  "hi" as const,
  -0 as const,
  0n as const,
  "" as const,
  null,
  undefined
).pipe(takeWhile(Boolean)); // $ExpectType Observable<false | "" | 0 | 0n | "hi" | null | undefined>
```

should expect this, instead:

```ts
const c = of(
  false as const,
  0 as const,
  "hi" as const,
  -0 as const,
  0n as const,
  "" as const,
  null,
  undefined
).pipe(takeWhile(Boolean)); // $ExpectType Observable<"hi">
```

The goal of this PR is to make the Boolean constructor typings of `takeWhile` (when the `inclusive` flag is set to `false`) consistent with the typings of `filter`.

### DIFF
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -707,9 +707,9 @@ export declare function takeLast<T>(count: number): MonoTypeOperatorFunction<T>;
 
 export declare function takeUntil<T>(notifier: ObservableInput<any>): MonoTypeOperatorFunction<T>;
 
-export declare function takeWhile<T>(predicate: BooleanConstructor): OperatorFunction<T, Exclude<T, Falsy> extends never ? never : T>;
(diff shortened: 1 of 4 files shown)


josepot:
> Not sure if I understood correctly, but looks like current signature behaves correct? Takewhile doesn't change what source emits but only limits when to stop emit. Would you mind explain bit as I may miss something?

If the `inclusive` flag is set to `false` (which it's its default value) then when used with `Boolean` then the resulting observable won't emit falsy values. For instance:

```ts
new Observable<string | null>().pipe(
  takeWhile(Boolean)
)
```

will only emit `strings`, it would never emit `null`.

yunuseon (approved):

benlesh:
Thank you, @josepot!

PR #6635, opened 2021-10-08
chore: fix docs build

jakovljevic-mladen:
**Description:**
~For some reason, `remark-html` with version 13 doesn't generate code examples and some guide files correctly. I tried increasing `remark` version to 12, 13 and 15, but it didn't work for multiple reasons, so I had to lower the `remark-html` from version 13 to version 12.~

For example, this is how a code example looked like (without fix from this PR):

image

And this is the blank page from `/api` route:
image

This PR fixes these issues.

Edit: please check [updated description](#issuecomment-939751362).

None


jakovljevic-mladen:
I managed to upgrade `remark` version to v12 and to keep `remark-html` at v13. To do this, I had to copy some stuff from Angular.

Related: angular/angular#24000 and angular/angular#41721

Issue #6641, opened 2021-10-15
Incompatible with `node12` TypesScript module resolution


demurgos:
## Bug Report

**Current Behavior**
I am trying out the upcoming TypeScript 4.5.0 version (currently in beta), and its new `"moduleResolution": "node12"`.

With the following `tsconfig.json`:
```json
{
  "compilerOptions": {
    "module": "node12",
    "moduleResolution": "node12",
    "target": "es2021",
    "noImplicitAny": true
  }
}
```

I get the following error when trying to import from `"rxjs"`:

```
main.mts:1:28 - error TS7016: Could not find a declaration file for module 'rxjs'. '/data/projects/various/rxjs-ts-node12/node_modules/rxjs/dist/cjs/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/rxjs` if it exists or add a new declaration (.d.ts) file containing `declare module 'rxjs';`

1 import { Observable } from "rxjs";
                             ~~~~~~


Found 1 error.
```

**Expected behavior**

The type declaration for `rxjs` should be resolved.

**Reproduction**

The `tsconfig.json` should be enough to reproduce the issue. Here is a repo set-up to reproduce the issue: <https://github.com/demurgos/rxjs-ts-node12>.

**Environment**
- Runtime: Node 12+ (with ESM support)
- RxJS version: 7.4.0
- (If bug is related) Loader, build configuration: N/A

**Possible Solution**

Add a `types` field above `node` to the `exports` config in the `package.json`.

**Additional context/Screenshots**
- [TypeScript PR to enable `node12`](https://github.com/microsoft/TypeScript/pull/45884)
- [TS 4.5.0 beta announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/)

ievgennaida:
Same problem when     "module": "esnext" is used.
RxJS version: 7.4.0
Typescript 4.4.4

kwonoj:
I think we may need few internal try to understand some of behaviors.

One of notable thing is rxjs does not support native esm yet. I'm not sure what's implication if we add this property for both of cases, if tsc try to import as native esm vs. other tooling's esm handling behavior such as tree shaking.

kwonoj:
Peeking this bit more, there is specific concern from TSC's core team iteslf https://github.com/microsoft/TypeScript/issues/46452. 

[...]

PR #6642, opened 2021-10-15
fix: Add types to the package export map

demurgos:
**Description:**

This commit adds `types` mapping for the package `exports`. It makes the package compatible with the upcoming `node12` module resolution in TypeScript 4.5.0.

**BREAKING CHANGE:** N/A

**Related issue (if exists):** #6641

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,36 +16,43 @@
   "sideEffects": false,
   "exports": {
     ".": {
+      "types": "./dist/types/index.d.ts",
       "node": "./dist/cjs/index.js",
       "es2015": "./dist/esm/index.js",
       "default": "./dist/esm5/index.js"
     },
     "./ajax": {
+      "types": "./dist/types/ajax/index.d.ts",
       "node": "./dist/cjs/ajax/index.js",
       "es2015": "./dist/esm/ajax/index.js",
       "default": "./dist/esm5/ajax/index.js"
     },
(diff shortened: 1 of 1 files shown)


kwonoj:
Ref original issue's comment https://github.com/ReactiveX/rxjs/issues/6641#issuecomment-955063190 .

kwonoj:
I'm going to close this for now based on Typescript Team's decision (https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-rc/#beta-delta) to not include this for the release:

> If you try to use these settings in TypeScript 4.5 RC or stable, you’ll receive an error message directing you to use a nightly build instead

We'll eventually try to support latest tsc features as much to align with ecosystem, but given upstream tsc would like to observe things more carefully and it's not availble for general purpose I don't think we can release our package with this at this moment.

frank-dspeed:
@kwonoj reopen and release this was wrong communicated it did ship it is released

kwonoj:
Isn't this same as https://github.com/ReactiveX/rxjs/pull/6802 filed recently?

frank-dspeed:
@kwonoj your correct and andrew is from the typescript team ok i got confused because the person who created this pull request posted that he did a pull request before and then i reviewed this.

but andrews pull is newer so ok your correct merge #6802 LGTM +1

PR #6652, opened 2021-10-24
Feature: Add getValue() in ReplaySubject to get the last emitted value

sagrawal31:
*I'm a first-time contributor (not a consumer) to RxJS. Pardon my lack of understanding.*

**Description:**

Just like `BehaviorSubject`, this PR will add a method `getValue()` in `ReplaySubject` so that the owner can always get the last emitted value (if not, returns `null`). This change makes the `ReplaySubject` usable like a state (maybe this can be a different *Subject class itself).

Once this MR is approved, I'll update the documentation.

### DIFF
--- a/src/internal/ReplaySubject.ts
+++ b/src/internal/ReplaySubject.ts
@@ -37,6 +37,7 @@ import { dateTimestampProvider } from './scheduler/dateTimestampProvider';
 export class ReplaySubject<T> extends Subject<T> {
   private _buffer: (T | number)[] = [];
   private _infiniteTimeWindow = true;
+  private _lastValue: T | null = null;
 
   /**
    * @param bufferSize The size of the buffer to replay on subscription
@@ -55,6 +56,14 @@ export class ReplaySubject<T> extends Subject<T> {
     this._windowTime = Math.max(1, _windowTime);
   }
 
+  get value(): T | null {
+    return this.getValue();
+  }
+
+  getValue(): T | null {
+    return this._lastValue;
+  }
+
   next(value: T): void {
     const { isStopped, _buffer, _infiniteTimeWindow, _timestampProvider, _windowTime } = this;
     if (!isStopped) {
@@ -62,7 +71,7 @@ export class ReplaySubject<T> extends Subject<T> {
       !_infiniteTimeWindow && _buffer.push(_timestampProvider.now() + _windowTime);
     }
     this._trimBuffer();
-    super.next(value);
+    super.next((this._lastValue = value));
   }
 
   /** @internal */


yunuseon:
Hey @sagrawal31 ,
unfortunately a getValue() Method would only make sense for a ReplaySubject with the buffer of 1.
ReplaySubjects with a buffer size of 1 are very similiar to BehaviorSubject, from the docs: 

https://rxjs.dev/api/index/class/ReplaySubject
> BehaviorSubject is similar to new ReplaySubject(1), with a couple fo exceptions:
> 1. BehaviorSubject comes "primed" with a single value upon construction.
> 2. ReplaySubject will replay values, even after observing an error, where BehaviorSubject will not.

sagrawal31:
Thanks for the response @yunuseon. While I understand your point, the `getValue()` is returning the last/latest value only, isn't it?

benlesh:
@sagrawal31 the simplest thing to do in this case would be to just subscribe to the `ReplaySubject` to update some local variable you can access.

```ts
let hasValue = false;
let value: T | null = null;

replaySubject.subscribe(v => {
  hasValue = true;
  value = v;
});
```

On of the problems with you proposed solution is what if the `ReplaySubject` is a `ReplaySubject<null | number>`? (meaning it might get `null --- 1 --- 2 --- null --- null --- 3` etc.) How would you know it received a value yet? That's not a problem with a `BehaviorSubject`.

In the end, there are LOTS of ways to solve this with existing functionality: A `tap` to update some local variables, then `share` through a `ReplaySubject` `connector`. Just subscribe with a `BehaviorSubject` to get a stream of the most recent values, etc. etc. 

I would just think about my use case and do the _simplest_ thing. (most likely the example above)

sagrawal31:
Thanks, @benlesh. For now, I have been using it by extending ReplaySubject like this-

```typescript
import {Observer, ReplaySubject, Subscription} from 'rxjs';

/**
 * A wrapper over {@link ReplaySubject} to also provide the last value. This basically kind of mix the
 * BehaviorSubject to {@link ReplaySubject};
 *
 * @see https://github.com/ReactiveX/rxjs/pull/6652
 * @author Shashank Agrawal
 */
export class StateSubject<T> extends ReplaySubject<T> {

    private lastValue: T;

    /**
     * Returns the last emitted value.
     */
    getValue(): T {
        return this.lastValue;
    }

    /**
     * Emits a new value.
     *
     * @see https://stackoverflow.com/a/69696916/2405040
     * @param value
     */
    next(value?: T): void {
        this.lastValue = value;
        super.next(value);
    }

    // @ts-ignore
    subscribe(observer?: Partial<Observer<T>>): Subscription {
        return super.subscribe(observer);
    }

}

```

benlesh:
That seems okay.. but as a general rule, I'd advise against inheriting from classes you don't own. Either way, that seems unlikely to break in the future.

Issue #6672, opened 2021-11-11
AsapScheduler race condition w/ nested schedules (Cannot read properties of undefined (reading 'execute'))


matthewjh:
## Bug Report

**Current Behavior**
On our non-trivial, non-public UI codebase that heavily uses rx.js, we sporadically see the following error arising from code that uses the `AsapScheduler`:

TypeError: Cannot read properties of undefined (reading 'execute')
            at AsapScheduler.flush (node_modules/rxjs/src/internal/scheduler/AsapScheduler.ts:19:8

This is thrown from https://github.com/ReactiveX/rxjs/blob/1f7b9c5b3339d7dbb9303e2e436c62f40ad2ff47/src/internal/scheduler/AsapScheduler.ts#L16

I have debugged this in-situ and have arrived at what is, I think, a working understanding how the exception arises. I believe it to be a problem with nested `AsapTask` schedules.

Note there have been reports of similar issues in #4690 and #2697, which were closed without resolution. I think it likely all these share this cause.

1. Let's say an `AsapAction`, A, is scheduled. The scheduler's actions are `[A]`. When A is executed, it will schedule another `AsapAction`, B.
2. When the scheduler asynchronously flushes, it first sets `this._scheduled = undefined;`
https://github.com/ReactiveX/rxjs/blob/master/src/internal/scheduler/AsapScheduler.ts#L5
3. Then it iterates through the actions and executes them. As A is the only action, it will call `action.execute()` on A. It `shifts` A off so the actions queue is now `[]`.
(Here, `index` is initially -1 and count is the length of the queue on entering `flush`, which is 1.)
```ts
 do {
      if ((error = action.execute(action.state, action.delay))) {
        break;
      }
    } while (++index < count && (action = actions.shift()));
```
4. In the execution of A, A schedules `AsapAction` B (as mentioned). The action queue is now `[B]`. As `asapScheduler._scheduled` was set to `undefined` on entering `AsapScheduler#flush`, the following code will create a new async task that will flush the scheduler:

https://github.com/ReactiveX/rxjs/blob/master/src/internal/scheduler/AsapAction.ts#L21

5. On returning from `action.execute` on A, the while expression will be evaluated. Bear in mind that we mutated `actions` in (4). The first part is true: `(++index < count) === (++(-1) < 1) === (0 < 1) === true`
The second part, `action = actions.shift()` will pop B from the actions queue and set the current action to B.
6. The next turn of the loop will call `action.execute()` on B.
[...]

Issue #4690, opened 2019-04-05
Scheduler.flush throws unclear error when actions is empty (cannot read property 'execute' of undefined)


notnarb:
## Bug Report

**Current Behavior**
Calling .flush() on a scheduler (e.g. animationFrameScheduler) with no pending actions causes an unclear exception to be thrown:
```
TypeError: Cannot read property 'execute' of undefined
at AnimationFrameScheduler.flush (https://rxjs-pc1ga8.stackblitz.io/turbo_modules/rxjs@6.4.0/internal/scheduler/AnimationFrameScheduler.js:31:32)
at Object.eval (https://rxjs-pc1ga8.stackblitz.io/~/index.ts:12:36)
at eval (https://rxjs-pc1ga8.stackblitz.io/~/index.ts:19:4)
at eval (https://rxjs-pc1ga8.stackblitz.io/~/index.ts:20:3)
at eval (<anonymous>)
at Qt (https://c.staticblitz.com/d/webcontainer.e1e926a8972546110c3.js:15:30145)
at https://c.staticblitz.com/d/webcontainer.e1e926a8972546110c3.js:15:38799
at U (https://c.staticblitz.com/d/webcontainer.e1e926a8972546110c3.js:15:13565)
at https://c.staticblitz.com/d/webcontainer.e1e926a8972546110c3.js:15:13207
at boot (https://rxjs-pc1ga8.stackblitz.io/dev/boot:6:3)
```

**Reproduction**
- REPL or Repo link: https://stackblitz.com/edit/rxjs-pc1ga8?devtoolsheight=60

```ts
import {animationFrameScheduler} from 'rxjs';
animationFrameScheduler.flush();
```

**Expected behavior**
Either one of:
 * not throwing when there are no pending actions
 * throwing a clearer error (e.g. 'Attempting to flush an empty queue')

**Environment**
- Runtime: Chrome v73
- RxJS version: 6.4.x

**Possible Solution**

See expected behavior. This appears to affect all schedulers that have this `do while` loop without verifying that a valid action has been `shift`'d off of `this.actions`

This may be working as intended (and no change desired for performance reasons) in which case just having this issue open and google-able would provide additional documentation

**Additional context/Screenshots**

#2697 Appears to have been the same issue given their stacktrace: https://github.com/ReactiveX/rxjs/blob/5.4.1/src/scheduler/AsapScheduler.ts#L17

A clearer error (or no error at all) would probably have pointed them in the right direction.

---

[...]

Issue #2697, opened 2017-06-26
TypeError: undefined is not an object (evaluating 'e.execute') in AsapScheduler.flush


crooy:
**RxJS version:**
5.4.1

**Actual behavior:**

```
TypeError: undefined is not an object (evaluating 'e.execute')
flush — AsapScheduler.ts:17
(anonymous function)
exports — _invoke.js:5
(anonymous function) — _task.js:31
_ — _task.js:19
y — _task.js:23
```

**Additional information:**
The error occurs in our cordova ReactJS + RxJs app, at the moment our UI switches from not logged in to logged in. I cannot find what code produces this error sadly. I suspect its somehow related to unsubscribing various streams. We use lots of streams that are mapped, filtered, published, refcounted, scanned etc.

This error does not create any issues for the user as far as i can find.

kwonoj:
Unfortunately it isn't possible to tell if this is occurring in library's behavior or user code with given information. If there's reproducible code snippet it'll be helpful.

kwonoj:
I'm closing this issue as inactive / lacks of information. Feel freely reopen if there is reproducible code snippet can be shared.

PR #6674, opened 2021-11-13
fix: scheduler unsubscription scenarios

cartant:
**Description:**

This PR adds failing tests and fixes for the following scenarios:

- In the `AnimationFrameScheduler` and the `AsapScheduler`, it was possible for actions scheduled within a `flush` call - i.e. scheduled by actions executed with said `flush` call - to be themselves executed within the _same_ `flush` call. This would only occur if actions were unsubscribed, as that would 'move' the newly-scheduled actions to a lower index in the `actions` queue - rendering the `count`-based mechanism unreliable for determining which actions should be executed.
- In the `AnimationFrameAction` and the `AsapAction`, the cancellation of the animation frame and the clearing of the microtask depended upon whether or not the `actions` queue was empty. However, that's insufficient, as the `actions` queue can contain actions that are to be executed in separate `flush` calls. To determine whether cancellation is required, it's necessary to check that the `actions` queue has _no actions_ that share the _same_ async id.

Note that if the animation frame or microtask mentioned in the second scenario is not cancelled, `flush` is called with an empty `actions` queue and an error is effected:

```
TypeError: Cannot read properties of undefined (reading 'execute')
    at AsapScheduler.flush (node_modules/rxjs/src/internal/scheduler/AsapScheduler.ts:19:8
```

Version 6 is going to have the same problem - this has been a long-standing issue that has been hard to reproduce - so we should open another PR after this is merged to cherry-pick the change into that version.

**Related issue:** #6672 and #4690 and #2697

### DIFF
--- a/spec/schedulers/AsapScheduler-spec.ts
+++ b/spec/schedulers/AsapScheduler-spec.ts
@@ -218,4 +218,50 @@ describe('Scheduler.asap', () => {
     }, 0, 0);
     scheduledIndices.push(0);
   });
+
+  it('should execute actions scheduled when flushing in a subsequent flush', (done) => {
+    const sandbox = sinon.createSandbox();
+    const stubFlush = (sandbox.stub(asapScheduler, 'flush')).callThrough();
+
+    let a: Subscription;
+    let b: Subscription;
+    let c: Subscription;
+
+    a = asapScheduler.schedule(() => {
+      expect(stubFlush).to.have.callCount(1);
+      c = asapScheduler.schedule(() => {
+        expect(stubFlush).to.have.callCount(2);
+        sandbox.restore();
+        done();
+      });
+    });
+    b = asapScheduler.schedule(() => {
+      expect(stubFlush).to.have.callCount(1);
+    });
+  });
+
+  it('should execute actions scheduled when flushing in a subsequent flush when some actions are unsubscribed', (done) => {
+    const sandbox = sinon.createSandbox();
+    const stubFlush = (sandbox.stub(asapScheduler, 'flush')).callThrough();
+
+    let a: Subscription;
+    let b: Subscription;
+    let c: Subscription;
+
+    a = asapScheduler.schedule(() => {
+      expect(stubFlush).to.have.callCount(1);
+      c = asapScheduler.schedule(() => {
+        expect(stubFlush).to.have.callCount(2);
+        sandbox.restore();
+        done();
+      });
+      b.unsubscribe();
+    });
+    b = asapScheduler.schedule(() => {
+      expect(stubFlush).to.have.callCount(1);
+    });
+  });
+
+  it.skip('should properly cancel an unnecessary flush', (done) => {
+  });
 });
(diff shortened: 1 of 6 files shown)


spec/schedulers/AsapScheduler-spec.ts
```diff
@@ -39,9 +39,10 @@ describe('Scheduler.asap', () => {
 
   it('should cancel asap actions when delay > 0', () => {
     testScheduler.run(({ cold, expectObservable, flush, time }) => {
-      const setImmediateSpy = sinon.spy(immediateProvider, 'setImmediate');
-      const setSpy = sinon.spy(intervalProvider, 'setInterval');
-      const clearSpy = sinon.spy(intervalProvider, 'clearInterval');
+      const sandbox = sinon.createSandbox();
+      const setImmediateSpy = sandbox.spy(immediateProvider, 'setImmediate');
+      const setSpy = sandbox.spy(intervalProvider, 'setInterval');
+      const clearSpy = sandbox.spy(intervalProvider, 'clearInterval');
```
cartant, line 45: `sandbox` wasn't used consistently throughout this file, so I added a commit - https://github.com/ReactiveX/rxjs/pull/6674/commits/0dd8f02d9d79d6e2f4a9dabc9c712168f05a2d10 - to address this.

matthewjh:
Thanks @cartant

src/internal/scheduler/AsapScheduler.ts
```diff
@@ -4,24 +4,32 @@ import { AsyncScheduler } from './AsyncScheduler';
...
-    let index = -1;
     action = action || actions.shift()!;
-    const count = actions.length;
 
     do {
       if ((error = action.execute(action.state, action.delay))) {
         break;
       }
-    } while (++index < count && (action = actions.shift()));
+    } while ((action = actions[0]) && action.id === flushId && actions.shift());
```
benlesh, line 27: It really feels like we should clean up this loop, but for now I think this change is solid.

cartant: No arguments from me re: clean up.

benlesh (approved):
Nice work. I can't wait until we have an opportunity to really rework the schedulers in the future. Some of what we have going on in here is a bit tangled.

benlesh:
@cartant FYI: This isn't trivial to cherry-pick into 6.x, mostly because the tests don't work. The actual runtime code port is pretty straight forward.

benlesh:
@cartant here is my attempt at cherry-picking the commit over: https://github.com/ReactiveX/rxjs/pull/6693/files

I don't have strong feelings about this fix making 6.x or not, given it's not really been a huge issue until now, but it might be nice to land it.  I don't really want to waste a lot of time refactoring our testing strategies for the 6.x branch, but that's where all the sticking points are right now.

cartant:
I'll take a look at the cherry-pick PR. The main reason for suggesting that it be cherry picked is that this issue has been occurring since version 5 - #2697 and #4690 - but no one had offered a reproduction of the problem until recently.

PR #6690, opened 2021-11-29
fix: race condition with Interop Observables

josepot:
**Description:**

I'm working on a library that exposes [Interop Observables](https://ncjamieson.com/how-to-use-interop-observables/) and I've noticed that in some situations, depending on which package gets imported first, the user gets the following error:

```bash
TypeError: You provided an invalid object where a stream was expected. You can provide an Observable, Promise, ReadableStream, Array, AsyncIterable, or Iterable.
```

After some debugging I found that the culprit is this line of code:

```ts
export const observable: string | symbol = (() => (typeof Symbol === 'function' && Symbol.observable) || '@@observable')();
```

The problem with that code is that if it runs before the package(s) that are responsible for polyfilling `Symbol.observable` get evaluated, then it's already too late to polyfill `Symbol.observable`.

A simple solution would be to make that code lazy, so that if later on a package polyfills `Symbol.observable` then everything should work as it's expected:

```ts
export const observable: () => string | symbol = (() => typeof Symbol === 'function' && Symbol.observable) || '@@observable';
```

Which is the change that I'm proposing in this PR.

### DIFF
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -415,7 +415,7 @@ export interface ObjectUnsubscribedError extends Error {
 
 export declare const ObjectUnsubscribedError: ObjectUnsubscribedErrorCtor;
 
-export declare const observable: string | symbol;
+export declare const observable: () => string | symbol;
 
 export declare class Observable<T> implements Subscribable<T> {
     operator: Operator<any, T> | undefined;
--- a/spec/helpers/interop-helper-spec.ts
+++ b/spec/helpers/interop-helper-spec.ts
@@ -7,7 +7,7 @@ describe('interop helper', () => {
   it('should simulate interop observables', () => {
     const observable: any = asInteropObservable(of(42));
     expect(observable).to.not.be.instanceOf(Observable);
-    expect(observable[symbolObservable]).to.be.a('function');
+    expect(observable[symbolObservable()]).to.be.a('function');
   });
 
   it('should simulate interop subscribers', () => {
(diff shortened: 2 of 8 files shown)


josepot:
Oh, wow! I just noticed that this `observable` is part of the public API :scream: So, this would be a breaking change... :worried: 

I think that this must be a mistake, IMO it only makes sense to expose a non lazy version of it if RxJS itself polyfills `Symbol.observable` like it used to do it in the past... Otherwise, it should be a function, right?

benlesh:
In general, polyfills should always happen _first_. So the issue here is if someone wants `Symbol.observable` interop, they'd need to polyfill it upfront. 

We can't really accept this PR because, like you pointed out, it would be a breaking change. 

If we wanted to be "lazy" about dealing with `Symbol.observable`, one thing that could be done is to move its set up to the constructor of `Observable`:

```ts
constructor() {
   this[Symbol.observable ?? '@@observable'] = ...
}
```

However, that shouldn't really be necessary, if people are using polyfills properly.

josepot:
Thanks @benlesh ! That makes sense.

One last question, though: I know that RxJS used to polyfill in the past, and that for some reason (that I currently ignore) it stopped doing that on v6. However, the more that I think about it, the more I realize that either all libraries that deal with InteropObservables try to polyfill, or InteropObservables won't be as interoperable as they were intended to be... At least in the sense that it can become quite challenging for the consumers of those libraries to get it right. 

What would be the harm on having this tiny polyfill on RxJS? 

```ts
if (!Symbol["observable"]) {
  Object.defineProperty(Symbol, "observable", {
    value: Symbol("observable")
  });
}
```

PR #6691, opened 2021-11-29
fix: when crossDomain is using default the X-Requested-With header is no longer added incorrectly

### LINKED ISSUE #6663: Rxjs 7.0+ breaking change requires preflight requests for Ajax calls that do not specify crossDomain
## Bug Report

**Current Behavior**

In 6.6 crossDomain is true by default, and therefore the X-Requested-With header is not added, and thus a standard Ajax request does not require preflight requests and one's servers do not need to implement support for OPTIONS requests.

Code in 6.6: https://github.com/ReactiveX/rxjs/blob/6.6.7/src/internal/observable/dom/AjaxObservable.ts#L205
The `request` object is populated at https://github.com/ReactiveX/rxjs/blob/6.6.7/src/internal/observable/dom/AjaxObservable.ts#L181
And this `request` object contains defaults, so has `crossDomain: true`.

In 7.0+ crossDomain is still true by default, however, X-Requested-With header is ALWAYS added to ALL requests made by AJAX, regardless, unless crossDomain is explicitly provided as false. This is because the default value of true is missing at the point in time at which the header is added.

Code in 7.4: https://github.com/ReactiveX/rxjs/blob/7.4.0/src/internal/ajax/ajax.ts#L337
The `config` object is directly from the caller and does not contain defaults because the defaults are not added until later on: https://github.com/ReactiveX/rxjs/blob/7.4.0/src/internal/ajax/ajax.ts#L355

**Expected behavior**

No preflight request for a standard simple ajax request with no crossDomain option specified (which should default to true)

**Reproduction**

RxJS 7.4 Codepen: https://codepen.io/Driskell/pen/MWvQymO?editors=1111
RxJS 6.6 Codepen: https://codepen.io/Driskell/pen/yLovOxK?editors=1111

Open console in your browser and click the button. In Safari on macOS you can see on 7.4 preflight response is not successful, but on 6.6 it does not attempt one.

**Environment**

- Runtime: Safari, Chrome, etc.
- RxJS version: 7.0+

**Possible Solution**

Check if `config.crossDomain` is defined first, or destructure it with a default on the first line of the function: https://github.com/ReactiveX/rxjs/blob/7.4.0/src/internal/ajax/ajax.ts#L282

N/A

driskell:
**Description:**

When this header is added incorrectly, it triggers preflight requests. This caused a breaking change from rxjs 6.x with any backend request URL that does not support preflight requests (OPTIONS method). See the #6663 for more details.

Issue traces back to https://github.com/ReactiveX/rxjs/commit/0d66ba458f07fba51cfc73440d01ef453c24cda7 which is really strange as it seems to change tests; likely just an accidental change. As an example just because I know I changed a lot of tests here to fix them... If you look at https://github.com/ReactiveX/rxjs/commit/0d66ba458f07fba51cfc73440d01ef453c24cda7#diff-0f6e5e92622cfbc19d5344e07b65569c748bce03635ad9be2ffb87321b12d228R96 you can see that the original test changes because now it expects a header in the request the original did not expect. So possibly it was thought that the header was being set but not checked - but actually due to deep equals it was already correct and the header was not meant to be set. There's likely other commits that further embedded the bad behaviour into the tests. For example the upload/download progress the check on the request object shows `crossDomain: true` but then it shows the header is also verified to be there even though when `true` it should never be there.

Some other changes I did to tests are a result of a feature that was built after the broken behaviour was introduced, and that's this feature: https://github.com/ReactiveX/rxjs/commit/1a2c2e49482a460778ea92c7f6a92e58cc3e87bb

**BREAKING CHANGE: (I don't believe it is)**

This fixes a breaking change from v6->v7. I do not believe this fix itself would break any v7 code as it should only mean code that was preflighting when it wasn't in v6 no longer preflights again. Unless some code is explicitly written that relies on the preflight this shouldn't cause any impact.

Fixes #6663

### DIFF
--- a/spec/observables/dom/ajax-spec.ts
+++ b/spec/observables/dom/ajax-spec.ts
@@ -103,7 +103,6 @@ describe('ajax', () => {
       'content-type': 'kenny/loggins',
       'fly-into-the': 'Dangah Zone!',
       'take-a-ride-into-the': 'Danger ZoooOoone!',
-      'x-requested-with': 'XMLHttpRequest',
     });
     // Did not mutate the headers passed
     expect(obj.headers).to.deep.equal({
@@ -129,6 +128,7 @@ describe('ajax', () => {
         url: '/some/path',
         xsrfCookieName: 'foo',
         xsrfHeaderName: 'Custom-Header-Name',
+        crossDomain: false,
       };
 
       ajax(obj).subscribe();
@@ -188,10 +188,31 @@ describe('ajax', () => {
       const request = MockXMLHttpRequest.mostRecent;
 
       expect(request.url).to.equal('/some/page');
-      expect(request.requestHeaders).to.deep.equal({
-        'x-requested-with': 'XMLHttpRequest',
-      });
+      expect(request.requestHeaders).to.deep.equal({});
     });
+  });
+
+  it('should not set the X-Requested-With if crossDomain is true', () => {
+    ajax({
+      url: '/test/monkey',
+      method: 'GET',
+      crossDomain: true,
+    }).subscribe();
+
+    const request = MockXMLHttpRequest.mostRecent;
+
+    expect(request.requestHeaders).to.not.have.key('x-requested-with');
+  });
+
+  it('should not set the X-Requested-With if crossDomain is not present (it defaults to true)', () => {
+    ajax({
+      url: '/test/monkey',
+      method: 'GET',
+    }).subscribe();
+
+    const request = MockXMLHttpRequest.mostRecent;
+
+    expect(request.requestHeaders).to.not.have.key('x-requested-with');
   });
 
   it('should set the X-Requested-With if crossDomain is false', () => {
@@ -590,9 +611,7 @@ describe('ajax', () => {
 
       expect(MockXMLHttpRequest.mostRecent.url).to.equal('/flibbertyJibbet');
       expect(MockXMLHttpRequest.mostRecent.data).to.equal(body);
-      expect(MockXMLHttpRequest.mostRecent.requestHeaders).to.deep.equal({
-        'x-requested-with': 'XMLHttpRequest',
-      });
+      expect(MockXMLHttpRequest.mostRecent.requestHeaders).to.deep.equal({});
     });
 
     it('should send the URLSearchParams straight through to the body', () => {
@@ -789,7 +808,6 @@ describe('ajax', () => {
       expect(request.url).to.equal('/flibbertyJibbet');
       expect(request.requestHeaders).to.deep.equal({
         'content-type': 'application/json;charset=utf-8',
-        'x-requested-with': 'XMLHttpRequest',
       });
 
       request.respondWith({
@@ -820,9 +838,7 @@ describe('ajax', () => {
 
       expect(request.method).to.equal('POST');
       expect(request.url).to.equal('/flibbertyJibbet');
-      expect(request.requestHeaders).to.deep.equal({
-        'x-requested-with': 'XMLHttpRequest',
-      });
+      expect(request.requestHeaders).to.deep.equal({});
 
       request.respondWith({
         status: 204,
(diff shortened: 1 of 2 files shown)


spec/observables/dom/ajax-spec.ts
```diff
@@ -1154,9 +1167,7 @@ describe('ajax', () => {
         async: true,
         body: undefined,
         crossDomain: true,
-        headers: {
-          'x-requested-with': 'XMLHttpRequest',
-        },
+        headers: {},
```
driskell, line 1170: This hopefully helps understand why I had to update many tests. It shows the bad behaviour in 4 lines - `crossDomain` is `true` but at the same time we were expecting the header that should never be sent when `crossDomain` is `true`

benlesh:
Hi, @driskell! Thank you for finding this issue and coming up with a fix. I've added a commit to what you did and started a new PR here #6696 ... Sorry for the confusion, I was using the Github Web Editor, and it would not let me add a commit to this PR for some reason.

Basically, the problem with this PR was that it changed the defaults of `ajax` while at the same time cleaning up some messy code. Please review the other PR and check to make sure it's satisfactory, and that there are no issues.

src/internal/ajax/ajax.ts
```diff
@@ -279,7 +279,20 @@ export function fromAjax<T>(config: AjaxConfig): Observable<AjaxResponse<T>> {
     // Here we're pulling off each of the configuration arguments
     // that we don't want to add to the request information we're
     // passing around.
-    const { queryParams, body: configuredBody, headers: configuredHeaders, ...remainingConfig } = config;
+    const configWithDefaults = {
+      // Default values
+      async: true,
+      crossDomain: true,
```
benlesh, line 285: Moving this up here changes some of the conditional branches below, and therefor would (from what I can tell) result in a breaking change. Please see the changes to your PR in the new PR (linked in my other comment)

Thanks! :)

driskell: Hi @benlesh - Which conditionals does it change? I thought I checked and it will only change the crossDomain conditional to use the correct default.

The default of crossDomain always has been true. Otherwise every single CORS request will default to sending a preflight which is a breaking change from v6.

See here default in v6: https://github.com/ReactiveX/rxjs/blob/6.x/src/internal/observable/dom/AjaxObservable.ts#L163

driskell:
> Basically, the problem with this PR was that it changed the defaults of ajax while at the same time cleaning up some messy code. Please review the other PR and check to make sure it's satisfactory, and that there are no issues.

Sorry for the multiple comments. Just so it’s tracked in the main comments area - as far as I could tell this resets the defaults back to v6 because as I can tell there shouldn’t have been a change. It doesn’t seem to make sense to me to change crossDomain to default to false. Here is v6: https://github.com/ReactiveX/rxjs/blob/6.x/src/internal/observable/dom/AjaxObservable.ts#L163

I did the cleanup here because I wanted to avoid separating the setting of defaults of crossDomain and other settings and it seemed to make no logic change except to crossDomain which was the issue. It also avoided a check for crossDomain undefined so seemed a cleaner fix.

src/internal/ajax/ajax.ts
```diff
@@ -279,7 +279,20 @@ export function fromAjax<T>(config: AjaxConfig): Observable<AjaxResponse<T>> {
     // Here we're pulling off each of the configuration arguments
     // that we don't want to add to the request information we're
     // passing around.
```
driskell, line 281: Just realised this should move down?

src/internal/ajax/ajax.ts
```diff
@@ -334,7 +347,7 @@ export function fromAjax<T>(config: AjaxConfig): Observable<AjaxResponse<T>> {
     // None of this is necessary, it's only being set because it's "the thing libraries do"
     // Starting back as far as JQuery, and continuing with other libraries such as Angular 1,
     // Axios, et al.
-    if (!config.crossDomain && !('x-requested-with' in headers)) {
+    if (!configWithDefaults.crossDomain && !('x-requested-with' in headers)) {
```
driskell, line 350: Maybe to match the below conditional should use remainingConfig

src/internal/ajax/ajax.ts
driskell: Checking I can see only two parts of this used below: withCredentials and crossDomain. The former defaults false and so when it was undefined would mean no logic change so exact same behaviour. So the only impact is crossDomain, which as noted should default true not false to match v6 and behave as expected.

PR #6693, opened 2021-11-29
fix: schedulers will no longer error while rescheduling and unsubscri…

benlesh:
attn @cartant


.prettierrc.json
```diff
@@ -0,0 +1,13 @@
+{
```
benlesh, line 1: I only added this to keep my IDE from messing with me. haha.

benlesh:
Given that this is old as hell, and no one has complained about this behavior. I'm just going to close it.

Issue #3935, opened 2018-07-18
WebSocket created during close of a previous one is never closed


ChristofferGersen:
## Bug Report

**Current Behavior**
When there are no more subscriptions to a `WebSocketSubject`, then the underlying `WebSocket` is closed. If you then create a new subscription on the `WebSocketSubject`, before the `onclose` of the previous `WebSocket` has fired, then the newly created `WebSocket` will never be closed.

If you combine this with the `multiplex` function, then some subscribe/unsubscribe messages will not be send as well.

**Reproduction**
1. Create `WebSocketSubject`
2. Create two subscriptions on `WebSocketSubject`
3. Wait a second
4. Unsubscribe both subscriptions and immediately create two new ones
5. Repeat steps 3 and 4
6. Unsubscribe both subscriptions

https://stackblitz.com/edit/typescript-n7jv6k

In case the stackblitz does not work, here is the code it contains:

<details><summary>Show imports and utility class</summary>
<p>

```ts
import { empty, interval, merge } from "rxjs";
import { webSocket } from "rxjs/webSocket";
import { switchMap, take } from "rxjs/operators";

function log(message: string) {
  console.log(message);
}

function error(err: any) {
  console.error(err);
}

class LoggingWebSocket implements WebSocket {
  private static nextSocketId = 0;
  readonly socketId = LoggingWebSocket.nextSocketId++;
  readonly CONNECTING: number = 0;
  readonly OPEN: number = 1;
  readonly CLOSING: number = 2;
  readonly CLOSED: number = 3;

  binaryType = "blob";
  readonly bufferedAmount: number = 0;
  readonly extensions: string = "";
  onclose: ((this: WebSocket, ev: CloseEvent) => any) | null = null;
  onerror: ((this: WebSocket, ev: Event) => any) | null = null;
  onmessage: ((this: WebSocket, ev: MessageEvent) => any) | null = null;
  onopen: ((this: WebSocket, ev: Event) => any) | null = null;
  readonly protocol: string;
  readyState: number = this.CONNECTING;
  constructor(
    public readonly url: string,
    public protocols: string | string[] = ""
  ) {
    this.protocol = Array.isArray(protocols) ? protocols[0] || "" : protocols;
    log("Created socket " + this.socketId);
    setTimeout(() => {
      this.readyState = this.OPEN;
      log("Opened socket " + this.socketId);
      this.dispatchEvent({ type: "open" });
    }, 100);
  }
[...]

PR #6708, opened 2021-12-06
fix(WebSocketSubject): handle slow WebSocket close

### LINKED ISSUE #4650: WebSocketSubject fails to recreate socket when close isn't immediate
## Bug Report

**Current Behavior**
If the close of a websocket takes a little bit of time subsequent calls to multiplex fail to send any subscription messages and close the socket immediately. See reproduction below.

**Reproduction**
- https://stackblitz.com/edit/rxjs-n2kgl6

```js
import { interval } from "rxjs";
import { webSocket } from "rxjs/webSocket";
import { switchMap, take } from "rxjs/operators";

const socket = webSocket<any>({
  url: "ws://localhost/ws"
});

interval(2_000)
  .pipe(
    take(2),
    switchMap(id => {
        return socket.multiplex(
          () => `sub ${id}`,
          () => `unsub ${id}`,
          value => value === id
        );
    })
  )
  .subscribe({
    next: _ => { },
    error: err => console.error(err),
    complete: () => console.log("Complete")
  });
```

**Expected behavior**
A new call to multiplex should subscribe and stay open.

**Environment**
- 6.4.0

maknapp:
**Description:**
The close event on WebSocket does not always occur immediately after running the close function. If the close event is occurs after a new WebSocket is opened, then the reset needs to be skipped. This checks to make sure the socket being reset is the one that matches the event.

The MockWebSocket class skipped the state between CLOSING and CLOSED. I removed the automatic closing so that the CLOSED state must be trigger manually. This caused many of the existing tests to expect CLOSING instead - I changed the codes instead of triggering close because it did not affect the tests.

I also changed the readyState codes to use an enum for clarity. If that is an unwanted change I can revert back to comments.

**Related issues:**
Closes #4650, #3935

### DIFF
--- a/spec/observables/dom/webSocket-spec.ts
+++ b/spec/observables/dom/webSocket-spec.ts
@@ -7,6 +7,13 @@ const root: any = (typeof globalThis !== 'undefined' && globalThis)
   || (typeof self !== 'undefined' && self)
   || global;
 
+enum WebSocketState {
+  CONNECTING = 0,
+  OPEN = 1,
+  CLOSING = 2,
+  CLOSED = 3
+}
+
 /** @test {webSocket}  */
 describe('webSocket', () => {
   let __ws: any;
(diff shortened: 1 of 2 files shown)


benlesh (approved):

benlesh:
This seems like a reasonable fix. Thank you for submitting the PR @maknapp. It seems this has been an issue for a while. The WebSocketSubject can definitely use some love. This is appreciated.

PR #6711, opened 2021-12-07
docs(operators): shareReplay refCount is false by default

samal-rasmussen:
The default value of refCount is not documented currently, so I added some info on it, so that people are not caught unaware.

This is well described here:
https://ncjamieson.com/whats-changed-with-sharereplay/


benlesh:
Thank you, @samal-rasmussen

PR #6738, opened 2022-01-05
fix: use "unknown" type for operators that ignore input values

Wyzard256:
**Description:**

After upgrading an application's typescript-eslint dependency from 4.30 to 5.9, I found that the "@typescript-eslint/no-unsafe-argument" rule was flagging my use of the RxJS `mapTo` and `ignoreElements` operators, because they return functions that operate on observables of `any`.  Since the operator functions ignore the values from the source observable, they don't need the dynamic typing that `any` provides for *accessing* the values.  Using `unknown` instead makes it clear (to both eslint and humans) that the operator function doesn't have any expectations about the type of observable it's given.

This PR changes `OperatorFunction<any, ...>` to `OperatorFunction<unknown, ...>` in the return type of `ignoreElements` and the `mapTo` family of operators.  No implementation changes are needed in the functions, since the associated values are never used anyway.  This shouldn't be a breaking change: `unknown` is the same as `any` from a caller's perspective (no restrictions on what can be passed in), and `OperatorFunction<unknown, ...>` is assignable to `OperatorFunction<any, ...>` so anything that explicitly expects the latter is still OK.


benlesh (approved):

Wyzard256:
Looks like the CI build failed because it's an API change?  I'm not familiar with api_guardian typical usage and it's not mentioned in the contributor guide, so I don't know whether there's an action I should take (like editing `api_guard/dist/types/index.d.ts` as part of this PR), or if it's something that a maintainer will handle.

PR #6678, opened 2021-11-19
docs: update operator imports in examples

jakovljevic-mladen:
**Description:**
As of RxJS version 7.2.0, most operators can be imported from `'rxjs'` export site. This PR updates code examples in most places. However, content from `*.md` files from `docs_app/content` folder is not updated because I'm still waiting for #6548 to be merged. Once that one gets merged, I will update files from `docs_app/content`.

Also, please forgive me for trying to be consistent and please try to ignore changes such as:
- removing brackets: `(x)` -> `x`
- removing trailing commas
- fixing whitespace issues
- adding missing `;`
- removing H3 headers (`###`) from examples - most of them are not present in most examples, so I removed almost all remaining ones
- replacing `"` with `'`
- and so on...

**Related issue (if exists):**
Closes #6650

Issue #6650, opened 2021-10-21
[Docs] update code examples to use new import


niklas-wortmann:
Out of the discussion from #6548, we should update the code examples replacing the old (<7.2) way of importing operators, with the new one.
e.g.

`import {map} from 'rxjs/operators`;

would be

`import { map } from 'rxjs'`

jakovljevic-mladen:
I will take this one.

PR #6739, opened 2022-01-05
docs: update operator imports in remaining examples

jakovljevic-mladen:
**Description:**
Fix remaining examples from `.md` files from `docs_app/content/` folder.

#6650 and #6678

### DIFF
--- a/docs_app/content/deprecations/multicasting.md
+++ b/docs_app/content/deprecations/multicasting.md
@@ -34,7 +34,8 @@ Instead of creating a [ConnectableObservable](/api/index/class/ConnectableObserv
 
 <!-- prettier-ignore -->
 ```ts
-import { ConnectableObservable, Subject, timer } from 'rxjs';
+import { ConnectableObservable, timer, Subject } from 'rxjs';
+
 // deprecated
 const tick$ = new ConnectableObservable(
   timer(1_000),
@@ -44,7 +45,8 @@ tick$.connect();
 
 <!-- prettier-ignore -->
 ```ts
-import { connectable, Subject, timer } from 'rxjs';
+import { connectable, timer, Subject } from 'rxjs';
+
 // suggested refactor
 const tick$ = connectable(timer(1_000), {
   connector: () => new Subject()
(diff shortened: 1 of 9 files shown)


docs_app/content/deprecations/multicasting.md
```diff
@@ -34,7 +34,8 @@ Instead of creating a [ConnectableObservable](/api/index/class/ConnectableObserv
 
 <!-- prettier-ignore -->
 ```ts
-import { ConnectableObservable, Subject, timer } from 'rxjs';
```
benlesh, line 37: I'm not sure why this was changed? Was there some sort of automated fix?

docs_app/content/deprecations/multicasting.md
```diff
@@ -44,7 +45,8 @@ tick$.connect();
 
 <!-- prettier-ignore -->
 ```ts
-import { connectable, Subject, timer } from 'rxjs';
+import { connectable, timer, Subject } from 'rxjs';
```
benlesh, line 48: same here.

benlesh (approved):
In general, nothing to block the PR on, but I am curious why some of the changes are just reordering imports that were otherwise fine. Is there a tool you're using?

docs_app/content/deprecations/multicasting.md
jakovljevic-mladen: I don't use any tool, all I use is WebStorm and [this handful shortcut](https://www.jetbrains.com/webstorm/guide/tips/move-elements-left-right/) which works in this line as well, which is amazing, since this is not actual code, but a comment.

But, why I changed this line? Because I did the same set of changes in examples below this one - as I said in #6678, it was much easier for me to keep track of all objects I need. What I did with every changed example is to run the docs script, wait to re-render the example in browser and then I opened StackBlitz to verify that everything is working just fine. In the end, when SB opened and was ready, I checked every imported object to see if it is used - those that were not used were never highlighted in the code below - this is how ordering objects by the time they were used helped me.

In this particular example, it was just my OCD to make things consistent through entire file, nothing else. I don't ever expect anyone else to keep doing the same consistency changes as I did, sorry about that.

PR #6743, opened 2022-01-08
chore: update docs app to Angular 13

jakovljevic-mladen:
**Description:**
This PR updates docs app to the latest Angular version - v13.1.1. The update was preformed by using `ng update` command which changed some files and dependencies. Some other files were manually adjusted.

None


docs_app/angular.json
```diff
@@ -198,16 +198,6 @@
...
-        },
-        "lint": {
-          "builder": "@angular-devkit/build-angular:tslint",
-          "options": {
-            "tsConfig": [
-              "src/tsconfig.app.json",
-              "src/tsconfig.spec.json"
-            ],
-            "exclude": []
-          }
```
fasidOnGit, line 210: this is precisely related to `"@angular-devkit/build-angular": "~13.1.2",` package. you can still upgrade to angular 13 and **not** upgrade `@angular-devkit/build-angular` (stay in v12). this way you can put back the lint. and probably we can address the linting and the afore-mentioned package upgrade after https://github.com/ReactiveX/rxjs/pull/6744 is done. wdyt ?

fasidOnGit (changes requested):

docs_app/angular.json
jakovljevic-mladen: TBH, I'd rather keep everything in one PR - it's easier to maintain and to keep track. So, I'd rather keep this removed and reintroduced later, if this is OK to you?

fasidOnGit: >  I'd rather keep this removed

I will leave that to the gatekeepers of this library here.

IMHO, I wouldn't remove linting since we open the possibility of code that goes against our linting rules (without people being aware). So, I'd rather migrate to eslint in this PR. or do no upgrade `@angular-devkit/build-angular` to 13 until linting is figured out.

benlesh: I'll be honest, temporarily disabling linting in the docs app is probably one of my last concerns. But I'll bring it up at the Core Team meeting.

benlesh:
Core team: Proceed and disable linting in docs-app for now, with a follow up to get it linting with eslint.

PR #6749, opened 2022-01-11
Removing deprecated empty and never functions

benlesh:
Just removing `empty()` and `never()` finally. (version 8 only)

### DIFF
--- a/src/internal/observable/empty.ts
+++ b/src/internal/observable/empty.ts
@@ -1,5 +1,4 @@
 import { Observable } from '../Observable';
-import { SchedulerLike } from '../types';
 
 /**
  * A simple Observable that emits no items to the Observer and immediately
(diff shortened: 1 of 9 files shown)


src/internal/observable/empty.ts
```diff
@@ -26,66 +25,3 @@ import { SchedulerLike } from '../types';
...
- *   mergeMap(x => x % 2 === 1 ? of('a', 'b', 'c') : empty()),
- * );
- * result.subscribe(x => console.log(x));
- *
- * // Results in the following to the console:
- * // x is equal to the count on the interval, e.g. (0, 1, 2, 3, ...)
- * // x will occur every 1000ms
- * // if x % 2 is equal to 1, print a, b, c (each on its own)
- * // if x % 2 is not equal to 1, nothing will be output
- * ```
```
jakovljevic-mladen, line 72: In #6746, I preserved these examples. Could you please review that PR?

src/internal/observable/empty.ts
```diff
@@ -1,5 +1,4 @@
 import { Observable } from '../Observable';
```
jakovljevic-mladen, line 1: Nit (maybe completely unimportant, but still): since many other files contain both uppercase and lowercase letters in file name, for consistency reasons, this one could be renamed like this: `empty.ts` -> `EMPTY.ts`. I'm not sure if it makes sense to do it, this is just a reminder. If it makes sense, the same renaming would be needed for all `-spec` and `NEVER` related files.

benlesh: I agree. I think we probably need a pass to do git moves for a lot of files that are named silly things, honestly. But we can probably do that in another PR.

kwonoj (approved):

PR #6734, opened 2021-12-28
docs: fix links in code snippets

jakovljevic-mladen:
**Description:**
I'm once again sorry for the very large PR. This PR fixes issue where [`autoLinkCode`](https://github.com/ReactiveX/rxjs/blob/bca49bdb0dd78f3828e818d64676153c1a491dc0/docs_app/tools/transforms/angular-base-package/post-processors/auto-link-code.js#L23) wasn't able to produce links in code snippets. For example, if there was a snippet like this (in any exported `.ts` or `.md` file):

```ts
import { from, Subject } from 'rxjs';
import { multicast } from 'rxjs/operators';
 
const source = from([1, 2, 3]);
const subject = new Subject();
const multicasted = source.pipe(multicast(subject));
 
// These are, under the hood, `subject.subscribe({...})`:
multicasted.subscribe({
  next: (v) => console.log(`observerA: ${v}`)
});
multicasted.subscribe({
  next: (v) => console.log(`observerB: ${v}`)
});
 
// This is, under the hood, `source.subscribe(subject)`:
multicasted.connect();
```

`autoLinkCode` would generate links wherever `from`, `Subject` and `multicast` were written because `from`, `Subject` and `multicast` are valid documentation pages.

The real reason why this `autoLinkCode` failed to do some processing was that it was actually excluded from `postProcessHtml.plugins`. At first, it [was added](https://github.com/ReactiveX/rxjs/blob/bca49bdb0dd78f3828e818d64676153c1a491dc0/docs_app/tools/transforms/angular-base-package/index.js#L141) to the `postProcessHtml.plugins` array, but later, `plugins` array [was overwritten](https://github.com/ReactiveX/rxjs/blob/bca49bdb0dd78f3828e818d64676153c1a491dc0/docs_app/tools/transforms/angular-api-package/index.js#L198) with another array (with a `embedMarbleDiagramsPostProcessor` plugin). Now, I fixed it to push `embedMarbleDiagramsPostProcessor` to an existing array instead of creating a new one.

`autoLinkCode` was adjusted so that it follows [the same code structure from Angular project](https://github.com/angular/angular/blob/8ebc946c0e7bf80d26ec8268acb4ff0af9e5c34a/aio/tools/transforms/angular-base-package/post-processors/auto-link-code.js).

[...]

PR #6750, opened 2022-01-11
docs(identity): add identity docs

jakovljevic-mladen:
**Description:**
Now that #6734 is merged, accessing to `identity` page is much more likely since `<code>` tags contain references to it. It's time to document it.

image

None

### DIFF
--- a/src/internal/util/identity.ts
+++ b/src/internal/util/identity.ts
@@ -1,3 +1,29 @@
+/**
+ * This function takes one parameter and just returns it. Simply put,
+ * this is like `<T>(x: T): T => x`.
+ *
+ * ## Example
+ *
+ * This is useful in some cases when using things like `mergeMap`
+ *
+ * ```ts
+ * import { interval, take, map, range, mergeMap, identity } from 'rxjs';
+ *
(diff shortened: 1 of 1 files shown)


src/internal/util/identity.ts
```diff
@@ -1,3 +1,29 @@
...
+ * const result$ = source$.pipe(
+ *   map(i => range(i)),
+ *   mergeMap(identity) // same as mergeMap(x => x)
+ * );
+ *
+ * result$.subscribe({
+ *   next: console.log
+ * });
+ * ```
+ *
```
benlesh, line 23: One of the main use cases I have for this is with a ternary to selectively apply an operator:

```ts
source$.pipe(
  condition ? mergeMap(doSomething) : identity
)
```

It would be nice to have an example of that.

jakovljevic-mladen: I'll add.

jakovljevic-mladen: Added (thought, as the second example). If you'd like this example to be the first one, I can swap them.

benlesh (approved):

PR #6755, opened 2022-01-11
feat(onErrorResumeNextWith): renamed `onErrorResumeNext` and exported from the top level

benlesh:
The `onErrorResumeNext` operator is deprecated due to a rename. The rename was done so we can move all operator exports to the top level `rxjs` export site. We were not currently exporting that operator.

This commit also refactors to logic to be based on the creation function, which is more widely used

### DIFF
--- a/src/internal/operators/onErrorResumeNextWith.ts
+++ b/src/internal/operators/onErrorResumeNextWith.ts
@@ -1,15 +1,11 @@
-import { Observable } from '../Observable';
 import { ObservableInputTuple, OperatorFunction } from '../types';
-import { operate } from '../util/lift';
-import { innerFrom } from '../observable/innerFrom';
 import { argsOrArgArray } from '../util/argsOrArgArray';
-import { createOperatorSubscriber } from './OperatorSubscriber';
-import { noop } from '../util/noop';
+import { onErrorResumeNext as oERNCreate } from '../observable/onErrorResumeNext';
 
-export function onErrorResumeNext<T, A extends readonly unknown[]>(
+export function onErrorResumeNextWith<T, A extends readonly unknown[]>(
   sources: [...ObservableInputTuple<A>]
 ): OperatorFunction<T, T | A[number]>;
-export function onErrorResumeNext<T, A extends readonly unknown[]>(
+export function onErrorResumeNextWith<T, A extends readonly unknown[]>(
   ...sources: [...ObservableInputTuple<A>]
 ): OperatorFunction<T, T | A[number]>;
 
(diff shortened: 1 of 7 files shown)


benlesh:
This one is just one we missed. I don't like the late deprecation, but I somehow doubt the operator version of this is in wide use.

jakovljevic-mladen:
[importing.md](https://github.com/ReactiveX/rxjs/blob/17380f1f05/docs_app/content/guide/importing.md#how-to-migrate) file should be updated as well.

src/internal/operators/onErrorResumeNextWith.ts
```diff
@@ -85,41 +81,19 @@ export function onErrorResumeNext<T, A extends readonly unknown[]>(
...
-
-    subscribeNext();
-  });
+  return (source) => oERNCreate(source, ...nextSources);
 }
+
+/**
+ * @deprecated Renamed. Use {@link onErrorResumeNextWith} instead. Will be removed in v8.
+ */
+export const onErrorResumeNext = onErrorResumeNextWith;
```
jakovljevic-mladen, line 99: Shouldn't `onErrorResumeNext` stay in this file and `onErrorResumeNextWith` get exported from another file (`onErrorResumeNextWith.ts`)? This is what's done with `combineLatest`, `concat` and others...

src/operators/index.ts
```diff
@@ -57,7 +57,7 @@ export { mergeWith } from '../internal/operators/mergeWith';
 export { min } from '../internal/operators/min';
 export { multicast } from '../internal/operators/multicast';
 export { observeOn } from '../internal/operators/observeOn';
-export { onErrorResumeNext } from '../internal/operators/onErrorResumeNext';
+export { onErrorResumeNext } from '../internal/operators/onErrorResumeNextWith';
```
jakovljevic-mladen, line 60: Suggested change:
```
export { onErrorResumeNext } from '../internal/operators/onErrorResumeNext';
export { onErrorResumeNextWith } from '../internal/operators/onErrorResumeNextWith';
```

jakovljevic-mladen (commented):
I think some things are not aligned with other similar changes, for example in `combineLatestWith` or `concat`...

kbrilla:
any update on this PR? It's waiting for over half a year already.

src/internal/operators/onErrorResumeNextWith.ts
benlesh: I'm indifferent about that. That's all internal implementation and has no bearing on users.

jakovljevic-mladen: I think we may have issues with the docs builder, though I'm not completely sure. Let me check it.

jakovljevic-mladen: I can't really make the docs work properly. The code changes from this PR seem to create multiple issues.

One of the issues that this function has is described in #7009.

The other is this:

image

The dgeni didn't properly infer the type. The inferred type seems to be: `{ <T, A extends any, unknown>(): any; <T, A extends any, unknown>(): any; <T, A extends any, unknown>(): any; };` which doesn't really mean anything.

I will try to dig deeper into the issue tomorrow.

jakovljevic-mladen: It looks like other constants are having the same issue. E.g. [`flatMap`](https://rxjs.dev/api/index/const/flatMap):

image

So, I don't think this is something we can do about by moving to another file, which would be OK only for consistency reasons, nothing else.

jakovljevic-mladen (approved):

PR #6769, opened 2022-01-14
feat(operators): combineAll operator

EarthyOrange:
**Description:**
An operator to combine values from source observables. The values emitted by early observers are saved until the last observer emits, following which all the previously saved values are emitted.

```
first$: ---a---b---c--
second$: ---------------d--
result$: [ad, bd, cd]
```

### DIFF
--- a/src/internal/util/tetris.ts
+++ b/src/internal/util/tetris.ts
@@ -1,0 +1,50 @@
+export class Tetris<T extends any> {
+  private readonly arrays: Array<Array<T>>;
+  private readonly lastEmissionIndices: Array<number>;
+
+  constructor(readonly columns: number = 2) {
+    this.arrays = new Array<Array<T>>();
+    this.lastEmissionIndices = new Array<number>(columns);
+    for (let i = 0; i < columns; i++) {
+      this.arrays[i] = new Array<T>();
+      this.lastEmissionIndices[i] = -1;
+    }
+  }
+
+  getNext(): Array<T> {
+    const next = new Array<T>();
+    if (!this.hasNext()) {
+      return [];
+    }
+    for (let column = 0; column < this.columns; column++) {
+      if (this._hasNewIn(column)) {
+        this.lastEmissionIndices[column] += 1;
+      }
+      next.push(this.arrays[column][this.lastEmissionIndices[column]]);
+    }
+    return next;
+  }
+
+  private _hasNewIn(column: number): boolean {
+    if (this.lastEmissionIndices[column] < this.arrays[column].length - 1) {
+      return true;
+    }
+    return false;
+  }
+
+  hasNext(): boolean {
+    let hasNext: boolean = false;
+    for (let column = 0; column < this.columns; column++) {
+      if (this.arrays[column].length === 0) {
+        return false;
+      }
+      hasNext = hasNext || this._hasNewIn(column);
+    }
+    return hasNext;
+  }
+
+  pushToColumn(column: number, value: T): Tetris<T> {
+    this.arrays[column].push(value);
+    return this;
+  }
+}
--- a/src/internal/util/tetris.ts
+++ b/src/internal/util/tetris.ts
@@ -18,6 +18,7 @@ export class Tetris<T extends any> {
     }
     for (let column = 0; column < this.columns; column++) {
       if (this._hasNewIn(column)) {
+        delete this.arrays[column][this.lastEmissionIndices[column]];
         this.lastEmissionIndices[column] += 1;
       }
       next.push(this.arrays[column][this.lastEmissionIndices[column]]);
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "rxjs",
-  "version": "7.5.2",
+  "version": "7.6.0",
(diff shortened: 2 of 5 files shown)


EarthyOrange:
@benlesh @cartant Can you please take a look and let me know what your thoughts on the changes? Please consider it a draft change. I want a green light before I add more to this PR. Thanks in advance.

kwonoj:
Same as prior feature implementation requests (https://github.com/ReactiveX/rxjs/pull/5939#issuecomment-758284916, https://github.com/ReactiveX/rxjs/pull/5089, and others)

Core team strongly suggest to create new features via userland custom operators instead of including it in core as we generally trying to reduce core api surface, also with `pipe` it is straightforward to create a new operator as separate library. If a new features should require core lib change we'd like to ask to have some details what part of core lib change is required & usecases it should be included in core.

EarthyOrange:
@kwonoj I have been implementing operators in my library and would have preferred to do the same for this operator as well. But looking at the implementation of the combineAll operator, it looked like I needed to have access to some internals like OperatorSubscriber and executeSchedule methods. If there is a replacement of these internals, that is already present in the exposed APIs, I will be happy to use those and close this PR.

From what I further understand from your comment is that the direction of this library development now is towards making it a framework which can be used to implement the operators in user libraries or perhaps even make the current operators available in a separate library which would rely on this core library. Is that right?

benlesh:
It seems like this PR is reimplementing `combineLatestAll`? I guess I'm not sure I understand this PR. Is there an issue that this is solving?

src/internal/util/tetris.ts
```diff
@@ -0,0 +1,51 @@
+export class Tetris<T extends any> {
```
benlesh, line 1: Based off of the name, I'm not sure what the intent of this class is. Why is it called "Tetris"?

EarthyOrange: @benlesh 

Apologies for the poor documentation. I wasn't sure if this change would be accepted so I just over-wrote a  deprecated operator for now, to demonstrate the change. It should, of-course, be renamed to something more appropriate and the code broken down and moved to a different folder (Happy to do it).

**Setup**: On subscription to 2 observables (which have the replay(buffer: 5) and share operators) the first one emits 4 values (a, b, c, d) and the second one emits 1 (e).
**Expectation**: Since both the hot observables have values ready to be forwarded, the expectation was that combineLatest would give (ae, be, ce, de).
**Actual**: The received value is (de).

The operator is similar to combineLatestAll, but differs from it by storing all the values that it receives from the time of subscription in the Tetris object.
Consider the example in the [description](https://github.com/ReactiveX/rxjs/pull/6769#issue-1103188742). The combineLatestAll operator would have just emitted the combination `cd`. But this operator stored the values emitted by the first observable until the second one emitted its first, following which it sequentially emitted all the values from the first observable paired with the one value of the second.

The purpose of the Tetris object is:
 - to maintain one array of emitted values per observable (column value in constructor)
 - to keep track of the last emitted values (lastEmittedIndices)
 - to let us know if there is a new available value (hasNext)
 - return an array of most recently emitted values, in case of a new emission (getNext)

I chose the name Tetris because I played a lot of it when I was a kid. When I saw how the values were "falling" and "disappearing" in the columns, it reminded me of it. It can/should be renamed to something more appropriate that betrays its intent a bit more clearly.

benlesh: I see. So what you're trying to solve here is the condition where the first values of the combined observables all show up synchronously, and you might miss a few values from the "earlier" observables while waiting for the subsequent values to arrive? 

I'm not sure if we can land this in RxJS core, as it's so similar to combineLatest, and probably confusing. I've seen requests for things like this before, but usually it's for people wanting this:

```
a$:   ---1----2-----4-----5-----
b$;   ------------3----------6----
out$: ---A----B---C--D-----E--F--- 

A: [1, null]
B: [2, null]
C: [2, 3]
D: [4, 3]
E: [5, 3]
F: [5, 6]
```

And I *think* that operator already exists in @cartant's rxjs-etc library.

EarthyOrange: <pre><code>
a$:   ---1----2-----------4-----5-----
b$;   ------------3----------6----
out$: ------------(AB)----C--D--E--- 

A: [1, 3]
B: [2, 3]
C: [4, 3]
D: [4, 6]
E: [5, 6]
</pre></code>

EarthyOrange: @benlesh The example posted in your comment doesn't reflect the way this operator works. Can you please recheck the expected output I have posted and confirm if this operator indeed exists in rxjs-etc?

benlesh:
Hi @EarthyOrange — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

PR #6774, opened 2022-01-15
feat: Add allValuesFrom function

Wyzard256:
**Description:**

This PR adds an `allValuesFrom` function that's similar to `firstValueFrom` and `lastValueFrom`, but the promise resolves to an array of all the values emitted by the observable, instead of just the first or last.  (Unlike `firstValueFrom` and `lastValueFrom`, it's not an error if the observable completes without emitting anything: the promise just resolves to an empty array in that case.)

The implementation is a trivial combination of the `toArray` operator with either `firstValueFrom` or `lastValueFrom` (they're equivalent when used with `toArray`), and applications can easily accomplish the same thing directly by using an expression like `firstValueFrom(anObservable.pipe(..., toArray()))`, but the interaction between the two might not be clear to a reader — especially since the `toArray()` could be at the end of a lengthy multi-line pipe, distant from the `firstValueFrom`.  Putting the combination into a function named `allValuesFrom` makes it obvious what the result will be.  And although applications can easily implement the function themselves, it's closely related to `firstValueFrom` and `lastValueFrom` so I think it makes sense for the library to provide it.

The jsdoc and tests are adapted from the ones for `firstValueFrom` and `lastValueFrom`, so similar in structure.

### DIFF
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -1,3 +1,5 @@
+export declare function allValuesFrom<T>(source: Observable<T>): Promise<T[]>;
+
 export declare const animationFrame: AnimationFrameScheduler;
 
 export declare function animationFrames(timestampProvider?: TimestampProvider): Observable<{
--- a/spec/allValuesFrom-spec.ts
+++ b/spec/allValuesFrom-spec.ts
@@ -1,0 +1,39 @@
+import { interval, allValuesFrom, EMPTY, throwError, of } from 'rxjs';
+import { expect } from 'chai';
+import { finalize, take } from 'rxjs/operators';
+
+describe('allValuesFrom', () => {
+  it('should emit all the values as a promise', async () => {
+    let finalized = false;
+    const source = interval(2).pipe(take(4), finalize(() => (finalized = true)));
+    const result = await allValuesFrom(source);
+    expect(result).to.deep.equal([0, 1, 2, 3]);
+    expect(finalized).to.be.true;
+  });
+
+  it('should produce empty arrays for empty observables', async () => {
+    const source = EMPTY;
+    const result = await allValuesFrom(source);
+    expect(result).to.be.empty;
+  });
+
+  it('should error for errored observables', async () => {
+    const source = throwError(() => new Error('blorp!'));
+    let error: any = null;
+    try {
+      await allValuesFrom(source);
+    } catch (err) {
+      error = err;
+    }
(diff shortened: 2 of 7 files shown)


kwonoj:
Same as prior feature implementation requests (https://github.com/ReactiveX/rxjs/pull/5939#issuecomment-758284916, https://github.com/ReactiveX/rxjs/pull/5089, and others)

Core team strongly suggest to create new features via userland custom operators instead of including it in core as we generally trying to reduce core api surface, also with `pipe` it is straightforward to create a new operator as separate library. If a new features should require core lib change we'd like to ask to have some details what part of core lib change is required & usecases it should be included in core.

Wyzard256:
Makes sense for totally-new features, but this is more like a small extension to a feature that's already present.  It'd seem weird to have `allValuesFrom` live in a different library than `firstValueFrom` and `lastValueFrom` since they're so similar and related.

(Applications can easily implement this function directly, though, so it's not a big problem to have it absent from the library.)

kwonoj:
> this is more like a small extension to a feature that's already present

We  actually started to deprecate some of those extensions even. For example `mapTo` is deprecated on behalf of `map`. Our general recommendation is if a feature can be implemented trivially via existing operator it'd better to be used in each app / or userland. The reason core api surface is not changing rapidly is since we already have user's relying on those features which makes deprecation / breaking changes really hard. That's also one reason core team attempts to not to add new api surface for the features can be composed with existing operators.

EarthyOrange:
@kwonoj I have been implementing operators in my library and would have preferred to do the same for this operator as well. But looking at the implementation of the combineAll operator, it looked like I needed to have access to some internals like OperatorSubscriber and executeSchedule methods. If there is a replacement of these internals, that is already present in the exposed APIs, I will be happy to use those and close this PR.

From what I further understand from your comment is that the direction of this library development now is towards making it a framework which can be used to implement the operators in user libraries or perhaps even make the current operators available in a separate library which would rely on this core library. Is that right?

benlesh:
Hi @Wyzard256 — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

PR #6790, opened 2022-02-02
fix: remove broken types reference in package.json

corinnaerin:
**Description:**

The `types` field in `package.json` references a type definition file that does not exist. This causes `eslint-plugin-import` to fail when it tries to resolve the reference: 

image

`example.ts` looks like this:

```typescript
import { BehaviorSubject } from "rxjs";

const a = new BehaviorSubject(1);
console.log(a.getValue());
```

TypeScript successfully resolves the package, but `eslint-plugin-import` does not (I'm using the latest version of this package). Removing the broken `types` reference fixes the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,6 @@
   "main": "./dist/cjs/index.js",
   "module": "./dist/esm5/index.js",
   "es2015": "./dist/esm/index.js",
-  "types": "index.d.ts",
   "typesVersions": {
     ">=4.2": {
       "*": [


kwonoj:
This is dupe of https://github.com/ReactiveX/rxjs/issues/6466#issuecomment-859186944

and its comments to upstream comment https://github.com/microsoft/TypeScript/issues/43034#issuecomment-823423227 that we are intentionally following TypeScript core team's suggestion for supporting type resolutions with supported typescript version.

corinnaerin:
Oh ok sorry! I did a search of past issues, I guess I didn't search thoroughly enough.

I will open a bug report with `eslint-plugin-import` instead for them to support this use case.

PR #6797, opened 2022-02-04
Fix 6783 patched binds v2

benlesh:
This just has an alternative approach to #6796 ..

So we can choose.


cartant:
I think this version highlights that this is not really a reasonable change. I mean, this is literally saying "don't call `bind` on the passed observer methods because it's possible they could have an non-standard/broken `bind` implementation".

kwonoj:
I shared my opinion in original issue already, I'm still bit failed to understand if there's reason rxjs should provide its core implementation for the cases of user / application breaks spec behavior of javascript itself. I know we _can_, but reason why is not I'm fully getting yet.

getify:
I explained the *WHY* here: https://github.com/ReactiveX/rxjs/issues/6783#issuecomment-1030491106

cartant (approved):
Having read the detailed use case in [this comment](https://github.com/ReactiveX/rxjs/issues/6783#issuecomment-1030492906), I'm in favour of this version of the change. However, I think it would benefit from a having a comment added explaining that `Function.prototype.bind` is used directly for compatibility with Monio. Maybe a link to Kyle's comment?

PR #6800, opened 2022-02-07
chore: isolated module set to true, exports split to export type

ziv:
**Description:**

In order to be more compatible with ESM and other bundlers or TypeScript environment (Babel and Deno, for example), **isolatedModules** flag should be turned on.

This PR contain only this change in **tsconfig** and fix some build error (replace "export" with "export type" for interfaces).
Reference:
https://www.typescriptlang.org/tsconfig#isolatedModules

All tests and build pass.


jakovljevic-mladen:
Hi @ziv, unfortunately, we already had an attempt to introduce `export type` syntax to the repo, but it failed due to docs app being unable to generate documentation for these types (`GlobalConfig`, `RetryConfig` and all others). Please take a look at [this PR](https://github.com/ReactiveX/rxjs/pull/6895) (and [this comment](https://github.com/ReactiveX/rxjs/pull/6895#issuecomment-1090749387)) and the [reverting PR](https://github.com/ReactiveX/rxjs/pull/6943).

I'd close this PR, but since it also includes `isolatedModules: true` change, I'll check with the core team how/should we include this.

benlesh:
Closing this. It doesn't seem like this is solving any issues we're currently having. Thank you for your contribution, @ziv .

Issue #6803, opened 2022-02-07
Productize operator creation internals


benlesh:
For version 8, we should try to productize our operator creation methods to give people something more terse, and perhaps more safe, to create their own operators.

benlesh:
This is effectively in place now for 8.x

PR #6804, opened 2022-02-07
refactor: add `createOperatorSubscriber` abstraction

benlesh:
Simply hides the creation of `new OperatorSubscriber` in a function called `createOperatorSubscriber`.

This is part of a larger plan to:

1. Productize our internal means of creating operators so other folks can use it. (this is just the first step, I do not plan on exposing `createOperatorSubscriber` at this time). The idea is to move away from using classes all over our operators for both minification reasons and because then later, when we productize things, we don't have to export classes that people will try to subclass.
2. Make it easier to refactor the class underpinning our operators by abstracting it away

NOTE: This does not attempt to change anything about the special case in `groupBy`, which will require some thought.

Related: #6803

### DIFF
--- a/src/internal/observable/ConnectableObservable.ts
+++ b/src/internal/observable/ConnectableObservable.ts
@@ -3,7 +3,7 @@ import { Observable } from '../Observable';
 import { Subscriber } from '../Subscriber';
 import { Subscription } from '../Subscription';
 import { refCount as higherOrderRefCount } from '../operators/refCount';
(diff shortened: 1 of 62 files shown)


cartant (approved):
LGTM. Can we remove the export from `export class OperatorSubscriber`? (Reviewing this on my phone, so it's not easy to check whether the export is really needed, now.)

benlesh:
@cartant not yet, we need to merge #6806 first, and update this a bit.

PR #6815, opened 2022-02-08
refactor(SafeSubscriber): optimize perf for ordinary observers

benlesh:
No longer require function binding if we aren't using the deprecated next context. This should improve performance in the common path of consumers subscribing with an object or even with a function.

Adds a simple class `ConsumerObserver` which is mostly meant to optimize the number of function refrences created. We should never expose this externally.

Related #6783

### DIFF
--- a/src/internal/Subscriber.ts
+++ b/src/internal/Subscriber.ts
@@ -155,9 +155,10 @@ class ConsumerObserver<T> implements Observer<T> {
   constructor(private partialObserver: Partial<Observer<T>>) {}
 
   next(value: T): void {
-    if (this.partialObserver.next) {
+    const { partialObserver } = this;
+    if (partialObserver.next) {
       try {
-        this.partialObserver.next(value);
+        partialObserver.next(value);
       } catch (error) {
         handleUnhandledError(error);
       }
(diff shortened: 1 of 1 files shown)


src/internal/Subscriber.ts
```diff
@@ -155,65 +195,52 @@ export class SafeSubscriber<T> extends Subscriber<T> {
...
         context = Object.create(observerOrNext);
         context.unsubscribe = () => this.unsubscribe();
+        partialObserver = {
+          next: observerOrNext.next && bind(observerOrNext.next, context),
+          error: observerOrNext.error && bind(observerOrNext.error, context),
+          complete: observerOrNext.complete && bind(observerOrNext.complete, context),
+        };
       } else {
-        context = observerOrNext;
+        partialObserver = observerOrNext;
```
benlesh, line 226: any other partial observer passed is used as-is.

benlesh:
I don't have any performance metrics on this, unfortunately, nor do I have time to create them. But I have no doubt that using the passed partial observer directly will be an order of magnitude faster than pulling its functions off and binding them before using them.

src/internal/Subscriber.ts
```diff
@@ -147,6 +147,46 @@ function bind<Fn extends (...args: any[]) => any>(fn: Fn, thisArg: any): Fn {
...
+/**
+ * Internal optimization only, DO NOT EXPOSE.
+ * @internal
+ */
+class ConsumerObserver<T> implements Observer<T> {
+  constructor(private partialObserver: Partial<Observer<T>>) {}
+
+  next(value: T): void {
+    if (this.partialObserver.next) {
+      try {
```
cartant, line 159: You could access the `partialObserver` via `this` only once here: `const { partialObserver } = this` or whatever. There's no need to access it twice. Same for the other methods.

cartant (approved):

Issue #6786, opened 2022-01-28
Monorepo and Splitting Out To Different Packages


benlesh:
## Goals

1. We've wanted to move to a monorepo for quite some time. This is so we can move the docs app out of the project for core components, and possible move things into our monorepo to ensure compatability &mdash; for example chrome tools or the eslint plugin.
2. We've discussed many times splitting `Observable` off to be a standalone package.
3. We may want to consider splitting `rxjs/testing`, `rxjs/ajax`, etc, off into their own packages as well.
4. We want to do this in the slowest, most easy to migrate to way possible.

## Monorepo and multiple packages

* I think we should try to button up new feature work and the like for version 7 (which we are very close to doing, if we're not done already) and move to work on version 8 in earnest.
* In version 8, we'll keep the same exports (with regards to `rxjs`, `rxjs/operators`, `rxjs/ajax`, et al) but split all of those off into their own packages under `@rxjs/observable`, `@rxjs/operators`, `@rxjs/ajax`, etc. We'll want to bikeshed ideas like `@rxjs/subjects` or `@rxjs/multicasting` etc. There's a lot of ground to cover there, and I don't want to get bogged down on that in this discussion.
* In version 8 we're going to reexport from the new packages. For example, `rxjs` will `export { Observable } from '@rxjs/observable';` etc.

## Standalone Observable

* Moving away from `lift`: If we want a standalone `Observable`, we should move away from lift, because it requires too much knowledge of the inner working of RxJS in order to create custom operators that leverage it properly.
* We can start supporting Observable that doesn't have `lift` as a non-breaking change anytime. This is completely contained within our `operate` utility function now. We may even start supporting this in v7, as it should be completely innocuous. It's just a different call pattern. I actually have some work done around this, and everything looks solid.
* I think we're going to need to leave `pipe` on our standalone observable package, as it's central to our particular design for how to use RxJS. Even though `pipe` might not be immediately useful to users of `Observable` that don't care about operators. (libraries that export `Observable`, etc).

## Long Term:

1. We wouldn't want to force people to move over to `@rxjs/observable` et al until version 9. That should be a long ways off.
2. Operator creation needs to be consumerized. We have our out internal way of creating operators, and we need to come up with a way of making that consumable and understandable to the masses. It has some advantages, like properly unsubscribing from firehose inner observables. (Currently it's impossible to do this properly with plan observer implementations) (A token/signal approach would help this)
3. Considerations around using abort signals, etc.

[...]

PR #6816, opened 2022-02-08
feat: Allow operators to operate on any Subscribable without lift

benlesh:
* Adds a code path for operating on a Subscribable without lift.
* Loosens the type of `OperatorFunction` and therefore `MonoTypeOperatorFunction` to go from `Subscribable<In>` to `Subscribable<Out>`.

This is done in preparation for splitting `Observable` into its own package, where `lift` doesn't really need to exist in the public API.

Related #6803 #6786

### DIFF
--- a/spec/operators/non-lifted-spec.ts
+++ b/spec/operators/non-lifted-spec.ts
@@ -1,0 +1,79 @@
+/** @prettier */
+import { expect } from 'chai';
+import { Observer, pipe, Subject, Unsubscribable, map, filter, take } from 'rxjs';
+
+/**
+ * An observable that does not have lift.
+ */
+class PrimitiveObservableFixture<T> {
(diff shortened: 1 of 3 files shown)


benlesh:
NOTE: I'm not at all committed to this approach. This is just something I wanted to get a draft out of while I was thinking about it. More as a spike than anything.

We could just as well handle this by allowing `from` to convert `Subscribable`: `from(otherSource).pipe(map, filter, etc)`.

The real thing I'm considering is whether or not it's good to have `lift` in the public API of an observable published without operators. Perhaps the real play here is to split off `Observable` and have a special `LiftedObservable` that we use for operators, which is fine once we productize how we create operators (#6803).

benlesh:
Actually, the more I think about it, the more I think I want to take another approach. 🤔

PR #6817, opened 2022-02-08
Subscriber refactor for version 8

benlesh:
This is completely based off of #6815, but taken with a commit-by-commit refactor of related pieces.

Highlights:

1. `Subscriber` constructor is no longer deprecated. It can be used to create a "safe subscriber" like `new Subscriber(nextHandler)` or `new Subscriber({ next, error })`, etc.
2. Long-deprecated `Subscriber.create` API has been removed.
3. `SafeSubscriber` removed. Everything is now simplified.

**NOTE: Please review one commit at a time. I tried to keep the individual commits small and focused.**


src/internal/Subscriber.ts
```diff
@@ -135,66 +121,46 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
   }
 }
 
-/**
- * This bind is captured here because we want to be able to have
- * compatibility with monoid libraries that tend to use a method named
- * `bind`. In particular, a library called Monio requires this.
- */
-const _bind = Function.prototype.bind;
+class ConsumerObserver<T> implements Observer<T> {
```
benlesh, line 124: This is from #6815

src/internal/Subscriber.ts
```diff
@@ -135,66 +121,46 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
...
-
-export class SafeSubscriber<T> extends Subscriber<T> {
-  constructor(observerOrNext?: Partial<Observer<T>> | ((value: T) => void) | null) {
-    super();
+  next(value: T): void {
+    if (this.partialObserver.next) {
+      try {
+        this.partialObserver.next(value);
+      } catch (error) {
+        reportUnhandledError(error);
```
benlesh, line 132: Note the difference here because we don't have to worry about deprecated error behavior.

src/internal/Subscriber.ts
```diff
@@ -226,3 +192,11 @@ export const EMPTY_OBSERVER: Readonly<Observer<any>> & { closed: true } = {
   error: defaultErrorHandler,
   complete: noop,
 };
+
+function isObserver<T>(value: any): value is Observer<T> {
+  return value && isFunction(value.next) && isFunction(value.error) && isFunction(value.complete);
+}
+
+export function isSubscriber<T>(value: any): value is Subscriber<T> {
```
benlesh, line 200: These were just moved.

src/internal/operators/expand.ts
```diff
@@ -82,7 +82,9 @@ export function expand<T, O extends ObservableInput<unknown>>(
       // General merge params
       source,
       subscriber,
-      project,
+
+      // HACK: Cast because TypeScript seems to get confused here.
+      project as (value: T, index: number) => ObservableInput<ObservedValueOf<O>>,
```
benlesh, line 87: This was super weird, but seems like an inference quirk with TypeScript.

src/internal/operators/ignoreElements.ts
```diff
@@ -40,6 +41,6 @@ import { noop } from '../util/noop';
  */
 export function ignoreElements(): OperatorFunction<unknown, never> {
   return operate((source, subscriber) => {
-    source.subscribe(createOperatorSubscriber(subscriber, noop));
+    source.subscribe(createOperatorSubscriber(subscriber as Subscriber<unknown>, noop));
```
benlesh, line 44: `Subscriber<never>` doesn't match `Subscriber<any>`.

src/internal/Subscriber.ts
```diff
@@ -18,45 +18,31 @@ import { timeoutProvider } from './scheduler/timeoutProvider';
...
-    if (destination) {
-      this.destination = destination;
-      // Automatically chain subscriptions together here.
-      // if destination is a Subscription, then it is a Subscriber.
-      if (isSubscription(destination)) {
-        destination.add(this);
-      }
-    } else {
-      this.destination = EMPTY_OBSERVER;
+    this.destination = isSubscriber(destination) ? destination : createSafeObserver(destination);
```
benlesh, line 40: important magic here!

If we have a `Subscriber`, then just use that, otherwise we don't know what it is, and we need to make a safe observer.

If you step through the commits, you find out that this is all `SafeSubscriber` was doing, which is why we removed it.

src/internal/Subscriber.ts
```diff
@@ -135,66 +121,46 @@ export class Subscriber<T> extends Subscription implements Observer<T> {
...
- * compatibility with monoid libraries that tend to use a method named
- * `bind`. In particular, a library called Monio requires this.
- */
-const _bind = Function.prototype.bind;
+class ConsumerObserver<T> implements Observer<T> {
+  constructor(private partialObserver: Partial<Observer<T>>) {}
 
-function bind<Fn extends (...args: any[]) => any>(fn: Fn, thisArg: any): Fn {
-  return _bind.call(fn, thisArg);
-}
```
benlesh, line 147: The bind stuff can be removed! Yay.

src/internal/Subscriber.ts
```diff
@@ -18,45 +18,31 @@ import { timeoutProvider } from './scheduler/timeoutProvider';
...
-  }
-
   /** @deprecated Internal implementation detail, do not use directly. Will be made internal in v8. */
   protected isStopped: boolean = false;
   /** @deprecated Internal implementation detail, do not use directly. Will be made internal in v8. */
-  protected destination: Subscriber<any> | Observer<any>; // this `any` is the escape hatch to erase extra type param (e.g. R)
+  protected destination: Subscriber<T> | Observer<T>; // this `any` is the escape hatch to erase extra type param (e.g. R)
 
   /**
-   * @deprecated Internal implementation detail, do not use directly. Will be made internal in v8.
```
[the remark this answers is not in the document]
benlesh: 🤔 🤔 🤔 You're probably right. I hate subclassing. That said, `Subscriber` is exposed anyhow, at least as a type. 🤔 🤔 🤔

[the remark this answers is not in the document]
benlesh: We'll need to address preventing `Subscriber` subclassing in a follow up.

cartant (approved):

Issue #6764, opened 2022-01-13
debounceTime does not emit when queueScheduler is used


ghostlytalamaur:
### Describe the bug

If synchronous scheduler is used and there are no other tasks debounceTime does not emit value to subscriber.

### Expected behavior

debounceTime should work properly with queueScheduler as it was in v6
#### expected output
```emit!```
#### actual output
No emit at all

### Reproduction code

```typescript
import { BehaviorSubject, queueScheduler } from 'rxjs';
import { debounceTime } from 'rxjs/operators';

const data$ = new BehaviorSubject('emit!');

data$.pipe(debounceTime(0, queueScheduler)).subscribe(console.log);
```

### Reproduction URL

https://stackblitz.com/edit/q7u9jg-hykb4x

### Version

7.5.1

PR #6826, opened 2022-02-11
fix(debounceTime): handle synchronous schedulers

dancras:
**Description:**

Using `debounceTime(0, queueScheduler)` was not emitting any values. `scheduler.schedule()` emits immediately but logic in `debounceTime` was expecting it to always return before emitting.

Fixes https://github.com/ReactiveX/rxjs/issues/6764

### DIFF
--- a/spec/operators/debounceTime-spec.ts
+++ b/spec/operators/debounceTime-spec.ts
@@ -1,6 +1,6 @@
 /** @prettier */
 import { expect } from 'chai';
-import { NEVER, of, Subject } from 'rxjs';
+import { NEVER, of, Subject, queueScheduler } from 'rxjs';
 import { AnimationFrameAction } from 'rxjs/internal/scheduler/AnimationFrameAction';
 import { AnimationFrameScheduler } from 'rxjs/internal/scheduler/AnimationFrameScheduler';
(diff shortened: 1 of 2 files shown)


benlesh:
I see the issue.

I think that a cleaner solution would be to create a `Subscription` prior to the `subscriber.schedule` call, and leverage that, something like:

```ts
if (!activeTask) {
  subscriber.add(activeTask = new Subscription());
  activeTask.add(scheduler.schedule(emitWhenIdle, dueTime));
}
```

It's perhaps a little more resource intensive per schedule, but it's certainly less complicated.

Let me know if that works out.

dancras:
That is a nice solution. Can confirm it works and cleans up the scheduler.schedule subscription even after the fact in the synchronous case. Learned a little more about Subscriptions, thanks.

benlesh:
@dancras Apologies on dropping the ball on landing this one. I think there may be another solution here, and we might still try to get this into the 7.x branch.

benlesh:
I've added a commit and moved it over here. I'll try to make sure you get credit for the commits, @dancras

https://github.com/ReactiveX/rxjs/pull/7161

PR #6832, opened 2022-02-18
fix(interop): allow Rx Observable to be consumed as InteropObservable at type level

### LINKED ISSUE #6777: Symbol.observable is missing from the Observable class on the type level
### Describe the bug

RxJS can **consume** InteropObservables that implement `Symbol.observable` method but it doesn't allow for reverse - it doesn't make its own observables interop (at the type level) so they can't be consumed by other libs that would like to leverage `Symbol.observable` for this purpose.

### Expected behavior

It should be possible to create a `from` method that would rely on the input argument having a `Symbol.observable` method and convert Rx observable to a different observable type.

### Reproduction code

```typescript
import { Observable, interval, InteropObservable, Subscribable } from "rxjs"; // types: 7.5.2

declare const obs1: Observable<number>;
const obs2 = interval(100);

type Test1 = typeof obs1 extends InteropObservable<infer T> ? T : never;
//   ^?
type Test2 = typeof obs2 extends InteropObservable<infer T> ? T : never;
//   ^?

declare class MyObservable<T> {
  subscribe(cb: (a: T) => void): void;
  // just make this class nominal to avoid relying on structural subtyping with the Observable from RxJS
  private tag: string;
}

// strategy based on the one in RxJS itself
// https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/src/internal/types.ts#L90
type ObservableInput<T> = MyObservable<T> | InteropObservable<T>;

// https://github.com/ReactiveX/rxjs/blob/5d0e41ee46337f56dba6d9a1ef938002b261ccac/src/internal/types.ts
type ObservedValueOf<O> = O extends ObservableInput<infer T> ? T : never;

// https://github.com/ReactiveX/rxjs/blob/888c753e272443bbb879f9ffb1a8145a7773c865/src/internal/observable/from.ts#L6
declare function from<O extends ObservableInput<any>>(
  input: O
): MyObservable<ObservedValueOf<O>>;

// actual
const converted1 = from(obs1);
const converted2 = from(obs2);

// expected result (without having to cast this)
const converted3 = from(obs1 as typeof obs1 & { [Symbol.observable]: () => Subscribable<number> });
//    ^?
const converted4 = from(obs2 as typeof obs2 & { [Symbol.observable]: () => Subscribable<number> });
//    ^?
```

### Reproduction URL

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgeQEYGcCmUBuBDVAG0wBo5gA7GbfQsgSSuwjDS1wOLIGUBXDAMZRgqTpjgBfOADMoEEHABEUAB4ArdIoDccAPS64MAJ5hM6AFxwA7ADoArDYBMAKGcATTAMJ4o4gRAp0eAgMAEZLNhoxAB4KXhBUbAA+LWd-QOCMRzgAXnImDkIAClCABlKASlTnY1M4ABUzGFDcwxNMCGk4EPQWzBVqCjd0OEZqOVYMKKJMaMppbAakuAB+BrhLCkwcbFT9OAOAPRWa9oam7Lzajq6e7P7B4dGClkiOGbmKBagl1fXN7a7Zz7I4ndyeby+OBePDoEYAWSMb3wH3qywQzgO6H46CEIkwRQEqEsRTwlnqFVyyxwEGAbgqlhpdNSB32al4QTgIDwAGtxDAABbAEYwuFwCjySh4QiGCBwPBMtxwXyEIyUADm3QocCCUF4AhgvCg0p1-FqGrgAHdgILDALxMixDI5AoAEoqABS3ExcDAwnw1EMeHVll1GtSElc+11eGo6qMcFEWCVATt4gC4kocHdXvIMCwhGkwIMApgMDAFn06ptAv4Nn8IF0rsweANwB2AA1dKoNLoiCFdHY3KVMAAWUKYMcANgAzDOrNI7FO3KJlwBOPAT6RrmcADnKjlQjinoQEAlbunQUAEuko4wo0t013QNnzAGIADJr0qnOqOmaMGAvAwNEaKtIi-7EKBywAD7POMrxTO8UFotU+yluWla6NWgp1g2TYtm2nbduo6B9oQA5DiO46TqOs7zouy6rm4G5bju+6lIex6nueN5Xjed7YA+hBPu0L75r+DpITsbgAGrSrwmDINI0TIMseTIHADyYEMIyQZggHAZ83y-Gs9QbOKgJQGhJZlhW5hVjWeHyARrYwO2mBdj2ZH9qgui7gFAhWHYM6YI4ViOKOo4zqgsW7lYa7btI0ioKEeC7qEo52HgVi5TOAi7lOdiXtet4FMJug9NMxC6LI8ivugn5TuCMJQtIvAUG2qZ1SAqlaQMOlPPphkgXgFBGEkSRFD6lBATAETOAycAQdJMTIpgckKUpKlqSkUYGG5vDSmkAScukOywBtLR5D1RQ9KEVQnRk0IBBd1BuJczryHdWSPcW-WmAaG3KmYvCEPARTWoKEDAXAAoKhaMByuenKCsKFRPWdr3YO9M6tLd93yiM1ydN0YRwAAZIgcAANrcEYCQQIQNhVchmAALokpSOTLHwgjCKIHxxAkySSI9IKgpj8DnTjG2jvjLo-eg2Swm0pik3clPU3TDOoEzLOrTMnNwEU3O8zieKC1BwuJFAywSOLBgHHAxzOEAA

### Version

7.5.2

### Environment

_No response_

### Additional context

_No response_

Andarist:
**Description:**

This fixes the interop story for other libraries that would like to consume Rx as interop observable. This weird, weird structure ensures that the `Symbol.observable` type is emitted correctly to the `.d.ts` file - it's actually not emitted directly but this computed property is left in the class declaration and its type is `Symbol.observable`, so the whole thing works.

I got a hint that this might work from [this comment](https://github.com/microsoft/TypeScript/issues/42601#issuecomment-771582430) in the TS repo - since it has mentioned "dotted names". So I've figured out that it's worth a try here and, somewhat, surprisingly I've managed to get this working.

**Related issue (if exists):** fixes #6777

### DIFF
--- a/package.json
+++ b/package.json
@@ -201,7 +201,7 @@
     "tslint-etc": "1.13.10",
     "tslint-no-toplevel-property-access": "0.0.2",
     "tslint-no-unused-expression-chai": "0.0.3",
-    "typescript": "~4.2.0",
+    "typescript": "~4.3.0",
     "validate-commit-msg": "2.14.0",
     "web-streams-polyfill": "^3.0.2",
     "webpack": "^4.31.0"
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -2,7 +2,7 @@ import { Operator } from './Operator';
 import { SafeSubscriber, Subscriber } from './Subscriber';
 import { isSubscription, Subscription } from './Subscription';
 import { TeardownLogic, OperatorFunction, Subscribable, Observer } from './types';
-import { observableTypeRef } from './symbol/observable';
+import { observable as Symbol_observable } from './symbol/observable';
 import { pipeFromArray } from './util/pipe';
 import { config } from './config';
 import { isFunction } from './util/isFunction';
@@ -340,7 +340,7 @@ export class Observable<T> implements Subscribable<T> {
    * @method Symbol.observable
    * @return {Observable} this instance of the observable
    */
-  [observableTypeRef.symbol](): Subscribable<T> {
+  [Symbol_observable](): Subscribable<T> {
     return this;
   }
 
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -2,7 +2,7 @@ import { Operator } from './Operator';
 import { SafeSubscriber, Subscriber } from './Subscriber';
 import { isSubscription, Subscription } from './Subscription';
 import { TeardownLogic, OperatorFunction, Subscribable, Observer } from './types';
-import { observable as Symbol_observable } from './symbol/observable';
+import { observableTypeRef } from './symbol/observable';
 import { pipeFromArray } from './util/pipe';
 import { config } from './config';
 import { isFunction } from './util/isFunction';
@@ -340,7 +340,7 @@ export class Observable<T> implements Subscribable<T> {
    * @method Symbol.observable
    * @return {Observable} this instance of the observable
    */
-  [Symbol_observable]() {
+  [observableTypeRef.symbol](): Subscribable<T> {
     return this;
   }
 
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -340,6 +340,7 @@ export class Observable<T> implements Subscribable<T> {
    * @method Symbol.observable
    * @return {Observable} this instance of the observable
    */
+  /** @ts-ignore mute errors for people without skipLibCheck on older versions of TS, remove this comment when we drop support for TS<4.3 */
   [Symbol_observable](): Subscribable<T> {
     return this;
   }
--- a/src/internal/symbol/observable.ts
+++ b/src/internal/symbol/observable.ts
@@ -1,2 +1,5 @@
 /** Symbol.observable or a string "@@observable". Used for interop */
-export const observable: string | symbol = (() => (typeof Symbol === 'function' && Symbol.observable) || '@@observable')();
+export const observable = (() => (typeof Symbol === 'function' && Symbol.observable) || '@@observable')();
+export const observableTypeRef = {
+  symbol: observable as Exclude<typeof observable, string>,
+};
--- a/src/internal/symbol/observable.ts
+++ b/src/internal/symbol/observable.ts
@@ -1,5 +1,2 @@
 /** Symbol.observable or a string "@@observable". Used for interop */
-export const observable = (() => (typeof Symbol === 'function' && Symbol.observable) || '@@observable')();
-export const observableTypeRef = {
-  symbol: observable as Exclude<typeof observable, string>,
-};
+export const observable: typeof Symbol.observable = (() => (typeof Symbol === 'function' && Symbol.observable) || '@@observable')() as any;
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -431,13 +431,13 @@ export interface ObjectUnsubscribedError extends Error {
 
 export declare const ObjectUnsubscribedError: ObjectUnsubscribedErrorCtor;
 
-export declare const observable: string | typeof Symbol.observable;
+export declare const observable: typeof Symbol.observable;
 
 export declare class Observable<T> implements Subscribable<T> {
     operator: Operator<any, T> | undefined;
     source: Observable<any> | undefined;
     constructor(subscribe?: (this: Observable<T>, subscriber: Subscriber<T>) => TeardownLogic);
-    [observableTypeRef.symbol](): Subscribable<T>;
+    [Symbol_observable](): Subscribable<T>;
     forEach(next: (value: T) => void): Promise<void>;
     forEach(next: (value: T) => void, promiseCtor: PromiseConstructorLike): Promise<void>;
     lift<R>(operator?: Operator<T, R>): Observable<R>;
--- a/api_guard/dist/types/index.d.ts
+++ b/api_guard/dist/types/index.d.ts
@@ -431,12 +431,13 @@ export interface ObjectUnsubscribedError extends Error {
 
 export declare const ObjectUnsubscribedError: ObjectUnsubscribedErrorCtor;
 
-export declare const observable: string | symbol;
+export declare const observable: string | typeof Symbol.observable;
 
 export declare class Observable<T> implements Subscribable<T> {
     operator: Operator<any, T> | undefined;
     source: Observable<any> | undefined;
     constructor(subscribe?: (this: Observable<T>, subscriber: Subscriber<T>) => TeardownLogic);
+    [observableTypeRef.symbol](): Subscribable<T>;
     forEach(next: (value: T) => void): Promise<void>;
     forEach(next: (value: T) => void, promiseCtor: PromiseConstructorLike): Promise<void>;
     lift<R>(operator?: Operator<T, R>): Observable<R>;
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,13065 @@
 {
   "name": "rxjs",
-  "version": "7.5.2",
-  "lockfileVersion": 1,
+  "version": "7.5.4",
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "rxjs",
+      "version": "7.5.4",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "tslib": "^2.1.0"
+      },
+      "devDependencies": {
+        "@angular-devkit/build-optimizer": "0.4.6",
+        "@angular-devkit/schematics": "^11.0.7",
+        "@swc/core": "^1.2.128",
+        "@swc/helpers": "^0.3.2",
+        "@types/chai": "^4.2.11",
+        "@types/lodash": "4.14.102",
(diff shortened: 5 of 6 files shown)


src/internal/Observable.ts
```diff
@@ -340,7 +340,7 @@ export class Observable<T> implements Subscribable<T> {
    * @method Symbol.observable
    * @return {Observable} this instance of the observable
    */
-  [Symbol_observable]() {
+  [observableTypeRef.symbol](): Subscribable<T> {
```
Andarist, line 343: An explicit type annotation is required here because with the inferred `Observable<T>` TS can't infer correctly the type using `ObservedValueOf` and it ends up being `unknown`. I think that is caused by the fact that `Observable<T>["subscribe"]` is overloaded and TS is only using the last declared overload when processing a conditional type - so it fails to match the correct signature.

src/internal/symbol/observable.ts
```diff
@@ -1,2 +1,5 @@
 /** Symbol.observable or a string "@@observable". Used for interop */
-export const observable: string | symbol = (() => (typeof Symbol === 'function' && Symbol.observable) || '@@observable')();
+export const observable = (() => (typeof Symbol === 'function' && Symbol.observable) || '@@observable')();
+export const observableTypeRef = {
+  symbol: observable as Exclude<typeof observable, string>,
+};
```
Andarist, line 5: Yeah, this is weird - I don't know how to declare it in a better way that would preserve the `typeof Symbol.observable` for this. You could use this [TS playground](https://www.typescriptlang.org/play?ts=4.5.4#code/KYDwDg9gTgLgBAbwL4CgAmwDGAbAhlYOAc2wgCNdtEU44BLAOxmCgDNdNCBlATwFsyEbAGEIDAM4woAV0wxo1WrQK40Y7DzjlxALjjj+g7DTipUKTGMlxS1gLxxeAoQDptKAPQelAPQD8nt4A7nQwABYQ0vDiEHyEANaMaFqscLhwQVC4YGAscKzSDHJ0YlpQ+nzhcOHchq7a1Ty5xMAw4jYQkhZW0fxwDgAUAwCU-QB8cAMwTcAQqU5G-XYOAOQFRTAlDCtwAGS7jnXYbmTiowA+53ArAAI32iwAbrhk2MArwyMA3IG+Ad0SXp8YS4STAZIOAx8NLtACiIBw0gwAB5prk5vp+AAafRSRhEMYoAHWKEAdSyOTyDgQmL4eihMLg8MRKLRs1SUJxkig+ImqC8fyJoEgsDgOFB7QAgoo4ABtWwwAC6ehG40QZiF4Gg8HF4naACEZbKocrJqM7BNkChzMLtWK8Hq4MIjVCQWC0KbVRb1dbNSKdQ72gARF38cnZXJQFwmlXmy0alCMZhsDiEUmhMIASSYLAgYGRABVLSZZQt6qdPXG4AWfjataK2XBM+JMzBSdBEgwiP04CHQMwGGh2unwtnk3nkbgGDwJn44ABGOB6AAMvyU67g-iAA) to experiment with this more

Andarist: Actually, I've managed to simplify this by casting the IFFE's result to any and assigning to `: typeof Symbol.observable`. For some reason `as typeof Symbol.observable` doesn't work and this was what I was previously attempting to do.

src/internal/Observable.ts
```diff
@@ -340,7 +340,8 @@ export class Observable<T> implements Subscribable<T> {
    * @method Symbol.observable
    * @return {Observable} this instance of the observable
    */
-  [Symbol_observable]() {
+  /** @ts-ignore mute errors for people without skipLibCheck on older versions of TS, remove this comment when we drop support for TS<4.3 */
```
Andarist, line 343: this has to be in a separate comment and right above the computed property for this to work 🤷‍♂️

package.json
```diff
@@ -201,7 +201,7 @@
     "tslint-etc": "1.13.10",
     "tslint-no-toplevel-property-access": "0.0.2",
     "tslint-no-unused-expression-chai": "0.0.3",
-    "typescript": "~4.2.0",
+    "typescript": "~4.3.0",
```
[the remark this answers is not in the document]
benlesh: We need to be backwards compatible  with TS 4.2 or this is a breaking change. 🤔

benlesh:
Sorry I'm just getting to reviewing this. This is a big deal if things work. However we need to support TS 4.2 (not necessarily fixing this issue for 4.2, we just can't break compilation for users of 4.2) otherwise this has to land in 8.x

package.json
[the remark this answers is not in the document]
Andarist: This change was only required to **emit** different `.d.ts`. It shouldn't break TS 4.2 consumers because they only consume the emitted `.d.ts`.

Take a look at the 4.2 playground [here](https://www.typescriptlang.org/play?ts=4.2.3#code/PTAEHMFMDtIJwIYBdIBNQFcDOBLa5QAVAZVABYA6AZgFgAoSADwAcB7OJUAbwF8BueqkgBjADYI4kCKNYAjBKO71QK0HhRwAZgmFTiATwC2s1qIDCraFiRwMwpOyV1VL0JISpLo-aDlYAXKBYRiaiAs6qPPRRdEJiElLCltZBRgDqiMzM8IFcyqrBhoFI+tmsmqAGxqYUfuH89ExsHKBx4pKg8VhYoACCTjFN7JxtCZ3i3aAAQgONLMOtIu2JEz1mswzzLaMdXT0AIk4uANqFGQhZ8BSFALqBABQAlKAAvAB8mNBCmnho4S7MOCsZiBaAYYzwaL0dTwbS6UBpHBIAAWAEloBpgQAeQgfPIRFTHKqhWqyLB3UBPV4fQj1ej0ECgFDWPAEJJCel0JJWEavUCwADuoH2T05jJROB6AssAHJOLIpAhujhwNAELJRFI8ERSJQAEygJVqTjCBDQOWfIRwaxm9AoqTBDEIRhqaCgAAGe2F7qZyOQah6kiwbCsOA1Uk0jgUMgFrN9koZYAAAkgsABaJoiJAZuBAuBzZom5KcJW4VXqzWBREo9GY5hYjBfSA-WCoD4vVpisCe1begNBJA4USKABu8B8hjsyMwWArUgFyJgxs6rEMkB6Dl89rgoEBwPgg-XvtYGHAyIoOv7AEcMEipOBWKx0P6FXHtBhRNmHJrEBjQAulyRFc1w3VhdwkXB8AAGlAWQ8FQVkYMgJBhALBZuRSMEin5cEFR3DtUAoPdmHoIA). It works as expected. 

We can also check the same code with TS 4.3 [here](https://www.typescriptlang.org/play?ts=4.3.5#code/PTAEHMFMDtIJwIYBdIBNQFcDOBLa5QAVAZVABYA6AZgFgAoSADwAcB7OJUAbwF8BueqkgBjADYI4kCKNYAjBKO71QK0HhRwAZgmFTiATwC2s1qIDCraFiRwMwpOyV1VL0JISpLo-aDlYAXKBYRiaiAs6qPPRRdEJiElLCltZBRgDqiMzM8IFcyqrBhoFI+tmsmqAGxqYUfuH89ExsHKBx4pKg8VhYoACCTjFN7JxtCZ3i3aAAQgONLMOtIu2JEz1mswzzLaMdXT0AIk4uANqFGQhZ8BSFALqBABQAlKAAvAB8mNBCmnho4S7MOCsZiBaAYYzwaL0dTwbS6UBpHBIAAWAEloBpgQAeQgfPIRFTHKqhWqyLB3UBPV4fQj1ej0ECgFDWPAEJJCel0JJWEavUCwADuoH2T05jORrHYCH0AEImcicD0BewANZYObNTjclIIbo4cDQBCyUSQQKIlHozHMLEYL6QH6wVAfF6tTlDFrazhgor88GyeB81AUQHA+hAA) and verify that with this version this property can be understood in full and used as expected, allowing Rx observable to be treated as `InteropObservable`.

The only downside to the current state of this PR is that with TS 4.2 people might get errors because of this pattern if they dont use `skipLibCheck`. We can suppress it with `/** @ts-ignore */` though and that should make it compatible with all of the requirements.

Let me know what do you think and I will followup with any changes/answers as needed.

Andarist:
@benlesh as far as I know - this doesn't break anything for TS 4.2 consumers. It just doesn't fix that issue for them. The `@ts-ignore` comment looks suspicious but it's actually what makes it "safe" for TS 4.2. I could do further testing with this PR if you want.

benlesh:
Hi @Andarist — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

PR #6840, opened 2022-02-18
chore(groupBy): convert groupBy specs to run mode

tmair:
**Description:**
This PR converts `groupBy` operator tests to run mode. This should be the last operator test that was using run mode.

*Note*: The defaults for `runMode` for the static marble parsing utilities in `TestScheduler` was changed to `true` because the corresponding methods are only used for the `groupBy` tests.

None

### DIFF
--- a/src/internal/testing/TestScheduler.ts
+++ b/src/internal/testing/TestScheduler.ts
@@ -221,7 +221,7 @@ export class TestScheduler extends VirtualTimeScheduler {
   }
 
   /** @nocollapse */
-  static parseMarblesAsSubscriptions(marbles: string | null, runMode = true): SubscriptionLog {
+  static parseMarblesAsSubscriptions(marbles: string | null, runMode = false): SubscriptionLog {
     if (typeof marbles !== 'string') {
       return new SubscriptionLog(Infinity);
     }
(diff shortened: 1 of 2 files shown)


benlesh (approved):

src/internal/testing/TestScheduler.ts
```diff
@@ -221,7 +221,7 @@ export class TestScheduler extends VirtualTimeScheduler {
   }
 
   /** @nocollapse */
-  static parseMarblesAsSubscriptions(marbles: string | null, runMode = false): SubscriptionLog {
+  static parseMarblesAsSubscriptions(marbles: string | null, runMode = true): SubscriptionLog {
```
cartant, line 224: Unfortunately, I don't think we should change this default, as the method is part of the `TestScheduler`'s public API. It's not marked as internal, so consumers will have it in their TypeScript definition and could be calling it. It should be marked as internal, but it isn't. Same goes for the `parseMarbles` method.

jakovljevic-mladen: Yeah, other libs depend on this, for example: https://github.com/timdp/swirly/blob/de269cbb93c2f82f98115347cb1ec2e42db3001d/packages/swirly-parser-rxjs/src/index.ts#L4

tmair: I did not think of that. I changed the defaults back to how they where and am setting the `runMode` parameter from within the tests itself.

benlesh: Good catch, @cartant, I missed that entirely.

benlesh (approved):

PR #6851, opened 2022-03-01
feat(shareReplay): use another observable to control resets

Evertt:
This feature, first introduced for the `share` operator, is equally as desirable for the `shareReplay` operator. It allows one to, for example, set up a delayed reset. For Observables that are a little expensive to set up, but not so expensive that it makes sense to keep them around forever.

### DIFF
--- a/src/internal/operators/shareReplay.ts
+++ b/src/internal/operators/shareReplay.ts
@@ -1,3 +1,4 @@
+import { Observable } from '../Observable';
 import { ReplaySubject } from '../ReplaySubject';
 import { MonoTypeOperatorFunction, SchedulerLike } from '../types';
 import { share } from './share';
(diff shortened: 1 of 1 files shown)


benlesh:
I think I'd rather guide users to just use `share`. IMHO, we should probably deprecate `shareReplay` at some point in the future.

Evertt:
> we should probably deprecate `shareReplay` at some point in the future

Really? Why? It's such a pleasant operator to have.

diginikkari:
I was just pumping to this and wondered how to solve it. I found refCount observable to be very handy when playing with suspense with react. If anyone else pumps in to this, here is how I used `share`-operator instead of `shareReplay` to manage this:
```typescript
      observable.pipe(
      share({
        connector: () => new ReplaySubject(1),
        resetOnRefCountZero: () => timer(1000),
      }))
```

Issue #6367, opened 2021-05-05
Version 7 -> 8 Roadmap


benlesh:
# TBD (not actionable yet)

This issue is to come up with what the roadmap is for version 7-8.

### In Version 7

- [ ] `groupBy` config (#5679)
- [ ] Add all exports to top level
- [ ] ??? TBD

### In Version 8

- [ ] Remove _all_ deprecated APIs.
- [ ] Drop all Internet Explorer support into a pit of flames! 🔥 IE 🔥 🥳 
- [ ] ??? TBD

criskrzysiu:
🔥 IE 🔥 🥳 - every time I see that statement in some repo it makes my day better

MaximeBernard:
Since "??? TBD" seems to imply we can suggest anything, what about https://github.com/ReactiveX/rxjs/issues/2686?

It seems to be a recurring "issue" people are having to configure headers for every request (most popular seem to be `Bearer`, `CorrelationId` and `Content-Type`).

While it's not "hard" to make your own `ajax`, I feel it's a bit sad everyone has to write that piece of code and it might be helpful for everyone to have an helper provided by `rxjs`.

PowerKiKi:
Wouldn't publishing multiple packages for what is essentially the same code actually increase risk of non tree shakable use-case and thus increase bundle size ?

As seen with lodash, if one of my dependency decides to depends on the future `@rxjs/observable`, but my app depends on the current `rxjs`, then all the code from `@rxjs/observable` will essentially be bundled twice, won't it ?

And of course it would be much worse if we actually publish a package for each operator. Then a single operator could be bundled three times if my deps happened to depend on `rxjs`, `@rxjs/operators` and say `@rxjs/operators-map`.

I could also imagine compatibility issues if one of my dependency depends on an old version of  `@rxjs/observable` and my app pass one of those object to a new, incompatible version of an operator coming from the latest version of `rxjs`. This kind of issue already happened with the `graphql` package where they had to add manual check in their code to throw error if it happens.

That seems a lot of tricky tradeoffs to attempt to solve the issue of bundle size that is already properly solved by tree shaking.

Is that really worth it ? :thinking:

alfaproject:
It looks like the items for version 7 are all done except one. Is there any progress? Feels a bit stagnated ):

akakira:
Do you honestly think this signature change on the subscribe, etc. methods is critical enough to cause billions of lines of code change across the internet?

[...]

PR #6859, opened 2022-03-04
Deprecate retryWhen and repeatWhen

benlesh:
Deprecates `retryWhen` and `repeatWhen`, as the same can be accomplished in a MUCH simpler way with the `delay` argument to both `retry` and `repeat`. Not to be removed until v9 or v10. (A long-long time away, still).

Related to #6367.

### DIFF
--- a/spec-dtslint/operators/repeatWhen-spec.ts
+++ b/spec-dtslint/operators/repeatWhen-spec.ts
@@ -1,0 +1,22 @@
+import { of } from 'rxjs';
+import { repeatWhen } from 'rxjs/operators';
+
+it('should infer correctly', () => {
+  const o = of(1, 2, 3).pipe(repeatWhen(errors => errors)); // $ExpectType Observable<number>
+});
+
+it('should infer correctly when the error observable has a different type', () => {
+  const o = of(1, 2, 3).pipe(repeatWhen(repeatWhen(errors => of('a', 'b', 'c')))); // $ExpectType Observable<number>
+});
+
+it('should enforce types', () => {
+  const o = of(1, 2, 3).pipe(repeatWhen()); // $ExpectError
+});
+
+it('should enforce types of the notifier', () => {
+  const o = of(1, 2, 3).pipe(repeatWhen(() => 8)); // $ExpectError
+});
+
+it('should be deprecated', () => {
+  const o = of(1, 2, 3).pipe(repeatWhen(() => of(true))); // $ExpectDeprecation
+});
--- a/spec-dtslint/operators/retryWhen-spec.ts
+++ b/spec-dtslint/operators/retryWhen-spec.ts
@@ -16,3 +16,7 @@ it('should enforce types', () => {
 it('should enforce types of the notifier', () => {
   const o = of(1, 2, 3).pipe(retryWhen(() => 8)); // $ExpectError
 });
+
+it('should be deprecated', () => {
+  const o = of(1, 2, 3).pipe(retryWhen(() => of(true))); // $ExpectDeprecation
+});
--- a/src/internal/operators/repeatWhen.ts
+++ b/src/internal/operators/repeatWhen.ts
@@ -37,6 +37,7 @@ import { createOperatorSubscriber } from './OperatorSubscriber';
  * which a user can `complete` or `error`, aborting the repetition.
  * @return A function that returns an Observable that that mirrors the source
  * Observable with the exception of a `complete`.
+ * @deprecated Will be removed in v9 or v10. Use {@link repeat}'s `delay` option instead.
  */
 export function repeatWhen<T>(notifier: (notifications: Observable<void>) => Observable<any>): MonoTypeOperatorFunction<T> {
   return operate((source, subscriber) => {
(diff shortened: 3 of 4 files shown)


cartant (approved):

BenniG82:
Hi Folks, Hi @benlesh,

I think there may be cases where the current implementation of retryWhen is extremely useful.
See this sample:
https://stackblitz.com/edit/angular-ivy-2pooz5?file=src%2Fapp%2Fapp.component.ts

In this sample retryWhen only retries in cases where the server responds with HTTP-Status 504 (Gateway Timeout).

I racked my brain but I couln't find a solution for this use case without retryWhen.

Sorry if posting a comment to a PR is not desired but this looked like a good place to me :)

benlesh:
Hey, @BenniG82... the new `retry` configuration with `delay` makes this a lot easier. I've forked your example and simplified it the "new way" here:

https://stackblitz.com/edit/angular-ivy-hewest?file=src/app/app.component.ts

Also, a general "pro tip": Within a function like you'd use in `retry`, `retryWhen`, `mergeMap`, et al, you can just `throw error` instead of needing to `return throwError(() => error)`. Same outcome, less code.

BenniG82:
Hi @benlesh,
thank you very much for pointing this out to me. A colleague of mine and I must be blind: We both looked at https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/retry.ts and didn't notice the union type (and the JS Doc) for `delay` in `RetryConfig`.
Now this deprecation makes sense to me :)

In case someone finds this conversation, this may be useful:
**Deprecated** `retryWhen` code:
```
  loadFromUrl(url: string): Observable<any> {
    return this.httpClient.get(url).pipe(
      retryWhen(this.retryNotifier)
    );
  }

  private retryNotifier(errorAttempt: Observable<any>): Observable<any> {
    return errorAttempt.pipe(
      switchMap((error, retryCount) => {
        console.log(error, retryCount);
        if (error.status === 504 && retryCount < 2) {
          return timer(1000).pipe(first());
        }
        throw error;
      })
    );
  }
```

Using `retry`:
```
  loadFromUrl(url: string): Observable<any> {
    return this.httpClient.get(url).pipe(
      retry({
        count: 2,
        delay: this.retryNotifier,
      })
    );
  }
  retryNotifier(error: any, retryCount: number) {
    console.log(error, retryCount);

    if (error.status === 504) {
      return timer(1000);
    }

    throw error;
  }
```

Also thank you for the throwError "pro tip" :+1:

m40l:
Hi @benlesh , 

I'm in a similar situation as BenniG82 where I don't think repeat({delay => observer}) is an adequate replacement for repeatWhen(). I'm trying to use repeat() to repeat a fromFetch request, but I would like to throttle the repeats, like so 

https://stackblitz.com/edit/rxjs-tqvnju?file=index.ts

```
// Original
fromFetch('www.google.com')
  .pipe(repeatWhen(() => fromEvent(document, 'click').pipe(throttleTime(5000))))
  .subscribe(() => console.log('repeatWhen'));

// Suggested non-deprecated alternative
fromFetch('www.google.com')
  .pipe(
    repeat({
      delay: () => fromEvent(document, 'click').pipe(throttleTime(5000)),
    })
  )
  .subscribe(() => console.log('repeat'));
```
The original works one would expect, because repeatWhen only generates the observer once. I think the problem is that repeat resubscribes to the observer generated by delay() after every time fromFetch completes, which is resetting the throttle timer.

I think it would be possible to use repeat({ delay() }) if there was a way for the click Observable to globally throttle all its subscribers, but throwing those search terms into google didn't yield anything. I would really appreciate some insight as to how to migrate my code.

m40l:
Nevermind! Although it 's slightly unintuitive, I found a clean solution!
```
fromFetch('www.google.com')
  .pipe(
    repeat({
      delay: () => fromEvent(document, 'click').pipe(skipUntil(timer(5000)))),
    })
  )
  .subscribe(() => console.log('repeat'));
```

If you know the delay observer will be resubscribed each time, you can use skipUntil to ignore any requests to repeat the fetch for the first 5 seconds *after each request*.

jimbarrett33:
Replacing `repeatWhen()` with a "simple" delay doesn't cover a case where you want to repeat, or more specifically stop repeating, based on something other than time. For example, I have a custom operator that has an input/configuration where the caller can specify when to stop repeating a successful HTTP call based on data being returned from call (or when a retry limit is reached).  

I may be missing something but overall I guess I just to don't get the deprecation of things that work, have been implemented and used by a lot of people for a long time, just to possibly save a few lines of code to make things more simple.

dariomannu:
I made a little example in which a UI (a dialog box) can help decide the course of action, whether to retry, cancel or start over.

[on StackBlitz](https://stackblitz.com/edit/rxjs-retry-rimmel-express?file=pages%2Flib%2Fretry-dialog.ts,pages%2Flib%2Fretryable-fetch.ts,pages%2Findex.ts)


I finally understood myself from an example by @benlesh how `retry` is indeed simpler than `retryWhen`, although it's still absolutely unintuitive from the docs alone.


Maybe all this is confusing because "delay" is English for just "doing something later", not for "implementing a custom logic-driving stream that will decide whether to proceed, retry the previous step or give up".


I think it would be really better to have a basic `delay(n: number)` operator separate from an advanced, switching operator called something else, along the lines of "switchRepeat", "switchRetry", "retryOn", even "retryWhen" is a clearer name IMHO...

PR #6860, opened 2022-03-04
chore: deprecate MapTo variants

benlesh:
Deprecating MapTo variants, as they were only wrappers around the Map variants, and added unnecessary API surface area.

related #6367

### DIFF
--- a/src/internal/operators/switchMapTo.ts
+++ b/src/internal/operators/switchMapTo.ts
@@ -2,7 +2,7 @@ import { switchMap } from './switchMap';
 import { ObservableInput, OperatorFunction, ObservedValueOf } from '../types';
 import { isFunction } from '../util/isFunction';
 
-/* tslint:disable:max-line-length */
+/** @deprecated Will be removed in v9. Use {@link mergeMap} instead: `mergeMap(() => result)` */
 export function switchMapTo<O extends ObservableInput<unknown>>(observable: O): OperatorFunction<unknown, ObservedValueOf<O>>;
 /** @deprecated The `resultSelector` parameter will be removed in v8. Use an inner `map` instead. Details: https://rxjs.dev/deprecations/resultSelector */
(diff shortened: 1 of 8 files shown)


cartant (commented):
Some of the dtslint tests are failing:

```
ERROR: /home/runner/work/rxjs/rxjs/spec-dtslint/operators/mapTo-spec.ts:17:47 - No deprecation found
ERROR: /home/runner/work/rxjs/rxjs/spec-dtslint/operators/mergeMapTo-spec.ts:80:56 - No deprecation found
```

It looks like the `mapTo` signature that's being matched doesn't have a deprecation tag and `mergeMapTo` does not appear to have any deprecation tags.

cartant (approved):

src/internal/operators/switchMapTo.ts
```diff
@@ -2,7 +2,7 @@ import { switchMap } from './switchMap';
 import { ObservableInput, OperatorFunction, ObservedValueOf } from '../types';
 import { isFunction } from '../util/isFunction';
 
-/* tslint:disable:max-line-length */
+/** @deprecated Will be removed in v9. Use {@link mergeMap} instead: `mergeMap(() => result)` */
```
balazserdeszlogmein, line 5: Why is `mergeMap` the recommended substitution for `switchMapTo`?

src/internal/operators/switchMapTo.ts
```diff
@@ -55,6 +54,7 @@ export function switchMapTo<T, R, O extends ObservableInput<unknown>>(
  * `resultSelector`) every time a value is emitted on the source Observable,
  * and taking only the values from the most recently projected inner
  * Observable.
+ * @deprecated Will be removed in v9. Use {@link mergeMap} instead: `mergeMap(() => result)`
```
timothyalcaide, line 57: Same as above...

PR #6873, opened 2022-03-09
fix: `share` and `connect` no longer bundle scheduling code by default

### LINKED ISSUE #6872: Don't use `from(x)` in the library, use `innerFrom`
Cases where we are using `from(x)` instead of `innerFrom(x)` cause the inclusion of `scheduled`, which then includes a host of other things, `observeOn`, `subscribeOn`, etc.

Related to conversation here https://github.com/trpc/trpc/pull/1592#issuecomment-1062462319

benlesh:
resolves #6872

### DIFF
--- a/src/internal/observable/innerFrom.ts
+++ b/src/internal/observable/innerFrom.ts
@@ -1,7 +1,7 @@
 import { isArrayLike } from '../util/isArrayLike';
 import { isPromise } from '../util/isPromise';
 import { Observable } from '../Observable';
-import { ObservableInput, ReadableStreamLike } from '../types';
+import { ObservableInput, ObservedValueOf, ReadableStreamLike } from '../types';
 import { isInteropObservable } from '../util/isInteropObservable';
 import { isAsyncIterable } from '../util/isAsyncIterable';
 import { createInvalidObservableTypeError } from '../util/throwUnobservableError';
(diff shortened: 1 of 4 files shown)


benlesh:
Before: 

image

After:

image

src/internal/observable/innerFrom.ts
```diff
@@ -12,6 +12,7 @@ import { isFunction } from '../util/isFunction';
 import { reportUnhandledError } from '../util/reportUnhandledError';
 import { observable as Symbol_observable } from '../symbol/observable';
 
+export function innerFrom<O extends ObservableInput<any>>(input: O): Observable<ObservedValueOf<O>>;
```
cartant, line 15: Can we just change the signature instead of adding an overload signature - which will hide the implementation signature anyway?

cartant (approved):

src/internal/observable/innerFrom.ts
benlesh: I did that and the guts of the implementation hated me for it. I'll fiddle with it again before I merge it, but I figure it's not super important since it's not a publicly exposed function.

benlesh: Upon review. I'm going to do a follow up PR on this, because I think maybe there are some issues with `onErrorResumeNext`'s types.

PR #6889, opened 2022-03-14
fix: animationFrameScheduler and asapScheduler no longer executing actions

### LINKED ISSUE #6854: animationFrameScheduler incorrect behavior
### Describe the bug

AnimationFrameScheduler doesn't execute scheduled tasks under certain circumstances.

If AnimationFrameScheduler gets a new id (e.g. new task is scheduled) before the flush method empties current **actions** array in the do-while loop we encounter wrong behavior.
In this case we enter the recycleAsyncId method of AnimationFrameAction . 
Then recycleAsyncId sets _scheduled field to undefined. 
After that we add new actions to the **actions** array in requestAsyncId, start requestAnimationFrame and enter flush method again on browser repaint. 
But this time **actions** array in the flush method has two types of _action_ with different id's. 
What is the most important is that the first element of the actions array has id which is different from the current frame id. 
This leads us to false condition "action.id === flushId" (in do-while loop) because action.id is the id of the previous frame but flushId is the id of the current frame. 
Since that moment we are unable to execute any action in the **actions** array and empty it.

Problem occured after commit e35f589e2ca10ab2d2d69f7e9fe60727edc4c53d
in src/internal/scheduler/AnimationFrameAction.ts lines 30-33

### Expected behavior

**Expected console log output:**
default task
task1
task2
default task
task1
task2

### Reproduction code

```typescript
let reschedule = true;
function log() {
  console.log('default task');
  animationFrameScheduler.schedule(() => console.log('task1'));
  animationFrameScheduler.schedule(() => console.log('task2'));
  if (reschedule) {
    this.schedule();
    reschedule = false;
  }
}

animationFrameScheduler.schedule(log);
```

### Reproduction URL

https://stackblitz.com/edit/1gjlhu?devtoolsheight=50&file=index.ts

### Version

7.5.1

### Environment

_No response_

### Additional context

animationFrameScheduler from rxjs version 7.1.0 worked correctly in this case

ajafff:
**Description:**

`AnimationFrameAction` and `AsapAction`'s `recycleAsyncId` should only reset `scheduler_scheduled` if it's the same `id` we are currently recycling. This fixes the issue described in #6854 where actions scheduled from another action have already set `scheduler._scheduled`, which is then cleared when the currently executing action is done.
This works because there are at most 2 schedules for these schedulers: the currently executing one and the next schedule. This also means we no longer `cancelAnimationFrame`/`clearImmediate` the current schedule after its last action is done (which does nothing for correctness, but avoid unnecessary work).

Fixes #6854

### DIFF
--- a/spec/schedulers/AsapScheduler-spec.ts
+++ b/spec/schedulers/AsapScheduler-spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
 import * as sinon from 'sinon';
-import { asapScheduler, Subscription, SchedulerAction, merge } from 'rxjs';
+import {asapScheduler, Subscription, SchedulerAction, merge, animationFrameScheduler} from 'rxjs';
 import { delay } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
@@ -287,5 +287,28 @@ describe('Scheduler.asap', () => {
       sandbox.restore();
       done();
     });
+  });
+
+  it('should handle actions scheduled during flush before current action is rescheduled', (done) => {
+    const sandbox = sinon.createSandbox();
+
+    const result: string[] = [];
+    let reschedule = true;
+    function work(this: SchedulerAction<unknown>) {
+      result.push('work');
+      if (reschedule) {
+        asapScheduler.schedule(() => result.push('task 1'));
+        asapScheduler.schedule(() => result.push('task 2'));
+        this.schedule();
+        expect(result).to.deep.equal(['work']);
+        reschedule = false;
+      } else {
+        expect(result).to.deep.equal(['work', 'task 1', 'task 2', 'work']);
+        sandbox.restore();
+        done();
+      }
+    }
+    asapScheduler.schedule(work);
+    expect(result).to.deep.equal([]);
   });
 });
--- a/spec/schedulers/AnimationFrameScheduler-spec.ts
+++ b/spec/schedulers/AnimationFrameScheduler-spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
 import * as sinon from 'sinon';
-import { animationFrameScheduler, Subscription, merge } from 'rxjs';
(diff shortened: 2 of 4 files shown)


spec/schedulers/AsapScheduler-spec.ts
```diff
@@ -288,4 +288,25 @@ describe('Scheduler.asap', () => {
...
+  it('should handle actions scheduled during flush before current action is rescheduled', (done) => {
+    const sandbox = sinon.createSandbox();
+
+    const result: string[] = [];
+    let reschedule = true;
+    function work(this: SchedulerAction<unknown>) {
+      result.push('work');
+      if (reschedule) {
+        asapScheduler.schedule(() => result.push('task 1'));
+        asapScheduler.schedule(() => result.push('task 2'));
```
benlesh, line 301: Can we add an assertion right here to make sure that the above tasks didn't execute synchronously in this spot? It's not so much that I think they did, it's that I want to make sure they're "scheduled" and don't just happen to mimic the proper outcome by virtue of when we checked.

Please do the same for the other test? Thanks, @ajafff!

Otherwise, this looks great.

benlesh (changes requested):

spec/schedulers/AsapScheduler-spec.ts
ajafff: done

andRave: @benlesh could you please review it again? Can't upgrade above 7.4.0 due to this problem.

pmoleri: OMG. This is a serious bug and produces all kind of issues on applications that use animationFrameScheduler.
I didn't notice about this PR (and created a similar one myself), almost 2 years later! [!7444](https://github.com/ReactiveX/rxjs/pull/7444)

Can we please review one of them and merge it?

pmoleri:
Hi @ajafff. My PR #7444 which addressed the same issue as this one has been merged and released on [`7.8.2`](https://github.com/ReactiveX/rxjs/commit/e5351d02e225e275ac0e497c7b66eaa5f0c88791).

PR #6890, opened 2022-03-14
docs: remove redundant deprecated section

jakovljevic-mladen:
**Description:**
This PR removes redundant section related to the deprecated stuff. This was added in #6379, but since this section existed before that PR, I will remove it as duplicated content.

image

With this part removed, it looks like this:

image

None


ajafff:
On a related note: it seems only the implementation signature of `pluck` is marked as deprecated. This causes TypeScript and other tooling (like eslint-plugin-deprecation) to not display uses of the operator as deprecated.

jakovljevic-mladen:
@ajafff, PRs that fix such issues are welcome.

benlesh (approved):

PR #6892, opened 2022-03-14
fix: animationFrameScheduler prematurely executing actions without delay

### LINKED ISSUE #6891: animationFrameScheduler with delay causes incorrect behavior (and exception)
### Describe the bug

If `AnimationFrameScheduler#schedule` is used with a `delay > 0`, the action is scheduled as `AsyncAction`. Executing the action takes the same code path as `AnimationFrameAction`.
So if the `AsyncAction` executes while `AnimationFrameAction`s are scheduled, it executes those too, which is too early as the browser is not yet preparing the next frame.

With the fix in #6889 it will throw an error, when `rAF` tries to execute the actual `AnimationFrameAction`s, but the queue is already empty: 
```
Uncaught TypeError: Cannot read properties of undefined (reading 'execute')
      at AnimationFrameScheduler.flush (src\internal\scheduler\AnimationFrameScheduler.ts:24:27)
      at C:\code\rxjs\src\internal\scheduler\AnimationFrameAction.ts:21:121
```

### Expected behavior

`AsyncAction` to be handled separately. Ideally we could use `asyncScheduler.schedule(...)` instead of `super.request/recycleAsyncId`.

### Reproduction code

```typescript
animationFrameScheduler.schedule(() => console.log('async'), 1);
animationFrameScheduler.schedule(() => console.log('animation frame'));
```

### Reproduction URL

_No response_

### Version

7.5.5

### Environment

_No response_

### Additional context

I could try to fix this, but I'm not sure how to test this properly. The code above flaky, because the timeout of 1ms is not guaranteed to be executed before the next animation frame. I guess I could mock the relevant `animationFrameProvider`

`asapScheduler` uses similar code, but seems not to be affected, because timeouts will never be executed before microtasks.

ajafff:
**Description:**

If `AnimationFrameScheduler#flush` is called with an `action` as parameter, it's an action with `delay > 0`. Only execute this action and don't touch `scheduler._scheduled` and `scheduler.actions` because the action it not affected by any of them.

This branch is based on #6889 to avoid merge conflicts in the spec file.

Fixes: #6891

### DIFF
--- a/spec/schedulers/AnimationFrameScheduler-spec.ts
+++ b/spec/schedulers/AnimationFrameScheduler-spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
 import * as sinon from 'sinon';
-import { animationFrameScheduler, Subscription, merge } from 'rxjs';
+import {animationFrameScheduler, Subscription, merge, SchedulerAction} from 'rxjs';
 import { delay } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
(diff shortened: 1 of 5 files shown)


benlesh:
Hi @ajafff — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

Issue #3561, opened 2018-04-13
already-handled errors are also "unhandled"


mattflix:
**RxJS version:**
5.5.8

**Code to reproduce:**
```javascript
window.onerror = function() {
  console.log(`onerror event fired anyway`);
};

Rx.Observable
  .of(`something`)
  .observeOn(Rx.Scheduler.async)
  .flatMap(() =>
    Rx.Observable
      .create(
        observer => {
          throw `trantrum`
        }
      )
  )
  .catch(
    error => {
      console.log(`totally handled the error`);
      return Rx.Observable.empty();
    }
  )
  .subscribe();
```
**Expected behavior:**
```
totally handled the error
```

**Actual behavior:**
```
totally handled the error
onerror event fired anyway
```

**Additional information:**
This behavior seems to have been introduced in 5.5.0.  Versions < 5.5.0, such as 5.4.3, exhibit the expected behavior.

If the `Observable.create()` call is replaced with `Observable.defer()` (that still synchronously throws an error), we get the expected behavior, even with versions > 5.4.3, including 5.5.8.

If this behavior is expected, it seems completely surprising, and very painful to deal with.  Why create observables with supposedly self-contained error-handling logic, and yet have to ALSO deal with the onerror events?

mattflix:
See also #3560.

benlesh:
This is because you're expected to handle your own errors inside of the init function passed to the `Observable` constructor (aka `Observable.create`), then hand those errors to the `observer.error` yourself.

```ts
new Observable(() => { throw new Error('test'); });

// vs

new Observable(o => o.error(new Error('test')));
```

mattflix:
@benlesh, your comment doesn't seem to explain why the error is delivered BOTH to the `.catch()` to also to the unhandled error callback.  (See my repro case.)

Are you saying that behavior (one error going to two handlers in contradiction with each other; handled vs. unhandled) is by-design?  If so, can you provide the rationale behind that?

Issue #3560, opened 2018-04-13
subscribe() insists on throwing already-handled errors


mattflix:
**RxJS version:**
5.5.8

**Code to reproduce:**
```javascript
try {
  Rx.Observable
    .create(
      observer => {
        throw `trantrum`
      }
    )
    .catch(
      error => {
        console.error(`totally handled the error: ${ error }`);
        return Rx.Observable.empty();
      }
    )
    .subscribe();
}
catch (error) {
  console.error(`subscribe() blew up anyway: ${ error }`);
}

```
**Expected behavior:**
```
totally handled the error: trantrum
```

**Actual behavior:**
```
totally handled the error: trantrum
subscribe() blew up anyway: trantrum
```

**Additional information:**
This behavior seems to have been introduced in 5.5.6.  Versions < 5.5.6 exhibit the expected behavior.

If the `Observable.create()` call is replaced with `Observable.defer()` (that still synchronously throws an error), we get the expected behavior, even with versions >= 5.5.6.

If this new 5.5.6 behavior is expected, it seems completely surprising, and very painful to deal with.  Why create observables with supposedly self-contained error-handling logic, and yet have to litter my code with try/catch blocks?

jayphelps:
Whoops! 🤡 

What ever happens be _very_ careful as https://github.com/ReactiveX/rxjs/commit/541b49db0ca4079760f92f670a49a8f06dc83dd2 fixed a ton of people who's errors were otherwise being swallowed silently. Hopefully test coverage will protect us, but just in case!

mattflix:
See also #3561.

jayphelps:
@mattflix is it not a duplicate?

mattflix:
@jayphelps, perhaps so, but #3561 requires an asynchronous element.  This one does not.

cartant:
The two issues refer to the same problem; the only difference is the stack frame from which the error is re-thrown.

The problem is that the error [caught in `_trySubscribe`](https://github.com/ReactiveX/rxjs/blob/5.5.8/src/Observable.ts#L216-L222) is [re-thrown](https://github.com/ReactiveX/rxjs/blob/5.5.8/src/Observable.ts#L207-L209) - because `syncErrorThrown ` is set to `true` - even though it's reported via `sink.error`.

The repro can be made simpler, as the `catch` operator can be removed. This snippet will still reproduce the problem:

```js
try {
  Rx.Observable
    .create(observer => { throw `trantrum`; })
[...]

PR #6893, opened 2022-03-14
7.5.5

Aljumaily:
**Description:**
The `error` property in an observable doesn't stop the program from crashing:

```trypscript
new Observable((channel) => {
    channel.error(`observable x1$ encountered an error`)
}).subscribe({
    next: (x: number) => {
        console.log(x)
    },
    error: (error: string) => {//still crashes the program
        console.log(`the following error is caught: ${error}`);
    },
    complete: () => {
        console.log()
    }
});
```

The fix is to use `try`/`catch`:
```trypscript
try{
    new Observable((channel) => {
        channel.error(`observable x1$ encountered an error`)
    }).subscribe({
        next: (x: number) => {
            console.log(x)
        },
        error: (error: string) => {//still gives an error
            console.log(`the following error is caught: ${error}`);
        },
        complete: () => {
            console.log()
        }
    });
}catch(e){
    console.log(`Error is caught and will not crash the program: ${e}`);
}
```
What is the appropriate approach to handle this situation?

[3560](https://github.com/ReactiveX/rxjs/issues/3560), [3561](https://github.com/ReactiveX/rxjs/issues/3561). It has been four years since v5.5.8 and it still persists in v7.5.5.


kwonoj:
Please use our gh issue template with minimally reproducible examples. We cannot audit all of commits included in this PR wothout having details.

PR #6894, opened 2022-03-14
Marble Diagrams: Dead Link Correction

webia1:
A dead link in the documentation is corrected.


jakovljevic-mladen:
I think that there is no need to fix anything here. The image loading is working with the existing `src` path as well - it's just the fact that the latest docs app is not yet publicly released. I'm really sorry about that, I will try to get it deployed as soon as possible.

BTW, this has been fixed in #6558 - it's been reverted to the value it originally had when it was first introduced in #4151.

PR #6895, opened 2022-03-15
Typescript 4.6 compatibility

sundbry:
Hello Rxjs team! A few patches here to support the latest Typescript 4.6.

**Description:** Fixes compilation under typescript 4.6.

There are three commits here:

1) fixes misc. failures in the type checker by adding some type casts where union types were assigned. 

2) In this patch, the return value on the type annotation for setInterval/setTimeout was set to `number`. This is [incorrect under node.js](https://bobbyhadz.com/blog/typescript-settimeout-type), where they return a Timer object. The type annotation is fixed to return a union of `number | NodeJS.Timer` instead.

3) adds support for compiling with `isolatedModules: true` (such as is required in next.js, babel, swc) by using the `export type` feature (supported since typescript 3.8.)


benlesh (approved):

src/internal/scheduler/timerHandle.ts
```diff
@@ -0,0 +1 @@
+export type TimerHandle = number | NodeJS.Timeout;
```
benlesh, line 1: fancy. I think I usually just do `ReturnType<typeof setTimeout>`. LOL... because I don't ever trust that I'm right.

benlesh:
This is breaking documentation generation. Looking back at this, I'm not entirely sure why this change was necessary. 🤔

sundbry:
> This is breaking documentation generation. Looking back at this, I'm not entirely sure why this change was necessary. 🤔

The type signature was invalid under node.js. It was necessary to correct the error, as the newer typescript releases detect it.

benlesh:
the `export type X from Y` bit is what is breaking the docs. Were those fixes necessary for correction of your type errors? Or was it just about the return type of `setTimeout`?

src/internal/scheduler/timerHandle.ts
hakimio, line 1: @sundbry This breaks compatibility with yarn v3 pnp: https://github.com/yarnpkg/berry/actions/runs/3521483267/jobs/5903377561

Issue #6910, opened 2022-03-25
exhaustAll never completes when lastValueFrom is used


mlegenhausen:
### Describe the bug

`exhaustAll` does not behave the same as `exhaustMap(identity)`

### Expected behavior

I would expect that `exhaustAll` should behave the same as `exhaustMap(identity)`

### Reproduction code

```typescript
import { exhaustAll, exhaustMap, firstValueFrom, lastValueFrom, of, identity } from 'rxjs'

const a = of(of(1))

const a1 = await firstValueFrom(a.pipe(exhaustMap(identity)))
assert.deepStrictEqual(a1, 1) // success

const a2 = await firstValueFrom(a.pipe(exhaustAll()))
assert.deepStrictEqual(a2, 1) // success

const a3 = await lastValueFrom(a.pipe(exhaustMap(identity)))
assert.deepStrictEqual(a3, 1) // success

const a4 = await lastValueFrom(a.pipe(exhaustAll()))
assert.deepStrictEqual(a4, 1) // timeout cause `a4` never resolves
```

### Version

7.5.5

PR #6911, opened 2022-03-25
fix: exhaustAll use exhaustMap to handle synchronous completion of in…

ajafff:
…ner Observable

**Description:**

Removes the implementation of `exhaustAll` and reuses `exhaustMap(identity)` instead.

Fixes: #6910


benlesh:
This seems like a fine fix, considering that `exhaustMap` is basically doing the same thing as `exhaustAll` only with the projection.

benlesh (approved):

PR #6919, opened 2022-04-02
chore(interval): convert interval specs to run mode

tmair:
**Description:**
This PR converts `interval` unit tests to run mode.

None

### DIFF
--- a/spec/observables/interval-spec.ts
+++ b/spec/observables/interval-spec.ts
@@ -14,35 +14,6 @@ describe('interval', () => {
     rxTestScheduler = new TestScheduler(observableMatcher);
   });
 
-  it('should create an observable emitting periodically', () => {
-    rxTestScheduler.run(({ expectObservable, time }) => {
-      const values = {
-        a: 0,
-        b: 1,
-        c: 2,
-        d: 3,
-        e: 4,
-        f: 5,
-      };
-
-      const period = time('--|           ');
-      //                     --|
-      //                       --|
-      //                         --|
-      //                           --|
-      //                             --|
-      //                               --|
-      const expected = '   --a-b-c-d-e-f-';
-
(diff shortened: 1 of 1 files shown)


spec/observables/interval-spec.ts
```diff
@@ -1,64 +1,101 @@
...
+      const expected = '   --a-b-c-d-e-f-';
+
+      const e1 = interval(period).pipe(
+        take(6), // make it actually finite, so it can be rendered
+        concat(NEVER) // but pretend it's infinite by not completing
+      );
+
+      expectObservable(e1).toBe(expected, values);
+    });
   });
```
jakovljevic-mladen, line 44: I'm not sure that we need this test at all. It seems that this test used to be used as a placeholder for the generation of the marble diagram images.

Since we don't create marble diagram images out of the tests anymore, comments `// make it actually finite, so it can be rendered` and `// but pretend it's infinite by not completing` are obsolete.

If the entire test does not make any value (or, if there is a test that does something similar), I would suggest removing it.

However, if the test makes any value, at least removal of these comments would be a good idea. And probably `concat(NEVER)` is not needed as well...

tmair: I don't exactly know if it can be removed. I also guess, that it was used to render diagram images.

It seems that the test next test in the file asserts pretty much the same as this one.

jakovljevic-mladen: Yeah, many files contain very similar first and second test, just because the first one was used for marble rendering purposes. I suggest removing this test, or at least removing comments related to the marble diagrams. However, I'd like to hear @cartant and/or @benlesh's opinion.

cartant: Yeah, I agree. Let's remove the test.

cartant (approved):

PR #6934, opened 2022-04-16
docs: fix mergeMap/switchMap copy/paste error

cartant:
**Description:**

Fixes a copy/paste error in the `switchMapTo` TSDoc - the replacement should be `switchMap` and not `mergeMap` - that was mentioned [here](https://github.com/ReactiveX/rxjs/pull/6860#pullrequestreview-914312284) and [here](https://github.com/ReactiveX/rxjs/pull/6860#pullrequestreview-919493278).


jakovljevic-mladen (approved):
LGTM

PR #6936, opened 2022-04-19
docs: Inline templates, correct type errors that revealed.

benlesh:
This change simply uses inline template instead of `templateUrl`. I've always hated `templateUrl` haha. 

Anyhow, doing this reveals more type errors and gives better feedback in VS Code (if you're using Angular Language Service).


docs_app/src/app/shared/search-results/search-results.component.ts
```diff
@@ -45,14 +71,16 @@ export class SearchResultsComponent implements OnChanges {
     }
```
jakovljevic-mladen, line 71: Since you added check in `ngOnChanges`, this part: `if (!search) { return []; }` can be removed.

jakovljevic-mladen (approved):
LGTM. One check can be removed since it was moved to another place.

PR #6940, opened 2022-04-20
race synchronous emissions now stop subsequent observable subscriptions

snesin:
Looking at the code, if an observable synchronously emits, subsequent observables will not be subscribed to. Synchronous errors and completions were noted as a breaking change, so adding emits to the list.

**Description:**

**BREAKING CHANGE:** 

### DIFF
--- a/docs_app/content/6-to-7-change-summary.md
+++ b/docs_app/content/6-to-7-change-summary.md
@@ -107,7 +107,7 @@ This document contains a detailed list of changes between RxJS 6.x and RxJS 7.x,
 #### race
 
 - Generic signatures have changed. Do not explicitly pass generics.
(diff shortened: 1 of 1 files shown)


cartant (commented):
It looks to me like this change is incorrect, as the version 6 implementation - AFAICT - would stop subscribing to subsequent sources when it received a synchronous next emission from a source. See the use of `this.hasFirst` below:

https://github.com/ReactiveX/rxjs/blob/881cacdc99a7ebae219b595004c632f7358f730d/src/internal/observable/race.ts#L107-L115

So that behaviour wasn't changed in v7. That is, stopping subscribing on a sync next is not a breaking change.

Issue #6716, opened 2021-12-11
zip() throws (swallowed) error when closed synchronously in next() handler


chrisguttandin:
### Describe the bug

After enabling logging for notifications on stopped observables with `config.onStoppedNotification` I noticed that `zip()` sometimes throws an error which can't be delivered anymore in case the observable gets closed from within the next handler.

Firefox: `TypeError: buffers is null`
Chrome: `TypeError: Cannot read properties of null (reading 'some')`

The error originates from here: https://github.com/ReactiveX/rxjs/blob/081ca2ba7290aa3084c1477a6d4bcc573bf478f6/src/internal/observable/zip.ts#L91

### Expected behavior

I would not expect this to happen. I think this case should be handled internally.

### Reproduction code

```typescript
import { NEVER, concat, config, first, of, zip } from 'rxjs';

config.onStoppedNotification = (notification) => {
  if (notification.kind === 'E') {
    throw notification.error;
  }
};

const source$ = concat(of(1), NEVER);

zip(source$).pipe(first()).subscribe();
```

### Reproduction URL

https://stackblitz.com/edit/rxjs-qxrbts?devtoolsheight=60&file=index.ts

### Version

7.4.0

### Environment

I discovered this in an Angular App (v13.1) which also uses NgRx (v13) but I guess it's not relevant for this issue.

### Additional context

I think adding a simple check if `buffers` is already null before calling `buffers.some()` should fix the issue. I would be happy to provide a PR with this fix.

truebden:
I just faced this issue in my code too and it would be great to have the PR merged 😉

benlesh:
Hi @chrisguttandin — thank you for taking the time to open this issue and for sharing your thoughtful feedback with the RxJS project. I’m sorry it took us so long to respond. Given how much time has passed and the lack of recent activity, I’m going to close this for now so the tracker reflects the work we’re likely to act on. If this is still relevant on a current RxJS release, please feel welcome to open a new issue with updated context and a reproduction. Thank you again for your contribution and patience.

— Ben

PR #6955, opened 2022-05-04
fix(zip): handle when unsubscribe is called from within next

chrisguttandin:
**Description:**

This PR makes sure the error reported here (#6716) doesn't get thrown anymore.

**BREAKING CHANGE:**  There are no breaking changes.

**Related issue (if exists):** #6716

I've added a test case which covers the bug but I wasn't sure if there might be a better way to test this. Please let me know if I need to change anything.

### DIFF
--- a/src/internal/observable/zip.ts
+++ b/src/internal/observable/zip.ts
@@ -90,7 +90,7 @@ export function zip(...args: unknown[]): Observable<unknown> {
                   // If any one of the sources is both complete and has an empty buffer
                   // then we complete the result. This is because we cannot possibly have
                   // any more values to zip together.
-                  if (buffers.some((buffer, i) => !buffer.length && completed[i])) {
+                  if (buffers?.some((buffer, i) => !buffer.length && completed[i])) {
                     destination.complete();
                   }
                 }
--- a/spec/observables/zip-spec.ts
+++ b/spec/observables/zip-spec.ts
@@ -1,6 +1,6 @@
 /** @prettier */
 import { expect } from 'chai';
-import { queueScheduler as rxQueueScheduler, zip, from, scheduled } from 'rxjs';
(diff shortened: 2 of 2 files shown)


src/internal/observable/zip.ts
```diff
@@ -88,7 +88,7 @@ export function zip(...args: unknown[]): Observable<unknown> {
                   // If any one of the sources is both complete and has an empty buffer
                   // then we complete the result. This is because we cannot possibly have
                   // any more values to zip together.
-                  if (buffers.some((buffer, i) => !buffer.length && completed[i])) {
+                  if (buffers !== null && buffers.some((buffer, i) => !buffer.length && completed[i])) {
```
josepot, line 91: nit:

```suggestion
                  if (buffers?.some((buffer, i) => !buffer.length && completed[i])) {
```

chrisguttandin: I initially thought using an explicit `!== null` check is better as it results in a boolean which some people (including myself :-)) prefer. But I think your suggestion is more in line with the existing RxJS codebase.

I've changed it. Thanks for the review.

chrisguttandin:
Sorry for tagging you @benlesh. I know it sucks. But given that this PR already had it's first anniversary and I rebased it a couple of times I hope it's okay. I would really appreciate if you could take a look. The PR only adds a single character and a test.

Many thanks in advance.

truebden:
I just faced this issue in my code too and it would be great to have the PR merged 😉

benlesh:
Hi @chrisguttandin — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

PR #6991, opened 2022-06-11
Evade override mistake

erights:
See https://github.com/Agoric/agoric-sdk/blob/master/patches/inquirer%2B%2Brxjs%2B7.5.5.patch

The original code used assignment to override the `constructor` property inherited from `Error.prototype`. However, if `Error.prototype` is frozen, as it is under Hardened JS (aka SES) or under the Node frozen intrinsics flag, then this assignment fails due to the JavaScript "override mistake".

`enumerable: true` would accurately preserve the behavior of the original assignment, but I'm guessing that was not intentional. For an actual error subclass, this property would not be enumerable, so my PR currently proposes that. But either would work, so let me know if you'd like me to change it.

**Description:**

**BREAKING CHANGE:** 

https://github.com/endojs/endo/issues/576
https://github.com/Agoric/agoric-sdk/pull/5569

### DIFF
--- a/src/internal/util/createErrorClass.ts
+++ b/src/internal/util/createErrorClass.ts
@@ -14,7 +14,13 @@ export function createErrorClass<T>(createImpl: (_super: any) => any): T {
   };
 
   const ctorFunc = createImpl(_super);
-  ctorFunc.prototype = Object.create(Error.prototype);
-  ctorFunc.prototype.constructor = ctorFunc;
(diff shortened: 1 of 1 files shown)


erights:
Any progress on this?

benlesh:
I’m inclined to approve this, on the face of it. However is it possible to add a test for this?

benlesh:
Inactive PR... no response and no tests.

PR #6996, opened 2022-06-21
Feature/add some operator

Yozmo:
**Description:**

Creating `some` boolean operator which emits a value whenever or not some value emitted by the source Observable satisfies the predicate function provided.


Yozmo:
> Isn't this the same as `first()` with a predicate function?
> 
> ```
> import { range, first } from 'rxjs';
> 
> range(1, 10)
>   .pipe(
>     first(v => v === 5),
>   )
>   .subscribe(console.log);
> ```
> 
> If there's no matching value I could add `catchError(() => EMPTY)`.
> 
> https://stackblitz.com/edit/rxjs-h3dt7s

Like for `Array.prototype.find()` and `Array.prototype.some()`, one is returning the element itself, and the other one is returning a boolean in case there is an element which satisfies the predicate.
There are cases where you just want to know if the source Observable emits a certain value, without keeping the element.

kwonoj:
We strongly recommend to create a new userland package for the new operators / features before including it as core api surface, as core team tries to remove existing interface / suppress new one to not to increase api surfaces anymore.

I may miss something, but so far it looks like this should be possible to create as userland operator without internals of core behavior, which I'd recommend to do so.

Yozmo:
> We strongly recommend to create a new userland package for the new operators / features before including it as core api surface, as core team tries to remove existing interface / suppress new one to not to increase api surfaces anymore.
> 
> I may miss something, but so far it looks like this should be possible to create as userland operator without internals of core behavior, which I'd recommend to do so.

I understood the part where the goal is not to increase the api surface anymore, but I didn't fully understand the part with ' create a new userland package for the new operators / features before including it as core api surface'. Could you explain this one, please.

kwonoj:
Means publish as separate npm package.

demensky:
This operator can be simply reproduced with `findIndex` and `map`:

```ts
export function some<T>(predicate: (value: T, index: number) => boolean): OperatorFunction<T, boolean> {
  return pipe(
    findIndex(predicate),
    map((index) => index !== -1)
  );
}
```

```ts
of(2, 4, 6, 8)
  .pipe(some((x) => x % 2 !== 0))
  .subscribe(console.log); // false

of(2, 4, 5, 6)
  .pipe(some((x) => x % 2 !== 0))
  .subscribe(console.log); // true

of()
  .pipe(some((x) => x % 2 !== 0))
  .subscribe(console.log); // false
```

Yozmo:
> This operator can be simply reproduced with `findIndex` and `map`:
> 
> ```ts
> export function some<T>(predicate: (value: T, index: number) => boolean): OperatorFunction<T, boolean> {
>   return pipe(
>     findIndex(predicate),
>     map((index) => index !== -1)
>   );
> }
> ```
> 
> ```ts
> of(2, 4, 6, 8)
>   .pipe(some((x) => x % 2 !== 0))
>   .subscribe(console.log); // false
> 
> of(2, 4, 5, 6)
>   .pipe(some((x) => x % 2 !== 0))
>   .subscribe(console.log); // true
> 
> of()
>   .pipe(some((x) => x % 2 !== 0))
>   .subscribe(console.log); // false
> ```

Totally agree with you. But if we are looking this way, many other operators can be reproduced combining other operators. Some of them literally just call another operator (`interval` calling `timer`, `min` and `max` both are using `reduce` and so on).
My implementation for `some` seemed simpler than other combinations, and I thought it would be useful to be added, as an oposite to `every` which already exists. But that was just an idea.

demensky:
@Yozmo There are already many operators in this codebase. Each new operator will need to be: tested (mocha & dtslint), documented and supported. The more "operators" the more difficult it is for new users of this library to find the right one.

For example, operators such as [`multicast`](https://rxjs.dev/api/operators/multicast), [`publish`](https://rxjs.dev/api/operators/publish), [`publishBehavior`](https://rxjs.dev/api/operators/publishBehavior), [`publishLast`](https://rxjs.dev/api/operators/publishLast), [`publishReplay`](https://rxjs.dev/api/operators/publishReplay) were very powerful, but it takes more effort to figure out which is better to use. Later these were replaced by the updated [`share`](https://rxjs.dev/api/operators/share) operator, it became much simpler and at the same time there were features that were previously very difficult to implement without custom operators (such as delayed unsubscribe).

About the fact that operators reuse each other:
For example, the [`mapTo`](https://rxjs.dev/api/operators/mapTo) operator is sugar for the [`map`](https://rxjs.dev/api/operators/map) operator. [It is marked as **deprecated**](https://github.com/ReactiveX/rxjs/blob/7.5.5/src/internal/operators/mapTo.ts#L44).
I think that I can make an issue for such operators as [`min`](https://rxjs.dev/api/operators/min) and [`max`](https://rxjs.dev/api/operators/max).

```ts
foo.pipe(max((a, b) => a.age > b.age ? 1 : -1));
foo.pipe(reduce((a, b) => a.age > b.age ? a : b)));
```

```ts
foo.pipe(max());
foo.pipe(reduce(Math.max));
```

kwonoj:
> many other operators can be reproduced combining other operators.

This is exact reason core team slowly discontinue those kind of operators over time, which takes long time due to it is a breaking change.

For the same reason, we recommend to publish new operator as userland to avoid same practices we've done, such as expanding api first then take long time to find out we need to reduce those surfaces.

Yozmo:
Thanks for clarifying

Issue #6760, opened 2022-01-12
shareReplay with refCount stops working


ghostlytalamaur:
### Describe the bug

When subscribe to shared observable and use `take(1)` operator, internal connection to the source observable gets broken.

### Expected output:
```
first: 123
second: 123
```

### Actual output
```
first: 123
```

### Reproduction code

```typescript
import { Subject } from 'rxjs';
import { shareReplay, take } from 'rxjs/operators';

const subj$$ = new Subject<number>();
const shared$ = subj$$.pipe(shareReplay({ refCount: true, bufferSize: 1 }));

shared$.pipe(take(1)).subscribe((value) => console.log('first:', value));
subj$$.next(123);

shared$.subscribe((value) => console.log('second:', value));
```

### Reproduction URL

https://stackblitz.com/edit/q7u9jg

### Version

7.5.1

EarthyOrange:
Removing `.pipe(take(1))` makes the 7.5.1 snippet work. So it looks like `take` is somehow affecting the `shareReplay` operator.

lincolnthree:
Confirmed I am also seeing behavior where `shareReplay({ bufferSize: 1, refCount: true })` stops emitting. My theory is also that it is causing the source observable to somehow complete when it should not, and it is not reconnecting.

lincolnthree:
I think this might be related: https://github.com/ReactiveX/rxjs/pull/5634/files#diff-44fa2a928f593fdd8b981dac7737a65f6b0ddc844300af18a539c59b683b5467R130 

Looks like shareReplay now calls `complete()` on the subscribed subject/connector when the source completes. I don't think this this was in the original `shareReplay` behavior?

Am I missing something?

henry-alakazhang:
I'm experiencing this as well. Specifically, it happens when you subscribe multiple times, then unsubscribe and resubscribe again (is this called a reentrant subscription?). The code in the original post is a good example since multiple things subscribe to `shared$`.

Specifically, it happens if the code emits and unsubscribes immediately. If it emits after the first subscription, things seem to work as expected. It seems like it's a race condition in the setup code and the reset at refcount 0 logic.

Excuse my atrocious console logging, but this is what I see when a subscription is made and the sourece emits after a delay:

<img width="341" alt="image" src="https://user-images.githubusercontent.com/8922125/174015877-a9cebce5-6c8a-4350-abef-82b94a45330c.png">

Whereas this is what happens if the source emits immediately:

[...]

PR #7005, opened 2022-06-29
fix(share): Prevent setup/reset race condition in share with refCount

henry-alakazhang:
**Description:** Fixes the bug described in #6760, which seems to be a race condition.

The conditions are as such:
- Have an observable that emits immediately, but does not complete (eg. a `BehaviorSubject`, or something with `startWith`)
- Share it with `shareReplay` (or any `share` with `refCount: true`)
- Subscribe multiple times simultaneously, then unsubscribe immediately (can happen with `firstValueFrom`)
- The shared observable then gets into a bad state where it stays subscribed to the source observable but doesn't emit new values when next subscribed to.

The reason this happens is because `share` adds the reset subscription before subscribing to the source. As a result, if we subscribe and unsubscribe simultaneously, the reset block can actually run before subscribing to the source. When it gets reset, the `connection` state gets reset as well, and it subscribes to the source again even though it's already been reset.

In the PR, I've fixed it via adding an extra check before subscribing to make sure we actually need to subscribe to the source. I feel like this is only patches over the issue and some other edge case may still be broken, but there are some comments in the code saying that the reset subscription needs to be before the source subscription, so I didn't want to move anything else around.

I added a test as well, though it tests a symptom (an excess subscription) rather than the real impact (doesn't emit again later). Couldn't get the latter working.

**BREAKING CHANGE:** No

**Related issue (if exists):** #6760

### DIFF
--- a/spec/operators/shareReplay-spec.ts
+++ b/spec/operators/shareReplay-spec.ts
@@ -2,7 +2,7 @@ import { expect } from 'chai';
 import * as sinon from 'sinon';
 import { shareReplay, mergeMapTo, retry, take } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
-import { Observable, Operator, Observer, of, from, defer, pipe } from 'rxjs';
+import { Observable, Operator, Observer, of, from, defer, pipe, combineLatest, firstValueFrom, BehaviorSubject } from 'rxjs';
 import { observableMatcher } from '../helpers/observableMatcher';
 
 /** @test {shareReplay} */
@@ -271,6 +271,20 @@ describe('shareReplay', () => {
     }).pipe(shareReplay({ bufferSize: 1, refCount: true }));
     source.subscribe();
     source.subscribe();
+    expect(subscriptions).to.equal(1);
+  });
+
+  it('should only subscribe once each with multiple synchronous subscriptions and unsubscriptions ', async () => {
+    // This may seem very specific, but it's a regression test for https://github.com/ReactiveX/rxjs/issues/6760
+
+    let subscriptions = 0;
+    const source = defer(() => {
+      ++subscriptions;
+      // Needs to be an observable that doesn't complete
+      return new BehaviorSubject(1);
+    }).pipe(shareReplay({ bufferSize: 1, refCount: true }));
+
+    firstValueFrom(combineLatest([source, source]));
     expect(subscriptions).to.equal(1);
   });
 
--- a/spec/operators/shareReplay-spec.ts
+++ b/spec/operators/shareReplay-spec.ts
@@ -284,7 +284,7 @@ describe('shareReplay', () => {
       return new BehaviorSubject(1);
     }).pipe(shareReplay({ bufferSize: 1, refCount: true }));
 
-    firstValueFrom(combineLatest([source, source]));
+    await firstValueFrom(combineLatest([source, source]));
     expect(subscriptions).to.equal(1);
   });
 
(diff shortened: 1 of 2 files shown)


jkossis:
@benlesh curious what your thoughts are on this. It looks like this may have cropped up with the simplification of multicast operators in 7.x.

benlesh:
This is not the bug you're looking for. haha.

**TLDR: This looks like a bug with `combineLatest`**

`refCount: true` should allow resubscription to the source if all subscriptions to the result are unsubscribed. In `firstValueFrom`, it immediately unsubscribes after getting the first value. Since the source given to `combineLatest([source, source])` emits synchronously, what happens is as follows:

1. Subscribe to `combineLatest` result
2. `combineLatest` subscribes to first reference to `source`
3. `refCount` in `source` is `0` (and no connection was made yet), so we subscribe, incrementing `refCount` to `1`.
4. `source` synchronously emits `1`. 
5. `firstValueFrom` gets that first value and resolves, [unsubscribing from source synchronously](https://github.com/ReactiveX/rxjs/blob/47fa8d555754b18887baf15e22eb3dd91bf8bfea/src/internal/firstValueFrom.ts#L61-L62).
6. Unsubscribing from source decrements refCount back down to 0.
7. `combineLatest` subscribes to second reference to `source`.
8. `refCount` in `source` is `0` (and no connection was made yet), so we subscribe, incrementing `refCount` to `1`.
9. `source` synchronously emits `1`. 
10. `firstValueFrom` is already resolved, this is a noop.
6. Unsubscribing from source decrements refCount back down to 0.

So... I'd argue that the bug is _actually_ in combineLatest, which should halt additional subscriptions if the outer subscription has closed. I'll have a fix for that shortly.

This fix, however, will cause a regression. In that this is expected behavior:

```ts
let subscriptions = 0;
const source = new Observable((subscriber) => {
  subscriptions++;
  subscriber.next(1);
}).pipe(shareReplay({ bufferSize: 1, refCount: true }));

source.subscribe().unsubscribe();
source.subscribe();

console.assert(subscriptions === 2, 'subscriptions should be 2');
```

spec/operators/shareReplay-spec.ts
```diff
@@ -274,6 +274,20 @@ describe('shareReplay', () => {
...
+    // This may seem very specific, but it's a regression test for https://github.com/ReactiveX/rxjs/issues/6760
+
+    let subscriptions = 0;
+    const source = defer(() => {
+      ++subscriptions;
+      // Needs to be an observable that doesn't complete
+      return new BehaviorSubject(1);
+    }).pipe(shareReplay({ bufferSize: 1, refCount: true }));
+
+    firstValueFrom(combineLatest([source, source]));
```
benlesh, line 287: did you mean to await this?

benlesh, line 287:
```suggestion
    await firstValueFrom(combineLatest([source, source]));
```

benlesh:
Yeah. Exploring this, I see the issue. It also looks like it's time for me to clean up share (again) 😩 . I experimented with it a bit, and I think this is the best fix for now. Thank you.

jkossis:
@benlesh is there any way to get a minor/patch publish for including this change on 7.x?

jkossis:
@benlesh it seems like part of the issue (or a tangential issue) is, with the refactor of `share`, we lost some of the granularity previously provided. 

For example, `publishReplay + refCount` used to unsubscribe from source when the reference count dropped to 0, while preserving the underlying `ReplaySubject`. With the newly configurable `share`, those two things are coupled with `resetOnRefCountZero`. So you can have the behavior of unsubscribe from source + generate new `ReplaySubject`, or neither. But not a mix.

tl;dr `resetOnRefCountZero` has become a proxy for `shareReplay`'s `refCount` configuration. We have lost the functionality previously provided by `publishReplay + refCount`.

PR #7019, opened 2022-07-10
withLatestFrom refactor

Yozmo:
<!--

**Description:**
Replace the `hasValue` array used to mark whether or not the other sources have emitted a value. Using `remainingValues` which it's initialized with the number of `inputs.length` and a boolean `hasFirstValue` (which mark if the current inner observable has emitted its first value or not) can easily replace the old array.


Yozmo:
Closing the PR because includes by mistake some changes that are not supposed to be in the PR.

PR #7045, opened 2022-08-24
Fix wording

ic-768:
"Create customize..." is grammatically incorrect and was changed based on the context.


benlesh (approved):

benlesh:
Thank you, @ic-768

PR #7052, opened 2022-09-01
fix(windowWhen): don't signal on complete

### LINKED ISSUE #6627: Update marble diagram images for operators that accept "notifiers"
Since RxJS v7 has been released, many operators that accept "notifiers" have been adjusted to listen to only "next" notifications. This has become mandatory as per [RxJS Core Semantics](https://github.com/ReactiveX/rxjs/blob/7dd7f3c4ab/docs_app/content/guide/core-semantics.md) guide:

> Operators that accept "notifiers" (as described above), MUST ONLY recognized next values from the notifier as "notifications". Emitted completions may not be used a source of notification.

However, many marble diagram images have to be updated as they don't reliably picture this behavior. For example, `audit` operator displays this image:
audit

A user "reading" this image might be deceived to think that an Observable that has been passed to `audit` operator needs to complete before destination will receive next "next" notification. This is not true now.

Marble diagram images of these operators need to be updated:
- [x] [audit](https://rxjs.dev/api/operators/audit) (fixed in PR: #6636)
- [x] [debounce](https://rxjs.dev/api/operators/debounce) (fixed in PR: #6639)
- [x] [throttle](https://rxjs.dev/api/operators/throttle) (fixed in PR: #6638)
- [x] [bufferWhen](https://rxjs.dev/api/operators/bufferWhen) (fixed in PR: #6767)
- [ ] [windowWhen](https://rxjs.dev/api/operators/windowWhen) (PR: #7052)

### LINKED ISSUE #6876: windowWhen should NOT be accepting completions as notification signals!
See tests linked here:

https://github.com/ReactiveX/rxjs/issues/6627#issuecomment-1014303852

jakovljevic-mladen:
**Description:**
This PR is split to two commits:
- in the first, it aligns `windowWhen` with `bufferWhen` - it fixes a now considered bug, so that it stops accepting `complete` notifications from Observable emitted by `closingSelector`;
- in another commit, it updates documentation and marble diagram:
  

Closes #6876
Closes #6627


benlesh:
Haven't reviewed yet, but I know the bug. It IS a bug, but the fix is also a breaking change. So we need to punt until next major.

jakovljevic-mladen:
Switching to `8.x` branch.

Issue #7018, opened 2022-07-08
debounceTime operator with animationFrameScheduler stops working unexpectedly


EoinGriffin-AI:
### Describe the bug

I've got a problem this (custom) operator in RxJS (7.5.4). It just... stops working sometimes. What's going on? How can I fix this?

```import { animationFrameScheduler, Observable, OperatorFunction } from 'rxjs';
import { debounceTime, tap } from 'rxjs/operators';


export function debounceAnimated<T>(): OperatorFunction<T, T> {
  return (source$: Observable<T>) => source$.pipe(
    tap(() => console.log('A')),
    debounceTime(0, animationFrameScheduler),
    tap(() => console.log('B')),
  );
}```

This operator works just fine most of the time. I see in my app that the action using this operator works correctly, and in the console I see a bunch of A's followed by a bunch of B's, as expected when the debounceTime fires.

The problem is that sometimes this operator just... breaks somehow. I get a whole bunch of A's when I perform the user action (in this case using the mouse scroll to zoom in and out of a chart), but the B's stop completely and downstream observable subscriptions stop working too. The application doesn't recover when this happens. Everything else seems to be fine, CPU usage is low, and other components that don't use this operator continue to work normally.

-------
-------

For a bit more context, this feature has been working fine for a few years, but a recent RxJS upgrade (we upgraded from 6.6.2 to 7.5.4 to get a necessary bugfix) broke the previous implementation of this operator and so now I'm trying to find some alternative. The previous implementation used this code, but it no longer debounces anything when written like this: ```debounce(() => EMPTY.pipe(observeOn(scheduler)))```

I also posted this same question on StackOverflow, in case anyone wants to follow there too.
https://stackoverflow.com/questions/72916854/rxjs-debouncetime-operator-with-animationframescheduler-stops-working-unexpected

### Expected behavior

debounceTime with animationFrameScheduler should continue to work indefinitely as long as there are active subscribers.

### Reproduction code

```typescript
In my case this code is run many times. Several Observables (maybe 10?) use this custom debounceAnimated operator and it gets his about 50 times and so logs 'A' 50 times on a single mouse scroll event. Obviously I don't want to re-render my UI 50 times, which is why I need this to work.

 { animationFrameScheduler, Observable, OperatorFunction } from 'rxjs';
import { debounceTime, tap } from 'rxjs/operators';


export function debounceAnimated<T>(): OperatorFunction<T, T> {
  return (source$: Observable<T>) => source$.pipe(
[...]

PR #7059, opened 2022-09-08
fix(schedulers): improve performance of animationFrameScheduler and asapScheduler

### LINKED ISSUE #7017: Performance issue with multiple concurrent uses of animationFrameScheduler
### Describe the bug

Since version 7.5.0, there is a huge performance loss with animationFrameScheduler when there are a lot of timers emitting concurrently using the scheduler. The problem can be painfully obvious on slower machines and was found in our application using Angular CDK virtual scroll.

There was no such performance issue in version 7.4.0. The problem have probably been created by [this code change](https://github.com/reactivex/rxjs/commit/e35f589e2ca10ab2d2d69f7e9fe60727edc4c53d)

The performance issue is big enough to prevent the migration of our application to RxJs 7.5 + as it makes some parts of it near unusable.

### Expected behavior

Independant observables using the animationFrameScheduler should not have a big performance impact on each others.

### Reproduction code

```typescript
See the Stackblitz code.

See the result in the console using rxJs 7.5.0 vs using rxJs 7.4.0. The result is dependant on your computer's performance, but should be quite obvious regardless
```

### Reproduction URL

https://stackblitz.com/edit/rxjs-afsbug?file=package.json,index.ts

### Version

7.5.0

### Environment

_No response_

### Additional context

Works properly and is a lot more performant before 7.5.0

benlesh:
+ Changes the check for existing action ids to simply check the last action in the queue to see if its id matches. Previously we were doing an O(n) loop on each execution of an action to check to see if the scheduling id needed to be recycled. This is problematic in AsapScheduler and AnimationFrameScheduler, where we're not reusing an interval. Since AsapScheduler and AnimationFrameScheduler reuse the most recent action id until their scheduled microtask or animation frame fires, the last action in the actions queue array is all we really need to check (rather than checking them all with `some`). O(1) vs O(n).
+ Refactors a weird conditional gaff from `if ((X && A) || (!X && B))` to just be `if (X ? A : B)`
+ Appropriately types action ids
    
resolves #7017
related #7018
related #6674

### DIFF
--- a/src/internal/scheduler/AnimationFrameAction.ts
+++ b/src/internal/scheduler/AnimationFrameAction.ts
@@ -2,13 +2,14 @@ import { AsyncAction } from './AsyncAction';
 import { AnimationFrameScheduler } from './AnimationFrameScheduler';
 import { SchedulerAction } from '../types';
 import { animationFrameProvider } from './animationFrameProvider';
+import { TimerHandle } from './timerHandle';
 
 export class AnimationFrameAction<T> extends AsyncAction<T> {
   constructor(protected scheduler: AnimationFrameScheduler, protected work: (this: SchedulerAction<T>, state?: T) => void) {
     super(scheduler, work);
   }
 
-  protected requestAsyncId(scheduler: AnimationFrameScheduler, id?: any, delay: number = 0): any {
+  protected requestAsyncId(scheduler: AnimationFrameScheduler, id?: TimerHandle, delay: number = 0): TimerHandle {
     // If delay is greater than 0, request as an async action.
     if (delay !== null && delay > 0) {
       return super.requestAsyncId(scheduler, id, delay);
@@ -20,7 +21,8 @@ export class AnimationFrameAction<T> extends AsyncAction<T> {
     // the current animation frame request id.
     return scheduler._scheduled || (scheduler._scheduled = animationFrameProvider.requestAnimationFrame(() => scheduler.flush(undefined)));
   }
-  protected recycleAsyncId(scheduler: AnimationFrameScheduler, id?: any, delay: number = 0): any {
+
+  protected recycleAsyncId(scheduler: AnimationFrameScheduler, id?: TimerHandle, delay: number = 0): TimerHandle | undefined {
     // If delay exists and is greater than 0, or if the delay is null (the
     // action wasn't rescheduled) but was originally scheduled as an async
     // action, then recycle as an async action.
@@ -30,8 +32,8 @@ export class AnimationFrameAction<T> extends AsyncAction<T> {
     // If the scheduler queue has no remaining actions with the same async id,
     // cancel the requested animation frame and set the scheduled flag to
     // undefined so the next AnimationFrameAction will request its own.
-    if (!scheduler.actions.some((action) => action.id === id)) {
-      animationFrameProvider.cancelAnimationFrame(id);
+    if (id != null && !scheduler.actions.some((action) => action.id === id)) {
+      animationFrameProvider.cancelAnimationFrame(id as number);
       scheduler._scheduled = undefined;
     }
     // Return undefined so the action knows to request a new async id if it's rescheduled.
--- a/src/internal/scheduler/AnimationFrameAction.ts
+++ b/src/internal/scheduler/AnimationFrameAction.ts
@@ -26,13 +26,14 @@ export class AnimationFrameAction<T> extends AsyncAction<T> {
     // If delay exists and is greater than 0, or if the delay is null (the
     // action wasn't rescheduled) but was originally scheduled as an async
     // action, then recycle as an async action.
-    if ((delay != null && delay > 0) || (delay == null && this.delay > 0)) {
+    if (delay != null ? delay > 0 : this.delay > 0) {
       return super.recycleAsyncId(scheduler, id, delay);
     }
     // If the scheduler queue has no remaining actions with the same async id,
     // cancel the requested animation frame and set the scheduled flag to
     // undefined so the next AnimationFrameAction will request its own.
-    if (id != null && !scheduler.actions.some((action) => action.id === id)) {
+    const { actions } = scheduler;
+    if (id != null && actions[actions.length - 1]?.id !== id) {
       animationFrameProvider.cancelAnimationFrame(id as number);
       scheduler._scheduled = undefined;
     }
--- a/src/internal/scheduler/AsapAction.ts
+++ b/src/internal/scheduler/AsapAction.ts
@@ -26,13 +26,14 @@ export class AsapAction<T> extends AsyncAction<T> {
     // If delay exists and is greater than 0, or if the delay is null (the
     // action wasn't rescheduled) but was originally scheduled as an async
     // action, then recycle as an async action.
-    if ((delay != null && delay > 0) || (delay == null && this.delay > 0)) {
+    if (delay != null ? delay > 0 : this.delay > 0) {
       return super.recycleAsyncId(scheduler, id, delay);
     }
     // If the scheduler queue has no remaining actions with the same async id,
     // cancel the requested microtask and set the scheduled flag to undefined
     // so the next AsapAction will request its own.
-    if (id != null && !scheduler.actions.some((action) => action.id === id)) {
+    const { actions } = scheduler;
+    if (id != null && actions[actions.length - 1]?.id !== id) {
       immediateProvider.clearImmediate(id);
       scheduler._scheduled = undefined;
     }
--- a/src/internal/scheduler/AsapAction.ts
+++ b/src/internal/scheduler/AsapAction.ts
@@ -2,13 +2,14 @@ import { AsyncAction } from './AsyncAction';
 import { AsapScheduler } from './AsapScheduler';
 import { SchedulerAction } from '../types';
 import { immediateProvider } from './immediateProvider';
+import { TimerHandle } from './timerHandle';
 
 export class AsapAction<T> extends AsyncAction<T> {
   constructor(protected scheduler: AsapScheduler, protected work: (this: SchedulerAction<T>, state?: T) => void) {
     super(scheduler, work);
   }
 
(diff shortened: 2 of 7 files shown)


src/internal/scheduler/AnimationFrameAction.ts
```diff
@@ -2,13 +2,14 @@ import { AsyncAction } from './AsyncAction';
...
 import { animationFrameProvider } from './animationFrameProvider';
+import { TimerHandle } from './timerHandle';
 
 export class AnimationFrameAction<T> extends AsyncAction<T> {
   constructor(protected scheduler: AnimationFrameScheduler, protected work: (this: SchedulerAction<T>, state?: T) => void) {
     super(scheduler, work);
   }
 
-  protected requestAsyncId(scheduler: AnimationFrameScheduler, id?: any, delay: number = 0): any {
+  protected requestAsyncId(scheduler: AnimationFrameScheduler, id?: TimerHandle, delay: number = 0): TimerHandle {
```
benlesh, line 12: Most changed lines are just around typing these ids appropriately.

src/internal/scheduler/AnimationFrameAction.ts
```diff
@@ -20,18 +21,20 @@ export class AnimationFrameAction<T> extends AsyncAction<T> {
...
     return scheduler._scheduled || (scheduler._scheduled = animationFrameProvider.requestAnimationFrame(() => scheduler.flush(undefined)));
   }
-  protected recycleAsyncId(scheduler: AnimationFrameScheduler, id?: any, delay: number = 0): any {
+
+  protected recycleAsyncId(scheduler: AnimationFrameScheduler, id?: TimerHandle, delay: number = 0): TimerHandle | undefined {
     // If delay exists and is greater than 0, or if the delay is null (the
     // action wasn't rescheduled) but was originally scheduled as an async
     // action, then recycle as an async action.
-    if ((delay != null && delay > 0) || (delay == null && this.delay > 0)) {
+    if (delay != null ? delay > 0 : this.delay > 0) {
```
benlesh, line 29: Simplify conditional `((X && A) || (!X && B))` to `(X ? A : B)`

src/internal/scheduler/AsapAction.ts
```diff
@@ -20,17 +21,19 @@ export class AsapAction<T> extends AsyncAction<T> {
...
     return scheduler._scheduled || (scheduler._scheduled = immediateProvider.setImmediate(scheduler.flush.bind(scheduler, undefined)));
   }
-  protected recycleAsyncId(scheduler: AsapScheduler, id?: any, delay: number = 0): any {
+
+  protected recycleAsyncId(scheduler: AsapScheduler, id?: TimerHandle, delay: number = 0): TimerHandle | undefined {
     // If delay exists and is greater than 0, or if the delay is null (the
     // action wasn't rescheduled) but was originally scheduled as an async
     // action, then recycle as an async action.
-    if ((delay != null && delay > 0) || (delay == null && this.delay > 0)) {
+    if (delay != null ? delay > 0 : this.delay > 0) {
```
benlesh, line 29: Simplify conditional `((X && A) || (!X && B))` to `(X ? A : B)`

src/internal/scheduler/AsapAction.ts
```diff
@@ -20,17 +21,19 @@ export class AsapAction<T> extends AsyncAction<T> {
...
       return super.recycleAsyncId(scheduler, id, delay);
     }
     // If the scheduler queue has no remaining actions with the same async id,
     // cancel the requested microtask and set the scheduled flag to undefined
     // so the next AsapAction will request its own.
-    if (!scheduler.actions.some((action) => action.id === id)) {
+    const { actions } = scheduler;
+    if (id != null && actions[actions.length - 1]?.id !== id) {
```
benlesh, line 36: Perf improvement here.  `arr.some(x => x.id === id)` to `arr[arr.length - 1]?.id === id`. Which is O(n) to O(1) in hot path code.

src/internal/scheduler/AnimationFrameAction.ts
```diff
@@ -20,18 +21,20 @@ export class AnimationFrameAction<T> extends AsyncAction<T> {
...
       return super.recycleAsyncId(scheduler, id, delay);
     }
     // If the scheduler queue has no remaining actions with the same async id,
     // cancel the requested animation frame and set the scheduled flag to
     // undefined so the next AnimationFrameAction will request its own.
-    if (!scheduler.actions.some((action) => action.id === id)) {
-      animationFrameProvider.cancelAnimationFrame(id);
+    const { actions } = scheduler;
+    if (id != null && actions[actions.length - 1]?.id !== id) {
```
benlesh, line 36: Perf improvement here.  `arr.some(x => x.id === id)` to `arr[arr.length - 1]?.id === id`. Which is O(n) to O(1) in hot path code.

src/internal/scheduler/AsyncAction.ts
```diff
@@ -58,23 +59,26 @@ export class AsyncAction<T> extends Action<T> {
...
+  protected recycleAsyncId(_scheduler: AsyncScheduler, id?: TimerHandle, delay: number | null = 0): TimerHandle | undefined {
     // If this action is rescheduled with the same delay time, don't clear the interval id.
     if (delay != null && this.delay === delay && this.pending === false) {
       return id;
     }
     // Otherwise, if the action's delay time is different from the current delay,
     // or the action has been rescheduled before it's executed, clear the interval id
-    intervalProvider.clearInterval(id);
+    if (id != null) {
+      intervalProvider.clearInterval(id);
```
benlesh, line 79: This was being called with `undefined` at times before.

src/internal/scheduler/QueueAction.ts
```diff
@@ -21,20 +20,21 @@ export class QueueAction<T> extends AsyncAction<T> {
...
 
     if ((delay != null && delay > 0) || (delay == null && this.delay > 0)) {
       return super.requestAsyncId(scheduler, id, delay);
     }
+
     // Otherwise flush the scheduler starting with this action.
-    return scheduler.flush(this);
+    scheduler.flush(this);
+
+    return 1;
```
benlesh, line 38: Previously, this was returning `true`, which isn't a `TimerHandle` type. I changed it to a `1`. Perhaps less memory efficient, but it's code-efficient and in-character for a `TimerHandle`.

cartant: Was this really returning `true`? AFAICT, `QueueScheduler` has no `flush`:

https://github.com/ReactiveX/rxjs/blob/4afbc16f053ffbfbd4f45d677056ff00d7711de7/src/internal/scheduler/QueueScheduler.ts#L3-L4

And the `flush` in `AsyncScheduler` - which `QueueScheduler` extends - returns `void`:

https://github.com/ReactiveX/rxjs/blob/4afbc16f053ffbfbd4f45d677056ff00d7711de7/src/internal/scheduler/AsyncScheduler.ts#L27

IDK what the consequences of returning something truthy will be when it previously appeared to return something falsy.

cartant (commented):
I've read through this and checking the `id` of only the last element in the `actions` array makes sense. I'm not sure about returning `1` where the return value of `flush` used to be returned, though - I left a comment. I'll have another look at it in the morning. A little too tired to think that one through ATM.

src/internal/scheduler/AnimationFrameAction.ts
```diff
@@ -20,18 +21,20 @@ export class AnimationFrameAction<T> extends AsyncAction<T> {
...
+      animationFrameProvider.cancelAnimationFrame(id as number);
       scheduler._scheduled = undefined;
```
cartant, line 38: This doesn't relate to a change that's made in this PR, but I'm not at all sure that setting `_scheduled` to `undefined` here is correct (this happens in the `AsapScheduler`, too). I think this was done - pretty sure, by me - when looking at explicit unsubscriptions of actions - because the `unsubscribe` calls `recycleAsyncId`:

https://github.com/ReactiveX/rxjs/blob/4afbc16f053ffbfbd4f45d677056ff00d7711de7/src/internal/scheduler/AsyncAction.ts#L140

But if we've come into this block under different circumstances, it might make no sense. E.g. if there's an action at the end of the array that was scheduled from within a flush - and has a different `id` - there'll have been a corresponding `requestAnimationFrame` call and `_scheduled` will be holding the handle for that call.

This stuff is so complicated. Will look again at this tomorrow, but I have a bad feeling it's wrong. (Again, not a change in this PR; it was already like this.)

benlesh:
@cartant on `QueueAction` `requestAsyncId` returning `1`. It was actually returning `void` before, now that I look at it. However, that's not a valid `TimerHandle` and the (obviously crazy) API is expecting a `TimerHandle` to be returned, even though it's not really using it in this case. I could change it to `0` so that it's a valid `TimerHandle`, but also falsy. OR I could go through and make all of the other `requestAsyncId` overrides return `TimerHandle | undefined`. Personally, I think that returning `0` and adding a comment might be the more prudent action here, so I'll preemptively do that.

This whole area of the library isn't long for this world. I want to deprecate it and replace it in version 8 with it fully phased out by version 9.

cartant (approved):
> Personally, I think that returning 0 and adding a comment might be the more prudent action here, so I'll preemptively do that.

I think that will be fine. Looking at the code, the worst thing that seems possible is that if `schedule` is called on the action and a delay is specified, `clearInterval` will be called with 0, here:

https://github.com/ReactiveX/rxjs/blob/4afbc16f053ffbfbd4f45d677056ff00d7711de7/src/internal/scheduler/AsyncAction.ts#L77

because this 0 will pass check:

https://github.com/ReactiveX/rxjs/blob/4afbc16f053ffbfbd4f45d677056ff00d7711de7/src/internal/scheduler/AsyncAction.ts#L51-L53

And that should be no big deal.

Later, I'll have a look at the possible problem I mentioned [here](https://github.com/ReactiveX/rxjs/pull/7059#discussion_r967057021), but that shouldn't hold up this PR.

PR #7060, opened 2022-09-09
perf(animationFrames): uses fewer Subscription instances

benlesh:
Leverages a single id variable instead of creating a new child Subscription for each scheduled animation frame. This is important because animation frames are scheduled very rapidly and this helps prevent CPU and GC thrashing.

Related #7018


benlesh:
NOTE: Will do a follow up to remove `animationFrameProvider.schedule`, since it's not public and we're not using it after this PR.

cartant (approved):

PR #7073, opened 2022-09-29
feat(buffer): closingNotifier should support ObservableInput

jakovljevic-mladen:
**Description:**
This PR adds support for `buffer`'s `closingNotifier` to accept `ObservableInput`.

#6972

### DIFF
--- a/spec/operators/buffer-spec.ts
+++ b/spec/operators/buffer-spec.ts
@@ -1,5 +1,5 @@
 import { buffer, mergeMap, take, window, toArray } from 'rxjs/operators';
-import { EMPTY, NEVER, throwError, of, Subject } from 'rxjs';
+import { EMPTY, NEVER, throwError, of, Subject, interval } from 'rxjs';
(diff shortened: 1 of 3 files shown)


spec/operators/buffer-spec.ts
```diff
@@ -324,6 +324,43 @@ describe('Observable.prototype.buffer', () => {
...
   });
 
+  it('should buffer when Promise resolves', (done) => {
+    const e1 = interval(3).pipe(take(5));
+    const expected = [
+      [0, 1],
+      [2, 3, 4],
+    ];
+
+    e1.pipe(buffer(new Promise<void>((resolve) => setTimeout(() => resolve(), 8)))).subscribe({
```
benlesh, line 334: Nit: (Can be fixed in a FLUP)... we don't need to wait 8 ms... we can just do `Promise.resolve()`.

benlesh (approved):

Issue #5312, opened 2020-02-13
WebSocket: Connection not closed when sole Subscription isStopped after an error


kostadin-mandulov:
## Bug Report

**Current Behavior**

When sole `WebSocketSubject` `Subscription` encounters an error, it `isStopped` and no longer handles anything, but the underlying connection is not closed.

**Reproduction**

Source: [rxjs-ws-error-handling-test.zip](https://github.com/ReactiveX/rxjs/files/4197650/rxjs-ws-error-handling-test.zip)

1. Have `node` and `npm` installed. I've used versions `12.15.0` and `6.13.7` respectively. 
2. `npm install` 
3. `node server.js`
4. Make sure the `index.html` is served via `localhost`. I've used Python 3: `python3 -m http.server 1234`
5. Open your favourite modern browser and open its DevTools.
6. Navigate to `http://localhost:1234/`
7. Observing the logs in the console and the WS connection, we can see the following behaviour:
    - [0s] Connection is established (`WebSocketSubject`'s `openObserver` handler called).
    - [0s] A message is received by the server (`Subscription`'s `next` handler called).
    - [2s] A faulty message causes an error (`Subscription`'s `error` handler called). The `Subscription` is now stopped, but the WebSocket connection remains opened (DevTools > Network > WS).
    - [4s] A message is received by the WebSocket connection but `Subscription`'s `next` handler is NOT called.
    - [6s] The server closes the connection (`WebSocketSubject`'s `closeObserver` handler called).

**Expected behavior**

If all of a `WebSocketSubject`'s Subscriptions are stopped / closed, I believe the connection should also close.

**Environment**

- Runtime: 
    - Node v12.15.0
    - NPM v6.13.7
    - Chrome v79.0.3945.130
    - Firefox v73.0
    - Microsoft Edge v44.18362.449.0
- RxJS version: 6.5.4

keradus:
I hit the same problem. any help on this one, folks ?

ZenwalkerD:
I think i am facing same issue! Any solution? In dev window i still see that specific WS connection shown as 101 but not Finished even though i close the WebSocketSubject or do complete() call on it.

splincode:
any news?

kostadin-mandulov:
[...]

PR #7077, opened 2022-09-29
fix(WebSocketSubject): Closes socket when closing socket observer

bever1337:
**Description:**

When the output observer is closed because of de-serialization errors, ensure the WebSocketSubject is reset and the WebSocket is closed.

**Related issue (if exists):** (#5312)

### DIFF
--- a/src/internal/observable/dom/WebSocketSubject.ts
+++ b/src/internal/observable/dom/WebSocketSubject.ts
@@ -359,7 +359,12 @@ export class WebSocketSubject<T> extends AnonymousSubject<T> {
         const { deserializer } = this._config;
         observer.next(deserializer!(e));
       } catch (err) {
+        this._resetState();
         observer.error(err);
+        if (socket && (socket.readyState === 1 || socket.readyState === 0)) {
+          // close the socket after the observer so additional WebSocket events do not race to end the observer
+          socket.close();
+        }
       }
     };
   }


bever1337:
I don't believe there's any marble testing to be done since this work is making native features observable. In my experience, there is no equivalent framework like `msw` for mocking and testing websockets.

bever1337:
I need to research further, but I think the multiplexer can also create the same state where the websocket subject has errored but the underlying socket remains open.

maknapp:
There is a `MockWebSocket` in `webSocket-spec.ts` to write tests with. There seems to be only [one test](https://github.com/ReactiveX/rxjs/blob/master/spec/observables/dom/webSocket-spec.ts#L398) for deserializer errors, and it does not check for ws.close. There are certainly more tests to add in this area.

benlesh:
Hi @bever1337 — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

PR #7081, opened 2022-10-09
feat(distinct): `flush` supports `ObservableInput`

jeremymwells:
**Description:**
This PR adds support for distinct's flush argument to accept ObservableInput.

https://github.com/ReactiveX/rxjs/issues/6367

### DIFF
--- a/src/internal/operators/distinct.ts
+++ b/src/internal/operators/distinct.ts
@@ -1,8 +1,8 @@
-import { Observable } from '../Observable';
-import { MonoTypeOperatorFunction } from '../types';
+import { MonoTypeOperatorFunction, ObservableInput } from '../types';
 import { operate } from '../util/lift';
 import { createOperatorSubscriber } from './OperatorSubscriber';
 import { noop } from '../util/noop';
+import { innerFrom } from '../observable/innerFrom';
 
 /**
  * Returns an Observable that emits all items emitted by the source Observable that are distinct by comparison from previous items.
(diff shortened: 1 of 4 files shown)


src/internal/operators/distinct.ts
```diff
@@ -74,6 +75,11 @@ export function distinct<T, K>(keySelector?: (value: T) => K, flushes?: Observab
       })
     );
 
-    flushes?.subscribe(createOperatorSubscriber(subscriber, () => distinctKeys.clear(), noop));
+    let flush$ = flushes as InteropObservable<any> | ObservableInput<any> | undefined | any;
+    if (flushes && isInteropObservable(flushes)) {
+      flush$ = fromInteropObservable(flushes);
+    }
+
+    flush$?.subscribe(createOperatorSubscriber(subscriber, () => distinctKeys.clear(), noop));
```
jakovljevic-mladen, line 83: Hey @jeremymwells, thanks for the PR. Instead of using `fromInteropObservable`, you should be using `innerFrom`, like this:

```suggestion
    flushes && innerFrom(flushes).subscribe(createOperatorSubscriber(subscriber, () => distinctKeys.clear(), noop));
```

src/internal/operators/distinct.ts
```diff
@@ -57,11 +58,11 @@ import { noop } from '../util/noop';
  * @see {@link distinctUntilKeyChanged}
  *
  * @param {function} [keySelector] Optional function to select which value you want to check as distinct.
- * @param {Observable} [flushes] Optional Observable for flushing the internal HashSet of the operator.
+ * @param {ObservableInput} [flushes] Optional ObservableInput for flushing the internal HashSet of the operator.
```
jakovljevic-mladen, line 61: Also, if you can, please remove anything between `{}`. The docs generator scripts are smart enough to understand params' type only by writing their name. If you'd write it like this:

```suggestion
 * @param flushes Optional `ObservableInput` for flushing the internal HashSet of the operator.
```

That would be nice. Fixing it for `keySelector` would be nice as well. Thanks.

jeremymwells: Thanks, @jakovljevic-mladen. Cheers. Just pushed the changes.

benlesh (approved):

PR #7088, opened 2022-10-17
feat(window): windowBoundaries should support ObservableInput

jakovljevic-mladen:
Description:
This PR adds support for `window`'s `windowBoundaries` to accept `ObservableInput`.

Related issue (if exists):
#6972

### DIFF
--- a/spec-dtslint/operators/window-spec.ts
+++ b/spec-dtslint/operators/window-spec.ts
@@ -6,5 +6,10 @@ it('should infer correctly', () => {
 });
 
 it('should enforce types', () => {
-  of(1).pipe(window('')); // $ExpectError
+  of(1).pipe(window()); // $ExpectError
+  of(1).pipe(window(6)); // $ExpectError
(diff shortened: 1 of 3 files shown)


spec-dtslint/operators/window-spec.ts
```diff
@@ -6,5 +6,10 @@ it('should infer correctly', () => {
 });
 
 it('should enforce types', () => {
-  of(1).pipe(window('')); // $ExpectError
```
jakovljevic-mladen, line 9: Since `String` [implements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/@@iterator) `Iterable` interface and since `Iterable` is declared in `ObservableInput`, `window` now accepts string as a valid input type. I didn't want to move this test to `'should infer correctly'` section so I don't confuse anyone reading these tests.

benlesh (approved):

PR #7114, opened 2022-11-11
fix(webSocket): manage serializer throwing exceptions

### LINKED ISSUE #7111: "WebSocketSubject.error must be called with an object with an error code" when serializer throws exception
### Describe the bug

When sending a message to a `WebSocketSubject` that causes the serializer to throw an exception, rxjs throws instead a different exception:

```
TypeError: WebSocketSubject.error must be called with an object with an error code, and an optional reason: { code: number, reason: string }
    at Object.eval [as error] (WebSocketSubject.js:156:26)
    at ConsumerObserver.error (Subscriber.js:118:25)
    at Subscriber._error (Subscriber.js:82:24)
    at Subscriber.error (Subscriber.js:59:12)
    at Object.eval [as next] (WebSocketSubject.js:145:31)
    at ConsumerObserver.next (Subscriber.js:108:25)
    at Subscriber._next (Subscriber.js:78:22)
    at Subscriber.next (Subscriber.js:51:12)
    at ReplaySubject._subscribe (ReplaySubject.js:55:18)
    at Observable._trySubscribe (Observable.js:43:19)
    at Subject._trySubscribe (Subject.js:117:43)
    at eval (Observable.js:37:115)
    at errorContext (errorContext.js:28:5)
    at Observable.subscribe (Observable.js:35:47)
    at socket.onopen [as _onopen] (WebSocketSubject.js:168:32)
```

### Expected behavior

The original error should be thrown instead

### Reproduction code

```typescript
const socket$ = webSocket({
  url: 'http://www.example.com/',
  serializer: message => {
    return JSON.stringify(message);
  }
});

socket$.subscribe({
  next: (v) => console.log('next', v),
  error: e => console.error('caught', e),
  complete: () => console.log('complete')
});

socket$.next({
  value: BigInt(5)
});
```

### Reproduction URL

https://codesandbox.io/s/lively-river-dqm1qr?file=/src/index.ts:197-527

### Version

7.5.7

### Environment

_No response_

### Additional context

_No response_

voliva:
Fixes #7111 

**Description:**

I see the behaviour of the serializer was not really specified, I added a couple of tests to cover it.

This is the solution that's less disruptive with the previous behaviour:

- If the serializer was throwing an object `{ code, reason }` it won't be effected by this change: The socket will be closed with those parameters.
- Otherwise, the socket is still getting closed, with the generic code 1000.
- To not swallow the exception, I've rethrown it. This way the consumer can catch errors when sending them (as if it's sending a message with an incorrect format). For this I had to change SafeSubscriber to Subscriber, as it wasn't propagating the exception back to `.next(` caller.

I think all of this points are open for discussion though. A couple of questions:

- **Should the connection get closed when a message can't be serialized?** Personally I think it shouldn't, but I didn't want to change this behaviour in this PR without discussing it.
- **Should the error happen on `.next(`, or should it be thrown on the observer?** I think throwing them on `.next(` makes more sense, because it lets the producer know there was an exception. The observer expects errors to come only from the websocket.


spec/observables/dom/webSocket-spec.ts
```diff
@@ -416,6 +416,52 @@ describe('webSocket', () => {
...
+      subject.subscribe({ next: (x: any) => {
+        expect(x).to.equal('this should not happen');
+      }, error: (err: any) => {
+        expect(err).to.equal('this should not happen');
+      } });
+
+      const socket = MockWebSocket.lastSocket;
+      socket.open();
+
+      expect(() => subject.next('weee!')).to.throw('I am a bad error');
```
benlesh, line 458: Issue: We wouldn't want this to synchronously throw. No call to `next()` should sync throw, really.

src/internal/observable/dom/WebSocketSubject.ts
```diff
@@ -296,18 +296,25 @@ export class WebSocketSubject<T> extends AnonymousSubject<T> {
...
+      this.destination = new Subscriber<T>({
+        next: (x: T) => {
           if (socket!.readyState === 1) {
             try {
               const { serializer } = this._config;
               socket!.send(serializer!(x!));
-            } catch (e) {
-              this.destination!.error(e);
+            } catch (e: any) {
+              if (e && e.code) {
```
benlesh, line 306: Nit: `e?.code` is a little more terse.

src/internal/observable/dom/WebSocketSubject.ts
```diff
@@ -296,18 +296,25 @@ export class WebSocketSubject<T> extends AnonymousSubject<T> {
 
       const queue = this.destination;
 
-      this.destination = Subscriber.create<T>(
-        (x) => {
+      this.destination = new Subscriber<T>({
```
benlesh, line 299: This is a solid change.

src/internal/observable/dom/WebSocketSubject.ts
```diff
@@ -296,18 +296,25 @@ export class WebSocketSubject<T> extends AnonymousSubject<T> {
...
+        next: (x: T) => {
           if (socket!.readyState === 1) {
             try {
               const { serializer } = this._config;
               socket!.send(serializer!(x!));
-            } catch (e) {
-              this.destination!.error(e);
```
benlesh, line 306: I'd say the actual bug was here. We clearly have a handle to the wrong thing. 🤔

benlesh (commented):
I don't think we can accept this PR as-is. There's clearly a bug, but we need to discuss what the correct behavior is.

In general, the WebSocketSubject should _emit_ the error... but also it seems like we should close the socket with the proper code when that happens.

spec/observables/dom/webSocket-spec.ts
voliva: I see your point, but at the same time the serializer kicks in on `subject.next()`.

In a way, the serializer option is just a shorthand so that the consumer doesn't need to do this every time:

```ts
subject.next(serialize(message))
```

which obviously would throw synchronously.

Just food for thought whenever you discuss about this.

voliva:
Hey @benlesh I've just updated the PR with the decision taken at #7111. Can you take another look please?

benlesh:
Hi @voliva — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

PR #7128, opened 2022-12-03
chore: remove ts-api-guardian, as it is outmoded

benlesh:
Because it was stupid that I ever listened to the person that wanted me to do this to begin with.


jakovljevic-mladen:
What about [`api_guard`](https://github.com/ReactiveX/rxjs/tree/master/api_guard) folder? Can you remove it as well?

jakovljevic-mladen (approved):

PR #7130, opened 2022-12-09
build(package): ensure NodeJS considers ESM .js files as ESM

devversion:
**Concrete issue**

When you import `rxjs` from NodeJS from an ESM file, the loading
of the RxJS files fails with syntax errors because NodeJS treats
them as CommonJS.

**Background:**

NodeJS treats JavaScript files as ESM when one of the following is true:

 * The file is explicitly named `.mjs`
 * The file is part of a directory where the closest `package.json` has
   `type: module`.

Both things are not applying for the ESM output in `dist/esm` and
`dist/esm5` so these ESM artifacts cannot be used directly in NodeJS
because NodeJS will attempt loading them as CommonJS.

Note that this was not noticeable with e.g. bundlers like Webpack
as those do not rely on the NodeJS semantics for "detecting" ESM.

#6321

### DIFF
--- a/tools/prepare-package.js
+++ b/tools/prepare-package.js
@@ -27,10 +27,14 @@ klawSync(ESM5_ROOT, {
 })
   .map((item) => item.path.slice(`${__dirname}/${ESM5_ROOT}`.length))
   .map((fileName) => {
-    if (!bo) {return fileName;}
+    if (!bo) {
+      return fileName;
+    }
     const fullPath = path.resolve(__dirname, ESM5_ROOT, fileName);
     // The file won't exist when running build_test as we don't create the ESM5 sources
-    if (!fs.existsSync(fullPath)) {return fileName;}
+    if (!fs.existsSync(fullPath)) {
+      return fileName;
+    }
     const content = fs.readFileSync(fullPath).toString();
     const transformed = bo.transformJavascript({
       content: content,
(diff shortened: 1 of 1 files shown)


benlesh:
Hey, @devversion, sorry for the slow response here. Anyone from the Angular team could have reached out to me directly to get my eyes on this sooner.

The problem I see is that this looks like a breaking change. So we'd be unable to release it until 8.x. Am I incorrect about that?

benlesh:
I think another problem is that we're not really outputting "real" ESM, like I don't think we have file extensions in the output, etc. So this would probably break in newer versions of NodeJS anyway?

devversion:
Thanks @benlesh! Good question with the breaking change. I think you're actually right that this would break the ecosystem as Webpack for example specifically works with this "invalid ESM", ignoring missing extensions **just** because there is no explicit `.mjs` extension, or `{type: module}` nearby. https://webpack.js.org/configuration/module/#resolvefullyspecified

So yeah, this seems only useful to v8. I saw you there is other planned ESM work for v8, so feel free to close this if it would rather interfere then.

devversion:
For NodeJS use specifically, I think the output is already broken and this would improve the situation, but still not solve it completely- due to the missing extensions. In practice, no NodeJS user is ever getting the ESM output of `rxjs` at all because the `node` condition always points to the CJS output (and this is [_non-configurable_](https://github.com/nodejs/node/blob/0c460518e8c0728294cfbb5313d347a62856a2c5/lib/internal/modules/esm/utils.js#L49))

devversion:
I'm closing this as I don't know what the next steps are, but I'm happy to re-open!

PR #7156, opened 2023-01-19
feat: removed deprecated call pattern with `resultSelector`

demensky:
**Description:** see individual commits.

**BREAKING CHANGE:** All deprecated call pattern with `resultSelector` is no longer available. [Read more](https://rxjs.dev/deprecations/resultSelector).


benlesh (approved):
Approved. I had to fix a couple of things. 

The biggest callout @demensky is that I changed `switchScan` to use a direct subscription, as `tap` is a little heavy-handed. It might be better if we at some point just made a `switchInternals` helper for the library like we did with `mergeMap` et al... or even just made `switchMap` based on `switchScan`... but I'll have to think about that. This is fine for now.

benlesh:
Great work, @demensky, thank you for saving me the time here.

PR #7158, opened 2023-01-19
feat(map): removed deprecated `map(project, thisArg)` call pattern

demensky:
BREAKING CHANGE: `map(project, thisArg)` call pattern is no longer available. Use Function#bind method: `map(project.bind(thisArg))`


benlesh (dismissed):

benlesh (approved):

benlesh:
Awesome. Thanks, @demensky

PR #7159, opened 2023-01-19
feat: removed deprecated call pattern in `find` & `findIndex`

demensky:
**BREAKING CHANGE:** `find(predicate, thisArg)` call pattern is no longer available. Use `find(predicate.bind(thisArg))`
**BREAKING CHANGE:** `findIndex(predicate, thisArg)` call pattern is no longer available. Use `findIndex(predicate.bind(thisArg))`


benlesh (approved):

benlesh:
Thank you again, @demensky!

PR #7163, opened 2023-01-21
feat(endWith): removed deprecated `endWith(value, scheduler)` call pattern

demensky:
**BREAKING CHANGE:** `endWith(value, scheduler)` call pattern is no longer available. [Read more](https://rxjs.dev/deprecations/scheduler-argument).


benlesh (approved):

benlesh:
Thanks again, @demensky

PR #7164, opened 2023-01-21
Removed deprecated symbols that have a new name

demensky:
Exported symbols marked as "_Renamed to $SYMBOL_NAME. Will be removed in v8_" have been removed. The entire project was also searched and all references to deleted entities were updated.

**BREAKING CHANGE:** removed deprecated symbols that have a new name


demensky:
I can’t figure out why “_CI / Node 14 build_” is crashing. I installed the same version of node locally but the tests pass.

@benlesh, maybe this is some kind of anomaly?

example

benlesh:
Not sure... It might have something to do with the transpiled output using a feature that doesn't exist in Node 14, and being somehow different than what is happening locally. Different TS version locally maybe?

demensky:
@benlesh I restarted and it's gone. It looks like it's some kind of floating bug in old node or tests in the repository.

benlesh (approved):

benlesh:
Thank you, @demensky

PR #7165, opened 2023-01-23
perf(operator): simplify the implementation of the `endWith` operator

demensky:
I would like to propose a simplification and optimization of the `endWith` operator.

Some operators may spend significantly more CPU and RAM than you expect from them. When trying to use the debugging tools, you can notice that often the problem lies in the fact that operators can consist of other operators. For example, in dealing with [`count`](https://github.com/ReactiveX/rxjs/blob/71a26ac155cb0f48c8054843282e7bbb5ed4e6fe/src/internal/operators/count.ts#L59-L61) you can see that the `reduce` operator is used in the implementation, but the code is quite simple and can hardly affect performance in any way. But if you take the operator from this PR ([`endWith`](https://github.com/ReactiveX/rxjs/blob/71a26ac155cb0f48c8054843282e7bbb5ed4e6fe/src/internal/operators/endWith.ts#L54-L58)), you can see that the relatively "heavy" operator `concatAll` is used in combination with the `fromArrayLike` functions.

I suggest trying to avoid using other operators in the implementation of operators if their implementation has a large amount of code that under no circumstances will be required (dead code) for the functionality of operators.

As an example, the difference in the number of created objects in this code:

```ts
new Observable<string>((subscriber) => {
  subscriber.next('a');
  subscriber.next('b');
  subscriber.next('c');
  subscriber.complete();
})
  .pipe(endWith('e', 'f', 'g'))
  .subscribe({
    next: (next) => {
      console.log('next', next);
    },
    complete: () => {
      console.log('complete');
    },
    error: (error) => {
      console.log('error', error);
    },
  });
```

|          | `Observable` | `Subscriber` | `Subscription` |
| -------- | -----------: | -----------: | -------------: |
| `master` |            5 |            4 |              5 |
| this PR  |            2 |            2 |              3 |

I would be very happy to hear reasoned criticism. 😄

### DIFF
--- a/src/internal/observable/innerFrom.ts
+++ b/src/internal/observable/innerFrom.ts
@@ -1,3 +1,4 @@
+import { subscribeToArray } from '../util/subscribeToArray';
 import { isArrayLike } from '../util/isArrayLike';
 import { isPromise } from '../util/isPromise';
 import { Observable } from '../Observable';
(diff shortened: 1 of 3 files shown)


benlesh:
LOL! I had forgotten that file was even in there. It wasn't even used. :)

PR #7166, opened 2023-01-24
feat(startWith): remove deprecated scheduler signature.

benlesh:
+ Removes deprecated scheduler signature
+ Improves performance slightly
+ Moves `subscribeToArray` closer to usage.

### DIFF
--- a/spec/operators/startWith-spec.ts
+++ b/spec/operators/startWith-spec.ts
@@ -1,7 +1,6 @@
 import { expect } from 'chai';
-import { startWith, mergeMap, take } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
-import { of, Observable } from 'rxjs';
+import { of, Observable, startWith, mergeMap, take } from 'rxjs';
 import { observableMatcher } from '../helpers/observableMatcher';
 
 /** @test {startWith} */
@@ -143,21 +142,6 @@ describe('startWith', () => {
     });
   });
 
-  it('should allow unsubscribing explicitly and early', () => {
-    testScheduler.run(({ hot, expectObservable, expectSubscriptions }) => {
-      const e1 = hot('  ---a--b----c--d--|');
-      const unsub = '   ---------!        ';
-      const e1subs = '  ^--------!        ';
-      const expected = 's--a--b---        ';
-      const values = { s: 's', a: 'a', b: 'b' };
-
-      const result = e1.pipe(startWith('s', testScheduler));
-
-      expectObservable(result, unsub).toBe(expected, values);
-      expectSubscriptions(e1.subscriptions).toBe(e1subs);
-    });
-  });
-
   it('should not break unsubscription chains when result is unsubscribed explicitly', () => {
     testScheduler.run(({ hot, expectObservable, expectSubscriptions }) => {
       const e1 = hot('  ---a--b----c--d--|');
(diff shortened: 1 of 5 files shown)


src/internal/operators/startWith.ts
```diff
@@ -1,7 +1,7 @@
-import { concat } from '../observable/concat';
-import { OperatorFunction, SchedulerLike, ValueFromArray } from '../types';
-import { popScheduler } from '../util/args';
+import { subscribeToArray } from '../observable/innerFrom';
+import { OperatorFunction, ValueFromArray } from '../types';
 import { operate } from '../util/lift';
+import { createOperatorSubscriber } from './OperatorSubscriber';
 
 // Devs are more likely to pass null or undefined than they are a scheduler
 // without accompanying values. To make things easier for (naughty) devs who
```
demensky, line 7: After the scheduler was removed, these signatures became useless. Maybe we should remove them?

demensky: Are signatures with `null` and `undefined` still relevant?

src/internal/observable/innerFrom.ts
```diff
@@ -64,12 +63,34 @@ export function fromInteropObservable<T>(obj: any) {
...
+  //    something like a `take`, `takeWhile`, or other synchronous unsubscription
+  //    has already unsubscribed.
+  // 2. In this form, reentrant code can alter that array we're looping over.
+  //    This is a known issue, but considered an edge case. The alternative would
+  //    be to copy the array before executing the loop, but this has
+  //    performance implications.
+  const length = array.length;
+  for (let i = 0; i < length && !subscriber.closed; i++) {
+    subscriber.next(array[i]);
+  }
```
demensky, line 90: [Here](https://github.com/ReactiveX/rxjs/blob/f93e5c86a37af3b553a950b73eb346ee2806eddb/src/internal/observable/innerFrom.ts#L93-L95) we exit the function as soon as `subscriber.closed` becomes `true`. Does it make sense to do something similar here?

jakovljevic-mladen: I think this is covered in the `for` loop. This part: `&& !subscriber.closed` should ensure that.

src/internal/operators/startWith.ts
jakovljevic-mladen: I think they aren't relevant anymore.

src/internal/observable/innerFrom.ts
demensky: It's just that in [example](https://github.com/ReactiveX/rxjs/blob/f93e5c86a37af3b553a950b73eb346ee2806eddb/src/internal/observable/innerFrom.ts#L93-L95), when `subscriber.closed` becomes `true`, then `subscriber.complete()` is not called, but in `subscribeToArray` it is called.

jakovljevic-mladen: If `subscriber.complete()` gets called, but subscription is already closed, it won't make any difference. Subscriber will not get `complete` notification multiple times, so this is OK I think.

demensky: Yes, I know that `complete` won't be notified several times. I just wanted to ask and suggest doing something like this:

```suggestion
  for (let i = 0; i < length; i++) {
    if (subscriber.closed) {
      return;
    }
    subscriber.next(array[i]);
  }
```

Nevermind, it's just my curiosity.

demensky (approved):

spec/operators/startWith-spec.ts
```diff
@@ -143,21 +142,6 @@ describe('startWith', () => {
...
 
-  it('should allow unsubscribing explicitly and early', () => {
-    testScheduler.run(({ hot, expectObservable, expectSubscriptions }) => {
-      const e1 = hot('  ---a--b----c--d--|');
-      const unsub = '   ---------!        ';
-      const e1subs = '  ^--------!        ';
-      const expected = 's--a--b---        ';
-      const values = { s: 's', a: 'a', b: 'b' };
-
-      const result = e1.pipe(startWith('s', testScheduler));
```
demensky, line 154: Just a reinsurance. Are you sure that these tests should be deleted? Isn't it enough just to remove `testScheduler`?

src/internal/observable/innerFrom.ts
demensky: @jakovljevic-mladen
I think it still needs to be fixed because `onStoppedNotification` will be called [here](https://github.com/ReactiveX/rxjs/blob/6c9930253a165303935ac096eeda65778dd8a3bc/src/internal/Subscriber.ts#L88).

```ts
import { config, Observable } from 'rxjs';

config.onStoppedNotification = (notification) => {
  console.log('onStoppedNotification', notification);
};

new Observable<never>((subscriber) => {
  subscriber.complete();
  subscriber.complete(); // log: onStoppedNotification {kind: "C", value: undefined, error: undefined}
}).subscribe();
```

demensky:
```suggestion
  for (let i = 0; i < length; i++) {
    if (subscriber.closed) {
      return;
    }
    subscriber.next(array[i]);
  }
```

I double-checked and made sure that if you do not return inside the loop, then `onStoppedNotification` will be called.

```ts
config.onStoppedNotification = (notification) => {
  console.log('onStoppedNotification', notification);
};

EMPTY.pipe(startWith(1, 2, 3), take(2)).subscribe();
```

demensky (changes requested):

src/internal/observable/innerFrom.ts
benlesh: It is _mildly_ more efficient, I suppose. A conditional followed by a return, rather than a conditional followed by a function call, a conditional and a return. It's not likely to make any performance difference though, I'd guess. Even at high-speed.

demensky: We are talking about a false positive `onStoppedNotification` call.

demensky: The difference is that `return` prevents [`subscriber.complete()`](https://github.com/ReactiveX/rxjs/blob/95f8d958eb500be6d66263b06ed4263ad6808c19/src/internal/observable/innerFrom.ts#L91).

spec/operators/startWith-spec.ts
benlesh: Yeah, they're invalid tests now. This would test that it would start with `'s'` and then an instance of `testScheduler`. :)

PR #7167, opened 2023-01-24
remove deprecations: `from` with scheduler, `combineLatest` with scheduler

benlesh:
Two of the more gnarly ones.

- `innerFrom` becomes `from`.
- `combineLatest` gets a LOT of crap removed from it. (there's more to go there, but that can be done in a follow up)


src/internal/observable/from.ts
```diff
@@ -69,36 +75,117 @@ export function from<O extends ObservableInput<any>>(input: O, scheduler: Schedu
...
+  throw createInvalidObservableTypeError(input);
+}
+
+/**
+ * Creates an RxJS Observable from an object that implements `Symbol.observable`.
+ * @param obj An object that properly implements `Symbol.observable`.
  */
-export function from<T>(input: ObservableInput<T>, scheduler?: SchedulerLike): Observable<T> {
-  return scheduler ? scheduled(input, scheduler) : innerFrom(input);
+export function fromInteropObservable<T>(obj: any) {
```
jakovljevic-mladen, line 119: Do these functions need to be exported? They don't seem to be used anywhere else.

benlesh: Good call. Done.

benlesh:
Reviewers: You'll want to review this commit by commit to make it easier on yourselves.

src/internal/observable/from.ts
```diff
@@ -65,36 +75,117 @@ export function from<O extends ObservableInput<any>>(input: O): Observable<Obser
...
- * ```
- *
  * @see {@link fromEvent}
  * @see {@link fromEventPattern}
+ * @see {@link scheduled}
  *
- * @param {ObservableInput<T>} A subscription object, a Promise, an Observable-like,
- * an Array, an iterable, or an array-like object to be converted.
- * @param {SchedulerLike} An optional {@link SchedulerLike} on which to schedule the emission of values.
- * @return {Observable<T>}
```
jakovljevic-mladen, line 100: I'd keep this `@return`. Something like this:

```
@return An Observable converted from the `input`.
```

src/internal/observable/from.ts
```diff
@@ -65,36 +75,117 @@ export function from<O extends ObservableInput<any>>(input: O): Observable<Obser
...
+ * @param input A subscription object, a Promise, an Observable-like,
+ * an Array, an iterable, async iterable, or an array-like object to be converted.
```
jakovljevic-mladen, line 83: `async iterable` may need an article: `an async iterable`.

src/internal/observable/from.ts
```diff
@@ -1,8 +1,18 @@
...
+import { isIterable } from '../util/isIterable';
+import { isReadableStreamLike, readableStreamLikeToAsyncGenerator } from '../util/isReadableStreamLike';
+import { Subscriber } from '../Subscriber';
+import { isFunction } from '../util/isFunction';
+import { reportUnhandledError } from '../util/reportUnhandledError';
+import { observable as Symbol_observable } from '../symbol/observable';
 
-export function from<O extends ObservableInput<any>>(input: O): Observable<ObservedValueOf<O>>;
 /**
  * Creates an Observable from an Array, an array-like object, a Promise, an iterable object, or an Observable-like object.
```
jakovljevic-mladen, line 17: I think we should cover all [ObservableInput](https://rxjs.dev/api/index/type-alias/ObservableInput) here. Also, `an Array` and `an array-like` are the same I guess, so I think that these can be merged to one.

If this is not relevant for this PR, we can fix it in another one. If you decide to include such a fix here, please update `@param input` on the bottom of the JSDoc as well.

benlesh:
@jakovljevic-mladen maybe we can have a follow-up to revamp the `from` docs? Right now I'd just like to get the removal landed, as this one effects a lot of stuff.

jakovljevic-mladen:
@benlesh, I agree. I had a plan to write much more `ObservableInput` docs, we have some open issues regarding that. I could cover both `from` and `ObservableInput` in a single PR.

jakovljevic-mladen (approved):
LGTM, only minor docs issues that _could_ be considered to be fixed right away (an article and `@return`).

src/internal/observable/of.ts
```diff
@@ -79,5 +80,5 @@ export function of<A extends readonly unknown[]>(...values: A): Observable<Value
  */
 export function of<T>(...args: Array<T | SchedulerLike>): Observable<T> {
   const scheduler = popScheduler(args);
-  return from(args as T[], scheduler);
+  return scheduler ? scheduled(args as T[], scheduler) : from(args as T[]);
```
demensky, line 83: Maybe `fromArrayLike` would be better here?

```suggestion
  return scheduler ? scheduled(args as T[], scheduler) : fromArrayLike(args as T[]);
```

src/internal/observable/of.ts
```diff
@@ -79,5 +80,5 @@ export function of<A extends readonly unknown[]>(...values: A): Observable<Value
  */
 export function of<T>(...args: Array<T | SchedulerLike>): Observable<T> {
   const scheduler = popScheduler(args);
-  return from(args as T[], scheduler);
+  return scheduler ? scheduled(args as T[], scheduler) : fromArrayLike(args as T[]);
```
jeremymwells, line 83: `fromArrayLike` is not exported from `from`. Should this just be `from`?

demensky: There can be nothing here but an array. `from` [contains a lot of code](https://github.com/ReactiveX/rxjs/blob/314b204acc9a8923830789c948727febb5b521e1/src/internal/observable/from.ts#L92-L109) that converts various types into `Observable` which will never be executed in the context of `of`.

jeremymwells (approved):

PR #7168, opened 2023-01-24
feat(combineLatest): remove spread params deprecated signature.

benlesh:
+ Removes the `combineLatest(a$, b$, c$)` signature, so we can standardize on `combineLatest([a$, b$, c$])` and `combineLatest({ a: a$, b: b$ })`.

NOTE: It appears we forgot to deprecate the `resultSelector` signature. Which isn't even supported for the newer object argument syntax (and never was).

### DIFF
--- a/src/internal/observable/combineLatest.ts
+++ b/src/internal/observable/combineLatest.ts
@@ -1,15 +1,13 @@
 import { Observable } from '../Observable';
 import { ObservableInput, ObservedValueOf, ObservableInputTuple } from '../types';
-import { argsArgArrayOrObject } from '../util/argsArgArrayOrObject';
+import { arrayOrObject } from '../util/argsArgArrayOrObject';
 import { Subscriber } from '../Subscriber';
-import { from } from './from';
 import { identity } from '../util/identity';
-import { mapOneOrManyArgs } from '../util/mapOneOrManyArgs';
-import { popResultSelector } from '../util/args';
 import { createObject } from '../util/createObject';
 import { createOperatorSubscriber } from '../operators/OperatorSubscriber';
 import { AnyCatcher } from '../AnyCatcher';
 import { EMPTY } from './empty';
+import { from } from './from';
 
 // combineLatest(any)
 // We put this first because we need to catch cases where the user has supplied
@@ -30,14 +28,6 @@ export function combineLatest<A extends readonly unknown[]>(sources: readonly [.
 export function combineLatest<A extends readonly unknown[], R>(
   sources: readonly [...ObservableInputTuple<A>],
   resultSelector: (...values: A) => R
-): Observable<R>;
-
-// combineLatest(a, b, c)
-/** @deprecated Pass an array of sources instead. The rest-parameters signature will be removed in v8. Details: https://rxjs.dev/deprecations/array-argument */
-export function combineLatest<A extends readonly unknown[]>(...sources: [...ObservableInputTuple<A>]): Observable<A>;
-/** @deprecated Pass an array of sources instead. The rest-parameters signature will be removed in v8. Details: https://rxjs.dev/deprecations/array-argument */
-export function combineLatest<A extends readonly unknown[], R>(
-  ...sourcesAndResultSelector: [...ObservableInputTuple<A>, (...values: A) => R]
 ): Observable<R>;
 
 // combineLatest({a, b, c})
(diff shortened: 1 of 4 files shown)


benlesh:
@demensky would you mind reviewing this?

demensky:
@benlesh, I'll try to do a review as soon as possible.

src/internal/observable/combineLatest.ts
```diff
@@ -188,117 +158,88 @@ export function combineLatest<T extends Record<string, ObservableInput<any>>>(
...
+  resultSelector?: (...value: any[]) => R
+): Observable<R> | Observable<ObservedValueOf<O>[]> {
+  const parts = arrayOrObject(sources);
 
-  const { args: observables, keys } = argsArgArrayOrObject(args);
+  if (!parts) {
+    throw new TypeError('sources must be an array or object');
+  }
+
+  const { args: observables, keys } = parts;
```
demensky, line 179: I propose to separate `arrayOrObject` and `argsArgArrayOrObject`, make a check "sources must be an array or object" inside `arrayOrObject`. This will make it easier for us if/when we want to make similar changes to `zip` and `forkJoin`.

demensky (approved):

src/internal/util/argsArgArrayOrObject.ts
```diff
@@ -10,21 +10,29 @@ const { getPrototypeOf, prototype: objectProto, keys: getKeys } = Object;
...
+    const result = arrayOrObject(first);
+    if (result) {
+      return result;
     }
   }
 
   return { args: args as T[], keys: null };
 }
 
+export function arrayOrObject<T, O extends Record<string, T>>(first: T | T[] | O) {
```
jeremymwells, line 22: @benlesh- do you want to split this out (per @demensky)?

src/internal/observable/combineLatest.ts
```diff
@@ -188,117 +158,88 @@ export function combineLatest<T extends Record<string, ObservableInput<any>>>(
...
       observables as ObservableInput<ObservedValueOf<O>>[],
-      scheduler,
       keys
-        ? // A handler for scrubbing the array of args into a dictionary.
-          (values) => createObject(keys, values)
-        : // A passthrough to just return the array
-          identity
+        ? (values: any[]) => createObject(keys, values)
+        : resultSelector
+        ? (values: readonly any[]) => resultSelector(...values)
```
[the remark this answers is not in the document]
benlesh: This is what Prettier did. 🤷

src/internal/util/argsArgArrayOrObject.ts
benlesh: Nah. It's fine in the same file for now. Every file we add actually adds a little overhead when bundling.

src/internal/observable/combineLatest.ts
benlesh: Let's do this in a follow up. I need to think about the best approach a little bit. Incremental refactors.

Every change to thrown errors is a change to our external API, and therefore a breaking change.

for example, if someone is doing this, and we change the error message to be more specific, we'll break them:

```ts
try {
  combineLatest(someVariable)
} catch (err) {
  if (err.message === 'sources must be an array or object') {
    // do something.
  }
}
```

I'm not saying we shouldn't make the change, just that we should make sure we're moving deliberatly.

src/internal/observable/combineLatest.ts
```diff
@@ -168,18 +158,25 @@ export function combineLatest<T extends Record<string, ObservableInput<any>>>(
  * @see {@link merge}
  * @see {@link withLatestFrom}
  *
- * @param {ObservableInput} [observables] An array of input Observables to combine with each other.
+ * @param sources An array of input Observables to combine with each other.
  * An array of Observables must be given as the first argument.
- * @param {function} [project] An optional function to project the values from
+ * @param project An optional function to project the values from
```
jakovljevic-mladen, line 163: Name of this parameter is not `project`, please rename to [`resultSelector`](https://github.com/ReactiveX/rxjs/pull/7168/files#diff-93ab8d5b23bdc9e3322b558953476f6efe920958972b445cdebfd6b666c678a7R171).

jakovljevic-mladen (approved):
LGTM, only a minor docs issue (related to parameter rename).

src/internal/observable/combineLatest.ts
ktrz: +1 on that

ktrz (commented):
Just +1 on Mladen's comment. Otherwise, LGTM

PR #7171, opened 2023-01-25
refactor(type): discard redundant `ValueFromArray`

demensky:
TypeScript has an easy way to get the type of array's value. This method works even in [older versions](https://www.typescriptlang.org/play?ts=3.3.3#code/C4TwDgpgBAysBOBLAdgcwIL3gQxFAvFAM4IqoDaAugNwCwAUAwCYQDGANtvNAGYCuyVsEQB7ZFB4iRACmwAuWKTSYcIcsj4BbAEYR4lAJQKAbiMRM6jepJkBybV1sHLQA). As you can see from the `endWith` example, this method calculates the type more reliably, this allowed us to remove the `as` hack.


benlesh (changes requested):
Technically this is a breaking change. Anyone that was using this exported type would be effected. We need to at least mark it as such.

demensky:
Reverted back `ValueFromArray` and marked it as `@deprecated`.

LcsGa (approved):

PR #7176, opened 2023-01-31
fix(throttle): properly handle default ThrottleConfig values

### LINKED ISSUE #7146: Unexpected `false` default value for `leading` when only `trailing` is specified in config parameter
### Describe the bug

The config object to `throttle` and `throttleTime` behaves in an unexpected way regarding `leading`. 

The documentation states that the config parameter "`Defaults to { leading: true, trailing: false }`" but if a config parameter is specified the the default value for `leading` becomes false.

In other words, if someone wants to change the behaviour for `trailing` to `true` and therefore adds a config parameter `{ trailing: true }` this also changes the behavior of `leading` to be `false`. 

I was first tricked by this but then realised I have to specify both `leading` and `trailing` even if I just want to change the behavior of `trailing`.  But then I got PR review comments on the updated code from two independent people that I did not have to specify `leading: true` since that is the default value. 

So it seems like the current behaviour is quite unexpected since they also thought that default value for `leading` would not be affected by specifying `trailing`.

### Expected behavior

I would expect that `leading` is set to the default value (`true`) but this code would cause `leading: false`:
```ts
  throttle(1000, undefined, { trailing: true }
```

I expect the reproduction code to print the same as if I do not specify `trailing`, ie:
```
0
6
13
20
26
```

but it prints 

```
6
13
20
26
```

### Reproduction code

```typescript
timer(0, 150)
  .pipe(throttleTime(1000, undefined, { trailing: true }))
  .subscribe((v) => console.log(v));
```

### Reproduction URL

https://stackblitz.com/edit/rxjs-wehnfb?file=index.ts

### Version

7.5.5

### Environment

_No response_

### Additional context

I believe the last line in this code defines the current behavior:

```ts
* @param durationSelector A function
 * that receives a value from the source Observable, for computing the silencing
 * duration for each source value, returned as an Observable or a Promise.
 * @param config a configuration object to define `leading` and `trailing` behavior. Defaults
 * to `{ leading: true, trailing: false }`.
 * @return A function that returns an Observable that performs the throttle
 * operation to limit the rate of emissions from the source.
 */
export function throttle<T>(
  durationSelector: (value: T) => ObservableInput<any>,
  config: ThrottleConfig = defaultThrottleConfig
    const { leading, trailing } = config;
```

jakovljevic-mladen:
**Description:**
This PR is split into two commits. The first one fixes an issue described in #7146, while the second one adds `ThrottleConfig` docs.

This issue should be probably cherry picked to the `7.x` branch.

Fixes #7146

### DIFF
--- a/src/internal/operators/throttleTime.ts
+++ b/src/internal/operators/throttleTime.ts
@@ -1,5 +1,5 @@
 import { asyncScheduler } from '../scheduler/async';
-import { defaultThrottleConfig, throttle } from './throttle';
+import { throttle, ThrottleConfig } from './throttle';
 import { MonoTypeOperatorFunction, SchedulerLike } from '../types';
 import { timer } from '../observable/timer';
 
@@ -55,7 +55,7 @@ import { timer } from '../observable/timer';
 export function throttleTime<T>(
   duration: number,
   scheduler: SchedulerLike = asyncScheduler,
-  config = defaultThrottleConfig
+  config?: ThrottleConfig
 ): MonoTypeOperatorFunction<T> {
   const duration$ = timer(duration, scheduler);
   return throttle(() => duration$, config);
--- a/src/internal/operators/throttleTime.ts
+++ b/src/internal/operators/throttleTime.ts
@@ -47,7 +47,7 @@ import { timer } from '../observable/timer';
  * internally by the optional `scheduler`.
  * @param scheduler The {@link SchedulerLike} to use for
  * managing the timers that handle the throttling. Defaults to {@link asyncScheduler}.
- * @param config a configuration object to define `leading` and
+ * @param config A configuration object to define `leading` and
  * `trailing` behavior. Defaults to `{ leading: true, trailing: false }`.
  * @return A function that returns an Observable that performs the throttle
  * operation to limit the rate of emissions from the source.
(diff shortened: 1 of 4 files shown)


src/internal/operators/throttleTime.ts
```diff
@@ -47,15 +47,15 @@ import { timer } from '../observable/timer';
...
  * managing the timers that handle the throttling. Defaults to {@link asyncScheduler}.
- * @param config a configuration object to define `leading` and
+ * @param config A configuration object to define `leading` and
  * `trailing` behavior. Defaults to `{ leading: true, trailing: false }`.
  * @return A function that returns an Observable that performs the throttle
  * operation to limit the rate of emissions from the source.
  */
 export function throttleTime<T>(
   duration: number,
   scheduler: SchedulerLike = asyncScheduler,
```
jakovljevic-mladen, line 57: Should `throttleTime` merge `scheduler` and `config` parameters effectively introducing a `ThrottleTimeConfig` interface which would extend `ThrottleConfig` interface by adding `scheduler?: SchedulerLike` property to it?

It would effectively remove the need to use `throttleTime` like this: `throttleTime(1000, undefined, { trailing: true })` to `throttleTime(1000, { trailing: true })`.

@benlesh, thoughts?

benlesh: Yes, this should probably be `throttleTime(number, config)` where `config` has a `scheduler` option. We'll need to deprecate the other signatures that take a scheduler as the second argument, but allow the scheduler to be passed with the config.

benlesh (approved):

PR #7177, opened 2023-01-31
Reduction in size of core piece `Subscriber` and `Subscription`, update deps, internalize deprecated props, drop Node 14.

benlesh:
feat: Reduction in size of core piece `Subscriber` and `Subscription`, internalize deprecated props

+ Removes specialized parent subscription tracking in favor of just adding teardowns to children to remove themselves from parents. This is what you'd have to do with any other similar contemporary api. A note has been added to the docs warning not to leave long-lived child subscriptions that have been removed from parents. This is a corner case we shouldn't be shipping extra code for.
+ Refactors and adds more descriptive comments to tests that are checking the counts of registered finalizers to more definitively show what the test is testing and help understanding of the tests.
+ Eliminates unnecessary creation and in-package-only export/import of `EMPTY_SUBSCRIPTION`.
+ Streamlines subscriber check on `subscription`.  The majority of the time we're either getting a `Subscriber` that we own and trust, or we're getting a user-supplied function or observer. Previous shape-checks could cause issues where an object that was "shaped" properly would not have appropriate error handling safety applied. We can only really trust our own `Subscriber` instances. This does mean that in the case of our own `Subjects` they will be wrapped in a `Subscriber` instance. This can be addressed in a follow-up.
+ Adds additional tests for to ensure errors are reported when thrown inside of destinations handed to the `Subscriber` constructor.

**BREAKING CHANGE: Deprecated `subscriber.isStopped` has been made internal. Use `subscriber.closed` or create a custom observer that tracks `isStopped` state.**

**BREAKING CHANGE: Deprecated `subscriber.destination` has been made internal. If you need to access the subscriber's destination for some reason, you will have to trap the value ahead of time. If this is something you really need, please file and issue and bring it to our attention. There shouldn't be any use cases for this.**

---

chore: update ts-node, mocha, chai, and sinon-chai
    
We had to update ts-node for the newer version of TypeScript. This PR showed an issue where when we used `??=` in the code, CI would fail for Node 14... but it would fail with a nonsense error. Trial and elimination showed that the error was caused when the syntax error was being handled by mocha in ts-node in Node 14. It caused an error like `Unknown pluginType "ui"`. Updating all of these dependencies resolved the issue so I could see the actual error.
    
 ---
 
 feat: Dropping support for Node 14.
    
 + Remove Node 14 from CI
    
**BREAKING CHANGE: Node 14 is no longer supported in version 8. End of life for Node 14 is April 30, 2023. There's no reason for RxJS version 8 to support it, as we won't even be out of beta at that point.**


spec/Subscriber-spec.ts
```diff
@@ -179,6 +177,114 @@ describe('Subscriber', () => {
...
+        },
+        complete() {
+          throw new Error('thrown from complete');
+        }
+      });
+
+      subscriber.next()
+    });
+
+    it('should report errors thrown from a full observer even if it is also shaped like a subscription', (done) => {
```
benlesh, line 255: This test will fail in master. The others will pass, I was just being thorough.

src/internal/Subscription.ts
```diff
@@ -126,51 +110,17 @@ export class Subscription implements SubscriptionLike {
...
-          // We don't add closed subscriptions, and we don't add the same subscription
-          // twice. Subscription unsubscribe is idempotent.
-          if (teardown.closed || teardown._hasParent(this)) {
-            return;
-          }
-          teardown._addParent(this);
+        if (teardown && 'add' in teardown) {
+          // If teardown is a subscription, we can make sure that if it
+          // unsubscribes first, it removes itself from this subscription.
+          teardown.add(() => {
```
benlesh, line 116: No longer keeping track of "parents" internally, instead we're optimizing for a world where child subscriptions remove themselves if they unsubscribe. The side effect to this is if you add and remove a child subscription from a parent subscription a number of times (an extreme edge case), or if you remove a child subscription from a parent, then leak the child subscription for some reason, it will keep a reference to the parent until you unsubscribe it or garbage collect it. These are corner cases, and we're able to remove a lot of weight if we just instruct people to clean up after themselves. :)

src/internal/Subscription.ts
```diff
@@ -186,26 +136,18 @@ export class Subscription implements SubscriptionLike {
...
   remove(teardown: Exclude<TeardownLogic, void>): void {
     this._finalizers?.delete(teardown);
-
-    if (teardown instanceof Subscription) {
-      teardown._removeParent(this);
-    }
   }
 }
 
-export const EMPTY_SUBSCRIPTION = Subscription.EMPTY;
```
benlesh, line 200: This wasn't externally exposed, and overall didn't need to exist.

src/internal/Subscription.ts
```diff
@@ -186,26 +136,18 @@ export class Subscription implements SubscriptionLike {
...
-
-export function isSubscription(value: any): value is Subscription {
```
benlesh, line 202: This was only used in the Observable `subscribe` call, and honestly was more robust than was necessary.

src/internal/Subscriber.ts
```diff
@@ -184,22 +174,6 @@ function handleStoppedNotification(notification: ObservableNotification<any>, su
   onStoppedNotification && timeoutProvider.setTimeout(() => onStoppedNotification(notification, subscriber));
 }
 
-/**
- * The observer used as a stub for subscriptions where the user did not
- * pass any arguments to `subscribe`. Comes with the default error handling
- * behavior.
- */
-export const EMPTY_OBSERVER: Readonly<Observer<any>> & { closed: true } = {
```
benlesh, line 192: Not used at all. Removed.

src/internal/Subscriber.ts
```diff
@@ -164,16 +164,6 @@ function createSafeObserver<T>(observerOrNext?: Partial<Observer<T>> | ((value:
   return new ConsumerObserver(!observerOrNext || isFunction(observerOrNext) ? { next: observerOrNext ?? undefined } : observerOrNext);
 }
 
-/**
- * An error handler used when no error handler was supplied
- * to the SafeSubscriber -- meaning no error handler was supplied
- * do the `subscribe` call on our observable.
- * @param err The error to handle
- */
-function defaultErrorHandler(err: any) {
```
benlesh, line 173: Only used in the removed `EMPTY_OBSERVER`.

src/internal/Subscriber.ts
```diff
@@ -184,22 +174,6 @@ function handleStoppedNotification(notification: ObservableNotification<any>, su
...
- * behavior.
- */
-export const EMPTY_OBSERVER: Readonly<Observer<any>> & { closed: true } = {
-  closed: true,
-  next: noop,
-  error: defaultErrorHandler,
-  complete: noop,
-};
-
-function isObserver<T>(value: any): value is Observer<T> {
```
benlesh, line 199: Only used in `isSubscriber`, which is being removed.

src/internal/Subscriber.ts
```diff
@@ -184,22 +174,6 @@ function handleStoppedNotification(notification: ObservableNotification<any>, su
...
-  return value && isFunction(value.next) && isFunction(value.error) && isFunction(value.complete);
-}
-
-export function isSubscriber<T>(value: any): value is Subscriber<T> {
```
benlesh, line 203: This isn't useful. The only way we can guarantee safety is applied during subscription is to check that it's OUR instance of `Subscriber`. Removed in favor of a simple `instanceof` check.

src/internal/Subscription.ts
```diff
@@ -28,8 +27,6 @@ export class Subscription implements SubscriptionLike {
    */
   public closed = false;
 
-  private _parentage: Subscription[] | Subscription | null = null;
```
benlesh, line 31: Almost all of the code removed in this file (which is substantial) comes from no longer tracking parents on children, and instead just having the children remove themselves. Side effects of that are explained in other comments.

src/internal/Subscription.ts
```diff
@@ -85,7 +69,7 @@ export class Subscription implements SubscriptionLike {
           } catch (err) {
             errors = errors ?? [];
             if (err instanceof UnsubscriptionError) {
-              errors = [...errors, ...err.errors];
+              errors.push(...err.errors);
```
benlesh, line 72: Previously, we were unnecessarily creating a new array every time we ran into an `UnsubscriptionError` inside of our unsubscription. This is probably a corner case, but it's silly that we were doing that.

src/internal/Subscriber.ts
trxcllnt: Are we worried about what happens if someone has two versions of Rx in the same app? i.e. some random package returns an `rxjs@v6` subscription that is added to an `rxjs@v8` subscriber. Or is that handled by always wrapping the unknown observer via `createSafeObserver()`?

jeremymwells: I think removing `isSubscriber()` makes sense. `createSafeObserver()` should yield normalized observer behavior(s).

jeremymwells (approved):

src/internal/Subscriber.ts
benlesh: @trxcllnt it will wrap it in the safe observer. There's a possibility that in a mixed environment mixing one RxJS's copy of Observable, and another's copy of some operator might result in subscriptions not being chained properly, but I haven't confirmed that. In any case, that's a corner case I wouldn't want to punish everyone else for in terms of size. I'll research that to see if it's even possible. If there's a small fix that doesn't cost everyone else, then we'll fix it. If not, we'll have to document workarounds, etc.

src/internal/Subscription.ts
```diff
@@ -186,26 +136,18 @@ export class Subscription implements SubscriptionLike {
    *
    * All finalizer instances are removed to free up memory upon unsubscription.
    *
+   * TIP: In instances you're adding and removing _Subscriptions from other Subscriptions_, you should
```
jakovljevic-mladen, line 139: This tip is a bit hard for me to understand (most probably because I'm not native English speaker). Is there a way to rephrase it a bit?

Also, I think it has a grammatical error here: `...this a non-issue,...`.

.github/workflows/ci_main.yml
```diff
@@ -34,8 +34,8 @@ jobs:
       - name: dtslint
         run: npm run dtslint
       - name: import
-        if: ${{ matrix.node == '14' }}
+        if: ${{ matrix.node == '18' }}
         run: npm run test:import
       - name: import
-        if: ${{ matrix.node == '14' || matrix.node == '15' }}
+        if: ${{ matrix.node == '16' || matrix.node == '18' }}
```
jakovljevic-mladen, line 40: This `if` is not needed at all since we now only have version 16 and 18.

jakovljevic-mladen (approved):
LGTM, only minor nits/comments.

PR #7183, opened 2023-02-11
refactor: remove legacy code

demensky:
Main points:

- [x] Drop Internet Explorer support into a pit of flames! 🔥 IE 🔥 🥳
- [x] Remove hacks when inheriting from native error classes. This makes no sense if ES2015+ is used.
- [ ] Remove support for ES5.
- [ ] [Switch compilation to ES2022](https://github.com/ReactiveX/rxjs/issues/6321#issuecomment-1423258843).

If you know any other places that are legacy and need to be deleted/refactor, welcome to the comments.

**BREAKING CHANGE:** `EmptyError` cannot be used if `target` is below ES2015
**BREAKING CHANGE:** `NotFoundError` cannot be used if `target` is below ES2015
**BREAKING CHANGE:** `SequenceError` cannot be used if `target` is below ES2015
**BREAKING CHANGE:** `TimeoutError` cannot be used if `target` is below ES2015
**BREAKING CHANGE:** `ObjectUnsubscribedError` cannot be used if `target` is below ES2015
**BREAKING CHANGE:** `ArgumentOutOfRangeError` cannot be used if `target` is below ES2015
**BREAKING CHANGE:** `AjaxError` cannot be used if `target` is below ES2015
**BREAKING CHANGE:** `AjaxTimeoutError` cannot be used if `target` is below ES2015
**BREAKING CHANGE:** `UnsubscriptionError` cannot be used if `target` is below ES2015
**BREAKING CHANGE:** `ajax` no longer supports IE


brodo:
Hi, can you tell my why you closed this PR? I stumbled across the `createErrorClass` function recently and it has more problems besides being 'hacky':

- It does not create 'native' errors and thus can not be identified with Node.js's `util.types.isNativeError()` function.
- It does instantiate the stack trace eagerly. This means that error creation is about 10 times slower compared to a normal error (in my micro benchmark).
- All stack traces contain createErrorClass, which is not pretty

PR #7184, opened 2023-02-12
Drop all Internet Explorer support into a pit of flames! 🔥 IE 🔥

demensky:
**BREAKING CHANGE:** `ajax` no longer supports IE


src/internal/ajax/getXHRResponse.ts
```diff
@@ -1,37 +1,11 @@
...
-      if ('response' in xhr) {
-        return xhr.response;
-      } else {
-        // IE
-        const ieXHR: any = xhr;
-        return ieXHR.responseText;
-      }
-    }
-  }
+  return xhr.responseType === 'document' ? xhr.responseXML : xhr.response;
```
benlesh, line 10: Honestly, we should just inline this if it's only used in one spot. It's a small amount of code now.

demensky: [Judging by the specification](https://xhr.spec.whatwg.org/#the-response-attribute), the use of `responseXML` is redundant, since we will get what we need anyway.

> Otherwise, if [this](https://webidl.spec.whatwg.org/#this)’s [response type](https://xhr.spec.whatwg.org/#response-type) is "document", [set a document response](https://xhr.spec.whatwg.org/#document-response) for [this](https://webidl.spec.whatwg.org/#this).

So I replaced `getXHRResponse(xhr)` with `xhr.response`.

benlesh (approved):

benlesh:
Thanks, @demensky

PR #7188, opened 2023-02-17
refactor(fromEvent): Make fromEvent a lot smaller

benlesh:
+ `fromEvent` no longer relies on `mergeMap` or `from` to handle the case of an `ArrayLike` of event targets (for example a `NodeList`).
+ Overall code size reduction, should also be a lot faster on subscribe.
+ Added an error message for invalid input when an `ArrayLike` comes with non-event-like items in it.

### DIFF
--- a/src/internal/observable/fromEvent.ts
+++ b/src/internal/observable/fromEvent.ts
@@ -251,7 +251,6 @@ export function fromEvent<T>(
   }
 
   return new Observable<T>((subscriber) => {
-    const targets: ArrayLike<any> = !isValidTarget && isArrayLike(target) ? target : [target];
     const handler = (...args: any[]) => subscriber.next(1 < args.length ? args : args[0]);
     if (isValidTarget) {
       // Valid event targets, even if they have a `length` property
@@ -260,8 +259,8 @@ export function fromEvent<T>(
     } else {
       // If it wasn't a valid event target, it must be an array-like.
       // Subscribe to each item in the array-like.
-      for (let i = 0; i < targets.length && !subscriber.closed; i++) {
-        const subTarget = targets[i];
+      for (let i = 0; i < target.length && !subscriber.closed; i++) {
+        const subTarget = target[i];
         doSubscribe(handler, subscriber, subTarget, eventName, options);
       }
     }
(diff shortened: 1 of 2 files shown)


demensky (commented):
I suggest calculating `targets` earlier and eliminating the code for the case where `target` is a single item.

```ts
export function fromEvent<T>(
  target: any,
  eventName: string,
  options?: EventListenerOptions | ((...args: any[]) => T),
  resultSelector?: (...args: any[]) => T
): Observable<T> {
  if (isFunction(options)) {
    resultSelector = options;
    options = undefined;
  }

  if (resultSelector) {
    return fromEvent<T>(target, eventName, options as EventListenerOptions).pipe(mapOneOrManyArgs(resultSelector));
  }

  let targets: ArrayLike<any>;

  if (isNodeStyleEventEmitter(target) || isJQueryStyleEventEmitter(target) || isEventTarget(target)) {
    targets = [target];
  } else if (isArrayLike(target)) {
    targets = target;
  } else {
    throw new TypeError('Invalid event target');
  }

  return new Observable<T>((subscriber) => {
    const handler = (...args: any[]) => subscriber.next(1 < args.length ? args : args[0]);

    for (let i = 0; i < targets.length && !subscriber.closed; i++) {
      doSubscribe(handler, subscriber, targets[i], eventName, options);
    }
  });
}
```

src/internal/observable/fromEvent.ts
```diff
@@ -240,66 +239,54 @@ export function fromEvent<T>(
...
-  if (!add) {
+  if (!isValidTarget && !isArrayLike(target)) {
     throw new TypeError('Invalid event target');
   }
 
   return new Observable<T>((subscriber) => {
-    // The handler we are going to register. Forwards the event object, by itself, or
-    // an array of arguments to the event handler, if there is more than one argument,
-    // to the consumer.
+    const targets: ArrayLike<any> = !isValidTarget && isArrayLike(target) ? target : [target];
```
benlesh, line 254: This was a mistake I left in here. :\

benlesh:
@demensky that's not a bad suggestion, but given the common case is a single event target, I'd rather not make that case pay the penalty of an additional array allocation and a for loop.

There was a bit in there were I was needlessly creating a `targets` array, and I've removed that. It was cruft.

demensky (approved):

Issue #6857, opened 2022-03-03
Implement Symbol.asyncIterator on Observable


benlesh:
### Discussed in https://github.com/ReactiveX/rxjs/discussions/6779

<div type='discussions-op-text'>

<sup>Originally posted by **benlesh** January 21, 2022</sup>
Thinking about the backpressure-related use cases for interop between async iterables and observable, I think I'd consider it an improvement to ergonomics to implement `Symbol.asyncIterator` on `Observable`. [See this comment here](https://github.com/ReactiveX/rxjs/discussions/6334#discussioncomment-2018168) about handling backpressure. There are some really cool/easy/clever things that can be done with this functionality.

Here's a few things to consider:

1. [rxjs-for-await](https://www.npmjs.com/package/rxjs-for-await) has a good amount of usage already.
2. `concatMap` has _exactly the same issue_ where users need to "understand there is buffering" in some cases, and so far, I haven't seen many people trip over that. In fact, many tutorials and documents steer people towards `concatMap` for this buffered, one-at-a-time behavior more often than not.
3. `for await` isn't going away any time soon, and &mdash; other than callbacks &mdash; is the only real native way to iterate async values (one-at-a-time like `concatMap`, of course).
4. Subscribing to an observable with `for await` is obviously non-cancellable, as there's no subscription or even an opportunity to pass a signal or the like, so it's unlikely to "replace" calling `subscribe` in the hearts and minds of users.
5. Provides even better interop with IxJS and JavaScript in general.

### For those new to this, here is what is being proposed (roughly):

```ts
for await (const value of someObservable$) {
   await sleep(1000);
   const subValue = await getAnotherValue(value);
   doSomething(subValue);
}
```

Which would roughly map 1-to-1 with this RxJS behavior:

```ts
someObservable$.pipe(
  concatMap(async (value) => {
    await sleep(1000);
    const subValue = await getAnotherValue(value);
    doSomething(subValue);
  })
)
.subscribe();
```

cc @cartant @kwonoj </div>

benlesh:
Core Team: silently approved for 8.x

vitaly-t:
I think this will make code confusing, especially for those who never used an Observable as an iterable.

Consider the following:

```ts
for await (const a of $myObservable) {
    // set of operations
}
```

It would not be a typical way to use an observable, but rather unique. So when someone sees such code, he will be confused as to what is happening here underneath.

[...]

PR #7189, opened 2023-02-17
feat(Observable): now implements `Symbol.asyncIterator`

benlesh:
Adds async iterator support to `Observable`!

#6779 
#6857

### DIFF
--- a/spec/Observable-spec.ts
+++ b/spec/Observable-spec.ts
@@ -5,6 +5,7 @@ import { Observable, config, Subscription, Subscriber, Operator, NEVER, Subject,
 import { map, filter, count, tap, combineLatestWith, concatWith, mergeWith, raceWith, zipWith, catchError, share} from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from './helpers/observableMatcher';
+import { result } from 'lodash';
 
 function expectFullObserver(val: any) {
   expect(val).to.be.a('object');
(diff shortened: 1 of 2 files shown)


spec/Observable-spec.ts
```diff
@@ -5,6 +5,7 @@ import { Observable, config, Subscription, Subscriber, Operator, NEVER, Subject,
 import { map, multicast, refCount, filter, count, tap, combineLatestWith, concatWith, merge, race, zip, catchError, publish, publishLast, publishBehavior, share} from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from './helpers/observableMatcher';
+import { result } from 'lodash';
```
demensky, line 8: I'm guessing it's just a typo.

```suggestion
```

benlesh: Lmfao

src/internal/Observable.ts
```diff
@@ -392,4 +392,123 @@ export class Observable<T> implements Subscribable<T> {
...
+        });
+      },
+      throw: (err): Promise<IteratorResult<T>> => {
+        subscription?.unsubscribe();
+        // NOTE: I did some research on this, and as of Feb 2023, Chrome doesn't seem to do
+        // anything with pending promises returned from `next()` when `throw()` is called.
+        // However, for consumption of observables, I don't want RxJS taking the heat for that
+        // quirk/leak of the type. So we're going to reject all pending promises we've nexted out here.
+        handleError(err);
+        return Promise.resolve({ value: undefined, done: true });
```
jpark011, line 498: it should reject on throw?

```suggestion
        return Promise.reject(err);
```

benlesh: No. The promise you get back from an async generator `throw` call doesn't normally reject.

jpark011: I might be misunderstanding, but I think an async generator should reject according to [AsyncGenerator.prototype.throw](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-asyncgenerator-prototype-throw)? 🧐

Also if `throw` does not reject, the catch clause in this test case will be skipped! (I'm not sure if this is intended though)
https://github.com/benlesh/rxjs/blob/e7968fde17cffc3ff35cf770c3d7f794af1053a5/spec/Observable-spec.ts#L1425-L1445

Forgot to say thank you for amazing work!! This feature is what I've been waiting for so long... 💜 RxJS

benlesh: Revisiting this and testing in Chrome console:

```ts
async function* myGen() {
   while (true) {
      yield 'hi'
   }
}

const g = myGen()

g.throw(new Error('wee')).catch(err => {
  console.log(`The throw call returned a rejected promise: ${err}`)
})
```

The throw call does indeed return a rejected error in this case.

Thanks for catching this. Sorry for the confusion.

demensky:
I suggest to play it safe and do it first as a regular method, and after a while do it with `Symbol.asyncIterator`. So there will be additional time to decide whether this was the optimal solution.

For example, the current implementation will trigger a microtask on each event, even if we already have events stored in the queue.

In streams, we usually work not with individual items, but with their chunks `AsyncGenerator<T[], void, void>`. Here, too, a similar approach suggests itself, returning an array of accumulated events.

```ts
for await (const events of source$) {
   console.log(events); // event accumulated between iterations
}
```

At the same time, nothing will interfere with us to work with each element separately. At the same time, it will not be unnecessary to provoke microtasks (and parasite Promise).

```ts
for await (const events of source$) {
  for (const event of events) {
     console.log(event);
  }
}
```

I hope that I was able to convey my thought correctly. Sorry if I wrote inaudibly somewhere. English is not my bearing language.

benlesh:
@demensky The library https://github.com/benlesh/rxjs-for-await as existed for quite some time. Every single use of that library I've found uses `eachValueFrom`. There's probably an argument to be made for pulling the rest of that library in here... however, given that the behavior here is nearly identical to `concatMap`, and people don't often trip over that, the core team has decided that this is the proper route.

demensky (approved):

spec/Observable-spec.ts
```diff
@@ -972,4 +973,239 @@ describe('Observable.lift', () => {
...
+      });
+
+      const results: number[] = [];
+      for await (const value of source) {
+        results.push(value);
+        break;
+      }
+
+      expect(results).to.deep.equal([1]);
+    });
```
ktrz, line 1009: Either a test name is misleading here or the unsubscription is not tested here. Are we expecting the subscription to be unsubscribed earlier if we break out of the loop?

Based on the test name I would imagine something like this:
```ts
    it('should unsubscribe if the for-await-of loop is broken', async () => {
      let state = 'idle';
      const source = new Observable<number>((subscriber) => {
        subscriber.next(1);
        setTimeout(() => subscriber.next(2), 100);
        setTimeout(() => subscriber.next(3), 200);
        setTimeout(() => subscriber.complete(), 500);

        return () => {
          state = 'unsubscribed'
        }
      });

      expect(state).to.equal('idle')

      const results: number[] = [];
      for await (const value of source) {
        results.push(value);
        break;
      }

      expect(state).to.equal('unsubscribed')
      expect(results).to.deep.equal([1]);
    });
```

src/internal/Observable.ts
```diff
@@ -392,4 +392,123 @@ export class Observable<T> implements Subscribable<T> {
...
+      },
+      return: (): Promise<IteratorResult<T>> => {
+        subscription?.unsubscribe();
+        // NOTE: I did some research on this, and as of Feb 2023, Chrome doesn't seem to do
+        // anything with pending promises returned from `next()` when `throw()` is called.
+        // However, for consumption of observables, I don't want RxJS taking the heat for that
+        // quirk/leak of the type. So we're going to resolve all pending promises we've nexted out here.
+        handleComplete();
+        return Promise.resolve({ value: undefined, done: true });
+      },
```
ktrz, line 508: TIL
I didn't know about this part of the iteration protocol. This is great!

ktrz (commented):
Looks great! I just left one question regarding the tests

benlesh:
CORE TEAM: Approval.

src/internal/Observable.ts
```diff
@@ -392,4 +392,123 @@ export class Observable<T> implements Subscribable<T> {
...
+              }
+            },
+            error: handleError,
+            complete: handleComplete,
+          });
+        }
+
+        // If we already have some values in our buffer, we'll return the next one.
+        if (values.length) {
+          return Promise.resolve({ value: values.shift()!, done: false });
```
bakkot, line 473: This is arguably a violation of the AsyncIterator protocol in the case that the value produced by the observable is a Promise - it will result in you seeing a Promise as the value in a for-await loop, as in

```js
for await (let item of observable) {
  assert(item instanceof Promise); // not possible with built-in async iterators
}
```

The language generally tries to prevent this from happening - that is, if you `yield` a Promise in an AsyncGenerator it will be `await`'d, and if you `for await` over a sync iterable containing Promises then the Promises will be `await`'d before each iteration of the loop.

Also the [async iterator helpers proposal](https://github.com/tc39/proposal-async-iterator-helpers) is intending to add helpers like `map` (etc) to the async iterator prototype, and the values returned by the callbacks passed to the helpers will be `await`d, which similarly prevents you from observing a Promise in the `value` slot of an iteration result from an AsyncIterator.

---

Now, this is kind of a weird case. Nothing in the language formally prevents you from putting a Promise there, it just avoids doing so itself. There was a [whole debate](https://docs.google.com/presentation/d/1U6PivKbFO0YgoFlrYB82MtXf1ofCp1xSVOODOvranBM/edit#slide=id.p) about how to handle this during the design of async iterables. But the upshot was, to quote one of the slides about the design (option 3) which was ultimately chosen,

> you could manually assemble an async iterator whose next() returns promises whose value is a promise, and we wouldn't unwrap them, and consumers would see the inner promise. That's fine; it's a contract violation.

That's what's happening here.

---

If you want to avoid the contract violation and match the rest of the language better, you can do something like `return Promise.resolve(values.shift()).then(value => ({ value, done: false }))` (though I haven't thought about the error case - probably you need to unsubscribe in that case).

src/internal/Observable.ts
```diff
@@ -392,4 +392,123 @@ export class Observable<T> implements Subscribable<T> {
...
+    };
+
+    const handleComplete = () => {
+      completed = true;
+      while (deferreds.length) {
+        deferreds.shift()![0]({ value: undefined, done: true });
+      }
+    };
+
+    return {
```
bakkot, line 454: You may (or may not) want to make this object inherit from `AsyncIterator.prototype`, which already exists but in [the async-iterator-helpers proposal](https://github.com/tc39/proposal-async-iterator-helpers) will gain additional helper methods like `map`. (Prior to that proposal shipping, the only thing inheriting gives you is the implementation of `Symbol.asyncIterator` you have below.)

`AsyncIterator` isn't currently a global, but you can today access `AsyncIterator.prototype` by doing `(async function*(){})().__proto__.__proto__.__proto__`.

---

That said, in that proposal the new `AsyncIterator.from` helper is the intended way for users to get async iterators out of things, and that will wrap the result up in a wrapper which inherits from `AsyncIterator.prototype`, so `AsyncIterator.from(observable)` will have the new helper methods on it either way. So even in the current state the only way users could get at a thing without the new helpers would be by doing `observable[Symbol.asyncIterator]()`. Still, that's a thing they might do, so I think it would make sense to inherit.

benlesh: We don't want to do that. This an interop point, I don't want people to deoptimize a pure push type like Observable to a pull/push type like AsyncIterable just to `map`. We have a `map` we export.

src/internal/Observable.ts
benlesh: I think you're conflating Async Generators and AsyncIterable here. While it's true that yielding a Promise in a generator will cause the promise to resolve before the value comes out in the async iteration result, there's no hard-and-fast rule that says "you can't have an AsyncIterator<Promise<T>>". If it wasn't allowed to work that way, then the `for await` loop would resolve the value.

In any case, this is a pretty extreme corner case, as I've never seen anyone create an `Observable<Promise<T>>`, and it seems even more unlikely they'd do that with the intent to interop to an async function, unless they REALLY wanted to do that for some reason.

I don't really want to add code that will further deoptimize this unless there's a compelling reason.

As far as async iterator helpers, I think I answered that in another comment.

bakkot: > I think you're conflating AsyncGenerator semantics and AsyncIterable consumption here. 

Well, yes and no. There's another case, which is, sync iterators get wrapped up into async iterators by `for await` loops, and the wrapper `await`s any promises yielded by the sync iterator specifically so as to not expose Promises to `for await`.

And right now async generators and the async-from-sync wrapper are the only things in the language which produce async iterators, so everything (all two things) in the language ensures this property is maintained. The point about async iterator helpers was just to say that the helpers will add a new built-in way of producing async iterators (from other iterators), and they'll maintain this property as well (though mostly by necessity - you really want to be able to pass an async function to `.map` on an async iterator, which means you have to await its result anyway).

But yes, as you say nothing in the language stops you from doing this when constructing async iterators manually. Perfectly reasonable to decide it's not worth the overhead in order to maintain consistency in this edge case.

jayphelps (changes requested):
Generally good, but some points that I at least wanted to run by you.

spec/Observable-spec.ts
benlesh: I'll change this one around to test an active subscription count.

src/internal/Observable.ts
```diff
@@ -347,4 +347,123 @@ export class Observable<T> implements Subscribable<T> {
...
+      next: (): Promise<IteratorResult<T>> => {
+        if (!subscription) {
+          // We only want to start the subscription when the user starts iterating.
+          subscription = this.subscribe({
+            next: (value) => {
+              if (deferreds.length) {
+                deferreds.shift()![0]({ value, done: false });
```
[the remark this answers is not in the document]
benlesh: A solid suggestion. Implemented.

jayphelps (approved):

src/internal/Observable.ts
```diff
@@ -347,4 +347,126 @@ export class Observable<T> implements Subscribable<T> {
...
+   * main();
+   * ```
+   */
+  [Symbol.asyncIterator](): AsyncGenerator<T, void, void> {
+    let subscription: Subscription | undefined;
+    let hasError = false;
+    let error: any;
+    let completed = false;
+    const values: T[] = [];
+    const deferreds: [(value: any) => void, (reason: any) => void][] = [];
```
jakovljevic-mladen, line 392:
```suggestion
    const deferreds: [(value: IteratorResult<T>) => void, (reason: unknown) => void][] = [];
```

src/internal/Observable.ts
```diff
@@ -347,4 +347,126 @@ export class Observable<T> implements Subscribable<T> {
...
+   *  }
+   * }
+   *
+   * main();
+   * ```
+   */
+  [Symbol.asyncIterator](): AsyncGenerator<T, void, void> {
+    let subscription: Subscription | undefined;
+    let hasError = false;
+    let error: any;
```
jakovljevic-mladen, line 389: If errors should be typed `unknown` in V8, we might wanna do it right away:

```suggestion
    let error: unknown;
```

src/internal/Observable.ts
```diff
@@ -347,4 +347,126 @@ export class Observable<T> implements Subscribable<T> {
...
+    let completed = false;
+    const values: T[] = [];
+    const deferreds: [(value: any) => void, (reason: any) => void][] = [];
+
+    const handleError = (err: any) => {
```
jakovljevic-mladen, line 394:
```suggestion
    const handleError = (err: unknown) => {
```

jakovljevic-mladen (approved):
LGTM, only minor types issues that could be considered fixing.

PR #7190, opened 2023-02-20
docs: remove examples-package

jakovljevic-mladen:
**Description:**
This project does not have complex examples like Angular, thus we don't have to have infrastructure for this. Since we're just library, our examples are always written in a single file. The plan is to remove as much Angular specific stuff from the docs app as possible.

None


benlesh (approved):

jakovljevic-mladen:
Cherry-picked to `7.x`.

PR #7197, opened 2023-02-28
fix(asapScheduler): No longer stops after being scheduled twice during flush

### LINKED ISSUE #7196: asapScheduler: Scheduling inside of an executing action only works once
Originally reported here: https://github.com/ReactiveX/rxjs/issues/6747#issuecomment-1360958245

Failing test:
```ts
  it('scheduling inside of an executing action more than once should work', (done) => {
    const results: any[] = [];

    asapScheduler.schedule(() => {
      results.push(1)
      asapScheduler.schedule(() => results.push(2));
      asapScheduler.schedule(() => results.push(3));
    });

    setTimeout(() => {
      // This should always fire after two recursively scheduled microtasks.
      expect(results).to.deep.equal([1, 2, 3]);
      done() 
    })
  })
```

benlesh:
Fixes an issue where trying to share a microtask cause the scheduler to trip over itself. Instead, each scheduled item will get its own microtask.

resolves #7196

### DIFF
--- a/spec/schedulers/AsapScheduler-spec.ts
+++ b/spec/schedulers/AsapScheduler-spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
 import * as sinon from 'sinon';
-import { asapScheduler, Subscription, SchedulerAction, merge } from 'rxjs';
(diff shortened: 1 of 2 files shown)


benlesh:
Closed by #7197. (Thanks, @trxcllnt)

PR #7198, opened 2023-02-28
fix(asapScheduler): No longer stops after scheduling twice during flush

trxcllnt:
Fixes an issue where trying to share a microtask cause the scheduler to trip over itself.

resolves ReactiveX#7196

A non-breaking alternative fix to PR https://github.com/ReactiveX/rxjs/pull/7197


benlesh (approved):

benlesh:
Also applied to 8.x (master)

Issue #7201, opened 2023-03-01
RFC: When to remove `Observable#lift`?


benlesh:
This is an issue to track to the deprecation and removal of `lift` in v9.

`lift` was a really interesting idea and a revolutionary thing we added back in v5. The initial promises it held were around performance, the idea that you could add custom "methods" to an inherited custom `Observable` and keep them after calling a native method, and that enabled things like `Subject` to still be a `Subject` after you operated on it.

## Why we'd want to remove it

1. It means we're implementing our operators the _same way_ we'd recommend other people implement their operators.
2. It means that we're compatible with any "same shaped" observable implementation more directly. )See test in associated PR here: https://github.com/ReactiveX/rxjs/pull/7202/commits/6403d8a224267faca1c42a086a3d53802b8f9afb#diff-c8b60d7f1c5a1e59f4a311cdc025d29918fef109c66428feaa6a7aa193803928R283-R305)
3. It reduces code size and code complexity.
4. It preps us for a future where a standalone RxJS `Observable` doesn't have anything "RxJS-specific" on it, (except maybe `Observable#pipe`, which is just generally useful).
5. We stopped using "methods" for operators years ago, and we've long recommended against people inheriting from `Observable`. In fact, it's an industry standard that subclassing types you don't own is a bad practice unless it's documented API (React.Component, or something like that). So `lift` doesn't have the same purpose it used to.

## Remove in v8 or v9?

So there's some debate to be had here with the Core Team:

While we've already had `lift` deprecated for a _very long time_. We're making it internal in v8 (per the comments in the codebase). And the "lifted observable" behavior is still supported, even if it's not supported by TypeScript.

Should we completely remove `lift` in v8? We can get away with this since it's been slated to be made internal (meaning it's not a public API in version 8). In other words, if it's not a public API, no one should have any expectations of behaviors associated with it beyond things that RxJS _itself_ does.

The _only_ issue I see is currently RxJS supports the composition of our subject variants through the operator chain, via lift. This is a little known aspect of RxJS that isn't used much, in no smart part because TypeScript can't support it:

```ts
const subject = new Subject<number>()

const lifted = subject.pipe(map(n => n + '?'), map(s => s + '!');

lifted.subscribe(console.log);

(lifted as any).next(42); // logs "42?!"
```

The workaround if we remove `lift` entirely is the obvious and correct thing:

```ts
const subject = new Subject<number>()

[...]

PR #7202, opened 2023-03-01
feat(Observable): RxJS doesn't even lift.

benlesh:
Removes `lift`, `source`, and `operator` from the Observable API. These have been deprecated for more than 3 years. It's time.

#7201


demensky:
image

src/internal/Subject.ts
```diff
@@ -108,7 +100,7 @@ export class Subject<T> extends Observable<T> implements SubscriptionLike {
 
   /** @internal */
   protected _subscribe(subscriber: Subscriber<T>): Subscription {
-    this._throwIfClosed();
+    // this._throwIfClosed();
```
demensky, line 103: Is there any additional change planned?

```suggestion
```

src/internal/util/lift.ts
```diff
@@ -17,16 +9,5 @@ export function hasLift(source: any): source is { lift: InstanceType<typeof Obse
 export function operate<T, R>(
   init: (liftedSource: Observable<T>, subscriber: Subscriber<R>) => (() => void) | void
 ): OperatorFunction<T, R> {
```
demensky, line 11:
```suggestion
export function operate<T, R>(
  init: (source: Observable<T>, subscriber: Subscriber<R>) => (() => void) | void
): OperatorFunction<T, R> {
```

Probably JSDoc should also be updated.

src/internal/util/lift.ts
```diff
@@ -17,16 +9,5 @@ export function hasLift(source: any): source is { lift: InstanceType<typeof Obse
...
-  return (source: Observable<T>) => {
```
demensky, line 20: Well, the file name is no longer true.

benlesh: I'll address this in a followup.

src/internal/util/lift.ts
benlesh: I'll address this in a followup.

benlesh:
CORE TEAM: Approval from core team. No one wants to review it. Ben will just have to trust @demensky.

@kolodny said he'll try. But also that Netflix doesn't review code.

benlesh:
I blame @jayphelps

jayphelps:
@benlesh so do I.

spec/operators/delay-spec.ts
```diff
@@ -237,7 +237,7 @@ describe('delay', () => {
       const result = e1.pipe(
         repeatWhen((notifications) => {
           const delayed = notifications.pipe(delay(t));
-          subscribeSpy = sinon.spy((delayed as any)['source'], 'subscribe');
+          subscribeSpy = sinon.spy(notifications as any, 'subscribe');
```
kolodny, line 240: Small nit but I believe you can do `sinon.spy<any>(notifications)` instead

spec/operators/map-spec.ts
```diff
@@ -279,4 +279,29 @@ describe('map', () => {
...
+    it('can operate on any Observable-shaped object', () => {
+      const source = {
+        subscribe(observer: Observer<number>) {
+          observer.next(1);
+          observer.next(2);
+          observer.next(3);
+          observer.complete();
+          return {
+            unsubscribe() {
+              // noop
```
kolodny, line 293: Do we want to verify this was called? If not can we remove this?

kolodny (approved):

Issue #6899, opened 2022-03-18
ignoreElements operator changing the generic type before itself in the pipeline


188599:
### Describe the bug

In summary, this is what is happening:

```ts
const $source = of(1).pipe(
  // x is of type unkown here
  tap(x => console.log(x)),
  ignoreElements()
);

const $source2 = of(1).pipe(
  // x is of type number here
  tap(x => console.log(x))
);
```

### Expected behavior

```ts
const $source = of(1).pipe(
  // x should be of type number here
  tap(x => console.log(x)),
  ignoreElements()
);
```

### Reproduction URL

https://stackblitz.com/edit/rxjs-maxffn?file=index.ts

### Version

7.5.2

### Additional context

This may be a regression bug, or at least I don't recall this happening in v6.

188599:
Tested the same thing on v6, it seems that back then it would still change the generic type, but it was to any.

188599:
Didn't find #6765 before opening the issue, sorry.

Closing it.

Issue #6765, opened 2022-01-13
ignoreElements breaks inferred types in proceeding operators


robmcm:
### Describe the bug

Using the ignoreElements operator seems to break the type inference in operators preceding it. It looks like in version 7 the inferred type is now unknown, where as before it was any, causing the issue to now be more noticeable. 

Example: https://codesandbox.io/s/epic-dijkstra-79vsv?file=/src/index.ts:23-462 

### Expected behavior

Types shouldn't be broken when using ignoreElements.

### Reproduction code

```typescript
import { ignoreElements, tap } from "rxjs/operators";
import { of, merge } from "rxjs";

type TestObject = { payload: string };
const testInstance: TestObject = { payload: "test" };

const test$ = of(testInstance);

const output$ = merge(
  test$.pipe(
    tap((x) => x.payload), // x is typed as unknown here
    ignoreElements()
  ),
  test$.pipe(
    tap((x) => x.payload) // x is TestObject here
  )
);

output$.subscribe(console.log);
```

### Reproduction URL

https://codesandbox.io/s/epic-dijkstra-79vsv?file=/src/index.ts:23-462

### Version

7+

robmcm:
I noticed a similar issue was reported in May last year (https://github.com/ReactiveX/rxjs/issues/5460) with version 6.5.5. This seems to be reproducible in codesandbox when updating to rxjs:7.5.2 and typescript: 4.5.4.

The resolution wasn't clear if it was indeed a non-issue with rxjs cc   @EnricoPicci @tonivj5 @cartant

tonivj5:
@robmcm it's related to Typescript config and strictness:

[...]

Issue #5460, opened 2020-05-28
ignoreElements forces to "any" the type of the input variable of a preceding operator


EnricoPicci:
## Bug Report

**Current Behavior**
When `ignoreElements()` is used after an operator which expects an input, the type of the input of the preceding operator is forced to `any`

**Reproduction**
- https://stackblitz.com/edit/rxjs-3zaonl

In the following code, the type inferred for the variable `data` is `any` while it should be `string`

```js
  let obs$: Observable<string>
  obs$.pipe(
    tap(data => {
    }),
    ignoreElements()
  );
```

**Expected behavior**
The inferred type of the variable of the variable should not change

**Possible Solution**
The solution is to use generics to define the type of `ignoreElements()` like 

```js
export declare function ignoreElements<T>(): OperatorFunction<T, never>;
```

I tried to create a PR but was not able to push the branch to https://github.com/ReactiveX/rxjs.git/

cartant:
It should infer `never`. It does for me:

<img width="484" alt="ignoreElements" src="https://user-images.githubusercontent.com/3878593/83145077-9d25e100-a137-11ea-8802-d19f6979deba.png">

https://github.com/ReactiveX/rxjs/blob/3d6e4c5bf2ba2eec31fdf2498a9fcc9ad364511f/src/internal/operators/ignoreElements.ts#L33

tonivj5:
I think what @EnricoPicci is reporting is that `data` is inferred as `any` when `ignoreElements` operator is used.

From its stackblitz:

```ts
  obs$.pipe(
    tap(data /* <--- is any */ => {
    }),
    ignoreElements()
  );

  obs$.pipe(
    tap(data /* <--- is string */ => {
    }),
  );
```

cartant:
With the latest versions of RxJS and TypeScript, it's fine in VS code, so IMO this is not an RxJS issue:

<img width="503" alt="ignored" src="https://user-images.githubusercontent.com/3878593/83149379-589d4400-a13d-11ea-870c-4837f86956af.png">

tonivj5:
You're right @cartant :+1:

It works in other platfroms (using latest) https://codesandbox.io/s/great-surf-f7bi5?file=/src/index.ts

EnricoPicci:
Thanks, it is clearly not an rxjs issue

Issue #5236, opened 2020-01-16
add generic type to ignoreElements() signature


mkurcius:
## Bug Report

**Current Behavior**
Current signature of ignoreElements() looks as follow:
`function ignoreElements(): OperatorFunction<any, never>`
this `any` as first generic type cause this code to compile:
```ts
interface SomeObject {
  x: number;
}

const source$: Observable<SomeObject> = of({x: 1}, {x: 2}, {x: 3});
source$.pipe(
  tap(obj => console.log(obj.y)),
  ignoreElements()
);
```
where it should fail because `SomeObject` does not have property `y`

if you remove `ignoreElements()` code does not compile as expected

**Reproduction**
```ts
interface SomeObject {
  x: number;
}

const source$: Observable<SomeObject> = of({x: 1}, {x: 2}, {x: 3});
source$.pipe(
  tap(obj => console.log(obj.y)),
  ignoreElements()
);
```

**Expected behavior**
Operators inside pipe should infer proper type

**Environment**
- RxJS version: verified on 6.4.0 but newer version should has same bug

**Possible Solution**
change signature to: `function ignoreElements<T>(): OperatorFunction<T, never>`

cartant:
Not an RxJS bug, IMO. An error is effected for me, using TypeScript version 3.7.4. What version of TypeScript are you using?

mkurcius:
Ok, it may be it, I am using Typescript 3.5.3 (version which comes with Angular 8)

cartant:
Closing this as it does not appear to be an RxJS issue and because specifying a type parameter that is used only in the return type is deemed to be poor practice.

PR #7204, opened 2023-03-02
fix(ignoreElements): change the return type from unknown to Generic Type

tristan-laroubine:
**Description:**
The old signature of the function forced the previous types to unknown.

Taking into account @cartant `s comment https://github.com/ReactiveX/rxjs/issues/5236#issuecomment-575399386, losing the typing for the previous operators forces to duplicate the typing information when it is not necessary.
Losing the explicit typing and as a side effect duplicating the type information is also a bad practice.

### Before [[stackblitz]](https://stackblitz.com/edit/typescript-zpw3s7?file=index.ts)

```typescript
of('a string').pipe(
  tap((value) => {
    /** string **/
  }),
  tap((value) => {
    /** unknow **/
  }),
  ignoreElements(),
  tap((value) => {
    /** never **/
  })
);
```
### Now

```typescript
of('a string').pipe(
  tap((value) => {
    /** string **/
  }),
  tap((value) => {
    /** string **/
  }),
  ignoreElements(),
  tap((value) => {
    /** never **/
  })
);

```

**Related issue :**
https://github.com/ReactiveX/rxjs/issues/6765
https://github.com/ReactiveX/rxjs/issues/6899
https://github.com/ReactiveX/rxjs/issues/5236
https://github.com/ReactiveX/rxjs/issues/5460

### DIFF
--- a/spec-dtslint/operators/ignoreElements-spec.ts
+++ b/spec-dtslint/operators/ignoreElements-spec.ts
@@ -17,3 +17,18 @@ it('should not break the inference of type', () => {
     ignoreElements(),
   );
 });
+
+it('should not break the inference of types when piped in a larger chain', () => {
+  const o$ = of('a string').pipe(
+    tap((o) => {
+      const t = o; // $ExpectType string
+    }),
+    tap((o) => {
+      const t = o; // $ExpectType string
+    }),
+    ignoreElements(),
+    tap((o) => {
+      const t = o; // $ExpectType never
+    }),
+  );
+});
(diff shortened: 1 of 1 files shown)


benlesh:
This looks okay to me, but we could use a `dtslint` test for it.

benlesh:
@tristan-laroubine I rebased this PR for you and pushed up some changes:

1. The added dtslint test passed without any changes.
2. To be thorough, I added the exact example you posted in the original issue as a dtslint test. It also passed without any changes.


...Therefore I'm not going to merge the change, but I'll probably merge the tests.

Maybe check your TypeScript version?

Also, just a friendly reminder to everyone (and I'm guilty of this myself) but make sure to thoroughly test what you're asserting as a required change in a PR, in an all-volunteer repo like this time is at a premium. :) That said, every contribution is absolutely appreciated and welcome.

benlesh (approved):
Approved, but the commit needs to read that it's just an added test. The actual TS change was reverted.

fgirardey:
Can it depends whether TypeScript is configured as strict or not?

LcsGa (approved):

PR #7207, opened 2023-03-07
fix: debounceTime with asapScheduler deadlock because of nested calls

yuri-apanasik:
**Description:**

**BREAKING CHANGE:** 

### DIFF
--- a/src/internal/scheduler/AsapScheduler.ts
+++ b/src/internal/scheduler/AsapScheduler.ts
@@ -4,7 +4,15 @@ import { AsyncScheduler } from './AsyncScheduler';
 export class AsapScheduler extends AsyncScheduler {
   public flush(action?: AsyncAction<any>): void {
     this._active = true;
+
+    const { actions } = this;
+    let error: any;
+    action = action || actions.shift()!;
+
     // The async id that effects a call to flush is stored in _scheduled.
+    // But in case of nested asyncScheduler calls _scheduled could be cleared,
+    // so we take flushId from the first action to guarantee execution of all
+    // actions with the same id.
     // Before executing an action, it's necessary to check the action's async
     // id to determine whether it's supposed to be executed in the current
     // flush.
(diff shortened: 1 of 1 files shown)


benlesh:
Hi, @yuri-apanasik, thank you for your contribution. Please add a test and confirm that the test is broken in master _prior_ to changing any code.

yuri-apanasik:
Hi @benlesh, as a wrote here (https://github.com/ReactiveX/rxjs/pull/7207) I have a problem with test, behavior is different in testing environment and runtime. Do you have any ideas why this difference could be?

src/internal/scheduler/AsapScheduler.ts
```diff
@@ -4,7 +4,15 @@ import { AsyncScheduler } from './AsyncScheduler';
 export class AsapScheduler extends AsyncScheduler {
   public flush(action?: AsyncAction<any>): void {
     this._active = true;
+
+    const { actions } = this;
+    let error: any;
+    action = action || actions.shift()!;
```
LcsGa, line 10:
```suggestion
    action = action ?? actions.shift()!;
```

LcsGa (approved):

pmoleri:
Hi @yuri-apanasik 
I think that the correct code should be analogous to this patch:
https://github.com/ReactiveX/rxjs/pull/7444/files#diff-64da125812f733a97ec0d18250db7664156743309052001c7a5eeaf2109a822b
With `action.id` taking precedence and `_scheduled` only cleaned when it's their case.

> Hi @benlesh, as a wrote here (#7207) I have a problem with test, behavior is different in testing environment and runtime. Do you have any ideas why this difference could be?

In my PR I also noticed some asap tests that were failing until I changed the way in which fakeTimers were registered.
You could check if that fixes (or make fail) your tests too,

yuri-apanasik:
Hi @pmoleri , to tell the truth I even forgot about this PR... As I remember and as I can see in the code, global logic with both patches (your or mine) will be the same. I would say we can wait until at least one PR will be merged by maintainers and adjust another PR so code in both animationFrame/asap schedulers is consistent.
Thank you for the hint with tests. I'll try to fix them on the weekend.

pmoleri:
@yuri-apanasik 

> global logic with both patches (your or mine) will be the same

The code communicates different intents, although, in practice it might end up being equivalent because of specifics about asapScheduler.

In your code  _scheduled``  is taking precedence as flushid and is always clearing the _scheduled, even if the dispatched action was in the parameter with a different id.
This has 2 consequences:
 - It may flush unrelated actions
 - It will clear the _scheduled when it's not its turn.

In asapScheduler, in my experience `_scheduled` actions _always_ get flushed before any other delayed action, so those consequences don't have any negative impact. In animationFrameScheduler it's critical, because delayed action can indeed happen before `_scheduled` ones having very negative consequences.

PR #7211, opened 2023-03-08
fix(subjects): All subjects will now correctly be `closed`

### LINKED ISSUE #7200: Windows from `windowX` operators shouldn't throw if subscribed to "late"
We have a test confirming that if you subscribe to a window after the parent subscription has been unsubscribed it should throw an `ObjectUnsubscribedError`. That honestly doesn't make any sense:

1. It's a corner case with an esoteric operator (we only test it in `windowToggle`) I doubt most people hit.
2. It doesn't make sense. The window is `complete` it's not an error.
3. It seems to be a side effect of an implementation detail of `Subject#asObservable` where `Subject` "knows" about `Observable#source`.

I want to faze out `Observable#lift` for v9 (and by proxy `Observable#operator` and `Observable#source`), and this is sort of in the way.

### LINKED ISSUE #7209: BUG: `Subject#closed` not set on `complete` or `error` and `ObjectUnsubscribedError` is nonsense.
Issues:

1. `ObjectUnsubscribedError` is only thrown when a `Subject` is `closed`.
2. `closed` is not `true` if the subject is completed or errored.
3. There's _HOT PATH_ code in every `next`, `error`, `complete`, and `subscribe` of every `Subject` to check to see if we need to throw an `ObjectUnsubscribedError`. 
4. No one has seen the benefit of `ObjectUnsubscribedError` in practice.

Related: #7200

benlesh:
...after `complete()` and `error()`. Removed `ObjectUnsubscribedError`. Unsubscribed subjects will no longer throw.

BREAKING CHANGE: `ObjectUnsubscribedError` is no longer exported. The constructor was long deprecated. If you need a similar error, you can construct and throw your own. RxJS no longer throws this error.

BREAKING CHANGE: `Subject#closed` is no longer writeable. That was never an intended API. If you need to silently close a `Subject`, call `Subject#unsubscribe()`.

BREAKING CHANGE: Windows from `windowToggle` will no longer throw if leaked and subscribed outside of the main subscription. There's no real use case for this behavior. If you find you needed this behavior, please file and issue.

fixes #7200
fixes #7209

feat(Subject): `isStopped`, `observers`, `hasError`, and `thrownError` no longer in public API.
    
    BREAKING CHANGE: `isStopped`, `observers`, `thrownError` and `hasError` are no longer part of the public API. The only known use case for any of these was detecting that a Subject was observed, like so: `if (subject.observers.length > 0)`. That shouldn't be done. Instead use `if (subject.observed)`.


src/internal/Subject.ts
```diff
@@ -102,21 +97,19 @@ export class Subject<T> extends Observable<T> implements SubscriptionLike {
 
   /** @internal */
   protected _trySubscribe(subscriber: Subscriber<T>): TeardownLogic {
```
pavelxdd, line 99: This method can probably be removed, since it just propagades the call to parent class.

spec/Subject-spec.ts
```diff
@@ -345,33 +345,35 @@ describe('Subject', () => {
...
-
-    subject.subscribe(function (x) {
-      expect(x).to.equal(expected.shift());
-    });
+    const results: any[] = []
+    subject.subscribe(x => results.push(x));
 
     subject.next('foo');
     subject.unsubscribe();
-    expect(() => subject.next('bar')).to.throw(ObjectUnsubscribedError);
```
jakovljevic-mladen, line 373: This needs to be documented somehow.

spec/Subject-spec.ts
```diff
@@ -345,33 +345,35 @@ describe('Subject', () => {
...
 
-  it('should not allow values to be nexted after it is unsubscribed', (done) => {
+  it('should not allow values to be nexted after it is unsubscribed', () => {
     const subject = new Subject<string>();
-    const expected = ['foo'];
-
-    subject.subscribe(function (x) {
-      expect(x).to.equal(expected.shift());
-    });
+    const results: any[] = []
```
jakovljevic-mladen, line 369: If possible, please turn Git hooks on so that Prettier formats this (missing `;`). The issue is present in multiple places.

src/internal/Subject.ts
```diff
@@ -14,17 +13,23 @@ import { arrRemove } from './util/arrRemove';
  * Subject, and you can call next to feed values as well as error and complete.
  */
 export class Subject<T> extends Observable<T> implements SubscriptionLike {
-  closed = false;
+  /** @internal */
+  _closed = false;
+
+  /**
+   * Will return true if this subject has been closed and is no longer accepting new values.
```
jakovljevic-mladen, line 20: I'd add docs somewhere to indicate that `Subject`s don't throw anymore if used after close and that that behavior is new as of V8.

src/internal/Subject.ts
```diff
@@ -14,17 +13,23 @@ import { arrRemove } from './util/arrRemove';
...
+  get closed() {
+    return this._closed;
+  }
 
   private currentObservers: Observer<T>[] | null = null;
 
-  /** @deprecated Internal implementation detail, do not use directly. Will be made internal in v8. */
+  /** @internal */
   observers: Observer<T>[] = [];
-  /** @deprecated Internal implementation detail, do not use directly. Will be made internal in v8. */
```
jakovljevic-mladen, line 23: _Will be ~made internal~ **removed** in v8._ 😂 Sorry about that...

jakovljevic-mladen (approved):
LGTM, needs docs updates. Either in this PR or in a follow-up.

spec/Subject-spec.ts
demensky: Do we happen to have a task for this?  I'll try to do this when I have time.

jakovljevic-mladen: We have git hooks enabled, contributors just have to perform `git commit` without `--no-verify` option.

benlesh: The specs don't have prettier running on all of them because it messes up the marble diagrams when the test scheduler is being used "the old way" (not using `TestScheduler#run()`). I'm not sure if everything has been ported yet. But I wouldn't want to include a massive formatting change in the same commit (or even PR) as a material change to the code. It would need to be done in a follow up.

src/internal/Subject.ts
benlesh: Removed/internal doesn't matter. People shouldn't be using anything marked internal. Because we could break them without warning.

spec/subjects/BehaviorSubject-spec.ts
```diff
@@ -26,12 +26,27 @@ describe('BehaviorSubject', () => {
...
   });
 
-  it('should throw an ObjectUnsubscribedError if getValue() is called and the BehaviorSubject has been unsubscribed', () => {
+  it('should be closed if unsubscribed', () => {
     const subject = new BehaviorSubject('hi there');
     subject.unsubscribe();
-    expect(() => {
-      subject.getValue();
-    }).to.throw(ObjectUnsubscribedError);
+    expect(subject.closed).to.be.true;
```
ktrz, line 32: What is the current behavior of calling `BehaviorSubject#getValue()` after the subject has been unsubscribed? Does it still just return the last value? Do we want to capture that in a test?

PR #7217, opened 2023-03-14
Change workflow permissions to be minimally scoped.

### LINKED ISSUE #7216: Change workflow permissions to be minimally scoped.
### Describe the bug

It is a known behavior of github workflow to grant write-all permission. Since it can be exploited by an attacker, it is consider a good practice to always use credentials that are minimally scoped. This is a recommendation from both the [OpenSSF Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and the [Github](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions).

### Expected behavior

I've noticed the [rebase.yml](https://github.com/ReactiveX/rxjs/blob/master/.github/workflows/rebase.yml) is following this practice: the top level permission is set to read and the write permission needed is granted on the job level.

I want to apply the same to the [ci_main.yml](https://github.com/ReactiveX/rxjs/blob/master/.github/workflows/ci_main.yml) and [ci_ts_latest](https://github.com/ReactiveX/rxjs/blob/master/.github/workflows/ci_ts_latest.yml).

I'll submit a PR with the changes and will be available to any questions or concerns, besides any request changes!

Thanks!

### Reproduction code

_No response_

### Reproduction URL

_No response_

### Version

None

### Environment

_No response_

### Additional context

_No response_

joycebrum:
**Description:**

- [ci_main.yml](https://github.com/ReactiveX/rxjs/blob/master/.github/workflows/ci_main.yml) top level permission set to contents read. Tested with success at https://github.com/joycebrum/rxjs/actions/runs/4419925844.
- [ci_ts_latest](https://github.com/ReactiveX/rxjs/blob/master/.github/workflows/ci_ts_latest.yml) top level permission ser to contents read. Tested with success at https://github.com/joycebrum/rxjs/actions/runs/4419925846

Closes #7216

### DIFF
--- a/.github/workflows/ci_main.yml
+++ b/.github/workflows/ci_main.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     types: ['opened', 'reopened', 'synchronize']
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci_ts_latest.yml
+++ b/.github/workflows/ci_ts_latest.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     types: ['opened', 'reopened', 'synchronize']
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest


benlesh:
Whoa! Thank you very much, @joycebrum!

benlesh:
Note: Applied to 7.x and 6.x branches as well... just in case.

PR #6913, opened 2022-03-28
chore: include docs build scripts in pipeline + fix scripts

jakovljevic-mladen:
**Description:**
To avoid issues that we had multiple times so far where the docs build was broken after some PR merges, I'm adding running docs build and generation scripts to the pipelines. This only runs `setup` and `build` scripts from the `docs_app`, it does not deploy app remotely.

**Related issue (if exists):**
None, but one of the latest issue we had was fixed in #6881. We want to avoid having such PRs by including these runs. Also, running current `build` script will fail pipelines until #6912 is merged.

PR #7227, opened 2023-03-20
Remove old docs annotations

jakovljevic-mladen:
**Description:**
The main purpose of this PR is to remove old docs annotations. E.g.

```ts
/**
 * @param {someType} someProp Property description.
 */
```

From this example, `{someType}` part is now removed from all `@param` and `@return` tag definitions. The docs generator called these types the Catharsis types which may have been required by the old docs, but aren't with the Angular docs.

Also, many other unused tag definitions are removed (like `@class` or `@nocollapse`). Plus some more Angular specific stuff.

Many removed items are separated by commits for easier review.

None.


jakovljevic-mladen (commented):
@benlesh, I added as much comments as I could find for the easier review, but I also found a couple of other issues that could be fixed, I'll commit them later.

Also, please note that this PR fixes two issues, both of them described in the PR description. And, I really hope this is the last _100+ files changed_ PR from me 🤣

p.s. We now run docs generation script in the pipelines (as of #6913), therefore we're sure that these changes work and that they don't break the docs generation.

src/internal/Subject.ts
```diff
@@ -133,15 +132,14 @@ export class Subject<T> extends Observable<T> implements SubscriptionLike {
...
    * code that uses the Observable.
-   * @return {Observable} Observable that the Subject casts to
+   * @return Observable that this Subject casts to.
    */
   asObservable(): Observable<T> {
     return new Observable((subscriber) => this.subscribe(subscriber));
   }
 }
 
 /**
```
demensky, line 142: Is it a rudiment?

src/internal/ajax/errors.ts
demensky, line 1: Maybe you should wait with this file until the transition to es2015+?

src/internal/observable/forkJoin.ts
```diff
@@ -137,11 +137,9 @@ export function forkJoin<T extends Record<string, ObservableInput<any>>>(
  * @see {@link combineLatest}
  * @see {@link zip}
  *
- * @param {...ObservableInput} args Any number of Observables provided either as an array or as an arguments
+ * @param args Any number of `ObservableInputs`s provided either as an array or as an object
```
demensky, line 140:
```suggestion
 * @param args Any number of `ObservableInput`s provided either as an array or as an object
```

src/internal/observable/zip.ts
```diff
@@ -45,8 +45,10 @@ export function zip<A extends readonly unknown[], R>(
  * // { age: 29, name: 'Beer', isDev: false }
  * ```
  *
- * @param sources
- * @return {Observable<R>}
+ * @param args Any number of `ObservableInputs`s provided either as an array or as an object
```
demensky, line 48:
```suggestion
 * @param args Any number of `ObservableInput`s provided either as an array or as an object
```

src/internal/operators/timeout.ts
```diff
@@ -82,8 +82,6 @@ export interface TimeoutErrorCtor {
  * create your own error class and use that.
  *
  * @see {@link timeout}
- *
- * @class TimeoutError
```
demensky, line 86: Maybe you should wait with this file until the transition to es2015+?

src/internal/util/ArgumentOutOfRangeError.ts
```diff
@@ -17,8 +17,6 @@ export interface ArgumentOutOfRangeErrorCtor {
  * @see {@link elementAt}
  * @see {@link take}
  * @see {@link takeLast}
- *
- * @class ArgumentOutOfRangeError
```
demensky, line 21: Maybe you should wait with this file until the transition to es2015+?

src/internal/util/EmptyError.ts
```diff
@@ -19,11 +19,12 @@ export interface EmptyErrorCtor {
  * @see {@link single}
  * @see {@link firstValueFrom}
  * @see {@link lastValueFrom}
- *
- * @class EmptyError
```
demensky, line 23: Maybe you should wait with this file until the transition to es2015+?

src/internal/util/NotFoundError.ts
```diff
@@ -15,8 +15,6 @@ export interface NotFoundErrorCtor {
  * observable sequence.
  *
  * @see {@link operators/single}
- *
- * @class NotFoundError
```
demensky, line 19: Maybe you should wait with this file until the transition to es2015+?

src/internal/util/ObjectUnsubscribedError.ts
```diff
@@ -16,8 +16,6 @@ export interface ObjectUnsubscribedErrorCtor {
  *
  * @see {@link Subject}
  * @see {@link BehaviorSubject}
- *
- * @class ObjectUnsubscribedError
```
demensky, line 20: Maybe you should wait with this file until the transition to es2015+?

src/internal/util/SequenceError.ts
```diff
@@ -15,8 +15,6 @@ export interface SequenceErrorCtor {
  * values arriving on the observable.
  *
  * @see {@link operators/single}
- *
- * @class SequenceError
```
demensky, line 19: Maybe you should wait with this file until the transition to es2015+?

demensky (commented):
Very cool change! I have a couple of questions and comments.

src/internal/Subject.ts
jakovljevic-mladen: Not really 😁

benlesh (approved):
Spot checked about 90% of the files. Approved out of an abundance of trust.

src/internal/util/ArgumentOutOfRangeError.ts
jakovljevic-mladen: Not sure why? 🙂 The reason why I'm removing these tag definitions is because they don't bring any value. At first, it may seem like they do, but they don't.

For example, the value from the `@see` tag definition two lines above is used [here](https://github.com/ReactiveX/rxjs/blob/f04fe72eba2f117041e1a9c1794f16473ecae697/docs_app/tools/transforms/templates/api/includes/see-also.html). Also, someone reading the code using an editor/IDE (or even Github) could use this information to _see_ `elementAt` and/or `take`/`takeLast`.

Therefore, there are two ways these tag definitions bring value: to the docs app and to the user. `@class` is not being used anywhere: docs site does not use this info from this tag def* and users reading this wouldn't benefit much (the [code below this tag def](https://github.com/ReactiveX/rxjs/pull/7227/files#diff-e696f1c182b04fde2eaf96783e6489b2ab52a66eaedfc2e534c90d1a1efa73c2L23) clearly indicates an object of the `ArgumentOutOfRangeErrorCtor` type).

The third issue is maintainability: renaming stuff (which was done in this library a lot of times) often leaves docs content outdated. Removing stuff like this removes the need to update docs in multiple places which lowers the possibility of having the outdated content.

*The docs app uses the `dgeni` library to [infer](https://github.com/angular/dgeni-packages/blob/4fac18fb8e03265718a11b52cdcb05e72e9fba0c/typescript/src/api-doc-types/ClassExportDoc.ts#L14) the `docType` and to display it [here](https://github.com/ReactiveX/rxjs/blob/f04fe72eba2f117041e1a9c1794f16473ecae697/docs_app/tools/transforms/templates/api/includes/class-overview.html#L5).

jakovljevic-mladen:
> Approved out of an abundance of trust.

Thanks @benlesh!

Two notes:
- Changes from this branch will only land to `master` (for now), therefore, if we spot anything wrong with the content, we can always fix it later (next.rxjs.dev is still not updated with changes from v8).
- Since changes from this PR are in a lot of ways related to the `7.x` branch, I will try to cherry pick them there, but, since these two branches diverged a lot now, I may need to create a new PR for the `7.x` branch.

jakovljevic-mladen:
@benlesh, I managed to cherry-pick this commit to `7.x`, no changes to the code were made. I added back some docs that were removed from `master`.

PR #7228, opened 2023-03-22
refactor(takeLast): Move back to using a ring buffer

benlesh:
This change moves the code back to using a ring buffer for the value collection of takeLast. This should be much more efficient when writing to, and reading from, the buffer.

NOTE: I tried to use github.dev for this, we'll see if it passes CI or not. (lol).

### DIFF
--- a/src/internal/operators/takeLast.ts
+++ b/src/internal/operators/takeLast.ts
@@ -46,40 +46,40 @@ export function takeLast<T>(count: number): MonoTypeOperatorFunction<T> {
   return count <= 0
     ? () => EMPTY
     : operate((source, subscriber) => {
-      // This is a ring buffer that will hold our values
-      let ring = new Array<T>(count)
-      // This counter is how we 
-      let counter = 0
-      source.subscribe(
-        createOperatorSubscriber(
-          subscriber,
-          (value) => {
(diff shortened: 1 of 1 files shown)


src/internal/operators/takeLast.ts
demensky, line 1: Please format the file with Prettier. This will reduce the diff.

src/internal/operators/takeLast.ts
```diff
@@ -46,36 +46,40 @@ export function takeLast<T>(count: number): MonoTypeOperatorFunction<T> {
...
-              }
-              subscriber.complete();
-            },
-            // Errors are passed through to the consumer
-            undefined,
-            () => {
-              // During finalization release the values in our buffer.
-              buffer = null!;
+      // This is a ring buffer that will hold our values
+      let ring = new Array(count)
```
demensky, line 50: any-array

I propose to explicitly specify the type parameter. This will ensure that only `T` is assigned to the array.

```suggestion
      let ring = new Array<T>(count)
```

demensky (changes requested):

jakovljevic-mladen (approved):
LGTM, only a couple of nits.

Issue #7203, opened 2023-03-01
RFC: A new way of "piping"


benlesh:
Given the move away from lift: #7201 #7202 

We can make our observable even more compatible with other observables by migrating us away from the `pipe` method. **This is not something that would happen immediately, rather over a very, very long period**. The idea is to increase compatibility and improve ergonomics.

## Problem: It's annoying to have to `from(x)` types in order to use our operators with them

There's a common issue where people want to convert other things to observables in order to use our operators. Whether it's an array or a promise, there are times were you need to do `from` and the only reason you're doing it is because you want that `.pipe` method, and you want an observable to operate on.

## Problem: `pipe` is a method

1. It's not treeshakable
2. It's unlikely to exist on "other observables".

The second one is more of a problem, honestly. If people implement code with the expectation that `pipe` exists as a method on whatever observable they get, then even while we've made sure our operators will work with any "same-shaped" observable, that custom code will break, because it's expecting `pipe` to be a method.

## Proposal

A new `rx` method that accepts an `ObservableInput` as the first argument, and a rest of operators:

```ts
export function rx<A, B>(source: ObservableInput<A>, a: OperatorFunction<A, B>): Observable<B>;
export function rx<A, B, C>(source: ObservableInput<A>, a: OperatorFunction<A, B>, b: OperatorFunction<B, C>): Observable<C>;
export function rx<A, B, C, D>(source: ObservableInput<A>, a: OperatorFunction<A, B>, b: OperatorFunction<B, C>, c: OperatorFunction<C, D>): Observable<D>;
// ... and so on
export function rx<T>(source: ObservableInput<T>, ...operators: OperatorFunction<T, any>[]): Observable<unknown> {
  returm pipeArray(operators)(from(source))
}
```

(Technically, `rx` is just a standard functional pipe with some specific types, it could be the same `pipe` method we expose under the hood)

The usage would be as such:

```ts
import { rx, map, delay } from 'rxjs'


rx(source$, map(n => n + n)).subscribe(console.log)

// or

rx(somePromise, delay(1000)).subscribe(console.log)

// or

rx(someAsyncIterable, map(n => n + n)).subscribe(console.log)

// ...and so on.
```

eduardoRoth:
I like the approach!

TheLarkInn:
> It's not tree shakable 

💯

UserGalileo:
Seems great!

Is "rx" the final proposed name? I know it'd be a breaking change, but what if...

```ts
[...]

PR #7229, opened 2023-03-23
feat(rx): A new method of piping has been added

benlesh:
+ Adds `rx` a new method of piping observables. This allows any valid observable input to be passed as the first argument, and it will implicitly convert the argument to an Observable and pass it to the function that may be present in the second argument, the return value of THAT function is then passed an argument to the next function, and so on
+ Corrects the types of `Observable#pipe` such that anything can be returned at any step of the functional chain, however the first pipeable function must accept an `Observable<T>`.
+ Adds dtslint and runtime tests for `rx` and `pipe`.

NOTE: This does NOT deprecate`Observable#pipe`. That will be done in a follow up, and we need to define what timeline we'll take to remove that, as it's an API that is broadly used - could be v9, could be v10, could be never. At this point, this is alpha/beta functionality.

BREAKING CHANGE: `Observable#pipe` now allows any pipeable unary function as an argument, just so long as the types properly compose, this means in some cases it will now return `unknown` instead of `Observable<unknown>` the workaround is just to cast the result for those cases. (or you can break your operator piping into smaller pipe sets)

Addresses #7203 

### Note for later: Make sure to update squashed commit message to reflect `rx` and not `r`.


src/internal/util/r.ts
LcsGa, line 1: Won't the `operators[]` syntax be implemented?

benlesh:
After some Core Team discussion, and some testing of the syntax, we've decided to move the implementation to where it takes an array of operators as the second argument. One of the example cases that compelled this change was this:

```ts
r(
  combineLatest([a, b]),
  map(([a, b]) => a + b),
  filter(n => n > 10),
)
.subscribe(console.log)
```

In the above scenario, muscle-memory for RxJS users leads them to believe that `combineLatest` is an operator, where it is in fact creating an observable.

The following was deemed to be more readable:

```ts
r(
  combineLatest([a, b]),
  [
    map(([a, b]) => a + b),
    filter(n => n > 10),
  ],
)
.subscribe(console.log)
```

### This PR is currently broken

I knowingly pushed it up this way so it can be addressed later. There's a problem with the TypeScript types when we do this. TypeScript can't seem to figure out how to type the operators as they go through the array.

benlesh:
cc @DanielRosenwasser 🙏❤️

jannispl:
Personally I am not a huge fan of the new array notation. One of my main issues when dealing with RxJS operators is getting the damn brackets right when refactoring code. Introducing new ones won't help.

benlesh:
After speaking offline with the TypeScript folks, there's just no way to properly type the array notation. So this is going back to the more "flat" approach.

src/internal/util/r.ts
benlesh: Operators are just unary functions... meaning a functions that take one argument. Even in the current pipe functions in RxJS 6.x and 7.x, any unary function will work, it's just our types didn't support it. We're correcting that here.

spec/util/pipe-spec.ts
```diff
@@ -31,3 +31,37 @@ describe('pipe', () => {
     expect(c(someObj)).to.equal(someObj);
   });
 });
+
+describe('r', () => {
```
cartant, line 35: Nitpick:
```suggestion
describe('rx', () => {
```

src/internal/Observable.ts
```diff
@@ -256,72 +255,79 @@ export class Observable<T> implements Subscribable<T> {
...
+    op1: UnaryFunction<Observable<T>, A>,
+    op2: UnaryFunction<A, B>,
+    op3: UnaryFunction<B, C>,
+    op4: UnaryFunction<C, D>,
+    op5: UnaryFunction<D, E>,
+    op6: UnaryFunction<E, F>,
+    op7: UnaryFunction<F, G>,
+    op8: UnaryFunction<G, H>,
+    op9: UnaryFunction<H, Observable<I>>,
     ...operations: OperatorFunction<any, any>[]
```
cartant, line 317: Can we replace the `any` usage here with `unknown`?

src/internal/Observable.ts
```diff
@@ -256,72 +255,79 @@ export class Observable<T> implements Subscribable<T> {
...
+    op1: UnaryFunction<Observable<T>, A>,
+    op2: UnaryFunction<A, B>,
+    op3: UnaryFunction<B, C>,
+    op4: UnaryFunction<C, D>,
+    op5: UnaryFunction<D, E>,
+    op6: UnaryFunction<E, F>,
+    op7: UnaryFunction<F, G>,
+    op8: UnaryFunction<G, H>,
+    op9: UnaryFunction<H, I>,
+    ...operations: UnaryFunction<any, any>[]
```
cartant, line 329: Can we replace the `any` usage here with `unknown`?

src/internal/Observable.ts
```diff
@@ -344,7 +350,7 @@ export class Observable<T> implements Subscribable<T> {
    *   .subscribe(x => console.log(x));
    * ```
    */
-  pipe(...operations: OperatorFunction<any, any>[]): Observable<any> {
+  pipe(...operations: UnaryFunction<any, any>[]): unknown {
```
cartant, line 353: Can we replace the `any` usage here with `unknown`?

src/internal/util/rx.ts
```diff
@@ -0,0 +1,113 @@
...
+  source: ObservableInput<A>,
+  fn2: UnaryFunction<Observable<A>, B>,
+  fn3: UnaryFunction<B, C>,
+  fn4: UnaryFunction<C, D>,
+  fn5: UnaryFunction<D, E>,
+  fn6: UnaryFunction<E, F>,
+  fn7: UnaryFunction<F, G>,
+  fn8: UnaryFunction<G, H>,
+  fn9: UnaryFunction<H, I>,
+  ...fns: UnaryFunction<any, any>[]
```
cartant, line 70: Same question as with `pipe`: Can we replace the `any` usage here with `unknown`?

cartant (approved):
LGTM, just a few questions about `any` usage and a nitpick.

src/internal/Observable.ts
```diff
@@ -256,72 +255,79 @@ export class Observable<T> implements Subscribable<T> {
...
+    op3: UnaryFunction<B, C>,
+    op4: UnaryFunction<C, D>,
+    op5: UnaryFunction<D, E>,
+    op6: UnaryFunction<E, F>,
+    op7: UnaryFunction<F, G>,
+    op8: UnaryFunction<G, H>,
+    op9: UnaryFunction<H, Observable<I>>,
     ...operations: OperatorFunction<any, any>[]
   ): Observable<unknown>;
+  pipe<A, B, C, D, E, F, G, H, I>(
```
kolodny, line 319: I really wish we either had a dedicated pipe operator in js or at least had some way to use recursive/conditional types powerful enough to support not needing to do this.

src/internal/util/rx.ts
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * This means it can be used for anything involving unary functions, just so long as the first unary function accepts an observable as its argument,
+ * and as long as the first argument to `r()` is a valid {@link ObservableInput}.
+ *
+ * This is the same as an ordinary functional {@link pipe}, except it has an implicit `from` as the second argument.
+ *
+ * The following are equivalent:
+ *
+ * ```ts
+ * r(source, map(x => x + 1), filter(x => x % 2 === 0));
```
kolodny, line 86: Just need to update the comment to reflect the `r(source, [map(x => x + 1), filter(x => x % 2 === 0)]);` change

src/internal/util/rx.ts
```diff
@@ -0,0 +1,113 @@
...
+ * pipe(from(source), map(x => x + 1), filter(x => x % 2 === 0));
+ * pipe(source, from, map(x => x + 1), filter(x => x % 2 === 0));
```
kolodny, line 88: Probably also worth calling out that `rx([1, 2, 3]);` is an `Observable<number>` and not `Observable<number[]>` in the documentation (it's already in the type tests)

kolodny (approved):
🚀🚀🚀

src/internal/util/rx.ts
```diff
@@ -0,0 +1,113 @@
...
+ * This means it can be used for anything involving unary functions, just so long as the first unary function accepts an observable as its argument,
+ * and as long as the first argument to `r()` is a valid {@link ObservableInput}.
+ *
+ * This is the same as an ordinary functional {@link pipe}, except it has an implicit `from` as the second argument.
+ *
+ * The following are equivalent:
+ *
+ * ```ts
+ * r(source, map(x => x + 1), filter(x => x % 2 === 0));
+ * pipe(from(source), map(x => x + 1), filter(x => x % 2 === 0));
```
jakovljevic-mladen, line 87: This doesn't seem to be valid. This does:
```suggestion
 * pipe(() => from(source), map(x => x + 1), filter(x => x % 2 === 0));
```

src/internal/util/rx.ts
```diff
@@ -0,0 +1,113 @@
...
+ * all functions have been called, and the result of the last function is returned.
+ *
+ * This means it can be used for anything involving unary functions, just so long as the first unary function accepts an observable as its argument,
+ * and as long as the first argument to `r()` is a valid {@link ObservableInput}.
+ *
+ * This is the same as an ordinary functional {@link pipe}, except it has an implicit `from` as the second argument.
+ *
+ * The following are equivalent:
+ *
+ * ```ts
```
jakovljevic-mladen, line 85: Adding the source would help a lot. E.g.

```const source = Promise.resolve(1);```

or

```const source = [1, 2, 3];```

**Review by jakovljevic-mladen — CHANGES_REQUESTED** [review]
A lot of minor issues found: `r` to `rx` rename should be updated in couple of places, some generics issues need to be addressed and docs needs more info and some updates.

File: src/internal/util/rx.ts (hunk 9 above)
Comment (benlesh) [reply src/internal/util/rx.ts:87]: Fixed. It was a little different than what that said.

Comment (benlesh) [reply src/internal/util/rx.ts:87]: I fixed the types up a bit.

File: src/internal/util/rx.ts (hunk 10 above)
Comment (benlesh) [reply src/internal/util/rx.ts:85]: I added more context.

File: src/internal/util/rx.ts [hunk 11]
```diff
@@ -0,0 +1,113 @@
...
+ * const promise = r(
+ *   of(1, 2, 3),
+ *   async (source) => {
+ *     const result = [];
+ *     await source.forEach(x => result.push(x));
+ *     return result;
+ *   },
+ * });
+ * ````
+ */
```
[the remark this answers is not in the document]
benlesh: I fixed the types up, which should resolve that a little.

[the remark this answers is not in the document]
benlesh: I also added `@param` and `@return` docs.

jakovljevic-mladen (approved):
LGTM overall, there are just some nitpicks that I'd like to be considered fixing.

PR #7231, opened 2023-03-23
fix(subject): Faster subscriptions management in Subject

dubzzz:
**Description:**

The current implementation of the Subject was backing itself on an array of observers. When one observer unsubscribed itself, it was performing a linear scan on the array of observers and dropping the item from the array of observers. The algorithm complexity of this operation was O(n) with n being the number of observers connected to the Subject.

It caused code like:

```js
const allSubscriptions = [];
const source = new Subject();
for (let index = 0 ; index !== 1_000_000 ; ++index) {
  allSubscriptions.push(source.subscribe()); // rather quick
}
for (const subscription of allSubscriptions) {
  subscription.unsubscribe(); // taking very long
}
```

The proposed approach consists into changing our backing collection (an array) into a Map. But we lose somehow the set capability on subject.observers (might possibly be patched in some way).

The command `cross-env TS_NODE_PROJECT=tsconfig.mocha.json mocha --config spec/support/.mocharc.js "spec/**/Subje*-spec.ts"` with the new test added:

- Now: 452ms
- Before: 10s

**BREAKING CHANGE:** 

Subject.observers has been impacted. If needed we can probably make it backward compatible in some ways. Here is what changed:

- Now a getter property, was previously a basic property
- The value change from one call to another -we recompute it anytime we access it)
- There is no setter anymore
- The value is never null

### DIFF
--- a/spec/Subject-spec.ts
+++ b/spec/Subject-spec.ts
@@ -71,7 +71,7 @@ describe('Subject', () => {
       }, complete: done }
     );
 
-    expect(subject.observers.length).to.equal(2);
+    expect(subject.observers.size).to.equal(2);
     subject.next('foo');
     subject.next('bar');
     subject.complete();
@@ -385,11 +385,11 @@ describe('Subject', () => {
       //noop
     });
 
-    expect(subject.observers.length).to.equal(2);
+    expect(subject.observers.size).to.equal(2);
     sub1.unsubscribe();
-    expect(subject.observers.length).to.equal(1);
+    expect(subject.observers.size).to.equal(1);
     sub2.unsubscribe();
-    expect(subject.observers.length).to.equal(0);
+    expect(subject.observers.size).to.equal(0);
     done();
   });
 
--- a/spec/Subject-spec.ts
+++ b/spec/Subject-spec.ts
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import { Subject, ObjectUnsubscribedError, Observable, AsyncSubject, Observer, of, config, throwError, concat } from 'rxjs';
+import { Subject, ObjectUnsubscribedError, Observable, AsyncSubject, Observer, of, config, throwError, concat, Subscription } from 'rxjs';
 import { AnonymousSubject } from 'rxjs/internal/Subject';
 import { catchError, delay, map, mergeMap } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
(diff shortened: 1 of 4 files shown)


spec/Subject-spec.ts
```diff
@@ -683,6 +683,30 @@ describe('Subject', () => {
       expect(true).to.be.true;
     });
   });
+
+  describe('many subscribers', () => {
+    it('should be able to subscribe and unsubscribe huge amounts of subscribers', () => {
+      let numResultsReceived = 0;
+      const allSubscriptions: Subscription[] = [];
+      const source = new Subject<number>();
+      const numSubscribers = 100000;
```
dubzzz, line 692: With 100k items, the version of the code currently on master exceeds the timeout of mocha on a GitHub Codespaces (it reached 10s contrary to new code taking around 400ms). But I don't really know which bar I should put their as the more powerful runners will be the more unlikely the failure even if we reintroduce a slow version. Maybe you have suggestion of it 🤔

kwonoj:
I believe we had similar sort of PR and did not made changes, but search fails me tonfind exact one.

Also in my opinion this seems for extreme cases: if you really have more than even 10,000 subscriptions I'd expect app's architecture need to be revisited. What are the real world problem we'd like to fix with breaking changes?

dubzzz:
> Also in my opinion this seems for extreme cases: if you really have more than even 10,000 subscriptions I'd expect app's architecture need to be revisited. What are the real world problem we'd like to fix with breaking changes?

Good point! On my side, I patched the issue locally by packing all my subscriptions behind a single one, itself pipping data back to my original subscribers. My case was a grid containing thousands of cells in which each cell subscribes to a shared store to see updates. My local fix has been to drop the subscribe done by the cell to an home-made version of a subscribe dealing with adding and removing subscribed callbacks into a Set that will be read by the single real subscription done on the source subject.

I though about submitting this potential fix as actually even if the problem is not that horrible in small cases, we still have extra time taken by a linear look-up while we could just have a immediate delete by switching to a Map. Idea was to improve a lot very huge case but also smaller ones.

dubzzz:
I can probably start to work onto measuring the performance impact on various cases including: 1/2/10/100 subscribers and time to subscribe/emit/unsubscribe. So that we can check whether or not it could bring negative (clearly a no-go) or positive performance impact on small and nominal cases 🤔

kwonoj:
Before doing any additional work may need some archaeology to confirm if there was a public / internal discussions for this and why we didn't. It would be bad if you spend amount of time to prepare PR then we find out some historical reason we can't land this.

benlesh:
I added a test that passes in `master` locally just to make sure this didn't break anything. However, the tests have been running for > 13 minutes now, so I suspect something is wrong. This is a decent idea, but it makes the changes in the wrong spots, and removes other optimizations.

benlesh:
lol...using the in-browser IDE is great, except I can't run the tests. I should really start using Stackblitz for this. 🤔

benlesh:
Actually... wait, I'm moving too fast because I was excited about this optimization. This particular code won't work. Using the Subscription as the key means that in some cases the same Subscription ends up trampling things. These cases are non-obvious because we're returning a Subscriber instance as the Subscription in a lot of cases (as an optimization). That Subscriber instance is often passed through the subscribe call as-is (another optimization), this means that if I rollback to @dubzzz's original commit, a large number of tests fail. The test I added also fails.

We can't accept this as-is... but I'll think about this a bit. This can probably be optimized.

benlesh:
Since I destroyed this one with the Web IDE, I'm going to close it and reopen it in it's original state at #7234 .

PR #7234, opened 2023-03-25
fix(subject): Faster subscriptions management in Subject

benlesh:
Since I HORRIBLY jacked up @dubzzz's PR at #7231 by attempting to use the GitHub Web IDE, I'm starting this new PR that rolls back to just @dubzzz's change, and the test that I added.

### DIFF
--- a/spec/Subject-spec.ts
+++ b/spec/Subject-spec.ts
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import { Subject, ObjectUnsubscribedError, Observable, AsyncSubject, Observer, of, config, throwError, concat } from 'rxjs';
+import { Subject, ObjectUnsubscribedError, Observable, AsyncSubject, Observer, of, config, throwError, concat, Subscription } from 'rxjs';
 import { AnonymousSubject } from 'rxjs/internal/Subject';
 import { catchError, delay, map, mergeMap } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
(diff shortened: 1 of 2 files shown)


benlesh:
I like the idea here, but I immediately spotted that this breaks re-entrant subscriptions.

To following will behave differently before and after this change:

```ts
const subject = new Subject()

subject.subscribe(console.log)
subject.subscribe((value) => {
  console.log(value);
  subject.subscribe(console.log);
})
```

This is because before we were protecting ourselves from re-entrant subscriptions modifying the observers we were looping over.

I also spotted that we're removing an optimization we added that prevented us from needing to copy an Array over and over for this case. (That said, I also noticed that in that code we should be using `for(;;)` array index looping and we're not. This would avoid creating a new iterator for each `next` call)

In any case, @dubzzz's commit is an interesting idea, but unfortunately breaks a lot of our tests. We cannot accept it as is. I'll leave this open for now. And I'll think about it some more.

dubzzz:
> unfortunately breaks a lot of our tests

Oups, sorry about that. I thought I launched then all. I'll re-run them locally to see what goes wrong in the coming days

dubzzz:
> following will behave differently before and after this change

Will try to find a way to make it work with it.

> unfortunately breaks a lot of our tests

CI run seems ok with the current version. Is re-entrant case checked in CI run?

dubzzz:
> following will behave differently before and after this change

Will try to find a way to make it work with it.

> unfortunately breaks a lot of our tests

CI run seems ok with the current version. Is re-entrant case checked in CI run? I'll investigate them closer.

dubzzz:
Hi @benlesh,

Not sure I can push on your PR 🤔 
The branch is under benlesh:pr/dubzzz/7231-1. Maybe I should re-open it.

Maybe I can re-open one?

Note: I have a fix for the re-entrant stuff. I'm checking performance of the new code with the pointers you sent me.

dubzzz:
I reopened it with everything fixed in theory. I also added benchmark results. See https://github.com/ReactiveX/rxjs/pull/7240

PR #7235, opened 2023-03-26
feat(operator): add `delayEach` operator

Puszkarek:
This commit introduces a new operator, delayEach, that delays the emission of each item from the source Observable by a given time duration. The operator is similar to the existing delay operator but applies the delay individually to each item.

**Description:**
This PR introduces a new operator, `delayEach`, that delays the emission of each item from the source Observable by a given time duration. It is similar to the existing `delay` operator, but instead of delaying the entire sequence, it delays each item individually.

##### Motivation
`delayEach` can be helpful in various scenarios where it's necessary to introduce a consistent delay between each item emitted by an Observable. This operator provides a simple and concise way to achieve this without having to create a custom implementation each time.

##### Use Cases
1. Animating elements in a UI sequentially, where each element should appear with a fixed time delay between them.
2. Rate-limiting requests made to an API, ensuring that there is a consistent delay between each request to avoid hitting rate limits.
3. Introducing a delay in a stream of events, such as logs or messages, to simulate real-time behavior during testing or development.

##### Example
```ts
import { delayEach } from './your-custom-operators';

const clicks = fromEvent(document, 'click');
const delayedClicks = clicks.pipe(
  delayEach(3000)
);
delayedClicks.subscribe(x => console.log(x));
```

Before finalizing this PR, I would like to hear your opinions on the usefulness of this operator and whether it would be a valuable addition to the library. If you agree, I will proceed with creating the necessary tests to ensure its functionality and reliability.

Looking forward to your feedback!

### DIFF
--- a/docs_app/content/guide/operators.md
+++ b/docs_app/content/guide/operators.md
@@ -227,6 +227,7 @@ Also see the [Join Creation Operators](#join-creation-operators) section above.
 - [`tap`](/api/operators/tap)
 - [`delay`](/api/operators/delay)
 - [`delayWhen`](/api/operators/delayWhen)
+- [`delayEach`](/api/operators/delayEach)
 - [`dematerialize`](/api/operators/dematerialize)
 - [`materialize`](/api/operators/materialize)
 - [`observeOn`](/api/operators/observeOn)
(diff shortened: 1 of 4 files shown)


kwonoj:
Hi, please refer prior PRs similar with these, i.e https://github.com/ReactiveX/rxjs/pull/6996#issuecomment-1163004529 .

Unless there are absolute reasons, we do recommend to publish it as userland operator since we are trying to reduce api surfaces.

PR #7246, opened 2023-04-09
docs(fromEvent): update the example to match the actual API

### LINKED ISSUE #7245: FromEvent Docs: No overload matches this call.
### Describe the bug

As per the https://rxjs.dev/api/index/function/fromEvent in order to use this operator with capture option we can pass the three arguments to it. 
image
But LSP is showing the following error.
image

### Expected behavior

As per the documentation.
**DOM EventTarget**

This is an object with addEventListener and removeEventListener methods.

In the browser, addEventListener accepts - apart from event type string and event handler function arguments - **optional third parameter, which is either an object or boolean, both used for additional configuration how and when passed function will be called.** When [fromEvent](https://rxjs.dev/api/index/function/fromEvent) is used with event target of that type, you can provide this values as third parameter as well.

### Reproduction code

```typescript
Open the following example in Docs using StackBlitz "Use addEventListener with capture option" (https://rxjs.dev/api/index/function/fromEvent)
```

### Reproduction URL

_No response_

### Version

~7.5.0

### Environment

_No response_

### Additional context

_No response_

demensky:
Updates the example according to the API.

Closes #7245

### DIFF
--- a/src/internal/observable/fromEvent.ts
+++ b/src/internal/observable/fromEvent.ts
@@ -203,9 +203,13 @@ export function fromEvent<T, R>(
  * ```ts
  * import { fromEvent } from 'rxjs';
  *
- * const clicksInDocument = fromEvent(document, 'click', true); // note optional configuration parameter
- *                                                              // which will be passed to addEventListener
- * const clicksInDiv = fromEvent(someDivInDocument, 'click');
+ * const div = document.createElement('div');
+ * div.style.cssText = 'width: 200px; height: 200px; background: #09c;';
+ * document.body.appendChild(div);
+ *
+ * // note optional configuration parameter which will be passed to addEventListener
+ * const clicksInDocument = fromEvent(document, 'click', { capture: true });
+ * const clicksInDiv = fromEvent(div, 'click');
  *
  * clicksInDocument.subscribe(() => console.log('document'));
  * clicksInDiv.subscribe(() => console.log('div'));


src/internal/observable/fromEvent.ts
```diff
@@ -204,8 +204,8 @@ export function fromEvent<T, R>(
  * ```ts
  * import { fromEvent } from 'rxjs';
  *
- * const clicksInDocument = fromEvent(document, 'click', true); // note optional configuration parameter
- *                                                              // which will be passed to addEventListener
+ * // note optional configuration parameter which will be passed to addEventListener
+ * const clicksInDocument = fromEvent(document, 'click', { capture: true });
  * const clicksInDiv = fromEvent(someDivInDocument, 'click');
```
jakovljevic-mladen, line 209: This line fails as well. Can you please create a `div` so it doesn't fail? There are many examples in code base, e.g.

```ts
const div = document.createElement('div');
div.style.cssText = 'width: 200px; height: 200px; background: #09c;';
document.body.appendChild(div);
```

```suggestion
 * const clicksInDiv = fromEvent(div, 'click');
```

jakovljevic-mladen (commented):
LGTM, only a minor suggestion so that this example becomes fully functional.

src/internal/observable/fromEvent.ts
demensky: @jakovljevic-mladen something like this?

```diff
import { fromEvent } from 'rxjs';
+
+ const div = document.createElement('div');
+ div.style.cssText = 'width: 200px; height: 200px; background: #09c;';
+ document.body.appendChild(div);

// note optional configuration parameter which will be passed to addEventListener
const clicksInDocument = fromEvent(document, 'click', { capture: true });
- const clicksInDiv = fromEvent(someDivInDocument, 'click');
+ const clicksInDiv = fromEvent(div, 'click');

clicksInDocument.subscribe(() => console.log('document'));
clicksInDiv.subscribe(() => console.log('div'));
```

jakovljevic-mladen: Yes.

demensky: Done.

jakovljevic-mladen:
@demensky, can you please update PR by changing target to `master` branch? After merge, I usually manually cherry-pick to `7.x`, I don't do that the other way around.

jakovljevic-mladen (approved):
LGTM. Thanks for changing target to the `master`.

PR #7248, opened 2023-04-11
feat(repeat): add lastValue as second param when delay is a function

AisonSu:
**Description:**
In some cases, we need to use the repeat operator in RxJS to continuously retrieve values such as tokens. However, sometimes the retrieval interval needs to be determined based on the expiration time in the last retrieved value. With this change, we have added support for passing the last retrieved value as the second argument to the delay option when it is a function. This makes it easier to handle such scenarios.


kwonoj:
Hi, we do not expand core api surface without collecting usecases & discussions. Please attempt to create userland operator  for the custom behaviors, and if it's being widely used please create a RFC to disucss if it needs to land in core library or not.

PR #7249, opened 2023-04-15
Remove operate

benlesh:
refactor: Remove `operate` function
    
+ Inlines all operator function creation as just arrow functions. After `lift` was removed, there is no use for the `operate` function.
    
Related #7202
    
refactor(find): reduce complexity in function calls
    
+  Removes use of higher-order function, as it's no longer necessary, because we're not using `operate`, because we don't even `lift`.
    
refactor(scan/reduce): reduce complexity in function calls
    
+  Removes use of higher-order function, as it's no longer necessary, because we're not using `operate`, because we don't even `lift`.

### Reviewer Note: Best reviewed per commit.


src/internal/operators/audit.ts
```diff
@@ -1,7 +1,7 @@
 import { Subscriber } from '../Subscriber';
 import { MonoTypeOperatorFunction, ObservableInput } from '../types';
 
-import { operate } from '../util/lift';
```
benlesh, line 4: The `operate` function is being removed everywhere.

Look for a conversion from:  `operate(fn)` to `(source) => new Observable(subscriber => fn(source, subscriber))` (basically)

src/internal/operators/audit.ts
```diff
@@ -51,46 +51,47 @@ import { createOperatorSubscriber } from './OperatorSubscriber';
  * emissions from the source Observable.
  */
 export function audit<T>(durationSelector: (value: T) => ObservableInput<any>): MonoTypeOperatorFunction<T> {
-  return operate((source, subscriber) => {
-    let hasValue = false;
-    let lastValue: T | null = null;
-    let durationSubscriber: Subscriber<any> | null = null;
-    let isComplete = false;
+  return (source) =>
+    new Observable((subscriber) => {
```
benlesh, line 55: Similar to here.

src/internal/operators/share.ts
```diff
@@ -176,7 +176,7 @@ export function share<T>(options: ShareConfig<T> = {}): MonoTypeOperatorFunction
       conn?.unsubscribe();
     };
 
-    return operate<T, T>((source, subscriber) => {
```
benlesh, line 179: Here before, we needed to make sure `lift` composed types, so we used a higher-order function. That is no longer necessary, so we can just inline things now.

That's why this one looks a little different.

src/internal/operators/scanInternals.ts
```diff
@@ -16,47 +16,47 @@ export function scanInternals<V, A, S>(
   seed: S,
   hasSeed: boolean,
   emitOnNext: boolean,
-  emitBeforeComplete?: undefined | true
+  emitBeforeComplete: undefined | true,
+  source: Observable<V>,
+  subscriber: Subscriber<any>
```
benlesh, line 21: Higher order function no longer needed here. Affects `reduce` and `scan` a little.

src/internal/operators/find.ts
```diff
@@ -51,28 +50,31 @@ export function find<T>(predicate: (value: T, index: number, source: Observable<
...
-  return operate(createFind(predicate, 'value'));
+  return (source) => new Observable((subscriber) => createFind(predicate, 'value', source, subscriber));
 }
 
-export function createFind<T>(predicate: (value: T, index: number, source: Observable<T>) => boolean, emit: 'value' | 'index') {
+export function createFind<T>(
+  predicate: (value: T, index: number, source: Observable<T>) => boolean,
+  emit: 'value' | 'index',
+  source: Observable<T>,
+  subscriber: Subscriber<any>
```
benlesh, line 60: Higher order function no longer needed here. Affects `reduce` and `scan` a little.

src/internal/operators/share.ts
```diff
@@ -237,7 +237,7 @@ export function share<T>(options: ShareConfig<T> = {}): MonoTypeOperatorFunction
         });
         from(source).subscribe(connection);
       }
-    })(wrapperSource);
```
jakovljevic-mladen, line 240: I think there's a test for this, but this is reminder to double check.

jakovljevic-mladen (approved):
🎉 LGTM, only minor issues related to the leftover comments that I'd like to be fixed.

This will finally make the code look like what we suggest to people when we instruct them how to build custom operators.

demensky (approved):

src/internal/operators/scanInternals.ts
```diff
@@ -16,47 +16,47 @@ export function scanInternals<V, A, S>(
   seed: S,
   hasSeed: boolean,
   emitOnNext: boolean,
-  emitBeforeComplete?: undefined | true
+  emitBeforeComplete: undefined | true,
```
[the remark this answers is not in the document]
benlesh: Seems reasonable.

src/internal/operators/share.ts
```diff
@@ -149,7 +149,7 @@ export function share<T>(options: ShareConfig<T> = {}): MonoTypeOperatorFunction
   // _operator function_ is called when the complete pipeline is composed via a
   // call to a source observable's `pipe` method - not when the static `pipe`
   // function is called.
-  return (wrapperSource) => {
+  return (source) => {
```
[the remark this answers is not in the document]
benlesh: Honestly. I'm just going to remove the comment. It existed to comment on a location where we resolved what should have been an obvious bug.

PR #7252, opened 2023-04-17
Fix 6805: remove strange groupBy behavior

### LINKED ISSUE #6805: Should `groupBy` keep connections after the result subscription is unsubscribed?
We currently have [special case handling](https://github.com/ReactiveX/rxjs/blob/123a0f256e0022e87dfa249c895670ee745def53/src/internal/operators/groupBy.ts#L273-L290) for `groupBy` that purposefully introduces the following seemingly non-deterministic behavior:

[Stackblitz](https://stackblitz.com/edit/rxjs-d2gj5u?file=index.ts)

```ts
import { tap, interval, groupBy } from 'rxjs';

const subscription = interval(100)
  .pipe(
    groupBy((n) => (n % 2 === 0 ? 'even' : 'odd')),
    tap((grouped$) => {
      grouped$.subscribe((n) => console.log(grouped$.key, n));
    })
  )
  .subscribe();

// Toggle to true or false to try different scenarios.
const unsubImmediately = true;

// Unsubscribe here and it stops! (no groups have been recieved yet).
// Result subscription was unsubscribed before any groups could be subscribed to.
if (unsubImmediately) {
  subscription.unsubscribe();
} else {
  setTimeout(() => {
    // Here the source subscription goes on forever!!
    // We have 2 groups, and someone's subscribed to them in a way they're not unsubscribed
    // when the result subscription is.
    subscription.unsubscribe();
  }, 1000);
}
```

Basically, if anything subscribes to a grouped observable and doesn't unsubscribe, unsubscribing from the outer subscription will not unsubscribe from the source. If there are no subscriptions to the grouped observables, then the subscription to the source will be unsubscribed.

## What is the issue?

This is what `groupBy` has been doing since long before I came onto the project. However, my gut tells me this behavior is inconsistent with the rest of the framework, saving the default behavior of `shareReplay()`, which has itself caused a lot of confusion. The behavior, as demonstrated by the Stackblitz above, can be hard to reason about if event sources or subscription management isn't deterministic.

This behavior is such a one-off thing it's wildly unlike any other operator we export.

## Proposed change

I feel that outer subscriptions to the resulting observable returned by `groupBy` should unsubscribe/disconnect the source from the groups when unsubscribed. In this case, the grouped observables, if still subscribed, could either: A) unsubscribe on their own, or B) emit an error saying they were disconnected. I feel like `A` is probably the proper solution, but we could even make that configurable.

(This would be a BREAKING change for `groupBy`, in the context that grouped observables would shut down if the source is disconnected)

benlesh:
fix(groupBy): will no longer leak inner subscriptions
    
`groupBy` no longer supports the behavior where inner subscriptions will cause the outer subscription to stay connected after consumer unsubscribes from result.
    
Resolves #6805
    
BREAKING CHANGE: `groupBy` no longer allows grouped observable subscriptions to stay connected after parent subscription unsubscribed. If you need this behavior, don't unsubscribe from the parent.
    
refactor: Remove direct instantiation of `OperatorSubscriber`
    
+ Stopped exporting `OperatorSubscriber` class from module
+ Updated `onErrorResumeNext` and a `Subject` test to use `createOperatorSubscriber`

Actual stream of the work is here, if that helps: https://youtu.be/rHgED1U6XDY

### DIFF
--- a/src/internal/operators/groupBy.ts
+++ b/src/internal/operators/groupBy.ts
@@ -2,7 +2,7 @@ import { Observable } from '../Observable';
 import { from } from '../observable/from';
 import { Subject } from '../Subject';
 import { ObservableInput, Observer, OperatorFunction, SubjectLike } from '../types';
-import { createOperatorSubscriber, OperatorSubscriber } from './OperatorSubscriber';
+import { createOperatorSubscriber } from './OperatorSubscriber';
 
 export interface BasicGroupByOptions<K, T> {
   element?: undefined;
@@ -165,12 +165,6 @@ export function groupBy<T, K, R>(
       // next call from the source.
       const handleError = (err: any) => notify((consumer) => consumer.error(err));
 
-      // The number of actively subscribed groups
-      let activeGroups = 0;
-
-      // Whether or not teardown was attempted on this subscription.
-      let teardownAttempted = false;
-
       // Capturing a reference to this, because we need a handle to it
       // in `createGroupedObservable` below. This is what we use to
       // subscribe to our source observable. This sometimes needs to be unsubscribed
@@ -178,7 +172,7 @@ export function groupBy<T, K, R>(
       // in cases where a user unsubscribes from the main resulting subscription, but
       // still has groups from this subscription subscribed and would expect values from it
       // Consider:  `source.pipe(groupBy(fn), take(2))`.
-      const groupBySourceSubscriber = new OperatorSubscriber(
+      const groupBySourceSubscriber = createOperatorSubscriber(
         subscriber,
         (value: T) => {
           // Because we have to notify all groups of any errors that occur in here,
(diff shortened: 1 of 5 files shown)


jakovljevic-mladen (approved):
LGTM, only one comment from the code needs to be updated.

src/internal/operators/groupBy.ts
```diff
@@ -165,20 +165,14 @@ export function groupBy<T, K, R>(
...
       // subscribe to our source observable. This sometimes needs to be unsubscribed
       // out-of-band with our `subscriber` which is the downstream subscriber, or destination,
       // in cases where a user unsubscribes from the main resulting subscription, but
       // still has groups from this subscription subscribed and would expect values from it
       // Consider:  `source.pipe(groupBy(fn), take(2))`.
-      const groupBySourceSubscriber = new OperatorSubscriber(
+      const groupBySourceSubscriber = createOperatorSubscriber(
         subscriber,
         (value: T) => {
           // Because we have to notify all groups of any errors that occur in here,
```
ktrz, line 178: GH didn't allow me to put a comment in the exact spot lines 189-191 [here](https://github.com/ReactiveX/rxjs/pull/7252/files#diff-454b92b7ec8b4a19e5475e7e52f58f37ad6230c46835eeb02f3cbee0e32c0ac6R189-R191) but I think the comment also needs revisiting as the mentioned reference counting seem to be removed now.

ktrz (commented):
I've added one additional note about a comment that probably needs revisiting. Other than that LGTM

PR #7258, opened 2023-05-07
chore(test): remove global test scheduler 

tmair:
**Description:**
This PR removes the global `rxTestScheduler` from the test infrastructure. To get this working it was necessary to also remove the tests for the global test scheduler.

In addition the custom mocha UI was also removed because it is no loger necessary.


benlesh (approved):

benlesh:
This is great, @tmair ... it looks like there's a conflict though.

spec/schedulers/TestScheduler-spec.ts
```diff
@@ -1,15 +1,14 @@
+/** @prettier */
```
jakovljevic-mladen, line 1: I hope this is the last PR where we have this line included. Can we update `.prettierrc.json` and completely remove `"overrides"` filed from it? 🤔

benlesh:
Thank you so much, @tmair !

spec/schedulers/TestScheduler-spec.ts
tmair: @jakovljevic-mladen I created a draft PR that removes the pragma and the `spec` directory from the prettier formatting override. Please have a look at #7275. If we would like to merge those changes I can update the PR and promote it to a proper PR.

jakovljevic-mladen: > If we would like to merge those changes I can update the PR and promote it to a proper PR.

I think we do. Great @tmair, thanks a bunch!

PR #7261, opened 2023-05-11
Version 8: Build changes - ES2022, no more tslib, etc. etc.

benlesh:
Please review by commit, there are several changes and cleanups made.

Highlights:

+ No more dependency on `tslib`!!
+ Outputting ES2022
+ No longer publishing UMD bundles
+ No longer publishing ES5
+ No longer publishing extraneous `es2015` folder
+ Updates package.json
+ removes a bunch of unused dependencies
+ removes `check-side-effects`!

NOTE: `check-side-effects` is yet another unmaintained side project (I was badgered into using) we depended on from the Angular team. It broke when we updated to output ES2022, because it couldn't handled things like `?.`. There are no more updates for us to pull from them, we were on the latest version.


package.json
```diff
@@ -19,50 +18,43 @@
       "types": "./dist/types/index.d.ts",
       "node": "./dist/cjs/index.js",
       "require": "./dist/cjs/index.js",
-      "es2015": "./dist/esm/index.js",
-      "default": "./dist/esm5/index.js"
+      "default": "./dist/esm/index.js"
```
demensky, line 21: There is no `"type": "module"` in `package.json`, so the file specified in `"default"` will be parsed as CommonJS, right? But there is a file with ES modules.

```json
".": {
  "types": "./dist/types/index.d.ts",
  "node": "./dist/cjs/index.js",
  "import": "./dist/esm/index.js",
  "default": "./dist/cjs/index.js"
},
```

demensky: Also, maybe you should move [`"node"`](https://nodejs.org/api/packages.html#conditional-exports) down or remove it altogether?

> `"node"` - matches for any Node.js environment. Can be a CommonJS or ES module file. _In most cases explicitly calling out the Node.js platform is not necessary._

```js
import {timer} from 'rxjs'; // version from current PR

timer(2000).subscribe(() => {
  console.trace('2000');
});
```

```
% node main.js

Trace: 2000
    at Object.next (file:///example/main.js:4:11)
    at ConsumerObserver.next (example/node_modules/rxjs/dist/cjs/internal/Subscriber.js:99:33)
    at Subscriber._next (example/node_modules/rxjs/dist/cjs/internal/Subscriber.js:70:26)
    at Subscriber.next (example/node_modules/rxjs/dist/cjs/internal/Subscriber.js:41:18)
    at AsyncAction.work (example/node_modules/rxjs/dist/cjs/internal/observable/timer.js:28:28)
    at AsyncAction._execute (example/node_modules/rxjs/dist/cjs/internal/scheduler/AsyncAction.js:79:18)
    at AsyncAction.execute (example/node_modules/rxjs/dist/cjs/internal/scheduler/AsyncAction.js:67:26)
    at AsyncScheduler.flush (example/node_modules/rxjs/dist/cjs/internal/scheduler/AsyncScheduler.js:38:33)
    at listOnTimeout (node:internal/timers:573:17)
    at process.processTimers (node:internal/timers:514:7)
```

NodeJS imports the CommonJS (cjs) version instead of the ES version.

.github/workflows/ci_main.yml
```diff
@@ -33,7 +33,6 @@ jobs:
       - name: test
         run: |
           npm test
```
demensky, line 35:
```suggestion
        run: npm test
```

LcsGa (approved):

package.json
benlesh: Honestly, I didn't change anything that was already working. It's still working, this is the minimal change to get the right JS target, I don't want to change it unless it's breaking someone.

demensky: Agree. I will try to do a thorough research and make a separate issue for this.

demensky (approved):

PR #7264, opened 2023-05-16
Adding operate

benlesh:
This is the execution of the design from #7253.

**Please review per commit**.

The work here is broken into commits to help understand the changes.

The key bits of work are:

1. `new Subscriber` now accepts an `overrides` argument as the second parameter. This argument contains the same set of arguments we used to pass to `createOperatorSubscriber`. In short, passing in `next` will "override" `_next`, `error` overrides `_error`, etc.
3. `Subscriber` constructor has been re-deprecated (it's currently deprecated in 7.x). There's no reason anyone should be creating a subscriber instance with `new Subscriber`, really.
4. The new `Subscriber` constructer with 2 arguments is marked `@internal`, and is not a public API.
5. A publicly exposed function called `operate` has been added. This basically wraps the new `Subscriber` constructor instance.
6. Operators were updated, one-by-one, in separate commits. **There are no functional changes to these!!**, all that was done was moving from `createOperatorSubscriber(destination, next, complete, error, finalize)` to `operate({ destination, next, error, complete, finalize })`. That's it.
7. Only one test needed to be updated, and it was testing the `createOperatorSubscriber` as a means of creating operator-related `Subscriber` instances and subscribing to a `Subject` with them.

If it helps reviewers, I live-streamed the bulk of the work here: https://youtu.be/YHyt3fHIEAk

- Updating `Subscriber` to accept the overrides work and explanation: https://youtu.be/YHyt3fHIEAk?t=475
- Adding `operate` function: https://youtu.be/YHyt3fHIEAk?t=1304
- Updating `Subscriber` signature type overloads: https://youtu.be/YHyt3fHIEAk?t=1491
- Adding unit tests for `operate`: https://youtu.be/YHyt3fHIEAk?t=1718
- Deprecating and internalizing `Subscriber` signatures (with dtslint testing): https://youtu.be/YHyt3fHIEAk?t=3420
- Example: updating `catchError`: https://youtu.be/YHyt3fHIEAk?t=4210
- Example: updating `takeLast`: https://youtu.be/YHyt3fHIEAk?t=4454
- This is where I realized I needed to make sure the `finalize` override/handler is fired _after_ the subscriber cleans up: https://youtu.be/YHyt3fHIEAk?t=6368


LcsGa (approved):

spec/Subscriber-spec.ts
```diff
@@ -310,4 +311,131 @@ describe('Subscriber', () => {
...
+
+    const subscriber = operate({
+      destination,
+      complete() {
+        throw 'boop!'
+      }
+    });
+
+    subscriber.complete();
+    expect(complete).not.to.have.been.called;
```
LeonardoBraga, line 424: `calledOnce`?

benlesh: Nah. I don't expect it to be called at all.

src/internal/Subscriber.ts
```diff
@@ -172,3 +249,31 @@ function handleStoppedNotification(notification: ObservableNotification<any>, su
 function hasAddAndUnsubscribe(value: any): value is Subscription {
   return value && isFunction(value.unsubscribe) && isFunction(value.add);
 }
+
+export interface OperateConfig<In, Out> extends SubscriberOverrides<In> {
```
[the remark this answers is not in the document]
benlesh: A good idea.

PR #7269, opened 2023-05-24
Create Walit

Walithack:
IMG_20230524_132850_530


benlesh:
I fixed your comment. Nice picture. I'm not sure what this PR is, but if you're practicing something, it would be nicer if you did it in your own repository. You can do that by clicking on your username, going under "repositories" and there should be a button to create one. Otherwise, I'm sure there's a lot of YouTube videos on how to do it.

Best of luck do you!

PR #7273, opened 2023-05-26
Create SECURITY.md

joycebrum:
Closes #7265 

I've created the SECURITY.md file considering the [report vulnerability through security advisory](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability), which is a new github feature.

If you're interested in the GitHub's feature, it must be [activated for the repository](https://docs.github.com/en/code-security/security-advisories/repository-security-advisories/configuring-private-vulnerability-reporting-for-a-repository):
1. Open the repo's settings
2. Click on [Code security & analysis](https://github.com/ReactiveX/rxjs/settings/security_analysis)
3. Click "Enable" for "Private vulnerability reporting (Beta)"

If you rather not enable it there is also the possibility to receive the vulnerability report through an email, in this case just let me know which email it would be and I'll submit the change.

Besides that, feel free to edit or suggest any changes to this document, it is supposed to reflect the amount of effort the team can offer to handle vulnerabilities.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -10,4 +10,4 @@ If you have discovered a security vulnerability in this project, please report i
 
 Please disclose it at [security advisory](https://github.com/ReactiveX/rxjs/security/advisories/new).
 
-Although we will be working to solve any security issue as fast as possible, it is also important to notice that, in accordance with Apache 2.0 terms, no rxjs contributor can be liable for damages, including the ones caused by a security issue.
+Although we will be working to solve any security issue as fast as possible, it is also important to notice that, in accordance with Apache 2.0 terms, no RxJS contributor can be liable for damages, including the ones caused by a security issue.


joycebrum:
Hi, just pinging to know how is going this PR Review. Would I need to address any changes to this policy to be a better suit for rxjs?

benlesh (commented):
In general, I think this is a good idea. However, it seems like it should mention something about the license agreement, and the waiver of liability there in.

We, of course, want to fix any security issues as rapidly as possible. And a library like this is actually pretty unlikely to get security issues.  But also as a volunteer, I don't really feel like being sued personally for damages that people feel were caused by not responding to security issue within a certain time period. Especially if there's a document that we approved that seems to be some sort of agreement on our part.

joycebrum:
Hi @benlesh thanks for the return. I've rephrased the policy to not explicitly mention a deadline for public disclosure (instead asking reporters to align with you before disclosing anything about the security issue). 

Besides, I've added a paragraph to remind users about the Apache Liability aspect. Let me know if you think this would be better in the beginning of the doc.

benlesh:
Thanks, @joycebrum .. I think this is good. I am only concerned at this point, because I'm not entirely sure I'll notice the security advisories, as I'm not conditioned to check them, nor do I know quite how they work.

PR #7279, opened 2023-06-03
type: fix `ignoreElements` generic type

caikan:
**Description:**
If `ignoreElements` follows `tap` immediately, the type of the `tap` parameter will be incorrectly inferred to be `unknown`.
```ts
import { ignoreElements, of, tap } from 'rxjs';

function test() {
  const data = { value: 1 }
  of(data).pipe(
    tap(data => console.log(data.value)),
    // ---------------------^~~~ ['data' is of type 'unknown'.ts(18046)]
    ignoreElements(),
  )
}
```

### DIFF
--- a/src/internal/operators/ignoreElements.ts
+++ b/src/internal/operators/ignoreElements.ts
@@ -38,7 +38,7 @@ import { noop } from '../util/noop';
  * `complete` or `error`, based on which one is called by the source
  * Observable.
  */
-export function ignoreElements(): OperatorFunction<unknown, never> {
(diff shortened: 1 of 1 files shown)


kwonoj:
Isn't this exactly same as https://github.com/ReactiveX/rxjs/pull/7204#issuecomment-1555260495 ?

LcsGa (approved):

benlesh:
I think this was already done in another commit.

Thank you so much for the PR, though, @caikan!!

PR #7281, opened 2023-06-05
fix(mergeWith): works correctly with an Array

### LINKED ISSUE #7278: error when `mergeWith` argument is an array
### Describe the bug

`mergeWith` can accept an `ObservableInput` as an argument, but when the argument is an array (it is a legal ObservableInput), an error will be reported:
`You provided 'xxx' where a stream was expected. You can provide an Observable, Promise, ReadableStream, Array, AsyncIterable, or Iterable.`

### Expected behavior

Treat the array as a normal `ObservableInput`

### Reproduction code

```typescript
import { EMPTY, mergeWith } from 'rxjs';

EMPTY.pipe(mergeWith([1])).subscribe(console.log);
```

### Reproduction URL

https://stackblitz.com/edit/rxjs-aixohl?file=index.ts

### Version

7.8

### Environment

_No response_

### Additional context

_No response_

josepot:
**Description:**

Fixes #7278 

I've also ran into this issue in the past.

#7278

### DIFF
--- a/src/internal/operators/merge.ts
+++ b/src/internal/operators/merge.ts
@@ -1,6 +1,5 @@
 import { ObservableInput, ObservableInputTuple, OperatorFunction, SchedulerLike } from '../types';
 import { operate } from '../util/lift';
-import { argsOrArgArray } from '../util/argsOrArgArray';
 import { mergeAll } from './mergeAll';
 import { popNumber, popScheduler } from '../util/args';
 import { from } from '../observable/from';
@@ -23,7 +22,6 @@ export function merge<T, A extends readonly unknown[]>(
 export function merge<T>(...args: unknown[]): OperatorFunction<T, unknown> {
   const scheduler = popScheduler(args);
   const concurrent = popNumber(args, Infinity);
-  args = argsOrArgArray(args);
 
   return operate((source, subscriber) => {
     mergeAll(concurrent)(from([source, ...(args as ObservableInput<T>[])], scheduler)).subscribe(subscriber);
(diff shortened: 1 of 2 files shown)


src/internal/operators/merge.ts
```diff
@@ -23,7 +22,6 @@ export function merge<T, A extends readonly unknown[]>(
 export function merge<T>(...args: unknown[]): OperatorFunction<T, unknown> {
   const scheduler = popScheduler(args);
   const concurrent = popNumber(args, Infinity);
-  args = argsOrArgArray(args);
```
josepot, line 26: This was clearly a mistake, because these `args` are being spread inside the from afterwords, so we always want the array, even when it only has one value.

LcsGa (approved):

src/internal/operators/merge.ts
benlesh: The typings on line 9 show that we were trying to handle the case of `mergeWith([a$, b$, c$])`...

I'm on the fence here. There are two issues at play:

1. `a$.pipe(mergeWith(b$, c$))` is more ergonomic
2. The `concurrency` limit should be provided via a configuration object at the end at some point, which means that `mergeWith([...], { concurrency: 2 })` is a LOT cleaner to implement, but `mergeWith(..., { concurrency: 2 })` is fine as well just obnoxious to implement.

Also, it's technically a breaking change from intent, but there's clearly been a bug that didn't always allow the intent.... and no one has complained. 🤔

benlesh: You're right... I was just getting back to this, and even dating back to RxJS 5, we didn't support `[a$, b$, c$]` as an argument. I'm probably just thinking of `combineLatest` or something else.

benlesh (approved):

PR #7286, opened 2023-06-14
Reverse if condition to remove nested code in Subscription.ts

thyecust:
**Description:** As a suggestion, we can remove nested code to make it clearer and easy to read.

**Related issue (if exists):** No.


kwonoj:
Sorry, this doesn't make a meaningful changes to the codebase and readibility various across people, while core team agress with current code style. Appreciate for the effort but we won't likely land this.

thyecust:
Thx for your reply

Issue #7280, opened 2023-06-04
defaultIfEmpty does not give correct type for null or undefined when strictNullChecks is false in tsconfig


shessafridi:
### Describe the bug

`defaultIfEmpty(null)` or `defaultIfEmpty(undefined)` always returns the observable of type `Observable<any>`

### Expected behavior

When strictNullChecks is off null type is ignored so it should just return the type of the source observable as it is.

Expected behaviour in `strictNullChecks: false`

`const test$ = of(1).pipe(defaultIfEmpty(null)); // Observable<number>`

Instead we get

`const test$ = of(1).pipe(defaultIfEmpty(null)); // Observable<any>`

### Reproduction code

```typescript
const test = of(1).pipe(defaultIfEmpty(null));
```

### Version

7.8.1

### Environment

"typescript": "~5.0.4",
 "rxjs": "^7.8.1"

### Additional context

I am currently using a custom wrapper operator built over `defaultIfEmpty` to fix the issue, here is the code for it.

```
import { defaultIfEmpty, Observable } from 'rxjs';

export function defaultIfEmptyFixed<T, R extends null>(
  defaultValue: R
): (source: Observable<T>) => Observable<T | null>;

export function defaultIfEmptyFixed<T, R extends undefined>(
  defaultValue: R
): (source: Observable<T>) => Observable<T | undefined>;

export function defaultIfEmptyFixed<T, R>(
  defaultValue: R
): (source: Observable<T>) => Observable<T | R> {
  return (input: Observable<T>) =>
    input.pipe(defaultIfEmpty<T, R>(defaultValue));
}
```

benlesh:
Hi @shessafridi — thank you for taking the time to open this issue and for sharing your thoughtful feedback with the RxJS project. I’m sorry it took us so long to respond. Given how much time has passed and the lack of recent activity, I’m going to close this for now so the tracker reflects the work we’re likely to act on. If this is still relevant on a current RxJS release, please feel welcome to open a new issue with updated context and a reproduction. Thank you again for your contribution and patience.

— Ben

PR #7287, opened 2023-06-19
Fixed defaultIfEmpty incorrect type with strictNullChecks off

shessafridi:
**Description:**

`defaultIfEmpty(null)` or `defaultIfEmpty(undefined)` always returns the observable of type `Observable<any>`

### Expected behavior

When strictNullChecks is off null type is ignored so it should just return the type of the source observable as it is.

Expected behaviour in `strictNullChecks: false`

`const test$ = of(1).pipe(defaultIfEmpty(null)); // Observable<number>`

Instead we get

`const test$ = of(1).pipe(defaultIfEmpty(null)); // Observable<any>`

This PR fixes the issue by adding explicit overloads for null and undefined

**Related issue (if exists):** #7280

### DIFF
--- a/src/internal/operators/defaultIfEmpty.ts
+++ b/src/internal/operators/defaultIfEmpty.ts
@@ -37,6 +37,19 @@ import { operate } from '../Subscriber';
  * specified `defaultValue` if the source Observable emits no items, or the
  * values emitted by the source Observable.
  */
+
+
+// To make things easier for (naughty) devs who use the
+// `strictNullChecks: false` TypeScript compiler option, these
+// overloads with explicit null and undefined values are included.
+
+export function defaultIfEmpty<T, R extends null>(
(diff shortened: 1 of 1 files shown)


shessafridi:
ping @benlesh

shessafridi:
ping @jakovljevic-mladen

conblem:
What happenened to this pr?

benlesh:
Honestly, I remember glancing at it, and thinking about it, but deciding (at the time) that it doesn't seem like something we want to support. If people want to turn off strict settings in TypeScript, that's their business. I left it here because I wasn't totally convinced I was correct, but every time I revisited it, I had the same thought. That it wasn't the right thing to do.

Sorry @shessafridi I should have more openly written out my thoughts. Thank you for your contribution.

PR #7172, opened 2023-01-25
fix(subscriber): strict type signature for next method

raveclassic:
**Description:**

This PR fixes type signature of `Subscriber.prototype.next` method that was previously taking optional argument which was violating generic constraint `T`.

**BREAKING CHANGE:** `Subscriber.prototype.next` now takes strict argument.
```typescript
new Observable(subscriber => {
  subscriber.next() // throws in compile time (expected, caused by this PR)
})

new Observable<string>(subscriber => {
  subscriber.next() // throws in compile time (expected, caused by this PR)
})

new Observable<void>(subscriber => {
  subscriber.next() // ok
})
```

**Related issue (if exists):** https://github.com/ReactiveX/rxjs/issues/2852

**Additional changes:**
- update package version in lockfile to `7.8.0` (running `npm i`)

PR #7290, opened 2023-06-19
fix(Subscriber): next method no longer has optional value argument

### LINKED ISSUE #2852: Subject.prototype.next() should only accept T, not T | undefined

benlesh:
Resolves #2852
Related #7172


jakovljevic-mladen (approved):

josepot:
I just faced a nasty a runtime error that would have been prevented if this fix had been shipped. So, my question is: will a patch be release with this fix? or this is going to be considered a "breaking change" and we will have to wait for version 8? 🙏

FWIW: I really think that this should be a patch.

PR #7297, opened 2023-06-28
[Idea] Make it possible for BehaviorSubject to emit an initial "summary" value to late subscribers

vkryachko:
**Description:**

**Note: this PR is intended to open a discussion on whether this change is considered broadly useful**

BehaviorSubject currently allows emitting the initial or latest value when it's subscribed to. However there is a use case that it non-trivial to model with it.

Imagine a BehaviorSubject is used to stream log events and consumers can subscribe and render them. However the logs subscribers will receive depends on when they subscribe, late subscribers won't see log events that happened before they subscribed.

This change proposes to make it possible for late subscribers to receive a "summary" value upon initial subscription, e.g.

```typescript
interface LogEvent {
  readonly log: string;
}

const events = new BehaviorSubject<LogEvent>(
    {log: "start"},
    (prev, next) => {log: `${prev.log}\n${next.log}`.slice(-2000)});
```

With this code, late subscribers will receive up to 2000 characters of previously emitted logs and then keep receiving new log lines as long as they are subscribed.
While this example it contrived, the proposed change is general enough to allow for more interesting folds to generate a meaningful "summary" value.

Thoughts?

**BREAKING CHANGE:** NA

### DIFF
--- a/src/internal/BehaviorSubject.ts
+++ b/src/internal/BehaviorSubject.ts
@@ -31,6 +31,7 @@ export class BehaviorSubject<T> extends Subject<T> {
   }
 
   next(value: T): void {
-    super.next((this._value = this._valueFold(this._value, value)));
+    this._value = this._valueFold(this._value, value);
+    super.next(value);
   }
 }
--- a/src/internal/BehaviorSubject.ts
+++ b/src/internal/BehaviorSubject.ts
@@ -7,7 +7,7 @@ import { Subscription } from './Subscription';
  * value whenever it is subscribed to.
  */
 export class BehaviorSubject<T> extends Subject<T> {
-  constructor(private _value: T) {
+  constructor(private _value: T, private readonly _valueFold: (currentValue: T, newValue: T) => T = (_, v) => v) {
     super();
   }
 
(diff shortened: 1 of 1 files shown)


voliva:
If you don't need to "summarise" and only need to keep the last 2000 logs, you can just use `new ReplaySubject(2000)` instead of `new BehaviourSubject()`.

But if for whatever reason you want to summarise it, then I think you can just break it in a separate observable:

```ts
const events = new Subject<LogEvent>(); // BehaviourSubject not needed

const eventSummary = connectable(events.pipe(
  scan((prev, next) => {log: `${prev.log}\n${next.log}`.slice(-2000)}, {log: "start"}),
), {
  connector: () => new ReplaySubject(1)
});
eventSummary.connect()
```

In a way, if I subscribe to `events` I expect not to receive the last log event, that's why I think `Subject` is usually more suitable.
And if I need the summary I can subscribe to `eventSummary`, or even combine them both so I can receive the summary followed by the new ones.

benlesh:
Hi @vkryachko — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

PR #7302, opened 2023-07-10
update observable docs

RavenColEvol:
Reduce complexity. There is no need to explain what consumer will do as it is self explanatory it only makes it harder to understand by adding unneccesary data.

**Description:**

**BREAKING CHANGE:** 


kwonoj:
I believe PR is redundant, doc stands for the new users and the current one is helpful for the user who want to understand concepts.

PR #7321, opened 2023-08-18
Refactor: no internal pipe method use

benlesh:
Removes internal usage of the `pipe` method, so that in the future, other observables that do not implement `pipe` will work with our operators.

This PR doesn't have the explicit goal of ensuring operators work with "observable shaped things", but it's a step in that direction.


benlesh:
Under current Core team agreement, I'm merging this without additional review, as it's been up a long time and is landing in an alpha version.

PR #7322, opened 2023-08-18
refactor: isolate reliance on SchedulerAction `this` context

benlesh:
This goes through and moves all of the reliance on the `this` context for scheduling to a single point, `executeSchedule`. It also moves common complexities to that helper function. All of this is done in preparation for creating newer, simplified schedulers


benlesh:
Under current Core team agreement, I'm merging this without additional review, as it's been up a long time and is landing in an alpha version.

PR #7325, opened 2023-08-19
fix(TestScheduler): remove applyMixins use

benlesh:
In some environments, the `applyMixins` function doesn't work as expected when used with modern JavaScript classes. This was causing issues when code was trying to access `this.logSubscribedFrame` and `this.logUnsubscribedFrame`.

This didn't appear until we switched over to ES2022 in CJS exports, and it doesn't appear to be an issue in all runtimes. In any case, simplifying the code solved the issue. Tested against some live tests in another repo, I don't have a solid way to duplicate it here. It's sort of a fluke/one-off.


src/internal/testing/ColdObservable.ts
```diff
@@ -1,35 +1,30 @@
...
-  scheduler: Scheduler;
-  // @ts-ignore: Property has no initializer and is not definitely assigned
-  logSubscribedFrame: () => number;
-  // @ts-ignore: Property has no initializer and is not definitely assigned
-  logUnsubscribedFrame: (index: number) => void;
+  logSubscribedFrame = logSubscribedFrame;
+  logUnsubscribedFrame = logUnsubscribedFrame;
 
-  constructor(public messages: TestMessage[], scheduler: Scheduler) {
-    super(function (this: Observable<T>, subscriber: Subscriber<any>) {
```
benlesh, line 20: Moved this super argument out to just be declared as a method. It's just less complicated.

src/internal/testing/ColdObservable.ts
```diff
@@ -1,35 +1,30 @@
...
 
-export class ColdObservable<T> extends Observable<T> implements SubscriptionLoggable {
+export class ColdObservable<T> extends Observable<T> {
   public subscriptions: SubscriptionLog[] = [];
-  scheduler: Scheduler;
-  // @ts-ignore: Property has no initializer and is not definitely assigned
-  logSubscribedFrame: () => number;
-  // @ts-ignore: Property has no initializer and is not definitely assigned
-  logUnsubscribedFrame: (index: number) => void;
+  logSubscribedFrame = logSubscribedFrame;
```
benlesh, line 11: Instead of mixins I'm just adding the methods as function properties on the classes.

src/internal/testing/ColdObservable.ts
```diff
@@ -1,35 +1,30 @@
...
+  logUnsubscribedFrame = logUnsubscribedFrame;
 
-  constructor(public messages: TestMessage[], scheduler: Scheduler) {
-    super(function (this: Observable<T>, subscriber: Subscriber<any>) {
-      const observable: ColdObservable<T> = this as any;
-      const index = observable.logSubscribedFrame();
```
benlesh, line 22: In particular, in some environments with CJS and ES2022, applyMixins was not adding the mixing properties to the prototype, and we were getting errors here.

cartant (approved):

PR #7335, opened 2023-09-11
fix: prevent recursive try-catch memory leak in mergeInternals

### LINKED ISSUE #7334: mergeMap with concurrency and synchronous inner observables leads to a memory leak and outer observable never completes.
### Describe the bug

A memory leak occurs when using mergeMap with a concurrency and a synchronous inner observable, apparently due to the nested try-catch in the doInner function.

Somewhat relates to https://github.com/ReactiveX/rxjs/issues/3609 but isn't the exact same bug. 

In this case, you don't actually get the `RangeError`, and instead receive no feedback at all, it appears the observable just hangs (despite the source and the inner observables all completing). 

You can prevent this error by editing the mergeInternals doInnerSub function to not reference the subscriber in the catch.

https://github.com/ReactiveX/rxjs/blob/7.8.1/src/internal/operators/mergeInternals.ts#L100-L128

You can actually leave the try-catch in place, simply commenting out that subscriber reference is enough to prevent the memory leak.

This suggests that the issue is a memory leak due to the way nodejs is holding on to the reference to subscriber in case there is an error (forgive me for not being to articulate that more clearly :p )

### Expected behavior

No memory leak should occur

### Reproduction code

```typescript
import { EMPTY, from, timer } from 'rxjs';
import { finalize, last, mergeMap } from 'rxjs/operators';

let i = 0;
from(Array(600).keys())
  .pipe(
    mergeMap(
      (n) =>
        (n < 100 ? timer(0) : EMPTY).pipe(
          finalize(() => console.log(++i, 'inner complete')),
        ),
      10,
    ),
    last(),
    finalize(() => console.log('outer complete')),
  )
  .subscribe({
    next: (v) => console.log(),
    error: (err) => console.error(err),
  });
```

You may have to tweak the size of the initial Array to reproduce the error.
```


### Reproduction URL

https://rxjs-playground-test-1kzutu.stackblitz.io

### Version

7.8.1

### Environment

node v18.15.0

### Additional context

_No response_

jonapgar-groupby:
fixes #7334

**Related issue: https://github.com/ReactiveX/rxjs/issues/7334**

### DIFF
--- a/src/internal/operators/mergeInternals.ts
+++ b/src/internal/operators/mergeInternals.ts
@@ -66,9 +66,16 @@ export function mergeInternals<T, R>(
     // This is checked during finalization to see if we should
     // move to the next item in the buffer, if there is on.
     let innerComplete = false;
-
+     
+    let projected;
+    try {
+      projected = project(value, index++);
+    } catch (err) {
+      subscriber.error(err)
+      return
+    }
     // Start our inner subscription.
-    innerFrom(project(value, index++)).subscribe(
+    innerFrom(projected).subscribe(
       createOperatorSubscriber(
         subscriber,
         (innerValue) => {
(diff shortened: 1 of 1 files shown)


benlesh:
Hi @jonapgar-groupby — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

PR #7339, opened 2023-09-13
fix(fromEvent): allow `fromEvent` to handle symbols as event names

### LINKED ISSUE #7338: Bug: Support for Symbol for `fromEvent`
### Describe the bug

Node Event Emitter supports `Symbol` for `on/off` https://nodejs.org/api/events.html#emitteroneventname-listener.
The current implementation of RxJS can use them as provided the example bellow it's just that the types that are used don't support them.

### Expected behavior

Symbols are accepted as types.

### Reproduction code

```typescript
import { EventEmitter } from 'node:events';

const emitter = new EventEmitter();
const protectedEvent = Symbol('Emitted');

// Fails  TypeChecking
const sourceOne = fromEvent(emitter, protectedEvent);
// Passes TypeChecking
const sourceTwo = fromEvent(emitter, protectedEvent as unknown as string);

sourceOne.subscribe((val) => console.log(val));
sourceTwo.subscribe((val) => console.log(val));

emitter.emit(protectedEvent, {
  message: 'Hello World',
});
```

### Reproduction URL

_No response_

### Version

7.8.1

### Environment

Node: 20.3.1 (should work for 16 LTS and above)

### Additional context

This useful so that you can have internal methods not exposed the the user of libary.

demensky:
Allow passing a value of type `symbol` to `eventName` if a `NodeStyleEventEmitter` is passed to `source`, according to [the documentation](https://nodejs.org/api/events.html#emitteraddlistenereventname-listener).

Closes #7338

### DIFF
--- a/packages/rxjs/spec-dtslint/observables/fromEvent-spec.ts
+++ b/packages/rxjs/spec-dtslint/observables/fromEvent-spec.ts
@@ -56,6 +56,13 @@ it('should support a node-style source', () => {
   const b = fromEvent<B>(nodeStyleSource, "exit"); // $ExpectType Observable<B>
 });
 
+it('should support a node-style source and symbol eventName', () => {
+  const SYMBOL_EVENT = Symbol();
+  const source: NodeStyleEventEmitter = nodeStyleSource;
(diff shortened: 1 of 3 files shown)


benlesh (approved):

benlesh:
Thank you, @demensky !

PR #7344, opened 2023-09-20
Async dispatched config handlers

### LINKED ISSUE #7343: onStoppedNotification and onUnhandledError shouldn't be allowed to flip synchronous via provider delegates
We currently allow the TestScheduler (or anyone who wants to fiddle with the delegate for `timeoutProvider`) to allow `config.onUnhandledError` and `config.onStoppedNotification` to synchronously dispatch. The problem with that is if either of those throw while synchronous, it will unwind the stack and cause all sorts of crazy stuff. The provider delegates are meant primarily so that TestScheduler can synchronously control what is going on with observable emissions. `onUnhandledError` and `onStoppedNotification` are firmly outside of that lifecycle and should always be dispatched in their own call stack to prevent issues.

benlesh:
This is removing dependency on `timeoutProvider` in `Observable` because observable depends on `Subscriber`, and `Subscriber` users `timeoutProvider` for both `onUnhandledError` and `onStoppedNotification`, when it doesn't really need to. That's not particularly what `timeoutProvider` is for.

This also preps us to publish `@rxjs/observable` by limiting the dependencies that `Observable` has.

Resolves #7343

### DIFF
--- a/packages/rxjs/spec/operators/share-spec.ts
+++ b/packages/rxjs/spec/operators/share-spec.ts
@@ -18,24 +18,11 @@ import {
 } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
-import { SinonSpy, spy } from 'sinon';
+import { spy } from 'sinon';
 
 const syncNotify = of(1);
 const asapNotify = scheduled(syncNotify, asapScheduler);
 const syncError = throwError(() => new Error());
-
-function spyOnUnhandledError(fn: (spy: SinonSpy) => void): void {
-  const prevOnUnhandledError = config.onUnhandledError;
-
-  try {
-    const onUnhandledError = spy();
-    config.onUnhandledError = onUnhandledError;
-
-    fn(onUnhandledError);
-  } finally {
-    config.onUnhandledError = prevOnUnhandledError;
-  }
-}
 
 /** @test {share} */
 describe('share', () => {
@@ -810,9 +797,22 @@ describe('share', () => {
       });
     });
 
-    it('should not reset on refCount 0 if reset notifier errors before emitting any value', () => {
-      spyOnUnhandledError((onUnhandledError) => {
+    describe('when config.onUnhandledError is set', () => {
+      afterEach(() => {
+        config.onUnhandledError = null;
+      });
+
+      it('should not reset on refCount 0 if reset notifier errors before emitting any value', (done) => {
         const error = new Error();
+        let calls = 0;
+
+        config.onUnhandledError = spy((err) => {
+          calls++;
+          expect(err).to.equal(error);
+          if (calls === 2) {
+            done();
+          }
+        });
 
         rxTest.run(({ hot, cold, expectObservable, expectSubscriptions }) => {
           const source = hot('       ---1---2---3---4---(5 )---|');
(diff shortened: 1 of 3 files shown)


packages/rxjs/spec/operators/share-spec.ts
```diff
@@ -856,89 +855,86 @@ describe('share', () => {
...
-        const expected = '       ---1---(2 )------1---(2|)  ';
-        const subscription = '   ^------(- )----------(- )  ';
-        const firstPause = cold('       (- )---|            ');
-        const reset = cold('            (- )-r              ');
-        // reset: '                                   (- )-r'
-
-        const sharedSource = source.pipe(share({ resetOnComplete: () => reset, resetOnRefCountZero }));
-
-        const result = concat(sharedSource, firstPause, sharedSource);
+  it('should not call "resetOnRefCountZero" on error', () => {
```
benlesh, line 887: Everything under this section is really a whitespace change. I'm not sure why the diff is flipping out here. This was just from reorganizing things in the file I think.

packages/rxjs/spec/operators/share-spec.ts
```diff
@@ -810,9 +797,22 @@ describe('share', () => {
       });
     });
 
-    it('should not reset on refCount 0 if reset notifier errors before emitting any value', () => {
-      spyOnUnhandledError((onUnhandledError) => {
+    describe('when config.onUnhandledError is set', () => {
```
benlesh, line 800: Adding this `describe` block is what I think messed with the whitespace diff in other parts of the file.

packages/rxjs/spec/operators/share-spec.ts
```diff
@@ -856,89 +855,86 @@ describe('share', () => {
           expectObservable(result, subscription).toBe(expected);
           expectSubscriptions(source.subscriptions).toBe(sourceSubs);
         });
-
-        expect(onUnhandledError).to.have.been.calledOnce;
```
benlesh, line 860: These tests were errant, in that they were able to check to see if `onUnhandledError` was dispatched synchronously, which it was never supposed to be.

packages/rxjs/src/internal/Subscriber.ts
```diff
@@ -267,7 +266,7 @@ function createSafeObserver<T>(observerOrNext?: Partial<Observer<T>> | ((value:
  */
 function handleStoppedNotification(notification: ObservableNotification<any>, subscriber: Subscriber<any>) {
   const { onStoppedNotification } = config;
-  onStoppedNotification && timeoutProvider.setTimeout(() => onStoppedNotification(notification, subscriber));
+  onStoppedNotification && setTimeout(() => onStoppedNotification(notification, subscriber));
```
benlesh, line 269: One important bit is here.

packages/rxjs/src/internal/util/reportUnhandledError.ts
```diff
@@ -11,7 +10,7 @@ import { timeoutProvider } from '../scheduler/timeoutProvider';
  * @param err the error to report
  */
 export function reportUnhandledError(err: any) {
-  timeoutProvider.setTimeout(() => {
+  setTimeout(() => {
```
benlesh, line 13: The other important bit is here.

jakovljevic-mladen (approved):

PR #7351, opened 2023-10-03
feat(from): change from accepted arguments

ynoplanetashka2:
added support for function/async-function
as argument for "from"


josepot:
> added support for function/async-function as argument for "from"

https://rxjs.dev/api/index/function/defer

ynoplanetashka2:
@josepot 
_defer_ works with _observable's_.
as far as i know there is no convenient way to create an `observable` from `async-function`.
while u can create `observable` from `promise` with "from", your `observable` won't rerun `promise` on subscriptions.

PR #7358, opened 2023-10-25
chore: remove redundant files

Andarist:


operators/package.json
Andarist, line 1: with the monorepo support this alias directory is meant to be in the `packages/rxjs` so it's redundant in the root

package-lock.json
Andarist, line 1: from what I see Lerna is configured to using Yarn so that's your package manager of choice and thus this npm-specific lockfile is redundant

benlesh (approved):

benlesh:
Thank you, @Andarist

PR #7363, opened 2023-10-27
refactor: testing matcher and logging

benlesh:
Refactors the error matcher to accommodate the fact that we know the shape of the things we are testing. This is only a change to our own matchers and is not publicly exposed. This was tested by intentionally breaking several tests and checking outcomes. The debugging information provided by the assertion is also improved.

I can't believe I didn't spend some time doing this ages ago. There's still room for improvement here, but this does make things easier to debug.

image
image
image

### DIFF
--- a/packages/rxjs/spec/helpers/observableMatcher.ts
+++ b/packages/rxjs/spec/helpers/observableMatcher.ts
@@ -1,49 +1,175 @@
 import * as _ from 'lodash';
 import * as chai from 'chai';
+import { ErrorNotification, NextNotification, ObservableNotification } from 'rxjs';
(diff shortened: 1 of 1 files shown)


benlesh:
Rethinking this change. As most people will be using a deep equals algorithm that is less forgiving with notifications structures. (i.e. matching `{ kind: 'N', value: 1 }` with `{ kind: 'N', value: 1, error: undefined }`.

kwonoj:
fyi I made a custom matcher once: 

image
image

https://github.com/kwonoj/rx-sandbox/blob/master/src/assert/marbleAssert.ts

PR #7368, opened 2023-10-29
chore: migrate to eslint

43081j:
This just removes tslint and its associated packages, replacing them with eslint and `@typescript-eslint`.

Any custom rule config we had has been discarded so we can instead rely on the recommended configs from both eslint and tseslint.

let me know if you want this. if you want any specific rules too, just let me know.


packages/rxjs/spec/observables/from-spec.ts
```diff
@@ -7,7 +7,7 @@ import { ReadableStream } from 'web-streams-polyfill';
 import { observableMatcher } from '../helpers/observableMatcher';
 
 function getArguments<T>(...args: T[]) {
-  return arguments;
```
benlesh, line 10: This isn't correct. We actually wanted to test to make sure that `from(arguments)` worked (this is an old pattern that I suspect most people don't use), but it's really testing "ArrayLike", where `arguments` is the most well-used "ArrayLike".

benlesh (changes requested):
The only thing I saw in there is where we changed `getArguments`, which was literally trying to get an `Arguments` instance on purpose. I think if that's changed back, we're good to go with this.

I REALLY appreciate this PR.

43081j:
no worries, i've reverted that and added an ignore.

we may have also found a bug in the prefer-const eslint rule, or just a limitation. in that anywhere we do this pattern:

```ts
let x;

doThingLaterOn(() => {
  x();
});
doThingNow(() => {
  x = whatever;
});
```

it tells us to use `const` but we can't ofc. I just ignored those lines

benlesh (approved):

benlesh (changes requested):
Oops. It looks like we have a failing test. Womp womp.

packages/rxjs/spec/observables/fromEvent-spec.ts
```diff
@@ -352,7 +352,7 @@ describe('fromEvent', () => {
     };
 
     function selector(x: number, y: number, z: number) {
-      return [].slice.call(arguments);
+      return [].slice.call(x, y, z);
```
benlesh, line 355: This looks like an issue... it's not the same. Just leave the old code as is and add an ignore, I think.

43081j:
not sure why CI is failing... something to do with core-js in the dev site's build? 🤔 any idea?

benlesh:
@43081j I think if you rebase now, it may resolve the issue.

43081j:
sorry for the delay @benlesh , have rebased now 👍

43081j:
rebased again fyi

benlesh:
The failure here seems weird. I'm going to go ahead and merge this, and sort it out separately.

benlesh:
Thank you again, @43081j ... this is a very helpful PR

43081j:
no problem, thanks for taking time to review

i also couldn't figure the failure out since i couldn't repro it locally, will be interesting to find out what it was 👀

PR #7367, opened 2023-10-29
fix(docs): fix missing overloads in docs when overload count is less than 3

tmair:
**Description:**
This PR fixes an issue in the docs app, that does not display the overlays if the number of overloads is 2. The overload is printed at the top of the document.

**Related issue (if exists):** #5281

See also AIO Commit: https://github.com/angular/angular/commit/e0c736066cfd3fb4d784c2a335e6214ed6d6ec21

Issue #5281, opened 2020-01-30
Operator overload signatures are not shown in docs app


cartant:
### Description Of The Issue

In the docs app, overload signatures are shown for methods on `Observable` - e.g. five overload signatures are shown for `subscribe`.

However, overload signatures are not shown for operators. That makes documenting `shareReplay` somewhat difficult, as it has a number of signatures - including a signature that takes a single config object. The same problem exists for the config object signature added for `retry` in [this PR](https://github.com/ReactiveX/rxjs/pull/5280).

It's not clear what needs to be done to effect the showing of overload signatures for operators and I'm not sure what it is that effects them for `Observable#subscribe` - as  there doesn't seem to be anything specific to overloads in the TSDoc.

dzhavat:
It might sound funny but I just added a 3rd signature function to the `shareReplay` operator, rebuild the docs, launched them, and now the Overloads section is there. 

![image](https://user-images.githubusercontent.com/1096332/73575716-0d1da680-4479-11ea-8f30-b10c9f01fc88.png)

![image](https://user-images.githubusercontent.com/1096332/73575851-57068c80-4479-11ea-950e-36501f038c52.png)

I did the opposite with the `mergeMap` operator. Deleted a signature function to just two (from three) and the Overloads section is not there. So there's something with having a specific number of signatures before they appear in the docs.

cartant:
@dzhavat Nice find.

dzhavat:
Looks like my observation above is correct. Here's how the Overloads section is build :) 

https://github.com/ReactiveX/rxjs/blob/5170728f0262d65ab43937da9cbd621d6a5b4739/docs_app/tools/transforms/templates/api/function.template.html#L19

jakovljevic-mladen:
Closed in #7367.

PR #7369, opened 2023-10-30
fix(docs): fix missing overloads in docs when overload count is less than 3

tmair:
**Description:**
This PR fixes the display of overloads in the docs application if there are less than 3 overloads. 

Backport of: #7367

**Related issue (if exists):** #5281

### DIFF
--- a/docs_app/tools/transforms/templates/api/function.template.html
+++ b/docs_app/tools/transforms/templates/api/function.template.html
@@ -3,7 +3,7 @@
 {% extends 'export-base.template.html' -%}
 
 {% block overview %}
-{% if doc.overloads.length > 0 and doc.overloads < 3 -%}
+{% if doc.overloads.length > 0 and doc.overloads.length < 3 -%}
   {% for overload in doc.overloads -%}
     {$ memberHelpers.renderOverloadInfo(overload, 'function-overload', doc) $}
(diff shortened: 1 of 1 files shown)


jakovljevic-mladen:
@benlesh, now we're running into issue where CI fails on `7.x` branch. Error:

```
npm ERR! notsup Required: {"node":"^18.17.0 || >=20.5.0"}
npm ERR! notsup Actual:   {"npm":"8.19.4","node":"v16.20.2"}
```

How do we solve this?

BTW, change from this PR does not impact in any way anything that is tested during CI run, so I will probably merge it later. Just want to make sure you know that we're having issues with CI run.

benlesh:
> How do we solve this?

I'm not quite sure yet. I'll look into it. In the meantime, I think this can be merged.

benlesh:
@tmair thank you for this PR! BTW (and this isn't a big deal). When doing anything on the docs site, even if it's a "fix", just use `docs: whatever the change is` for the message. `fix(???): ???` will show up in the library's changelog.

tmair:
> @tmair thank you for this PR! BTW (and this isn't a big deal). When doing anything on the docs site, even if it's a "fix", just use `docs: whatever the change is` for the message. `fix(???): ???` will show up in the library's changelog.

Oh, I was not aware of this. I will try to remember this for any future PRs

Issue #7359, opened 2023-10-27
Logo on MPM website points to a 404


Lykos:
### Describe the bug

If you go to the [NPM site of RxJS](https://www.npmjs.com/package/rxjs), there is a broken image with the alt text "RxJS Logo". It points to https://github.com/reactivex/rxjs/blob/HEAD/docs_app/src/assets/images/logos/Rx_Logo_S.png which doesn't exist.

### Expected behavior

The logo should point to https://github.com/ReactiveX/rxjs/blob/0bd47eab10dec89f245b888f1f26e03cb36d2a78/apps/rxjs.dev/src/assets/images/logos/Rx_Logo_S.png which shows the incredibly cute logo of RxJS.

### Reproduction code

```typescript
Just open https://www.npmjs.com/package/rxjs and you'll see the problem.
```

### Reproduction URL

https://www.npmjs.com/package/rxjs

### Version

7.8.1

### Additional context

This is not an issue of the RxJS code itself but rather with its NPM configuration. If you tell me where this is configured, I can even try to fix it myself.

jakovljevic-mladen:
I don't know how it works.

The logo is now moved to `https://raw.githubusercontent.com/reactivex/rxjs/HEAD/apps/rxjs.dev/src/assets/images/logos/Rx_Logo_S.png`, but current stable version is now way off of the `HEAD` so pointing to `https://raw.githubusercontent.com/reactivex/rxjs/HEAD/docs_app/src/assets/images/logos/Rx_Logo_S.png` doesn't really work.

We should probably make it work to point to `7.x` branch, or at least to the latest stable version of the library.

Because it works with `https://raw.githubusercontent.com/reactivex/rxjs/7.x/docs_app/src/assets/images/logos/Rx_Logo_S.png` or `https://raw.githubusercontent.com/reactivex/rxjs/7.8.1/docs_app/src/assets/images/logos/Rx_Logo_S.png`.

Maybe @benlesh has some option to adjust on NPM?

benlesh:
No, we just need to publish 7.x again with the change. It's low-priority, but fixable with a publish.

benlesh:
[...]

PR #7371, opened 2023-11-03
in the npm websie the icon disappeared

giorgioGTelian:
**Description:**

**BREAKING CHANGE:** 


jakovljevic-mladen:
Hello @giorgioGTelian, thank you for the PR.

I don't think this is the right way of fixing this issue. This PR adds a completely new folder structure that does match what NPM does on their site. This is discussed in #7359 and the solution was proposed in [this comment](https://github.com/ReactiveX/rxjs/issues/7359#issuecomment-1784198119).

I'm gonna close this as the fix for NPM is not by modifying anything in the repo.

giorgioGTelian:
sure thank you for the clarification
i kinda suspected that was not the correct solution but i wanted to draw attention to it anyway.
 hope that wasn't too much of an hussle.
take care!

jakovljevic-mladen:
Not at all. You too.

PR #7374, opened 2023-11-10
Observable gets its own package - Stage 1

benlesh:
Okay, really this is "stage 2 or 3" as we've already done some work elsewhere.

This PR needs to be reviewed per-commit. It's really just moving things around. The goal here is to just move everything under `Observable.ts` that _required_ for observable. Right now that's a little more than you'd think. But we can shave it down after.

Again, review this per-commit so it's easier to understand what all changed and why. See the commit messages for more details.

**Note that I've moved `from` under the Observable.ts file** This is intentional. There is little to no chance that code using Observable, particularly with any operator will not _want_ from. The W3C Observable has `Observable.from`, and I'm going to propose that we move it back on. I've never seen a bundle that used Observable that DIDN'T include `from`. It's ubiquitous, and it's very likely that users of raw `Observable` will want to convert native types to observables.


packages/rxjs/src/internal/Observable.ts
```diff
@@ -466,3 +1008,375 @@ export class Observable<T> implements Subscribable<T> {
...
+  Own,
+  InteropObservable,
+  ArrayLike,
+  Promise,
+  AsyncIterable,
+  Iterable,
+  ReadableStreamLike,
+}
+
+export function getObservableInputType(input: unknown): ObservableInputType {
```
benlesh, line 1242: NEW: I've centralized the checking of `ObservableInput<T>` here. We had it in two places before, and order mattered. We also had to duplicate the error throwing behavior. This is now used in both `from` and `schedule`.

packages/rxjs/src/internal/Observable.ts
```diff
@@ -341,7 +883,7 @@ export class Observable<T> implements Subscribable<T> {
    * in the order they were passed in.
    */
   pipe(...operations: UnaryFunction<any, any>[]): unknown {
-    return pipeFromArray(operations)(this);
+    return operations.reduce(pipeReducer, this as any);
```
benlesh, line 886: Refactored more than a move: This is so small, I couldn't see a reason not to just inline it for now. the `pipeFromArray` was used in `Observable#pipe`, `pipe`, and `rx`, and we'll need to separate it up when they're in different packages.

packages/rxjs/src/internal/Observable.ts
benlesh, line 1: Yes, it's a big file now. Once we have an `@rxjs/observable` package, I'll move the "internal" stuff out of here, and reorganize a bit. BUT! I might still keep most things in one file, as all of the import/export logic adds some "death-by-a-thousand-cuts" bloat to bundled chunks.

niklas-wortmann (approved):
Went high level over it, looks good but quite a lot of changes :D

packages/rxjs/src/internal/Observable.ts
```diff
@@ -1241,3 +1232,28 @@ export function subscribeToArray<T>(array: ArrayLike<T>, subscriber: Subscriber<
...
+
+export enum ObservableInputType {
+  Own,
+  InteropObservable,
+  ArrayLike,
+  Promise,
+  AsyncIterable,
+  Iterable,
+  ReadableStreamLike,
+}
```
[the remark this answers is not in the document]
benlesh: No. This will never be public. I could switch this to numeric constants at some point, I suppose. It's an implementation detail.

PR #7380, opened 2023-11-20
chore: set primary node version and yarn v1 package manager

JamesHenry:
**Description:**

In getting started with the repo, my global yarn installation of latest v4 tried to upgrade the repo.

This PR adds a volta config so that those that use it to manager their node toolchain can benefit from automatic resolution of node and yarn, and also adds a `packageManager` definition for corepack users.

This should make it easier for first time contributors to get up and running faster.

### DIFF
--- a/.github/workflows/ci_main.yml
+++ b/.github/workflows/ci_main.yml
@@ -35,10 +35,8 @@ jobs:
       - name: rxjs dtslint
         run: yarn workspace rxjs dtslint
       - name: rxjs test:import
-        if: ${{ matrix.node == '18' || matrix.node == '20' }}
         run: yarn workspace rxjs test:import
       - name: rxjs test:esm
-        if: ${{ matrix.node == '18' || matrix.node == '20' }}
         run: yarn workspace rxjs test:esm
       - name: rxjs.dev build
         run: yarn workspace rxjs.dev build --prod
--- a/package.json
+++ b/package.json
@@ -14,11 +14,10 @@
     ]
   },
   "license": "Apache-2.0",
-  "packageManager": "yarn@1.22.21",
-  "volta": {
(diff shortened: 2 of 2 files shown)


.github/workflows/ci_main.yml
```diff
@@ -35,10 +35,8 @@ jobs:
       - name: rxjs dtslint
         run: yarn workspace rxjs dtslint
       - name: rxjs test:import
-        if: ${{ matrix.node == '18' || matrix.node == '20' }}
```
JamesHenry, line 38: I imagine these conditions originally distinguished these steps, but as it stands now we only run the jobs on v18 and v20 so the condition is redundant and I went ahead and cleaned them up

kwonoj (changes requested):
I believe setting engines field should be enough, if not we should set corepack related config only instead of having other 3rd party tools.

JamesHenry:
@kwonoj just to be clear, no third party tools are introduced here, purely config for volta users to benefit from if they already use the tool. It will be safely ignored by everything else and will not impact installs in any way. Do you still have an objection to it?

kwonoj:
> no third party tools are introduced here, purely config for volta users

I am well aware of volta and the what config those. I'm against our pkg manifest contains those tool specific configs. If there's a separate config file that's probably ok.

JamesHenry:
@kwonoj Ok I have replaced the volta config with an appropriate engines reference. Naturally this is only for the monorepo itself, rxjs users won't be impacted

kwonoj (approved):

JamesHenry:
Thanks for the approval @kwonoj are you able to approve the CI run?

JamesHenry:
@kwonoj following up on approving the CI run please

PR #7381, opened 2023-11-20
Remove Subject.create, refactor WebSocketSubject

benlesh:
An incremental refactor of `WebSocketSubject` to clean it up quite a bit.

Includes the following **BREAKING CHANGES**:

1. `WebSocketSubject` is no longer `instanceof Subject`. It really didn't make any sense to do this, it didn't leverage anything from `Subject` to begin with.
2. `Subject.create`, a long-deprecated API has been removed. If you recall, `Subject.create(observer, observable)` basically "glued" the observer to the observable.

This removes `AnonymousSubject`, which was never exported, and was overall weird.

### DIFF
--- a/packages/rxjs/spec/Subject-spec.ts
+++ b/packages/rxjs/spec/Subject-spec.ts
@@ -1,6 +1,5 @@
 import { expect } from 'chai';
 import { Subject, Observable, AsyncSubject, Observer, of, config, Subscription, Subscriber, noop, operate } from 'rxjs';
-import { AnonymousSubject } from 'rxjs/internal/Subject';
 import { delay } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from './helpers/observableMatcher';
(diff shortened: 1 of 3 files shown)


packages/rxjs/spec/Subject-spec.ts
```diff
@@ -448,104 +447,6 @@ describe('Subject', () => {
     expect(subject.observed).to.equal(false);
   });
 
-  it('should have a static create function that works', () => {
```
benlesh, line 451: Removed because Subject.create is no longer a thing.

packages/rxjs/spec/Subject-spec.ts
```diff
@@ -448,104 +447,6 @@ describe('Subject', () => {
...
-
-    expect(nexts).to.deep.equal(['a', 'b', 'c']);
-    expect(complete).to.be.true;
-    expect(error).to.be.a('undefined');
-
-    expect(output).to.deep.equal([1, 2, 3, 4, 5]);
-    expect(outputComplete).to.be.true;
-  });
-
-  it('should have a static create function that works also to raise errors', () => {
```
benlesh, line 500: Removed because Subject.create is no longer a thing.

packages/rxjs/spec/Subject-spec.ts
```diff
@@ -781,30 +682,3 @@ describe('Subject', () => {
     expect(results).to.deep.equal([1, 1, 2, 2, 'complete']);
   });
 });
-
-describe('AnonymousSubject', () => {
```
benlesh, line 785: Removed because AnonymousSubject and Subject.create is are longer things.

packages/rxjs/src/internal/observable/dom/WebSocketSubject.ts
```diff
@@ -151,47 +148,53 @@ const WEBSOCKETSUBJECT_INVALID_ERROR_OBJECT =
...
+
+  private _hasError = false;
+
+  private _error: any;
+
+  private _isComplete = false;
+
+  private _subscriberCounter = 0;
+
+  private _subscribers = new Map<number, Subscriber<Out>>();
```
ssilve1989, line 167: Should these be real private fields without the ugly `_` prefix instead of TS private?

ssilve1989 (approved):

packages/rxjs/src/internal/observable/dom/WebSocketSubject.ts
benlesh: Yeah. I think overall, RxJS is due for an overhaul in this area. The upside is we can't accidentally publicly expose any `#private` properties. The downside is you can't destructure them. Which is mostly a superficial downside.

benlesh:
Thank you for the review @ssilve1989! ❤️

packages/rxjs/src/internal/observable/dom/WebSocketSubject.ts
demensky: @benlesh, I might try to do this.

packages/rxjs/src/internal/observable/dom/WebSocketSubject.ts
```diff
@@ -151,47 +148,53 @@ const WEBSOCKETSUBJECT_INVALID_ERROR_OBJECT =
...
-          }
-        }
-      }
+    const userConfig = typeof urlConfigOrSource === 'string' ? { url: urlConfigOrSource } : urlConfigOrSource;
+    this._config = {
+      ...DEFAULT_WEBSOCKET_CONFIG,
+      // Setting this here because a previous version of this allowed
+      // WebSocket to be polyfilled later than DEFAULT_WEBSOCKET_CONFIG
+      // was defined.
+      WebSocketCtor: WebSocket,
```
demensky, line 181: Will this throw an exception if the `WebSocket` is not available globally?

`WebSocketCtor` becomes almost useless.

PR #7382, opened 2023-11-21
style: automatic type-only imports

demensky:
Automatically add `type` to `import`/`export` if the symbol is used only for type annotations and declarations.

* TypeScript: [Type-Only Imports and Export](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export).
* Eslint: [@typescript-eslint/consistent-type-imports](https://typescript-eslint.io/rules/consistent-type-imports/).

Fixes: https://x.com/BenLesh/status/1724808188988633490


jakovljevic-mladen:
Hehe, have I now completely missed the point?

It looks like `export type` creates an issue for us, but this PR is doing the `import type` 😄

However, I'd still like to know whether the docs app has issues with these changes.

demensky:
@jakovljevic-mladen, I checked the docs app and didn't find any problems with it.

Thanks to you, I also decided to add `export type`, checked it again and still found no problems. Apparently the problem is no longer relevant.

Example

P.S. I deleted `apps/rxjs.dev/src/generated` before starting the docs app build.

jakovljevic-mladen:
Actually, I don't think we had an issue with `ShareConfig` page being rendered, but with links that point to the `ShareConfig` page or any other exported type page that we have.

So, to test it, please open [`config`](https://rxjs.dev/api/index/const/config) page [locally](http://localhost:4200/api/index/const/config) and verify that link from [this line](https://github.com/ReactiveX/rxjs/blob/72bc92191ab959e27a969dc4476e14d95416573f/src/internal/config.ts#L5) works and that it points to [GlobalConfig](https://rxjs.dev/api/index/interface/GlobalConfig) page.

demensky:
@jakovljevic-mladen. Here's video proof. Although I would be very grateful if you also checked locally.

[Example](https://github.com/ReactiveX/rxjs/assets/10235949/0c819b44-dda4-4462-85f5-240130f4156e)

jakovljevic-mladen:
Cool, I think that should be good enough. I also checked pipelines and it seems that we don't have issues there as well, so I think it's fine.

packages/rxjs/spec-dtslint/Observable-spec.ts
```diff
@@ -1,4 +1,5 @@
-import { Observable, of, OperatorFunction, map, filter } from 'rxjs';
+import type { OperatorFunction} from 'rxjs';
```
benlesh, line 1:
```suggestion
import type { OperatorFunction } from 'rxjs';
```

packages/rxjs/.eslintrc.json
```diff
@@ -3,10 +3,15 @@
...
   ],
+  "parserOptions": {
+    "project": ["./tsconfig.json"]
+  },
   "rules": {
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/no-unused-vars": "off",
     "@typescript-eslint/no-unsafe-declaration-merging": "off",
+    "@typescript-eslint/consistent-type-imports": "warn",
+    "@typescript-eslint/consistent-type-exports": "warn",
```
benlesh, line 14: Question: Should these be "error" instead of warn?

benlesh (approved):

demensky:
@benlesh I almost managed to add `"error"`. I also discovered that eslint also checks `dist`. I will create a separate PR.

packages/rxjs/.eslintrc.json
demensky: #7391. I also discovered that eslint also checks dist. I will create a separate PR.

PR #7383, opened 2023-11-21
feat(rxjs.dev): add api list deprecated filter

daniel-melanson:
**Description:**

Adds a "Show deprecated" checkbox to optionally filter out deprecated items.

**BREAKING CHANGE:** N/A

**Related issue (if exists):** N/A

### DIFF
--- a/apps/rxjs.dev/src/app/custom-elements/api/api-list.component.spec.ts
+++ b/apps/rxjs.dev/src/app/custom-elements/api/api-list.component.spec.ts
@@ -42,6 +42,18 @@ describe('ApiListComponent', () => {
     });
   }
 
+  /**
+   * Expectation Utility: Assert that filteredSections has no items
+   *
+   * Subscribes to `filteredSections` and performs expectation within subscription callback.
+   */
+  function expectNoItems() {
+    component.filteredSections.subscribe(filtered => {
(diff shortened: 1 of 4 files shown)


benlesh:
This a cool idea. @niklas-wortmann or @jakovljevic-mladen can you please have a look?

jakovljevic-mladen (changes requested):
Hi @daniel-melanson, thank you a lot for this PR, it looks really cool and I think we should definitely have such option. However, we had it before, but it was for some reason removed and it was never restored back.

The original version had a dropdown select option, but that option was commented for some reason in the template. All other functionality around that option remained in the code (and probably in the tests). This commented code has been removed in #6936.

I would strongly suggest we go in that direction because dropdown select option is capable of having more options than we could have with a checkbox. Could you please review the code removed from the mentioned PR and try to include it in the current code base?

The code I'm referring to looked like this:

```
<!-- <aio-select (change)="setStatus($event.option)"
  [options]="statuses"
  [selected]="status"
  label="Status:">
</aio-select> -->
```

And instead of having status (in the `statuses` array in the component class):

```
{ value: 'security-risk', title: 'Security Risk' }
```

we could have:

```
{ value: 'stable', title: 'Stable' }
```

Because we probably won't ever introduce potentially dangerous code.

PR #7384, opened 2023-11-23
refactor: remove duplicate configuration code in Observable

Leehyunji0715:
## Description
Remove duplicate config code from the Observable.ts file and instead use the config.ts file.

## Changes
- Remove duplicate config code from Observable.ts file. (line 383 - 420)
- Update index.ts file
```typescript
/* Configuration */
export { config, GlobalConfig } from './internal/config.js';
```

### DIFF
--- a/.github/workflows/ci_main.yml
+++ b/.github/workflows/ci_main.yml
@@ -35,10 +35,8 @@ jobs:
       - name: rxjs dtslint
         run: yarn workspace rxjs dtslint
       - name: rxjs test:import
-        if: ${{ matrix.node == '18' || matrix.node == '20' }}
         run: yarn workspace rxjs test:import
       - name: rxjs test:esm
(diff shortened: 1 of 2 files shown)


packages/rxjs/src/index.ts
```diff
@@ -12,12 +12,13 @@
 /// <reference path="./testing/index.ts" />
 /* eslint-enable @typescript-eslint/triple-slash-reference */
 
+/* Configuration */
+export { config, GlobalConfig } from './internal/config.js';
```
jakovljevic-mladen, line 16: I believe this should be the other way around. We should probably remove `/internal/config.ts` file.

Leehyunji0715: I see, I will rollback the change and try to remove the `/internal/config.ts` file

PR #7392, opened 2023-11-30
chore(test): use fakeTimers with setTimeout in buffer, skipUntil and window

jsaguet:
**Description:**

After checking out the rxjs repo and running the unit tests, I noticed these 3 tests were always failing. (I'm running Windows 10 with Node 20)

The issue is probably that setTimeout does not necessarily execute at the specified delay.
I tried to reproduce the test in [StackBlitz](https://stackblitz.com/edit/rxjs-qa2f7v?file=index.ts) and I had the exact same issue when using timeouts so close to each other.

The PR updates these tests to use fake timers to make sure the tests work as expected.

### DIFF
--- a/packages/rxjs/spec/operators/buffer-spec.ts
+++ b/packages/rxjs/spec/operators/buffer-spec.ts
@@ -3,6 +3,7 @@ import { EMPTY, NEVER, throwError, of, Subject, interval } from 'rxjs';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
 import { expect } from 'chai';
+import * as sinon from 'sinon';
 
 /** @test {buffer} */
 describe('Observable.prototype.buffer', () => {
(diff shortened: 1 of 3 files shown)


pmoleri:
I had the same issue when running the tests on Windows 10 node@18.18.2.
It's very annoying to contribute when there are unrelated failing tests.

I came across this interesting stackoverflow question: https://stackoverflow.com/questions/65177373/odd-behavior-with-setinterval-in-node-js-windows-only-works-in-linux that seems to explain the issue.

`fakeTimers` seems to be the way to go if we don't want to slow down the tests to be consistent in different platforms.

benlesh:
Hi @jsaguet — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

PR #7415, opened 2023-12-20
chore: improve release script

benlesh:
+ Ensure the proper upstream is always selected
+ Display the options provided
+ Fixes a 1GB number to be a proper 1GB


benlesh:
cc @JamesHenry

PR #7416, opened 2023-12-27
chore: migrate to expecttype eslint rules

43081j:
This migrates to using an eslint plugin for `$ExpectType` and friends so we can do it as part of the lint script.

For now, `typescript` had to be added to the root project since the rxjs site is still on an older typescript which confuses eslint (since it ends up being hoisted by yarn).

In future, we should probably keep the two projects in sync, using the same typescript version. However, this will mean bumping the angular version in the rx site (which is its own chunky job).

**Questions**

- do we still need `tsconfig.types.json`? and `tsconfig.types.spec.json`. im not sure what use these have anymore, as we use a separate mocha tsconfig during testing anyway

### DIFF
--- a/packages/rxjs/spec-dtslint/operators/expand-spec.ts
+++ b/packages/rxjs/spec-dtslint/operators/expand-spec.ts
@@ -21,10 +21,6 @@ it('should support concurrent parameter', () => {
   const o = of(1, 2, 3).pipe(expand(value => of(1), 47)); // $ExpectType Observable<number>
 });
 
-it('should support a scheduler', () => {
-  const o = of(1, 2, 3).pipe(expand(value => of(1), 47, asyncScheduler)); // $ExpectType Observable<number>
-});
-
 it('should enforce types', () => {
   const o = of(1, 2, 3).pipe(expand()); // $ExpectError
 });
--- a/.github/workflows/ci_main.yml
+++ b/.github/workflows/ci_main.yml
@@ -30,8 +30,6 @@ jobs:
         run: yarn workspace rxjs build
       - name: rxjs test
         run: yarn workspace rxjs test
-      - name: rxjs dtslint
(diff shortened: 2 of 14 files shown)


43081j:
@benlesh fyi this seems to have also caught some leftover dtslint assertions for things that have since been removed

e.g. `every` supporting a `thisArg`

the jsdoc was also mentioning those params still, so i removed that in this PR.

pinging you because you made those changes iirc, so want to be sure this is right

packages/rxjs/spec-dtslint/operators/expand-spec.ts
```diff
@@ -1,51 +1,47 @@
...
-});
-
-it('should enforce scheduler type', () => {
-  const o = of(1, 2, 3).pipe(expand(value => of(1), 47, 'foo')); // $ExpectError
-});
-
-it('should support union types', () => {
-  const o = of(1).pipe(expand(x => typeof x === 'string' ? of(123) : of('test'))); // $ExpectType Observable<string | number>
-});
+import { of, asyncScheduler } from 'rxjs';
```
43081j, line 1: the whitespace change seems to be that the original file had CRLF. my editor read the `editorconfig` and changed it to `LF`

43081j:
any chance you can take a look at this @benlesh ? 👀 

low priority ofc, assuming you do want the change!

benlesh:
Hi @43081j — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

PR #7419, opened 2023-12-28
fix: remove node export map entries

43081j:
These entries cause node to resolve to commonjs in all cases, rather than respecting the consumer's module `type`.

For example, a consumer with `"type": "module"` will still resolve to the commonjs entrypoint since node will choose the `node` entry before the `default` entry.

### DIFF
--- a/packages/rxjs/package.json
+++ b/packages/rxjs/package.json
@@ -16,43 +16,36 @@
   "exports": {
     ".": {
       "types": "./dist/types/index.d.ts",
-      "node": "./dist/cjs/index.js",
       "require": "./dist/cjs/index.js",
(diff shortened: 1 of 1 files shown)


43081j:
this has now uncovered that the published ESM doesn't work in node anyhow, since rxjs' `type` is not `"module"`.

it means we resolve to the esm entrypoint successfully now (and didn't before), but the resolved file is treated as commonjs.

we would have to rename all the esm output to `*.mjs` as far as i can tell

43081j:
abandoning this since it sounds like there's a plan to do it as part of a larger change further ahead in future

will just have to manage for now

PR #7427, opened 2024-01-11
fix(rxjs): removed unnecessary peer deps

### LINKED ISSUE #7426: rxjs@8.0.0-alpha.13 has peer deps?! Oops
Apparently, when I migrated to a monorepo, something automated added peer deps of `typescript` and `@types/node`... that's completely silly. Those are dev dependencies. There's no reason anyone consuming `rxjs` needs those.

🤦

benlesh:
resolves #7426


packages/rxjs/package.json
```diff
@@ -193,9 +193,5 @@
       "pre-commit": "lint-staged",
       "commit-msg": "validate-commit-msg"
     }
-  },
-  "peerDependencies": {
```
benlesh, line 197: This is what I set out to remove.

apps/rxjs.dev/package.json
```diff
@@ -77,9 +77,10 @@
     "@types/jasmine": "~3.6.0",
     "@types/jasminewd2": "^2.0.3",
     "@types/lunr": "^2.3.3",
-    "@types/node": "^20.6.3",
+    "@types/node": "^20.11.0",
     "@types/svgo": "^1.3.3",
     "@types/trusted-types": "^2.0.7",
+    "@types/yamljs": "^0.2.34",
```
benlesh, line 83: Ran into issues with TypeScript building the dev app and had to add some dev deps and update types in a few files.

PR #7428, opened 2024-01-13
chore: update nx to latest, simplify config

JamesHenry:
Contains all the latest fixes and DX improvements relevant to `nx release`

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -13,6 +13,7 @@
     }
   },
   "release": {
+    "projects": ["rxjs"],
     "releaseTagPattern": "{version}",
     "changelog": {
       "git": {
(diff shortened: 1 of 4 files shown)


nx.json
```diff
@@ -25,15 +26,10 @@
       },
       "projectChangelogs": true
     },
-    "groups": {
```
JamesHenry, line 28: A single group is no longer required, we can use the projects shorthand and not need the nested structure (and not need to give it a name)

scripts/release.js
```diff
@@ -56,8 +56,6 @@ const yargs = require('nx/node_modules/yargs');
 
     const { workspaceVersion, projectsVersionData } = await releaseVersion({
       specifier: options.version,
-      // stage package.json updates to be committed later by the changelog command
-      stageChanges: true,
```
JamesHenry, line 60: This is essentially always needed for the common case, so it's true by default now

nx.json
```diff
@@ -13,27 +13,19 @@
     }
   },
   "release": {
+    "projects": ["rxjs"],
     "releaseTagPattern": "{version}",
     "changelog": {
-      "git": {
```
JamesHenry, line 18: Committing and tagging by default after/as part of the changelog generation step just makes a lot sense based on more and more real world usage, so we updated the latest nx to make it so and can streamline the config. As before the logs, in both dry-run and non dry-run, explain when and what git commands are/would be running

PR #7430, opened 2024-01-18
feat: add @rxjs/observable package and update rxjs to use it

JamesHenry:
This PR adds a new package `@rxjs/observable`, and updates the existing `rxjs` package to import from it. It contains all low level values and types related to `Observable`.


benlesh:
Thank you again, @JamesHenry!

PR #7439, opened 2024-02-07
feat(fromEvent): complete the stream after the first value on one time events

Deku-nattsu:
**Description:**
A small quality of life change to fromEvent when the once option is supplied, it makes sense to automatically close the stream after the first emission since the observable will only emit once. this will remove the need to manually unsubscribe every time you work with one time events.

**BREAKING CHANGE:** 


Deku-nattsu:
oops i forgot about the case where you can pass an array of targets, i am setting this to draft for the moment

PR #7444, opened 2024-02-20
fix(animationFrameScheduler): some tasks are never flushed and sometimes it breaks completely

pmoleri:
`animationFrameScheduler` has a few issues that makes it miss tasks and even start accumulating tasks that never get flushed.

 - It's clearing the _scheduled in every flush, but some of the flushes are not for the next scheduled frame, they're regular async timeouts.
 - because it might be picking the incorrect flushId it may flush tasks before the next frame
 - after flushing and recycling the id it's clearing the `_scheduled` even though current id might be a different one.

**Related issues:**
 - #7018
 - #6891
 - #7196
   already fixed but similar cause

Failing tests are included in a independent commit and then fixed.

Note:
The changes could also be ported to AsapScheduler for consistency (also I think it's more clear).
However, I couldn't reproduce the issues there because:
 - AsapScheduler already contains one of the fixes
 - In AsapScheduler I couldn't schedule a task to happen before the `_scheduled` one.

### DIFF
--- a/spec/schedulers/AnimationFrameScheduler-spec.ts
+++ b/spec/schedulers/AnimationFrameScheduler-spec.ts
@@ -165,11 +165,15 @@ describe('Scheduler.animationFrame', () => {
     let runDelayed = false;
     let runFirst = false;
     animationFrame.schedule(() => {
-      animationFrame.schedule(() => { runFirst = true; });
       animationFrame.schedule(() => {
         if (!runDelayed) {
           done(new Error('Delayed action did not run'));
-        } else if (!runFirst) {
+          return;
+        }
+        runFirst = true;
+      });
+      animationFrame.schedule(() => {
+        if (!runFirst) {
           done(new Error('First action did not run'));
         } else {
           done();
@@ -177,7 +181,9 @@ describe('Scheduler.animationFrame', () => {
       });
 
       // This action will execute before the next frame because the delay is less than the one of the frame
-      animationFrame.schedule(() => { runDelayed = true; }, 1);
+      animationFrame.schedule(() => {
+        runDelayed = true;
+      }, 1);
     });
   });
 
--- a/spec/schedulers/AnimationFrameScheduler-spec.ts
+++ b/spec/schedulers/AnimationFrameScheduler-spec.ts
@@ -141,6 +141,46 @@ describe('Scheduler.animationFrame', () => {
     }
   });
 
+  it('should schedule next frame actions from a delayed one', (done) => {
+    animationFrame.schedule(() => {
+      animationFrame.schedule(() => { done(); });
+    }, 1);
+  });
+
+  it('should schedule 2 actions for a subsequent frame', (done) => {
+    let runFirst = false;
+    animationFrame.schedule(() => {
+      animationFrame.schedule(() => { runFirst = true; });
+      animationFrame.schedule(() => {
+        if (runFirst) {
+          done();
+        } else {
+          done(new Error('First action did not run'));
+        }
+      });
+    });
+  });
+
+  it('should handle delayed action without affecting next frame actions', (done) => {
+    let runDelayed = false;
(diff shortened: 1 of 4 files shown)


pmoleri:
Hi @benlesh. I noticed that you fixed 7196 for asapScheduler, but animationFrameScheduler has similar issue, and worse even (reason explained in this PR description).
Any chance that you can review this PR? It also includes tests that make it fail without the fix.

BlueCat0:
any update yet?

pmoleri:
@BlueCat0 Unfortunately it seems that no one had the time to review it.
Once you understand the code (which takes a while) the bug is pretty clear and the resolution too.

To avoid the bug altogether, never use animationFrameScheduler with a delay, you can use the asyncScheduler for that purpose. In fact they would behave exactly the same, as animationFrameScheduler doesn't really schedule into a frame when delay is used.

Note that using animationFrameScheduler with a delay in one place may impact in a different place.

ChristophWieske:
@pmoleri Your PR is open for quite some time. Has there been another PR which made yours obsolete? Or is this still waiting for a review?

I encounter issues with `animationFrameScheduler` in 7.8.1 which are maybe related.

So, is the bug you are trying to fix here still unfixed?

pmoleri:
Hi @ChristophWieske 

The issue is actually quite serious, it can affect any Angular app combining animationFrameScheduler with a delay other than 0.
It's very hard to troubleshoot, it leads to anomalous behavior or crashes depending on the case.

In the 7.x branch the files have not been touched in 3 years, so I'm positive that the issue is still there.
I've included tests to make it fail (if you omit the fix).

In master the files have been moved around and the imports changed but the code is exactly the same.
So it must happen too.

pmoleri:
Hi @benlesh. This PR fixes real bugs and it has tests to confirm the bug and the fix.
The changes are small and I consider them sound.

I made the PR to 7.x, but I could re-target to master and backport later.
I can make the changes that you consider best. Thanks.

trxcllnt (approved):

spec/schedulers/AnimationFrameScheduler-spec.ts
```diff
@@ -141,6 +141,46 @@ describe('Scheduler.animationFrame', () => {
...
+          done(new Error('Delayed action did not run'));
+        } else if (!runFirst) {
+          done(new Error('First action did not run'));
+        } else {
+          done();
+        }
+      });
+
+      // This action will execute before the next frame because the delay is less than the one of the frame
+      animationFrame.schedule(() => { runDelayed = true; }, 1);
```
trxcllnt, line 180: I'm curious which order in which these actions are expected to execute.

Is `runFirst` expected to be true when `runDelayed` is set to true? e.g.:

```suggestion
      animationFrame.schedule(() => {
        if (!runFirst) {
          done(new Error('First action did not run'));
        } else {
          runDelayed = true;
        }
      }, 1);
```

pmoleri: No, it's the opposite. The action queued with a delay smaller than animationFrame runs first.
That's how animationFrameScheduler is implemented, actions with delay other than zero are scheduled purely based on time and not actual animation frames.
I'll push a change to make it more clear.

pmoleri: There. I pushed the change. Now each scheduled actions asserts the previous one. Instead of having everything in the last one.

sherlock1982:
Would be nice to have it fixed as it is used by a number of libs causing random issues left and right.

benlesh:
Thank you for your contribution, @pmoleri. Apologies for the long wait.

pmoleri:
Thank you @benlesh. I created this PR against 7.x.
Do you want me to create another one against master?

Also, are you planing a new 7.x release soon?

PR #7451, opened 2024-02-29
chore: publish correct package types

benlesh:
- Also removes generated .tshy directory from source control


.gitignore
```diff
@@ -13,6 +13,7 @@ typings/
 # Generated
 dist/
 .tshy-build
+.tshy
```
benlesh, line 16: the files under `.tshy` are apparently generated on every build.

JamesHenry: Ok cool, I purely based it on the tshy repo itself having that folder checked in https://github.com/isaacs/tshy

@isaacs sorry for the direct ping but do you have the official view on this please? Is there any disadvantage to ignoring it?

isaacs: I usually check it in because I use the tsconfig files tshy generates in order to run tests and build docs in CI. But you can check it in or not, as you see fit. Whenever you build, tshy will re-generate those files as needed anyhow.

JamesHenry: Thanks @isaacs!

PR #7472, opened 2024-05-10
Improve type inference with `zen-observable`

forivall:
**Description:**

While trying to consume a `zen-observable` into rxjs via `from`, the types weren't being inferred. I added a facscimille of the zen-observable types as `CompatObservable` to the tests to figure out how it was failing, and discovered the solution of adding a `InteropSubscribable` with an overload, which allowed the types to infer properly. This might be a shortcoming of typescript itself, but, in the meantime, this seems to work well.

### DIFF
--- a/spec-dtslint/helpers.ts
+++ b/spec-dtslint/helpers.ts
@@ -32,3 +32,35 @@ export const g$ = of(new G());
 export const h$ = of(new H());
 export const i$ = of(new I());
 export const j$ = of(new J());
+
+interface Subscription {
+  closed: boolean;
+  unsubscribe(): void;
+}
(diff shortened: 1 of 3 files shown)


benlesh:
Hi @forivall — thank you for the time, care, and thoughtfulness you put into this pull request. I’m sorry we weren’t able to respond in a reasonable timeframe. Since the codebase and surrounding context have moved on and this hasn’t seen recent activity, I’m going to close it for now. That’s a reflection of our delayed triage, not a lack of appreciation for your contribution. If the change is still relevant, please feel welcome to open a fresh pull request against the current codebase. Thank you again for contributing to RxJS and for your patience.

— Ben

PR #7487, opened 2024-06-28
Replace polyfill.io with a Cloudflare equivalent

tnfssc:
**Description:**

This PR replaces the `polyfill.io` with a Cloudflare equivalent because `polyfill.io` can no longer be trusted.

> polyfill.io, a popular JavaScript library service, can no longer be trusted and should be removed from websites.

Source: [https://blog.cloudflare.com/automatically-replacing-polyfill-io-links-with-cloudflares-mirror-for-a-safer-internet](https://blog.cloudflare.com/automatically-replacing-polyfill-io-links-with-cloudflares-mirror-for-a-safer-internet)

### DIFF
--- a/apps/rxjs.dev/src/index.html
+++ b/apps/rxjs.dev/src/index.html
@@ -84,7 +84,7 @@
   <script>
     if (!Object.assign) {
       // polyfill other non-evergreen browsers in a blocking way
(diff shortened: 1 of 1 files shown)


jakovljevic-mladen (approved):
LGTM. Thanks for fixing this.

PR #7499, opened 2024-09-13
docs: mention sample and withLatestFrom in each other's docs

robintown:
withLatestFrom and sample are pretty similar: the latter just skips emitting when the sampled observable hasn't emitted anything new. I recently found myself using sample when I should've been using withLatestFrom, and would've benefited from it being linked in the docs.


jakovljevic-mladen (approved):
Yeah, they kinda are the same, so we can include them in each other's `@see` section. LGTM, thanks. We will merge this once we solve issue with pipelines (it may take some time, sorry, we don't have anyone available right now to fix it).

Dustin4444 (approved):

PR #7502, opened 2024-09-19
fix: make `observerOrNext` in `Observable#subscribe` nullable

imjordanxd:
**Description:**
This PR makes the `observerOrNext` param in ``Observable#subscribe`. No logic change is required, as `subscribe` is already overloaded with `null`, so the logic to check for this is already in place. This is also aligned with what's in the V8 branch.

- mobx-utils: https://github.com/mobxjs/mobx-utils/issues/324

### DIFF
--- a/src/internal/Observable.ts
+++ b/src/internal/Observable.ts
@@ -64,7 +64,7 @@ export class Observable<T> implements Subscribable<T> {
     return observable;
   }
 
-  subscribe(observerOrNext?: Partial<Observer<T>> | ((value: T) => void)): Subscription;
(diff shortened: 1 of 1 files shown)


src/internal/Observable.ts
```diff
@@ -64,7 +64,7 @@ export class Observable<T> implements Subscribable<T> {
     return observable;
   }
 
-  subscribe(observerOrNext?: Partial<Observer<T>> | ((value: T) => void)): Subscription;
+  subscribe(observerOrNext?: Partial<Observer<T>> | ((value: T) => void) | null): Subscription;
```
imjordanxd, line 67: for some reason, all the build steps fail locally for me 🤷

PR #7503, opened 2024-09-29
docs(types): typos

lor3k:
**Description:**
Typos in docs

**BREAKING CHANGE:** 


jakovljevic-mladen (approved):
LGTM. It looks like we have some issues with pipelines, so we can't merge this until they are resolved.

JamesHenry:
Not only is there the type error on the latest TS workflow, in the latest Node v20 (`20.18.0`) there is now a failing unit test.

I debugged and fixed it here: https://github.com/ReactiveX/rxjs/pull/7512

PR #7511, opened 2024-10-29
chore: migrate nx to latest v20, bump typescript-eslint to v7

JamesHenry:
**Description:**

I ran `nx migrate latest` to bring the repo up to date with the latest nx and `@nx/*` packages, including some minor configuration optimizations.

I bumped typescript-eslint packges to v7 as well. We can do the migration to ESLint v9 and typescript-eslint v8 in a separate PR.

This PR can go in once #7512 is merged which fixes all existing issues with the repo's CI that are unrelated to this.

PR #7512, opened 2024-10-29
chore: fix latest compile TS and Node 20 workflow failures

JamesHenry:
**Description:**

In PR #7511 I discovered that Node 18 tests are working fine, but it looks like the latest Node 20 (`20.18.0`) which the Node 20 job is configured to use, has actually implemented `FinalizationRegistry` which was added to the tests 4 years ago (!) but has not actually been in use because of it not being present globally in Node until now.

This is obviously a very surprising discovery that this test code has not actually ever been executed in the last 4 years, and in reading the MDN entry for `FinalizationRegistry` (a thing I was not familiar with at all), it seems like something we'd want to avoid in deterministic tests across multiple environments.

Therefore I replaced its usage entirely with a `WeakRef` which should hopefully cover the desired logic on Node v20.

(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry)

The PR then ultimately also fixes the long standing TS compilation issues with latest TS.

### DIFF
--- a/packages/rxjs/tsconfig.json
+++ b/packages/rxjs/tsconfig.json
@@ -9,7 +9,7 @@
     "moduleResolution": "node",
     "stripInternal": true,
     "noEmit": true,
-    "lib": ["esnext", "dom"],
+    "lib": ["es2022", "dom"],
     "target": "esnext",
     "baseUrl": ".",
     "paths": {


JamesHenry:
Ha, so this did indeed fix v20, but has caused v18 to fail... I will see if I can repro locally. Perhaps it just needs a few retries on the GC check...

JamesHenry:
It turns out `global.gc` does not exist by default in Node 18. We could either:

- skip the tests in Node 18
- always run node explicitly for the mocha tests and pass it the `--expose-gc` flag which sets `global.gc`. Again, it's set by default in Node 20.

I decided to go with the latter, but let me know if you would prefer the former.

JamesHenry:
NOTE: This will still be blocked until the type error on latest TS is resolved

PR #7582, opened 2025-11-16
remove a bunch of 'any' types

dynst:
**Description:**

Replace a bunch of `any` types with either `unknown` or a more specific type.

But two are harder to replace, changing them causes type errors:

https://github.com/ReactiveX/rxjs/blob/c15b37f81ba5f5abea8c872b0189a70b150df4cb/packages/observable/src/observable.ts#L262
https://github.com/ReactiveX/rxjs/blob/c15b37f81ba5f5abea8c872b0189a70b150df4cb/packages/observable/src/observable.ts#L498

```
  Type 'Subscriber<T>' is not assignable to type 'Subscriber<unknown>'.
    Type 'unknown' is not assignable to type 'T'.
```

**BREAKING CHANGE:** 

Part of the 8.x roadmap in #6367 

>  Replace *all* external uses of `any` with `unknown`. (This could likely be non-breaking, FWIW.)

### DIFF
--- a/packages/observable/src/observable.spec.ts
+++ b/packages/observable/src/observable.spec.ts
@@ -358,10 +358,10 @@ describe('Observable', () => {
         for await (const value of source) {
           results.push(value);
         }
-      } catch (err: any) {
+      } catch (err: unknown) {
         thrownError = err;
       }
-
+      expect(thrownError instanceof Error).to.be.true;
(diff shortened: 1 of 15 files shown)


dynst:
Can this make it into the next major release? https://github.com/ReactiveX/rxjs/releases/tag/9.0.0-beta.0

Issue #7536, opened 2025-03-03
7.8.2 introduces breaking changes


ilyakonrad:
### Describe the bug

Calling `.next()` with no arguments like this
```
new Observable((observer) => {
  observer.next()
}
```
shows error
```
Expected 1 arguments, but got 0.ts(2554)

Subscriber.d.ts(42, 10): An argument for 'value' was not provided.
```

### Expected behavior

No breaking changes in patch version updates.

### Version

7.8.2

### Additional context

Duplicate of https://github.com/ReactiveX/rxjs/issues/7534.

hakimio:
I guess now you have to explicitly specify the generic type:
```ts
new Observable<void>((observer) => {
  observer.next()
}
```

Either the default generic should be set to `void` (`class Observable<T = void>`) for the `Observable` or this breaking change should be postponed until `v8`.

dgadelha:
The [changelog mentions this as part of the 7.8.2 patch release](https://github.com/ReactiveX/rxjs/blob/e5351d02e225e275ac0e497c7b66eaa5f0c88791/CHANGELOG.md?plain=1#L7):

>  **subscriber:** strict type signature for next method ([#7172](https://github.com/reactivex/rxjs/issues/7172)) ([0e2ef5e](https://github.com/reactivex/rxjs/commit/0e2ef5e1142699b028bc3624aae9b24c3e3aaccf))

Whereas the PR explicitly has on the description:

> **BREAKING CHANGE:** `Subscriber.prototype.next` now takes strict argument.

Yet, it was released in a patch version.

hakimio:
@dgadelha 
Breaking changes do not belong in patch releases. This should have been postponed for next major or the typing introduced in a non-breaking way like I mentioned above with a good default.

dgadelha:
@hakimio I'm with you! I just pointed out that the author of the PR explicitly mentioned this was a breaking change, yet it was still released as a patch version. I’ve updated my comment to make that clearer :)

samwightt:
Jumping in again, this is introducing type errors in our codebase where there were none previously. This is absolutely a breaking change and should've been included in a major version bump, along with guidance on how to remediate.

SvetlaGeorgieva:
Yep, and the docs are incorrect at the moment, as they still mention usage of `next()` with no parameters, which is no longer allowed: https://rxjs.dev/guide/subject#void-subject

We also experienced the breaking changes in our project.

PR #7585, opened 2025-12-05
fix 'new Subject()' and 'new Observable()'

dynst:
**Description:**

As pointed out in #7536, the docs for rxjs 7+ have specifically said this since #5307:

https://github.com/ReactiveX/rxjs/blob/c15b37f81ba5f5abea8c872b0189a70b150df4cb/apps/rxjs.dev/content/guide/subject.md#L368

https://github.com/ReactiveX/rxjs/blob/c15b37f81ba5f5abea8c872b0189a70b150df4cb/apps/rxjs.dev/content/guide/subject.md#L377

But the `<T = void>` was removed for unexplained reasons in #5430 even though the docs were never updated.

**BREAKING CHANGE:** 

Code that depends on `new Subject()` returning `Subject<unknown>` will break.

This also had to fix tests missing type parameters from #3449 #5750 #6216 #6478

### DIFF
--- a/spec-dtslint/Subject-spec.ts
+++ b/spec-dtslint/Subject-spec.ts
@@ -2,11 +2,10 @@ import { Subject } from 'rxjs';
 
 describe('Subject', () => {
   it('should handle no generic appropriately', () => {
-    const s1 = new Subject(); // $ExpectType Subject<unknown>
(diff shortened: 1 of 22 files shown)


dynst:
The alternative to this PR is just fixing the documentation in #7586

PR #7590, opened 2026-01-24
feat(rxjs.dev): handle api list fetch failure with fallback

djenadiamirredouane:
Description:
Handle API list fetch failures on rxjs.dev: log the error and fall back to an empty list to avoid a blank page.

BREAKING CHANGE: None

Related issue (if exists):
N/A


djenadiamirredouane:
@jakovljevic-mladen can you take a look please ?

PR #7608, opened 2026-06-23
fix(takeLast): prevent RangeError when count is Infinity

JSap0914:
## Problem

`takeLast(Infinity)` throws a `RangeError: Invalid array length` at subscription time:

```ts
// Throws: RangeError: Invalid array length
of(1, 2, 3).pipe(takeLast(Infinity)).subscribe(console.log);
```

The cause is line 50 of `takeLast.ts`:

```ts
let ring = new Array<T>(count);  // new Array(Infinity) → RangeError
```

JavaScript forbids creating an array with a non-integer or infinite length argument.

## Fix

Use a growable array (`[]`) when `count` is non-finite. The ring-buffer index arithmetic already handles `Infinity` correctly because `n % Infinity === n` for every finite `n`, so no further changes to the logic are required:

```ts
let ring: T[] = isFinite(count) ? new Array<T>(count) : [];
```

After the fix, `takeLast(Infinity)` behaves like `takeLast(n)` where `n >= source.length`: it buffers every value and emits them all on completion.

## Tests

Two new tests added to `takeLast-spec.ts`:
- Synchronous assertion that no error is thrown and all values are emitted
- Marble test verifying the observable shape matches the source

> *This fix was developed with AI assistance.*

### DIFF
--- a/packages/rxjs/spec/operators/takeLast-spec.ts
+++ b/packages/rxjs/spec/operators/takeLast-spec.ts
@@ -1,4 +1,5 @@
 import { takeLast, mergeMap } from 'rxjs/operators';
+import { expect } from 'chai';
 import { of } from 'rxjs';
 import { TestScheduler } from 'rxjs/testing';
 import { observableMatcher } from '../helpers/observableMatcher';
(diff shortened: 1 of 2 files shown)


ajafff:
IMO `takeLast(Infinity)` should be equal to `identity`, i.e. not changing the source sequence at all, so it needs no buffer.

On the other hand that would change the existing behavior of only emitting the items on complete to emitting immediately...

benlesh (approved):
Thank you for your contribution, @JSap0914.

PR #7635, opened 2026-08-28
fix(takeLast): handle Infinity count

yowainwright:
## Description

`takeLast(Infinity)` currently allocates `new Array(Infinity)`, which throws `RangeError: Invalid array length`.

This update assigns a growable buffer only for `Infinity`, while preserving the existing finite-count ring buffer and non-positive early-complete behavior. 
- The regression test verifies that values are still emitted on completion.

## Related issue (if exists)

- Refs #7608


lbesecker195 (approved):
Traced the ring-buffer math for `amount = Infinity`: `counter % Infinity` is `counter` for every finite counter, `start` resolves to `0`, and `total` becomes `counter`, so the completion loop emits every buffered value in order. Only the allocation needed to change. Test covers it. LGTM.

